PR #8100, opened 2019-05-02
Fixes #26720 - CV Publish: copy correct errata

parthaa:
This commit tweaks the cv publish algorithm slightly to copy errata
selectively.

Prior to this commit errata that got copied to the destination repo were
based on the RPMs in the destination repo.
This lead to issues in dealing with rpms that belong to 2 errata. For
example if we had a content view filter that says "Include only Errata
between X and Y days" and same rpm belonged to 2 errata, then both
errata would automatically get included in the destination repo even
if one of the errata was issued much after Y or before X.

This commit addresses that issue by only copying errata that matched the
filters. We were already using this logic to copy rpms belonging to
errata. Now we copy the appropriate errata the same way.


jturel:
@parthaa can you add some test coverage here?

parthaa:
Closing this this PR for  https://github.com/Katello/katello/pull/8112

PR #8101, opened 2019-05-03
Fixes #26734 - Sync plan run without repos

sjha4:
Issue:

When a sync plan without products or repositories is run, it fails with the error: "No syncable repositories found for selected products and options."

As per the issue, we shouldn't fail the run in this case. Instead, we should let the job succeed without actually running the repository sync action.


sjha4:
[test katello]

PR #8104, opened 2019-05-07
Fixes #26752 - remove react-router package

glekner:
adding `react-router-dom` to core broke nightlies, we think it may be related to `react-router` package. @sharvit 
`react-router` can be safely removed because react-router-dom uses all its exports

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,6 @@
     "react-ellipsis-with-tooltip": "^1.0.7",
     "react-helmet": "^5.2.0",
     "react-redux": "^5.0.6",
-    "react-router": "^4.2.0",
     "react-router-bootstrap": "0.24.4",
     "react-router-dom": "^4.2.2",
     "redux": "^3.7.2",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -72,7 +72,6 @@
     "react-ellipsis-with-tooltip": "^1.0.7",
     "react-helmet": "^5.2.0",
     "react-redux": "^5.0.6",
-    "react-router": "^4.2.0",
     "react-router-bootstrap": "0.24.4",
     "react-router-dom": "^4.2.2",
```
tbrisker: Shouldn't we remove this as well and use the react-router-dom provided by core?

waldenraines: Yes I believe so

jturel: Makes sense to me. I tested without and it's fine (did a fresh npm install)

tbrisker:
ping @glekner @jturel what's the status here?

jturel:
> ping @glekner @jturel what's the status here?

waiting for react-router-dom removal

jturel:
[test katello]

package.json
sharvit: I believe it's still needed for dev-env, perhaps it should move to be a `devDependency`

PR #8110, opened 2019-05-14
Fixes #26794 - Multi-entitlement missing on sub. details

jturel:
This was missing due to backend changes and UI changes from several releases ago.

To test, just load up the details page for any subscription and it'll probably show 'false' for the Multi-entitlement field which would have been missing before.

To see a 'true' value - you'll need a subscription manifest with a variety of subscriptions and one of them is likely to have it. One example is "Red Hat Enterprise Linux Server, Premium (Physical or Virtual Nodes)"

### DIFF
--- a/app/models/katello/subscription.rb
+++ b/app/models/katello/subscription.rb
@@ -42,6 +42,10 @@ module Katello
       pools.any?(&:recently_expired?)
     end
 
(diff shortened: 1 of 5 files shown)


jturel:
[test katello]

jturel:
[test katello]

PR #8111, opened 2019-05-14
Fixes #26685 - Show syspurpose info on sub. details

jturel:
We should show System Purpose information associated with subscriptions on their details page. This PR adds that and calls it out specifically in a new appropriately named section since it's a named feature.

**To test with this PR checked out:**

- Create a custom product, for example, 'PurposeProduct'.
- Set some syspurpose attributes on it by connecting to your local postgres candlepin db:

```
select uuid from cp2_products where name = 'PurposeProduct';
insert into cp2_product_attributes(name, value, product_uuid) values ('addons', 'Test Addon1,Test Addon2', 'REPLACEME');
insert into cp2_product_attributes(name, value, product_uuid) values ('role', 'Test Role', 'REPLACEME');
insert into cp2_product_attributes(name, value, product_uuid) values ('usage', 'Development', 'REPLACEME');
```

Also set a service level from the rails console:

```
sub = Katello::Pool.find(...).subscription
sub.support_level = "Premium"
sub.save
```

- Restart the tomcat service to refresh the product cache `systemctl restart tomcat`

Go to the details page for this subscription (Content -> Subscriptions) and you should see all of the above information as you entered it. The fields are named to identically match what we show on content host details so there's no confusion.


jturel:
Thanks @jjeffers & @sjha4 !

PR #8112, opened 2019-05-14
Fixes #26720 - CV Publish: copy correct errata

parthaa:
This commit makes the following changes to the cv publish process
Before This Commit:
Errata would get copied over to the destination repo if ANY of the
packages belonging to the errata were copied.

After this Commit:
Errata would get copied over to the destination repo only if ALL
of the packages belonging to the errata were copied.

This commit also includes changes to hide modular errata in a content
view filter (since all modules and module rpms are copied over, all the
modular errata will also get copied automatically.)


app/models/katello/erratum.rb
```diff
@@ -110,6 +120,10 @@ def self.list_filenames_by_clauses(repo, clauses)
           where(statement).pluck(:filename)
     end
 
+    def modular?
+      ::Katello::Erratum.modular.where(:id => id).exists?
+    end
```
jlsherrill: why is this method needed?  what exactly is the intent?

parthaa: Good call. Removing it. I was planning to show if an errata has modules or not. But that can be done in another way if needed so removing it.

app/models/katello/repository.rb
```diff
@@ -265,30 +265,40 @@ def product_type
...
+        errata_with_package_counts.keep_if { |id| errata_with_package_counts[id] == errata_with_packages_in_library[id] }
+        unless errata_with_package_counts.empty?
+          partial_errata = self.errata.where("#{Katello::Erratum.table_name}.id NOT IN (?)", errata_with_package_counts.keys)
+        end
       end
+
+      partial_errata
     end
 
     def empty_errata!
```
jlsherrill: should this method get renamed?

jlsherrill:
few things, but overall looks good. now to do some testing

jlsherrill:
Tested a few scenarios and I'm happy with the results. Just a few small comments

parthaa:
[test katello]

PR #6366, opened 2016-10-01
Fixes #16764: Return archived content view version repositories

ehelms:
Updates the repositories API to return the archived repositories
when a content view version ID is supplied instead of returning
all repositories across all environments. If a user specifies an
environment or the library parameter, the repositories returned
should be limited to that environment or library.

PR #8114, opened 2019-05-16
Fixes #26301 - Fetch correct rpm counts in cv version details

sjha4:
**To test:**

1. Create a repo with some content.
2. Add repo to cv and publish
3. Remove some content from repo
4. Re-publish repo
5. Go to CV > Version >  Yum Repositories
6. You should see the correct counts for the specific versions.

Most of the API changes come from the PR: https://github.com/Katello/katello/pull/6366 which tackles returning archived repo when environment is not specified.

### DIFF
--- a/app/controllers/katello/api/v2/repositories_controller.rb
+++ b/app/controllers/katello/api/v2/repositories_controller.rb
@@ -79,6 +79,7 @@ module Katello
     param :file_id, String, :desc => N_("Id of a file to find repositories that contain the file")
     param :ostree_branch_id, String, :desc => N_("Id of an ostree branch to find repositories that contain that branch")
     param :library, :bool, :desc => N_("show repositories in Library and the default content view")
+    param :archived, :bool, :desc => N_("show archived repositories")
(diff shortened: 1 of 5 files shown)


app/controllers/katello/api/v2/repositories_controller.rb
```diff
@@ -132,22 +131,23 @@ def index_relation_product(query)
     def index_relation_content_view(query)
       if params[:content_view_version_id]
         query = query.where(:content_view_version_id => params[:content_view_version_id])
-        query = Katello::Repository.where(:id => query.select(:library_instance_id)) if params[:library]
+        query = query.archived unless ::Foreman::Cast.to_bool params[:library] || params[:environment_id]
+        query = Katello::Repository.where(:id => query.select(:library_instance_id)) if ::Foreman::Cast.to_bool(params[:library])
```
sjha4: Feel like the meaning of params[:library] changes with this..This would suggest that without if we are not looking for libraries in all environments, we return archived repos. 
Leaning towards a separate archived param to make things less confusing..

sjha4:
[test katello]

jturel:
ACK. Looks good from the code side. Deferring to testing by @jjeffers for functionality.

PR #8115, opened 2019-05-16
Fixes #26530 - Don't reset Add Ons on UI updates

jturel:
System Purpose Addons were being reset under certain conditions when updating a content host via the UI. The `scope.purposeAddonsList` was being initialized as an empty list until the addons selector was expanded which caused the empty list to be sent to the server if the addons list was never expanded or modified.

I've changed the code so that addons are only sent to the server if they've potentially (meaning the addons list was opened but not necessarily modified in any way) been updated.

**To test:**

**Old behavior:**
- Set up a custom product with some addons similar to #8111 
- From the details page of a content host, set one or more addons
-  **Refresh the page**
- Update the host in some way but don't modify or open the purpose addons selector
- Addons will be cleared out

**New behavior:**
- Apply this PR and repeat the above steps for the content host. The addons should be preserved.
- Extra credit: try to break things by unsetting some or all addons, updating the host in other ways, etc. Everything should work as expected.

### DIFF
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/details/content-host-details.controller.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/details/content-host-details.controller.js
@@ -32,7 +32,6 @@ angular.module('Bastion.content-hosts').controller('ContentHostDetailsController
         $scope.defaultRoles = ['Red Hat Enterprise Linux Server', 'Red Hat Enterprise Linux Workstation', 'Red Hat Enterprise Linux Compute Node'];
         $scope.defaultServiceLevels = ['Self-Support', 'Standard', 'Premium'];
 
-        $scope.purposeAddonsList = [];
         $scope.purposeAddonsCount = 0;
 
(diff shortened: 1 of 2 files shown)


engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/details/content-host-details.controller.js
```diff
@@ -186,7 +185,7 @@ angular.module('Bastion.content-hosts').controller('ContentHostDetailsController
                 });
 
                 angular.forEach(addOns, function (addOn) {
-                    $scope.purposeAddonsList.push({"name": addOn, "selected": $scope.host.subscription_facet_attributes.purpose_addons.indexOf(addOn) > -1});
+                    $scope.purposeAddonsList.push({"name": addOn, "selected": purposeAddons.indexOf(addOn) > -1});
```
jturel: This is just a cleanup to reference an already defined variable referencing the same values

jturel:
[test katello]

PR #8116, opened 2019-05-16
Fixes #26810 - Don't run AutoAttach on every host update

jturel:
A while back host update was changed to trigger auto-attach at the time it was being enabled(#4444). Since then the code has regressed such that any host UI update will trigger an auto-attach which is not intended. It's especially bad for use cases like System Purpose where the user will want to update 1-4 attributes before running auto-attach manually.

If the ::Update action isn't given a params hash (such as during a host checkin/fact update) the consumer params are taken from the subscription facet. If autoheal is true then an auto attach is performed. This PR prevents that from happening and will once again only trigger auto-attach on host update if the flag is being set to true.

**To test**

- check out this PR
- register a content host to your devel box (vagrant up katello-devel)
- from the details UI for the host, perform some updates and check Monitor->Tasks dynflow console for the relevant tasks. There should not be a planned action for AutoAttachSubscriptions.
- from the command line of your client, disable auto attach: `subscription-manager auto-attach --disable`. This should not trigger an auto attach (verify same as above)
- now, enable auto-attach `subscription-manager auto-attach --enable` - you should see that AutoAttachSubscriptions has ran during the host update

### DIFF
--- a/app/lib/actions/katello/host/update.rb
+++ b/app/lib/actions/katello/host/update.rb
@@ -10,12 +10,14 @@ module Actions
           sequence do
             host.content_facet.save! if host.content_facet
 
+            auto_attach_enabled_via_checkin = consumer_params.try(:[], 'autoheal')
+
             if host.subscription_facet
               consumer_params ||= host.subscription_facet.consumer_attributes
               cp_update = plan_action(::Actions::Candlepin::Consumer::Update, host.subscription_facet.uuid, consumer_params)
             end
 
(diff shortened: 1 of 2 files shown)


jturel:
[test prprocessor]

parthaa:
Tested. Works well. APJ

sjha4:
[test katello]

sjha4:
[test katello]

jturel:
[test katello]

PR #8117, opened 2019-05-17
Fixes #26811 - Select all host installable errata

parthaa:
This commit lets you select "all" installable errata on content hosts
errata page. Prior to this one was able to only select the errata on the
first page. This commit tries to address that by facilitating bulk
params for errata apply.
In addition it provides functionality in the UI to facilitate the same.

### DIFF
--- a/app/controllers/katello/api/v2/host_errata_controller.rb
+++ b/app/controllers/katello/api/v2/host_errata_controller.rb
@@ -86,7 +86,7 @@ module Katello
         search_errata = @host.content_facet.installable_errata
         search_errata = search_errata.search_for(bulk_params[:included][:search])
         if @errata.any?
-          ::Katello::Erratum.where(errata_id: @errata).or(::Katello::Erratum.where(errata_id: search_errata))
+          @errata = ::Katello::Erratum.where(errata_id: @errata).or(::Katello::Erratum.where(errata_id: search_errata))
         else
           @errata = search_errata
         end
--- a/app/controllers/katello/api/v2/host_errata_controller.rb
+++ b/app/controllers/katello/api/v2/host_errata_controller.rb
@@ -86,13 +86,13 @@ module Katello
         search_errata = @host.content_facet.installable_errata
         search_errata = search_errata.search_for(bulk_params[:included][:search])
         if @errata.any?
-          ::Katello::Erratum.where("errata_id in (?) OR errata_id in (?)", @errata, search_errata)
+          ::Katello::Erratum.where(:errata_id: @errata).or(::Katello::Erratum.where(:errata_id: search_errata))
         else
           @errata = search_errata
         end
       end
 
-      @errata = @errata.where('errata_id not in (?)', bulk_params[:excluded][:ids]) unless bulk_params[:excluded][:ids].blank?
+      @errata = @errata.where.not(:errata_id: bulk_params[:excluded][:ids]) unless bulk_params[:excluded][:ids].blank?
 
       if bulk_params[:included][:ids].blank? && bulk_params[:included][:search].nil?
         fail HttpErrors::BadRequest, _("No errata have been specified.")
--- a/app/controllers/katello/api/v2/host_errata_controller.rb
+++ b/app/controllers/katello/api/v2/host_errata_controller.rb
@@ -72,6 +72,34 @@ module Katello
       respond_for_async :resource => task
     end
 
+    def find_bulk_errata_ids(bulk_params)
+      #works on a structure of param_group bulk_params and transforms it into a list of errata_ids
+      bulk_params[:included] ||= {}
+      bulk_params[:excluded] ||= {}
+      @errata = []
+
+      unless bulk_params[:included][:ids].blank?
+        @errata = @host.content_facet.installable_errata.where(:errata_id => bulk_params[:included][:ids])
+      end
+
+      if bulk_params[:included][:search]
+        search_errata = @host.content_facet.installable_errata
+        search_errata = search_errata.search_for(bulk_params[:included][:search])
+        if @errata.any?
+          ::Katello::Erratum.where("errata_id in (?) OR errata_id in (?)", @errata, search_errata)
+        else
+          @errata = search_errata
+        end
+      end
+
+      @errata = @errata.where('errata_id not in (?)', bulk_params[:excluded][:ids]) unless bulk_params[:excluded][:ids].blank?
+
+      if bulk_params[:included][:ids].blank? && bulk_params[:included][:search].nil?
+        fail HttpErrors::BadRequest, _("No errata have been specified.")
+      end
+      @errata.pluck(:errata_id)
+    end
+
     protected
 
     def index_relation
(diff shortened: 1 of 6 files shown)


parthaa:
UI - Behavior

apply-errata

parthaa:
Also important to check the api endpoint, since now it accepts bulk parameters . Look at the https://github.com/Katello/katello/pull/8117/files#diff-462c66df3ff934dde90914065e2aabd5 unit tests for various  search options

jturel:
[test katello]

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/content/content-host-errata.controller.js
```diff
@@ -155,6 +153,12 @@ angular.module('Bastion.content-hosts').controller('ContentHostErrataController'
             }
         };
 
+        errataNutupane.enableSelectAllResults();
+        // if ($location.search()['select_all']) {
```
jturel: Can this commented block go away?

parthaa: removed :). Good call

app/controllers/katello/api/v2/host_errata_controller.rb
```diff
@@ -17,6 +17,16 @@ def resource_class
...
 
+    def_param_group :bulk_errata_ids do
+      param :included, Hash, :required => true, :action_aware => true do
+        param :search, String, :required => false, :desc => N_("Search string for erratum to perform an action on")
+        param :ids, Array, :required => false, :desc => N_("List of errata ids to perform an action on, (ex: RHSA-2019:1168)")
+      end
+      param :excluded, Hash, :required => true, :action_aware => true do
+        param :ids, Array, :required => false, :desc => N_("List of errata ids to exclude and not run an action on, (ex: RHSA-2019:1168)")
+      end
+    end
```
johnpmitsch: Can included and excluded be used together? might be a good thing to mention in the API docs

app/controllers/katello/api/v2/host_errata_controller.rb
```diff
@@ -61,6 +72,34 @@ def applicability
...
+
+      unless bulk_params[:included][:ids].blank?
+        @errata = @host.content_facet.installable_errata.where(:errata_id => bulk_params[:included][:ids])
+      end
+
+      if bulk_params[:included][:search]
+        search_errata = @host.content_facet.installable_errata
+        search_errata = search_errata.search_for(bulk_params[:included][:search])
+        if @errata.any?
+          ::Katello::Erratum.where("errata_id in (?) OR errata_id in (?)", @errata, search_errata)
```
johnpmitsch: Couldn't you combine @errata and search_errata if they are both arrays and use `::Katello::Erratum.where(errata_id: combined_array)`?

There is also the `.or` active record method that would also work

app/controllers/katello/api/v2/host_errata_controller.rb
```diff
@@ -61,6 +72,34 @@ def applicability
...
+        else
+          @errata = search_errata
+        end
+      end
+
+      @errata = @errata.where('errata_id not in (?)', bulk_params[:excluded][:ids]) unless bulk_params[:excluded][:ids].blank?
```
johnpmitsch: why not use AR's `.not`?

```ruby
@errata.where.not(errata_id: bulk_params[:excluded][:ids])
```

johnpmitsch:
The selecting works fine, I'm getting an error actually applying the errata, but I don't think it is from your PR (good ole taxonomies) `Invalid single-table inheritance type: Location is not a subclass of Taxonomy`

I'll dig more tomorrow. only a couple comments but it looks good!

app/controllers/katello/api/v2/host_errata_controller.rb
parthaa: I like this idea. However there could be a 1000 errata ids matching a search criteria, using the query as above will give AR the opportunity to be smart and do something like ```errata_id in (select errata_id from .... where(..search..) )``` . As opposed to hydrating the 1000 ids in a query

parthaa: AR support OR now. Going to use that

parthaa:
[test katello]

app/controllers/katello/api/v2/host_errata_controller.rb
```diff
@@ -61,6 +72,34 @@ def applicability
...
+          @errata = ::Katello::Erratum.where(errata_id: @errata).or(::Katello::Erratum.where(errata_id: search_errata))
+        else
+          @errata = search_errata
+        end
+      end
+
+      @errata = @errata.where.not(errata_id: bulk_params[:excluded][:ids]) unless bulk_params[:excluded][:ids].blank?
+
+      if bulk_params[:included][:ids].blank? && bulk_params[:included][:search].nil?
+        fail HttpErrors::BadRequest, _("No errata have been specified.")
```
johnpmitsch: "No errata has been specified"

app/controllers/katello/api/v2/host_errata_controller.rb
johnpmitsch: Here is what I would suggest based on the behavior I am seeing:

```diff
diff --git a/app/controllers/katello/api/v2/host_errata_controller.rb b/app/controllers/katello/api/v2/host_errata_controller.rb
index 398808c..0c34460 100644
--- a/app/controllers/katello/api/v2/host_errata_controller.rb
+++ b/app/controllers/katello/api/v2/host_errata_controller.rb
@@ -18,11 +18,11 @@ module Katello
     end
 
     def_param_group :bulk_errata_ids do
-      param :included, Hash, :required => true, :action_aware => true do
+      param :included, Hash, :desc => N_("Errata to exclusively include in the action"), :required => true, :action_aware => true do
         param :search, String, :required => false, :desc => N_("Search string for erratum to perform an action on")
         param :ids, Array, :required => false, :desc => N_("List of errata ids to perform an action on, (ex: RHSA-2019:1168)")
       end
-      param :excluded, Hash, :required => true, :action_aware => true do
+      param :excluded, Hash, :desc => N_("Errata to explicitly exclude in the action. All other applicable errata will be included in the action, unless an include paramete
         param :ids, Array, :required => false, :desc => N_("List of errata ids to exclude and not run an action on, (ex: RHSA-2019:1168)")
       end
     end
```
(with better formatting of course)

johnpmitsch:
Everything is working well! Just a suggestion on the API and a typo

johnpmitsch:
@parthaa I'm thinking about the structure of these params, I wonder if this is the best way to structure the API? It's a bit confusing looking at the docs since there is already an `errata_ids` param, and now there is an `include` where I can also pass in errata ids?

What about something like this?

- `errata_ids` - the existing errata ids param
- `search` - a query to search for applicable errata
-  `exclude` - errata ids that can be passed in to exclude from the search query

This seems to fit the UI use case. select all would use the search query, and if they are unselected, they can be added to 'exclude'. Let me know missing something, but I do think these API params are confusing as is and we should explore other options to simplify them

parthaa:
> Found something else, just sending `exclude` is raising an error.

addressed. even added unit test for this case.

parthaa:
> 'm thinking about the structure of these params, I wonder if this is the best way to structure the API? It's a bit confusing looking at the docs since there is already an `errata_ids` param, and now there is an `include` where I can also pass in errata ids?

I deprecated the ```errara_ids``` parameter. We are forced to use the nutupane's selectAll parameters for the bulk apply and hence that format was chosen.

johnpmitsch:
@parthaa it's working well! I think this may be a thing with bastion, but I am still getting asked if I want to select all the items when there is less than the per-page amount:

```
All 3 items on this page are selected. Select all 3.
```

We can file a separate issue for this, but figured I would bring it up! :)

PR #8118, opened 2019-05-17
Fixes #26825 - Allow some DMI UUID duplicates on registration

jturel:
It turns out that when dmidecode can't determine a UUID there are some possible substitutions it will make -'Not Present' and 'Not Settable'

https://github.com/nima/python-dmidecode/blob/92f117b13b3ef2506bcf9c15ff1e151206c6faee/src/dmidecode.c#L456

This PR allows those possible values to exist as duplicates since it's an unavoidable situation. I also added a check for an empty string which shouldn't happen and may be unnecessary to look for.

**To test**

- register the katello-client forklift box to your env
- alter the hostname of the client so it doesnt conflict with the existing registration
```
hostnamectl set-hostname whatever.example.com
```
- override the dmi.system.uuid fact to check for these cases:
```
[root@katello-client-c vagrant]# cat /etc/rhsm/facts/uuid.facts 
{"dmi.system.uuid": ""}
```
- modify the local DB to match your overridden value from the rails console
```
fv = ::FactValue.where(host_id: ::Host.where(name: 'katello-client-a.jturel.example.com')).where(fact_name_id: Katello::RhsmFactName.find_by(name: 'dmi::system::uuid')).first
=> #<FactValue id: 26717, value: "Not Settable", fact_name_id: 71, host_id: 174, updated_at: "2019-04-17 21:13:58", created_at: "2019-04-05 02:56:10">

fv.update_attributes(value: "Not Present")
```

- clean and register the system with the new hostname and overridden fact
```
subscription-manager clean
subscription-manager register
```
- Registration should succeed and you can find the duplicate hosts with the query in the code here

### DIFF
--- a/app/models/katello/host/subscription_facet.rb
+++ b/app/models/katello/host/subscription_facet.rb
@@ -29,6 +29,8 @@ module Katello
 
       attr_accessor :facts
 
+      DMI_UUID_ALLOWED_DUPS = ['', 'Not Settable', 'Not Present'].freeze
+
       def host_type
         host_facts = self.host.facts
         host_facts["virt::host_type"] || host_facts["hypervisor::type"]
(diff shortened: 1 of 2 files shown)


jturel:
@evgeni do you an environment to test this in? It won't merge cleanly into a 6.5 but it should be straightforward

evgeni:
I have a prod box with a few hundred users, something with dogs ;)

But if you remind me next week, I can setup a reproducer outside prod too :)

jturel:
> I have a prod box with a few hundred users, something with dogs ;)
> 
> But if you remind me next week, I can setup a reproducer outside prod too :)

Not ringing any bells ;)

I tested in my devel environment but I wanted to make sure it worked in the scenario where you encountered the problem just to be sure. Will ping you next week!

PR #8122, opened 2019-05-22
Fixes #26855 - don't define params that are in the group already

evgeni:
the resource group already lists name, label and description, no need to
add them explicitly to the endpoint


evgeni:
test failure unrelated

jlsherrill:
[test katello]

jlsherrill:
[test katello]

jlsherrill:
Thanks @evgeni

PR #8123, opened 2019-05-22
Fixes #26239 - Index pulp2 refactor and pulp3

sjha4:
Refactoring and implementing as per design discussion with Justin.
To Do:
- [x] Test with yum metadata files
- [x] Rebase on top of contents_changed flow
- [x] Refactor pulp_batch_unit to be used for all and repository.
- [x] Tests
- [x] Artifact sha details with https://pulp.plan.io/issues/4827


app/services/katello/pulp3/file_unit.rb
```diff
@@ -0,0 +1,44 @@
...
+      end
+
+      def self.pulp_data(href)
+        content_unit = SmartProxy.pulp_master!.pulp3_api.content_file_files_read(href)
+        content_unit.as_json
+      end
+
+      def update_model(model)
+        custom_json = {}
+        custom_json['checksum'] = backend_data['sha256']
```
sjha4: Based on this https://pulp.plan.io/issues/4827 , we should be able to pick up sha 256 once the bindings are updated.

sjha4:
[test katello]

app/models/katello/repository.rb
```diff
@@ -750,12 +750,12 @@ def index_linked_repo
       end
     end
 
-    def index_content(full_index = false)
+    def index_content(_full_index = false)
```
jlsherrill: we can likely remove this full_index option now

app/services/katello/pulp/docker_tag.rb
```diff
@@ -3,6 +3,36 @@ module Pulp
...
+              model_class.where(:pulp_id => unit['_id']).first_or_create
+            end
+            service = self.new(model.pulp_id)
+            service.backend_data = unit
+            service.update_model(model)
+            pulp_ids << model.pulp_id
+          end
+        end
+        import_all pulp_ids unless pulp_ids.blank?
+      end
```
jlsherrill: can you do most of this in a call to 'super' ?

app/services/katello/pulp/pulp_content_unit.rb
```diff
@@ -22,10 +22,143 @@ def self.pulp_data(uuid)
...
+            break if (response.blank? && criteria[:skip] != 0)
+            response = block.call
+            yielder.yield response
+            criteria[:skip] += page_size
+          end
+        end
+      end
+
+      def self.import_all(pulp_ids = nil, options = {})
+        index_repository_association = options.fetch(:index_repository_association, true) && model_class.manage_repository_association
```
jlsherrill: can't we remove the index_repository_assocation option?

app/services/katello/pulp/pulp_content_unit.rb
```diff
@@ -22,10 +22,143 @@ def self.pulp_data(uuid)
...
+                repo_unit_id[repo_pulp_id] << unit_json['_id']
+              end
+            end
+          end
+
+          repo_unit_id.each do |repo_pulp_id, unit_pulp_ids|
+            sync_repository_associations(Katello::Repository.find_by(:pulp_id => repo_pulp_id), :pulp_ids => unit_pulp_ids, :additive => additive)
+          end
+        end
+      end
```
jlsherrill: we're only supporting indexing a content unit and associating it with a single repository, so this can probably be simplified?

app/services/katello/pulp3/repository.rb
```diff
@@ -135,7 +135,11 @@ def ssl_remote_options
       end
 
       def lookup_version(href)
-        pulp3_api.repositories_versions_read(href)
+        pulp3_api.repositories_versions_read href
+      rescue PulpcoreClient::ApiError => e
+        Rails.logger.error "Exception when calling RepositoriesApi->repositories_versions_read: #{e}"
+        nil
+      end
```
jlsherrill: i'm curious why this was needed?

app/services/katello/pulp3/repository/file.rb
```diff
@@ -91,6 +91,13 @@ def get_distribution(href)
           raise e if e.code != 404
           nil
         end
+
+        def version_details
+          pulp3_api.repositories_versions_read(repo.version_href)
+        rescue PulpFileClient::ApiError => e
+          raise e if e.code != 404
+          nil
```
jlsherrill: curious why this was needed?

jjeffers: @jlsherrill I saw the same thing used for a similar read operation in the original create PR (for distributions_file_file_read ?). Why check for and trap exceptions for specific pulp operations?

app/services/katello/pulp3/repository.rb
sjha4: Probably don't need to catch exceptions here and allow method to fail in case there are errors.

sjha4:
Couple questions:
1. If the import_all just creates content unit records and doesn't add repository associations, is it really required?
2. Based on 1, do we need the reimport task. Do we refactor it to loop over all repositories and call index on each?

jlsherrill:
> **Couple questions:**
> 
>     1. If the import_all just creates content unit records and doesn't add repository associations, is it really required?

Yes, but only for the use case where we add some attribute to be indexed to a model (say add a new field to our errata model), and want to import that into the katello db for all errata.  Its not common but does happen on occasion, and we put it in an upgrade rake task for one specific model.

> 
>     2. Based on 1, do we need the reimport task. Do we refactor it to loop over all repositories and call index on each?

We likely don't need the task for content objects.   The current 'reimport' task rarely solves any problems a user has, and i think is often used to just 'try' to fix problems.  I think we should just hold off and see if it ever becomes necessary, it should be very simple to add if so.  

> 
> Will tackle test failures after the next round of review.. smile
Great questions!

app/services/katello/pulp/pulp_content_unit.rb
sjha4: The additive option seemed unnecessary after all the changes. Now we are getting ids to add and ids to remove, forming queries and executing them.

app/services/katello/pulp3/repository/file.rb
jlsherrill: it makes sense for deletes IMO, but that is a good question, why it was  being done for reads originally.  I guess looking over that specific case, it was handling the situation where a distribution had been deleted in pulp, but we were still referencing it.  

In these cases, I don't think it can occur until we start deleting repository versions (if we ever do)

app/models/katello/concerns/pulp_database_unit.rb
```diff
@@ -55,85 +55,47 @@ def pulp_data(pulp_id)
...
-                    where repository_id = #{source_repo.id} and pulp_id not in (select pulp_id
-                    from #{self.table_name} where repository_id = #{dest_repo.id})"
-
+      def import_for_repository(repository)
+        pulp_ids = []
+        service_class = SmartProxy.pulp_master!.content_service(content_type)
+        service_class.pulp_units_batch_for_repo(repository).each do |units|
+          units.each do |unit|
+            unit = unit.with_indifferent_access
+            unit['_id'] = unit['_href'] ? unit['_href'] : unit['_id']
```
jlsherrill: This logic won't work because pulp2 units have an '_href' as well, but we don't use it.  What about doing something like 'service_class.unit_identifier' and have that return '_id' or '_href' ?

app/models/katello/concerns/pulp_database_unit.rb
```diff
@@ -55,85 +55,47 @@ def pulp_data(pulp_id)
...
 
-        process_block = lambda do |units|
+      # Don't know if this makes any sense anymore if we do not update repository associations?
+      def import_all(pulp_ids = nil)
+        service_class = SmartProxy.pulp_master!.content_service(content_type)
+        service_class.pulp_units_batch_all(pulp_ids).each do |units|
           units.each do |unit|
             unit = unit.with_indifferent_access
             model = Katello::Util::Support.active_record_retry do
               self.where(:pulp_id => unit['_id']).first_or_create
```
jlsherrill: shouldn't you use 'service_class.unit_identifier' here instead of '_id' ?

app/models/katello/concerns/pulp_database_unit.rb
```diff
@@ -55,85 +55,47 @@ def pulp_data(pulp_id)
...
-
+      def import_for_repository(repository)
+        pulp_ids = []
+        service_class = SmartProxy.pulp_master!.content_service(content_type)
+        service_class.pulp_units_batch_for_repo(repository).each do |units|
+          units.each do |unit|
+            unit = unit.with_indifferent_access
+            unit['_id'] = unit[service_class.unit_identifier]
+            model = Katello::Util::Support.active_record_retry do
+              self.where(:pulp_id => unit['_id']).first_or_create
```
jlsherrill: instead of storing the resulting identifier in '_id', would it make sense to just refer to 'service_class.unit_identifier' when you need to ?

app/models/katello/concerns/pulp_database_unit.rb
```diff
@@ -55,85 +55,47 @@ def pulp_data(pulp_id)
...
-      def import_all(pulp_ids = nil, options = {})
-        index_repository_association = options.fetch(:index_repository_association, true) && self.manage_repository_association
-        service_class = SmartProxy.pulp_master!.content_service(content_type)
+      def unit_id_field
+        "#{self.name.demodulize.underscore}_id"
+      end
 
-        process_block = lambda do |units|
+      # Don't know if this makes any sense anymore if we do not update repository associations?
+      def import_all(pulp_ids = nil)
```
jlsherrill: testing this, when uploading an rpm, it doesn't seem to get associated to the repository (the rpm count doesn't increase).   I think its because filtered_index_content.rb calls import_all() and passes in the unit ids, but that no longer associates them to the repository.

sjha4: Yes..We will need to handle that for upload content for rpm and deb if we remove the repository association part from import_all..
We could pass an optional parameter repository to import_all and then call
` sync_repository_associations(repository, :pulp_ids => pulp_ids) ` after importing units as we do in import_for_repository. We could use that in filtered_index_content.rb.

sjha4: Pushed changes based on the last few comments in a separate commit https://github.com/Katello/katello/pull/8123/commits/f15c8adc87215bfa5f4ce28a96c14e6b30f0237f

jlsherrill:
So now when i upload an rpm to a yum repository it clears out all the existing indexed data, and only shows 1 rpm in the repo (I think this might have been what 'additive' was used for) :)

sjha4:
> So now when i upload an rpm to a yum repository it clears out all the existing indexed data, and only shows 1 rpm in the repo (I think this might have been what 'additive' was used for) :)

Aah..I think I created a repo and immediately synced it before uploading. Everything works under 5 minutes :) I'll add the addtive option back..

app/models/katello/concerns/pulp_database_unit.rb
```diff
@@ -55,85 +55,50 @@ def pulp_data(pulp_id)
...
-      # Import all units of a single type and refresh their repository associations
-      def import_all(pulp_ids = nil, options = {})
-        index_repository_association = options.fetch(:index_repository_association, true) && self.manage_repository_association
-        service_class = SmartProxy.pulp_master!.content_service(content_type)
+      def unit_id_field
+        "#{self.name.demodulize.underscore}_id"
+      end
 
-        process_block = lambda do |units|
+      # Don't know if this makes any sense anymore if we do not update repository associations?
```
sjha4: One final push to remove this comment! :stuck_out_tongue:

PR #8127, opened 2019-05-23
Fixes #26857 - Pings pulp3 services before actions

parthaa:
Adds bindings for services checking before anu pulp3 actions. This
commit pings pulp3, gets its response and determines if pulp3 is well
and alive before running a pulp3 action.


jturel:
When I run `ping` without a proxy id it gives an error:

```
irb(main):013:0> Katello::Ping.ping                                                                                                                                          
2019-05-28T04:02:23 [W|app|] Couldn't find SmartProxy without an ID                                                                                                          
 | /home/vagrant/foreman/.vendor/ruby/2.5.0/gems/activerecord-5.2.1/lib/active_record/relation/finder_methods.rb:429:in `find_with_ids'
 | /home/vagrant/foreman/.vendor/ruby/2.5.0/gems/activerecord-5.2.1/lib/active_record/relation/finder_methods.rb:69:in `find'   
 | /home/vagrant/foreman/.vendor/ruby/2.5.0/gems/activerecord-5.2.1/lib/active_record/querying.rb:5:in `find'           
 | /home/vagrant/foreman/.vendor/ruby/2.5.0/gems/activerecord-5.2.1/lib/active_record/core.rb:160:in `find'
 | /home/vagrant/katello/app/models/katello/ping.rb:34:in `block in ping_pulp3_without_auth'                                        
 | /home/vagrant/katello/app/models/katello/ping.rb:95:in `exception_watch'                                                                                                  
 | /home/vagrant/katello/app/models/katello/ping.rb:33:in `ping_pulp3_without_auth'                                                                                          
 | /home/vagrant/katello/app/models/katello/ping.rb:16:in `ping'
```

Looks like the existing `pulp_url` method handles that case - should `pulp3_url` do the same?

jturel:
The ping code works well outside of the question above. But how can I test the integration with pulp actions? Creating a file repo locally is using pulp2. I thought it would be using pulp3 now? I've refreshed my local foreman-proxy and see pulp3 where I have the file plugin installed and things look good there.

From my status API output:

```
    "versions": [
        {
            "component": "pulpcore",
            "version": "3.0.0rc2"
        },
        {
            "component": "pulpcore-plugin",
            "version": "0.1.0rc2"
        },
        {
            "component": "pulp_file",
            "version": "0.0.1b11"
        }
    ]
```

sjha4:
Probably need to add pulp3 as a service in the middleware. Also, would make sense to check only pulp3 for Pulp3 actions and not pulp and vice-versa?

Also, for https://github.com/Katello/katello/blob/a1911c7b68208c54d0f8667dfbc12a17bebf240c/app/lib/actions/middleware/backend_services_check.rb#L30 ,
 I think we don't pass capsule_id as an argument in most pulp3 actions. Instead we are passing the smart_proxy itself.

app/lib/actions/pulp3/abstract.rb
```diff
@@ -2,6 +2,7 @@ module Actions
   module Pulp3
     class Abstract < Actions::Base
       middleware.use ::Actions::Middleware::RemoteAction
+      middleware.use Actions::Middleware::PulpServicesCheck
```
jturel: I think @sjha4 might be saying to create a new middleware that checks pulp3 only:

https://github.com/Katello/katello/blob/a1911c7b68208c54d0f8667dfbc12a17bebf240c/app/lib/actions/middleware/pulp_services_check.rb#L5

jturel:
@ekohl all good?

ekohl:
This feels very wrong to me. There are so many internals parsed. Katello shouldn't care about Pulp internals like Redis or its database. *If* you want to ping, then the Pulp3 status API should have a summary field that states whether the entire API is alive, optionally with a human readable description.

However, IMHO it's still a bad idea to ping around actions since it introduces a race condition. It may work before an action, but during the service may fail. That means you still need to do error handling. If you have error handling in place, then pinging beforehand is not needed. Pulp should also provide an atomic API.

jturel:
oops. I meant to ping you on another PR @ekohl. Sorry about that- but thanks for the feedback!

jlsherrill:
@ekohl hey, so there is some history behind this.  A while back pulp2 was quite unstable and would often fail.  Typically we would notice it had failed in the middle of a multi-step foreman-tasks action.   When this happens for things like a cv publish or prote, the task is left paused and it waits for the user to attempt correct the situation and resume it.  The problem is that this was happening a lot, and we added a health check before a task is executed to ensure that pulp is healthy prior to running an action.  This helped reduce the frequency that the user got stuck with a paused task and had to deal all those implications.  When doing pulp3 we're just extending this same check to those pulp3 actions.

parthaa:
> I think we don't pass capsule_id as an argument in most pulp3 actions. Instead we are passing the smart_proxy itself.

@sjha4 good call on the pulp3 check. Implemented you suggestion and seems to be working well in my tests.
@jturel if you wanted to test the ping - assuming you have pulp3 setup - checkout the PR and just create/sync a file repo.  
Try the following command
```
$ sudo journalctl --all|grep status

Jun 12 06:29:56 zeta.partello.example.com gunicorn[10201]: 127.0.0.1 - - [12/Jun/2019:06:29:56 +0000] "GET /pulp/api/v3/status/ HTTP/1.1" 200 1118 "-" "rest-client/2.0.2 (linux-gnu x86_64) ruby/2.5.3p105"
```
hopefully you should see an output like above.

app/lib/actions/middleware/backend_services_check.rb
```diff
@@ -26,8 +26,8 @@ def plan(*args)
       protected
 
       def capsule_id(args)
```
jturel: I found this method too difficult to understand for 2 lines of code. Since we're changing it- what do you think about this implementation?

```
def capsule_id(args)                                                                                                                                                         
  capsule_id = nil                                                                                                                                                           
  args.each do |arg|                                                                                                                                                         
    next unless arg.is_a? Hash                                                                                                                                               
                                                                                                                                                                             
    capsule_id = arg[:capsule_id] || arg[:smart_proxy_id]                                                                                                                    
    break if capsule_id                                                                                                                                                      
  end                                                                                                                                                                        
  capsule_id                                                                                                                                                                 
end
```

parthaa: Nice suggestion. Updated :)

app/models/katello/ping.rb
```diff
@@ -137,6 +143,25 @@ def pulp_without_auth(url)
...
+          fail _("Pulp database connection issue at %s.") % url
+        end
+
+        if json['redis_connection'] && json['redis_connection']['connected'] != true
+          fail _("Pulp redis connection issue at %s.") % url
+        end
+
+        if json["online_workers"].empty?
+          fail _("Not all necessary pulp workers running at %s.") % url
+        end
```
sjha4: So the status api returns the pulp-resource-manager service as a worker. If we just stop the workers then the tasks end up waiting for workers to be started. I guess we can ask for a separate key for workers other than the resource manager itself or check if the count>1 of workers?

sjha4:
Tested pulp3 create, update, delete, sync, indexing and CV create/publish/promote/delete with file repos. Also works as expected for Pulp2 for the same tasks. :heavy_check_mark:  
Works great and handles both up and down states well other than the one point in the comment above.

katello.gemspec
```diff
@@ -44,6 +44,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency "anemone"
   gem.add_dependency "pulpcore_client"
   gem.add_dependency "pulp_file_client", "< 1.0"
+  gem.add_dependency "pulp_docker_client"
```
sjha4: Did you mean to add this dependency for this PR?

PR #8132, opened 2019-05-28
Fixes #26895 - avoid auto complete on repo update

jlsherrill:
the repo upstream username and password can
trigger autocomplete if a user or password is saved.
In firefox we can easily avoid this, but in chrome we cannot,
so instead of hiding these fields, we can just not render them

### DIFF
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/views/repository-info.html
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/views/repository-info.html
@@ -82,7 +82,7 @@
           readonly="denied('edit_products', product)">
       </dd>
 
-      <span ng-show="!product.redhat">
+      <span ng-if="!product.redhat">
         <dt translate>Upstream Authorization</dt>
         <dd bst-edit-custom="repository.upstream_auth_exists"
             readonly="denied('edit_products', product)"
@@ -95,11 +95,13 @@
           <input id="upstream_username"
                  name="upstream_username"
                  type="name"
+                 autocomplete="off"
                  ng-model="repository.upstream_username"/>
           <div translate>Upstream Password</div>
           <input id="upstream_password"
                  name="upstream_password"
                  type="password"
+                 autocomplete="off"
                  ng-model="repository.upstream_password"/>
         </dd>
       </span>


jlsherrill:
This was kinda tricky to reproduce, i had to do all of this (but all may not be required):

1)  access the server over https with the correct hostname of the system
2) access the server using a trusted ca (import the server's ca as trusted)
3) logout of foreman and log back in, chrome should ask to save the password, confirm that
4) go to a custom repository, save the upstream username and password, confirm if asked to save
5) on a redhat repository change the download policy or 'mirror on sync' option, if you don't get the error immediately save it again
6) repeat steps 1-5 while standing on your left foot only, if it still doesn't reproduce, try on your right foot

jlsherrill:
Note that to properly test this, you'll also need https://github.com/Katello/katello/pull/8133

jlsherrill:
[test katello]

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/views/repository-info.html
```diff
@@ -82,7 +82,7 @@ <h4 translate>Sync Settings</h4>
           readonly="denied('edit_products', product)">
       </dd>
 
-      <span ng-show="!product.redhat">
+      <span ng-if="!product.redhat">
```
jturel: I was able to reproduce after performing a pirouette on my head while juggling several bowling pins with my feet - and it fixes the problem! :+1: 

I noticed one thing that seems to related to s/ng-show/ng-if -> when I load the repo details page for a red hat product the 'Upstream Authorization' fields is shown and disappears shortly after (removal from DOM). Though I think it's the crux of your change - is it required / can be done in a different way?

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/views/repository-info.html
```diff
@@ -95,11 +95,13 @@ <h4 translate>Sync Settings</h4>
           <input id="upstream_username"
                  name="upstream_username"
                  type="name"
+                 autocomplete="off"
```
jturel: Was this changed as a potential safeguard around caching angular could be doing?

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/views/repository-info.html
jlsherrill: I could change this to ng-if="product && !product.redhat"  I think that would make it 'pop in' to screen rather than 'pop out'.  I would think this would happen even with ng-show though, but i'm not quite sure.

jlsherrill: Also its probably less of an issue with a production server where multiple requests can be handled at the same time (and faster)

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/views/repository-info.html
jlsherrill: This is 'the' way of disabling auto complete on a form input, however chrome no longer seems to support this.  My thinking is that we should prefer to have it autocomplete disabled for this field and do our best to do that where we can.   Although maybe we should allow it?

jlsherrill: To clarify, this problem is solved even without this line, but it seemed like a good idea to me

jturel: gotcha. makes sense to not autocomplete since the values are getting saved

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/views/repository-info.html
jturel: A similar behavior happens on master - I guess I'm just not looking at repo details for Red Hat products too much. You're right - should be better in prod, so I'm OK with it.

PR #8133, opened 2019-05-28
Fixes #26870 - properly adapt repo update task

jlsherrill:
as part of a recent refactoring, the pulp2 repo update
task was being called improperly.  This adds an adapter action
in the orchestration layer.

### DIFF
--- a/app/lib/actions/katello/repository/update.rb
+++ b/app/lib/actions/katello/repository/update.rb
@@ -31,7 +31,7 @@ module Actions
                                        gpg_url: repository.yum_gpg_key_url)
           end
           if root.pulp_update_needed?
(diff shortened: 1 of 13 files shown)


jlsherrill:
TODO: add a test

jlsherrill:
@jjeffers our 'scenario' tests do this, see:  https://github.com/Katello/katello/blob/master/test/scenarios/scenario_test.rb  & https://github.com/Katello/katello/blob/master/test/scenarios/annotations/README.md

but we haven't added any for pulp3.   I'm not sure if its possible to plan out a full task without actually executing it, i believe it stops at the first layer of planning.  I'll look into it.

jlsherrill:
Added a full integration-style test to cover this, as i don't think there is an easy way to plan this fully out without running it

PR #8135, opened 2019-05-28
Fixes #26902 - Add uuid param back to CV puppet environment

johnpmitsch:
This param was removed [here](https://github.com/Katello/katello/commit/0d5a1ebe7d722113d16146d6f98b6d2a39c351c0#diff-b3f64de16869054a88473910f7fef5cfL12), rightfully so since it only grabs a single puppet module, which is what the show call is for. However, we should deprecate the param like we have the other one in the POST call


johnpmitsch:
a related concern: does apipie not show deprecated warnings in the docs? I'm not seeing any (even after regenerating the docs)

johnpmitsch:
Since [my question](https://github.com/Katello/katello/pull/8135#issuecomment-496543708) is a larger concern I will merge this and I've reached out to @iNecas to clarify

PR #8136, opened 2019-05-28
Fixes #26907 - always verify ssl on capsule sync

jlsherrill:


jlsherrill:
[test katello]

jlsherrill:
How to test:

1) setup a foreman proxy with content
2) associate some lifecycle env to the smart proxy
3) promote some yum repo in a CV to the lifecycle env
4) manually sync the capsule 
5) manually sync the capsule again

check the capsule sync task in dynflow and look for the refresh step and verify that NO tasks are listed in that step's output.

sjha4:
Tested that no pulp tasks are listed in the RefreshNeeded step with the changes. :heavy_check_mark: 

I don't understand the flow here though and how setting ssl_validation to true is affecting this.

akofink:
Is this PR really to fix https://projects.theforeman.org/issues/26907?

mccun934:
@akofink agreed that the change is odd that the effect is so severe. @jlsherrill can you clarify why ssl validation forces a full sync?

jlsherrill:
Hi yes, this is really a fix to that issue.  Let me explain a bit. 

On every capsule sync, before katello actually kicks off a capsule sync on a repo, it checks:
1.  Does this repository exist on the remote pulp server (the capsule/foreman proxy)
2.  If it doesn't, it creates it.  
3.  If it does exist, it compares the configuration of the repository, its importer and distributors with what is expected.
4.  If there is a difference, it updates the importer or distributors to match the expected configuration
5.  Anytime pulp syncs or publishes a repo, if its configuration has had an update on it, it will bypass any sync/publish optimizations that cause it to be skipped.  This is because it may need to fully sync or publish for those new changes to take effect

Now in this case, on a previous release, we were configuring importers with:

```{"feed":"http://myurl", "ssl_validation": VALUE}```

where VALUE is the setting that was configured by the user for that repository.  Now this setting is intended to control the ssl validation when syncing upstream, but in this case it was controlling it on capsule syncs as well!  This was a bug, and during recent refactoring we fixed this by simply omitting the setting:

```{"feed": "http://myurl"}```

And this worked fine on new installs, since the default value of ssl_validation is true.  However on old installs, katello would compare the expected configuration:

```{"feed": "http://myurl"}```
to what was on the pulp server:
```{"feed":"http://myurl", "ssl_validation": VALUE}```

and find a mismatch, it would then update the pulp importer with the expected configuration of `{"feed": "http://myurl"}`

However since the update in pulp only actually updates values to those that were passed in (and ssl_validation was omitted), an 'update' would occur that doesn't actually update anything, and the pulp configuration would actually be ```{"feed":"http://myurl", "ssl_validation": VALUE}```

Repeat this for every sync, and it would always find a mismatch and attempt to update it, but never actually bringing it inline with the expected configuration.   

We likely should also change this to use a PUT instead of a PATCH, so it would change all values of the configuration.  I'm a little hesitant to make that chnage for the existing pulp2 code, but will definitely use that method for pulp3 repository updating.

PR #8137, opened 2019-05-28
Fixes #26243 - remove file content

jjeffers:
This PR introduces support for removing file repository content (backed by pulp3 servers).

To test this PR:
- Create a file repository with an upstream URL of `https://repos.fedorapeople.org/repos/pulp/pulp/fixtures/file/`
- Sync the repository.
- Navigate to the repository details  at `/products/n/repositories/y`and then to the content files page. Observe that under the repository details thee are 3 files, "1.iso", "2.iso", "3.iso".
- You can also confirm the file contents from the console, for example
`Katello::Repository.find(y).repository_files`.  Note also the repository version_href .
- Select "1.iso" and click the "Remove Files" button.
- After the task completes, the repository should only have content files "2.iso" and "3.iso" associated.
- From the console you can verify that the repository version_href has been updated.


app/lib/actions/pulp/orchestration/repository/remove_units.rb
```diff
@@ -0,0 +1,15 @@
+module Actions
+  module Pulp
+    module Orchestration
+      module Repository
+        class RemoveUnits < Pulp::Abstract
+          def plan(_repository, _smart_proxy, options)
+            sequence do
+              plan_action(Actions::Repository::RemoveUnits, options)
+            end
```
sjha4: You probably meant `Actions::Pulp::Repository::RemoveUnits` here..

jjeffers: You mean, without the `sequence` term?

sjha4: The namespace seems to be off here.. The remove units action would be under pulp module?
Also, the sequence do block can be ommited for a singular action.

app/lib/actions/pulp3/repository/remove_units.rb
```diff
@@ -0,0 +1,21 @@
...
+        def plan(repository, smart_proxy, options)
+          plan_self(repository_id: repository.id,
+                    smart_proxy_id: smart_proxy.id,
+                    options: options)
+        end
+
+        def invoke_external_task
+          repo = ::Katello::Repository.find(input[:repository_id])
+          content_unit_ids = input[:options][:contents]
+          file_units = ::Katello::RepositoryFile.find(content_unit_ids)
```
sjha4: I think we need to narrow this down with repository. So the query would be something like:
` Katello::RepositoryFile.where(:file_id => content_unit_ids :repository_id => repo.id)`

app/lib/actions/pulp/orchestration/repository/remove_units.rb
```diff
@@ -0,0 +1,13 @@
+module Actions
+  module Pulp
+    module Orchestration
+      module Repository
+        class RemoveUnits < Pulp::Abstract
+          def plan(_repository, _smart_proxy, options)
+            plan_action(Actions::Pulp::Repository::RemoveUnits, options)
```
jlsherrill: can we get rid of the repo_id param going into RemoveUnits orchestration tasks, and use the repository being passed in?

app/services/katello/pulp3/repository/file.rb
```diff
@@ -91,6 +91,19 @@ def get_distribution(href)
...
+        rescue PulpFileClient::ApiError => e
+          raise e if e.code != 404
+          nil
+        end
+
+        def remove_content(content_units)
+          data = PulpcoreClient::RepositoryVersionCreate.new(
+            remove_content_units: content_units.map(&:pulp_id))
+          pulp3_api.repositories_versions_create(repository_reference.repository_href, data)
+        end
```
jlsherrill: can we move this to pulp3/repository.rb ? doesn't look tied to file at all

jlsherrill:
couple of small comments, but otherwise looks good!!! :+1:

app/lib/actions/pulp/orchestration/repository/remove_units.rb
jjeffers: As long as we are ok refactoring pulp2 things, sure. I'm assuming that's what we're talking about here.

jlsherrill: I actually meant to leave the pulp2 action alone, but in this orchestration action, just do something like:
```
options[:repo_id] = repository.id
```

and remove the 'repo_id being passed in from the  RemoveContent action.  That way, the RemoveContent action doesn't have any pulp2-specific code, which this 'repo_id' param is.  Does that make sense?

jjeffers: Oh! I see what you meant. Makes sense.

jlsherrill:
@jjeffers looks like rubocop failed and a couple tests need updating

app/lib/actions/pulp/orchestration/repository/remove_units.rb
```diff
@@ -0,0 +1,14 @@
+module Actions
+  module Pulp
+    module Orchestration
+      module Repository
+        class RemoveUnits < Pulp::Abstract
+          def plan(_repository, _smart_proxy, options)
+            options[:repo_id] = repository_id
```
jlsherrill: I think you meant 'repository.id'  This currently causes a failure when trying to remove units from a pulp2 repo

katello.gemspec
```diff
@@ -43,7 +43,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency "runcible", ">= 2.11.0", "< 3.0.0"
   gem.add_dependency "anemone"
   gem.add_dependency "pulpcore_client"
-  gem.add_dependency "pulp_file_client", "~> 3.0.0rc2.dev.1559507270"
+  gem.add_dependency "pulp_file_client", "~> 3.0.0rc2.dev.1558441126"
```
jlsherrill: i think you have to do '= 3.0.0rc2.dev.1558441126'

PR #8138, opened 2019-05-29
Fixes #26912 - Better error for subs page on 410

johnpmitsch:
When a Susbscription Allocation is removed from Red Hat Subscription Manager and you
still import that manifest, you will see `410 GONE` errors, meaning the subscription allocation
no longer exists. This commit catches that error and responds with a more friendly message:

```
The Subscription Allocation providing the imported manifest has been removed.
Please create a new Subscription Allocation and import the new manifest.
```
sub

### DIFF
--- a/app/models/katello/upstream_pool.rb
+++ b/app/models/katello/upstream_pool.rb
@@ -13,11 +13,18 @@ module Katello
           included_fields: included_field_params(quantities_only)
         )
 
-        upstream_response = CP_POOL.get(params: cp_params)
+        upstream_response = fetch_upstream_with_error_handling(cp_params)
         pools = response_to_pools(upstream_response, pool_id_map: pool_id_map)
         total = upstream_response.headers[total_count_header] || pools.count
 
         respond(pools, total)
+      end
+
+      def fetch_upstream_with_error_handling(cp_params)
+        CP_POOL.get(params: cp_params)
+      rescue RestClient::Gone
+        raise "The Subscription Allocation providing the imported manifest has been removed. " \
+              "Please create a new Subscription Allocation and import the new manifest."
       end
 
       def respond(pools, total)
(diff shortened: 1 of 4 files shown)


johnpmitsch:
I thought about adding another catch for `401 unauthorized`, but that is most commonly seen when using manifests from the staging CDN, so not really user-facing.

app/models/katello/upstream_pool.rb
```diff
@@ -13,13 +13,20 @@ def fetch_pools(params)
...
         pools = response_to_pools(upstream_response, pool_id_map: pool_id_map)
         total = upstream_response.headers[total_count_header] || pools.count
 
         respond(pools, total)
       end
 
+      def fetch_upstream_with_error_handling(cp_params)
+        CP_POOL.get(params: cp_params)
+      rescue RestClient::Gone
+        raise "The Subscription Allocation providing the imported manifest has been removed. " \
```
jturel: Error looks good!

I have one suggestion around the error handling w.r.t `RestClient::Gone` which is I don't think that we should have HTTP / HTTP client lib errors being handled in our models.

We'll probably move away from it one day and the smaller the affected area the better it will be. One straightforward solution is to rescue `RestClient::Gone` in the resource class `Resources::Candlepin::UpstreamPool` and raise a new custom error, something like Katello::Errors::UpstreamConsumerGone and rescue from it in this code. What do you think?

jturel:
Any idea on the LOE to make it so that I can't attempt to modify rows in the subscription list (the Red Hat ones) when this error is happening? It'll stop no-op actions from getting planned

johnpmitsch:
@jturel Updated to move the error to our Error class and the rescue to the Resource class

> Any idea on the LOE to make it so that I can't attempt to modify rows in the subscription list (the Red Hat ones) when this error is happening? It'll stop no-op actions from getting planned

It looks like this is handled for the upstream subscriptions page, its just the local page that will still load:

Here is the upstream page:
here

Does this seem ok?

johnpmitsch:
@jturel The UI part is not as straightforward as I thought, so I can file another issue for that!

johnpmitsch:
filed! https://projects.theforeman.org/issues/27020

PR #8141, opened 2019-06-03
[CP] Katello 3.12 Cherrypicks from Toolbelt

chris1984:


chris1984:
@jturel here is the updated list of cherry-picks from toolbelt for 3.12

jturel:
Looks good! If this is the last of picks for GA I think it'd be appropriate to version bump and add an updated changelog

chris1984:
@jturel updated with the version bump and updated changelog from toolbelt.

jturel:
my only request here would be to completely regenerate the changelog so everything inside is under the 3.12.0 heading (there are two now). otherwise looks good!

chris1984:
@jturel updated with fresh changelog!

PR #8142, opened 2019-06-04
Fixes #26443 - properly remove gpg from content

jlsherrill:
at some point in the past, either candlepin stopped
accepting nil as a valid value for a gpg key url
or katello started sending nil instead of empty string.

This starts sending empty string to properly unset a gpg
key url

### DIFF
--- a/app/lib/actions/candlepin/product/content_update.rb
+++ b/app/lib/actions/candlepin/product/content_update.rb
@@ -19,7 +19,7 @@ module Actions
                      id: input[:content_id],
                      name: input[:name],
                      contentUrl: input[:content_url],
-                     gpgUrl: input[:gpg_key_url],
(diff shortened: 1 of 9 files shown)


jlsherrill:
Reproduce:

1.  create a yum repo
2.  create a content credential of a gpg key (any contents work)
3.  register a system with sub-man and subscribe it to the yum repo
4.  notice in /etc/yum.repos.d/redhat.repo there is no gpg key
5.  Associate the gpg key with the repository on the repo details page
6.  run sub-man refresh && yum repolist
7.  notice in /etc/yum.repos.d/redhat.repo there is a gpg key for the repo
8.  Remove the gpg key from the repo on the repo details page
9.  notice in /etc/yum.repos.d/redhat.repo thegpg key for the repo is now gone.

jlsherrill:
[test katello]

PR #8143, opened 2019-06-04
Fixes #26947 - correct metadata gen logic

jlsherrill:
a logic check was introduced, but during the
planning phase, the value is actually unresolved
output from a previous step which is always truthy
during the plan phase

### DIFF
--- a/test/actions/katello/repository/metadata_generate_test.rb
+++ b/test/actions/katello/repository/metadata_generate_test.rb
@@ -72,5 +72,29 @@ module Actions
       assert_action_planed_with(action, pulp_publish_class, yum_repo, SmartProxy.pulp_master,
                                 yum_action_options)
     end
+
+    it 'plans a yum refresh with matching content set to some deferred object' do
+      action = create_action(action_class)
+      not_falsey = Object.new
+      plan_action(action, yum_repo, :matching_content => not_falsey, :repository_creation => false)
+
+      yum_action_options = action_options.clone
+      yum_action_options[:matching_content] = not_falsey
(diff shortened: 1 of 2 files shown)


jlsherrill:
To test this, 
1.  create a yum repo 'foo'
2.  add it to a NEW content view named 'bar', publish
3.  look on the filesystem for:  /var/lib/pulp/published/yum/https/repos/Default_Organization/Library/bar
4.  On master, it will not exist.  To repeat this delete the content view version completely (so there are no versions), and repeat step 2)

jlsherrill:
[test katello]

jlsherrill:
[test katello]

test/actions/katello/repository/metadata_generate_test.rb
```diff
@@ -72,5 +72,17 @@ module Actions
       assert_action_planed_with(action, pulp_publish_class, yum_repo, SmartProxy.pulp_master,
                                 yum_action_options)
     end
+
+    it 'plans a yum refresh with matching content true, but during repository creation mode' do
```
jturel: This is a valid test case but I don't think it's the one described in the fix (it passes when I revert your code change). Should there be a test to assert that a publish action wasn't planned with the `not_falsey `value or something similar?

jlsherrill: Good eye :+1:

jlsherrill: updated and added a 2nd test

jturel:
[test katello]

jlsherrill:
[test katello]

PR #8144, opened 2019-06-05
Fixes #26956 - update axiom dep

jlsherrill:


jlsherrill:
no idea if this will actually pass :)

package.json
```diff
@@ -53,7 +53,7 @@
     "redux-thunk": "^2.2.0"
   },
   "dependencies": {
-    "axios": "^0.17.1",
+    "axios": "^0.19.0",
```
waldenraines: You may be able to just remove this dependency (and also `axios-mock-adapter` as well) since it's actually [provided by foreman](https://github.com/theforeman/foreman/blob/develop/package.json#L72).  Although this change may also require changes to tests, I'm not sure.

jlsherrill: @waldenraines  react tests did fail when i removed it, thoughts?

waldenraines: I'd have to know why they failed.  It probably means that it needs to be included differently in the test.

waldenraines:
@jlsherrill actually, it will be easier to just update this dependency until the work on the foreman vendor is completed and then we can remove this dependency all together.

Sorry for the back and forth.

package.json
waldenraines: Disregard the above ;)

PR #8145, opened 2019-06-05
Fixes #26904 - improve vcr testing

jlsherrill:
This commit improves vcr testing in a couple ways:
1.  It does not record pending task requests.  This means that all
tasks will appear to finish correctly after the first task poll during
playback.  This provides smaller cassettes as well as faster test runs.
In my testing i saw about a 30-50% speed improvement in test execution
for task-heavy tests.

2.  executing the tests while running mode=all will now delete
the cassettes before recording.  This prevents users from continuing
to re-record the same cassette over and over which has happened in
the past.

3.  re-recording all tests showed that some tests were broken in ways due to recent refactoring that were only seen when re-recording, those have been fixed


jturel:
These changes sound great. I guess the VCRs need a regen now

app/models/katello/glue/pulp/repo.rb
```diff
@@ -137,21 +137,6 @@ def package_group_count
...
-        task = PulpSyncStatus.using_pulp_task(pulp_tasks) do |t|
-          t.organization = organization
-          t.parameters ||= {}
-          t.parameters[:options] = options
-        end
-        task.save!
-        return [task]
-      end
-
       def clone_file_metadata(to_repo)
```
sjha4: :+1:

app/lib/actions/pulp/abstract_async_task.rb
```diff
@@ -125,9 +125,16 @@ def external_task=(external_task_data)
...
+        new_tasks = (spawned_task_ids - current_list.map { |task| task['task_id'] }).map do |task_id|
           task_resource.poll(task_id)
         end
+
+        if new_tasks.empty?
+          []
+        else
+          spawned = new_tasks.map { |t| t['spawned_tasks'].map { |st| st['task_id'] } }.flatten
+          new_tasks + get_new_tasks(current_list + new_tasks, spawned)
+        end
```
jlsherrill: My changes actually exposed a bug (albiet one that would be hard to see and have very little consequences.  If a task has a 'spawned_task' in pulp we would poll it as well as part of the async task.  However if the spawned_task was found on the first poll of the task, it was ignored.  This change basically recursively fetches spanwed_tasks until there are no more.

test/support/pulp/task_support.rb
```diff
@@ -26,6 +26,7 @@ def self.wait_on_tasks(task_list, options = {})
 
     def self.any_task_running(async_tasks)
       async_tasks.each do |t|
+        next if t.finished?
```
jlsherrill: this avoids re-polling tasks that are already finished, such as if 1 out of 3 are finished, just poll the two that aren't.

jlsherrill:
@jturel this is finally green, i left some comments explaining a few of the changes in the non-cassette commit: https://github.com/Katello/katello/pull/8145/commits/d619dbe6690eff2fc77eaacc43f3396d1800e7ca

app/lib/actions/pulp/abstract_async_task.rb
jturel: What's an action I can do that will test the `else` here?

jlsherrill: the vcr tests reproduce it, but it would be very hard to do manually.  You would need to start a task and have it finish before the first task poll, and also it needs kick off another task.  The timeframe here would be a few milliseconds i imagine. 

The tests reproduce it because they are only recording the final task poll

jturel: Gotcha - in that case I'll defer to the green tests!

jturel:
Nooooo more merge conflicts

jturel:
APJ :)

PR #8147, opened 2019-06-06
Fixes #26763 - Pass string representation of time in host info

tbrisker:
Otherwise Puppet 6 can't handle the host enc output (see
https://tickets.puppetlabs.com/browse/PUP-9506 for more info)

### DIFF
--- a/app/models/katello/host/info_provider.rb
+++ b/app/models/katello/host/info_provider.rb
@@ -28,7 +28,7 @@ module Katello
           'label' => host.content_view.try(:label),
           'latest-version' => host.content_view.try(:latest_version),
           'version' => content_version.try(:version),
-          'published' => content_version.try(:created_at).try(:time),
+          'published' => content_version.try(:created_at).try(:time).to_s,
           'components' => content_view_components
         }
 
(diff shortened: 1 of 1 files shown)


app/models/katello/host/info_provider.rb
```diff
@@ -28,7 +28,7 @@ def content_view_info
           'label' => host.content_view.try(:label),
           'latest-version' => host.content_view.try(:latest_version),
           'version' => content_version.try(:version),
-          'published' => content_version.try(:created_at).try(:time),
+          'published' => content_version.try(:created_at).try(:time).to_s,
```
jlsherrill: should this be .try(:to_s)?  otherwise you'd end up with 'nil'  right?

tbrisker:
```
[1] pry(main)> nil.to_s
=> ""

jlsherrill:
Oh nevermind then!  I guess I was misremembering!

jlsherrill:
[test katello]

jlsherrill:
[test katello]

ci timed out

jlsherrill:
[test katello]

transient test failure :(

PR #8148, opened 2019-06-06
Fixes #26973 - include pagination helper in app

jlsherrill:
A recent change in foreman meant that the application
helper now requires the pagination helper.  In foreman
this gets included automatically in controllers, but not in
katello


jlsherrill:
@tbrisker i debated fixing in foreman, but wasn't sure if other plugins would hit this as well or if it was something special about katello.  If we do see it in other plugins, we may want to go that route.

tbrisker:
@jlsherrill katello is the only one that does `clear_helpers` afaik

PR #8149, opened 2019-06-10
Fixes #27011 - Correctly handle files in the gemspec

ekohl:
This ensures gem build katello.gemspec can be used from any directory.


jlsherrill:
Thanks @ekohl !

PR #8152, opened 2019-06-10
fixes #26531 - Search Purpose Addons

sseelam2:
This PR is created to search for hosts using purpose addons. Addons can be viewed under System Purpose inside Host's content.
**To test this PR:**
-  Create a Custom Product
-  Set some addons on it by connecting to your local postgres candlepin db
```
select uuid from cp2_products where name = 'CustomProduct';
insert into cp2_product_attributes(name, value, product_uuid) values ('addons', 'Test Addon1,Test Addon2', 'REPLACEME');
```

- Restart the tomcat service to refresh the product cache ` systemctl restart tomcat`

Now you should be able to see these addons under System Purpose. Add these addons to the host by selecting them and clicking on save.
To perform search on addons, go to Hosts and in the search box, type -
`addon = "Name of the addon you have added to the host" `and press enter. 
You should be able to see the hosts which have the addon that you have searched for. Also while performing a search, the text should be auto completed.``

### DIFF
--- a/app/models/katello/host/subscription_facet.rb
+++ b/app/models/katello/host/subscription_facet.rb
@@ -4,12 +4,16 @@ module Katello
       audited :associated_with => :host, :associations => [:pools], :except => [:last_checkin]
       self.table_name = 'katello_subscription_facets'
       include Facets::Base
+      include DirtyAssociations
 
       belongs_to :user, :inverse_of => :subscription_facets, :class_name => "::User"
       belongs_to :hypervisor_host, :class_name => "::Host::Managed", :foreign_key => "hypervisor_host_id"
 
       has_many :subscription_facet_activation_keys, :class_name => "Katello::SubscriptionFacetActivationKey", :dependent => :destroy, :inverse_of => :subscription_facet
       has_many :activation_keys, :through => :subscription_facet_activation_keys, :class_name => "Katello::ActivationKey"
+
+      has_many :subscription_facet_purpose_addons, :class_name => "Katello::SubscriptionFacetPurposeAddon", :dependent => :destroy, :inverse_of => :subscription_facet
+      has_many :purpose_addons, :class_name => "Katello::PurposeAddon", :through => :subscription_facet_purpose_addons
 
       has_many :subscription_facet_pools, :class_name => "Katello::SubscriptionFacetPool", :dependent => :delete_all, :inverse_of => :subscription_facet
       has_many :pools, :through => :subscription_facet_pools, :class_name => "Katello::Pool"
@@ -19,13 +23,13 @@ module Katello
 
       has_many :compliance_reasons, :class_name => "Katello::ComplianceReason", :dependent => :destroy, :inverse_of => :subscription_facet
 
-      serialize :purpose_addons, Array
-
       validates :host, :presence => true, :allow_blank => false
 
       DEFAULT_TYPE = 'system'.freeze
 
-      accepts_nested_attributes_for :installed_products
+      accepts_nested_attributes_for :installed_products, :purpose_addons
+
+      dirty_has_many_associations :purpose_addons
 
       attr_accessor :facts
 
@@ -53,7 +57,9 @@ module Katello
         self.update_installed_products(consumer_params['installedProducts']) if consumer_params.key?('installedProducts')
         self.purpose_role = consumer_params['role'] unless consumer_params['role'].nil?
         self.purpose_usage = consumer_params['usage'] unless consumer_params['usage'].nil?
-        self.purpose_addons = consumer_params['addOns'] unless consumer_params['addOns'].nil?
+        unless consumer_params['addOns'].nil?
+          self.purpose_addon_ids = consumer_params['addOns'].map { |addon_name| ::Katello::PurposeAddon.find_or_create_by(name: addon_name).id }
+        end
 
         unless consumer_params['releaseVer'].blank?
           release = consumer_params['releaseVer']
@@ -127,7 +133,7 @@ module Katello
           :autoheal => autoheal,
           :usage => purpose_usage,
           :role => purpose_role,
-          :addOns => purpose_addons,
+          :addOns => purpose_addons.pluck(:name),
           :serviceLevel => service_level,
           :releaseVer => release_version,
           :environment => {:id => self.candlepin_environment_id},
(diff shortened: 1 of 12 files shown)


jturel:
[test katello]

jturel:
[test katello]

jturel:
[test katello]

jturel:
[test katello]

jturel:
[test katello]

ianballou:
I've given this a functional test and I can confirm that it works!  Just one correction, the search term was `addon = ...` rather than `addons = ...`.  Auto complete worked too.

sseelam2:
> Just one correction, the search term was `addon = ...` rather than `addons = ...`. Auto complete worked too.

Ah. Thank you for noticing that. I'll edit my comment.

app/models/katello/host/subscription_facet.rb
```diff
@@ -53,7 +57,6 @@ def import_database_attributes(consumer_params = candlepin_consumer.consumer_att
         self.update_installed_products(consumer_params['installedProducts']) if consumer_params.key?('installedProducts')
         self.purpose_role = consumer_params['role'] unless consumer_params['role'].nil?
         self.purpose_usage = consumer_params['usage'] unless consumer_params['usage'].nil?
-        self.purpose_addons = consumer_params['addOns'] unless consumer_params['addOns'].nil?
```
jturel: I looked through the code and there are some instances where we update the subscription facet attributes not based on what was passed in (via UI update or client host checkin). In these situations we update the subscription facet based on the values in candlepin. Notice the signature of the method which does this:

```
def import_database_attributes(consumer_params = candlepin_consumer.consumer_attributes)
```

So, I think we need to restore this behavior that was removed. Since we no longer have purpose_addons as an attribute you can use the same approach as the host extensions with purpose_addon_ids

sseelam2: Yes. That makes sense. I have updated the subscription_facet.rb and also have added the related test. Could you please review.

jturel:
[test katello]

app/models/katello/host/subscription_facet.rb
jturel: Looks good. Can you add the assertion back which tests the functionality?

app/controllers/katello/concerns/api/v2/hostgroups_controller_extensions.rb
```diff
@@ -18,7 +18,6 @@ module ApiPieExtensions
 
       included do
         include ApiPieExtensions
-
```
jturel: This might sound picky but it's better to not introduce unnecessary/unrelated changes. The space that was here should be put back :)

jturel:
[test katello]

db/migrate/20190605014649_add_purpose_addons.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    create_table :katello_subscription_facet_purpose_addons do |t|
+      t.references :purpose_addon, index: { name: :katello_sub_facet_purpose_addons_paid }
+      t.references :subscription_facet, index: { name: :katello_sub_facet_purpose_addons_sfid }
+    end
+
+    add_foreign_key :katello_subscription_facet_purpose_addons, :katello_subscription_facets, column: :subscription_facet_id, name: :katello_sub_facet_purpose_addon_facet_id
+    add_foreign_key :katello_subscription_facet_purpose_addons, :katello_purpose_addons, column: :purpose_addon_id, name: :katello_sub_facet_purpose_addon_purpose_addon_id
+
+    Katello::Host::SubscriptionFacet.pluck(:id, :purpose_addons).each do |facet|
```
jturel: I was most concerned if this migration would work properly or not - and it seems to work great. Here's what I did:

Set up some data:

```
irb(main):010:0> Katello::Host::SubscriptionFacet.where.not(purpose_addons: []).pluck(:id, :purpose_addons)
=> [[265, ["TestAddon1", "TestAddon2"]], [257, ["TestAddon1", "TestAddon2"]], [271, ["TestAddon3"]], [274, ["TestAddon1"]]]
```

Applied code change & migrated db, got these results:

```
irb(main):003:0> Katello::PurposeAddon.pluck(:name)
=> ["TestAddon1", "TestAddon2", "TestAddon3"]

Katello::Host::SubscriptionFacet.find(265).purpose_addons.pluck(:name)
=> ["TestAddon1", "TestAddon2"]

Katello::Host::SubscriptionFacet.find(257).purpose_addons.pluck(:name)
=> ["TestAddon1", "TestAddon2"]

Katello::Host::SubscriptionFacet.find(271).purpose_addons.pluck(:name)
=> ["TestAddon3"]

Katello::Host::SubscriptionFacet.find(274).purpose_addons.pluck(:name)
=> ["TestAddon1"]

Katello::SubscriptionFacetPurposeAddon.pluck(:subscription_facet_id, :purpose_addon_id)
=> [[265, 1], [265, 2], [257, 1], [257, 2], [271, 3], [274, 1]]
```

lib/katello/plugin.rb
```diff
@@ -211,7 +211,7 @@
     :content_facet_attributes => [:content_view_id, :lifecycle_environment_id, :content_source_id,
                                   :host, :kickstart_repository_id],
     :subscription_facet_attributes => [:release_version, :autoheal, :purpose_usage, :purpose_role, :service_level, :host,
-                                       {:installed_products => [:product_id, :product_name, :arch, :version]}, :facts, :hypervisor_guest_uuids => [], :purpose_addons => []]
+                                       {:installed_products => [:product_id, :product_name, :arch, :version]}, :facts, :hypervisor_guest_uuids => [], :purpose_addon_ids => []]
```
jturel: Whenever I update purpose addons from the UI I see this in the logs:

```
2019-06-18T01:03:01 [D|app|ffe23364] Unpermitted parameter: :purpose_addons
```

It would be nice if we could omit this. I think we need to add back purpose_addons here, and I think we need purpose_addon_ids too.

jturel:
Works really well, left just one comment on some logging around unpermitted params.

db/migrate/20190605014649_add_purpose_addons.rb
sseelam2: Yes. This looks like the tables are set up the way we want. :+1:

lib/katello/plugin.rb
sseelam2: Adding the :purpose_addons => [] back does not seem to help because it takes more precedence over purpose_addon_ids, so an instance of a String is being accepted rather than Katello::PurposeAddon and this breaks the functionality.

sseelam2: As you suggested, deleting the :purpose_addons key after updating the :purpose_addon_ids in the before action under hosts_controller_extensions.rb  work.

jturel:
[test katello]

PR #8153, opened 2019-06-10
Fixes #26981 -  /katello/api/repository_sets/:id returns too many repositories

ianballou:
To reproduce:

- Import a manifest with Red Hat content.
- Check the katello/api/v2/repository_sets endpoint and pick one set that exists there.  Note that "repositories" in the set should be empty.
- Enable the chosen RH repo in Foreman.
- Create two Content Views, with each being in a different Lifecycle Environment.
- Add the same enabled repo to each of the two Content Views and publish.
- Check the chosen product in the repository_sets API endpoint again.  "repositories" under that product should now be populated and have duplicates with only different IDs.
- Apply this PR to Katello.
- Check the repository_sets API endpoint again.  This time, under the chosen product's repositories list, there should no longer be any duplicate repositories.

### DIFF
--- a/test/models/product_content_test.rb
+++ b/test/models/product_content_test.rb
@@ -33,9 +33,12 @@ module Katello
       repo = katello_repositories(:fedora_17_x86_64)
       repo.root.update_attributes!(product: @product, content_id: @content_id)
 
-      assert_equal repo.clones.count + 1, @product_content.repositories.size
-
       assert_includes @product_content.repositories, repo
+
+      @product_content.repositories.each do |repository|
+        assert repository.in_default_view?
+        assert_equal @product_content.product, repository.product
+      end
     end
 
     def test_enabled
--- a/app/models/katello/product_content.rb
+++ b/app/models/katello/product_content.rb
@@ -36,7 +36,7 @@ module Katello
 
     # used by Katello::Api::V2::RepositorySetsController#index
     def repositories
-      Katello::Repository.where(:root_id => product.root_repositories.has_url.where(:content_id => content.cp_content_id))
+      Katello::Repository.in_default_view.where(:root_id => product.root_repositories.has_url.where(:content_id => content.cp_content_id))
     end
   end
 end


chris1984:
ok  to test

parthaa:
Can you add a unit test :)

ianballou:
@parthaa sure thing

parthaa:
look into https://github.com/Katello/katello/blob/master/test/models/product_content_test.rb#L30-L39

ianballou:
Looks like the fix is now causing 
```assert_equal repo.clones.count + 1, @product_content.repositories.size``` 
to fail.  Need to figure out what this was trying to test, since it's now incorrect.

**Comment by ianballou** [comment]
@parthaa I added a unit test.  It fails without the fix in this PR and passes with it, so I think it covers what we need.  Let me know what you think

File: test/models/product_content_test.rb [hunk 1]
```diff
@@ -36,6 +36,10 @@ def test_repositories
       assert_equal repo.clones.count + 1, @product_content.repositories.size
 
       assert_includes @product_content.repositories, repo
+
+      @product_content.repositories.each do |clone|
+        assert clone.in_default_view?
+      end
```
jturel: Thanks for adding an assertion to verify your fix. Based on our IRC discussion I advise removing the assertion above which checks the count of the repositories. I think it's meaningless and I regret writing that :-1: 

Your new assertion is much better and defines the behavior of the change you made. You might also add an assertion in the loop which verifies that the product on each repo matched the product_content object ie

```
assert_equal @product_content.product, clone.product
```

Maybe `repo` is a better name than `clone` since it describes the type of object being iterated over?

ianballou: Okay, I'll remove the previous assertion.  And yeah I like the name `repo` better, I just wasn't sure if that would cause confusion with the other repo that's also being tested.  I'll change :+1:

PR #8157, opened 2019-06-11
Fixes #27014 - Smart Proxy set initial loc and org

sjha4:
The way I tested this was to run `bundle exec rails katello:reset` in my foreman dir. Another way is to spin up a new dev box and check the smart proxy created.
**Issue**: The default smart proxy created was not assigned the default location and org.
**Expected**: The default smart proxy should be assigned thye default org and location on create.

### DIFF
--- a/app/models/katello/concerns/smart_proxy_extensions.rb
+++ b/app/models/katello/concerns/smart_proxy_extensions.rb
@@ -151,12 +151,11 @@ module Katello
       end
 
       def pulp_mirror?
-        self.features.map(&:name).include?(PULP_NODE_FEATURE)
+        self.has_feature? PULP_NODE_FEATURE
       end
 
       def pulp_master?
-        # use map instead of pluck in case the features aren't saved yet during create
-        self.features.map(&:name).include?(PULP_FEATURE)
+        self.has_feature? PULP_FEATURE
       end
 
       #deprecated methods
@@ -168,7 +167,7 @@ module Katello
       end
 
       def associate_default_locations
-        return unless pulp_master?
+        return unless self.pulp_master?
         ['puppet_content', 'subscribed_hosts'].each do |type|
           default_location = ::Location.unscoped.find_by_title(
             ::Setting[:"default_location_#{type}"])
--- a/katello.gemspec
+++ b/katello.gemspec
@@ -43,7 +43,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency "runcible", ">= 2.11.0", "< 3.0.0"
   gem.add_dependency "anemone"
   gem.add_dependency "pulpcore_client"
-  gem.add_dependency "pulp_file_client", "3.0.0rc2.dev.1558441126"
+  gem.add_dependency "pulp_file_client", "< 1.0"
 
   # UI
   gem.add_dependency "deface", '>= 1.0.2', '< 2.0.0'
--- a/app/views/overrides/smart_proxies/_environment_tab.html.erb
+++ b/app/views/overrides/smart_proxies/_environment_tab.html.erb
@@ -1,4 +1,4 @@
-<% if !@smart_proxy.new_record? && @smart_proxy.features.pluck(:name).include?(SmartProxy::PULP_NODE_FEATURE) -%>
+<% if !@smart_proxy.new_record? && @smart_proxy.has_feature?(SmartProxy::PULP_NODE_FEATURE) -%>
   <li id="kt_environments_tab">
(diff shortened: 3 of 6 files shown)


jlsherrill:
Can we add a test for this?

app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -156,7 +156,7 @@ def pulp_mirror?
 
       def pulp_master?
         # use map instead of pluck in case the features aren't saved yet during create
-        self.features.map(&:name).include?(PULP_FEATURE)
+        self.smart_proxy_features.any? { |smart_proxy_feature| smart_proxy_feature.feature.try(:name).try(:eql?, PULP_FEATURE) }
```
parthaa: can you rewrite it to 
```
self.smart_proxy_features.any? { |smart_proxy_feature| PULP_FEATURE == smart_proxy_feature.feature.try(:name)}
```

jturel: This code makes me think that smart_proxy_features isn't a relation since we're using 'any?' vs 'where' is that right? Why isn't it a relation? I *thought* that the features were a model object. Just wondering for my own knowledge

parthaa: @jturel so smart_proxy used to be related to features via HABTM relationship. That recently got changed here to has_many through -> https://github.com/theforeman/foreman/commit/9080d4e5bc4fe7c7628cc1fce738302aceaed7ce

since the association to org and locations happens "before_create" all the entries are not fully baked in the db yet. That is causing this side effect where ```self.smart_proxy_features``` returns a list of features while ```self.features   ``` turns up empty. So we need to resort to looking at smart_proxy_features model directly

sjha4:
Pushed a change to pulp_mirror check based on offline suggestion by @parthaa

app/models/katello/concerns/smart_proxy_extensions.rb
jturel: Got it. Thanks @parthaa !

app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -155,8 +155,7 @@ def pulp_mirror?
       end
 
       def pulp_master?
-        # use map instead of pluck in case the features aren't saved yet during create
-        self.features.map(&:name).include?(PULP_FEATURE)
+        self.smart_proxy_features.map(&:feature).any? { |f| PULP_FEATURE == f.name }
```
ekohl: Isn't this equivalent?

```suggestion
        self.has_feature?(PULP_FEATURE)
```

app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -155,8 +155,7 @@ def pulp_mirror?
       end
```
ekohl: This feature looks like it may need a similar treatment

ekohl:
Other places that probably should use `has_feature?` are:

https://github.com/Katello/katello/blob/ea3d7cb751b22d0abe0d8d3175fefcc4a4484fcd/app/views/overrides/smart_proxies/_environment_tab.html.erb#L1
https://github.com/Katello/katello/blob/ea3d7cb751b22d0abe0d8d3175fefcc4a4484fcd/app/views/overrides/smart_proxies/_environment_tab_pane.html.erb#L1

app/models/katello/concerns/smart_proxy_extensions.rb
sjha4: Yes..This works well..Wasn't aware of this method. Will update! :+1:

ekohl: https://github.com/theforeman/foreman/commit/8f657a8422a1ab85defaa9a999cc79c32ccfb1c6 has been there for a while so surprised nobody noticed this given the author worked on katello :)

sjha4:
Ready to merge!

katello.gemspec
```diff
@@ -43,7 +43,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency "runcible", ">= 2.11.0", "< 3.0.0"
   gem.add_dependency "anemone"
   gem.add_dependency "pulpcore_client"
-  gem.add_dependency "pulp_file_client", "3.0.0rc2.dev.1558441126"
+  gem.add_dependency "pulp_file_client", "< 1.0"
```
ekohl: FYI: https://github.com/Katello/katello/commit/f4a7b6bb1347e05a648374a06ae41c29338d9518 already did this.

parthaa:
ACK

PR #8158, opened 2019-06-11
Fixes #27009 - adds a default http proxy setting for content

jjeffers:
This PR creates a new Setting in the "Content" section under Administer->Settings in the UI.

The default is an empty value, meaning no global default value for content syncing will be set for organizations or repositories (remotes).

When clicked, the input will include any http proxies defined in the application. The value use is the proxy url. It's possible to set the setting value to a "nil" value by selecting "no global default".

### DIFF
--- a/app/models/katello/concerns/http_proxy_extensions.rb
+++ b/app/models/katello/concerns/http_proxy_extensions.rb
@@ -11,13 +11,11 @@ module Katello
         changes = self.previous_changes
         if changes.key?(:name)
           previous_name = changes[:name].first
-          content_settings = Setting::Content.arel_table
           setting = Setting::Content.unscoped.
             where(name: "content_default_http_proxy").
-            where(content_settings[:value].matches("%#{previous_name}%")).
             first
 
-          if setting && !previous_name.blank?
+          if setting && setting.value == previous_name && !previous_name.blank?
             setting.update_attribute(:value, self.name)
           end
         end
--- a/app/models/katello/concerns/http_proxy_extensions.rb
+++ b/app/models/katello/concerns/http_proxy_extensions.rb
@@ -1,0 +1,19 @@
+module Katello
+  module Concerns
+    module HttpProxyExtensions
+      extend ActiveSupport::Concern
+
+      included do
+        after_save :update_default_proxy_setting
+      end
+
+      def update_default_proxy_setting
+        setting = Setting.where(name: "content_default_http_proxy").first
+
+        if setting
+          setting.update_attribute(:value, self.name)
+        end
+      end
+    end
+  end
+end
--- a/app/models/katello/concerns/http_proxy_extensions.rb
+++ b/app/models/katello/concerns/http_proxy_extensions.rb
@@ -16,7 +16,7 @@ module Katello
             where(value: previous_name.to_yaml).
             first
 
-          if setting and not previous_name.blank?
+          if setting && !previous_name.blank?
             setting.update_attribute(:value, self.name)
           end
         end
--- a/app/models/katello/concerns/http_proxy_extensions.rb
+++ b/app/models/katello/concerns/http_proxy_extensions.rb
@@ -14,6 +14,10 @@ module Katello
           setting.update_attribute(:value, self.name)
         end
       end
+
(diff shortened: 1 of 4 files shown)


jlsherrill:
Can this pr is relatively small, can we go ahead and handle the renaming of the default http proxy as part of this pr?

jjeffers:
@jlsherrill tested with the latest pulpfile client and didn't encounter any problems. Should we keep the pin, or just specify the version as "< 1" as in master branch?

jlsherrill:
Id leave it as is in master

jjeffers:
@jlsherrill can you elaborate on what you meant by "handle the renaming of the default http proxy"?

jlsherrill:
What we chatted about yesterday, you set a default proxy and then go and
rename it, we need the setting to update.

It'd involve adding an extension/concern to the http proxy model (monkey
patching like we do for a lot of models).  Engine.rb loads a bunch of
these.  You'd probably add an after save hook to do the rename.

jjeffers:
@jlsherrill if I understand he settings correctly, the only value that's being stored in the record is the proxy url. If we rename the HttpProxy (change the proxy.name), I don't think this will change the Setting::Content record, will it?

jlsherrill:
Oh, I didn't catch that, we should probably store the name instead of the
url, you potentially could have two proxies with the same url.

app/models/setting/content.rb
```diff
@@ -10,14 +10,26 @@ def self.load_defaults
...
+      value_method: 'name',
+      text_method: 'name_and_url'
+    }]
 
     self.transaction do
       [
+        self.set('content_default_http_proxy', N_("Default HTTP Proxy for syncing content"),
+                        nil, N_('Default http proxy'),
+                        nil,
+                        collection: proc { http_proxy_select }, include_blank: "no global default"
```
jlsherrill: We probably want to i18n that 'no global default'

jlsherrill:
As discussed in person but documenting here, if i:

1. have no http proxies, setting has defaulted to default value
2.  create an http proxy
3.  navigate to the settings page, the added setting shows the newly created http proxy as selected even though it isn't

app/models/katello/concerns/http_proxy_extensions.rb
```diff
@@ -8,10 +8,17 @@ module HttpProxyExtensions
       end
 
       def update_default_proxy_setting
-        setting = Setting.where(name: "content_default_http_proxy").first
+        changes = self.previous_changes
+        if changes.key?(:name)
+          previous_name = changes[:name].first
+          setting = Setting.
+            where(name: "content_default_http_proxy").
+            where(value: previous_name.to_yaml).
```
jlsherrill: why did you have to call .to_yaml?

jjeffers: If you attempt to query based on the proxy name, the where clause will match nothing. The setting values are converted to yaml during the assignment of values.
See 
https://github.com/theforeman/foreman/blob/a31356d3f35ed8f2da7eb1501c8cf9a2551855e3/app/models/setting.rb#L116

app/models/katello/concerns/http_proxy_extensions.rb
```diff
@@ -11,9 +11,10 @@ def update_default_proxy_setting
         changes = self.previous_changes
         if changes.key?(:name)
           previous_name = changes[:name].first
-          setting = Setting.
+          content_settings = Setting::Content.arel_table
+          setting = Setting::Content.unscoped.
             where(name: "content_default_http_proxy").
-            where(value: previous_name.to_yaml).
+            where(content_settings[:value].matches("%#{previous_name}%")).
```
jlsherrill: this seems kinda wrong as 'foo' would match 'foobar'.  What about something like:
```
setting = Setting::Content.unscoped.find_by(name: "content_default_http_proxy")
if setting.value == previous_name:
  UPDATE_NAME
end
```

jlsherrill:
Looks good to me! just a rubocop and a test failure.

jlsherrill:
@jjeffers feel free to merge, just don't forget to squash ;)

PR #8162, opened 2019-06-12
Fixes #27040 - RH Recommended Repos shows RHEL8

parthaa:
Add RHEL 8 channels to the recommended repositories


bbuckingham:
If not already included, please ensure that the RHEL 8 tools are also included.

PR #8163, opened 2019-06-13
Fixes #27037 - update proxy task

jjeffers:
This PR creates a new rake task for katello. The task "update_default_http_proxy" takes a http proxy name as defined under the Infrastructure->Http Proxies. 

Example:
```
[vagrant@centos7-katello-devel foreman]$ bundle exec rails katello:update_default_http_proxy -- -n a_new_proxy -u http://192.168.23.131 -s admin -p redhat
...
Content default http proxy set to a_new_proxy (http://192.168.23.131).
```

### DIFF
--- a/lib/katello/tasks/update_content_default_http_proxy.rake
+++ b/lib/katello/tasks/update_content_default_http_proxy.rake
@@ -33,15 +33,15 @@ namespace :katello do
     User.current = User.anonymous_api_admin
     http_proxy = HttpProxy.where(name: options[:name]).first
 
+    uri = URI(options[:url])
+    username = uri.user
+    password = uri.password
+
     if http_proxy
       setting.update_attribute(:value, http_proxy.name)
-      http_proxy.update_attribute(:url, options[:url])
+      http_proxy.update_attributes(url: options[:url], username: username, password: password)
       puts "Content default http proxy set to #{http_proxy.name_and_url}."
     else
-
-      uri = URI(options[:url])
-      username = uri.user
-      password = uri.password
       new_proxy = ::HttpProxy.new(name: options[:name], url: options[:url],
                                 username: username, password: password,
                                 organizations: Organization.all,
--- a/lib/katello/tasks/update_content_default_http_proxy.rake
+++ b/lib/katello/tasks/update_content_default_http_proxy.rake
@@ -33,7 +33,6 @@ namespace :katello do
     User.current = User.anonymous_api_admin
     http_proxy = HttpProxy.where(name: options[:name]).first
 
-
     if http_proxy
       setting.update_attribute(:value, http_proxy.name)
       http_proxy.update_attribute(:url, options[:url])
--- a/lib/katello/tasks/update_content_default_http_proxy.rake
+++ b/lib/katello/tasks/update_content_default_http_proxy.rake
@@ -1,0 +1,22 @@
+require File.expand_path("../engine", File.dirname(__FILE__))
+
+namespace :katello do
+  desc "Sets the content default http proxy to an existing http proxy based on supplied URL."
+  task :update_default_http_proxy, [:proxy_name] => [ :environment] do |task, args|
+    setting = ::Setting::Content.where(name: 'content_default_http_proxy').first
+
+    http_proxy = HttpProxy.where(name: args[:proxy_name]).first
+
+    if http_proxy
+      setting.update_attribute(:value, http_proxy.name)
+      puts "Content default http proxy set to #{http_proxy.name_and_url}."
+    else
+      $stderr.print "No http proxy found with name \"#{args[:proxy_name]}\"."
+    end
+  end
+
+  task http_proxy_list: [:environment] do
+    ::HttpProxy.all.each { |proxy| puts "#{proxy.name_and_url}" }
+  end
+end
+
--- a/lib/katello/tasks/update_content_default_http_proxy.rake
+++ b/lib/katello/tasks/update_content_default_http_proxy.rake
@@ -2,7 +2,7 @@ require File.expand_path("../engine", File.dirname(__FILE__))
 
 namespace :katello do
   desc "Sets the content default http proxy to an existing http proxy based on supplied URL."
-  task :update_default_http_proxy => :environment do |_task, args|
+  task :update_default_http_proxy => :environment do
     setting = ::Setting::Content.where(name: 'content_default_http_proxy').first
     options = {}
     o = OptionParser.new
@@ -17,7 +17,7 @@ namespace :katello do
       puts o
       exit
     end
-    ordered_args = o.order!(args.to_a) {}
+    ordered_args = o.order!(ARGV) {}
     o.parse!(ordered_args)
 
     unless options.key?(:name)
--- a/lib/katello/tasks/update_content_default_http_proxy.rake
+++ b/lib/katello/tasks/update_content_default_http_proxy.rake
@@ -48,7 +48,7 @@ namespace :katello do
                                 locations: Location.all)
       if new_proxy.save!
         setting.update_attribute(:value, new_proxy.name)
-        puts "Default content http proxy set to \"#{new_proxy.name_and_url}\"."
+        puts "Default content http proxy set to #{new_proxy.name_and_url}."
       end
     end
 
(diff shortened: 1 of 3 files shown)


jjeffers:
[test katello]

lib/katello/tasks/update_content_default_http_proxy.rake
```diff
@@ -0,0 +1,59 @@
...
+      $stderr.print("ERROR: Missing required option for --name HTTP_PROXY_NAME")
+      exit 2
+    end
+
+    User.current = User.anonymous_api_admin
+    http_proxy = HttpProxy.where(name: options[:name]).first
+
+    if http_proxy
+      setting.update_attribute(:value, http_proxy.name)
+      puts "Content default http proxy set to #{http_proxy.name_and_url}."
```
jlsherrill: what if the name is the same, but some option got updated like password?

lib/katello/tasks/update_content_default_http_proxy.rake
```diff
@@ -0,0 +1,59 @@
...
+        new_proxy = ::HttpProxy.new(name: options[:name], url: options[:url],
+                                  username: username, password: password,
+                                  organizations: Organization.all,
+                                  locations: Location.all)
+        if new_proxy.save!
+          setting.update_attribute(:value, new_proxy.name)
+          puts "Default content http proxy set to \"#{new_proxy.name_and_url}\"."
+        end
+      else
+        $stderr.print("ERROR: No http proxy found with name \"#{options[:name]}\".")
```
jlsherrill: this error doesn't make sense for this condition does it?  If the user didn't pass in :url, its complaining about not finding the proxy?

jjeffers: You are correct, that's in the wrong place.

lib/katello/tasks/update_content_default_http_proxy.rake
jjeffers: Ok, so if they use the "name" that matches an existing proxy, then this should update that proxy with the new url, in addition to setting the global default proxy?

lib/katello/tasks/update_content_default_http_proxy.rake
jjeffers: If the name is given without a url, then we're trying to set the global default to the proxy with that name. But if that named proxy doesn't exist, then that's why the error is given.

Is my assumption about the condition where the name is given without a url correct?

lib/katello/tasks/update_content_default_http_proxy.rake
jlsherrill: yes, the end goal is that the user can run the installer with the same options they do today, the installer will call this rake task.  Currently the users use it to add or update global proxy configuration.

lib/katello/tasks/update_content_default_http_proxy.rake
jlsherrill: since i think really this is meant to be just used to create/update an http proxy and also make it global, i would make name and url both required.

jlsherrill: as i mentioned above, i see this as mainly to be used by the installer, not by an end user.  An end user that wanted to update this information could use the api/cli

lib/katello/tasks/update_content_default_http_proxy.rake
```diff
@@ -0,0 +1,57 @@
...
+      $stderr.print("ERROR: Missing required option for --url HTTP_PROXY_URL")
+      exit 2
+    end
+
+    User.current = User.anonymous_api_admin
+    http_proxy = HttpProxy.where(name: options[:name]).first
+
+    if http_proxy
+      setting.update_attribute(:value, http_proxy.name)
+      http_proxy.update_attribute(:url, options[:url])
```
jlsherrill: don't we want to update username and password too?

jjeffers: Good point!

lib/katello/tasks/update_content_default_http_proxy.rake
```diff
@@ -0,0 +1,57 @@
+require File.expand_path("../engine", File.dirname(__FILE__))
+
+namespace :katello do
+  desc "Sets the content default http proxy to an existing http proxy based on supplied URL."
+  task :update_default_http_proxy => :environment do
+    setting = ::Setting::Content.where(name: 'content_default_http_proxy').first
+    options = {}
+    o = OptionParser.new
+    o.banner = "Usage: rake katello:update_content_default_http_proxy [options]"
```
jlsherrill: i've not ever written a rake task to use an option parser, how do you execute it with options?

```
$ rake  katello:update_default_http_proxy 
ERROR: Missing required option for --name HTTP_PROXY_NAME

[vagrant@devel2 foreman]$ rake  katello:update_default_http_proxy  --name foo
invalid option: --name
[vagrant@devel2 foreman]$ rake  katello:update_default_http_proxy  --name=foo
invalid option: --name=foo
[vagrant@devel2 foreman]$ rake  --name=foo katello:update_default_http_proxy  
invalid option: --name=foo
[vagrant@devel2 foreman]$ rake  --name foo katello:update_default_http_proxy  
invalid option: --name
[vagrant@devel2 foreman]$ rake  katello:update_default_http_proxy  -n foo
/home/vagrant/git/foreman/lib/foreman.rb:8: warning: already initialized constant Foreman::UUID_REGEXP
/home/vagrant/git/foreman/lib/foreman.rb:8: warning: previous definition of UUID_REGEXP was here
** Invoke katello:update_default_http_proxy (first_time)
** Invoke environment (first_time)
** Execute (dry run) environment
** Execute (dry run) katello:update_default_http_proxy
rake aborted!
Don't know how to build task 'foo' (See the list of available tasks with `rake --tasks`)
/home/vagrant/.gem/ruby/gems/rake-12.3.2/exe/rake:27:in `<top (required)>'
(See full trace by running task with --trace)

```

jjeffers: Take a look at the PR comment near the top for how I used it.

`bundle exec rails katello:update_default_http_proxy -- -n another_proxy -u http://admin:admin@192.168.23.131:8888`

Note that you have to use the `--` so that the options are seen by rake, I think otherwise they are interpreted as options for the ruby interpreter.

jlsherrill: ohh right the -- .  Mind updating the usage to include that?

jlsherrill: oh and you might print usage if name is missing, otherwise you have to know the -- trick to even print help text

jjeffers: It was frustrating to figure out because the recommended pattern for using optparse is to supply args in the task definition. However, that approach had issues when trying to run them via a test. So I decided to use the ARGV array instead.

lib/katello/tasks/update_content_default_http_proxy.rake
```diff
@@ -0,0 +1,57 @@
...
+    User.current = User.anonymous_api_admin
+    http_proxy = HttpProxy.where(name: options[:name]).first
+
+    uri = URI(options[:url])
+    username = uri.user
+    password = uri.password
+
+    if http_proxy
+      setting.update_attribute(:value, http_proxy.name)
+      http_proxy.update_attributes(url: options[:url], username: username, password: password)
```
jlsherrill: since we're passing in user/pass as part of the url, i think we'd want to strip it from the url when saving it both here and below, otherwise the password is displayed as part of the url even if its not used that way in the application.

jlsherrill: also, anywhere we print the url, don't print the password, we don't want it appearing in logs whenever possible.

jlsherrill:
One last comment, but then i think this is good!

lib/katello/tasks/update_content_default_http_proxy.rake
jlsherrill: yeah, i think what you have is fine, its probably better than just using ENV variables which we typically do, we just want to make sure its obvious how to run it when you don't pass any options :)

lib/katello/tasks/update_content_default_http_proxy.rake
jlsherrill: looks like the url is still getting saved with the username and password, which we don't want.

jjeffers:
[test katello]

jlsherrill:
[test katello]

jjeffers:
[test katello]

PR #8164, opened 2019-06-18
Fixes #26991 - Add Ansible Collections

sjha4:
To test, copy the change in katello.yaml.example to foreman/config/settings.plugins.d/katello.yaml and restart server after installing the binding gem. The one-VM setup for Pulp3 has the pulp_ansible enabled.

- Read documentation for the pulp-ansible plugin here : https://pulp-ansible.readthedocs.io/en/latest/
- Read API documentation for the plugin here: https://pulp-ansible.readthedocs.io/en/latest/restapi.html
- Ruby binding gem for pulp_ansible can be found here: https://rubygems.org/gems/pulp_ansible_client

### DIFF
--- a/app/controllers/katello/api/v2/repositories_controller.rb
+++ b/app/controllers/katello/api/v2/repositories_controller.rb
@@ -54,6 +54,7 @@ module Katello
       param :deb_architectures, String, :desc => N_("comma separated list of architectures to be synched from deb-archive")
       param :ignore_global_proxy, :bool, :desc => N_("if true, will ignore the globally configured proxy when syncing")
       param :ignorable_content, Array, :desc => N_("List of content units to ignore while syncing a yum repository. Must be subset of %s") % RootRepository::IGNORABLE_CONTENT_UNIT_TYPES.join(",")
+      param :ansible_collection_whitelist, String, :desc => N_("Name of collection to sync from URL")
     end
 
     def_param_group :repo_create do
@@ -451,6 +452,7 @@ module Katello
              ]
 
       keys += [{:docker_tags_whitelist => []}, :docker_upstream_name] if params[:action] == 'create' || @repository&.docker?
+      keys += [:ansible_collection_whitelist] if params[:action] == 'create' || @repository&.ansible_collection?
       keys += [:label, :content_type] if params[:action] == "create"
       if params[:action] == 'create' || @repository.custom?
         keys += [:url, :gpg_key_id, :ssl_ca_cert_id, :ssl_client_cert_id, :ssl_client_key_id, :unprotected, :name,
@@ -482,6 +484,7 @@ module Katello
       root.upstream_username = repo_params[:upstream_username] if repo_params.key?(:upstream_username)
       root.upstream_password = repo_params[:upstream_password] if repo_params.key?(:upstream_password)
       root.ignorable_content = repo_params[:ignorable_content] if root.yum? && repo_params.key?(:ignorable_content)
+      root.ansible_collection_whitelist = repo_params[:ansible_collection_whitelist] if root.ansible_collection?
 
       if root.ostree?
         root.ostree_upstream_sync_policy = repo_params[:ostree_upstream_sync_policy]
--- a/app/lib/actions/pulp3/orchestration/repository/generate_metadata.rb
+++ b/app/lib/actions/pulp3/orchestration/repository/generate_metadata.rb
@@ -5,11 +5,12 @@ module Actions
         class GenerateMetadata < Pulp3::Abstract
           def plan(repository, smart_proxy, options = {})
             options[:contents_changed] = (options && options.key?(:contents_changed)) ? options[:contents_changed] : true
+            publication_content_type = !::Katello::RepositoryTypeManager.find(repository.content_type).pulp3_skip_publication
             sequence do
               plan_action(Actions::Pulp3::Repository::CreateVersion, repository, smart_proxy) if options[:repository_creation]
-              if options[:source_repository]
+              if options[:source_repository] && publication_content_type
                 repository.update_attributes!(publication_href: options[:source_repository].publication_href)
-              else
+              elsif publication_content_type
                 plan_action(Actions::Pulp3::Repository::CreatePublication, repository, smart_proxy, options)
               end
 
--- a/app/models/katello/ansible_collection.rb
+++ b/app/models/katello/ansible_collection.rb
@@ -11,7 +11,7 @@ module Katello
 
     scoped_search :on => :name, :complete_value => true
     scoped_search :on => :namespace, :complete_value => true
-    scoped_search :on => :checksum
+    scoped_search :on => :version, :complete_value => true
 
     def self.default_sort
       order(:name)
--- a/app/models/katello/ansible_collection.rb
+++ b/app/models/katello/ansible_collection.rb
@@ -1,0 +1,32 @@
+module Katello
+  class AnsibleCollection < ApplicationRecord
+    include Concerns::PulpDatabaseUnit
+
+    self.table_name = 'katello_ansible_collections'
+
+    CONTENT_TYPE = 'ansible collection'.freeze
+
+    has_many :repository_ansible_collections, :class_name => "Katello::RepositoryAnsibleCollection", :dependent => :destroy, :inverse_of => :ansible_collection, :foreign_key => :ansible_collection_id
+    has_many :repositories, :through => :repository_ansible_collections, :class_name => "Katello::Repository"
+
+    scoped_search :on => :name, :complete_value => true
+    scoped_search :on => :namespace, :complete_value => true
+    scoped_search :on => :checksum
+
+    def self.default_sort
+      order(:name)
+    end
+
+    def self.repository_association_class
+      RepositoryAnsibleCollection
+    end
+
+    def self.unit_id_field
+      'ansible_collection_id'
+    end
+
+    def self.total_for_repositories(repos)
+      self.in_repositories(repos).count
+    end
+  end
+end
--- a/app/models/katello/repository.rb
+++ b/app/models/katello/repository.rb
@@ -25,6 +25,7 @@ module Katello
     PUPPET_TYPE = 'puppet'.freeze
     DOCKER_TYPE = 'docker'.freeze
     OSTREE_TYPE = 'ostree'.freeze
+    ANSIBLE_COLLECTION_TYPE = 'ansible_collection'.freeze
 
     define_model_callbacks :sync, :only => :after
 
@@ -90,6 +91,9 @@ module Katello
     has_many :repository_module_streams, class_name: "Katello::RepositoryModuleStream", dependent: :delete_all
     has_many :module_streams, through: :repository_module_streams
 
+    has_many :repository_ansible_collections, :class_name => "Katello::RepositoryAnsibleCollection", :dependent => :delete_all
+    has_many :ansible_collections, :through => :repository_ansible_collections
+
     # rubocop:disable HasAndBelongsToMany
     # TODO: change this into has_many :through association
     has_and_belongs_to_many :filters, :class_name => "Katello::ContentViewFilter",
@@ -116,6 +120,7 @@ module Katello
     scope :puppet_type, -> { with_type(PUPPET_TYPE) }
     scope :docker_type, -> { with_type(DOCKER_TYPE) }
     scope :ostree_type, -> { with_type(OSTREE_TYPE) }
+    scope :ansible_collection_type, -> { with_type(ANSIBLE_COLLECTION_TYPE) }
     scope :non_puppet, -> { with_type(RepositoryTypeManager.repository_types.keys - [PUPPET_TYPE]) }
     scope :non_archived, -> { where('environment_id is not NULL') }
     scope :archived, -> { where('environment_id is NULL') }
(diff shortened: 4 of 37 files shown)


app/lib/actions/katello/repository/metadata_generate.rb
```diff
@@ -15,14 +15,16 @@ def plan(repository, options = {})
...
 
           plan_pulp_action([Pulp::Repository::DistributorPublish, Pulp3::Orchestration::Repository::GenerateMetadata],
                         repository, SmartProxy.pulp_master,
                         :force => force,
                         :source_repository => source_repository,
                         :matching_content => matching_content,
                         :dependency => dependency,
-                        :repository_creation => repository_creation)
+                        :repository_creation => repository_creation,
+                        :publication_content_type => publication_content_type)
```
jlsherrill: any reason this needs to be determined here?  can it just be handled in the pulp3 orchestration layer?

app/models/katello/repository.rb
```diff
@@ -145,15 +150,15 @@ class Repository < Katello::Model
...
+    delegate :yum?, :docker?, :puppet?, :deb?, :file?, :ostree?, :ansible_collection?, :to => :root
     delegate :name, :label, :docker_upstream_name, :url, :to => :root
 
     delegate :name, :created_at, :updated_at, :major, :minor, :gpg_key_id, :gpg_key, :content_id, :arch, :label, :url, :unprotected,
              :content_type, :product_id, :checksum_type, :docker_upstream_name, :mirror_on_sync, :"mirror_on_sync?",
              :download_policy, :verify_ssl_on_sync, :"verify_ssl_on_sync?", :upstream_username, :upstream_password,
              :ostree_upstream_sync_policy, :ostree_upstream_sync_depth, :deb_releases, :deb_components, :deb_architectures,
              :ignore_global_proxy, :ssl_ca_cert_id, :ssl_ca_cert, :ssl_client_cert, :ssl_client_cert_id, :ssl_client_key_id,
-             :ssl_client_key, :ignorable_content, :description, :docker_tags_whitelist, :to => :root
+             :ssl_client_key, :ignorable_content, :description, :docker_tags_whitelist, :ansible_collection_whitelist, :to => :root
```
jlsherrill: lets not delegate new attributes to root, and just call them on root directly?

app/services/katello/pulp3/ansible_collection.rb
```diff
@@ -0,0 +1,47 @@
...
+        response = {}
+        Enumerator.new do |yielder|
+          loop do
+            page_opts = page_opts.with_indifferent_access
+            break unless (response["next"] || page_opts["page"] == 1)
+            response = SmartProxy.pulp_master!.pulp3_api.content_ansible_collections_list page_opts
+            response = response.as_json.with_indifferent_access
+            yielder.yield response[:results]
+            page_opts[:page] += 1
+          end
```
jlsherrill: can we DRY this up between file and ansible_collection?

app/services/katello/pulp3/repository/ansible_collection.rb
```diff
@@ -0,0 +1,95 @@
...
+              pulp3_api.remotes_ansible_collection_delete(href)
+            end
+          else
+            if repo.remote_href?
+              pulp3_api.remotes_ansible_collection_partial_update(repo.remote_href, remote_options)
+            else
+              create_remote
+            end
+          end
+        end
```
jlsherrill: I think we could DRY this up as well

katello.gemspec
```diff
@@ -44,6 +44,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency "anemone"
   gem.add_dependency "pulpcore_client"
   gem.add_dependency "pulp_file_client", "< 1.0"
+  gem.add_dependency "pulp_ansible_client", "0.2.0b1.dev0.1560866833"
```
jlsherrill: Reminder for me to do a packaging PR before this gets merged

app/controllers/katello/api/v2/repositories_controller.rb
```diff
@@ -482,6 +484,7 @@ def construct_repo_from_params(repo_params)
       root.upstream_username = repo_params[:upstream_username] if repo_params.key?(:upstream_username)
       root.upstream_password = repo_params[:upstream_password] if repo_params.key?(:upstream_password)
       root.ignorable_content = repo_params[:ignorable_content] if root.yum? && repo_params.key?(:ignorable_content)
+      root.ansible_collection_whitelist = repo_params[:ansible_collection_whitelist] if root.ansible_collection?
```
jlsherrill: As part of adding this new type, it'd be great to identify places where we can make the code more extensible, and hardcode type-specific entries less often.  Maybe lets get together and brainstorm and document some ways of doing it, even if we do it at a later time.

sjha4: In this case specifically, we could just ignore the check and let the field be set to nil?

sjha4:
[test katello]

sjha4:
[test katello]

app/lib/actions/pulp3/orchestration/repository/generate_metadata.rb
```diff
@@ -5,11 +5,12 @@ module Repository
         class GenerateMetadata < Pulp3::Abstract
           def plan(repository, smart_proxy, options = {})
             options[:contents_changed] = (options && options.key?(:contents_changed)) ? options[:contents_changed] : true
+            publication_content_type = !(repository.content_type == "ansible_collection")
```
jlsherrill: could we add some smarts into the repository type registration to know about this?  Maybe something like:

```
Katello::RepositoryTypeManager.register(::Katello::Repository::FILE_TYPE) do
  allow_creation_by_user true
  service_class Katello::Pulp::Repository::File
  pulp3_service_class Katello::Pulp3::Repository::File
  pulp3_plugin 'pulp_file'
  pulp3_uses_publications

  content_type Katello::FileUnit, :pulp2_service_class => ::Katello::Pulp::FileUnit, :pulp3_service_class => ::Katello::Pulp3::FileUnit, :user_removable => true
end

```

to indicate that file type uses publications (and anything that doesn't declare that, does not)

app/models/katello/repository.rb
sjha4: Sticking with this for now based on chats with Justin.. Will refactor if required later.

katello.gemspec
jlsherrill: PR opened: https://github.com/theforeman/foreman-packaging/pull/3858

app/controllers/katello/api/v2/repositories_controller.rb
```diff
@@ -54,6 +54,7 @@ def custom_index_relation(collection)
       param :deb_architectures, String, :desc => N_("comma separated list of architectures to be synched from deb-archive")
       param :ignore_global_proxy, :bool, :desc => N_("if true, will ignore the globally configured proxy when syncing")
       param :ignorable_content, Array, :desc => N_("List of content units to ignore while syncing a yum repository. Must be subset of %s") % RootRepository::IGNORABLE_CONTENT_UNIT_TYPES.join(",")
+      param :ansible_collection_whitelist, String, :desc => N_("Name of collection to sync from URL")
```
sean797: Calling this `ansible_collection_whitelist` implies it should be an array to me.

db/migrate/20190617142328_create_katello_ansible_collections.rb
```diff
@@ -0,0 +1,26 @@
...
+      t.string :pulp_id, :null => false, :limit => 255
+      t.string :checksum
+      t.string :name
+      t.string :namespace
+      t.string :version
+      t.timestamps
+    end
+
+    add_index :katello_ansible_collections, :pulp_id, :unique => true, :name => 'katello_ansible_collections_pulp_id_index'
+    add_index :katello_ansible_collections, [:id, :pulp_id, :name, :version, :namespace], :name => 'katello_ansible_collections_fields_index'
```
sean797: Do we really need to index every attribute?

app/models/katello/ansible_collection.rb
```diff
@@ -0,0 +1,32 @@
...
+    self.table_name = 'katello_ansible_collections'
+
+    CONTENT_TYPE = 'ansible collection'.freeze
+
+    has_many :repository_ansible_collections, :class_name => "Katello::RepositoryAnsibleCollection", :dependent => :destroy, :inverse_of => :ansible_collection, :foreign_key => :ansible_collection_id
+    has_many :repositories, :through => :repository_ansible_collections, :class_name => "Katello::Repository"
+
+    scoped_search :on => :name, :complete_value => true
+    scoped_search :on => :namespace, :complete_value => true
+    scoped_search :on => :checksum
```
sean797: Can you add a search for `version` ?

app/controllers/katello/api/v2/repositories_controller.rb
sean797: Also, the description should say `Ansible Collection` IMHO

sean797: Aplogies, one more thing - Are you going to have a `whitelist` function for other content types with Pulp 3? If so, it would make sence to use a more generic name.

jlsherrill: we do also have a whitelist for docker too, thats not a bad idea!  We'd have to depreciate the old docker one, and move it over to the common one.  Although then if pulp ever supported whitelisting of multiple types within a repository, that wouldn't work too well.... hrmmmm  Let me think about it

db/migrate/20190617142328_create_katello_ansible_collections.rb
sjha4: Removed indexing for all fields other than pulp_id.

sjha4:
[test katello]

sjha4:
@jlsherrill : updated pulp_ansible_client gem to 0.2.0b1.dev0.1561471371

sjha4:
Pinning gem to <=0.2.0b1.dev0.1561386170

db/migrate/20190618034438_add_ansible_collection_whitelist_to_katello_root_repositories.rb
```diff
@@ -0,0 +1,5 @@
+class AddAnsibleCollectionWhitelistToKatelloRootRepositories < ActiveRecord::Migration[5.2]
+  def change
+    add_column :katello_root_repositories, :ansible_collection_whitelist, :string
```
jlsherrill: I think we want this to be :text as to not have a 255 char limit

jlsherrill:
image

* Could we 'suggest' the default galaxy url? Kinda like what we do with docker.
* white list seems required (i can create without a whitelist but can't sync), we should probably indicate that if that is intentional in pulp
* an example of what to type into the whitelist would be very helpful, when i try to do multiple moduels i can't seem to get it to work and the errors i've seen imply i can paste 'requirements' in 
* should whitelist be a larger text box so you can put more text in it easily?

sjha4:
> Could we 'suggest' the default galaxy url? Kinda like what we do with docker.

Added a help text `"URL of the ansible content sharing hub you want to sync. Example: https://galaxy.ansible.com"`
> white list seems required (i can create without a whitelist but can't sync), we should probably indicate that if that is intentional in pulp

Added a model validation.
> an example of what to type into the whitelist would be very helpful, when i try to do multiple moduels i can't seem to get it to work and the errors i've seen imply i can paste 'requirements' in

Added a help text ` Name / Comma-separated names of ansible collection(s) you want to sync from the galaxy. ex: newswangerd.collection_demo, devoperate.base`
> should whitelist be a larger text box so you can put more text in it easily?

Made the field a text-area field.

jlsherrill:
Looking good!  

* When syncing i end up with the message 'New Content Units: 0'  even when new units had been synced.
* I'm getting an error when re-syncing a repo, the task starts but fails with error: Missing the required parameter 'repository_version_href' when calling RepositoriesApi.repositories_versions_read

sjha4:
@jlsherrill: I have a few issues open with pulp to address these:
1. https://pulp.plan.io/issues/5023 : Ansible Collection sync does not have data in the progress report
2. https://pulp.plan.io/issues/4997 : Pulp3 Ansible Collection remote doesn't seem to be able to sync more than 1 repo
3. https://pulp.plan.io/issues/5018 : Add artifact sha details to the content ansible collection list API

These should take care of the issues with sync and indexing.

sjha4:
> I'm getting an error when re-syncing a repo, the task starts but fails with error: Missing the required parameter 'repository_version_href' when calling RepositoriesApi.repositories_versions_read . This may not be ansible specific

This is ansible specific. There is a fix for this available but that is post the binding changes so it's not working with this PR. The issue tracking this was https://pulp.plan.io/issues/4920 wherein new repo versions were not created on re-sync. That is what is causing the error you see on re-sync.

sjha4:
Thanks @jlsherrill , @sean797 ..

app/controllers/katello/api/v2/repositories_controller.rb
jlsherrill: so i chatted with samir about this, and i initially liked the idea, but if we ever have some plugin that have multiple whitelists for content types, or some other kind of format, we'd need other parameters for that.  

I think we're going to have this explosion in repository attributes regardless that we need to work on dealing with that.  As we continue to add content types it may become apparent that this was a bad decision, and we can revisit it :)

PR #8166, opened 2019-06-19
Fixes #27080 - pulp3 not in ping if disabled

parthaa:
Pulp3 status information will not show up in ping if pulp3 is disabled
on the smart proxy. Check the issue for the steps.


app/models/katello/ping.rb
```diff
@@ -13,6 +13,10 @@ class << self
       def ping(services: SERVICES, capsule_id: nil)
         result = {}
         services.each { |service| result[service] = {} }
+        if capsule_id && SmartProxy.find(capsule_id).pulp3_enabled?
```
jlsherrill: could capsule_id be nil here?  (Should you default to master if it is?

app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -136,6 +136,10 @@ def backend_service_type(repository)
         end
       end
 
+      def pulp3_enabled?
+        !self.capabilities('Pulp3').nil?
```
jlsherrill: This isn't the best way to test, i think has_feature?('Pulp3')  is better.  What you're doing is fetching the capabiltiies and checking if its nil, which is kind of a side effect of it not having the pulp3 feature

app/models/katello/ping.rb
parthaa: Yes. It can be null. ```curl -k http://localhost:3000/katello/api/ping``` had nil for capsule

parthaa:
[test katello]

PR #8167, opened 2019-06-19
Fixes #27062 - Fail to delete subscriptions from UI

ianballou:
Two issues are being solved here:  
- Firstly, the bug described in [#27062](https://projects.theforeman.org/issues/27062) occurred because the upstream consumer was deleted sometime before trying to delete the subscription in the UI.  The issue here was the ambiguous error messages in the task.  Now, the upstream consumer is checked and the error messaged introduced [here](https://github.com/Katello/katello/commit/444db0c5d3d91182c1873c6f1deb5281b7d58a23) is displayed.
    - To reproduce the issue, import a fresh manifest and delete the upstream consumer (subscription allocation).  Then try to delete a subscription that belongs to that manifest.
- Secondly, during testing, it was discovered that there was another ambiguous error message in the task when trying to delete a subscription that had previously been deleted.  The fix is to hide the error, refresh the manifest, and log the missing subscription in the task Dynflow.
  - To reproduce the issue, import a fresh manifest and delete one of the subscriptions.  Delete the manifest and re-upload it.  Try to delete the subscription again.

### DIFF
--- a/app/lib/actions/katello/upstream_subscriptions/remove_entitlement.rb
+++ b/app/lib/actions/katello/upstream_subscriptions/remove_entitlement.rb
@@ -6,10 +6,13 @@ module Actions
 
         input_format do
           param :entitlement_id
+          param :sub_name
         end
 
         def run
           output[:response] = ::Katello::Resources::Candlepin::UpstreamConsumer.remove_entitlement(input[:entitlement_id])
+        rescue ::Katello::Errors::UpstreamEntitlementGone
+          output[:response] = _("%s has already been deleted" % input[:sub_name])
         end
 
         def run_progress_weight
--- a/app/lib/actions/katello/upstream_subscriptions/remove_entitlements.rb
+++ b/app/lib/actions/katello/upstream_subscriptions/remove_entitlements.rb
@@ -3,6 +3,7 @@ module Actions
     module UpstreamSubscriptions
       class RemoveEntitlements < Actions::Base
         def plan(pool_ids = [])
+          ::Katello::Resources::Candlepin::UpstreamConsumer.get(:include_only => [:uuid])
           ids = pool_ids.uniq.compact
           fail _("No pool IDs were provided.") if ids.blank?
           fail _("Current organization is not set.") unless ::Organization.current
@@ -13,7 +14,9 @@ module Actions
 
               fail _("Provided pool with id %s has no upstream entitlement" % pid) if pool.upstream_entitlement_id.nil?
 
-              plan_action(::Actions::Katello::UpstreamSubscriptions::RemoveEntitlement, entitlement_id: pool.upstream_entitlement_id)
+              sub_name = pool.subscription.name
+
+              plan_action(::Actions::Katello::UpstreamSubscriptions::RemoveEntitlement, entitlement_id: pool.upstream_entitlement_id, sub_name: sub_name)
             end
 
             plan_action(::Actions::Katello::Organization::ManifestRefresh, ::Organization.current)
--- a/app/lib/katello/resources/candlepin/upstream_consumer.rb
+++ b/app/lib/katello/resources/candlepin/upstream_consumer.rb
@@ -9,10 +9,21 @@ module Katello
             super(id)
           end
 
+          # Overrides the HttpResource get method to check if the upstream
+          # consumer exists.
+          def get(params)
+            includes = params.key?(:include_only) ? "&" + included_list(params.delete(:include_only)) : ""
+            JSON.parse(super(path + hash_to_query(params) + includes, self.default_headers).body)
+          rescue RestClient::Gone
+            raise ::Katello::Errors::UpstreamConsumerGone
+          end
+
           def remove_entitlement(entitlement_id)
             fail ArgumentError, "No entitlement ID given to remove." if entitlement_id.blank?
 
             self["entitlements/#{entitlement_id}"].delete
+          rescue RestClient::NotFound
+            raise ::Katello::Errors::UpstreamEntitlementGone
           end
 
           def export(url, client_cert, client_key, ca_file)
--- a/test/actions/katello/upstream_subscriptions/remove_entitlement_test.rb
+++ b/test/actions/katello/upstream_subscriptions/remove_entitlement_test.rb
@@ -8,11 +8,16 @@ describe ::Actions::Katello::UpstreamSubscriptions::RemoveEntitlement do
   before :all do
     @org = FactoryBot.create(:katello_organization)
     set_organization(@org)
-    @planned_action = create_and_plan_action(subject, org_id: @org.id, entitlement_id: 'foo')
+    @planned_action = create_and_plan_action(subject, org_id: @org.id, entitlement_id: 'foo', sub_name: 'Test Sub')
   end
 
   it 'runs' do
     ::Katello::Resources::Candlepin::UpstreamConsumer.expects(:remove_entitlement).with('foo')
     run_action @planned_action
+  end
+
+  it 'outputs a message if entitlement is gone' do
+    ::Katello::Resources::Candlepin::UpstreamConsumer.expects(:remove_entitlement).with('foo').raises(::Katello::Errors::UpstreamEntitlementGone)
+    assert_equal("Test Sub has already been deleted", run_action(@planned_action).output[:response])
   end
 end
--- a/app/lib/katello/errors.rb
+++ b/app/lib/katello/errors.rb
@@ -123,5 +123,7 @@ module Katello
           "Please create a new Subscription Allocation and import the new manifest.")
       end
     end
+
+    class UpstreamEntitlementGone < StandardError; end
   end
 end
--- a/test/actions/katello/upstream_subscriptions/remove_entitlements_test.rb
+++ b/test/actions/katello/upstream_subscriptions/remove_entitlements_test.rb
@@ -23,10 +23,12 @@ describe ::Actions::Katello::UpstreamSubscriptions::RemoveEntitlements do
     ::Katello::Pool.expects(:find).with(pool1.id).returns(pool1)
     ::Katello::Pool.expects(:find).with(pool2.id).returns(pool2)
 
+    ::Katello::Resources::Candlepin::UpstreamConsumer.expects(:get).returns(true)
+
     plan_action(@action, [pool1.id, pool2.id])
 
-    assert_action_planned_with(@action, @remove_action_class, entitlement_id: 'a')
-    assert_action_planned_with(@action, @remove_action_class, entitlement_id: 'b')
+    assert_action_planned_with(@action, @remove_action_class, entitlement_id: 'a', sub_name: 'basic subscription')
+    assert_action_planned_with(@action, @remove_action_class, entitlement_id: 'b', sub_name: 'other subscription')
     assert_action_planned_with(@action, @manifest_action_class, @org)
   end
 
@@ -34,17 +36,27 @@ describe ::Actions::Katello::UpstreamSubscriptions::RemoveEntitlements do
     pool1 = katello_pools(:pool_one)
     pool1.expects(:upstream_entitlement_id).returns(nil)
     ::Katello::Pool.expects(:find).with(pool1.id).returns(pool1)
+    ::Katello::Resources::Candlepin::UpstreamConsumer.expects(:get).returns(true)
 
     error = proc { plan_action(@action, [pool1.id]) }.must_raise RuntimeError
     error.message.must_match(/upstream/)
   end
 
+  it 'raises an error when the pool has no upstream allocation' do
+    pool1 = katello_pools(:pool_one)
+    ::Katello::HttpResource.expects(:get).raises(::Katello::Errors::UpstreamConsumerGone)
+
+    proc { plan_action(@action, [pool1.id]) }.must_raise ::Katello::Errors::UpstreamConsumerGone
+  end
+
   it 'raises an error when given no pool ids' do
+    ::Katello::Resources::Candlepin::UpstreamConsumer.expects(:get).returns(true)
     error = proc { plan_action(@action, []) }.must_raise RuntimeError
     error.message.must_match(/provided/)
   end
 
   it 'raises an error when no organization is set' do
+    ::Katello::Resources::Candlepin::UpstreamConsumer.expects(:get).returns(true)
     set_organization(nil)
     error = proc { plan_action(@action, [:foo]) }.must_raise RuntimeError
     error.message.must_match(/is not set/)


app/lib/actions/katello/upstream_subscriptions/remove_entitlements.rb
```diff
@@ -3,17 +3,30 @@ module Katello
     module UpstreamSubscriptions
       class RemoveEntitlements < Actions::Base
         def plan(pool_ids = [])
+          organization = Object::Organization.current
+          url = "#{organization.owner_details['upstreamConsumer']['apiUrl']}#{organization.owner_details['upstreamConsumer']['uuid']}"
```
jturel: the URL and Cert code here - I feel this knowledge should be limited to the resource class. Could this be moved to
`Katello::Resources::Candlepin::UpstreamConsumer.get` ?

ianballou: I think that's a good way to make this code reusable.  `UpstreamConsumer.get` already exists since it's inherited, did you mean I should overload that?

jturel: ah! does that method already work then? doesn't look like we're using it anywhere so it might be broken in some way

ianballou: I haven't tried using `UpstreamConsumer.get` since I haven't seen us using it at all, I've only seem `UpstreamConsumer.resource(...).get` like I'm doing on line 13.
Maybe instead I could write some methods in `upstream_consumer.rb` to fetch the url, cert, and key?

jturel: as far as I know most of the resource classes have a .get method so it may come to implementing one (for consistency & convenience)  to do exactly what you did here in the action

ianballou:
@jturel is the get method I created along the lines of what you were thinking? The behaviour is the same.

app/lib/katello/resources/candlepin/upstream_consumer.rb
```diff
@@ -9,6 +9,15 @@ def path(id = upstream_consumer_id)
             super(id)
           end
 
+          def get(organization)
```
jturel: It looks like the UpstreamConsumer model already has a lot of this implemented. This approach seems to work just the same:

```
def get
  JSON.parse(resource.get)
end
```

app/lib/actions/katello/upstream_subscriptions/remove_entitlements.rb
```diff
@@ -3,17 +3,26 @@ module Katello
     module UpstreamSubscriptions
       class RemoveEntitlements < Actions::Base
         def plan(pool_ids = [])
+          organization = Object::Organization.current
+
+        begin
```
jturel: I think the RestClient::Gone exception shouldn't be reaching up this high for a few reasons. I prefer to see knowledge of our HTTP client (RestClient) limited to the Resource:: namespace. That'll make our lives easier when we inevitably change clients.

The answer to this is to have the .get method itself raise UpstreamConsumerGone similar to this:

https://github.com/Katello/katello/blob/fe7244b506f3439757d821be3516eb2f460ac97a/app/lib/katello/resources/candlepin/upstream_pool.rb#L8-L12

app/lib/actions/katello/upstream_subscriptions/remove_entitlements.rb
```diff
@@ -3,17 +3,26 @@ module Katello
...
+        end
+
           ids = pool_ids.uniq.compact
           fail _("No pool IDs were provided.") if ids.blank?
           fail _("Current organization is not set.") unless ::Organization.current
 
           sequence do
             ids.each do |pid|
               pool = ::Katello::Pool.find(pid)
+              sub_name = pool.subscription.name
```
jturel: this assignment can probably be moved to below the 'fail' since in that situation it's a meaningless assignment :)

ianballou:
@jturel I think I addressed all of your comments.

jturel:
[test katello]

jturel:
[test katello]

app/lib/actions/katello/upstream_subscriptions/remove_entitlement.rb
```diff
@@ -6,10 +6,13 @@ class RemoveEntitlement < Actions::Base
 
         input_format do
           param :entitlement_id
+          param :sub_name
         end
 
         def run
           output[:response] = ::Katello::Resources::Candlepin::UpstreamConsumer.remove_entitlement(input[:entitlement_id])
+        rescue RestClient::NotFound
```
jturel: Can you give remove_entitlement the same treatment and raise a custom error which masks RestClient::NotFound?

app/lib/actions/katello/upstream_subscriptions/remove_entitlements.rb
```diff
@@ -3,6 +3,8 @@ module Katello
     module UpstreamSubscriptions
       class RemoveEntitlements < Actions::Base
         def plan(pool_ids = [])
+          ::Katello::Resources::Candlepin::UpstreamConsumer.get
```
jturel: Candlepin APIs accept an 'includes' parameter in order to filter out the attributes returned in the response. Since we don't actually need any of the data here it'd be great to add that ability to .get so we can get only a single attribute for performance reasons.

You can refer to existing examples in the resource classes but the usage would end up something like this:

```
::Katello::Resources::Candlepin::UpstreamConsumer.get(includes: [:uuid])
```

app/lib/actions/katello/upstream_subscriptions/remove_entitlement.rb
jturel: Would be nice to have a test for the behavior change here (adding a custom message into the output response)

ianballou: Sounds good, my latest commit should address these.

app/lib/actions/katello/upstream_subscriptions/remove_entitlement.rb
```diff
@@ -6,10 +6,13 @@ class RemoveEntitlement < Actions::Base
 
         input_format do
           param :entitlement_id
+          param :sub_name
         end
 
         def run
           output[:response] = ::Katello::Resources::Candlepin::UpstreamConsumer.remove_entitlement(input[:entitlement_id])
+        rescue ::Katello::Errors::EntitlementGone
```
jturel: Nice. One thing, can we call it UpstreamEntitlementGone? Just to make it a little more clear since we could conceivably have an error of the same name pertaining to the local Candlepin :)

ianballou: Sure thing!

ianballou:
@jturel I think I've covered all your comments!

jturel:
[test katello]

jturel:
I'm testing the scenario in which I've removed an upstream entitlement already and I'm trying to get it off my Katello. When I press Delete the action raises an error:

```
2019-06-26T20:47:47 [E|bac|] Entitlement with id 2fafe7458a4e4677ad2c7f4bf04de884 could not be found. (Katello::Errors::CandlepinError) 
 | /home/vagrant/katello/app/lib/actions/middleware/propagate_candlepin_errors.rb:21:in `rescue in propagate_candlepin_errors'                      
 | /home/vagrant/katello/app/lib/actions/middleware/propagate_candlepin_errors.rb:18:in `propagate_candlepin_errors'                                                 
 | /home/vagrant/katello/app/lib/actions/middleware/propagate_candlepin_errors.rb:9:in `run'                                                                         
 | /home/vagrant/foreman/.vendor/ruby/2.5.0/gems/dynflow-1.2.3/lib/dynflow/middleware/stack.rb:22:in `call'                            
 | /home/vagrant/foreman/.vendor/ruby/2.5.0/gems/dynflow-1.2.3/lib/dynflow/middleware/stack.rb:26:in `pass'                               
 | /home/vagrant/foreman/.vendor/ruby/2.5.0/gems/dynflow-1.2.3/lib/dynflow/middleware.rb:18:in `pass'                                                        
 | /home/vagrant/foreman/.vendor/ruby/2.5.0/gems/foreman-tasks-0.15.6/app/lib/actions/middleware/rails_executor_wrap.rb:14:in `block in run'                                  | /home/vagrant/foreman/.vendor/ruby/2.5.0/gems/activesupport-5.2.1/lib/active_support/execution_wrapper.rb:87:in `wrap'                         
 | /home/vagrant/foreman/.vendor/ruby/2.5.0/gems/foreman-tasks-0.15.6/app/lib/actions/middleware/rails_executor_wrap.rb:13:in `run'        
```

So, your new message isn't making it to the task output and the task is in warning state. I think it's because you didn't change ::Katello::Resources::Candlepin::UpstreamConsumer.remove_entitlement to actually raise your new error

jturel:
For testing the scenario in which I want to remove a subscription and the upstream consumer is gone, I see the error notification, but the task remains in error state and doesn't refresh the manifest to remove any leftover subscriptions from the removed upstream allocation.

Let me think about whether or not we should prevent the error state and proceed with the manifest refresh.

ianballou:
@jturel thanks for catching the un-raised entitlement error, I fixed it in my latest push. Looks like I was having issues with code reloading :) 

As for the missing upstream consumer, my idea was that we wouldn't refresh the manifest since that would end up failing anyway.  I can see though that leaving the errored task is ugly, do you think it should be a warning instead?

jturel:
[test katello]

app/lib/actions/katello/upstream_subscriptions/remove_entitlement.rb
```diff
@@ -6,10 +6,13 @@ class RemoveEntitlement < Actions::Base
 
         input_format do
           param :entitlement_id
+          param :sub_name
         end
 
         def run
           output[:response] = ::Katello::Resources::Candlepin::UpstreamConsumer.remove_entitlement(input[:entitlement_id])
+        rescue ::Katello::Errors::UpstreamEntitlementGone
+          output[:response] = "entitlement (subscription) with name \"%s\" has already been deleted" % input[:sub_name]
```
jturel: What do you think about reworking this message? "%s has already been deleted". My thought is that it's a little more concise. We already have the context of what object we are dealing with when we are looking at the message in the dynflow console (::Remove**Entitlement**). I don't feel too strongly about it, so up to you!

jturel:
> I noticed that refreshing a manifest that doesn't exist results in a task that completes but gives a warning like Unit 0f096a1f-dff9-407c-b8a8-c2ce1271c606 has been deleted. I wonder if we should make that error-out instead and give the upstream consumer gone error.

Does it happen only when pressing the Refresh button? If so, I think we could take care of it as a separate issue. Let me know what you think.

Left one comment on messaging but other than that the behavior is perfect!

app/lib/actions/katello/upstream_subscriptions/remove_entitlement.rb
jturel: Oh! In any case this string needs to be surrounded by _("") so that it can be translated

ianballou:
> Does it happen only when pressing the Refresh button? If so, I think we could take care of it as a separate issue. Let me know what you think.

It does appear only to happen when hitting the Refresh button, yeah.   I suppose that would happen during a refresh happening outside of clicking the button, but my PR here should block that anyway.  That can be a separate issue.

app/lib/actions/katello/upstream_subscriptions/remove_entitlement.rb
ianballou: Yeah I was thinking it's a bit long.  I think shortening it works fine.

ianballou: Pushed the change here!

test/actions/katello/upstream_subscriptions/remove_entitlement_test.rb
```diff
@@ -15,4 +15,9 @@
     ::Katello::Resources::Candlepin::UpstreamConsumer.expects(:remove_entitlement).with('foo')
     run_action @planned_action
   end
+
+  it 'outputs a message if entitlement is gone' do
+    ::Katello::Resources::Candlepin::UpstreamConsumer.expects(:remove_entitlement).with('foo').raises(::Katello::Errors::UpstreamEntitlementGone)
+    assert_equal("entitlement (subscription) with name \"\" has already been deleted", run_action(@planned_action).output[:response])
```
jturel: test needs update too :)

ianballou: Just caught that!

test/actions/katello/upstream_subscriptions/remove_entitlement_test.rb
```diff
@@ -15,4 +15,9 @@
     ::Katello::Resources::Candlepin::UpstreamConsumer.expects(:remove_entitlement).with('foo')
     run_action @planned_action
   end
+
+  it 'outputs a message if entitlement is gone' do
+    ::Katello::Resources::Candlepin::UpstreamConsumer.expects(:remove_entitlement).with('foo').raises(::Katello::Errors::UpstreamEntitlementGone)
+    assert_equal(" has already been deleted", run_action(@planned_action).output[:response])
```
jturel: I'd like to see this test with some given input like this:

```
+++ b/test/actions/katello/upstream_subscriptions/remove_entitlement_test.rb
@@ -8,7 +8,7 @@ describe ::Actions::Katello::UpstreamSubscriptions::RemoveEntitlement do
   before :all do
     @org = FactoryBot.create(:katello_organization)
     set_organization(@org)
-    @planned_action = create_and_plan_action(subject, org_id: @org.id, entitlement_id: 'foo')
+    @planned_action = create_and_plan_action(subject, org_id: @org.id, entitlement_id: 'foo', sub_name: 'Awesome Sub')
   end
 
   it 'runs' do
@@ -18,6 +18,6 @@ describe ::Actions::Katello::UpstreamSubscriptions::RemoveEntitlement do
 
   it 'outputs a message if entitlement is gone' do
     ::Katello::Resources::Candlepin::UpstreamConsumer.expects(:remove_entitlement).with('foo').raises(::Katello::Errors::UpstreamEntitlementGone)
-    assert_equal(" has already been deleted", run_action(@planned_action).output[:response])
+    assert_equal("Awesome Sub has already been deleted", run_action(@planned_action).output[:response])
   end
 end
```

ianballou: Gotcha, fixed it

jturel:
[test katello]

PR #8170, opened 2019-06-20
Fixes #27089 - pin pulp gems

jlsherrill:
to prevent future changes from breaking us


sjha4:
Looks good to me.. :+1:  Will test the core gem in a little bit since it was published today..

PR #8171, opened 2019-06-20
Fixes #26930 - support multi-repo composite publishing

jlsherrill:
this adds support for pulp3 for publishing a composite repo
with multiple components that contain the same repository

### DIFF
--- a/app/lib/actions/pulp/repository/copy_all_units.rb
+++ b/app/lib/actions/pulp/repository/copy_all_units.rb
@@ -2,7 +2,7 @@ module Actions
   module Pulp
     module Repository
       class CopyAllUnits < Pulp::AbstractAsyncTask
-        def plan(source_repo, _smart_proxy, target_repo, options = {})
+        def plan(target_repo, _smart_proxy, source_repo, options = {})
           filter_ids = options.fetch(:filters, nil)&.map(&:id)
           rpm_filenames = options.fetch(:rpm_filenames, nil)
           solve_dependencies = options.fetch(:solve_dependencies, false)
--- a/app/lib/actions/katello/repository/clone_contents.rb
+++ b/app/lib/actions/katello/repository/clone_contents.rb
@@ -13,13 +13,11 @@ module Actions
 
           sequence do
             if copy_contents
-              source_repositories.each do |repository|
-                plan_pulp_action([Pulp3::Orchestration::Repository::CopyAllUnits, Pulp::Repository::CopyAllUnits],
-                            repository,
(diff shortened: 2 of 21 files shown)


jlsherrill:
TODO:

* [ ] write tests


To test:

1.  create a file repository and sync it, such as https://repos.fedorapeople.org/repos/pulp/pulp/fixtures/file/
2.  remove all files from it but 1 file, such as '1.iso'
3.  Publish the file repo in a content view named 'component1'
4.  resync the file repo
5.  remove all files from the file repo again except one, but leave a different file '2.iso'
6.  Publish the file repo in a content view named 'component2'
7.  create a new content view named 'composite', and make it a composite
8.  select both component1 and component2 and publish
9.  verify that both 1.iso and 2.iso are in the resulting content view.

app/lib/actions/pulp/repository/copy_all_units.rb
```diff
@@ -2,16 +2,18 @@ module Actions
...
           rpm_filenames = options.fetch(:rpm_filenames, nil)
           solve_dependencies = options.fetch(:solve_dependencies, false)
 
-          plan_self(source_repo_id: source_repo.id,
-                    target_repo_id: target_repo.id,
-                    filter_ids: filter_ids,
-                    solve_dependencies: solve_dependencies,
-                    rpm_filenames: rpm_filenames)
+          source_repos.each do |source_repo|
+            plan_self(source_repo_id: source_repo.id,
```
sjha4: We should check if we get a single repo here? I see one example here: https://github.com/Katello/katello/blob/c31deec0391d013f37b67364b9f5c0c859ad27bf/app/lib/actions/katello/repository/export.rb#L72

jlsherrill: oh right, plan_self won't work in that case!  i think i'll need to refactor out the run phase to another action.

sjha4:
Not sure I'm doing this right but with the steps above, the published composite content view seems to have 3 file units. Even the one that was deleted for both repos.

app/lib/actions/pulp/repository/copy_all_units.rb
jlsherrill: I updated this, but haven't tested my changes yet.  Will do so shortly

sjha4:
@jlsherrill: There are conflicts on this..I will test this again to see if I am able to see the issue pre changes.. :smile:

jlsherrill:
@sjha4 rebased!

PR #8172, opened 2019-06-21
Fixes #27066 - Corrects content source error message

wbclark:
Fixes #27066 - Clearing content source from hostgroup fails with error to 'unable to flatten recursive array'

On line 14, the error msg was defined to be the errors array itself, with the intended message added. When the msg gets added to the array on line 26, this caused the errors array to become a recursive array.

Thanks to @parthaa for the assistance with this issue!


jturel:
[test katello]

PR #8173, opened 2019-06-21
Fixes #27102 - hide rhel 8 meta kickstart repos

jlsherrill:


jlsherrill:
[test katello]

app/controllers/katello/api/v2/repository_sets_controller.rb
```diff
@@ -44,9 +44,7 @@ def available_repositories
 
       repos = repos.select do |repo|
         if repo[:path].include?('kickstart')
-          variants = ['Server', 'Client', 'ComputeNode', 'Workstation']
-          has_variant = variants.any? { |v| repo[:substitutions][:releasever].try(:include?, v) }
-          has_variant ? repo[:enabled] : true
+          repo[:substitutions][:releasever].include?('.') || repo[:enabled]
```
jturel: when I expanded `rhel-7-desktop-beta-kickstart` I got a stack trace:
```
2019-06-21T20:26:06 [E|app|] NoMethodError: undefined method `include?' for nil:NilClass                                    
 | /home/vagrant/katello/app/controllers/katello/api/v2/repository_sets_controller.rb:47:in `block in available_repositories'          
 | /home/vagrant/katello/app/controllers/katello/api/v2/repository_sets_controller.rb:45:in `select'                                   
 | /home/vagrant/katello/app/controllers/katello/api/v2/repository_sets_controller.rb:45:in `available_repositories'
```

Might indicate a problem with the data (didn't look into it) but would be good to show something nicer in that case. On master I get just 'x86_64' as a repo to enable and it doesnt explode when I do

jlsherrill: ahh release version is probably not a variable in that case, in which case its nil.  Good catch!

jlsherrill: updated check and added a test case

jlsherrill:
[test katello]

jlsherrill:
[test katello]

jlsherrill:
[test katello]

PR #8174, opened 2019-06-24
Refs #12723 - Fix content count factory for debian content

mdellweg:


jlsherrill:
looks like there is a small test breakage (ignore the react test failure)

mdellweg:
Oh yeah, that was a trivial oversight.

app/views/foreman/smart_proxies/_content_tab.html.erb
```diff
@@ -8,9 +8,11 @@
       <th translate>Last Published</th>
       <th translate>Hosts</th>
       <th translate>Products</th>
+      <th translate>Apt repos</th>
       <th translate>Yum repos</th>
       <th translate>Container Image repos</th>
-      <th translate>Packages</th>
+      <th translate>Debian Packages</th>
+      <th translate>Redhat Packages</th>
```
jturel: 'Red Hat Packages' :)

jlsherrill: I'd actually prefer 'RPM Packages'   but good job @jturel for keeping brand in mind ;)

jturel: yeah good call on 'RPM Packages'!

mdellweg: RPM Packages it is... ;)
Also changed Debian (the distribution) to Deb (the file format).

PR #8179, opened 2019-06-24
Fixes #26996 - support docker content

ianballou:
To test this out:

- Create a new repository:
    - type: docker
    - upstream url: https://registry.hub.docker.com/
    - upstream repository name: library/busybox
- Click 'save'
- Go to the new docker repository
- Ensure that settings can be edited without error
    - Especially ensure that Limit Sync Tags can be emptied after being set to something.
- Finally, try deleting the repository

### DIFF
--- a/app/services/katello/pulp3/repository/docker.rb
+++ b/app/services/katello/pulp3/repository/docker.rb
@@ -1,0 +1,55 @@
+require 'pulp_docker_client'
+
+module Katello
+  module Pulp3
+    class Repository
+      class Docker < ::Katello::Pulp3::Repository
+        def api_client
+          PulpDockerClient::ApiClient.new(smart_proxy.pulp3_configuration(PulpDockerClient::Configuration))
+        end
+
+        def api_exception_class
+          PulpDockerClient::ApiError
+        end
+
+        def remote_class
+          PulpDockerClient::DockerRemote
+        end
+
+        def remotes_api
+          PulpDockerClient::RemotesDockerApi.new(api_client)
+        end
+
+        def distribution_class
+          PulpDockerClient::DockerDistribution
+        end
+
+        def distributions_api
+          PulpDockerClient::DistributionsDockerApi.new(api_client)
+        end
+
+        def remote_options
+          if root.url.blank?
+            super
+          else
+            options = {url: root.url, upstream_name: root.docker_upstream_name}
+            if root.docker_tags_whitelist && root.docker_tags_whitelist.any?
+              options[:whitelist_tags] = root.docker_tags_whitelist.join(",")
+            else
+              options[:whitelist_tags] = nil
+            end
+            common_remote_options.merge(options)
+          end
+        end
+
+        def distribution_options(path)
+          {
+            base_path: path,
+            repository_version: repo.version_href,
+            name: "#{backend_object_name}"
+          }
+        end
+      end
+    end
+  end
+end
--- a/lib/katello/repository_types/docker.rb
+++ b/lib/katello/repository_types/docker.rb
@@ -1,5 +1,8 @@
 Katello::RepositoryTypeManager.register(::Katello::Repository::DOCKER_TYPE) do
   service_class Katello::Pulp::Repository::Docker
+  pulp3_service_class Katello::Pulp3::Repository::Docker
+  pulp3_skip_publication true
+  pulp3_plugin 'pulp_docker'
 
   content_type Katello::DockerManifest, :priority => 1, :pulp2_service_class => ::Katello::Pulp::DockerManifest, :user_removable => true
   content_type Katello::DockerManifestList, :priority => 2, :pulp2_service_class => ::Katello::Pulp::DockerManifestList
--- a/katello.gemspec
+++ b/katello.gemspec
@@ -45,6 +45,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency "pulpcore_client"
   gem.add_dependency "pulp_file_client", "< 3.0.0"
   gem.add_dependency "pulp_ansible_client"
+  gem.add_dependency "pulp_docker_client"
 
   # UI
   gem.add_dependency "deface", '>= 1.0.2', '< 2.0.0'
--- a/test/actions/pulp3/orchestration/docker_create_test.rb
+++ b/test/actions/pulp3/orchestration/docker_create_test.rb
@@ -1,0 +1,26 @@
+require 'katello_test_helper'
+
+module ::Actions::Pulp3
+  class DockerCreateTest < ActiveSupport::TestCase
+    include Katello::Pulp3Support
+
+    def setup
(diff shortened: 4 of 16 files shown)


chris1984:
ok to test

app/lib/actions/pulp3/orchestration/repository/generate_metadata.rb
```diff
@@ -7,9 +7,9 @@ def plan(repository, smart_proxy, options = {})
             options[:contents_changed] = (options && options.key?(:contents_changed)) ? options[:contents_changed] : true
             sequence do
               plan_action(Actions::Pulp3::Repository::CreateVersion, repository, smart_proxy) if options[:repository_creation]
-              if options[:source_repository]
+              if options[:source_repository] && repository.content_type != "docker"
                 repository.update_attributes!(publication_href: options[:source_repository].publication_href)
-              else
+              elsif repository.content_type != "docker"
```
jlsherrill: we're trying to make content types be as 'pluggable' as possible, so we might want to add this logic to the repository type registration

app/lib/pulp/v3/api.rb
```diff
@@ -29,16 +30,41 @@ def configure(opts)
...
+      delegate :'content_docker_manifest_list_tags_read', to: :docker_content_api
+      delegate :'content_docker_manifest_lists_create', to: :docker_content_api
+      delegate :'content_docker_manifest_lists_list', to: :docker_content_api
+      delegate :'content_docker_manifest_lists_read', to: :docker_content_api
+      delegate :'content_docker_manifest_tags_create', to: :docker_content_api
+      delegate :'content_docker_manifest_tags_list', to: :docker_content_api
+      delegate :'content_docker_manifest_tags_read', to: :docker_content_api
+      delegate :content_docker_manifests_create, to: :docker_content_api
+      delegate :content_docker_manifests_list, to: :docker_content_api
+      delegate :content_docker_manifests_read, to: :docker_content_api
```
jlsherrill: hrmm, i will think of a way to make this more flexible

app/services/katello/pulp3/repository/docker.rb
```diff
@@ -0,0 +1,101 @@
...
+          response = pulp3_api.remotes_docker_docker_create(remote_docker_data)
+          repo.update_attributes!(:remote_href => response._href)
+        end
+
+        def remote_options
+          if root.docker_tags_whitelist.any?
+            common_remote_options.merge({url: root.url, upstream_name: root.docker_upstream_name, whitelist_tags: root.docker_tags_whitelist})
+          else
+            # TODO: Remove placeholder once whitelist_tags supports empty string
+            common_remote_options.merge({url: root.url, upstream_name: root.docker_upstream_name, whitelist_tags: "placeholder!"})
```
jlsherrill: Lets go ahead and open  up a issue to track this

ianballou: I'm going to replace `"placeholder!"` with empty string once I un-draft this PR.  The fix in pulp-docker should be out soon.

app/lib/actions/pulp3/orchestration/repository/generate_metadata.rb
ianballou: I'm going to wait and see what we decide for the Ansible PR since it has similar changes being made.

ianballou:
`nil` whitelist_tags  works now with the latest Pulp 3. The Pulp 3 bindings changes however will cause some changes, especially in ` app/lib/pulp/v3/api.rb` since the API method names are no longer unique.

app/services/katello/pulp3/repository/docker.rb
ianballou: (Taken care of in latest commit)

lib/katello/repository_types/docker.rb
```diff
@@ -1,5 +1,7 @@
 Katello::RepositoryTypeManager.register(::Katello::Repository::DOCKER_TYPE) do
   service_class Katello::Pulp::Repository::Docker
+  pulp3_service_class Katello::Pulp3::Repository::Docker
+  pulp3_plugin 'pulp_docker'
```
sjha4: Does docker content not have a publication?

sjha4: Nvm..missed your earlier comment..Also, you could remove pry as it's not letting your tests run..

ianballou: It doesn't, no

ianballou: > Nvm..missed your earlier comment..Also, you could remove pry as it's not letting your tests run..

Oh my bad, I missed that!

app/lib/pulp/v3/api.rb
```diff
@@ -1,11 +1,12 @@
 require 'pulpcore_client'
```
ianballou: This file will need to be split out in order to work with the new bindings.

ianballou: This file will need to be split out in order to work with the new bindings.  Multiple API endpoints here will have methods with the same name.

ianballou:
I'll add that the code here, as it is, is able to create and update repositories.  Let me know if anyone finds that there are issues with this basic functionality.

ianballou:
Create, update, and delete works here and the changes were verified in the Pulp 3 database.

ianballou:
My PR here is going to be changing quite a bit to work with the new Pulp bindings: https://github.com/Katello/katello/pull/8190/files

ianballou:
I just force pushed the new docker code that supports the bindings changes.

app/services/katello/pulp3/repository/docker.rb
```diff
@@ -0,0 +1,49 @@
...
+
+        def remote_options
+          if root.url.blank?
+            super
+          else
+            if root.docker_tags_whitelist.any?
+              common_remote_options.merge(url: root.url, upstream_name: root.docker_upstream_name, whitelist_tags: root.docker_tags_whitelist.join(","))
+            else
+              common_remote_options.merge(url: root.url, upstream_name: root.docker_upstream_name, whitelist_tags: nil)
+            end
```
jlsherrill: could probably simplify this a bit to not repeat the full line

jlsherrill:
Looks good!  I'd like to see a repository creation vcr cassette recorded, similar to https://github.com/Katello/katello/blob/master/test/actions/pulp3/orchestration/file_create_test.rb

app/services/katello/pulp3/repository/docker.rb
ianballou: Fixed this in my latest commit.

ianballou:
I added some tests, let me know if I need to bolster them at all.

ianballou:
Failures here are from Ansible looks like a rebase will fix it when #8190 merges.

jlsherrill:
@ianballou this now needs a rebase

ianballou:
@jlsherrill rebase complete

PR #8178, opened 2019-06-24
Fixes #27100 - Organization and location assginment to global default http proxy.

jjeffers:
This PR adds model AR hooks to assign all organizations and locations to the HttpProxy indicated by the global http proxy Setting value. Additionally, after creating a new Organization or Location, that record is associated with the global default HttpProxy.

Example of how to verify this change from the console:
```[vagrant@centos7-katello-devel foreman]$ bundle exec rails console
...
Loading development environment (Rails 5.2.1)
Failed to load console gems, starting anyway
irb(main):002:0> Setting::Content.where(name: 'content_default_http_proxy').first
=> #<Setting::Content id: 11, name: "content_default_http_proxy", value: "--- a_new_proxy\n...\n", description: "Default HTTP Proxy for syncing content", category: "Setting::Content", settings_type: nil, default: "--- \n...", created_at: "2019-06-18 18:17:15", updated_at: "2019-06-24 14:53:21", full_name: "Default http proxy", encrypted: false>
irb(main):003:0> setting = Setting::Content.where(name: 'content_default_http_proxy').first
=> #<Setting::Content id: 11, name: "content_default_http_proxy", value: "--- a_new_proxy\n...\n", description: "Default HTTP Proxy for syncing content", category: "Setting::Content", settings_type: nil, default: "--- \n...", created_at: "2019-06-18 18:17:15", updated_at: "2019-06-24 14:53:21", full_name: "Default http proxy", encrypted: false>
irb(main):008:0> default_proxy = HttpProxy.where(name: setting.value).first
=> #<HttpProxy id: 52, name: "a_new_proxy", url: "http://192.168.23.131", username: "admin", password: "redhat">
=> #<ActiveRecord::Associations::CollectionProxy [#<Organization id: 1, name: "Default Organization", type: "Organization", created_at: "2019-06-18 18:17:27", updated_at: "2019-06-18 18:17:27", ignore_types: [], description: nil, label: "Default_Organization", ancestry: nil, title: "Default Organization", manifest_refreshed_at: nil>]>
```
Then, create a new Organization, and reload that default proxy.
```
irb(main):010:0> default_proxy.organizations
[...]

PR #8181, opened 2019-06-25
Fixes #27121 - migrate proxy usage to global content proxy

jlsherrill:
This migrates the following to use the globally set
contetn proxy:

* Repository Discovery
* Manifest actions (refresh, entitlement change)
* Redhat Repository enablement

### DIFF
--- a/app/models/katello/concerns/http_proxy_extensions.rb
+++ b/app/models/katello/concerns/http_proxy_extensions.rb
@@ -5,6 +5,12 @@ module Katello
 
       included do
         after_save :update_default_proxy_setting
+
+        def self.default_global_content_proxy
+          if Setting[:content_default_http_proxy]
+            HttpProxy.unscoped.find_by(name: Setting[:content_default_http_proxy])
+          end
+        end
       end
 
       def update_default_proxy_setting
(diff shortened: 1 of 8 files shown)


app/models/katello/concerns/http_proxy_extensions.rb
```diff
@@ -5,6 +5,12 @@ module HttpProxyExtensions
 
       included do
         after_save :update_default_proxy_setting
+
+        def self.default_global_content_proxy
```
jjeffers: Oh, this is a great addition!

jjeffers: @jlsherrill should we refactor other queries where we try to find the global default to use this method?

jlsherrill: good idea! i'll see if i can find em all

jjeffers: I think there will be a a couple more in https://github.com/Katello/katello/pull/8178.

jlsherrill: i don't actually see any in the current code base, but will change in once i rebase against #8178

jlsherrill: updated, let me know if i missed any

PR #8183, opened 2019-06-25
[WIP] Add cypress test for content credentials page

johnpmitsch:
Adds [cypress](cypress.io) to do end to end UI testing on Katello. This commit starts out with the content credentials page.

We'll use it in headless mode so it can run in the VM, but it can be used with the GUI too.

[link to community post explaining more about cypress and my proposal](https://community.theforeman.org/t/rfc-integration-tests-end-to-end-tests/12142/17?u=john_mitsch)

To use, do the following on a Katello VM:
- `npm install --save-dev cypress`
- `yum install xorg-x11-server-Xvfb.x86_64 gtk3 libXScrnSaver-devel GConf2-devel`
- `./node_modules/.bin/cypress run`
- see test output for video and screenshot location

You will need to start the server that it runs against. It runs fastest with [assets precompiled](https://github.com/theforeman/forklift/blob/master/docs/development.md#run-the-rails-server-without-a-webpack-server-running)

Best practices and tips for using cypress (I'll add this to docs somewhere)
- Stub all server requests, use cypress for front-end testing only
- Log in once in `before()` block and then store the session for each test. Cypress.Cookies.preserveOnce('_session_id', 'remember_token')
- You can debug cypress by using `DEBUG=cypress:*` prepended before the cypress command

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,5 @@ node_modules
 npm-debug.log
 /foreman/
 package-lock.json
+test/cypress/videos
+test/cypress/screenshots
--- a/cypress.json
+++ b/cypress.json
@@ -1,0 +1,11 @@
+{
+  "baseUrl": "https://localhost",
+  "fixturesFolder": "test/cypress/fixtures",
+  "integrationFolder": "test/cypress/integration",
+  "pluginsFile": "test/cypress/plugins/index.js",
+  "screenshotsFolder": "test/cypress/screenshots",
+  "videosFolder": "test/cypress/videos",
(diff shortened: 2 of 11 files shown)


johnpmitsch:
There is a [video of the output here](https://imgur.com/2O26Wrm)

johnpmitsch:
This is fully functional, I'm just keeping WIP while there are discussions around this topic and we can discuss CI integration as well

johnpmitsch:
going to close this as there are larger discussions around testing, but this can be used as an example for anyone who needs it

PR #8185, opened 2019-06-28
Fixes #26821 - move cp owner delete to finalize phase

jlsherrill:
We are currently hitting a deadlock sometimes while
deleting an organization.  The current understanding is that
the following is occuring:

1. in run phase: delete org in candlepin
2. in finalize: delete product AR models and all their associations
3. in finalize: delete the org AR model and all their associations

As part of 1), messages are sent to the candlepin event queue that are
picked up and handled asyncronously.  This asyncronous task will
try to delete a specific pool and if that pools product is being deleted
in 2) there is the possibility of a deadlock.

To work around this, lets reorder things a little to be like this:

1.  in finalize phase: delete the org in katello
2.  in finalize phase delete the org in candlepin

this delays all the messages about pool deletion until after
the AR relations are destroyed

In addition i hit an error trying to delete RH content in candlepin
'content ID is locked'.  This happens when trying to delete rh content
from manifests.  With candlepin 2.0 we don't need to remove the content
since the owner delete will handle it

### DIFF
--- a/app/lib/actions/candlepin/owner/destroy.rb
+++ b/app/lib/actions/candlepin/owner/destroy.rb
@@ -6,7 +6,7 @@ module Actions
           param :label
         end
 
-        def run
+        def finalize
           output[:response] = ::Katello::Resources::Candlepin::Owner.destroy(input[:label])
         end
       end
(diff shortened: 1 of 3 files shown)


ekohl:
Test failure looks relevant.

tbrisker:
[test katello]

jturel:
Conceptually it makes sense and org delete was successful for me locally for an org with a decent amount of products. Let's see what happens in nightlies over the week!

PR #8187, opened 2019-06-30
Refs #26393 - import boot time information

ares:
needs https://github.com/theforeman/foreman/pull/6843

### DIFF
--- a/app/models/katello/rhsm_fact_parser.rb
+++ b/app/models/katello/rhsm_fact_parser.rb
@@ -72,6 +72,10 @@ module Katello
     def ipmi_interface
     end
 
+    def boot_timestamp
+      facts['proc_stat.btime']&.to_i
+    end
+
     private
 
     def get_rhsm_ip(interface)


app/models/katello/rhsm_fact_parser.rb
```diff
@@ -72,6 +72,10 @@ def environment
     def ipmi_interface
     end
 
+    def boot_time
+      facts['proc_stat::btime']&.to_i
```
ezr-ondrej: Shouldn't this be `proc_stat.btime`?

ares:
of course, thanks for pointing that out

ares:
@ezr-ondrej I belive you tested this recently, perhaps you could ack :-)

jlsherrill:
[test katello]

jlsherrill:
have never seen this test failure before, but let's [test katello]

ares:
:green_apple:

PR #8188, opened 2019-07-01
Fixes #27194 - Run CV auto-publish serially

jturel:
This PR fixes the locking problems that occur when component content view publish result in multiple publishes for the same composite view coming in at/around the same time.

I would have liked to go with a more dynflow-y or foreman-tasks focused solution but what I came up with using Polling module brought in a race condition coming out of ContentView#create_new_version that was not very natural to fix:

```
PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "version_index"
DETAIL:  Key (content_view_id, major, minor)=(19, 1, 0) already exists.
: INSERT INTO "katello_content_view_versions" ("content_view_id", "major", "created_at", "updated_at") VALUES ($1, $2, $3, $4) RETURNING "id"
PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "version_index"
DETAIL:  Key (content_view_id, major, minor)=(19, 1, 0) already exists.
: INSERT INTO "katello_content_view_versions" ("content_view_id", "major", "created_at", "updated_at") VALUES ($1, $2, $3, $4) RETURNING "id"
```

**To test:**

- Create an auto-publishable  composite content view with two components (select 'Latest') each having a single empty repo
- Publish the component repos at the same time
- Observe one of the auto-publishes for the composite will fail with a locking issue

Applying this PR will fix that by serially running the auto-publishes thru the event queue (thanks for the suggestion @jlsherrill !)

For future note I think it would be cool to be able to chain actions serially based on the input of other actions something like :
```
class MyAction < Actions::BaseAction
  chain_on_input do
    :content_view_id
   end
   def plan
     ...
    end
end
```

TODO:

- [x] decide on how to remove events that never seen to complete. purge by age? max retry attempts?

### DIFF
--- a/app/lib/actions/katello/content_view/publish.rb
+++ b/app/lib/actions/katello/content_view/publish.rb
@@ -21,12 +21,7 @@ module Actions
             end
           end
 
-          if options[:minor] && options[:major]
-            version = content_view.create_new_version(options[:major], options[:minor])
-          else
-            version = content_view.create_new_version
-          end
-
+          version = version_for_publish(content_view, options)
           library = content_view.organization.library
           history = ::Katello::ContentViewHistory.create!(:content_view_version => version,
                                                           :user => ::User.current.login,
@@ -60,6 +55,8 @@ module Actions
             plan_action(Katello::Foreman::ContentUpdate, library, content_view)
             plan_action(ContentView::ErrataMail, content_view, library)
             plan_self(history_id: history.id, content_view_id: content_view.id,
+                      auto_publish_composite_ids: auto_publish_composite_ids(content_view),
+                      content_view_version_name: version.name,
                       content_view_version_id: version.id,
                       environment_id: library.id, user_id: ::User.current.id)
           end
@@ -70,34 +67,13 @@ module Actions
         end
 
         def run
-          view = ::Katello::ContentView.find(input[:content_view_id])
-          version = ::Katello::ContentViewVersion.find(input[:content_view_version_id])
-          output[:content_view_id] = view.id
-          output[:content_view_version_id] = version.id
-          unless view.composite?
-            output[:composite_version_auto_published] = []
-            output[:composite_view_publish_failed] = []
-            output[:composite_auto_publish_task_id] = []
-
-            # Iterate through the list of composites
-            # this component belongs to
-            view.component_composites.each do |cv_component|
-              if cv_component.latest? && cv_component.composite_content_view.auto_publish?
-                description = _("Auto Publish - Triggered by '%{component}'") %
-                                { :component => version.name }
-                begin
-                  task = ForemanTasks.async_task(::Actions::Katello::ContentView::Publish,
-                                          cv_component.composite_content_view,
-                                          description,
-                                          :triggered_by => version)
-                  output[:composite_auto_publish_task_id] << task.id
-                  output[:composite_version_auto_published] << task.input[:content_view_version_id]
-                rescue StandardError
-                  ::Katello::UINotifications::ContentView::AutoPublishFailure.deliver!(
-                                              cv_component.composite_content_view)
-                  output[:composite_view_publish_failed] << cv_component.composite_content_view.id
-                end
-              end
+          metadata = {
+            description: _("Auto Publish - Triggered by '%s'") % input[:content_view_version_name],
+            triggered_by: input[:content_view_version_id]
+          }
+          input[:auto_publish_composite_ids].each do |composite_id|
+            ::Katello::EventQueue.push_event(::Katello::Events::AutoPublishCompositeView::EVENT_TYPE, composite_id) do |attrs|
+              attrs[:metadata] = metadata
             end
           end
         end
@@ -136,6 +112,18 @@ module Actions
           end
           content_view.repos(content_view.organization.library).find_all do |repo|
             !library_instances.include?(repo.library_instance_id)
+          end
+        end
+
+        def auto_publish_composite_ids(content_view)
+          content_view.auto_publish_components.pluck(:composite_content_view_id)
+        end
+
+        def version_for_publish(content_view, options)
+          if options[:minor] && options[:major]
+            content_view.create_new_version(options[:major], options[:minor])
+          else
+            content_view.create_new_version
           end
         end
       end
--- a/app/lib/actions/katello/event_queue/poller_thread.rb
+++ b/app/lib/actions/katello/event_queue/poller_thread.rb
@@ -25,15 +25,24 @@ module Actions
 
         def run_event(event)
           @logger.debug("event_queue_event: #{event.event_type}, #{event.object_id}")
-          ::User.as_anonymous_admin do
-            ::Katello::EventQueue.event_class(event.event_type).new(event.object_id).run
+
+          event_instance = nil
+          begin
+            ::User.as_anonymous_admin do
+              event_instance = ::Katello::EventQueue.create_instance(event)
+              event_instance.run
+            end
+          rescue => e
+            @logger.error("event_queue_error: #{event.event_type}, #{event.object_id}")
+            @logger.error(e.message)
+            @logger.error(e.backtrace.join("\n"))
+          ensure
+            if event_instance.try(:retry)
+              ::Katello::EventQueue.reschedule_event(event)
+              @logger.warn("event_queue_rescheduled: type=#{event.event_type} object_id=#{event.object_id}")
+            end
+            ::Katello::EventQueue.clear_events(event.event_type, event.object_id, event.created_at)
           end
-        rescue => e
-          @logger.error("event_queue_error: #{event.event_type}, #{event.object_id}")
-          @logger.error(e.message)
-          @logger.error(e.backtrace.join("\n"))
-        ensure
-          ::Katello::EventQueue.clear_events(event.event_type, event.object_id, event.created_at)
         end
 
         def poll_for_events(suspended_action)
--- a/app/lib/actions/katello/event_queue/poller_thread.rb
+++ b/app/lib/actions/katello/event_queue/poller_thread.rb
@@ -24,7 +24,7 @@ module Actions
         end
 
         def run_event(event)
-          @logger.debug("event_queue_event: #{event.event_type}, #{event.object_id}")
+          @logger.debug("event_queue_event: type=#{event.event_type}, object_id=#{event.object_id}")
 
           event_instance = nil
           begin
@@ -33,13 +33,17 @@ module Actions
               event_instance.run
             end
           rescue => e
-            @logger.error("event_queue_error: #{event.event_type}, #{event.object_id}")
+            @logger.error("event_queue_error: type=#{event.event_type}, object_id=#{event.object_id}")
             @logger.error(e.message)
             @logger.error(e.backtrace.join("\n"))
           ensure
             if event_instance.try(:retry)
-              ::Katello::EventQueue.reschedule_event(event)
-              @logger.warn("event_queue_rescheduled: type=#{event.event_type} object_id=#{event.object_id}")
+              result = ::Katello::EventQueue.reschedule_event(event)
+              if result == :expired
+                @logger.warn("event_queue_event_expired: type=#{event.event_type} object_id=#{event.object_id}")
+              elsif !result.nil?
+                @logger.warn("event_queue_rescheduled: type=#{event.event_type} object_id=#{event.object_id}")
+              end
             end
             ::Katello::EventQueue.clear_events(event.event_type, event.object_id, event.created_at)
           end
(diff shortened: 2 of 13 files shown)


jturel:
[test katello]

app/lib/actions/katello/content_view/publish.rb
```diff
@@ -70,35 +71,9 @@ def humanized_name
         end
 
         def run
-          view = ::Katello::ContentView.find(input[:content_view_id])
-          version = ::Katello::ContentViewVersion.find(input[:content_view_version_id])
-          output[:content_view_id] = view.id
```
jturel: what would the reason be to take values from the input and put them into the output? seems unnecessary

app/models/katello/events/publish_component_composites.rb
```diff
@@ -0,0 +1,29 @@
...
+        @version = ::Katello::ContentViewVersion.find_by_id(cvv_id)
+      end
+
+      def run
+        return unless version
+
+        description = _("Auto Publish - Triggered by '%s'") % version.name
+        version.content_view.auto_publish_components.each do |component|
+          begin
+            ForemanTasks.sync_task(::Actions::Katello::ContentView::Publish,
```
jlsherrill: should this be a sync_task?  that'd cause the event queue to be held up until the publish is complete

app/lib/actions/katello/content_view/publish.rb
```diff
@@ -70,35 +71,9 @@ def humanized_name
...
-                rescue StandardError
-                  ::Katello::UINotifications::ContentView::AutoPublishFailure.deliver!(
-                                              cv_component.composite_content_view)
-                  output[:composite_view_publish_failed] << cv_component.composite_content_view.id
-                end
-              end
-            end
+          if input[:auto_publish]
+            event = ::Katello::EventQueue.push_event(::Katello::Events::PublishComponentComposites::EVENT_TYPE, input[:content_view_version_id])
+            output[:composite_auto_publish_event_id] = event.id
```
jturel: We lost some insight into the tasks that get scheduled for auto publish. Not sure if putting the event_id here is useful at all since it'll probably be gone if it ever needs to be looked at. Though since auto-publish should be more reliable now the output here should be less valuable in general.

app/lib/actions/katello/content_view/publish.rb
jlsherrill: agreed, i think it is unnecessary

app/lib/actions/katello/content_view/publish.rb
jlsherrill: yeah, its probably not useful at all, i'd not bother

app/models/katello/events/publish_component_composites.rb
jlsherrill: oh i see, this is the reason its more reliable :)
I kinda envisioned it a tad differently:

1)  the composite id is put on the queue so that way its a single task that need to get started
2) The tasks are started asynchronously 
3) if one fails to publish, it gets put back on the queue for another ~5 minutes or so (the queue class will need new functionality to support this)

jturel: Ah yeah, you did mention that when we talked. The only issue I see is that I don't think we can correlate back to the version which triggered the action (unless we put more data on the event?) so we can't build the `description` text

jlsherrill: oh hrmmm, that does complicate it a bit.......  we could add that as a column on the event model, or maybe a serialized column for arbitrary data?

jturel: Maybe it would be possible to async create the task in ContentView::Publish in a suspended state and the event could try to run the task. Dunno if that's possible out of the box

jturel:
@jlsherrill updated per your suggestion *but* I'm seeing some very odd behavior when rescheduling (marking the task not in progress and setting the process_after timestamp) - it appears as though something, somewhere (separate thread?) is immediately marking the task as in progress. I left some logging and an event which will throw an error when ran in place. you can check it like so:

```
Katello::EventQueue.push_event('auto_publish_composite', 1)

Katello::Event.all

#<Katello::Event id: 1065, object_id: 1, event_type: "auto_publish_composite", in_progress: true, created_at: "2019-07-03 19:15:22", updated_at: "2019-07-03 19:19:29", metadata: {}, process_after: "2019-07-03 19:19:59">
```

I'm expecting in_progress to be false here (test case shows reschedule code works) - am I missing something?

app/lib/actions/katello/event_queue/poller_thread.rb
```diff
@@ -26,12 +26,15 @@ def close
...
-            ::Katello::EventQueue.event_class(event.event_type).new(event.object_id).run
+            ::Katello::EventQueue.create_instance(event).run
           end
         rescue => e
           @logger.error("event_queue_error: #{event.event_type}, #{event.object_id}")
           @logger.error(e.message)
           @logger.error(e.backtrace.join("\n"))
+
+          rescheduled = ::Katello::EventQueue.reschedule_event(event)
+          @logger.warn("event_queue_rescheduled: type=#{event.event_type} object_id=#{event.object_id}") if rescheduled
```
jlsherrill: I think we should likely catch a specific error such that:

* we don't reschedule an event for some unexpected error that is likely to not be recoverable (forever and ever)
* we dont' reschedule an event that isn't meant to be rescheduled

My thinking is that below, we explicitly check for a 'lock' exception, and then throw a specific 'reschedule' exception that is caught here. 

Thoughts?

jturel: The way the code is written right now won't reschedule every failed event run - only those where the concrete class implements the `retry_seconds` seconds method.

Are you saying that we should tightly control the retry behavior for particular modes of failure that we determine rather than just retrying if `retry_seconds` is implemented (which would retry on any failure)?

jlsherrill: yeah, i think we should only retry on a specific failure (a lock exception) if possible

jturel:
@jlsherrill have a look! I didn't use any special error to rescue from but rather a 'retry' flag on the event instance. I like that better than using exceptions to handle the flow since we can check 'retry' in the ensure block

db/migrate/20190702182118_add_metadata_process_after_to_katello_event.rb
```diff
@@ -0,0 +1,6 @@
+class AddMetadataProcessAfterToKatelloEvent < ActiveRecord::Migration[5.2]
+  def change
+    add_column :katello_events, :metadata, :string
```
parthaa: prefer `text` to `string` here. That seems to be the default for all serializable columns.

app/services/katello/event_queue.rb
```diff
@@ -1,15 +1,30 @@
 module Katello
   class EventQueue
+    MAX_AGE = 6.hours
```
parthaa: might make sense to mark "metadata" as serializable. I would prefer key/value style hashes

jturel: check event.rb - unless you mean something else?

db/migrate/20190702182118_add_metadata_process_after_to_katello_event.rb
jturel: done

app/services/katello/event_queue.rb
parthaa: my bad :)

jturel:
[test katello]

parthaa:
This was the script i ran to test this PR
```
#!/bin/bash
shopt -s expand_aliases
alias hammer='BUNDLE_GEMFILE=~/git/hammer-cli-katello/Gemfile bundle exec hammer'

set -e
set -x

ORG="org-$RANDOM"
PRODUCT="test_product"
hammer organization create --name=$ORG
PRODUCT_OPT="--organization=$ORG --product=$PRODUCT"
hammer product create --organization=$ORG --name=$PRODUCT
hammer repository create $PRODUCT_OPT --name=yum --content-type=yum --url="https://partha.fedorapeople.org/test-repos/pteradactly-with-dino-errata/"

hammer content-view create  --organization=$ORG --name=joy1
hammer content-view add-repository --organization=$ORG --repository=yum --product=$PRODUCT --name=joy1

hammer content-view create  --organization=$ORG --name=joy2
hammer content-view add-repository --organization=$ORG --repository=yum --product=$PRODUCT --name=joy2

hammer content-view create  --organization=$ORG --name=joy3
hammer content-view add-repository --organization=$ORG --repository=yum --product=$PRODUCT --name=joy3

hammer content-view create --organization=$ORG --composite --auto-publish=yes --name=joy-composite1 
hammer content-view component add --organization=$ORG --latest --component-content-view=joy1 --composite-content-view=joy-composite1
hammer content-view component add --organization=$ORG --latest --component-content-view=joy2 --composite-content-view=joy-composite1
hammer content-view component add --organization=$ORG --latest --component-content-view=joy3 --composite-content-view=joy-composite1


hammer content-view create --organization=$ORG --composite --auto-publish=yes --name=joy-composite2
hammer content-view component add --organization=$ORG --latest --component-content-view=joy1 --composite-content-view=joy-composite2
hammer content-view component add --organization=$ORG --latest --component-content-view=joy2 --composite-content-view=joy-composite2
hammer content-view component add --organization=$ORG --latest --component-content-view=joy3 --composite-content-view=joy-composite2


hammer content-view create --organization=$ORG --composite --auto-publish=yes --name=joy-composite3
hammer content-view component add --organization=$ORG --latest --component-content-view=joy1 --composite-content-view=joy-composite3
hammer content-view component add --organization=$ORG --latest --component-content-view=joy2 --composite-content-view=joy-composite3
hammer content-view component add --organization=$ORG --latest --component-content-view=joy3 --composite-content-view=joy-composite3

hammer content-view publish --organization=$ORG --name=joy1 &
hammer content-view publish --organization=$ORG --name=joy2 &
hammer content-view publish --organization=$ORG --name=joy3 &
```
Wait for the concurrent publishes to finish

followed by 

```
$ hammer content-view list --organization=$ORG
```

you should see an output that looks like
```
----------------|---------------------------|-----------|---------------------|---------------
CONTENT VIEW ID | NAME                      | COMPOSITE | LAST PUBLISHED      | REPOSITORY IDS
----------------|---------------------------|-----------|---------------------|---------------
47              | Default Organization View |           | 2019/07/09 21:36:19 |               
48              | joy1                      |           | 2019/07/09 21:41:27 | 149           
49              | joy2                      |           | 2019/07/09 21:41:26 | 149           
50              | joy3                      |           | 2019/07/09 21:41:27 | 149           
51              | joy-composite1            | true      | 2019/07/09 21:41:41 | 159, 158, 160 
52              | joy-composite2            | true      | 2019/07/09 21:41:45 | 159, 158, 160 
53              | joy-composite3            | true      | 2019/07/09 21:41:48 | 159, 158, 160 
----------------|---------------------------|-----------|---------------------|---------------
```

Note all the 3 composites published automatically at the same time. You can retest by just running
```
hammer content-view publish --organization=$ORG --name=joy1 &
hammer content-view publish --organization=$ORG --name=joy2 &
hammer content-view publish --organization=$ORG --name=joy3 &
```
 and checking the cv list again

Works well. APJ

jturel:
[test katello]

jturel:
@jlsherrill any additional feedback for me here?

app/services/katello/event_queue.rb
jlsherrill: To make sure i follow correctly, how we prevent things from being rescheduled past the MAX_AGE, it looks like if an event is to be retried:

1. we update the event's process_after in the future by 'retry_seconds'
2. we then delete any event who was created at more than 6 hours ago

Assuming this is correct, i think this is fine.  I think we should probably log something if MAX_AGE has expired though for traceability.

jlsherrill:
One small comment, otherwise it looks fine form a code perspective

app/services/katello/event_queue.rb
jturel: Yes, but there's a problem with that now that I thought through it some more: I envisioned a scenario where the task queue gets backed up, and for some reason the Sat server is stopped for some time > MAX_AGE. When it comes back up - *poof* all the events get deleted because they are too old after first event gets run. All because I tied MAX_AGE to `clear_events`.

I'm pushing an update to make the `reschedule_event` method itself aware of MAX_AGE, and if it doesn't get rescheduled it'll get cleared out as usual in PollerThread. Will add the logging too

jturel: Should be good now :+1:

jturel:
@parthaa some small changed since you last tested. would you give it one more run-through before I merge?

parthaa:
Ack

jturel:
Thanks guys!

PR #8189, opened 2019-07-01
Fixes #27169 - Pass arg to delete_orphaned_content

sjha4:


jlsherrill:
Looks good!  Would you mind adding a quick test?

sjha4:
Added test for the task..

PR #8190, opened 2019-07-01
Fixes #27204 - use newer pulp clients

jlsherrill:
This change migrates to the newer pulp3 client bindings
while also restructuring our application to use the more
consistent nature of the bindings.  This also makes it easier
to add in new content types, reducing the amount of repeated
code and the amount of code needed for a new type


jlsherrill:
I'm sure there will be more test failures, but this should largely be ready for review.  

Ideally i wouldn't have had to rerecord any casettes (as the binding method changes shoudln't have changed any api), however a recent commit (https://github.com/pulp/pulpcore/commit/9e6e5d36e594ffff85aac0dc28aa77e06874b971) removed the 'validate' option so any bindings after that change will refuse to accept it.

app/lib/pulp/v3/api.rb
```diff
@@ -1,108 +0,0 @@
...
-      def ansible_remotes_api
-        @ansible_remotes_api ||= PulpAnsibleClient::RemotesApi.new
-      end
-
-      def ansible_content_api
-        @ansible_content_api ||= PulpAnsibleClient::ContentApi.new
-      end
-    end
-  end
-end
```
sjha4: Yayy! :tada:

app/services/katello/pulp3/repository/ansible_collection.rb
```diff
@@ -1,72 +1,43 @@
...
 module Katello
   module Pulp3
     class Repository
       class AnsibleCollection < ::Katello::Pulp3::Repository
-        def create_content_remote
-          remote_collection_data = PulpAnsibleClient::CollectionRemote.new(remote_options)
-          response = pulp3_api.remotes_ansible_collection_create(remote_collection_data)
-          response._href
+        def api_client
+          PulpAnsibleClient::ApiClient.new(smart_proxy.pulp3_configuration(PulpAnsibleClient::Configuration))
```
sjha4: This should probably mimic the file service method: https://github.com/Katello/katello/pull/8190/files#diff-300f565091a88861433c44ce4f5005e7R11

jlsherrill: strange that this didn't trigger any test failures

jlsherrill: fixed!

jlsherrill:
ian realized that PulpFileClient::ApiError was being referenced in repository.rb, so i changed it to allow the service class to declare the type specific api error class

app/services/katello/pulp3/repository/ansible_collection.rb
sjha4: Yes..I saw this on indexing..CRUD worked fine..We don't have collection tests for sync/indexing as of now..

jlsherrill:
[test katello]

jlsherrill:
[test katello]

katello.gemspec
```diff
@@ -42,9 +42,9 @@ Gem::Specification.new do |gem|
   # Pulp
   gem.add_dependency "runcible", ">= 2.11.0", "< 3.0.0"
   gem.add_dependency "anemone"
-  gem.add_dependency "pulpcore_client", "<= 3.0.0rc3.dev0.1561038430"
-  gem.add_dependency "pulp_file_client", "<= 0.1.0b1.dev.1560792388"
-  gem.add_dependency "pulp_ansible_client", "<= 0.2.0b1.dev0.1561386170"
+  gem.add_dependency "pulpcore_client"
+  gem.add_dependency "pulp_file_client"
```
jlsherrill: pin this to less than 3.0

parthaa:
APJ

PR #8191, opened 2019-07-01
Fixes #26767 - Custom products on bulk content host manage page

sjha4:
**Steps to Reproduce:**
1. Create a custom product
2. Create a repository in a custom product
3. Register content host(s) to your box
4. Goto Hosts -> Content Hosts
5. Select one or more hosts with checkbox
6. Select Action (Dropdown) -> Manage Repository Sets

Actual results:
Repository list is empty (in case you don't have any redhat products) with the message 'No enabled Repository Sets provided through subscriptions.'
This list only shows redhat products.

Expected results:
Ability to change Repository Sets for hosts selected, like in Host -> Content Hosts ->host -> Repository Sets

### DIFF
--- a/app/controllers/katello/api/v2/repository_sets_controller.rb
+++ b/app/controllers/katello/api/v2/repository_sets_controller.rb
@@ -19,6 +19,7 @@ module Katello
     param :enabled, :bool, :required => false, :desc => N_("If true, only return repository sets that have been enabled. Defaults to false")
     param :with_active_subscription, :bool, :required => false, :desc => N_("If true, only return repository sets that are associated with an active subscriptions")
     param :organization_id, :number, :desc => N_("organization identifier"), :required => false
+    param :with_custom, :bool, :required => false, :desc => N_("If true, return custom repository sets along with redhat repos")
     param_group :search, Api::V2::ApiController
     add_scoped_search_description_for(Katello::ProductContent)
     def index
@@ -116,7 +117,8 @@ module Katello
       end
 
       relation = relation.where(Katello::Content.table_name => {:name => params[:name]}) if params[:name].present?
-      relation.redhat
+      relation = relation.redhat unless ::Foreman::Cast.to_bool(params[:with_custom])
+      relation
     end
 
     def find_product_content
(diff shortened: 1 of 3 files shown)


sjha4:
[test katello]

sjha4:
[test katello]

jturel:
@sjha4 this works well, but I noticed an existing problem that is arguably made more prevalent with this change.

Manage Repository Sets presents a list that is not a superset of all the selected hosts available repo sets, but rather all repo sets for the org. This is inconsistent with the other options in the dropdown because their lists sum up the applicable items for the selected hosts. And with this change we've made that list even longer.

In contrast if I select 3 hosts then the errata shown under Manage Errata are applicable to at least 1 of the selected hosts. This might be a good opportunity to make Repository Sets consistent with the other dropdown options since we are making changes in this area via the bulk_hosts_controller.

Example API:

```
    api :POST, "/hosts/bulk/repository_sets", N_("Fetch available repository sets for hosts.")                                                                               
    param_group :bulk_params                                                                                                                                                 
    def repository_sets                                                                                                                                                      
      content_finder = ProductContentFinder.new(                                                                                                                             
          :consumables => @hosts,                                                                                                                                            
          :match_subscription => true,                                                                                                                                       
          :match_environment => false                                                                                                                                        
      )                                                                                                                                                                      
                                                                                                                                                                             
      respond_for_index(...)                                                                                                                                                 
    end
```

I could be missing something. Maybe @parthaa can comment to weigh in on the suggestion just in case it's not a good one

sjha4:
@jturel: Thanks for the review. I have pushed a second commit trying to implement what you suggested. The only problem which still exists is that if we have a superset of repository sets provided by subscriptions, we would still end up disabling/enabling repo sets for hosts which are not subscribed to the product.
I had this discussion with @jlsherrill around showing all repos on the bulk manage page. AFAIR, he said we can enable a repo for a host without the associated subscription and then add the subscription later to be able to use the repo. In which case restricting the list to a superset of repos would be more confusing than showing all?

jturel:
@sjha4 i actually didn't realize we could do the overrides while a host isn't subscribed; i guess that's a useful feature! looks like we shouldn't go with my suggestion after all. sorry that you wrote most of the code for it already :disappointed:

ianballou:
I'm not sure what the status of this PR is, but I gave it a try.  I found that the Red Hat repositories I have also disappear from the Repository Sets Management window.

sjha4:
Can you see only custom products in the Content Host> Select a few content hosts> Actions> Manage Repository Sets page? Do you have subscriptions for the redhat product attached to the content host?

ianballou:
Before applying your PR I saw only Red Hat related content on the Manage
Repo Sets page. After applying your PR, the Manage Repo Sets page was
blank.  I don't have anything attached to the content host.

PR #8192, opened 2019-07-02
Fixes #27199 - adds http proxy and policy fields

jjeffers:
This PR adds http proxy policy and http proxy fields so that a user can associate a proxy policy (none, use the global, or use a specific HTTP proxy). Be default a newly created repository will use the global default HTTP proxy.

This also introduces additional form elements to the UI on repository creation and on the info-details form in place editing of the proxy values.


jjeffers:
[test katello]

jlsherrill:
so far this looks good, i don't have any major comments

jjeffers:
[test katello]

app/services/katello/pulp/repository.rb
```diff
@@ -90,8 +90,12 @@ def sync(overrides = {})
       end
 
       def create
-        smart_proxy.pulp_api.extensions.repository.create_with_importer_and_distributors(repo.pulp_id, generate_importer,
-                                                                              generate_distributors, display_name: root.name)
+        importer = generate_importer
+        smart_proxy.pulp_api.extensions.repository.create_with_importer_and_distributors(
+          repo.pulp_id,
+          importer,
```
jlsherrill: any reason to not just call generate_importer here?  (wondering if this was from leftover debugging)

app/services/katello/pulp/repository.rb
```diff
@@ -247,6 +251,32 @@ def index_content_units
           content_type.pulp2_service_class
         end
       end
+
+      def proxy_options
+        if repo.root.http_proxy_policy == RootRepository::NO_DEFAULT_HTTP_PROXY
+          return {}
```
jlsherrill: i think its actually more than returning nothing, i think we should return empty string for proxy 'host'.  Also the same below

jlsherrill: and the reason is that it makes this change independent of the installer dropping in a 'global' proxy config for pulp2, that can continue to happen, but if we use empty string, it'll be ignored.  (eventually we'll remove it though)

db/migrate/20190701174401_add_http_proxy_policy_to_repository_root.rb
```diff
@@ -0,0 +1,6 @@
+class AddHttpProxyPolicyToRepositoryRoot < ActiveRecord::Migration[5.2]
+  def change
+    add_column :katello_root_repositories, :http_proxy_policy, :string,
+      default: Katello::RootRepository::NO_DEFAULT_HTTP_PROXY
```
jlsherrill: this looks good, we probably want to import the 'skip global proxy' column that is currently there and drop it

app/services/katello/pulp/repository.rb
jjeffers: yes, I was spelunking!

db/migrate/20190701174401_add_http_proxy_policy_to_repository_root.rb
jjeffers: I was wondering if we need to add a migration that sets the proxy policy based on what the "ignore global proxy" column value is? I wasn't sure how commonly this setting is used or how important it is to carry that information forward.

jlsherrill: yep, i think we do.  Its common enough to carry its intent forward.

jjeffers: Should the default be, instead of 'No Default",set "use global default", and then set the records that already have the "ignore" column as true set to to "no default"?

jlsherrill: yep, exactly!  The current default is 'use global default' (which can mean no proxy), so all repos on the new model should be set to that unless ignore_global_proxy is set to true

jjeffers: That's what I thought. As it is said, so let it be written.

jjeffers:
[test katello]

jjeffers:
[test katello]

app/services/katello/pulp/repository.rb
```diff
@@ -247,6 +250,29 @@ def index_content_units
...
+          uri = URI(proxy.url)
+          proxy_options = {
+            proxy_host: uri.host,
+            proxy_port: uri.port,
+            proxy_username: proxy.username,
+            proxy_password: proxy.password
+          }
+          return proxy_options
+        end
+        {}
```
jlsherrill: I think proxy_host should be empty string here, as if there is no global proxy defined, it will pass no option, and thus use the old pulp2 global proxy (which we don't want).

test/services/katello/pulp/rpm_test.rb
```diff
@@ -38,6 +38,31 @@ def test_update_model_is_idempotent
...
+      def test_with_global_http_proxy
+        @repo.root.update(http_proxy_policy: RootRepository::GLOBAL_DEFAULT_HTTP_PROXY)
+        RepositorySupport.create_and_sync_repo(@repo)
+      end
+
+      def teardown
+        RepositorySupport.destroy_repo(@repo)
+        User.current = nil
+      end
+    end
```
jlsherrill: this is more for testing the rpm unit class, i think this test would go better here: https://github.com/Katello/katello/blob/master/test/services/katello/pulp/repository/yum_test.rb

jjeffers:
[test katello]

jjeffers:
[test katello]

jjeffers:
[test katello]

app/models/katello/concerns/http_proxy_extensions.rb
```diff
@@ -4,7 +4,10 @@ module HttpProxyExtensions
       extend ActiveSupport::Concern
 
       included do
+        has_many :root_repositories, :class_name => "::Katello::RootRepository", :foreign_key => :http_proxy_id,
+          :inverse_of => :http_proxy, :dependent => :nullify
         after_save :update_default_proxy_setting
+        after_save :update_repository_proxy_details
```
jlsherrill: this should probably be after_update ,and actually update_default_proxy_setting, likely should be too

app/models/katello/concerns/http_proxy_extensions.rb
```diff
@@ -17,16 +20,29 @@ def update_default_proxy_setting
...
+
+          repos =
+            Repository.joins(:root).where("#{RootRepository.table_name}.http_proxy_id = ?", id) +
+            Repository.joins(:root).where("#{RootRepository.table_name}.http_proxy_policy = ?", RootRepository::GLOBAL_DEFAULT_HTTP_PROXY)
+          repos.uniq.each do |repo|
+            ForemanTasks.async_task(
+              ::Actions::Katello::Repository::UpdateHttpProxyDetails,
+              repo)
+          end
+        end
```
jlsherrill: this isn't taking into account that the http proxy that was updated might not be the global proxy, we'd only want to update repos set to 'GLOBAL_DEFAULT_HTTP_PROXY' if the one being updated is the global proxy.

jjeffers:
[test katello]

app/models/katello/concerns/http_proxy_extensions.rb
```diff
@@ -17,16 +20,33 @@ def update_default_proxy_setting
...
+        changes = self.previous_changes
+        if changes.key?(:url) || changes.key?(:username) || changes.key?(:password)
+
+          root_repos = RootRepository.with_selected_proxy(id)
+
+          if self == HttpProxy.default_global_content_proxy
+            root_repos += RootRepository.with_global_proxy
+          end
+
+          repos = root_repos.collect(&:repositories)
```
jlsherrill: you probably could just collect  (&:library_instance), we don't to update the cv repos

app/models/katello/concerns/http_proxy_extensions.rb
```diff
@@ -17,16 +20,33 @@ def update_default_proxy_setting
...
+          repos.flatten.uniq.each do |repo|
+            ForemanTasks.async_task(
+              ::Actions::Katello::Repository::UpdateHttpProxyDetails,
+              repo)
```
jlsherrill: Can we make this into a bulk action?

jjeffers: We sure can. My understanding is that the failure handling  is different for BulkActions, does that matter here?

app/services/katello/pulp3/repository.rb
```diff
@@ -249,6 +252,22 @@ def ssl_remote_options
...
+
+        proxy = repo.root.http_proxy
+        if repo.root.http_proxy_policy == RootRepository::GLOBAL_DEFAULT_HTTP_PROXY
+          proxy = HttpProxy.default_global_content_proxy
+        end
+
+        if proxy
+          return proxy.full_url
+        end
+        nil
```
jlsherrill: couldn't you just call repo.root.http_proxy&.full_url for this entire method?

jjeffers: Did you mean for just the if clause or for the entire `remote_proxy_url` method?

jlsherrill: the whole method (which probably means we don't actually need it split out as a method

app/models/katello/concerns/http_proxy_extensions.rb
jlsherrill: since its an async task, the failure would seen as about the same

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/http-proxy-service.js
```diff
@@ -0,0 +1,53 @@
...
+
+        this.displayHttpProxyName = function (proxies, proxyId) {
+            var findProxy = function(proxy) {
+                return proxy.id === this.id;
+            };
+
+            var foundProxy = proxies.find(findProxy, { id: proxyId });
+
+            if (proxies.length === 0) {
+                return "No HTTP Proxies found";
```
jlsherrill: this needs a translate()

lib/katello/tasks/update_content_default_http_proxy.rake
```diff
@@ -3,7 +3,7 @@ require File.expand_path("../engine", File.dirname(__FILE__))
 namespace :katello do
   desc "Sets the content default http proxy to an existing http proxy based on supplied URL."
   task :update_default_http_proxy => :environment do
-    setting = ::Setting::Content.where(name: 'content_default_http_proxy').first
+    setting = ::Setting.find_by_name('content_default_http_proxy')
```
jlsherrill: nitpick  i think we prefer .find_by(name: 'content_default_http_proxy')

jlsherrill:
few small comments, i'll start doing testing

jlsherrill:
* When creating a repository, i have to select an option, it should default to "Global default" IMO to reduce the amount of selections
* The styling around the selects on the new repository page do not match existing selects
* Is there any way to get the edit proxy options to behave more like the upstream auth where its a single save/cancel?  

image

jlsherrill:
* Pulp 2 updates don't seem to be propagated, i think because https://github.com/Katello/katello/blob/master/app/models/katello/root_repository.rb#L270  wasn't updated for those two new options

app/models/katello/concerns/http_proxy_extensions.rb
```diff
@@ -17,16 +20,32 @@ def update_default_proxy_setting
...
+        if changes.key?(:url) || changes.key?(:username) || changes.key?(:password)
+
+          root_repos = RootRepository.with_selected_proxy(id)
+
+          if self == HttpProxy.default_global_content_proxy
+            root_repos += RootRepository.with_global_proxy
+          end
+
+          repos = root_repos.collect(&:library_instance)
+          ForemenTasks.async_task(
```
jlsherrill: typo here Foreman singular :)

app/models/katello/concerns/http_proxy_extensions.rb
```diff
@@ -17,16 +20,32 @@ def update_default_proxy_setting
...
+            ::Actions::BulkAction,
+            Actions::Katello::Repository::RefreshRepository,
+            repos.flatten.uniq.order_by_root(:name))
```
jlsherrill: also getting undefined method `order_by_root' for #<Array:0x00007fcaecbeb1d0>

jjeffers:
>     * When creating a repository, i have to select an option, it should default to "Global default" IMO to reduce the amount of selections
> 
>     * The styling around the selects on the new repository page do not match existing selects
> 
>     * Same styling difference on the edit page too
> 
>     * Is there any way to get the edit proxy options to behave more like the upstream auth where its a single save/cancel?
> 
> 
> image

I noticed that the OStree repo type form elements had selects that were not also styled properly. Is that a bug we should file?

jlsherrill:
@jjeffers yep, i think an issue for that would be fine (or if its trivial, we can just do it here, either way.

jjeffers:
[test katello]

jlsherrill:
Small UI issue:

* on the edit repo page, the value is always blank:

image

jlsherrill:
Noticed a gap in functionality.  If you set a repository to use the 'global' policy and then change the global http proxy setting, the repository does not get updated

jjeffers:
[test katello]

engines/bastion_katello/app/assets/javascripts/bastion_katello/http-proxies/http-proxies.module.js
```diff
@@ -0,0 +1,14 @@
+/**
+ *  * @ngdoc module
+ *   * @name  Bastion.http-proxies
+ *    *
+ *     * @description
+ *      *    Module for http proxies
+ *       */
```
jlsherrill: double comment?

engines/bastion_katello/app/assets/javascripts/bastion_katello/http-proxies/http-proxy.factory.js
```diff
@@ -0,0 +1,21 @@
+/**
+ *  * @ngdoc service
+ *   * @name  Katello.http-proxies.factory:HttpProxy
+ *    *
+ *     * @requires BastionResource
+ *      *
+ *       * @description
+ *        *   Provides a BastionResource for Http Proxies.
+ *         */
```
jlsherrill: same thing, double comment?

jjeffers:
[test katello]

jjeffers:
[test katello]

jlsherrill:
:green_heart:  :green_apple: :green_book: :green_salad:

jjeffers:
> green_heart green_apple green_book green_salad

So this is good for a merge?!?! :)

jlsherrill:
merge merge merge

PR #8195, opened 2019-07-03
Fixes #27223 - manage global proxy via seeds

jlsherrill:
during the cross over period where the content proxy
is set via the installer, but managed via the global
content Setting, we need something to populate the
data from the installer, so this seed can do that for us

### DIFF
--- a/db/seeds.d/115-http_proxy.rb
+++ b/db/seeds.d/115-http_proxy.rb
@@ -1,0 +1,25 @@
+if (SETTINGS[:katello][:cdn_proxy] && SETTINGS[:katello][:cdn_proxy][:host])
+  config = SETTINGS[:katello][:cdn_proxy]
+  uri = URI(config[:host])
+  uri.user = nil
+  uri.password = nil
+  uri.port = config[:port] if config[:port]
+
+  name = uri.host
+  if (http_proxy = ::HttpProxy.where(name: name).first)
+    http_proxy.update_attributes!(url: uri.to_s,
+                                  username: config[:user],
+                                  password: config[:password])
+  else
+    ::HttpProxy.create!(name: name,
+                        url: uri.to_s,
+                        username: config[:user],
+                        password: config[:password])
+  end
+
+  if Setting[:content_default_http_proxy] != name
+    Setting[:content_default_http_proxy] = name
+  end
+else
+  Setting[:content_default_http_proxy] = ''
+end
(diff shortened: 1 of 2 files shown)


jlsherrill:
added a test and refactor our seeds test to be faster, went from ~45 seconds down to ~4 seconds.  This no longer does a full db seed run, but that is done as part of jenkins anyways: https://ci.theforeman.org/blue/organizations/jenkins/katello-pr-test/detail/katello-pr-test/945/pipeline/146

db/seeds.d/115-http_proxy.rb
```diff
@@ -0,0 +1,25 @@
+if (SETTINGS[:katello][:cdn_proxy] && SETTINGS[:katello][:cdn_proxy][:host])
+  config = SETTINGS[:katello][:cdn_proxy]
+  uri = URI(config[:host])
```
jjeffers: When I am specifying the cdn setting in the katello config, the host is not picked up unless I format the string as a URL (including the HTTP scheme). For example, "http://centos7-katello-devel.example.com/".

Is this the real meaning of the :cdn_proxy, :host entry? I was expecting either an IP address or a hostname to work here.

test/lib/tasks/seeds_test.rb
```diff
@@ -1,72 +1,108 @@
...
+        run_proxy_seed
+
+        assert_empty Setting[:content_default_http_proxy]
+      end
+
+      test "Creates proxy and assigns if not existing" do
+        refute ::HttpProxy.find_by(:name => 'foo.com')
+
+        SETTINGS[:katello][:cdn_proxy] = {
+          host: 'http://foo.com/',
```
jjeffers: This might answer my question from above, but let me know if this is the expected format for host in theforeman/katello config.

db/seeds.d/115-http_proxy.rb
jlsherrill: the install does require a full uri "http://foo.com", and not just a hostname, this is to indicate whether http or https is being used.  
example:
```
# foreman-installer  --katello-proxy-url=redhat.com
Resetting puppet server version param...
Parameter katello-proxy-url invalid: redhat.com must match one of (?i-mx:^https?:\/\/)
```

jjeffers: Ok, so it looks like this is the established convention. Thanks!

PR #8199, opened 2019-07-07
Fixes #27242 - Update CDN url might break feed url.

hao-yu:
When updating the CDN url, some repositories' url are updating
multiple times in a single foreman task if they are used in any
content view. This causes the url path to replace by an empty
string due to a bug in the code that can't handle url update
multiple time safely. This commit fixed the issue.

### DIFF
--- a/app/lib/actions/katello/provider/update.rb
+++ b/app/lib/actions/katello/provider/update.rb
@@ -15,22 +15,21 @@ module Actions
         private
 
         def update_url(provider, base_url)
-          current_base_url = provider.repository_url
           provider.update_attributes!(:repository_url => base_url)
 
           if provider.redhat_provider?
             provider.products.enabled.each do |product|
-              update_repository_urls(product, current_base_url, base_url)
+              update_repository_urls(product)
             end
           end
         end
 
-        def update_repository_urls(product, current_base_url, new_base_url)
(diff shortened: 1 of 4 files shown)


jlsherrill:
[test katello]

hao-yu:
Fixed robocop warning

jlsherrill:
[test katello]

parthaa:
Code looks ok. Can you give me steps to test this PR ?

hao-yu:
Hi

Before patching this, you can follow the steps below to reproduce the issue first:
1. Enable any Red Hat Repository and sync it with CDN.
2. Create a CV and add that repository to the CV, publish and promote the CV. 
3. Update the CDN url to something like "http://satellite.example.com/pub/exports" and wait for the task to finish.
4. Check the Dynflow task console, you should see duplicate "Actions::Katello::Repository::Update" and "Actions::Pulp::Repository::Refresh" actions for the same repo.
5. Check the feed url of the repository in the web ui and you should see that the it becomes "http://satellite.example.com/pub/exports".

Patch this commit then Restart httpd and foreman tasks.
1. Update the CDN url again and wait for the task to finish.
2. Check the Dynflow task console again, you shouldn't see any duplicate actions.
3. The feed url of the repo should be fixed.

PR #8200, opened 2019-07-08
Fixes #27248 - intorduce separate queue for host-related tasks

iNecas:
To prevent the user-related actions to be highly influenced by the
hosts-related tasks (such as errata applicability calculations), the
work for the hosts should be put on dedicated queue with independent set
of worker threads to process, so if the host-workers get starved, the
non-host work could still flow.

This PR is introducing new queue `hosts_queue` to be assigned to
host-related tasks and setting `host_tasks_workers_pool_size` to set how
many workers should be in the queue.

To achieve the old behaviour of one shared queue, one can set
`host_tasks_workers_pool_size` to 0.

### DIFF
--- a/app/lib/actions/katello/host/attach_subscriptions.rb
+++ b/app/lib/actions/katello/host/attach_subscriptions.rb
@@ -2,6 +2,10 @@ module Actions
   module Katello
     module Host
       class AttachSubscriptions < Actions::EntryAction
+        def queue
+          ::Katello::HOST_TASKS_QUEUE
+        end
+
         def plan(host, pools_with_quantities_params)
           action_subject(host)
           sequence do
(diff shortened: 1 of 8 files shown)


mccun934:
@iNecas is there associated PRs against Dynflow or Foreman Tasks that are required?

iNecas:
Nope, all the required changes on Dynflow/Tasks side are in since Sat 6.4 times. This is all the changes that would be needed to make use of this.

jturel:
I think we might want to include the following actions to use the hosts queue:

app/lib/actions/katello/host/hypervisors.rb
app/lib/actions/katello/host/auto_attach_subscriptions.rb

Let me know if you omitted them for a particular reason

jturel:
Other than my previous point it works well - I see the new queue being used and when I set the pool size for it to 0 the host actions revert to the default queue :+1:

adamruzicka:
@jturel I added the two actions you mentioned to the host queue as well in here https://github.com/Katello/katello/pull/8276 . I'm afraid @iNecas won't have time to continue with this

PR #8204, opened 2019-07-09
Fixes #27257 - properly check os name for debian

jlsherrill:


engines/bastion_katello/app/assets/javascripts/bastion_katello/hosts/host.factory.js
```diff
@@ -27,7 +27,7 @@ angular.module('Bastion.hosts').factory('Host',
         };
 
         resource.prototype.isDebEnabled = function() {
-            return angular.isDefined(this.operatingsystem_name) && (this.operatingsystem_name.includes("Debian") || this.operatingsystem_name.includes("Ubuntu"));
+            return angular.isObject(this.operatingsystem_name) && (this.operatingsystem_name.includes("Debian") || this.operatingsystem_name.includes("Ubuntu"));
```
jlsherrill: i used isObject because iirc we were favoring using angular over lodash

jturel:
For content hosts without an operating system it works great! But for my Debian box I don't see the Deb Packages tab.

I added a test case to test the true case and it still fails:

```
     it("determines if its debian with debian os", function() {
        host = new Host({operatingsystem_name: 'Debian 9'})
        expect(host.isDebEnabled()).toEqual(true)
     });
```

jlsherrill:
@jturel reworked some stuff and added tests cases!

jturel:
[test katello]

PR #8205, opened 2019-07-09
Fixes #27258 - Install errata only to selected hosts

jturel:
From Hosts -> Content Hosts, selecting a subset of hosts for errata installation via REX initiates the action for all hosts in the lists - not just the selected ones. The root cause is that the form was sending an empty string for the `scoped_search` param which matches all hosts in BulkHostsExtensions. I think that's the correct behavior so I did this:

- removed the scoped_search param from errata and modularity bulk actions form submissions (looks like they were unused)
- adjusted remote_execution_controller to remove the 'excluded' option - it doesn't appear to be used by any REX action and was not in the proper location in the hash anyway

To test with this PR applied:

- Set up a content host w/ some applicable errata(Host A), and one or more other content hosts
- From the Content Hosts lists select HostA 
- Select Action -> Manage Errata -> select some errata -> install via remove execution

The following page should indicate that the errata will be installed only on HostA and not all of them

### DIFF
--- a/app/controllers/katello/remote_execution_controller.rb
+++ b/app/controllers/katello/remote_execution_controller.rb
@@ -31,13 +31,15 @@ module Katello
 
       def hosts
         host_ids = params[:host_ids].is_a?(String) ? params[:host_ids].split(',') : params[:host_ids]
-        find_bulk_hosts('edit_hosts',
+
+        bulk_params = {
           included: {
             ids: host_ids,
-            search: params[:scoped_search],
-            excluded: params[:excluded]
+            search: params[:scoped_search]
           }
-                       )
+        }
+
+        find_bulk_hosts('edit_hosts', bulk_params)
       end
 
       def inputs
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-host-bulk-module-streams-modal.html
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-host-bulk-module-streams-modal.html
@@ -9,7 +9,6 @@
       <input type="hidden" name="remote_action" ng-value="moduleStreamActionFormValues.remoteAction"/>
       <input type="hidden" name="module_stream_action" ng-value="moduleStreamActionFormValues.moduleStreamAction"/>
       <input type="hidden" name="module_spec" ng-value="moduleStreamActionFormValues.moduleSpec"/>
-      <input type="hidden" name="scoped_search" ng-value="moduleStreamActionFormValues.search"/>
       <input type="hidden" name="host_ids" ng-value="moduleStreamActionFormValues.hostIds"/>
       <input type="hidden" name="customize" ng-value="moduleStreamActionFormValues.customize"/>
       <input type="hidden" name="authenticity_token" ng-value="moduleStreamActionFormValues.authenticityToken"/>
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-errata-modal.html
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-errata-modal.html
@@ -5,7 +5,6 @@
     <form id="errataActionForm" name="errataActionForm" class="form" method="post" action="/katello/remote_execution">
       <input type="hidden" name="name" ng-value="errataActionFormValues.errata"/>
       <input type="hidden" name="remote_action" ng-value="errataActionFormValues.remoteAction"/>
-      <input type="hidden" name="scoped_search" ng-value="errataActionFormValues.search"/>
       <input type="hidden" name="host_ids" ng-value="errataActionFormValues.hostIds"/>
       <input type="hidden" name="authenticity_token" ng-value="errataActionFormValues.authenticityToken"/>
       <input type="hidden" name="customize" ng-value="errataActionFormValues.customize"/>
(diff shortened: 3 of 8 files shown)


jturel:
[test katello]

jturel:
[test katello]

jturel:
[test katello]

jturel:
[test katello]

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-errata-modal.html
```diff
@@ -5,7 +5,6 @@ <h4 data-block="modal-header">Content Host Errata Management</h4>
     <form id="errataActionForm" name="errataActionForm" class="form" method="post" action="/katello/remote_execution">
       <input type="hidden" name="name" ng-value="errataActionFormValues.errata"/>
       <input type="hidden" name="remote_action" ng-value="errataActionFormValues.remoteAction"/>
-      <input type="hidden" name="scoped_search" ng-value="errataActionFormValues.search"/>
```
sseelam2: I don't think we need the below line in `content-hosts-bulk-errata-modal.controller.js` since we removed the form input for search.

https://github.com/Katello/katello/blob/4a99b51048fc0816c3b4237f91448b39ae9af388/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-errata-modal.controller.js#L54

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-host-bulk-module-streams-modal.html
```diff
@@ -9,7 +9,6 @@ <h3 translate ng-hide="selectedModuleStreamOption === 'available'">Available Mod
       <input type="hidden" name="remote_action" ng-value="moduleStreamActionFormValues.remoteAction"/>
       <input type="hidden" name="module_stream_action" ng-value="moduleStreamActionFormValues.moduleStreamAction"/>
       <input type="hidden" name="module_spec" ng-value="moduleStreamActionFormValues.moduleSpec"/>
-      <input type="hidden" name="scoped_search" ng-value="moduleStreamActionFormValues.search"/>
```
sseelam2: `search` is being set in `content-hosts-bulk-module-streams-modal.controller.js`. Should we put back the above line in content-host-bulk-module-streams-modal.html ?

https://github.com/Katello/katello/blob/4a99b51048fc0816c3b4237f91448b39ae9af388/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-module-streams-modal.controller.js#L42

jturel: I'll need to look into this but it appears there's some issue with the modularity modal - will need to take a look at that

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-errata-modal.html
jturel: I think you're right. The field was called 'scoped_search' anyway - so this was not working from what I can see

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-host-bulk-module-streams-modal.html
jturel: Actually, I think it's dead code so I'll remove it from the controller

jturel:
@parthaa @sjha4 can one/both take a look at this? I think you're familiar with the bulk stuff.

In some of bulk host modal controller I've removed code which sets a `search` value on the form that does not exist in the HTML. I've also removed 'scoped_search' from some of the modal views. I think both of them are not needed because those forms work entirely with the *selected* hosts. Am I right in saying that?

jturel:
@parthaa @sjha4 any input on this one regarding the open question above?

parthaa:
code looks ok. Going to run some tests

jturel:
Thank you @parthaa.

I'll wait for an ACK from @sseelam2 since she is reviewing also

app/controllers/katello/remote_execution_controller.rb
```diff
@@ -31,13 +31,15 @@ def prepare_composer
...
         host_ids = params[:host_ids].is_a?(String) ? params[:host_ids].split(',') : params[:host_ids]
-        find_bulk_hosts('edit_hosts',
+
+        bulk_params = {
           included: {
             ids: host_ids,
-            search: params[:scoped_search],
-            excluded: params[:excluded]
+            search: params[:scoped_search]
           }
```
sjha4: > In some of bulk host modal controller I've removed code which sets a `search` value on the form that does not exist in the HTML. I've also removed 'scoped_search' from some of the modal views. I think both of them are not needed because those forms work entirely with the _selected_ hosts. Am I right in saying that?

One thing is we could add the excluded hash to the bulk_params outside the included hash:
The bulk_hosts_extenstion looks for this parameter.https://github.com/Katello/katello/blob/05ef323476286317e282ee93a43ba2504e9c09f7/app/controllers/katello/concerns/api/v2/bulk_hosts_extensions.rb#L30. 
I am not sure if people use this outside the UI though.

jturel: In terms of the remote execution controller it's not used through the UI and also considering that the it's not an API-related controller I decided to remove it since it really had no function that I could determine

sjha4: So from what I understood, we post on /katello/remote_execution route which redirects to this controller but I might be wrong. I think it's safe to drop the excluded param since this has not worked for a long time anyway so nothing outside the modal controller is probably concerned with this.. :+1:

jturel: Exactly!

sseelam2:
It looks good. ACK :+1:

PR #8206, opened 2019-07-10
Fixes #27267 - Updates for rubocop 0.71.0

jturel:
Some things have changed with rubocop in that, for example, the rails and performance cops are now separate gems, so regenerating this todo.yaml was a little tricker this time and I'll lay out what I had to do in order to regenerate the todo:

```
[vagrant@centos7-devel foreman]$ bundle show | grep rubocop
  * rubocop (0.71.0)
  * rubocop-checkstyle_formatter (0.4.0)
  * rubocop-performance (1.4.0)
  * rubocop-rails (2.0.1)
```

In katello dir:
```
gem install rubocop --version=0.71.0
gem install rubocop-rails --version=2.0.1
gem install rubocop-performance --version=1.4.0
```

I also had to add the new requirements to .rubocop.yaml (see the changes)

Then regenerating worked correctly: `rubocop --auto-gen-config --exclude-limit=0`


.rubocop.yml
```diff
@@ -1,3 +1,7 @@
+require:
+  - rubocop-performance
+  - rubocop-rails
```
jturel: without this there will be a lot of 'unknown cop' messages

PR #8207, opened 2019-07-10
Fixes #27268 - Expose ansible collection endpoints

sjha4:


jturel:
[test katello]

app/views/katello/api/v2/ansible_collections/base.json.rabl
```diff
@@ -0,0 +1,8 @@
+object @resource
+
+attributes :id
+attributes :name
+attributes :namespace
+attributes :version
+attributes :checksum
+attributes :pulp_id
```
jlsherrill: i know we do this for other types, but i wonder if we should stop providing a pulp_id, i know i'd like to see us do that long term, maybe we can go ahead and start with ansible.

jlsherrill:
Overall looks good and seems to work fine, one small comment and then its good!

app/views/katello/api/v2/ansible_collections/base.json.rabl
sjha4: Removed pulp_id from rabl.

PR #8208, opened 2019-07-11
Fixes #26245 - Sync pulp3 capsule - file content

parthaa:
This commit contains updates for pulp3 proxy work. Its started with breaking the capsule content to Pulp2 and Pulp3 capsule content tasks.
Before this PR, capsule content sync using
```ForemanTasks.sync_task(Actions::Katello::CapsuleContent::Sync, SmartProxy.last, {}) ``` would return the following plan 
pulp2-capsule
Capsule Content would do the following
1. CapsuleContent::Sync - Fetch all the repositories that need to be created/updated
2. RefreshNeeded - Refresh the importer and exporter settings for repos that need updates. This was a fairly expensive operation done at the plan phase.
3. CreateRepos - Create a repositories on capsule that need to be created. The feed of this repo on the capsule points to the main server.
4. UnassociateUnits - Clear the repository so that the sync "mirrors" the master repo
5. SyncCapsule - Finally Sync the capsule on the mirror 

This has now been modified to account for both pulp2 and pulp3
```ForemanTasks.sync_task(Actions::Katello::CapsuleContent::Sync, SmartProxy.last, {}) ``` would return the following plan
pulp3-capsule

This PR Modifies the following aspects
1. Enables sync across a pulp capsule supporting both pulp2/pulp3 repository types. We do not want to configure all capsules to be pulp3 only. For example "file" repo's are available for both pulp2 and pulp3. Now instead of expecting the user to update all capsules to pulp3 on sat upgrade, we make the code flexible enough to handle both pulp2 or pulp3 file repp. 
2. Plans different orchestration steps based on pulp2/3. 
3. Pulp3 requires around 3 or 4 additional artifacts for repo creation. These include creating a repo, a remote, version, publication and distribution
2. RefreshRepos - This replaces RefreshNeeded. It creates/updates repo mirror artifacts based on its checks. Unline RefreshNeeded this code runs the same logic inside the "run" phase.
4. The pulp3 repositories run sync, generate metadata and create/update distribution tasks.
5. The pulp2 repos follow the traditional path via SmartProxySync and SyncCapsule tasks.

Other changes include adding "mirror" methods to pulp3 repository service class. I noticed that creating repos on mirror's needed different configuration from creating the same on master, hence it made sense to break it into separate methods.

### CODE REVIEW

File: app/lib/actions/pulp/orchestration/repository/smart_proxy_sync.rb [hunk 1]
```diff
@@ -0,0 +1,50 @@
...
+                            capsule_id: smart_proxy.id,
+                            force: true)
+              end
+              if repository.is_a?(::Katello::Repository) &&
+                repository.distribution_bootable? &&
+                repository.download_policy == ::Runcible::Models::YumImporter::DOWNLOAD_ON_DEMAND
+                plan_action(Katello::Repository::FetchPxeFiles,
+                            id: repository.id,
+                            capsule_id: smart_proxy_service.smart_proxy.id)
+              end
```
jlsherrill: I think this block that fetches FetchPxeFiles could be put on the main sync capsule task, not in the pulp2 specific one

app/lib/actions/katello/capsule_content/sync_capsule.rb
```diff
@@ -0,0 +1,49 @@
...
+          repository = ::Katello::Repository.find(repository_id) if repository_id
+          skip_metadata_check = options.fetch(:skip_metadata_check, false)
+
+          concurrence do
+            available_repositories(smart_proxy, environment, content_view, repository).each do |repo|
+              plan_pulp_action([Actions::Pulp::Orchestration::Repository::SmartProxySync,
+                                Actions::Pulp3::CapsuleContent::Sync],
+                                 repo, smart_proxy,
+                                 content_view: content_view,
+                                 environment: environment,
```
jlsherrill: if we're passing in each repo, do we need to pass content_view and environment here?

app/lib/actions/pulp3/capsule_content/create_publication.rb
```diff
@@ -0,0 +1,31 @@
...
+          sequence do
+            if !::Katello::RepositoryTypeManager.find(repository.content_type).pulp3_skip_publication
+              action = plan_self(:repository_id => repository.id, :smart_proxy_id => smart_proxy.id, :contents_changed => options[:contents_changed], :options => options)
+              plan_action(RefreshDistribution, repository, smart_proxy,
+                            :tasks => action.output,
+                            :use_repository_version => false,
+                            :contents_changed => options[:contents_changed])
+            else
+              plan_action(RefreshDistribution, repository, smart_proxy,
+                            :tasks => action.output,
```
jlsherrill: action here is undefined, so it won't work properly :)  should refresh_distribution look up the latest version?

app/lib/actions/pulp3/capsule_content/generate_metadata.rb
```diff
@@ -0,0 +1,15 @@
+module Actions
+  module Pulp3
+    module CapsuleContent
+      class GenerateMetadata < Pulp3::Abstract
+        def plan(repository, smart_proxy, options = {})
+          options[:contents_changed] = (options && options.key?(:contents_changed)) ? options[:contents_changed] : true
+          sequence do
+            plan_action(CreateVersion, repository, smart_proxy)
```
jlsherrill: why are you creating a version for metadata generation?

app/lib/actions/pulp3/capsule_content/create_publication.rb
jlsherrill: oh it looks like refreshDistribution doesn't need an action if use_repository_version is true, so maybe just omit this

app/lib/actions/pulp3/capsule_content/sync.rb
```diff
@@ -0,0 +1,34 @@
...
+    module CapsuleContent
+      class Sync < Pulp3::Abstract
+        def plan(repository, smart_proxy, options = {})
+          content_view = options[:content_view]
+          environment = options[:environment]
+          sequence do
+            refresh_options = {}
+            refresh_options[:content_view_id] = content_view.id if content_view
+            refresh_options[:environment_id] = environment.id if environment
+            refresh_options[:repository_id] = repository.id if repository
```
jlsherrill: this hash (and all the options) seem completely unused

app/services/katello/pulp3/repository.rb
```diff
@@ -45,8 +45,22 @@ def distribution_options
...
+            base_path: path,
+            name: "#{backend_object_name}"
+        }
+        ret[:publication] = options[:publication] if options.key? :publication
+        ret[:repository_version] = options[:repository_version] if options.key? :repository_version
+        ret
+      end
+
       def remote_options
-        fail NotImplementedError
```
jlsherrill: why remove this? the reason it wasn't implemented is that the plugins are meant to implement it

app/services/katello/pulp3/repository.rb
```diff
@@ -57,21 +71,26 @@ def api_client
...
+      def create_mirror_remote
+        remote_file_data = remote_class.new(mirror_remote_options)
+        response = remotes_api.create(remote_file_data)
+      end
+
       def create_remote
         remote_file_data = remote_class.new(remote_options)
         response = remotes_api.create(remote_file_data)
-        repo.update_attributes!(:remote_href => response._href)
+        repo.update_attributes!(:remote_href => response._href) if @smart_proxy.pulp_master?
```
jlsherrill: is this check needed, look like create_mirror_remote does not call this

app/services/katello/pulp3/repository.rb
```diff
@@ -83,6 +102,11 @@ def remote_partial_update
         remotes_api.partial_update(repo.remote_href, remote_options)
       end
 
+      def delete_mirror_remote
+        remote = list_remotes(name: backend_object_name).first
+        remotes_api.delete(remote._href) if remote
+      end
```
jlsherrill: this method seems to not be used

app/services/katello/pulp3/repository.rb
```diff
@@ -123,8 +152,36 @@ def distribution_reference(path)
...
+        create_mirror_remote
+      end
+
+      def create_mirror
+        repositories_api.create(name: backend_object_name)
+      end
+
       def create
-        unless repository_reference
+        unless  repository_reference
```
jlsherrill: extra space

app/services/katello/pulp3/repository.rb
```diff
@@ -266,13 +395,65 @@ def ssl_remote_options
...
+          ssl_client_certificate: ueber_cert[:cert],
+          ssl_client_key: ueber_cert[:key],
+          ssl_ca_certificate: ::Cert::Certs.ca_cert,
+          ssl_validation: true
+        }
+      end
+
+      def external_url(force_https = false)
+        uri = URI.parse(::SmartProxy.pulp_master.pulp3_url)
+        uri.scheme = (root.unprotected && !force_https) ? 'http' : 'https'
```
jlsherrill: this should always be https

app/services/katello/pulp3/repository.rb
```diff
@@ -45,8 +45,22 @@ def distribution_options
...
 
+      def distribution_mirror_options(path, options = {})
+        ret  = {
+            base_path: path,
+            name: "#{backend_object_name}"
+        }
+        ret[:publication] = options[:publication] if options.key? :publication
+        ret[:repository_version] = options[:repository_version] if options.key? :repository_version
+        ret
+      end
```
jlsherrill: if you could move this lower in the file, the intent was to keep all the method that simpley `fail NotImplementedError`  together

app/lib/actions/katello/capsule_content/sync_capsule.rb
parthaa: we are not passing in each repo here. Its across all available repos for the capsule

app/lib/actions/katello/capsule_content/refresh_repos.rb
```diff
@@ -0,0 +1,56 @@
...
+        def invoke_external_task
+          tasks = []
+          environment = ::Katello::KTEnvironment.find_by(:id => input[:environment_id]) if input[:environment_id]
+          repository = ::Katello::Repository.find_by(:id => input[:repository_id]) if input[:repository_id]
+          if repository.nil? && input[:repository_id]
+            repository = ::Katello::ContentViewPuppetEnvironment.find(input[:repository_id])
+            repository = repository.nonpersisted_repository
+          end
+          content_view = ::Katello::ContentView.find_by(:id => input[:content_view_id]) if input[:content_view_id]
+          smart_proxy = SmartProxy.find(input[:smart_proxy_id])
```
jlsherrill: use .unscoped here on the smart proxy lookup

app/lib/actions/katello/capsule_content/sync.rb
```diff
@@ -39,9 +39,11 @@ def plan(smart_proxy, options = {})
...
                                              repository: repository)
-          plan_action(Actions::Pulp::Orchestration::Repository::RefreshRepos, smart_proxy, refresh_options)
-          plan_action(Actions::Pulp3::Orchestration::Repository::RefreshRepos, smart_proxy, refresh_options)
-          plan_action(SyncCapsule, smart_proxy, refresh_options)
+          sequence do
+            plan_action(Actions::Pulp::Orchestration::Repository::RefreshRepos, smart_proxy, refresh_options)
+            plan_action(Actions::Pulp3::Orchestration::Repository::RefreshRepos, smart_proxy, refresh_options)
+            plan_action(SyncCapsule, smart_proxy, refresh_options)
+          end
         end
```
sjha4: Yayyy! Fixed all issues.. :smile: Will run through use-cases on the PR. Meanwhile, I see the repo, version and the contents so passes smoke test. :+1:

parthaa:
[test katello]

app/services/katello/pulp3/repository.rb
```diff
@@ -123,6 +147,40 @@ def distribution_reference(path)
...
+        [:ssl_client_certificate, :ssl_client_key, :ssl_ca_certificate].each do |key|
+          computed_options[key] = Digest::SHA256.hexdigest(computed_options[key].chomp)
+        end
+        computed_options
+      end
+
+      def create_mirror_entities
+        create_mirror
+        create_mirror_remote unless mirror_remote_href
+      end
```
sjha4: Do we need this check at creation time?

app/lib/actions/pulp3/capsule_content/sync.rb
```diff
@@ -0,0 +1,26 @@
...
+          sequence do
+            plan_self(:repository_id => repository.id, :smart_proxy_id => smart_proxy.id, :options => options)
+            plan_action(GenerateMetadata, repository, smart_proxy, options)
+          end
+        end
+
+        def invoke_external_task
+          repo = ::Katello::Repository.find(input[:repository_id])
+          output[:pulp_tasks] = repo.backend_service(::SmartProxy.unscoped.find(input[:smart_proxy_id])).sync_mirror
+        end
```
sjha4: I don't know how often full capsule sync is run but we would be syncing all repositories on the capsule on every refresh with this?

app/lib/actions/katello/capsule_content/refresh_repos.rb
```diff
@@ -0,0 +1,55 @@
...
+        def act_on_repo?(_repo)
+          fail NotImplementedError
+        end
+
+        def plan(smart_proxy, options = {})
+          plan_self(:smart_proxy_id => smart_proxy.id,
+                    :environment_id => options[:environment]&.id,
+                    :content_view_id => options[:content_view_id]&.id,
+                    :repository_id => options[:repository]&.id)
+        end
```
sjha4: Works well for pulp3 Library repos per my testing. For CVs in other repos, I see this error: http://pastebin.test.redhat.com/787719 on capsule sync. 
Probably cause of this : `:content_view_id => options[:content_view_id]&.id,`

sjha4: Pulp2 yum works well for library env..For CV in other envs I see the above error for pulp2 and 3..

app/services/katello/pulp3/repository.rb
parthaa: Yes. The idea here is that the case where you have a remote but not a repo also needs to get handled. This line will handle that scenario.

app/lib/actions/pulp3/capsule_content/sync.rb
parthaa: Actually only syncable repos are synced. This was the case even in the prior implementation -> 

https://github.com/Katello/katello/blob/master/app/lib/actions/katello/capsule_content/sync.rb#L43-L60

Its the same set here also.

sjha4:
Works for pulp2 and pulp3 for CV repos. :heavy_check_mark:

jlsherrill:
ACKity ACK ACK

PR #8209, opened 2019-07-11
Fixes #28907 - API doc does mention org/loc in create/update

ofedoren:
Related to: https://github.com/theforeman/foreman/pull/6892

Updates /api/v2/organizations/[create | update]
enpoints to mention the possibility to assing locations via location_ids param.

### DIFF
--- a/app/controllers/katello/api/v2/organizations_controller.rb
+++ b/app/controllers/katello/api/v2/organizations_controller.rb
@@ -32,6 +32,7 @@ module Katello
         param :environment_ids, Array, N_("Environment IDs"), :required => false
         param :subnet_ids, Array, N_("Subnet IDs"), :required => false
         param :label, String, :required => false
+        param :location_ids, Array, N_("Associated location IDs"), :required => false
       end
     end
 
@@ -55,13 +56,9 @@ module Katello
     end
 
     api :PUT, '/organizations/:id', N_('Update organization')
-    # The organization param hash below is redefined from foreman's ::Api::V2::TaxonomiesController
-    # resource param_group instead of reusing the param_group :resource scoped from TaxonomiesController.
-    # This is because name substitutions of the param group's name from :resource to :organization are limited
-    # to the inclusion of a modules.
     param :id, :number, :desc => N_("organization ID"), :required => true
     param :redhat_repository_url, String, :desc => N_("Red Hat CDN URL")
-    param_group :resource, ::Api::V2::TaxonomiesController
+    param_group :resource
     def update
       if params.key?(:redhat_repository_url)
         org_params = params.permit(:redhat_repository_url).to_h


evgeni:
[test katello]

app/controllers/katello/api/v2/organizations_controller.rb
```diff
@@ -61,7 +62,7 @@ def show
     # to the inclusion of a modules.
     param :id, :number, :desc => N_("organization ID"), :required => true
     param :redhat_repository_url, String, :desc => N_("Red Hat CDN URL")
-    param_group :resource, ::Api::V2::TaxonomiesController
+    param_group :resource
```
evgeni: Could you explain why this is needed and probably also remove the comment in lines 59-62?

My understanding is that `PUT` (currently) uses the `resource` definition from Foreman, not the one from Katello, and that is actually wrong, as Foreman's allows parents (Katello doesn't)

At least that's what I see when I look at the Katello docs at https://theforeman.org/plugins/katello/3.12/api/apidoc/v2/organizations/update.html and https://theforeman.org/plugins/katello/3.12/api/apidoc/v2/organizations/create.html

cc @mdellweg for more API consistency "fun"

evgeni: Fun fact, the comment was from a commit removing the v2 taxonomies controller:
https://github.com/Katello/katello/commit/5dfda340f90f93039de436203b6b5b2f66186c6a but the actual code that commit added was removed in df52a9bd77928099e74c495122e85e04185660b3 and v2 taxonimies added back in 96c101bb3cc6e20625b4ea9ad16a29737a7d221d

and then the new, local `:resource` was defined in a08199b78ceb149992f34cd4f56b0cab829d03fd but only applied to create, not update.

@jlsherrill @bbuckingham @ehelms @parthaa can we have a sit down at a nice :fire: and y'all tell us about "The History of the Organizations Controller"?

jlsherrill: I don't remember the full history, but i think we did it the way we did for a few reasons:

1) at the time there was no way to add in apipie params (there now is)
2) we wanted to hook in dynflow into the org CRUD process (there should be other ways of doing that)
3) at the time there was no way to add in json to rendered rabl, I think this is likely still a gap, but we may be able to overcome it in some other way.

@bbuckingham and @ehelms may remember more reasons since they did the initial enginification, but if these 3 are the only reasons we may be able to look at migrating away from our current setup.

bbuckingham: Correct.  This one goes way way back to when Katello was an independent rails application and was being converted to a rails engine (aka. Foreman plugin).  At that time, both apps had an organizations controller and Katello needed to add additional attributes to it.   When this was initially done, there was no simple way to add those attributes without defining a separate controller.  Much has changed and improved in the last 5 years (except perhaps this lovely piece of code that has lived on :)).

evgeni:
Is there any reason this is `Waiting on contributor` besides that the references redmine is not in Katello?

chris1984:
Starting to test

ofedoren:
@chris1984, done :)

PR #8210, opened 2019-07-11
Fixes #27279 - only save upstream auth when intentional.

jlsherrill:


jlsherrill:
This was kinda tricky to reproduce, i had to do all of this (but all may not be required):

*    access the server over https with the correct hostname of the system
*    access the server using a trusted ca (import the server's ca as trusted)
*    logout of foreman and log back in, chrome should ask to save the password, confirm that
*    go to a custom repository, click the 'upstream auth' field, click cancel
*    try to save another field
*   you'll notice the username and pass are saved in the database

I can't say this pr resolves all issues with upstream auth and saved fields, as the root cause is that we can't disable auto-fill on those fields, and chrome seems adamant about saving the password.

jlsherrill:
[test katello]

PR #8212, opened 2019-07-15
Fixes #27288 - CCV environment repo shouldn't be a master

hao-yu:
An environment repository in a composite content view may
mistakenly be considered as a master repository when its
archived repository has more than 1 source repositories.
This causes the environment repository to fetch the contents
from Pulp instead of its archived repository while doing the
"index_content". Since no contents are associated to the
enviroment repository in Pulp so nothing will be indexed.
This causes the content hosts to get 0 installable content.


jlsherrill:
[test katello]

PR #8214, opened 2019-07-16
[WIP] Provide UI for Ansible Collections

sjha4:
WIP


webpack/scenes/AnsibleCollections/Details/AnsibleCollectionDetails.js
```diff
@@ -0,0 +1,128 @@
+import React, { Component } from 'react';
+import { Nav, NavItem, TabPane, TabContent, TabContainer, Grid, Row, Col } from 'patternfly-react';
+import BreadcrumbsBar from 'foremanReact/components/BreadcrumbBar';
+import { PropTypes } from 'prop-types';
+import { LoadingState } from '../../../move_to_pf/LoadingState';
+import api from '../../../services/api';
+import AnsibleCollectionDetailsInfo from './AnsibleCollectionDetailsInfo';
+import AnsibleCollectionDetailRepositories from '../../ModuleStreams/Details/Repositories/ModuleStreamDetailRepositories';
```
sjha4: @johnpmitsch : This repositories component would be a complete rewrite for Collections. We could move this out as a utility component?

webpack/scenes/AnsibleCollections/Details/AnsibleCollectionDetailsInfo.js
```diff
@@ -0,0 +1,46 @@
...
+//   /* eslint-disable no-restricted-syntax, react/jsx-closing-tag-location */
+//   for (const key of mapping.keys()) {
+//     rows.push(<tr key={key}>
+//       <td><b>{mapping.get(key)}</b></td>
+//       <td>{details[key]}</td>
+//     </tr>);
+//   }
+//   /* eslint-enable no-restricted-syntax, react/jsx-closing-tag-location */
+//   return rows;
+// };
```
sjha4: Same with this function.

webpack/scenes/AnsibleCollections/Details/AnsibleCollectionDetails.js
johnpmitsch: +1

webpack/scenes/AnsibleCollections/Details/AnsibleCollectionDetailsInfo.js
johnpmitsch: I think you could make a shared Detail component as well

johnpmitsch:
@sjha4 The react code looks good, I didn't get a chance to run it, but I think you are on the right track by thinking about de-duplicating the code. This is the format of a lot of our content pages and we should try to have common components. 

My thought would be to have a `content` folder in `webpack/components` that holds these common content components.

something like
```
-- webpack/
---- components/
------ content/
-------- ContentPage.js
-------- ContentTable.js
-------- Details/
---------- ContentDetailsPage.js
---------- ContentDetailsInfo.js
---------- ContentDetailRepositories.js
```

The details page may be tricky because of the number of tabs and pages will vary. I can go over different approaches with you, but we should keep things flexible and allow custom components to be inserted into these pages. Module stream artifacts and profiles are good examples of specific components that wouldn't be shared, but should be able to be inserted into a common component.

Also would love to hear what @Katello/ui thinks! To recap, @sjha4 is adding an ansible collections page, which is mostly similar to the existing module streams page and any future simple content pages we convert to React.

webpack/components/Content/ContentPage.js
```diff
@@ -0,0 +1,54 @@
...
+  content: PropTypes.shape({}).isRequired,
+  tableSchema: PropTypes.arrayOf(PropTypes.shape({})).isRequired,
+  onSearch: PropTypes.func.isRequired,
+  getAutoCompleteParams: PropTypes.func.isRequired,
+  updateSearchQuery: PropTypes.func.isRequired,
+  initialInputValue: PropTypes.string.isRequired,
+  onPaginationChange: PropTypes.func.isRequired,
+};
+
+export default ContentPage;
```
sjha4: @johnpmitsch : Does this look ok to you? We have a component ContentPage corresponding to the main list page for the content..I will work on the details page if this looks like a good place to start..

johnpmitsch: At a glance, that looks fine to me!

webpack/components/Content/ContentDetails.js
```diff
@@ -0,0 +1,104 @@
...
+              <Col sm={12}>
+                <Nav bsClass="nav nav-tabs">
+                  <NavItem eventKey={1}>
+                    <div>{__('Details')}</div>
+                  </NavItem>
+                  <NavItem eventKey={2}>
+                    <div>{__('Repositories')}</div>
+                  </NavItem>
+                  {tabHeaders(tabLabels)}
+                </Nav>
```
sjha4: This dynamically generates tab headers based on tab labels passed.

webpack/components/Content/ContentDetails.js
```diff
@@ -0,0 +1,104 @@
...
+                      <ModuleStreamDetailProfiles profiles={profiles} /> :
+                      __('No profiles to show')}
+                  </Col>
+                </Row>
+              </TabPane>
+              <TabPane eventKey={4}>
+                <Row>
+                  <Col sm={12}>
+                    {artifacts && artifacts.length ?
+                      <ModuleStreamDetailArtifacts artifacts={artifacts} /> :
```
sjha4: Here, for ModuleStreamDetailProfiles and ModuleStreamDetailArtifacts, even if we generate it here, we would need to import these pages in this common component. How do we go about achieving dynamic imports based on what we are trying to render? I didn't find any good resources on that and I basically verified that React's `SyntaxError: 'import' and 'export' may only appear at the top level` works.. :smile:

webpack/components/Content/ContentDetails.js
```diff
@@ -0,0 +1,82 @@
...
+                <Row>
+                  <Col sm={12}>
+                    {repositories && repositories.length ?
+                      <ModuleStreamDetailRepositories repositories={repositories} /> :
+                      __('No repositories to show')}
+                  </Col>
+                </Row>
+              </TabPane>
+              {props.children}
+            </TabContent>
```
sjha4: @johnpmitsch: props.children here..

webpack/scenes/ModuleStreams/Details/ModuleStreamDetails.js
```diff
@@ -63,70 +63,26 @@ class ModuleStreamDetails extends Component {
...
+          <TabPane eventKey={4}>
+            <Row>
+              <Col sm={12}>
+                {artifacts && artifacts.length ?
+                  <ModuleStreamDetailArtifacts artifacts={artifacts} /> :
+                  __('No artifacts to show')}
+              </Col>
+            </Row>
+          </TabPane>
+        </ContentDetails>
```
sjha4: And pass tabs here on the content specific details page..

sjha4:
@johnpmitsch: The 4th commit on the PR has a working solution with the schema passing approach based on what you suggested.

johnpmitsch:
@sjha4 Looks good!

Some suggestions:
- Don't include this part in the schema, render it in the component
```jsx
<TabPane eventKey={3}>
        <Row>
          <Col sm={12}>
```
- Include all the tabs including the info and repositories tabs in the schema, this will allow you to use the index for the `eventKey` and also keeps this implementation flexible.

So you would just have

```jsx
schema.map((tabs, index) =>
    <TabPane eventKey={i}>
        <Row>
          <Col sm={12}>
            {tab.tabBody}
          </Col>
         </Row>
      </TabPane>
```

webpack/components/Content/Details/ContentDetails.js
```diff
@@ -0,0 +1,94 @@
...
+        <div>{child.tabHeader}</div>
+      </NavItem>
+    );
+    return (
+      tabs
+    );
+  };
+
+  const tabPanes = () => {
+    const children = props.children;
```
johnpmitsch: what is props.children used for?

sjha4: I have to clean up some code from these commits.. :smile:

sjha4:
> @sjha4 Looks good!
> 
> Some suggestions:
> 
> * Don't include this part in the schema, render it in the component
> 
> ```js-jsx
> <TabPane eventKey={3}>
>         <Row>
>           <Col sm={12}>
> ```
> 
> * Include all the tabs including the info and repositories tabs in the schema, this will allow you to use the index for the `eventKey` and also keeps this implementation flexible.
> 
> So you would just have
> 
> ```js-jsx
> schema.map((tabs, index) =>
>     <TabPane eventKey={i}>
>         <Row>
>           <Col sm={12}>
>             {tab.tabBody}
>           </Col>
>          </Row>
>       </TabPane>
> ```

Sounds good..I'll make these changes..

sjha4:
Pushed changes with the above 2 points addressed. There might be some useless consts and imports etc lying around..I'll work on the lint etc once the flow starts looking ok..

johnpmitsch:
@sjha4 looks really good, let me know when you want me to review and I'll review more thoroughly! :+1:

sjha4:
@johnpmitsch: This should be ready for review. I am working on tests for the new pages and components.

webpack/components/Content/Details/TableSchema.js
```diff
@@ -4,7 +4,7 @@ import { translate as __ } from 'foremanReact/common/I18n';
 import {
```
johnpmitsch: I think `RepositoriesTableSchema` or even the very verbose `ContentDetailsRepositoriesTableSchema` would be a better name for this file so its more clear where it fits in

webpack/components/Content/Details/ContentDetails.js
```diff
@@ -0,0 +1,74 @@
...
+
+  const tabHeaders = () => {
+    const tabs = schema.map((node, index) => (
+      <NavItem eventKey={index + 1}>
+        <div>{node.tabHeader}</div>
+      </NavItem>
+    ));
+    return (
+      tabs
+    );
```
johnpmitsch: Would just `return tabs` work?

webpack/components/Content/Details/ContentDetails.js
```diff
@@ -0,0 +1,74 @@
...
+      <TabPane eventKey={index + 1}>
+        <Row>
+          <Col sm={12}>
+            {node.tabContent}
+          </Col>
+        </Row>
+      </TabPane>
+    ));
+    return (
+      tabPane
```
johnpmitsch: wondering the same for this line

webpack/containers/Application/overrides.scss
```diff
@@ -14,6 +14,10 @@ html .pagination-pf-pagesize.btn-group {
   margin-top: -6px !important;
 }
 
+td {
+  overflow: auto;
+}
+
```
johnpmitsch: what is this for? Asking because it will affect all `td`'s

sjha4: So there is https://projects.theforeman.org/issues/25622 which is around the same issue.  I can remove this and have this tackled outside of this PR or leave it here if this indeed fixes it properly.

webpack/components/Content/Details/TableSchema.js
sjha4: Renamed to ContentDetailRepositoryTableSchema.

johnpmitsch:
When creating a repository and selecting the type, it shows as `ansible_collection`. I think we should avoid the underscored name in the UI. Can we show `ansible collection` instead?

webpack/components/Content/ContentPage.js
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Grid, Col, Row, Form, FormGroup } from 'react-bootstrap';
+import Search from '../../components/Search/index';
+import ContentTable from './ContentTable';
+
+const ContentPage = props => (
```
johnpmitsch: The way you have things here works fine, but you can destructure the props object and assign to variables instead of calling props each time

for example: `const ContentPage = ({ header, onSearch, etc... }) => (`

webpack/scenes/AnsibleCollections/AnsibleCollectionsTableSchema.js
```diff
@@ -0,0 +1,60 @@
...
+            <Link to={urlBuilder('ansible_collections', '', rowData.id)}>{rowData.name}</Link>
+          </td>
+        ),
+      ],
+    },
+  },
+  {
+    property: 'namespace',
+    header: {
+      label: __('NameSpace'),
```
johnpmitsch: `NameSpace` doesn't seem like a user-friendly term, I would change (the user-facing portion of) this to be "Author" or something similar. Let me know if I'm missing some convention though.

I wonder if we should show just list these with the format `author/name`? So the Name column would show `geerlingguy/ntp`. You could even modify the API to return a display name configured like this (and could sort results alphabetically). Just a thought!

johnpmitsch:
Do you think the ability to search on `checksum` would be helpful?

webpack/scenes/AnsibleCollections/AnsibleCollectionsTableSchema.js
sjha4: The` namespace` field name is dicated by the galxy.yml metadata structure: https://docs.ansible.com/ansible/devel/dev_guide/collections_galaxy_meta.html 
There are more fields we plan to index in katello and one of them is `authors`. I think keeping the names consistent with the yml metadata would make it less confusing for now?

sjha4:
> Do you think the ability to search on `checksum` would be helpful?

Good point..We do provide search on checksum on file units so I guess it would make sense here.

webpack/scenes/AnsibleCollections/AnsibleCollectionsTableSchema.js
johnpmitsch: @Rohoover Can you provide some UX guidance on how to present this information? 

We are adding the ability for users to sync down collections from [ansible galaxy](https://galaxy.ansible.com/home). [Here](https://galaxy.ansible.com/search?tags=system&keywords=&order_by=-relevance) is an example list of collections in ansible galaxy, and [here](https://galaxy.ansible.com/newswangerd/collection_demo) is an example of a collection named "collection_demo"

Right now,  our table looks like this: 
here

What do you think is the best way to display these fields?

johnpmitsch:
I know this isn't part of the UI portion of things, but I'm not able to sync some repos
These two result in errors: `geerlingguy.java`, `robertdebock.bootstrap`
But these two sync fine: `dcos.cluster` , `newswangerd.collection_demo`

This is the error I'm seeing:
```
An error occurred during the sync 
{:code=>nil, :description=>"404 Client Error: Not Found for url: https://galaxy.ansible.com/api/v2/collections/robertdebock/bootstrap/:\n- you can use --ignore-errors to skip failed collections and finish processing the list.", :traceback=>"  File \"/usr/local/lib/pulp/lib64/python3.6/site-packages/rq/worker.py\", line 822, in perform_job\n    rv = job.perform()\n  File \"/usr/local/lib/pulp/lib64/python3.6/site-packages/rq/job.py\", line 605, in perform\n    self._result = self._execute()\n  File \"/usr/local/lib/pulp/lib64/python3.6/site-packages/rq/job.py\", line 611, in _execute\n    return self.func(*self.args, **self.kwargs)\n  File \"/usr/local/lib/pulp/src/pulp-ansible/pulp_ansible/app/tasks/collections.py\", line 73, in sync\n    repository_spec_strings=repository_spec_strings,\n  File \"/usr/local/lib/pulp/lib64/python3.6/site-packages/ansible_galaxy/actions/install.py\", line 212, in install_repository_specs_loop\n    force_overwrite=force_overwrite)\n  File \"/usr/local/lib/pulp/lib64/python3.6/site-packages/ansible_galaxy/actions/install.py\", line 99, in install_repositories_matching_repository_specs\n    force_overwrite=force_overwrite)\n  File \"/usr/local/lib/pulp/lib64/python3.6/site-packages/ansible_galaxy/actions/install.py\", line 317, in install_repositories\n    force_overwrite=force_overwrite)\n  File \"/usr/local/lib/pulp/lib64/python3.6/site-packages/ansible_galaxy/actions/install.py\", line 391, in install_repository\n    raise_without_ignore(ignore_errors, e)\n  File \"/usr/local/lib/pulp/lib64/python3.6/site-packages/ansible_galaxy/actions/install.py\", line 39, in raise_without_ignore\n    raise exceptions.GalaxyError(message)\n"}
Missing the required parameter 'repository_version_href' when calling RepositoriesVersionsApi.read
```

config/routes.rb
```diff
@@ -26,6 +26,11 @@
 
   match '/module_streams' => 'react#index', :via => [:get]
 
+  match '/ansible_collections' => 'react#index', :via => [:get]
+  match '/ansible_collections/*page' => 'react#index', :via => [:get]
+
+  match '/ansible_collections' => 'react#index', :via => [:get]
```
johnpmitsch: is this a duplicate of line 29?

webpack/scenes/AnsibleCollections/Details/AnsibleCollectionsSchema.js
```diff
@@ -0,0 +1,33 @@
...
+  ['version', __('Version')],
+  ['checksum', __('Checksum')],
+]);
+
+export default (detailInfo) => {
+  const { repositories } = detailInfo;
+
+  return [
+    {
+      key: 1,
```
johnpmitsch: Did using the index of the array not work for the `eventKey` prop for the tabs?

config/routes.rb
sjha4: Looks like it.. :smile:  Will edit..

webpack/components/Content/Details/ContentDetails.js
```diff
@@ -0,0 +1,70 @@
...
+import { LoadingState } from '../../../move_to_pf/LoadingState/index';
+
+const ContentDetails = (props) => {
+  const { contentDetails, schema } = props;
+  const { loading } = contentDetails;
+
+  const tabHeaders = () => {
+    const tabs = schema.map(node => (
+      <NavItem key={node.key} eventKey={node.key}>
+        <div>{node.tabHeader}</div>
```
sjha4: So this map here needed a key or else there was this every list item must have unique key error I was hitting. So instead of using index which I was unsure of regarding ordering here in the tabHeader vs the tabPane for the key, I thought it would be better and more controlled to provide a key in the schema.

johnpmitsch:
@sjha4 I think the only remaining thing is looking into if we are able to catch the error and show a better one when syncing roles?

sjha4:
> I think the only remaining thing is looking into if we are able to catch the error and show a better one when syncing roles?

@johnpmitsch : I will handle that separately with https://projects.theforeman.org/issues/27565.

webpack/scenes/AnsibleCollections/AnsibleCollectionsTableSchema.js
Rohoover: @johnpmitsch Overall this is really consistent with the rest of Satellite. 
A few thoughts"
1) Is "NameSpace" the correct column header? In Ansible the tooltip denotes this content as "Author"
2) Is there a reason we don't show the Quality Score? Is it because of its community origins?
3) The Ansible UI allows you to install different versions, are we mimicking this?
4) I'd recommend an export button at the top.

sjha4: @Rohoover :
> Is "NameSpace" the correct column header? In Ansible the tooltip denotes this content as "Author"

So the exact [description](https://docs.ansible.com/ansible/devel/dev_guide/collections_galaxy_meta.html#collections-galaxy-meta) of the namespace field is: 
```
The namespace of the collection.
This can be a company/brand/organization or product namespace under which all content lives.
May only contain alphanumeric characters and underscores. Additionally namespaces cannot start with underscores or numbers and cannot contain consecutive underscores.
```
I am open to using a more generic term on UI if needed.
****

> Is there a reason we don't show the Quality Score? Is it because of its community origins?

The quality score is not indexed in katello right now. These are newer fields that pulp has now and we plan to index those in a future PR. We do plan to bring it in.

****
> The Ansible UI allows you to install different versions, are we mimicking this?

Yes, a user should be able to sync different versions to his repo if he needs to.

johnpmitsch: > So the exact [description](https://docs.ansible.com/ansible/devel/dev_guide/collections_galaxy_meta.html#collections-galaxy-meta) of the namespace field is:
> 
> ```
> The namespace of the collection.
> This can be a company/brand/organization or product namespace under which all content lives.
> May only contain alphanumeric characters and underscores. Additionally namespaces cannot start with underscores or numbers and cannot contain consecutive underscores.
> ```
> 
> I am open to using a more generic term on UI if needed.

I am in favor of "Author" in the UI since the user is more likely to see it listed as this in Ansible Galaxy than see the namespace metadata field, which I'm not entirely sure the typical user would see. The backend fields could still use `namespace`, I don't see an issue with this.  

But I defer to @rohoover's judgement :)

Rohoover: Let's go with "Author". Namespace to me, isn't a real word and parity between the apps will be helpful to users. 

Is the synching of different versions happening in the detailed page upon click of "collection_demo"?
Is each version going to be a separate line item in this table?

sjha4: @Rohoover
> Is the synching of different versions happening in the detailed page upon click of "collection_demo"? Is each version going to be a separate line item in this table?

So the collection_demo in the table would be already synced. If there are multiple versions, they are treated as separate records and would be separate line items in the table.

There's currently some discussions around how we sync in https://pulp.plan.io/issues/5250. We will probably let users specify certain versions when syncing and those versions would get synced.

> Let's go with "Author". Namespace to me, isn't a real word and parity between the apps will be helpful to users.

I will update this to Author.

sjha4:
@Rohoover , @johnpmitsch 

Updated Namespace -> Author:

List Page:
List Page

Details Page:
Detail

I have 2 issues opened following discussions on this thread to address the backend changes necessary to handle them on the UI:
1. [Throw descriptive error when trying to sync roles from Galaxy
](https://projects.theforeman.org/issues/27579)
2. [Index additional fields for collection version from pulp](https://projects.theforeman.org/issues/27608)

sjha4:
Following https://github.com/Katello/katello/pull/8267, the new flow would be to create an ansible collection repo with only URLs in the form `https://galaxy-url/api/v2/collections/namespace/name` .
Some example URLs would be:
https://galaxy.ansible.com/api/v2/collections/testing/k8s_demo_collection
https://galaxy.ansible.com/api/v2/collections/devoperate/base

sjha4:
@johnpmitsch: 
Does [this change](https://github.com/Katello/katello/compare/0cf4f0ba3cd623bf38ce9993f72f32a280006f7f..42eaa20403e7a7039bef3aa38279839d4833011f) look good to you for hiding the whitelist field? I am keeping track of the pulp task that needs to support this and will revert when it's ready.

johnpmitsch:
@sjha4 How do users know to sync based on the API url? I would say at the very least the Example url should be updated or add a way in the form to sync ansible galaxy roles based on the user passing in `author.name`.

Also, a comment in the code about why `&& false` was added could help to clarify why the field is hidden to other devs not in the loop on this change.

sjha4:
> @sjha4 How do users know to sync based on the API url? I would say at the very least the Example url should be updated or add a way in the form to sync ansible galaxy roles based on the user passing in `author.name`.

Updated the example URL for this PR and we can discuss how we want users to consume this based on how the related backend tasks shape up.

> Also, a comment in the code about why `&& false` was added could help to clarify why the field is hidden to other devs not in the loop on this change.

Added comments with link to pulp story to add some context.

sjha4:
[test katello]
Random test failure..

johnpmitsch:
@sjha4 It could be something with my setup, but getting an error now syncing `https://galaxy.ansible.com/api/v2/collections/debops/debops` https://paste.fedoraproject.org/paste/jXJtS9dDOC~E8ZIz0dTCPw

johnpmitsch:
@sjha4 
This is what I am seeing. I'm using just this PR checked out (not combined with the other one) and tried restarting my server and a hard refresh. The page will display the search results and then load the full base page

Peek 2019-08-23 16-45

sjha4:
> The search on the ansible collections page is acting strange, when I search for something, it displays the results, and then the page goes back to the full results page and then search is cleared. Let me know if you see the same.

I am not seeing this in my setup. :confused: 

> Also, can we make namespace searchable as "author" since that is what we show in the UI?

So I was looking at how to do this and with the scoped_search gem, the labels are returned from the server with a common logic to extract labels based on field names in a concern. 
The cleanest solution for this would be to rename the field on the model and I have opened an issue for this https://projects.theforeman.org/issues/27696 to tackle it outside this PR with a wider discussion around naming.

johnpmitsch:
> > The search on the ansible collections page is acting strange, when I search for something, it displays the results, and then the page goes back to the full results page and then search is cleared. Let me know if you see the same.
> 
> I am not seeing this in my setup. 

I'll try again and maybe someone else can confirm its not causing issues

> 
> > Also, can we make namespace searchable as "author" since that is what we show in the UI?
> 
> So I was looking at how to do this and with the scoped_search gem, the labels are returned from the server with a common logic to extract labels based on field names in a concern.
> The cleanest solution for this would be to rename the field on the model and I have opened an issue for this https://projects.theforeman.org/issues/27696 to tackle it outside this PR with a wider discussion around naming.

Can you use the `rename` parameter? [example here](https://github.com/Katello/katello/blob/fc8c3f9682f58e0855f35bbf429ebde84a4d525b/app/models/katello/erratum.rb#L33)

PR #8217, opened 2019-07-17
Fixes #26999 - support syncing docker repositories in pulp3 

ianballou:
Docker Pulp 3 sync is working, minus the presenters.  Note that, to relate the repository ID properly to a tag, I needed to add the 
```ruby 
model.repository_id = repository.id unless many_repository_associations
```
line to `pulp_database_unit.rb`.  We realized that the `manage_repository_association` method is really just checking if the relation type is many-to-many, so the name was updated to `many_repository_associations` to better suit it.  A test will be added to ensure this doesn't break.

To test this draft PR, try to sync your favourite Docker repo.  See that the repository version was updated in Pulp 3, and see that manifests and tags were created.  

Note that, due to https://pulp.plan.io/issues/5146, there will almost definitely be fewer schema version 1 manifests from the Pulp 3 sync.  Seeing none is normal.

TODO: Add tests and get the sync progress tracking working.

### DIFF
--- a/app/services/katello/pulp3/docker_manifest.rb
+++ b/app/services/katello/pulp3/docker_manifest.rb
@@ -1,0 +1,36 @@
+module Katello
+  module Pulp3
+    class DockerManifest < PulpContentUnit
+      include LazyAccessor
+
+      def self.content_api
+        PulpDockerClient::ContentManifestsApi.new(Katello::Pulp3::Repository::Docker.api_client(SmartProxy.pulp_master!))
+      end
+
+      def self.ids_for_repository(repo_id)
+        repo = Katello::Pulp3::Repository::Docker.new(Katello::Repository.find(repo_id), SmartProxy.pulp_master)
+        repo_content_list = repo.content_list
+        repo_content_list.map { |content| content.try(:_href) }
+      end
+
+      def self.content_unit_list(page_opts)
+        page_opts[:media_type] = "application/vnd.docker.distribution.manifest.v1+json"
+        data_v1 = self.content_api.list(page_opts)
+        page_opts[:media_type] = "application/vnd.docker.distribution.manifest.v2+json"
+        data_v2 = self.content_api.list(page_opts)
+
+        filtered = {}
+        filtered["count"] = data_v1.count + data_v2.count
+        filtered["results"] = data_v1.results + data_v2.results
+        filtered
+      end
+
+      def update_model(model)
+        custom_json = {}
+        custom_json['schema_version'], = backend_data['schema_version']
+        custom_json['digest'], = backend_data['digest']
+        model.update_attributes!(custom_json)
+      end
+    end
+  end
+end
--- a/app/services/katello/pulp3/docker_tag.rb
+++ b/app/services/katello/pulp3/docker_tag.rb
@@ -1,0 +1,28 @@
+module Katello
+  module Pulp3
+    class DockerTag < PulpContentUnit
+      include LazyAccessor
+
+      def self.content_api
+        PulpDockerClient::ContentManifestTagsApi.new(Katello::Pulp3::Repository::Docker.api_client(SmartProxy.pulp_master!))
+      end
+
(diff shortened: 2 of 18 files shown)


app/services/katello/pulp3/docker_manifest.rb
```diff
@@ -0,0 +1,36 @@
...
+        page_opts[:media_type] = "application/vnd.docker.distribution.manifest.v1+json"
+        data_v1 = self.content_api.list(page_opts)
+        page_opts[:media_type] = "application/vnd.docker.distribution.manifest.v2+json"
+        data_v2 = self.content_api.list(page_opts)
+
+        filtered = {}
+        filtered["count"] = data_v1.count + data_v2.count
+        filtered["results"] = data_v1.results + data_v2.results
+        filtered
+      end
```
jlsherrill: Does this work properly with pagination?  i would think it would make a lot of extra requests potentially if it worked at all for large repos.  Is it possible to pass media_type as an array or do some other type of filtering so that we can just make a single request?

app/services/katello/pulp3/docker_tag.rb
```diff
@@ -0,0 +1,28 @@
...
+      end
+
+      def update_model(model)
+        custom_json = {}
+        taggable = ::Katello::DockerManifest.find_by(:pulp_id => backend_data['tagged_manifest'])
+        if taggable == nil
+          taggable = ::Katello::DockerManifestList.find_by(:pulp_id => backend_data['tagged_manifest'])
+        end
+        custom_json['docker_taggable'] = taggable
+        custom_json['name'], = backend_data['name']
```
jlsherrill: stray comma?

app/services/katello/pulp3/docker_manifest.rb
ianballou: Filtering by both media types would definitely be preferable, I'm seeing if that's possible in Pulp 3.

ianballou: I've added a story for it here https://pulp.plan.io/issues/5157

app/services/katello/pulp3/repository/docker.rb
```diff
@@ -4,7 +4,7 @@ module Katello
   module Pulp3
     class Repository
       class Docker < ::Katello::Pulp3::Repository
-        def api_client
+        def self.api_client(smart_proxy)
           PulpDockerClient::ApiClient.new(smart_proxy.pulp3_configuration(PulpDockerClient::Configuration))
         end
```
sjha4: You may want to look at https://github.com/Katello/katello/pull/8224 and add a similar RepositorySyncURL class for docker content. Also, you can test progress presenter with https://github.com/Katello/katello/pull/8223 and see if it works with docker content.

jlsherrill:
reviewing this now, needs a rebase!

jlsherrill:
@ianballou still showing conflicting files

ianballou:
@jlsherrill we should be all good now.  I was trying to figure out when `pulp_database_unit.rb` changed and how it effected my code.

app/services/katello/pulp3/docker_manifest.rb
jlsherrill: Thought about this a bit more and i think its fine for now

jlsherrill: i opened https://projects.theforeman.org/issues/27425 to track it

ianballou: ok, and I think the solution will be the filtering that the Pulp team is adding.

jlsherrill:
i suspect you re-recorded these cassettes using some broken pulp client gems that had been released, you might try again

ianballou:
Cassettes are re recorded! Just need to see the tests pass again.

ianballou:
I've updated some tests, now everything should pass just fine.

jlsherrill:
thanks @ianballou

PR #8220, opened 2019-07-18
Fixes #27277 - Fix pulp2 file indexing

sjha4:
**To Test:**
**************
On a dev box with pulp2, i.e regular dev box:
Create a repo with content type as file, save and sync. 
(Ex URL to sync from: https://repos.fedorapeople.org/repos/pulp/pulp/fixtures/file2/)
Notice the 400 Bad request error in the indexing task.

This PR should fix that and you should be able to see the indexed file contents.


test/services/katello/pulp/rpm_test.rb
```diff
@@ -57,7 +57,7 @@ def teardown
       end
     end
 
-    class RpmTest < RpmTestBase
+    class RpmTest < RpmTestBaseRpm
       def test_find
         package = Rpm.find(@package_id)
```
sjha4: Renaming this base class name to see if this will cause failure that should happen with the line Katello::Rpm.import_for_repository(@repo, true)

sjha4:
Unrelated test errors. 
[test katello]

sjha4:
Will give this one more try..
[test katello]

tbrisker:
[test katello] (please ignore, checking jenkins)

jturel:
@sjha4 the sync task is a success but I synced the repo you linked and I see '0' for the files count. I think that indicates the indexing didn't import the data we expected since there are 3 files in the repo. Did you see the same or should I try something else on my end?

sjha4:
@jturel: If you go back to the repositories page, the content count is not refreshed immediately. You can refresh the page to see updated count.

jturel:
Strange. I gave it a re-test and then I saw the counts get updated. But as of this morning the old repo was still showing 0

jturel:
Can you leave some comments on the PR around the changes to some of the files under test/ ? I'm not quite following why some of the changes were made

test/services/katello/pulp/srpm_test.rb
```diff
@@ -33,46 +33,5 @@ def test_update_model_is_idempotent
...
-        refute_empty backend_rpm .provides
-      end
-
-      def test_ignored_fields
-        refute_includes Katello::Pulp::Rpm::PULP_SELECT_FIELDS, 'changelog'
-        refute_includes Katello::Pulp::Rpm::PULP_SELECT_FIELDS, 'repodata'
-        refute_includes Katello::Pulp::Rpm::PULP_SELECT_FIELDS, 'filelist'
-      end
-    end
   end
```
sjha4: This was duplicated in test/services/katello/pulp/rpm_test.rb and for some reason the tests in the original class were not running on Jenkins. There was a bug in the test https://github.com/Katello/katello/pull/8220/files#diff-20f25fb764ba969e5a677000d85050bbL50 but was never caught cause it wasn't being run. Instead the same tests were being run from here. There is no need to duplicate tests hence removing this and fixing the issue in the original class.

test/services/katello/pulp/repository/file_test.rb
```diff
@@ -10,10 +10,30 @@ class FileBaseTest < ::ActiveSupport::TestCase
         def setup
           @master = FactoryBot.create(:smart_proxy, :default_smart_proxy)
           @custom = katello_repositories(:generic_file)
+          @repo = katello_repositories(:pulp3_file_1)
+          RepositorySupport.create_and_sync_repo(@repo)
+        end
```
sjha4: We had a fixture for files but we weren't saving it to DB in earlier tests. Hence we missed the error in distribution path and the label containing a blank space. With this test we are saving,syncing and indexing in VCR test which highlighted these issues in our fixtures.

test/support/pulp/repository_support.rb
```diff
@@ -30,8 +30,8 @@ def self.create_and_sync_repo(repo)
     def self.create_repo(repo)
       FactoryBot.create(:smart_proxy, :default_smart_proxy) unless ::SmartProxy.pulp_master
 
-      repo.relative_path = repo.puppet? ? PULP_TMP_DIR : 'test_path'
-      repo.root.url = repo.puppet? ? @puppet_repo_url : @repo_url
+      repo.relative_path = (repo.puppet? ? PULP_TMP_DIR : 'test_path') unless repo.file?
+      repo.root.url = (repo.puppet? ? @puppet_repo_url : @repo_url) unless repo.file?
       repo.root.download_policy = :immediate if repo.yum?
```
sjha4: We specify relative path and sync url in file fixtures so we do not need to overwrite these for files. I am not sure if we have tests saving other types of repos but doesn't seem to affect them.

test/services/katello/pulp/repository/ostree_test.rb
```diff
@@ -22,7 +22,13 @@ def delete_repo(repo)
...
           service = Katello::Pulp::Repository::Ostree.new(@repo, @mirror)
-
+          Cert::Certs.stubs(:ueber_cert).returns({})
+          Katello::Pulp::Repository::Ostree.any_instance.stubs(:mirror_importer_connection_options).returns(
+                                                                                                         ssl_client_cert: nil,
+                                                                                                           ssl_client_key: nil,
+                                                                                                           ssl_ca_cert: nil,
+                                                                                                           ssl_validation: false
+          )
           assert_include service.generate_mirror_importer.feed, URI(SmartProxy.pulp_master.pulp_url).host
```
sjha4: This was needed cause with the existing VCR the tests seem to pass. But re-recording the cassette fails looking for the cert during the test. Stubbing it inside the method should make the re-recording more stable.

PR #8222, opened 2019-07-22
Fixes #27367 - custom repo default download policy immediate

jlsherrill:
This PR, separates the default download policy for custom repos and for redhat repos, and sets the default for redhat repos to 'on_demand', but custom repos to immediate.


ianballou:
Should we add some kind of warning for people who decide to choose on demand for their custom repos? I feel like people might still try to change it to on demand because it seems better on the surface.

jlsherrill:
@ianballou  added some to the help text

parthaa:
can you add a test if its not a lot of work..

jlsherrill:
@parthaa added

PR #8223, opened 2019-07-22
Fixes #27370 - Pulp3 Presenter

sjha4:
To test:

Create and sync repos in Pulp3
The progress presenter should look like screenshots below:
****
**1st time file sync:**
file_1
*******
**2nd time file sync** : Ignores tasks that dont run in pulp, i.e total is null
file_2
*****
**Ansible Sync**
ansible

### DIFF
--- a/app/lib/actions/pulp3/repository/presenters/content_unit_presenter.rb
+++ b/app/lib/actions/pulp3/repository/presenters/content_unit_presenter.rb
@@ -4,7 +4,7 @@ module Actions
       module Presenters
         class ContentUnitPresenter < AbstractSyncPresenter
           def progress
-            total_units == 0 ? 0.01 : finished_units.to_f / total_units
+            total_units == 0 ? 0.1 : finished_units.to_f / total_units
           end
 
           private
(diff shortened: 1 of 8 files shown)


app/lib/actions/pulp3/repository/presenters/content_unit_presenter.rb
```diff
@@ -12,7 +12,16 @@ def progress
           def humanized_details
             ret = []
             ret << _("Cancelled.") if cancelled?
-            ret << _("New Content Units: %s") % finished_units
+            ret << _("Content Units found: %s") % finished_units
```
sjha4: This line **"Content Units found:"** results in inconsistent data. We have different messages for all content types which can return this value. For example: In Ansible the message would be "Downloading Collections", in Files it would be "Parsing Metadata". 
We could possibly leave this out to make this generic enough or figure out a way to track different messages for different content type which might take some additional work. Looking for ideas on this one.

sjha4:
[test katello]

sjha4:
Based on some offline discussion with Justin, the following is the new screen:
File_progress test
Ansible_sync_progress

sjha4:
[test katello]

sjha4:
[test katello]

jlsherrill:
retested docker and it looks like the task status is looking like this:
```
  progress_reports:
  - message: Downloading tag list
    state: completed
    total: 1
    done: 1
  - message: Downloading Artifacts
    state: running
    done: 415
  - message: Associating Content
    state: running
    done: 641
  - message: Processing Tags
    state: completed
    total: 135
    done: 135
```

notice that downloading artifacts does not have a total, we should probably assume that total == done if total doesn't exist

sjha4:
@jlsherrill : I think we should raise this as a bug with the pulp team cause total and done are both expected fields for the progress report.

sjha4:
Pushed the change to account for missing total values for running tasks. Opened https://pulp.plan.io/issues/5193 for docker plugin to fix this in Pulp. The presenter should work normally without any changes when the fix is available.

PR #8224, opened 2019-07-22
Fixes #27373 - Pass mirror true for pulp3 repo syncs

sjha4:
**To Test**
*******
**Files:**
1. Create a file repo with pulp3 with upstream URL as https://repos.fedorapeople.org/repos/pulp/pulp/fixtures/file2/
2. Sync the repo and verify that 3 file units were indexed. (1, 2 and 3.iso 's)
3. Update the upstream URL to : https://repos.fedorapeople.org/pulp/pulp/demo_repos/test_file_repo/
4. Re-sync:
i) **With this PR**: There should be 3 new content units (test, test2, test3.iso 's)
ii) **Earlier**: There are now 6 content units for the repository which is not what we want.
******************
**Ansible Collections:**
1. Create a collection repo with pulp3 with upstream URL as https://galaxy.ansible.com and whitelist set to newswangerd.collection_demo
2. Sync the repo and verify that 1 collection unit was indexed.
3. Update the whitelist to devoperate.base
4. Re-sync:
i) **With this PR**: There should be 1 total content unit
ii) **Earlier**: There are now 2 collection units for the repository (1 from the old whitelist and 1 from the updated whitelist)

**To Do:**

- [x] Add tests around this since this part is not tested in files/ collections
- [x] Test this with the fix from https://pulp.plan.io/issues/5167


sjha4:
[test katello]

app/lib/actions/pulp3/repository/sync.rb
```diff
@@ -10,7 +10,9 @@ def plan(repo, smart_proxy, options = {})
 
         def invoke_external_task
           repo = ::Katello::Repository.find(input[:repo_id])
-          output[:pulp_tasks] = repo.backend_service(::SmartProxy.find(input[:smart_proxy_id])).sync
+          mirror_param = input.dig(:options, :options, :remove_missing)
+          mirror = mirror_param.nil? ? true : mirror_param
```
jlsherrill: why are we not looking on the repository to see if mirror_on_sync is set?

jlsherrill: also, this should be set in the importer, not at sync time: https://github.com/Katello/katello/blob/master/app/services/katello/pulp/repository/yum.rb#L8

app/services/katello/pulp3/repository.rb
```diff
@@ -151,8 +151,9 @@ def delete(href = repository_reference.try(:repository_href))
         end
       end
 
-      def sync
-        [remotes_api.sync(repo.remote_href, repository: repository_reference.repository_href)]
+      def sync(mirror = true)
+        repository_sync_url_data = client_class::RepositorySyncURL.new(repository: repository_reference.repository_href, mirror: mirror)
+        [remotes_api.sync(repo.remote_href, repository_sync_url_data)]
```
jlsherrill: similarly to pulp2, i'd expect this set in the remote instead of at sync time

app/lib/actions/pulp3/repository/sync.rb
sjha4: Updated this to use mirror_on_sync field and send it to remote with sync call.

app/services/katello/pulp3/repository.rb
sjha4: @jlsherrill : Based on our offline discussion and the fact that we don't have mirror option on remote in pulp3, I'll leave this as is if it's ok to pass this at time of sync.

jlsherrill: can you just just reference @repo.root.mirror_on_sync here instead of passing it in as a variable?

sjha4: Updated..

sjha4:
Need to re-record cassettes post https://github.com/Katello/katello/pull/8267..

sjha4:
Pushed updated tests and cassettes post https://github.com/Katello/katello/pull/8267.

app/services/katello/pulp3/repository.rb
```diff
@@ -276,11 +267,7 @@ def refresh_mirror_distributions(options = {})
       def refresh_distributions
         path = relative_path
         dist_ref = distribution_reference(path)
-        if dist_ref
-          update_distribution(path)
-        else
-          create_distribution(path)
-        end
+        dist_ref ? update_distribution(path) : create_distribution(path)
```
sjha4: Modified some code in the class to bring it under 379.. #rubocop

test/actions/pulp3/orchestration/file_sync_test.rb
```diff
@@ -40,5 +40,45 @@ def test_sync
...
+      ForemanTasks.sync_task(
+          ::Actions::Pulp3::Orchestration::Repository::Update,
+          @repo,
+          @master)
+
+      ForemanTasks.sync_task(::Actions::Pulp3::Orchestration::Repository::Sync, @repo, @master, sync_args)
+      @repo.reload
+      @repo.index_content
+      post_count_content = ::Katello::RepositoryFile.where(:repository_id => @repo.id).count
+      assert_equal pre_count_content + 3, post_count_content
```
jlsherrill: I would probably prefer we don't rely on hosted repos whenever possible.  For yum its relatively easy as we can use the fixtures in ./test/fixtures/test_repos/  and add more repos there.  I'd probably rather do that then rely on pulp ones.  I can help you generate a yum repo there if you've not done it before.

sjha4: @jlsherrill : Does this https://github.com/Katello/katello/pull/8224/commits/cfbf4af092f8ed74d76acd19c611b142d5aa3076 look ok?

jlsherrill: Yep! looks great!

sjha4:
[test katello]
random failure..Katello::ContentViewComponentTest

PR #8226, opened 2019-07-23
Fixes #27390 - fix repo-link in content credential

m-bucher:


app/views/katello/api/v2/content_credentials/show.json.rabl
```diff
@@ -23,7 +23,9 @@ child :root_repositories => :gpg_key_repos do
   attribute :id
   attribute :name
   attribute :content_type
-
+  node :library_instance do |root|
```
m-bucher: Not sure if it is better to use `node` here, because we only get one value
or to use
```ruby
child :library_instance do |_library_instance|
  attributes :id
end
```

jlsherrill: node is fine, but i'd probably rename it to library_instance_id since its just the id.   If we wanted to include more information we could use child and have multipel attributes, but that doesn't seem necessary here.

m-bucher: Sounds good, renamed to `library_instance_id`

PR #8052, opened 2019-03-28
Fixes #25113 - Refactor Org handling in React

johnpmitsch:
This commit refactors org handling in our React pages by updating the `withOrganization` component to handle the Organization logic and then passing the Organization to the wrapped component. This keeps the logic separate and always tracks the Organization in Foreman.

My understanding of how Foreman handles an organization is this:

- Foreman passes the current org title from [the rails side](https://github.com/theforeman/foreman/blob/84ae69c6bf191837e8c7b4ccadb4583eebeda172/app/helpers/layout_helper.rb#L39) to react when the components [are mounted](https://github.com/theforeman/foreman/blob/84ae69c6bf191837e8c7b4ccadb4583eebeda172/app/helpers/layout_helper.rb#L3)
- A currentOrganization prop is passed to Layout from Redux.
  - The layout reducer [starts with "Any Organization"](https://github.com/theforeman/foreman/blob/84ae69c6bf191837e8c7b4ccadb4583eebeda172/webpack/assets/javascripts/react_app/components/Layout/LayoutReducer.js#L21)
  - The [`changeOrganization`](https://github.com/theforeman/foreman/blob/84ae69c6bf191837e8c7b4ccadb4583eebeda172/webpack/assets/javascripts/react_app/components/Layout/LayoutActions.js#L40-L45) action can change this value in Redux
- Foreman checks if the `currentOrganization` is the same as what is being passed in from Rails [here](https://github.com/theforeman/foreman/blob/84ae69c6bf191837e8c7b4ccadb4583eebeda172/webpack/assets/javascripts/react_app/components/Layout/Layout.js#L49-L59) and updating if they are not the same.
- So as long as the `Layout` component is rendering, the Org in Redux always equals the org set in Rails

Then in Katello:

- The index file for the top-level `Application` Component [loads the organization](https://github.com/Katello/katello/blob/master/webpack/containers/Application/index.js#L22-L25)
  - This fetches the organization details [with an action](https://github.com/Katello/katello/blob/master/webpack/scenes/Organizations/OrganizationActions.js#L14-L35) and uses the [`orgId()` function](https://github.com/Katello/katello/blob/master/webpack/services/api/index.js#L116-L122) which pulls the organization from a data attribute on the html element itself
  - Then it is [saved by the reducer](https://github.com/Katello/katello/blob/master/webpack/scenes/Organizations/OrganizationReducer.js#L22)
[...]

PR #8022, opened 2019-03-14
Fixes #25113: ensure subscription pages have org.

waldenraines:
https://projects.theforeman.org/issues/25113

PR #8227, opened 2019-07-23
Fixes #25113: ensure subscription pages have org

jeremylenz:
Subscription pages that require an organization to be selected weren't
working properly when 'Any Organization' was selected in the top
toolbar.

* Use WithOrganization HOC to ensure org is selected when required
* Enclose the submit button in `<a>` tag (removes the need for
`redirectPath` because redirects are handled by the 302 response returned
to the browser)
* Update test & snapshot

https://projects.theforeman.org/issues/25113

### DIFF
--- a/webpack/components/SelectOrg/SetOrganization.js
+++ b/webpack/components/SelectOrg/SetOrganization.js
@@ -25,25 +25,15 @@ class SetOrganization extends Component {
 
   onSelectItem(e) {
     this.setState({
-      item: e.target.options[e.target.selectedIndex].text,
       id: e.target.value,
       disabled: false,
     });
   }
 
   onSend() {
-    const {
-      history,
-      changeCurrentOrganization,
-      redirectPath,
-    } = this.props;
-    const { id } = this.state;
-
-    changeCurrentOrganization(`${id}`).then(() =>
-      history.push({
-        pathname: redirectPath,
-        state: { orgChanged: this.state.item },
-      }));
+    this.setState({
+      disabled: true,
+    });
   }
 
   render() {
@@ -51,6 +41,8 @@ class SetOrganization extends Component {
       list,
       loading,
     } = this.props;
+
+    const { id } = this.state;
 
     return (
       <div id="select-org" className="well col-sm-6 col-sm-offset-3">
@@ -79,9 +71,11 @@ class SetOrganization extends Component {
               </div>
 
               <div className="col-sm-3">
-                <Button disabled={this.state.disabled} className="btn btn-primary" onClick={this.onSend}>
-                  {__('Select')}
-                </Button>
+                <a href={`/organizations/${id}/select`}>
+                  <Button disabled={this.state.disabled} className="btn btn-primary" onClick={this.onSend}>
+                    {__('Select')}
+                  </Button>
+                </a>
               </div>
             </div>
           </Form>
@@ -95,11 +89,9 @@ class SetOrganization extends Component {
 SetOrganization.propTypes = {
   list: PropTypes.arrayOf(PropTypes.object),
   loading: PropTypes.bool.isRequired,
-  changeCurrentOrganization: PropTypes.func.isRequired,
   history: PropTypes.shape({
     push: PropTypes.func,
   }).isRequired,
-  redirectPath: PropTypes.string.isRequired,
   getOrganiztionsList: PropTypes.func.isRequired,
 };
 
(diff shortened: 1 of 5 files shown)


jeremylenz:
@amirfefer take a look!  It's my first one.. be gentle ;)

waldenraines:
[test katello]

waldenraines:
I'll wait for @amirfefer and/or @johnpmitsch to have a look before I merge since they were also involved in this.

waldenraines:
[test katello]

waldenraines:
[test katello]

johnpmitsch:
Hey @jeremylenz,

I found this fixes the error from "Flow 2" on the issue. I still was able to reproduce "Flow 1". I'm fine with addressing these separately though.

I do still run into an issue with switching from "Any Org" where the org showing and the subscriptions don't match. Here's what I'm doing:
- Starting with "Another Org" on subscription page
- Clicking on a sub to view the details
- Selecting "Any Organization"
- Selecting "Default Organization"
- Clicking the subscriptions breadcrumb
- Subscriptions show from "Another Org", but "Default Organization" org shows in masthead (this is the issue)
- If I refresh the page, we get the expected results of Default Organization's subs showing.
Subscription_issues_july_2019

[In my own attempt](https://github.com/Katello/katello/pull/8052) to solve this and other organization handling issues, we ultimately arrived at the conclusion that the org handling needs to be done completely on the Foreman side and Katello needs a "hook" of some sort to base these changes off of, so this is the solution I think is best for the issue I mention above.

I want to make sure I understand the intention of the PR and if there is remaining work planned in this area as the issue is only partially fixed, but if this PR is meant to be a temporary fix, then it seems fine to me :+1:

jeremylenz:
@johnpmitsch you're right!  looks like I missed this when I was testing it. 

@waldenraines thoughts?  I think it's clear that org handling needs work as mentioned above, but think this fix is still worthwhile for now.

waldenraines:
> I want to make sure I understand the intention of the PR and if there is remaining work planned in this area as the issue is only partially fixed, but if this PR is meant to be a temporary fix, then it seems fine to me

Yeah, this PR is meant to be a better version of my [`$window.location.reload()` hack](https://github.com/Katello/katello/pull/8022).  

I still think that we should fix the aforementioned issue here if we can but I think we should timebox the investigation into fixing it (to maybe a day?) since we know we want to refactor this organization handing.

johnpmitsch:
> I still think that we should fix the aforementioned issue here if we can but I think we should timebox the investigation into fixing it (to maybe a day?)

+1 since it's a bit of a rabbit hole.

I'm fine with the change as is, but you might want to break the associated issue into multiple ones or make QE aware that this is a partial fix with some known caveats.

waldenraines:
> I'm fine with the change as is, but you might want to break the associated issue into multiple ones or make QE aware that this is a partial fix with some known caveats.

That's a good point.  @jeremylenz do you want to take a quick look at this and see if it's possible to fix it?  Maybe timebox it to a short period of time just to see if it's possible to fix quickly.  If not, we can open some other bugs to address these other issues.

jeremylenz:
@waldenraines Per our in-person discussion just now, it looks like this will take significant time to fix. 
 So we'd like to go ahead and merge this PR to address Flow 2, and leave Flow 1 as part of our more comprehensive look at how Katello deals with permissions.

waldenraines:
> So we'd like to go ahead and merge this PR to address Flow 2, and leave Flow 1 as part of our more comprehensive look at how Katello deals with permissions.

@johnpmitsch thoughts?  

Basically flow 1 requires us to be able to check to see if a user has permission to view the subscription and handle it accordingly.  Since we have no way in react of checking to see if the user has the proper permissions for a given subscription we'll have to fix this as part of a comprehensive permission effort as @jeremylenz mentions above.

waldenraines:
> * "Flow 1" from the redmine
> * The organization mismatch from [my comment](https://github.com/Katello/katello/pull/8227#issuecomment-514687258)

I believe that both of these will be addressed by the newly created issue [#27451](https://projects.theforeman.org/issues/27451) which suggests that we rewrite the organization switcher in foreman in react so that it stores the organization in the foreman redux store.  Once we do that we can fix katello to use that store and we'll only have one single source of truth for what organization is selected.

waldenraines:
I'm going to go ahead and merge this and call flow 2 fixed.  I have also updated the BZ with relevant information and linked a couple of related issues to [#25113](https://projects.theforeman.org/issues/25113).

waldenraines:
@jeremylenz congratulations on your first katello PR being merged!

johnpmitsch:
@waldenraines excellent, thanks for the update.

@jeremylenz congrats! :)

PR #8234, opened 2019-07-24
Fixes #26792: fix autocomplete search selection.

waldenraines:
Do not automatically select the first autocomplete result which will
prevent enter from being interpreted as "search for the first result".

https://projects.theforeman.org/issues/26792

### DIFF
--- a/engines/bastion/app/assets/javascripts/bastion/layouts/partials/table.html
+++ b/engines/bastion/app/assets/javascripts/bastion/layouts/partials/table.html
@@ -16,6 +16,7 @@
                    ng-trim="false"
                    stop-event="click"
                    uib-typeahead="item.label for item in table.autocomplete($viewValue)"
+                   typeahead-focus-first="false"
                    typeahead-template-url="components/views/autocomplete-scoped-search.html"
                    typeahead-min-length="0"/>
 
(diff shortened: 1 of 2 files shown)


waldenraines:
[test katello]

waldenraines:
> This change works perfectly in the "Packages" page. But it does not work as expected in the "Subscriptions" page. @waldenraines

Seems like I'll need to fix the react implementation as well.  Will do, thanks @sseelam2!

jlsherrill:
@waldenraines still planning on fixing the react implementation? or want to do that as part of another pr?

waldenraines:
> Seems like I'll need to fix the react implementation as well. Will do, thanks @sseelam2!

I was able to fix this on the katello react pages but only by making enter select the autocomplete item (rather than search for it).  Unfortunately I couldn't find a way for the library we're using in katello's react pages to not autoselect the first item.

Related to this is https://projects.theforeman.org/issues/27615 which is the technical debt of migrating katello's react search inputs to match foreman's.  This effort is too much scope for this bug fix so I just took the easy way out :smirk:

waldenraines:
@jlsherrill @sseelam2 this is ready for another review please.

sseelam2:
@waldenraines I retested this PR and it works as you mentioned. ACK :+1:

waldenraines:
Thanks @sseelam2!

PR #8099, opened 2019-05-02
Refs #26135 - upload/delete/view srpms on a repo

chris1984:

PR #8237, opened 2019-07-26
Fixes #26240 - Support uploading files to repo

sjha4:
**To test:**
**1st step:** (This is because we get an access error on /var/lib/pulp/uploads)
sudo chmod -R 777 /var/lib/pulp/

1. Create a file repo using pulp3
2. Go to repo and upload a file to the repository.

*****************************
**To Do:**
- [x] Make uploads more generic
- [x] Test upload for different extensions
- [x]  Tests

### DIFF
--- a/app/lib/actions/pulp3/repository/save_artifact.rb
+++ b/app/lib/actions/pulp3/repository/save_artifact.rb
@@ -2,17 +2,18 @@ module Actions
   module Pulp3
     module Repository
       class SaveArtifact < Pulp3::Abstract
-        def plan(file, repository, tasks, unit_type_id)
-          file_name = file[:filename]
-          plan_self(:file_name => file_name, :repository_id => repository.id, :tasks => tasks, :unit_type_id => unit_type_id)
+        def plan(file, repository, tasks, unit_type_id, options = {})
+          options[:file_name] = file[:filename]
+          plan_self(:repository_id => repository.id, :tasks => tasks, :unit_type_id => unit_type_id, :options => options)
         end
 
         def run
           artifact_href = input[:tasks].last[:created_resources].first
           content_type = input[:unit_type_id]
           content_backend_service = SmartProxy.pulp_master.content_service(content_type)
-          data = content_backend_service.content_class.new(relative_path: input[:file_name], _artifact: artifact_href)
-          content_unit = content_backend_service.content_api.create(data)
+          upload_options = {_artifact: artifact_href}.merge(input[:options])
+          content = content_backend_service.create_content(upload_options.with_indifferent_access)
+          content_unit = content_backend_service.content_api.create(content)
           output[:content_unit_href] = content_unit._href
         end
       end
--- a/app/lib/actions/pulp3/repository/save_artifact.rb
+++ b/app/lib/actions/pulp3/repository/save_artifact.rb
@@ -1,0 +1,21 @@
+module Actions
+  module Pulp3
+    module Repository
+      class SaveArtifact < Pulp3::Abstract
+        def plan(file, repository, tasks, unit_type_id)
+          file_name = file[:filename]
+          plan_self(:file_name => file_name, :repository_id => repository.id, :tasks => tasks, :unit_type_id => unit_type_id)
+        end
+
+        def run
+          artifact_href = input[:tasks].last[:created_resources].first
+          repo = ::Katello::Repository.find(input[:repository_id])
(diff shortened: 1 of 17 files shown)


sjha4:
@jlsherrill : This works for uploading from the UI. Planning to handle the cli uploads in a separate PR.

app/lib/actions/pulp3/repository/upload_file.rb
```diff
@@ -0,0 +1,84 @@
...
+          offset = 0
+          response = ""
+          File.open(input[:file], "rb") do |file|
+            total_size = File.size(file)
+            upload_data = upload_class.new({size: total_size})
+            response = uploads_api.create(upload_data)
+            upload_href = response._href
+            chunksize = upload_chunk_size
+            offset = 0
+            sha256 = Digest::SHA256.new
```
jlsherrill: would it be simpler to just do something like:
```
Digest::SHA256.file(filename)
```

instead of doing this append method?

app/services/katello/pulp3/repository.rb
```diff
@@ -101,6 +109,14 @@ def repository_versions_api
         PulpcoreClient::RepositoriesVersionsApi.new(core_api_client)
       end
 
+      def uploads_api
+        PulpcoreClient::UploadsApi.new(core_api_client)
+      end
```
jlsherrill: So today we can only upload 1 type of content per repository type (rpm for yum, file for file, etc...).  This pr: https://github.com/Katello/katello/pull/8099 however changes this assumption and allows the user to pass in what content_type they are wanting to upload.  

So we may want to make this uploads_api exist on the content type service classes themselves.

I also have some concerns that the api may be different across pulp plugins, but we'll just have to see how that goes as we do more :)

jlsherrill: and i would add, to do this properly, we probably need to pull in his work first, to get some of the changes needed in the repository type manager

sjha4: Pulled in Chris's PR and updated the flow accordingly. This should be ready for review.

sjha4:
[test katello]

sjha4:
[test katello]

app/lib/actions/pulp3/repository/save_artifact.rb
```diff
@@ -0,0 +1,21 @@
...
+          file_name = file[:filename]
+          plan_self(:file_name => file_name, :repository_id => repository.id, :tasks => tasks, :unit_type_id => unit_type_id)
+        end
+
+        def run
+          artifact_href = input[:tasks].last[:created_resources].first
+          content_type = input[:unit_type_id]
+          content_backend_service = SmartProxy.pulp_master.content_service(content_type)
+          data = content_backend_service.content_class.new(relative_path: input[:file_name], _artifact: artifact_href)
+          content_unit = content_backend_service.content_api.create(data)
```
jlsherrill: Is this relative_path a file specific thing?  I'm trying to understand why it would be needed when creating a content unit generically

sjha4: You are right..It is content specific..I see this on files and docker but not on collections..I will change this..

app/lib/actions/pulp3/repository/save_artifact.rb
```diff
@@ -2,17 +2,18 @@ module Actions
...
         def run
           artifact_href = input[:tasks].last[:created_resources].first
           content_type = input[:unit_type_id]
           content_backend_service = SmartProxy.pulp_master.content_service(content_type)
-          data = content_backend_service.content_class.new(relative_path: input[:file_name], _artifact: artifact_href)
-          content_unit = content_backend_service.content_api.create(data)
+          upload_options = {_artifact: artifact_href}.merge(input[:options])
+          content = content_backend_service.create_content(upload_options.with_indifferent_access)
+          content_unit = content_backend_service.content_api.create(content)
           output[:content_unit_href] = content_unit._href
```
sjha4: I was thinking we use the options hash to add whatever needs to be added and on the content specific class we can check if all required options for that content_type are not nil. I am not sure what options will be needed for uploading other content types right now or whether those options will need to be passed to this action. Maybe we can look at that when supporting upload for different content types.

jlsherrill: yeah, i think thats fine for now

PR #7969, opened 2019-02-06
Fixes #25034 - No Duplicate repos on LCE UI

parthaa:
Click on a Lifecycle Env Details page. Notice that there is a "All
Content View" selction option and duplicate repository names displayed
in the "repositories"  section. The duplication is because the same
repository is in multiple content views.
This commmit tries to address this issue by removing the "All Content
Views" option which is probably never used. Instead it auto selects the
first available content view and displays repositories belonging to
that.

PR #8238, opened 2019-07-26
Fixes #27452 - Autoload LCE Content Views table

jturel:
https://github.com/Katello/katello/pull/7969 took steps to remove duplicate repos from showing under the various tabs when looking at the detail view of a Lifecycle Environment. It unintentionally stopped the Content Views tab from loading data in its table. This change restores that functionality.

**To test**

- Promote a Content View with some repos to any Lifecycle Environment
- Click Content -> Lifecycle Environments and click on the name of the destination environment
- Observe that under the Content Views tab the one promoted in the first step is shown. This was previously not working.

This change **does not** affect the intended behavior introduced by #7969 in that the other tabs require a CV to be selected before showing any data in the table.

### DIFF
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/environments/content.service.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/environments/content.service.js
@@ -28,7 +28,8 @@
                 state: 'content-views',
                 resource: 'ContentView',
                 display: translate('Content Views'),
-                params: {nondefault: true}
+                params: {nondefault: true},
+                autoLoad: true
             }, {
                 state: 'repositories',
                 resource: 'Repository',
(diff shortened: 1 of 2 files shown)


engines/bastion_katello/app/assets/javascripts/bastion_katello/environments/content.service.js
```diff
@@ -87,11 +88,14 @@
         };
 
         this.buildNutupane = function (params) {
-            var nutupane;
+            var nutupane, type, disableAutoLoad;
+
+            type = this.getCurrentContentType();
+            disableAutoLoad = type.autoLoad !== true;
```
johnpmitsch: Any reason to not just use `!type.autoload` below?

johnpmitsch: though I can see how assigning it to a variable is more readable and explicit

jturel: Good suggestion - I like it better! Updated

PR #8240, opened 2019-07-30
Fixes #27479 - Republish migrated deb repos on upgrade

quba42:
Some future Katello version (as yet unknown?) will include pulp version `2.20.0`.
Pulp version `2.20.0` includes pulp_deb `1.10.0`.
In pulp_deb `1.10.0` we fixed a bug (https://pulp.plan.io/issues/4138) that necessitates a best effort db migration.

This migration is capable of fixing the bug for affected Katello repos. (CV version repos, LCENV repos, katello root repos, etc.)
However, once these repos have been migrated in the database, they should be republished ("republish repository metadata") for the bug fix to take full effect.
To facilitate this the migration records all affected repos in `/var/lib/pulp/0004_deb_repo_republish_candidates.txt`.

The rake task added by this PR will automatically trigger the needed publish tasks during upgrade.
Since I do not yet know what Katello version will include Pulp `2.20.0`, I have assumed Katello `3.14` for now, and made this a "Draft pull request".

### DIFF
--- a/db/seeds.d/111-upgrade_tasks.rb
+++ b/db/seeds.d/111-upgrade_tasks.rb
@@ -9,6 +9,7 @@ UpgradeTask.define_tasks(:katello) do
     {:name => 'katello:upgrades:3.11:import_yum_metadata'},
     {:name => 'katello:upgrades:3.11:update_puppet_repos'},
     {:name => 'katello:upgrades:3.11:clear_checksum_type', :task_name => 'katello:upgrades:3.8:clear_checksum_type'},
-    {:name => 'katello:upgrades:3.12:remove_pulp2_notifier'}
+    {:name => 'katello:upgrades:3.12:remove_pulp2_notifier'},
+    {:name => 'katello:upgrades:3.13:republish_deb_metadata'}
   ]
 end
--- a/lib/katello/tasks/upgrades/3.13/republish_deb_metadata.rake
+++ b/lib/katello/tasks/upgrades/3.13/republish_deb_metadata.rake
@@ -1,0 +1,22 @@
+namespace :katello do
+  namespace :upgrades do
+    namespace '3.13' do
+      desc "Regenerate the metadata for repositories affected by pulp_deb migration 0004"
+      task :republish_deb_metadata, [:input_file] => ["environment"] do |task, args|
+        User.current = User.anonymous_api_admin
+        input_file = args[:input_file] || "/var/lib/pulp/0004_deb_repo_republish_candidates.txt"
+        if File.readable?(input_file)
+          pulp_ids = File.read(input_file).each_line.map(&:strip) || []
+          repos = Katello::Repository.where(:pulp_id => pulp_ids)
+          puts _("Starting BulkMetadataGenerate task.")
+          task = ForemanTasks.async_task(Actions::Katello::Repository::BulkMetadataGenerate, repos, :force => true)
+          puts _("Please check that the task #{task.id} completes successfully.")
+          puts _('You can manually re-trigger this task by running "foreman-rake katello:upgrades:3.13:republish_deb_metadata"')
+        else
+          puts _("Input file #{input_file} was not readable.")
+          puts _('You can manually use an alternate input file by running "foreman-rake katello:upgrades:3.13:republish_deb_metadata[<path>]"')
+        end
+      end
+    end
+  end
+end


lib/katello/tasks/upgrades/3.14/republish_deb_metadata.rake
```diff
@@ -0,0 +1,24 @@
+namespace :katello do
+  namespace :upgrades do
+    namespace '3.14' do
+      desc "Regenerate the metadata for repositories affected by pulp_deb migration 0004"
+      task :republish_deb_metadata => ["environment"] do
+        User.current = User.anonymous_api_admin
+        input_file_name = "/var/lib/pulp/0004_deb_repo_republish_candidates.txt"
```
ekohl: I would strongly recommend to avoid relying on a shared filesystem. This will break on split deployments. Those are currently rare, but they are high on my wishlist. It will also make Katello much more container friendly.

lib/katello/tasks/upgrades/3.14/republish_deb_metadata.rake
```diff
@@ -0,0 +1,24 @@
...
+        if File.file?(input_file_name)
+          pulp_ids = []
+          File.open(input_file_name, "r") do |file_handle|
+            file_handle.each_line do |pulp_id|
```
ekohl: Isn't this equivalent?
```ruby
pulp_ids = file_handle.each_line.map(&:strip)
```

m-bucher: Or even shorter :smile::
```ruby
pulp_ids = File.read(input_file_name).each_line.map(&:strip)
```

lib/katello/tasks/upgrades/3.14/republish_deb_metadata.rake
quba42: I will keep that in mind for the future, but I don't think it is worth the effort to change this for the current case. The existing db migration on the pulp side is pretty much set in stone, (though it might be possible to parse the same information from the existing logs as well).

This rake task is essentially a convenience improvement for a presumably very small number of Katello deb content users. If the `/var/lib/pulp/0004_deb_repo_republish_candidates.txt` file is missing on the disk the task just won't do anything (which is desirable for all the Katello users who don't have any deb content).

Everything fixed by this task can also be achieved by manual (though cumbersome) action. The main thing that won't work without running the rake task, is synchronizing deb content to smart proxy (which has never worked in the past).

ekohl: Can you help me understand what's in this file? Is it all Debian repositories or a subset? Would be ok to query the Pulp server for all Deb repos and process them all?

quba42: Sure, I feel like I should have anticipated this question and answered it preemptively.

The file contains a list of pulp_ids (one per line) of every repository that had to be migrated by pulp_deb migration 0004.

This will be any Debian or Ubuntu repositories affected by a certain bug in pulp_deb. Known examples include all releases from the `debian-security` repository, as well as any Ubuntu releases that have a `-` in their name (e.g.: `bionic-backports`, `bionic-updates`, etc.)

quba42: Since "republishing metadata" for deb type repositories, takes some amount of time (bearable but not super efficient), I would not recommend republishing all deb type repositories instead. (Though that would also solve the problem).

lib/katello/tasks/upgrades/3.14/republish_deb_metadata.rake
quba42: I tried to use @m-bucher's version, but I have yet to test it.

lib/katello/tasks/upgrades/3.14/republish_deb_metadata.rake
ekohl: So it's not all repositories but a subset. It sounds like this can be ok, but I would make sure it also works when the file exists but is not readable (because of permissions for example).

Another option to consider is making the path configurable (with its current value as the default). This would allow users of a split deployment to copy the file and manually run the task. https://www.viget.com/articles/protip-passing-parameters-to-your-rake-tasks/ has some examples.

jlsherrill:
pulp 2.20 should be shipped with katello 3.13, mind updating the versioning to that?

ekohl:
> pulp 2.20 should be shipped with katello 3.13, mind updating the versioning to that?

Could you tell the release team that? Until now all our systems have been using 2.19 (repoclosure was even using 2.18).

jlsherrill:
@ekohl i'm working on doing the prs now :)

quba42:
> pulp 2.20 should be shipped with katello 3.13, mind updating the versioning to that?

Done.

lib/katello/tasks/upgrades/3.14/republish_deb_metadata.rake
quba42: I replaced `File.file?` with `File.readable?`. I may have another look at the parameters later.

lib/katello/tasks/upgrades/3.14/republish_deb_metadata.rake
quba42: I tested and it worked.

lib/katello/tasks/upgrades/3.14/republish_deb_metadata.rake
quba42: I now added the ability to manually rerun the rake task using a different input file (supplied via a parameter). I also adjusted the output to reflect this.

Can I resolve this discussion?

ekohl: Yes, this addresses my concerns. Thanks!

jlsherrill:
[test katello]

jlsherrill:
rubocop issue:

```
[2019-08-02T13:59:03.448Z] /home/jenkins/workspace/katello-pr-test/db/seeds.d/111-upgrade_tasks.rb:13:5: E: Lint/Syntax: unexpected token tLBRACE

[2019-08-02T13:59:03.448Z] (Using Ruby 2.3 parser; configure using TargetRubyVersion parameter, under AllCops)

[2019-08-02T13:59:03.448Z]     {:name => 'katello:upgrades:3.13:republish_deb_metadata'}

[2019-08-02T13:59:03.448Z]     ^

[2019-08-02T13:59:03.448Z] /home/jenkins/workspace/katello-pr-test/db/seeds.d/111-upgrade_tasks.rb:14:3: E: Lint/Syntax: unexpected token tRBRACK

[2019-08-02T13:59:03.448Z] (Using Ruby 2.3 parser; configure using TargetRubyVersion parameter, under AllCops)

[2019-08-02T13:59:03.448Z]   ]

[2019-08-02T13:59:03.448Z]   ^
```

quba42:
Ugh! I forgot a comma. Hurray for this: (https://community.theforeman.org/t/new-trailing-comma-style-in-foreman-git/14551)

evgeni:
[test katello]

quba42:
So, Jenkins says: "Avoid comma after the last item of an array."
Accoding to this (https://community.theforeman.org/t/new-trailing-comma-style-in-foreman-git/14551) community discussion we "now require trailing commas".

Which is it? Do we now force different conventions within Katello and Foreman, or am I just missing something?

jlsherrill:
katello uses its own rubocop configuration which is somewhat different than foremans.  We hadn't really discussed the trailing comma after the foreman repo made their change.  So right now it looks like we require no trailing comma

quba42:
I fixed the rubocop issue. I am hoping the Katello tests will come back green now...

jlsherrill:
[test katello]

jlsherrill:
Thanks @quba42 !

PR #8242, opened 2019-07-30
Fixes #27210 - support more generic remove on pulp3

jlsherrill:
this adds support for more generic content
removal from a repo with pulp3


sjha4:
I do need to test this for docker content..Didn't realize this was to address that as well..

sjha4:
Tested..Works well for pulp3 Docker..ReACK! :+1:

PR #8244, opened 2019-07-31
Fixes #27485 - encrypt discovery password within task

jlsherrill:


jlsherrill:
By default dev envs do not have an encryption key setup, so if you run: 
 rake  security:generate_encryption_key

you can then examine the task and see the password encrypted

jturel:
[test katello]

test/actions/katello/repository/discover_test.rb
```diff
@@ -0,0 +1,43 @@
...
+    include FactoryBot::Syntax::Methods
+
+    let(:action_class) { ::Actions::Katello::Repository::Discover }
+
+    def setup
+      get_organization #ensure we have an org label
+    end
+
+    def test_discovers_without_encryption
+      EncryptionKey.const_set(:ENCRYPTION_KEY, nil)
```
jturel: I didn't have an encryption key in my dev env so this broke.

Maybe better to mock it out completely:

```
module ::EncryptionKey                                                                                                                                                         
  ENCRYPTION_KEY = nil                                                                                                      
end
```

similar for the following test

app/lib/actions/katello/repository/discover.rb
```diff
@@ -1,8 +1,9 @@
 module Actions
   module Katello
     module Repository
-      class Discover < Actions::Base
+      class Discover < Actions::EntryAction
```
jturel: What'd this do?

jturel:
works well!

```
---
url: REDACTED
content_type: yum
upstream_username: admin
upstream_password: encrypted-OEluTnN2K2ozTnRlM1NjRFVkVFljOFlYVHpTd0JpMUhBaFNCWFhZcmNxdz0tLWxRdUJzeitIQzVKZU9rMDdFWVJJYWc9PQ==--9866752dc75dca63893aa92b4964a0ed476c3d03
search: "*"
locale: en
current_request_id: 
current_timezone: UTC
current_user_id: 4
current_organization_id: 2
current_location_id: 1
```

Left some notes on testing

app/lib/actions/katello/repository/discover.rb
jlsherrill: This is a class that extends Actions::Base, and really has no affect here, but in general any action we call from controllers should extend it instead of Base.  It allows for things like locking and has a few other features

test/actions/katello/repository/discover_test.rb
jlsherrill: Thanks, updated!

PR #8248, opened 2019-08-05
Fixes #27497 - Support dmi.system.uuid override

jturel:
In some circumstances content hosts can report the same `dmi.system.uuid` fact as other hosts. This fact needs to be unique and this PR provides a way to work around the problem as a stopgap until the problem is (ideally) fixed in the infrastructure.

Also in this PR is a relocation and minor refactor of the registration bits in `SubscriptionFacet`. I've moved most of the code to `RegistrationManager` and split up the methods slightly to reflect the logical parts of the registration which makes thinking about, reading, and testing the code much easier. Also removed some dead code.

**How does it work?**
A new setting is provided which allows users to input the DMI UUID which is being reported by multiple systems. Doing so will have the following effects:
- During registration a random UUID will be generated and its value sent to Candlepin. It will also be stored as a host parameter for reference during checkin
- During checkin, if the host is reporting one of the duplicate values, the value is taken from the host parameter instead of what was already in the facts. If no host parameter exists (accidental deletion, whatever) a new one is generated and stored as a host parameter.

**To test**
- Register a content host (ex katello-client from forklift)
- Run `subscription-manager clean` on the client so it believes it's unregistered but preserves its data on the server
- On the same client, change its hostname so that registering will attempt to create a new profile.
- Attempt to register (`subscription-manager register`) it should fail with an error

Now, we can work around this problem.

- Take the dmi.system.uuid from the client: `subscription-manager facts | grep dmi.system.uuid`
- In the UI, go to Administer -> Settings -> Content
- Find the new setting 'Host Duplicate DMI UUIDs' and input the UUID you found in array format
- Registering the system should work, and it should have a new DMI UUID on the server - not on the client

You can verify that with some queries:

```
::Host.find(197).fact_values.where(fact_name_id: Katello::RhsmFactName.find_by_name('dmi::system::uuid'))

::Host.find(197).subscription_facet.candlepin_consumer.consumer_attributes[:facts]['dmi.system.uuid']

::Host.find(197).parameters

```

The UUIDs should all match.

- From the client, refresh the facts `subscription-manager facts --update`
- The UUID on the server side should be unchanged
- Deleting the host parameter and refreshing facts again should generate and store and new UUID


app/services/katello/candlepin/consumer.rb
```diff
@@ -44,10 +44,6 @@ def entitlement_status
         consumer_attributes[:entitlementStatus]
       end
 
-      def update(params)
```
jturel: Dead code

jturel:
@jlsherrill I updated to use the setting like you suggested yesterday - so now this is randomly generating the DMI UUID in those cases. Not sure what to do about system checkins when the DMI UUID has been overridden. Some options:

- also randomize DMI UUID on checkin when using the setting
- store it in our DB and override the value received in facts
- as part of the usage of the setting, require the user to override the fact on the client in /etc/rhsm/ to match what was randomly generated (or use their own)

jlsherrill:
I could see a 4th option:

* if a host is checking in (not registering), just pull the previously randomized UUID from the existing facts?

johnpmitsch:
This works well!

I have a couple questions:

- How does one know to use this setting? Going through the workflow it wasn't very obvious without previous knowledge of the setting. Ideally, subscription-manager would say something about it, but if that isn't realistic, we could show a message in our logs.
- Is there any concern about the uuid on the client and the Katello server not matching? I see it is an override so it seems safe but not sure if there are any larger implications.

I know there is some backstory so bear with me as I'm just familiarizing myself with everything :bear: 
 Code and test coverage looks good :+1:

jturel:
> * How does one know to use this setting? Going through the workflow it wasn't very obvious without previous knowledge of the setting. Ideally, subscription-manager would say something about it, but if that isn't realistic, we could show a message in our logs.

I'll need to think about this more, but it's a very good question. I want this to be hidden enough that users don't start using it from some unknown reason / misunderstanding on their part, but also provide guidance when this problem is encountered (it should be really rare). First thought is having a distinct error message that would be searchable and turn up a kbase article/ Logging too, like you mentioned.

> * Is there any concern about the uuid on the client and the Katello server not matching? I see it is an override so it seems safe but not sure if there are any larger implications.

While I'd prefer them to match, I don't think it's critical that they do as long as the server is sufficiently tricked. The client itself doesn't do anything based on the dmi uuid - it just reports it (to my knowledge). If there is a kbase for this we could also instruct the user to override the DMI UUID on the client so that it matches the overridden value on the server if they want

bbuckingham:
> > * How does one know to use this setting? Going through the workflow it wasn't very obvious without previous knowledge of the setting. Ideally, subscription-manager would say something about it, but if that isn't realistic, we could show a message in our logs.
> 
> I'll need to think about this more, but it's a very good question. I want this to be hidden enough that users don't start using it from some unknown reason / misunderstanding on their part, but also provide guidance when this problem is encountered (it should be really rare). First thought is having a distinct error message that would be searchable and turn up a kbase article/ Logging too, like you mentioned.
> 

This may not be ideal or a good option, but mentioning in case it would be desirable.  Today, in the UI (Hosts -> Content Hosts) we have a page for 'Register a Host'.  We could add a section on advanced configuration or debugging/troubleshooting to it.

jturel:
[test katello]

jturel:
> > > * How does one know to use this setting? Going through the workflow it wasn't very obvious without previous knowledge of the setting. Ideally, subscription-manager would say something about it, but if that isn't realistic, we could show a message in our logs.
> > 
> > 
> > I'll need to think about this more, but it's a very good question. I want this to be hidden enough that users don't start using it from some unknown reason / misunderstanding on their part, but also provide guidance when this problem is encountered (it should be really rare). First thought is having a distinct error message that would be searchable and turn up a kbase article/ Logging too, like you mentioned.
> 
> This may not be ideal or a good option, but mentioning in case it would be desirable. Today, in the UI (Hosts -> Content Hosts) we have a page for 'Register a Host'. We could add a section on advanced configuration or debugging/troubleshooting to it.

Not a bad idea - but I think I prefer just doing something in our usual upstream docs at theforeman.org so it's easy to link to and make updates to

jturel:
@johnpmitsch @jjeffers I pushed an update to improve the error messaging by indicating what the problematic DMI UUID of the host is.

Also opened a docs PR to instruct how to use the two available methods to work around: https://github.com/theforeman/theforeman.org/pull/1441

I think this should be good enough - to have something to point folks running into this (hopefully extremely rare) problem toward. Let me know what you think!

PR #8249, opened 2019-08-05
Refs #26135 - Update srpm rabl for hammer listing/info

chris1984:
Without this, the listing command is incomplete, also removed sourcerpm since it does not relate to srpms, this was copied from the rpm rabl. I moved filename to the top since it was not seeing it in the index

### DIFF
--- a/app/views/katello/api/v2/srpms/show.json.rabl
+++ b/app/views/katello/api/v2/srpms/show.json.rabl
@@ -1,5 +1,5 @@
 object @resource
 
-attributes :id, :pulp_id
+attributes :id, :pulp_id, :name, :filename
 attributes :created_at, :updated, :version, :arch, :release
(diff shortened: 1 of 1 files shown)


chris1984:
Without this, the listing command is incomplete, also removed sourcerpm since it does not relate to srpms, this was copied from the rpm rabl.

chris1984:
@jlsherrill when you get time, this was the rabl change I was talking about

chris1984:
Before pr:
```bash
--
ID
--
1 
--
```
With pr:
```bash
---|-----------|------------------------------
ID | NAME      | FILENAME                     
---|-----------|------------------------------
1  | garmindev | garmindev-0.3.4-9.el7.src.rpm
---|-----------|------------------------------
```

jlsherrill:
Thanks @chris1984

PR #8250, opened 2019-08-05
Fixes #27516 - Create os without spaces

m-bucher:
This should ensure we only use the first part of Ubuntu release code-names (e.g. 'bionic' instead of 'bionic beaver'). Debian will not be affected, because they only use one word as their code-names.

Not sure if this is the only section that has to be fixed.

### DIFF
--- a/app/models/katello/rhsm_fact_parser.rb
+++ b/app/models/katello/rhsm_fact_parser.rb
@@ -61,7 +61,7 @@ module Katello
 
     def os_release_name(os_name)
       if os_name.match(::Operatingsystem::FAMILIES['Debian'])
-        facts['distribution.id']
+        facts['distribution.id']&.split&.first&.downcase
       end
     end
 
--- a/test/models/rhsm_fact_parser_test.rb
+++ b/test/models/rhsm_fact_parser_test.rb
@@ -77,12 +77,14 @@ module Katello
 
     def test_operatingsystem_ubuntu
       @facts['distribution.name'] = 'Ubuntu GNU/Linux'
-      @facts['distribution.version'] = '19'
-      @facts['distribution.id'] = 'disco'
+      @facts['distribution.version'] = '19.04'
+      @facts['distribution.id'] = 'Disco Dingo'
 
       assert_equal parser.operatingsystem.release_name, 'disco'
       assert_equal parser.operatingsystem.name, 'Ubuntu'
       assert_equal parser.operatingsystem.type, 'Debian'
+      assert_equal parser.operatingsystem.major, '19'
+      assert_equal parser.operatingsystem.minor, '04'
     end
 
     def test_uname_architecture


app/models/katello/rhsm_fact_parser.rb
```diff
@@ -61,7 +61,7 @@ def operatingsystem
 
     def os_release_name(os_name)
       if os_name.match(::Operatingsystem::FAMILIES['Debian'])
-        facts['distribution.id']
+        facts['distribution.id']&.split(' ')&.first&.downcase
```
jturel: split(' ') is the same as .split without any args

jturel:
@m-bucher hey, what's the consequence of not doing this?

m-bucher:
@jturel ,for instance, rendering of templates fails. This is mostly a direct result of using a string containing a whitespace as release-name in the Operatingsystem.
However, Katello will enforce doing this by either not finding an existing Operatingsystem that uses the correct release-name or creating a new Operatingsystem with the wrong one:
```
Error rendering the Preseed finish template
 URI::InvalidURIError: bad URI(is not URI?): http://archive.ubuntu.com/ubuntu/dists/bionic beaver/main/installer-x86_64/current/images/netboot/ubuntu-installer/x86_64
 /home/vagrant/.rvm/rubies/ruby-2.5.1/lib/ruby/2.5.0/uri/rfc3986_parser.rb:67:in `split'
 /home/vagrant/.rvm/rubies/ruby-2.5.1/lib/ruby/2.5.0/uri/rfc3986_parser.rb:73:in `parse'
 /home/vagrant/.rvm/rubies/ruby-2.5.1/lib/ruby/2.5.0/uri/common.rb:237:in `parse'
 /home/vagrant/foreman/app/services/medium_providers/default.rb:42:in `medium_vars_to_uri'
 /home/vagrant/foreman/app/services/medium_providers/default.rb:20:in `medium_uri'
 /home/vagrant/foreman/app/services/medium_providers/default.rb:29:in `unique_id'
 /home/vagrant/foreman/app/models/operatingsystem.rb:202:in `pxe_prefix'
```
As already mentioned, a better solution is probably for subscription-manager to send the correct facts in the first place.
I vote for adding this nevertheless :wink:

app/models/katello/rhsm_fact_parser.rb
m-bucher: agreed, looks nicer :+1:

jturel:
@m-bucher that makes sense - thank you. could I bother you to add a test case?

m-bucher:
@jturel , sure thing.
I guess it's OK that I modified the existing one to reflect the facts that will come from the subscription-manager.

I also took the liberty to add the version-parsing to the test, which is also rather special on Ubuntu.
I am still not sure if splitting up the Ubuntu-"version" is correct, because the actual major version is "19.04".
Using the "19" as major would mean that the next major release in 2019 ("19.10") will be regarded as a minor release. Even though the release-name will change to "Eoan Ermine".

Maybe I am overthinking this :wink:

jturel:
> @jturel , sure thing.
> I guess it's OK that I modified the existing one to reflect the facts that will come from the subscription-manager.
> 

Perfectly fine!

> I also took the liberty to add the version-parsing to the test, which is also rather special on Ubuntu.
> I am still not sure if splitting up the Ubuntu-"version" is correct, because the actual major version is "19.04".

Thanks for adding coverage :)

> Using the "19" as major would mean that the next major release in 2019 ("19.10") will be regarded as a minor release. Even though the release-name will change to "Eoan Ermine".
> 
> Maybe I am overthinking this

I think the way it is maps to how we think about major and minor versions for an operating system, so it's OK as far as I can tell. Also, I'm not aware of it causing any problems. Of course we can always revisit.

PR #8251, opened 2019-08-05
Fixes #27518 - proxy options for sync

jjeffers:
This change fixes a problem where pulp2 requires the proxy_host value in an importer config for to use the format of "scheme://host", and not just "host".

### DIFF
--- a/app/services/katello/pulp/repository.rb
+++ b/app/services/katello/pulp/repository.rb
@@ -264,7 +264,7 @@ module Katello
         if proxy
           uri = URI(proxy.url)
           proxy_options = {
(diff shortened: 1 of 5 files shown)


app/services/katello/pulp/repository.rb
```diff
@@ -264,7 +264,7 @@ def proxy_options
         if proxy
           uri = URI(proxy.url)
           proxy_options = {
-            proxy_host: uri.host,
+            proxy_host: uri.scheme + '://' + uri.host,
```
jlsherrill: i'd probably default to an http scheme if its nil

PR #8258, opened 2019-08-06
Fixes #27426: katello sends invalid page number request

jeremylenz:
When you have no Red Hat Repositories enabled and try to enable one, the 
newly enabled repository would not show until after a page refresh.  
This was due to an invalid page number (0) in the request


waldenraines:
[test katello]

PR #8259, opened 2019-08-06
Fixes #27524 - fix docker cv publishing

ianballou:
To reproduce issue:
1) Create a custom docker repo with tags (fedora/ssh perhaps)
2) Sync it
3) Add it to a content view
4) Try to publish new version
5) It will fail

To test this patch:
1) Try the above steps
2) Check that there are no errors
3) Check that docker tags were indeed copied


test/fixtures/models/katello_root_repositories.yml
```diff
@@ -150,6 +150,17 @@ busybox2_root:
...
+ssh_root:
+  name:                 ssh
+  content_id:           1
+  content_type:         docker
+  label:                ssh
+  product_id:           <%= ActiveRecord::FixtureSet.identify(:puppet_product) %>
+  gpg_key_id:           <%= ActiveRecord::FixtureSet.identify(:fedora_gpg_key) %>
+  unprotected:           <%= true %>
+  url:                  'https://registry-1.docker.io'
+  docker_upstream_name: 'fedora/ssh'
```
jlsherrill: lets not add another docker repo to fixtures.  Lets try to adapt one that already exists if what we have isn't sufficient.

test/services/katello/pulp/repository/docker_test.rb
```diff
@@ -42,6 +44,19 @@ def test_create
...
+          service = Katello::Pulp::Repository::Docker.new(@ssh, @master)
+          service.create
+          RepositorySupport.create_and_sync_repo(@ssh)
+
+          @ssh.index_content
+          assert_equal @ssh.docker_tags.count, 2
+        ensure
+          delete_repo(@ssh)
+        end
+
```
jlsherrill: I'm not sure this test actually reproduces the problem?  I think you'd have to then create some content_view repo as well, and then index, passing in the source repo.  Although there'd be nothing inthe content view repo.  Maybe you could copy units over to the cv repo prior

ianballou: Ok, I'll bolster this a bit

ianballou:
Looks like I need to re record some VCRs before getting all the tests to pass.

chris1984:
Works :)

Before patch:

```ruby
2019-08-08T13:56:44 [E|bac|] PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_katello_docker_tags_on_pulp_id"
13:56:44 rails.1   |  | DETAIL:  Key (pulp_id)=(5b3b0767-4e0b-4029-b7c7-da1a618aade7) already exists.
13:56:44 rails.1   |  | : insert into katello_docker_tags (repository_id, pulp_id, name,docker_taggable_id,docker_taggable_type)
13:56:44 rails.1   |  |                     select 12 as repository_id, pulp_id, name,docker_taggable_id,docker_taggable_type from katello_docker_tags
13:56:44 rails.1   |  |                     where repository_id = 11 and pulp_id not in (select pulp_id
13:56:44 rails.1   |  |                     from katello_docker_tags where repository_id = 12) (ActiveRecord::RecordNotUnique)
```
https://nimb.ws/RlaHdL

After patch:

https://nimb.ws/w5Jx2y

test/services/katello/pulp/repository/docker_test.rb
ianballou: I've updated the test.

ianballou:
rebased and now I'm getting some Pulp3 Ansible errors, hm...

sjha4:
[test katello]

test/services/katello/pulp/repository/docker_test.rb
```diff
@@ -42,6 +44,26 @@ def test_create
...
+          service2 = Katello::Pulp::Repository::Docker.new(@repo_copy, @master)
+          service2.create
+          RepositorySupport.create_and_sync_repo(@repo)
+          RepositorySupport.create_repo(@repo_copy)
+
+          @repo.index_content
+          assert_equal @repo.docker_tags.count, 248
+          service.copy_contents(@repo_copy)
+          @repo_copy.index_content(:source_repository => @repo)
+          assert_equal @repo_copy.docker_tags.count, 248
```
jlsherrill: we can't rely on a specific count, as the next person that runs the test in live mode may get a different result.  Something like 

assert  @repo.docker_tags.count > 0
assert_equal @repo.docker_tags.count, @repo_copy.docker_tags.count

should be sufficient.   Also  i think this test needs to be re-recorded after rebasing against master to get my vcr fixes that avoid duplicate recordings

ianballou: Rebased and test fixed

PR #8260, opened 2019-08-06
Fixes #27530 - allow service ping regardless of scope

jlsherrill:


chris1984:
Starting to review

chris1984:
Works :)

Before patch:

https://nimb.ws/a4U1cJ

After patch:

https://nimb.ws/BqlPWs

PR #8261, opened 2019-08-07
Fixes #27418 - Support docker pull with pulp3

johnpmitsch:
This supports docker pull with pulp3. Crane (pulp2) is still supported and the
registry URL is dynamically determined based on the support for Docker blobs.

Previously, the RegistryResource class would run code when loaded, setting up
the inherited class variables. This commit moves this and the new logic to a class
method. Since there are now backend calls (to check for pulp3) to set up the class
variables, it seems like this should be a more deliberate action. This also makes testing
much easier.

I'm open to other suggestions on how to handle the class inheritance and the class variables.

To test:
- Modify foreman/config/settings.plugins.d/katello.yaml, adding the pulp_registry_url
```yaml
  :container_image_registry:
    :crane_url: https://localhost:5000
    :pulp_registry_url: http://localhost:24816
    :registry_ca_cert_file: /etc/pki/katello/certs/katello-default-ca.crt
```
- Set up docker on pulp3 box. You will need an up-to-date box since https://github.com/pulp/pulp_docker/pull/391 was merged
- Sync repos, you can use https://cloud.docker.com/u/jomitsch/repository/docker/jomitsch/workflow-test for a small one.
- docker login $HOSTNAME
- docker pull $HOSTNAME/default_organization-docker-workflow_test (check name in docker repo details)
- docker pull should be successful

It's worth trying out some other repos or registries (like quay.io) and also we should ensure pulp2 functionality is not affected

Installer PRs:
https://github.com/theforeman/puppet-katello_devel/pull/207
https://github.com/theforeman/puppet-katello/pull/296

### DIFF
--- a/app/lib/katello/resources/registry.rb
+++ b/app/lib/katello/resources/registry.rb
@@ -12,22 +12,37 @@ module Katello
 
         def self.get(path, headers = {:accept => :json})
           logger.debug "Sending GET request to Registry: #{path}"
-          client = RegistryResource.rest_client(Net::HTTP::Get, :get, path)
+          client = RegistryResource.load_class.rest_client(Net::HTTP::Get, :get, path)
           client.get(headers)
         end
       end
 
       class RegistryResource < HttpResource
-        if SETTINGS[:katello][:container_image_registry]
-          cfg = SETTINGS[:katello][:container_image_registry]
-          url = cfg[:crane_url]
-          uri = URI.parse(url)
-          self.prefix = uri.path
-          self.site = "#{uri.scheme}://#{uri.host}:#{uri.port}"
-          self.ca_cert_file = cfg[:crane_ca_cert_file]
-        end
-
         class << self
+          def load_class
+            container_config = SETTINGS.dig(:katello, :container_image_registry)
+            registry_url = nil
+            pulp_master = ::SmartProxy.pulp_master
+
+            # Pulp 3 has its own registry
+            if pulp_master && pulp_master.content_pulp3_support?(::Katello::DockerBlob::CONTENT_TYPE)
+              registry_url = pulp_master.setting('Pulp3', 'content_app_url')
+              # Assume the registry uses the same CA as the Smart Proxy
+              ca_cert_file = Setting[:ssl_ca_file]
+            elsif container_config
+              registry_url = container_config[:crane_url]
+              ca_cert_file = container_config[:registry_ca_cert_file]
+            end
+
+            fail Errors::ContainerRegistryNotConfigured unless registry_url
+
+            uri = URI.parse(registry_url)
+            self.prefix = uri.path
+            self.site = "#{uri.scheme}://#{uri.host}:#{uri.port}"
+            self.ca_cert_file = ca_cert_file
+            self
+          end
+
           def process_response(response)
             debug_level = response.code >= 400 ? :error : :debug
             logger.send(debug_level, "Registry request returned with code #{response.code}")
--- a/test/factories/smart_proxy_factory.rb
+++ b/test/factories/smart_proxy_factory.rb
@@ -21,7 +21,8 @@ FactoryBot.modify do
 
         smart_proxy_feature = proxy.smart_proxy_features.select { |spf| spf.feature_id == v3_feature.id }.first
         smart_proxy_feature.capabilities = plugins
-        smart_proxy_feature.settings = {pulp_url: "https://#{Socket.gethostname}"}
+        smart_proxy_feature.settings = { pulp_url: "https://#{Socket.gethostname}",
+                                         content_app_url: "http://localhost:24816" }
         smart_proxy_feature.save!
       end
     end
--- a/test/lib/resources/registry_test.rb
+++ b/test/lib/resources/registry_test.rb
@@ -1,0 +1,24 @@
+require 'katello_test_helper'
+
+module Katello
+  module Resources
+    class RegistryTest < ActiveSupport::TestCase
+      before do
+        @crane_url = "https://localhost:5000"
+        SETTINGS[:katello][:container_image_registry] = {
+          crane_url: @crane_url
+        }
+      end
+
+      def test_pulp3_registry_url
+        pulp_master = FactoryBot.create(:smart_proxy, :default_smart_proxy, :with_pulp3)
+        ::SmartProxy.expects(:pulp_master).at_least_once.returns(pulp_master)
+        assert_equal Registry::RegistryResource.load_class.site, 'http://localhost:24816'
+      end
+
+      def test_crane_registry_url
+        assert_equal Registry::RegistryResource.load_class.site, @crane_url
+      end
+    end
+  end
+end
--- a/app/controllers/katello/api/registry/registry_proxies_controller.rb
+++ b/app/controllers/katello/api/registry/registry_proxies_controller.rb
@@ -79,10 +79,7 @@ module Katello
 
     def authorize_repository_write
       @repository = find_writable_repository
-      unless @repository
-        not_found params[:repository]
-        return false
-      end
+      return item_not_found(params[:repository]) unless @repository
       true
     end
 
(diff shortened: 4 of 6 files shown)


sjha4:
Don't have enough context around this to ACK but following the steps, docker pull was successful on a host registered to a pulp3-enabled server. :heavy_check_mark: 
docker pull

.rubocop_todo.yml
```diff
@@ -122,13 +122,6 @@ Metrics/ClassLength:
...
   Max: 14
 
-# Offense count: 875
-# Cop supports --auto-correct.
-# Configuration parameters: AutoCorrect, AllowHeredoc, AllowURI, URISchemes, IgnoreCopDirectives, IgnoredPatterns.
-# URISchemes: http, https
-Metrics/LineLength:
-  Max: 200
-
 # Offense count: 23
```
sjha4: :+1:

ekohl:
In https://github.com/theforeman/puppet-katello/pull/296 I started a rather large design discussion about this. Please read it before merging :)

app/lib/katello/resources/registry.rb
```diff
@@ -12,22 +12,38 @@ def self.logger
...
+            registry_url = nil
+
+            if container_config
+              crane_url = container_config[:crane_url]
+              pulp_registry_url = container_config[:pulp_registry_url]
+              registry_url = crane_url if crane_url
+              # pulp 3 acts as its own registry
+              if pulp_registry_url && content_pulp3_support?(::Katello::DockerBlob::CONTENT_TYPE)
+                registry_url = pulp_registry_url
+              end
```
sjha4: Would this be clearer? 
`registry_url = content_pulp3_support?(::Katello::DockerBlob::CONTENT_TYPE) ?  pulp_registry_url: crane_url`

johnpmitsch:
I updated this to use the implementation in https://github.com/theforeman/smart_proxy_pulp/commit/1c8e95338a81d11a6ed5bc7249196bc9f695f3aa#diff-956144ad423c270b7224b0028bc0ebd9

To test, you'll need to apply this patch locally to `/usr/share/gems/gems/smart_proxy_pulp-1.4.1`

You can just apply this quick patch for the gem version (which doesn't include test code)
```
diff --git a/lib/smart_proxy_pulp_plugin/pulp3_plugin.rb b/lib/smart_proxy_pulp_plugin/pulp3_plugin.rb
index a224dce..d2e94c7 100644
--- a/lib/smart_proxy_pulp_plugin/pulp3_plugin.rb
+++ b/lib/smart_proxy_pulp_plugin/pulp3_plugin.rb
@@ -4,15 +4,18 @@ module PulpProxy
   class Pulp3Plugin < ::Proxy::Plugin
     plugin "pulp3", ::PulpProxy::VERSION
     default_settings :pulp_url => 'https://localhost/pulp/',
+                     :content_app_url => 'https://localhost:24816/',
                      :pulp_dir => '/var/lib/pulp',
                      :pulp_content_dir => '/var/lib/pulp/content',
                      :mirror => false
 
     load_validators :url => ::PulpProxy::Validators::PulpUrlValidator
     validate :pulp_url, :url => true
+    validate :content_app_url, :url => true
 
     expose_setting :pulp_url
     expose_setting :mirror
+    expose_setting :content_app_url
     capability( lambda do ||
       Pulp3Client.capabilities
     end)
```

then update `/etc/foreman-proxy/settings.d/pulp3.yml` adding `:content_app_url: http://localhost:24816/`

restart foreman-proxy service, and in the UI refresh the main proxy.

Then you can sync docker repos and pull them as previously specified (no need to update the settings in the old instructions)

I also added some error handling for not found repositories and tags (the message wasn't parsable by the docker/podman client before). Now it displays a message when a repo or tag is not found, similar to dockerhub.

```
[root@pulp3-again ~]# docker pull pulp3-again.jomitsch.example.com/pizza
Using default tag: latest
Trying to pull repository pulp3-again.jomitsch.example.com/pizza ... 
unknown: pizza was not found!

[root@pulp3-again ~]# docker pull pulp3-again.jomitsch.example.com/default_organization-docker-workflow_test:cupcakes
Trying to pull repository pulp3-again.jomitsch.example.com/default_organization-docker-workflow_test ... 
unknown: cupcakes was not found!
```

app/lib/katello/resources/registry.rb
```diff
@@ -12,22 +12,37 @@ def self.logger
...
+            if pulp_master && pulp_master.content_pulp3_support?(::Katello::DockerBlob::CONTENT_TYPE)
+              registry_url = pulp_master.setting('Pulp3', 'content_app_url')
+              # Assume the registry uses the same CA as the Smart Proxy
+              ca_cert_file = Setting[:ssl_ca_file]
+            elsif container_config
+              registry_url = container_config[:crane_url]
+              ca_cert_file = container_config[:registry_ca_cert_file]
+            end
+
+            logger.send(:error, "No container registry url specified") unless registry_url
```
ekohl: Should this be a harder error? `URI.parse(nil)` raises an exception and can happen in misconfigurations.

johnpmitsch:
@ekohl updated to throw a proper error

test/lib/resources/registry_test.rb
```diff
@@ -0,0 +1,25 @@
...
+        SETTINGS[:katello][:container_image_registry] = {
+          crane_url: "https://localhost:5000"
+        }
+      end
+
+      def test_pulp3_registry_url
+        pulp_master = FactoryBot.create(:smart_proxy, :default_smart_proxy, :with_pulp3)
+        ::SmartProxy.expects(:pulp_master).at_least_once.returns(pulp_master)
+        assert_equal Registry::RegistryResource.load_class.site,
+                     ::SmartProxy.pulp_master.setting("Pulp3", "content_app_url")
```
ekohl: This may be personal style, but I like to have explicit checks to avoid silently checking `nil` equals `nil`.

```suggestion
                     'http://localhost:24816')
```

test/lib/resources/registry_test.rb
```diff
@@ -0,0 +1,25 @@
...
+      end
+
+      def test_crane_registry_url
+        crane_url = SETTINGS[:katello][:container_image_registry][:crane_url]
+        assert_equal Registry::RegistryResource.load_class.site, crane_url
```
ekohl: Same here:
```suggestion
        assert_equal Registry::RegistryResource.load_class.site, 'https://localhost:5000'
```

johnpmitsch:
I removed the part that updates rubocop, I think that should be dealt with separately

test/factories/smart_proxy_factory.rb
```diff
@@ -21,7 +21,8 @@
 
         smart_proxy_feature = proxy.smart_proxy_features.select { |spf| spf.feature_id == v3_feature.id }.first
         smart_proxy_feature.capabilities = plugins
-        smart_proxy_feature.settings = {pulp_url: "https://#{Socket.gethostname}"}
+        smart_proxy_feature.settings = { pulp_url: "https://#{Socket.gethostname}",
```
ekohl: Unrelated, but typically I like to explicit static values such as `https://pulp.example.com` since this you know that if there is some integration test, it will not magically work sometimes because the host is configured. It also sets the expectation that pulp can be on a remote host and that should be supported.

test/factories/smart_proxy_factory.rb
```diff
@@ -21,7 +21,8 @@
...
+                                         content_app_url: "http://localhost:24816" }
```
ekohl: Slightly similar remark here.

test/factories/smart_proxy_factory.rb
johnpmitsch: I get the sentiment, but this was pre-existing but I'll leave it for this PR and we can consider removing it outside of this

test/factories/smart_proxy_factory.rb
johnpmitsch: this is will just be a string comparison so I think its fine to use `localhost`

johnpmitsch:
@ekohl updated the tests to use fixed values

sjha4:
Tested :heavy_check_mark: 
Pulp2:
docker_pull
Pulp3 works well too. :+1:

johnpmitsch:
@ekohl can you approve/comment if you are ok with the latest changes?

PR #8262, opened 2019-08-09
Fixes #27450: Check permissions for React pages

jeremylenz:
To test this PR:

1.  Create a User with no roles (no permissions)
2.  Log in as (or impersonate) that user.  
3.  Navigate manually (by typing in the URL bar) to `/subscriptions` and `/redhat_repositories`. 

You should now see the permission denied message for those two pages.

[Edit: Please also test "Flow 1" (see below)]


webpack/redux/reducers/RedHatRepositories/enabled.js
```diff
@@ -57,40 +58,46 @@ const changeRepoState = (state, repoToChange, stateDiff) => (
...
+      return changeRepoState(state, action.repository, { loading: true });
+    case DISABLE_REPOSITORY_FAILURE:
+      return changeRepoState(state, action.payload.repository, { loading: false });
+    case ENABLED_REPOSITORIES_REQUEST:
+      if (!action.silent) {
+        return state.set(['loading'], true);
+      }
+      return state;
+    case ENABLED_REPOSITORIES_SUCCESS:
+      ({ page, per_page, subtotal, results } = action.response); // eslint-disable-line
```
sjha4: How does changing `const` to `let` help in this and similar case? Asking cause we do use `const` in some other reducers..

webpack/redux/reducers/RedHatRepositories/sets.js
```diff
@@ -10,6 +11,7 @@ import {
 
 import { initialState } from './sets.fixtures.js';
 
+
```
waldenraines: Empty new line here.

webpack/components/PermissionDenied/index.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { Button, Icon } from 'patternfly-react';
+import PropTypes from 'prop-types';
+
+const PermissionDenied = ({ missingPermissions }) => {
```
waldenraines: I hate to ask this but should this live in foreman?

webpack/redux/reducers/RedHatRepositories/enabled.js
```diff
@@ -57,40 +58,46 @@ const changeRepoState = (state, repoToChange, stateDiff) => (
   )
 );
 
+// eslint-disable-next-line consistent-return
```
waldenraines: Why disable this here?

webpack/redux/reducers/RedHatRepositories/enabled.js
jeremylenz: @sjha4 This was just me fighting with ESLint. 
1.  First, I tried using `const` in the case statement.  It I wasn't allowed to have a declaration within the case statement.
2. Then, I moved the `const` declaration up to the top.  But since those values aren't always defined, it threw errors.
3. So, I ended up declaring them at the top and assigning the values in the case.

waldenraines: Can you use `const` with that line disabled?  Also if we're going to disable this line shouldn't we disable just for that particular rule?  @sharvit @amirfefer any other ideas here?

webpack/redux/reducers/RedHatRepositories/enabled.js
jeremylenz: @waldenraines good catch, I don't think this is needed anymore.  I put it in before I added the return on line 76.

webpack/components/PermissionDenied/index.js
jeremylenz: @waldenraines Idk.. what's our rule for what should live in Foreman vs. Katello?   This component is basically a copy of a Foreman ERB template, re-written in React.  To me it makes sense to have it live here, because that's where we are using the data.  If it lived in Foreman what would the difference be?  We would just import it from a different location?

If we're going to update Foreman anyway, it would also make sense to have the API give us an array of missing permissions, rather than parsing them from the error message string..

webpack/redux/reducers/RedHatRepositories/enabled.js
jeremylenz: @waldenraines  Not sure what I was doing before, but today I was able to get Eslint to use `const` in the case statement instead, which is much better.  thanks for the suggestion!

webpack/redux/reducers/RedHatRepositories/enabled.js
```diff
@@ -58,39 +59,45 @@ const changeRepoState = (state, repoToChange, stateDiff) => (
...
+      return changeRepoState(state, action.repository, { loading: true });
+    case DISABLE_REPOSITORY_FAILURE:
+      return changeRepoState(state, action.payload.repository, { loading: false });
+    case ENABLED_REPOSITORIES_REQUEST:
+      if (!action.silent) {
+        return state.set(['loading'], true);
+      }
+      return state;
+    case ENABLED_REPOSITORIES_SUCCESS: // eslint-disable-line no-case-declarations
+      const {
```
sjha4: Does it help to wrap the case blocks in brackets. As per documentation for [eslint-disable-line no-case-declarations](https://eslint.org/docs/rules/no-case-declarations), that should be accepted.

jeremylenz: @sjha4 cool, I'm sold.  I'll add the block.  thanks!

webpack/components/PermissionDenied/index.js
waldenraines: I think we should both things mentioned above.  We don't have to port this to foreman right now but, to me, it makes sense that it lives there since it will eventually be used in both foreman and katello.

@amirfefer @sharvit thoughts?

jeremylenz: Maybe I'm beginning to realize what the `move_to_foreman` folder is for...

jeremylenz:
Per conversation today we are going to move the `<PermissionDenied>` component to Foreman and modify this PR accordingly.

webpack/scenes/RedHatRepositories/index.js
```diff
@@ -30,6 +31,14 @@ class RedHatRepositoriesPage extends Component {
     const { enabledRepositories, repositorySets } = this.props;
     const { repoParams } = createEnabledRepoParams(enabledRepositories);
 
+    if (repositorySets.missingPermissions && repositorySets.missingPermissions.length > 0) {
```
sharvit: Notice this mechanism:
```js
if (!havePermissions()) {
  return <PermissionDenied />
} else { continue }
```

It is also used in the `SubscriptionsPage` and we might adopt it to other pages.
Can it be reused as a HOC?
So your logic can be:
```js
if (!havePermissions()) {
  return <PermissionDenied />
} else { <Component /> }
```

webpack/scenes/RedHatRepositories/index.js
```diff
@@ -30,6 +31,14 @@ class RedHatRepositoriesPage extends Component {
...
+      return <PermissionDenied missingPermissions={repositorySets.missingPermissions} />;
+    }
+    if (enabledRepositories.missingPermissions
+      && enabledRepositories.missingPermissions.length > 0) {
```
amirfefer: `.length > 0` is redundant, because 0 is `false` in js

webpack/scenes/RedHatRepositories/index.js
```diff
@@ -98,11 +107,13 @@ RedHatRepositoriesPage.propTypes = {
   enabledRepositories: PropTypes.shape({
     loading: PropTypes.bool,
     search: PropTypes.shape({}),
+    missingPermissions: PropTypes.array,
```
amirfefer: if `missingPermissions` can be only strings, you can be more explicit 
`PropTypes.arrayOf(PropTypes.string)`

webpack/redux/reducers/RedHatRepositories/enabled.js
```diff
@@ -58,39 +59,46 @@ const changeRepoState = (state, repoToChange, stateDiff) => (
...
-      repositories: [],
-      loading: false,
-      error: action.error,
-    });
+    case ENABLED_REPOSITORIES_FAILURE:
+      return Immutable({
+        repositories: [],
+        loading: false,
+        error: get(action, ['error']),
+        missingPermissions: parseMissingPermissions(get(action, ['result', 'response', 'data', 'error', 'details'])),
```
amirfefer: I prefer readable code, how about something like:

``` js 
const errors = get(action, ['result', 'response', 'data', 'error', 'details']);
// some code

missingPermissions: parseMissingPermissions(errors);
```

I prefer that the reducer will get the filtered data already, so this logic IMHO should be in the action.

webpack/redux/reducers/helpers.js
```diff
@@ -0,0 +1,14 @@
+// eslint-disable-next-line import/prefer-default-export
+export function parseMissingPermissions(errMsg) {
+  const defaultMsg = ['unknown'];
+  try {
+    // 1. "Missing one of the required permissions: view_subscriptions, edit_manifest"
+    // 2. ["Missing one of the required permissions", "view_subscriptions, edit_manifest"]
+    // 3. ["view_subscriptions", "edit_manifest"]
+    const missingPermissions = errMsg.split(': ')[1].split(', ');
```
amirfefer: I'm not sure if this convention exists in foreman and its plugins

webpack/redux/reducers/helpers.js
```diff
@@ -0,0 +1,14 @@
+// eslint-disable-next-line import/prefer-default-export
+export function parseMissingPermissions(errMsg) {
```
amirfefer: can you use an arrow function? 
 ``` js
export const parseMissingPermissions = errMsg => { //some code}

webpack/scenes/RedHatRepositories/index.js
```diff
@@ -15,6 +15,7 @@ import SearchBar from './components/SearchBar';
 import RecommendedRepositorySetsToggler from './components/RecommendedRepositorySetsToggler';
 import { getSetsComponent, getEnabledComponent } from './helpers';
 import api from '../../services/api';
+import PermissionDenied from '../../components/PermissionDenied/index';
```
amirfefer: `index` file is automatically imported

```js
import PermissionDenied from '../../components/PermissionDenied
```

webpack/redux/reducers/helpers.js
jeremylenz: @amirfefer you're right, it's definitely hacky.  I think we should just have the Foreman API give us the array we need.

webpack/redux/reducers/helpers.js
jeremylenz: @amirfefer  I can, I guess I just used a regular function because I wasn't worried about the value of `this`.  I guess I didn't realize the other helper files have arrow functions.  They're prettier anyway so I will change it :)

waldenraines:
Requires https://github.com/theforeman/foreman/pull/6979

jeremylenz:
Updated to fix Flow 1.

webpack/redux/reducers/RedHatRepositories/enabled.js
jeremylenz: @waldenraines  ^^

webpack/redux/reducers/RedHatRepositories/enabled.js
```diff
@@ -91,12 +91,14 @@ export default (state = initialState, action) => {
         search: action.search,
       });
     }
+    // eslint-disable-next-line no-case-declarations
```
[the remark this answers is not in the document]
jeremylenz: @waldenraines Yes, it was resolved by adding a block { } to the case.   That eslint-disable shouldn't be there any more, were you looking at an old version somehow?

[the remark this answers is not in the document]
jeremylenz: @waldenraines sorry, I had removed it in a different place but not this one.  updating now!

webpack/redux/reducers/RedHatRepositories/enabled.js
```diff
@@ -58,39 +58,47 @@ const changeRepoState = (state, repoToChange, stateDiff) => (
...
-      repositories: [],
-      loading: false,
-      error: action.error,
-    });
+    case ENABLED_REPOSITORIES_FAILURE: {
+      return Immutable({
+        repositories: [],
+        loading: false,
+        error: get(action, ['error']),
+        missingPermissions: get(action, ['result', 'response', 'data', 'error', 'missing_permissions']),
```
laviro: `lodash.get` is usually used to set up a default value when undefined is returned from the entered path,
For this case I would use the regular destructing. If it's not too difficult there's already the action destructing above
which can be reused.

webpack/redux/reducers/RedHatRepositories/sets.js
```diff
@@ -40,6 +40,7 @@ export default (state = initialState, action) => {
       return Immutable({
         error: payload,
         loading: false,
+        missingPermissions: get(payload, ['response', 'data', 'error', 'missing_permissions']),
```
laviro: Can be destructed too as other cases here are using `payload.response`

webpack/redux/reducers/RedHatRepositories/sets.fixtures.js
```diff
@@ -137,6 +137,7 @@ export const successState = Immutable({
 });
 
 export const errorState = Immutable({
+  error: { response: { data: { error: { missing_permissions: ['unknown'] } } } },
```
laviro: Would it be possible to use camel case here?
it can be confusing to have both `missing_permissions` and `missingPermissions` in the same state.

webpack/redux/reducers/RedHatRepositories/enabled.js
laviro: another thought - in my opinion the actions is the place to write "dirty" code that do all of the adjustments,
and the reducer should receive it as clean as possible.

webpack/scenes/RedHatRepositories/RedHatRepositoriesPage.js
```diff
@@ -0,0 +1,118 @@
...
+  loadData() {
+    this.props.loadEnabledRepos();
+    this.props.loadRepositorySets({ search: { filters: ['rpm'] } });
+  }
+
+  render() {
+    const { enabledRepositories, repositorySets } = this.props;
+    const { repoParams } = createEnabledRepoParams(enabledRepositories);
+
+    if (repositorySets.missingPermissions && repositorySets.missingPermissions.length > 0) {
```
laviro: you can use `lodash.isEmpty`:
```if ( !isEmpty(repositorySets.missingPermissions) )```
and
```if ( !isEmpty(enabledRepositories.missingPermissions) )```
looks cleaner

File: webpack/scenes/RedHatRepositories/RedHatRepositoriesPage.js [hunk 18]
```diff
@@ -0,0 +1,118 @@
...
+import api from '../../services/api';
+
+class RedHatRepositoriesPage extends Component {
+  componentDidMount() {
+    this.loadData();
+  }
+
+  loadData() {
+    this.props.loadEnabledRepos();
+    this.props.loadRepositorySets({ search: { filters: ['rpm'] } });
```
laviro: should we use `async & await` here?

webpack/scenes/RedHatRepositories/RedHatRepositoriesPage.js
```diff
@@ -0,0 +1,118 @@
...
+              <RecommendedRepositorySetsToggler
+                enabled={repositorySets.recommended}
+                onChange={value => this.props.updateRecommendedRepositorySets(value)}
+                className="recommended-repositories-toggler"
+              />
+            </div>
+            <LoadingState loading={repositorySets.loading} loadingText={__('Loading')}>
+              {getSetsComponent(
+                repositorySets,
+                (pagination) => {
```
laviro: would it make sense to move it into a class handler? brackets can be removed too.

webpack/scenes/RedHatRepositories/RedHatRepositoriesPage.js
```diff
@@ -0,0 +1,118 @@
+/* eslint-disable import/no-extraneous-dependencies */
+/* eslint import/no-unresolved: [2, { ignore: [foremanReact/*] }] */
+/* eslint-disable import/no-unresolved */
+
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { Grid, Row, Col } from 'react-bootstrap';
```
laviro: I think `patternfly-react` is using `react-bootstrap` already :)

webpack/redux/reducers/RedHatRepositories/sets.fixtures.js
jeremylenz: @laviro The snake case here is just because that's how Rails gives it to us. Is there a best practice in general for exactly where in the React world it should make the jump from snake case to camel case?  I see snake case variables in JS all the time and agree that it's messy.

webpack/redux/reducers/RedHatRepositories/enabled.js
jeremylenz: @laviro good point, the latter makes sense to me! I moved the "dirty" data traversal to the action creator.

Regarding `lodash.get`, In addition to providing default values, `get` also guards against throwing errors if one of your intermediary keys is undefined (for example, `result` or `data` etc.)
After moving it to the action, I tried to use destructuring to get the value, and I had something like
```
const { response: { data: { error: { missing_permissions } } } } = error
```
But it broke another test.  So I switched back to using `get`.
I know I could still use destructuring by using default values, like 
```
const { response: { data: { error: { missing_permissions } =  {}} =  {}} = error
```
But I feel like this is hard to understand and easy to get the syntax wrong, whereas lodash `get` is very straightforward.

Until JS optional chaining becomes part of JavaScript, I feel like lodash `get` is a good solution.

webpack/scenes/RedHatRepositories/RedHatRepositoriesPage.js
jeremylenz: @laviro Cool, didn't know that one :)

webpack/scenes/RedHatRepositories/RedHatRepositoriesPage.js
jeremylenz: @laviro I don't think it's needed here, since React takes care of the rendering whenever the data comes back and neither dataset is dependent on the other.

webpack/scenes/RedHatRepositories/RedHatRepositoriesPage.js
jeremylenz: @laviro Yes, I think this entire page could use a refactor.  But maybe in another PR.

webpack/scenes/RedHatRepositories/RedHatRepositoriesPage.js
jeremylenz: @laviro Yeah I remember looking at that too and thinking why bootstrap?  But see the previous comment :smiley:

webpack/redux/reducers/RedHatRepositories/sets.fixtures.js
laviro: usually I would prefer to rename it while receiving it from the API if possible.
Otherwise it will fastly spread to other areas of the JS code.

webpack/scenes/RedHatRepositories/RedHatRepositoriesPage.js
laviro: Yeah, I agree..
My first impression was that you wrote it from scratch.

webpack/redux/reducers/RedHatRepositories/sets.fixtures.js
laviro: usually I would prefer to rename it while receiving it from the API if possible.
Otherwise it will spread fast into other areas of the JS code.

webpack/scenes/RedHatRepositories/RedHatRepositoriesPage.js
laviro: Thanks, sorry, my first impression was that you wrote it from scratch.

PR #8232, opened 2019-07-24
Fixes #27251 - Allow Host profile reuse during kickstart

jturel:
The most recent change to our registration process made it a requirement
that a host profile could not be associated to unless it was unregistered first.

This change aims to help with that.

- Unregistering a host is no longer required to assume the profile of
an existing host when registering. The only criteria are that the host's name
and dmi.system.uuid fact match the profile and are sufficiently unique

- If a host's dmi.system.uuid has changed but the hostname is the same, profile reuse is allowed
*if* the build flag is set to true which indicates that the registration is likely happening via kickstart
and rebuild may have changed the dmi.system.uuid

**To test**

**scenario 1**: hostname and dmi.system.uuid haven't changed, but system is still registered results in successful registration

- spin up katello-client so that it registers to your dev box with this change (or client of your choice)
- on the client run `subscription-manager clean`
- `subscription-manager register` should **succeed**

**scenario 2**: hostname is the same, dmi.system.uuid changed, host is not in build mode. Registration fails.

- run `subscription-manager clean`
- using the same client, modify the dmi.system.uuid by creating a custom fact in /etc/rhsm/facts/uuid.facts with content `{"dmi.system.uuid": "my-custom-uuid"}`
- `subscription-manager register` should **fail** since there exists a host profile with the same name but different UUID

**scenario 3**: hostname is the same, dmi.system.uuid changed, host is in build mode. Registration **succeeds**.

- With client configured for scenario 2, set the hosts build flag to true from the rails console:
```
::Host.find_by_name('katello-client.jturel.example.com').update_attributes!(build: true)
```
This simulates that the host is being rebuilt and the next registration will come from the kickstart
- Registering from the client should now succeed even with the altered dmi.system.uuid in place

PR #8263, opened 2019-08-09
Fixes #27580 - Allow register to profile without DMI UUID

jturel:
bootstrap.py usage was broken by #8232 since we didn't have a test case for the workflow used by the script. This issue is high priority in the downstream context and while #8248 fixes the problem we likely need to deliver a minimal changeset. ie this must be merged before #8242 

**To test (without change)**

- Spin up katello-client
- attempt to register it with bootstrap.py (get it from /pub/ of your katello instance) example:
```
python /root/bootstrap.py --login=admin --password=changeme --server=centos7-devel.jturel.example.com --location='Default Location' --organization='Default Organization' -g MyHostGroup -a testkey --add-domain --force
```

It will fail with error: `Please unregister or remove hosts which match this host before registering`

Check out this PR and run the bootstrap.py command again and the registration will succeed!

### DIFF
--- a/app/models/katello/host/subscription_facet.rb
+++ b/app/models/katello/host/subscription_facet.rb
@@ -272,9 +272,17 @@ module Katello
 
         if hosts_size == 1
           host = hosts.first
-          found_uuid = host.fact_values.find { |fv| fv.fact_name_id == uuid_fact_id }
 
-          return host if host.name == host_name && (host.build || found_uuid&.value == host_uuid)
+          if host.name == host_name
+            unless host.build
+              found_uuid = host.fact_values.where(fact_name_id: uuid_fact_id).first
+              if found_uuid && found_uuid.value != host_uuid
+                fail Katello::Errors::RegistrationError, _("This host is reporting a DMI UUID that differs from the existing registration.")
+              end
+            end
+
+            return host
+          end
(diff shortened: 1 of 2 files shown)


sjha4:
[test katello]

evgeni:
I didn't try to *run* this, but the code logic looks exactly how I've imagined the fix after filing the BZ on Friday. @jturel++ for super quick fix!

jjeffers:
Followed the test steps as outlined in the PR summary. 
I first created a host group "Test8263".

As root:
```./bootstrap.py --login=admin --password=changeme --server=centos7-katello-devel.jjeffers.example.com --location='Default Location' --organization='Default Organization' -g Test8263 -a testkey --add-domain --force```

The process seemed to succeed except for 2 steps:
```[ERROR], [2019-08-12 13:52:09], EXITING: [0 element in array for search key 'title="Test8263"' in API '/api/v2/hostgroups'. Please note that all searches are case-sensitive.] failed to execute properly.
[ERROR], [2019-08-12 13:52:09], EXITING: [Please also ensure that the user has permissions to view the searched objects. Fatal error.] failed to execute properly.```

jturel:
Thanks @jjeffers !

PR #8269, opened 2019-08-13
Fixes #27221 - Verify Act Key Content Overrides

johnpmitsch:
This commit adds validation that the content label sent with the the content override matches to
an actual label in the organization. Previously, you could submit any label and get a successful
response back, but nothing actually updated. Now, a proper error message should be returned by
the API. Since multiple activation keys can be used on client, it is best to validate the
label exists in the org rather than just the key.

The 'content_override' param was deprecated previously, but not actually removed (this is a whole other
issue with how we handle deprecations). It is removed in this PR

### DIFF
--- a/app/controllers/katello/api/v2/activation_keys_controller.rb
+++ b/app/controllers/katello/api/v2/activation_keys_controller.rb
@@ -213,36 +213,31 @@ module Katello
 
     api :PUT, "/activation_keys/:id/content_override", N_("Override content for activation_key")
     param :id, :number, :desc => N_("ID of the activation key"), :required => true
-    param :content_override, Hash, :desc => N_("Content override parameters"), :deprecated => true do
-      param :content_label, String, :desc => N_("Label of the content"), :required => true
-      param :value, String, :desc => N_("Override to a boolean value or 'default'"), :required => true
-    end
-    param :content_overrides, Array, :desc => N_("Array of Content override parameters") do
+    param :content_overrides, Array, :desc => N_("Array of Content override parameters to be added in bulk") do
       param :content_label, String, :desc => N_("Label of the content"), :required => true
       param :value, String, :desc => N_("Override value. Provide a boolean value if name is 'enabled'"), :required => false
       param :name, String, :desc => N_("Override parameter key or name. Note if name is not provided the default name will be 'enabled'"), :required => false
       param :remove, :bool, :desc => N_("Set true to remove an override and reset it to 'default'"), :required => false
     end
     def content_override
-      content_overrides = []
-      if params[:content_override]
-        ::Foreman::Deprecation.api_deprecation_warning("The parameter content_override will be removed in Katello 3.5. Please update to use the content_overrides parameter.")
-        content_override_params = {:content_label => params[:content_override][:content_label]}
-        value = params[:content_override][:value]
-        if value == "default"
-          content_override_params[:remove] = true
-        elsif value
-          content_override_params[:value] = ::Foreman::Cast.to_bool(value)
+      if params[:content_overrides]
+        organization = @activation_key.organization
+        specified_labels = []
+        content_override_values = params[:content_overrides].map do |override|
+          specified_labels << override[:content_label]
+          validate_content_overrides_enabled(override)
         end
-        content_overrides << content_override_params
-      elsif params[:content_overrides]
-        content_overrides = params[:content_overrides]
-      end
+        specified_labels.uniq!
+        existing_labels = organization.contents.where(label: specified_labels).uniq
 
-      content_override_values = content_overrides.map do |override_params|
-        validate_content_overrides_enabled(override_params)
+        unless specified_labels.size == existing_labels.size
+          missing_labels = specified_labels - existing_labels.pluck(:label)
+          msg = "Content label(s) \"#{missing_labels.join(", ")}\" were not found in the Organization \"#{organization}\""
+          fail HttpErrors::BadRequest, _(msg)
+        end
+
+        @activation_key.set_content_overrides(content_override_values)
       end
-      @activation_key.set_content_overrides(content_override_values)
       respond_for_show(:resource => @activation_key)
     end
 
--- a/test/controllers/api/v2/activation_keys_controller_test.rb
+++ b/test/controllers/api/v2/activation_keys_controller_test.rb
@@ -2,6 +2,7 @@
 
 require "katello_test_helper"
 
+# rubocop:disable Metrics/ClassLength
 module Katello
   class Api::V2::ActivationKeysControllerTest < ActionController::TestCase
     include Support::ForemanTasks::Task
@@ -386,24 +387,18 @@ module Katello
       denied_perms = [@view_permission, @create_permission, @destroy_permission]
 
       assert_protected_action(:content_override, allowed_perms, denied_perms) do
-        put(:content_override, params: { :id => @activation_key.id, :content_label => 'some-content', :name => 'enabled', :value => 1 })
+        put(:content_override,
+            params: { id: @activation_key.id,
+                      content_overrides: [{:content_label => 'fedora', :name => "enabled", :value => true}]
+                    }
+           )
       end
     end
 
-    def test_content_override
-      ActivationKey.any_instance.expects(:set_content_overrides).returns(true)
-
-      put(:content_override, params: { :id => @activation_key.id, :content_override => {:content_label => 'some-content',
-                                                                                        :value => 1} })
-
-      assert_response :success
-      assert_template 'api/v2/activation_keys/show'
-    end
-
-    def test_content_override_bulk
-      overrides = [{:content_label => 'some-content', :name => "enabled", :value => true},
-                   {:content_label => 'some-content1', :value => 0},
-                   {:content_label => 'some-content3', :name => "mirrorlist", :remove => true}
+    def test_bulk_content_override
+      overrides = [{:content_label => 'fedora', :name => "enabled", :value => true},
+                   {:content_label => 'fedora', :value => 0},
+                   {:content_label => 'fedora', :name => "mirrorlist", :remove => true}
                   ]
 
       expected_content_labels = overrides.map { |o| o[:content_label] }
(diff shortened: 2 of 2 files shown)


johnpmitsch:
To test:

- Create an activation key, associate subscriptions
- List product content for subscription and see the content labels
```
hammer> activation-key product-content --name bug --organization-id 2
```
- Try to update a valid content label and see the proper error message is returned
```
hammer> activation-key content-override --name bug --content-label coffee --value 0 --organization-id 2
Could not update content override:
  Content label "coffee" was not found in the Organization "Default Organization"
```
- Update a valid content label and ensure it succeeds
```
hammer> activation-key content-override --name bug --content-label rhel-7-server-rpms --value 0 --organization-id 2
Updated content override.
hammer> activation-key product-content --name bug --organization-id 2
-----|------------------------------------------|--------------------|------------------|----------
ID   | NAME                                     | LABEL              | DEFAULT ENABLED? | OVERRIDE 
-----|------------------------------------------|--------------------|------------------|----------
1792 | Red Hat Enterprise Linux 7 Server (RPMs) | rhel-7-server-rpms | yes              | enabled:0
-----|------------------------------------------|--------------------|------------------|----------
```
- Do the same with a custom override (not the default "enabled")
```
hammer> activation-key content-override --name bug --content-label coffee --override-name ham --value sandwich --organization-id 2
Could not update content override:
  Content label "coffee" was not found in the Organization "Default Organization"
hammer> activation-key content-override --name bug --content-label rhel-7-server-rpms --value 1 --organization-id 2
Updated content override.
hammer> activation-key product-content --name bug --organization-id 2
-----|------------------------------------------|--------------------|------------------|------------------------
ID   | NAME                                     | LABEL              | DEFAULT ENABLED? | OVERRIDE               
-----|------------------------------------------|--------------------|------------------|------------------------
1792 | Red Hat Enterprise Linux 7 Server (RPMs) | rhel-7-server-rpms | yes              | ham:sandwich, enabled:1
-----|------------------------------------------|--------------------|------------------|------------------------
```

app/controllers/katello/api/v2/activation_keys_controller.rb
```diff
@@ -213,33 +213,28 @@ def remove_subscriptions
 
     api :PUT, "/activation_keys/:id/content_override", N_("Override content for activation_key")
     param :id, :number, :desc => N_("ID of the activation key"), :required => true
-    param :content_override, Hash, :desc => N_("Content override parameters"), :deprecated => true do
+    param :content_override, Hash, :desc => N_("Content override parameters") do
       param :content_label, String, :desc => N_("Label of the content"), :required => true
       param :value, String, :desc => N_("Override to a boolean value or 'default'"), :required => true
     end
```
johnpmitsch: If this single `content_override` param can only be used for `enabled`, I think it would be good to make that more clear in the API docs, it doesn't seem clear to me

app/controllers/katello/api/v2/activation_keys_controller.rb
```diff
@@ -213,33 +213,28 @@ def remove_subscriptions
...
-      elsif params[:content_overrides]
-        content_overrides = params[:content_overrides]
-      end
+      content_overrides += [params[:content_override]] if params[:content_override]
+      content_overrides += params[:content_overrides] if params[:content_overrides]
+      organization = @activation_key.organization
+      all_content_labels = organization.content_labels
 
       content_override_values = content_overrides.map do |override_params|
+        unless override_params[:content_label] && all_content_labels.include?(override_params[:content_label])
```
jturel: I think this check would not catch when someone sends a nil value for the content_label

Also, since we aren't ever using the values of `all_content_labels` it might be faster to check presence in the org like this:

```
unless organization.content.find_by_label(override_params[:content_label])

end
```
Another option would be to use a `where` and check the result size vs input size or something like that. That's probably better when sending a lot of overrides

Would it make sense to move this check inside of `validate_content_overrides_enabled` and address this problem in a few other places?

app/controllers/katello/api/v2/activation_keys_controller.rb
```diff
@@ -213,33 +213,28 @@ def remove_subscriptions
...
-        if value == "default"
-          content_override_params[:remove] = true
-        elsif value
-          content_override_params[:value] = ::Foreman::Cast.to_bool(value)
-        end
-        content_overrides << content_override_params
-      elsif params[:content_overrides]
-        content_overrides = params[:content_overrides]
-      end
+      content_overrides += [params[:content_override]] if params[:content_override]
```
jturel: Are we no longer deprecating this param?

jturel: I see - you opted not to. Couldn't we go ahead and do it?

jturel:
Left a few notes - overall it's a good change :+1:

app/controllers/katello/api/v2/activation_keys_controller.rb
johnpmitsch: I don't see a huge reason to, I mentioned that in my commit message:
> The 'content_override' param was deprecated previously, but not actually removed (this is a whole other
issue with how we handle deprecations). This param would just be used for the 'enabled' parameter and
was hardcoded to update only that. I don't see a huge reason to remove it, so I removed the deprecation warning.
and I updated to code to pass it as a bulk param so it uses the same validations as the bulk parameters.

app/controllers/katello/api/v2/activation_keys_controller.rb
johnpmitsch: good suggestions, I'll look into them :+1:

app/controllers/katello/api/v2/activation_keys_controller.rb
johnpmitsch: I'm not sure if we should, but maybe we can discuss further. 

My view is it works fine as is and I think with some clarification in the API docs that the single param is hardcoded to change "enabled", it could stay useful and not conflict with anything else. This to me doesn't feel like a parameter changed, but additional functionality to add bulk params was added so I don't see a huge harm in keeping it around and not deprecating.

But open to other opinions from the team!

sjha4:
[test katello]

app/controllers/katello/api/v2/activation_keys_controller.rb
johnpmitsch: Thinking about this more....

I could also see the `content_override` parameter being actually removed as it looks like this is the only place the API supports this and I see `content_overrides` is the pattern everywhere else. I was worried about the UI and hammer, but it looks like those have been updated to use `content_overrides` (though we should confirm these workflows)

After all, users have had since 3.5 to update :laughing: 

I can see an argument for either way, its nice to not break APIs when possible, but also we have a new pattern and have propery deprecated... 

Thats enough thinking out loud for me... @jturel curious to hear your thoughts :wink:

jturel: I'm all for removing it since the deprecation warning has been there for ~2.5 years. Since you mentioned that hammer is using the new param I'm especially convinced. I'd wager the set of users automating this through the API and still using the old parameter are small. Do it!

jturel: If you do remove my only request is to make sure any test coverage that is lost is updated to use the new param (if there are no tests for a particular behavior w. that new param already)

app/controllers/katello/api/v2/activation_keys_controller.rb
johnpmitsch: The label check will work well (I didn't realize you can do `organization.contents`, thanks for pointing that out!)

For moving the check inside of `validate_content_overrides_enabled`, it does require an organization to be known, which doesn't seem as clear in other areas such as [host bulk actions](https://github.com/Katello/katello/blob/master/app/controllers/katello/api/v2/hosts_bulk_actions_controller.rb#L190-L206) so I think it can just stay here

jturel: So how about:

```
labels = override_params.map { |override| override[:content_label] }
contents = organization.content.where(label: labels)

unless contents.size == labels.size # one more more weren't found
   missing_labels = contents.pluck(:label) - labels
   missing_labels.each |label|
     # error stuff
    end
end

File: app/controllers/katello/api/v2/activation_keys_controller.rb [hunk 4]
```diff
@@ -213,36 +213,31 @@ def remove_subscriptions
...
-          content_override_params[:value] = ::Foreman::Cast.to_bool(value)
+      if params[:content_overrides]
+        organization = @activation_key.organization
+        specified_labels = []
+        content_override_values = params[:content_overrides].map do |override|
+          specified_labels << override[:content_label]
+          validate_content_overrides_enabled(override)
+        end
+        specified_labels.uniq!
+        existing_labels = organization.contents.where(label: specified_labels).pluck(:label).uniq
```
jturel: Minor thing - by doing `pluck` here we've actually loaded the labels into memory and based on line 233 we might not need to use them for anything. It's a slight optimization to leave this as a relation (I think?) only pluck the labels if we need them for the error messages inside the `unless` block

johnpmitsch: good point, updated!

app/controllers/katello/api/v2/activation_keys_controller.rb
```diff
@@ -213,36 +213,31 @@ def remove_subscriptions
...
+        specified_labels = []
+        content_override_values = params[:content_overrides].map do |override|
+          specified_labels << override[:content_label]
+          validate_content_overrides_enabled(override)
+        end
+        specified_labels.uniq!
+        existing_labels = organization.contents.where(label: specified_labels)
+
+        unless specified_labels.size == existing_labels.size
+          missing_labels = specified_labels - existing_labels.pluck(:label).uniq
```
jturel: I think content labels in an org are unique but I don't think there's harm in adding it. Though you might want to add the .uniq on the existing_labels relation so the size comparison is is .uniq on both sides.

Testing now!

jturel:
Works great. Let one note on .uniq placement.

I made another observation:

When I added an override for a Red Hat repo that I haven't enabled (but was in the manifest) the override is created(good) but it doesn't show up in the Repository Sets lists (product content API)

the key:

```
    "content_overrides": [
        {
            "content_label": "rhel-ha-for-rhel-7-for-system-z-beta-source-rpms",
            "name": "enabled",
            "value": "0"
        }
    ]
```

I think that override should be represented but I'm not sure if it's a bug. @jlsherrill do you know if this has regressed, always been that way, and should it not be included for any reason? I can file an issue to add it if we all think it should be listed.

Side note:

The UI calls: `/api/v2/activation_keys/8/product_content?content_access_mode_all=true&content_access_mode_env=false&enabled=true&full_result=true&include_available_content=true&organization_id=2&page=NaN&paged=true&per_page=20&search=`

params enabled, full_result, and include_available_content  aren't valid anymore

johnpmitsch:
@jturel I would file a bug to at least get rid of the extra params in the UI (please mark as trivial/easy so we can get to it during our backlog grooming!)

I'm trying to follow all the relations, but does the override make the associated product enabled? It looks like its [finding the organization's enabled products](https://github.com/Katello/katello/blob/master/app/services/katello/product_content_finder.rb#L25) to find all the repository sets.

jturel:
> @jturel I would file a bug to at least get rid of the extra params in the UI (please mark as trivial/easy so we can get to it during our backlog grooming!)
> 
Done! https://projects.theforeman.org/issues/27669

> I'm trying to follow all the relations, but does the override make the associated product enabled? It looks like its [finding the organization's enabled products](https://github.com/Katello/katello/blob/master/app/services/katello/product_content_finder.rb#L25) to find all the repository sets.

Would need to look deeper myself but that looks like a good starting point. Let's see what Justin thinks and I'll file another issue for that accordingly.

PR #8271, opened 2019-08-13
Fixes #26914 - ignore OS facts if applicable

jjeffers:
This PR addresses a long standing issue concerning distribution facts sent by hosts.

At some point CentOS stopped reporting minor versions and the fact parser considers a missing minor version as a fact to be ignored. This has the side effect of showing a "blank" when a CentOS host is registered with Katello.

If not ignored, then hosts with an updated operating system fact would revert to the minorless operating system, created during the fact parsing. This might lead to hosts "reverting" to the wrong version.

The compromise is to ignore operating system fact updates _if_ the current host has an operating system with a major version and the incoming OS fact is "Centos <major version>". 

Otherwise, it's possible the host has a different OS or perhaps the CentOS major version has changed, say from "7" to "8".

Manual test suggested:
- register a CentOS host with `subscription-manager register`
- observer the operating system with either `hammer` or via the UI, it should report "CentOS 7", where the "7" is whatever major version you are using.
- Create a new operating system with a minor version, say "CentOS 7.6"
- update the host as managed, then assign the new operating system with the minor version
- run the `subscription-manager facts --update`
- the reported OS in Katello should not change to the "Centos 7", where the minor version is missing

### DIFF
--- a/app/models/katello/host/subscription_facet.rb
+++ b/app/models/katello/host/subscription_facet.rb
@@ -212,7 +212,38 @@ module Katello
         return if host.build? || rhsm_facts.nil?
         rhsm_facts[:_type] = RhsmFactName::FACT_TYPE
         rhsm_facts[:_timestamp] = Time.now.to_s
+        if ignore_os?(host.operatingsystem, rhsm_facts)
+          rhsm_facts[:ignore_os] = true
+        end
         host.import_facts(rhsm_facts)
+      end
+
+      def self.ignore_os?(host_os, rhsm_facts)
+
+        if host_os.nil?
+          return false
+        end
+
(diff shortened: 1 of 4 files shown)


app/models/katello/host/subscription_facet.rb
```diff
@@ -291,6 +268,45 @@ def remove_subscriptions(pools_with_quantities)
         ForemanTasks.sync_task(Actions::Katello::Host::RemoveSubscriptions, self.host, pools_with_quantities)
       end
 
+      def self.find_host(facts, organization)
```
jturel: don't go undoing all my hard work now ;)

jjeffers:
The rebase was botched.

PR #8272, opened 2019-08-13
Fixes #27614 - Add Sub page loads no resource

sseelam2:
Changes in this PR -
1. Filtered the subscriptions that are displayed when "Add Subscriptions" button is clicked on the Subsciptions page
2. Removed hyperlinks from subscription name because we locally do not have the subscription information that are pulled from the customer portal. We could redirect to the Red Hat Customer Portal.
Any feedback would be appreciated. 

To check this PR:
1. Navigate to "Subscriptions" under "Content"
2. Import a manifest
3. Click on "Add Subscriptions" 
4. The Subscriptions that are displayed should not contain the Subscriptions that are already added. That is this page should only contain the new subscriptions that are loaded from the Red Hat Customer Portal.
5. Subscription names should not be hyperlinked.

### DIFF
--- a/webpack/scenes/Subscriptions/UpstreamSubscriptions/UpstreamSubscriptionsTableSchema.js
+++ b/webpack/scenes/Subscriptions/UpstreamSubscriptions/UpstreamSubscriptionsTableSchema.js
@@ -1,7 +1,6 @@
 import React from 'react';
 import { FormGroup, FormControl, ControlLabel, HelpBlock } from 'react-bootstrap';
 import { translate as __ } from 'foremanReact/common/I18n';
-import { urlBuilder } from 'foremanReact/common/urlHelpers';
 import {
   headerFormatter,
   cellFormatter,
@@ -32,7 +31,7 @@ export const columns = (controller, selectionController) => [
       formatters: [
         (value, { rowData }) => (
           <td>
-            <a href={urlBuilder('subscriptions', '', rowData.id)}>
+            <a href={`https://access.redhat.com/management/subscriptions/${rowData.subscription_id}`} rel="noopener noreferrer" target="_blank">
               {rowData.product_name}
             </a>
           </td>


sseelam2:
[test katello]

jturel:
@sseelam2 hey, I thought about this a little more and I'm leaning going in a slightly different direction:

1. don't filter the results. this will preserve the existing behavior of the page
2. rather than remove the links,  redirect to the portal as you mentioned

(2) likely means a change in UpstreamSubscriptionsController to add the subscription ID field since it's not already there (at least I don't think it is). Once we have it we can make the url: https://access.redhat.com/management/subscriptions/:subscription_id

Let me know if you have any concerns about that approach

sseelam2:
@jturel Yeah. Your suggestions totally make sense. It's better to provide the user with subscription information than just remove the links. I have pushed the branch with requested changes and it's ready for a retest.

webpack/scenes/Subscriptions/UpstreamSubscriptions/UpstreamSubscriptionsTableSchema.js
```diff
@@ -32,7 +31,7 @@ export const columns = (controller, selectionController) => [
       formatters: [
         (value, { rowData }) => (
           <td>
-            <a href={urlBuilder('subscriptions', '', rowData.id)}>
+            <a href={`https://access.redhat.com/management/subscriptions/${rowData.subscription_id}`}>
```
jturel: I like it a lot. I think the last thing to do is make this open in a new window

sseelam2: @jturel Yes. That sounds better. Pushed the changes.

jturel:
@sseelam2 eslint is complaining :sob: 

```
34:103  error  Using target="_blank" without rel="noopener noreferrer" is a security risk: see https://mathiasbynens.github.io/rel-noopener  react/jsx-no-target-blank
```

sseelam2:
> @sseelam2 eslint is complaining 
> 
> ```
> 34:103  error  Using target="_blank" without rel="noopener noreferrer" is a security risk: see https://mathiasbynens.github.io/rel-noopener  react/jsx-no-target-blank
> ```

@jturel I used included rel="noopener" and pushed the branch. What is the Redmine issue about?

jturel:
> > @sseelam2 eslint is complaining
> > ```
> > 34:103  error  Using target="_blank" without rel="noopener noreferrer" is a security risk: see https://mathiasbynens.github.io/rel-noopener  react/jsx-no-target-blank
> > ```
> 
> @jturel I included rel="noopener" and pushed the branch. Hopefully that helps. What is the Redmine issue about?

You also need 'noreferrer' so maybe on the next push it will work. Not sure why it gave an internal server error

PR #7961, opened 2019-02-04
Fixes #25978 - Debian Errata support

m-bucher:
Add support for Debian/Ubuntu errata using external service, which:

1. parses Debian Security Announcements (DSA) and Ubuntu Security Notices (USN): https://github.com/ATIX-AG/errata_parser

1. serves the resulting errata: https://github.com/ATIX-AG/errata_server

PR #8274, opened 2019-08-13
Fixes #26975 - Errata/package repository filter

sjha4:
**To Test**

1. Create a yum repository "Repo1" with errata content. (ex: https://fedorapeople.org/groups/katello/fakerepos/zoo)
2. Create a yum repository "Repo2" without errata content. (ex: https://repos.fedorapeople.org/repos/pulp/pulp/fixtures/rpm-with-vendor/)
3. Sync Repo1 and Repo2
4. Go to Content > Errata > 
5. In the repository filter, you should see only Repo1.
Earlier, you would see both Repo1 and Repo2.

You can also create an empty yum repo and verify that it doesn't show up in either Errata or Packages page repository filter.

### DIFF
--- a/app/controllers/katello/api/v2/repositories_controller.rb
+++ b/app/controllers/katello/api/v2/repositories_controller.rb
@@ -90,6 +90,7 @@ module Katello
     param :description, String, :desc => N_("description of the repository")
     param :available_for, String, :desc => N_("interpret specified object to return only Repositories that can be associated with specified object.  Only 'content_view' & 'content_view_version' are supported."),
           :required => false
+    param :with_content, RepositoryTypeManager.enabled_content_types, :desc => N_("only repositories having at least one of the specified content type ex: rpm , erratum")
     param_group :search, Api::V2::ApiController
     add_scoped_search_description_for(Repository)
     def index
@@ -119,6 +120,7 @@ module Katello
 
     def index_relation
       query = Repository.readable
+      query = query.with_content(params[:with_content]) if params[:with_content]
       query = index_relation_product(query)
       query = query.with_type(params[:content_type]) if params[:content_type]
       query = query.where(:root_id => RootRepository.where(:name => params[:name])) if params[:name]
--- a/app/models/katello/repository.rb
+++ b/app/models/katello/repository.rb
@@ -123,6 +123,7 @@ module Katello
     scope :archived, -> { where('environment_id is NULL') }
     scope :in_published_environments, -> { in_content_views(Katello::ContentView.non_default).where.not(:environment_id => nil) }
     scope :order_by_root, ->(attr) { joins(:root).order("#{Katello::RootRepository.table_name}.#{attr}") }
+    scope :with_content, ->(content) { joins(Katello::RepositoryTypeManager.find_content_type(content).model_class.repository_association_class.name.demodulize.underscore.pluralize.to_sym).distinct }
 
     scoped_search :on => :name, :relation => :root, :complete_value => true
     scoped_search :rename => :product, :on => :name, :relation => :product, :complete_value => true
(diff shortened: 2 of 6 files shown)


jturel:
I like this a lot. I worked with ATIX folks on something similar on another open PR: https://github.com/Katello/katello/pull/7961 I'll ask that they update to use this approach once merged

app/models/katello/repository.rb
```diff
@@ -123,6 +123,7 @@ class Repository < Katello::Model
     scope :archived, -> { where('environment_id is NULL') }
     scope :in_published_environments, -> { in_content_views(Katello::ContentView.non_default).where.not(:environment_id => nil) }
     scope :order_by_root, ->(attr) { joins(:root).order("#{Katello::RootRepository.table_name}.#{attr}") }
+    scope :with_content, ->(content) { where(:id => Katello::RepositoryTypeManager.find_content_type(content).model_class.repository_association_class.select(:repository_id)) }
```
jturel: I think it could be possible to do a join here which is probably better than a subquery. What do you think?

sjha4: @jturel : How does this look to you:
joins(Katello::RepositoryTypeManager.find_content_type(content).model_class.repository_association_class.name.demodulize.underscore.pluralize.to_sym)

app/controllers/katello/api/v2/repositories_controller.rb
```diff
@@ -90,6 +90,7 @@ def custom_index_relation(collection)
     param :description, String, :desc => N_("description of the repository")
     param :available_for, String, :desc => N_("interpret specified object to return only Repositories that can be associated with specified object.  Only 'content_view' & 'content_view_version' are supported."),
           :required => false
+    param :with_content, String, :desc => N_("only repositories having at least one of the specified content type ex: rpm , erratum")
```
jturel: Do we have any precedent in the code that can autogenerate the available types for the api docs? I seem to recall something but not sure where

app/models/katello/repository.rb
```diff
@@ -123,6 +123,7 @@ class Repository < Katello::Model
     scope :archived, -> { where('environment_id is NULL') }
     scope :in_published_environments, -> { in_content_views(Katello::ContentView.non_default).where.not(:environment_id => nil) }
     scope :order_by_root, ->(attr) { joins(:root).order("#{Katello::RootRepository.table_name}.#{attr}") }
+    scope :with_content, ->(content) { joins(Katello::RepositoryTypeManager.find_content_type(content).model_class.repository_association_class.name.demodulize.underscore.pluralize.to_sym) }
```
jturel: This looks good - but it needs a .distinct :+1:

app/controllers/katello/api/v2/repositories_controller.rb
sjha4: Right..I thought about that but couldn't see a straightforward way to get all content_types..I'll think about this a little..

app/models/katello/repository.rb
sjha4: Added..

app/controllers/katello/api/v2/repositories_controller.rb
```diff
@@ -90,6 +90,7 @@ def custom_index_relation(collection)
     param :description, String, :desc => N_("description of the repository")
     param :available_for, String, :desc => N_("interpret specified object to return only Repositories that can be associated with specified object.  Only 'content_view' & 'content_view_version' are supported."),
           :required => false
+    param :with_content, RepositoryTypeManager.enabled_content_types, :desc => N_("only repositories having at least one of the specified content type ex: rpm , erratum")
     param_group :search, Api::V2::ApiController
```
sjha4: Added allowed values here. API looks like: 
with_content

PR #8276, opened 2019-08-14
Fixes #27248 - introduce separate queue for host-related tasks

adamruzicka:
This addresses comments from https://github.com/Katello/katello/pull/8200


jturel:
@adamruzicka thanks for moving this forward! can you rename your commit to "Refs #..." ? I'm not sure if using two "Fixes # ..." on the same PR will affect any other tooling, so let's be safe

adamruzicka:
No problem, done.

PR #8278, opened 2019-08-14
Fixes #26914 - ingore os fact if applicable

jjeffers:
This PR addresses a long standing issue concerning distribution facts sent by hosts.

At some point CentOS stopped reporting minor versions and the fact parser considers a missing minor version as a fact to be ignored. This has the side effect of showing a "blank" when a CentOS host is registered with Katello.

If not ignored, then hosts with an updated operating system fact would revert to the minorless operating system, created during the fact parsing. This might lead to hosts "reverting" to the wrong version.

The compromise is to ignore operating system fact updates if the current host has an operating system with a major version and the incoming OS fact is "Centos ".

Otherwise, it's possible the host has a different OS or perhaps the CentOS major version has changed, say from "7" to "8".

1. Manual test suggested:
- register a CentOS host with subscription-manager register
- observe the operating system with either hammer or via the UI, it should report "CentOS 7", where the "7" is whatever major version you are using.
- Create a new operating system with a minor version, say "CentOS 7.6"
- update the host as managed, then assign the new operating system with the minor version
- run the subscription-manager facts --update
- the reported OS in Katello should not change to the "Centos 7", where the minor version is missing

2. Manual test suggested:
- register a CentOS host with subscription-manager register
- observe the operating system with either hammer or via the UI, it should report "CentOS 7", where the "7" is whatever major version you are using - in other words, it's not blank or missing in the hosts summary

### DIFF
--- a/test/models/host/subscription_facet_test.rb
+++ b/test/models/host/subscription_facet_test.rb
@@ -15,6 +15,15 @@ module Katello
                                      :lifecycle_environment => library, :organization => org)
     end
     let(:subscription_facet) { host.subscription_facet }
+    let(:centos_76) do
+      FactoryBot.create(:operatingsystem,
+        :with_os_defaults, :with_associations,
+        name: 'CentOS',
+        major: 7,
+        minor: 6,
+        type: "Redhat",
+        title: "CentOS 7.6")
+    end
   end
 
   class SubscriptionFacetSystemPurposeTest < SubscriptionFacetBase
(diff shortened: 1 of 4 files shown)


jlsherrill:
This looks good!  I will do some testing.

test/models/host/subscription_facet_test.rb
```diff
@@ -215,6 +215,71 @@ def test_update_foreman_facts
       assert_includes values.map(&:name), '_timestamp'
     end
 
+    def test_update_foreman_facts_with_centos_no_minor_version
```
jturel: nitpick: our unwritten convention is test_<method_name>_<use_case> so I'd prefer `test_update_facts_with_centos_no_minor_version`

test/models/host/subscription_facet_test.rb
```diff
@@ -215,6 +215,71 @@ def test_update_foreman_facts
...
+      host.operatingsystem = ::Operatingsystem.create(
```
jturel: Take a look and see if we have a fixture we could use for the operatingsystem, or perhaps a factorybot factory (it might exist in foreman)

jjeffers: There are foreman fixtures/factories, it appears.

PR #8279, opened 2019-08-14
[CP] Katello 3.12.3 Cherrypicks from Toolbelt

chris1984:


chris1984:
@jturel here are the cherrypicks for 3.12.3 tests are green :)

jturel:
@chris1984 nice! I looked through the 3.13 release to see if there was anything else worth bringing over and nothing jumped out. I think this is good.

PR #8286, opened 2019-08-19
Fixes #27532 - Sync plan api docs

sjha4:


chris1984:
Looks good @sjha4 ACK

PR #8288, opened 2019-08-20
Fixes #27667 - Add register client directive

xprazak2:
This extracts the last step for client registration
into a directive, which can be overriden more easily
than a template when modified instructions for client
registration steps are needed.

### DIFF
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/views/register-client.html
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/views/register-client.html
@@ -1,0 +1,13 @@
+<li>
+  <p translate>Install client package:</p>
+
+  <p translate>To report package & errata information:</p>
+  <pre><code>yum -y install katello-host-tools</code></pre>
+
+  <p translate>To optionally report tracer information:</p>
+  <pre><code>yum -y install katello-host-tools-tracer</code></pre>
+
+  <p translate>For remote actions and reporting package & errata information:</p>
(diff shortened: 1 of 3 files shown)


engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/views/register-client.html
```diff
@@ -0,0 +1,13 @@
...
+  <p translate>To report package & errata information:</p>
+  <pre><code>yum -y install katello-host-tools</code></pre>
+
+  <p translate>To optionally report tracer information:</p>
+  <pre><code>yum -y install katello-host-tools-tracer</code></pre>
+
+  <p translate>For remote actions and reporting package & errata information:</p>
+  <pre><code>yum -y install katello-agent</code></pre>
+  <p translate>Learn more about these packages in the <a href="https://theforeman.org/plugins/katello/">Documentation</a> </p>
+</li>
```
sjha4: As per the issue report, do we also want to add:
```
If you want to use client tools such as katello-agent or puppet, the satellite-tools repository will need to be enabled

subscription-manager repos --enable=rhel-\*-satellite-tools-\*-rpms
```

xprazak2: The satellite-tools repo is not really relevant for upstream, this change makes it possible to override/modify the registration steps.

The instructions to enable repo can then be easily added in downstream branding.

PR #8289, opened 2019-08-20
Refs #27643 - Revert registry URL change

sseelam2:


jturel:
Marked DONTMERGE until we know for sure this needs to be reverted

jlsherrill:
@jturel can we remove the DONOTMERGE now?

PR #8290, opened 2019-08-21
Fixes #27617 - purge orphaned repo versions

jjeffers:
This PR introduces support for deleting orphaned repository versions in Pulp3 servers.

Currently the Katello rake task that deletes orphaned content doesn't handle these orphaned repository version.

An example scenario:
Create a file repository with an upstream URL. Synchronize that repository. Then, change the upstream URL and re-sryncrhonize. The pulp server will now have 2 repository versions, although only the latest version is reflected in the repository reference in Katello. The first version is not referenced and therefore an orphan.

After applying this PR, the rake task `katello:delete_orphaned_content` will identify and delete orphaned repository versions.

(A follow-on PR will introduce functionality to purge pulp3 mirrors of orphaned repositories.)

### DIFF
--- a/app/lib/actions/pulp3/repository/delete_orphan_repository_versions.rb
+++ b/app/lib/actions/pulp3/repository/delete_orphan_repository_versions.rb
@@ -1,0 +1,16 @@
+module Actions
+  module Pulp3
+    module Repository
+      class DeleteOrphanRepositoryVersions < Pulp3::AbstractAsyncTask
+        def plan(smart_proxy)
+          plan_self(:smart_proxy_id => smart_proxy.id)
+        end
+
+        def run
+          smart_proxy = SmartProxy.find(input[:smart_proxy_id])
+          output[:pulp_tasks] = ::Katello::Pulp3::Repository.delete_orphan_repository_versions(smart_proxy)
+        end
+      end
+    end
+  end
+end
--- a/app/services/katello/pulp3/repository.rb
+++ b/app/services/katello/pulp3/repository.rb
@@ -210,12 +210,66 @@ module Katello
         repositories_api.update(repository_reference.try(:repository_href), name: backend_object_name)
       end
 
-      def list(args)
-        repositories_api.list(args).results
+      def list(options)
+        repositories_api.list(options).results
       end
 
-      def self.list(smart_proxy, args)
-        ::Katello::Pulp3::Repository.new(nil, smart_proxy).list(args)
+      def self.versions_list(smart_proxy, href, options)
+        repository_versions_api = ::Katello::Pulp3::Repository.new(nil, smart_proxy).repository_versions_api
+        fetch_from_list { |page_opts| repository_versions_api.list(href, page_opts.merge(options)) }
+      end
+
+      def self.list(smart_proxy, options)
+        repositories_api = ::Katello::Pulp3::Repository.new(nil, smart_proxy).repositories_api
+        fetch_from_list do |page_opts|
+          repositories_api.list(page_opts.merge(options))
+        end
+      end
+
+      def self.fetch_from_list
+        page_size = SETTINGS[:katello][:pulp][:bulk_load_size]
+        page_opts = { "offset" => 0, limit: page_size }
+        response = {}
+
+        results = []
+
+        loop do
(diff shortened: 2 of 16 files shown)


jlsherrill:
are you planning on handling pulp mirrors as part of this?

jjeffers:
Mirrors: yes, in this PR, but I wanted to talk to you first about it.

jjeffers:
[test katello]

jjeffers:
> Mirrors: yes, in this PR, but I wanted to talk to you first about it.

Decided to defer mirror work for a follow on PR.

jjeffers:
[test katello]

jjeffers:
[test katello]

jjeffers:
[test katello]

app/services/katello/pulp3/repository.rb
```diff
@@ -218,6 +218,33 @@ def self.list(smart_proxy, args)
         ::Katello::Pulp3::Repository.new(nil, smart_proxy).list(args)
       end
 
+      def self.repository_versions(smart_proxy)
+        current_repos = list(smart_proxy, {})
```
jlsherrill: i don't think this would work across pages, would it?

app/services/katello/pulp3/repository.rb
```diff
@@ -218,6 +218,33 @@ def self.list(smart_proxy, args)
...
+            list(href, {}).results.collect(&:_href)
+        end
+
+        version_hrefs.flatten.uniq
+      end
+
+      def self.orphan_repository_versions(smart_proxy)
+        version_hrefs = repository_versions(smart_proxy)
+
+        version_hrefs.flatten.reject { |href| ::Katello::Repository.find_by(version_href: href) }
```
jlsherrill: would something like 
```
version_hrefs - ::Katello::Repository.find_by(version_href: version_hrefs).pluck(:version_href)
```
be more efficient?

app/services/katello/pulp3/repository.rb
jjeffers: You know, you are right, it would not work across pages. Good point.

lib/katello/tasks/delete_orphaned_content.rake
```diff
@@ -16,6 +16,11 @@ namespace :katello do
 
   def remove_orphan(proxy)
     ForemanTasks.async_task(Actions::Katello::CapsuleContent::RemoveOrphans, proxy)
+    if proxy.pulp3_enabled? && proxy == SmartProxy.pulp_master
+      ForemanTasks.async_task(
+        Actions::Pulp3::Orchestration::Repository::RemoveOrphans, proxy)
+    end
```
jlsherrill: would it make sense to put this logic inside of Actions::Katello::CapsuleContent::RemoveOrphans ?

jjeffers: I wasn't sure: I could see that it could, but it's not just for capsule operations. I am totally OK with moving it.

jlsherrill: i think we should move it mainly to reduce the # of tasks that are kicked off

app/lib/actions/pulp3/repository/delete_orphan_repository_versions.rb
```diff
@@ -0,0 +1,16 @@
...
+  module Pulp3
+    module Repository
+      class DeleteOrphanRepositoryVersions < Pulp3::Abstract
+        def plan(smart_proxy)
+          plan_self(:smart_proxy_id => smart_proxy.id)
+        end
+
+        def run
+          smart_proxy = SmartProxy.find(input[:smart_proxy_id])
+          ::Katello::Pulp3::Repository.delete_orphan_repository_versions(smart_proxy)
```
jlsherrill: we probably want this to be an 'abstract_async_task' and have this delete_orphan_repository_versions() method return a list of tasks to track.'

app/services/katello/pulp3/repository.rb
jlsherrill: Looks like its still not being paged?

jjeffers:
[test katello]

PR #8291, opened 2019-08-22
Fixes #25876 - Return 422 on a locked task

xprazak2:
When trying to delete the same org
simultaneously multiple times,
we may hit lock on a task.
In that case, return 422 rather than 500.


xprazak2:
I no longer get the PG::ForeignKeyViolation as the issue in Redmine describes, it is prevented by the task lock.

xprazak2:
I found out the issue is elsewhere and will require a different fix, closing.

PR #8295, opened 2019-08-23
Fixes #26226 - Filter Module Streams on CV

parthaa:
This commit adds UI, API and publish bindings for filtering module
streams in a content view.
Check out https://partha.fedorapeople.org/CVModuleFilters/ for the screen shots

### DIFF
--- a/app/controllers/katello/api/v2/content_view_filter_rules_controller.rb
+++ b/app/controllers/katello/api/v2/content_view_filter_rules_controller.rb
@@ -34,6 +34,11 @@ module Katello
     param :end_date, String, :desc => N_("erratum: end date (YYYY-MM-DD)")
     param :types, Array, :desc => N_("erratum: types (enhancement, bugfix, security)")
     param :date_type, String, :desc => N_("erratum: search using the 'Issued On' or 'Updated On' column of the errata. Values are 'issued'/'updated'")
+    param :module_streams, Array, :desc => N_("Array of module streams to add") do
+      param :name, String, :desc => N_("Module Streams name"), :required => true
+      param :stream, String, :desc => N_("Module Stream's stream"), :required => true
+    end
+
     def create
       rule_clazz = ContentViewFilter.rule_class_for(@filter)
 
@@ -44,6 +49,11 @@ module Katello
       rules += (rule_params[:errata_ids] || []).map do |errata_id|
         rule_clazz.create!(rule_params.except(:errata_ids)
           .merge(filter: @filter, errata_id: errata_id))
+      end
+      if rule_params[:module_streams]
+        rules += rule_params[:module_streams].map do |module_stream|
+          rule_clazz.create!(filter: @filter, name: module_stream[:name], stream: module_stream[:stream])
+        end
       end
 
       if rules.empty?
@@ -125,7 +135,7 @@ module Katello
       @rule_params ||= params.fetch(:content_view_filter_rule, {}).
             permit(:uuid, :version, :min_version, :max_version, :architecture,
                     :errata_id, :start_date, :end_date, :date_type,
-                    :types => [], :errata_ids => [], name: [])
+                    :types => [], :module_streams => [:name, :stream], :errata_ids => [], name: [])
     end
 
     def process_errata_ids(select_all_params)
--- a/app/controllers/katello/api/v2/content_view_filter_rules_controller.rb
+++ b/app/controllers/katello/api/v2/content_view_filter_rules_controller.rb
@@ -34,7 +34,7 @@ module Katello
     param :end_date, String, :desc => N_("erratum: end date (YYYY-MM-DD)")
     param :types, Array, :desc => N_("erratum: types (enhancement, bugfix, security)")
     param :date_type, String, :desc => N_("erratum: search using the 'Issued On' or 'Updated On' column of the errata. Values are 'issued'/'updated'")
-    param :module_stream_ids, String, :desc => N_("module stream ids")
+    param :module_stream_ids, Array, :desc => N_("module stream ids")
     def create
       rule_clazz = ContentViewFilter.rule_class_for(@filter)
 
--- a/app/controllers/katello/api/v2/content_view_filter_rules_controller.rb
+++ b/app/controllers/katello/api/v2/content_view_filter_rules_controller.rb
@@ -34,11 +34,7 @@ module Katello
     param :end_date, String, :desc => N_("erratum: end date (YYYY-MM-DD)")
     param :types, Array, :desc => N_("erratum: types (enhancement, bugfix, security)")
     param :date_type, String, :desc => N_("erratum: search using the 'Issued On' or 'Updated On' column of the errata. Values are 'issued'/'updated'")
-    param :module_streams, Array, :desc => N_("Array of module streams to add") do
-      param :name, String, :desc => N_("Module Streams name"), :required => true
-      param :stream, String, :desc => N_("Module Stream's stream"), :required => true
-    end
-
+    param :module_stream_ids, String, :desc => N_("module stream ids")
     def create
       rule_clazz = ContentViewFilter.rule_class_for(@filter)
 
@@ -50,9 +46,10 @@ module Katello
         rule_clazz.create!(rule_params.except(:errata_ids)
           .merge(filter: @filter, errata_id: errata_id))
       end
-      if rule_params[:module_streams]
-        rules += rule_params[:module_streams].map do |module_stream|
-          rule_clazz.create!(filter: @filter, name: module_stream[:name], stream: module_stream[:stream])
+      if rule_params[:module_stream_ids]
+        rules += (rule_params[:module_stream_ids] || []).map do |module_stream_id|
+          rule_clazz.create!(rule_params.except(:module_stream_ids)
+            .merge(filter: @filter, module_stream_id: module_stream_id))
         end
       end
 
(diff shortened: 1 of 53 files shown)


omkarkhatavkar:
@parthaa I tried quick glance today for UI, found some issues mentioned below. 

Module Stream Link with not shown here. 
Repositories for Filter  animal_repo_module_stream_filter

Module Stream Count shown wrong after all modules are applied in exclude filter with having Solve Dependencies is 'Yes'. This is not reproducible if  Solve Dependencies is 'No'.
Versions for Content View  Content_view_2

We can get rid of this message as we have module stream filters 
Package Filter  package filter

Currently, Publishing content views is failing with the below steps 
1. Add all module streams in the filter and publish the content view 
2. Now remove all modules from the filter and publish the filter 
Versions for Content View  Content_view_3
centos7 katello devel okhatavk example com foreman_tasks tasks 002b9cba fa87 45ca 98dc 8a2bf3a5b116

Minor UI Issue, while removing module stream list we are showing multiple success message not similar while adding. we should stick only UI success message?   
Content Views_1

I will be testing this further in detailed.

omkarkhatavkar:
Module Stream Filter Create hangs in UI, not happens with other filters  

New Filter for Content View  New_Testing_Content_View_1

omkarkhatavkar:
@parthaa One More thing, I was not able to create the content view filter rule. tried some options, can you please that as well.   

```
Option details:
  Following parameters accept format defined by its schema (bold are required):

  --module-streams    "name=string\,stream=string, ... "
hammer> 

hammer> content-view filter rule create --name "exclude duck module stream" --content-view-id 9 --content-view-filter-id 24 --module-streams "name='duck', stream='0'"
Could not create the filter rule:
  Missing arguments for 'module_streams[0][stream]', 'module_streams[1][name]'
hammer> content-view filter rule create --name "exclude duck module stream" --content-view-id 9 --content-view-filter-id 24 --module-streams "name=duck, stream=0"
Could not create the filter rule:
  Missing arguments for 'module_streams[0][stream]', 'module_streams[1][name]'

```

parthaa:
> ```
> Option details:
>   Following parameters accept format defined by its schema (bold are required):
> 
>   --module-streams    "name=string\,stream=string, ... "
> hammer> 
> 
> hammer> content-view filter rule create --name "exclude duck module stream" --content-view-id 9 --content-view-filter-id 24 --module-streams "name='duck', stream='0'"
> Could not create the filter rule:
>   Missing arguments for 'module_streams[0][stream]', 'module_streams[1][name]'
> hammer> content-view filter rule create --name "exclude duck module stream" --content-view-id 9 --content-view-filter-id 24 --module-streams "name=duck, stream=0"
> Could not create the filter rule:
>   Missing arguments for 'module_
> ```

@omkarkhatavkar  I think this should be a separate hammer bug. The --module-streams parameter in the api creates "multiple" rules - one for each module-stream. 
We should probably tweak the hammer call to have flags for module name and stream. Something like
```
hammer> content-view filter rule create --name "exclude duck module stream" --content-view-id 9 --content-view-filter-id 24 --module-name='duck'  --module-stream "0" 

```

omkarkhatavkar:
> @omkarkhatavkar I think this should be a separate hammer bug. The --module-streams parameter in the api creates "multiple" rules - one for each module-stream.
> We should probably tweak the hammer call to have flags for module name and stream. Something like
> 
> ```
> hammer> content-view filter rule create --name "exclude duck module stream" --content-view-id 9 --content-view-filter-id 24 --module-name='duck'  --module-stream "0" 
> ```

Make Sense I will raise the new bug for that

omkarkhatavkar:
@parthaa I saw today another issue really weird one
Steps :
1. added all module streams in the filter -> publish the filter 
2. remove all module streams from the filter -> publish 
3. Change the Content view Depedancy => Solve 'Yes'   
4. Now try to all module stream filter 
ezgif com-video-to-gif (3)

omkarkhatavkar:
> Module Stream Link with not shown here.
### PASS
> Module Stream Count shown wrong after all modules are applied in exclude filter with having Solve Dependencies is 'Yes'. This is not reproducible if Solve Dependencies is 'No'.
 ### Still Failing 
> We can get rid of this message as we have module stream filters
 ### PASS
> Currently, Publishing content views is failing with the below steps
> 1. Add all module streams in the filter and publish the content view
> 2. Now remove all modules from the filter and publish the filter

> Module Stream Filter Create hangs in UI, not happens with other filters
### PASS
> Minor UI Issue, while removing module stream list we are showing multiple success message not similar while adding. we should stick only UI success message?
### Can be Ignored

omkarkhatavkar:
@Partha: One change may be required here to make sure that API docs get updated  
https://github.com/Katello/katello/blob/e3e0f9596358449141783e9f206cb67cc6affab3/app/controllers/katello/api/v2/content_view_filters_controller.rb#L37

omkarkhatavkar:
@parthaa Facing this issue, can you please retry. 
Module Stream Filter  module stream filter

omkarkhatavkar:
> @parthaa Facing this issue, can you please retry.
> Module Stream Filter module stream filter

ignore it was db: migrate was not run

omkarkhatavkar:
Products

parthaa:
Products

omkarkhatavkar:
**Case 1:** 
1. Add Module Streams Exclude filter on duck and kangaroo and Add include filter on Errata filter with duck showing below behaviour ( screenshot). We are missing to show walrus is this expected as we including errata and then excluding module streams?       
Versions for Content View  Content_view_2

app/controllers/katello/api/v2/content_view_filter_rules_controller.rb
```diff
@@ -34,6 +34,7 @@ def index_relation
     param :end_date, String, :desc => N_("erratum: end date (YYYY-MM-DD)")
     param :types, Array, :desc => N_("erratum: types (enhancement, bugfix, security)")
     param :date_type, String, :desc => N_("erratum: search using the 'Issued On' or 'Updated On' column of the errata. Values are 'issued'/'updated'")
+    param :module_stream_ids, String, :desc => N_("module stream ids")
```
omkarkhatavkar: is this will String or Array similar like errata_ids ?

parthaa: Good catch its array

parthaa:
> @parthaa One More Question
> **Case 1:**
> 
> 1. Add Module Streams Exclude filter on duck and kangaroo and Add include filter on Errata filter with duck showing below behaviour ( screenshot). We are missing to show walrus is this expected as we including errata and then excluding module streams?
>    Versions for Content View  Content_view_2

Is Depsolve turned on ?

omkarkhatavkar:
> Is Depsolve turned on ?
@parthaa  No. You can try this out. Not even sure how much this will practically happen

parthaa:
> > Is Depsolve turned on ?
> > @parthaa  No. You can try this out. Not even sure how much this will practically happen

Ok Include filters are first run and then the excludes happen on them. So if you included an errata but excluded modules, then the result should be "Modules in the Errata - Modules Excluded"

app/models/katello/module_stream.rb
```diff
@@ -56,6 +56,22 @@ def self.available_for_hosts(hosts)
...
+      return where("1=0") if name_streams.blank?
+      ns = name_streams.shift
+      query = where(name: ns.first, stream: ns.last)
+      name_streams.each do |name, stream|
+        query = query.or(where(name: name, stream: stream))
+      end
+      query
+    end
+
+    def self.uniquify_by_name_streams(name_streams)
```
johnpmitsch: where is this used?

parthaa: Good call. Going to remove it (left over stray method from the name/stream implementation).

omkarkhatavkar:
@parthaa please take a look on added automated tests for this PR here https://github.com/SatelliteQE/robottelo/pull/7339 (whenever you get a chance)

johnpmitsch:
@parthaa I wonder if we need the description in the table of module streams in the filter? They can be lengthy and make the list very long and busy:

here

johnpmitsch:
@parthaa it looks like dependency solving does actually affect module streams?
here

I'm guessing this is intended, if so, I think we should change some of the dep solving help text in the content view details page, new CV page, and settings page. It currently says:
`This will solve RPM dependencies on every publish of this Content View. `, we should mention module streams or rephrase this

app/models/katello/content_view_module_stream_filter.rb
```diff
@@ -0,0 +1,26 @@
...
+      return if module_stream_rules.blank?
+
+      module_stream_ids = module_stream_rules.map(&:module_stream_id)
+      module_streams_in(module_stream_ids) unless module_stream_ids.empty?
+    end
+
+    private
+
+    def module_stream_arel
+      ::Katello::ModuleStream.arel_table
```
johnpmitsch: Why use arel in this file, does AR not provide the capability to do what you need?

parthaa: So this query is only  generating the where clause, instead of hard coding "(item in 111,24,5)" as string I feel arel table is a better approach

parthaa:
[test katlelo]

parthaa:
[test katello]

parthaa:
> Code and functionality looks good, just a few comments and open questions!

updated

parthaa:
[test katello]

app/models/katello/content_view_module_stream_filter.rb
```diff
@@ -0,0 +1,26 @@
...
+    CONTENT_TYPE = ModuleStream::CONTENT_TYPE
+    has_many :module_stream_rules, :dependent => :destroy, :foreign_key => :content_view_filter_id,
+                             :class_name => "Katello::ContentViewModuleStreamFilterRule"
+
+    validates_lengths_from_database
+
+    def generate_clauses(_repo)
+      return if module_stream_rules.blank?
+
+      module_stream_ids = module_stream_rules.map(&:module_stream_id)
```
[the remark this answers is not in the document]
parthaa: addressed

PR #8296, opened 2019-08-23
Fixes #27739 - Reg. broken when DMI UUID changed after unregister

jturel:
If a host is unregistered and its DMI UUID has changed by the next time it attempts to register (but the hostname is the same) it will not work. It should work so long as the DMI UUID is unique and the host record on the server is not already registered.

**To test**

- register content host to your env
- unregister it (subscription-manager unregister for example)
- change the content host's DMI UUID:

```
echo '{"dmi.system.uuid": "this-is-made-up"}' > /etc/rhsm/facts/uuid.facts
```
- attempt to register and see that it fails stating that the DMI UUID matches another host

- apply this PR and see that the registration succeeds

### DIFF
--- a/test/services/katello/registration_manager_test.rb
+++ b/test/services/katello/registration_manager_test.rb
@@ -34,7 +34,7 @@ module Katello
         def setup
           @org = get_organization
           @klass = Katello::RegistrationManager
-          @host = FactoryBot.create(:host, organization: @org)
+          @host = FactoryBot.create(:host, :with_subscription, organization: @org)
           @uuid_fact_name = RhsmFactName.create(name: 'dmi::system::uuid')
         end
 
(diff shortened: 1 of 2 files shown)


test/services/katello/registration_manager_test.rb
```diff
@@ -89,6 +89,16 @@ def test_existing_host_null_uuid
 
           assert @klass.validate_hosts(hosts, @org, @host.name, 'different-uuid')
         end
+
+        def test_existing_uuid_mistmatch_unregistered
```
timogoebel: typo: `mismatch`?

```suggestion
        def test_existing_uuid_mismatch_unregistered
```

jlsherrill:
[test katello]

bbuckingham:
[test katello]

PR #8297, opened 2019-08-23
Fixes #27395 - Missing CentOS env groups in content view versions

ianballou:
Adds support for copying over environment groups in yum repos.

Note that this only copies over the environment groups.  The environment group information is still not kept track of in Katello.  The environment groups don't appear in the UI and are not installable by package actions.  This implementation should be introduced in a following PR.

To test the patch:

1) Sync CentOS 7 repo http://mirror.centos.org/centos/7/os/x86_64/
2) Add that repo to a content view and publish
3) Subscribe a katello client to that content view and enable the synced repo
4) On the client run 
```yum --disablerepo="*" --enablerepo="<your repo id>" grouplist```

You should see environment groups
```
...
Available Environment Groups:
   Minimal Install
   Compute Node
   Infrastructure Server
   File and Print Server
   Basic Web Server
   Virtualization Host
   Server with GUI
   GNOME Desktop
   KDE Plasma Workspaces
   Development and Creative Workstation
...
```

Alternatively, you could check the repodata/<characters>-comps.xml file in your content view's pulp repo for the environments.

### DIFF
--- a/katello.gemspec
+++ b/katello.gemspec
@@ -40,7 +40,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency "apipie-rails", ">= 0.5.14"
 
   # Pulp
-  gem.add_dependency "runcible", ">= 2.11.0", "< 3.0.0"
+  gem.add_dependency "runcible", ">= 2.12.1", "< 3.0.0"
   gem.add_dependency "anemone"
 
   #pulp3
--- a/app/services/katello/pulp/repository/yum.rb
+++ b/app/services/katello/pulp/repository/yum.rb
@@ -91,7 +91,8 @@ module Katello
           modular_includes = {filters: {unit: { 'filename' => { '$in' => repo.rpms.modular.pluck(:filename) } }}}
           tasks << smart_proxy.pulp_api.extensions.rpm.copy(repo.pulp_id, destination_repo.pulp_id, modular_includes)
 
-          [:srpm, :errata, :package_group, :yum_repo_metadata_file, :distribution, :module, :module_default].each do |type|
+          [:srpm, :errata, :package_group, :package_environment,
+           :yum_repo_metadata_file, :distribution, :module, :module_default].each do |type|
             tasks << smart_proxy.pulp_api.extensions.send(type).copy(repo.pulp_id, destination_repo.pulp_id)
           end
           tasks
--- a/test/fixtures/vcr_cassettes/katello/service/repository/rpm_with_proxy/create_with_global_http_proxy.yml
+++ b/test/fixtures/vcr_cassettes/katello/service/repository/rpm_with_proxy/create_with_global_http_proxy.yml
@@ -2,26 +2,26 @@
 http_interactions:
 - request:
     method: get
-    uri: https://devel.localhost.example.com/pulp/api/v2/repositories/debian_9/?details=true
+    uri: https://centos7-katello-devel-2.cannolo.example.com/pulp/api/v2/repositories/debian_9/?details=true
     body:
       encoding: US-ASCII
       base64_string: ''
     headers:
       Accept:
       - application/json
-      Accept-Encoding:
-      - gzip, deflate
       User-Agent:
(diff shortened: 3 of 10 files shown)

### CODE REVIEW

**Comment by ianballou** [comment]
Don't merge until https://github.com/Katello/runcible/pull/217 is through

**Comment by ianballou** [comment]
[test katello]

**Comment by jturel** [comment]
How can I verify the functionality?

**Comment by ianballou** [comment]
@jturel I've updated the description with how to test

**Comment by ianballou** [comment]
[test katello]

File: katello.gemspec [hunk 1]
```diff
@@ -40,7 +40,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency "apipie-rails", ">= 0.5.14"
 
   # Pulp
-  gem.add_dependency "runcible", ">= 2.11.0", "< 3.0.0"
+  gem.add_dependency "runcible", ">= 2.12.0", "< 3.0.0"
```
jlsherrill: we need to open a packaging pr for the new runcible, i'll do that today

jturel:
@ianballou my environment has runcible-2.12.0 but I didn't check out your PR yet. I synced the centos7 repo and I saw the group info:

```
[root@katello-client vagrant]# yum --disablerepo="*" --enablerepo="Default_Organization_Centos7_Centos7_x86_64" grouplist
Loaded plugins: enabled_repos_upload, fastestmirror, package_upload, product-id, search-disabled-repos, subscription-manager
There is no installed groups file.
Maybe run: yum groups mark convert (see man yum)
Loading mirror speeds from cached hostfile
Available Environment Groups:
   Minimal Install
   Compute Node
   Infrastructure Server
   File and Print Server
   Basic Web Server
   Virtualization Host
   Server with GUI
   GNOME Desktop
   KDE Plasma Workspaces
   Development and Creative Workstation
Available Groups:
   Compatibility Libraries
   Console Internet Tools
   Development Tools
   Graphical Administration Tools
   Legacy UNIX Compatibility
   Scientific Support
   Security Tools
   Smart Card Support
   System Administration Tools
   System Management
```

I can also groupinstall without any problems. Is this expected to work without this Katello change? It doesn't seem like your runcible change would cause this to work on its own. Was it already being synced somehow?

katello.gemspec
jlsherrill: opened pr here: https://github.com/theforeman/foreman-packaging/pull/4056

ianballou:
@jturel I am a little surprised, I'll test it out myself and see what's going on.  Maybe @jlsherrill knows why it's working with only the runcible update?

jturel:
Here's what I did:

**Prior to applying the PR:**

- synced Centos7 to Library
- Content host in Library saw the env. groups (expected current behavior)
- Created CV-A with Centos7 repo, published version 1
- Changed CV of host to CV-A
- Host was unable to see the env. groups (expected current behavior)

***applied the pr***

- published CV-A version 2 (expecting new metadata)
- cleared yum cache on client, could *not* see the env groups (**i think this is incorrect**)

- Created CV-B with Centos7 and published version 1
- Changed CV of my host to CV-B
- Cleared yum cache on client, env groups where there (good, new behavior)

It appears this doesn't work when publishing a new version of an existing CV and I think it's reasonable to expect that it should- what do you think @ianballou  ?

jlsherrill:
i think the reason you didn't seem when publishing a new cv version, but it worked fine with a new cv is the 'check matching content' that doesn't re-clone/generate the yum metadata if no changes are detected.   And with this change we aren't detecting 'changes' with regard to environmental groups.    You can force this by forcing a 'metadata generate' for the version in the lifcycle environment that has the groups but isn't showing them. 

To do this, we'd likely need to add indexing support for environment groups :/     Which is maybe a little more than we want to do in this Pr, but could do in a followup pr?

jturel:
I forgot about that. In that case I'm good with the scope of changes here!

ianballou:
[test katello]

ianballou:
[test katello]

PR #8299, opened 2019-08-25
Fixes #27699 - update @theforeman/vendor to ^1.4.0

sharvit:
This should fix the broken nightly

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
-
 language: node_js
 node_js:
-  - '6.14' # previous LTS, currently in EPEL
-  - '8' # current LTS
-  - '10' # future LTS
+  - '8'
+  - '10'
+  - '12'
 script: ./travis_run_js_tests.sh
--- a/package.json
+++ b/package.json
@@ -25,10 +25,10 @@
   "devDependencies": {
     "@storybook/react": "^3.2.17",
     "@storybook/storybook-deployer": "^2.0.0",
-    "@theforeman/vendor-dev": "^0.1.1",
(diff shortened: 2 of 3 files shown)


tbrisker:
Looks like test failure may be related. We also need the matching change in packaging.

sharvit:
packaging update:
https://github.com/theforeman/foreman-packaging/pull/4035

sharvit:
Jenkins failing because it is running `eslint` on node-6.

Therefore, this PR bring back Travis and run the `eslint` on Travis using node 6-12.
Created a PR in foreman-infra to remove eslint: https://github.com/theforeman/foreman-infra/pull/1116

tbrisker:
i've re-enabled travis on the katello repo, closing and reopening to trigger build for this pr.

sharvit:
Thanks @tbrisker,

tbw, I let `travis` to run the test as well, just because it will still finish before `Jenkins` so "why not?".

tbrisker:
looks like this isn't a matter of node version this time :cry:

sharvit:
Seems like updating `eslint` solves that issue but i'm still trying to figure out what went wrong.

ekohl:
Looks like node 12 still fails on Travis.

sharvit:
The issue is because version 0.2.0 of the vendor added `react-loading-skeleton` which uses an unmatched version of `@babel/types` to what the `babel-eslint` uses.
npm :-1: 

Anyway, I updating `babel-eslint` fixes the issue because it uses a newer version of `@babel/types`.
- Running `eslint` on node-6 won't work anymore.

PR #8305, opened 2019-08-27
Fixes #27718 - cannot remove puppet module from cv by uuid

ianballou:
Test steps:
1. Create a repo with a puppet module
2. Sync the repo
3. Create a content view with the puppet module
4. Remove the puppet module using the hammer command: 
`hammer content-view puppet-module remove --content-view-id=<id> --uuid=<uuid>`


ianballou:
@chris1984 and I discovered that bug exists already in Hammer.  The content view puppet module doesn't have a UUID if its version is Latest.

ianballou:
@chris1984 it still says you've requested changes, is it all set now?

jlsherrill:
thanks @ianballou !

PR #8319, opened 2019-08-28
[WIP]Fixes #19712 - Allow single default capsule

sjha4:
Add validation to allow only one default capsule.

**To test:**

- Update your default capsule with a new name and a new dummy URL.
- Create a new capsule with the original name and original URL. 
- It should error out with the error message: Only one default capsule is allowed

Create a mirror smart proxy against this instance as an additional test to confirm that we can create a mirror smart proxy.

### DIFF
--- a/app/lib/katello/validators/smart_proxy_default_capsule_validator.rb
+++ b/app/lib/katello/validators/smart_proxy_default_capsule_validator.rb
@@ -2,7 +2,10 @@ module Katello
   module Validators
     class SmartProxyDefaultCapsuleValidator < ActiveModel::Validator
       def validate(record)
-        record.errors[:base] << _("Only one default capsule is allowed") if (record.has_feature?('Pulp') && !SmartProxy.with_features('Pulp').blank?)
+        if (record.has_feature?('Pulp') && !SmartProxy.with_features('Pulp').blank? &&
+            !(SmartProxy.with_features('Pulp').count == 1 && SmartProxy.with_features('Pulp').include?(record)))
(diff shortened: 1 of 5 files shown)


sjha4:
[test katello]

app/lib/katello/validators/smart_proxy_default_capsule_validator.rb
```diff
@@ -0,0 +1,9 @@
+module Katello
+  module Validators
+    class SmartProxyDefaultCapsuleValidator < ActiveModel::Validator
+      def validate(record)
+        record.errors[:base] << _("Only one default capsule is allowed") if (record.has_feature?('Pulp') && SmartProxy.pulp_master && SmartProxy.pulp_master != record)
```
jturel: Should `record.pulp_master?` be used instead of  `record.has_feature?('Pulp')` ?

jturel: @sjha4 watcha think?

sjha4: @jturel: So this is a little complicated to do on the katello side alone and this fix only stops updating a second pulp master. We need one small change on foreman side to make it easier to stop creation..I will open a PR in foreman and change this accordingly..
Will mark WIP for now and get back to this..

jturel: Sounds good!

sjha4:
Depends on : https://github.com/theforeman/foreman/pull/7055

sjha4:
Closing this out.Will revisit this.

PR #8322, opened 2019-08-29
Fixes #27608 - add Ansible Coll tags and description

jlsherrill:


sjha4:
One easy way I found is to replace https://github.com/Katello/katello/blob/17c31ff3883ab58ead81da6cb98dbce68c0ef429/webpack/components/Content/Details/ContentDetailInfo.js#L13

with 
`<td>{Array.isArray(details[key]) ? details[key].join(', ') : details[key]}</td>`

PR #8323, opened 2019-08-29
Fixes #27737 - consider proxy capabilities for pulp selector

jlsherrill:


jlsherrill:
[test katello]

jlsherrill:
@sjha4 or @ianballou would one of you mind looking over this once again?

app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -139,14 +139,18 @@ def pulp3_enabled?
...
-        type.pulp3_plugin.present? && pulp3_enabled?
+        pulp3_repository_type_support?(repository.content_type)
       end
 
-      def content_pulp3_support?(content_type)
-        content_type = content_type.is_a?(String) ? content_type : content_type.model_class::CONTENT_TYPE
-        type = Katello::RepositoryTypeManager.find_repository_type content_type
-        type.pulp3_plugin && SmartProxy.pulp_master!.capabilities('Pulp3').try(:include?, type.pulp3_plugin)
+      def pulp3_repository_type_support?(repository_type)
+        repository_type = Katello::RepositoryTypeManager.repository_types[repository_type] if repository_type.is_a?(String)
```
ianballou: Should we be checking for `nil` here? When I tested with an unknown repo type, it errored out on the next line with `NoMethodError (undefined method `pulp3_plugin' for nil:NilClass)`

app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -139,14 +139,18 @@ def pulp3_enabled?
...
+        repository_type.pulp3_plugin.present? && pulp3_enabled? && self.capabilities('Pulp3').try(:include?, repository_type.pulp3_plugin)
+      end
+
+      def pulp3_content_support?(content_type)
+        content_type = RepositoryTypeManager.find_content_type(content_type) if content_type.is_a?(String)
```
ianballou: Same as above, should we check `nil` here?  Might as well error out if it is.

jlsherrill:
@ianballou updated!

ianballou:
The strings for `repository_type` and `content_type` end up being blank since they're being replaced with `nil`.

app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -140,14 +140,22 @@ def pulp3_enabled?
...
+        fail "Cannot find repository type #{repository_type}, is it enabled?" unless repository_type_obj
+        repository_type_obj.pulp3_plugin.present? && pulp3_enabled? && self.capabilities(PULP3_FEATURE).try(:include?, repository_type_obj.pulp3_plugin)
+      end
+
+      def pulp3_content_support?(content_type)
+        content_type_obj = content_type.is_a?(String) ? RepositoryTypeManager.find_content_type(content_type) : content_type
+        fail "Cannot find content type #{content_type}." unless content_type_obj
+
+        found_type = Katello::RepositoryTypeManager.repository_types.values.find { |repo_type| repo_type.content_types.include?(content_type_obj) }
+        fail "Cannot find repository type #{found_type}, is it enabled?" unless found_type
```
ianballou: `found_type` will always be nil here, probably don't have to print it out

app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -140,14 +140,22 @@ def pulp3_enabled?
...
-        type = Katello::RepositoryTypeManager.find_repository_type content_type
-        type.pulp3_plugin && SmartProxy.pulp_master!.capabilities(PULP3_FEATURE).try(:include?, type.pulp3_plugin)
+      def pulp3_repository_type_support?(repository_type)
+        repository_type_obj = repository_type.is_a?(String) ? Katello::RepositoryTypeManager.repository_types[repository_type] : repository_type
+        fail "Cannot find repository type #{repository_type}, is it enabled?" unless repository_type_obj
+        repository_type_obj.pulp3_plugin.present? && pulp3_enabled? && self.capabilities(PULP3_FEATURE).try(:include?, repository_type_obj.pulp3_plugin)
+      end
+
+      def pulp3_content_support?(content_type)
+        content_type_obj = content_type.is_a?(String) ? RepositoryTypeManager.find_content_type(content_type) : content_type
```
ianballou: `RepositoryTypeManager` here doesn't have `Katello::` like the others do.   Still works but I thought I'd point it out.

app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -140,14 +140,22 @@ def pulp3_enabled?
...
-        type.pulp3_plugin.present? && pulp3_enabled?
+        pulp3_repository_type_support?(repository.content_type)
       end
 
-      def content_pulp3_support?(content_type)
-        content_type = content_type.is_a?(String) ? content_type : content_type.model_class::CONTENT_TYPE
-        type = Katello::RepositoryTypeManager.find_repository_type content_type
-        type.pulp3_plugin && SmartProxy.pulp_master!.capabilities(PULP3_FEATURE).try(:include?, type.pulp3_plugin)
+      def pulp3_repository_type_support?(repository_type)
+        repository_type_obj = repository_type.is_a?(String) ? Katello::RepositoryTypeManager.repository_types[repository_type] : repository_type
```
parthaa: Is this better
```repositoy_type =  Katello::RepositoryTypeManager.repository_types[repository_type] if repository_type.is_a?(String)```

jlsherrill: I originally had that, but consider the case where a repository_type string is passed in of an invalid type.  repository_type is set to nil and the line below can't print the repository_type

PR #8324, opened 2019-08-29
Fixes #27738 - View matching content for CV filter rule

johnpmitsch:
Currently, it is difficult to know what packages a content view package filter rule will match until you
publish a Content View. This commit adds the ability to see matching content for a filter rule in the API
and in the UI.

### DIFF
--- a/app/controllers/katello/concerns/api/v2/repository_content_controller.rb
+++ b/app/controllers/katello/concerns/api/v2/repository_content_controller.rb
@@ -10,7 +10,7 @@ module Katello
         before_action :find_optional_organization, :only => [:index, :auto_complete_search]
         before_action :find_environment, :only => [:index, :auto_complete_search]
         before_action :find_content_view_version, :only => [:index, :auto_complete_search]
-        before_action :find_filter, :only => [:index, :auto_complete_search]
+        before_action :find_filter, :find_filter_rule, :only => [:index, :auto_complete_search]
         before_action :find_content_resource, :only => [:show]
       end
 
@@ -23,6 +23,7 @@ module Katello
       param :organization_id, :number, :desc => N_("organization identifier")
       param :content_view_version_id, :number, :desc => N_("content view version identifier")
       param :content_view_filter_id, :number, :desc => N_("content view filter identifier")
+      param :content_view_filter_rule_id, :number, :desc => N_("content view filter rule identifier")
       param :repository_id, :number, :desc => N_("repository identifier")
       param :environment_id, :number, :desc => N_("environment identifier")
       param :ids, Array, :desc => N_("ids to filter content by")
@@ -79,12 +80,20 @@ module Katello
 
         collection = filter_by_repos(repos, collection)
         collection = filter_by_ids(params[:ids], collection) if params[:ids]
+
         @filter = ContentViewFilter.find(params[:filterId]) if params[:filterId]
         if params[:available_for] == "content_view_filter" && self.respond_to?(:available_for_content_view_filter)
           collection = self.available_for_content_view_filter(@filter, collection) if @filter
         else
-          collection = filter_by_content_view_filter(@filter, collection) if @filter
+          # Filtering by the CV filter rule makes filtering by the CV filter redundant, keeping these
+          # exclusive to keep the queries simple.
+          if @filter_rule
+            collection = filter_by_content_view_filter_rule(@filter_rule, collection)
+          elsif @filter
+            collection = filter_by_content_view_filter(@filter, collection)
+          end
         end
+
         collection = self.custom_index_relation(collection) if self.respond_to?(:custom_index_relation)
         collection
       end
@@ -105,9 +114,12 @@ module Katello
         end
       end
 
-      def filter_by_content_view_filter(filter, collection)
-        ids = filter.send("#{singular_resource_name}_rules").pluck(:uuid)
-        filter_by_ids(ids, collection)
+      def filter_by_content_view_filter(_filter, _collection)
+        fail NotImplementedError, "Unsupported content type for content view filter parameter"
+      end
+
+      def filter_by_content_view_filter_rule(_filter, _collection)
+        fail NotImplementedError, "Unsupported content type for content view filter rule parameter"
       end
 
       def repos
@@ -193,6 +205,19 @@ module Katello
         end
       end
 
+      def find_filter_rule
+        filter_rule_id = params[:content_view_filter_rule_id]
+
+        if filter_rule_id
+          @filter_rule = filter_rule_class_name.constantize.find_by(:id => filter_rule_id)
+
+          unless @filter_rule
+            fail HttpErrors::NotFound, _("Couldn't find %{type} Filter with id %{id}") %
+              {:type => resource_name, :id => filter_rule_id}
+          end
+        end
+      end
+
       def resource_class
         "Katello::#{controller_name.classify}".constantize
       end
@@ -203,6 +228,10 @@ module Katello
 
       def filter_class_name
         "Katello::ContentView#{controller_name.classify}Filter"
+      end
+
+      def filter_rule_class_name
+        "Katello::ContentView#{controller_name.classify}FilterRule"
       end
 
       def resource_name(_i18n = true)
--- a/app/lib/katello/api/v2/rendering.rb
+++ b/app/lib/katello/api/v2/rendering.rb
@@ -43,8 +43,7 @@ module Katello
 
           render :template => "katello/api/v2/#{resource_name}/#{action}",
                  :status => status,
-                 :locals => { :object_name => options[:object_name],
-                              :root_name => options[:root_name] },
+                 :locals => options.slice(:object_name, :root_name, :locals),
                  :layout => "katello/api/v2/layouts/#{options[:layout]}"
         end
 
--- a/app/models/katello/content_view_package_filter.rb
+++ b/app/models/katello/content_view_package_filter.rb
@@ -36,13 +36,13 @@ module Katello
     def query_rpms(repo, rule)
       query_name = rule.name.tr("*", "%")
       query = Rpm.in_repositories(repo).non_modular.where("#{Rpm.table_name}.name ilike ?", query_name)
-      if rule.architecture
+      if rule.architecture.present?
         query_arch = rule.architecture.tr("*", "%")
         query = query.where("#{Rpm.table_name}.arch ilike ?", query_arch)
       end
-      if !rule.version.blank?
+      if rule.version.present?
         query = query.search_version_equal(rule.version)
-      elsif !rule.min_version.blank? || !rule.max_version.blank?
+      elsif rule.min_version.present? || rule.max_version.present?
         query = query.search_version_range(rule.min_version, rule.max_version)
       end
       query.pluck("#{Rpm.table_name}.filename")
--- a/app/models/katello/content_view_package_filter.rb
+++ b/app/models/katello/content_view_package_filter.rb
@@ -31,11 +31,18 @@ module Katello
       { 'filename' => { "$in" => package_filenames } } unless package_filenames.empty?
     end
 
-    protected
+    def applicable_rpms
+      Rpm.in_repositories(self.applicable_repos)
+    end
 
     def query_rpms(repo, rule)
+      rpms = Rpm.in_repositories(repo)
+      query_rpms_from_collection(rpms, rule).pluck("#{Rpm.table_name}.filename")
+    end
+
+    def query_rpms_from_collection(collection, rule)
       query_name = rule.name.tr("*", "%")
-      query = Rpm.in_repositories(repo).non_modular.where("#{Rpm.table_name}.name ilike ?", query_name)
+      query = collection.non_modular.where("#{Rpm.table_name}.name ilike ?", query_name)
       if rule.architecture.present?
         query_arch = rule.architecture.tr("*", "%")
         query = query.where("#{Rpm.table_name}.arch ilike ?", query_arch)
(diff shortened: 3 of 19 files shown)


johnpmitsch:
Marking this as a draft because the UI is not finished and I would like to discuss with the team and UX. However, the functionality is all there and reviews are welcome.

johnpmitsch:
@rohoover 
I would like some help designing this UI! Right now, its more of a functional POC. The idea is to show a list of packages in a content view that match a filter rule. Basically the "Show Matching Content" button appears on any created filter rule and you can click it to see the list.

Here are some examples:

A small list of matched packages for the walrus rule
here

A large list of matched packages for the 389-ds-base rule
here

When no packages are matched (the blah rule)
here

What do you think the best way to display these packages is?

ianballou:
I gave this a try, I think it works pretty well.  What was a bit interesting was when there were many matching packages (like when I filtered * on my CentOS 7 repo).  The biggest issue there happened when editing the rule. To hit 'save' I had to scroll all the way to the bottom.  Maybe the Save and Cancel buttons could be placed up above the matching content where the "edit" and "show matching content buttons" are?

johnpmitsch:
> I gave this a try, I think it works pretty well. What was a bit interesting was when there were many matching packages (like when I filtered * on my CentOS 7 repo). The biggest issue there happened when editing the rule. To hit 'save' I had to scroll all the way to the bottom. Maybe the Save and Cancel buttons could be placed up above the matching content where the "edit" and "show matching content buttons" are?

That's interesting, I didn't know you could submit `*`, good to know! The UI part of this is meant to be a proof of concept (sorry if I didn't make that clear). The packages will be displayed in a better way, but I just stuck them in the table for now. I'm thinking we could use a modal with a scrollbar or a collapsible row on the table to display the matching packages. I'm waiting for @rohoover to weigh in to help come up with a more functional UI and I'll update to use what we decide when I get some feedback :) 

But I'm glad it is functioning as expected! Thanks for testing that aspect of things.

ianballou:
Ah yeah that makes more sense.  Well from a usability standpoint it seems to really get the job done! I'll take a closer look at the code too.

ianballou:
Looks like all but one of the errors here are from the Runcible upgrade issues, so those should clear up next time.

johnpmitsch:
[test katello]

app/views/katello/api/v2/content_view_filter_rules/matching_content.json.rabl
```diff
@@ -0,0 +1,7 @@
+object @rule
+
+attributes :id
+
+node(:matching_content) do |_|
+  @rule.class.name == Katello::ContentViewPackageFilterRule.to_s ? @rule.matching_rpms : []
```
johnpmitsch: This will error out  in the controller if an unsupported filter rule type is used, so a blank array should never be shown, but I added it in case anything slips by

johnpmitsch:
Updated to use a modal, but I think I would rather put a parameter to the normal CV filter rule show call to show with matching content rather than use a new endpoint, so keeping this PR as a draft until that happens. Adding a parameter to the CV filter rule will ensure the functionality is added to hammer automatically and the permissions are already set up correctly.

Here are a few screenshots of the modal

Small list:
here

Larger list with scroll:
here

No content found:
here

app/views/katello/api/v2/content_view_filter_rules/matching_content.json.rabl
ianballou: Should we show an error if there is an unsupported filter rule type?

ianballou:
Also I think using the modal does look better!

ianballou:
I noticed an issue: when this PR is checked out, angular pages like Content Views reloads constantly.  It stops when I checkout master and reload the page myself.  I remember @johnpmitsch you mentioned seeing this before.

engines/bastion/app/assets/stylesheets/bastion/overrides.scss
```diff
@@ -77,6 +77,7 @@ html .toolbar-pf.table-view-pf-toolbar-external .toolbar-pf-filter {
 .modal-body {
   overflow-x: hidden;
   overflow-y: scroll;
+  max-height: 75vh; // prevent modal from going below current view causing two scrollbars
```
johnpmitsch: The modal will go below the current viewport height causing two scrollbars and part of it to be hidden. This will limit the body of it to 75% of the viewport and the modal will be fixed. The user can then scroll just the modal. I changed it for the long list of packages in the matching filter modal, but here is another example for the host collections page:

Before (notice there are two scrollbars and much of the modal is hidden):
here

After, modal is in full view and the only scroll is in the modal:
here

app/views/katello/api/v2/content_view_filter_rules/matching_content.json.rabl
johnpmitsch: An API call with `matching_content` to any rule that isn't a package filter rule will return an error saying this rule type isn't supported yet, I just added a way to catch this in the rabl in case anything gets by. The UI won't show any buttons for non-package rules so the user can't calll for matching content there.

The implementation has changed with the API change, but I kept the same error when it's not a package rule.

johnpmitsch:
[test katello]

app/controllers/katello/api/v2/content_view_filter_rules_controller.rb
```diff
@@ -67,8 +67,20 @@ def create
     api :GET, "/content_view_filters/:content_view_filter_id/rules/:id", N_("Show filter rule info")
     param :content_view_filter_id, :number, :desc => N_("filter identifier"), :required => true
     param :id, :number, :desc => N_("rule identifier"), :required => true
+    param :matching_content, :bool,
+          :desc => N_("Show content in Content View that matches the rule. Only package filters are " \
```
sean797: Does `s/package/RPM/` make sense here? since I don't think this works for deb

app/controllers/katello/api/v2/content_view_filter_rules_controller.rb
```diff
@@ -67,8 +67,20 @@ def create
...
+                      "currently supported. Defaults to false.")
     def show
-      respond :resource => @rule
+      locals = {}
+      if ::Foreman::Cast.to_bool(params[:matching_content] ||= false)
+        if @filter.type == Katello::ContentViewPackageFilter.to_s
+          locals[:matching_content] = @rule.matching_rpms
+        else
+          fail "Only Package Filters can return matching content at this time"
```
sean797: ditto

johnpmitsch:
> One question, would it make sense to filter the matching content alphabetically? It could be helpful to have when matching with wildcards.

That's a good idea. The only thing is when the alphabetic sort gets down to the versions they don't sort that well.  for example, foo-3.1.3 could show after foo-3.10.0. But I suppose it's not sorting that great as is, so i'll take a look :eyes:

ianballou:
I was mostly thinking about if RPMs with the same name could be grouped
together, maybe the version sorting could be less important.  I think the
code looks all good too by the way.

johnpmitsch:
@parthaa I updated to use `applicable_repos`, so now the affected repositories is taken into account.

@ianballou I used the RPM default sort in the query so now the rpms are sorted as we do elsewhere, which takes into account both name and version.

I also found https://projects.theforeman.org/issues/27885 but will address separately

app/controllers/katello/api/v2/content_view_filter_rules_controller.rb
johnpmitsch: Thanks, we should probably look at updating this whole file, but I did update the places I added as you suggested :+1:

johnpmitsch:
I did wind up fixing https://projects.theforeman.org/issues/27885 in this PR, but kept it as a separate commit (and will merge as a separate commit). Sorry for the late addition, but I was able to use some of the existing test fixtures I added in this PR and its much easier to test with this PR.

@parthaa and/or @ianballou can you be sure to test this? It should be a quick test, see the issue details and commit message for details

johnpmitsch:
[test katello]

ianballou:
Tested the new changes and it works as intended :+1:

johnpmitsch:
I just discussed this offline and am going to keep all the same functionality but move the logic in the API to the respository_content_controller.rb to keep it flexible for other content types in the future and to respond with the content in the format we expect

We also discussed addding a paginated table to the modal, but that may not happen in this PR. It will be more of a concern with other content types like errata, where more information needs to be displayed.

I'll comment when I've updated with these changes!

app/lib/katello/api/v2/rendering.rb
```diff
@@ -43,8 +43,7 @@ def respond_with_template(action, resource_name, options = {}, &_block)
 
           render :template => "katello/api/v2/#{resource_name}/#{action}",
                  :status => status,
-                 :locals => { :object_name => options[:object_name],
-                              :root_name => options[:root_name] },
+                 :locals => options.slice(:object_name, :root_name, :locals),
```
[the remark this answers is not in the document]
johnpmitsch: this isn't applicable to this PR, but I would like to keep this change in as it allows us to use a useful feature in Rabl

johnpmitsch:
@parthaa @jlsherrill @ianballou I still need to update some tests/linting/cleanup but I was able to use the repository content controller and use a paginated table. Here is what this looks like:

cv_check_matching_content_pagination

Some changes were needed to the nutupane factory to support multiple tables on one page, I added some details in code comments and will add more here

engines/bastion/app/assets/javascripts/bastion/components/nutupane.factory.js
```diff
@@ -64,7 +64,14 @@ angular.module('Bastion.components').factory('Nutupane',
...
             params.paged = true;
-            params.page = $location.search().page || 1;
+
+            locationPage = $location.search().page;
+            // the $location page can return NaN as a String, explicitly checking for that here
+            if (!locationPage || isNaN(locationPage)) {
+              params.page = 1;
+            } else {
+              params.page = locationPage;
+            }
```
johnpmitsch: This was necessary because sometimes the location is not paginated and will return "NaN" which somehow gets coerced into a string, so it becomes truthy and was not using the 1 as a default. This then throws off the Nutupane params and is also showing up in the URL params.

I added a more verbose check for a truthy value and `isNaN`, which checks for `"NaN"`

engines/bastion/app/assets/javascripts/bastion/components/nutupane.factory.js
```diff
@@ -118,8 +126,9 @@ angular.module('Bastion.components').factory('Nutupane',
 
                     table.rows = response.results;
 
-                    table.resource.page = parseInt(response.page, 10);
-                    params.page = table.params.page = parseInt(response.page, 10);
+                    responsePage = response.page || 1;
```
johnpmitsch: This will return `undefined` if the response does not have a page number, which it doesn't in a response for a single object. For example, for CV Filter Rules, we GET a single CV filter and then iterate through the `rules` attribute returned. There is no paging information in this response. This returns undefined and sets the page as undefined.

It has been working fine because the CV filter rules page is not paginated, but adding paginated pages to the same page exposed the issue.

app/controllers/katello/concerns/api/v2/repository_content_controller.rb
```diff
@@ -105,9 +114,12 @@ def sort_options
         end
       end
 
-      def filter_by_content_view_filter(filter, collection)
-        ids = filter.send("#{singular_resource_name}_rules").pluck(:uuid)
-        filter_by_ids(ids, collection)
```
johnpmitsch: This is only applicable to package groups (which uses uuid) and is failing for other content type queries. I moved this to raise `NotImplemented` errors if the child class doesn't implement the method and moved this implementation to package groups.

With this PR, this is overridden in the following child classes:
```
app/controllers/katello/api/v2/docker_manifests_controller.rb:    def filter_by_content_view_filter(filter)
app/controllers/katello/api/v2/errata_controller.rb:    def filter_by_content_view_filter(filter, collection)
app/controllers/katello/api/v2/module_streams_controller.rb:    def filter_by_content_view_filter(filter, collection)
app/controllers/katello/api/v2/package_groups_controller.rb:    def filter_by_content_view_filter(filter, collection)
app/controllers/katello/api/v2/packages_controller.rb:    def filter_by_content_view_filter(filter, collection)
```

johnpmitsch:
[test katello]

johnpmitsch:
@ianballou @jlsherrill @parthaa this is ready for re-review. 

I was testing out the various filters, the errata filter doesn't seem to be showing any results to select from and I'm seeing the same in master. I'll look more tomorrow, I could be missing something

johnpmitsch:
[test katello]

ianballou:
[test katello]

parthaa:
just the single comment. lgtm otherwise

app/models/katello/content_view_package_filter.rb
```diff
@@ -31,21 +31,32 @@ def self.generate_rpm_clauses(package_filenames = [])
...
+    end
+
+    def query_rpms_from_repo(repo, rule)
+      rpms = Rpm.in_repositories(repo).non_modular
+      query_rpms_from_collection(rpms, rule)
+    end
+
+    def query_rpms_from_collection(collection, rule)
       query_name = rule.name.tr("*", "%")
-      query = Rpm.in_repositories(repo).non_modular.where("#{Rpm.table_name}.name ilike ?", query_name)
```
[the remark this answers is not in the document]
johnpmitsch: are you thinking it should be `collection.non_modular` below?

johnpmitsch:
@parthaa Good call on the non_modular part, they were actually showing up in the filters but would not be actually filtered in the content_view. I moved that to the filter_by_collection method so it will always use non_modular rpms, both in the index call and the actual filter. It's in a separate commit if you would like to see the change.

johnpmitsch:
[test katello]

johnpmitsch:
[test katello]

johnpmitsch:
[test katello]

PR #8327, opened 2019-09-03
Fixes #27768 - Fix package action on host collection

xprazak2:
@jturel, how does this look? Should not revert [#27258](https://github.com/Katello/katello/commit/0ec25db78db6a9b9b9024219d4da26d2a21f9938#diff-7330cd901481c6b6825a0b2bdaab9738)...

### DIFF
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/host-collections/details/host-collection-details.controller.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/host-collections/details/host-collection-details.controller.js
@@ -38,7 +38,8 @@ angular.module('Bastion.host-collections').controller('HostCollectionDetailsCont
         $scope.getHostIds = function() {
             return {
                 included: {
-                    search: 'host_collection_id = %s'.replace('%s', $scope.$stateParams.hostCollectionId)
+                    search: 'host_collection_id = %s'.replace('%s', $scope.$stateParams.hostCollectionId),
+                    ids: $scope.hostCollection ? $scope.hostCollection.host_ids : []
                 }
             };
         };


jturel:
Thanks for the fix @xprazak2 . It looks good but I'm on PTO this week. @sseelam2 can you take a look? You tested my related PR. Looks like I missed a similar but different scenario with Host Collections.

sseelam2:
@jturel Sure. I'll take a look at it.

sseelam2:
@xprazak2  I tried installing a package to my host in the host collection through remote execution and it's failing. I used this upstream URL https://inecas.fedorapeople.org/fakerepos/severity_zoo/.Maybe I'm not using the right test data. Could you please provide me with something that I can use to test this PR.

sseelam2:
@jturel  @xprazak2  I'm getting the below error when I install a package with REX. Please let me know if I am missing something.
> [Foreman::Exception]: The smart proxy task 0e28a756-bf50-42ed-b878-1ef60f21b17a failed. (Foreman::Exception)

I created a product with this upstream URL  (https://repos.fedorapeople.org/pulp/pulp/fixtures/rpm/).
I also created a Host and subscribed the above product to it and added this host to a host collection. I tried installing the bear-4.1-1.noarch  and camel-0.1-1.noarch package using REX and got the above error.

xprazak2:
@sseelam2, thank you for testing! It is difficult to say why the task failed from that error, it could be caused by a number of reasons such as networking or authentication problems. There could be something in the task details/output or logs on proxy that could point you in the right direction. 

But the fact that the task even starts is a good sign - it shows that target host ids are submitted in the form. Without this fix, the whole page just crashes on form submission:

no-hosts

sseelam2:
@xprazak2 Thanks for clearing that up. After applying your PR, when I try to install a package through REX, a job starts running which wasn't the case without your change. This PR looks good to me as well. 
ACK :+1:

PR #8328, opened 2019-09-03
Fixes #27771 - updates pulp3 gems

jjeffers:
This PR updates the pulp gems and VCR tapes for related pulp tests.

### DIFF
--- a/test/actions/pulp3/orchestration/file_sync_test.rb
+++ b/test/actions/pulp3/orchestration/file_sync_test.rb
@@ -41,6 +41,30 @@ module ::Actions::Pulp3
       assert_equal repository_reference.repository_href + "versions/2/", @repo.version_href
     end
 
+    def test_sync_with_pagination
+      @repo.root.update_attributes(:url => "https://repos.fedorapeople.org/repos/pulp/pulp/fixtures/file-many/", :mirror_on_sync => false)
+      ForemanTasks.sync_task(
+          ::Actions::Pulp3::Orchestration::Repository::Update,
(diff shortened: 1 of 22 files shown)


jlsherrill:
@sjha4  or @ianballou would one of you mind looking over this again? i've changed it quite a bit

jjeffers:
[test katello]

test/actions/pulp3/orchestration/file_sync_test.rb
```diff
@@ -41,6 +41,24 @@ def test_sync
...
     end
 
+    def test_sync_with_pagination
+      @repo.root.update_attributes(:url => "https://repos.fedorapeople.org/repos/pulp/pulp/fixtures/file-many/", :mirror_on_sync => false)
+      ForemanTasks.sync_task(
+          ::Actions::Pulp3::Orchestration::Repository::Update,
+          @repo,
+          @master)
+
+      SETTINGS[:katello][:pulp][:bulk_load_size] = 10
```
jlsherrill: this persists across tests, so i think any test that runs after this one would fail.  This would be random for each test run, so this explains all the failures you're seeing :)

I think setting it back in an ensure would probably fix it!

jlsherrill: i'd delete all the pulp2 vcr rerecords, as those likely will fail after this change (and maybe some of the others too)

jlsherrill:
Can you remove the pulp2 service class cassette changes to make sure that they run without any changes?

PR #8329, opened 2019-09-03
Fixes #27776 - Run metadata generation on right capsule

sjha4:


sjha4:
[test katello]

jlsherrill:
Why is this being opened against 3.12 exactly?

sjha4:
> Why is this being opened against 3.12 exactly?

Must have needed coffee..From what I checked earlier seemed to me that master is fixed and we only needed this downstream..Checking this again, it is needed on master so I'll reopen this against master..

sjha4:
Closing this..

PR #8331, opened 2019-09-04
Fixes #27783 - CVV status error in default CV

ianballou:
The user should not be able to browse to the Default Content View since it isn't meant to be interacted with.  It should only be manually accessed via URL when needed for debugging.

### DIFF
--- a/app/views/foreman/smart_proxies/_content_tab.html.erb
+++ b/app/views/foreman/smart_proxies/_content_tab.html.erb
@@ -28,7 +28,7 @@
     </tr>
     <tr ng-repeat-end ng-repeat="cv in env.content_views" ng-show="isEnvronmentExpanded(env)">
       <td>
-        <a href="/content_views/{{cv.id}}/versions" target="_self">
+        <a href="{{ fetchUrl(cv.default, cv.id) }}" target="_self">
           {{ cv.name }}
         </a>
       </td>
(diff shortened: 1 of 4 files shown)


engines/bastion_katello/app/assets/javascripts/bastion_katello/capsule-content/capsule-content.controller.js
```diff
@@ -63,6 +63,10 @@ angular.module('Bastion.capsule-content').controller('CapsuleContentController',
             }
         }
 
+        $scope.fetchUrl = function (cvId) {
+            return cvId === 1 ? '/products' : '/content_views/' + cvId + '/versions';
```
jlsherrill: I think i mentioned this on irc, but putting it here:  Checking for id==1 will only work for the first org :)  there should be a 'default' attribute you can use

ianballou: oh gotcha, didn't know it would only work for the first organization.  I'll update.

ianballou:
@jlsherrill should be all set now.

app/views/foreman/smart_proxies/_content_tab.html.erb
```diff
@@ -28,7 +28,7 @@
     </tr>
     <tr ng-repeat-end ng-repeat="cv in env.content_views" ng-show="isEnvronmentExpanded(env)">
       <td>
-        <a href="/content_views/{{cv.id}}/versions" target="_self">
+        <a href="{{ fetchUrl(cv.default, cv.id) }}" target="_self">
```
jturel: I think the name fetchUrl is a bit misleading since it doesn't fetch anything. Something like contentViewVersionUrl or productsOrVersionUrl are a bit clearer, or something else :)

jturel:
Closing in favor of https://github.com/Katello/katello/pull/8337 being merged. thanks @ianballou !

PR #8333, opened 2019-09-05
Fixes #27795 - properly remove uneeded repos from FPC

jlsherrill:


app/services/katello/pulp/smart_proxy_repository.rb
```diff
@@ -58,7 +58,7 @@ def current_repositories_data(environment = nil, content_view = nil)
       end
 
       def orphaned_repos
-        @smart_proxy.pulp_repositories.map { |x| x["id"] } - current_repositories.map { |x| x.pulp_id }
+        @smart_proxy.pulp_repositories.map { |x| x["id"] } - repos_available_to_capsule.map { |x| x.pulp_id }
```
jlsherrill: current_repositories returns 'repositories that are on the capsule that should be on the capsule'.   While 'repos_available_to_capsule' are repos that should be on the capsule regardless of whether they are at the moment or not

jlsherrill:
[test katello]

PR #8334, opened 2019-09-06
Clean up Pulp3 code

ekohl:
This applies some best practices of using constants where possible.

It also uses the scheme from the pulp_url setting rather than hardcoding to https making it easier to deploy in development. It still maintains the hardcoded username/password because I'm unsure what the reason behind this is.

### DIFF
--- a/app/models/katello/concerns/smart_proxy_extensions.rb
+++ b/app/models/katello/concerns/smart_proxy_extensions.rb
@@ -147,11 +147,11 @@ module Katello
       def content_pulp3_support?(content_type)
         content_type = content_type.is_a?(String) ? content_type : content_type.model_class::CONTENT_TYPE
         type = Katello::RepositoryTypeManager.find_repository_type content_type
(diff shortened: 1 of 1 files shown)


jlsherrill:
Thanks @ekohl the user/pass are temporary and we're going to move to ssl-based auth soon

app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -146,17 +147,21 @@ def pulp3_support?(repository)
...
-      def pulp3_host!
-        url = self.setting('Pulp3', 'pulp_url')
+      def pulp3_uri!
+        url = self.setting(PULP3_FEATURE, 'pulp_url')
         fail "Cannot determine pulp3 url, check smart proxy configuration" unless url
-        URI.parse(url).host
+        URI.parse(url)
+      end
+
+      def pulp3_host!
```
ekohl: I kept this in for compatibility but I don't think this is used anywhere now btw.

PR #8336, opened 2019-09-06
Fixes #27605 - Adding generic first row for grouped subs

sseelam2:
This PR is still a work in progess. I would appreciate suggestions. 
This PR is changing the view of the subscriptions with multiple entitlements.

**To test this PR -**
- Create a manifest with a subscription having multiple entitlements in the Red Hat customer portal. 
- Click on manage manifest and import this manifest under the Subscriptions tab.
- When this is viewed in the UI (Content -> Subscriptions), the first row for this subscription must be collapsible having the subscription name and product ID and everything else set to NA. When '>' is clicked it should show all the subscriptions under the generic row  with the same product ID.

### DIFF
--- a/webpack/move_to_foreman/components/common/table/components/TableSelectionCell.js
+++ b/webpack/move_to_foreman/components/common/table/components/TableSelectionCell.js
@@ -5,17 +5,18 @@ import { translate as __ } from 'foremanReact/common/I18n';
 import { noop } from 'foremanReact/common/helpers';
 
 const TableSelectionCell = ({
-  id, before, after, label, checked, onChange, ...props
+  id, before, after, label, checked, hide, onChange, ...props
 }) => (
   <Table.SelectionCell>
     {before}
+    {!hide &&
     <Table.Checkbox
       id={id}
       label={label}
       checked={checked}
       onChange={onChange}
       {...props}
-    />
+    />}
     {after}
   </Table.SelectionCell>
 );
@@ -26,6 +27,7 @@ TableSelectionCell.propTypes = {
   after: PropTypes.node,
   label: PropTypes.string,
   checked: PropTypes.bool,
+  hide: PropTypes.bool,
   onChange: PropTypes.func,
 };
 
@@ -34,6 +36,7 @@ TableSelectionCell.defaultProps = {
   after: null,
   label: __('Select row'),
   checked: false,
+  hide: false,
   onChange: noop,
 };
 
--- a/webpack/move_to_foreman/components/common/table/formatters/EntitlementsInlineEditFormatter.js
+++ b/webpack/move_to_foreman/components/common/table/formatters/EntitlementsInlineEditFormatter.js
@@ -5,12 +5,18 @@ import { Table, FormControl, FormGroup, HelpBlock, Spinner } from 'patternfly-re
 import { validateQuantity } from '../../../../../scenes/Subscriptions/SubscriptionValidations';
 
 const renderValue = (value, additionalData, onActivate) => {
-  const { available, upstream_pool_id: upstreamPoolId } = additionalData.rowData;
-  if (available < 0 || !upstreamPoolId) {
+  const { available, upstream_pool_id: upstreamPoolId, collapsible } = additionalData.rowData;
+
+  if (collapsible) {
+    return (
+      <td>{__('NA')}</td>
+    );
+  } else if (available < 0 || !upstreamPoolId) {
     return (
       <td>{available < 0 ? __('Unlimited') : available}</td>
     );
   }
+
   return (
     <td className="editable">
       <div
@@ -34,6 +40,7 @@ const renderEdit = (hasChanged, onChange, value, additionalData) => {
   const {
     upstreamAvailable, upstreamAvailableLoaded, maxQuantity,
   } = additionalData.rowData;
+
   const className = hasChanged(additionalData)
     ? 'editable editing changed'
     : 'editable editing';
--- a/webpack/move_to_foreman/components/common/table/formatters/collapseableAndSelectionCellFormatter.js
+++ b/webpack/move_to_foreman/components/common/table/formatters/collapseableAndSelectionCellFormatter.js
@@ -4,8 +4,10 @@ import CollapseSubscriptionGroupButton from '../components/CollapseSubscriptionG
 
 export default (collapseableController, selectionController, additionalData) => {
   const shouldShowCollapseButton = collapseableController.isCollapseable(additionalData);
+  const isGenericRow = additionalData.rowData.collapsible;
 
   return selectionCellFormatter(
+    isGenericRow,
     selectionController,
     additionalData,
     shouldShowCollapseButton && (
--- a/webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTableHelpers.js
+++ b/webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTableHelpers.js
@@ -30,18 +30,39 @@ const buildTableRow = (subscription, availableQuantities, updatedQuantity) => {
       quantity: updatedQuantity[subscription.id],
     };
   }
+
   return baseSubscription;
+};
+
+const buildTableCollapseRow = (subscriptionGroup) => {
+  const first = subscriptionGroup.subscriptions[0];
+  const heading = {
+    id: 0,
+    collapsible: true,
+    contract_number: 'NA',
+    start_date: 'NA',
+    end_date: 'NA',
+    consumed: 'NA',
+    product_id: first.product_id,
+    name: first.name,
+    virt_only: first.virt_only,
+    hypervisor: first.hypervisor,
+  };
+  return heading;
 };
 
 const buildTableRowsFromGroup = (subscriptionGroup, availableQuantities, updatedQuantity) => {
   const { open, subscriptions } = subscriptionGroup;
 
-  // build row for each subscription
-  if (open) {
-    return subscriptions.map(subscription =>
-      buildTableRow(subscription, availableQuantities, updatedQuantity));
+  if (subscriptions.length > 1) {
+    const rows = [];
+    rows.push(buildTableCollapseRow(subscriptionGroup));
+    if (open) {
+      subscriptions.forEach(sub =>
+        rows.push(buildTableRow(sub, availableQuantities, updatedQuantity)));
+    }
+    return rows;
   }
-
   // build row only for the first subscription in the group
   const [firstSubscription] = subscriptions;
   return [buildTableRow(firstSubscription, availableQuantities, updatedQuantity)];
--- a/webpack/scenes/Subscriptions/components/SubscriptionsTable/components/Table.js
+++ b/webpack/scenes/Subscriptions/components/SubscriptionsTable/components/Table.js
@@ -28,10 +28,8 @@ const Table = ({
 
   const groupingController = {
     isCollapseable: ({ rowData }) =>
-      // it is the first subscription in the group
-      rowData.id === groupedSubscriptions[rowData.product_id].subscriptions[0].id &&
       // the group contains more then one subscription
-      groupedSubscriptions[rowData.product_id].subscriptions.length > 1,
+      rowData.collapsible,
     isCollapsed: ({ rowData }) => !groupedSubscriptions[rowData.product_id].open,
     toggle: ({ rowData }) => toggleSubscriptionGroup(rowData.product_id),
   };
--- a/webpack/move_to_foreman/components/common/table/formatters/selectionCellFormatter.js
+++ b/webpack/move_to_foreman/components/common/table/formatters/selectionCellFormatter.js
@@ -2,14 +2,17 @@ import React from 'react';
 import TableSelectionCell from '../components/TableSelectionCell';
 
 export default (
+  isGenericRow,
   selectionController,
   additionalData,
   before,
   after,
 ) => (
+
   <TableSelectionCell
     id={`select${additionalData.rowIndex}`}
     checked={selectionController.isSelected(additionalData)}
+    hide={isGenericRow}
     onChange={() => selectionController.selectRow(additionalData)}
     before={before}
     after={after}
--- a/webpack/scenes/Subscriptions/components/SubscriptionsTable/__tests__/SubscriptionsTable.fixtures.js
+++ b/webpack/scenes/Subscriptions/components/SubscriptionsTable/__tests__/SubscriptionsTable.fixtures.js
@@ -2,7 +2,7 @@ import Immutable from 'seamless-immutable';
 
 const groupedSubscriptions = Immutable({
   RH00001: {
-    open: false,
+    open: true,
     subscriptions: [
       {
         id: 1,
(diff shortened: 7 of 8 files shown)


jturel:
@sseelam2 I saw this was in ready for review but it appears to be a work in progress. Was there something specific you're looking for feedback on?

sseelam2:
@jturel  Sorry I forgot to mention that. I still have not figured out how to make the entitlements column NA for the generic row. And also I'm not sure why '>' this symbol comes up in the first two rows when it should be present only on the first row.

jturel:
@sseelam2 that's OK. I am also curious if you talked with anyone from the UX side on this. Maybe it's better to *not* collapse/group the subscriptions, but instead list them individually. That makes it easier from our side but I think I can also see why we have the grouping.

sseelam2:
@jturel I discussed with Roxanne about this. She went through the description of the bug and mentioned that she had seen something just like this earlier and asked me to go ahead with this solution.

webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTableHelpers.js
```diff
@@ -69,6 +70,17 @@ export const groupSubscriptionsByProductId = ({ results: subscriptions }) => {
     grouped[subscription.product_id].subscriptions.unshift(subscription);
   });
 
+  for (product_id in grouped) {
```
jturel: I see this code is changing the `groupSubscriptionsByProductId` function. I see what you're trying to do but I don't think this is the right place to do it. At this level of the code we don't want to affect the grouping since it's already correct.

Notice that this function is used here: https://github.com/Katello/katello/blob/3a485398c78a42ad02fb19989a5d09f919f147ea/webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTable.js#L33-L38

I see there is a function called 'buildTableRows' and I think that might be the better place to do this but I haven't looked at the implemtnation of that function.

jturel:
Kicked off a dolly build

webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTableHelpers.js
sseelam2: @jturel Since buildTableRows gets called every time we collapse, If I write the same logic inside buildTableRows it results in multiple generic rows. I'm trying to find a way around it if possible.

sseelam2: @jturel I updated the code. Let me know if this looks good. I still need to figure out how to make entitlements "NA".

webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTableHelpers.js
```diff
@@ -36,6 +36,19 @@ const buildTableRow = (subscription, availableQuantities, updatedQuantity) => {
 const buildTableRowsFromGroup = (subscriptionGroup, availableQuantities, updatedQuantity) => {
   const { open, subscriptions } = subscriptionGroup;
 
+  if (subscriptionGroup.subscriptions.length > 1) {
+    const temp = JSON.parse(JSON.stringify(subscriptionGroup.subscriptions[0]));
```
jturel: is the JSON parsing here necessary? I feel like the data should already be an object we can work with

sseelam2: @jturel It was a read only object. Couldn't modify it. So I created a copy.

webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTableHelpers.js
```diff
@@ -36,6 +36,19 @@ const buildTableRow = (subscription, availableQuantities, updatedQuantity) => {
...
+    temp.id=0
+    temp.contract_number = 'NA';
+    temp.start_date = 'NA';
+    temp.end_date = 'NA';
+    temp.consumed = 'NA';
+
+    if (subscriptionGroup.subscriptions[0].id != 0) {
+      subscriptionGroup.subscriptions.unshift(temp);
```
jturel: I don't really like how we're creating a 'fake' subscription by adding it to the subscriptions list to create a row - but it does work!

Here's another approach to get around modifying the subscriptions array which I think is better practice to keep our data and view separate:

```
diff --git a/webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTableHelpers.js b/webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTableHelpers.js
index f4f1a3b..f580299 100644
--- a/webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTableHelpers.js
+++ b/webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTableHelpers.js
@@ -33,13 +33,33 @@ const buildTableRow = (subscription, availableQuantities, updatedQuantity) => {
   return baseSubscription;
 };
 
+const buildTableCollapseRow = (subscriptionGroup) => {
+  const { subscriptions } = subscriptionGroup;
+  const first = subscriptions[0];
+
+  let heading = {};
+  heading.contract_number = 'NA';
+  heading.start_date = 'NA';
+  heading.end_date = 'NA';
+  heading.consumed = 'NA';
+  heading.product_id = first.product_id;
+  heading.name = first.name;
+
+  return heading;
+}
+
 const buildTableRowsFromGroup = (subscriptionGroup, availableQuantities, updatedQuantity) => {
   const { open, subscriptions } = subscriptionGroup;
 
-  // build row for each subscription
-  if (open) {
-    return subscriptions.map(subscription =>
-      buildTableRow(subscription, availableQuantities, updatedQuantity));
+  let rows = [];
+
+  if (subscriptions.length > 1) {
+    rows.push(buildTableCollapseRow(subscriptionGroup));
+
+    // build row for each subscription
+    if (open) {
+      return rows.concat(subscriptions.map(sub => buildTableRow(sub, availableQuantities, updatedQuantity)));
+    }
   }
 
   // build row only for the first subscription in the group
diff --git a/webpack/scenes/Subscriptions/components/SubscriptionsTable/components/Table.js b/webpack/scenes/Subscriptions/components/SubscriptionsTable/components/Table.js
index 33d6465..81e88a5 100644
--- a/webpack/scenes/Subscriptions/components/SubscriptionsTable/components/Table.js
+++ b/webpack/scenes/Subscriptions/components/SubscriptionsTable/components/Table.js
@@ -29,7 +29,7 @@ const Table = ({
   const groupingController = {
     isCollapseable: ({ rowData }) =>
       // it is the first subscription in the group
-      rowData.id === groupedSubscriptions[rowData.product_id].subscriptions[0].id &&
+      //rowData.id === groupedSubscriptions[rowData.product_id].subscriptions[0].id &&
       // the group contains more then one subscription
       groupedSubscriptions[rowData.product_id].subscriptions.length > 1,
     isCollapsed: ({ rowData }) => !groupedSubscriptions[rowData.product_id].open,
```
I don't think what I pasted is fully working. We'll also want to remove the checkbox from the collapsible row since it doesn't really have a purpose (it's not an actual subscription).

Let me know if I didn't explain this well.

sseelam2: @jturel I completely agree with you. I tried doing something along the same lines and it was not working properly. So I switched to this solution. But I do agree that creating a fake subscription was not right.

jlsherrill:
Can we make the commit message a little more descriptive?

sseelam2:
@jlsherrill We are not sure how we should be displaying the subscriptions yet. @jturel and I have a meeting with Roxanne this friday to figure out a better way of displaying the subscriptions.

webpack/scenes/Subscriptions/components/SubscriptionsTable/components/Table.js
```diff
@@ -27,7 +27,7 @@ const Table = ({
   const groupingController = {
     isCollapseable: ({ rowData }) =>
       // it is the first subscription in the group
-      rowData.id === groupedSubscriptions[rowData.product_id].subscriptions[0].id &&
+      rowData.id === 0 &&
```
jturel: How is checking the id is 0 significant?

jturel: Oh, I think it's so we don't get the '>' on the expanded rows also?

sseelam2: @jturel yes that's exactly why.

jturel: I wonder if there's some other distinguishing characteristic about the row (maybe something else we can add) to identify when we should collapse rather than id === 0. What if we had an Id later? That would break the functionality because it's not obvious that ID is significant in that way.

jturel: What about putting a 'collapsible' field on the custom row we are building?

sseelam2: @jturel Subscriptions already have an id field which starts from 1. So I just used id=0 to represent a generic row. But using a collapsible field sounds good to me.

webpack/scenes/Subscriptions/components/SubscriptionsTable/components/Table.js
```diff
@@ -27,7 +27,7 @@ const Table = ({
   const groupingController = {
     isCollapseable: ({ rowData }) =>
       // it is the first subscription in the group
-      rowData.id === groupedSubscriptions[rowData.product_id].subscriptions[0].id &&
+      rowData.collapsible &&
       // the group contains more then one subscription
       groupedSubscriptions[rowData.product_id].subscriptions.length > 1,
```
jturel: the groupedSubscriptions check might be redundant now that we have the collapsible field on the row itself

jturel:
@sseelam2 looks like the build is green now. is this ready for re-review? just checking in case you were planning any more changes

sseelam2:
> @sseelam2 looks like the build is green now. is this ready for re-review? just checking in case you were planning any more changes

Yes, It is ready for re-review. 👍

jturel:
@jeremylenz hey! I'll take care of the end-to-end testing of the PR. Would you mind looking at it from the JS perspective and make sure that what we came up with isn't breaking any rules?

webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTableHelpers.js
```diff
@@ -30,18 +30,40 @@ const buildTableRow = (subscription, availableQuantities, updatedQuantity) => {
...
+  const rows = [];
 
-  // build row for each subscription
-  if (open) {
-    return subscriptions.map(subscription =>
-      buildTableRow(subscription, availableQuantities, updatedQuantity));
+  if (subscriptions.length > 1) {
+    rows.push(buildTableCollapseRow(subscriptionGroup));
+    if (open) {
+      return rows.concat(subscriptions.map(sub =>
```
jeremylenz: What about, instead of `return rows.concat` if we were to use `subscriptions.forEach(sub => rows.push(...` ? 

That way, on line 65 you're just returning `rows` no matter what, and it's a little easier to see what's happening.

webpack/move_to_foreman/components/common/table/components/TableSelectionCell.js
```diff
@@ -5,17 +5,17 @@ import { translate as __ } from 'foremanReact/common/I18n';
 import { noop } from 'foremanReact/common/helpers';
 
 const TableSelectionCell = ({
-  id, before, after, label, checked, onChange, ...props
+  id, before, after, label, checked, hide, onChange, ...props
 }) => (
   <Table.SelectionCell>
     {before}
-    <Table.Checkbox
+    {!hide && <Table.Checkbox
```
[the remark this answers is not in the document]
sseelam2: alright. thanks for the feedback. will move it to the next line. :+1:

webpack/move_to_foreman/components/common/table/formatters/collapseableAndSelectionCellFormatter.js
```diff
@@ -4,8 +4,10 @@ import CollapseSubscriptionGroupButton from '../components/CollapseSubscriptionG
 
 export default (collapseableController, selectionController, additionalData) => {
   const shouldShowCollapseButton = collapseableController.isCollapseable(additionalData);
+  const isGenericRow = additionalData.rowData.collapsible;
```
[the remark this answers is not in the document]
sseelam2: The table rows are built only if there are any subscriptions and these rows are passed in as additional data to collapseableAndSelectionCellFormatter.js . So I don't think addtionalData can be empty. Correct me if I'm wrong @jturel.

webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTableHelpers.js
```diff
@@ -30,18 +30,40 @@ const buildTableRow = (subscription, availableQuantities, updatedQuantity) => {
...
   }
+
   return baseSubscription;
 };
 
+const buildTableCollapseRow = (subscriptionGroup) => {
+  const first = subscriptionGroup.subscriptions[0];
+  const heading = {};
+
+  heading.id = 0;
```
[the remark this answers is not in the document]
sseelam2: This makes sense. Thanks :+1:

webpack/move_to_foreman/components/common/table/formatters/collapseableAndSelectionCellFormatter.js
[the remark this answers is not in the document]
jeremylenz: @sseelam2 `rowData` is the one at issue.  If `additionalData` is defined but `additionalData.rowData` isn't, it will error.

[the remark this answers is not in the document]
sseelam2: additonalData doesn't get created if there is no rowData. additonalData contains additional info about the row. So I don't think we'll end up in that situation. But I would like @jturel to confirm. Just to be extra sure about this.

webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTableHelpers.js
sseelam2: Nice catch.  Thanks for your feedback @jeremylenz .

sseelam2:
@jturel PR is ready for testing :)

webpack/move_to_foreman/components/common/table/formatters/collapseableAndSelectionCellFormatter.js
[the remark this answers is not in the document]
jturel: That sounds correct :)

webpack/move_to_foreman/components/common/table/formatters/EntitlementsInlineEditFormatter.js
```diff
@@ -6,11 +6,17 @@ import { validateQuantity } from '../../../../../scenes/Subscriptions/Subscripti
 
 const renderValue = (value, additionalData, onActivate) => {
   const { available, upstream_pool_id: upstreamPoolId } = additionalData.rowData;
-  if (available < 0 || !upstreamPoolId) {
+
+  if (additionalData.rowData.collapsible) {
```
jturel: Can we pull out `collapsible` via destructuring on line 8? What's here works fine, but that would be more consistent with the surrounding code

webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTableHelpers.js
```diff
@@ -30,18 +30,39 @@ const buildTableRow = (subscription, availableQuantities, updatedQuantity) => {
...
+    name: first.name,
+    virt_only: first.virt_only,
+    hypervisor: first.hypervisor,
+  };
+  return heading;
+};
+
 const buildTableRowsFromGroup = (subscriptionGroup, availableQuantities, updatedQuantity) => {
   const { open, subscriptions } = subscriptionGroup;
+  const rows = [];
```
jturel: I think rows can be initialized inside the below if statement, so that we don't initialize it unnecessarily

sseelam2:
Thanks @jturel and @jeremylenz for the feedback! :smile:

PR #8341, opened 2019-09-10
Fixes #27825 - Set autopublish for non-lib puppet

parthaa:
When creating puppet repos on a capsule

Previously:
puppet_install_dist would have auto_publish to true even for library
repos.

This commit:
puppet_install_dist would have auto_publish to true only for
non-library repos.

### DIFF
--- a/app/services/katello/pulp/repository/puppet.rb
+++ b/app/services/katello/pulp/repository/puppet.rb
@@ -18,7 +18,7 @@ module Katello
           importer_class.new(config.merge(mirror_importer_connection_options))
         end
 
-        def auto_publish?
+        def puppet_install_auto_publish?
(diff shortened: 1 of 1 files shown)


app/services/katello/pulp/repository/puppet.rb
```diff
@@ -18,8 +18,15 @@ def generate_mirror_importer
           importer_class.new(config.merge(mirror_importer_connection_options))
         end
 
+        def auto_publish?
```
jlsherrill: can you rename this to be more specific?  puppet_install_auto_publish?  or something like that

parthaa:
[test katello]

jlsherrill:
[test katello]

PR #8344, opened 2019-09-11
Fixes #27840 - Setting to toggle host profile stealing

jturel:
To test:

- register a client to your dev box with this PR checked out
- `subscription-manager clean` on the client and change its DMI UUID:

```
echo '{"dmi.system.uuid": "this-is-made-up1"}' > /etc/rhsm/facts/uuid.facts
```
- re-register the client. it should succeed (crux of this change) as long as the client's DMI UUID is unique across hosts.

To test the reverse: toggle the new Host Profile Assume setting to false. Repeating the test (with a new UUID) will result in a registration error

### DIFF
--- a/app/models/setting/content.rb
+++ b/app/models/setting/content.rb
@@ -137,6 +137,9 @@ class Setting::Content < Setting
                  N_("If hosts fail to register because of duplicate DMI UUIDs " \
                     "add their comma-separated values here. Subsequent registrations will generate a unique DMI UUID for the affected hosts."),
                  [], N_('Host Duplicate DMI UUIDs')),
+        self.set('host_profile_assume', N_("Allow new Host registrations to assume registered profiles with matching hostname " \
+                    "as long as the registering DMI UUID is not used by another host."),
+                 true, N_('Host Profile Assume')),
         self.set('host_tasks_workers_pool_size', N_("Amount of workers in the pool to handle the execution of host-related tasks. When set to 0, the default queue will be used instead. Restart of the dynflowd/foreman-tasks service is required."),
                  5, N_('Host Tasks Workers Pool Size'))
       ].each { |s| self.create! s.update(:category => "Setting::Content") }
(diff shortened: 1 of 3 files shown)


app/models/setting/content.rb
```diff
@@ -137,6 +137,9 @@ def self.load_defaults
                  N_("If hosts fail to register because of duplicate DMI UUIDs " \
                     "add their comma-separated values here. Subsequent registrations will generate a unique DMI UUID for the affected hosts."),
                  [], N_('Host Duplicate DMI UUIDs')),
+        self.set('host_profile_steal', N_("Allow new Host registrations to overtake registered profiles with matching hostname " \
```
mccun934: how about "host_profile_assume"  and: "Allow new Host registrations to assume registered profiles with matching hostname as long as the registering DMI UUID is unique."

jturel: updated!

app/models/setting/content.rb
```diff
@@ -137,6 +137,9 @@ def self.load_defaults
                  N_("If hosts fail to register because of duplicate DMI UUIDs " \
                     "add their comma-separated values here. Subsequent registrations will generate a unique DMI UUID for the affected hosts."),
                  [], N_('Host Duplicate DMI UUIDs')),
+        self.set('host_profile_assume', N_("Allow new Host registrations to assume registered profiles with matching hostname " \
+                    "as long as the registering DMI UUID is unique."),
+                 true, N_('Host Profile Assume')),
```
jlsherrill: Trying to understand this working.   "as long as registering DMI uuid is unique".  Do you mean 'unique other than the host its taking' ?  I assume it already would take the host's profile if its uuid matched that, right?

jturel: Correct! It's kind of a strange behavior to articulate. Let me know if you have a suggestion

jlsherrill: "as long as the registering DMI UUID is not used by another host" ?

jturel: I'm good with that, updating

jturel:
@jlsherrill  updated!

PR #8348, opened 2019-09-16
Fixes #27863 - settings initialization improvement

ezr-ondrej:
As theforeman/foreman#6902 got merge, we need to refactor to avoid double initialization.

~~Needs foreman `>= 1.24`, but I see no foreman version dependency in `engine.rb` file.~~
Edit: found it.


ezr-ondrej:
@jlsherrill rebased :)

jlsherrill:
thanks so much @ezr-ondrej !!

PR #8349, opened 2019-09-17
Fixes #27864 - publish cv in any location

jlsherrill:


app/lib/actions/katello/repository/metadata_generate.rb
```diff
@@ -10,15 +10,15 @@ def plan(repository, options = {})
...
           else
             matching_content = options.fetch(:matching_content, false)
           end
 
           plan_pulp_action([Pulp::Repository::DistributorPublish, Pulp3::Orchestration::Repository::GenerateMetadata],
-                        repository, SmartProxy.find(smart_proxy_id),
+                        repository, smart_proxy,
                         :force => force,
                         :source_repository => source_repository,
                         :matching_content => matching_content,
```
sjha4: @jlsherrill : I remember changing this earlier. Is this is a refactor or was this causing the issue?

jlsherrill: SmartProxy.find() respects location and permission settings, so we can't ever use it as is in these types of actions (but instead need to use .unscoped).  In this case the smart proxy id was being passed in, and had already been looked up, so i figured we could just use it as is

jlsherrill:
@chris1984 oh sorry, by 'any location' i meant any location, even ones that aren't the 'default location'.  I did NOT mean the actual setting "Any Location".  Terrible wording is terrible ;)

chris1984:
ACK updated testing shows still working:

Screenshot of proxy not in new location:

https://nimb.ws/82faMZ

Screenshot of publish working:

https://nimb.ws/zdFlNK

jlsherrill:
Thanks @chris1984 and @sjha4 !!!

PR #8350, opened 2019-09-17
Fixes #27870 - Remove Candlepin Events API

sseelam2:
Removing events tab from content hosts and returning null for the events api.


app/controllers/katello/api/v2/host_subscriptions_controller.rb
```diff
@@ -55,8 +55,7 @@ def auto_attach
     api :GET, "/hosts/:host_id/subscriptions/events", N_("List subscription events for the host")
     param :host_id, Integer, :desc => N_("Id of the host"), :required => true
     def events
-      collection = full_result_response(@host.subscription_facet.candlepin_consumer.events.map { |e| OpenStruct.new(e) })
-      respond_for_index :collection => collection
+      respond_for_index :collection => {}
```
jturel: I think this should technically be an Array rather than Hash

app/services/katello/candlepin/consumer.rb
```diff
@@ -13,7 +13,6 @@ class Consumer
       CP_TYPES = [SYSTEM, HYPERVISOR, CANDLEPIN].freeze
 
       lazy_accessor :entitlements, :initializer => lambda { |_s| Resources::Candlepin::Consumer.entitlements(uuid) }
-      lazy_accessor :events, :initializer => lambda { |_s| Resources::Candlepin::Consumer.events(uuid) }
```
jturel: I think the events method itself can also be removed from the Resource class

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/details/views/content-host-details.html
```diff
@@ -69,23 +69,11 @@
...
         </a>
-        <ul uib-dropdown-menu>
-          <li>
-            <a ui-sref="content-host.subscriptions.list" translate>
-              Subscriptions
-            </a>
-          </li>
-          <li>
-            <a ui-sref="content-host.events" translate>
-              Events
```
jturel: I think there should be another angular template for the actual contents of the Events tab which can be removed

sseelam2: Oh yes. I forgot to remove that. I'm on it.

app/services/katello/candlepin/consumer.rb
jturel: Looks like the Owner resource also has an events API and we aren't using it. We can remove that too

```
[vagrant@centos7-devel katello]$ git grep "\.events" app/
app/models/katello/glue/candlepin/owner.rb:        lazy_accessor :events, :initializer => lambda { |_s| Resources::Candlepin::Owner.events(label) }
```

app/controllers/katello/api/v2/host_subscriptions_controller.rb
```diff
@@ -55,8 +55,7 @@ def auto_attach
     api :GET, "/hosts/:host_id/subscriptions/events", N_("List subscription events for the host")
     param :host_id, Integer, :desc => N_("Id of the host"), :required => true
     def events
-      collection = full_result_response(@host.subscription_facet.candlepin_consumer.events.map { |e| OpenStruct.new(e) })
-      respond_for_index :collection => collection
+      respond_for_index :collection => []
```
jturel: Let's add a deprecation warning to this API. Check out our usage of `::Foreman::Deprecation.api_deprecation_warning` in some controllers. Since this code will land in 3.14 we can mention that the API will be removed in 3.16

app/controllers/katello/api/v2/host_subscriptions_controller.rb
sseelam2: @jturel collection was a hash, that's why I had returned an empty hash earlier. I am getting this error message if I use an array.
`{"displayMessage":"no implicit conversion of Symbol into Integer","errors":["no implicit conversion of Symbol into Integer"]}`

Ideally the events api should be returning the below - 
`{"total":null,"subtotal":null,"page":null,"per_page":null,"error":null,"search":null,"sort":{"by":null,"order":null},"results":null}`

So I'm changing it back to an empty hash. Let me know if that makes sense.

jturel: I see what's happening. The collection should actually be a hash as you mentioned because of the view template:


https://github.com/Katello/katello/blob/32e7c74eba444e9c20048b0e587b1aca2d19cf3e/app/views/katello/api/v2/host_subscriptions/events.json.rabl#L5


I think `results` should not be null in the response, though. That could break scripts or maybe even hammer in some way (not sure) if someone is trying to iterate over null. Can you put a `results` key in the hash with value of []?

sseelam2: Okay yes. That makes sense.

app/controllers/katello/api/v2/host_subscriptions_controller.rb
```diff
@@ -52,11 +52,11 @@ def auto_attach
...
     end
 
-    api :GET, "/hosts/:host_id/subscriptions/events", N_("List subscription events for the host")
+    api :GET, "/hosts/:host_id/subscriptions/events", N_("List subscription events for the host"), :deprecated => true
     param :host_id, Integer, :desc => N_("Id of the host"), :required => true
     def events
-      collection = full_result_response(@host.subscription_facet.candlepin_consumer.events.map { |e| OpenStruct.new(e) })
-      respond_for_index :collection => collection
+      ::Foreman::Deprecation.api_deprecation_warning("Events API is being deprecated and will be removed in Katello 3.16.")
+      respond_for_index :collection => {results: []}
```
jturel: Ugh, sorry @sseelam2 .  I noticed that with this, the paging and other result metadata in the response doesn't get populated. I think we need to use the `full_result_reponse` method like before, but with an empty collection. Try it out!

jturel: ie collection => full_result_response([]) should do the trick

sseelam2: @jturel this is ready to retest.

PR #8352, opened 2019-09-18
Fixes #27878 - hide Registered Through for hypervisor profiles

snagoor:
hide Registered Through option for hypervisor profiles on the Content Hosts page

### DIFF
--- a/app/views/katello/api/v2/subscription_facet/base.json.rabl
+++ b/app/views/katello/api/v2/subscription_facet/base.json.rabl
@@ -1,4 +1,4 @@
-attributes :id, :uuid, :last_checkin, :service_level, :release_version, :autoheal, :registered_at, :registered_through, :purpose_role, :purpose_usage
+attributes :id, :uuid, :last_checkin, :service_level, :release_version, :autoheal, :registered_at, :registered_through, :purpose_role, :purpose_usage, :hypervisor
 
 child :user => :user do
   attributes :id, :login
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/details/views/content-host-info.html
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/details/views/content-host-info.html
@@ -54,8 +54,8 @@
           </a>
         </dd>
 
-        <dt translate>Registered Through</dt>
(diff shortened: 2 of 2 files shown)


jturel:
Couldn't a Hypervisor be registered through a Capsule / Foreman Proxy w/ Content ?

snagoor:
@jturel agreed, when virt-who used in VDSM mode, we register the hypervisor directly to either Satellite/Capsule which indeed has a `registered_through` value set. So, to satisfy both these conditions can I tweak the condition as below?
~~~
ng-show="!host.subscription_facet_attributes.hypervisor && host.subscription_facets.registered_through == null"
~~~

jturel:
I'm still not sure what you're trying to solve.

Is it that, for cases where the hypvervisor is created by the virt-who checkin, you want to hide "Registered Through: Unknown" completely (rather than showing 'Unknown')?

snagoor:
> I'm still not sure what you're trying to solve.
> 
> Is it that, for cases where the hypvervisor is created by the virt-who checkin, you want to hide "Registered Through: Unknown" completely (rather than showing 'Unknown')?

Yes, that's what I am trying to accomplish. +1:

jturel:
> > I'm still not sure what you're trying to solve.
> > Is it that, for cases where the hypvervisor is created by the virt-who checkin, you want to hide "Registered Through: Unknown" completely (rather than showing 'Unknown')?
> 
> Yes, that's what I am trying to accomplish. +1

Cool, thank you. Got it! How about this:

```
ng-show="host.subscription_facet_attributes.registered_through || !host.subscription_facet_attributes.hypervisor"
```

snagoor:
@jturel, I think one of the condition is enough to hide `Registered Through` this from showing up on UI? 
~~~
ng-show="!host.subscription_facet_attributes.hypervisor"
~~~
should be sufficient IMHO, Correct me if I am wrong.

jturel:
> @jturel, I think one of the condition is enough to hide `Registered Through` this from showing up on UI?
> 
> ```
> ng-show="!host.subscription_facet_attributes.hypervisor"
> ```
> 
> should be sufficient IMHO, Correct me if I am wrong.

I think we need both checks because if a hypervisor is registered directly with subscription-manager we should still show registered-through, since the value will be set

snagoor:
@jturel modified PR as per your suggestion. Please review and let me know if any further changes are required.

PR #8358, opened 2019-09-20
Fixes #27882 - Container CV version has duplicate tag count

ianballou:
Non-archived repos were erroneously increasing the docker tag count for latest cv versions with docker content. 

To test:
1) Sync a repo with docker content
2) Create a content view with that repo and promote it once
3) Notice that there's double the tags than there should be under the latest CV version's Content
4) Apply this PR
5) Notice that the tag count is now correct

### DIFF
--- a/app/models/katello/content_view_version.rb
+++ b/app/models/katello/content_view_version.rb
@@ -298,7 +298,8 @@ module Katello
     end
 
     def docker_tags
-      ::Katello::DockerMetaTag.where(:id => RepositoryDockerMetaTag.where(:repository_id => repositories.docker_type).select(:docker_meta_tag_id))
+      # Don't count tags from non-archived repos; this causes count errors
+      ::Katello::DockerMetaTag.where(:id => RepositoryDockerMetaTag.where(:repository_id => repositories.archived.docker_type).select(:docker_meta_tag_id))
     end
 
     def debs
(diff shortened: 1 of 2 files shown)


app/models/katello/content_view_version.rb
```diff
@@ -298,7 +298,8 @@ def docker_tag_count
     end
 
     def docker_tags
```
jturel: Can you modify the test here so that it would fail without your change?

https://github.com/Katello/katello/blob/c5116b3e40fc6dcfd569cb0bd81207f8dbaf7195/test/models/content_view_version_test.rb#L94

ianballou: I've added a part to the test where an un-archived repo with a docker_meta_tag is introduced.  This returns a count of 2 rather than one if looking at `repositories.docker_type` rather than `repositories.archived.docker_type`.

ianballou: I've added a part to the test where an un-archived repo with a docker_meta_tag is introduced. This returns a count of 2 rather than one if looking at repositories.docker_type rather than repositories.archived.docker_type.

jturel:
I'm still seeing duplicate repos:
Screenshot from 2019-09-23 17-08-37


I created a new product & repo for busybox, synced it, and published it in a new content view

ianballou:
@jturel what I fixed was the docker tag count on the CV version page.  I think it's intentional that we see the two repos there, but maybe we don't have to.

jturel:
> @jturel what I fixed was the docker tag count on the CV version page where all of the included content is listed (you should see 135 there instead of double that). Before, the tags of the two repos were being added up. I think it's intentional that we see the two repos on the page in your screenshot, but maybe we don't have to.

That ... makes a lot more sense! :laughing:

PR #8359, opened 2019-09-23
Fixes #27211 - Publish Pulp 3 Docker CV with Filters

ianballou:
Adds CV filter support for Pulp 3 CVs with Docker content.  I've also re-factored some of the logic in `copy_all_units.rb` to hopefully make it easier to understand, but let me know if anyone has better ideas for re-factoring that.

To test:
1) Spin up a Pulp 3 dev box: https://projects.theforeman.org/projects/katello/wiki/Pulp_3_Integration
2) Sync a repo with some Docker content
3) Publish a CV with some Docker content using both include and exclude filters
4) Ensure correct content count in Katello (keeping in mind that [this bug](https://projects.theforeman.org/issues/27882) might still be around)
5) Ensure correct content count in Pulp 3 by querying the new CV version repo's version_href using the Pulp 3 API

~~Note: to test this correctly, the API endpoint in the Pulp 3 recursive-add binding code needs to be updated to include /pulp/api/v3/.  Ensure `local_var_path = '/pulp/api/v3/docker/recursive-add/'` in `docker_recursive_add_api.rb`~~

~~Related Pulp 3 issue: https://pulp.plan.io/issues/5486~~
The above has been fixed by bumping the docker client version.


app/services/katello/pulp3/repository/docker.rb
```diff
@@ -49,6 +57,36 @@ def distribution_options(path)
...
+          whitelist_ids = []
+          blacklist_ids = []
+          existing_units = source_repository.docker_manifests
+          filters.each do |filter|
+            if filter.inclusion
+              whitelist_ids += filter.get_pulp_ids(source_repository)
+            else
+              blacklist_ids += filter.get_pulp_ids(source_repository)
+            end
+          end
```
jlsherrill: does this handle the case where i *only* blacklist an item?  Whereby i'd expect everything to be copied except that one item?    It looks like it might copy nothing, but i haven't fully tested it yet

ianballou: It does. I check if there's no whitelist, and if there's none, I copy all tags minus the blacklist files on line 85.

jlsherrill: oh doh!  i missed that line, looks good!   I will do some testing

ianballou:
Looks like VCRs are having some issues, re-recording them now.

app/models/katello/content_view_docker_filter.rb
```diff
@@ -21,6 +21,15 @@ def generate_clauses(repo)
...
     end
 
+    def get_pulp_ids(repo)
+      manifest_tags = []
+      self.docker_rules.each do |rule|
+        manifest_tags.concat(query_manifests(repo, rule))
+      end
+
+      manifest_tags
+    end
```
jlsherrill: we usually try to avoid 'get_' methods, maybe just name it 'pulp_ids'?  or 'content_unit_pulp_ids' to be more specific/

test/actions/pulp3/orchestration/copy_all_units_test.rb
```diff
@@ -6,16 +6,81 @@ class CopyAllUnitsTest < ActiveSupport::TestCase
...
+      ForemanTasks.sync_task(::Actions::Pulp3::Orchestration::Repository::CopyAllUnits,
+                             @docker_clone, @master, [@docker_repo], filters: [filter])
+      @docker_clone.reload
+      @docker_clone.index_content
+      ::Katello::DockerMetaTag.import_meta_tags([@docker_clone])
+
+      refute_nil(@docker_repo.version_href)
+      refute_nil(@docker_clone.version_href)
+      assert_not_equal @docker_repo.version_href, @docker_clone.version_href
+      assert_equal @docker_clone.docker_tags.pluck(:name).sort, ["latest", "uclibc"]
```
jlsherrill: nice job on the .sort :+1:

jlsherrill:
@ianballou nice job!!! Works great and testing looks sufficient.  One small nitpick and a rubocop failure, but looks good!

PR #8361, opened 2019-09-23
Fixes #27875 - Content View missing from Host Collections Add Table

ianballou:
Does anyone have an opinion about whether I use `content_view.name` or `content_view_name`?

To test:

1) Ensure there's at least one host registered with Foreman
2) Check Host Collections -> (Some Collection) -> List Hosts -> Add
3) Check that the Content View isn't blank

### DIFF
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/host-collections/details/views/host-collection-add-hosts.html
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/host-collections/details/views/host-collection-add-hosts.html
@@ -44,7 +44,7 @@
           </a>
         </td>
         <td bst-table-cell >{{ host.content_facet_attributes.lifecycle_environment.name}}</td>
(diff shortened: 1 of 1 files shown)


engines/bastion_katello/app/assets/javascripts/bastion_katello/host-collections/details/views/host-collection-add-hosts.html
```diff
@@ -44,7 +44,7 @@
           </a>
         </td>
         <td bst-table-cell >{{ host.content_facet_attributes.lifecycle_environment.name}}</td>
-        <td bst-table-cell >{{ host.content_facet_attributes.name}}</td>
+        <td bst-table-cell >{{ host.content_facet_attributes.content_view.name}}</td>
```
jturel: This seems more correct than `content_view_name`. What makes me say that is we added 'content_view' directly on the API response after `content_view_name`, so I'd interpret that to mean `content_view_name` is the legacy attribute

PR #8362, opened 2019-09-23
Fixes #27911 - new pulp3 cancel tasks api

jlsherrill:


sjha4:
**State of task in pulp without change:** 
Goes to completed:
https://codebeautify.org/jsonviewer/cbaec1cd

**State of task in pulp with change: **
Goes to cancelled.
https://codebeautify.org/jsonviewer/cb36133f

**Minor note:** The progress reports say running for subtasks in case of cancellation. However, the task is cancelled.

PR #8363, opened 2019-09-24
Fixes #27714 - Generate bastion config on load

parthaa:
This commit adds a config generator option to bastion so that it reloads
the configurations on every page load. Prior to this commit the plugins
were returning a static list of config items, which means if some one
changed "remote_execution_by_default" setting, the config would not get
regenerated in the UI.

### DIFF
--- a/engines/bastion/lib/bastion.rb
+++ b/engines/bastion/lib/bastion.rb
@@ -24,6 +24,7 @@ module Bastion
     }
 
     Bastion.plugins.each do |_name, plugin|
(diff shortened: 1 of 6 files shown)


parthaa:
1. Goto -> any katello pages like Products
2. Inspect Element and search for the word remoteExecutionByDefault
3. You should see something like here
4. Administer -> Settings -> Content, set remote_execution_by_default to Yes
5. Go to the any katello page with bastion
6. Inspect Element and search for the word remoteExecutionByDefault 
7. You should see something like here


Before the PR the remoteExecutionByDefault value would not get updated without restarting the server

PR #8364, opened 2019-09-26
Fixes #27910 - Hammer support to upload duplicate files

sjha4:
This PR contains supporting changes to allow duplicate content uploads in pulp3 from hammer (https://github.com/Katello/hammer-cli-katello/pull/699).

### DIFF
--- a/app/controllers/katello/api/v2/content_uploads_controller.rb
+++ b/app/controllers/katello/api/v2/content_uploads_controller.rb
@@ -9,8 +9,12 @@ module Katello
     api :POST, "/repositories/:repository_id/content_uploads", N_("Create an upload request")
     param :repository_id, :number, :required => true, :desc => N_("repository id")
     param :size, :number, :required => true, :desc => N_("Size of file to upload")
+    param :checksum, String, :required => true, :desc => N_("Checksum of file to upload")
+    param :content_type, RepositoryTypeManager.uploadable_content_types.map(&:label), :required => false, :desc => N_("content type ('deb', 'docker_manifest', 'file', 'ostree', 'puppet_module', 'rpm', 'srpm')")
     def create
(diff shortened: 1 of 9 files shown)


sjha4:
[test katello]

app/controllers/katello/api/v2/content_uploads_controller.rb
```diff
@@ -9,8 +9,12 @@ class Api::V2::ContentUploadsController < Api::V2::ApiController
     api :POST, "/repositories/:repository_id/content_uploads", N_("Create an upload request")
     param :repository_id, :number, :required => true, :desc => N_("repository id")
     param :size, :number, :required => true, :desc => N_("Size of file to upload")
+    param :checksum, String, :required => true, :desc => N_("Checksum of file to upload")
```
jlsherrill: since this param is not used for pulp2, can we make it optional (not required?).  I'd prefer to not break api compatibility until we absolutely have to.  Maybe have a check in the pulp3 content service that raises an exception if its not passed in for pulp3 types.

sjha4: So in hammer, regardless of pulp2/pulp3, we are passing checksum to keep the hammer part consistent. So this check, while useless for pulp2 is satisfied from hammer. It is easier and cleaner this way. What do you think?

jlsherrill: right, i'm not worried as much about hammer, as i am about other api users.  Which it probably still works if you're not using pulp2 right?  so in that sense it is optional under certain situations?

Can you make it optional and just always pass it with hammer?  That way individual api users won't be super confused

jlsherrill: I opened this https://projects.theforeman.org/issues/28106 to track making it required.

jjeffers:
LGTM - tested earlier, and the Pulp3 uploads worked as indicated.

PR #8366, opened 2019-09-26
Fixes #27674 - Move event pollers off Dynflow

jturel:
This PR moves our two event listeners 'Monitor Event Queue'  and  'Listen On Candlepin Events' out of Dynflow and into a runner that we control entirely. The daemon will be started inside the rails process similar to Dynflow(in dev) and will be restarted on code changes.

In production environments we use Passenger's [smart spawning](https://www.phusionpassenger.com/library/indepth/ruby/spawn_methods/#the-smart-spawning-method) which runs multiple processes forked from a parent process which preloads the Rails code. The preloader is where the daemon will run, and so we have only one instance even if passenger is configured to run a large number of Foremans.

This change is also compatible in the future when we switch from Passenger to Puma for production. The plan is to run Puma in a single process (vs clustered mode), but with multiple threads. In this scenario we will still end up with only a single instance of the new event daemon. Even in clustered mode, due to how Puma forks its worker processes, we still end up with a single instance of the daemon.

**Simulating Events**

Test incoming Katello events (among the usual workflows) via rails console like this:
```
Katello::Event.create!(event_type: 'import_host_applicability', object_id: 16)
```

Test Candlepin events by, among other ways, changing the service level on a content host, or adding/removing subscriptions.
```
Katello::Resources::Candlepin::Consumer.update(::Host.find(214).subscription_facet.uuid, {'role' => SecureRandom.uuid})
```

**Monitoring**
- the "About" page prints out processed ct/error ct/queue depth for katello and candlepin event processors

**Considerations**

- I tried to make the pollers somewhat durable like they were in Dynflow. If the database or qpid connections are lost they will indefinitely attempt to reconnect and continue processing. This is an area I'd really like to be scrutinized and get feedback on.

- I noticed that the Qpid library has its own mechanism to reconnect to the broker when connection is lost. I tried to use it but it appears to block the main thread when doing so, which I felt wasn't great (cant even ctrl+c the process)

TODO:

- [x] more tests
- [x] some load testing in a prod environment
- [x] custom loggers


app/services/katello/event_monitor/poller_thread.rb
```diff
@@ -0,0 +1,85 @@
+module Katello
+  module EventMonitor
+    class PollerThread
+      SLEEP_INTERVAL = 2
```
jturel: Where did the 2 second value come from? https://devblogs.microsoft.com/oldnewthing/20190924-00/?p=102915

lib/katello/engine.rb
```diff
@@ -85,15 +85,6 @@ class Engine < ::Rails::Engine
       end
     end
 
-    initializer "katello.initialize_cp_listener", :before => :finisher_hook do
```
jturel: I'm thinking we could run both of the event monitors here just like in the rake task for dev?

app/services/katello/event_monitor/poller_thread.rb
jlsherrill: hah, i think it was just the idea of we don't want it to wait too long (30 seconds might be too long).  It can be changed to 1 or 3 most likely

lib/katello/engine.rb
jlsherrill: makes sense, as long as it doesn't cause any issues :+1:

app/services/katello/event_daemon.rb
```diff
@@ -0,0 +1,45 @@
...
+        Rails.logger.debug("Not starting Katello Event Daemon in a console session")
+        in_console = true
+      end
+
+      return if in_console
+
+      should_run = if via_rake
+                     true
+                   else
+                     !Foreman.in_rake? && !Rails.env.test? && SETTINGS[:katello][:with_event_monitor]
```
jlsherrill: i assume the setting was to allow a developer to choose to not run it?

app/services/katello/event_monitor/poller_thread.rb
jlsherrill: CandlepinListeningService had a similar sleep, but was just 1 second

jlsherrill:
Overall +1 to the general idea/flow

app/services/katello/event_daemon.rb
jturel: Yep, it's false by default and I plan to puppetize it to true for dev

app/services/katello/event_monitor/poller_thread.rb
jturel: I think I like 3 too

lib/katello/engine.rb
jturel: I think I resolved the daemon running in the spring process by checking $PROGRAM_NAME in EventDaemon!

app/services/katello/event_daemon.rb
jturel: Also so that we can start rails in production mode in a dev environment, if we want, with the event monitor running which we wouldn't do in a real deployment

app/services/katello/event_daemon.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      if $PROGRAM_NAME.match(/spring/)
+        Rails.logger.info("Spring is running in the environment, not starting Katello Event Daemon")
+        return false
+      end
+
+      if Rails.env == 'production' && $PROGRAM_NAME.match(/dynflow/)
+        Rails.logger.info("This appears to be the Dynflow process, not starting Katello Event Daemon")
+        return false
+      end
```
jlsherrill: what about a random rake task, would it get started in those?

jturel: See few lines down, checking `Foreman.in_rake?` there

jlsherrill: ahh!!! :+1:

chris1984:
[test katello]

ehelms:
From some side conversations, the concerns I raised were around monitoring of these two pollers once they are out of dynflow. The request was two fold:

 1) Can we add custom logger for these to identify issues specific to them in production.log
 2) Can we add status reporting to the Ping API and About page to quickly know if they are functioning properly

We also discussed can we do any error rate detection for reporting to give admins an indicator that either is stuck or misbehaving.

jturel:
[test katello]

jturel:
@ehelms here's what the about page looks like now:

daemon_status

It uses the rails cache to store the status values since the poller is not running in the process handling web requests. It stores values for 24 hours or clears them out when the server is restarted.

jturel:
> From some side conversations, the concerns I raised were around monitoring of these two pollers once they are out of dynflow. The request was two fold:
> 
>     1. Can we add custom logger for these to identify issues specific to them in production.log

Will do
> 
>     2. Can we add status reporting to the Ping API and About page to quickly know if they are functioning properly

Done
> 
> 
> We also discussed can we do any error rate detection for reporting to give admins an indicator that either is stuck or misbehaving.

I don't remember talking about that. How to you think we should accomplish it? And can we do it in as a follow-on to this PR?

jturel:
@ehelms  updated!

app/services/katello/event_monitor/poller_thread.rb
```diff
@@ -0,0 +1,114 @@
...
+              @logger.warn("event_queue_event_expired: type=#{event.event_type} object_id=#{event.object_id}")
+            elsif !result.nil?
+              @logger.warn("event_queue_rescheduled: type=#{event.event_type} object_id=#{event.object_id}")
+            end
+          end
+          ::Katello::EventQueue.clear_events(event.event_type, event.object_id, event.created_at)
+        end
+      end
+
+      def poll_for_events
```
johnpmitsch: @jturel I'm not sure if this is still applicable here, but we had to wrap the calls inside the thread in a `Rails.application.executor.wrap`, see https://github.com/Katello/katello/pull/7261/files because of the dev server hanging on code changes.

The discussion and some rails docs are in the PR message and I can rack my brain for more details if you need them. can you take a look and see if its still a concern here :)

ehelms:
The About page looks great! I think the failure count there addresses the error detection as an introductory level. The general gist of the error detection was for admins to know if one of the pollers is acting in a way that they should investigate and potentially fix. Avoiding the case if we can of them dieing without anyone noticing and critical functionality being non-operational without anyone knowing.

Mind posting an example output from the ping API? (I'm lazy and don't want to spin up a Katello :P)

johnpmitsch:
> (I'm lazy and don't want to spin up a Katello :P)

@ehelms Sounds like you could [kill two birds with one stone](https://github.com/theforeman/forklift/pull/1028)  :stuck_out_tongue_winking_eye:

app/services/katello/event_monitor/poller_thread.rb
jturel: I'm glad you mentioned this! I took great care and tracked down the discussion when I came across that application executor bit. Take a look at my changes in engine.rb to see how I accomplished code reloading. The gist of it is that when code is reloaded the two event listeners are completely stopped and started again on the next request, and that seems to work perfectly. I have not seen any hangups!

jturel:
> The About page looks great! I think the failure count there addresses the error detection as an introductory level. The general gist of the error detection was for admins to know if one of the pollers is acting in a way that they should investigate and potentially fix. Avoiding the case if we can of them dieing without anyone noticing and critical functionality being non-operational without anyone knowing.
> 
> Mind posting an example output from the ping API? (I'm lazy and don't want to spin up a Katello :P)

Thanks! API looks like this:

```
{
"status": "ok",
"services": {
"pulp": {
"status": "ok",
"duration_ms": "35"
},
"pulp_auth": {
"status": "ok",
"duration_ms": "15"
},
"candlepin": {
"status": "ok",
"duration_ms": "16"
},
"candlepin_auth": {
"status": "ok",
"duration_ms": "11"
},
"foreman_tasks": {
"status": "ok",
"duration_ms": "2"
},
"katello_events": {
"status": "ok",
"message": "1 Processed, 1 Failed, 0 in queue",
"duration_ms": "1"
},
"candlepin_events": {
"status": "ok",
"message": "6 Processed, 0 Failed, 0 in queue",
"duration_ms": "18"
}
}
}
```

One thing to be aware of, and may be worth discussing, is that I did not build any condition to set status to "fail" (unless there's an exception while pulling the status). My thinking was that the current behavior of the app is not hindered if these two listeners aren't running, so I wanted to be consistent with that. I think a fail state could prevent some rake tasks from running, or may affect hammer ping usage in undesirable ways.

app/services/katello/event_monitor/poller_thread.rb
johnpmitsch: Thanks, good to hear you put some thought into it before omitting, I'll poke around and review!

lib/katello/engine.rb
```diff
@@ -26,6 +26,7 @@ class Engine < ::Rails::Engine
         :redhat_repository_url => 'https://cdn.redhat.com',
         :consumer_cert_rpm => 'katello-ca-consumer-latest.noarch.rpm',
         :consumer_cert_sh => 'katello-rhsm-consumer',
+        :event_monitor_enabled => true,
```
johnpmitsch: Where is this setting used?

ehelms:
Is there an opportunity to have the message be a structured response for easier parsing? e.g.

```
message: {
  "processed": 6,
  "failed": 1,
  "queued": 0
}
```

Maybe for now we don't change the overall status but the individual status for the queues does get set to failed? Ideally, that would trigger an overall status of "warning" perhaps rather than failed.

johnpmitsch:
> My thinking was that the current behavior of the app is not hindered if these two listeners aren't running, so I wanted to be consistent with tha

Could you elaborate on this? Wouldn't functionality be significantly affected if these listeners aren't running?

jturel:
> Is there an opportunity to have the message be a structured response for easier parsing? e.g.
> 
> ```
> message: {
>   "processed": 6,
>   "failed": 1,
>   "queued": 0
> }
> ```
> 

Outside the scope of this PR in my opinion but it seems useful

> Maybe for now we don't change the overall status but the individual status for the queues does get set to failed? Ideally, that would trigger an overall status of "warning" perhaps rather than failed.

I'll see if that's possible, I like the idea

lib/katello/engine.rb
jturel: I think I refactored its usage away but I'd like to add it back. Never know when that knob may be handy.

jturel:
> > My thinking was that the current behavior of the app is not hindered if these two listeners aren't running, so I wanted to be consistent with tha
> 
> Could you elaborate on this? Wouldn't functionality be significantly affected if these listeners aren't running?

Katello can run just fine without them since the operations are totally asynchronous due to being event driven. That's the benefit of using messages. If qpid is down and messages get backed up, things will start moving when qpid comes back online. No visible errors or breakage of critical workflows (ignoring pulp2, but I'm speaking from katello perspective). Whereas if Candlepin is not running Katello is pretty much unusable.

In production right now, if the event listeners are not running, we don't say via hammer ping for example that service is degraded. That may or may not be correct but I want the focus of this PR to be 'get 1:1 functionality with production today, but without dynflow'  I'm going to look into Eric's suggestion to see if it's feasible ('warning' status)

jeremylenz:
+1 to @ehelms comment above about structuring the response for easier parsing.  I dealt with the same situation a few weeks ago, and ended up putting together a really wonky frontend solution that was traversing the human-readable sentence looking for colons and commas, and parsing it into something usable.  I then realized that was ridiculous and just changed the API.

jturel:
> +1 to @ehelms comment above about structuring the response for easier parsing. I dealt with the same situation a few weeks ago, and ended up putting together a really wonky frontend solution that was traversing the human-readable sentence looking for colons and commas, and parsing it into something usable. I then realized that was ridiculous and just changed the API.

It doesn't feel like the right time to undertake this change. All of the Katello::Ping messages are strings and changing this would mean it could be a string or a hash!

jeremylenz:
@jturel ah, so it's not just a few lines of RABL then.  Makes sense :)

ehelms:
> . All of the Katello::Ping messages are strings and changing this would mean it could be a string or a hash!

That is fair -- it would seem to warrant a new field for this metadata. I can live without the structured data as long as soon as somebody starts trying to parse that message we implement it :)

I think it worth breaking down what the result of either listener being down is (this is based on my knowledge so correct me where I go astray):

 * Candlepin events: consumption updates would get out of date, and a user may experience something like a system registration that fails to lack of available pools because the Katello and Candlepin data have drifted. This is more of an inconvenience.

 * Katello queue: applicability calculations will be out of date and if a critical security errata were to arrive, there is potential that an affected system may not be accurately reporting that it needs the fix leaving it vulnerable. This is critical path.

I do appreciate the argument for 1:1 parity. Its hard to at a minimum, use this as a forum for discussion around both ensuring this switch maintains the level of visibility and action for the admin as well as identify improvements we can make now or in follow up PRs. 

Given the dynflow changes are pushing off of 1.24, is the intent for this to still go into 3.14?

app/services/katello/event_monitor/poller_thread.rb
```diff
@@ -0,0 +1,114 @@
...
+        }
+      end
+
+      def self.reset_status
+        Rails.cache.write(PROCESSED_COUNT_CACHE_KEY, 0)
+        Rails.cache.write(FAILED_COUNT_CACHE_KEY, 0)
+      end
+
+      def initialize(logger = nil)
+        @logger = logger || ::Foreman::Logging.logger('katello/katello_events')
```
ehelms: :fireworks:

lib/katello/engine.rb
```diff
@@ -26,6 +26,7 @@ class Engine < ::Rails::Engine
         :redhat_repository_url => 'https://cdn.redhat.com',
         :consumer_cert_rpm => 'katello-ca-consumer-latest.noarch.rpm',
         :consumer_cert_sh => 'katello-rhsm-consumer',
+        :event_daemon_enabled => true,
```
ehelms: What happens when I set this to false? Is there a occasion to do that?

jturel: It doesn't run the event listeners. It could be useful if you want to create a backlog of events for load testing, or another reason we'll inevitably run into!

app/services/katello/event_daemon.rb
```diff
@@ -0,0 +1,60 @@
+module Katello
+  class EventDaemon
+    def self.runnable?
+      if defined?(Rails::Console)
+        Rails.logger.info("Not starting Katello Event Daemon in the Rails Console")
+        return false
+      end
+
+      if $PROGRAM_NAME.match(/spring/)
+        Rails.logger.info("Spring is running in the environment, not starting Katello Event Daemon")
```
jeremylenz: You once made me say "erratum" so I feel like I have the right to point out these comma splices.  :wink:
```suggestion
        Rails.logger.info("Spring is running in the environment; not starting Katello Event Daemon")
```

jturel:
> > . All of the Katello::Ping messages are strings and changing this would mean it could be a string or a hash!
> 
> That is fair -- it would seem to warrant a new field for this metadata. I can live without the structured data as long as soon as somebody starts trying to parse that message we implement it :)

:fearful: 
> 
> I think it worth breaking down what the result of either listener being down is (this is based on my knowledge so correct me where I go astray):
> 
>     * Candlepin events: consumption updates would get out of date, and a user may experience something like a system registration that fails to lack of available pools because the Katello and Candlepin data have drifted. This is more of an inconvenience.

I'm not sure registration would be blocked since that's all proxied to candlepin. Pool data would be out of date though. Host could appear to be subscribed to a subscription that's gone. Content host compliance status & syspurpose status would be stale also.

> 
>     * Katello queue: applicability calculations will be out of date and if a critical security errata were to arrive, there is potential that an affected system may not be accurately reporting that it needs the fix leaving it vulnerable. This is critical path.
> 

The two event listeners are tied together for the pool related actions. Katello events do errata applicability like you mentioned, and also handle auto-publish of composite content views.



> 
> I do appreciate the argument for 1:1 parity. Its hard to at a minimum, use this as a forum for discussion around both ensuring this switch maintains the level of visibility and action for the admin as well as identify improvements we can make now or in follow up PRs.
> 
> Given the dynflow changes are pushing off of 1.24, is the intent for this to still go into 3.14?

Yes, 3.14!

parthaa:
> > Is there an opportunity to have the message be a structured response for easier parsing? e.g.
> > ```
> > message: {
> >   "processed": 6,
> >   "failed": 1,
> >   "queued": 0
> > }
> > ```
> 
> Outside the scope of this PR in my opinion but it seems useful

I wonder if you could keep the message the same but add a second
```
result:  {
  "processed": 6,
   "failed": 1,
  "queued": 0
}
```

Sure its redundant but message could be construed as a description field

parthaa:
> * The suggested Rails console commands in the PR description create candlepin_events and katello_events which show on the About page
> * Logging messages are added in Rails logger
> * New log messages now appear when you start the Rails server, alerting you that it's using 3 threads
> * The two Foreman tasks mentioned no longer appear

Tested the same myself :) .. Doubly Verified:)

johnpmitsch:
Had a bit of strange behavior... not sure what happened but I'll do my best to explain:

After checking out the PR and restart services and the foreman server, I was seeing events in queue but not processed
```
katello_events	OK	0 Processed, 0 Failed, 4 in queue
```
I had imported a manifest, registered a host, attached subscriptions to the host and ran a couple command line events, all went fine and no errors

I also didn't see any candlepin events processed or in queue

My server was just running for a while and then I restarted it and saw an error
 ```
E, [2019-10-24T00:17:34.336113 #23201] ERROR -- /parallel-executor-core: uninitialized constant Actions::Katello::EventQueue (NameError)
```
and [the traceback is here](https://paste.fedoraproject.org/paste/AeLPctKvHdPpuRmSbJ~odg)

I restarted the server again, and then I saw in the logs all these messages coming through and processed. The good thing is they were all processed fine, this is reflected in the count:
```
katello_events	OK	8 Processed, 0 Failed, 0 in queue
candlepin_events	OK	27 Processed, 0 Failed, 0 in queue
```
I'm not sure if this is a fluke, but given the issues we see with multi-threaded processes raising errors, its probably worth trying to re-create the error and test the stability of these processes.

I am out tomorrow, but can check in and also will look more Friday.

**Comment by jturel** [comment]
> > > Is there an opportunity to have the message be a structured response for easier parsing? e.g.
> > > ```
> > > message: {
> > >   "processed": 6,
> > >   "failed": 1,
> > >   "queued": 0
> > > }
> > > ```
> > 
> > 
> > Outside the scope of this PR in my opinion but it seems useful
> 
> I wonder if you could keep the message the same but add a second
> 
> ```
> result:  {
>   "processed": 6,
>    "failed": 1,
>   "queued": 0
> }
> ```
> 
> Sure its redundant but message could be construed as a description field

I like the idea, but I think that could be done as part of a larger analysis of our status API & ping service

**Comment by jturel** [comment]
> Had a bit of strange behavior... not sure what happened but I'll do my best to explain:
> 
> After checking out the PR and restart services and the foreman server, I was seeing events in queue but not processed
> 
> ```
> katello_events	OK	0 Processed, 0 Failed, 4 in queue
> ```
> 
> I had imported a manifest, registered a host, attached subscriptions to the host and ran a couple command line events, all went fine and no errors
> 
> I also didn't see any candlepin events processed or in queue
> 
> My server was just running for a while and then I restarted it and saw an error
> 
> ```
> E, [2019-10-24T00:17:34.336113 #23201] ERROR -- /parallel-executor-core: uninitialized constant Actions::Katello::EventQueue (NameError)
> ```
> 
> and [the traceback is here](https://paste.fedoraproject.org/paste/AeLPctKvHdPpuRmSbJ~odg)
> 
> I restarted the server again, and then I saw in the logs all these messages coming through and processed. The good thing is they were all processed fine, this is reflected in the count:
> 
> ```
> katello_events	OK	8 Processed, 0 Failed, 0 in queue
> candlepin_events	OK	27 Processed, 0 Failed, 0 in queue
> ```
> 
> I'm not sure if this is a fluke, but given the issues we see with multi-threaded processes raising errors, its probably worth trying to re-create the error and test the stability of these processes.
> 
> I am out tomorrow, but can check in and also will look more Friday.

My guess is this is somehow spring related - try stopping or restarting spring

File: app/services/katello/event_daemon.rb (hunk 9 above)
Comment (jturel) [reply app/services/katello/event_daemon.rb]: I can appreciate that :) will update

**Comment by jturel** [comment]
@ehelms katello and candlepin events will now show a WARN status which does not affect overall app status. WARN means the queue depth is > 1000 messages. I chose WARN because the system could be running slowly and  catching up. I opted to not look at the error count because it's misleading in this context. Since error count is cached for 24 hours (or until service is restarted) reporting FAIL because some events failed 12 hours ago doesn't seem great to me. What do you think?

**Comment by johnpmitsch** [comment]
> My guess is this is somehow spring related - try stopping or restarting spring

@jturel I always stop spring so I don't think that is it. Not sure what is going on, but I'll see if I can reproduce consistently and share a way to reproduce if I can

File: app/models/katello/ping.rb [hunk 10]
```diff
@@ -2,7 +2,9 @@ module Katello
   class Ping
     OK_RETURN_CODE = 'ok'.freeze
     FAIL_RETURN_CODE = 'FAIL'.freeze
+    WARN_RETURN_CODE = 'WARN'.freeze
```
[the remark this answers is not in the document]
jturel: Yeah, I preferred the upcase a e s t h e t i c

File if you want to see consistency here :+1:

app/services/katello/candlepin_event_listener.rb
```diff
@@ -0,0 +1,104 @@
...
+
+    def self.run
+      @thread.kill if @thread
+
+      # run in own thread so connecting to qpid won't block the main process
+      @thread = Thread.new do
+        initialize_listening_service
+        start_service
+
+        ActiveRecord::Base.connection_pool.with_connection do
```
johnpmitsch: @jturel Revisiting the `Rails.application.executor.wrap` method, I was looking at [the Rails docs on threading](https://guides.rubyonrails.org/threading_and_code_execution.html) and this part jumped out at me:

> Prior to Rails 5.0, some of these were handled by separate Rack middleware classes (such as ActiveRecord::ConnectionAdapters::ConnectionManagement), or directly wrapping code with methods like ActiveRecord::Base.connection_pool.with_connection. The Executor replaces these with a single more abstract interface.

It seems like they are suggesting to use either ` `Rails.application.executor.wrap`  or `
Rails.application.reloader.wrap` instead of `ActiveRecord::Base.connection_pool.with_connection` for Rails 5 apps. We should probably stick to convention Rails specifies here, but maybe you saw some reason to use the `with_connection` method over the application wrappers. Any thoughts?

jturel: I didn't realize that was handled by the executor! When I was researching I just came across with_connection. The application.executor should work, but just want to mention that all that needs to be considered here (I think) is the database connection since the reloading is handled in engine.rb So, some preciseness is lost in the intention of the call, but the documentation does seem to indicate using the executor is preferred.

jturel:
[test katello]

jturel:
@ehelms @johnpmitsch 

It appears that feedback has tapered off. If there isn't anything else, please endorse with an approval. I say this especially since 3.14 branching is today and it'd be great to not have to pick this over.

@johnpmitsch the issue you reported around the stack trace coming from dynflow - my feeling is that it wouldn't be an issue in production since we do things more gracefully there when updating such as ensuring no running tasks and bringing services up and down at the right times. it looks like something was left lingering around in dynflow in reference to the old action, but mine cleaned up after itself, or so it seems.

app/services/katello/candlepin_event_listener.rb
jturel: Updated to use Rails.application.executor.wrap

johnpmitsch:
> it looks like something was left lingering around in dynflow in reference to the old action, but mine cleaned up after itself, or so it seems.

Gotcha, just something to keep an eye out for when this is merged and as it goes through our testing process, I agree it is not worth delaying if we can't reproduce.

jturel:
> I can't say anything from a runtime perspective. From the design perspective looks great! I wouldn't mind an issue filed to look into notifications if queue depth gets too large to pro-actively warn a user that something is amiss.

Thank you. Filed https://projects.theforeman.org/issues/28141

PR #8367, opened 2019-09-30
Fixes #27773 - Prefer Pulp 2 for some content types

ianballou:
This PR enables content types to be configured to use either Pulp 2 or Pulp 3 when both exist in one Katello environment.  The default is Pulp 3.

Things to test:
1) Create and sync some Pulp 2 repo
2) Create and sync some Pulp 3 repo
3) Proxy sync Pulp 2 and Pulp 3 repositories (note that Apache isn't configured to serve Pulp 2 content using the Pulp 3 dev environment, which will cause the proxy sync to fail)

### DIFF
--- a/app/models/katello/concerns/smart_proxy_extensions.rb
+++ b/app/models/katello/concerns/smart_proxy_extensions.rb
@@ -143,10 +143,19 @@ module Katello
         pulp3_repository_type_support?(repository.content_type)
       end
 
+      def pulp2_preferred_for_type?(repository_type)
+        if SETTINGS[:katello][:use_pulp_2_for_content_type].nil? ||
+            SETTINGS[:katello][:use_pulp_2_for_content_type][repository_type.to_sym].nil?
+          return false
+        else
+          return SETTINGS[:katello][:use_pulp_2_for_content_type][repository_type.to_sym]
+        end
+      end
+
       def pulp3_repository_type_support?(repository_type)
         repository_type_obj = repository_type.is_a?(String) ? Katello::RepositoryTypeManager.repository_types[repository_type] : repository_type
         fail "Cannot find repository type #{repository_type}, is it enabled?" unless repository_type_obj
-        repository_type_obj.pulp3_plugin.present? && pulp3_enabled? && self.capabilities(PULP3_FEATURE).try(:include?, repository_type_obj.pulp3_plugin)
+        repository_type_obj.pulp3_plugin.present? && pulp3_enabled? && self.capabilities(PULP3_FEATURE).try(:include?, repository_type_obj.pulp3_plugin) && !pulp2_preferred_for_type?(repository_type_obj.id)
       end
 
       def pulp3_content_support?(content_type)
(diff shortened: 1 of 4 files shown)


app/lib/actions/katello/capsule_content/sync.rb
```diff
@@ -33,8 +33,15 @@ def plan(smart_proxy, options = {})
...
-          smart_proxy.ping_pulp if repositories.any? { |repo| !smart_proxy.pulp3_support?(repo) }
+
+          smart_proxy.ping_pulp3 if repositories.any? do |repo|
+            smart_proxy.pulp3_support?(repo) &&
+            !smart_proxy.pulp2_preferred_for_type?(repo.content_type.to_sym)
+          end
+          smart_proxy.ping_pulp if repositories.any? do |repo|
+            !smart_proxy.pulp3_support?(repo) ||
+            smart_proxy.pulp2_preferred_for_type?(repo.content_type.to_sym)
+          end
```
jlsherrill: i would just remove this for capsule syncing, as we won't ever have a capsule with mixed types, and this setting only applies to the 'master' pulp server.

app/lib/actions/pulp3/orchestration/repository/refresh_repos.rb
```diff
@@ -15,7 +15,8 @@ def fetch_proxy_service(smart_proxy)
           end
 
           def act_on_repo?(repo, smart_proxy)
-            smart_proxy.pulp3_support?(repo)
+            smart_proxy.pulp3_support?(repo) &&
```
jlsherrill: if .pulp2_support? supports this setting, is it needed here?

app/lib/actions/pulp/orchestration/repository/refresh_repos.rb
```diff
@@ -16,7 +16,8 @@ def fetch_proxy_service(smart_proxy)
           end
 
           def act_on_repo?(repo, smart_proxy)
-            !smart_proxy.pulp3_support?(repo)
+            !smart_proxy.pulp3_support?(repo) ||
+              smart_proxy.pulp2_preferred_for_type?(repo.content_type.to_sym)
```
jlsherrill: same question :)

jlsherrill: couple more down below as well

app/lib/actions/katello/pulp_selector.rb
```diff
@@ -14,7 +14,11 @@ def plan_optional_pulp_action(backend_actions, repository, smart_proxy, *args)
       end
 
       def select_pulp_action(backend_actions, repository, smart_proxy)
-        backend_type = smart_proxy.backend_service_type(repository)
+        if smart_proxy.pulp2_preferred_for_type?(repository.content_type.to_sym)
+          backend_type = "pulp2"
+        else
+          backend_type = smart_proxy.backend_service_type(repository)
+        end
```
jlsherrill: since backend_service_type calls pulp3_support?, i don't think we need this particular check

jlsherrill:
overall works well!  just some duplicate checking i think

PR #8370, opened 2019-10-03
Fixes #27974 - RHSM checkin requests stuck for long time

hao-yu:
Cache the list of virtual guests of the hypervisors before going
into the database transaction. This will prevent the transaction
to lock the updating rows in subscription facet table for long time
and causes all RHSM checkin requests to stuck. Also improve the
update query to minimize the number of rows to be locked.

### DIFF
--- a/app/services/katello/candlepin/consumer.rb
+++ b/app/services/katello/candlepin/consumer.rb
@@ -100,9 +100,10 @@ module Katello
       end
 
       def virtual_guests
+        return @virtual_guests unless @virtual_guests.nil?
         return [] if self.uuid.nil?
         guest_uuids = Resources::Candlepin::Consumer.virtual_guests(self.uuid).map { |guest| guest['uuid'] }
-        ::Host.joins(:subscription_facet).where("#{Katello::Host::SubscriptionFacet.table_name}.uuid" => guest_uuids)
+        @virtual_guests = ::Host.joins(:subscription_facet).where("#{Katello::Host::SubscriptionFacet.table_name}.uuid" => guest_uuids)
       end
 
       def virtual_host
--- a/test/actions/katello/host/hypervisors_update_test.rb
+++ b/test/actions/katello/host/hypervisors_update_test.rb
@@ -49,6 +49,8 @@ module Katello::Host
         @host.subscription_facet.destroy!
         @host.destroy!
 
+        ::Katello::Resources::Candlepin::Consumer.expects(:virtual_guests).never
+
         action = create_action(::Actions::Katello::Host::HypervisorsUpdate)
 
         plan_action(action, :hypervisors => @hypervisor_results)
@@ -59,6 +61,32 @@ module Katello::Host
         @host = Host.find_by(:name => @hypervisor_name)
         assert_not_nil @host.subscription_facet
         assert_equal @facts['hypervisor.type'], @host.facts['hypervisor::type']
+      end
+
+      it 'update guests hypervisor' do
+        guests = []
+        original_host = FactoryBot.create(:host, :with_subscription, :content_view => @content_view,
+                                          :lifecycle_environment => @content_view_environment, :organization => @organization)
+
+        3.times do
+          guests << FactoryBot.create(:host, :with_subscription, :content_view => @content_view,
+                                      :lifecycle_environment => @content_view_environment, :organization => @organization)
+        end
+
+        guests.first.subscription_facet.update_attributes!(:hypervisor_host_id => original_host.id)
+        guests.sort!
+        guest_uuids = guests.map { |guest| { 'uuid' => guest.subscription_facet.uuid } }
+
+        # Delete :guestIds to make katello to fetch the virtual guests from Candlepin
+        @consumer.delete(:guestIds)
+        ::Katello::Resources::Candlepin::Consumer.expects(:virtual_guests).once.returns(guest_uuids)
+
+        action = create_action(::Actions::Katello::Host::HypervisorsUpdate)
+        plan_action(action, :hypervisors => @hypervisor_results)
+        action = run_action(action)
+
+        action.state.must_equal :success
+        assert_equal guests, @host.subscription_facet.virtual_guests.sort
       end
 
       it 'existing hypervisor, no facet' do
--- a/test/services/katello/candlepin/consumer_test.rb
+++ b/test/services/katello/candlepin/consumer_test.rb
@@ -62,6 +62,16 @@ module Katello
 
         assert_equal nil, Candlepin::Consumer.distribution_to_puppet_os('RedHot')
       end
+
+      def test_virtual_guests
+        hosts = [hosts(:one), hosts(:two)]
+        guest_uuids = hosts.map { |host| { 'uuid' => host.subscription_facet.uuid } }
+
+        Resources::Candlepin::Consumer.expects(:virtual_guests).once.returns(guest_uuids)
+
+        assert_equal hosts, @consumer.virtual_guests.sort
+        assert_equal hosts, @consumer.virtual_guests.sort
+      end
     end
   end
 end
--- a/app/lib/actions/katello/host/hypervisors_update.rb
+++ b/app/lib/actions/katello/host/hypervisors_update.rb
@@ -18,6 +18,8 @@ module Actions
               load_resources
             end
 
+            load_hosts_guests
+
             ActiveRecord::Base.transaction do
               @hosts.each do |uuid, host|
                 update_host(uuid, host)
@@ -27,6 +29,16 @@ module Actions
             @hosts.each do |uuid, host|
               update_facts(uuid, host)
             end
+          end
+        end
+
+        def load_hosts_guests
+          @hosts.each do |uuid, host|
+            host.subscription_facet ||= host.build_subscription_facet(uuid: uuid)
+            # Only preload the virtual guests if 'guestIds' is not returned by Candlepin
+            consumer = @candlepin_attributes[uuid]
+            next unless consumer && consumer.try(:[], 'guestIds').empty?
+            host.subscription_facet.candlepin_consumer.virtual_guests
           end
         end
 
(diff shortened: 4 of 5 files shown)


jturel:
[test katello]

hao-yu:
hmm...the failed test doesn't seem related to this patch and it is passed in my vm.

jturel:
[test katello]

hao-yu:
Just wondering how is the review and testing of this patch going? I think it will be good to have this issue solved asap since this is quite critical for Satellite that running virt-who.

jturel:
@hao-yu  I'm reviewing now. I'll need to prepare some test data to ensure it's achieving the desired effect but the approach makes sense to me

test/services/katello/candlepin/consumer_test.rb
```diff
@@ -62,6 +62,16 @@ def test_distribution_to_puppet_os
 
         assert_equal nil, Candlepin::Consumer.distribution_to_puppet_os('RedHot')
       end
+
+      def test_virtual_guests
+        guests = []
```
jturel: Thanks for adding a test for this method!

What if we wrote it this way, to make it more readable, and reduce interaction with the database by using an existing 'fixture' instead of creating three new hosts?

```
      def test_virtual_guests                                                                                                                                                
              facet = katello_subscription_facets(:one)                                                                                                                      
              guests = [facet.host]                                                                                                                                          
              guest_uuids = [{'uuid' => facet.uuid}]                                                                                                                         
                                                                                                                                                                             
              Resources::Candlepin::Consumer.expects(:virtual_guests).once.returns(guest_uuids)                                                                              
                                                                                                                                                                             
              assert_equal guests, @consumer.virtual_guests                                                                                                                  
              assert_equal guests, @consumer.virtual_guests                                                                                                                  
      end
```

test/actions/katello/host/hypervisors_update_test.rb
```diff
@@ -59,6 +73,7 @@ class HypervisorsUpdateTest < ActiveSupport::TestCase
         @host = Host.find_by(:name => @hypervisor_name)
         assert_not_nil @host.subscription_facet
         assert_equal @facts['hypervisor.type'], @host.facts['hypervisor::type']
+        assert_equal @guests, @host.subscription_facet.virtual_guests.sort
```
jturel: It's hard to tell by looking what is achieved by adding this assertion + the additional code you added to the setup method

An entirely new test in this file could be the solution (moving the setup code there too). Since there is no explicit test, someone could very easily come along and unknowingly remove the behavior which you are trying to test!

test/services/katello/candlepin/consumer_test.rb
hao-yu: Sure. will do.

test/actions/katello/host/hypervisors_update_test.rb
hao-yu: Since this test file is to the the hypervisor update, my reason to add the extra codes to in the setup method is to make the data reusable across tests in the future. Minimum @guests and @hosts data are the basic data for the testing the hypervisor update I guess. The new assertion that I added is to strengthen the test case which is what we should check for the hypervisor update run. Maybe adding a comment explaining what the assertion trying to achieve will work?

jturel: Looking at it now, I understand better. I agree with your approach - up to you if you want to add a comment about it, but it makes sense if you understand what the code is trying to do :)

hao-yu: I ended up adding a new test. :D

test/services/katello/candlepin/consumer_test.rb
hao-yu: Done

hao-yu:
I also fix the codes to preload virtual guests only if guestIds is not returned by Candlepin

jturel:
[test katello]

jturel:
Hey, this is still on my to-do list, but I haven't been able to reproduce the original issue yet by following the steps in the BZ.

hao-yu:
@jturel Are you following the comment 5 in the bugzilla? The requests will stuck even longer time like 4mins+ if you have 20+ hypervisors and 5000 content hosts/guests.

jturel:
> @jturel Are you following the comment 5 in the bugzilla? The requests will stuck even longer time like 4mins+ if you have 20+ hypervisors and 5000 content hosts/guests.

I have 10 hypervisors and 200 guests for each one. Will have time tomorrow to check it out again - thanks for your patience

jturel:
At last I was able to test after reproducing the problem and I must say: nice fix! I didn't realize how much time was spent loading the virtual guests. In my environment it took 40 seconds (almost the entire time it takes to complete the whole action) to fetch the guest data from candlepin. I took a look at the API and it turns out there's a much faster endpoint for loading the virtual guests. It returns a smaller dataset and is a good fit since we only need the UUID. I did this to test it out:

```
diff --git a/app/lib/katello/resources/candlepin/consumer.rb b/app/lib/katello/resources/candlepin/consumer.rb
index a56e7f6..2c376c9 100644
--- a/app/lib/katello/resources/candlepin/consumer.rb
+++ b/app/lib/katello/resources/candlepin/consumer.rb
@@ -150,6 +150,13 @@ module Katello
             return []
           end
 
+          def virtual_guest_ids(uuid)
+            response = Candlepin::CandlepinResource.get(join_path(path(uuid), 'guestids'), self.default_headers).body
+            ::Katello::Util::Data.array_with_indifferent_access JSON.parse(response)
+          rescue RestClient::Exception
+            return []
+          end
+
           def virtual_host(uuid)
             response = Candlepin::CandlepinResource.get(join_path(path(uuid), 'host'), self.default_headers).body
             if response.present?
diff --git a/app/services/katello/candlepin/consumer.rb b/app/services/katello/candlepin/consumer.rb
index 6dab118..0c22a5e 100644
--- a/app/services/katello/candlepin/consumer.rb
+++ b/app/services/katello/candlepin/consumer.rb
@@ -102,7 +102,7 @@ module Katello
       def virtual_guests
         return @virtual_guests unless @virtual_guests.nil?
         return [] if self.uuid.nil?
-        guest_uuids = Resources::Candlepin::Consumer.virtual_guests(self.uuid).map { |guest| guest['uuid'] }
+        guest_uuids = Resources::Candlepin::Consumer.virtual_guest_ids(self.uuid).map { |guest| guest['guestId'] }
         @virtual_guests = ::Host.joins(:subscription_facet).where("#{Katello::Host::SubscriptionFacet.table_name}.uuid" => guest_uuids)
       end
 
```

This way takes less than 1 second (0.74) to load the guests for 20 hypervisors! Would you be interested in including this as part of your change? Together I think this becomes an even better enhancement.

jturel:
@jlsherrill I've asked Hao to change to the `/consumers/:uuid/guestids` API to load virtual guests on the Consumer service. I was surprised we hadn't been using it since hypervisor checkins are known resource hog. Any concerns / do you share my surprise too ?

hao-yu:
@jturel the proposing api is fetching the guest ids but we are looking for guest uuids so the api won't work unless we make the following change too.

`diff --git a/app/services/katello/candlepin/consumer.rb b/app/services/katello/candlepin/consumer.rb
index f961e6b..24b01ec 100644
--- a/app/services/katello/candlepin/consumer.rb
+++ b/app/services/katello/candlepin/consumer.rb
@@ -100,9 +100,13 @@ module Katello
       end
 
       def virtual_guests
+        return @virtual_guests unless @virtual_guests.nil?
         return [] if self.uuid.nil?
-        guest_uuids = Resources::Candlepin::Consumer.virtual_guests(self.uuid).map { |guest| guest['uuid'] }
-        ::Host.joins(:subscription_facet).where("#{Katello::Host::SubscriptionFacet.table_name}.uuid" => guest_uuids)
+        guest_ids = Resources::Candlepin::Consumer.virtual_guest_ids(self.uuid).map { |guest| guest['guestId'].downcase }
+        fact_name_id = RhsmFactName.find_by(:name => 'virt::uuid').id
+        @virtual_guests = ::Host.joins(:fact_values).
+                                 where("lower(#{FactValue.table_name}.value) IN (?) AND #{FactValue.table_name}.fact_name_id = ?", guest_ids, fact_name_id).
+                                 uniq
       end`

hao-yu:
@jturel submitted the new patch. Also added 2 new fixture files. not sure if i done it correctly. Let me know if there is any issue. Thanks.

hao-yu:
According to my test, this change has significant speed up the hypervisor update tasks. Previously took about 1.30min now only 20secs.

app/services/katello/candlepin/consumer.rb
```diff
@@ -100,9 +100,13 @@ def pool_ids
...
 
       def virtual_guests
+        return @virtual_guests unless @virtual_guests.nil?
         return [] if self.uuid.nil?
-        guest_uuids = Resources::Candlepin::Consumer.virtual_guests(self.uuid).map { |guest| guest['uuid'] }
-        ::Host.joins(:subscription_facet).where("#{Katello::Host::SubscriptionFacet.table_name}.uuid" => guest_uuids)
+        guest_ids = Resources::Candlepin::Consumer.virtual_guest_ids(self.uuid).map { |guest| guest['guestId'].downcase }
+        fact_name_id = RhsmFactName.find_by(:name => 'virt::uuid').id
+        @virtual_guests = ::Host.joins(:fact_values).
+                                 where("lower(#{FactValue.table_name}.value) IN (?) AND #{FactValue.table_name}.fact_name_id = ?", guest_ids, fact_name_id).
```
hao-yu: Hmm... I am kind of worry this query when the facts table are extremely large.

jturel: I agree with you. Maybe it would be better to go back to your original implementation and remove the virtual_guest_ids call.

I think there are two possible future improvements:

1) store the guest id on SubscriptionFacet for fast lookup based on /guestids API
2) new Candlepin API to retrieve guest UUIDs only (this is my preference)

jturel: I filed a bug for Candlepin to provide an API for us https://bugzilla.redhat.com/show_bug.cgi?id=1772159

In the future we can switch to use it

hao-yu: Great! Thanks for creating the bug.  I have rewinded the commit. Let me know if you found any issue.

jturel:
[test katello]

PR #8373, opened 2019-10-08
Fixes #28018 - Show installed pkgs for Report Template

snagoor:
Creating a new method `host_installed_packages` for it to be used in the Report Templates for fetching the installed packages on the host.

One of the use-case is to generate a report to see if a particular package is installed on all hosts example. If `katello-agent` is installed or not on all Content Hosts then we can do the below from the Report Templates

~~~
'Katello Agent Status': host_installed_packages(host).include?('katello-agent')
~~~

### DIFF
--- a/app/lib/katello/concerns/base_template_scope_extensions.rb
+++ b/app/lib/katello/concerns/base_template_scope_extensions.rb
@@ -6,7 +6,8 @@ module Katello
       module Overrides
         def allowed_helpers
           super + [:errata, :host_subscriptions, :host_applicable_errata_ids, :host_applicable_errata_filtered,
-                   :host_latest_applicable_rpm_version, :load_pools, :load_errata_applications, :host_content_facet]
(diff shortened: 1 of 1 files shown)


app/lib/katello/concerns/base_template_scope_extensions.rb
```diff
@@ -38,6 +39,10 @@ def host_latest_applicable_rpm_version(host, package)
         host.applicable_rpms.where(name: package).order(:version_sortable).limit(1).pluck(:nvra).first
       end
 
+      def host_installed_packages(host)
+        host.installed_packages.map(&:name)
```
sjha4: I think we should rename this to host_installed_package_names since we are only returning names.

snagoor:
@sjha4 update the method name to `host_installed_package_names` as per your suggestion.

snagoor:
[test katello]

snagoor:
@sjha4 closing this in favor of https://github.com/Katello/katello/pull/8389

PR #8378, opened 2019-10-10
Fixes #28037 - Incremental Update additional repos

parthaa:
This commit handles incremental update with additional_repos parameters.
Content from one repo can depend on content from another repo, this is
especially true with RHEL 8. This causes a problem during depsolving.
For example if we have a module/errata in one repo pointing packages in
another repo, we'd like to ideally copy the dependencies across both
repositories in the same operation.
In the incremental update case for example the user may request a
certain errata to be copied over along with their dependencies.
To address this Pulp 2.21 updated their copy api to accept
"additional repos" (https://pulp.plan.io/issues/5067.)

This commit sets up the appropriate mapping for this operation. So if
the content view has depsolve turned on or if its incremental katello
will create a 'repository_mapping' which maps source repo to a
destination repo. The entries in this hash consist of all the
repositories in the content view version minus the main one.
For example
CVV1 - repos A,B
On incremental update CVV1 -> CVV 1.1
The copy operation for content in repo A will  have {b:dest-b} as the
"additional_repos" aka "repository mapping".
This mapping is then passed to pulp copy api.

This change has a couple of complexities. We need all the destination
repositories to be created before we call the copy operations with the
addition_repos mapping. To handle this a new CreateRepos plan was
created and the CloneToVersion plan was modified to accept a
destination repo. The CloneToVersion call now expects the destination
repo to have been created before copying.
Changes were made to both IncrementalUpdate and Publish plans to
accomodate the new structure.

### DIFF
--- a/app/lib/actions/katello/content_view_version/incremental_update.rb
+++ b/app/lib/actions/katello/content_view_version/incremental_update.rb
@@ -16,8 +16,9 @@ module Actions
         end
 
         # rubocop:disable Metrics/MethodLength
+        # rubocop:disable Metrics/AbcSize
         def plan(old_version, environments, options = {})
-          dep_solve = options.fetch(:resolve_dependencies, false)
+          dep_solve = options.fetch(:resolve_dependencies, true)
           description = options.fetch(:description, '')
           content = options.fetch(:content, {})
           new_components = options.fetch(:new_components, [])
@@ -36,11 +37,25 @@ module Actions
                                                           :notes => description)
 
           copy_action_outputs = []
+          repos_to_clone = repos_to_copy(old_version, new_components)
 
           sequence do
+            repository_mapping = plan_action(ContentViewVersion::CreateRepos, new_content_view_version, repos_to_clone).repository_mapping
+
+            repos_to_clone.each do |source_repos|
+              plan_action(Repository::CloneToVersion,
+                          source_repos,
+                          new_content_view_version,
+                          repository_mapping[source_repos],
+                          incremental: true)
+            end
+
             concurrence do
-              repos_to_copy(old_version, new_components).each do |source_repos|
-                copy_action_outputs += copy_repos(source_repos, new_content_view_version, content, dep_solve)
+              repos_to_clone.each do |source_repos|
+                copy_action_outputs += copy_repos(repository_mapping[source_repos],
+                                                  new_content_view_version,
+                                                  content,
+                                                  dep_solve)
               end
 
               sequence do
@@ -64,13 +79,14 @@ module Actions
           end
         end
 
-        def copy_repos(source_repos, new_version, content, dep_solve)
+        def copy_repos(new_repo, new_version, content, dep_solve)
           copy_output = []
           sequence do
-            new_repo = plan_action(Repository::CloneToVersion, source_repos, new_version, :incremental => true).new_repository
             solve_dependencies = new_version.content_view.solve_dependencies || dep_solve
             copy_output += copy_deb_content(new_repo, solve_dependencies, content[:deb_ids])
-            copy_output += copy_yum_content(new_repo, solve_dependencies, content[:package_ids], content[:errata_ids])
+            copy_output += copy_yum_content(new_repo, solve_dependencies,
+                                            content[:package_ids],
+                                            content[:errata_ids])
 
             plan_action(Katello::Repository::MetadataGenerate, new_repo)
             plan_action(Katello::Repository::IndexContent, id: new_repo.id)
@@ -121,8 +137,10 @@ module Actions
         def run
           content = { ::Katello::Erratum::CONTENT_TYPE => [],
                       ::Katello::Rpm::CONTENT_TYPE => [],
+                      ::Katello::ModuleStream::CONTENT_TYPE => [],
                       ::Katello::Deb::CONTENT_TYPE => [],
-                      ::Katello::PuppetModule::CONTENT_TYPE => []}
+                      ::Katello::PuppetModule::CONTENT_TYPE => []
+                    }
 
           input[:copy_action_outputs].each do |copy_output|
             copy_output[:pulp_tasks].each do |pulp_task|
@@ -132,6 +150,8 @@ module Actions
                 case type
                 when ::Katello::Erratum::CONTENT_TYPE
                   content[::Katello::Erratum::CONTENT_TYPE] << unit['id']
+                when ::Katello::ModuleStream::CONTENT_TYPE
+                  content[::Katello::ModuleStream::CONTENT_TYPE] << "#{unit['name']}:#{unit['stream']}:#{unit['version']}"
                 when ::Katello::Rpm::CONTENT_TYPE
                   content[::Katello::Rpm::CONTENT_TYPE] << ::Katello::Util::Package.build_nvra(unit)
                 when ::Katello::Deb::CONTENT_TYPE
@@ -222,7 +242,7 @@ module Actions
             unless deb_ids.blank?
               copy_outputs << plan_action(Pulp::Repository::CopyUnits, new_repo.library_instance, new_repo,
                                           ::Katello::Deb.with_identifiers(deb_ids),
-                                        resolve_dependencies: dep_solve).output
+                                          incremental_update: dep_solve).output
             end
           end
           copy_outputs
@@ -233,14 +253,14 @@ module Actions
           if new_repo.content_type == ::Katello::Repository::YUM_TYPE
             unless errata_ids.blank?
               copy_outputs << plan_action(Pulp::Repository::CopyUnits, new_repo.library_instance, new_repo,
-                                        ::Katello::Erratum.with_identifiers(errata_ids),
-                                        resolve_dependencies: dep_solve).output
+                                          ::Katello::Erratum.with_identifiers(errata_ids),
+                                          incremental_update: dep_solve).output
             end
 
             unless package_ids.blank?
               copy_outputs << plan_action(Pulp::Repository::CopyUnits, new_repo.library_instance, new_repo,
-                                        ::Katello::Rpm.with_identifiers(package_ids),
-                                        resolve_dependencies: dep_solve).output
+                                          ::Katello::Rpm.with_identifiers(package_ids),
+                                          incremental_update: dep_solve).output
             end
           end
           copy_outputs
--- a/app/lib/actions/pulp/repository/copy_units.rb
+++ b/app/lib/actions/pulp/repository/copy_units.rb
@@ -8,19 +8,16 @@ module Actions
                       target_repo_id: target_repo.id,
                       class_name: units.first.class.name,
                       unit_ids: units.pluck(:id),
-                      resolve_dependencies: options[:resolve_dependencies])
+                      incremental_update: options[:incremental_update])
           end
         end
 
         def invoke_external_task
           source_repo = ::Katello::Repository.find(input[:source_repo_id])
           target_repo = ::Katello::Repository.find(input[:target_repo_id])
-
           units = input[:class_name].constantize.where(:id => input[:unit_ids])
-
-          override_config = ::Katello::Repository.build_override_config(input)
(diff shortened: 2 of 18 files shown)


jturel:
[test katello]

parthaa:
To Test this PR
- Make sure your pulp is 2.21
- Turn pulp rest logger in config/settings.yaml 
```
:loggers:
  :katello/pulp_rest:
    :enabled: true
```
- Create and sync the following  repos -> https://partha.fedorapeople.org/test-repos/separated/modules-only/ and https://partha.fedorapeople.org/test-repos/separated/modular-rpms-only/ . The former only module definitions refering to rpms from some other repo  and the later repo has those missing rpms.
- Add both those repositories to a content view
- Setup a content view filter to include the Module Stream ->  "Walrus" version "20180707144203"
- Publish this content view 
- Notice  that only 1 module stream gets copied over.  No modular packages belonging to the module stream are copied over as such. This is because the rpms referred to by the "Walrus" module are in the rpms-only repo.
- Now go to the cv details and update the Content View to "Solve Dependencies" - Yes
- Publish the content view again.
- Notice that 1 module stream and 1 package got copied over. (i.e. the package in the rpms-only also got copied)

Grep the devel logs for the -> additional_repos you should see a call similar to 
```
\"override_config\":{\"recursive_conservative\":true,\"additional_repos\":{\"1c414117-2a20-4795-90eb-aa12f133c970\":\"1-foo-v14_0-1c414117-2a20-4795-90eb-aa12f133c970\"}}}"
```
- Try incrementally updating version by running a hammer command like ```hammer content-view version incremental-update --content-view-version-id=21 --errata-ids=RHEA-2012:0059```
- Make sure the minor version has 1 errata, 3 module streams and 3 packages

- Finally Remove the module filters and add an includes for Errata By Id. Include 'RHEA-2012:0059'
- Publish the cv and you should notice the errata + 2 module streams +  2 packages got copied to the destination

app/models/katello/glue/pulp/repo.rb
```diff
@@ -47,16 +47,25 @@ def self.needs_distributor_updates(repos, smart_proxy)
...
           config
         end
+
+        def self.generate_yum_dependent_repositories(cloned_repos = {})
+          ret = {}
+          cloned_repos.each do |source, dest|
+            ret[source.first.pulp_id] = dest.pulp_id if source.first.yum?
+          end
+          ret
+        end
```
jlsherrill: does this need to go in the glue layer?  Also, can it refer to repositories and NOT pulp ids?  Leave it to the service layer to figure out what is needed.

jlsherrill: if this returned a "RepositoryMapper" object, you could also convert it to a pulp_id -> pulp_id hash if that is useful in a service layer.

app/lib/actions/katello/content_view_version/incremental_update.rb
```diff
@@ -222,25 +242,31 @@ def copy_deb_content(new_repo, dep_solve, deb_ids)
...
+                                        resolve_dependencies: dep_solve,
+                                        incremental: true).output
             end
           end
           copy_outputs
         end
 
-        def copy_yum_content(new_repo, dep_solve, package_ids, errata_ids)
+        def copy_yum_content(new_repo, dep_solve, package_ids, errata_ids, repository_mapping)
+          new_repository_mapping = Hash[repository_mapping.map { |source_repos, dest| [source_repos.map(&:library_instance), dest] }]
```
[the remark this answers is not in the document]
jlsherrill: and to further this point, i see: 
``` ::Katello::Repository.generate_yum_dependent_repositories(new_repository_mapping)```

Its not really really clear what this is taking in, nor what its returning.  Some object that holds the mapping and maybe can do transformations on it might be helpful

File: app/models/katello/repository.rb [hunk 3]
```diff
@@ -276,29 +276,34 @@ def product_type
...
         "INNER JOIN #{repository_rpm} on #{repository_rpm}.rpm_id = #{rpm}.id").
-        where("#{repository_rpm}.repository_id" => repo.id).
-        where("#{repository_errata}.repository_id" => repo.id).
-        group("#{errata}.id").count
+         where(where_clause.join(" OR ")).
+         group("#{errata}.id").count
     end
 
-    def partial_errata
+    def partial_errata(repository_mapping = {})
```
jlsherrill: why exactly is repository_mapping needed to figure out partially included errata in a repository?

app/lib/actions/katello/content_view_version/incremental_update.rb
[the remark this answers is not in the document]
jlsherrill: Some further thought, would it be possible to calculate the repository mapping when we need it?  That way we wouldn't have to pass it around all over the place

app/models/katello/glue/pulp/repo.rb
parthaa: I need a way to serialize. I originally had {soure_repo:destination_repo} full blown objects. If you want I can do {source_repo_id:destination_repo_id} and then hav ethe service layer convert that

parthaa:
@jlsherrill  I ve removed repository mapping from many actions and updated the code according to our discussion. feel free to review

app/models/katello/repository.rb
```diff
@@ -315,6 +315,10 @@ def remove_partial_errata!
       found
     end
 
+    def siblings
+      content_view_version.archived_repos.where.not(:id => id)
+    end
```
jlsherrill: :+1: good naming!

app/services/katello/pulp/repository/yum.rb
```diff
@@ -75,19 +75,33 @@ def regenerate_applicability
...
+            key = sibling.library_instance? ? sibling.id : sibling.library_instance_id
+            source_repo_map[key] = sibling.pulp_id
+          end
+
+          Hash[destination_repo.siblings.map { |sibling| [source_repo_map[sibling.library_instance_id], sibling.pulp_id] }]
+        end
+
         def copy_contents(destination_repo, options = {})
-          override_config = Katello::Repository.build_override_config(options)
+          override_config = Katello::Repository.build_override_config(
```
jlsherrill: would you be open to moving the build_override_config into this service class? (since its pulp2 related).  Its only used by one more action (app/lib/actions/pulp/repository/copy_units.rb)

jlsherrill: you could probably have it handle injecting the mapping there as well (if we're dep solving) instead of .merge'ing it in below

app/services/katello/pulp/repository/yum.rb
```diff
@@ -75,19 +75,33 @@ def regenerate_applicability
...
+                                                options.merge(repository_mapping: generate_mapping(destination_repo)))
           rpm_copy_clauses, rpm_remove_clauses = generate_copy_clauses(options[:filters]&.yum(false), options[:rpm_filenames])
-          tasks = [smart_proxy.pulp_api.extensions.rpm.copy(repo.pulp_id, destination_repo.pulp_id,
-                   rpm_copy_clauses.merge(:override_config => override_config))]
+          tasks = []
+          if rpm_copy_clauses
+            tasks << smart_proxy.pulp_api.extensions.rpm.copy(repo.pulp_id, destination_repo.pulp_id,
```
jlsherrill: kind of a nitpick, but can you move the 'override_config' in this if statement?  That way its more clear its only used for rpm copying

jlsherrill: to clarify, move the definition of override_config into this if statement

jlsherrill:
Overall this is good!  I'm not sure if your testing instructions need updating now that we're only really worried about:

1. package deps across repositories (such as a package in some module in Appstream requiring some package in base)
2.  Module dependencies

Besides a few suggestions, i think this is mostly good to go from a code perspective.  If no one else can do testing, i can work on it on monday

parthaa:
> Besides a few suggestions, i think this is mostly good to go from a code perspective. If no one else can do testing, i can work on it on monday

I updated the test instructions and code as per your suggestions. Let me know if you have other comments

app/lib/actions/pulp/repository/copy_units.rb
```diff
@@ -8,19 +8,16 @@ def plan(source_repo, target_repo, units, options = {})
...
         def invoke_external_task
           source_repo = ::Katello::Repository.find(input[:source_repo_id])
           target_repo = ::Katello::Repository.find(input[:target_repo_id])
-
           units = input[:class_name].constantize.where(:id => input[:unit_ids])
-
-          override_config = ::Katello::Repository.build_override_config(input)
-
-          source_repo.backend_service(SmartProxy.pulp_master).copy_units(target_repo, units, override_config)
+          source_repo.backend_service(SmartProxy.pulp_master).copy_units(target_repo, units, input)
```
jlsherrill: can you pluck out what is needed in this input?

app/services/katello/pulp/repository.rb
```diff
@@ -259,7 +259,9 @@ def remove_unnecessary_distributors
         tasks
       end
 
-      def copy_units(destination_repo, units, override_config = {})
+      def copy_units(destination_repo, units, input = {})
+        override_config = {}
+        override_config = build_override_config(destination_repo, input) if respond_to? :build_override_config
```
jlsherrill: can we rename input to options here?

app/services/katello/pulp/repository/yum.rb
jlsherrill: thoughts on this?

ianballou:
When selecting dep solving there's a warning that says "Also, certain scenarios involving errata may still cause dependency errors.".  Does this fix that or can there still be dependency errors?

parthaa:
> When selecting dep solving there's a warning that says "Also, certain scenarios involving errata may still cause dependency errors.". Does this fix that or can there still be dependency errors? Curious if we can remove that warning.

I think that warning is still relevant. Dep Solving is best-effort only not the perfect solution ...

app/lib/actions/katello/content_view_version/incremental_update.rb
```diff
@@ -36,11 +37,25 @@ def plan(old_version, environments, options = {})
...
 
           sequence do
+            repository_mapping = plan_action(ContentViewVersion::CreateRepos, new_content_view_version, repos_to_clone).repository_mapping
+
+            repos_to_clone.each do |source_repos|
+              plan_action(Repository::CloneToVersion,
+                          source_repos,
+                          new_content_view_version,
+                          repository_mapping[source_repos],
+                          incremental: true)
```
jlsherrill: i just realized you were dep solving the initial clone_to_version for each repository, we don't actually want to do that right?  we just want to dep solve the things we're adding as part of the incremetnal update?

jlsherrill: unless i misunderstood?

parthaa: Yes that is unclear, but CloneToVersion uses 'incremental' for a different reason.
If you looked at
https://github.com/Katello/katello/blob/master/app/lib/actions/katello/repository/clone_to_version.rb#L8-L11
You would see the following line
```filters = incremental ? [] : content_view.filters.applicable(repositories.first)```
That is the only purpose of the 'incremental' flag

parthaa: We are not really dep solving  in CloneToVersion unless the content view has dep solve turned on.

parthaa:
[test katello]

parthaa:
[test katello]

app/services/katello/pulp/repository/yum.rb
ianballou: I agree with this too

parthaa: updated

parthaa:
[test katello]

parthaa:
[test katello]

parthaa:
[test katello]

PR #8380, opened 2019-10-10
Fixes #28038 - Update bastion katello test instructions

johnpmitsch:


engines/bastion_katello/README.md
```diff
@@ -8,9 +8,8 @@ The only real difference, as far as the user is concerned, is that the applicati
 
 ```
 cd ./engines/bastion_katello
-sudo npm update -g grunt-cli
+sudo npm install -g grunt-cli
 npm install
-npm install ../bastion/
```
jlsherrill: actually i think this installs bastion from the npm registry instead of from local git, which i don't think we want to do?

PR #8382, opened 2019-10-11
Fixes #28039 - shouldn't dep solve if no filters present

ianballou:
Dep resolving during CV publish should only occur if there are filters.
To test:
1) Enable `katello/pulp_rest` under `loggers` in foreman/settings.yaml
2) Set logging level to debug in foreman/settings.yaml
3) Make a CV with dependency solving on
4) Add yum content to the CV and a filter and publish it
5) Search the Foreman logs for `recursive_conservative`.  The publish should have logged this.
6) Do another publish but this time without any filters
7) Search again for `recursive_conservative`.  You should see it only for the first publish you made, not the latest one without filters.

### DIFF
--- a/app/models/katello/glue/pulp/repo.rb
+++ b/app/models/katello/glue/pulp/repo.rb
@@ -47,7 +47,7 @@ module Katello
 
         def self.build_override_config(options)
           config = {}
-          if options[:solve_dependencies] || options[:resolve_dependencies]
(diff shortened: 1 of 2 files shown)


parthaa:
LGTM. Can ya add a test?

ianballou:
@parthaa I've added my test

johnpmitsch:
[test katello]

PR #8385, opened 2019-10-16
Fixes #28074 - Add deprecate to api/ui for puppet/ostree repos

chris1984:
This PR adds a deprecate message to new/update operations on repos with the content type ostree/puppet.

I have verified it works correctly with the API and the UI as I am seeing this when I create/update a repo of those content types:

`15:25:01 rails.1   | 2019-10-16T15:25:01 [W|app|] DEPRECATION WARNING: Your API call uses deprecated behavior, Repository types of 'Puppet' and 'ostree' will no longer be supported in Katello 3.16. (called from plan at /home/vagrant/katello/app/lib/actions/katello/repository/create.rb:18)`

I added this to the types in the UI, but it looks awful:

https://s.nimbusweb.me/share/3430766/dkoyy1t22417an74r5nl

@jeremylenz I am trying to do something like Justin did here:

https://github.com/Katello/katello/pull/8376

But it seems we are dynamically pulling the content types here:

https://github.com/Katello/katello/blob/master/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/repository.factory.js#L23

Do you have a better way to add `(depreciated)` next to ostree/puppet like Justin did with download-policy?

### DIFF
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/new/views/new-repository.html
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/new/views/new-repository.html
@@ -39,6 +39,11 @@
         </select>
       </div>
 
+      <div class="alert alert-warning" ng-if="repository.content_type === 'ostree' || repository.content_type === 'puppet'">
+        <span class="pficon pficon-warning-triangle-o"></span>
+        <strong>Deprecation Warning: </strong>Puppet and OSTree will no longer be supported in Katello 3.16.
+      </div>
+
       <div bst-form-group label="{{ 'Restrict to Architecture' | translate }}" ng-show="repository.content_type === 'yum'">
         <select id="architecture_restricted"
                 name="architecture_restricted"
--- a/app/lib/actions/katello/repository/create.rb
+++ b/app/lib/actions/katello/repository/create.rb
@@ -14,6 +14,10 @@ module Actions
             ::Foreman::Deprecation.api_deprecation_warning("Background download_policy will be removed in Katello 3.16.  Any background repositories will be converted to Immediate")
           end
 
+          if root['content_type'] == 'puppet' || root['content_type'] == 'ostree'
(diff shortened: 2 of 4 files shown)


engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/new/views/new-repository.html
```diff
@@ -30,7 +30,7 @@ <h4 translate> Basic Information </h4>
                   type="text"/>
       </div>
 
-      <div bst-form-group label="{{ 'Type' | translate }}">
+      <div bst-form-group label="{{ 'Type (Puppet and ostree will no longer be supported in Katello 3.16.)' | translate }}">
```
jturel: I think the messaging is a good idea, but it might be better to add this as a separate block of text that way the translation for 'Type' won't be broken

jeremylenz:
@chris1984 Maybe it would look better with a patternfly 'Warning' widget?

https://www.patternfly.org/v3/pattern-library/widgets/#alerts

deprecation


You could remove the label attribute from your div and insert something like this instead:

```html
<div class="alert alert-warning">
  <span class="pficon pficon-warning-triangle-o"></span>
  <strong>Deprecation Warning: </strong>Puppet and Ostree will no longer be supported in Katello 3.16.
</div>
```

As a bonus, you could have the warning only show when one of those is selected (Angular can definitely do that, but I'd want to ask someone who knows it better than me)

chris1984:
@jturel @jlsherrill 

Updated and looking good :)

https://nimb.ws/nUz8jf recording of new repo

https://nimb.ws/k0pY4X ostree repo info page with warning

https://nimb.ws/rfNJj2 puppet repo info page with warning

chris1984:
@ianballou updated with your suggestion :)

chris1984:
[test katello]

chris1984:
@ianballou can I get a merge please :) Tests are green

PR #8387, opened 2019-10-16
Fixes #28026 - Pulp 3 attribute name changes

ianballou:
Updates attribute names to match recent name changes in Pulp 3.  Related: https://www.redhat.com/archives/pulp-dev/2019-October/msg00031.html


jlsherrill:
test failures i think are just tests that assume magic numbers.  asserting that content was added or  removed would be a better test.

ianballou:
> test failures i think are just tests that assume magic numbers. asserting that content was added or removed would be a better test.

I've changed it so it only checks for more content instead of the exact amount.

test/actions/pulp3/orchestration/ansible_collection_sync_test.rb
```diff
@@ -93,7 +93,7 @@ def test_sync_mirror_true
 
       collections = ::Katello::AnsibleCollection.in_repositories(@repo)
       post_count_content = collections.count
-      assert_equal pre_count_content + 2, post_count_content
+      assert_operator pre_count_content, :<, post_count_content
```
jlsherrill: actually looking at these two tests again, i see the first is testing mirror is false, the second is tesitng mirror is true. 
For mirror is false, i think the count check is good, but we should probably test also that the post_content also includes everything that is in the pre_content

for when mirror_on_sync is true, we should not really check the count at all but instead verify that none of the pre_content is in the post_content

ianballou: sounds good, I think I've taken care of that now.

PR #8388, opened 2019-10-17
Fixes #28080 - Remove RHSM facts when unregistering

jturel:
**To test**

1. register a client ie host.example.com (subman register)
2. unregister the client (subman unregister)
3. change hostname to host-a.example.com (hostnamectl)
4. attempt re-register (subman register)

(4) would have failed with an error, but now it should succeed since we don't keep the old registration's facts around

### DIFF
--- a/app/models/katello/concerns/host_managed_extensions.rb
+++ b/app/models/katello/concerns/host_managed_extensions.rb
@@ -79,6 +79,10 @@ module Katello
         self.organization.label
       end
 
+      def rhsm_fact_values
+        self.fact_values.joins(:fact_name).where("#{::FactName.table_name}.type = '#{Katello::RhsmFactName}'")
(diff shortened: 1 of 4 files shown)


jturel:
[test katello]

jturel:
@jlsherrill any thoughts on why we may have not deleted the RHSM facts while unregistering before? I'm guessing it's because it didn't really matter, but now it does - or can at least

jlsherrill:
@jturel seems very reasonable, i think the only reason we didn't previously was because no one ever asked for it :)

jturel:
[test katello]

jturel:
[test katello]

jturel:
[test katello]

chris1984:
Starting to test

PR #8390, opened 2019-10-17
Fixes #28085 - fixing columns in content view history

sseelam2:
Content view -> History -> Action column is broken when foreman tasks are deleted

Steps to Reproduce:
1. Create a content view, publish, promote and delete
2. Go to the content view page -> select the content view -> History tab
3. Action and status column displays correct information
4. Delete foreman-tasks and check Action and Status column

Actual results:
The Action column is blank for delete action and the Status column is blank for all the actions.

Expected results:
The action and status column should display normally after foreman tasks are deleted.


jturel:
[test katello]

chris1984:
Starting to test

PR #8391, opened 2019-10-17
Fixes #25876 - Disable auditing on org destroy

johnpmitsch:
When destroying an organization and environment, audits are created that are linked back to the Taxonomy (org or loc). This creates an issue when the taxonomy has been deleted and audits are created referring to the taxonomy.

This only shows up during concurrent actions. I'm not sure exactly why, maybe because of the multiple tasks modifying the same table means the organization will be destroyed while the audits are taking longer to write.

I was able to reproduce pretty consistently with script run on a hammer devel box:
```ruby

require 'net/http'

ITERATIONS=5

hammer="BUNDLE_GEMFILE=~/hammer-cli-foreman/Gemfile bundle exec hammer"

org_names = []
ITERATIONS.times do
  org_names << (0...8).map { (65 + rand(26)).chr }.join
end

puts "Creating Orgs\n"

base_create_command = "organization create"

org_names.each_with_index do |org_name, i|
  puts "creating org #{i+1}"
  cmd = `#{hammer} #{base_create_command} --name #{org_name} --label #{org_name}`
  puts cmd
end

puts "Deleting Orgs\n"

base_delete_command = "organization delete"

threads = []

org_names.each_with_index do |org_name, i|
  # spawn a new thread for each url
  threads << Thread.new do
    sleep rand(0..1)
    cmd = `#{hammer} #{base_delete_command} --name #{org_name}`
    puts "\nOrg #{i+1} deletion task complete \n#{cmd}\n"
  end
end

threads.each { |t| t.join }

puts "All Done!"
```

With this patch, I'm able to do 20 iterations without seeing deadlock or race condition issues.

### DIFF
--- a/app/lib/actions/katello/environment/destroy.rb
+++ b/app/lib/actions/katello/environment/destroy.rb
@@ -42,7 +42,11 @@ module Actions
 
         def finalize
           environment = ::Katello::KTEnvironment.find(input['kt_environment']['id'])
-          environment.destroy!
+
+          # CapsuleLifecycleEnvironment can cause issues when auditing, it will try to associate the audit to the deleted taxonomy
+          ::Katello::CapsuleLifecycleEnvironment.without_auditing do
+            environment.destroy!
+          end
         end
       end
     end


johnpmitsch:
[test katello]

app/lib/actions/katello/environment/destroy.rb
```diff
@@ -42,7 +42,11 @@ def humanized_input
 
         def finalize
           environment = ::Katello::KTEnvironment.find(input['kt_environment']['id'])
+
+          # Don't audit anything during this action as it can sometimes try to save audits to the taxonomy that was deleted.
+          Audited.auditing_enabled = false
           environment.destroy!
+          Audited.auditing_enabled = true
```
jlsherrill: I'm not sure we can use this method of disabling auditing since this is being run in a multi-threaded environment and this disables it globally and doesn't seem to do so in a threadsafe way.

What about something like KTEnvironment.without_auditing{  environment.destroy!  } ?  and same for below.  I'm not actually sure if this will disable audting for just the KTEnvironment class or everything that would be affected in that block, but its worth looking inot

johnpmitsch: ya, I wasn't sure how safe this would be, that is a good point.

I tried just specifying the audits for just the org or the environment, but it looks like the audits are still created for the dependent actions, like removing the default content view and cv env, which will still be audited. 

I'll look again, I can try nesting together a bunch of classes or figuring out a way to dynamically use `without_auditing` for a class and all its dependents. Maybe overriding the destroy action for the org and environment class would be a better place to do it too.

johnpmitsch:
I was able to narrow down the problem class to `Katello::CapsuleLifecycleEnvironment` and not audit it on delete. I'm unable to reproduce this issue after doing so.

@rplevka would there be a way to test this PR to see if it fixes QE's issue? I only ask because I was able to reproduce the error you were seeing, but the traceback was sligthly different, happening on the environment destroy action, not the org destroy action.

I have a [way to debug](https://gist.github.com/johnpmitsch/02268da2997fafb46ba3e2f625fd1f37) if you are still seeing the issue even with this patch. The change in that gist will log the association that is being audited and causing the issue. There may be more associations we need to not audit when an org is deleted

johnpmitsch:
@jlsherrill since you commented on a previous iteration, could you please confirm your concerns are addressed before I merge?

> @rplevka would there be a way to test this PR to see if it fixes QE's issue? I only ask because I was able to reproduce the error you were seeing, but the traceback was sligthly different, happening on the environment destroy action, not the org destroy action.

An update on this, I spoke with @rplevka off-thread and its going to best to see what happens to automation when this gets in a snap and take it from there!

jlsherrill:
@johnpmitsch yes, my concerns are alleviated.  This kinda feels like a one-off solution and the problem could come back, but i think its okay for now.

johnpmitsch:
> This kinda feels like a one-off solution and the problem could come back, but i think its okay for now.

Agreed, there probably needs to be a proper fix in the audited gem and/or foreman to not audit all associations on destroy

PR #8396, opened 2019-10-21
Fixes #28103 - allow API user to search pools using upstream_pool_id

wbclark:
In developing foreman ansible modules, it would be very nice for users of the modules to be able to use upstream_pool_id in playbooks which are written before Katello is deployed and any downstream pool ID has been created. Allowing users of the Katello API to search pools using upstream_pool_id will enable the module developers to use this functionality and create a better experience for users of the modules.


evgeni:
[test katello]

evgeni:
ok to test

evgeni:
[test foreman]

PR #8403, opened 2019-10-30
Refs #27674 - start daemon threads in single worker

jturel:
This PR builds on the work I did in #8366 to address some feedback I received in discourse: https://community.theforeman.org/t/changes-to-event-handling-in-katello/15963

Concern was raised about running the threads in the Passenger or Puma 'preloader' (the process from which the workers are forked) since we lose the benefits of being managed in the worker pool which is useful in cases where memory bloat occurs, or something  else goes wrong.

Now, the event daemon will run in a single worker, whether there is one or 100 of them. The tricky part was to ensure that multiple workers would not try to start the daemon simultaneously. Since this is happening cross-process, I opted to use https://ruby-doc.org/core-2.5.0/File.html#method-i-flock which can maintain an exclusive lock on a file at the OS level, so it acts as a mutex.

**The flow is like this:**

- Worker A starts serving a request and checks to see if the Daemon is already started, if so, it does nothing and continues startup
- If the daemon is not running, a lock is placed on an arbitrary file, and daemon startup begins, writing its pidfile
- If Worker B serves a request at the same time and also finds the daemon not started it waits for the file lock to be released (it's imperceptible since daemon start is fast)
- Worker B sees that the daemon is already started by Worker A, and continues as usual

This approach gives improved resiliency to the event daemon. If the daemon's worker process dies another worker will take its place in 1 minute or less depending if the fact daemon run state is cached or not.

Watch the daemon recover if its current process is killed:

```
last_daemon_pid = nil

loop do
  log_line = `grep 'event daemon started' /var/log/foreman/production.log | tail -1`
  daemon_pid = log_line.match(/process=(\d+)/)[1]

  if daemon_pid.nil? || daemon_pid == last_daemon_pid
    puts "a daemon was never started, stopping"
    puts log_line
    break
  end

  last_daemon_pid = daemon_pid
  Process.kill('INT', daemon_pid.to_i)

  puts "killed daemon process=#{daemon_pid}"

  sleep 15
end
```

### DIFF
--- a/app/services/katello/event_daemon.rb
+++ b/app/services/katello/event_daemon.rb
@@ -1,19 +1,39 @@
 module Katello
   class EventDaemon
-    PID_CACHE_KEY = 'katello_event_daemon_pid'.freeze
-
     class << self
+      def initialize
+        FileUtils.touch(settings[:lock_file])
+      end
+
       def settings
         SETTINGS[:katello][:event_daemon]
       end
 
-      def cached_pid
-        Rails.cache.fetch(PID_CACHE_KEY)
+      def pid
+        return unless pid_file && File.exist?(pid_file)
+
+        File.open(pid_file) { |f| f.read.to_i }
+      end
+
+      def pid_file
+        settings[:pid_file]
+      end
+
+      def write_pid_file
+        return unless pid_file
+
+        File.open(pid_file, 'w') { |f| f.puts Process.pid }
+
+        at_exit do
+          if pid == Process.pid
+            File.unlink(pid_file) if pid_file && File.exist?(pid_file)
+          end
+        end
       end
 
       def stop
+        return unless pid == Process.pid
         services.values.each(&:close)
-        Rails.cache.delete(PID_CACHE_KEY)
       end
 
       def start(worker: false)
@@ -26,7 +46,8 @@ module Katello
 
           start_services
 
-          Rails.cache.write(PID_CACHE_KEY, Process.pid)
+          write_pid_file
+
           Rails.logger.info("Katello event daemon started process=#{Process.pid} multiprocess=#{settings[:multiprocess]}")
         ensure
           lockfile.flock(File::LOCK_UN)
@@ -34,7 +55,7 @@ module Katello
       end
 
       def started?
-        Process.kill(0, cached_pid)
+        Process.kill(0, pid)
         true
       rescue Errno::ESRCH, TypeError # process no longer exists or we had no PID cached
         false
--- a/app/services/katello/event_daemon.rb
+++ b/app/services/katello/event_daemon.rb
@@ -3,6 +3,7 @@ module Katello
     class << self
       def initialize
         FileUtils.touch(lock_file)
+        @cache = ActiveSupport::Cache::MemoryStore.new
       end
 
       def settings
@@ -73,7 +74,10 @@ module Katello
       end
 
       def runnable?
-        !started? && settings[:enabled] && !::Foreman.in_rake? && !Rails.env.test?
+        # avoid accessing the disk on each request
+        @cache.fetch('katello_event_daemon_runnable', expires_in: 1.minute) do
+          !started? && settings[:enabled] && !::Foreman.in_rake? && !Rails.env.test?
+        end
       end
 
       def services
--- a/app/services/katello/event_daemon.rb
+++ b/app/services/katello/event_daemon.rb
@@ -23,17 +23,12 @@ module Katello
         return unless pid_file
 
         File.open(pid_file, 'w') { |f| f.puts Process.pid }
-
-        at_exit do
-          if pid == Process.pid
-            File.unlink(pid_file) if pid_file && File.exist?(pid_file)
-          end
-        end
       end
 
       def stop
         return unless pid == Process.pid
         services.values.each(&:close)
+        File.unlink(pid_file) if pid_file && File.exist?(pid_file)
       end
 
       def start(worker: false)
@@ -45,8 +40,11 @@ module Katello
           return if started? # ensure it wasn't started while we waited for the lock
 
           start_services
-
           write_pid_file
+
+          at_exit do
+            stop
+          end
 
           Rails.logger.info("Katello event daemon started process=#{Process.pid} multiprocess=#{settings[:multiprocess]}")
         ensure
@@ -63,10 +61,6 @@ module Katello
 
       def start_services
         services.values.each(&:run)
-
-        at_exit do
-          stop
-        end
       end
 
       def runnable?
--- a/app/services/katello/event_daemon.rb
+++ b/app/services/katello/event_daemon.rb
@@ -76,7 +76,7 @@ module Katello
           return false
         end
 
-        if Rails.env.production? && $PROGRAM_NAME.match(/dynflow/)
+        if $PROGRAM_NAME.match(/dynflow/)
           Rails.logger.debug("This appears to be the Dynflow process; not starting Katello Event Daemon")
           return false
         end
--- a/app/services/katello/event_daemon.rb
+++ b/app/services/katello/event_daemon.rb
@@ -64,12 +64,7 @@ module Katello
       end
 
       def runnable?
-        return false if !settings[:enabled] || ::Foreman.in_rake? || Rails.env.test?
-
-        if started?
-          Rails.logger.debug("Katello Event Daemon already running; not attempting startup")
-          return false
-        end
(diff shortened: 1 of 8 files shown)


app/services/katello/event_monitor/poller_thread.rb
```diff
@@ -81,18 +81,18 @@ def poll_for_events
...
         @thread = Thread.new do
           begin
-            Rails.application.executor.wrap do
-              @logger.info("Polling Katello Event Queue")
-              ::Katello::EventQueue.reset_in_progress
-              loop do
+            @logger.info("Polling Katello Event Queue")
+            ::Katello::EventQueue.reset_in_progress
+            loop do
+              Rails.application.executor.wrap do
```
jturel: This fixes a deadlock when reloading code within the request cycle such as when hammer requests will issue a reload of the application code. @ianballou thanks for pointing this one out. Rails' DebugLocks middleware helped me see that my executor.wrap was not in the right place! https://api.rubyonrails.org/classes/ActionDispatch/DebugLocks.html

jturel:
@ehelms @jlsherrill any thoughts on using this approach in production?

jturel:
[test katello]

ehelms:
> For those of us single process devs

What scenario is a single process dev? Is this a "non-standard" development environment?

lib/katello/engine.rb
```diff
@@ -26,7 +26,12 @@ class Engine < ::Rails::Engine
         :redhat_repository_url => 'https://cdn.redhat.com',
         :consumer_cert_rpm => 'katello-ca-consumer-latest.noarch.rpm',
         :consumer_cert_sh => 'katello-rhsm-consumer',
-        :event_daemon_enabled => true,
+        :event_daemon => {
+          enabled: true,
+          multiprocess: true,
+          lock_file: '/tmp/katello_event_daemon.lock',
+          pid_file: "#{Rails.root.join('tmp')}/pids/katello_event_daemon.pid"
+        },
```
ehelms: In what cases would I as a Foreman administrator want to change each of these settings?

ehelms:
If I look ahead to future architectures, does this model work if I:

 a) Have multiple application servers on the same host?
 b) Have multiple application servers on different hosts?
 c) Anything in Rails 6 that would break this model? (or enhance it)

app/services/katello/event_daemon.rb
```diff
@@ -1,60 +1,95 @@
...
 
-    def self.services
-      {
-        candlepin_events: ::Katello::CandlepinEventListener,
-        katello_events: ::Katello::EventMonitor::PollerThread
-      }
+        if Rails.env.production? && $PROGRAM_NAME.match(/dynflow/)
+          Rails.logger.debug("This appears to be the Dynflow process; not starting Katello Event Daemon")
+          return false
+        end
```
jlsherrill: This seems to allow it to run if the program name matches /dynflow/ as long as its not a production environment, which seems unnecessary.  (You could just remove the production? check), unless i'm missing some scenario?

jturel:
> > For those of us single process devs
> 
> What scenario is a single process dev? Is this a "non-standard" development environment?

Using `foreman start` (puma, multiple processes, clustered) vs `rails s` (puma, single process, non-clustered)

app/services/katello/event_daemon.rb
jturel: Good catch I guess dynflow could theoretically be started on its own in dev

lib/katello/engine.rb
jturel: I looked closely and in prod we symlink the rails tmp directory to /var/run/foreman. That's great! So I think I can simply hard-code these. My concern was about writable directories between dev and prod

jturel:
> If I look ahead to future architectures, does this model work if I:
> 
> a) Have multiple application servers on the same host?

This seems like an unrealistic scenario. What am I missing?

> b) Have multiple application servers on different hosts?

I don't think it would differ from dynflow in that case today. Dynflow would also be running across hosts right? Candlepin event handling should be OK since we're plucking messages from a real message queue. Not sure how the katello events side would handle being run across multiple hosts. Addressing problems there wouldn't necessarily happen at this level of the code, but rather where those events are being plucked and handled from the DB.

> c) Anything in Rails 6 that would break this model? (or enhance it)

I don't think so

app/services/katello/event_daemon.rb
jturel: Fixed

app/services/katello/event_daemon.rb
```diff
@@ -1,60 +1,99 @@
...
 
-    def self.started?
-      @pid.present?
-    end
+        if $PROGRAM_NAME.match(/spring/)
+          Rails.logger.debug("Spring is running in the environment; not starting Katello Event Daemon")
+          return false
+        end
+
+        if $PROGRAM_NAME.match(/dynflow/)
```
jturel: @adamruzicka hey, are you aware of any better way to see if this code is running in the same process as dynflow?

adamruzicka: In this context by "dynflow" you mean an executor?

jturel: I mean anywhere that dynflow is loading the rails application code (initializers being triggered, for example). Maybe that's the executor, but maybe that's only one such place?

jturel:
[test katello]

app/services/katello/event_daemon.rb
jturel: Reworked the code so I don't need to check this any more :)

app/services/katello/event_daemon.rb
```diff
@@ -1,60 +1,91 @@
...
-    end
+      def pid_dir
+        Rails.root.join('tmp', 'pids')
+      end
 
-    def self.start
-      if runnable?
-        services.values.each(&:run)
+      def lock_file
+        Rails.root.join('tmp', 'katello_event_daemon.lock')
```
adamruzicka: Not an issue for now, but if we try to stuff Foreman+Katello into containers, this path won't necessarily have to be shared between containers.

app/services/katello/event_daemon.rb
adamruzicka: I'm glad you did, I couldn't really come up with a reliable way which would work in all scenarios

jlsherrill:
>  b) Have multiple application servers on different hosts?

I think with a small amount of work, these tasks should be safe to run multiple of (meaning if one was running on every application host, that would be okay).  Maybe we should track that work for the medium term.

app/services/katello/event_daemon.rb
jturel: Good point. By the time we get there I'd like to make this all safe to be running multiple instances of

app/services/katello/event_daemon.rb
```diff
@@ -1,60 +1,91 @@
...
+        false
+      end
+
+      def start_services
+        services.values.each(&:run)
+      end
+
+      def runnable?
+        # avoid accessing the disk on each request
+        @cache.fetch('katello_event_daemon_runnable', expires_in: 1.minute) do
```
jturel: This adds a nice performance benefit over a lot of requests (100k in these results)

```
=> [#<Benchmark::Tms:0x0000000013d78828 @label="runnable_cached", @real=5.545715187967289, @cstime=0.0, @cutime=0.0, @stime=0.1517280000000003, @utime=1.3842180000000042, @total=1.5359460000000045>]


=> [#<Benchmark::Tms:0x000000000999ae68 @label="runnable_uncached", @real=79.99794052902143, @cstime=0.0, @cutime=0.0, @stime=1.329981, @utime=14.650019, @total=15.98>]```

ehelms:
Sorry, slow to reply.

>> a) Have multiple application servers on the same host?
> This seems like an unrealistic scenario. What am I missing?

With puma, we can do this by starting multiple services on different ports. For example, foreman@3001, foreman@3002 and then load balancing across them. I'm not as clear when this model vs. Puma in cluster mode is more applicable for scale / resiliency. 

Also, my questions are just for my own knowledge and to ensure we thought about them not to block review.

jturel:
> Sorry, slow to reply.
> 
> > > a) Have multiple application servers on the same host?
> > > This seems like an unrealistic scenario. What am I missing?
> 
> With puma, we can do this by starting multiple services on different ports. For example, foreman@3001, foreman@3002 and then load balancing across them. I'm not as clear when this model vs. Puma in cluster mode is more applicable for scale / resiliency.
> 
> Also, my questions are just for my own knowledge and to ensure we thought about them not to block review.

Interesting idea! This design holds up in that scenario because the lock file which prevents multiple daemons from being started uses a lock at the OS level so it works across processes.

PR #8409, opened 2019-11-01
Fixes #28188 - audit repo syncs and fallback to audit times

jlsherrill:
this adds auditing of repository syncs, upon successful
syncs.  Display of last syncs times will use audit times
if the last sync tasks have been removed.

### DIFF
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/views/product-repositories.html
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/views/product-repositories.html
@@ -88,8 +88,11 @@
           <td bst-table-cell>{{ repository.content_type }}</td>
           <td bst-table-cell>
             <span ng-show="repository.url">
-              <span ng-show="repository.last_sync == null" translate>
+              <span ng-show="repository.last_sync == null && repository.last_sync_words == null" translate>
                 Not Synced
+              </span>
+              <span ng-show="repository.last_sync == null && repository.last_sync_words" translate>
+                Completed {{ repository.last_sync_words }} ago
               </span>
               <span ng-show="repository.last_sync !== null && repository.last_sync.ended_at == null" translate>
                 <a href="/foreman_tasks/tasks/{{repository.last_sync.id}}">{{ repository.last_sync.result | capitalize}}</a>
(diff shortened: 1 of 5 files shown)


jlsherrill:
Note there are some usability issues with custom audit actions, addressed here: https://github.com/theforeman/foreman/pull/7150

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/views/product-repositories.html
```diff
@@ -88,9 +88,12 @@
           <td bst-table-cell>{{ repository.content_type }}</td>
           <td bst-table-cell>
             <span ng-show="repository.url">
-              <span ng-show="repository.last_sync == null" translate>
+              <span ng-show="repository.last_sync == null && repository.last_sync_words == null" translate>
                 Not Synced
               </span>
+              <span ng-show="repository.last_sync == null && repository.last_sync_words" translate>
+                Completed {{ repository.last_sync_words }} ago
+              </span>
```
johnpmitsch: @jlsherrill what do you think about putting the logic in these two spans in the controller?

johnpmitsch:
@jlsherrill works well! I synced on repository on master, and one with this PR, the syncs done with the PR created audits. I deleted the tasks for both syncs, and the one made with this PR still showed a last sync, while the one done with master did not (as expected):
witha

Just one question around moving the UI logic in the view to the control

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/views/product-repositories.html
johnpmitsch: I would have to look at it more, I only suggested it as it seems like a lot of logic is going on in this view and there is some room for improvement. Maybe it would be too complicated to do for that cell since there are various structures for each option

jlsherrill: thanks yeah, i initially agreed, but when i went to make the change it seemed more awkward as to not being worth it.

johnpmitsch: seems fine as is then, I approved :)

PR #8413, opened 2019-11-04
[CP] For Katello 3.13.2

jjeffers:


app/models/katello/rhsm_fact_parser.rb
```diff
@@ -53,8 +53,14 @@ def operatingsystem
 
       os_name = ::Katello::Candlepin::Consumer.distribution_to_puppet_os(name)
       major, minor = version.split('.')
-      if os_name && !invalid_centos_os?(os_name, minor)
-        os_attributes = {:major => major, :minor => minor || '', :name => os_name, :release_name => os_release_name(os_name)}
+      unless facts['ignore_os']
```
jturel: Did you hit a merge conflict in this area? I think there's more code related to the ignore_os fact which is missing which is the reason for me asking. We may want to just pick that PR into this release to avoid the merge conflict altogether

PR #8415, opened 2019-11-04
Fixes #28189 - Refresh repo when needed

parthaa:
When a manifest refresh is triggered all the RH repos update their
distributors and importers, even if those repos did not change. This
commit makes a repo update its importers and distributors only if
needed.

### DIFF
--- a/app/services/katello/pulp/repository.rb
+++ b/app/services/katello/pulp/repository.rb
@@ -11,8 +11,10 @@ module Katello
       end
 
       def backend_data(force = false)
-        return smart_proxy.pulp_api.extensions.repository.retrieve_with_details(repo.pulp_id) if (repo.pulp_id && force)
-        @backend_data ||= smart_proxy.pulp_api.extensions.repository.retrieve_with_details(repo.pulp_id) if repo.pulp_id
+        if (repo.pulp_id && (force || @backend_data.nil?))
+          @backend_data = smart_proxy.pulp_api.extensions.repository.retrieve_with_details(repo.pulp_id)
+        end
+        @backend_data
       rescue RestClient::ResourceNotFound
         nil
       end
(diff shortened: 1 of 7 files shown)


parthaa:
To test this
In master
- Import a manifest
- Enable a couple of RH repos
- Refresh manifest
- Goto the dynflow task for the RefreshManifest and check for  ```Actions::Pulp::Repository::Refresh (success) ```
 - Make sure you see a bunch of pulp tasks in the output for the ```Pulp::Repository::Refresh```

Now check out the PR
- Refresh manifest
- Goto the dynflow task for the RefreshManifest and check for  ```Actions::Pulp::Repository::Refresh (success) ```
 - Make sure you see ```pulp_tasks:[]```  in the output for the ```Pulp::Repository::Refresh```

parthaa:
[test katello]

jjeffers:
@parthaa is there a test case where the `refresh_if_needed` is triggered? I think I see that the test change shows the condition where it is not triggered. Correct me if I'm reading that incorrectly.

parthaa:
> @parthaa is there a test case where the `refresh_if_needed` is triggered? I think I see that the test change shows the condition where it is not triggered. Correct me if I'm reading that incorrectly.

https://github.com/Katello/katello/pull/8415/files#diff-9a8be58a012da76042ea5e21ad94e40aR160

I test if refresh_if_needed does not call refresh

app/services/katello/pulp/repository.rb
```diff
@@ -199,6 +201,10 @@ def proxy_host_importer_value
         root.ignore_global_proxy ? "" : nil
       end
 
+      def refresh_if_needed
+        refresh if needs_importer_updates? || needs_distributor_updates?
+      end
```
jlsherrill: this is the same logic as mirror_needs_updates?  maybe we could combine those into one method?

parthaa: updated

parthaa:
[test katello]

parthaa:
[test katello]

parthaa:
[test katello]

PR #8416, opened 2019-11-05
Fixes #28206 - consistent pulp3 sync status

jlsherrill:


jlsherrill:
[test katello]

app/lib/actions/pulp3/repository/presenters/content_unit_presenter.rb
```diff
@@ -15,6 +15,7 @@ def humanized_details
             ret << _("Total tasks: ") + ": #{finished_units}/#{total_units}"
             ret << "--------------------------------"
             progress_reports = sync_task.try(:[], 'progress_reports') || []
+            progress_reports = progress_reports.sort_by { |pr| pr.try(:[], 'message') }
```
sjha4: Looks good..Even though I am not sure that alphabetically is the natural order for these messages. Maybe in the order they execute but pulp needs to give us that.

sjha4:
[test katello]

app/lib/actions/pulp3/repository/presenters/content_unit_presenter.rb
jlsherrill: Yeah, the problem was that the ordering wasn't consistent coming from pulp itself.  But this could be considered a pulp bug, i will chat with them before merging

jlsherrill:
I did have a discussion with them and they didn't seem to have any strong opinions either way, but were open to changing it in the future.  The current ordering isn't really set to any particular order.  We can revisit this if we ever want to. Thanks!

PR #8421, opened 2019-11-07
Fixes #28219 - setting 201 as response code for POST

sseelam2:


jturel:
[test katello]

chris1984:
Starting to review

jturel:
The change requests that the response code returns 201 instead of 200 so it looks like this might not be completely working - definitely needs a test case!

chris1984:
> The change requests that the response code returns 201 instead of 200 so it looks like this might not be completely working - definitely needs a test case!

Good catch, I read it wrong. Will remove my review

jturel:
@sseelam2 check the redmine issue linked on the BZ. The PR associated with it references a method called `respond_for_create` maybe we can use that? This solution makes me thing we'll respond with a 201 anywhere that we don't set a response code ourselves. That doesn't seem correct: 

```
The HTTP 201 Created success status response code indicates that the request has succeeded and has led to the creation of a resource. The new resource is effectively created before this response is sent back and the new resource is returned in the body of the message, its location being either the URL of the request, or the content of the Location header.

The common use case of this status code is as the result of a POST request.
```

Let me know if I have misinterpreted your change

sseelam2:
> The change requests that the response code returns 201 instead of 200 so it looks like this might not be completely working - definitely needs a test case!

@jturel Only for a POST request right? So maybe I could set the response to 201 inside respond_for_create if it gets successfully created.

jturel:
> > The change requests that the response code returns 201 instead of 200 so it looks like this might not be completely working - definitely needs a test case!
> 
> @jturel Only for a POST request right? So maybe I could set the response to 201 inside respond_for_create if it gets successfully created.

Correct, since we use POST for creating resources in the REST world. Try it and see. I recommend starting with a failing test case before making any change :)

sseelam2:
[test katello]

test/controllers/api/v2/host_collections_controller_test.rb
```diff
@@ -61,7 +61,7 @@ def test_show
     def test_create
       post :create, params: { :organization_id => @organization, :host_collection => {:name => 'Collection A', :description => 'Collection A, World Cup 2014'} }
 
-      assert_response :success
+      assert_response (201)
```
jturel: Let's be consistent with our style. I'd like to see `assert_response 201` or `assert_response(201)` (note no space when using parenthesis)

jturel:
I think the commit & PR need a more accurate name since the change isn't actually to set 201 response code for POST requests, but for create requests (they just happen to be HTTP POST).

test/controllers/api/v2/host_collections_controller_test.rb
sseelam2: rubocop complained about it as well.

jturel:
LGTM!

@chris1984 wanna give it another test?

chris1984:
Sounds good will test again and report back

PR #8424, opened 2019-11-08
Fixes #28102 - Add Pulp3 to reset rake task

chris1984:
Cleaned up some stuff that was left around from EL6 days as well

### DIFF
--- a/lib/katello/tasks/reset.rake
+++ b/lib/katello/tasks/reset.rake
@@ -1,29 +1,50 @@
 namespace :katello do
   namespace :reset_backends do
-    service_stop = "sudo /sbin/service %s status > /dev/null && sudo /sbin/service %s stop"
-    service_start = "sudo /sbin/service %s start"
+    # Switch to foreman-maintain when Pulp3 services are added to it
+    service_stop = "sudo /usr/bin/systemctl status %s > /dev/null && sudo /usr/bin/systemctl stop %s"
+    service_start = "sudo /usr/bin/systemctl start %s"
 
-    task :pulp do
-      SERVICES = %w(httpd pulp_workers pulp_resource_manager pulp_celerybeat).freeze
-      system(service_stop.gsub("%s", "mongod"))
+    task :pulp_legacy do
+      SERVICES = %w(httpd pulp_workers pulp_resource_manager pulp_celerybeat squid qpidd pulp_streamer).freeze
+
+      puts "\e[33mStarting Pulp2 Reset\e[0m\n\n"
 
       SERVICES.each { |s| system(service_stop.gsub("%s", s)) }
-      system("sudo rm -rf /var/lib/mongodb/pulp_database*")
-      system(service_start.gsub("%s", "mongod"))
-      sleep(10)
-      fail "Cannot migrate pulp database" unless system("sudo -u apache /usr/bin/pulp-manage-db")
+      system(service_start.gsub("%s", 'rh-mongodb34-mongod'))
+      system("mongo pulp_database --eval 'db.dropDatabase();'")
+      fail "\e[31mCannot Migrate Pulp2 Database\e[0m\n\n" unless system('sudo -u apache /usr/bin/pulp-manage-db')
 
       SERVICES.each { |s| system(service_start.gsub("%s", s)) }
-      puts "Pulp database reset."
+      puts "\e[32mPulp2 Database Reset Complete\e[0m\n\n"
+    end
+
+    task :pulp do
+      SERVICES = %w(redis pulpcore-api pulpcore-resource-manager pulpcore-content).freeze
+
+      puts "\e[33mStarting Pulp3 Reset\e[0m\n\n"
+
+      SERVICES.each { |s| system(service_stop.gsub("%s", s)) }
+      system("sudo systemctl stop 'pulpcore-worker@*' --all")
+      system("sudo runuser - postgres -c 'dropdb pulp -p 7878'")
+      system("sudo runuser - postgres -c 'createdb pulp -p 7878'")
+      Dir.chdir('/usr/local/lib/pulp/bin') do
+        fail "\e[31mCannot migrate Pulp3 database\e[0m\n\n" unless system("sudo -u pulp PULP_SETTINGS='/etc/pulp/settings.py' DJANGO_SETTINGS_MODULE='pulpcore.app.settings' ./django-admin migrate --no-input")
+        puts "\e[33mRecreating Admin User\e[0m\n\n"
+        system("sudo -u pulp PULP_SETTINGS='/etc/pulp/settings.py' DJANGO_SETTINGS_MODULE='pulpcore.app.settings' ./django-admin reset-admin-password --password password")
+      end
+
+      SERVICES.each { |s| system(service_start.gsub("%s", s)) }
+      system("sudo systemctl start 'pulpcore-worker@*' --all")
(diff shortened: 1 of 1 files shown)


lib/katello/tasks/reset.rake
```diff
@@ -1,29 +1,48 @@
 namespace :katello do
   namespace :reset_backends do
-    service_stop = "sudo /sbin/service %s status > /dev/null && sudo /sbin/service %s stop"
-    service_start = "sudo /sbin/service %s start"
+    service_stop = "sudo /usr/bin/systemctl status %s > /dev/null && sudo /usr/bin/systemctl stop %s"
+    service_start = "sudo /usr/bin/systemctl start %s"
```
jturel: Any reason to not use foreman-maintain here?

chris1984: Right now foreman-maintain does not know about the pulp3 services when I was seeing what restarts on the dev box. Once it does I am happy to change it over to that.

jturel: Ah! makes sense

lib/katello/tasks/reset.rake
```diff
@@ -1,29 +1,46 @@
...
+      system("runuser - postgres -c 'dropdb pulp -p 7878'")
+      fail 'Cannot migrate Pulp database' unless system("#{PULP_DIR}/django-admin migrate auth --no-input")
+      fail 'Cannot migrate Pulp database' unless system("#{PULP_DIR}/django-admin migrate --no-input")
+
+      services.each { |s| system(service_start.gsub("%s", s)) }
+      puts 'Pulp3 database reset.'
+    end
+
+    task :candlepin do
       system(service_stop.gsub("%s", tomcat))
```
ianballou: I think you're missing 
```ruby
tomcat = File.exist?('/var/lib/tomcat') ? 'tomcat' : 'tomcat6'
```

I hit this: 
```
NameError: undefined local variable or method `tomcat' for main:Object
/home/vagrant/katello/lib/katello/tasks/reset.rake:40:in `block (3 levels) in <top (required)>'
...
```

ianballou:
After applying the line I suggested above I hit this error with Pulp 2 DB: https://paste.fedoraproject.org/paste/1RiRZqeuxNTI0uH56dv7Tw

lib/katello/tasks/reset.rake
chris1984: Fixed, forgot to quote as string since I removed the variable

chris1984:
> After applying the line I suggested above I hit this error with Pulp 2 DB: https://paste.fedoraproject.org/paste/1RiRZqeuxNTI0uH56dv7Tw

Starting to look at 2nd failure

chris1984:
@ianballou can you try with the updated pr and let me know if it works. It looks like that was coming from Pulp3 so I updated the fail message to alert better which version is failing. It looks like it could not find the settings.py so I set that manually here: https://github.com/Katello/katello/pull/8424/files#diff-0f71407350d301b3670d977a704b4e49R33

ianballou:
@chris1984 the system call to set `DJANGO_SETTINGS_MODULE` wasn't working so I replaced it with ```ruby
ENV['DJANGO_SETTINGS_MODULE'] = '/etc/pulp/settings.py
```

That caused the following error: https://paste.fedoraproject.org/paste/dsYctAF9ZIH1F5vSXvMM8Q

**Comment by ianballou** [comment]
@chris1984 I think we need to set `DJANGO_SETTINGS_MODULE` to `pulp.app.settings`. Found this https://pulp.plan.io/issues/2819

**Comment by chris1984** [comment]
@ianballou updated and fully working, and now with colors and better descriptions of what is going on :)

**Comment by sjha4** [comment]
The reset worked great. I could see pulp 2/3 dbs getting dropped and created and migrated.
A useful addition here would be to create the admin user for the pulp3 db on the lines of https://github.com/pulp/ansible-pulp/blob/0f0c0844a771358fe1590de110d13128af5dbb9e/roles/pulp-database/tasks/main.yml#L37 to allow connecting to the db after reset.

**Comment by chris1984** [comment]
@sjha4 added, snippet showing it is working:

```shell
Successfully set password for "admin" user.
Pulp3 Database Reset Complete
```

chris1984:
@ianballou Samir acked the pr with recreating the admin user, are you fine with acking as well or did you want to wait till the new pulp3 works.

lib/katello/tasks/reset.rake
```diff
@@ -1,29 +1,57 @@
...
+
+    task :pulp do
+      # Pull active workers from Redis to see how many we need to start up
+      worker_count = `redis-cli --scan --pattern "rq:workers:reserved-resource-worker*" | wc -l`.chomp.to_i
+
+      services = %w(redis pulpcore-api pulpcore-resource-manager pulpcore-content)
+      while worker_count > 0
+        services.push("pulpcore-worker@#{worker_count}")
+        worker_count -= worker_count
+      end
```
jlsherrill: using this: 
```
systemctl start 'pulpcore-worker@*' --all
```

i think is much simpler and works even if the workers are down

lib/katello/tasks/reset.rake
```diff
@@ -1,29 +1,57 @@
...
+      services.each { |s| system(service_stop.gsub("%s", s)) }
+      system("sudo runuser - postgres -c 'dropdb pulp -p 7878'")
+      system("sudo runuser - postgres -c 'createdb pulp -p 7878'")
+      # rubocop:disable Style/BlockDelimiters
+      Dir.chdir('/usr/local/lib/pulp/bin') {
+        fail "\e[31mCannot migrate Pulp3 database\e[0m\n\n" unless system("sudo -u pulp PULP_SETTINGS='/etc/pulp/settings.py' DJANGO_SETTINGS_MODULE='pulpcore.app.settings' ./django-admin migrate --no-input")
+        puts "\e[33mRecreating Admin User\e[0m\n\n"
+        system("sudo -u pulp PULP_SETTINGS='/etc/pulp/settings.py' DJANGO_SETTINGS_MODULE='pulpcore.app.settings' ./django-admin reset-admin-password --password password")
+      }
+      # rubocop:enable Style/BlockDelimiters
```
jlsherrill: is there a reason you didn't use block delimiters here? like:
```
Dir.chdir(path) do 
  Code
  Code 
end
```

chris1984: I can write it like that, didnt know it would work that way without the braces, all the examples on rubocop I saw were .each and .item. I will fix that too

jlsherrill: yep, all blocks can be written either as:
```
method {   
}
```

or 
```
method do 
end
```

These two ways of doing it are exactly the same.

we don't allow {} for multi-line blocks as a style rule, meaning its only allowed when its written in one line:
```
method {    }
```

chris1984:
@jlsherrill thanks for the help, updated it. @ianballou updated based on Justin's comments

lib/katello/tasks/reset.rake
```diff
@@ -33,10 +54,15 @@ namespace :katello do
     SmartProxy.where(name: hostname, url: "https://#{hostname}:9090").first_or_create!
   end
 
-  desc "Resets Foreman/Katello development environemnt. WARNING: This will destroy all your Foreman and Katello data."
+  desc 'Resets the Foreman/Katello development environemnt. WARNING: This will destroy all your Foreman, Katello and Pulp data.'
```
ianballou: `%s/environemnt/environment`

PR #8430, opened 2019-11-12
Fixes #28241 - upgrade vendor to v3

sharvit:
The main upgrades are `redux-hooks`. `babel` v7 and `jest` v24


sharvit:
Updated to use `@theforeman/babel`

webpack/scenes/RedHatRepositories/components/RepositorySet.js
```diff
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import { ListView, Icon } from 'patternfly-react';
 
 import RepositoryTypeIcon from './RepositoryTypeIcon';
+// eslint-disable-next-line import/no-named-as-default
 import RepositorySetRepositories from './RepositorySetRepositories';
```
jeremylenz: Why disable the rule instead of importing the named import? (`import { RepositorySetRepositories }` instead of `import RepositorySetRepositories`)  And why is this needed for upgrading the vendor package?

sharvit: Because the named import is not connected to redux.
I wanted to make it work without a refactor to the component, eventually, it should migrate to the new folder structure so you will have one export for each file.

jeremylenz: ah ok, makes sense

ekohl:
This is needed for nightly to work again. Could a Katello dev review this?

waldenraines:
This is causing https://projects.theforeman.org/issues/28423.  When I downgrade vendor to 2.x the subscriptions page loads without error.

jeremylenz:
That means one of the dependencies of vendor v3 must be causing it

sharvit:
@waldenraines I assume it's the `react-router-dom` update we introduced in version `3.4.0`, see:
- https://github.com/theforeman/foreman-js/blob/master/packages/vendor-core/CHANGELOG.md#340-2019-12-04
- https://github.com/theforeman/foreman-js/commit/f792279

But it is odd, I remember myself checking the subscription page works before the release.
Can you try using vendor version `v3.3.2` and verify it apear with `v3.4.0`?

@laviro can you give some insights? looks like `katello` need to update something.

laviro:
https://github.com/react-bootstrap/react-router-bootstrap/issues/256
seem you were right @sharvit, someone already opened an issue 2 days ago about the compatibility with `react-router-dom v5.x`

sharvit:
Should we revert react-router-dom back to version 4?

laviro:
upgrading `react-router-bootstrap to "^0.25.0"` seem to fix it:

screencapture-centos7-luna-devel-rlavi-example-com-subscriptions-1576141511566

laviro:
upgrade here: https://github.com/theforeman/foreman-js/pull/95

PR #8433, opened 2019-11-13
Fixes #27920 - refactor errata modeling to support pulp3

sjha4:
My steps to reproduce and test:
1. Download https://github.com/Katello/runcible/tree/master/test/fixtures/repositories/zoo5 and start HTTP server.
2. Sync this repo with katello running Pulp3
3. Sync repo https://fedorapeople.org/groups/katello/fakerepos/zoo2/
4. Notice that duplicate errata are created.

With this change, you should see no duplicate errata. Also if you update the local repo to add packages etc to the errata, you should see those getting added to the combined errata.


app/models/katello/erratum.rb
```diff
@@ -144,6 +144,30 @@ def module_stream_objects
       return streams.flatten.uniq
     end
 
+    def self.import_for_repository(repository)
+      pulp_ids = []
+      service_class = SmartProxy.pulp_master!.content_service(content_type)
+      fetch_only_ids = !repository.content_view.default? &&
+          !repository.repository_type.unique_content_per_repo
```
jlsherrill: since this is errata, i think we can assume 'repository.repository_type.unique_content_per_repo' is false, correct?

app/models/katello/erratum.rb
```diff
@@ -144,6 +144,30 @@ def module_stream_objects
...
+            end
+            service = service_class.new(model.errata_id)
+            service.backend_data = unit
+            model.repository_id = repository.id unless many_repository_associations
+            service.update_model(model)
+          end
+          pulp_ids << errata_id
+        end
+      end
+      sync_repository_associations(repository, :errata_ids => pulp_ids) if self.many_repository_associations
```
jlsherrill: shouldn't that be :pulp_ids => pulp_ids? ?

jlsherrill: or else, shouldn't sync_repository_associations be enhanced to handle :errata_ids ?

jlsherrill: i think it would make more sense to just pull the pulp_id from the errata (looked up via the errata_id) and use sync_repository_associations as is?

jlsherrill: In which case you might be able to distill the differences into just a method that does something like, unless i'm missing other differences:
```
               self.where(:errata_id => errata_id).first_or_create(:pulp_id => errata_id)
```

sjha4: hmm..looks like this PR is missing a commit..I am trying to add errata_id support to sync_repository_associations.

sjha4: Actually..I think the changes to the sync_repository_associations are in the pulp_database_unit file above..

sjha4: > In which case you might be able to distill the differences into just a method that does something like, unless i'm missing other differences:
> 
> ```
>                self.where(:errata_id => errata_id).first_or_create(:pulp_id => errata_id)
> ```

I'll test this out..

jlsherrill: oh hah, sorry i missed it!  but yes, i'd like to see if there's a way to not have to duplicate this method completely

sjha4: I think as long as the pulp2_3 migrated errata has pulp_id as errata_id instead of pulp_href, setting identifier to id for errata should work fine thanks to this: https://github.com/Katello/katello/blob/b5e7d05c134a864e8f1971b803fd2fadc8648bb0/app/models/katello/concerns/pulp_database_unit.rb#L112

Added a separate commit to try that.

sjha4:
[test katello]

jturel:
There's a lot of new fixtures in here especially relating to RPM. I'm not against it, but I am curious if it would be better suited as one of our BATS tests?

app/models/katello/concerns/pulp_database_unit.rb
```diff
@@ -127,7 +127,7 @@ def import_for_repository(repository)
 
       def sync_repository_associations(repository, options = {})
         additive = options.fetch(:additive, false)
-        pulp_ids = options.fetch(:pulp_ids)
+        pulp_ids = options.fetch(:pulp_ids, nil)
```
jturel: Not asking for change here, but I like to use `.dig` in this case since it will return nil instead of raising an error

jturel:
Here were my testing steps:

- Synced both repos on the same product and I saw duplicated errata in Content -> Errata
- Removed the zoo2 repo, and saw the duplicates were removed
- Checked out this PR
- Created zoo2 again and synced it, and I still see duplicates:

dup_errata

sjha4:
@jturel: So with this PR, we are trying to set pulp_id as errata_id. If you synced one repo before the change, the pulp_id would still be the pulp_href and then with this change the second errata would have pulp_id set to errata_id which are different. 
Could you delete both repos, run "curl -X "DELETE" http://admin:password@localhost:24817/pulp/api/v3/orphans/" and sync again?

jturel:
Got it.

I have three errata that look like this:

```
=> [#<Katello::Erratum:0x000000000f14e088
  id: 8,
  pulp_id: "RHEA-2012:0059",
  errata_id: "RHEA-2012:0059",
  created_at: Wed, 29 Jan 2020 21:15:59 UTC +00:00,
  updated_at: Wed, 29 Jan 2020 21:15:59 UTC +00:00,
  issued: nil,
  updated: nil,
  errata_type: "enhancement",
  severity: "",
  title: "Duck_Kangaroo_Erratum",
  solution: "",
  description: "Duck_Kangaro_Erratum description",
  summary: "",
  reboot_suggested: false,
  migrated_pulp3_href: nil>,
 #<Katello::Erratum:0x000000000f14def8
  id: 9,
  pulp_id: "RHEA-2010:0001",
  errata_id: "RHEA-2010:0001",
  created_at: Wed, 29 Jan 2020 21:15:59 UTC +00:00,
  updated_at: Wed, 29 Jan 2020 21:15:59 UTC +00:00,
  issued: nil,
  updated: nil,
  errata_type: "security",
  severity: "",
  title: "Empty errata",
  solution: "",
  description: "Empty errata",
  summary: "",
  reboot_suggested: false,
  migrated_pulp3_href: nil>,
 #<Katello::Erratum:0x000000000f14dd90
  id: 10,
  pulp_id: "RHEA-2010:0002",
  errata_id: "RHEA-2010:0002",
  created_at: Wed, 29 Jan 2020 21:15:59 UTC +00:00,
  updated_at: Wed, 29 Jan 2020 21:15:59 UTC +00:00,
  issued: nil,
  updated: nil,
  errata_type: "security",
  severity: "",
  title: "One package errata",
  solution: "",
  description: "One package errata",
  summary: "",
  reboot_suggested: false,
  migrated_pulp3_href: nil>]
```

sjha4:
Merging...Thanks @jturel , @jlsherrill !

PR #8434, opened 2019-11-14
Fixes #28238: pass result of await into reload.

waldenraines:
Pass the result of async-await into the reload
handler so that it works correctly.

https://projects.theforeman.org/issues/28238


waldenraines:
[test]

waldenraines:
[test katello]

waldenraines:
> One preexisting thing we should fix at some point: When you click the + button to enable a repo, there's no spinner or other UI to let you know the request is pending. This may cause users to click multiple times thinking nothing is happening.

You know, I saw that on some repos and I think it was related to enabling, disabling, and then enabling again.  I definitely saw the spinner on a repo while enabling it so I'm not 100% sure they cause.

PR #8435, opened 2019-11-15
fixes #28211 - Added last checkin macro

domitea:


jlsherrill:
@domitea this looks good!  I would use Refs instead of 'fixes' to link it to that issue assuming more work needs to be done to issue 28211.

ares:
there's no more work that needs to happen in there I guess, should we move the ticket to katello project? This is probably only about making this available in safe mode.

PR #8437, opened 2019-11-18
Fixes #28194 - remove rpm content

jjeffers:
This PR adds a test for RPM content removal. Based on previous Pulp3/yum work already merged, no implementation changes were made.

To verify that the content removal works, create a yum repository and sync with a pulp3 enabled smart proxy. Observe the content is added in the pulp3 repository. Select a content unit (rpm) and then remove it. Check that Katello recognizes that the content is no longer associated with the repository.


jturel:
I see this is marked as needing review on our internal board but it has an ACK - please update if warranted :)

PR #8441, opened 2019-11-19
Fixes #28292 - update to rubocop 0.75

jlsherrill:
and get rid of all warnings


jlsherrill:
This might depend on https://github.com/theforeman/foreman/pull/7161  but if its :green_heart:  we can go ahead and merge

PR #8443, opened 2019-11-19
Fixes #27366 - account for missing content type in pulp3 check

jjeffers:
This change prevents errors in removing content view versions where puppet components are not present in a production foreman install.

Previously, a non-persisted repository with content type "puppet" would be supplied to a pulp3 destroy action. This content type does not exist in the RepositoryTypeManager dictionary of content types. This would throw an exception during the task execution, and the overall CVV removal task would never complete.


jjeffers:
[test katello]

app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -140,6 +140,10 @@ def pulp3_enabled?
 
       def pulp3_support?(repository)
         type = Katello::RepositoryTypeManager.repository_types[repository.content_type]
+        if type.nil?
```
jturel: Sorry, I have nothing else to offer but to ask: isn't `return unless type` more idiomatic?

jjeffers: An easy change. I'll make it so.

test/models/concerns/smart_proxy_extensions_test.rb
```diff
@@ -60,4 +60,22 @@ def test_save_with_organization_location
       assert_equal @proxy_mirror.lifecycle_environments.all.count, 0
     end
   end
+
+  class SmartProxyPulp3Test < ActiveSupport::TestCase
+    def setup
+      @master = FactoryBot.create(:smart_proxy, :default_smart_proxy, :with_pulp3)
```
jturel: It's probably the right time to create a fixture for a pulp3 proxy (maybe there already is one!)

jlsherrill: Its tricky because you can only have 1 'default' smart proxy or things get really wonky, so if we have a default one in fixtures then we'd have to delete it or modify it anytime we wanted to test pulp2.  

We'd also have to figure out to add onto the existing foreman fixtures since they are defined there (and fixtures don't have a way to do that easily like factories).

jturel: Oh good points. Probably more trouble than it's worth right now :)

jturel:
Anything that gets cherry-picked here to fix the builds should be re-triaged for 3.13 :)

jlsherrill:
oh i didn't realize this had been opened against 3.13, it needs to be merged into master first.  Everything goes into master first :)

jturel:
> oh i didn't realize this had been opened against 3.13, it needs to be merged into master first. Everything goes into master first :)

oops, thought it was a backport !

jjeffers:
[test katello]

jjeffers:
> > oh i didn't realize this had been opened against 3.13, it needs to be merged into master first. Everything goes into master first :)
> 
> oops, thought it was a backport !

I didn't realize it hadn't been merged into master as of yet.

PR #8444, opened 2019-11-19
Fixes #24517: reduce virt required dash emphasis

waldenraines:
Make the dash have less emphasis in the requires virt
who column.

https://projects.theforeman.org/issues/24517

Before:
gscreenshot_2019-11-19-121653

After:

gscreenshot_2019-11-19-121554

### DIFF
--- a/webpack/scenes/Subscriptions/SubscriptionsPage.scss
+++ b/webpack/scenes/Subscriptions/SubscriptionsPage.scss
@@ -42,3 +42,9 @@
   margin-top: 3px;
   width: 10px;
 }
+
+span.fa.fa-minus::before {
+  content: "\2014";
+  color: gray;
+  font-size: 1em;
+}


pgagne:
Looks good.

jeremylenz:
Don't hate me...

Using the hyphen was fine when the `font-size` was the same, but using the smaller font size makes it only about half the width of the check mark.  (I see why you didn't use `font-weight` instead; we apparently don't have the lighter font available.)

We could use an em-dash(`–`) instead.  This combo looks good to me:
```css
content: "–";
font-size: 1.5em;
color: gray;
```
virtwho_lite

waldenraines:
> We could use an em-dash(`–`) instead.

Usually I wouldn't want to change a patternfly icon but I think it's fine in this case as it's just a dash.  I'll make that change.

waldenraines:
> Usually I wouldn't want to change a patternfly icon but I think it's fine in this case as it's just a dash. I'll make that change.

Actually I end up with this by using your proposed changes:

gscreenshot_2019-11-20-143122

Which actually makes the emphasis more noticeable.

waldenraines:
@jeremylenz see what you think about this.

Screenshot:

gscreenshot_2019-11-20-151109

jeremylenz:
ACK :+1:

PR #8445, opened 2019-11-19
Fixes #28310 - Restore content access mode banner

jturel:
This was accidentally removed during removal of the legacy subs pages. Ping me for a golden ticket manifest so this can be tested.

The banner is expected to be shown on content host and activation key details pages.


sseelam2:
@jturel Looks good to me. I can see the banner in hosts and the activation key pages. :+1:

PR #8446, opened 2019-11-20
Fixes #28317 - Host update out of dynflow

parthaa:


parthaa:
[test prprocessor]

jturel:
@partha can you add a few comments here about any considerations you had to take / workarounds / things to be aware of as a reviewer? Just want to make sure I don't miss something :)

chris1984:
[test prprocessor]

parthaa:
So to test this

1. Register a host
1. ```subscription-manager facts --update```. This should trigger host update actions. May be you can change some fact like dmi.uuid mutiple times and run the same call and verify that the host facts were updated.
1. Updating the content view/release version  for host as a host bulk action should also trigger this,

To Review this PR codewise I 'd start with -> changes in   ```katello/concerns/subscription_facet_host_extensions.rb ``` . The rest of the changes all came out due to this.

Main change was just removal of the foreman triggering. Instead we are just doing that aspect in the before save.

I also moved the host update unit tests to HostManagedExtensions.

parthaa:
What I was worried about was performance with 100s of simultaneous host update requests. Going  to check with @mccun934 on that,
That being said one of the better features in my change I think is that Candlepin failures will fail the transaction in Katello. So we would not have a dirty data issue where katello thinks everything is good while candlepin failed.

mccun934:
@parthaa will subscription-manager block waiting for the completion of the API request so if I client calls:

```
subscription-manager facts --update
subscription-manager facts --update
```
the second call will not start until the 1st one finishes?

mccun934:
looks like it blocks, which IMO is a good thing:

```
2019-11-22T13:23:38 [I|app|22681423] Started PUT "/rhsm/consumers/4c0a7d22-36fb-40ab-bbf7-fa7395d1ffae" for 10.8.101.9 at 2019-11-22 13:23:38 -0500
2019-11-22T13:23:38 [I|app|22681423] Processing by Katello::Api::Rhsm::CandlepinProxiesController#facts as JSON
2019-11-22T13:23:38 [I|app|22681423]   Parameters: {"facts"=>"[FILTERED]", "id"=>"4c0a7d22-36fb-40ab-bbf7-fa7395d1ffae"}
2019-11-22T13:23:38 [I|app|22681423] Current user set to 4c0a7d22-36fb-40ab-bbf7-fa7395d1ffae (regular)
2019-11-22T13:23:39 [I|app|22681423] Import facts for 'aaa-katest-dynxgzhtiuuwhikj.lab.redhat.com' completed. Added: 0, Updated: 2, Deleted 0 facts
2019-11-22T13:23:39 [I|app|22681423] Completed 200 OK in 249ms (Views: 0.1ms | ActiveRecord: 64.2ms)
```

mccun934:
Awesome throughput performance with this patch:

- Default foreman-tasks configuration to complete 50,000 Host::Update tasks: **~8hours**
- Tuned foreman-tasks foreman-tasks configuration with multiple executors + threads to complete 50,000 Host::Update tasks: **~2hours**
- This PR's Host::Update patch to complete 50,000 Host::Updates : **44minutes**

over **10x** the default configuration's throughput performance.

app/models/setting/content.rb
```diff
@@ -114,8 +114,6 @@ def self.default_settings
                N_('Default Location where new Puppet content will be put upon Content View publish'),
                nil, N_('Default Location Puppet content'), nil,
                :collection => proc { Hash[Location.unscoped.all.map { |loc| [loc[:title], loc[:title]] }] }),
-      self.set('host_update_lock', N_("Allow multiple concurrent Actions::Katello::Host::Update calls for one host to be processed at the same time."),
```
tbrisker: This probably also needs a migration to clean up the existing setting from the db

tbrisker: https://github.com/Katello/katello/pull/8502

PR #8449, opened 2019-11-21
Fixes #28321 - Remove link from sub name in collapsible row

sseelam2:
Removing invalid hyperlink from subscription name in the collapsible row.


webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionNameFormatter.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
+import { urlBuilder } from 'foremanReact/common/urlHelpers';
+
+export const subscriptionNameFormatter = (value, { rowData }) => {
+  let cellContent;
+
+  if (rowData.collapsible) {
+    cellContent = (
+      <span>
```
waldenraines: Nitpick: why are these `<span>`s necessary?

jturel: Good point - the tests aren't passing because of the inclusion of the tag

sseelam2: @waldenraines They are not actually. Thanks.

sseelam2:
> Tested and it works nicely. Is it feasible to add a test?

@jturel Sure, I'll look into it.

waldenraines:
> @jturel Sure, I'll look into it.

You'd probably be able to make use of the `groupedSubscriptions` fixture in https://github.com/Katello/katello/blob/master/webpack/scenes/Subscriptions/components/SubscriptionsTable/__tests__/SubscriptionsTable.fixtures.js#L3-L59

Let me know if you have any questions.

sseelam2:
> > @jturel Sure, I'll look into it.
> 
> You'd probably be able to make use of the `groupedSubscriptions` fixture in https://github.com/Katello/katello/blob/master/webpack/scenes/Subscriptions/components/SubscriptionsTable/__tests__/SubscriptionsTable.fixtures.js#L3-L59
> 
> Let me know if you have any questions.

@waldenraines Thank you . I'm looking into snapshot testing as well. Do you think I'm going in the right direction?

waldenraines:
> @waldenraines Thank you . I'm looking into snapshot testing as well. Do you think I'm going in the right direction?

Yeah, the test that uses the snapshot (https://github.com/Katello/katello/blob/master/webpack/scenes/Subscriptions/components/SubscriptionsTable/__tests__/SubscriptionsTable.test.js#L7) makes use of the fixtures I mentioned above.  You'd just want to render the page with the grouped subscriptions and use that to snapshot the results.

sseelam2:
@waldenraines @jturel I added a test case. Let me know if it looks good!

PR #8451, opened 2019-11-21
Fixes #27366 - handles missing puppet content type in CVV removal

jjeffers:
This PR address the issues discovered in https://projects.theforeman.org/issues/27366 - when katello is installed without puppet CVV removal triggers an error, which blocks completion of the CVV removal task.

To reproduce the original error, set the puppet content in `config/settings.plugin.d/katello.yml` to false. Restart spring and the foreman web app. Create a product and a new repository, then sync that repo. Finally, create a Content View with that repository. Create 2 new CV versions, and then promote the oldest one. 

Delete the older CV version, opting to delete all environments. 

You will encounter an error, which on master manifests as:
`"RuntimeError: Cannot find repository type puppet, is it enabled?"` and the task will be blocked.

After applying this PR you should no longer have any errors when trying to remove e CV version when puppet is not a known content type.

### DIFF
--- a/app/lib/actions/katello/content_view_puppet_environment/destroy.rb
+++ b/app/lib/actions/katello/content_view_puppet_environment/destroy.rb
@@ -4,7 +4,9 @@ module Actions
       class Destroy < Actions::EntryAction
         def plan(puppet_env)
           action_subject(puppet_env)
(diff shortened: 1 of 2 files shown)


jjeffers:
> Works and code looks good! Just need to fix the RuboCop error.

For some reason the `ensure` keyword in that context is causing the error.

PR #8452, opened 2019-11-21
Fixes #28326 - Improve event daemon resilience

jturel:
If the polling thread for either of the event listeners dies we would be left in an unrecoverable state ie. events would not be handled until the daemon process is recycled by passenger or puma, or the server is altogether restarted. That is hopefully a never-happens scenario but now we are guarded from it by having a special monitor thread on the EventDaemon class itself. It will periodically ping the services and attempt to revive them if they aren't running. This design will continue to work well for us when we switch to Artemis.

**Notable changes in this PR:**

- `EventDaemon::Monitor` is now responsible for starting the services the first time (replacing old startup code) so that it's regularly tested at runtime.
- The Monitor updates the cached status of each of the services. Previously it was updated on every incoming event in the poller threads. The disadvantage there is that our app is using the file cache so it's 'slow', and if the cache expired or were cleared we would not correctly know the status. The Monitor ensures the cache is always updated even when no events are coming in.
- The health of each service is determined by the `:running` key of its status. It reports whether the thread is still running and any relevant connections are still established. With this in mind I've changed how we report service status in the Ping API. I removed the WARN state that was added in #8366 and replaced it with the FAIL state since we can truly know if it's running or not.
- Since we accurately know if the pollers are running I removed reporting of the queue depth on the Ping API as there is no specific action to take based on that number.

**To test:**

This is how to create a candlepin and katello event from the console respectively:

```
Katello::Resources::Candlepin::Consumer.update(::Host.last.subscription_facet.uuid, {'role' => SecureRandom.uuid})
```
```
Katello::Event.create!(event_type: 'import_host_applicability', object_id: 16)
```

With that in mind, exiting behavior can be verified by looking for the following events after making a request to the server for the first time:

```
2019-11-22T15:07:02 [I|app|94fd8786] Katello event daemon started process=7705
2019-11-22T15:07:02 [I|kat|94fd8786] Candlepin Event Listener started
2019-11-22T15:07:03 [I|kat|94fd8786] Polling Katello Event Queue
```

It's easiest to test the restart functionality of the monitor by starting and stopping qpidd:
```
systemctl stop qpidd
```
I found it useful to see the behavior by stopping qpidd before I started my rails server and once again after startup (with event daemon already running) to see that it was graceful in both cases. That should look something like this:

```
2019-11-22T15:12:19 [I|kat|94fd8786] Stopping Candlepin Listening Service
2019-11-22 15:12:19 [Security] warning Connect failed: Connection refused
2019-11-22T15:12:19 [E|kat|94fd8786] Unable to establish candlepin events connection: Failed to connect (reconnect disabled)

<qpidd is started again>

2019-11-22T15:12:49 [I|kat|94fd8786] Stopping Candlepin Listening Service
2019-11-22T15:12:49 [I|kat|94fd8786] Candlepin Event Listener started
``` 

Similar can be performed for the Katello events side by stopping postgresql and seeing that it also recovers when postgres is back

**About page looks like this:**
about_page

### DIFF
--- a/app/services/katello/candlepin_event_listener.rb
+++ b/app/services/katello/candlepin_event_listener.rb
@@ -1,8 +1,6 @@
 module Katello
   class CandlepinEventListener
-    PROCESSED_COUNT_CACHE_KEY = 'candlepin_events_processed'.freeze
-    FAILED_COUNT_CACHE_KEY = 'candlepin_events_failed'.freeze
-    AMQP_QUEUE_NAME = 'event.org.candlepin.audit.AMQPBusPublisher'.freeze
+    STATUS_CACHE_KEY = 'candlepin_events_status'.freeze
 
     CandlepinEvent = Struct.new(:message_id, :subject, :content)
 
@@ -10,38 +8,17 @@ module Katello
     @failed_count = 0
     @processed_count = 0
 
-    def self.start_service
-      loop do
-        begin
-          result = Katello::CandlepinListeningService.instance.start
-
-          break if result == :connected
-
-          @logger.info("Attempting to restart Candlepin Listening Service")
-          sleep 5
-        end
-      end
-    end
-
     def self.run
-      @thread.kill if @thread
+      initialize_listening_service
 
-      # run in own thread so connecting to qpid won't block the main process
-      @thread = Thread.new do
-        Rails.application.executor.wrap do
-          initialize_listening_service
-          start_service
+      result = Katello::CandlepinListeningService.instance.start
+      return unless result == :connected
 
-          Katello::CandlepinListeningService.instance.poll_for_messages do |message|
-            if message[:result]
-              result = message[:result]
-              event = CandlepinEvent.new(result.message_id, result.subject, result.content)
-              act_on_event(event)
-            elsif message[:error]
-              @logger.error("Disconnected from Candlepin Listening Service, reconnecting")
-              start_service
-            end
-          end
+      Katello::CandlepinListeningService.instance.poll_for_messages do |message|
+        if message[:result]
+          result = message[:result]
+          event = CandlepinEvent.new(result.message_id, result.subject, result.content)
+          act_on_event(event)
         end
       end
     rescue => e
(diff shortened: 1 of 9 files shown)


app/services/katello/candlepin_event_listener.rb
```diff
@@ -1,47 +1,24 @@
...
-    AMQP_QUEUE_NAME = 'event.org.candlepin.audit.AMQPBusPublisher'.freeze
+    STATUS_CACHE_KEY = 'candlepin_events_status'.freeze
 
     CandlepinEvent = Struct.new(:message_id, :subject, :content)
 
     @logger = ::Foreman::Logging.logger('katello/candlepin_events')
     @failed_count = 0
     @processed_count = 0
 
-    def self.start_service
```
jturel: Since the Monitor will try to restart the service if it loses connection (`status[:running]` would be false) this recovery code is totally redundant

app/services/katello/candlepin_event_listener.rb
```diff
@@ -1,47 +1,24 @@
...
-        end
-      end
-    end
-
     def self.run
-      @thread.kill if @thread
+      initialize_listening_service
 
-      # run in own thread so connecting to qpid won't block the main process
-      @thread = Thread.new do
```
jturel: Since the Monitor is running in a background thread we don't need to worry about blocking the main thread here any more, so I simplified this code

chris1984:
I like with the patch the ping changes showing more detail etc as well ❤️

chris1984:
Tested Katello/Candlepin threads by adding a pry in `lib/katello/middleware/event_daemon.rb`

Then running:

Candlepin:

```ruby
Katello::CandlepinListeningService.instance.instance_eval{@thread}
Katello::CandlepinListeningService.instance.instance_eval{@thread}.kill
```

We can see the thread starting up:
`2019-12-05T12:45:34 [I|kat|d4d80253] Candlepin Event Listener started`

Katello:

```ruby
Katello::EventMonitor::PollerThread.instance.instance_eval { @thread }
Katello::EventMonitor::PollerThread.instance.instance_eval { @thread }.kill
```

Process starting up
```ruby
2019-12-05T12:43:02 [I|app|e7aed641] Processing by Katello::Api::V2::KatelloPingController#index as HTML
2019-12-05T12:43:02 [I|app|e7aed641]   Parameters: {"api_version"=>"v2"}
2019-12-05T12:43:02 [I|kat|d4d80253] Stopping Katello Event Monitor
2019-12-05T12:43:02 [I|kat|d4d80253] Polling Katello Event Queue
```

PR #8453, opened 2019-11-22
Fixes #28327 - Allow users in org to access subs

sjha4:
**Current flow:**
Cannot access katello/api/v2/subscriptions/:id without specifying org id

**Updated flow:**
Allow access to katello/api/v2/subscriptions/:id  if the current user belongs to the org of the subscription and deny permission if not.

**To Test:**
Create an organization
Create a user and add the new org to the user.
Try to access a subscription in a different org. Should see the error.
Add user to the subscription.organization. Should see the subscription without error.

### DIFF
--- a/app/controllers/katello/api/v2/subscriptions_controller.rb
+++ b/app/controllers/katello/api/v2/subscriptions_controller.rb
@@ -77,8 +77,9 @@ module Katello
     def show
       org_id = Organization.current&.id
       @resource = Katello::Pool.with_identifier(params[:id])
-      if @resource.organization_id != org_id
-        fail ActiveRecord::RecordNotFound, N_('This subscription is not relevant to the current organization.')
+      fail ActiveRecord::RecordNotFound, N_('Subscription not found') unless @resource
+      if @resource.organization_id != org_id && !User.current.organizations&.pluck(:id)&.include?(@resource.organization_id)
+        fail ActiveRecord::RecordNotFound, N_('This subscription is not relevant to the current user and organization.')
       end
       respond(:resource => @resource)
     end
--- a/test/controllers/api/v2/subscriptions_controller_test.rb
+++ b/test/controllers/api/v2/subscriptions_controller_test.rb
@@ -86,6 +86,35 @@ module Katello
       assert_template 'api/v2/subscriptions/index'
     end
 
+    def test_show_with_org_id
+      @subscription = katello_pools(:pool_one)
+      @org = get_organization(:empty_organization)
+      get :show, params: { :id => @subscription.id, :organization_id => @org.id }
+
+      assert_response :success
+      assert_template 'api/v2/subscriptions/show'
+    end
+
+    def test_show_without_user_org_id
+      @subscription = katello_pools(:pool_one)
+      @org = get_organization(:empty_organization)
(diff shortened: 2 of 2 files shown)


app/controllers/katello/api/v2/subscriptions_controller.rb
```diff
@@ -77,8 +77,8 @@ def index_relation
     def show
       org_id = Organization.current&.id
       @resource = Katello::Pool.with_identifier(params[:id])
-      if @resource.organization_id != org_id
-        fail ActiveRecord::RecordNotFound, N_('This subscription is not relevant to the current organization.')
+      if @resource.organization_id != org_id && !User.current.organizations&.pluck(:id)&.include?(@resource.organization_id)
+        fail ActiveRecord::RecordNotFound, N_('This subscription is not relevant to the current user and organization.')
```
jturel: Your change is consistent with what's here and it fixes the problem, but I am curious about what's here (cc @jeremylenz )

If try to lookup a subscription in an organization which I don't have access to I should get a 404, and that is what happens. However, I should not be informed of whether the subscription exists or not. Since I don't have access to the organization it's irrelevant. On the other hand if that subscription *is* in my organization and I lack permission to view it, that would be permission denied error (403).

@jeremylenz is the message you added here for consumption in the UI?

jturel: When I give the API a bogus subscription ID I get a stack trace. Might be worth addressing here as well:

```
curl -X GET -k -u qauser:changeme http://localhost:3000/katello/api/v2/organizations/4/subscriptions/5000
{"displayMessage":"undefined method `organization_id' for nil:NilClass","errors":["undefined method `organization_id' for nil:NilClass"]}```

Comment (jeremylenz) [reply app/controllers/katello/api/v2/subscriptions_controller.rb]: @jturel Yes, this was to address the user flow of changing the org while you happen to be on the subscription details page.  The error message shows up as a Toast notification.

Comment (jeremylenz) [reply app/controllers/katello/api/v2/subscriptions_controller.rb]: See [comment](https://github.com/Katello/katello/pull/8262#issuecomment-527635305) on 25113

**Comment by sjha4** [comment]
@jturel , @jeremylenz : Does this look good?

**Comment by jeremylenz** [comment]
@sjha4 I'll defer to @jturel; looks fine to me!

**Comment by jturel** [comment]
> @jturel , @jeremylenz : Does this look good?

Looks like it will do the trick! I'll give it a test

File: app/controllers/katello/api/v2/subscriptions_controller.rb (hunk 1 above)
Comment (jturel) [reply app/controllers/katello/api/v2/subscriptions_controller.rb]: Thanks for confirming!

It looks like the Pool model isn't fully integrated into our permission system:

```
[7] pry(#<Katello::Api::V2::SubscriptionsController>)> ::Katello::Pool.authorized
NoMethodError: undefined method `authorized' for #<Class:0x00007f7ff4f60c18>
from /home/vagrant/foreman/.vendor/ruby/2.5.0/gems/activerecord-5.2.1/lib/active_record/dynamic_matchers.rb:22:in `method_missing'
```

That's a separate issue and I know we have one other files for Repositories so I won't file another since I think we'll look at permissions at a high level when addressing it. That said, this change does seem to address what was reported

PR #8455, opened 2019-11-25
Fixes #28349 - refactors importer comparison

jjeffers:
This PR addresses a pulp2 bug where the proxy password is returned as a masked string.

```
>> in the pulp2 db

> db.repo_importers.find({})
{ "_id" : ObjectId("5dd6ce3506c71a7bd3d7623f"), "repo_id" : "5bc050e0-88d7-4f8d-8a75-ab2516b53b74", "importer_type_id" : "yum_importer", "config" : { "feed" : "https://repos.fedorapeople.org/pulp/pulp/fixtures/rpm-unsigned/", "remove_missing" : true, "proxy_host" : "", "ssl_validation" : true, "proxy_password" : "", "download_policy" : "immediate", "proxy_username" : "" }, "last_updated" : ISODate("2019-11-22T14:52:10.174Z"), "last_override_config" : { "num_threads" : 4, "validate" : true }, "_ns" : "repo_importers", "scratchpad" : { "repomd_revision" : 1573300917, "repomd_checksum" : "734ef2769824aeb357165a7959b0c942fad7ce0ba5d8bde12c145d7f4fa0c231" }, "last_sync" : "2019-11-21T17:50:14Z" }

>> Check repos details using pulp-admin

[vagrant@centos7-katello-devel-stable foreman]$ pulp-admin -u admin -p admin repo list --details --repo-id 5bc050e0-88d7-4f8d-8a75-ab2516b53b74
+----------------------------------------------------------------------+
                              Repositories
+----------------------------------------------------------------------+

Id:                   5bc050e0-88d7-4f8d-8a75-ab2516b53b74
Display Name:         test
Description:          None
Content Unit Counts:  
  Erratum:           4
  Package Category:  1
  Package Group:     2
  Package Langpacks: 1
  Rpm:               35
Notes:                
Scratchpad:           
  Checksum Type: sha256
Importers:            
  Config:               
    Download Policy: immediate
    Feed:            https://repos.fedorapeople.org/pulp/pulp/fixtures/rpm-unsig
                     ned/
    Proxy Host:      
    Proxy Password:  *****
    Proxy Username:  
   [...]
```

This condition triggers repo refreshes unnecessarily.

### DIFF
--- a/app/services/katello/pulp/importer_comparison.rb
+++ b/app/services/katello/pulp/importer_comparison.rb
@@ -6,21 +6,25 @@ module Katello
           return false
         end
 
-        if generated_importer.config['proxy_host'] == ""
-          generated_importer.config.delete('proxy_host')
-          generated_importer.config.delete('proxy_username')
-          generated_importer.config.delete('proxy_password')
-          generated_importer.config.delete('proxy_port')
-          capsule_importer['config'].delete('proxy_host')
-          capsule_importer['config'].delete('proxy_username')
-          capsule_importer['config'].delete('proxy_port')
-          capsule_importer['config'].delete('proxy_password')
+        generated_config = generated_importer.config
+        capsule_config = capsule_importer['config']
+        if generated_config['proxy_host'] == ""
+          generated_config.delete('proxy_host')
+          generated_config.delete('proxy_username')
+          generated_config.delete('proxy_password')
+          generated_config.delete('proxy_port')
+          capsule_config.delete('proxy_host')
+          capsule_config.delete('proxy_username')
+          capsule_config.delete('proxy_password')
+          capsule_config.delete('proxy_port')
         end
-        if generated_importer.config['proxy_password'] == "" && capsule_importer['config']['proxy_password'] == "*****"
-          generated_importer.config.delete('proxy_password')
-          capsule_importer['config'].delete('proxy_password')
+
+        if generated_config['proxy_password'] == "" && capsule_config['proxy_password'] == "*****"
+          generated_config.delete('proxy_password')
+          capsule_config.delete('proxy_password')
         end
-        generated_importer.config.compact == capsule_importer['config'].compact
+
+        generated_config.compact == capsule_config.compact
       end
     end
   end
--- a/app/services/katello/pulp/importer_comparison.rb
+++ b/app/services/katello/pulp/importer_comparison.rb
@@ -9,14 +9,11 @@ module Katello
         generated_config = generated_importer.config
         capsule_config = capsule_importer['config']
         if generated_config['proxy_host'] == ""
-          generated_config.delete('proxy_host')
-          generated_config.delete('proxy_username')
-          generated_config.delete('proxy_password')
-          generated_config.delete('proxy_port')
-          capsule_config.delete('proxy_host')
-          capsule_config.delete('proxy_username')
-          capsule_config.delete('proxy_password')
-          capsule_config.delete('proxy_port')
+          proxy_keys = %w(proxy_host proxy_username proxy_password)
+          proxy_keys.each do |key|
+            generated_config.delete(key)
+            capsule_config.delete(key)
+          end
         end
 
         if generated_config['proxy_password'] == "" && capsule_config['proxy_password'] == "*****"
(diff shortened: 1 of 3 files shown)


jturel:
:fire: :fire: :fire: :fire: :fire:

app/services/katello/pulp/importer_comparison.rb
```diff
@@ -0,0 +1,31 @@
...
+          capsule_config.delete('proxy_host')
+          capsule_config.delete('proxy_username')
+          capsule_config.delete('proxy_password')
+          capsule_config.delete('proxy_port')
+        end
+
+        if generated_config['proxy_password'] == "" && capsule_config['proxy_password'] == "*****"
+          generated_config.delete('proxy_password')
+          capsule_config.delete('proxy_password')
+        end
```
jlsherrill: to clarify, if `generated_config['proxy_password']` is set, it will always refresh the importer.  Is that what we want?

app/services/katello/pulp/importer_comparison.rb
```diff
@@ -0,0 +1,31 @@
...
+        capsule_config = capsule_importer['config']
+        if generated_config['proxy_host'] == ""
+          generated_config.delete('proxy_host')
+          generated_config.delete('proxy_username')
+          generated_config.delete('proxy_password')
+          generated_config.delete('proxy_port')
+          capsule_config.delete('proxy_host')
+          capsule_config.delete('proxy_username')
+          capsule_config.delete('proxy_password')
+          capsule_config.delete('proxy_port')
```
jlsherrill: kind of a nitpick, but thoughts on something like this instead:
```
proxy_keys = %w(proxy_host proxy_username proxy_password)
proxy_keys.each do |key| 
   generated_config.delete(key)
   capsule_config.delete(key)
end
```

app/services/katello/pulp/repository.rb
```diff
@@ -139,7 +140,8 @@ def needs_importer_updates?
         repo_details = backend_data
         return unless repo_details
         capsule_importer = repo_details["importers"][0]
-        !importer_matches?(capsule_importer)
+        generated_importer = generate_importer
+        !importer_matches?(generated_importer, capsule_importer)
```
jlsherrill: another nitpick, but i'd probably not store the generate_importer return output and just call it in the next line

app/services/katello/pulp/importer_comparison.rb
jjeffers: Is your suggestion to _always_ ignore `proxy_password` in the comparison? Since pulp2 will always return that masked string (if set previously to any value), it seems like any comparison will fail.

ianballou:
I gave this a test by calling `refresh_if_needed` directly on the repository's `backend_service`.  Without the PR I could see the unnecessary actions, and with it applied the actions went away.

app/services/katello/pulp/importer_comparison.rb
jlsherrill: possibly, but i was more opening up debate :)  
I think this is the safter thing to do is the way you have it, although it makes https://github.com/Katello/katello/pull/8415 not really work if you're using an http proxy, although thats not really a problem introduced by this pr.  So i guess yes lets leave it the way you have it :+1:   but we should probably open a bug around this to discuss how to properly address it (especially with pulp3)

jjeffers: Hopefully this _isn't_ an issue in pulp3.

jjeffers: We could ask pulp2 to be fixed to resolve this. That's another option.

jlsherrill: I checked and pulp3 returns the full proxy url:
```
{"pulp_href":"/pulp/api/v3/remotes/file/file/c6688395-58f8-461a-bd3e-145083741543/","pulp_created":"2019-11-26T19:11:49.606085Z","name":"proxytest-137322","url":"http://foo.com/bar/PULP_MANIFEST","ca_cert":null,"client_cert":null,"client_key":null,"tls_validation":true,"proxy_url":"http://boobar:boobar@foo.com","pulp_last_updated":"2019-11-26T19:12:43.872594Z","download_concurrency":20,"policy":"immediate"}
```

however i suspect that might be a bug ;)

I think this is fine for now!

jlsherrill:
don't forget to squash!

PR #8458, opened 2019-12-03
Fixes #28413 - Create product with permissions

parthaa:
This commit relaxes a permission requirement to create custom
products. Prior to this commit the custom product create required a user
do to have create_products, edit_products, and view_subscriptions.
However it is not entirely clear to the user that view_subscriptions is
needed to create a product. Since unlimited subscriptions happens to be
a side effect of Product creation, it makes sense for create/edit
products to be permission enough.

This commit precludes the need for view_subscriptions to create
unlimited subscription on product create.

### DIFF
--- a/app/models/katello/glue/candlepin/product.rb
+++ b/app/models/katello/glue/candlepin/product.rb
@@ -101,13 +101,19 @@ module Katello
 
       def import_custom_subscription
         fail _("Cannot import a custom subscription from a redhat product.") if self.redhat?
-        sub = nil
-        ::Katello::Util::Support.active_record_retry do
-          sub = ::Katello::Subscription.where(:cp_id => self.cp_id, :organization_id => self.organization.id).first_or_create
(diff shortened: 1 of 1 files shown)


app/models/katello/glue/candlepin/product.rb
```diff
@@ -101,13 +101,19 @@ def add_content(content)
...
+        User.as_anonymous_admin do
+          sub = nil
+          ::Katello::Util::Support.active_record_retry do
+            sub = ::Katello::Subscription.where(:cp_id => self.cp_id, :organization_id => self.organization.id).first_or_create
+          end
+          unless sub.persisted?
+            message = _("Subscription was not persisted - %{error_message}") % {:error_message => sub.errors.full_messages.join("; ")}
+            fail HttpErrors::UnprocessableEntity, message
+          end
+          sub.import_data
```
jlsherrill: would it make sense to extract any of this to another method?

PR #8460, opened 2019-12-04
[DONTMERGE] Test the tests

jturel:


test/actions/katello/event_queue_monitor_test.rb
```diff
@@ -8,7 +8,7 @@ class Actions::Katello::EventQueueMonitorTest < ActiveSupport::TestCase
     let(:polling_class) { ::Actions::Katello::EventQueue::PollerThread }
     let(:suspended_class) { ::Actions::Katello::EventQueue::SuspendedAction }
     let(:planned_action) do
-      action_class.any_instance.expects(:singleton_lock!)
+      action_class.any_instance.expects(:singleton_lock!).returns
```
jturel: @jjeffers The unit tests are green with this change. I don't think the dynflow version lock would be required. Give it a shot!

jjeffers: I'll give that a shot, thanks for looking into this.

PR #8464, opened 2019-12-07
Fixes #28445: Update require-dir dependency

ehelms:
This is designed to fix the angular-ui tests. But currently the react-ui tests are also failing.

### DIFF
--- a/engines/bastion_katello/package.json
+++ b/engines/bastion_katello/package.json
@@ -2,6 +2,24 @@
   "name": "bastion_katello",
   "version": "1.0.0",
   "dependencies": {
-    "bastion": "file:../bastion"
+    "bastion": "file:../bastion",
+    "eslint": "~0.14.1",
+    "eslint-plugin-angular": "0.0.3",
+    "generator-bastion": "~0.1.2",
+    "grunt": "~0.4.5",
+    "grunt-angular-gettext": "~0.2.15",
+    "grunt-bower-task": "yatskevich/grunt-bower-task#v0.5.0",
+    "grunt-concurrent": "~1.0.0",
+    "grunt-eslint": "~6.0.0",
+    "grunt-htmlhint": "~0.4.1",
+    "grunt-karma": "^1.0.0",
+    "jasmine-core": "^2.4.1",
+    "karma": "^0.13.22",
+    "karma-coverage": "^1.0.0",
+    "karma-jasmine": "^1.0.2",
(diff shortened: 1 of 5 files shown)


ehelms:
Now also includes a fix for the broken UI snapshots in the React UI as well. This is designed to fix all current CI issues.

ekohl:
I think the grunt failure might be due to us installing a very old version on our slaves that's no longer compatible with node 12.

engines/bastion_katello/package.json
```diff
@@ -2,6 +2,24 @@
...
+    "grunt-htmlhint": "~0.4.1",
+    "grunt-karma": "^1.0.0",
+    "jasmine-core": "^2.4.1",
+    "karma": "^0.13.22",
+    "karma-coverage": "^1.0.0",
+    "karma-jasmine": "^1.0.2",
+    "karma-ng-html2js-preprocessor": "^1.0.0",
+    "karma-phantomjs-launcher": "^1.0.0",
+    "require-dir": "~1.2.0",
+    "phantomjs-prebuilt": "2.1.10"
```
ehelms: Re-defining all of these here is kinda ugly, but this code rarely changes so I think it's OK. Something in updating to NodeJS 10+ seems to have affected how all the relative imports work with Grunt.

webpack/components/Content/Details/__tests__/__snapshots__/ContentDetails.test.js.snap
```diff
@@ -7,7 +7,7 @@ exports[`Content Details Info should render and contain appropriate components 1
     loadingText="Loading"
     timeout={300}
   >
-    <ForwardRef
+    <Uncontrolled(TabContainer)
```
waldenraines: Is this intentional?  Seems Odd.

jeremylenz: @waldenraines I think this is related to a patternfly change.

ehelms: Foreman had similar changes once vendor v3 was released -- https://github.com/theforeman/foreman/commit/ae55b7eff2f2e653981781d3b73fc7d1361f1952

jturel:
Any interest in assuaging prprocesser?

ehelms:
Fixed prprocessor

PR #8468, opened 2019-12-09
Fixes #28427 - Caches counts in a CVV

parthaa:
Content view versions list api ran a count query for every content type
belonging to a content view. This means atleast 10 queries for each
content type and can get out of hand when one is returning over 40
versions in a list api call

This commit addresses the above issue by storing the counts in a
serializable hash for each content view version on every publish and
that cached result is returned when content_counts is called.
This should result in a speed up in the api response

### DIFF
--- a/app/lib/actions/katello/content_view/publish.rb
+++ b/app/lib/actions/katello/content_view/publish.rb
@@ -92,6 +92,8 @@ module Actions
         end
 
         def finalize
+          version = ::Katello::ContentViewVersion.find(input[:content_view_version_id])
+          version.update_content_counts!
           # update errata applicability counts for all hosts in the CV & Library
           ::Katello::Host::ContentFacet.where(:content_view_id => input[:content_view_id],
                                               :lifecycle_environment_id => input[:environment_id]).each do |facet|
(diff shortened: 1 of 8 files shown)


parthaa:
screenshot-zeta sparta example com_3000-2019 12

parthaa:
> I noticed an issue with CVV Docker tag counts with this PR. The Docker tag count appears to include tags from both schemas. In my case there should be 8, but it's 10 and two of the tags belong in schemas 1 and 2.

Hmm can you verify this is not the case in master?

ianballou:
@parthaa I was switching between master and this PR and then refreshing the page.  The tag count was flipping between 8 and 10.  It could be a bug that your PR uncovered perhaps?

parthaa:
> @parthaa I was switching between master and this PR and then refreshing the page. The tag count was flipping between 8 and 10. It could be a bug that your PR uncovered perhaps?

You are right. Just fixed :). Try again

parthaa:
> You are right. Just fixed :). Try again

Note you'd need to run -> ```Katello::ContentViewVersion.all.each(&:update_content_counts!)``` in your rails console to see the counts updated.

parthaa:
[test katello]

parthaa:
[test katello]

PR #8471, opened 2019-12-10
Refs #28194 - correct flaky result ordering

jjeffers:
This change updates a test that introduced an ordering of an AR collection which causes flaky test outcomes.

### DIFF
--- a/test/actions/pulp3/orchestration/rpm_remove_units_test.rb
+++ b/test/actions/pulp3/orchestration/rpm_remove_units_test.rb
@@ -12,6 +12,7 @@ module ::Actions::Pulp3
 
       create_and_sync_repo(@repo, @master)
       @repo.reload
+      @rpm = Katello::Rpm.find_by(filename: "frog-0.1-1.noarch.rpm")
     end
 
     def create_and_sync_repo(repo, proxy)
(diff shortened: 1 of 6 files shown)


test/actions/pulp3/orchestration/rpm_remove_units_test.rb
```diff
@@ -12,6 +12,7 @@ def setup
 
       create_and_sync_repo(@repo, @master)
       @repo.reload
+      @rpm = Katello::Rpm.all.where(filename: "frog-0.1-1.noarch.rpm")[0]
```
jlsherrill: Few nitpicks:  i'd probably look in the repo for the rpm instead of querying it across Rpm.all (i think that'd be less likely to fail in the future).  I'd also use .find_by instead of .where()[0]

test/actions/pulp3/orchestration/rpm_remove_units_test.rb
```diff
@@ -34,15 +35,15 @@ def create_and_sync_repo(repo, proxy)
     end
 
     def test_remove_rpm
-      content_unit = @repo.repository_rpms.first
+      content_unit = @repo.repository_rpms.where(rpm_id: @rpm.id)[0]
```
jlsherrill: can you just reference @rpm ?

PR #8472, opened 2019-12-11
Fixes #28183 - handle repo refresh with pulp3

jlsherrill:
This does a few things:

1) ensures that manifest refreshing will examine any affected redhat repo 
      and check to see if the distribution or remote needs updating and do that
2) Changes how we track distributions.  Previously we were mainly doing it 
     by id.  The DistributionReference object really wasn't used all that much, 
    and we'd just look up by path and then make any changes we need.  
    However if we are looking up by path, it makes it impossible to update 
    the distribution when the path changes (because at this point in the
    service layer, we don't know what the old path is).  So this adjusts 
    the DistributionReference to store the repository_id instead of 
    the root_repository_id.  Its actually quite complicated to migrate 
   this data, so this will simply delete any distribution reference that 
   exists, so you really need to reset your katello before using this.


jlsherrill:
TODO: 
- [ ] add more tests

To test:

1) import a manifest
2) enable a yum repository
3) refresh the manifest
4) examine the dynflow task of the manifest refresh and look for the RefreshIfNeeded and look for the ouptut.  You should see no tasks.  (althought its possible you may, if you do refresh again and check the new task, where you should see no tasks)
5) find the id of the yum repository
6) update an option that would affect the remote:  ```  Katello::Repository.find(6).root.update_attributes(:download_policy => 'immediate')  ```
7)  update an option that would affect the distribution:  ``` Katello::Repository.find(6).update_attributes(:relative_path => '/foo/bar')  ```
8)  Refresh the manifest
9) go to the dynflow console and look at the RefreshIfNeeded task.  Verify that you see 2 tasks.

ianballou:
Testing worked as prescribed! Reading through the code now.

PR #8473, opened 2019-12-11
Fixes #16333 - sort CCV version based on IDs

snagoor:
Sorting order doesn't work based on `version`, switching it to be sorted based on `id`. Tested this on katello 3.14 and works as expected.


chris1984:
Starting to test

chris1984:
Thanks @snagoor 🍻

PR #8474, opened 2019-12-11
Fixes #28478 - Pulp 3 worker rename fix

ianballou:
To test:
1) Make a brand-new Pulp 3 dev box.
2) Check the 'About' page in Katello.  Notice Katello thinks Pulp 3 isn't running even though all of the Pulp 3 services are running.
3) Apply this PR.
4) Refresh the page.
5) See that Katello now says Pulp 3 is OK.

### DIFF
--- a/app/models/katello/ping.rb
+++ b/app/models/katello/ping.rb
@@ -193,10 +193,9 @@ module Katello
         end
 
         workers = json["online_workers"] || []
-        resource_manager_exists = workers.any? { |worker| worker["name"].include?("resource-manager@") }
-        reservered_resource_worker_exists = workers.any? { |worker| worker["name"] =~ /reserved-resource-worker-./ }
+        resource_manager_exists = workers.any? { |worker| worker["name"].include?("resource-manager") }
 
-        unless resource_manager_exists && reservered_resource_worker_exists
+        unless resource_manager_exists && workers.count > 1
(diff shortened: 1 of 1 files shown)


app/models/katello/ping.rb
```diff
@@ -193,8 +193,8 @@ def pulp3_without_auth(url)
         end
 
         workers = json["online_workers"] || []
-        resource_manager_exists = workers.any? { |worker| worker["name"].include?("resource-manager@") }
-        reservered_resource_worker_exists = workers.any? { |worker| worker["name"] =~ /reserved-resource-worker-./ }
+        resource_manager_exists = workers.any? { |worker| worker["name"].include?("resource-manager") }
+        reservered_resource_worker_exists = workers.any? { |worker| worker["name"] =~ /#{Socket.gethostbyname(Socket.gethostname).first}/ }
```
jlsherrill: I would probably avoid trying to match the hostname (which wouldn't work if we ever support multi-box scenarios).  Maybe its good enough to just count any workers otehr than the resource-manager?

ianballou: yeah that's true, I'll just make sure there's at least one other worker

ianballou:
@jlsherrill all set now.

ianballou:
[test katello]

ianballou:
[test katello]

jlsherrill:
[test katello]

PR #8475, opened 2019-12-11
Fixes #28489 - properly handle no-arch repos

jlsherrill:
the rhel8 tools repo does not have an arch variable
within its content url, and the UI seemed to confuse
the displayArch with what is actually needed to enable
or disable a repo


jlsherrill:
[test katello]

jlsherrill:
[test katello]

PR #8476, opened 2019-12-12
Fixes #28423 - bump react-router-bootstrap

laviro:
bump react-router-bootstrap in foreman vendor-core
to the latest version (released 9 months ago)
and it seem to solve the issue

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,9 +26,9 @@
     "@babel/core": "^7.7.0",
     "@storybook/react": "^3.2.17",
     "@storybook/storybook-deployer": "^2.0.0",
-    "@theforeman/builder": "^3.3.2",
-    "@theforeman/env": "^3.3.2",
-    "@theforeman/vendor-dev": "^3.3.2",
+    "@theforeman/builder": "^3.8.0",
+    "@theforeman/env": "^3.8.0",
+    "@theforeman/vendor-dev": "^3.8.0",
     "axios-mock-adapter": "^1.10.0",
     "babel-eslint": "^10.0.3",
     "babel-jest": "^24.9.0",
@@ -54,7 +54,7 @@
     "redux-mock-store": "^1.3.0"
   },
   "dependencies": {
-    "@theforeman/vendor": "^3.3.2",
+    "@theforeman/vendor": "^3.8.0",
     "angular": "1.5.5",
     "bootstrap-select": "1.12.4",
     "downshift": "^1.28.0",
--- a/webpack/scenes/Subscriptions/UpstreamSubscriptions/__tests__/__snapshots__/UpstreamSubscriptionsPage.test.js.snap
+++ b/webpack/scenes/Subscriptions/UpstreamSubscriptions/__tests__/__snapshots__/UpstreamSubscriptionsPage.test.js.snap
@@ -200,11 +200,7 @@ exports[`upstream subscriptions page should render 1`] = `
         >
           Submit
         </Button>
-        <LinkContainer
-          activeClassName="active"
-          exact={false}
-          replace={false}
-          strict={false}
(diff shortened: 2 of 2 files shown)


waldenraines:
> I restarted rails and I still get the error in the description. I even tried the big hammer of `rm -fr node_modules` before `npm i`, no luck.

Hmm, the plot thickens.  After I `rm -fr node_modules` in foreman and `npm i` this fixes the issue for me.  I'm not 100% sure that this actually fixes the issue or if it's the release of the new vendor but foreman having 3.8.0 is the ticket.

laviro:
That's really weird, I can confirm what you saw,
@waldenraines @jeremylenz @sharvit @amirfefer what do you think could cause Foreman's `react-router-bootstrap` update to affect katello? 

maybe there is a wrapper to the LinkContainer in SubscriptionToolbar which is imported from Foreman?  I am off for the weekend, glad it fixed it anyway.. but it's worth investigating more...

jeremylenz:
Katello using vendor 3.8 doesn't seem to affect the version used by Foreman, and thus
```
npm list | grep react-router-bootstrap
```
shows 
```
react-router-bootstrap@0.24.4
```

until I upgrade all of Foreman's `@theforeman/*` packages to `^3.8.0` and `npm install`.  After doing this, the above command showed 
```
react-router-bootstrap@0.25.0
```
and the issue was fixed for me.

laviro:
I encountered the same flow as Walden mentioned:
1. Katello upgraded to vendor 3.8.0
2. Foreman using 3.7.x 
3. Subscription page is blank with the error
4. Upgrading Foreman to use vendor 3.8.0
5. Subscription is working as expected.

That's why I think there's sort of a "leak" from Foreman to Katello

jeremylenz:
My guess is it's something to do with how module names are resolved.. maybe it's looking first in the `vendor` provided by Foreman and then only if it doesn't find it it will go to Katello's `node_modules`.

waldenraines:
I'm fairly certain that all of the duplicate dependencies in katello's `package.json` are only there for testing.  In other words, I think that anything in foreman's `package.json` overrides katello's.

sharvit:
> I'm fairly certain that all of the duplicate dependencies in katello's `package.json` are only there for testing. In other words, I think that anything in foreman's `package.json` overrides katello's.

It's true!

In **testing and in the storybook** environment, it will run against your **local** `vendor-dev`.
But in a **server** environment (dev or prod), it will run against the `vendor` supplied by the **foreman core**.

IMO opinion it's a good practice to have the `vendor` as a dependency so at least you declare about the vendor version range your plugin can work with. It won't fix or change anything but it will give developers a clue information compatibility,

sharvit:
I'm wondering if we should change the policy and put the `vendor` as a `peerDependency`?

The benefits would be:
1. **Faster** - It won't install the vendor when running `npm i`
2. **Less confusion** - Developers would understand that the vendor should be supplied by someone else.

waldenraines:
> I'm wondering if we should change the policy and put the `vendor` as a `peerDependency`?

I like this idea, I'll try it.

jeremylenz:
@sharvit will this cause problems with packaging? https://github.com/theforeman/npm2rpm#peerdependencies

waldenraines:
> @sharvit will this cause problems with packaging? https://github.com/theforeman/npm2rpm#peerdependencies

It sounds like this shouldn't be a problem on all but the most ancient of machines:

> On npm versions less than 3.0.0...

Also, I'm not sure that we even run the tests post packaging so it shouldn't matter in that sense.

waldenraines:
Anywho, we can continue this conversation in https://github.com/Katello/katello/pull/8477

PR #8478, opened 2019-12-13
Fixes #28459 - index Pulp 3 package groups and SRPMs

ianballou:
This PR adds support for Pulp 3 package groups and SRPMs.

To test:
1) Create a yum repo with the upstream url https://repos.fedorapeople.org/pulp/pulp/fixtures/srpm/
2) Sync this repo
3) Check that it has 3 SRPMs and 2 package groups (2 errata too).

### DIFF
--- a/app/services/katello/pulp3/package_group.rb
+++ b/app/services/katello/pulp3/package_group.rb
@@ -1,0 +1,66 @@
+module Katello
+  module Pulp3
+    class PackageGroup < PulpContentUnit
+      include LazyAccessor
+
+      lazy_accessor :optional_package_names, :mandatory_package_names,
+                    :conditional_package_names, :default_package_names, :_id,
+                    :repository_memberships,
+                    :initializer => :backend_data
+
+      def _id
+        backend_data['pulp_href']
+      end
+
+      # Package type is now an integer:
+      # 1. default
+      # 2. optional
+      # 3. conditional
+      # 4. mandatory
+      # 5. unknown
+      # https://github.com/rpm-software-management/libcomps/blob/01a4759894cccff64d2561614a58281adf5ce859/libcomps/src/comps_docpackage.h#L36
+      def default_package_names
+        package_names_of_type(1)
+      end
+
+      def optional_package_names
+        package_names_of_type(2)
+      end
+
+      def conditional_package_names
+        package_names_of_type(3)
+      end
+
+      def mandatory_package_names
+        package_names_of_type(4)
+      end
+
+      def unknown_package_names
+        package_names_of_type(5)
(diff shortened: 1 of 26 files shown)


ianballou:
[test katello]

app/services/katello/pulp3/rpm.rb
```diff
@@ -19,6 +19,20 @@ def self.ids_for_repository(repo_id)
...
+        ia64, alpha, alphaev5, alphaev56, alphapca56, alphaev6, alphaev67,
+        sparc, sparcv8, sparcv9, sparc64, sparc64v, sun4, sun4c, sun4d, sun4m, sun4u,
+        armv3l, armv4b, armv4l, armv5tel, armv5tejl, armv6l, armv7l,
+        mips, mipsel,
+        ppc, ppciseries, ppcpseries, ppc64, ppc8260, ppc8560, ppc32dy4,
+        m68k, m68kmint, atarist, atariste, ataritt, falcon, atariclone, milan, hades,
+        Sgi,
+        rs6000,
+        i370, s390x, s390,
+        noarch"
```
johnpmitsch: why a huge string vs. an array which you call `.join(", ")` on?

app/services/katello/pulp3/package_group.rb
```diff
@@ -0,0 +1,67 @@
...
+        package_names_of_type(4)
+      end
+
+      def unknown_package_names
+        package_names_of_type(5)
+      end
+
+      def package_names_of_type(type)
+        packages = []
+        backend_data['packages'].each { |package| packages << package['name'] if package['type'] == type }
```
johnpmitsch: How about something like this to avoid initializing the empty array?
```ruby
def package_names_of_type(type)
  filtered_packages = backend_data['packages'].select { |package| package['type'] == type }
  filtered_packages.map { |package| package['name'] }
end
```

app/services/katello/pulp3/rpm.rb
johnpmitsch: To clarify, I find arrays are easier to update and manage on multiple lines, but not sure if there is a reason for the string

ianballou: yeah I'll make it an array.  This is temporary and I was hoping to change it before merging but I'm not sure if it'll be fixed in pulp soon enough

app/services/katello/pulp3/package_group.rb
ianballou: Sure that works

ianballou:
@johnpmitsch think I addressed your questions.  Architecture spacing is done by platform.

PR #8480, opened 2019-12-17
Fixes #28446 - reverting snap update

jjeffers:
This change reverts (via `npm test -- -u`) changes made that originally were needed to address an upgrade to a dependency in npm land. Now that update was reverted, and so we need to revert our tests.


jlsherrill:
For the commit message, this should not Ref since its a ~couple weeks old, but should instead 'Fixes'

PR #8483, opened 2019-12-19
Fixes #28551 - use sqlite friendly not_null add_column

jlsherrill:


jlsherrill:
To test:

1. edit foreman/config/database.yml and comment out the test db, add:
```
test:
  adapter: sqlite3
  database: db/foo.sqlite
```

2. run: 
```
RAILS_ENV=test be rake db:migrate
```

It will fail without this change, and pass with it.
Make sure to revert it back after you are done :)

jlsherrill:
This fixed the rpm build error: http://koji.katello.org/kojifiles/work/tasks/8778/258778/build.log

jjeffers:
Tested, and it worked just as it said on the tin.

PR #8484, opened 2019-12-19
Fixes #28350 - docker manifests not showing tags

ianballou:
To test:

1) Sync a Docker repo like library/hello-world.
2) Look at the repo's manifests and manifest lists.
3) See that the tags are populated.

### DIFF
--- a/app/views/katello/api/v2/docker_manifest_lists/show.json.rabl
+++ b/app/views/katello/api/v2/docker_manifest_lists/show.json.rabl
@@ -4,7 +4,8 @@ attributes :id, :schema_version, :digest, :manifest_type
 
 child :docker_tags => :tags do
   attributes :associated_meta_tag_identifier => :id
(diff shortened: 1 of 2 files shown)


app/views/katello/api/v2/docker_manifest_lists/show.json.rabl
```diff
@@ -4,7 +4,7 @@ attributes :id, :schema_version, :digest, :manifest_type
 
 child :docker_tags => :tags do
   attributes :associated_meta_tag_identifier => :id
-  attributes :repository_id, :name
+  attributes :repository, :name
```
jlsherrill: is it possible to do this without changing the api?

something like  attributes :repository => :repository_id

ianballou: @jlsherrill yeah this works fine.

PR #8487, opened 2020-01-02
Fixes #28621 - properly show repo after disabling

jlsherrill:
for a couple of reasons, a repo will not show up properly
without a page refresh:

1) the content has no release version substitution, in which
   case the repo comparison will fail when comparing undefined to null
2) the server stores 'noarch' when there is no arch substituion, and the
   comparision does not recognize this

### DIFF
--- a/webpack/redux/reducers/RedHatRepositories/repositorySetRepositories.js
+++ b/webpack/redux/reducers/RedHatRepositories/repositorySetRepositories.js
@@ -1,3 +1,4 @@
+import _ from 'lodash';
 import Immutable from 'seamless-immutable';
 
 import {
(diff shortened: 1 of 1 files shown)


webpack/redux/reducers/RedHatRepositories/repositorySetRepositories.js
```diff
@@ -12,9 +13,23 @@ import {
...
 
+const normalizeArch = (arch) => {
+  if (arch === 'noarch') {
+    return undefined;
+  }
+  return arch;
+};
+
+const substitutionMatches = (value1, value2) => {
+  if (_.isEmpty(value1) && _.isEmpty(value2)) {
```
parthaa: Cant you do 
```((_.isEmpty(value1) && _.isEmpty(value2)) || value1 == value2 ```

jlsherrill: i kinda like the more explicitness of what i have

PR #8491, opened 2020-01-07
Fixes #28659 - show correct subscriptions after delete

sseelam2:
This PR loads the correct subscriptions after a delete action is performed. Previously the correct subscriptions were loaded only after manually refreshing the page.

To check this PR:
1. Navigate to the Subscriptions tab.
2. Select few subscriptions and click on the delete button.
3. Subscriptions page should show subscriptions without the deleted subscriptions.


sseelam2:
[test katello]

sseelam2:
[test katello]

sseelam2:
[test katello]

jturel:
[test katello]

jturel:
I like how the feedback when pressing the 'Delete' button is immediate. It feels very responsive!

Although, there could be confusion in some cases: if the delete request fails there will be a notification stating that on the screen, but the sub(s) will be removed from the list. This could leave the user wondering if something went really bad.

Here's a potential way to work around that:

- Since the deletion fires an external task we could monitor that task for completion and *then* trigger the filter on the table (I think we do something like this on Red Hat Repositories) to omit the deleted subs.
- After confirming the Delete dialog we can show an additional notification on the screen that says 'The subscriptions are being deleted which can take a few minutes. The table will be refreshed automatically when complete' (or something to that effect)
- As soon as Delete is confirmed, the table is refreshed for no reason (this was there prior to your change). We may stop that reload altogether since there's no way the delete task will complete by the time the table reloads after confirmation. It's a minor detail but I think it would be nice to include here.

Let me know what you think (they are just suggestions). Would be wise to ask Rox as well since this is very much in the UX area.

sseelam2:
@jturel Thank you for the feedback. My understanding was that the subscriptions will be filtered only if the DELETE_SUBSCRIPTIONS_SUCCESS action is dispatched because I'm setting the filtered subscriptions as response only for this action.  So if there is an error, DELETE_SUBSCRIPTIONS_SUCCESS will not be dispatched and the subscriptions will not be removed. Let me know if I'm missing something.

sseelam2:
@jturel But that being said, if I do two deletes immediately one after the other I see this error - 
"Required lock is already taken by other running tasks. Please inspect their state, fix their errors and resume them. Required lock: read Conflicts with tasks:". I don't think it's letting me delete because the previous delete is still in progress. So as you suggested, monitoring the delete task for completion and then filtering will do the trick.

jturel:
> @jturel But that being said, if I do two deletes immediately one after the other I see this error -
> "Required lock is already taken by other running tasks. Please inspect their state, fix their errors and resume them. Required lock: read Conflicts with tasks:". I don't think it's letting me delete because the previous delete is still in progress. So as you suggested, monitoring the delete task for completion and then filtering will do the trick.

You may want to ask Walden or Jeremy for pointers in this area if there isn't an obvious example

sseelam2:
@jturel PR is ready for a retest.

sseelam2:
[test katello]

jturel:
@sseelam2 This is definitely a great solution. Now we don't need to worry about any special messaging!

I have one enhancement request for the functionality, and one issue I found.

The enhancement: when I confirm the delete the table is refreshed despite the task being in progress. There's no need to refresh the table, and I think we can safely stop that from happening. I noticed that this refresh doesn't happen when importing or refreshing a manifest, so that would be consistent too.

The issue: sometimes, but not all the time, once the task completes the page reloads with an error and is blank. Here's a screenshot of my console:

subscriptions_failure

webpack/scenes/Subscriptions/SubscriptionActions.js
```diff
@@ -126,7 +127,9 @@ export const deleteSubscriptions = poolIds => async (dispatch) => {
   };
 
   try {
+    const POLL_TASK_INTERVAL = 5000;
```
jturel: Is it possible / does it make sense to move this constant into TaskActions so we can use it in multiple places?

jturel: Also looks like SubcriptionsPage.js has the same constant defined, so it could potentially be reused there too

webpack/scenes/Subscriptions/SubscriptionsPage.js
```diff
@@ -41,14 +41,14 @@ class SubscriptionsPage extends Component {
     const [task] = tasks;
 
     if (numberOfTasks > 0) {
-      if (currentOrg === task.input.organization.id) {
+      if (currentOrg === task.input.current_organization_id) {
```
jturel: I guess the delete subscriptions action input doesn't have the organization? Wondering if this was actually necessary to change

webpack/scenes/Subscriptions/SubscriptionActions.js
sseelam2: I am putting it in `TaskConstants.js ` because this is being used by manifest tasks and also by subscription task.

webpack/scenes/Subscriptions/SubscriptionsPage.js
sseelam2: > I guess the delete subscriptions action input doesn't have the organization? Wondering if this was actually necessary to change

Yes it does not have organization. It threw an error because it couldn't find organization inside input.

jturel: Cool. Fortunately it looks like that doesn't break anything (manifest import and refresh, I think the only two other task-related actions on the page)

sseelam2:
> @sseelam2 This is definitely a great solution. Now we don't need to worry about any special messaging!
> 
> I have one enhancement request for the functionality, and one issue I found.
> 
> The enhancement: when I confirm the delete the table is refreshed despite the task being in progress. There's no need to refresh the table, and I think we can safely stop that from happening. I noticed that this refresh doesn't happen when importing or refreshing a manifest, so that would be consistent too.
> 
> The issue: sometimes, but not all the time, once the task completes the page reloads with an error and is blank. Here's a screenshot of my console:
> 
> subscriptions_failure

I did not set pollingATask to true when starting the delete subscription task. I believe setting this should fix the issue.

webpack/scenes/Subscriptions/SubscriptionsPage.js
```diff
@@ -139,19 +138,19 @@ class SubscriptionsPage extends Component {
...
 
   deleteManifest = async () => {
-    this.startManifestTask();
+    this.startTask();
     await this.props.deleteManifest();
     this.cleanUpManifestTask();
   };
 
   refreshManifest = async () => {
-    this.startManifestTask();
```
jturel: Deleting subscriptions is a manifest task. What were your thoughts when renaming this method?

sseelam2: Oh yes! cause we are deleting subscriptions from a manifest. My bad. I renamed it because we are deleting a subscription and not doing any major task on the manifest as a whole. But deleting a subscription changes the manifest. So startManifestTask() makes sense to me. I will rename it back.

jturel:
@swetha this looks great and now works consistently with manifest import and refresh!

I noticed one last issue which I think is on master but might be worth addressing on this PR: After deleting subscriptions and the table is reloaded the 'Delete' button is still active as if I have rows selected. In fact, clicking delete will tell me I have some number of them selected, but they are not (they were deleted). This will cause problems when trying to delet more subscriptions. Can you see how to clear the selection / disable the delete button after performing a delete?

webpack/scenes/Subscriptions/SubscriptionReducer.js
```diff
@@ -50,7 +49,6 @@ export default (state = initialState, action) => {
   switch (action.type) {
     case SUBSCRIPTIONS_REQUEST:
     case UPDATE_QUANTITY_REQUEST:
-    case DELETE_SUBSCRIPTIONS_REQUEST:
```
jturel: Also in this file is:    

```
case DELETE_SUBSCRIPTIONS_SUCCESS:
      return state.set('loading', false);
```

I *think* this is no longer required

sseelam2: Yes. we don't need to set `loading` to `false` anymore. I am now using `DELETE_SUBSCRIPTIONS_SUCCESS` to set `deleteButtonDisabled` to `true`.

sseelam2:
I am seeing this error `Uncaught (in promise) undefined` in my console now. Let me know if you guys @jturel @jeremylenz  see it as well. The error is not very informative. I'm trying to find a solution for it.

sseelam2:
> @swetha this looks great and now works consistently with manifest import and refresh!
> 
> I noticed one last issue which I think is on master but might be worth addressing on this PR: After deleting subscriptions and the table is reloaded the 'Delete' button is still active as if I have rows selected. In fact, clicking delete will tell me I have some number of them selected, but they are not (they were deleted). This will cause problems when trying to delet more subscriptions. Can you see how to clear the selection / disable the delete button after performing a delete?

made the required changes.

webpack/scenes/Subscriptions/SubscriptionReducer.js
sseelam2: Seems like we need to set loading to false because loading is set to true in the initial state. A test case failed. So I added this - `state.set('loading', false) `  back.

webpack/scenes/Subscriptions/SubscriptionReducer.js
```diff
@@ -91,7 +89,9 @@ export default (state = initialState, action) => {
     }
 
     case DELETE_SUBSCRIPTIONS_SUCCESS:
-      return state.set('loading', false);
+      return state
+        .set('deleteButtonDisabled', true)
```
jturel: In addition to disabling the button we need to clear whatever is keeping track of the selection.

After I delete 1 subscription and it completes, I select another 1, the modal says this:

```
Are you sure you want to delete 2 subscription(s)? This action will remove the subscription(s) and refresh your manifest. All systems using these subscription(s) will lose them and also may lose access to updates and Errata.
```

It should say 1 subscription

jturel: @sseelam2 have you had a chance to look into this?

sseelam2: @jturel I am working on it. I am looking into how to change the state of `selectedRows` because that is keeping track of the selection and it is not getting updated once the delete action is complete.

sseelam2: Used refs to change the state of the child component from its parent!

jeremylenz: @sseelam2 I have to give you props for your unorthodox methods, which I definitely would not have thought of.  But look at what the React folks [say here](https://reactjs.org/docs/refs-and-the-dom.html#when-to-use-refs):

> Your first inclination may be to use refs to “make things happen” in your app. If this is the case, take a moment and think more critically about where state should be owned in the component hierarchy. Often, it becomes clear that the proper place to “own” that state is at a higher level in the hierarchy. See the [Lifting State Up](https://reactjs.org/docs/lifting-state-up.html) guide for examples of this.

I'd have to hear a very compelling argument for not using state or Redux to solve this.

sseelam2: Ah. I thought it was safe to use refs. Updated the PR. Lifted the state up :+1:

webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTable.js
```diff
@@ -78,40 +77,33 @@ class SubscriptionsTable extends Component {
...
     return ({
       allRowsSelected: () => checkAllRowsSelected(),
       selectAllRows: () => {
         if (checkAllRowsSelected()) {
-          this.setState(
-            { selectedRows: [] },
-            updateDeleteButton,
-          );
+          this.props.onSelectedRowsChange([]);
+          this.setState(updateDeleteButton);
```
jeremylenz: It looks like this is passing the `updateDeleteButton` function to `this.setState`.  Since `updateDeleteButton` just invokes `this.props.toggleDeleteButton()`, and doesn't return an object describing the new state, this `setState` call doesn't make sense to me.  Can probably change it to
```js
if (checkAllRowsSelected()) {
  this.props.onSelectedRowsChange([]);
  this.props.toggleDeleteButton(false);  // will always be false because we just deselected all rows
  ...
```

webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTable.js
```diff
@@ -78,40 +77,33 @@ class SubscriptionsTable extends Component {
...
         } else {
-          this.setState(
-            { selectedRows: allSubscriptionResults.map(row => row.id) },
-            updateDeleteButton,
-          );
+          this.props.onSelectedRowsChange(allSubscriptionResults.map(row => row.id));
+          this.setState(updateDeleteButton);
```
jeremylenz: Same comment for this one, except pass `true`

webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTable.js
jeremylenz: Explanation of ^^

In the previous code, `updateDeleteButton` was the 2nd argument to `this.setState`, meaning it's a callback that React runs after the state update is completed.  But in the new code, it's the 1st argument, meaning it's expected to be an "[updater](https://reactjs.org/docs/react-component.html#setstate)" function that returns the new state.

webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTable.js
```diff
@@ -305,6 +297,8 @@ SubscriptionsTable.propTypes = {
       displayName: PropTypes.string,
     }),
   }),
+  selectedRows: PropTypes.array.isRequired,
```
jeremylenz: I think eslint might make you change this to `PropTypes.shape([]).isRequired`

webpack/scenes/Subscriptions/SubscriptionsPage.js
```diff
@@ -201,6 +213,7 @@ class SubscriptionsPage extends Component {
     });
 
     const onDeleteSubscriptions = (selectedRows) => {
+      this.startManifestTask();
       this.props.deleteSubscriptions(selectedRows);
```
jeremylenz: I don't see anywhere `selectedRows` is updated after deleting.  Shouldn't `selectedRows` be cleared out after deleting?

jturel:
I can give this another test today but looks like there are some JS lints. You can verify everything on your side without pushing code if you like (I like it a lot since I don't have everything eslint expects memorized!):

```
npm test

npm run lint
```

jturel:
I'm seeing an error in the console when loading the page:

```
foreman-vendor.bundle-v4.0.6-development-2be37b10db4fa73cba5e.js:209314 Warning: Failed prop type: Invalid prop `selectedRows` of type `array` supplied to `SubscriptionsTable`, expected `object`.
    in SubscriptionsTable (created by SubscriptionsPage)
    in SubscriptionsPage (created by ConnectFunction)
    in ConnectFunction (created by Headers)
    in Headers (created by CheckOrg)
    in CheckOrg (created by ConnectFunction)
    in ConnectFunction (created by Context.Consumer)
    in Route
    in div
    in Unknown (created by Application)
    in Router (created by BrowserRouter)
    in BrowserRouter (created by Application)
    in Application (created by ConnectFunction)
    in ConnectFunction (created by I18nProviderWrapper(Connect(Application)))
    in IntlProvider (created by I18nProviderWrapper(Connect(Application)))
    in I18nProviderWrapper(Connect(Application)) (created by StoreProvider(I18nProviderWrapper(Connect(Application))))
    in Provider (created by StoreProvider(I18nProviderWrapper(Connect(Application))))
    in StoreProvider(I18nProviderWrapper(Connect(Application))) (created by DataProvider(StoreProvider(I18nProviderWrapper(Connect(Application)))))
    in DataProvider(StoreProvider(I18nProviderWrapper(Connect(Application))))

```

webpack/scenes/Subscriptions/SubscriptionsPage.js
sseelam2: Yes you are right. didn't push the right code. I have updated it now.

jturel:
@sseelam2 looking good, but I found a regressions with 'select all' checkbox and the Delete button:

- refresh page
- click 'Select All' checkbox (select all rows)
- Notice that 'Delete' button isn't enabled
- Click 'Select All' again (deselect all rows)
- 'Delete' button is enabled (should be disabled b/c no rows are selected)

webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTable.js
```diff
@@ -78,40 +77,34 @@ class SubscriptionsTable extends Component {
...
         } else {
           selectedRows.push(rowData.id);
         }
-
-        this.setState(
-          { selectedRows },
-          updateDeleteButton,
-        );
+        this.props.onSelectedRowsChange(selectedRows);
+        if (selectedRows.length > 0) {
```
jturel: this if/else can be rewritten I think:

```
this.props.toggleDeleteButton(selectedRows.length > 0);
```

webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTable.js
```diff
@@ -78,40 +77,34 @@ class SubscriptionsTable extends Component {
...
+          this.props.toggleDeleteButton(true);
         }
       },
       selectRow: ({ rowData }) => {
-        let { selectedRows } = this.state;
+        let { selectedRows } = this.props;
         if (selectedRows.includes(rowData.id)) {
           selectedRows = selectedRows.filter(e => e !== rowData.id);
         } else {
           selectedRows.push(rowData.id);
```
jturel: The bug we talked about on IRC:

```
Uncaught ImmutableError {name: "MyError", message: "The push method cannot be invoked on an Immutable data structure.", stack: "Error↵    at new ImmutableError (https://centos7-k…0.8-development-2be37b10db4fa73cba5e.js:236922:5)"}
```

We can fix it like this, I think. Try it out by replacing the .push with:

```
selectedRows = selectedRows.concat(rowData.id);
```

PR #8495, opened 2020-01-07
Fixes #28676 - Single Candlepin call to create activation key

jturel:
We can now create an activation key with service level and release version without a separate update call to Candlepin. This PR cleans up the code in order to do that! The related BZ mentions that creating a key without an `auto_attach` value will result in a null value on the resulting key even though it should be `true`. This is fixed as well.
 
To test:

Create an activation key through API via POST to `/katello/api/v2/activation_keys`:
```
{
    "organization_id": "1",
    "name": "testkey",
    "service_level": "Self-support",
    "auto_attach": false,
    "release_version": "7Server"
}
```

The important aspect is that these values reach the key in Candlepin. Verify with rails console:

```
Katello::Resources::Candlepin::ActivationKey.get(Katello::ActivationKey.last.cp_id)
```

Check the releaseVer, autoAttach, and serviceLevel keys in particular. If no auto-attach value is specified in the API call then the default should be `true`.

### DIFF
--- a/test/actions/katello/activation_key_test.rb
+++ b/test/actions/katello/activation_key_test.rb
@@ -16,8 +16,10 @@ module ::Actions::Katello::ActivationKey
     let(:candlepin_input) do
       {  :organization_label => activation_key.organization.label,
          :auto_attach => true,
-         :purpose_usage => katello_activation_keys(:purpose_attributes_key).purpose_usage,
-         :purpose_role => katello_activation_keys(:purpose_attributes_key).purpose_role,
+         :service_level => 'Self-support',
+         :release_version => activation_key.release_version,
+         :purpose_usage => activation_key.purpose_usage,
+         :purpose_role => activation_key.purpose_role,
          :purpose_addons => [katello_purpose_addons(:addon).name]
       }
     end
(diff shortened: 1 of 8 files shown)


jturel:
[test katello]

jturel:
[test katello]

test/actions/katello/activation_key_test.rb
```diff
@@ -16,16 +16,18 @@ class CreateTest < TestBase
     let(:candlepin_input) do
       {  :organization_label => activation_key.organization.label,
          :auto_attach => true,
-         :purpose_usage => katello_activation_keys(:purpose_attributes_key).purpose_usage,
-         :purpose_role => katello_activation_keys(:purpose_attributes_key).purpose_role,
+         :service_level => 'Self-support',
```
sseelam2: I am just curious. Why not add service_level as well to the fixtures? I see that you have added release version.

sseelam2:
@jturel The functionality seems to be working as specified. Tested creating an activation key, updating it and also checking if the default value for auto attach is true through api. :+1:

test/actions/katello/activation_key_test.rb
jturel: We don't store the service level on the Katello::ActivationKey (but it's there in candlepin). But maybe double check!

jturel:
[test katello]

jturel:
[test katello]

jturel:
Thank you @sseelam2 and @parthaa !

jturel:
The build is green: https://ci.theforeman.org/blue/organizations/jenkins/katello-pr-test/detail/katello-pr-test/2689/pipeline

But the wrong one is linked to the PR since it was stuck and later cancelled. I'm merging based on this.

jturel:
Oh, it's blocking me!

[test katello]

PR #8497, opened 2020-01-08
Fixes #28535 - fix product task results loading

jjeffers:
This PR changes the foreman-tasks version to fix an issue with NULL ids returned in the results. Currently this version is the earliest that the foreman-tasks plugin will include the fix.

The release version of foreman-tasks is not yet ready, so it's expected this PR won't pass tests until then.


jlsherrill:
@jjeffers i don't think this is actually needed, if the fix lands in foreman-tasks only, then that fix is sufficient.

PR #8498, opened 2020-01-08
Fixes #28556 - Parse Oracle Linux facts properly

jturel:
Fixing per issue reported here: https://community.theforeman.org/t/task-actions-update-results-in-error/16665/4

### DIFF
--- a/app/services/katello/candlepin/consumer.rb
+++ b/app/services/katello/candlepin/consumer.rb
@@ -152,6 +152,10 @@ module Katello
           'Debian'
         elsif name =~ /ubuntu/
           'Ubuntu'
+        elsif name =~ /oracle/
+          'OracleLinux'
+        else
+          'Unknown'
         end
       end
     end
--- a/app/models/katello/rhsm_fact_parser.rb
+++ b/app/models/katello/rhsm_fact_parser.rb
@@ -66,7 +66,7 @@ module Katello
     end
 
     def os_release_name(os_name)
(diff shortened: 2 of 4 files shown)


jturel:
@jlsherrill any concerns with this? If so I'd like to see if the reporter is willing to test it

jlsherrill:
Looks good to me.  I'd be fine ACKin'g based on user testing

pa-jberanek:
I stumbled across this and fixed it myself, for reference:

https://github.com/Katello/katello/pull/8501

app/services/katello/candlepin/consumer.rb
```diff
@@ -152,6 +152,10 @@ def self.distribution_to_puppet_os(name)
           'Debian'
         elsif name =~ /ubuntu/
           'Ubuntu'
+        elsif name =~ /oracle/
+          'Oracle'
```
jlsherrill: I noticed that the other PR uses 'OracleLinux' and yours uses 'Oracle', what is the proper puppet OS name?

jturel: Looks like OracleLinux is the correct one

https://github.com/puppetlabs/facter/blob/79e5c272cea78dba2fbfac668335529630f79716/lib/inc/facter/facts/os.hpp#L55

jturel:
[test katello]

jturel:
This has been tested & verified by the reporter!

PR #8501, opened 2020-01-09
Katello::Candlepin::Consumer.distribution_to_puppet_os:

pa-jberanek:
Added detection of Oracle Linux, and added missing default case.

### DIFF
--- a/app/services/katello/candlepin/consumer.rb
+++ b/app/services/katello/candlepin/consumer.rb
@@ -144,6 +144,8 @@ module Katello
           'RedHat'
         elsif name =~ /centos/
           'CentOS'
+        elsif name =~ /oracle/
+          'OracleLinux'
         elsif name =~ /fedora/
           'Fedora'
         elsif name =~ /sles/ || name =~ /suse.*enterprise.*/
@@ -152,6 +154,8 @@ module Katello
           'Debian'
         elsif name =~ /ubuntu/
           'Ubuntu'
+        else
+          'Unknown'
         end
       end
     end


tbrisker:
ok to test

pa-jberanek:
I wonder if it should be "Oracle Linux" rather than "OracleLinux" though

evgeni:
> I wonder if it should be "Oracle Linux" rather than "OracleLinux" though

No, because OS families can't contain spaces.

pa-jberanek:
> > I wonder if it should be "Oracle Linux" rather than "OracleLinux" though
> 
> No, because OS families can't contain spaces.

Ah, OK.

By the way this out of the blue PR came from me finding that "subscription-manager register" started returning error 500 when run from Oracle Linux systems.

pa-jberanek:
I also wonder if this is the area of code (or app/models/katello/rhsm_fact_parser.rb) where the issue of the OS release not being shown in Katello for CentOS 7 could be fixed once and for all...

pa-jberanek:
OK, that'll be a "no"...

```
$ sudo subscription-manager facts |grep -i version
distribution.version: 7
```

jturel:
I have a PR opened for this since yesterday https://github.com/Katello/katello/pull/8498

Do you mind if we take mine over yours since I already have test coverage, redmine issue, etc?

pa-jberanek:
> I have a PR opened for this since yesterday #8498
> 
> Do you mind if we take mine over yours since I already have test coverage, redmine issue, etc?

No worries at all, but yours is missing the 'else' clause so still open to causing the error encountered for Oracle Linux...

jturel:
> > I have a PR opened for this since yesterday #8498
> > Do you mind if we take mine over yours since I already have test coverage, redmine issue, etc?
> 
> No worries at all, but yours is missing the 'else' clause so still open to causing the error encountered for Oracle Linux...

Do you have a stack trace? Also, can you provide the output of `subscription-manager facts | grep distribution` from the oracle client? Sounds like you're running into something slightly different than what my PR is addressing, but I can adjust it.

pa-jberanek:
> > > I have a PR opened for this since yesterday #8498
> > > Do you mind if we take mine over yours since I already have test coverage, redmine issue, etc?
> > 
> > 
> > No worries at all, but yours is missing the 'else' clause so still open to causing the error encountered for Oracle Linux...
> 
> Do you have a stack trace? Also, can you provide the output of `subscription-manager facts | grep distribution` from the oracle client? Sounds like you're running into something slightly different than what my PR is addressing, but I can adjust it.

Sorry, I mean the **possibility** of the issue for a **different** distribution. The backtrace occurred because the function returned null instead of a string in the case of an unrecognised distribution.

The backtrace and facts are very similar - for an 8.1 server:

```
distribution.id: 8.1
distribution.name: Oracle Linux Server
distribution.version: 8.1
distribution.version.modifier: 1
```

jturel:
> Sorry, I mean the **possibility** of the issue for a **different** distribution. The backtrace occurred because the function returned null instead of a string in the case of an unrecognised distribution.
> 
> The backtrace and facts are very similar - for an 8.1 server:
> 
> ```
> distribution.id: 8.1
> distribution.name: Oracle Linux Server
> distribution.version: 8.1
> distribution.version.modifier: 1
> ```

Understood. Updated my PR accordingly. Thank you!

jeremylenz:
[test katello]

jturel:
Closing in favor of #8498

PR #8506, opened 2020-01-10
Fixes #28715 - Correct permission for reports

ranjan:


sjha4:
[test katello]

sjha4:
[test katello]

sjha4:
@ranjan : Could you test this with the permission view_tasks from miscellaneous list? Would be interesting to see if that work today?

ranjan:
@sjha4, No it doesn't work for view-tasks(miscellaneous)

PR #8507, opened 2020-01-10
Fixes #28714 - Limit the certificate end date to late 2049

hao-yu:
Subscription-manager (python-rhsm) can't read the certificate
with end date deyond 2049 year correctly. It will read the date
as 1949 year which causing the pool of the custom products not
accessible.

### DIFF
--- a/app/lib/katello/resources/candlepin/product.rb
+++ b/app/lib/katello/resources/candlepin/product.rb
@@ -72,8 +72,13 @@ module Katello
 
           def create_unlimited_subscription(owner_key, product_id, start_date)
             start_date ||= Time.now
-            # End it 100 years from now
-            end_date ||= start_date + 10_950.days
+
+            # Subscription-manager (python-rhsm) can't read the certificate with end date deyond
+            # 2049 year correctly. Refer the links below for more details:
+            # https://bugzilla.redhat.com/show_bug.cgi?id=1789654
+            # https://github.com/candlepin/candlepin/blob/5b87865f304555c112982af4fbc83a1c463d37b2
+            # /server/src/main/java/org/candlepin/model/UeberCertificateGenerator.java#L247
+            end_date = Time.parse('2049-12-01 00:00:00 +0000')
 
             pool = {
               'startDate' => start_date,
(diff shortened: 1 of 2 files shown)


evgeni:
ok to test

evgeni:
This fails `test_scenarios – Scenarios::RepositoryCreateTest` because the VCR tape doesn't match anymore :(

app/lib/katello/resources/candlepin/product.rb
```diff
@@ -72,8 +72,13 @@ def remove_content(owner_label, product_id, content_id)
 
           def create_unlimited_subscription(owner_key, product_id, start_date)
             start_date ||= Time.now
-            # End it 100 years from now
-            end_date ||= start_date + 10_950.days
+
+            # Subscription-manager (python-rhsm) can't read the certificate with end date deyond
```
ekohl:
```suggestion
            # Subscription-manager (python-rhsm) can't read the certificate with end date beyond
```

app/lib/katello/resources/candlepin/product.rb
```diff
@@ -72,8 +72,13 @@ def remove_content(owner_label, product_id, content_id)
...
+            # 2049 year correctly. Refer the links below for more details:
+            # https://bugzilla.redhat.com/show_bug.cgi?id=1789654
```
ekohl: Most of the comments are private, such as linking to RFC 5280 that states:
> CAs conforming to this profile MUST always encode certificate validity dates through the year 2049 as UTCTime; certificate validity dates in 2050 or later MUST be encoded as GeneralizedTime.

app/lib/katello/resources/candlepin/product.rb
```diff
@@ -72,8 +72,13 @@ def remove_content(owner_label, product_id, content_id)
...
+            # https://github.com/candlepin/candlepin/blob/5b87865f304555c112982af4fbc83a1c463d37b2
+            # /server/src/main/java/org/candlepin/model/UeberCertificateGenerator.java#L247
```
ekohl: Not sure how useful it is to split a URL over multiple lines

app/lib/katello/resources/candlepin/product.rb
jlsherrill: i can make that comment not-private

jlsherrill:
@hao-yu thank you for your fix!   i've recorded the cassettes and included your fix as part of https://github.com/Katello/katello/pull/8508 with your commit.  I've also addressed a few of @ekohl comments there.  Closing this in favor of that

PR #8509, opened 2020-01-10
Fixes #28721 - Add a "Disabled" state for Subs

parthaa:
This commit adds a `Disabled` status for the golden ticketed
organizations. Its "disabled" to indicate that compliance calculations
will be ignored for this host.

### DIFF
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/subscriptions/views/content-access-mode-banner.html
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/subscriptions/views/content-access-mode-banner.html
@@ -1,5 +1,5 @@
 <div bst-alert="info" ng-show="contentAccessMode === 'org_environment'">
   <span translate>
-    Access to repositories is unrestricted in this organization. Hosts can consume all repositories available in the Content View they are registered to, regardless of subscription status.
+    This organization has Simple Content Access enabled. Access to repositories is unrestricted in this organization. Hosts can consume all repositories available in the Content View they are registered to, regardless of subscription status.
   </span>
 </div>
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/subscriptions/views/content-access-mode-banner.html
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/subscriptions/views/content-access-mode-banner.html
@@ -1,5 +1,5 @@
 <div bst-alert="info" ng-show="contentAccessMode === 'org_environment'">
   <span translate>
-    This organization has Simple Content Access enabled. Access to repositories is unrestricted in this organization. Hosts can consume all repositories available in the Content View they are registered to, regardless of subscription status.
+This organization has Simple Content Access enabled. Hosts can consume from all repositories in their Content View regardless of subscription status.
   </span>
 </div>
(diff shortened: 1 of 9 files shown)


sseelam2:
@parthaa I created an org for golden ticket and imported a manifest which is golden ticket enabled and created a host in this org. I see the subscription status on this host as "disabled". I think it is working as expected. 
Question - So do all hosts with simple content access have a subscription status as disabled by default?

sseelam2:
@parthaa Should we also show "disabled" in the Host Subscription Status under the Dashboard?

parthaa:
> Question - do all hosts with simple content access have a subscription status as disabled by default?

Yup. As long as golden ticket is enabled that should be the case.

parthaa:
> @parthaa Should we also show "disabled" in the Host Subscription Status under the Dashboard?

Good call. will look into it

jturel:
One thing I'm curious about. If I have an ordinary manifest imported, and my host is properly subscribed (green) and then I import a golden ticket manifest, do we get an event from Candlepin which updates the host status so that the host status will be in sync with the manifest?

parthaa:
Made the requisite dashboard changes for disabled subscription status

Dashboard for disabled sub status

Also changed the color for the disabled status from orange to green on the content hosts subscriptions page.

parthaa:
> One thing I'm curious about. If I have an ordinary manifest imported, and my host is properly subscribed (green) and then I import a golden ticket manifest, do we get an event from Candlepin which updates the host status so that the host status will be in sync with the manifest?

Not right now. Subscription status wont be changed until the system checks-in
Create an RFE here -> https://bugzilla.redhat.com/show_bug.cgi?id=1790994 for cp to provide us an event on golden ticket enablement.

engines/bastion_katello/app/assets/javascripts/bastion_katello/subscriptions/views/content-access-mode-banner.html
```diff
@@ -1,5 +1,5 @@
 <div bst-alert="info" ng-show="contentAccessMode === 'org_environment'">
   <span translate>
-    Access to repositories is unrestricted in this organization. Hosts can consume all repositories available in the Content View they are registered to, regardless of subscription status.
+    This organization has Simple Content Access enabled. Access to repositories is unrestricted in this organization. Hosts can consume all repositories available in the Content View they are registered to, regardless of subscription status.
```
jturel: Looking at it again, I think we can tighten this text up a little bit. Let me know what you think about:

`This organization has Simple Content Access enabled. Hosts can consume from all repositories in their Content View regardless of subscription status.`

I removed (imo) repetative 'this organization' and text around 'unrestricted' usage - because it *is* restricted to the content view.

jturel: Lemme know what you think

parthaa: Updated

jturel:
Looks good code-wise. I'll let @sseelam2 give it another test for functionality!

sseelam2:
@jturel @parthaa looks good to me :+1:

jturel:
[test katello]

PR #8510, opened 2020-01-10
Fixes #28657 - adds a rake task to copy migrated pulp3 href to pulp ids

jjeffers:
This PR introduces a rake task that copies the `migrated_pulp3_href` values into the `pulp_id` column for supported content types (file, docker).

### DIFF
--- a/app/services/katello/pulp3/migration.rb
+++ b/app/services/katello/pulp3/migration.rb
@@ -6,6 +6,11 @@ module Katello
       attr_accessor :smart_proxy
       GET_QUERY_ID_LENGTH = 90
 
+      REPOSITORY_TYPES = [
+        Katello::Repository::FILE_TYPE,
+        Katello::Repository::DOCKER_TYPE
+      ].freeze
+
       def initialize(smart_proxy, repository_type_labels)
         @smart_proxy = smart_proxy
         @repository_type_labels = repository_type_labels
@@ -25,6 +30,14 @@ module Katello
 
       def create_and_run_migration
         start_migration(create_migration)
+      end
+
+      def self.content_types_for_migration
+        content_types = REPOSITORY_TYPES.collect do |repository_type_label|
+          Katello::RepositoryTypeManager.repository_types[repository_type_label].content_types_to_index
+        end
+
+        content_types.flatten
       end
 
       def import_pulp3_content
--- a/lib/katello/tasks/pulp3_content_switchover.rake
+++ b/lib/katello/tasks/pulp3_content_switchover.rake
@@ -1,0 +1,21 @@
+require File.expand_path("../engine", File.dirname(__FILE__))
+
+namespace :katello do
+  desc "Runs a Pulp 3 migration of pulp3 hrefs to pulp ids for supported content types."
+  task :pulp3_content_switchover => :environment do
+    content_types = Katello::Pulp3::Migration.content_types_for_migration
+
+    content_types.each do |content_type|
+      if content_type.model_class.where(migrated_pulp3_href: nil).any?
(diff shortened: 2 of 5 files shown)


lib/katello/tasks/pulp3_content_switchover.rake
```diff
@@ -0,0 +1,19 @@
...
+
+    content_models.each do |model|
+      if model.where(migrated_pulp3_href: nil).any?
+        $stderr.print("ERROR: at least one #{model.table_name} record has migrated_pulp3_href NULL value")
+        exit 1
+      end
+    end
+
+    content_models.each do |model|
+      model.update_all("pulp_id = migrated_pulp3_href")
```
sjha4: Wonder if something like `model.where.not("pulp_id=migrated_pulp3_href").update_all("pulp_id=migrated_pulp3_href")` will make this more efficient on re-runs or cause memory overloads in case of large number of records like packages?

jjeffers: A good point. I wasn't sure about every context it would be run in, and I don't see a reason to not add this clause. Unless @jlsherrill has an objection, I'll add it. Thanks!

jlsherrill: that seems reasonable!

app/services/katello/pulp3/migration.rb
```diff
@@ -6,6 +6,13 @@ class Migration
       attr_accessor :smart_proxy
       GET_QUERY_ID_LENGTH = 90
 
+      CONTENT_TYPES = [
+        Katello::FileUnit,
+        Katello::DockerManifestList,
+        Katello::DockerManifest,
+        Katello::DockerTag
+      ].freeze
```
jlsherrill: i'd really like there to be a single 'definition' for the content types in pulp we are migrating.  Currently the full migration rake task defines it here:  
https://github.com/Katello/katello/blob/master/lib/katello/tasks/pulp3_migration.rake#L4

but i think it makes sense to move that here, add docker, and then derive the content types from the repository like this is doing:

https://github.com/Katello/katello/blob/master/app/services/katello/pulp3/migration.rb#L32

That way, all of these migration tasks have a single 'source' for what should be migrated

app/services/katello/pulp3/migration.rb
```diff
@@ -6,6 +6,11 @@ class Migration
       attr_accessor :smart_proxy
       GET_QUERY_ID_LENGTH = 90
 
+      REPOSITORY_TYPES = [
+        'file',
+        'docker'
+      ].freeze
```
jlsherrill: mind referencing the constants like Katello::Repository::FILE_TYPE ?

lib/katello/tasks/pulp3_content_switchover.rake
```diff
@@ -0,0 +1,19 @@
...
+
+namespace :katello do
+  desc "Runs a Pulp 3 migration of pulp3 hrefs to pulp ids for supported content types."
+  task :pulp3_content_switchover => :environment do
+    content_types = Katello::Pulp3::Migration.content_types_for_migration
+
+    content_types.each do |content_type|
+      if content_type.model_class.where(migrated_pulp3_href: nil).any?
+        $stderr.print("ERROR: at least one #{content_type.model_class.table_name} record has migrated_pulp3_href NULL value")
+        exit 1
```
jlsherrill: would it make sense to just raise an exception?

jjeffers: Previous rake tasks seemed to have used SystemExit. I don't have a strong opinion either way.

jlsherrill:
One small comment, but everything else looks good!  Should be good to merge after that

PR #8518, opened 2020-01-14
Fixes #28722 - Upgrade task to set modular in rpm

parthaa:
Pulp recently fixed a bug related to the the is_modular flag of a
package. https://pulp.plan.io/issues/5942

This commit adds a task to reindex the is_modular attribute of an
rpm based on the values in the pulp db store.

This upgrade script is meant to be run after
sudo -u apache /usr/bin/pulp-manage-db

### DIFF
--- a/lib/katello/tasks/upgrades/3.15/reindex_rpm_modular.rake
+++ b/lib/katello/tasks/upgrades/3.15/reindex_rpm_modular.rake
@@ -2,34 +2,21 @@ namespace :katello do
   namespace :upgrades do
     namespace '3.15' do
       desc "Reindex the is_modular attribute of an rpm"
-      def update_modular(pulp_response)
-        modular = []
-        non_modular = []
-        pulp_response.each do |rpm|
-          if rpm["is_modular"]
-            modular << rpm["_id"]
-          else
-            non_modular << rpm["_id"] unless rpm["is_modular"]
-          end
-        end
-        ::Katello::Rpm.where(pulp_id: modular).update_all(modular: true)
-        ::Katello::Rpm.where(pulp_id: non_modular).update_all(modular: false)
-      end
-
       task :reindex_rpm_modular, [:input_file] => ["environment"] do
         User.current = User.anonymous_api_admin
-        criteria = { :fields => ["is_modular"],
+        criteria = { :fields => [],
                      :limit => SETTINGS[:katello][:pulp][:bulk_load_size],
-                     :skip => 0
+                     :skip => 0,
+                     :filters => {'is_modular' => {'$eq' => true}}
                    }
         content_type = Katello.pulp_server.extensions.rpm.content_type
         batch = ::Katello::Pulp::PulpContentUnit.pulp_units_batch(criteria, criteria[:limit]) do
           Katello.pulp_server.resources.unit.search(content_type, criteria)
         end
-
-        batch.each do |pulp_response|
-          update_modular(pulp_response)
-        end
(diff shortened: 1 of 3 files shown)


parthaa:
To test this
* Sync centos8 appstream ->  http://mirror.linux.duke.edu/pub/centos/8/AppStream/x86_64/os
* Content-> Packages 
* Search for the "git" package. Click on the details
* Notice that the modular attribute is set to true (which is wrong because git does not belong to module)
* Now apply the pulp patch from -> https://github.com/pulp/pulp_rpm/pull/1579
* Stop the pulp services
```
for x in httpd pulp_resource_manager pulp_celerybeat pulp_workers
do
  sudo service $x stop
done
```
* run ```sudo -u apache /usr/bin/pulp-manage-db```
* restart pulp
```
for x in httpd pulp_resource_manager pulp_celerybeat pulp_workers
do
  sudo service $x restart
done
```
* Finally run the upgrade task listed in this PR
```bundle exec rake katello:upgrades:3.15:reindex_rpm_modular```
This should classify the modular and non modular rpms.

Look at the git package again in the UI and you should see that modular is correctly set to false

File: lib/katello/tasks/upgrades/3.15/reindex_rpm_modular.rake [hunk 1]
```diff
@@ -0,0 +1,22 @@
+namespace :katello do
+  namespace :upgrades do
+    namespace '3.15' do
+      desc "Reindex the is_modular attribute of an rpm"
+      task :reindex_rpm_modular, [:input_file] => ["environment"] do
+        User.current = User.anonymous_api_admin
+        rpms = Katello.pulp_server.resources.unit.search(Katello.pulp_server.extensions.rpm.content_type, :fields => ["is_modular"])
```
jlsherrill: I think this needs to be paged, what would happen if the user had ~100K rpms?

lib/katello/tasks/upgrades/3.15/reindex_rpm_modular.rake
```diff
@@ -0,0 +1,36 @@
...
+        end
+        ::Katello::Rpm.where(pulp_id: modular).update_all(modular: true)
+        ::Katello::Rpm.where(pulp_id: non_modular).update_all(modular: false)
+      end
+
+      task :reindex_rpm_modular, [:input_file] => ["environment"] do
+        User.current = User.anonymous_api_admin
+        criteria = { :fields => ["is_modular"],
+                     :limit => SETTINGS[:katello][:pulp][:bulk_load_size],
+                     :skip => 0
```
jlsherrill: i think it might be worth it to limit to ONLY those rpms that have is_modular set to true.  That way we'd likely cut down the migration time by like ~90%, or even more depending on if the user has rhel 8 or centos 8 synced

parthaa: > i think it might be worth it to limit to ONLY those rpms that have is_modular set to true. That way we'd likely cut down the migration time by like ~90%, or even more depending on if the user has rhel 8 or centos 8 synced

updated

parthaa:
> I'm still looking through the code but I've tested this and it works!

can ya try again :)

chhith:
Hello, I can't find those 2 files modularity.py and 0048_correct_is_modular_field.py on my pulp server.
The version of pulp is 2.16.1. Do I have to update the version ? Or Can I just create those both files ?

Kinds

jlsherrill:
Hello, the pulp fix for this has not been released yet:  https://pulp.plan.io/issues/5942

PR #8520, opened 2020-01-15
Fixes #28751 - Allow admin users to access subscriptions from all organizations

jeremylenz:
If the current user is an admin, don't check if the requested subscription matches the current user & organization.

### DIFF
--- a/app/controllers/katello/api/v2/subscriptions_controller.rb
+++ b/app/controllers/katello/api/v2/subscriptions_controller.rb
@@ -75,10 +75,9 @@ module Katello
     param :organization_id, :number, :desc => N_("Organization identifier")
     param :id, :number, :desc => N_("Subscription identifier"), :required => true
     def show
-      org_id = Organization.current&.id
       @resource = Katello::Pool.with_identifier(params[:id])
       fail ActiveRecord::RecordNotFound, N_('Subscription not found') unless @resource
-      if @resource.organization_id != org_id && !User.current.organizations&.pluck(:id)&.include?(@resource.organization_id)
+      unless @resource.readable?
         fail ActiveRecord::RecordNotFound, N_('This subscription is not relevant to the current user and organization.')
       end
       respond(:resource => @resource)
--- a/app/models/katello/authorization/subscription.rb
+++ b/app/models/katello/authorization/subscription.rb
@@ -10,7 +10,9 @@ module Katello
 
     module ClassMethods
       def readable
-        authorized(:view_subscriptions)
+        return all if User.current.admin?
+
+        authorized(:view_subscriptions).where(organization_id: User.current.organization_ids)
       end
     end
   end
--- a/app/models/katello/authorization/pool.rb
+++ b/app/models/katello/authorization/pool.rb
@@ -2,9 +2,13 @@ module Katello
   module Authorization::Pool
     extend ActiveSupport::Concern
 
+    def readable?
+      self.class.readable.where(id: self.id).any?
+    end
+
     module ClassMethods
       def readable
-        where(:subscription_id => Katello::Subscription.authorized(:view_subscriptions))
+        where(:subscription_id => Katello::Subscription.readable)
       end
     end
   end
--- a/test/controllers/api/v2/subscriptions_controller_test.rb
+++ b/test/controllers/api/v2/subscriptions_controller_test.rb
@@ -95,9 +95,23 @@ module Katello
       assert_template 'api/v2/subscriptions/show'
     end
 
-    def test_show_without_user_org_id
+    def test_show_admin_user_can_access_subs_from_different_org
+      @subscription = katello_pools(:pool_one)
+      get :show, params: { :id => @subscription.id }
+
+      assert_response :success
+      assert_template 'api/v2/subscriptions/show'
+    end
+
+    def test_show_non_admin_cannot_access_subs_from_different_org
       @subscription = katello_pools(:pool_one)
       @org = get_organization(:empty_organization)
+      # use a non-admin user with view_subscriptions permission
+      non_admin_user = User.find(users(:one).id)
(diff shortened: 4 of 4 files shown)


sjha4:
[test katello]

jeremylenz:
Discussion on the related BZ:

> I'm able to reproduce the issue when using the endpoint GET subscriptions/:id.
>
> However, using GET organizations/:id/subscriptions/:id may work around the issue.  (that's what the web UI does.)
>
> Which API endpoint are you using to make the request?  Would using the endpoint above be a viable workaround?

Jitendra responded:

_Hey Jeremy,_

_Yes, I am using the same endpoint with which you could repro the issue: `katello/api/v2/subscriptions`_

_I think the workaround is viable and we can all-together remove the `katello/api/v2/subscriptions` that has nothing to do with org._

@jturel @sjha4 If we were to remove that endpoint, it would require more changes to the front end and might break other things...  I'm gravitating toward merging this as-is and keeping the endpoints the same.  Thoughts?

sjha4:
@jeremylenz : I think the scope of this PR is allowing admin access to subscriptions via the subscriptions/:id endpoint when they have access to the same using org/:id/sub/:id endpoint. If we agree on the validity of the requirement, I don't see issues in merging this.

app/controllers/katello/api/v2/subscriptions_controller.rb
```diff
@@ -200,6 +199,11 @@ def available_for_activation_key
       @activation_key.available_subscriptions
     end
 
+    def subscription_not_relevant?
```
jturel: There are a *lot* of negations going on in this code and it's hurting my brain. 💀 

One suggestion is to frame this method positively ie `subscription_relevant?` and rework the logic accordingly

However, I think it could be clarified significantly doing something like this:

```

def permitted_org_ids
  [User.current.organizations&.pluck(:id), Organization.current&.id].flatten.uniq.compact
end


unless User.current.admin? || !permitted_org_ids.include?(@resource.organization_id)
  # raise error
end
```

Now I can read the logic the same way I would describe the intended behavior: unless the user is an admin or subscription's org is permitted then raise an error.

The conditional could even be written to avoid the negation: `unless User.current.admin || permitted_org_ids.exclude?(@resource.organization_id)`

jturel: I was thinking about this some more and thought there could be a way to use our permissions system to handle this rather than doing this check in the controller. Your controller tests pass with this:

controller:

```
unless @resource.readable?
  # raise error
end
```

app/models/katello/authorization/subscription.rb
```
module ClassMethods
  def readable
    return all if User.current.admin?

    authorized(:view_subscriptions).where(organization_id: User.current.organization_ids)
  end
end
```

app/models/katello/authorization/pool.rb
```
    def readable?
      self.class.readable.where(id: self.id).any?
    end

    module ClassMethods
      def readable
        where(:subscription_id => Katello::Subscription.readable)
      end
    end
```

We end up with more reusable code, and all the checks are done at the db via 3 queries:

```
Katello::Pool Load (0.3ms)  SELECT  "katello_pools".* FROM "katello_pools" WHERE (katello_pools.id = (128914721) or katello_pools.cp_id = ('128914721')) ORDER BY "katello_pools"."id" ASC LIMIT $1  [["LIMIT", 1]]

Filter Load (2.1ms)  SELECT DISTINCT "filters".* FROM "filters" INNER JOIN "filterings" ON "filterings"."filter_id" = "filters"."id" INNER JOIN "permissions" ON "permissions"."id" = "filterings"."permission_id" INNER JOIN "roles" ON "filters"."role_id" = "roles"."id" INNER JOIN "cached_user_roles" ON "roles"."id" = "cached_user_roles"."role_id" LEFT JOIN taxable_taxonomies ON (filters.id = taxable_taxonomies.taxable_id AND taxable_type = 'Filter') LEFT JOIN taxonomies ON (taxonomies.id = taxable_taxonomies.taxonomy_id) WHERE "cached_user_roles"."user_id" = $1 AND (permissions.resource_type = 'Katello::Subscription') AND (permissions.name = 'view_subscriptions') AND (taxable_taxonomies.id IS NULL OR (taxonomies.type = 'Organization') OR (taxonomies.type = 'Location'))  [["user_id", 980190962]]

Katello::Subscription Exists (0.3ms)  SELECT  1 AS one FROM "katello_subscriptions" WHERE "katello_subscriptions"."organization_id" = $1 AND "katello_subscriptions"."id" = $2 LIMIT $3  [["organization_id", 447626438], ["id", 904941504], ["LIMIT", 1]]
``` 

The code is organized so that viewing subscriptions (actually, they are Katello::Pool) use the Katello::Subscription permissions, at least for reading.

jeremylenz: > There are a _lot_ of negations going on in this code and it's hurting my brain. 
> ```ruby
> unless User.current.admin? || permitted_org_ids.include?(@resource.organization_id)
>   # raise error
> end
> ```
> 
> Now I can read the logic the same way I would describe the intended behavior: unless the user is an admin or subscription's org is permitted then raise an error.

I was thinking the same thing but couldn't think of a way to make it more clear with an `unless`!  jturel++

jeremylenz:
Updated with @jturel 2nd suggestion above

app/models/katello/authorization/subscription.rb
```diff
@@ -4,13 +4,11 @@ module Authorization::Subscription
 
     include Authorizable
 
-    def readable?
```
jturel: The tests are passing => this code my be dead (or untested). that said, I'd rather leave it in since it's not related to this change.

jeremylenz: @jturel ok, adding it back

jturel:
Merging, thanks guys!

PR #8522, opened 2020-01-16
Fixes #28777 - sync_summary no result in API

m-bucher:
From within the API-Controller `started_at` returns timezoned-format that does not work with the following `where`-statement.

### DIFF
--- a/app/models/katello/glue/pulp/repos.rb
+++ b/app/models/katello/glue/pulp/repos.rb
@@ -97,7 +97,7 @@ module Katello
       def last_repo_sync_task_group
         if last_repo_sync_task
           started_after = last_repo_sync_task.started_at - 30.seconds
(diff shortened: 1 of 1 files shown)


m-bucher:
@jlsherrill , when I cherry-picked this fix to Katello-3.12 our internal test found that the working `sync_summary`-attribute adds a lot of loading time to the product-index API call (and therefore the products overview-page).

Tested with 70-80 products with multiple repos each and a pagination of 100 products-per-page.
Resulting in the products-page taking approx. 5 minutes to load.

Can you confirm that for the nightly builds?
I do not like to fix something and at the same time kill the good user-experience :-(

m-bucher:
Looks like this is due to an change in https://github.com/Katello/katello/pull/8536, where I changed the method that supplies the data.
Gonna discuss this over there.

PR #8525, opened 2020-01-16
Fixes #28660 - post pulp3 migration rake task

jjeffers:
This PR introduces a post-migration check for pulp3 content. The task checks that pulp3 repositories have `version_href` and `remote_href` values. For non-archived repositories, the task also checks that a distribution reference exists.

### DIFF
--- a/lib/katello/tasks/pulp3_post_migration_check.rake
+++ b/lib/katello/tasks/pulp3_post_migration_check.rake
@@ -1,0 +1,32 @@
+require File.expand_path("../engine", File.dirname(__FILE__))
+
+namespace :katello do
+  desc "Runs a post Pulp3 migration check for supported content types."
+  task :pulp3_post_migration_check => :environment do
+    repository_types = Katello::Pulp3::Migration::REPOSITORY_TYPES
+
+    repository_types.each do |type|
+      filter = 'version_href is NULL OR remote_href is NULL'
+
+      unless type == Katello::Repository::DOCKER_TYPE
+        filter += ' OR publication_href is NULL'
+      end
+      repositories = Katello::Repository.with_type(type).where(filter)
+
+      if repositories.any?
+        $stderr.print("ERROR: #{type} repository #{repositories.first.id} has a NULL value for remote_href, version_href, publication_href\n")
+        exit 1
+      end
+
+      non_archived_repositories = Katello::Repository.with_type(type).non_archived
+      with_no_distribution_references = non_archived_repositories
+        .left_outer_joins(:distribution_references)
+        .where(katello_distribution_references: { id: nil })
+
+      if with_no_distribution_references.any?
+        $stderr.print("ERROR: A non-archive #{type} repository #{with_no_distribution_references.first.id} did not have a distribution reference\n")
+        exit 1
+      end
+    end
+  end
+end
--- a/test/lib/tasks/pulp3_post_migration_check_test.rb
+++ b/test/lib/tasks/pulp3_post_migration_check_test.rb
@@ -1,0 +1,153 @@
+require 'katello_test_helper'
+require 'rake_test_helper'
+require 'rake'
+
+module Katello
+  class PostPulp3MigrationCheckTest < ActiveSupport::TestCase
+    def setup
+      @fake_pulp3_href = 'fake_pulp3_href'
+      @another_fake_pulp3_href = 'another_fake_pulp3_href'
+
+      Katello::Pulp3::Migration::REPOSITORY_TYPES.each do |type|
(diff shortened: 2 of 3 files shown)


lib/katello/tasks/pulp3_post_migration_check.rake
```diff
@@ -0,0 +1,29 @@
...
+      end
+
+      repositories = Katello::Repository.with_type(type)
+        .where.not(remote_href: nil)
+        .where.not(version_href: nil)
+
+      non_archived_repositories = repositories.non_archived
+      if non_archived_repositories
+        .left_outer_joins(:distribution_references)
+        .where(katello_distribution_references: { id: nil }).any?
```
ianballou: Should this line be `.where(katello_distribution_references: { href: nil }).any?`

I'm wondering because we're checking hrefs for the other two fields (version_href, remote_href).

I did notice that `id` and `href` for the `katello_distribution_references` table are required to be non-null, so I'm not sure if this check is actually necessary.

ianballou: Never mind, I see what's happening here.

ianballou:
Should we print the offending repository IDs?  I imagine if I hit this error I would want some more info on how to fix.  Also, does a user necessarily know what a "non-archived" repository is?

jjeffers:
@ianballou Good suggestions, thank you. What is a better way to express `non-archive`?

ianballou:
@jjeffers well, actually if someone understands what a distribution reference is then I guess `non-archived` won't trip them up either.  Not sure how verbose we need to be here.

ianballou:
Oh and I tested this out and it works :+1:

I'll ACK after we figure out the best user output, everything else looks good.

jlsherrill:
I think publication_href also needs to be checked, apologies i forgot to mention it in the issue description!

lib/katello/tasks/pulp3_post_migration_check.rake
```diff
@@ -0,0 +1,30 @@
+require File.expand_path("../engine", File.dirname(__FILE__))
+
+namespace :katello do
+  desc "Runs a post Pulp3 migration check for supported content types."
+  task :pulp3_post_migration_check => :environment do
+    repository_types = Katello::Pulp3::Migration::REPOSITORY_TYPES
+
+    repository_types.each do |type|
+      repositories = Katello::Repository.with_type(type)
+        .where('version_href is NULL OR remote_href is NULL OR publication_href is NULL')
```
ianballou: Docker repositories will never have a non-nil `publication_href`.  We need to skip that check.

lib/katello/tasks/pulp3_post_migration_check.rake
```diff
@@ -0,0 +1,33 @@
...
+      if repositories.any?
+        $stderr.print("ERROR: at least one #{type} repository record has a NULL value for remote_href, version_href, publication_href\n")
```
ianballou: I'm still hitting this error, I think you need to add your Docker filter before it.

lib/katello/tasks/pulp3_post_migration_check.rake
```diff
@@ -0,0 +1,38 @@
...
+        filter += ' OR publication_href is NULL'
+      end
+      repositories = Katello::Repository.with_type(type).where(filter)
+
+      if repositories.any?
+        $stderr.print("ERROR: at least one #{type} repository record has a NULL value for remote_href, version_href, publication_href\n")
+        exit 1
+      end
+
+      repositories = Katello::Repository.with_type(type)
```
ianballou: Is it necessary to filter out the repos where remote, version, or publication href are `nil`?  If they are `nil`, I think the execution wouldn't make it this far due to the `exit 1` on line 18.

jjeffers: Fantastic catch!

lib/katello/tasks/pulp3_post_migration_check.rake
```diff
@@ -0,0 +1,30 @@
...
+      if repositories.any?
+        $stderr.print("ERROR: #{type} repository #{repositories.first.id} has a NULL value for remote_href, version_href, publication_href\n")
+        exit 1
+      end
+
+      non_archived_repositories = Katello::Repository.with_type(type).non_archived
+      if non_archived_repositories
+        .left_outer_joins(:distribution_references)
+        .where(katello_distribution_references: { id: nil }).any?
+        $stderr.print("ERROR: at least one non-archived #{type} repository did not have a distribution reference\n")
```
ianballou: Just need the repository id output here and I'm good with it!

jjeffers:
[test katello]

PR #8528, opened 2020-01-16
Fixes #28425 - Change service name to Foreman URL

chris1984:
This appears to be dead code:

https://github.com/Katello/katello/blob/d9f5d173cb8a604caae3ad20c3f237d06c0f846b/app/models/katello/glue/candlepin/repository.rb#L17-L22

Did not come up with pry and removing it all together did not impact anything and the redhat.repo file on the client is still updating correctly when cycling between two different gpg keys. Did not see any tracebacks in the rails server output and all api calls to /rhsm etc were 200.

I looked at the database and in `katello(katello_contents)` and `candlepin(cp2_content)` it looks like we store the GPG URL as such:

`../../katello/api/v2/repositories/1/gpg_key_content`

I was able to watch the database in real-time using Navicat, when doing a `rhsm refresh` on the client I could see the database GPG entries being changed so the above method is no longer in use.

Waiting on the 2nd part to hear more back from Ewoud and to see if there any test failures.

### DIFF
--- a/app/controllers/katello/api/rhsm/candlepin_proxies_controller.rb
+++ b/app/controllers/katello/api/rhsm/candlepin_proxies_controller.rb
@@ -266,7 +266,7 @@ module Katello
     def get_parent_host(headers)
       hostnames = headers["HTTP_X_FORWARDED_SERVER"]
       host = hostnames.split(",")[0] if hostnames
-      host || SETTINGS[:fqdn]
+      host || URI.parse(Setting[:foreman_url]).host
     end
 
     private
--- a/test/controllers/api/rhsm/candlepin_proxies_controller_test.rb
+++ b/test/controllers/api/rhsm/candlepin_proxies_controller_test.rb
@@ -395,6 +395,7 @@ module Katello
     describe "get parent host" do
       it "can get parent host" do
         capsule = "foocapsule.example.com"
+        Setting[:foreman_url] = 'https://foreman.example.com'
 
         host_and_capsule = {"HTTP_X_FORWARDED_SERVER" => "#{capsule}, foo.example.com"}
         just_capsule = {"HTTP_X_FORWARDED_SERVER" => "#{capsule}"}
@@ -402,7 +403,7 @@ module Katello
 
         assert_equal @controller.get_parent_host(host_and_capsule), "#{capsule}"
         assert_equal @controller.get_parent_host(just_capsule), "#{capsule}"
-        assert_equal SETTINGS[:fqdn], @controller.get_parent_host(nil_host)
+        assert_equal 'foreman.example.com', @controller.get_parent_host(nil_host)
       end
     end
   end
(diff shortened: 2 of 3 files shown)


app/controllers/katello/api/rhsm/candlepin_proxies_controller.rb
```diff
@@ -266,7 +266,7 @@ def serials
     def get_parent_host(headers)
       hostnames = headers["HTTP_X_FORWARDED_SERVER"]
       host = hostnames.split(",")[0] if hostnames
-      host || SETTINGS[:fqdn]
+      host || Setting[:foreman_url]
```
ekohl: URL is an URL, not a hostname.

jlsherrill:
@chris1984 actually the reason removing that method doesn't affect anything is that its defined a 2nd time in app/models/katello/repository.rb and thats taking precedent, and is the correct implementation :)

chris1984:
```shell
[root@client ~]# subscription-manager register --org="Default_Organization" --environment="Library"
Registering to: katello.katello.lan:443/rhsm
Username: admin
Password:
The system has been registered with ID: 4dedf224-b0d0-4154-a21a-2d1cce1094a9
The registered system name is: client.katello.lan
```
Shows registered through the katello server:

https://nimb.ws/OYsH7D

Installed a caspule and tried to register a client through the capsule and shows correct register through fqdn

```shell
[root@client ~]# subscription-manager register --org="Default_Organization" --environment="Library"
Registering to: capsule.katello.lan:8443/rhsm
Username: admin
Password:
The system has been registered with ID: 17db2950-1f0c-407a-a9ec-89b9e8c79c51
The registered system name is: client.katello.lan
```
Screenshot:

https://nimb.ws/a8WBqY

test/controllers/api/rhsm/candlepin_proxies_controller_test.rb
```diff
@@ -402,7 +402,7 @@ def test_regenerate_indentity_certificates
 
         assert_equal @controller.get_parent_host(host_and_capsule), "#{capsule}"
         assert_equal @controller.get_parent_host(just_capsule), "#{capsule}"
-        assert_equal SETTINGS[:fqdn], @controller.get_parent_host(nil_host)
+        assert_equal URI.parse(Setting[:foreman_url]).host, @controller.get_parent_host(nil_host)
```
ekohl: I'm never a fan of having the same code in tests as in the actual code. IMHO it'd be better to have setup code that ensures `Setting[:foreman_url]` is `https://foreman.example.com` and a static assertion:
```suggestion
        assert_equal 'foreman.example.com', @controller.get_parent_host(nil_host)
```

chris1984: @ekohl updated the test to reflect what you wanted to see which makes more sense how you explained it. Does this look ok?

PR #8531, opened 2020-01-17
Fixes #27647: enable true select all on HC errata

waldenraines:
Include true select all functionality on the host collection errata
modal.

https://projects.theforeman.org/issues/27647


waldenraines:
Requires https://github.com/Katello/katello/pull/8526

waldenraines:
Note that this only adds the ability to select all errata on the host collection errata modal.  It does not address any edge cases that are broken on all tables such as using select all, deselecting something from page one, going to page two and submitting the form.  It also doesn't add true bulk functionality via the bulk action includes/excludes.  Our entire table functionality (and potentially the bulk APIs) need a rework for that to be possible.

PR #8536, opened 2020-01-21
Fixes #28818 - Add aggregated sync-state

m-bucher:
This adds a method to aggregate the last sync-status for all repositories of a product (that a sync-task exists for) and:
- uses this data for `sync_summary`-method
- adds an additional attribute to Product-API that shows the "worst" sync-status that any of the last repositories of a product has ('running' is currently defined worse/more interesting than 'error')

The latter should be added to the hammer-cli output for Katello-Products.

### DIFF
--- a/app/models/katello/glue/pulp/repos.rb
+++ b/app/models/katello/glue/pulp/repos.rb
@@ -69,7 +69,7 @@ module Katello
 
       def sync_summary
         summary = {}
-        last_repo_sync_task_group.each do |task|
+        last_repo_sync_task_by_repoid.values.each do |task|
           summary[task.result] ||= 0
           summary[task.result] += 1
         end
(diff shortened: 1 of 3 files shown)


chris1984:
Starting to test this

m-bucher:
small improvement: https://github.com/Katello/katello/pull/8557

app/models/katello/glue/pulp/repos.rb
```diff
@@ -69,7 +69,7 @@ def sync_status
 
       def sync_summary
         summary = {}
-        last_repo_sync_task_group.each do |task|
+        last_repo_sync_task_by_repoid.values.each do |task|
```
m-bucher: @jlsherrill, @chris1984 looks like this also makes the product-index request slow, because `last_repo_sync_task_by_repoid()` is pretty slow.
I will look into making this faster.

Otherwise, I expect that we will have to remove this change again for the release.

m-bucher: addressed in https://github.com/Katello/katello/pull/8562

PR #8537, opened 2020-01-21
Fixes #28831 - adjust id to bigint

jlsherrill:


jlsherrill:
we probably want to have a discussion if any other tables need this adjustment

PR #8541, opened 2020-01-26
Fixes #28383 - Reinstate Activation Key autocomplete

jturel:
This depends on https://github.com/theforeman/foreman/pull/7387

FireShot Capture 003 - Edit testgroup - centos7-katello-devel jturel example com
FireShot Capture 004 - Edit testgroup - centos7-katello-devel jturel example com

Try to break it. Make sure that the host parameter in the database (::Parameter.all) is added/updated/removed correctly based on the selections made.


app/assets/javascripts/katello/hosts/activation_key_edit.js
```diff
@@ -26,17 +27,14 @@ function ktLoadActivationKeys() {
...
+        },
+        error: function() {
+          $("#ak-load-error").show();
         },
-        error: ktErrorLoadingActivationKeys
     });
 }
 
-function ktErrorLoadingActivationKeys(error) {
-    $.jnotify("Error while loading activation keys from Katello", { type: "error", sticky: true });
```
jturel: jnotify has been gone from Foreman for a while, so I changed how we show an error

app/assets/javascripts/katello/hosts/activation_key_edit.js
```diff
@@ -127,51 +120,24 @@ function ktAkTab() {
...
+    });
+
     ktHideParams();
     ktParamToAkInput();
     ktLoadActivationKeys();
 }
 
 $(document).on('ContentLoad', function(){
-
-    $("#kt_activation_keys").parents("form").submit(ktAkInputToParam);
```
jturel: the host/hostgroup parameter is now updated by listening for changes in the redux store rather than on submit

ianballou:
Is autocomplete supposed to work when the host group doesn't have a selected lifecycle environment and content view? I suppose I could see why we'd try to steer people away from entering activation keys if the host group doesn't belong to the same LE + CV, but the user can also just enter the activation key as an arbitrary string with "New selection: blah".

jturel:
> Is autocomplete supposed to work when the host group doesn't have a selected lifecycle environment and content view? I suppose I could see why we'd try to steer people away from entering activation keys if the host group doesn't belong to the same LE + CV, but the user can also just enter the activation key as an arbitrary string with "New selection: blah".

Yep! Keys can be entered which don't belong to the CV and LCE - and they don't even have to exist. Double check on dogfood just to be sure

ianballou:
@jturel ah, ok.  I'm noticing that the autocomplete isn't showing up unless I select a lifecycle environment and a content view.  Sounds like that isn't expected?

ianballou:
I'll re-test in case there need to be changes to the foreman PR.

app/assets/javascripts/katello/hosts/activation_key_edit.js
```diff
@@ -127,51 +125,24 @@ function ktAkTab() {
 }
 
 function ktOnLoad() {
+    tfm.store.observeStore('typeAheadSelect', function(items) {
+        if (items.kt_activation_keys) {
+            var selected = items.kt_activation_keys.selected;
+            selected = selected ? selected : [];
```
jeremylenz: I think this could be 
```js
var selected = items.kt_activation_keys.selected || [];
```

jturel:
> Thanks @jturel!
> 
> I created a host group with content view Default Organization View and lifecycle environment Library.
> I also created three Activation Keys with the same org / lifecycle environment.
> 
>     * white_check_mark All options show up in autocomplete
> 
>     * white_check_mark I am able to enter an arbitrary value also
> 
> 
> When I enter an arbitrary value, it does show up* in the pre-filled input field of the TypeAheadSelect when I return to the Activation Keys tab. However, it is not saved as a new Activation Key and is not a new option for autocomplete. Is this expected behavior?

- No new activation key will be created
- I didn't add it to autocomplete values, consistent with behavior before it was broken (also, I think it's the right thing)

> 
> When I enter Activation Keys into the TypeAheadSelect, and click Submit to save the host group, and then return to the Activation Key tab:
> 
>     * Sometimes I see the previously-selected activation keys pre-filled in the TypeAheadSelect
> 
>     * *Sometimes the TypeAheadSelect input field is blank (the autocomplete options do show, though.)  Looks like this:
>       hostgroup_activation_keys
>
The saved values should *always* show as selected in the input whether they were actual activation keys or not. But bogus keys or keys not part of the CV+LCE won't be in the autocomplete options

> 
> Which is the expected behavior? I haven't been able to get consistent behavior.
> 
> Also one minor suggestion below

jturel:
@jeremylenz @ianballou 

I've pushed an update to this PR that fixed a bug which would unintentionally clear out the parameter in the Db:

- set some keys on a hostgroup(saving doesnt matter)
- clear all selections
- add some selections back
- save
- host parameter is gone!

I also pushed some changes to address feedback in the Foreman PR and it should be more or less complete.

@ianballou since you're going to be out - do you mind delegating final ACK to @jeremylenz since it may happen while you're away?

app/assets/javascripts/katello/hosts/activation_key_edit.js
```diff
@@ -82,28 +80,25 @@ function ktSetParam(name, value) {
             paramContainer.find("input[name*='name']").val(name);
         }
         paramContainer.find("textarea").val(value);
+        paramContainer.find("input[ type = 'hidden' ]").val(0);
```
jturel: This one-liner fixed the bug w.r.t the host parameter getting cleared out

jturel:
@jeremylenz pushed a separate commit so you can see how I'm handling trimming spaces - give it a whirl

PR #8549, opened 2020-01-29
Fixes #28886 - App applicability associations to InstalledPackage

ianballou:
Adds version, release, epoch, and arch to InstalledPackage to prep for the new Pulp 3 Katello applicability.  Note that there is a database migration.  The migration tries to find the correct epoch among packages that Katello has indexed.  If it cannot find any, it sets the epoch to 0.  This may be incorrect, but the next package upload should update the epoch.

To test: 
1) Register a content host
2) Makes sure the package profile was uploaded
3) Checkout this PR
4) Run the migration
5) Ensure that the new columns exist for the existing InstalledPackage(s)
6) Install something (like `sl` for example)
7) Check the installed packages again to see that new ones were uploaded
8) Play around with the package uploads and see if anything breaks
9) Another testing idea: create your own `InstalledPackage` entries with nvras that match some RPM and see if they migrate properly.  Try editing the epoch of the RPMs to ensure the migration is correctly updating the `InstalledPackage` epochs to match the RPM epochs.

### DIFF
--- a/app/models/katello/concerns/host_managed_extensions.rb
+++ b/app/models/katello/concerns/host_managed_extensions.rb
@@ -110,17 +110,26 @@ module Katello
       end
 
       def import_package_profile_in_bulk(simple_packages)
-        nvras = simple_packages.map { |sp| sp.nvra }
-        found = InstalledPackage.where(:nvra => nvras).select(:id, :nvra).to_a
-        found_nvras = found.map(&:nvra)
-        new_packages = simple_packages.select { |sp| !found_nvras.include?(sp.nvra) }
+        nvreas = simple_packages.map { |sp| sp.nvrea }
+        found = InstalledPackage.where(:nvrea => nvreas).select(:id, :nvrea).to_a
+        found_nvreas = found.map(&:nvrea)
 
+        new_packages = simple_packages.select { |sp| !found_nvreas.include?(sp.nvrea) }
+
+        installed_packages = []
         new_packages.each do |simple_package|
-          ::Katello::Util::Support.active_record_retry do
-            found << InstalledPackage.where(:nvra => simple_package.nvra, :name => simple_package.name).first_or_create!
-          end
+          installed_packages << InstalledPackage.new(:nvrea => simple_package.nvrea,
+                                          :nvra => simple_package.nvra,
+                                          :name => simple_package.name,
+                                          :epoch => simple_package.epoch,
+                                          :version => simple_package.version,
+                                          :release => simple_package.release,
+                                          :arch => simple_package.arch)
         end
-        found
+        InstalledPackage.import(installed_packages, validate: false, on_duplicate_key_ignore: true)
+
+        found << installed_packages
+        found.flatten
       end
 
       def import_enabled_repositories(repos)
--- a/db/migrate/20200129172534_add_epoch_version_release_arch_to_katello_installed_packages.rb
+++ b/db/migrate/20200129172534_add_epoch_version_release_arch_to_katello_installed_packages.rb
@@ -1,0 +1,40 @@
+class AddEpochVersionReleaseArchToKatelloInstalledPackages < ActiveRecord::Migration[5.2]
+  def up
+    add_column :katello_installed_packages, :nvrea, :string
+    add_column :katello_installed_packages, :epoch, :string
+    add_column :katello_installed_packages, :version, :string
+    add_column :katello_installed_packages, :release, :string
+    add_column :katello_installed_packages, :arch, :string
+
+    epoch_non_0 = ::Katello::Rpm.where.not(epoch: [0, nil]).pluck(:nvra, :epoch).to_h
+    installed_packages = []
+
+    ::Katello::InstalledPackage.reset_column_information
+    ::Katello::InstalledPackage.find_each do |pkg|
+      epoch = epoch_non_0[pkg.nvra] || "0"
+
+      attributes_hash = ::Katello::Util::Package.parse_nvrea(pkg.nvra)
+      attributes_hash[:epoch] = epoch
+      attributes_hash[:nvra] = pkg.nvra
+      if epoch == "0"
+        attributes_hash[:nvrea] = pkg.nvra
+      else
+        attributes_hash[:nvrea] = "#{pkg.name}-#{epoch}:#{attributes_hash[:version]}-"\
+                                  "#{attributes_hash[:release]}.#{attributes_hash[:arch]}"
+      end
+
+      installed_packages << ::Katello::InstalledPackage.new(attributes_hash)
+    end
+    ::Katello::InstalledPackage.import(installed_packages, validate: false, batch_size: 50_000,
+                                       on_duplicate_key_update: {conflict_target: [:nvra],
+                                                                 columns: [:nvrea, :epoch, :version, :release, :arch]})
+  end
+
+  def down
+    remove_column :katello_installed_packages, :nvrea, :string
+    remove_column :katello_installed_packages, :epoch, :string
+    remove_column :katello_installed_packages, :version, :string
+    remove_column :katello_installed_packages, :release, :string
+    remove_column :katello_installed_packages, :arch, :string
+  end
+end
--- a/app/models/katello/installed_package.rb
+++ b/app/models/katello/installed_package.rb
@@ -8,10 +8,11 @@ module Katello
     has_many :host_installed_packages, :class_name => "Katello::HostInstalledPackage", :dependent => :destroy, :inverse_of => :installed_package
 
     scoped_search :on => :name, :complete_value => true
+    scoped_search :on => :nvrea
     scoped_search :on => :nvra
-
-    def nvrea
-      nvra
-    end
+    scoped_search :on => :epoch
+    scoped_search :on => :version
+    scoped_search :on => :release
+    scoped_search :on => :arch
   end
 end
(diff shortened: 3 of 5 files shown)


app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -110,17 +110,26 @@ def import_package_profile(simple_packages)
...
+          installed_packages << InstalledPackage.new(:nvrea => simple_package.nvrea,
+                                          :nvra => simple_package.nvra,
+                                          :name => simple_package.name,
+                                          :epoch => simple_package.epoch,
+                                          :version => simple_package.version,
+                                          :release => simple_package.release,
+                                          :arch => simple_package.arch)
         end
-        found
+        InstalledPackage.import(installed_packages)
```
jlsherrill: I'd probably add:  validate: false  and probably  on_duplicate_key_ignore: true

jlsherrill: and to clarify, typically when we're just 'inserting' data from some other source, we are okay ignoring validations because we typically don't have validations

ianballou: That sounds good, I was trying to decide before if I needed those.

db/migrate/20200129172534_add_epoch_version_release_arch_to_katello_installed_packages.rb
```diff
@@ -0,0 +1,44 @@
...
+      release = tokens[-1]
+      version = tokens[-2]
+
+      pkg.arch = arch
+      pkg.release = release
+      pkg.version = version
+
+      pkg.nvrea = "#{epoch}:#{pkg.name}-#{version}-#{release}.#{arch}"
+      pkg.save!
+    end
```
jlsherrill: can we simplify any of this  via this method?    https://github.com/Katello/katello/blob/master/app/lib/katello/util/package.rb#L22-L30

db/migrate/20200129172534_add_epoch_version_release_arch_to_katello_installed_packages.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
```
jlsherrill: Should we set the epoch to zero for all the other installed packages after this runs?

db/migrate/20200129172534_add_epoch_version_release_arch_to_katello_installed_packages.rb
```diff
@@ -0,0 +1,44 @@
...
+    add_column :katello_installed_packages, :arch, :string
+
+    epoch_non_0 = ::Katello::Rpm.where.not(epoch: 0)
+
+    ::Katello::InstalledPackage.reset_column_information
+    ::Katello::InstalledPackage.all.each do |pkg|
+      epoch = epoch_non_0.where(nvra: pkg.nvra).first
+      if epoch.nil?
+        epoch = '0'
+      end
```
jlsherrill: i think you could do this above to accomplish the same thing: 
`Katello::Rpm.where.not(:epoch => [0, nil]`

jlsherrill: actually i misunderstood what this was doing, i see now that its looping over every installed package to update it.  I did some testing on a user's db and this seemed really slow.  Something like this:

```
epoch_non_0 = ::Katello::Rpm.where.not(epoch: 0)
installed_packages = Katello::InstalledPackage.where(:nvra => epoch_non_0.select(:nvra)).each do |pkg|
   #DO STUFF 
end
Katello::InstalledPackage.where(:epoch => nil).update_all(:epoch => 0)
```

was much faster.  Here's some stats from the db:

irb(main):077:0> Katello::Rpm.count
=> 198944
irb(main):078:0> Katello::InstalledPackage.count
=> 381726
> installed_packages = Katello::InstalledPackage.where(:nvra => epoch_non_0.select(:nvra)).count
=> 3068

without any of the .save! or update_all  it finished in about ~100 seconds which seems resonable for  a large dataset.

db/migrate/20200129172534_add_epoch_version_release_arch_to_katello_installed_packages.rb
```diff
@@ -0,0 +1,44 @@
...
+      tokens = nvra.split('-')
+      release = tokens[-1]
+      version = tokens[-2]
+
+      pkg.arch = arch
+      pkg.release = release
+      pkg.version = version
+
+      pkg.nvrea = "#{epoch}:#{pkg.name}-#{version}-#{release}.#{arch}"
+      pkg.save!
```
jlsherrill: i think saving these to an array and then using the bulk importing library would speed this up a lot

db/migrate/20200129172534_add_epoch_version_release_arch_to_katello_installed_packages.rb
ianballou: @jlsherrill  If I filter the packages using `Katello::InstalledPackage.where(:nvra => epoch_non_0.select(:nvra))`, then only `InstalledPackage`s with a matching `Rpm` will have the `nvrea`, `version`, `release`, and `arch` filled in.  That's why I needed to loop through each individual `InstalledPackage`s.  If you're okay with those fields being `nil` then I can switch to the strategy you mentioned.

db/migrate/20200129172534_add_epoch_version_release_arch_to_katello_installed_packages.rb
ianballou: Looking into this, for some reason, even if I do `::Katello::InstalledPackage.import(installed_packages, validate: false, on_duplicate_key_update: :all)` there are still errors, like `PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_katello_installed_packages_on_nvra"
DETAIL:  Key (nvra)=(cheetah-1.25.3-5.noarch) already exists.`

ianballou: Oh I see why, activerecord import's `on_duplicate_key_update` requires PostgreSQL 9.5+, but in dev we're on `"PostgreSQL 9.2.24 on x86_64-redhat-linux-gnu, compiled by gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-28), 64-bit"`

jlsherrill:
We should be upgraded to 12 in the latest nightly, how old is your Dev env?

jlsherrill:
Oh right, we're updating other stuff too!  My mind was focused on epoch

ianballou:
@jlsherrill it looks like there are some more pending PRs until postgres 12 is in the dev box: https://github.com/theforeman/foreman-installer/pull/463

I'll bring up a fresh box just in case it is in though.

ianballou:
I've made some fixes to the migration.  It should be a bit more efficient and organized now.  I'm still looping through all the installed packages, but now I'm using Active Record Import.

ianballou:
[test katello]

db/migrate/20200129172534_add_epoch_version_release_arch_to_katello_installed_packages.rb
```diff
@@ -0,0 +1,35 @@
...
+    add_column :katello_installed_packages, :nvrea, :string
+    add_column :katello_installed_packages, :epoch, :string
+    add_column :katello_installed_packages, :version, :string
+    add_column :katello_installed_packages, :release, :string
+    add_column :katello_installed_packages, :arch, :string
+
+    epoch_non_0 = ::Katello::Rpm.where.not(epoch: [0, nil]).pluck(:nvra, :epoch)
+
+    ::Katello::InstalledPackage.reset_column_information
+    installed_packages = ::Katello::InstalledPackage.all.collect do |pkg|
```
jlsherrill: I'd probably use InstalledPackage.find_each for this: https://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_each

db/migrate/20200129172534_add_epoch_version_release_arch_to_katello_installed_packages.rb
```diff
@@ -0,0 +1,38 @@
...
+    add_column :katello_installed_packages, :release, :string
+    add_column :katello_installed_packages, :arch, :string
+
+    epoch_non_0 = ::Katello::Rpm.where.not(epoch: [0, nil]).pluck(:nvra, :epoch)
+    installed_packages = []
+
+    ::Katello::InstalledPackage.reset_column_information
+    ::Katello::InstalledPackage.find_each do |pkg|
+      #epoch = epoch_non_0.select{ |nvra_epoch| nvra_epoch[0] == pkg.nvra }.first&.second || "0"
+      epoch_non_0_map = epoch_non_0.map.to_h
```
jlsherrill: you can move this out of the loop to only do it once?  Also don't need the previous line

jlsherrill:
This all looks good and performs well for me now!  good work :+1: 

Do you mind updating the host_managed_extension_test.rb for import_package_profile tests, to take nvrea and epoch into consideration?

ianballou:
@jlsherrill I added to the test, let me know if you want more there.

ianballou:
I found a package where the epoch comes after the name in the NVREA for some reason, I'll investigate that tomorrow.  Example: `"atomic-1:1.13.8-1.el7.x86_64"`.

jlsherrill:
test looks fine, let me know what you find out about that package

ianballou:
@jlsherrill looks like the epoch placement is unrelated to my PR (also since it's on an RPM rather than an InstalledPackage).  Seems like `Util::Package.build_nvrea` wants to put epoch after the name.

ianballou:
Should be all good now.

PR #8550, opened 2020-01-30
Fixes  #28894 - load the pools from AR database instead of Candlepin

matt8754:
In case there is a large list of pools, this will improve the speed when loading up the subscription page.

### DIFF
--- a/app/models/katello/glue/candlepin/candlepin_object.rb
+++ b/app/models/katello/glue/candlepin/candlepin_object.rb
@@ -3,12 +3,6 @@ module Katello
     extend ActiveSupport::Concern
 
     module ClassMethods
-      def get_for_organization(organization)
-        # returns objects from AR database rather than candlepin data
-        pool_ids = self.get_for_owner(organization.label).collect { |x| x['id'] }
-        self.where(:cp_id => pool_ids)
-      end
-
       def get_candlepin_ids(organization)
         self.get_for_owner(organization.label).map { |subscription| subscription["id"] }
       end
--- a/app/controllers/katello/api/v2/subscriptions_controller.rb
+++ b/app/controllers/katello/api/v2/subscriptions_controller.rb
@@ -65,7 +65,7 @@ module Katello
       return available_for_activation_key if params[:available_for] == "activation_key"
       collection = Pool.readable
       collection = collection.where(:unmapped_guest => false)
-      collection = collection.get_for_organization(Organization.find(params[:organization_id])) if params[:organization_id]
+      collection = collection.where(organization: Organization.find(params[:organization_id])) if params[:organization_id]
       collection = collection.for_activation_key(@activation_key) if params[:activation_key_id]
       collection
     end
--- a/test/controllers/api/v2/subscriptions_controller_test.rb
+++ b/test/controllers/api/v2/subscriptions_controller_test.rb
@@ -34,15 +34,14 @@ module Katello
     end
 
     def test_index
-      Pool.expects(:get_for_organization).returns(Pool.all)
       get :index, params: { :organization_id => @organization.id }
-
+      results = JSON.parse(@response.body)['results']
+      assert_equal results.count, Pool.all.size
       assert_response :success
       assert_template 'api/v2/subscriptions/index'
     end
 
     def test_index_csv
-      Pool.expects(:get_for_organization).returns(Pool.all)
       get :index, params: {:format => 'csv', :organization_id => @organization.id }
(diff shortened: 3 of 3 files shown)


chris1984:
[ok to test]

chris1984:
@matt8754 how is a good way to test this, how many subs/hosts do you need?

jturel:
This is a good change. I can see an unnecessary call being made to candlepin

before:
```
INFO  org.candlepin.common.filter.LoggingFilter - Request: verb=GET, uri=/candlepin/owners/Default_Organization/pools?add_future=true&attribute=unmapped_guests_only:!true
INFO  org.candlepin.common.filter.LoggingFilter - Request: verb=GET, uri=/candlepin/pools/4028fac46fc959b7016fc976f5800a30
INFO  org.candlepin.common.filter.LoggingFilter - Request: verb=GET, uri=/candlepin/owners/Default_Organization/products/SER0482/?
INFO  org.candlepin.common.filter.LoggingFilter - Request: verb=GET, uri=/candlepin/pools/4028fac46fd2c37b016ffd7bc1e8039c
INFO  org.candlepin.common.filter.LoggingFilter - Request: verb=GET, uri=/candlepin/owners/Default_Organization/products/RH00005/?
INFO  org.candlepin.common.filter.LoggingFilter - Request: verb=GET, uri=/candlepin/pools/4028fac46fd2c37b016ff327e9520167
INFO  org.candlepin.common.filter.LoggingFilter - Request: verb=GET, uri=/candlepin/owners/Default_Organization/products/605248846626/?
```

after:
```
INFO  org.candlepin.common.filter.LoggingFilter - Request: verb=GET, uri=/candlepin/pools/4028fac46fc959b7016fc976f5800a30
INFO  org.candlepin.common.filter.LoggingFilter - Request: verb=GET, uri=/candlepin/owners/Default_Organization/products/SER0482/?
INFO  org.candlepin.common.filter.LoggingFilter - Request: verb=GET, uri=/candlepin/pools/4028fac46fd2c37b016ffd7bc1e8039c
INFO  org.candlepin.common.filter.LoggingFilter - Request: verb=GET, uri=/candlepin/owners/Default_Organization/products/RH00005/?
INFO  org.candlepin.common.filter.LoggingFilter - Request: verb=GET, uri=/candlepin/pools/4028fac46fd2c37b016ff327e9520167
INFO  org.candlepin.common.filter.LoggingFilter - Request: verb=GET, uri=/candlepin/owners/Default_Organization/products/605248846626/?
```

I verified that this does not affect the output of the API.

app/models/katello/glue/candlepin/candlepin_object.rb
```diff
@@ -5,8 +5,7 @@ module Glue::Candlepin::CandlepinObject
     module ClassMethods
       def get_for_organization(organization)
         # returns objects from AR database rather than candlepin data
-        pool_ids = self.get_for_owner(organization.label).collect { |x| x['id'] }
-        self.where(:cp_id => pool_ids)
+        self.where(:organization => organization)
```
jturel: It looks like this method is only used in one place, and I don't think it makes sense to have a separate method anymore since it only has a single `where` clause.

Let's remove this method and update subscriptions_controller.rb:

```
+++ b/app/controllers/katello/api/v2/subscriptions_controller.rb
@@ -65,7 +65,7 @@ module Katello
       return available_for_activation_key if params[:available_for] == "activation_key"
       collection = Pool.readable
       collection = collection.where(:unmapped_guest => false)
-      collection = collection.get_for_organization(Organization.find(params[:organization_id])) if params[:organization_id]
+      collection = collection.where(organization: Organization.find(params[:organization_id])) if params[:organization_id]
       collection = collection.for_activation_key(@activation_key) if params[:activation_key_id]
       collection
     end
```

I think the tests will need to be updated. You can verify locally with:
```
ktest ./test/controllers/api/v2/subscriptions_controller_test.rb
```

matt8754:
> @matt8754 how is a good way to test this, how many subs/hosts do you need?

Ideally, it would be good to have 3K+ hypervisors with the VDC sub attached. This would match the customer's env from which the issue happens.

app/models/katello/glue/candlepin/candlepin_object.rb
matt8754: Thanks, Jonathon. This is exactly what I came out first. However, I was worried about breaking the tests as they call get_for_organization. I'll take a look further for that.

jturel:
[test katello]

PR #8552, opened 2020-01-31
Fixes #28476 - Add option for CDN SSL Versions to Settings

chris1984:


chris1984:
@ekohl Would you be fine with a rake task that reads the yaml file and if there is an option then we update the setting in Katello? If it was just in the installer moving around I would propose a migration there but a rake task seems appropriate, thoughts?

ekohl:
An alternative is to use a foreman-installer hook. If the answer is present, we can call `foreman-rake config -k cdn_ssl_version -v #{answers['katello']['cdn_ssl_version']}` after running the installer and then drop the answer.

For the direct users of puppet-katello we can add a note in the release notes but I doubt they use this setting.

chris1984:
> An alternative is to use a foreman-installer hook. If the answer is present, we can call `foreman-rake config -k cdn_ssl_version -v #{answers['katello']['cdn_ssl_version']}` after running the installer and then drop the answer.
> 
> For the direct users of puppet-katello we can add a note in the release notes but I doubt they use this setting.

The hook route seems the easiest and less error prone. I will make a hook then, I just know hooks are generally looked down upon. I will make a pr to the installer with a `refs` for you to review.

ekohl:
> I just know hooks are generally looked down upon

Not entirely true, but hooks should only be used when they make sense. I think this is one of those cases.

chris1984:
[test katello]

chris1984:
[test katello]

chris1984:
@ekohl hope your travels were good, did you get a chance to see what i had for the installer hook and the issues calling the answers method?

chris1984:
@ekohl I think once this is merged https://github.com/theforeman/puppet-katello/pull/320 I think this one is good then? The installer one was merged this morning.

chris1984:
@jturel can I get a review please on this, the other 2 prs are merged.

app/models/setting/content.rb
```diff
@@ -28,6 +29,9 @@ def self.default_settings
                       nil,
                       collection: proc { http_proxy_select }, include_blank: N_("no global default")
               ),
+      self.set('cdn_ssl_version', N_("SSL version used to communicate with the CDN"),
```
jturel, line 32: The description is the same as the setting name. How about changing the name here to 'CDN SSL version' ?

app/lib/katello/resources/cdn.rb
```diff
@@ -66,7 +66,7 @@ def http_downloader
           # Run the following command in rails console to figure out other
           # valid constants in other ruby versions
           # "OpenSSL::SSL::SSLContext::METHODS"
-          net.ssl_version = SETTINGS[:katello][:cdn_ssl_version] if SETTINGS[:katello].key?(:cdn_ssl_version)
+          net.ssl_version = Setting[:cdn_ssl_version] if Setting[:cdn_ssl_version]
```
jturel, line 69: Since we can't validate the value of setting when we set or modify it, we might want to perform some validation in this file where we use that value since something bogus is probably going to raise a not very helpful error.

One solution might be to add a constant in this file:

SUPPORTED_SSL_VERSIONS = ['SSLv23', 'TLSv1']

Then in the constructor we can check:

```
if Setting[:cdn_ssl_version] && !SUPPORTED_SSL_VERSIONS.include?(Setting[:cdn_ssl_version))
   fail("Invalid SSL version specified. Check the 'CDN SSL Version' setting")
end
```

Let me know if you like the idea. This should give a useful error when enabling a repository when the SSL version is misconfigured.

app/models/setting/content.rb
```diff
@@ -14,6 +14,7 @@ def self.default_settings
     download_policies = proc { hashify_parameters(::Runcible::Models::YumImporter::DOWNLOAD_POLICIES) }
     proxy_download_policies = proc { hashify_parameters(::SmartProxy::DOWNLOAD_POLICIES) }
     dependency_solving_options = proc { hashify_parameters(['conservative', 'greedy']) }
+    cdn_ssl_versions = proc { hashify_parameters(['SSLv23', 'TLSv1']) }
```
jturel, line 17: If we go with my suggestion around the SUPPORTED_SSL_VERSIONS constant we can reference if here ie:

```
proc { hashify_parameters(Katello::Resources::CDN::SUPPORTED_SSL_VERSIONS) }
```

test/models/setting_test.rb
```diff
@@ -3,24 +3,34 @@
...
       assert setting.valid?
     end
 
+    def test_cdn_ssl_setting
+      setting = Setting.where(:name => 'cdn_ssl_version').first
+      setting.value = nil
+      assert setting.valid?
+      setting.value = 'SSLv23'
+      assert setting.valid?
+      setting.value = 'TLSv1'
```
jturel, line 27: The tests here might be a little redundant. What I mean is that if we give it something bogus, like 'foo' will it be invalid? Checking both  SSLv23 and TLSv1 is not really increasing the coverage in any way as far as I can tell, but let me know if I'm missing something.

We can also increase the readability of the code by adding some whitespace:

```
   def test_cdn_ssl_setting
      setting = Setting.where(:name => 'cdn_ssl_version').first

      setting.value = nil
      assert setting.valid?

      setting.value = 'SSLv23'
      assert setting.valid?

      setting.value = 'TLSv1'
      assert setting.valid?
end
```

app/lib/katello/resources/cdn.rb
ekohl: There's a list of valid options, shouldn't that be sufficient validation?

jturel: @ekohl i'm not sure how those options come in to play. in the UI I can still enter any value that I want and I don't see any reference to the valid options, even on the CLI

```
[vagrant@centos7-hammer-devel ~]$ hammer settings set --name cdn_ssl_version --value "whatever"
Setting [cdn_ssl_version] updated to [whatever].
```
```
vagrant@centos7-hammer-devel ~]$ hammer settings list | grep cdn_ssl_version
cdn_ssl_version                                        | SSL version used to communicate with the CDN                | whatever                                                                         | SSL version used to communicate with the CDN      
```

Unless I'm missing something I think the validation is useful because it appears there's none and the user does not even know the valid options

jturel: Tested and found that setting some random value for SSL version doesn't seem to hurt anything when making the connection. I guess it's ignored and the version is renegotiated, so maybe the validation really is unnecessary. I feel less strongly about the validation now since it's not a total failure when the value is invalid.

jturel: Hmm wait, now I am seeing a stack trace:

```
16:37:00 rails.1   | 2020-04-02T16:37:00 [W|not|23dac3ad] Failed to handle notifications - this is most likely a bug: unknown SSL method `whatever'                                     
16:37:00 rails.1   | 2020-04-02T16:37:00 [D|not|23dac3ad] /opt/rh/rh-ruby25/root/usr/share/ruby/openssl/ssl.rb:209:in `ssl_version='                                                          16:37:00 rails.1   |  | /opt/rh/rh-ruby25/root/usr/share/ruby/openssl/ssl.rb:140:in `block in set_params'                          
16:37:00 rails.1   |  | /opt/rh/rh-ruby25/root/usr/share/ruby/openssl/ssl.rb:140:in `each'                                                                                                    
16:37:00 rails.1   |  | /opt/rh/rh-ruby25/root/usr/share/ruby/openssl/ssl.rb:140:in `set_params'                                                                                              16:37:00 rails.1   |  | /opt/rh/rh-ruby25/root/usr/share/ruby/net/http.rb:971:in `connect'                                                                                                    
16:37:00 rails.1   |  | /opt/rh/rh-ruby25/root/usr/share/ruby/net/http.rb:920:in `do_start'                                                                                                   
16:37:00 rails.1   |  | /opt/rh/rh-ruby25/root/usr/share/ruby/net/http.rb:909:in `start'                                                                                                      
16:37:00 rails.1   |  | /home/vagrant/katello/app/lib/katello/resources/cdn.rb:94:in `get'
```

chris1984: I will put the validation in that you talked about then :)

chris1984:
@jturel updated! Screenshot of new setting change:

https://nimbusweb.me/s/share/4045678/fz6dtewbym4awma8swg1

app/lib/katello/resources/cdn.rb
```diff
@@ -66,7 +67,10 @@ def http_downloader
           # Run the following command in rails console to figure out other
           # valid constants in other ruby versions
           # "OpenSSL::SSL::SSLContext::METHODS"
-          net.ssl_version = SETTINGS[:katello][:cdn_ssl_version] if SETTINGS[:katello].key?(:cdn_ssl_version)
+          if Setting[:cdn_ssl_version] && !SUPPORTED_SSL_VERSIONS.include?(Setting[:cdn_ssl_version])
```
jturel, line 70: This is working well, but I think it would be better to move the check to the initialize method in this class so we can bail out as soon as possible. Example:

```
          @ssl_version = Setting[:cdn_ssl_version]                                                                                                                                            
          if @ssl_version && !SUPPORTED_SSL_VERSIONS.include?(@ssl_version)                                                                                                                   
            fail("Invalid SSL version specified. Check the 'CDN SSL Version' setting")                                                                                                        
          end 
```

Then in this method we can set the value:

```
net.ssl_version = @ssl_version
```

test/lib/resources/cdn_test.rb
```diff
@@ -0,0 +1,19 @@
...
+  def test_http_downloader_tls
+    cdn_resource = Katello::Resources::CDN::CdnResource.new('http://foo.com')
+    Setting[:cdn_ssl_version] = 'TLSv1'
+    assert_equal cdn_resource.http_downloader.ssl_version, 'TLSv1'
+  end
+
+  def test_http_downloader_no_version
+    cdn_resource = Katello::Resources::CDN::CdnResource.new('http://foo.com')
+    assert_nil cdn_resource.http_downloader.ssl_version
+  end
```
jturel, line 18: Let's add a test here to check for the raised error when we initialize the CdnResource - you can find some examples in other test with assert_raises iirc

chris1984:
@jturel thanks, updated!

app/lib/katello/resources/cdn.rb
```diff
@@ -66,7 +71,7 @@ def http_downloader
           # Run the following command in rails console to figure out other
           # valid constants in other ruby versions
           # "OpenSSL::SSL::SSLContext::METHODS"
-          net.ssl_version = SETTINGS[:katello][:cdn_ssl_version] if SETTINGS[:katello].key?(:cdn_ssl_version)
+          net.ssl_version = net.ssl_version = @ssl_version
```
jturel, line 74: I don't think you meant to write this (double assignment)

chris1984:
@jturel I like to keep it interesting ;) copy and paste fail. Updated

test/lib/resources/cdn_test.rb
```diff
@@ -0,0 +1,26 @@
...
+class CdnResourceTest < ActiveSupport::TestCase
+  def test_http_downloader_v2
+    cdn_resource = Katello::Resources::CDN::CdnResource.new('http://foo.com')
+    Setting[:cdn_ssl_version] = 'SSLv23'
+    assert_equal cdn_resource.http_downloader.ssl_version, 'SSLv23'
+  end
+
+  def test_http_downloader_tls
+    cdn_resource = Katello::Resources::CDN::CdnResource.new('http://foo.com')
+    Setting[:cdn_ssl_version] = 'TLSv1'
```
jturel, line 11: issue with the failing tests is that the Setting must be set before creating the cdn_resource

PR #8553, opened 2020-02-04
Fixes #28913 - Add pulp3_href to Repository Errata

sjha4:
To Do:

- [x] Index pulp3_href for errata.
- [x] Test pulp2 indexing/sync.
- [x] Tests

### DIFF
--- a/app/models/katello/concerns/pulp_database_unit.rb
+++ b/app/models/katello/concerns/pulp_database_unit.rb
@@ -131,7 +131,6 @@ module Katello
       end
 
       def sync_repository_associations(repository, options = {})
-        associated_ids, id_href_map_for_repository = [], {}
         additive = options.fetch(:additive, false)
         pulp_id_href_map = options.dig(:pulp_id_href_map) || {}
         pulp_ids = options.dig(:pulp_ids) || pulp_id_href_map.try(:keys)
@@ -144,37 +143,38 @@ module Katello
           self.repository_association_class.where(:repository_id => repository).pluck(unit_id_field)
         end
 
-        if !additive && (delete_ids = existing_ids - associated_ids).any?
-          query = "DELETE FROM #{self.repository_association_class.table_name} WHERE repository_id=#{repository.id} AND #{unit_id_field} IN (#{delete_ids.join(', ')})"
-          ActiveRecord::Base.connection.execute(query)
-        end
-
-        unless (new_ids = associated_ids - existing_ids).empty?
-          self.repository_association_class.import(db_columns_sync, db_values(new_ids, id_href_map_for_repository, repository), validate: false)
+        ActiveRecord::Base.transaction do
+          if !additive && (delete_ids = existing_ids - associated_ids).any?
+            query = "DELETE FROM #{self.repository_association_class.table_name} WHERE repository_id=#{repository.id} AND #{unit_id_field} IN (#{delete_ids.join(', ')})"
+            ActiveRecord::Base.connection.execute(query)
+          end
+          unless (new_ids = associated_ids - existing_ids).empty?
+            self.repository_association_class.import(db_columns_sync, db_values(new_ids, id_href_map_for_repository, repository), validate: false)
+          end
         end
       end
 
       def copy_repository_associations(source_repo, dest_repo)
-        queries = []
         if many_repository_associations
-          queries << "delete from #{repository_association_class.table_name} where repository_id = #{dest_repo.id} and
+          delete_query = "delete from #{repository_association_class.table_name} where repository_id = #{dest_repo.id} and
                          #{unit_id_field} not in (select #{unit_id_field} from #{repository_association_class.table_name} where repository_id = #{source_repo.id})"
-
-          self.repository_association_class.import(db_columns_copy, db_values_copy(source_repo, dest_repo), validate: false)
+          ActiveRecord::Base.transaction do
+            ActiveRecord::Base.connection.execute(delete_query)
+            self.repository_association_class.import(db_columns_copy, db_values_copy(source_repo, dest_repo), validate: false)
+          end
         else
           columns = column_names - ["id", "pulp_id", "created_at", "updated_at", "repository_id"]
-
+          queries = []
           queries << "delete from #{self.table_name} where repository_id = #{dest_repo.id} and
                           pulp_id not in (select pulp_id from #{self.table_name} where repository_id = #{source_repo.id})"
           queries << "insert into #{self.table_name} (repository_id, pulp_id, #{columns.join(',')})
                     select #{dest_repo.id} as repository_id, pulp_id, #{columns.join(',')} from #{self.table_name}
                     where repository_id = #{source_repo.id} and pulp_id not in (select pulp_id
                     from #{self.table_name} where repository_id = #{dest_repo.id})"
-        end
-
-        ActiveRecord::Base.transaction do
-          queries.each do |query|
-            ActiveRecord::Base.connection.execute(query)
+          ActiveRecord::Base.transaction do
+            queries.each do |query|
+              ActiveRecord::Base.connection.execute(query)
+            end
           end
         end
       end
--- a/app/models/katello/concerns/pulp_database_unit.rb
+++ b/app/models/katello/concerns/pulp_database_unit.rb
@@ -146,6 +146,9 @@ module Katello
         end
 
         unless new_ids.empty?
+          # column = columns
+          # value = values(new_ids, repository)
+          # self.repository_association_class.import column, value, validate:false
           inserts = new_ids.map { |unit_id| "(#{unit_id.to_i}, #{repository.id.to_i}, '#{Time.now.utc.to_s(:db)}', '#{Time.now.utc.to_s(:db)}')" }
           queries << "INSERT INTO #{table_name} (#{attribute_name}, repository_id, created_at, updated_at) VALUES #{inserts.join(', ')}"
         end
@@ -183,6 +186,14 @@ module Katello
 
       def with_pulp_id(unit_pulp_ids)
         where('pulp_id in (?)', unit_pulp_ids)
+      end
+
+      def columns
+        [unit_id_field.to_sym, :repository_id, :created_at, :updated_at]
+      end
+
+      def values(new_ids, repository)
+        new_ids.map { |unit_id| [unit_id.to_i, repository.id.to_i, Time.now.utc.to_s(:db), Time.now.utc.to_s(:db)] }
       end
     end
   end
--- a/app/models/katello/concerns/pulp_database_unit.rb
+++ b/app/models/katello/concerns/pulp_database_unit.rb
@@ -43,6 +43,10 @@ module Katello
         self::CONTENT_TYPE
       end
 
+      def backend_identifier_field
+        nil
+      end
+
       def content_unit_class
         "::Katello::Pulp::#{self.name.demodulize}".constantize
       end
@@ -110,7 +114,7 @@ module Katello
           units.each do |unit|
             unit = unit.with_indifferent_access
             pulp_id = unit[service_class.unit_identifier]
-            backend_identifier = unit[service_class.backend_unit_identifier] if service_class.backend_unit_identifier
+            backend_identifier = unit.dig(service_class.backend_unit_identifier)
             unless fetch_only_ids
               model = Katello::Util::Support.active_record_retry do
                 self.where(:pulp_id => pulp_id).first_or_create
@@ -121,7 +125,7 @@ module Katello
               service.update_model(model)
             end
             pulp_ids << pulp_id
-            pulp_id_href_map.merge!(pulp_id => backend_identifier) if service_class.backend_unit_identifier
+            pulp_id_href_map.merge!(pulp_id => backend_identifier) if backend_identifier
           end
         end
         sync_repository_associations(repository, :pulp_ids => pulp_ids, :pulp_id_href_map => pulp_id_href_map) if self.many_repository_associations
@@ -129,9 +133,9 @@ module Katello
 
       def sync_repository_associations(repository, options = {})
         additive = options.fetch(:additive, false)
-        pulp_ids = options.fetch(:pulp_ids, nil)
-        pulp_id_href_map = options.fetch(:pulp_id_href_map, nil)
-        id_map_for_repository, table_name, attribute_name = {}, self.repository_association_class.table_name, unit_id_field
+        pulp_ids = options.dig(:pulp_ids)
+        pulp_id_href_map = options.dig(:pulp_id_href_map)
+        id_map_for_repository = {}
         if pulp_ids
           ids_for_repository = with_pulp_id(pulp_ids).pluck(:id, :pulp_id)
           associated_ids = ids_for_repository.map { |ids| ids[0] }
@@ -141,19 +145,18 @@ module Katello
           end
         end
         existing_ids = self.repository_association_class.uncached do
-          self.repository_association_class.where(:repository_id => repository).pluck(attribute_name)
+          self.repository_association_class.where(:repository_id => repository).pluck(unit_id_field)
         end
         new_ids = associated_ids - existing_ids
         delete_ids = existing_ids - associated_ids
         queries = []
 
         if delete_ids.any? && !additive
-          queries << "DELETE FROM #{table_name} WHERE repository_id=#{repository.id} AND #{attribute_name} IN (#{delete_ids.join(', ')})"
+          queries << "DELETE FROM #{self.repository_association_class.table_name} WHERE repository_id=#{repository.id} AND #{unit_id_field} IN (#{delete_ids.join(', ')})"
         end
 
         unless new_ids.empty?
-          backend_identifier = id_map_for_repository && !id_map_for_repository.empty?
-          self.repository_association_class.import db_columns(backend_identifier), db_values(new_ids, id_map_for_repository, backend_identifier, repository), validate: false
+          self.repository_association_class.import(db_columns_sync, db_values(new_ids, id_map_for_repository, repository), validate: false)
         end
 
         ActiveRecord::Base.transaction do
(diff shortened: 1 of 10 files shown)


jturel:
Looking to test this. I think I can validate the functionality but need help understanding how it fits into the overall picture to make sure I can give good feedback

app/models/katello/concerns/pulp_database_unit.rb
```diff
@@ -178,12 +182,49 @@ def copy_repository_associations(source_repo, dest_repo)
...
         end
         ActiveRecord::Base.connection.execute(delete_query)
-        ActiveRecord::Base.connection.execute(insert_query)
+        ActiveRecord::Base.connection.execute(insert_query) if insert_query
       end
 
       def with_pulp_id(unit_pulp_ids)
         where('pulp_id in (?)', unit_pulp_ids)
       end
+
```
jturel: Throughout this code there is a lot of checking `defined?` and calling `to_sym`. I think it could be avoidable. What if BACKEND_IDENTIFIER_FIELD was a method like `backend_unit_identifier`

```
def backend_identifier_field
  :pulp3_href
end
```

I think it would ultimately improve the code complexity and readability. For example:

```
def db_columns(backend_identifier)
  [self.unit_id_field, self.backend_identifier_field, :repository_id, :created_at, :updated_at].compact
end

def db_columns_copy(....)
  [self.unit_id_field, self.backend_identifier_field, :repository_id].compact
end
```

if we need to keep some `to_sym` calls we can do `[].compact.map(&:to_sym)`

Ruby gives us a lot of tools to avoid conditionals. Less code duplication, and fewer paths through the code.

app/models/katello/concerns/pulp_database_unit.rb
```diff
@@ -178,12 +182,49 @@ def copy_repository_associations(source_repo, dest_repo)
...
+        backend_identifier ? [unit_id_field.to_sym, self::BACKEND_IDENTIFIER_FIELD.to_sym, :repository_id, :created_at, :updated_at] : [unit_id_field.to_sym, :repository_id, :created_at, :updated_at]
+      end
+
+      def db_columns_copy
+        (defined? self::BACKEND_IDENTIFIER_FIELD) ? [self.unit_id_field.to_sym, self::BACKEND_IDENTIFIER_FIELD.to_sym, :repository_id] : [self.unit_id_field.to_sym, :repository_id]
+      end
+
+      def db_values_copy(source_repo, dest_repo)
+        db_values = []
+        source_units = self.repository_association_class.where(:repository_id => source_repo.id).to_a #source_repo.send(self::CONTENT_TYPE.pluralize)
```
jturel: The comments at the end of the line can be removed.

Why do we cast `to_a` ?

app/models/katello/concerns/pulp_database_unit.rb
```diff
@@ -178,12 +182,49 @@ def copy_repository_associations(source_repo, dest_repo)
...
+        dest_units = self.repository_association_class.where(:repository_id => dest_repo.id).to_a #dest_repo.send(self::CONTENT_TYPE.pluralize)
+        source_unit_ids = source_units.pluck(unit_id_field.to_sym)
+        dest_unit_ids = dest_units.pluck(unit_id_field.to_sym)
+        new_units = source_units.select { |su| (source_unit_ids - dest_unit_ids).include? su[unit_id_field] }
+        if defined? self::BACKEND_IDENTIFIER_FIELD
```
jturel: if backend_identifier field becomes a method this whole block gets simplified:

```
new_units.each do |unit|
  db_values << [unit[unit_id_field], self.backend_identifier_field, dest_repo.id].compact
end
```

app/models/katello/concerns/pulp_database_unit.rb
```diff
@@ -178,12 +182,49 @@ def copy_repository_associations(source_repo, dest_repo)
...
+        else
+          new_units.each do |unit|
+            import_unit = [unit[unit_id_field], dest_repo.id]
+            db_values << import_unit
+          end
+        end
+        db_values
+      end
+
+      def db_values(new_ids, pulp_id_href_map, backend_identifier, repository)
```
jturel: We can remove the if conditional I think:

```
new_ids.map do |unit_id|
  [unit_id.to_i, pulp_id_href_map.dig(unit_id), repository.id.to_i, Time.now.utc.to_s(:db), Time.now.utc.to_s(:db)].compact
end
```

app/models/katello/concerns/pulp_database_unit.rb
```diff
@@ -120,34 +121,39 @@ def import_for_repository(repository)
...
       def sync_repository_associations(repository, options = {})
         additive = options.fetch(:additive, false)
         pulp_ids = options.fetch(:pulp_ids, nil)
-        associated_ids = with_pulp_id(pulp_ids).pluck(:id) if pulp_ids
-
-        table_name = self.repository_association_class.table_name
-        attribute_name = unit_id_field
-
+        pulp_id_href_map = options.fetch(:pulp_id_href_map, nil)
+        id_map_for_repository, table_name, attribute_name = {}, self.repository_association_class.table_name, unit_id_field
```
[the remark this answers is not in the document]
jturel: It's also not consistent with the surrounding code

jturel:
Code looks good :+1: 

I'll give it a test today!

jturel:
When running `rake katello:reimport` I'm getting a stack trace:
```
Importing Katello::Erratum                                                                                                                                    
rake aborted!                                                                                                                                                 
NoMethodError: undefined method `pulp_units_batch_all' for Katello::Pulp3::Erratum:Class                                                   
Did you mean?  pulp_units_batch_for_repo                                                                                                                      
/home/vagrant/katello/app/models/katello/concerns/pulp_database_unit.rb:92:in `import_all'                                                        
/home/vagrant/katello/lib/katello/tasks/reimport.rake:41:in `block (3 levels) in <top (required)>'
/home/vagrant/katello/lib/katello/tasks/reimport.rake:39:in `each'
/home/vagrant/katello/lib/katello/tasks/reimport.rake:39:in `block (2 levels) in <top (required)>'
```

sjha4:
> When running `rake katello:reimport` I'm getting a stack trace:
> 
> ```
> Importing Katello::Erratum                                                                                                                                    
> rake aborted!                                                                                                                                                 
> NoMethodError: undefined method `pulp_units_batch_all' for Katello::Pulp3::Erratum:Class                                                   
> Did you mean?  pulp_units_batch_for_repo                                                                                                                      
> /home/vagrant/katello/app/models/katello/concerns/pulp_database_unit.rb:92:in `import_all'                                                        
> /home/vagrant/katello/lib/katello/tasks/reimport.rake:41:in `block (3 levels) in <top (required)>'
> /home/vagrant/katello/lib/katello/tasks/reimport.rake:39:in `each'
> /home/vagrant/katello/lib/katello/tasks/reimport.rake:39:in `block (2 levels) in <top (required)>'
> ```

Filed an issue to deprecate this task and skip for pulp3 content types: https://projects.theforeman.org/issues/29209

app/models/katello/concerns/pulp_database_unit.rb
```diff
@@ -120,34 +125,38 @@ def import_for_repository(repository)
...
-        pulp_ids = options.fetch(:pulp_ids, nil)
-        associated_ids = with_pulp_id(pulp_ids).pluck(:id) if pulp_ids
-
-        table_name = self.repository_association_class.table_name
-        attribute_name = unit_id_field
-
+        pulp_ids = options.dig(:pulp_ids)
+        pulp_id_href_map = options.dig(:pulp_id_href_map)
+        id_map_for_repository = {}
+        if pulp_ids
```
jturel: when would this be false?

app/models/katello/concerns/pulp_database_unit.rb
```diff
@@ -175,15 +181,41 @@ def copy_repository_associations(source_repo, dest_repo)
...
+
+      def db_columns_copy
+        [self.unit_id_field.to_sym, backend_identifier_field, :repository_id].compact
+      end
+
+      def db_values_copy(source_repo, dest_repo)
+        db_values = []
+        source_units = self.repository_association_class.where(:repository_id => source_repo.id)
+        dest_units = self.repository_association_class.where(:repository_id => dest_repo.id)
+        source_unit_ids = source_units.pluck(unit_id_field.to_sym)
```
jturel: We actually don't need to call to_sym!

app/models/katello/concerns/pulp_database_unit.rb
```diff
@@ -175,15 +181,41 @@ def copy_repository_associations(source_repo, dest_repo)
...
+        dest_unit_ids = dest_units.pluck(unit_id_field.to_sym)
+        new_units = source_units.select { |su| (source_unit_ids - dest_unit_ids).include? su[unit_id_field] }
```
jturel: Would it be better to move `(source_unit_ids - dest_unit_ids)` into a variable out of the select loop?

```
missing_unit_ids = source_unit_ids - dest_unit_ids
new_unit_db_values = source_units.map do |su|
  next unless missing_unit_ids.include?(su[unit_id_field])
  [unit[unit_id_field], unit[backend_identifier_field], dest_repo.id].compact
end
new_unit_db_values.compact
```

jturel:
Looks good. I imported a repo with 3 errata and got this:

```
 #<Katello::RepositoryErratum:0x0000000013d306b8
  id: 5,
  erratum_id: 5,
  repository_id: 2,
  created_at: Thu, 27 Feb 2020 20:55:06 UTC +00:00,
  updated_at: Thu, 27 Feb 2020 20:55:06 UTC +00:00,
  erratum_pulp3_href: "/pulp/api/v3/content/rpm/advisories/2d4c2186-7b63-41b1-9efc-202d3a3a3c67/">,
 #<Katello::RepositoryErratum:0x0000000013d30578
  id: 6,
  erratum_id: 6,
  repository_id: 2,
  created_at: Thu, 27 Feb 2020 20:55:06 UTC +00:00,
  updated_at: Thu, 27 Feb 2020 20:55:06 UTC +00:00,
  erratum_pulp3_href: "/pulp/api/v3/content/rpm/advisories/bb1cf835-a219-4824-bc5b-eaf48b59cc82/">,
 #<Katello::RepositoryErratum:0x0000000013d30438
  id: 7,
  erratum_id: 7,
  repository_id: 2,
  created_at: Thu, 27 Feb 2020 20:55:06 UTC +00:00,
  updated_at: Thu, 27 Feb 2020 20:55:06 UTC +00:00,
  erratum_pulp3_href: "/pulp/api/v3/content/rpm/advisories/d614b64c-4c08-4b3c-8c5f-8ba0a7ed2da3/">

Left a couple notes on the code :)

```

app/models/katello/concerns/pulp_database_unit.rb
jturel: Hmm. Is there a way to find the missing units with a single active record query rather than running two queries and taking the difference?

app/models/katello/concerns/pulp_database_unit.rb
sjha4: I believe when a content unit gets deleted in a repo version upon sync..

sjha4:
@jturel : Updated as per your suggestions..

app/models/katello/concerns/pulp_database_unit.rb
```diff
@@ -175,15 +181,40 @@ def copy_repository_associations(source_repo, dest_repo)
...
+
+      def db_columns_sync
+        [unit_id_field, backend_identifier_field, :repository_id, :created_at, :updated_at].compact
+      end
+
+      def db_columns_copy
+        [unit_id_field, backend_identifier_field, :repository_id].compact
+      end
+
+      def db_values_copy(source_repo, dest_repo)
```
jturel: So it looks like this method is ultimately trying to find the units in the source repo that are missing from the destination repo for copying?

Would a query like this find those missing units?

`missing_units = self.repository_association_class.where(repository: source_repo).where.not(repository: dest_repo)`

Then this method just returns:
```
missing_units.map do |unit|
  [unit.unit_id_field, unit.backend_identifier_field, dest_repo.id].compact
end
```

jturel: I guess the query would need to consider the unit_id_field as well

sjha4: Made a couple of changes based on this..The query above works great and using it..While testing that realized backend_identifier_field is a method invocation causing sql queries of it's own..So moving both invocations outside loop..

app/models/katello/concerns/pulp_database_unit.rb
jturel: ACK. The way this code is (and was) written is that it assumes that `associated_ids` will always be defined on this line: 

```  
new_ids = associated_ids - existing_ids
```

This is making me think about how to clean up this code because it is a little hard to follow due to the organization and some of the data manipulation going on. I think the time is now for a refactor of the method. Let me know what you think about this. Do you agree with it? Does it still work?

```
     def sync_repository_associations(repository, options = {})                                                                                                                                                                                                                
        additive = options.fetch(:additive, false)                                                                                                                                                                                                                              
        pulp_ids = options.fetch(:pulp_ids, []) # seems like we always give pulp_ids, ie it is required                                                                                                                                                                         
                                                                                                                                                                                                                                                                                
        existing_ids = self.repository_association_class.uncached do                                                                                                                                                                                                            
          self.repository_association_class.where(:repository_id => repository).pluck(unit_id_field)                                                                                                                                                                            
        end                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                
        ids_for_repository = with_pulp_id(pulp_ids).pluck(:id, :pulp_id)                                                                                                                                                                                                        
        associated_ids = ids_for_repository.map(&:first)                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                
        unless additive                                                                                                                                                                                                                                                         
          delete_ids = existing_ids - associated_ids                                                                                                                                                                                                                            
          ActiveRecord::Base.connection.execute(                                                                                                                                                                                                                                
            "DELETE FROM #{self.repository_association_class.table_name}                                                                                                                                                                                                        
            WHERE repository_id=#{repository.id}                                                                                                                                                                                                                                
            AND #{unit_id_field} IN (#{delete_ids.join(', ')})"                                                                                                                                                                                                                 
          ) if delete_ids.any?                                                                                                                                                                                                                                                  
        end                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                
        new_ids = associated_ids - existing_ids                                                                                                                                                                                                                                 
        unless new_ids.empty?                                                                                                                                                                                                                                                   
          id_map_for_repository = {}                                                                                                                                                                                                                                            
          pulp_id_href_map = options.fetch(:pulp_id_href_map, {})                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                
          if pulp_id_href_map.any?                                                                                                                                                                                                                                              
            ids_for_repository.map do |pair|                                                                                                                                                                                                                                    
              # pair[0] is the unit id                                                                                                                                                                                                                                          
              # pair[1] is the pulp_id                                                                                                                                                                                                                                          
              id_map_for_repository[pair[0]] = pulp_id_href_map[pair[1]]                                                                                                                                                                                                        
            end                                                                                                                                                                                                                                                                 
          end                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                
          self.repository_association_class.import(db_columns_sync, db_values(new_ids, id_map_for_repository, repository), validate: false)                                                                                                                                     
        end                                                                                                                                                                                                                                                                     
      end
```

app/models/katello/concerns/pulp_database_unit.rb
```diff
@@ -119,71 +124,87 @@ def import_for_repository(repository)
...
-          inserts = new_ids.map { |unit_id| "(#{unit_id.to_i}, #{repository.id.to_i}, '#{Time.now.utc.to_s(:db)}', '#{Time.now.utc.to_s(:db)}')" }
-          queries << "INSERT INTO #{table_name} (#{attribute_name}, repository_id, created_at, updated_at) VALUES #{inserts.join(', ')}"
-        end
-
-        ActiveRecord::Base.transaction do
-          queries.each do |query|
-            ActiveRecord::Base.connection.execute(query)
-          end
+        unless (new_ids = associated_ids - existing_ids).empty?
+          self.repository_association_class.import(db_columns_sync, db_values(new_ids, id_href_map_for_repository, repository), validate: false)
```
jturel: I realized that the delete and imports are happening in two separate transactions now. Before they were happening together. Should we wrap these in a single transaction?

sjha4: Looking at this, do you think wrapping both conditional blocks in a transaction would make sense or is that too big of a block?
```
ActiveRecord::Base.transaction do
  if !additive && (delete_ids = existing_ids - associated_ids).any?
    query = "DELETE FROM #{self.repository_association_class.table_name} WHERE repository_id=#{repository.id} AND #{unit_id_field} IN (#{delete_ids.join(', ')})"
    ActiveRecord::Base.connection.execute(query)
  end
  unless (new_ids = associated_ids - existing_ids).empty?
    self.repository_association_class.import(db_columns_sync, db_values(new_ids, id_href_map_for_repository, repository), validate: false)
  end
end
```

jturel: Should be quick operations so I am OK with wrapping those in the transaction block

sjha4:
[test katello]

jturel:
After the updates: deleted my repo, created and synced, results look good!

```
=> [#<Katello::RepositoryErratum:0x00000000105266a0
  id: 8,
  erratum_id: 5,
  repository_id: 3,
  created_at: Thu, 05 Mar 2020 02:56:00 UTC +00:00,
  updated_at: Thu, 05 Mar 2020 02:56:00 UTC +00:00,
  erratum_pulp3_href: "/pulp/api/v3/content/rpm/advisories/2d4c2186-7b63-41b1-9efc-202d3a3a3c67/">,
 #<Katello::RepositoryErratum:0x0000000010526218
  id: 9,
  erratum_id: 6,
  repository_id: 3,
  created_at: Thu, 05 Mar 2020 02:56:00 UTC +00:00,
  updated_at: Thu, 05 Mar 2020 02:56:00 UTC +00:00,
  erratum_pulp3_href: "/pulp/api/v3/content/rpm/advisories/bb1cf835-a219-4824-bc5b-eaf48b59cc82/">,
 #<Katello::RepositoryErratum:0x0000000010525de0
  id: 10,
  erratum_id: 7,
  repository_id: 3,
  created_at: Thu, 05 Mar 2020 02:56:00 UTC +00:00,
  updated_at: Thu, 05 Mar 2020 02:56:00 UTC +00:00,
  erratum_pulp3_href: "/pulp/api/v3/content/rpm/advisories/d614b64c-4c08-4b3c-8c5f-8ba0a7ed2da3/">]

```

sjha4:
Thanks @jturel! Merging .. :smile:

PR #8554, opened 2020-02-04
Fixes #26878 - Adds Deprecation Warning testing

johnpmitsch:
This adds a central way to handle deprecation warnings and tests them against the Katello version to ensure they are removed.

### DIFF
--- a/app/controllers/katello/api/v2/host_subscriptions_controller.rb
+++ b/app/controllers/katello/api/v2/host_subscriptions_controller.rb
@@ -158,7 +158,7 @@ module Katello
     def content_override
       content_overrides = []
       if params[:content_label]
-        ::Foreman::Deprecation.api_deprecation_warning("The parameter content_label will be removed in Katello 3.5. Please update to use the content_overrides parameter.")
+        ::Katello::Deprecation.api_deprecation_warning("host_subs_content_label", caller(0))
         content_override_params = {:content_label => params[:content_label]}
         value = params[:value]
         if value == "default"
--- a/app/services/katello/deprecation.rb
+++ b/app/services/katello/deprecation.rb
@@ -1,0 +1,28 @@
+module Katello
+  class Deprecation
+    class << self; attr_accessor :deprecations end
+
+    # Structure of Deprecation warning object:
+    # reference_key : {
+    #   removal_version: Katello version that the deprecated behavior will be removed by. This value is used in
+    #                    testing to compare to the current Katello version, failing if it is less than the current version.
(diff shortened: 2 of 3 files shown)


jlsherrill:
I'm not a big fan of this approach for a couple reasons:

1.  I'm generally not a fan of a 'registry' of strings that get referenced elsewhere in code.  If some registry entry got removed without the corresponding place, it might be easy to forget about it (although  hopefully a test would catch it)
2.  passing in caller(0) isn't super great and could easily get forgotten, although there may be a solution to that

Imagine something like this:
```
Katello::Deprecation.warning("3.15", _("This thing is being removed, please use this other thing")
```
and then write a source code test that would catch it, similar to what is in https://github.com/Katello/katello/blob/master/test/source_code_test.rb

I see this as similar to translations that pick up on _() without a similar 'registry'

Thoughts?

johnpmitsch:
@jlsherrill its a fair point that we would have to remove two entries, the registry entry and the reference in the source code, and that is cause for concern.

> and then write a source code test that would catch it, similar to what is in https://github.com/Katello/katello/blob/master/test/source_code_test.rb

Could you expand on this part? I looked at that file initially but am not sure how this kind of test would work.

jlsherrill:
Here's an example, it might could be cleaned up a tad:

```
  describe 'foo' do
    it 'has no deprecation warnings it should not have' do
      SourceCode.new('**/*.{rb}').check_lines("remove these deprecations!") do |line|
        if line =~ /Katello::Deprecation\.warning/
          version = line[/\d+\.\d+/]
          assert version, "Could not determine removal version"
          Gem::Version.new(version) > Gem::Version.new(Katello::VERSION)
        else
          true
        end
      end
    end
  end
```

When adding this:
```
          Katello::Deprecation.warning('3.12', 'do something')
```
to a file, i now get:
```
  1) Failure:
SourceCodeTest::foo#test_0001_has no deprecation warnings it should not have [../katello/test/source_code_test.rb:48]:
remove these deprecations!
check lines:
 - /home/vagrant/git/katello/test/services/katello/pulp3/migration_test.rb:26: Katello::Deprecation.warning('3.12', 'do something').
Expected [["          Katello::Deprecation.warning('3.12', 'do something')\n", 26, "/home/vagrant/git/katello/test/services/katello/pulp3/migration_test.rb"]] to be empty.
```

The entire test run took about ~3 seconds

johnpmitsch:
closing in favor of https://github.com/theforeman/tool_belt/pull/237

PR #8556, opened 2020-02-07
Fixes #28958 - disable not inited lifecycle-button

m-bucher:
Keep 'Create Environment Path'-button disabled until it is initialized


sbernhard:
[test katello]

PR #8558, opened 2020-02-10
Bump master version to 3.16

jjeffers:


ekohl:
For future reference, you should only do this after packaging has been branched. Right now we can't build any katello nightly packages and this creates more work for branching.

jturel:
> For future reference, you should only do this after packaging has been branched. Right now we can't build any katello nightly packages and this creates more work for branching.

Didn't think of that. My mistake!

PR #8559, opened 2020-02-10
Bumps version for 3.15 RC1, adds changelog.

jjeffers:


jturel:
I think we may be stuck here since Foreman hasn't branched yet, and foreman-infra will need an update: https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy

jjeffers:
[test katello]

PR #8561, opened 2020-02-11
Fixes #28988 - Purge modular errata

parthaa:
This commit properly purges modular errata.
Before this fix the purge_partial_errata would check on the packages
belonging to the errata and if they matched will keep the errata in the
destination. However with modular errata we need to inaddition check if
the module exists in the destination repo before copying the errata
over. This commit handles that appropriately.

### DIFF
--- a/app/models/katello/repository.rb
+++ b/app/models/katello/repository.rb
@@ -301,17 +301,53 @@ module Katello
         group("#{errata}.id").count
     end
 
+    def  self.errata_with_module_stream_counts(repo)
+      repository_errata = Katello::RepositoryErratum.table_name
+      errata = Katello::Erratum.table_name
+      erratum_package = Katello::ErratumPackage.table_name
+      repository_module_stream = Katello::RepositoryModuleStream.table_name
+      module_stream = Katello::ModuleStream.table_name
+      msep = ::Katello::ModuleStreamErratumPackage.table_name
+      ::Katello::Erratum.joins(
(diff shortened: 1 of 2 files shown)


johnpmitsch:
@parthaa  I'm having trouble reproducing the original issue. I synced the repo attached to the issue and attached to a CV, but my errata filter successfully filters the errata, my understanding is it should not filter correctly.

Here is my filter:
Screenshot from 2020-02-18 09-56-03

and (while on master) I am getting the count of 9 Packages, 3 Module Streams in the published CVV, suggesting the filter actually did filter out the errata, which isn't the result of "The excluded errata still persists in the destination repo". 

Anything I should be doing differently?

parthaa:
@johnpmitsch ok I did this

1. synced appstream from staging
1. Created a filter for excluding this errata -> RHBA-2019:49742
1. Publish the and notice that the version still had RHBA-2019:49742
1. Delete the excludes filter
1. Create a errata date exclude filters with "exclude after august 2019"
1. Publish again
1. Notice that you 'll see quite a few errata after 08/2019 in the published version
1. Try the same with my PR and you should notice a difference.

johnpmitsch:
@parthaa Everything works well! can we look into adding testing?

parthaa:
> Everything works well! can we look into adding testing?

Added

johnpmitsch:
[test katello]

PR #8562, opened 2020-02-12
Refs #28818 - Fix sync_summary select

m-bucher:
Today I learned: If you want something done right/fast, do it in SQL :wink: 

This should fix the last performance problem mentioned in https://github.com/Katello/katello/pull/8536

### DIFF
--- a/app/models/katello/glue/pulp/repos.rb
+++ b/app/models/katello/glue/pulp/repos.rb
@@ -107,12 +107,25 @@ module Katello
       end
 
       def last_repo_sync_task_by_repoid
-        latest_tasks = {}
-        last_repo_sync_tasks.each do |t|
-          repoid = t.input["repository"]["id"]
-          latest_tasks[repoid] = t unless latest_tasks.key?(repoid) && latest_tasks[repoid].started_at > t.started_at
-        end
-        latest_tasks
+        all_repos = repos(self.library, nil, false)
+        base_combined_table = ForemanTasks::Task::DynflowTask
+          .where(:label => ::Actions::Katello::Repository::Sync.name)
+          .joins(:locks)
+          .where("foreman_tasks_locks.resource_id in (?) and foreman_tasks_locks.resource_type = ?", all_repos.pluck(:id), ::Katello::Repository.name)
+
+        max_per_repoid = ForemanTasks::Task::DynflowTask
+          .joins(:locks)
+          .select("#{ForemanTasks::Task::DynflowTask.table_name}.*, inner_select.resource_id")
+          .joins(
+            "INNER JOIN (#{base_combined_table
+              .select("MAX(foreman_tasks_tasks.started_at) AS started_at", "foreman_tasks_locks.resource_id AS resource_id")
+              .group("foreman_tasks_locks.resource_id")
+              .to_sql}) inner_select
+            ON inner_select.started_at = #{ForemanTasks::Task::DynflowTask.table_name}.started_at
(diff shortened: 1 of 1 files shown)


chris1984:
Starting to test

chris1984:
Works good, thanks @m-bucher

app/models/katello/glue/pulp/repos.rb
```diff
@@ -107,12 +107,25 @@ def last_repo_sync_task_group
       end
 
       def last_repo_sync_task_by_repoid
```
jturel: @m-bucher @chris1984 

This method is very troublesome to me. My thoughts:

- We should be extremely deliberate when breaking away from typical active record queries. There is a time and place for it, though it could mean that there is a larger issue to be addressed needing design.

- We should not expose so much of the internal details of the ForemanTasks schema to Katello's own code. This is very brittle and I would not consider it a good practice.

- When writing SQL queries by hand we should limit their scope to Katello's own schema.

I see how this solution was arrived upon. There was already a (smaller) query referencing the Foreman Tasks schema:

```
last_repo_sync_tasks.where("#{ForemanTasks::Task::DynflowTask.table_name}.started_at > '%s'", started_after.utc).uniq
```

I guess it was tempting to build on it, but now we are coupled to ForemanTasks schema even closer.

It looks like this code should be moved to ForemanTasks by exposing and API that will be supported and tested. I mean a Ruby API, not REST. It would remove all knowledge of the internals out of Katello so that this code does not become a burden for us later.

m-bucher: I get that. For now this was the fastest way to get it running.

I agree, it would be better to add something like a `get_last_task_for_resource_id()` method to foreman-tasks, where others may also profit from it (I believe it is not an uncommon task).

@jturel Can you point me to the place in foreman-tasks, where it would make sense to add such a method, so I may work on it?

jturel: I totally understand and we are always grateful for your contributions.

As far as where in foreman-tasks this goes I am helpless!

@adamruzicka can you guide us to the right place in foreman-tasks to provide a ruby method to approximate the query here?

adamruzicka: I'd just put it to https://github.com/adamruzicka/foreman-tasks/blob/master/app/models/foreman_tasks/task.rb .

A couple of notes:
- How many repos are we trying to find the last task for? If it is a smaller number, I'd be in favor of having an api for "get last task for a SINGLE resource" and just run that several times. It would make the code much simpler

- maybe [Task.search_by_generic_resource](https://github.com/adamruzicka/foreman-tasks/blob/master/app/models/foreman_tasks/task/search.rb#L4) could be expanded to properly handle searches with `IN`. This way, `base_combined_table` could be rewritten as:
```
ids = all_repos.pluck(:id).join(',')
label = ::Actions::Katello::Repository::Sync.name
type = Katello::Repository.name
ForemanTasks::Task.search_for("label = #{label} and resource_type = #{type} and resource_id ^ (#{ids})")
```

m-bucher: > How many repos are we trying to find the last task for? If it is a smaller number, I'd be in favor of having an api for "get last task for a SINGLE resource" and just run that several times. It would make the code much simpler

In most (sane) cases I would expect not more than 10, but in theory it is not limited.
If someone wants to have 100+ repositories in one product, it can be done as far as I know.

PR #8431, opened 2019-11-12
POC for artemis events

jturel:
**To test with external broker**

Set up candlepin to look for artemis externally https://www.candlepinproject.org/docs/candlepin/external_artemis_setup.html
- For the broker.xml use  https://gist.github.com/jturel/bc0d428b473ed5d6adfb8adc08408184
- Restart tomcat

**To test with candlepin's embedded broker**
- Replace `/var/lib/tomcat/webapps/candlepin/WEB-INF/classes/broker.xml` with https://gist.github.com/jturel/bc0d428b473ed5d6adfb8adc08408184
- Restart tomcat

Artemis supports last value queues which prevents dup messages from residing in a queue simultaneously: https://activemq.apache.org/components/artemis/documentation/2.4.0/last-value-queues.html This a key feature in the katello events side maintaining parity with existing functionality.

**TODO**
[ ] RFC
[ ] packaging for stomp gem
[ ] removal of qpid-messaging gem packaging
[ ] puppet-katello updates to land broker.xml
[ ] puppet-candlepin updates to point to external broker.xml
[ ] squash before merging

PR #8563, opened 2020-02-12
Fixes #29000 - Consume Candlepin events via STOMP

jturel:
This PR is the result of my POC (#8431) and [RFC](https://community.theforeman.org/t/rfc-goodbye-qpid-hello-artemis-evolving-messaging-in-katello/17016) to implement the Artemis message broker and consume Candlepin messages via the STOMP protocol.

**To test**
Spin up a new (second) development environment using the related installer changes so there's zero-configuration other than checking out this PR once provisioned. Use a box definition like this:

```
centos7-katello-artemis:                                                                                                                                                                        
  box: centos7                                                                                                                                                                                
  ansible:                                                                                                                                                                                    
    playbook: 'playbooks/katello_devel.yml'                                                                                                                                                   
    group: 'devel'                                                                                                                                                                            
    variables:                                                                                                                                                                                
      ssh_forward_agent: true                                                                                                                                                                 
      foreman_installer_module_prs:                                                                                                                                                           
        - theforeman/candlepin/143                                                                                                                                                            
        - theforeman/certs/275                                                                                                                                                                
        - theforeman/katello/327                                                                                                                                                              
        - theforeman/katello_devel/224
```

At some point the installer will fail because the candlepin version in nightly (as of 4/16) is not 3.10. At this point install both RPMs here: http://koji.katello.org/koji/buildinfo?buildID=57101 and run `vagrant provosion` on the box. Installer should complete now. After checking out this PR you'll need to `bundle install` from the foreman dir in order to bring in the stomp gem.

The first request to the server should indicate the the candlepin_event queue has been subscribed to:

```
2020-02-13T19:53:10 [I|app|b3ff7cd3] Katello event daemon started process=31288
2020-02-13T19:53:10 [I|app|b3ff7cd3] Processing by Katello::Api::V2::KatelloPingController#index as */*
2020-02-13T19:53:10 [I|app|b3ff7cd3]   Parameters: {"api_version"=>"v2"}
2020-02-13T19:53:10 [I|app|b3ff7cd3] Subscribed to katello.candlepin.candlepin_events
2020-02-13T19:53:10 [I|kat|b3ff7cd3] Polling Katello Event Queue
```

From here the goal is to make sure that Candlepin events are processed as normal. The filtering block in /etc/candlepin/broker.xml determines which messages will reach our queue to be handled.

- Pool create and delete events are tested by importing and deleting a manifest
- Entitlement create and delete are tested by adding and removing subscriptions from a content host
- Compliance events are tested by adding/removing subscriptions or changing system purpose on a content host

There's debug logging for the received events:

```
2020-02-13T20:07:44 [D|kat|b3ff7cd3] message received from subscriptions queue 
2020-02-13T20:07:44 [D|kat|b3ff7cd3] message subject: compliance.created
```

Finally, the listener must be resilient. Get creative and try to break it (randomly restarting tomcat for example), and study the code for weaknesses. If tomcat is down the rails server should continue to serve requests - ie it will not hang indefinitely waiting for the event listener to start. But once tomcat comes back up, you should see that it once again subscribes and handles messages.

**Design**
My main goal was to maintain the behaviors we had with the qpid implementation in terms of resiliency and the messages that reach the listener. I've also created a pattern so that the listener's connection is set up with configuration settings in engine.rb This makes the listener completely unaware of the implementation details - it just subscribes and handles messages. 

The code also makes it easy to add new types of messaging connections so this code will not have to be rewritten from the ground up if we change message brokers in the future. New connection types can be added and instantiated using the `Katello::Messaging::Connection` class. If we can't extract qpid's arrow from our hearts, we can add a new 'QpidConnection' for example.


ianballou:
I've tested this out and I haven't been able to break it so far! I'm not 100% done studying the code yet.

app/services/katello/candlepin_event_listener.rb
```diff
@@ -2,81 +2,66 @@ module Katello
...
     @processed_count = 0
 
-    def self.run
-      initialize_listening_service
+    def self.logger
+      ::Foreman::Logging.logger('katello/candlepin_events')
+    end
+
+    def self.running?
+      @running == true && client.running?
```
ianballou: What happens if these two are in disagreement? That sounds like some kind of failure condition.

jturel: You're exactly right. If this method returns false then the event daemon will restart the CandlepinEventListener. See Katello::EventDaemon::Monitor for how that's handled.

This would happen in a situation where the connection is somehow closed or if the listening thread in the stomp client dies, leaving an open connection but a state where messages are no longer handled

ianballou:
Looked at the code and I don't see any issues! I'll revisit and finish the review when this PR is unblocked by the remaining work.

ekohl:
There's 2 merge conflicts now. Is there anything this is blocked on? Anything on the platform side?

ehelms:
@ekohl It's in collaboration with this installer change -- https://github.com/theforeman/puppet-candlepin/pull/143

lib/katello/engine.rb
```diff
@@ -48,6 +48,13 @@ class Engine < ::Rails::Engine
           :oauth_secret => 'katello',
           :ca_cert_file => nil,
           :bulk_load_size => 1000
+        },
+        :candlepin_events => {
+          :broker_host => 'localhost',
+          :broker_port => 61_613,
```
ehelms: Is the underscore intentional?

ehelms:
Based on our discussions, I think this would need configuration to indicate what certificates to use when communicating with Artemis. Which I would then assume end up here https://github.com/theforeman/puppet-katello/pull/327/files#diff-c48970cf4dfa5d2a2124a59cac2bf934 as parameters that we configure.

lib/katello/engine.rb
jturel: I just do what rubocop tells me to do like a good programmer ;)

https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/NumericLiterals

jturel:
> Based on our discussions, I think this would need configuration to indicate what certificates to use when communicating with Artemis. Which I would then assume end up here https://github.com/theforeman/puppet-katello/pull/327/files#diff-c48970cf4dfa5d2a2124a59cac2bf934 as parameters that we configure.

Agreed!

jturel:
@ianballou this is ready to test again. I've updated the testing steps and that should make the setup simpler but it does involve setting up a second vm. Not much has changed code-wise except to enable SSL and to ensure resilience from outages.

Let me know if I can help with anything!

jturel:
Thanks @ianballou !!

PR #8567, opened 2020-02-17
Fixes #29029 - set debian repository architecture

sbernhard:


app/lib/actions/helpers/content_helper.rb
```diff
@@ -0,0 +1,13 @@
...
+      def format_arches(repo)
+        if repo.content_type == ::Katello::Repository::DEB_TYPE
+          repo.deb_architectures
+        else
+          repo.arch == "noarch" ? nil : repo.arch
+        end
+      end
+    end
+  end
+end
```
jlsherrill: would it make sense to just put this method on the root_repository model itself?  Then you could just call root.format_aches  ind and not have to include a module where you want to use it?

sbernhard: Sure. I will move it to the root repo. Thanks!

PR #8573, opened 2020-02-19
Fixes #28952 - allow srpm content pulpcore uploads from hammer

jjeffers:
This change allows for SRPM content type uploads into yum repositories.

To test, upload an srpm via hammer. You can verify the package was uploaded via hammer as well as examining the puilp schema.
```
[vagrant@centos7-hammer-devel ~]$ hammer product create --name test_srpm --organization-id 1
Product created.

[vagrant@centos7-hammer-devel ~]$ hammer repository create --name test_srpm_content --product-id 1 --content-type yum
Repository created.

[vagrant@centos7-hammer-devel ~]$ hammer repository  upload-content --id 1 --path ./test-srpm01-1.0-1.src.rpm --content-type srpm
Successfully uploaded file 'test-srpm01-1.0-1.src.rpm'

[vagrant@centos7-hammer-devel ~]$ hammer repository info --id 1
ID:                 1
Name:               test_srpm_content
Label:              test_srpm_content
[...]
Content Counts:     
    Packages:       0
    Source RPMS:    1
    Package Groups: 0
    Errata:         0
    Module Streams: 0
```

The pulp schema:
```
pulp=# select name, arch from rpm_package;
    name     | arch 
-------------+------
 test-srpm01 | src
(1 row)
```
This same information should be reflected in the web UI as well.
Finally, pulp2 srpm uploads via hammer should work just as well after this change.


jjeffers:
> @jjeffers is this for pulp 2 or 3? I see a test for 3 but should we test for both?

@chris1984 This is just for pulpcore, so I don't recall testing pulp2 specifically.

jjeffers:
> @jjeffers is this for pulp 2 or 3? I see a test for 3 but should we test for both?

You can test for both. I double checked with hammer uploads and it should world for both pulp2 and pulpcore (srpm content).

app/lib/actions/katello/repository/filtered_index_content.rb
```diff
@@ -23,10 +23,10 @@ def run
             unit_ids = search_units(repo)
             ::Katello::Deb.import_all(unit_ids, repo)
           elsif repo.yum?
-            unit_ids = search_units(repo)
-            if input[:content_type] == 'srpm'
-              ::Katello::Srpm.import_all(unit_ids, repo)
+            if input[:content_type] == ::Katello::Srpm::CONTENT_TYPE
+              ::Katello::Srpm.import_for_repository(repo)
```
jlsherrill: why did you decide to move out the unit_ids lookup?  This causes the repository to be fully indexed (for srpms)

app/services/katello/pulp3/content.rb
```diff
@@ -5,17 +5,29 @@ class Content
...
         end
 
+        def content_href_by_checksum(content_type, content_list_results, checksum, content_api)
+          if content_type == FileUnit::CONTENT_TYPE
+            content_unit_href = content_list_results.first.pulp_href unless content_list_results.empty?
+            return {"content_unit_href" => content_unit_href} if content_unit_href
+          else
+            content_list_results.each do |package|
+              package_details = content_api.read(package.pulp_href)
+              return {"content_unit_href" => package.pulp_href} if package_details.pkg_id == checksum
```
jlsherrill: Is this with or without the changes that the pulp team was going to add?

I don't quite get the logic here, its fetching all rpms with the same checksum, passing that to this fucntion, then this is looping over the results and fetching each full unit (is that required?), but then its comparing the pkg_id to the expected checksum, what is pkg_id here?

jjeffers: Currently there is no "digest" option for other plugins, so what you get is a list of all packages. There is a request for pulp_rpm to add the ability to filter a list by checksum, or an equivalent, but it doesn't exist at the moment.

For large package lists, this seems like a workaround that will become cumbersome. I couldn't think of another way to do this. Hopefully the pulp team will add this and we can refactor to take advantage of it.

For each ContentPackage object returned, the `pkg_id` is the checksum (according the RpmPackage documentation).

app/lib/actions/katello/repository/filtered_index_content.rb
jjeffers: I can't recall why I did that, but I think it was because of some issues with the unit search and pulpcore.

jlsherrill: actually looking at this, it seems hard coded to go straight to pulp2.  I don't think this was every properly converted to handle rpms for pulp3 either.    I think we need to refactor this quite a bit.  The caller of this (FinishUpload) passes in a filter for pulp2 that fetches all the RPMs/Srpms uploaded in the past 5 minutes.  But this won't work with pulp3, in fact pulp3 provides the href for uploaded contents, so we shouldn't really need to do this anyways.  We should be able to take the provided hrefs from the upload and pass it into this (or into a pulp3 related action).   I'm a bit skeptical that rpm uploads work properly tbh!

app/services/katello/pulp3/content.rb
jlsherrill: so i guess the call to `content_backend_service.content_api.list("digest": checksum)` just returns all rpms?    I'm not sure that'll work properly since its not paged.  I'd kinda be in favor of just letting this fail for duplicate uploads until the pulp team adds digest support.  It would simplify it a bit too

app/lib/actions/katello/repository/filtered_index_content.rb
jjeffers: Right, as of right now, there is only the pulp2 path for unit_search. 

And yup, completely forgot to account for paging in the list operation!

jjeffers:
Speaking of missing digest options, https://github.com/Katello/katello/blob/2a3aea0a655db2a51277552d60fec01b43791bcc/app/lib/actions/pulp3/orchestration/repository/upload_content.rb#L9 uses the same filter options. But this only is effective for pulp_file at the moment.

app/lib/actions/katello/repository/filtered_index_content.rb
jlsherrill: and this is still waiting on an update to try to pass in the unit href's, correct?

jjeffers: Yes, if we can extract a valid content_unit_href/pulp_id, then we should be able to use `import_all`.

jjeffers: The alternative is to refactor the unit search to have both pulp2/pulpcore solutions.

app/lib/actions/katello/repository/filtered_index_content.rb
```diff
@@ -23,10 +23,15 @@ def run
             unit_ids = search_units(repo)
             ::Katello::Deb.import_all(unit_ids, repo)
           elsif repo.yum?
-            unit_ids = search_units(repo)
-            if input[:content_type] == 'srpm'
+            if input[:dependency] && input[:dependency][:content_unit_href]
```
jlsherrill: lets not use 'dependency' here since we're actually using data from it.   Soething like 'upload_task' for a name would be a lot more clear

app/lib/actions/pulp3/repository/commit_upload.rb
```diff
@@ -0,0 +1,22 @@
...
+    module Repository
+      class CommitUpload < Pulp3::AbstractAsyncTask
+        def plan(repository, smart_proxy, upload_href, sha256)
+          plan_self(:repository_id => repository.id, :smart_proxy_id => smart_proxy.id, :upload_href => upload_href, :sha256 => sha256)
+        end
+
+        def invoke_external_task
+          if input[:skip]
+            output[:pulp_tasks] = nil
+          else
```
jlsherrill: could this just be re-written as  'unless input[:skip]' ?  i'm not sure setting output[:pulp_tasks] to nil is really worth anything?

jlsherrill: also i don't even see skip being passed into plan_self, so i'm not sure how it would ever be set?

app/lib/actions/katello/repository/filtered_index_content.rb
jlsherrill: so to clarify, i'd rename what gets passed in as well

jlsherrill:
Code looks good, i think samir is going to do some manual testing.

jlsherrill:
@jjeffers do you mind openeing a followup issue about coming and fixing the already uploaded case, once we're at a newer pulp version?

jjeffers:
> @jjeffers do you mind openeing a followup issue about coming and fixing the already uploaded case, once we're at a newer pulp version?

Sure thing :+1:

sjha4:
Other than the duplicate file upload issue, everything else worked for me..Also, realized that while yum sync is additive, file sync is not so when you re-sync a repo after uploading content to it, the uploaded content just disappears..I wonder if we have an issue tracking that..

PR #8574, opened 2020-02-20
Fixes #29094 - Prevent setting auto-attach on Activation Key when in Simple Content Access

jturel:
This change prevents setting the auto-attach preference on an activation key when in simple content access and shows 'Not Applicable' in place of the selection. Per discussion with UX it also changes the preference checkbox to a select dropdown when not in SCA.

This does not affect the behavior of the UI since setting auto-attach to true or false has no positive or negative effect.

This also fixes the changes introduced by #8529 in the scenario when not in SCA. The checkbox was only changed to a dropdown in one place, and the one that was changed did not work at all.

Ask me for a manifest that can be used to test :)

### DIFF
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/activation-keys/details/views/activation-key-subscriptions.html
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/activation-keys/details/views/activation-key-subscriptions.html
@@ -1,21 +1,24 @@
 <span page-title ng-model="activationKey">{{ 'Subscriptions for Activation Key:' | translate }} {{ activationKey.name }}</span>
 
 <section>
-  <h4 translate>Activation Key Type:</h4>
-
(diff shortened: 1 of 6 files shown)


engines/bastion_katello/app/assets/javascripts/bastion_katello/activation-keys/details/views/activation-key-subscriptions.html
```diff
@@ -1,21 +1,24 @@
 <span page-title ng-model="activationKey">{{ 'Subscriptions for Activation Key:' | translate }} {{ activationKey.name }}</span>
 
 <section>
-  <h4 translate>Activation Key Type:</h4>
```
jturel: I'm not sure what was meant by 'activation key type' so I made it consistent with the Content Host Details tab instead

PR #8578, opened 2020-02-24
Fixes #29142 - delete all http proxies in tests

ianballou:
Required for https://github.com/theforeman/foreman/pull/7458


ianballou:
[test katello]

PR #8580, opened 2020-02-24
Fixes #29146 - Remove upstream idCert from organization details API

jturel:
I can't think of a good reason to expose these credentials which allow complete access to the upstream subscription allocation. If users want to manipulate it we already have an API to do that. Portal RHSM also has its own API. The only place in the UI that uses the upstream consumer details is the Content -> Subscriptions page and I tested that it still works

API output looks like this:

```
"owner_details": {
"id": "4028fac4701add1401701adebd430001",
"key": "Default_Organization",
"created": "2020-02-06T14:19:29+0000",
"updated": "2020-02-25T14:29:11+0000",
"lastRefreshed": "2020-02-25T14:29:11+0000",
"virt_who": false,
"upstreamConsumer": {
"name": "jturel-centos7-devel",
"uuid": "38239dd3-2cc2-4551-bf85-f4d8f8b40cfb",
"webUrl": "access.redhat.com/management/subscription_allocations/"
}
},
```

This PR also provides a rake task to be used by the satellite receptor plugin which is currently using our API to get the id cert and key : https://github.com/project-receptor/satellite-receptor-installer/blob/330ee399e0f6506da979dd74199ff000597216c0/roles/satellite-receptor/tasks/extract_orgs.yml

The rake task gives the receptor a plugin a way to access the certs. It produces output like this:

@j00bar can you confirm these details would be sufficient for the receptor plugin?

```
[                                                                                                                                                                                             
  {                                                                                                                                                                                           
    "id": 114267492,                                                                                                                                                                          
    "redhat_account_number": 12400203,                                                                                                                                                        
    "cert": "<the actual cert>",                                                                                                                                                        
    "key": "<the actual key>"                                                                                                                                                           
  }                                                                                                                                                                                           
]
```

The task can be invoked in dev like this

```
bundle exec rake katello:receptor:extract_orgs OUTPUT_FILE=/tmp/receptor.json
```

or with `foreman-rake` in production environments


jturel:
[test katello]

app/models/katello/content.rb
```diff
@@ -28,7 +28,7 @@ def repositories
     end
 
     def redhat?
-      self.products.first.try(:redhat?)
+      products.redhat.any?
```
jturel: When looking around the codebase I found the opportunity to clean up some of the determinations around upstream / redhat subscriptions

app/models/katello/subscription.rb
```diff
@@ -24,10 +25,7 @@ def self.using_virt_who
     end
 
     def redhat?
-      # for custom subscriptions, there is no separate marketing and engineering product
-      #   so query our Products table and check there
-      product = Katello::Product.where(:cp_id => self.cp_id, :organization => self.organization).first
-      product.nil? || product.redhat?
```
jturel: Not sure if the `product.nil?` check was actually meaningful?

test/katello_test_helper.rb
```diff
@@ -98,6 +98,12 @@ module FixtureTestCase
 
     @@admin = ::User.unscoped.find(FIXTURES['users']['admin']['id'])
     User.current = @@admin
+
+    Organization.all.each do |org|
```
jturel: For my tests I needed the organizations to have their labels set - which they don't because those fixtures are defined in Foreman and we relied on `get_organization` to do that for us. I tried to come up with a better solution which ended up requiring some VCRs to be regenerated. I'm not exactly sure *why* but the results look sane to me.

lib/katello/tasks/receptor/extract_orgs.rake
```diff
@@ -0,0 +1,27 @@
+namespace :katello do
+  namespace :receptor do
```
ehelms: Was just perusing and got curious if the receptor namespace here was required given its a generic-ish rake task.

ehelms:
Just wanted to check on something since as I understand it this is changing the scope of knowledge. The previous implementation implied that an org admin had access to this data. This change removes that and replaces with anybody having root access on the box having access to that information and org admin's loosing that access. (and yea, I realize if you have root on the box you can poke at the database as well).

I agree it's a fair argument as to why an org admin would need that information. But I did want to ensure I understood this was taking data away from one set of authorization in favor of a (more explicit) granting to a different set of authorization.

lib/katello/tasks/receptor/extract_orgs.rake
jturel: I wanted to namespace this in way that might make a user second guess using this task - which they would not have a reason to. Maybe a printed warning would help in clarifying that the result / output of the task has no guarantees other than to serve the needs of the receptor plugin

jturel:
> Just wanted to check on something since as I understand it this is changing the scope of knowledge. The previous implementation implied that an org admin had access to this data. This change removes that and replaces with anybody having root access on the box having access to that information and org admin's loosing that access. (and yea, I realize if you have root on the box you can poke at the database as well).
> 
> I agree it's a fair argument as to why an org admin would need that information. But I did want to ensure I understood this was taking data away from one set of authorization in favor of a (more explicit) granting to a different set of authorization.

I don't think there's a sound reason to provide the credentials over the API. The rake task seemed like a good alternative that can be used readily by the ansible role that needs it, with the added benefit of likely being available to a smaller audience

jturel:
[test katello]

app/models/katello/subscription.rb
jeremylenz: yeah that check kind of breaks my brain.  What you replaced it with looks much better to me.

jturel: much cleaner with the right scopes in place!

jturel:
@j00bar @adamruzicka can you guys take a look at the example output of this rake task and ensure it meets the needs of the receptor plugin installer?

jturel:
Rebased due to #8607 and made a small change to include the `upstreamConsumer` key and show it as null if no manifest is imported to be consistent with the current API

PR #8584, opened 2020-02-25
Fixes #29042 - Applied errata report

sjha4:
https://github.com/theforeman/foreman/pull/6843/files removed uptime_facts and uptime_seconds is now coming from reported_data.


sjha4:
Thanks @ezr-ondrej , @jlsherrill !

PR #8586, opened 2020-02-26
fixes #29271 - Show values instead of references to object

domitea:


xprazak2:
Works for me, you might need to create a separate issue in katello so that the check for redmine issue passes.

app/lib/katello/concerns/base_template_scope_extensions.rb
```diff
@@ -20,7 +20,7 @@ def errata(id)
       end
 
       def host_subscriptions(host)
-        host.subscriptions
+        host.subscriptions.map(&:name).join(', ')
```
ares: We shouldn't be changimg existing macros, people may rely on their behavior. How about introducing `host_subscriprions_names` tjat would internally call `host_subscriptions`? The same for `host_products_names` below

app/lib/katello/concerns/base_template_scope_extensions.rb
```diff
@@ -35,6 +40,22 @@ def host_products(host)
...
+        if products = host_products(host)
+          products.map(&:name).join(', ')
+        else
+          ''
+        end
+      end
+
+      def host_collections(host)
+        host.host_collections
+      end
```
adamruzicka: Is this really necessary?

ares: that's another disucssion, but given this file already has macros like this (`host_subscriptions` e.g.) I'd prefer consistency

app/lib/katello/concerns/base_template_scope_extensions.rb
```diff
@@ -35,6 +40,22 @@ def host_products(host)
...
+
+      def host_collections_names(host)
+        host_collections(host).map(&:name).join(", ")
```
ares: here you don't have to use the macro from above, inside of this method, I think using `host.host_collections` would make more sense

app/lib/katello/concerns/base_template_scope_extensions.rb
```diff
@@ -35,6 +40,22 @@ def host_products(host)
         host_subscription_facet(host)&.installed_products
       end
 
+      def host_products_names(host)
+        if products = host_products(host)
```
ares: can this ever return nil? I assume it's either empty array or array with records. given `[].join(', ') # => ""` this whole method can be a one liner

```
def host_products_names(host)
  products.map(&:name).join(', ')
end
```

unless I'm missing something

adamruzicka: It can be nil if the host doesn't have a subscription facet

ares: correct, ignore me then

ares:
```
[2020-03-05T12:28:40.149Z] Offenses:
[2020-03-05T12:28:40.149Z] 
[2020-03-05T12:28:40.149Z] /home/jenkins/workspace/katello-pr-test/app/lib/katello/concerns/base_template_scope_extensions.rb:44:21: W: Lint/AssignmentInCondition: Use == if you meant to do a comparison or wrap the expression in parentheses to indicate you meant to assign in a condition.
[2020-03-05T12:28:40.149Z]         if products = host_products(host)
[2020-03-05T12:28:40.149Z]                     ^
[2020-03-05T12:28:40.149Z] 
[2020-03-05T12:28:40.149Z] 1713 files inspected, 1 offense detected
```

tests are failing due to rubocop, please address that and the comment for `host_collections_names` then it's good for merge imho

domitea:
@ares code has been improved. I think it's ready for merge.

app/lib/katello/concerns/base_template_scope_extensions.rb
```diff
@@ -32,7 +37,24 @@ def host_sla(host)
       end
 
       def host_products(host)
-        host_subscription_facet(host)&.installed_products
+        host.content_facet&.installed_products
```
jturel: This is raising an error: `undefined method `installed_products' for #<Katello::Host::ContentFacet:0x0000000006ccfd20>`

What was the reason for changing this method?

jturel:
@domitea  I think we'll need a corresponding Foreman PR to update the report template: https://github.com/theforeman/foreman/blob/develop/app/views/unattended/report_templates/entitlements.erb

I'd like that have that so I can test both to see that everything works end-to-end

jlsherrill:
@jturel looks like there's a pr here: https://github.com/theforeman/community-templates/pull/693/

app/lib/katello/concerns/base_template_scope_extensions.rb
domitea: Bad improvement :/

domitea:
@ares, @jturel code improved, I think it's ready for merge

tbrisker:
@jturel @jjeffers can we get this into the next 3.15 RC (or maybe even next 3.14 release)? this can be quite annoying issue I believe.

jjeffers:
We will triage it today - I don't know of a reason why we can't pick it into 3.15.

jturel:
> @jturel @jjeffers can we get this into the next 3.15 RC (or maybe even next 3.14 release)? this can be quite annoying issue I believe.

my first thought was "no problem" but if there's any risk of added churn in the release cycle because of getting the community templates synced I'd rather leave it for the 3.16 release. This report is new so I doubt anyone has encountered the problem yet. What do you think @tbrisker ?

tbrisker:
> > @jturel @jjeffers can we get this into the next 3.15 RC (or maybe even next 3.14 release)? this can be quite annoying issue I believe.
> 
> my first thought was "no problem" but if there's any risk of added churn in the release cycle because of getting the community templates synced I'd rather leave it for the 3.16 release. This report is new so I doubt anyone has encountered the problem yet. What do you think @tbrisker ?

The report was already added in 3.14 if i understand correctly. if we pull back the macros to 3.14 or 3.15 we can also pull the template back to the matching foreman version, in any case adding the macros doesn't require the template to be there, just enables the changes to it.

jturel:
> > > @jturel @jjeffers can we get this into the next 3.15 RC (or maybe even next 3.14 release)? this can be quite annoying issue I believe.
> > 
> > 
> > my first thought was "no problem" but if there's any risk of added churn in the release cycle because of getting the community templates synced I'd rather leave it for the 3.16 release. This report is new so I doubt anyone has encountered the problem yet. What do you think @tbrisker ?
> 
> The report was already added in 3.14 if i understand correctly. if we pull back the macros to 3.14 or 3.15 we can also pull the template back to the matching foreman version, in any case adding the macros doesn't require the template to be there, just enables the changes to it.

Triaged to 3.15 in case you didn't see.

No 3.14.2 planned at the moment and I'd like to keep it that way ;)

jjeffers:
@jturel I saw that on the roadmap, we'll get it in.

tbrisker:
@jjeffers please notice when cherry-picking this also has a followup pr, https://github.com/Katello/katello/pull/8622

PR #7678, opened 2018-09-06
Fixes #24775 - Use REX for errata installation

mdellweg:
Big thank you to @adamruzicka !

PR #8588, opened 2020-02-26
Fixes #29200 - deprecate katello-agent

jlsherrill:


jlsherrill:
note that deprecating the api endpoints didn't seem to cause it to show up in hammer as deprecated.  I need to investigate that

jturel:
Probably a good time to refresh some of the phrasing, etc on https://github.com/Katello/katello/blob/master/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/views/register-client.html

jlsherrill:
Good pointer @jturel !!  I also added (deprecated) to apipie action descriptions so they'd show up more visibliy in hammer --help output

jlsherrill:
Good pointer @jturel !!  I also added (deprecated) to apipie action descriptions so they'd show up more visibliy in hammer --help output

jlsherrill:
Good pointer @jturel !!  I also added (deprecated) to apipie action descriptions so they'd show up more visibliy in hammer --help output

app/controllers/katello/api/v2/host_packages_controller.rb
```diff
@@ -26,7 +27,7 @@ def index
       respond_for_index(:collection => collection)
     end
 
-    api :PUT, "/hosts/:host_id/packages/install", N_("Install packages remotely")
+    api :PUT, "/hosts/:host_id/packages/install", N_("Install packages remotely (deprecated)"), :deprecated => true
```
jturel: Can you tell me what you have in mind as far as deprecating these APIs? I assume that it means we'll remove them, and so the command `hammer host package install` (among the others here) will be gone. It might be better to keep these APIs and delegate to remote execution, because it seems much more natural to type the former (you can guess it if you've used hammer a few times) vs `hammer job-invocation create --job-template 'Install Package - Katello SSH Default' --inputs=...`

We have another open PR in the spirit of keeping this API around, but supporting REX too: https://github.com/Katello/katello/pull/7678

jlsherrill: ahh thats an interesting idea.   Do you envision the UI using these APIs still?  Or would it redirect to the REX UI as it does today?

jlsherrill: regardless, updated to remove those api deprecation warnings, but am curious your thoughts on that

jlsherrill:
[test katello]

app/controllers/katello/api/v2/host_packages_controller.rb
jturel: Hmm good question. It's nice how the UI takes you to the job invocation details automatically via `remote_execution_controller`. We could remove the `remote_execution_controller` and provide the job invocation details in the API responses from this controller. UI can redirect based on that, or just provide notification with the link.

engines/bastion_katello/app/assets/javascripts/bastion_katello/common/views/katello-agent-notice.html
```diff
@@ -0,0 +1,8 @@
+<section>
+  <p bst-alert="warning" >
+    <span translate>
+      Katello-agent is deprecated and will be removed in Katello 4.0.  Consider migrating to using Remote Execution.
```
jturel: nit: is 'migrating to using' correct? I'm actually not sure, but something about it sounds a little strange to my ear. Should it be 'use' or could we remove that word altogether?

jlsherrill: 'migrating to Remote Execution' sounds better, updated

jlsherrill:
[test katello]

PR #8589, opened 2020-02-27
Fixes #29202 - Disallow commas in activation key name

jturel:
With subscription-manager we can register with multiple activation keys separated by commas. This is problematic when we try to explode a string of comma-separated activation key names when one or more of them contain commas - we can't find the nonexistent keys.

Katello itself relies on the ability to register in this way on the host/hostgroup edit page through the 'Activation Key's tab => we store multiple selections separated by commas in a host parameter which ultimately gets passed directly to subscription-manager during provisioning.

Here's what the user sees now:

```
hammer activation-key update --id=1 --lifecycle-environment-id=13 --organization-id=1
Could not update the activation key:
  Validation failed: Name cannot contain commas
```

### DIFF
--- a/app/models/katello/activation_key.rb
+++ b/app/models/katello/activation_key.rb
@@ -34,6 +34,7 @@ module Katello
     validates_lengths_from_database
     validates_with Validators::KatelloNameFormatValidator, :attributes => :name
     validates :name, :presence => true
+    validates :name, :format => { without: /,/, message: _('cannot contain commas') }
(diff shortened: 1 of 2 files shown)


sseelam2:
Tested this PR through hammer and UI. I am not able to create or update activation key names with a comma. It looks good to me.

jturel:
Thanks @sseelam2 and @chris1984 !

PR #8590, opened 2020-02-28
[CP] 3.15 RC1 Picks

jturel:
This remediates the KATELLO-3.15 branch. I went through the redmines and found the ones that were Closed and Triaged but were missing the 3.15 target version. We still have a number of older (pre 3.15) issues that are missing a target version. I guess there's nothing to do with those now but triage them to where they actually landed: https://tinyurl.com/wkturzz

I only saw one oddity: cherry-picking reports that #8441 is missing, and there's actually a diff when I pick it. It's definitely in this branch though. I'm ignoring it unless the build fails.

git diff between this branch and ead2d60c9045e8a2687b7bb41c8a0473bb090ff4 where we bumped to 3.16 on master looks sane to me.

@jjeffers @jlsherrill please double check my work :)


jturel:
Looking at the changelog there are a lot of issues under 'other'. I'm going to recategorize some redmines so they show under the right headings

jlsherrill:
#8441 i manually cherrypicked because it was merged after branching.  So i think what you saw makes sense

jturel:
> #8441 i manually cherrypicked because it was merged after branching. So i think what you saw makes sense

Updated tool_belt: https://github.com/theforeman/tool_belt/pull/253

jturel:
Changelog updated

PR #8591, opened 2020-02-28
Fixes #29149 - Informative message in auto-attach api sca

sseelam2:
This PR helps in providing an error message when auto-attach api is called in simple content access mode.


app/lib/actions/katello/host/auto_attach_subscriptions.rb
```diff
@@ -10,6 +10,9 @@ def queue
 
         def plan(host)
           action_subject(host)
+          if ::Organization.find_by_id(host.organization_id).simple_content_access?
+            fail ::Katello::HttpErrors::BadRequest, _("Auto attach is disabled in Simple Content Access mode")
```
jturel: While it makes perfect sense to raise the error here since this action could be called in several places, we have been making efforts for performance to move certain actions out of dynflow.

For this reason, I think we should raise the error wherever this action is called from. That way we won't be filling the Monitor -> Tasks view with a bunch of 'Failed' tasks. That way we are more likely to have successful records in the tasks list

app/lib/actions/katello/host/auto_attach_subscriptions.rb
```diff
@@ -10,6 +10,9 @@ def queue
 
         def plan(host)
           action_subject(host)
+          if ::Organization.find_by_id(host.organization_id).simple_content_access?
```
jturel: I think this can be simpilified: `host.organization.simple_content_access?`

app/lib/actions/katello/host/auto_attach_subscriptions.rb
jturel: We should be clear that it's the organization in simple content access mode: `This host's organization is in Simple Content Access mode. Auto-attach is disabled` is a little more precise. Let me know what you think

jturel:
Oh by the way: the commit message length isn't limited anymore so we can be more descriptive in commit names now without trying to condense everything!

test/controllers/foreman/organizations_controller_test.rb
```diff
@@ -29,4 +29,15 @@ def test_use_newly_created_org
     assert_redirected_to step2_organization_path(:id => org.to_param)
     assert_equal session[:organization_id], org.id
   end
+
+  test "should not be able to auto attach when host is in simple content access mode" do
+    organization = FactoryBot.create(:organization)
+    organization.stubs(:simple_content_access?).returns(true)
+
+    if organization.simple_content_access?
```
jturel: We don't need this `if` since you are stubbing simple content access to true. We want to make this assertion always.

This test doesn't appear to be issuing any http request so it's not actually testing anything. You'll want to add your test to this file instead: `test/controllers/api/v2/organizations_controller_test.rb`. You'll see the other tests for auto attach there

test/controllers/api/v2/host_subscriptions_controller_test.rb
```diff
@@ -75,6 +75,13 @@ def test_auto_attach
       assert_template 'api/v2/host_subscriptions/index'
     end
 
+    def test_auto_attach_simple_content_access
+      @host.organization.stubs(:simple_content_access?).returns(true)
+      if @host.organization.simple_content_access?
```
jturel: Ditto on the if statement

jturel: This test also appears to not test anything - it does not issue a request for auto attach. Did you try running it locally with ktest?

sseelam2: Yes. I did run it locally with ktest and it passed. Let me know if I am understanding this right. I should be issuing a request for auto attach instead of checking this `@host.organization.simple_content_access?` and asserting for a bad request.

app/controllers/katello/api/v2/organizations_controller.rb
```diff
@@ -136,6 +137,10 @@ def releases
 
     api :POST, "/organizations/:id/autoattach_subscriptions", N_("Auto-attach available subscriptions to all hosts within an organization. Asynchronous operation."), :deprecated => true
     def autoattach_subscriptions
```
jturel: This API was deprecated a couple years back and it looks like it's not being used anywhere. In the spirit of the recent PR which removed other deprecations I think this one is a good candidate for the same. Sound good @sseelam2  @jlsherrill @johnpmitsch ?

johnpmitsch: It sounds like it is a good candidate for removal, just would need to make sure the UI and hammer is not using it or that usage is modified.

I'm surprised there was no warning that went along with the apipie deprecation (though I'm not sure if those are even showing after the community discussion). But I see it has been 4 years since it has been added so is probably safe to remove.

jturel: Totally agree. @sseelam2 I think we can remove this method, its tests, and its route in config/routes

app/controllers/katello/api/v2/organizations_controller.rb
```diff
@@ -4,11 +4,11 @@ class Api::V2::OrganizationsController < ::Api::V2::OrganizationsController
 
     include Api::V2::Rendering
     include ForemanTasks::Triggers
+    include Api::V2::ErrorHandling
```
jturel: I guess this isn't needed now!

sseelam2: yes! thank you

sseelam2:
[test katello]

jturel:
[test katello]

jturel:
[test katello]

PR #8592, opened 2020-02-28
Fixes #29199 - re-import mutable types during p3 migrate

jlsherrill:
Docker tags and errata are mutable.  Which means that running
a migration again could result in the units having new hrefs in
pulp3

### DIFF
--- a/app/services/katello/pulp3/migration.rb
+++ b/app/services/katello/pulp3/migration.rb
@@ -11,6 +11,11 @@ module Katello
         Katello::Repository::DOCKER_TYPE
       ].freeze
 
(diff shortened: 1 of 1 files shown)


jlsherrill:
How to test:

1) configure docker to be used for pulp2
2) create and sync  docker repo pointed at   fedora/ssh
3) run 'rake katello:pulp3_migration'
4) in the rails console run:  "Katello::DockerTag.all".   Take note of migrated_pulp3_href  
5) change repository configuration to point to fedora/couchdb, and sync it
6) re-run 'rake katello:pulp3_migration'
7) re-run the console command

notice that the hrefs have changed

PR #8593, opened 2020-03-02
Fixes #29125 - Only migrate use_pulp_2_for_content_type types

ianballou:
This PR ensures that `katello:pulp3_migration` and `katello:pulp3_content_switchover` only migrate repository types that have `:use_pulp_2_for_content_type:` as `true`.

To test:
1) Make a Pulp 3 development box
2) Apply this PR
3) Add some repo types (file and docker perhaps) under `:use_pulp_2_for_content_type:` in `foreman/config/settings.plugins.d/katello.yaml` and make their values `true`
4) Stick a `binding.pry` after line 8 of `lib/katello/tasks/pulp3_content_switchover.rake`
5) Run `bundle exec rake katello:pulp3_content_switchover`
6) Check that `content_types` matches what you put in `katello.yaml`
7) Stick a `binding.pry` on the first line of `def create_migrations`
8) Ensure that `@repository_types` returns the types under `:use_pulp_2_for_content_type:`

### DIFF
--- a/app/services/katello/pulp3/migration.rb
+++ b/app/services/katello/pulp3/migration.rb
@@ -17,8 +17,13 @@ module Katello
       ].freeze
 
       def initialize(smart_proxy, repository_types = REPOSITORY_TYPES)
-        @smart_proxy = smart_proxy
+        if (repository_types - smart_proxy.supported_pulp_types[:pulp3][:overriden_to_pulp2]).any?
+          fail ::Katello::Errors::Pulp3MigrationError, _("Pulp 3 migration cannot run. Types %s have already been migrated.") %
+            (repository_types - smart_proxy.supported_pulp_types[:pulp3][:overriden_to_pulp2]).join(', ')
+        end
+
         @repository_types = repository_types
+        @smart_proxy = smart_proxy
       end
 
       def api_client
@@ -41,8 +46,8 @@ module Katello
         create_migrations.map { |href| start_migration(href) }
       end
 
-      def self.content_types_for_migration
-        content_types = REPOSITORY_TYPES.collect do |repository_type_label|
+      def content_types_for_migration
+        content_types = @repository_types.collect do |repository_type_label|
           Katello::RepositoryTypeManager.repository_types[repository_type_label].content_types_to_index
         end
 
--- a/app/lib/katello/errors.rb
+++ b/app/lib/katello/errors.rb
@@ -115,6 +115,8 @@ module Katello
       end
     end
 
+    class Pulp3MigrationError < StandardError; end
+
     class PulpError < StandardError
       def self.from_task(task)
         if %w(error canceled).include?(task[:state])
--- a/lib/katello/tasks/pulp3_content_switchover.rake
+++ b/lib/katello/tasks/pulp3_content_switchover.rake
@@ -3,7 +3,8 @@ require File.expand_path("../engine", File.dirname(__FILE__))
 namespace :katello do
   desc "Runs a Pulp 3 migration of pulp3 hrefs to pulp ids for supported content types."
   task :pulp3_content_switchover => :environment do
-    content_types = Katello::Pulp3::Migration.content_types_for_migration
+    migration_service = Katello::Pulp3::Migration.new(SmartProxy.pulp_master)
+    content_types = migration_service.content_types_for_migration
 
     content_types.each do |content_type|
       if content_type.model_class.where(migrated_pulp3_href: nil).any?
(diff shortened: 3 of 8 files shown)


ianballou:
The VCR recordings were recorded with a newer version of pulp-2to3-migration, so the tests will fail until that newer version is built.

ianballou:
The failing tests are passing for me locally, looking into this.

ianballou:
[test katello]

app/lib/actions/pulp3/content_migration.rb
```diff
@@ -9,7 +9,8 @@ def plan(smart_proxy = SmartProxy.pulp_master)
       end
 
       def invoke_external_task
-        migration_service = ::Katello::Pulp3::Migration.new(smart_proxy)
+        migration_service = ::Katello::Pulp3::Migration.new(smart_proxy,
+                                                            smart_proxy.supported_pulp_types[:pulp3][:overriden_to_pulp2])
```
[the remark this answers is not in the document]
jlsherrill: ohhh i see,   I think instead of passing in only certain types (meaning if nothing was to be migrated, it would pass in []), we probably instead want to check and actually throw an error.  thoughts?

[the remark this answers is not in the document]
ianballou: @jlsherrill are you thinking I should add a check for `supported_pulp_types[:pulp3][:overriden_to_pulp2]` being empty and error out if so? But otherwise keep the `migration_service` line the same?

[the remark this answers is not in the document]
jlsherrill: compare that to what is being requested to be migrated, but yes, let the migration service dictate what is being attempted to be migrated

ianballou:
[test katello]

app/services/katello/pulp3/migration.rb
```diff
@@ -16,9 +16,21 @@ class Migration
         Katello::Erratum
       ].freeze
 
-      def initialize(smart_proxy, repository_types = REPOSITORY_TYPES)
+      def initialize(smart_proxy, repository_types = nil)
+        if repository_types.nil?
+          @repository_types = smart_proxy.supported_pulp_types[:pulp3][:overriden_to_pulp2]
```
jlsherrill: wouldn't this cause all types to be migrated?   Meaning if its set to migration only File and Docker (in the constant above), but yum is overriden to use pulp2, that'd get migrated too?

ianballou: Yeah, I'm missing a check for that.

ianballou: I pushed the new check.  The code should now error out if someone tries to create a migration service with un-migratable types in `:use_pulp_2_for_content_type`.

app/services/katello/pulp3/migration.rb
```diff
@@ -16,9 +16,21 @@ class Migration
...
+        else
+          @repository_types = repository_types
+        end
+
+        if @repository_types.empty?
+          fail ::Katello::Errors::Pulp3MigrationError, _("Nothing to migrate.")
+        elsif (@repository_types - REPOSITORY_TYPES).any?
+          fail ::Katello::Errors::Pulp3MigrationError, _("Repository types %s cannot be migrated.") %
+            (@repository_types - REPOSITORY_TYPES).join(', ')
+        end
```
jlsherrill: Currently only the REPOSITORY_TYPES are supported for migration, wouldn't it make sense to do something like this:
@repository_types = REPOSITORY_TYPES

Shouldn't we throw an error if 

REPOSITORY_TYPES -  smart_proxy.supported_pulp_types[:pulp3][:overriden_to_pulp2]

returns anything?  Remember that overriden to pulp2 is essentially telling katello what hasn't been migrated yet, so i think what you have is erroring if something hasn't been migrated yet, but wasn't requested for migration, which i don't think is a situation we need to be checking.

ianballou: What do you mean by "requested for migration"? I thought the types in `smart_proxy.supported_pulp_types[:pulp3][:overriden_to_pulp2]` were the types being requested for migration.

If only 'file' is in `smart_proxy.supported_pulp_types[:pulp3][:overriden_to_pulp2]`, I take that as meaning we should only migrate file.  Wouldn't `REPOSITORY_TYPES - smart_proxy.supported_pulp_types[:pulp3][:overriden_to_pulp2]` in that case return 'docker' and fail?

ianballou: Do you think that the migration should silently pass if there's nothing to migrate? Then the only error case would be if something is requested to be migrated that can't be.  That would just mean I take out the `@repository_types.empty?` check.

ianballou:
Think this is all set now.

PR #8599, opened 2020-03-05
Fixes #29260 - Add -y flag for yum command

loitho:
Hi there,
As described in : https://projects.theforeman.org/issues/29260
And https://community.theforeman.org/t/using-sudo-su-for-privilege-escalation/17635/4

The Katello Ansible Default templates that are using Ansible shell commands directly for creating Yum action like update are missing a "-y" flag which makes them hang 

This fixes the problem by adding the missing -y flag on each template that was missing it.

### DIFF
--- a/app/views/foreman/job_templates/install_errata_-_katello_ansible_default.erb
+++ b/app/views/foreman/job_templates/install_errata_-_katello_ansible_default.erb
@@ -15,4 +15,4 @@ kind: job_template
 %>
 
 <% advisories = input(:errata).split(',').map { |e| "--advisory=#{e}" }.join(' ') %>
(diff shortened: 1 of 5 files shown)


ares:
:+1: 

question for katello maintainers - is there a sync script that would automate this? should we drop templates from this repo and start seeding it from community templates? templates can contain requirements from now, so they wouldn't be seeded if Katello is not available

ianballou:
@ares we don't have any processes right now in Katello for keeping the job templates in this repo synced with the community templates.  I agree that it would be better to either add a way to sync or start having Katello seed from the community templates.  I will make an issue on redmine about it.

ianballou:
[test katello]

PR #8600, opened 2020-03-09
Fixes #29314 - add deprecation for background download settings

jlsherrill:


jlsherrill:
image

jlsherrill:
[test katello]

app/models/setting/content.rb
```diff
@@ -11,7 +11,12 @@ def self.hashify_parameters(parameters)
   end
 
   def self.default_settings
-    download_policies = proc { hashify_parameters(::Runcible::Models::YumImporter::DOWNLOAD_POLICIES) }
+    download_policies = proc do
+      hash = hashify_parameters(::Runcible::Models::YumImporter::DOWNLOAD_POLICIES)
+      hash['background'] = 'background (orphaned)'
+      hash
+    end
+
```
sjha4: This works as expected on the settings page. :heavy_check_mark: 
Couple of questions:
1. Why are we using orphaned instead of deprecated?
2. Should this apply to proxy_download_policies as well?

jlsherrill: 1.  apparently i was talking about orphans to someone at the time and my wires got crossed!
2. yes!!!

jlsherrill: updated!

PR #8601, opened 2020-03-09
Fixes #29316 - Ignore missing candlepin content on product delete

jturel:
Rather than raise an error which would break the dynflow task, denote that the content is gone


mccun934:
Tested this by reproducing the error condition (manually deleted cp2_content rows in candlepin db) and attempting to delete the repo. Error occurred:

`Content with ID "*********" could not be found. (Katello::Errors::CandlepinError)`

tried patch in this PR but my environment did not contain:

https://github.com/Katello/katello/pull/8106

with the combination of this PR + #8106 it worked great.

PR #8605, opened 2020-03-10
Fixes #29321 - Indexing large repo

sjha4:
To do:

- [x] Measure time for sync/index with/without the change

### DIFF
--- a/test/models/file_unit_test.rb
+++ b/test/models/file_unit_test.rb
@@ -43,5 +43,18 @@ module Katello
       ids = ['href'] * 70_000 + [@file_one.pulp_id]
       assert_equal 1, FileUnit.with_pulp_id(ids).count
     end
+
+    def test_large_sync_repository_association
+      i, ids, ids_href_map = 0, [], {}
+      while (i < 70_000)
+        ids[i] = ["href#{i}"]
+        ids_href_map["href#{i}"] = nil
+        i += 1
+      end
+      Katello::FileUnit.import([:pulp_id], ids)
+      Katello::FileUnit.sync_repository_associations(@repo, :pulp_id_href_map => ids_href_map)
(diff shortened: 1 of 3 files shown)


jlsherrill:
Do you think it'd be possible to add a test that would have caught this (without having to actually sync a repo with thousands of units)?

I'm thinking you could create 50,000 fake file units with ActiveRecordImport fairly quickly, and then grab all the ids and pass that to the sync_repository_association method?

sjha4:
@jlsherrill : Added a couple of tests for the 2 types of flows, file and errata with 70000 units. Is that what you had in mind?

test/models/file_unit_test.rb
```diff
@@ -43,5 +43,18 @@ def test_large_query
...
     end
+
+    def test_large_sync_repository_association
+      i, ids, ids_href_map = 0, [], {}
+      while (i < 70_000)
+        ids[i] = ["href#{i}"]
+        ids_href_map["href#{i}"] = nil
+        i += 1
+      end
+      Katello::FileUnit.import([:pulp_id], ids)
```
jlsherrill: Adding  validate: false shaves off about 20 seconds from this method call.  It still takes about ~14 seconds, but i think thats okay.  I don't know that i see a ton of value for having an errata test and a file test, any reason to have the errata test in addition?

sjha4: Right..I dont think we need the validate:true for this. 
The errata test follows a different path where there are actually mappings in the id_href_map..So the 2 flows are the map having nil values for all keys on all content types minus errata and map having actual pulp3_hrefs for errata that we want to store on the katello_repository_erratum table.

sjha4:
Thanks @jlsherrill ! Merging..

PR #8606, opened 2020-03-10
Refs #29322 - Add custom scoped_search parameter

jeremylenz:
[SAT-E-459]  Updated with [testing instructions](https://github.com/Katello/katello/pull/8606#issuecomment-600827102) 3/18

This PR contains updates to enable improvements to the Entitlements report template.

* Add `days_until_expiration` method to `Pool` - Allows the addition of a new column to the Entitlements report, _Subscription Expires In_
* Allow access via Jail so that you can use it in a report template
* Add a scoped_search parameter, `with_pools_expiring_in_days`, with a custom external method that joins Hosts and Pools and fetches the required data.

Inside a report template, this allows you to do 
```
load_hosts(search: 'pools_expiring_in_days = 30')
```

Or anywhere else,
```
Host.search_for('pools_expiring_in_days = 30')
```
/
and get all hosts with pools expiring in the next 30 days.

### DIFF
--- a/app/models/katello/concerns/host_managed_extensions.rb
+++ b/app/models/katello/concerns/host_managed_extensions.rb
@@ -61,6 +61,14 @@ module Katello
         scoped_search :relation => :host_traces, :on => :application, :complete_value => true, :rename => :trace_app, :only_explicit => true
         scoped_search :relation => :host_traces, :on => :app_type, :complete_value => true, :rename => :trace_app_type, :only_explicit => true
         scoped_search :relation => :host_traces, :on => :helper, :complete_value => true, :rename => :trace_helper, :only_explicit => true
+
+        scoped_search relation: :pools, on: :end_date, ext_method: :find_with_pools
+
+        def self.find_with_pools(key, operator, value)
+          conditions = sanitize_sql_for_conditions(["katello_pools.#{key} #{operator} ?", value_to_sql(operator, value)])
+          host_ids = ::Host.joins(:pools).where(conditions).ids
+          { :conditions => "hosts.id IN(#{host_ids.join(',')})" }
+        end
       end
 
       def correct_kickstart_repository
--- a/app/models/katello/concerns/host_managed_extensions.rb
+++ b/app/models/katello/concerns/host_managed_extensions.rb
@@ -51,6 +51,8 @@ module Katello
         before_save :correct_puppet_environment
         before_validation :correct_kickstart_repository
 
+        scope :with_pools_expiring_in_days, ->(days) { joins(:pools).merge(Katello::Pool.expiring_in_days(days)).distinct }
+
         scoped_search :relation => :host_collections, :on => :id, :complete_value => false, :rename => :host_collection_id, :only_explicit => true, :validator => ScopedSearch::Validators::INTEGER
         scoped_search :relation => :host_collections, :on => :name, :complete_value => true, :rename => :host_collection
         scoped_search :relation => :installed_packages, :on => :nvra, :complete_value => true, :rename => :installed_package, :only_explicit => true
@@ -65,26 +67,17 @@ module Katello
         scoped_search relation: :pools, on: :pools_expiring_in_days, ext_method: :find_with_expiring_pools, only_explicit: true
 
         def self.find_with_expiring_pools(_key, _operator, days_from_now)
-          minimum_pool_date = (Date.today + days_from_now.to_i.days).strftime('%Y-%m-%d')
-          host_ids = self.joins(:pools).where(["katello_pools.end_date < ?", minimum_pool_date]).distinct.ids
-          if host_ids.blank?
-            return {
-              :conditions => "katello_pools.end_date < ?",
-              :parameter => [minimum_pool_date],
-              :joins => :pools
-            }
-          end
-          {
-            :conditions => "hosts.id IN (#{host_ids.join(', ')})"
+          host_ids = with_pools_expiring_in_days(days_from_now).ids
+          scoped_search_params = {
+            :conditions => "katello_pools.end_date < ?",
+            :parameter => [days_from_now.to_i.days.from_now.end_of_day],
+            :joins => :pools
           }
+          if host_ids.present?
+            scoped_search_params = { :conditions => "hosts.id IN (#{host_ids.join(', ')})" }
+          end
+          scoped_search_params
         end
-      end
-
-      # show a single host's expiring pools
-      def pools_expiring_in_days(days_from_now)
-        return self.pools if days_from_now.blank?
-        minimum_pool_date = (Date.today + days_from_now.to_i.days).strftime('%Y-%m-%d')
-        self.pools.where(["katello_pools.end_date < ?", minimum_pool_date]).order(:end_date)
       end
 
       def correct_kickstart_repository
@@ -317,5 +310,9 @@ end
 class ::Host::Managed::Jail < Safemode::Jail
   allow :content_source, :subscription_manager_configuration_url, :rhsm_organization_label,
         :host_collections, :comment, :pools, :hypervisor_host, :lifecycle_environment, :content_view,
-        :installed_packages, :pools_expiring_in_days
+        :installed_packages
+end
+
+class ActiveRecord::Associations::CollectionProxy::Jail < Safemode::Jail
+  allow :expiring_in_days
 end
--- a/app/models/katello/concerns/host_managed_extensions.rb
+++ b/app/models/katello/concerns/host_managed_extensions.rb
@@ -51,12 +51,6 @@ module Katello
         before_save :correct_puppet_environment
         before_validation :correct_kickstart_repository
 
-        # show all hosts which have expiring pools
-        scope :with_pools_expiring_in_days, ->(days_from_now) do
-          minimum_pool_date = (Date.today + days_from_now.to_i.days).strftime('%Y-%m-%d')
-          joins(:pools).where(["katello_pools.end_date < ?", minimum_pool_date])
-        end
-
         scoped_search :relation => :host_collections, :on => :id, :complete_value => false, :rename => :host_collection_id, :only_explicit => true, :validator => ScopedSearch::Validators::INTEGER
         scoped_search :relation => :host_collections, :on => :name, :complete_value => true, :rename => :host_collection
         scoped_search :relation => :installed_packages, :on => :nvra, :complete_value => true, :rename => :installed_package, :only_explicit => true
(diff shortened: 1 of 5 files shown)


jeremylenz:
Entitlements report template update: https://github.com/theforeman/community-templates/pull/705

jeremylenz:
Cross-post from theforeman/community-templates#705 :
Updated to use `search` instead of filtering in the erb.
By using scoped_search and a custom external method, I am able to get scoped_search to return only hosts _with expiring pools_.  However, when iterating through each host, `host.pools` would still return _all_ of that host's pools, not just the expiring ones.  So, I also added an instance method to `Host`, `pools_expiring_in_days`, so that none of the filtering has to be done inside the report template.  Thanks @jlsherrill for the help

Also added a test.

jjeffers:
@jeremylenz I haven't followed all of associated discussions re this PR, but I am happy to test locally, to complete the review. Do you recommend a particular workflow to see the changes and verify?

jeremylenz:
@jjeffers With both this PR and the community-templates PR checked out, you can do one of the 2 methods in [this](https://github.com/theforeman/community-templates/pull/705#issuecomment-597289365) comment.  Let me know if you have any issues.  thanks!

jeremylenz:
Better testing instructions:

1. Have at least one Host with or without expiring subscriptions
2. check out this PR
3. Go to Monitor > Report Templates
4. Clone the Entitlements report
5. Edit your new clone and replace the ERB contents with those of [entitlements.erb](https://github.com/theforeman/community-templates/blob/2ae2a49b23655a701f2f10462135f9a25a9674ea/report_templates/entitlements.erb) in community-templates [PR #705](https://github.com/theforeman/community-templates/pull/705)
6. Manually add the input on the Inputs tab (should look like this:)
input
7. Go back out to the Report Templates page and click Generate on your cloned report.
8. For "Days from Now," leave blank to show all entitlements, or choose an option to show expiring subscriptions.  (Note the report will be blank if you don't have any hosts with subscriptions expiring in that number of days)

* Notice new column _Subscription Ends In_
* Should be sorted by subscription end date

app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -66,16 +66,23 @@ def update_os_from_facts
...
         def self.find_with_expiring_pools(_key, _operator, days_from_now)
           minimum_pool_date = (Date.today + days_from_now.to_i.days).strftime('%Y-%m-%d')
+          host_ids = self.joins(:pools).where(["katello_pools.end_date < ?", minimum_pool_date]).distinct.ids
+          if host_ids.blank?
+            return {
+              :conditions => "katello_pools.end_date < ?",
+              :parameter => [minimum_pool_date],
+              :joins => :pools
+            }
+          end
```
jeremylenz: `scoped_search` was causing a SQL syntax error whenever the host_ids returned zero hosts.  This workaround looks weird but seems to work fine.

You may be asking, why not just return this no matter what?  It's because `scoped_search` doesn't allow any way to do `SELECT DISTINCT`, so it was giving me duplicate Host objects and thus, duplicate report rows.  I work around that by doing `.distinct` myself and then just giving scoped_search the ID's.

app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -66,16 +66,23 @@ def update_os_from_facts
...
-            :conditions => "katello_pools.end_date < ?",
-            :parameter => [minimum_pool_date]
+            :conditions => "hosts.id IN (#{host_ids.join(', ')})"
           }
         end
       end
 
       # show a single host's expiring pools
       def pools_expiring_in_days(days_from_now)
+        return self.pools if days_from_now.blank?
```
jeremylenz: This handles the situation where the user leaves the "Days from Now" input blank.  It will result in a report with all pools for all hosts.

jeremylenz:
@jjeffers updated :)

app/models/katello/concerns/host_managed_extensions.rb
jturel: Whenever I see a method like this which refers to an association in the name: *pools*_in_expiring_in_days this tells me "use a scope" that would entail placing a scope definition on the Pool model so we could call `host.pools.expiring_in_days`

That way we have a more generic solution. We could call Katello::Pool.expiring_in_days directly if we wanted!

app/models/katello/pool.rb
```diff
@@ -50,8 +50,15 @@ def active?
...
 
+    # used for notification bell
     def expiring_soon?
-      (end_date.to_date - Date.today) <= Setting[:expire_soon_days]
+      expiring_in_days >= 0 &&
+        expiring_in_days <= Setting[:expire_soon_days].to_i
+    end
+
+    # used for entitlements report template
+    def expiring_in_days
```
jturel: This method name leads me to believe it expects a parameter named 'days' even though that makes no sense because it's an instance method and we're looking at a single pool rather than querying a set.  What do you think about `days_until_expiration`

app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -61,6 +61,30 @@ def update_os_from_facts
         scoped_search :relation => :host_traces, :on => :application, :complete_value => true, :rename => :trace_app, :only_explicit => true
         scoped_search :relation => :host_traces, :on => :app_type, :complete_value => true, :rename => :trace_app_type, :only_explicit => true
         scoped_search :relation => :host_traces, :on => :helper, :complete_value => true, :rename => :trace_helper, :only_explicit => true
+
+        scoped_search relation: :pools, on: :pools_expiring_in_days, ext_method: :find_with_expiring_pools, only_explicit: true
+
+        def self.find_with_expiring_pools(_key, _operator, days_from_now)
+          minimum_pool_date = (Date.today + days_from_now.to_i.days).strftime('%Y-%m-%d')
+          host_ids = self.joins(:pools).where(["katello_pools.end_date < ?", minimum_pool_date]).distinct.ids
```
jturel: Based on my recommendation of using a scope I think you could call that here ie `pools.expiring_in_days.distinct.ids` Sorry if this feedback is all irrelevant because of scoped search, but I *think* it should be OK

app/models/katello/concerns/host_managed_extensions.rb
jturel: In response to your comment about returning early from this method - I don't think we should do it especially if we're using a scope.

Your template (https://github.com/theforeman/community-templates/pull/705/files#diff-b757c54cdf8d66853a76e863d2269309R23) could be updated to look like this

```
pools = host.pools
pools = pools.expiring_in_days(days_from_now) if days_from_now

pools.each do |pool|
...
```

jeremylenz:
Updated to address @jturel comments

app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -61,6 +63,21 @@ def update_os_from_facts
         scoped_search :relation => :host_traces, :on => :application, :complete_value => true, :rename => :trace_app, :only_explicit => true
         scoped_search :relation => :host_traces, :on => :app_type, :complete_value => true, :rename => :trace_app_type, :only_explicit => true
         scoped_search :relation => :host_traces, :on => :helper, :complete_value => true, :rename => :trace_helper, :only_explicit => true
+
+        scoped_search relation: :pools, on: :pools_expiring_in_days, ext_method: :find_with_expiring_pools, only_explicit: true
+
+        def self.find_with_expiring_pools(_key, _operator, days_from_now)
```
jturel: Does this simplification work?

```
        def self.find_with_expiring_pools(_key, _operator, days_from_now)                                                                                     
          host_ids = with_pools_expiring_in_days(days_from_now).ids                                                                                           
          condition = if host_ids.any?                                                                                                                        
                        "hosts.id IN (#{host_ids.join(', ')})"                                                                                                
                      else                                                                                                                                    
                        '1=0'                                                                                                                                 
                      end                                                                                                                                     
          { conditions: condition }                                                                                                                           
        end 
```

I don't understand what the initial definition of `scoped_search_params` is needed for (the case when there are no matching hosts). It seems like if there are no matching hosts then we weren't going to show any results anyway - regardless of any addition scoped search filters, so no need to join pools, filter them, etc.

jeremylenz: The `:conditions` value becomes the `WHERE` clause.  I agree that it could use some simplification, but I don't think `WHERE 1=0` would work..?
(see [this](https://github.com/Katello/katello/pull/8606#discussion_r395337525) comment for why we need the initial definition.)

jeremylenz: How about something like `hosts.id = 0` ?

jturel: But 1=0 is three characters long ;)

jturel: Also, please add a test for this in HostManagedExtensionsTest by calling search_for

jeremylenz:
@jturel @jjeffers Updated :)

app/models/katello/pool.rb
```diff
@@ -100,8 +110,12 @@ def default_sort
       Pool.joins(:subscription).order("subscription.name")
     end
 
-    class Jail < ::Safemode::Jail
-      allow :id, :name, :available, :quantity, :product_id, :contract_number, :type, :account_number, :start_date, :end_date, :organization, :consumed
+    class Pool::Jail < ::Safemode::Jail
```
jturel: was this change necessary or intentional? it seems weird because this class is basically called 'Katello::Pool::Pool::Jail` which is a little redundant

jeremylenz: It was to be consistent with how the jail is defined in `host_managed_extensions.rb` but I didn't realize this would result in this weird naming.  Changing it back.

app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -61,6 +63,19 @@ def update_os_from_facts
...
+          host_ids = with_pools_expiring_in_days(days_from_now).ids
+          scoped_search_params = {
```
jturel: To recreate the whole params hash based on the condition is an unusual way to write this code. Why not conditionally set the conditions key instead? I had my prior suggestion but here's another way:

```
scoped_search_params = {}
if host_ids.any? # any? is more idiomatic than present?
  scoped_search_params[:condition] = ....
else
 scoped_search_params[:condition] ...
end
```

jeremylenz: @jturel sorry, forgot to update that

jjeffers:
Tested this and was able to see the new columns, etc, with appropriate expiration days listed for each subscription/host.

jeremylenz:
[test katello]

jturel:
[test katello]

PR #8607, opened 2020-03-11
Refs #28659 - Show progress bar for all manifest actions

jturel:
This PR builds on the work in #8491 and fixes two bugs:

- Updating the entitlement quantity of an existing subscription left the page in a weird state with a console error
```
Uncaught (in promise) TypeError: Cannot read property 'owner_details' of undefined
    at _callee$ (UpdateDialog.js:25)
    at tryCatch (runtime.js:45)
    at Generator.invoke [as _invoke] (runtime.js:271)
    at Generator.prototype.<computed> [as next] (runtime.js:97)
    at asyncGeneratorStep (UpdateDialog.js:7)
    at _next (UpdateDialog.js:7)
```
- Updating the entitlement quantity of an existing subscription now shows a progress bar
- Adding new subscriptions sometimes revealed an error because the task id prop type was a number instead of a string (they are uuids):
```
react_devtools_backend.js:6 Warning: Failed prop type: Invalid prop `upstreamSubscriptions.task.id` of type `string` supplied to `UpstreamSubscriptionsPage`, expected `number`.
    in UpstreamSubscriptionsPage (created by Context.Consumer)
    in withRouter(UpstreamSubscriptionsPage) (created by ConnectFunction)
```
- Adding new subscriptions now shows a progress bar

Will comment directly on the code where the changes are not obvious!

Testing this page is complicated because there is a lot of functionality. Definitely have your JS console open to find any errors. At least the following should be verified:

- Manifest import
- Manifest delete
- Removing subscriptions
- Adding subscriptions
- Updating subscription entitlement quantity
- Duplication/Rollup subscriptions (add some quantity of a subscription, then add another quantity of the same subscription to create the collapsible row)


webpack/scenes/Subscriptions/SubscriptionActions.js
```diff
@@ -88,10 +88,10 @@ export const updateQuantity = (quantities = {}) => async (dispatch) => {
 
   try {
     const { data } = await api.put(`/organizations/${orgId()}/upstream_subscriptions`, params);
-    return dispatch({
-      type: UPDATE_QUANTITY_SUCCESS,
-      response: data,
-    });
+
+    dispatch(pollTaskUntilDone(data.id, {}, POLL_TASK_INTERVAL, Number(orgId())));
```
jturel: pollTaskUntilDone ensures that the progress modal shows up immediately rather than on the next poll that the page does naturally

webpack/scenes/Subscriptions/SubscriptionActions.js
```diff
@@ -88,10 +88,10 @@ export const updateQuantity = (quantities = {}) => async (dispatch) => {
...
+
+    return dispatch({ type: UPDATE_QUANTITY_SUCCESS });
```
jturel: Removed the payload - it is not used

webpack/scenes/Subscriptions/SubscriptionReducer.js
```diff
@@ -87,11 +82,7 @@ export default (state = initialState, action) => {
 
     case DELETE_SUBSCRIPTIONS_SUCCESS:
       return state
-        .set('deleteButtonDisabled', true)
-        .set('loading', false);
```
jturel: There were a lot of actions setting loading to false. I don't see the point especially now that things are using the progress bar. Let me know if I missed something with that

webpack/scenes/Subscriptions/SubscriptionsPage.js
```diff
@@ -109,9 +109,9 @@ class SubscriptionsPage extends Component {
   async pollTasks() {
     const { pollBulkSearch, organization } = this.props;
 
-    if (organization && organization.owner_details) {
+    if (organization) {
       pollBulkSearch({
-        action: `organization '${organization.owner_details.displayName}'`,
```
jturel: I don't know what the intention of this code was. This filtering was not working as intended such that it was not finding in-progress tasks for the organization. Now it filters in a more succinct way - by org ID and action labels. Let me know if I'm missing the true intent of this method

webpack/scenes/Subscriptions/SubscriptionsPage.js
```diff
@@ -143,7 +143,7 @@ class SubscriptionsPage extends Component {
   cleanUpManifestTask = async () => {
     await renderTaskStartedToast(this.props.taskDetails);
     setTimeout(() => this.props.bulkSearch({
-      action: `organization '${this.props.organization.owner_details.displayName}'`,
+      organization_id: this.props.organization.id,
```
jturel: Updated similar to the above

jturel:
@johnpmitsch  @jeremylenz  I requested your reviews here since you're familiar with the functions of the page and the react code that makes it happen. Let me know if I can clarify on any of the changes here!

jlsherrill:
Running through the scenarios, i did:
* import manifest
* refresh manifest
* add a subscription
* add a 2nd instace of that same subscription
I then noticed an error:
```
katello.js:53356 Uncaught (in promise) undefined
asyncGeneratorStep @ katello.js:53356
_throw @ katello.js:53358
Promise.then (async)
asyncGeneratorStep @ katello.js:53356
_next @ katello.js:53358
(anonymous) @ katello.js:53358
(anonymous) @ katello.js:53358
handleDoneTask @ SubscriptionsPage.js:133
componentDidUpdate @ SubscriptionsPage.js:60
commitLifeCycles @ react-dom.development.js:20064
commitLayoutEffects @ react-dom.development.js:22813
callCallback @ react-dom.development.js:347
invokeGuardedCallbackDev @ react-dom.development.js:397
invokeGuardedCallback @ react-dom.development.js:454
commitRootImpl @ react-dom.development.js:22585
unstable_runWithPriority @ scheduler.development.js:643
runWithPriority$2 @ react-dom.development.js:11305
commitRoot @ react-dom.development.js:22414
runRootCallback @ react-dom.development.js:21554
(anonymous) @ react-dom.development.js:11353
unstable_runWithPriority @ scheduler.development.js:643
runWithPriority$2 @ react-dom.development.js:11305
Pattern not found  (press RETURN)
```

jlsherrill:
Didn't look at the code, just did some manual testing, will leave that more react-knowledgeable folks

jturel:
> Its strange, i went through again and added a new subscription (that didn't exist yet), and i saw the same error immediately, it actually popped up twice. I then added another new subscription and didn't see it at all man_shrugging

Nice find. It appears to be some kind of race condition. Looking into it. Kinda tricky since it's not reliably producible

jeremylenz:
@jturel My immediate hunch is that it's something with the pollTaskUntilDone function and would be resolved once https://projects.theforeman.org/issues/28889 is addressed.

jeremylenz:
`Uncaught (in promise) undefined` usually means you should look for a promise `.then` without a `.catch`, or an `await` without a `try..catch` block.  I'll take a further look a bit later

jturel:
@johnpmitsch @jeremylenz pushed a change that I think fixes the race condition.

You can attempt to recreate the subscription just by adding new subscription (Add Subscriptions button)

webpack/scenes/Subscriptions/SubscriptionReducer.js
```diff
@@ -14,11 +14,7 @@ import {
   SUBSCRIPTIONS_QUANTITIES_FAILURE,
   SUBSCRIPTIONS_COLUMNS_REQUEST,
   UPDATE_SUBSCRIPTION_COLUMNS,
-  UPDATE_QUANTITY_REQUEST,
-  UPDATE_QUANTITY_SUCCESS,
-  UPDATE_QUANTITY_FAILURE,
```
johnpmitsch: Mentioned off-thread, but commenting here for visibility: These are still dispatched, should this be handled differently? Right now, it will fall to the default state, suggesting no dispatch was needed.

johnpmitsch:
I'll double check my environment, but I'm seeing the delete subs progress bar hang around even after the task has finished.

It feels like the "kicking off a task via the API and poll for it while displaying it's status" workflow could use a hard look and perhaps an overhaul to make things more stable, but I know you are trying to get a short-term fix in. I'll dig into the code more and see if I have any ideas as well.

jturel:
> I'll double check my environment, but I'm seeing the delete subs progress bar hang around even after the task has finished.
> 
> It feels like the "kicking off a task via the API and poll for it while displaying it's status" workflow could use a hard look and perhaps an overhaul to make things more stable, but I know you are trying to get a short-term fix in. I'll dig into the code more and see if I have any ideas as well.

Agree!

How long is it sticking around for exactly? Does it go away?

jlsherrill:
i saw the same thing, and it never went away.  Was stuck at 0%

jturel:
@johnpmitsch @jlsherrill @jeremylenz I made some small adjustments to the code to try and ensure that 1) the page's state gets updated when there are no tasks and 2) close the modal when there are no tasks

one quirk I am seeing is that sometimes the success notification pops up twice, but I haven't seen any stuck progress bars

jturel:
@johnpmitsch @jeremylenz pushed another update - don't be afraid of the size!

As I was trying to figure out the stuck progress bar edge case (it was a race condition with TASK_BULK_SEARCH) I saw a lot of weird behavior thanks to the redux dev tools: unnecessary loading of subscriptions, *multiple* bulk tasks pollers, and a whole lot of logic related to tasks in `componentWillUpdate` that made me think "there has to be a better way" so I made these improvements:

- moved single and bulk task pollers into SubscriptionsActions from SubscriptionsPage
- pollBulkTasks now understands how to pause and cancel itself. It will pause when a single task is being polled (the progress modal is shown) and resume when polling is finished. It will cancel when the page unmounts. This was necessary because the pollers were still running when navigating to Add Subscriptions
- the state is handled by a few keys in the store: `pollingATask`, `bulkTasksPaused`, and `bulkTasksCancelled`

The result I've seen is that the page works much more smoothly w.r.t the manifest actions. When testing I recommend watching the redux actions to see the flow clearly. Let me know if you have any questions!

webpack/scenes/Subscriptions/SubscriptionActions.js
```diff
@@ -75,7 +80,49 @@ export const loadSubscriptions = (extendedParams = {}) => async (dispatch) => {
...
+  if (pollingATask) {
+    return dispatch({ type: 'ALREADY_POLLING_TASK' });
+  }
+
+  return dispatch(pollTaskUntilDone(
+    task.id,
+    {},
+    POLL_TASK_INTERVAL,
+    task.input.current_organization_id,
+  // eslint-disable-next-line promise/prefer-await-to-then
```
jeremylenz: I don't see a reason to override the eslint rule here?  You can use a `try..finally` block with async/await (`catch` is not required.)

webpack/scenes/Subscriptions/__tests__/SubscriptionsActions.test.js
```diff
@@ -145,6 +157,50 @@ describe('subscription actions', () => {
...
+        data: {},
+        response: mockFinishedTask,
+      });
+
+      await store.dispatch(handleTask(mockPendingTask));
+
+      expect(store.getActions()).toEqual(handleTaskActions);
+    });
+
+    it('does not poll if already polling', () => {
```
jeremylenz: Interval middleware would take care of this as well, with its unique keys

jturel:
I'm looking at the possibility of using Foreman's 'Interval' middleware to simplify the polling code a lot. stay tuned

jeremylenz:
@jturel Was just about to say that!! See pending review

jturel:
@jeremylenz updated!

webpack/scenes/Tasks/TaskActions.js
```diff
@@ -1,133 +1,41 @@
...
+export const stopPollingTasks = key => stopInterval(bulkSearchKey(key));
 
-    return dispatch({ type: TASK_BULK_SEARCH_CANCELLED });
-  };
+const getTask = (key, task) => get({
+  key,
+  url: `${foremanTasksApi.baseApiPath}/tasks/${task.id}`,
+});
 
-export const pollTaskUntilDone = (taskId, extendedParams = {}, interval, orgId) =>
```
jeremylenz: :tada: :tada: :tada:

webpack/scenes/Tasks/helpers.js
```diff
@@ -1,6 +1,13 @@
 import React from 'react';
 import { translate as __ } from 'foremanReact/common/I18n';
 import { urlBuilder } from 'foremanReact/common/urlHelpers';
+import { selectDoesIntervalExist } from 'foremanReact/redux/middlewares/IntervalMiddleware/IntervalSelectors';
+
+export const bulkSearchKey = key => `${key}_TASK_SEARCH`;
+export const pollTaskKey = key => `${key}_POLL_TASK`;
+
+export const isPollingTask = (state, key) => selectDoesIntervalExist(state, pollTaskKey(key));
+export const isPollingTasks = (state, key) => selectDoesIntervalExist(state, bulkSearchKey(key));
```
jeremylenz: These are selectors, so they should be in a new file `TaskSelectors.js` and named `selectIsPollingTask` and `selectIsPollingTasks`

webpack/scenes/Subscriptions/__tests__/SubscriptionsActions.test.js
```diff
@@ -18,21 +18,35 @@ import {
...
   enableDeleteButton,
 } from '../SubscriptionActions';
 
+import { getTaskPendingResponse, getTaskSuccessResponse } from '../../Tasks/__tests__/task.fixtures';
+
 const mockStore = configureMockStore([thunk]);
-const store = mockStore({ subscriptions: Immutable({}) });
+const store = mockStore(Immutable({
+  intervals: {
+    SUBSCRIPTIONS_TASK_SEARCH: 5,
```
jeremylenz: Can you use [jest.useFakeTimers()](https://jestjs.io/docs/en/timer-mocks) instead of using a real 5ms interval?

jturel: I'm unclear on what that would accomplish since the test isn't actually executing the interval polling but making a determination based on what's in the store. Maybe I'm missing your point.

jeremylenz: I see, I guess it's not applicable here.  Carry on  :yum:

jturel:
@jeremylenz updated!

jturel:
@jlsherrill @johnpmitsch now that the changes are finalized and have an ack I want to give you the chance to test again. Up to you!

jturel:
[test katello]

jturel:
Merging. Thanks for all the great reviewing here!

PR #8610, opened 2020-03-16
Fixes #29355 - allow CVE id in safe mode

ares:
to test, apply [this](https://github.com/theforeman/community-templates/pull/709) to your report template and generate, cves column will contain CVE identifiers instead of internal AR object id


ares:
Please also let me know if this hits Katello 3.17, I'll need to update the report template accrodingly

jlsherrill:
@ares apologies, this fell through the cracks.

jlsherrill:
@ares this will make 3.16 (foreman 2.1)

ares:
thanks Justin, redmine still says 3.17 but I'll udpate the template so it requires 3.16+

PR #8611, opened 2020-03-16
Fixes #29357: remove vendor as a dependency.

waldenraines:
Vendor isn't used directly as a dependency in katello but is rather
provided by foreman.  Remove vendor from the list of dependencies.

https://projects.theforeman.org/issues/29357


waldenraines:
Not needed because of https://github.com/Katello/katello/pull/8612

PR #8612, opened 2020-03-16
Fixes #29358 - Upgrade foreman-js meta-package

johnpmitsch:
This provides patternfly 4 by way of the foreman-js metapackages. The actual production dependency comes from foreman, where it is already upgraded. This only upgrades the dev packages

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,8 +26,8 @@
     "@babel/core": "^7.7.0",
     "@storybook/react": "^3.2.17",
     "@storybook/storybook-deployer": "^2.0.0",
-    "@theforeman/builder": "^4.0.2",
-    "@theforeman/vendor-dev": "^3.8.0",
+    "@theforeman/builder": "^4.2.0",
+    "@theforeman/vendor-dev": "^4.2.0",
     "axios-mock-adapter": "^1.10.0",
     "babel-eslint": "^10.0.3",
     "babel-jest": "^24.9.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -63,9 +63,6 @@
     "react-bootstrap": "^0.32.1",
     "react-helmet": "^5.2.0"
   },
-  "peerDependencies": {
-    "@theforeman/vendor": ">= 3.8.0"
-  },
```
johnpmitsch: This peer dependency is causing some errors during npm commands such as `npm ls`, mainly `unmet peer dependency`

My understanding of [the docs](https://npm.github.io/using-pkgs-docs/package-json/types/peerdependencies.html) is a peer dependency should be assumed to be installed at the same level of the dependency tree. So for this to be a peer dependency, the structure would be:
```
- Katello 1.0.0
  - foo
  - bar
- @theforeman/vendor
  - foo
  - bar
```
In this case, our katello package is checking if `@theforeman/vendor` is installed at the same level, and it is not because `@theforeman/vendor` is a child of the foreman package

I can't see a way to handle this with npm package.json entries, so I added a comment above as a key/value (JSON doesn't support comments). We used to not specify `@theforeman/vendor`, which works technically as it is assumed to come from foreman, but confusing to devs when checking the package.json for it's version, the comment will hopefully clear up the confusion without causing any errors or noise with npm operations.

johnpmitsch: here is a real world example of a peer dependency: https://github.com/facebook/react/blob/297f7588c46dc50f7439cf197d28d7686bd8d45d/packages/react-dom/package.json#L25

react-dom is assumed to be installed along react, but is not a child dep of it

This doesn't seem like it is the case for `@theforeman/vendor` and `katello`

johnpmitsch: To give more details

On master I see:

A warning on npm install:
```
npm WARN katello@1.0.0 requires a peer of @theforeman/vendor@>= 3.8.0 but none is installed. You must install peer dependencies yourself.
```

An error on `npm ls`:
```
npm ERR! peer dep missing: @theforeman/vendor@>= 3.8.0, required by katello@1.0.0
```

both of which are removed with this change

PR #8613, opened 2020-03-17
Fixes #29013 - Can view record in different organization

sjha4:
- [x] Add to other pages that need this.

**Steps to reproduce:**
1. Create a sync plan/product/activation-key etc under org1
2. Go to details page of the created resource
3. Change organization to a different organization.
4. User able to see the resource. You can also see the resource if you directly go to the record URL with it's id.

**Proposed Approach:**
Show a warning notification in this case - when user has access to the record but is in a different org that the record belongs to.

### DIFF
--- a/app/controllers/katello/api/v2/api_controller.rb
+++ b/app/controllers/katello/api/v2/api_controller.rb
@@ -198,9 +198,9 @@ module Katello
     end
 
     def check_resource_organization(resource, organization_id)
-      return unless resource && resource&.organization_id && organization_id
+      return unless resource && resource.try(:organization_id) && organization_id
       if resource.organization_id != organization_id.to_i
-        fail HttpErrors::BadRequest, _("The requested resource does not belong to the specified organization")
+        fail HttpErrors::BadRequest, _("The requested resource does not belong to the specified Organization")
       end
     end
 
--- a/app/controllers/katello/api/v2/api_controller.rb
+++ b/app/controllers/katello/api/v2/api_controller.rb
@@ -192,6 +192,18 @@ module Katello
       end
     end
 
+    def respond_for_show(options = {})
+      check_resource_organization(options[:resource], params[:organization_id])
+      super
+    end
+
+    def check_resource_organization(resource, organization_id)
+      return unless resource && resource&.organization_id && organization_id
+      if resource.organization_id != organization_id.to_i
+        fail HttpErrors::BadRequest, _("The requested resource does not belong to the specified organization")
+      end
+    end
+
     def find_host_with_subscriptions(id, permission)
       @host = resource_finder(::Host::Managed.authorized(permission, ::Host::Managed), id)
       fail HttpErrors::BadRequest, _("Host has not been registered with subscription-manager") if @host.subscription_facet.nil?
--- a/app/controllers/katello/api/v2/api_controller.rb
+++ b/app/controllers/katello/api/v2/api_controller.rb
@@ -198,7 +198,7 @@ module Katello
     end
 
     def check_resource_organization(resource, organization_id)
-      return unless resource && resource.try(:organization_id) && organization_id
+      return unless resource.try(:organization_id) && organization_id
       if resource.organization_id != organization_id.to_i
         fail HttpErrors::BadRequest, _("The requested resource does not belong to the specified Organization")
       end
--- a/app/controllers/katello/api/v2/environments_controller.rb
+++ b/app/controllers/katello/api/v2/environments_controller.rb
@@ -67,7 +67,7 @@ module Katello
     param :id, :number, :desc => N_("ID of the environment"), :required => true
     param :organization_id, :number, :desc => N_("ID of the organization")
     def show
-      respond
+      respond(resource: @environment)
     end
 
     api :POST, "/environments", N_("Create an environment")
--- a/app/controllers/katello/api/v2/host_collections_controller.rb
+++ b/app/controllers/katello/api/v2/host_collections_controller.rb
@@ -19,8 +19,9 @@ module Katello
 
     api :GET, "/host_collections/:id", N_("Show a host collection")
     param :id, :number, :desc => N_("Id of the host collection"), :required => true
+    param :organization_id, :number, :desc => N_("organization identifier"), :required => false
     def show
-      respond
+      respond(:resource => @host_collection)
     end
 
     api :GET, "/host_collections", N_("List host collections")
--- a/app/controllers/katello/api/v2/products_controller.rb
+++ b/app/controllers/katello/api/v2/products_controller.rb
@@ -86,7 +86,6 @@ module Katello
       respond_for_create(:resource => product)
     end
 
-    api :GET, "/organizations/:organization_id/products/:id", N_("Show a products")
     api :GET, "/products/:id", N_("Show a product")
     param :organization_id, :number, :desc => N_("Organization ID")
     param :id, :number, :desc => N_("product numeric identifier"), :required => true
--- a/app/controllers/katello/api/v2/products_controller.rb
+++ b/app/controllers/katello/api/v2/products_controller.rb
@@ -86,7 +86,9 @@ module Katello
       respond_for_create(:resource => product)
     end
 
+    api :GET, "/organizations/:organization_id/products/:id", N_("Show a products")
     api :GET, "/products/:id", N_("Show a product")
+    param :organization_id, :number, :desc => N_("Organization ID")
     param :id, :number, :desc => N_("product numeric identifier"), :required => true
     def show
       find_product(:includes => [{:root_repositories => {:repositories => :environment}}])
--- a/app/controllers/katello/api/v2/repositories_controller.rb
+++ b/app/controllers/katello/api/v2/repositories_controller.rb
@@ -260,6 +260,7 @@ module Katello
 
     api :GET, "/repositories/:id", N_("Show a repository")
     param :id, :number, :required => true, :desc => N_("repository ID")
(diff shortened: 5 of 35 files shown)


sjha4:
[test katello]

engines/bastion_katello/test/common/check-organization.service.test.js
```diff
@@ -0,0 +1,44 @@
...
+    CurrentOrganization = '1';
+    $provide.value('Notification', Notification);
+    $provide.value('CurrentOrganization', CurrentOrganization);
+  }));
+
+  beforeEach(inject(function($injector) {
+    CheckOrganizationHandler = $injector.get('CheckOrganizationHandler');
+  }));
+
+  describe("Provides a function to handle organization matching", function() {
```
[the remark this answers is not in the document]
sjha4: Updated to use describe("Organization-matching function")

sjha4:
[test katello]

sjha4:
> Sorry, meant to say this before: Instead of "selected organization" I think it would be helpful to have the org name in the message: `Resource does not belong to organization 'org1'` Thoughts?

@jeremylenz : So I was looking at doing this and this involves another API call to fetch the organization name from API using CurrentOrganization. I am not sure we should be doing that for a notification unless the existing message is not clear enough. Thoughts?

jeremylenz:
@sjha4 ah yeah, I'm fine just leaving it then!

sjha4:
[test katello]

sjha4:
@jturel , @jlsherrill : Wanted to run this by you..
With this, (Actually even without this) the behavior on subscription page vs other katello pages is different. 
On subscription detail page, for users outside the org of the subscription irrespective of their RBAC, we show an error message like the subscription is not relevant to user/organization and don't show any details.
 On other katello pages, we are adding a warning message like image above and allowing users to view records outside their selected organization if they have role based access to it.

sjha4:
[test katello]

jturel:
I think this is a good short-term fix. Although, I think it's incorrect to show a resource if it does not belong to the currently selected organization regardless of permissions. We're talking about broad API changes to fix that, though.

IMO the behavior should be like this:

As a user attempting to view an existing resource...
- I get a 403 if I am lacking permissions (API/UI)
- I get a 404 if I am permitted but the resources organization does not match Organization.current (UI case)
- I get a 400 if I am permitted but I have specified organization_id parameter which doesn't match the resource (API case)

jlsherrill:
I generally agree, but in a lot of cases the server doesn't have the 'current' org within the URL.  While it could check 'Org.current', we've tried to keep the api stateless and any checking of Org.current would make it not stateless.  So a request to  /katello/api/v2/repository/:id   wouldn't know what the org is.  The UI could detect this and throw a 'fake' HTTP code page i guess, but maybe we should add proper org-scoped URLs to get us there.

jturel:
> I generally agree, but in a lot of cases the server doesn't have the 'current' org within the URL. While it could check 'Org.current', we've tried to keep the api stateless and any checking of Org.current would make it not stateless. So a request to /katello/api/v2/repository/:id wouldn't know what the org is. The UI could detect this and throw a 'fake' HTTP code page i guess, but maybe we should add proper org-scoped URLs to get us there.

Good point. I guess we should always be scoping by organization_id in the UI then

sjha4:
@jturel @jlsherrill : Wanted to understand expectations with this PR and any to-do items here? Should I add org checks on these show methods in the API and return 404s? Or is warning on the UI the way to go?

jturel:
It'd be cool to have a generic way to solve this problem in the API.

With that in mind my wishful thinking would be for this PR to:

1) create an`enforce_resource_organization` filter for the API controller which might look like:
```
def enforce_resource_organization(resource = @resource)
  if resource && resource.respond_to?(:organization_id) && params[:organization_id] && resource.organization_id != params[:organization_id]
    fail HttpBadRequest("Requested resource does not exist in the specified organization")
  end
end
```
2) For Sync Plan #show, use `find_resource` if it does not already

3) For Sync Plan details, specify `organization_id` via UI API call if not already.

3) Keep your UI change here to catch the cases not otherwise handled

Over time we can fix the other areas to be come more consistent.

This is just my idea, not my specific request, but I'm ready to work with you on it if you're up for it!

app/controllers/katello/api/v2/api_controller.rb
```diff
@@ -192,6 +192,14 @@ def find_default_organization_and_or_environment
...
     end
 
+    def respond_for_show(options = {})
+      if (options.try(:[], :resource).try(:organization_id) && params[:organization_id] &&
+         (options[:resource].organization_id.to_s != params[:organization_id].to_s))
+        fail HttpErrors::NotFound, _("Resource does not belong to the specified organization")
+      end
+      super
+    end
+
```
sjha4: So using this we can check for resource organization for show methods if organization id is passed to the API. For cases where organization id is not passed to the API, the UI will show the warning as earlier..

app/controllers/katello/api/v2/api_controller.rb
```diff
@@ -192,6 +192,14 @@ def find_default_organization_and_or_environment
       end
     end
 
+    def respond_for_show(options = {})
+      if (options.try(:[], :resource).try(:organization_id) && params[:organization_id] &&
```
jturel: I'd like to suggest a bit of a refactor here, to separate the concerns:

```
    def respond_for_show(options = {})                                                                                                                                                        
      check_resource_organization(options[:resource], params[:organization_id])                                                                                                               
      super                                                                                                                                                                                   
    end                                                                                                                                                                                       
                                                                                                                                                                                              
    def check_resource_organization(resource, organization_id)                                                                                                                                
      return unless resource && resource&.organization_id && organization_id                                                                                                                  
                                                                                                                                                                                              
      if resource.organization_id != organization_id.to_i                                                                                                                                     
        fail HttpErrors::BadRequest, _("The requested resource does not belong to the specified Organization")                                                                                              
      end                                                                                                                                                                                     
    end
```

I think this makes the code more readable by exiting out of `check_resource_organization` if we don't have all of the data to make the comparison. And so the comparison is easier to read! Another distinction: we should raise ::BadRequest instead of ::NotFound because the request is invalid in nature. I also made the message a little 'friendlier'

engines/bastion_katello/app/assets/javascripts/bastion_katello/common/check-organization.service.js
```diff
@@ -0,0 +1,21 @@
...
+    'use strict';
+
+    /**
+     * @ngdoc service
+     * @name  Bastion.common.service:CheckOrganizationHandler
+     *
+     * @description
+     *   Provides common functionality in handling Resource/Organization warnings
+     */
+    function CheckOrganizationHandler(CurrentOrganization, Notification, translate) {
```
jturel: Now that we have messaging coming out of the API controller I am wondering if we should include this any longer. It kind of introduces technical debt since we ideally handle this at the API now.

I also found that we have something in bastion_katello called 'ApiErrorHandler' and we can make it work with your API change like this:

```
        function handleError(response, $scope, defaultErrorMessage) {                                                                                                                         
            var errors = (response.data && response.data.errors) ? response.data.errors : [];                                                                                                 
            var error = (response.data && response.data.error) ? response.data.error : null;                                                                                                  
                                                                                                                                                                                              
            if (errors.length > 0) {                                                                                                                                                          
                angular.forEach(errors, function (error) {                                                                                                                                    
                    Notification.setErrorMessage(error);                                                                                                                                      
                });                                                                                                                                                                           
            } else if (error && response.status >= 400 && response.status <= 404) {                                                                                                           
                Notification.setErrorMessage(error.message);                                                                                                                                  
            } else {                                                                                                                                                                          
                Notification.setErrorMessage(defaultErrorMessage);                                                                                                                            
            }                                                                                                                                                                                 
                                                                                                                                                                                              
            if ($scope && $scope.hasOwnProperty('panel')) {                                                                                                                                   
                $scope.panel.error = true;                                                                                                                                                    
            }                                                                                                                                                                                 
        } 
```

This way, for 400 and 404 errors we can show the real message. I think that's a lot more informative than 'Something went wrong while retrieving the resource' but we still fall back to it.

I realize this will not achieve the same effect as your CheckOrganizationService because we don't pass the organization_id parameter everywhere in the UI. I think it would be worthwhile to:

- remove the CheckOrganizationService
- improve the ApiErrorHandler per my above example
- wherever you've added CheckOrganizationService, check if we are passing organization_id. If not, let's file an issue to fix those areas in 3.16. I'd be open to fixing them here, but up to you.

sjha4: I was testing this and it doesn't seem like this refactoring is needed. We do pass the api error message to the handleError method and it displays accordingly.

jturel:
[test katello]

jturel:
@sjha4 this looks really good! I noticed just two things:

- On content host details, rather than showing the new error, I see "something went wrong when loading the resource"
- lifecycle environment #show seems to not work, but can be fixed like this:
```
+++ b/app/controllers/katello/api/v2/environments_controller.rb
@@ -67,7 +67,7 @@ module Katello
     param :id, :number, :desc => N_("ID of the environment"), :required => true
     param :organization_id, :number, :desc => N_("ID of the organization")
     def show
-      respond
+      respond(resource: @environment)
     end
```

sjha4:
> * On content host details, rather than showing the new error, I see "something went wrong when loading the resource"

The way the content host page behaves, the best I was able to do is to display the api error `Resource host not found by id '1'` that comes from the foreman api instead of the default "something went wrong when loading the resource"

PR #8615, opened 2020-03-17
Fixes #29368 - skips planning errata installation 

jjeffers:
This PR skips errata installation on hosts when there is no applicable errata for that host.

To test this, add content hosts to a host collection. 

Create a content view that includes a repository that includes errata (like https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/). Create a version of that content view and publlsh it.

Add one of the hosts in the host collection to the content view.

Install `walrus` package, then downgrade that package on the host that belongs to the content view. Add that host to the content view. There should now be applicable errata for that host, but not others in the host collection.

Go to *Host Collections* -> *Errata Installation*. Note that the number of affected hosts should be 1, but go ahead and select an applicable errata and then "Install Selected".

After the task has completed, examine the dynflow console. There should be a subplan for the ApplicableErrataInstall action for each host. Hoever, there should only be planned and executed actions (Actions::Katello::Host::Erratum::Install, et al) _only_ for hosts that would have applicable errata.

(Prevously every host would have the Erratum::Install action planned for it.)

### DIFF
--- a/app/lib/actions/katello/host/erratum/applicable_errata_install.rb
+++ b/app/lib/actions/katello/host/erratum/applicable_errata_install.rb
@@ -12,8 +12,18 @@ module Actions
 
             applicable_errata = host.content_facet.applicable_errata
             applicable_errata = applicable_errata.with_identifiers(errata_ids) unless options[:update_all]
-            plan_action(Actions::Katello::Host::Erratum::Install, host, applicable_errata.pluck(:errata_id))
-            plan_self(:hostname => host.name)
+            if applicable_errata.empty?
+              plan_self(:hostname => host.name, :skip => true)
+            else
+              plan_action(Actions::Katello::Host::Erratum::Install, host, applicable_errata.pluck(:errata_id))
+              plan_self(:hostname => host.name, :skip => false)
+            end
+          end
+
+          def run
+            if input[:skip]
+              output[:response] = _("No applicable errata for %s, skipping" % input[:hostname])
+            end
           end
 
           def humanized_name
--- a/test/actions/katello/host/applicable_errata_install_test.rb
+++ b/test/actions/katello/host/applicable_errata_install_test.rb
@@ -1,0 +1,51 @@
+require 'katello_test_helper'
+
+module ::Actions::Katello::Host::Erratum
+  class TestBase < ActiveSupport::TestCase
+    include Dynflow::Testing
+    include Support::Actions::Fixtures
+
+    class Installtest < TestBase
+      let(:action_class) { ::Actions::Katello::Host::Erratum::ApplicableErrataInstall }
+
+      let(:uuid) { 'uuid' }
+      let(:content_facet) { katello_content_facets(:content_facet_one) }
+      let(:host) do
+        host_mock = mock('a_host', content_facet: content_facet).mimic!(::Host::Managed)
+        host_mock.stubs('name').returns('foobar')
+        host_mock
+      end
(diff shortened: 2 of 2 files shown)


jturel:
needs a rebase :)

jturel:
@jjeffers looks like this needs another rebase .. somehow :ballot_box_with_check:

test/actions/katello/host/applicable_errata_install_test.rb
```diff
@@ -0,0 +1,51 @@
...
+  class TestBase < ActiveSupport::TestCase
+    include Dynflow::Testing
+    include Support::Actions::Fixtures
+
+    class Installtest < TestBase
+      let(:action_class) { ::Actions::Katello::Host::Erratum::ApplicableErrataInstall }
+
+      let(:uuid) { 'uuid' }
+      let(:content_facet) { mock('a_system').mimic!(::Katello::Host::ContentFacet) }
+      let(:host) do
```
jturel: I'd like to see one of the host fixtures here  rather than a mock - I think we have some with content facets set up!

app/lib/actions/katello/host/erratum/applicable_errata_install.rb
```diff
@@ -12,7 +12,9 @@ def plan(host, options = {})
 
             applicable_errata = host.content_facet.applicable_errata
             applicable_errata = applicable_errata.with_identifiers(errata_ids) unless options[:update_all]
-            plan_action(Actions::Katello::Host::Erratum::Install, host, applicable_errata.pluck(:errata_id))
+            unless applicable_errata.empty?
+              plan_action(Actions::Katello::Host::Erratum::Install, host, applicable_errata.pluck(:errata_id))
```
jturel: Any opinion on providing a message to the task output here (I think it's possible) to the effect of "No applicable errata, skipping". I'm imagining a user or support comparing the task details between the two variants and wondering about the differences in the sub-plans

test/actions/katello/host/applicable_errata_install_test.rb
```diff
@@ -0,0 +1,51 @@
...
+  class TestBase < ActiveSupport::TestCase
+    include Dynflow::Testing
+    include Support::Actions::Fixtures
+
+    class Installtest < TestBase
+      let(:action_class) { ::Actions::Katello::Host::Erratum::ApplicableErrataInstall }
+
+      let(:uuid) { 'uuid' }
+      let(:content_facet) { katello_content_facets(:content_facet_one) }
+      let(:host) do
```
jturel: I thought we had a fixture for a host with a content facet (seems like I was wrong?). If not, this will do to avoid unnecessary mocking: `@host = FactoryBot.create(:host, :with_content)`

jjeffers: Seems like attempting to use the factory for the host begins a cascade of dependencies for content views, etc.

app/lib/actions/katello/host/erratum/applicable_errata_install.rb
jjeffers: How about a task output like so? (As seen in the raw task output for the ApplicableErrataInstall for a host that is skipped.)
Screenshot from 2020-03-25 08-21-32

jturel: Perfect!

test/actions/katello/host/applicable_errata_install_test.rb
jturel: Sounds like what you have is simpler and probably more appropriate for this case then. Let's leave it. Thanks for trying

PR #8616, opened 2020-03-18
Fixes #29366 - distinguish between Wrokstation and Server

ares:
I haven't tested in real life however this should be all we need, if it's not working it would be great if someone could take over.

Please see https://github.com/theforeman/foreman/pull/7528 for more details.

### DIFF
--- a/app/models/katello/rhsm_fact_parser.rb
+++ b/app/models/katello/rhsm_fact_parser.rb
@@ -61,6 +61,10 @@ module Katello
           os_attributes[:release_name] = release_name
         end
 
+        if facts['distribution.name'] == 'Red Hat Enterprise Linux Workstation'
+          os_name += '_Workstation'
+        end
+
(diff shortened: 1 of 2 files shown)


jlsherrill:
@ares couple of test failures and a rubocop failure.  What was the plan for puppet OS flipping it back to Red Hat?

I also think we need to update: https://github.com/Katello/katello/blob/master/app/models/katello/concerns/redhat_extensions.rb#L32-L41  to handle Workstation?

ares:
The plan is linked in the first comment - https://github.com/theforeman/foreman/pull/7528, see https://community.theforeman.org/t/rhel7-os-detection-based-on-facts-update-your-parsers/17843 for more details

I'd be happy if someone could take this over in katello repo, I can't tell when I'll get to fix tests and the rubocop.

sjha4:
[test katello]

sjha4:
Closing this and opening https://github.com/Katello/katello/pull/8735 with this commit.

PR #8620, opened 2020-03-19
Fixes #29381 - switch patternfly to npm for bastion

jlsherrill:


engines/bastion/package.json
```diff
@@ -18,7 +18,9 @@
     "karma-coverage": "^1.0.0",
     "karma-jasmine": "^1.0.2",
     "karma-ng-html2js-preprocessor": "^1.0.0",
-    "require-dir": "~1.2.0"
+    "require-dir": "~1.2.0",
+    "patternfly": "4.0.0-alpha.2",
+    "angular-patternfly": "4.0.0-alpha.2"
```
waldenraines: Maybe I'm not understanding the issue but do we really want `angular-patternfly` 4.0?

jlsherrill: notice that was there before (in bower.json).  This is not the new patternfly 4.0, this is the initial attempt at a patternfly 4.0 which from what i can tell was abandoned.   I'm just moving it to package.json and calling out the patternfly dep explicitly.

PR #8579, opened 2020-02-24
Fixes #29145 - Add custom evr type column to katello_rpms and katello_installed_packages

ianballou:
Includes custom sql types, functions, and triggers that automatically generate items in the new "evr" column for katello_rpms and katello_installed_packages. The migrations that add these types are all reversible. The logic for creating the evr type is mostly pulled from here: https://github.com/ggainey/pulp_startup/blob/master/applicability/evr_t.sql. The evr type is treated as a string by ActiveRecord, but this shouldn't be an issue since the EVR type shouldn't be used directly in Ruby code.

To test: 
1) Ensure there are some RPMs and InstalledPackages in your database.  
2) Install this RPM: https://fedorapeople.org/groups/katello/releases/yum/nightly/pulpcore/el7/x86_64/rh-postgresql12-postgresql-evr-0.0.1-2.el7.x86_64.rpm
2) Checkout my branch and run the migration (ignore the new warning about the evr_t type). 
3) Check that the existing RPMs and InstalledPackages have something other than `nil` for evr.  
4) Sync in or create some RPMs and see that the evr column for katello_rpms is updated. 
5) Create some InstalledPackages and see that evr is updated too.  The evr column will show up as a weird string, but that's okay since it'll be used in the background. 
6) Check that the evr columns are updated if the epochs, versions, or releases are updated.
7) Try rolling back the migration.

TODO:

- [x] Get postgresql-evr package into foreman-packaging (https://github.com/theforeman/foreman-packaging/pull/4795)
- [x] Get postgresql-evr installed on the Jenkins slaves

PR #8621, opened 2020-03-19
Fixes #29098 - Katello Applicability config and generation w/ RPMs and Errata

ianballou:
This PR introduces package applicability calculations (with RPMs and not modules) using information indexed in Katello alone.  New classes:

1) `Actions::Katello::Applicability::Hosts::Generate`
  - Generates applicability for a list of hosts
2) `Actions::Katello::Applicability::Host::Generate`
  - Generates applicability for a single host
3) `Actions::Katello::Applicability::Repository::Regenerate`
  - Generates applicability for all hosts bound to a repository
4) `Katello::Events::GenerateHostApplicability`
 - Event to allow for simultaneous applicability generation requests
5) `Katello::Applicability::ApplicableContentHelper`
  - A new helper class for Katello Applicability

Current issues/questions:
-  The decisions on which applicability agent to use are `if` statements at the moment.  Should there be some kind of applicability selector like the pulp selector?  I'm thinking not since there aren't too many tasks to switch.
  - This also relates to dropping the consumer-related code when we drop Pulp 2.  Should we start that separation now? Will the `if` statements be sufficient until then?

TODO
- [x] Add more tests

Blocked by https://github.com/Katello/katello/pull/8579

### DIFF
--- a/app/lib/actions/katello/repository/sync.rb
+++ b/app/lib/actions/katello/repository/sync.rb
@@ -59,7 +59,9 @@ module Actions
               plan_action(Pulp::Repository::Download, :pulp_id => repo.pulp_id, :options => {:verify_all_units => true}) if validate_contents && repo.yum?
               plan_action(Katello::Repository::MetadataGenerate, repo, :force => true) if skip_metadata_check && repo.yum?
               plan_action(Katello::Repository::ErrataMail, repo, nil, contents_changed)
-              plan_action(Pulp::Repository::RegenerateApplicability, :repository_id => repo.id, :contents_changed => contents_changed) if generate_applicability
+              if generate_applicability
+                regenerate_applicability(repo, contents_changed)
+              end
             end
             plan_self(:id => repo.id, :sync_result => output, :skip_metadata_check => skip_metadata_check, :validate_contents => validate_contents,
                       :contents_changed => contents_changed)
@@ -97,6 +99,14 @@ module Actions
           pulp_action = planned_actions(Pulp::Repository::Sync).first
           if (pulp_task = Array(pulp_action.external_task).first)
             pulp_task.fetch(:task_id)
+          end
+        end
+
+        def regenerate_applicability(repo, contents_changed)
+          if SETTINGS[:katello][:katello_applicability]
+            plan_action(Actions::Katello::Applicability::Repository::Regenerate, :repo_id => repo.id, :contents_changed => contents_changed)
+          else
+            plan_action(Pulp::Repository::RegenerateApplicability, :repository_id => repo.id, :contents_changed => contents_changed)
           end
         end
 
--- a/app/services/katello/applicability/applicable_content_helper.rb
+++ b/app/services/katello/applicability/applicable_content_helper.rb
@@ -1,0 +1,97 @@
+module Katello
+  module Applicability
+    class ApplicableContentHelper
+      attr_accessor :content_facet, :content_unit_class, :bound_library_instance_repos
+
+      def initialize(content_facet, content_unit_class, bound_library_instance_repos)
+        self.content_facet = content_facet
+        self.content_unit_class = content_unit_class
+        self.bound_library_instance_repos = bound_library_instance_repos
+      end
+
+      def calculate_and_import
+        if self.bound_library_instance_repos.any?
+          to_add, to_remove = applicable_differences
+          ActiveRecord::Base.transaction do
+            insert(to_add) unless to_add.blank?
+            remove(to_remove) unless to_remove.blank?
+          end
+        end
+      end
+
+      def fetch_content_ids
+        if self.content_unit_class == ::Katello::Erratum
+          # Query for all Errata ids that are attached to the host's applicable packages
+          query = 'select katello_repository_errata.erratum_id as id from katello_repository_errata
+                       inner join katello_erratum_packages
+                          on  katello_repository_errata.erratum_id  = katello_erratum_packages.erratum_id
+                       inner join katello_rpms
+                          on katello_rpms.nvra = katello_erratum_packages.nvrea
+                       inner join katello_content_facet_applicable_rpms on
+                           katello_content_facet_applicable_rpms.rpm_id = katello_rpms.id
+                       where
+                            "katello_content_facet_applicable_rpms"."content_facet_id" = :content_facet_id
+                       AND katello_repository_errata.repository_id IN (:repo_ids)'
+
+          return Katello::Erratum.find_by_sql([query, {content_facet_id: content_facet.id, repo_ids: self.bound_library_instance_repos} ]).map(&:id).uniq
+        elsif self.content_unit_class == ::Katello::ModuleStream
+          fail NotImplementedError
+        else
+          # Query for applicable RPM ids
+          return ::Katello::Rpm.joins("INNER JOIN katello_repository_rpms ON \
+                                      katello_rpms.id = katello_repository_rpms.rpm_id").
+                                      joins("INNER JOIN katello_installed_packages ON \
+                                            katello_rpms.name = katello_installed_packages.name AND \
+                                            katello_rpms.arch = katello_installed_packages.arch AND \
+                                            katello_rpms.evr > katello_installed_packages.evr").
+                                            joins("INNER JOIN katello_host_installed_packages ON \
+                                                  katello_installed_packages.id = \
+                                                  katello_host_installed_packages.installed_package_id WHERE \
+                                                  katello_repository_rpms.repository_id in \
+                                                  (#{self.bound_library_instance_repos.join(',')}) \
+                                                  and katello_host_installed_packages.host_id = \
+                                                  #{self.content_facet.host.id}").pluck(:id).uniq
+        end
+      end
+
+      def applicable_differences
+        consumer_ids = content_facet.send(applicable_units).pluck("#{content_unit_class.table_name}.id")
+        content_ids = fetch_content_ids
+
+        to_remove = consumer_ids - content_ids
+        to_add = content_ids - consumer_ids
+
+        [to_add, to_remove]
+      end
+
+      def insert(applicable_ids)
+        unless applicable_ids.empty?
+          inserts = applicable_ids.map { |applicable_id| "(#{applicable_id.to_i}, #{content_facet.id.to_i})" }
+          sql = "INSERT INTO #{content_facet_association_class.table_name} (#{content_unit_association_id}, content_facet_id) VALUES #{inserts.join(', ')}"
+          ActiveRecord::Base.connection.execute(sql)
+        end
+      end
+
+      def remove(applicable_ids)
+        content_facet_association_class.where(:content_facet_id => content_facet.id, content_unit_association_id => applicable_ids).delete_all
+      end
+
+      def content_unit_association_id
+        "#{content_unit_class.name.demodulize.underscore}_id".to_sym
+      end
+
+      def content_facet_association_class
+        # Example: ContentFacetErratum
+        self.content_unit_class.content_facet_association_class
+      end
+
+      def content_units
+        content_unit_class.name.demodulize.pluralize.underscore.to_sym
+      end
+
+      def applicable_units
+        "applicable_#{content_units}".to_sym
+      end
+    end
+  end
+end
--- a/test/actions/katello/applicability/generate_applicability_test.rb
+++ b/test/actions/katello/applicability/generate_applicability_test.rb
@@ -1,0 +1,33 @@
+require 'katello_test_helper'
+
+module ::Actions::Katello::Applicability::Host
+  class GenerateApplicabilityTest < ActiveSupport::TestCase
+    include Dynflow::Testing
+    include Support::Actions::Fixtures
+    include FactoryBot::Syntax::Methods
+
+    before :all do
+      User.current = users(:admin)
+      @host = FactoryBot.build(:host, :with_content, :with_subscription, :content_view => katello_content_views(:library_dev_view),
+                               :lifecycle_environment => katello_environments(:library))
+      SETTINGS[:katello][:katello_applicability] = true
+    end
+
+    after :all do
+      SETTINGS[:katello][:katello_applicability] = false
+    end
+
+    describe 'Host Generate Applicability using Katello Applicability' do
+      let(:action_class) { ::Actions::Katello::Host::GenerateApplicability }
+
+      it 'plans' do
+        action = create_action action_class
+
+        plan_action action, [@host]
+
+        assert_action_planed_with(action, Actions::Katello::Applicability::Hosts::Generate,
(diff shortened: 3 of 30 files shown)


app/lib/actions/katello/katello_applicability/generate_host_applicability.rb
```diff
@@ -0,0 +1,69 @@
...
+
+        def run
+          content_facet = ::Host.find(input[:host_id]).content_facet
+
+          content_uuids = ::Katello::Rpm.joins("inner join katello_installed_packages on \
+                               katello_rpms.name = katello_installed_packages.name and \
+                               katello_rpms.evr > katello_installed_packages.evr").
+                               joins("INNER JOIN katello_host_installed_packages ON \
+                                     katello_installed_packages.id = katello_host_installed_packages.installed_package_id \
+                                     WHERE katello_host_installed_packages.host_id = #{input[:host_id]}").pluck(:pulp_id)
```
jlsherrill: couldn't you just grab the rpm id directly?  instead of converting to pulp_ids and then converting back to rpm ids on line 38?

app/lib/actions/katello/katello_applicability/generate_host_applicability.rb
```diff
@@ -0,0 +1,69 @@
...
+        input_format do
+          param :host_id, Integer
+        end
+
+        def run
+          content_facet = ::Host.find(input[:host_id]).content_facet
+
+          content_uuids = ::Katello::Rpm.joins("inner join katello_installed_packages on \
+                               katello_rpms.name = katello_installed_packages.name and \
+                               katello_rpms.evr > katello_installed_packages.evr").
```
jlsherrill: i think you also need to compare 'arch'

app/lib/actions/katello/katello_applicability/generate_host_applicability.rb
ianballou: I see what you mean.  I'll take a look at that.

app/lib/actions/katello/katello_applicability/generate_host_applicability.rb
```diff
@@ -0,0 +1,69 @@
...
+          content_unit_class.name.demodulize.pluralize.underscore.to_sym
+        end
+
+        def applicable_units
+          "applicable_#{content_units}".to_sym
+        end
+      end
+    end
+  end
+end
```
jlsherrill: This looks like it only does 'packages',  IS there another issue to then calculate errata from this list?

jlsherrill: or do you want to do it here too?

ianballou: Yeah I was starting with packages first.  I'm okay with doing errata applicability in this PR too, it might not be a big next step.

jlsherrill: i'd also probably either move this into a model or a service for easier testing

ianballou: That sounds good.  I'm thinking it could mostly fit in a new `app/services/katello/katello_applicability/applicable_content_helper.rb`

ianballou:
Errata applicability is going to be part of this PR as well.

jturel:
I have nothing to offer but some suggestions about the naming of the various classes:

What do you think about:

`Actions::Katello::Applicability::Hosts::Generate` 

`Actions::Katello::Applicability::Host::Generate`

`Actions::Katello::Applicability::Repository::Regenerate`

`Katello::Events::GenerateHostApplicability`

`Katello::Applicability::ApplicableContentHelper`

To me this is more concise and consistent with existing conventions

ianballou:
@jturel I like that naming scheme!  I was going to ask for some help there.

app/services/katello/applicability/applicable_content_helper.rb
```diff
@@ -0,0 +1,81 @@
...
+
+        # FIXME: Debug return
+        return [to_add, to_remove]
+      end
+
+      def content_ids
+        if self.content_unit_class == ::Katello::Erratum
+          # Query for all Errata ids that are attached to the host's applicable packages
+          return ::Katello::Erratum.where(id: ::Katello::ErratumPackage.
+                                          where(nvrea: self.content_facet.applicable_rpms.pluck(:nvra)).
```
ianballou: Is this comparison between `nvrea` and `nvra` okay? It seem like internally they end up being the same.  Perhaps I should compare `filename` instead?

ianballou:
One question for any reviewers: are my tasks set up correctly? Wasn't sure exactly which ones should be considered an EntryAction.  Also I think I need to clean up the task names, currently the tasks just show up as `Generate` in the task browser.

app/lib/actions/katello/applicability/host_generate.rb
```diff
@@ -0,0 +1,32 @@
...
+            param :host_id, Integer
+          end
+
+          def run
+            content_facet = ::Host.find(input[:host_id]).content_facet
+
+            package_content_ids = ::Katello::Applicability::ApplicableContentHelper.new(::Katello::Rpm, content_facet).calculate_and_import
+            erratum_content_ids = ::Katello::Applicability::ApplicableContentHelper.new(::Katello::Erratum, content_facet).calculate_and_import
+
+            output[:response] = "package_content_ids were #{package_content_ids}. erratum_content_ids were #{erratum_content_ids}."
```
jlsherrill: we probably don't need this in the task response, we could include a count?

ianballou: Yeah, that was for me during debugging

app/services/katello/applicability/applicable_content_helper.rb
```diff
@@ -0,0 +1,80 @@
...
+          fail NotImplementedError
+        else
+          # Query for applicable RPM ids
+          return ::Katello::Rpm.joins("INNER JOIN katello_installed_packages ON \
+                                      katello_rpms.name = katello_installed_packages.name AND \
+                                      katello_rpms.arch = katello_installed_packages.arch AND \
+                                      katello_rpms.evr > katello_installed_packages.evr").
+                                      joins("INNER JOIN katello_host_installed_packages ON \
+                                            katello_installed_packages.id = katello_host_installed_packages.installed_package_id \
+                                            WHERE katello_host_installed_packages.host_id = #{self.content_facet.host.id}").pluck(:id)
```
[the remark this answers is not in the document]
ianballou: This should be fixed now.  Also added a `.uniq` since I was seeing key errors with duplicate repositories.

ianballou:
[test katello]

app/lib/actions/katello/repository/sync.rb
```diff
@@ -97,6 +99,14 @@ def pulp_task_id
           end
         end
 
+        def regenerate_applicability(repo)
+          if SETTINGS[:katello][:katello_applicability]
+            plan_action(Actions::Katello::Applicability::Repository::Regenerate, :repo_id => repo.id)
```
jlsherrill, line 104: We probably want to pass contents_changed here, to skip applicability calcs if nothing has changed.  We can tackle that as a 2nd issue, or in this one. Preferences?

app/services/katello/applicability/applicable_content_helper.rb
```diff
@@ -0,0 +1,88 @@
...
+                                      katello_rpms.id = katello_repository_rpms.rpm_id").
+                                      joins("INNER JOIN katello_installed_packages ON \
+                                            katello_rpms.name = katello_installed_packages.name AND \
+                                            katello_rpms.arch = katello_installed_packages.arch AND \
+                                            katello_rpms.evr > katello_installed_packages.evr").
+                                            joins("INNER JOIN katello_host_installed_packages ON \
+                                                  katello_installed_packages.id = \
+                                                  katello_host_installed_packages.installed_package_id WHERE \
+                                                  katello_repository_rpms.repository_id in \
+                                                  (#{self.content_facet.bound_repositories.pluck(:id).join(',')}) \
```
jlsherrill, line 44: it looks like you're using the 'bound' repositories, and not the library instances?

app/services/katello/applicability/applicable_content_helper.rb
```diff
@@ -0,0 +1,88 @@
...
+      end
+
+      def content_ids
+        if self.content_unit_class == ::Katello::Erratum
+          # Query for all Errata ids that are attached to the host's applicable packages
+          return ::Katello::Erratum.joins("INNER JOIN katello_repository_errata ON katello_errata.id = \
+                                          katello_repository_errata.erratum_id").
+                                          where(id: ::Katello::ErratumPackage.
+                                                where(nvrea: self.content_facet.applicable_rpms.pluck(:nvra)).
+                                                pluck(:erratum_id)).pluck(:id).uniq
```
jlsherrill, line 29: I think we have to use a similar mechanism as below to limit the errata to what is in the host's bound repositories (or the library instances of them)

test/actions/katello/applicability/generate_applicability_test.rb
```diff
@@ -0,0 +1,29 @@
...
+  class GenerateApplicabilityTest < ActiveSupport::TestCase
+    include Dynflow::Testing
+    include Support::Actions::Fixtures
+    include FactoryBot::Syntax::Methods
+
+    before :all do
+      User.current = users(:admin)
+      @host = FactoryBot.build(:host, :with_content, :with_subscription, :content_view => katello_content_views(:library_dev_view),
+                               :lifecycle_environment => katello_environments(:library))
+      SETTINGS[:katello][:katello_applicability] = true
```
jlsherrill, line 13: anytime we set this to true, we need to set it back to false in the test teardown, that should make it not affect other tests

app/lib/actions/katello/repository/sync.rb
ianballou: It sounds like a relatively minor change for the PR, I'll add it in.

app/services/katello/applicability/applicable_content_helper.rb
ianballou: Are you talking about the `bound_repository`s' library instances?  If I check my host's `bound_repositories`, all of their `library_instance`s return `nil`.

ianballou: They are nil because they are already library instances.  I need to grab the library instance if `library_instance` isn't nil.

app/services/katello/applicability/applicable_content_helper.rb
```diff
@@ -0,0 +1,93 @@
...
+      def content_ids
+        bound_library_instance_repos = self.content_facet.bound_repositories.collect do |repo|
+          repo.library_instance.nil? ? repo : repo.library_instance
+        end
+
+        if self.content_unit_class == ::Katello::Erratum
+          # Query for all Errata ids that are attached to the host's applicable packages
+          return ::Katello::Erratum.joins("INNER JOIN katello_repository_errata ON \
+          katello_errata.id = katello_repository_errata.erratum_id").
+          where(id: ::Katello::ErratumPackage.where(nvrea: self.content_facet.applicable_rpms.pluck(:nvra)).
```
jlsherrill, line 32: if you change to .select(:nvra)  it avoids a 2nd query

app/services/katello/applicability/applicable_content_helper.rb
```diff
@@ -0,0 +1,93 @@
...
+                pluck(:erratum_id)).where("katello_repository_errata.repository_id in \
```
jlsherrill, line 33: if you use .select(:erratum_id)  it also avoids another query

jlsherrill:
i'd add an index on

katello_rpms: (name, arch, evr)
katello_erratum_packages (erratum_id, nvrea)

that seemed to avoid some full table scans in postgresql

app/services/katello/applicability/applicable_content_helper.rb
```diff
@@ -0,0 +1,91 @@
...
+        end
+
+        if self.content_unit_class == ::Katello::Erratum
+          # Query for all Errata ids that are attached to the host's applicable packages
+          return ::Katello::Erratum.joins("INNER JOIN katello_repository_errata ON \
+          katello_errata.id = katello_repository_errata.erratum_id").
+          where(id: ::Katello::ErratumPackage.where(nvrea: self.content_facet.applicable_rpms.select(:nvra)).
+                select(:erratum_id)).where("katello_repository_errata.repository_id in \
+                (#{bound_library_instance_repos.pluck(:id).join(',')})").pluck(:id).uniq
+        elsif self.content_unit_class == ::Katello::ModuleStream
```
jlsherrill, line 33: This errata query seems fairly slow for me, this was much faster:

```
          query = 'select katello_repository_errata.erratum_id  as id from katello_repository_errata
                       inner join katello_erratum_packages
                          on  katello_repository_errata.erratum_id  = katello_erratum_packages.erratum_id
                       inner join katello_rpms
                          on katello_rpms.nvra = katello_erratum_packages.nvrea
                       inner join katello_content_facet_applicable_rpms on
                           katello_content_facet_applicable_rpms.rpm_id = katello_rpms.id
                       where
                            "katello_content_facet_applicable_rpms"."content_facet_id" = :content_facet_id
                       AND katello_repository_errata.repository_id IN (:repo_ids)'

          Katello::Erratum.find_by_sql([query, {content_facet_id: content_facet.id, repo_ids: bound_library_instance_repos.pluck(:id)} ]).map(&:id).uniq

```

Creating an index on katello_rpms (nvra) seemed to help with the previous query, but backing it seems to not have a noticable effect with the new query, something to keep an eye on

app/services/katello/applicability/applicable_content_helper.rb
```diff
@@ -0,0 +1,91 @@
...
+                                                  (#{bound_library_instance_repos.pluck(:id).join(',')}) \
+                                                  and katello_host_installed_packages.host_id = \
+                                                  #{self.content_facet.host.id}").pluck(:id).uniq
+        end
+      end
+
+      def applicable_differences
+        consumer_ids = content_facet.send(applicable_units).pluck("#{content_unit_class.table_name}.id")
+        to_remove = consumer_ids - content_ids
+        to_add = content_ids - consumer_ids
```
jlsherrill, line 56: i realized this is calling content_ids twice! so its calling the queries 2x unnecessarily.

ianballou: Totally missed that! I'll cache content_ids then

ianballou:
@jlsherrill I've refactored the `bound_repos` related code like we discussed.

ianballou:
[test katello]

ianballou:
[test katello]

ianballou:
[test katello]

ianballou:
This PR wasn't rebased which might be why the previous one is failing,  I've rebased this one and we'll see if it causes a failure still.

PR #8622, opened 2020-03-23
refs #29271 - template macros return array instead of string

domitea:
Fixing correct result for several macros. Current implementation takes values from array and join them together. This fix remain array of values in array.

### DIFF
--- a/app/lib/katello/concerns/base_template_scope_extensions.rb
+++ b/app/lib/katello/concerns/base_template_scope_extensions.rb
@@ -25,7 +25,7 @@ module Katello
       end
 
       def host_subscriptions_names(host)
-        host.subscriptions.map(&:name).join(', ')
+        host.subscriptions.map(&:name)
       end
 
       def host_content_facet(host)
@@ -43,9 +43,9 @@ module Katello
       def host_products_names(host)
         products = host_products(host)
         if products
-          products.map(&:name).join(', ')
+          products.map(&:name)
         else
-          ''
+          []
         end
       end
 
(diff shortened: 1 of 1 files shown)


app/lib/katello/concerns/base_template_scope_extensions.rb
```diff
@@ -43,7 +43,7 @@ def host_products(host)
       def host_products_names(host)
         products = host_products(host)
         if products
-          products.map(&:name).join(', ')
+          products.map(&:name)
         else
           ''
```
tbrisker: should this be `[]` instead?

jturel: Likely the else can be removed entirely

jturel:
@domitea is the intent here that the templates will be responsible for turning the arrays into comma-separated strings?

app/lib/katello/concerns/base_template_scope_extensions.rb
tbrisker: i think the condition is needed because `host_products` can return nil

domitea:
@jturel yes, it is.

app/lib/katello/concerns/base_template_scope_extensions.rb
jturel: Oh :confounded:

tbrisker:
test failures are unrelated, due to added rubocop plugin (https://community.theforeman.org/t/katello-master-source-release-850-failed/17902)

jturel:
[test katello]

jturel:
[test katello]

PR #8624, opened 2020-03-23
Fixes #29395 - Address cops for rubocop-minitest

jturel:


jturel:
[test katello]

mmoll:
@jturel For the auto-correct to work better you'd need rubocop-minitest 0.8 (relased today)

jturel:
[test katello]

jturel:
> @jturel For the auto-correct to work better you'd need rubocop-minitest 0.8 (relased today)

Thanks, I did see that. I decided to change a lot of those to user assert_equal instead because it looked so ugly

PR #8625, opened 2020-03-24
Fixes #29397 - improve tracer upload performance

jturel:
- use active_record_import
- use delete_all vs destroy_all
- don't reach into candlepin to return the host

### DIFF
--- a/app/models/katello/concerns/host_managed_extensions.rb
+++ b/app/models/katello/concerns/host_managed_extensions.rb
@@ -208,11 +208,13 @@ module Katello
       end
 
       def import_tracer_profile(tracer_profile)
-        self.host_traces.destroy_all
+        host_traces.delete_all
+        traces = []
         tracer_profile.each do |trace, attributes|
-          self.host_traces.create!(:application => trace, :helper => attributes[:helper], :app_type => attributes[:type])
+          traces << { host_id: self.id, application: trace, helper: attributes[:helper], app_type: attributes[:type] }
         end
-        self.update_trace_status
+        Katello::HostTracer.import(traces, validate: false)
+        update_trace_status
       end
 
       def subscription_status
--- a/app/controllers/katello/api/rhsm/candlepin_proxies_controller.rb
+++ b/app/controllers/katello/api/rhsm/candlepin_proxies_controller.rb
@@ -151,7 +151,7 @@ module Katello
       User.as_anonymous_admin do
         @host.import_tracer_profile(params[:traces])
       end
-      render :json => Resources::Candlepin::Consumer.get(@host.subscription_facet.uuid)
+      render json: { displayMessage: _("Tracer profile uploaded successfully") }
     end
 
     def available_releases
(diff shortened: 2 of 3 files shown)


app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -208,11 +208,13 @@ def sync_package_associations(new_installed_package_ids)
...
       def import_tracer_profile(tracer_profile)
-        self.host_traces.destroy_all
+        host_traces.delete_all
+        traces = []
         tracer_profile.each do |trace, attributes|
-          self.host_traces.create!(:application => trace, :helper => attributes[:helper], :app_type => attributes[:type])
+          traces << { host_id: self.id, application: trace, helper: attributes[:helper], app_type: attributes[:type] }
         end
-        self.update_trace_status
+        Katello::HostTracer.import(traces)
```
jlsherrill: if we don't need validation (which we likely don't), you can add :validate => false to make it even faster.

jlsherrill:
these changes all make sense +1 to them, have you been able to 'simulate' what was seen in a users environment with concurrent uploads?

app/models/katello/concerns/host_managed_extensions.rb
jturel: Good call, I don't think we need validation here

jturel:
> these changes all make sense +1 to them, have you been able to 'simulate' what was seen in a users environment with concurrent uploads?

I didn't bother since it's a lot of setup ... 100+ systems with REX. Is there a simplified approach you can think of? What I keyed in on was the fact that when user than the REX job serially they had 100% success and no timeouts with `time katello-tracer-upload`. It regularly timed out with 100+ parallel jobs. That pointed me to a server side fix. I'm looking at another issue separately that brings the performance of that command into question, from the client side.  Though, I can't help but wonder if they are seeing performance problems both client and server side.

jlsherrill:
oh i definitely think its a server side fix for sure, I'm just not sure if these changes will fully solve the issue or further is needed.   We might be able to test this with a docker container made specifically to generate this kinda data (although that would be a lot of setup as well).  Would it be possible to have the user test a patch (even if the results come after we merge this pr).  

Do you have any timings showing what a single large upload took before and after the change? 
I'm just surprised that even the old code would take >15 minutes with ~160 clients trying to upload them, it really doesn't seem that large unless there was some sort of lock contention going on which these changes likely reduce considerably but may not fully eliminate.

jturel:
Have an idea on how to test. stay tuned

jturel:
Alright, I created 200 hosts with content-host-d and wrote this script: https://gist.github.com/jturel/1dd1ad2ae688569421044e3f2a7db57c

I tested before and after with and without the request in a Thread block. Got these results uploading 100 traces per host:
```
before in serial
total time: 147.558109362

before in parallel
total time: 48.315363307

after in serial
total time: 42.246104491

after parallel
total time: 22.000553987  
```

On one occasion I saw a few timeouts with the parallel test of the old code, if that means anything. I think the 15+ minute timeouts had something to do with the client-side, but this change definitely makes the server faster.

jlsherrill:
nice investigation!  I still suspect we aren't fully reproducing the user's issue, but i think this will go a long way

PR #8630, opened 2020-03-25
Fixes #29411 - add descriptions to objects

jlsherrill:
to ensure they pass when running with foreman


jturel:
did something change or is this a long-standing intermittence?

jlsherrill:
These tests were failing on foreman prs only, which has happened before and it occured only when comparing two null values, so something like:

assert_equal null, null

this never happens on katello and i never figured out what was different, why it wasn't happening on katello i never figured out.  It also doesn't seem like anything changed in the area, so why it just started happening on foreman tests is a mystery as well!

PR #8633, opened 2020-03-27
Fixes #29323 - publish a yum repository in a CV with filters

jjeffers:
WARNING: requires pulp_rpm 3.3.2

This PR allows publishing a yum repository in a content view using filters.

At this time the copy API will not affect drpms or modules. Rpm packges, package groups, package environments, errata, distribution trees, and metadatafiles are all copied as appropriate. What's appropriate? Distribution tree and metadatafile content is always copied. Other content types are copied if they are specified via a filter. Also any content that includes previously identified content for copy (for example an erratum that lists a package included in an initial package filter). 

PACKAGE (RPM) FILTERS

A common use case is to create a product with a yum repository. For this example, we'll use https://repos.fedorapeople.org/pulp/pulp/fixtures/rpm/ as the repository remote URL. After creating the repository in Katello, sync the repository. 

Note that the sync'd repository has 35 packages, 4 errata, and 2 package groups.

Then create a content view that includes this yum repository.

Create a yum name inclusion filter with the package name of "shark". 
Screenshot from 2020-05-04 15-33-03

Publish a new version of the content view.

There should now be a new content view (CV) version, but now with only 3 packages and 1 errata listed.
Screenshot from 2020-05-04 15-34-39

Note that the packages include packages "shark", but also associated packages from the included errata. The errata indicated the "shark" package, but also includes 2 other RPMs, so those are copied as well in the new version.

Screenshot from 2020-05-04 15-35-57

PACKAGE GROUP FILTERS

Similarly, create a content view, add the same repository. Create a new package group filter for the content view, targeting the "birds" package group.

Screenshot from 2020-05-05 08-45-03

Publish a new version and examine the packages, errata, etc of that version.
Screenshot from 2020-05-05 08-46-20
Note that the package group "birds" is copied into the new repository version, but only that group, since none of the other packages were listed as part of the other package group.
Screenshot from 2020-05-05 08-47-40

ERRATA FILTERS

Create a CV with the same repository, but add an errata filter that specifies a particular erratum by id
Screenshot from 2020-05-05 09-07-30

Publish that version. Observe the packages are only the ones includes in the errata, and the only errata included is the one indicated in the filter.
Screenshot from 2020-05-05 09-09-18
Screenshot from 2020-05-05 09-09-34


jjeffers:
[test katello]

app/services/katello/pulp3/repository/yum.rb
```diff
@@ -59,9 +60,54 @@ def partial_repo_path
...
+          content_unit_hrefs += source_repository.srpms.pluck(:pulp_id)
+          errata_to_include = filter_by_pulp_href(source_repository.errata, content_unit_hrefs)
+          content_unit_hrefs += errata_to_include.collect { |erratum| erratum.repository_errata.pluck(:erratum_pulp3_href) }.flatten.uniq
+
+          copy_units(source_repository.version_href, content_unit_hrefs, dependency_solving)
+        end
+
+        def package_group(package_group_href, options = {})
+          api.content_package_groups_api.read(package_group_href, options)
+        end
```
jlsherrill, line 107: I think this may already be handled here: https://github.com/Katello/katello/blob/master/app/services/katello/pulp3/package_group.rb  or at least might be a better place to put it if not

jjeffers: I don't even recall _why_ that exists!

ianballou:
I realized above that the filtering above wasn't working because of the included Sea_Erratum. However, the error I got above is still an issue.

jjeffers:
> Is this change in behaviour on purpose?
 
Yes, but it could be wrong. I'll follow up with @jlsherrill

jjeffers:
> Also is this PR supposed to support dependency solving?

There was an issue with the dependency solving settings in the copy API, so I have not yet tested that functionality.

app/models/katello/content_view_package_filter.rb
```diff
@@ -7,10 +7,14 @@ class ContentViewPackageFilter < ContentViewFilter
...
-    # repo repository that match parameters
+    # repo repository that match parameters if the repository content type
+    # is managed by pulp2.
+    #
+    # Returns a set of pulpcore hrefs (pulp_ids) if the repository content
+    # type is managed by pulpcore.
     #
     # @param repo [Repository] a repository containing packages to filter
-    # @return [Array] an array of hashes with MongoDB conditions
+    # @return [Array] an array of hashes with MongoDB conditions or pulpcore hrefs
```
ianballou, line 17: I'm a little confused here, how would `generate_clauses` return pulpcore hrefs?

jjeffers: That comment was left over from a very early revision, before I realized that overloading the existing methods with pulpcore specific support wasn't a good idea. It's misleading so thank you for noticing it!

jjeffers:
> I've found a difference between Pulp 2 and Pulp 3 with dep solving. If you include all versions of walrus with Pulp 3, you only get the two walrus packages. With Pulp 2, all of the walrus 0.71 deps are there. It looks like the include filter with Pulp 3 is excluding dependencies from other packages.

Can you send me the log of the copy_content data sent to pulp_rpm?

ianballou:
`40183:2020-05-21T17:14:44 [D|kat|] Calling API: RpmCopyApi.copy_content ...
40201:2020-05-21T17:14:44 [D|kat|] API called: RpmCopyApi#copy_content
40211: | {"pulp_href":"/pulp/api/v3/tasks/8ab062c4-a324-4c71-b091-328371afe010/","pulp_created":"2020-05-21T17:14:44.339451Z","state":"completed","name":"pulp_rpm.app.tasks.copy.copy_content","started_at":"2020-05-21T17:14:44.441502Z","finished_at":"2020-05-21T17:14:44.605646Z","error":null,"worker":"/pulp/api/v3/workers/faf1dc90-71ef-491a-b9f9-117aa33a2692/","parent_task":null,"child_tasks":[],"task_group":null,"progress_reports":[],"created_resources":["/pulp/api/v3/repositories/rpm/rpm/3dbe1db2-4692-45c0-81af-df1abe2f7815/versions/3/"],"reserved_resources_record":["/pulp/api/v3/repositories/rpm/rpm/d54fb797-434f-4ec1-aa60-7d673e474c4a/","/pulp/api/v3/repositories/rpm/rpm/3dbe1db2-4692-45c0-81af-df1abe2f7815/"]}
40215: | Data: #<PulpcoreClient::Task:0x00007f4fb5072d60 @pulp_href="/pulp/api/v3/tasks/8ab062c4-a324-4c71-b091-328371afe010/", @pulp_created=Thu, 21 May 2020 17:14:44 +0000, @state="completed", @name="pulp_rpm.app.tasks.copy.copy_content", @started_at=Thu, 21 May 2020 17:14:44 +0000, @finished_at=Thu, 21 May 2020 17:14:44 +0000, @worker="/pulp/api/v3/workers/faf1dc90-71ef-491a-b9f9-117aa33a2692/", @child_tasks=[], @progress_reports=[], @created_resources=["/pulp/api/v3/repositories/rpm/rpm/3dbe1db2-4692-45c0-81af-df1abe2f7815/versions/3/"], @reserved_resources_record=["/pulp/api/v3/repositories/rpm/rpm/d54fb797-434f-4ec1-aa60-7d673e474c4a/", "/pulp/api/v3/repositories/rpm/rpm/3dbe1db2-4692-45c0-81af-df1abe2f7815/"]>
`

@jjeffers let me know if you need any other logs

app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -124,8 +124,8 @@ def pulp3_configuration(config_class)
           config.ssl_client_key = ::Cert::Certs.ssl_client_key
           config.debugging = true
           config.logger = ::Foreman::Logging.logger('katello/pulp_rest')
-          config.username = self.setting(PULP3_FEATURE, 'username')
```
jjeffers, line 127: @ianballou I think this shouldn't be in there. I need to revert this change.

app/lib/actions/pulp3/repository/remove_all_units.rb
```diff
@@ -0,0 +1,18 @@
+module Actions
+  module Pulp3
+    module Repository
+      class RemoveAllUnits < Pulp3::AbstractAsyncTask
```
ianballou, line 4: Where do you use this action? I see `remove_all_content` being used but not through `RemoveAllUnits`

jjeffers: It _was_ used at one point to work around an issue in pulp_rpm. But it no longer is used. We should remove it!

PR #8640, opened 2020-03-31
Fixes #29449 - some fixes for ext. pulp3-server

m-bucher:
This is probably not the most efficient way to do it, so comments/improvements are welcome.

### DIFF
--- a/app/models/katello/concerns/smart_proxy_extensions.rb
+++ b/app/models/katello/concerns/smart_proxy_extensions.rb
@@ -124,6 +124,8 @@ module Katello
           config.ssl_client_key = ::Cert::Certs.ssl_client_key
           config.debugging = true
           config.logger = ::Foreman::Logging.logger('katello/pulp_rest')
+          config.username = self.setting(PULP3_FEATURE, 'username')
+          config.password = self.setting(PULP3_FEATURE, 'password')
         end
       end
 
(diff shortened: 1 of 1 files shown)


app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -124,6 +124,8 @@ def pulp3_configuration(config_class)
           config.ssl_client_key = ::Cert::Certs.ssl_client_key
           config.debugging = true
           config.logger = ::Foreman::Logging.logger('katello/pulp_rest')
+          config.username = self.setting(PULP3_FEATURE, 'username') 
+          config.password = self.setting(PULP3_FEATURE, 'password')
```
jlsherrill: the smart proxy has to be configured to actually send these settings: https://github.com/theforeman/smart_proxy_pulp/blob/master/lib/smart_proxy_pulp_plugin/pulp_plugin.rb#L10

otherwise i think adding this option is fine (although i don't see our installed deployments actually using it).

m-bucher: Like so: https://github.com/theforeman/smart_proxy_pulp/pull/23

app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -182,7 +184,7 @@ def pulp3_host!
         pulp3_uri!.host
       end
 
-      def pulp3_url(path = '/pulp/api/v3/')
+      def pulp3_url(path = '/pulp/api/v3')
```
jlsherrill, line 187: My guess is that you only hit this because you were pointing straight at the pulpcore-api rather than going through apache?  I don't see the issue with the normal proxied-apache setup, but I think the change makes sense :+1:

jlsherrill:
Not sure why travis-ci is stuck here.  Going to try to close and reopen

PR #8638, opened 2020-03-30
Fixes #28969 - Adding 'Subscriptions expiring soon' email to email preferences

sseelam2:
Adding a new email notification to the Email Preferences tab of the User edit screen. Users can subscribe to get notifications about subscriptions expiring in 30 days.

PR #8646, opened 2020-04-06
Fixes #28970 - Customize email notification for 'Subscriptions Expiring Soon'

jeremylenz:
~(WIP until #8638 is merged and I finish tests)~

#### Add 'Days from Now' dropdown to the email preferences page
email_notification_1
#### Changes the 'Subscriptions Expiring Soon' email notification to send a list of subscriptions expiring within XX days (no longer hard-coded to 30)
info

### DIFF
--- a/app/mailers/katello/subscription_mailer.rb
+++ b/app/mailers/katello/subscription_mailer.rb
@@ -11,7 +11,7 @@ module Katello
       end
 
       set_locale_for(user) do
-        mail(:to => user.mail, :subject => _("You have subscriptions expiring within #{days_from_now} days"))
+        mail(:to => user.mail, :subject => _("You have subscriptions expiring within %s days") % days_from_now)
       end
     end
 
--- a/app/mailers/katello/subscription_mailer.rb
+++ b/app/mailers/katello/subscription_mailer.rb
@@ -4,12 +4,14 @@ module Katello
 
     def subscription_expiry(options)
       user = ::User.find(options[:user])
+      days_from_now = options[:query]
+
       ::User.as(user.login) do
-        @pools = Katello::Pool.readable.expiring_in_days(65)
+        @pools = Katello::Pool.readable.expiring_in_days(days_from_now)
       end
 
       set_locale_for(user) do
-        mail(:to => user.mail, :subject => _("Subscriptions expiring soon"))
+        mail(:to => user.mail, :subject => _("You have subscriptions expiring within #{days_from_now} days"))
       end
     end
 
--- a/app/mailers/katello/subscription_mailer.rb
+++ b/app/mailers/katello/subscription_mailer.rb
@@ -5,7 +5,7 @@ module Katello
     def subscription_expiry(options)
       user = ::User.find(options[:user])
       ::User.as(user.login) do
-        @pools = Katello::Pool.readable.expiring_in_days(30)
+        @pools = Katello::Pool.readable.expiring_in_days(65)
       end
 
       set_locale_for(user) do
--- a/test/mailers/subscription_mailer_test.rb
+++ b/test/mailers/subscription_mailer_test.rb
@@ -13,20 +13,22 @@ module Katello
                         :subscription_type => 'report')
 
       @user.mail_notifications << MailNotification[:subscriptions_expiring_soon]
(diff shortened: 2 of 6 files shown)


app/mailers/katello/subscription_mailer.rb
```diff
@@ -0,0 +1,29 @@
+module Katello
+  class SubscriptionMailer < ApplicationMailer
+    after_action :prevent_sending_blank_report
+
+    def subscription_expiry(options)
+      user = ::User.find(options[:user])
+
+      mail_notification = MailNotification.find_by(name: :subscriptions_expiring_soon)
```
jturel, line 8: The audit mailer doesn't do a lookup like this and it's using the mail_query field from what I can tell:

https://github.com/theforeman/foreman/blob/develop/app/mailers/audit_mailer.rb

jeremylenz: @jturel Updated, thanks!

jeremylenz:
Rebased to current master and test updated

test/mailers/subscription_mailer_test.rb
```diff
@@ -13,20 +13,22 @@ def setup
...
       @pool_not_expiring_soon = FactoryBot.create(:katello_pool,
                                                   :not_expiring_soon,
+                                                  :with_organization,
                                                   cp_id: "1234",
                                                   subscription_id: ActiveRecord::FixtureSet.identify(:other_subscription))
       ActionMailer::Base.deliveries = []
     end
 
-    def setup_expiring_pool
+    def set_up_expiring_pool
```
jturel, line 26: Instead of putting this into a method, we could use a `let` ie

```
let(:expiring_pool) do
 FactoryBot.create(....)
end
```

Lets are lazily evaluated, so the pool won't be created until `expiring_pool` is called as if it were a local variable. And calling it multiple times will return the created pool, not invoke the block multiple times. Seems like the right fit for these test scenarios.

jeremylenz: @jturel Thanks, updated

test/mailers/subscription_mailer_test.rb
```diff
@@ -81,5 +94,17 @@ def test_omits_non_expiring_subscription
       assert_equal rows.length, 2
       assert Katello::Pool.readable.size > 1
     end
+
+    def test_days_from_now_mail_query
+      pool_expiring_soon.save
```
jturel, line 99: I think if you do `FactoryBot.create` vs `FactoryBot.build` you don't need the `save` calls~

app/mailers/katello/subscription_mailer.rb
```diff
@@ -4,12 +4,14 @@ class SubscriptionMailer < ApplicationMailer
...
+      days_from_now = options[:query]
+
       ::User.as(user.login) do
-        @pools = Katello::Pool.readable.expiring_in_days(30)
+        @pools = Katello::Pool.readable.expiring_in_days(days_from_now)
       end
 
       set_locale_for(user) do
-        mail(:to => user.mail, :subject => _("Subscriptions expiring soon"))
+        mail(:to => user.mail, :subject => _("You have subscriptions expiring within #{days_from_now} days"))
```
jturel, line 14: For translations we shouldn't use interpolation, but this format instead (that's the reason for the test failure):

```
_("You have subscriptions expiring within %s days") % days_from_now
```

test/mailers/subscription_mailer_test.rb
jeremylenz: @jturel For this test I wanted to ensure that both records were present when `deliver` is run.  Because the `let`s are lazily evaluated, they might not get created until the asserts, which would mean that this test isn't really testing anything.  (I could probably change it for the other tests though)

app/mailers/katello/subscription_mailer.rb
jeremylenz: thanks!  I was about to ask about that :smile:

test/mailers/subscription_mailer_test.rb
jturel: ah ok. if you use create vs build then you can call just `pool_expiting_soon` here

jeremylenz: @jturel Thanks, updated!

PR #8649, opened 2020-04-07
[DO NOT MERGE] Fixes #29499 - Fix Linting on TypeAhead component

johnpmitsch:


johnpmitsch:
Figured out that I needed to add the existing rules from our eslint configurations (I think it comes from the airbnb style guide) if modifying a rule. So needed:

```javascript
    "no-unused-vars": ["error", {
        "vars": "all", // existing option
        "args": "after-used", // existing option
        "ignoreRestSiblings": true, // existing option
        "argsIgnorePattern": "^_" // new option
    }],
```
 instead of
```javascript
    // existing options are overwritten
    "no-unused-vars": ["error", {
        "argsIgnorePattern": "^_" // new option
    }],
```

You can see the eslint config with ` ./node_modules/.bin/eslint --print-config webpack/.eslintrc.js`, which is how I figured this out

I'm going to close this out and fix this in the original PR

PR #8650, opened 2020-04-08
Fixes #29442 - Enable cert-auth

sjha4:
To Do:
- [x] Figure out behavior on proxy - https://github.com/Katello/katello/pull/8731
- [x] Test content consumption on client
- [x] Tests


app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -129,6 +129,19 @@ def pulp3_configuration(config_class)
...
+          uri = pulp3_uri!
+          config.host = uri.host
+          config.scheme = uri.scheme
+          config.ssl_client_cert = ::Cert::Certs.ssl_client_cert
+          config.ssl_client_key = ::Cert::Certs.ssl_client_key
+          config.debugging = true
+          config.logger = ::Foreman::Logging.logger('katello/pulp_rest')
+        end
+        cert_config
+      end
```
jlsherrill, line 143: what is this used for exactly? (how is it different than the pulp3_configuration method?

app/services/katello/pulp3/api/content_guard.rb
```diff
@@ -0,0 +1,35 @@
...
+          PulpCertguardClient::ApiClient.new(smart_proxy.pulp3_cert_configuration(PulpCertguardClient::Configuration))
+        end
+
+        def rhsm_api_client
+          PulpCertguardClient::ContentguardsRHSMApi.new(api_client)
+        end
+
+        def create(name = "RHSMCertGuard")
+          data = PulpCertguardClient::CertguardRHSMCertGuard.new(name: name, ca_certificate: Cert::Certs.ca_cert)
+          rhsm_api_client.create(data)
```
jlsherrill, line 21: we probably want to handle a create conflict here, in case two get created with the same name in pulp, the conflict will be 'ignored' and the initial will be recorded/used

app/services/katello/pulp3/repository/file.rb
```diff
@@ -8,12 +8,14 @@ def copy_content_for_source(source_repository, _options = {})
...
-        def distribution_options(path)
-          {
+        def distribution_options(path, unprotected = true)
+          options = {
             base_path: path,
             publication: repo.publication_href,
             name: "#{generate_backend_object_name}"
           }
+          options[:content_guard] = Katello::Pulp3::ContentGuard.first.pulp_href unless unprotected
+          options
```
jlsherrill, line 18: would it make sense to have a 'common_distribution_options' method to handle the cert guard piece?

app/models/katello/concerns/smart_proxy_extensions.rb
sjha4: I was seeing some issues with configuring the cert_guard. This was in trying to work around that..Will test as this might not be needed any more..

app/services/katello/pulp3/api/content_guard.rb
jlsherrill: i think this is still unresolved?

sjha4: Verfied that name is indeed unique for contentguard. So catching exception should work.

app/models/katello/pulp3/content_guard.rb
```diff
@@ -0,0 +1,14 @@
+module Katello
+  module Pulp3
+    class ContentGuard < Katello::Model
```
jturel, line 3: Just trying to learn - is there a reason this class extends Katello::Model?

app/lib/actions/pulp3/ContentGuard/refresh.rb
```diff
@@ -0,0 +1,19 @@
+module Actions
+  module Pulp3
+    module ContentGuard
+      class Refresh < Pulp3::Abstract
+        def plan(smart_proxy, options = {})
+          return if (::Katello::Pulp3::ContentGuard.count > 0 || options.try(:[], :update))
```
jturel, line 6: How does this class seem to respond to enumerable method like count and first?

jturel: I just realized it's an active record model :)

app/models/katello/concerns/smart_proxy_extensions.rb
sjha4: This isn't needed..Tested with the default pulp_config method.

sjha4:
Need to re-record cassettes with unprotected set to true for existing tests. Added a test with unprotected=false which is content_guard specific. Will add some more tests and update VCRs..

sjha4:
[test katello]

app/services/katello/pulp3/api/content_guard.rb
```diff
@@ -0,0 +1,45 @@
...
+
+        def api_client
+          PulpCertguardClient::ApiClient.new(smart_proxy.pulp3_configuration(PulpCertguardClient::Configuration))
+        end
+
+        def rhsm_api_client
+          PulpCertguardClient::ContentguardsRhsmApi.new(api_client)
+        end
+
+        def create(name = "RHSMCertGuard")
```
jturel, line 23: Perhaps ca_certificate should be an argument here - with the default being `Cert::Certs.ca_cert`

sjha4: Good point..will update :+1:

jlsherrill:
One enhancement i think we should do, and this wasn't done as part of pulp2 so maybe it should be another issue, but it may be pretty simple:

* If i change the 'unprotected' attribute of a repo in library, update all distributions of that repo across all CV's to match the root repo's setting

app/lib/actions/pulp3/repository/update_cv_repository_cert_guard.rb
```diff
@@ -0,0 +1,19 @@
...
+      class UpdateCVRepositoryCertGuard < Pulp3::Abstract
+        def plan(repository, _smart_proxy)
+          plan_self(:root_id => repository.root.id)
+        end
+
+        def run
+          root = ::Katello::RootRepository.find(input[:root_id])
+          cv_repositories = root.repositories - [root.library_instance]
+          cv_repositories.each do |repo|
+            ForemanTasks.async_task(::Actions::Pulp3::Repository::RefreshDistribution, repo, SmartProxy.pulp_master)
```
jlsherrill, line 13: since '::Actions::Katello::Repository::UpdateCVRepoCertGuard' is being run async anyways, we can just call refresh distribution as part of a 'plan_task' within this action's plan phase.  No need to spin them all off as async_tasks

app/lib/actions/katello/repository/update.rb
```diff
@@ -50,9 +50,15 @@ def plan(root, repo_params)
 
         def run
           repository = ::Katello::Repository.find(input[:repository_id])
+          output[:repository_id] = input[:repository_id]
           ForemanTasks.async_task(Katello::Repository::MetadataGenerate, repository)
         end
 
+        def finalize
+          repository = ::Katello::Repository.find(output[:repository_id])
+          ForemanTasks.async_task(::Actions::Katello::Repository::UpdateCVRepoCertGuard, repository, SmartProxy.pulp_master)
```
jlsherrill, line 59: is it possible to only do this if we detect changes to the 'unprotected' attribute?

app/lib/actions/katello/repository/update_cv_repo_cert_guard.rb
```diff
@@ -0,0 +1,13 @@
+module Actions
+  module Katello
+    module Repository
+      class UpdateCVRepoCertGuard < Actions::Base
+        include Actions::Katello::PulpSelector
+
+        def plan(repository, smart_proxy)
+          plan_optional_pulp_action([::Actions::Pulp3::Repository::UpdateCVRepositoryCertGuard], repository, smart_proxy)
+        end
+      end
```
jlsherrill, line 10: Can you add a 'humanized_name' method to show a more friendly name in the tasks list?  Maybe something that doesn't reference cert guard, like 'Updating repository authentication configuration' or something...

jlsherrill:
This looks awesome!  and works flawlessly.  Just a couple comments and this is good to go

katello.gemspec
```diff
@@ -52,6 +52,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency "pulp_container_client", ">= 1.3.0", "< 1.4.0"
   gem.add_dependency "pulp_rpm_client", ">=3.3.0", "< 3.4.0"
   gem.add_dependency "pulp_2to3_migration_client", "> 0.1", "< 0.3.0"
+  gem.add_dependency "pulp_certguard_client", "0.1.0rc5.dev01588849857"
```
jlsherrill, line 55: can you relax this a bit   '>= 0.1.0rc5',  '< 2.0' 

I think that will cover the upcoming 1.0 ga release of the plugin

sjha4: Oh..missed this comment..pushing now with the change..

sjha4:
[test katello]

sjha4:
[test katello]

sjha4:
[test katello]

PR #8651, opened 2020-04-08
Fixes #29502 - use Rails 6 compatible angular-rails-templates version

mmoll:


mmoll:
RPM PR: https://github.com/theforeman/foreman-packaging/pull/5012

PR #8658, opened 2020-04-14
Fixes #29454 - container tags don't show all CVVs

ianballou:
To test:

**For Pulp 2 and Pulp 3:**

1) Sync some Docker repos
2) Add some of those repos to content views
3) Publish the content views, add some to an environment or two
4) Go to Content -> Container Image Tags
5) Click on tags and make sure that the Lifecycle Environments table makes sense
    - All environments and content view versions that the tags exist in (non-archived) should be listed there

Questions/Issues/Additions:

- Sometimes if you click on a link in the Lifecycle Environment table and then click the back button on the browser, the table fails to load with the JS error `"TypeError: "self.table.allSelected is not a function"`  It looks like this happens on master, curious how to fix this.
- I discovered that, for Pulp 2, not all Docker tags are being shown on the Container Image Tags page.  I've also fixed that in this PR.
  - The cause was `search_in_tags(DockerTag.in_repositories(repos), grouped)` in `docker_meta_tag.rb` messing up when Pulp 2 Docker repositories are in content views.  I've adjusted the query to make it work.  Please test adding docker repos to content views and environments and checking that the global tags page makes sense.


app/controllers/katello/api/v2/repositories_controller.rb
```diff
@@ -119,7 +121,11 @@ def index
     end
 
     def index_relation
-      query = Repository.readable
+      if ::Foreman::Cast.to_bool params[:archived]
+        query = params[:docker_tag_id].nil? ? Repository.readable : DockerMetaTag.find(params[:docker_tag_id]).repositories.readable
+      else
+        query = params[:docker_tag_id].nil? ? Repository.non_archived.readable : DockerMetaTag.find(params[:docker_tag_id]).repositories.non_archived.readable
```
ianballou, line 127: I realized I need to re-work the default behavior here since it's different from normal.

app/controllers/katello/api/v2/repositories_controller.rb
```diff
@@ -120,6 +122,13 @@ def index
 
     def index_relation
       query = Repository.readable
+      unless params[:docker_tag_id].nil?
```
jturel, line 125: Something feels a bit off to me about the changes to this controller and the angular controller. I think it's because they are both treating DocketMetaTags as a special case. I want to throw out an idea for you, and you tell me if will help make things cleaner, worse, or whatever: I see we have a docker_tags_controller. Can we add a 'def repositories' method (ie /api/v2/docker_tags/:id/repositories) that knows what do do for docker tags specifically? Then we can use that from the UI controller

engines/bastion_katello/app/assets/javascripts/bastion_katello/docker-tags/details/docker-tag-environments.controller.js
```diff
@@ -11,27 +11,36 @@
...
                 'search': $location.search().search || "",
                 'sort_by': 'name',
                 'sort_order': 'ASC',
-                'paged': false,
-                'ids[]': ids
+                'paged': false
             };
+            var ids;
+            var nutupane;
+            if ($scope.tag.repositories.length > 1) {
```
jturel, line 34: I don't understand how we might be looking at a repository *or* a docker tag in this controller. Are we using this controller for multiple views? It feels like we need a separate controller and view for this rather than handle special cases. Let me know if I'm missing the point of your solution here.

app/controllers/katello/api/v2/repositories_controller.rb
ianballou: Sure, I can't think of any immediate reason why that wouldn't work here. It is a cleaner and more reusable solution.

engines/bastion_katello/app/assets/javascripts/bastion_katello/docker-tags/details/docker-tag-environments.controller.js
ianballou: This is for a single view, which is the environments tab for each individual tag.  So the reason why we need to have two scenarios (tag and repo) is that the environments/CVVs page for a docker tag cannot be created using the same strategy for both Pulp 2 and Pulp 3.  For Pulp 2 each repo gets its own copy of a tag so the `related_tags` method works as intended.  For Pulp 3, however, tags are shared between repos so `related_tags` only ends up returning itself.  My solution was to have this page be "tag-centric" for Pulp 2 and "repo-centric" for Pulp 3 in the background, but have the page look identical to the user.

I see what you mean about splitting this into different controllers, do you think it would work out cleanly to use one bastion_katello controller over another depending on if we're using Pulp 3 or not?  I'm not sure myself at the moment, I'd have to see how bastion_katello chooses which controller to use.  Also since you mentioned maintainability, the Pulp 2 (tag-centric) method of displaying this page will be dropped eventually along with Pulp 2.

jturel: Now I understand! Thanks to your explanation I think we can keep this front-end changes as you've made them because half of it will be tech-debt which we can clean up when pulp2 is gone

app/models/katello/docker_meta_tag.rb
```diff
@@ -107,7 +107,7 @@ def self.search_in_tags(tags, grouped = false)
       if grouped
         grouped_fields = "#{table_name}.name, katello_repositories.root_id"
         ids = distinct.select("ON (#{grouped_fields}) #{table_name}.id").joins(:repositories)
-        where(:id => ids).where("#{self.table_name}.schema1_id in (#{sql}) or #{self.table_name}.schema2_id in (#{sql})")
+        self.where(:id => ids)
```
jturel, line 110: Might be worthwhile to add a unit test here

ianballou:
@jturel I think I've addressed your comments.

ianballou:
I'm a little confused why I'm seeing the error `katello/api/v2/docker_tags/repositories not found`.  When I look in the permission actions, I figured it should be checking for something more like `katello/api/v2/docker_tags/repositories/index`.  I tried adding that to the permissions list and it didn't seem to work

jturel:
Will give this another look today. Thank you for the updates - looks good at a glance.

config/routes/api/v2.rb
```diff
@@ -163,6 +163,9 @@ class ActionDispatch::Routing::Mapper
             get :auto_complete_name
             get :compare
           end
+          member do
+            get 'repositories', :action => :repositories
```
jturel, line 167: I think this can just be `get :repositories`

engines/bastion_katello/app/assets/javascripts/bastion_katello/docker-tags/docker-tag-repositories.factory.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * @ngdoc service
+ * @name  Bastion.docker-tags.factory:DockerTagRepositories
+ *
+ * @requires BastionResource
+ *
+ * @description
+ *   Provides a BastionResource for Docker Tag Repositories
+ */
+angular.module('Bastion.docker-tags').factory('DockerTagRepositories',
```
jturel, line 10: I think we don't need a whole new file for this  but let me know if you already tried the below:

I believe the DockerTag factory can be expanded to add the repositories endpoint and that would be invoked with `DockerTag.getRepositories` or something like that. You can refer to another file that makes me think it's possible:  https://github.com/Katello/katello/blob/master/engines/bastion_katello/app/assets/javascripts/bastion_katello/hosts/host-bulk-action.factory.js

BTW - does this autocomplete stuff in here actually work?

app/controllers/katello/api/v2/docker_tags_controller.rb
```diff
@@ -25,6 +25,18 @@ def index
       end
     end
 
+    api :GET, "/docker_tags/:id/repositories", N_("List of repositories for a docker meta tag")
+    param :archived, :bool, :desc => N_("include archived repositories")
+    def repositories
```
jturel, line 30: Works well. I think this was the way to go - thanks for adding it

engines/bastion_katello/app/assets/javascripts/bastion_katello/docker-tags/docker-tag-repositories.factory.js
ianballou: Looks like autocomplete is broken here now.  I will investigate if autocomplete can be fixed without too much trouble since we're filtering on repositories in the docker tags controller.  If not, I'll revert to editing the repositories controller like I was doing before.  It isn't worth getting too caught up on this.

ianballou: Edit: looks like autocomplete was never there to begin with, but the filter was indeed broken.  The new repo-centric table was missing it, I'm going to add it in my next commit.

ianballou:
@jturel I think I've addressed all of your comments. I managed to get rid of the `docker-tag-repositories` factory too, it was a smaller change than I expected.

engines/bastion_katello/app/assets/javascripts/bastion_katello/docker-tags/docker-tags.controller.js
```diff
@@ -44,5 +44,17 @@ angular.module('Bastion.docker-tags').controller('DockerTagsController',
             }
             return versions.join(", ");
         };
+
+        $scope.getRepositoryNames = function (tag) {
+            var names = [];
+            var i;
+            for (i = 0; i < tag.repositories.length; ++i) {
```
jturel, line 51: Just throwing this out there as the stylistic choice is up to you - but there's a mix of 2 styles in this method.

I think it could be rewritten like:
```
return tag.repositories.map(function(repo) {
  return repo.name;
}).filter(function(item, index) {
  return names.indexOf(item) >= index;
}).sort().join(', ');
```

app/views/katello/api/v2/repositories/base.json.rabl
```diff
@@ -57,3 +57,11 @@ end
 child :content_view => :content_view do |_repo|
   attribute :id, :name
 end
+
+child :content_view_version do
```
jturel, line 61: For these added attributes you may want to take a look at the dev log when loading repositories API (both old and new) to make sure there are no n+1 queries. The 'bullet' gem logs that data for us. Would be good to pre-load these objects in the controllers in that case otherwise this could make things slow

app/controllers/katello/api/v2/docker_tags_controller.rb
```diff
@@ -25,6 +25,18 @@ def index
...
+      if ::Foreman::Cast.to_bool params[:archived]
```
jturel, line 31: stylistic choice, up to you here, but we tend to write these things like this:

```
repos = DockerMetaTag.find(params[:id]).repositories
repos = repos.non_archived unless ::Foreman::Cast.to_bool(params[:archived])
```

That said - it looks like you're only using the non-archived version of this API call in the code. maybe we don't need to support querying both ways?

ianballou: I can remove the archived parameter until someone needs it

config/routes/api/v2.rb
```diff
@@ -163,6 +163,9 @@ class ActionDispatch::Routing::Mapper
             get :auto_complete_name
             get :compare
           end
+          member do
+            get :repositories, :action => :repositories
```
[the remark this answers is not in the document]
ianballou: oh I misunderstood your last comment there, I'll try it out.

engines/bastion_katello/app/assets/javascripts/bastion_katello/docker-tags/docker-tags.controller.js
```diff
@@ -44,5 +44,17 @@ angular.module('Bastion.docker-tags').controller('DockerTagsController',
...
+                names.push(tag.repositories[i].name);
+            }
+
+            return names.filter(function(item, index) {
+                return names.indexOf(item) >= index;
```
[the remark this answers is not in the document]
ianballou: Yeah, it was getting duplicates before.

engines/bastion_katello/app/assets/javascripts/bastion_katello/docker-tags/docker-tags.controller.js
ianballou: this works fine for me!

app/views/katello/api/v2/repositories/base.json.rabl
ianballou: Is it logged by default? I'm not seeing anything about N+1 in my logs.

jturel:
This all looks good to me for pulp2 and pulp3. Last thing is the JS test failure. I believe the ruby test failures are on master and related to applicability?

ianballou:
> This all looks good to me for pulp2 and pulp3. Last thing is the JS test failure. I believe the ruby test failures are on master and related to applicability?

I'm figuring out the JS test failure now.  As for the Ruby test failures, I've got a PR out for them now.

ianballou:
[test katello]

PR #8659, opened 2020-04-14
Fixes #29527 - Load distributor version from a constant

jturel:
This updates the distributor version to 'sat-6.7' and loads it from a constant. This will enable us to override the version with our downstream theme gem, so that we won't have to worry about forgetting to update it in the upstream where it's less crucial. That'll work because we can define the override based on the downstream version.

To test:

- create a manifest and select a version prior to '6.7'.
- import that manifest and refresh it.
- portal should reflect 6.7 as the version instead of whatever was set intially

### DIFF
--- a/app/models/katello/glue/provider.rb
+++ b/app/models/katello/glue/provider.rb
@@ -1,5 +1,7 @@
 module Katello
   module Glue::Provider
+    DISTRIBUTOR_VERSION = 'sat-6.7'.freeze
+
     def self.included(base)
       base.send :include, InstanceMethods
     end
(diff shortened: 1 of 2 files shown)


jlsherrill:
One thing that might be interesting is if we want to report 'upstream' or something, such that we could differentiate between a satellite and katello.  Obviously that would need coordination, but might make more sense than having a hard coded sat version, and may provide some insight.

jturel:
> One thing that might be interesting is if we want to report 'upstream' or something, such that we could differentiate between a satellite and katello. Obviously that would need coordination, but might make more sense than having a hard coded sat version, and may provide some insight.

It's a good idea. I'll keep it in mind as we discuss including the .z digit in the future. Would be nice for our team to have the right access to add those versions rather than involving IT too.

PR #8661, opened 2020-04-15
[CP] Katello 3.15 GA cherrypicks

jjeffers:


jturel:
Looks good! Few things

- https://projects.theforeman.org/issues/28476 should be retriaged to 3.16
- needs version bump
- the rubocop failures appear to be related to the fact that we have a newer version on master. I think we'll need a commit here to make rubocop happy on this branch. I think the fix is to change TargetRubyVersion to 2.5 in .rubocop.yml

jturel:
Well, setting the target ruby version just made things worse. In this case I think the simple fix is going to be to change the task in pulp3_content_switchover.rake to have a 'begin' block rather than the newer style that omits it.

jjeffers:
[test katello]

PR #8662, opened 2020-04-16
Fixes #29488 - Make modular errata applicable

parthaa:
Prior to this commit the applicable errata was not properly
calculated.

Content hosts upload a list of packages, enabled repos and modulemd
information (aka profiles) on every dnf transaction.
* When katello receives this information it passes on pulp to calculate
  Packages, Errata and Module Streams that are applicable.
* Applicable errata is calculated against the "library_instance" of the
  repositories bound to the content host.

However module streams introduce some quirkiness. The prior upload
module stream method would just pass the actual module stream that is
available to the system.

This commit passes the module stream available in the library instance
of the bound repos instead to pulp.

So when new errata get added overtime the new errata would automatically
be marked "applicable".

### DIFF
--- a/app/lib/actions/katello/host/upload_profiles.rb
+++ b/app/lib/actions/katello/host/upload_profiles.rb
@@ -33,16 +33,42 @@ module Actions
           Dynflow::Action::Rescue::Skip
         end
 
+        def self.upload_modules_to_pulp(available_streams, host)
+          query_name_streams = available_streams.map do |profile|
+            ::Katello::ModuleStream.where(profile.slice(:name, :stream))
+          end
+
+          query_name_streams = query_name_streams.inject(&:or)
+
+          bound_library_instances = host.content_facet.bound_repositories.map(&:library_instance_or_self)
+          query = ::Katello::ModuleStream.in_repositories(bound_library_instances).
+                                          select(:name, :stream, :version, :context, :arch).
+                                            merge(query_name_streams)
+
+          updated_profiles = query.map do |module_stream|
+            module_stream.slice(:name, :stream, :version, :context, :arch)
+          end
+
+          # We also need to pass module streams that are not found in the ModuleStream table
+          # but are present on the content host
+          unassociated_profiles = available_streams.select do |profile|
+            updated_profiles.none? { |p| p[:name] == profile[:name] && p[:stream] == profile[:stream] }
+          end
+
+          ::Katello::Pulp::Consumer.new(host.content_facet.uuid).
+              upload_module_stream_profile(updated_profiles + unassociated_profiles)
+        rescue RestClient::ResourceNotFound
+          Rails.logger.warn("Host with ID %s was not known to Pulp, continuing" % host.id)
+        end
+
         def import_module_streams(payload, host)
           enabled_payload = payload.map do |profile|
-            profile.slice("name", "stream", "version", "context", "arch") if profile["status"] == "enabled"
+            profile.slice("name", "stream", "version", "context", "arch").with_indifferent_access if profile["status"] == "enabled"
           end
           enabled_payload.compact!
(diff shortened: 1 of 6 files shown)


parthaa:
Steps to test this PR
Make sure you have the latest pulp2 `pulp-server-2.21.1-1.el7.noarch` rpm installed in your dev env. 

On master branch 
* Sync RHEL8 appstream
* Create a content view that has RHEL 8 appstream
* Create an errata excludes filter to exclude RHSA-2020:0579
* Publish the content view
* Have a rhel8 host consume off it

On the host run the following
```
$ dnf -y module install nodejs:10
$ dnf uploadprofile     
```

Now back on the UI

* Content -> Errata
* Search for 'errata_id = RHSA-2020:0579'

You should see that `RHSA-2020:0579` has 0 Applicable Errata and 0 Installable

* Now change over this PR branch
* On the rhel8 host run 
```
$ dnf uploadprofile    
```
Get back to the UI

* Content -> Errata
* Search for 'errata_id = RHSA-2020:0579'
You should see something like

 image text`` 

- showing one applicable host for `RHSA-2020:0579`

app/lib/actions/katello/host/upload_profiles.rb
```diff
@@ -35,11 +35,32 @@ def rescue_strategy
...
+            profile.slice("name", "stream", "version", "context", "arch").with_indifferent_access if profile["status"] == "enabled"
           end
           enabled_payload.compact!
+          # Get the library instances
+          bound_library_instances = host.content_facet.bound_repositories.map(&:library_instance_or_self)
 
-          ::Katello::Pulp::Consumer.new(host.content_facet.uuid).upload_module_stream_profile(enabled_payload)
+          query_name_streams = enabled_payload.map do |profile|
+            ::Katello::ModuleStream.where(profile.slice(:name, :stream))
+          end
```
jlsherrill, line 46: can we put the query_name_streams.inject(&:org) here?  i think it could easily be missed down below.  I originally thought you had forgot to put it in

parthaa: updated

parthaa:
[test katello]

jlsherrill:
This works well +1 to the code and functionality.  But one question, should we add in a migration rake task to fix the issue for upgrades (since the pulp consumers will be wrong).  Do you know under what conditions they will be updated?  Would it require all rhel8 systems to have a package update occur?

parthaa:
[test katello]

parthaa:
> This works well +1 to the code and functionality. But one question, should we add in a migration rake task to fix the issue for upgrades (since the pulp consumers will be wrong). Do you know under what conditions they will be updated? Would it require all rhel8 systems to have a package update occur?

Good point. I would say yes we'd likely need to look at all rhel8 systems. Would need to think a bit more on this. Debating if filing a new issue for migration is a better idea ?

parthaa:
@jlsherrill finally added migration :)

parthaa:
[test katello]

lib/katello/tasks/upgrades/3.16/update_applicable_el8_hosts.rake
```diff
@@ -0,0 +1,29 @@
...
+                                          non_archived)
+        hosts.each do |host|
+          available_streams = ::Katello::HostAvailableModuleStream.joins(:available_module_stream).
+                                                enabled.where(:host_id => host).select(:name, :stream)
+          ::Actions::Katello::Host::UploadProfiles.upload_modules_to_pulp(available_streams, host)
+        end
+      end
+    end
+  end
+end
```
jlsherrill, line 29: if we want this to run automatically, we have to add this to the seeds file for upgrade tasks

PR #8667, opened 2020-04-17
Fixes #27501 - Filter private keys in Katello logs

jeremylenz:
This change filters out RSA private keys before they're logged.  In the log file, the private key is replaced with `[private key filtered]`.

To test:

```
tail -f log/development.log | grep "BEGIN RSA PRIVATE KEY"
```
* Navigate to Katello pages (subscriptions, Red Hat Repositories, etc.)
* Verify that no private keys are logged
* Compare results with `master`

### DIFF
--- a/app/controllers/katello/api/registry/registry_proxies_controller.rb
+++ b/app/controllers/katello/api/registry/registry_proxies_controller.rb
@@ -488,7 +488,7 @@ module Katello
 
     def process_action(method_name, *args)
       ::Api::V2::BaseController.instance_method(:process_action).bind(self).call(method_name, *args)
-      Rails.logger.debug "With body: #{response.body}\n" unless route_name == 'pull_blob'
+      Rails.logger.debug "With body: #{filter_sensitive_data(response.body)}\n" unless route_name == 'pull_blob'
     end
 
     def item_not_found(item)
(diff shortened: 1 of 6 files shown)


jeremylenz:
Updated.  To retest, on a katello-client do
```
subscription-manager refresh
```
in addition to the testing above.

chris1984:
Restesting

jturel:
Untested, but the concept looks good. I'm unconcerned. :eyeglasses:

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -12,7 +12,9 @@ class Api::Registry::RegistryProxiesController < Api::V2::ApiController
     before_action :authorize_repository_write, only: [:push_manifest]
     skip_before_action :check_content_type, only: [:start_upload_blob, :upload_blob, :finish_upload_blob,
                                                    :chunk_upload_blob, :push_manifest]
-    skip_after_action :log_response_body, :only => [:pull_blob]
+    if defined?(log_response_body)
```
jturel, line 15: why check if its defined?

jeremylenz: When you run `skip_after_action` Rails doesn't define the method.. it was causing the web app not to load because `skip_after_action` had already been run on the Api::BaseController from which this inherits.

jeremylenz: After confirming this fixes the issue I'll probably just not run `skip_after_action` here, instead of checking

jeremylenz: Still not sure why @chris1984 and Jenkins are seeing this issue, but I didn't see the error locally

chris1984:
Thanks @jeremylenz

PR #8669, opened 2020-04-18
Fixes #29565 - Remove outdated Rails workarounds

mmoll:


mmoll:
Finally :green_heart:.
I hope the tests are meaningful enough for the real app not to break...

PR #8670, opened 2020-04-20
Fixes #29228 - Ignore 404 on delete

sjha4:
To test:
1. Create a repo and sync
2. curl --cert /etc/pki/katello/certs/pulp-client.crt --key /etc/pki/katello/private/pulp-client.key https://`hostname`/pulp/api/v3/repositories/file/file/  to get the **repo_href.**
3. curl -X DELETE --cert /etc/pki/katello/certs/pulp-client.crt --key /etc/pki/katello/private/pulp-client.key https://`hostname`/pulp/api/v3/repositories/file/file/**repo_href**/
4. Delete the repo in katello UI. It should succeed.

### DIFF
--- a/app/services/katello/pulp3/api/core.rb
+++ b/app/services/katello/pulp3/api/core.rb
@@ -78,15 +78,19 @@ module Katello
           PulpcoreClient::Upload
         end
 
+        def ignore_404_exception(*)
+          yield
+        rescue self.class.api_exception_class => e
+          raise e unless e.code == 404
+          nil
+        end
+
         def delete_orphans
           [orphans_api.delete]
         end
 
         def delete_remote(remote_href)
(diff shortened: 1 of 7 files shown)


app/services/katello/pulp3/api/core.rb
```diff
@@ -84,6 +84,9 @@ def delete_orphans
 
         def delete_remote(remote_href)
           remotes_api.delete(remote_href)
+        rescue self.class.api_exception_class => e
+          raise e if e.code != 404
+          nil
```
jlsherrill, line 89: can we add some helper like:
ignore_404 {  remotes_api.delete(remote_href) }

jlsherrill:
[test katello]

PR #8671, opened 2020-04-20
Fixes #29576 - Run linting from Travis CI config

johnpmitsch:
Travis has been failing with `eslint: command not found`

My best guess is running the linting command from a bash script was causing some sort of environmental difference and eslint couldn't be found on the $PATH

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: node_js
 node_js:
   - '10'
-script: ./travis_run_js_tests.sh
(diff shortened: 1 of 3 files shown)


johnpmitsch:
For record-keeping purposes the `npx` is what fixed the travis issue, so the global package must have been missing on some travis builds. The removal of the wrapper script didn't actually fix anything, but was kept in to keep things succinct

PR #8675, opened 2020-04-22
Fixes #28971 - Generating and adding entitlements expiring soon report link in email

sseelam2:
This PR generates the entitlements expiring soon report link and attaches it in the email notification.

### DIFF
--- a/app/helpers/katello/subscription_mailer_helper.rb
+++ b/app/helpers/katello/subscription_mailer_helper.rb
@@ -1,0 +1,23 @@
+module Katello
+  module SubscriptionMailerHelper
+    include Rails.application.routes.url_helpers
+    include ActionView::Helpers::UrlHelper
+
+    def report_url
+      base_url = report_data_report_template_path(@report_template.id, job_id: @provider_job_id)
+      "#{Setting[:foreman_url]}#{base_url}"
+    end
+
+    def report_link
+      link_to _("View a report of the affected hosts"), report_url
+    end
+
+    def start_report_task(days_from_now)
+      @report_template = ReportTemplate.find_by(name: "Subscription - Entitlement Report")
+      template_input_id = @report_template.template_inputs.find_by_name("Days from Now").id.to_s
+      params = { format: 'csv', template_id: @report_template.id, input_values: { template_input_id => { value: days_from_now} } }
+      composer = ReportComposer.new(params)
+      @provider_job_id = composer.schedule_rendering.provider_job_id
+    end
+  end
+end
--- a/app/mailers/katello/subscription_mailer.rb
+++ b/app/mailers/katello/subscription_mailer.rb
@@ -1,6 +1,8 @@
 module Katello
   class SubscriptionMailer < ApplicationMailer
+    helper :'katello/subscription_mailer'
     after_action :prevent_sending_blank_report
+    include SubscriptionMailerHelper
 
     def subscription_expiry(options)
       user = ::User.find(options[:user])
@@ -8,6 +10,13 @@ module Katello
 
       ::User.as(user.login) do
         @pools = Katello::Pool.readable.expiring_in_days(days_from_now)
+        @affected_hosts = ::Host::Managed.with_pools_expiring_in_days(days_from_now)
+      end
+
+      if @affected_hosts.any?
+        start_report_task(days_from_now)
+        @report_url = report_url
+        @report_link = report_link
       end
 
       set_locale_for(user) do
--- a/app/views/katello/subscription_mailer/subscription_expiry.html.erb
+++ b/app/views/katello/subscription_mailer/subscription_expiry.html.erb
@@ -41,5 +41,7 @@
       <% end %>
     </table>
 </div>
+<br/>
+  <%= @report_link %>
 </body>
 </html>
--- a/app/views/katello/subscription_mailer/subscription_expiry.text.erb
+++ b/app/views/katello/subscription_mailer/subscription_expiry.text.erb
@@ -13,3 +13,7 @@
   <%= _("Days Remaining") %>:       <%= pool.days_until_expiration %>
 
 <% end %>
+
+<%if @report_url %>
+  <%= _("View a report of the affected hosts") %>:               <%= @report_url %>
+<% end %>
--- a/db/seeds.d/106-mail_notifications.rb
+++ b/db/seeds.d/106-mail_notifications.rb
@@ -34,7 +34,7 @@ User.as(::User.anonymous_api_admin.login) do
     },
 
     {:name => :subscriptions_expiring_soon,
-     :description => N_('A list of subscriptions expiring within 30 days'),
+     :description => N_('A list of subscriptions expiring soon'),
      :mailer => 'Katello::SubscriptionMailer',
      :method => 'subscription_expiry',
      :subscription_type => 'report',
--- a/test/mailers/subscription_mailer_test.rb
+++ b/test/mailers/subscription_mailer_test.rb
@@ -44,6 +44,9 @@ module Katello
                         cp_id: "1234",
                         subscription_id: ActiveRecord::FixtureSet.identify(:other_subscription))
 
+      template = FactoryBot.create(:report_template, name: 'Subscription - Entitlement Report')
+      FactoryBot.create(:template_input, name: 'Days from Now', template_id: template.id)
+
       ActionMailer::Base.deliveries = []
     end
 


app/helpers/katello/subscription_mailer_helper.rb
```diff
@@ -0,0 +1,17 @@
+module Katello
+  module SubscriptionMailerHelper
+    include Rails.application.routes.url_helpers
+
+    def report_url
+      params = { format: 'html', template_id: ReportTemplate.find_by(name: "Subscription - Entitlement Report").id , input_values: { TemplateInput.find_by_name("Days from Now").id.to_s  => { value: UserMailNotification.first.mail_query.to_s } } }
```
jeremylenz, line 6: (Repeating our IRC convo) Remember to refactor this to not use `UserMailNotification.first`

jeremylenz, line 6: `ReportTemplate.find_by(name: "Subscription - Entitlement Report").id` is used twice in this method; I suggest DRYing it up by assigning it to a variable.

jeremylenz: For readability, it might be better to also assign `TemplateInput.find_by_name("Days from Now").id.to_s` and the `mail_query` value to variables as well.

app/helpers/katello/subscription_mailer_helper.rb
```diff
@@ -0,0 +1,17 @@
+module Katello
+  module SubscriptionMailerHelper
+    include Rails.application.routes.url_helpers
+
+    def report_url
```
jeremylenz, line 5: This method is doing two things - running the report and building the url.  I think it should be broken into two methods.  `report_url` can call the `start_report_task` method.

app/views/katello/subscription_mailer/subscription_expiry.text.erb
```diff
@@ -11,5 +11,6 @@
   <%= _("Start Date") %>:           <%= pool.start_date %>
   <%= _("End Date") %>:             <%= pool.end_date %>
   <%= _("Days Remaining") %>:       <%= pool.days_until_expiration %>
+  <%= _("Report") %>:               <%= report_url %>
```
jeremylenz, line 14: This shouldn't be inside the loop

app/views/katello/subscription_mailer/subscription_expiry.html.erb
```diff
@@ -40,6 +40,7 @@
         </tr>
       <% end %>
     </table>
+  <%= _("Report") + link_to(report_url) %>
```
jeremylenz, line 43: I think this should be
```ruby
<%= link_to _("Report"), report_url %>
```

app/helpers/katello/subscription_mailer_helper.rb
jturel, line 6: I think CSV might be a more suitable format for the report since it's both human and machine readable.

app/helpers/katello/subscription_mailer_helper.rb
```diff
@@ -0,0 +1,17 @@
...
+      params = { format: 'html', template_id: ReportTemplate.find_by(name: "Subscription - Entitlement Report").id , input_values: { TemplateInput.find_by_name("Days from Now").id.to_s  => { value: UserMailNotification.first.mail_query.to_s } } }
+
+      composer = ReportComposer.new(params)
+      job = composer.schedule_rendering
+
+      base_url = report_data_report_template_path(ReportTemplate.find_by(name: "Subscription - Entitlement Report").id)
```
jturel, line 11: It's been a while but I think you can specify job_id as a hash parameter to `report_data_report_template_path` to avoid a bit of the string interpolation at the next line

app/helpers/katello/subscription_mailer_helper.rb
```diff
@@ -0,0 +1,25 @@
...
+    end
+
+    def start_report_task(params)
+      composer = ReportComposer.new(params)
+      job = composer.schedule_rendering
+      job.provider_job_id
+    end
+  end
+end
+
```
jturel, line 24: nit: two unnecessary lines at end of file

app/helpers/katello/subscription_mailer_helper.rb
```diff
@@ -0,0 +1,25 @@
...
+      mail_query = UserMailNotification.first.mail_query.to_s
+
+      params = { format: 'csv', template_id: report_id , input_values: { days_from_now => { value: mail_query} } }
+
+      provider_job_id = start_report_task(params)
+      base_url = report_data_report_template_path(report_id, job_id: provider_job_id)
+      "#{Setting[:foreman_url]}#{base_url}"
+    end
+
+    def start_report_task(params)
```
jturel, line 17: I think this method should just return the result of `composer.schedule_rendering` that way we can gather more data if needed beyond the provider_job_id without having to change the method later on

app/helpers/katello/subscription_mailer_helper.rb
```diff
@@ -0,0 +1,25 @@
...
+    include Rails.application.routes.url_helpers
+
+    def report_url
+      report_id = ReportTemplate.find_by(name: "Subscription - Entitlement Report").id
+      days_from_now = TemplateInput.find_by_name("Days from Now").id.to_s
+      mail_query = UserMailNotification.first.mail_query.to_s
+
+      params = { format: 'csv', template_id: report_id , input_values: { days_from_now => { value: mail_query} } }
+
+      provider_job_id = start_report_task(params)
```
jturel, line 12: It seems more correct to have `SubscriptionMailer` call `start_report_task` but let me know if that is not possible for some reason

app/helpers/katello/subscription_mailer_helper.rb
```diff
@@ -0,0 +1,25 @@
+module Katello
+  module SubscriptionMailerHelper
+    include Rails.application.routes.url_helpers
+
+    def report_url
+      report_id = ReportTemplate.find_by(name: "Subscription - Entitlement Report").id
```
jturel, line 6: These queries, I think, would be better off taking advantage of the helper and create methods, such as this:

```
def report_template
  @report_template ||= ReportTemplate.find_by_name("Subscription - Entitlement Report")
end
```

Idea here is that we can use this in multiple places, and memoize the value so the query will be ran only once. we can also grab more data than just the `id` if we need to for some reason later on, with fewer changes to the code.

app/helpers/katello/subscription_mailer_helper.rb
```diff
@@ -0,0 +1,25 @@
...
+      days_from_now = TemplateInput.find_by_name("Days from Now").id.to_s
+      mail_query = UserMailNotification.first.mail_query.to_s
```
jturel, line 8: The `mail_query` is available to us from the `SubscriptionMailer` so this method could be called from the mailer class with a new argument providing the mail_query

jlsherrill:
test failures are unrelated fixed by @ianballou, but rubocop failure is legit

app/helpers/katello/subscription_mailer_helper.rb
sseelam2: Yes! That's a great idea. thanks.

app/mailers/katello/subscription_mailer.rb
```diff
@@ -1,10 +1,13 @@
 module Katello
   class SubscriptionMailer < ApplicationMailer
+    helper :'katello/subscription_mailer'
     after_action :prevent_sending_blank_report
+    include SubscriptionMailerHelper
 
     def subscription_expiry(options)
       user = ::User.find(options[:user])
       days_from_now = options[:query]
+      start_report_task(days_from_now)
```
jturel, line 10: Last comment from me - just need to ensure we don't start the report task if we're not going to send the email. I suppose we could do it at the bottom of this method `start_report_task(days_from_now) unless @pools.empty?`

app/helpers/katello/subscription_mailer_helper.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    def report_url
+      base_url = report_data_report_template_path(@report_template.id, job_id: @provider_job_id)
+      "#{Setting[:foreman_url]}#{base_url}"
+    end
+
+    def start_report_task(mail_query)
+      @report_template ||= ReportTemplate.find_by(name: "Subscription - Entitlement Report")
+      days_from_now = TemplateInput.find_by_name("Days from Now").id.to_s
+      params = { format: 'csv', template_id: @report_template.id , input_values: { days_from_now => { value: mail_query} } }
```
jeremylenz, line 13: I think it's less confusing to name this `template_input_id`.

app/helpers/katello/subscription_mailer_helper.rb
jeremylenz: Fyi, I asked for feedback about this on Aha. Will pass on the response here

app/views/katello/subscription_mailer/subscription_expiry.html.erb
```diff
@@ -40,6 +40,9 @@
         </tr>
       <% end %>
     </table>
+  <% if @pools.any? %>
```
jturel, line 43: Is this necessary? We don't send the email if there are no pools to begin with

app/views/katello/subscription_mailer/subscription_expiry.text.erb
```diff
@@ -13,3 +13,6 @@
   <%= _("Days Remaining") %>:       <%= pool.days_until_expiration %>
 
 <% end %>
+<% if @pools.any? %>
```
jturel, line 16: Ditto my comment for the ERB

app/views/katello/subscription_mailer/subscription_expiry.html.erb
sseelam2: html.erb gets evaluated even if the pools are empty but the mail does not get delivered. I have moved the logic into the helper.

app/helpers/katello/subscription_mailer_helper.rb
```diff
@@ -0,0 +1,18 @@
+module Katello
+  module SubscriptionMailerHelper
+    include Rails.application.routes.url_helpers
+
+    def report_url
+      base_url = report_data_report_template_path(@report_template.id, job_id: @provider_job_id) unless @pools.empty?
```
jturel, line 6: It would be better to not have this code know anything about @pools since it doesn't use that value if it's not empty

a clearer check would be `if @report_template && @provider_job_id`

app/helpers/katello/subscription_mailer_helper.rb
```diff
@@ -0,0 +1,18 @@
...
+      "#{Setting[:foreman_url]}#{base_url}"
+    end
+
+    def start_report_task(mail_query)
```
jturel, line 10: the variable name `mail_query` is a little vague, wouldn't `days_from_now` be clearer?

app/helpers/katello/subscription_mailer_helper.rb
```diff
@@ -0,0 +1,18 @@
...
+      @report_template ||= ReportTemplate.find_by(name: "Subscription - Entitlement Report")
```
jturel, line 11: May have misunderstood my original feedback here - but the way this is written now makes the `||=` unnecessary - can just use normal assignment

app/helpers/katello/subscription_mailer_helper.rb
```diff
@@ -0,0 +1,18 @@
...
+      template_input_id = TemplateInput.find_by_name("Days from Now").id.to_s
```
jturel, line 12: I just learned something. TemplateInput 'names' are not unique - so I think we need to make sure that we are looking at the TemplateInput with the name 'Days from Now' that belongs to our report template. think 'where clause'

jturel: Also, the to_s on the ID is probably unnecessary

jeremylenz: @jturel I think the `.to_s` may be necessary since Foreman tries the hash key as a string [here](https://github.com/theforeman/foreman/blob/62ef3e89d576611a74f132af84a67c750cb02e0d/app/models/report_composer.rb#L158)

app/views/katello/subscription_mailer/subscription_expiry.text.erb
```diff
@@ -13,3 +13,5 @@
   <%= _("Days Remaining") %>:       <%= pool.days_until_expiration %>
 
 <% end %>
+
+<%= _("Report") %>:               <%= report_url %>
```
jeremylenz, line 17: To me this is almost too pithy.  What about 
* "Full report"
* "View the full report"
* "View the full report, including affected hosts"

Thoughts?

jturel: +1 to 'view the full report'

app/views/katello/subscription_mailer/subscription_expiry.html.erb
```diff
@@ -40,6 +40,7 @@
         </tr>
       <% end %>
     </table>
+  <%= link_to _("Report"), report_url %>
```
jeremylenz, line 43: On my dev box the `report_url` seems to use `http` and not `https`, which causes errors for me.  When I manually change the url to use `https`, it works.  Not sure if this will be an issue in production.. @jturel ?

app/helpers/katello/subscription_mailer_helper.rb
```diff
@@ -0,0 +1,18 @@
...
+      params = { format: 'csv', template_id: @report_template.id, input_values: { template_input_id => { value: mail_query} } }
+      composer = ReportComposer.new(params)
```
jeremylenz, line 14: The params look good to me.  However, when I go to the report page in the web UI, I'm seeing the following:
days_from_now_error
Error stack: http://pastebin.test.redhat.com/860473

app/views/katello/subscription_mailer/subscription_expiry.html.erb
jeremylenz, line 43: I think this should be outside the `<div>`, maybe even with a `<br />` to separate it.  (See picture)

email_with_report_link

jturel: Should be fine - though not sure why Setting[:foreman_url] is using http in dev

app/views/katello/subscription_mailer/subscription_expiry.text.erb
sseelam2: yes. "View the full report" sounds good. thanks!

app/views/katello/subscription_mailer/subscription_expiry.html.erb
```diff
@@ -41,5 +41,6 @@
       <% end %>
     </table>
 </div>
+<br> <%= link_to _("View the full report"), report_url %>
```
jturel, line 44: I think the tag should be `<br/>` and the link (@report_link per my other feedback) on a new line

app/helpers/katello/subscription_mailer_helper.rb
```diff
@@ -0,0 +1,20 @@
+module Katello
+  module SubscriptionMailerHelper
+    include Rails.application.routes.url_helpers
+
+    def report_url
+      if @provider_job_id && @report_template
```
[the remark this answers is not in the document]
sseelam2: Is there any particular reason we need two methods here? is the below sufficient?
```
    def report_link
      base_url = report_data_report_template_path(@report_template.id, job_id: @provider_job_id)
      link_to _("View the full report"), "#{Setting[:foreman_url]}#{base_url}"
    end
```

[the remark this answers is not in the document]
jturel: I haven't seen what the delivered text format email looks like but I was thinking that the `@report_link` would be in the HTML email, and `@report_url` would be used in the text email

app/helpers/katello/subscription_mailer_helper.rb
jeremylenz: @sseelam2 I think this is why I'm getting the error on my end.  `TemplateInput.find_by_name("Days from Now")` finds the first `TemplateInput` with that name, but since I've been working on this feature I've actually amassed _four_ `TemplateInputs` with that name.  The one it happens to find isn't the one attached to the `@report_template`, and therefore the required input is not satisfied.

jeremylenz:
```ruby
[14] pry(main)> TemplateInput.where(name: "Days from Now").map(&:id)
=> [15, 17, 16, 23]

[15] pry(main)> composer = ReportComposer.new({:user_id => 2, "format"=>"csv", "template_id"=>138, "input_values"=>{"15"=>{"value"=>"180"}}})
[16] pry(main)> composer.valid?
=> false

[17] pry(main)> composer = ReportComposer.new({:user_id => 2, "format"=>"csv", "template_id"=>138, "input_values"=>{"23"=>{"value"=>"180"}}})
[18] pry(main)> composer.valid?
=> true

Comment (jturel) [reply app/helpers/katello/subscription_mailer_helper.rb:13]: Need to use a where clause on the report template:

```
template_input_id = @report_template.template_inputs.find_by_name(...).id.to_s
```

File: app/helpers/katello/subscription_mailer_helper.rb [hunk 24]
```diff
@@ -0,0 +1,23 @@
...
+    include Rails.application.routes.url_helpers
+    include ActionView::Helpers::UrlHelper
+
+    def report_url
+      base_url = report_data_report_template_path(@report_template.id, job_id: @provider_job_id)
+      "#{Setting[:foreman_url]}#{base_url}"
+    end
+
+    def report_link
+      link_to _("View the full report"), "#{Setting[:foreman_url]}#{report_url}"
```
jeremylenz, line 12: Both of these use `Setting[:foreman_url]`, so I'm ending up with
```
http://centos7-katello-devel.redhatlaptop.example.comhttp//centos7-katello-devel.redhatlaptop.example.com/templates/report_templates/138/report_data?job_id=1a0b4978-6b7e-43b8-a9ab-4534b998cee4
```

app/helpers/katello/subscription_mailer_helper.rb
jeremylenz: update: We're correctly finding the `template_input_id` now, so this error is resolved. :partying_face:

jeremylenz:
To test this PR:

### Setup
1. Make sure you have subscriptions expiring soon
1. In My Account -> Email Preferences, sign up to receive the Subscriptions Expiring Soon email "daily":
daily
2. In Settings, make sure your Foreman URL is correct and uses `https`
3. In Settings -> Email Settings, set up Foreman to send you emails (the settings that worked for me are below, thanks @jturel)
email_settings
4. Click the "Test email" button and verify that you receive the test email.
5. On your dev box, run
```
bundle exec rake reports:daily
```
6. Wait 0-2 minutes and check your email.  You should have an email like the following:
email_not
7. Click the link in the email to "View the full report".  Verify that the browser downloads the report and you see this screen:
download_page

jturel:
[test katello]

app/mailers/katello/subscription_mailer.rb
```diff
@@ -10,6 +12,12 @@ def subscription_expiry(options)
         @pools = Katello::Pool.readable.expiring_in_days(days_from_now)
       end
 
+      if @pools.any?
```
jturel, line 15: I was testing this and overall it's great. I noticed one quirk that we can easily fix here:

- the subscriptions report represents the expiring subscriptions in a **host-wise manner**
- in my environment I had subscriptions expiring but **none of them were attached to my hosts** and so the report was empty

Improve this by checking if any hosts are subscribed to these pools:
```
@affected_hosts = Host::Managed.with_pools_expiring_in_days(days_from_now) 
```

The code in this class needs to be updated so that the report url and link are only included in the email if there are some `@affected_hosts`. I now think the report link text should be renamed from 'View full report' to 'View a report of the affected hosts' or to indicate more clearly what the full report contains vs the email

app/helpers/katello/subscription_mailer_helper.rb
```diff
@@ -0,0 +1,23 @@
...
+    include Rails.application.routes.url_helpers
+    include ActionView::Helpers::UrlHelper
+
+    def report_url
+      base_url = report_data_report_template_path(@report_template.id, job_id: @provider_job_id)
+      "#{Setting[:foreman_url]}#{base_url}"
+    end
+
+    def report_link
+      link_to _("View the full report"), "#{report_url}"
```
jturel, line 12: I think the interpolation around report_url isn't needed

jeremylenz:
[test katello]

johnpmitsch:
[test katello]

app/views/katello/subscription_mailer/subscription_expiry.text.erb
```diff
@@ -13,3 +13,5 @@
   <%= _("Days Remaining") %>:       <%= pool.days_until_expiration %>
 
 <% end %>
+
+<%= _("View a report of the affected hosts") %>:               <%= @report_url %>
```
jturel, line 17: We should only render this line if we have a `@report_url`

app/mailers/katello/subscription_mailer.rb
```diff
@@ -1,13 +1,24 @@
...
     after_action :prevent_sending_blank_report
+    include SubscriptionMailerHelper
 
     def subscription_expiry(options)
       user = ::User.find(options[:user])
       days_from_now = options[:query]
 
       ::User.as(user.login) do
         @pools = Katello::Pool.readable.expiring_in_days(days_from_now)
+        require "pry"; binding.pry;
```
jturel, line 13: pry :)

sseelam2: ah thanks! :)

app/views/katello/subscription_mailer/subscription_expiry.text.erb
sseelam2: ah yes. will include an if condition there.

sseelam2:
Thank you @jturel and @jeremylenz for the great feedback!

PR #8683, opened 2020-04-28
Fixes #29644 - only call to pulp3 on cv delete if needed

jlsherrill:


pcreech:
Test failure:

```
Response body: {"displayMessage":"undefined method `pulp3_enabled?' for nil:NilClass","errors":["undefined method `pulp3_enabled?' for nil:NilClass"]}
```

PR #8686, opened 2020-04-28
Refs #29647 - remove implicit scopes

mmoll:


mmoll:
Only "Refs" because this will need to get re-checked with core aned Katello on the 6.1 update, but with the current deprecation tracking, these are the only ones in Katello.

PR #8689, opened 2020-04-29
Fixes #29653 - fix 0 epoch package profile upload

jlsherrill:
This resolves a couple of issues:
1. Epochs come in with an epoch of 0 as an integeter,
   but the code to calculate nvrea only was checking for a string
2. If we tried to insert an installed package, but got a conflict
   likely due to a race condition (but also problem #1), the code
   would try to insert an inst. package id of zero, which was not
   correct.  this adds code to lookup any installed packages in case
   a conflict occurs, and log an error if at least one could not be found.
3. Adds a migration to correct 1)

### DIFF
--- a/app/services/katello/pulp/simple_package.rb
+++ b/app/services/katello/pulp/simple_package.rb
@@ -13,8 +13,8 @@ module Katello
       end
 
       def nvrea
-        if epoch == "0"
-          @nvra
+        if epoch.nil? || epoch.to_s == "0"
+          nvra
         else
           "#{@name}-#{@epoch}:#{@version}-#{@release}.#{@arch}"
         end
--- a/db/migrate/20200501155054_installed_package_unique_nvrea.rb
+++ b/db/migrate/20200501155054_installed_package_unique_nvrea.rb
@@ -1,0 +1,64 @@
+class InstalledPackageUniqueNvrea < ActiveRecord::Migration[5.2]
+  def fix_missing_attributes
+    #bug in dynflow may have resulted in old code running and not properly populating fields
+    # This block is basically copied from 20200129172534_add_epoch_version_release_arch_to_katello_installed_packages.rb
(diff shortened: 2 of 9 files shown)


app/services/katello/pulp/simple_package.rb
```diff
@@ -13,7 +13,7 @@ def nvra
       end
 
       def nvrea
-        if epoch == "0"
+        if epoch.to_s == "0"
           @nvra
```
ianballou, line 17: I realized that @nvra is going to return nil here, it should be just `nvra`.

ianballou: It looks like I hit this because I am running Pulp 3.

db/migrate/20200501155054_installed_package_unique_nvrea.rb
```diff
@@ -0,0 +1,64 @@
...
+      end
+
+      installed_packages << ::Katello::InstalledPackage.new(attributes_hash)
+    end
+    ::Katello::InstalledPackage.import(installed_packages, validate: false, batch_size: 50_000,
+                                       on_duplicate_key_update: {conflict_target: [:nvra],
+                                                                 columns: [:nvrea, :epoch, :version, :release, :arch]})
+  end
+
+  def consolidate_duplicate_nvreas
```
ianballou, line 27: Why is this method necessary?  Are duplicate installed packages being created as part of the bug?

jlsherrill: In theory, it shouldn't happen, but anytime we take a column and make it unique we need to ensure there are no duplicates.  Otherwise users would be stuck in the middle of a db migration

jlsherrill:
user confirmed (in private email) that this resolved the issue.  i'll work on adding tests and fixing failed tests

jlsherrill:
[test katello]

jlsherrill:
[test katello]

PR #8691, opened 2020-04-30
Fixes #29674 - add github action to count commits

jlsherrill:


.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,20 @@
...
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v1
+      with:
+        fetch-depth: 30
+    - name: Checkout master
+      run: git fetch origin master
+    - name: create local master branch
+      run: git branch master origin/master
+    - name: Conform Action
```
johnpmitsch, line 18: how about calling this something more specific like "Checking for more than one commit"?

jlsherrill: errr  yes, that was a holdover from the old project, thanks for catching

johnpmitsch:
If we could have a way to say this is non-blocking it would help avoid some confusion on whether it needs to be passing before merging, maybe just by putting "non-blocking" in the title of the action?

johnpmitsch:
@jlsherrill I think somewhere it should specify that it is checking for more than one commit, I don't see that being obvious with the titles right now

PR #8692, opened 2020-04-30
Fixes #29553 - Katello Applicability of rpms with modules

ianballou:
Note: this functionality only works with Pulp 3 due to the use of its module stream <-> rpm relationship.

Extends the logic of `fetch_content_ids` in Katello Applicability's `ApplicableContentHelper` to take enabled module streams into account when calculating RPM applicability.

RPMs that are applicable must:
1) Exist within a host's bound library instance repositories
2) Belong to the host
3) Either:
  a) Not be modular, or,
  b) Exist in the host's enabled modules

My PR here is a work in progress because I think the queries could be combined in a more efficient way.

### DIFF
--- a/app/services/katello/applicability/applicable_content_helper.rb
+++ b/app/services/katello/applicability/applicable_content_helper.rb
@@ -21,37 +21,61 @@ module Katello
 
       def fetch_content_ids
         if self.content_unit_class == ::Katello::Erratum
-          # Query for all Errata ids that are attached to the host's applicable packages
-          query = 'select katello_repository_errata.erratum_id as id from katello_repository_errata
-                       inner join katello_erratum_packages
-                          on  katello_repository_errata.erratum_id  = katello_erratum_packages.erratum_id
-                       inner join katello_rpms
-                          on katello_rpms.nvra = katello_erratum_packages.nvrea
-                       inner join katello_content_facet_applicable_rpms on
-                           katello_content_facet_applicable_rpms.rpm_id = katello_rpms.id
-                       where
-                            "katello_content_facet_applicable_rpms"."content_facet_id" = :content_facet_id
-                       AND katello_repository_errata.repository_id IN (:repo_ids)'
-
-          return Katello::Erratum.find_by_sql([query, {content_facet_id: content_facet.id, repo_ids: self.bound_library_instance_repos} ]).map(&:id).uniq
+          fetch_errata_content_ids
         elsif self.content_unit_class == ::Katello::ModuleStream
           fail NotImplementedError
         else
-          # Query for applicable RPM ids
-          return ::Katello::Rpm.joins("INNER JOIN katello_repository_rpms ON \
-                                      katello_rpms.id = katello_repository_rpms.rpm_id").
-                                      joins("INNER JOIN katello_installed_packages ON \
-                                            katello_rpms.name = katello_installed_packages.name AND \
-                                            katello_rpms.arch = katello_installed_packages.arch AND \
-                                            katello_rpms.evr > katello_installed_packages.evr").
-                                            joins("INNER JOIN katello_host_installed_packages ON \
-                                                  katello_installed_packages.id = \
-                                                  katello_host_installed_packages.installed_package_id WHERE \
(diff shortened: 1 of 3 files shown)


app/services/katello/applicability/applicable_content_helper.rb
```diff
@@ -38,19 +38,41 @@ def fetch_content_ids
...
+              katello_rpms.id = katello_repository_rpms.rpm_id").
+            joins("INNER JOIN katello_installed_packages ON
+              katello_rpms.name = katello_installed_packages.name AND
+              katello_rpms.arch = katello_installed_packages.arch AND
+              katello_rpms.evr > katello_installed_packages.evr").
+            joins("INNER JOIN katello_host_installed_packages ON
+              katello_installed_packages.id = katello_host_installed_packages.installed_package_id WHERE
+                katello_repository_rpms.repository_id in (#{self.bound_library_instance_repos.join(',')}) and
+                katello_host_installed_packages.host_id = #{self.content_facet.host.id} and
+                (katello_rpms.modular = FALSE" + rpms_in_modules_addendum + ')').pluck(:id).uniq
```
jlsherrill, line 75: This is a tad simpler, and works in a single query (ignoring the bound repos being passed in):

```
          enabled_module_stream_ids = ::Katello::ModuleStream.
            joins("inner join katello_available_module_streams on
              katello_module_streams.name = katello_available_module_streams.name and
                  katello_module_streams.stream = katello_available_module_streams.stream").
            joins("inner join katello_host_available_module_streams on
              katello_available_module_streams.id = katello_host_available_module_streams.available_module_stream_id where
                katello_host_available_module_streams.host_id = #{self.content_facet.host.id} and
                katello_host_available_module_streams.status = 'enabled'").select(:id)

          query =  ::Katello::Rpm.
            joins("INNER JOIN katello_repository_rpms ON
              katello_rpms.id = katello_repository_rpms.rpm_id").
            joins("INNER JOIN katello_installed_packages ON
              katello_rpms.name = katello_installed_packages.name AND
              katello_rpms.arch = katello_installed_packages.arch AND
              katello_rpms.evr > katello_installed_packages.evr").
            joins("LEFT JOIN katello_module_stream_rpms ON
                   katello_rpms.id = katello_module_stream_rpms.rpm_id").
            joins("INNER JOIN katello_host_installed_packages ON
              katello_installed_packages.id = katello_host_installed_packages.installed_package_id").
            where("katello_repository_rpms.repository_id in (#{self.bound_library_instance_repos.join(',')})").
            where("katello_host_installed_packages.host_id = #{self.content_facet.host.id}").
            where("katello_module_stream_rpms.module_stream_id is null or katello_module_stream_rpms.module_stream_id in (:enabled_module_streams)", :enabled_module_streams => enabled_module_stream_ids).
            pluck(:id).uniq
```

jlsherrill: also, its probably time to split these 3 types of content out into their own method

jlsherrill: can you convert the other where() clauses to also use the prepared statement format like 3rd where clause there? Its safer, and you dont have to manually join()

jlsherrill:
looking at the query and playing around in postgresql, i think adding an index on  on katello_host_available_module_streams (host_id, available_module_stream_id, status)  would help

app/services/katello/applicability/applicable_content_helper.rb
ianballou: What do you mean by not having to manually join?

ianballou:
@jlsherrill I've addressed your comments.

app/services/katello/applicability/applicable_content_helper.rb
jlsherrill: i think my brain malfunctioned, i have no idea what i meant by that, but it looks like you got what i meant!

jlsherrill: ohhhh  the  array  join(',')  is what i meant!

ianballou: Ah that makes sense! Well that's taken care of then.

PR #8694, opened 2020-05-01
Fixes #29563 - Applicable packages page loading error

ianballou:
Fixes the broken applicable packages page.

### DIFF
--- a/app/controllers/katello/api/v2/packages_controller.rb
+++ b/app/controllers/katello/api/v2/packages_controller.rb
@@ -4,7 +4,7 @@ module Katello
     include Katello::Concerns::Api::V2::RepositoryContentController
 
     before_action :find_repositories, :only => [:auto_complete_name, :auto_complete_arch]
-    before_action :find_host, :only => :index
+    before_action :find_hosts, :only => :index
 
     def auto_complete_name
       page_size = Katello::Concerns::FilteredAutoCompleteSearch::PAGE_SIZE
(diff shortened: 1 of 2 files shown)


jlsherrill:
mind adding a test?

app/controllers/katello/api/v2/packages_controller.rb
```diff
@@ -55,7 +55,7 @@ def custom_index_relation(collection)
       upgradable = ::Foreman::Cast.to_bool(params[:packages_restrict_upgradable])
       if applicable || upgradable
         hosts = @host ? [@host] : ::Host::Managed.authorized("view_hosts")
```
jturel, line 57: The way this code is organized seems unnecessarily confusing. It's not because of your change and I know you're trying to make a quick fix. A way to clarify this code overall could be to refactor `find_host` into `find_hosts`:

```
    def find_hosts                                                                                                                                                                            
      @hosts = ::Host::Managed.authorized("view_hosts")                                                                                                                                       
      if params[:host_id]                                                                                                                                                                     
        @hosts = @hosts.where(:id => params[:host_id])                                                                                                                                        
        fail HttpErrors::NotFound, _('Could not find a host with id %s') % params[:host_id] if @hosts.empty?                                                                                  
      end                                                                                                                                                                                     
      @hosts = @hosts.where(:organization_id => params[:organization_id]) if params[:organization_id]                                                                                         
    end
```

then the `custom_index_relation` becomes
```
    def custom_index_relation(collection)                                                                                                                                                     
      applicable = ::Foreman::Cast.to_bool(params[:packages_restrict_applicable]) || @hosts.any?                                                                                                     
      upgradable = ::Foreman::Cast.to_bool(params[:packages_restrict_upgradable])                                                                                                             
                                                                                                                                                                                              
      if applicable || upgradable                                                                                                                                    
        if upgradable                                                                                                                                                                         
          collection = collection.installable_for_hosts(@hosts)                                                                                                                               
        elsif applicable                                                                                                                                                                      
          collection = collection.applicable_to_hosts(@hosts)                                                                                                                                 
        end                                                                                                                                                                                   
      end                                                                                                                                                                                     
      collection                                                                                                                                                                              
    end
```

I think the end result is the same and to me the code is a lot easier to understand

jturel: I'll leave it up to you as to whether you want to make this change, but I do think it's an improvement especially because it will always return a relation for `@hosts` which is generally useful!

ianballou: Looks like it should work well! I'll try it out, always like an improvement.

ianballou:
@jlsherrill @jturel I've updated the tests and switched the code over to the `find_hosts` style.

test/controllers/api/v2/packages_controller_test.rb
```diff
@@ -9,6 +9,7 @@ def models
       @host = hosts(:one)
       @simple_filter = katello_content_view_filters(:simple_filter)
       @one_package_rule = katello_content_view_package_filter_rules(:one_package_rule)
+      @org = ::Organization.find_by(name: "Empty Organization")
```
jturel, line 12: we have a helper, `get_organization`, that will load an org from fixtures so I'd recommend using that. it loads the empty organization by default (no argument) otherwise it loads the fixture name specified!

ianballou: Thanks, I've pushed the changes now.

jturel:
I ran into this bug while working on another change - merging so I can continue :)

PR #8695, opened 2020-05-01
Fixes #29678 - do not lock smart proxy sync

jlsherrill:


wbclark:
Tested on a reproducer machine and confirmed that it resolved the issue

sjha4:
The code change should create non-exclusive locks as needed and tested above. Ack :+1:

PR #8696, opened 2020-05-04
Fixes #29709 - Remove deleted repositories post migration

sjha4:
To test:
On pulp2 create a file repo
Run bundle exec rails katello:pulp3_migration
Delete file repo on pulp2
Run bundle exec rails katello:pulp3_migration
See that the migrated file repo is not deleted in pulp3
Run bundle exec rails katello:pulp3_post_migration_check
Verify that the deleted file repo is deleted in pulp3

This also needs to fix https://projects.theforeman.org/issues/29595 to not have unmigrated docker tags from orphaned repos.

### DIFF
--- a/app/lib/actions/pulp3/orphan_cleanup/delete_orphaned_migrated_repositories.rb
+++ b/app/lib/actions/pulp3/orphan_cleanup/delete_orphaned_migrated_repositories.rb
@@ -1,0 +1,15 @@
+module Actions
+  module Pulp3
+    module OrphanCleanup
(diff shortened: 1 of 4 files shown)


ianballou:
After the discussion in our meeting I tested out this bug again: https://projects.theforeman.org/issues/29595

This PR doesn't seem to fix the issue, but I'm not sure it has to.

sjha4:
Thanks Ian..Ya..I checked that too with the delete orphaned content..Will merge this and move that card along..

PR #8697, opened 2020-05-04
Fixes #29692 - Enable installation of Tracer from the UI

jturel:
This PR adds a button to install the katello-host-tools-tracer package from the host details page and Traces tab if it's not installed. Also adds a tooltip on the details page to explain what Tracer does. Clicking the button will redirect to the package actions page with the form input pre-filled with 'katello-host-tools-tracer' so it can be installed via agent or REX

May wanna test this in conjunction with https://github.com/Katello/katello-host-tools/pull/111 for an easy 2 for 1.  found the bug while implementing this

### DIFF
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/content/content-host-packages-actions.controller.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/content/content-host-packages-actions.controller.js
@@ -3,12 +3,19 @@
  * @name  Bastion.content-hosts.controller:ContentHostPackagesActionsController
  *
  * @requires $scope
+ * @requires $location
  *
  * @description
  *   Provides the functionality for the content host package actions.
  */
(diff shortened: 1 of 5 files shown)


jturel:
> One comment: After the task was complete and showing a result of "success", I clicked `katello-client-3` in the breadcrumb bar and the page appeared immediately, but still looked like the Before picture. I had to refresh the page in the browser to see the After. This is my Angular ignorance showing.. but how hard would it be to force an update so that's not needed?


Good question. I don't think we're doing that anywhere so there could be some effort involved. I'll see if it's not too much work though.

jturel:
@jeremylenz I got it working. Way easier than I thought. I learned how to do it from here: https://stackoverflow.com/questions/28800426/what-is-on-in-angularjs

jeremylenz:
@jturel Beautiful, works great now :smile: 

APJ

jturel:
[test katello]

PR #8698, opened 2020-05-05
Fixes #29209 - Do not import_all for pulp3 content types

sjha4:
To test:
**On master:** 
1. On a box with file, docker, (yum) enabled on pulp3 run bundle exec rails katello:reimport 
2. Notice the failures. import_all is not supported with pulp3 content types since content api endpoints in pulp3 do not return associated repositories.

**On this branch:**
1. bundle exec rails katello:reimport
2. Notice output ignoring all pulp3 content types.

### DIFF
--- a/lib/katello/tasks/reimport.rake
+++ b/lib/katello/tasks/reimport.rake
@@ -14,27 +14,23 @@ namespace :katello do
   desc "Reimports information from backend systems"
   task :reimport => ["environment", "katello:check_ping"] do
     User.current = User.anonymous_admin #set a user for orchestration
-
     Dir.glob(Katello::Engine.root.to_s + '/app/models/katello/*.rb').each { |file| require file }
 
-    models = [Katello::Erratum,
-              Katello::PackageGroup,
-              Katello::PuppetModule,
-              Katello::Rpm,
-              Katello::Srpm,
-              Katello::ModuleStream,
-              Katello::YumMetadataFile,
-              Katello::FileUnit,
-              Katello::Subscription,
-              Katello::Pool,
-              Katello::DockerManifest,
-              Katello::DockerManifestList,
-              Katello::DockerTag,
-              Katello::ContentViewPuppetEnvironment,
-              Katello::Content]
+    models = [
+      Katello::Subscription,
+      Katello::Pool,
+      Katello::ContentViewPuppetEnvironment,
+      Katello::Content
+    ]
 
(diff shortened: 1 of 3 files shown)


lib/katello/tasks/reimport.rake
```diff
@@ -35,7 +41,10 @@ namespace :katello do
 
     models << Katello::OstreeBranch if Katello::RepositoryTypeManager.find(Katello::Repository::OSTREE_TYPE).present?
     models << Katello::Deb if Katello::RepositoryTypeManager.find(Katello::Repository::DEB_TYPE).present?
-
+    ignore_content_types.each do |model|
+      puts "\e[33mIgnoring pulp3 enabled content type : #{model.name}\e[0m\n"
+    end
+    models -= ignore_content_types
```
jlsherrill, line 47: Can we use the RepositoryTypeManager to get rid of the list of content models?  I think there are some that are missing.

Something like (haven't checked actual syntax):
```
models = [ 
              Katello::Subscription,
              Katello::Pool,
              Katello::Content
]
Katello::RepositoryTypeManager.repository_types.each |repo_type|
    if SmartProxy.pulp_master.pulp3_support(repo_type)
       puts "Ignoring types: #{repo_type.indexable_content_types}"
   else
      models += repo_type.indexable_content_types
   end
end
```

jlsherrill: that could also get rid of the check for ostree and deb_type, since it would be built into the 'loop'

sjha4: Updated with the changes as you pointed. Added Katello::ContentViewPuppetEnvironment to list of models to import.

lib/katello/tasks/reimport.rake
```diff
@@ -14,27 +14,23 @@ namespace :katello do
...
-    models << Katello::Deb if Katello::RepositoryTypeManager.find(Katello::Repository::DEB_TYPE).present?
+    models = [
+      Katello::Subscription,
+      Katello::Pool,
+      Katello::ContentViewPuppetEnvironment,
+      Katello::Content
+    ]
+
+    Katello::RepositoryTypeManager.repository_types.each_value do |repo_type|
+      indexable_types = repo_type.content_types.select { |c| c.index }
```
jlsherrill, line 27: nitpick, but you can use  repo_type.content_types_to_index  instead of checking it here :)

jlsherrill:
tested and everything works great, just the one small comment

PR #8702, opened 2020-05-06
Fixes #29448 - do not import puppet if not enabled

jlsherrill:
this adds a check when we trigger an  import of puppet classes
as part of a CV publish.  If a smart proxy does not have the feature,
then we know that import cannot take place

### DIFF
--- a/app/lib/katello/foreman.rb
+++ b/app/lib/katello/foreman.rb
@@ -21,7 +21,8 @@ module Katello
             foreman_environment.save!
           end
 
-          if (foreman_smart_proxy = SmartProxy.pulp_master)
+          foreman_smart_proxy = SmartProxy.pulp_master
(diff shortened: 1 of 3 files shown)


jlsherrill:
You can test this by:

1.  creating a puppet repo
2.  upload a puppet module
3.  shut down your puppet server 'systemctl stop puppetserver'
4. publish a cv with a puppet module  <- this will fail
5. edit /etc/foreman-proxy/settings.d/puppet.yml  set :enabled:  to 'false'
6.  restart the foreman-proxy service
7. refresh the smart proxy in the UI
8. resume the CV publish, try to publish again.  It should now work.

jlsherrill:
hrmm, that would be unrelated, but i'll take a look

jlsherrill:
pushed a fix for that issue.  The count caching we were doing wasn't accounting for puppet explicitly.

PR #8644, opened 2020-04-02
Fixes #29298 - Add new Content View table

johnpmitsch:
This adds a new content view table using patternfly 4 to the page "/labs/content_views"

PR #8703, opened 2020-05-07
Fixes #29717 - Fix React tests

johnpmitsch:
Recently nock was introduced to testing. Importing `nock` from a place we also use `axios-mock-adapter` was causing issues as these mock network requests differently. Nock mocks from the nodejs level and axios-mock-adapter mocks the axios library. So initializing axios-mock-adapter and nock caused axios-mock-adapter to try to interecept the request before nock could.

We didn't see this locally because our dev env is on nodejs 10 and jenkins is on 12. I'm not sure why some runs passed and others have not.

Its likely a snapshot has to be re-recorded on nodejs 12 as well for the SubscriptionActions failure.

### DIFF
--- a/webpack/mockRequest.js
+++ b/webpack/mockRequest.js
@@ -1,6 +1,5 @@
 import axios from 'axios';
 import MockAdapter from 'axios-mock-adapter';
-import nock from 'nock';
 
 // TODO: figure out way to reuse this from foreman
 export const mock = new MockAdapter(axios);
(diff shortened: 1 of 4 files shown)


johnpmitsch:
I updated this to use `afterEach` for the Subscriptions Actions test. It seemed to be an asynchronous bug, from what I could tell a mock was sticking around from a previous test. The test would pass if I wrapped the `expect` assertion with a `setTimeout` function. I think the recent changes to use foremanReact caused some slight timing changes, but I'm not quite sure exactly what was happening, but I can say for sure asynchronous code had something to do with it. Using `beforeEach` seems to ensure a clean environment better than `afterEach` callback. Maybe we want to make this change elsewhere, but lets start with this and see if we have any more issues.

One other thing we can think about changing is making sure our CI tests check out katello as a sibling of foreman and run tests that way, so this would be the same as a devel environment. Currently foreman is checked out inside of katello in the CI environment. This is especially important if we are using code from Foreman in testing with `foremanReact`. It seems like there is an issue with this format if one does not install npm packages for Foreman (and only for Katello as we do now in Jenkins), but perhaps there is a better fix. It could make sense to install packages for both foreman and katello in CI, especially if we could use Foreman's dependencies during testing and remove Katello's duplicate dev dependencies. We could even consider solely using github actions or travis for JS testing. Seeing these issues in the devel environment would prevent this kind of thing from being checked into master, even if it is somewhat transient of an error.

Just wanted to share some broader thoughts, everything should be fixed now :)

jturel:
Nice analysis. So, is the nock stuff actually relevant here? I'm not against merging it in - just asking.

johnpmitsch:
> Nice analysis. So, is the nock stuff actually relevant here? I'm not against merging it in - just asking.

thanks and ya, the tests that were introduced in https://github.com/Katello/katello/pull/8644 use nock and were failing as well as the subscription ones. Interestingly it happened transiently as well. There may be more at play here but this was as far as I could figure out for now. Hopefully we don't see more async or mock issues like this but I can dig further if we do!

PR #8705, opened 2020-05-08
Fixes #29604 - Katello Applicability calculate module streams

ianballou:
Adds Katello Applicability support for calculating which module streams have upgrades available.

Note: this PR requires the Pulp 3 RPM plugin

Start by enabling Katello Applicability in `katello.yaml`:
```
:katello_applicability: true
```

One way to test:
1) Sync this repo: https://partha.fedorapeople.org/test-repos/pteradactly-with-dino-errata/
2) Run these commands on a registered and properly subscribed el8 system: 
  - `dnf module enable duck`
  - `dnf install  duck-0.6-1.noarch`
3) Check that `duck-0.7-1.noarch` is an applicable package and that the duck module is marked "Upgrade Available"

### DIFF
--- a/app/services/katello/applicability/applicable_content_helper.rb
+++ b/app/services/katello/applicability/applicable_content_helper.rb
@@ -23,7 +23,7 @@ module Katello
         if self.content_unit_class == ::Katello::Erratum
           fetch_errata_content_ids
         elsif self.content_unit_class == ::Katello::ModuleStream
-          fail NotImplementedError
+          fetch_module_stream_content_ids
         else
           fetch_rpm_content_ids
         end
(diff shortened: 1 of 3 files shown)


app/services/katello/applicability/applicable_content_helper.rb
```diff
@@ -45,6 +45,22 @@ def fetch_errata_content_ids
...
+        query = 'select katello_repository_module_streams.module_stream_id as id from katello_repository_module_streams
+                     inner join katello_module_stream_rpms
+                        on  katello_repository_module_streams.module_stream_id  = katello_module_stream_rpms.module_stream_id
+                     inner join katello_rpms
+                        on katello_rpms.id = katello_module_stream_rpms.rpm_id
+                     inner join katello_content_facet_applicable_rpms on
+                         katello_content_facet_applicable_rpms.rpm_id = katello_rpms.id
+                     where
+                          "katello_content_facet_applicable_rpms"."content_facet_id" = :content_facet_id
+                     AND katello_repository_module_streams.repository_id IN (:repo_ids)'
```
jeremylenz, line 59: nit: I prefer SQL keywords to be uppercase (SELECT, INNER JOIN etc.) but if you don't like that it should at least be consistent within the same SQL statement.

jeremylenz: If there's ever a migration in the future that changes a table name etc., this would have to be manually updated.  Is this query too difficult to do with Active Record `.joins`, `.ids`, etc?

app/services/katello/applicability/applicable_content_helper.rb
```diff
@@ -45,6 +45,22 @@ def fetch_errata_content_ids
...
+
+        return Katello::ModuleStream.find_by_sql([query, {content_facet_id: content_facet.id, repo_ids: self.bound_library_instance_repos} ]).map(&:id).uniq
```
jeremylenz, line 61: I think SELECT DISTINCT may be faster than making Ruby do it with `.uniq`?  Also Active Record has a `.distinct` method

app/services/katello/applicability/applicable_content_helper.rb
ianballou: @jeremylenz I can look at changing the query to use more Active Record methods, I don't think we need the manual query necessarily.  I'll change the errata method above if so too.

app/services/katello/applicability/applicable_content_helper.rb
ianballou: I can change this and the errata method to distinct, sure

ianballou:
@jeremylenz I've cleaned up some things and replaced the `uniq` with a `SELECT DISTINCT`.

It doesn't look like that hard-coded table names can be avoided due to the custom joins going on here.

jeremylenz:
[test katello]

jlsherrill:
Code looks good to me too!

PR #8706, opened 2020-05-11
Fixes #29730 - Warn if resolving Traces will reboot the host

jturel:
- spin up katello-client
- install katello-host-tools-tracer
- upgrade the kernel
- resolving the trace from the host details or bulk hosts UI will warn of the restart involved

- downgrade the kernel
- the restart required messaging shouldn't be shown


chris1984:
Starting to test

PR #8277, opened 2019-08-14
Add Deb Package applicability

m-bucher:
This PR depends on the `deb_version_cmp` PostgreSQL-function, introduced in #7961.

PR #8708, opened 2020-05-12
Fixes #29768 - Use debversion extension for sorting deb-versions

m-bucher:
This requires that `rh-postgresql12-postgresql-debversion` package is installed.
Additionally the extension must be activated in the database:
```
CREATE EXTENSION debversion;
```
Not sure if this can/should be done in the migration-script.

Having this merged would make it possible to merge/review #7961 and #8277, because they rely on a possibility to compare Debian version numbers.

### DIFF
--- a/app/models/katello/deb.rb
+++ b/app/models/katello/deb.rb
@@ -29,5 +29,10 @@ module Katello
     def self.total_for_repositories(repos)
       self.in_repositories(repos).count
     end
+
+    def version=(version)
+      super(version)
+      self.version_sortable = version
+    end
   end
 end
--- a/config/initializers/postgresql_custom_types.rb
+++ b/config/initializers/postgresql_custom_types.rb
@@ -1,0 +1,39 @@
+require 'active_record/connection_adapters/postgresql_adapter'
+
+module ActiveRecord
+  module ConnectionAdapters
+    if const_defined? :PostgreSQLAdapter
+      class PostgreSQLAdapter
+        NATIVE_DATABASE_TYPES.merge!(
+          debversion: { name: 'debversion' }
+        )
+      end
+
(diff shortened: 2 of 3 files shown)


m-bucher:
OK, so the test fails because `db/schema.rb` is not correctly created:
```
# Could not dump table "katello_debs" because of following StandardError
#   Unknown type 'debversion' for column 'version_sortable'
```
Any ideas?

m-bucher:
Patched the `debversion` into the PostgreSQLAdapter :grin:

chris1984:
@m-bucher what are the steps for me to test this? Just install the package and run a db:migrate/seed?

m-bucher:
@chris1984 yes, if the postgresql-debversion package is installed, the migration should work.

chris1984:
Awesome will test it Monday

chris1984:
@m-bucher 

To testing the sorting, do I just need to upload a few deb packages of different names, or the same name but with different versions?

m-bucher:
@chris1984, either that or you just sync any apt-repository (I do not recommend Debian stable repository because it is fairly large :grin: ).
Something like Ubuntu focal-security should have a reasonable size, because it is fairly new:
URL | Release | Architecture
------------ | ------------- | -----
http://archive.ubuntu.com/ubuntu/ | focal-security | amd64

Then you can compare the results of `ORDER BY version` and `ORDER BY version_sortable` in the DB-console.
At least that's what I did :smiley:

jlsherrill:
This looks good to me, just waiting on https://github.com/theforeman/puppet-katello/pull/341/files so we don't break nightlies

jlsherrill:
I think this is good to go. Thanks @m-bucher !

PR #8709, opened 2020-05-13
Fixes #29790 - New macro to show only Red Hat subscriptions

snagoor:
Adding this new macro `host_redhat_subscriptions` to report only Red Hat subscription consumption. 

Here the assumption is that the `support_level` field for Custom Subscriptions are always null.

### DIFF
--- a/app/lib/katello/concerns/base_template_scope_extensions.rb
+++ b/app/lib/katello/concerns/base_template_scope_extensions.rb
@@ -8,7 +8,8 @@ module Katello
           super + [:errata, :host_subscriptions, :host_applicable_errata_ids, :host_applicable_errata_filtered,
                    :host_latest_applicable_rpm_version, :load_pools, :load_errata_applications, :host_content_facet,
                    :host_sla, :host_products, :sub_name, :sub_sku, :registered_through, :last_checkin, :host_collections,
-                   :host_subscriptions_names, :host_subscriptions, :host_products_names, :host_collections_names]
+                   :host_subscriptions_names, :host_subscriptions, :host_products_names, :host_collections_names,
+                   :host_redhat_subscription_names]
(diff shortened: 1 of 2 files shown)


app/lib/katello/concerns/base_template_scope_extensions.rb
```diff
@@ -28,6 +29,10 @@ def host_subscriptions_names(host)
         host.subscriptions.map(&:name)
       end
 
+      def host_redhat_subscriptions(host)
+        host.subscriptions.where.not(:support_level => nil).map(&:name)
```
jturel, line 33: We can improve this check to find a redhat subscription by adding a scope here:

https://github.com/Katello/katello/blob/d6cb33076d4606676086c0159282aaef5022fda9/app/models/katello/subscription.rb#L13

It would look like this:
```
scope :redhat, ->() { joins(:products).merge(Katello::Product.redhat).distinct }
```

So you can call:

```
host.subscriptions.redhat.pluck(:name)
```

This generates an optimal query, and we can re-use the new scope in other areas of the code in the future :)

app/lib/katello/concerns/base_template_scope_extensions.rb
```diff
@@ -28,6 +29,10 @@ def host_subscriptions_names(host)
         host.subscriptions.map(&:name)
       end
 
+      def host_redhat_subscriptions(host)
```
jturel, line 32: It all works well. To be consistent with the other methods I think this one should end in _names such as `host_redhat_subscription_names`

snagoor: @jturel thank you, modified as suggested.

jturel:
[test katello]

PR #8711, opened 2020-05-13
Fixes #29791 - Disable subscription attach for Simple Content Access

jeremylenz:
[sat-e-492] As a user, I should not be able to attach a subscription to a host with SCA enabled

* `PUT hosts/:host_id/subscriptions/add_subscriptions` will now return a Bad Request
* Content Hosts page:
  * Subscription Status section does not show
* Content Hosts -> details -> Subscriptions:
  * Tab does not show
  * Simple Content Access Banner is visible, explaining that you don't need subscriptions

- [x] Disallow API call to attach subscriptions
- [x] Don’t show Subscriptions tab in Content Host details page
- [x] Don't show ‘Subscription Status’ table column
- [x] Change messaging in Subscription section of Content Host Details

### DIFF
--- a/app/controllers/katello/api/v2/host_subscriptions_controller.rb
+++ b/app/controllers/katello/api/v2/host_subscriptions_controller.rb
@@ -134,6 +134,10 @@ module Katello
       param :quantity, :number, :desc => N_("Quantity of this subscriptions to add"), :required => true
     end
     def add_subscriptions
+      if @host.organization.simple_content_access?
+        fail ::Katello::HttpErrors::BadRequest, _("This host's organization has Simple Content Access enabled.  Attaching subscriptions is neither necessary nor allowed.")
+      end
+
       pools_with_quantities = params.require(:subscriptions).map do |sub_params|
         PoolWithQuantities.new(Pool.with_identifier(sub_params['id']), sub_params['quantity'].to_i)
       end
--- a/app/controllers/katello/api/v2/host_subscriptions_controller.rb
+++ b/app/controllers/katello/api/v2/host_subscriptions_controller.rb
@@ -135,7 +135,7 @@ module Katello
     end
     def add_subscriptions
       if @host.organization.simple_content_access?
-        fail ::Katello::HttpErrors::BadRequest, _("This host's organization has Simple Content Access enabled.  Attaching subscriptions is neither necessary nor allowed.")
+        fail ::Katello::HttpErrors::BadRequest, _("This host's organization is in Simple Content Access mode. Attaching subscriptions is disabled.")
       end
 
       pools_with_quantities = params.require(:subscriptions).map do |sub_params|
--- a/app/controllers/katello/api/v2/hosts_bulk_actions_controller.rb
+++ b/app/controllers/katello/api/v2/hosts_bulk_actions_controller.rb
@@ -10,6 +10,7 @@ module Katello
     before_action :find_deletable_hosts, :only => [:destroy_hosts]
     before_action :find_readable_hosts, :only => [:applicable_errata, :installable_errata, :available_incremental_updates]
     before_action :find_errata, :only => [:available_incremental_updates]
+    before_action :find_organization, :only => [:add_subscriptions]
     before_action :deprecate_katello_agent, :only => [:install_content, :update_content, :remove_content]
 
     before_action :validate_content_action, :only => [:install_content, :update_content, :remove_content]
@@ -301,9 +302,7 @@ module Katello
     end
 
     def validate_organization
-      org_id = params[:organization_id] || params[:included]&.[](:params)&.[](:organization_id)
-      fail HttpErrors::BadRequest, _("Organization ID is required") if org_id.blank?
-      @organization = Organization.find org_id
+      fail HttpErrors::BadRequest, _("Organization ID is required") if @organization.blank?
     end
 
     def validate_host_collection_membership_limit
--- a/app/controllers/katello/api/v2/hosts_bulk_actions_controller.rb
+++ b/app/controllers/katello/api/v2/hosts_bulk_actions_controller.rb
@@ -13,6 +13,7 @@ module Katello
     before_action :deprecate_katello_agent, :only => [:install_content, :update_content, :remove_content]
 
     before_action :validate_content_action, :only => [:install_content, :update_content, :remove_content]
+    before_action :validate_organization, :only => [:add_subscriptions]
 
     # disable *_count fields on erratum rabl, since they perform N+1 queries
     before_action :disable_erratum_hosts_count
@@ -173,6 +174,10 @@ module Katello
       param :quantity, :number, :desc => N_("Quantity of this subscriptions to add"), :required => true
     end
     def add_subscriptions
+      if @organization.simple_content_access?
+        fail HttpErrors::BadRequest, _("The specified organization is in Simple Content Access mode. Attaching subscriptions is disabled")
+      end
+
       pools_with_quantities = params.require(:subscriptions).map do |sub_params|
         PoolWithQuantities.new(Pool.find(sub_params['id']), sub_params['quantity'])
       end
@@ -293,6 +298,12 @@ module Katello
 
     def find_deletable_hosts
       find_bulk_hosts(:destroy_hosts, params)
+    end
+
+    def validate_organization
+      org_id = params[:organization_id] || params[:included]&.[](:params)&.[](:organization_id)
+      fail HttpErrors::BadRequest, _("Organization ID is required") if org_id.blank?
+      @organization = Organization.find org_id
     end
 
     def validate_host_collection_membership_limit
--- a/app/controllers/katello/concerns/api/v2/bulk_hosts_extensions.rb
+++ b/app/controllers/katello/concerns/api/v2/bulk_hosts_extensions.rb
@@ -5,19 +5,19 @@ module Katello
 
       def find_bulk_hosts(permission, bulk_params, restrict_to = nil)
         #works on a structure of param_group bulk_params and transforms it into a list of systems
-        organization = find_organization
+        find_organization
         bulk_params[:included] ||= {}
         bulk_params[:excluded] ||= {}
         @hosts = []
 
         unless bulk_params[:included][:ids].blank?
           @hosts = ::Host::Managed.authorized(permission).where(:id => bulk_params[:included][:ids])
-          @hosts = @hosts.where(:organization_id => organization.id) if organization
+          @hosts = @hosts.where(:organization_id => @organization.id) if @organization
         end
 
         if bulk_params[:included][:search]
           search_hosts = ::Host::Managed.authorized(permission)
-          search_hosts = search_hosts.where(:organization_id => organization.id) if params[:organization_id]
+          search_hosts = search_hosts.where(:organization_id => @organization.id) if @organization
           search_hosts = search_hosts.search_for(bulk_params[:included][:search])
           if @hosts.any?
             @hosts = ::Host.where(id: @hosts).or(::Host.where(id: search_hosts))
@@ -38,7 +38,7 @@ module Katello
       end
 
       def find_organization
-        @organization ||= Organization.find(params[:organization_id])
+        @organization ||= Organization.find_by_id(params[:organization_id])
       end
     end
   end
--- a/app/controllers/katello/concerns/api/v2/bulk_hosts_extensions.rb
+++ b/app/controllers/katello/concerns/api/v2/bulk_hosts_extensions.rb
@@ -38,7 +38,7 @@ module Katello
       end
 
       def find_organization
-        Organization.find_by_id(params[:organization_id])
+        @organization ||= Organization.find(params[:organization_id])
       end
     end
   end
--- a/app/controllers/katello/concerns/api/v2/bulk_hosts_extensions.rb
+++ b/app/controllers/katello/concerns/api/v2/bulk_hosts_extensions.rb
@@ -17,7 +17,7 @@ module Katello
 
         if bulk_params[:included][:search]
           search_hosts = ::Host::Managed.authorized(permission)
-          search_hosts = search_hosts.where(:organization_id => organization_id) if params[:organization_id]
+          search_hosts = search_hosts.where(:organization_id => organization.id) if params[:organization_id]
           search_hosts = search_hosts.search_for(bulk_params[:included][:search])
           if @hosts.any?
             @hosts = ::Host.where(id: @hosts).or(::Host.where(id: search_hosts))
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-subscriptions-modal.controller.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-subscriptions-modal.controller.js
@@ -30,7 +30,7 @@ angular.module('Bastion.content-hosts').controller('ContentHostsBulkSubscription
             function getBulkSubscriptionParams() {
                 var bulkParams = hostIds;
                 bulkParams.subscriptions = SubscriptionsHelper.getSelectedSubscriptionAmounts($scope.table);
-                bulkParams.organization_id = CurrentOrganization; //eslint-disable-line camelcase
+                bulkParams['organization_id'] = CurrentOrganization; // eslint-disable-line camelcase
                 return bulkParams;
             }
 
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-subscriptions-modal.controller.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-subscriptions-modal.controller.js
@@ -30,7 +30,7 @@ angular.module('Bastion.content-hosts').controller('ContentHostsBulkSubscription
             function getBulkSubscriptionParams() {
                 var bulkParams = hostIds;
                 bulkParams.subscriptions = SubscriptionsHelper.getSelectedSubscriptionAmounts($scope.table);
-                bulkParams.organization_id = CurrentOrganization;
+                bulkParams.organization_id = CurrentOrganization; //eslint-disable-line camelcase
                 return bulkParams;
             }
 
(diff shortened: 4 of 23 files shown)


jturel:
@jeremylenz looking back through an email (check your inbox) around SCA, I think the desire here now is to completely hide the Subscriptions tab. Will be good to collaborate with UX on this (cc @RoHoover)

Rohoover:
Would it be desirable to show the tab, and just put messaging in the tab (no functionality) explaining SCA?

jturel:
> Would it be desirable to show the tab, and just put messaging in the tab (no functionality) explaining SCA?

I could go one way or the other, but mostly wanted to confirm that removing the tab was actually what we talked about in that mtg, and I'm pretty sure it was.

jeremylenz:
> I could go one way or the other, but mostly wanted to confirm that removing the tab was actually what we talked about in that mtg, and I'm pretty sure it was.

Roxanne and I discussed this on Friday and she couldn't remember the discussion specifically, but she and I agreed that removing the tab completely isn't ideal because the inconsistency may confuse some users, and there isn't anywhere else in the UI where we remove/hide elements for features that are disabled.  (We do for missing permissions, but that's a different domain.)

On the other hand, it's less work to just remove it entirely..  but that's not a great reason.

jeremylenz:
Updated to not show the Subscriptions tab when Simple Content Access is enabled:

content_host_sca

(Can always add more granular messaging later.)

jeremylenz:
Added messaging to content host details (info) page
content_host_details_sca

app/controllers/katello/api/v2/host_subscriptions_controller.rb
```diff
@@ -134,6 +134,10 @@ def remove_subscriptions
       param :quantity, :number, :desc => N_("Quantity of this subscriptions to add"), :required => true
     end
     def add_subscriptions
+      if @host.organization.simple_content_access?
+        fail ::Katello::HttpErrors::BadRequest, _("This host's organization has Simple Content Access enabled.  Attaching subscriptions is neither necessary nor allowed.")
```
jeremylenz, line 138: Thoughts on this wording?  I'm not sure I'm happy with it.

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/content-hosts.routes.js
```diff
@@ -134,7 +134,7 @@ angular.module('Bastion.content-hosts').config(['$stateProvider', function ($sta
         controller: 'ContentHostAddSubscriptionsController',
         templateUrl: 'content-hosts/details/views/content-host-add-subscriptions.html',
         ncyBreadcrumb: {
-            label: "{{ 'Add Subscriptipons' | translate }}",
+            label: "{{ 'Add Subscriptions' | translate }}",
```
jeremylenz, line 137: glad we finally get to fix this

jturel:
Looking great.

There are two other places where we can currently attach subscriptions:

1. Hosts -> Content Hosts -> Select Action -> Manage Subscriptions
2. Hosts -> Host Collections -> <collection> -> Subscription Management

To be consistent with the theme of "SCA by default" I think we should hide these links. These two modals share an API as far as I know (see the bulk hosts controller) so we should provide messaging like you already have for a single host

jturel:
> Added messaging to content host details (info) page
> content_host_details_sca

Seeing how this looks in practice - I don't think it makes sense to show 'Details and 'Auto-Attach Details' any more. Nothing useful shows there. It does 'undo' some of the work we did in 3.14 to show 'Not Applicable' but I think this is the right change in the 'SCA by default' world. Thoughts?

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/details/views/content-host-info.html
```diff
@@ -89,10 +89,19 @@ <h4 translate>Subscriptions</h4>
...
           {{ host.subscription_status_label }}
         </dd>
+        <dd ng-if="simpleContentAccessEnabled">
+          <!-- always display green icon for SCA -->
+          <i
+            ng-class="getHostStatusIcon(0)"
+            title = "{{ 'This organization has Simple Content Access enabled. Hosts can consume from all repositories in their Content View regardless of subscription status.' | translate }}"
+            >
+          </i>
+          {{ 'Simple Content Access' | translate }}
```
jturel, line 103: I'm tempted to handle this change in the subscription_status.rb to push it further back in the stack - and so it'll give a more meaningful result in the API, and hammer. See what it looks like now, with errata status included for comparison:

```
"errata_status": 1,
"errata_status_label": "Could not calculate errata status, ensure host is registered and the katello-host-tools package is installed",
"subscription_status": 5,
"subscription_status_label": "Disabled",
```

IOW I think we can change the label to reflect the messaging you've provided here.

app/controllers/katello/api/v2/host_subscriptions_controller.rb
jturel: For auto-attach we say `This host's organization is in Simple Content Access mode. Auto-attach is disabled` so I think we can be consistent and say `This host's organization is in Simple Content Access mode. Attaching subscriptions is disabled` Concise and explanatory enough, maybe?

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/details/views/content-host-info.html
```diff
@@ -89,10 +89,19 @@ <h4 translate>Subscriptions</h4>
...
         <dt translate>Subscription Status</dt>
-        <dd>
+        <dd ng-if="!simpleContentAccessEnabled">
           <i ng-class="getHostStatusIcon(host.subscription_global_status)"></i>
           {{ host.subscription_status_label }}
         </dd>
+        <dd ng-if="simpleContentAccessEnabled">
+          <!-- always display green icon for SCA -->
+          <i
+            ng-class="getHostStatusIcon(0)"
```
jturel, line 99: Instead of doing this you can update getHostStatusIcon to understand `5`  

https://github.com/Katello/katello/blob/bb672414d89e51eee88453ea73b91d1250594343/app/models/katello/subscription_status.rb#L3

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/details/views/content-host-info.html
jturel: Or maybe it can become just 'Simple Content Access' rather than the full message? Whichever feels right

jeremylenz: Is `Katello::Candlepin::Consumer::ENTITLEMENTS_DISABLED` used _only_ for SCA?

jturel: Yep!

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/details/views/content-host-info.html
jeremylenz: `getHostStatusIcon` is called with `host.subscription_global_status` which calls `to_global`, which only gives values from `::HostStatus::Global::ERROR`, `OK`, or `WARN` (0-2).  The '5' / `DISABLED` is returned not from `to_global`, but from `to_status`, which is then translated in `to_global` to `::HostStatus::Global::OK`.  All that being said, I don't think it makes sense to add a `5` value to `getHostStatusIcon`, since those values line up with the `to_global` return values.

jeremylenz: Updating to remove the `ng-if`, since the only thing we'll lose is the tooltip.  We get the correct green icon already thanks to the `DISABLED` status.

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/details/views/content-host-info.html
jeremylenz: Thanks, updating the label in `subscription_status.rb` to read "Simple Content Access" instead.

jturel:
> > These two modals share an API as far as I know (see the bulk hosts controller) so we should provide messaging like you already have for a single host
> 
> @jturel I removed the links. Not sure where I need to add the messaging; can you please clarify?

Cool. the two modals share an API in the 'bulk hosts controller' - so we should add messaging to that API too, I think (like you did for a single host subscription attach).

jeremylenz:
> Seeing how this looks in practice - I don't think it makes sense to show 'Details and 'Auto-Attach Details' any more

I agree. Removed :)

app/controllers/katello/api/v2/host_subscriptions_controller.rb
jeremylenz: @jturel looks good to me; updated.

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-subscriptions-modal.controller.js
```diff
@@ -30,6 +30,7 @@ angular.module('Bastion.content-hosts').controller('ContentHostsBulkSubscription
             function getBulkSubscriptionParams() {
                 var bulkParams = hostIds;
                 bulkParams.subscriptions = SubscriptionsHelper.getSelectedSubscriptionAmounts($scope.table);
+                bulkParams.organization_id = CurrentOrganization;
```
jeremylenz, line 33: Once I changed our API to return the error if Simple Content Access is enabled, this created two new requirements:

1. The incoming request must have an `organization_id` param (this was already required in the API docs, but was not actually enforced in the Rails controller)
2. The API needs to check the specified organization for Simple Content Access and respond accordingly

API requests sent by Bastion from this dialog were not even including `organization_id` as a param, despite the `:required => true` in the API docs.   I added it to `bulkParams` here to avoid the issue.

app/controllers/katello/api/v2/hosts_bulk_actions_controller.rb
```diff
@@ -295,6 +300,12 @@ def find_deletable_hosts
       find_bulk_hosts(:destroy_hosts, params)
     end
 
+    def validate_organization
```
jturel, line 303: This controller includes a concern which defines a method called `find_organization`

https://github.com/Katello/katello/blob/2fd5de8ae422881bf457c1d5d359f6a87836e803/app/controllers/katello/concerns/api/v2/bulk_hosts_extensions.rb#L40-L42

Perhaps you can update that method to use some memoization:

```
      def find_organization
        @organization ||= Organization.find_by_id(params[:organization_id])
      end
```

jeremylenz:
[test katello]

test/controllers/api/v2/concerns/bulk_hosts_extensions_test.rb
```diff
@@ -15,10 +15,10 @@ def initialize(params = {})
 
   class Api::V2::BulkHostsExtensionsTest < ActiveSupport::TestCase
     def models
-      @host1 = FactoryBot.create(:host)
-      @host2 = FactoryBot.create(:host)
-      @host3 = FactoryBot.create(:host)
       @organization = get_organization
+      @host1 = FactoryBot.create(:host, organization: @organization)
```
jturel, line 19: we have some fixtures for hosts and I think they have an organization already - try `hosts(:one)` for example and see if we can avoid creating these. if it's not a drop-in replacement then don't worry about it

app/controllers/katello/api/v2/hosts_bulk_actions_controller.rb
```diff
@@ -295,6 +301,10 @@ def find_deletable_hosts
...
+      fail HttpErrors::BadRequest, _("Organization ID is required") if @organization.blank?
```
jturel, line 305: `find_organization` will raise an error if it can't find the organization since you changed it to use `find` vs `find_by_id` which just returns nil if the org doesn't exist. This error wouldn't be raised as it is now. My recommendation is to change it back so that we aren't affecting the other APIs using the concern with this behavior change (even though we want the org_id to be required.), and perhaps file a separate issue that we can address next sprint or something.

app/controllers/katello/concerns/api/v2/bulk_hosts_extensions.rb
```diff
@@ -17,7 +17,7 @@ def find_bulk_hosts(permission, bulk_params, restrict_to = nil)
 
         if bulk_params[:included][:search]
           search_hosts = ::Host::Managed.authorized(permission)
-          search_hosts = search_hosts.where(:organization_id => organization_id) if params[:organization_id]
```
jturel, line 20: Curious about this. Was organization_id an undefined variable? are we missing test coverage here? We should add it :)

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/details/views/content-host-info.html
```diff
@@ -94,15 +94,14 @@ <h4 translate>Subscriptions</h4>
           {{ host.subscription_status_label }}
         </dd>
 
-        <dt translate>Details</dt>
+        <dt translate ng-if="!simpleContentAccessEnabled">Details</dt>
         <dd>
```
jturel, line 98: Does this <dd> also need the ng-if check?

app/controllers/katello/concerns/api/v2/bulk_hosts_extensions.rb
jeremylenz: Yeah!  Maybe I'm missing something major, but as far as I could tell `organization_id` is not defined and would just return `nil` always.  And it probably went unchallenged because none of the `bulk_hosts_extensions` tests call `find_bulk_hosts` with an `organization_id` param.  The controller actions just pass through params as-is, so my guess is the only time we'd run into this bug is if a user were making API requests without using the web UI or Hammer.

jturel: Dang. Did you run into an error when you changed the tests, or did you just happen to notice?

jeremylenz: I just ran into it when I was working on the tests.  It doesn't even try to evaluate that variable unless `params[:organization_id]` is truthy -- which is one of the things I had to change for the tests.

jturel: Cool. bonus bug fix! we must not be using the included[:search] path much if at all since it looks like this has been broken since 2016

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/details/views/content-host-info.html
jturel: ack. wonder if it was making it into the HTML...

jeremylenz: My guess is that since it's looping through `reason in host.subscription_facet_attributes.compliance_reasons`, the `dd` and `ul` were there but nothing was visible, since my host didn't happen to have any compliance reasons.

test/controllers/api/v2/concerns/bulk_hosts_extensions_test.rb
jeremylenz: Thanks, updating.

app/controllers/katello/api/v2/hosts_bulk_actions_controller.rb
jeremylenz: Thanks, didn't realize the error behavior was different.   Updating

app/controllers/katello/concerns/api/v2/bulk_hosts_extensions.rb
jeremylenz: Now that I'm [initializing](https://github.com/Katello/katello/blob/7e57aab509b71c837fec6012e7d2f61329c521f7/test/controllers/api/v2/concerns/bulk_hosts_extensions_test.rb#L32) the `TestController` to send `params[:organization_id]` with each test,
```ruby
@controller = TestController.new(organization_id: @organization.id)
```
I think that should take care of giving us test coverage on the problematic line.

jeremylenz: (I had added a test but realized it's not needed because ^^)

jturel: Gotcha. I think we're good to go w.r.t that!

jturel:
UI changes look good, and the functionality we've prevented in the UI still work in the non-SCA mode.

Hammer looks good too:

```
$ hammer host subscription attach --subscription-id=1 --host-id=8
                                                                                  
Failed to attach subscriptions to the host:
  This host's organization is in Simple Content Access mode. Attaching subscriptions is disabled.
```

And so does the bulk attach API:

```
$ curl -X PUT -u admin:changeme -H 'Content-Type: application/json' http://localhost:3000/api/v2/hosts/bulk/add_subscriptions?organization_id=9 -d '{"included":{"ids":[8]}, "subscriptions":[{"id": 359}]}'

{"displayMessage":"The specified organization is in Simple Content Access mode. Attaching subscriptions is disabled","errors":["The specified organization is in Simple Content Access mode. Attaching subscriptions is disabled"]}
```

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-subscriptions-modal.controller.js
```diff
@@ -30,6 +30,7 @@ angular.module('Bastion.content-hosts').controller('ContentHostsBulkSubscription
             function getBulkSubscriptionParams() {
                 var bulkParams = hostIds;
                 bulkParams.subscriptions = SubscriptionsHelper.getSelectedSubscriptionAmounts($scope.table);
+                bulkParams.organization_id = CurrentOrganization; // eslint-disable-line camelcase
```
jturel, line 33: Maybe we can just do `bulkParams['organization_id']` to work around the lint?

jeremylenz: :exploding_head:

jeremylenz:
[test katello]

PR #8713, opened 2020-05-14
Fixes #29832 - Added pagination to new CV page

johnpmitsch:
This moves the patternfly 4 table component to a new location and add pagination the CV table located at `labs/content_views`


johnpmitsch:
A few notes:

- The table has moved keeping it uncoupled from the content view components and getting it ready to move to foremanReact.
- Pagination is assumed in `TableWrapper` currently, we may want to make this optional in the future but I don't want to make assumptions just yet.
- It just occured to me that I should maybe just dynamically generate `webpack/scenes/ContentViews/__tests__/hundredContentViews.fixtures.json` with javascript and convert to JSON since they are all the same. This would make it easier to update if we drop or add any values to the API. I'll update that if there are no objections

johnpmitsch:
[test katello]

webpack/scenes/ContentViews/Table/ContentViewsTable.js
```diff
@@ -3,29 +3,29 @@ import PropTypes from 'prop-types';
...
+const ContentViewTable = ({ response, status, error }) => {
   const [table, setTable] = useState({ rows: [], columns: [] });
   const [rowMapping, setRowMapping] = useState({});
   const loading = status === STATUS.PENDING;
+  const { results, ...metadata } = response;
 
   useEffect(
     () => {
-      if (!loading && items && items.length > 0) {
+      if (!loading && results && results.length > 0) {
```
jeremylenz, line 19: I think the issue with the flashing `EmptyStateMessage` is happening because `status` changes to `RESOLVED` before `rows` is populated.  This is while `tableDataGenerator` is running.

The second problem is that the `setRowMapping` in the `useEffect` in `ContentViewsTable.js` actually causes the same `useEffect` to run again, because `JSON.stringify(rowMapping)` is in the dependency array.  This means `tableDataGenerator` runs a second time, doubling the time on screen for the `EmptyStateMessage`.

I was playing around (for way too long) and I was able to prevent the `EmptyStateMessage` by comparing the size of `rowMapping` with `results.length`:

```js
const resultsLength = 0 || results && results.length
  const rowMappingLength = 0 || Object.keys(rowMapping) && Object.keys(rowMapping).length
  const displaySpinner = (status === STATUS.PENDING) || resultsLength !== rowMappingLength
  useEffect(
    () => {
      if (status = STATUS.PENDING && results && results.length > 0) {
...
```

I forced the status to `STATUS.PENDING` when rendering the table:
```js
return (
    <TableWrapper
      {...{
        rows,
        error,
        status: displaySpinner ? STATUS.PENDING : status,
...
```

This gets rid of the EmptyStateMessage flash.

There's probably a more elegant way to do it, but you get the idea.

To address the second problem, I think it would improve load time if you [memoize](https://reactjs.org/docs/hooks-reference.html#usememo) `tableDataGenerator`, since it's running multiple times with the same results data.

webpack/components/Table/TableWrapper.js
```diff
@@ -0,0 +1,108 @@
...
+import EmptyStateMessage from './EmptyStateMessage';
+import Loading from './Loading';
+
+/* Patternfly 4 table wrapper */
+const TableWrapper = ({
+  status, cells, rows, error, emptyTitle, emptyBody, metadata, fetchItems, ...extraTableProps
+}) => {
+  const dispatch = useDispatch();
+  const { foremanPerPage = 20 } = useForemanSettings();
+  // setting pagination to local state so it doesn't disappear when page reloads
```
jeremylenz, line 24: Seems like two sources of truth, `metadata` and `useForemanSettings()`.  Is there a better way?

Looking in `ContentViewsTable.js`, I see `metadata` is part of `response`, which Foreman is sending us anyway.  Isn't Foreman sending us the same pagination settings in the API response that we get thru React context and `useForemanSettings`?  If the data is the same, can this be simplified to use a single source of truth?

webpack/components/Table/TableWrapper.js
```diff
@@ -0,0 +1,108 @@
...
+  const [total, setTotal] = useState(0);
+
+  useEffect(() => {
+    const { total: newTotal, page: newPage, per_page: newPerPage } = metadata;
+    if (newTotal) setTotal(parseInt(newTotal, 10));
+    if (newPage) setPage(parseInt(newPage, 10));
+    if (newPerPage) setPerPage(parseInt(newPerPage, 10));
+  }, [metadata]);
+
+  const buildMainTable = () => {
```
jeremylenz, line 36: This is a function that returns JSX, therefore it's a React component and should be refactored as such (name should be TitleCase; render as regular JSX instead of inside brackets; etc.)  I've never liked the pattern of building the JSX to return within a React component by using helper functions that return JSX.  Instead let's use the amazing composability that JSX gives to us natively.  (No need to extract to a separate file, though.. I'm fine with multiple small components living in the same file if it makes sense.)

webpack/components/Table/TableWrapper.js
```diff
@@ -0,0 +1,108 @@
...
+    if (status === STATUS.PENDING) return (<Loading />);
+    // Can we display the error message?
+    if (status === STATUS.ERROR) return (<EmptyStateMessage error={error} />);
+    // Can we prevent flash of empty row message while rows are loading with data?
```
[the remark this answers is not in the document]
jeremylenz: updated, see other comment for possible solution :)

webpack/components/Table/TableWrapper.js
```diff
@@ -0,0 +1,108 @@
...
+/* Patternfly 4 table wrapper */
+const TableWrapper = ({
+  status, cells, rows, error, emptyTitle, emptyBody, metadata, fetchItems, ...extraTableProps
+}) => {
+  const dispatch = useDispatch();
+  const { foremanPerPage = 20 } = useForemanSettings();
+  // setting pagination to local state so it doesn't disappear when page reloads
+  const [perPage, setPerPage] = useState(foremanPerPage);
+  const [page, setPage] = useState(1);
+  const [total, setTotal] = useState(0);
```
jeremylenz, line 27: Could the pagination-related stuff be extracted into a `usePagination()` hook?  It would be awesome if we could reuse that across different tables, etc.

webpack/scenes/ContentViews/ContentViewSelectors.js
```diff
@@ -5,8 +5,7 @@ import {
 } from 'foremanReact/redux/API/APISelectors';
 import CONTENT_VIEWS_KEY from './ContentViewsConstants';
```
jeremylenz, line 6: Sorry I didn't catch this in the previous PR.. The value of the key should be uppercase snake case (CONTENT_VIEWS) to be consistent with the rest of our Redux actions:
redux

webpack/scenes/ContentViews/__tests__/contentViewPage.test.js
```diff
@@ -97,3 +99,45 @@ test('Can handle unpublished Content Views', async () => {
...
+  const { queryByText, getByLabelText } = renderWithApiRedux(<ContentViewsPage />, renderOptions);
+
+  // Wait for first paginated page to load and assert only the first page of results are present
+  await waitFor(() => {
+    expect(queryByText(results[0].name)).toBeInTheDocument();
+    expect(queryByText(results[19].name)).toBeInTheDocument();
+    expect(queryByText(results[21].name)).not.toBeInTheDocument();
+  });
+
+  // Label comes from patternfly, if this test fails, check if patternfly updated the label.
```
jeremylenz, line 130: Nice.  This will be useful down the road :smile:

webpack/scenes/ContentViews/__tests__/hundredContentViews.fixtures.json
```diff
@@ -0,0 +1,1514 @@
+{
```
jeremylenz, line 1: The sheer size of this seems wasteful.  Would it work to create this data programmatically?

jeremylenz: update: don't mind me, I just read your update above.. seems like we're on the same page :yum:

webpack/scenes/ContentViews/Table/ContentViewsTable.js
jeremylenz: Just realized my solution breaks when you change the per-page setting.  :shrug:

webpack/components/Table/TableWrapper.js
johnpmitsch: There is a default per page setting in Foreman. You can play around with this by going to `Settings > General` tab and changing `Entries per page` to 30 or something like that and you will see 30 load at first. This actually isn't working correctly in the angular Katello pages from what I can tell :/ but it is working on the Subscriptions page so it seems to be accounted for in the pf3 pages.

I'm updating this so the logic is:
- Get per page from Foreman settings first with 20 as a fallback and use 0 as a default for total and 1 for page/
- Update page and perPage, and total (though total shouldn't change here) when pagination is updated
- Updating page, perPage, and total when metadata comes back in API response. The API sends back the page and per-page of what was sent so this will stay the same, but we need the total to be set by the response.

I have this a separate function inside the TableWrapper right now and will push up soon. I'm not sure I see the use case of making it a separate hook, the data is pretty specific to the pf4 table and it looks like the pf3 pagination is structured differently. Is there a particular component you have in mind?

I also updated to use the pagination options from Foreman, so when you set a default pagination setting, it is now an option. This is the same as how the pf3 table works. For example, 30 is the default per page setting and is now listed as an option in the dropdown:
h

webpack/components/Table/TableWrapper.js
johnpmitsch: I agree in principle, but I went with this way because I'm not crazy about the two alternatives:
- Put inside JSX - The logic for this many conditions is harder to read (IMO) when all in JSX and needs to be escaped
- Move to separate component - having to do more prop drilling by repeating the props and passing them once again

I'm not opposed to either of these as things grow, but I just don't see enough issue with the helper function to go with an alternative as things are right now. I'll definitely update the name to TitleCase as you suggested and maybe we can discuss the structure more as things progress?

jeremylenz:
```js
{MainTable()}
```
can become 
```js
<MainTable />
```
and it just works.  Since the component is defined from within the same scope, it has access to `status`, `error`, `emptyTitle`, and `emptyBody` without explicitly passing them.  Yay JavaScript!

webpack/scenes/ContentViews/__tests__/basicContentViews.fixtures.js
```diff
@@ -0,0 +1,31 @@
...
+    error: null,
+    search: null,
+    sort: {
+      by: 'name',
+      order: 'asc',
+    },
+    results: [],
+  };
+
+  Array(amount).fill().map((_, i) => response.results.push(buildContentView(i)));
```
jeremylenz, line 26: Should be a `forEach` since the result is thrown away.
Actually this can be simplified a bit with 
```js
[...Array(amount).keys()]
```

webpack/scenes/ContentViews/Table/ContentViewsTable.js
```diff
@@ -58,37 +58,48 @@ const ContentViewTable = ({
         return { ...prev, [contentViewId]: updatedMap };
       });
     } else {
-      // remove the row completely by assigning it to a throwaway variable
-      // eslint-disable-next-line camelcase, no-unused-vars
-      const { [contentViewId]: _throwaway, ...newMap } = rowMapping;
-      setRowMapping(newMap);
+      setRowMapping(prev => ({ ...prev, [contentViewId]: {} }));
```
jeremylenz, line 61: :+1: :+1:

webpack/components/Table/TableWrapper.js
```diff
@@ -0,0 +1,111 @@
...
+      <Pagination
+        itemCount={total}
+        page={page}
+        perPage={perPage}
+        onSetPage={(_evt, updated) => onPaginationUpdate({ page: updated })}
+        onPerPageSelect={(_evt, updated) => onPaginationUpdate({ per_page: updated })}
+        perPageOptions={usePaginationOptions().map(p => ({ title: p.toString(), value: p }))}
+        variant="top"
+      />
+      {MainTable()}
```
jeremylenz, line 75: See response to the comment from the earlier review
```suggestion
      <MainTable />
```

webpack/__mocks__/foremanReact/Root/Context/ForemanContext.js
```diff
@@ -0,0 +1,3 @@
+// eslint-disable-next-line import/prefer-default-export
+export const useForemanSettings = () => ({ perPage: 20 });
+export const usePaginationOptions = () => [5, 10, 20, 50];
```
jeremylenz, line 3: Why did you have to mock these?

webpack/components/Table/TableWrapper.js
jeremylenz: Thanks for the explanation.  Makes sense not to have a separate hook if there's nowhere else you would use it.

jeremylenz:
If I change to a new page, and then go back to the old page, it still reloads the data from the API.  Can it display the data from Redux instead, if it already has it?

jeremylenz:
Also, I'm still seeing the first issue above where I switch pages and it displays Loading spinner -> Old page -> new page.

johnpmitsch:
> Also, I'm still seeing the first issue above where I switch pages and it displays Loading spinner -> Old page -> new page.

Ya, I meant to mention that, that one I have to figure out still. I'm not sure why that is happening

webpack/__mocks__/foremanReact/Root/Context/ForemanContext.js
johnpmitsch: They seem to eventually [dip into ruby code](https://github.com/theforeman/foreman/blob/17d7be599c88ab953d10a396eccdc93d30a69e88/app/helpers/application_helper.rb#L524) and don't seem to cooperate in a testing environment, at least in Katello. I just figured its a small thing that is easy to mock.

johnpmitsch: This is the error if you are curious, it is a bit obscure, but a little digging showed me its these methods that are causing it, probably due to the ruby portion not being present. If you want to recreate you can just rename this mock file:

```
  ● Can handle pagination

    Error: Uncaught [Error: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
    1. You might have mismatching versions of React and the renderer (such as React DOM)
    2. You might be breaking the Rules of Hooks
    3. You might have more than one copy of React in the same app
    See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.]

      3 | // output and traceback for actual error.
      4 | global.console.error = (error) => {
    > 5 |   throw new Error(error);
        |         ^
      6 | };
      7 | 

      at reportException (node_modules/jsdom/lib/jsdom/living/helpers/runtime-script-errors.js:66:24)
      at invokeEventListeners (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:209:9)
      at HTMLUnknownElementImpl._dispatch (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:119:9)
      at HTMLUnknownElementImpl.dispatchEvent (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:82:17)
      at HTMLUnknownElementImpl.dispatchEvent (node_modules/jsdom/lib/jsdom/living/nodes/HTMLElement-impl.js:30:27)
      at HTMLUnknownElement.dispatchEvent (node_modules/jsdom/lib/jsdom/living/generated/EventTarget.js:157:21)
      at Object.invokeGuardedCallbackDev (node_modules/react-dom/cjs/react-dom.development.js:237:16)
      at invokeGuardedCallback (node_modules/react-dom/cjs/react-dom.development.js:292:31)
      at beginWork$1 (node_modules/react-dom/cjs/react-dom.development.js:23203:7)
      at performUnitOfWork (node_modules/react-dom/cjs/react-dom.development.js:22157:12)
      at CustomConsole.Object.<anonymous>.global.console.error (webpack/global_test_setup.js:5:9)
      at VirtualConsole.<anonymous> (node_modules/jsdom/lib/jsdom/virtual-console.js:29:45)
      at reportException (node_modules/jsdom/lib/jsdom/living/helpers/runtime-script-errors.js:70:28)
      at invokeEventListeners (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:209:9)
      at HTMLUnknownElementImpl._dispatch (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:119:9)
      at HTMLUnknownElementImpl.dispatchEvent (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:82:17)
      at HTMLUnknownElementImpl.dispatchEvent (node_modules/jsdom/lib/jsdom/living/nodes/HTMLElement-impl.js:30:27)
      at HTMLUnknownElement.dispatchEvent (node_modules/jsdom/lib/jsdom/living/generated/EventTarget.js:157:21)
      at Object.invokeGuardedCallbackDev (node_modules/react-dom/cjs/react-dom.development.js:237:16)
      at invokeGuardedCallback (node_modules/react-dom/cjs/react-dom.development.js:292:31)
      at beginWork$1 (node_modules/react-dom/cjs/react-dom.development.js:23203:7)
      at performUnitOfWork (node_modules/react-dom/cjs/react-dom.development.js:22157:12)
      at workLoopSync (node_modules/react-dom/cjs/react-dom.development.js:22130:22)
      at performSyncWorkOnRoot (node_modules/react-dom/cjs/react-dom.development.js:21756:9)
      at scheduleUpdateOnFiber (node_modules/react-dom/cjs/react-dom.development.js:21188:7)
      at updateContainer (node_modules/react-dom/cjs/react-dom.development.js:24373:3)
      at node_modules/react-dom/cjs/react-dom.development.js:24758:7
      at unbatchedUpdates (node_modules/react-dom/cjs/react-dom.development.js:21903:12)
      at legacyRenderSubtreeIntoContainer (node_modules/react-dom/cjs/react-dom.development.js:24757:5)
      at Object.render (node_modules/react-dom/cjs/react-dom.development.js:24840:10)
      at node_modules/@testing-library/react/dist/pure.js:86:25
      at batchedUpdates$1 (node_modules/react-dom/cjs/react-dom.development.js:21856:12)
      at act (node_modules/react-dom/cjs/react-dom-test-utils.development.js:929:14)
      at render (node_modules/@testing-library/react/dist/pure.js:82:26)
      at renderWithApiRedux (webpack/test-utils/react-testing-lib-wrapper.js:36:15)
      at _callee5$ (webpack/scenes/ContentViews/__tests__/contentViewPage.test.js:121:43)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:45:40)
      at Generator.invoke [as _invoke] (node_modules/regenerator-runtime/runtime.js:274:22)
      at Generator.prototype.<computed> [as next] (node_modules/regenerator-runtime/runtime.js:97:21)
      at asyncGeneratorStep (webpack/scenes/ContentViews/__tests__/contentViewPage.test.js:27:103)
      at _next (webpack/scenes/ContentViews/__tests__/contentViewPage.test.js:29:194)
      at webpack/scenes/ContentViews/__tests__/contentViewPage.test.js:29:364
      at Object.<anonymous> (webpack/scenes/ContentViews/__tests__/contentViewPage.test.js:29:97)

  ● Can handle pagination

    Error: Uncaught [Error: The above error occurred in the <TableWrapper> component:
        in TableWrapper (created by ContentViewTable)
        in ContentViewTable (created by ContentViewsPage)
        in ContentViewsPage (created by Context.Consumer)
        in withRouter(ContentViewsPage)
        in Router (created by MemoryRouter)
        in MemoryRouter
        in Provider

    Consider adding an error boundary to your tree to customize error handling behavior.
    Visit https://fb.me/react-error-boundaries to learn more about error boundaries.]

      3 | // output and traceback for actual error.
      4 | global.console.error = (error) => {
    > 5 |   throw new Error(error);
        |         ^
      6 | };
      7 | 

      at reportException (node_modules/jsdom/lib/jsdom/living/helpers/runtime-script-errors.js:66:24)
      at Timeout.callback [as _onTimeout] (node_modules/jsdom/lib/jsdom/browser/Window.js:680:7)
      at CustomConsole.Object.<anonymous>.global.console.error (webpack/global_test_setup.js:5:9)
      at VirtualConsole.<anonymous> (node_modules/jsdom/lib/jsdom/virtual-console.js:29:45)
      at reportException (node_modules/jsdom/lib/jsdom/living/helpers/runtime-script-errors.js:70:28)
      at Timeout.callback [as _onTimeout] (node_modules/jsdom/lib/jsdom/browser/Window.js:680:7)
```

webpack/components/Table/TableWrapper.js
johnpmitsch: Thanks, I updated. I figured you meant move it outside the parent function so was confused, but it makes more sense now :)

johnpmitsch:
@jeremylenz I updated, I actually found an issue with the pagination and the row mapping that affected the expansion page

So when you went to the first page and clicked on a column for a CV with the id of 2, the row mapping updated to something like:
```js
const rowMap = {
...
2: { rowIndex: 5, expandedColumn: true } 
...
};
```

Then you would go to the next page, but all the row mappings from the previous page would stay. So you would click on that same row and the mapping would update to:
```js
const rowMap = {
...
2: { rowIndex: 5, expandedColumn: true } 
...
12: { rowIndex: 5, exandedColumn: false }
};
```

Since they both had the same rowIndex, the column didn't update since the helper function found the content view id by searching row indexes.

I updated the mapping to be $ROWINDEX: { $CONTENTVIEWID, $EXPANDEDCOLUMN }

so the above would be
```js
const rowMap = {
...
5: { id: 2, expandedColumn: true } 
...
};
```
on the first page and
```js
const rowMap = {
...
5: { id: 12, expandedColumn: true } 
...
};
```
on the second, since the row index will be overwritten. This also prevents the data structure from growing too large, if someone has 1000s of CVs and is cycling through the pages, the rowMapping will just get as large as whatever the largest page size is. Probably not a huge performance difference, but worth considering!

> Ya, I meant to mention that, that one I have to figure out still. I'm not sure why that is happening

Mind if we do a follow up for this pagination issue? Its not super clear since its mostly handled on the pf side.

jeremylenz:
@johnpmitsch I think a follow-up is fine to improve the content view caching / Redux stuff, but can you see if you can at least fix the [rendering](https://github.com/Katello/katello/pull/8713#issuecomment-631444937) issue?   It should just be a matter of a little more tinkering with `tableStatus`.. if it gets too involved we can always push to the next PR, but I think it's worth a shot?

johnpmitsch:
@jeremylenz Updated the table status to only return status if:
- results are undefined (either an error or pending request)
- results are zero (no CVs present)
- all id's from results are accounted for in rowMapping (results are returned and table has been rebuilt)

The caching of pages seems more involved as we discussed off-thread, I filed an issue here and added it to the tracker: https://projects.theforeman.org/issues/29904

webpack/scenes/ContentViews/Table/ContentViewsTable.js
```diff
@@ -48,47 +42,64 @@ const ContentViewTable = ({
...
 
+  // Prevents flash of "No Content" before rows are loaded
+  const tableStatus = () => {
+    if (typeof results === 'undefined') return status; // will handle errored state
+    const resultsLength = results && results.length;
+    const rowMappingLength = Object.keys(rowMapping) && Object.keys(rowMapping).length;
+    if (resultsLength === 0) return status; // handles no CVs present
+    if (rowMappingLength > 0 && resultsLength > 0) {
+      const rowMappingIds = Object.values(rowMapping).map(map => map.id);
+      const resultsIds = Array.from(results.map(result => result.id));
```
johnpmitsch, line 68: If you are wondering why I need `Array.from`, the results object is an `Immutable` object

johnpmitsch:
ugg.. I just realized when closing an expansion pane the Loading icon shows again :disappointed:, will look into it

johnpmitsch:
> ugg.. I just realized when closing an expansion pane the Loading icon shows again , will look into it

This we because of deleting the whole object in the row mapping when an expansion pane is closed, so the id's didn't match between results and rowMapping id's when checking the table status. I kept the `id` in now and just checking for `expandedColumn` to see if the row mapping needs to be updated.

johnpmitsch:
> I'm still seeing the old rows flash before it renders the new ones. May be harder to solve than we thought.. I'll leave it up to you how to proceed.

@jeremylenz I'm not seeing that, can you confirm the latest changes are there? I even tried with CPU throttling and couldn't reproduce. Here is what it looks like for me:

Peek 2020-05-21 13-10

webpack/scenes/ContentViews/Table/tableDataGenerator.js
```diff
@@ -100,34 +100,33 @@ const buildDetailDropdowns = (id, rowIndex, openColumn) => {
...
+  const updatedRowMap = { ...rowMapping };
   const contentViews = results || [];
   const columns = buildColumns();
   const rows = [];
 
-  contentViews.forEach((contentView, rowIndex) => {
+  contentViews.forEach((contentView) => {
     const { id } = contentView;
+    const rowIndex = rows.length;
     // hasOwnProperty syntax because of https://eslint.org/docs/rules/no-prototype-builtins
```
jeremylenz, line 117: Obsolete comment

PR #8714, opened 2020-05-15
Fixes #29838 - add Red Hat Subscriptions consumed macro

snagoor:


chris1984:
@snagoor looks like there is a rubocop failure, if you want to fix and push up, we can start testing.

snagoor:
> @snagoor looks like there is a rubocop failure, if you want to fix and push up, we can start testing.

@chris1984, thank you. Can you please hold on to this one. Seems that the `consumed` value shown here is for the entire Subscription Pool, not the consumed value of the individual host. 

So, I'm working on  getting the `quantity_consumed`  https://github.com/Katello/katello/blob/abbff485b07e94f8dd1dc1caeff4fc23f4e1b356/app/views/katello/api/v2/host_subscriptions/index.json.rabl#L7 to be shown here, which is the accurate count for an individual host.

jturel:
@snagoor looks like this needs a rebase now :)

app/lib/katello/concerns/base_template_scope_extensions.rb
```diff
@@ -28,6 +29,10 @@ def host_subscriptions_names(host)
         host.subscriptions.map(&:name)
       end
 
+      def host_redhat_subcriptions_consumed(host)
+        host.pools.redhat.pluck(:consumed).sum
```
tbrisker, line 33: You can 
```suggestion
        host.pools.redhat.sum(:consumed)
```
instead and let the DB do the calculation instead of pulling all rh pools to memory and summing it.

jturel:
Closing this out in favor of https://github.com/Katello/katello/pull/9091

PR #8716, opened 2020-05-15
Fixes #29847 - Cannot enable dep solving or auto publish

ianballou:
Dep solving and auto publish were broken by Rails 6 it appears.  Adding these attributes to `view_params` fixes the issue.


mmoll:
Could there get a test created, that would have caught that issue? :bomb:

ianballou:
@mmoll just pushed a test that catches the issue

PR #8717, opened 2020-05-15
Fixes #29554 - Katello Applicability host processing queue

ianballou:
This PR allows multiple host applicabilities to be calculated per Dynflow task.

I recommend testing with Pulp 3 yum content.

We want to make sure firstly that applicability still works as expected with this PR.

We also need to test that the host queue works as expected.  One idea could be to set `applicability_batch_size` in the settings to a smaller number, e.g. 5, and then calculate the applicabilities for 15 hosts.  Ideally, 3 tasks should pop up.  The first 3 applicability events will consume 5 hosts each, and the rest will do nothing.


ianballou:
I noticed that, even if I spawn a bunch of generate applicability events, only one is run if they all have the same `object_id` passed to them.  This could potentially leave hosts in the queue if the same host needs to be calculated multiple times.  At the moment I am passing the `host_id` to the event because it needed `object_id` to not be `nil`.

ianballou:
To fix my issue above, I will change the code to allow one event to spawn multiple generate applicability tasks, rather than one applicability task per event.

app/lib/actions/katello/host/generate_applicability.rb
```diff
@@ -19,7 +19,14 @@ def plan(hosts, use_queue = true)
...
 
         def finalize
-          unless SETTINGS[:katello][:katello_applicability]
+          if SETTINGS[:katello][:katello_applicability]
+            input[:host_ids].each do |host_id|
+              ::Katello::ApplicableHostQueue.push_host(host_id)
+            end
+            input[:host_ids].count.times do
+              ::Katello::EventQueue.push_event(::Katello::Events::GenerateHostApplicability::EVENT_TYPE, 0)
+            end
```
jlsherrill, line 28: i think we can likely just push once here, right?

app/models/setting/content.rb
```diff
@@ -145,7 +145,9 @@ def self.default_settings
                   "as long as the registering DMI UUID is not used by another host."),
                true, N_('Host Profile Assume')),
       self.set('host_tasks_workers_pool_size', N_("Amount of workers in the pool to handle the execution of host-related tasks. When set to 0, the default queue will be used instead. Restart of the dynflowd/foreman-tasks service is required."),
-               5, N_('Host Tasks Workers Pool Size'))
+               5, N_('Host Tasks Workers Pool Size')),
+      self.set('applicability_batch_size', N_("Number of host applicability calculations to process per task."),
+               50, N_('Applicability Batch Size'))
```
jlsherrill, line 150: i suspect we'll want something larger asa a default than 50, but let me do some testing!

app/services/katello/applicable_host_queue.rb
```diff
@@ -0,0 +1,21 @@
...
+    end
+
+    def self.push_host(host_id)
+      HostQueueElement.create!({ host_id: host_id })
+    end
+
+    def self.pop_hosts(amount = self.batch_size)
+      queue = HostQueueElement.order(:created_at => 'asc')[0...amount]
+      HostQueueElement.delete(queue)
+      queue
```
jlsherrill, line 18: i think this won't de-duplicate host queue element items, if i'm reading it right.  

Also, in AR .limit(amount) would be a lot faster than using array methods [0..amount] (i think)

ianballou: Do you think a good strategy for removing the duplicates would be to retrieve the popped hosts, delete them from the DB, and then return the batched host list with a `.uniq`?

jlsherrill: I think so, i'd probably do something like:
HostQueueElement.group(:host_id).select("host_id, MIN(created_at) as created_at").limit(amount)
and then you can just delete_all by host_id and pass the results out

ianballou:
@jlsherrill I've fixed everything up

app/lib/actions/katello/applicability/repository/regenerate.rb
```diff
@@ -12,8 +12,9 @@ class Regenerate < Actions::EntryAction
 
           def run
             ::Katello::Repository.find(input[:repo_id]).content_facets.each do |facet|
```
jlsherrill, line 14: I just realized, i think this will skip host's that are attached to a Content View instance of the repo.  We probably need to look up content_facets associated with any clone of repo_id  as well

jlsherrill: and to make clear, i think this was an issue on the initial applicabiliyt pr

ianballou: Oh yep I can see that happening. I'll fix

app/models/setting/content.rb
jlsherrill: actually after testing, when doing ~1000 profiles, it really didn't make any difference between 30, 50, or 100.  So i think a default of 50 is fine

app/lib/actions/katello/applicability/hosts/bulk_generate.rb
```diff
@@ -2,17 +2,22 @@ module Actions
...
-              ::Katello::EventQueue.push_event(::Katello::Events::GenerateHostApplicability::EVENT_TYPE, host_id)
+              content_facet = ::Host.find(host_id).content_facet
+              content_facet.calculate_and_import_applicability
             end
           end
 
+          def resource_locks
+            :link
+          end
+
```
jlsherrill, line 20: I think we want to add this method to this action:
```
        def queue
          ::Katello::HOST_TASKS_QUEUE
        end
```

so that these run in the 'host queue'

app/models/katello/host_queue_element.rb
```diff
@@ -0,0 +1,5 @@
+module Katello
+  class HostQueueElement < Katello::Model
+    serialize :metadata, Hash
```
jlsherrill, line 3: what is this used for?

jlsherrill: specifically the ` serialize :metadata, Hash`

ianballou: That was something I thought I needed but do not. Gone now!

ianballou:
@jlsherrill I've addressed your comments.

app/lib/actions/katello/applicability/repository/regenerate.rb
```diff
@@ -11,9 +11,18 @@ class Regenerate < Actions::EntryAction
           end
 
           def run
-            ::Katello::Repository.find(input[:repo_id]).content_facets.each do |facet|
-              ::Katello::EventQueue.push_event(::Katello::Events::GenerateHostApplicability::EVENT_TYPE, facet.host.id)
+            content_facets = []
+            repo = ::Katello::Repository.find(input[:repo_id])
+            content_facets << repo.content_facets
+            content_facets << repo.clones.collect { |clone| clone.content_facets }
```
jlsherrill, line 17: It would be a lot more efficient to look up all the clones and then query the content facets for all of them (so its not a host query per repo).  I looked and there's actually already a method for that:
Repository#hosts_with_applicability

jlsherrill:
one comment, and a rubocop issue.  Otherwise looks good and works well!

app/lib/actions/katello/applicability/repository/regenerate.rb
```diff
@@ -11,9 +11,14 @@ class Regenerate < Actions::EntryAction
           end
 
           def run
-            ::Katello::Repository.find(input[:repo_id]).content_facets.each do |facet|
-              ::Katello::EventQueue.push_event(::Katello::Events::GenerateHostApplicability::EVENT_TYPE, facet.host.id)
+            hosts = ::Katello::Repository.find(input[:repo_id]).hosts_with_applicability
```
jlsherrill, line 14: This is kind of a nitpick, but since there's a test failure (that might be an ordering issue), might as well update this too: 
This could return 10s of thousands of hosts, so i'd probably optimize here to only return ids:
```host_ids = ::Katello::Repository.find(input[:repo_id]).hosts_with_applicability.pluck(:id)```
and not require the full host to be loaded

File: test/actions/katello/applicability/repository_regenerate_applicability_test.rb [hunk 9]
```diff
@@ -0,0 +1,30 @@
...
+    include Support::Actions::Fixtures
+    include FactoryBot::Syntax::Methods
+
+    before :all do
+      User.current = users(:admin)
+      @host = FactoryBot.build(:host, :with_content, :with_subscription, :content_view => katello_content_views(:library_dev_view),
+                               :lifecycle_environment => katello_environments(:library))
+      @host.save!
+      @repo = katello_repositories(:fedora_17_x86_64_duplicate)
+      SETTINGS[:katello][:katello_applicability] = true
```
jlsherrill, line 15: I think you need an  after(:all) here to set it back to false

PR #8721, opened 2020-05-19
[CP] Release 3.16.0.rc1 + cherry picks

ianballou:


jturel:
Looks good! There are a *lot* of issues under 'Other' in the changelog which means we didn't do a good job of updating the category during triage. Can you look through those and re-categorize whichever ones can fit into a better category? It's not an exact science, and it's ok if we have a few left under Other, but not 20+

jturel:
tip: use [CP] in the PR name when opening PRs that contain cherry-picks. Otherwise the bot will add this PR in the redmines of the individual issues which we don't want

jturel:
https://projects.theforeman.org/issues/27501 was in the list of cherry-picks but I don't see it here

ianballou:
@jturel I added 27501 to the ignore list since it is already on the branch

ianballou:
@jturel the new changelog has better categories now. Should be all set!

jturel:
note: we don't squash on release branches so we can revert things if necessary, etc

PR #8722, opened 2020-05-20
Fixes #29905 - Remove duplicate npm dependencies and add github action

johnpmitsch:
Related community discussions:
https://community.theforeman.org/t/rfc-dont-duplicate-npm-packages-in-katello-and-plugins-for-react-testing/18725
https://community.theforeman.org/t/running-plugin-tests-with-github-actions/18342

This PR:
- Removes dependencies that are already in Foreman
- Configures jest to use the dependencies from Foreman
- Configures eslint to use the dependencies from Foreman
- Adds a GH action for React testing and linting

My proposal is:
- Merge this PR and the github action
- ~When we are comfortable that the github action is working as expected, remove the react tests from Jenkins and make this github action blocking~
- Since Jenkins would need updates to install foreman's npm packages with this change, what I think would be best is to remove the react tests from Jenkins in favor of the GH action once merged.

Related PRs
- Remove job from Jenkins: https://github.com/theforeman/foreman-infra/pull/1367 - this will have to be merged right after this change is merged to Katello
- Adds the "find Foreman" logic as an npm package in foreman-js so this logic can be done in other plugins: https://github.com/theforeman/foreman-js/pull/138 - This won't block this PR but we should switch to using the package when it is available.


johnpmitsch:
I still have to figure out linting since we are using some configuration from the foreman metapackages, which are removed in this PR.

.github/workflows/react_tests.yml
```diff
@@ -0,0 +1,50 @@
+# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
+
+name: React Tests
+
+on:
+  push:
```
jeremylenz, line 7: Any particular reason for running the tests on `push` to master as well as pull requests?

.github/workflows/react_tests.yml
```diff
@@ -0,0 +1,50 @@
...
+    - name: Use Node.js
+      uses: actions/setup-node@v1
+      with:
+        node-version: '12.x'
+    - uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: 2.6
+    # We could update the postinstall action for foreman to look for an environment variable for plugin webpack dirs
+    # before kicking off the ruby script to find them, setting that would eliminate the ruby dep and running `npm i` in Katello.
+    - name: Install bundler (needed to find Foreman plugins)
```
jeremylenz, line 34: I thought I remembered reading somewhere bundler is preinstalled on Github action runners?

.github/workflows/react_tests.yml
```diff
@@ -0,0 +1,50 @@
...
+        path: foreman # checkout Foreman and Katello as siblings, same as dev env
+        ref: develop
+    - name: Install Foreman dependencies
+      run:  npm i
+      working-directory: ./foreman
+    - name: Install Katello dependencies
+      run:  npm i
+    - name: Run Katello tests
+      run: npm test
+    - name: Run Katello linting
```
jeremylenz, line 49: I still think it would be _super_ valuable if this were a separate job, so you could see at a glance if the failure is in linting or tests.

webpack/scenes/Subscriptions/Manifest/__tests__/__snapshots__/ManageManifestModal.test.js.snap
```diff
@@ -5,7 +5,7 @@ exports[`manage manifest modal should render 1`] = `
   id="manageManifestModal"
   title="Manage Manifest"
 >
-  <Uncontrolled(Tabs)
+  <ForwardRef
```
jeremylenz, line 8: Patternfly strikes again :smile:

webpack/test-utils/findForeman.js
```diff
@@ -0,0 +1,29 @@
+const path = require('path');
+const fs = require('fs');
+
+const getForemanLocation = () => {
```
jeremylenz, line 4: Consider dropping `get` from the names `getForemanLocation` and `getForemanRelativePath`.  That way the function names reference what they return, rather than the task they perform.

webpack/test-utils/findForeman.js
```diff
@@ -0,0 +1,29 @@
...
+
+  if (!fullPath) throw new Error(notFound);
+  return fullPath;
+};
+
+// Get a subdirectory within Foreman
+const getForemanRelativePath = (relativeForemanPath) => {
+  const foremanLocation = getForemanLocation();
+  const notFound = `Could not find ${relativeForemanPath} in ${foremanLocation}`;
+  const foremanRelativePath = path.join(foremanLocation, relativeForemanPath);
```
jeremylenz, line 24: Another suggestion for a bit less naming confusion:
```suggestion
  const result = path.join(foremanLocation, relativeForemanPath);
```

webpack/test-utils/findForeman.js
```diff
@@ -0,0 +1,29 @@
...
+    const possibleForemanPath = path.join(currentDir, relativeForemanPath);
+    if (fs.existsSync(possibleForemanPath)) fullPath = possibleForemanPath;
+  });
+
+  if (!fullPath) throw new Error(notFound);
+  return fullPath;
+};
+
+// Get a subdirectory within Foreman
+const getForemanRelativePath = (relativeForemanPath) => {
```
jeremylenz, line 21: Too much naming confusion here.  Is there a better name for `relativeForemanPath`? 
```suggestion
const getForemanRelativePath = (innerPath) => {
```

johnpmitsch:
[test katello]

.github/workflows/react_tests.yml
johnpmitsch: It would be nice, but we'll have to think about how to separate out the common actions like the checkout structure and npm installing, probably in another action. 

Seeing as they are pretty quick, maybe we can hold off on that for now?

johnpmitsch:
@jeremylenz thanks for the feedback, updated! The Jenkins job has also been updated to accommodate this change.

I also was curious if GH checkout action would base your PR on the master branch, which it looks like it does. I added a line (which is now removed) to check the git log and the latest commit was:

a91df3103b43d6c9e06347168ec8c24231e8f913 Merge f07b4352bac77024d37b086bc2f25e8df92a8844 into 394c8b6a3daaf59bb8942b0f555f030a4a107a52

which is the latest commit in the branch f07b4352bac77024d37b086bc2f25e8df92a8844 merged into the latest commit in Katello 394c8b6a3daaf59bb8942b0f555f030a4a107a52

This means we are testing the PR as it would be if merged into Katello, not the state of the branch, which could be missing commits if it hasn't been rebased on master in a while. This is what we do in Jenkins too.

Just want to record this small investigation here in case the question ever comes up again or others are wondering the same :)

.github/workflows/react_tests.yml
jeremylenz: This may be the only time this code will be touched in years.  I think it would be pretty easy to separate out.. you could just rename the existing `build` job to `tests` and add another one (in the same file) for `linting`.  For that you probably could just install Katello only, `npm i`, and then do `npx eslint`.  And the testing and linting would run concurrently!  
If there are any complications at all then I'm fine with the code as-is, but I still think it's worth a shot.

johnpmitsch: thanks for clarifying, I thought you meant to separate out into different actions which would (afaict) require publishing an action to share the common steps.

Eslint does need foreman checked out and installed for a couple reasons: to use the eslint config from foreman, which exists in `vendor-dev` and to check packages for the `    'import/no-extraneous-dependencies'` check. Both of these could maybe be improved in the future but that is the best I can do with the current setup.

I kept the build steps the same and set them up to run the jobs concurrently, currently they aren't working but I'll poke around!

johnpmitsch:
[test katello]

PR #8725, opened 2020-05-25
Refs #29330 - fix install_all errata

m-bucher:
@sbernhard @jlsherrill REX always tried to install all applicable errata, even if only a subset had been selected.

### DIFF
--- a/app/controllers/katello/remote_execution_controller.rb
+++ b/app/controllers/katello/remote_execution_controller.rb
@@ -43,7 +43,7 @@ module Katello
       end
 
       def errata_inputs
-        if params[:install_all]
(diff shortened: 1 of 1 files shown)


app/controllers/katello/remote_execution_controller.rb
```diff
@@ -43,7 +43,7 @@ def hosts
       end
 
       def errata_inputs
-        if params[:install_all]
+        if params[:install_all] == 'true'
```
jlsherrill, line 46: Using  ::Foreman::Cast.to_bool(params[:install_all])  would be preferred here

m-bucher: I know such a method existed :smile: 
Thanks for the hint.

PR #8726, opened 2020-05-26
Fixes #28112 - encrypt upstream password for root repository in audits

piyushtiw:
Bug fix for http://kanboard-foreman-teamc-kanboard.int.open.paas.redhat.com/project/2/task/6147

### DIFF
--- a/db/migrate/20200526015005_encrypt_password_in_audit.rb
+++ b/db/migrate/20200526015005_encrypt_password_in_audit.rb
@@ -1,0 +1,33 @@
+class EncryptPasswordInAudit < ActiveRecord::Migration[6.0]
+  def encrypt_password_for_audit(audit)
+    changes = audit.audited_changes
+    name = :upstream_password.to_s
+    change = changes[name]
+
+    if change.is_a? Array
+      changes[name] = change.map { |_| AuditExtensions::REDACTED }
+    elsif change
+      changes[name] = AuditExtensions::REDACTED
+    end
+
+    audit.save!
+  end
+
+  def encrypt_upstream_password_for_root_repository_audits
+    ::Katello::RootRepository.all.each do |repository|
+      repository.audits.each do |audit|
+        if audit.audited_changes.key? "upstream_password"
+          encrypt_password_for_audit(audit)
+        end
+      end
+    end
+  end
+
+  def up
+    encrypt_upstream_password_for_root_repository_audits
+  end
+
+  def down
+    # Do nothing
+  end
+end


jturel:
[test katello]

db/migrate/20200526015005_encrypt_password_in_audit.rb
```diff
@@ -0,0 +1,29 @@
...
+        if audit.audited_changes.key? "upstream_password"
+          changes = audit.audited_changes
+          name = :upstream_password.to_s
+          change = changes[name]
+
+          if change.is_a? Array
+            changes[name] = change.map { |_| AuditExtensions::REDACTED }
+          elsif change
+            changes[name] = AuditExtensions::REDACTED
+          end
```
parthaa, line 14: Could refactor this logic a bit ? may be in the root repo add a `update_password_for_audit(audit)` method and call it here.

parthaa:
There are 2 issues here 

1. The audit logs for the root repository are showing the upstream_password in clear text. This issue is addressed by this PR.
2. The upstream_password for root_repository is stored in clear text in the database.  The password needs to be encrypted. Either we fix it here or file a separate issue. I 'd recommend filing a separate one and linking the downstream bz.

parthaa:
> 2\. The upstream_password for root_repository is stored in clear text in the database.  The password needs to be encrypted. Either we fix it here or file a separate issue. I 'd recommend filing a separate one and linking the downstream bz. Look at https://github.com/theforeman/foreman/blob/develop/app/models/http_proxy.rb#L10 for example on how to do that (think you have to just include Encryptable and encrypts)

Piyush filed https://projects.theforeman.org/issues/29931

parthaa:
> 2\. The upstream_password for root_repository is stored in clear text in the database.  The password needs to be encrypted. Either we fix it here or file a separate issue. I 'd recommend filing a separate one and linking the downstream bz. Look at https://github.com/theforeman/foreman/blob/develop/app/models/http_proxy.rb#L10 for example on how to do that (think you have to just include Encryptable and encrypts)

@piyushtiw filed -> https://projects.theforeman.org/issues/29931

app/models/katello/root_repository.rb
```diff
@@ -333,6 +334,28 @@ def format_arches
       end
     end
 
+    def update_audit_password
+      audit = audits.last
+
+      if audit.audited_changes.key? "upstream_password"
+        encrypt_password_for_audit(audit)
+      end
+    end
```
jlsherrill, line 343: is audits.last really guaranteed to return the previously saved audit entry? 

You might could order by :created_at to try to ensure that, but i feel like it would be better to extend the audit model and add a before_save callback to modify the audited_changes hash in a similar way that you've done here (if the associated model is a RootRepository)

jlsherrill: here's where we 'extend' other models/classes: https://github.com/Katello/katello/blob/master/lib/katello/engine.rb#L155-L170

parthaa: might wanna explore https://github.com/theforeman/foreman/blob/develop/app/models/lookup_keys/lookup_key.rb#L2

piyushtiw:
Please refer to https://github.com/Katello/katello/pull/8729/files

PR #8728, opened 2020-05-26
Fixes #29933 - log error when repo scan errors

jlsherrill:
with an unexpected exception


jlsherrill:
The lack of this logging made debugging an issue a lot harder than it should have been

PR #8730, opened 2020-05-27
Fixes #29942 - Prevent duplicate toasts on Subscriptions page

jeremylenz:
Multiple toast notifications were appearing when importing a manifest:

manifest_success

During API interval polling of the subscription task state, Katello is supposed to look for the first non-pending "stopped / success" response, stop polling, and send a Toast notification informing the user.  I'm not certain, but the issue was probably due to multiple "success" messages coming back because polling was not stopped soon enough.

Whatever the cause, this change will prevent multiple "task finished" toasts for the same task.

### DIFF
--- a/webpack/scenes/Subscriptions/SubscriptionActions.js
+++ b/webpack/scenes/Subscriptions/SubscriptionActions.js
@@ -106,7 +106,7 @@ export const resetTasks = () => (dispatch) => {
 
 export const handleTask = task => async (dispatch, getState) => {
   if (selectIsPollingTask(getState(), SUBSCRIPTIONS)) {
-    if (!task.pending) {
+    if (!task.pending && task.result !== 'pending') {
       dispatch(stopPollingTask(SUBSCRIPTIONS));
       dispatch(toastTaskFinished(task));
       dispatch(resetTasks());
--- a/webpack/scenes/Tasks/TaskActions.js
+++ b/webpack/scenes/Tasks/TaskActions.js
@@ -5,8 +5,16 @@ import { stopInterval, withInterval } from 'foremanReact/redux/middlewares/Inter
 import { foremanTasksApi } from '../../services/api';
 import { bulkSearchKey, pollTaskKey, taskFinishedToast } from './helpers';
 
-export const toastTaskFinished = task => async dispatch =>
+const finishedTasks = {};
+
+export const toastTaskFinished = task => async (dispatch) => {
+  if (task.id) {
+    // Keep track of tasks we've already notified about to prevent duplicate toasts from appearing
+    if (finishedTasks[task.id]) return;
+    finishedTasks[task.id] = true;
+  }
   dispatch(addToast(taskFinishedToast(task)));
+};
 
 const taskBulkSearchParams = params => ({
   search: Object.entries(propsToSnakeCase(params))


webpack/scenes/Subscriptions/SubscriptionActions.js
```diff
@@ -106,7 +106,7 @@ export const resetTasks = () => (dispatch) => {
 
 export const handleTask = task => async (dispatch, getState) => {
   if (selectIsPollingTask(getState(), SUBSCRIPTIONS)) {
-    if (!task.pending) {
+    if (!task.pending && task.result !== 'pending') {
```
jeremylenz, line 109: `task.pending` will be true if `task.state === 'pending'`.  So I'm not sure why we'd ever have a state of "pending" and not a result of "pending," but figured it can't hurt to address that case..

jturel, line 109: I wonder if this should be in a selector in TaskSelectors

jeremylenz: `selectTaskIsDefinitelyNotPendingOrAtLeastWerePrettySure`

jeremylenz: Added the selector.   Since we're inside a thunk as opposed to a React component, I had to use `getState()` so it feels kind of clunky to me.  I used a separate commit just in case.. let me know your thoughts

webpack/scenes/Tasks/TaskActions.js
```diff
@@ -5,8 +5,16 @@ import { stopInterval, withInterval } from 'foremanReact/redux/middlewares/Inter
 import { foremanTasksApi } from '../../services/api';
 import { bulkSearchKey, pollTaskKey, taskFinishedToast } from './helpers';
 
-export const toastTaskFinished = task => async dispatch =>
+const finishedTasks = {};
+
+export const toastTaskFinished = task => async (dispatch) => {
```
jturel, line 10: I'm wondering if this method care about whether we've already toasted about some task already, keeping in mind that this is probably the only area of the code affected by our problem.

Thinking more, maybe the issue isn't related to the task status at all. I have a theory it could be related to the SubscriptionsPage component being re-rendered while a manifest task is in progress.

See: https://github.com/Katello/katello/blob/master/webpack/scenes/Subscriptions/SubscriptionsPage.js#L45-L47
and
https://github.com/Katello/katello/blob/master/webpack/scenes/Subscriptions/SubscriptionActions.js#L109-L115

The problematic flow might be like this:

-> task polled
-> task has transitioned to finished
-> componentWillUpdate (task changed)
-> handleTask called
-> componentWillUpdate is called while handleTask is already executing, and now **handleTask is called again with the same completed task**

You might need to prevent this second `handleTask` from being called. Assuming this is the problem, is there a React way to handle it? Fixing the problem at this level seems like the way to go if it really is the problem, and it's probably pretty easy to write a test for.

jeremylenz: @jturel Thanks, you sent me on a journey :)

We were doing too much in `handleTask` anyway, so I split that into 2 actions.  I also moved all the logic to the component, which both feels better AND allows us to avoid using `getState()` in the thunk.  The new home for the logic is `componentDidUpdate` (I think that's what you meant above), which I chose so that we could use the `prevProps` for easy comparison.

jeremylenz:
Also, updated the `SubscriptionsTable.js` to fix the bug where expanded rows would collapse as soon as you click to select a row.  (The cause was that a variable called `groupedSubscriptions` was being cleared out every time new props were received.)  In the process I discovered that two functions were using React `setState` [incorrectly](https://reactjs.org/docs/faq-state.html#why-is-setstate-giving-me-the-wrong-value).  Refactored using the function form of `setState` so that we can rely on the previous state values.

webpack/scenes/Tasks/TaskActions.js
jturel: It looks awesome!!

jturel:
> Also, updated the `SubscriptionsTable.js` to fix the bug where expanded rows would collapse as soon as you click to select a row. (The cause was that a variable called `groupedSubscriptions` was being cleared out every time new props were received.) In the process I discovered that two functions were using React `setState` [incorrectly](https://reactjs.org/docs/faq-state.html#why-is-setstate-giving-me-the-wrong-value). Refactored using the function form of `setState` so that we can rely on the previous state values.

Usually I'm all for including minor fixes like this, but this time we need to handle it differently since this change is gonna get cherry-picked into a downstream build where we only want things explicitly targeted to it. We have two feasible options IMO:

1. isolate the secondary change to a separate commit in this PR and use 'Fixes' for https://projects.theforeman.org/issues/26954
2. create a second PR with the secondary change

webpack/scenes/Subscriptions/SubscriptionsPage.js
```diff
@@ -39,11 +39,16 @@ class SubscriptionsPage extends Component {
 
   componentDidUpdate(prevProps) {
     const {
-      organization, task, handleTask,
+      organization, task, handleStartTask, handleFinishedTask, isTaskPending, isPollingTask,
     } = this.props;
-
     if (task) {
-      handleTask(task);
+      if (isPollingTask) {
```
jturel, line 45: Let's get some test coverage here too :)

jeremylenz:
> create a second PR with the secondary change

Removed the commit from this PR and raised https://github.com/Katello/katello/pull/8744

webpack/scenes/Subscriptions/SubscriptionsPage.js
jturel: I haven't actually tested these changes yet but I think there's an issue here. Based on the way the code reads, I think we'll only update the progress bar modal when we start polling a task and when that task completes so we'll miss the 'progress' aspect of the modal.

jeremylenz:
Tests updated & passing

PR #8731, opened 2020-05-27
Fixes #29995 - Capsule protected content

sjha4:
On the forklift branch pulp3, update  roles/katello_pulp3/tasks/main.yml with https://gist.github.com/sjha4/6506fe2ecadf51524a0463b212b1f9c3

Spin up a capsule using steps here:
https://projects.theforeman.org/projects/katello/wiki/Pulp_3_Integration#Smart-Proxy-Setup
Next update your box following steps: 
https://gist.github.com/sjha4/c58367185083f6ecd130a6e97c414b49

Now create a protected/unprotected repo on your server and verify that the content guard is set up properly on the capsule and the repo is protected as expected on the capsule.


app/lib/actions/pulp3/capsule_content/generate_metadata.rb
```diff
@@ -6,6 +6,7 @@ class GenerateMetadata < Pulp3::AbstractAsyncTask
         def plan(repository, smart_proxy, options = {})
           options[:contents_changed] = (options && options.key?(:contents_changed)) ? options[:contents_changed] : true
           sequence do
+            plan_action(RefreshContentGuard, smart_proxy) unless repository.unprotected
```
jlsherrill, line 9: we might need to run this regardless of repository.unprotected in case the value for that changed

app/services/katello/pulp3/repository_mirror.rb
```diff
@@ -145,6 +146,7 @@ def refresh_distributions(options = {})
         dist_params = {}
         dist_params[:publication] = options[:publication] if options[:publication]
         dist_params[:repository_version] = version_href if options[:use_repository_version]
+        dist_params[:content_guard_href] = options[:content_guard_href] if options[:content_guard_href]
```
jlsherrill, line 149: would it make more sense to 'lookup' the content guard href here and set it if needed?

jlsherrill: the other options on the distributor are created as part of the distribution_options method in this file

jturel:
@sjha4 can you add some notes about the goals of the PR and how to validate that when testing?

app/lib/actions/pulp3/capsule_content/generate_metadata.rb
jlsherrill: right, but what if you 'unset' the unprotected flag, in which case 'unprotected' will be false and this won't ever get called, resulting in the repo still being protected, right?

jlsherrill: oh, right, the actual association to the content guard is done as part of refreshrepos

app/services/katello/pulp3/repository_mirror.rb
```diff
@@ -87,9 +93,8 @@ def remote_options
         base_options = common_remote_options
         if (type_specific_options = repo_service.try(:mirror_remote_options))
           base_options.merge(type_specific_options)
-        else
-          base_options.merge(url: remote_feed_url)
         end
+        base_options.merge(url: remote_feed_url)
       end
```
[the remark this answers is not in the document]
jlsherrill: i think this would break the docker syncing which is supplying its own 'url', which gets overridden here.  You might wanna move this above the if statement

jlsherrill:
one comment around docker, but otherwise looks good!

app/lib/actions/pulp3/capsule_content/generate_metadata.rb
jlsherrill, line 9: i think it would be better to put this at more of a global  capsule sync level, that way it only executes once per smart_proxy.

jlsherrill: even though it may run regardless of whether its needed (without the 'unless repository.unprotected') it'd only run once

app/lib/actions/katello/capsule_content/sync.rb
```diff
@@ -32,6 +32,10 @@ def plan(smart_proxy, options = {})
 
           smart_proxy_helper = ::Katello::SmartProxyHelper.new(smart_proxy)
           repositories = smart_proxy_helper.repos_available_to_capsule(environment, content_view, repository)
+          if repositories.any? { |repo| smart_proxy.pulp3_support?(repo) }
+            smart_proxy.ping_pulp3
+            plan_action(Actions::Pulp3::CapsuleContent::RefreshContentGuard, smart_proxy)
+          end
           smart_proxy.ping_pulp3 if repositories.any? { |repo| smart_proxy.pulp3_support?(repo) }
```
sjha4, line 39: Calling refreshContentGuard here so it is planned only once. The Action itself does not perform anything if there is a Content Guard configured.

jlsherrill: i'm not sure that we need a ping_pulp3 here, planning Actions::Pulp3::CapsuleContent::RefreshContentGuard should take care of that because it inherits from  Pulp3::AbstractAsyncTask, which inherits from Pulp3::Abstract, which includes the Actions::Middleware::Pulp3ServicesCheck check

sjha4: Oh..I meant to remove the 2nd of these ping_pulp3 lines..Let me test after removing both..

sjha4: Updated..

sjha4:
[test katello]

sjha4:
[test katello]

jlsherrill:
[test katello]

PR #8732, opened 2020-05-27
Fixes #29948 - Unable to re-import subscriptions in large environment

jturel:
Prevent active record from generating a query that postgres doesn't like (we use this approach in other places too).

I couldn't reproduce on my dev box, so recommend doing it downstream.

Testing the approach is simple from the console:

```
uuids = []
100000.times { |x| uuids << SecureRandom.uuid }
Katello::Host:::SubscriptionFacet.where(uuid: uuids)
```

This raises an error: `ActiveRecord::StatementInvalid (PG::ProtocolViolation: ERROR: invalid message format)`

But running the query as it is in the PR works:

```
Katello::Host:::SubscriptionFacet.where('uuid in (?)', uuids)
```

You can test the actual functionality by running `Katello::Pool.import_all` from the rails console. I also added a unit test since we have 0 coverage here it seems!

### DIFF
--- a/app/models/katello/glue/candlepin/pool.rb
+++ b/app/models/katello/glue/candlepin/pool.rb
@@ -164,7 +164,7 @@ module Katello
       def import_hosts
         uuids = Resources::Candlepin::Pool.consumer_uuids(self.cp_id)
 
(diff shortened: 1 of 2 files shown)


jturel:
> I have seen this change to get around a similar issue before so Ack on the code part..Strangely on dev box this is working for me on the console and I see the error downstream..Don't fully understand why that is? Rails version? I tested on 6.7 downstream..

Good point. I bet it is the rails version! Thanks for the review

PR #8735, opened 2020-05-28
Fixes #29366 - distinguish between Wrokstation and Server

sjha4:
Opening to move this forward: https://github.com/Katello/katello/pull/8616


sjha4:
[test katello]

PR #8736, opened 2020-05-29
Fixes #29956 - Lock default job templates

stejskalleos:


sjha4:
Thanks @stejskalleos , @adamruzicka ! Merging..

PR #8744, opened 2020-06-02
Fixes #26954 - Fix auto-closing rows on Subscriptions table

jeremylenz:
(Broken out into a separate PR from #8730)

This change updates `SubscriptionsTable.js` to fix the bug where expanded rows would collapse as soon as you click to select a row.  (The cause was that a variable called `groupedSubscriptions` was being cleared out every time new props were received.)  In the process I discovered that two functions were using React `setState` [incorrectly](https://reactjs.org/docs/faq-state.html#why-is-setstate-giving-me-the-wrong-value).  Refactored using the function form of `setState` so that we can rely on the previous state values.

### DIFF
--- a/webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTableHelpers.js
+++ b/webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTableHelpers.js
@@ -76,13 +76,17 @@ export const buildTableRows = (groupedSubscriptions, availableQuantities, update
   return rows;
 };
 
-export const groupSubscriptionsByProductId = ({ results: subscriptions }) => {
+export const groupSubscriptionsByProductId = (
+  { results: subscriptions },
+  prevGroupedSubscriptions,
+) => {
   const grouped = {};
 
   subscriptions.forEach((subscription) => {
     if (grouped[subscription.product_id] === undefined) {
+      const prevOpenState = prevGroupedSubscriptions?.[subscription.product_id]?.open || false;
       grouped[subscription.product_id] = {
-        open: false,
+        open: prevOpenState,
         subscriptions: [],
       };
     }
--- a/webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTable.js
+++ b/webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTable.js
@@ -29,7 +29,10 @@ class SubscriptionsTable extends Component {
       nextProps.subscriptions !== undefined &&
       !isEqual(nextProps.subscriptions, prevState.subscriptions)
     ) {
-      const groupedSubscriptions = groupSubscriptionsByProductId(nextProps.subscriptions);
+      const groupedSubscriptions = groupSubscriptionsByProductId(
+        nextProps.subscriptions,
+        prevState.groupedSubscriptions,
+      );
       const rows = buildTableRows(
         groupedSubscriptions,
         nextProps.subscriptions.availableQuantities,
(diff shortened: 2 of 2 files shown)


webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTableHelpers.js
```diff
@@ -76,13 +76,17 @@ export const buildTableRows = (groupedSubscriptions, availableQuantities, update
...
-export const groupSubscriptionsByProductId = ({ results: subscriptions }) => {
+export const groupSubscriptionsByProductId = (
+  { results: subscriptions },
+  prevGroupedSubscriptions,
+) => {
   const grouped = {};
 
   subscriptions.forEach((subscription) => {
     if (grouped[subscription.product_id] === undefined) {
+      const prevOpenState = prevGroupedSubscriptions?.[subscription.product_id]?.open || false;
```
jeremylenz, line 87: Using JS optional chaining operator here (`?.`).  It's similar to Ruby's `&.` operator.  Based on this [canIuse entry](https://caniuse.com/#search=%3F.) confirming support in most modern browsers, and [this Babel PR](https://github.com/babel/babel/pull/10811) adding it to `@babel/preset-env` which will cover older browsers, I think it's safe to start using.  (more discussion in this Babel [issue](https://github.com/babel/babel/issues/10809))

jeremylenz: @sharvit @amirfefer @laviro  thoughts?

johnpmitsch:
@jeremylenz works well!

A couple thoughts:

1) `updateRows` and `toggleSubscriptionGroup` look pretty similar, what do you think about combining them? I could see updateRows taking an optional `groupId` as the second arg and modifying `groupedSubscriptions` if it is present.

2) I would love to start adding react-testing-library tests from the page or table level along with these bug fixes to prevent regressions. I understand we don't have them for subscriptions already so it may be a lot of overhead to set up. Any interest in looking into it?

webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTableHelpers.js
jeremylenz: looks like `foo?.bar` would transpile to
```js
foo === null || foo === void 0 ? void 0 : foo.bar;
```
Ugly but that's Babel's job anyway :smile:

johnpmitsch: Curious to hear if there are other considerations, but seems fine from my perspective :+1: Thanks for checking!

laviro: Looks like a great JS feature :fireworks: 
I would like to see it all over our code :)

jeremylenz:
> 1. `updateRows` and `toggleSubscriptionGroup` look pretty similar, what do you think about combining them?

My thought with the changes here was simply to move from the incorrect form of `setState` (where it assumes that `this.state` is up to date and then depends on that value to set the new state) to the function form (which provides `prevState` for reliable comparison.)  I wasn't aiming for any other logic changes or refactoring.

The code does look very similar, but I feel like the purposes are different enough that it wouldn't make sense to combine into one function.  `toggleSubscriptionGroup` expands or collapses a row, while `updateRows` updates a quantity.  The two methods also end up in two different table props (`inlineEditController` and `toggleSubscriptionGroup`) which is another argument for keeping them separate.

>  I would love to start adding react-testing-library tests from the page or table level along with these bug fixes to prevent regressions.

Do you mean just adding a test that asserts the bug is fixed?  Or rewriting the existing tests with r-t-l?  I was hoping to keep this PR quick and easy, since it was a spinoff of another PR, but I'm open to discussion..

johnpmitsch:
> > 1.
> 
> My thought with the changes here was simply to move from the incorrect form of `setState` (where it assumes that `this.state` is up to date and then depends on that value to set the new state) to the function form (which provides `prevState` for reliable comparison.) I wasn't aiming for any other logic changes or refactoring.
> 

It seems like a good opportunity to clean up this area, which has gotten quite bloated. Both functions recreate the table so seemed like a good way to clean up the file, but I hadn't tested anything out. It's not essential to the fix, just a suggestion.

> The code does look very similar, but I feel like the purposes are different enough that it wouldn't make sense to combine into one function. `toggleSubscriptionGroup` expands or collapses a row, while `updateRows` updates a quantity. The two methods also end up in two different table props (`inlineEditController` and `toggleSubscriptionGroup`) which is another argument for keeping them separate.

Ideally we submit tests that recreate the behavior when submitting a fix. r-t-lib seems like the best candidate to do so since it allows you to act more like a user, I could see a test on the `SubscriptionsTable` being valuable, doing it from the Table will make it so we don't have to mock http calls, which I imagine would take some time to figure out. We don't have to start with this PR as I know some setup will be involved, but submitting tests with bug fixes is a pattern we should keep in mind as we build out and maintain the UI.

Both were just suggestions, the code does what it intends to do so ACK from me! :+1:

PR #8746, opened 2020-06-04
Refs #29929 - disable subscription through activation key in SCA

piyushtiw:
What was done:
Remove subscription tab from activation_key page for SCA enabled org.
Check for SCA enabled org while adding subscription through activation-key in controller
Remove subscription status column from content host association table at activation key page.
Add test for activation key controller changes.

Not included:
If we come on subscription page through url, we will display the subscription page then but because we have changed the add_subscription method in controller, user can only see the page, can not do anything.

### DIFF
--- a/test/controllers/api/v2/activation_keys_controller_test.rb
+++ b/test/controllers/api/v2/activation_keys_controller_test.rb
@@ -461,9 +461,19 @@ module Katello
       allowed_perms = [@update_permission]
       denied_perms = [@view_permission, @create_permission, @destroy_permission]
 
+      Organization.any_instance.stubs(:simple_content_access?).returns(false)
+
       assert_protected_action(:add_subscriptions, allowed_perms, denied_perms, [@organization]) do
         post(:add_subscriptions, params: { :organization_id => @organization.id, :id => @activation_key.id, :subscription_id => 123 })
       end
+    end
+
+    def test_add_subscriptions_fails
+      Organization.any_instance.stubs(:simple_content_access?).returns(true)
+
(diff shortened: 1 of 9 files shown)


piyushtiw:
[test katello]

jturel:
[test katello]

test/controllers/api/v2/activation_keys_controller_test.rb
```diff
@@ -461,11 +461,21 @@ def test_add_subscriptions_protected
...
         post(:add_subscriptions, params: { :organization_id => @organization.id, :id => @activation_key.id, :subscription_id => 123 })
       end
     end
 
+    def test_add_subscriptions_fails
+      Organization.any_instance.stubs(:simple_content_access?).returns(true)
+
+      post(:add_subscriptions, params: { :organization_id => @organization.id, :id => @activation_key.id, :subscription_id => 123 })
+
+      assert_response 400
```
jeremylenz, line 476:
```suggestion
      assert_response :bad_request
```
Using the descriptor makes it so you don't have to remember HTTP status codes :)

jeremylenz:
@jturel Could you confirm that we didn't miss any other places in the UI where you can attach a sub to an AK?

jturel:
> @jturel Could you confirm that we didn't miss any other places in the UI where you can attach a sub to an AK?

it can only be done through Content -> Activation Keys so LGTM!

piyushtiw:
[test katello]

jeremylenz:
[test katello]

PR #8749, opened 2020-06-04
Fixes #29753 - Support skipping srpms

sjha4:
**To test:**
1. Create yum repo with srpm content remote. ex: (https://repos.fedorapeople.org/repos/pulp/pulp/fixtures/srpm/)
2. Chose srpm as ignorable_content when creating repo.
3. Sync and verify that srpm content is not synced.


sjha4:
The tests don't seem to be failing all at once..new errors in every batch..I am working on re-recording VCRs but the overall PR is ready for review.

jjeffers:
APJ

jlsherrill:
@sjha4 looks like there's a TON of deleted cassettes, i think you just need to restore those

PR #8754, opened 2020-06-08
Fixes #29751 - Pulp 3 incremental update

ianballou:
Pulp 3 incremental update support for Yum repositories.  To test, try doing various incremental update commands just like you would with Pulp 2.

Note, since this is with Pulp 3 don't forget to enable `:katello_applicability:` in the config.

Potential test steps:

1) Create a content view with yum content
2) Filter out some errata and RPMs
3) Run various incremental update commands.  Incremental update lets you make a new content view version with extra RPMs and errata that you specify.  For example:
```
hammer content-view version incremental-update --organization "Default Organization" --content-view-version-id 148 --errata-ids 5,13 --package-ids 64,69
```
You should notice that dependency solving is on, like with Pulp 2.

You can also try applying errata from the Content -> Errata page.

Questions:
1) Is it okay that the added content is being calculated from the content view version rather from the Pulp task output?  This seemed necessary since the copy task doesn't always result in a new Pulp repository version, which would cause no content to be reported.


ianballou:
I realized I need to fix up the reported content output if a new Pulp repo version isn't created.  No new content will be reported for that incremental update since the content exists in the previous version.

There also needs to be a report for module streams.  It looks like that was not yet implemented for Pulp 2.

ianballou:
> With how the code is now, no new content will be reported as copied if it already exists in the latest repo version during the incremental update.
> 
> There also needs to be a report for module streams. It looks like that was not yet implemented for Pulp 2.

I've pushed up new code that fixes both of these.

app/lib/actions/katello/content_view_version/incremental_update.rb
```diff
@@ -142,22 +145,47 @@ def run
...
+                  case type
+                  when ::Katello::Erratum::CONTENT_TYPE
+                    content[::Katello::Erratum::CONTENT_TYPE] << unit['id']
+                  when ::Katello::ModuleStream::CONTENT_TYPE
+                    content[::Katello::ModuleStream::CONTENT_TYPE] << "#{unit['name']}:#{unit['stream']}:#{unit['version']}"
+                  when ::Katello::Rpm::CONTENT_TYPE
+                    content[::Katello::Rpm::CONTENT_TYPE] << ::Katello::Util::Package.build_nvra(unit)
+                  when ::Katello::Deb::CONTENT_TYPE
+                    content[::Katello::Deb::CONTENT_TYPE] << "#{unit['name']}_#{unit['version']}_#{unit['architecture']}"
+                    SmartProxy.pulp_master
```
sjha4, line 185: Is this needed?

ianballou: Good catch, I don't know where that came from!

parthaa:
Seems like its not doing dep solving across multiple repositories.

- Create a product with the following repoisitories

- Create and sync the  repositories pointing to -> ```https://partha.fedorapeople.org/test-repos/separated/modules-rpms/``` and ```https://partha.fedorapeople.org/test-repos/separated/rpm-deps/```

So the special thing about these 2 repos is that the dependencies for rpms in the module-rpms are in rpm-deps

```
* Module-Rpms: Test repository with modules and rpms. There are 2 RPM dependencies missing in this repository. 
* Rpm-Deps-Only: Contains the dependent ‘shark’ and ‘stork’ rpms but no modules.
So the dependency tree looks like

  [walrus-0.71] - repository - Module-Rpms
  ├── walrus - RPM - repository - Module-Rpms
      └── whale - RPM - repository - Module-Rpms
          ├── shark - RPM - repository - Rpm-Deps-Only
          └── stork - RPM - repository - Rpm-Deps-Only

```

- Create a content view with those repos
- Create a excludes package filter that excludes "*", i.e exclude all non modular  rpms
- Publish the CV
- I would expect to see something like
expected 
with the version details for `Yum Repositories` appearing like
expectedrepos

- Instead I see
Actua
with the `Yum Repositories` for the version detail appearing like
ActuaRepo

(this is a probably an unconnected bug just making a note here :))

- Note: `WALRUS-2013:0002` errata has walrus-0.71 as its rpm dependency. So incrementally updating that should automatically copy walrus-0.71 and all its dependencies across BOTH repos.
 - Now make a note of the generated version and try the following line
```hammer content-view version incremental-update --organization "Default Organization" --content-view-version-id 2 --errata-ids "WALRUS-2013:0002"
```
Now when it completes I would expect to see 
```
 hammer content-view version incremental-update --organization "Default Organization" --content-view-version-id 10 --errata-ids "WALRUS-2013:0002"
[.....................................................................................................................................................................................................................................] [100%]
Content View: test version 1.1
Added Content:
  Errata:
        WALRUS-2013:0002
  Packages:
        shark-0.1-1.noarch
        stork-0.12-2.noarch
        walrus-0.71-1.noarch
        whale-0.2-1.noarch
```

Instead I see
```
[.....................................................................................................................................................................................................................................] [100%]
Content View: test version 1.1
Added Content:
  Errata:
        WALRUS-2013:0002
  Module Streams:
        walrus:0.71:20190707165243
  Packages:
        walrus-0.71-1.noarch
```
Which means the dependencies did not get copied at all.

ianballou:
I may need to refactor the code to support cross-repo dep solving.  The Pulp 3 copy API requires a source-dest pair for each additional repo that should be considered in the dep solving.  With how the incremental update code is now, the destination repo href isn't available to copy units.

I think the destination repo is guaranteed to already be created, so it might be a matter of just passing the right destinstaion repo information to copy_units.  This would be temporary, since after Pulp 2 is dropped we will change incremental update to do the entire copy at once.

parthaa:
> I may need to refactor the code to support cross-repo dep solving. The Pulp 3 copy API requires a source-dest pair for each additional repo that should be considered in the dep solving. With how the incremental update code is now, the destination repo href for sibling repos isn't available to copy_units.
> 
> I think the destination repo is guaranteed to already be created, so it might be a matter of just passing the right destinstaion repo information to copy_units. This would be temporary, since after Pulp 2 is dropped we will change incremental update to do the entire copy at once.

If you need to handle cross repo dep solving as a different issue I am ok with that. Just file a bug (or check with other titans and figure out :) )

ianballou:
@parthaa I should be able to finish that up in this PR, it'll need to be there for 3.16 anyway.

ianballou:
I'm almost finished with the refactoring to support cross-repo dep solving during incremental update.  After this PR is finished, another PR will be required to make cross-repo dep solving work with content view filtering.  This will likely require a rewrite of Pulp3's `CopyAllUnits`, CopyContent, and an update of `copy_units` in `katello/pulp3/repository/yum.rb`.  My refactor is including new tasks to copy units between multiple repos at once and to save multiple created repo versions too.  These will be helpful for the refactoring needed for content view filtering w/ dep solving.

After this PR is merged, we'll need to create a redmine issue for the above.

ianballou:
@parthaa @sjha4 this is ready for a re-review now.  I had to make significant changes to get the cross-repo dep solving to work.  Try to break it!

I'll need to update the tests.

parthaa:
> @parthaa @sjha4 this is ready for a re-review now. I had to make significant changes to get the cross-repo dep solving to work. Try to break it!

Works well with my test scenario. +1

app/lib/actions/pulp3/repository/save_versions.rb
```diff
@@ -0,0 +1,41 @@
...
+              unversioned_href == version_href[0..-2].rpartition('/').first
+            end
+            # Successive incremental updates won't generate a new repo version, so fetch the latest Pulp 3 repo version
+            new_version_href ||= ::Katello::Pulp3::Api::Yum.new(SmartProxy.pulp_master!).
+              repositories_api.read(dest_repo.backend_service(SmartProxy.pulp_master).
+              repository_reference.repository_href).latest_version_href
+
+            if new_version_href
+              dest_repo.update(version_href: new_version_href)
+              dest_repo.index_content
```
sjha4, line 32: Do we still need to index if version doesn't change?

app/services/katello/pulp3/repository/yum.rb
```diff
@@ -70,28 +70,53 @@ def partial_repo_path
...
+                source_repo_version = ::Katello::Repository.find(source_repo).version_href
+
+                dest_repo = ::Katello::Repository.find(dest_repo_map[:dest_repo])
+                dest_repo_href = ::Katello::Pulp3::Repository::Yum.new(dest_repo, SmartProxy.pulp_master).repository_reference.repository_href
+                data.config << {
+                  source_repo_version: source_repo_version,
+                  dest_repo: dest_repo_href,
+                  dest_base_version: dest_repo_map[:base_version],
+                  content: content_unit_hrefs.sort
+                }
```
sjha4, line 103: Does this _need_ to be sorted?

app/lib/actions/pulp3/repository/save_versions.rb
ianballou: Ah no, we don't. I'll refactor.

app/services/katello/pulp3/repository/yum.rb
ianballou: It was sorted before so that the VCR tests would work.  Maybe there's a way to keep that sorting on the testing side of things?

ianballou:
Also I do realize I should add some more tests.

app/lib/actions/pulp3/repository/save_versions.rb
ianballou: Also I realize that SaveVersions shouldn't take source-dest repo pairs, that's incremental-update specific.  I'm going to refactor so that SaveVersions just takes a list of repositories.

app/services/katello/pulp3/repository/yum.rb
sjha4: or maybe sort outside the loop once.

sjha4:
This looks good to me..I will do 1 last round of testing with the changes here..

PR #8755, opened 2020-06-09
Refs #30021 - correct pulp 2to3 migration client dependency

evgeni:


evgeni:
RPM builds otherwise fail with:

```
Gem loading error: Could not find 'pulp_2to3_migration_client' (< 0.3.0, > 0.02b2) - did find: [pulp_2to3_migration_client-0.2.0b2]
```

PR #8758, opened 2020-06-09
Fixes #30077 - unsorted content-views in hostgroup

m-bucher:
This was probably also the case for lifecyle_environments and content-source, so I adapted them as well.


jlsherrill:
[test katello]

PR #8761, opened 2020-06-10
Refs #29928 remove subscription widgets

piyushtiw:
What was done
=> Remove subscriptions widgets from Dashboard when SCA is enabled.
=> Remove "Add subscriptions widgets" from manage button drop down when SCA is enbled.

### DIFF
--- a/app/helpers/katello/concerns/dashboard_helper_extensions.rb
+++ b/app/helpers/katello/concerns/dashboard_helper_extensions.rb
@@ -28,6 +28,16 @@ module Katello
       def unsubscribed_hypervisor_count
         host_query.search_for("subscription_status = unsubscribed_hypervisor").size
       end
+
+      def removed_widgets
+        widgets = super
+
+        if Organization.current&.simple_content_access?
+          widgets.reject! { |widget| ::Widget.singleton_class::SUBSCRIPTION_TEMPLATES.include? widget[:template] }
+        end
+
+        widgets
+      end
     end
   end
 end
--- a/app/models/katello/concerns/widget_extensions.rb
+++ b/app/models/katello/concerns/widget_extensions.rb
@@ -1,0 +1,23 @@
+module Katello
+  module Concerns
+    module WidgetExtensions
+      extend ActiveSupport::Concern
+
+      module ClassMethods
+        SUBSCRIPTION_TEMPLATES = %w[subscription_status_widget subscription_widget].freeze
+
+        def without_subscription_widgets
+          where.not(template: ::Widget.singleton_class::SUBSCRIPTION_TEMPLATES)
+        end
+
+        def available
+          if Organization.current&.simple_content_access?
+            super.without_subscription_widgets
+          else
+            super
+          end
+        end
+      end
+    end
+  end
+end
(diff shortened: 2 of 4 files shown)


piyushtiw:
[test katello]

app/models/katello/concerns/widget_extensions.rb
```diff
@@ -0,0 +1,23 @@
+module Katello
+  module Concerns
+    module WidgetExtensions
+      extend ActiveSupport::Concern
+
+      included do
+        SUBSCRIPTION_TEMPLATES = %w[subscription_status_widget subscription_widget].freeze
+
+        scope :remove_subscription_widgets, -> { where.not(template: ::Widget::SUBSCRIPTION_TEMPLATES) }
```
jeremylenz, line 9:
```suggestion
        scope :without_subscription_widgets, -> { where.not(template: ::Widget::SUBSCRIPTION_TEMPLATES) }
```
`remove` to me implies an action, but in reality this is more of a filter.

app/helpers/katello/concerns/dashboard_helper_extensions.rb
```diff
@@ -1,6 +1,10 @@
 module Katello
   module Concerns
     module DashboardHelperExtensions
+      include DashboardHelper
+
+      alias_method :removed_widgets_org, :removed_widgets
```
jeremylenz, line 6:
```suggestion
      alias_method :removed_widgets_orig, :removed_widgets
```

piyushtiw:
> Thanks @piyushtiw
> 
> Taking an initial look -
> 
> * I had an SCA manifest already loaded, but I still saw the two Subscription widgets on the dashboard.
> * I removed them by clicking the X on each of them
> * I then tried to add them again, but as expected, the action menu said "Nothing to add"
> 
> I would expect that the widgets would already be removed for me if SCA is enabled; as a user I should not have to remove them myself.
> 
> Also, I tried to reload the page at this point and got the following error:
> 
> ```
> NameError in Dashboard#index
> Showing /home/vagrant/foreman/app/views/dashboard/index.html.erb where line #3 raised:
> 
> uninitialized constant Widget::SUBSCRIPTION_TEMPLATES
> ```

@jeremylenz Have you restarted the server after coming to this branch?

jeremylenz:
@piyushtiw No, I was just reloading the page.  :see_no_evil:   

Works as expected now :)

app/models/katello/concerns/widget_extensions.rb
```diff
@@ -0,0 +1,23 @@
...
+  module Concerns
+    module WidgetExtensions
+      extend ActiveSupport::Concern
+
+      included do
+        SUBSCRIPTION_TEMPLATES = %w[subscription_status_widget subscription_widget].freeze
+
+        scope :without_subscription_widgets, -> { where.not(template: ::Widget::SUBSCRIPTION_TEMPLATES) }
+
+        singleton_class.send(:alias_method, :old_available, :available)
```
jlsherrill, line 11: A better way to do this is to use 'prepend':  https://www.justinweiss.com/articles/rails-5-module-number-prepend-and-the-end-of-alias-method-chain/

We have an example of the usage here:  https://github.com/Katello/katello/blob/master/lib/katello/engine.rb#L225   
where by it defines a method in the extension that overrides the original method, and you can call super() within it to call the original method.

mmoll:
[test katello]

parthaa:
[test katello]

PR #8762, opened 2020-06-10
Fixes #30082 - drop audit record creation when change in content_id

kgaikwad:


rplevka:
I tested this PR on a downstream 6.8 build and it seems to work fine.
- I tried to create a repo pre and post applying this change
- I tested that the 'unwanted' audit records created prior running the migration were successfully removed post running it

PR #8765, opened 2020-06-11
Fixes #30100 - reset http proxy in repos if deleted

jjeffers:
This PR handles cases where repositories are set to use a particular HTTP Proxy and then that HTTP Proxy is deleted.

After the deletion, the repositories are set to use the global default instead of still displaying the now deleted proxy. Likewise, any repository set to use a deleted global default proxy will now display the "Global Default (None)" instead of incorrectly displaying the old, deleted global proxy name.

Finally, the Content setting for the global default is reset to the "blank" option ("no global default"), which is the value shown when katello is first initialized (and there is no HTTP Proxy).

### DIFF
--- a/app/models/katello/concerns/http_proxy_extensions.rb
+++ b/app/models/katello/concerns/http_proxy_extensions.rb
@@ -8,11 +8,40 @@ module Katello
           :inverse_of => :http_proxy, :dependent => :nullify
         after_update :update_default_proxy_setting
         after_commit :update_repository_proxy_details
+        before_destroy :remove_references_to_proxy
 
         def self.default_global_content_proxy
           if Setting[:content_default_http_proxy]
             HttpProxy.unscoped.find_by(name: Setting[:content_default_http_proxy])
           end
+        end
+      end
+
+      def repositories_with_proxy(proxy_id)
+        root_repos = RootRepository.with_selected_proxy(proxy_id)
+
+        if self == HttpProxy.default_global_content_proxy
+          root_repos += RootRepository.with_global_proxy
+        end
+
+        root_repos
+      end
+
+      def remove_references_to_proxy
+        root_repos = repositories_with_proxy(nil).uniq.sort
+
+        setting = Setting.find_by(name: 'content_default_http_proxy')
+        if setting&.value && setting.value == self.name
+          setting.update_attribute(:value, '')
+        end
+
+        unless root_repos.empty?
+          ForemanTasks.async_task(
+            ::Actions::BulkAction,
+            ::Actions::Katello::Repository::Update,
+            root_repos,
+            http_proxy_policy: Katello::RootRepository::GLOBAL_DEFAULT_HTTP_PROXY,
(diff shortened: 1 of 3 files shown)


app/models/katello/concerns/http_proxy_extensions.rb
```diff
@@ -16,6 +17,32 @@ def self.default_global_content_proxy
...
       end
 
+      def update_references_to_proxy
+        root_repos = RootRepository
+          .where(:http_proxy_policy => Katello::RootRepository::USE_SELECTED_HTTP_PROXY)
+          .where(:http_proxy_id => nil)
+
+        if self == HttpProxy.default_global_content_proxy
+          root_repos += RootRepository.with_global_proxy
+        end
```
jlsherrill, line 27: we made a helper method for these lines:  https://github.com/Katello/katello/blob/9e9228f4d29601efefe8e95b36ca60fe9f199978/app/models/katello/root_repository.rb#L101

jlsherrill: actually that doesn't handle the global case, i swear we had one, let me see

jlsherrill: actually this method looks extremely similar to update_repository_proxy_details below, can you somehow combine them or reuse most of the logic?

jjeffers: Using the `with_selected_proxy(nil)` seems to yield different results, but I'm not sure why. Still looking into that. I do plan to try and consolidate some of the commonality between the different hooks.

app/models/katello/concerns/http_proxy_extensions.rb
```diff
@@ -16,6 +17,34 @@ def self.default_global_content_proxy
...
+        root_repos = RootRepository.with_selected_proxy(proxy_id)
+
+        if self == HttpProxy.default_global_content_proxy
+          root_repos += RootRepository.with_global_proxy
+        end
+
+        root_repos
+      end
+
+      def update_references_to_proxy
```
jlsherrill, line 30: small nitpick, but i think 'remove_references_to_proxy' would be more clear what its doing?

jlsherrill:
gonna do some testing, other than one small nitpick it looks good

app/models/katello/concerns/http_proxy_extensions.rb
jjeffers: I concur!

app/models/katello/concerns/http_proxy_extensions.rb
```diff
@@ -8,6 +8,7 @@ module HttpProxyExtensions
           :inverse_of => :http_proxy, :dependent => :nullify
         after_update :update_default_proxy_setting
         after_commit :update_repository_proxy_details
+        before_destroy :update_references_to_proxy
```
jlsherrill, line 11: forget to update this?

jturel:
[test katello]

PR #8768, opened 2020-06-11
Fixes #30114 - Stop Pulp 2 consumer binding if Katello Applicability is in use

ianballou:
Pulp 2 consumer binding should never happen if Katello Applicability (Pulp 3 applicability) is in use.

### DIFF
--- a/test/models/host/content_facet_test.rb
+++ b/test/models/host/content_facet_test.rb
@@ -67,6 +67,20 @@ module Katello
       content_facet_rec = host1.associated_audits.where(auditable_id: content_facet1.id)
       assert content_facet_rec, "No associated audit record for content_facet"
     end
+
+    def test_pulp2_binding_with_katello_applicability
+      SETTINGS[:katello][:katello_applicability] = true
+      org = taxonomies(:empty_organization)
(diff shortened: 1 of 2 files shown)


jlsherrill:
Did you include an extra commit here by mistake?

ianballou:
@jlsherrill I did, it's fixed now.

test/models/host/content_facet_test.rb
```diff
@@ -67,6 +67,20 @@ def test_audit_for_content_facet
...
+      org = taxonomies(:empty_organization)
+      host = ::Host::Managed.create!(:name => 'foohost', :managed => false, :organization_id => org.id)
+      content_facet = Katello::Host::ContentFacet.create!(
+        :content_view_id => view.id, :lifecycle_environment_id => library.id, :host => host
+      )
+
+      ::Katello::Host::ContentFacet.expects(:propagate_yum_repos).never
+      content_facet.update_bound_repositories([::Katello::Repository.find_by(pulp_id: "Fedora_17")])
+
+      SETTINGS[:katello][:katello_applicability] = nil
```
jlsherrill, line 82: its better to put this in an 'ensure' block so that even if something fails it will still unset the value

ianballou:
@jlsherrill all fixed.

PR #8770, opened 2020-06-15
[CP] 3.16 RC3 cherry picks + version bump

ianballou:


jturel:
I missed triage last week so it's possible there was a discussion around this one, but I don't see how it fits into 3.16 (seems like 4.0 to me) https://github.com/Katello/katello/pull/8756 do you know more about it @ianballou ?

ianballou:
> I missed triage last week so it's possible there was a discussion around this one, but I don't see how it fits into 3.16 (seems like 4.0 to me) #8756 do you know more about it @ianballou ?

I don't remember why, @chris1984 do you remember why we set the issue above to 3.16?

ianballou:
@jturel I've cleaned up the cherry picks.

PR #8772, opened 2020-06-15
Fixes #18035 - filter repos for CV with perms

jlsherrill:
This commit adds permission filtering to associations
within content view create and update.  the idea is that
we can easily spread this to other controllers as well.

Not only are associations filtered, but any array within the 
parameters is checked for permissions on update/create. 
An error is thrown if the correct class to check has not been defined.

### DIFF
--- a/app/controllers/katello/api/v2/api_controller.rb
+++ b/app/controllers/katello/api/v2/api_controller.rb
@@ -5,6 +5,7 @@ module Katello
     include Api::V2::Rendering
     include Api::V2::ErrorHandling
     include ::Foreman::Controller::CsvResponder
+    include Concerns::Api::V2::AssociationsPermissionCheck
 
     # support for session (thread-local) variables must be the last filter in this class
     include Foreman::ThreadSession::Cleaner
--- a/app/controllers/katello/api/v2/content_views_controller.rb
+++ b/app/controllers/katello/api/v2/content_views_controller.rb
@@ -22,6 +22,13 @@ module Katello
       param :solve_dependencies, :bool, :desc => N_("Solve RPM dependencies by default on Content View publish, defaults to false")
     end
 
+    def filtered_associations
+      {
+        :component_ids => Katello::ContentViewVersion,
+        :repository_ids => Katello::Repository
+      }
+    end
+
     api :GET, "/organizations/:organization_id/content_views", N_("List content views")
     api :GET, "/content_views", N_("List content views")
     param :organization_id, :number, :desc => N_("organization identifier")
--- a/app/controllers/katello/concerns/api/v2/associations_permission_check.rb
+++ b/app/controllers/katello/concerns/api/v2/associations_permission_check.rb
@@ -1,0 +1,67 @@
+module Katello
+  module Concerns
+    module Api::V2::AssociationsPermissionCheck
+      extend ActiveSupport::Concern
+
+      # The purpose of this module is to protect a controller from a user creating or updating some association
+      # when they do not have permissions to view the associated items.  An example would be adding random repository ids to
+      # content view.
+      # To support this, within the controller define a method such as:
+      #     def filtered_associations
+      #       {
+      #         :component_ids => Katello::ContentViewVersion,
+      #         :repository_ids => Katello::Repository
+      #       }
+      #     end
+      # This assumes that the parameters are 'wrapped'.  So the above in the content_views_controller, actually looks at
+      #  a subhash of 'content_view'
+
+      included do
+        before_action :check_association_ids, :only => [:create, :update]
+      end
+
+      def check_association_ids
+        if filtered_associations
+          wrapped_params = params[self._wrapper_options.name]
+          find_param_arrays(wrapped_params).each do |key_path|
+            if (model_class = filtered_associations.with_indifferent_access.dig(*key_path))
+              param_ids = wrapped_params.dig(*key_path)
+              filtered_ids = model_class.readable.where(:id => param_ids).pluck(:id)
+              if (unfound_ids = param_ids_missing(param_ids, filtered_ids)).any?
+                fail HttpErrors::NotFound, _("One or more ids (%{ids}) were not found for %{assoc}.  You may not have permissions to see them.") %
+                    {ids: unfound_ids, assoc: key_path.last}
+              end
+            else
+              fail _("Unfiltered params array: %s.") % key_path
+            end
+          end
+        else
+          Rails.logger.warn("#{self.class.name} may has unprotected associations, see associations_permission_check.rb for details.") if ENV['RAILS_ENV'] == 'development'
+        end
+      end
+
+      def filtered_associations
+        #should return {} when supported by all controllers
+        nil
+      end
+
+      def param_ids_missing(param_ids, filtered_ids)
+        param_ids.map(&:to_i).uniq - filtered_ids.map(&:to_i).uniq
+      end
+
+      #returns an array of list of keys pointing to an array in a params hash i.e.:
+      # {"a"=> {"b" => [3]}}  =>  [["a", "b"]]
+      def find_param_arrays(hash = params)
+        list_of_paths = []
+        hash.each do |key, value|
+          if value.is_a?(ActionController::Parameters) || value.is_a?(Hash)
+            list_of_paths += find_param_arrays(value).compact.map { |inner_keys| [key] + inner_keys }
+          elsif value.is_a?(Array)
+            list_of_paths << [key]
+          end
+        end
+        list_of_paths.compact
+      end
+    end
+  end
+end
--- a/test/controllers/api/v2/concerns/associations_permission_check_test.rb
+++ b/test/controllers/api/v2/concerns/associations_permission_check_test.rb
@@ -1,0 +1,78 @@
+# encoding: utf-8
+
+require "katello_test_helper"
+
+module Katello
+  class TestAssociationIdController
+    def initialize(params, filtered_associations)
+      @params = params
+      @filtered_associations = filtered_associations
+    end
+
+    def self.before_action(*_args)
+    end
+
+    include Concerns::Api::V2::AssociationsPermissionCheck
+
+    attr_accessor :filtered_associations
+    attr_reader :params
+
+    def _wrapper_options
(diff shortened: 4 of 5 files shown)


app/controllers/katello/api/v2/api_controller.rb
```diff
@@ -216,5 +217,44 @@ def check_disconnected
             "is prohibited. If you would like to change this, please update the content setting 'Disconnected mode'."
       fail HttpErrors::BadRequest, _(msg) if Setting[:content_disconnected]
     end
+
+    def filtered_associations
+      nil
+    end
+
+    def filter_association_ids
+      if filtered_associations
```
jlsherrill, line 226: the thought is that we would eventually remove this 'if' and just fail if its not defined.

jlsherrill:
still need to write a test for this, but curious what people think about approach

app/controllers/katello/api/v2/api_controller.rb
```diff
@@ -216,5 +217,44 @@ def check_disconnected
...
+        find_param_arrays.each do |key_path|
+          if (model_class = filtered_associations.with_indifferent_access.dig(*key_path))
+            filtered_ids = model_class.readable.where(:id => params.dig(*key_path)).pluck(:id)
```
jlsherrill, line 229: I made this silently ignore items that can't be read, but now i'm wondering if thats what we should do.  I imagine a UI or cli that would be taking a list of associated Ids and adding one item to it.  With the current logic, if the user can't read one of those items it is silently removed from the content view because they couldn't read it.  Does that make the most sense? or does rejecting the entire request if any items in the association list are unreadable make the most sense?

app/controllers/katello/api/v2/api_controller.rb
```diff
@@ -216,5 +217,44 @@ def check_disconnected
...
+            replace_params_value(key_path, filtered_ids)
+          else
+            fail _("Unfiltered params array: %s") % key_path
+          end
+        end
+      end
+    end
+
+    def replace_params_value(key_path, value, hash = params)
```
jturel, line 238: api_controller.rb is pretty beefy. any appetite for moving these methods into a module?

app/controllers/katello/api/v2/api_controller.rb
jturel: I think we should definitely reject the whole request

app/controllers/katello/api/v2/content_views_controller.rb
```diff
@@ -22,6 +22,15 @@ class Api::V2::ContentViewsController < Api::V2::ApiController
       param :solve_dependencies, :bool, :desc => N_("Solve RPM dependencies by default on Content View publish, defaults to false")
     end
 
+    def filtered_associations
```
jturel, line 25: Did you add the filtered_associations here to serve as kind of a documentation? I was going to ask to move it to the private block down below but I noticed it was right next to the content_view param group

app/controllers/katello/api/v2/api_controller.rb
```diff
@@ -216,5 +217,44 @@ def check_disconnected
...
+    def find_param_arrays(hash = params)
+      list_of_paths = []
+      hash.each do |key, value|
+        if value.is_a?(ActionController::Parameters)
+          list_of_paths += find_param_arrays(value).compact.map { |inner_keys| [key] + inner_keys }
+        elsif value.is_a?(Array)
+          list_of_paths << [key]
+        end
+      end
+      list_of_paths.compact
```
jturel, line 257: +1 to defensive programming, but curious if we could actually get nils in this array

app/controllers/katello/api/v2/api_controller.rb
```diff
@@ -216,5 +217,44 @@ def check_disconnected
...
+          end
+        end
+      end
+    end
+
+    def replace_params_value(key_path, value, hash = params)
+      if key_path.length == 1
+        hash[key_path.first] = value
+      else
+        replace_params_value(key_path[1..-1], value, hash[key_path[0]])
```
jturel, line 242: Two recursive methods in one PR!! :lollipop:

app/controllers/katello/api/v2/content_views_controller.rb
jlsherrill: Yeah, i had put it at the top so it would be somewhat obvious, and next tot he apidoc declaration you can easily match up the associations from the apidoc.  I could see it here or down where the 'view_params' method is defined.  IT couldn't be private because its being called from its parents class.  It could be protected though

app/controllers/katello/api/v2/api_controller.rb
jlsherrill: +1

app/controllers/katello/api/v2/api_controller.rb
jlsherrill: all

app/controllers/katello/api/v2/api_controller.rb
jturel: if these are moved to a module then filtered_associations could raise a NotImplementedError

app/controllers/katello/api/v2/content_views_controller.rb
jturel: Cool. I'm good with that

jlsherrill:
Pushed an update. the biggest changes:

* Now reject the request if one id is not readable
* Moved to a concern
* Handles parameters passed in as wrapped or unwrapped.  A test caught this, but now you define the inner part of the wrapped paramters, so in this example:

```
    def filtered_associations
      {
        :component_ids => Katello::ContentViewVersion,
        :repository_ids => Katello::Repository
      }
    end
```

within the controller.  The underlying code uses the wrap_parameter internals to figure out the wrap'd param name to locate things inside the param hash (params[:content_view] in this case.

jturel:
I'm running into an issue while testing this: 
```
$ hammer -u nonadminuser -p changeme content-view add-repository --id=3 --repository-id=4
Could not add repository:
  Field '1' not recognized for searching!
```

This user has viewer perms only. Looks like it could be Authorizer or scoped_search related. Did you run into anything like this? The scoped_search query builder is getting a '1=1' in the search string which I think is invalid since it's not a real field like the error says


```
2020-06-29T12:56:00 [E|app|9130caa3] ScopedSearch::QueryNotSupported: Field '1' not recognized for searching!                                                                                                                                 
 | /home/vagrant/foreman/.vendor/ruby/2.5.0/gems/scoped_search-4.1.8/lib/scoped_search/query_builder.rb:498:in `to_single_field_sql'                                                                                                          
 | /home/vagrant/foreman/.vendor/ruby/2.5.0/gems/scoped_search-4.1.8/lib/scoped_search/query_builder.rb:519:in `to_sql'                                                                       
 | /home/vagrant/foreman/.vendor/ruby/2.5.0/gems/scoped_search-4.1.8/lib/scoped_search/query_builder.rb:54:in `build_find_params'                                                             
 | /home/vagrant/foreman/.vendor/ruby/2.5.0/gems/scoped_search-4.1.8/lib/scoped_search/query_builder.rb:22:in `build_query'                                                                   
 | /home/vagrant/foreman/app/services/authorizer.rb:97:in `build_filtered_scope_components'                                                                                                   
 | /home/vagrant/foreman/app/services/authorizer.rb:56:in `find_collection'                                                                                                                   
 | /home/vagrant/foreman/app/models/concerns/authorizable.rb:56:in `authorized_as'                                                                                                            
 | /home/vagrant/foreman/app/models/concerns/authorizable.rb:94:in `authorized'                                                                                                               
 | /home/vagrant/katello/app/models/katello/authorization/repository.rb:23:in `readable'                                                                                                      
 | /home/vagrant/katello/app/controllers/katello/concerns/api/v2/associations_permission_check.rb:29:in `block in check_association_ids'                                                                                                      
 | /home/vagrant/katello/app/controllers/katello/concerns/api/v2/associations_permission_check.rb:26:in `each'                                                                                
 | /home/vagrant/katello/app/controllers/katello/concerns/api/v2/associations_permission_check.rb:26:in `check_association_ids' 
```

jlsherrill:
Updated with test:

```
[vagrant@centos7-hammer-devel ~]$ hammer -u add -p add  content-view add-repository --organization-id=1 --name=perms  --repository-id=12
The repository has been associated.
[vagrant@centos7-hammer-devel ~]$ hammer -u add -p add  content-view add-repository --organization-id=1 --name=perms  --repository-id=14
Could not add repository:
  One or more ids ([14]) were not found for repository_ids.  You may not have permissions to see them.
```

jlsherrill:
oh but still looking into the association reflection!

jlsherrill:
I don't actually see an easy way to get the class given an arbitrary parameter.  I could make lots of assumptions but i don't see a simple way to go from  ContentView 'repository_ids' =>  to the Repository class without either making a lot of assumptions or making it crazy complex

jturel:
Gave it a retest and it's working great in conjunction with https://github.com/theforeman/foreman/pull/7787

Created a user who could not see a particular repo whose ID is 4

```
[vagrant@centos7-katello-devel foreman (build_scoped_search_condition)]$ hammer -u nonadminuser -p changeme repository list | grep Needed
```

Confusingly, I can still see the repo details because `find_repository` in RepositoriesController doesn't use `.readable` in its query
```
hammer -u nonadminuser -p changeme repository show --id=4
.....
```

I'm unable to add the repo as expected:
```
[vagrant@centos7-katello-devel foreman (build_scoped_search_condition)]$ hammer -u nonadminuser -p changeme content-view add-repository --id=3 --repository-id=4
Could not add repository:
  One or more ids ([4]) were not found for repository_ids.  You may not have permissions to see them.
```

When I remove the filter I can add the repo :1st_place_medal: 

```
[vagrant@centos7-katello-devel foreman (build_scoped_search_condition)]$ hammer -u nonadminuser -p changeme repository list | grep Needed
4  | NeededErrata                                           | WithErrata                      | yum          | https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/

[vagrant@centos7-katello-devel foreman (build_scoped_search_condition)]$ hammer -u nonadminuser -p changeme content-view add-repository --id=3 --repository-id=4
The repository has been associated.
```

PR #8773, opened 2020-06-15
Fixes #26977 - Host sub status resets on build

parthaa:
This commit enables one to ensure that the Host Status is reset to unknown when
the "build" button is set.

### DIFF
--- a/app/models/katello/concerns/host_managed_extensions.rb
+++ b/app/models/katello/concerns/host_managed_extensions.rb
@@ -52,6 +52,9 @@ module Katello
         before_validation :correct_kickstart_repository
         before_update :check_host_registration, :if => proc { organization_id_changed? }
 
+        after_validation :queue_reset_content_host_status
+        register_rebuild(:queue_reset_content_host_status, N_("Content_Host_Status"))
+
         scope :with_pools_expiring_in_days, ->(days) { joins(:pools).merge(Katello::Pool.expiring_in_days(days)).distinct }
 
         scoped_search :relation => :host_collections, :on => :id, :complete_value => false, :rename => :host_collection_id, :only_explicit => true, :validator => ScopedSearch::Validators::INTEGER
@@ -81,6 +84,29 @@ module Katello
         if subscription_facet
           fail ::Katello::Errors::HostRegisteredException
         end
+      end
+
+      def reset_katello_status
+        self.host_statuses.where(type: ::Katello::HostStatusManager::STATUSES.map(&:name)).each do |status|
+          status.update!(:status => status.class.const_get(:UNKNOWN))
+        end
+        self.host_statuses.reload
+        true
+      end
+
+      def reset_content_host_status
+        logger.debug "Scheduling host status cleanup"
+        queue.create(id: "reset_content_host_status_#{id}", name: _("Mark Content Host Statuses as Unknown for %s") % self,
+          priority: 200, action: [self, :reset_katello_status])
+      end
+
+      def queue_reset_content_host_status
+        should_reset_content_host_status? && reset_content_host_status
+      end
+
+      def should_reset_content_host_status?
+        return false unless self.is_a?(::Host::Base)
+        !new_record? && build && self.changes.key?('build')
       end
 
       def correct_kickstart_repository
--- a/test/models/concerns/host_managed_extensions_test.rb
+++ b/test/models/concerns/host_managed_extensions_test.rb
@@ -50,6 +50,22 @@ module Katello
       refute_includes found, other_host
     end
 
+    def test_host_status_reset
+      host = FactoryBot.create(:host)
+      host.host_statuses.delete_all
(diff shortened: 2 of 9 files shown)


test/models/concerns/host_managed_extensions_test.rb
```diff
@@ -43,6 +43,16 @@ def test_full_text_search
       refute_includes found, other_host
     end
 
+    def test_host_status_reset
+      host = FactoryBot.create(:host)
+      host.host_statuses.delete_all
+      Katello::SubscriptionStatus.create!(host: host, :status => Katello::SubscriptionStatus::VALID)
+      host.reload
+      assert_equal Katello::SubscriptionStatus::VALID, host.host_statuses.first.status
+      host.reset_katello_status
```
jturel, line 52: if you set build to true on the host and save rather than calling reset_katello_status then we can test the full behavior

app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -76,6 +79,31 @@ def self.find_with_expiring_pools(_key, _operator, days_from_now)
...
+      end
+
+      def reset_subscription_status
+        logger.debug "Scheduling host status cleanup"
+        queue.create(id: "reset_subscription_status_#{id}", name: _("Mark Subscription, and other Status as Unknown for %s") % self,
+          priority: 200, action: [self, :reset_katello_status])
+      end
+
+      def queue_reset_subscription_status
+        should_reset_subscription_status? && reset_subscription_status
```
jturel, line 99: You've been doing too much react development ;)

app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -76,6 +79,31 @@ def self.find_with_expiring_pools(_key, _operator, days_from_now)
         end
       end
 
+      def reset_katello_status
+        self.host_statuses.where("#{HostStatus::Status.table_name}.type like :query", query: "Katello%").each do |status|
+          if status.class.const_defined?(:UNKNOWN)
```
jturel, line 84: Maybe we should have a requirement which states that all of our statuses have an UNKNOWN?

app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -51,6 +51,9 @@ def update_os_from_facts
         before_save :correct_puppet_environment
         before_validation :correct_kickstart_repository
 
+        after_validation :queue_reset_subscription_status
+        register_rebuild(:queue_reset_subscription_status, N_("Subscription_Status"))
```
jturel, line 55: Some of the references to "subscription status" need to be reworded because we are doing more than that. Something like 'Content Host Status' ?

test/models/concerns/host_managed_extensions_test.rb
parthaa: Hmm I didnt want to deal with the host "queue" in my test case. Will play with it though.

jturel: If it's a pain then don't worry about it

app/models/katello/concerns/host_managed_extensions.rb
parthaa: lol

app/models/katello/concerns/host_managed_extensions.rb
parthaa: Hmm how can we enforce that?

jturel: in plugin.rb we hook into the statuses: https://github.com/Katello/katello/blob/1621c4f1c54253c58d5a27a88df512bbbb42d88f/lib/katello/plugin.rb#L269-L276

We could:

- somewhere (engine.rb?) define a 'statuses' public method to get a list of all the classes
- change plugin.rb to iterate over that to register the statuses
- call that same public method here in order to get the class names for this query (vs using like)
- adjust the test to create one of each of the status classes (this part will enforce that they all provide an UNKNOWN value)

thoughts?

test/models/concerns/host_managed_extensions_test.rb
parthaa: Yeh tried simpler version but its a pain because I dont think the queue runs in a test env when we save?

app/models/katello/host_status_manager.rb
```diff
@@ -0,0 +1,13 @@
+module Katello
```
jturel, line 1: do you think app/services might be a better place for this?

parthaa: sure

jturel:
@parthaa just needs a rebase ... APJ :)

parthaa:
[test katello]

sbernhard:
Should this be cherry-picked to Katello 3.16?

PR #8775, opened 2020-06-16
[CP] Katello 3.15.2

jjeffers:


katello.gemspec
```diff
@@ -31,7 +31,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency "rest-client"
 
   gem.add_dependency "rabl"
-  gem.add_dependency "foreman-tasks", ">= 0.14.1"
+  gem.add_dependency "foreman-tasks", "<= 1.2.0"
```
jturel, line 34: I think you might want this to be < 1.2.0

jturel: See here http://yum.theforeman.org/plugins/2.0/el7/x86_64/

PR #8777, opened 2020-06-16
Fixes #30112 - Add host registration check before changing organization

ianballou:
Hosts need to be unaffiliated from Candlepin before the organization is changed.  Currently, it is allowed to do so, which causes numerous bugs for the host's content facet.

My PR blocks org changes in a couple of places and tells the user to unregister first:
1) When changing the `organization_id` through the API.  For example:
```
echo '{"id": "centos8.cannolo.example.com", "host": {"organization_id": 3} }' | http PUT https://admin:changeme@centos7-katello-devel-2.cannolo.example.com/api/hosts/centos8.cannolo.example.com
```
returns, in response:
```
{
    "error": {
        "message": "Content host must be unregistered first before performing this action."
    }
}
```

2) When using "Assign Organization" from the `/hosts` page.  A message such as "Unregister host centos8.cannolo.example.com before assigning an organization" will appear and block the org update.

To test, try both scenarios above on a registered content host.

### DIFF
--- a/app/controllers/katello/concerns/hosts_controller_extensions.rb
+++ b/app/controllers/katello/concerns/hosts_controller_extensions.rb
@@ -18,6 +18,17 @@ module Katello
       included do
         prepend Overrides
 
+        def update_multiple_taxonomies(type)
+          registered_host = @hosts.detect { |host| host.subscription_facet }
+          unless registered_host.nil?
+            error _("Unregister host %s before assigning an organization") % registered_host.name
+            redirect_back_or_to hosts_path
+            return
+          end
+
+          super
+        end
+
         def destroy
           if Katello::RegistrationManager.unregister_host(@host, :unregistering => false)
             process_success redirection_url_on_host_deletion
--- a/app/lib/katello/errors.rb
+++ b/app/lib/katello/errors.rb
@@ -39,6 +39,12 @@ module Katello
       end
     end
 
+    class HostRegisteredException < StandardError
+      def message
+        _("Content host must be unregistered before performing this action.")
+      end
+    end
+
     class EmptyBulkActionException < StandardError
       def message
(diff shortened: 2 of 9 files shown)


jjeffers:
Tested this as described, first verifying that katello allows organization assignment to registered hosts. Then proceeded to apply the PR, and attempted to assign organization changes to the registered host, which was rejected and the warning displayed in both the UI and the api request, respectively. Works as described!

ianballou:
@jjeffers I had to push some update to make the tests happy.  It wasn't anything major but I think it might be worth a quick look-over before I merge.

ianballou:
[test katello]

app/controllers/katello/concerns/hosts_controller_extensions.rb
```diff
@@ -18,6 +18,17 @@ def action_permission
       included do
         prepend Overrides
 
+        def update_multiple_taxonomies(type)
+          registered_host = @hosts.detect { |host| host.subscription_facet }
+          unless registered_host.nil?
+            error "Unregister host #{registered_host.name} before assigning an organization"
```
jturel, line 24: Formatting strings in this way make translations more challenging. It should be: `_("Unregister host %s before assigning an Organization") % registered_host.name`

ianballou: oh right, I'll update!

app/lib/katello/errors.rb
```diff
@@ -39,6 +39,12 @@ def message
       end
     end
 
+    class HostRegisteredException < StandardError
+      def message
+        _("Content host must be unregistered first before performing this action.")
```
jturel, line 44: `first before` probably could be only `before`

ianballou:
@jturel I've updated

PR #8778, opened 2020-06-17
Fixes #29694 - Handles race condition on pool deletion

parthaa:
This commit handles the race condition where multiple
pool destroy messages  are run on the same pool.

### DIFF
--- a/app/services/katello/candlepin/message_handler.rb
+++ b/app/services/katello/candlepin/message_handler.rb
@@ -81,9 +81,8 @@ module Katello
       end
 
       def delete_pool
-        if pool
-          Rails.logger.info "deleting pool #{pool.id} from Katello"
-          pool.destroy!
+        if Katello::Pool.where(:cp_id => pool_id).destroy_all.any?
+          Rails.logger.info "deleted pool #{pool_id} from Katello"
         end
       end
     end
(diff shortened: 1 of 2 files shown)


test/services/candlepin/message_handler_test.rb
```diff
@@ -136,5 +136,13 @@ def test_delete_pool
       @handler.delete_pool
       assert_empty Katello::Pool.find_by(:cp_id => @pool_id)
     end
+
+    def test_delete_pool_on_null
+      pool = @handler.pool
+      @handler.stubs(:pool).returns(pool).then.returns(nil)
```
jturel, line 142: Could we just delete the pool instead of stub?

app/services/katello/candlepin/message_handler.rb
```diff
@@ -81,9 +81,8 @@ def import_pool(index_hosts = true)
       end
 
       def delete_pool
-        if pool
-          Rails.logger.info "deleting pool #{pool.id} from Katello"
-          pool.destroy!
+        if Katello::Pool.where(:cp_id => pool_id).destroy_all
+          Rails.logger.info "deleting pool #{pool_id} from Katello"
```
jturel, line 85: This should be 'deleted' instead of 'deleting' now

jturel:
@parthaa works well, just two small notes

test/services/candlepin/message_handler_test.rb
parthaa: deleting the pool wont fail it ? The double stub is to simulate the original race condition.

jturel: Fair, but this stub also does nothing now :)

PR #8780, opened 2020-06-18
Fixes #30155 - fix add/remove subscription api response

piyushtiw:
What was done:

Attaching of a subscription via API call does not provide the correct output, it provides the subscription state of the host before the subscription was attached so we do not see if the actual subscription was correctly attached.

Changed response to accommodate the recent changes(either add or remove subscription).

### DIFF
--- a/app/controllers/katello/api/v2/host_subscriptions_controller.rb
+++ b/app/controllers/katello/api/v2/host_subscriptions_controller.rb
@@ -39,7 +39,9 @@ module Katello
       respond_for_index :collection => @collection
     end
 
-    def index_response
+    def index_response(reload_host = false)
+      # Host needs to be reloaded because of lazy accessor
+      @host.reload if reload_host
       entitlements = @host.subscription_facet.candlepin_consumer.entitlements
       subscriptions = entitlements.map { |entitlement| ::Katello::HostSubscriptionPresenter.new(entitlement) }
       full_result_response(subscriptions)
@@ -53,7 +55,7 @@ module Katello
       end
 
       sync_task(::Actions::Katello::Host::AutoAttachSubscriptions, @host)
-      respond_for_index(:collection => index_response, :template => "index")
+      respond_for_index(:collection => index_response(true), :template => "index")
     end
 
     api :DELETE, "/hosts/:host_id/subscriptions/", N_("Unregister the host as a subscription consumer")
@@ -124,7 +126,7 @@ module Katello
       end
 
       sync_task(::Actions::Katello::Host::RemoveSubscriptions, @host, pool_id_quantities.values)
-      respond_for_index(:collection => index_response, :template => "index")
+      respond_for_index(:collection => index_response(true), :template => "index")
     end
 
     api :PUT, "/hosts/:host_id/subscriptions/add_subscriptions", N_("Add a subscription to a host")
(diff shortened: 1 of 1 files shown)


test/controllers/api/v2/host_subscriptions_controller_test.rb
```diff
@@ -101,8 +101,11 @@ def test_add_subscriptions
         assert_equal [1], pools_with_quantities[0].quantities.map(&:to_i)
       end
 
-      post :add_subscriptions, params: { :host_id => @host.id, :subscriptions => [{:id => @pool.id, :quantity => "1"}] }
```
jturel, line 104: Did these pass for you locally when running with ktest? I suspect they would fail because the tests won't actually make API calls to Candlepin (unless we record the VCR cassettes). If they are failing, I really think we can omit these test changes.

jturel:
[test katello]

test/controllers/api/v2/host_subscriptions_controller_test.rb
piyushtiw: I was working still on it, Cause after adding the subscription for host, total count did not change. Shall I remove it and would not test anything for this change at all?

jturel: Yeah I'm OK with no unit test for this one. Whosoever verifies the change can do it manually - and the comment you added explains the change so no one will remove that line later on.

jturel:
[test katello]

app/controllers/katello/api/v2/host_subscriptions_controller.rb
```diff
@@ -220,5 +222,9 @@ def action_permission
         fail ::Foreman::Exception.new(N_("unknown permission for %s"), "#{params[:controller]}##{params[:action]}")
       end
     end
+
+    def remove_or_add_subscriptions_api?
```
jturel, line 226: I see where you're going here and I like it. What about just adding a new parameter:

```
def index_response(reload_host = false)

end
```

And update `add_subscriptions`, `remove_subscriptions`, and `auto_attach` methods to call this method with `true`? Another benefit is that if anyone adds a new endpoint in the controller and they want to call `index_subscriptions` they will have to think about whether the host needs to be reloaded or not - so maybe we'll avoid this kind of bug in the future.

That way we don't need to maintain this logic.

jturel: Looks great. Can you also add the (true) to `auto_attach` ? This method also attaches subscriptions and I suspect it has the same bug

app/controllers/katello/api/v2/host_subscriptions_controller.rb
```diff
@@ -39,7 +39,9 @@ def index
       respond_for_index :collection => @collection
     end
 
-    def index_response
+    # params :reload_host, Boolean, :default => false, :desc => "Host needs to be reloaded because of lazy accessor"
```
jturel, line 42: This format of documentation is specific to our external API but `index_response` is internal to the controller. So I think we can remove this comment or go back to the old one :)

Here's how that style of comment affects the api docs:https://theforeman.org/plugins/katello/3.15/api/apidoc/v2/host_subscriptions.html

jturel:
[test katello]

PR #8783, opened 2020-06-19
Fixes #30161 - Attach subscriptions task failed in finalize

hao-yu:
For some reasons User.current is not set when the after_commit
foreman hook is triggered in the finalize method. The nil
User.current is causing foreman hook to raise an error
because it is not able to set the ENV['FOREMAN_HOOKS_USER'].
This issue seems to be related to the fiber-local and
thread-local storage in the Thread. If I set the User.current
using thread-local then it will work without issue. It will
also work if I remove the transaction block around the finalize
method. After some testing, I found that the easiest way to
prevent this issue is to move the code from the finalize method
to the run method.

### DIFF
--- a/app/lib/actions/katello/host/attach_subscriptions.rb
+++ b/app/lib/actions/katello/host/attach_subscriptions.rb
@@ -28,10 +28,12 @@ module Actions
           end
         end
 
-        def finalize
-          ::Katello::Pool.where(:id => input[:pool_ids]).map { |pool| pool.import_data(false) }
-          host = ::Host.find_by(:id => input[:host_id])
-          host.subscription_facet.import_database_attributes
+        def run
+          ActiveRecord::Base.transaction do
+            ::Katello::Pool.where(:id => input[:pool_ids]).map { |pool| pool.import_data(false) }
+            host = ::Host.find_by(:id => input[:host_id])
+            host.subscription_facet.import_database_attributes
+          end
         end
 
         def rescue_strategy
--- a/app/lib/actions/katello/host/remove_subscriptions.rb
+++ b/app/lib/actions/katello/host/remove_subscriptions.rb
@@ -23,10 +23,12 @@ module Actions
           plan_self(:host_id => host.id, :host_name => host.name, :pool_ids => pool_ids)
         end
 
-        def finalize
-          ::Katello::Pool.where(:cp_id => input[:pool_ids]).map { |pool| pool.import_data(false) }
-          host = ::Host.find_by(:id => input[:host_id])
-          host.subscription_facet.import_database_attributes
+        def run
+          ActiveRecord::Base.transaction do
+            ::Katello::Pool.where(:cp_id => input[:pool_ids]).map { |pool| pool.import_data(false) }
+            host = ::Host.find_by(:id => input[:host_id])
+            host.subscription_facet.import_database_attributes
+          end
         end
 
         def rescue_strategy


jturel:
@hao-yu how can someone interested in testing this reproduce the error that you saw?

hao-yu:
@jturel as I described in the bugzilla and the foreman issue:

Steps to Reproduce:
1. Add any hook script in "/usr/share/foreman/config/hooks/host/managed/after_commit"
2. Restart the satellite
3. Go to the web UI -> Content Hosts -> Attach a subscription to any hosts.

Let me know if you need more information.

jturel:
> @jturel as I described in the bugzilla and the foreman issue:
> 
> Steps to Reproduce:
> 
> 1. Add any hook script in "/usr/share/foreman/config/hooks/host/managed/after_commit"
> 2. Restart the satellite
> 3. Go to the web UI -> Content Hosts -> Attach a subscription to any hosts.
> 
> Let me know if you need more information.

Sorry, I completely missed that. Thanks!

jturel:
@hao-yu sorry for the delay here. I was able to reproduce the problem, but I don't think I agree with the fix. This PR only addresses these specific actions in Katello, whereas the problem appears to be in the interaction between foreman tasks (dynflow) and foreman hooks as you alluded to on the BZ.

I came up with a quick fix which doesn't solve the issue of User.current being nil, but it does solve the problem of an error being raised which affects the task result. What do you think about it? https://github.com/jturel/foreman_hooks/commit/9e77acf163f5de24e22e683303b2673833c9b827

hao-yu:
Hi @jturel . I also thought about fixing jturel/foreman_hooks@9e77acf before submitting this patch but I didn't feel comfortable to have nil current user and not sure about what will be the impact of doing this. In my opinion, this might cause issue to the hook that required a user to do something and also likely to break existing hooks. Therefore, I decided to just fix this particular action. I think this is the only action that affected by this issue so far.

jturel:
@adamruzicka I'm curious if the behavior noted by the PR description indicates anything specific to you? There is definitely some strange (meaning I don't currently understand it) interaction going on between foreman hooks and dynflow. The code here addressed it, but it doesn't seem right. Evidently the KeepCurrentUser middleware doesn't work around the problem either. Not asking for you to deep dive into it, but maybe something is obvious to you!

adamruzicka:
A couple key points:
- Entire finalize phase runs in a single transaction which starts before any step is run and is commited after all steps are run
- `after_commit` hook runs when the transaction is commited. Not sure about this one, but that's my interpretation
- `KeepCurrentUser` middleware wraps each step separately. Before a step is executed it restores the user and when the step is done it unsets the user again.

So for example with an action with finalize phase with a single step:
1) Finalize phase starts
2) Transaction begins
3) KeepCurrentUser restores the current user
4) the step runs and makes some changes to an ActiveRecord object
5) KeepCurrentUser unsets the current user
6) Transaction from 2 is commited
7) Hooks are run, but current user is already unset

Does that make sense?

hao-yu:
I think "KeepCurrentUser" is unrelated to this bug. This method is to temporary replace the original User.current with the user who run the task (from task input) and then restore it after running the step. We don't need it in this action.

The issue that we are having here is the original "User.current" becoming nil when running the foreman hook triggered by after_commit.

As I mentioned in the description, this issue is fixed when I remove the transaction block around the finalize or when I set the User.current using thread-local. Thus, I guess it might be something causing the foreman-hook to run outside of the fiber-local scope or something.

adamruzicka:
> This method is to temporary replace the original User.current with the user who run the task (from task input) and then restore it after running the step. We don't need it in this action.

Well, you need some user to be set, right? Inside the executor `User.current` is `nil` while not running any step. So the "restore after running the step" sets it back to `nil`.

> The issue that we are having here is the original "User.current" becoming nil when running the foreman hook triggered by after_commit.

We are saying the same thing but with different words. The hook runs at the end of the phase, after the current user was already "restored" (unset).

> this issue is fixed when I remove the transaction block around the finalize

With this the changes were commited inside the step, while the current user was still set by the middleware. This leads me to believe this is an issue with current user being unset too early and not fiber local thingy issue. Or at least I hope, 

> when I set the User.current using thread-local

Where and how did you do this?

hao-yu:
> Well, you need some user to be set, right? Inside the executor User.current is nil while not running any step. So the "restore after running the step" sets it back to nil.

Yes, I think you are right. I just realized that "KeepCurrentUser" is loaded by the engine by default. I believe "KeepCurrentUser unsets the current user" to nil is the root cause because I am able to fix the issue by commenting out https://github.com/theforeman/foreman-tasks/blob/master/lib/foreman_tasks/engine.rb#L111 and then set the User.current manually in https://github.com/Katello/katello/blob/master/app/lib/actions/katello/host/attach_subscriptions.rb#L35

hao-yu:
Probably need to adjust the order of the middleware or something? IIRC the transaction middleware is loaded first so it is the most outside block.

adamruzicka:
Edit the order so KeepCurrentUser is called before the transaction one and make it span the entire phase instead of step execution.

adamruzicka:
@hao-yu could you check if https://github.com/theforeman/foreman-tasks/pull/592 helps?

hao-yu:
@adamruzicka Yes. I tried it in Satellite 6.7 and it fixed the issue.

jturel:
@hao @adamruzicka  It seems like the next step is to close this PR and associate the foreman-tasks PR with the downstream BZ?

hao-yu:
@jturel yeah, agree

PR #8785, opened 2020-06-22
Fixes #30184 - Add set_dependent_action :destroy for Hostgroup::ContentFacet

jturel:
Errors like this were getting raised if the hostgroup had a content facet (lce, cv, kickstart, or content source are set):

```
PG::ForeignKeyViolation: ERROR: update or delete on table "hostgroups" violates foreign key constraint "fk_rails_434951f902" on table "katello_hostgroup_content_facets" DETAIL: Key (id)=(4) is still referenced from table "katello_hostgroup_content_facets".
```

### DIFF
--- a/app/models/katello/concerns/hostgroup_extensions.rb
+++ b/app/models/katello/concerns/hostgroup_extensions.rb
@@ -5,6 +5,7 @@ module Katello
 
       included do
         before_save :add_organization_for_environment
+
         has_one :kickstart_repository, :through => :content_facet
         has_one :content_source, :through => :content_facet
         has_one :content_view, :through => :content_facet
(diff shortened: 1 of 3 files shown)


jturel:
@ShimShtein  any concerns with this?

ShimShtein:
It should be done automatically here: https://github.com/theforeman/foreman/blob/develop/app/models/concerns/facets/model_extensions_base.rb#L83

ShimShtein:
I think you are missing `set_dependent_action :destroy`, as it is defined in https://github.com/theforeman/foreman/blob/develop/developer_docs/how_to_create_a_plugin.asciidoc#how-to-register-a-facet

jturel:
> I think you are missing `set_dependent_action :destroy`, as it is defined in https://github.com/theforeman/foreman/blob/develop/developer_docs/how_to_create_a_plugin.asciidoc#how-to-register-a-facet

That did the trick. Thanks!

ShimShtein:
ACK from code point of view. Still need testing though

jturel:
> ACK from code point of view. Still need testing though

ty! it's on our board so someone on my team should test

jturel:
> When I try to delete my host group, it says "My host group is used by katello-client.cannolo.example.com". This is definitely better than the error that was thrown before, but the host group still isn't deleted. Is that what you're expecting?

that's valid.

create a new hostgroup on master which isn't associated to any hosts and give it a lifecycle environment for example. The delete will fail with a foreign key error, and this pr fixes that!

PR #8787, opened 2020-06-22
Katello translation updates

ianballou:


jturel:
I *think* this will be green once the rubocop lint is fixed

PR #8790, opened 2020-06-26
Fixes #30227 - Update tests to use fixtures.pulpproject.org

daviddavis:
https://repos.fedorapeople.org/pulp/pulp/fixtures/ is being taken down.


daviddavis:
So... uh anyone with a Katello box interested in re-recording these VCR cassettes?

PR #8792, opened 2020-06-29
Fixes #29689 - check that the HttpProxy table exists 

jjeffers:
This PR adds a check for the HttpProxy table before setting validations. Otherwise, it's possible that the validation will throw an error because the model itself isn't present in the schema. This possibility increases as seen in scenarios with an external database for the foreman schema.

This also may help eliminate some occasional test failures seen in CI.


chris1984:
[test katello]

PR #8795, opened 2020-06-29
Fixes #30239 - Fix breaking changes for patternfly upgrade

johnpmitsch:
@patternfly/react-core is being upgrade from v3 to v4 and there are major version changes. Note that both versions are considered "patternfly 4", the version of the design library, its just the react package that is updating with some more new breaking changes.

When https://github.com/theforeman/foreman-js/pull/142 is merged and released, `npm update` in Foreman and test the content view page at `labs/content_views` works. The subscription and rh-repos pages should not be affected by this change, but its worth clicking through those to make sure.

Update: this was merged, run `npm update` in Foreman to test

### DIFF
--- a/webpack/components/Table/TableWrapper.js
+++ b/webpack/components/Table/TableWrapper.js
@@ -1,5 +1,5 @@
 import React, { useEffect, useState } from 'react';
-import { Pagination, Flex, FlexItem, FlexModifiers } from '@patternfly/react-core';
+import { Pagination, Flex, FlexItem } from '@patternfly/react-core';
 
 import PropTypes from 'prop-types';
 import { useDispatch } from 'react-redux';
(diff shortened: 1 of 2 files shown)


johnpmitsch:
[test katello]

webpack/components/Table/TableWrapper.js
```diff
@@ -53,7 +53,7 @@ const TableWrapper = ({
         <FlexItem>
           <Search patternfly4 {...{ onSearch, getAutoCompleteParams }} />
         </FlexItem>
-        <FlexItem breakpointMods={[{ modifier: FlexModifiers['align-right'] }]}>
+        <FlexItem align={{ default: 'alignRight' }}>
```
jeremylenz, line 56: oh I bet this is what's causing the Jenkins test failure on the other PR..

johnpmitsch: yessir! this would be it

PR #8798, opened 2020-07-01
Refs #29925 - Allow the user to toggle Simple Content Access on/off

piyushtiw:
Update the Manage Manifest screen to allow the toggle simple content access


jturel:
[test katello]

webpack/scenes/Subscriptions/Manifest/ManifestConstants.js
```diff
@@ -14,6 +14,10 @@ export const MANIFEST_HISTORY_REQUEST = 'MANIFEST_HISTORY_REQUEST';
 export const MANIFEST_HISTORY_SUCCESS = 'MANIFEST_HISTORY_SUCCESS';
 export const MANIFEST_HISTORY_FAILURE = 'MANIFEST_HISTORY_FAILURE';
 
+export const MANIFEST_SCA_TOGGLE_REQUEST = 'MANIFEST_SCA_TOGGLE_REQUEST';
```
jeremylenz, line 17: As a new name for these actions (moving away from Toggle) I suggest `SET_SIMPLE_CONTENT_ACCESS_REQUEST` etc.

webpack/scenes/Subscriptions/Manifest/ToggleManifestSCA.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Switch } from '@patternfly/react-core';
+
+class SimpleSwitch extends React.Component {
```
jeremylenz, line 5: This is a new component so I would prefer it to use Hooks  :smile:

webpack/scenes/Subscriptions/Manifest/ToggleManifestSCA.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Switch } from '@patternfly/react-core';
```
jeremylenz, line 3: I think this will be our first PF4 component on a PF3 page.  Nice!  I'm on board, but want to double check with @johnpmitsch - do you foresee any issues here?

app/controllers/katello/api/v2/upstream_subscriptions_controller.rb
```diff
@@ -77,6 +77,19 @@ def ping
       render json: { status: 'OK' }
     end
 
+    api :PUT, "/organizations/:organization_id/toggle_simple_content_access",
+      N_("Toggle simple sontent access for a manifest")
+    param :organization_id, :number, :desc => N_("Organization ID"), :required => true
+    def toggle_simple_content_access
```
jturel, line 83: what do you think about creating two APIs:

`/organizations/:id/simple_content_access/enable` and `organizations/:id/simple_content_access/disable` ?

app/services/katello/toggle_simple_content_access.rb
```diff
@@ -0,0 +1,47 @@
+module Katello
+  class ToggleSimpleContentAccess
```
jturel, line 2: This is a good abstraction, but it would be more appropriate I think to create an 'Action' class that we can execute from the controller, which itself will 'plan' the manifest refresh.

I think it could live at `app/lib/actions/katello/organization/simple_content_acces/{enable,disable}.rb` (two distinct actions)

app/controllers/katello/api/v2/upstream_subscriptions_controller.rb
```diff
@@ -77,6 +77,19 @@ def ping
       render json: { status: 'OK' }
     end
 
+    api :PUT, "/organizations/:organization_id/toggle_simple_content_access",
+      N_("Toggle simple sontent access for a manifest")
+    param :organization_id, :number, :desc => N_("Organization ID"), :required => true
```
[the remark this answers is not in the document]
jeremylenz: update: I like @jturel 's solution better, with different endpoints for enable and disable.  Then the param won't be needed.

webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
```diff
@@ -243,6 +245,17 @@ class ManageManifestModal extends Component {
                             >
                               <Icon type="pf" name="info" />
                             </OverlayTrigger>
+                            <div id="manifest-toggle-sca-switch">
+                              <SimpleSwitch
+                                toggleSCA={this.props.toggleSCAManifest}
+                                hideModal={this.hideModal}
+                                simpleContentAccess={simpleContentAccess}
+                                canToggleSimpleContentAccess={
```
jturel, line 253: You'll have to check but I think `canToggleSimpleContentAccess` can be set like this: `canToggleSimpleContentAccess={disableManifestActions}` depending on how disableManifestActions is defined

webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
```diff
@@ -243,6 +245,17 @@ class ManageManifestModal extends Component {
                             >
                               <Icon type="pf" name="info" />
                             </OverlayTrigger>
+                            <div id="manifest-toggle-sca-switch">
+                              <SimpleSwitch
```
jturel, line 249: This is weird. I think it should reference your react component name

webpack/scenes/Subscriptions/Manifest/ToggleManifestSCA.js
jturel, line 5: A better name for this component (and its file) would be `SimpleContentAccessToggle`. We don't need to use 'manifest' since we are already in a folder called /Manifest/

webpack/scenes/Subscriptions/Manifest/ToggleManifestSCA.js
johnpmitsch: as long as it doesn't look too out of places, I don't see an issue with it :+1:

webpack/scenes/Subscriptions/Manifest/ToggleManifestSCA.js
```diff
@@ -0,0 +1,39 @@
...
+
+    return (
+      <Switch
+        id="simple-switch"
+        isChecked={isChecked}
+        onChange={this.handleChange}
+        isDisabled={!this.props.canToggleSimpleContentAccess}
+      />
+    );
+  }
```
johnpmitsch, line 29: If this is just the Switch, any reason to create a separate component for it?  Seems like this could all be done in the parent component.

If you are concerned about the size of `ManageManifestModal`, I'm all for refactoring to use smaller components or breaking a part out, but I think we should look at larger groups of components to abstract out.

piyushtiw: My reasoning behind doing it in a separate component is, The responsibility of ManageManifestModal is showing data related to manifest in a modal while doing actions on manifest(e.g. toggling simple content action) should be done in separate component as it has its own state. Any thoughts?

johnpmitsch: I think we could handle the state fine in the parent component, if it gets complex a separate component is fine, I'm just not sure its doing more than shuffling props to Switch, and I don't see it as a reusable component (for example if there was custom CSS to make the switch look different, that could justify breaking it out to be re-used again).

Here is my suggestion for keeping it in `ManageManifesModal`: 
```javascript
// Add to state in constructor
this.state = {
  contentAccessChecked: props.simpleContentAccess
}

// Inside render
<Switch
 id="content-access-toggle-switch"
 isChecked={contentAccessChecked}
 onChange={() => { 
   this.setState({ contentAccessChecked });
   simpleContentAccess ? disableSimpleContentAccess() : enableSimpleContentAccess();
   this.hideModal();
 }}
  // double check my logic here, any of these things would mean the switch is disabled? 
  isDisabled={ disableManifestActions || !manifestExists(organization) || actionInProgress} />
```

If the above achieves the same thing as what you are doing now, I don't see a reason to break this out to another component. I'm open to other opinions from @jturel and @jeremylenz too!

piyushtiw: Thanks @johnpmitsch for such a detailed explanation. I will do the same as you mentioned.

piyushtiw: @johnpmitsch what about extracting all the data related to simple content access to newer component(suggested by Jturel? It will refactor the `ManageManifesModal` in a larger group also ?

jturel: I think that's the way to go: moving the rest of the 'stuff' currently on the ManageManifest modal related to this SCA toggle into the component itself. That way we are much closer to having a truly reusable component that we can drop on any page.

johnpmitsch: +1 to refactoring a larger group of components out of ManageManifestModal, that would be great :+1:

app/controllers/katello/api/v2/upstream_subscriptions_controller.rb
```diff
@@ -77,6 +77,22 @@ def ping
       render json: { status: 'OK' }
     end
 
+    api :PUT, "/organizations/:organization_id/enable_simple_content_access",
```
jturel, line 80: I think these endpoints should be organized like this `/organization/:id/simple_content_access/enable` rather then `enable_simple_content_access` - it's more consistent with our other APIs

app/lib/actions/katello/organization/simple_content_access/toggle.rb
```diff
@@ -0,0 +1,38 @@
...
+
+          SIMPLE_CONTENT_ACCESS_DISABLED_VALUE = "entitlement".freeze
+          SIMPLE_CONTENT_ACCESS_ENABLED_VALUE = "org_environment".freeze
+
+          attr_reader :organization
+
+          def plan(organization_id)
+            @organization = ::Organization.find(organization_id)
+
+            fail N_('Not allowed to toggle Simple content access') unless consumer
```
jturel, line 16: To get the upstream consumer there is also the option of `Katello::Resources::Candlepin::UpstreamCandlepinResource.upstream_consumer` which will raise an error if there is no manifest imported like we have in other areas - so we wouldn't need a new message here

johnpmitsch:
@piyushtiw The new React component abstraction looks great! 

You'll have to add some more tests. Unfortunately the `ManageManifestModal` component is using [shallow rendering](https://github.com/Katello/katello/blob/bab3adfd6b7b599e08512bb34a9a13752cb088b3/webpack/scenes/Subscriptions/Manifest/__tests__/ManageManifestModal.test.js#L15) which doesn't render child components, including the new `SimpleContentAccess`.

This exact scenario is why I proposed we move away from enzyme snapshots and to react-testing-library functional testing [here](https://community.theforeman.org/t/using-react-testing-library-in-katello/18120), which is now available to use. IMHO a refactoring like this should not need testing updates other than the ones for new functionality and the existing tests should be confirmation that there are no regressions introduced by refactoring a component into multiple ones.

But we don't have react-testing-library tests for this area so your options are:

A) Add `react-testing-library` tests for only the functionality you introduced. 

I would recommend adding them for `ManageManifestModal`, which will be easier than for `SubscriptionPage`. Since react-testing-library renders all child components, it will render `SimpleContentAccess` too. I would suggest something like this:
- Render `ManageManifestModal`
- find toggle switch
- click toggle switch
- assert proper API call is made

There are some examples [here](https://github.com/Katello/katello/pull/8808/files#diff-ae55c0abb5d8463717cd4dba33095e9f) and [here](https://github.com/Katello/katello/blob/master/webpack/scenes/ContentViews/__tests__/contentViewPage.test.js) and I'm available for questions and/or pair-programming :smile:

OR

B) Add enzyme snapshot tests for `SimpleContentAccess`, similar to `ManageManifestModal` tests

I would prefer we do A as that is the new method of testing React functionality and they can be expanded upon in the future, but I can understand the argument to add snapshot tests as that is what is there pre-existing.

webpack/scenes/Subscriptions/Manifest/SimpleContentAccess.js
```diff
@@ -0,0 +1,85 @@
...
+    enableSimpleContentAccess,
+    disableSimpleContentAccess,
+    setActionInProgress,
+  } = props;
+
+  const [isChecked, setChecked] = useState(simpleContentAccessValue);
+
+  useEffect(() => {
+    setChecked(simpleContentAccessValue);
+  });
```
jeremylenz, line 20: Should add a dependency array so this effect won't run after every render, but will only run when that prop changes:

```suggestion
  useEffect(() => {
    setChecked(simpleContentAccessValue);
  }, [simpleContentAccessValue]);
```

webpack/scenes/Subscriptions/Manifest/SimpleContentAccess.js
```diff
@@ -0,0 +1,85 @@
...
+    } else {
+      enableSimpleContentAccess();
+    }
+  }
+
+  function handleChange() {
+    setChecked(!isChecked);
+    setActionInProgress();
+    toggleSimpleContentAccess();
+    props.hideModal();
```
jeremylenz, line 34: This happens too fast and it's a bit jarring.  I click the switch, the modal disappears immediately and all of a sudden I'm refreshing the manifest.

I would be in favor of waiting to hide the modal for a bit, just so the user can at least see the switch toggle..  maybe run `hideModal()` after a `setTimeout` of 500ms?  @johnpmitsch thoughts?

webpack/scenes/Subscriptions/Manifest/SimpleContentAccess.js
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  return (
+    <Fragment>
+      {simpleContentAccessValue ? __('Yes') : __('No')}
```
jeremylenz, line 39: Maybe @johnpmitsch 's `boolToYesNo()` should become a helper function :smile:

webpack/scenes/Subscriptions/Manifest/SimpleContentAccess.js
```diff
@@ -0,0 +1,85 @@
...
+        <Switch
+          id="simple-switch"
+          isChecked={isChecked}
+          onChange={handleChange}
+          isDisabled={!canToggleSimpleContentAccess}
+        />
+        <OverlayTrigger
+          overlay={
+            <Tooltip id="sca-tooltip">
+              {__('Toggling Simple Content Access will refresh your manifest.')}
```
jeremylenz, line 62: sca_toggle

Having two info icons sandwiching the switch just looks too funny to me.  For the `Toggling Simple Content Access will refresh your manifest.`, I was envisioning regular text that's on the screen all the time, below the switch and above the Choose File button.

Also, to me the remaining info icon (line 42 above) makes more sense next to the text "Simple Content Access," not next to the switch.

webpack/scenes/Subscriptions/Manifest/SimpleContentAccess.js
jeremylenz: come to think of it, `setChecked` should probably also be in the deps array

webpack/scenes/Subscriptions/Manifest/SimpleContentAccess.js
johnpmitsch: It is a bit jarring, at least the switch indicates an action though. The upload manifest workflow is a bigger concern IMO as it uploads as soon as you select a file.

If we can figure out a way to improve it we can do that, I'm not sure we should set a timeout though. It could be worth asking UX.

I'm really fine with how it is now though, I think its going to be hard to improve in isolation without refactoring all the loading/async workflows.

johnpmitsch:
@piyushtiw A few asthetic comments:

- In the UI, why have the "Yes" and the switch? The switch is already indicating if it is turned on or off. Can we remove the Yes/No?
- the "Simple Content Access" label should be bold like the other labels. Can we get it on one line too?
- Can we put Simple Content Access on top of USMA (whatever that stand for), that seems to be showing the manifest name which is more related to the upload file buttons, so they should be grouped together.
- Can we add more spacing between USMA, simple content access, and the upload file buttons? This appears as one big mangled group and its hard to determine what part on the right is related to the part on the left.

johnpmitsch:
> * Can we add more vertical spacing between USMA, simple content access, and the upload file buttons? This appears as one big mangled group and its hard to determine what part on the right is related to the part on the left.

pf4 has a `TextList` option for lists like this. I'm not sure if it will play nice with the pf3 stuff here, but you could think about refactoring the `Subscription Manifest` fields to use this. It is being used in the [new CV page](https://github.com/Katello/katello/pull/8808/files#diff-9bdfb8248e1486424edda8b71d0da183) and it helps align the labels and values of the fields and adds spacing.

piyushtiw:
> > * Can we add more vertical spacing between USMA, simple content access, and the upload file buttons? This appears as one big mangled group and its hard to determine what part on the right is related to the part on the left.
> 
> pf4 has a `TextList` option for lists like this. I'm not sure if it will play nice with the pf3 stuff here, but you could think about refactoring the `Subscription Manifest` fields to use this. It is being used in the [new CV page](https://github.com/Katello/katello/pull/8808/files#diff-9bdfb8248e1486424edda8b71d0da183) and it helps align the labels and values of the fields and adds spacing.

I used TextList but it needs lots of css changes so I made some changes in the same code and it is looking loke this
Screenshot from 2020-07-09 22-30-02

johnpmitsch:
> I used TextList but it needs lots of css changes so I made some changes in the same code and it is looking loke this

Nice! Looking cleaner.

What do you think about switching USMA and Simple Content Access rows? The USMA seems more related to the choose file buttons.

adjacent question: @jturel should we update the `USMA` label, what do you think that stands for and is that something users would be expected to know?

johnpmitsch:
@piyushtiw patternfly design guidelines recommend putting the tooltip on the field label instead of the value, can we move the tooltip there for Simple Content Access? https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/embedded-assistance#popovers

jturel:
> adjacent question: @jturel should we update the `USMA` label? what do you think that stands for and is that something users would be expected to know?

I raised this once before and I'm having a hard time remembering what the justification was for keeping it (this was back when we built the page), but I really dislike it.

USMA stands for "Upstream Subscription Management Allocation" iirc, and there is no chance that any user knows what it means.

I think it should just be called "Subscription Allocation" since that's what it's referred to as in the portal

jturel:
I think we should remove the :heavy_check_mark: from inside the toggle so that we're consistent with the UX on the Portal

webpack/test-utils/react-testing-lib-wrapper.js
```diff
@@ -22,26 +23,29 @@ process.env.DEBUG_PRINT_LIMIT = 999999;
 function renderWithRedux(
   component,
   {
-    namespace, // redux namespace
-    initialState = { response: {}, status: STATUS.RESOLVED },
+    apiNamespace, // namespace if using API middleware
```
johnpmitsch, line 26: It actually doesn't seem to be affected, but for consistency can you update [here](https://github.com/Katello/katello/blob/master/webpack/scenes/ContentViews/__tests__/contentViewPage.test.js#L18) to `apiNamespace`?

webpack/scenes/Subscriptions/Manifest/__tests__/SimpleContentAccess.test.js
```diff
@@ -0,0 +1,105 @@
...
+
+  await waitFor(() => { expect(toggleButton).toBeInTheDocument(); });
+  expect(toggleButton.checked).toEqual(true);
+
+  fireEvent.click(toggleButton);
+
+  expect(toggleButton.checked).toEqual(false);
+
+  assertNockRequest(getscope, done);
+  assertNockRequest(updatescope, done);
```
johnpmitsch, line 104: For both of these tests, just pass the `done` callback to the last API call you expect to happen.  In this case it would be:
```javascript
  assertNockRequest(getscope);
  assertNockRequest(updatescope, done);
```
Its not a huge deal to pass it earlier, but it makes sure jest waits for the response to finish before moving on to the next test.

johnpmitsch:
Looks like Simple Content Access and the USMA isn't properly aligned, also the tooltip is on a new line:
here

> What do you think about switching USMA and Simple Content Access rows? The USMA seems more related to the choose file buttons.

What do you think about updating this as well?

webpack/scenes/Subscriptions/Manifest/__tests__/SimpleContentAccess.test.js
```diff
@@ -0,0 +1,105 @@
+import React from 'react';
+import { cleanup } from '@testing-library/react';
+import { renderWithRedux, fireEvent, waitFor } from 'react-testing-lib-wrapper';
```
johnpmitsch, line 3: Update these to `patientlyWaitFor` now with https://github.com/Katello/katello/commit/883baf75db86703a3c5efa63baca17e6388c46ea merged

webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
```diff
@@ -107,13 +112,19 @@ class ManageManifestModal extends Component {
...
 
+    const { actionInProgress } = this.state;
+
     const showRedHatProviderDetails = canEditOrganizations;
     const showSubscriptionManifest = (canImportManifest || canDeleteManifest);
     const showManifestTab = (showRedHatProviderDetails || showSubscriptionManifest);
-
-    const { actionInProgress } = this.state;
+    const disableSCASwitch = (disableManifestActions ||
+      !manifestExists(organization) ||
```
jeremylenz, line 125: Please rebase on latest master and update this to use the new Redux selector `selectIsManifestImported` from `OrganizationSelectors.js` (see [comment](https://github.com/Katello/katello/pull/8810#discussion_r453858147) on #8810) instead of `manifestExists`

webpack/scenes/Subscriptions/Manifest/SimpleContentAccess.js
```diff
@@ -0,0 +1,103 @@
...
+  const [isChecked, setChecked] = useState(simpleContentAccessValue);
+
+  useEffect(() => {
+    if (isChecked !== simpleContentAccessValue) {
+      setChecked(simpleContentAccessValue);
+    }
+  }, [simpleContentAccessValue]);
+
+  const toggleSimpleContentAccess = () => {
+    if (simpleContentAccessValue) {
```
jeremylenz, line 27: I think this prop should be named `isEnabled` or something similar, so that this line becomes more self-explanatory

webpack/scenes/Subscriptions/Manifest/SimpleContentAccess.js
```diff
@@ -0,0 +1,103 @@
...
+            placement="bottom"
+            trigger={['hover', 'focus']}
+            rootClose={false}
+          >
+            <span>{__('Simple Content Access')}</span>
+          </OverlayTrigger>
+          <OverlayTrigger
+            overlay={
+              <Tooltip id="sca-refresh-tooltip">
+                {__('Toggling Simple Content Access will refresh your manifest.')}
```
jeremylenz, line 67: I still think this particular text should be a regular span, not inside a tooltip.

webpack/scenes/Subscriptions/Manifest/SimpleContentAccess.js
```diff
@@ -0,0 +1,103 @@
...
+          className="control-label"
+          htmlFor="usmaFile"
+          style={{
+            paddingTop: '0',
+            position: 'absolute',
+          }}
+        >
+          <OverlayTrigger
+            overlay={
+              <Tooltip id="sca-tooltip">
```
jeremylenz, line 54: Seems very weird to have a tooltip over a span, without the info icon.  I think the info icon should be restored here.

webpack/scenes/Subscriptions/Manifest/SimpleContentAccess.js
```diff
@@ -0,0 +1,103 @@
...
+    hideModal,
+  } = props;
+
+  const [isChecked, setChecked] = useState(simpleContentAccessValue);
+
+  useEffect(() => {
+    if (isChecked !== simpleContentAccessValue) {
+      setChecked(simpleContentAccessValue);
+    }
+  }, [simpleContentAccessValue]);
```
jeremylenz, line 24: This still feels off to me, because we're trying to keep props in sync with internal state.

Here's how `simpleContentAccessValue` currently affects the state of our switch:
1. Get the value via prop.
2. Set the state to the Prop value.
3. If the prop changes, update the state.
4. If the state changes (due to user clicking), call `disableSimpleContentAccess` or `enableSimpleContentAccess`, not based on the state, but based on the current prop value.  This gives the advantage of immediate change in response to user action, but also gives us a period of time (while the task is finishing) where the state is out of sync with props and the visual state of the switch is not accurate.

After some more thinking, want to get your thoughts on this possible alternative:

1. The Switch will not toggle immediately based on user clicks.  Instead, when the user clicks, the switch will disable without changing value and we will show a small loading spinner.
2. The loading spinner will remain (with the disabled switch) for as long as the task takes to complete.
3. When the task is complete, the loading spinner will clear and the switch state will change to reflect the completion of the task.

(The modal could still be closed for 2 and 3 if we decide to keep it that way)

This way, the state of the switch always reflects our best knowledge, and there is never a moment when anything is inaccurate.  In addition, we still get the immediate UI response upon user click.

jeremylenz: This approach will also simplify this component, since you won't need the `useEffect` to track the prop anymore.

jeremylenz: update: It looks like much of this is already implemented (the loading spinner is there).  The missing piece is that the switch shouldn't change value until the task is complete

webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
```diff
@@ -198,52 +186,32 @@ class ManageManifestModal extends Component {
...
                       <Grid>
-                        <h5>{__('Subscription Manifest')}</h5>
+                        <h3>{__('Subscription Manifest')}</h3>
                         <hr />
                         <Row>
-                          <Col sm={4}>
+                          <Col sm={5}>
                             <ControlLabel
                               className="control-label"
                               htmlFor="usmaFile"
```
jeremylenz, line 195: This `<ControlLabel>` creates an html `<label>` element, and the `htmlFor` binds it to the form control having the ID `"usmaFile"`.  The effect is that when I click the words "Subscription Allocation," the file open dialog shows.

To fix it I think this can be moved to line 216 right before the `<FormControl>`.

jeremylenz:
@jturel @johnpmitsch I think this is very close!  Can you take another look and see if you have any additional concerns?

piyushtiw:
Screenshot from 2020-07-21 13-32-16

In my Screen It is looking like the above. Still I have added CSS to keep the text left align.

jeremylenz:
per IRC convo, turns out I had my font size set to 20 in Chrome settings.

webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
```diff
@@ -198,52 +186,36 @@ class ManageManifestModal extends Component {
...
+                            isSimpleContentAccessEnabled={simpleContentAccess}
+                            canToggleSimpleContentAccess={!disableSCASwitch}
+                          />
+                          <Col sm={5}>
+                            {canImportManifest &&
+                              <ControlLabel
+                                className="control-label"
+                                style={{ paddingTop: '10 px' }}
+                              >
+                                <div>{__('Upload new manifest')}</div>
```
jeremylenz, line 213: I'm realizing this should probably say "Import new manifest" to match the empty state on the subscriptions page.  Also it's more accurate since it's not really an upload.  (sorry for the one-off)

piyushtiw: > I'm realizing this should probably say "Import new manifest" to match the empty state on the subscriptions page. Also it's more accurate since it's not really an upload. (sorry for the one-off)

If the manifest is already there we can upload the file and import new manifest but it will not enable the new manifest. So if the manifest is already there should we disable the choose new file option?

jeremylenz: @piyushtiw No, I think you can import an updated manifest over an existing one as long as the new one is not older than the existing one (right @jturel ?)

jturel: yep, uploading a new manifest will replace the old one

webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
```diff
@@ -108,13 +87,22 @@ class ManageManifestModal extends Component {
       isManifestImported,
       canEditOrganizations,
       simpleContentAccess,
+      enableSimpleContentAccess,
+      disableSimpleContentAccess,
+      taskInProgress,
+      manifestActionInProgress,
     } = this.props;
 
+    const actionInProgress = (taskInProgress || manifestActionInProgress);
```
jturel, line 96: What's the difference between `taskInProgress` and `manifestActionInProgress`. High chance they're the same.

webpack/scenes/Subscriptions/Manifest/SimpleContentAccess.js
```diff
@@ -0,0 +1,85 @@
...
+  const toggleSimpleContentAccess = () => {
+    if (isSimpleContentAccessEnabled) {
+      disableSimpleContentAccess();
+    } else {
+      enableSimpleContentAccess();
+    }
+  };
+
+  const handleChange = () => {
+    toggleSimpleContentAccess();
```
jturel, line 26: I think this conversation was touched upon so let's chat about it before making the change I am about to suggest: When I toggle the modal immediately disappears, there is a moment where nothing happens, and then the progress bar modal appears. If we use 'await' here, then it should behave a little more smoothly. The modal will close once the task has been started, and the modal should pop up at that time because the reducer will have put it into the state.

webpack/scenes/Subscriptions/Manifest/SimpleContentAccess.js
```diff
@@ -0,0 +1,85 @@
...
+    hideModal();
```
jturel, line 27: I don't think it makes sense for this component to be aware of the modal. It might not even be used in the context of a modal. Since the task from toggling is put into the state of the subscriptions page, then the modal should be closed when that task reaches the props.

jturel:
It looks really good and I left a few notes.

webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
piyushtiw: The `taskInProgress` prop will have true value when we will get the success response of any manifest Action(I.e. Refresh, upload, delete, Enable/Disable SCA). While, when the action request is in progress `manifestActionInProgress` will have true value. So if we will get failure from any Manifest Action request, Only  `manifestActionInProgress` value will get change.

webpack/scenes/Subscriptions/Manifest/SimpleContentAccess.js
piyushtiw: We can hide the modal when the task starts, In between the modal can be remain open and we can disable all the buttons.

johnpmitsch:
@piyushtiw It looks really good, thanks for improving the whole modal's look and writing functional tests!

> * Can we put Simple Content Access on top of USMA (whatever that stand for), that seems to be showing the manifest name which is more related to the upload file buttons, so they should be grouped together.

I am still wondering about this, I'm asking if we can do:
```
Simple Content Access
Subscription Allocation
Import New Manifest
```
as the order? The reason being Subscription Allocation and Import New Manifest are related and seem like a more natural grouping.

The only other thing is it could be a good idea to ask UX what the best way to display `Toggling Simple Content Access will refresh your manifest.` would be? Maybe even just put it in italics? but given the improvements here, I'm fine with it as is too.

piyushtiw:
> @piyushtiw It looks really good, thanks for improving the whole modal's look and writing functional tests!
> 
> > * Can we put Simple Content Access on top of USMA (whatever that stand for), that seems to be showing the manifest name which is more related to the upload file buttons, so they should be grouped together.
> 
> I am still wondering about this, I'm asking if we can do:
> 
> ```
> Simple Content Access
> Subscription Allocation
> Import New Manifest
> ```
> 
> as the order? The reason being Subscription Allocation and Import New Manifest are related and seem like a more natural grouping.
> 
> The only other thing is it could be a good idea to ask UX what the best way to display `Toggling Simple Content Access will refresh your manifest.` would be? Maybe even just put it in italics? but given the improvements here, I'm fine with it as is too.

Screenshot from 2020-07-22 09-07-06

Thanks @johnpmitsch for suggestion. I have made the changes accordingly.

webpack/scenes/Subscriptions/Manifest/SimpleContentAccess.js
jturel: That sounds good in combination with removing the hideModal() functionality of the component per my below comment. Let's do it :)

piyushtiw: @jturel I have made the changes and we are hiding the modal when task is getting started. I have made the changes for refresh action also. We can make changes for upload and delete Manifest action also but because it has more than one modal open, it was looking weird to me to hide 2 modal in time of few seconds, So I have not made changes there. If you want I can also remove `hideModal` function call from those actions.

webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
```diff
@@ -198,52 +193,35 @@ class ManageManifestModal extends Component {
...
                           </Col>
-                          <Col sm={8}>
+                          <Col sm={7}>
                             {getManifestName()}
                           </Col>
-                        </Row>
-                        <Row>
-                          <Col sm={4}>
-                            <div>{__('Simple Content Access')}</div>
+                          <Col sm={5}>
```
jeremylenz, line 213:
```suggestion
                        </Row>
                        <Row>
                          <Col sm={5}>
```
Not strictly required, but this will prevent Subscription Allocation and Import New Manifest from being in the same row

webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
```diff
@@ -198,52 +193,35 @@ class ManageManifestModal extends Component {
...
+                            {canImportManifest &&
+                              <ControlLabel
+                                className="control-label"
```
jeremylenz, line 216:
```suggestion
 
```
The explicit className isn't needed since Patternfly already adds the `control-label` class.  In my html this made the `<label>` have a class of `"control-label control-label"`.  This line can be removed.

webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
```diff
@@ -198,52 +193,35 @@ class ManageManifestModal extends Component {
...
+                                style={{ paddingTop: '10 px' }}
```
jeremylenz, line 217:
```suggestion
                                style={{ paddingTop: '10px' }}
```
With the space there, the style tag isn't picked up.

webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
```diff
@@ -20,31 +20,21 @@ class ManageManifestModal extends Component {
...
   }
 
-  static getDerivedStateFromProps(newProps, prevState) {
-    if (
-      !isEqual(newProps.taskInProgress, prevState.actionInProgress)
-    ) {
-      return {
-        actionInProgress: newProps.taskInProgress,
-      };
+  componentWillMount() {
```
jeremylenz, line 28: The React folks _really_ don't like the `componentWillMount` method and in general strongly discourage it.  We'll have to start calling it `UNSAFE_componentWillMount` once we upgrade to React v17.  However, I can see why it makes sense here.  Our use case is rather unique and I don't think it breaks any of their [guidelines](https://reactjs.org/docs/react-component.html#unsafe_componentwillmount), so I'm okay with it.

PR #8800, opened 2020-07-01
Fixes #30271 - Add foreman-js link script

laviro:
it is useful when developing with the `foreman-js` repo

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,8 @@
     "lint": "npx eslint webpack/",
     "lint:fix": "npx eslint --fix webpack/",
     "lint:test": "npm run lint && npm test",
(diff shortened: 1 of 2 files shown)


johnpmitsch:
@laviro since we use `@foreman` packages from foreman, I've always ran the foreman-js link there. We only specify `@theforeman/builder` and `@theforeman/find-foreman` in Katello's packages.json. Do you intend this script just for those packages?

What scenarios do you see this being useful for?

laviro:
I wasn't sure if it's really passed through foreman in jest
but I guess I'll need to avoid running this line: https://github.com/theforeman/foreman-js/pull/143/files#diff-f5b93a9b235dba09ab963eef65e8da8eR60 if it's Katello, thanks @johnpmitsch

laviro:
closing for now :)

PR #8802, opened 2020-07-01
Fixes #30168 - support running without pulp2 enabled

jlsherrill:


jlsherrill:
TODO: add tests, further test

ehelms:
Seeing issues with content view publish:

```
 ✗ publish content view
   (in test file fb-content-katello.bats, line 264)
     `hammer content-view publish --organization="${ORGANIZATION}" \' failed with status 70
   Could not publish the content view:
     There was an issue with the backend service pulp: 404 Not Found
```

ehelms:
Here is the ping:

```
[root@pipeline-katello-server-nightly-centos8 bats]# curl https://`hostname`/api/v2/ping | json_reformat
100   465    0   465    0     0   2012      0 --:--:-- --:--:-- --:--:--  2004
{
    "results": {
        "foreman": {
            "database": {
                "active": true,
                "duration_ms": "0"
            }
        },
        "katello": {
            "services": {
                "candlepin": {
                    "status": "ok",
                    "duration_ms": "31"
                },
                "candlepin_auth": {
                    "status": "ok",
                    "duration_ms": "27"
                },
                "foreman_tasks": {
                    "status": "ok",
                    "duration_ms": "2"
                },
                "katello_events": {
                    "status": "ok",
                    "message": "0 Processed, 0 Failed",
                    "duration_ms": "0"
                },
                "candlepin_events": {
                    "status": "FAIL",
                    "message": "Not running",
                    "duration_ms": "0"
                },
                "pulp3": {
                    "status": "ok",
                    "duration_ms": "84"
                }
            },
            "status": "ok"
        }
    }
}
```

jlsherrill:
For devs doing testing:

* Disable pulp2 from smart proxy, edit /etc/foreman-proxy/settings.d/pulp.yml change ':enabled:" to 'false'
* restart foreman-proxy
* comment out 'pulp' section in config/settings.plugins.d/katello.yaml
* refresh smart-proxy features in UI

ekohl:
> * Disable pulp2 from smart proxy, edit /etc/foreman-proxy/settings.d/pulp.yml change ':enabled:" to 'false'
> * restart foreman-proxy
> * refresh smart-proxy features in UI

Any reason you're not using `foreman-installer --foreman-proxy-pulp-enabled false` to do this? Is that because the foreman-proxy-content module breaks things?

> comment out 'pulp' section in config/settings.plugins.d/katello.yaml

Don't forget to restart the foreman service after this.

jlsherrill:
@ekohl mostly because the testing notes were focused on a devel environment.  I've had bad luck re-running the installer in the past on a devel environment so i provided manual steps.

app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -204,11 +210,11 @@ def pulp3_url(path = '/pulp/api/v3')
       end
 
       def pulp_mirror?
-        self.has_feature? PULP_NODE_FEATURE
+        self.has_feature?(PULP_NODE_FEATURE) || (self.has_feature?(SmartProxy::PULP3_FEATURE) && self.setting(SmartProxy::PULP3_FEATURE, 'mirror'))
       end
 
       def pulp_master?
-        self.has_feature? PULP_FEATURE
+        self.has_feature?(PULP_FEATURE) || (self.has_feature?(SmartProxy::PULP3_FEATURE) && !self.setting(SmartProxy::PULP3_FEATURE, 'mirror'))
```
ekohl, line 217: I think that when a setting exists, implies it has the feature. If a feature is removed, there are no settings stored. So I think the `has_feature?` is redundant if you check `mirror` is `false`.

jlsherrill: Yeah, but we're also trying to differentiate between not having the feature and having the feature set to false.   If it doesn't have the feature its not a pulp_master, if the mirror setting is false its not a pulp_master, so i think it makes sense to differentiate here.  

On the 'pulp_mirror?' function we could do that though and just check that the setting is true.  I'll update that!

ekohl: Don't we always set the value? There is a default (https://github.com/theforeman/smart_proxy_pulp/blob/3eca0c86a3414fb5e8fdeb29dfd918d2a39f0ffa/lib/smart_proxy_pulp_plugin/pulpcore_plugin.rb#L8) so isn't this the same?
```suggestion
        self.has_feature?(PULP_FEATURE) || self.setting(SmartProxy::PULP3_FEATURE, 'mirror') == false
```

jlsherrill: oh you mean an explicit check on == false.  That works for me

app/services/katello/registration_manager.rb
```diff
@@ -183,6 +185,10 @@ def destroy_host_record(host_id)
         Rails.logger.warn("Attempted to destroy host %s but host is already gone." % host_id)
       end
 
+      def pulp2_supported?
+        SmartProxy.pulp_master.nil? || SmartProxy.pulp_master.has_feature?(SmartProxy::PULP_FEATURE)
```
ekohl, line 189: For my understanding: if pulp_master is nil, why does that imply Pulp 2 is supported?

jlsherrill: honestly, i'm not sure, lets get rid of it

jlsherrill:
Requires:  https://github.com/theforeman/foreman/pull/7864

jjeffers:
@jlsherrill tested this after disabling pulp2 per the suggestions.
Then applied both https://github.com/theforeman/foreman/pull/7864 and this PR changes. 

I was able to create a yum repo, sync that repo, and then create and publish a CV. I didn't encounter any issues, so the changes appear to work as intended.

PR #8803, opened 2020-07-01
[CP] Fixes #30283 - Lock theforeman js packages to 4.3.0

jturel:


package.json
```diff
@@ -56,8 +56,8 @@
     "react-test-renderer": "^16.0.0",
     "redux-mock-store": "^1.3.0"
   },
-  "_comment": "We don't include @theforeman/vendor because it's assumed to be present in Foreman",
   "dependencies": {
+    "@theforeman/vendor-core": "4.3.0",
```
jturel, line 60: @johnpmitsch this PR is a cherry-pick of of your master lint fixes, and I'm also pinning theforeman js to 4.3.0

I had to explicitly add this because it *still* keeps pulling 4.10 due to versioning in theforeman/vendor. I think it shouldn't be a problem though ... but let me know what you think

johnpmitsch: if its just for testing purposes that is fine. The dependencies in both production and testing come from Foreman, but I guess this is overriding it for the tests and it matches what is shipped. :+1: 

Not sure what the procedure is now, but we should perhaps think about creating release branches with a package-lock.json that matches what is shipped in the repo. This was the test npm package versions match the production ones

jturel: Agreed! package-lock.json was committed in the 2.1-stable branch, but we should probably do the same in the future. For now I'd like to keep this change small since we're late in the release cycle

jturel:
@johnpmitsch can I get an ACK ? :)

PR #8804, opened 2020-07-02
Fixes #30286 - batch repos on capsule sync

jlsherrill:


wbclark:
Also, thanks @jlsherrill for the collab and quick delivery of this patch!

jlsherrill:
@wbclark yes, this could lead to slower overall syncs, because one batch has to fully complete before the next starts.   This would be most noticeable on the first sync to a capsule when the most work is done.  When most of the syncs are less than a minute its probably not going to be a large deal.  A larger batch size would decrease the chances for this to happen. 

We probably want to make this size configurable so that the user can increase it if they desire.  By using a large value (such as 10,000), they'd effectively be disabling it.

But even if the syncs are somewhat slower (i think it would be a low percentage increase), this change is probably worth it.  We may want to increase the default to 30 or so just to reduce the chances more

jlsherrill:
[test katello]

app/lib/actions/katello/capsule_content/sync_capsule.rb
```diff
@@ -11,19 +11,23 @@ def plan(smart_proxy, options = {})
...
           smart_proxy_helper = ::Katello::SmartProxyHelper.new(smart_proxy)
-          concurrence do
-            smart_proxy_helper.repos_available_to_capsule(environment, content_view, repository).each do |repo|
-              plan_pulp_action([Actions::Pulp::Orchestration::Repository::SmartProxySync,
-                                Actions::Pulp3::CapsuleContent::Sync],
-                                 repo, smart_proxy,
-                                 skip_metadata_check: skip_metadata_check)
+          sequence do
+            smart_proxy_helper.repos_available_to_capsule(environment, content_view, repository).in_groups_of(Setting[:foreman_proxy_content_batch_size], false) do |repos|
+              concurrence do
```
parthaa, line 16: nitpick. Is the indent off here?

PR #8806, opened 2020-07-02
fixing #30296 - Fixing not listed macros

domitea:
This fix solves problem where Katello template macros was not listed in template page at help section but they're working. The reason why is not working is fairly simple. In current implementation in `BaseTemplateScopeExtensions` where we're overriding  `allowed_helpers` in `Foreman::Renderer::Configuration`. However, this extension is pointed to `::Foreman::Renderer::Scope::Base` so these macros are not in `allowed_helpers`. 

So this problem why I make this PR, solution is fairy simple - I've split macros and `allowed_helpers` to two different modules that they're injected to correct classes. It's only draft for now because I don't know if it's best solution for this problem ;)

### DIFF
--- a/lib/katello/engine.rb
+++ b/lib/katello/engine.rb
@@ -154,7 +154,6 @@ module Katello
 
       # Lib Extensions
       ::Foreman::Renderer::Scope::Variables::Base.include Katello::Concerns::RendererExtensions
-      ::Foreman::Renderer::Scope::Base.include Katello::Concerns::BaseTemplateScopeExtensions
 
       # Model extensions
       ::Environment.include Katello::Concerns::EnvironmentExtensions
--- a/app/lib/katello/concerns/base_template_scope_extensions.rb
+++ b/app/lib/katello/concerns/base_template_scope_extensions.rb
@@ -3,20 +3,6 @@ module Katello
     module BaseTemplateScopeExtensions
       extend ActiveSupport::Concern
 
-      module Overrides
-        def allowed_helpers
-          super + [:errata, :host_subscriptions, :host_applicable_errata_ids, :host_applicable_errata_filtered,
-                   :host_latest_applicable_rpm_version, :load_pools, :load_errata_applications, :host_content_facet,
-                   :host_sla, :host_products, :sub_name, :sub_sku, :registered_through, :last_checkin, :host_collections,
-                   :host_subscriptions_names, :host_subscriptions, :host_products_names, :host_collections_names,
-                   :host_redhat_subscription_names, :registered_at]
-        end
-      end
-
-      included do
-        prepend Overrides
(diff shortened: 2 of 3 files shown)


domitea:
@tbrisker commit message updated

app/lib/katello/concerns/renderer_configuration_extension.rb
```diff
@@ -0,0 +1,21 @@
...
+                   :host_latest_applicable_rpm_version, :load_pools, :load_errata_applications, :host_content_facet,
+                   :host_sla, :host_products, :sub_name, :sub_sku, :registered_through, :last_checkin, :host_collections,
+                   :host_subscriptions_names, :host_subscriptions, :host_products_names, :host_collections_names,
+                   :host_redhat_subscription_names, :registered_at]
+        end
+      end
+
+      included do
+        prepend Overrides
+      end
```
tbrisker, line 18: Not crucial, but can we prepend `::Foreman::Renderer::Configuration` directly instead of including this module which then prepends the `Overrides` module?

lib/katello/engine.rb
```diff
@@ -155,6 +155,7 @@ class Engine < ::Rails::Engine
       # Lib Extensions
       ::Foreman::Renderer::Scope::Variables::Base.include Katello::Concerns::RendererExtensions
       ::Foreman::Renderer::Scope::Base.include Katello::Concerns::BaseTemplateScopeExtensions
+      ::Foreman::Renderer::Configuration.include Katello::Concerns::RendererConfigurationExtension
```
tbrisker, line 158: Perhaps i did not explain myself well, what I suggested was changing this to 
```suggestion
      ::Foreman::Renderer::Configuration.prepend Katello::Concerns::RendererConfigurationExtension
```

And changing the new module to:
```ruby
module Katello
  module Concerns
    module RendererConfigurationExtension
      def allowed_helpers
        super + [:errata, :host_subscriptions, :host_applicable_errata_ids, :host_applicable_errata_filtered,
                 :host_latest_applicable_rpm_version, :load_pools, :load_errata_applications, :host_content_facet,
                 :host_sla, :host_products, :sub_name, :sub_sku, :registered_through, :last_checkin, :host_collections,
                 :host_subscriptions_names, :host_subscriptions, :host_products_names, :host_collections_names,
                 :host_redhat_subscription_names, :registered_at]
      end
    end
  end
end
```

However, looking at the Foreman plugin registry, I think this monkey patching is not actually needed at all - we can use the `allowed_template_helpers` method to register these helpers:

https://github.com/theforeman/foreman/blob/develop/app/registries/foreman/plugin.rb#L367-L374

jlsherrill:
tested and worked great! Thanks @domitea !!

PR #8808, opened 2020-07-02
Fixes #29996 - Adds Content View Details Page

johnpmitsch:
Adds new page at `/labs/content_views/:id`

Breadcrumbs and "Publish New Content View" button still need to be added, that will be done in follow-up PRs. This adds the name header, the tabs, and the details tab.

Requires https://github.com/theforeman/foreman/pull/7261


johnpmitsch:
Here is a screenshot:

here

I still need to update the content view index page, for now it is a simple link on the name, but type and name should be separate columns.

johnpmitsch:
Here is the update workflow for a text box and checkbox. It is hard to see, but there are two loading steps. the steps are:
- Send PUT request to update CV
- Show loading icon inside tab
- Show notification based on status of request
- Send GET request for updated CV info
- Show loading icon on page
- New data is loaded and page shows normally

cv_details_update

This is different than what we do today, we currently don't send another GET request after updating. Doing this has the advantage of always making sure the text is showing the correct information after update and not managing two sets of state (the state of the form and the state of the actual CV). It was a bit tricky to do both and have it stay correct based on successful or unsuccessful update. I don't think it's much overhead to do a GET call for a single object, so went with this approach.

I know the loading icons could be improved, perhaps to a loading icon just on entry being updated, but it does require handling state for each detail entry and its hard to show it for two API calls, the PUT and the GET request. I just went with the above approach to start, I think in production it will hardly be noticeable having both loading icons and I do like having the confidence of a GET request to load the latest data after updating.

johnpmitsch:
> One thing I noticed, if you try to navigate to a content view that doesn't exist, the server errors out trying to look up that ID. I think this is a content views controller bug:
> 
> ```
> 2020-07-06T16:31:41 [I|app|f7157133] Processing by Katello::Api::V2::ContentViewsController#show as JSON
> 2020-07-06T16:31:41 [I|app|f7157133]   Parameters: {"api_version"=>"v2", "id"=>"999"}
> 2020-07-06T16:31:41 [D|tax|f7157133] Current location set to Default Location
> 2020-07-06T16:31:41 [D|tax|f7157133] Current organization set to Default Organization
> 2020-07-06T16:31:41 [E|app|f7157133] ActiveRecord::RecordNotFound: Couldn't find Katello::ContentView with 'id'=999
> ```
> 
> unable_to_connect
> 
> I also noticed that searching for a content view with a non-integer ID results in a blank page, will there be some sort of "resource not found" page? I noticed that the route is described as so:
> 
> ```js
> path: 'labs/content_views/:id([0-9]+)',
> ```
> 
> Perhaps non-integer IDs are just being rejected without any further processing.

Thanks for pointing that out, I just checked the existing behavior for the subscriptions (also react) page.

If you go to subscriptions/id where id is an integer that doesn't exist, it shows an error notification "Subscription not found". We should probably do the same here with Content Views. I imagine it would if the server responded with a 404 since there is error handling on the API request.

I'll check what the server is doing and confirm that a 404 would be handled appropriately in the front-end, then see if the server side can be fixed here or a follow-up issue makes more sense.

If you go to subscriptions/id where id is a non-integer, it also shows a blank page, we should probably have a default catch-all 404 page. I'm not sure if it will make more sense to handle here or in a follow up, but will look to see if its an easy add.

johnpmitsch:
@jlsherrill Do we still need `Force Puppet Environment` option?

jlsherrill:
i think when this lands we won't have puppet, so nope, you can get rid of it in the UI

MariSvirik:
After


@johnpmitsch @Rohoover 
There are just a couple of things, I would like to comment on:

- **Spacing** (please, follow the spacing displayed at the picture - red color) We can discuss further global 20 px between Nav and page content on Friday. 

- **Font size** (please, follow the displayed values at the picture - blue color)

- **Components**

**Alert** see [my mock](https://marvelapp.com/prototype/1121a8de/screen/70946341) and [PF4 documentation](https://www.patternfly.org/v4/documentation/react/components/alert#component-title). Alert component was updated in the latest PF4 release.
As we are not referring to the Content View page, but to specific content view you should capitalized just C because it's the first letter in the sentence: 'Content view updated'

**Labels** (Name, Label, Description etc) A correct component is '[Field label](https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/forms)' Font size in this component is surprisingly smaller than the regular text.

**Icon** 
It should be **fa-pencil-alt**.  Recommendations to update icons are [here.](https://www.patternfly.org/v4/design-guidelines/styles/icons#component-title)
Screenshot 2020-07-03 at 11 45 47

- Be careful with the **color of icon buttons** (pencil, question mark). They are actionable. When it's default color should be pf-global --icon--light #6A6E73 and when it's hovered or selected pf-global --icon--dark  #151515.

- 👍 Good idea to use a switch instead of checkbox and structure that information the way you did. 

- 👍 Good job that you haven't forgotten to add CV type. (I did after so many iterations 😄)
It's already added to [my mocks ](https://marvelapp.com/prototype/1121a8de/screen/69017447)additionally as well.
Screenshot 2020-07-07 at 16 08 36 
Now the label is called 'Type'. It is more consistent with the main page.

- 🤔 Somehow tabs seem to be off (maybe it's because of the spacing ). Please, double-check component usage.

On Friday we can discuss with devs what kind of information the label 'Label' displays and how it behaves.

johnpmitsch:
I realized we need "Auto Publish" for composite CV details too, will update with that

johnpmitsch:
@MariSvirik I updated the main page as well: 

here

It's not complete yet, the expansions still need to be filled out and the details tab will be a blue link that expands the expansion, but I've moved "Type" to it's own column and removed the repositories tab

webpack/components/EditableTextInput/EditableTextInput.js
```diff
@@ -0,0 +1,98 @@
...
+    const sharedProps = {
+      value: inputValue,
+      onChange: v => setInputValue(v),
+    };
+
+    if (textArea) {
+      return <TextArea {...sharedProps} aria-label={`text area ${label}`} />;
+    }
+    return <TextInput {...sharedProps} type="text" aria-label={`text input ${label}`} />;
+  };
```
jeremylenz, line 36: I don't think this const is necessary.  Why not just put everything inline in JSX?
```javascript
{
  textArea && <TextArea value={inputValue} onChange={v => setInputValue(v)} aria-label={`text area ${label}` />
}
{ 
  !textArea && <TextInput value={inputValue} onChange={v => setInputValue(v)} aria-label={`text input ${label}`} />
}
```
Having more of the markup in JSX like this lets me get a clearer picture in my head of the structure of the page, and saves me from having to jump around the code when reading it.

webpack/components/EditableTextInput/EditableTextInput.js
```diff
@@ -0,0 +1,98 @@
...
+
+  if (working) return <Loading size="sm" />;
+  return (
+    <Fragment>
+      {editing ?
```
jeremylenz, line 41: Prefer conditional rendering to avoid the long ternary:
```javascript
<Fragment>
  {editing &&
    <Split>
     ...
   </Split>
  }
  {!editing &&
  <Split>
     ...
   </Split>
  }
</Fragment>
```

webpack/components/TabbedView/TabbedView.js
```diff
@@ -0,0 +1,43 @@
...
+import {
+  Page, Tabs, Tab, TabTitleText, PageSection, PageSectionVariants,
+} from '@patternfly/react-core';
+
+const TabbedView = ({ tabs }) => {
+  const [activeTabKey, setActiveTabKey] = useState(0);
+  const handleTabClick = (_event, tabIndex) => setActiveTabKey(tabIndex);
+
+  return (
+    <React.Fragment>
```
jeremylenz, line 12: Fragment isn't necessary since there's only one child

webpack/scenes/ContentViews/ContentViewSelectors.js
```diff
@@ -7,6 +7,9 @@ import CONTENT_VIEWS_KEY from './ContentViewsConstants';
 
 export const selectContentViews = state => selectAPIResponse(state, CONTENT_VIEWS_KEY) || {};
 
+// export const selectContentView = (state, id) =>
+// selectAPIResponse(state, `${CONTENT_VIEWS_KEY}${id}`) || {};
```
jeremylenz, line 11: leftover?

webpack/scenes/ContentViews/Details/ContentViewDetailSelectors.js
```diff
@@ -13,3 +13,6 @@ export const selectCVDetailStatus =
 
 export const selectCVDetailError =
   (state, cvId) => selectAPIError(state, `${CONTENT_VIEWS_KEY}_${cvId}`);
+
+
+export const selectCVUpdating = state => state.katello?.contentViewDetails?.updating;
```
jeremylenz, line 18: Optional chaining party :tada: 

Since this returns a boolean, prefer naming like `selectIsCVUpdating`

webpack/scenes/ContentViews/Details/ContentViewDetails.js
```diff
@@ -0,0 +1,116 @@
...
+    description,
+    composite,
+    force_puppet_environment: forcePuppetEnvironment,
+    solve_dependencies: solveDependencies,
+  } = details;
+  const info = {
+    name: {
+      value: name,
+      label: __('Name'),
+    },
```
jeremylenz, line 27: Took me 15+ minutes to get into my head how this object literal ends up as JSX.  It was confusing to me (for some reason) that the object keys end up as `key`s, though that actually makes sense.  Even so, what you're doing here looks _very_similar to JSX markup assigning different props to each child, but it's in the object literal form so it's hard to comprehend.

Is it possible to do some refactoring of `ContentViewInfo.js` and possibly turn all of these inner objects into props that you assign to React components?  I just think that down the line it would make it that much easier to comprehend and maintain.

jeremylenz: Refactoring in this way would also allow you to eliminate some of the confusion I had around "why does this object have different keys than the previous one?" by using PropTypes and default props.

webpack/scenes/ContentViews/Details/ContentViewDetails.js
```diff
@@ -0,0 +1,116 @@
...
+      boolean: true,
+      tooltip: __('This option will solve RPM and Module Stream dependencies on every publish of ' +
+               'this Content View. Dependency solving significantly increases publish time ' +
+               '(publishes can take over three times as long) and filters will be ignored when ' +
+               'adding packages to solve dependencies. Also, certain scenarios involving errata ' +
+               'may still cause dependency errors.'),
+    },
+  };
+
+  const tabs = [
```
jeremylenz, line 63: Was about to request the same for this const, but realized the Patternfly API probably will make it impossible.  It's ok for this one, though, since it's more straightforward.

webpack/components/EditableTextInput/EditableTextInput.js
johnpmitsch: I'm not sure if this is the best practice, do you see any examples or docs with this way of writing? I would rather change to an if..else than value...!value.

I think using ternaries is more just a matter of style preference. React docs provide both ternary and if..else as an example  so both seem to be acknowledged as legitimate values: https://reactjs.org/docs/conditional-rendering.html#inline-if-else-with-conditional-operator.

If you feel strongly about it, I can change to if..else, but I would like to stick with the ways the docs provide

johnpmitsch: if..else would have to be outside the render function, as shown in the top of the page

jeremylenz: React doc recommending this pattern [here](https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator)

The React docs also illustrate using a long [ternary](https://reactjs.org/docs/conditional-rendering.html#inline-if-else-with-conditional-operator), but caution that it's "less obvious what's going on"

Personally I strongly prefer the former (but you knew that ;)

webpack/scenes/ContentViews/Details/ContentViewDetails.js
johnpmitsch: ya, I'm a little torn on this, passing an array of objects and iterating on it is not the best pattern IMO (pf tables *cough* *cough*). I would rather be repetitive with small, declarative components.

I could make this into a `TabWrapper`  or `ForemanTab` that looks like `<TabWrapper title={title} ><MyComponent /></TabWrapper>` for each tab, so the TabWrapper could keep the styling and labels consistent

webpack/components/EditableTextInput/EditableTextInput.js
johnpmitsch: > React doc recommending this pattern [here](https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator)

That is for an inline if logic, not if...else logic like we have here.

I've never seen the `&&` used for if...else logic and I don't see the React docs recommending that. I wouldn't do `value...!value` in normal javascript, so I can't really see the argument for it. I'm going to keep it as is, but I'm open to other opinions or examples.

johnpmitsch: My understanding of the choices the docs give for if..else are:
- ternary operator
- if..else statement outside of render/return statement
- break into separate components

jeremylenz: I would go for breaking into separate components before I had a ternary that long.

webpack/scenes/ContentViews/Details/ContentViewDetails.js
jeremylenz: sounds good to me!

webpack/components/EditableTextInput/EditableTextInput.js
johnpmitsch: let me know what you think of what I've updated to

webpack/scenes/ContentViews/Details/ContentViewDetails.js
johnpmitsch: agreed, this started out simple but got a little too complicated. I updated this to use a new `ActionableDetail` component instead of iterating over this object

webpack/scenes/ContentViews/Details/ContentViewDetails.js
johnpmitsch: I actually couldn't get a wrapper component around a tab and still have it work well in the `Tabs` component. It was using anonymous functions, which didn't seem to be able to be recognized by `Tabs` as the active tab open. I tried different function syntax and hardcoding values but nothing worked. I think it has something to do with how it is implemented in patternfly.

I changed it back to the original tabs array that is iterated through. I think what would be best is to move this component to Foreman, detail out the issue and options, and open up the discussion to see if there are suggestions.

johnpmitsch:
@MariSvirik @rohoover Here are some updated screenshots:

CV with no description
here

CV with description and solve deps
here

composite
image

johnpmitsch:
@jlsherrill Do we display `label` in CV details any particular reason? Should we keep that in the new page?

webpack/utils/helpers.js
```diff
@@ -28,9 +28,9 @@ export const getResponseErrorMsgs = ({ data, actionType } = {}) => {
     const customMessage = getCustomMessage(actionType, data.displayMessage);
     const messages =
       customMessage ||
-      data.errors ||
       data.displayMessage ||
       data.message ||
+      data.errors ||
```
johnpmitsch, line 33: `errors` wasn't formatted well in an error which created a `undefined` notification, but it was provided with `displayMessage`. 

It seems that we should be displaying `message` or `displayMessage` in a notification before we dig into the errors object to look for a message.

webpack/components/EditableTextInput/EditableTextInput.js
johnpmitsch: I moved it into JSX but as a ternary for the reasons mentioned in the other comment

webpack/utils/helpers.js
jeremylenz: I remembered fixing something similar in #8643... too bad we don't have some sort of standard for this.

webpack/components/EditableTextInput/EditableTextInput.js
jeremylenz: Totally fine with `if` with early returns.  :+1:

webpack/components/EditableTextInput/EditableTextInput.js
jeremylenz: 👿 

Actually this is fine, it's not too long.. :smile:

webpack/utils/helpers.js
```diff
@@ -81,12 +81,15 @@ export const sendErrorNotifications = messages => (dispatch) => {
 export const apiError = (actionType, result, additionalData = {}) => (dispatch) => {
   const messages = getResponseErrorMsgs(result.response);
 
-  const dataExtenstion = {
-    messages,
-    ...additionalData,
-  };
+  // If no actionType passed in, only create notification and skip dispatching action
+  if (actionType) {
+    const dataExtenstion = {
```
jeremylenz, line 86:
```suggestion
 let dataExtension;
  if (actionType) {
    dataExtension = {
```
If no actionType is passed in, `dataExtension` will never be defined, which will cause a `ReferenceError: dataExtension is not defined`

MariSvirik:
@johnpmitsch it looks good. 

1. Again, make sure to use sentence case (almost all the time).

- Instead of 'Solve Dependencies' use 'Solve dependencies'

- Instead of 'Auto Publish' use 'Autopublish' (I asked the content team and they prefer this version of the word)

2. A pencil icon in description (the one that has content) seems to be too far from the description content.

3. 'None provided' should be written in the regular Red Hat text (16) - the same as the rest. (Inventory displays it the same way)- I'm still trying to confirm this one.

johnpmitsch:
> Thanks @johnpmitsch, looking really good! Just a couple things
> 
> * When I edit a description, I want to be able to press Enter to commit.  is this possible?

Good thinking, I added it in

> * When I clicked the heavy_check_mark I got an error
> 
> ```
> foreman-vendor.bundle-v4.11.1-development-8f1002a20f1552db88e3.js:399366 Warning: Failed prop type: The prop `value` is marked as required in `ActionableDetail`, but its value is `null`.
> ```
> 
> (update: I think that was because I hadn't checked out [theforeman/foreman#7261](https://github.com/theforeman/foreman/pull/7261) )

Ya, that should be it, but let me know if you see it again.

> 
> * When I click the page it gets a border around it that looks odd (also happens with some other elements)
>   Content_View_Details

I noticed that too, it happens on the main page as well, I'm going to have to look into it more

johnpmitsch:
> @johnpmitsch it looks good.
> 
> 1. Again, make sure to use sentence case (almost all the time).
> 
> * Instead of 'Solve Dependencies' use 'Solve dependencies'
> * Instead of 'Auto Publish' use 'Autopublish' (I asked the content team and they prefer this version of the word)
> 

Thanks, updated!

> 1. A pencil icon in description (the one that has content) seems to be too far from the description content.

It seems to be better now, I'm not sure if it's because something I changed since then or it's the text wrapping in the first screenshot that created space :man_shrugging: 

> 2. 'None provided' should be written in the regular Red Hat text (16) - the same as the rest. (Inventory displays it the same way)- I'm still trying to confirm this one.

It actually is RH text at 16px, just in italics, should I change anything?

Here is an updated screenshot:

here

PR #8810, opened 2020-07-02
Refs #29927 - Add messaging & links for SCA

jeremylenz:
* Add Subscription Watch button to Subscriptions page
* Add SubWatch link to SCA banner on Subscriptions page

sub_watch_subs_page

https://www.awesomescreenshot.com/video/339889?key=f7f1ca0b5e6317e17f2ef85c760d722a


jeremylenz:
* _As a user, I want to understand why Subscription Status is 'Disabled' (tooltip)_
* _As a user, I want to understand why Auto-Attach is 'Not applicable' (tooltip)_

@jturel Currently on Content Host details, we just show Subscription Status as "Simple Content Access" (not "Disabled"), we show the explanatory SCA banner at the top, and we don't show Auto Attach at all.  So I think these stories may not be needed anymore.  Thoughts?  Is any other messaging still needed?

sca_content_host_details

jeremylenz:
https://www.awesomescreenshot.com/image/4999133/ca2d86d1fffb26995497e770fe439049

sideangleside:
Note, if Simple Content Access is enabled, you'd also want to explain why 'system purpose status' is disabled/unknown as well. See [bz1854138](https://bugzilla.redhat.com/show_bug.cgi?id=1854138)

jeremylenz:
@sideangleside I thought we're not disabling System Purpose for SCA (see your [comment on Aha](https://redhatmbu.aha.io/comments/6827140165120710884)) So shouldn't the System Purpose messaging be the same?

jturel:
> @sideangleside I thought we're not disabling System Purpose for SCA (see your [comment on Aha](https://redhatmbu.aha.io/comments/6827140165120710884)) So shouldn't the System Purpose messaging be the same?

You can still set system purpose in SCA, but there is no status. ie you can't be Mismatched if you aren't dealing with subscriptions

jturel:
Looks good! The only change I think we need here is for the button to only show if we've got a manifest imported. When no manifest is imported the user is probably not a Red Hat customer, and so swatch is not relevant to them

webpack/scenes/Organizations/OrganizationSelectors.js
```diff
@@ -1,4 +1,18 @@
...
+
+// for use in ManageManifestModal to replace getManifestName()
+export const selectManifestHref = state =>
+  selectManifestName(state) && [
+    'https://',
+    selectOrganizationState(state).owner_details.upstreamConsumer.webUrl,
+    selectOrganizationState(state).owner_details.upstreamConsumer.uuid,
+  ].join('/');
+
+export const selectIsManifestImported = state =>
```
jturel, line 14: It looks like we have two ways of making this determination now. Do you think it should stay that way, or could it be consolidated? 

https://github.com/Katello/katello/blob/ac0620f2d614d6ad690008f3f5a2d7848972ce6e/webpack/scenes/Subscriptions/SubscriptionHelpers.js#L6-L7

jeremylenz: good catch!  It should probably be consolidated, but I'd say the one in `SubscriptionHelpers.js` is basically a selector, but in the wrong file.  So we should consolidate in favor of `selectIsManifestImported`

jeremylenz:
[test katello]

jeremylenz:
rebased to fix merge conflict

PR #8813, opened 2020-07-07
Fixes #30336 - Fix React test intermittent failure

johnpmitsch:
We are seeing this issue happen with jenkins, it seems to run quite a bit slower than GH actions.

I believe the `waitFor` command is timing out as it waits for the text to appear on the screen. react-testing-library doesn't have a way to increase the default (1000ms) waitFor afaict, so we can set it in the waitFor commands where needed. If it continues to be a problem, we can think about making a wrapper function that calls waitFor with a longer timeout.


johnpmitsch:
If we still see this issue or see a related one, please comment here or contact me. It passed a test run, but since it is an intermittent issue, that is not a guarantee that the issue is fixed!

PR #8820, opened 2020-07-08
Fixes #30343 - Sync plan fails

ranjan:
Sync plan fails when root_repo has no library_instance


sjha4:
I don't follow this completely - _when root_repo has no library_instance_.. I would think this shouldn't happen at all..Do you have any steps to reproduce this?

PR #8821, opened 2020-07-08
Fixes #33783 - Support connected Inter-Server Sync

jturel:

This change brings back the ability to sync Red Hat content between two Katello instances. There was a way to achieve this when we were on pulp2 but that went away due to the lack of pulp3 publishing a file structure to disk that mimics the Red Hat CDN.

It works by introducing a new `CdnConfiguration` model which allows the user to set the necessary information to talk to the API of another Katello. That information is used to configure a new `KatelloCdn` resource that extends the existing `CdnResource` class. By providing similar (and some new) APIs, the `KatelloCdn` can be used by the /repository_sets/:id/available_repositories API to know which repositories are available for enablement, and properly configure any enabled repo to pull content from the upstream instance by using the org's debug certificate as the auth mechanism.

Testing this requires two Katello instances- a devel box and a 4.2 box for example. Only the downstream box (devel) needs to have this change applied which also makes this feature portable to older Katellos using pulp3.

I used the following script in my own testing to execute the workflow with two orgs on a single Katello instance and it could be adapted for two boxes if you want: https://gist.github.com/jturel/2e18c3784f3abeefcbfb3905f5372e66

The workflow is:

**Upstream Katello setup**
- Configure some Red Hat content. => enable a small repo like satellite-tools
- Create a personal access token for a user who would have access to the repository sets API (could be admin...)

**Downstream Katello setup**
- Apply this code change, migrate DB
- Import a manifest which provides the same repo you enabled above
- Create a Content Credential which has the CA certificate of the upstream server
- Update the `CdnConfiguration` of your downstream org to talk to the upstream server:

```
curl http://localhost:3000/katello/api/v2/organizations/$ORG_ID/cdn_configuration -u admin:changeme -X PUT -H 'Content-type: application/json' \
  -d '{"username": "$UPSTREAM_KATELLO_USER'", "password": "$PERSONAL_ACCESS_TOKEN", "organization_label": "'$UPSTREAM_KATELLO_ORG_LABEL", "url": "https://$UPSTREAM_KATELLO_FQDN", "ssl_ca_credential_id": "$CONTENT_CREDENTIAL_ID"}'
```

- Browse the Red Hat repositories of the downstream Katello. Expanding any repo set should call back to the API of your upstream server to fetch the available paths. If you get an error the CdnConfiguration is likely wrong. If it works - good job following my instructions so far :)

- Try enabling a repository which is **not** enabled on the upstream server. You should get an intelligible error.
- Now enable the same repository that you enabled on the upstream server at the start. It should be enable-able and you should be able to sync it. Check the traffic on your upstream server to verify. Also check the repository details and note that the upstream URL uses a path from the upstream instance rather than cdn.redhat.com

**What I really want you to test**
- Analyze my design and understand how it works. Where does it fail?
- Are there any workflows you can execute that give a bad user experience or just don't work?
- Check the inline comments I've left on the code
- What happens when you revert the downstream CdnConfiguration ie set everything to empty? Can you successfully sync from the Red Hat CDN again?


jturel:
> Thanks for doing this investigation work @jturel
> 
> Seems really well thought-out and straightforward. Overall it really does make a lot of sense to me, nice work! After reading the doc I think I understand (mostly)
> 
> some questions:
> 
> * The general gist of this seems to be "If we're using the RedHat CDN, carry on as normal.  But if the CDN is another satellite, do lots of certificate magic and then 2 API requests."  Right?  Could it be done in a single API request?

Hopefully not too much magic. Just take the certs of the upstream Satellite, create Content Credentials for them, and associate to your 'CDNConfiguration' which will _manifest_ on the Manage Manifest modal.

Technically, it could be done with a single API call, but I think that would mean we need a new API. The approach here really keeps it simple by staying true to our current API. That said, I'll give it another hard look when I pick this back up (next sprint hopefully).

> * What is a "debug" certificate normally used for?  Is there something unorthodox about using them in this way?

A debug certificate (which can be retrieved over the API -> /organizations/:id/debug_certificate) provides a way for folks to get content off their Katello without registering a content host, I guess for experimental purposes. There are probably other uses I'm not thinking of. I think it's fair to say this is a bit unorthodox because we're bringing these certs into real workflow. At the same time it's a great fit, and I don't think we're getting too coupled to the certs. We could move to some other auth mechanism in the future.

app/lib/actions/katello/repository_set/available_repositories.rb
```diff
@@ -0,0 +1,51 @@
+module Actions
+  module Katello
+    module RepositorySet
+      class AvailableRepositories < Actions::AbstractAsyncTask
+        def plan(product:, content_id:)
+          results = []
+          cdn_config = product.organization.cdn_configuration
+
+          if ::Katello::Resources::CDN::CdnResource.redhat_cdn?(cdn_config.url)
```
jlsherrill, line 9: One consequence of this change is that it will assume anything that isn't 'cdn.redhat.com' is another katello instance.  So something like cdn.stage.redhat.com or a manually configured http:// cdn would not work i presume (although stage is maybe the better use case).   I guess the user could manually set the SETTING to their url, but i'm not sure the installer supports that.   Maybe having a cdn type field or other configuration would help make this clear cut?

jturel:
@jlsherrill I think you'd mentioned a concern about how to update the debug cert if it changes upstream. Currently the API to update the CDN will sync the debug cert to any repos that need it, and that should be simple enough to let the user do via API+Hammer. Do you think that's sufficient enough?

parthaa:
Seems to work well in my tests. Can you add something to `organization show.json.rabl` to include CDN Configuration information ?

parthaa (approved):
Works well for me. ACK

app/models/katello/cdn_configuration.rb
```diff
@@ -0,0 +1,20 @@
...
+    # test deleting the CA Cred...
+    belongs_to :ssl_ca_credential, :class_name => "Katello::ContentCredential", :inverse_of => :ssl_ca_cdn_configurations
+
+    encrypts :password
+
+    def ssl_ca
+      ssl_ca_credential&.content
+    end
+
+    def redhat?
```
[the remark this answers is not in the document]
jlsherrill: I think with the validations above, its basically saying if any of these are set, all of them have to be set and thus its not a redhat CDN, which i think is good.  I'm not sure that the resulting validation messages will be clear about that?  Its probably not terrible, but maybe room for improvement.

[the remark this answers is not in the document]
jlsherrill: and a clear UI may go a long way towards that too (I haven't seen the UI yet)

db/migrate/20211019192121_create_cdn_configuration.katello.rb
```diff
@@ -0,0 +1,28 @@
...
+    add_foreign_key :katello_cdn_configurations, :taxonomies, name: 'katello_cdn_configurations_organization_id', column: :organization_id
+    add_foreign_key :katello_cdn_configurations, :katello_content_credentials, name: 'katello_cdn_configurations_ssl_ca_credential_id', column: :ssl_ca_credential_id
+
+    ::Organization.all.each do |org|
+      Katello::CdnConfiguration.where(
+        organization: org,
+        url: org.redhat_provider.repository_url || ::Katello::Resources::CDN::CdnResource.redhat_cdn_url
+      ).first_or_create!
+    end
+  end
```
jlsherrill, line 23: should we drop the redhat provider url column here?

jlsherrill:
overall design looks good, couple of small questions

app/models/katello/cdn_configuration.rb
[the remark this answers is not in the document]
jturel: I see what you mean- let me take a look

db/migrate/20211019192121_create_cdn_configuration.katello.rb
jturel: Good idea, but I I'd like to file a redmine and do it separately. Reason being is that this code might get ported back to 6.9 perhaps as a hotfix for some customers who want to sync RH content from their 7.0. In order to limit the size+risk `of the change, and especially because the `OrganizationCreator' (thing that replaced the Org::Create dynflow task) isn't in 6.9 and moved creation of providers etc. around, I'd rather have the unused field left over in the code+db than break something. Lemme know if I'm not making any sense :)

jturel: Taking a look - maybe it's not so bad

jturel:
[test katello]

jturel:
@jlsherrill updated! take a look at the top few commits

jturel:
Ooops, fixing something :)

jturel:
Fixed!

app/controllers/katello/api/v2/subscriptions_controller.rb
```diff
@@ -108,8 +108,7 @@ def upload
 
       # repository url
       if (repo_url = params[:repository_url])
-        @provider.repository_url = repo_url
-        @provider.save!
+        @organization.cdn_configuration.update!(url: repo_url)
```
jturel, line 111: @parthaa @jlsherrill I wonder if we should remove the repository_url param from this API as part of this change? My thinking is that this code doesn't run the CdnConfiguration::Update (previously Provider::Update) action when making the change - so I don't see its purpose. Notice that the organizations controller does run the dynflow action - and that's what the existing UI uses.

jturel: The API param doesn't show up currently in hammer or foreman-documentation repo so it should be safe

jlsherrill: i think it would be fine to remove it entirely, since its broken as is

jturel:
[test katello]

jlsherrill (approved):
ACK!

PR #8824, opened 2020-07-08
Fixes #30350 - adds the scoped_search form parameter 

jjeffers:
This PR repairs bulk host actions for errata (and packages) modals from the content hosts summary page.

The "scoped_search" form parameter was missing and here we attempt to populate that value before the form is submitted.

To test this change, register at least as many content hosts as the page volume allows, plus 1 more. (You want more hosts to be selectable than can be shown in 1 page of results.)

Screenshot from 2020-07-10 13-51-13

Next, select top checkbox for all hosts, then click the "Select All ..." link near the top to the results table. Then choose an action from the Select Action option, picking "Manage Errata".

Assuming there is applicable errata indicated for at least one of the hosts, select that errata. Then click "Install" with the "remote execution" option. The subsequent task to install the errata should be applied against all hosts (not just the ones shown on the first page).

Note that the same errors appear for "Manage Packages", and this PR repairs that functionality as well.

### DIFF
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-errata-modal.controller.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-errata-modal.controller.js
@@ -57,12 +57,15 @@ angular.module('Bastion.content-hosts').controller('ContentHostsBulkErrataModalC
         $scope.initialLoad = true;
         $scope.remoteExecutionPresent = BastionConfig.remoteExecutionPresent;
         $scope.remoteExecutionByDefault = BastionConfig.remoteExecutionByDefault;
+        $scope.allHostsSelected = hostIds.allResultsSelected;
 
         $scope.errataActionFormValues = {
             authenticityToken: $window.AUTH_TOKEN.replace(/&quot;/g, '')
         };
 
-        if (hostIds.included.ids) {
+        if ($scope.allHostsSelected) {
+            $scope.errataActionFormValues.scopedSearch = hostIds.included.search;
+        } else if (hostIds.included.ids.length > 0) {
             $scope.errataActionFormValues.hostIds = hostIds.included.ids.join(',');
         }
 
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-packages-modal.controller.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-packages-modal.controller.js
@@ -119,6 +119,7 @@ angular.module('Bastion.content-hosts').controller('ContentHostsBulkPackagesModa
 
             $scope.content.confirm = false;
             $scope.packageActionFormValues.customize = customize;
+            $scope.allHostsSelected = selectedHosts.allResultsSelected;
 
             if (!action) {
                 action = $scope.content.action;
@@ -134,7 +135,9 @@ angular.module('Bastion.content-hosts').controller('ContentHostsBulkPackagesModa
                 $scope.packageActionFormValues.remoteAction = 'package_' + action;
             }
 
-            if (selectedHosts.included.ids) {
+            if ($scope.allHostsSelected) {
+                $scope.packageActionFormValues.scopedSearch = selectedHosts.included.search;
+            } else if (selectedHosts.included.ids.length > 0) {
                 $scope.packageActionFormValues.hostIds = selectedHosts.included.ids.join(',');
             }
 
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-packages-modal.html
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-packages-modal.html
@@ -38,6 +38,7 @@
       <input type="hidden" name="host_ids" ng-value="packageActionFormValues.hostIds"/>
       <input type="hidden" name="authenticity_token" ng-value="packageActionFormValues.authenticityToken"/>
       <input type="hidden" name="customize" ng-value="packageActionFormValues.customize"/>
+      <input type="hidden" ng-if="allHostsSelected" name="scoped_search" ng-value="packageActionFormValues.scopedSearch"/>
     </form>
 
     <div ng-form name="systemContentForm" class="form" ng-hide="content.workingMode || denied('edit_hosts')" novalidate>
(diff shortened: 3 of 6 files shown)


jturel:
This change does fix the problem but it took me a little while to figure out why you had to go the route you did, and I learned some things:

- the hostIds data is derived from the `getAllSelectedResults` function in Nutupane (the container object of the content hosts table)

https://github.com/Katello/katello/blob/ab15cbb4235c11bc3aa9c0a91066ec9420b63f20/engines/bastion/app/assets/javascripts/bastion/components/nutupane.factory.js#L286-L310

- that function creates a data structure that we basically send as-is to the API
- i don't like how tightly coupled some random function in our JS code is unintentionally coupled to our API

OK, rant over :)

The `hostIds` doesn't give you an easy way to ask "did we have all hosts selected?" and that's what you're trying to determine by checking the length of the included ids. So, I'm wondering if we can tag on an extra attribute in the object returned by `getAllSelectedResults` which can allow us to answer that question more easily. It should help clean up the logic a bit, and make the intention clearer. Unfortunately you'll need to do a bit of a read in the code to see if we're going to end up sending that new value to the API or something. If not, then I think it's a quick and worthwhile change. Let me know what you think

jjeffers:
@jturel thank you for the feedback. This is precisely the sort of review I like to get.

I think the request to make the state of the selection more apparent is a good one. Let me see what we can come up with.

jjeffers:
@jturel I don't think the additional attribute is passed directly to the API. In both controllers the returned data from nutupane `getAllSelectedResults` is then injected into form values, and not directly supplied.

I'm not sure what the other controllers do, so I'm looking into that now.

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-packages-modal.html
```diff
@@ -38,6 +38,7 @@ <h4 data-block="modal-header" translate>Update Packages</h4>
       <input type="hidden" name="host_ids" ng-value="packageActionFormValues.hostIds"/>
       <input type="hidden" name="authenticity_token" ng-value="packageActionFormValues.authenticityToken"/>
       <input type="hidden" name="customize" ng-value="packageActionFormValues.customize"/>
+      <input type="hidden" ng-if="scopedSearch" name="scoped_search" ng-value="errataActionFormValues.scopedSearch"/>
```
jturel, line 41: I think this should be `packageActionFormValues.scopedSearch`

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-packages-modal.controller.js
```diff
@@ -134,10 +135,15 @@ angular.module('Bastion.content-hosts').controller('ContentHostsBulkPackagesModa
                 $scope.packageActionFormValues.remoteAction = 'package_' + action;
             }
 
-            if (selectedHosts.included.ids) {
+            if (selectedHosts.included.ids.length > 0) {
                 $scope.packageActionFormValues.hostIds = selectedHosts.included.ids.join(',');
             }
 
+            if (selectedHosts.allResultsSelected) {
+                $scope.scopedSearch = true;
```
jturel, line 143: what about this?

```
$scope.allHostsSelected = selectedHosts.allResultsSelected;

if ($scope.allHostsSelected) {
...
}
```

That way you don't need to track `scopedSearch` separately and you can use it in the view as part of your conditional logic

jjeffers: Good idea!

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-errata-modal.controller.js
```diff
@@ -57,15 +57,21 @@ angular.module('Bastion.content-hosts').controller('ContentHostsBulkErrataModalC
...
             authenticityToken: $window.AUTH_TOKEN.replace(/&quot;/g, '')
         };
 
-        if (hostIds.included.ids) {
+        if (hostIds.included.ids.length > 0) {
             $scope.errataActionFormValues.hostIds = hostIds.included.ids.join(',');
         }
 
+        if ($scope.allHostsSelected) {
+            $scope.scopedSearch = true;
```
jturel, line 71: Don't need this line, otherwise LGTM :)

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-packages-modal.controller.js
```diff
@@ -134,10 +135,14 @@ angular.module('Bastion.content-hosts').controller('ContentHostsBulkPackagesModa
                 $scope.packageActionFormValues.remoteAction = 'package_' + action;
             }
 
-            if (selectedHosts.included.ids) {
+            if (selectedHosts.included.ids.length > 0) {
                 $scope.packageActionFormValues.hostIds = selectedHosts.included.ids.join(',');
             }
 
+            if ($scope.allHostsSelected) {
```
jturel, line 142: oh, one thing you'll want to check on (sorry, i just realized)

The way the code is written will send both the ids and the scoped search value. From what I can tell, I think we only care about one or the other. Let me know if you get a different impression from the UI. If this is true, then we should only send one or the other:

```
if ($scope.allHostsSelected) {
 ...
} elseif (selectedHosts.included.ids.length > 0) {
...
}
```

jjeffers: Based on the testing I did, once the scoped search is populated, the host ids didn't seem to make any difference if present.

In any case, seems like this would reduce confusion in the future.

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-packages-modal.controller.js
```diff
@@ -134,7 +135,9 @@ angular.module('Bastion.content-hosts').controller('ContentHostsBulkPackagesModa
                 $scope.packageActionFormValues.remoteAction = 'package_' + action;
             }
 
-            if (selectedHosts.included.ids) {
+            if ($scope.allHostsSelected) {
+                $scope.packageActionFormValues.scopedSearch = selectedHosts.included.search;
+            } elseif (selectedHosts.included.ids.length > 0) {
```
jturel, line 140: I guess it should be 'else if' :roll_eyes:

jjeffers: yeah the linter smacked my hand

PR #8826, opened 2020-07-09
Refs #30354 - use public setting interface in tests

ezr-ondrej:
https://github.com/theforeman/foreman/pull/7702 is trying to move the unchanged settings to memory only.
That requires accessing the setting only through the public interface.

This is changing it wherever possible in tests, improving some while at it.


ezr-ondrej:
Ping, this should not cause any regressions in tests, it's just a different way to set settings during tests :upside_down_face:

ezr-ondrej:
Ping? :)

jturel:
Since it's a bit older, gonna re-test just to be safe, but LGTM!

[test katello]

PR #8832, opened 2020-07-10
Fixes #30368 - Make Candlepin CA file optional

ekohl:
This change makes the Candlepin CA file optional by falling back to the Foreman CA file. The result is a reduced configuration in most deployments.

For Pulp the certificate is now read in the same way as Candlepin. Previously it partly relied on the CA being in the global allowed CA. This may be an issue in some cases, but in the default deployment it isn't. Following the general SSL config makes the configuration more predictable for users. It can also be easier in a containerized setup or on a system where the admin is not allowed to modify the system CA certificates. The example config is now also consistent with reality.

The verify_ssl option is dropped from the ping model. This isn't respected elsewhere and it's misleading to have a valid ping only to have it fail at runtime.

### DIFF
--- a/app/lib/katello/http_resource.rb
+++ b/app/lib/katello/http_resource.rb
@@ -18,8 +18,8 @@ module Katello
 
     include Katello::Concerns::FilterSensitiveData
 
-    class_attribute :consumer_secret, :consumer_key, :ca_cert_file, :prefix, :site, :default_headers,
-                    :ssl_client_cert, :ssl_client_key
+    class_attribute :consumer_secret, :consumer_key, :prefix, :site, :default_headers,
+                    :ssl_client_cert, :ssl_client_key, :ssl_ca_file
 
     attr_reader :json
 
@@ -128,7 +128,7 @@ module Katello
                    :authorize_path => "",
                    :access_token_path => ""}
 
-        params[:ca_file] = self.ca_cert_file unless self.ca_cert_file.nil?
+        params[:ca_file] = self.ssl_ca_file unless self.ssl_ca_file.nil?
         # New OAuth consumer to setup signing the request
         consumer = OAuth::Consumer.new(self.consumer_key,
                             self.consumer_secret,
@@ -147,7 +147,7 @@ module Katello
           :open_timeout => SETTINGS[:katello][:rest_client_timeout],
           :timeout => SETTINGS[:katello][:rest_client_timeout]
         }
-        options[:ssl_ca_file] = self.ca_cert_file unless self.ca_cert_file.nil?
+        options[:ssl_ca_file] = self.ssl_ca_file unless self.ssl_ca_file.nil?
         options[:ssl_client_cert] = self.ssl_client_cert unless self.ssl_client_cert.nil?
         options[:ssl_client_key] = self.ssl_client_key unless self.ssl_client_key.nil?
 
--- a/app/lib/katello/resources/registry.rb
+++ b/app/lib/katello/resources/registry.rb
@@ -54,7 +54,6 @@ module Katello
             uri = URI.parse(content_app_url)
             self.prefix = "/pulpcore_registry/"
             self.site = "#{uri.scheme}://#{uri.host}:#{uri.port}"
-            self.ssl_ca_file = Setting[:ssl_ca_file]
             pulp_primary.pulp3_ssl_configuration(self, :net_http)
 
             self
--- a/app/lib/katello/resources/registry.rb
+++ b/app/lib/katello/resources/registry.rb
@@ -54,7 +54,7 @@ module Katello
             uri = URI.parse(content_app_url)
             self.prefix = "/pulpcore_registry/"
             self.site = "#{uri.scheme}://#{uri.host}:#{uri.port}"
-            self.ca_cert_file = Setting[:ssl_ca_file]
+            self.ssl_ca_file = Setting[:ssl_ca_file]
             pulp_primary.pulp3_ssl_configuration(self, :net_http)
 
             self
--- a/app/models/katello/ping.rb
+++ b/app/models/katello/ping.rb
@@ -252,12 +252,11 @@ module Katello
       end
 
       def backend_status(url, backend)
-        ca_file = SETTINGS[:katello][backend][:ca_cert_file]
+        ca_file = ::Cert::Certs.backend_ca_cert_file(backend)
         request_id = ::Logging.mdc['request']
 
         options = {}
         options[:ssl_ca_file] = ca_file unless ca_file.nil?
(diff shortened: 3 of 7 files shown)


ehelms:
Related, and based on some discussion with @jlsherrill , today Pulp communication relies on the the CA being trusted by the system (the whole use of trusted_ca module in the installer) and we should switch to doing this explicitly -- https://github.com/Katello/katello/blob/68ff922e84b66df05f2b7c767c0cd8f2a7328611/app/models/katello/concerns/smart_proxy_extensions.rb#L119

ekohl:
I took a stab at this. In doing so, I found 2 settings that were removed but still had references to them. It then also unifies the certificate handling in the certs service. There are a few more changes, documented in the commit message. Please read it and let me hear your thoughts.

app/lib/katello/resources/registry.rb
```diff
@@ -33,10 +33,10 @@ def load_class
               registry_url = uri.to_s
 
               # Assume the registry uses the same CA as the Smart Proxy
-              ca_cert_file = Setting[:ssl_ca_file]
+              ca_cert_file = ::Cert::Certs.backend_ca_cert_file(:pulp3)
             elsif (container_config = SETTINGS.dig(:katello, :container_image_registry))
               registry_url = container_config[:crane_url]
-              ca_cert_file = container_config[:registry_ca_cert_file]
```
ekohl, line 39: I debated this change. It suggests nobody uses this since the [installer sets `crane_ca_cert_file`](https://github.com/theforeman/puppet-katello/blob/2e15151a6d324149071c4e240bed78b5d977adf6/templates/katello.yaml.erb#L42)

app/services/cert/certs.rb
```diff
@@ -24,10 +24,14 @@ def self.ssl_client_key_filename
       Setting['pulp_client_key']
     end
 
+    def self.backend_ca_cert_file(backend)
+      SETTINGS.dig(:katello, backend, :ca_cert_file) || Setting[:ssl_ca_file]
```
ehelms, line 28: I honestly cannot think of a scenario where we would have or want a different CA cert for Candlepin or Pulp given our architecture. @jlsherrill can you think of any reason to not just set the CA cert used to Setting[:ssl_ca_file] and reduce the potential configuration options?

ekohl: I didn't know how breaking we would be if we dropped it immediately and if that's acceptable.

jlsherrill: unless there is some quirk related with a 3rd party CA i can't think of this should be fine.  I would imagine all sorts of things would not work properly if you started using different CAs for differnet subsystems.

ekohl: There is one case. Pulp 3 can use a different CA than the Smart Proxy. For example, one signed by a publicly trusted CA.

jturel:
[test katello]

app/models/katello/ping.rb
```diff
@@ -226,10 +226,11 @@ def fetch_proxy(capsule_id)
       end
 
       def backend_status(url, backend)
-        ca_file = SETTINGS[:katello][backend][:ca_cert_file]
         options = {}
-        options[:ssl_ca_file] = ca_file unless ca_file.nil?
-        options[:verify_ssl] = SETTINGS[:katello][backend][:verify_ssl] if SETTINGS[:katello][backend].key?(:verify_ssl)
+        if (ca_file = ::Cert::Certs.backend_ca_cert_file(backend))
+          options[:ssl_ca_file] = ca_file
```
jturel, line 231: Was this removed because we don't ever set the `verify_ssl` option for any of the backends?

ekohl: Indeed. It didn't make any sense to have separate logic for ping when the actual resources don't respect the setting.

jturel:
One test failure seems relevant

```
test_verify_ueber_cert_no_change – Katello::CertsTest
```

ekohl:
Rebased to resolve merge conflict and to get recent tests. I've also split of unrelated work to https://github.com/Katello/katello/pull/9268 since it doesn't impact this PR.

jturel:
Haven't tested yet but the test failure definitely seems related

ekohl:
Given stabilization is coming soon and my lack of time, I'm afraid we'll have to push this to the next Katello release.

parthaa:
@ekohl  is this PR still valid ? Rebase if so

ekohl:
I've tried to rebase it, but I'm a bit unclear about the current Pulp 2 vs Pulp 3 status. I still see traces of Runcible, but isn't that Pulp 2-only?

I now think this unifies everything. It does mean the system store isn't really used anymore (unless you don't configure anything I think), but that's probably a good thing.

Perhaps @ehelms and @evgeni could also think about what kind of behavior we would prefer to have from a platform point of view.

app/lib/katello/http_resource.rb
```diff
@@ -19,7 +19,7 @@ def initialize(params)
     include Katello::Concerns::FilterSensitiveData
 
     class_attribute :consumer_secret, :consumer_key, :ca_cert_file, :prefix, :site, :default_headers,
-                    :ssl_client_cert, :ssl_client_key
+                    :ssl_client_cert, :ssl_client_key, :ssl_ca_file
```
ekohl, line 22: I now see there already is `ca_cert_file` for this. I don't know what other classes can be passes into `pulp3_ssl_configuration` so I'm looking for guidance here.

ekohl: I included a commit to rename it for consistency with other tools and naming for the other `ssl_` attrs.

config/katello.yaml.example
```diff
@@ -17,22 +17,16 @@
...
+    # specified, Foreman's ssl_ca_file setting is used.
     :ca_cert_file:
     # :bulk_load_size: 1000
   # Setup your pulp environment here
   :pulp:
     :sync_threads: 4
     # refers to the apache certificate
-    # (typically /etc/pki/tls/certs/localhost.crt) location that is needed
-    # to connect to pulp over https.
+    # Optional CA file to user to verify HTTPS connections to Candlepin. If not
```
ehelms, line 28:
```suggestion
    # Optional CA file to user to verify HTTPS connections to Pulp. If not
```

ehelms (approved):
Clean re-factoring that helps set up future simplifications.

ekohl:
I had hoped @evgeni could test if this resolves his issue in containers before merging.

ehelms:
> I had hoped @evgeni could test if this resolves his issue in containers before merging.

I liked the change from a re-factoring stand point. If this doesn't solve his issue, then we can iterate on solving that.

evgeni:
https://github.com/theforeman/foreman-quadlet/pull/43 is not failing on SSL errors anymore, so it certainly helped :)

PR #8833, opened 2020-07-13
Bumps gem to resolve packaging dependency.

jjeffers:


jturel:
@jjeffers can you catch me up? how can I verify that this is the right change?

jturel:
The foreman 2.0 repos have foreman-tasks 1.1.1 which this change is disallowing. Is there gonna be a foreman-tasks 1.1.3 ?

jturel:
Ok, I get it now. This seems like it will be OK, but we need a version bump for the gem too

evgeni:
@jturel fwiw, 1.1.3 is packaged (today) so you'll see it in the repos tomorrow

PR #8837, opened 2020-07-14
Fixes #30409 - Require Remote Execution

jlsherrill:


jturel:
we need a packaging pr too right?

jlsherrill:
we do! although it looks like there is a react failure, ping @johnpmitsch !

jlsherrill:
But to clarify for something simple like this, usually we merge the gemspec change and then open a packaging pr

johnpmitsch:
> we do! although it looks like there is a react failure, ping @johnpmitsch !
Since we hit this [here](https://github.com/Katello/katello/pull/8813) too, I opened up a PR to address this everywhere https://github.com/Katello/katello/pull/8838

I believe the `waitFor` function should be the end of the road for the timeouts, its the only ones I've seen recently, we've increased the jest one and the  *knocks on wood*

jlsherrill:
[test katello]

PR #8840, opened 2020-07-16
Fixes #29742 - registries search and catalog

jjeffers:
This PR updates docker registry API responses to allow unauthenticated access to `/v2/_catalog` requests. Repositories with container images are included in these requests, the results are limited to repositories within lifecycle environments with "registry_unauthenticated_pull" set to true.

There is a potential for confusion with the podman client command "search": 
When used without any search terms, podman issues a request for the _catalog_ via `/v2/_catalog/` endpoint, which accepts token authentication. 

Podman defers to the deprecated `/v1_search` endpooint when using the "search" command with a search term. Podman **does not** supply an authentication token in this case, even if previously authenticated/logged in. For this reason, the "search" command with a search term can only show repositories where "regitstry_unauthenticated_pull" for that repository's LCE is set to true.

An example of how to test this:
Create 2 content views ("public" and "private"), adding 2 distinct docker repositories ("hello-world" and "busybox"), respectively. Publish the CVs, and then promote the version to the different LCEs.

Modify the "public" LCE so the 'registry_unauthenticated_pull" is set to true.

Then, attempt search command unauthenticated with no search terms (a "catalog" request"):
```
[jjeffers@localhost forklift]$ podman logout -a
Removed login credentials for all registries
[jjeffers@localhost forklift]$ podman search --tls-verify=false centos7-katello-devel-stable.example.com/
INDEX         NAME                                                                                               DESCRIPTION   STARS   OFFICIAL   AUTOMATED
example.com   centos7-katello-devel-stable.example.com/default_organization-public-cv2-docker_test-hello-world   example.com   0  
```
Next, try the same command with a search term:
```
[jjeffers@localhost forklift]$ podman search --tls-verify=false centos7-katello-devel-stable.example.com/hello-world
INDEX         NAME                                                                                               DESCRIPTION   STARS   OFFICIAL   AUTOMATED
example.com   centos7-katello-devel-stable.example.com/default_organization-public-cv2-docker_test-hello-world   example.com   0 
```
And the same query, but with a search term for an image in the restricted LCE:
```podman search --tls-verify=false centos7-katello-devel-stable.example.com/busybox```
This should return no results.

Repeat the commands but after logging in via podman (and establishing an authenticated token):
```
podman login --tls-verify=false centos7-katello-devel-stable.example.com -u admin -p changeme
Login Succeeded!
[jjeffers@localhost forklift]$ podman search --tls-verify=false centos7-katello-devel-stable.example.com/
INDEX         NAME                                                                                                DESCRIPTION   STARS   OFFICIAL   AUTOMATED
example.com   centos7-katello-devel-stable.example.com/default_organization-docker_test-busybox                   example.com   0                  
example.com   centos7-katello-devel-stable.example.com/default_organization-library-cv1-docker_test-busybox       example.com   0                  
example.com   centos7-katello-devel-stable.example.com/default_organization-cv1-1_0-docker_test-busybox           example.com   0                  
example.com   centos7-katello-devel-stable.example.com/default_organization-private-cv1-docker_test-busybox       example.com   0                  
example.com   centos7-katello-devel-stable.example.com/default_organization-docker_test-hello-world               example.com   0                  
example.com   centos7-katello-devel-stable.example.com/default_organization-library-cv2-docker_test-hello-world   example.com   0                  
example.com   centos7-katello-devel-stable.example.com/default_organization-cv2-1_0-docker_test-hello-world       example.com   0                  
example.com   centos7-katello-devel-stable.example.com/default_organization-public-cv2-docker_test-hello-world    example.com   0 
[jjeffers@localhost forklift]$ podman search --tls-verify=false centos7-katello-devel-stable.example.com/hello-world
INDEX         NAME                                                                                               DESCRIPTION   STARS   OFFICIAL   AUTOMATED
example.com   centos7-katello-devel-stable.example.com/default_organization-public-cv2-docker_test-hello-world   example.com   0 
```
Note that re-issuing the `podman search --tls-verify=false centos7-katello-devel-stable.example.com/busybox` returns no results. This is due to podman not sending the search request with an authentication token.
```

### DIFF
--- a/app/controllers/katello/api/registry/registry_proxies_controller.rb
+++ b/app/controllers/katello/api/registry/registry_proxies_controller.rb
@@ -6,8 +6,8 @@ module Katello
     before_action :confirm_push_settings, only: [:start_upload_blob, :upload_blob, :finish_upload_blob,
                                                  :chunk_upload_blob, :push_manifest]
     skip_before_action :authorize
-    before_action :optional_authorize, only: [:token]
-    before_action :registry_authorize, except: [:token, :v1_search]
+    before_action :optional_authorize, only: [:token, :catalog]
+    before_action :registry_authorize, except: [:token, :v1_search, :catalog]
     before_action :authorize_repository_read, only: [:pull_manifest, :tags_list]
     before_action :authorize_repository_write, only: [:push_manifest]
     skip_before_action :check_content_type, only: [:start_upload_blob, :upload_blob, :finish_upload_blob,
@@ -39,20 +39,7 @@ module Katello
                                              "scope=\"repository:registry:pull,push\""
     end
 
-    def optional_authorize
-      @repository = find_scope_repository
-      if @repository && (@repository.environment.registry_unauthenticated_pull || ssl_client_authorized?(@repository.organization.label))
-        true
-      else
-        authorize
-      end
-    end
-
-    def registry_authorize
-      @repository = find_readable_repository
-      return true if ['GET', 'HEAD'].include?(request.method) && @repository && !require_user_authorization?
-
-      token = request.headers['Authorization']
+    def set_user_by_token(token, redirect_on_failure = true)
       if token
         token_type, token = token.split
         if token_type == 'Bearer' && token
@@ -63,10 +50,34 @@ module Katello
           end
         elsif token_type == 'Basic' && token
           return true if authorize
-          redirect_authorization_headers
+          redirect_authorization_headers if redirect_on_failure
           return false
         end
       end
+      false
+    end
+
+    def optional_authorize
+      @repository = find_scope_repository
+      if @repository && (@repository.environment.registry_unauthenticated_pull || ssl_client_authorized?(@repository.organization.label))
+        true
+      elsif params['action'] == 'catalog'
+        set_user_by_token(request.headers['Authorization'], false)
+      elsif (params['action'] == 'token' && params['scope'].blank? && params['account'].blank?)
+        true
+      else
+        authorize
+      end
+    end
+
+    def registry_authorize
+      @repository = find_readable_repository
+      return true if ['GET', 'HEAD'].include?(request.method) && @repository && !require_user_authorization?
+
+      is_user_set = set_user_by_token(request.headers['Authorization'])
+
+      return true if is_user_set
+
       redirect_authorization_headers
       render_error('unauthorized', :status => :unauthorized)
       return false
@@ -87,12 +98,8 @@ module Katello
     # Also include repositories in lifecycle environments with registry_unauthenticated_pull=true
     def readable_repositories
       table_name = Repository.table_name
-      in_products = Repository.in_product(Katello::Product.authorized(:view_products)).select(:id)
-      in_environments = Repository.where(:environment_id => Katello::KTEnvironment.authorized(:view_lifecycle_environments)).select(:id)
       in_unauth_environments = Repository.joins(:environment).where("#{Katello::KTEnvironment.table_name}.registry_unauthenticated_pull" => true).select(:id)
-      in_content_views = Repository.joins(:content_view_repositories).where("#{ContentViewRepository.table_name}.content_view_id" => Katello::ContentView.readable).select(:id)
-      in_versions = Repository.joins(:content_view_version).where("#{Katello::ContentViewVersion.table_name}.content_view_id" => Katello::ContentView.readable).select(:id)
-      Repository.where("#{table_name}.id in (?) or #{table_name}.id in (?) or #{table_name}.id in (?) or #{table_name}.id in (?) or #{table_name}.id in (?)", in_products, in_content_views, in_versions, in_environments, in_unauth_environments)
+      Repository.readable.or(Repository.joins(:root).where("#{table_name}.id in (?)", in_unauth_environments))
     end
 
     def find_readable_repository
@@ -105,7 +112,8 @@ module Katello
     end
 
     def require_user_authorization?(repository = @repository)
-      !repository || (!repository.environment.registry_unauthenticated_pull && !ssl_client_authorized?(repository.organization.label))
+      !(params['action'] == 'token' && params['scope'].blank? && params['account'].blank?) &&
+        (!repository || (!repository.environment.registry_unauthenticated_pull && !ssl_client_authorized?(repository.organization.label)))
     end
 
     def ssl_client_authorized?(org_label)
@@ -293,6 +301,14 @@ module Katello
     end
 
     def v1_search
+      # Checks for podman client and issues a 404 in that case. Podman
+      # examines the response from a /v1_search request. If the result
+      # is a 4XX, it will then proceed with a request to /_catalog
+      if request.headers['HTTP_USER_AGENT'].downcase.include?('libpod')
+        render json: {}, status: :not_found
+        return
+      end
+
       authenticate # to set current_user, not to enforce
       options = {
         resource_class: Katello::Repository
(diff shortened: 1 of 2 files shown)


app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -43,6 +43,20 @@ def optional_authorize
...
+      elsif params['action'] == 'catalog'
+        token = request.headers['Authorization']
+        if token
+          token_type, token = token.split
+          if token_type == 'Bearer' && token
+            personal_token = PersonalAccessToken.find_by_token(token)
+            if personal_token && !personal_token.expired?
+              User.current = User.unscoped.find(personal_token.user_id)
+            end
+          end
```
jlsherrill, line 55: This looks heavily duplicated from the registry_authorize method, could we combine these two into one?  (I also notice this isn't handling the 'basic' token, whatever that is, so one method that handles all tokens (optionally) would be good i think

jlsherrill:
>  Podman does not supply an authentication token in this case, even if previously authenticated/logged in. For this reason, the "search" command with a search term can only show repositories where "regitstry_unauthenticated_pull" for that repository's LCE is set to true.

This seems to indicate that it should:  https://github.com/containers/podman/issues/5405
Trying locally i def. don't see the basic auth credentials being sent up, so i wonder if its something we're doing wrong.  I'm gonna dig around in the podman source and see if i can figure it out.

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -6,8 +6,8 @@ class Api::Registry::RegistryProxiesController < Api::V2::ApiController
     before_action :confirm_push_settings, only: [:start_upload_blob, :upload_blob, :finish_upload_blob,
                                                  :chunk_upload_blob, :push_manifest]
     skip_before_action :authorize
-    before_action :optional_authorize, only: [:token]
-    before_action :registry_authorize, except: [:token, :v1_search]
+    before_action :optional_authorize, only: [:token, :catalog]
+    before_action :registry_authorize, except: [:token, :v1_search, :catalog]
```
jlsherrill, line 10: while we are here, the '/v2/' endpoint seems to require auth which isn't really intentional, the clients seem to ignore the 401.  Can you add :ping to this except: []   to fix that?

jlsherrill: actually.... docker hub does this too, so maybe that excepted, ignore for now i guess

app/controllers/katello/api/registry/registry_proxies_controller.rb
jjeffers: It is duplicated and consolidating the logic would be a good thing.

app/controllers/katello/api/registry/registry_proxies_controller.rb
jjeffers: I wasn't sure why docker client did this too, which was confusing. Based on the podman discussions it seems as if the design goal was to match docker client as closely as possible. Maybe they'll fix this eventually?

jlsherrill:
i didn't have a ton of luck building podman from source (in order to add some debug logging), so i took a different route and setup my own private registry according to https://docs.docker.com/registry/deploying/ 

I got it up and running requiring basic auth and podman search worked as it claims to send auth with 'podman search'.  After digging around and trying a few things, i added some debug logging against the private registry:

podman  --log-level=debug search devel2.balmora.example.com:5001/dock

```
DEBU[0000] GET https://devel2.balmora.example.com:5001/v2/ 
DEBU[0000] Ping https://devel2.balmora.example.com:5001/v2/ status 401 
DEBU[0000] GET https://devel2.balmora.example.com:5001/v1/search?n=25&q=dock 
DEBU[0000] error getting search results from v1 endpoint "devel2.balmora.example.com:5001": invalid status code from registry 404 (Not Found) 
DEBU[0000] trying to talk to v2 search endpoint         
DEBU[0000] GET https://devel2.balmora.example.com:5001/v2/_catalog 
```

Notice that v1 search threw a 404!   Podman then fetched the full catalog and filtered the results client side.  I removed the v1 search route within katello and then saw a similar behavior with katello's registry.    "Docker search" doesn't do this and seems just to error with a 404 if v1/search isn't there.  (This means that docker search does not work at all against any 'official' private registry)

So it seems that its not possible to support 'docker search', since that *only* talks to the v1/search api and also support 'podman search' since that seems to somewhat depend on it to not exist.  Another option is to add a setting that would toggle the v1/search api being on or off.  I kinda like having a setting to turn it on or off, but i kinda just like removing it and putting our support behind podman.    Curious if you have any thoughts!

jjeffers:
Based on what I have read it seems as if it would be easier on us to remove the v1_search support, especially if podman falls back on the catalog functionality. 

That being said, I can see this causing issues with clients who still use docker client. So perhaps we can make v1_search an option, but communicate that we will deprecate it at some point.

jlsherrill:
@jjeffers do you think a Setting would be best to control it? or do you think there's some better way?

jjeffers:
> @jjeffers do you think a Setting would be best to control it? or do you think there's some better way?
Let's try a setting. I'll also open a topic for discussion to gauge community response to this proposed change.

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -293,6 +301,11 @@ def v1_ping
     end
 
     def v1_search
+      if SETTINGS[:katello][:block_registry_v1_search]
+        render json: {}, status: :not_found
+        return
+      end
```
jlsherrill, line 307: I would suggest making this a Setting instead of a SETTING (defining it in app/models/settings/content.rb)   Anything that is a SETTING requires installer updates to be able to consistently set it.

jlsherrill: Also if its a Setting, i'd probably recommend we change it to be positive instead of negative (allow_v1_docker_search) and default it to true.

jlsherrill: So while i was looking at this just now, i realized we might could use the HTTP_USER_AGENT to determien whether to throw a 404 or not.  Thoughts?

From docker:
```
172.17.0.1 - - [27/Jul/2020:17:37:50 +0000] "GET /v1/search?q=&n=25 HTTP/1.1" 404 19 "" "docker/1.13.1 go/go1.12.7 kernel/5.6.13-100.fc30.x86_64 os/linux arch/amd64 UpstreamClient(Docker-Client/1.13.1 \\(linux\\))"
```


from podman:
```
172.17.0.1 - - [27/Jul/2020:17:41:54 +0000] "GET /v1/search?n=25&q=foo HTTP/1.1" 404 19 "" "libpod/1.8.0"
```

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -105,7 +112,8 @@ def find_readable_repository
     end
 
     def require_user_authorization?(repository = @repository)
-      !repository || (!repository.environment.registry_unauthenticated_pull && !ssl_client_authorized?(repository.organization.label))
+      !(params['action'] == 'token' && params['scope'].blank? && params['account'].blank?) &&
```
jlsherrill, line 115: what was the reason to add this line?

jjeffers: It was to allow for unscoped requests to the `token` endpoint, so that unauthenticated requests could proceed to the `_catalog` method.

jjeffers: The original version specifically prohibited unscoped token requests.

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -293,6 +301,11 @@ def v1_ping
     end
 
     def v1_search
+      if request.headers['HTTP_USER_AGENT'].downcase.include?('libpod')
```
jlsherrill, line 304: Everything looks good and seems to work well.  Could we add a comment here explaining that podman checks explicitly for a 404 before trying to call /catalog  ?

jjeffers: You bet!

PR #8841, opened 2020-07-16
Fixes #30425 - fix page redirect during smart proxy sync

jlsherrill:
This was likely caused by the recent upgrade to angular,
although the exact root cause is not fully clear.  I believe
this code was meant to handle a link to a smart proxy page from
another bastion page, however this did not even work on older
builds (where this bug was not present).  No such links exist
today, but if you try to add one, it will not work.

Thus this change does not cause any regressions, but if we
ever add a link to the smart proxy details page from a bastion page
it will (still) not redirect to that page.


jlsherrill:
To test:

1) setup a smart proxy with content, assigned to a LCE
2) go to Infrastructure > smart proxies
3) click a smart proxy
4) sync the smart proxy
5)  while its syncing, click one of the tabs on that page such as 'logs'

Without this bug fix, the page will refresh and go back to the main tab.  With this fix it will not do a full refresh at all.

jlsherrill:
[test katello]

PR #8843, opened 2020-07-20
Fixes #30448 - Initialize event daemon in to_prepare

jturel:
Sometimes reloading the code in dev will put the event daemon in a weird state because the related classes haven't been initialized. It appears to be related to when we load the classes.

This change:

- Moves the CandlepinEventListener configuration out of its own initializer and into `config.to_prepare` so that it happens upon each code reload and not just when the app starts. It avoids this error:

```
undefined method `call' for nil:NilClass
/home/vagrant/katello/app/services/katello/candlepin_event_listener.rb:36:in `run' 
```

- Moves the `EventDaemon.initialize` and related code to the top of the `config.to_prepare` block. In the event of bad code being loaded (ie syntax error), it prevents the following error from being raised:

```
NoMethodError (undefined method `fetch' for nil:NilClass):
| /home/vagrant/katello/app/services/katello/event_daemon.rb:123:in `runnable?'     
```

- Decouples the desired services from `EventDaemon` by providing a `register_service` method similar to the `EventQueue` class. This is a totally unrelated change, but since I was in the area I thought it was a good improvement to make.

Overall, none of this affects the runtime behavior of the event daemon and you can test that events are still processed from the rails console (or however you like):

```
Katello::Resources::Candlepin::Consumer.update(::Host.first.subscription_facet.uuid, {'role' => SecureRandom.uuid})
Katello::EventQueue.push_event('generate_host_applicability', 100)
```

### DIFF
--- a/app/services/katello/event_daemon.rb
+++ b/app/services/katello/event_daemon.rb
@@ -1,5 +1,8 @@
 module Katello
   class EventDaemon
+    @services = {}
+    @cache = ActiveSupport::Cache::MemoryStore.new
+
     class Monitor
       def initialize(service_classes)
         @service_classes = service_classes
@@ -44,7 +47,6 @@ module Katello
     class << self
       def initialize
         FileUtils.touch(lock_file)
-        @cache = ActiveSupport::Cache::MemoryStore.new
       end
 
       def settings
@@ -80,7 +82,7 @@ module Katello
         return unless pid == Process.pid
         @monitor_thread.kill
         @cache.clear
-        services.values.each(&:close)
+        @services.values.each(&:close)
         File.unlink(pid_file) if pid_file && File.exist?(pid_file)
       end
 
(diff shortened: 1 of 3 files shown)


jjeffers:
Tested this against a development image, after also updating for the issue resolved in https://github.com/theforeman/puppet-katello_devel/pull/241.

Found that before applying this PR, reloading application in the console would fail as described. Also noticed that the CandlepinEventListener.client_factory would evaluate with a nil result.

Applied the PR and retested. Code reloading didn't generate the errors. Checked CandlepinEventListener.client_factory and it evaluated to a proc. I wasn't able to generate the second error ("NoMethodError (undefined method `fetch' for nil:NilClass") before applying the PR, but it seems as if new initialization approach would resolve that.

jturel:
Thanks @jjeffers - can I get an ACK?

PR #8844, opened 2020-07-20
Fixes #30451 - attach subscription to unregstered host gives useful e…

piyushtiw:
…rror message

Issue => Attach bulk subscription api is giving following error -
Internal Service Error (ISE)
NoMethodError: undefined method `candlepin_consumer' for nil:NilClass

Fix=> returning useful error message.

### DIFF
--- a/app/lib/actions/katello/host/attach_subscriptions.rb
+++ b/app/lib/actions/katello/host/attach_subscriptions.rb
@@ -8,12 +8,16 @@ module Actions
 
         def plan(host, pools_with_quantities_params)
           action_subject(host)
+
+          subscription_facet = host.subscription_facet
+          fail _("Register host '%s' before attaching subscriptions") % host.name unless subscription_facet
+
           sequence do
             pool_ids = []
             pools_with_quantities = pools_with_quantities_params.map do |pool_with_quantity|
               ::Katello::PoolWithQuantities.fetch(pool_with_quantity)
             end
-            existing_pool_ids = host.subscription_facet.candlepin_consumer.pool_ids
+            existing_pool_ids = subscription_facet.candlepin_consumer.pool_ids
             pools_with_quantities.each do |pool_with_quantities|
               unless existing_pool_ids.include?(pool_with_quantities.pool.cp_id.to_s)
                 pool_ids << pool_with_quantities.pool.id
--- a/test/actions/katello/host/attach_subscriptions_test.rb
+++ b/test/actions/katello/host/attach_subscriptions_test.rb
@@ -15,7 +15,7 @@ module Katello::Host
     describe 'attach subscriptions' do
       let(:action_class) { ::Actions::Katello::Host::AttachSubscriptions }
 
-      it 'plans' do
+      it 'plans success' do
         action = create_action action_class
         action.expects(:action_subject).with(@host)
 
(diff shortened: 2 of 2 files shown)


test/actions/katello/host/attach_subscriptions_test.rb
```diff
@@ -33,5 +33,22 @@ class AttachSubscriptionsTest < ActiveSupport::TestCase
...
+    describe 'attach subscriptions failure' do
+      let(:action_class) { ::Actions::Katello::Host::AttachSubscriptions }
+
+      it 'plans' do
+        action = create_action action_class
+        action.expects(:action_subject).with(@host)
+
+        @host.expects(:subscription_facet).returns(nil)
+
+        pools_with_quantities = [::Katello::PoolWithQuantities.new(@pool, [1, 2])]
```
jturel, line 46: For this test I don't think the pools_with_quantities comes into play at all since we error out earlier. IMO we can omit this

app/lib/actions/katello/host/attach_subscriptions.rb
```diff
@@ -8,12 +8,16 @@ def queue
 
         def plan(host, pools_with_quantities_params)
           action_subject(host)
+
+          subscription_facet = host.subscription_facet
+          fail _("Subscription cannot be attached to the host because of no subscription facet.") unless subscription_facet
```
jturel, line 13: The BZ explains how this action was called through the bulk subscription attach API. Because of that I think we should include the name of the host in the error message, so the user knows which one is unregistered.

`_("Please register host '%s' before attaching subscriptions") % host.name`

parthaa: Wouldn't that be clear on the dynflow console? May be useful in hammer though

jturel: Would it? I think the dynflow console is extremely hit or miss with the information it contains. Either way, this is the kind of info that should come out of the API. It's a bad experience to have to dig into dynflow to find out what went wrong and the best thing we can do is make users as unaware of dynflow as is possible.

app/lib/actions/katello/host/attach_subscriptions.rb
```diff
@@ -8,12 +8,16 @@ def queue
 
         def plan(host, pools_with_quantities_params)
           action_subject(host)
+
+          subscription_facet = host.subscription_facet
+          fail _("Please register host '%s' before attaching subscriptions") % host.name unless subscription_facet
```
jturel, line 13: Functionality works great, but looking at the wording again - let's remove the word 'please'. Sorry for suggesting it :)

test/actions/katello/host/attach_subscriptions_test.rb
```diff
@@ -33,5 +33,22 @@ class AttachSubscriptionsTest < ActiveSupport::TestCase
                                           :quantity => 2, :pool_uuid => @pool.cp_id
       end
     end
+
+    describe 'attach subscriptions failure' do
+      let(:action_class) { ::Actions::Katello::Host::AttachSubscriptions }
```
jturel, line 38: This `let` is duplicated between this test and the test above. you can move it to the outer block to be shared by both :)

PR #8845, opened 2020-07-21
Fixes #30452 - Improve errata status reporting

jturel:
I've seen a few threads on discourse that have questioned the Errata Status messaging. This PR aims to clarify that messaging by separating out the cases:
 - host is unregistered
 - no package information reported
 - no enabled repo information reported

https://community.theforeman.org/t/strange-behaviour-with-host-status-and-the-sync-fixing-my-issue-but-then-showing-zero-packages/19673

### DIFF
--- a/app/models/katello/errata_status.rb
+++ b/app/models/katello/errata_status.rb
@@ -9,17 +9,28 @@ module Katello
       N_("Errata")
     end
 
+    def errata_status_installable
+      @installable ||= Setting[:errata_status_installable]
+    end
+
+    def profiles_reporter_package_name
+      if host.content_facet.host_tools_installed?
+        Katello::Host::ContentFacet::HOST_TOOLS_PACKAGE_NAME
+      else
+        Katello::Host::ContentFacet::SUBSCRIPTION_MANAGER_PACKAGE_NAME
+      end
+    end
+
     def to_label(_options = {})
-      installable = Setting[:errata_status_installable]
       case status
       when NEEDED_SECURITY_ERRATA
-        installable ? N_("Security errata installable") : N_("Security errata applicable")
+        errata_status_installable ? N_("Security errata installable") : N_("Security errata applicable")
       when NEEDED_ERRATA
-        installable ? N_("Non-security errata installable") : N_("Non-security errata applicable")
+        errata_status_installable ? N_("Non-security errata installable") : N_("Non-security errata applicable")
       when UP_TO_DATE
         N_("All errata applied")
       when UNKNOWN
-        N_("Could not calculate errata status, ensure host is registered and the katello-host-tools package is installed")
+        N_("No installed packages and/or enabled repositories have been reported by %s." % profiles_reporter_package_name)
       else
         N_("Unknown errata status")
       end
@@ -41,19 +52,17 @@ module Katello
     end
 
     def to_status(_options = {})
-      return UNKNOWN if host.content_facet.nil?
-
-      if Setting[:errata_status_installable]
-        errata = host.content_facet.try(:installable_errata)
-      else
-        errata = host.content_facet.try(:applicable_errata)
-      end
+      errata = if errata_status_installable
+                 host.content_facet.try(:installable_errata)
+               else
+                 host.content_facet.try(:applicable_errata)
+               end
 
       if errata.security.any?
         NEEDED_SECURITY_ERRATA
       elsif errata.any?
         NEEDED_ERRATA
-      elsif host.content_facet.bound_repositories.empty?
+      elsif host.installed_packages.empty? || host.content_facet.bound_repositories.empty?
         UNKNOWN
       else
         UP_TO_DATE
--- a/app/models/katello/host/content_facet.rb
+++ b/app/models/katello/host/content_facet.rb
@@ -5,6 +5,9 @@ module Katello
       self.table_name = 'katello_content_facets'
       include Facets::Base
 
+      HOST_TOOLS_PACKAGE_NAME = 'katello-host-tools'.freeze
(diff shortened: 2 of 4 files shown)


app/models/katello/errata_status.rb
```diff
@@ -9,17 +10,30 @@ def self.status_name
       N_("Errata")
     end
 
+    def errata_status_installable
+      Setting[:errata_status_installable]
+    end
+
+    def profiles_reporter_package_name
```
jturel, line 17: This check isn't 100% accurate, but is more precise than it was. It could be improved by checking that the subscription-manager that's installed is >= the version which introduced package+repo profile reporting

app/models/katello/errata_status.rb
```diff
@@ -53,7 +69,7 @@ def to_status(_options = {})
         NEEDED_SECURITY_ERRATA
       elsif errata.any?
         NEEDED_ERRATA
-      elsif host.content_facet.bound_repositories.empty?
+      elsif host.content_facet.bound_repositories.empty? || host.installed_packages.empty?
```
jturel, line 72: I could still see this being split out into separate messages for 
- missing package data
- missing repositories data

jturel:
@jjeffers @jlsherrill you guys think this is a worthwhile clarification to make?

app/models/katello/errata_status.rb
```diff
@@ -9,17 +10,30 @@ def self.status_name
...
       when NEEDED_ERRATA
-        installable ? N_("Non-security errata installable") : N_("Non-security errata applicable")
+        errata_status_installable ? N_("Non-security errata installable") : N_("Non-security errata applicable")
       when UP_TO_DATE
         N_("All errata applied")
+      when UNREGISTERED
+        N_("Could not calculate errata status because the host is not registered.")
       when UNKNOWN
-        N_("Could not calculate errata status, ensure host is registered and the katello-host-tools package is installed")
+        N_("Could not calculate errata status because %s has not reported any package or repository information." % profiles_reporter_package_name)
```
jlsherrill, line 36: I think this is very valuable and i've wanted to add something to help 'debug' applicability issues, this is a good first step.  

One issue though is that this saying that it hasn't reported package or repo information, but thats not actually what we're checking.  I think it would be more valuable to report specifically that 'no package' or 'no repo' information has been reported based off the actual contents of those tables.

app/models/katello/errata_status.rb
jlsherrill: oh yes, i mentioned that above, i think its valuable to do that!

app/models/katello/concerns/content_facet_host_extensions.rb
```diff
@@ -8,7 +8,9 @@ module ContentFacetHostExtensions
           :security_needed => Katello::ErrataStatus::NEEDED_SECURITY_ERRATA,
           :errata_needed => Katello::ErrataStatus::NEEDED_ERRATA,
           :updated => Katello::ErrataStatus::UP_TO_DATE,
-          :unknown => Katello::ErrataStatus::UNKNOWN
```
jturel, line 11: This is basically an API change. I could keep 'unknown' but map it to ::UNREGISTERED?

app/models/katello/errata_status.rb
jturel: @jlsherrill Updated. I think I captured what you meant although the wording is a little nuanced. Let me know if it's what you were thinking / clear enough.

jlsherrill:
This all looks good to me from a code perspective, i'll leave it to @jjeffers to test (as per discussion)

app/models/katello/errata_status.rb
```diff
@@ -33,17 +50,21 @@ def to_global(_options = {})
...
+      when UNREGISTERED
         ::HostStatus::Global::WARN
       else
         ::HostStatus::Global::WARN
       end
     end
 
     def to_status(_options = {})
-      return UNKNOWN if host.content_facet.nil?
+      return UNREGISTERED if host.content_facet.nil?
```
jturel, line 65: @jlsherrill When the content facet has no UUID this status is considered irrelevant so it won't be shown in the UI. I think that makes this check redundant and therefore the UNKNOWN/UNREGISTERED status unnecessary. Am I not thinking of some (edge)case?

jjeffers:
@jlsherrill @jturel Thanks to this PR, we actually were able to discover another issue with [errata status updates in pulp3](https://github.com/Katello/katello/pull/8887). 

After applying the changes in that PR, I was able to generate different errata status states for a registered RHEL content host. The messages appeared and worked as described.

app/models/katello/errata_status.rb
jlsherrill: As long as it also doesn't affect the global status in that case, i think you are correct

jturel:
@jjeffers after our conversation yesterday I opted to take a step back on this PR. Previously I'd removed UNKNOWN and added NO_PACKAGES and NO_REPOSITORIES states. After messing around with the errata_status search in the Content Hosts list I found that it's useful to have to only search by one (ie unknown) status rather than two to find problematic hosts. It also felt somewhat wrong to effectively break the API by removing the UNKNOWN state (even though I intended to fix that here) and I couldn't work around it easily. Now I've just added clearer messaging to the UNKNOWN status. Let me know what you think of the wording!

I'll file a separate PR for the issue we discussed yesterday

jjeffers:
>  Let me know what you think of the wording!
> 
> I'll file a separate PR for the issue we discussed yesterday

The wording is fine. It's clear what the process detects as "unknown" here for errata status.

Sounds like a separate PR is prudent. I look forward to what you come up with.

PR #8850, opened 2020-07-23
Fixes #30479 - API to export and use pulp3

parthaa:
This commit creates the initial functionality to export a content view
version via the pulp3 exporter.
Things in this commit
1) Exporter updates to point to the right pulp3 api
2) After generating tar you should be able to access the location in the
   foreman task.

### DIFF
--- a/app/controllers/katello/api/v2/content_view_versions_controller.rb
+++ b/app/controllers/katello/api/v2/content_view_versions_controller.rb
@@ -84,31 +84,39 @@ module Katello
 
     api :POST, "/content_view_versions/:id/export", N_("Export a content view version"), :deprecated => true
     param :id, :number, :desc => N_("Content view version identifier"), :required => true
-    param :export_to_iso, :bool, :desc => N_("Export to ISO format"), :required => false
-    param :iso_mb_size, :number, :desc => N_("maximum size of each ISO in MB"), :required => false
-    param :since, Date, :desc => N_("Optional date of last export (ex: 2010-01-01T12:00:00Z)"), :required => false
+    param :export_to_iso, :bool, :desc => N_("Export to ISO format. Relevant only for Pulp 2 repositories"), :required => false
+    param :iso_mb_size, :number, :desc => N_("maximum size of each ISO in MB. Relevant only for Pulp 2 repositories"), :required => false
+    param :since, Date, :desc => N_("Optional date of last export (ex: 2010-01-01T12:00:00Z). Relevant only for Pulp 2 repositories"), :required => false
     def export
-      ::Foreman::Deprecation.api_deprecation_warning("Export is being deprecated and will be removed in a future version of Katello. Use hammer content-view version export instead.")
-      if params[:export_to_iso].blank? && params[:iso_mb_size].present?
-        fail HttpErrors::BadRequest, _("ISO export must be enabled when specifying ISO size")
-      end
-
-      if (repos = @version.content_view.on_demand_repositories).any?
-        fail HttpErrors::BadRequest, _("This content view has on demand repositories that cannot be exported: %{repos}" % {repos: repos.pluck(:label).join(", ")})
-      end
-
-      if params[:since].present?
-        begin
-          params[:since].to_datetime
-        rescue
-          raise HttpErrors::BadRequest, _("Invalid date provided.")
+      if SmartProxy.pulp_master.pulp3_repository_type_support?(Katello::Repository::YUM_TYPE)
+        invalid_params = [:export_to_iso, :iso_mb_size, :since].reject { |param| params[param].blank? }
+        unless invalid_params.empty?
+          fail HttpErrors::BadRequest, _("Invalid parameters provided - %{params}. These are only relevant for Pulp 2 repositories" % { params: invalid_params.join(', ')})
         end
+        task = async_task(::Actions::Pulp3::Orchestration::ContentViewVersion::Export, @version)
+      else
+        ::Foreman::Deprecation.api_deprecation_warning("Export is being deprecated and will be removed in a future version of Katello. Use hammer content-view version export instead.")
+        if params[:export_to_iso].blank? && params[:iso_mb_size].present?
+          fail HttpErrors::BadRequest, _("ISO export must be enabled when specifying ISO size")
+        end
+
+        if (repos = @version.content_view.on_demand_repositories).any?
+          fail HttpErrors::BadRequest, _("This content view has on demand repositories that cannot be exported: %{repos}" % {repos: repos.pluck(:label).join(", ")})
+        end
+
+        if params[:since].present?
+          begin
+            params[:since].to_datetime
+          rescue
+            raise HttpErrors::BadRequest, _("Invalid date provided.")
+          end
+        end
+        task = async_task(::Actions::Katello::ContentViewVersion::Export, @version,
+                          ::Foreman::Cast.to_bool(params[:export_to_iso]),
+                          params[:since].try(:to_datetime),
+                          params[:iso_mb_size])
       end
 
-      task = async_task(::Actions::Katello::ContentViewVersion::Export, @version,
-                        ::Foreman::Cast.to_bool(params[:export_to_iso]),
-                        params[:since].try(:to_datetime),
-                        params[:iso_mb_size])
       respond_for_async :resource => task
     end
 
--- a/app/lib/actions/pulp3/content_view_version/export.rb
+++ b/app/lib/actions/pulp3/content_view_version/export.rb
@@ -1,0 +1,19 @@
+module Actions
+  module Pulp3
+    module ContentViewVersion
+      class Export < Pulp3::AbstractAsyncTask
+        input_format do
+          param :content_view_version_id, Integer
+          param :smart_proxy_id, Integer
+          param :exporter_data, Hash
+        end
+
+        def invoke_external_task
+          cvv = ::Katello::ContentViewVersion.find(input[:content_view_version_id])
+          ::Katello::Pulp3::ContentViewVersion.new(smart_proxy: smart_proxy,
+                                                   content_view_version: cvv).create_export(input[:exporter_data][:pulp_href])
+        end
+      end
+    end
+  end
+end
--- a/app/lib/actions/pulp3/orchestration/content_view_version/export.rb
+++ b/app/lib/actions/pulp3/orchestration/content_view_version/export.rb
@@ -1,0 +1,60 @@
+module Actions
+  module Pulp3
+    module Orchestration
+      module ContentViewVersion
+        class Export < Actions::EntryAction
+          input_format do
+            param :smart_proxy_id, Integer
+            param :exporter_data, Hash
+          end
+
+          output_format do
+            param :exported_file_name, String
+            param :exported_file_checksum, String
+          end
+
+          def plan(content_view_version)
+            action_subject(content_view_version)
+            unless File.directory?(Setting['pulpcore_export_destination'])
+              fail ::Foreman::Exception, N_("Unable to export. 'pulpcore_export_destination' setting is not set to a valid directory.")
+            end
+
+            sequence do
+              smart_proxy = SmartProxy.pulp_master!
+              action_output = plan_action(::Actions::Pulp3::ContentViewVersion::CreateExporter,
+                                     content_view_version_id: content_view_version.id,
+                                     smart_proxy_id: smart_proxy.id).output
+
+              plan_action(::Actions::Pulp3::ContentViewVersion::Export,
+                                     content_view_version_id: content_view_version.id,
+                                     smart_proxy_id: smart_proxy.id,
+                                     exporter_data: action_output[:exporter_data]
+                                     )
+
+              plan_self(exporter_data: action_output[:exporter_data], smart_proxy_id: smart_proxy.id)
+
+              plan_action(::Actions::Pulp3::ContentViewVersion::DestroyExporter,
+                            smart_proxy_id: smart_proxy.id,
+                            exporter_data: action_output[:exporter_data])
+            end
+          end
+
+          def run
+            smart_proxy = ::SmartProxy.find(input[:smart_proxy_id])
+            api = ::Katello::Pulp3::Api::Core.new(smart_proxy)
+            export_data = api.export_api.list(input[:exporter_data][:pulp_href]).results.first
+            output[:exported_file_name], output[:exported_file_checksum] = export_data.output_file_info.first
+          end
+
+          def humanized_name
+            _("Export")
+          end
+
+          def rescue_strategy
+            Dynflow::Action::Rescue::Skip
+          end
+        end
+      end
+    end
+  end
+end
--- a/app/models/setting/content.rb
+++ b/app/models/setting/content.rb
@@ -105,6 +105,8 @@ class Setting::Content < Setting
                5000, N_('Pulp Docker registry port')),
       self.set('pulp_export_destination', N_("On-disk location for exported repositories"),
                "/var/lib/pulp/katello-export", N_('Pulp export destination filepath')),
+      self.set('pulpcore_export_destination', N_("On-disk location for pulp 3 exported repositories"),
+               "/var/lib/pulp/exports", N_('Pulp 3 export destination filepath')),
       self.set('pulp_client_key', N_("Path for ssl key used for pulp server auth"),
                "/etc/pki/katello/private/pulp-client.key", N_('Pulp client key')),
       self.set('pulp_client_cert', N_("Path for ssl cert used for pulp server auth"),
--- a/app/services/katello/pulp3/content_view_version.rb
+++ b/app/services/katello/pulp3/content_view_version.rb
@@ -1,0 +1,64 @@
+module Katello
+  module Pulp3
+    class ContentViewVersion
+      def initialize(smart_proxy:, content_view_version: nil)
+        @smart_proxy = smart_proxy
+        @content_view_version = content_view_version
+      end
+
+      def exporter_name
+        @content_view_version.name.gsub(/\s/, '_')
+      end
+
+      def generate_exporter_id
+        "#{@content_view_version.organization.label}_#{exporter_name}"
+      end
+
+      def generate_exporter_path
+        export_path = "#{@content_view_version.content_view}/#{@content_view_version.version}".gsub(/\s/, '_')
+        "#{@content_view_version.organization.label}/#{export_path}"
+      end
+
+      def api
+        ::Katello::Pulp3::Api::Core.new(@smart_proxy)
+      end
+
+      def repository_hrefs
+        version_hrefs.map { |href| version_href_to_repository_href(href) }.uniq
+      end
+
+      def version_hrefs
+        if @content_view_version.default?
+          @content_view_version.repositories.yum_type.pluck(:version_href).compact
+        else
+          @content_view_version.archived_repos.yum_type.pluck(:version_href).compact
+        end
+      end
+
+      def version_href_to_repository_href(version_href)
+        version_href.split("/")[0..-3].join("/") + "/"
+      end
+
+      def create_exporter(export_base_dir: Setting['pulpcore_export_destination'])
+        api.exporter_api.create(name: generate_exporter_id,
+                                path: "#{export_base_dir}/#{generate_exporter_path}",
+                                repositories: repository_hrefs)
+      end
+
+      def create_export(exporter_href)
+        [api.export_api.create(exporter_href, { versions: version_hrefs })]
+      end
+
+      def fetch_export(exporter_href)
+        api.export_api.list(exporter_href).results.first
+      end
+
+      def destroy_exporter(exporter_href)
+        export_data = fetch_export(exporter_href)
+        api.exporter_api.partial_update(exporter_href, :last_export => nil)
+        api.export_api.delete(export_data.pulp_href) unless export_data.blank?
+        api.exporter_api.delete(exporter_href)
+      end
+    end
+  end
+end
--- a/app/lib/actions/katello/repository/export.rb
+++ b/app/lib/actions/katello/repository/export.rb
@@ -15,7 +15,7 @@ module Actions
           end
 
           unless File.writable?(Setting['pulp_export_destination'])
-            fail ::Foreman::Exception, N_("Unable to export, 'pulp_export_destination' setting is not a writable directory.")
+            fail ::Foreman::Exception, N_("Unable to export. 'pulp_export_destination' setting is not a writable directory.")
           end
 
           repo_pulp_ids = repos.collect do |repo|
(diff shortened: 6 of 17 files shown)


parthaa:
To test this API

- Make sure you have >= pulpcore-3.4.1
- Set up the export directory
```
$ mkdir /tmp/exports
$ chown -R apache:pulp /tmp/exports
```
- Include the directory in /etc/pulp/settings.py (this will be added as an installer option in a different pr.)
```
ALLOWED_EXPORT_PATHS= ["/tmp/exports"]
```
- Restart pulp services
```
sudo systemctl restart pulpcore-resource-manager pulpcore-content pulpcore-api pulpcore-worker@1 pulpcore-worker@2
```
- katello ui -  Goto Admin-> Settings
- Change the`pulp_export_destination` in content tab to `/tmp/exports`
- Create/Sync a yum repo 
- Publish a Content View associated to this (you can also use default content view if you so choose)
- now go to console to the  hammer-cli-katello git checkout and execute the following
```
$ hammer content-view version export-legacy --id=1 
[.....................................................................................................................................................................................................................................] [100%]
``` 
- Verify that a tar gz got created in /tmp/exports
```
$ ls -lh /tmp/exports/Default_Organization/Default_Organization_View_1.0
-rw-r--r--. 1 pulp pulp 282K Jul 23 00:57 export-b2c15190-6d5b-48af-9571-ad47b1ebab73-20200723_0057.tar.gz
```

Back in the UI
- Monitor -. Tasks 
- Search for `action = Export` to pick the export task
- Click on Dynflow Console
- Examine the `3: Actions::Pulp3::ContentViewVersion::Export ` in the run tab
- you should see something like below with the output file info (marked in red)
Dynflow-Console

app/lib/actions/pulp3/content_view_version/export.rb
```diff
@@ -0,0 +1,33 @@
...
+          smart_proxy = SmartProxy.find(input[:smart_proxy_id])
+          api = ::Katello::Pulp3::Api::Core.new(smart_proxy)
+          export_base_dir = input[:export_base_dir]
+          repository_hrefs = cvv.repositories.map { |repo| repo.repository_reference[:repository_href] }
+          exporter_data = api.exporter_api.create(name: "#{cvv.generate_exporter_id}-#{rand(9999)}",
+                                                  path: "#{export_base_dir}/#{cvv.generate_exporter_path}",
+                                                  repositories: repository_hrefs)
+          tasks = [api.export_api.create(exporter_data.pulp_href, {})]
+          output[:exporter_data] = exporter_data
+          output[:export_data] = api.export_api.list(exporter_data.pulp_href).results.first
```
jlsherrill, line 21: i don't know that you are guaranteed to be able to access export_data here.  Consider if all workers are busy syncing some repos or something else.   This should appear in the task as a 'created_resource', if it doesnt that is a pulp bug IMO.  Assuming it does, you can just pass the pulp_tasks to the tasks that need them and grab it out of created_resources there

parthaa: Hmm good point. I have not been able to reproduce this issue, but I can see it being possible. Let me see if I can make that issue go away :)

parthaa: I have updated the actions a bit. Resolve if you are happy with the outcome

app/lib/actions/katello/content_view_version/export.rb
```diff
@@ -2,35 +2,45 @@ module Actions
...
+          param :exported_file_checksum, String
         end
 
-        def finalize
-          history = ::Katello::ContentViewHistory.find(input[:history_id])
-          history.status = ::Katello::ContentViewHistory::SUCCESSFUL
-          history.save!
+        def plan(content_view_version)
+          unless File.directory?(Setting['pulp_export_destination'])
+            fail ::Foreman::Exception, N_("Unable to export, 'pulp_export_destination' setting is not set to a valid directory.")
```
jturel, line 17: Grammar nitpick: this message should be two sentences or joined with 'because' instead of a comma :)

jturel, line 17: Interesting - we don't do much raising Foreman::Exception. Is there a special significance?

app/lib/actions/katello/content_view_version/export.rb
```diff
@@ -2,35 +2,45 @@ module Actions
...
+
+        def run
+          smart_proxy = SmartProxy.find(input[:smart_proxy_id])
+          api = ::Katello::Pulp3::Api::Core.new(smart_proxy)
+          export_data = api.export_api.list(input[:exporter_data][:pulp_href]).results.first
+          output[:exported_file_name], output[:exported_file_checksum] = export_data.output_file_info.first
+        end
+
+        def humanized_name
+          _("Export")
```
jturel, line 43: Haven't tested yet, but if this shows up as 'Export' in the task list then we should give some more details. I've seen in other tasks we show product and repo names where appropriate. Maybe we can show the CV name and version?

app/lib/actions/pulp3/content_view_version/export.rb
```diff
@@ -0,0 +1,32 @@
...
+          param :smart_proxy_id, Integer
+          param :content_view_version_id, Integer
+          param :export_base_dir, String
+        end
+
+        def export(cvv)
+          smart_proxy = SmartProxy.find(input[:smart_proxy_id])
+          api = ::Katello::Pulp3::Api::Core.new(smart_proxy)
+          export_base_dir = input[:export_base_dir]
+          repository_hrefs = cvv.repositories.map { |repo| repo.repository_reference[:repository_href] }
```
jturel, line 15: Do you think it'd be advantageous to pass the repository hrefs as the input to this action? That would help decouple this pulp3 focused code from Katello. We don't use anything else off the CVV

app/models/katello/content_view_version.rb
```diff
@@ -184,6 +185,18 @@ def products(env = nil)
       end
     end
 
+    def exporter_name
```
jturel, line 188: From what I can tell these 3 methods seem useful only in the pulp3 context. If we have a place, I think they could go into one of the pulp3 service classes, or perhaps stay in the action where used

app/lib/actions/pulp3/content_view_version/export.rb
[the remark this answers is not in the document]
parthaa: Yes. So to make an export happen I have to

1. create an exporter
1. create an export referencing this exporter

I wanted the href of the created export and had this originally but as @jlsherrill pointed out the create task may have not completed before the export href is generated. So it got moved outside this task

app/models/katello/content_view_version.rb
parthaa: Yeh but they belong to the version. I can try creating a "app/services/katello/pulp3/content_view_version.rb" and have it included.

parthaa: I thought about this and feel the pulp3 separation as unnecessary given this PR  is going to only work with pulp3 anyway

app/lib/actions/pulp3/content_view_version/export.rb
parthaa: I thought about this but I saw code passing in Katello models in pulp3 so decided to be consistent. Fine with passing a list of hrefs if you prefer :)

app/lib/actions/katello/content_view_version/export.rb
parthaa: Shouldnt that information be in the input params anyway ?

app/models/katello/content_view_version.rb
parthaa: AFAIK we dont have the logic to say "include this module only if pulp3" in any of the models . @jlsherrill correct me if wrong.

app/lib/actions/katello/content_view_version/export.rb
parthaa: That was previous code :) . I don;t see why it needs to be special. (wonder if hammer behaves differently when it sees a foreman exception ?)

app/models/katello/content_view_version.rb
jlsherrill: we don't, but a service class wouldn't be a module that you include anyways. It would be its own object potentially, something like:
```
exporter = Katello::Pulp3::Exporter.new(cvv)
exporter.generate_exporter_id
expoter.exporter_id
```

app/lib/actions/pulp3/content_view_version/export.rb
```diff
@@ -0,0 +1,32 @@
...
+    module ContentViewVersion
+      class Export < Pulp3::AbstractAsyncTask
+        input_format do
+          param :smart_proxy_id, Integer
+          param :content_view_version_id, Integer
+          param :export_base_dir, String
+        end
+
+        def export(cvv)
+          smart_proxy = SmartProxy.find(input[:smart_proxy_id])
```
jlsherrill, line 12: you can just call 'smart_proxy' here, Pulp3::AbstractTask defines the 'smart_proxy' method

app/lib/actions/pulp3/content_view_version/export.rb
```diff
@@ -0,0 +1,32 @@
...
+          api = ::Katello::Pulp3::Api::Core.new(smart_proxy)
+          export_base_dir = input[:export_base_dir]
```
jlsherrill, line 14: curious, does this need to be overridden in the action?  It looks like it comes from a setting, could that just be looked up?

app/lib/actions/pulp3/content_view_version/export.rb
jlsherrill: we do typically try to pass models instead of hrefs, as that even more so ties it to pulp3.   But in general it does seem like these actions have more pulp3 logic in them than most others.  In fact most don't call api methods directly, they let the service class do that for them.  Sometimes that is warranted so its not necessary a terrible thing.

I do kinda wonder if the action planning looked more like this:
```
sequence do 
     exporter_output = plan_action(Actions::pulp3::ContentViewVersion::CreateExporter).output
     plan_action(Actions::pulp3::ContentViewVersion::Export, exporter_output[:exporter_href])
     plan_action(::Actions::Pulp3::ContentViewVersion::DestroyExporter, exporter_output[:exporter_href)
end
```

That would allow the actions to be smaller and i think let you push the api calls and logic into an Exporter service class.  We've tried to stick with dumb actions as much as possible and two things help with that:  1) having an action do one and only one things, 2) using the orchestration features of dynflow to string them together.

app/lib/actions/pulp3/content_view_version/export.rb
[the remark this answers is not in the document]
parthaa: Updated!. Check now

app/lib/actions/katello/content_view_version/export.rb
parthaa: updated

app/lib/actions/pulp3/content_view_version/export.rb
parthaa: updated

app/models/katello/content_view_version.rb
parthaa: updated. Created a service class -> `app/services/katello/pulp3/content_view_version.rb` where all this logic is handled.

app/lib/actions/pulp3/content_view_version/export.rb
parthaa: updated.

app/lib/actions/pulp3/content_view_version/export.rb
parthaa: updated. Moved it to the service class

app/services/katello/pulp3/content_view_version.rb
```diff
@@ -0,0 +1,50 @@
+module Katello
+  module Pulp3
+    class ContentViewVersion
```
jturel, line 3: Having all this code in one place makes a lot of sense to me. Thanks for the refactor!

app/controllers/katello/api/v2/content_view_versions_controller.rb
```diff
@@ -84,31 +84,35 @@ def republish_repositories
 
     api :POST, "/content_view_versions/:id/export", N_("Export a content view version"), :deprecated => true
     param :id, :number, :desc => N_("Content view version identifier"), :required => true
-    param :export_to_iso, :bool, :desc => N_("Export to ISO format"), :required => false
-    param :iso_mb_size, :number, :desc => N_("maximum size of each ISO in MB"), :required => false
-    param :since, Date, :desc => N_("Optional date of last export (ex: 2010-01-01T12:00:00Z)"), :required => false
+    param :export_to_iso, :bool, :desc => N_("Export to ISO format, relevant only on Pulp 2 repositories"), :required => false
```
jturel, line 87: In general, if our users are knowing about the difference between pulp3 and pulp2 we've failed. I see what you're trying to do here, though. I would be fine with removing this or logging a warning if it's used against the wrong implementation. If you wanna keep it, it should be two sentences :)

app/controllers/katello/api/v2/content_view_versions_controller.rb
```diff
@@ -84,31 +84,35 @@ def republish_repositories
...
+    param :iso_mb_size, :number, :desc => N_("maximum size of each ISO in MB, relevant only on Pulp 2 repositories"), :required => false
+    param :since, Date, :desc => N_("Optional date of last export (ex: 2010-01-01T12:00:00Z), relevant only on Pulp 2 repositories"), :required => false
     def export
-      ::Foreman::Deprecation.api_deprecation_warning("Export is being deprecated and will be removed in a future version of Katello. Use hammer content-view version export instead.")
-      if params[:export_to_iso].blank? && params[:iso_mb_size].present?
-        fail HttpErrors::BadRequest, _("ISO export must be enabled when specifying ISO size")
-      end
+      if  SmartProxy.pulp_master.pulp3_repository_type_support?(Katello::Repository::YUM_TYPE)
```
jturel, line 91: Nitpick - extra space after the `if`

app/lib/actions/pulp3/orchestration/content_view_version/export.rb
```diff
@@ -0,0 +1,59 @@
...
+
+          def run
+            smart_proxy = ::SmartProxy.find(input[:smart_proxy_id])
+            api = ::Katello::Pulp3::Api::Core.new(smart_proxy)
+            export_data = api.export_api.list(input[:exporter_data][:pulp_href]).results.first
+            output[:exported_file_name], output[:exported_file_checksum] = export_data.output_file_info.first
+          end
+
+          def humanized_name
+            _("Export")
```
jturel, line 49: I'd really like to be more descriptive here. 'Export' says nothing about what's going on; seeing the word Export five times in the task list requires more digging if you're looking for a specific one; what was exported?; etc

Contrast to CV publish: `Publish content view 'Pulp3View'; organization 'Default Organization'` this information is way more useful, especially to someone who just wants to look at the tasks and get a general sense of what's going on, let alone the user trying to find their specific Export task

app/controllers/katello/api/v2/content_view_versions_controller.rb
```diff
@@ -84,31 +84,35 @@ def republish_repositories
 
     api :POST, "/content_view_versions/:id/export", N_("Export a content view version"), :deprecated => true
     param :id, :number, :desc => N_("Content view version identifier"), :required => true
-    param :export_to_iso, :bool, :desc => N_("Export to ISO format"), :required => false
-    param :iso_mb_size, :number, :desc => N_("maximum size of each ISO in MB"), :required => false
-    param :since, Date, :desc => N_("Optional date of last export (ex: 2010-01-01T12:00:00Z)"), :required => false
+    param :export_to_iso, :bool, :desc => N_("Export to ISO format, relevant only on Pulp 2 repositories"), :required => false
+    param :iso_mb_size, :number, :desc => N_("maximum size of each ISO in MB, relevant only on Pulp 2 repositories"), :required => false
+    param :since, Date, :desc => N_("Optional date of last export (ex: 2010-01-01T12:00:00Z), relevant only on Pulp 2 repositories"), :required => false
     def export
```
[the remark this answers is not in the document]
parthaa: Good point. I don't think we should  support setting destination directory. Pulp is limited in where it exports things permissions and selinux and all that Jazz. We don't have an easy way to add new destinations without restarting pulp.

app/services/katello/pulp3/content_view_version.rb
parthaa: Great suggestion!

app/controllers/katello/api/v2/content_view_versions_controller.rb
parthaa: Alternately I can raise errors on unsupported parameters if you prefer ? So if your yum is pulp3 and you passed in pulp2 params I can complain. Would you like that?

jturel: That sounds good to me!

app/lib/actions/pulp3/orchestration/content_view_version/export.rb
parthaa: Should be resolved now :) 
```
Export content view version 'hah 1.0'; content view 'hah'; organization 'Default Organization'
```

app/models/katello/repository.rb
```diff
@@ -237,6 +237,10 @@ def library_instance?
       self.content_view.default?
     end
 
+    def repository_reference
```
jturel, line 240: Unused now?

app/lib/actions/pulp3/orchestration/content_view_version/export.rb
jturel: I don't see any change here

app/controllers/katello/api/v2/content_view_versions_controller.rb
jturel: Do you plan to update per above?

app/lib/actions/pulp3/orchestration/content_view_version/export.rb
parthaa: Oh the change was -> https://github.com/Katello/katello/pull/8850/files#diff-476181c4440753c9b39d487820d7679bR17 to make this work? Are you not seeing this work when you run from the PR?

app/models/katello/repository.rb
parthaa: removed :+1:

app/controllers/katello/api/v2/content_view_versions_controller.rb
parthaa: updated

app/services/katello/pulp3/content_view_version.rb
```diff
@@ -0,0 +1,65 @@
...
+        else
+          @content_view_version.archived_repos.pluck(:version_href)
+        end
+      end
+
+      def version_href_to_repository_href(version_href)
+        version_href.split("/")[0..-3].join("/") + "/"
+      end
+
+      def create_exporter(export_base_dir: nil)
```
jturel, line 42: You could just set the keyword arg to Setting[:pulp_export_destination] by default!

jturel:
I gave this a re-test after the code changes and upgrading to pulpcore 3.5.0. Now my export looks as expected. I have a  pulpcore.app.modelresource.ArtifactResource.json which contains references to the files in the artifacts/ directory within the export:

```
  {
    "pulp_id": "cae6aa8d-ce7e-44b5-8b98-9f798b6bfc0d",
    "pulp_created": "2020-07-02 19:27:55",
    "pulp_last_updated": "2020-07-02 19:27:55",
    "file": "artifact/2c/5df6b51df167eb01246dce3049668cbe688033b9abff24464b0e05cdeb20ac",
    "size": 2446,
    "md5": "5fa2b27cc8f5e738a76d3a3c8b73a44a",
    "sha1": "1a690f53f33d0c9378b723cacab1954d80f8ac7f",
    "sha224": "f71765e25f5fb7e7c3678b44e25ebcfd0834b6bb6eca101f327d9297",
    "sha256": "2c5df6b51df167eb01246dce3049668cbe688033b9abff24464b0e05cdeb20ac",
    "sha384": "5fd131930e16bb91d19526884d0ba8333d459e1af3bed8273bc1495c6ab249451d9446e03f220b60c2f53dc789eec66e",
    "sha512": "435b036646df53b66bd8c54dd7ded0ad1a6b8c2992ca7521da2aecbe16f383ba35a83e733688ea0f5fd35bdbd4060e923ce11c5f58b949b5cedda8aa7299372e"
  }
```

app/models/setting/content.rb
```diff
@@ -104,7 +104,7 @@ def self.default_settings
       self.set('pulp_docker_registry_port', N_("The port used by Pulp Crane to provide Docker Registries"),
                5000, N_('Pulp Docker registry port')),
       self.set('pulp_export_destination', N_("On-disk location for exported repositories"),
-               "/var/lib/pulp/katello-export", N_('Pulp export destination filepath')),
+               "/var/lib/pulp/exports", N_('Pulp export destination filepath')),
```
jlsherrill, line 107: are there selinux implications to changing this for pulp2 installs?

jlsherrill: or permission/installer/packaging implications?

jlsherrill:
need an installer pr to add this to the settings.py:  ALLOWED_EXPORT_PATHS= ["/var/lib/pulp/exports"]

Also need to determine if there are installer/selinux needs to make exporting to that path work, but all of that can be done async to this PR (before or after)

parthaa:
> need an installer pr to add this to the settings.py: ALLOWED_EXPORT_PATHS= ["/var/lib/pulp/exports"]
> 
> Also need to determine if there are installer/selinux needs to make exporting to that path work, but all of that can be done async to this PR (before or after)

Have created a Kanboard task for the next sprint.

app/models/setting/content.rb
parthaa: hmm are you saying we should leave this alone for this PR and have the installer determine the settings (based on pulp3/2) ?

parthaa: We discussed further and decided to have a new setting for `pulpcore_export_directory` because the permissions are quite  different between `/var/lib/pulp/katello-exports` and `/var/lib/pulp/exports`

jlsherrill: +1 to this!

parthaa: updated.

app/models/setting/content.rb
```diff
@@ -104,7 +104,9 @@ def self.default_settings
       self.set('pulp_docker_registry_port', N_("The port used by Pulp Crane to provide Docker Registries"),
                5000, N_('Pulp Docker registry port')),
       self.set('pulp_export_destination', N_("On-disk location for exported repositories"),
-               "/var/lib/pulp/katello-export", N_('Pulp export destination filepath')),
+               "/var/lib/pulp/katello-exports", N_('Pulp export destination filepath')),
```
jlsherrill, line 107: didn't you mean to drop the 's'?

jlsherrill: errrrr "Did you mean to drop the 's'?"

PR #8851, opened 2020-07-24
Fixes #30189 - support force full sync with pulp3

jlsherrill:


jlsherrill:
[test katello]

app/services/katello/pulp3/repository.rb
```diff
@@ -193,10 +193,12 @@ def delete(href = repository_reference.try(:repository_href))
         ignore_404_exception { api.repositories_api.delete(href) } if href
       end
 
-      def sync
+      def sync(options = {})
         sync_url_params = {remote: repo.remote_href, mirror: repo.root.mirror_on_sync}
-        skip_type_param = skip_types
-        sync_url_params[:skip_types] = skip_type_param if skip_type_param
+
+        sync_url_params[:optimize] = options[:optimize] if options.key?(:optimize) == false
```
jlsherrill, line 199: even thought optimize should always be false for non-yum repos, i think it makes sense  to move this to yum.rb

app/lib/actions/pulp3/repository/save_version.rb
```diff
@@ -3,35 +3,39 @@ module Pulp3
     module Repository
       class SaveVersion < Pulp3::Abstract
         def plan(repository, options)
-          plan_self(:repository_id => repository.id, :tasks => options[:tasks], :repository_details => options[:repository_details])
+          fail "Cannot accept tasks and repository_details into Save Version." if options[:tasks].present? && options[:repository_details].present?
+          plan_self(:repository_id => repository.id, :tasks => options[:tasks], :repository_details => options[:repository_details], :force_fetch_version => options.fetch(:force_fetch_version, false))
         end
```
sjha4, line 8: How are we using this? Is there a way for users to set force_fetch_version to true?

jlsherrill: updated to pass this in through the pulp3 repo sync orchestration action

also added a test that tests this functionality.  You can test manually too:

```
> Katello::Repository.first.version_href
=> "/pulp/api/v3/repositories/rpm/rpm/0d068318-16a4-4f1f-bc82-6a4fc296ad5e/versions/3/"
```
Find a repo and reset its version_href to zero (note you may want to use .find(id) here if you have more than 1 repo.
```
> Katello::Repository.first.update_attributes!(version_href:  "/pulp/api/v3/repositories/rpm/rpm/0d068318-16a4-4f1f-bc82-6a4fc296ad5e/versions/0/")
```

Re-index the repo at version zero:

```
> Katello::Repository.first.index_content
> Katello::Repository.first.rpms
0
```

Now do an advanced sync, skipping the metadata check.  You should see the version href go back to the newest version, and the proper content get indexed from that latest version. 

This workflow simulates what happens if a repo gets updated outside of katello (or katello fails to index it properly), and the user performs this action to correct it

jlsherrill:
rebased against master

PR #8852, opened 2020-07-24
Fixes #30333: Pulp 3 yum cross-repo dependency solving for content views

ianballou:
Fixes included in this PR:

1) Cross-repo dep solving now supported with Pulp 3 yum content views
2) Composite content view publishing fixed for Pulp 3 yum repositories
3) Incremental update refactoring and fixes for:
  a) Pulp 3 incremental update uses a non-deletable repo version 0 if there are no content units to copy
  b) Pulp 3 incremental update doesn't handle repo version saving if the content units already exist in the repo

This PR include new "multi" versions of the following actions:
`CloneContents`, `CloneToVersion`, `CopyAllUnits`, and `CopyContent`

These actions use a repository mapping to manipulate multiple repositories at once, which is required to use the Pulp 3 yum copy API properly.

At the moment, these actions work only for Pulp 3 yum repositories.  To ensure that this PR doesn't get too big, I [separate out the Pulp 3 yum content](https://github.com/Katello/katello/compare/master...ianballou:30333-cross-repo-dep-solving?expand=1#diff-620d80d1e16354da4a45be23cd269036R44) if dependency solving is `true` for the content view.

Things to test (**for Pulp 2 and Pulp 3**):
1) Content view publishing:
  a) With 1 repo, with and without filters with and without dep solving on
  b) With > 1 repos, with and without filters with and without dep solving on
2) Incremental updating:
  a) With multiple RPMs and errata, along with 1a and 1b above
  b) With RPMs and errata that don't exist in any of the content view library instance repos
  c) With RPMs and errata that already exist in the content view versions
3) Composite content view workflows:
  a) Publishing using 2+ repos with combinations from 1a and 1b
  b) Incremental updating using combinations from 2a, 2b, and 2c
      -> Instead of incrementally updating the composite view itself, incrementally update the composite component content views and use the `--propagate-all-composites true` option

I need to write some tests for this.  I'm putting it out a bit early so people can start testing things out.

NOTES:

I've seen Pulp 3 fail to perform dependency solving for some RPMs (like zsh) in the RHEL8 BaseOS repo.  Keep an eye out for issues like that.  Pulp 3 issue: https://pulp.plan.io/issues/7202

### DIFF
--- a/app/lib/actions/katello/content_view/publish.rb
+++ b/app/lib/actions/katello/content_view/publish.rb
@@ -4,6 +4,7 @@ module Actions
     module ContentView
       class Publish < Actions::EntryAction
         include ::Katello::ContentViewHelper
+        attr_accessor :version
         # rubocop:disable Metrics/MethodLength,Metrics/AbcSize
         def plan(content_view, description = "", options = {})
           action_subject(content_view)
@@ -22,7 +23,11 @@ module Actions
             end
           end
 
+          # Add non-override components back in
+          options[:override_components] = include_other_components(options[:override_components], content_view)
+
           version = version_for_publish(content_view, options)
+          self.version = version
           library = content_view.organization.library
           history = ::Katello::ContentViewHistory.create!(:content_view_version => version,
                                                           :user => ::User.current.login,
@@ -38,7 +43,7 @@ module Actions
           end
 
           sequence do
-            plan_action(ContentView::AddToEnvironment, version, library)
+            plan_action(ContentView::AddToEnvironment, version, library) unless options[:skip_promotion]
             repository_mapping = plan_action(ContentViewVersion::CreateRepos, version, source_repositories).repository_mapping
 
             # Split Pulp 3 Yum repos out of the repository_mapping.  Only Pulp 3 RPM plugin has multi repo copy support.
@@ -56,7 +61,7 @@ module Actions
                     plan_action(Repository::CloneToVersion, repositories, version, repository_mapping[repositories],
                                 :repos_units => options[:repos_units])
                   end
-                  plan_action(Repository::CloneToEnvironment, repository_mapping[repositories], library)
+                  plan_action(Repository::CloneToEnvironment, repository_mapping[repositories], library) unless options[:skip_promotion]
                 end
               end
 
@@ -66,16 +71,18 @@ module Actions
             end
             has_modules = content_view.publish_puppet_environment?
             plan_action(ContentViewPuppetEnvironment::CreateForVersion, version)
-            plan_action(ContentViewPuppetEnvironment::Clone, version, :environment => library,
-                :puppet_modules_present => has_modules)
-            plan_action(Candlepin::Environment::SetContent, content_view, library, content_view.content_view_environment(library))
-            plan_action(Katello::Foreman::ContentUpdate, library, content_view)
-            plan_action(ContentView::ErrataMail, content_view, library)
+            unless options[:skip_promotion]
+              plan_action(ContentViewPuppetEnvironment::Clone, version, :environment => library,
+                  :puppet_modules_present => has_modules)
+            end
+            plan_action(Candlepin::Environment::SetContent, content_view, library, content_view.content_view_environment(library)) unless options[:skip_promotion]
+            plan_action(Katello::Foreman::ContentUpdate, library, content_view) unless options[:skip_promotion]
+            plan_action(ContentView::ErrataMail, content_view, library) unless options[:skip_promotion]
             plan_self(history_id: history.id, content_view_id: content_view.id,
                       auto_publish_composite_ids: auto_publish_composite_ids(content_view),
                       content_view_version_name: version.name,
                       content_view_version_id: version.id,
-                      environment_id: library.id, user_id: ::User.current.id)
+                      environment_id: library.id, user_id: ::User.current.id, skip_promotion: options[:skip_promotion])
           end
         end
 
@@ -96,6 +103,7 @@ module Actions
 
           output[:content_view_id] = input[:content_view_id]
           output[:content_view_version_id] = input[:content_view_version_id]
+          output[:skip_promotion] = input[:skip_promotion]
         end
 
         def rescue_strategy_for_self
@@ -106,9 +114,11 @@ module Actions
           version = ::Katello::ContentViewVersion.find(input[:content_view_version_id])
           version.update_content_counts!
           # update errata applicability counts for all hosts in the CV & Library
-          ::Katello::Host::ContentFacet.where(:content_view_id => input[:content_view_id],
-                                              :lifecycle_environment_id => input[:environment_id]).each do |facet|
-            facet.update_applicability_counts
+          unless input[:skip_promotion]
+            ::Katello::Host::ContentFacet.where(:content_view_id => input[:content_view_id],
+                                                :lifecycle_environment_id => input[:environment_id]).each do |facet|
+              facet.update_applicability_counts
+            end
           end
 
           history = ::Katello::ContentViewHistory.find(input[:history_id])
@@ -116,7 +126,7 @@ module Actions
           history.save!
           environment = ::Katello::KTEnvironment.find(input[:environment_id])
           view = ::Katello::ContentView.find(input[:content_view_id])
-          if SmartProxy.sync_needed?(environment) && Setting[:foreman_proxy_content_auto_sync]
+          if SmartProxy.sync_needed?(environment) && Setting[:foreman_proxy_content_auto_sync] && !input[:skip_promotion]
             ForemanTasks.async_task(ContentView::CapsuleSync,
                                     view,
                                     environment)
@@ -125,6 +135,20 @@ module Actions
         end
 
         private
+
+        def include_other_components(override_components, content_view)
+          if override_components.present?
+            content_view.components.each do |component|
+              component_has_override = override_components.detect do |override_component|
+                component.content_view_id == override_component.content_view_id
+              end
+              unless component_has_override
+                override_components << component
+              end
+            end
+            override_components
+          end
+        end
 
         def repos_to_delete(content_view)
           if content_view.composite?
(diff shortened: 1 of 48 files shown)


app/lib/actions/katello/content_view/publish.rb
```diff
@@ -40,11 +40,22 @@ def plan(content_view, description = "", options = {})
             plan_action(ContentView::AddToEnvironment, version, library)
             repository_mapping = plan_action(ContentViewVersion::CreateRepos, version, source_repositories).repository_mapping
 
+            # Split Pulp 3 Yum repos out of the repository_mapping.  Only Pulp 3 RPM plugin has multi repo copy support.
+            separated_repo_map = separated_repo_mapping(repository_mapping, content_view)
+
+            if separated_repo_map[:pulp3_yum_depsolve].keys.flatten.present? &&
+                SmartProxy.pulp_master.pulp3_support?(separated_repo_map[:pulp3_yum_depsolve].keys.flatten.first) &&
+                content_view.solve_dependencies
+              plan_action(Repository::MultiCloneToVersion, separated_repo_map[:pulp3_yum_depsolve], version)
```
jlsherrill, line 49: thinking ahead to when pulp2 is gone, would it simplify things to always use MultiCloneToVersion even if solve_dependencies is false?  I assume the multi-repo copy api supports turning off dep solving?

jlsherrill: thinking logically, it then becomes:

if pulp3 supports yum, use MultiCloneToVersion for yum repos, otherwise use CloneToVersion

app/lib/actions/katello/repository/multi_clone_contents.rb
```diff
@@ -0,0 +1,67 @@
...
+
+        def metadata_generate(source_repositories, new_repository, filters)
+          metadata_options = {}
+
+          if source_repositories.count == 1 && filters.empty?
+            metadata_options[:source_repository] = source_repositories.first
+          end
+
+          check_matching_content = ::Katello::RepositoryTypeManager.find(new_repository.content_type).metadata_publish_matching_check
+          if new_repository.environment && source_repositories.count == 1 && check_matching_content && !SmartProxy.pulp_master.pulp3_support?(new_repository)
```
jlsherrill, line 40: is it possible for it not to support pulp3 here?

app/lib/actions/pulp3/repository/multi_copy_content.rb
```diff
@@ -0,0 +1,31 @@
...
+          plan_self(options.merge(:repo_id_map => repo_id_map,
+                                  :smart_proxy_id => smart_proxy.id,
+                                  :filter_ids => filter_ids.flatten))
+        end
+
+        def invoke_external_task
+          repo_id_map = {}
+
+          input[:repo_id_map].each do |source_repo_ids, dest_repo_id|
+            repo_id_map[JSON.parse(source_repo_ids)] = dest_repo_id
```
jlsherrill, line 23: any idea why the JSON.parse here was needed?

app/services/katello/pulp3/repository/yum.rb
```diff
@@ -70,7 +70,35 @@ def partial_repo_path
...
+
+          content_unit_hrefs.sort!
+          if content_unit_hrefs.any?
+            data = PulpRpmClient::Copy.new
+            data.dependency_solving = dependency_solving
+            data.config = []
+            repo_id_map.each do |source_repo_ids, dest_repo_id_map|
+              dest_repo = ::Katello::Repository.find(dest_repo_id_map[:dest_repo])
+              dest_repo_href = ::Katello::Pulp3::Repository::Yum.new(dest_repo, SmartProxy.pulp_master).repository_reference.repository_href
+              tasks << remove_all_content_from_repo(dest_repo_href)
```
jlsherrill, line 84: how would this work with incremental update?  Wouldn't this always clear out the content ?

jlsherrill: (and i'm very probably misunderstanding the flow here)

app/lib/actions/pulp3/repository/multi_copy_content.rb
ianballou: My repo ID map ended up coming back as a string for some reason so I had to do a JSON.parse.

app/lib/actions/katello/content_view/publish.rb
ianballou: Yeah that definitely makes sense to me, when writing this I knew that we'd want to use MultiCloneToVersion without dep solving too at some point.  I didn't include yet because I was worried about taking too much time getting this PR into 3.16, but I can fix it if we do have time.  The potentially time-consuming bit would just be getting the composite content view functionality right.

app/services/katello/pulp3/repository/yum.rb
ianballou: The idea with incremental update is that the dest_base_version determines what content is should already be in the repo when copying.  `remove_all_content_from_repo` then would generate a new repo version which would not affect the incremental update.  Without the `remove_all_content_from_repo`, my exclude filtering does not work during content view publishing with dep solving.

Perhaps I could skip this content removal during incremental update by checking if `dest_repo_id_map[:base_version]` exists.

app/lib/actions/katello/repository/multi_clone_contents.rb
ianballou: That line doesn't make sense, I will revisit it.

app/services/katello/pulp3/repository/yum.rb
ianballou:
```
unless dest_repo_id_map[:base_version]
  tasks << remove_all_content_from_repo(dest_repo_href)
end
```

app/lib/actions/katello/content_view/publish.rb
ianballou: I started implementing the changes to use MultiCloneToVersion for all Pulp 3 Yum content, and it's been working well so far.  I need to do a little testing, but I've been able to get composite content views to "merge" pretty well.

ianballou: I pushed a new commit that has my work so far.

ianballou: Looks like the only part of this that isn't working right is incremental update and `--propagate-all-composites true`.  It does work if I simply publish the composite, though.

app/lib/actions/pulp3/repository/multi_copy_content.rb
jlsherrill: i see that too.  It looks like json doesn't support an array as a hash key, so it converts it to a string

ianballou:
TODO: 
- [ ] Fix incremental update with composites
- [ ] Make per-repo filtering work
- [ ] Write tests

jlsherrill:
Things i tested:

1) publishing a content view w/o dep solving  :heavy_check_mark:  
2) publishing a content view w/ cross repo dep solving :heavy_check_mark: 
3) incremental update w/ cross repo dep solving :heavy_check_mark: 
4) incremental update w/ cross repo dep solving w/ propagate-all-composites=true  :x: 

Only issue i found was that 4) didn't properly update the composite with the new component version, because the composite was not set to use 'latest'.

ianballou:
@jlsherrill I've fixed the issue where composites weren't publishing properly when they weren't set to latest.  There is an issue now however where the new composite version information isn't output properly, only the info for the components is output. It'll be easier to fix it once Pulp 3 is dropped, but if we need to fix it before that perhaps we should make another issue so that this isn't delayed too much longer.

I think your test cases pretty much cover the big things.  There are some corner cases like incremental updating with no content units or content units that don't exist among the CV repos, but I've been checking those regularly.

After we think this PR is settled I can test Pulp 2 once more just to make sure that no regressions are being introduced.

jlsherrill:
```
 There is an issue now however where the new composite version information isn't output properly, only the info for the components is output
```
can you clarify what you mean by that?

ianballou:
@jlsherrill so when you do an incremental update, the API/Hammer output shows the new CVV created and the content added.  If you used `--propagate-all-composites`, and composite CVs were incrementally updated, you'd also see their new versions in the API/Hammer output.

I realized now that just publishing instead of incrementally updating the composites caused the version + added content output to not be included for them.  That info came from the incremental update action, which is no longer being run (`plan_self` is skipped).

It's not a huge issue, but just something else that will have to be fixed sooner or later.

jlsherrill:
okay, i think thats acceptable.  Will do some more testing!

ianballou:
@jlsherrill, great.  Also, now that #8869 is merged, I'm adding in the package env stuff to my `multi_copy_units`.

I'm adding a big TODO to fix the tests.  It seems like proper dep-solving is causing a bunch to fail.  Plus, there are plenty of VCRs to re-record.

ianballou:
Since some of this will be difficult to test via unit tests, I will write a bunch of BATS tests after this merges.

app/lib/actions/katello/content_view/publish.rb
```diff
@@ -32,19 +33,29 @@ def plan(content_view, description = "", options = {})
                                                           :triggered_by => options[:triggered_by]
                                                          )
           source_repositories = []
-          content_view.publish_repositories do |repositories|
+          content_view.publish_repositories(options[:override_cvvs]) do |repositories|
```
jlsherrill, line 36: can we rename 'override_cvvs' to 'override_component_versions'  I think its a bit more clear

app/lib/actions/katello/content_view_version/incremental_update.rb
```diff
@@ -30,17 +31,26 @@ def plan(old_version, environments, options = {})
           validate_environments(environments, old_version)
 
           new_minor = old_version.content_view.versions.where(:major => old_version.major).maximum(:minor) + 1
+          if SmartProxy.pulp_master.pulp3_support?(old_version.repositories.first) && is_composite
```
jlsherrill, line 34: i saw something similar above, checking that the first repo in the cvv is supported under pulp3.  Whats the reason for that?  What would happen if i published a mixed set of content (file under pulp2, rpm under pulp3)?

jlsherrill:
Another issue i found, i *think* that the way propagate all composites, is if i say to incrementally update content view version id 4, it:

1.  finds all the composite content view versions that include id 4 that are included in a Lifecycle environment 
2.  creates a new incremental update for each of those composites
3.  Promotes each to the corresponding lifecycle env that the composite version was in

in step 3, for me, i'd expect it to be promoted to Dev, but instead it gets promoted to Library.  (Probably due to switch to using the publish action, maybe the promote-to environments could be passed into publish?)

PR #8856, opened 2020-07-28
Fixes #30509 - test fixes for new bindings

jlsherrill:


app/services/katello/pulp3/migration.rb
```diff
@@ -35,11 +35,15 @@ def migration_plan_api
...
+        Pulp2to3MigrationClient::Pulp2ContentApi.new(api_client)
+      rescue NameError
+        Pulp2to3MigrationClient::Pulp2contentApi.new(api_client) #backwards compatible
       end
 
       def pulp2_repositories_api
-        Pulp2to3MigrationClient::Pulp2repositoriesApi.new(api_client)
+        Pulp2to3MigrationClient::Pulp2RepositoriesApi.new(api_client)
+      rescue NameError
+        Pulp2to3MigrationClient::Pulp2repositoriesApi.new(api_client) #backwards compatible
```
jlsherrill, line 46: i left the old constants here so we can cherry pick this and have more flexibility with packaging.  It will just work everywhere

PR #8857, opened 2020-07-29
Fixes #30520 - Fix tests for HTTP proxy setting

xprazak2:
Fix for tests when https://github.com/theforeman/foreman/pull/7775 is merged


tbrisker:
Core has been merged, [test katello] should be green (other than the known chromium issue)

tbrisker (approved):
Tests have passed with this change, the failures are due to the chromium issue.

PR #8859, opened 2020-07-29
Fixes #30522 - Hide syspurpose when org is in SCA mode

chris1984:
To test,

* Snapshot your Katello environment and run

* `bundle exec rake katello:reset`

* Register 2 or more content hosts to the organization so the syspurpose status populates

* Apply patch

* Upload a manifest that has SCA enabled and check if the hosts show anything for simple content access in the all hosts and hosts - content hosts tab.

* Try to refresh a manifest as well

### DIFF
--- a/app/lib/actions/katello/organization/manifest_refresh.rb
+++ b/app/lib/actions/katello/organization/manifest_refresh.rb
@@ -61,6 +61,7 @@ module Actions
 
         def finalize
           org = ::Organization.find(input[:organization_id])
+          subject_organization.clear_syspurpose_status if subject_organization.simple_content_access?(cached: false)
           org.clear_manifest_expired_notifications
           subject_organization.audit_manifest_action(_('Manifest refreshed'))
         end
--- a/app/models/katello/concerns/organization_extensions.rb
+++ b/app/models/katello/concerns/organization_extensions.rb
@@ -228,6 +228,11 @@ module Katello
           subscriptions.select(&:expiring_soon?)
         end
 
+        def clear_syspurpose_status
+          host_purpose = HostStatus::Status.where(type: ::Katello::HostStatusManager::PURPOSE_STATUS.map(&:to_s)).where('host_id in (?)', self.hosts.pluck(:id))
+          host_purpose.destroy_all
+        end
+
         def notification_recipients_ids
           users = User.unscoped.all.find_all do |user|
             user.can?(:import_manifest) && user.can?(:view_organizations, self)
--- a/app/models/katello/glue/candlepin/owner.rb
+++ b/app/models/katello/glue/candlepin/owner.rb
@@ -35,8 +35,10 @@ module Katello
         self.owner_details['contentAccessMode']
       end
 
-      def simple_content_access?
-        content_access_mode == "org_environment"
+      def simple_content_access?(cached: true)
+        Rails.cache.fetch("#{self.label}_simple_content_access?", expires_in: 1.minute, force: !cached) do
+          content_access_mode == "org_environment"
+        end
       end
 
       def generate_debug_cert
--- a/app/models/katello/purpose_status.rb
+++ b/app/models/katello/purpose_status.rb
@@ -1,8 +1,8 @@
 module Katello
   class PurposeStatus < HostStatus::Status
-    MATCHED = 0
+    UNKNOWN = 0
     MISMATCHED = 1
-    UNKNOWN = 2
+    MATCHED = 2
     NOT_SPECIFIED = 3
 
     def self.status_map
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/details/views/content-host-info.html
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/details/views/content-host-info.html
@@ -118,8 +118,8 @@
       <div style="display: inline-block"><i class="pficon-info" title="{{ 'System Purpose allows you to set the system\'s intended use on your network and improves the accuracy of auto attaching subscriptions.' | translate }}"></i></div>
 
       <dl class="dl-horizontal dl-horizontal-left">
-        <dt translate>System Purpose Status</dt>
-        <dd>
+        <dt translate ng-hide="simpleContentAccessEnabled">System Purpose Status</dt>
+        <dd ng-hide="simpleContentAccessEnabled">
           <i ng-class="getHostPurposeStatusIcon(host.purpose_status)"></i>
           {{ host.purpose_status_label }}
         </dd>
--- a/test/actions/katello/organization_test.rb
+++ b/test/actions/katello/organization_test.rb
@@ -144,6 +144,7 @@ module ::Actions::Katello::Organization
       rhel7 = katello_repositories(:rhel_7_x86_64)
       acme_org.stubs(:owner_details).returns({})
       acme_org.products.stubs(:redhat).returns([rhel7.product])
+      Organization.any_instance.stubs(:simple_content_access?).returns(true)
       action.stubs(:rand).returns('1234')
 
       stub_action_locking!(action)
@@ -239,7 +240,7 @@ module ::Actions::Katello::Organization
       acme_org = get_organization(:empty_organization)
       rhel7 = katello_repositories(:rhel_7_x86_64)
       acme_org.products.stubs(:redhat).returns([rhel7.product])
-
+      Organization.any_instance.stubs(:simple_content_access?).returns(true)
       stub_action_locking!(action)
       plan_action(action, acme_org, '/tmp/1234.zip', false)
       assert_difference 'Audit.count', 1 do
@@ -319,6 +320,7 @@ module ::Actions::Katello::Organization
       acme_org = get_organization(:empty_organization)
       rhel7 = katello_repositories(:rhel_7_x86_64)
       acme_org.products.stubs(:redhat).returns([rhel7.product])
+      Organization.any_instance.stubs(:simple_content_access?).returns(true)
 
       stub_action_locking!(action)
       plan_action(action, acme_org)
--- a/test/models/concerns/organization_extensions_test.rb
+++ b/test/models/concerns/organization_extensions_test.rb
@@ -26,5 +26,19 @@ module Katello
       end
       assert_equal @org.manifest_refreshed_at.to_i, current_time.to_i
     end
+
+    def test_clear_syspurpose_status
+      host = @org.hosts.first
+
+      Katello::HostStatusManager::PURPOSE_STATUS.each do |status_class|
+        HostStatus::Status.create(host: host, type: status_class.to_s)
+      end
+
+      assert_equal 5, host.host_statuses.count
+
+      @org.clear_syspurpose_status
+
+      assert_equal 0, host.host_statuses.count
+    end
   end
 end
--- a/test/models/host/subscription_facet_test.rb
+++ b/test/models/host/subscription_facet_test.rb
@@ -90,6 +90,7 @@ module Katello
     end
 
     def test_update_purpose_status
+      host.organization.stubs(:simple_content_access?).returns(false)
       subscription_facet.update_purpose_status(sla_status: :mismatched,
                                                role_status: :unknown,
                                                usage_status: :not_specified,
--- a/test/models/purpose_status_test.rb
+++ b/test/models/purpose_status_test.rb
@@ -2,12 +2,12 @@ require 'katello_test_helper'
 
 module Katello
   module PurposeStatusTests
-    def test_status_matched
-      purpose_mock = mock(purpose_method => :matched)
+    def test_status_unknown
+      purpose_mock = mock(purpose_method => :inexistent)
       Katello::Candlepin::Consumer.any_instance.expects(:system_purpose).returns(purpose_mock)
 
-      assert_equal Katello::PurposeStatus::MATCHED, status.to_status
-      assert_equal 'Matched', status.to_label
+      assert_equal Katello::PurposeStatus::UNKNOWN, status.to_status
+      assert_equal 'Unknown', status.to_label
     end
 
     def test_status_mismatched
@@ -28,13 +28,13 @@ module Katello
       assert_equal 'Not Specified', status.to_label
     end
 
-    def test_status_unknown
-      purpose_mock = mock(purpose_method => :inexistent)
+    def test_status_matched
+      purpose_mock = mock(purpose_method => :matched)
       Katello::Candlepin::Consumer.any_instance.expects(:system_purpose).returns(purpose_mock)
       status.status = status.to_status
 
-      assert_equal Katello::PurposeStatus::UNKNOWN, status.status
-      assert_equal 'Unknown', status.to_label
+      assert_equal Katello::PurposeStatus::MATCHED, status.status
+      assert_equal 'Matched', status.to_label
     end
 
     def test_status_override
--- a/app/lib/actions/katello/organization/manifest_delete.rb
+++ b/app/lib/actions/katello/organization/manifest_delete.rb
@@ -40,6 +40,7 @@ module Actions
         end
 
         def finalize
+          subject_organization.clear_syspurpose_status if subject_organization.simple_content_access?(cached: false)
           subject_organization.audit_manifest_action(_('Manifest deleted'))
         end
       end
--- a/app/lib/actions/katello/organization/manifest_import.rb
+++ b/app/lib/actions/katello/organization/manifest_import.rb
@@ -45,6 +45,7 @@ module Actions
         end
 
         def finalize
+          subject_organization.clear_syspurpose_status if subject_organization.simple_content_access?(cached: false)
           subject_organization.clear_manifest_expired_notifications
           subject_organization.audit_manifest_action(_('Manifest imported'))
         end
--- a/app/services/katello/candlepin/event_handler.rb
+++ b/app/services/katello/candlepin/event_handler.rb
@@ -53,6 +53,7 @@ module Katello
 
       def reindex_purpose_status
         reindex_consumer do
+          return if subscription_facet.host.organization.simple_content_access?
           subscription_facet.update_purpose_status(role_status: message_handler.system_purpose.role_status,
                                                    usage_status: message_handler.system_purpose.usage_status,
                                                    addons_status: message_handler.system_purpose.addons_status,
--- a/app/services/katello/host_status_manager.rb
+++ b/app/services/katello/host_status_manager.rb
@@ -9,5 +9,12 @@ module Katello
       Katello::PurposeAddonsStatus,
       Katello::PurposeStatus,
       Katello::TraceStatus].freeze
+
+    PURPOSE_STATUS = [
+      Katello::PurposeStatus,
+      Katello::PurposeAddonsStatus,
+      Katello::PurposeRoleStatus,
+      Katello::PurposeSlaStatus,
+      Katello::PurposeUsageStatus].freeze
   end
 end
--- a/db/migrate/20200818192230_update_system_purpose_status.rb
+++ b/db/migrate/20200818192230_update_system_purpose_status.rb
@@ -1,0 +1,12 @@
+class UpdateSystemPurposeStatus < ActiveRecord::Migration[6.0]
+  def change
+    purpose_types = Katello::HostStatusManager::PURPOSE_STATUS.map(&:to_s)
+
+    # load both sets of host statuses and *then* update them to make sure we update the correct statuses
+    unknown_statuses = ::HostStatus::Status.where(type: purpose_types, status: 2).pluck(:id)
+    matched_statuses = ::HostStatus::Status.where(type: purpose_types, status: 0).pluck(:id)
+
+    ::HostStatus::Status.where(id: unknown_statuses).update_all(status: Katello::PurposeStatus::UNKNOWN) # 2 => 0
+    ::HostStatus::Status.where(id: matched_statuses).update_all(status: Katello::PurposeStatus::MATCHED) # 0 => 2
+  end
+end
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/content-hosts-helper.service.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/content-hosts-helper.service.js
@@ -41,7 +41,7 @@ angular.module('Bastion.content-hosts').service('ContentHostsHelper',
         this.getHostPurposeStatusIcon = function (statusCode) {
             var code = parseInt(statusCode);
 
-            if (code === 0) { // matched
+            if (code === 2) { // matched
                 return 'pficon pficon-ok';
             } else if (code === 1) { // mismatched
                 return 'pficon pficon-warning-triangle-o';
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/details/content-host-details-info.controller.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/details/content-host-details-info.controller.js
@@ -10,13 +10,14 @@
  * @requires Organization
  * @requires CurrentOrganization
  * @requires CurrentHostsHelper
+ * @requires simpleContentAccessEnabled
  *
  * @description
  *   Provides the functionality for the content host details action pane.
  */
 angular.module('Bastion.content-hosts').controller('ContentHostDetailsInfoController',
-    ['$scope', '$q', 'translate', 'HostSubscription', 'ContentView', 'Organization', 'CurrentOrganization', 'ContentHostsHelper',
-    function ($scope, $q, translate, HostSubscription, ContentView, Organization, CurrentOrganization, ContentHostsHelper) {
+    ['$scope', '$q', 'translate', 'HostSubscription', 'ContentView', 'Organization', 'CurrentOrganization', 'ContentHostsHelper', 'simpleContentAccessEnabled',
+    function ($scope, $q, translate, HostSubscription, ContentView, Organization, CurrentOrganization, ContentHostsHelper, simpleContentAccessEnabled) {
         function doubleColonNotationToObject(dotString) {
             var doubleColonObject = {}, tempObject, parts, part, key, property;
             for (property in dotString) {
@@ -45,6 +46,7 @@ angular.module('Bastion.content-hosts').controller('ContentHostDetailsInfoContro
         $scope.showCVAlert = false;
         $scope.editContentView = false;
         $scope.disableEnvironmentSelection = false;
+        $scope.simpleContentAccessEnabled = simpleContentAccessEnabled;
         $scope.environments = [];
 
         $scope.environments = Organization.readableEnvironments({id: CurrentOrganization});
--- a/engines/bastion_katello/test/content-hosts/content-hosts-helper.service.test.js
+++ b/engines/bastion_katello/test/content-hosts/content-hosts-helper.service.test.js
@@ -11,7 +11,7 @@ describe('Controller: ContentHostsController', function() {
     }));
 
     it("provides a way to get the status icon for system purpose on a host.", function() {
-        expect(ContentHostsHelper.getHostPurposeStatusIcon(0)).toBe("pficon pficon-ok");
+        expect(ContentHostsHelper.getHostPurposeStatusIcon(2)).toBe("pficon pficon-ok");
         expect(ContentHostsHelper.getHostPurposeStatusIcon(1)).toBe("pficon pficon-warning-triangle-o");
     });
 
--- a/engines/bastion_katello/test/content-hosts/details/content-host-details-info.controller.test.js
+++ b/engines/bastion_katello/test/content-hosts/details/content-host-details-info.controller.test.js
@@ -83,7 +83,8 @@ describe('Controller: ContentHostDetailsInfoController', function() {
             Host: Host,
             ContentView: ContentView,
             Organization: Organization,
-            CurrentOrganization: 'ACME_Corporation'
+            CurrentOrganization: 'ACME_Corporation',
+            simpleContentAccessEnabled: false
         });
     }));
 
--- a/test/services/katello/candlepin/event_handler_test.rb
+++ b/test/services/katello/candlepin/event_handler_test.rb
@@ -13,6 +13,7 @@ module Katello
       ::Katello::Pool.any_instance.stubs(:import_data).returns(true)
       ::Katello::Candlepin::MessageHandler.any_instance.stubs(:get_pool_by_reference_id).returns(pool_two)
       ::Katello::Candlepin::MessageHandler.any_instance.stubs(:subscription_facet).returns(subscription_facet)
+      subscription_facet.host.organization.stubs(:simple_content_access?).returns(false)
     end
 
     def message(subject, content = {})


app/models/katello/concerns/organization_extensions.rb
```diff
@@ -228,6 +228,11 @@ def expiring_subscriptions
           subscriptions.select(&:expiring_soon?)
         end
 
+        def clear_syspurpose_status
+          host_purpose = HostStatus::Status.where(type: 'Katello::PurposeStatus')
```
jturel, line 232: This is going to clear out the system purpose status for all hosts - in every organization!

You can limit to the relevant org like this:

```
HostStatus::Status.where(type: 'Katello::PurposeStatus', host: organization.hosts)
```

The issue we've seen with queries like this is that postgresql will not like when there are a LOT of host ids in the subquery, so we can form it a bit differently to avoid that:

```
HostStatus::Status.where(type: 'Katello::PurposeStatus').where('host_id in (?)', organization.hosts.pluck(:id))
```

That said, I think we should clear out all of the other purpose status types since they'd just be hanging around for no reason: PurposeRoleStatus, PurposeSlaStatus, PurposeAddonsStatus. You can pass multiple types as an array in your query!

Of course, you should verify that everything I said is accurate :+1:

app/models/katello/purpose_status.rb
```diff
@@ -38,6 +38,7 @@ def self.to_label(status)
     end
 
     def self.to_status(status, purpose_method, options)
+      return nil if status.host.organization.simple_content_access?
```
jturel, line 41: What does this do?

app/models/katello/purpose_status.rb
```diff
@@ -38,6 +38,7 @@ def self.to_label(status)
     end
 
     def self.to_status(status, purpose_method, options)
```
jturel, line 40: In this class there's a 'relevant?' method which needs to update its logic to check whether the host's org is in SCA mode. In that case it should return false

jturel:
Regarding the test failures: they are failing because of your new code which checks if the org is in SCA. That's expected since `simple_content_access?` issues an HTTP request into candlepin. In this case we don't want to record new VCR cassettes because we aren't concerned with testing that functionality (the `simple_content_access?` method tests themselves should be testing that, if anything...).

The thing to do here is, for the tests which are now failing, to stub the response of `simple_content_access?` to return true or false based on what the test is trying to do. I'm guessing for the current tests you can return `false` and for new tests you can return `true`, broadly speaking.

app/models/katello/purpose_status.rb
jturel: Some extra credit reading would be to see how `relevant?` is used in Foreman

app/models/katello/concerns/organization_extensions.rb
chris1984: Good catch on that, I was able to find the correct way to delete all of the stuff I need to and use your way with a change in the hosts select statement using self.hosts to get the hosts.

app/models/katello/purpose_status.rb
chris1984: I don't even remember why 😆, confirmed with my delete it still works correctly, so just removing it.

app/models/katello/concerns/organization_extensions.rb
```diff
@@ -228,6 +228,13 @@ def expiring_subscriptions
           subscriptions.select(&:expiring_soon?)
         end
 
+        def clear_syspurpose_status
+          # rubocop:disable LineLength
+          host_purpose = HostStatus::Status.where(type: ['Katello::PurposeStatus','Katello::PurposeAddonsStatus','Katello::PurposeUsageStatus','Katello::PurposeRoleStatus','Katello::PurposeSlaStatus']).where('host_id in (?)', self.hosts.pluck(:id))
```
jturel, line 233: This is looking really good. Now we see that the line is very long. We shouldn't bypass rubocop whenever possible, and this is very easy to work around in this case.

You can pull the list of types out of the query and put them somewhere else. Even just a variable above this query in the same method would do - but there's somewhere better: https://github.com/Katello/katello/blob/master/app/services/katello/host_status_manager.rb

Create a new constant there with the class constants - not strings - in an array. You can then reference that constant here and call to_s on its elements to get strings: `MY_CONSTANT.map(&:to_s)`. This achieves not only addressing the rubocop lint, but now all the purpose statuses are grouped together which could be beneficial in the future for new code that might need to know about them. We can change the value of that constant in one place and the code referencing it should continue to work.

Here's a challenge that I would _highly_ recommend: before changing this code, write a passing unit test that has assertions for all the intended behavior of this method. Then go about refactoring as I suggested. If the tests pass then you got it working. If not, you created a bug! Adding tests at the very end of a code change is much less useful than writing them at the start (you know I had to sneak that in there). This is the perfect opportunity to try out some TDD

app/models/katello/purpose_status.rb
```diff
@@ -65,7 +65,11 @@ def to_status(options = {})
     end
 
     def relevant?(_options = {})
-      host.subscription_facet.try(:uuid)
```
jturel, line 68: This can be a one-liner: `host.subscription_facet.try(:uuid) && !host.organization.simple_content_access?`

It says "It's relevant if the hosts's subscription facet has a uuid (it's registered) and the host's org is NOT in SCA mode"

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/details/views/content-host-info.html
```diff
@@ -119,7 +119,7 @@ <h4 translate>Subscriptions</h4>
 
       <dl class="dl-horizontal dl-horizontal-left">
         <dt translate>System Purpose Status</dt>
-        <dd>
+        <dd ng-hide="simpleContentAccessEnabled">
```
jturel, line 122: This doesn't seem to be working - not sure why

test/models/concerns/organization_extensions_test.rb
```diff
@@ -26,5 +26,24 @@ def test_audit_manifest_action
...
+      HostStatus::Status.create(id: 2, host_id: host.id, type: 'Katello::PurposeAddonsStatus')
+      HostStatus::Status.create(id: 3, host_id: host.id, type: 'Katello::PurposeUsageStatus')
+      HostStatus::Status.create(id: 4, host_id: host.id, type: 'Katello::PurposeRoleStatus')
+      HostStatus::Status.create(id: 5, host_id: host.id, type: 'Katello::PurposeSlaStatus')
+
+      assert_equal HostStatus::Status.count, 5
+
+      Organization.any_instance.stubs(:simple_content_access?).returns(true)
+
+      host_purpose = HostStatus::Status.where(type: ['Katello::PurposeStatus','Katello::PurposeAddonsStatus','Katello::PurposeUsageStatus','Katello::PurposeRoleStatus','Katello::PurposeSlaStatus'])
```
jturel, line 43: I think you meant to test your `clear_syspurpose_status` method? ie `some_organization.clear_syspurpose_status`

Looks like this test class already has an org to use defined in its setup method!

test/models/purpose_status_test.rb
```diff
@@ -3,6 +3,7 @@
 module Katello
   module PurposeStatusTests
     def test_status_matched
+      Organization.any_instance.stubs(:simple_content_access?).returns(false)
```
jturel, line 6: I see you've added this stub to each test - while it works it seems like we can take some measure to DRY this up. If you add a setup method to the PurposeStatusTest class in this file, you can achieve that:

```
    def setup
      host.organization.stubs(:simple_content_access?).returns(false)
    end
```

Notice how this stubs `host.organization` vs `Organization.any_instance`. This is the preferred way to do it whereever possible so that we stub the minimum amount of behavior

test/actions/katello/organization_test.rb
```diff
@@ -144,6 +144,8 @@ class ManifestRefreshTest < TestBase
       rhel7 = katello_repositories(:rhel_7_x86_64)
       acme_org.stubs(:owner_details).returns({})
       acme_org.products.stubs(:redhat).returns([rhel7.product])
+      acme_org.stubs(:simple_content_access).returns(false)
+      Organization.any_instance.stubs(:simple_content_access?).returns(false)
```
jturel, line 148: Is this stub necessary? Seems like the one directly above should be sufficient

test/models/host/subscription_facet_test.rb
```diff
@@ -90,6 +90,7 @@ def test_purpose_status_search
     end
 
     def test_update_purpose_status
+      Organization.any_instance.stubs(:simple_content_access?).returns(false)
```
jturel, line 93: My preference would be to stub `host.organization` vs `Organization.any_instance` if possible here

test/controllers/api/v2/hosts_controller_test.rb
```diff
@@ -14,6 +14,7 @@ def models
   end
 
   def host_index_and_show(host)
+    Organization.any_instance.stubs(:simple_content_access?).returns(false)
```
jturel, line 17: Are the stubs in this class necessary? You didn't change the API in this PR..

chris1984: @jturel If I take it out, I am getting a VCR failure shown below:
```
  1) Failure:
Api::V2::HostsControllerTest#test_with_content [/home/vagrant/katello/test/controllers/api/v2/hosts_controller_test.rb:19]:
Expected {"message"=>"\n\n================================================================================\nAn HTTP request has been made that VCR does not know how to handle:\n  GET https://localhost:8443/candlepin/owners/Empty_Organization\n\nThere is currently no cassette in use. There are a few ways\nyou can configure VCR to handle this request:\n\n  * If you're surprised VCR is raising this error\n    and want insight about how VCR attempted to handle the request,\n    you can use the debug_logger configuration option to log more details [1].\n  * If you want VCR to record this request and play it back during future test\n    runs, you should wrap your test (or this portion of your test) in a\n    `VCR.use_cassette` block [2].\n  * If you only want VCR to handle requests made while a cassette is in use,\n    configure `allow_http_connections_when_no_cassette = true`. VCR will\n    ignore this request since it is made when there is no cassette [3].\n  * If you want VCR to ignore this request (and others like it), you can\n    set an `ignore_request` callback [4].\n\n[1] https://www.relishapp.com/vcr/vcr/v/3-0-3/docs/configuration/debug-logging\n[2] https://www.relishapp.com/vcr/vcr/v/3-0-3/docs/getting-started\n[3] https://www.relishapp.com/vcr/vcr/v/3-0-3/docs/configuration/allow-http-connections-when-no-cassette\n[4] https://www.relishapp.com/vcr/vcr/v/3-0-3/docs/configuration/ignore-request\n================================================================================\n\n"} to be nil.
```

I did clean remove it in a few other areas that it was not needed.

test/actions/katello/organization_test.rb
chris1984: Fixed and removed.

test/models/host/subscription_facet_test.rb
chris1984: Updated, inspecting the attributes, I do see now it is better to use the host's organization compared to the other way.

test/models/concerns/organization_extensions_test.rb
chris1984: Fixed and updated, based on what we talked about. I will start the refactor work that you mentioned above shortly.

app/models/katello/purpose_status.rb
chris1984: Fixed, thanks for explaining it, makes more sense now :)

test/models/purpose_status_test.rb
chris1984: Updated, I see how in the area I wrote my main test in how there was a setup method. 

Looking closer at how it works I can see that it creates the top level setup for the test, that the methods can either use ie org, or a setting that will apply to all test methods. 

Let me know if I am reading that correct.

test/controllers/api/v2/hosts_controller_test.rb
jturel: Can you look at the stack trace in test.log and see where the error is raised from? I'm missing how your changes would be affecting the hosts API at all

test/models/purpose_status_test.rb
jturel: Yep - the setup method is run before each test! https://www.rubypigeon.com/posts/minitest-cheat-sheet/

app/models/katello/glue/candlepin/owner.rb
```diff
@@ -36,7 +36,9 @@ def content_access_mode
       end
 
       def simple_content_access?
```
jturel, line 38: this probably isn't working since you need to add the `cached` keyword argument to this method (default to true)

app/lib/actions/katello/organization/manifest_refresh.rb
```diff
@@ -61,6 +61,7 @@ def humanized_name
 
         def finalize
           org = ::Organization.find(input[:organization_id])
+          subject_organization.clear_syspurpose_status(cached: false) if subject_organization.simple_content_access?
```
jturel, line 64: the 'cached' param is on the simple_content_access method :)

test/actions/katello/organization_test.rb
```diff
@@ -144,6 +144,7 @@ class ManifestRefreshTest < TestBase
       rhel7 = katello_repositories(:rhel_7_x86_64)
       acme_org.stubs(:owner_details).returns({})
       acme_org.products.stubs(:redhat).returns([rhel7.product])
+      acme_org.stubs(:simple_content_access?).returns(false)
```
jturel, line 147: The `finalize` method reloads the org so this stub won't work. I think you need to use Organization.any_instance here unfortunately

test/models/concerns/organization_extensions_test.rb
```diff
@@ -26,5 +26,21 @@ def test_audit_manifest_action
       end
       assert_equal @org.manifest_refreshed_at.to_i, current_time.to_i
     end
+
+    def test_clear_syspurpose_status
+      host = @org.hosts.first
+
+      HostStatus::Status.create(id: 1, host_id: host.id, type: 'Katello::PurposeStatus')
```
jturel, line 33: `id` can be removed here - active record will create that for you!

with your new constant in the `manager` class you could also do this:

```
Katello::HostStatusManager::PURPOSE_STATUS.each do |status_class|
  HostStatus::Status.create(host: host, type: status_class.to_s)
end
```

test/models/concerns/organization_extensions_test.rb
```diff
@@ -26,5 +26,21 @@ def test_audit_manifest_action
...
+      HostStatus::Status.create(id: 2, host_id: host.id, type: 'Katello::PurposeAddonsStatus')
+      HostStatus::Status.create(id: 3, host_id: host.id, type: 'Katello::PurposeUsageStatus')
+      HostStatus::Status.create(id: 4, host_id: host.id, type: 'Katello::PurposeRoleStatus')
+      HostStatus::Status.create(id: 5, host_id: host.id, type: 'Katello::PurposeSlaStatus')
+
+      assert_equal HostStatus::Status.count, 5
```
jturel, line 39: For the assertions in this test it's more accurate to say "this host has no purpose statuses left" rather than "there are no status records in the db at all (even though it is true)" because your code change intends to achieve the former, not the latter. If someone came a long and added a fixture for some other host to give it a status, then this test would fail :(

`assert_equal 5, host.host_statuses.count` should be more future-proof and accurate!

test/actions/katello/organization_test.rb
jturel, line 147: Since we're testing an action class I think it'd be beneficial to return `true` for your stub here, and actually execute your method call! (ditto for other actions)

jturel:
@chris1984 for completeness I think we should give the manifest delete action the same treatment as import  & refresh

test/models/concerns/organization_extensions_test.rb
jturel: This refactor is totally up to you - just sharing a different approach. Should still remove `id` however!

jturel:
@chris1984 found one easily fixed issue. When registering a new host or changing the system purpose values on a host the statuses will re-appear in the DB. That's because we're still processing the event from Candlepin. We should disable it here: https://github.com/Katello/katello/blob/817cb4b274f82d645079d32e4aeb05a4c1830e4d/app/services/katello/candlepin/event_handler.rb#L56 by returning early ie `return if subscription_facet.host.organization.simple_content_access?`

Note that I didn't include cached: false since we want to leverage the cache here - because we could be processing a lot of events.

app/models/katello/purpose_status.rb
```diff
@@ -65,7 +65,7 @@ def to_status(options = {})
...
+      host.subscription_facet.try(:uuid) && !host.organization.simple_content_access?
```
jturel, line 68: Per my recent comment: https://github.com/Katello/katello/pull/8859#issuecomment-674956576

I think this check is no longer necessary since there should be no cases where we are in SCA mode and there will be any purpose statuses in the db. You can remove it and verify the behavior when registering a new host in SCA mode, changing syspurpose values, and making sure they don't come back in the db, show in the UI, etc. That also means you can remove all of the stubs you added in the Controller tests since that's where `relevant?` is primarily called!

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/details/views/content-host-info.html
jturel, line 122: I think this same check needs to be done on the above <dt> since I can still see the "System Purpose Status" text

jturel:
Just about there! I found an interesting side effect of this change that we have to consider now: when we take an org back out of SCA mode (this shouldn't be a common case) any host will show its system purpose status as matched (when it really could be anything - we don't have a record of it in the db....)

The status should be 'Unknown' since it's unknown. Eventually we should get an event from Candlepin that'll reset it to the correct status. So, to get it to show Unknown properly we need two changes:

```
diff --git a/app/models/katello/purpose_status.rb b/app/models/katello/purpose_status.rb
index 78dde0f..3d5377b 100644
--- a/app/models/katello/purpose_status.rb
+++ b/app/models/katello/purpose_status.rb
@@ -1,8 +1,8 @@
 module Katello
   class PurposeStatus < HostStatus::Status
-    MATCHED = 0
+    UNKNOWN = 0
     MISMATCHED = 1
-    UNKNOWN = 2
+    MATCHED = 2
     NOT_SPECIFIED = 3
 
     def self.status_map
diff --git a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/content-hosts-helper.service.js b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/content-hosts-helper.service.js
index 033fd10..a681bec 100644
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/content-hosts-helper.service.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/content-hosts-helper.service.js
@@ -41,7 +41,7 @@ angular.module('Bastion.content-hosts').service('ContentHostsHelper',
         this.getHostPurposeStatusIcon = function (statusCode) {
             var code = parseInt(statusCode);
 
-            if (code === 0) { // matched
+            if (code === 2) { // matched
                 return 'pficon pficon-ok';
             } else if (code === 1) { // mismatched
                 return 'pficon pficon-warning-triangle-o';

```

Basically it's just swapping the numeric value of MATCHED and UNKNOWN statuses. The reason this works is that when the status object is NOT in the db its default is 0 which means it's Unknown. The related JS change ensures that we only show the green checkmark for Matched. This is the best idea I could come up with, let me know what you think though.

chris1984:
Manual tests:

Registered 3 clients and confirmed `purpose status` attributes were in the `host_status` table

verfied UI in all hosts - hosts and the content host's content page

### Non SCA Manifest

Uploaded manifest -  `purpose status` attributes still present

Refreshed manifest - `purpose status` attributes still present

Registered a new client - `purpose status` attributes and new ones added

Deleted manifest - `purpose status` attributes still present

### SCA Manifest

Uploaded manifest -  `purpose status` attributes removed from db

Refreshed manifest - `purpose status` attributes still gone from db

Registered a new client - Didn't add any new `purpose status` attributes

Deleted manifest - `purpose status` attributes still gone from db

Switched org back to non-SCA - On content host details page, showing Unkown and System Purpose Status is unhidden:

https://nimbusweb.me/s/share/4547537/gcldruvgsuelccbatk3d

test/models/purpose_status_test.rb
```diff
@@ -2,12 +2,16 @@
 
 module Katello
   module PurposeStatusTests
-    def test_status_matched
-      purpose_mock = mock(purpose_method => :matched)
+    def setup
+      host.organization.stubs(:simple_content_access?).returns(false)
```
jturel, line 6: I think this stub + setup method is no longer needed!

chris1984: Cleaned up the method since we changed things around in the other areas.

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/details/views/content-host-info.html
chris1984: Fixed

test/actions/katello/organization_test.rb
chris1984: Changed to true on upload/refresh/delete. Tests pass and we are now testing my method :)

app/lib/actions/katello/organization/manifest_refresh.rb
chris1984: Moved, prob a copy pasta fail :)

test/actions/katello/organization_test.rb
chris1984: Switched them all to the `Organization.any_instance` stub

test/models/concerns/organization_extensions_test.rb
chris1984: Did the refactor, looks cleaner now too. Thanks for the suggestion, learning new things with this PR :) 👍

test/models/concerns/organization_extensions_test.rb
chris1984: Updated the test to your suggestion, I see with `puts` how it is more accurate and better to have a test that is adaptive compared to something that could fail down the line.

app/models/katello/purpose_status.rb
chris1984: Removed this, and confirmed it works the way we want with the manual testing comment. Less changes the better even though this PR has grown quite a bit haha.

chris1984:
[test katello]

app/models/katello/purpose_status.rb
```diff
@@ -1,8 +1,8 @@
 module Katello
   class PurposeStatus < HostStatus::Status
```
jturel, line 2: Because of the changes in this file, I realize that for users in non-sca mode we are going to be telling them that their Matched hosts are Unknown and vice-versa, so we need a migration in order to correct the data. It's minor and should look something like below. Check out the rails docs on creating a migration if you've never done it before!

```
purpose_types = Katello::HostStatusManager::PURPOSE_STATUS.map(&:to_s)

# load both sets of host statuses and *then* update them to make sure we update the correct statuses
unknown_statuses = ::HostStatus::Status.where(type: purpose_types, status: 2)
matched_statuses = ::HostStatus::Status.where(type: purpose_types, status: 0)

unknown_statuses.update_all(status: Katello::PurposeStatus::UNKNOWN) # 2 => 0
matched_statuses.update_all(status: Katello::PurposeStatus::MATCHED) # 0 => 2
```

chris1984: @jturel updated with a migration. Tested that it works by manually editing my `PurposeStatus` records from 3 to 0 then running the migration and now they are a 2.

jturel: The migration didn't work as expected for me. I ended up seeing ALL the purpose status objects have '2' as the status.

before migration:
```
 #<Katello::PurposeStatus:0x0000000013cc2410 id: 38, type: "Katello::PurposeStatus", status: 0, host_id: 1, reported_at: [FILTERED]>,
 #<Katello::PurposeSlaStatus:0x0000000013cc2000 id: 39, type: "Katello::PurposeSlaStatus", status: 2, host_id: 1, reported_at: [FILTERED]>]
```

after migration:
```
  #<Katello::PurposeStatus:0x0000000013b0ca08 id: 38, type: "Katello::PurposeStatus", status: 2, host_id: 1, reported_at: [FILTERED]>,
 #<Katello::PurposeSlaStatus:0x0000000013b0c800 id: 39, type: "Katello::PurposeSlaStatus", status: 2, host_id: 1, reported_at: [FILTERED]>]
```

suggested change to migration:
```
diff --git a/db/migrate/20200818192230_update_system_purpose_status.rb b/db/migrate/20200818192230_update_system_purpose_status.rb
index 32526b5..76cc7a0 100644
--- a/db/migrate/20200818192230_update_system_purpose_status.rb
+++ b/db/migrate/20200818192230_update_system_purpose_status.rb
@@ -3,10 +3,10 @@ class UpdateSystemPurposeStatus < ActiveRecord::Migration[6.0]
     purpose_types = Katello::HostStatusManager::PURPOSE_STATUS.map(&:to_s)
 
     # load both sets of host statuses and *then* update them to make sure we update the correct statuses
-    unknown_statuses = ::HostStatus::Status.where(type: purpose_types, status: 2)
-    matched_statuses = ::HostStatus::Status.where(type: purpose_types, status: 0)
+    unknown_statuses = ::HostStatus::Status.where(type: purpose_types, status: 2).pluck(:id)
+    matched_statuses = ::HostStatus::Status.where(type: purpose_types, status: 0).pluck(:id)
 
-    unknown_statuses.update_all(status: Katello::PurposeStatus::UNKNOWN) # 2 => 0
-    matched_statuses.update_all(status: Katello::PurposeStatus::MATCHED) # 0 => 2
+    ::HostStatus::Status.where(id: unknown_statuses).update_all(status: Katello::PurposeStatus::UNKNOWN) # 2 => 0
+    ::HostStatus::Status.where(id: matched_statuses).update_all(status: Katello::PurposeStatus::MATCHED) # 0 => 2
   end
 end

```

chris1984: @jturel updated, here is before and after:

Before:
```shell
"id"	"type"	"status"	"host_id"	"reported_at"
"2"	"Katello::PurposeSlaStatus"	"3"	"1"	"17/8/2020 18:09:50.072425"
"3"	"Katello::PurposeRoleStatus"	"3"	"1"	"17/8/2020 18:09:50.092667"
"4"	"Katello::PurposeUsageStatus"	"3"	"1"	"17/8/2020 18:09:50.115356"
"5"	"Katello::PurposeAddonsStatus"	"3"	"1"	"17/8/2020 18:09:50.183421"
"1"	"Katello::SubscriptionStatus"	"0"	"1"	"17/8/2020 18:09:53.219876"
"7"	"Katello::ErrataStatus"	"1"	"1"	"17/8/2020 18:09:59.541609"
"9"	"Katello::PurposeSlaStatus"	"3"	"2"	"17/8/2020 18:11:04.315267"
"10"	"Katello::PurposeRoleStatus"	"3"	"2"	"17/8/2020 18:11:04.336124"
"11"	"Katello::PurposeUsageStatus"	"3"	"2"	"17/8/2020 18:11:04.362418"
"12"	"Katello::PurposeAddonsStatus"	"3"	"2"	"17/8/2020 18:11:04.381024"
"8"	"Katello::SubscriptionStatus"	"2"	"2"	"17/8/2020 18:11:06.463509"
"14"	"Katello::ErrataStatus"	"1"	"2"	"17/8/2020 18:11:13.00389"
"16"	"Katello::PurposeSlaStatus"	"3"	"3"	"17/8/2020 18:12:02.345523"
"17"	"Katello::PurposeRoleStatus"	"3"	"3"	"17/8/2020 18:12:02.350816"
"18"	"Katello::PurposeUsageStatus"	"3"	"3"	"17/8/2020 18:12:02.35574"
"19"	"Katello::PurposeAddonsStatus"	"3"	"3"	"17/8/2020 18:12:02.361128"
"15"	"Katello::SubscriptionStatus"	"2"	"3"	"17/8/2020 18:12:04.964302"
"21"	"Katello::ErrataStatus"	"1"	"3"	"17/8/2020 18:12:10.045735"
"6"	"Katello::PurposeStatus"	"0"	"1"	"17/8/2020 18:09:50.200822"
"20"	"Katello::PurposeStatus"	"0"	"3"	"17/8/2020 18:12:02.366202"
"13"	"Katello::PurposeStatus"	"0"	"2"	"17/8/2020 18:11:04.397999"
```

After updated migration:
```shell
"id"	"type"	"status"	"host_id"	"reported_at"
"2"	"Katello::PurposeSlaStatus"	"3"	"1"	"17/8/2020 18:09:50.072425"
"3"	"Katello::PurposeRoleStatus"	"3"	"1"	"17/8/2020 18:09:50.092667"
"4"	"Katello::PurposeUsageStatus"	"3"	"1"	"17/8/2020 18:09:50.115356"
"5"	"Katello::PurposeAddonsStatus"	"3"	"1"	"17/8/2020 18:09:50.183421"
"1"	"Katello::SubscriptionStatus"	"0"	"1"	"17/8/2020 18:09:53.219876"
"7"	"Katello::ErrataStatus"	"1"	"1"	"17/8/2020 18:09:59.541609"
"9"	"Katello::PurposeSlaStatus"	"3"	"2"	"17/8/2020 18:11:04.315267"
"10"	"Katello::PurposeRoleStatus"	"3"	"2"	"17/8/2020 18:11:04.336124"
"11"	"Katello::PurposeUsageStatus"	"3"	"2"	"17/8/2020 18:11:04.362418"
"12"	"Katello::PurposeAddonsStatus"	"3"	"2"	"17/8/2020 18:11:04.381024"
"8"	"Katello::SubscriptionStatus"	"2"	"2"	"17/8/2020 18:11:06.463509"
"14"	"Katello::ErrataStatus"	"1"	"2"	"17/8/2020 18:11:13.00389"
"16"	"Katello::PurposeSlaStatus"	"3"	"3"	"17/8/2020 18:12:02.345523"
"17"	"Katello::PurposeRoleStatus"	"3"	"3"	"17/8/2020 18:12:02.350816"
"18"	"Katello::PurposeUsageStatus"	"3"	"3"	"17/8/2020 18:12:02.35574"
"19"	"Katello::PurposeAddonsStatus"	"3"	"3"	"17/8/2020 18:12:02.361128"
"15"	"Katello::SubscriptionStatus"	"2"	"3"	"17/8/2020 18:12:04.964302"
"21"	"Katello::ErrataStatus"	"1"	"3"	"17/8/2020 18:12:10.045735"
"6"	"Katello::PurposeStatus"	"2"	"1"	"17/8/2020 18:09:50.200822"
"20"	"Katello::PurposeStatus"	"2"	"3"	"17/8/2020 18:12:02.366202"
"13"	"Katello::PurposeStatus"	"2"	"2"	"17/8/2020 18:11:04.397999"
```

jturel: Your Before state didn't have any of the purpose statuses with status == 2, so we can't be sure that this worked fully. Can you check with some purpose states in 0 and 2, and make sure the status is swapped between them after migration?

chris1984: Updated testing results:

Before:

```shell
"id"	"type"	"status"	"host_id"	"reported_at"
"1"	"Katello::SubscriptionStatus"	"0"	"1"	"17/8/2020 18:09:53.219876"
"7"	"Katello::ErrataStatus"	"1"	"1"	"17/8/2020 18:09:59.541609"
"8"	"Katello::SubscriptionStatus"	"2"	"2"	"17/8/2020 18:11:06.463509"
"14"	"Katello::ErrataStatus"	"1"	"2"	"17/8/2020 18:11:13.00389"
"15"	"Katello::SubscriptionStatus"	"2"	"3"	"17/8/2020 18:12:04.964302"
"21"	"Katello::ErrataStatus"	"1"	"3"	"17/8/2020 18:12:10.045735"
"2"	"Katello::PurposeSlaStatus"	"2"	"1"	"17/8/2020 18:09:50.072425"
"3"	"Katello::PurposeRoleStatus"	"2"	"1"	"17/8/2020 18:09:50.092667"
"4"	"Katello::PurposeUsageStatus"	"2"	"1"	"17/8/2020 18:09:50.115356"
"5"	"Katello::PurposeAddonsStatus"	"2"	"1"	"17/8/2020 18:09:50.183421"
"6"	"Katello::PurposeStatus"	"2"	"1"	"17/8/2020 18:09:50.200822"
"9"	"Katello::PurposeSlaStatus"	"0"	"2"	"17/8/2020 18:11:04.315267"
"10"	"Katello::PurposeRoleStatus"	"0"	"2"	"17/8/2020 18:11:04.336124"
"11"	"Katello::PurposeUsageStatus"	"0"	"2"	"17/8/2020 18:11:04.362418"
"12"	"Katello::PurposeAddonsStatus"	"0"	"2"	"17/8/2020 18:11:04.381024"
"13"	"Katello::PurposeStatus"	"0"	"2"	"17/8/2020 18:11:04.397999"
"16"	"Katello::PurposeSlaStatus"	"0"	"3"	"17/8/2020 18:12:02.345523"
"17"	"Katello::PurposeRoleStatus"	"0"	"3"	"17/8/2020 18:12:02.350816"
"18"	"Katello::PurposeUsageStatus"	"0"	"3"	"17/8/2020 18:12:02.35574"
"19"	"Katello::PurposeAddonsStatus"	"0"	"3"	"17/8/2020 18:12:02.361128"
"20"	"Katello::PurposeStatus"	"0"	"3"	"17/8/2020 18:12:02.366202"
```

After:

```shell
"id"	"type"	"status"	"host_id"	"reported_at"
"1"	"Katello::SubscriptionStatus"	"0"	"1"	"17/8/2020 18:09:53.219876"
"7"	"Katello::ErrataStatus"	"1"	"1"	"17/8/2020 18:09:59.541609"
"8"	"Katello::SubscriptionStatus"	"2"	"2"	"17/8/2020 18:11:06.463509"
"14"	"Katello::ErrataStatus"	"1"	"2"	"17/8/2020 18:11:13.00389"
"15"	"Katello::SubscriptionStatus"	"2"	"3"	"17/8/2020 18:12:04.964302"
"21"	"Katello::ErrataStatus"	"1"	"3"	"17/8/2020 18:12:10.045735"
"2"	"Katello::PurposeSlaStatus"	"0"	"1"	"17/8/2020 18:09:50.072425"
"3"	"Katello::PurposeRoleStatus"	"0"	"1"	"17/8/2020 18:09:50.092667"
"4"	"Katello::PurposeUsageStatus"	"0"	"1"	"17/8/2020 18:09:50.115356"
"5"	"Katello::PurposeAddonsStatus"	"0"	"1"	"17/8/2020 18:09:50.183421"
"6"	"Katello::PurposeStatus"	"0"	"1"	"17/8/2020 18:09:50.200822"
"9"	"Katello::PurposeSlaStatus"	"2"	"2"	"17/8/2020 18:11:04.315267"
"10"	"Katello::PurposeRoleStatus"	"2"	"2"	"17/8/2020 18:11:04.336124"
"11"	"Katello::PurposeUsageStatus"	"2"	"2"	"17/8/2020 18:11:04.362418"
"12"	"Katello::PurposeAddonsStatus"	"2"	"2"	"17/8/2020 18:11:04.381024"
"13"	"Katello::PurposeStatus"	"2"	"2"	"17/8/2020 18:11:04.397999"
"16"	"Katello::PurposeSlaStatus"	"2"	"3"	"17/8/2020 18:12:02.345523"
"17"	"Katello::PurposeRoleStatus"	"2"	"3"	"17/8/2020 18:12:02.350816"
"18"	"Katello::PurposeUsageStatus"	"2"	"3"	"17/8/2020 18:12:02.35574"
"19"	"Katello::PurposeAddonsStatus"	"2"	"3"	"17/8/2020 18:12:02.361128"
"20"	"Katello::PurposeStatus"	"2"	"3"	"17/8/2020 18:12:02.366202"
```

PR #8862, opened 2020-07-30
Fixes #30524 - Add content_views/:id/repositories/show_all endpoint

johnpmitsch:
Adds a way to see all repositories, both added and available to a content view. This is useful for the new CV UI page.

### DIFF
--- a/app/controllers/katello/api/v2/content_view_repositories_controller.rb
+++ b/app/controllers/katello/api/v2/content_view_repositories_controller.rb
@@ -1,0 +1,47 @@
+module Katello
+  class Api::V2::ContentViewRepositoriesController < Api::V2::ApiController
+    # There are other Content View Repository related endpoints that are in RepositoriesController
+    include Katello::Concerns::FilteredAutoCompleteSearch
+
+    before_action :find_content_view
+    before_action :find_organization_from_cv
+
+    # content_views/:id/repositories/show_all
+    # Shows all repositories, added and available to add, for a content view
+    # Undocumented endpoint since the functionality exists in separate calls already.
+    # This was created for ease of pagination and search for the UI
+    param :id, :number, desc: N_("Content View id"), required: true
+    param :only, ["available", "added"], desc: N_("only show a subset of results, either 'available' or 'added'")
+    def show_all
+      query = Repository.readable
+      ids = []
+      ids += @organization.default_content_view.versions.first.repositories.pluck(:id) unless params[:only] == 'added'
+      added_ids = @content_view.repositories.pluck(:id)
+      if params[:only] == 'available'
+        ids -= added_ids
+      else
+        ids += added_ids
+      end
+      query = query.where(id: ids.uniq)
+      sorted_ids = query.sort_by { |repo| repo.in_content_view?(@content_view) ? 0 : 1 }.pluck(:id)
+      sorted_query = Katello::Repository.where(id: sorted_ids)
+
+      # Make sure the added content views are returned first
+      options = { resource_class: Katello::Repository,
+                  deterministic_order: "array_position(array[#{sorted_ids.join(',')}], #{Katello::Repository.table_name}.id)" }
+      repos = scoped_search(sorted_query, nil, nil, options)
+
+      respond_for_index(:collection => repos)
+    end
+
+    private
+
+    def find_content_view
+      @content_view = ContentView.find(params[:content_view_id])
+    end
+
+    def find_organization_from_cv
+      @organization = @content_view.organization
+    end
+  end
+end
--- a/app/controllers/katello/api/v2/content_view_repositories_controller.rb
+++ b/app/controllers/katello/api/v2/content_view_repositories_controller.rb
@@ -24,12 +24,11 @@ module Katello
         CAST (#{kcvr}.repository_id as BOOLEAN) ASC, #{krr}.name
       SQL
 
-      query = Katello::Repository.in_default_view.in_organization(@organization)
+      query = Katello::Repository.readable.in_default_view.in_organization(@organization)
       # Use custom sort to perform the join and order since we need to order by specific content_view
       # and the ORDER BY query needs access to the katello_content_view_repositories table
       custom_sort = ->(sort_query) { sort_query.joins(:root).joins(join_query).order(order_query) }
-      options = { resource_class: Katello::Repository,
-                  custom_sort: custom_sort }
+      options = { resource_class: Katello::Repository, custom_sort: custom_sort }
       repos = scoped_search(query, nil, nil, options)
 
       respond_for_index(:collection => repos)
--- a/app/controllers/katello/api/v2/content_view_repositories_controller.rb
+++ b/app/controllers/katello/api/v2/content_view_repositories_controller.rb
@@ -11,25 +11,10 @@ module Katello
     # Undocumented endpoint since the functionality exists in separate calls already.
     # This was created for ease of pagination and search for the UI
     param :id, :number, desc: N_("Content View id"), required: true
-    param :only, ["available", "added"], desc: N_("only show a subset of results, either 'available' or 'added'")
     def show_all
-      query = Repository.readable
-      ids = []
-      ids += @organization.default_content_view.versions.first.repositories.pluck(:id) unless params[:only] == 'added'
-      added_ids = @content_view.repositories.pluck(:id)
-      if params[:only] == 'available'
-        ids -= added_ids
-      else
-        ids += added_ids
-      end
-      query = query.where(id: ids.uniq)
-      sorted_ids = query.sort_by { |repo| repo.in_content_view?(@content_view) ? 0 : 1 }.pluck(:id)
-      sorted_query = Katello::Repository.where(id: sorted_ids)
-
-      # Make sure the added content views are returned first
-      options = { resource_class: Katello::Repository,
-                  deterministic_order: "array_position(array[#{sorted_ids.join(',')}], #{Katello::Repository.table_name}.id)" }
-      repos = scoped_search(sorted_query, nil, nil, options)
+      query = Katello::Repository.all_for_content_view(@content_view.id);
+      options = { resource_class: Katello::Repository, fixed_query: true }
+      repos = scoped_search(query, nil, nil, options)
 
       respond_for_index(:collection => repos)
     end
--- a/app/controllers/katello/api/v2/content_view_repositories_controller.rb
+++ b/app/controllers/katello/api/v2/content_view_repositories_controller.rb
@@ -12,8 +12,24 @@ module Katello
     # This was created for ease of pagination and search for the UI
     param :id, :number, desc: N_("Content View id"), required: true
     def show_all
-      query = Katello::Repository.all_for_content_view(@content_view.id);
-      options = { resource_class: Katello::Repository, fixed_query: true }
+      kr = Katello::Repository.table_name
+      krr = Katello::RootRepository.table_name
+      kcvr = Katello::ContentViewRepository.table_name
+      join_query = <<-SQL
+        LEFT OUTER JOIN #{kcvr}
+        ON #{kr}.id = #{kcvr}.repository_id
+        AND #{kcvr}.content_view_id = #{@content_view.id}
+      SQL
+      order_query = <<-SQL
+        CAST (#{kcvr}.repository_id as BOOLEAN) ASC, #{krr}.name
+      SQL
+
+      query = Katello::Repository.in_default_view.in_organization(@organization)
+      # Use custom sort to perform the join and order since we need to order by specific content_view
+      # and the ORDER BY query needs access to the katello_content_view_repositories table
+      custom_sort = ->(sort_query) { sort_query.joins(:root).joins(join_query).order(order_query) }
+      options = { resource_class: Katello::Repository,
+                  custom_sort: custom_sort }
       repos = scoped_search(query, nil, nil, options)
 
       respond_for_index(:collection => repos)
--- a/app/views/katello/api/v2/content_view_repositories/show_all.json.rabl
+++ b/app/views/katello/api/v2/content_view_repositories/show_all.json.rabl
@@ -1,0 +1,8 @@
+object false
+
+extends "katello/api/v2/common/metadata"
+
+child @collection[:results] => :results do
+  extends 'katello/api/v2/repositories/base'
+  node(:added_to_content_view) { |repo| repo.in_content_view?(@content_view) }
+end
--- a/app/controllers/katello/api/v2/api_controller.rb
+++ b/app/controllers/katello/api/v2/api_controller.rb
@@ -66,6 +66,7 @@ module Katello
       resource = options[:resource_class] || resource_class
       includes = options.fetch(:includes, [])
       group = options.fetch(:group, nil)
+      deterministic_order = options.fetch(:deterministic_order, "#{query.table_name}.id DESC")
       params[:full_result] = true if options[:csv]
       blank_query = resource.none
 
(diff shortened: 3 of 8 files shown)


app/controllers/katello/api/v2/content_view_repositories_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+      sorted_ids = query.sort_by { |repo| repo.in_content_view?(@content_view) ? 0 : 1 }.pluck(:id)
+      sorted_query = Katello::Repository.where(id: sorted_ids)
+
+      # Make sure the added content views are returned first
+      options = { resource_class: Katello::Repository,
+                  deterministic_order: "array_position(array[#{sorted_ids.join(',')}], #{Katello::Repository.table_name}.id)" }
+      repos = scoped_search(sorted_query, nil, nil, options)
+
+      respond_for_index(:collection => repos)
+    end
```
johnpmitsch, line 27: I'm not sure if this is the best way to achieve what we are trying to, but we need to ensure that that the added content views are returned first in the results and that stays the same across pagination and regardless of the amount of repositories addded to a CV.

With scoped search, this needs to be done in SQL, but there isn't an attribute to cleanly order by to tell if a repository is added to a specific content view.

This is grabbing all the ids of the available and added repositories to a content view and then sorting them by if they have been added to that content view. From there, queries for all the ids and specifies to use the sql sort by ids in the scoped_search method by the new 'deterministic_order' option. 

This way everything stays searchable and paginated and the added content views will always be on the very top of the results.

johnpmitsch:
We discussed making this undocumented since the functionality is already available with two calls to `content_views/:id/repositories` with different parameters to get available and added repos

This would allow us to make breaking changes and structure it the way we need in the UI.

Any objections to this approach?

app/controllers/katello/api/v2/content_view_repositories_controller.rb
jeremylenz: Can you use a JOIN statement to make a single SQL result set with a simple boolean "added/available" column, and then return that?  It may take some SQL thinking to get it working, and we may have to graduate from the SQL that Active Record can give us, but I think it's at least worth investigating.

Using `JOIN`s will be much better than running `in_content_view?` on every row, because it could turn possibly hundreds of SQL queries into one.

`.uniq` and `.sort_by` can get mighty slow in Ruby when you have a lot of records to process.  `.distinct` and `.order` are much faster since they use SQL.  And you could order first by added, then by ID.  For the `[:only]` params, you could use `.where` clauses to limit the results.

johnpmitsch:
I'm having trouble getting the SQL to play nice with the scoped search method, converting to a draft for now. Here is a recap of the progress so far:

Goal: Add API call to get all Repositories for a Content View, both added and available.
Requirements:
- Have a indicator of whether or not the repository was added to the specific content view in the API response
- Added repositories should be shown first in the response
- Search and pagination should not affect the ordering, added repositories should be showed first in the response. For example, if CV foo has 20 repositories added to it and I query with a page size of 5, the first 4 pages will be the added repositories.

The root of the issue is that we don't have a column on the repositories db table to sort by to tell if a repository was added to a specific content view, and we can't add one since it is meant for a specific content view so this data needs to be dynamically determined. This makes it hard to work with the existing scoped_search method since it's not as simple as ordering by "name" or similar.

Attempt 1: Query the repositories then sort the repositories by added or available to the CV in Rails, grab those ids, then tell SQL to use that exact id sort order. See a316f04e7a132d0782bf68f1a7a47f0b49994a51 
- issues: This works fine, but is a very roundabout way of doing so and probably is not the best performance-wise.

Attempt 2: Use SQL to perform joins to get the relevant info and add an "added_to_content_view"  column. Sort based on that column. See f370439902e6e73804c7d529ef342248e8899373
- issues: This works in SQL, but some modifications have to be made to scoped_search, since it re-queries the collection by id. This also is proving to be very brittle. For example, `.count` doesn't work but `.count(:all)` and `.size` do. (I think this is [a known bug](https://github.com/rails/rails/issues/36402))
```sql
LINE 3: WHERE  "katello_repositories"."id" IN
                                           ^
katello=# SELECT Count(*)
FROM   "katello_repositories"
WHERE  "katello_repositories"."id" IN
              (SELECT DISTINCT katello_repositories.*,
                        CASE
                          WHEN katello_content_view_repositories.content_view_id
                               = 4
                                                              THEN 1
                          ELSE 0
                        END AS "added_to_content_view",
                        "katello_repositories"."id"
                                       FROM   katello_repositories
              LEFT JOIN katello_content_view_repositories
                     ON ( katello_repositories.id =
                          katello_content_view_repositories.repository_id )
              LEFT JOIN katello_content_views
                     ON ( katello_content_view_repositories.content_view_id =
                          katello_content_views.id )
                                       WHERE  ( katello_content_views.id = 4
                                                 OR
              katello_content_views.id IS NULL )
                                       ORDER  BY added_to_content_view DESC,
                                                 katello_repositories.id)
;
ERROR:  subquery has too many columns
```
It also doesn't work with some methods like `.pluck(:id)`, where the custom column doesn't exist
```
[9] pry(main)> Katello::Repository.all_for_content_view(4).pluck(:id)
ActiveRecord::StatementInvalid: PG::SyntaxError: ERROR:  syntax error at or near ""added_to_content_view""
LINE 2: ) ORDER BY         CASE WHEN EXISTS "added_to_content_view" ...
```

Right now I'm leaning towards going back to attempt 1 and seeing if there is a way to make things more performant there, but want to share the progress so far and see if there are any other ideas.

@jeremylenz @jturel @jlsherrill any thoughts?

jeremylenz:
I'd feel strongly against returning to Attempt 1.  I feel like we're on the right track with Attempt 2 and it's so close!

Without investigating, I'm guessing that the inability to run certain methods is because the data type of your result isn't an `ActiveRecord::Relation` (could be an array or something else), but either way I'm pretty sure there's still a way and it's worth doing some more poking.

Sometimes the Active Record query methods just can't do what you want.  But this is ok, since it's just SQL.  For example, there's no real difference between `.pluck(:id)` and `SELECT table_name.id FROM ...`  Maybe @chris1984 or another SQL guru can give us some more tips :wink: 

I guess in general the things to pay attention to are:

1. What Ruby data type you are getting from each query result? Array, ActiveRecord::Relation, something else? (Raw SQL, scoped_search and Active Record query methods may all behave differently)
2. What SQL does Active Record write for you?  Is that the SQL you want?
3. What SQL does scoped_search write for you? Is that the SQL you want?
4. Once the query is complete, is it in the form you want?

I think it's worth further investigation with the above in mind.

jeremylenz:
also just had a thought: [External search methods](https://github.com/wvanbergen/scoped_search/wiki/Search-definition#external-search-methods) in scoped_search may be helpful here

chris1984:
@johnpmitsch for the SQL command, can you run an explain on it to see the time it takes for the transaction to complete?

https://www.postgresql.org/docs/9.1/sql-explain.html

jlsherrill:
@johnpmitsch this seems to work within the console, would something similar work within your controller?
```
Katello::Repository.in_default_view.joins(:root).joins("LEFT OUTER JOIN katello_content_view_repositories on katello_repositories.id = katello_content_view_repositories.repository_id and katello_content_view_repositories.content_view_id = 5").order('CAST (katello_content_view_repositories.repository_id as BOOLEAN) ASC, katello_root_repositories.name')
```

passing in a sort option via the api isn't going to work the same way it does with an index call (which uses scoped search today), i think we'd need to decide if its that important here to sort on anything other than name

johnpmitsch:
@jeremylenz updated thanks to @jlsherrill's suggestion, we determined off-thread it would work inside a custom_sort and still keep the appropriate sort order.

app/views/katello/api/v2/content_view_repositories/show_all.json.rabl
```diff
@@ -0,0 +1,8 @@
+object false
+
+extends "katello/api/v2/common/metadata"
+
+child @collection[:results] => :results do
+  extends 'katello/api/v2/repositories/base'
+  node(:added_to_content_view) { |repo| repo.in_content_view?(@content_view) }
```
jeremylenz, line 7: Won't this add a SQL query for each result row?  Perhaps the query in `content_view_repositories_controller#show_all` could be even further improved by using Active Record [`.select`](https://guides.rubyonrails.org/active_record_querying.html#selecting-specific-fields), something like

```rb
query = Katello::Repository.readable.in_default_view.in_organization(@organization).select("#{kr}.*, #{krr}.*, #{kcvr}.*, (#{kcvr}.repository_id IS NOT NULL) AS added")
```

(not sure if that condition is correct, double-check me)

Then each result will gain an `.added` method that returns true or false.  So here, in your rabl, you could just say 
```rb
attributes :added
```
with no need to call a method that runs more SQL.  Hopefully this would still work with `scoped_search`..  :crossed_fingers: 

And as a side bonus, it would simplify the `order_query` so you could just say `.order("added")`

jlsherrill: Because @content_view is an instance variable, its repositories association should get lazy only once, and not cause a query per repo, unless i misreading something.  Does that make sense?

jeremylenz: @jlsherrill aah yeah, I didn't think about that! Makes sense.

johnpmitsch: @jeremylenz are you all set with this or would like some changes here? 

The issue I ran into with using a custom column is we ultimately put everything in a subquery since we [re-query by ID](https://github.com/Katello/katello/blob/c50717c04f6b885d91da4280ae5b0de59cddc5b9/app/controllers/katello/api/v2/api_controller.rb#L81) so a custom column like that is not accessible when it comes to the rabl. I tried to modify scoped_search, but it didn't prove to be easy without making breaking changes.

This was the source of a lot of trouble here, but doing everything in the custom_search including the ordering allowed it to work as expected and be sorted. It does require the method to be called in the rabl to tell if it was added_for_repo, but I don't see a huge performance hit with that, we actually do a lot of queries per-repo, here are some for just one repo, notice they take 1ms or less: (not that we couldn't speed up this and many other db queries in our app :smile:)

```
2020-08-20T00:56:34 [D|sql|31374920]   ↳ /home/vagrant/katello/app/models/katello/repository.rb:721:in `docker_meta_tag_count'
2020-08-20T00:56:34 [D|sql|31374920]    (0.8ms)  SELECT COUNT(*) FROM "katello_rpms" INNER JOIN "katello_repository_rpms" ON "katello_rpms"."id" = "katello_repository_rpms"."rp
m_id" WHERE "katello_repository_rpms"."repository_id" = $1  [["repository_id", 110]]
2020-08-20T00:56:34 [D|sql|31374920]   ↳ /home/vagrant/katello/app/views/katello/api/v2/repositories/base.json.rabl:33:in `block in cached_source_450191144403171490'
2020-08-20T00:56:34 [D|sql|31374920]    (1.0ms)  SELECT COUNT(*) FROM "katello_srpms" INNER JOIN "katello_repository_srpms" ON "katello_srpms"."id" = "katello_repository_srpms"
."srpm_id" WHERE "katello_repository_srpms"."repository_id" = $1  [["repository_id", 110]]
2020-08-20T00:56:34 [D|sql|31374920]   ↳ /home/vagrant/katello/app/views/katello/api/v2/repositories/base.json.rabl:34:in `block in cached_source_450191144403171490'
2020-08-20T00:56:34 [D|sql|31374920]   CACHE  (0.0ms)  SELECT COUNT(*) FROM "katello_rpms" INNER JOIN "katello_repository_rpms" ON "katello_rpms"."id" = "katello_repository_rpm
s"."rpm_id" WHERE "katello_repository_rpms"."repository_id" = $1  [["repository_id", 110]]
2020-08-20T00:56:34 [D|sql|31374920]   ↳ /home/vagrant/katello/app/views/katello/api/v2/repositories/base.json.rabl:35:in `block in cached_source_450191144403171490'
2020-08-20T00:56:34 [D|sql|31374920]    (0.7ms)  SELECT COUNT(*) FROM "katello_package_groups" INNER JOIN "katello_repository_package_groups" ON "katello_package_groups"."id" =
 "katello_repository_package_groups"."package_group_id" WHERE "katello_repository_package_groups"."repository_id" = $1  [["repository_id", 110]]
2020-08-20T00:56:34 [D|sql|31374920]   ↳ /home/vagrant/katello/app/views/katello/api/v2/repositories/base.json.rabl:36:in `block in cached_source_450191144403171490'
2020-08-20T00:56:34 [D|sql|31374920]    (0.7ms)  SELECT COUNT(*) FROM "katello_errata" INNER JOIN "katello_repository_errata" ON "katello_errata"."id" = "katello_repository_err
ata"."erratum_id" WHERE "katello_repository_errata"."repository_id" = $1  [["repository_id", 110]]
2020-08-20T00:56:34 [D|sql|31374920]   ↳ /home/vagrant/katello/app/views/katello/api/v2/repositories/base.json.rabl:37:in `block in cached_source_450191144403171490'
2020-08-20T00:56:34 [D|sql|31374920]    (0.7ms)  SELECT COUNT(*) FROM "katello_puppet_modules" INNER JOIN "katello_repository_puppet_modules" ON "katello_puppet_modules"."id" =
 "katello_repository_puppet_modules"."puppet_module_id" WHERE "katello_repository_puppet_modules"."repository_id" = $1  [["repository_id", 110]]
2020-08-20T00:56:34 [D|sql|31374920]   ↳ /home/vagrant/katello/app/views/katello/api/v2/repositories/base.json.rabl:38:in `block in cached_source_450191144403171490'
2020-08-20T00:56:34 [D|sql|31374920]    (0.9ms)  SELECT COUNT(*) FROM "katello_files" INNER JOIN "katello_repository_file_units" ON "katello_files"."id" = "katello_repository_f
ile_units"."file_unit_id" WHERE "katello_repository_file_units"."repository_id" = $1  [["repository_id", 110]]
2020-08-20T00:56:34 [D|sql|31374920]   ↳ /home/vagrant/katello/app/views/katello/api/v2/repositories/base.json.rabl:39:in `block in cached_source_450191144403171490'
2020-08-20T00:56:34 [D|sql|31374920]    (0.6ms)  SELECT COUNT(*) FROM "katello_debs" INNER JOIN "katello_repository_debs" ON "katello_debs"."id" = "katello_repository_debs"."de
b_id" WHERE "katello_repository_debs"."repository_id" = $1  [["repository_id", 110]]
2020-08-20T00:56:34 [D|sql|31374920]   ↳ /home/vagrant/katello/app/views/katello/api/v2/repositories/base.json.rabl:40:in `block in cached_source_450191144403171490'
2020-08-20T00:56:34 [D|sql|31374920]    (0.7ms)  SELECT COUNT(*) FROM "katello_module_streams" INNER JOIN "katello_repository_module_streams" ON "katello_module_streams"."id" =
 "katello_repository_module_streams"."module_stream_id" WHERE "katello_repository_module_streams"."repository_id" = $1  [["repository_id", 110]]
2020-08-20T00:56:34 [D|sql|31374920]   ↳ /home/vagrant/katello/app/views/katello/api/v2/repositories/base.json.rabl:41:in `block in cached_source_450191144403171490'
2020-08-20T00:56:34 [D|sql|31374920]    (0.7ms)  SELECT COUNT(*) FROM "katello_ansible_collections" INNER JOIN "katello_repository_ansible_collections" ON "katello_ansible_collections"."id" = "katello_repository_ansible_collections"."ansible_collection_id" WHERE "katello_repository_ansible_collections"."repository_id" = $1  [["repository_id", 110]]
2020-08-20T00:56:34 [D|sql|31374920]   ↳ /home/vagrant/katello/app/views/katello/api/v2/repositories/base.json.rabl:42:in `block in cached_source_450191144403171490'
2020-08-20T00:56:34 [D|sql|31374920]   Katello::Repository Exists? (1.1ms)  SELECT 1 AS one FROM "katello_repositories" INNER JOIN "katello_content_view_repositories" ON "katello_repositories"."id" = "katello_content_view_repositories"."repository_id" WHERE "katello_content_view_repositories"."content_view_id" = $1 AND "katello_repositories"."id" = $2 LIMIT $3  [["content_view_id", 10], ["id", 110], ["LIMIT", 1]]
```

jeremylenz: After Justin's explanation I think my concern is addressed.. I'm good with this :+1:

PR #8863, opened 2020-07-30
Fixes #30530 - Add repair-sync on capsule for deb

m-bucher:


app/lib/actions/pulp/orchestration/repository/smart_proxy_sync.rb
```diff
@@ -15,6 +15,7 @@ def plan(repository, smart_proxy, options)
               end
```
m-bucher, line 15: Not sure if we should do this for `deb` as well?

jlsherrill: We do it so that the capsule sync is exact, because yum and puppet were the only ones to support  'remove_missing' in pulp2.     For example:

1.  Create a repo with pkgs a,b,c in some repo
2.  sync it to a remote smart proxy
3.  remove package b from the repo on the katello server
4.  Sync it again to the smart proxy

In order to have b removed from the smart proxy you either:  a) need to use remove_missing b) if remove_missing isn't available, you just have to remove everything from the repo

option b) only really works for small repos, so ideally the debian plugin would support 'remove_missing' and not do this 'UnassociateUnits' madness

jlsherrill: But that might be best left for a pulp3 feature :)

quba42:
Do we also need to adjust the docs/text for the smart proxy sync as part of this change?

m-bucher:
@quba42 added `deb` to description

jlsherrill:
Thanks @m-bucher !

PR #8872, opened 2020-07-31
Fixes #30541 - properly import errata dates from pulp3

jlsherrill:


app/services/katello/pulp3/erratum.rb
```diff
@@ -26,8 +26,10 @@ def self.ids_for_repository(repo_id)
       end
 
       def update_model(model)
-        keys = %w(title id severity issued type description reboot_suggested solution updated summary)
+        keys = %w(title id severity issued_date type description reboot_suggested solution updated_date summary)
```
parthaa, line 29: Did this change between pulp2 and 3 ?

jlsherrill: it did, what i'm not 100% sure about is if it ever worked with pulp3.  Since there was no test for it, my guess is no

parthaa: do we have to keep support for both?

parthaa: Oh wait it looks like this code will only run for  pulp3 erratum

PR #8869, opened 2020-07-31
Fixes #30532 - CV publish with separate package env copy task

sjha4:
Currently blocked on https://pulp.plan.io/issues/7248

To test:
1. Set up pulp_rpm 3.5.0 and then apply this patch https://github.com/pulp/pulp_rpm/pull/1794 and restart pulp
2. Sync RHEL 8 Appstream repo
3. Create content view with any filter and publish.
4. Look at the pulp3 data with something like:
curl --cert /etc/pki/katello/certs/pulp-client.crt --key /etc/pki/katello/private/pulp-client.key https://`hostname`/pulp/api/v3/repositories/rpm/rpm/`997aa028-3467-47a5-95b3-d92dbaf43b1e`/versions/ | python -m json.tool
of the target repo to make sure all package environment were added to the latest version.

PR #8873, opened 2020-07-31
Fixes #30389 - Optimze false for repo syncs

sjha4:
1. Create and Sync a repo with mirror set to true.
2. Re-sync the repo.
3. Contents should not get unassociated.

### DIFF
--- a/app/services/katello/pulp3/repository.rb
+++ b/app/services/katello/pulp3/repository.rb
@@ -193,8 +193,12 @@ module Katello
         ignore_404_exception { api.repositories_api.delete(href) } if href
       end
 
+      def sync_params
+        {remote: repo.remote_href, mirror: repo.root.mirror_on_sync}
+      end
+
       def sync
-        sync_url_params = {remote: repo.remote_href, mirror: repo.root.mirror_on_sync}
+        sync_url_params = sync_params
         skip_type_param = skip_types
         sync_url_params[:skip_types] = skip_type_param if skip_type_param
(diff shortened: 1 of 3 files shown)


app/services/katello/pulp3/repository.rb
```diff
@@ -194,7 +194,7 @@ def delete(href = repository_reference.try(:repository_href))
       end
 
       def sync
-        sync_url_params = {remote: repo.remote_href, mirror: repo.root.mirror_on_sync}
+        sync_url_params = {remote: repo.remote_href, mirror: repo.root.mirror_on_sync, optimize: false}
```
jlsherrill, line 197: actually, will this work for non-yum repos?  (will it be ignored)?

jlsherrill: Doesn't look like:  `optimize` is not a valid attribute in `PulpFileClient::RepositorySyncURL`. Please check the name to make sure it's valid. List of attributes: [:remote, :mirror]


I think we need to only add it if its a yum type

sjha4: Waiting for tests to fail to find out and re-record.. :)

sjha4: Yes..needs to be for only rpm..Updating..

sjha4:
Moved this to https://github.com/Katello/katello/pull/8869 with gem update and re-recorded VCRs. Closing this.

PR #8876, opened 2020-08-05
Fixes #30561 - allow sortable_version macro

lzap:
This is currently untested as I am reinstalling my devsetup from scratch, but I'd appreciate quick look and test as we have a BZ for this.

Thanks!

### DIFF
--- a/lib/katello/plugin.rb
+++ b/lib/katello/plugin.rb
@@ -218,7 +218,6 @@ Foreman::Plugin.register :katello do
        :parent => :lab_features_menu,
        :turbolinks => false
 
-  allowed_template_helpers :subscription_manager_configuration_url, :repository_url
   extend_template_helpers Katello::KatelloUrlsHelper
   extend_template_helpers Katello::Concerns::BaseTemplateScopeExtensions
 
--- a/app/lib/katello/concerns/base_template_scope_extensions.rb
+++ b/app/lib/katello/concerns/base_template_scope_extensions.rb
@@ -154,6 +154,10 @@ module Katello
       end
       # rubocop:enable Metrics/MethodLength
 
+      def sortable_version(version)
+        Util::Package.sortable_version(version)
+      end
+
       private
 
       def host_subscription_facet(host)


lib/katello/plugin.rb
```diff
@@ -218,8 +218,9 @@
        :parent => :lab_features_menu,
        :turbolinks => false
 
-  allowed_template_helpers :subscription_manager_configuration_url, :repository_url
+  allowed_template_helpers :subscription_manager_configuration_url, :repository_url, :sortable_version
```
tbrisker, line 221: No need to add here, this is handled by `extend_template_helpers` (in fact, the other two are also redundant)

lib/katello/plugin.rb
```diff
@@ -218,8 +218,9 @@
...
   extend_template_helpers Katello::KatelloUrlsHelper
+  extend_template_helpers Katello::KatelloVersionHelper
   extend_template_helpers Katello::Concerns::BaseTemplateScopeExtensions
```
tbrisker, line 224: Should we just add this method to the base template scope extensions? considering it is not actually a helper that is used elsewhere.

lib/katello/plugin.rb
lzap: Is it safe to remove the whole line? We will be backporting this one.

lib/katello/plugin.rb
lzap: Do you mean to move this into core? There was actually NACK on this thus I am sending my PR into Katello.

tbrisker: No, i mean instead of adding another module, add the new method in Katello::Concerns::BaseTemplateScopeExtensions which is where all the template extensions live

lib/katello/plugin.rb
tbrisker: it is, but it can be done in a separate commit:

https://github.com/theforeman/foreman/blob/develop/app/registries/foreman/plugin.rb#L578

lzap:
Ok now I get it, all done!

tbrisker:
[test katello]

PR #8877, opened 2020-08-05
Fixes #30580 - Correctly handles org selection

parthaa:
The fix for https://projects.theforeman.org/issues/28998 upgraded
angular ui router to 1.0.25. This was a major update and there were some
deprecated changes deleted.
Things like org-selector and other places in
the bastion code used events like stateChangeStart or stateChangeError
or stateChangeSuccess. Support for these got removed in the update.

This commit adds the necessary stateEvents.js file and module which was
added as a backward compatibility adapter script.
It adds the script below to the vendors and loads them in memory
https://cdnjs.cloudflare.com/ajax/libs/angular-ui-router/1.0.25/stateEvents.js

This commit allows for Any Organization content->products page to
properly redirect user to the org selector.

### DIFF
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/organizations/fenced-pages.service.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/organizations/fenced-pages.service.js
@@ -21,7 +21,10 @@ angular.module('Bastion.organizations').service('FencedPages', ['$state',
             'products',
             'puppet-modules',
             'subscriptions',
-            'sync-plans'
+            'sync-plans',
+            'files',
+            'debs',
+            'ostree-branches'
         ];
 
         function getRootPath(path) {
@@ -43,7 +46,13 @@ angular.module('Bastion.organizations').service('FencedPages', ['$state',
 
         this.isFenced = function (toState) {
             var stateUrl = $state.href(toState);
-            return fencedPages.indexOf(getRootPath(stateUrl)) !== -1;
+            var rootPath;
+            if (_.isEmpty(stateUrl)) {
+                rootPath = toState.templateUrl.replace('_', '-').split('/')[0];
+            } else {
+                rootPath = getRootPath(stateUrl);
+            }
+            return fencedPages.indexOf(rootPath) !== -1;
         };
     }
 ]);
--- a/engines/bastion_katello/test/organizations/organization-selector.controller.test.js
+++ b/engines/bastion_katello/test/organizations/organization-selector.controller.test.js
@@ -17,7 +17,7 @@ describe('Controller: OrganizationSelectorController', function() {
         $window = {location: {href: ''}};
 
         Organization.select = function (params) {
-            return {'$promise': {catch: function (func) { func.call() } }};
+            return {'$promise': { then: function (func) { func.call() } }};
         };
 
         $controller('OrganizationSelectorController', {
@@ -33,7 +33,7 @@ describe('Controller: OrganizationSelectorController', function() {
     });
 
     it("should provide selecting an organization in the backend", function () {
-        $scope.$broadcast('$stateChangeSuccess', {}, {toState: '/product'});
+        $scope.$stateParams = {"toState": '/foo'};
         $scope.selectOrganization({id: 1, name: 'Default Organization'});
 
         expect($window.location.href).toBeDefined();
--- a/webpack/containers/Application/config.js
+++ b/webpack/containers/Application/config.js
@@ -38,19 +38,19 @@ export const links = [
   },
   {
     path: 'module_streams',
-    component: withHeader(ModuleStreams, { title: __('Module Streams') }),
+    component: WithOrganization(withHeader(ModuleStreams, { title: __('Module Streams') })),
   },
   {
     path: 'module_streams/:id([0-9]+)',
-    component: withHeader(ModuleStreamDetails, { title: __('Module Stream Details') }),
+    component: WithOrganization(withHeader(ModuleStreamDetails, { title: __('Module Stream Details') })),
   },
   {
     path: 'ansible_collections',
-    component: withHeader(AnsibleCollections, { title: __('Ansible Collections') }),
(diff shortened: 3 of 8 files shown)


parthaa:
To test this
1) Select `Any Organization` in the org selector
2) Select `Content -> Products`
3) It should ask you to pick an org instead of showing an empty products list page.

sjha4:
Works correctly on katello index pages..
1. Content > Files page seems to not redirect to the org selector page. 
2. Do we want to change behavior of the details page etc?

parthaa:
> One page that didn't redirect to org selection page was Content > Files page.

Fixed this and a couple of others including OS Tree Branches . The only one not working is the ```Content -> Module Streams``` page. Trying to figure that out. May be create a separate issue ?

sjha4:
Works for me with all pages except the module streams (and Ansible collections) page as you mentioned. Although I don't get any error or anything with Any org selected. Instead I see records across orgs on the page which doesn't seem too bad although it's a slightly different behavior that other pages. 
Should the behavior of "Any Organization" be to show records across all orgs? That seems more logical based on selection  "Any Organization" but we do want to check user org permissions etc so it might be a broader discussion. We can tackle that in a separate issue given this BZ is to address a regression.

engines/bastion_katello/app/assets/javascripts/bastion_katello/organizations/fenced-pages.service.js
```diff
@@ -21,7 +21,11 @@ angular.module('Bastion.organizations').service('FencedPages', ['$state',
             'products',
             'puppet-modules',
             'subscriptions',
-            'sync-plans'
+            'sync-plans',
+            'files',
+            'module-streams',
```
jturel, line 26: I'm not sure this will affect a react page?

jeremylenz: I don't see how it would

webpack/containers/Application/config.js
```diff
@@ -38,11 +38,11 @@ export const links = [
   },
   {
     path: 'module_streams',
-    component: withHeader(ModuleStreams, { title: __('Module Streams') }),
+    component: WithOrganization(withHeader(ModuleStreams, { title: __('Module Streams') })),
   },
   {
     path: 'module_streams/:id([0-9]+)',
-    component: withHeader(ModuleStreamDetails, { title: __('Module Stream Details') }),
+    component: WithOrganization(withHeader(ModuleStreamDetails, { title: __('Module Stream Details') })),
```
jturel, line 45: Do ansible collections need the same treatment?

parthaa: oh quite possible @jjeffers @sajha ^

parthaa: hmm samir stated the same above. Going to check

parthaa:
> Works for me with all pages except the module streams (and Ansible collections) page as you mentioned. Although I don't get any error or anything with Any org selected. Inst

Should work fine for module streams and ansible collections now.

The only pages where this doesnt happen is the details page. For some reason the ```$state.href``` returns empty for angular 1.0 vs prior.

```
> $state.href("product.repositories")
null


In 6.7
> $state.href("product.repositories")
"/products//repositories"

```

webpack/containers/Application/config.js
sjha4: You didn't give your treatment to ansible collections with the last commit.. :smile:

parthaa:
The details page should be fixed.

webpack/containers/Application/config.js
parthaa: Should be fixed for Ansible collections also

engines/bastion_katello/app/assets/javascripts/bastion_katello/organizations/fenced-pages.service.js
```diff
@@ -42,8 +45,14 @@ angular.module('Bastion.organizations').service('FencedPages', ['$state',
         };
 
         this.isFenced = function (toState) {
-            var stateUrl = $state.href(toState);
-            return fencedPages.indexOf(getRootPath(stateUrl)) !== -1;
+            var stateUrl = $state.href(toState),
+            rootPath;
```
jeremylenz, line 49: Declaring multiple variables in a single `var` statement is fine, but it's confusing if it's on two lines.  I would just say
```js
var rootPath;
```
instead

engines/bastion_katello/test/organizations/organization-selector.controller.test.js
```diff
@@ -16,8 +16,11 @@ describe('Controller: OrganizationSelectorController', function() {
         $state = $injector.get('$state');
         $window = {location: {href: ''}};
 
+        var selectFunction = function (func) { func.call() };
         Organization.select = function (params) {
-            return {'$promise': {catch: function (func) { func.call() } }};
+            return {'$promise': { then:selectFunction,
+                                 catch: selectFunction }
```
jeremylenz, line 22: I'm assuming this is something similar to [$q](https://docs.angularjs.org/api/ng/service/$q) ?  Doesn't make much sense to me, but I'll trust you since you're the Angular expert ;)

Looks like `catch` isn't being used anymore?  (even if it is, doesn't make sense to me to have the same callback for `then` and `catch`)

parthaa: I m trying to mock the promise method in the unit test and have it run accordingly. I dont think the catch is necessary here going to remove.

engines/bastion_katello/app/assets/javascripts/bastion_katello/organizations/fenced-pages.service.js
parthaa: updated

parthaa:
[test katello]

engines/bastion_katello/test/organizations/organization-selector.controller.test.js
parthaa: updated

PR #8879, opened 2020-08-07
Fixes #30591 - ManageManifestModal: remove componentWillMount and han…

jturel:
…dle updates based on prop changes

This PR does the following:

* prevents unnecessary re-mounting of ManageManifestModal by removing the `key` prop on `SubscriptionsPage` where it renders the component.
* removes componentWillMount in ManageManifestModal
* removes unnecessary state handling ie `hideModalAfterAction`
* implements `componentDidUpdate` to handle:
  -  hiding the modal when the task starts
  - reloading the organization and manifest history when a task completes
* renamed `manifestActionInProgress` to `manifestActionStarted` to clarify its meaning and usage
* add a selector for the same
* Fix tooltip not showing up on the Delete button in ManageManifestModal
* Teach SubscriptionsReducer to set `manifestActionInProgress` when user updates quantity of a Red Hat subscription in the table, or uses the table Delete button to remove subscriptions

**To test**
Verify that when any manifest action happens on the Subscriptions page that all other manifest actions are disabled (when *_REQUEST is dispatched)

Manifest actions include importing, refreshing, deleting the manifest, changing the quantity of subscriptions through the table, or deleting subscriptions from the manifest through the table.

When a manifest task begins the manifest modal should close ( when *_SUCCESS is dispatched)

After the task completes, everything should go back to its normal state


jturel:
@jeremylenz updated - let me know if that fixes it!

jturel:
Thanks @jeremylenz ! Now I'll get it downstream. Stay tuned

PR #8880, opened 2020-08-10
Fixes #30619 - Add apt and zypper to job template install_errata_-_katello_ansible_default.erb

knoppi:
Current version only supports installing upgrades through yum, added check and code for zypper as well as apt


chris1984:
[test katello]

PR #8882, opened 2020-08-10
[CP] 3.16.0 GA cherry picks and version bump

ianballou:


sjha4:
@ianballou : Looks good to go to me..Are we waiting for anything else or just an ack?

ianballou:
@sjha4 just an ACK, everything else is good.

PR #8883, opened 2020-08-10
Fixes #30624 - don't exclude lib repo on compare

jlsherrill:


jlsherrill:
instructions to reproduce are in the issue, updated a test to also reproduce the issue

jjeffers (approved):
Tested this with a CVV published with a filter that excluded a single package.

Before applying PR:
```
{
  "total": 31,
  "subtotal": 31,
  "page": "1",
  "per_page": 20,
  "error": null,
  "search": null,
  "sort": {
    "by": null,
    "order": null
  },
```

After:
```
{
  "total": 32,
  "subtotal": 32,
  "page": "1",
  "per_page": 20,
  "error": null,
  "search": null,
  "sort": {
    "by": null,
    "order": null
  },
```

PR #8884, opened 2020-08-11
Fix #30626 - filtering applied errata by host search

ezr-ondrej:
There is a desire to have applied errata filtrable by host to provide similar functionality to a applicable errata.

I must say it feels quite hacky, but the request is valid, so OTOH I feel we should do our best to implement it.

### DIFF
--- a/app/lib/katello/concerns/base_template_scope_extensions.rb
+++ b/app/lib/katello/concerns/base_template_scope_extensions.rb
@@ -87,7 +87,7 @@ module Katello
       end
 
       # rubocop:disable Metrics/MethodLength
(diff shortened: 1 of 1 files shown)


ezr-ondrej:
Ping? Any opinions?

sjha4 (approved):
Looks good to me with the associated report change adding optional host filter. Ack :+1: 
Thanks @ezr-ondrej !

sjha4:
@ezr-ondrej : Should we wait for the foreman PR to get merged? Should be ok with the default nil value either way. Thoughts?

ezr-ondrej:
The core pr depends on this one :)
This adds the support for the change in the template.

sjha4:
Thanks @ezr-ondrej .. Merging!

PR #8890, opened 2020-08-14
Refs #29926 - Add  Simple Content Access status in Organization details

piyushtiw:
Changes Done = Added simple content access column at org list page

Changes need to be done = In Organization details, Add a new section to Organization edit screen.
Show whether Simple Content Access is enabled and allow toggle.


piyushtiw:
@jeremylenz Please have a look.

PR #8893, opened 2020-08-17
Fixes #30655 - Make katello reset work on pip installed directory

parthaa:


lib/katello/tasks/reset.rake
```diff
@@ -27,7 +27,10 @@ namespace :katello do
       system("sudo systemctl stop 'pulpcore-worker@*' --all")
       system("sudo runuser - postgres -c 'dropdb pulpcore'")
       system("sudo runuser - postgres -c 'createdb pulpcore'")
-      Dir.chdir('/usr/lib/python3.6/site-packages/pulpcore') do
+      dir = '/usr/local/lib/python3.6/site-packages/pulpcore'
+      dir = '/usr/lib/python3.6/site-packages/pulpcore' unless Dir.exist?(dir)
+
+      Dir.chdir(dir) do
```
jlsherrill, line 33: actually you can just always chdir to  /tmp and it should work fine!

parthaa: updated! works well

PR #8896, opened 2020-08-19
Fixes #30664 - add installable models for host

snagoor:
added new `installable_errata` and `installable_rpms` models for Report Templates that needs purely `installable` report data.

### DIFF
--- a/app/lib/katello/concerns/base_template_scope_extensions.rb
+++ b/app/lib/katello/concerns/base_template_scope_extensions.rb
@@ -78,6 +78,18 @@ module Katello
         host.applicable_rpms.where(name: package).order(:version_sortable).limit(1).pluck(:nvra).first
       end
 
+      def host_installable_errata_ids(host)
+        host.installable_errata.map(&:errata_id)
+      end
+
+      def host_installable_errata_filtered(host, filter = '')
+        host.installable_errata.includes(:cves).search_for(filter)
+      end
+
+      def host_latest_installable_rpm_version(host, package)
+        host.installable_rpms.where(name: package).order(:version_sortable).limit(1).pluck(:nvra).first
+      end
+
       def load_pools(search: '', includes: nil)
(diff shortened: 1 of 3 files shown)


ares:
This needs a rebase now

snagoor:
> This needs a rebase now

@ares done :+1:

app/lib/katello/concerns/base_template_scope_extensions.rb
```diff
@@ -78,6 +78,18 @@ def host_latest_applicable_rpm_version(host, package)
         host.applicable_rpms.where(name: package).order(:version_sortable).limit(1).pluck(:nvra).first
       end
 
+      def host_installable_errata_ids(host)
+        host.installable_errata.map(&:errata_id)
```
jlsherrill, line 82: using  .pluck()  would be far more efficient

jlsherrill:
```
         host.installable_errata.pluck(&:errata_id)
```

app/models/katello/host/content_facet.rb
```diff
@@ -15,12 +15,14 @@ class ContentFacet < Katello::Model
 
       has_many :content_facet_errata, :class_name => "Katello::ContentFacetErratum", :dependent => :delete_all, :inverse_of => :content_facet
       has_many :applicable_errata, :through => :content_facet_errata, :class_name => "Katello::Erratum", :source => :erratum
+      has_many :installable_errata, :through => :content_facet_errata, :class_name => "Katello::Erratum", :source => :erratum
```
jlsherrill, line 18: I don't think this makes sense, as the errata in this association are 'applicable', but not necessarily installable.  You'd need to call the installable_errata' method below to calculate that.  Same comment for the installable_rpms relationship down below

ezr-ondrej:
@snagoor can you take a look and rebase, please?

PR #8897, opened 2020-08-19
Fixes #30480: Cleanup orphaned root repositories

ianballou:
There were a few cases before where (mysteriously) orphaned root repositories were causing some trouble.  I've updated the `correct_repositories` rake task to work with Pulp 3 and remove orphaned RootRepositories (nil `library_instance`).

To test:

1) Try `rake correct_repositories` on a machine using Pulp 3
2) Create a RootRepository that has a nil `library_instance`
3) Run `rake correct_repositories` and see that the RootRepository is deleted

### DIFF
--- a/lib/katello/tasks/repository.rake
+++ b/lib/katello/tasks/repository.rake
@@ -65,6 +65,10 @@ namespace :katello do
         handle_missing_repo(repo)
       end
     end
+
+    ::Katello::RootRepository.orphaned.each do |root_repo|
+      handle_missing_root_repo(root_repo)
+    end
   end
 
   desc "Correct missing pulp repositories for puppet environments. Specify CONTENT_VIEW=name and LIFECYCLE_ENVIRONMENT=name to narrow repositories.  COMMIT=true to perform operation."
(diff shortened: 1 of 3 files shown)


jturel:
@ianballou  - can you add some notes on how to verify this?

ianballou:
@jturel I've added the test steps.

lib/katello/tasks/repository.rake
```diff
@@ -65,6 +65,12 @@ namespace :katello do
         handle_missing_repo(repo)
       end
     end
+
+    ::Katello::RootRepository.find_each do |root_repo|
+      if root_repo.library_instance.nil? && root_repo.repositories.empty?
+        handle_missing_root_repo(root_repo)
+      end
+    end
```
jturel, line 73: I think the nil check on library_instance is unnecessary if also checking that the repos are empty, in which case you could find such root repos like this: `Katello::RootRepository.where.not(id: Katello::Repository.pluck(:root_id).uniq)`

You may even want to create an `orphaned` scope on RootRepository if that works - up to you.

lib/katello/tasks/repository.rake
```diff
@@ -152,6 +163,10 @@ namespace :katello do
     end
   end
 
+  def handle_missing_root_repo(root_repo)
+    root_repo.destroy!
```
jturel, line 167: Didn't check but I think the roots are being destroyed even when COMMIT is false

test/lib/tasks/repository_test.rb
```diff
@@ -103,12 +110,68 @@ def test_correct_repositories_missing_library_repo_commit
...
       Katello::Repository.stubs(:in_environment).returns(Katello::Repository.where(:id => @library_repo))
       Runcible::Extensions::Repository.any_instance.expects(:retrieve).once.with(@library_repo.pulp_id).raises(RestClient::ResourceNotFound)
+      SmartProxy.any_instance.stubs(:pulp3_support?).returns(false)
 
       ForemanTasks.expects(:sync_task).with(::Actions::Katello::Repository::Create, @library_repo)
 
       Rake.application.invoke_task('katello:correct_repositories')
     end
 
+    def test_correct_repositories_pulp3
```
jturel, line 120: Nice test coverage! :fire:

ianballou:
@jturel I think I've addressed your comments! I went with the `RootRepository` scope.

PR #8898, opened 2020-08-19
Fixes #30641 - CV publish fails if checksum type changes

ianballou:
Fixes untested erroneous code that was running if a user changes their repo's checksum type before publishing it in a content view.

To test:
1) Make sure you're using Pulp 3 for yum content
2) Create and sync a repo
3) Add that repo to a content view
4) Publish the content view
5) Change the repo's checksum type to sha1
6) Publish the content view again.  This will fail without my PR.

### DIFF
--- a/test/actions/katello/repository/multi_clone_contents_test.rb
+++ b/test/actions/katello/repository/multi_clone_contents_test.rb
@@ -1,0 +1,37 @@
+require 'katello_test_helper'
+
+module Actions
+  describe Katello::Repository::MultiCloneContents do
+    include Dynflow::Testing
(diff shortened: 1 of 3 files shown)


test/actions/katello/repository/multi_clone_contents_test.rb
```diff
@@ -0,0 +1,38 @@
+require 'katello_test_helper'
+require 'pry'
```
ianballou, line 2: Need to remove this

jturel:
@ianballou can you leave some notes on how to verify?

ianballou:
@jturel I've added test steps.

PR #8899, opened 2020-08-20
Fixes #30671 - Send UI Notification when Bastion table cannot load it…

jturel:
…s resource

Looks like a very old regression. The way we were handling errors from the table resource was confusing and seemingly wrong! A simple way to test is to go to the content hosts list and search 'errata_status = foo'. Previously there'd be an error about an unhandled promise in the console, but now it's cleaned up and sends a UI notification. Since this touches a critical area, I recommend doing any action that would refresh the table (such as paging, searching) and make sure everything looks good and no errors in console.

### DIFF
--- a/engines/bastion/app/assets/javascripts/bastion/components/nutupane.factory.js
+++ b/engines/bastion/app/assets/javascripts/bastion/components/nutupane.factory.js
@@ -97,7 +97,6 @@ angular.module('Bastion.components').factory('Nutupane',
             self.load = function () {
                 var deferred = $q.defer(),
                     responsePage,
-                    resourceCall,
                     table = self.table,
                     existingTable = TableCache.getTable(getTableName());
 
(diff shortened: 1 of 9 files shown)


engines/bastion/app/assets/javascripts/bastion/components/nutupane.factory.js
```diff
@@ -155,14 +150,14 @@ angular.module('Bastion.components').factory('Nutupane',
                     table.working = false;
                     table.refreshing = false;
                     table.initialLoad = false;
-                });
+                }).catch(function(response) {
+                    table.working = false;
+                    table.refreshing = false;
 
-                if (resourceCall && resourceCall.$promise && resourceCall.$promise.catch) {
```
jturel, line 160: I tried to send the Notification in this block but the 'unhandled promise' error was still being thrown in the browser console. use $promise.then.catch is the correct way to handle this, so I adjusted the code and fixed related test failures.

jturel:
@jeremylenz updated!

jeremylenz:
@jturel Changes look great, no more console error :)

Looks like some Angular tests still need adjusting, but ACK pending that.

jturel:
test fix pushed. APJ?

jeremylenz:
I was slower than J.  ACK (remember to squash ;)

PR #8900, opened 2020-08-20
Fixes #30488 - Make download_concurrency configurable

sjha4:
The current default download concurrency is 20 for pulp3. With this change, https://github.com/pulp/pulpcore/pull/832 the new default is 10. This Katello change exposes that as a field on remotes so we can configure it on katello repos if needed.
To test:
1. Create a repo
2. Check the remote in pulp 
curl --cert /etc/pki/katello/certs/pulp-client.crt --key /etc/pki/katello/private/pulp-client.key https://`hostname`/pulp/api/v3/remotes/rpm/rpm/ | python -m json.tool
3. Verify that the remote is created with the default value.
4. Open to suggestions on this part. To change the concurrency and propagate change to pulp, go to console.
```
repo = Katello::Repository.find(_id_) 
repo.root.update!(:download_concurrency=>15)
smart_proxy = SmartProxy.pulp_master
repo.backend_service(smart_proxy).update_remote

```
5. Re-sync the repo. Verify it succeeds with updated value.

### DIFF
--- a/app/controllers/katello/api/v2/repositories_controller.rb
+++ b/app/controllers/katello/api/v2/repositories_controller.rb
@@ -43,6 +43,7 @@ module Katello
       param :docker_upstream_name, String, :desc => N_("Name of the upstream docker repository")
       param :docker_tags_whitelist, Array, :desc => N_("Comma separated list of tags to sync for Container Image repository")
       param :download_policy, ["immediate", "on_demand", "background"], :desc => N_("download policy for yum repos (either 'immediate', 'on_demand', or 'background (deprecated)')")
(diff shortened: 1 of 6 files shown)


sjha4:
So I had a chat with Justin about this yesterday and he is OK with adding this to the API and have the download concurrency be updated from hammer.
Users can use that like below :
hammer -r repository update --download-concurrency 15 --id 1
Also, update --help should show them the help text for update:
Screenshot from 2020-08-25 09-58-52

Also, note that although this field would be relevant only for pulp3, the model still has the field on pulp2 repos and we are not adding any checks in hammer/api to stop users from updating this on pulp2 repos.

PR #8904, opened 2020-08-21
Refs #29926 - Add Simple Content Access to Org details

jeremylenz:
Continuing from #8890 

* Add Simple Content Access column to Org index
* Add Simple Content Access toggle switch to Org details
  - The change does not execute immediately; rather, it is deferred until you click Submit
  - Flash notification directs you to the Tasks page to monitor progress
* Add a UI (bell) notification when the SCA enable/disable is complete.  Bonus: this also works when toggling SCA from the Subscriptions page.


app/views/overrides/organizations/_edit_override.html.erb
```diff
@@ -1,18 +1,34 @@
...
+    isSimpleContentAccessEnabled: true,
+    )
+    %>
+  <% else %>
+    <%= react_component(
+    'SCAContainer',
+    isSimpleContentAccessEnabled: false,
+    )
+    %>
+  <% end %>
```
tbrisker, line 32:
```suggestion
    <%= react_component(
    'SCAContainer',
    isSimpleContentAccessEnabled: @taxonomy.simple_content_access?,
    )
    %>
```

app/controllers/katello/concerns/organizations_controller_extensions.rb
```diff
@@ -42,9 +43,22 @@ def create
         end
       end
 
+      def update
+        super
+        if params[:content_access_mode] == 'simple_content_access'
+          async_task(::Actions::Katello::Organization::SimpleContentAccess::Enable, params[:id])
+          info "Enabling Simple Content Access for organization #{@taxonomy.name}.", link: { text: "View progress on the Tasks page", href: "/foreman_tasks/tasks" }
```
jeremylenz, line 50: Users may enable/disable SCA for an organization that is not `Organization.current`.  Also, the Org details page is run with organization unscoped, so we don't even know what org is selected anyway.  For these reasons, it makes more sense to refer users to the Tasks page rather than the Manage Manifest modal.

app/lib/actions/katello/organization/simple_content_access/toggle.rb
```diff
@@ -24,10 +25,23 @@ def plan(organization_id)
...
+            task_success_notification.deliver!(
+              subject_organization
+            )
+          end
+
           private
 
           def consumer
-            @consumer ||= ::Katello::Resources::Candlepin::UpstreamCandlepinResource.upstream_consumer
+            @consumer ||= @organization.owner_details['upstreamConsumer']
```
jeremylenz, line 44: had to change this to enable dealing with an org that is not `Organization.current`

webpack/scenes/Subscriptions/Manifest/SCAContainer.js
```diff
@@ -0,0 +1,61 @@
...
+import SimpleContentAccess from './SimpleContentAccess';
+
+const SCAContainer = (props) => {
+  const {
+    isSimpleContentAccessEnabled,
+  } = props;
+  const [enabled, setEnabled] = useState(isSimpleContentAccessEnabled);
+  const hiddenRailsInput = useRef(document.querySelector('#content_access_mode'));
+
+  const infoMsgOverride = enabledVal => (enabledVal !== isSimpleContentAccessEnabled) || undefined;
```
jeremylenz, line 13: I should probably add comments here, but a note about how this works:
* The `isSimpleContentAccessEnabled` prop is set by the Rails view in `_edit_override.html.erb` and sets the initial state of the toggle switch.
* The Rails view also creates a hidden `<input>` element which is accessed here via `useRef`.  Its value is kept in sync with the state of the switch.  This is so it can eventually end up as `params[:content_access_mode]` when the form is submitted.
* `infoMsgOverride` is a function that decides what the text below the toggle should be.  If the user is about to change the content access mode, it displays the informative messages below; if not, the `undefined` value triggers the defaultProp in `<SimpleContentAccess />`.

webpack/scenes/Subscriptions/Manifest/SCAContainer.js
```diff
@@ -0,0 +1,61 @@
...
+import { translate as __ } from 'foremanReact/common/I18n';
+import PropTypes from 'prop-types';
+import SimpleContentAccess from './SimpleContentAccess';
+
+const SCAContainer = (props) => {
+  const {
+    isSimpleContentAccessEnabled,
+  } = props;
+  const [enabled, setEnabled] = useState(isSimpleContentAccessEnabled);
+  const hiddenRailsInput = useRef(document.querySelector('#content_access_mode'));
```
jturel, line 11: Could this possibly be given as a prop, or should it? In the context of this container I don't think we care about Rails or whether it's hidden so I'd recommend a name along the lines of 'scaFormInput'

app/lib/actions/katello/organization/simple_content_access/toggle.rb
jturel: Nice. I really dislike the areas of our code that rely on Org.current

app/controllers/katello/concerns/organizations_controller_extensions.rb
jturel, line 50: Did you mean to link to the specific task vs the tasks list?

app/controllers/katello/concerns/organizations_controller_extensions.rb
```diff
@@ -42,9 +43,22 @@ def create
...
+        end
+        if params[:content_access_mode] == 'entitlement'
+          async_task(::Actions::Katello::Organization::SimpleContentAccess::Disable, params[:id])
+          info "Disabling Simple Content Access for organization #{@taxonomy.name}.", link: { text: "View progress on the Tasks page", href: "/foreman_tasks/tasks" }
```
jturel, line 54: Ditto my question on the link to the tasks page

webpack/scenes/Subscriptions/Manifest/SCAContainer.js
```diff
@@ -0,0 +1,61 @@
...
+
+  const infoMsgOverride = enabledVal => (enabledVal !== isSimpleContentAccessEnabled) || undefined;
+  const getContentAccessModeValue = (newValue) => {
```
jturel, line 14: Observations
- I see that `infoMsgOVerride` changes the messaging on screen. It seems a little strange to mention that there's a change pending. We don't do that for any of the other form inputs, why this one? IMO it just adds more complexity in this component.
- We should avoid having the UI layer know about 'content access mode' and the related values. That really is a Candlepin concept and should be kept on the backend

Suggestions:
- rename the hidden input field to 'simple_content_access' and have its default value be @org.simple_content_access?
- implies the hidden field is now only true/false
- on the backend, if we **detect a change** between the current @org.simple_content_access? and the form value, perform the right action

webpack/scenes/Subscriptions/Manifest/SCAContainer.js
jeremylenz: I don't think a DOM element could be given as a prop since it's starting out in Ruby, but the name change makes sense to me.  Will update

app/controllers/katello/concerns/organizations_controller_extensions.rb
jeremylenz: It would be nice.  I didn't really look into if I have access to the task ID at that point.  Will investigate & update

webpack/scenes/Subscriptions/Manifest/SCAContainer.js
jeremylenz: > We don't do that for any of the other form inputs, why this one? 

There were 2 ways I was thinking about consistency here: (1) the form element should behave the same way as all other form elements on the page (we already talked about this one), and (2) this same toggle is used both here and in the Manage Manifest modal on the Subscriptions page.  On the Subscriptions page, the change executes immediately, so users might also expect it to execute immediately here.  That's why I added the `infoMsgOverride` dynamic text.

>  the hidden field is now only true/false

The `value` attribute of the [`<input>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/hidden) HTML element is a [`DOMString`](https://developer.mozilla.org/en-US/docs/Web/API/DOMString), so I don't think trying to store a boolean value there would work well.

That said, I don't have any strong feelings about the name or text value of the attribute.  We just need to read the string from that value attribute and send it as a param to Rails.  If we want to call it `simple_content_access` and then have the value be something like `enable` or `disable`, I'd be fine with that.

> on the backend, if we detect a change between the current @org.simple_content_access? and the form value, perform the right action

It's doing this right now, but on the front end.  If we were to move this logic to the back end, it would remove the ability to have the `infoMsgOverride`.

jeremylenz: I just realized that using a switch here goes against [this Patternfly guideline](https://www.patternfly.org/v4/documentation/core/components/switch#overview):

_Use a switch when your user needs to perform instantaneous actions without confirmation._

_Use checkbox if your user has to perform additional steps for changes to become effective._

Going to look into changing this to a dumb checkbox.  This may remove the need for a React component entirely.

jturel: Using a checkbox really would make this a lot simpler considering the patternfly guidance around when to use the switch and how it should behave (real time, instantaneous action triggering). +1 from me.

jturel: > It's doing this right now, but on the front end. If we were to move this logic to the back end, it would remove the ability to have the `infoMsgOverride`.

I perceive it this way: your impl on the front end is to enable dynamic messaging, but it has a side-effect of not triggering a manifest action on the backend if the user didn't touch the toggle since it will send 'unchanged' as the value. This tightly couples the front and back end. If someone comes along later and changes the form (intentionally or not) it would be easy for the `update` method to start triggering the SCA toggle action on each `update`. No matter what the front end is doing, it's safer to have that defensive check in the controller to prevent future bugs.

app/controllers/katello/concerns/organizations_controller_extensions.rb
jeremylenz: Updated

jeremylenz:
@jturel updated:
* Switched to using a checkbox as recommended by Patternfly
* added link to specific task page

webpack/scenes/Subscriptions/Manifest/SimpleContentAccess.js
```diff
@@ -11,6 +11,8 @@ const SimpleContentAccess = (props) => {
     isSimpleContentAccessEnabled,
     enableSimpleContentAccess,
     disableSimpleContentAccess,
+    infoMessage,
```
jturel, line 14: i guess all changes here can be reverted?

app/controllers/katello/concerns/organizations_controller_extensions.rb
```diff
@@ -42,9 +43,36 @@ def create
         end
       end
 
+      def update
```
jturel, line 46: would love a unit test for this :)

app/views/overrides/organizations/_edit_override.html.erb
```diff
@@ -8,11 +8,19 @@
...
+
   <%= field(f, _('Debug Certificate'),
             :help_inline => _('This certificate allows a user to view the repositories in any environment from a browser.')) do
     button_tag _('Generate and Download'), :id => :download_debug_cert_key,
                :type => :button, :style => 'margin-top: 8px;', :class => 'btn btn-small',
                'data-url' => "/katello/api/v2/organizations/#{@taxonomy.id}/download_debug_certificate"
   end %>
 
-<% end%>
+  <%= field(f, _('Simple Content Access'),
```
jturel, line 19: The toggle is showing up for orgs that don't have a manifest imported - that should be prevented

app/controllers/katello/concerns/organizations_controller_extensions.rb
```diff
@@ -42,9 +43,36 @@ def create
...
+        super
```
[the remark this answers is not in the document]
jeremylenz: If I put `super` at the bottom then you get the flash notifications in this order:
```
Disabling Simple Content Access for Org B
Successfully updated Org B
```
I think that's confusing because it implies that the SCA toggle was already successful.

app/views/overrides/organizations/_edit_override.html.erb
```diff
@@ -8,11 +8,19 @@
...
+            :help_inline => _('Toggling Simple Content Access will refresh your manifest.')) do
+    tag1 = hidden_field_tag 'simple_content_access', '0'
```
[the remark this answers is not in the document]
jeremylenz: Yeah, Rails does it for you if you use the form helpers.  (see ["Gotcha"](https://api.rubyonrails.org/v6.0.3.2/classes/ActionView/Helpers/FormHelper.html#method-i-check_box)) But since it's not an attribute of `Organization` that we're editing I just decided to do it myself :)

app/controllers/katello/concerns/organizations_controller_extensions.rb
```diff
@@ -42,9 +43,36 @@ def create
...
+        return if @taxonomy.simple_content_access? == sca_param
```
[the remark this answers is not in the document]
jeremylenz: The explicit true/false check is on purpose, due to the checkbox "gotcha" I mentioned above.  I didn't want a future change to the ERB resulting in no `simple_content_access` param being sent, and then being mistaken for "please turn SCA off".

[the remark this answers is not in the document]
jeremylenz: I didn't know about `Foreman::Cast.to_bool`!  Our helpers can be a double-edged sword but sometimes they're super useful.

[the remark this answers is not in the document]
jturel: > The explicit true/false check was on purpose, due to the checkbox "gotcha" I mentioned above. I didn't want a future change to the ERB resulting in no `simple_content_access` param being sent, and then being mistaken for "please turn SCA off".

Yep. My example avoids that same problem in a more obvious way (imo) ;)

webpack/scenes/Subscriptions/Manifest/SimpleContentAccess.js
jeremylenz: I kept the changes in `SimpleContentAccess.js` (namely, adding the `colWidths` and `infoMessage` props) because I still think they make the component more flexible and reusable.  (also I think I'd have to revert it manually :confused:)   would you be open to keeping them?

jturel: They do add that to the component, but I think there's a very high chance we'll never benefit from those changes so I'd prefer to see them removed. However, I'll defer to you!

If you did want to restore the file: `git reset upstream/master webpack/scenes/Subscriptions/Manifest/SimpleContentAccess.js`

app/views/overrides/organizations/_edit_override.html.erb
```diff
@@ -8,11 +8,21 @@
...
+
   <%= field(f, _('Debug Certificate'),
             :help_inline => _('This certificate allows a user to view the repositories in any environment from a browser.')) do
     button_tag _('Generate and Download'), :id => :download_debug_cert_key,
                :type => :button, :style => 'margin-top: 8px;', :class => 'btn btn-small',
                'data-url' => "/katello/api/v2/organizations/#{@taxonomy.id}/download_debug_certificate"
   end %>
 
-<% end%>
+  <% if @taxonomy.manifest_imported? %>
```
jturel, line 19: Thinking about this a bit more. On the manifest modal manifest actions like this are disabled if the upstream subscriptions controller 'ping' api fails. it will fail under circumstances like:

- no manifest imported
- the system is in 'disconnected' mode
- the manifest is expired

If we can be consistent that is ideal. A way to achieve this is to override the `edit` method (assuming that's what it's called) in the same controller you overrode `update` and provide a new instance variable for the view `@can_toggle_sca` which will be true/false determined by code similar to here https://github.com/Katello/katello/blob/master/app/controllers/katello/api/v2/api_controller.rb#L215-L224

app/views/overrides/organizations/_edit_override.html.erb
```diff
@@ -8,11 +8,21 @@
     <%= selectable_f(f, :service_level, options_for_select(service_level_options, service_level_selected), {}, :label => _('Default System SLA')) %>
   <% end %>
 
+
```
jturel, line 11: Big fan of whitespace here but this one seems unnecessary!

test/controllers/foreman/organizations_controller_test.rb
```diff
@@ -29,4 +29,30 @@ def test_use_newly_created_org
...
     assert_equal session[:organization_id], org.id
   end
+
+  def test_simple_content_access_toggle
+    # use org with SCA enabled
+    org = Organization.first
+    Organization.any_instance.stubs(:simple_content_access?).returns true
+    TaxHost.any_instance.stubs(:check_for_orphans).returns false
+    # assert correct task will be started
+    @controller.expects(:async_task).with do |action_class, _org|
```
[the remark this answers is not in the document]
jeremylenz: I thought so too but it didn't seem work without the block!  I'll try it again

test/controllers/foreman/organizations_controller_test.rb
```diff
@@ -29,4 +29,30 @@ def test_use_newly_created_org
...
+    end
+    # send request to disable SCA
+    put :update, params: { id: org.id, simple_content_access: false }
+  end
+
+  def test_simple_content_access_unchanged
+    # use org with SCA enabled
+    org = Organization.first
+    Organization.any_instance.stubs(:simple_content_access?).returns true
+    TaxHost.any_instance.stubs(:check_for_orphans).returns false
```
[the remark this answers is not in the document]
jeremylenz: Still gets the VCR error :sob:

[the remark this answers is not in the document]
jturel: booo. just checked foreman and they are using organization2 for their update test with no such stub. mind giving that a try? if it doesn't work let's not worry about it.

https://github.com/theforeman/foreman/blob/develop/test/controllers/organizations_controller_test.rb#L37

test/controllers/foreman/organizations_controller_test.rb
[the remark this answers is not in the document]
jeremylenz: I had to convert `org.id` to a string. better now :)

jeremylenz:
@jturel updated, thanks!

test/controllers/foreman/organizations_controller_test.rb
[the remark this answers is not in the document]
jeremylenz: `organization2` works!!

https://github.com/theforeman/foreman/blob/develop/test/fixtures/taxonomies.yml#L17

No idea why.  I don't see anything different about the fixtures.

app/controllers/katello/concerns/organizations_controller_extensions.rb
```diff
@@ -42,9 +48,27 @@ def create
         end
       end
 
+      def update
+        super
```
jturel, line 52: Mind adding a comment here about the positioning of super for the flash message ordering, to make it clear?

app/controllers/katello/concerns/organizations_controller_extensions.rb
```diff
@@ -3,8 +3,14 @@ module Concerns
     module OrganizationsControllerExtensions
       extend ActiveSupport::Concern
       include ForemanTasks::Triggers
+      include Foreman::Controller::Flash
 
       module Overrides
+        def edit
```
jturel, line 9: Can you add a test for this method? Since we're overriding foreman behavior it's a critical place to test!

jeremylenz:
@jturel Updated / added 2 tests and one assertion helper method.   let me know what you think :)

test/controllers/foreman/organizations_controller_test.rb
```diff
@@ -29,4 +29,54 @@ def test_use_newly_created_org
...
+  end
+
+  def test_edit_override_can_toggle
+    org = get_organization(:organization2)
+    Organization.any_instance.stubs(:service_level)
+    Organization.any_instance.stubs(:service_levels).returns []
+    Katello::UpstreamConnectionChecker.any_instance.expects(:can_connect?).returns true
+    Organization.any_instance.expects(:simple_content_access?).returns true
+    get :edit, params: { id: org.id }
+    assert_instance_variable_set :can_toggle_sca, true
```
jturel, line 70: Because you have an expectation that the org received `simple_content_access?` this assertion is redundant since that method wouldn't have been called were `@can_toggle_sca` false. IMO, remove this assertion and add an another checking that the response was successful.

Also, checking instance variables is unusual. Better to test the response body to see how it was affected, but I don't think that's necessary here because the `simple_content_access` assertion does the trick  :)

jeremylenz: Checking the response body feels a bit like testing the view to me, but this is `controller_test` so we should be testing the controller.

Also I realized that `@response.body` won't contain what we would expect anyway, because our overrides are inserted using Deface after the response has already been sent.

I decided on testing the instance variable since really the test should only be testing the code in the `edit` method, and setting the variable is all that code really does.   In light of this I feel like it makes sense to keep this assertion.  Thoughts?

jturel: Testing how that instance variable influences the response is what's important and your expectations are already set up perfectly by asserting that `simple_content_access?` was called as a result of `@can_toggle_sca` being true

test/controllers/foreman/organizations_controller_test.rb
```diff
@@ -29,4 +29,54 @@ def test_use_newly_created_org
...
+  end
+
+  def test_edit_override_no_manifest_imported
+    org = get_organization(:organization2)
+    Katello::UpstreamConnectionChecker.any_instance.stubs(:assert_can_upstream_ping).raises(Katello::Errors::NoManifestImported)
```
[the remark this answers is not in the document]
jeremylenz: @jturel Ok, removed!

test/controllers/foreman/organizations_controller_test.rb
```diff
@@ -29,4 +29,44 @@ def test_use_newly_created_org
...
+    put :update, params: { id: org.id, simple_content_access: true }
+  end
+
+  def test_edit_override_can_toggle
+    org = get_organization(:organization2)
+    Organization.any_instance.stubs(:service_level)
+    Organization.any_instance.stubs(:service_levels).returns []
+    Katello::UpstreamConnectionChecker.any_instance.expects(:can_connect?).returns true
+    Organization.any_instance.expects(:simple_content_access?).returns true
+    get :edit, params: { id: org.id }
```
jturel, line 69: Mind adding the success response to these 4 tests? Sorry I missed that

jeremylenz: @jturel done, `update` actually responds with 302 Found so I asserted accordingly.

jeremylenz:
test failures unrelated

jturel:
[test katello]

jturel:
[test katello]

PR #8906, opened 2020-08-24
Fixes #30702 - Regression in re-assigning host location for registered hosts

jturel:
We recently made a change which prevented changing host Organization if it is registered in Katello. It inadvertently broke the ability to change the Location of a registered Content Host.

This change:
- restores ability to change location of registered content hosts
- prevents error from being raised when "changing" host into its current org (think about Any Org + Any Loc use case)
- avoid N+1 query in loading host subscription facets

To test: Try to break stuff while changing Org or Loc from Hosts -> All Hosts

### DIFF
--- a/test/controllers/foreman/hosts_controller_test.rb
+++ b/test/controllers/foreman/hosts_controller_test.rb
@@ -8,6 +8,7 @@ class HostsControllerTest < ActionController::TestCase
   def models
     @library = katello_environments(:library)
     @library_dev_staging_view = katello_content_views(:library_dev_staging_view)
+    @host = hosts(:one)
   end
 
   def setup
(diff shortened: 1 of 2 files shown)


test/controllers/foreman/hosts_controller_test.rb
```diff
@@ -58,17 +80,29 @@ def setup
...
-      assert_equal "attachment; filename=\"hosts-#{Date.today}.csv\"",
-        response.headers["Content-Disposition"]
-      buf = response.stream.instance_variable_get(:@buf)
-      assert buf.is_a? Enumerator
-      assert_equal "Name,Subscription Status,Installable Updates - Security,Installable Updates - Bug \
-Fixes,Installable Updates - Enhancements,Installable Updates - Package Count,OS,Environment,\
-Content View,Registered,Last Checkin\n",
-        buf.next
-      assert_equal 3, buf.count
+      assert_equal "attachment; filename=\"hosts-#{Date.today}.csv\"", response.headers["Content-Disposition"]
```
jturel, line 86: this test started failing when i was messing with the other tests. i figured it could use some tidying up!

PR #8908, opened 2020-08-26
Fixes #30703 - allow CV version to be used in templates

ares:
This method is necessary for enabling content view version be listed for
a host in the report templates. With this, it's possible to get the
version like

host.content_view.version(host.lifecycle_environment).to_s

### DIFF
--- a/app/models/katello/content_view.rb
+++ b/app/models/katello/content_view.rb
@@ -241,6 +241,10 @@ module Katello
       environments.include?(env)
     end
 
+    apipie :method, 'Returns the Katello::ContentViewVersion for a given Lifecycle Environment' do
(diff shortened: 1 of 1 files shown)


ares:
This is the most straight forward way, however we could also add a method to the host object, so we'd only call `host.content_view_version` and enable that in the safe mode. It would automatically pass its LC env and handle the case there's no env or content view associated.

Let me know what you think, it shouldn't be too hard to do. I was unable to find such method in katello repo, but if there's some already, please let me know.

jlsherrill:
should we add docs here: https://github.com/Katello/katello/blob/master/app/models/katello/content_view.rb#L728

ares:
Yes, thanks, this patch predates the documentation but added now.

ares:
imho react tests failure is unrelated, @jlsherrill could you take a quick look and merge pls?

jlsherrill (approved):

jlsherrill:
thx for the reminder!  and for the pr!

PR #8910, opened 2020-08-26
Fixes #30322 - Clean orphan content upon repo deletion

sjha4:


app/models/katello/concerns/pulp_database_unit.rb
```diff
@@ -211,6 +211,10 @@ def db_values_copy(source_repo, dest_repo)
       def db_values(new_ids, pulp_id_href_map, repository)
         new_ids.map { |unit_id| [unit_id.to_i, pulp_id_href_map.dig(unit_id), repository.id.to_i, Time.now.utc.to_s(:db), Time.now.utc.to_s(:db)].compact }
       end
+
+      def clean_orphans
+        (self.where("id not in (?)", repository_association_class.pluck(unit_id_field.to_sym))).destroy_all if many_repository_associations && repository_association_class
+      end
     end
```
sjha4, line 218: Note needed..will remove after 1 jenkins run.

sjha4:
[test katello]

app/models/katello/docker_meta_tag.rb
```diff
@@ -131,6 +131,7 @@ def self.with_identifiers(ids)
 
     def self.cleanup_tags
       self.where(:schema2_id => nil, :schema1_id => nil).delete_all
+      self.where("id not in (?) OR (schema2_id IS NULL AND schema1_id IS NULL)", Katello::RepositoryDockerMetaTag.pluck(:docker_meta_tag_id)).delete_all
```
jturel, line 134: I think the second half of this query is a duplicate of the line above: schema2_id IS NULL AND schema1_id IS NULL

jturel, line 134: Since we need to do this cleanup against RepositoryDockerMetaTag doesn't it indicate that we don't have our associations configured properly?

sjha4: We don't have dependent deletes on content units from the Repository content unit join models currently. Mostly because they are many to many and also we don't delete content units in katello unless the unit gets deleted in pulp. That deletion happens as part of orphan cleanup task.

sjha4: Oh I did mean to remove the above query..let me do that..

sjha4: Updated..

app/lib/actions/katello/repository/destroy.rb
```diff
@@ -52,9 +53,10 @@ def handle_redhat_content(repository)
           end
         end
 
-        def delete_record(repository)
+        def delete_record(repository, options = {})
           repository.destroy!
           repository.root.destroy! if repository.root.repositories.empty?
+          ::Katello::DockerMetaTag.cleanup_tags if options[:docker_cleanup]
```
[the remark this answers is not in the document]
sjha4: It would add the DockerMetaTag cleanup method to flow of to all repo deletes..might get confusing when the query appears in logs when deleting other types of repos?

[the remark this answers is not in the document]
jturel: True. I could go either way :)

app/lib/actions/katello/repository/destroy.rb
```diff
@@ -36,7 +36,8 @@ def plan(repository, options = {})
 
         def finalize
           repository = ::Katello::Repository.find(input[:repository][:id])
-          delete_record(repository)
+          docker_cleanup = repository.content_type == "docker"
```
jturel, line 39: Is there any advantage of passing this new options hash through rather than just checking the repo content_type in `delete_record` ?

jturel, line 39: Extra credit for changing this to `Katello::Repository::DOCKER_TYPE` :)

sjha4: I wanted to have the check as an option for the method. Although we are not using this method anywhere else so we can do it either way. I will just take your suggestion.. Hash or not? :smile:

jturel: Let's leave it! was curious if there was a specific purpose though

PR #8912, opened 2020-08-28
Fixes #30736 - refactor cert guard updating

jlsherrill:
previously when updating the CDN url, a task was triggered
to update each repository and as part of that to update cert guard
data.  Due to a small logic bug, this was happening even when it should
not have been.  But in addition to this, the place that it was triggered
could lead to tasks being left in pending state.

This refactor moves UpdateCvRepoCertGuard into a orchestration action,
which will optionally get called if the repo is pulp3.  This means that under
pulp2, a repo will not have this action triggered for it at all saving some
computation.


jlsherrill:
To test:

1.  under pulp2 and pulp3, enable a RH repo and then update the cdn url.  You should not see any additional cert guard tasks get spawned.

2.  Under pulp3, change 'publish via http' to true or false, and verify that you can or can't access it via the http:// url to the repo when you do this.

app/lib/actions/katello/repository/update.rb
```diff
@@ -26,28 +26,23 @@ def plan(root, repo_params)
...
-                             SmartProxy.pulp_master)
-            plan_self(:repository_id => root.library_instance.id, :update_cv_cert_protected => update_cv_cert_protected)
+            sequence do
+              plan_pulp_action([::Actions::Pulp::Orchestration::Repository::Refresh,
+                                ::Actions::Pulp3::Orchestration::Repository::Update],
+                               repository,
+                               SmartProxy.pulp_master)
+              plan_self(:repository_id => root.library_instance.id, :update_cv_cert_protected => update_cv_cert_protected)
+              plan_optional_pulp_action([::Actions::Pulp3::Orchestration::Repository::TriggerUpdateRepoCertGuard], repository, ::SmartProxy.pulp_master)
+            end
```
sjha4, line 36: Should we condition this action TriggerUpdateRepoCertGuard with `update_cv_cert_protected`?

jlsherrill: doh! yes good catch!

jlsherrill: updated

sjha4:
[test katello]
Code looks good..testing now..

jlsherrill:
[test katello]    master fix was just merged

PR #8913, opened 2020-08-31
Fixes #30079 - workaround for ISE when search terms are used 

jjeffers:
This PR is a workaround, not the ultimate resolution, for an observed defect in the scoped_search gem. See [this issue](https://github.com/wvanbergen/scoped_search/issues/200) for details.

Once that issue is resolved, then the complete_value can be restored.


chris1984:
[test katello]

PR #8916, opened 2020-08-31
Fixes #30749 - Add /owners/:id/system_purpose to Candlepin Proxies Controller

jturel:
New versions of subscription-manager attempt to retrieve the org's system purpose values from Katello, and we need to add that endpoint to our allowlist since we proxy those requests through the candlepin.

Can be tested from a client like this:
```
curl https://centos7-katello-devel.jturel.example.com/rhsm/owners/Default_Organization/system_purpose --cert /etc/pki/consumer/cert.pem --key /etc/pki/consumer/key.pem
```

It should respond with something like:

```
{"owner":{"id":"4028fac873da183e0173da1da6e20000","key":"Default_Organization","displayName":"Default Organization","href":"/owners/Default_Organization"},"systemPurposeAttributes":{"addons":[],"roles":["","Red Hat Enterprise Linux Server"],"usage":["Development/Test","Production"],"support_level":["Standard","Self-Support"]}}
```

Previously it would be a 404!

### DIFF
--- a/app/controllers/katello/api/rhsm/candlepin_proxies_controller.rb
+++ b/app/controllers/katello/api/rhsm/candlepin_proxies_controller.rb
@@ -26,7 +26,7 @@ module Katello
     before_action :set_organization_id, :except => [:hypervisors_update, :async_hypervisors_update]
     before_action :find_hypervisor_organization, :only => [:async_hypervisors_update, :hypervisors_update]
 
-    before_action :check_content_type, :except => :async_hypervisors_update
+    before_action :check_media_type, :except => :async_hypervisors_update
 
     prepend_before_action :convert_owner_to_organization_id, :except => [:hypervisors_update, :async_hypervisors_update], :if => lambda { params.key?(:owner) }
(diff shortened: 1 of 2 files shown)


app/controllers/katello/api/rhsm/candlepin_proxies_controller.rb
```diff
@@ -26,7 +26,7 @@ class Api::Rhsm::CandlepinProxiesController < Api::V2::ApiController
     before_action :set_organization_id, :except => [:hypervisors_update, :async_hypervisors_update]
     before_action :find_hypervisor_organization, :only => [:async_hypervisors_update, :hypervisors_update]
 
-    before_action :check_content_type, :except => :async_hypervisors_update
+    before_action :check_media_type, :except => :async_hypervisors_update
```
jturel, line 29: There was a deprecation warning from Foreman, so I fixed that too.

```
18:05:00 rails.1   | 2020-08-31T18:05:00 [W|app|c7dbac04] DEPRECATION WARNING: You are using a deprecated behavior, it will be removed in version 2.3, #check_content_type is renamed to #check_media_type (called from repackage_message at /home/vagrant/katello/app/controllers/katello/api/rhsm/candlepin_proxies_controller.rb:35)
```

PR #8919, opened 2020-09-02
Fixes #30758 - Refresh Subscription and Purpose status when toggling SCA

jturel:
Builds on recent work by @chris1984 (https://github.com/Katello/katello/pull/8859) to process the `owner_content_access_mode.modified` event (turns out there is one!) to:
- consolidate places where we might clear syspurpose status (any manifest action) into a single event handler
- in the same event handler, handle case of switching out of SCA mode, and recalculate host purpose + subscription status

Installer PR to prevent filtering out this new event: https://github.com/theforeman/puppet-candlepin/pull/164
You can modify /etc/candlepin/broker.xml and add this filter `(EVENT_TARGET='OWNER_CONTENT_ACCESS_MODE' and EVENT_TYPE='MODIFIED')` be sure to restart tomcat afterward

**To Test**
 - register a content host
 - import a manifest that can toggle SCA
 - observe the host sub status switch from $WHATEVER to 'Simple Content Access' and vice versa
- observe  the host purpose status switch from $WHATEVER to hidden, and vice versa

### DIFF
--- a/app/services/katello/candlepin/message_handler.rb
+++ b/app/services/katello/candlepin/message_handler.rb
@@ -91,8 +91,17 @@ module Katello
       end
 
       def handle_content_access_mode_modified
-        org_label = Katello::Util::Model.labelize(target_name)
-        org = ::Organization.find_by!(label: org_label)
+        # Ideally the target_name would be the Candlepin Owner key
+        # Since it's the displayName, and we don't update that after org creation, there could be a mismatch
+        # For now, find the Candlepin Owner displayName from this event, and tie it back to a Katello org based on owner key
+        owners = Katello::Resources::Candlepin::Owner.all
+        owner = owners.find { |o| o['displayName'] == target_name }
+
+        unless owner
+          fail("Candlepin Owner %s could not be found" % target_name)
+        end
+
+        org = ::Organization.find_by!(label: owner['key'])
         hosts = org.hosts
 
         if event_data['contentAccessMode'] == 'org_environment'
@@ -104,7 +113,7 @@ module Katello
             Katello::Resources::Candlepin::Consumer.compliance(uuid)
             Katello::Resources::Candlepin::Consumer.purpose_compliance(uuid)
           rescue => e
-            Rails.logger.error("Error encountered while fetching consumer compliance for #{uuid}: #{e.message}")
+            Rails.logger.error("Error encountered while fetching compliance for consumer #{uuid}: #{e.message}")
           end
         end
 
--- a/app/services/katello/candlepin/message_handler.rb
+++ b/app/services/katello/candlepin/message_handler.rb
@@ -20,6 +20,10 @@ module Katello
         @event_data ||= (data = content['eventData']) ? JSON.parse(data) : {}
       end
 
+      def target_name
+        content['targetName']
+      end
+
       def status
         event_data['status']
       end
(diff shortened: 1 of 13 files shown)


jturel:
@jeremylenz updated to handle your case when the org name has changed. extremely good catch! I'm going to file a bug against Candlepin so we get the "owner key" as part of the event. That will match our org label always since it cannot be changed. We can roll with this workaround for now though.

test/services/candlepin/message_handler_test.rb
```diff
@@ -145,4 +124,43 @@ def test_delete_pool_on_null
...
+      @handler.handle_content_access_mode_modified
+    end
+
+    def test_sca_disabled
+      Katello::HostStatusManager.expects(:clear_syspurpose_status).never
+      Katello::HostStatusManager.expects(:update_subscription_status_to_sca).never
+      Organization.any_instance.expects(:simple_content_access?).with(cached: false)
+      @handler.expects(:event_data).returns('contentAccessMode' => 'entitlement').twice
+
+      @org.hosts.joins(:subscription_facet).count.times do
```
jeremylenz, line 158: Can this number change for each test run?  If not, should probably be hard-coded and avoid the DB query

test/services/candlepin/message_handler_test.rb
```diff
@@ -84,11 +72,8 @@ def test_create_pool_on_host
     end
   end
 
-  class EntitlementDeleted < MessageHandlerTestBase
-    def setup
-      super
-      @handler = load_handler('entitlement.deleted')
-    end
+  class EntitlementDeletedTest < MessageHandlerTestBase
+    let(:event_name) { 'entitlement.deleted' }
```
jeremylenz, line 76: thanks for cleaning up the tests with `let` :smile:

app/services/katello/candlepin/message_handler.rb
```diff
@@ -85,6 +89,36 @@ def delete_pool
...
+          cp_consumer_uuids = hosts.joins(:subscription_facet).pluck("#{Katello::Host::SubscriptionFacet.table_name}.uuid")
+          cp_consumer_uuids.each do |uuid|
+            Katello::Resources::Candlepin::Consumer.compliance(uuid)
+            Katello::Resources::Candlepin::Consumer.purpose_compliance(uuid)
+          rescue => e
+            Rails.logger.error("Error encountered while fetching compliance for consumer #{uuid}: #{e.message}")
+          end
+        end
+
+        org.simple_content_access?(cached: false)
```
jeremylenz, line 120: Seems like this return value isn't used anywhere; is it necessary?

jeremylenz (commented):
Thanks @jturel, works as expected now!

* Used "Org B" which had the naming problem before
* Disabled SCA - my host showed "Fully entitled" immediately
* Re-enabled SCA - my host showed "Simple Content Access" immediately

Just a couple questions/comments below

app/services/katello/candlepin/message_handler.rb
jturel: Here I'm refreshing our cached value for `simple_content_access?` so that any request which reads it 'soon' (the cache has a 1 min lifetime) will be accurate. We cache it because it's a frequent call that we make, and saves requests into Candlepin. Just a coincidence that it's the last line of the method!

test/services/candlepin/message_handler_test.rb
jturel: It could change since this organization is one of our fixtures

jeremylenz (approved):
Thanks for the answers

ACK

PR #8922, opened 2020-09-02
Fixes #30766 - Candlepin error with repo deletion during product destroy

ianballou:
Candlepin throws an error when multiple [remove_content](https://github.com/Katello/katello/blob/master/app/lib/actions/candlepin/product/content_remove.rb) calls happen concurrently.  To get around this, I've made the `Actions::Katello::Product::ContentDestroy` actions run sequentially before the concurrent `Actions::Katello::Repository::Destroy` actions.

To test:

1) Create a product
2) Add a few Docker repos
3) Delete the product
4) See the error in question
5) Apply my PR
6) Do 1-3 again
7) See that the product is deleted

### DIFF
--- a/app/lib/actions/katello/repository/destroy.rb
+++ b/app/lib/actions/katello/repository/destroy.rb
@@ -7,9 +7,11 @@ module Actions
 
         # options:
         #   skip_environment_update - defaults to false. skips updating the CP environment
+        #   destroy_content - can be disabled to skip Candlepin remove_content
         def plan(repository, options = {})
           skip_environment_update = options.fetch(:skip_environment_update, false) ||
               options.fetch(:organization_destroy, false)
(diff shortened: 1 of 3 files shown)


app/lib/actions/katello/repository/destroy.rb
```diff
@@ -29,7 +31,9 @@ def plan(repository, options = {})
             if repository.redhat?
               handle_redhat_content(repository) unless skip_environment_update
             else
-              handle_custom_content(repository) unless skip_environment_update
+              if destroy_content
+                handle_custom_content(repository) unless skip_environment_update
```
jlsherrill, line 35: we should probably combine those statemetns:

if destroy_content && !skip_environment_update

jlsherrill (approved):

PR #8923, opened 2020-09-02
Fixes #30280 - replaces 'master' with 'primary'

jjeffers:


jeremylenz (commented):
I think we should also address  https://github.com/Katello/katello/blob/9dc14ebc5e8d5351b15e1d9c8bb82855b7964bbd/app/models/katello/rhsm_fact_parser.rb#L116

jjeffers:
[test katello]

jturel:
```
test/controllers/api/v2/content_view_versions_controller_test.rb:      SmartProxy.stubs(:pulp_master).returns(FactoryBot.create(:smart_proxy, :default_smart_proxy, :with_pulp3))
```

Found this last occurrence of pulp_master - but other than that looks good

jeremylenz (approved):
Agree with @jturel on the version.

clicked around a bit and seems not to break anything!  ACK

PR #8928, opened 2020-09-04
Fixes #30752 - support sles token via url param

jlsherrill:
this extends the current support for sles tokens (by putting a param
at the end like  http://url/path?token) by parsing that token and
passing it to pulp3

### DIFF
--- a/app/services/katello/pulp3/repository/yum.rb
+++ b/app/services/katello/pulp3/repository/yum.rb
@@ -8,11 +8,17 @@ module Katello
         include Katello::Util::PulpcoreContentFilters
 
         def remote_options
-          if root.url.blank?
-            common_remote_options.merge(url: nil, policy: root.download_policy)
-          else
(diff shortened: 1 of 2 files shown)


jlsherrill:
@m-bucher would you be able to test this?  I don't have a sles account to actually check.  

I thought about adding a special user-supplied field for this, but a) it would break existing use cases, b) wouldn't make much sense with pulp2

So i think this is okay for now.   This PR still needs some tests though

m-bucher:
> @m-bucher would you be able to test this?

Tested with katello-repository creatd by foreman_scc_manager plugin and worked nicely.
Thanks @jlsherrill

m-bucher (approved):

jlsherrill:
[test katello]

sjha4 (approved):
Tested usual flow without sles token. Code looks good. ACK based on @m-bucher's testing of the functionality.

PR #8933, opened 2020-09-08
Fixes #30804 - Add a default status to selectAPIStatus selector

amirfefer:
Following this change in foreman - https://github.com/theforeman/foreman/pull/7956 we need to add `STATUS.PENDING` value to the `selectAPIStatus` selector


johnpmitsch (approved):
works well, thanks @amirfefer!

PR #8934, opened 2020-09-09
[CP] 3.16.1 cherry picks part 1

ianballou:
There are still bugs being waited on for the 3.16.1 release, but I'm starting the cherry picks early since there are a bunch of commits.


ianballou:
I had to add a ceiling to the foreman-tasks version since Foreman 2.1 is not compatible with foreman-tasks 3

jturel:
I see a rubocop complaint - was there a merge conflict incorrectly resolved?

ianballou:
@jturel yep and I just fixed it!  I think it should all be correct now.  Had to shuffle some bugs around between 3.16.1 and 3.17 due to dependencies.

PR #8938, opened 2020-09-11
Fixes #30819 - Allow to set prefix for subscription_manager_configuration_url

stejskalleos:


app/helpers/katello/katello_urls_helper.rb
```diff
@@ -6,12 +6,8 @@ def foreman_settings_url(schema = 'http')
...
     end
 
-    def subscription_manager_configuration_url(host = nil, rpm = true)
-      prefix = if host&.content_source
-                 "http://#{host.content_source.hostname}"
-               else
-                 foreman_settings_url
-               end
+    def subscription_manager_configuration_url(host = nil, rpm = true, prefix = nil)
+      prefix ||= host&.content_source ? "http://#{host.content_source.hostname}" : foreman_settings_url
```
ares, line 10: Turns out we'll need to do this slightly differently. We need to only get the hostname part of the url and use that instead of host&.content_source.hostname. Something like this

```ruby
def subscription_manager_configuration_url(host = nil, rpm = true, hostname = nil)
  prefix = if host&.content_source
￼                 "http://#{hostname || host.content_source.hostname}"
￼               else
￼                 foreman_settings_url
￼               end

...
```
My appologies for sending you the wrong direction first. We can do the parsing either in the template or here, but we must only take the hostname to ensure, the resulting url using http.

ares (changes requested):
There's one change we need to make, we need to make sure, we still use `http`

app/helpers/katello/katello_urls_helper.rb
ares: one clarification: the logic in fact is, if hostname is set, use that with http://, otherwise remains

stejskalleos: Maybe we shouldn't expect just `hostname`, but parse it from the `url argument`,
like this:
```ruby
def subscription_manager_configuration_url(host = nil, rpm = true, url = nil)
if url
  prefix = "http://#{URI(url).host}"
else
  prefix = host&.content_source ? "http://#{host.content_source.hostname}" : foreman_settings_url
end
```
That way users won't have to parse the URL manually. Thoughts?

ares: After a short discussion, it seems better to take just hosname here, the extraction should happen in the controller that renders the template. There are more urls and hostnames we'll need to prepare to variables prior the template rendering.

stejskalleos:
Rebased & updated

app/helpers/katello/katello_urls_helper.rb
```diff
@@ -6,12 +6,13 @@ def foreman_settings_url(schema = 'http')
...
-      prefix = if host&.content_source
-                 "http://#{host.content_source.hostname}"
-               else
-                 foreman_settings_url
-               end
+    def subscription_manager_configuration_url(host = nil, rpm = true, hostname = nil)
+      if hostname
+        prefix = "http://#{hostname}"
+      else
+        prefix = host&.content_source ? "http://#{host.content_source.hostname}" : foreman_settings_url
```
evgeni, line 13: can't this be `host&.content_source&.hostname or foreman_settings_url`?

jlsherrill (approved):

jlsherrill:
ack'd waiting on chrmoium :(

app/helpers/katello/katello_urls_helper.rb
jlsherrill: it would be missing the http:// part of the url, although now i'm wondering why its not https://    This should be https:// right?

ares: Should it? I think the katello-consumer rpm is not exposed via https but even if it is, for the bootstrap purpose, client at that time does not trust the certificate, the rpm contains it. Regardless of that, currently the method hardcodes that, we only allow specifying proxy hostname while keeping the same protocol. I believe this has been rebased now, is chromium alreayd resolved? Tests still seem red.

evgeni: The RPM is exposed on both HTTP and HTTPS today. But yeah, the certificate is not trusted at this step yet.

(bootstrap.py has a rather ugly hack to install RPMs from insecure HTTPS urls…)

Chromium issue should be fixed.

stejskalleos: Looks like we are green. @evgeni @jlsherrill can we merge?

evgeni: I have no hulk powers here.

ekohl: Wouldn't this be cleaner?
```ruby
      prefix = if hostname
                 "http://hostname"
               elsif host&.content_source
                 "http://#{host.content_source.hostname}"
               else
                 foreman_settings_url
               end
```

ekohl:
I am wondering if host and hostname wouldn't be ambiguous. The only solution I can think of would be to split the macro into two.

jlsherrill:
i think its okay for now, i think host is a bad name in general but that :ship:  has :boat: .  Going ahead and merging.

PR #8940, opened 2020-09-11
Refs #30630 - update ansible client to 0.3

jlsherrill:


sjha4 (approved):
Smoke tested with create, sync and delete.
Works as expected. ACK :+1:

PR #8945, opened 2020-09-14
Fixes #30735 - refactor tracer resolve actions.

chris1984:


engines/bastion_katello/app/assets/javascripts/bastion_katello/hosts/host-bulk-action.factory.js
```diff
@@ -26,7 +26,8 @@ angular.module('Bastion.hosts').factory('HostBulkAction',
             releaseVersion: {method: 'PUT', params: {action: 'release_version'}},
             availableIncrementalUpdates: {method: 'POST', isArray: true, params: {action: 'available_incremental_updates'}},
             moduleStreams: {method: 'POST', params: {action: 'module_streams'}},
-            traces: {method: 'POST', params: {action: 'traces'}}
+            traces: {method: 'POST', params: {action: 'traces'}},
+            resolve: {method: 'PUT', isArray: true, params: {action: 'resolve'}}
```
jturel, line 30: let's call this something like resolveTraces since it's a little ambiguous in this context (this factory can do all sorts of stuff)

test/controllers/api/v2/host_tracer_controller_test.rb
```diff
@@ -28,53 +28,25 @@ def test_index
...
+      ForemanTasks::Task.stubs(:find).returns(ForemanTasks::Task.new)
 
-      assert_response :success
-
-      response_body = JSON.parse(response.body)
-      assert_equal response_body, [job_invocation]
-    end
+      Katello::HostTraceManager.expects(:resolve_traces).with([trace_one]).returns([composer])
 
-    def test_resolve_reboot_service
```
jturel, line 47: This test and the others which test the grouping are now behaviors of the HostTracesManager, so these tests should be moved to a file which tests that class. Good new is since we're not dealing with any controller stuff, they become even simpler

test/controllers/api/v2/hosts_bulk_actions_controller_test.rb
```diff
@@ -2,7 +2,13 @@
 
 require "katello_test_helper"
 
+unless Katello.with_remote_execution?
```
jturel, line 5: This controller test doesn't invoke any code which references JobInvocationComposer, so this can be removed. It should be moved to the new test file for HostTracesManager. On current code, we have remote execution as a gem dependency, so it will always be defined. Although this code will be cherry-picked to an older branch which doesn't have that dep, so better to preserve it (in host traces manager test).

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-traces-modal.controller.js
```diff
@@ -46,7 +45,7 @@ angular.module('Bastion.content-hosts').controller('ContentHostsBulkTracesContro
                 });
             };
             /* eslint-disable camelcase */
-            HostTracesResolve.resolve({trace_ids: traceids}, onSuccess, onFailure);
+            HostBulkAction.resolve({trace_ids: traceids}, onSuccess, onFailure);
```
jturel, line 48: Nice!

engines/bastion_katello/app/assets/javascripts/bastion_katello/hosts/host-bulk-action.factory.js
chris1984: Updated to `resolve_traces` and the other to `fetch_traces`

test/controllers/api/v2/hosts_bulk_actions_controller_test.rb
chris1984: Removed from both controller tests and added to the service class test.

test/controllers/api/v2/host_tracer_controller_test.rb
chris1984: Yep, moved the 3 main tests to the service class tests, and kept 1 for the main resolve and 1 for the bulk resolve test.

test/services/katello/host_trace_manager.rb
```diff
@@ -0,0 +1,64 @@
...
+      trace_two = Katello::HostTracer.create(host_id: @host2.id, application: 'rsyslog', app_type: 'daemon', helper: 'systemctl restart rsyslog')
+      helper = {:helper => trace_two.helper}
+      traces = [trace_one, trace_two]
+      job_invocation = {"description" => "Restart Services", "id" => 1, "job_category" => "Katello"}
+
+      JobInvocationComposer.expects(:for_feature).with(:katello_service_restart, [@host1.id, @host2.id], helper).returns(mock(trigger: true, job_invocation: job_invocation))
+
+      Katello::HostTraceManager.resolve_traces(traces)
+
+      result = [job_invocation]
```
jturel, line 26: Does this actually test the result of `resolve_traces` ?

engines/bastion_katello/app/assets/javascripts/bastion_katello/hosts/host-bulk-action.factory.js
```diff
@@ -26,7 +26,8 @@ angular.module('Bastion.hosts').factory('HostBulkAction',
             releaseVersion: {method: 'PUT', params: {action: 'release_version'}},
             availableIncrementalUpdates: {method: 'POST', isArray: true, params: {action: 'available_incremental_updates'}},
             moduleStreams: {method: 'POST', params: {action: 'module_streams'}},
-            traces: {method: 'POST', params: {action: 'traces'}}
+            traces: {method: 'POST', params: {action: 'fetch_traces'}},
+            resolve: {method: 'PUT', isArray: true, params: {action: 'resolve_traces'}}
```
jturel, line 30: I was expecting to see this called `resolve_traces` but it's still called `resolve` (you changed only the server side, and the previous name of `traces` was fine - the 'fetch' is essentially implied)

lib/katello/permissions/host_permissions.rb
```diff
@@ -26,7 +26,8 @@
   'katello/api/v2/hosts_bulk_actions/content_overrides',
   'katello/api/v2/hosts_bulk_actions/environment_content_view',
   'katello/api/v2/hosts_bulk_actions/release_version',
-  'katello/api/v2/hosts_bulk_actions/traces',
+  'katello/api/v2/hosts_bulk_actions/fetch_traces',
```
jturel, line 29: This controller method and route should be renamed back to 'traces' from 'fetch_traces'

test/services/katello/host_trace_manager.rb
chris1984: I thought it was, I am really having trouble getting a return value of an array, instead of the mock object.

test/controllers/api/v2/host_tracer_controller_test.rb
```diff
@@ -28,53 +22,25 @@ def test_index
...
 
-    def test_resolve_group_by_helper
+    def test_resolve
       trace_one = Katello::HostTracer.create(host_id: @host1.id, application: 'rsyslog', app_type: 'daemon', helper: 'systemctl restart rsyslog')
-      trace_two = Katello::HostTracer.create(host_id: @host2.id, application: 'rsyslog', app_type: 'daemon', helper: 'systemctl restart rsyslog')
-      helper = {:helper => trace_two.helper}
-
-      job_invocation = {"description" => "Restart Services", "id" => 1, "job_category" => "Katello"}
-      JobInvocationComposer.expects(:for_feature).with(:katello_service_restart, [@host1.id, @host2.id], helper).returns(mock(trigger: true, job_invocation: job_invocation))
+      task_output = {"id" => nil, "label" => nil, "started_at" => nil, "ended_at" => nil, "state" => nil, "result" => nil, "external_id" => nil,
```
jturel, line 27: It looks like this works but imho it's purely a coincidence that this output is matching with the below `ForemanTasks::Task.new`. I recommend using the same task in both places. ie

```
task = ForemanTasks::Task.new
....

Which you can return from the stubbed find method, and you can call `to_json` on it to make your assertion in the response body

**Review by jturel — CHANGES_REQUESTED** [review]

File: test/services/katello/host_trace_manager.rb (hunk 5 above)
Comment (jturel) [reply test/services/katello/host_trace_manager.rb:26]: First think you'll want to do is actually test the response of `resolve_traces` so something like this:

```
result = Katello::HostTraceManager.resolve_traces(traces)

assert_equal 1, result.size
```

File: engines/bastion_katello/app/assets/javascripts/bastion_katello/hosts/host-bulk-action.factory.js (hunk 6 above)
Comment (chris1984) [reply engines/bastion_katello/app/assets/javascripts/bastion_katello/hosts/host-bulk-action.factory.js:30]: Updated, I was trying to be fancy with the fetch but I see now what you mean. Thanks for ruining my fun :laughing:

File: lib/katello/permissions/host_permissions.rb (hunk 7 above)
Comment (chris1984) [reply lib/katello/permissions/host_permissions.rb:29]: Updated

File: test/services/katello/host_trace_manager.rb (hunk 5 above)
Comment (jturel) [reply test/services/katello/host_trace_manager.rb:26]: It would be good to ensure that this works with an AR relation as input rather than just an array so you could do this: `Katello::HostTracer.where(id: trace.id)`. Mentioning this since that's how we use this class in the controllers

File: test/services/katello/host_trace_manager.rb [hunk 9]
```diff
@@ -0,0 +1,64 @@
...
+    def test_resolve_group_by_helper_one_invocation
+      trace_one = Katello::HostTracer.create(host_id: @host1.id, application: 'rsyslog', app_type: 'daemon', helper: 'systemctl restart rsyslog')
+      trace_two = Katello::HostTracer.create(host_id: @host2.id, application: 'rsyslog', app_type: 'daemon', helper: 'systemctl restart rsyslog')
+      helper = {:helper => trace_two.helper}
+      traces = [trace_one, trace_two]
+      job_invocation = {"description" => "Restart Services", "id" => 1, "job_category" => "Katello"}
+
+      JobInvocationComposer.expects(:for_feature).with(:katello_service_restart, [@host1.id, @host2.id], helper).returns(mock(trigger: true, job_invocation: job_invocation))
+
+      Katello::HostTraceManager.resolve_traces(traces)
```
jturel, line 24: It would be good to ensure that this works with an AR relation as input rather than just an array so you could do this: Katello::HostTracer.where(id: trace.id). Mentioning this since that's how we use this class in the controllers

chris1984: Makes sense, changed to reflect the way we resolve the traces instead of an array in the test and use the AR relation for the resolve.

test/controllers/api/v2/host_tracer_controller_test.rb
chris1984: Updated

test/services/katello/host_trace_manager.rb
chris1984: Updated

chris1984:
Went through and tested UI on a single host and bulk. I was able to fetch and resolve traces both ways and saw REX kick off the job and host tracer profile upload, even though the REX job hung, which is something with the way rex is on the dev box. 

I also tested with my hammer pr and was able to list/resolve as well from that.

test/controllers/api/v2/hosts_bulk_actions_controller_test.rb
```diff
@@ -380,7 +381,7 @@ def test_get_host_traces
       Katello::HostTracer.create(host_id: @host1.id, application: 'kernel', app_type: 'static', helper: 'reboot the system')
       Katello::HostTracer.create(host_id: @host2.id, application: 'dbus', app_type: 'static', helper: 'reboot the system')
 
-      post :traces, params: {
+      get :traces, params: {
```
jturel, line 384: afaik this method is still a POST so I'm curious how this actually works! Can you see what's going on here?

chris1984: Updated to POST to match the bulk controller, here are the results from both:

With POST: 
```ruby
[1] pry(#<Katello::Api::V2::HostsBulkActionsControllerTest>)> response.body
=> "{\"total\":2,\"subtotal\":2,\"page\":1,\"per_page\":20,\"error\":null,\"search\":null,\"sort\":{\"by\":\"application\",\"order\":\"desc\"},\"results\":[{\"id\":890,\"application\":\"kernel\",\"helper\":\"reboot the system\",\"app_type\":\"static\",\"host_id\":980191011,\"host\":\"host49\",\"reboot_required\":true},{\"id\":891,\"application\":\"dbus\",\"helper\":\"reboot the system\",\"app_type\":\"static\",\"host_id\":980191012,\"host\":\"host50\",\"reboot_required\":true}]}\n"
[2] pry(#<Katello::Api::V2::HostsBulkActionsControllerTest>)> response.body.class
=> String

[1] pry(#<Katello::Api::V2::HostsBulkActionsController>)> collection
=> {:results=>
  [#<Katello::HostTracer:0x0000000018b61238 id: 900, host_id: 980191001, application: "kernel", helper: "reboot the system", app_type: "static">,
   #<Katello::HostTracer:0x0000000018b61080 id: 901, host_id: 980191002, application: "dbus", helper: "reboot the system", app_type: "static">],
 :subtotal=>2,
 :total=>2,
 :page=>1,
 :per_page=>20,
 :error=>nil}
[2] pry(#<Katello::Api::V2::HostsBulkActionsController>)> foo = respond_for_index(:collection => collection, :template => '../../../api/v2/host_tracer/index')
=> "{\"total\":2,\"subtotal\":2,\"page\":1,\"per_page\":20,\"error\":null,\"search\":null,\"sort\":{\"by\":\"application\",\"order\":\"desc\"},\"results\":[{\"id\":900,\"application\":\"kernel\",\"helper\":\"reboot the system\",\"app_type\":\"static\",\"host_id\":980191001,\"host\":\"host39\",\"reboot_required\":true},{\"id\":901,\"application\":\"dbus\",\"helper\":\"reboot the system\",\"app_type\":\"static\",\"host_id\":980191002,\"host\":\"host40\",\"reboot_required\":true}]}\n"
[3] pry(#<Katello::Api::V2::HostsBulkActionsController>)> foo.class
=> ActionView::OutputBuffer
```

With GET:

```ruby
[1] pry(#<Katello::Api::V2::HostsBulkActionsControllerTest>)> response.body
=> "{\"total\":2,\"subtotal\":2,\"page\":1,\"per_page\":20,\"error\":null,\"search\":null,\"sort\":{\"by\":\"application\",\"order\":\"desc\"},\"results\":[{\"id\":893,\"application\":\"kernel\",\"helper\":\"reboot the system\",\"app_type\":\"static\",\"host_id\":980191013,\"host\":\"host51\",\"reboot_required\":true},{\"id\":894,\"application\":\"dbus\",\"helper\":\"reboot the system\",\"app_type\":\"static\",\"host_id\":980191014,\"host\":\"host52\",\"reboot_required\":true}]}\n"
[2] pry(#<Katello::Api::V2::HostsBulkActionsControllerTest>)> response.class
=> ActionDispatch::TestResponse

[1] pry(#<Katello::Api::V2::HostsBulkActionsController>)> collection
=> {:results=>
  [#<Katello::HostTracer:0x00000000184b3378 id: 897, host_id: 980190995, application: "kernel", helper: "reboot the system", app_type: "static">,
   #<Katello::HostTracer:0x00000000184b31c0 id: 898, host_id: 980190996, application: "dbus", helper: "reboot the system", app_type: "static">],
 :subtotal=>2,
 :total=>2,
 :page=>1,
 :per_page=>20,
 :error=>nil}
[2] pry(#<Katello::Api::V2::HostsBulkActionsController>)> foo = respond_for_index(:collection => collection, :template => '../../../api/v2/host_tracer/index')
=> "{\"total\":2,\"subtotal\":2,\"page\":1,\"per_page\":20,\"error\":null,\"search\":null,\"sort\":{\"by\":\"application\",\"order\":\"desc\"},\"results\":[{\"id\":897,\"application\":\"kernel\",\"helper\":\"reboot the system\",\"app_type\":\"static\",\"host_id\":980190995,\"host\":\"host33\",\"reboot_required\":true},{\"id\":898,\"application\":\"dbus\",\"helper\":\"reboot the system\",\"app_type\":\"static\",\"host_id\":980190996,\"host\":\"host34\",\"reboot_required\":true}]}\n"
[3] pry(#<Katello::Api::V2::HostsBulkActionsController>)> foo.class
=> ActionView::OutputBuffer
```

app/controllers/katello/api/v2/host_tracer_controller.rb
```diff
@@ -2,50 +2,28 @@ module Katello
...
-
-      job_invocations = []
-
-      composers.each do |composer|
-        composer.trigger
-        job_invocations << composer.job_invocation
-      end
-
-      render json: job_invocations
+      render json: task
```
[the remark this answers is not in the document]
chris1984: It looks like we have to define a task instead if we use this:
```ruby
task = async_task(::Actions::Katello::ContentView::Promote,
                        @version, @environments, is_force, params[:description])
respond_for_async :resource => task
```

Since we are having the service class and the Composer take care of that, I think we should keep it as is. Also the way we are returning the task ID, we eliminate the need for the hammer-foreman-tasks pr that I had originally opened. 

Given the time crunch and with it working currently with hammer/ui, I vote for keeping it the way it is now, and when we address new issues, we can experiment with it there. I am happy to make a redmine for it so we have record of it.

chris1984:
@jturel updated the tests with what we talked about.

test/services/katello/host_trace_manager_test.rb
```diff
@@ -0,0 +1,65 @@
+require 'katello_test_helper'
+
+unless Katello.with_remote_execution?
+  class JobInvocationComposer
```
jturel, line 4: Can you confirm that these tests pass when we call into this block? Simplest way would be to just comment out the `unless` and `end` that way, when this reaches downstream where we'll enter into this block, that the test still pass!

jturel:
I don't see any issues with the code! will give it a test later tonight

test/services/katello/host_trace_manager_test.rb
chris1984: They do still pass when I comment out `unless` and `end`

```ruby
[vagrant@centos7-katello-devel-stable foreman]$ ktest ../katello/test/services/katello/host_trace_manager_test.rb 
/home/vagrant/foreman/lib/foreman.rb:8: warning: already initialized constant Foreman::UUID_REGEXP
/home/vagrant/foreman/lib/foreman.rb:8: warning: previous definition of UUID_REGEXP was here
   -> 0.0005s
Run options: --seed 50217

# Running:

...

Finished in 2.459055s, 1.2200 runs/s, 6.0999 assertions/s.

3 runs, 15 assertions, 0 failures, 0 errors, 0 skips
Coverage report Rcov style generated for Unit Tests to /home/vagrant/foreman/coverage/rcov
Coverage report generated for Unit Tests to /home/vagrant/foreman/coverage. 0 / 0 LOC (100.0%) covered.
[vagrant@centos7-katello-devel-stable katello]$ git diff
diff --git a/test/services/katello/host_trace_manager_test.rb b/test/services/katello/host_trace_manager_test.rb
index 8cc693f..397dbbf 100644
--- a/test/services/katello/host_trace_manager_test.rb
+++ b/test/services/katello/host_trace_manager_test.rb
@@ -1,8 +1,8 @@
 require 'katello_test_helper'
 
-unless Katello.with_remote_execution?
+#unless Katello.with_remote_execution?
   class JobInvocationComposer
-  end
+#  end
 end
 
 module Katello
```

jturel:
In dev, when going to resolve bulk traces, I see this: Do you see the same? 

```
No route matches [PUT] "/katello/api/v2/traces/resolve"
```

jturel:
If you push any more commits to fix above, please do them as additional Refs # ... commits on top of where's here so it's easier to review what changed!

app/controllers/katello/api/v2/hosts_bulk_actions_controller.rb
```diff
@@ -236,6 +237,16 @@ def traces
       respond_for_index(:collection => collection, :template => '../../../api/v2/host_tracer/index')
     end
 
+    api :PUT, "/hosts/bulk/resolve_traces", N_("Resolve traces for one or more hosts")
+    param_group :bulk_params
+    param :trace_ids, Array, :required => true, :desc => N_("Array of Trace IDs")
+    def resolve_traces
+      traces = Katello::HostTracer.resolvable.where(id: params[:trace_ids])
```
jturel, line 244: This code currently allows resolving traces for hosts other than --host-id !

The code should be something like: `@host.traces.resolvable.where(id: params[:trace_ids])`

And if `traces` ends up being empty, I think the API should throw an error

app/controllers/katello/api/v2/host_tracer_controller.rb
```diff
@@ -2,50 +2,28 @@ module Katello
...
       collection = scoped_search(index_relation, :application, :asc, :resource_class => ::Katello::HostTracer)
       respond_for_index(:collection => collection)
     end
 
-    api :PUT, "/traces/resolve", N_("Resolve Traces")
+    api :PUT, "/hosts/:host_id/traces/resolve", N_("Resolve Traces")
+    param :host_id, :number, :required => true, :desc => N_("ID of the host")
     param :trace_ids, Array, :required => true, :desc => N_("Array of Trace IDs")
     def resolve
       traces = Katello::HostTracer.resolvable.where(id: params[:trace_ids])
```
jturel, line 21: This code currently allows resolving traces for hosts other than --host-id !

The code should be something like: `@host.traces.resolvable.where(id: params[:trace_ids])`

And if `traces` ends up being empty, I think the API should throw an error

jturel: Call `.to_sql` on the above example to see how it affects the query

app/controllers/katello/api/v2/host_tracer_controller.rb
```diff
@@ -2,50 +2,28 @@ module Katello
...
+      result = Katello::HostTraceManager.resolve_traces(traces)
+
+      task = ForemanTasks::Task.find(result.first.job_invocation.task_id)
```
jturel, line 24: We're getting back a list of job invocations from HostTraceManager.resolve so I think this should be written: `result.first.task_id`

jturel:
Overall tests well in conjunction with https://github.com/Katello/hammer-cli-katello/pull/739/files - left my comments

test/services/katello/host_trace_manager_test.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+end
+
+module Katello
+  class HostTraceManagerTest < ActiveSupport::TestCase
+    def setup
+      @host1 = FactoryBot.create(:host)
```
jturel, line 11: I think we can just use the fixtures `hosts(:one)` and two like we did previously

chris1984: Updated to use the fixtures instead of making more hosts.

app/controllers/katello/api/v2/host_tracer_controller.rb
chris1984: Updated the return value, thanks for catching this

jturel:
Now I see a good error when I try to resolve inexistent traces or ones that don't belong to my host:

```
[vagrant@centos7-hammer-devel hammer-cli-katello]$ hammer host traces resolve --host-id=3 --trace-ids=100
Could not resolve traces:
  The requested traces were not found for this host
```

jturel (approved):
Tested this in the UI for regressions and with the related hammer PR. Everything looks good. APJ!

Take a picture and frame this beautiful PR that covers numerous technical areas - adding a new api - refactoring an existing API - related UI refacotring - implementing a service class - with all the test coverage you'd ever want.

TLDR: NICE

jturel:
Oh there's one more thing ...

....





.....







....

Just remember to squash and merge :)

chris1984:
[test katello]

chris1984:
[test katello]

jlsherrill:
confirmed that angular ci runs locally. going ahead and merging.

PR #8949, opened 2020-09-16
Fixes #30371 - Fix timezone display on sync date field in UI

sjha4:
Opening this for better ideas to fix this. 

To test:
1. Create a sync plan
2. Go to Logged in user (top right corner dropdown)>  My Account
3. The default Timezone should be Browser timezone. 
4. Change it to some different timezone.
5. Refresh sync plan page. Notice the sync plan start date field does not respected this timezone.

After this change, it should display start date in the user account timezone and not the browser timezone similar to next sync date.

The problem I ran into is that while the next sync date is a react component , the start date is a bastion bst-edit-custom field and I wasn't able to make it respect account timezone over browser timezone...Hence falling back to the API.

### DIFF
--- a/app/models/katello/sync_plan.rb
+++ b/app/models/katello/sync_plan.rb
@@ -121,6 +121,14 @@ module Katello
       errors.add :base, _("Start Date and Time can't be blank") if self.sync_date.nil?
     end
 
+    def sync_date_sans_tz
+      if User.current.try(:timezone)
+        return self.sync_date.strftime('%a, %d %b %Y %H:%M:%S')
+      else
+        sync_date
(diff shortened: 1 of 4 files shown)


sjha4:
Pushed an update to fix behavior on updating the start date. Should work correctly now.

jjeffers (commented):
Screenshot from 2020-09-16 08-38-56

Looks like the expression of the month is truncated in one field, but not in the other. Not sure if this is something we need to change.

jjeffers (approved):

sjha4:
[test katello]

PR #8955, opened 2020-09-18
Fixes #30873 - activation key authorization

jlsherrill:


test/controllers/api/v2/activation_keys_controller_test.rb
```diff
@@ -80,12 +80,23 @@ def test_show_protected
...
     end
 
+    def stub_view_environment(env)
+      Katello::KTEnvironment.stubs(:readable).returns(stub(:find_by => env))
+    end
+
+    def stub_view_content_view(view)
+      Katello::ContentView.stubs(:readable).returns(stub(:find_by => view))
+    end
+
```
parthaa, line 90: Can you add a test asserting show specific instance with good and bad perms

test/controllers/api/v2/activation_keys_controller_test.rb
```diff
@@ -80,12 +80,23 @@ def test_show_protected
...
     def test_create_protected
+      stub_view_environment(@library)
+      stub_view_content_view(@view)
+
```
parthaa, line 94: You could do something similar to https://github.com/Katello/katello/pull/8957/files#diff-950d343e8ded0e9a59fbe8561c825bb6L427-L440 instead of stubbing. I dont mind either way

jlsherrill:
updated @parthaa

parthaa (approved):
Works well

PR #8956, opened 2020-09-18
Fixes #30880 - Permission support to validate 404

parthaa:
This commit enables checking authorizing a specific instance and ensure
that it handles 404 correctly.
It also adds support for multi possible permissions for same action.
For example destroy_content_views and promote_or_remove_content_views
Either support 'removing' a content view.
This commit adds support to handle that.

### DIFF
--- a/test/controllers/api/v2/repositories_controller_test.rb
+++ b/test/controllers/api/v2/repositories_controller_test.rb
@@ -185,17 +185,15 @@ module Katello
       assert_response_ids response, ids
     end
 
-    # def test_index_with_content_view_version_id_and_library
-    #   # version  = @organization.default_content_view_version
-    #   # ids = version.repositories.pluck(:library_instance_id).reject(&:blank?).uniq
-    #   ids = @view.versions.first.repositories.pluck(:library_instance_id).reject(&:blank?).uniq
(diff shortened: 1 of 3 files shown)


test/controllers/api/v2/repositories_controller_test.rb
```diff
@@ -543,6 +541,15 @@ def test_show_protected
       end
     end
 
+    def test_show_protected_specific_instance
+      allowed_perms = [{:name => @read_permission, :search => "name=\"#{@repository.product.name}\"" }]
+      denied_perms = [{:name => @read_permission, :search => "name=\"#{@redhat_repository.product.name}\"" }]
+
+      assert_protected_action(:show, allowed_perms, denied_perms, expect_404: true) do
```
jlsherrill, line 548: what do you think about having a shortcut method  'assert_protected_object'  , which just calls assert_protected_action, passing expect_404?

parthaa: added

jlsherrill (approved):
APJ

parthaa:
[test katello]

parthaa:
[test katello]

PR #8959, opened 2020-09-18
Fixes #30878 - content credentials authorization

jlsherrill:
and drop gpg key permissions

### DIFF
--- a/db/migrate/20201012172713_remove_gpg_key_perms.rb
+++ b/db/migrate/20201012172713_remove_gpg_key_perms.rb
@@ -1,0 +1,23 @@
+class RemoveGpgKeyPerms < ActiveRecord::Migration[6.0]
+  def permission_map
+    {
+      Permission.find_by(name: :view_gpg_keys) => Permission.find_by(name: :view_content_credentials),
+      Permission.find_by(name: :edit_gpg_keys) => Permission.find_by(name: :edit_content_credentials),
+      Permission.find_by(name: :destroy_gpg_keys) => Permission.find_by(name: :destroy_content_credentials),
+      Permission.find_by(name: :create_gpg_keys) => Permission.find_by(name: :create_content_credentials)
+    }
+  end
+
+  def up
+    perms = permission_map
+    perms.each do |old_perm, new_perm|
+      Filtering.where(permission_id: old_perm.id).update_all(:permission_id => new_perm.id) if old_perm
+    end
+    names = perms.keys.compact.map(&:name)
+    Permission.where(:name => names).destroy_all if names.any?
+  end
+
+  def down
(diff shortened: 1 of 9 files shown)


jlsherrill:
[test katello]

jlsherrill:
[test katello]

lib/katello/permission_creator.rb
```diff
@@ -150,50 +149,34 @@ def content_view_permissions # rubocop:disable Metrics/MethodLength
...
-                         },
-                         :resource_type => 'Katello::GpgKey'
-      @plugin.permission :destroy_gpg_keys,
-                         {
-                           'katello/api/v2/gpg_keys' => [:destroy]
-                         },
-                         :resource_type => 'Katello::GpgKey'
-    end
-
     def content_credential_permissions
```
jturel, line 152: `view_gpg_keys` is referenced in lib/katello/plugin.rb and should be `view_content_credentials`

You might also consider updating the `humanized_name` of the GPG Key model to `Content Credentials` so that the Filters UI makes more send (resource called GPG Keys but perm is *_content_credentials...)

Would you also be interested in deprecating the gpg_keys routes in favor of content_credentials? Ya opened up a can of worms here ;)

jlsherrill: oh man, i thought these were already deprecated. let me do that

jlsherrill: removed the view_gpg_keys, references, changed humanized_name, deprecated the API's via apipie  properly (previously there was just a log message), and added a migration. 

The migration deletes permissions and moves any existing *gpg_key permissions to content credentials permissions.  The downside is that you may end up with duplicate perms on a filter, but i think thats better than missing them outright.

db/migrate/20201012172713_remove_gpg_key_perms.rb
```diff
@@ -0,0 +1,22 @@
...
+      Permission.find_by(name: :create_gpg_keys) => Permission.find_by(name: :create_content_credentials)
+    }
+  end
+
+  def up
+    perms = permission_map
+    perms.each do |old_perm, new_perm|
+      Filtering.where(permission_id: old_perm.id).update_all(:permission_id => new_perm.id)
+    end
+    Permission.where(:name => perms.keys.map(&:id)).destroy_all
```
jturel, line 16: name vs id mismatch

jlsherrill: doh! in my testing i checked the filterings but not the permission!

jturel: <3 tests

jturel (approved):
APJ!

PR #8966, opened 2020-09-24
Fixes #30907 - Pulp 3 Docker proxy syncing does not work

ianballou:
Fixes issues with Docker proxy syncing.

To test:
1) Set up a content-enabled smart proxy with Pulp 3
2) Sync some Docker content to that proxy.  This will fail without my PR
3) Check that the content was actually synced via the Pulp 3 API or by looking in the Pulp database

### DIFF
--- a/app/services/katello/pulp3/repository_mirror.rb
+++ b/app/services/katello/pulp3/repository_mirror.rb
@@ -110,8 +110,11 @@ module Katello
         api.remotes_list(name: backend_object_name).first
       end
 
-      def sync
-        repository_sync_url_data = api.class.repository_sync_url_class.new(remote: remote_href, mirror: true)
+      def sync(options = {})
+        sync_params = repo_service.sync_url_params(options)
+        sync_params[:remote] = remote_href
+        sync_params[:mirror] = true
+        repository_sync_url_data = api.class.repository_sync_url_class.new(sync_params)
         [api.repositories_api.sync(repository_href, repository_sync_url_data)]
       end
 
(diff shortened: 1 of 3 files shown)


app/services/katello/pulp3/repository_mirror.rb
```diff
@@ -110,8 +110,10 @@ def fetch_remote
         api.remotes_list(name: backend_object_name).first
       end
 
-      def sync
-        repository_sync_url_data = api.class.repository_sync_url_class.new(remote: remote_href, mirror: true)
+      def sync(options = {})
+        options.delete(:optimize) unless repo.content_type == "yum"
+        options.delete(:skip_types) unless repo.content_type == "yum"
```
jlsherrill, line 115: could we move the knowledge about these options into the pulp3 yum repo service?

Also i'm surprised 'skip_types' was being passed in, that shouldn't ever be used with a smart proxy sync

ianballou: Sure thing.  I was deleting skip_types just because I saw it was a possible parameter for yum repos.  I can remove that deletion if skip_types will never be there.

jlsherrill: its fine to delete, i was more curious if you actually saw it being passed in, as that could indicate something unexpected was going on

ianballou: Think I've fixed it.  Not able to test functionally yet due to my broken proxy env, but the tests pass at least.

ianballou:
[test katello]

ianballou:
[test katello]

app/services/katello/pulp3/repository_mirror.rb
ianballou: Tested the functionality, still works

jlsherrill (approved):

PR #8971, opened 2020-09-30
Fixes #30937 - be dynflow client on post migration check

jlsherrill:


lib/katello/tasks/pulp3_post_migration_check.rake
```diff
@@ -1,8 +1,9 @@
 require File.expand_path("../engine", File.dirname(__FILE__))
+load "#{Katello::Engine.root}/lib/katello/tasks/common.rake"
 
 namespace :katello do
   desc "Runs a post Pulp3 migration check for supported content types."
-  task :pulp3_post_migration_check => :environment do
+  task :pulp3_post_migration_check => [:environment, :disable_dynflow] do
```
adamruzicka, line 6: Would be cool if Katello could switch over to the definition of this from Foreman https://github.com/theforeman/foreman/blob/develop/lib/tasks/dynflow.rake#L33 and eventually got rid of the katello-specific way of doing things

jlsherrill: yes! thank you for mentioning this.  I will open an issue and put it on our sprint to change this across the board.  Will leave this change as is for now

ianballou (approved):
ACK pending jenkins!

lib/katello/tasks/pulp3_post_migration_check.rake
jlsherrill: opened https://projects.theforeman.org/issues/30941

PR #8973, opened 2020-09-30
Fixes #30944 - pass registry redirects to client

jlsherrill:
for blobs and manifests, pulp3 will redirect us to
the content app with a specially generated auth token
we can just pass this back to the client and redirect them
for improved performance


jlsherrill:
tested with both pulp2 (where redirects won't happen), and pulp3 where they will

ianballou (approved):
works for me!

PR #8975, opened 2020-10-01
Fixes #30824 - Optimize capsule sync by tracking sync history

sjha4:


app/lib/actions/katello/capsule_content/sync_capsule.rb
```diff
@@ -3,24 +3,28 @@ module Katello
...
+          smart_proxy_history_ids = []
           sequence do
             repos = repos_to_sync(smart_proxy, environment, content_view, repository)
 
             repos.in_groups_of(Setting[:foreman_proxy_content_batch_size], false) do |repo_batch|
               concurrence do
                 repo_batch.each do |repo|
-                  plan_pulp_action([Actions::Pulp::Orchestration::Repository::SmartProxySync,
-                                    Actions::Pulp3::CapsuleContent::Sync],
+                  action_needed = action_needed?(repo, smart_proxy, skip_metadata_check)
```
jlsherrill, line 20: instead of doing this on a per-repo basis, could we make this a scope on the Repository object?    It might have te in the repos_to_sync() method, but being able to build that into the list would be a lot more efficient

app/lib/actions/katello/capsule_content/sync_capsule.rb
```diff
@@ -47,9 +56,38 @@ def repos_to_sync(smart_proxy, environment, content_view, repository)
           end
         end
 
+        def action_needed?(repo, smart_proxy, skip_metadata_check)
+          if skip_metadata_check
+            sp_history = ::Katello::SmartProxySyncHistory.where(:smart_proxy_id => smart_proxy.id, :repository_id => repo.id)
+            sp_history.try(:delete_all)
+            true
+          else
+            sp_history = ::Katello::SmartProxySyncHistory.where(:smart_proxy_id => smart_proxy.id, :repository_id => repo.id)
```
jlsherrill, line 65: I think you need a   finished_at != nil  in the where clause too

app/lib/actions/katello/capsule_content/sync_capsule.rb
```diff
@@ -47,9 +56,38 @@ def repos_to_sync(smart_proxy, environment, content_view, repository)
...
+            sp_history.empty?
+          end
+        end
+
+        def create_smart_proxy_sync_history(repo, smart_proxy)
+          #Delete old record
+          sp_history = ::Katello::SmartProxySyncHistory.where(:smart_proxy_id => smart_proxy.id, :repository_id => repo.id)
+          sp_history.delete_all
```
jlsherrill, line 73: nitpick but i think you could just do this in one line

app/lib/actions/katello/repository/capsule_sync.rb
```diff
@@ -12,7 +12,7 @@ def plan(repo)
               smart_proxies = ::SmartProxy.with_environment(repo.environment)
               unless smart_proxies.blank?
                 plan_action(::Actions::BulkAction, ::Actions::Katello::CapsuleContent::Sync, smart_proxies,
-                            :repository_id => repo.id)
+                            :repository_id => repo.id, :skip_metadata_check => false)
```
jlsherrill, line 15: do you remember what this was needed for?

app/lib/actions/katello/repository/import_upload.rb
```diff
@@ -80,6 +82,12 @@ def results_to_json(results)
           end
           json_results
         end
+
+        def clear_smart_proxy_history(repository)
+          repos_to_clear = ::Katello::Repository.where(:root_id => repository.root_id).pluck(:id)
+          sp_history = ::Katello::SmartProxySyncHistory.where(:repository_id => repos_to_clear)
+          sp_history.try(:delete_all)
```
jlsherrill, line 89: would it make sense to put this on the Repository object itself?

app/lib/actions/katello/repository/remove_content.rb
```diff
@@ -52,6 +52,12 @@ def finalize
           end
         end
 
+        def clear_smart_proxy_history(repository)
+          repos_to_clear = ::Katello::Repository.where(:root_id => repository.root_id).pluck(:id)
+          sp_history = ::Katello::SmartProxySyncHistory.where(:repository_id => repos_to_clear)
+          sp_history.try(:delete_all)
+        end
```
jlsherrill, line 59: It looks like here you are clearing the history across all instances of that Root Repositroy, but i think we should just do it for that particular repository, Thoughts?

jlsherrill:
I think we also need to 'clear the history' if you sync a repo and its contents have changed

sjha4:
@jlsherrill : Updated..Do the changes look like what you had in mind?

app/lib/actions/katello/repository/capsule_sync.rb
sjha4: Removed this..I was trying something for repo sync and forgot to clear smart_proxy_sync_history when contents change upon sync..Updated.

app/lib/actions/helpers/smart_proxy_sync_history_helper.rb
```diff
@@ -0,0 +1,17 @@
...
+  module Helpers
+    module SmartProxySyncHistoryHelper
+      def self.included(base)
+        base.middleware.use ::Actions::Middleware::RecordSmartProxySyncHistory
+      end
+
+      def done?
+        is_done = super
+        if is_done
+          ::Katello::SmartProxySyncHistory.where(:id => output[:smart_proxy_history_ids]).update_all(finished_at: Time.now)
```
jlsherrill, line 11: hah, i just realized we're not actually checking for errors here.  We want to delete it if there are errors

jlsherrill: or just not update it as we discussed

app/lib/actions/middleware/record_smart_proxy_sync_history.rb
```diff
@@ -0,0 +1,17 @@
...
+  module Middleware
+    class RecordSmartProxySyncHistory < Dynflow::Middleware
+      def run(*args)
+        smart_proxy_history_ids = []
+        if (action.input[:repository_id] && (action.input[:smart_proxy_id] || action.input[:capsule_id]) && !self.action.output[:smart_proxy_history_ids])
+          repo = ::Katello::Repository.find(action.input[:repository_id])
+          smart_proxy_id = action.input[:smart_proxy_id] || action.input[:capsule_id]
+          smart_proxy = ::SmartProxy.find(smart_proxy_id)
+          smart_proxy_history_ids << repo.create_smart_proxy_sync_history(smart_proxy)
+          self.action.output[:smart_proxy_history_ids] = smart_proxy_history_ids
```
jlsherrill, line 11: a little niptick, i think only one id can exist, so maybe rename  output[:smart_proxy_history_ids] to smart_proxy_history_id ?

jlsherrill:
One issue i found, if you do a 'complete sync' on the smart proxy details page, it syncs all repos as expected.   However if you then do a regular 'optimized sync', it still syncs all repositories.  (a 2nd optimized sync doesn't).  It seems like the history events aren't being recreated during a complete sync.

app/lib/actions/helpers/smart_proxy_sync_history_helper.rb
```diff
@@ -0,0 +1,17 @@
...
+    module SmartProxySyncHistoryHelper
+      def self.included(base)
+        base.middleware.use ::Actions::Middleware::RecordSmartProxySyncHistory
+      end
+
+      def done?
+        is_done = super
+        if is_done
+          ::Katello::SmartProxySyncHistory.where(:id => output[:smart_proxy_history_id]).update_all(finished_at: Time.now)
+        end
```
jlsherrill, line 12: this still needs to check for errors, right?

sjha4: Leaning towards leaving the record as is. In case, there's failures, we will have a sync history record without finished_at which will be cleared upon the next sync..

jlsherrill: but if there are errors, it won't ever sync again, right?
Say i sync my some my capsule and it errors, so i sync it again.  If finished_at is set, it won't sync again?

jlsherrill: and when i say 'it won't ever sync again' i mean until there are changes to the repo

jlsherrill: oh and one little nitpick, you might change it to:
```
           ::Katello::SmartProxySyncHistory.where(:id => output[:smart_proxy_history_id], :finished_at => nil).update_all(finished_at: Time.now)
```
So it won't re-update anything if done? is called multiple times after its completed

sjha4: > and when i say 'it won't ever sync again' i mean until there are changes to the repo

So it will pick up all repos without a sync history record as well as records without the finished_at set for the next sync.

sjha4:
[test katello]

sjha4:
[test katello]

sjha4:
[test katello]

sjha4:
[test katello]

sjha4:
[test katello]

app/models/katello/smart_proxy_sync_history.rb
```diff
@@ -0,0 +1,8 @@
+module Katello
+  class SmartProxySyncHistory < Katello::Model
+    self.table_name = 'katello_smart_proxy_sync_history'
+
+    belongs_to :smart_proxy, :class_name => "::SmartProxy", :dependent => :destroy, :inverse_of => :smart_proxy_sync_histories
+    belongs_to :repository, :class_name => "Katello::Repository", :dependent => :destroy, :inverse_of => :smart_proxy_sync_histories
```
jlsherrill, line 6: I think you need to remove these :dependent => :destroy blocks.  I manually destroyed the sync history in the rails console and it blew away my repos and smart proxy ;)

jlsherrill:
one small comment about the dependent destroy, but otherwise this is good2go.  We may want to hold for the larger scale testing before merging though

app/models/katello/smart_proxy_sync_history.rb
sjha4: Wow..definitely don't want that..:open_mouth:

jlsherrill (approved):
ACK!

sjha4:
[test katello]

sjha4:
Metrics from the sat-box:
1st run without history records:
`reproducer`/foreman_tasks/tasks/5d3a582e-6dd8-4d20-a992-9b708f6cf720 - 5 hours 22 mins
2nd run with history tracking
`reproducer`/foreman_tasks/tasks/689b225d-9af2-4667-ac08-a567c6e15670 - 11 mins

PR #8976, opened 2020-10-01
Fixes #30954 - inc update fails if cvv has mixed-Pulp content

ianballou:
Fixes the issue by limiting the repositories that need module_default copying to yum ones.

To test: incrementally update a content view version that has a Pulp 2 yum repo and a Pulp 3 non-yum repo.


jlsherrill (approved):
ACK   reproduced the issue and confirmed that this resolved it

PR #8977, opened 2020-10-02
Fixes #30961 - End point for /export_api_usable

parthaa:
This commit adds an endpoint to the content view versions controller to
verify if pulp3 yum is enabled.

### DIFF
--- a/app/controllers/katello/api/v2/content_view_versions_controller.rb
+++ b/app/controllers/katello/api/v2/content_view_versions_controller.rb
@@ -99,8 +99,9 @@ module Katello
                         :template => '../../../api/v2/content_view_version_export_histories/index')
     end
 
-    api :GET, "/content_view_versions/export_api_status", N_("true if the export api is pulp3 ready and usable")
+    api :GET, "/content_view_versions/export_api_status", N_("true if the export api is pulp3 ready and usable. This API is intended for use by hammer-cli only.")
     def export_api_status
+      ::Foreman::Deprecation.api_deprecation_warning("export_api_status is being deprecated and will be removed in a future version of Katello.")
       render json: { api_usable: SmartProxy.pulp_primary.pulp3_repository_type_support?(Katello::Repository::YUM_TYPE) }, status: :ok
     end
 
--- a/app/controllers/katello/api/v2/content_view_versions_controller.rb
+++ b/app/controllers/katello/api/v2/content_view_versions_controller.rb
@@ -99,6 +99,11 @@ module Katello
                         :template => '../../../api/v2/content_view_version_export_histories/index')
     end
 
+    api :GET, "/content_view_versions/export_api_status", N_("true if the export api is pulp3 ready and usable")
+    def export_api_status
+      render json: { api_usable: SmartProxy.pulp_primary.pulp3_repository_type_support?(Katello::Repository::YUM_TYPE) }, status: :ok
+    end
+
     api :POST, "/content_view_versions/:id/export", N_("Export a content view version")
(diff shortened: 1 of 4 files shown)


jlsherrill:
partha: i'd name it  /export_api_status      
and return  {api_usable: true}

parthaa:
``` bash
$ curl -X GET -s -k -u admin:changeme "http://$(hostname -f):3000/katello/api/content_view_versions/export_api_status" | jq
{
  "api_usable": true
}

```

parthaa:
> partha: i'd name it /export_api_status
> and return {api_usable: true}

updated

jturel:
With pulp3:

```
[vagrant@centos7-katello-devel katello]$ curl http://localhost:3000/katello/api/v2/content_view_versions/export_api_status -u admin:changeme
{"api_usable":true}
```

without pulp3:

```
[vagrant@centos7-katello-devel katello]$ curl http://localhost:3000/katello/api/v2/content_view_versions/export_api_status -u admin:changeme
{"api_usable":false}
```

Behavior looks good!

app/controllers/katello/api/v2/content_view_versions_controller.rb
```diff
@@ -99,6 +99,11 @@ def export_histories
                         :template => '../../../api/v2/content_view_version_export_histories/index')
     end
 
+    api :GET, "/content_view_versions/export_api_status", N_("true if the export api is pulp3 ready and usable")
+    def export_api_status
+      render json: { api_usable: SmartProxy.pulp_primary.pulp3_repository_type_support?(Katello::Repository::YUM_TYPE) }, status: :ok
```
jturel, line 104: There should be a usage of foreman's deprecation warning in here so that we can remove this in 4.0 (see existing examples in the code) and will be picked up by the deprecations tracker

app/controllers/katello/api/v2/content_view_versions_controller.rb
```diff
@@ -99,6 +99,11 @@ def export_histories
                         :template => '../../../api/v2/content_view_version_export_histories/index')
     end
 
+    api :GET, "/content_view_versions/export_api_status", N_("true if the export api is pulp3 ready and usable")
```
jturel, line 102: I'd personally add text in the description to the effect of "This API is intended for use by hammer-cli only" *just in case*

test/controllers/api/v2/content_view_versions_controller_test.rb
```diff
@@ -136,6 +150,15 @@ def test_export_histories_protected
       end
     end
 
+    def test_export_api_status_protected
+      allowed_perms = [@view_permission]
+      denied_perms = [@create_permission, @update_permission, @destroy_permission]
```
jturel, line 155: Should `create` be in allowed perms?

parthaa: You bring a good point. May be it should be `export_content_views` not read

parthaa: Settled for read + export

parthaa: updated

app/controllers/katello/api/v2/content_view_versions_controller.rb
parthaa: updated

app/controllers/katello/api/v2/content_view_versions_controller.rb
parthaa: Added

jturel (approved):
APJ!

PR #8980, opened 2020-10-05
Fixes #30994 - Add ability to set system purpose via bulk actions

chris1984:


app/lib/actions/katello/host/update_system_purpose.rb
```diff
@@ -0,0 +1,28 @@
+module Actions
+  module Katello
+    module Host
+      class UpdateSystemPurpose < Actions::EntryAction
+        def plan(host, service_level, purpose_role, purpose_usage)
+          if host.content_facet
```
jturel, line 6: Probably a bit more precise to check if the subscription_facet exists rather than content_facet, but I guess it works.

Style comment: you could do this on one line if you want: `fail ("") unless host.subscription_facet

app/lib/actions/katello/host/update_system_purpose.rb
```diff
@@ -0,0 +1,28 @@
+module Actions
+  module Katello
+    module Host
+      class UpdateSystemPurpose < Actions::EntryAction
+        def plan(host, service_level, purpose_role, purpose_usage)
```
jturel, line 5: We also need to support purpose_addons

app/lib/actions/katello/host/update_system_purpose.rb
```diff
@@ -0,0 +1,28 @@
...
+          if host.content_facet
+            host.subscription_facet.service_level = service_level
+            host.subscription_facet.purpose_role = purpose_role
+            host.subscription_facet.purpose_usage = purpose_usage
+
+            host.update_candlepin_associations
```
jturel, line 11: subscription facet has a before_update callback which calls update_candlepin_associations, so you shouldn't need to call it here. See subscription_facet_host_extensions.rb for the definition

You can also update multiple attributes at once: `host.subscription_facet.update(service_level: 'foo', purpose_role: 'bar')` etc

app/lib/actions/katello/host/update_system_purpose.rb
```diff
@@ -0,0 +1,28 @@
...
+            plan_self(:hostname => host.name)
+          else
+            fail _("Host %s has not been registered with subscription-manager.") % host.name
+          end
+        end
+
+        def humanized_name
+          if input.try(:[], :hostname).nil?
```
jturel, line 19: I don't think I'm following this code. Can the hostname be nil?

test/actions/katello/host/update_system_purpose_test.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      it 'plans' do
+        service_level = 'Standard'
+        purpose_role = 'Red Hat Enterprise Linux Server'
+        purpose_usage = 'Production'
+        action = create_action action_class
+
+        @host.expects(:update_candlepin_associations)
+
+        assert @host.subscription_facet.service_level = 'Standard'
```
jturel, line 30: This isn't exactly what you're looking for. This assertion says 'assert that I can set the service_level to 'Standard', and it's not actually testing your action. What you want to do is, after `plan_action`, assert that the service_level is equal to what was given to the action

app/lib/actions/katello/host/update_system_purpose.rb
chris1984: Updated and made it cleaner

app/lib/actions/katello/host/update_system_purpose.rb
chris1984: Updated

app/lib/actions/katello/host/update_system_purpose.rb
chris1984: After our talk on IRC kept it the way it was since we have to update purpose addons in a different way and host.save works better

app/lib/actions/katello/host/update_system_purpose.rb
chris1984: I removed this, I saw it used in other tasks, so just copied and pasted. After reading it does not make sense.

test/actions/katello/host/update_system_purpose_test.rb
chris1984: Updated the test to assert correctly after the plan action and added a 2nd test for the duplicate purpose addon we talked about

chris1984:
[ test katello]

chris1984:
[test katello]

chris1984:
[test katello]

app/lib/actions/katello/host/update_system_purpose.rb
```diff
@@ -0,0 +1,31 @@
...
+          host.subscription_facet.purpose_usage = purpose_usage
+
+          if host.subscription_facet.purpose_addons.pluck(:id) == purpose_addon_objects.pluck(:id)
+            host.subscription_facet.purpose_addons.clear
+            host.subscription_facet.purpose_addons << purpose_addon_objects
+          else
+            host.subscription_facet.purpose_addons << purpose_addon_objects
+          end
+
+          host.subscription_facet.save
```
jturel, line 21: `.save` will return false if the save fails, whereas `.save!` will raise an error. I think you'd likely want the latter here

app/lib/actions/katello/host/update_system_purpose.rb
```diff
@@ -0,0 +1,31 @@
...
+          plan_self(:hostname => host.name)
+        end
+
+        def humanized_name
+          _("Update for host %s") % input[:hostname]
```
jturel, line 26: 'Update for host <hostname>' is a little bit vague. We could have an endless amount of actions that update a host. Mentioning that it's updating system purpose would be more clear

test/actions/katello/host/update_system_purpose_test.rb
```diff
@@ -0,0 +1,51 @@
...
+      it 'plans' do
+        service_level = 'Standard'
+        purpose_role = 'Red Hat Enterprise Linux Server'
+        purpose_usage = 'Production'
+        purpose_addons = ['Test Addon']
+        action = create_action action_class
+
+        plan_action action, @host, service_level, purpose_role, purpose_usage, purpose_addons
+
+        assert_equal(@host.subscription_facet.service_level, 'Standard')
```
jturel, line 26: for these assertions, why not reference the variables you defined above instead of 'Standard' in this case

test/actions/katello/host/update_system_purpose_test.rb
```diff
@@ -0,0 +1,51 @@
...
+    before :all do
+      User.current = users(:admin)
+      @host = FactoryBot.create(:host, :with_subscription)
+    end
+
+    describe 'update system purpose' do
+      let(:action_class) { ::Actions::Katello::Host::UpdateSystemPurpose }
+
+      it 'plans' do
+        service_level = 'Standard'
```
jturel, line 18: These variables with syspurpose values look duplicated b/t the two tests. consider making them instance variables defined in the  setup method.

test/actions/katello/host/update_system_purpose_test.rb
```diff
@@ -0,0 +1,51 @@
...
+    include Dynflow::Testing
+    include Support::Actions::Fixtures
+    include FactoryBot::Syntax::Methods
+
+    before :all do
+      User.current = users(:admin)
+      @host = FactoryBot.create(:host, :with_subscription)
+    end
+
+    describe 'update system purpose' do
```
jturel, line 14: I'd really like to see us move away from this spec style of test for consistency and for debugging. For example, you can't use ktest to run just this test in the file, so it can get really cumbersome.

Do you mind reformatting the test into unit style? ie `def test_update_system_purpose` ?

app/controllers/katello/api/v2/hosts_bulk_actions_controller.rb
```diff
@@ -247,6 +247,23 @@ def resolve_traces
       render json: result
     end
 
+    api :PUT, "/hosts/bulk/system_purpose", N_("Assign system purpose attributes on one or more hosts")
+    param_group :bulk_params
+    param :service_level, String, :desc => N_("Service level of host")
+    param :purpose_role, String, :desc => N_("Role of host")
+    param :purpose_usage, String, :desc => N_("Usage of host")
+    param :purpose_addons, Array, :desc => N_("Sets the system add-ons")
+    def system_purpose
```
jturel, line 256: How does this API behave when I don't provide one of the params ? What gets set on the host, and what happens to the existing value?

IMO the behavior should be something like:
- not passing the param at all (nil value) should not modify the given syspurpose value
- passing in an empty string (or empty array in case of addons) should unset the addons

app/lib/actions/katello/host/update_system_purpose.rb
```diff
@@ -0,0 +1,31 @@
...
+        def plan(host, service_level, purpose_role, purpose_usage, purpose_addons)
+          fail _("Host %s has not been registered with subscription-manager.") % host.name unless host.subscription_facet
+
+          purpose_addon_objects = ::Katello::PurposeAddon.where(name: purpose_addons)
+
+          host.subscription_facet.service_level = service_level
+          host.subscription_facet.purpose_role = purpose_role
+          host.subscription_facet.purpose_usage = purpose_usage
+
+          if host.subscription_facet.purpose_addons.pluck(:id) == purpose_addon_objects.pluck(:id)
```
jturel, line 14: Does refactoring this whole block into the following work?

```
if purpose_addon_objects.any?
  host.subscription_facet.purpose_addons =  purpose_addon_objects
end
```

jturel (changes requested):

app/controllers/katello/api/v2/hosts_bulk_actions_controller.rb
chris1984: Would the correct place to address this be here or in the task? The task seems like the appropriate place for it.

jturel: Agree on the task - passing the params from the controller without any checking or manipulation here is ideal - can centralize that in the task

chris1984: Updated

app/lib/actions/katello/host/update_system_purpose.rb
chris1984: Updated

test/actions/katello/host/update_system_purpose_test.rb
chris1984: Updated

test/actions/katello/host/update_system_purpose_test.rb
chris1984: Updated

test/actions/katello/host/update_system_purpose_test.rb
chris1984: Updated

app/lib/actions/katello/host/update_system_purpose.rb
chris1984: Updated

app/controllers/katello/api/v2/hosts_bulk_actions_controller.rb
```diff
@@ -247,6 +247,23 @@ def resolve_traces
...
+      service_level = params[:service_level]
```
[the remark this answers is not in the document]
chris1984: Updated

app/lib/actions/katello/host/update_system_purpose.rb
chris1984: Updated

app/lib/actions/katello/host/update_system_purpose.rb
```diff
@@ -0,0 +1,27 @@
...
+    module Host
+      class UpdateSystemPurpose < Actions::EntryAction
+        def plan(host, service_level, purpose_role, purpose_usage, purpose_addons)
+          fail _("Host %s has not been registered with subscription-manager.") % host.name unless host.subscription_facet
+          purpose_addon_objects = ::Katello::PurposeAddon.where(name: purpose_addons).first_or_create! unless purpose_addons.nil?
+
+          host.subscription_facet.service_level = service_level unless service_level.nil?
+          host.subscription_facet.purpose_role = purpose_role unless purpose_role.nil?
+          host.subscription_facet.purpose_usage = purpose_usage unless purpose_usage.nil?
+          host.subscription_facet.purpose_addons.clear if purpose_addons == [] && !purpose_addons.nil?
```
jturel, line 12: I'm seeing a fair bit of conditionals related to the addons which sparked my interest. The more logic in the code -> the more execution paths thru the code -> the harder the code is to understand -> the harder the code is  to maintain -> the harder the code is to test!

Let's think about the requirements:
- if the input `purpose_addons` is nil, preserve whatever purpose addons are there on the facet. ie do nothing
- if the input `purpose addons` is an empty array, clear the purpose addons
- if the input  `purpose_addons` is an array with some items, replace the current purpose addons on the facet entirely

```

if purpose_addons
  if purpose_addons.empty?
    host.subscription_facet.purpose_addons = [] # i don't think we should use clear if we can avoid - it runs an extra query, but not sure if active record is smart enough to understand the empty array here
  else
    katello_purpose_addons = ::Katello::PurposeAddon.where(name: purpose_addons).first_or_create!
    host.subscription_facet.purpose_addons = katello_purpose_addons
  end
end
```

I'm not asking you to copy and paste what I wrote here, but feel free to use it or do something like it. I mainly wanted to illustrate how easy we can make the code to understand by keeping related things to their own block, if possible.

jturel (changes requested):

app/lib/actions/katello/host/update_system_purpose.rb
```diff
@@ -0,0 +1,31 @@
...
+          host.subscription_facet.service_level = service_level unless service_level.nil?
+          host.subscription_facet.purpose_role = purpose_role unless purpose_role.nil?
+          host.subscription_facet.purpose_usage = purpose_usage unless purpose_usage.nil?
+
+          if purpose_addons
+            if purpose_addons.empty?
+              host.subscription_facet.purpose_addons = []
+            else
+              purpose_addon_objects = ::Katello::PurposeAddon.where(name: purpose_addons).first_or_create!
+              host.subscription_facet.purpose_addons << purpose_addon_objects unless purpose_addon_objects.id == host.subscription_facet.purpose_addons.pluck(:id).first
```
jturel, line 17: I think you're trying to prevent setting the purpose addons if they match what was given. I think active record might actually be smart enough to do that already in that it won't issue a query to update them. You can test that quickly in the console and SQL logging will help: `ActiveRecord::Base.logger = Logger.new(STDOUT)
`

chris1984: @jturel 

I tried what you suggested and see the SQL, I am still getting duplicates:

```ruby
[16] pry(main)> foo.subscription_facet.purpose_addons
=> [#<Katello::PurposeAddon:0x000000000d0f2870 id: 1, name: "test">, #<Katello::PurposeAddon:0x000000000c544758 id: 1, name: "test">, #<Katello::PurposeAddon:0x000000000c544758 id: 1, name: "test">]
[17] pry(main)> 
```

jturel: Interesting! Let's pause on that for now. Have you tested adding multiple addons via the API?

jturel: Yep I think your code needs to be something more like this `addons = ['one', 'two'].map { |x| Katello::PurposeAddon.find_or_create_by(name: x) }`

jturel: I was thinking more about your optimization, and what the value of it is. In most cases, if you're using this API, there is going to be a difference in the addons  and we would run queries to confirm that only to update the DB anyway. Since we're only deleting and creating at most a handful of rows, maybe better to do without, which simplifies this code. `subscription_facet.purpose_addons = purpose_addon_objects`. What do you think?

chris1984: Can this work for multiples? I wrote a unit test and it passes:

```ruby
if purpose_addons
            if purpose_addons.empty?
              host.subscription_facet.purpose_addons = []
            else
              purpose_addons.each do |addon|
                purpose_addon_objects = ::Katello::PurposeAddon.where(name: addon).first_or_create!
              binding.pry
                host.subscription_facet.purpose_addons << purpose_addon_objects #unless purpose_addon_objects.id == host.subscription_facet.purpose_addons.pluck(:id).first
              binding.pry
              end
            end
          end
```

Test:

```ruby
def test_update_system_purpose_multiple_addons
      plan_action @action, @host, @service_level, @purpose_role, @purpose_usage, ['Addon One', 'Addon Two']

      assert_equal(@host.subscription_facet.service_level, @service_level)
      assert_equal(@host.subscription_facet.purpose_role, @purpose_role)
      assert_equal(@host.subscription_facet.purpose_usage, @purpose_usage)
      assert_equal(@host.subscription_facet.purpose_addons.count, 2)
      assert_equal(@host.subscription_facet.purpose_addons.pluck(:name).first, 'Addon One')
      assert_equal(@host.subscription_facet.purpose_addons.pluck(:name).second, 'Addon Two')
    end
```

chris1984: Still looking at the duplicate issue

chris1984: For the duplicate issue, can we keep the unless in there since activerecord is not picking up on what you suggested.

chris1984: oh nm I see you using `find_or_create`

jturel: When we do 'subscription_facet.purpose_addons = purpose_addon_objects' that does a complete replacement - so no duplicates unless those were passed into the api. so you'd want to find a way to make the array elements unique

chris1984: Ah so the `<<` is adding to it and adding the dupes compared to the `=` where it just overrides.

chris1984: > I was thinking more about your optimization, and what the value of it is. In most cases, if you're using this API, there is going to be a difference in the addons and we would run queries to confirm that only to update the DB anyway. Since we're only deleting and creating at most a handful of rows, maybe better to do without, which simplifies this code. `subscription_facet.purpose_addons = purpose_addon_objects`. What do you think?

I am fine with changing it to that, I will work on updating it and the unit tests

chris1984: @jturel updated with the new changes, we got it down to just 2 lines and are creating things that are passed in. Let me know what needs to be changed next.

test/actions/katello/host/update_system_purpose_test.rb
```diff
@@ -0,0 +1,80 @@
...
+
+      assert_equal(@host.subscription_facet.service_level, @service_level)
+      assert_equal(@host.subscription_facet.purpose_role, @purpose_role)
+      assert_equal(@host.subscription_facet.purpose_usage, @purpose_usage)
+      assert_equal(@host.subscription_facet.purpose_addons.count, 2)
+      assert_equal(@host.subscription_facet.purpose_addons.pluck(:name).first, 'Addon One')
+      assert_equal(@host.subscription_facet.purpose_addons.pluck(:name).second, 'Addon Two')
+    end
+
+    def test_update_host_duplicate_purpose
```
jturel, line 41: These first three tests seem very similar to me. You could use ['Addon One', 'Addon Two', 'Addon Two'] and you'd have the same all the same assertions. Let me know if missing something

jturel (commented):
Code looks good. Will give it another test tomorrow :)

chris1984:
@jturel updated based on what we talked about and cleaned up the duplicate tests :)

app/lib/actions/katello/host/update_system_purpose.rb
```diff
@@ -0,0 +1,31 @@
...
+    module Host
+      class UpdateSystemPurpose < Actions::EntryAction
+        def plan(host, service_level, purpose_role, purpose_usage, purpose_addons)
+          fail _("Host %s has not been registered with subscription-manager.") % host.name unless host.subscription_facet
+
+          host.subscription_facet.service_level = service_level unless service_level.nil?
+          host.subscription_facet.purpose_role = purpose_role unless purpose_role.nil?
+          host.subscription_facet.purpose_usage = purpose_usage unless purpose_usage.nil?
+
+          if purpose_addons
```
jturel, line 12: Do the tests still pass if this is consolidated:

```
if purpose_addons
  purpose_addon_objects = purpose_addons.reject(&:blank?).uniq.map.....
  host.subscription_facet.purpose_addons = purpose_addon_objects
end

chris1984: Yeah they do, will clean it up to that +1

chris1984: Pushed changes

jturel (approved):
APJ! Very nice.

PR #8983, opened 2020-10-09
Fixes #24166 - Add requiredTags to custom repos

jeremylenz:
This PR allows you to add __OS Versions__ to repositories in custom products.  On RHEL systems, this will automatically disable the repo on any system that doesn't have a matching "provided tag."

### Repo edit screen
no_restrict

### New repository screen
new_repo_norestrict

## How it works

Consider a custom product with custom repository.  Traditionally these custom repos are always enabled by default.  With this new feature, however, repos can be disabled for certain OS versions via Candlepin's `requiredTags` field.

#### Example with RHEL 7 repo and RHEL 7 host
1. A custom repo is created with "Restrict to OS Version" set to "Red Hat Enterprise Linux 7"
2. This causes the Candlepin product to be updated with `requiredTags` set to `rhel-7`
3. On a RHEL 7 host, there exists a product certificate in `/etc/pki/product` which has "provided tags" `rhel-7,rhel-7-server`
4. Since the host provides all of the required tags, the repo is enabled by default.

#### Example with RHEL 7 repo and RHEL 6 host
1. A custom repo is created with "Restrict to OS Version" set to "Red Hat Enterprise Linux 7"
2. This causes the Candlepin product to be updated with `requiredTags` set to `rhel-7`
3. On a RHEL 6 host, there exists a product certificate in `/etc/pki/product` which has "provided tags" `rhel-6,rhel-6-server`, but not `rhel-7`
4. Since the host does not provide the required tags, the repo is disabled.

#### Example with CentOS 7 host
1. A custom repo is created with "Restrict to OS Version" set to "Red Hat Enterprise Linux 7"
2. This causes the Candlepin product to be updated with `requiredTags` set to `rhel-7`
3. On the CentOS host, no matching product certificate exists
4. Since the host does not provide the required tags, the repo is disabled.

__Note:__ For custom repositories with _no_ OS versions set, the product will be enabled by default regardless of OS version.  This is not a change from previous behavior.

| Provided tags (on host) | Required tags (repo) | Result |
| ------------- | ------------- | ------------- |
| rhel-7, rhel-7-server | rhel-7 | Enabled
| rhel-7, rhel-7-server | (none) | Enabled
| rhel-7, rhel-7-server | rhel-8 | Disabled
| rhel-7, rhel-7-server | rhel-7, rhel-8 | Disabled*
| (none) | rhel-7 | Disabled
| (none) | rhel-7, rhel-8 | Disabled
| (none) | (none) | Enabled

As we can see, required tags are _required_; they live up to their name.  In order for the custom repo to be enabled, a content host must provide __all__ of the required tags.  Therefore, this feature will only be useful for RHEL hosts.

*You may think that adding multiple required tags would allow a repo to be enabled on multiple OS versions.  But this is not how it works.  A repo with both `rhel-7` and `rhel-8` tags would be disabled on all systems, because there are no systems that would list both `rhel-7` and `rhel-8` in 'Provided Tags'.  For this reason, this feature only sets a single required tag and does not allow multiples.

### Web UI

* Create a custom product and repo
* Set 'Restrict to OS Version'
* Save the repo
* View the repo details and verify 'Restrict to OS Version' value was saved

### Check that Candlepin is updated correctly

```ruby
# rails console
prod_all = Katello::Resources::Candlepin::Product.all('Default_Organization').select { |p| !p['productContent'].nil? }.map { |q| q['productContent'] }.flatten.uniq
prod_all.find { |p| p['content']['name'] == 'My New Repo' }

=> {"content"=>
  {"created"=>"2020-10-12T19:23:43+0000",
   "updated"=>"2020-10-19T16:22:09+0000",
   "uuid"=>"4028face7504ef7e017541ab07e60565",
   "id"=>"1602530623303",
   "type"=>"yum",
   "label"=>"Default_Organization_SomeProduct_asdf",
   "name"=>"asdf",
   "vendor"=>"Custom",
   "contentUrl"=>"/custom/SomeProduct/asdf",
   "requiredTags"=>"rhel-8,rhel-7-server,rhel-7-workstation",
   "gpgUrl"=>nil,
   "modifiedProductIds"=>[],
   "arches"=>nil,
   "requiredProductIds"=>[],
   "metadataExpire"=>1,
   "releaseVer"=>nil},
 "enabled"=>true}
```
Look for the `requiredTags` value.

### Test with a host
_Note:_ The Restrict to OS Version feature only works with RHEL hosts.  However, for testing purposes you can make a CentOS host look like a RHEL host by obtaining the RHEL product certificate in step 2 below.  On actual RHEL hosts, you can skip step 2.

1. Register a host to Katello.
2. Obtain a certificate from https://access.redhat.com/labs/rhpc/  (recommended: Red Hat Enterprise Linux Server ->  Version 7.9 -> x86_64 -> Generate)
3. Ensure your new product certificate is in `/etc/pki/product`

```
% rct cat-cert /etc/pki/product/FOO.pem
Product:
    ID: 69
    Name: Red Hat Enterprise Linux Server
    Version: 7.5
    Arch: x86_64
    Tags: rhel-7,rhel-7-server
    Brand Type: 
    Brand Name:
```

This confirms that the "Provided Tags" on your host are now `rhel-7,rhel-7-server`.

4. Save a custom product & repo with 'Restrict to OS Version' set to 'Red Hat Enterprise Linux 7'.  This will cause the Candlepin product content to have the `rhel-7` tag.
5. Ensure the host is subscribed to that product
6. `subscription-manager refresh` on the host
7.  Find and view the entitlement certificate for the custom product:
```
% rct cat-cert /etc/pki/entitlements/FOO.pem
Content:
	Type: yum
	Name: repo7
	Label: Default_Organization_SomeProduct_repo7
	Vendor: Custom
	URL: /Default_Organization/Library/custom/SomeProduct/repo7
	GPG: 
	Enabled: True
	Expires: 1
	Required Tags: rhel-7
	Arches: ALL
```

The Required Tags should match what was saved in the web UI.
 
Since the 'Required Tags' are a subset of the 'Provided Tags', `subscription-manager repos` should show the repo as enabled:
```
Repo ID:   Default_Organization_SomeProduct_repo7
Repo Name: repo7
Repo URL:  https://centos7-katello-devel.redhatlaptop.example.com/pulp/repos/Default_Organization/Library/custom/SomeProduct/repo7
Enabled:   1
```

Test again with mismatching Required Tags:

```
subscription-manager refresh; subscription-manager repos
```
The repo should be disabled (`Enabled: 0`)


app/models/katello/root_repository.rb
```diff
@@ -85,6 +85,10 @@ class RootRepository < Katello::Model
       :in => HTTP_PROXY_POLICIES,
       :message => _("must be one of the following: %s") % HTTP_PROXY_POLICIES.join(', ')
     }
+    validates :required_tags, allow_nil: true, format: {
```
parthaa, line 88: hmm I i'd think required_tags will be an array (like `:docker_tags_whitelist`). Any particular reason why you want to store it as string?

jeremylenz: That's the way Candlepin takes it in - a comma-separated string with no spaces.  It's stored in that format both in the product certificate and in the `requiredTags` value in the object Candlepin returns.

parthaa: use serializable like -> https://github.com/Katello/katello/blob/master/app/models/katello/root_repository.rb#L5

jeremylenz:
```
=> {"content"=>
  {"created"=>"2020-10-12T19:41:32+0000",
   "updated"=>"2020-10-12T19:41:32+0000",
   "uuid"=>"4028face7504ef7e01751e5510ca01b0",
   "id"=>"1602531692751",
   "type"=>"yum",
   "label"=>"Default_Organization_SomeProduct_qwer",
   "name"=>"qwer",
   "vendor"=>"Custom",
   "contentUrl"=>"/custom/SomeProduct/qwer",
   "requiredTags"=>"rhel-7-workstation,rhel-7-server",
   "gpgUrl"=>nil,
   "modifiedProductIds"=>[],
   "arches"=>nil,
   "requiredProductIds"=>[],
   "metadataExpire"=>1,
   "releaseVer"=>nil},
 "enabled"=>true}

```

engines/bastion/app/assets/javascripts/bastion/components/bst-edit.directive.js
```diff
@@ -335,6 +335,74 @@ angular.module('Bastion.components')
             unbindWatcher();
         });
     }])
+    .directive('bstCheckboxList', function () {
```
jturel, line 338: What's this new component do for us?

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/new/views/new-repository.html
```diff
@@ -39,6 +39,23 @@ <h4 translate> Basic Information </h4>
         </select>
       </div>
 
+      <div
+        bst-form-group label="{{ 'Required Tags' | translate }}"
```
jturel, line 43: I'm having a premonition .... we are receiving feedback to change what this is called ... maybe we should get buy in from PM on what the call this in the user-facing sense

engines/bastion/app/assets/javascripts/bastion/components/bst-edit.directive.js
jeremylenz: I'm honestly not sure if a new component is the way to go here.  I adapted it from the existing `bst-edit-multiselect` so that I could eliminate the "Save" and "Cancel" buttons.

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/new/views/new-repository.html
jeremylenz: yes, wording is difficult as always!  Also looking for feedback on the API description

engines/bastion/app/assets/javascripts/bastion/components/bst-edit.directive.js
jeremylenz: also eliminated the "X" button that was active while editing, and eliminated the ability to exit edit mode

app/models/katello/root_repository.rb
parthaa: But that is an idiosyncrasy related from Katello to Candlepin  right and not hammer/ui -> katello ? Are you saying candlepin requiring comma separated in requiredtags forces us in someway ?  Couldnt we just `requiredTags.join(',')` ?

jeremylenz: No it's not forced, I just thought it made sense to store it in the same format as cp.  I don't really feel strongly so I'll go ahead and change it

app/models/katello/root_repository.rb
```diff
@@ -336,7 +345,7 @@ def format_arches
       property :url, String, desc: 'Returns repository source URL'
     end
     class Jail < ::Safemode::Jail
-      allow :name, :label, :docker_upstream_name, :url
+      allow :name, :label, :docker_upstream_name, :url, :required_tags
```
jeremylenz, line 348: thought it'd be nice to allow this in report templates

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/new/views/new-repository.html
jeremylenz: @jturel commented on Aha.

db/migrate/20201008204114_add_required_tags_to_katello_root_repositories.rb
```diff
@@ -0,0 +1,5 @@
+class AddRequiredTagsToKatelloRootRepositories < ActiveRecord::Migration[6.0]
+  def change
+    add_column :katello_root_repositories, :required_tags, :string
```
parthaa, line 3:
```suggestion
    add_column :katello_root_repositories, :required_tags, :text
```

https://medium.com/@songyishu.yang/increase-text-limit-in-rails-44dc254f7c0b

engines/bastion/app/assets/javascripts/bastion/components/views/bst-edit.html
```diff
@@ -14,6 +14,8 @@
     <div ng-switch-when="addRemove">
       <div ng-include="'components/views/bst-edit-add-remove-cancel.html'"></div>
     </div>
+    <div ng-switch-when="none">
```
parthaa, line 17: Interesting. Why was this needed ?

parthaa:
Code wise looks ok other than my 1 suggestion. Going to test in a few

parthaa:
Missing unit tests. Guess that is pending ?

engines/bastion/app/assets/javascripts/bastion/components/views/bst-edit.html
jeremylenz: I'm using the new `bst-checkbox-list` in the New Repository form, where the Save button is at the bottom of the form.  I didn't want it to show its own 'save' and 'cancel' buttons here, when there is another one at the bottom of the form.

jeremylenz: I suppose this div could be eliminated, but I think it's better to be explicit and cover all `buttonConfig` options.

app/controllers/katello/api/v2/repositories_controller.rb
```diff
@@ -35,14 +35,15 @@ def custom_index_relation(collection)
 
     def_param_group :repo do
       param :url, String, :desc => N_("repository source url")
+      param :required_tags, Array, :desc => N_("comma-separated list of tags used to identify whether the repository should be disabled on a client with a nonmatching release version")
```
jeremylenz, line 38: @jturel what do you think of the wording here?

Also unsure about the accuracy of "comma-separated list."  The browser sends it as an actual JSON array, yet `:docker_tags_whitelist` also says "comma-separated list," yet is an `Array`.  Is that because Hammer turns a comma-separated list into the correct format when it makes the API request?  Should it just say "list" here?

jturel: Yeah, I think 'comma-separated' is a bit redundant when we're talking about an API field that is specified as an array!

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/required-tags.service.js
```diff
@@ -0,0 +1,59 @@
...
+        this.isRequiredTagSelected = function (tag, repo) {
+            if (!repo.required_tags) {
+                return false;
+            }
+            return !!repo.required_tags.find(function (reqTag) {
+                return reqTag === tag;
+            });
+        };
+
+        this.getRequiredTagsOptions = function () {
```
jturel, line 23: What happens when someone uses the API to set a value that isn't represented here?

jeremylenz: hmm..

I think it would break things.  It wouldn't be shown as an option, and then would be overwritten when you saved the repo in the UI.

jeremylenz: @jturel good catch :)

jeremylenz: @jturel Updated to address this.  If a repo has any tags in addition to the defaults, `getRequiredTagsOptions` will now honor them and it should otherwise work like normal.

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/required-tags.service.js
```diff
@@ -0,0 +1,77 @@
...
+ * @ngdoc service
+ * @name  Bastion.products.details.repositories.service:RequiredTags
+ *
+ * @description
+ *   Helper functions for repo requiredTags.
+ */
+
+angular
+    .module('Bastion.repositories')
+    .service('RequiredTags', function () {
```
parthaa, line 12: can ya add a unit test for this? look at https://github.com/Katello/katello/blob/master/engines/bastion_katello/test/products/details/repositories/checksum.service.test.js for example

parthaa:
Other than my comment on adding an service test, LGTM code wise

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/repository-details-info.controller.js
```diff
@@ -14,194 +14,208 @@
...
-                    repository.commaTagsWhitelist = null;
+                    repository["docker_tags_whitelist"] = [];
                 }
-                Notification.setSuccessMessage(translate('Repository Saved.'));
-            }, function (response) {
-                deferred.reject(response);
-                _.each(response.data.errors, function (errorMessage) {
-                    Notification.setErrorMessage(translate("An error occurred saving the Repository: ") + errorMessage);
+                /* eslint-disable camelcase */
+                repository.required_tags = $scope.requiredTagsParam();
```
jeremylenz, line 106: Most changes shown in this file are just fixing spacing.  The only real changes are this line and adding the required tags methods at the bottom.

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/new/new-repository.controller.js
```diff
@@ -17,144 +17,156 @@
...
+                $scope.architecture = results;
+                $scope.repository.arch = results[0].id;
+            });
+
+            $scope.save = function (repository) {
+                if (repository.content_type === 'ostree') {
+                    repository.unprotected = false;
+                }
+                if (repository.content_type === 'yum') {
+                    repository.required_tags = $scope.requiredTagsParam();
```
jeremylenz, line 127: same here, I don't know why the diff is so weird.  Only real changes to this file are here and the requiredTags methods

app/models/katello/root_repository.rb
```diff
@@ -214,6 +221,20 @@ def ensure_valid_docker_tags_whitelist
       end
     end
 
+    def ensure_valid_required_tags
+      return if required_tags.all?(&:blank?)
```
jturel, line 225: Should this be `required_tags.empty?`

app/models/katello/root_repository.rb
```diff
@@ -214,6 +221,20 @@ def ensure_valid_docker_tags_whitelist
...
+      # A host must provide ALL required tags in order for the repo to be enabled.
+      # So required_tags such as ['rhel-7', 'rhel-8'] is not allowed, since the repo would always be disabled.
+      if required_tags.length > 1
+        errors.add(:required_tags, N_("incompatible: tags must all belong to the same OS.  Try again with a single required tag."))
```
jturel, line 229: What do you think about "repositories can only have one required tag"?

app/models/katello/root_repository.rb
```diff
@@ -214,6 +221,20 @@ def ensure_valid_docker_tags_whitelist
...
+      end
+      required_tags.each do |tag|
+        unless ALLOWED_REQUIRED_TAGS.include?(tag)
+          errors.add(:required_tags, N_("must be one of: #{ALLOWED_REQUIRED_TAGS.join(', ')}"))
```
jturel, line 233: Format string should be used here ie `N_("must be one of: %s") % ALLOWED_REQUIRED_TAGS....`

app/controllers/katello/api/v2/repositories_controller.rb
```diff
@@ -35,14 +35,15 @@ def custom_index_relation(collection)
 
     def_param_group :repo do
       param :url, String, :desc => N_("repository source url")
+      param :required_tags, Array, :desc => N_("Identifies whether the repository should be disabled on a client with a non-matching OS version. Pass [] to enable regardless of OS version. Maximum length 1; allowed tags are: #{Katello::RootRepository::ALLOWED_REQUIRED_TAGS.join(', ')}")
```
jturel, line 38: I'd rather not call this parameter required_tags, which is candlepin-centric and not really matching up with how we identify it in the UI. Something like `os_releases` would be more appropriate imo

app/models/katello/root_repository.rb
```diff
@@ -3,6 +3,7 @@ class RootRepository < Katello::Model
     audited :except => [:content_id]
     serialize :ignorable_content
     serialize :docker_tags_whitelist
+    serialize :required_tags
```
jturel, line 6: Similar to API I think this is too candlepin-centric

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/new/new-repository.controller.js
jturel: thank goodness for 'Hide whitespace changes'

app/models/katello/root_repository.rb
jeremylenz: Needed to account for both `['']` and `[]` which this does in a single line.  (I think angular may be sending the former sometimes)

jturel: I think [''] is invalid

jeremylenz: retested and I'm not sending any `['']` any more.  So changing to `.empty?`.

app/controllers/katello/api/v2/repositories_controller.rb
jeremylenz: per discussion, going with `os_versions`

app/services/katello/product_content_importer.rb
```diff
@@ -10,7 +10,7 @@ class ProductContentImporter
     #       "name":"Red Hat Satellite 5.7 (ISOs)",
     #       "vendor":"Red Hat",
     #       "contentUrl":"/content/dist/rhel/server/6/$releasever/$basearch/satellite/5.7/iso",
-    #       "requiredTags":"rhel-6-server",
+    #       "osVersions":"rhel-6",
```
jturel, line 13: I think this was an unintentional sed replace?

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/views/repository-info.html
```diff
@@ -36,14 +36,29 @@ <h4 translate>Basic Information</h4>
     <h4 translate>Sync Settings</h4>
     <dl class="dl-horizontal dl-horizontal-left">
       <span ng-show="repository.content_type === 'yum'" >
-        <dt translate>Restrict to architecture</dt>
+        <dt>
+          <span translate>Restrict to architecture</span>
+          <i class="pficon-info" title="{{ 'The repository will be enabled by default on content hosts with the selected architecture. Select \'Default\' to enable regardless of architecture.' | translate }}"></i>
```
jturel, line 41: What do you think about making the current 'Default' option 'No restriction' instead? IMO that is more clear than Default, and removes the need to explain that in the tooltip (Select Default ...) Likewise for OS versions

app/models/katello/root_repository.rb
```diff
@@ -214,6 +221,21 @@ def ensure_valid_docker_tags_whitelist
...
+      # A host must provide ALL required tags in order for the repo to be enabled.
+      # So os_versions such as ['rhel-7', 'rhel-8'] is not allowed, since the repo would always be disabled.
+      if os_versions.length > 1
+        errors.add(:os_versions, N_("invalid: Repositories can only require one OS version."))
+      end
+      os_versions.each do |tag|
+        unless ALLOWED_OS_VERSIONS.include?(tag)
+          errors.add(:os_versions, N_("must be one of: %s" % ALLOWED_OS_VERSIONS.join(', ')))
+        end
+      end
```
jturel, line 236: how about enhancing this validation to check the content_type for yum?

```
[14] pry(main)> docker_root.content_type
=> "docker"
[15] pry(main)> docker_root.os_versions = ['rhel-7']
=> ["rhel-7"]
[16] pry(main)> docker_root.valid?
=> true
```

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/views/repository-info.html
```diff
@@ -36,14 +36,29 @@ <h4 translate>Basic Information</h4>
...
             on-save="save(repository)">
         </dd>
       </span>
 
+      <span ng-show="repository.content_type === 'yum'" >
+        <dt>
+          <span translate>Restrict to <br />OS version</span>
+          <i class="pficon-info" title="{{ 'The repository will be enabled by default on content hosts with the selected OS version.  Select \'Default\' to enable regardless of OS version.' | translate }}"></i>
+        </dt>
+        <dd bst-edit-select="repository.os_versions.length ? formatOSVersions() : 'Default'"
```
jturel, line 55: Is it feasible to get the UI to continue showing the full OS name rather than the tag value after I made a selection?

jturel (changes requested):
Looking good!

app/services/katello/product_content_importer.rb
jeremylenz: i love that you assume I used `sed` :laughing:

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/views/repository-info.html
jeremylenz: I do like 'No restriction' as a name.  But I do not think rjerrido will appreciate his tooltip going away completely.

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/views/repository-info.html
jeremylenz: Sure it's feasible.   Showing the tag value was on purpose though.  It's more compact on screen and matches the value you'd have to send if you were using hammer or the API.  Saving it with 'Red Hat Enterprise Linux 7' and then seeing it saved as 'rhel-7' is a good way for me as a user to grok that that's the value that is actually sent, which I imagine is useful for users of the CLI/API.  :man_shrugging:

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/views/repository-info.html
jeremylenz: what if we change it to 'No Restriction' and keep the tooltips, and just delete the 'Select Default...' sentence?  Is that what you were thinking?

jturel: Just what I was thinking!

app/services/katello/product_content_importer.rb
jturel: Is there any other way? ;)

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/views/repository-info.html
jturel: I guess I can roll with that!

jeremylenz:
@jturel updated! :)

app/lib/actions/katello/product/content_create.rb
```diff
@@ -8,12 +8,13 @@ def plan(root)
...
-                                           type:        root.content_type,
-                                           arches:      root.format_arches,
-                                           label:       root.custom_content_label,
-                                           content_url: root.custom_content_path)
+                                           owner:         root.product.organization.label,
+                                           name:          root.name,
+                                           type:          root.content_type,
+                                           arches:        root.format_arches,
+                                           label:         root.custom_content_label,
+                                           os_versions:   root.os_versions&.join(','),
```
jturel, line 16: maybe another find/replace fail?

app/lib/actions/katello/repository/update.rb
```diff
@@ -32,7 +33,9 @@ def plan(root, repo_params)
                         :gpg_key_url => repository.yum_gpg_key_url,
                         :label => content.label,
                         :type => root.content_type,
-                        :arches => root.format_arches)
+                        :arches => root.format_arches,
+                        :os_versions => root.os_versions&.join(','),
```
jturel, line 37: ditto

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/repository-details-info.controller.js
```diff
@@ -14,194 +14,208 @@
...
                 } else {
-                    repository.commaTagsWhitelist = null;
+                    repository["docker_tags_whitelist"] = [];
                 }
-                Notification.setSuccessMessage(translate('Repository Saved.'));
-            }, function (response) {
-                deferred.reject(response);
-                _.each(response.data.errors, function (errorMessage) {
-                    Notification.setErrorMessage(translate("An error occurred saving the Repository: ") + errorMessage);
+                /* eslint-disable camelcase */
```
jturel, line 105: Does this need to get re-enabled?

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/repository-details-info.controller.js
```diff
@@ -14,194 +14,208 @@
...
+                    deferred.resolve(response);
+                    if (!_.isEmpty(response["docker_tags_whitelist"])) {
+                        repository.commaTagsWhitelist = repository["docker_tags_whitelist"].join(", ");
+                    } else {
+                        repository.commaTagsWhitelist = null;
+                    }
+                    Notification.setSuccessMessage(translate('Repository Saved.'));
+                }, function (response) {
+                    deferred.reject(response);
+                    Notification.setErrorMessage(translate("An error occurred saving the Repository: ") + response.data.displayMessage);
```
jturel, line 117: Was this not working as it was written before?

engines/bastion/app/assets/javascripts/bastion/components/formatters/array-to-string.filter.js
```diff
@@ -10,7 +10,13 @@
  */
 angular.module('Bastion.components.formatters').filter('arrayToString', [function () {
     return function (toFormat, separator) {
+        if (!toFormat) {
```
jturel, line 13: Is this change in use in this PR?

app/lib/actions/katello/product/content_create.rb
jeremylenz: It stays `os_versions` in the Katello action.  It only gets transformed to `requiredTags` in the Candlepin action.  I was guessing that's how you'd want it...

engines/bastion/app/assets/javascripts/bastion/components/formatters/array-to-string.filter.js
jeremylenz: This may be left over from `bst-checkbox-list` which was removed, let me check

app/lib/actions/katello/product/content_create.rb
jturel: Oops, ignore me. I thought that was going into candlepin.

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/repository-details-info.controller.js
jeremylenz: It was displaying the notification as, for example:
```
An error occurred saving the Repository: invalid: Repositories can only require one OS version.
```
With this change it displays as
```
An error occurred saving the Repository: Os versions invalid: Repositories can only require one OS version.
```

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/repository-details-info.controller.js
jeremylenz: The first question that comes to my mind is 'why aren't we just disabling the line?' The answer is that whatever cursed version of eslint is being used here doesn't seem to work with `eslint-disable-next-line`.  `eslint-disable` is used already further down on line 165, so I'd argue for just keeping the first of the two occurrences (here) and calling it good?  We use snake case far too often in these controllers to worry about disabling/enabling it each time.

jturel: Lol. Fair!

engines/bastion/app/assets/javascripts/bastion/components/formatters/array-to-string.filter.js
jeremylenz: Looks like that's the case.  Reverting.

jeremylenz:
@jturel updated :)

jturel (approved):
Simply awesome! Nice work @jeremylenz 

Works just as I'd expected, tested with a RHEL7 client

jeremylenz:
[test katello]

jeremylenz:
[test katello]

jeremylenz:
[test katello]

jeremylenz:
[test katello]

PR #8989, opened 2020-10-20
Refs #30994 - Add ability to set system purpose via bulk actions

chris1984:
This PR gives the ability to set system purpose values on content hosts via the bulk actions. 

* You can add/update/delete `Service Level` `Usage Type` `Role` and `Add ons`
* The `Add ons` selection will be hidden unless there are purpose addons present in Candlepin.
* By default the form opens up to not change anything if you were to press assign. Under that is an option to clear the values. 
* Since the `Add ons` selection is hidden, running assign on hosts will not touch the purpose addon value in the host subscription facet. 

The changes can then be seen by going into the individual content host details page.

### DIFF
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-system-purpose-modal.controller.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-system-purpose-modal.controller.js
@@ -1,0 +1,112 @@
+/**
+ * @ngdoc object
+ * @name  Bastion.content-hosts.controller:ContentHostsBulkSystemPurposeModalController
+ *
+ * @requires $scope
+ * @requires $uibModalInstance
+ * @requires HostBulkAction
+ * @requires Organization
+ * @requires CurrentOrganization
+ * @requires Notification
+ * @requires hostIds
+ *
+ * @description
+ *   A controller for providing bulk action functionality for setting system purpose values
+ */
+angular.module('Bastion.content-hosts').controller('ContentHostsBulkSystemPurposeModalController',
+    ['$scope', '$uibModalInstance', 'HostBulkAction', 'Organization', 'CurrentOrganization', 'Notification', 'hostIds',
+    function ($scope, $uibModalInstance, HostBulkAction, Organization, CurrentOrganization, Notification, hostIds) {
+
+        $scope.organization = Organization.get({id: CurrentOrganization});
+
+        $scope.purposeAddonsList = function () {
+            var defaultOptions = ['No Change', 'None (Clear)'];
+            if ($scope.organization.system_purposes && $scope.organization.system_purposes.addons) {
+                return defaultOptions.concat($scope.organization.system_purposes.addons);
+            }
+            return [];
+        };
+
+        $scope.defaultUsages = ['No change', 'None (Clear)', 'Production', 'Development/Test', 'Disaster Recovery'];
+        $scope.defaultRoles = ['No change', 'None (Clear)', 'Red Hat Enterprise Linux Server', 'Red Hat Enterprise Linux Workstation', 'Red Hat Enterprise Linux Compute Node'];
+        $scope.defaultServiceLevels = ['No change', 'None (Clear)', 'Self-Support', 'Standard', 'Premium'];
+
+        $scope.hostCount = hostIds.included.ids.length;
+
+        $scope.selectedUsages = $scope.defaultUsages[0];
+        $scope.selectedRoles = $scope.defaultRoles[0];
+        $scope.selectedServiceLevels = $scope.defaultServiceLevels[0];
+
+        $scope.selected = {
+           addons: undefined
+        };
+
+        $scope.selectedItemToParam = function (item) {
+            var mapping = {
+                "None (Clear)": "",
+                "No change": null,
+                "": []
+            };
+            if (Array.isArray(item)) {
+                return $scope.selectedAddonsToParam(item);
+            }
+            if (mapping.hasOwnProperty(item)) {
+                return mapping[item];
+            }
+            return item;
+        };
+
+        $scope.selectedAddonsToParam = function (addons) {
+            var intentOptions = ['No Change', 'None (Clear)'];
+
+            var userIntent = intentOptions.filter(function(val) {
+                return addons.indexOf(val) !== -1;
+            });
+
+            if (userIntent.length === 0) {
+                return addons;
+            }
+
+            if (userIntent.includes('No Change')) {
+                return null;
+            }
+
+            if (userIntent.includes('None (Clear)') && addons.length === 1) {
+                return [];
+            } if (userIntent.includes('None (Clear)') && addons.length > 1) {
+                addons.shift();
+                return addons;
+            }
+        };
+
+        function actionParams() {
+            var params = hostIds;
+
+            params['purpose_usage'] = $scope.selectedItemToParam($scope.selectedUsages);
+            params['purpose_role'] = $scope.selectedItemToParam($scope.selectedRoles);
+            params['purpose_addons'] = $scope.selectedItemToParam($scope.selectedAddons);
+            params['service_level'] = $scope.selectedItemToParam($scope.selectedServiceLevels);
+
+            return params;
+        }
+
+        $scope.performAction = function () {
+            HostBulkAction.systemPurpose(actionParams(), function (task) {
+            $scope.ok();
+            $scope.transitionTo('content-hosts.bulk-task', {taskId: task.id});
+        }, function (response) {
+            angular.forEach(response.data.errors, function (error) {
+                Notification.setErrorMessage(error);
+            });
+        });
+        };
+
+        $scope.ok = function () {
+            $uibModalInstance.close();
+        };
+
+        $scope.cancel = function () {
+            $uibModalInstance.dismiss('cancel');
+        };
+    }]
+);
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-system-purpose-modal.html
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-system-purpose-modal.html
@@ -1,0 +1,78 @@
+<div data-extend-template="components/views/bst-modal.html">
+  <h4 data-block="modal-header" translate>Content Host System Purpose</h4>
+
+  <div data-block="modal-body">
+    <h4 translate>Assign System Purpose:</h4>
+
+    <div class="row">
+      <div class="col-sm-12">
+        <div bst-global-notification></div>
+      </div>
+    </div>
+
+    <form name="contentHostContentForm" class="form" ng-hide="content.workingMode">
+      <div>
+        <label translate>Service Level:</label>
+
+        <select type="select"
+                ng-options="item for item in defaultServiceLevels"
+                ng-model="selectedServiceLevels">
+        </select>
+      <br /><br />
+
+        <label translate>Usage Type:</label>
+
+        <select type="select"
+                ng-options="item for item in defaultUsages"
+                ng-model="selectedRoles">
+        </select>
+        <br /><br />
+
+        <label translate>Role:</label>
+
+        <select type="select"
+                ng-options="item for item in defaultRoles"
+                ng-model="selectedUsages">
+        </select>
+        <br /><br />
+
+        <label ng-hide="!purposeAddonsList().length" translate>Add ons:</label>
+        <div class="help-block" style="text-align:left;">
+          <p translate>ctrl-click or shift-click to select multiple Add ons</p>
+        </div>
+
+        <select multiple ng-multiple="true"
+                ng-hide="!purposeAddonsList().length"
+                ng-options="item for item in purposeAddonsList()"
+                ng-model="selectedAddons">
+        </select>
+        <br /><br />
+
+      </div>
+
+      <div bst-alert="info" ng-show="showConfirm">
+        <span translate>
+          Set System Purpose values on {{ hostCount }} selected content hosts?
+        </span>
+        <div>
+          <br />
+          <button type="button" class="btn btn-primary" ng-click="showConfirm = false; performAction()" translate>Assign</button>
+        </div>
+      </div>
+
+      <button class="btn btn-primary"
+              type="button"
+              ng-hide="showConfirm"
+              ng-click="showConfirm = true;">
+        <span translate>Assign</span>
+      </button>
+
+    </form>
+  </div>
+
+  <div data-block="modal-footer">
+    <button type="button" class="btn btn-default" ng-click="cancel()" translate>
+      Cancel
+    </button>
+  </div>
+</div>
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/content-host-modal-helper.service.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/content-host-modal-helper.service.js
@@ -91,6 +91,17 @@ angular.module('Bastion.content-hosts').service('ContentHostsModalHelper', ['$ui
             });
         };
 
+        this.openSystemPurposeModal = function() {
+            $uibModal.open({
+                templateUrl: 'content-hosts/bulk/views/content-hosts-bulk-system-purpose-modal.html',
+                controller: 'ContentHostsBulkSystemPurposeModalController',
+                size: 'lg',
+                resolve: {
+                    hostIds: this.resolveFunc()
+                }
+            });
+        };
+
         this.openModuleStreamsModal = function() {
             $uibModal.open({
                 templateUrl: 'content-hosts/bulk/views/content-host-bulk-module-streams-modal.html',
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/content-hosts.controller.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/content-hosts.controller.js
@@ -160,6 +160,11 @@ angular.module('Bastion.content-hosts').controller('ContentHostsController',
             ContentHostsModalHelper.openModuleStreamsModal();
         };
 
+        $scope.openSystemPurposeModal = function () {
+            nutupane.invalidate();
+            ContentHostsModalHelper.openSystemPurposeModal();
+        };
+
         $scope.openTracesModal = function () {
             nutupane.invalidate();
             ContentHostsModalHelper.openTracesModal();
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/views/content-hosts.html
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/views/content-hosts.html
@@ -52,6 +52,10 @@
         </li>
 
         <li role="menuitem" ng-show="permitted('edit_hosts')" ng-class="{disabled: table.numSelected === 0}">
+          <a ng-click="openSystemPurposeModal()" disable-link="table.numSelected === 0" translate>Manage System Purpose</a>
+        </li>
+
+        <li role="menuitem" ng-show="permitted('edit_hosts')" ng-class="{disabled: table.numSelected === 0}">
           <a ng-click="openTracesModal()" disable-link="table.numSelected === 0" translate>Manage Host Traces</a>
         </li>
 
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/host-collections/details/host-collection-details.controller.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/host-collections/details/host-collection-details.controller.js
@@ -77,6 +77,10 @@ angular.module('Bastion.host-collections').controller('HostCollectionDetailsCont
             ContentHostsModalHelper.openSubscriptionsModal();
         };
 
+        $scope.openSystemPurposeModal = function () {
+            ContentHostsModalHelper.openSystemPurposeModal();
+        };
+
         $scope.openModuleStreamsModal = function () {
             ContentHostsModalHelper.openModuleStreamsModal();
         };
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/host-collections/details/views/host-collection-info.html
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/host-collections/details/views/host-collection-info.html
@@ -100,6 +100,12 @@
           </a>
         </li>
 
+        <li>
+          <a translate ng-click="openSystemPurposeModal()">
+            System Purpose Management
+          </a>
+        </li>
+
         <li bst-feature-flag="lifecycle_environments">
           <a translate ng-click="openEnvironmentModal()">
             Change assigned Lifecycle Environment or Content View
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/hosts/host-bulk-action.factory.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/hosts/host-bulk-action.factory.js
@@ -27,7 +27,8 @@ angular.module('Bastion.hosts').factory('HostBulkAction',
             availableIncrementalUpdates: {method: 'POST', isArray: true, params: {action: 'available_incremental_updates'}},
             moduleStreams: {method: 'POST', params: {action: 'module_streams'}},
             traces: {method: 'POST', params: {action: 'traces'}},
-            resolveTraces: {method: 'PUT', isArray: true, params: {action: 'resolve_traces'}}
+            resolveTraces: {method: 'PUT', isArray: true, params: {action: 'resolve_traces'}},
+            systemPurpose: {method: 'PUT', params: {action: 'system_purpose'}}
         });
 
     }]
--- a/engines/bastion_katello/test/content-hosts/bulk/content-hosts-bulk-system-purpose-modal.controller.test.js
+++ b/engines/bastion_katello/test/content-hosts/bulk/content-hosts-bulk-system-purpose-modal.controller.test.js
@@ -1,0 +1,105 @@
+
+describe('Controller: ContentHostsBulkSystemPurposeModalController', function() {
+    var $scope, $uibModalInstance, BulkAction, Organization, CurrentOrganization, hostIds;
+
+    beforeEach(module('Bastion.content-hosts', 'Bastion.test-mocks'));
+
+    beforeEach(inject(function($injector) {
+        BulkAction = {
+            systemPurpose: function() {}
+        };
+        $httpBackend = $injector.get('$httpBackend');
+
+        CurrentOrganization = '1';
+        hostIds = {included: {ids: [1, 2]}};
+
+        mockOrg = {
+            id: '1',
+            service_levels: ['Premium'],
+            system_purposes: {
+              roles: ['custom-role'],
+              usage: ['custom-usage'],
+              addons: ['custom-addon']
+            }
+          }
+  
+          Organization = $injector.get('Organization');
+          spyOn(Organization, 'get').and.callThrough();
+          $httpBackend.expectGET('katello/api/v2/organizations/1').respond(mockOrg);
+
+        $uibModalInstance = {
+            close: function () {},
+            dismiss: function () {}
+        };
+
+        hostIds = {included: {ids: [1, 2]}};
+    }));
+
+    beforeEach(inject(function($controller, $rootScope) {
+        $scope = $rootScope.$new();
+
+        $controller('ContentHostsBulkSystemPurposeModalController', {
+            $scope: $scope,
+            $uibModalInstance: $uibModalInstance,
+            hostIds: hostIds,
+            HostBulkAction: BulkAction,
+            Organization: Organization,
+            CurrentOrganization: CurrentOrganization
+        });
+    }));
+
+    it("checks options to show correctly", function() {
+        var serviceLevels = $scope.defaultServiceLevels;
+        var usageRoles = $scope.defaultRoles;
+        var usages = $scope.defaultUsages;
+        $httpBackend.flush();
+        var addons = $scope.purposeAddonsList();
+
+        expect(serviceLevels.length).toEqual(5);
+        expect(serviceLevels.sort()).toEqual(['No change', 'None (Clear)', 'Self-Support', 'Standard', 'Premium'].sort());
+        expect(usageRoles.length).toEqual(5);
+        expect(usageRoles.sort()).toEqual(['No change', 'None (Clear)', 'Red Hat Enterprise Linux Server', 'Red Hat Enterprise Linux Workstation', 'Red Hat Enterprise Linux Compute Node'].sort());
+        expect(usages.length).toEqual(5);
+        expect(usages.sort()).toEqual(['No change', 'None (Clear)', 'Production', 'Development/Test', 'Disaster Recovery'].sort());
+        expect(addons.length).toEqual(3);
+        expect(addons.sort()).toEqual(['No Change', 'None (Clear)', 'custom-addon'].sort());
+    });
+
+    it("should perform the function and transition", function() {
+        var params = _.extend(hostIds, {purpose_usage: null, purpose_role: null, service_level: null, purpose_addons: null});
+        spyOn(BulkAction, 'systemPurpose');
+
+        $scope.performAction();
+        spyOn($scope, "transitionTo");
+        expect(BulkAction.systemPurpose).toHaveBeenCalledWith(params, jasmine.any(Function), jasmine.any(Function));
+    });
+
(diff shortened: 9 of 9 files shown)


engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-system-purpose-modal.html
```diff
@@ -0,0 +1,80 @@
...
+
+        <select type="select"
+                readonly="denied('edit_hosts', host)"
+                ng-options="roles for roles in defaultRoles"
+                ng-model="selected.purpose_role">
+                <option value="" ng-if="false"></option>
+        </select>
+        <br /><br />
+
+        <label translate>Add ons:</label>
```
jeremylenz, line 54: Need to make sure that leaving this blank won't clear all add-ons.

jeremylenz (changes requested):
Thanks @chris1984 !

Tested and works as expected; comments below.

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-system-purpose-modal.html
```diff
@@ -0,0 +1,80 @@
...
+
+    <div bst-alert="info" ng-show="showConfirm">
+      <span translate>
+        Are you sure you want to set the System Purpose values on the content host(s) selected?
+      </span>
+      <div>
+        <button type="button" class="btn btn-default" ng-click="showConfirm = false; performAction()" translate>Yes</button>
+        <button type="button" class="btn btn-default" ng-click="showConfirm = false; cancel()" translate>No</button>
+      </div>
+    </div>
```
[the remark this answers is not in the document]
chris1984: Move the confirmation blob to the bottom of the modal. When the user clicks the first Assign button, it is replaced by the confirmation blob. 

- Updated and it does this now

The confirmation blob also covers / replaces the modal's Cancel button - Since moving it to the bottom I am not hitting this. 

- Tried with Opera and Firefox

In the confirmation blob, change the button titles to 'Assign' and 'Cancel'. The 'Assign' button should have class btn-primary and 'Cancel' should have class btn-default (in other words, reverse the default and primary). The different colors will make it more clear that 'Assign' and 'Cancel' are very different actions.

- Removed the `no` from the confirmation blob and just left 'assign' with the new button color/class. I kept cancel at the bottom right and changed the class to `default` Working on the other feed back.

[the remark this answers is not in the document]
jeremylenz: > The confirmation blob also covers / replaces the modal's Cancel button

sorry, I meant that the confirmation blob _should_ cover or replace the modal's Cancel button, so that you don't have two Cancel buttons on screen.  Looks like you've addressed this now, thx :slightly_smiling_face:

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-system-purpose-modal.html
jeremylenz: (to be clear, I didn't test this)

chris1984:
@jeremylenz how do these look:

Default modal load:

https://nimbusweb.me/s/share/4763985/w8xz6w071vydt3e2dnt0

Drop Downs:

https://nimbusweb.me/s/share/4763986/zhddi7ayq7yqtklxs9pk

Confirm blob:

https://nimbusweb.me/s/share/4763987/x9mx33yta4pfi348i029

I am still working on getting the logic behind the scenes to work with those new values in the array and will tackle the addons next if that looks good to you?

jeremylenz:
@chris1984 screen shots look much better!  The confirmation blob in particular is 1000% better.

(For the drop-downs, I'm assuming they will actually drop down in the right place and not next to Add-ons as in the screenshot)

chris1984:
@jeremylenz This is how we have them in the content-host page, I can make the dropdowns below the field if you want? IE

Service Level:
<dropdown>

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-system-purpose-modal.html
chris1984: I fixed this by default setting the value to `null`
https://github.com/Katello/katello/pull/8989/files#diff-36d24e3abb899cdb34ffec17f2a4458ce1baa7efeb5540fc5ba40abf3a01e87cR70

Which does not take any action from the unit test here:

https://github.com/Katello/katello/pull/8980/files#diff-19db80ef1ed1c8bebb89697baa0db9bc664be7f730d1301d874b99af864e813dR32

chris1984:
@jeremylenz updated and addressed the feedback.

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-system-purpose-modal.controller.js
```diff
@@ -0,0 +1,97 @@
...
+        function actionParams() {
+            var params = hostIds;
+
+            if ($scope.selectedUsages === "None (Clear)") {
+                params['purpose_usage'] = "";
+            } else if ($scope.selectedUsages === "No change") {
+                params['purpose_usage'] = null;
+            } else {
+                params['purpose_usage'] = $scope.selectedUsages;
+            }
```
jeremylenz, line 51: Could DRY this up a bit by having a function like
```js
$scope.selectedItemToParam = function (item) {
  var mapping = { 
     "None (Clear)": "",
     "No change": null,
  };
  if (mapping.hasOwnProperty(item) {
    return mapping[item];
  }
  return item;
}
```
Then you could just say 
```js
params['purpose_usage'] = $scope.selectedItemToParam($scope.selectedUsages);
params['purpose_role'] = $scope.selectedItemToParam($scope.selectedRoles);
```
etc.

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-system-purpose-modal.html
```diff
@@ -0,0 +1,77 @@
...
+                ng-options="item for item in purposeAddonsList"
+                ng-model="selected.addons">
+        </select>
+        <br /><br />
+
+      </div>
+
+      <div bst-alert="info" ng-show="showConfirm">
+        <span translate>
+          Are you sure you want to set the System Purpose values on the content host(s) selected?
```
jeremylenz, line 54: Forgot to say this before, but it would be super helpful if this could be a dynamic message like
```
Set System Purpose values on 422 selected content hosts?
```
```suggestion
          Set System Purpose values on {{ hostIds.length }} selected content hosts?
```

jeremylenz (commented):
Changes look really good!  Left a couple code comments

also I'm not sure how to verify system purpose addons is functioning correctly.. do you have an easy way to test that?

jturel:
@chris1984 do you mind adding a description to the PR outlining what it does and what expected outcomes should be? That can help you in terms of thinking through use cases (and identifying missed ones). It goes a long way in helping reviewers contribute and QE in verifying the feature later on!

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-system-purpose-modal.html
chris1984: I could not get that to work, using the console I was able to find it like this `hostIds.included.ids.length` but putting it in the view did not work so just made a variable to reference the count. Not sure if I am doing something wrong or just how our page behaves. Tried `{{ table.numSelected }}` as well but no luck

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-system-purpose-modal.controller.js
```diff
@@ -0,0 +1,97 @@
...
+ * @requires hostIds
+ *
+ * @description
+ *   A controller for providing bulk action functionality for setting system purpose values
+ */
+angular.module('Bastion.content-hosts').controller('ContentHostsBulkSystemPurposeModalController',
+    ['$scope', '$uibModalInstance', 'HostBulkAction', 'Organization', 'CurrentOrganization', 'Notification', 'hostIds',
+    function ($scope, $uibModalInstance, HostBulkAction, Organization, CurrentOrganization, Notification, hostIds) {
+
+        $scope.purposeAddonsCount = 0;
```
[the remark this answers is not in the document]
chris1984: How does something like this look?

```javascript
$scope.purposeAddonsList = function () {
            var addonsList = [];
            var purposeAddonsCount;

            $scope.organization = Organization.get({id: CurrentOrganization}, function(org) {
                purposeAddonsCount += org.system_purposes.addons.length;
                if (purposeAddonsCount > 0) {
                    addonsList.concat(org.system_purposes.addons);
                }
            });
            return addonsList;
        };
```

chris1984:
@jeremylenz updated.

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-system-purpose-modal.html
```diff
@@ -0,0 +1,77 @@
...
+        <label translate>Role:</label>
+
+        <select type="select"
+                readonly="denied('edit_hosts', host)"
+                ng-options="item for item in defaultRoles"
+                ng-model="selectedUsages">
+        </select>
+        <br /><br />
+
+        <label ng-hide="purposeAddonsList.length === 0" translate>Add ons:</label>
```
jeremylenz, line 41:
```suggestion
        <label ng-hide="purposeAddonsList().length === 0" translate>Add ons:</label>
```
This is a function now, so remember to call the function :wink: 

If you want to be fancy, `0` in JavaScript is falsey, so instead of `purposeAddonsList().length === 0` you could just write `!purposeAddonsList().length`.  But some prefer the non-fancy version which is more explicit.  Whatever you're more drawn to is good with me.

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-system-purpose-modal.html
```diff
@@ -0,0 +1,77 @@
...
+        <select multiple ng-multiple="true"
+                ng-hide="purposeAddonsList.length === 0"
```
jeremylenz, line 43:
```suggestion
                ng-hide="purposeAddonsList().length === 0"
```

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-system-purpose-modal.html
```diff
@@ -0,0 +1,77 @@
...
+                readonly="denied('edit_hosts', host)"
+                ng-options="item for item in purposeAddonsList"
```
jeremylenz, line 45:
```suggestion
                ng-options="item for item in purposeAddonsList()"
```

I _think_ this will work, but angular is doing some magic here..

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-system-purpose-modal.controller.js
```diff
@@ -0,0 +1,93 @@
...
+ * @description
+ *   A controller for providing bulk action functionality for setting system purpose values
+ */
+angular.module('Bastion.content-hosts').controller('ContentHostsBulkSystemPurposeModalController',
+    ['$scope', '$uibModalInstance', 'HostBulkAction', 'Organization', 'CurrentOrganization', 'Notification', 'hostIds',
+    function ($scope, $uibModalInstance, HostBulkAction, Organization, CurrentOrganization, Notification, hostIds) {
+
+        $scope.purposeAddonsList = function () {
+            var addonsList = [];
+            var purposeAddonsCount;
```
jeremylenz, line 22: This isn't used outside or inside this function, so it's not needed anymore
```suggestion
```

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-system-purpose-modal.controller.js
```diff
@@ -0,0 +1,93 @@
...
+
+            $scope.organization = Organization.get({id: CurrentOrganization}, function(org) {
+                purposeAddonsCount += org.system_purposes.addons.length;
+                if (purposeAddonsCount > 0) {
```
jeremylenz, line 26:
```suggestion
                if (org.system_purposes.addons.length > 0) {
```

jeremylenz (commented):
Almost there!  some more code comments below

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-system-purpose-modal.html
jeremylenz: it may just be my lack of knowledge of Angular.  I'll see if i can figure it out

jeremylenz (commented):
one more thing that came to mind: Shouldn't this also be added to the Host Collections page?

Depending on how much work it is, could either add to this PR or make a new issue for it.

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-system-purpose-modal.html
chris1984: Updated to the fancy way :)

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-system-purpose-modal.html
chris1984: Updated to the fancy way :)

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-system-purpose-modal.html
chris1984: Updated

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-system-purpose-modal.controller.js
chris1984: Updated

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-system-purpose-modal.controller.js
chris1984: Removed

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-system-purpose-modal.html
```diff
@@ -0,0 +1,77 @@
...
+        <label translate>Role:</label>
+
+        <select type="select"
+                readonly="denied('edit_hosts', host)"
+                ng-options="item for item in defaultRoles"
+                ng-model="selectedUsages">
+        </select>
+        <br /><br />
+
+        <label ng-hide="!purposeAddonsList.length" translate>Add ons:</label>
```
jeremylenz, line 41:
```suggestion
        <label ng-hide="!purposeAddonsList().length" translate>Add ons:</label>
```

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-system-purpose-modal.html
```diff
@@ -0,0 +1,77 @@
...
+        <select multiple ng-multiple="true"
+                ng-hide="!purposeAddonsList.length"
```
jeremylenz, line 43:
```suggestion
                ng-hide="!purposeAddonsList().length"
```

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-system-purpose-modal.controller.js
```diff
@@ -0,0 +1,93 @@
...
+                addonsList.concat($scope.organization.system_purposes.addons);
+            }
+            return addonsList;
+        };
+
+        $scope.defaultUsages = ['No change', 'None (Clear)', 'Production', 'Development/Test', 'Disaster Recovery'];
+        $scope.defaultRoles = ['No change', 'None (Clear)', 'Red Hat Enterprise Linux Server', 'Red Hat Enterprise Linux Workstation', 'Red Hat Enterprise Linux Compute Node'];
+        $scope.defaultServiceLevels = ['No change', 'None (Clear)', 'Self-Support', 'Standard', 'Premium'];
+
+        $scope.hostCount = hostIds.included.ids.length;
```
jeremylenz, line 37: :+1: :+1:

jeremylenz (commented):
love the dynamic host count!  Still need to test with addons but after that I think it's good to go

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-system-purpose-modal.controller.js
```diff
@@ -0,0 +1,93 @@
...
+        });
+
+        $scope.purposeAddonsList = function () {
+            var addonsList = [];
+
+            if ($scope.organization.system_purposes && $scope.organization.system_purposes.addons.length > 0) {
+                addonsList.concat($scope.organization.system_purposes.addons);
+            }
+            return addonsList;
+        };
```
jeremylenz, line 31: My basic JS knowledge failure here... :smile:  `concat` doesn't modify the array in place; it creates a new array.  So line 28 is evaluated and then the result thrown out, and `addonsList` stays empty.

While we're fixing it may as well simplify even further:
```suggestion
           if ($scope.organization.system_purposes && $scope.organization.system_purposes.addons) {
                return $scope.organization.system_purposes.addons;
            }
            return [];
        };
```

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-system-purpose-modal.html
jeremylenz, line 41: please add help note here "ctrl-click or shift-click to select multiple"

jeremylenz (commented):
Got the addons tested!

modal_addons

It's a bit ugly but this use case is pretty 'edgey', so I won't complain about that too much ;)

A couple things are a bit unclear to the user:
* Not selecting any addons means "no change"
* There isn't a "None (Clear)" option
* You can select multiple addons by dragging the mouse or ctrl-clicking

After resolving the comments below, selecting one or more addons works as expected.  I verified by looking at host details:
host_detail_addons

I think it'd be worth it to see how much work it would be to add None (Clear) and No Change options to be consistent with other fields.  If it's a lot we can work something out, but ideally we should be consistent here.

chris1984:
@jeremylenz let me know how this looks, I think I got all the edge cases you were describing.

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-system-purpose-modal.html
```diff
@@ -0,0 +1,81 @@
...
+                readonly="denied('edit_hosts', host)"
+                ng-options="item for item in defaultUsages"
+                ng-model="selectedRoles">
+        </select>
+        <br /><br />
+
+        <label translate>Role:</label>
+
+        <select type="select"
+                readonly="denied('edit_hosts', host)"
```
jturel, line 35: Two things are coming to mind for the readonly check:
- this modal is already not accessible unless you can edit_hosts so maybe this check is redundant?
- the `host` variable, I think, is undefined. there's a chance this is not working as you expect

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-system-purpose-modal.controller.js
```diff
@@ -0,0 +1,102 @@
...
+ * @requires HostBulkAction
+ * @requires Organization
+ * @requires CurrentOrganization
+ * @requires Notification
+ * @requires hostIds
+ *
+ * @description
+ *   A controller for providing bulk action functionality for setting system purpose values
+ */
+angular.module('Bastion.content-hosts').controller('ContentHostsBulkSystemPurposeModalController',
```
jturel, line 16: Needs test for this controller

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-system-purpose-modal.html
chris1984: prob not working, was a copy/paste from another. Will remove it.

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-system-purpose-modal.controller.js
chris1984: Will start to work on them once @jeremylenz acks it so I have a working set of things to base it off of

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-system-purpose-modal.controller.js
```diff
@@ -0,0 +1,102 @@
...
+
+            if (angular.isUndefined(params['purpose_addons'])) {
+                params['purpose_addons'] = null;
+            } else if (params['purpose_addons'].includes('No Change')) {
+                params['purpose_addons'] = null;
+            } else if (params['purpose_addons'] === "") {
+                params['purpose_addons'] = [];
+            }
+            
+            if (params['purpose_addons'].includes('None (Clear)') && params['purpose_addons'].length > 1) {
```
jeremylenz, line 76: Need to add a check here to make sure `params['purpose_addons']` is defined, in case we just set it to `null` above.  I'm getting a console error 
```
TypeError: Cannot read property 'includes' of null
```

chris1984: @jeremylenz updated and not getting the null anymore

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-system-purpose-modal.controller.js
```diff
@@ -0,0 +1,102 @@
...
+
+        function actionParams() {
+            var params = hostIds;
+
+            params['purpose_usage'] = $scope.selectedItemToParam($scope.selectedUsages);
+            params['purpose_role'] = $scope.selectedItemToParam($scope.selectedRoles);
+            params['purpose_addons'] = $scope.selectedItemToParam($scope.selectedAddons);
+            params['service_level'] = $scope.selectedItemToParam($scope.selectedServiceLevels);
+
+            if (angular.isUndefined(params['purpose_addons'])) {
```
jturel, line 68: This block seems like it could be handled in the `mapping` section of the `selectedItemToParam` method. It also has a seemingly redundant check for `No change` which should have already been set to `null` by `selectedItemToParam`

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-system-purpose-modal.controller.js
```diff
@@ -0,0 +1,102 @@
...
+                params['purpose_addons'] = null;
+            } else if (params['purpose_addons'].includes('No Change')) {
+                params['purpose_addons'] = null;
+            } else if (params['purpose_addons'] === "") {
+                params['purpose_addons'] = [];
+            }
+            
+            if (params['purpose_addons'] && params['purpose_addons'].includes('None (Clear)') && params['purpose_addons'].length > 1) {
```
jturel, line 76: Is it possible that `.includes('None (Clear)')` would ever evaluate to true at this point?

jeremylenz: yes, because `selectedItemToParam` isn't modifying `params['purpose_addons']` because it's an array

jeremylenz, line 76: Per our discussion the other day, since it's possible for the user to make illogical selections, we want to cover the following situations:
1. If both 'No change' and other addon items are selected, 'No change' wins
2. If 'No change' and 'None' are both selected, 'No change' wins 
3. If both 'None (Clear)' and others are selected, None is ignored and the others win

After thinking about it for a while, I arrived at an alternate way of thinking about it.  

#### Input

You have a multi-select with several options:
* No Change
* None (Clear)
* addon1
* addon2
* addon3

The first two options are what I would call possible user _intents_, whereas the rest of the options are actual selected addons.  

#### Outputs

We want to end up with a param that looks like one of the following:
* `[]` for `None (Clear)`
* `null` for `No Change`
* Otherwise, the actual array of add-ons

So I think it makes sense to add a `$scope.selectedAddonsToParam` which will do the following:

1. Divide the state of the checkbox into _intents_ and _selectedAddons_
2. Figure out what the user _intent_ is;
3. Use that to figure out which of the above three to return.

This would accomplish the same thing as the if-else blocks here.

We could modify `$scope.selectedItemToParam` to check for arrays and also call our new method:
```js
$scope.selectedItemToParam = function (item) {
          console.log({item})
            var mapping = {
                "None (Clear)": "",
                "No change": null,
                "": [],
            };
            if (Array.isArray(item)) {
              return $scope.selectedAddonsToParam(item);
            }
            if (mapping.hasOwnProperty(item)) {
                return mapping[item];
            }
            return item;
        };
```

jeremylenz (commented):
Works as expected, comments below

chris1984:
@jeremylenz updated. Glad I was on the same page as you as what I came up with for the function, tells me I am not a total noob with JS lol

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-system-purpose-modal.controller.js
```diff
@@ -0,0 +1,117 @@
...
+
+        $scope.organization = Organization.get({id: CurrentOrganization}, function(org) {
+            return org.system_purposes.addons.length;
+        });
+
+        $scope.purposeAddonsList = function () {
+            if ($scope.organization.system_purposes && $scope.organization.system_purposes.addons) {
+                $scope.defaultOptions = ['No Change', 'None (Clear)'];
+                $scope.addons = $scope.organization.system_purposes.addons.map(function (addon) {
+                    return addon; });
```
jeremylenz, line 28: If `$scope.defaultOptions` isn't used anywhere outside this function, it can just be a regular variable.  Same with `$scope.addons`:
```js
var defaultOptions = ['No Change', 'None (Clear)'];
var addons;
if ($scope.organization.system_purposes && $scope.organization.system_purposes.addons) {
                ...
                $scope.addOnList = defaultOptions.concat(addons);
```

actually, `addons` and `addOnList` probably aren't needed at all:  
* Doing this `.map` on `addons` would just return `$scope.organization.system_purposes.addons` unaltered, so there's no reason to do the map
* `$scope.addonList` is assigned just before returning it, so instead of `return $scope.addonList` we can just return the expression directly without assigning it to a scope property.

```js
return $scope.defaultOptions.concat($scope.organization.system_purposes.addons);

File: engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-system-purpose-modal.controller.js [hunk 21]
```diff
@@ -0,0 +1,117 @@
...
+                return $scope.selectedAddonsToParam(item);
+            }
+            if (mapping.hasOwnProperty(item)) {
+                return mapping[item];
+            }
+            return item;
+        };
+
+        $scope.selectedAddonsToParam = function (addons) {
+            var selectedIntents = ['No Change', 'None (Clear)'];
```
jeremylenz, line 65: I think `intents` or `intentOptions` is a better name for this variable.  As long as `selected` isn't in the name.

jeremylenz (commented):
Looking good!  Just a couple JavaScript cleanup comments

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-system-purpose-modal.controller.js
chris1984: Much cleaner now :)

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-system-purpose-modal.controller.js
chris1984: Went with `intentOptions `

chris1984:
Updated, starting to work on unit tests now, since the code part should be good.

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-system-purpose-modal.controller.js
```diff
@@ -0,0 +1,102 @@
...
+ * @description
+ *   A controller for providing bulk action functionality for setting system purpose values
+ */
+angular.module('Bastion.content-hosts').controller('ContentHostsBulkSystemPurposeModalController',
+    ['$scope', '$uibModalInstance', 'HostBulkAction', 'Organization', 'CurrentOrganization', 'Notification', 'hostIds',
+    function ($scope, $uibModalInstance, HostBulkAction, Organization, CurrentOrganization, Notification, hostIds) {
+
+        $scope.organization = Organization.get({id: CurrentOrganization}, function(org) {
+            return org.system_purposes.addons.length;
+        });
```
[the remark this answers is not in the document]
jeremylenz: @chris1984 remember to resolve this one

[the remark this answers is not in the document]
chris1984: Updated with latest push, thanks for catching this one I missed.

jeremylenz:
@chris1984 can you please make a Redmine issue for adding this to host collections?

chris1984:
> @chris1984 can you please make a Redmine issue for adding this to host collections?

@jeremylenz here is the issue for the host-collections:

https://projects.theforeman.org/issues/31221

chris1984:
[test katello]

chris1984:
@jeremylenz updated

jeremylenz (approved):
Tested with host collections page and works as expected.

Unit tests also look good.

nice job @chris1984!

ACK

PR #8990, opened 2020-10-20
Fixes #31115 - Old puppet modules remain after inc update

ianballou:
As of Katello 3.12 (perhaps 3.11 too), Katello does not remove old puppet module versions when incremental updating.  This causes promotion issues due to the duplicate puppet modules.  This PR searches for the older versions and removes them as part of the incremental update.

To test:

1) Get a hold of multiple versions of 2 or 3 different puppet modules
2) Upload the puppet modules to a repository
3) Create a content view with all of the older puppet modules and publish
4) Incrementally update the version you just created with the newest puppet modules
5) See that the older puppet modules are removed
6) Try promoting the incremental content view version.  This would fail before.

### DIFF
--- a/app/lib/actions/katello/content_view_version/incremental_update.rb
+++ b/app/lib/actions/katello/content_view_version/incremental_update.rb
@@ -91,7 +91,8 @@ module Actions
               sequence do
                 new_puppet_environment = plan_action(Katello::ContentViewPuppetEnvironment::Clone, old_version,
                                                    :new_version => new_content_view_version).new_puppet_environment
-                copy_action_outputs += copy_puppet_content(new_puppet_environment, content[:puppet_module_ids]) unless content[:puppet_module_ids].blank?
+                check_puppet_module_duplicates(content[:puppet_module_ids])
+                copy_action_outputs += copy_puppet_content(new_puppet_environment, content[:puppet_module_ids], old_version) unless content[:puppet_module_ids].blank?
               end
             end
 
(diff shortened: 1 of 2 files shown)


ianballou:
Updates:
1) Checking for name and author instead of just name.  There can be puppet modules with the same name.
2) Now cannot run incremental update if multiple puppet modules with the same name & author are input.
Example error output:
```
# hammer content-view version incremental-update --puppet-module-ids=3,9,10,11 --content-view-version-id=25
Adding multiple versions of the same Puppet Module is not supported by incremental update.  The following Puppet Modules have duplicate versions in the incremental update content list: ["apt-puppetlabs", "stdlib-puppetlabs"]
```

jjeffers:
> Updates:
> 
>     1. Checking for name and author instead of just name.  There can be puppet modules with the same name.
> 
>     2. Now cannot run incremental update if multiple puppet modules with the same name & author are input.
>        Example error output:
> 
> 
> ```
> # hammer content-view version incremental-update --puppet-module-ids=3,9,10,11 --content-view-version-id=25
> Adding multiple versions of the same Puppet Module is not supported by incremental update.  The following Puppet Modules have duplicate versions in the incremental update content list: ["apt-puppetlabs", "stdlib-puppetlabs"]
> ```

Works great!
```
[vagrant@centos7-hammer-devel ~]$ hammer content-view version incremental-update --content-view-version-id 20 --puppet-modules puppetdb
Warning: An error occured while loading module hammer_cli_foreman_remote_execution.
An error occurred incrementally updating the content view:
  Adding multiple versions of the same Puppet Module is not supported by incremental update.  The following Puppet Modules have duplicate versions in the incremental update content list: ["puppetdb-puppetlabs"]
```

jjeffers (approved):
Test this attempting an incremental update with mulitple versions of the same puppet module, by module id and by name. In both cases, the incremental update refused this invalid input.

When presented with a non-duplicate id list, the incremental update proceeded and the older versions of the puppet modules were not included, as desired.

ianballou:
Thanks @jjeffers! I just need to fix one rubocop issue and then I will merge.

PR #8998, opened 2020-10-22
Add a Slot to manage manifest modal

laviro:
Adding a slot to be able to extend the manage manifest modal from core/plugins,
@shimshtein @amirfefer @johnpmitsch @jturel it would be awesome if we could get this in before the upcoming code-freeze :)

What has been done here:
- Moved the two main components upper in the component
  (didn't want to create a huge refactor in this PR and move them to another files)
- Used the `Slot` component from foreman with the id of `'katello-manage-manifest-form'`
- Used the `Fill` component in the `ComponentDidUpdate` section. the same `Fill` can be used from another plugin.

### DIFF
--- a/webpack/__mocks__/foremanReact/components/common/Fill/GlobalFill.js
+++ b/webpack/__mocks__/foremanReact/components/common/Fill/GlobalFill.js
@@ -1,3 +1,0 @@
-
-const addGlobalFill = () => jest.fn();
-export default addGlobalFill;
--- a/webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
+++ b/webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import { Grid, Col, Row, Tabs, Tab, Form, FormGroup, FormControl, ControlLabel } from 'react-bootstrap';
 import { Button, Spinner } from 'patternfly-react';
 import ForemanModal from 'foremanReact/components/ForemanModal';
+import Slot from 'foremanReact/components/common/Slot';
 import { translate as __ } from 'foremanReact/common/I18n';
 import TooltipButton from '../../../components/TooltipButton';
 import { LoadingState } from '../../../components/LoadingState';
@@ -250,17 +251,17 @@ class ManageManifestModal extends Component {
                                 />
                               }
                               {canDeleteManifest &&
-                              <React.Fragment>
-                                <TooltipButton
-                                  disabled={!isManifestImported || actionInProgress}
-                                  bsStyle="danger"
-                                  onClick={this.showDeleteManifestModal}
-                                  title={__('Delete')}
-                                  tooltipId="delete-manifest-button-tooltip"
-                                  tooltipText={this.disabledTooltipText()}
-                                  tooltipPlacement="top"
-                                />
-                              </React.Fragment>
+                                <React.Fragment>
+                                  <TooltipButton
+                                    disabled={!isManifestImported || actionInProgress}
+                                    bsStyle="danger"
+                                    onClick={this.showDeleteManifestModal}
+                                    title={__('Delete')}
+                                    tooltipId="delete-manifest-button-tooltip"
+                                    tooltipText={this.disabledTooltipText()}
+                                    tooltipPlacement="top"
+                                  />
+                                </React.Fragment>
                               }
                             </div>
                             <ForemanModal title={__('Confirm delete manifest')} id={DELETE_MANIFEST_MODAL_ID}>
@@ -280,6 +281,7 @@ class ManageManifestModal extends Component {
                     </FormGroup>
                   </React.Fragment>
                 }
+                <Slot id="katello-manage-manifest-form" multi />
               </Form>
             </Tab>
           }
--- a/webpack/test-utils/react-testing-lib-wrapper.js
+++ b/webpack/test-utils/react-testing-lib-wrapper.js
@@ -5,6 +5,7 @@ import React from 'react';
 import thunk from 'redux-thunk';
 import Immutable from 'seamless-immutable';
 import { APIMiddleware, reducers as apiReducer } from 'foremanReact/redux/API';
+import { reducers as fillReducers } from 'foremanReact/components/common/Fill';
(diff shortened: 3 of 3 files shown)


laviro:
@johnpmitsch to enable the Slot&fill mechanism to actually work during test, can we use the code from Foreman, or do we need to mock the whole thing?

If it makes trouble I would just add the Slot to the bottom of the modal, how does it sounds @ShimShtein ?

johnpmitsch:
> @johnpmitsch to enable the Slot&fill mechanism to actually work during test, can we use the code from Foreman, or do we need to mock the whole thing?

Yes, you can use the code from Foreman, all Foreman code is available during testing. You can of course still mock too.

johnpmitsch:
Is there an accompanying PR that uses the Slot and Fill or is that coming later?

webpack/__mocks__/foremanReact/components/common/Slot/index.js
```diff
@@ -0,0 +1 @@
+export default jest.fn;
```
johnpmitsch, line 1: I think this has to be `() => () => {}` or maybe `()  => jest.fn`. You also don't have to mock Foreman components, have you tried not mocking it?

laviro:
> Is there an accompanying PR that uses the Slot and Fill or is that coming later?

Yes, there is a POC that adds a fill to the bottom of the modal in https://github.com/theforeman/foreman_rh_cloud/pull/399

Screenshot (21)

webpack/__mocks__/foremanReact/components/common/Slot/index.js
laviro: > Yes, you can use the code from Foreman, all Foreman code is available during testing. You can of course still mock too.

thanks, I removed those mocks as the issue wasn't there actually

laviro:
@amirfefer can you take a look on the failing tests?
I think additional setup is needed to use the [Foreman SlotsRegistry](https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/services/SlotsRegistry/index.js)

webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
```diff
@@ -22,6 +24,16 @@ class ManageManifestModal extends Component {
...
+
+    /**
+     * the Fills need to exist before adding the Slot,
+     * calling it on DidMount was too late as the Slot was already created.
+     * TODO: move these 2 components into a different file and connect them to redux,
+     * later register the fills in a global_js file or when the app launches.
+      */
+    const { ProviderDetails, SubscriptionManifest } = this;
+    addGlobalFill('katello-manage-manifest-form', 'ProviderDetails', <ProviderDetails key={1} />, 200);
+    addGlobalFill('katello-manage-manifest-form', 'SubscriptionManifest', <SubscriptionManifest key={2} />, 100);
```
laviro, line 36: @amirfefer does it makes sense to register those on the constructor before the Slot is created?
otherwise the test didn't pass as there is an error when the Slot exist before the Fills.

amirfefer (commented):
Is there a chance to just extract the fills outside of this file scope without a big refactoring? this file is enormous

laviro:
> Is there a chance to just extract the fills outside of this file scope without a big refactoring? this file is enormous

We could do that, but I thought it would be better to do as a follow-up from the risk of not breaking anything before the coming release, also it might be good to connect those components to redux and make them a standalone - so we could call the register fills from another file, maybe the global-js file.

webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
amirfefer: IMHO it doesn't seem like a best practice especially due to the enormous file size, and it requires a future refactoring.
How about just adding a slot for now to extend the manifest modal? especially due to the upcoming code freeze I think it saves us for refactoring this page atm.

laviro:
As per [comment](https://github.com/Katello/katello/pull/8998#discussion_r511946260) and also Amir and I have talked offline,
we think it might be better in this stage to just add the slot to the bottom of the modal - without touching the two other components.
and on a follow up, we should refactor this file and register the fills of the two other components in a proper way.

ShimShtein:
Well... I think because of the tight schedule, it would be acceptable to add the slot at the bottom and refactor the component later.

laviro:
ready for another review :)

amirfefer (approved):
Thanks @laviro  for a quick update :+1: 
LGTM

ShimShtein:
ACK from me too

johnpmitsch:
[test katello]

johnpmitsch (approved):
I didn't test it out with any "Fill" functionality, but I don't see it affecting anything, so APJ!

laviro:
In order to test it from another plugin or even from Foreman,
you can do something like:

```js
import { addGlobalFill } from 'foremanReact/components/common/Fill/GlobalFill';

const Test = props => <h1>Test!</h1>;

addGlobalFill(
    'katello-manage-manifest-form',
    'test-case',
    <Test/>,
    50
  );
```

johnpmitsch:
[test katello]

johnpmitsch:
@laviro jenkins seems to be having some issues, it keeps aborting the tests early, we are looking into it!

laviro:
just rebased, can you take another look?

webpack/__mocks__/foremanReact/components/common/Fill/GlobalFill.js
```diff
@@ -1,3 +0,0 @@
-
-const addGlobalFill = () => jest.fn();
-export default addGlobalFill;
```
johnpmitsch, line 3: @laviro was this file deleted before? Looks like it is still used here https://github.com/Katello/katello/blob/0c44a2fa68e8fca1a1e6b86582512b603809d6d6/webpack/fills_index.js#L2

johnpmitsch: oh right, that is from `foremanReact`, nevermind! :smiley:

laviro: yeah, mocks are not needed anymore :)

laviro:
Thanks @johnpmitsch @ShimShtein @amirfefer !

PR #8999, opened 2020-10-23
Fixes #31150 - content view component auth fixes

jlsherrill:


parthaa:
I tested it and most of it seems to work ok except for a couple of things
I have  user with the `edit_content_views` perm and a filter criteria of `composite=true`
I'd expect the following to fail
```
$ hammer -u raja -p changeme content-view component add   --component-content-view-id=8 --composite-content-view-id=9  --latest
Component added to content view.
```

looks like it doesn't do a look up on the perms for the objects being added.

jlsherrill:
updated partha to throw a 404 on the add

parthaa (approved):

jlsherrill:
[test katello]

jlsherrill:
[test katello]

PR #9001, opened 2020-10-23
Fixes #31152 - add auth for sync controller

ianballou:
It seems like the `SyncController` is all set permissions-wise.  I figured I would change the `find_*` methods to match our new naming.  Let me know if I missed anything.

### DIFF
--- a/app/controllers/katello/api/v2/sync_controller.rb
+++ b/app/controllers/katello/api/v2/sync_controller.rb
@@ -15,23 +15,23 @@ module Katello
     # used in unit tests
     def find_object
       if params.key?(:product_id)
-        @obj = find_product
+        @obj = find_syncable_product
       elsif params.key?(:repository_id)
-        @obj = find_repository
+        @obj = find_syncable_repository
       else
         fail HttpErrors::NotFound, N_("Couldn't find subject of synchronization") if @obj.nil?
       end
       @obj
     end
 
-    def find_product
+    def find_syncable_product
       fail _("Organization required") if @organization.nil?
       @product = Product.syncable.find_by_cp_id(params[:product_id], @organization)
       fail HttpErrors::NotFound, _("Couldn't find product with id '%s'") % params[:product_id] if @product.nil?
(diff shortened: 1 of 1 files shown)


app/controllers/katello/api/v2/sync_controller.rb
```diff
@@ -15,23 +15,23 @@ def index
...
       end
       @obj
     end
 
-    def find_product
+    def find_syncable_product
       fail _("Organization required") if @organization.nil?
       @product = Product.syncable.find_by_cp_id(params[:product_id], @organization)
       fail HttpErrors::NotFound, _("Couldn't find product with id '%s'") % params[:product_id] if @product.nil?
       @product
```
sjha4, line 31: We might also want to use this.
throw_resource_not_found(name: 'product', id: params[:product_id]) if @product.nil?
same for repository find.
To test, I synced a product and tried going to https://`hostname`/katello/api/v2/organizations/1/products/439917364448/sync/ which doesn't seem to work.

https://`hostname`/katello/api/v2/products/439917364448/sync?organization_id=1 seems to hit the controller but doesn't return anything.

https://`hostname`/katello/api/v2/repositories/1/sync/ returns 404..

ianballou: `https://hostname/katello/api/v2/organizations/1/products/439917364448/sync/` needs to use the string id for the organization rather than the number id

ianballou: The 404 is interesting, looks like it's trying to find something from pulp 2 even if you're using pulp 3. Looking into it.

ianballou: Oh I see, sync_status is Pulp 2 - specific.

ianballou:
Closing this since the future of this controller is uncertain.  It seems like it hasn't worked for a number of years now and no one has brought it up since.  I'm opening a separate issue to fix or delete this controller.

PR #9002, opened 2020-10-23
Fixes #31149 - updates 404 conditions in before action filters

jjeffers:


chris1984 (approved):
ACK @jjeffers Merge away!

PR #9003, opened 2020-10-23
Fixes #31146 - content view filter auth fixes

parthaa:


chris1984 (approved):
ACK, I was able to create/delete/update/view a filter with a limited permissions user account.

swadeley:
Bug 1923209 - hammer content-view filter delete fails

PR #9005, opened 2020-10-23
Fixes #31158 - fix auth for CVVs controller

jlsherrill:


app/models/katello/authorization/content_view_version.rb
```diff
@@ -10,9 +10,32 @@ def all_hosts_editable?(lifecycle_environment)
     end
 
     module ClassMethods
+      def with_content_view_scope(scope)
+        joins(:content_view).where("#{Katello::ContentView.table_name}.id" => ::Katello::ContentView.send(scope))
```
jturel, line 14: If you include the Authorizable module in here you can do this: `joins_authorized(::Katello::ContentView, :view_content_views)` for example

jturel:
left one note on authorization, but all else looks good. will test if you end up making the suggested change

app/models/katello/authorization/content_view_version.rb
jturel: @jlsherrill maybe you'd prefer the merge approach? `joins(:content_view).merge(Katello::ContentView.readable)` for example

jturel:
Needs a rebase now :-/

app/models/katello/authorization/content_view_version.rb
jlsherrill: i'm hesitant to add Authorizable due to the check_permissions_after_save hook it adds, but i like .merge()

jlsherrill:
updated and rebased!

jturel:
[test katello]

jturel (approved):
I tried to break it but I couldn't! APJ

jlsherrill:
[test katello]

PR #9006, opened 2020-10-23
Fixes #31159 - Add authorization for HostBulkActionsController

jturel:
Addresses the various resources in HostBulkActionsController and adds a new helper method `throw_each_resource_not_found`

### DIFF
--- a/app/controllers/katello/api/v2/hosts_bulk_actions_controller.rb
+++ b/app/controllers/katello/api/v2/hosts_bulk_actions_controller.rb
@@ -7,7 +7,7 @@ module Katello
     before_action :find_host_collections, :only => [:bulk_add_host_collections, :bulk_remove_host_collections]
     before_action :find_environment, :only => [:environment_content_view]
     before_action :find_content_view, :only => [:environment_content_view]
-    before_action :find_editable_hosts, :except => [:destroy_hosts, :applicable_errata, :installable_errata, :resolve_traces]
+    before_action :find_editable_hosts, :except => [:destroy_hosts, :resolve_traces]
     before_action :find_deletable_hosts, :only => [:destroy_hosts]
     before_action :find_readable_hosts, :only => [:applicable_errata, :installable_errata, :available_incremental_updates]
     before_action :find_errata, :only => [:available_incremental_updates]
--- a/app/controllers/katello/api/v2/hosts_bulk_actions_controller.rb
+++ b/app/controllers/katello/api/v2/hosts_bulk_actions_controller.rb
@@ -12,6 +12,7 @@ module Katello
     before_action :find_readable_hosts, :only => [:applicable_errata, :installable_errata, :available_incremental_updates]
     before_action :find_errata, :only => [:available_incremental_updates]
     before_action :find_organization, :only => [:add_subscriptions]
+    before_action :find_traces, :only => [:resolve_traces]
     before_action :deprecate_katello_agent, :only => [:install_content, :update_content, :remove_content]
 
     before_action :validate_content_action, :only => [:install_content, :update_content, :remove_content]
@@ -241,8 +242,7 @@ module Katello
     param_group :bulk_params
     param :trace_ids, Array, :required => true, :desc => N_("Array of Trace IDs")
     def resolve_traces
-      traces = Katello::HostTracer.resolvable.where(id: params[:trace_ids])
-      result = Katello::HostTraceManager.resolve_traces(traces)
+      result = Katello::HostTraceManager.resolve_traces(@traces)
 
       render json: result
     end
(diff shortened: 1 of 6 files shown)


app/controllers/katello/api/v2/hosts_bulk_actions_controller.rb
```diff
@@ -314,7 +314,9 @@ def find_errata
     end
 
     def find_host_collections
-      @host_collections = HostCollection.where(:id => params[:host_collection_ids])
+      throw_each_resource_not_found(name: 'host collection', expected_ids: params[:host_collection_ids]) do
```
jturel, line 317: @jlsherrill thoughts on using the block pattern for `throw_resource_not_found` ?

app/controllers/katello/concerns/api/v2/authorization.rb
```diff
@@ -42,6 +42,15 @@ def throw_resource_not_found(name: resource_name, id: params[:id])
         fail HttpErrors::NotFound, _("Could not find %{name} resource with id %{id}") % {id: id, name: name}
       end
 
+      def throw_each_resource_not_found(name:, expected_ids: [])
+        resources = yield
+        found_ids = resources.map(&:id)
+        missing_ids = expected_ids.map(&:to_i) - found_ids
+        missing_ids.each do |id|
+          throw_resource_not_found(name: name, id: id)
```
parthaa, line 50: I'd rather have all the missing ids and one error thrown.

app/controllers/katello/api/v2/hosts_bulk_actions_controller.rb
parthaa: whats the alternative? Passing the expected_ids as parameter ?

jturel:
@parthaa updated!

app/controllers/katello/api/v2/hosts_bulk_actions_controller.rb
```diff
@@ -12,6 +12,7 @@ class Api::V2::HostsBulkActionsController < Api::V2::ApiController
     before_action :find_readable_hosts, :only => [:applicable_errata, :installable_errata, :available_incremental_updates]
```
parthaa, line 12: Do you happen to know why both ```:find_editable_hosts``` and ```find_readable_hosts``` are needed for applicable and installable errata? Might be an opportunity to fix that. 

```
    before_action :find_editable_hosts, :except => [:destroy_hosts, :applicable_errata, :installable_errata, :resolve_traces]
    before_action :find_deletable_hosts, :only => [:destroy_hosts]
    before_action :find_readable_hosts, :only => [:applicable_errata, :installable_errata, :available_incremental_updates]
```

jturel: It seems totally unnecessary - removed it

app/controllers/katello/concerns/api/v2/bulk_hosts_extensions.rb
```diff
@@ -3,37 +3,41 @@ module Concerns
     module Api::V2::BulkHostsExtensions
       extend ActiveSupport::Concern
 
+      def bulk_hosts_relation(permission, org)
```
jturel, line 6: @parthaa after testing it, I don't think it makes much sense to throw error about the missing ids here. The reason being is that the scoped search or restrict parameters can alter the output and so ids which were included may ultimately be excluded. Opted to keep current behavior, but I did a few things to clean this concern up:

- fail at the start if no ids or search given
- framed a few of the `unless` into `if` for comprehension
- de-duped some code for the hosts relation
- used `merge` for the query when a scoped search is given + other improvement in that area

None of the tests broke, and I added another one for the .merge usage since we didn't have coverage in that case

app/controllers/katello/concerns/api/v2/bulk_hosts_extensions.rb
```diff
@@ -3,37 +3,41 @@ module Concerns
...
+        if bulk_params[:included][:ids].present?
+          @hosts = @hosts.where(id: bulk_params[:included][:ids])
         end
 
         if bulk_params[:included][:search]
-          search_hosts = ::Host::Managed.authorized(permission)
-          search_hosts = search_hosts.where(:organization_id => @organization.id) if @organization
-          search_hosts = search_hosts.search_for(bulk_params[:included][:search])
-          if @hosts.any?
-            @hosts = ::Host.where(id: @hosts).or(::Host.where(id: search_hosts))
```
jturel, line 23: I believe .merge gives us the same functionality here without the ugliness

parthaa (approved):
Tested in the UI. Seems to work well. APJ

PR #9008, opened 2020-10-26
Fixes #31167 - fix authorization for repository sets controller

jjeffers:
This PR attempts to shore up the authorization deficiencies in Katello in the repository sets controller. Substantial modifications were also incorporated from https://github.com/jjeffers/katello/pull/1 authored by @jturel 

The additional PR cleaned up many resource methods and simplified obtaining readable products by subscription.

### DIFF
--- a/app/controllers/katello/api/v2/repository_sets_controller.rb
+++ b/app/controllers/katello/api/v2/repository_sets_controller.rb
@@ -4,8 +4,11 @@ module Katello
 
     include Katello::Concerns::FilteredAutoCompleteSearch
 
-    before_action :find_product_or_organization
+    before_action :set_readable_product_scope, only: [:index, :show, :available_repositories, :auto_complete_search]
+    before_action :set_editable_product_scope, only: [:enable, :disable]
+    before_action :find_product
     before_action :custom_product?
+    before_action :find_organization
     before_action :find_product_content, :except => [:index, :auto_complete_search]
 
     resource_description do
@@ -102,7 +105,8 @@ module Katello
 
     def index_relation
       if @product.nil?
-        relation = @organization.product_contents.displayable
+        authorized_product_contents = Katello::ProductContent.joins(:product).merge(@product_scope)
+        relation = @organization.product_contents.merge(authorized_product_contents).displayable
       else
         relation = @product.displayable_product_contents
       end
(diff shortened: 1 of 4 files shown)


jjeffers:
[test katello]

jjeffers:
[test katello]

jjeffers:
[test katello]

jturel:
@jjeffers - I opened a PR against your branch with my feedback since I wanted to experiment with the changes: https://github.com/jjeffers/katello/pull/1

jjeffers:
[test katello]

jturel:
I was testing this out with a restricted user and the experience I had has convinced me that **the permissions on this API should be based on products rather than subscriptions**. Lengthy explanation:

As the restricted user I needed a large collection of permissions to visit the Red Hat Repositories page without warnings popping up around missing permissions. They were `view_products`, `view_subscriptions`, `view_organizations`, and `view_sync_plans `(lol). Then, in order to enable or disable a repo set, you need the `import_manifest` permission.  I couldn't imagine being an admin trying to set up a minimal user who can use this API or RH repos page to any degree. I want to make our software easier to use.

At a conceptual level, making this API based on products seems correct. Whether it's from the UI or the API, we are not thinking in a subscription sense at all, but rather in a products and repositories sense. `product_id` is a (optional) parameter of the repository sets  APIs.  `view_subscriptions` and `import_manifest` become even more confusing in that way. Moving toward sca-by-default subscriptions will be taking a back seat.

I think the product perms are just more correct. Repository sets are directly related to products and enabling or disabling repo sets creates or destroys repositories. No subscriptions in sight!

Achieving this is a small code change on top of this PR: https://github.com/jturel/katello/commit/4cba34ecc425a1c6f03f9668381f33725733e916

@jjeffers @jlsherrill what do you think of my proposal?

jjeffers:
> @jjeffers @jlsherrill what do you think of my proposal?

I think this was one of @jlsherrill's concerns as well (is the permission better based on subscriptions or products?). We really didn't have a strong lean, so I am very happy to get your input. Thank you!

jturel (approved):
APJustin :)

jjeffers:
@jlsherrill any thoughts on this? If none, I will merge, etc.

jlsherrill:
Yep i agree!!! ACK  on using product perms. Merge away!

PR #9009, opened 2020-10-26
Fixes #31169 - Fix ping if pulpcore is disabled for downstream

chris1984:


app/models/katello/ping.rb
```diff
@@ -5,20 +5,22 @@ class Ping
     PACKAGES = %w(katello candlepin pulp qpid foreman tfm hammer).freeze
 
     class << self
+      def pulpcore_enabled # for downstream 6.9, remove in 6.10
+        system('systemctl is-enabled pulpcore-api.service')
+      end
```
jlsherrill, line 10: lets add one more check here, Only check systemctl if these SETTINGS are set:

```
  :use_pulp_2_for_content_type:
    :docker: false
    :file: false
    :yum: false
```

so it'd be like:
```
if SETTING[:katello][:use_pulp_2_for_content_type] && SETTING[:katello][:use_pulp_2_for_content_type][:yum] && SETTING[:katello][:use_pulp_2_for_content_type][:docker] && SETTING[:katello][:use_pulp_2_for_content_type][:file]]  && system('systemctl is-enabled pulpcore-api.service')
```

This will 'narrow' down the scope of when we actually have to go to systemctl

jlsherrill: we want to check when these are true, updated my snippet to show that :)

chris1984: Thanks for the snippet :) Updated with that

chris1984:
@jlsherrill here is the output of the ping test on my dev box:

```shell
[vagrant@katellodevel foreman]$ ktest ../katello/test/models/ping_test.rb 
Run options: --seed 17752

# Running:

......................

Finished in 18.434147s, 1.1934 runs/s, 3.0378 assertions/s.

22 runs, 56 assertions, 0 failures, 0 errors, 0 skips
Coverage report Rcov style generated for Unit Tests to /home/vagrant/foreman/coverage/rcov
Coverage report generated for Unit Tests to /home/vagrant/foreman/coverage. 0 / 0 LOC (100.0%) covered.
```

app/models/katello/ping.rb
jlsherrill: oh shoot, we need to check the opposite, sorry for leading you wrong, so we actually want:
```
SETTING[:katello][:use_pulp_2_for_content_type].nil? || 
( !SETTING[:katello][:use_pulp_2_for_content_type][:yum] &&
  !SETTING[:katello][:use_pulp_2_for_content_type][:docker] && 
  !SETTING[:katello][:use_pulp_2_for_content_type][:file]) || 
system('systemctl is-enabled pulpcore-api.service')
```

so the logic being either:
1) use_pulp_2_for_content_type is not set at all 
OR
2) use_pulp_2_for_content_type is set to false for all types
OR
3) systemctl is-enabled pulpcore-api.service   returns enabled


Couple of other things:
* the code as you have it just does an if check, doesn't actually return anything :)  i think you can drop the if and just return the output of the large expression

chris1984: Updated, thanks for the assist and catch on the if statement :)

chris1984:
@jlsherrill updated with your feedback and here is the result of my local test

```ruby
[vagrant@katellodevel foreman]$ ktest ../katello/test/models/ping_test.rb
Run options: --seed 50342

# Running:

......................

Finished in 12.280539s, 1.7915 runs/s, 4.5601 assertions/s.

22 runs, 56 assertions, 0 failures, 0 errors, 0 skips
Coverage report Rcov style generated for Unit Tests to /home/vagrant/foreman/coverage/rcov
Coverage report generated for Unit Tests to /home/vagrant/foreman/coverage. 0 / 0 LOC (100.0%) covered.
```

chris1984:
[test katello]

app/models/katello/ping.rb
jlsherrill: This needs to be SETTINGS, not SETTING

Don't ever trust pseudo code without trying it yourself ;)

chris1984:
@jlsherrill updated and tested this time ;) Will make sure to always test given out going forward. Updated the `system` call to be `!system()` since it was not coming back true correctly. 

Here is with `pulpcore-api` enabled and default dev box setup:

```shell
[root@pulp ~]# hammer ping
database:         
    Status:          ok
    Server Response: Duration: 0ms
candlepin:        
    Status:          ok
    Server Response: Duration: 75ms
candlepin_events: 
    Status:          ok
    message:         0 Processed, 0 Failed
    Server Response: Duration: 0ms
candlepin_auth:   
    Status:          ok
    Server Response: Duration: 51ms
katello_events:   
    Status:          ok
    message:         0 Processed, 0 Failed
    Server Response: Duration: 0ms
pulp:             
    Status:          ok
    Server Response: Duration: 97ms
pulp_auth:        
    Status:          ok
    Server Response: Duration: 47ms
pulp3:            
    Status:          ok
    Server Response: Duration: 348ms
foreman_tasks:    
    Status:          ok
    Server Response: Duration: 5ms
```

Here is with `pulpcore-api` disabled and running the installer to enable pulp2 for `yum` content type.

```shell
[root@pulp ~]# hammer ping
database:         
    Status:          ok
    Server Response: Duration: 0ms
candlepin:        
    Status:          ok
    Server Response: Duration: 63ms
candlepin_events: 
    Status:          ok
    message:         0 Processed, 0 Failed
    Server Response: Duration: 0ms
candlepin_auth:   
    Status:          ok
    Server Response: Duration: 45ms
katello_events:   
    Status:          ok
    message:         0 Processed, 0 Failed
    Server Response: Duration: 0ms
pulp:             
    Status:          ok
    Server Response: Duration: 148ms
pulp_auth:        
    Status:          ok
    Server Response: Duration: 65ms
foreman_tasks:    
    Status:          ok
    Server Response: Duration: 3ms
```

chris1984:
[test katello]

chris1984:
[test katello]

jlsherrill (approved):

app/models/katello/ping.rb
```diff
@@ -5,20 +5,25 @@ class Ping
     PACKAGES = %w(katello candlepin pulp qpid foreman tfm hammer).freeze
 
     class << self
+      def pulpcore_enabled # for downstream 6.9, remove in 6.10
+        SETTINGS[:katello][:use_pulp_2_for_content_type].nil? || (!SETTINGS[:katello][:use_pulp_2_for_content_type][:yum] &&
+        !SETTINGS[:katello][:use_pulp_2_for_content_type][:docker] &&
+        !SETTINGS[:katello][:use_pulp_2_for_content_type][:file]) ||
+        system('systemctl is-enabled pulpcore-api.service')
```
evgeni, line 12: I know this is already merged, but I'd think that should fail with SELinux enabled, as I don't think Apache/Puma can run `systemctl`?

chris1984: When I was in pry, I was able to see the output of `system('systemctl is-enabled pulpcore-api.service')` and `true` `false` come back correctly.

evgeni: Was that in a prod setup with selinux in enforcing mode? Anyways, just a hunch. We'll see if that fails in prod early enough ;)

chris1984: hehe. The pry was in my devel setup. I did test it also with prod vagrant nightly without a debugger and got the ping output to change.

lzap:
Guys, we have a report from QA that SELinux prevents `systemctl` call and I am definitely not adding that into our policy. Please work on a different solution, @ekohl had some concerns about doing this check and that pulp core can be even installed on a different host.

https://bugzilla.redhat.com/show_bug.cgi?id=1967856

Thanks.

evgeni:
7mo-ago me says hi: https://github.com/Katello/katello/pull/9009#discussion_r521558073

jlsherrill:
i somehow missed this conversation 7 moths ago :(

For 6.10 this can easily be removed (and really won't ever be called anyways i don't think).  

For 6.9, the downstream requirements (that differ from upstream) made this difficult:
a) pulpcore services should be stopped after installation
b) pulpcore services should be started/enabled to run the migration
c) pulpcore services should be stopped/disabled after the migration is run
d) there should be no downtime

I think the only alternative is to change: https://github.com/theforeman/foreman_maintain/blob/master/definitions/scenarios/content.rb#L29-L31 
to do everything via the installer, which would mean that after the migration runs:

1) the pulpcore smart proxy plugin needs to be disabled
2) the smart proxy needs to be refreshed
3) the pulpcore services need to be stopped
4) have zero downtime


The installer doesn't appear to do 3) today, but i guess foreman-maintain could continue to do that after 1) & 2).   It looks like apache and the foreman-proxy services get restarted, so 4) wouldn't actually be true.  Those services restart so quickly maybe its not a big deal?  curious your thoughts @ekohl

lzap:
The `systemctl is-enabled` performs the check via DBUS from what I can tell from `strace` call, however, to make this easier for you I am okay adding a rule for symlink existence check:

```
# stat /etc/systemd/system/multi-user.target.wants/pulp_celerybeat.service
  File: ‘/etc/systemd/system/multi-user.target.wants/pulp_celerybeat.service’ -> ‘/usr/lib/systemd/system/pulp_celerybeat.service’
  Size: 47        	Blocks: 0          IO Block: 4096   symbolic link
Device: fd00h/64768d	Inode: 1076407038  Links: 1
Access: (0777/lrwxrwxrwx)  Uid: (    0/    root)   Gid: (    0/    root)
Context: system_u:object_r:systemd_unit_file_t:s0
Access: 2021-06-09 05:32:50.905911245 -0400
Modify: 2018-08-30 13:37:33.200202289 -0400
Change: 2018-08-30 13:37:33.200202289 -0400
 Birth: -
```

Adding a read rule for `systemd_unit_file_t` should not be a problem, chances are that this is already in the base policy by default and if not, I can add this temporarily.

lzap:
Added a PR into katello-selinux if you agree with that.

ekohl:
> the pulpcore smart proxy plugin needs to be disabled

An alternative is to change the plugin to determine if it's enabled at runtime via some way (place a file somewhere, call systemctl, $option3). This could be exposed as a capability. Then the scenario would refresh the features on the Smart Proxy after doing so and Katello should understand Pulp is unavailable. That would solve the downtime constraint.

Placing a file and noting it's only to be used for migrations could mean you can enable the service to be started at boot to survive restarts.

It may be a more complicated approach, but it's what I somehow assumed we'd build when we discussed this.

To be clear: I think Katello should **never** assume it's on the same system as Pulp, not even for the primary storage. That also means we shouldn't build these hacks even for a one time migration.

jlsherrill:
keep in mind that we're fixing this for katello 3.18 only (or downstream 6.9).  There's no reason for this to exist in katello 4.0 or newer. 

And i agree @ekohl having it on the smart proxy side is probably a better solution, but at this point i think we need to go with the least complex/least amount of backports required solution.

jlsherrill:
@lzap i like that approach and should be really simple to add.  I see two solutions:

1) remove the systemctl check completely for 4.0/4.1/master
2) swich to the file based check for 3.18

lzap:
Whatever you prefer, the SELinux patch was tested by Stephen so feel free to use it!

ekohl:
> And i agree @ekohl having it on the smart proxy side is probably a better solution, but at this point i think we need to go with the least complex/least amount of backports required solution.

Agreed. We should have done that back when we built it but now it's too late.

> * remove the systemctl check completely for 4.0/4.1/master
> * swich to the file based check for 3.18

That sounds good to me.

PR #9012, opened 2020-10-27
Fixes #31182 - Add Activation Keys to Global Registration UI form

stejskalleos:
Add Activation key select to Registration Form (Hosts > All Hosts > Register Host)

katello

### DIFF
--- a/app/views/foreman/hosts/_registration.html.erb
+++ b/app/views/foreman/hosts/_registration.html.erb
@@ -1,0 +1,12 @@
+<div class='form-group'>
+  <label class='col-md-2 control-label'>
+    <%= _('Activation Key(s)') %>
+    <% help = _('Activation key(s) for Subscription Manager. Required for CentOS and Red Hat Enterprise Linux. Multiple keys add separated by comma, example: key1,key2,key3.') %>
+    <a rel="popover" data-content="<%= help %>" data-trigger="focus" data-container="body" data-html="true" tabindex="-1">
+      <span class="pficon pficon-info "></span>
+    </a>
+  </label>
+  <div class='col-md-4'>
+    <%= text_field_tag 'activation_key', params[:activation_key], class: 'form-control' %>
+  </div>
+</div>
--- a/test/controllers/foreman/registration_controller_test.rb
+++ b/test/controllers/foreman/registration_controller_test.rb
@@ -1,0 +1,51 @@
+require 'katello_test_helper'
+
+class RegistrationControllerTest < ActionController::TestCase
+  def setup
+    @tax_params = { organization: taxonomies(:organization1).id, location: taxonomies(:location1).id }
+
+    setup_controller_defaults(false)
+    setup_foreman_routes
+    login_user(User.find(users(:admin).id))
+  end
+
+  def test_one_key
+    post :create, params: { activation_key: ' key1 ' }.merge(@tax_params)
+    assert_includes(assigns(:command), 'activation_key=key1')
+  end
+
+  def test_two_keys
+    post :create, params: { activation_key: ' key1 , key2 , ' }.merge(@tax_params)
+    assert_includes(assigns(:command), 'activation_key=key1%2Ckey2')
+  end
+
+  def test_formatting1
+    post :create, params: { activation_key: ' , ,,key1, , key2, , ,,,  , ,' }.merge(@tax_params)
+    assert_includes(assigns(:command), 'activation_key=key1%2Ckey2')
+  end
+
+  def test_key_with_space
+    post :create, params: { activation_key: ' key1 with space ' }.merge(@tax_params)
+    assert_includes(assigns(:command), 'activation_key=key1+with+space')
+  end
+
+  def test_keys_with_spaces
+    post :create, params: { activation_key: ' key1 with space , key2 with space ' }.merge(@tax_params)
+    assert_includes(assigns(:command), 'activation_key=key1+with+space%2Ckey2+with+space')
+  end
+
+  def test_formatting_keys
+    post :create, params: { activation_key: ' key1 with space , key2 with space ,, ,, , ' }.merge(@tax_params)
+    assert_includes(assigns(:command), 'activation_key=key1+with+space%2Ckey2+with+space')
+  end
+
+  def test_empty_key
+    post :create, params: { activation_key: '' }.merge(@tax_params)
+    refute_includes(assigns(:command), 'activation_key=')
+  end
+
+  def test_nil_key
+    post :create, params: @tax_params
+    refute_includes(assigns(:command), 'activation_key=')
+  end
+end
--- a/app/controllers/katello/concerns/registration_controller_extensions.rb
+++ b/app/controllers/katello/concerns/registration_controller_extensions.rb
@@ -1,0 +1,16 @@
+module Katello
+  module Concerns
+    module RegistrationControllerExtensions
+      extend ActiveSupport::Concern
+
+      included do
+        before_action :format_activation_key, only: [:create]
+      end
+
+      def format_activation_key
+        return if params[:activation_key].blank?
+        params[:activation_key] = params[:activation_key].split(',').map(&:strip).reject(&:blank?).join(',')
+      end
+    end
+  end
+end
--- a/lib/katello/engine.rb
+++ b/lib/katello/engine.rb
@@ -178,6 +178,7 @@ module Katello
       ::SmartProxiesController.include Katello::Concerns::SmartProxiesControllerExtensions
       ::Foreman::Plugin.fact_importer_registry.register(Katello::RhsmFactName::FACT_TYPE, Katello::RhsmFactImporter)
       ::FactParser.register_fact_parser(Katello::RhsmFactName::FACT_TYPE, Katello::RhsmFactParser)
+      ::RegistrationController.include Katello::Concerns::RegistrationControllerExtensions
 
       #Helper Extensions
       ::SmartProxiesController.class_eval do
--- a/lib/katello/plugin.rb
+++ b/lib/katello/plugin.rb
@@ -3,7 +3,7 @@ require 'katello/repository_types.rb'
 require 'katello/host_status_manager.rb'
 # rubocop:disable Metrics/BlockLength
 Foreman::Plugin.register :katello do
-  requires_foreman '>= 1.24'
+  requires_foreman '>= 2.3'
 
   sub_menu :top_menu, :content_menu, :caption => N_('Content'),
            :icon => 'fa fa-book', :after => :monitor_menu do
@@ -282,6 +282,11 @@ Foreman::Plugin.register :katello do
 
   # Extend Global Registration endpoint
   extend_allowed_registration_vars :activation_key
+
+  # Extend Global Registration UI
+  extend_page 'registration/_form' do |cx|
+    cx.add_pagelet :global_registration, name: N_('Katello'), partial: 'foreman/hosts/registration', priority: 100, id: 'katello'
+  end
 
   extend_page "smart_proxies/show" do |cx|
     cx.add_pagelet :details_content,


app/views/foreman/hosts/_registration.html.erb
```diff
@@ -0,0 +1,7 @@
+<% activation_keys = Katello::ActivationKey.authorized(:view_activation_keys).select(:name) %>
+<div class='form-group'>
+  <label class='col-md-2 control-label'><%= _('Activation Key') %></label>
+  <div class='col-md-4'>
+    <%= select_tag 'activation_key', options_from_collection_for_select(activation_keys, :name, :name, params[:activation_keys]), class: 'form-control', include_blank: '' %>
```
ares, line 5: should this be `params[:activation_key]` instead? note the singular form

ares (changes requested):
I'm not sure how katello bumps the dependency, but this requires Foreman 2.3, probably should change https://github.com/Katello/katello/blob/master/lib/katello/plugin.rb#L6

Otherwise I've tested this and works fine, just comment on the value to display selected AK properly.

ares (approved):
Tested, works well. Could someone with merge permissions please merge?

ares:
For the record, failures seems unrelated

```
[2020-10-29T11:52:38.789Z] 185 tests, 317 assertions, 0 failures, 0 errors, 0 skips
[2020-10-29T11:52:38.789Z] rake aborted!
```

jeremylenz:
[test katello]

jeremylenz (changes requested):
This is a cool feature, thanks @stejskalleos!

When registering to Katello it's possible to use multiple activation keys.  (for example, one AK for CV, another one for subs, etc.)  More info in [this post](https://access.redhat.com/blogs/1169563/posts/2867891).  That said, allowing multiple AK's in the UI here may complicate things quite a bit, particularly because ordering is important.  And it's certainly supported to register with just a single AK, but it may not match some users' workflow.  If you're feeling ambitious you could look at the `TypeAheadSelect` component or something similar.   But honestly I can also see the argument for keeping it simple here and staying with a single selection.

However I do see one issue:

If I don't select an activation key, the template inserts the subscription-manager command like this:
```
subscription-manager register --org=Default_Organization --activationkey=""
```
This would cause an error:
```
[root@katello-client-2 vagrant]# subscription-manager register --org=Default_Organization --activationkey=""
Error: Must specify an activation key
```

If an activation key is not specified, the `--activationkey` flag should be completely left out of the `subscription-manager` command.  However, if you don't use an activation key, the Foreman username and password is required during registration:

```
[root@katello-client-2 vagrant]# subscription-manager register --org=Default_Organization
Registering to: centos7-katello-devel.redhatlaptop.example.com:443/rhsm
Username: 
```

You can use the `--username` and `--password` options, but I think it would probably be simpler to make "Activation Key" a required field for RHEL and CentOS.

stejskalleos:
@jeremylenz Thanks for the comment!

> When registering to Katello it's possible to use multiple activation keys.

Wow, being honest here, I had absolutely no idea that more than one key can be used for the registration.
We definitively need to fix this, but not sure right now how to do it, because we need this PR merged ASAP (because of Sat 6.9)

> That said, allowing multiple AK's in the UI here may complicate things quite a bit, particularly because ordering is important. And it's certainly supported to register with just a single AK, but it may not match some users' workflow. If you're feeling ambitious you could look at the TypeAheadSelect component or something similar. But honestly I can also see the argument for keeping it simple here and staying with a single selection.

Would you be okay with the following fix: (?)
Let's change `activation_key(s)` field to regular `text field`, so user can specify keys like this: `key1,key2`
There is a plan to do refactoring of the form, moving from Rails helpers to React, plus do some actual UX work how to make the form more user-friendly. (This is just first proof-of-concept)
When I'll start working on it, I will change the field to `TypeAheadSelect` component with all the proper validations.
I created task for it: https://projects.theforeman.org/issues/30440, to be we won't forget it.

> If an activation key is not specified, the --activationkey flag should be completely left out of the subscription-manager command.

Since it's a template thing, I mentioned it in the https://projects.theforeman.org/issues/30440 and I'll fix it in the Foreman core project.
For now, registration will fail with nice descriptive error, user immediately knows what is happening so it's not a big issue IMHO.

> However, if you don't use an activation key, the Foreman username and password is required during registration

For security reasons we don't want to use login:password anywhere in the templates. Only `JWT authorization` tokens, like this:
```
curl -X GET "https://foreman.example.com/register?location_id=2&organization_id=1" -H 'Authorization: Bearer eyJhbGciOiJIUzI1NiJ9.eyJ1c2VyX2lkIjo0LCJpYXQiOjE2MDQzOTUzMDEsImp0aSI6IjdkYTY2NWNjYWY0MWY4YmMxZjkyMjg0Yzk3Y2U2NTdhYmM4NDk0MzNlNzM2MmUyMDhjMTg3MWE2Yjc0ZWU5YjAiLCJleHAiOjE2MDQ0MDk3MDF9.7Iuhbtfdixu-a_7JdX8T0j67MJO-9g9yCASxe3T4Oa0' | bash
```

> You can use the --username and --password options, but I think it would probably be simpler to make "Activation Key" a required field for RHEL and CentOS.

Yeah, as I mentioned above, for security reasons we don't want to do that or even encourage users to do it.

jeremylenz:
@stejskalleos I'd be ok with the free-text field as a temporary solution before the form moves to React.  (just make sure to trim whitespace between each key ;)

I agree that passing `--username` and `--password` is not a good idea here.  So, for any registrations where the template uses subscription-manager, 

* `--activationkey` will always be needed
* the second bullet point in https://projects.theforeman.org/issues/31240 is not relevant
* "Activation Key" should become a required field on this form if the user chooses RHEL or CentOS

I think the changes above would be a reasonable course of action.

stejskalleos:
@jeremylenz  Updated & rebased, added some tests for formatting

app/views/foreman/hosts/_registration.html.erb
```diff
@@ -0,0 +1,14 @@
+<% activation_keys = Katello::ActivationKey.authorized(:view_activation_keys).select(:name) %>
```
jturel, line 1: can this go in a controller?

test/controllers/foreman/registration_controller_test.rb
```diff
@@ -0,0 +1,55 @@
+require 'katello_test_helper'
+
+class RegistrationControllerTest < ActionController::TestCase
```
jturel, line 3: We tend to not write new tests in the spec style. do you mind changing these to unit style?

app/views/foreman/hosts/_registration.html.erb
```diff
@@ -0,0 +1,14 @@
...
+
+  <label class='col-md-2 control-label'>
+    <%= _('Activation Key(s)') %>
+    <% help = _('Activation key(s) for Subscription Manager. Multiple keys add separated by comma, example: key1,key2,key3') %>
+    <a rel="popover" data-content="<%= help %>" data-trigger="focus" data-container="body" data-html="true" tabindex="-1">
+      <span class="pficon pficon-info "></span>
+    </a>
+  </label>
+  <div class='col-md-4'>
+    <%= text_field_tag 'activation_key', params[:activation_key], class: 'form-control' %>
```
jeremylenz, line 12: If this is left blank it still sends
```
subscription-manager register --org=Default_Organization --activationkey=""
```

I think "Activation Key" needs to be a required field (if a relevant OS is selected).

The output from my test:
```
[root@katello-client vagrant]# curl -X GET "https://centos7-katello-devel.redhatlaptop.example.com/register?location_id=2&operatingsystem_id=1&organization_id=1" -H 'Authorization: Bearer eyJhbGciOiJIUzI1NiJ9.eyJ1c2VyX2lkIjo0LCJpYXQiOjE2MDQ2MDI0MjQsImp0aSI6ImMzMjI3ODJlOTliMzg0M2EwMzY5NTBmYmI3ZDIxNGQzM2E3NmI1N2YyYzg3ZDk5OWE0NTU1MWFiNDA1MWE5YjUiLCJleHAiOjE2MDQ2MTY4MjR9.PF-V4F08dGHdcGzXV3qiqME6s6_28GymAKNOErlpk40' | bash
100 10433    0 10433    0     0  41014      0 --:--:-- --:--:-- --:--:-- 41074
#
# Running registration
#
100  9137  100  9137    0     0  1338k      0 --:--:-- --:--:-- --:--:-- 1487k
Loaded plugins: enabled_repos_upload, fastestmirror, package_upload, product-id, search-disabled-repos, subscription-manager

This system is not registered with an entitlement server. You can use subscription-manager to register.

Examining /tmp/tmp.8XjmjPu8RK.rpm: katello-ca-consumer-centos7-katello-devel.redhatlaptop.example.com-1.0-1.noarch
/tmp/tmp.8XjmjPu8RK.rpm: does not update installed package.
Nothing to do
Uploading Enabled Repositories Report
Cannot upload enabled repos report, is this client registered?
Error: Must specify an activation key
This system is not yet registered. Try 'subscription-manager register --help' for more information.
100   207    0   164  100    43    912    239 --:--:-- --:--:-- --:--:--   916
ERROR: not_found
Host was not found by the subscription UUID: '', this can happen if the host is registered already, but not to this Foreman
```

test/controllers/foreman/registration_controller_test.rb
jeremylenz: I'd also like to see a test added asserting that the blank activation key field is handled properly.

jeremylenz (commented):
I did a couple tests of actual client registration.  The sub-man command seems to work OK with activation key, but not when left blank (see below).

jeremylenz (commented):
also I noticed that it gives me an error `Must run as root` unless I `sudo su` before running the command.  (It's not enough to add `sudo` to the command.)  Maybe the `subscription-manager` command in the script should have `sudo`..

test/controllers/foreman/registration_controller_test.rb
stejskalleos: > We tend to not write new tests in the spec style. do you mind changing these to unit style?

Changed.

> I'd also like to see a test added asserting that the blank activation key field is handled properly.

Its there:
https://github.com/Katello/katello/pull/9012/files#diff-d17fa6ea78f8ea3cee9a2327b2c863311d7e9fa82064069c714a4ab1b1cc35ecR42-R51

app/views/foreman/hosts/_registration.html.erb
stejskalleos: > If this is left blank it still sends

Yes, but since I can't change template from Katello, I'll do it in this ticket: https://projects.theforeman.org/issues/31240

> I think "Activation Key" needs to be a required field (if a relevant OS is selected).

Yes, I agree, but for now I'd like to keep it as it is and add this validation when we will refactor the form.
Added comment to https://projects.theforeman.org/issues/31240 so we won't forget it.

stejskalleos:
> also I noticed that it gives me an error Must run as root unless I sudo su before running the command. (It's not enough to add sudo to the command.) Maybe the subscription-manager command in the script should have sudo..

There is check for root user in Global Registration template, see:
https://github.com/theforeman/foreman/blob/develop/app/views/unattended/provisioning_templates/registration/global_registration.erb#L22

test/controllers/foreman/registration_controller_test.rb
```diff
@@ -1,55 +1,51 @@
...
+    setup_foreman_routes
+    login_user(User.find(users(:admin).id))
+  end
+
+  test do
+    post :create, params: { activation_key: ' key1 ' }.merge(@tax_params)
+    assert_includes(assigns(:command), 'activation_key=key1')
+  end
+
+  test do
```
jeremylenz, line 17: Also please name the tests, for example instead of 
```ruby
test do
...
end
```
instead write
```ruby
def test_activation_key_param_trims_whitespace
...
end
```

app/views/foreman/hosts/_registration.html.erb
jeremylenz: Sounds good to me (as long as one doesn't end up getting released before the other).

In the meantime maybe it's good to add 
```
Required for CentOS and Red Hat Enterprise Linux.
```
to the help text above.  (in other words, remind users that it's needed even before the validation is done)

jeremylenz (commented):
Getting closer, just one more comment below :)

stejskalleos:
@jeremylenz  @jturel Rebased & Updated

jturel (approved):
LGTM! I'll give @jeremylenz the pleasure of final ACK + merge

jeremylenz (approved):

PR #9013, opened 2020-10-28
Fixes #31189 - Add import-only flag to Content Views

jturel:
This PR introduces the idea of an import-only content view. an import-only CV is used in a downstream Katello solely for the purpose of being published when importing a CV that's been exported from a different (upstream) Katello server.

**The following restrictions are put into place when set to import-only:**

- it may not be published directly
- may not be composite
- may not do dep-solving
- may only have yum repos associated
- may not have filters associated

**Now we:**
- show import_only in the API
- show import-only in the CV table for easy identification
- can set import-only when creating a new CV, and not after creation
- hide the import-only UI aspects + api docs if yum content is not managed by pulp3

**Please especially give feedback on:**
- new messaging, api descriptions
- should there be any more restrictions?
- can you break it by somehow getting into a weird or conflicting state?

### DIFF
--- a/app/lib/actions/katello/content_view/publish.rb
+++ b/app/lib/actions/katello/content_view/publish.rb
@@ -6,10 +6,10 @@ module Actions
         include ::Katello::ContentViewHelper
         attr_accessor :version
         # rubocop:disable Metrics/MethodLength,Metrics/AbcSize,Metrics/CyclomaticComplexity
-        def plan(content_view, description = "", options = {})
+        def plan(content_view, description = "", options = {importing: false})
           action_subject(content_view)
-          import_only = options.fetch(:import_only, false)
-          content_view.check_ready_to_publish! unless import_only
+
+          content_view.check_ready_to_publish!(importing: options[:importing])
 
           if options[:repos_units].present?
             valid_labels_from_cv = content_view.repositories.map(&:label)
@@ -52,7 +52,8 @@ module Actions
 
             if separated_repo_map[:pulp3_yum].keys.flatten.present? &&
                 SmartProxy.pulp_primary.pulp3_support?(separated_repo_map[:pulp3_yum].keys.flatten.first)
-              if import_only
+
+              if options[:importing]
                 handle_import(version, options.slice(:path, :metadata))
               else
                 plan_action(Repository::MultiCloneToVersion, separated_repo_map[:pulp3_yum], version)
--- a/app/models/katello/content_view_filter.rb
+++ b/app/models/katello/content_view_filter.rb
@@ -157,6 +157,10 @@ module Katello
       if self.content_view.composite?
         errors.add(:base, _("cannot contain filters if composite view"))
       end
+
+      if self.content_view.import_only?
+        errors.add(:base, _("cannot add filter to import-only view"))
+      end
     end
 
     def validate_filter_repos(errors, content_view)
--- a/app/models/katello/content_view_repository.rb
+++ b/app/models/katello/content_view_repository.rb
@@ -7,6 +7,10 @@ module Katello
                                 Repository::DEB_TYPE
                                ].freeze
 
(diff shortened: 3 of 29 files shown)


app/lib/actions/katello/content_view/publish.rb
```diff
@@ -6,10 +6,10 @@ class Publish < Actions::EntryAction
         include ::Katello::ContentViewHelper
         attr_accessor :version
         # rubocop:disable Metrics/MethodLength,Metrics/AbcSize,Metrics/CyclomaticComplexity
-        def plan(content_view, description = "", options = {})
+        def plan(content_view, description = "", options = {importing: false})
```
parthaa, line 9: This importing option is unnecessary imo. why not just do content_view.import_only?

jturel: We need a way to identify if the publish action was called directly (via API for example) vs the CV import action in order to know whether we should allow the publish of an import_only CV. Lemme know if I'm missing the point though

parthaa: I might be missing something here.  In both cases the thing will fail if you are trying publish if your cv is import_only (I see that you already do this in the model?)
I dont know why it should matter how the import call was made..

jturel: If we don't have a flag like this, how do we allow publishing an import_only CV through CV import, but deny the same publish when performed through the ordinary publish API? Seems like we need to know when we are performing an import

parthaa: Ah good point. That makes sense.Lets keep it your way.

parthaa:
Need a better term than 'Import-only'. 
May be 'Importing' is better?

parthaa:
Details-for-Content-View-import-only

import-only cannot be unset. The idea here is that publish is takes stuff from the library and copies them on to the new version while import takes it from the dump and handles incremental updates etc. Importing content views are  purely reliant on what was created prior.

parthaa:
Tested importing and error cases. Seems to handle well +1
Just make the import-only unchangeable (similar to the composite case) + think about a different name for the feature.

app/models/katello/content_view_filter.rb
```diff
@@ -157,6 +157,10 @@ def validate_content_view
       if self.content_view.composite?
         errors.add(:base, _("cannot contain filters if composite view"))
       end
+
+      if self.content_view.import_only?
+        errors.add(:base, _("cannot add filter to import-only view"))
```
parthaa, line 162: Can you add test for this

app/models/katello/content_view_repository.rb
```diff
@@ -27,7 +31,7 @@ def content_view_composite
     end
 
     def ensure_repository_type
-      unless ALLOWED_REPOSITORY_TYPES.include?(repository.content_type)
+      unless allowed_repository_types.include?(repository.content_type)
```
parthaa, line 34: add a test for this

app/models/katello/content_view_filter.rb
jturel: See changes in test/models/content_view_filter_test.rb

app/models/katello/content_view_repository.rb
jturel: See test/models/content_view_repository_test.rb

jturel:
@parthaa updated to make import_only immutable after create

parthaa:
Can you also add validation for  `content_view_puppet_modules` .
Don't want puppet modules assigned for a cv

parthaa (approved):
nice work

jturel:
@parthaa updated to hide import-only UI + api doc as discussed - take a look! I tested that it behaves in hammer as expected

jturel:
[test katello]

parthaa:
ack

PR #9014, opened 2020-10-28
Fixes #31199 - don't require container_image_registry for registry

jlsherrill:
with pulp3


ehelms:
Tested on a EL8 with Pulp 3 only spun, this fixed `podman login`:

```
[root@pipeline-katello-server-nightly-centos8 vagrant]# podman login $HOSTNAME -u admin -p changeme
Login Succeeded!
```

This results from having dropped this section of the config during deployment of Pulp 3 only:

https://github.com/theforeman/puppet-katello/blob/master/templates/katello.yaml.erb#L39-L41

ehelms (approved):

PR #9016, opened 2020-10-29
Fixes #31214 - Inc update requires puppet content type

ianballou:
Incremental update assumes Pulp 2 & Puppet content are available on the server.  This causes issues if Pulp 2 is not installed at all.

### DIFF
--- a/app/lib/actions/katello/content_view_version/incremental_update.rb
+++ b/app/lib/actions/katello/content_view_version/incremental_update.rb
@@ -88,11 +88,15 @@ module Actions
                 end
               end
 
-              sequence do
-                new_puppet_environment = plan_action(Katello::ContentViewPuppetEnvironment::Clone, old_version,
-                                                   :new_version => new_content_view_version).new_puppet_environment
-                check_puppet_module_duplicates(content[:puppet_module_ids])
-                copy_action_outputs += copy_puppet_content(new_puppet_environment, content[:puppet_module_ids], old_version) unless content[:puppet_module_ids].blank?
+              if SmartProxy.pulp_primary.has_feature?(SmartProxy::PULP_FEATURE) && SETTINGS[:katello][:content_types][:puppet]
+                sequence do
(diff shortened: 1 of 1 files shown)


ianballou:
[test katello]

app/lib/actions/katello/content_view_version/incremental_update.rb
```diff
@@ -89,10 +89,12 @@ def plan(old_version, environments, options = {})
               end
 
               sequence do
-                new_puppet_environment = plan_action(Katello::ContentViewPuppetEnvironment::Clone, old_version,
-                                                   :new_version => new_content_view_version).new_puppet_environment
-                check_puppet_module_duplicates(content[:puppet_module_ids])
-                copy_action_outputs += copy_puppet_content(new_puppet_environment, content[:puppet_module_ids], old_version) unless content[:puppet_module_ids].blank?
+                if SmartProxy.pulp_primary.has_feature?(SmartProxy::PULP_FEATURE) && SETTINGS[:katello][:content_types][:puppet]
```
ekohl, line 92: Not too familiar with the code, but does it make sense to have the `if` outside of `sequence`?

ianballou: ah yeah, good point. I'll move it.

ekohl (approved):

app/lib/actions/katello/content_view_version/incremental_update.rb
jlsherrill: it really doesn't matter in this case.  The following is basically the same:

```
if blah
  sequence 
  do 
end
```
```
sequence do 
  if blah
  end
end
```

The first one would technically be slightly more efficient, but really it shouldn't matter

ekohl: I figured it would indeed be a micro-optimization, depending on how expensive a call to `sequence` is.

ianballou:
[test katello]

jlsherrill (approved):
ACK

reproduced the issue and confirmed this solves it

PR #9017, opened 2020-10-30
Fixes #31134 - Disable SCA toggle in UI if the organization is ineligible

jturel:
This PR makes use of the new Candlepin API introduced in https://github.com/candlepin/candlepin/pull/2829 which allows us to check if the organization in Hosted candlepin (portal) is eligible for SCA.

The React UI will call the new API in the upstream subscriptions controller, while the ERB view will call the `simple_content_access_eligible?` method on the Org's `upstream_consumer`

**I still need to test this against an actual candlepin build (linked in the kanboard card), but wanted to get the PR open for feedback on approach as the API interaction is a relatively minor detail**


app/controllers/katello/api/v2/upstream_subscriptions_controller.rb
```diff
@@ -77,15 +77,22 @@ def ping
       render json: { status: 'OK' }
     end
 
-    api :PUT, "/organizations/:organization_id/simple_content_access/enable",
+    api :GET, "/organizations/:organization_id/upstream_subscriptions/simple_content_access/eligible",
```
jturel, line 80: I debated having a separate endpoing (as it is here - `/eligible`) vs just a GET for `/simple_content_access` which we could later, for whatever reason, add more data in the response as we need. That may never happen, so I opted for this and I'm interested in feedback here

app/services/katello/candlepin/upstream_consumer.rb
```diff
@@ -0,0 +1,28 @@
+module Katello
+  module Candlepin
+    class UpstreamConsumer
+      def initialize(organization)
+        @organization = organization
+      end
+
+      def simple_content_access_eligible?
```
jturel, line 8: Based on the output of the new candlepin API (see the PR in description), confirm that this will parse and handle the response appropriately

webpack/scenes/Subscriptions/SubscriptionsPage.js
```diff
@@ -79,6 +80,10 @@ class SubscriptionsPage extends Component {
     }
 
     if (hasUpstreamConnection) {
+      if (hasUpstreamConnection !== prevProps.hasUpstreamConnection) {
```
jturel, line 83: This is so the check is performed when hasUpstreamConnection has changed and is true, and not on every re-render

webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
```diff
@@ -344,6 +347,7 @@ ManageManifestModal.defaultProps = {
   isManifestImported: false,
   canEditOrganizations: false,
   simpleContentAccess: false,
+  simpleContentAccessEligible: false,
```
jeremylenz, line 350: I clicked on my manifest modal and saw the switch in the state of "on, but disabled" for a second.  If SCA is already on, it wouldn't make much sense for it not to be available.

During that moment when the API request is in flight and we don't know the answer, should the switch be enabled by default rather than disabled?
Should this change according to whether SCA is already enabled?

jeremylenz (commented):
Looking nice @jturel 

On the manifest modal, might be useful to have a "disabled reason" tooltip like we do for the delete button.  I can imagine getting support inquiries if the switch is disabled.

More initial comments below

app/controllers/katello/api/v2/upstream_subscriptions_controller.rb
jturel: I like the way you're thinking, but I disagree in this case. As devs you and I know that this set of APIs does talk to hosted candlepin. However, they would not change their interface at all if hosted Candlepin were rewritten in JS (please no) or if the Candlepin API scheme totally changed. Terms like 'Upstream Subscriptions' and 'Simple Content Access' are broader. When we made this controller we needed to differentiate between subscriptions activities happening in the Katello org and the Upstream org, so that's where the name came from. If I added this API under `/katello/api/v2/organizations/simple_content_access/eligible` the question that the API is asking has changed away from "is this **upstream** organization eligible for SCA"

Also keep in mind that my tweak on this line is not changing the API endpoint - it's correcting the apipie doc to match the real API.

app/controllers/katello/api/v2/upstream_subscriptions_controller.rb
```diff
@@ -77,15 +77,22 @@ def ping
...
+      N_("Check if the specified organization is eligible for Simple Content Access")
+    def simple_content_access_eligible
+      eligible = @organization.upstream_consumer.simple_content_access_eligible?
+      render json: { simple_content_access_eligible: eligible }
```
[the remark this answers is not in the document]
jturel: Well they may need an anger management course for yelling at you over something like that ;)

Generally, I agree. But what's the solution? Creating a RABL template with a single line for an API that is unlikely to change feels like blind adherence to some best practice that, imo, decreases readability in this case. The controller is encapsulated entirely in 2 lines of code. I'd like to leave it. If the response of this API should grow it's dead simple to refactor into a RABL template!

webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
jturel: Anything can happen. It's entirely possible that the imported manifest could have SCA toggled on but at some later point the org in the portal disabled SCA toggling across the board. Let me think about the 2nd part of your question

jturel: I'll need to test this:

- i have an imported manifest with SCA enabled
- my upstream org has SCA disabled globally
- am I still able to toggle to 'off' in Katello? (need to test and handle this case, if that works which I think it should)

app/controllers/katello/api/v2/upstream_subscriptions_controller.rb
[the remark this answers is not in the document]
jeremylenz: that's fair! :smile: 

(and I may have been exaggerating a bit about the yelling)

jturel:
> Looking nice @jturel
> 
> On the manifest modal, might be useful to have a "disabled reason" tooltip like we do for the delete button. I can imagine getting support inquiries if the switch is disabled.
> 
> More initial comments below

re: tooltip - good idea

jturel:
@jeremylenz thoughts on this messaging?

image

jeremylenz:
@jturel hmm, I think I'd want something that implies that SCA is out of Katello's control.  As a user I don't think I'd know what "eligible for Simple Content Access" means, let alone "upstream subscription allocation".  And my first question would be "why?" which is the question this is supposed to prevent.  This is tricky...

How about 

"Toggling Simple Content Access is disabled upstream"
"Toggling Simple Content Access is disabled in the customer portal"
"Toggling Simple Content Access is disabled by the manifest's organization admin"
"Toggling Simple Content Access is unavailable due to your organization admin settings"

jturel:
> @jturel hmm, I think I'd want something that implies that SCA is out of Katello's control. As a user I don't think I'd know what "eligible for Simple Content Access" means, let alone "upstream subscription allocation". And my first question would be "why?" which is the question this is supposed to prevent. This is tricky...
> 
> How about
> 
> "Toggling Simple Content Access is disabled upstream"
> "Toggling Simple Content Access is disabled in the customer portal"
> "Toggling Simple Content Access is disabled by the manifest's organization admin"
> "Toggling Simple Content Access is unavailable due to your organization admin settings"

"The upstream subscription allocation is not eligible for Simple Content Access because it has been disabled by the organization administrator" ?

webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
jturel: Updated to allow disabling SCA even if the upstream org is ineligible. That way users can switch back to entitlement mode if they wish, but they won't be able to toggle it back on.

jeremylenz:
> "The upstream subscription allocation is not eligible for Simple Content Access because it has been disabled by the organization administrator" ?

_The upstream subscription allocation is not eligible for Simple Content Access_

This part is 100% technically accurate but feels too wordy to me.  What if we just skip to the end?

"Simple Content Access has been disabled by the organization administrator."

jturel:
@jeremylenz updated!

jeremylenz:
looks like the commits all have an extra `#` :man_shrugging:

jturel:
> looks like the commits all have an extra `#`

that's OK it shall be squashed

jturel:
Oh prprocessor is complaining ... so dramatic

jturel:
squashed+rebased

webpack/scenes/Subscriptions/Manifest/SimpleContentAccess.js
```diff
@@ -21,6 +22,14 @@ const SimpleContentAccess = (props) => {
     }
   };
 
+  const simpleContentAccessText = () => {
+    if (!simpleContentAccessEligible && !isSimpleContentAccessEnabled) {
+      return __('Simple Content Access has been disabled by the upstream organization administrator.');
```
jeremylenz, line 27: While the API request is in flight, this message is shown.  I think it would make more sense to default to the other message instead.  (I know it's only half a second but I feel like details like this make for a better user experience.)

This will require uncoupling the enablement of the switch and the message that's shown.

jeremylenz (approved):
Tested with hard-coded CP response and works as expected

nice work @jturel 

LGTM

PR #9018, opened 2020-10-30
Fixes #31209 - Get repo type filter for CV page from API

johnpmitsch:
The values for this filter can be dynamic based on what is enabled on the server, this gets everything from the server. Error and loading states are also handled by the dropdown.

### DIFF
--- a/webpack/components/SelectableDropdown/SelectableDropdown.js
+++ b/webpack/components/SelectableDropdown/SelectableDropdown.js
@@ -9,8 +9,8 @@ const SelectableDropdown = ({
 }) => {
   const [isOpen, setIsOpen] = useState(false);
   const icon = () => {
-    if (error) return <ErrorCircleOIcon color="red" />;
-    if (loading) return <Spinner size="sm" />;
+    if (error) return <span aria-label={`${title} error`}><ErrorCircleOIcon color="red" /></span>;
+    if (loading) return <span aria-label={`${title} spinner`}><Spinner size="sm" /></span>;
     return null;
   };
   const onSelect = (_event, selection) => {
--- a/webpack/components/SelectableDropdown/__tests__/SelectableDropdown.test.js
+++ b/webpack/components/SelectableDropdown/__tests__/SelectableDropdown.test.js
@@ -1,0 +1,45 @@
+import React from 'react';
+import { render } from 'react-testing-lib-wrapper';
+
+import SelectableDropdown from '../SelectableDropdown';
+
+const type = 'Breakfast';
+
+test('Can dynamically load options', async () => {
+  const { rerender, queryByText, queryByLabelText } = render(<SelectableDropdown
+    items={[]}
+    title={type}
+    selected=""
+    setSelected={jest.fn}
+    placeholderText={type}
+    loading
+    error={false}
+  />);
+  expect(queryByLabelText(`${type} spinner`)).toBeInTheDocument();
+
+  rerender(<SelectableDropdown
+    items={['donut', 'croissant', 'bear claw']}
+    title={type}
+    selected=""
+    setSelected={jest.fn}
+    placeholderText={type}
+    loading={false}
+    error={false}
+  />);
+  expect(queryByLabelText(`${type} spinner`)).not.toBeInTheDocument();
+  expect(queryByText('donut')).toBeInTheDocument();
+});
+
+
+test('Can handle error', async () => {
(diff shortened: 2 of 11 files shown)


johnpmitsch:
Here is what the dropdown looks like:
image

Most likely the user won't see any loading, but here is an example if the server takes a while to load and the API throws an error:

repo_types

Notice only the filter behavior is blocked by the error and the toast notification shows the error. To recreate this, you can add `sleep 10; raise 'error!'` under [here](https://github.com/Katello/katello/blob/152529f808562bf3a1653b8d16c716e1ef7e018c/app/controllers/katello/api/v2/repositories_controller.rb#L253)

johnpmitsch:
It's functional but realized need to update some tests with the new API calls and while I'm at it, this dynamic functionality calls for a separate test for `SelectableDropdown`

webpack/scenes/ContentViews/Details/ContentViewDetailActions.js
```diff
@@ -63,4 +64,11 @@ export const getContentViewRepositories = (cvId, params, status) => {
   });
 };
 
+export const getRepositoryTypes = () => async dispatch => dispatch(get({
```
jeremylenz, line 67: In general `dispatch` shouldn't be used in action creators.  I've never liked that feature of `redux-thunk` because it makes it too easy to muddle the purposes of the different parts of Redux.  I don't think this is needed anyway -- you can just return the action itself and dispatch it in the component.  This way your action creator doesn't need to be a thunk, and can become a simple action instead:

```suggestion
export const getRepositoryTypes => get({
```
And the API middleware will take care of the rest :)

johnpmitsch:
> It's functional but realized need to update some tests with the new API calls and while I'm at it, this dynamic functionality calls for a separate test for `SelectableDropdown`

Tests added, this is all ready for review!

webpack/components/SelectableDropdown/__tests__/SelectableDropdown.test.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import { render } from 'react-testing-lib-wrapper';
+
+import SelectableDropdown from '../SelectableDropdown';
+
+const type = 'Breakfast';
```
jeremylenz, line 6: :fried_egg: :yum:

webpack/components/SelectableDropdown/SelectableDropdown.js
```diff
@@ -1,12 +1,19 @@
...
+
 
 const SelectableDropdown = ({
-  items, title, selected, setSelected,
+  items, title, selected, setSelected, loading, error,
 }) => {
   const [isOpen, setIsOpen] = useState(false);
-  const onSelect = (event, selection) => {
+  const icon = () => {
+    if (error) return <span aria-label={`${title} error`}><ErrorCircleOIcon color="red" /></span>;
```
jeremylenz, line 12: Not that it's going to happen often, but it may be helpful to have a `title` HTML attribute here that says something like "error retrieving repo types from the server."  Then users wouldn't be left wondering what the red X means.

jeremylenz (changes requested):
Works as expected.  Loads so fast I can't see the spinner without using React Devtools.  nice work @johnpmitsch 

Just one ask below about a Redux action.

webpack/scenes/ContentViews/Details/ContentViewDetailActions.js
johnpmitsch: I forgot we have this now https://github.com/theforeman/foreman/pull/7841 - I can update!

webpack/components/SelectableDropdown/SelectableDropdown.js
johnpmitsch: I wasn't able to get the title attribute to show any popover either on the span or the icon itself. In this usage, there is the toast notification, won't that be enough info?

johnpmitsch:
@jeremylenz updated!

webpack/components/SelectableDropdown/SelectableDropdown.js
jeremylenz: I can roll with that.  (I was simulating the error state by changing props with React devtools, so didn't see the toast on screen.)

jeremylenz (approved):
Changes LGTM, thanks @johnpmitsch!

PR #9019, opened 2020-10-30
Fixes #31223 - Pulp 3 task cancelling does not work

ianballou:
Pulp 3 tasks can't be cancelled.  To test, try cancelling a Pulp 3 action, like a content migration.

PR #9020, opened 2020-10-30
Fixes #31195 - Add rake task that kills running pulp 3 migration tasks

ianballou:
This rake task will kill the migration dynflow action and the related Pulp 3 migration task.  Currently it will not cancel ImportMigration since that action isn't skippable.   Cancelling the migration Pulp 3 task is the most important part, so ImportMigration cancelling will come in another PR later.

Dependent on https://github.com/Katello/katello/pull/9019

To test:

1) Apply #9019 and this PR
2) Sync some content with a Pulp 2 enabled Katello server
3) Migrate that content with `rake katello:pulp3_migration`
4) Once the task is running, try cancelling it with `rake katello:pulp3_migration_abort`
5) Check that the task is actually cancelled.  It will not cancel if the last step (ImportMigration) is running.

### DIFF
--- a/lib/katello/tasks/pulp3_migration_abort.rake
+++ b/lib/katello/tasks/pulp3_migration_abort.rake
@@ -1,0 +1,17 @@
+namespace :katello do
+  desc "Cancels all running Pulp 2 to 3 migration tasks."
+  task :pulp3_migration_abort => ["environment"] do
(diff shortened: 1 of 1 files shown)


ianballou:
TODO: check if the Pulp 3 task groups are cancelled as well.

sjha4 (approved):
pulp3_migration_abort cancels the migration task as expected with the task cancelled in pulp.
Ack :+1:

PR #9021, opened 2020-11-02
i18n - pulling from tx

jturel:


app/lib/actions/katello/host/update_system_purpose.rb
```diff
@@ -19,7 +19,11 @@ def plan(host, service_level, purpose_role, purpose_usage, purpose_addons)
         end
 
         def humanized_name
-          _("Updating System Purpose for host %s") % input[:hostname]
+          if input&.dig(:hostname)
```
jturel, line 22: Without this change I couldn't generate the new PO files due to a nil error. Changed the code to mirror other actions

chris1984 (approved):
APJ

ianballou (approved):
Looks fine to me!

PR #9027, opened 2020-11-06
Fixes #31268 - copy all units on pulp3 cv publish

jlsherrill:


app/services/katello/pulp3/repository/yum.rb
```diff
@@ -160,7 +160,7 @@ def copy_content_chunked(data)
             data_dup.config[i][:content] = leftover_units.pop(copy_amount)
             unit_copy_counter += copy_amount
             # Do copy call if limit is reached or if we're under the limit but on the last repo config.
-            if unit_copy_counter >= UNIT_LIMIT || (i == data_dup.config.size - 1 && leftover_units.empty?)
+            if unit_copy_counter != 0
```
jlsherrill, line 163: @ianballou  wanted to get your opinion about this.  It seems to fix the problem, but i feel like i'm not understanding the full purpose of this if statement.   Basically given a set of repos:

RepoA:  1001 RPMs
Repo B:  5 Rpms

It hits this loop once and copies 1000 rpms into Repo A.
It then seems to go through the loop a 2nd time where unit_copy_counter is 1 and thus doesn't copy anything.

ianballou: I see the issue. I put 
```ruby
if unit_copy_counter >= UNIT_LIMIT || (i == data_dup.config.size - 1 && leftover_units.empty?)
```
because I was aiming for the second copy to have 1 RPM from RepoA and the 5 RPMs from RepoB.  This doesn't work though because I clear out data_dup.config[i] [here](893a1c298832d5213af1a68f9aaae82f2507bd2414371bce47fa7c962d477089R170).  I remember I hit some issue with prior data_dup config entries not being empty so I had to clear them.

Changing the `if` statement to
```ruby
if unit_copy_counter != 0
```
is better from a complexity standpoint.  It'll result in a bit more copy calls (1000 RepoA, 1 RepoA, 5 RepoB), but I think that's fine.

app/services/katello/pulp3/repository/yum.rb
```diff
@@ -160,7 +160,7 @@ def copy_content_chunked(data)
             data_dup.config[i][:content] = leftover_units.pop(copy_amount)
             unit_copy_counter += copy_amount
             # Do copy call if limit is reached or if we're under the limit but on the last repo config.
```
ianballou, line 162: This comment won't be relevant any more with your change.

app/services/katello/pulp3/repository/yum.rb
jlsherrill: Thats kinda what i suspected might have been the intent, and yeah i think at this scale (10,000 rpms per copy call), a few extra ones aren't gonna be a big deal.  I'll get that comment cleaned up

ianballou:
`test_copy_content_chunked_limits_units_copied` will need a small update.

jlsherrill:
[test katello]

ianballou:
I think we might need a rebase.

jlsherrill:
[test katello]

ianballou (approved):
Looks good to me!

PR #9036, opened 2020-11-13
Fixes #30828 - conditionally skip content indexing

jjeffers:
This PR introduces the use of the "content_changed" input to index content actions in additional opportunities during a CV publish. The goal for this change is to eliminate unnecessary index operations when it's not needed (no changes detected).

### DIFF
--- a/app/lib/actions/katello/check_matching_content.rb
+++ b/app/lib/actions/katello/check_matching_content.rb
@@ -1,0 +1,17 @@
+module Actions
+  module Katello
+    module CheckMatchingContent
+      def check_matching_content(new_repository, source_repositories)
+        check_matching_content = ::Katello::RepositoryTypeManager.find(new_repository.content_type).metadata_publish_matching_check
+        if new_repository.environment && source_repositories.count == 1 && check_matching_content
+          match_check_output = plan_action(Katello::Repository::CheckMatchingContent,
+                                           :source_repo_id => source_repositories.first.id,
+                                           :target_repo_id => new_repository.id).output
+
+          return match_check_output[:matching_content]
+        end
+        false
+      end
+    end
+  end
+end
--- a/app/lib/actions/katello/repository/check_matching_content.rb
+++ b/app/lib/actions/katello/repository/check_matching_content.rb
@@ -23,8 +23,10 @@ module Actions
             yum_metadata_files = yum_metadata_files_match?(source_repo, target_repo)
             checksum_match = (target_repo.saved_checksum_type == source_repo.saved_checksum_type)
 
+            published = target_repo.backend_service(SmartProxy.pulp_primary).published?
+
             output[:checksum_match] = checksum_match
-            output[:matching_content] = yum_metadata_files && srpms_match && rpms && errata && package_groups && distributions && target_repo.published? && checksum_match
+            output[:matching_content] = yum_metadata_files && srpms_match && rpms && errata && package_groups && distributions && published && checksum_match
           end
 
           if source_repo.content_type == ::Katello::Repository::DEB_TYPE
--- a/app/lib/actions/katello/repository/multi_clone_contents.rb
+++ b/app/lib/actions/katello/repository/multi_clone_contents.rb
@@ -3,6 +3,8 @@ module Actions
     module Repository
       class MultiCloneContents < Actions::Base
         include Actions::Katello::PulpSelector
+        include Actions::Katello::CheckMatchingContent
+
         def plan(extended_repo_mapping, options)
           generate_metadata = options.fetch(:generate_metadata, true)
           copy_contents = options.fetch(:copy_contents, true)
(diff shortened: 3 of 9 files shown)


jjeffers:
[test katello]

jjeffers:
[test katello]

jjeffers:
[test katello]

jjeffers:
[test katello]

jjeffers:
[test katello]

jjeffers:
[test katello]

app/lib/actions/katello/repository/multi_clone_contents.rb
```diff
@@ -18,34 +20,29 @@ def plan(extended_repo_mapping, options)
 
             concurrence do
               extended_repo_mapping.each do |source_repos, dest_repo_map|
+                dest_repo_map[:matching_content] = check_matching_content(dest_repo_map[:dest_repo], source_repos)
```
ianballou, line 23: `dest_repo_map[:matching_content]` is always false for me in this scenario:
1) Sync the `needed_errata` repo
2) Add it to the content view
3) Publish n times

I'm assuming that indexing shouldn't be happening after the first CV publish.  I noticed that `dest_repo_map[:dest_repo]`, which is fed to `check_matching_content`, is archived and has no RPMs.  Seems like is is why my indexing always occurs.

jjeffers: It appears that on a subsquent publish, the index content action is invoked because the sub-action, CheckMatchingContent returns the output of "matching_content" as false, which is confusing since it's the same repository.

ianballou: I believe I was referring to the index action being planned.  I remember seeing the index action happening more times than I expected. If you're not seeing this then I might have to go back and try again, it's been a while.

jjeffers: Justin pointed out that these changes included a boolean that is actually a deferred value not set until the run phase of the action, so it is always truthy. This is probably part of the issue! Hold off on any retesting until I can refactor.

jjeffers:
@ianballou ok fixed the issue that Justin spotted, should be worth a retest at this point.

ianballou:
@jjeffers I will re-test soon.

ianballou:
[test katello]

ianballou (commented):
Functionally it seems like everything is working! I just need to do some more corner case tests.

app/lib/actions/katello/repository/check_matching_content.rb
```diff
@@ -23,8 +23,10 @@ def run
             yum_metadata_files = yum_metadata_files_match?(source_repo, target_repo)
             checksum_match = (target_repo.saved_checksum_type == source_repo.saved_checksum_type)
 
+            published = target_repo.backend_service(::SmartProxy.pulp_primary).published?
```
ianballou, line 26: This line is causing ```NoMethodError: undefined method `pulp3_support?' for nil:NilClass``` in some of the tests

jjeffers: @ianballou ah yes, I suspected that approach was not ideal. I will see about passing a relevant smart proxy parameter into this action. Thanks!

jjeffers:
[test katello]

jjeffers:
@ianballou green!

ianballou:
One thing is confusing me. In the photo below, there are the Index actions in the concurrent boxs (blue) and the Index actions in the sequential boxes (orange).  It doesn't seem like the blue ones are getting skipped. Should they be?

The skipping is definitely working for the sequential Index actions.

Screenshot from 2021-01-11 14-49-55

jjeffers:
@ianballou check the output from the Index actions. The index content checks for checksum and content matching before running the action, and if the run action is skipped (vs just not planned at all), there should be an output value indicating the action was skipped.

Not sure about the GenerateMetadata action - I didn't modify anything related there in this PR.

jjeffers:
@ianballou what was the scenario that you tested above?

ianballou:
@jjeffers I have a content view with a couple of repositories in it without any filters. I just published a new content view version.

jjeffers:
[test katello]

ianballou (commented):
I re-published my repositories and now the resulting content view version is empty. I noticed `post_action_skipped: true` on the concurrent `IndexContent`s but not the sequential `IndexContent`s. Are they out of sync?

ianballou:
Looks like it's back to how it was before, where the concurrent `IndexContent` actions are always running but the sequential ones are being skipped.

app/lib/actions/katello/check_matching_content.rb
```diff
@@ -0,0 +1,17 @@
+module Actions
+  module Katello
+    module CheckMatchingContent
+      def check_matching_content(new_repository, source_repositories)
+        check_matching_content = ::Katello::RepositoryTypeManager.find(new_repository.content_type).metadata_publish_matching_check
+        if new_repository.environment && source_repositories.count == 1 && check_matching_content
```
ianballou, line 6: I'm finding that `matching_content` is false for my concurrent `IndexContent` still because `new_repository.environment` is `nil`.  It seems like this is always the case for me.

jjeffers: Great catch!

jjeffers: As per discussion with @jlsherrill if the target repo is an archive repo, this is expected.

ianballou (approved):
Looking good to me!

app/lib/actions/katello/repository/check_matching_content.rb
```diff
@@ -23,8 +23,10 @@ def run
             yum_metadata_files = yum_metadata_files_match?(source_repo, target_repo)
             checksum_match = (target_repo.saved_checksum_type == source_repo.saved_checksum_type)
 
+            published = target_repo.backend_service(SmartProxy.pulp_primary).published?
```
hao-yu, line 26: It appears that this will break the pulp2 support because this new method is not added to the Katello::Pulp::Repository. Is it intended?

jlsherrill: i had noticed this too a while back on my development environment, but this commit went into katello 4.0 which no longer includes pulp2.  Due to that, i didn't file a bug as it shouldn't ever show up for a user.  Under what version/environment are you seeing this be an issue?

hao-yu: I don't see any issue. I just notice this change when working on a patch (https://github.com/Katello/katello/pull/9348) in the master branch for Satellite 6.9 so I thought I will just ask to confirm you guys know about it.

jlsherrill: yes, this could NOT be cherry picked to 6.9 (or katello 3.18) as is

PR #9039, opened 2020-11-16
Fixes #31312 - only import modified errata on re-migration

jlsherrill:
this uses the previous successful migration as a timestamp
from which to look for modified errat that need to be re-imported


jlsherrill:
[test katello]

ianballou (approved):
Works for me!

PR #9041, opened 2020-11-17
Fixes #31321 - deleted host causes applicability error

ianballou:
If a host is missing, or if it doesn't have a content facet, the rescue block now will skip it instead of erroring out the entire task.

To test:

1) Make a content host
2) `ForemanTasks.sync_task(::Actions::Katello::Applicability::Hosts::BulkGenerate, host_ids: [<real id>,<random fake id>])`
3) Check the logs for `Host with ID <random fake id> does not exist. Skipping applicability calculation.`
4) Delete the content host's content facet
5) `ForemanTasks.sync_task(::Actions::Katello::Applicability::Hosts::BulkGenerate, host_ids: [<real id>])`
6) Check the logs for `Host with ID <read id> is not a content host. Skipping applicability calculation.`

### DIFF
--- a/app/lib/actions/katello/applicability/hosts/bulk_generate.rb
+++ b/app/lib/actions/katello/applicability/hosts/bulk_generate.rb
@@ -9,8 +9,12 @@ module Actions
 
           def run
             input[:host_ids].each do |host_id|
-              content_facet = ::Host.find(host_id).content_facet
-              content_facet.calculate_and_import_applicability
+              content_facet = ::Katello::Host::ContentFacet.find_by_host_id(host_id)
+              if content_facet.present?
+                content_facet.calculate_and_import_applicability
+              else
+                Rails.logger.warn(_("Content Facet for host with id %s is non-existent. Skipping applicability calculation.") % host_id)
+              end
             end
           end
 
--- a/test/actions/katello/applicability/generate_applicability_test.rb
+++ b/test/actions/katello/applicability/generate_applicability_test.rb
@@ -29,5 +29,23 @@ module ::Actions::Katello::Applicability::Host
         ForemanTasks.sync_task(action_class, [@host, @host, @host, @host, @host])
       end
     end
+
+    describe 'BulkGenerate does not error out with missing hosts' do
+      let(:action_class) { ::Actions::Katello::Applicability::Hosts::BulkGenerate }
+
+      it 'runs' do
+        deleted_id = @host.id
+        @host.subscription_facet.destroy
+        @host.content_facet.destroy
+        @host.destroy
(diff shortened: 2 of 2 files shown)


ianballou:
[test katello]

app/lib/actions/katello/applicability/hosts/bulk_generate.rb
```diff
@@ -10,7 +10,14 @@ class BulkGenerate < Actions::EntryAction
           def run
             input[:host_ids].each do |host_id|
               content_facet = ::Host.find(host_id).content_facet
-              content_facet.calculate_and_import_applicability
+            rescue ActiveRecord::RecordNotFound
```
jturel, line 13: In general I like to avoid having exception handling to control flow, saving it for those instances where we really can't avoid it. In this case you could you find_by_id vs find to avoid raising an error. Though, looking at the code I don't think we need the host object at all, so maybe better to do `content_facet = Katello::Host::ContentFacet.find_by_host_id(host_id)`

Looks like your test will still pass with this tweak. Thanks for adding test coverage!

ianballou: I didn't know `find_by_host_id` existed, seems perfect.

ianballou:
@jturel I've addressed your comment.

app/lib/actions/katello/applicability/hosts/bulk_generate.rb
jturel: Rails magic!

app/lib/actions/katello/applicability/hosts/bulk_generate.rb
```diff
@@ -9,8 +9,12 @@ class BulkGenerate < Actions::EntryAction
 
           def run
             input[:host_ids].each do |host_id|
-              content_facet = ::Host.find(host_id).content_facet
-              content_facet.calculate_and_import_applicability
+              content_facet = ::Katello::Host::ContentFacet.find_by_host_id(host_id)
+              if content_facet.present?
+                content_facet.calculate_and_import_applicability
+              else
+                Rails.logger.warn(_("Content Facet for host with %s is non-existent. Skipping applicability calculation.") % host_id)
```
jturel, line 16: I think you meant "with id %s"

jturel:
Works perfectly! Left one comment on the warning message

app/lib/actions/katello/applicability/hosts/bulk_generate.rb
ianballou: Oops, I've fixed this.

jturel (approved):
APJ!

ianballou:
[test katello]

jturel:
[test katello]

PR #9043, opened 2020-11-18
Fixes #31337 - Katello sends unauth. docker repos at proxy sync time

ianballou:
To be tested with https://github.com/Katello/smart_proxy_container_gateway/pull/2 if it's not already merged.  Use the same setup instructions.  Test instructions become simpler.  Just sync your smart proxy and curl the smart proxy api for the unauthenticated repo list: 
```
curl https://`hostname`/v2/unauthenticated_repository_list
```

### DIFF
--- a/db/seeds.d/104-proxy.rb
+++ b/db/seeds.d/104-proxy.rb
@@ -16,7 +16,7 @@ User.as(::User.anonymous_api_admin.login) do
     fail "Unable to create proxy feature: #{format_errors feature}"
   end
 
-  ["Pulp", "Pulp Node", "Pulpcore"].each do |input|
+  ["Pulp", "Pulp Node", "Pulpcore", "Container_Gateway"].each do |input|
     f = Feature.where(:name => input).first_or_create
     fail "Unable to create proxy feature: #{format_errors f}" if f.nil? || f.errors.any?
   end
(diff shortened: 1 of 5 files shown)


lib/katello/engine.rb
```diff
@@ -202,6 +202,7 @@ class Engine < ::Rails::Engine
       require "#{Katello::Engine.root}/app/services/katello/puppet_class_importer_extensions"
       require "#{Katello::Engine.root}/lib/proxy_api/pulp"
       require "#{Katello::Engine.root}/lib/proxy_api/pulp_node"
+      require "#{Katello::Engine.root}/lib/proxy_api/container_gateway"
```
jlsherrill, line 205: did you forget to add this file? :)

ianballou: I sure did! I'll add it in a second.

db/seeds.d/104-proxy.rb
```diff
@@ -16,7 +16,7 @@ def format_errors(model = nil)
     fail "Unable to create proxy feature: #{format_errors feature}"
   end
 
-  ["Pulp", "Pulp Node", "Pulpcore"].each do |input|
+  ["Pulp", "Pulp Node", "Pulpcore", "Container_Gateway"].each do |input|
```
ianballou, line 19: I was hoping the feature would show up as "Container Gateway" without the underscore, but it seems the Foreman feature mapper adds the underscore.  It's not a huge deal, but if anyone knows a workaround without renaming the plugin let me know.

jjeffers (approved):
Tested this with previous smart proxy setup and eveything worked as expected. The only hiccup was a katello seed that didn't populate the smart proxy feature "ContainerGateway". After adding that via the console and refreshing the capsule, I was able to see the list of repos as per the testing instructions on the proxy via http/curl.

PR #9044, opened 2020-11-19
Fixes #31344 - Update @theforeman/builder

johnpmitsch:


package.json
```diff
@@ -26,7 +26,7 @@
     "@sheerun/mutationobserver-shim": "^0.3.3",
     "@testing-library/jest-dom": "^5.3.0",
     "@testing-library/react": "^10.0.2",
-    "@theforeman/builder": "^4.6.0",
+    "@theforeman/builder": "^6.0.0",
```
tbrisker, line 29: Should this be `>= 6.0.0` to ensure it doesn't break on the next update?

johnpmitsch: we could change to that for builder for now. Ideally we don't have to update all the foreman-js metapackages at the same time so a major version change in `@theforeman/builder` did mean a breaking change. I'm opening up a discussion on this topic.

johnpmitsch: @tbrisker updated to use `>=`

evgeni:
[test katello]

evgeni:
The three failing tests are unrelated (Pulp3 VCR tapes don't match) and also fail on other PRs.

tbrisker (approved):
Thanks @johnpmitsch !

PR #9045, opened 2020-11-19
Refs #30832 - check for pulp2 presence rather than pulp3 yum support

jturel:
Can't install errata via pulp2 if there's no consumer profile, so check for presence of pulp2 instead of whether yum content is handled by pulp3

### DIFF
--- a/app/lib/actions/katello/host/upload_package_profile.rb
+++ b/app/lib/actions/katello/host/upload_package_profile.rb
@@ -39,7 +39,7 @@ module Actions
             Rails.logger.warn("Host with ID %s has no content facet, continuing" % host_id)
           else
             begin
-              unless SmartProxy.pulp_primary&.pulp3_repository_type_support?(::Katello::Repository::YUM_TYPE)
+              if SmartProxy.pulp_primary&.has_feature?(SmartProxy::PULP_FEATURE)
                 ::Katello::Pulp::Consumer.new(host.content_facet.uuid).upload_package_profile(profile)
               end
               simple_packages = profile.map { |item| ::Katello::Pulp::SimplePackage.new(item) }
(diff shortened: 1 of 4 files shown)


app/lib/actions/katello/host/upload_package_profile.rb
```diff
@@ -39,7 +39,7 @@ def self.upload(host_id, profile)
             Rails.logger.warn("Host with ID %s has no content facet, continuing" % host_id)
           else
             begin
-              unless SmartProxy.pulp_primary&.pulp3_repository_type_support?(::Katello::Repository::YUM_TYPE)
+              if SmartProxy.pulp_primary&.has_feature?(SmartProxy::PULP_FEATURE)
```
ekohl, line 42: Shouldn't it check if yum is actually handled by Pulp 2?

jlsherrill: Pulp 2 doesn't support reporting specific content types/plugins, i supposed we could check that yum is enabled for katello and pulp2 is enabled, but i don't know that running with yum disabled is actually supported.

ekohl: Doesn't it still make sense to check `pulp2_preferred_for_type?` for yum? Maybe these questions are more for my own understanding of flows.

jlsherrill:
[test katello]

app/lib/actions/katello/host/upload_package_profile.rb
jlsherrill: the problem is that even if pulp3 is the preferred type, if you are using pulp2 for katello-agent actions (errata install), pulp2 requires a profile to be uploaded.  This is likely a bug in pulp2 from long ago when pulp actually needed this information to install an errata, but now it really doesn't (but thats moot at this point).  So this needs to continue to upload a package profile to pulp if katello-agent is planning to be used with pulp2.

ekohl: Thanks, that makes sense.

jlsherrill (approved):

PR #9047, opened 2020-11-23
Fixes #31334 - fix export transient failure

jlsherrill:


jlsherrill:
no idea if this will actually fix it.... lets see

jlsherrill:
i don't think this is on the right track

jlsherrill:
[test katello]

jeremylenz (approved):
Thanks @jlsherrill, tests pass for me and the mocking seems much safer than redefining the method.  LGTM :+1:

PR #9048, opened 2020-11-25
WIP: Fixes #31395 - Debian filter in Content View without version comparison

Manisha15:
Add a Debian Filter in Content View to filter deb packages on basis of name and architecture.

### DIFF
--- a/app/controllers/katello/api/v2/content_view_filter_rules_controller.rb
+++ b/app/controllers/katello/api/v2/content_view_filter_rules_controller.rb
@@ -22,7 +22,7 @@ module Katello
     api :POST, "/content_view_filters/:content_view_filter_id/rules",
         N_("Create a filter rule. The parameters included should be based upon the filter type.")
     param :content_view_filter_id, :number, :desc => N_("filter identifier"), :required => true
-    param :name, Array, of: [String], :desc => N_("package, package group, or docker tag names")
+    param :name, Array, of: [String], :desc => N_("deb, package, package group, or docker tag names")
     param :uuid, String, :desc => N_("package group: uuid")
     param :version, String, :desc => N_("package: version")
     param :architecture, String, :desc => N_("package: architecture")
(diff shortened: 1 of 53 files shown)


sbernhard:
[test katello]

m-bucher:
[test katello]

m-bucher:
[test katello]

jlsherrill:
Taking a look at this, any planned support for pulp3?

quba42:
> Taking a look at this, any planned support for pulp3?

The key word here is "planned". To be more precise we plan to do a more full featured version of filters for the pulp 3 implementation (that also includes version comparisons). However, to do that I believe we need to add "advanced copy" features similar to those in `pulp_rpm` to `pulp_deb` as well. And to add those we need to add a psql postgres extension for Debian version comparisons (which we also need for the Debian errata story for Katello). This is high on our internal priority list, but the actual implementation has not yet started...

jlsherrill:
Hitting a couple issues:

1) filter didnt' seem to work:
I synced http://deb.theforeman.org/ buster nightly and got 159 debs
I then created an exclude filter to exlude 'foreman'.  Publishing i still get 159 debs.  Setting arch to 'amd64' didn't seem to change it at all.
I also checked the filesystem to see if it was an indexing issue, but the generated metadata seemed to have that same number

2) showing matching content seems to be missing info:
image


Also, katello 3.18 has released, which is the last release with pulp2.  ~90% of this code would be useful for pulp3 filteringi'm wondering though if it makes sense to include this in 4.0 when it doesn't work?  I'm not necessarily opposed to merging it, but might it make more sense to merge with pulp3 support to some degree?

Manisha15:
I have fixed the issue in separate commit for better visibility of changes. If the functionality is approved I can squash the commits into one.

The Apt Repositories in CV Version is not showing correct number of packages. It is showing all packages instead of filtered ones.  Other than this GUI issue, the filter works correctly.

jlsherrill:
Hey sorry, i meant to ask again and reiterate about the dropping of pulp2.  I don't think it makes sense to merge this into master without pulp 3 support, as master is now tracking the 2nd release of katello without pulp2.   If we merged this now (or even when this pr was opened) for the following katello release it wouldn't have worked because there was no pulp2 in 4.0 or 4.1.  The simplest path forward would be to get pulp3 support (and i know there were pulp-debian chnges needed).

Manisha15:
Changing the status to WIP as we are working on adding pulp3 support.

PR #9054, opened 2020-11-30
Fixes #30867 - Tag a manifest from hammer (pulp3)

sjha4:
To test:

Create a docker repository fedora/ssh for example and sync
From hammer, try adding a new tag.
hammer -d repository update --id $repo_id --organization-id $org_id --docker-tag some_name --docker-digest sha256:a6ecbb1553353a08936f50c275b010388ed1bd6d9d84743c7e8e7468e2acd82e

This should create a new tag some_name in the repository.

### DIFF
--- a/app/lib/actions/pulp3/orchestration/repository/import_upload.rb
+++ b/app/lib/actions/pulp3/orchestration/repository/import_upload.rb
@@ -1,3 +1,4 @@
+# rubocop:disable Metrics/MethodLength
 module Actions
   module Pulp3
     module Orchestration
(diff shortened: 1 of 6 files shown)


sjha4:
[test katello]

ianballou (commented):
I'm having an issue where the uploaded tags are not always indexed.  I upload one tag, and I don't see the update.  I upload another tag, only then do I see my first tag, but not the second.  **After poking around a bit, it appears that the repository version for the repository is off by one.**

Also, if I upload a tag that has the same name as another tag but a different digest, the old tag will be silently overwritten.  There is still the off-by-one issue in this case too, so to see it I had to upload another tag to see the overwrite occur.


This might not be a new issue, but I noticed as well that the error message for a digest that doesn't exist is simply `Failed to upload tag 'potatotaco2' to repository.`.  This message doesn't provide much help for debugging.

sjha4:
> I'm having an issue where the uploaded tags are not always indexed. I upload one tag, and I don't see the update. I upload another tag, only then do I see my first tag, but not the second. **After poking around a bit, it appears that the repository version for the repository is off by one.**

This should be fixed with last commit

> Also, if I upload a tag that has the same name as another tag but a different digest, the old tag will be silently overwritten. There is still the off-by-one issue in this case too, so to see it I had to upload another tag to see the overwrite occur.

I tested this. This is same as existing pulp2 behavior. 
 
> This might not be a new issue, but I noticed as well that the error message for a digest that doesn't exist is simply `Failed to upload tag 'potatotaco2' to repository.`. This message doesn't provide much help for debugging.

This is also same as pulp2. I think we can print more details with changes in hammer.

ianballou (approved):
Works for me :+1:

PR #9057, opened 2020-12-01
Fixes #31418 - avoid duplicating notifications

ares:
Please see https://github.com/theforeman/foreman/pull/8165 for more background. This is quite severe, all email notifications doubled. There's more issues like this thanks to wide use of `load` in katello tasks, but that's out of the scope for this PR.


jlsherrill (approved):
Looks like there's a rubocop issue in master, will take a look

chris1984:
[test katello]

PR #9060, opened 2020-12-01
Fixes #31343 - New CV Modal

sjha4:
To-do
- [x] Lint
- [x]  Tests


webpack/scenes/ContentViews/Create/ContentViewCreateActions.js
```diff
@@ -0,0 +1,18 @@
...
+import api, { orgId } from '../../services/api';
+import CONTENT_VIEWS_KEY from './ContentViewsConstants';
+
+export const createContentViewsParams = extraParams => ({
+  organization_id: orgId(),
+  nondefault: true,
+  ...extraParams,
+});
+
+const createContentView = extraParams => get({
```
johnpmitsch, line 11: I know this is still a draft, but here are some small suggestions that hopefully will save some time :) 

is it a `get` or a `post`? Looks like there is a `post` available [here](https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/redux/API/APIActions.js#L24). I think you can drop the `type` param too.

There is also an `errorToast` and `successToast` param available [here](https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/redux/API/APIRequest.js#L20-L21 ), I think these will be helpful here as well for success and error handling feedback.

johnpmitsch:
What do you think about putting all the redux actions into `webpack/scenes/ContentViews/ContentViewsActions.js`? It could easily share the `createContentViewParams` and keeps things in one place.

webpack/scenes/ContentViews/Create/CreateContentViewModal.js
```diff
@@ -0,0 +1,34 @@
...
+const CreateContentViewModal = ({ show }) => {
+  const [modalOpen, setModalOpen] = useState(show);
+  useEffect(
+    () => {
+      setModalOpen(show);
+    },
+    [show]
+  );
+   return (
+      <Fragment>
```
johnpmitsch, line 14: This was was wrong in my example :flushed: - you can drop the `Fragment` wrapping tags as its one component

webpack/scenes/ContentViews/Create/CreateContentViewForm.js
```diff
@@ -0,0 +1,94 @@
...
+  const [label, setLabel] = useState('');
+  const [description, setDescription] = useState('');
+  const [composite, setComposite] = useState(false);
+  const toggleComposite = () => {
+    setComposite(!composite);
+  };
+  const [dependencies, setDependencies] = useState(false);
+  const toggleDependencies = () => {
+    setDependencies(!dependencies);
+  };
```
johnpmitsch, line 23: are these wrapper functions going away? I don't see them in use anywhere

webpack/scenes/ContentViews/ContentViewsActions.js
```diff
@@ -15,4 +22,25 @@ const getContentViews = extraParams => get({
   params: createContentViewsParams(extraParams),
 });
 
+const cvCreateSuccess = (response, dispatch) => {
+  const { data: { name } } = response;
+  return dispatch(addToast({
+    type: 'success',
+    message: __(' Content view ' + name + ' created'),
+  }));
+};
```
johnpmitsch, line 31: could the `successToast` param work for the notification? this also helps us not pass in the dispatch below

https://github.com/theforeman/foreman/blob/a58aa42c9b5aa39ea2209dd01a06334ddfbd777d/webpack/assets/javascripts/react_app/redux/API/APIRequest.js#L20

some of the other actions didn't use this as it wasn't a feature in the API middleware at the time (we should update this at some point)

johnpmitsch: same question error handling for using `errorToast`

webpack/scenes/ContentViews/Create/CreateContentViewForm.js
sjha4: Yes..these aren't needed..

webpack/scenes/ContentViews/ContentViewsActions.js
sjha4: Updated..

webpack/scenes/ContentViews/Create/ContentViewCreateSelectors.js
```diff
@@ -0,0 +1,18 @@
...
+export const selectCreateContentViews = state =>
+  selectAPIResponse(state, CREATE_CONTENT_VIEW_KEY) || {};
+
+export const selectCreateContentViewStatus = state =>
+  selectAPIStatus(state, CREATE_CONTENT_VIEW_KEY) || STATUS.PENDING;
+
+export const selectCreateContentViewError = state =>
+  selectAPIError(state, CREATE_CONTENT_VIEW_KEY);
+
+export const selectIsCVSaving = state => state.katello?.contentViewCreate?.saving;
```
johnpmitsch, line 18: You should be able to use the status selector to check this, we've been going with this recently over setting `saving` in a reducer.

webpack/scenes/ContentViews/ContentViewsActions.js
```diff
@@ -15,4 +17,23 @@ const getContentViews = extraParams => get({
...
 });
 
+const cvSuccessToast = (response) => {
+  const { data: { name } } = response;
+  return __(`Content view ${name} created`);
+};
+
+const cvErrorToast = (error) => {
+  const messages = getResponseErrorMsgs(error.response);
+  return __(messages);
```
johnpmitsch, line 27: I don't think translate works dynamically, though I haven't come to a conclusive answer, generally I use it when there are fixed strings.

Error messages might be translated on the server end? :thinking:

webpack/scenes/ContentViews/Create/CreateContentViewForm.js
```diff
@@ -0,0 +1,154 @@
...
+
+  useEffect(
+    () => {
+      setLabel(name.replace(/ /g, '_'));
+    },
+    [name],
+  );
+
+  if (redirect) {
+    return (<Redirect to={redirectUrl} />);
```
johnpmitsch, line 54: Something about setting the `redirectUrl` as a hook still doesn't feel right, what about doing this?
```js
  if (redirect) {
    const { id } = response;
    return (<Redirect to={`/labs/content_views/${id}`} />);
```
and remove the hook from the effect above

webpack/scenes/ContentViews/ContentViewsActions.js
```diff
@@ -15,4 +17,23 @@ const getContentViews = extraParams => get({
...
+};
+
+export const createContentView = params => post({
+  type: API_OPERATIONS.POST,
+  key: CREATE_CONTENT_VIEW_KEY,
+  url: api.getApiUrl('/content_views'),
+  params,
+  successToast: response => cvSuccessToast(response),
+  errorToast: error => cvErrorToast(error),
+});
```
johnpmitsch, line 37: nice usage of API middleware :+1:

webpack/scenes/ContentViews/Create/ContentViewCreateReducer.js
```diff
@@ -0,0 +1,23 @@
+import Immutable from 'seamless-immutable';
+import {
+  CREATE_CONTENT_VIEW,
+  CREATE_CONTENT_VIEW_FAILURE,
+  CREATE_CONTENT_VIEW_SUCCESS,
```
jeremylenz, line 5: Foreman's API middleware automatically sets the action type from the key you generated in `ContentViewsConstants.js`, `CONTENT_VIEW_CREATE`.  (see https://github.com/theforeman/foreman/blob/fb285ce995975fb255151d9385a16554b8d51295/webpack/assets/javascripts/react_app/redux/API/APIActionTypeGenerator.js#L10)  As a result, the actions that happen are not those referenced here, but rather `CONTENT_VIEW_CREATE_REQUEST` and `CONTENT_VIEW_CREATE_SUCCESS`:

cv_create_redux_actions

Handling of the request-pending state also comes for free with the API middleware: https://github.com/theforeman/foreman/blob/fb285ce995975fb255151d9385a16554b8d51295/webpack/assets/javascripts/react_app/redux/API/APIReducer.js#L20

The raw API responses are also stored by the API middleware.  This reducer and constants may not be needed.

webpack/scenes/ContentViews/Create/ContentViewCreateSelectors.js
jeremylenz: +1, I'd prefer checking the pending state with `selectAPIStatus` from `foremanReact/redux/API/APISelectors.js`

webpack/scenes/ContentViews/Create/CreateContentViewModal.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Modal, ModalVariant } from '@patternfly/react-core';
```
jeremylenz, line 3: I may be biased, but consider using the `ForemanModal` component to be consistent with other modals.  Usage details here: http://foreman-storybook.surge.sh/?path=/story/components-foremanmodal-foremanmodal-usage--foreman-modal-basics

It hasn't been upgraded yet to use the PF4 modal component, but hopefully that shouldn't make too much visual difference..  (If so then it's fine to leave as is.)

webpack/scenes/ContentViews/Create/CreateContentViewFormComponents.js
```diff
@@ -0,0 +1,70 @@
...
+);
+
+export const labelAutoPublish = () => (
+  <Flex>
+    <FlexItem spacer={{ default: 'spacerSm' }}>Auto Publish </FlexItem>
+    <FlexItem>
+      <Popover
+        bodyContent={
+            __('Applicable only for composite views. Auto publish composite view when a new version of a component content view is created.' +
+            'Also note auto publish will only happen when the component is marked latest.')
```
jeremylenz, line 55: this sentence was probably there already, but I don't understand what it means

webpack/scenes/ContentViews/Create/CreateContentViewFormComponents.js
```diff
@@ -0,0 +1,70 @@
+import { translate as __ } from 'foremanReact/common/I18n';
+import React from 'react';
+import { Popover, Flex, FlexItem } from '@patternfly/react-core';
+import { HelpIcon } from '@patternfly/react-icons';
+import ContentViewIcon from '../components/ContentViewIcon';
+
+
+export const labelComposite = () => (
```
jeremylenz, line 8: A function that returns JSX is a React component, so the name should be styled as such: `LabelComposite`

webpack/scenes/ContentViews/Create/CreateContentViewFormComponents.js
```diff
@@ -0,0 +1,70 @@
...
+  <Flex>
+    <FlexItem spacer={{ default: 'spacerNone' }}><ContentViewIcon composite /></FlexItem>
+    <FlexItem>Composite Content View</FlexItem>
+  </Flex>
+);
+
+export const labelComponent = () => (
```
jeremylenz, line 15: same comment as above

webpack/scenes/ContentViews/Create/CreateContentViewFormComponents.js
```diff
@@ -0,0 +1,70 @@
...
+  <Flex>
+    <FlexItem spacer={{ default: 'spacerNone' }}><ContentViewIcon composite={false} /></FlexItem>
+    <FlexItem>Component Content View</FlexItem>
+  </Flex>
+);
+
+export const labelDependencies = () => (
```
jeremylenz, line 22: same comment as above

webpack/scenes/ContentViews/Create/CreateContentViewFormComponents.js
```diff
@@ -0,0 +1,70 @@
...
+          className="pf-c-form__group-label-help"
+        >
+          <HelpIcon noVerticalAlign />
+        </button>
+      </Popover>
+    </FlexItem>
+  </Flex>
+);
+
+export const labelAutoPublish = () => (
```
jeremylenz, line 48: same comment as above

webpack/scenes/ContentViews/Create/CreateContentViewForm.js
```diff
@@ -0,0 +1,154 @@
...
+  const [name, setName] = useState('');
+  const [label, setLabel] = useState('');
+  const [description, setDescription] = useState('');
+  const [composite, setComposite] = useState(false);
+  const [component, setComponent] = useState(true);
+  const [autoPublish, setAutoPublish] = useState(false);
+  const [importOnly, setImportOnly] = useState(false);
+  const [dependencies, setDependencies] = useState(false);
+  const [redirect, setRedirect] = useState(false);
+  const [redirectUrl, setRedirectURL] = useState('/labs/content_views');
```
jeremylenz, line 22: Take a look at the `ForemanForm` component: http://foreman-storybook.surge.sh/?path=/story/components-foreman-form--basic-foreman-form

It uses a library called [Formik](https://formik.org/) and it may help reduce some of this boilerplate for keeping track of the form fields' state.

webpack/scenes/ContentViews/Create/CreateContentViewForm.js
```diff
@@ -0,0 +1,154 @@
...
+    dependencies,
+    auto_publish: (autoPublish && composite),
+    import_only: importOnly,
+  }));
+
+  useEffect(
+    () => {
+      setLabel(name.replace(/ /g, '_'));
+    },
+    [name],
```
jeremylenz, line 50: Love this :heart_eyes:

webpack/scenes/ContentViews/Create/CreateContentViewForm.js
```diff
@@ -0,0 +1,154 @@
...
+        <FormGroup isInline fieldId="autoPublish">
+          <Checkbox
+            id="autoPublish"
+            name="autoPublish"
+            label={labelAutoPublish()}
+            isChecked={autoPublish}
+            onChange={checked => setAutoPublish(checked)}
+          />
+        </FormGroup>}
+      <FormGroup isInline fieldId="importOnly">
```
jeremylenz, line 132: Would be nice to have a description for this as well

jeremylenz (commented):
Welcome to React world @sjha4 :)

Comments below

webpack/scenes/ContentViews/Create/CreateContentViewModal.js
sjha4: I think @johnpmitsch wanted to be on PF4 for the modal and form on this page instead of using Foreman Form and modal.

webpack/scenes/ContentViews/Create/ContentViewCreateReducer.js
sjha4: Ya..meant to remove these..I was using this before in an earlier version so these are lingering from there.

webpack/scenes/ContentViews/Create/CreateContentViewForm.js
johnpmitsch: @jeremylenz so `ForemanForm` would wrap the pf4 input components? 

If so, I think it would be ok, the only built-in pf3 component I see is the alert [here](https://github.com/theforeman/foreman/blob/c18d67464d81943b77feeb0e24ec649574860ab7/webpack/assets/javascripts/react_app/components/common/forms/Form.js#L3), ideally we upgrade it but its fine for the time being IMO

jeremylenz: @johnpmitsch Yeah I don't see any issue with having PF4 components as children of the form.  `ForemanModal` is a little bit more tricky because it does have the PF3 modal components baked in..

jeremylenz: Another consideration is I'm not sure if ForemanForm has been updated to the newer version of Formik that works well with hooks.  probably should check that before going too far.

sjha4: I am not entirely sure that ForemanForm is handling pf4 components well as children..It only seems to work if I use TextField from foreman forms (https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/components/common/forms/TextField/TextField.js).
Otherwise it doesn't tie the input to formik state which seems to be expected since we don't have formik wrappers for the pf4 form components AFAIU. Does that make sense?

jeremylenz: Ok, looks like the existing Foreman components aren't keeping up.  Per our discussion just now I guess we should go with the current approach for now and discuss how to upgrade ForemanForm later.

webpack/scenes/ContentViews/Create/CreateContentViewForm.js
sjha4: Added desc..

webpack/scenes/ContentViews/Create/CreateContentViewFormComponents.js
sjha4: So for a composite content view, publishing any child cv which was included with "Always use latest version" option will re-publish the parent as well.

webpack/scenes/ContentViews/Create/ContentViewCreateSelectors.js
sjha4: Updated to use STATUS.PENDING

webpack/scenes/ContentViews/Create/CreateContentViewFormComponents.js
jeremylenz: I guess the "is marked latest" was a bit confusing to me.

maybe something like this?
```
Automatically publish a new version of the composite content view whenever one of its component content views is published.
(Also note auto-publish will only happen for component views that use the 'Always use latest version' option.)
```

sjha4: Ya this seems better explained. Updated.

webpack/scenes/ContentViews/Create/CreateContentViewFormComponents.js
```diff
@@ -0,0 +1,93 @@
...
+  </Flex>
+);
+
+export const LabelImportOnly = () => (
+  <Flex>
+    <FlexItem spacer={{ default: 'spacerSm' }}>Import Only </FlexItem>
+    <FlexItem>
+      <Popover
+        bodyContent={
+          __('Designate whether this Content View is for importing from an upstream server. Import-only Content Views can not be published directly.')
```
jeremylenz, line 78: I think they like us to use lower case for the new stuff; also one typo:

```suggestion
          __('Designate whether this content view is for importing from an upstream server. Import-only content views cannot be published directly.')
```

webpack/scenes/ContentViews/Create/CreateContentViewFormComponents.js
```diff
@@ -0,0 +1,93 @@
...
+  </Flex>
+);
+
+export const LabelDependencies = () => (
+  <Flex>
+    <FlexItem spacer={{ default: 'spacerSm' }}>Solve Dependencies </FlexItem>
+    <FlexItem>
+      <Popover
+        bodyContent={
+            __('This will solve RPM and Module Stream dependencies on every publish of this Content View. ' +
```
jeremylenz, line 28:
```suggestion
            __('This will solve RPM and module stream dependencies on every publish of this content view. ' +
```

webpack/scenes/ContentViews/Create/CreateContentViewModal.js
```diff
@@ -0,0 +1,28 @@
...
+import CreateContentViewForm from './CreateContentViewForm';
+
+const CreateContentViewModal = ({ show, setIsOpen }) => (
+  <Modal
+    title="Create Content View"
```
jeremylenz, line 8: Should check with Maria to make sure we're getting the capitalization right for the headings.

webpack/scenes/ContentViews/Create/CreateContentViewForm.js
```diff
@@ -0,0 +1,165 @@
...
+      </FormGroup>
+      <FormGroup isInline fieldId="component">
+        <Radio
+          id="component"
+          name="component"
+          role="radio"
+          label={LabelComponent()}
+          isChecked={component}
+          onChange={(checked) => { setComponent(checked); setComposite(!checked); }}
+          description="Consists of repositories"
```
jeremylenz, line 106: not sure if I missed this, but I really like the simplicity of these descriptions! :+1:

webpack/scenes/ContentViews/Create/CreateContentViewFormComponents.js
```diff
@@ -0,0 +1,93 @@
...
+
+export const LabelImportOnly = () => (
+  <Flex>
+    <FlexItem spacer={{ default: 'spacerSm' }}>Import Only </FlexItem>
+    <FlexItem>
+      <Popover
+        bodyContent={
+          __('Designate whether this Content View is for importing from an upstream server. Import-only Content Views can not be published directly.')
+        }
+        aria-label="Import Only"
```
jeremylenz, line 80:
```suggestion
        aria-label="Import-only"
```

webpack/scenes/ContentViews/Create/CreateContentViewFormComponents.js
```diff
@@ -0,0 +1,93 @@
...
+export const LabelAutoPublish = () => (
+  <Flex>
+    <FlexItem spacer={{ default: 'spacerSm' }}>Auto Publish </FlexItem>
+    <FlexItem>
+      <Popover
+        bodyContent={
+            __('Automatically publish a new version of the composite content view whenever one of its component content views is published. ' +
+          '(Also note auto-publish will only happen for component views that use the "Always use latest" option.)')
+          }
+        aria-label="Auto Publish"
```
jeremylenz, line 57: To be consistent with the new details page this should be
```suggestion
        aria-label="Autopublish"
```
(however, let's talk about that at our next meeting)

jeremylenz (commented):
Thanks @sjha4 

Just a few style nits below, and then I think this will be good to go

webpack/scenes/ContentViews/Create/__tests__/createContentView.test.js
```diff
@@ -0,0 +1,90 @@
...
+
+  assertNockRequest(createscope, done);
+});
+
+test('Form closes itself upon save', async (done) => {
+  const createscope = nockInstance
+    .post(cvCreatePath, createDetails)
+    .reply(201, createdCVDetails);
+  const { queryByText, getByLabelText } = renderWithRedux(form);
+  expect(queryByText('Description')).toBeInTheDocument();
```
johnpmitsch, line 49: A subtle thing with react-testing-library that took me a bit to figure out is when to use `getBy*` and `queryBy*`, I finally landed on:
`getBy*` is for when you expect the element to be present
`queryBy*` is for when you don't expect the element to be present

The big difference is the error message for `getBy*` is more informative since it raises an error when the element is not there, vs. `queryBy*` returns null and you get an error message about not being able to find null

Can you update anything that is expected to be there to `getBy*`?

sjha4: Updated to use getBy* for expected elements and queryBy for elements not expected to be present.
The difference I saw 
getByText('Description')).not.toBeInTheDocument(); was it doesn't fail at _.not.toBeInTheDocument()_ but instead fails at getByText itself if element is not in document.

webpack/scenes/ContentViews/Create/CreateContentViewFormComponents.js
johnpmitsch: Good improvement! How do you feel about these tweaks?

> Automatically publish a new version of the composite content view whenever one of its content views is published.
Auto-publish will only happen for component views that use the 'Always use latest version' option.

Removing the parenthesis and the `Also note` as it seems like important enough information to state directly, and removing the 'component' part from content views, we haven't  really settled on a term and I think the context makes it clear that it's a 'child' content view.

jeremylenz: @johnpmitsch :+1:  to the above

webpack/scenes/ContentViews/Create/CreateContentViewModal.js
johnpmitsch: +1 I do see it lower case on the button and the modal title in the mockup, we should change it to that or confirm with Maria that it can be changed.

johnpmitsch:
@sjha4 The tooltips don't seem to be working on hover, only on click. This is different than the details page, is this intentional?

webpack/scenes/ContentViews/Create/CreateContentViewFormComponents.js
```diff
@@ -0,0 +1,93 @@
...
+      <Popover
+        bodyContent={
+            __('This will solve RPM and module stream dependencies on every publish of this content view. ' +
+            'Dependency solving significantly increases publish time (publishes can take over three times as long) ' +
+            'and filters will be ignored when adding packages to solve dependencies. ' +
+            'Also, certain scenarios involving errata may still cause dependency errors.')
+          }
+        aria-label="Solve-dependencies"
+        position="right"
+      >
```
johnpmitsch, line 35: Seems like we should DRY this up with webpack/scenes/ContentViews/Details/ContentViewInfo.js, looking at the structure sharing just the text would make the most sense IMO

webpack/scenes/ContentViews/Create/CreateContentViewFormComponents.js
```diff
@@ -0,0 +1,93 @@
...
+
+export const LabelAutoPublish = () => (
+  <Flex>
+    <FlexItem spacer={{ default: 'spacerSm' }}>Auto Publish </FlexItem>
+    <FlexItem>
+      <Popover
+        bodyContent={
+            __('Automatically publish a new version of the composite content view whenever one of its component content views is published. ' +
+          '(Also note auto-publish will only happen for component views that use the "Always use latest" option.)')
+          }
```
johnpmitsch, line 56: Same note as above, can we share this text with ContentViewInfo.js? We can use the updated version there too

johnpmitsch:
I'm not seeing the solve dependencies option working, it doesn't seem to be saving with that attribute set to true

johnpmitsch:
@sjha4 Some comments, but overall looks and works well! Nice use of hooks and testing :+1:

sjha4:
> @sjha4 The tooltips don't seem to be working on hover, only on click. This is different than the details page, is this intentional?

It's a popover with a click..Should probably change to tooltip..will look into using the text from details page..

> I'm not seeing the solve dependencies option working, it doesn't seem to be saving with that attribute set to true

Looks like I let eslint --fix change that param to a different name.. :)

webpack/scenes/ContentViews/Create/__tests__/createContentView.test.js
```diff
@@ -0,0 +1,94 @@
...
+  fireEvent.change(getByLabelText('input_name'), { target: { value: '1232123' } });
+
+  await patientlyWaitFor(() => { expect(getByLabelText('input_label')).toHaveAttribute('value', '1232123'); });
+
+  getByLabelText('create_content_view').click();
+
+  assertNockRequest(createscope, done);
+});
+
+test('Form closes itself upon save', async (done) => {
```
johnpmitsch, line 44: nitpick - you are testing the save functionality here, I would argue that is the bigger part you are testing, so could call this `Content view create call is sent to API and modal closes upon save`

webpack/scenes/ContentViews/Create/CreateContentViewModal.js
sjha4: Updated to use proper case from mockups..

webpack/scenes/ContentViews/Create/CreateContentViewFormComponents.js
sjha4: Updated with the new text. :+1:

webpack/scenes/ContentViews/helpers.js
```diff
@@ -0,0 +1,74 @@
...
+  'Also, certain scenarios involving errata may still cause dependency errors.');
+
+export const autoPublishHelpText = __('Automatically publish a new version of the composite content view whenever one of its content views is published. ' +
+  'Autopublish will only happen for component views that use the \'Always use latest version\' option.');
+
+export const importOnlyHelpText = __('Designate whether this content view is for importing from an upstream server. ' +
+  'Import-only content views can not be published directly.');
+
+
+export const LabelComposite = () => (
```
johnpmitsch, line 35: Are these `Label*` components really helpers? I'm only seeing them used in one place and I don't think of a helpers file having components, I think they could be in the file they are being used in or a separate file.

sjha4: Re-using LabelAutoPublish in the info page right now. I think you wanted to move it out of the form page in a separate file and I had it in a create_form_component.js file but then moved it to helper file to keep it together..

webpack/scenes/ContentViews/ContentViewsActions.js
```diff
@@ -1,6 +1,8 @@
-import { API_OPERATIONS, get } from 'foremanReact/redux/API';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { API_OPERATIONS, get, post } from 'foremanReact/redux/API';
```
jeremylenz, line 2: Check out https://github.com/theforeman/foreman/pull/8153

Now that this has been merged, it may simplify our API-handling code even more!
(great work @amirfefer)

johnpmitsch:
> Should we hide the import-only option for composite content views? This is what happens when trying to use that option:
> 
> `Validation failed: Import only Import-only Content Views can not be Composite`

@sjha4 Changes look good, but still wondering about this? I see it disappear in the current page's form but not seeing it disappear here.

Bonus points if you want to update to the API hooks as mentioned by @jeremylenz but since those were just merged I think it's fair to do as a follow-up.

sjha4:
> > Should we hide the import-only option for composite content views? This is what happens when trying to use that option:
> > `Validation failed: Import only Import-only Content Views can not be Composite`
> 
Pushed the change. :+1: 

> 
> Bonus points if you want to update to the API hooks as mentioned by @jeremylenz but since those were just merged I think it's fair to do as a follow-up.

I propose merging this with the current middleware usage and trying the new useAPI hook in a fresh PR.

webpack/scenes/ContentViews/helpers.js
```diff
@@ -0,0 +1,13 @@
...
+export const dependenciesHelpText = __('This will solve RPM and module stream dependencies on every publish of this content view. ' +
+  'Dependency solving significantly increases publish time (publishes can take over three times as long) ' +
+  'and filters will be ignored when adding packages to solve dependencies. ' +
+  'Also, certain scenarios involving errata may still cause dependency errors.');
+
+export const autoPublishHelpText = __('Automatically publish a new version of the composite content view whenever one of its content views is published. ' +
+  'Autopublish will only happen for component views that use the \'Always use latest version\' option.');
+
+export const importOnlyHelpText = __('Designate whether this content view is for importing from an upstream server. ' +
+  'Import-only content views can not be published directly.');
```
jeremylenz, line 12:
```suggestion
  'Import-only content views cannot be published directly.');
```

jeremylenz:
> I propose merging this with the current middleware usage and trying the new useAPI hook in a fresh PR.

ACK after addressing the one typo above :)

johnpmitsch (approved):
Looks great! :+1:

sjha4:
Merging. Thanks @johnpmitsch and @jeremylenz for guiding this along!

PR #9066, opened 2020-12-04
Fixes #31121 - add migration timing estimate to pulp 3 migration stats

ianballou:
To test:

1) Have a Katello server running with Pulp 2 for yum
2) Sync a variety of yum content, both immediate and on-demand
3) Run `rake katello:pulp3_migration_stats` and see the timing estimate.  Check that it makes sense and adheres to the formula:
```
# On Demand RPMs: (6.46E-04)*(#RPMs) + -3.22
# Immediate RPMs: (9.39E-04)*(#RPMs) + -3
# Repositories: 0.0746*(#Repos) + -2.07

Total the 3 to find the estimate.  Note that the y-intercept is negative due to the linear trend lines.  Therefore, if negative, we assume it'll take fewer than 5 minutes.
```

Note that if you try testing the timing estimate on a development box, it is likely to take more time than estimated.

### DIFF
--- a/lib/katello/tasks/pulp3_migration_stats.rake
+++ b/lib/katello/tasks/pulp3_migration_stats.rake
@@ -3,17 +3,37 @@ namespace :katello do
   task :pulp3_migration_stats => [:environment] do
     User.current = User.anonymous_admin
 
-    migrated_rpms = ::Katello::Rpm.where('pulp_id LIKE ?', '%/pulp/api/v3/content/rpm/packages/%').
+    migrated_rpm_count = ::Katello::Rpm.where('pulp_id LIKE ?', '%/pulp/api/v3/content/rpm/packages/%').
       or(::Katello::Rpm.where.not(migrated_pulp3_href: nil)).count
-    migrated_errata = ::Katello::RepositoryErratum.where.not(erratum_pulp3_href: nil).count
-    migrated_repos = ::Katello::Repository.where.not(version_href: nil).count
-    migratable_repos = ::Katello::Repository.count - ::Katello::Repository.puppet_type.count -
+    migrated_erratum_count = ::Katello::RepositoryErratum.where.not(erratum_pulp3_href: nil).count
+    migrated_repo_count = ::Katello::Repository.where.not(version_href: nil).count
+    migratable_repo_count = ::Katello::Repository.count - ::Katello::Repository.puppet_type.count -
       ::Katello::Repository.ostree_type.count - ::Katello::Repository.deb_type.count
 
+    on_demand_rpm_count = Katello::RepositoryRpm.where(:repository_id => Katello::Repository.yum_type.on_demand).distinct.count
+    on_demand_unmigrated_rpm_count = on_demand_rpm_count - migrated_rpm_count
+    immediate_unmigrated_rpm_count = ::Katello::Rpm.count - migrated_rpm_count - on_demand_unmigrated_rpm_count
+
+    # On Demand RPMs: (6.46E-04)*(#RPMs) + -3.22
+    # Immediate RPMs: (9.39E-04)*(#RPMs) + -3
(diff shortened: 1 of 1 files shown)


lib/katello/tasks/pulp3_migration_stats.rake
```diff
@@ -3,17 +3,39 @@ namespace :katello do
...
-    migrated_repos = ::Katello::Repository.where.not(version_href: nil).count
-    migratable_repos = ::Katello::Repository.count - ::Katello::Repository.puppet_type.count -
+    migrated_erratum_count = ::Katello::RepositoryErratum.where.not(erratum_pulp3_href: nil).count
+    migrated_repo_count = ::Katello::Repository.where.not(version_href: nil).count
+    migratable_repo_count = ::Katello::Repository.count - ::Katello::Repository.puppet_type.count -
       ::Katello::Repository.ostree_type.count - ::Katello::Repository.deb_type.count
 
+    immediate_rpms = ::Katello::Repository.yum_type.collect do |repo|
+      repo.rpms unless repo.on_demand?
+    end
```
jlsherrill, line 15: I think you could turn this into a query that would be much faster.  Also i would imagine if you can stick to rpm ids, it'd be faster too

something like:
Katello::RepositoryRpm.where(:repository_id => Katello::Repository.yum_type).select(:id).uniq

jlsherrill: updated previous comment to fix .select()

ianballou: Yeah, I got a bit stuck on the on_demand check. I'll have to think about how to add that in.

jlsherrill: i think there's an 'on_demand' scope

ianballou:
@jlsherrill I've improved the rpm query.

lib/katello/tasks/pulp3_migration_stats.rake
jeremylenz: Also `.distinct` may be a bit faster if that's possible, since it will do the duplicate detection via SQL instead of Ruby.

jlsherrill: oh yeah, and since really all you care about is the count, you can drop the select:

Katello::RepositoryRpm.where(:repository_id => Katello::Repository.yum_type.on_demand).distinct.count

lib/katello/tasks/pulp3_migration_stats.rake
```diff
@@ -3,17 +3,38 @@ namespace :katello do
...
+      select(:rpm_id).uniq.count
+    on_demand_unmigrated_rpm_count = on_demand_rpm_count - migrated_rpm_count
+    immediate_unmigrated_rpm_count = ::Katello::Rpm.count - migrated_rpm_count - on_demand_unmigrated_rpm_count
+
+    # On Demand RPMs: (6.46E-04)*(#RPMs) + -3.22
+    # Immediate RPMs: (9.39E-04)*(#RPMs) + -3
+    # Repositories: 0.0746*(#Repos) + -2.07
+    migration_minutes = (0.000646 * on_demand_unmigrated_rpm_count - 3.22 +
+                         0.000943 * immediate_unmigrated_rpm_count - 3 +
+                         0.0746 * migratable_repo_count).to_i
```
jlsherrill, line 23: trying to understand this, is the 0.0746 meaning roughly .07 of a minute? (i.e. 4 seconds) ?

ianballou: The formulas here are linear trendlines for the data I collected. The time units for the y-axis were in minutes. 0.0746 was just the slope that came along with the trendline equation.

ianballou:
Should be all set.

jlsherrill (approved):

ianballou:
[test katello]

PR #9067, opened 2020-12-07
Fixes #31465 - show sync widget for SP with only pulp3

jlsherrill:


app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -214,7 +214,7 @@ def redirect_client
 
     def pull_blob
       headers = {}
-      headers['Accept'] = request.headers['Accept'] if request.headers['Accept']
+      #headers['Accept'] = request.headers['Accept'] if request.headers['Accept']
```
ianballou, line 217: Leftovers? :)

ianballou (commented):
Some issues I noticed:
1) The "Lifecycle Environments" tab is still missing from the "Edit smart proxy" page
2) The disk space indicator is gone (I'm guessing that's more of a pulpcore proxy plugin issue?)

app/controllers/katello/api/registry/registry_proxies_controller.rb
jlsherrill: hah yes!

jlsherrill:
1) good catch on this!
2) looking into disk widget, we may pull that into a 2nd isssue/pr, let me see how involved it is

jlsherrill:
@ianballou looking into the pulp disk stuff i think it might be better as a 2nd pr in which we could a) remove the pulp2 part & b) add the pulp3 part.  I opened an issue here:  https://projects.theforeman.org/issues/31506  I'm open to reconsider if you disagree.

EDIT; Actually i realize we probably need to keep the pulp2 part to support older smart proxies, i will update the issue to focus on just adding support for pulp3

ianballou:
@jlsherrill putting the disk usage stuff in a 2nd PR sounds good to me. I figured that might be the case.

ianballou (approved):
Looks good to me!

PR #9071, opened 2020-12-08
Fixes #31479 - If org has SCA enabled, show SCA for hypervisors

chris1984:
This PR adds the correct status message to hypervisors when an org has SCA enabled.

Screenshot with patch:

https://nimbusweb.me/s/share/4963201/5xp3idmmzci8nws453nl

To test:

* Create an org if not using a fresh devel box
* Import a manifest
* Toggle SCA on
* Register a content host
* Install virt-who on the content host and have it report in
* Check the subscription status of the hypervisor in the content host page.

### DIFF
--- a/app/models/katello/subscription_status.rb
+++ b/app/models/katello/subscription_status.rb
@@ -43,9 +43,10 @@ module Katello
 
     def to_status(options = {})
       return UNKNOWN unless host.subscription_facet.try(:uuid)
-      status_override = 'unsubscribed_hypervisor' if host.subscription_facet.hypervisor && !host.subscription_facet.candlepin_consumer.entitlements?
+      return DISABLED if host.organization.simple_content_access?
+      status_override = 'unsubscribed_hypervisor' if host.subscription_facet.unsubscribed_hypervisor?
       status_override ||= options.fetch(:status_override, nil)
-      status = status_override || Katello::Candlepin::Consumer.new(host.subscription_facet.uuid, host.organization.label).entitlement_status
+      status = status_override || host.subscription_facet.candlepin_consumer.entitlement_status
 
       case status
       when Katello::Candlepin::Consumer::ENTITLEMENTS_DISABLED
--- a/test/models/subscription_status_test.rb
+++ b/test/models/subscription_status_test.rb
@@ -8,6 +8,10 @@ module Katello
 
     let(:status) { host.get_status(Katello::SubscriptionStatus) }
 
+    def setup
+      ::Organization.any_instance.stubs(:simple_content_access?).returns(false)
+    end
+
     def stub_status(status)
       Katello::Candlepin::Consumer.any_instance.stubs(:entitlement_status).returns(status)
     end
@@ -37,8 +41,13 @@ module Katello
     end
 
     def test_to_status_unsubscribed_hypervisor
-      stub_status('unsubscribed_hypervisor')
+      host.subscription_facet.stubs(:unsubscribed_hypervisor?).returns(true)
       assert_equal Katello::SubscriptionStatus::UNSUBSCRIBED_HYPERVISOR, status.to_status
+    end
+
+    def test_sca_enabled
+      ::Organization.any_instance.stubs(:simple_content_access?).returns(true)
+      assert_equal Katello::SubscriptionStatus::DISABLED, status.to_status
     end
 
     def test_no_subscription_facet
--- a/app/models/katello/host/subscription_facet.rb
+++ b/app/models/katello/host/subscription_facet.rb
@@ -282,6 +282,10 @@ module Katello
         name.gsub('_', '-').chomp('.').downcase
       end
 
+      def unsubscribed_hypervisor?
+        self.hypervisor && !self.candlepin_consumer.entitlements?
+      end
+
       def candlepin_consumer
         @candlepin_consumer ||= Katello::Candlepin::Consumer.new(self.uuid, self.host.organization.label)
       end
--- a/test/actions/katello/host/hypervisors_test.rb
+++ b/test/actions/katello/host/hypervisors_test.rb
@@ -72,6 +72,7 @@ module Katello::Host
       data = data.with_indifferent_access
       create_org(org_label)
       org = Organization.find_by(label: org_label)
+      ::Organization.any_instance.stubs(:simple_content_access?).returns(false)
       host_names_expected = ["virt-who-2e78f643-1d2f-45d1-b191-d931147cbde1-#{org.id}", "virt-who-261c4dca-702f-42b3-b8ef-2a72b77f7ec2-#{org.id}"]
 
       assert_equal 0, org.hosts.count
@@ -92,6 +93,7 @@ module Katello::Host
       data = data.with_indifferent_access
       create_org(org_label)
       org = Organization.find_by(label: org_label)
+      ::Organization.any_instance.stubs(:simple_content_access?).returns(false)
       host_names_expected = ["virt-who-more-useful-identifier-#{org.id}", "virt-who-261c4dca-702f-42b3-b8ef-2a72b77f7ec2-#{org.id}"]
       reporter_id = 100
       task = Katello::Resources::Candlepin::Consumer.async_hypervisors(owner: org_label, reporter_id: reporter_id, raw_json: data.to_json)
--- a/test/actions/katello/host/hypervisors_update_test.rb
+++ b/test/actions/katello/host/hypervisors_update_test.rb
@@ -48,6 +48,8 @@ module Katello::Host
       @host.subscription_facet.destroy!
       @host.destroy!
 
+      ::Organization.any_instance.stubs(:simple_content_access?).returns(false)
+
       ::Katello::Resources::Candlepin::Consumer.expects(:virtual_guests).never
 
       action = create_action(::Actions::Katello::Host::HypervisorsUpdate)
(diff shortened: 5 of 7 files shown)


test/models/subscription_status_test.rb
```diff
@@ -41,6 +41,11 @@ def test_to_status_unsubscribed_hypervisor
       assert_equal Katello::SubscriptionStatus::UNSUBSCRIBED_HYPERVISOR, status.to_status
     end
 
+    def test_hypervisor_sca_enabled
```
jturel, line 44: Does this test pass without your code change in place?

chris1984: I will comment it out and let you know :)

app/models/katello/subscription_status.rb
```diff
@@ -43,7 +43,13 @@ def to_global(_options = {})
 
     def to_status(options = {})
       return UNKNOWN unless host.subscription_facet.try(:uuid)
-      status_override = 'unsubscribed_hypervisor' if host.subscription_facet.hypervisor && !host.subscription_facet.candlepin_consumer.entitlements?
+      if host.subscription_facet.hypervisor
```
jturel, line 46: This looks a bit complex now, so I'm questioning how optimal the approach is. The code (before you modified it) also made me wonder why we handle unsubscribed_hypervisor separately. And I guess it's because we need information from the subscription facet so we can't rely completely on the `Katello::Candlepin::Consumer` service class.

As far as this method is concerned, we ultimately want to 'not care' about unsubscribed_hypervisor if the org is in sca mode. So you should check SCA first, then based on that check for unsubscribed hypervisor. Here's how the code might be refactored:

To help us move some of this logic out, edit  subscription_facet.rb
```
def unsubscribed_hypervisor?
  !self.organization.simple_content_access && self.hypervisor && self.candlepin_consumer.entitlements?
end
```

Now we can use it here:
```
status_override = 'unsubscribed_hypervisor' if host.subscription_facet.unsubscribed_hypervisor?
```

app/models/katello/subscription_status.rb
```diff
@@ -43,7 +43,13 @@ def to_global(_options = {})
...
+        if host.organization.simple_content_access?
+          status_override = 'simple_content_access'
+        elsif !host.subscription_facet.candlepin_consumer.entitlements?
+          status_override = 'unsubscribed_hypervisor'
+        end
+      end
       status_override ||= options.fetch(:status_override, nil)
       status = status_override || Katello::Candlepin::Consumer.new(host.subscription_facet.uuid, host.organization.label).entitlement_status
```
jturel, line 54: Extra credit: The `unsubscribed_hypervisor?` check may have already initialized a Katello::Candlepin::Consumer object. See subscription_facet.rb#candlepin_consumer So, we can use it again here:

```
status = status_override || host.subscription_facet.candlepin_consumer.entitlement_status
```

app/models/katello/subscription_status.rb
```diff
@@ -58,6 +64,8 @@ def to_status(options = {})
         INVALID
       when 'unsubscribed_hypervisor'
         UNSUBSCRIBED_HYPERVISOR
+      when 'simple_content_access'
```
jturel, line 67: I don't think this is needed based on my suggestions

test/models/subscription_status_test.rb
jturel: BTW the reason I am asking is because it doesn't appear that this test is using a simple content access org

app/models/katello/subscription_status.rb
chris1984: Removed

app/models/katello/subscription_status.rb
chris1984: Updated to this, thank you for the pointer

app/models/katello/subscription_status.rb
chris1984: Updated to this, thank you for the pointer and help.

chris1984:
@jturel updated, had to stub `simple_content_access?` on the tests because vcr errors kept failing. Let me know if that looks fine?

test/models/subscription_status_test.rb
```diff
@@ -17,27 +17,32 @@ def test_get_status
...
     def test_to_status_invalid
+      host.organization.stubs(:simple_content_access?).returns(false)
       stub_status(Katello::Candlepin::Consumer::ENTITLEMENTS_INVALID)
       assert_equal Katello::SubscriptionStatus::INVALID, status.to_status
     end
 
     def test_to_status_unsubscribed_hypervisor
-      stub_status('unsubscribed_hypervisor')
+      host.organization.stubs(:simple_content_access?).returns(false)
+      host.subscription_facet.stubs(:unsubscribed_hypervisor?).returns(true)
```
jturel, line 45: Are both stubs needed?

jturel: I think you had a new test case here - where'd it go? 🔍

chris1984: Got lost with my editing, adding it back and removed the uneeded sub on line 44

test/models/subscription_status_test.rb
```diff
@@ -17,30 +17,40 @@ def test_get_status
...
 
     def test_to_status_unsubscribed_hypervisor
-      stub_status('unsubscribed_hypervisor')
+      host.subscription_facet.stubs(:unsubscribed_hypervisor?).returns(true)
       assert_equal Katello::SubscriptionStatus::UNSUBSCRIBED_HYPERVISOR, status.to_status
     end
 
+    def test_sca_enabled_hypervisor
+      host.subscription_facet.stubs(:unsubscribed_hypervisor?).returns(false)
+      stub_status(Katello::Candlepin::Consumer::ENTITLEMENTS_DISABLED)
```
jturel, line 50: is stub_status needed? ie does the code path which is getting stub get reached?

test/models/subscription_status_test.rb
```diff
@@ -17,30 +17,40 @@ def test_get_status
     end
 
     def test_to_status_valid
+      host.organization.stubs(:simple_content_access?).returns(false)
```
jturel, line 20: optional: since most of these test cases use this same stub, it *could* go in a before method at the top. up to you

test/models/subscription_status_test.rb
chris1984: The `stub_status` is needed because it fails with a vcr error, looking at pry it fails on this line unless I stub it:

https://github.com/Katello/katello/pull/9071/files#diff-746cf731242664a742efafbdcd690f3bceb15b60d66a320d25e3f8aeea73ff3cR48

test/models/subscription_status_test.rb
chris1984: Refactored to a `setup` method. Will start to look at the other test failures.

test/models/subscription_status_test.rb
```diff
@@ -37,10 +41,16 @@ def test_to_status_invalid
     end
 
     def test_to_status_unsubscribed_hypervisor
-      stub_status('unsubscribed_hypervisor')
+      host.subscription_facet.stubs(:unsubscribed_hypervisor?).returns(true)
       assert_equal Katello::SubscriptionStatus::UNSUBSCRIBED_HYPERVISOR, status.to_status
     end
 
+    def test_sca_enabled_hypervisor
+      host.subscription_facet.stubs(:unsubscribed_hypervisor?).returns(false)
```
jturel, line 49: If I am understanding the BZ properly, we need to show DISABLED status for hypervisors in SCA mode. So the assertion for this test is correct but the setup is wrong . The unsubscribed_hypervisor? stub should be returning true here. Does that sound right?

In which case I think the below stub_status could be eliminated, and if not let's understand why.

jturel: Ah, right. unsubscribed_hypervisor itself checks for sca mode. Maybe that should be removed and then the status class can do this:

```
status_override = Katello::Candlepin::Consumer::ENTITLEMENTS_DISABLED if host.organization.simple_content_access?
status_override ||= 'unsubscribed_hypervisor' if host.subscription_facet.unsubscribed_hypervisor?
status_override ||= options.fetch(:status_override, nil)
status = status_override || host.subscription_facet.candlepin_consumer.entitlement_status
```

That will make the test case more clear so unsubscribed_hypervisor can be stubbed to true which illustrates the requirement. Don't think that will mess with any tests you've encountered

jturel: Let me look at this code a bit more. I think we need a small adjustment.

jturel: This method probably needs a bit of a refactor, but let's keep it simple for now. Try (and test) my suggestion above

chris1984:
@jturel thanks again for the help and review, updated and still working. Screenshot from the UI:

https://nimbusweb.me/s/share/4989181/7urdfat0a1tiaysfkszm

test/models/subscription_status_test.rb
```diff
@@ -37,10 +41,15 @@ def test_to_status_invalid
     end
 
     def test_to_status_unsubscribed_hypervisor
-      stub_status('unsubscribed_hypervisor')
+      host.subscription_facet.stubs(:unsubscribed_hypervisor?).returns(true)
       assert_equal Katello::SubscriptionStatus::UNSUBSCRIBED_HYPERVISOR, status.to_status
     end
 
+    def test_sca_enabled_hypervisor
```
jturel, line 48: nitpick: since we now check for SCA at the top of the status method, the host type (hypervisor or not) is irrelevant. so, would prefer test_sca_enabled here

jturel:
looks good! i know the testing was a bit wild with this one, so you might wanna check that all the stubs added now are actually necessary if you didn't already, though I suspect they're valid even after all the changes

test/models/subscription_status_test.rb
chris1984: > looks good! i know the testing was a bit wild with this one, so you might wanna check that all the stubs added now are actually necessary if you didn't already, though I suspect they're valid even after all the changes

Sounds good, will address the nit and test commenting out the stubs and see what fails.

jturel (approved):
APJ

PR #9075, opened 2020-12-11
Fixes #31496: Fix error when allSelected() not yet defined

ehelms:
When assets are compiled, intermittently during the loading of
the products page, the self.table.allSelected() function is not
yet defined when this segment of code is first encountered. This
is a result of the products page calling refresh after disabling
load rather than calling load explicitly to enable repo discovery
button.


ehelms:
@ekohl This was an issue identified by smoker intermittently

jturel (approved):
I don't see the error any more with this PR. Thanks @ehelms

PR #9076, opened 2020-12-11
Fixes #31498 - de-duplicate activation key names at registration time

jturel:
previously `subscription-manager register --activationkey="keyOne,keyOne"` would break. Now we sanitize the given keys!


parthaa:
Code looks good

parthaa (approved):

parthaa:
Tested. Ack

PR #9077, opened 2020-12-14
Fixes #31504 - improved content-import error handling

parthaa:
This commit tries to gracefully handle the case where the foreman user
is not of pulp group and removes that requirement.
Ti achieve this it  does the following:
1) Ignores checking the 'pulp' permissions in the import path. This job
   is best left to pulp (https://pulp.plan.io/issues/7549)
2) Makes sure that the toc name is gotten from the exporter task output
   rather than scanning the directory. This removes the requirement of
   foreman/vagrant user having to check a pulp directory.
3) Removes the import 'skip' strategy on failed import. Instead it
   pauses on error and requires the user to skip and delete the content
   view version.


chris1984 (approved):
ACK, was able to export and import without any permission errors:
```bash
[vagrant@dhcp182-208 ~]$  hammer content-export complete version --content-view=view --organization=export-6123  --version='1.0'
[...............................................................................................................................................................................................................................][100%]
Generated /var/lib/pulp/exports/export-6123/view/1.0/2020-12-15T17-45-55-05-00/metadata.json

[vagrant@dhcp182-208 ~]$ hammer content-import version --organization=import-8506 --path=/var/lib/pulp/exports/export-6123/view/1.0/2020-12-15T17-45-55-05-00/ --content-view-id=6
[...............................................................................................................................................................................................................................] [100%]
```

PR #9079, opened 2020-12-17
Fixes #27772 - Use sync_history for content details on smart proxy

sjha4:
This PR removes pulp2 calls to display content details for a smart proxy and uses smart_proxy_history instead.

To test:

Create a few repos, cv etc and sync a smart proxy and go to content tab. The page should look like it does today on master.


app/services/katello/pulp3/smart_proxy_repository.rb
```diff
@@ -82,6 +82,21 @@ def orphan_repositories
         end
         repo_map
       end
+
+      def current_repositories_data(environment = nil, content_view = nil)
```
sjha4, line 86: This is not needed right now since the smart_proxy_service is always pulp2 here, not sure if or what to do about it?

https://github.com/Katello/katello/blob/3e3dc6dae8c2b0c01e62ab8cf6db935a920a430c/app/models/katello/concerns/smart_proxy_extensions.rb#L349

app/services/katello/pulp/content_counts_calculator.rb
```diff
@@ -24,18 +24,18 @@ def calculate
...
+            counts[:package_groups] += repo.package_groups.count
+            counts[:errata] += repo.errata.count
           when repo_type?(repo, 'docker')
             counts[:docker_repositories] += 1
-            counts[:docker_images] += get_unit_count(repo, 'docker_image')
+            counts[:docker_images] += repo.docker_manifests.count
           when repo_type?(repo, 'puppet')
             counts[:puppet_repositories] += 1
-            counts[:puppet_modules] += get_unit_count(repo, 'puppet_module')
+            counts[:puppet_modules] += repo.puppet_modules.count
```
sjha4, line 38: Puppet count won't work since we don't maintain puppet repo sync history. We could remove the puppet column from the page?

jlsherrill:
When we had discussed this previously it was mentioned that the actual counts aren't very helpful, and we thought about replacing it with a very simple table of repos and their last sync time (from the sync history) (possibly organized by CV/LCE).    Thoughts on that?  

I'm not opposed going this route, but i think the display of content counts isn't all that useful.

sjha4:
> When we had discussed this previously it was mentioned that the actual counts aren't very helpful, and we thought about replacing it with a very simple table of repos and their last sync time (from the sync history) (possibly organized by CV/LCE). Thoughts on that?
> 
> I'm not opposed going this route, but i think the display of content counts isn't all that useful.

The only place where the content counts can be of help between 2 capsules on different env  or some env vs library is when there are filters and small number of cvs and repos..I don't see how the raw count would be all that helpful in larger deployments..

jlsherrill:
Still testing, but a couple of issues i found:
1) if i associate to a LCE and click the content ta b(without syncing), i get:
```
Uncaught TypeError: Cannot read property 'forEach' of undefined
    at SmartProxyContentTable.js:77
```
2) if pulp isn't installed on the capsule, the content tab doesn't show up, i think thats covered here:  https://github.com/Katello/katello/blob/master/lib/katello/plugin.rb#L266   mind updating that?

jlsherrill:
actually i get that error even with history entries:
```SmartProxyContentTable.js:77 Uncaught TypeError: Cannot read property 'forEach' of undefined
    at SmartProxyContentTable.js:77
    at buildrows (SmartProxyContentTable.js:70)
    at SmartProxyContentTable.js:166
    at commitHookEffectList (react-dom.development.js:19986)
    at commitPassiveHookEffects (react-dom.development.js:20016)
    at HTMLUnknownElement.callCallback (react-dom.development.js:347)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:397)
    at invokeGuardedCallback (react-dom.development.js:454)
    at flushPassiveEffectsImpl (react-dom.development.js:22868)```

File: app/services/katello/pulp3/smart_proxy_repository.rb [hunk 3]
```diff
@@ -82,6 +82,21 @@ def orphan_repositories
...
+        @pulp_repositories ||= smart_proxy.smart_proxy_sync_histories.map { |sync_history| sync_history.repository }
+
+        repos = Katello::Repository
+        repos = repos.in_environment(environment) if environment
+        repos = repos.in_content_views([content_view]) if content_view
+        puppet_envs = Katello::ContentViewPuppetEnvironment
+        puppet_envs = puppet_envs.in_environment(environment) if environment
+        puppet_envs = puppet_envs.in_content_view(content_view) if content_view
```
jlsherrill, line 94: we can just delete this

jlsherrill: the content view puppet environment code

jlsherrill:
* get rid of content counts in API 
* Consolidate repo counts into just 'repositories'
* remove products column
* Up to date? (using history event)
* sync_status.json.rabl needs to look for PULPCORE feature too (not just PULP_NODE)

app/services/katello/pulp3/smart_proxy_repository.rb
sjha4: Deleted this for pulp3 service, still present for pulp service class..Not removing it there..

sjha4:
> * get rid of content counts in API
> * Consolidate repo counts into just 'repositories'
> * remove products column
> * Up to date? (using history event)
> * sync_status.json.rabl needs to look for PULPCORE feature too (not just PULP_NODE)

Screenshot from 2021-01-19 12-28-24
 Current columns and headings. Added few icons. :smile:  The content view type icon is what we are using in the new CV pages for single/composite.

app/views/katello/api/v2/capsule_content/sync_status.json.rabl
```diff
@@ -18,15 +18,13 @@ child @lifecycle_environments => :lifecycle_environments do
     @capsule.environment_syncable?(env)
   end
 
-  if @capsule.has_feature?(SmartProxy::PULP_NODE_FEATURE)
+  if @capsule.has_feature?(SmartProxy::PULP_NODE_FEATURE) || @capsule.has_feature?(SmartProxy::PULPCORE)
```
jlsherrill, line 21: this should be PULP3_FEATURE  (sorry i typed it from memory in my comment and got it wrong)

jlsherrill:
https://github.com/Katello/katello/blob/master/lib/katello/plugin.rb#L266 still needs updating too

app/services/katello/pulp3/smart_proxy_repository.rb
```diff
@@ -82,6 +82,31 @@ def orphan_repositories
...
+      end
+
+      def repos_in_sync_history
+        smart_proxy.smart_proxy_sync_histories.map { |sync_history| sync_history.repository }
+      end
+
+      def current_repositories_data(environment = nil, content_view = nil)
+        @pulp_repositories ||= repos_in_sync_history
+        repo_ids = repos_in_env_cv(environment, content_view)
+        @pulp_repositories.select { |r| repo_ids.include?(r['pulp_id']) }
```
jlsherrill, line 101: i think we can get rid of all this 'pulp_id' comparisons.  This was there before because we were comparing to to results back from pulp2, but since we're just working within our own db, i think we can simplify it quite a bit

app/services/katello/pulp3/smart_proxy_repository.rb
```diff
@@ -82,6 +82,31 @@ def orphan_repositories
...
+      end
+
+      def up_to_date?(environment = nil, content_view = nil)
+        @pulp_repositories ||= repos_in_sync_history
+        synced_repos = @pulp_repositories.pluck(:pulp_id)
+        repo_ids = repos_in_env_cv(environment, content_view)
+        (repo_ids - synced_repos).empty?
+      end
```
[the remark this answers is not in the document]
jlsherrill: as an example, i think some of this code could be ported to the smart_proxy class.  I would maybe not copy the code as is, but instead implement the needed functionality in the SmartProxy class

app/views/katello/api/v2/capsule_content/sync_status.json.rabl
```diff
@@ -18,15 +18,13 @@ child @lifecycle_environments => :lifecycle_environments do
...
   end
 
-  if @capsule.has_feature?(SmartProxy::PULP_NODE_FEATURE)
+  if @capsule.has_feature?(SmartProxy::PULP_NODE_FEATURE) || @capsule.has_feature?(SmartProxy::PULP3_FEATURE)
     node :counts do |env|
-      counts = {
+      {
         :content_hosts => env.hosts.authorized("view_hosts").count,
         :content_views => env.content_views.non_default.count,
         :products => env.products.enabled.count
```
jlsherrill, line 26: i would probably drop the content_hosts and products api part as wel

webpack/scenes/SmartProxy/__tests__/SmartProxyContentTest.js
```diff
@@ -0,0 +1,31 @@
...
+    .query(true)
+    .reply(200, smartProxyContent);
+
+  const { getByText, queryByText } = renderWithRedux(contentTable);
+  await patientlyWaitFor(() => expect(getByText('Description')).toBeInTheDocument());
+  expect(queryByText('Content view')).toBeInTheDocument();
+  expect(queryByText('Type')).toBeInTheDocument();
+  expect(queryByText('Last published')).toBeInTheDocument();
+  expect(queryByText('Repositories')).toBeInTheDocument();
+  expect(queryByText('Synced to smart proxy')).toBeInTheDocument();
```
johnpmitsch, line 28: these `queryByText` can be `getByText` if they are expected to be in the document, it will show a better error message if it's not present.

webpack/scenes/SmartProxy/SmartProxyContentTable.js
```diff
@@ -0,0 +1,154 @@
...
+    setRow(newRows);
+  };
+
+
+  useEffect(() => fetchWithParams(), []);
+
+  useEffect(() => {
+    if (status !== STATUS.PENDING && response) {
+      const { lifecycle_environments: env } = response;
+      // setLoaded(true);
```
johnpmitsch, line 135: I'm guessing this isn't needed anymore?

sjha4: Updated :+1:

app/views/katello/api/v2/capsule_content/sync_status.json.rabl
sjha4: Updated to address comments..Also moved common methods to smart proxy..( Removed puppet env from pulp2 as well..I'm guessing that's ok?)

jlsherrill (approved):
ACK!

sjha4:
Merging..Thanks @jlsherrill , @johnpmitsch !

PR #9082, opened 2020-12-18
Fixes #31532 - Regenerate entitlement certs on custom repo update for RHEL 6.8 and older

jeremylenz:
The "Restrict to OS Version" feature doesn't work for clients on RHEL 6.8 and older.

This fixes that by doing the following:

* When a custom repo is updated, see if the 'OS Versions' field was changed
* If so, and `rhel-6` is involved, get a list of Candlepin consumers for every client that uses RHEL 6.8 and older
* Send a request to Candlepin to regenerate the entitlement certificates for each consumer in the list

After this change, `subscription-manager refresh` should cause the certificates to be properly refreshed on RHEL 6.8 clients, including the 'Required Tags' field.  This will allow the "Restrict to OS Version" feature to work properly on these clients.

Opening this as a draft because this approach is not ideal.  Ideally, we would only send the request for the _relevant_ consumers, that is, those that are on the older OS _and_ ~~subscribe to the product containing the repo that was just changed~~.  (_update_: this would not work because of SCA) Even more ideally, we could send a single bulk request to Candlepin, but I don't think that's possible.

### DIFF
--- a/app/lib/actions/katello/repository/update.rb
+++ b/app/lib/actions/katello/repository/update.rb
@@ -9,6 +9,10 @@ module Actions
         def plan(root, repo_params)
           repository = root.library_instance
           action_subject root.library_instance
+          action_params = {
+            repository_id: root.library_instance.id,
+            cp_consumers: []
+          }
 
           repo_params[:url] = nil if repo_params[:url] == ''
           update_cv_cert_protected = repo_params.key?(:unprotected) && (repo_params[:unprotected] != repository.unprotected)
@@ -47,17 +51,14 @@ module Actions
             if root.previous_changes.key?('os_versions')
               # If Restrict to OS Version is changing _from_ or _to_ 'rhel-6'...
               if root.previous_changes['os_versions'].any? { |ch| ch.include?('rhel-6') }
-                # ...force regeneration of entitlement certs on RHEL 6.8 and older
+                # ...force regeneration of entitlement certs on RHEL 6.8 and older (see run method)
                 Rails.logger.info("Looking for RHEL 6.8 and older clients")
                 cp_consumers = ::Katello::Host::ContentFacet
                   .joins(host: :operatingsystem)
                   .where("operatingsystems.major = '6' AND operatingsystems.minor < '9' AND operatingsystems.name = 'RedHat'")
                   .map(&:uuid)
-                # TODO: make this an action
-                Rails.logger.info("Marking entitlement certs dirty for #{pluralize(cp_consumers.length, 'client')}")
-                cp_consumers.each do |uuid|
-                  ::Katello::Resources::Candlepin::Entitlement.regenerate_entitlement_certificates_for_consumer(uuid, true)
-                end
+                action_params[:cp_consumers] = cp_consumers
+                action_params[:org_label] = root.organization.label
               end
             end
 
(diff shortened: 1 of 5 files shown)


app/lib/actions/katello/repository/update.rb
```diff
@@ -42,6 +43,24 @@ def plan(root, repo_params)
...
+                Rails.logger.info("Looking for RHEL 6.8 and older clients")
+                cp_consumers = ::Katello::Host::ContentFacet
+                  .joins(host: :operatingsystem)
+                  .where("operatingsystems.major = '6' AND operatingsystems.minor < '9' AND operatingsystems.name = 'RedHat'")
+                  .map(&:uuid)
+                # TODO: make this an action
+                Rails.logger.info("Marking entitlement certs dirty for #{pluralize(cp_consumers.length, 'client')}")
+                cp_consumers.each do |uuid|
+                  ::Katello::Resources::Candlepin::Entitlement.regenerate_entitlement_certificates_for_consumer(uuid, true)
+                end
```
jeremylenz, line 60: If we decide to go with this approach I'll make this a Dynflow action.

app/lib/actions/katello/repository/update.rb
```diff
@@ -42,6 +43,24 @@ def plan(root, repo_params)
...
                                        gpg_url: repository.yum_gpg_key_url)
+
+            if root.previous_changes.key?('os_versions')
+              # If Restrict to OS Version is changing _from_ or _to_ 'rhel-6'...
+              if root.previous_changes['os_versions'].any? { |ch| ch.include?('rhel-6') }
+                # ...force regeneration of entitlement certs on RHEL 6.8 and older
+                Rails.logger.info("Looking for RHEL 6.8 and older clients")
+                cp_consumers = ::Katello::Host::ContentFacet
+                  .joins(host: :operatingsystem)
+                  .where("operatingsystems.major = '6' AND operatingsystems.minor < '9' AND operatingsystems.name = 'RedHat'")
```
jeremylenz, line 54: if there's ever a RHEL 6.10 this `<` will break

jlsherrill: Another option would be to look for hosts that have a subscription manager installed that is older than what works.  That would be more accurate IMO

app/lib/actions/katello/repository/update.rb
jlsherrill: +1 and it should spawn off another task, not run within this one.  We do this for example here:  https://github.com/Katello/katello/blob/master/app/lib/actions/katello/repository/update.rb#L62

app/lib/actions/katello/repository/update.rb
jeremylenz: Yeah that would be ideal but how would I know the installed sub-man version for all my hosts?

jeremylenz:
Updated:
* add new actions `Actions::Katello::Consumer::RegenerateEntitlementCertificate(s)`
* move the code that actually calls Candlepin to new methods in `Katello::Candlepin::Consumer` and `Katello::Resources::Candlepin::Consumer`
* update `Actions::Katello::Repository::Update` to spawn the new action using `async_task`
* clean up the above action to avoid calling `plan_self` inside a conditional

jeremylenz:
Updated to run the action in batches of 10 hosts each (speeds up performance by 10x)

jeremylenz:
Closing after discussion with Rich Jerrido and @mccun934.  Support for older RHEL 6 clients is not a requirement; these users should upgrade at least to RHEL6.latest.

PR #9083, opened 2020-12-21
Fixes #31540 - delay CVV calcs until needed

jlsherrill:


jlsherrill:
@parthaa curious your thought on this solution.  The reproducer in question had 150 million repo<=>rpm associations.  Delaying the calculation till its needed seemed like the most simple solution.  the initial request will be slow, but any additional requests will be fast.

parthaa:
> @parthaa curious your thought on this solution. The reproducer in question had 150 million repo<=>rpm associations. Delaying the calculation till its needed seemed like the most simple solution. the initial request will be slow, but any additional requests will be fast.

1) Won't the rest call to fetch the count timeout if there are 150 mill repo-rpms association?
2) I'd ve expected -> https://github.com/Katello/katello/blob/master/app/lib/actions/katello/content_view/publish.rb#L125 and https://github.com/Katello/katello/blob/master/app/lib/actions/katello/content_view_version/incremental_update.rb#L267 to be commented out or removed ?

jlsherrill:
@parthaa on the reproducer there were ~4500 content view versions.  Each CVV only took ~5 seconds to calculate, but doing them all during the migration took ~4 hours.  On the api, it would only return ~20 at at a time, so I think we would be okay.  The only other solution (i can think of) to this would be to create an upgrade rake task that runs in the background.  This is simpler in design, but i'm open to going that route.

jlsherrill:
@parthaa and i intentionally left those in place, as its no problem to spend 5 seconds calculating that at CV publish time.  Just doing them all at once takes forever.

parthaa:
Ah this is purely for migration I guess. Yeh good with your proposed solution

parthaa (approved):
Simple migration test ran ok with this change
ACK

PR #9084, opened 2020-12-22
Fixes #31544 - properly check for content source on new hostgroup

jlsherrill:


jlsherrill:
see issue for reproducer info, added test to reproduce the problem

jturel (approved):
ACK

PR #9085, opened 2020-12-23
Fixes #31446 - Copy Content Views UI

sjha4:
The copy flow is functional with this PR.
Working on:
- [x] Tests


johnpmitsch:
Works well and code looks good, will take another look after the tests are added!

webpack/scenes/ContentViews/Table/tableDataGenerator.js
```diff
@@ -27,7 +27,7 @@ const buildRow = (contentView, openColumn) => {
     id, composite, name, environments, versions, last_published: lastPublished,
   } = contentView;
   const row = [
-    { title: <ContentViewIcon composite={composite ? true : undefined} /> },
+    { title: <ContentViewIcon composite={composite ? true : undefined} />, props: { id, name } },
```
johnpmitsch, line 30: @sjha4 is this still needed? Looks like the info is coming from the row data that is set up below.

sjha4: Right..This is not needed..

webpack/scenes/ContentViews/Copy/__tests__/copyContentView.test.js
```diff
@@ -0,0 +1,53 @@
...
+  fireEvent.change(getByLabelText('input_name'), { target: { value: 'cv copy' } });
+
+  getByLabelText('copy_content_view').click();
+  // Form closes it self on success
+  await patientlyWaitFor(() => {
+    expect(queryByText('Name')).not.toBeInTheDocument();
+  });
+
+  assertNockRequest(copyscope, done);
+});
```
johnpmitsch, line 53: One thing I just noticed - this test is pretty similar to the other one, I'm fine with this test being the only test and checking multiple things during the test (the api call is made and the form is closed), these longer tests make sense in functional testing IMO

sjha4: Updated..

johnpmitsch (approved):

PR #9087, opened 2020-12-23
Refs #31550 - Add Timer class to Katello::Logging

jeremylenz:
Katello debugging tools: Currently you can do
```rb
::Katello::Logging.time('message', {data: mydata} ) do
  # code that takes time
end
```
But you have to pass in any variables used by the block to the data object.  Also you can only start and stop the timer within that single block.

This PR adds a Timer class to `Katello::Logging`.

Usage:
```rb
mytimer = ::Katello::Logging::Timer.new('my_timer').start
  # code that takes time
mytimer.stop
```

It will produce logs that look like this:
```
2020-12-23T18:39:04 [I|app|] Timer my_timer starting at 2020-12-23 18:39:04 +0000
2020-12-23T18:39:22 [I|app|] Timer my_timer stopping at 2020-12-23 18:39:22 +0000: 18.39 sec
```

Or, find the timer by key.  Useful for when you aren't in the same method / context:
```rb
def method_a
  ::Katello::Logging::Timer.new('my_timer').start
  # code that takes time
end

def method_b
  # more code that takes time
  ::Katello::Logging::Timer.find_by_key('my_timer').stop
end
```

Other benefits:
* `::Katello::Logging::Timer.new('my_timer')` and `::Katello::Logging::Timer.new('my_timer').start` will both return the timer object. 
* Unlimited number of concurrent timers
* `stop` method both logs and returns the log message (also updated the existing `time` class method to do the same)

### DIFF
--- a/app/lib/katello/logging.rb
+++ b/app/lib/katello/logging.rb
@@ -9,32 +9,37 @@ module Katello
       data_string = data.map { |k, v| "#{k}=#{v}" }.join(' ')
 
       logger.send(level, "#{message} #{data_string}")
-      "#{message} #{data_string}"
     end
 
     class Timer
       def initialize(key = "default")
         @key = key
         @start_time = Time.now
-        @@timers ||= {}
-        @@timers[key] = self
+        Thread.current[:timers] ||= {}
+        Thread.current[:timers][key] = self
       end
 
       def start
+        Rails.logger.info "Timer #{@key} already started; resetting start time" if @start_time
         Rails.logger.info "Timer #{@key} starting at #{Time.now}"
         @start_time = Time.now
         self
       end
 
       def stop
+        fail ::StandardError, "Timer #{@key} is not started" unless @start_time
         duration = (Time.now - @start_time).truncate(2)
         @start_time = nil
         Rails.logger.info "Timer #{@key} stopping at #{Time.now}: #{duration} sec"
-        "Timer #{@key} stopping at #{Time.now}: #{duration} sec"
       end
 
       def self.find_by_key(key)
-        @@timers[key]
+        if Thread.current&.[](:timers)&.[](key)
+          Thread.current[:timers][key]
+        else
+          Rails.logger.warn "Timer #{key} not found on current thread; creating a new timer"
+          self.new(key)
+        end
       end
     end
   end
--- a/app/lib/katello/logging.rb
+++ b/app/lib/katello/logging.rb
@@ -9,6 +9,33 @@ module Katello
       data_string = data.map { |k, v| "#{k}=#{v}" }.join(' ')
 
       logger.send(level, "#{message} #{data_string}")
+      "#{message} #{data_string}"
+    end
+
+    class Timer
+      def initialize(key = "default")
+        @key = key
+        @start_time = Time.now
+        @@timers ||= {}
+        @@timers[key] = self
+      end
+
+      def start
+        Rails.logger.info "Timer #{@key} starting at #{Time.now}"
+        @start_time = Time.now
+        self
+      end
+
+      def stop
+        duration = (Time.now - @start_time).truncate(2)
+        @start_time = nil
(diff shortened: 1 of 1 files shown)


jlsherrill:
Out of curiosity why is this Refs and not Fixes?

app/lib/katello/logging.rb
```diff
@@ -9,6 +9,33 @@ def self.time(message, data: {}, logger: Rails.logger, level: :info)
...
       logger.send(level, "#{message} #{data_string}")
+      "#{message} #{data_string}"
+    end
+
+    class Timer
+      def initialize(key = "default")
+        @key = key
+        @start_time = Time.now
+        @@timers ||= {}
+        @@timers[key] = self
```
jlsherrill, line 20: I'm not sure how thread safe this is given that its using class variables.  I wonder if we should use Thread local varaibles?  (Thread.current[:key]) ?

app/lib/katello/logging.rb
```diff
@@ -9,6 +9,33 @@ def self.time(message, data: {}, logger: Rails.logger, level: :info)
...
+
+      def stop
+        duration = (Time.now - @start_time).truncate(2)
+        @start_time = nil
+        Rails.logger.info "Timer #{@key} stopping at #{Time.now}: #{duration} sec"
+        "Timer #{@key} stopping at #{Time.now}: #{duration} sec"
+      end
+
+      def self.find_by_key(key)
+        @@timers[key]
```
jlsherrill, line 37: since this depends on the local process that is running, imagine setting some timer in the plan phase of a dynflow action and then stopping it in the finalize (this wouldn't work at all), but it might be better to handle that gracefully?  I.e. Return a new timer instead of nil if one isn't found with a given key (maybe with a warning in that timer's message that the data may not be accurate)?

jeremylenz:
> Out of curiosity why is this Refs and not Fixes?

I think I meant the redmine to be a Feature and not a Bug, and you can't "fix" a feature.  I will change the redmine if that's ok?

jeremylenz:
@jlsherrill Added usage of `Thread.current` and some other misc. cleanups & error handling.

jturel:
Where's this gonna be used?

jeremylenz:
@jturel It's used primarily for debugging Katello, as is the existing file.  (this is what I was using to time my Katello actions when I was working on https://github.com/Katello/katello/pull/9082)

app/lib/katello/logging.rb
```diff
@@ -9,6 +9,40 @@ def self.time(message, data: {}, logger: Rails.logger, level: :info)
       data_string = data.map { |k, v| "#{k}=#{v}" }.join(' ')
 
       logger.send(level, "#{message} #{data_string}")
+      "#{message} #{data_string}"
```
jturel, line 12: Why return this string?

jturel:
> @jturel It's used primarily for debugging Katello, as is the existing file. (this is what I was using to time my Katello actions when I was working on #9082)

Ok. If you haven't already you may wanna give this a shout in the larger standup so others are aware of the utility, and that way it'll get some use.

app/lib/katello/logging.rb
jeremylenz: @jturel It's a feature! :smile: 

When I was using it in the console, I would have to go searching in the log file to find the message.  Returning it guarantees that it's displayed on the screen.  (I now see that the log is also displayed.. not sure if I just missed that before)

jturel: Since the log message is displayed in the console if you're using this method there, I'd prefer to not return this string as that's not the intent of the method.

jeremylenz: @jturel Ok, removed

jeremylenz:
updated to placate Rubocop

jlsherrill (approved):
ACK from me, anything else @jturel ?

jturel (approved):
LGTM!

PR #9089, opened 2021-01-04
Fixes #31105 - Truncate description of CV env in candlepin

sjha4:
To test: create a CV with a very long description and publish.
You'll see action :  Actions::Candlepin::Environment::Create fail with error: `description: size must be between 0 and 255`

With this change we truncate the Candlepin env description to 255 chars.


jturel (approved):
ACK! Works for me with very long descriptions and when there is no description. Thanks @sjha4 !

PR #9093, opened 2021-01-07
Fixes #31586 - use pulp2 if pulp3 does not support a type

jlsherrill:


jlsherrill:
The steps to reproduce might be a little weird:

1.  Shut down and disable all the pulpcore services
2. Set all content types to use pulp2 (true) in :use_pulp_2_for_content_type:
3. Manually delete the capabilities of the SmartProxy pulpcore feature in the rails console
4. Try to create a puppet repository

ekohl:
> Manually delete the capabilities of the SmartProxy pulpcore feature in the rails console

I think that if pulpcore is unavailable, the smart proxy doesn't expose any capabilities so refreshing features should be sufficient.

jlsherrill:
@ekohl the issue is that in this 'specialized' downstream deployment, pulp3 is installed but not running, so when the smart proxy is refreshed it does not get any of the capabilities (since the pulpcore api server is down)

ianballou (commented):
Works fine for me! There's just a test failure.

ekohl:
> @ekohl the issue is that in this 'specialized' downstream deployment, pulp3 is installed but not running, so when the smart proxy is refreshed it does not get any of the capabilities (since the pulpcore api server is down)

What I'm saying is that I think you can also clear capabilities by stopping pulpcore* and then refreshing features. That would be the same thing as going into the console and clearing them. So the reproduction steps could be easier. Maybe it was also for my own understanding because we can see that in real deployments (pulpcore can be down when a user refreshes features). It's always good to understand a system.

I wasn't commenting on the exact PR. That in itself looks like it makes sense to me.

ianballou:
I started seeing `A smart proxy seems to have been refreshed without pulpcore being running.  You may want to ` when creating yum repositories following the same set up steps above.

jlsherrill:
i think i've got it correct now!  Also included a small fix for the switchover task

ianballou (approved):
Works for me! This also fixes the content switchover for 3.17.

PR #9095, opened 2021-01-07
Fixes #31382 - Hostgroup content facet migration fix

parthaa:
Added unscoped when checking for anonymous_admin.

### DIFF
--- a/db/migrate/20150930183738_migrate_content_hosts.rb
+++ b/db/migrate/20150930183738_migrate_content_hosts.rb
@@ -329,7 +329,7 @@ class MigrateContentHosts < ActiveRecord::Migration[4.2]
   # rubocop:disable Metrics/MethodLength
   # rubocop:disable Metrics/AbcSize
   def up
-    if User.where(:login => User::ANONYMOUS_API_ADMIN).first.nil?
+    unless User.unscoped.where(:login => User::ANONYMOUS_API_ADMIN).exists?
       logger.warn("Foreman anonymous admin does not exist, skipping content host migration.")
       return
     end
--- a/db/migrate/20200514092553_move_katello_fields_from_hostgroups.katello.rb
+++ b/db/migrate/20200514092553_move_katello_fields_from_hostgroups.katello.rb
@@ -1,6 +1,6 @@
 class MoveKatelloFieldsFromHostgroups < ActiveRecord::Migration[6.0]
   def up
-    if User.where(login: User::ANONYMOUS_ADMIN).exists?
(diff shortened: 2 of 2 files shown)


db/migrate/20150930183738_migrate_content_hosts.rb
```diff
@@ -329,7 +329,7 @@ def update_hypervisor_consumer(hypervisor)
   # rubocop:disable Metrics/MethodLength
   # rubocop:disable Metrics/AbcSize
   def up
-    if User.where(:login => User::ANONYMOUS_API_ADMIN).first.nil?
+    if User.unscoped.where(:login => User::ANONYMOUS_API_ADMIN).first.nil?
```
ekohl, line 332: Not needed, but shouldn't this also be `exists?` like the other?
```suggestion
    if User.unscoped.where(:login => User::ANONYMOUS_API_ADMIN).exists?
```

db/migrate/20150930183738_migrate_content_hosts.rb
```diff
@@ -329,7 +329,7 @@ def update_hypervisor_consumer(hypervisor)
   # rubocop:disable Metrics/MethodLength
   # rubocop:disable Metrics/AbcSize
   def up
-    if User.where(:login => User::ANONYMOUS_API_ADMIN).first.nil?
+    if User.unscoped.where(:login => User::ANONYMOUS_API_ADMIN).exists?
```
parthaa, line 332:
```suggestion
    unless User.unscoped.where(:login => User::ANONYMOUS_API_ADMIN).exists?
```

chris1984:
@parthaa is there a way to reproduce this?

parthaa:
> @parthaa is there a way to reproduce this?

- I installed 3.16
- created a hostgroup that has a cv, content source,  a lifecycle env, an os and arch (no repo id, but dont think it matters)
- upgraded to 3.17 using the instructions here -> https://theforeman.org/plugins/katello/nightly/upgrade/index.html (I used 3.17 instead of nightly.)

chris1984 (approved):
ACK works good:

Without the patch after upgrade:

https://nimb.ws/VkyTLu

With patch after upgrade:

https://nimb.ws/IiQWmY

parthaa:
[test katello]

parthaa:
[test katello]

chris1984:
[test katello]

parthaa:
[test katello]

chris1984:
[test katello]

parthaa:
[test katello]

PR #9097, opened 2021-01-12
Fixes #31615 - Document arch parameter for repositories

jturel:
This was preventing repo create and update from taking arch as a parameter via hammer


chris1984 (approved):
ACK, able to see the param with hammer now.

jturel:
[test katello]

PR #9098, opened 2021-01-13
Refs #31586 - load common.rake for pulp3 content switchover

evgeni:
in b6532ce we added disable_dynflow to the switchover task, but didn't
load the common.rake that defines it.

### DIFF
--- a/lib/katello/tasks/pulp3_content_switchover.rake
+++ b/lib/katello/tasks/pulp3_content_switchover.rake
@@ -1,3 +1,5 @@
+load "#{Katello::Engine.root}/lib/katello/tasks/common.rake"
+
 require File.expand_path("../engine", File.dirname(__FILE__))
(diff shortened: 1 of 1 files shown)


lib/katello/tasks/pulp3_content_switchover.rake
```diff
@@ -1,3 +1,5 @@
+load "#{Katello::Engine.root}/lib/katello/tasks/common.rake"
+
 require File.expand_path("../engine", File.dirname(__FILE__))
```
ekohl, line 3: Unrelated, but I wonder why this isn't:
```suggestion
require File.expand_path("engine", File.dirname(__dir__))
```

jlsherrill (approved):

lib/katello/tasks/pulp3_content_switchover.rake
jlsherrill: i'm not sure this line is even needed, but lets go ahead and merge this

PR #9102, opened 2021-01-14
Fixes #31359 - do not delete repo version 0

jlsherrill:
if we try to delete a repo version of zero,
just ignore it.

### DIFF
--- a/app/services/katello/pulp3/repository.rb
+++ b/app/services/katello/pulp3/repository.rb
@@ -262,8 +262,12 @@ module Katello
         create_version(:base_version => from_repository.version_href)
       end
 
+      def version_zero?
+        repo.version_href.ends_with?('/versions/0/')
+      end
+
       def delete_version
-        ignore_404_exception { api.repository_versions_api.delete(repo.version_href) }
+        ignore_404_exception { api.repository_versions_api.delete(repo.version_href) } unless version_zero?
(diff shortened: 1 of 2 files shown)


jlsherrill:
see redmine for reproducer steps

jjeffers (approved):
Test this as per reproduction steps.
Works as described, a+++ would delete CV version again.

APJ

app/services/katello/pulp3/repository.rb
```diff
@@ -262,8 +262,12 @@ def copy_version(from_repository)
         create_version(:base_version => from_repository.version_href)
       end
 
+      def version_zero?
+        repo.version_href.ends_with?('/versions/0/')
```
jjeffers, line 266: I guess this works but... seems like should be a better way to identify the 0th version without parsing the path? (Pretty sure the pulp apis do not provide this.)

jjeffers (approved):

app/services/katello/pulp3/repository.rb
jlsherrill: We could fetch that version and check the "number" field, but i was trying to not add an additional call to pulp.

jlsherrill:
[test katello]

jlsherrill:
[test katello]

PR #9104, opened 2021-01-19
Fixes #31676 - simplify smart proxy sync checks

jlsherrill:


app/lib/actions/katello/capsule_content/sync.rb
```diff
@@ -30,18 +30,18 @@ def plan(smart_proxy, options = {})
 
           fail _("Action not allowed for the default smart proxy.") if smart_proxy.pulp_primary?
 
-          smart_proxy_helper = ::Katello::SmartProxyHelper.new(smart_proxy)
-          repositories = smart_proxy_helper.combined_repos_available_to_capsule(environment, content_view, repository)
-
-          smart_proxy.ping_pulp if repositories.any? { |repo| !smart_proxy.pulp3_support?(repo) }
```
jlsherrill, line 36: This was causing an error for me (was checking pulp2 when it shouldn't), and when i went to debug, it was working fine.  However this code isn't needed, as the Actions should ping pulp2 or 3 based on if they are planned or not

jjeffers (approved):
LGTM.

Tested a capsule sync as well with these changes (pulp3 yum content), everything looked fine.

PR #9105, opened 2021-01-19
[WIP] Fixes #31644 - check ping 503 retry

jjeffers:
This adds a simple check for 503 http responses and then retries for a given number of times before failing.

Tested this locally by modifying /etc/httpd/conf.d/05-foreman-ssl.conf:
```
 <Location "/pulp/api/v3">
    ...
    ProxyPass  http://localhost:8080/
    ProxyPassReverse http://localhost:8080/
   ...
  </Location>
```

At port 8080 a [simple web application](https://gist.github.com/jjeffers/8e3434ed674628ba5e6a23c65a4baf3f) returns a 503.

In some situations a pulp3 migration will fail because the environment might include pulp3 services that have started but are not yet serving requests. This change provides a little insurance in that case.

### DIFF
--- a/lib/katello/tasks/reimport.rake
+++ b/lib/katello/tasks/reimport.rake
@@ -4,11 +4,25 @@ namespace :katello do
   desc "Runs a katello ping and prints out the statuses of each service"
   task :check_ping => :environment do
     ::User.current = ::User.anonymous_admin
-    ping_results = Katello::Ping.ping
-    if ping_results[:status] != "ok"
-      pp ping_results
-      fail("Not all the services have been started. Check the status report above and try again.")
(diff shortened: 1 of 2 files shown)


jjeffers:
[test katello]

jturel:
I see this is triaged for 3.18.3 but hasn't moved in a while. Still relevant?

jjeffers:
> I see this is triaged for 3.18.3 but hasn't moved in a while. Still relevant?

@jlsherrill @ehelms was the fix for this merged elsewhere?

jlsherrill:
it was an installer/puppet module change. I think we can close this now !

PR #9107, opened 2021-01-20
Fixes #31270 - Fix applicability with modular vs non-modular rpms

ianballou:
This PR addresses an issue Katello/Pulp 3 applicability where newer non-modular RPMs are marked as applicable for older modular RPMs.  This is not possible since `dnf` locks down modular packages when a related module is enabled, which blocks the installation of non-modular RPMs with the same name as the locked modular ones.

To test:
1) Sync needed-errata (https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/), modules-rpms (https://partha.fedorapeople.org/test-repos/separated/modules-rpms/), and rpm-deps (https://partha.fedorapeople.org/test-repos/separated/rpm-deps/)
2) Register an `EL8` content host and subscribe it
3) `yum module enable walrus:0.71`
4) `dnf uploadprofile --force-upload` for good measure
5) Take a look at the host's applicability.  `walrus-5.21` should be marked as applicable
  - This is wrong because a non-modular RPM should not be applicable for a modular RPM when its module stream is enabled
6) Switch to my PR
7) Regenerate applicability again
8) `walrus-5.21` should disappear from the host's applicability

TODO:
- [x] Add some tests
- [ ] Can this be optimized?


jlsherrill (approved):
ACK  

I reproduced the problem and this pr solved it.  Performance seems on par with current levels

PR #9109, opened 2021-01-25
Fixes #31716 - Add support for linked tabs and subroutes

johnpmitsch:
This adds support to our usage of Patternfly 4 tabs for the following:
1) link directly to tabs
2) have the back button work when switching tabs
3) support subpages inside of tabs (clicking a row inside a table in a tab will open a details page in that tab)


johnpmitsch:
@amirfefer @jeremylenz This is what I mentioned today, it's influenced by Ansible's awx (Tower's upstream) implentation, which was suggested on the patternfly slack channel for working with react router and tabs.

johnpmitsch:
Going to `labs/content_views/5` will automatically go to the default versions tab: `/labs/content_views/5#versions`
Clicking on a new tab will update the url hash, for example `/labs/content_views/5#filters`
Then if you click on a row in the filters tab, it will update with a query param for that content id and render the detail page (which is a placeholder for now). `/labs/content_views/6#filters?subContentId=1`

sjha4:
Functionality testing:
Lands on /labs/content_views/2#versions
Able to navigate to /labs/content_views/2#details , /labs/content_views/2#repositories, /labs/content_views/2#filters, /labs/content_views/2#history tabs.
Able to refresh the page and land on the individual tabs based on route.
:heavy_check_mark:

johnpmitsch:
@sjha4 thanks! I also moved the logic to get the hash and the query params to helper function w/ some tests too.

webpack/components/RoutedTabs/RoutedTabs.js
```diff
@@ -0,0 +1,89 @@
...
+    <Tabs activeKey={getActiveTab()} onSelect={handleTabSelect}>
+      {tabs.map((tab) => {
+        const { key, title } = tab;
+
+        return (
+          <Tab
+            aria-label={title}
+            eventKey={key}
+            key={key}
+            link={`${baseUrl}#${key}`}
```
jeremylenz, line 62: If you pass `href` instead of `link` here, the tab will render as an `<a>` rather than a `<button>`.  Thoughts on changing this?

The differences are
* As it is now, I get `@@router/LOCATION_CHANGE` Redux actions when I click Back and Forward, but not when I click one of the tabs.
* If it's rendered as an `<a>`, I get `@@router/LOCATION_CHANGE` actions both on click and on browser navigation.
* `<a>` feels slightly more correct since clicking it changes the URL in the URL bar and navigates you to another place.  But I can also see an argument for the other side.
* The CSS styling would have to be overridden (rendering as `<a>` makes every tab underlined.)

More food for thought: https://css-tricks.com/use-button-element/#when-links-make-more-sense

webpack/components/RoutedTabs/RoutedTabs.js
```diff
@@ -0,0 +1,89 @@
...
+            title={<TabTitleText>{title}</TabTitleText>}
```
jeremylenz, line 63: You can put [`<TabTitleIcon>`s in tab titles too](https://patternfly-react.surge.sh/components/tabs#filled-with-icons).  May be nice to avoid swallowing that ability here.  I know we shouldn't build features before they're required, but on the other hand I don't like to unnecessarily disable the innate abilities of Patternfly with our wrappers.  (`...props` works great for this in general, but wouldn't work here because PF loves JSX in props rather than making their components composable. :roll_eyes: )

jeremylenz (commented):
Thanks @johnpmitsch 

Tested on new CV page and works great!  Tabs load direct from initial request and once you're on the page navigation works and is fast.

Love the way everything is put together.  Just a couple discussion points below.

webpack/components/RoutedTabs/RoutedTabs.js
johnpmitsch: I think its a good idea but I just keep running into issues with patternfly that way, here is what happens when I just add a `className` to the tab, without any change to the stylesheet:

hi

full change (remove the className prop makes it go back to normal with underlined links)
```diff
diff --git a/webpack/components/RoutedTabs/RoutedTabs.js b/webpack/components/RoutedTabs/RoutedTabs.js
index 9f73c52..93f66b7 100644
--- a/webpack/components/RoutedTabs/RoutedTabs.js
+++ b/webpack/components/RoutedTabs/RoutedTabs.js
@@ -1,9 +1,9 @@
 import React, { useEffect } from 'react';
 import { shape, string, number, element, arrayOf } from 'prop-types';
-import { Tab, Tabs, TabTitleText } from '@patternfly/react-core';
+import { Tab, Tabs, TabsComponent, TabTitleText } from '@patternfly/react-core';
 import { useHistory, useLocation } from 'react-router-dom';
-import './RoutedTabs.scss';
 import paramsFromHash from '../../utils/paramsFromHash';
+import './RoutedTabs.scss';
 
 /*
   Creates tabs that change url hash based on the tab selected.
@@ -50,16 +50,17 @@ const RoutedTabs = ({ tabs, baseUrl, defaultTabIndex }) => {
   };
 
   return (
-    <Tabs activeKey={getActiveTab()} onSelect={handleTabSelect}>
+    <Tabs activeKey={getActiveTab()} onSelect={handleTabSelect} component={TabsComponent.nav}>
       {tabs.map((tab) => {
         const { key, title } = tab;
 
         return (
           <Tab
             aria-label={title}
+            className="kplain-tab-anchors"
             eventKey={key}
             key={key}
-            link={`${baseUrl}#${key}`}
+            href={`${baseUrl}#${key}`}
             title={<TabTitleText>{title}</TabTitleText>}
           >
             <div className="tab-body-with-spacing">
```

webpack/components/RoutedTabs/RoutedTabs.js
johnpmitsch: added an optional `titleComponent` prop!

webpack/components/RoutedTabs/RoutedTabs.js
jeremylenz: In order for `className` to work it would have to be passed down to a native HTML element.  Looks like Patternfly doesn't pass down the `className` in this way for `<Tab>`.

jeremylenz: also I'd be curious to see what the console error is when you get the blank screen :)

jeremylenz: In addition to the edited comment above, another option would be just to change the SCSS file for this component to find the tabs without relying on class names, and style as we wish.

jeremylenz: all of that said, don't spend too much time on this.  I'm fine with keeping as is if it's going to take too long.

johnpmitsch: The pf4 docs say `className` should `additional classes added to the Tab`, but it feels like it's overriding their own. I think passing in any of the pf4 class names would be brittle, they could change outside of our control and we would have to keep in-sync

> all of that said, don't spend too much time on this. I'm fine with keeping as is if it's going to take too long.

I'm leaning this way, I tried the href's when I started and gave up then too :man_shrugging:

johnpmitsch:
[test katello]

jeremylenz (approved):
Title prop changes look good, thanks @johnpmitsch 

APJ

jeremylenz:
[test katello]

johnpmitsch:
[test katello]

johnpmitsch:
[test katello]

PR #9111, opened 2021-01-25
Fixes #31724 - fix cv errata filter pkg calculation

jlsherrill:


jturel:
This is triaged to 3.18.3 - is that still correct?

PR #9112, opened 2021-01-25
Fixes #31655 - UX changes for create and copy modals

sjha4:
Depends on https://github.com/theforeman/foreman-js/pull/253 to use Tile component.
Requirements:
Create Screen : https://marvelapp.com/prototype/1121a8de/screen/76054859

Copy Screen : https://marvelapp.com/prototype/1121a8de/screen/74756540

Create Form Updates:
Create

Copy Form Updates:
Copy

### DIFF
--- a/webpack/scenes/ContentViews/Copy/CopyContentViewModal.js
+++ b/webpack/scenes/ContentViews/Copy/CopyContentViewModal.js
@@ -5,17 +5,27 @@ import CopyContentViewForm from './CopyContentViewForm';
 
 const CopyContentViewModal = ({
   cvId, cvName, show, setIsOpen,
-}) => (
-  <Modal
-    title={`Copy content view ${cvName}`}
-    variant={ModalVariant.large}
-    isOpen={show}
-    width="50%"
-    onClose={() => { setIsOpen(false); }}
-    appendTo={document.body}
-  ><CopyContentViewForm cvId={cvId} setModalOpen={setIsOpen} />
-  </Modal>
-);
+}) => {
+  const description = (
+    <p>
+      This will create a copy of <b>{cvName}</b>, including details,
+      repositories, and filters. Generated data such
+      as history, tasks and versions will not be copied.
+    </p>
+  );
+  return (
+    <Modal
+      title="Copy content view"
+      variant={ModalVariant.small}
+      isOpen={show}
+      description={description}
+      onClose={() => {
+        setIsOpen(false);
+      }}
+      appendTo={document.body}
+    ><CopyContentViewForm cvId={cvId} setModalOpen={setIsOpen} />
+    </Modal>);
+};
 
 CopyContentViewModal.propTypes = {
   cvId: PropTypes.string,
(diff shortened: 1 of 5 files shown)


johnpmitsch:
@sjha4 
How do you feel about this change?
```diff
diff --git a/webpack/scenes/ContentViews/Create/CreateContentViewForm.js b/webpack/scenes/ContentViews/Create/CreateContentViewForm.js
index d52bfcc..3f9ea64 100644
--- a/webpack/scenes/ContentViews/Create/CreateContentViewForm.js
+++ b/webpack/scenes/ContentViews/Create/CreateContentViewForm.js
@@ -97,8 +97,8 @@ const CreateContentViewForm = ({ setModalOpen }) => {
         />
       </FormGroup>
       <FormGroup isInline fieldId="type" label="Type">
-        <Grid>
-          <GridItem span={5} rowSpan={3}>
+        <Grid hasGutter>
+          <GridItem span={6}>
             <Tile
               isStacked
               aria-label="component_tile"
@@ -108,11 +108,10 @@ const CreateContentViewForm = ({ setModalOpen }) => {
               onClick={() => { setComponent(true); setComposite(false); }}
               isSelected={component}
             >
-              Single CV consists of repositories
+              Single content view consisting of repositories
             </Tile>
           </GridItem>
-          <GridItem span={2} rowSpan={2} />
-          <GridItem span={5} rowSpan={3}>
+          <GridItem span={6}>
             <Tile
               isStacked
               aria-label="composite_tile"
```
It will then use gutters for spacing, also spelling out of `content views` will wrap the text and make the Tiles even:

here

cc @MariSvirik

johnpmitsch:
`content view` can be lowercase in `Composite content view` and `Component content view` in the Tile's title (@marisvirik correct me if I'm wrong)

webpack/scenes/ContentViews/Copy/CopyContentViewModal.js
```diff
@@ -5,17 +5,27 @@ import CopyContentViewForm from './CopyContentViewForm';
...
-  ><CopyContentViewForm cvId={cvId} setModalOpen={setIsOpen} />
-  </Modal>
-);
+}) => {
+  const description = (
+    <p>
+      This will create a copy of <b>{cvName}</b>, including details,
+      repositories, and filters.<br />Generated data such
+      as history, tasks and versions will not be copied.
+    </p>
```
johnpmitsch, line 14: The wrapping looks a little odd here as it will split into two sentences if the name is long enough:

here

You can use backticks and backslash to make a long string and split over lines, removing the `br` to let it wrap as needed. I think `<b>` can be included in the string and still render, but we can come up with something if that is not the case!

```js
`testing that \
this should be \
one long string`
-> "testing that this should be one long string"
```

johnpmitsch (commented):
Just a couple cosmetic suggestions but otherwise looks good :+1:

MariSvirik:
Looks reaaaally good @sjha4 @johnpmitsch  👍.
Use sentence also in checkboxes: 'Solve dependencies' and 'Import only'. And don't forget to use sentence case also for the text in the button: 'Create content view'

johnpmitsch (approved):
Looks great!

sjha4:
[test katello]

sjha4:
React test failures were related..Updated tests..

webpack/scenes/ContentViews/Copy/CopyContentViewModal.js
jeremylenz: Should be a `<strong>` instead, see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b :)

PR #9113, opened 2021-01-25
Fixes #31725 - properly use dynflow:client with check_ping in tasks

jlsherrill:


jjeffers (approved):
Applied this change to a Katello 3.18 production image after reproducing the error.

The change resolved the issue.

PR #9121, opened 2021-01-27
Fixes #31740 - katello_events and candlepin_events intermittently show FAIL/not started

jeremylenz:
This PR includes some chores and refactoring, and also attempts to fix ~~a bug~~ two bugs.

### chores

* move `event_daemon.rb` to `event_daemon/runner.rb`
* move `Katello::EventDaemon` to `Katello::EventDaemon::Runner`
* move `Katello::EventDaemon::Monitor` to its own file
* move all `Rails.cache` use to `Katello::EventDaemon::Runner`

### how we fix the actual problem

* Don't clear `Rails.cache` on `stop`.  thanks @jturel for inspiration

### how to test

1. In `app/models/katello/ping.rb`, add the following at the top of the `ping` method:
```rb
        if Katello::EventDaemon::Runner.pid && Process.pid != Katello::EventDaemon::Runner.pid
          Process.kill(3, Katello::EventDaemon::Runner.pid)
          sleep(10)
        end
```
2. From another tab, `curl localhost:3000/api/v2/ping` until it sleeps
3. If needed, ping while it's sleeping from a third tab
4. You should see it come back FAIL / not running
5. Now apply the patch and do it again


app/services/katello/event_daemon/runner.rb
```diff
@@ -0,0 +1,94 @@
+module Katello
```
jeremylenz, line 1: This is the file formerly known as `event_daemon.rb`

app/services/katello/event_daemon/monitor.rb
```diff
@@ -8,6 +8,7 @@ def initialize(service_classes)
       def start
         error = nil
         status = nil
+        check_services(error, status)
```
jeremylenz, line 11: added this because I was getting a FAIL on the first request

app/services/katello/event_daemon/runner.rb
```diff
@@ -40,18 +41,26 @@ def write_pid_file
...
 
         def start
+          Rails.logger.info "((((((((((((((NOT RUNNABLE))))))))))))))" unless runnable?
           return unless runnable?
           lockfile = File.open(lock_file, 'r')
           begin
             lockfile.flock(File::LOCK_EX)
-            return if started? # ensure it wasn't started while we waited for the lock
+            strtd = started?
+            Rails.logger.info("******************** ABORTING start; already started *************") if strtd
```
jeremylenz, line 62: ignore all my silly log messages; they will be gone eventually!

app/services/katello/event_daemon/runner.rb
```diff
@@ -87,7 +96,19 @@ def runnable?
...
         end
+
+        # writes status to cache if refresh is true or if there is no value cached
+        # otherwise, reads from cache
+        def cached_status(service_class:, status: nil, refresh: true)
+          cache_key = @cache_keys[service_class]
+          Rails.logger.debug "Writing status to cache #{cache_key}: #{status}" if refresh
+          Rails.cache.fetch(cache_key, force: refresh) do
+            status
+          end
```
jeremylenz, line 109: I went through several implementations and arrived at this one.  Originally this took a `cache_key` argument, but after I started removing the `STATUS_CACHE_KEY` constants from each service class I discovered that it's much simpler to just pass in the class here.  This is because `EventDaemon::Runner` is the only class that has access to the _names_ that the services were registered with, and the cache_key is derived from the name ( `"#{name.downcase}_status"`, i.e. `katello_events_status`.)

app/services/katello/event_daemon/monitor.rb
```diff
@@ -0,0 +1,52 @@
+module Katello
```
jturel, line 1: I vote to move runner and monitor to app/lib vs app/services (I did this too as part of my katello-agent pr) reason being that they are not "service classes" in the sense that they encapsulate logic is used by other areas of the app. Really this is part of an underlying subsystem of the app - therefore lib seems more appropriate! The distinction is a bit vague...

app/services/katello/event_daemon/monitor.rb
jturel: Was that problematic in some way?

app/models/katello/ping.rb
```diff
@@ -65,14 +70,20 @@ def event_daemon_status(status, result)
...
+          )
           event_daemon_status(status, result)
         end
       end
 
       def ping_candlepin_events(result)
         exception_watch(result) do
-          status = Katello::CandlepinEventListener.status(refresh: false)
+          status = Katello::EventDaemon::Runner.cached_status(
+            service_class: Katello::CandlepinEventListener,
```
jturel, line 84: I'd prefer to see the service name references vs the service class ie `:candlepin_events`

app/services/katello/event_daemon/monitor.rb
jeremylenz: Isn't it always problematic to get a FAIL on a hammer ping?  (For our friend the original bug reporter it seems to be.)

app/models/katello/ping.rb
jeremylenz: How will `Katello::Ping` know the name?  It seemed better to me to leave that to `EventDaemon::Runner` (see my other comment).  but thoughts welcome

app/services/katello/event_daemon/runner.rb
```diff
@@ -0,0 +1,115 @@
...
+        def runnable?
+          # avoid accessing the disk on each request
+          @cache.fetch('katello_event_daemon_runnable', expires_in: 1.minute) do
+            !started? && settings[:enabled] && !::Foreman.in_rake? && !Rails.env.test?
+          end
+        end
+
+        def register_service(name, klass)
+          @services[name] = klass
+          @cache_keys[klass] = "#{name.downcase}_status"
```
jturel, line 99: you may elect to change the structure of the service registry so we can store more data about the services now and in the future

current example:
```
{
 :candlepin_events => ::Katello::CandlepinEventListener
}
```

possibility:

```
{
  :candlepin_events => {
     class: Katello::CandlepinEventListener
     cache_key: "candlepin_events_status"
  }
}
```

app/services/katello/event_daemon/monitor.rb
jturel: You're saying hammer ping always fails on the first request without this addition?

app/models/katello/ping.rb
jturel: Just like you do here, but give it the service name vs the service class. Since you're now asking the Runner for the status it's better to ask by the 'friendly name' rather than the class which implements the service - that's now an implementation detail we can hide through the Runner

jeremylenz: but that will cause a lot of difficulty in Monitor and other places where we do `@service_classes.each` loops

jturel: I think it pretty much comes down to making [this change](https://github.com/Katello/katello/pull/9121#discussion_r566278311) and the rest should fall into place

app/services/katello/event_daemon/runner.rb
```diff
@@ -0,0 +1,115 @@
...
+        end
+
+        # writes status to cache if refresh is true or if there is no value cached
+        # otherwise, reads from cache
+        def cached_status(service_class:, status: nil, refresh: true)
```
jturel, line 104: Why does this method take status as a param? shouldn't it be calling .status on the service class directly in all cases?

app/services/katello/event_daemon/runner.rb
```diff
@@ -0,0 +1,115 @@
...
+
+        def stop
+          return unless pid == Process.pid
+          Rails.logger.info "<<<<<<<<<<<<<<< STOPPING RUNNER >>>>>>>>>>>>>>>>>"
+          @monitor_thread.kill
+          @cache.clear
+          @services.values.each do |service_class|
+            service_class.close
+            Rails.logger.info "Clearing cache #{@cache_keys[service_class]}"
+            Rails.cache.delete(@cache_keys[service_class])
```
jturel, line 50: How vital is it that we clear the cache here?

jeremylenz: I didn't really think about it, I was just moving the existing functionality over from the individual service classes.  But to answer the question, once the service is `close`d then our cached data is inaccurate anyway, so it makes sense to me to clear it.

app/services/katello/event_daemon/monitor.rb
```diff
@@ -18,13 +18,11 @@ def start
       end
 
       def check_services(error, status)
+        Katello::EventDaemon::Runner.cached_status(
+          refresh: true
```
jeremylenz, line 22: I changed `cached_status` so that `service_name` is optional as well.  If not provided it will just update all the statuses.

app/models/katello/ping.rb
jeremylenz: @jturel ok, updated!

app/services/katello/event_daemon/monitor.rb
jeremylenz: Well, I'm doing `curl localhost:3000/api/v2/ping` but yes.  It fails for the first few seconds.

jeremylenz: I think it's because `check_services` is the only place that does `service_class.run`.  Seems like that should happen when the daemon is started..

app/services/katello/event_daemon/runner.rb
jeremylenz: Instead of storing the cache key with each service class we decided to use a single cache for all services, so I left `@services` as-is for now and eliminated `@cache_keys`.  (just posting this for the record here)

app/services/katello/event_daemon/runner.rb
jeremylenz: :heavy_check_mark: Solution

app/services/katello/event_daemon/runner.rb
```diff
@@ -0,0 +1,113 @@
...
+
+        def service_statuses
+          result = {}
+          @services.each do |service_name, service_class|
+            result[service_name] = service_class.status
+          end
+          result
+        end
+
+        def cached_status(service_name = nil)
```
jturel, line 105: IMO this method could/should be renamed to `service_status`. The cache aspect is not relevant to this external API, although you and I both know where the status comes from!

that simplifies things significantly because this also removes the logic in the last line of the method which alters the return value (not a fan of that pattern personally):

```
def service_status(service_name)
  Rails.cache.read(STATUS_CACHE_KEY)&.dig(service_name)
end
```

This is unneeded now, but in the future if we wanted all services status, we just add another method and keep the two concepts separate:
```
def services_status
  Rails.cache.read(STATUS_CACHE_KEY)
end
```

app/services/katello/event_daemon/monitor.rb
```diff
@@ -0,0 +1,54 @@
...
+              service_class.status
+            rescue => error
+              Rails.logger.error("Error occurred while starting #{service_class}")
+              Rails.logger.error(error.message)
+              Rails.logger.error(error.backtrace.join("\n"))
+            ensure
+              error = nil
+            end
+          end
+          write_statuses_to_cache
```
jturel, line 41: This implementation means that we are calling .status on all the service classes an additional time. What about building the entire status during the execution of `check_statuses`?

for example:

```
def initialize(...)
   ...
  @service_statuses = {}
end
```

near the top of check_services:
```
  statuses[service_name] = service_class.status
````

and at the end

```
Rails.cache.write(KEY, @service_statuses)
```

File: app/services/katello/event_daemon/runner.rb (hunk 1 above)
Comment (jturel) [reply app/services/katello/event_daemon/runner.rb:1]: image

**Review by jturel — CHANGES_REQUESTED** [review]

File: app/services/katello/event_daemon/runner.rb (hunk 11 above)
Comment (jeremylenz) [reply app/services/katello/event_daemon/runner.rb]: The future `services_status` will go great right next to `service_statuses` :laughing:

**Comment by jeremylenz** [comment]
@jturel updated :)

**Comment by jeremylenz** [comment]
@jturel squashed, shuffled & scrubbed!

File: app/lib/katello/event_daemon/monitor.rb [hunk 13]
```diff
@@ -0,0 +1,50 @@
...
+        loop do
+          Rails.application.executor.wrap do
+            check_services(error, status)
+          end
+          sleep 15
+        end
+      end
+
+      def check_services(error, status)
+        @service_statuses = {}
```
jturel, line 21: I think you can just use the one you created in the initializer here

app/lib/katello/event_daemon/monitor.rb
```diff
@@ -0,0 +1,50 @@
...
+              @service_statuses[service_name] = service_class.status
+            rescue => error
+              Rails.logger.error("Error occurred while starting #{service_class}")
+              Rails.logger.error(error.message)
+              Rails.logger.error(error.backtrace.join("\n"))
+            ensure
+              error = nil
+            end
+          end
+          Rails.cache.write(
```
jturel, line 42: Let's add a 1 minute expiry to this cache key. That way we can be more sure of the accuracy in case, somehow, the monitor itself rails

jturel:
It appears there is a bug in CandlepinEventListener when close is called. It could be that `running?` is incorrectly returning false? because I never see 'Closing candlepin event listener' logged.

https://github.com/Katello/katello/blob/b3ec5d421be54305bfdc618e4849d4e4ffaa689a/app/services/katello/candlepin_event_listener.rb#L21

I was comparing to main branch, and in the reproducer path once the daemon process is killed, the CandlepinEventListener status still shows running! I think it's worth trying to fix this now since we're looking at the code.

test/lib/event_daemon/runner_test.rb
```diff
@@ -0,0 +1,42 @@
+require 'katello_test_helper'
+
+module Katello
```
jturel, line 3: These tests should be in the EventDaemon namespace too

app/lib/katello/event_daemon/monitor.rb
jeremylenz: Seems that since we're now using `Rails.cache.read` instead of `fetch`, it will just return `nil` if the expiry time has passed.  Unlike `fetch`, it won't refresh anything.  So I'm not sure of the benefit of expiring it, unless we were to switch back to using `fetch`.

jeremylenz: > in case, somehow, the monitor itself rails

perhaps you meant ... goes off the rails? :smile:

jturel: The underlying question here is "how long do we want this cache key to live?" Options include:

- briefly (seconds)
- short amount of time (minutes)
- a long time (hours or days)
- indefinitely (until the cache is cleared)

My suggestion was to prevent us from running into a situation where the Monitor is not running (who knows why), but the cached status (living until cache clear in the current implementation)  gives the impression that everything is OK. By giving it a reasonable expiry (1 minute) we can be sure that within the last minute that Monitor has set the status that we are looking at. I see it as a safety mechanism

jturel: Hmm but maybe this is a bad idea cause then the ping API will show Not Running in between clearing the cache and the next tick of the monitor thread :(

jeremylenz: Another option would be to ditch the 15-second loop altogether and return to using `Rails.cache.fetch` with expiration.  That way when you Katello::Ping, it would either give you the (valid) cached data or update it on the spot.  But I feel like that's not really in scope of this PR.

app/lib/katello/event_daemon/monitor.rb
```diff
@@ -0,0 +1,47 @@
...
+          if error || !@service_statuses[service_name]&.dig(:running)
+            begin
+              service_class.close
+              service_class.run
+              @service_statuses[service_name] = service_class.status
+            rescue => error
+              Rails.logger.error("Error occurred while starting #{service_class}")
+              Rails.logger.error(error.message)
+              Rails.logger.error(error.backtrace.join("\n"))
+            ensure
```
jturel, line 35: I think this ensure can go away since error will not be shared across ticks of the check_services

app/lib/katello/event_daemon/monitor.rb
```diff
@@ -0,0 +1,47 @@
...
+              error = nil
+            end
+          end
+          Rails.cache.write(
```
jturel, line 39: It's probably safe + preferable to write the cache just once per loop of `check_services` (this is writing on for each service which I realize is how it was before, essentially)

jturel (changes requested):

jeremylenz:
@jturel Updated addressing comments; also removed `@thread&.kill` as per IRC discussion.

test/lib/event_daemon/monitor_test.rb
```diff
@@ -39,7 +39,33 @@ def test_monitor_not_running
 
         MockService.expects(:run)
 
-        monitor.check_services(nil, nil)
+        monitor.check_services
+      end
+
+      def test_monitor_starting
+        Rails.cache.expects(:write).with(
```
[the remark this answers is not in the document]
jeremylenz: I thought making a test double of `write` would be more correct than reading the "real" Rails cache, since this test isn't trying to test `Rails.cache`.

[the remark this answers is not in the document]
jeremylenz: For some reason I have a rule in my head "mock whatever you're not testing" so that's how I was thinking.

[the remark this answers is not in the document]
jturel: Works for me

jturel (approved):
It works great! Thanks @jeremylenz 

APJ

jturel:
I've seen the error in the tests before and I fixed it like this: https://github.com/Katello/katello/blob/8fc20655bf6e349e93ec6d9475488a8453f0cc6e/test/lib/messaging/stomp_connection_test.rb#L3

PR #9124, opened 2021-01-28
Fixes #29592 - properly clear upstream auth

jlsherrill:
on repo details page


jlsherrill:
[test katello]

jjeffers (approved):
Set the upstream authentication details, checked that the display and the underlying root repository in the console reflected the values. Applied this PR and observed the "x" icon working to clear the upstream authentication values back to nil, both in the UI and in the stored root repository fields.

PR #9125, opened 2021-01-28
Fixes #31748 - Applied Errata report not working

ranjan:


jlsherrill:
[test katello]

ares:
I don't think this is necessarily the right fix. The filter in this case is empty array so later it results in the empty list. We should use the filter only if it's not nil I think. cc @ezr-ondrej who wrote the original patch.

app/lib/katello/concerns/base_template_scope_extensions.rb
```diff
@@ -117,7 +117,7 @@ def load_errata_applications(filter_errata_type: nil, include_last_reboot: 'yes'
                               select: 'foreman_tasks_tasks.*,template_invocations.id AS template_invocation_id',
                               search: search
         )
-        only_host_ids = ::Host.search_for(host_filter).pluck(:id) if host_filter
+        only_host_ids = ::Host.search_for(host_filter).pluck(:id)
```
ezr-ondrej, line 120: We should fix this instead, without the filter we fetch all the hosts, that might be quite a lot.

ezr-ondrej: something along the lines of

```ruby
# line #132
seen_host_ids &= only_host_ids if only_host_ids
```
```ruby
# line #138
next unless only_host_ids && only_host_ids.include?(task.input['host']['id'].to_i)
```

ranjan: Thanks @ezr-ondrej , I have update the PR

app/lib/katello/concerns/base_template_scope_extensions.rb
```diff
@@ -129,13 +129,14 @@ def load_errata_applications(filter_errata_type: nil, include_last_reboot: 'yes'
...
+          seen_host_ids &= only_host_ids if only_host_ids
 
           # preload errata in one query for this batch
           preloaded_errata = Katello::Erratum.where(:errata_id => seen_errata_ids).pluck(:errata_id, :errata_type)
-          preloaded_hosts = ::Host.where(:id => seen_host_ids & only_host_ids).includes(:reported_data)
+          preloaded_hosts = ::Host.where(:id => seen_host_ids).includes(:reported_data)
 
           batch.each do |task|
-            next unless only_host_ids.include?(task.input['host']['id'].to_i)
+            next unless seen_host_ids && seen_host_ids.include?(task.input['host']['id'].to_i)
```
ezr-ondrej, line 139: `seen_host_ids` is always array, so we wouldn't need `seen_host_ids &&` part, but I think, we should not change it to `seen_host_ids` here so it should be:
```suggestion
            next unless only_host_ids && only_host_ids.include?(task.input['host']['id'].to_i)
```
1. `seen_host_ids` will be empty array if `include_last_reboot != 'yes'` in such case we would end up with empty report for every `include_last_reboot == 'no'`
2. `seen_host_ids` doesn't need to be complete set even for `include_last_reboot == 'yes'`. Some hosts (hosts without perms, deleted hosts, etc...) may be missing and we still want to include their errata application in the report.

ezr-ondrej (changes requested):
Just one issue, otherwise I believe we are good to go :+1:

app/lib/katello/concerns/base_template_scope_extensions.rb
ezr-ondrej: Shame on me. Bad conditional.
```suggestion
            next unless only_host_ids.nil? || only_host_ids.include?(task.input['host']['id'].to_i)
```

ranjan: @ezr-ondrej++ PR updated.

ezr-ondrej (approved):
Thanks @ranjan! Works for me :+1:

ezr-ondrej:
I've fixed the redmine ticket's project, so reruning the redmine issue check should make it green, but I don't have permissions to re-run it :)
@jlsherrill could you please take a look? This should get backported.

ezr-ondrej:
Or maybe @jturel? :P

ares (approved):

jlsherrill:
This isn't working in master:
This line:
https://github.com/theforeman/foreman/blob/develop/app/views/unattended/report_templates/host_-_applied_errata.erb#L62
is throwing this error:
undefined method '#empty?' for NilClass::Jail (NilClass)).   

Once that is fixed, this seems to work fine.

jlsherrill:
would it be possible to write a test that covers this functionality ?

ezr-ondrej:
Fixing the nil values: https://github.com/theforeman/foreman/pull/8350

ezr-ondrej:
@jlsherrill the `#empty?` is now `#blank?` so we should not crash anymore.
I agree test would be nice, I can try so in a followup PR.

jlsherrill (approved):

PR #9127, opened 2021-01-29
Fixes #30939 - remove deprecations for 4.0 release

jjeffers:
These are identified as definitely to be removed for katello 4.0. There will be a separate PR for any additional deprecations to be removed that may not make the first release candidate.


app/controllers/katello/api/v2/upstream_subscriptions_controller.rb
```diff
@@ -2,7 +2,6 @@ module Katello
   class Api::V2::UpstreamSubscriptionsController < Api::V2::ApiController
     before_action :find_organization
     before_action :check_upstream_connection
-    before_action :deprecated, only: [:simple_content_access_eligible, :enable_simple_content_access, :disable_simple_content_access]
```
jturel, line 5: We just added these deprecations so unfortunately it's too soon to remove them!

jjeffers: We should adjust the annotation in the deprecation method `it will be removed in Katello 4.0`.

jjeffers:
[test katello]

app/controllers/katello/api/rhsm/candlepin_dynflow_proxy_controller.rb
```diff
@@ -31,19 +29,6 @@ def upload_profiles
...
-        param :name, String, :required => true
-        param :architecture, String, :required => true
-        param :version, String, :required => true
-      end
-    end
-    param :id, String, :desc => N_("UUID of the system"), :required => true
-    def deb_package_profile
-      upload_profiles
-    end
-
```
jlsherrill, line 46: @jturel since you already reviewed this and it looks like you messed around with this deb_package_profile api previously, can you give some backstory on what it was used for and why its going away?

app/controllers/katello/api/v2/repositories_controller.rb
```diff
@@ -45,7 +44,7 @@ def custom_index_relation(collection)
       param :checksum_type, String, :desc => N_("Checksum of the repository, currently 'sha1' & 'sha256' are supported")
       param :docker_upstream_name, String, :desc => N_("Name of the upstream docker repository")
       param :docker_tags_whitelist, Array, :desc => N_("Comma-separated list of tags to sync for Container Image repository")
-      param :download_policy, ["immediate", "on_demand", "background"], :desc => N_("download policy for yum repos (either 'immediate', 'on_demand', or 'background (deprecated)')")
+      param :download_policy, ["immediate", "on_demand", "background"], :desc => N_("download policy for yum repos (either 'immediate' or 'on_demand')")
```
jlsherrill, line 47: I think we need to remove 'background' here from the list of acceptable values (and add a migration to change the db from 'background' => 'immediate' for all RootRepository objects that are set to background

app/controllers/katello/api/rhsm/candlepin_dynflow_proxy_controller.rb
jturel: ATIX added this in support of deb package actions before subscription-manager could upload the deb package profile (they had a separate script to upload the package data). Subman has been able to do that for a while now, so I think it's safe to remove this API.

jlsherrill (approved):
LGTM!

jturel (approved):
ACK. I recommend squashing so that you've got 1 commit for 30939 and another for 31617 (git rebase -i should do the trick)

jjeffers:
> ACK. I recommend squashing so that you've got 1 commit for 30939 and another for 31617 (git rebase -i should do the trick)

Not sure why 31617 is in there.

jjeffers:
[test katello]

PR #9131, opened 2021-02-01
Fixes #31770 - fix start_date and type in CH subscriptions page

snagoor:


jturel:
@snagoor I think the root cause of this issue is that a number of files were inadvertently deleted from the code some time ago now. Those files are:

https://github.com/Katello/katello/blob/KATELLO-3.13/engines/bastion_katello/app/assets/javascripts/bastion_katello/subscriptions/subscription-start-date.directive.js
https://github.com/Katello/katello/blob/KATELLO-3.13/engines/bastion_katello/app/assets/javascripts/bastion_katello/subscriptions/subscription-type.directive.js
https://github.com/Katello/katello/blob/KATELLO-3.13/engines/bastion_katello/app/assets/javascripts/bastion_katello/subscriptions/views/subscription-start-date.html
https://github.com/Katello/katello/blob/KATELLO-3.13/engines/bastion_katello/app/assets/javascripts/bastion_katello/subscriptions/views/subscription-type.html

If you'll add these files back to the code I'm happy to test!

snagoor:
@jturel Thank you for the review. I might be wrong, but removal of this code was intentional as a part of migrating to react https://github.com/Katello/katello/pull/8314

Also, current change was enough for me to show the missing `Starts` and `Type` columns in the Subscriptions page of the Content Host.

**Before**

before

**After**

after

jturel:
> @jturel Thank you for the review. I might be wrong, but removal of this code was intentional as a part of migrating to react #8314
> 
> Also, current change was enough for me to show the missing `Starts` and `Type` columns in the Subscriptions page of the Content Host.
> 
> **Before**
> 
> before
> 
> **After**
> 
> after

@snagoor you're correct that they were removed as part of the migration to react, but I don't think the developer realized that those files were still in use. If you apply them back to the code (if you're using our dev environment) you'll see that they have an effect.  If you look into those files you can see they do a few other things with respect to how the data is presented which is different than what your change currently does. That's why I think they should be restored!

chris1984:
@snagoor 

I tested this pr with those files back in the Katello environment and I am seeing the start date and the guests of:

https://1drv.ms/u/s!AmcYeL73Lgj6hlqLVwZ03TGJvWnP?e=ojwrch

snagoor:
Apologies @jturel I couldn't build a dev environment myself. Facing multiple issues as described in https://github.com/theforeman/forklift/issues/1288

I just added those 4 files back to this patch and removed the changes I made. 

@chris1984 Thank you for testing this and indeed we are fixing one more bug with this patch https://bugzilla.redhat.com/show_bug.cgi?id=1843710

jturel:
Thanks @snagoor  !

Two things:

1. the .html files should be in the `views` folder ie `engines/bastion_katello/app/assets/javascripts/bastion_katello/subscriptions/views/`
2. the date formatting used in `engines/bastion_katello/app/assets/javascripts/bastion_katello/subscriptions/views/subscription-start-date.html` is no longer correct and can be changed to: `<date date="subscription.start_date" />{{ checkFutureDate(subscription.start_date) }}` which I think is what you had in the original version :)

snagoor:
@jturel addressed both points you have mentioned. `me--` missed on the HTML file location.

Finally I was able to setup a dev environment to test this, with few hiccups https://github.com/theforeman/forklift/issues/1288#issuecomment-776605202

jturel:
> @jturel addressed both points you have mentioned. `me--` missed on the HTML file location.
> 
> Finally I was able to setup a dev environment to test this, with few hiccups [theforeman/forklift#1288 (comment)](https://github.com/theforeman/forklift/issues/1288#issuecomment-776605202)

Thank you @snagoor! It works perfectly. Sorry about your trouble with the dev env, but looks like it should already be fixed for next time.

jturel (approved):
ACK

PR #9133, opened 2021-02-01
Fixes #31653 - New UI: Add/Remove repositories to CV

sjha4:
Add/remove repositories to content views
To-do:
- [x] Bulk
- [x] Actions in kebab
- [x] Tests

### DIFF
--- a/webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js
+++ b/webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js
@@ -1,6 +1,16 @@
 import React, { useState, useEffect } from 'react';
 import { useSelector, shallowEqual, useDispatch } from 'react-redux';
-import { Bullseye, Split, SplitItem } from '@patternfly/react-core';
+import {
+  Bullseye,
+  Split,
+  SplitItem,
+  Button,
+  ActionList,
+  ActionListItem,
+  Dropdown,
+  DropdownItem,
+  KebabToggle,
+} from '@patternfly/react-core';
 import { TableVariant, fitContent } from '@patternfly/react-table';
 import { STATUS } from 'foremanReact/constants';
 import { translate as __ } from 'foremanReact/common/I18n';
@@ -9,13 +19,14 @@ import PropTypes from 'prop-types';
 
 import TableWrapper from '../../../../components/Table/TableWrapper';
 import onSelect from '../../../../components/Table/helpers';
-import { getContentViewRepositories, getRepositoryTypes } from '../ContentViewDetailActions';
+import { getContentViewRepositories, getRepositoryTypes, updateContentView } from '../ContentViewDetailActions';
 import {
   selectCVRepos,
   selectCVReposStatus,
   selectCVReposError,
   selectRepoTypes,
   selectRepoTypesStatus,
+  selectCVDetails,
 } from '../ContentViewDetailSelectors';
 import { ADDED, NOT_ADDED, ALL_STATUSES } from '../../ContentViewsConstants';
 import ContentCounts from './ContentCounts';
@@ -41,6 +52,7 @@ const ContentViewRepositories = ({ cvId }) => {
   const error = useSelector(state => selectCVReposError(state, cvId), shallowEqual);
   const repoTypesResponse = useSelector(state => selectRepoTypes(state), shallowEqual);
   const repoTypesStatus = useSelector(state => selectRepoTypesStatus(state), shallowEqual);
+  const details = useSelector(state => selectCVDetails(state, cvId), shallowEqual);
 
   const [rows, setRows] = useState([]);
   const [metadata, setMetadata] = useState({});
@@ -49,6 +61,8 @@ const ContentViewRepositories = ({ cvId }) => {
   const [statusSelected, setStatusSelected] = useState(ALL_STATUSES);
   // repoTypes object format: [displayed_value]: API_value
   const [repoTypes, setRepoTypes] = useState({});
+  const [bulkActionOpen, setBulkActionOpen] = useState(false);
+  const [bulkActionEnabled, setBulkActionEnabled] = useState(false);
 
   const columnHeaders = [
     { title: __('Type'), transforms: [fitContent] },
@@ -84,17 +98,9 @@ const ContentViewRepositories = ({ cvId }) => {
           title: <RepoAddedStatus added={addedToCV || statusSelected === ADDED} />,
         },
       ];
-
-      newRows.push({ cells });
+      newRows.push({ repoId: id, cells });
     });
     return newRows;
-  };
-
-  const getCVReposWithOptions = (params = {}) => {
-    const allParams = { ...params };
-    if (typeSelected !== 'All repositories') allParams.content_type = repoTypes[typeSelected];
-
-    return getContentViewRepositories(cvId, allParams, statusSelected);
   };
 
   useEffect(() => {
@@ -111,6 +117,11 @@ const ContentViewRepositories = ({ cvId }) => {
     dispatch(getRepositoryTypes());
   }, []);
 
+  useEffect(() => {
+    const rowsAreSelected = rows.some(row => row.selected);
+    setBulkActionEnabled(rowsAreSelected);
+  }, [rows]);
+
   // Get repo type filter selections dynamically from the API
   useEffect(() => {
     if (repoTypesStatus === STATUS.RESOLVED && repoTypesResponse) {
@@ -126,12 +137,76 @@ const ContentViewRepositories = ({ cvId }) => {
     }
   }, [JSON.stringify(repoTypesResponse), repoTypesStatus]);
 
+  const toggleBulkAction = () => {
+    setBulkActionOpen(!bulkActionOpen);
+  };
+
+  const onAdd = (repos) => {
+    const { repository_ids: repositoryIds = [] } = details;
+    dispatch(updateContentView(cvId, { repository_ids: repositoryIds.concat(repos) }));
+  };
+
+  const onRemove = (repos) => {
+    const reposToDelete = [].concat(repos);
+    const { repository_ids: repositoryIds = [] } = details;
+    const deletedRepos = repositoryIds.filter(x => !reposToDelete.includes(x));
+    dispatch(updateContentView(cvId, { repository_ids: deletedRepos }));
+  };
+
+  const addBulk = () => {
+    const reposToAdd = [];
+    rows.forEach(row => row.selected && reposToAdd.push(row.repoId));
+    onAdd(reposToAdd);
+  };
+
+  const removeBulk = () => {
+    const reposToDelete = [];
+    rows.forEach(row => row.selected && reposToDelete.push(row.repoId));
+    onRemove(reposToDelete);
+  };
+
+  const actionResolver = (rowData, { _rowIndex }) => {
+    if (rowData.parent || rowData.compoundParent || rowData.noactions) return null;
+    const { repository_ids: repositoryIds } = details;
+    return [
+      {
+        title: 'Add',
+        isDisabled: repositoryIds && repositoryIds.includes(rowData.repoId),
+        onClick: (_event, rowId, rowInfo) => {
+          onAdd(rowInfo.repoId);
+        },
+      },
+      {
+        title: 'Remove',
+        isDisabled: repositoryIds && !repositoryIds.includes(rowData.repoId),
+        onClick: (_event, rowId, rowInfo) => {
+          onRemove(rowInfo.repoId);
+        },
+      },
+    ];
+  };
+
+  const getCVReposWithOptions = (params = {}) => {
+    const allParams = { ...params };
+    if (typeSelected !== 'All repositories') allParams.content_type = repoTypes[typeSelected];
+
+    return getContentViewRepositories(cvId, allParams, statusSelected);
+  };
+
   const emptyContentTitle = __("You currently don't have any repositories to add to this content view.");
   const emptyContentBody = __('Please add some repositories.'); // needs link
   const emptySearchTitle = __('No matching repositories found');
   const emptySearchBody = __('Try changing your search settings.');
   const activeFilters = (typeSelected && typeSelected !== allRepositories) ||
     (statusSelected && statusSelected !== ALL_STATUSES);
+  const dropdownItems = [
+    <DropdownItem aria-label="bulk_add" key="bulk_add" isDisabled={!bulkActionEnabled} component="button" onClick={addBulk}>
+      Add
+    </DropdownItem>,
+    <DropdownItem aria-label="bulk_remove" key="bulk_remove" isDisabled={!bulkActionEnabled} component="button" onClick={removeBulk}>
+      Remove
+    </DropdownItem>,
+  ];
 
   return (
     <TableWrapper
@@ -144,6 +219,7 @@ const ContentViewRepositories = ({ cvId }) => {
         emptySearchBody,
         searchQuery,
         updateSearchQuery,
+        actionResolver,
         error,
         status,
         activeFilters,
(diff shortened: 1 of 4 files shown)


sjha4:
The remove button needs to be in the action list kebab. Otherwise, the functionality is working for me..
Screenshot from 2021-02-01 12-43-17

sjha4:
Screenshot from 2021-02-02 11-32-39

Moved bulk actions to kebab..

webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js
```diff
@@ -109,6 +180,9 @@ const ContentViewRepositories = ({ cvId }) => {
 
   useEffect(() => {
     dispatch(getRepositoryTypes());
+    if (Object.keys(details).length === 0 && detailStatus !== STATUS.RESOLVED) {
+      dispatch(getContentViewDetails(cvId));
```
sjha4, line 184: Had to add this for tests and to render ContentViewRepositories directly as an independent component.

johnpmitsch: I'd like to look into this more, generally I'd like to avoid adding something just for testing, I can help think about possible solutions.

webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js
```diff
@@ -60,6 +76,62 @@ const ContentViewRepositories = ({ cvId }) => {
...
+  const toggleBulkAction = () => {
+    setBulkActionOpen(!bulkActionOpen);
+  };
+  const onAdd = (repos) => {
+    const { repository_ids: repositoryIds } = details;
+    dispatch(updateContentView(cvId, { repository_ids: repositoryIds.concat(repos) }));
+  };
+
+  const onRemove = (repos) => {
+    const reposToDelete = [].concat(repos);
```
johnpmitsch, line 88: This is valid, but `const reposToDelete = [...repos];` is the more 'modern' way to copy an array

webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js
```diff
@@ -60,6 +76,62 @@ const ContentViewRepositories = ({ cvId }) => {
...
+    const { repository_ids: repositoryIds } = details;
+    const deletedRepos = repositoryIds.filter(x => !reposToDelete.includes(x));
```
johnpmitsch, line 90: naming nitpick - is this the deleted repos, or the repos that should remain in the CV? If the latter, can we call it something like `remainingRepos` or `stillAddedRepos` (or another suggestion)?

webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js
```diff
@@ -25,6 +37,7 @@ import RepoIcon from './RepoIcon';
 import SelectableDropdown from '../../../../components/SelectableDropdown';
 import { capitalize } from '../../../../utils/helpers';
 
+
```
johnpmitsch, line 40: I thought eslint picked these things up, but I don't think this blank line is needed

webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js
```diff
@@ -60,6 +76,62 @@ const ContentViewRepositories = ({ cvId }) => {
...
+    dispatch(updateContentView(cvId, { repository_ids: deletedRepos }));
+  };
+
+  const addBulk = () => {
+    const reposToAdd = [];
+    rows.forEach((row) => {
+      if (row.selected) {
+        reposToAdd.push(row.repoId);
+      }
+    });
```
johnpmitsch, line 100: I don't mind it spread out, but can be in one line like this `rows.forEach(row => row.selected && reposToAdd.push(row.repoId))`, same can be done in `removeBulk` below

johnpmitsch: actually you could do this:

```javascript
 const addBulk = () => {
    const reposToAdd = rows.filter(row => row.selected);
    onAdd(reposToAdd);
 }

File: webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js (hunk 2 above)
Comment (sjha4) [reply webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js:88]: So repos param here can be either a single repo or array of repos coming from bulk remove..

Comment (johnpmitsch) [reply webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js:87]: concat makes more sense then, thanks!

**Comment by johnpmitsch** [comment]
@sjha4 From the mockups, it looks like the "Add repositories" button should be disabled if nothing is checked. Maybe we should put a tooltip too when disabled "Please select repositories to add". I also wonder if the kebab should disappear until repositories are checked?

Another consideration - what if both the Added and Not Added repositories are checked and I click "Add"?

**Comment by sjha4** [comment]
> what if both the Added and Not Added repositories are checked and I click "Add"?

This throws an error right now..Should we change that? Is that a thing users will do?

> From the mockups, it looks like the "Add repositories" button should be disabled if nothing is checked. Maybe we should put a tooltip too when disabled "Please select repositories to add". I also wonder if the kebab should disappear until repositories are checked?

Updated to disable button and hide kebab until repositories are selected.

**Comment by sjha4** [comment]
[test katello]

**Comment by johnpmitsch** [comment]
> > what if both the Added and Not Added repositories are checked and I click "Add"?
> 
> This throws an error right now..Should we change that? Is that a thing users will do?

Right now it says `Validation failed: Repository has already been taken`, it would be nice if the error message was a little more helpful or we handle this in the UI somehow

**Comment by johnpmitsch** [comment]
> Updated to disable button and hide kebab until repositories are selected.

Nice, this looks good for the basic cases. However, there are some areas that are confusing:

1. I am in one view such as "All", check a repository, and switch to another one such as "Available", the button and the kebab stay highlighted, but nothing changes in this view. I can click "Add repositories" and then see a notification of "Content View updated", but nothing has changed. If buttons and kebab could disable/hide again after switching views, I think that would be preferred.

2. When in the "Not Added" view, the "Add repositories" button is still visible. Should we be limiting this or changing the button? I won't make a lot of sense in that view and only cause an errored call to be made if clicked. I could see only showing an "Remove repositories" button (and no kebab) in that particular view. Perhaps an open question for @marisvirik as well!

Let me know if you want to discuss the detail of implementing a fix for either issue!

File: webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js (hunk 5 above)
Comment (johnpmitsch) [reply webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js]: its still valid, but the second comment with `rows.filter` avoids initializing the empty array

File: webpack/scenes/ContentViews/Details/Repositories/__tests__/contentViewAddRemove.test.js [hunk 6]
```diff
@@ -0,0 +1,135 @@
...
+
+  await patientlyWaitFor(() => expect(getByText(firstRepo.name)).toBeTruthy());
+  expect(getByLabelText('Select all rows')).toBeInTheDocument();
+  expect(getByLabelText('add_repositories')).toHaveAttribute('aria-disabled', 'true');
+  getByLabelText('Select all rows').click();
+  getByLabelText('add_repositories').click();
+  assertNockRequest(autocompleteScope);
+  assertNockRequest(scope);
+  assertNockRequest(repoAddscope);
+  assertNockRequest(scope, done);
```
johnpmitsch, line 97: you shouldn't have to assert the same scope twice, you can keep the one with the `done` only, same for the others in this file

webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js
```diff
@@ -60,6 +76,56 @@ const ContentViewRepositories = ({ cvId }) => {
...
+      {
+        title: 'Remove',
+        isDisabled: repositoryIds && !repositoryIds.includes(rowData.repoId),
+        onClick: (_event, rowId, rowInfo) => {
+          onRemove(rowInfo.repoId);
+        },
+      },
+    ];
+  };
+
```
johnpmitsch, line 128: can you move all of these functions underneath the `useEffect` hooks? see [react docs]( https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level). They can move unless they are used in the effect hooks.

(`getCVReposWithOptions` can move as well, that was my mistake!)

webpack/scenes/ContentViews/Details/Repositories/SelectToggleHelper.js
```diff
@@ -0,0 +1,20 @@
...
+  } else {
+    newRows = [...rows];
+    newRows[rowId].selected = isSelected;
+  }
+
+  if (newRows.some(row => row.selected)) {
+    setBulkActionDisabled(false);
+  } else {
+    setBulkActionDisabled(true);
+  }
```
johnpmitsch, line 15: I think one thing that would help is if you moved this bulk action logic to an effect hook in the actual component:

```javascript
useEffect(() => {
    const rowsAreSelected = rows.some(row => row.selected)
    setBulkActionDisabled(!rowsAreSelected)
}, [JSON.parse(rows)]
```
This should keep up with the statuses switching if I'm reading things right! Also, `bulkActionEnabled` would be a little more readable both here and in the component IMO, double negatives tend to hurt my brain :brain:

webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js
johnpmitsch: Finally was able to look at this, you can add a default value of a blank array for repositoryIds in that section. You can also pass an  `initialApiState` value to the `renderOptions` object argument to the `renderWithRedux` function. This will allow you to pass in more repository ids that will show initially.

This was my attempt at fixing things, feel free to use anything here:

```diff
diff --git a/webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js b/webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js
index d519036..6a3f9b0 100644
--- a/webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js
+++ b/webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js
@@ -80,7 +80,7 @@ const ContentViewRepositories = ({ cvId }) => {
     setBulkActionOpen(!bulkActionOpen);
   };
   const onAdd = (repos) => {
-    const { repository_ids: repositoryIds } = details;
+    const { repository_ids: repositoryIds=[] } = details;
     dispatch(updateContentView(cvId, { repository_ids: repositoryIds.concat(repos) }));
   };
 
@@ -172,9 +172,6 @@ const ContentViewRepositories = ({ cvId }) => {
 
   useEffect(() => {
     dispatch(getRepositoryTypes());
-    if (Object.keys(details).length === 0 && detailStatus !== STATUS.RESOLVED) {
-      dispatch(getContentViewDetails(cvId));
-    }
   }, []);
 
   // Get repo type filter selections dynamically from the API
diff --git a/webpack/scenes/ContentViews/Details/Repositories/__tests__/contentViewAddRemove.test.js b/webpack/scenes/ContentViews/Details/Repositories/__tests__/contentViewAddRemove.test.js
index 8f9109d..729f608 100644
--- a/webpack/scenes/ContentViews/Details/Repositories/__tests__/contentViewAddRemove.test.js
+++ b/webpack/scenes/ContentViews/Details/Repositories/__tests__/contentViewAddRemove.test.js
@@ -15,7 +15,7 @@ const repoTypesResponse = [{ name: 'deb' }, { name: 'docker' }, { name: 'file' }
 const cvDetailData = require('../../__tests__/contentViewDetails.fixtures.json');
 
 const cvDetailsPath = api.getApiUrl('/content_views/1');
-const cvAddparams = { repository_ids: [58, 62, 64, 107, 106] };
+const cvAddparams = { repository_ids: [107, 106] };
 const cvRepoAddData = require('./contentViewRepoAdd.fixture.json');
 
 let firstRepo;
@@ -32,12 +32,7 @@ beforeEach(() => {
     .get(api.getApiUrl('/repositories/repository_types'))
     .query(true)
     .reply(200, repoTypesResponse);
-  nockInstance
-    .persist()
-    .get(cvDetailsPath)
-    .query(true)
-    .reply(200, cvDetailData);
-});
+ });
 
 afterEach(() => {
   nock.cleanAll();
@@ -79,6 +74,11 @@ test('Can add repositories', async (done) => {
     .put(cvDetailsPath, cvAddparams)
     .reply(200, cvRepoAddData);
 
+  const cvDetailScope = nockInstance
+    .get(cvDetailsPath)
+    .query(true)
+    .reply(200, cvDetailData);
+
   const scope = nockInstance
     .get(cvAllRepos)
     .query(true)
@@ -105,5 +105,6 @@ test('Can add repositories', async (done) => {
   });
   assertNockRequest(autocompleteScope);
   assertNockRequest(repoAddscope);
+  assertNockRequest(cvDetailScope);
   assertNockRequest(scope, done);
 });
diff --git a/webpack/scenes/ContentViews/Details/Repositories/__tests__/contentViewDetailRepos.test.js b/webpack/scenes/ContentViews/Details/Repositories/__tests__/contentViewDetailRepos.test.js
index b32310d..ebfc146 100644
--- a/webpack/scenes/ContentViews/Details/Repositories/__tests__/contentViewDetailRepos.test.js
+++ b/webpack/scenes/ContentViews/Details/Repositories/__tests__/contentViewDetailRepos.test.js
@@ -31,11 +31,6 @@ beforeEach(() => {
     .get(api.getApiUrl('/repositories/repository_types'))
     .query(true)
     .reply(200, repoTypesResponse);
-  nockInstance
-    .persist()
-    .get(cvDetailsPath)
-    .query(true)
-    .reply(200, cvDetailData);
 });
 
 afterEach(() => {
```

johnpmitsch:
@sjha4 The options disabling is working well when the status changes, as nothing is selected :+1: 

One thing that we touched on is what happens to the options when the status changes, should we disable the options that aren't relevant for the status?

For example, this is what I was envisioning with "Not added" repos, disabling the 'remove' option:
Screenshot from 2021-03-09 11-28-31

And for the Added repos, changing the button to "Remove repositories" and disabling the "add" option in the dropdown:
Screenshot from 2021-03-09 11-30-41

@MariSvirik thoughts?

webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js
johnpmitsch: @sjha4 were you able to look at this?

webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js
johnpmitsch: @sjha4 any reason to not use filter here? ^

sjha4: We actually need the repoId on the row in the reposToAdd array. Is there a JS way to pluck that from filtered rows?

johnpmitsch:
> @sjha4 The options disabling is working well when the status changes, as nothing is selected 
> 
> One thing that we touched on is what happens to the options when the status changes, should we disable the options that aren't relevant for the status?
> 
> For example, this is what I was envisioning with "Not added" repos, disabling the 'remove' option:
> Screenshot from 2021-03-09 11-28-31
> 
> And for the Added repos, changing the button to "Remove repositories" and disabling the "add" option in the dropdown:
> Screenshot from 2021-03-09 11-30-41
> 
> @MariSvirik thoughts?

This is still an open question but we can address this separately

sjha4:
Thanks @johnpmitsch ! Merging.. :)

MariSvirik:
@johnpmitsch @sjha4 
The first part (disabling irrelevant options) is correct. The second part:  if the table is filtered out, the button should change - the way you do. But I need to double-check if there is a pattern to change the buttons in other places (e.g. cloud.redhat.com). 

If the table is mixed (added and not added) and if a user has selected just 'Not added repos' we should probably use a disabled button 'Add repositories'.

PR #9137, opened 2021-02-02
Fixes #32093 - Remove registration template auto-assign to new OS

stejskalleos:
Remove the auto-association of registration template to Red Hat OS,
Foreman now assigns the template to all operating systems.
See https://projects.theforeman.org/issues/31663

### DIFF
--- a/app/models/katello/concerns/operatingsystem_extensions.rb
+++ b/app/models/katello/concerns/operatingsystem_extensions.rb
@@ -17,8 +17,6 @@ module Katello
           TemplateKind.all.each do |kind|
             if name == ::Operatingsystem::REDHAT_ATOMIC_HOST_OS && kind.name == "provision"
               provisioning_template_name = Setting["katello_default_atomic_provision"]
-            elsif kind.name == 'registration'
-              provisioning_template_name = 'Linux registration default'
             else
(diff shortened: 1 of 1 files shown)


jturel:
Issue 31402 is since closed and part of an old release. This PR should have a new issue of its own. If it should be part of 3.18.x like the original was then indicate that on the new redmine :)

ares:
@stejskalleos please update the redmine issue name, otherwise seems ready and should get in, when the core PR gets in

jturel:
Hey @stejskalleos thanks for the PR! Is this actually intended as part of #31402? That bug has a downstream BZ associated, so merging this as-is will mean that it's gonna get picked up by the downstream cherry-pick process which may not be desired at this point. If that's OK then we can merge as-is. If not, then there should be an entirely new issue for this change.

jturel:
Ah, I see there is a Redmine referencing this PR but the commit and PR name here don't refer to it. Guess the PR + commit just need an update :) 

https://projects.theforeman.org/issues/32093

stejskalleos:
@jturel 
Rebased & updated, changed the redmine issue to https://projects.theforeman.org/issues/32093

jturel (approved):
ACK!

Thanks @stejskalleos !

PR #9139, opened 2021-02-04
Add endpoint to show added and available content for cv filter

johnpmitsch:
This is mostly for UI purposes, but I don't see a reason to keep it hidden like we did for the cv repos addition. It will return all the added and available content for a content view filter, right now it's set up for package group filters, which has the `added_to_cv_filter` boolean in the response json. The response can be modified like this for other types too.

An example request with httpie:
`http -a admin:changeme centos7-katello-devel-stable.example.com/katello/api/v2/package_groups filterId==1 show_all_for==content_view_filter`

mockup is here: https://marvelapp.com/prototype/1121a8de/screen/73236934

### DIFF
--- a/app/controllers/katello/concerns/api/v2/repository_content_controller.rb
+++ b/app/controllers/katello/concerns/api/v2/repository_content_controller.rb
@@ -189,7 +189,7 @@ module Katello
 
       def find_filter
         # TODO: in v2.rb some routes use "filters", others use "content_view_filters"
-        filter_id = params[:content_view_filter_id] || params[:filter_id]
+        filter_id = params[:content_view_filter_id] || params[:filter_id] || params[:filterId]
 
         if filter_id
           scoped = ContentViewFilter.all
--- a/app/controllers/katello/concerns/api/v2/repository_content_controller.rb
+++ b/app/controllers/katello/concerns/api/v2/repository_content_controller.rb
@@ -28,6 +28,7 @@ module Katello
       param :repository_id, :number, :desc => N_("repository identifier")
       param :environment_id, :number, :desc => N_("environment identifier")
       param :ids, Array, :desc => N_("ids to filter content by")
+      param :include_filter_ids, :bool, desc: N_("Includes associated content view filter ids in response")
       param_group :search, ::Katello::Api::V2::ApiController
       def index
         sort_by, sort_order, options = sort_options
@@ -216,7 +217,7 @@ module Katello
 
       def check_show_all_and_available_params
         if params[:show_all_for] && params[:available_for]
-          fail HttpErrors::UnprocessableEntity, _("params 'show_all_for' and 'available_for' must be used independently")
+          fail HttpErrors::BadRequest, _("params 'show_all_for' and 'available_for' must be used independently")
         end
       end
 
@@ -286,9 +287,9 @@ module Katello
         else
           # Filtering by the CV filter rule makes filtering by the CV filter redundant, keeping these
           # exclusive to keep the queries simple.
-          if @filter_rule
+          if filter_rule
             collection = filter_by_content_view_filter_rule(filter_rule, collection)
-          elsif @filter
+          elsif filter
             collection = filter_by_content_view_filter(filter, collection)
           end
         end
--- a/app/controllers/katello/concerns/api/v2/repository_content_controller.rb
+++ b/app/controllers/katello/concerns/api/v2/repository_content_controller.rb
@@ -70,6 +70,8 @@ module Katello
       end
 
       param :available_for, :string, :desc => N_("Return content that can be added to the specified object.  The values 'content_view_version' and 'content_view_filter are supported.")
+      param :show_all_for, :bool,
+            :desc => N_("Returns content that can be both added and is currently added to the object. The value 'content_view_filter' is supported")
       param :filterId, :integer, :desc => N_("Content View Filter id")
       def index_relation
         if @version && params[:available_for] == "content_view_version" && self.respond_to?(:available_for_content_view_version)
@@ -83,17 +85,7 @@ module Katello
         collection = filter_by_ids(params[:ids], collection) if params[:ids]
 
         @filter = ContentViewFilter.find(params[:filterId]) if params[:filterId]
-        if params[:available_for] == "content_view_filter" && self.respond_to?(:available_for_content_view_filter)
-          collection = self.available_for_content_view_filter(@filter, collection) if @filter
-        else
-          # Filtering by the CV filter rule makes filtering by the CV filter redundant, keeping these
-          # exclusive to keep the queries simple.
-          if @filter_rule
-            collection = filter_by_content_view_filter_rule(@filter_rule, collection)
-          elsif @filter
-            collection = filter_by_content_view_filter(@filter, collection)
-          end
-        end
+        collection = handle_cv_filter(collection, @filter, @filter_rule, params)
 
         collection = self.custom_index_relation(collection) if self.respond_to?(:custom_index_relation)
         collection
@@ -279,6 +271,25 @@ module Katello
           fail HttpErrors::NotFound, _("Could not find %{content} with id '%{id}' in repository.") %
             {content: resource_name, id: params[:id]}
         end
+      end
+
+      def handle_cv_filter(collection, filter, filter_rule, params)
+        if params[:show_all_for] == "content_view_filter" && self.respond_to?(:available_for_content_view_filter)
+          available = available_for_content_view_filter(@filter, collection)
+          added = filter_by_content_view_filter(@filter, collection)
+          collection = available.or(added)
+        elsif params[:available_for] == "content_view_filter" && self.respond_to?(:available_for_content_view_filter)
+          collection = self.available_for_content_view_filter(filter, collection) if filter
+        else
+          # Filtering by the CV filter rule makes filtering by the CV filter redundant, keeping these
+          # exclusive to keep the queries simple.
+          if @filter_rule
+            collection = filter_by_content_view_filter_rule(filter_rule, collection)
+          elsif @filter
+            collection = filter_by_content_view_filter(filter, collection)
+          end
+        end
+        collection
       end
     end
   end
--- a/app/controllers/katello/concerns/api/v2/repository_content_controller.rb
+++ b/app/controllers/katello/concerns/api/v2/repository_content_controller.rb
@@ -12,6 +12,7 @@ module Katello
         before_action :find_content_view_version, :only => [:index, :auto_complete_search]
         before_action :find_filter, :find_filter_rule, :only => [:index, :auto_complete_search]
         before_action :find_content_resource, :only => [:show]
+        before_action :check_show_all_and_available_params, only: [:index]
       end
 
       extend ::Apipie::DSL::Concern
@@ -83,9 +84,7 @@ module Katello
 
         collection = filter_by_repos(repos, collection)
         collection = filter_by_ids(params[:ids], collection) if params[:ids]
-
-        @filter = ContentViewFilter.find(params[:filterId]) if params[:filterId]
-        collection = handle_cv_filter(collection, @filter, @filter_rule, params)
+        collection = handle_cv_filter(collection, @filter, @filter_rule, params) if @filter || @filter_rule
 
         collection = self.custom_index_relation(collection) if self.respond_to?(:custom_index_relation)
         collection
@@ -215,6 +214,12 @@ module Katello
         end
       end
 
+      def check_show_all_and_available_params
+        if params[:show_all_for] && params[:available_for]
+          fail HttpErrors::UnprocessableEntity, _("params 'show_all_for' and 'available_for' must be used independently")
+        end
+      end
+
       def resource_class
         "Katello::#{controller_name.classify}".constantize
       end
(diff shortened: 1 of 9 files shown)


app/views/katello/api/v2/package_groups/index.json.rabl
```diff
@@ -4,4 +4,8 @@ extends "katello/api/v2/common/metadata"
 
 child @collection[:results] => :results do
   extends "katello/api/v2/package_groups/base"
+
+  if (params[:show_all_for] == "content_view_filter" && @filter)
+    node(:added_to_content_view_filter) { |package_group| package_group.in_content_view_filter?(@filter) }
```
jturel, line 9: Would it be more flexible if we returned an array of the CV filter IDs which this package group belongs to ? Sorry if it's a bad question - still forming up my understanding of the change

app/controllers/katello/concerns/api/v2/repository_content_controller.rb
```diff
@@ -280,6 +272,25 @@ def check_repo_for_content_resource
             {content: resource_name, id: params[:id]}
         end
       end
+
+      def handle_cv_filter(collection, filter, filter_rule, params)
```
jturel, line 276: Is it intentional that this method takes the filter as a param but refers to the `@filter` instance variable?

app/controllers/katello/concerns/api/v2/repository_content_controller.rb
```diff
@@ -70,6 +70,8 @@ def compare
       end
 
       param :available_for, :string, :desc => N_("Return content that can be added to the specified object.  The values 'content_view_version' and 'content_view_filter are supported.")
+      param :show_all_for, :bool,
```
jturel, line 73: Are show_all_for and available_for mutually exclusive? Would be helpful to raise an error if so

jturel:
@johnpmitsch can you point me to the wireframe or actual view (if it exists) where this will be used?

jturel:
Ok! I think I have a good understanding of the intent here and I have some suggestions.

**UI Goal (your requirements)**

- When looking at a particular Package Group Filter on a CV, see a unified list of package groups applicable to the CV and indicate whether each group is present in the filter or not.

**API Goal (my requirements)**
- Generalized, flexible API (try to make params and output not specific to a specific view)
- efficient querying

I came up with this: https://github.com/jturel/katello/commit/5cd89a19b313dceec1784660693b4b3798456ec4 which:

- establishes a few associations that makes it easy to ask the question 'to what filters does this PG belong' 
- replaces the two queries in `handle_cv_filter` joined up with an `or` into a single method: `all_for_content_view_filter` to simplify the query (added to package_groups_controller)
- the index endpoint understands a boolean parameter `include_filter_ids` which adds the associated filter ids to the response which can be used by the UI code to determine membership. This is good because it separates the concepts of "show me everything for this pg filter" and "show me the filters to which this PG belongs"

Invoke the API like this: `curl -u admin:changeme "http://localhost:3000/katello/api/v2/package_groups?filterId=3&show_all_for=content_view_filter&include_filter_ids=true"`

Output is like this: 
```
[
    {
      "id": 2,
      "name": "birds",
      "pulp_id": "/pulp/api/v3/content/rpm/packagegroups/a5f81227-0fdc-4e84-98df-152b8b959ba5/",
      "uuid": "/pulp/api/v3/content/rpm/packagegroups/a5f81227-0fdc-4e84-98df-152b8b959ba5/",
      "description": "",
      "repository": {
        "id": 13,
        "name": "zoo2",
        "product": {
          "id": 286,
          "name": "Animals"
        }
      },
      "filter_ids": [
        3
      ]
    },
    {
      "id": 1,
      "name": "mammals",
      "pulp_id": "/pulp/api/v3/content/rpm/packagegroups/a4d08e7f-866a-4264-9fd5-d33419e67290/",
      "uuid": "/pulp/api/v3/content/rpm/packagegroups/a4d08e7f-866a-4264-9fd5-d33419e67290/",
      "description": "",
      "repository": {
        "id": 13,
        "name": "zoo2",
        "product": {
          "id": 286,
          "name": "Animals"
        }
      },
      "filter_ids": []
    }
  ]

```

@johnpmitsch hopefully I didn't miss the point here. What do you think about my approach?

johnpmitsch:
> @johnpmitsch hopefully I didn't miss the point here. What do you think about my approach?

@jturel Thanks for the detailed investigation and commit! It seems like a reasonable approach, search and pagination will work fine afaict. The other remaining requirement is the added content must show first, not sure if that is happening now, but I think we can figure something out if not.

jturel:
> > @johnpmitsch hopefully I didn't miss the point here. What do you think about my approach?
> 
> @jturel Thanks for the detailed investigation and commit! It seems like a reasonable approach, search and pagination will work fine afaict. The other remaining requirement is the added content must show first, not sure if that is happening now, but I think we can figure something out if not.

no problem. that sorting has to happen on the API side I guess, cause of pagination?

johnpmitsch:
> no problem. that sorting has to happen on the API side I guess, cause of pagination?

Yup, exactly.

I know with the previous endpoint added that did this [here](https://github.com/Katello/katello/pull/8862/files#diff-8e4ab8f10a0f9413b066e80b3b093c810f6a989060199a5150a4bec539534b8cR30) we had to add a custom sort query since the sorting happens on the db side during scoped_search. I think the same will have to happen here (hopefully not as complex), but looking at it now!

johnpmitsch:
@jturel updated! As we discussed, the order isn't preserved but it's still returning the filter ids in the response. I also added a check for the mutually exclusive "show_all_for" and "available_for" params and a test for it.

johnpmitsch:
[test katello]

app/controllers/katello/concerns/api/v2/repository_content_controller.rb
```diff
@@ -223,6 +214,12 @@ def find_filter_rule
         end
       end
 
+      def check_show_all_and_available_params
+        if params[:show_all_for] && params[:available_for]
+          fail HttpErrors::UnprocessableEntity, _("params 'show_all_for' and 'available_for' must be used independently")
```
jturel, line 219: I think this should be a BadRequest vs UnprocessableEntity

app/models/katello/content_view_package_group_filter_rule.rb
```diff
@@ -7,6 +7,12 @@ class ContentViewPackageGroupFilterRule < Katello::Model
                :inverse_of => :package_group_rules,
                :foreign_key => :content_view_filter_id
 
+    has_one :package_group,
+            class_name: "Katello::PackageGroup",
+            foreign_key: :pulp_id,
+            primary_key: :uuid,
+            dependent: :nullify
```
jturel, line 14: Hmm, I wonder if this should be destroy? I'm having trouble thinking through the implications for having this filter rule in place where the group no longer exists

test/models/package_group_test.rb
```diff
@@ -35,5 +36,10 @@ def test_search_returns_none
     def test_search_by_repository
       assert_includes PackageGroup.search_for('repository = "Fedora 17 x86_64"'), @mammals_pg
     end
+
+    def test_in_content_view_filter
```
jturel, line 40: I think this is now obsolete :)

app/controllers/katello/concerns/api/v2/repository_content_controller.rb
jturel: This is still outstanding

jturel (commented):
Looking great. I'll give a re-test after we work through the comments together

johnpmitsch:
@jturel I updated the associations and your comments, I only added the `dependent: :destroy` on the package group's `has_many`, since it doesn't make sense to have the package group filter rule if the package group is gone. I'm thinking messing around with nullify isn't smart here because we are using pulp id's as foreign keys, so that shouldn't be something we ever modify or erase. I just ignored the rubocop warning there but let me know if you think its worth re-evaluating.

cc @jlsherrill can you check the new association between `ContentViewPackageGroupFilterRule` and `PackageGroup` makes sense?

johnpmitsch:
Also, it looks like we have to use `has_one` according to section 4.1 here https://guides.rubyonrails.org/v6.0.0/association_basics.html#has_many-association-reference

>The belongs_to association creates a one-to-one match with another model. In database terms, this association says that this class contains the foreign key. If the other class contains the foreign key, then you should use has_one instead.

app/models/katello/content_view_package_group_filter_rule.rb
```diff
@@ -7,6 +7,14 @@ class ContentViewPackageGroupFilterRule < Katello::Model
                :inverse_of => :package_group_rules,
                :foreign_key => :content_view_filter_id
 
+    # rubocop:disable Rails/HasManyOrHasOneDependent
+    has_one :package_group,
+            class_name: "Katello::PackageGroup",
+            inverse_of: :content_view_filter_rules,
+            primary_key: :uuid,
+            foreign_key: :pulp_id
+    # rubocop:enable Rails/HasManyOrHasOneDependent
```
jlsherrill, line 16: i think you want to use belongs_to here, thats what the rubocop was complaining about.   I think all of the options are the same and it works very similarly.

johnpmitsch:
@jlsherrill ty, updated!

jturel:
I'm seeing a change in behavior in the List/Remove tab for the filter - it appears to be showing everything available to the filter (2 package groups in my case)

The UI makes this API request, and I can see 2 results with this PR, and 0 on master (when I have no groups added to my particular filter): `/katello/api/v2/package_groups?filterId=4&full_result=true&organization_id=1&page=1&paged=true&per_page=20&search=&sort_by=name&sort_order=ASC`

jturel:
Querying the API with just `include_filter_ids=true` throws a SQL error:

```
2021-02-22T21:34:26 [I|app|ed9b786b] Processing by Katello::Api::V2::PackageGroupsController#index as */*      
2021-02-22T21:34:26 [I|app|ed9b786b]   Parameters: {"include_filter_ids"=>"true", "api_version"=>"v2"}
```
```
2021-02-22T21:34:27 [I|app|ed9b786b]   Rendered /home/vagrant/katello/app/views/katello/api/v2/package_groups/index.json.rabl within katello/api/v2/layouts/collection (Duration: 139.8ms | Allocations: 75762)                               2021-02-22T21:34:27 [E|app|ed9b786b] ActionView::Template::Error: PG::UndefinedColumn: ERROR:  column katello_package_groups.uuid does not exist                                                                                               ed9b786b | LINE 1: ...ters"."id" INNER JOIN "katello_package_groups" ON "katello_p...                                                                                                                                                        
 ed9b786b |                                                              ^                                    
 ed9b786b | HINT:  Perhaps you meant to reference the column "katello_package_groups.id".                                                                                                                                                     
 ed9b786b |                                                                                          
```

app/models/katello/package_group.rb
```diff
@@ -4,6 +4,11 @@ class PackageGroup < Katello::Model
 
     CONTENT_TYPE = "package_group".freeze
     has_many :roots, :through => :repositories, :class_name => "Katello::RootRepository"
+    has_many :content_view_filter_rules,
+             class_name: "Katello::ContentViewPackageGroupFilterRule",
+             primary_key: :pulp_id,
+             foreign_key: :uuid,
+             dependent: :destroy
```
jlsherrill, line 11: based on the error, i think you've got primary_key and foreign_key swapped here.

johnpmitsch:
@jturel @jlsherrill I can't get rubocop, the association tests, and the actual functionality to agree with the association, so how do we feel about just having a method to find content view filters for a package group? I updated to this in the latest commit, let me know what you think!

@jturel we have both `filterId` and `filter_id` as params and `filterId` was missing from the `find_filter` method which the new approach is relying on. I added it there and it is filtering correctly again :+1:

app/models/katello/package_group.rb
```diff
@@ -29,5 +29,9 @@ def package_names
       group = service_class.new(self.pulp_id)
       group.default_package_names + group.conditional_package_names + group.optional_package_names + group.mandatory_package_names
     end
+
+    def content_view_filters
+      Katello::ContentViewPackageGroupFilterRule.where(uuid: self.pulp_id).map(&:filter)
```
jturel, line 34: I think I'm OK w/ this although it looks like there's an N+1 situation now, which was one of the benefits of using the associations. I think you can solve that by throwing `.eager_load(:filter)` on the query which will create the join

app/models/katello/content_view_package_group_filter.rb
```diff
@@ -4,6 +4,10 @@ class ContentViewPackageGroupFilter < ContentViewFilter
 
     has_many :package_group_rules, :dependent => :destroy, :foreign_key => :content_view_filter_id,
                                    :class_name => "Katello::ContentViewPackageGroupFilterRule"
+    has_many :package_groups,
```
jturel, line 7: I guess this isn't needed now with your revised approach!

app/models/katello/package_group.rb
johnpmitsch: That's a good point, I added it here.

FWIW I played around with this and came up with a query that I think works for what we are trying to do:
```SQL
SELECT kcvf.id
FROM katello_content_view_filters AS kcvf
INNER JOIN katello_content_views AS kcv ON kcv.id = kcvf.content_view_id
INNER JOIN katello_content_view_package_group_filter_rules AS kcvpgfr ON kcvpgfr.content_view_filter_id = kcvf.id
INNER JOIN katello_package_groups AS kpg ON kcvpgfr.uuid = kpg.pulp_id
WHERE kpg.id = 2;
```
(replace 2 with the package group id), its pretty similar with to what you suggested, but the trouble is "railsifying" it.

I'm ok with whats there now, but if you think it's worth it we can figure out how to get the stars aligned with the associations :star2:

jturel: That query looks accurate. We've gone this far without having the associations I was looking to have added, and your approach is more straightforward so I'm happy to stick with it. Will give it a test today

jturel:
Tested with these combinations of API params and got the expected results:

`curl -u admin:changeme "http://localhost:3000/katello/api/v2/package_groups?include_filter_ids=true"`
`curl -u admin:changeme "http://localhost:3000/katello/api/v2/package_groups?include_filter_ids=true&filterId=1" | jq .`
`curl -u admin:changeme "http://localhost:3000/katello/api/v2/package_groups?include_filter_ids=true&filterId=1&show_all_for=content_view_filter" | jq .
`
`curl -u admin:changeme "http://localhost:3000/katello/api/v2/package_groups?include_filter_ids=true&filterId=1&available_for=content_view_filter" | jq .
`

Nice work!

jturel (approved):
ACK!

Friendly reminder to squash :)

PR #9140, opened 2021-02-04
Fixes #31812 - Avoid race conditions in CandlepinMessageHandler

jturel:
the `pool` and `subscription_facet` methods of CandlepinMessageHandler previously loaded from the db on each call. I refactored the code to use instance variables when the handler is initialized to avoid those and potential race conditions as a consequence where (for example) the Pool has already been deleted from the Katello db:
```
      def remove_pool_from_host
        return if self.subscription_facet.nil? || pool.nil?
        old_host_ids = pool.subscription_facets.pluck(:host_id)
        ::Katello::SubscriptionFacetPool.where(subscription_facet_id: self.subscription_facet.id,
                                               pool_id: pool.id).destroy_all
        pool.import_audit_record(old_host_ids)
      end
```

We could enter into the method (avoiding the first return) but get nil errors when referring to `pool.id` or `subscription_facet.id` again if those entities were deleted in that narrow window of time due to querying the db on each call.

There are also a few other defensive checks added!

### DIFF
--- a/test/services/candlepin/message_handler_test.rb
+++ b/test/services/candlepin/message_handler_test.rb
@@ -2,10 +2,11 @@ require 'katello_test_helper'
 
 module Katello
   class MessageHandlerTestBase < ActiveSupport::TestCase
+    let(:handler) { ::Katello::Candlepin::MessageHandler.new(@event) }
+
     def setup
       json = File.read("#{Katello::Engine.root}/test/fixtures/candlepin_messages/#{event_name}.json")
-      event = OpenStruct.new(subject: event_name, content: json)
-      @handler = ::Katello::Candlepin::MessageHandler.new(event)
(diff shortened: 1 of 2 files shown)


test/services/candlepin/message_handler_test.rb
```diff
@@ -2,10 +2,11 @@
 
 module Katello
   class MessageHandlerTestBase < ActiveSupport::TestCase
+    let(:handler) { ::Katello::Candlepin::MessageHandler.new(@event) }
```
jeremylenz, line 5: :heart:

jeremylenz (approved):
Thanks @jturel !

I think the passing tests are good enough to trust that this doesn't break anything, and the code change LGTM :)

PR #9143, opened 2021-02-05
Fixes #31828 - Delete client queue on unregister

jturel:
This PR mirrors the Pulp2 functionality of deleting the agent queue when a host is unregistered in that it will wait for ten minutes to actually delete the queue. If the client is still connected at unregistration time, goferd will re-create the queue which is the reason for the wait => hopefully the client has been turned off or destroyed by that point. The waiting is achieved by the existing mechanism of Katello's event queue which also decouples the unregistration process from Qpid.

**To test:**

- ensure your dev environment was created on or after 2.5.2021
- check out this pr
- register a content host (ie vagrant up centos7-katello-client) with katello-agent installed
- turn goferd off on the content host
- verify existence of client's queue. $UUID is derived from `subscription-manager identity` and can be found in many ways
```
qpid-stat --ssl-certificate=/etc/pki/katello/qpid_router_client.crt --ssl-key=/etc/pki/katello/qpid_router_client.key -b amqps://localhost:5671 -q | grep $UUID
```
- unregister the host using your favorite method
- use the above command to see the queue is still there
- check again in ten minutes and see that it got removed. you'll see something in your Rails log to this effect:
`19:13:00 rails.1   | 2021-02-05T19:13:00 [D|kat|] event_queue_event: type=delete_host_agent_queue, object_id=1`
- now use your imagination and try to break it so that I can fix it

### DIFF
--- a/app/services/katello/registration_manager.rb
+++ b/app/services/katello/registration_manager.rb
@@ -115,6 +115,8 @@ module Katello
         # This can be cleaned up later via clean_backend_objects.
         pulp_consumer_destory(host.content_facet.uuid) if host.content_facet.try(:uuid) && pulp2_supported?
 
+        delete_agent_queue(host) if host.content_facet.try(:uuid)
+
         host.subscription_facet.try(:destroy!)
 
         if unregistering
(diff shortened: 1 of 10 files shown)


app/services/katello/registration_manager.rb
```diff
@@ -266,6 +268,14 @@ def pulp_consumer_destory(host_uuid)
         Rails.logger.warn(_("Pulp Consumer %s has already been removed") % host_uuid)
       end
 
+      def delete_agent_queue(host)
+        queue_name = Katello::Agent::Dispatcher.host_queue_name(host)
```
jturel, line 272: Since the host record (including content facet) will be gone by the time we issue the queue delete to Qpid, we need a way to derive what its queue name would be, and that's what this method does.

parthaa (approved):
Works well. Tested unregister both via ui and subman.

PR #9145, opened 2021-02-05
Fixes #31799 - Use new async Candlepin API for manifest import

jeremylenz:
* Modify `::Katello::Resources::Candlepin::Owner.import` to use new async Candlepin API endpoint `{candlepin}/owners/{owner_key}/imports/async`
* Obtain the Candlepin job id from the response
* ~~Inside `::Actions::Katello::Organization::ManifestImport` and `ManifestRefresh`, trigger a new task that inherits from~~ Change `Actions::Candlepin::Owner::Import` to inherit from `Actions::Candlepin::AbstractAsyncTask` and poll the Candlepin task until finished
* also sets `humanized_input` which fixes bad grammar on Tasks page


app/lib/actions/katello/organization/manifest_import.rb
```diff
@@ -11,19 +11,25 @@ def plan(organization, path, force)
           manifest_update = organization.products.redhat.any?
 
           sequence do
-            plan_action(Candlepin::Owner::Import,
+            cp_import_task = plan_action(Candlepin::Owner::Import,
                         :label => organization.label,
                         :path => path,
                         :force => force)
-            plan_action(Candlepin::Owner::ImportProducts, :organization_id => organization.id)
+            concurrence do
```
jturel, line 18: Can these really be concurrent?

jeremylenz: I was only thinking of the Dynflow logic but I think you're right, I didn't consider the products may have changed after the import.  I'll change it back.

app/lib/actions/candlepin/owner/async_import.rb
```diff
@@ -0,0 +1,23 @@
...
+      # Candlepin::AbstractAsyncTask uses Actions::Base::Polling
+      # (which is an alias for Dynflow::Polling) and defines the run method to initiate polling.
+      # Therefore, don't override the run method here or in any dependent tasks.
+      class AsyncImport < Candlepin::AbstractAsyncTask
+        def poll_external_task(task_id = external_task[:id])
+          Rails.logger.info "Polling candlepin task"
+          task = ::Katello::Resources::Candlepin::Job.get(task_id, :result_data => true)
+          Rails.logger.info task['state']
+          task
+        end
```
jlsherrill, line 14: what was the reason to override this 'poll_external_task' method from Candlepin::AbstractAsyncTask ?

app/lib/actions/candlepin/owner/import.rb
```diff
@@ -11,7 +11,11 @@ class Import < Candlepin::Abstract
 
         def run
           options = input.slice(:force, :upstream)
-          output[:response] = ::Katello::Resources::Candlepin::Owner.import(input[:label], input[:path], options)
+          cp_response = JSON.parse(
+            ::Katello::Resources::Candlepin::Owner.import(input[:label], input[:path], options)
+          )
+          output[:task] = cp_response # this also sets external_task
+          output[:task_id] = cp_response['id']
```
jlsherrill, line 18: It looks like you're taking the cadlepin task  data and passing it to AsyncImport, but all this can be done within this one action.  If you just make Import inherit from Candlepin::AbstractAsyncTask,  rename the run method to 'invoke_external_task', and then have it return the task hash, it should all be done in one action with no need to pass it around

jlsherrill: ahh i see you modeled this after the hypervisors task.  Its kind of a special case, i would probably more model this after all of the pulp actions, which just initiate and poll in the same action

jeremylenz: @jlsherrill I originally had it like that but it wasn't actually initiating polling.  I figured out that it's because `Actions::Base::Polling` (aka Dynflow::Polling) defines the `run` method which is where the [first call of `invoke_external_task`](https://github.com/Dynflow/dynflow/blob/e47118cd6bc6f972bec38844a6b06dd8b0739cca/lib/dynflow/action/polling.rb#L19) is, so overriding the `run` method would prevent polling from ever starting.  That's why I had separated it into the two actions.

jlsherrill: right, but if you change it to invoke_external_task, it should work and you can do it within a single action, similar to: https://github.com/Katello/katello/blob/master/app/lib/actions/pulp3/repository/sync.rb#L11-L14

jeremylenz: > rename the run method to 'invoke_external_task'

Whoops, glossed right over that part. :see_no_evil: I'll give it a try

jeremylenz:
@jlsherrill combined into a single task per your suggestion :)

app/lib/actions/candlepin/owner/async_import.rb
jeremylenz: It was just for the logging.  It's removed now since `AbstractAsyncTask` handles it

app/lib/actions/katello/organization/manifest_import.rb
```diff
@@ -23,7 +23,7 @@ def plan(organization, path, force)
                 plan_action(Katello::Repository::RefreshRepository, repo)
               end
             end
-            plan_self
+            plan_self(candlepin_task: cp_import_task.output[:task])
```
jlsherrill, line 26: i don't think this 'candlepin_task' passing around is needed anymore, right?  (i see it being passed through to other actions, but its not really used)

app/lib/actions/candlepin/owner/async_import.rb
jeremylenz: Actually now that I think of it, I think I wasn't aware at the time that `AbstractAsyncTask` provides an implementation of `poll_external_task`.  I had just implemented it this way because I discovered `Katello::Resources::Candlepin::Job.get` was available and I needed it for the return value here.  I'm glad I arrived at basically the same code that was already there :smile:

app/lib/actions/katello/organization/manifest_import.rb
jeremylenz: I just really like seeing the candlepin task details on the Tasks page.

jeremylenz: I suppose it's not needed in both input and output though

jeremylenz: Problem is there's no way to have it just in output, which would make the most sense.  In order to get to output it has to be passed.

jlsherrill: you dig into the dynflow console to get at the subtask details and see it there, is that sufficient?

jeremylenz: For a developer I guess that's fine, but users would rarely if ever go into the dynflow console.  I feel like it's a better user experience to leave it in.

app/lib/actions/katello/organization/manifest_refresh.rb
```diff
@@ -70,12 +70,7 @@ def humanized_name
         # results in correct grammar on Tasks page,
         # e.g. "Refresh manifest for organization Default Organization"
         def humanized_input
-          [
-            [:organization, {
-              :text => "for organization '#{input[:organization_name]}'",
-              :link => "/organizations/#{input[:organization_id]}/edit"
-            }]
-          ]
+          "for organization '#{input[:organization_name]}'"
```
jeremylenz, line 73: Changed to returning just a string here.  For some reason it wasn't working before, but I think it was just because I hadn't restarted the server or something.

app/lib/actions/katello/organization/manifest_import.rb
jlsherrill: normal users probably aren't going to go looking for a candlepin task either i wouldn't think.  I ask this because we generate these task in sub-actions in dozens of places and we don't generally propagate task data up, so this would be an exception to what we do elsewhere.

jeremylenz: Okaaaaay, removing.  :cry:

jeremylenz: Added https://projects.theforeman.org/issues/31876 as an RFE

jlsherrill (approved):
ACK! Tested great!

jlsherrill:
don't forget to squash :)

app/lib/actions/katello/organization/manifest_import.rb
jeremylenz: @jlsherrill, see @adamruzicka comment here:  https://projects.theforeman.org/issues/31876#change-145787

Do you think it's possible for me to use that presenter in a way that would both give the better user experience and be consistent with katello code?

jlsherrill: what kind of information from the task is it you're interested in exposing?

jlsherrill: looking at the task output:

```
---
task:
  created: '2021-02-15T20:47:37+0000'
  updated: '2021-02-15T20:48:11+0000'
  id: 4028fa387792895c0177a772d8b300ec
  name: Import Manifest
  group: 
  origin: devel.balmora.example.com
  executor: devel.balmora.example.com
  principal: foreman_admin
  state: FINISHED
  previousState: RUNNING
  startTime: '2021-02-15T20:47:37+0000'
  endTime: '2021-02-15T20:48:11+0000'
  attempts: 1
  maxAttempts: 1
  statusPath: "/jobs/4028fa387792895c0177a772d8b300ec"
  resultData:
    created: 1613422091232
    updated: 1613422091232
    id: 4028fa387792895c0177a7735be03445
    status: SUCCESS_WITH_WARNING
    statusMessage: Default_Organization file imported successfully.One or more inactive
      subscriptions found in the file.
    fileName: import_b51c6c50c540001a6402.zip
    generatedBy: rhn-engineering-jsherrill
    generatedDate: 1598646848281
    upstreamConsumer:
      created: 1613422091232
      updated: 1613422091232
      id: 4028fa387792895c0177a7735be03446
      uuid: dc13b59f-2d6c-4072-930e-631abfd3c08b
      name: jsherrill
      type:
        created: 1613422057938
        updated: 1613422057938
        id: 4028fa387792895c0177a772d9d200ed
        label: satellite
        manifest: true
      ownerId: 4028fa387792895c0177928eb7630001
      contentAccessMode: 
      webUrl: access.redhat.com/management/subscription_allocations/
      apiUrl: https://subscription.rhsm.redhat.com/subscription/consumers/
  key: ImportJob
```

maybe the status message?

jlsherrill: here's an example of looking up a sub action and passing that as a presenter, in this case it uses the presenter of a sub-action: 
https://github.com/Katello/katello/blob/master/app/lib/actions/katello/repository/sync.rb#L102-L107

you could also just grab the output directly iirc, if you want, for example:

all_planned_actions(Pulp::Repository::Sync).first.output

app/lib/actions/candlepin/owner/import.rb
```diff
@@ -1,18 +1,24 @@
...
           param :path
           param :force
           param :upstream
         end
 
-        def run
+        def invoke_external_task
           options = input.slice(:force, :upstream)
-          output[:response] = ::Katello::Resources::Candlepin::Owner.import(input[:label], input[:path], options)
+          cp_response = JSON.parse(
```
jturel, line 14: How do you feel about doing the JSON parsing in the Owner.import method?

app/lib/actions/candlepin/owner/import.rb
```diff
@@ -1,18 +1,24 @@
...
+            ::Katello::Resources::Candlepin::Owner.import(input[:label], input[:path], options)
+          )
+          output[:task] = cp_response
+          output[:task_id] = cp_response['id']
+          poll_external_task # this return value sets external_task for future calls
```
jturel, line 19: Sorry I'm jumping on late here. Just thinking back to when I was doing to katello agent stuff with polling which I later switched away from.

My method would have looked more like this:

```
       def invoke_external_task
          options = input.slice(:force, :upstream)
          ::Katello::Resources::Candlepin::Owner.import(input[:label], input[:path], options)
        end
```

If the other stuff is really needed then ignore this

app/lib/actions/katello/organization/manifest_refresh.rb
```diff
@@ -6,7 +6,7 @@ class ManifestRefresh < Actions::AbstractAsyncTask
 
         include Helpers::Notifications
 
-        def plan(organization)
+        def plan(organization) # rubocop:disable Metrics/MethodLength
```
jturel, line 9: Maybe you can move the code block which does the repo sync stuff into a separate method to avoid disabling this?

```
do_repo_syncs_and_stuff if manifest_update
```

app/lib/actions/katello/organization/manifest_import.rb
jeremylenz: It turns out the two pieces of data I would want to include in the output are the same ones currently shown in Manifest History (Status and Message).  So after today's journey of discovery I think I'm going to just leave the UI as-is. :man_shrugging:

app/lib/actions/candlepin/owner/import.rb
jeremylenz: I think this was left over from when I was trying to figure out `Dynflow::Polling` and looking at https://github.com/Dynflow/dynflow/blob/e47118cd6bc6f972bec38844a6b06dd8b0739cca/lib/dynflow/action/polling.rb#L50

I said to myself, "Oh I better set `output[:task]`!"

jeremylenz:
@jturel @jlsherrill updated :)

jturel (approved):
untested, but LGTM. Nice work @jeremylenz

jlsherrill (approved):

jeremylenz:
[test katello]

jeremylenz:
[test katello]

jeremylenz:
[test katello]

jturel:
[test katello]

PR #9148, opened 2021-02-09
Fixes #31839 - Viewer role create CV button

sjha4:


jeremylenz (approved):
Thanks @sjha4 

Works as expected on the Content Views page:
* Admin user can see "Create New View" button
* Viewer user does not have "Create New View" button

On the new `/labs/content_views` page, Viewer user can still see the Create Content View button.  I think a separate issue should be created for this-- should probably make sure all the new CV pages behave as expected with permissions.

sjha4:
> ly make sure all the new CV pages behave as expected with permissions.



> On the new `/labs/content_views` page, Viewer user can still see the Create Content View button. I think a separate issue should be created for this-- should probably make sure all the new CV pages behave as expected with permissions.

@jeremylenz Ya I was thinking about that..I'll file a card..

PR #9149, opened 2021-02-09
[WIP] Fixes #29912 - support logging cid for pulp3

jjeffers:


app/services/katello/pulp3/api/ansible_collection.rb
```diff
@@ -25,7 +25,10 @@ def self.repository_sync_url_class
         end
 
         def api_client
-          PulpAnsibleClient::ApiClient.new(smart_proxy.pulp3_configuration(PulpAnsibleClient::Configuration))
+          client = PulpAnsibleClient::ApiClient.new(smart_proxy.pulp3_configuration(PulpAnsibleClient::Configuration))
+          request_id = ::Logging.mdc['request']
+          client.default_headers['Correlation-ID'] = request_id if request_id
+          client
```
jlsherrill, line 31: can we extract this out in some way so the code isn't repeated?

jlsherrill:
I think you're missing some of the Api classes for content types, like these: app/services/katello/pulp3/rpm.rb

app/services/katello/pulp3/api/ansible_collection.rb
jlsherrill: one way potentially would be to refactor all these classes to have an 'api_client_class' method, and then api_client could live in the parent class  "Katello::Pulp3::Api::Core" and have the logic to do all this

jlsherrill:
Its also missing all the Distributions, Publication, Remote, importer, exporter, api classes  (maybe some more?)

app/services/katello/pulp3/api/ansible_collection.rb
jjeffers: re extraction refactor:Absolutely. I considered it, but wanted to get eyes on the overall effort first.

jjeffers:
> I think you're missing some of the Api classes for content types, like these: app/services/katello/pulp3/rpm.rb

I am dismayed that there are client api instantiations outside the api classes.

jlsherrill:
looks like there's a rubocop failure, but ACK from a functionality/code perspective. 
Do you plan to open a PR to https://github.com/theforeman/puppet-pulpcore/blob/master/templates/pulpcore-api.service.erb  to add this logging option?  https://github.com/pulp/pulp_installer/blob/master/roles/pulp_api/templates/pulpcore-api.service.j2#L25

jjeffers:
> Do you plan to open a PR...?
https://github.com/theforeman/puppet-pulpcore/pull/175

jlsherrill (approved):

PR #9153, opened 2021-02-09
Fixes #31836 - Unclear error message for cvv promotion with bad perrms

ianballou:
This PR brings back the error message that reports which permissions could be missing that went away with https://github.com/Katello/katello/pull/8956

The root of the issue here is that `promote_or_remove_content_views_to_environments` has a special relationship to `promote_or_remove_content_views`.  The whole permissions model seems to expect permissions to be separate and not have relationships, which isn't always the case.  The PR here does seem like a bit of a workaround, but it also seems to be the least complicated fix at least.

A better long-term solution may be to formally declare these special relationships somehow.  I'm not sure if it's worth the time to do that here.

I'm curious to hear if there are ideas about better ways to solve the issue.  I'm not sure how this change will interact with other permissions so I'm marking it a draft for now.

To test:
1) Go through the scenario in the redmine
2) Test accessing other things with the UI and Hammer to make sure that my new messages doesn't break anything.

### DIFF
--- a/app/controllers/katello/concerns/api/v2/authorization.rb
+++ b/app/controllers/katello/concerns/api/v2/authorization.rb
@@ -39,7 +39,20 @@ module Katello
       end
 
       def throw_resource_not_found(name: resource_name, id: params[:id])
-        fail HttpErrors::NotFound, _("Could not find %{name} resource with id %{id}") % {id: id, name: name}
+        perms_message = "Potential missing permissions: " +
+          missing_permissions.map(&:name).join(', ')
+        fail HttpErrors::NotFound, _("Could not find %{name} resource with id %{id}. %{perms_message}") % {id: id, name: name, perms_message: perms_message}
+      end
+
+      def missing_permissions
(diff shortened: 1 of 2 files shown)


chris1984 (approved):
ACK, works good:

Before:
```shell
[vagrant@hammer ~]$ hammer -u test -p changeme content-view version promote --content-view-id 2 --organization-id 1 --to-lifecycle-environment-id 1
Could not promote the content view:
  Could not find content_view_version resource with id 2
```

With patch:
```shell
[vagrant@hammer ~]$ hammer -u test -p changeme content-view version promote --content-view-id 2 --organization-id 1 --to-lifecycle-environment-id 1
Could not promote the content view:
  Could not find content_view_version resource with id 2. Potential missing permissions: promote_or_remove_content_views, promote_or_remove_content_views_to_environments
```

ianballou:
For future reference, the weird permissions relationship between lifecycle environments and content views is rooted here: https://github.com/Katello/katello/blob/master/app/models/katello/authorization/content_view.rb#L24

PR #9155, opened 2021-02-10
Fixes #31861 - i18n - pulling from tx

jjeffers:


chris1984 (approved):

jjeffers:
@jturel since this will get merged (into master?) should this need a redmine issue number to be picked up by CPs for 4.0 release candidates?

jturel:
> @jturel since this will get merged (into master?) should this need a redmine issue number to be picked up by CPs for 4.0 release candidates?

I guess there could be an argument for having a Redmine with the translations PR, but it'll need a manual cherry-pick in this case. Ideally the translations are merged into main branch *before* the release branches (not sure of the procedure reflects that) and we never have to worry about picking it over in that case

jjeffers:
[test katello]

PR #9158, opened 2021-02-10
Fixes #31854 - Fix blank screen after editing entitlements

jeremylenz:
### The issue

Confirming a change to entitlements on the subscriptions table causes `componentDidUpdate` to run on `ManageManifestModal` even though it is not displayed on the screen.  `componentDidUpdate` then tries to close the Delete Manifest confirmation dialog, which doesn't exist yet.

### To recreate

Go to Content > Subscriptions
Refresh the page in the browser (this will clear the Redux store.  In order to get the error, the `deleteManifestModal` must not already be in the store.)
Edit the number of entitlements for a subscription
Click the checkmark and also confirm on the next confirmation dialog
At this point, the page will turn blank and you'll see a console error `SET_MODAL_CLOSED error: deleteManifestModal does not exist`

~~side note: Now that it's causing bugs, this is probably a good time to look at completing https://projects.theforeman.org/issues/30724.~~
_update_: See https://github.com/theforeman/foreman/pull/8322  (if that is merged, this PR will merely prevent a console warning rather than a blank page.)

### DIFF
--- a/webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
+++ b/webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
@@ -51,8 +51,11 @@ class ManageManifestModal extends Component {
   showDeleteManifestModal = () =>
     this.props.setModalOpen({ id: DELETE_MANIFEST_MODAL_ID });
 
-  hideDeleteManifestModal = () =>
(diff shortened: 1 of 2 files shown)


parthaa (approved):
Works great. ACK

jeremylenz:
tests passed locally, merging

PR #9159, opened 2021-02-11
Fixes #31862 - eslint error in Tasks/helpers.js

jturel:
Looks like the recent release of eslint-plugin-promise broke us due to this change https://github.com/xjamundx/eslint-plugin-promise/pull/202

While I think it could be a bug there, we can easily fix it to keep our builds passing and PRs merging, so that's what this PR does :)


parthaa (approved):

xjamundx:
I mean you could consider disabling the rule until we figure out a solution

PR #9161, opened 2021-02-11
Fixes #31863 - upgrade to 0.7.0 migration plugin (3.18)

jlsherrill:


jlsherrill:
I ran the migration tests live and there were no errors, if they pass with no changes, i think this is good to go

ianballou (approved):

PR #9163, opened 2021-02-11
Fixes #31864 - Add cell formatter for entitlements column

jeremylenz:
### Issue
While a manifest task is in progress (refresh, add subscription, delete, etc), the subscription entitlements count is showing as -1 for custom products.

### Cause
The `canManageSubscriptionAllocations` permission is set to a falsey value during this time.  (See `getEntitlementsFormatter` in `webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTableSchema.js`) This causes the formatter for the table cell to fall back to the `cellFormatter` instead of `entitlementsInlineEditFormatter`, and thus the raw value is passed directly to the table with no formatting.

### To recreate
On the Subscriptions page, make sure you have at least one custom product (the Entitlement column for that product should show quantity as 'Unlimited').
Now you can see the issue one of 3 ways:
1. Refresh the browser, and notice 'Unlimited' changes to '-1' only while the page is loading.
2. Log in as or impersonate a user with only Viewer role.  You will see '-1' even after the page is finished loading.
3. Start a manifest task (add subscriptions, delete subscriptions, refresh manifest, etc.)  You will see '-1' behind the task progress dialog.

### Solution
The formatting logic should be applied regardless of whether or not the user has edit permissions.  I added a new Patternfly cell formatter for the Entitlements cells, and made sure the formatting logic runs from both formatters.

### DIFF
--- a/webpack/components/pf3Table/formatters/EntitlementsInlineEditFormatter.js
+++ b/webpack/components/pf3Table/formatters/EntitlementsInlineEditFormatter.js
@@ -11,22 +11,26 @@ const renderValue = (rawValue, additionalData, onActivate) => {
   const value = getEntitlementsDisplayValue({
     rawValue, available, collapsible, upstreamPoolId,
   });
+  const editable = (value === rawValue);
 
   return (
-    <td className="editable">
-      <div
-        onClick={() => onActivate(additionalData)}
-        onKeyPress={(e) => {
-          if (e.keyCode === KEYCODES.ENTER) {
-            onActivate(additionalData);
-          }
-        }}
-        className="input"
-        role="textbox"
-        tabIndex={0}
-      >
-        {value}
-      </div>
+    <td className={editable ? 'editable' : ''}>
+      {editable &&
+        <div
+          onClick={() => onActivate(additionalData)}
+          onKeyPress={(e) => {
+            if (e.keyCode === KEYCODES.ENTER) {
+              onActivate(additionalData);
+            }
+          }}
+          className="input"
+          role="textbox"
+          tabIndex={0}
+        >
+          {value}
+        </div>
+      }
+      {!editable && value}
     </td>
   );
 };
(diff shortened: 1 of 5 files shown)


webpack/components/pf3Table/formatters/EntitlementsInlineEditFormatter.js
```diff
@@ -3,35 +3,34 @@ import { sprintf, translate as __ } from 'foremanReact/common/I18n';
...
-    );
-  } else if (available < 0 || !upstreamPoolId) {
-    return (
-      <td>{available < 0 ? __('Unlimited') : available}</td>
-    );
-  }
+  const value = getEntitlementsDisplayValue({
+    rawValue, available, collapsible, upstreamPoolId,
+  });
+  const editable = (value === rawValue);
```
jeremylenz, line 14: I'm not sure that this is the correct check.  :thinking:

jturel: This check aside for a moment, here's what I believe the overall functionality of this formatter should be so you can validate against that:

- If there are multiple subscriptions (iirc we group them by the name) then show NA because the row is `collapsible` and will never be editable as it does not correspond to a single subscription
- If there is an `upstreamPoolId` for the row (it maps to some subscription in the portal) then it should be `editable`
- If the value of `rawValue` (I think that's the quantity added in the manifest for example?) is -1, then show Unlimited as the text. I think this only relates to:
   - custom products
   - subscriptions created 'on the fly' in local Candlepin 
   - In other words these probably shouldn't be editable

jturel: All that said, it seems like this would work for `editable`
`const editable = !!upstreamPoolId;`

And a more defensive check could be:

`const editable = !!upstreamPoolId && rawValue != -1`

Because subscriptions on the portal are attached in some quantity or another.

And of course you should validate what I've told you here against your own understanding and testing! If this affects the existing test cases then those should be carefully reviewed

jeremylenz: It looked good to me, but failed the test "`renders unlimited for -1`" because it was making 'Unlimited' editable.  I changed the check to `typeof value === 'number'`.  This should cover both `NA` and `Unlimited`, and hopefully make `editable` correct.   let me know your thoughts.

jturel: that makes sense to me. code looks good. will test on monday!

jturel (approved):
Tested a number of scenarios on the subscriptions page and it all worked as expected without the anomaly mentioned in the bug report. ACK!

PR #9166, opened 2021-02-16
Fixes #31880 - correctly use SSL for qpid connections

jturel:
Correctly sets up the SSL connection and handles closing qpid connections a little better. I need to give this a more thorough test but my dev env is in an odd state..

To be tested in conjunction with https://github.com/theforeman/puppet-katello/pull/384 and https://github.com/theforeman/puppet-foreman_proxy_content/pull/327/files

### DIFF
--- a/app/lib/katello/qpid/connection.rb
+++ b/app/lib/katello/qpid/connection.rb
@@ -49,17 +49,18 @@ module Katello
       end
 
       class Receiver < ::Qpid::Proton::MessagingHandler
-        def initialize(url, address, handler)
+        def initialize(url, connection_options, address, handler)
           super()
           @url = url
+          @connection_options = connection_options.merge(
+            idle_timeout: 30
+          )
           @address = address
           @handler = handler
         end
 
         def on_container_start(container)
-          c = container.connect(@url,
-            idle_timeout: 4
-          )
+          c = container.connect(@url, @connection_options)
           c.open_receiver(@address)
         end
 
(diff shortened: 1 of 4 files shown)


lib/katello/engine.rb
```diff
@@ -58,8 +58,12 @@ class Engine < ::Rails::Engine
...
         :agent => {
           :client_queue_format => 'pulp.agent.%s',
-          :broker_url => 'amqp:ssl:localhost:5671',
-          :event_queue_name => 'katello.agent'
+          :event_queue_name => 'katello.agent',
+          :broker_url => 'amqps://localhost:5671',
+          # puppetize these
+          :broker_ssl_cert_file => '/etc/pki/katello/qpid_router_client.crt',
+          :broker_ssl_key_file => '/etc/pki/katello/qpid_router_client.key',
+          :broker_ssl_ca_file => '/etc/pki/katello/certs/katello-default-ca.crt'
```
ehelms, line 66: We do not need to introduce special certificates here for this. We can use the main Foreman client certificates which is a move we are making in general:

https://github.com/theforeman/puppet-foreman/blob/master/templates/settings.yaml.erb#L24-L26

app/lib/katello/qpid/connection.rb
```diff
@@ -49,17 +49,18 @@ def on_message(_delivery, message)
       end
 
       class Receiver < ::Qpid::Proton::MessagingHandler
-        def initialize(url, address, handler)
+        def initialize(url, connection_options, address, handler)
```
ehelms, line 52: The Sender method signature also needs an update -- https://github.com/Katello/katello/pull/9166/files#diff-d5d0f092b1171d0232fd04dc54b5b9f9dc095735857a4f8e7aecf0bd1f35dabbR7

ehelms (approved):
From a testing it on a nightly perspective, this looked good to me. I opened https://github.com/theforeman/puppet-katello/pull/385 to correspond with the needed URL change.

jturel:
[test katello]

jturel:
[test katello]

PR #9167, opened 2021-02-16
Fixes #31887 - Discard systemctl output

adamruzicka:
The command prints enabled/disabled to stdout which ends up in the terminal when
running any rake task which depends on katello:check_ping. Since we do not need
the output, let's redirect it to /dev/null.

### DIFF
--- a/app/models/katello/ping.rb
+++ b/app/models/katello/ping.rb
@@ -9,7 +9,7 @@ module Katello
         SETTINGS[:katello][:use_pulp_2_for_content_type].nil? || (!SETTINGS[:katello][:use_pulp_2_for_content_type][:yum] &&
         !SETTINGS[:katello][:use_pulp_2_for_content_type][:docker] &&
         !SETTINGS[:katello][:use_pulp_2_for_content_type][:file]) ||
-        system('systemctl is-enabled pulpcore-api.service')
+        system('systemctl is-enabled pulpcore-api.service &>/dev/null')
       end
 
       def services(capsule_id = nil)


app/models/katello/ping.rb
```diff
@@ -9,7 +9,7 @@ def pulpcore_enabled # for downstream 6.9, remove in 6.10
         SETTINGS[:katello][:use_pulp_2_for_content_type].nil? || (!SETTINGS[:katello][:use_pulp_2_for_content_type][:yum] &&
         !SETTINGS[:katello][:use_pulp_2_for_content_type][:docker] &&
         !SETTINGS[:katello][:use_pulp_2_for_content_type][:file]) ||
-        system('systemctl is-enabled pulpcore-api.service')
+        system('systemctl is-enabled pulpcore-api.service >/dev/null')
```
jlsherrill, line 12: should it  be `&> /dev/null` to capture stderr too?

adamruzicka: The offending line in output was from stdout, not sure if `systemctl is-enabled` does ever output anything to stderr. I don't mind redirecting both just to be extra sure.

Is `&>` a bash-specific syntax? And if so, should we strive for posix sh compatibility?

jlsherrill: it does still output 'disabled' if its disabled.  And if doesn't exist it outputs a longer error: 'Failed to get unit file state for pulpcore-api.service: No such file or directory'.  

`systemctl is-enabled -q `  seems to nicely hide the enabled/disabled text, but not the error.    I tested &> in plain 'sh' and it seemed to work fine?

jlsherrill (approved):

jlsherrill:
acked, but gotta wait for tests in master to pass

adamruzicka:
[test katello]

adamruzicka:
[test katello]

adamruzicka:
:green_apple:

jlsherrill:
oops, thanks for the ping on this!

PR #9175, opened 2021-02-17
Fixes #31894 - apply errata with REX from content->errata page

ianballou:
To test:

1) Install remote execution on your machine
2) Set the option Content > "Use remote execution by default" to true
3) Go to the WebUI > Content > Errata > select errata > apply errata > select hosts > Next > Apply Errata > Confirm
4) Notice that katello-agent is still trying to be used
5) Apply this PR
6) Try step 3 again

I wasn't able to get remote execution working properly on my development environment, but I did see the job run with the proper parameters for hosts and errata.

I don't fix bugs in the JavaScript often, so let me know if I'm doing anything weird.


chris1984 (approved):
ACK, works good thanks @ianballou 

I could not get the REX working either but was able to confirm I saw the same thing as you, so I can say this should work once it can be tested end to end on a production install. 

Thanks for including tests with the JS too.

PR #9178, opened 2021-02-18
Fixes #31910 - Hide k-agent dep warnings if setting is disabled.

chris1984:
This PR is a draft to show/hide the deprecation warnings for katello-agent if the setting is enabled in the installer. By default, katello-agent will be disabled so we hide the warnings in the UI, if the user enabled it then we will show them.

Client with katello-agent not installed:

https://nimb.ws/0qu6iK

Client with katello-agent installed:

https://nimb.ws/BOSiEk

Bulk update packages with katello-agent status set to true:

https://nimb.ws/OvgA4k

Bulk update packages with katello-agent status set to false:

https://nimb.ws/43e51x

Right now, I am keeping the PR as a draft as Eric Helms is working on a Smart Proxy feature which will set/configure the status. When that is made, I will update this PR to pull the state from that.

I am opening this in the interim for feedback etc.

To test toggle: 

https://github.com/Katello/katello/compare/master...chris1984:hide-agentwarn?expand=1#diff-7dc78088a8bc944d190599ecf572ea01ab9b85fc91415ddcadf17dd5e9b1638eR262 

To true/false and check the following pages:

- content-hosts-bulk-errata
- content-hosts-bulk-packages
- content-host-errata
- content-host-packages-actions
- content-host-packages-applicable
- content-host-packages-installed

### DIFF
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-errata-modal.html
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-errata-modal.html
@@ -27,7 +27,14 @@
       </div>
     </div>
 
-    <div data-extend-template="common/views/katello-agent-notice.html"></div>
+    <div ng-show="katelloAgentPresent" data-extend-template="common/views/katello-agent-notice.html"></div>
+    <div bst-feature-flag="remote_actions">
+      <p bst-alert="warning" ng-hide="hostToolingEnabled">
+        <span translate>
+          Performing host package actions is disabled because Katello is not configured for Remote Execution or Katello Agent.
+          </span>
+      </p>
+    </div>
 
     <div ng-show="showTable()" data-extend-template="layouts/partials/table.html">
       <div data-block="list-actions">
@@ -39,19 +46,19 @@
 
         <span uib-dropdown class="btn-group">
           <button class="btn btn-primary" type="button"
-                  ng-disabled="table.numSelected === 0 || table.working || table.numSelected === 0"
+                  ng-disabled="table.numSelected === 0 || table.working || table.numSelected === 0 || !hostToolingEnabled"
                   ng-click="showConfirm = true;">
             <span translate>Install Selected</span>
           </button>
           <button uib-dropdown-toggle class="btn btn-primary" type="button"
-                  ng-hide="!remoteExecutionPresent"
+                  ng-show="remoteExecutionPresent"
                   ng-disabled="table.numSelected === 0 || table.working || table.numSelected === 0"
                   type="button" id="use-remote-execution">
             <span class="caret"></span>
           </button>
 
           <ul uib-dropdown-menu class="dropdown-menu-right" role="menu" aria-labelledby="use-remote-execution">
-            <li role="presentation"><a ng-click="installErrataViaKatelloAgent()" role="menuitem" translate>via Katello agent</a></li>
+            <li ng-show="katelloAgentPresent" role="presentation"><a ng-click="installErrataViaKatelloAgent()" role="menuitem" translate>via Katello agent</a></li>
             <li role="presentation"><a ng-click="installErrataViaRemoteExecution(false)" role="menuitem" translate>via remote execution</a></li>
             <li role="presentation"><a ng-click="installErrataViaRemoteExecution(true)" role="menuitem" translate>via remote execution - customize first</a></li>
           </ul>
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/content/content-host-packages-applicable.controller.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/content/content-host-packages-applicable.controller.js
@@ -9,14 +9,19 @@
  * @requires HostPackage
  * @requires translate
  * @requires Nutupane
+ * @requires BastionConfig
  *
  * @description
  *   Provides the functionality for the content host packages list and actions.
  */
 angular.module('Bastion.content-hosts').controller('ContentHostPackagesApplicableController',
-    ['$scope', '$timeout', '$window', 'Package', 'HostPackage', 'translate', 'Nutupane',
-    function ($scope, $timeout, $window, Package, HostPackage, translate, Nutupane) {
+    ['$scope', '$timeout', '$window', 'Package', 'HostPackage', 'translate', 'Nutupane', 'BastionConfig',
+    function ($scope, $timeout, $window, Package, HostPackage, translate, Nutupane, BastionConfig) {
         var packagesNutupane, openEventInfo;
+
+        $scope.katelloAgentPresent = BastionConfig.katelloAgentPresent;
+        $scope.remoteExecutionPresent = BastionConfig.remoteExecutionPresent;
+        $scope.hostToolingEnabled = BastionConfig.hostToolingEnabled;
 
         $scope.getSelectedPackages = function () {
             var selected = $scope.table.getSelected();
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/content/views/content-host-errata.html
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/content/views/content-host-errata.html
@@ -5,7 +5,7 @@
   <h3 translate ng-show="selectedErrataOption === 'current'">Installable Errata</h3>
 </div>
 
-<div data-extend-template="common/views/katello-agent-notice.html"></div>
+<div ng-show="katelloAgentPresent" data-extend-template="common/views/katello-agent-notice.html"></div>
 
 <div ng-hide="host.hasContent()">
   <div data-extend-template="common/views/registration.html"></div>
@@ -24,10 +24,10 @@
   </div>
 
   <div bst-feature-flag="remote_actions">
-    <p bst-alert="info" ng-hide="host.content_facet_attributes.katello_agent_installed">
+    <p bst-alert="warning" ng-hide="hostToolingEnabled">
       <span translate>
-        If you are not using Remote Execution to manage content on this host, ensure that it has the katello-agent package installed.
-      </span>
+        Performing host package actions is disabled because Katello is not configured for Remote Execution or Katello Agent.
+        </span>
     </p>
   </div>
 
@@ -69,7 +69,7 @@
             <button class="btn btn-default" type="button"
                     translate
                     ng-hide="denied('edit_hosts', host)"
-                    ng-disabled="table.getSelected().length == 0"
+                    ng-disabled="table.getSelected().length == 0 || !hostToolingEnabled"
                     ng-click="openModal()">
               Apply Selected
             </button>
@@ -80,7 +80,7 @@
               <span class="caret"></span>
             </button>
             <ul uib-dropdown-menu role="menu" aria-labelledby="use-remote-execution">
-              <li role="presentation"><a ng-click="performViaKatelloAgent()" role="menuitem" tabindex="-1" href="#" translate>via Katello agent</a></li>
+              <li ng-show="katelloAgentPresent" role="presentation"><a ng-click="performViaKatelloAgent()" role="menuitem" tabindex="-1" href="#" translate>via Katello agent</a></li>
               <li role="presentation"><a ng-click="performViaRemoteExecution(false)" role="menuitem" tabindex="-1" href="#" translate>via remote execution</a></li>
               <li role="presentation"><a ng-click="performViaRemoteExecution(true)" role="menuitem" tabindex="-1" href="#" translate>via remote execution - customize first</a></li>
             </ul>
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/content/views/content-host-packages-installed.html
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/content/views/content-host-packages-installed.html
@@ -2,13 +2,13 @@
 
 <h3 translate>Installed Packages</h3>
 
-<div data-extend-template="common/views/katello-agent-notice.html"></div>
+<div ng-show="katelloAgentPresent" data-extend-template="common/views/katello-agent-notice.html"></div>
 
 <section ng-hide="denied('edit_hosts', host)" bst-feature-flag="remote_actions">
-  <p bst-alert="info" ng-hide="host.content_facet_attributes.katello_agent_installed">
+  <p bst-alert="warning" ng-hide="hostToolingEnabled">
     <span translate>
-      If you are not using Remote Execution to manage content on this host, ensure that it has the katello-agent package installed.
-    </span>
+      Performing host package actions is disabled because Katello is not configured for Remote Execution or Katello Agent.
+      </span>
   </p>
 </section>
 
@@ -24,7 +24,7 @@
   <div data-block="list-actions" bst-feature-flag="remote_actions">
     <button class="btn btn-default" type="button"
             ng-hide="denied('edit_hosts', host)"
-            ng-disabled="working || table.numSelected === 0"
+            ng-disabled="working || table.numSelected === 0 || !hostToolingEnabled"
             ng-click="removeSelectedPackages()">
       <span translate>Remove Selected</span>
     </button>
--- a/lib/katello/engine.rb
+++ b/lib/katello/engine.rb
@@ -253,6 +253,14 @@ module Katello
     (RemoteExecutionFeature rescue false) ? true : false
   end
 
+  def self.with_katello_agent?
+    !!SETTINGS.dig(:katello, :agent, :enabled)
+  end
+
+  def self.remote_execution_by_default?
+    self.with_katello_agent? ? Setting['remote_execution_by_default'] : true
+  end
+
   def self.with_ansible?
     (ForemanAnsible rescue false) ? true : false
   end
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-errata-modal.controller.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/content-hosts-bulk-errata-modal.controller.js
@@ -57,7 +57,9 @@ angular.module('Bastion.content-hosts').controller('ContentHostsBulkErrataModalC
         $scope.initialLoad = true;
         $scope.remoteExecutionPresent = BastionConfig.remoteExecutionPresent;
         $scope.remoteExecutionByDefault = BastionConfig.remoteExecutionByDefault;
+        $scope.katelloAgentPresent = BastionConfig.katelloAgentPresent;
         $scope.allHostsSelected = hostIds.allResultsSelected;
+        $scope.hostToolingEnabled = BastionConfig.hostToolingEnabled;
 
         $scope.errataActionFormValues = {
             authenticityToken: $window.AUTH_TOKEN.replace(/&quot;/g, '')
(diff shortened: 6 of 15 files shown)


engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/content/views/content-host-errata.html
```diff
@@ -5,7 +5,7 @@ <h3 translate ng-hide="selectedErrataOption === 'current'">Applicable Errata</h3
   <h3 translate ng-show="selectedErrataOption === 'current'">Installable Errata</h3>
 </div>
 
-<div data-extend-template="common/views/katello-agent-notice.html"></div>
+<div ng-show="host.content_facet_attributes.katello_agent_installed" data-extend-template="common/views/katello-agent-notice.html"></div>
```
jeremylenz, line 8: Here are the results I got, can you double check if this is what's expected?  Especially the bold ones (They may be correct but wanted to lay it out here just to be sure.)

### katello-agent status vs 4 banners
| - | `with_katello_agent` method return value | katello-agent installed? | banner status
| --- | --- | --- | --- |
| content-host-details-errata | true | yes | Present
content-host-details-package-actions | true | yes | Present
bulk-actions-errata modal | true | yes | Present
bulk-actions-package modal | true | yes | Present
| content-host-details-errata | true | no | __Present__
content-host-details-package-actions | true | no | __Present__
bulk-actions-errata modal | true | no | Not present
bulk-actions-package modal | true | no | Not present
| content-host-details-errata | false | yes | __Present__
content-host-details-package-actions | false | yes | __Present__
bulk-actions-errata modal | false | yes | Not present
bulk-actions-package modal | false | yes | Not present
| content-host-details-errata | false | no | Not present
content-host-details-package-actions | false | no | Not present
bulk-actions-errata modal | false | no | Not present
bulk-actions-package modal | false | no | Not present

lib/katello/engine.rb
```diff
@@ -258,6 +258,10 @@ def self.with_remote_execution?
     (RemoteExecutionFeature rescue false) ? true : false
   end
 
+  def self.with_katello_agent?
+    return true
```
jeremylenz, line 262: I'm assuming this is the part that will be updated later to pull the setting?

jeremylenz (commented):
Thanks @chris1984 

Banners look fine; see a couple thoughts below.

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/content/views/content-host-packages-installed.html
```diff
@@ -2,7 +2,7 @@
 
 <h3 translate>Installed Packages</h3>
 
-<div data-extend-template="common/views/katello-agent-notice.html"></div>
+<div ng-show="host.content_facet_attributes.katello_agent_installed" data-extend-template="common/views/katello-agent-notice.html"></div>
```
jturel, line 5: Why make some of these display based on presence of katello-agent package vs the enablement of the agent functionality on the server itself?

jeremylenz: right, I guess that's the same thing I am asking with the table above!

chris1984: I will update it and change it, I saw that we could display stuff based on that, rather than adding new stuff into the controller I figured we could use that in the view. I will switch it the way we discussed.

lib/katello/engine.rb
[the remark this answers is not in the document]
chris1984: Updated

chris1984: Yeah once the feature from Eric is made, this will be changed to something better than just `true` or `false` lol. I marked the PR as blocked by another so it does not get merged on accident.

chris1984:
@jeremylenz updated, let me know how that looks.

jturel:
To link it up, this is the behavior that Eric is working on that we'll be able to use here. The direction has changed from a smart proxy plugin to a simple config value. https://github.com/theforeman/puppet-katello/pull/387

lib/katello/engine.rb
jeremylenz: I never look at the tags.  I would change it to a draft PR or add [WIP] to the beginning of the title. :stuck_out_tongue:

jeremylenz (commented):
Changes LGTM, thanks @chris1984!

Looks like there are a few Angular tests that need updating.

Will ACK once that's resolved and theforeman/puppet-katello#387 is in.

lib/katello/engine.rb
chris1984: Changed to [WIP] since Github won't allow me to change it to a draft

ehelms:
Let me know if you want to base this on an explicit configuration flag or implicit absence of configuration. Right now https://github.com/theforeman/puppet-katello/pull/387/files#diff-510884a7414666a0d988133c5f869510210d883aa1580d1248e74f1d16ca7d91R26 is just not configuring that section of the file but if you have default values I could see that being a hard case to judge what to do on. If you prefer I can go the route of having:

```
:agent:
    :enabled: <%= @enable_katello_agent %>
    :broker_url: <%= @agent_broker_url %>
    :event_queue_name: <%= @agent_event_queue_name %>
```

chris1984:
> Let me know if you want to base this on an explicit configuration flag or implicit absence of configuration. Right now https://github.com/theforeman/puppet-katello/pull/387/files#diff-510884a7414666a0d988133c5f869510210d883aa1580d1248e74f1d16ca7d91R26 is just not configuring that section of the file but if you have default values I could see that being a hard case to judge what to do on. If you prefer I can go the route of having:
> 
> ```
> :agent:
>     :enabled: <%= @enable_katello_agent %>
>     :broker_url: <%= @agent_broker_url %>
>     :event_queue_name: <%= @agent_event_queue_name %>
> ```

The `:enabled: <%= @enable_katello_agent %>` looks good, that would be the easiest way to go with how we have the method setup.

ehelms:
> The `:enabled: <%= @enable_katello_agent %>` looks good, that would be the easiest way to go with how we have the method setup.

Updated my PR with this. Right now in the installer the support for this would look like:

Enabling:
```
foreman-installer --katello-enable-katello-agent true --foreman-proxy-content-enable-katello-agent true
```

Disabling on an existing setup:
```
foreman-installer --katello-enable-katello-agent false --foreman-proxy-content-enable-katello-agent false
```

On a proxy:
Enabling
```
foreman-installer --foreman-proxy-content-enable-katello-agent true
```

Disabling on an existing setup:
```
foreman-installer --foreman-proxy-content-enable-katello-agent false
```

chris1984:
@jeremylenz can I get an ack?

jturel:
@chris1984 Since we're going with option 1 here: https://community.theforeman.org/t/katello-package-actions-in-an-agent-less-world/22777 I'm curious if you're open to adding some additional work on this PR: _If the Agent is disabled, then in the UI should present only one "Perform" option (REX) instead of the dropdown._

Should be a small change and would save some churn with another card and more testing esp since this PR is already touching the UI in this area. I'm open to a second card, but wanted run it by you.

ehelms:
> The `:enabled: <%= @enable_katello_agent %>` looks good, that would be the easiest way to go with how we have the method setup.

I am mostly ready over on the installer side, and will be implementing this pattern. I think you'll want to add reading of that setting in this PR now and set it to false by default (I assume).

lib/katello/engine.rb
```diff
@@ -250,6 +250,10 @@ def self.with_remote_execution?
     (RemoteExecutionFeature rescue false) ? true : false
   end
 
+  def self.with_katello_agent?
+    SETTINGS[:katello][:agent][:enabled]
```
ehelms, line 254: You may want to add a default here https://github.com/Katello/katello/pull/9178/files#diff-7dc78088a8bc944d190599ecf572ea01ab9b85fc91415ddcadf17dd5e9b1638eR59

You might consider using dig here as well to future proof it if this setting got removed or wasn't present:

```
SETTINGS.dig(:katello, :agent, :enabled)
```

jturel:
Something else came to mind that could save us time by addressing in this PR now. Lots of text below but the actual change is not big imo.

1. Remove all 'remoteExecutionPresent / with_remote_execution?' checks. This can be done safely because Katello is always installed with REX now. Removing the REX rpm would remove the rubygem-katello RPM so I don't think it's possible to have Katello without REX. Further, I don't see a way to disable REX via a setting or anything like that.

2. Consider the 'remote_execution_by_default' setting. When katello-agent is fully removed that setting can be removed altogether. It still has a purpose today and we may want to tweak that by updating the default value to true. However, I learned that updating the default value to true would also affect upgrades to 4.1 and that wouldn't be desired. A 4.0 user might be using katello-agent and want to keep using it in 4.1 and changing the default value of that setting will break their workflows via UI at least. You could add a new method like you did for `with_katello_agent?` which would say `if we have katello agent, then respect the value of remote_execution_by_default otherwise true`. That likely implies updating the few places that reference the setting to look at your new method instead.

Please double-check my thinking here and let me know if you're up for it on this PR!

ehelms:
> * Remove all 'remoteExecutionPresent / with_remote_execution?' checks. This can be done safely because Katello is always installed with REX now. Removing the REX rpm would remove the rubygem-katello RPM so I don't think it's possible to have Katello without REX. Further, I don't see a way to disable REX via a setting or anything like that.
> *

A question. Do we want to tie Katello to REX as a hard requirement? The more we couple Katello to other plugins as a *hard requirement* the hard it becomes to release Katello. With the move away from katello-agent, and to a centralized use of REX for remote actions I feel like we should be aiming for *if* a user has installed and configured REX they can then perform package actions.



> 2\. Consider the 'remote_execution_by_default' setting. When katello-agent is fully removed that setting can be removed altogether. It still has a purpose today and we may want to tweak that by updating the default value to true. However, I learned that updating the default value to true would also affect upgrades to 4.1 and that wouldn't be desired. A 4.0 user might be using katello-agent and want to keep using it in 4.1 and changing the default value of that setting will break their workflows via UI at least. You could add a new method like you did for `with_katello_agent?` which would say `if we have katello agent, then respect the value of remote_execution_by_default otherwise true`. That likely implies updating the few places that reference the setting to look at your new method instead.

I think I read this as swap the defaults which I'd agree with if thats easy. If the two changes can be split that would work too. I can't merge the installer changes until we know that Katello can handle katello-agent not being present both from a runtime and 'makes sense to the user' perspective. We also have only two weeks to stabilization to get both sets of changes in.

jturel:
> > * Remove all 'remoteExecutionPresent / with_remote_execution?' checks. This can be done safely because Katello is always installed with REX now. Removing the REX rpm would remove the rubygem-katello RPM so I don't think it's possible to have Katello without REX. Further, I don't see a way to disable REX via a setting or anything like that.
> 
> A question. Do we want to tie Katello to REX as a hard requirement? The more we couple Katello to other plugins as a _hard requirement_ the hard it becomes to release Katello. With the move away from katello-agent, and to a centralized use of REX for remote actions I feel like we should be aiming for _if_ a user has installed and configured REX they can then perform package actions.
>
It's already a hard requirement (unless you mean something else) => https://github.com/theforeman/foreman-packaging/blob/dd071f1ee8e5bedbad98ad7c982e540fd8a3f9ea/packages/katello/rubygem-katello/rubygem-katello.spec#L38

Having REX optional in a katello-agent disabled environment means that a user could install Katello and come to find they can't perform remote actions because REX isn't there. Seems like bad UX even if it's well documented that it must be installed separately because it gives the more steps for the user and is not following the principle of least surprise. Is there overhead in having REX installed? IMO just install the thing always & everywhere. Actually, why isn't it just part of Foreman itself at this point ...
 
> > 2. Consider the 'remote_execution_by_default' setting. When katello-agent is fully removed that setting can be removed altogether. It still has a purpose today and we may want to tweak that by updating the default value to true. However, I learned that updating the default value to true would also affect upgrades to 4.1 and that wouldn't be desired. A 4.0 user might be using katello-agent and want to keep using it in 4.1 and changing the default value of that setting will break their workflows via UI at least. You could add a new method like you did for `with_katello_agent?` which would say `if we have katello agent, then respect the value of remote_execution_by_default otherwise true`. That likely implies updating the few places that reference the setting to look at your new method instead.
> 
> I think I read this as swap the defaults which I'd agree with if thats easy. If the two changes can be split that would work too. I can't merge the installer changes until we know that Katello can handle katello-agent not being present both from a runtime and 'makes sense to the user' perspective. We also have only two weeks to stabilization to get both sets of changes in.

Yep, it's just flipping the default based on the enabled state of katello-agent. We can't change the default value outright in the settings file (app/models/katello/setting.rb) because it would flip during the upgrade process even for users who wish to continue using katello-agent. The wrapper method described above allows us to put some logic around that setting to effectively flip the default.

jturel:
Thanks for the input @ehelms -  we'll leave the REX checks in place - that makes the scope of this PR smaller anyway!

chris1984:
@jturel how does this look?

ehelms:
As I understand the way this change is coded up, we can merge it prior to merging the installer changes. This will just assume katello-agent disabled by default without the changes to set the enabled flag to true? Trying to see how much we can avoid coordinated merges  (without breaking things).

chris1984:
@ehelms that is correct, now that it is reading if the setting is present. For now, it will just default to false since the installer change is not there. Does it look fine for all your feedback that you raised on it?

ehelms:
> Does it look fine for all your feedback that you raised on it?

Yes, thank you.

chris1984:
@jturel All of @ehelms comments have been addressed, can I get a retest/review again?

chris1984:
@jturel updated.

Here are my testing results:

## rex enabled and katello agent disabled:

* Bulk Errata - Agent dep warning removed, dropdown enabled

* Bulk Packages - Agent dep warning removed, dropdown enabled

* Packages Actions - Agent dep warning removed, dropdown enabled

* Packages Applicable - Agent dep warning removed, dropdown enabled

* Packages Installed - Agent dep warning removed, no drop down

* Errata - Agent dep warning removed, dropdown enabled

## rex disabled and katello agent enabled:

* Bulk Errata - Agent dep warning, dropdown removed

* Bulk Packages - Agent dep warning, dropdown removed

* Packages Actions - Agent dep warning, dropdown removed

* Packages Applicable - Agent dep warning, dropdown removed

* Packages Installed - Agent dep warning, dropdown removed

* Errata - Agent dep warning, dropdown removed

## Katello agent and rex both enabled:

* Bulk Errata - Agent dep warning, dropdown enabled

* Bulk Packages - Agent dep warning, dropdown enabled

* Packages Actions - Agent dep warning, dropdown enabled

* Packages Applicable - Agent dep warning, dropdown enabled

* Packages Installed - Agent dep warning, no drop down

* Errata - Agent dep warning, dropdown enabled

## Katello agent and rex both disabled:

* Bulk Errata - Warning showed and buttons grayed out

* Bulk Packages - Warning showed and buttons grayed out

* Packages Actions - Warning showed and buttons grayed out

* Packages Applicable - Warning showed and buttons grayed out

* Packages Installed - Warning showed and buttons grayed out

* Errata - Warning showed and buttons grayed out

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/content/content-host-packages-applicable.controller.js
```diff
@@ -26,6 +30,12 @@ angular.module('Bastion.content-hosts').controller('ContentHostPackagesApplicabl
             return selected;
         };
 
+        if ($scope.katelloAgentPresent === false && $scope.remoteExecutionPresent === false) {
```
jturel, line 33: It looks like this logic is duplicated a few times. I think it makes sense to put it as one of the attributes on the BastionConfig!

To check for `false` in this way isn't very idiomatic unless you explicitly need to check for false rather than something falsey, like undefined. It's a nitpick. I'd also write it as `hostToolingEnabled` to keep the logic framed more positively which usually keeps things easier to follow. so the pseudocode now becomes:

`hostToolingEnabled = (katelloAgentPresent || remoteExecutionPresent) ? true : false`

lib/katello/engine.rb
```diff
@@ -253,6 +253,14 @@ def self.with_remote_execution?
     (RemoteExecutionFeature rescue false) ? true : false
   end
 
+  def self.with_katello_agent?
+    !!SETTINGS.dig(:katello, :agent, :enabled)
+  end
+
+  def self.remote_execution_by_default?
+    self.with_katello_agent? ? Setting['remote_execution_by_default'] : true
```
jturel, line 261: Stunning!

chris1984:
not sure why me pushing closed it

chris1984:
@jturel updated.

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-errata-modal.html
```diff
@@ -27,7 +27,21 @@ <h4 data-block="modal-header">Content Host Errata Management</h4>
       </div>
     </div>
 
-    <div data-extend-template="common/views/katello-agent-notice.html"></div>
+    <div ng-show="katelloAgentPresent" data-extend-template="common/views/katello-agent-notice.html"></div>
+
+    <div bst-feature-flag="remote_actions">
+      <p bst-alert="info" ng-hide="katelloAgentPresent || !hostToolingEnabled">
+        <span translate>
+          If you are not using Remote Execution to manage content on this host, ensure that it has the katello-agent package installed.
```
jturel, line 35: I kinda want to remove this particular message altogether because it is effectively pushing someone to use the agent. Watcha think?

chris1984: I am for removing it, I will take it out and push up the changes.

chris1984: @jturel removed and updated

jturel (changes requested):
- When REX=true and AGENT=false I still see 'install via Katello Agent' on the Bulk update packages modal
- on Host -> Packages -> Actions ''Update all packages" is enabled when REX=true and AGENT=false but the page only works with AGENT (please confirm). Suggest disabling button in this case
- Host -> Packages -> Applicable  -> Update All Packages is enabled with REX=true AGENT=false but the page appears to work with AGENT only.
- Let's hide whether Katello Agent is installed on host details with AGENT=false

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/bulk/views/content-hosts-bulk-errata-modal.html
```diff
@@ -27,7 +27,14 @@ <h4 data-block="modal-header">Content Host Errata Management</h4>
       </div>
     </div>
 
-    <div data-extend-template="common/views/katello-agent-notice.html"></div>
+    <div ng-show="katelloAgentPresent" data-extend-template="common/views/katello-agent-notice.html"></div>
+    <div bst-feature-flag="remote_actions">
+      <p bst-alert="warning" ng-hide="hostToolingEnabled">
+        <span translate>
+          Remote execution is not present and katello-agent is not enabled. Performing host actions is disabled.
```
jturel, line 34: This is tricky to word properly because we need to be clear that it's the server that's configured this way, not the client. Something like:

```
Performing host package actions is disabled because Katello is not configured for Remote Execution or Katello Agent
```

chris1984: @jturel addressed all feedback and put in a second commit for you :)

jturel (approved):
APJ

The UI experience looks consistent across the board with various permutations of REX or AGENT enablement!

chris1984:
[test katello]

ehelms:
Odd failure given I don't see any React changes in this PR. Katello nightly has gone green finally from other changes. So I think we are safe to brea... I mean merge this and the installer changes :)

jturel:
Agree!

jeremylenz:
fyi, the React failure is fixed by https://github.com/theforeman/foreman/pull/8454

PR #9180, opened 2021-02-18
Fixes #31919 - fix migrated content import to update all units

jlsherrill:
in some cases we have multiple katello entries that match up to a single
pulp3 entry.  In these cases we can just update all units with the same
pulp2 id


jlsherrill:
you'll need to
1) sync a RH repo
2) promote it in a CV
3) do the migration (rake katello:pulp3_migration)
4) run the swtichover rake task  (rake katello:pulp3_content_switchover)

you should hit an error on 4).  After applying the patch, rerun 3) and then 4)

ianballou:
Works for me! Tested on production 3.18 (with the latest dynflow skip patched into the switchover rake task).

PR #9182, opened 2021-02-18
Refs #31788 - Can migrate CVV Export History correctly

parthaa:
Due to a botched previous commit to this issue if one tried to migrate
with existing export histories you'd hit the following issue
`undefined local variable or method `cvve' for

This commit fixes that

### DIFF
--- a/app/models/katello/content_view_version_export_history.rb
+++ b/app/models/katello/content_view_version_export_history.rb
@@ -8,13 +8,14 @@ module Katello
 
     belongs_to :content_view_version, :class_name => "Katello::ContentViewVersion", :inverse_of => :export_histories
     validates_lengths_from_database
+
     validates :content_view_version_id, :presence => true
     validates :destination_server, :uniqueness => { :scope => [:content_view_version_id, :destination_server, :path] }
-    validates :export_type, :inclusion => { :in => EXPORT_TYPES,
-                                            :allow_blank => false,
-                                            :message => _("Invalid export_type from one of the following: %s" % EXPORT_TYPES.join(', '))
-                                          }
-
+    validates :export_type, :presence => true,
+              :inclusion => { :in => EXPORT_TYPES,
+                              :allow_blank => false,
+                              :message => _("must be one of the following: %s" % EXPORT_TYPES.join(', '))
+                            }
     validates :metadata, :presence => true
     serialize :metadata, Hash
 
@@ -39,7 +40,7 @@ module Katello
     end
 
     def export_type_from_metadata
-      cvve.metadata[:incremental] ? INCREMENTAL : COMPLETE
+      metadata[:incremental] ? INCREMENTAL : COMPLETE
     end
 
     def set_export_type
--- a/db/migrate/20210218214048_change_default_content_view_version_export_history.rb
+++ b/db/migrate/20210218214048_change_default_content_view_version_export_history.rb
@@ -1,0 +1,5 @@
+class ChangeDefaultContentViewVersionExportHistory < ActiveRecord::Migration[6.0]
+  def change
+    change_column_default(:katello_content_view_version_export_histories, :export_type, nil)
+  end
+end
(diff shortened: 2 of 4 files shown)


parthaa:
- Publish a cv
- Get its cvv_id of the published version.
- Goto ```rails console```

Prior to this PR if you did (assuming your cvv id is 100)
```
>>  ::Katello::ContentViewVersionExportHistory.create!(content_view_version_id: 100, path: "/tmp2",  metadata:{export_type: 'incremental'})
=> #<Katello::ContentViewVersionExportHistory:0x000000001b47a908
 id: 33,
 created_at: Thu, 18 Feb 2021 21:29:18 UTC +00:00,
 updated_at: Thu, 18 Feb 2021 21:29:18 UTC +00:00,
 destination_server: nil,
 path: "/tmp2",
 content_view_version_id: 100,
 metadata: {:export_type=>"incremental"},
 export_type: "complete">
```

You'd notice that export_type is set to complete even though your metadata said it should be incremental.

After this PR + running `rake db:migrate`

```
>>  ::Katello::ContentViewVersionExportHistory.create!(content_view_version_id: 100, path: "/tmp2",  metadata:{export_type: 'incremental'})
=> #<Katello::ContentViewVersionExportHistory:0x000000001b47a808
 id: 34,
 created_at: Thu, 18 Feb 2021 21:29:18 UTC +00:00,
 updated_at: Thu, 18 Feb 2021 21:29:18 UTC +00:00,
 destination_server: nil,
 path: "/tmp2",
 content_view_version_id: 100,
 metadata: {:export_type=>"incremental"},
 export_type: "incremental">
```
You'd notice that export_type is set to correctly set to incremental.

app/models/katello/content_view_version_export_history.rb
```diff
@@ -8,17 +8,18 @@ class ContentViewVersionExportHistory < Katello::Model
...
     validates :destination_server, :uniqueness => { :scope => [:content_view_version_id, :destination_server, :path] }
-    validates :export_type, :inclusion => { :in => EXPORT_TYPES,
-                                            :allow_blank => false,
-                                            :message => _("Invalid export_type from one of the following: %s" % EXPORT_TYPES.join(', '))
-                                          }
-
+    validates :export_type, :presence => true,
+              :inclusion => { :in => EXPORT_TYPES,
+                              :allow_blank => false,
+                              :message => _("Invalid export_type from one of the following: %s" % EXPORT_TYPES.join(', '))
```
jeremylenz, line 17: While we're here, this message could be clearer:
```suggestion
                              :message => _("must be one of the following: %s" % EXPORT_TYPES.join(', '))
```

app/models/katello/content_view_version_export_history.rb
```diff
@@ -39,11 +40,11 @@ def self.latest(content_view, destination_server: nil)
     end
 
     def export_type_from_metadata
-      cvve.metadata[:incremental] ? INCREMENTAL : COMPLETE
+      metadata[:incremental] ? INCREMENTAL : COMPLETE
     end
 
     def set_export_type
-      self.export_type = export_type_from_metadata
+      self.export_type ||= export_type_from_metadata
```
jeremylenz, line 47: This still seems to me like trying to manage multiple sources of truth.  The export type is stored inside the `metadata` hash, and is also stored in the model's `export_type`.  What if `export_type` were not stored in the db as a separate column, and instead we could just add an `export_type` method to extract the value from `metadata`?  Then you wouldn't have to keep the two values in sync.  Would that cause problems anywhere?

jeremylenz (commented):
Thanks @parthaa 

Confirmed that the migration change works.  See some other thoughts below.. I'm not sure how strongly I feel but wanted to bring it up at least to discuss it

app/models/katello/content_view_version_export_history.rb
parthaa: This was discussed before but the gist is that we want to be able to query on export_type (`hammer content-export list --type=incremental`.) So it needs to be a column so that we can apply the scoped search magic. That guy ensures that if export_type is not set he pulls from metadata. The actual actions that create this History provide the export_type anyway so this is really not a big deal. These records are not manipulated after in any case.

jeremylenz: @parthaa Thanks for reminding me of the explanation, makes sense :)

jeremylenz (approved):
APJ :+1:

parthaa:
[test katello]

ehelms:
[test katello]

ehelms:
Apologies, I am playing with some new job configuration an accidentally left the PR trigger on.

PR #9183, opened 2021-02-19
Fixes #31923 - allow setting no proxy too

ares:
the blank value caused problem when user tries to reset this to the default (nil), because no proxy can be found


ares:
Ok I give up, the problem is that the value coming from the form is not empty string, it's the actual `no global default` string. Therefore the detection fails. `:include_blank` no longer sets empty string as a value for this select <option>. Settings is now in react and I didn't find the proper place to fix it.

PR #9190, opened 2021-02-22
Fixes #31760 - Add Katello versions to unversioned warnings

jeremylenz:
* `app/models/katello/host/info_provider.rb` had two parameters marked as deprecated with a code comment (possibly since 2015).  These values don't seem to be used anywhere in Katello, so I thought it's probably safe to remove them.
* For other unversioned deprecation warnings, added Katello 4.1 as a starting point.  Need feedback on whether any should be moved to 4.0 or later versions.
* I'm not sure if our automated deprecations tool picks up both `apipie` params and `::Foreman::Deprecation.api_deprecation_warning`s.  If it's not both, Redmine issues should be created for any deprecation that won't be called out by our tooling.

### DIFF
--- a/app/controllers/katello/api/v2/content_exports_controller.rb
+++ b/app/controllers/katello/api/v2/content_exports_controller.rb
@@ -29,7 +29,7 @@ module Katello
 
     api :GET, "/content_exports/api_status", N_("true if the export api is pulp3 ready and usable. This API is intended for use by hammer-cli only.")
     def api_status
-      ::Foreman::Deprecation.api_deprecation_warning("/content_exports/api_status is being deprecated and will be removed in a future version of Katello.")
+      ::Foreman::Deprecation.api_deprecation_warning("/content_exports/api_status is being deprecated and will be removed in Katello 4.1.")
       render json: { api_usable: SmartProxy.pulp_primary.pulp3_repository_type_support?(Katello::Repository::YUM_TYPE) }, status: :ok
     end
 
--- a/app/controllers/katello/api/v2/content_view_puppet_modules_controller.rb
+++ b/app/controllers/katello/api/v2/content_view_puppet_modules_controller.rb
@@ -9,7 +9,7 @@ module Katello
     param :content_view_id, :number, :desc => N_("content view identifier"), :required => true
     param :name, String, :desc => N_("name of the puppet module")
     param :author, String, :desc => N_("author of the puppet module")
-    param :uuid, String, :desc => N_("uuid of the puppet module"), :deprecated => true
+    param :uuid, String, :desc => N_("uuid of the puppet module.  Will be removed in Katello 4.1."), :deprecated => true
     param_group :search, ::Katello::Api::V2::ApiController
     add_scoped_search_description_for(ContentViewPuppetModule)
     def index
@@ -22,7 +22,7 @@ module Katello
     param :name, String, :desc => N_("name of the puppet module")
     param :author, String, :desc => N_("author of the puppet module")
     param :id, String, :desc => N_("the id of the puppet module to associate")
-    param :uuid, String, :desc => N_("the uuid of the puppet module to associate"), :deprecated => true
+    param :uuid, String, :desc => N_("the uuid of the puppet module to associate. Will be removed in Katello 4.1."), :deprecated => true
     def create
       params[:content_view_puppet_module][:uuid] ||= PuppetModule.find(params[:id]).try(:pulp_id) if params[:id]
       respond resource: ContentViewPuppetModule.create!(puppet_module_params.merge(content_view: @view))
@@ -41,7 +41,7 @@ module Katello
     param :id, :number, :desc => N_("puppet module ID"), :required => true
     param :name, String, :desc => N_("name of the puppet module")
     param :author, String, :desc => N_("author of the puppet module")
-    param :uuid, String, :desc => N_("the uuid of the puppet module to associate"), :deprecated => true
+    param :uuid, String, :desc => N_("the uuid of the puppet module to associate. Will be removed in Katello 4.1."), :deprecated => true
     def update
       @puppet_module.update!(puppet_module_params)
       respond :resource => @puppet_module
--- a/app/controllers/katello/api/v2/content_view_puppet_modules_controller.rb
+++ b/app/controllers/katello/api/v2/content_view_puppet_modules_controller.rb
@@ -9,7 +9,7 @@ module Katello
     param :content_view_id, :number, :desc => N_("content view identifier"), :required => true
     param :name, String, :desc => N_("name of the puppet module")
     param :author, String, :desc => N_("author of the puppet module")
-    param :uuid, String, :desc => N_("uuid of the puppet module.  Will be removed in Katello 4.1."), :deprecated => true
+    param :uuid, String, :desc => N_("uuid of the puppet module")
     param_group :search, ::Katello::Api::V2::ApiController
     add_scoped_search_description_for(ContentViewPuppetModule)
     def index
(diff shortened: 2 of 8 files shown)


app/controllers/katello/api/v2/content_view_puppet_modules_controller.rb
```diff
@@ -9,7 +9,7 @@ class Api::V2::ContentViewPuppetModulesController < Api::V2::ApiController
     param :content_view_id, :number, :desc => N_("content view identifier"), :required => true
     param :name, String, :desc => N_("name of the puppet module")
     param :author, String, :desc => N_("author of the puppet module")
-    param :uuid, String, :desc => N_("uuid of the puppet module"), :deprecated => true
+    param :uuid, String, :desc => N_("uuid of the puppet module.  Will be removed in Katello 4.1."), :deprecated => true
```
jeremylenz, line 12: When is Puppet itself being removed?  Should we just delete these "uuid" deprecations and wait til then?

app/controllers/katello/api/v2/content_view_versions_controller.rb
```diff
@@ -95,7 +95,7 @@ def export
         fail HttpErrors::BadRequest, _("Invalid usage for Pulp 3 repositories. "\
                                        "Use hammer content-export for Yum repositories")
       end
-      ::Foreman::Deprecation.api_deprecation_warning("Export is being deprecated and will be removed in a future version of Katello. Use hammer content-view version export instead.")
+      ::Foreman::Deprecation.api_deprecation_warning("Export is being deprecated and will be removed in Katello 4.1. Use hammer content-view version export instead.")
```
jeremylenz, line 98: @parthaa tagged you for this review to get your feedback here.  "Export is being deprecated" sounds scary, but I think this just refers to the pulp2 endpoint?  Is that correct? Is there a better way to say this?

app/controllers/katello/api/v2/content_view_puppet_modules_controller.rb
jjeffers: @jlsherrill do we know if/when we will utterly remove puppet support (not just content)?

jlsherrill: puppet content support is still in the code, but upgrading to katello 4.0 will remove all the pulp support, so this entire controller will be useless (and the UI will no longer even show up).

We just haven't gotten around to removing the code parts yet

app/controllers/katello/api/v2/content_view_versions_controller.rb
jlsherrill: This is correct, you might change it to suggest the new endpoint?  And similar to the puppet parts, even though the api is there, its useless in 4.0.  Its basically non-functional

app/controllers/katello/api/v2/content_view_puppet_modules_controller.rb
jeremylenz: >  Should we just delete these "uuid" deprecations and wait til then?

^ Based on what @jlsherrill just said it seems like this is a good course of action for now.  Let me know if anyone thinks differently

app/controllers/katello/api/v2/content_view_versions_controller.rb
jeremylenz: Ok, changing it to
```
Pulp 2 export is deprecated and will not work as of Katello 4.0.  To export using Pulp 3, use hammer content-export version.
```

jeremylenz:
@jlsherrill @parthaa Updated

app/controllers/katello/api/v2/organizations_controller.rb
```diff
@@ -22,7 +22,7 @@ class Api::V2::OrganizationsController < ::Api::V2::OrganizationsController
         param :smart_proxy_ids, Array, N_("Smart proxy IDs"), :required => false
         param :compute_resource_ids, Array, N_("Compute resource IDs"), :required => false
         param :medium_ids, Array, N_("Medium IDs"), :required => false
-        param :config_template_ids, Array, N_("Provisioning template IDs"), :required => false # FIXME: deprecated
+        param :config_template_ids, Array, N_("Provisioning template IDs. Will be removed in Katello 4.1."), required: false, deprecated: true
```
jlsherrill, line 25: this can be just removed completely. It hasn't actually worked since 2019: https://github.com/theforeman/foreman/commit/9b2913f383b16b36009e64f3b988b65e6ac272b4#diff-914259019ce9daa4464c1df28cf4fb803d34182acecc24cdd450e762ae1d19e7L26

app/controllers/katello/api/v2/errata_controller.rb
```diff
@@ -62,7 +62,7 @@ def custom_index_relation(collection)
       collection
     end
 
-    api :GET, "/content_view_versions/:id/available_errata", N_("Return errata that can be added to the Content View Version via an Incremental Update"), :deprecated => true
+    api :GET, "/content_view_versions/:id/available_errata", N_("Return errata that can be added to the Content View Version via an Incremental Update. Will be removed in Katello 4.1."), :deprecated => true
```
jeremylenz, line 65: :thinking:

app/controllers/katello/api/v2/host_errata_controller.rb
```diff
@@ -51,7 +51,7 @@ def index
 
     api :PUT, "/hosts/:host_id/errata/apply", N_("Schedule errata for installation")
     param :host_id, :number, :desc => N_("Host ID"), :required => true
-    param :errata_ids, Array, :desc => N_("List of Errata ids to install"), :required => false, :deprecated => true
+    param :errata_ids, Array, :desc => N_("List of Errata ids to install. Will be removed in Katello 4.1."), :required => false, :deprecated => true
```
jeremylenz, line 54: :thinking:

app/controllers/katello/api/v2/content_exports_controller.rb
```diff
@@ -29,7 +29,7 @@ def index
 
     api :GET, "/content_exports/api_status", N_("true if the export api is pulp3 ready and usable. This API is intended for use by hammer-cli only.")
     def api_status
-      ::Foreman::Deprecation.api_deprecation_warning("/content_exports/api_status is being deprecated and will be removed in a future version of Katello.")
+      ::Foreman::Deprecation.api_deprecation_warning("/content_exports/api_status is being deprecated and will be removed in Katello 4.1.")
```
jeremylenz, line 32: :thinking:

jeremylenz (commented):
Are we okay with 4.1 for the following deprecations?

jeremylenz:
> I'm not sure if our automated deprecations tool picks up both apipie params and ::Foreman::Deprecation.api_deprecation_warnings. If it's not both, Redmine issues should be created for any deprecation that won't be called out by our tooling.

@jjeffers @johnpmitsch Do you know which flavors of deprecations our tooling handles currently?

johnpmitsch:
> @jjeffers @johnpmitsch Do you know which flavors of deprecations our tooling handles currently?

I'm pretty sure toolbelt doesn't handle apipie deprecations directly, but it does search for `::Foreman::Deprecation.api_deprecation_warning`, so if that is included it will be caught. The api method or `Foreman::Deprecation.deprecation_warning` are the methods that are searched for.

jeremylenz:
@johnpmitsch ah okay, so the ones that just have `:deprecated => true` without that method call won't be caught.   I think that applies to API params that are being deprecated, but the endpoint itself is not.

johnpmitsch:
> @johnpmitsch ah okay, so the ones that just have `:deprecated => true` without that method call won't be caught. I think that applies to API params that are being deprecated, but the endpoint itself is not.

That sounds correct to me!

jeremylenz:
@johnpmitsch Ok, I will manually add a redmine to remove those deprecated features.  Also I might put in an RFE to the deprecation tool :yum:

jeremylenz:
Added https://projects.theforeman.org/issues/31996

jeremylenz:
@johnpmitsch @jjeffers @jlsherrill Any outstanding issues or objections?

jjeffers:
@jeremylenz no issues so far as I can see

jeremylenz:
I think this is ready.  Just needs at least one approving review ;)

jeremylenz:
I think this is ready.  Just needs at least one approving review ;)

johnpmitsch (approved):
:+1:

PR #9192, opened 2021-02-23
Fixes #31949 - ignore repo user/pass if blank

jlsherrill:


jlsherrill:
Note, to reproduce this problem:

1.  User a production install (its easier).  Nightly/3.18/4.0 all work
2. use firefox
3. import your server's CA cert into firefox
4. login to  foreman/katello and save the username password
5. Create a product
6. Go to the create repo page and start to create a yum repo, notice that the username and password are auto filled in.  Highlight the username and delete it, do the same for password.
7. try to save the repo
you'll get an error and the creation task will be paused

ianballou:
I'm having trouble reproducing the issue. When you say to import the servers CA into Firefox, do you mean `/etc/pki/katello/certs/katello-server-ca.crt` ?

jlsherrill:
@ianballou yeah, that should be it.  Is firefox trusting the server?   Are you accessing it over https:?

sjha4:
So, this is also reproducible by typing a random usename/password on the create repo page fields and then clearing it out before save. (On chrome)

ianballou:
> So, this is also reproducible by typing a random usename/password on the create repo page fields and then clearing it out before save. (On chrome)

I was able to reproduce the issue like this as well on Firefox. Neat!

ianballou (approved):
Works for me!

PR #9194, opened 2021-02-24
Fixes #31920 - disassociate disabled products from sync plan

sjha4:
**To test:**
1. Enable a redhat repository and add the Product to a sync-plan
2. Disable the repository.
3. The product should get removed from the sync plan. (If the product only contains the 1 repository)
4.  The product should not get removed from the sync plan if the product contains other enabled repositories. (Enable more than 1 repository from a repository set to test this flow)

### DIFF
--- a/db/migrate/20210224160921_remove_disabled_products_from_sync_plans.rb
+++ b/db/migrate/20210224160921_remove_disabled_products_from_sync_plans.rb
@@ -1,0 +1,6 @@
+class RemoveDisabledProductsFromSyncPlans < ActiveRecord::Migration[6.0]
+  def change
+    disabled = ::Katello::Product.redhat.where.not(:id => Katello::Product.enabled)
+    disabled.update_all(:sync_plan_id => nil)
+  end
+end
--- a/app/lib/actions/katello/repository/destroy.rb
+++ b/app/lib/actions/katello/repository/destroy.rb
@@ -58,6 +58,7 @@ module Actions
         end
 
         def delete_record(repository, options = {})
+          ::Katello::SyncPlan.remove_disabled_product(repository) if repository.redhat?
           repository.destroy!
           repository.root.destroy! if repository.root.repositories.empty?
           ::Katello::DockerMetaTag.cleanup_tags if options[:docker_cleanup]
--- a/app/models/katello/sync_plan.rb
+++ b/app/models/katello/sync_plan.rb
@@ -46,6 +46,13 @@ module Katello
       end
     end
 
+    def self.remove_disabled_product(repository)
+      if (product = repository.product) && product&.redhat? && (sync_plan = product&.sync_plan) && product&.repositories&.count == 1
+        sync_plan.product_ids = (sync_plan.product_ids - [product.id])
+        sync_plan.save!
+      end
+    end
+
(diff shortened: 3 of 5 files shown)


sjha4:
[test katello]

jlsherrill:
I almost wonder if we should just drop this restriction/validation?  Having a disabled product within a sync plan wouldn't cause a problem.  

If we leave the validation in place, we might should also write a migration to clean up any existing ones a user might have

sjha4:
> I almost wonder if we should just drop this restriction/validation? Having a disabled product within a sync plan wouldn't cause a problem.

When the last repository is disabled, the product is no longer shown on the product page but it still shows up on the sync plan page if we don't remove it. Might cause confusion.

> If we leave the validation in place, we might should also write a migration to clean up any existing ones a user might have

:+1:

sjha4:
Added a migration to handle existing/old sync plans with disabled products.

db/migrate/20210224160921_remove_disabled_products_from_sync_plans.rb
```diff
@@ -0,0 +1,14 @@
...
+  def change
+    ::Katello::SyncPlan.all.each do |sync_plan|
+      product_ids = []
+      sync_plan.products.each do |product|
+        product_ids << product.id if (product&.redhat? && product&.repositories&.count == 0)
+      end
+      if product_ids.size > 0
+        sync_plan.product_ids = sync_plan.product_ids - product_ids
+        sync_plan.save!
+      end
```
jlsherrill, line 11: I think you could do something simpler (and with better performance):
```
disabled = Katello::Product.redhat.where.not(:id => Katello::Product.enabled)
Katello::SyncPlanProduct.where(:product_id => disabled).delete_all
```

jlsherrill: ohhh there's no SyncPlanProduct model

jlsherrill: here we go:
```
disabled = Katello::Product.redhat.where.not(:id => Katello::Product.enabled)
disabled.update_all(:sync_plan_id => nil)
```

Note: untested :)

sjha4: This will fetch a lot of products for a big manifest..You think it's better to operate with sync_plans since there'll be fewer of those?

jlsherrill: it should execute as a single query which would be much more efficient

chris1984 (approved):
ACK, works good APJ also pending Justin's comment.

db/migrate/20210224160921_remove_disabled_products_from_sync_plans.rb
sjha4: Ok..I'll update and test.. :+1:

sjha4: Pushed a commit with the suggested changes..

jlsherrill (approved):

PR #9198, opened 2021-02-25
Fixes #31973 - Add deprecation warning on content-host register for new method

chris1984:


engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/views/register.html
```diff
@@ -8,6 +8,11 @@ <h2 translate>Content Host Registration</h2>
 
   <div data-block="content">
     <p translate>To register a content host to this server, follow these steps.</p>
+    <p bst-alert="warning" >
+      <span translate>
+        In a future version of Katello, we are planning to deprecate the <b>katello-ca-consumer</b> package and <b>bootstrap.py</b>. Use the new registration method <a href="https://{{ katelloHostname }}/hosts/register">Register Host</a>
```
jturel, line 13: I'd tweak the wording slightly to indicate that at this time the old methods are now (technically) deprecated (correct me if wrong!).

`
The following instructions are obsolete because registering with the *katello-ca-consumer* package and *bootstrap.py* have been deprecated and will be removed in a future version. They have been replaced by the *Register Host* workflow. Registering new hosts with the below instructions is not recommended.
`

I also removed the 'Katello' usage to make this more appropriate for the downstream. Whatcha think?

jeremylenz: > we are planning to deprecate the <b>katello-ca-consumer</b> package

The [global registration template](https://github.com/theforeman/foreman/blob/fce22f37651ba76a42c2ef6385342af4fb84bb12/app/views/unattended/provisioning_templates/registration/global_registration.erb#L74) still downloads the consumer rpm, no? 
```bash
CONSUMER_RPM=$(mktemp --suffix .rpm)
    curl --output $CONSUMER_RPM <%= subscription_manager_configuration_url(hostname: @url_host) %>
    yum localinstall $CONSUMER_RPM -y
```

evgeni: Bootstrap should only handle the cases GR can't. So soft deprecated? ;)

chris1984: @jeremylenz that makes the wording harder now lol. What do you think about adding something like:

```bash
Register Host is the new preferred way of registering a Content Host. Use Register Host <link>
```

jeremylenz: I like the direction!  It feels better to me not to style it as a "warning," but just letting users know about a new way to do it.  So the style of the banner can be "info" instead of "warning". And not mentioning bootstrap or deprecation makes the messaging a lot smoother.

As for the wording, how about
```
NEW: To register a new content host without following these manual steps, see <a>Register Host</a>.
```

chris1984: Works for me, will get it updated.

jturel: I like @jeremylenz suggestion!

chris1984: Updated! screenshot:

https://nimb.ws/9InBbZ

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/views/register.html
```diff
@@ -8,6 +8,11 @@ <h2 translate>Content Host Registration</h2>
 
   <div data-block="content">
     <p translate>To register a content host to this server, follow these steps.</p>
+    <p bst-alert="info" >
+      <span translate>
+        NEW: To register a content host without following these manual steps, <a href="https://{{ katelloHostname }}/hosts/register">Register Host</a>.
```
jeremylenz, line 13:
```suggestion
        NEW: To register a content host without following these manual steps, see <a href="https://{{ katelloHostname }}/hosts/register">Register Host</a>.
```

bbuckingham:
@chris1984, For this change, it is important that we are communicating to users that the current page is being deprecated and that the recommended flow for registration would be through the global registration page.

I like the idea of stating that there is a 'New' workflow; however, it may imply an 'alternative' versus deprecation.  As a result, users may not take the needed actions to update their processes to use that new workflow.

ahumbe:
The intent of this request was mainly to educate customers about the deprecation of the current registration method and make them try the new Global Registration way, so it would be great if we can show a real warning for users.

For announcing and educating customers about the katello-agent deprecation, we had added a warning message like " _Katello-agent is deprecated and will be removed in a future release._" on the Content Host page and it was very much effective, a lot of customers migrated to REX after looking at the deprecation warning.

jeremylenz:
@ahumbe @bbuckingham Global Registration doesn't yet provide feature parity with current registration methods, so I guess I was shying away from having a scary warning so that users won't be afraid they'll be left with nothing.  I suppose we just have to make sure to balance these two facts.

How does it look if we add mild deprecation wording like this?
```
These instructions will be removed in a future release. NEW: To register a content host without following these manual steps, see <a href="https://{{ katelloHostname }}/hosts/register">Register Host</a>
```

chris1984:
Updated

chris1984:
> @ahumbe @bbuckingham Global Registration doesn't yet provide feature parity with current registration methods, so I guess I was shying away from having a scary warning so that users won't be afraid they'll be left with nothing. I suppose we just have to make sure to balance these two facts.
> 
> How does it look if we add mild deprecation wording like this?
> 
> ```
> These instructions will be removed in a future release. NEW: To register a content host without following these manual steps, see <a href="https://{{ katelloHostname }}/hosts/register">Register Host</a>
> ```

@ahumbe what do you think of the updated wording? @bbuckingham suggested maybe instead we just create a MR downstream. What are your thoughts?

ahumbe:
@chris1984 @jeremylenz Okay, I got your point of view. 

The new message you have proposed looks good to me, it is helpful to make an impact.

chris1984:
[test katello]

chris1984:
@jeremylenz @jturel how does this look? Ashish said it looks good from his side in his last comment.

jturel (approved):
LGTM!

jeremylenz (approved):
ACK :+1:

PR #9206, opened 2021-03-04
Fixes #32010 - handle pkg groups better on cv publish

jlsherrill:
with pulp3, if you copy a package group over, it copies all packages
within that group.  This means that if you exclude a package that is
in a group, and tell pulp to copy over the group the package that is
excluded will get copied.  This is a behavior change from pulp2.
To handle this, we need to treat groups like errata, where by you need
everything in the group otherwise we can not copy it.


jjeffers (approved):
Tested this change with the pulp fixture at https://repos.fedorapeople.org/pulp/pulp/demo_repos/zoo/. There are 4 package groups, includes "birds" which includes the "stork" package. 

Initial exclude filter named the "stork" package, and a CV version publication showed that this package was copied despite that filter.

Applied the change, and then re-published the CV. The new version properly excluded the package.

APJ

jjeffers (approved):
Discussed with @jlsherrill re picking true package name from location_href and how that sometimes leads to wrong package names in katello. LGTM.

PR #9209, opened 2021-03-04
Fixes #31997 - Remove Puppet content support

ianballou:
Removes Puppet support.


ianballou:
[test katello]

ianballou:
I've marked this ready for review. I'm still checking that various things still work, but I figured it would be good if people can start looking at the changes.  There are still some references to Puppet, namely in Pulp 2 - related files.  I didn't remove all of these because they'll be taken out when more of Pulp 2 is removed.

ianballou:
Realized the `webpack/` directory still has puppet content. I'll need to clean that out next.

jlsherrill:
Found a few references: 
```
test/controllers/api/v2/repositories_controller_test.rb   has a puppet module upload test
test/factories/repository_factory.rb has a 'puppet trait'
test/factories/root_repository_factory.rb:  also has a puppet trait
```

otherwise looks good, let me do a little testing

db/migrate/20160701180402_add_sortable_version_to_puppet_modules.rb
```diff
@@ -13,6 +13,8 @@ def sortable_version(version)
   end
 
   def up
+    return unless Object.const_defined?('::Katello::PuppetModule')
+
     add_column :katello_puppet_modules, :sortable_version, :string
```
jlsherrill, line 18: can you move this return below the add_column?

jlsherrill: the reason i ask, is that its better to leave the structural changes to the db as is, but any data handling is fine to skip

db/migrate/20160701180402_add_sortable_version_to_puppet_modules.rb
```diff
@@ -21,6 +23,8 @@ def up
   end
 
   def down
+    return unless Object.const_defined?('::Katello::PuppetModule')
```
jlsherrill, line 26: and the above would make this unneeded.

db/migrate/20210119162528_delete_puppet_and_ostree_repos.rb
```diff
@@ -34,7 +41,7 @@ def up
         FakeContentViewPuppetEnvironment.delete_all
         FakePuppetModule.delete_all
 
-        Katello::Repository.puppet_type.delete_all
+        ::Katello::Repository.delete(puppet_repositories)
```
jlsherrill, line 44: shouldn't this be puppet_repositories.delete_all?

jlsherrill: or does find_by_sql return an array?   I'm not actual familiar with delete() and passing in objects

ianballou: Yeah I had an issue with that. `find_by_sql` was returning an array.

ianballou:
> Found a few references:
> 
> ```
> test/controllers/api/v2/repositories_controller_test.rb   has a puppet module upload test
> test/factories/repository_factory.rb has a 'puppet trait'
> test/factories/root_repository_factory.rb:  also has a puppet trait
> ```
> 
> otherwise looks good, let me do a little testing

I assumed that the puppet module in the upload test was fine since it was just testing a random file.  It doesn't seem specifically like a puppet repo related test. In fact, it's uploading to a yum repo.

jlsherrill:
[test katello]

db/migrate/20210302165636_remove_katello_puppet_modules.rb
```diff
@@ -0,0 +1,60 @@
+class RemoveKatelloPuppetModules < ActiveRecord::Migration[4.2]
+  def up
+    drop_table :katello_content_view_puppet_environment_puppet_modules
+    drop_table :katello_repository_puppet_modules
+    drop_table :katello_puppet_modules
```
jlsherrill, line 5: a couple more tables:
katello_content_view_puppet_environments
katello_content_view_puppet_modules

jlsherrill:
Other than the two missing table deletions, i think this is good to go

jlsherrill (approved):

ianballou:
Looks like I need to regenerate katello.pot after rebasing.

ianballou:
[test katello]

ianballou:
[test katello]

ianballou:
[test katello]

PR #9211, opened 2021-03-08
Fixes #32039 - Add auditing to export history

jeremylenz:
- [ ] Add an Import History that is auditable
- [ ] Make Export History auditable
- [ ] In Monitor > Audits, I should be able to see changes to Export History and the user that made the changes
- [ ] In Monitor > Audits, I should be able to see changes to Import History and the user that made the changes
- [ ] tests


jeremylenz:
audit1

jturel:
What do you think about not adding the metadata to the audit? One can always look up the export history if they want to see the metadata. IMO the most important info is already shown - the CV name in the comments, the CVV ID, and login of the exporting user

app/lib/actions/katello/content_view_version/create_export_history.rb
```diff
@@ -0,0 +1,67 @@
...
+
+          history = ::Katello::ContentViewVersionExportHistory.new(
+            content_view_version_id: input[:content_view_version_id],
+            destination_server: input[:destination_server],
+            path: path,
+            metadata: export_metadata
+          )
+          history.export_type = history.export_type_from_metadata
+          history.audit_comment = generate_audit_comment(history, User.current, cvv)
+          history.save!
```
jeremylenz, line 46: It's not normally recommended to do DB actions in the `run` method.  But

1. it's how it was before;
2. it's only a single record so shouldn't be too risky
3. I need access to all of the above data inside `input`, which is not populated within `plan`.  for example, if I try to use `input[:exporter_data][:pulp_href]` from inside `plan`, I get a value like `"Step(3)[:exporter_data][:pulp_href]"` (as a string) instead of the populated value.

app/lib/actions/pulp3/orchestration/content_view_version/export.rb
```diff
@@ -50,56 +50,32 @@ def plan(content_view_version:, destination_server: nil,
...
+                content_view_version_id: content_view_version.id,
+                from_content_view_version_id: from_content_view_version&.id,
+                destination_server: destination_server
+              ).output
               plan_action(::Actions::Pulp3::ContentViewVersion::DestroyExporter,
-                            smart_proxy_id: smart_proxy.id,
-                            exporter_data: action_output[:exporter_data])
+                smart_proxy_id: smart_proxy.id,
+                exporter_data: action_output[:exporter_data])
+              plan_self(exporter_data: action_output[:exporter_data], smart_proxy_id: smart_proxy.id,
```
jeremylenz, line 65: `plan_self` should normally be at the bottom of `plan` so I moved it

parthaa: I 'd recommend against this. You 'd want plan_self to run here before `::Actions::Pulp3::ContentViewVersion::DestroyExporter`

jeremylenz: Within `plan` the `input` and `output` values will be unpopulated, like in [the above comment](https://github.com/Katello/katello/pull/9211#discussion_r591857535), but in `run` they will be the actual values.  Due to this I think it should still work okay?

I checked and `exporter_data` does show up with the correct values in the UI (Raw tab):
```json
  "exporter_data": {
    "pulp_href": "/pulp/api/v3/exporters/core/pulp/b1eba19c-eaa9-4625-b2eb-342858321a6f/",
    "pulp_created": "2021-03-10T19:51:46.682+00:00",
    "name": "Test_Organization_Test_CV_1.0",
    "path": "/var/lib/pulp/exports/Test_Organization/Test_CV/1.0//2021-03-10T19-51-46-00-00",
    "repositories": [
      "/pulp/api/v3/repositories/rpm/rpm/8af5dca5-0d1c-4dbc-9a18-7b4df2766de1/"
    ]
  },
```

(However, I did run into a similar issue when I tried to plan the `CreateExportHistory` action after the `DestroyExporter` action.  That was because the pulp API request for the exporter returned a 404 after it was destroyed.)

parthaa: Oh I see you moved that to an action before. LGTM

db/migrate/20210309160925_create_katello_content_view_version_import_histories.rb
```diff
@@ -0,0 +1,11 @@
+class CreateKatelloContentViewVersionImportHistories < ActiveRecord::Migration[6.0]
+  def change
+    create_table :katello_content_view_version_import_histories do |t|
+      t.string :path
+      t.text :metadata
+      t.bigint :content_view_version_id
```
parthaa, line 6:
```suggestion
      t.references :content_view_version, index: { name: :katello_cvv_import_history_cvv_id }
```
Also add a unique index to speed up things and validate duplicate data like
```
    add_index :katello_content_view_version_import_histories, [:content_view_version_id, :path],
              unique: true, name: :katello_cvv_import_history_sat_cvv_id_path
```

app/models/katello/content_view_version_import_history.rb
```diff
@@ -0,0 +1,26 @@
+module Katello
+  class ContentViewVersionImportHistory < Katello::Model
```
parthaa, line 2: Wonder if should have an api end point to facilitate searching.
```hammer content-import list``` is a good goal imo. I am open to addressing that in a different PR, but if its not too hard I 'd recommend putting it here ?

File: app/lib/actions/katello/content_view_version/create_export_history.rb [hunk 5]
```diff
@@ -0,0 +1,67 @@
...
+          toc_path_info = output[:exported_file_checksum].find { |item| item.first.end_with?("toc.json") }
+          export_metadata[:toc] = File.basename(toc_path_info.first)
+
+          history = ::Katello::ContentViewVersionExportHistory.new(
+            content_view_version_id: input[:content_view_version_id],
+            destination_server: input[:destination_server],
+            path: path,
+            metadata: export_metadata
+          )
+          history.export_type = history.export_type_from_metadata
```
parthaa, line 44: Why do you need to do this ?? shouldn't https://github.com/Katello/katello/blob/master/app/models/katello/content_view_version_export_history.rb#L22 already do that ?

parthaa: I dont think its necessary try removing and see if it sets the wrong thing. If you Have to do it 
```suggestion
          history.set_export_type
```

app/lib/actions/katello/content_view_version/create_import_history.rb
```diff
@@ -0,0 +1,33 @@
+module Actions
+  module Katello
+    module ContentViewVersion
+      class CreateImportHistory < Actions::EntryAction
+        input_format do
+          param :content_view_version_id, Integer
+          param :path, String
+          param :metadata, Hash
+          param :cv_name, String
```
parthaa, line 9:
```suggestion
          param :content_view_name, String
```

app/lib/actions/katello/content_view_version/create_import_history.rb
```diff
@@ -0,0 +1,33 @@
...
+        end
+
+        output_format do
+          param :import_history_id, Integer
+        end
+
+        def run
+          user_full_name = "#{User.current.firstname} #{User.current.lastname}"
```
parthaa, line 17: https://github.com/theforeman/foreman/blob/develop/app/models/user.rb#L218
```suggestion
          user_full_name = User.current.to_label
```

app/lib/actions/katello/content_view_version/create_import_history.rb
```diff
@@ -0,0 +1,33 @@
...
+          history = ::Katello::ContentViewVersionImportHistory.create!(
+            content_view_version_id: input[:content_view_version_id],
+            path: input[:path],
+            metadata: input[:metadata],
+            audit_comment: "Content imported from #{input[:path]} into content view '#{input[:cv_name]}' by #{user_full_name}"
```
parthaa, line 22:
```suggestion
            audit_comment: "Content imported from #{input[:path]} into content view '#{input[:cv_name]}' by #{User.current.to_label}"
```

app/lib/actions/katello/content_view_version/create_export_history.rb
```diff
@@ -0,0 +1,67 @@
...
+        end
+
+        def generate_audit_comment(export_history, user, cvv)
+          export_descriptor = if cvv.content_view.name == "Export-Library"
+                                "library export"
+                              else
+                                "export of content view '#{cvv.content_view.name}' version #{cvv.version}"
+                              end
+          user_full_name = "#{User.current.firstname} #{User.current.lastname}"
+          "#{export_history.export_type&.capitalize} #{export_descriptor} created by #{user_full_name}"
```
parthaa, line 58:
```suggestion
          "#{export_history.export_type&.capitalize} #{export_descriptor} created by #{User.current.to_label}"
```

app/models/katello/content_view_version_import_history.rb
```diff
@@ -0,0 +1,26 @@
...
+    validates :content_view_version_id, presence: true
+    validates :metadata, presence: true
+    validates :path, presence: true
+    serialize :metadata, Hash
+
+    scope :with_organization_id, ->(organization_id) do
+      where(:content_view_version_id => ContentViewVersion.with_organization_id(organization_id))
+    end
+
+    scope :with_content_view_id, ->(cv_id) do
```
parthaa, line 17: why do we need this?

parthaa: I guess in the case of export_history we need these 2 scopes at https://github.com/Katello/katello/blob/master/app/controllers/katello/api/v2/content_exports_controller.rb#L24-L25

If you plan on adding similar endpoint to the imports I 'd be ok with these in the model.

parthaa:
Looks good code wise. I 'll take a look again once you have the unit tests and addressed the comments.

parthaa:
Add an association here
https://github.com/Katello/katello/blob/master/app/models/katello/content_view_version.rb#L26

```
    has_many :import_histories, :class_name => "Katello::ContentViewVersionImportHistory", :dependent => :destroy,
             :inverse_of => :content_view_version, :foreign_key => :content_view_version_id
```

app/lib/actions/katello/content_view_version/create_export_history.rb
jeremylenz: I did it because `generate_audit_comment` uses `export_type`:
```rb
"#{export_history.export_type&.capitalize} #{export_descriptor} created by #{user_full_name}"
```
However, I need to `generate_audit_comment` before the record is saved, so the `before_validation :set_export_type` won't have run yet.  I did consider doing `history.validate`, but left it like this so it was more clear that using this value was the reason.

app/models/katello/content_view_version_import_history.rb
jeremylenz: I modeled the import history model after the existing export history one, so tbh this was just left over from that :)

If I end up adding the endpoint I'll keep it in

app/lib/actions/katello/content_view_version/create_export_history.rb
jeremylenz: Didn't know about `to_label`, thanks!

app/models/katello/content_view_version_import_history.rb
jeremylenz: Agree, I don't think it'd be good to have `hammer content-export list` but no `hammer content-import list`.  I'll start working on it; should probably be small enough to fit in this PR.

.rubocop.yml
```diff
@@ -30,6 +30,10 @@ AllCops:
 Metrics/MethodLength:
   Description: 'Avoid methods longer than 30 lines of code.'
   Max: 30 # default is 10
+  CountAsOne:
+    - array
+    - hash
+    - heredoc
```
jeremylenz, line 36: I've always thought that `Metrics/MethodLength` was too literal in terms of how it counts lines of code.  I discovered that you can tell it to count arrays/hashes/etc as a single line.  I think this is a good default for Katello going forward.

jeremylenz:
```
Lint/RedundantCopDisableDirective: Unnecessary disabling of Metrics/MethodLength
```
looks like I may have to extract this to another PR :cry:

jeremylenz: Opened https://github.com/Katello/katello/pull/9231

app/lib/actions/katello/content_view_version/create_export_history.rb
parthaa: What do you think about renaming `set_export_type` to `update_export_type_from_metadata!`  instead and calling that. I just don't like the world outside that model to know that export_type is not set and its value is based of  `export_type_from_metadata`

app/lib/actions/katello/content_view_version/create_export_history.rb
```diff
@@ -0,0 +1,65 @@
...
+          )
+          history.export_type = history.export_type_from_metadata
+          history.audit_comment = generate_audit_comment(history, cvv)
+          history.save!
+          output[:export_history_id] = history.id
+          output[:path] = history.path
+        end
+
+        def generate_audit_comment(export_history, cvv)
+          export_descriptor = if cvv.content_view.name == "Export-Library"
```
parthaa, line 51: I'd rather you add something like https://github.com/Katello/katello/blob/master/app/models/katello/content_view.rb#L106-L108 and use that.
```ruby
    EXPORT_LIBRARY = "Export-Library".freeze

    def library_export?
      name == EXPORT_LIBRARY
    end
```

and update https://github.com/Katello/katello/blob/master/app/services/katello/pulp3/content_view_version/export.rb#L169-L170 to use this `EXPORT_LIBRARY` constant.   We do not want duplicated logic

app/lib/actions/katello/content_view_version/create_export_history.rb
```diff
@@ -0,0 +1,65 @@
...
+            metadata: export_metadata
+          )
+          history.export_type = history.export_type_from_metadata
+          history.audit_comment = generate_audit_comment(history, cvv)
+          history.save!
+          output[:export_history_id] = history.id
+          output[:path] = history.path
+        end
+
+        def generate_audit_comment(export_history, cvv)
```
parthaa, line 50: what do you think of moving this method to content_view_version itself or cvvexhistory ? Will make your tests easier and may be you can combine import/export audit comments in the same method ?
``` ruby
    def audit_comment(history:, user:)
       .....
    end 
```

I am okay either way just checkin.

app/lib/actions/katello/content_view_version/create_export_history.rb
```diff
@@ -0,0 +1,65 @@
+module Actions
+  module Katello
+    module ContentViewVersion
+      class CreateExportHistory < Actions::EntryAction
```
parthaa, line 4: Considering this file is doing a lot of things tied to pulp3 I think it needs to be in the same directory as the export/import for pulp3 -> `app/lib/actions/pulp3/orchestration/content_view_version`

parthaa:
took at stab at some of my suggested changes
```diff
diff --git a/app/lib/actions/katello/content_view_version/create_export_history.rb b/app/lib/actions/katello/content_view_version/create_export_history.rb
index 92b1de706f..4bf20c3655 100644
--- a/app/lib/actions/katello/content_view_version/create_export_history.rb
+++ b/app/lib/actions/katello/content_view_version/create_export_history.rb
@@ -40,22 +40,12 @@ module Actions
             path: path,
             metadata: export_metadata
           )
-          history.export_type = history.export_type_from_metadata
-          history.audit_comment = generate_audit_comment(history, cvv)
+          history.audit_comment = history.generate_audit_comment(user: User.current)
           history.save!
           output[:export_history_id] = history.id
           output[:path] = history.path
         end
 
-        def generate_audit_comment(export_history, cvv)
-          export_descriptor = if cvv.content_view.name == "Export-Library"
-                                "library export"
-                              else
-                                "export of content view '#{cvv.content_view.name}' version #{cvv.version}"
-                              end
-          "#{export_history.export_type&.capitalize} #{export_descriptor} created by #{User.current.to_label}"
-        end
-
         def humanized_name
           _("Create Export History")
         end
diff --git a/app/lib/actions/katello/content_view_version/create_import_history.rb b/app/lib/actions/katello/content_view_version/create_import_history.rb
index 73ec7fa5eb..719bce4d48 100644
--- a/app/lib/actions/katello/content_view_version/create_import_history.rb
+++ b/app/lib/actions/katello/content_view_version/create_import_history.rb
@@ -18,7 +18,9 @@ module Actions
             content_view_version_id: input[:content_view_version_id],
             path: input[:path],
             metadata: input[:metadata],
-            audit_comment: "Content imported from #{input[:path]} into content view '#{input[:content_view_name]}' by #{User.current.to_label}"
+            audit_comment: ::Katello::ContentViewVersionImportHistory.generate_audit_comment(path: input[:path],
+                                                                                             user: User.current,
+                                                                                             content_view_name: input[:content_view_name])
           )
           output[:import_history_id] = history.id
         end
diff --git a/app/models/katello/content_view.rb b/app/models/katello/content_view.rb
index 47c8c3ffb1..20440a4511 100644
--- a/app/models/katello/content_view.rb
+++ b/app/models/katello/content_view.rb
@@ -9,6 +9,7 @@ module Katello
 
     CONTENT_DIR = "content_views".freeze
     IMPORT_LIBRARY = "Import-Library".freeze
+    EXPORT_LIBRARY = "Export-Library".freeze
 
     belongs_to :organization, :inverse_of => :content_views, :class_name => "::Organization"
 
@@ -107,6 +108,10 @@ module Katello
       name == IMPORT_LIBRARY
     end
 
+    def library_export?
+      name.start_with? EXPORT_LIBRARY
+    end
+
     def content_host_count
       hosts.count
     end
diff --git a/app/models/katello/content_view_version_export_history.rb b/app/models/katello/content_view_version_export_history.rb
index a379a9054b..426cc870e7 100644
--- a/app/models/katello/content_view_version_export_history.rb
+++ b/app/models/katello/content_view_version_export_history.rb
@@ -2,6 +2,7 @@ module Katello
   class ContentViewVersionExportHistory < Katello::Model
     include Authorization::ContentViewVersionExportHistory
     audited except: :metadata
+    delegate :organization, to: :content_view_version
 
     COMPLETE = "complete".freeze
     INCREMENTAL = "incremental".freeze
@@ -20,7 +21,9 @@ module Katello
     validates :metadata, :presence => true
     serialize :metadata, Hash
 
-    before_validation :set_export_type, :if => -> { export_type.blank? }
+    after_initialize do |history|
+      history.set_export_type if history.metadata
+    end
 
     scope :with_organization_id, ->(organization_id) do
       where(:content_view_version_id => ContentViewVersion.with_organization_id(organization_id))
@@ -45,7 +48,20 @@ module Katello
     end
 
     def set_export_type
-      self.export_type = export_type_from_metadata
+      self.export_type ||= export_type_from_metadata
+    end
+
+    def organization_id
+      organization.id
+    end
+
+    def generate_audit_comment(user:)
+      export_descriptor = if content_view_version.content_view.library_export?
+                      "library export"
+                    else
+                      "export of content view version '#{content_view_version.name}'"
+                    end
+      "#{export_type&.capitalize} #{export_descriptor} created by #{user&.to_label}"
     end
   end
 end
diff --git a/app/models/katello/content_view_version_import_history.rb b/app/models/katello/content_view_version_import_history.rb
index ed6dc1c4c1..ebf56ec793 100644
--- a/app/models/katello/content_view_version_import_history.rb
+++ b/app/models/katello/content_view_version_import_history.rb
@@ -1,6 +1,7 @@
 module Katello
   class ContentViewVersionImportHistory < Katello::Model
     audited except: :metadata
+    delegate :organization, to: :content_view_version
 
     belongs_to :content_view_version, :class_name => "Katello::ContentViewVersion", :inverse_of => :import_histories
     validates_lengths_from_database
@@ -21,5 +22,17 @@ module Katello
     scoped_search :on => :content_view_id, :relation => :content_view_version, :validator => ScopedSearch::Validators::INTEGER, :only_explicit => true
     scoped_search :on => :content_view_version_id, :only_explicit => true, :validator => ScopedSearch::Validators::INTEGER
     scoped_search :on => :id, :only_explicit => true, :validator => ScopedSearch::Validators::INTEGER
+
+    def organization_id
+      organization.id
+    end
+
+    def self.generate_audit_comment(user:, path:, content_view_name:)
+      _("Content imported from %{path} into content view '%{name}' by %{user}") % {
+          path: path,
+          user: user.to_label,
+          name: content_view_name
+      }
+    end
   end
 end
diff --git a/app/services/katello/pulp3/content_view_version/export.rb b/app/services/katello/pulp3/content_view_version/export.rb
index 2b6b6b1f30..64cf3b6501 100644
--- a/app/services/katello/pulp3/content_view_version/export.rb
+++ b/app/services/katello/pulp3/content_view_version/export.rb
@@ -166,7 +166,7 @@ module Katello
         def self.find_library_export_view(create_by_default: false,
                                           destination_server:,
                                           organization:)
-          name = "Export-Library"
+          name = ::Katello::ContentView::EXPORT_LIBRARY
           name += "-#{destination_server}" unless destination_server.blank?
           select_method = create_by_default ? :first_or_create : :first
           ::Katello::ContentView.where(name: name, organization: organization).send(select_method)
```

app/models/katello/content_view_version_export_history.rb
```diff
@@ -2,6 +2,8 @@ module Katello
   class ContentViewVersionExportHistory < Katello::Model
     include Authorization::ContentViewVersionExportHistory
     audited except: :metadata
+    delegate :organization, to: :content_view_version
+    delegate :id, to: :organization, prefix: true
```
jeremylenz, line 6: This is the same as 
```rb
def organization_id
  organization.id
end
```
Rubocop didn't like when I tried to do the above, so I discovered that I can use `delegate` with `prefix: true` and it has the same effect.

app/lib/actions/katello/content_view_version/create_export_history.rb
jeremylenz: you're right, moving it

app/lib/actions/katello/content_view_version/create_export_history.rb
jeremylenz: Looks like adding the `after_initialize` callback in the model saves you from having to call it here! :+1:

app/models/katello/content_view_version_export_history.rb
parthaa: wow

.rubocop.yml
jturel: This should be removed now since you've got an open PR :scissors:

test/controllers/api/v2/content_imports_controller_test.rb
```diff
@@ -74,5 +74,11 @@ def test_version
       post :version, params: { content_view_id: @library_view.id, path: path, metadata: metadata_params}
       assert_response :success
     end
+
+    def test_history_index
+      get :index
+      assert_response :success
+      assert_template 'api/v2/content_view_version_import_histories/index'
+    end
```
parthaa, line 82: Can you add a permissions test ? look at -> https://github.com/Katello/katello/blob/master/test/controllers/api/v2/content_exports_controller_test.rb#L23

parthaa:
You need to add `:index` here -> https://github.com/Katello/katello/blob/master/lib/katello/permission_creator.rb#L144

jeremylenz:
@parthaa updated. here goes nothing!

config/initializers/audited.rb
```diff
@@ -0,0 +1,13 @@
+class KatelloAudit < Audited::Audit
+  def self.name
+    if super.start_with?("::")
+      super
+    else
+      "::#{super}"
+    end
+  end
```
jeremylenz, line 8: This should fix the `_reflect_on_association` errors.  @jturel does the way I added this Rails initializer look ok to you?  (I followed the instructions from Audited [here](https://github.com/collectiveidea/audited#custom-audit-model))

jeremylenz: for context:  This will cause `has_many` to be called correctly within the `audited` gem [here](https://github.com/collectiveidea/audited/blob/master/lib/audited/auditor.rb#L80)

jeremylenz:
zero_failed

parthaa (approved):
Ran a few tests and worked well. ACK!

app/models/katello/content_view_version_export_history.rb
```diff
@@ -44,7 +49,16 @@ def export_type_from_metadata
     end
 
     def set_export_type
-      self.export_type = export_type_from_metadata
+      self.export_type ||= export_type_from_metadata
+    end
+
+    def generate_audit_comment(user:)
+      export_descriptor = if content_view_version.content_view.name == "Export-Library"
```
jturel, line 56: I think I saw "Export-Library" as a constant somewhere

parthaa: Good call..
```suggestion
      export_descriptor = if content_view_version.content_view.library_export?
```

jeremylenz:
So @jturel and I discovered the cause of the `_reflect_on_association` errors:

Inside the `audited` gem, `has_many` was being called with `class_name` parameter of `"Katello::ContentViewVersionExportHistory::Audited::Audit"` when it should be just `"Audited::Audit"`. But why was this happening?

In Rails code, Jonathan discovered a method in .vendor/ruby/2.5.0/gems/activesupport-6.0.3.4/lib/active_support/dependencies.rb called `autoloadable_module?`:
```rb
    def autoloadable_module?(path_suffix)
      autoload_paths.each do |load_path|
        return load_path if File.directory? File.join(load_path, path_suffix)
      end
      nil
    end
```

When testing a `load_path` of `/home/vagrant/foreman/lib/katello/content_view_version_export_history/audited`, `File.directory?` was returning `true`, even though that path is _not_ a directory (and in fact, does not exist at all.)  This is confirmed with pry inside the test:
```
[4] pry(ActiveSupport::Dependencies)> dir
=> "/home/vagrant/foreman/lib/katello/content_view_version_export_history/audited"
[5] pry(ActiveSupport::Dependencies)> Dir.exist?("/home/vagrant/foreman/lib/katello/content_view_version_export_history/audited")
=> false
[6] pry(ActiveSupport::Dependencies)> File.directory?("/home/vagrant/foreman/lib/katello/content_view_version_export_history/audited")
=> true
```

So, we were considering possibilities:
1. the directory momentarily exists (unlikely)
2. the `File.directory` method has been redefined 

To test this, we used the `source_location` method like in [this article](https://blog.revathskumar.com/2012/12/ruby-check-whether-method-is-monkey-patched-or-not.html)

and discovered that the source location for the `File.directory?` code is...
```
"/home/vagrant/foreman/.vendor/ruby/2.5.0/gems/mocha-1.12.0/lib/mocha/stubbed_method.rb"
```

Now we're getting somewhere.  Turns out that in `export_test.rb`, there was this line:
```rb
File.expects(:directory?).returns(true).at_least_once
```

This was causing failures every time in `export_test`, and depending on test order, the failure could spill over to other tests.

It appears that deleting that line from the `export_test` both solves the issue itself, and doesn't require any other modifications - the test still passes without that mock.

test/actions/katello/content_view_version/export_test.rb
```diff
@@ -13,12 +12,12 @@ def setup
       @content_view = @repo.content_view
       @content_view_version = @repo.content_view_version
       @destination_server = 'dream-destination'
-      FileUtils.mkdir_p(Setting['pulpcore_export_destination'])
```
jturel: Since this failed due to a permissions issue, you could override the setting value to a Tmpdir -> https://ruby-doc.org/stdlib-2.5.1/libdoc/tmpdir/rdoc/Dir.html or `Rails.root.join('tmp')`

The tmpdir will be writable by jenkins, and I'd like to see that approach here because I don't want to get bitten by this again - although I see you've added a workaround.

jturel: One may say stub everything you aren't testing but I would say look what happens when you do that! ;)

jeremylenz: @jturel yeah, I was thinking about that but I decided to just improve the stub (and make sure it's unstubbed after the tests). I think the `unstub` will prevent the future biting, and feels more correct than actually altering the filesystem on the test machine.  thoughts?

jturel: It's not world-ending but I prefer creating our own prescribed dir because:

- someone might come and delete that line in the teardown if they see the tests pass (I think Jenkins will even be green if that test runs toward the end)
- Stubbing such a high level construct (File) tends to be brittle in general. *any* code that gets executed could affect the stub - so a gem update might cause it to fail for example.

TLDR - just writing our own file will probably prevent us from ever having to look at this test again for the same or similar reason we had to this time around and for me that's a huge plus

jturel: If you do want to keep the stub then I think the line you put in the teardown must have a comment about why it's critical cause removing it is not as innocuous as it seems

jeremylenz: (just for future Github readers) We ended up going with the previous approach of not stubbing `File.directory` at all.

jturel (approved):
Exquisite!

PR #9219, opened 2021-03-11
Fixes #32067 - Content Hosts + ActivationKey repository set UI fixes

parthaa:
This commit addresses multiple issues with the Content Hosts and  Activation Key repository
sets page.
These include
1) Auto complete search
2) Scoped search based on different things with regard to content
3) More consistent handling of select all and other facets.


app/models/katello/concerns/organization_extensions.rb
```diff
@@ -174,12 +174,8 @@ def syncable_content?
         end
 
         def enabled_product_content_for(roots)
```
jeremylenz, line 176: :carrot: :sweet_potato:

app/services/katello/product_content_finder.rb
```diff
@@ -18,7 +18,7 @@ def product_content
       if match_environment
         environment = consumable.lifecycle_environment
         view = consumable.content_view
-        return [] unless environment && view
+        return ProductContent.where("1=0") unless environment && view
```
jeremylenz, line 21: Ah, this is where we were returning the blank array :smile: 

`"1=0"` works fine, but it's always been just super weird to me.  What about `ProductContent.none` ?  That should always return an `ActiveRecord::Relation` as well.

app/controllers/katello/api/v2/repository_sets_controller.rb
```diff
@@ -122,7 +130,22 @@ def index_relation
 
       relation = relation.where(Katello::Content.table_name => {:name => params[:name]}) if params[:name].present?
       relation = relation.redhat unless ::Foreman::Cast.to_bool(params[:with_custom])
-      relation
+      index_relation_with_activation_key_overrides(relation)
+    end
+
+
+    def index_relation_with_activation_key_overrides(relation)
+      return relation if @activation_key.nil?
```
jeremylenz, line 138: I kind of prefer using Rails' `.blank?` over `.nil?` since it also covers empty strings, so it feels safer

app/controllers/katello/api/v2/repository_sets_controller.rb
```diff
@@ -163,5 +186,11 @@ def custom_product?
     def substitutions
       params.permit(:basearch, :releasever).to_h
     end
+
+   def find_authorized_activation_key
+      return unless params[:activation_key_id]
+      @activation_key = Katello::ActivationKey.readable.find_by(:id => params[:activation_key_id])
+      throw_resource_not_found(name: 'activation_key', id: params[:activation_key_id]) if @activation_key.nil?
```
jeremylenz, line 193: same as above

engines/bastion_katello/app/assets/javascripts/bastion_katello/activation-keys/details/activation-key-repository-sets.controller.js
```diff
@@ -14,20 +15,19 @@
  *   Provides the functionality for the activation-key products action pane.
  */
 angular.module('Bastion.activation-keys').controller('ActivationKeyRepositorySetsController',
-    ['$scope', 'translate', 'Nutupane', 'ActivationKey', 'ContentOverrideHelper', 'Notification', 'CurrentOrganization',
-    function ($scope, translate, Nutupane, ActivationKey, ContentOverrideHelper, Notification, CurrentOrganization) {
+    ['$scope', 'translate', 'Nutupane', 'ActivationKey','RepositorySet', 'ContentOverrideHelper', 'Notification', 'CurrentOrganization',
+    function ($scope, translate,  Nutupane, ActivationKey, RepositorySet, ContentOverrideHelper, Notification, CurrentOrganization) {
```
jeremylenz, line 19:
```suggestion
    function ($scope, translate, Nutupane, ActivationKey, RepositorySet, ContentOverrideHelper, Notification, CurrentOrganization) {
```

jeremylenz (commented):
Thanks @parthaa !

Tested and works as expected
* "Limit to Environment" check box filters repos accordingly
* Autocomplete search works again

A couple comments below

app/services/katello/product_content_finder.rb
parthaa: That sounds interesting :).. Lemme see

jeremylenz:
[test katello]

jeremylenz:
@parthaa Looks like at least some of the test failures are related :cry:

parthaa:
[test katello]

parthaa:
[test katello]

app/models/katello/concerns/organization_extensions.rb
```diff
@@ -174,12 +174,8 @@ def syncable_content?
...
-          return [] if roots.blank?
-          content_ids = roots.pluck(:content_id)
-
-          filtered_product_content do |pc|
-            content_ids.include?(pc.content.cp_content_id) && pc.product.enabled?
-          end
+          Katello::ProductContent.joins(:content).where(:product_id => self.products.enabled,
+                             "#{::Katello::Content.table_name}.cp_content_id" => roots.pluck(:content_id))
```
jturel, line 178: this might be a case to use select vs pluck?

jturel:
This pr looks awesome!!

app/models/katello/concerns/organization_extensions.rb
parthaa: I think pluck uses select internally ?

jturel:
Is it time to merge this?

parthaa:
@chris1984  @jturel @jeremylenz  - I added support for content hosts repository sets page also. Please give it a quick whirl .

jeremylenz (changes requested):
Content Host page repository sets seems to work fine! But I found one issue on the AK page.

Not sure if this is a regression or I just didn't notice before:

On the Activation Keys page, 
1. load the page
2. click Repository Sets or any other tab
3. click back to the Details tab

The Content View will be in edit mode, even though you didn't click the edit button:
ak_cv_edit_mode

jeremylenz (commented):
found some more issues with the autocomplete:

When I type "content_type =", it autocompletes to that point ok but doesn't show me possible values.  I'm on my own after that. 

When I start typing "enabled_by_default" and then click on the autocomplete suggestion, then press Space, the dropdown then contains a single item, "undefined".

jturel:
We groomed card 7805 on the board to fix that content view selection issue this sprint! I think it's the same one at least

parthaa:
> When I type "content_type =", it autocompletes to that point ok but doesn't show me possible values. I'm on my own after that.
> 
> When I start typing "enabled_by_default" and then click on the autocomplete suggestion, then press Space, the dropdown then contains a single item, "undefined".

Should be resolved. Check latest commit

parthaa:
> * load the page
> * click Repository Sets or any other tab
> * click back to the Details tab

It looks like a pre existing issue. I noticed it in one of our demo sats also. Not sure this PR is the right place to fix it. Created https://projects.theforeman.org/issues/32225 to address it

PR #9222, opened 2021-03-12
Fixes #32079 - CI reporting for pulpcore integration tests

jjeffers:


lib/katello/tasks/jenkins.rake
```diff
@@ -18,6 +18,13 @@ begin
       Rake::Task['jenkins:setup:minitest'].invoke
       Rake::Task['rake:test:katello:test'].invoke
     end
+
+    desc "Runs puplcore integration tests"
+    task 'katello:pulpcore' do
+      Rake::Task['jenkins:setup:minitest'].invoke
+      Rake::Task['rake:test:katello:pulpcore'].invoke
+    end
```
jlsherrill, line 26: should these just always run in live mode?

jjeffers: Not necessarily - the setup is primarily for generation of the JUnit style output, so this mirrors the same pulpcore test task in test.rake. The advantage here is that we are only interested in the pulp3 tests. In the pipeline, we issue the command:
`mode=all bundle exec rails jenkins:katello:pulpcore`.

jlsherrill (approved):

jjeffers:
[test katello]

jjeffers:
[test katello]

jjeffers:
[test katello]

jjeffers:
[test katello]

jjeffers:
[test katello]

PR #9227, opened 2021-03-16
Fixes #32106 - Default registration method

stejskalleos:
Make 'Global Registration' default method for host registration and
remove old Content host registration page
'Register content host' button redirects to Global Registration page


jturel:
@stejskalleos thanks for the PR, but I have some concerns:

- 4.0 *just* added all of the additional steps for non-EL distros to get clients registered and this removes that
- global registration is rather new, so perhaps removing this page entirely is too soon
- I don't think global registration itself accounts for all the documented steps for the various client distros which would leave users stuck (correct me if wrong there, I honestly have not tested of verified that)

stejskalleos:
@jturel

> 4.0 just added all of the additional steps for non-EL distros to get clients registered and this removes that
> global registration is rather new, so perhaps removing this page entirely is too soon

It goes with the plan for Satellite 6.10, where the global registration should be default & primary registration method (and also replacing `Bootstrap.py` since it's going to be deprecated).

> I don't think global registration itself accounts for all the documented steps for the various client distros which would leave users stuck (correct me if wrong there, I honestly have not tested of verified that)

It does, it is possible to register with it various distros (Red Hat, Debian ...)
See:
* [Documentation](https://docs.theforeman.org/nightly/Managing_Hosts/index-foreman-el.html#registering-a-host-to-project-using-the-global-registration-template_managing-hosts)
* (for Sat. 6.10) new Form version with new fields & features ([Foreman PR](https://github.com/theforeman/foreman/pull/8419), [Katello PR](https://github.com/Katello/katello/pull/9249), [REX PR](https://github.com/theforeman/foreman_remote_execution/pull/574))

ares:
Yeah, ideally this should get to 4.1 ideally, even though I understand it was just improved... is there anything that these pages offer that is not handled by the new registration? Also note there's new react version of the form planned to be merged for Foreman 2.5 enhancing the user experience, see https://github.com/theforeman/foreman/pull/8419 for the screenshots and https://github.com/Katello/katello/pull/9249 for the Katello part.

jturel:
[test katello]

jturel:
Thanks @stejskalleos + @ares. My concerns are addressed! APJ

jturel (approved):
ACK

PR #9229, opened 2021-03-16
Run react UI tests on all branches

ehelms:
The changes in https://github.com/theforeman/jenkins-jobs/pull/20 will drop React testing that normally would happen on release branches to ensure cherry picks introduce no regressions. This would leave a gap if we leave GH actions only testing PRs on master.


jlsherrill (approved):

chris1984 (approved):
ACK

PR #9232, opened 2021-03-16
Fixes #32100 - Unable to set HostGroup content source to capsule that isn't synced

ianballou:
Hosts don't validate that the content source and lifecycle environment match, so host groups shouldn't need to either.

To see the validation in action:

1) Load up the foreman console
2) Create a lifecycle environment and a pulp node smart proxy
3) Make a hostgroup (note that if you try to make a mismatched LCE and content source, the hostgroup creation will silently fail.  This might be due to a `create` vs `create!` situation.)
4) Create a new hostgroup content facet like so:
```ruby
::Katello::Hostgroup::ContentFacet.create!(hostgroup_id: <hostgroup_id>, kickstart_repository_id: nil, content_source_id: <mismatched_content_source_id>, content_view_id: nil, lifecycle_environment_id: <mismatched_LCE_id>)
```
5) See the error "The selected content source and lifecycle environment do not match".

### DIFF
--- a/app/lib/katello/validators/hostgroup_kickstart_repository_validator.rb
+++ b/app/lib/katello/validators/hostgroup_kickstart_repository_validator.rb
@@ -2,12 +2,6 @@ module Katello
   module Validators
     class HostgroupKickstartRepositoryValidator < ActiveModel::Validator
       def validate(facet)
-        # check content source first, otherwise it's meaningless to proceed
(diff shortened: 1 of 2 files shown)


ianballou:
[test katello]

ianballou:
[test katello]

chris1984 (approved):
ACK, works great!

PR #9234, opened 2021-03-18
Fixes #32119 - adjust task group error check

jlsherrill:
this adjusts the error check to handle 2 conditions
1) do not report an error while there are tasks still left to execute
2) if a task is cancelled and there are no tasks left to execute, report an error

### DIFF
--- a/app/services/katello/pulp3/task_group.rb
+++ b/app/services/katello/pulp3/task_group.rb
@@ -71,8 +71,11 @@ module Katello
       end
 
       def error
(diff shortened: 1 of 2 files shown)


jlsherrill:
There are some reproducer steps in the issue, but this review may end up being mostly a thought exercise

jlsherrill:
[test rpm]

ianballou:
[test katello]

ianballou (approved):
Gave it some thought and tried the reproducer steps. All seems good to me!

PR #9237, opened 2021-03-19
Fixes #32142 - import fewer errata at a time

jlsherrill:
during migration


ianballou (approved):
Works fine for me!

I synced rhel 7 optional and rhel 7 extras and then published them in 5 content view versions.  3657 errata per content view version.  The Pulp 3 migration and switchover went well. Afterwards, when publishing new content view versions, the amount of errata stayed the same.

PR #9238, opened 2021-03-22
Fixes #31968 - adds support for new pulpcore (c 3.11) feature names

jjeffers:


jjeffers:
[test katello]

chris1984 (approved):
ACK, looks fine

jjeffers:
@chris1984 thank you for reviewing this. I do have one concern, and I'm not 100% sure it's addressed in this PR. @jlsherrill every plugin is accounted for except for "core", but it wasn't clear to me how that is used by katello. The PR assumes that the feature names are used for checking supported repository types based on the capabilities of the "pulpcore" feature.

What might be missing without also making any changes to account for the new "core" (formerly "pulpcore") name?

jlsherrill:
@jjeffers we don't actually check for pulpcore, so i think that is fine.  One comment about your PR is that the repository_type definitions 'for example:. https://github.com/Katello/katello/blob/master/lib/katello/repository_types/file.rb#L6'  still refer to the old name, and then you're programmatically checking for the new name.   I think doing the opposite would be better, updating all the repository_type files to use the new name, and then programmatically supporting the old name.  Thoughts?

jjeffers:
>  I think doing the opposite would be better, updating all the repository_type files to use the new name, and then programmatically supporting the old name. Thoughts?

I had considered that, too. It would map better to the new names, since someone reading the code wouldn't wonder why there were not the same (until they read the changes in smart proxy extensions). Good point!

jlsherrill (approved):

jjeffers:
[test katello]

jjeffers:
[test katello]

jjeffers:
[test katello]

jjeffers:
[test katello]

PR #9249, opened 2021-03-29
Fixes #31809 - Host Registration - New Form

stejskalleos:
Refactoring of the Katello's form extension, now completely in React & with PF4 components.

**REQUIRES FOREMAN:** https://github.com/theforeman/foreman/pull/8419

### Summary
**Activation key(s)**
- `<Select>` instead of `<input>` field
- Showing key's environment for each option
- Scoped to selected organization, changing organization will reset selected the field
- Inheriting ACKs from host group
- If no keys are selected (by user or from host group), form can't be submitted.

**Life cycle environment** _(new field)_
- Scoped to selected organization, changing ORG will reset the field
- Inheriting environment from host group

**Force** _(new field)_
Remove any `katello-ca-consumer*` packages & run `subscription-manager register` with `--force` parameter.

**Ignore errors** _(new field)_
Ignore subscription-manager errors for `subscription-manager register` command

**API changes**
- New registration params: `:lifecycle_environment_id`, `:force` & `:ignore_subman_errors`
- `:activation_key` renamed to `:activation_keys` & changed to `Array`


webpack/components/extensions/RegistrationCommands/RegistrationCommandsPageHelpers.js
```diff
@@ -0,0 +1,21 @@
+// Activation Keys helpers
+
+export const validateAKField = (hostGroupId, userKeys, hgKeys) => {
+  if (hostGroupId === '') {
+    return (userKeys?.length > 0 ? 'success' : 'error');
```
laviro, line 5: Can you use constants from foreman core for SUCCESS, ERROR?

webpack/components/extensions/RegistrationCommands/RegistrationCommandsPageHelpers.js
```diff
@@ -0,0 +1,21 @@
...
+
+  return ((userKeys?.length > 0 || hgKeys?.length > 0) ? 'success' : 'error');
+};
+
+export const akHasValidValue = (hostGroupId, userKeys, hgKeys) => {
+  if (hostGroupId === '') {
+    return (userKeys?.length > 0);
+  }
+
+  return (hgKeys?.length > 0 || userKeys?.length > 0);
```
laviro, line 20: does it prevent from submitting the registration form?
I am not sure every user will actually need this plugin extension each time,
so maybe it should be `optional`?

webpack/components/extensions/RegistrationCommands/fields/Force.js
```diff
@@ -0,0 +1,47 @@
...
+          >
+            <button
+              className="pf-c-form__group-label-help"
+              onClick={e => e.preventDefault()}
+            >
+              <HelpIcon noVerticalAlign />
+            </button>
+          </Popover>
+        </span>
+      }
```
laviro, line 26: I see that the `label` component needed to be reinvented in almost all use-cases with the form,
can you move it into foreman's common dir?

laviro (commented):
works great! left a few inline comments

webpack/components/extensions/RegistrationCommands/RegistrationCommandsPageHelpers.js
stejskalleos: Yes, it is desired behavior, we don't want to allow submit the form without activation keys.

webpack/components/extensions/RegistrationCommands/RegistrationCommandsPageHelpers.js
stejskalleos: Which constants do you mean, I couldn't find any 'success' & 'error' constants. I need to have it in lowercase for the PF4 component `FormGroup` in `RegistrationCommands/fields/ActivationKeys.js`

laviro (approved):
Thanks @stejskalleos !! Tested with the core PR and looks great to me :)

webpack/components/extensions/RegistrationCommands/__tests__/LifeCycleEnvironment.test.js
```diff
@@ -0,0 +1,11 @@
+import { testComponentSnapshotsWithFixtures } from 'react-redux-test-utils';
+import LifeCycleEnvironment from '../fields/LifeCycleEnvironment';
```
jturel, line 2: Correct case here would be `LifecycleEnvironment` I think

webpack/components/extensions/RegistrationCommands/fields/IgnoreSubmanErrors.js
```diff
@@ -0,0 +1,38 @@
...
+import { FormGroup, Checkbox } from '@patternfly/react-core';
+import LabelIcon from 'foremanReact/components/common/LabelIcon';
+import { translate as __ } from 'foremanReact/common/I18n';
+
+const IgnoreSubmanErrors = ({ value, onChange, isLoading }) => (
+  <FormGroup fieldId="reg_katello_ignore">
+    <Checkbox
+      label={
+        <span>
+          {__('Ignore errors')}{' '}
```
jturel, line 14: Does the functionality ignore just subman errors or any error raised while registering? Just want to be sure that we are being clear. In general I don't think we should refer to the name of other tooling (subman here) if we can help it - that may change one day (and perhaps soon!)

webpack/components/extensions/RegistrationCommands/fields/LifeCycleEnvironment.js
```diff
@@ -0,0 +1,56 @@
...
+import { noop } from 'foremanReact/common/helpers';
+
+const LifeCycleEnvironment = ({
+  pluginValues, onChange, isLoading,
+  hostGroupEnvironment, lifeCycleEnvironments,
+}) => (
+  <FormGroup
+    label={__('Lifecycle enviroment')}
+    fieldId="reg_katello_lce"
+    helperText={hostGroupEnvironment && sprintf('From host group: %s', hostGroupEnvironment)}
```
jturel, line 19: I think something like Host Group should be capitalized? Not 100% sure on that. Where's our style guide?? :)

webpack/components/extensions/RegistrationCommands/fields/LifeCycleEnvironment.js
```diff
@@ -0,0 +1,56 @@
...
+    <FormSelect
+      value={pluginValues?.lifecycleEnvironmentId}
+      onChange={v => onChange({ lifecycleEnvironmentId: v })}
+      className="without_select2"
+      id="reg_katello_lce"
+      isDisabled={isLoading || lifeCycleEnvironments.length === 0}
+    >
+      <FormSelectOption
+        value=""
+        label={lifeCycleEnvironments.length === 0 ? __('No LCE to select') : ''}
```
jturel, line 30: s/LCE/Lifecycle Environment

app/controllers/katello/concerns/api/v2/registration_commands_controller_extensions.rb
```diff
@@ -1,31 +1,53 @@
...
 
-          def format_activation_key
-            return if registration_params[:activation_key].blank?
-            registration_params[:activation_key] = registration_params[:activation_key].split(',').map(&:strip).reject(&:blank?).join(',')
+        update_api(:create) do
+          param :registration_command, Hash do
+            param :activation_key, String, desc: N_('Activation key(s) for Subscription Manager. Required for CentOS and Red Hat Enterprise Linux. Multiple keys add separated by comma, example: key1,key2,key3.'), deprecated: true
+            param :activation_keys, Array, required: true, desc: N_('Activation key(s) for Subscription Manager. Required for CentOS and Red Hat Enterprise Linux. Required only if hostgroup has no activation keys')
+            param :lifecycle_environment_id, :number, required: false, desc: N_('Life cycle environment for the host.')
+            param :force, :bool, required: false, desc: N_('Remove any `katello-ca-consumer` rpms before registration and run subscription-manager with `--force` argument.')
```
jturel, line 41: I think this description is a little too specific - the katello-ca-consumer part is not very meaningful imo. How about "Clear any previous registration and run subscription-manager with `--force`" ?

app/controllers/katello/concerns/api/v2/registration_commands_controller_extensions.rb
```diff
@@ -1,31 +1,53 @@
...
+        extend ::Apipie::DSL::Concern
 
-          def format_activation_key
-            return if registration_params[:activation_key].blank?
-            registration_params[:activation_key] = registration_params[:activation_key].split(',').map(&:strip).reject(&:blank?).join(',')
+        update_api(:create) do
+          param :registration_command, Hash do
+            param :activation_key, String, desc: N_('Activation key(s) for Subscription Manager. Required for CentOS and Red Hat Enterprise Linux. Multiple keys add separated by comma, example: key1,key2,key3.'), deprecated: true
+            param :activation_keys, Array, required: true, desc: N_('Activation key(s) for Subscription Manager. Required for CentOS and Red Hat Enterprise Linux. Required only if hostgroup has no activation keys')
+            param :lifecycle_environment_id, :number, required: false, desc: N_('Life cycle environment for the host.')
```
jturel, line 40: Lifecycle Environment

app/controllers/katello/concerns/api/v2/registration_commands_controller_extensions.rb
```diff
@@ -1,31 +1,53 @@
...
-          end
+      module ApiPieExtensions
+        extend ::Apipie::DSL::Concern
 
-          def format_activation_key
-            return if registration_params[:activation_key].blank?
-            registration_params[:activation_key] = registration_params[:activation_key].split(',').map(&:strip).reject(&:blank?).join(',')
+        update_api(:create) do
+          param :registration_command, Hash do
+            param :activation_key, String, desc: N_('Activation key(s) for Subscription Manager. Required for CentOS and Red Hat Enterprise Linux. Multiple keys add separated by comma, example: key1,key2,key3.'), deprecated: true
```
jturel, line 38: I think the example can be removed in favor of just `Separate multiple activation key names with a comma`

app/controllers/katello/concerns/api/v2/registration_commands_controller_extensions.rb
```diff
@@ -1,31 +1,53 @@
...
-              end
+    module Api::V2::RegistrationCommandsControllerExtensions
+      extend ActiveSupport::Concern
+
+      module Overrides
+        def registration_args
+          args = super
+
+          if args['activation_key'].present?
+            args['activation_keys'] = args['activation_key']
```
jturel, line 11: So if I pass both `activation_key` and `activation_keys` it looks like the former wins by overwriting the latter?

How about starting this off like:

```
args['activation_keys'] ||= []

# parse activation_key param

args['activation_keys'] << parsed_activation_key_param
```

That should prevent the need to check for String and Array which are usually code smells in Ruby

app/controllers/katello/concerns/api/v2/registration_commands_controller_extensions.rb
jturel, line 38: I think we can also remove the 'for Subscription Manager' part to avoid coupling our domain objects with the client tooling

jturel (commented):
The new forms looks great @stejskalleos. Here's my first pass on feedback - all relatively minor :)

webpack/components/extensions/RegistrationCommands/fields/IgnoreSubmanErrors.js
stejskalleos: This is only for subscription-manager errors

webpack/components/extensions/RegistrationCommands/fields/LifeCycleEnvironment.js
stejskalleos: Me neither, according to this [comment](https://github.com/theforeman/foreman/pull/8419#issuecomment-823401017) it should be `host group`, since it's not name or a label for input.

app/controllers/katello/concerns/api/v2/registration_commands_controller_extensions.rb
stejskalleos: Well, since it was already released, we should keep it with flag `deprecated: true`, because removing it could break the workflow for the users who already use the registration API with `string` parameter

stejskalleos: I renamed it to `Activation key for subscription-manager client.`

ares:
this will need to be merged together with the Foreman core PR, merging just the core would result in katello tests broken, @jturel let me please know when you feel this PR is ready, so we can click the green button at once :-)

app/controllers/katello/concerns/api/v2/registration_commands_controller_extensions.rb
jturel: I don't mean to say remove the param altogether, I mean to update its description per my example

jturel (approved):
ACK

Thanks @stejskalleos! Once the Foreman PR is merged I think we can run tests here for a green build and then merge this one?

ares:
core PR merged, let's try this [test katello]

jturel:
[test katello]

jturel:
@stejskalleos looks like the react snapshots need an update then we can merge!

ares:
I'm afraid @stejskalleos is off already, if nightly is broken thanks to this and if that's feasible, can someone else regenerate the snapshots? Otherwise, the tests will remain broken at least until Monday.

PR #9250, opened 2021-03-29
Fixes #32224 - Expose Katello events for Webhooks

ofedoren:
This PR exposes new events for subscriprions within Foreman Webhooks plugin, to test this you might want to have it installed.

@lzap, you might want to test it and probably tell what is missing.

TODO:
 - [x] Add DSL docs
 
 NOTE: I've changed the order of initializers because currently the actions are being loaded after all plugin registration methods are called, which leads to addition of new observable events being impossible. Also `extend_observable_events` is available from `2.4`, so I've updated the required version as well.

### DIFF
--- a/app/lib/actions/katello/content_view/promote.rb
+++ b/app/lib/actions/katello/content_view/promote.rb
@@ -2,6 +2,9 @@ module Actions
   module Katello
     module ContentView
       class Promote < Actions::EntryAction
+        extend ApipieDSL::Class
+        include ::Actions::ObservableAction
+
         def plan(version, environments, is_force = false, description = nil, incremental_update = false)
           action_subject(version.content_view)
           version.check_ready_to_promote!(environments)
@@ -17,6 +20,28 @@ module Actions
               plan_action(Katello::ContentViewVersion::AfterPromoteHook, :id => version.id)
             end
           end
+        end
+
+        def environments
+          input['environments']
+        end
+
+        apipie :class, "A class representing #{self} object" do
+          desc 'This object is available as **@object** variable in
+                webhook templates when a corresponding event occures.
+                The following properties can be used to retrieve the needed information.'
+          name "#{class_scope}"
+          refs "#{class_scope}"
+          sections only: %w[all webhooks]
+          property :task, object_of: 'Task', desc: 'Returns the task to which this action belongs'
+          property :environments, array_of: String, desc: 'Returns the list of environments the content view was promoted to'
+        end
+        include Actions::Katello::JailConcern::Organization
+        include Actions::Katello::JailConcern::ContentView
+        class Jail < ::Actions::ObservableAction::Jail
+          allow :organization_id, :organization_name, :organization_label,
+                :content_view_id, :content_view_name, :content_view_label,
+                :environments
         end
       end
     end
--- a/app/lib/actions/katello/jail_concern/content_view.rb
+++ b/app/lib/actions/katello/jail_concern/content_view.rb
@@ -1,0 +1,30 @@
+module Actions
+  module Katello
+    module JailConcern
+      module ContentView
+        def content_view_id
+          input['content_view']['id']
+        end
+
+        def content_view_name
+          input['content_view']['name']
+        end
+
+        def content_view_label
+          input['content_view']['label']
+        end
+
+        def self.included(base)
+          super
+          base.instance_eval do
+            apipie :class do
+              property :content_view_id, Integer, desc: 'Returns the id of the content view'
+              property :content_view_name, String, desc: 'Returns the name of the content view'
+              property :content_view_label, String, desc: 'Returns the label of the content view'
+            end
+          end
+        end
+      end
+    end
+  end
+end
--- a/app/lib/actions/katello/repository/sync.rb
+++ b/app/lib/actions/katello/repository/sync.rb
@@ -3,8 +3,10 @@ module Actions
   module Katello
     module Repository
       class Sync < Actions::EntryAction
+        extend ApipieDSL::Class
         include Helpers::Presenter
         include Actions::Katello::PulpSelector
+        include ::Actions::ObservableAction
         middleware.use Actions::Middleware::ExecuteIfContentsChanged
 
         input_format do
@@ -123,6 +125,63 @@ module Actions
 
         def rescue_strategy
           Dynflow::Action::Rescue::Skip
+        end
+
+        def repository_id
+          input['repository']['id']
+        end
+
+        def repository_name
+          input['repository']['name']
+        end
+
+        def repository_label
+          input['repository']['label']
+        end
+
+        def product_id
+          input['product']['id']
+        end
+
+        def product_name
+          input['product']['name']
+        end
+
+        def product_label
+          input['product']['label']
+        end
+
+        def contents_changed
+          input['contents_changed']
+        end
+
+        def sync_result
+          input['sync_result']
+        end
+
+        apipie :class, "A class representing #{self} object" do
+          desc 'This object is available as **@object** variable in
+                webhook templates when a corresponding event occures.
+                The following properties can be used to retrieve the needed information.'
+          name "#{class_scope}"
+          refs "#{class_scope}"
+          sections only: %w[all webhooks]
+          property :task, object_of: 'Task', desc: 'Returns the task to which this action belongs'
+          property :repository_id, Integer, desc: 'Returns synced repository id'
+          property :repository_name, String, desc: 'Returns synced repository name'
+          property :repository_label, String, desc: 'Returns synced repository label'
+          property :product_id, Integer, desc: 'Returns product id the synced repository belongs to'
+          property :product_name, String, desc: 'Returns product name the synced repository belongs to'
+          property :product_label, String, desc: 'Returns product label the synced repository belongs to'
+          property :sync_result, Hash, desc: 'Returns Hash object with sync result'
+          property :contents_changed, one_of: [true, false], desc: 'Returns true if repository content was changed due to sync, false otherwise'
+        end
+        include Actions::Katello::JailConcern::Organization
+        class Jail < ::Actions::ObservableAction::Jail
+          allow :organization_id, :organization_name, :organization_label,
+                :repository_id, :repository_name, :repository_label,
+                :product_id, :product_name, :product_label,
+                :sync_result, :contents_changed
         end
       end
     end
--- a/app/models/katello/content_view.rb
+++ b/app/models/katello/content_view.rb
@@ -6,6 +6,7 @@ module Katello
     include Ext::LabelFromName
     include Katello::Authorization::ContentView
     include ForemanTasks::Concerns::ActionSubject
+    include Foreman::ObservableModel
 
     CONTENT_DIR = "content_views".freeze
     IMPORT_LIBRARY = "Import-Library".freeze
@@ -85,6 +86,17 @@ module Katello
     scoped_search :on => :organization_id, :complete_value => true, :only_explicit => true, :validator => ScopedSearch::Validators::INTEGER
     scoped_search :on => :label, :complete_value => true
     scoped_search :on => :composite, :complete_value => true
+
+    set_crud_hooks :content_view
+
+    apipie :class, desc: "A class representing #{model_name.human} object" do
+      name 'Content View'
+      refs 'ContentView'
+      sections only: %w[all additional]
+      prop_group :katello_idname_props, Katello::Model, meta: { friendly_name: 'Content View' }
+      property :label, String, desc: 'Returns label of the Content View'
+      property :organization, 'Organization', desc: 'Returns organization object'
+    end
 
     def self.in_environment(env)
       joins(:content_view_environments).
@@ -649,14 +661,8 @@ module Katello
       self.organization
     end
 
-    apipie :class, desc: "A class representing #{model_name.human} object" do
-      name 'Content View'
-      refs 'ContentView'
-      sections only: %w[all additional]
-      prop_group :katello_basic_props, Katello::Model, meta: { friendly_name: 'Content View' }
-    end
     class Jail < ::Safemode::Jail
-      allow :name, :label, :version
+      allow :id, :name, :label, :version, :organization
     end
   end
 end
--- a/lib/katello/engine.rb
+++ b/lib/katello/engine.rb
@@ -79,16 +79,6 @@ module Katello
       require 'katello/apipie/validators'
     end
 
-    # make sure the Katello plugin is initialized before `after_initialize`
-    # hook so that the resumed Dynflow tasks can rely on everything ready.
-    initializer 'katello.register_plugin', :before => :finisher_hook, :after => 'foreman_remote_execution.register_plugin' do
-      ::Foreman::AccessControl::Permission.prepend ::Katello::Concerns::PermissionExtensions
-      require 'katello/plugin'
-
-      # extend builtin permissions from core with new actions
-      require 'katello/permissions'
-    end
-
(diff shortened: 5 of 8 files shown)


lzap:
@jjeffers what are our chances getting this into 4.0? cheers :-)

jjeffers:
> @jjeffers what are our chances getting this into 4.0? cheers :-)
> 
> I will be testing this tomorrow.

Probably low chance, since we just built rc3.1. The next release should be the GA.

lzap:
Yeah that's what I thought. I was also hoping the patch would be even smaller, it looks like an initializer change is needed. Okay.

lib/katello/engine.rb
```diff
@@ -99,6 +89,17 @@ class Engine < ::Rails::Engine
...
+    # hook so that the resumed Dynflow tasks can rely on everything ready.
+    initializer 'katello.register_plugin', :before => :finisher_hook, :after => 'foreman_remote_execution.register_plugin' do
+      ::Foreman::AccessControl::Permission.prepend ::Katello::Concerns::PermissionExtensions
+      require 'katello/plugin'
+
+      # extend builtin permissions from core with new actions
+      require 'katello/permissions'
     end
 
     initializer "katello.set_dynflow_middlewares", :before => :finisher_hook do |_app|
```
lzap, line 105: I am not sure I understand the consequences of this, but since I'd like to ask for minor release inclusion for this patch I wonder if @adamruzicka would be able to confirm this is safe to do.

app/lib/actions/katello/actions_jail.rb
```diff
@@ -0,0 +1,33 @@
...
+      def organization_label
+        input['organization']['label']
+      end
+      class Jail < ::Actions::ObservableAction::Jail
+        allow :content_view_id, :content_view_name, :content_view_label,
+              :organization_id, :organization_name, :organization_label
+      end
+    end
+  end
+end
```
lzap, line 33: I am wondering how a concern will work in this case, class Jail is defined via the concern and then redefined later. I guess I need to test it.

By the way is the reason you inherit `::Actions::ObservableAction::Jail` instead of `Safemode::Jail` to also have `task` allowed?

lzap:
Nicely done, can you add those DSL docs? I am starting with testing right away. I was hoping that we only add a single line into plugin.rb that initializer reordering is unfortunate but I can test some edge-case scenarios (e.g. dynflow resume?).

lib/katello/engine.rb
adamruzicka: This looks reasonable to me and I don't see any immediate footguns. @jlsherrill anything I'm overlooking?

ofedoren:
@lzap, working on the docs, I've updated the PR so it's testable with Safemode enabled (I forgot that when I was writing this I had that disabled)

app/lib/actions/katello/actions_jail.rb
ofedoren: Well, I guess for actions we want have `task` object available to get some info. The first action that has that is `host_run_job_succeeded`, so I thought it might be useful for all the actions, thus I inherit `::Actions::ObservableAction::Jail`.

Also, I've updated the way the `Jail` gets possibility to be re-opened after the module is included.

app/lib/actions/katello/actions_jail.rb
```diff
@@ -0,0 +1,40 @@
...
+      def self.included(base)
+        super
+        base.instance_eval do
+          jail_class = Class.new(::Actions::ObservableAction::Jail) do
+            allow :content_view_id, :content_view_name, :content_view_label,
+                  :organization_id, :organization_name, :organization_label
+          end
+          base.const_set('Jail', jail_class)
+        end
+      end
```
lzap, line 37: I am just wondering if it is not cleaner just to define attributes in this module and simply repeating all allowed attributes for each class. But I am okay with this.

ofedoren: Sure it would be cleaner than some magic Ruby-woodoo, but repeating is not so good as well since one might forget to update allowed methods somewhere else.

lzap:
Rubocop failures.

ofedoren:
@lzap, updated. Test failures don't seem to be related..

app/lib/actions/katello/actions_common_jail.rb
```diff
@@ -0,0 +1,62 @@
...
+              allow :content_view_id, :content_view_name, :content_view_label,
+                    :organization_id, :organization_name, :organization_label
+            end
+          end
+          base.const_set('Jail', jail_class)
+        end
+      end
+    end
+  end
+end
```
lzap, line 62: This really calls for breaking up into two concerns: `ActionsContentViewJailConcern` and `ActionsOrganizationJailConcern`.

lzap (changes requested):
Other than that, I am good.

ofedoren:
@lzap, updated. I've renamed those a bit to have a structure.

Just a reminder for reviewers: if you want to test docs locally as well:

`FOREMAN_APIPIE_LANGS=en bundle exec rake apipie_dsl:cache` within `foreman` directory.

lzap (approved):
Thank you I am good.

To Katello reviewers: This patch introduces Jail support for Safemode, in order to expose methods they must be explicitly listed. We tried to find minimum amount of methods which gives reasonable power to webhooks users. We have landed on this rather flat list of methods because the dynflow actions carries the data in a hash.

app/lib/actions/katello/jail_concern/content_view.rb
```diff
@@ -0,0 +1,41 @@
...
+            jail_class = Class.new(::Actions::ObservableAction::Jail) do
+              allow :content_view_id, :content_view_name, :content_view_label
+            end
+            base.const_set('Jail', jail_class)
+          end
+        end
+      end
+    end
+  end
+end
```
lzap, line 41: I think I found a bug tho, I am getting safemode error for Content View create event, I think this concern is not included in the ContentView model at all. It is not a Dynflow action but ActiveRecord event as well, so it should be just a simple Jail.

```
2021-04-06T12:24:34 [I|app|06a8fc9e] ForemanWebhooks::EventSubscriber: content_view_created.event.foreman event received
2021-04-06T12:24:34 [E|app|06a8fc9e] Safemode::NoMethodError: undefined method '#content_view_id' for Katello::ContentView::Jail (Katello::ContentView)
```

app/models/katello/content_view.rb
```diff
@@ -86,6 +87,8 @@ class ContentView < Katello::Model
     scoped_search :on => :label, :complete_value => true
     scoped_search :on => :composite, :complete_value => true
 
+    set_crud_hooks :content_view
+
     def self.in_environment(env)
       joins(:content_view_environments).
         where("#{Katello::ContentViewEnvironment.table_name}.environment_id = ?", env.id)
```
lzap, line 94: I think organization and location (name,label,id) is worth exposing too for CV.

app/lib/actions/katello/jail_concern/content_view.rb
ofedoren: Well, for content view object it should be just `id`, not the `content_view_id`. There is (or at least should be, I don't remember) documentation for Content View class in `all`/`additional` section in DSL docs.

lzap: So it is used in other Dynflow action(s) but in Content View (CRUD) the Jail is missing, so actually here it is correct.

app/models/katello/content_view.rb
lzap: We are missing Jail for this model here.

lzap (changes requested):
Actually one missing jail.

ofedoren:
@lzap, updated. There was Jail for content views already, I've just updated it with `id` and `organization`.

lzap:
> @lzap, updated. There was Jail for content views already, I've just updated it with `id` and `organization`.

Oh my bad I started testing with `id` and I assumed there was no Jail class. Name would be maybe enough, but this is good, let me retest.

app/lib/actions/katello/content_view/promote.rb
```diff
@@ -2,6 +2,11 @@ module Actions
   module Katello
     module ContentView
       class Promote < Actions::EntryAction
+        extend ApipieDSL::Class
+        include ::Actions::ObservableAction
+        include Actions::Katello::JailConcern::Organization
+        include Actions::Katello::JailConcern::ContentView
```
lzap, line 8: Just for the record, this does not work atm, organization jail is overwritten in the include block by content view.

lzap (changes requested):
We are close, it does not work yet.

app/lib/actions/katello/content_view/promote.rb
lzap: Since the Jail.allow method concatenates to a list, you can call it multiple times. So I suggest you assume Jail already exists, move these includes down after Jail class definition and append from the included block.

ofedoren:
@lzap, thanks for noticing that `const_set` will replace the `Jail` with the second `include`. I've played a little bit more with class reopening and noticed that what works with plain Ruby in `irb` doesn't work in Rails. Well, even if it worked I agree that this complicates simple things and might break something in the future.

I've returned the simple way for `Jail`s and left mixins only for their main purpose.

app/lib/actions/katello/content_view/promote.rb
ofedoren: I've tried that one, but when I try something like `base::Jail` or `base.const_get(:Jail)` it returns an object that doesn't understand `.inspect` even.

lzap (approved):
Everything works as expected. Can we have somebody looking on this PR, there are many comments already so folks might be skipping it. @ehelms @jlsherrill thanks.

lib/katello/engine.rb
```diff
@@ -99,6 +89,17 @@ class Engine < ::Rails::Engine
...
+    end
+
+    # make sure the Katello plugin is initialized before `after_initialize`
+    # hook so that the resumed Dynflow tasks can rely on everything ready.
+    initializer 'katello.register_plugin', :before => :finisher_hook, :after => 'foreman_remote_execution.register_plugin' do
+      ::Foreman::AccessControl::Permission.prepend ::Katello::Concerns::PermissionExtensions
+      require 'katello/plugin'
+
+      # extend builtin permissions from core with new actions
+      require 'katello/permissions'
```
ehelms, line 102: I am confused by the moving around of this code. It appears to not change anything other than where it appears in the file?

app/lib/actions/katello/repository/sync.rb
```diff
@@ -124,6 +126,62 @@ def regenerate_applicability(repo, contents_changed)
...
+        end
+
+        def contents_changed
+          input['contents_changed']
+        end
+
+        def sync_result
+          input['sync_result']
+        end
+        apipie :class, "A class representing #{self} object" do
```
ehelms, line 161: Is this tied to `def sync_result` ? Otherwise, a blank line above this would help.

app/lib/actions/katello/repository/sync.rb
```diff
@@ -124,6 +126,62 @@ def regenerate_applicability(repo, contents_changed)
...
+          property :sync_result, Hash, desc: 'Returns Hash object with sync result'
+          property :contents_changed, one_of: [true, false], desc: 'Returns true if repository content was changed due to sync, false otherwise'
+        end
+        include Actions::Katello::JailConcern::Organization
+        class Jail < ::Actions::ObservableAction::Jail
+          allow :organization_id, :organization_name, :organization_label,
+                :repository_id, :repository_name, :repository_label,
+                :product_id, :product_name, :product_label,
+                :sync_result, :contents_changed
+        end
```
ehelms, line 184: Do I understand right that the `apipie` definition, the include JailConcern and the class Jail declaration are all tied to the same concept? Events?

ehelms:
> To Katello reviewers: This patch introduces Jail support for Safemode, in order to expose methods they must be explicitly listed. We tried to find minimum amount of methods which gives reasonable power to webhooks users. We have landed on this rather flat list of methods because the dynflow actions carries the data in a hash.

Is the addition of Jail support for Safemode directly tied to Webhooks support? Is that something that is a pre-requisite? Does Jail support without "events" make sense for Katello? One thing I am trying to understand from this while looking over the code is as a developer what parts of the changes are what I need to add to produce events? (and which parts are not 'directly' required)

lib/katello/engine.rb
ofedoren: Yes. From that perspective there are no changes at all. But the order does affects if the needed constants are available in plugin registration block.

app/lib/actions/katello/repository/sync.rb
ofedoren: Noticed, I've added missing blank lines.

app/lib/actions/katello/repository/sync.rb
ofedoren: Yes, for `actions` these are needed to fully support events in Foreman Webhooks that are being fired up if e.g. repo sync has happened.

`apipie` block is for documentation; JailConcern with public methods to use in webhook templates; Jails are to make those methods available in safe mode.

ofedoren:
@ehelms, short answer would be:

If a developer wants to expose events to be subscribed after then simple `include Foreman::ObservableModel` for models or `include ::Actions::ObservableAction` for actions should be enough. Other additions are related, but not required. Although without them there is no much of usage from the user end perspective since everything will be internal.

app/lib/actions/katello/repository/sync.rb
ehelms: Thanks. It seems like it would be useful to group all of the pieces related to webhooks together somehow but I'll let Katello core maintainers weigh in on the overall code structure. I appreciate the helpful details understanding. I was worried that it would be hard over time to track what is required to ensure Webhooks work (and subsequently do not break in the future) and what aspects are unrelated. Also, how to help developers who want to add additinal webhook "hooks" into the code know what to add.

lzap:
I can confirm what @ofedoren said, making events observable via foreman_webhooks is really that simple and we have documented this in the devel docs: https://github.com/theforeman/foreman/blob/develop/developer_docs/how_to_create_a_plugin.asciidoc#adding-subscribers

However if you turn on safemode, then the data coming in the webhook templates will be useless because there will be no way of getting any data out of it. Therefore the Jail and apidocs portions of the patch.

jlsherrill:
Wanted to update, i'm in the middle of testing this.  Overall looks good, will update again soon.

jlsherrill (approved):
ACK, tested the events and they all worked, rendering the template i created in each case.  I'm okay with the overall structure of the code here

lzap:
Thanks, I asked for 4.0.1 in the issue.

lzap:
I see Katello 4.0.1 being pushed to rubygems. I did set target release in RM to at least have a conversation about possibility of this patch to be cherry picked. Now I've sent my presentation recording to Foreman Birthday party where I announce Foreman Webhooks with Katello 4.0 support :-(

https://projects.theforeman.org/versions/1369

Should I move the target release to 4.0.2 with hope it would make the cut? :-)

ofedoren:
@lzap isn't it already there? https://github.com/Katello/katello/commits/KATELLO-4.0 / https://github.com/Katello/katello/compare/4.0.0...4.0.1

jjeffers:
> @lzap isn't it already there? https://github.com/Katello/katello/commits/KATELLO-4.0 / [4.0.0...4.0.1](https://github.com/Katello/katello/compare/4.0.0...4.0.1)

It should be in there - please see https://github.com/Katello/katello/pull/9359

lzap:
Holy smokes, I think I've used all of them. THANKS! :-)

PR #9251, opened 2021-03-29
Fixes #32221 - use eslint react-hooks rules

jeremylenz:
Foreman was enforcing [react-hooks rules](https://reactjs.org/docs/hooks-faq.html#what-exactly-do-the-lint-rules-enforce) (rules-of-hooks and exhaustive-deps) but Katello currently is not.

This change adds those rules to Katello's eslint configuration, and makes the necessary changes to follow them.

Lint warnings generated (and fixed) here include ones like:
```
/home/vagrant/katello/webpack/components/EditableTextInput/EditableTextInput.js
  36:6  warning  React Hook useEffect has missing dependencies: 'attribute', 'inputValue', and 'onEdit'. Either include them or remove the dependency array. If 'onEdit' changes too often, find the parent component that defines it and wrap that definition in useCallback  react-hooks/exhaustive-deps

/home/vagrant/katello/webpack/components/Table/TableWrapper.js
  54:20  warning  React Hook useEffect has a spread element in its dependency array. This means we can't statically verify whether you've passed the correct dependencies  react-hooks/exhaustive-deps

/home/vagrant/katello/webpack/scenes/ContentViews/Copy/CopyContentViewForm.js
  30:7  warning  React Hook useEffect has a complex expression in the dependency array. Extract it to a separate variable so it can be statically checked  react-hooks/exhaustive-deps

/home/vagrant/katello/webpack/scenes/ContentViews/Publish/CVPublishFinish.js
  42:9  warning  The 'pollPublishTask' function makes the dependencies of useDeepCompareEffect Hook (at line 92) change on every render. Move it inside the useDeepCompareEffect callback. Alternatively, wrap the definition of 'pollPublishTask' in its own useCallback() Hook  react-hooks/exhaustive-deps

  48:9  warning  The 'handleEndTask' function makes the dependencies of useDeepCompareEffect Hook (at line 109) change on every render. To fix this, wrap the definition of 'handleEndTask' in its own useCallback() Hook                                                         react-hooks/exhaustive-deps


```

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "eslint-plugin-jsx-a11y": "^6.0.2",
     "eslint-plugin-promise": "^4.2.1",
     "eslint-plugin-react": "^7.4.0",
+    "eslint-plugin-react-hooks": "^4.2.0",
     "jest": "^24.9.0",
     "nock": "^12.0.3",
     "react-test-renderer": "^16.0.0"
--- a/webpack/components/Table/TableWrapper.js
+++ b/webpack/components/Table/TableWrapper.js
@@ -30,31 +30,37 @@ const TableWrapper = ({
   const [total, setTotal] = useState(0);
 
   const updatePagination = (data) => {
+    console.log('updatePagination')
     const { subtotal: newTotal, page: newPage, per_page: newPerPage } = data;
     if (newTotal !== undefined) setTotal(parseInt(newTotal, 10));
     if (newPage !== undefined) setPage(parseInt(newPage, 10));
     if (newPerPage !== undefined) setPerPage(parseInt(newPerPage, 10));
   };
   // eslint-disable-next-line no-underscore-dangle
-  const _paginationParams = () => ({ per_page: perPage, page });
-  const paginationParams = useCallback(_paginationParams, [page, perPage]);
-
-  useEffect(() => updatePagination(metadata), [metadata]);
+  const paginationParams = () => ({ per_page: perPage, page });
+  // const paginationParams = useCallback(_paginationParams, [page, perPage]);
+let counter = React.useRef(1);
+  // useEffect(() => updatePagination(metadata), [metadata]);
 
   // The search component will update the search query when a search is performed, listen for that
   // and perform the search so we can be sure the searchQuery is updated when search is performed.
   useEffect(() => {
     const fetchWithParams = (allParams = {}) => {
-      dispatch(fetchItems({ ...paginationParams(), ...allParams }));
+      dispatch(fetchItems({ per_page: perPage, page, ...allParams }));
     };
     if (searchQuery || activeFilters) {
       // Reset page back to 1 when filter or search changes
+      console.log('fetchWithParams with searchQuery')
       fetchWithParams({ search: searchQuery, page: 1 });
     } else {
-      fetchWithParams();
+      console.log('fetchWithParams')
+      counter.current += 1;
+      // if (counter.current > 10) debugger;
+      // fetchWithParams();
+      if (counter.current < 10) fetchWithParams();
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [searchQuery, ...additionalListeners, activeFilters, dispatch, fetchItems, paginationParams]);
+  }, [searchQuery, ...additionalListeners, activeFilters, dispatch, fetchItems, page, perPage]);
 
   const getAutoCompleteParams = search => ({
     endpoint: autocompleteEndpoint,
--- a/webpack/components/Table/TableWrapper.js
+++ b/webpack/components/Table/TableWrapper.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useState, useCallback } from 'react';
+import React, { useEffect, useState } from 'react';
 import { Pagination, Flex, FlexItem } from '@patternfly/react-core';
 
 import PropTypes from 'prop-types';
@@ -30,7 +30,6 @@ const TableWrapper = ({
   const [total, setTotal] = useState(0);
 
   const updatePagination = (data) => {
-    console.log('updatePagination')
     const { subtotal: newTotal, page: newPage, per_page: newPerPage } = data;
     if (newTotal !== undefined) setTotal(parseInt(newTotal, 10));
     if (newPage !== undefined) setPage(parseInt(newPage, 10));
@@ -39,7 +38,7 @@ const TableWrapper = ({
   // eslint-disable-next-line no-underscore-dangle
   const paginationParams = () => ({ per_page: perPage, page });
   // const paginationParams = useCallback(_paginationParams, [page, perPage]);
-let counter = React.useRef(1);
+  const counter = React.useRef(1);
   // useEffect(() => updatePagination(metadata), [metadata]);
 
   // The search component will update the search query when a search is performed, listen for that
@@ -50,10 +49,8 @@ let counter = React.useRef(1);
     };
     if (searchQuery || activeFilters) {
       // Reset page back to 1 when filter or search changes
-      console.log('fetchWithParams with searchQuery')
       fetchWithParams({ search: searchQuery, page: 1 });
     } else {
-      console.log('fetchWithParams')
       counter.current += 1;
       // if (counter.current > 10) debugger;
       // fetchWithParams();
--- a/webpack/components/Table/TableWrapper.js
+++ b/webpack/components/Table/TableWrapper.js
@@ -55,7 +55,7 @@ const TableWrapper = ({
   }, [
     activeFilters,
     dispatch,
-    // fetchItems,
+    fetchItems,
     paginationParams,
     searchQuery,
     additionalListeners,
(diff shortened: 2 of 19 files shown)


jeremylenz:
For places where we use `JSON.stringify(response)` in the dependency array, we will get this:
```
 76:7  warning  React Hook useEffect has a complex expression in the dependency array. Extract it to a separate variable so it can be statically checked   react-hooks/exhaustive-deps
```
Provided that no other eslint rules are broken on those lines, I'm adding `// eslint-disable-line react-hooks/exhaustive-deps` for now.  We probably should find a better solution in the future.

jeremylenz:
I think this will need an ack from @theforeman/packaging since we modify `package.json`

webpack/components/Table/TableWrapper.js
```diff
@@ -35,23 +36,30 @@ const TableWrapper = ({
...
       // Reset page back to 1 when filter or search changes
       fetchWithParams({ search: searchQuery, page: 1 });
     } else {
       fetchWithParams();
     }
-  }, [searchQuery, ...additionalListeners]);
+  }, [
+    activeFilters,
+    dispatch,
+    // fetchItems,
```
jeremylenz, line 58: still working on this one..

jeremylenz: Updated. All tests & lints passing now :angel:

package.json
```diff
@@ -38,6 +38,7 @@
     "eslint-plugin-jsx-a11y": "^6.0.2",
     "eslint-plugin-promise": "^4.2.1",
     "eslint-plugin-react": "^7.4.0",
+    "eslint-plugin-react-hooks": "^4.2.0",
```
tbrisker, line 41: could we use [eslint-plugin-foreman](https://github.com/theforeman/foreman-js/tree/master/packages/eslint-plugin-foreman) which already has this and some of the other packages, as well as some basic rules config? 
If there are rules missing there that katello uses it would be great to add there and try to unify the linting as much as possible across the whole ecosystem.

jeremylenz: @tbrisker This sounds fine, however I tried adding `eslint-plugin-foreman` and ran `npx eslint webpack/`.  I ran into the following:

1. It still complained about not being able to find `eslint-plugin-react-hooks`.  So I added that back in to the package.json
2. Now `eslint` complains about not being able to find module `prettier`. I see that Foreman's package.json requires `prettier`, so I also add it to Katello.  Now I get:

```
✖ 3534 problems (3534 errors, 0 warnings)
  3531 errors and 0 warnings potentially fixable with the `--fix` option.
```

Considering the huge number of changes that `prettier` will require in Katello, I think it's better to do it this way:

1. Keep `eslint-plugin-react-hooks` here
2. In a separate PR, include both the addition of `eslint-plugin-foreman` and `prettier` (cc @Andrewgdewar).   Thoughts?

jeremylenz: side note: I also noticed that the version of eslint-plugin-react-hooks used in `eslint-plugin-foreman` is out of date:
```
"eslint-plugin-react-hooks": "^2.1.1"
```

jeremylenz:
Rebased after merge of https://github.com/Katello/katello/pull/9457

webpack/scenes/ContentViews/Details/Histories/ContentViewHistories.js
```diff
@@ -46,6 +38,14 @@ const ContentViewHistories = ({ cvId }) => {
...
 
   useDeepCompareEffect(() => {
+    const taskTypes = {
+      publish: 'Actions::Katello::ContentView::Publish',
+      promotion: 'Actions::Katello::ContentView::Promote',
+      removal: 'Actions::Katello::ContentView::Remove',
+      incrementalUpdate: 'Actions::Katello::ContentView::IncrementalUpdates',
+      export: 'Actions::Katello::ContentViewVersion::Export',
+    };
+
```
sjha4, line 48: Unrelated to the whole hooks rule but would it make more sense to have this in the Constants file?

webpack/scenes/ContentViews/Details/Versions/ContentViewVersions.js
```diff
@@ -96,7 +96,7 @@ const ContentViewVersions = ({ cvId }) => {
       cells={columnHeaders}
       variant={TableVariant.compact}
       autocompleteEndpoint={`/content_view_versions/auto_complete_search?content_view_id=${cvId}`}
-      fetchItems={params => getContentViewVersions(cvId, params)}
+      fetchItems={useCallback(params => getContentViewVersions(cvId, params), [cvId])}
```
sjha4, line 99: Do we also need this getContentViewVersions in the Dependency array here?

jeremylenz: It's defined outside of the component, so it's not needed in the dependency array according to the eslint rule.

webpack/scenes/ContentViews/Details/Histories/ContentViewHistories.js
jeremylenz: I can see that, but I feel like it makes sense to wait until it's used in multiple places.

jeremylenz:
From IRC:

> can you highlight some of the functionality that needs regression testing? it's not totally apparrent to me just by looking at the diff

The changes here primarily affect useEffect dependency arrays, so they change whether an effect is run. Effects can contain React state updates or API calls.  If a regression were introduced, it would be causing the effect to run either more or less often than it should.  

Hopefully our React testing library tests will catch regressions.  For example, one change earlier in this PR was causing two API requests to fire every time I clicked the next page button, when there should be only one.  The test did its job and failed spectacularly with nock matching errors.

As far as manual regression testing, I can't think of any particular testing needed other than to click around the new content view pages and perform as many actions as you can.

package.json
Andrewgdewar: Sounds like a plan.

jturel (commented):
This LGTM considering:

- my limited react experience
- the build is green
- the changes are made in a consistent way without any 'bonus' or confusing changes
- testing in the areas based on filename didn't expose any problems _except_ on the /hosts/register page:
```
react-dom.development.js:23093 Uncaught Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
    at checkForNestedUpdates (react-dom.development.js:23093)
    at scheduleUpdateOnFiber (react-dom.development.js:21164)
    at dispatchAction (react-dom.development.js:15660)
    at handlePluginValue (index.js:98)
    at index.js:20
    at commitHookEffectListMount (react-dom.development.js:19731)
    at commitPassiveHookEffects (react-dom.development.js:19769)
    at HTMLUnknownElement.callCallback (react-dom.development.js:188)
    at HTMLUnknownElement.dispatchEvent (<anonymous>)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:237)
```

jturel:
Merging this will warrant a kick of green PRs containing changes in the react code to make sure they're adhering to the new rules lest we merge them and find lint errors in subsequent jenkins runs

webpack/scenes/ContentViews/Details/Histories/ContentViewHistories.js
Andrewgdewar: I feel the same way, but to be fair, I don't like that we have constants inside of our react components, I feel they should be placed above the component to prevent clutter. (I do understand the arguement of scope, but its the same memory footprint.

jeremylenz:
Great catch @jturel!

This was a tricky one.

Turns out the `onChange` in the dependency array [here](https://github.com/Katello/katello/blob/bd611d9b4f8b38b48ffccde52d9c6ac592ba4bfe/webpack/components/extensions/RegistrationCommands/index.js#L21) was changing on every render, causing the effect to run infinitely.  So I went to apply my usual `useCallback` trick. But where is that function defined?

Turns out it's defined in Foreman thru Slot & Fill: https://github.com/theforeman/foreman/blob/71f97377b29b121dd05c84ee9993a847842e31ec/webpack/assets/javascripts/react_app/routes/RegistrationCommands/RegistrationCommandsPage/index.js#L262

So our `onChange` is actually `handlePluginValue`, which is also in Foreman.

I also found that adding `handleInvalidField` [here](https://github.com/Katello/katello/blob/bd611d9b4f8b38b48ffccde52d9c6ac592ba4bfe/webpack/components/extensions/RegistrationCommands/fields/ActivationKeys.js#L43) causes a similar problem.

I've opened https://github.com/theforeman/foreman/pull/8699 which addresses both in Foreman.

For the time being, I've removed the 'problematic' dependencies and disabled the rule in those two spots here.  But when the Foreman change is merged, we can reenable it.

sjha4:
I have played around with CV workflows on this branch and I didn't run into any issues. Changes look good to me. 👍🏽 Will kick tests on open PRs and bug you if there are things that need fixing..😄

jturel:
> I've opened [theforeman/foreman#8699](https://github.com/theforeman/foreman/pull/8699) which addresses both in Foreman.
> 
> For the time being, I've removed the 'problematic' dependencies and disabled the rule in those two spots here. But when the Foreman change is merged, we can reenable it.

Nice fix & that totally works for me

jturel (approved):
ACK! Great stuff

PR #9253, opened 2021-03-30
Fixes #32230 - fix SPC sync with pulp2 for container

jlsherrill:
rails appends charset=utf-8 to the CONTENT_TYPE header of all requests
by default.  This overrides that behavior as pulp2 does not like it
when syncing.


jjeffers (approved):
Test this change on production 3.18 katello and a 3.18 production foreman proxy with pulp2 backend. I was able to reproduce the error syncing content to the capsule. 

After applying the patch the 'fsLayers' issue was resolved.

PR #9255, opened 2021-03-30
Fixes #17423,#32234 - support sync timeout & pulp 3.11

jlsherrill:


app/lib/actions/pulp3/orchestration/content_view_version/export.rb
```diff
@@ -0,0 +1,102 @@
...
+                content_view_version_id: input[:content_view_version_id],
+                destination_server: input[:destination_server],
+                path: path,
+                metadata: export_metadata
+            )
+            output[:export_history_id] = history.id
+            output[:path] = path
+          end
+
+          def humanized_name
```
parthaa, line 91: This should remain in the parent action?

app/lib/actions/pulp3/orchestration/content_view_version/export.rb
```diff
@@ -0,0 +1,102 @@
...
+            _("Export")
+          end
+
+          def rescue_strategy
```
parthaa, line 95: This should remain in the parent action?

app/lib/actions/katello/content_view_version/export.rb
```diff
@@ -20,50 +20,21 @@ def plan(content_view_version:, destination_server: nil,
...
-              smart_proxy_id: smart_proxy.id,
-              exporter_data: action_output[:exporter_data])
-            plan_self(exporter_data: action_output[:exporter_data], smart_proxy_id: smart_proxy.id,
-              destination_server: destination_server,
-              content_view_version_id: content_view_version.id,
-              from_content_view_version_id: from_content_view_version&.id,
-              export_history_id: history_output[:export_history_id],
-              exported_file_checksum: history_output[:exported_file_checksum],
-              path: history_output[:path])
+            export_output = plan_action(::Actions::Pulp3::Orchestration::ContentViewVersion::Export,
```
parthaa, line 27: wonder if you can just capture all the plan params to an options hash and pass that to plan action.
That way it becomes amply clear that this is like a delegating action

app/lib/actions/pulp3/orchestration/content_view_version/export.rb
```diff
@@ -0,0 +1,102 @@
...
+
+          output_format do
+            param :exported_file_checksum, Hash
+            param :export_path, String
+          end
+          def plan(content_view_version:, destination_server: nil,
+                   chunk_size: nil, from_history: nil,
+                   validate_incremental: true,
+                   fail_on_missing_content: false)
+            action_subject(content_view_version)
```
parthaa, line 23: dont need this right since the parent action is doing it?

app/lib/actions/pulp3/orchestration/content_view_version/export.rb
```diff
@@ -0,0 +1,102 @@
...
+            unless File.directory?(Setting['pulpcore_export_destination'])
```
parthaa, line 24: I think @jeremylenz changed this part. Might want to get the latest copy of the parent action from master and then port appropriately

app/lib/actions/pulp3/orchestration/content_view_version/export.rb
jlsherrill: yeah, i think so

app/lib/actions/pulp3/orchestration/content_view_version/export.rb
jlsherrill: nope!

app/lib/actions/pulp3/orchestration/content_view_version/export.rb
```diff
@@ -0,0 +1,86 @@
+module Actions
```
jeremylenz, line 1: We moved these `pulp3/orchestration` actions back into the Katello namespace recently - for example, this one moved to `Actions::Katello::ContentViewVersion::Export` - because `pulp3` doesn't make as much sense when there's no more pulp2.  (I think it was @jturel suggestion) But I wouldn't want to see it moved back without a good reason..

jlsherrill: So the reason i suggested changing this is that we could run:
```
rake test:katello:test:pulp3 
```
to have all the vcr tests against pulp3 get run.  It did this by looking in all the pulp3 paths: 
https://github.com/Katello/katello/blob/master/lib/katello/tasks/test.rake#L158-L159

i discussed with partha about this and suggested we could either move the orchestration here as i've done and the test, or just move the test, and he was more in favor of moving them both.  I'm open to other suggestions

jturel: I like what was done on this PR - it makes later refactoring easier to have Actions::Katello::ContentViewVersion::Export more or less a generic shell that starts the (as of today, but maybe not some day in the future) pulp3 orchestration off.

app/lib/actions/pulp3/orchestration/content_view_version/export.rb
jlsherrill: correct!

ianballou (approved):
I'm cool with this from the sync timeout and Pulp 3.11 standpoint. It sounds like the other reviewers of this PR are happy with the import/export changes too?

jeremylenz (approved):
LGTM :+1:  remember to squash!

jlsherrill:
I'm actually not gonna squash because i think all 3 of these commits have value on their own.  thanks guys!

PR #9258, opened 2021-03-31
Fixes #31704 - Module stream copying tests need to be un-skipped

ianballou:
Tests un-skipped and new cassettes recorded.


parthaa (approved):
APJ

PR #9260, opened 2021-03-31
Fixes #32241 - Auto create CV on import

parthaa:
Automatically create a content view if one with the same name does not
exist in this organization. The name of the content view is provided in
the import metadata.

### DIFF
--- a/app/controllers/katello/api/v2/content_imports_controller.rb
+++ b/app/controllers/katello/api/v2/content_imports_controller.rb
@@ -1,6 +1,8 @@
 module Katello
   class Api::V2::ContentImportsController < Api::V2::ApiController
-    before_action :find_publishable_content_view, :only => [:version]
+    before_action :find_organization, :only => [:version]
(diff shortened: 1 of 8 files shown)


parthaa:
As steps to test

1. create a content view in the upstream or exporting org
2. Add a few immediate repos 
3. Publish
4. Export the the content view version via hammer
5. Then in the importing org create shell repos  with the same product name. Dont create content view
6. import via `hammer content-import version --organization=downstream --path=<> ` 
7. Notice that a  content view is created in the import org automatically with the appropriate repos and version imported
8. Notice that the library repositories are updated..

chris1984 (approved):
APJ, works well. Using hammer I was able to have the cv created for me with only just making the repos/product:

```bash
$ hammer content-import version --organization=import --path /var/lib/pulp/exports/Default_Organization/view/1.0/2021-04-05T15-51-44-04-00
[...............................................................................................................................] [100%]
```

PR #9261, opened 2021-04-01
Fixes #31867 - updates references to content to /pulp/content

jjeffers:


app/models/katello/host/content_facet.rb
```diff
@@ -32,10 +32,10 @@ class ContentFacet < Katello::Model
       validates_with Validators::ContentViewEnvironmentValidator
 
       def update_repositories_by_paths(paths)
-        prefixes = %w(/pulp/deb/ /pulp/repos/ /pulp/content/)
+        prefixes = %w(/pulp/content/)
```
jlsherrill, line 35: we still want to support clients fetching content on the old paths, while advertising the new paths, so this needs to stay

app/helpers/katello/katello_urls_helper.rb
```diff
@@ -39,11 +39,10 @@ def subscription_manager_configuration_url(host = nil, rpm = true, hostname: nil
 
     apipie :method, 'Generates an absolute path to the file' do
       required :content_path, String, desc: "Relative path to the file or it's name"
-      optional :content_type, String, desc: 'Content type', default: 'repos'
       optional :schema, String, desc: 'Optional URL schema for the content source', default: 'http'
       returns String, desc: 'Absolute path to a file'
     end
-    def repository_url(content_path, content_type = 'repos', schema = 'http')
+    def repository_url(content_path, schema = 'http')
```
jlsherrill, line 45: this is a good change, but would require an update in a couple places in foreman:  

$ git grep repository_url\(
app/views/unattended/provisioning_templates/provision/kickstart_default.erb:    repository_url(img_name, 'isos')
app/views/unattended/provisioning_templates/provision/kickstart_ovirt.erb:  liveimg_url = repository_url(liveimg_name, 'isos')

I wonder if it would make sense to leave the variable and just not use it:  "_content_type = nil"  at least for a while

app/models/katello/host/content_facet.rb
jjeffers: Thank you!

app/helpers/katello/katello_urls_helper.rb
jjeffers: I will open a follow on foreman PR later, after this is merged. But let's go with the default value as nil for content_type. I restored the optional parameter too until foreman has been changed.

app/models/katello/glue/pulp/repo.rb
```diff
@@ -361,16 +361,10 @@ def full_path(smart_proxy = nil, force_https = false)
...
         "#{pulp_uri.host.downcase}/#{container_repository_name}"
-      elsif file?
-        "#{scheme}://#{pulp_uri.host.downcase}/pulp/isos/#{relative_path}/"
       elsif ostree?
-        "#{scheme}://#{pulp_uri.host.downcase}/pulp/ostree/web/#{relative_path}"
-      elsif deb?
-        "#{scheme}://#{pulp_uri.host.downcase}/pulp/deb/#{relative_path}/"
-      elsif ansible_collection?
-        "#{scheme}://#{pulp_uri.host.downcase}/pulp/content/#{relative_path}/"
+        "#{scheme}://#{pulp_uri.host.downcase}/pulp/content/web/#{relative_path}"
```
jlsherrill, line 365: @sjha4 would it make sense to show a url of   `https://HOSTNAME/pulp_ansible/galaxy/#{relative_path}/`  here?

jlsherrill:
one question for samir, and then i think its all good

app/models/katello/glue/pulp/repo.rb
sjha4: Yes..That makes sense..👍🏽 
Can we change it to `https://HOSTNAME/pulp_ansible/galaxy/#{relative_path}/api `

app/models/katello/glue/pulp/repo.rb
```diff
@@ -361,16 +361,12 @@ def full_path(smart_proxy = nil, force_https = false)
       scheme   = (self.unprotected && !force_https) ? 'http' : 'https'
```
jlsherrill, line 361: Sorry one last comment :)

With pulp2, unprotected meant that it was only available via http (none were available via https without a special cert).  

But with pulp3, unprotected means that a user can browse the repo over https without a special cert.  Because of this, i think we should default to https on here (instead of defaulting to http).   This way a user browsing the repo details will get an https:// url by default. 

The only time this needs to return http i think is here: app/services/katello/managed_content_medium_provider.rb:12:      url = kickstart_repo.full_path(entity.content_source)

as its returning a kickstart url, it generally needs to be http://

jjeffers: If the default is https, then under what conditions would we set to http? Just kickstart urls?

jlsherrill: yes, i don't think theres any reason to return an http:// url otherwise.  the user could always try it via http (and maybe we should show both in the UI if both are available).  that might be another issue for another day though

sjha4 (approved):
ACK based on all of Justin's reviews and the ansible path change.. 👍🏽

PR #9262, opened 2021-04-05
Fixes #32008 - Candlepin API accepts empty string to unset a value.

lfu:
`nil` would be ignored by Candlepin API.

### DIFF
--- a/lib/katello/tasks/upgrades/4.1/sync_noarch_content.rake
+++ b/lib/katello/tasks/upgrades/4.1/sync_noarch_content.rake
@@ -1,0 +1,15 @@
+namespace :katello do
+  namespace :upgrades do
+    namespace '4.1' do
+      desc "Synchronize arches field of content in candlepin which should be unset."
+      task :sync_noarch_content => ['environment'] do
+        roots = Katello::RootRepository.joins(:product).merge(Katello::Product.custom).where(arch: 'noarch')
+        roots.each do |root|
+          Katello::Resources::Candlepin::Content.update(root.library_instance.organization.label, id: root.content_id, arches: '')
+        rescue RestClient::NotFound
+          #skip content not found
+        end
+      end
+    end
+  end
+end
--- a/test/lib/tasks/upgrades/4.1/sync_noarch_content_test.rb
+++ b/test/lib/tasks/upgrades/4.1/sync_noarch_content_test.rb
@@ -1,0 +1,27 @@
+require 'katello_test_helper'
+require 'rake'
+
+module Katello
+  class SyncNoarchContentTest < ActiveSupport::TestCase
+    def setup
+      Rake.application.rake_require 'katello/tasks/upgrades/4.1/sync_noarch_content'
+      Rake::Task['katello:upgrades:4.1:sync_noarch_content'].reenable
+      Rake::Task.define_task(:environment)
+    end
+
+    def test_sync_content_with_noarch_repository
+      count = Katello::RootRepository.joins(:product).merge(Katello::Product.custom).update_all(arch: 'noarch')
+      assert_operator count, :>=, 1
+      Katello::Resources::Candlepin::Content.expects(:update).times(count)
+
+      Rake.application.invoke_task('katello:upgrades:4.1:sync_noarch_content')
+    end
+
+    def test_no_sync_with_set_arch_repository
+      Katello::RootRepository.joins(:product).merge(Katello::Product.custom).where(arch: 'noarch').update_all(arch: 'i386')
+      Katello::Resources::Candlepin::Content.expects(:update).never
+
+      Rake.application.invoke_task('katello:upgrades:4.1:sync_noarch_content')
+    end
+  end
+end
--- a/app/lib/actions/candlepin/product/content_update.rb
+++ b/app/lib/actions/candlepin/product/content_update.rb
@@ -22,7 +22,7 @@ module Actions
                      contentUrl: input[:content_url],
                      gpgUrl: input[:gpg_key_url] || '', #candlepin ignores nil
                      type: input[:type],
-                     arches: input[:arches],
+                     arches: input[:arches] || '',
                      requiredTags: input[:os_versions],
                      label: input[:label],
                      metadataExpire: 1,
--- a/test/scenarios/scenario_test.rb
+++ b/test/scenarios/scenario_test.rb
@@ -12,6 +12,8 @@ module Scenarios
     end
 
     def test_scenarios
+      skip "Until we can figure out testing this with pulp3"
+
       @support = ScenarioSupport.new(User.current)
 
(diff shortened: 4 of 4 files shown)


jturel:
Verified! Nice.

I think there is another layer to this issue that we need to address. There are presumably users out there who have data in an unsynced state per the bug. An 'upgrade script' (a special kind of rake task that is run when users upgrade to a new Katello version) to correct any inconsistencies should be included with this change. Take a look at some examples in `lib/katello/tasks/upgrades/`

The algorithm will be something like: for all root repositories with a nil arch, call the candlepin product content resource (not the Action class) update to ensure that the arch is nil too. You can run our rake tasks from the foreman dir `bundle exec rake katello:upgrades:$KATELLO_VERSION:$TASK_NAME to test them out!

lib/katello/tasks/upgrades/4.0/sync_noarch_content.rake
```diff
@@ -0,0 +1,13 @@
+namespace :katello do
+  namespace :upgrades do
+    namespace '4.0' do
```
jturel, line 3: This should be in the 4.1 namespace since we're in the 4.1 dev cycle (ending in around May 4th!)

lib/katello/tasks/upgrades/4.0/sync_noarch_content.rake
```diff
@@ -0,0 +1,13 @@
...
+      desc "Synchronize noarch content in candlepin."
+      task :sync_noarch_content => ['environment'] do
```
jturel, line 5: This migration looks good! Can you add a test? You should be able to find some examples of tests for rake tasks existing already

test/lib/tasks/upgrades/4.1/sync_noarch_content_test.rb
```diff
@@ -0,0 +1,26 @@
...
+module Katello
+  class SyncNoarchContentTest < ActiveSupport::TestCase
+    def setup
+      Rake.application.rake_require 'katello/tasks/upgrades/4.1/sync_noarch_content'
+      Rake::Task['katello:upgrades:4.1:sync_noarch_content'].reenable
+      Rake::Task.define_task(:environment)
+    end
+
+    def test_sync_content_with_noarch_repository
+      count = Katello::RootRepository.where(arch: 'noarch').count
```
jturel, line 13: i'm worried that the test fixtures could change one day and maybe count would be 0. Does the `.times` method throw an error for 0? If not , might be worth adding an assertion to make sure we actually are hitting the Content.update method at least once

test/lib/tasks/upgrades/4.1/sync_noarch_content_test.rb
```diff
@@ -0,0 +1,26 @@
+require 'katello_test_helper'
+require 'rake'
+
+module Katello
```
jturel, line 4: Nice tests :)

lib/katello/tasks/upgrades/4.1/sync_noarch_content.rake
```diff
@@ -0,0 +1,13 @@
+namespace :katello do
+  namespace :upgrades do
+    namespace '4.1' do
+      desc "Synchronize arches field of content in candlepin which should be unset."
+      task :sync_noarch_content => ['environment'] do
+        roots = Katello::RootRepository.where(arch: 'noarch')
```
jturel, line 6: This query needs a bit more refinement to exclude Red Hat content (ie content that has come from your manifest). Updating RH content raises a 403 error with candlepin since it's locked for updating. You probably didn't hit that since you don't have a manifest yet!

This is an example of how you can limit the rootrepositories which are mapped to Custom products - ie non-RH products: `Katello::RootRepository.joins(:product).merge(Katello::Product.custom)`

The error I see for reference:

```
RestClient::Forbidden: Katello::Resources::Candlepin::Content: 403 Forbidden {"displayMessage":"content \"10313\" is locked","requestUuid":"b78e21a1-9883-4047-b405-271288fef649"} (PUT /candlepin/owners/Default_Organization/content/10313)
/home/vagrant/foreman/.vendor/ruby/2.5.0/gems/rest-client-2.1.0/lib/restclient/abstract_response.rb:249:in `exception_with_response'
/home/vagrant/foreman/.vendor/ruby/2.5.0/gems/rest-client-2.1.0/lib/restclient/abstract_response.rb:129:in `return!'
/home/vagrant/foreman/.vendor/ruby/2.5.0/gems/rest-client-2.1.0/lib/restclient/request.rb:836:in `process_result'
/home/vagrant/foreman/.vendor/ruby/2.5.0/gems/rest-client-2.1.0/lib/restclient/request.rb:743:in `block in transmit'
/home/vagrant/foreman/.vendor/ruby/2.5.0/gems/rest-client-2.1.0/lib/restclient/request.rb:727:in `transmit'
/home/vagrant/foreman/.vendor/ruby/2.5.0/gems/rest-client-2.1.0/lib/restclient/request.rb:163:in `execute'
/home/vagrant/foreman/.vendor/ruby/2.5.0/gems/rest-client-2.1.0/lib/restclient/request.rb:63:in `execute'
/home/vagrant/foreman/.vendor/ruby/2.5.0/gems/rest-client-2.1.0/lib/restclient/resource.rb:79:in `put'
/home/vagrant/katello/app/lib/katello/http_resource.rb:115:in `put'
/home/vagrant/katello/app/lib/katello/resources/candlepin/content.rb:33:in `update'
/home/vagrant/katello/lib/katello/tasks/upgrades/4.1/sync_noarch_content.rake:8:in `block (5 levels) in <top (required)>'
```

lib/katello/tasks/upgrades/4.1/sync_noarch_content.rake
```diff
@@ -0,0 +1,13 @@
+namespace :katello do
+  namespace :upgrades do
+    namespace '4.1' do
+      desc "Synchronize arches field of content in candlepin which should be unset."
+      task :sync_noarch_content => ['environment'] do
+        roots = Katello::RootRepository.joins(:product).merge(Katello::Product.custom).where(arch: 'noarch')
+        roots.each do |root|
+          Katello::Resources::Candlepin::Content.update(root.library_instance.organization.label, id: root.content_id, arches: '')
```
jturel, line 8: In my dev environment the data is a little bit messed up. I had a RootRepository but the candlepin Content was gone which made this raise a 404. If any users out there have some bad data this would break upgrades. Perhaps this update call should rescue from RestClient::NotFound and log the error instead of failing
```
RestClient::NotFound: Katello::Resources::Candlepin::Content: 404 Not Found {"displayMessage":"Content with ID \"1617221979927\" could not be found.","requestUuid":"d191baf3-222b-47db-99a0-b7b959d46d55"} (PUT /candlepin/owners/import-5905/content/1617221979927)
```

jturel (approved):
APJ!

 Excellent work @lfu !

PR #9266, opened 2021-04-06
Fixes #32261 - Manifest delete should only index subscriptions for current org

jturel:
This is bad because it needlessly re-imports all Subscriptions and Pools for all orgs and not just the org which had its manifest deleted.

**to test**

- create a few organizations
- tail candlepin.log
- import a manifest into one of them
- ensure no regression for subscriptions index by verifying there are no candlepin API requests at this time for any other org except the importing org such as `uri=/candlepin/owners/SOME_OTHER_ORG/products/?`
- delete the org's manifest
- verify the same - no candlepin API requests are made for any other org but the deleting org
- verify that the katello:import_subscriptions rake task queries *all* orgs which is the intended behavior (no regression)

### DIFF
--- a/app/models/katello/glue/provider.rb
+++ b/app/models/katello/glue/provider.rb
@@ -144,7 +144,7 @@ module Katello
           prod_content_importer.import
         end
 
-        self.index_subscriptions(self.organization)
+        self.index_subscriptions
         prod_content_importer
       end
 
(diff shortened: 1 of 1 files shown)


lfu:
LGTM 👍

jlsherrill (approved):
ACK based on @lfu's testing

PR #9267, opened 2021-04-06
Fixes #32233 - Katello should send user-repo info to container gateway

ianballou:
Causes Katello to send user and repo info to the container gateway during proxy sync time.  Katello will only send the user-repo mapping for users who exist in the Container Gateway's database.

Related Container Gateway PR: https://github.com/Katello/smart_proxy_container_gateway/pull/13

TODO:

- [x] Write tests

### DIFF
--- a/app/controllers/katello/api/registry/registry_proxies_controller.rb
+++ b/app/controllers/katello/api/registry/registry_proxies_controller.rb
@@ -93,20 +93,11 @@ module Katello
       true
     end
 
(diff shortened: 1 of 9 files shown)


jjeffers (approved):

jjeffers:
Tested this with docker client and podman via a smart proxy updated with https://github.com/Katello/smart_proxy_container_gateway/pull/13. Worked as expected!

PR #9269, opened 2021-04-07
Fixes #32262 - Avoid leaking file descriptors

ekohl:
Using File.open without closing it can leak file descriptors. File.read ensures the file is closed and as a nice bonus actually shorter.

This also shortens a lot of cases where the file was closed to make it shorter.

I'm not 100% sure everything is exactly the same, I'll note them inline.

### DIFF
--- a/app/models/katello/glue/provider.rb
+++ b/app/models/katello/glue/provider.rb
@@ -97,10 +97,7 @@ module Katello
         data = Resources::Candlepin::UpstreamConsumer.export("#{url}#{upstream['uuid']}/export", upstream['idCert']['cert'],
                                                              upstream['idCert']['key'], ca_file)
 
-        File.open(zip_file_path, 'w') do |f|
-          f.binmode
-          f.write data
-        end
+        File.write(zip_file_path, data, mode: 'wb')
 
         return true
       end
--- a/spec/models/gpg_key_spec.rb
+++ b/spec/models/gpg_key_spec.rb
@@ -12,7 +12,7 @@ module Katello
     describe "create gpg key" do
       before(:each) do
         @organization = get_organization
-        @test_gpg_content = File.open("#{Katello::Engine.root}/spec/assets/gpg_test_key").read
+        @test_gpg_content = File.read("#{Katello::Engine.root}/spec/assets/gpg_test_key")
       end
 
       it "should be successful with valid parameters" do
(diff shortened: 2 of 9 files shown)


app/models/katello/glue/provider.rb
```diff
@@ -97,10 +97,7 @@ def owner_upstream_export(upstream, zip_file_path, _options)
         data = Resources::Candlepin::UpstreamConsumer.export("#{url}#{upstream['uuid']}/export", upstream['idCert']['cert'],
                                                              upstream['idCert']['key'], ca_file)
 
-        File.open(zip_file_path, 'w') do |f|
-          f.binmode
-          f.write data
-        end
+        File.write(zip_file_path, data, mode: 'wb')
```
ekohl, line 100: I think opening with mode `wb` is the same as opening with mode `w` and then calling `binmode`, but please consider this carefully.

spec/models/gpg_key_spec.rb
```diff
@@ -12,7 +12,7 @@ module Katello
     describe "create gpg key" do
       before(:each) do
         @organization = get_organization
-        @test_gpg_content = File.open("#{Katello::Engine.root}/spec/assets/gpg_test_key").read
+        @test_gpg_content = File.read("#{Katello::Engine.root}/spec/assets/gpg_test_key")
```
ekohl, line 15: I considered 
```suggestion
        @test_gpg_content = Katello::Engine.root.join('spec', 'assets', 'gpg_test_key').read
```

test/lib/tasks/receptor/extract_orgs_test.rb
```diff
@@ -39,7 +39,7 @@ def test_output_file
 
       Rake.application.invoke_task('katello:receptor:extract_orgs')
 
-      data = JSON.parse(File.read(File.open(ENV['OUTPUT_FILE'])))
+      data = JSON.parse(File.read(ENV['OUTPUT_FILE']))
```
ekohl, line 42: I think this is the same, but please check. Might also have been a bug.

app/models/katello/glue/provider.rb
jjeffers: According to https://ruby-doc.org/core-2.5.0/IO.html#method-i-binmode they appear to be the same.

test/lib/tasks/receptor/extract_orgs_test.rb
jjeffers: The original was unnecessary syntax in 2.5, but not functionally a defect.  Good catch, thank you.

jjeffers (approved):

PR #9270, opened 2021-04-08
Fixes #32277 - make org creation idempotent and async from seeding

jturel:
**TODO**
- [x] new db field on org to track owner and library env creation in candlepin
- [x] actually test this
- [x] remove dead code unrelated to this pr to boost # of lines deleted
- [x] add debug cert creation to OrganizationCreator
- [x] fix tests / vcrs

**This PR**
- Adds a middleware which ensures all organizations are created as far as katello is concerned
- Removes the `Organization::Create` dynflow action (and a few related ones)
- Uses the new `OrganizationCreator` in the middleware and where the dynflow task was used
- Adds some new errors that are generally useful (for Pinging services and checking candlepin status more accurately)
- Consolidates some of the org related testing that seems to be spread around a lot
- Fires a UI notification if `create_all_organizations` fails so admins can fix the problem (candlepin down)

**to test**

- migrate the db

- tail the candlepin.log to understand the traffic pattern this generates
- hit any URL of the app multiple times
- notice the query to check non-created orgs won't be run on each request for that given rails process since the the middleware is initialized only once
-  multiple requests shouldn't cause a problem cause org create doesn't lock the org anymore
- remember that the org create API calls `OrganizationCreator.create!` whereas seeding calls `OrganizationCreator.seed!` to separate the two. this implies that this middleware should only be relevant to orgs that were seeded but not created through app usage (ie Default Organization)

To reset your orgs: `Organization.update_all(created_in_katello: false)`

- turn candlepin off `systemctl stop tomcat` or 'break' candlepin by moving /etc/candlepin/broker.xml somewhere else and doing `systemctl restart tomcat` and test some more to see the failure scenario

**try to break it so I can fix it**


jturel:
@jeremylenz @jlsherrill here's my rough draft - is this moving in the right direction?

jeremylenz:
> here's my rough draft - is this moving in the right direction?

Heck yeah!

ekohl (commented):
This question is for my understanding. Right now it's a boolean field. Does it make sense to store some ID as how Candlepin tracks it where `null` means it doesn't exist in Candlepin?

The reason I ask is that I had a discussion with @ares where we discussed an option of sharing a single Candlepin instance with multiple Katellos. That could help developers but long term I could also imagine you could connect to some Candlepin cloud instance and only spin up your own if you have a need it (performance, control of data, etc).

Of course I'm not asking for it here, but this PR feels like a potential step in that direction by decoupling Katello and Candlepin a bit. If it's a potentially feasible path, I do think right now is a good time to consider it and see if we align on it.

Other than that I'd like to say I am very happy to see this change.

jturel:
> This question is for my understanding. Right now it's a boolean field. Does it make sense to store some ID as how Candlepin tracks it where `null` means it doesn't exist in Candlepin?

I was going to do that at first but decided not to. My reasoning is that I didn't want to couple the new column name to any particular backend, but rather just the notion of "is everything all good as far as Katello is concerned?" That gives flexibility should the definition of what it means for an org to be `created_in_katello` to change. That definition shift can be addressed with a migration, upgrade script, and supporting changes to `OrganizationCreator` and the `Organization::Create` action.

In some sense I think we already have that piece of data recorded which enables the idempotency of the Org::Create action

https://github.com/Katello/katello/blob/master/app/models/katello/content_view.rb#L476-L478

> 
> The reason I ask is that I had a discussion with @ares where we discussed an option of sharing a single Candlepin instance with multiple Katellos. That could help developers but long term I could also imagine you could connect to some Candlepin cloud instance and only spin up your own if you have a need it (performance, control of data, etc).
> 

I think we'd be looking to store the Candlepin Owner Label on the Organization in this scenario which should probably be randomized in the case of some SaaS Candlepin. Don't think that adds anything to this PR though.

> Of course I'm not asking for it here, but this PR feels like a potential step in that direction by decoupling Katello and Candlepin a bit. If it's a potentially feasible path, I do think right now is a good time to consider it and see if we align on it.
> 
> Other than that I'd like to say I am very happy to see this change.

Glad to hear that :)

db/migrate/20210409033915_add_created_in_katello_to_taxonomy.rb
```diff
@@ -0,0 +1,9 @@
+class AddCreatedInKatelloToTaxonomy < ActiveRecord::Migration[6.0]
+  def up
+    add_column :taxonomies, :created_in_katello, :bool, default: false, null: true
```
ekohl, line 3: is `null: true` here to make the migration easier since in PostgreSQL adding a null column is cheap? From a data perspective I think NULL and false are treated equal but there is a risk that someone might query for where created_in_katello = false and not get null entries back so there is a risk for subtle bugs, right?

ekohl:
> I think we'd be looking to store the Candlepin Owner Label on the Organization in this scenario which should probably be randomized in the case of some SaaS Candlepin. Don't think that adds anything to this PR though.

Fair point. Might be a bit premature and it would be an easy migration otherwise.

db/migrate/20210409033915_add_created_in_katello_to_taxonomy.rb
jturel: Actually I wrote the migration while half asleep so I didn't give it ANY consideration. Good idea to make this non-nullable for the reason you mentioned.

app/lib/actions/katello/organization/create.rb
```diff
@@ -3,30 +3,33 @@ module Katello
...
+
           plan_self(:organization_id => organization.id)
         end
 
         def run
-          ::Organization.find_by(:id => input[:organization_id]).try(:load_debug_cert)
+          org = ::Organization.find_by(:id => input[:organization_id])
+          # do we need to create the debug cert here?
+          # why was it written with try...
+          org.try(:load_debug_cert)
```
jeremylenz, line 31: I did a `rake katello:reset` (seeds completed with no errors) and then attempted to load the web UI.

On the first attempt to load `/` I'm getting an Active Record error `Organization with id Default_Organization could not be found.`  I then refresh the browser and the login page shows, and everything works fine.  Here's what the dynflow console looked like:

org_create_error

and here's more of the stack:
```
- "/home/vagrant/foreman/.vendor/ruby/2.5.0/gems/rest-client-2.1.0/lib/restclient/abstract_response.rb:249:in
  `exception_with_response'"
- "/home/vagrant/foreman/.vendor/ruby/2.5.0/gems/rest-client-2.1.0/lib/restclient/abstract_response.rb:129:in
  `return!'"
- "/home/vagrant/foreman/.vendor/ruby/2.5.0/gems/rest-client-2.1.0/lib/restclient/request.rb:836:in
  `process_result'"
- "/home/vagrant/foreman/.vendor/ruby/2.5.0/gems/rest-client-2.1.0/lib/restclient/request.rb:743:in
  `block in transmit'"
- "/opt/rh/rh-ruby25/root/usr/share/ruby/net/http.rb:910:in `start'"
- "/home/vagrant/foreman/.vendor/ruby/2.5.0/gems/rest-client-2.1.0/lib/restclient/request.rb:727:in
  `transmit'"
- "/home/vagrant/foreman/.vendor/ruby/2.5.0/gems/rest-client-2.1.0/lib/restclient/request.rb:163:in
  `execute'"
- "/home/vagrant/foreman/.vendor/ruby/2.5.0/gems/rest-client-2.1.0/lib/restclient/request.rb:63:in
  `execute'"
- "/home/vagrant/foreman/.vendor/ruby/2.5.0/gems/rest-client-2.1.0/lib/restclient/resource.rb:69:in
  `post'"
- "/home/vagrant/katello/app/lib/katello/http_resource.rb:101:in `post'"
- "/home/vagrant/katello/app/lib/katello/resources/candlepin/owner.rb:94:in `generate_ueber_cert'"
- "/home/vagrant/katello/app/models/katello/glue/candlepin/owner.rb:56:in `generate_debug_cert'"
- "/home/vagrant/katello/app/models/katello/glue/candlepin/owner.rb:62:in `rescue
  in load_debug_cert'"
- "/home/vagrant/katello/app/models/katello/glue/candlepin/owner.rb:59:in `load_debug_cert'"
- "/home/vagrant/foreman/.vendor/ruby/2.5.0/gems/activesupport-6.0.3.6/lib/active_support/core_ext/object/try.rb:15:in
  `public_send'"
- "/home/vagrant/foreman/.vendor/ruby/2.5.0/gems/activesupport-6.0.3.6/lib/active_support/core_ext/object/try.rb:15:in
  `try'"
- "/home/vagrant/katello/app/lib/actions/katello/organization/create.rb:31:in `run'"
- "/home/vagrant/foreman/.vendor/ruby/2.5.0/gems/dynflow-1.4.7/lib/dynflow/action.rb:571:in
  `block (3 levels) in execute_run'"
```

From the Dynflow execution plan it looks like `Actions::Katello::Organization::Create` is planned as a concurrence with the two actions in `plan`, meaning it's trying to run `run` before the Candlepin actions are completed.

app/lib/actions/katello/organization/create.rb
```diff
@@ -3,30 +3,33 @@ module Katello
...
+          org.update!(created_in_katello: true)
         end
 
         def humanized_name
```
jeremylenz, line 35: This shows up in Foreman Tasks as just "Create," which is not ideal.  I think foreman-tasks will try to humanize the inputs as well and append them to the task name, so if you add a `plan_self` with  `organization: organization` it might just work, and say "Create organization Default Organization".  Otherwise, you can add a `humanized_input` method that returns the string to append.

app/lib/actions/katello/organization/create.rb
jeremylenz, line 31: I don't know why `.try` was used, but I think at this point `org` should not be undefined and we deserve any NoMethodErrors that causes.

(anyway I'm partial to `&.` anyway)

app/models/katello/glue/candlepin/owner.rb
```diff
@@ -16,8 +16,19 @@ def self.included(base)
...
     module InstanceMethods
+      def find_owner
+        Resources::Candlepin::Owner.find(self.label)
+      end
+
+      def candlepin_owner_exists?
+        find_owner
+        true
+      rescue RestClient::ResourceNotFound
+        false
```
jeremylenz, line 27: :heart_eyes:

app/services/katello/organization_creator.rb
```diff
@@ -0,0 +1,78 @@
...
+        end
+      end
+    end
+
+    def initialize(organization)
+      @organization = organization
+    end
+
+    def seed!
+      @organization.setup_label_from_name
```
jeremylenz, line 26: is this a good time to correct my grammar pet peeve and rename this method? :yum: 
```suggestion
      @organization.set_up_label_from_name
```

lib/katello/engine.rb
```diff
@@ -16,6 +16,9 @@ class Engine < ::Rails::Engine
 
       require 'katello/middleware/event_daemon'
       app.middleware.use(Katello::Middleware::EventDaemon)
+
+      require 'katello/middleware/organization_created_enforcer'
+      app.middleware.use(Katello::Middleware::OrganizationCreatedEnforcer)
```
jeremylenz, line 21: Does the middleware approach work better than controller before-action thingies?

test/actions/katello/organization_test.rb
```diff
@@ -42,22 +42,55 @@ def except_notification(action, action_result)
...
-      organization.expects(:create_redhat_provider)
-      organization.expects(:save!)
-      action.stubs(:action_subject).with(organization, any_parameters)
+    it 'plans and runs' do
+      organization.expects(:candlepin_owner_exists?).returns(false)
+      ::Organization.any_instance.expects(:load_debug_cert)
+      action.stubs(:action_subject).with(organization)
       plan_action(action, organization)
-      assert_action_planed_with(action,
+      assert_action_planned_with(action,
```
jeremylenz, line 53: :pray: 

We should remove that alias at some point..  :laughing:

jeremylenz (commented):
Thanks @jturel!

See test results & comments below

app/lib/actions/katello/organization/create.rb
jturel: Good observation. Should be fixed

app/lib/actions/katello/organization/create.rb
jturel: You nailed it. Should be fixed!

app/lib/actions/katello/organization/create.rb
jturel: I'm leaning that way too. Perhaps @jlsherrill will take us to school here...

app/services/katello/organization_creator.rb
jturel: I like it. But the module that defines this method is used in several models so I'd prefer to leave any change out of this PR :-/

test/actions/katello/organization_test.rb
jturel: One day ..

lib/katello/engine.rb
jturel: I like it better because the middleware approach will only check to run the task **about once per process** (in theory) unless there is a huge flood of requests coming in at startup time. Will be good to check what happens when candlepin is down and it needs to run the task though.

On the flip side, the Concern approach will check the org for each request at least how i've written it.

app/lib/actions/katello/organization/create.rb
jeremylenz: indeed it appears so! Resolving

jturel:
@jlsherrill @jeremylenz decided to take this a bit further and remove the Organization::Create dynflow task altogether. Thoughts?

app/services/katello/organization_creator.rb
```diff
@@ -0,0 +1,108 @@
...
+      if needs_candlepin_organization?
+        ::Katello::Resources::Candlepin::Owner.create(@organization.label, @organization.name)
+      end
+
+      ::Katello::ContentViewManager.create_candlepin_environment(
+        content_view_environment: @content_view_environment
+      )
+
+      @organization.created_in_katello = true
+      @organization.save!
```
jeremylenz, line 52: may be nice to add an audit comment here conveying that the Katello-ey-ness of the organization was consummated!

app/controllers/katello/concerns/organizations_controller_extensions.rb
```diff
@@ -30,7 +30,7 @@ def create
           if taxonomy_class == Organization
             begin
               @taxonomy = Organization.new(resource_params)
-              sync_task(::Actions::Katello::Organization::Create, @taxonomy)
+              ::Katello::OrganizationCreator.new(@taxonomy).create!
```
jeremylenz, line 33: Should our `create` calls from controllers, um, be in a transaction?

(since we don't have that protection from Dynflow anymore)

jeremylenz (commented):
@jturel Getting rid of that action seems fine to me!

app/services/katello/organization_creator.rb
jturel: I'm wondering what the value of that is. I view Audits as the record of things that happened which users care about and this is more of an internal process, not something they need to track or monitor.

jeremylenz: Maybe it would be better as a log then -- I was just thinking about what if something failed halfway through `seed` and the creation was left incomplete?  How could we make that easier to discover and address?

jturel: Keeping in mind that the OrganizationCreator should be resilient to failures at any step (failed runs will not impede subsequent ones), here are the scenarios:

the db:seed process fails => This will raise an error just like it would today.

org create via API, UI fails => Error will be raised as it would today

middleware for org creation fails => it pings candlepin before starting the process and if that ping fails then a UINotification is sent. any other error will be raised. that should likely change (since right now it will basically make the app unusable..) but I'm having trouble envisioning where this could break down such that an error gets raised in creating the handful of entities required for a fully-fledged org

jturel: Take a look at latest commit - it's not fully baked but I think it captures what you have in mind. I need to compare the current behavior to understand implications of removing the created candlepin environment and owner if there were any problems.

app/services/katello/organization_creator.rb
```diff
@@ -26,30 +26,37 @@ def initialize(organization)
...
-      @organization.setup_label_from_name
-      @organization.save!
-
-      create_library_environment
-      create_library_view
-      create_library_cvv
-      create_content_view_environment
-      create_anonymous_provider
-      create_redhat_provider
+      ActiveRecord::Base.transaction do
```
jeremylenz, line 29: Won't this fail if `OrganizationCreator` is run from `102-organizations.rb` within the transaction in `foreman_seeder`?  That's why my first thought was to put the `.transaction` in the controller rather than here, to ensure that we're not trying to nest the transactions.  I'm sure there's a better way I'm just not thinking of, though..

jturel: If you mean that simply using a transaction here will raise an error - no error!

parthaa:
Worked  for me as well :+1:

app/models/katello/glue/candlepin/owner.rb
```diff
@@ -16,8 +16,19 @@ def self.included(base)
...
 
     module InstanceMethods
+      def find_owner
+        Resources::Candlepin::Owner.find(self.label)
+      end
+
+      def candlepin_owner_exists?
+        find_owner
+        true
+      rescue RestClient::ResourceNotFound
```
jlsherrill, line 26: this method is a tad dangerous as it cannot detect between candlepin being down and the owner being non-existant.  Dynflow would 'ping' candlepin before the candlepin action, i wonder if this should too (instead of adding it everywhere this is called).  It might result in duplicates but these should be quick calls i think?

app/models/katello/glue/candlepin/environment.rb
```diff
@@ -11,6 +11,13 @@ def candlepin_info
         Resources::Candlepin::Environment.find(self.cp_id)
       end
 
+      def exists_in_candlepin?
+        candlepin_info
+        true
+      rescue RestClient::NotFound
```
jlsherrill, line 17: same question here

app/models/katello/glue/candlepin/owner.rb
jturel: Watcha think about pinging candlepin in the `OrganizationCreator` and instead of checking if the owner exists (assuming we get this far ie candlepin is up) we just try to create the owner and rescue the BadRequest if it already exists?

jlsherrill: does it return badrequest? or 409 conflict?    If we can detect a conflict over  a bad request i'd maybe be okay with it.  But if we get back a  bad request, it might be bad to assume it exists?

ekohl: You can also ping in the `rescue` if the 404 doesn't contain a guarantee from Candlepin. I don't know Candlepin well enough, but perhaps it always sends a header that you can check that it does indeed come from the application. If not, it could be an interesting conversation to have with upstream.

jturel: Thanks @ekohl for the header check inspiration. I went that route.

Take a look at the latest commit and see how the error handling looks @jlsherrill  - I think it's safer now

jturel:
Description updated to reflect where this has landed

app/lib/katello/http_resource.rb
```diff
@@ -137,13 +137,8 @@ def delete(a_path = nil, headers = {})
 
       # re-raise the same exception with nicer error message
       def raise_rest_client_exception(e, a_path, http_method)
-        # message method in rest-client is hardcoded - we need to override it
         msg = "#{name}: #{e.message} #{e.http_body} (#{http_method} #{a_path})"
-        (class << e; self; end).instance_eval do
```
jturel, line 142: I guess this was written in an age when the RestClientException didn't have the accessor for message...

test/models/organization_test.rb
```diff
@@ -3,7 +3,37 @@
...
 
 module Katello
-  class OrganizationTestDelete < ActiveSupport::TestCase
+  class OrganizationTest < ActiveSupport::TestCase
+    def test_org_valid_labels
+      assert Organization.new(:name => 'name', :label => 'org').valid?
+      assert Organization.new(:name => 'name', :label => 'with_underscore').valid?
+      assert Organization.new(:name => 'name', :label => 'With_Capital_letter').valid?
+      assert Organization.new(:name => 'name', :label => 'with_number').valid?
+      refute Organization.new(:name => 'name', :label => 'with\'space').valid?
```
jeremylenz, line 12:
```
irb(main):001:0> 'with\'space'
=> "with'space"
```
seems like this should be called `'with\'apostrophe'` ?

jturel: in my defense I copy/pasted these tests :)

app/models/katello/glue/candlepin/owner.rb
jlsherrill: I like this approach!

ehelms:
I gave this a run through in an install and upgrade pipeline. I got green runs in both cases. :+1: from a runtime perspective. I won't presume to code review though :)

db/seeds.d/109-katello-notification-blueprints.rb
```diff
@@ -84,6 +84,18 @@
     name: 'sca_disable_error',
     message: N_('Disabling Simple Content Access failed for \'%{subject}\'.'),
     level: 'error'
+  },
+  {
+    group: N_('System Status'),
+    name: 'system_status_error',
+    message: N_('Some services are not properly started. See the About page for more information.'),
```
jeremylenz, line 91: :+1: :+1: Just seeing this! Nice.

jeremylenz (approved):
* `db:seed` and `katello:reset` complete without errors
* orgs finish creation without errors
* `created_in_katello` gets set as needed

:heavy_check_mark: tried to break it & couldn't

nice work @jturel!  ACK

jturel:
squashed and rebase - will merge once green!

ehelms:
> squashed and rebase - will merge once green!

:green_apple: :green_book: :green_circle: :green_heart: :green_salad:

jturel:
image

ezr-ondrej:
@jturel I think that https://github.com/Katello/katello/blob/0e7f14ce3b639b36ab968d460cc441228d991b16/app/services/katello/organization_creator.rb#L49 causes second Audit for the created Org (update audit with the single - internal column change).

1) I'd expect that column to be not audited
2) we should not have additional update audit.

jturel:
@ezr-ondrej so the fix would be to not audit that column which would stop the additional update audit from being generated?

ezr-ondrej:
> so the fix would be to not audit that column which would stop the additional update audit from being generated?

Yeah, I guess that would be enough, tho saving the org twice seems not being needed.
But that I'd leave up to you, not auditing the column would fix the core issue :)

PR #8002, opened 2019-03-07
Fixes #26275: add react helper to sync status page

waldenraines:
Add the new react app helper to the sync status page to avoid an
error.

https://projects.theforeman.org/issues/26275

PR #9272, opened 2021-04-10
Fixes #32285 - use public API to update setting

ezr-ondrej:
We are updating setting for content proxy directly in database, but that
blocks efforts in core to migrate all settings into memory.
This would just break tests, as in production settings will be loaded
from database per request anyway, but it is still better to access
Setting through public API.

This was discovered by failing tests on #8002

### DIFF
--- a/app/models/katello/concerns/http_proxy_extensions.rb
+++ b/app/models/katello/concerns/http_proxy_extensions.rb
@@ -49,10 +49,8 @@ module Katello
         changes = self.previous_changes
         if changes.key?(:name)
           previous_name = changes[:name].first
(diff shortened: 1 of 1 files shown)


jlsherrill:
@ezr-ondrej rubocop failed which prevented the test from running.  This change makes total sense and i'm good to ack it assuming tests pass.

ezr-ondrej:
hehe, good catch by rubocop :+1:

jlsherrill (approved):

jlsherrill:
thanks @ezr-ondrej !

PR #9277, opened 2021-04-12
Fixes #32302 - delete orphans before switchover

jlsherrill:


app/services/katello/pulp3/migration_switchover.rb
```diff
@@ -24,6 +24,18 @@ def run
...
+        models = []
+        ::Katello::RepositoryTypeManager.repository_types.each_value do |repo_type|
+          indexable_types = repo_type.content_types_to_index
+          models += indexable_types&.map(&:model_class)
+          models.select! { |model| model.many_repository_associations }
+        end
+        models.each do |model|
+          model.joins("left join katello_#{model.repository_association} on #{model.table_name}.id = katello_#{model.repository_association}.#{model.unit_id_field}").where("katello_#{model.repository_association}.#{model.unit_id_field} IS NULL").destroy_all
+        end
+      end
```
jlsherrill, line 37: note that i essentially duplicated this code, which i'd normally avoid, but since this code is dead after 3.18 it seemed a simpler solution than refactoring other code just for this (and only in 3.18)

jlsherrill:
So you should be able to reproduce this by:


1. Sync some docker content
2. Publish it in a CV
3. Delete 1 or more CV versions
4. migrate and switchover

You could also try with some orphaned rpm content for bonus points

jjeffers (approved):
Tested this with the scenario described - no errors in migration, removal of published CVVs, and then re-migration, and finally switchover command.

PR #9283, opened 2021-04-14
Fixes #32256 - aborts a migration task if the Katello config is invalid

jjeffers:
Adds a check for Katello configuration - the `use_pulp_2_for_content_type` section must exist and at least 1 of the content types must be set to `true`. If not, the migration task fails.

### DIFF
--- a/lib/katello/tasks/check_config.rake
+++ b/lib/katello/tasks/check_config.rake
@@ -1,0 +1,17 @@
+namespace :katello do
+  task :check_config => ['environment'] do
+    desc "Task that can be run before a content migration to check that the configuration valid"
+
+    puts "Checking for valid Katello configuraton."
+    if SETTINGS[:katello][:use_pulp_2_for_content_type].nil?
+      fail _("Invalid Katello configuration: 'use_pulp_2_for_content_type' is missing. ")
+    end
+
+    if !SETTINGS[:katello][:use_pulp_2_for_content_type][:docker] &&
+      !SETTINGS[:katello][:use_pulp_2_for_content_type][:file] &&
+      !SETTINGS[:katello][:use_pulp_2_for_content_type][:yum] &&
+      !SETTINGS[:katello][:use_pulp_2_for_content_type][:deb]
+      fail _("Invalid Katello configuration: no content types in 'use_pulp_2_for_content_type' section are set to 'true'. ")
+    end
+  end
+end
--- a/lib/katello/tasks/check_config.rake
+++ b/lib/katello/tasks/check_config.rake
@@ -1,17 +1,18 @@
 namespace :katello do
   task :check_config => ['environment'] do
     desc "Task that can be run before a content migration to check that the configuration valid"
+    fail_msg = _("The system appears to already be using pulp3 with all content migrated.")
 
     puts "Checking for valid Katello configuraton."
     if SETTINGS[:katello][:use_pulp_2_for_content_type].nil?
-      fail _("Invalid Katello configuration: 'use_pulp_2_for_content_type' is missing. ")
+      fail fail_msg
     end
 
     if !SETTINGS[:katello][:use_pulp_2_for_content_type][:docker] &&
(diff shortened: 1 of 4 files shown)


lib/katello/tasks/pulp3_migration.rake
```diff
@@ -4,6 +4,23 @@ namespace :katello do
...
+    end
+
+    if !SETTINGS[:katello][:use_pulp_2_for_content_type][:docker] &&
+      !SETTINGS[:katello][:use_pulp_2_for_content_type][:file]
+      !SETTINGS[:katello][:use_pulp_2_for_content_type][:yum]
+      !SETTINGS[:katello][:use_pulp_2_for_content_type][:deb]
+      msg = _("Migration aborted, no content types in 'use_pulp_2_for_content_type' section are set to 'true'.")
+      $stderr.print(msg)
+      fail ForemanTasks::TaskError, task
+    end
```
jlsherrill, line 22: i had originally envisioned this to go in foreman-maintain, are you sure that foreman-maintain won't shut down pulp services aftewards in the 6.9 scenario ?

lib/katello/tasks/pulp3_migration.rake
```diff
@@ -4,6 +4,23 @@ namespace :katello do
     services = [:candlepin, :foreman_tasks, :pulp3, :pulp, :pulp_auth]
     Katello::Ping.ping!(services: services)
 
+    #check for valid yaml
+    puts "Checking for valid migration configuraton."
+    if SETTINGS[:katello][:use_pulp_2_for_content_type].nil?
+      msg = _("Migration aborted, 'use_pulp_2_for_content_type' is missing from Katello configuration.")
+      $stderr.print(msg)
+      fail ForemanTasks::TaskError, task
```
jlsherrill, line 12: i think you could just ```fail msg```  here, but that 'task' doesn't seem to actually exist

lib/katello/tasks/pulp3_migration.rake
```diff
@@ -4,6 +4,23 @@ namespace :katello do
...
+    end
+
+    if !SETTINGS[:katello][:use_pulp_2_for_content_type][:docker] &&
+      !SETTINGS[:katello][:use_pulp_2_for_content_type][:file]
+      !SETTINGS[:katello][:use_pulp_2_for_content_type][:yum]
+      !SETTINGS[:katello][:use_pulp_2_for_content_type][:deb]
+      msg = _("Migration aborted, no content types in 'use_pulp_2_for_content_type' section are set to 'true'.")
+      $stderr.print(msg)
+      fail ForemanTasks::TaskError, task
```
jlsherrill, line 21: same point here for the fail

lib/katello/tasks/check_config.rake
```diff
@@ -0,0 +1,17 @@
...
+    puts "Checking for valid Katello configuraton."
+    if SETTINGS[:katello][:use_pulp_2_for_content_type].nil?
+      fail _("Invalid Katello configuration: 'use_pulp_2_for_content_type' is missing. ")
+    end
+
+    if !SETTINGS[:katello][:use_pulp_2_for_content_type][:docker] &&
+      !SETTINGS[:katello][:use_pulp_2_for_content_type][:file] &&
+      !SETTINGS[:katello][:use_pulp_2_for_content_type][:yum] &&
+      !SETTINGS[:katello][:use_pulp_2_for_content_type][:deb]
+      fail _("Invalid Katello configuration: no content types in 'use_pulp_2_for_content_type' section are set to 'true'. ")
```
jlsherrill, line 14: we might wanna adjust these error messages a bit, something more along the lines of "The system appears to already be using pulp3 with all content migrated".  From the user perspective, these settings are kinda meaningless.  

I would adjust the one above to kinda same the same thing.  That section missing isn't invalid, it just means the content has already been migrated and switched over.  I think the same message for both of these is fine.

jlsherrill (changes requested):

PR #9286, opened 2021-04-14
Fixes #31827 - View component CVS for composite Content Views

sjha4:
- [x] Tests

### DIFF
--- a/app/controllers/katello/api/v2/content_view_components_controller.rb
+++ b/app/controllers/katello/api/v2/content_view_components_controller.rb
@@ -1,5 +1,6 @@
 module Katello
   class Api::V2::ContentViewComponentsController < Api::V2::ApiController
+    include Katello::Concerns::FilteredAutoCompleteSearch
     before_action :find_composite_content_view, :only => [:show, :index]
     before_action :find_composite_content_view_for_edit, :only => [:add_components, :remove_components, :update]
     before_action :find_authorized_katello_resource, :only => [:show, :update]
@@ -10,7 +11,12 @@ module Katello
         N_("List components attached to this content view")
     param :composite_content_view_id, :number, :desc => N_("composite content view identifier"), :required => true
     def index
-      respond :collection => index_response
+      sort_by, sort_order, options = sort_options
+      respond(:collection => scoped_search(index_relation, sort_by, sort_order, options))
+    end
+
+    def index_relation
+      @view.content_view_components.readable
     end
 
     def index_response
@@ -113,6 +119,21 @@ module Katello
     def component_params
       attrs = [:latest, :content_view_version_id, :content_view_id]
       params.require(:content_view_component).permit(*attrs)
+    end
+
+    def default_sort
+      %w(label asc)
+    end
+
+    def sort_options
+      case default_sort
+      when Array
+        [default_sort[0], default_sort[1], {}]
+      when Proc
+        [nil, nil, { :custom_sort => default_sort }]
+      else
+        fail "Unsupported default_sort type"
+      end
     end
   end
 end
--- a/webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentEnvironments.js
+++ b/webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentEnvironments.js
@@ -1,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Label, Flex, FlexItem } from '@patternfly/react-core';
+
+const ComponentEnvironments = ({ environments }) => {
+  const envList = environments.map(env =>
+    (
+      <FlexItem key={env.id} style={{ marginTop: '0.5em' }}>
+        <Label color="purple" href={`/lifecycle_environments/${env.id}`}>{`${env.name}`}</Label>
+      </FlexItem>
+    ));
+  return (
+    <Flex grow={{ default: 'grow' }} spaceItems={{ default: 'spaceItemsSm' }}>
+      {envList}
+    </Flex>
+  );
+};
+
+ComponentEnvironments.propTypes = {
+  environments: PropTypes.instanceOf(Array),
+};
+
+ComponentEnvironments.defaultProps = {
+  environments: [],
+};
+
+export default ComponentEnvironments;
+
--- a/webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentVersion.js
+++ b/webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentVersion.js
@@ -1,0 +1,45 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
+import PropTypes from 'prop-types';
+import { urlBuilder } from 'foremanReact/common/urlHelpers';
+import { translate as __ } from 'foremanReact/common/I18n';
+import {
+  Text,
+  TextContent,
+  TextVariants,
+} from '@patternfly/react-core';
+
+const ComponentVersion = ({ componentCV }) => {
+  const { latest, content_view_version: cvVersion, content_view: cv } = componentCV;
+  const {
+    id,
+    latest_version: latestVersion,
+  } = cv;
+  const { version } = cvVersion;
+  const latestDescription = __('Latest (automatically updates)');
+  const manualVersionText = (latestVersion === version) ? __('Latest version') : __(`New version is available: Version ${latestVersion}`);
+  return (
+    <>
+      <Link to={urlBuilder('labs/content_views', '', id)}> Version {version}</Link>
+      <TextContent>
+        <Text component={TextVariants.small}>{latest ? latestDescription : manualVersionText}</Text>
+      </TextContent>
+    </>
+  );
+};
+
+ComponentVersion.propTypes = {
+  componentCV: PropTypes.shape({
+    latest: PropTypes.bool.isRequired,
+    content_view_version: PropTypes.shape({
+      version: PropTypes.string.isRequired,
+    }),
+    content_view: PropTypes.shape({
+      id: PropTypes.number.isRequired,
+      latest_version: PropTypes.string.isRequired,
+    }),
+  }).isRequired,
+};
+
+export default ComponentVersion;
+
--- a/webpack/scenes/ContentViews/Details/ComponentContentViews/ContentViewComponents.js
+++ b/webpack/scenes/ContentViews/Details/ComponentContentViews/ContentViewComponents.js
@@ -1,0 +1,124 @@
+import React, { useState, useEffect } from 'react';
+import { useSelector } from 'react-redux';
+import {
+  Bullseye,
+} from '@patternfly/react-core';
+import { Link } from 'react-router-dom';
+import { TableVariant, fitContent, TableText } from '@patternfly/react-table';
+import { STATUS } from 'foremanReact/constants';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { urlBuilder } from 'foremanReact/common/urlHelpers';
+import PropTypes from 'prop-types';
+
+import TableWrapper from '../../../../components/Table/TableWrapper';
+import onSelect from '../../../../components/Table/helpers';
+import {
+  selectCVComponents,
+  selectCVComponentsStatus,
+  selectCVComponentsError,
+} from '../ContentViewDetailSelectors';
+import { getContentViewComponents } from '../ContentViewDetailActions';
+import AddedStatusLabel from '../../../../components/AddedStatusLabel';
+import ComponentVersion from './ComponentVersion';
+import ComponentEnvironments from './ComponentEnvironments';
+import ContentViewIcon from '../../components/ContentViewIcon';
+
+const ContentViewComponents = ({ cvId, details }) => {
+  const response = useSelector(state => selectCVComponents(state, cvId));
+  const status = useSelector(state => selectCVComponentsStatus(state, cvId));
+  const error = useSelector(state => selectCVComponentsError(state, cvId));
+
+  const [rows, setRows] = useState([]);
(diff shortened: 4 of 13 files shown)


sjha4:
[test katello]

sjha4:
[test katello]

webpack/scenes/ContentViews/Details/ComponentContentViews/ContentViewComponents.js
```diff
@@ -0,0 +1,114 @@
...
+    { title: __('Status') },
+    __('Description'),
+  ];
+  const loading = status === STATUS.PENDING;
+
+  const buildRows = (results) => {
+    const newRows = [];
+    results.forEach((componentCV) => {
+      const { content_view: cv, content_view_version: cvVersion } = componentCV;
+      const { environments, repositories } = cvVersion;
```
jeremylenz, line 49: I got a blank page with this error in the console:
```
ContentViewComponents.js:49 Uncaught TypeError: Cannot read property 'environments' of null
    at ContentViewComponents.js:49
    at buildRows (ContentViewComponents.js:47)
    at ContentViewComponents.js:77
    at commitHookEffectListMount (react-dom.development.js:19731)
    at commitPassiveHookEffects (react-dom.development.js:19769)
    at HTMLUnknownElement.callCallback (react-dom.development.js:188)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:237)
    at invokeGuardedCallback (react-dom.development.js:292)
    at flushPassiveEffectsImpl (react-dom.development.js:22853)
```

I think it was because my component views hadn't been published yet.  Once I published a new version for each of my two component views, the error was gone.

I'd also like to see a test for this :)

webpack/scenes/ContentViews/Details/ComponentContentViews/ContentViewComponents.js
```diff
@@ -0,0 +1,114 @@
...
+      const {
+        id,
+        name,
+        description,
+      } = cv;
+
+      const cells = [
+        { title: <Bullseye><ContentViewIcon composite={false} /></Bullseye> },
```
jeremylenz, line 57: The tooltip says "single"; shouldn't it say "component"?

webpack/scenes/ContentViews/Details/ComponentContentViews/ContentViewComponents.js
```diff
@@ -0,0 +1,114 @@
...
+    const { results, ...meta } = response;
+    setMetadata(meta);
+
+    if (!loading && results) {
+      const newRows = buildRows(results);
+      setRows(newRows);
+    }
+  }, [JSON.stringify(response)]);
+
+  const emptyContentTitle = __("You currently don't have any content views added to this content view.");
```
jeremylenz, line 82: I actually like what we have for the current page, which is: 
```
No Content Views belong to {composite_view_name}. Use the add tab to add Content Views.
```
```suggestion
  const emptyContentTitle = __("No content views belong to ${name}.");
```

jeremylenz (commented):
Off to a good start!  Some errors and comments below

webpack/scenes/ContentViews/Details/ComponentContentViews/ContentViewComponents.js
sjha4: I have a PR here to update the icon for Composite and Component..I'll add this change in the PR: https://github.com/Katello/katello/pull/9298

sjha4: Updated in the PR above.

webpack/scenes/ContentViews/Details/ComponentContentViews/ContentViewComponents.js
sjha4: Updated to display this with ${label}

webpack/scenes/ContentViews/Details/ComponentContentViews/ContentViewComponents.js
sjha4: Updated.. Working on tests.

sjha4: Pushed a test to cover this..

sjha4:
[test katello]

sjha4:
[test katello]

webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentVersion.js
```diff
@@ -0,0 +1,45 @@
...
+
+const ComponentVersion = ({ componentCV }) => {
+  const { latest, content_view_version: cvVersion, content_view: cv } = componentCV;
+  const {
+    id,
+    latest_version: latestVersion,
+  } = cv;
+  const { version } = cvVersion;
+  const latestDescription = __('Latest(automatically updates)');
+  const manualVersionText = (latestVersion === version) ? __('Latest Version') : __(`New Version is available: Version ${latestVersion}`);
```
jeremylenz, line 20: Is this the wording we decided on?  If so, only the first words should be capitalized:
```suggestion
  const manualVersionText = (latestVersion === version) ? __('Latest version') : __(`New version is available: Version ${latestVersion}`);
```

app/controllers/katello/api/v2/content_view_components_controller.rb
```diff
@@ -114,5 +120,19 @@ def component_params
...
+    def default_sort
+      %w(label asc)
+    end
+
+    def sort_options
+      if default_sort.is_a?(Array)
+        return [default_sort[0], default_sort[1], {}]
+      elsif default_sort.is_a?(Proc)
+        return [nil, nil, { :custom_sort => default_sort }]
+      else
```
jeremylenz, line 133: Not a big deal, but by the time an `elsif` is needed I kinda prefer `case`:
```rb
case default_sort
  when Array
    ...
  when Proc
   ...
  else
  ...
  end
```
Since Ruby cases don't have fall-through, you also wouldn't have to say `return` or `break` each time.

webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentVersion.js
```diff
@@ -0,0 +1,45 @@
...
+  return (
+    <React.Fragment>
```
jeremylenz, line 22: Pretty sure we can just use `<> ... </>` instead of `<React.Fragment> ... </React.Fragment>` now! It should work as long as you don't need to assign a `key` prop.

webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentEnvironments.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Label } from '@patternfly/react-core';
+
+const ComponentEnvironments = ({ environments }) => {
+  const envLabels = environments.map(env =>
+    <React.Fragment key={env.id}><Label color="purple" href={`/lifecycle_environments/${env.id}`}>{`${env.name}`}</Label><br /></React.Fragment>);
+  return envLabels;
+};
```
jeremylenz, line 9: I think this will be better if you don't assign it to a variable:
```suggestion
const ComponentEnvironments = ({ environments }) => {
  return environments.map(env =>
    (
      <React.Fragment key={env.id}>
        <Label color="purple" href={`/lifecycle_environments/${env.id}`}>{`${env.name}`}</Label>
        <br />
      </React.Fragment>
    ));
};
```

webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentVersion.js
```diff
@@ -0,0 +1,45 @@
...
+} from '@patternfly/react-core';
+
+const ComponentVersion = ({ componentCV }) => {
+  const { latest, content_view_version: cvVersion, content_view: cv } = componentCV;
+  const {
+    id,
+    latest_version: latestVersion,
+  } = cv;
+  const { version } = cvVersion;
+  const latestDescription = __('Latest(automatically updates)');
```
jeremylenz, line 19:
```suggestion
  const latestDescription = __('Latest (automatically updates)');
```

webpack/scenes/ContentViews/Details/ComponentContentViews/ContentViewComponents.js
```diff
@@ -0,0 +1,124 @@
...
+          title: <AddedStatusLabel added />,
+        },
+        { title: <TableText wrapModifier="truncate">{description || 'No description'}</TableText> },
+      ];
+      newRows.push({ cells });
+    });
+    return newRows;
+  };
+
+  useEffect(() => {
```
jeremylenz, line 73: Should probably move this down to after you're done defining constants.

webpack/scenes/ContentViews/Details/ContentViewDetailActions.js
```diff
@@ -117,4 +118,15 @@ export const getCVFilterRules = (filterId, params) => get({
   url: api.getApiUrl(`/content_view_filters/${filterId}/rules`),
 });
 
+export const getContentViewComponents = (cvId, params) => {
+  const apiParams = { ...params };
+  const apiUrl = `/content_views/${cvId}/content_view_components`;
+  return get({
+    key: cvDetailsComponentKey(cvId),
+    params: apiParams,
+    errorToast: error => __(`Something went wrong while retrieving the content view components! ${error}`),
```
jeremylenz, line 127: Does this have to be a function?

jeremylenz (commented):
Works as expected! A few more nits etc. below

webpack/scenes/ContentViews/Details/ContentViewDetailActions.js
sjha4: The error toast?

webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentVersion.js
sjha4: Updated.. 👍🏽

app/controllers/katello/api/v2/content_view_components_controller.rb
sjha4: Updated..

webpack/scenes/ContentViews/Details/ContentViewDetailActions.js
jeremylenz: Yeah, I was just curious why the value of `errorToast` is a function.

sjha4: It's cause we need the error as a param to form the error message.

We do need to parse through error messages better using the helper function we are using elsewhere but not in Detail Actions. Will push an update to do so..

https://github.com/Katello/katello/blob/628ee7ed2a40ec1ea33c1d84bd940035ad34bdec/webpack/utils/helpers.js#L26

app/controllers/katello/api/v2/content_view_components_controller.rb
```diff
@@ -114,5 +120,20 @@ def component_params
...
     end
+
+    def default_sort
+      %w(label asc)
+    end
+
+    def sort_options
+      case default_sort
+      when Array
+        return [default_sort[0], default_sort[1], {}]
```
jeremylenz, line 131: I don't think you need the `return`s

webpack/scenes/ContentViews/Details/ContentViewDetailActions.js
```diff
@@ -72,21 +71,21 @@ export const getContentViewRepositories = (cvId, params, status) => {
 export const getRepositoryTypes = () => get({
   type: API_OPERATIONS.GET,
   key: REPOSITORY_TYPES,
-  errorToast: error => __(`Something went wrong while retrieving the repository types! ${error}`),
+  errorToast: error => __(`Something went wrong while retrieving the repository types! Error: ${getResponseErrorMsgs(error.response)}`),
```
jeremylenz, line 74: I think explicitly saying `Error: ` isn't needed-- just in case you ever get an `error` with no `response` key, the text would make more sense without it.

jeremylenz (commented):
Still works as expected. Changes look good, getting very close :smiley: 

When there are multiple environments, it seems like it's giving the Version column too much width, and not enough width to the Environments column.
* Would like to see the column widths adjusted to give more space to Environments if possible
* If the environments must spill over to the next line, they need spacing so they're not touching.
* always good pleasure reading: https://css-tricks.com/snippets/css/a-guide-to-flexbox/ ;)

composite view components

jeremylenz:
Should we also be showing component views that are not Added?  I just realized it's only showing Added views.

sjha4:
> Should we also be showing component views that are not Added? I just realized it's only showing Added views.

Ya..We'll need that when we implement add/remove CVs from composite. We'll also need a new API endpoint I think covered here: https://projects.theforeman.org/issues/32303 but it might change based on implementation. 
Leaving it out of scope for this PR.

sjha4:
> Still works as expected. Changes look good, getting very close 
> 
> When there are multiple environments, it seems like it's giving the Version column too much width, and not enough width to the Environments column. Each Environment is on its own line, but the mockup shows they can flow horizontally and then wrap if needed.
> 
> * Would like to see the column widths adjusted to give more space to Environments if possible
> * If the environments must spill over to the next line, they need spacing so they're not touching.
> * always good pleasure reading: https://css-tricks.com/snippets/css/a-guide-to-flexbox/ ;)
> 
> composite view components

Screenshot from 2021-04-22 13-01-30

jeremylenz (approved):
Beautiful!

APJ :heavy_check_mark:

PR #9288, opened 2021-04-14
Fixes #32298 - consider pulp3 proxies for content sources

jlsherrill:


chris1984 (approved):
ACK, tested on my external devel box:

https://nimb.ws/hiYepQ

PR #9291, opened 2021-04-15
Fixes #32275 - ping correct Smart proxy for pulp2

jlsherrill:
instead of always pinging the central one

### DIFF
--- a/app/models/katello/ping.rb
+++ b/app/models/katello/ping.rb
@@ -97,10 +97,10 @@ module Katello
         end
       end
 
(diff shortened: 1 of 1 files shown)


jjeffers (approved):
LGTM - was this a bug before? Seem like it would be a problem in some situations.

jlsherrill:
yep, it was mistakenly pinging the pulp2 server on the main katello server (with auth), we never noticed that until it was no longer there :)

PR #9293, opened 2021-04-15
Fixes #31323 - change pulp bulk_load_size to a setting

ianballou:
Changes the `bulk_load_size` to a Setting instead of a SETTING.

### DIFF
--- a/app/models/setting/content.rb
+++ b/app/models/setting/content.rb
@@ -150,7 +150,9 @@ class Setting::Content < Setting
       self.set('autosearch_while_typing', N_('For pages that support it, automatically perform search while typing in search input.'),
                true, N_('Autosearch')),
       self.set('autosearch_delay', N_('If Autosearch is enabled, delay in milliseconds before executing searches while typing.'),
-               500, N_('Autosearch delay'))
+               500, N_('Autosearch delay')),
+      self.set('bulk_load_size', N_('The number of items fetched from a single paged Pulp API call.'), 2000,
+               N_('Pulp bulk load size'))
     ]
   end
 
--- a/app/services/katello/pulp/pulp_content_unit.rb
+++ b/app/services/katello/pulp/pulp_content_unit.rb
@@ -46,7 +46,7 @@ module Katello
         Katello.pulp_server.extensions.send(self.name.demodulize.underscore)
       end
 
-      def self.pulp_units_batch_all(unit_ids = nil, page_size = SETTINGS[:katello][:pulp][:bulk_load_size])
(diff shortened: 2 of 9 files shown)


app/services/katello/pulp/pulp_content_unit.rb
```diff
@@ -46,7 +46,7 @@ def self.unit_handler
         Katello.pulp_server.extensions.send(self.name.demodulize.underscore)
       end
 
-      def self.pulp_units_batch_all(unit_ids = nil, page_size = SETTINGS[:katello][:pulp][:bulk_load_size])
+      def self.pulp_units_batch_all(unit_ids = nil, page_size = ::Setting::Content.find_by(name: "bulk_load_size").value)
```
jlsherrill, line 49: Can't you just do  Setting[:bulk_load_size] ?

ianballou: Ah yeah, that should work.

app/models/setting/content.rb
```diff
@@ -150,7 +150,9 @@ def self.default_settings
       self.set('autosearch_while_typing', N_('For pages that support it, automatically perform search while typing in search input.'),
                true, N_('Autosearch')),
       self.set('autosearch_delay', N_('If Autosearch is enabled, delay in milliseconds before executing searches while typing.'),
-               500, N_('Autosearch delay'))
+               500, N_('Autosearch delay')),
+      self.set('bulk_load_size', N_('The amount of items fetched from a single paged Pulp API call.'), 100,
```
jeremylenz, line 154:
```suggestion
      self.set('bulk_load_size', N_('The number of items fetched from a single paged Pulp API call.'), 100,
```
what are the units?

jeremylenz (commented):
Does this require an installer change?

app/models/setting/content.rb
ianballou: I'm not sure what to call the units besides items from the API response.

ianballou:
> Does this require an installer change?

Looks like this doesn't require an installer change. We never added `bulk_load_size` support in puppet-katello.

app/models/setting/content.rb
jeremylenz: Oh, I got it now :)  Yeah I think "number of items" is clear enough then.

jlsherrill (approved):

ianballou:
@jlsherrill I changed the default bulk_load_size to 2000 from 100.

jlsherrill:
oh thank you, that would have been bad :)

PR #9299, opened 2021-04-20
Fixes #32358 - properly display Less Than rpm requires

jlsherrill:


chris1984 (approved):
APJ, works

jlsherrill:
[test katello]

PR #9300, opened 2021-04-20
Fixes #32420 - Enable deb content for 2to3 migration

quba42:
https://projects.theforeman.org/issues/32420

### DIFF
--- a/app/services/katello/pulp3/migration.rb
+++ b/app/services/katello/pulp3/migration.rb
@@ -18,7 +18,9 @@ module Katello
       ].freeze
 
       CORRUPTABLE_CONTENT_TYPES = [
-        Katello::Rpm, Katello::FileUnit
+        Katello::Rpm,
+        Katello::FileUnit,
+        Katello::Deb
       ].freeze
 
       def self.repository_types_for_migration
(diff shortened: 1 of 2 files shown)


quba42:
My current migration test scenario throws the following error within the Pulp migration task:
```
duplicate key value violates unique constraint \"core_basedistribution_base_path_key\"
DETAIL:  Key (base_path)=(Default_Organization/Library/custom/Test/tuxedo) already exists.
```
Sounds like it is trying to create duplicate distributions somehow... Will need to investigate.

jlsherrill:
@quba42 did you run the switchover multiple times?  (or run migration after switchover?)    That could cause the issue.

quba42:
> @quba42 did you run the switchover multiple times? (or run migration after switchover?) That could cause the issue.

I did not manually run the switchover multiple times, but since we initially did not set the `use_pulp_2_for_deb: true` setting the smart proxy switched itself to Pulp 3 at least once before we set it back to Pulp 2 and performed the migration. I also ran a bunch of migrations in various stages of building up the dev box. Now that I know what the process is I plan to build the scenario one more time in a clean way and retest.

quba42:
I retested on a fresh box, and everything appears to have migrated and switched over successfully and as expected.
I will now open an issue to reference from this PR, so that I can lose the "Draft" status.

jlsherrill:
thanks!  i plan to test this out today

jlsherrill:
ACK  i tested this with a debian repo synced and in a content view.  It worked well after working through this error:  https://pulp.plan.io/issues/8613?#note-4

but thats something i think that needs to be handled in the migration plugin

quba42:
> ACK i tested this with a debian repo synced and in a content view. It worked well after working through this error: https://pulp.plan.io/issues/8613?#note-4
> 
> but thats something i think that needs to be handled in the migration plugin

I think I have a fix https://github.com/pulp/pulp-2to3-migration/pull/358, just need to test it...

PR #9304, opened 2021-04-21
Fixes #32351 - Race condition in katello-agent dynflow actions

jturel:
This PR fixes a race condition where the agent message (ex: install package) is handled by the client *before* the plan phase of the respective AgentAction is finished. Because the transaction of `plan` hasn't completed, the ClientMessageHandler is not able to find the DispatchHistory to update.

The fix is two-fold:
- Refactor `Katello::Agent::Dispatcher.dispatch` so that it no longer creates the `DispatchHistory` objects. That is handled by the new method `create_histories` on the same class. The result is that `dispatch` is much more succinct
- This refactor allows the `DispatchHistory` objects to be created in `plan`, while the `.dispatch` call has been moved to the `run` phase. This means the DispatchHistory will always be found by the ClientMessageHandler.

A bonus of this PR is that the `BulkAgentAction` now uses the batching capability provided by Dynflow (default batch size is 1000). `spawn_plans` is called for each batch. Reference: https://github.com/Dynflow/dynflow/blob/master/lib/dynflow/action/with_bulk_sub_plans.rb

**To test:**

- Register a client
- Fire up the rails server and curl some URL so that the event handlers start up
- From the rails console invoke some package actions on the client:

`User.as_anonymous_admin { ForemanTasks.sync_task(::Actions::Katello::Host::Package::Remove, ::Host.where(id: 1).first, content: ['screen']) }`

`User.as_anonymous_admin { ForemanTasks.sync_task(::Actions::Katello::BulkAgentAction, ::Actions::Katello::Host::Package::Install, ::Host.where(id: 1), content: ['screen']) }`
- Do this enough times and you'll likely see the error shown in the redmine description
- Check out this PR, repeat the testing steps and all should be well
-  Think critically about this change and if it introduces any new problems / try to break it

### DIFF
--- a/app/lib/actions/katello/agent_action.rb
+++ b/app/lib/actions/katello/agent_action.rb
@@ -15,22 +15,15 @@ module Actions
       def plan(host, options)
         action_subject(host, :hostname => host.name, :content => options[:content])
 
-        # if already dispatched by bulk action use the provided history ID
-        dispatch_history_id = options.dig(:dispatch_histories, host.id.to_s)
-
-        unless dispatch_history_id
-          histories = ::Katello::Agent::Dispatcher.dispatch(
-            self.class.agent_message,
-            [host.id],
-            content: options[:content]
-          )
-
-          dispatch_history_id = histories.first.id
-        end
+        dispatch_history_id = options.dig(:dispatch_histories, host.id.to_s) || ::Katello::Agent::Dispatcher.create_histories(
+          host_ids: [host.id]
+        ).first.id
 
         plan_self(
           host_id: host.id,
-          dispatch_history_id: dispatch_history_id
+          dispatch_history_id: dispatch_history_id,
+          content: options[:content],
+          bulk: options[:bulk]
         )
       end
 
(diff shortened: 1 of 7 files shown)


ehelms (approved):
From my perspective, I tested this on a nightly install and no longer saw the error that the Redmine identified. The code structure looks good.

app/lib/actions/katello/bulk_agent_action.rb
```diff
@@ -3,19 +3,32 @@ module Katello
...
           dispatch_histories: grouped_histories,
-          content: args[:content]
+          type: agent_action.agent_message,
+          content: args[:content],
+          bulk: true
         }
         super(agent_action, hosts, options)
       end
+
+      def spawn_plans
```
parthaa, line 22: who calls this ?

app/lib/actions/katello/agent_action.rb
```diff
@@ -15,43 +15,39 @@ def agent_action_type
...
-            self.class.agent_message,
-            [host.id],
-            content: options[:content]
-          )
-
-          dispatch_history_id = histories.first.id
-        end
+        dispatch_history_id = options.dig(:dispatch_histories, host.id.to_s) || ::Katello::Agent::Dispatcher.create_histories(
+          host_ids: [host.id]
+        ).first.id
```
parthaa, line 20: ?
```suggestion
        ).first.id.to_s
```

app/lib/actions/katello/bulk_agent_action.rb
jturel: Dynflow! https://github.com/Dynflow/dynflow/blob/d7bde98a9ac370c64e0ee2532758481d4956bb9c/lib/dynflow/action/with_bulk_sub_plans.rb#L18

parthaa (approved):
LGTM. ACK

PR #9305, opened 2021-04-22
Fixes #32283 - Implement new CV index page mockup

sjha4:
Mockup: https://marvelapp.com/prototype/1121a8de/screen/73175915

- [x] Update tests for new page.


webpack/components/RoutedTabs/RoutedTabs.js
```diff
@@ -24,7 +24,7 @@ const RoutedTabs = ({
     if (matchedTab) {
       history.push(buildLink(matchedTab.key));
     } else {
-      history.push(buildLink(tabs[defaultTabIndex].key)); // go to first tab if no tab selected
+      history.replace(buildLink(tabs[defaultTabIndex].key)); // go to first tab if no tab selected
     }
```
sjha4, line 28: I'll remove this history.replace here and tackle it with the issue we have open around back button to keep things separate.

sjha4:
@jeremylenz : This is ready for review too..I am working on some tests and have to file some issues for things out of scope of 4.1 currently on this page..

webpack/components/RoutedTabs/RoutedTabs.js
jeremylenz: It shouldn't always be `replace` anyway.  We don't want to lose the functionality of being able to move between different tabs with the Back button.

webpack/scenes/ContentViews/ContentViewsCountHeader.js
```diff
@@ -0,0 +1,63 @@
...
+            compositeCVCount += 1;
+          } else {
+            componentCVCount += 1;
+          }
+        });
+        setComponentCount(componentCVCount);
+        setCompositeCount(compositeCVCount);
+      }
+    },
+    [JSON.stringify(response), status],
```
jeremylenz, line 34: I still hate doing `JSON.stringify` in the dependency array.  Would you consider adding https://github.com/kentcdodds/use-deep-compare-effect ?

webpack/scenes/ContentViews/ContentViewsCountHeader.js
```diff
@@ -0,0 +1,63 @@
+import React, { useState, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import { Spinner } from '@patternfly/react-core';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { STATUS } from 'foremanReact/constants';
+import ContentViewIcon from './components/ContentViewIcon';
+
+const ContentViewsCountHeader = ({ response, status }) => {
+  const [componentCount, setComponentCount] = useState(<Spinner size="sm" />);
```
jeremylenz, line 9: Using JSX as state values seems very weird to me.  I think it's Patternfly table formatters etc. creating bad habits.  I'd stick to primitive values for state, and keep the JSX in the component's render method.  In other words:

1. ~~Store only the counts here in state - `useState(0)`~~
_update:_ I don't think state is needed here.  see below 
2. Have `ContentViewIcon` take a `loading` prop and render the `<Spinner>` there (this will also simplify your `useEffect` below.)

webpack/scenes/ContentViews/ContentViewsCountHeader.js
jeremylenz: In fact, I don't think this `useEffect` is needed at all:

1. see comment above about `loading` - can pass the value of `loadingResponse` directly to the child component
2. For results, you can just use `filter`:
```js
const compositeCVCount = results.filter(contentView => !!contentView.composite).length
const componentCVCount = results.filter(contentView => !contentView.composite).length
```

Come to think of it, this completely eliminates the need to store the counts with `useState` as well.

(If this causes too many renders or performance issues, could possibly wrap it in a `useMemo`.)

webpack/scenes/ContentViews/components/ContentViewIcon.js
```diff
@@ -3,24 +3,30 @@ import PropTypes from 'prop-types';
...
-const ContentViewIcon = ({ composite }) => {
+const ContentViewIcon = ({ composite, pre, post }) => {
   const props = {
     title: composite ? 'composite' : 'component',
     className: 'svg-icon-centered',
   };
   return (
-    <div className="svg-centered-container">
+    <div aria-label="content_view_icon" className="svg-centered-container">
+      {pre || null}
```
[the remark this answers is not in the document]
jeremylenz: Consider renaming `pre` to reflect its actual use, maybe `count` or `cvCount`

jeremylenz (commented):
Thanks for pushing forward @sjha4 

here's some initial comments :)

webpack/scenes/ContentViews/ContentViewsCountHeader.js
sjha4: I will change this. I was surprised when it worked this way at all..😄

webpack/scenes/ContentViews/ContentViewsCountHeader.js
```diff
@@ -0,0 +1,63 @@
...
+    [JSON.stringify(response), status],
+  );
+
+  if (status === STATUS.ERROR) {
+    return <></>;
+  }
+
+  return (
+    <>
+      <ContentViewIcon composite={false} pre={componentCount || '0'} post={__('component content views')} />
```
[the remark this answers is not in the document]
sjha4: Hah, not sure what the tooltip is.

On that note, there's work needed on this counter. This is supposed to have the total count in the org regardless of page we are on. We don't have an API call at the moment to support that so currently this reflects the results on the page. 

Also, it needs to be a link that populates the search field on the page. We will be adding that as a UX improvement bit but it's lower on priority list right now. I don't mind moving this component out of this PR.

webpack/scenes/ContentViews/ContentViewsCountHeader.js
sjha4: We need to add an API request here to show org wide content view counts. The current request is paged and changes upon changing pages or searching etc. 
This will probably need a new API endpoint and action to show the org-wide counts.

webpack/scenes/ContentViews/ContentViewsCountHeader.js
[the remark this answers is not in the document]
sjha4: I have added a couple of issue for 2 things here: 
1. https://projects.theforeman.org/issues/32431
- The content view counter.
- Related content views field in the expanded details.
2. https://projects.theforeman.org/issues/32432
- Add ability to pass search params in the URL and filter results on the target page.

Will remove the counter component out of this PR.

webpack/scenes/ContentViews/Table/ContentViewsTable.js
```diff
@@ -26,15 +26,12 @@ const ContentViewTable = ({ response, status, error }) => {
...
-          rowMapping,
-        );
+        const { newRowMappingIds, ...tableData } = tableDataGenerator(results);
         setTable(tableData);
-        setRowMapping(updatedRowMap);
+        setRowMappingIds(newRowMappingIds);
       }
     },
-    [results, JSON.stringify(rowMapping)], // use JSON to check obj values eq not reference eq
+    [results, JSON.stringify(rowMappingIds)], // use JSON to check obj values eq not reference eq
```
jeremylenz, line 34: Same comment from before about the `JSON.stringify`

webpack/scenes/ContentViews/Table/tableDataGenerator.js
```diff
@@ -1,137 +1,86 @@
...
-    {
-      compoundParent: offsetColumn + 2,
-      cells: [
         {
-          title: (
-            <DetailsContainer {...containerProps(offsetColumn + 3)}>
-              <VersionsExpansion {...expansionProps} />
-            </DetailsContainer>),
-          props: { colSpan: 6 },
+          title: description ? <>{__('Description: ')}{__(description)}</> : <EmptyContent emptyText="No Description" />,
```
jeremylenz, line 64:
```suggestion
          title: description ? <>{__('Description: ')}{__(description)}</> : <EmptyContent emptyText="No description" />,
```
also, wasn't `emptyText` renamed?

webpack/scenes/ContentViews/components/EnvironmentLabels.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Label } from '@patternfly/react-core';
+
+const EnvironmentLabels = ({ environments }) => {
+  const envLabels = environments.map(env =>
+    <React.Fragment key={env.id}><Label color="purple" href={`/lifecycle_environments/${env.id}`}>{`${env.name}`}</Label></React.Fragment>);
+  return envLabels;
+};
```
jeremylenz, line 9:
```suggestion
const EnvironmentLabels = environments => environments.map(env =>
    <React.Fragment key={env.id}><Label color="purple" href={`/lifecycle_environments/${env.id}`}>{`${env.name}`}</Label></React.Fragment>);
```

jeremylenz (commented):
When I expand a table row I see the dividing line between them is just about in the center:
expanded

Whereas in the mockup it seems much more natural because the top space is smaller:
mockup_expanded

I feel like there's too much space around the top and bottom of the content of the top half.  Can you see if the spacing can be adjusted to make it look better?

webpack/scenes/ContentViews/Table/tableDataGenerator.js
sjha4: Ya..that's in the other PR..Will rebase this once that is merged..

sjha4:
> When I expand a table row I see the dividing line between them is just about in the center:
> expanded
> 
> Whereas in the mockup it seems much more natural because the top space is smaller:
> mockup_expanded
> 
> I feel like there's too much space around the top and bottom of the content of the top half. Can you see if the spacing can be adjusted to make it look better?

I found a noPadding prop to Patternfly tables. Makes the expanded row look like this:

Screenshot from 2021-04-27 16-57-09

jeremylenz:
> I found a noPadding prop to Patternfly tables. Makes the expanded row look like this:

I think to make it look like the mockup the noPadding would need to be in the regular row, not the expanded.

sjha4:
> > I found a noPadding prop to Patternfly tables. Makes the expanded row look like this:
> 
> I think to make it look like the mockup the noPadding would need to be in the regular row, not the expanded.

Interesting.. I changed that and it looks better. Let me set up some time with you tomorrow to go over how it looks with all the combinations of TableVariant.compact and noPadding..

webpack/scenes/ContentViews/Table/ContentViewsTable.js
sjha4: We don't need to check rowMappingIds in the dependency array since the value is generated from results in the first place. I removed it.

webpack/scenes/ContentViews/Table/ContentViewsTable.js
```diff
@@ -26,15 +27,12 @@ const ContentViewTable = ({ response, status, error }) => {
...
-          rowMapping,
-        );
+        const { newRowMappingIds, ...tableData } = tableDataGenerator(results);
         setTable(tableData);
-        setRowMapping(updatedRowMap);
+        setRowMappingIds(newRowMappingIds);
       }
     },
-    [results, JSON.stringify(rowMapping)], // use JSON to check obj values eq not reference eq
+    [results],
```
jeremylenz, line 35: Any variable that's not defined within the effect still needs to go in the dep array :)
```suggestion
    [results, loadingResponse, setTable, setRowMappingIds],
```

sjha4: Aah..That's right.. Updated.. 👍🏽 

So the current combination of compact table with padding for expanded children looks like this:

Screenshot from 2021-04-28 09-34-36

jeremylenz: I think that looks great!!

jeremylenz:
The table row height for a non-published content view is smaller than for a published one:
rowheights

Not immediately sure what the best solution is.. maybe something with `min-height` somewhere.  Thoughts?

sjha4:
> The table row height for a non-published content view is smaller than for a published one:
> rowheights
> 
> Not immediately sure what the best solution is.. maybe something with `min-height` somewhere. Thoughts?

So it's coming from the top-margin here:

https://github.com/Katello/katello/blob/0f3220e6c7ce4dd4aad3f95720bf3c94a9d38d9f/webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentEnvironments.js#L8

sjha4:
> > The table row height for a non-published content view is smaller than for a published one:
> > rowheights
> > Not immediately sure what the best solution is.. maybe something with `min-height` somewhere. Thoughts?
> 
> So it's coming from the top-margin here:
> 
> https://github.com/Katello/katello/blob/0f3220e6c7ce4dd4aad3f95720bf3c94a9d38d9f/webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentEnvironments.js#L8

It's actually not that somehow..Still investigating..
Published versions are `565.600 * 89.600` while unpublished ones are at `565.600 * 60.800` regardless of the the styling above..

sjha4:
> > The table row height for a non-published content view is smaller than for a published one:
> > rowheights
> > Not immediately sure what the best solution is.. maybe something with `min-height` somewhere. Thoughts?
> 
> So it's coming from the top-margin here:
> 
> https://github.com/Katello/katello/blob/0f3220e6c7ce4dd4aad3f95720bf3c94a9d38d9f/webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentEnvironments.js#L8

It's actually something else..Looking..

sjha4:
Well, it seems to be the <Label /> that's adding the extra padding..
Something like this makes the height difference negligible..
`<Label style={{paddingBottom: '0.1em', paddingTop: '0.1em'}}..`
And a top margin for these at 0.1em as well..

Screenshot from 2021-04-29 11-16-03

I'll push the change..Let me know how it looks..

webpack/scenes/ContentViews/Table/tableDataGenerator.js
```diff
@@ -1,137 +1,86 @@
...
-    {
-      compoundParent: offsetColumn + 2,
-      cells: [
         {
-          title: (
-            <DetailsContainer {...containerProps(offsetColumn + 3)}>
-              <VersionsExpansion {...expansionProps} />
-            </DetailsContainer>),
-          props: { colSpan: 6 },
+          title: description ? <>{__('Description: ')}{__(description)}</> : <InactiveText text="No Description" />,
```
jeremylenz, line 64: `__(description)` shouldn't be translated, since `__` should only be used on literal strings.  (`__('Description: ')` is fine though, to be clear)

jeremylenz: What are your thoughts on displaying the description without the word "Description"?  I know it's in the mockup, but it seems unnecessary to me. 
```suggestion
          title: description ? <>{description}</> : <InactiveText text={__("No description")} />,
```
(also remember not to capitalize the second word :)

jeremylenz (commented):
> Something like this makes the height difference negligible..
> I pushed the change diff ..Let me know how it looks..

Do you think that's enough padding?  I was actually hoping that the row heights would be the more-padded one, not the less.  But I'm on the fence.. let me know your thoughts.

webpack/scenes/ContentViews/Table/tableDataGenerator.js
sjha4: I don't mind removing it..The description text should be self explanatory..

webpack/scenes/ContentViews/Table/tableDataGenerator.js
```diff
@@ -1,137 +1,86 @@
...
-    {
-      compoundParent: offsetColumn + 2,
-      cells: [
         {
-          title: (
-            <DetailsContainer {...containerProps(offsetColumn + 3)}>
-              <VersionsExpansion {...expansionProps} />
-            </DetailsContainer>),
-          props: { colSpan: 6 },
+          title: description ? __(description) : <InactiveText text={__('No Description')} />,
```
jeremylenz, line 64:
```suggestion
          title: description ? __(description) : <InactiveText text={__('No description')} />,
```

webpack/scenes/ContentViews/Table/tableDataGenerator.js
```diff
@@ -1,137 +1,86 @@
...
-    {
-      compoundParent: offsetColumn + 2,
-      cells: [
         {
-          title: (
-            <DetailsContainer {...containerProps(offsetColumn + 3)}>
-              <VersionsExpansion {...expansionProps} />
-            </DetailsContainer>),
-          props: { colSpan: 6 },
+          title: description ? __(description) : <InactiveText text={__('No description')} />,
```
jeremylenz, line 64: Oh.. and the `description` variable still shouldn't be translated..
```suggestion
          title: description ? description : <InactiveText text={__('No description')} />,
```

sjha4: Don't know why I thought that would work..Haven't tested translation bits on the UI..

jeremylenz:
Row heights look great now!

On non-published content views, the table row text is vertically centered :heavy_check_mark: 
On published content views, the table row content is a bit below center :x:

sjha4:
> On published content views, the table row content is a bit below center

Updated to split margin between top and bottom:
https://github.com/Katello/katello/compare/4fa00484b319e376ee3c60dfb0e53905da555197..d895d8eaa41ccd4208b52cb706841c38bd505a97

Screenshot from 2021-05-03 11-54-15

jeremylenz (approved):
Looks great!  :100: 

APJ :+1:

PR #9306, opened 2021-04-22
Fixes #32393 - Add deb structured mode publishing

m-bucher:


quba42:
This brings the Pulp 3 Katello deb functionality more in line with what the Pulp 2 Katello deb functionality, so we probably should have added it from the get go.

We could consider making this configurable in the future (either within Katello or alternatively directly within pulp_deb).
Even if we do that, using both should remain the default.

ianballou (approved):
This looks good to me! I agree that having a configuration option would be nice, perhaps on the debian repository creation page.  Simple publishing is required for file uploads so we might want a warning about that too. Can someone make an issue to do that once this PR is merged?

quba42:
I have created the following issue: https://projects.theforeman.org/issues/32514

We have an internal issue that I will link with this as well, so while I cannot give an exact time frame, we _will_ work on this at some point.

PR #9308, opened 2021-04-22
Fixes #32402 - Remove more traces of the background download policy

ianballou:
Removes more  traces of the old background download policy.


jlsherrill (approved):

swadeley (approved):
ACK

PR #9310, opened 2021-04-23
Refs #31910 - don't assert duration of the agent ping test

evgeni:


test/models/ping_test.rb
```diff
@@ -164,7 +164,6 @@ def test_ping_katello_agent_enabled
 
       assert_equal 'ok', result[:status]
       assert_equal '0 Processed, 0 Failed', result[:message]
-      assert_equal '0', result [:duration_ms]
```
ekohl, line 167: Does it make sense to assert it's present?
```ruby
assert_includes(result, :duration_ms)
```

evgeni: maybe? but we don't do that in the other tests either…

jturel: Perfectly find without it imo.

jturel (approved):
Thanks @evgeni !

PR #9312, opened 2021-04-23
Fixes #32405 - Permit DMI change when host is in build mode

jcpunk:
I've got an edge condition where I'd like to allow host DMI to change if and only if a host is in build mode (ie I know I'm rebuilding it)

This patch should add an extra switch to let hosts get through, but is disabled by default so (a) the existing behavior is preserved and (b) folks who want this know they want it and can turn it on.

### DIFF
--- a/app/models/setting/content.rb
+++ b/app/models/setting/content.rb
@@ -143,6 +143,9 @@ class Setting::Content < Setting
       self.set('host_profile_assume', N_("Allow new Host registrations to assume registered profiles with matching hostname " \
                   "as long as the registering DMI UUID is not used by another host."),
                true, N_('Host Profile Assume')),
+      self.set('host_profile_assume_build_can_change', N_("Allow Host registrations to bypass 'Host Profile Assume' " \
+                  "as long as the host is in build mode."),
+               false, N_('Host Profile Can Change In Build')),
       self.set('host_tasks_workers_pool_size', N_("Amount of workers in the pool to handle the execution of host-related tasks. When set to 0, the default queue will be used instead. Restart of the dynflowd/foreman-tasks service is required."),
                5, N_('Host Tasks Workers Pool Size')),
       self.set('applicability_batch_size', N_("Number of host applicability calculations to process per task."),
--- a/test/services/katello/registration_manager_test.rb
+++ b/test/services/katello/registration_manager_test.rb
@@ -71,6 +71,26 @@ module Katello
           assert @klass.validate_hosts(hosts, @org, @host.name, 'different-uuid')
         end
 
+        def test_host_profile_assume_build_mode_only_not_in_build
+          @host.subscription_facet.update(dmi_uuid: 'existing_system_uuid')
+          Setting[:host_profile_assume] = false
+          Setting[:host_profile_assume_build_can_change] = true
+          refute @host.build
+          error = assert_raises(Katello::Errors::RegistrationError) { @klass.validate_hosts(hosts, @org, @host.name, 'different-uuid') }
+          assert_match(/DMI UUID that differs/, error.message)
+        end
+
+        def test_host_profile_assume_build_mode_only_in_build
+          Setting[:host_profile_assume] = false
+          Setting[:host_profile_assume_build_can_change] = true
+          @host = FactoryBot.create(:host, :with_subscription, :managed, organization: @org, build: true)
+          @host.subscription_facet.update(dmi_uuid: 'existing_system_uuid')
+          # if a registering client is matched by hostname to an existing profile
+          # but its UUID has changed *and* is still unique, also it is in build mode
+          # then allow the registration when enabled
+          assert @klass.validate_hosts(hosts, @org, @host.name, 'different-uuid')
+        end
+
         def test_existing_uuid_and_name
           @host.subscription_facet.update(dmi_uuid: 'host3-uuid')
 
--- a/app/services/katello/registration_manager.rb
+++ b/app/services/katello/registration_manager.rb
@@ -50,6 +50,16 @@ module Katello
         Katello::Host::SubscriptionFacet::DMI_UUID_ALLOWED_DUPS
       end
 
+      def dmi_uuid_change_allowed?(host, host_uuid_overridden)
+        if host_uuid_overridden
+          true
+        elsif host.build && Setting[:host_profile_assume_build_can_change]
+          true
+        else
+          Setting[:host_profile_assume]
+        end
+      end
+
       def find_existing_hosts(host_name, host_uuid)
         query = ::Host.unscoped.where("#{::Host.table_name}.name = ?", host_name)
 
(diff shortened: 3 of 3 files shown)


jturel:
Thanks @jcpunk we'll be reviewing this soon!

jturel:
@jcpunk is the name/hostname of the host also changing when it's getting rebuilt in your case?

jcpunk:
Alas, the hostname remains the same.

jturel:
Alright, I've looked at this code a bit more closely and I now grok the change and see the value in it!

I'd like to request a few changes with maintainability and readability in mind. The below refactor removes the `unless` usage (I love the keyword but it quickly makes code unreadable), and the `dmi_uuid_change_allowed?` method establishes a clear precedence of scenarios that allow the UUID to change. I also recommend changing the setting name to `host_profile_assume_build_mode_only` (or similar) to establish a relationship between the two to someone looking at the settings list.  `host_profile_assume` must be true for the new setting to take effect. Obviously, ensure that it meets your expectations. :)

inside validate_hosts method
```
        if hosts_size == 1                                                                                                                                                                                                                                                      
          host = hosts.first                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                
          if host.name == host_name                                                                                                                                                                                                                                             
            current_dmi_uuid = host.subscription_facet&.dmi_uuid                                                                                                                                                                                                                
            dmi_uuid_changed = current_dmi_uuid && current_dmi_uuid != host_uuid                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                
            if dmi_uuid_changed && !dmi_uuid_change_allowed?(host, host_uuid_overridden)                                                                                                                                                                                        
              registration_error("This host is reporting a DMI UUID that differs from the existing registration.")                                                                                                                                                              
            end                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                
            return true                                                                                                                                                                                                                                                         
          end                                                                                                                                                                                                                                                                   
        end
```

new method in RegistrationManager:

```
      def dmi_uuid_change_allowed?(host, host_uuid_overridden)                                                                                                                                                                                                                  
        if host_uuid_overridden                                                                                                                                                                                                                                                 
          true                                                                                                                                                                                                                                                                  
        elsif Setting[:host_profile_assume] && Setting[:host_profile_assume_build_mode_only]                                                                                                                                                                                    
          host.build                                                                                                                                                                                                                                                            
        elsif Setting[:host_profile_assume]                                                                                                                                                                                                                                     
          true                                                                                                                                                                                                                                                                  
        else                                                                                                                                                                                                                                                                    
          false                                                                                                                                                                                                                                                                 
        end                                                                                                                                                                                                                                                                     
      end
```

tests:
```
        def test_host_profile_assume_build_mode_only_not_in_build                                                                                                                                                                                                               
          @host.subscription_facet.update(dmi_uuid: 'existing_system_uuid')                                                                                                                                                                                                     
          Setting[:host_profile_assume] = true                                                                                                                                                                                                                                  
          Setting[:host_profile_assume_build_mode_only] = true                                                                                                                                                                                                                  
          refute @host.build                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                
          error = assert_raises(Katello::Errors::RegistrationError) { @klass.validate_hosts(hosts, @org, @host.name, 'different-uuid') }                                                                                                                                        
          assert_match(/DMI UUID that differs/, error.message)                                                                                                                                                                                                                  
        end                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                
        def test_host_profile_assume_build_mode_only_in_build                                                                                                                                                                                                                   
          Setting[:host_profile_assume] = true                                                                                                                                                                                                                                  
          Setting[:host_profile_assume_build_mode_only] = true                                                                                                                                                                                                                  
          @host = FactoryBot.create(:host, :with_subscription, :managed, organization: @org, build: true)                                                                                                                                                                       
          @host.subscription_facet.update(dmi_uuid: 'existing_system_uuid')                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                
          # if a registering client is matched by hostname to an existing profile                                                                                                                                                                                               
          # but its UUID has changed *and* is still unique, also it is in build mode                                                                                                                                                                                            
          # then allow the registration when enabled                                                                                                                                                                                                                            
        end 
```

jcpunk:
I think I successfully cloned this over to my tree.  I thought `host_profile_assume_build_can_change` might be cleaner to evaluate.  With `ok` I wasn't sure if folks would know which thing was "ok".

app/models/setting/content.rb
```diff
@@ -143,6 +143,9 @@ def self.default_settings
       self.set('host_profile_assume', N_("Allow new Host registrations to assume registered profiles with matching hostname " \
                   "as long as the registering DMI UUID is not used by another host."),
                true, N_('Host Profile Assume')),
+      self.set('host_profile_assume_build_can_change', N_("Allow Host registrations to bypass 'Host Profile Assume' " \
```
jturel, line 146: @jcpunk I think there's a discrepancy between this description and the actual behavior.

The wording states that the new setting supercedes `host_profile_assume` if the host is in build mode ie they are not coupled. The implementation states that `if host_profile_assume is true and host_profile_assume_build_can_change is true and the host is in build mode, then the DMI UUID may change when registering` ie they are coupled.

if the setting description is correct then I think the logic of `dmi_uuid_change_allowed?` would be more like this:

```
      def dmi_uuid_change_allowed?(host, host_uuid_overridden)                                                                                                                                
        if host_uuid_overridden                                                                                                                                                               
          true                                                                                                                                                                                
        elsif host.build && Setting[:host_profile_assume_build_can_change]                                                                                                                    
          true                                                                                                                                                                                
        else                                                                                                                                                                                  
          Setting[:host_profile_assume]                                                                                                                                                       
        end                                                                                                                                                                                   
      end 
```

In other words host_uuid_overridden > host.build & host_profile_assume_build_can_change > host_profile_assume

jcpunk: Yeah, that does look better.

I'll have a new patch up shortly.

test/services/katello/registration_manager_test.rb
```diff
@@ -71,6 +71,26 @@ def test_existing_host_mismatch_uuid
           assert @klass.validate_hosts(hosts, @org, @host.name, 'different-uuid')
         end
 
+        def test_host_profile_assume_build_mode_only_not_in_build
+          @host.subscription_facet.update(dmi_uuid: 'existing_system_uuid')
+          Setting[:host_profile_assume] = true
```
jturel, line 76: Since we've established this feature is works outside the bounds of host_profile_assume, this should be changed to `false` for both tests

jturel (commented):
Just left one note on the tests, and this is good to go. Sorry for the delay here.

Your second PR will need a rebase on top of master branch once this is merged, but I can continue to test it from my side before you push those changes :)

test/services/katello/registration_manager_test.rb
jcpunk: I believe I've made the requested changes.

jturel (approved):
ACK.

Nice work @jcpunk !!

PR #9313, opened 2021-04-23
Fixes #32406 - Prevent re-register unless host is in build mode

jcpunk:
This should help keep hosts from registering under a name that is already in use unless that host is in build mode.

### DIFF
--- a/app/models/setting/content.rb
+++ b/app/models/setting/content.rb
@@ -146,6 +146,8 @@ class Setting::Content < Setting
       self.set('host_profile_assume_build_can_change', N_("Allow Host registrations to bypass 'Host Profile Assume' " \
                   "as long as the host is in build mode."),
                false, N_('Host Profile Can Change In Build')),
+      self.set('host_re_register_build_only', N_("Allow hosts to re-register themselves only when they are in build mode"),
+               false, N_('Host Can Re-Register Only In Build')),
       self.set('host_tasks_workers_pool_size', N_("Amount of workers in the pool to handle the execution of host-related tasks. When set to 0, the default queue will be used instead. Restart of the dynflowd/foreman-tasks service is required."),
                5, N_('Host Tasks Workers Pool Size')),
       self.set('applicability_batch_size', N_("Number of host applicability calculations to process per task."),
--- a/app/services/katello/registration_manager.rb
+++ b/app/services/katello/registration_manager.rb
@@ -87,6 +87,10 @@ module Katello
           host = hosts.first
 
           if host.name == host_name
+            if !host.build && Setting[:host_re_register_build_only]
+              registration_error("Host with name %{host_name} is currently registered but not in build mode (host_re_register_build_only==True). Unregister the host manually or put it into build mode to continue.", host_name: host_name)
+            end
+
             current_dmi_uuid = host.subscription_facet&.dmi_uuid
             dmi_uuid_changed = current_dmi_uuid && current_dmi_uuid != host_uuid
             if dmi_uuid_changed && !dmi_uuid_change_allowed?(host, host_uuid_overridden)
--- a/test/services/katello/registration_manager_test.rb
+++ b/test/services/katello/registration_manager_test.rb
@@ -91,6 +91,19 @@ module Katello
           assert @klass.validate_hosts(hosts, @org, @host.name, 'different-uuid')
         end
 
+        def test_re_register_build_mode
+          @host = FactoryBot.create(:host, :with_subscription, :managed, organization: @org)
+          @host.subscription_facet.update(dmi_uuid: 'existing_system_uuid')
+          refute @host.build
+          Setting[:host_re_register_build_only] = true
+
+          error = assert_raises(Katello::Errors::RegistrationError) { @klass.validate_hosts(hosts, @org, @host.name, nil) }
+          assert_match(/currently registered/, error.message)
+
+          @host.update(build: true)
+          assert @klass.validate_hosts(hosts, @org, @host.name, 'existing_system_uuid')
+        end
+
         def test_existing_uuid_and_name
           @host.subscription_facet.update(dmi_uuid: 'host3-uuid')
 


app/services/katello/registration_manager.rb
```diff
@@ -79,6 +79,9 @@ def validate_hosts(hosts, organization, host_name, host_uuid, host_uuid_overridd
           if host.name == host_name
             unless Setting[:host_profile_assume] || host.subscription_facet.nil? || host.build || host_uuid_overridden
               current_dmi_uuid = host.subscription_facet.dmi_uuid
+              if not Setting[:host_re_register] and not host.build and current_dmi_uuid
```
jlsherrill, line 82: there's some rubocop errors:

 Use ! instead of not.
Use && instead of and.

jcpunk: I'll update and re-push.

jcpunk:
Looks like I'll need some help with the testing......

jturel:
@jcpunk I think I am missing how this is different than the host_profile_assume setting. Does that setting work for your case, and if not, let me know how it doesn't :)

app/models/setting/content.rb
```diff
@@ -136,6 +136,9 @@ def self.default_settings
                "latest package to solve a dependency even if it already does exist in the repository."),
                'conservative', N_('Content View Dependency Solving Algorithm'), nil,
                :collection => dependency_solving_options),
+      self.set('host_re_register', N_("Allow a Host to register itself even if another host by that name exists " \
```
jturel, line 139: This is already the default behavior, at least as the setting is written

app/services/katello/registration_manager.rb
```diff
@@ -79,6 +79,9 @@ def validate_hosts(hosts, organization, host_name, host_uuid, host_uuid_overridd
           if host.name == host_name
             unless Setting[:host_profile_assume] || host.subscription_facet.nil? || host.build || host_uuid_overridden
               current_dmi_uuid = host.subscription_facet.dmi_uuid
+              if !Setting[:host_re_register] && !host.build && current_dmi_uuid
```
jturel, line 82: I see the goal in this change but I think it needs to be handled in a different way for two reasons:

- host_re_register (as you've pointed out) is already the current behavior, so adding a setting for it feels strange
- the fact that the setting is bypassed if the host is in build mode seems to be a bit of an undocumented loophole


potential solution:

Rename the setting to `host_re_register_build_only` with the text being something like "Only allow previously registered hosts to re-register if they are in build mode"

Then the logic becomes something like:

```
if !host.build && Setting[:host_re_register_build_only] #putting the build check first avoids DB hit on setting lookup if host is in build mode already
 # raise error
end

#dmi uuid uniqueness enforcement happens down here
```

jturel (commented):
Left some notes on the approach, let me know your thoughts. To avoid even more merge conflicts I suggest not making any changes until https://github.com/Katello/katello/pull/9312 is merged

test/services/katello/registration_manager_test.rb
```diff
@@ -71,6 +71,24 @@ def test_existing_host_mismatch_uuid
           assert @klass.validate_hosts(hosts, @org, @host.name, 'different-uuid')
         end
 
+        def test_existing_host_already_registered
```
jturel, line 74: If you go with my approach above, this test would become something like:

```
        def test_re_register_build_mode                                                                                                                                                       
          @host.build = false                                                                                                                                                                 
          Setting[:host_re_register_build_only] = true                                                                                                                                        
                                                                                                                                                                                              
          error = assert_raises(Katello::Errors::RegistrationError) { @klass.validate_hosts(hosts, @org, @host.name) }                                                                        
          assert_match(/existing registration/, error.message)                                                                                                                                
                                                                                                                                                                                              
          @host.build = true                                                                                                                                                                  
          assert @klass.validate_hosts(hosts, @org, @host.name, 'existing_system_uuid')                                                                                                       
        end 
```

app/services/katello/registration_manager.rb
jcpunk: This makes **_a lot_** more sense!

I'll see about getting this written up as listed here once #9312 is merged.

jcpunk: I've taken a stab at the requested changes

test/services/katello/registration_manager_test.rb
jcpunk: I think I've adapted the test correctly....

app/services/katello/registration_manager.rb
```diff
@@ -83,6 +83,10 @@ def validate_hosts(hosts, organization, host_name, host_uuid, host_uuid_overridd
                              org_name: organization.name, host_name: host_name)
         end
 
+        if !host.build && Setting[:host_re_register_build_only]
```
jturel, line 86: This new block should come after the `  if host.name == host_name` line just below so that `host` is defined (but before the DMI UUID checking)

jcpunk: Oops.. pushed a fix.

jturel:
testing / reviewing now!

app/services/katello/registration_manager.rb
```diff
@@ -87,6 +87,10 @@ def validate_hosts(hosts, organization, host_name, host_uuid, host_uuid_overridd
           host = hosts.first
 
           if host.name == host_name
+            if !host.build && Setting[:host_re_register_build_only]
+              registration_error("Host with name %{host_name} is currently registered but not in build mode (host_re_register_build_only==True).  You can unregister the host manually or put it into build mode to continue.", host_name: host_name)
```
jturel, line 91: Looks like there are two spaces in between the sentences. Also recommend removing "you" ie `Unregister the host manually or <snip>`

test/services/katello/registration_manager_test.rb
```diff
@@ -91,6 +91,19 @@ def test_host_profile_assume_build_mode_only_in_build
           assert @klass.validate_hosts(hosts, @org, @host.name, 'different-uuid')
         end
 
+        def test_re_register_build_mode
```
jturel, line 94: These adjustment got the test working as expected for me:

```
[vagrant@centos7-katello-devel katello]$ git diff
diff --git a/test/services/katello/registration_manager_test.rb b/test/services/katello/registration_manager_test.rb
index e6bc4c3..b72819d 100644
--- a/test/services/katello/registration_manager_test.rb
+++ b/test/services/katello/registration_manager_test.rb
@@ -94,13 +94,13 @@ module Katello
         def test_re_register_build_mode
           @host = FactoryBot.create(:host, :with_subscription, :managed, organization: @org)
           @host.subscription_facet.update(dmi_uuid: 'existing_system_uuid')
-          @host.build = false
+          refute @host.build
           Setting[:host_re_register_build_only] = true
 
-          error = assert_raises(Katello::Errors::RegistrationError) { @klass.validate_hosts(hosts, @org, @host.name) }
+          error = assert_raises(Katello::Errors::RegistrationError) { @klass.validate_hosts(hosts, @org, @host.name, nil) }
           assert_match(/currently registered/, error.message)
 
-          @host.build = true
+          @host.update(build: true)
           assert @klass.validate_hosts(hosts, @org, @host.name, 'existing_system_uuid')
         end

```

jturel (changes requested):
The new behavior works perfectly! Just two small adjustments :)

app/services/katello/registration_manager.rb
jcpunk: Fixed :)

test/services/katello/registration_manager_test.rb
jcpunk: I think I've pushed a fix with the changes applied correctly.  My local testing is all messed up and everything fails.....

jturel (approved):
Changes look great. I'll merge so long as the build goes green!

jturel:
Merging! Thanks @jcpunk I greatly enjoyed our collaboration. Hopefully it won't be long before we do so again :)

PR #9318, opened 2021-04-27
Fixes #32426 - pass Docker-Content-Digest on container requests

jlsherrill:


jlsherrill:
The best way to test this would be to run a 4.0 'upgrade' pipeline.  You will likely still hit other errors though because of these two issues: https://pulp.plan.io/issues/8629 https://pulp.plan.io/issues/8631
so maybe its better to just test that this doesn't break podman/docker clients?

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -161,11 +161,9 @@ def pull_manifest
       end
 
       if (manifest_response = redirect_client { Resources::Registry::Proxy.get(@_request.fullpath, headers) })
-        #when pulp 2 is removed, this should no longer be needed, and all clients should be redirected
-        logger.debug filter_sensitive_data(manifest_response)
+        #for some requests, we get a redirect, but for others we get the actual manifest in response
```
jlsherrill, line 164: also i discovered that pulp doesn't always redirect on manifest requests, so i changed the comment here.

ianballou (approved):
I tested a bunch of `docker` and `podman` pulls, and they still work! Seems fine to me.

PR #9323, opened 2021-04-29
Fixes #32455 - fix untranslated strings

Ron-Lavi:
sharing the findings of `eslint-plugin-translation`,
@jeremylenz could you take a look? thanks :)

### DIFF
--- a/webpack/scenes/ContentViews/Copy/CopyContentViewModal.js
+++ b/webpack/scenes/ContentViews/Copy/CopyContentViewModal.js
@@ -1,17 +1,19 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Modal, ModalVariant } from '@patternfly/react-core';
+import { FormattedMessage } from 'react-intl';
+import { translate as __ } from 'foremanReact/common/I18n';
 import CopyContentViewForm from './CopyContentViewForm';
 
 const CopyContentViewModal = ({
   cvId, cvName, show, setIsOpen,
 }) => {
   const description = (
-    <p>
-      This will create a copy of <b>{cvName}</b>, including details,
-      repositories, and filters. Generated data such
-      as history, tasks and versions will not be copied.
-    </p>
+    <FormattedMessage
+      id="copy-cv-description"
+      values={{ cv: <b>{cvName}</b> }}
+      defaultMessage={__('This will create a copy of {cv}, including details, repositories, and filters. Generated data such as history, tasks and versions will not be copied.')}
+    />
   );
   return (
     <Modal
--- a/webpack/scenes/Subscriptions/SubscriptionsPage.js
+++ b/webpack/scenes/Subscriptions/SubscriptionsPage.js
@@ -1,4 +1,5 @@
 import React, { Component } from 'react';
+import { FormattedMessage } from 'react-intl';
 import PropTypes from 'prop-types';
 import Immutable from 'seamless-immutable';
 import { translate as __ } from 'foremanReact/common/I18n';
@@ -217,6 +218,21 @@ class SubscriptionsPage extends Component {
       },
     };
 
+    const SCAAlert = simpleContentAccess ? (
+      <Alert type="info">
+        <FormattedMessage
+          id="sca-alert"
+          values={{
+            subscriptionWatch: <a href={SUBSCRIPTION_WATCH_URL} target="_blank" rel="noreferrer">{__('Subscription Watch')}</a>,
+            br: <br />,
+          }}
+          defaultMessage={__(`This organization has Simple Content Access enabled.
+          Hosts are not required to have subscriptions attached to access repositories.
+          {br}
+          Learn more about your overall subscription usage at {subscriptionWatch}.`)}
+        />
+      </Alert>
+    ) : null;
     return (
       <Grid bsClass="container-fluid">
         <Row>
@@ -255,15 +271,7 @@ class SubscriptionsPage extends Component {
             />
 
             <div id="subscriptions-table" className="modal-container">
-              {simpleContentAccess && (
-                <Alert type="info">
-                  This organization has Simple Content Access enabled.
-                  Hosts are not required to have subscriptions attached to access repositories.
-                  <br />
-                  Learn more about your overall subscription usage at
-                  {' '}<a href={SUBSCRIPTION_WATCH_URL} target="_blank" rel="noreferrer">Subscription Watch</a>.
-                </Alert>
-              )}
+              {SCAAlert}
               <SubscriptionsTable
                 canManageSubscriptionAllocations={canManageSubscriptionAllocations}
                 loadSubscriptions={this.props.loadSubscriptions}
--- a/webpack/__mocks__/react-intl/index.js
+++ b/webpack/__mocks__/react-intl/index.js
@@ -1,3 +1,7 @@
+/* eslint-disable react/prop-types */
+import React from 'react';
+
 export const addLocaleData = jest.fn();
 export const FormattedRelative = jest.fn();
 export const intlShape = jest.fn();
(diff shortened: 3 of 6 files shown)


webpack/scenes/ContentViews/Details/Repositories/LastSync.js
```diff
@@ -1,12 +1,13 @@
...
 import { CheckCircleIcon, ExclamationTriangleIcon, CloseIcon } from '@patternfly/react-icons';
 import { foremanUrl } from 'foremanReact/common/helpers';
+import { translate as __ } from 'foremanReact/common/I18n';
 import PropTypes from 'prop-types';
 
 const LastSync = ({ lastSyncWords, lastSync }) => {
   if (lastSync && lastSyncWords) {
     let Icon;
-    let color = 'black';
+    let color = __('black');
```
jeremylenz, line 10: This will end up as a CSS style, so it shouldn't be translated

webpack/scenes/ContentViews/Copy/CopyContentViewModal.js
```diff
@@ -1,17 +1,21 @@
...
 }) => {
   const description = (
-    <p>
-      This will create a copy of <b>{cvName}</b>, including details,
-      repositories, and filters. Generated data such
-      as history, tasks and versions will not be copied.
-    </p>
+    <p
+      dangerouslySetInnerHTML={{
+        __html: __(`This will create a copy of <b>${cvName}</b>, including details,
```
jeremylenz, line 14: `dangerouslySetInnerHTML` is bad, no?  Too bad we don't have any other way to translate text with formatting. :thinking:

webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentVersion.js
```diff
@@ -20,7 +20,7 @@ const ComponentVersion = ({ componentCV }) => {
   const manualVersionText = (latestVersion === version) ? __('Latest version') : __(`New version is available: Version ${latestVersion}`);
   return (
     <>
-      <Link to={urlBuilder('labs/content_views', '', id)}> Version {version}</Link>
+      <Link to={urlBuilder('labs/content_views', '', id)}> {__('Version')} {version}</Link>
```
jeremylenz, line 23: This will end up as a string like "Version 5.0", so "Version" alone might be hard to translate.  Not sure what to do about it though; should probably mark it for translation anyway.

webpack/scenes/Subscriptions/SubscriptionsPage.js
```diff
@@ -271,11 +271,11 @@ class SubscriptionsPage extends Component {
...
               {simpleContentAccess && (
                 <Alert type="info">
-                  This organization has Simple Content Access enabled.
-                  Hosts are not required to have subscriptions attached to access repositories.
+                  {__(`This organization has Simple Content Access enabled.
+                  Hosts are not required to have subscriptions attached to access repositories.`)}
                   <br />
-                  Learn more about your overall subscription usage at
-                  {' '}<a href={SUBSCRIPTION_WATCH_URL} target="_blank" rel="noreferrer">Subscription Watch</a>.
+                  {__('Learn more about your overall subscription usage at')}
```
jeremylenz, line 277: This is another partial that will be difficult to translate without context.

jeremylenz (changes requested):
Thanks @laviro!

Should we also disable the eslint rule for the two unmodified files that github is complaining about?

webpack/scenes/ContentViews/Copy/CopyContentViewModal.js
laviro: yeah, we should investigate more about it,
though for now, I think it's the only solution...

webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentVersion.js
laviro: using `sprintf`, although for `Version X` I don't know how many option there are.

webpack/scenes/Subscriptions/SubscriptionsPage.js
laviro: Using dangerous sprintf here :D

webpack/scenes/ContentViews/Details/Repositories/LastSync.js
laviro: in french it could have a nice sound: the new css `noir` :D

webpack/scenes/Subscriptions/SubscriptionsPage.js
jeremylenz: What does dangerous sprintf mean?

laviro: just kidding, I used the `sprintf` method from core and injected it with `dangerouslySetInnerHTML` to keep the html tags

webpack/scenes/ContentViews/Details/Histories/ContentViewHistories.js
```diff
@@ -55,9 +55,9 @@ const ContentViewHistories = ({ cvId }) => {
       const taskType = task ? task.label : taskTypes[action];
 
       if (taskType === taskTypes.removal) {
-        return <React.Fragment> Deleted from <Label key="1" color="blue" href={`/lifecycle_environments/${environment.id}`}>{`${environment.name}`}</Label>{}</React.Fragment>;
+        return <React.Fragment> {__('Deleted from')} <Label key="1" color="blue" href={`/lifecycle_environments/${environment.id}`}>{`${environment.name}`}</Label>{}</React.Fragment>;
       } else if (taskType === taskTypes.promotion) {
-        return <React.Fragment> Promoted to <Label key="2" color="blue" href={`/lifecycle_environments/${environment.id}`}>{`${environment.name}`}</Label>{}</React.Fragment>;
+        return <React.Fragment> {__('Promoted to')} <Label key="2" color="blue" href={`/lifecycle_environments/${environment.id}`}>{`${environment.name}`}</Label>{}</React.Fragment>;
       } else if (taskType === taskTypes.publish) {
         return ('Published new version');
```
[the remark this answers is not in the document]
laviro: well the eslint rule, for now, catches JSX strings and variables declaration
next thing would be to catch also strings like this one, but maybe first we should change it into the syntax of `return 'string'` without the braces 👼🏼

webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentVersion.js
jeremylenz: It's the number of content view versions the user creates; I'd say it's usually between 1 and 20

jeremylenz (approved):
@laviro Needs a rebase but otherwise LGTM :+1:

jeremylenz:
@laviro what's the status here?

laviro:
Hey @jeremylenz, thanks for pinging me,
Looks like I will need to re-write the eslint package structure to modern structure because it does issues in other plugins, but we can use its suggestions and keep them in this PR

jeremylenz:
@laviro is this PR still valid?

Ron-Lavi:
updated after a long time :)
Ready for another round @jeremylenz

jeremylenz (commented):
Thanks for updating @Ron-Lavi!

@amirfefer has alerted us to a way to deal with formatting in translations that I think is better than `dangerouslySetInnerHTML`, could you take a look?  https://github.com/theforeman/foreman/pull/8837/files

Ron-Lavi:
@jeremylenz ready for another round

jeremylenz (approved):
Thanks @Ron-Lavi 

everything works as expected and looks good!

I would suggest changing the name of the PR before merging, since we didn't actually add the eslint plugin.

Ron-Lavi:
done @jeremylenz

PR #9324, opened 2021-04-29
Fixes #31023 - add katello_agent_enabled macro for provisioning

jturel:


app/lib/katello/concerns/base_template_scope_extensions.rb
```diff
@@ -278,6 +278,11 @@ def sortable_version(version)
         Util::Package.sortable_version(version)
       end
 
+      apipie :method, 'Returns true if Katello Agent infrastructure is enabled on the server'
+      def katello_agent_enabled
```
ekohl, line 282: We also have macros which follow the `?` convention. Would this be clearer?
```suggestion
      def katello_agent_enabled?
```

ekohl (approved):

jlsherrill (approved):

jlsherrill:
don't forget to squash

app/lib/katello/concerns/base_template_scope_extensions.rb
```diff
@@ -278,6 +278,11 @@ def sortable_version(version)
         Util::Package.sortable_version(version)
       end
 
+      apipie :method, 'Returns true if Katello Agent infrastructure is enabled on the server'
+      def katello_agent_enabled?
+        Katello.with_katello_agent?
```
jeremylenz, line 283: Should the `::` be prepended?

jturel: Works without in this case

jeremylenz: Oh, okay, uh, ACK then :yum:

PR #9327, opened 2021-04-30
Fixes #32464 - Inherited hostgroup values are not passed to host

hao-yu:
User is unable to create a host using a nested hostgroup because all values that inherited from the parent host group are not set, such has content_view_id, content_source_id, lifecycle_environment_id and kickstart_repository_id. This commit fixes the issue.


tbrisker:
[test katello]

ianballou:
[test katello]

tbrisker (approved):
Test failure seems unrelated, an error with some node module missing for angular tests.

tbrisker:
@jjeffers @jeremylenz this looks like a good candidate for cherry-picking to stable branches.

PR #9330, opened 2021-04-30
Fixes #32487: Use Foreman client certificates to communicate with Pulp

ehelms:
See https://community.theforeman.org/t/rfc-using-foreman-client-certificates-for-all-backend-service-communication/23391 for some reference

Requires:
  * https://github.com/theforeman/puppet-pulpcore/pull/186 -- MERGED
  * https://github.com/theforeman/puppet-foreman_proxy_content/pull/350

### DIFF
--- a/app/lib/katello/resources/registry.rb
+++ b/app/lib/katello/resources/registry.rb
@@ -31,9 +31,9 @@ module Katello
             uri = URI.parse(content_app_url)
             self.prefix = "/pulpcore_registry/"
             self.site = "#{uri.scheme}://#{uri.host}:#{uri.port}"
-            self.ssl_client_cert = ::Cert::Certs.ssl_client_cert
-            self.ssl_client_key = ::Cert::Certs.ssl_client_key
             self.ca_cert_file = Setting[:ssl_ca_file]
+            pulp_primary.pulp3_ssl_configuration(self)
+
             self
           end
 
--- a/app/services/cert/certs.rb
+++ b/app/services/cert/certs.rb
@@ -12,20 +12,28 @@ module Cert
       File.read(SETTINGS[:katello][:candlepin][:ca_cert_file])
     end
 
-    def self.ssl_client_cert
-      @ssl_client_cert ||= OpenSSL::X509::Certificate.new(File.read(ssl_client_cert_filename))
+    def self.ssl_client_cert(use_admin_as_cn_cert: false)
+      @ssl_client_cert ||= OpenSSL::X509::Certificate.new(File.read(ssl_client_cert_filename(use_admin_as_cn_cert: use_admin_as_cn_cert)))
     end
 
-    def self.ssl_client_cert_filename
-      Setting['pulp_client_cert']
+    def self.ssl_client_cert_filename(use_admin_as_cn_cert: false)
+      if use_admin_as_cn_cert
+        Setting[:pulp_client_cert]
+      else
+        Setting[:ssl_certificate]
+      end
     end
 
-    def self.ssl_client_key
-      @ssl_client_key ||= OpenSSL::PKey::RSA.new(File.read(ssl_client_key_filename))
+    def self.ssl_client_key(use_admin_as_cn_cert: false)
+      @ssl_client_key ||= OpenSSL::PKey::RSA.new(File.read(ssl_client_key_filename(use_admin_as_cn_cert: use_admin_as_cn_cert)))
     end
 
-    def self.ssl_client_key_filename
-      Setting['pulp_client_key']
+    def self.ssl_client_key_filename(use_admin_as_cn_cert: false)
+      if use_admin_as_cn_cert
+        Setting[:pulp_client_key]
+      else
+        Setting[:ssl_priv_key]
+      end
     end
 
     def self.verify_ueber_cert(organization)
--- a/app/models/katello/concerns/smart_proxy_extensions.rb
+++ b/app/models/katello/concerns/smart_proxy_extensions.rb
@@ -141,12 +141,14 @@ module Katello
       end
 
       def pulp3_ssl_configuration(config)
+        legacy_pulp_cert = !self.setting(PULP3_FEATURE, 'client_authentication')&.include?('client_certificates')
+
         if Faraday.default_adapter == :excon
-          config.ssl_client_cert = ::Cert::Certs.ssl_client_cert_filename
-          config.ssl_client_key = ::Cert::Certs.ssl_client_key_filename
+          config.ssl_client_cert = ::Cert::Certs.ssl_client_cert_filename(use_admin_as_cn_cert: legacy_pulp_cert)
+          config.ssl_client_key = ::Cert::Certs.ssl_client_key_filename(use_admin_as_cn_cert: legacy_pulp_cert)
         elsif Faraday.default_adapter == :net_http
-          config.ssl_client_cert = ::Cert::Certs.ssl_client_cert
-          config.ssl_client_key = ::Cert::Certs.ssl_client_key
+          config.ssl_client_cert = ::Cert::Certs.ssl_client_cert(use_admin_as_cn_cert: legacy_pulp_cert)
+          config.ssl_client_key = ::Cert::Certs.ssl_client_key(use_admin_as_cn_cert: legacy_pulp_cert)
         else
           fail "Unexpected faraday default_adapter #{Faraday.default_adapter}!  Cannot continue, this is likely a bug."
         end
--- a/app/services/katello/pulp/server.rb
+++ b/app/services/katello/pulp/server.rb
@@ -17,8 +17,8 @@ module Katello
             :debug => true
           },
           :cert_auth => {
-            :ssl_client_cert => ::Cert::Certs.ssl_client_cert,
-            :ssl_client_key => ::Cert::Certs.ssl_client_key
+            :ssl_client_cert => ::Cert::Certs.ssl_client_cert(use_admin_as_cn_pulp_cert: true),
+            :ssl_client_key => ::Cert::Certs.ssl_client_key(use_admin_as_cn_pulp_cert: true)
           }
         }
 


ehelms:
@jlsherrill @ekohl The part of this I was most unsure about was whether it made sense to keep these admin level settings for being able to specify a set of certificates to talk to Pulp with. If the goal is to focus on a single set of client certificates, keeping this seemed contrary to that and could reduce the amount of available settings. I can see the argument for "flexibility" but I'd prefer that argument only if we have a viable use case. Otherwise, let's simplify.

ekohl (changes requested):
If you drop the setting (and I think it makes sense), you also need a migration to drop it from the DB.

jlsherrill:
I think this generally makes sense, as is, and i'm 100% okay with dropping the setting.  I can't imagine why a user would want to change them. 

However, reading this, it would use the same cert (with the foreman fqdn) across all smart proxies.  Does the change here: https://github.com/theforeman/puppet-foreman_proxy_content/pull/350   looks for the foreman FQDN -> admin translation or is it using the smart proxy hostname?

ehelms:
> However, reading this, it would use the same cert (with the foreman fqdn) across all smart proxies. Does the change here: [theforeman/puppet-foreman_proxy_content#350](https://github.com/theforeman/puppet-foreman_proxy_content/pull/350) looks for the foreman FQDN -> admin translation or is it using the smart proxy hostname?

The puppet change now maps all smart-proxy trusted hosts which means each content proxy that is deployed should end up configured with the Foreman FQDN mapped to admin and not the smart-proxy's FQDN.

jlsherrill:
aaahhhh i see now.  Makes perfect sense!

db/migrate/20210507115801_drop_pulp_certificate_settings.rb
```diff
@@ -0,0 +1,6 @@
+class DropPulpCertificateSettings < ActiveRecord::Migration[6.0]
+  def change
+    Setting.where(name: 'pulp_client_key').delete_all
+    Setting.where(name: 'pulp_client_cert').delete_all
```
ekohl, line 4: https://guides.rubyonrails.org/active_record_querying.html#subset-conditions suggests this can be unified to
```suggestion
    Setting.where(name: ['pulp_client_cert', 'pulp_client_cert']).delete_all
```

ekohl (approved):

ekohl:
> I think this would break speaking to pulp2 running on a smart proxy, or speaking to a 4.0/4.1

It wouldn't break on remote smart proxies since we copy the trusted hosts from the Smart Proxy config. There you already need to add the right hosts for Foreman's authentication. However, we don't apply this mapping to 4.0/4.1 so that would break unless we pick the installer changes back. Good point.

Perhaps we should change the implementation to keep the setting around. We can then stop setting it in the installer ([here](https://github.com/theforeman/puppet-katello/blob/55a79bf134f463741e9673f3de9694542cd5d9c6/manifests/application.pp#L21-L31)). That means we use the modern version on fresh installs but keep using the old certificate on existing installations.

The challenge might then be certificates. If we stop deploying the admin certificate (but don't clean it up) we can continue using it. If it needs an update, it'll break. In https://github.com/theforeman/puppet-certs/pull/324 we already spoke about cleanup. Perhaps puppet-katello needs a configuration option to choose the behavior?

ehelms:
My take on reading and thinking is here is our options:

 1) We leave this alone and keep using admin based certificates for 2-3 releases while we let the Apache configuration land in all the needed places and then switch.
 2) We update this code to check what version of smart proxy we are talking to:
    a) if 2.6/3.0+ smart proxies then use the client certificates from Foreman
    b) if less than that version then use the older admin based certificates
    Then in 2-3 releases we drop the admin based certificates

I think 1 is simpler but 2 has the advantage of running against the newer code path and testing it along with setting new installations to be using that method from the get go.

ehelms:
Making a note that the method I am pursuing is instead of based on version, to drive it based on a capability coming from the smart-proxy itself -- https://github.com/theforeman/smart_proxy_pulp/pull/28/

ehelms:
The code for this change is merged and released in all the right places. The change is not yet present fully in nightly, however, this code change since it is based on the capability being present should work regardless. Thus I am marking this ready to review.

app/services/cert/certs.rb
```diff
@@ -12,20 +12,28 @@ def self.candlepin_client_ca_cert
...
-      @ssl_client_cert ||= OpenSSL::X509::Certificate.new(File.read(ssl_client_cert_filename))
+    def self.ssl_client_cert(use_admin_as_cn_cert: false)
+      @ssl_client_cert ||= OpenSSL::X509::Certificate.new(File.read(ssl_client_cert_filename(use_admin_as_cn_cert: use_admin_as_cn_cert)))
     end
 
-    def self.ssl_client_cert_filename
-      Setting['pulp_client_cert']
+    def self.ssl_client_cert_filename(use_admin_as_cn_cert: false)
+      if use_admin_as_cn_cert
+        '/etc/pki/katello/certs/pulp-client.crt'
```
ekohl, line 21: I'm generally not a fan of hardcoding and would prefer that the settings stick around until we've really removed them.

app/services/cert/certs.rb
```diff
@@ -12,20 +12,28 @@ def self.candlepin_client_ca_cert
...
+      else
+        Setting['ssl_certificate']
```
ekohl, line 23: In Foreman itself we always use symbols
```suggestion
        Setting[:ssl_certificate]
```

app/services/cert/certs.rb
```diff
@@ -12,20 +12,28 @@ def self.candlepin_client_ca_cert
...
+      @ssl_client_key ||= OpenSSL::PKey::RSA.new(File.read(ssl_client_key_filename(use_admin_as_cn_cert: use_admin_as_cn_cert)))
     end
 
-    def self.ssl_client_key_filename
-      Setting['pulp_client_key']
+    def self.ssl_client_key_filename(use_admin_as_cn_cert: false)
+      if use_admin_as_cn_cert
+        '/etc/pki/katello/private/pulp-client.key'
+      else
+        Setting['ssl_priv_key']
```
ekohl, line 35:
```suggestion
        Setting[:ssl_priv_key]
```

ekohl (changes requested):

app/lib/katello/resources/registry.rb
```diff
@@ -44,9 +44,9 @@ def load_class
             uri = URI.parse(registry_url)
             self.prefix = uri.path
             self.site = "#{uri.scheme}://#{uri.host}:#{uri.port}"
-            self.ssl_client_cert = ::Cert::Certs.ssl_client_cert
-            self.ssl_client_key = ::Cert::Certs.ssl_client_key
             self.ca_cert_file = ca_cert_file
+            pulp_primary.pulp3_ssl_configuration(self) if pulp_primary
```
ekohl, line 48: If not, should it fall back to the previous config? Should this actually happen at line 36?

app/models/setting/content.rb
```diff
@@ -106,9 +106,9 @@ def self.default_settings
                "/var/lib/pulp/katello-export", N_('Pulp export destination filepath')),
       self.set('pulpcore_export_destination', N_("On-disk location for pulp 3 exported repositories"),
                "/var/lib/pulp/exports", N_('Pulp 3 export destination filepath')),
-      self.set('pulp_client_key', N_("Path for ssl key used for pulp server auth"),
+      self.set('pulp_client_key', N_("Path for ssl key used for pulp server auth (deprecated)"),
```
ekohl, line 109: AFAIK today you still need a DB migration to change descriptions (since descriptions are still stored in the DB, until https://community.theforeman.org/t/setting-in-memory-only-changed-values-in-database/18976 is done)

ehelms: That's a bummer. And I don't suppose we have a built in way to mark settings as deprecated?

app/lib/katello/resources/registry.rb
ehelms: Fair question, I think I would need @ianballou thoughts here. There is a test where `pulp_primary` does not exist, and thus the cause of the `if` check here. That's either by design or something the test is doing just to be 'simple'. Can we have a registry resource without a smart-proxy? I think not, but let's see what is said.

ianballou: I can't see a case either when we would have a registry resource without a smart proxy.  Katello would be very broken without a `pulp_primary` smart proxy.

I have a PR coming that will give all tests a `pulp_primary`.  I'm almost certain the test you're referring to doesn't have a `pulp_primary` simply because there isn't one by default at the moment. Feel free to create one in the test to make it pass.

app/models/setting/content.rb
ekohl: I don't know. Perhaps @ezr-ondrej knows.

test/lib/resources/registry_test.rb
```diff
@@ -17,6 +17,8 @@ def test_pulp3_registry_url
       end
 
       def test_crane_registry_url
+        pulp_primary = FactoryBot.create(:smart_proxy, :default_smart_proxy, :with_pulp3)
+        ::SmartProxy.expects(:pulp_primary).at_least_once.returns(pulp_primary)
```
ehelms, line 21: @jlsherrill @ianballou this test is the one giving me a headache, and then I noticed that it stems from RegistryResource still supporting crane. Would I be better off first opening a PR to rip that chunk of code out? If, as I understand it, RegistryResource is for just talking to the registry on the primary Pulp server, we should never have crane present ever again at this point and this code is dead code.

jlsherrill: yes, +1 from me, we can rip this out

ehelms: Proposing doing clean up over here first -- https://github.com/Katello/katello/pull/9450

jlsherrill:
this needs a rebase now

jlsherrill (approved):
ACK

tested an old proxy and a new (confirming the feature).  all syncs worked and code looks good.  Thanks @ehelms !

PR #9331, opened 2021-04-30
Fixes #32421 - Replace --chunk-size-mb with --chunk-size-gb for hammer content export

lfu:
Fix the issue with hammer export that it fails with super large chunk size.


app/services/katello/pulp3/content_view_version/export.rb
```diff
@@ -91,6 +91,12 @@ def destroy_exporter(exporter_href)
           api.exporter_api.delete(exporter_href)
         end
 
+        def validate_chunk_size(size)
```
parthaa, line 94: can't  you just call this  from 'validate!' method in the same class ?

parthaa: Also add a unit test to test this condition

parthaa: Add the condition test here -> https://github.com/Katello/katello/blob/master/test/services/katello/pulp3/content_view_version/export_test.rb#L50

parthaa (approved):
LGTM Codewise!

I played a little bit with the param. Seems to work fine.

Might wanna add a couple of hammer  tests that checks for this parameter like here -> https://github.com/Katello/hammer-cli-katello/blob/master/test/functional/content_export/complete/version_test.rb#L60. But that's extra credit :)

parthaa:
[test katello]

chris1984 (approved):
ACK works well and parthas comments are addressed.

PR #9333, opened 2021-05-03
Fixes #32498 - 3.17 upgrade fails if there is stale hostgroup data

ianballou:
We need to check for the existence of all records on hostgroups being migrated into content facets.  For reasons unknown but probably due to bugs in older Katello versions, hostgroups could have references to objects (like content views) that no longer exist.  This causes the database migration to fail.  This PR checks for bad data and skips migrating hostgroup data into content facets as needed.

To test:
1) Set up a Katello 3.16 box
2) Create a hostgroup
3) Use the foreman console to set that hostgroup's `content_view_id` to one that doesn't exist
4) Update the Katello gem to 3.17 and run the database migrations (or just run the foreman-installer)
5) See the error from the redmine issue in the production logs
6) Apply this PR and re-migrate
7) Watch for the `Unable to save content facet hostgroup ...` error message
8) Reset the database and try again but with a proper hostgroup this time.  Try only populating one of the checked fields in `all_records_exist?(...)` such as the content source to test that method out.

### DIFF
--- a/db/migrate/20200514092553_move_katello_fields_from_hostgroups.katello.rb
+++ b/db/migrate/20200514092553_move_katello_fields_from_hostgroups.katello.rb
@@ -25,6 +25,16 @@ class MoveKatelloFieldsFromHostgroups < ActiveRecord::Migration[6.0]
     end
   end
 
+  # If an ID is not nil but the associated record does not exist, return false.
+  def all_records_exist?(content_source_id, kickstart_repository_id, content_view_id, lifecycle_environment_id)
+    return false if content_source_id.present? && ::SmartProxy.where(id: content_source_id).empty?
+    return false if kickstart_repository_id.present? && ::Katello::Repository.where(id: kickstart_repository_id).empty?
+    return false if content_view_id.present? && ::Katello::ContentView.where(id: content_view_id).empty?
+    return false if lifecycle_environment_id.present? && ::Katello::KTEnvironment.where(id: lifecycle_environment_id).empty?
+
+    true
+  end
+
   def copy_data_from_hostgroup
     hg_table = ::Hostgroup.arel_table
     hostgroups = ::Hostgroup.unscoped.where(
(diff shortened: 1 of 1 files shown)


chris1984 (approved):
ACK, works great:

Before Patch:

```ruby
[ WARN 2021-05-05T14:19:28 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:migrate]/Exec[foreman-rake-db:migrate]/returns: Caused by:
[ WARN 2021-05-05T14:19:28 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:migrate]/Exec[foreman-rake-db:migrate]/returns: ActiveRecord::InvalidForeignKey: PG::ForeignKeyViolation: ERROR:  insert or update on table "katello_hostgroup_content_facets" violates foreign key constraint "fk_rails_9cae1dc9a3"
[ WARN 2021-05-05T14:19:28 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:migrate]/Exec[foreman-rake-db:migrate]/returns: DETAIL:  Key (content_view_id)=(2) is not present in table "katello_content_views".
[ERROR 2021-05-05T14:20:05 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:migrate]/Exec[foreman-rake-db:migrate]: Failed to call refresh: '/usr/sbin/foreman-rake db:migrate' returned 1 instead of one of [0]
[ERROR 2021-05-05T14:20:05 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:migrate]/Exec[foreman-rake-db:migrate]: '/usr/sbin/foreman-rake db:migrate' returned 1 instead of one of [0]
Something went wrong! Check the log for ERROR-level output
  The full log is at /var/log/foreman-installer/katello.log
[ INFO 2021-05-05T14:21:24 verbose] All hooks in group post finished
[ INFO 2021-05-05T14:21:24 verbose] Installer finished in 209.642522721 seconds
```

With Patch:

```ruby
foreman-rake upgrade:run finished successfully!
  Success!
  * Foreman is running at https://k316.croberts.lan
  * To install an additional Foreman proxy on separate machine continue by running:

      foreman-proxy-certs-generate --foreman-proxy-fqdn "$FOREMAN_PROXY" --certs-tar "/root/$FOREMAN_PROXY-certs.tar"
  * Foreman Proxy is running at https://k316.croberts.lan:9090
  The full log is at /var/log/foreman-installer/katello.log
[ INFO 2021-05-05T14:43:53 verbose] All hooks in group post finished
[ INFO 2021-05-05T14:43:53 verbose] Installer finished in 414.362496317 seconds
```

PR #9334, opened 2021-05-04
Fixes #32137 - Deprecate package groups UI for content hosts

chris1984:
To test:

* Register a client to your Katello
* Check to see if there is a warning on package actions page
* Check for (Deprecated) in the dropdown in package actions

### DIFF
--- a/app/controllers/katello/api/v2/host_packages_controller.rb
+++ b/app/controllers/katello/api/v2/host_packages_controller.rb
@@ -15,7 +15,7 @@ module Katello
 
     def_param_group :packages_or_groups do
       param :packages, Array, :desc => N_("List of package names"), :required => false
-      param :groups, Array, :desc => N_("List of package group names"), :required => false
+      param :groups, Array, :desc => N_("List of package group names (Deprecated)"), :required => false
     end
 
     api :GET, "/hosts/:host_id/packages", N_("List packages installed on the host")
--- a/app/controllers/katello/api/v2/hosts_bulk_actions_controller.rb
+++ b/app/controllers/katello/api/v2/hosts_bulk_actions_controller.rb
@@ -120,7 +120,7 @@ module Katello
     param :content_type, String,
           :desc => N_("The type of content.  The following types are supported: 'package', 'package_group' and 'errata'."),
           :required => true
-    param :content, Array, :desc => N_("List of content (e.g. package names, package group names or errata ids)")
+    param :content, Array, :desc => N_("List of content (e.g. package names, package group names (Deprecated) or errata ids)")
     def install_content
       content_action
     end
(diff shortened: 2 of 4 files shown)


jlsherrill:
This also exists under the bulk actions, and probably needs to be deprecated there too: 
image

Couple of questions:
1) are we only removing this from the UI? (not the api or hammer?)
2) do you have a link to the new mockups?  I'm curious

chris1984:
> This also exists under the bulk actions, and probably needs to be deprecated there too:

Thanks for the review @jlsherrill 

I will update the UI there as well. It makes sense if we are going to remove it, to remove it from the API as well, I will work on getting deprecation messages there. 

As far as the mockups, I have not really been following but I think Maria has some in the "New Host Page" GChat room

chris1984:
@jlsherrill updated the UI and a few other places I missed, will make a hammer PR and add `(Deprecated)` to the package group command

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-views/details/filters/new-filter.controller.js
```diff
@@ -74,7 +74,7 @@ angular.module('Bastion.content-views').controller('NewFilterController',
         if ($scope.stateIncludes('content-view.yum')) {
             $scope.filterChoices = [
                 {id: 'rpm', name: translate('Package')},
-                {id: 'package_group', name: translate('Package Group')},
+                {id: 'package_group', name: translate('Package Group (Deprecated)')},
                 {id: 'erratumId', name: translate('Erratum - by ID')},
                 {id: 'erratumDateType', name: translate('Erratum - Date and Type')},
                 {id: 'modulemd', name: translate('Module Stream')}
```
jlsherrill, line 80: are we actually deprecating content view package group filters?

engines/bastion_katello/app/assets/javascripts/bastion_katello/content-views/versions/views/content-view-version.html
```diff
@@ -25,7 +25,7 @@
     <li ng-class="{active: isState('content-view.version.package-groups')}"
         ng-show="version.package_group_count !== 0">
       <a ui-sref="content-view.version.package-groups({versionId: version.id})">
-        <span translate>rpm Package Groups</span>
+        <span translate>rpm Package Groups (Deprecated)</span>
```
jlsherrill, line 28: I'm not sure we're removing package groups across the board ?  Just the host actions, right?

chris1984:
@jlsherrill updated just the content host parts :)

jlsherrill:
i mentioned it before, but we should probably add deprecation text to the apis for handling package groups, such as:

https://github.com/Katello/katello/blob/master/app/controllers/katello/api/v2/host_packages_controller.rb#L18
and
https://github.com/Katello/katello/blob/master/app/controllers/katello/api/v2/hosts_bulk_actions_controller.rb#L123

i think just modifying the description of those params is fine?

chris1984:
@jlsherrill updated the API descriptions, confirmed with new API docs, none of the tests in hammer will fail.

jlsherrill (approved):

PR #9337, opened 2021-05-04
Fixes #32441 - CV UI Publish Workflow

sjha4:
Needs to be tested with https://github.com/Katello/katello/pull/9352 to test promotion workflow.

To test:
rm -rf node_modules package-lock.json  && npm i from inside foreman directory.
You might need to run through this: https://gist.github.com/johnpmitsch/95b9d631935115ca91c46efecb115464


sjha4:
Need to confirm what to do with active tasks on versions page.
Publish Active History:
https://codebeautify.org/jsonviewer/cb762268
Promote Active History: 
https://codebeautify.org/jsonviewer/cb907f28

webpack/scenes/ContentViews/Publish/CVPublishFinish.js
```diff
@@ -0,0 +1,149 @@
...
+    dispatch(publishContentView({
+      id: cvId,
+      versionCount,
+      description,
+      environment_ids: checkedItems,
+      is_force_promote: (forcePromote.length > 0),
+    }));
+    setDescription('');
+    setCheckedItems([]);
+  }, []);
```
jeremylenz, line 52: I don't like `[]` for the dependency array, as it breaks the rules.  Options are
1. Add `setSaving, dispatch, cvId, versionCount, description, checkedItems, forcePromote, setDescription, setCheckedItems` to the dependency array;
2. Create a ref outside of the effect: `const done = useRef(false);` Then put `if (done.current) return;` at the top of the useEffect.  And set `done.current` to true after `setCheckedItems([]);`. Then you can just not pass any dependency array at all.

webpack/scenes/ContentViews/Publish/CVPublishFinish.js
```diff
@@ -0,0 +1,149 @@
...
+    setSaving(true);
+    const { id } = response;
+    if (id && status === STATUS.RESOLVED) {
+      setSaving(false);
+      dispatch(pollPublishTask(cvVersionPublishKey(cvId, versionCount), response));
+      setPolling(true);
+    } else if (status === STATUS.ERROR) {
+      setSaving(false);
+    }
+  }, [JSON.stringify(response), status, error]);
```
jeremylenz, line 64: I still hate `JSON.stringify`.. Have you thought any more about adding that `useDeepCompareEffect`?

sjha4: I tried out the deepcompare locally..Seems to work except I was hitting some snags on the filter page for some reason. 
I am adding a card to the board to take care of that on all pages. https://projects.theforeman.org/issues/32621

webpack/scenes/ContentViews/Publish/CVPublishFinish.js
sjha4: I thought this was the hook way of running the effect once on page's first render..That is what we need to do here in the final step of the wizard. Call publish and display progress..

jeremylenz: It's true that `[]` will cause the effect only to run once, but that technique really only works cleanly if you have no dependencies.  Once we start using the hooks eslint rules `[]` is not going to fly; we'll have to do something else.

sjha4: I see in foreman they use dispatch etc in the dependency array..Let me update that and verify that it still works..

ianballou (commented):
There seems to be weird behavior with the "Force Promotion" error message. It seems to only refer to the first lifecycle environment that you select. Examples with Dev, Test, Prod:
1) I select all 3 LCEs, the unselect Test.  There is no "Force Promotion" warning.
2) I select Prod, see the warning.  Then I select Test, then Dev, and the error doesn't go away

There is also an issue with the number of LCEs displayed in the warning. Examples:
1) Select Test and then Prod. Only Test will be in the warning
2) Select Prod and then Test. Both Test and Prod will be in the warning.

sjha4:
> There seems to be weird behavior with the "Force Promotion" error message. It seems to only refer to the first lifecycle environment that you select. Examples with Dev, Test, Prod:
> 
> 1. I select all 3 LCEs, the unselect Test.  There is no "Force Promotion" warning.
> 2. I select Prod, see the warning.  Then I select Test, then Dev, and the error doesn't go away
> 
> There is also an issue with the number of LCEs displayed in the warning. Examples:
> 
> 1. Select Test and then Prod. Only Test will be in the warning
> 2. Select Prod and then Test. Both Test and Prod will be in the warning.

The current logic to set forced promote array is not handling the above use-cases..Working on it..

sjha4:
@ianballou : Too many use-cases around selecting and unselecting.. 😄 
I think the last commit should fix the issues around that. Let me know if you see anything off.

ianballou:
The lifecycle environment selection is looking a lot better to me!

sjha4:
[test katello]

ianballou:
I don't fully understand all of the JS yet but I think I'm pretty much good to approve this from a functional standpoint. I noticed there is a window sizing issue. Maybe it's outside the scope of this PR?  I've attached an image. There is a point where the window is too small and you cannot scroll.
Screenshot from 2021-06-02 16-15-55

sjha4:
> I don't fully understand all of the JS yet but I think I'm pretty much good to approve this from a functional standpoint. I noticed there is a window sizing issue. Maybe it's outside the scope of this PR? I've attached an image. There is a point where the window is too small and you cannot scroll.
> Screenshot from 2021-06-02 16-15-55

I tried this out. Couple of issues here zIndex of navigation pane is set very high(1030 for some reason) and on purpose so everything else is behind it..And the wizard is appending to the whole body and sliding behind the pane when it has smaller real estate..I will try to run it through @jeremylenz if he has any ideas around this..

sjha4:
[test katello]

webpack/scenes/ContentViews/Publish/CVPublishFinish.js
```diff
@@ -0,0 +1,147 @@
...
+    return <Loading />;
+  }
+
+  if (polling && pollResponse) {
+    return (
+      <Stack hasGutter>
+        <StackItem span={12} rowSpan={3}>
+          <Bullseye>
+            <div style={{ height: '350px', width: '350px' }}>
+              <ChartDonut
```
jeremylenz, line 91: I feel like this is missing a visual indicator that something is happening.  It's too static.  I wonder if Patternfly has anything that can help..  how would it look if this were superimposed on a [medium Circle skeleton](https://patternfly-react.surge.sh/components/skeleton)?

Also it's too bad that there's no animation transition when the percentages update. But that's probably not worth spending time on.

webpack/scenes/ContentViews/Publish/CVPublishFinish.js
```diff
@@ -0,0 +1,147 @@
...
+    const { state, result } = pollResponse;
+    if (response && (state === 'paused' || result === 'error')) {
+      setTaskErrored(true);
+    }
+    if (response && (state === 'stopped' && result === 'success')) {
+      setPolling(false);
+      dispatch(handleFinishedTask(cvId, cvVersionPublishKey(cvId, versionCount), pollResponse));
+      dispatch(getContentViewDetails(cvId));
+      dispatch(getContentViews);
+      setIsOpen(false);
```
jeremylenz, line 75: When the task finishes, the modal closes completely and we don't get to see the completed donut.  Thoughts on keeping it onscreen for a half second or so?

webpack/scenes/ContentViews/Publish/CVPublishFinish.js
```diff
@@ -0,0 +1,147 @@
...
+                title="Publishing"
+                colorScale={taskErrored ? ['#FF0000', '#F08080'] : ['#008000', '#8FBC8F']}
+              />
+            </div>
+          </Bullseye>
+        </StackItem>
+        <StackItem isFilled> </StackItem>
+        <StackItem span={6}>
+          <Bullseye>
+            <Button onClick={() => { dispatch(handleFinishedTask(cvId, cvVersionPublishKey(cvId, versionCount), pollResponse, true)); setIsOpen(false); }} variant="primary" aria-label="publish_content_view">
```
jeremylenz, line 105: These buttons kinda look like they're in the center of the screen - there's too much space between them and the bottom of the modal.  Should they be moved down a bit?

webpack/scenes/ContentViews/Publish/PublishContentViewWizard.js
```diff
@@ -0,0 +1,132 @@
...
+            forcedPromoteList = [...forcedPromoteList, env];
+          } else if (!(checkedItems.filter(item => item.id === env.prior.id).length) &&
+            checkedItems.includes(env)) {
+            // Find and add first out of order environment
+            forcedPromoteList = [...forcedPromoteList, env];
+          }
+        });
+      });
+      setForcePromote(forcedPromoteList.filter(item => (checkedItems.includes(item))));
+    }
```
jeremylenz, line 95: This works fine but I still think it's hard to read and grok what's going on.  Are there ways you can refactor it to make it easier to comprehend?  (I actually like the recursion idea and think it makes it simpler in some ways, for example you don't have to have a special case for the first two paths.)

webpack/scenes/ContentViews/Publish/PublishContentViewWizard.js
```diff
@@ -0,0 +1,132 @@
...
+  }, [checkedItems, setCheckedItems]);
```
jeremylenz, line 96: All the other dependencies used should still be listed here (and I don't think you actually need `setCheckedItems` since it's not used)

webpack/scenes/ContentViews/components/EnvironmentPaths/EnvironmentPaths.scss
```diff
@@ -0,0 +1,8 @@
+.env-labels-with-pointer {
+  &:not(:last-child):after {
+    content: '>';
```
jeremylenz, line 3: Nice!  @sjha4 CSS master ;)

jeremylenz (commented):
Sorry for the delays, here finally are my comments :)

webpack/scenes/ContentViews/Publish/CVPublishForm.js
```diff
@@ -0,0 +1,103 @@
...
+              type="text"
+              id="description"
+              aria-label="input_description"
+              name="description"
+              value={description}
+              onChange={value => setDescription(value)}
+            />
+          </FormGroup>
+          <FormGroup label="Promote" fieldId="promote" style={{ marginBottom: '2em' }}>
+            <Switch
```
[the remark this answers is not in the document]
sjha4: The CSS you showed me earlier is in overrides.css and already applied. That didn't seem to remove the outline on this.
This however worked for me looking at the switch component:
```
.pf-c-switch {
  --pf-c-switch__input--focus__toggle--OutlineWidth: 0;
}

```

webpack/scenes/ContentViews/Publish/CVPublishFinish.js
sjha4: animate={{ duration: 500 }} This adds a bit of animation.

webpack/scenes/ContentViews/Publish/CVPublishFinish.js
sjha4: Updated as discussed to stick around for 500 milliseconds..There were extra dispatches in an action I showed you earlier. Removed that and the state update happens without warnings I was seeing earlier with setTimeout.

webpack/scenes/ContentViews/Publish/CVPublishFinish.js
sjha4: Increased size of status donut chart and buttons are down below.
Screenshot from 2021-06-14 22-37-00

jeremylenz: this looks much better! :+1:

webpack/scenes/ContentViews/Publish/PublishContentViewWizard.js
```diff
@@ -0,0 +1,134 @@
...
+    return checkedItems.includes(env) && isValid(getEnv(env.prior.id));
+  };
+
+  useEffect(() => {
+    const forcedList = [];
+    checkedItems.forEach((item) => {
+      if (!isValid(item)) forcedList.push(item);
+    });
+    setForcePromote(forcedList);
+  }, [checkedItems, setForcePromote]);
```
sjha4, line 98:
```
useEffect(() => {
    setForcePromote(checkedItems.map((item) => {
      if (!isValid(item)) return item;
    }).filter(item => item));
  }, [checkedItems, setForcePromote]);
```

This works too. Let me know which is better..

jeremylenz: I like a combination of the two:
1. Use the `forcedList` variable name to explain what we're doing;
2. Use `map` to consolidate and avoid using `forEach` + `push`.  _Update:_ I think it can be simplified even more by just using `filter`:
```suggestion
  useEffect(() => { 
   const forcedList = checkedItems.filter(item => !isValid(item)); 
   
   setForcePromote(forcedList);
   }, [checkedItems, setForcePromote]);
```

jeremylenz (commented):
I think the new method is a bit easier to understand.. Some comments below to streamline it even more.  let me know your thoughts :)

webpack/scenes/ContentViews/Publish/PublishContentViewWizard.js
```diff
@@ -0,0 +1,132 @@
...
+          }
+        });
+      });
+      setForcePromote(forcedPromoteList.filter(item => (checkedItems.includes(item))));
+    }
+  }, [checkedItems, setCheckedItems]);
+
+  return (
+    <Wizard
+      title={__('Publish')}
```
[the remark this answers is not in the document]
sjha4: Opened https://projects.theforeman.org/issues/32817 for Wizard and modal header colors.

sjha4:
Screenshot from 2021-06-15 15-59-18

Reused same behavior for success and error. Wizard closes with an error notification after displaying the red chart for 0.5 seconds.

jeremylenz (changes requested):
A couple problems:

* I saw "0000000000" momentarily flash on the screen where the percentage should be (this is while it was at 0%) but not sure how to recreate it
* After publishing a version, the modal does not reset.  I click the button and it immediately starts publishing again.

jeremylenz:
Btw, The animation looks great :smile: :+1:

webpack/scenes/ContentViews/Publish/PublishContentViewWizard.js
```diff
@@ -0,0 +1,130 @@
...
+  selectEnvironmentPaths,
+  selectEnvironmentPathsStatus,
+} from '../components/EnvironmentPaths/EnvironmentPathSelectors';
+
+const PublishContentViewWizard = ({
+  details, show, setIsOpen, currentStep, setCurrentStep,
+}) => {
+  const { name, id: cvId, version_count: versionCount } = details;
+  const [description, setDescription] = useState('');
+  // Change to userCheckedItems
```
jeremylenz, line 21: comment no longer necessary :)

webpack/scenes/ContentViews/Publish/PublishContentViewWizard.js
```diff
@@ -0,0 +1,130 @@
...
+      isFinishedStep: true,
+    },
+  ];
+
+  useEffect(
+    () => {
+      dispatch(getEnvironmentPaths());
+      setCurrentStep(1);
+    },
+    [],
```
jeremylenz, line 73: dependency array needs to be filled out

webpack/scenes/ContentViews/Publish/PublishContentViewWizard.js
```diff
@@ -0,0 +1,130 @@
...
+  );
+
+  useEffect(() => {
+    if (!environmentPathLoading) {
+      const { results } = environmentPathResponse || {};
+      setEnvPathFlat(results.map(result => result.environments).flatten());
+    }
+  }, [environmentPathResponse, environmentPathLoading]);
```
jeremylenz, line 81:
```suggestion
  }, [environmentPathResponse, environmentPathLoading, setEnvPathFlat]);
```

webpack/scenes/ContentViews/Publish/PublishContentViewWizard.js
```diff
@@ -0,0 +1,130 @@
...
+      dispatch(getEnvironmentPaths());
+      setCurrentStep(1);
+    },
+    [],
+  );
+
+  useEffect(() => {
+    if (!environmentPathLoading) {
+      const { results } = environmentPathResponse || {};
+      setEnvPathFlat(results.map(result => result.environments).flatten());
```
jeremylenz, line 79: Why keep `envPathFlat` in state?  I feel like it would work fine as a regular function..

jeremylenz (commented):
a few annoying comments below, sorry

webpack/scenes/ContentViews/Publish/PublishContentViewWizard.js
sjha4: It felt like an operation needed just once..

webpack/scenes/ContentViews/Publish/PublishContentViewWizard.js
sjha4: This is intentional. This is supposed to run just once on component load.

sjha4:
> A couple problems:
> 
> * I saw "0000000000" momentarily flash on the screen where the percentage should be (this is while it was at 0%) but not sure how to recreate it

Let's open an issue if you see this again. This might be from the animation? 

> * After publishing a version, the modal does not reset.  I click the button and it immediately starts publishing again.

This should be fixed now. I introduced this when adding the 0.5 sec wait and forgetting to reset currentStep back to 0.

Addressed other comments.

webpack/scenes/ContentViews/Publish/PublishContentViewWizard.js
jeremylenz: eslint isn't going to care about that once we turn on that rule ;)

jeremylenz: https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies

jeremylenz: Oh, and if you want to read the whole novel about it: 
https://github.com/facebook/react/issues/14920

webpack/scenes/ContentViews/Publish/PublishContentViewWizard.js
jeremylenz: Sounds like a job for [`useMemo`](https://reactjs.org/docs/hooks-reference.html#usememo), not `useState`!

webpack/scenes/ContentViews/Publish/PublishContentViewWizard.js
sjha4: They do discuss using props and functions with side effects on props in dependency array. Added the functions to dependency array..

webpack/scenes/ContentViews/Publish/PublishContentViewWizard.js
sjha4: Updated to use useMemo..My first use of it :)

webpack/scenes/ContentViews/Details/ContentViewDetailActions.js
```diff
@@ -139,4 +141,18 @@ export const getContentViewVersions = (cvId, params) => {
   });
 };
 
+export const pollPublishTask = (cvPublishVersionKey, task) => async (dispatch) => {
+  dispatch(startPollingTask(cvPublishVersionKey, task));
+};
+
+export const handleFinishedTask = (cvId, cvPublishVersionKey, task, pending = false) =>
```
jeremylenz, line 148: This has enough args to make it confusing.  Probably better to use an object and destructure.
```suggestion
export const handleFinishedTask = ({cvId, cvPublishVersionKey, task, pending: false}) =>
```

webpack/scenes/ContentViews/Publish/CVPublishFinish.js
jeremylenz:
```js
() => { dispatch(handleFinishedTask(cvId, cvVersionPublishKey(cvId, versionCount), pollResponse, true)); setIsOpen(false);
```
This is multiple lines of code; probably better to extract into a named function at this point.

webpack/scenes/ContentViews/Publish/CVPublishReview.js
```diff
@@ -0,0 +1,81 @@
...
+  const environmentPathResponse = useSelector(selectEnvironmentPaths);
+  const { results } = environmentPathResponse || {};
+  const environmentsAdded = (environmentResult) => {
+    let environmentsList = [];
+    let library = null;
+    environmentResult.forEach((path) => {
+      const {
+        environments,
+      } = path || {};
+      [library] = environments;
```
jeremylenz, line 24: :exploding_head: nice array destructure :smile:

webpack/scenes/ContentViews/Publish/CVPublishReview.js
```diff
@@ -0,0 +1,81 @@
...
+      const envInPath = environments.filter(env =>
+        userCheckedItems.filter(item => item.id === env.id).includes(env));
```
jeremylenz, line 26: filter-inception!  Can this be clarified with a comment or a named function?

webpack/scenes/ContentViews/Publish/CVPublishReview.js
```diff
@@ -0,0 +1,81 @@
...
+      { title: `Version ${nextVersion}` },
+      { title: <ComponentEnvironments environments={environmentsAdded(results)} /> },
+    ],
+  ];
+
+  return (
+    <Table
+      aria-label="Review Table"
+      cells={columnHeaders}
+      rows={rows}
```
jeremylenz, line 61: This is a pretty simple Table; would you consider using `TableComposable` instead?  It's going to allow you to decouple your JSX from the row and column data and make everything look so much cleaner!

webpack/scenes/ContentViews/Publish/__tests__/publishContentView.test.js
```diff
@@ -0,0 +1,127 @@
...
+
+const cvDetailData = require('../../Details/__tests__/contentViewDetails.fixtures.json');
+
+const cvPublishPath = api.getApiUrl('/content_views/1/publish');
+const publishResponseData = require('./publishResponse.fixture.json');
+
+const environmentPathsPath = api.getApiUrl('/organizations/1/environments/paths');
+const environmentPathsData = require('./environmentPaths.fixtures.json');
+
+test('Can call API and show Wizard', async (done) => {
```
jeremylenz, line 17: Would like to see tests added

* `Shows a warning when environments are selected out of order`
* `Doesn't show a warning when environments selected are the next in the lifecycle path`

webpack/scenes/ContentViews/Table/ContentViewsTable.js
```diff
@@ -9,30 +9,73 @@ import tableDataGenerator from './tableDataGenerator';
...
+      id: rowInfo.cvId.toString(),
+      name: rowInfo.cvName,
+      composite: rowInfo.cvComposite,
+      version_count: rowInfo.cvVersionCount,
+      next_version: rowInfo.cvNextVersion,
+    });
+    setIsPublishModalOpen(true);
+  };
+
+  // Prevents flash of "No Content" before rows are loaded
```
jeremylenz, line 49: Does it, though? :thinking:

webpack/scenes/ContentViews/Table/ContentViewsTable.js
```diff
@@ -144,24 +177,19 @@ const ContentViewTable = ({ response, status, error }) => {
       <React.Fragment>
         <CopyContentViewModal cvId={actionableCvId} cvName={actionableCvName} show={copy} setIsOpen={setCopy} aria-label="copy_content_view_modal" />
       </React.Fragment>
+      {isPublishModalOpen &&
```
jeremylenz, line 180: Using conditional rendering for modals prevents the modal opening / closing animations provided by Patternfly. Instead you should use the `isOpen` prop of the Patternfly `Modal` component.

jeremylenz (commented):
Almost there! Just some nitpicks below, looks great otherwise :+1:

webpack/scenes/ContentViews/Table/ContentViewsTable.js
sjha4: Ya, but this is only on the main index page. Other tables still have this issue.

webpack/scenes/ContentViews/Publish/__tests__/publishContentView.test.js
sjha4: Added a test for in order and out of order env. 👍🏽

webpack/scenes/ContentViews/Publish/CVPublishReview.js
sjha4: Updated to use TableComposable 👍🏽

webpack/scenes/ContentViews/Publish/CVPublishReview.js
sjha4: Simplified this whole thing a little and used useMemo this time. This adds library to the userCheckedItems to display in the review table. 

```
const promotedToEnvironments = useMemo(() => {
    if (!environmentPathLoading) {
      const { results } = environmentPathResponse || {};
      // 1st env in result's every path is Library
      return [results[0].environments[0]].concat(userCheckedItems);
    }
    return [];
  }, [environmentPathResponse, environmentPathLoading]);
```

webpack/scenes/ContentViews/Details/ContentViewDetailActions.js
sjha4: Updated to use an object instead of separate params.

webpack/scenes/ContentViews/Table/ContentViewsTable.js
sjha4: I tried this and this causes all sorts of issues which the tests also don't like..The component actually renders but remains closed, makes all the dispatch calls and everything. Will have to rely on conditional rendering to avoid that.

webpack/scenes/ContentViews/Publish/CVPublishFinish.js
```diff
@@ -0,0 +1,167 @@
...
+        <GridItem span={12} rowSpan={11}>
+          <Bullseye>
+            <div style={{ height: '500px', width: '500px' }}>
+              <ChartDonut
+                ariaDesc={__('Publishing Content View')}
+                ariaTitle={__('Publishing Content View')}
+                animate={{ duration: 100 }}
+                data={[{ x: 'Completed', y: pollResponse.progress ? pollResponse.progress * 100 : 0 }, { x: 'Pending', y: pollResponse.progress ? (100 - (pollResponse.progress * 100)) : 100 }]}
+                labels={({ datum }) => `${datum.x}: ${datum.y}%`}
+                title={pollResponse.progress ? `${(pollResponse.progress * 100).toFixed(0)} %` : 'Starting..'}
```
sjha4, line 108: I was able to reproduce the 00000 sort of issue. For some reason sometimes the API returns a very long decimal value in the task progress result instead of the usual decimal with precision 2 or 3. That was causing this title to be a very long number. .toFixed(0) should fix that since we don't really need decimal precision in % complete.

jeremylenz: Oooooooh!  Nice :)

sjha4:
[test katello]

sjha4:
[test katello]

sjha4:
[test katello]

sjha4:
[test katello]

webpack/scenes/ContentViews/components/EnvironmentPaths/EnvironmentPaths.js
```diff
@@ -0,0 +1,79 @@
...
+
+  const oncheckedChange = (checked, env) => {
+    if (checked) {
+      setUserCheckedItems([...userCheckedItems, env]);
+    } else {
+      setUserCheckedItems(userCheckedItems.filter(item => item.id !== env.id));
+    }
+  };
+
+  const buildformGroups = (results) => {
```
jeremylenz, line 25: I don't think this should be a function.  This is part of render.

1. Change the `results.forEach` to a `results.map`
3. Don't reassign with `concat` each time - Have each item in the `environments.map` return a React.Fragment with the `<FormGroup>` and the `<hr>`
4. Move it all inside the `<Form>...</Form>` on line 66.

jeremylenz (commented):
Turns out I had a few more comments :grimacing:

webpack/scenes/ContentViews/Publish/CVPublishFinish.js
```diff
@@ -0,0 +1,185 @@
...
+  const pollResponse = useSelector(state =>
+    selectPublishTaskPoll(state, cvVersionPublishKey(cvId, versionCount)), shallowEqual);
+  const pollResponseStatus = useSelector(state =>
+    selectPublishTaskPollStatus(state, cvVersionPublishKey(cvId, versionCount)), shallowEqual);
+
+  const pollPublishTask = (cvPublishVersionKey, task) => {
+    if (!polling) dispatch(startPollingTask(cvPublishVersionKey, task));
+  };
+
+  const progressCompleted = () => (pollResponse.progress ? pollResponse.progress * 100 : 0);
```
jeremylenz, line 45: May as well move the `.toFixed` here

jeremylenz (commented):
looking good! Giving it one more whirl..

jeremylenz (approved):
The last review was just small nitpicks, so I'll go ahead and APJ

Great work! :smile: :+1:

webpack/scenes/ContentViews/Publish/CVPublishFinish.js
sjha4: I did refactor this a little:
progressPending method is nothing but 100 - progressCompleted() so using that..

toFixed() returns a string so can't used it in the chart as data directly except the title and label.

PR #9339, opened 2021-05-05
Fixes #32225 - store akey original environment on the right scope

jturel:
Check the BZ for the basic repro steps. For a slightly more clever test:

- change the env + cv of the activation key
- change to a different tab
- change back to Details tab
- everything should look good (cv+env are accurate, no info messages shown for the CV field)

### DIFF
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/activation-keys/details/activation-key-details.controller.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/activation-keys/details/activation-key-details.controller.js
@@ -51,9 +51,13 @@ angular.module('Bastion.activation-keys').controller('ActivationKeyDetailsContro
             ];
         };
 
+        $scope.setOriginalEnvironment = function(newEnvironment) {
+            $scope.originalEnvironment = newEnvironment;
+        };
+
         $scope.activationKey = ActivationKey.get({id: $scope.$stateParams.activationKeyId}, function (activationKey) {
-            $scope.$broadcast('activationKey.loaded', activationKey);
             $scope.panel.loading = false;
+            $scope.setOriginalEnvironment(activationKey.environment);
         }, function (response) {
             $scope.panel.loading = false;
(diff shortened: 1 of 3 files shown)


jturel:
[test katello]

engines/bastion_katello/app/assets/javascripts/bastion_katello/activation-keys/details/activation-key-details.controller.js
```diff
@@ -51,9 +51,13 @@ angular.module('Bastion.activation-keys').controller('ActivationKeyDetailsContro
             ];
         };
 
+        $scope.setOriginalEnvironment = function(newEnvironment) {
+            $scope.originalEnvironment = newEnvironment;
+        };
+
         $scope.activationKey = ActivationKey.get({id: $scope.$stateParams.activationKeyId}, function (activationKey) {
-            $scope.$broadcast('activationKey.loaded', activationKey);
```
chris1984, line 55: What did the `$scope.$broadcast do`?

chris1984 (commented):
Starting to test. I would say if we were not changing activation keys down the road, we should prob add a test, but I am fine without one.

engines/bastion_katello/app/assets/javascripts/bastion_katello/activation-keys/details/activation-key-details.controller.js
jturel: It was responsible for sending an event that another scope could act upon. See above controller => `$scope.$on('activationKey.loaded', function () {` no longer needed!

jturel:
> Starting to test. I would say if we were not changing activation keys down the road, we should prob add a test, but I am fine without one.

no harm in adding some tests. was easier than I thought here- done!

chris1984 (approved):
ACK, code looks good and it works:

Before patch:

https://nimb.ws/XYNpfC

After patch:

https://nimb.ws/RvbJxi

Thanks for explanining the function question I had.

chris1984:
[test katello]

PR #9342, opened 2021-05-05
fixes #32515 - Parse Rocky Linux distribution properly

lumarel:
This PR will add the correct parsing of [Rocky Linux](https://rockylinux.org) to Katello.
Here also the reference about the [discussion](https://community.theforeman.org/t/client-recognition-for-rocky-linux-in-foreman-katello/23437).

If I can improve something I'm really open for recommendations, as I'm not really familiar with the code of Foreman/Katello.
Thank you!


jlsherrill:
[test katello]

jlsherrill:
This now needs a rebase after merging support for AlmaLinux :)

Looking at facter commit: https://github.com/puppetlabs/facter/commit/f4ca4f88d8e878071806d1895570848918fdc49a  it looks like this is just being called 'Rocky'  in puppet land?  Can you also confirm that `name =~ /rocky\s*linux/ `  matches what subscription-manager sends?

lumarel:
Thank you for reviewing @jlsherrill !
Okay as this is the first time for me to make a rebase, how do I do this without breaking the PR? ^^

And I was not sure which variable gets used by the katello, but if it is the `distribution.name` of the subscription manager it should be correct:
```bash
[root@r8-rc-dev ~]# subscription-manager facts
...
distribution.id: 8.3
distribution.name: Rocky Linux
distribution.version: 8.3
distribution.version.modifier: Unknown
...
```

But you are right about puppet, it is parsed differently there:
```bash
[root@r8-rc-dev ~]# puppet facts
{
...
  "values": {
...
    "operatingsystem": "Rocky",
    "operatingsystemmajrelease": "8",
    "operatingsystemrelease": "8.3",
    "os": {
      "architecture": "x86_64",
      "family": "RedHat",
      "hardware": "x86_64",
      "name": "Rocky",
      "release": {
        "full": "8.3",
        "major": "8",
        "minor": "3"
      },
...
    },
    "osfamily": "RedHat",
...
    "puppetversion": "6.22.1",
...
  },
...
}
```

jlsherrill:
actually despite the ugly history, it looks like i can squash this and it should be fine.  Lets see what the tests show :)

lumarel:
Yeah.. sorry this already happened to me once, wasn't looking too closely at the merge message

But okay thank you if this is possible :)

jlsherrill:
[test katello]

jlsherrill:
Oh wait, i was about to merge and realized that the Puppet 'name' still looks incorrect.  It should be 'Rocky'.  If you want to use this Pr or open a 2nd, either is fine with me

lumarel:
> 
> 
> Oh wait, i was about to merge and realized that the Puppet 'name' still looks incorrect. It should be 'Rocky'. If you want to use this Pr or open a 2nd, either is fine with me

Oh okay, so it has to match the puppet facts, this was what I wasn't sure about :)
I will recreate the PR for this, so also all the mess gets cleaned up 👍🏻

PR #9343, opened 2021-05-05
Fixes #32527 - Transient package groups test

jturel:
Saw this fail due to the sort order in a recent jenkins build


chris1984 (approved):
ACK

jeremylenz:
Maybe we should add an `assert_same_elements`.. seems like we'd use it quite a bit :thinking:

jturel:
[test katello]

jturel:
> Maybe we should add an `assert_same_elements`.. seems like we'd use it quite a bit thinking

+1 but I will deem it out of scope for this PR. Be the change and all that ...

PR #9295, opened 2021-04-16
Fixes #32334 - Subscriptions and Pools can cross organizations

jturel:

PR #9347, opened 2021-05-06
Refs #32334 - ensure import_all creates Pools with a subscription_id

jturel:
My recent PR https://github.com/Katello/katello/pull/9295 added non-null constraints+validations for subscription_id and organization_id on `Katello::Pool`. I missed updating one area of the code => `import_all` which would silently fail and not create any pools because it calls `first_or_create` without `!` (due to the missing subscription_id). This PR adjusts the `CandlepinObject` module to add a new method `import_candlepin_record` which the Pool model now overrides in order to ensure the subscription_id is provided.

If you run into issues I may need your development.log output. Since it logs tons of HTTP requests I ask that you please change to INFO logging in ~/foreman/config/logging.yaml under 'development' to suppress those. Restart the rails server after changing that!

Two ways I recommend to test this. The first is the Rails console:

```
Katello::Subscription.destroy_all
Katello::Pool.import_all
Katello::Pool.count => 0
```
The result will be zero here because of the issue mentioned above. Notice that there was no error raised. Assuming you have a manifest imported or a custom product created that number should be > 1. Repeating the steps with this PR checked out should show the correct # of Pool objects.

To test the actual workflow: delete any imported manifest and then import it again. Notice that the Subscriptions list may be missing some or all of the expected Subscriptions. That's because import_all didn't create them, and the CandlepinEventListener may have some luck in doing so. The results should be deterministic with this PR checked out! Repeat the test a few times or various permutations (SCA toggle, add or remove subscriptions, anything that re-imports the manifest) to make sure the fix sticks.

### DIFF
--- a/app/models/katello/glue/candlepin/candlepin_object.rb
+++ b/app/models/katello/glue/candlepin/candlepin_object.rb
@@ -3,18 +3,35 @@ module Katello
     extend ActiveSupport::Concern
 
     module ClassMethods
-      def get_candlepin_ids(organization)
-        self.get_for_owner(organization.label).map { |subscription| subscription["id"] }
+      def candlepin_records_by_id(organization)
+        records = get_for_owner(organization.label)
+        records_by_id = {}
+        records.each do |record|
+          records_by_id[record['id']] = record
+        end
+        records_by_id
       end
 
-      def import_candlepin_ids(organization)
-        candlepin_ids = self.get_candlepin_ids(organization)
-        candlepin_ids.each do |cp_id|
-          Katello::Util::Support.active_record_retry do
-            self.where(:cp_id => cp_id, :organization_id => organization.id).first_or_create unless cp_id.nil?
-          end
+      def import_candlepin_records(candlepin_objects, org)
+        candlepin_objects.each do |object|
+          import_candlepin_record(record: object, organization: org)
         end
-        candlepin_ids
+      end
+
+      def import_candlepin_record(record:, organization:)
+        db_attrs = {
+          cp_id: record['id'],
+          organization: organization
+        }
+
+        yield(db_attrs) if block_given?
+
+        persisted = nil
+        Katello::Util::Support.active_record_retry do
+          persisted = self.where(db_attrs).first_or_create!
+        end
+
+        persisted
       end
 
       def with_identifier(ids)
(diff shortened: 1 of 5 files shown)


app/models/katello/glue/candlepin/candlepin_object.rb
```diff
@@ -3,18 +3,35 @@ module Glue::Candlepin::CandlepinObject
...
-          Katello::Util::Support.active_record_retry do
-            self.where(:cp_id => cp_id, :organization_id => organization.id).first_or_create unless cp_id.nil?
-          end
+      def import_candlepin_records(candlepin_objects, org)
+        candlepin_objects.each do |object|
+          import_candlepin_record(record: object, organization: org)
+        end
+      end
+
+      def import_candlepin_record(record:, organization:)
```
jturel, line 21: Katello::Pool overrides this and calls `super` to inject the `subscription_id`

app/models/katello/pool.rb
```diff
@@ -24,8 +24,8 @@ class Pool < Katello::Model
     scope :upstream, -> { where.not(upstream_pool_id: nil) }
     scope :redhat, -> { joins(:products).merge(Katello::Product.redhat).distinct }
 
-    include Glue::Candlepin::Pool
     include Glue::Candlepin::CandlepinObject
+    include Glue::Candlepin::Pool
```
jturel, line 28: The order changed here so that the override of `import_candlepin_record` is picked up

test/models/pool_test.rb
```diff
@@ -107,16 +107,53 @@ def test_candlepin_data_rescue_gone
...
+          }
+        }
+      )
       Pool.expects(:in_organization).with(org).returns([@pool_one])
       @pool_one.expects(:import_data).once
       @pool_one.expects(:import_managed_associations).once
       Pool.import_all(org)
     end
 
+    def test_import_all_new_record
```
jturel, line 124: This test would fail on master branch, so added coverage for it here

test/models/subscription_test.rb
```diff
@@ -7,6 +7,32 @@ def setup
       @other = katello_subscriptions(:other_subscription)
     end
 
+    def test_import_all
```
jturel, line 10: This path is not broken on master, but Subscription also uses import_all so I decided to be safe and add test coverage now

jjeffers (approved):
Tested this PR on a katello installation where previously a new manifest would throw an error on import re: subscription_id values being null.

After applying the PR, the manifest and SCA works without errors or warnings - the expected repositories appear.

jturel:
[test katello]

jturel:
[test katello]

PR #9349, opened 2021-05-07
refs #31809 - Registration - fix incorrect LCE variable

stejskalleos:
Due to the incorrect variable [registration form component](https://github.com/Katello/katello/blob/master/webpack/components/extensions/RegistrationCommands/fields/LifecycleEnvironment.js#L14) is not loading life cycle environments.


tbrisker (approved):
Thanks @stejskalleos !

PR #9357, opened 2021-05-12
Fixes #32591 - use correct ping mechanism for katello-agent events

jturel:
While the katello-based agent infra was in development the Event Daemon ping code was rewritten - removing the need for caching the individual service status since the Daemon caches them now. This PR addresses that and also adds a small sleep after each service is started to give its thread some time to start.

Testing: The ping api should return 'ok' for katello_agent at app startup


ianballou (approved):
Looks good to me!

PR #9358, opened 2021-05-12
Fixes #32447 - Have task goto warn state instead of paused on error

chris1984:
With patch showing no regression on installing packages:

https://nimb.ws/M8blXr

Task now fails and does not goto a paused state:

https://nimb.ws/HlrxY7

@adamruzicka here is the pr, thanks for the pointer.

### DIFF
--- a/app/lib/actions/katello/agent_action.rb
+++ b/app/lib/actions/katello/agent_action.rb
@@ -49,6 +49,8 @@ module Actions
 
             schedule_timeout(timeout)
           end
+        when Dynflow::Action::Skip
+          # Do not fail and goto a paused state, instead skip and send the state to warning so we do not block other host actions
         when Dynflow::Action::Timeouts::Timeout
           process_timeout
         when 'accepted'


app/lib/actions/katello/agent_action.rb
```diff
@@ -28,6 +28,7 @@ def plan(host, options)
       end
 
       def run(event = nil)
+        return if event == Dynflow::Action::Skip
```
jturel, line 31: can you add an additional item in the below case statement with a blank body to achieve the same? Would like a unit test as well :)

app/lib/actions/katello/agent_action.rb
```diff
@@ -28,6 +28,8 @@ def plan(host, options)
       end
 
       def run(event = nil)
+        require 'pry'
```
jturel, line 31: Looks like something went wrong in this last push

app/lib/actions/katello/agent_action.rb
```diff
@@ -49,6 +49,7 @@ def run(event = nil)
 
             schedule_timeout(timeout)
           end
+        when Dynflow::Action::Skip # rubocop:disable Lint/EmptyWhen
```
jturel, line 52: if you add a comment inside the when block here you can a) avoid disabling this particular lint check b) make some future bug-fixer who has to read this code grateful for your comment explaining what this thing is (your PR name pretty much articulates that!)

chris1984: Updated with a comment :)

app/lib/actions/katello/agent_action.rb
chris1984: We decided this is not possible after discussion in IRC, marking resolved.

app/lib/actions/katello/agent_action.rb
chris1984: Outdated, resolved.

adamruzicka:
Looks good to me.

jturel:
```
[vagrant@centos7-katello-devel katello]$ hammer host package install --host-id=1 --packages=food
[.....................................................................................................................................................................................] [100%]
food: No package(s) available to install
Error: food: No package(s) available to install
[vagrant@centos7-katello-devel katello]$ hammer host package remove --host-id=1 --packages=screen
[.....................................................................................................................................................................................] [100%]
screen-4.1.0-0.27.20120314git3c2946.el7_9.x86_64
```

Behavior looks good when a failed agent action precedes one that should succeed (no lock error reported)!

vs master:

```
[vagrant@centos7-katello-devel katello]$ hammer host package install --host-id=1 --packages=food
[........................................................................................................................................                                              ] [75%]
food: No package(s) available to install
Error: food: No package(s) available to install
[vagrant@centos7-katello-devel katello]$ hammer host package install --host-id=1 --packages=screen
Could not install packages:
  Required lock is already taken by other running tasks.
  Please inspect their state, fix their errors and resume them.
  
  Conflicts with tasks:
  - http://centos7-katello-devel.jturel.example.com/foreman_tasks/tasks/8f284efc-d937-47e9-833f-ab615e1c9532


```

jturel (approved):
ACK

PR #9360, opened 2021-05-12
Fixes #32582 - old rh repo root arch stops disabling

ianballou:
The addition of https://github.com/Katello/katello/blob/master/app/models/katello/candlepin/repository_mapper.rb#L73 caused issues with disabling RH repos whose roots had an arch other than 'noarch' and a `root.content.content_url` that didn't have `$basearch`.

My PR adds a migration that changes these root repos to have 'noarch' as their architectures.

To test:
1) Enable a Red Hat repo whose repo-set (`::Katello::Content`) has a `content_url` that doesn't say `$basearch`.  One example I found was any RHEL 8 AppStream repo.
2) Find that repository's root and change the `arch` to something other than 'noarch'.
3) Try to disable the repo and see the error: `Repository not found`
4) Apply this PR and migrate
5) Try disabling the repo again

### DIFF
--- a/app/views/katello/api/v2/repository_sets/show.json.rabl
+++ b/app/views/katello/api/v2/repository_sets/show.json.rabl
@@ -38,16 +38,20 @@ child @resource.repositories => :repositories do
 end
 
 node :override do |pc|
-  pc.override
+  pc.override if pc.respond_to? :override
 end
 
 node :overrides do |pc|
-  pc.content_overrides.map do |override|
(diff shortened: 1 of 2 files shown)


jturel:
Does the migration address the root cause of the issue as well?

ianballou:
> Does the migration address the root cause of the issue as well?

When the line of code in my comment above was introduced it didn't take into account the older systems that had corrupt data due to the bug it fixed, so I suppose the migration is the root fix of the issue.

PR #9361, opened 2021-05-12
Fixes #29898,#31411 - applicability improvements

jlsherrill:
This pr resolves a couple issues:

1) Inserting hosts into the applicable host queue can cause a race condition where the 
    thread consuming the queue items are consuming at the same rate as the thread pushing host ids onto it.
    In that case you end up with applicability generation tasks that are only servicing a single host (or two)
2) run applicability once after a sync plan run.  This also refactors repo applicability generation to use db columns
    to determine if it needs to be run instead of the 'contents_changed' mechanism

### DIFF
--- a/app/controllers/katello/api/v2/products_bulk_actions_controller.rb
+++ b/app/controllers/katello/api/v2/products_bulk_actions_controller.rb
@@ -43,7 +43,6 @@ module Katello
       task = async_task(::Actions::BulkAction,
                         ::Actions::Katello::Repository::Sync,
                         syncable_repositories,
-                        nil,
                         :skip_metadata_check => skip_metadata_check,
                         :validate_contents => validate_contents)
 
--- a/app/controllers/katello/api/v2/repositories_controller.rb
+++ b/app/controllers/katello/api/v2/repositories_controller.rb
@@ -286,7 +286,7 @@ module Katello
         fail HttpErrors::BadRequest, _("attempted to sync without a feed URL")
       end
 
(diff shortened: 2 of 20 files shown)


jlsherrill:
To test this, i registered ~100 hosts with this script running on a single box over and over:

```
subscription-manager clean
echo "{\"dmi.system.uuid\":\"`uuidgen`\"}" > /etc/rhsm/facts/uuid.facts
echo "{\"virt.uuid\":\"`uuidgen`\"}" > /etc/rhsm/facts/virtuuid.facts

NAME=`uuidgen`
echo "{\"network.fqdn\":\"$NAME\"}" > /etc/rhsm/facts/fqdn.facts 
echo "{\"network.hostname\":\"$NAME\"}" > /etc/rhsm/facts/hostname.facts    

subscription-manager register --org=Default_Organization --environment=Library --user=admin --password=changeme
subscription-manager attach --pool=4028fa0d795cf40e01795d22d5aa0003
yum repolist &> /dev/null
```

Where the pool id is the pool of a product containing ~4 repos (or however many repos you want).

ianballou (approved):
I tested out the PR and it works fine! I tried doing bulk applicability regeneration while registering new hosts and I always saw the max amount of hosts per action.

While going through this, I realized that we don't recalculate applicability after deleting or uploading packages to a repository.  I wonder if we should.

jlsherrill:
We definitely should do it as part of upload and remove. i filed: https://projects.theforeman.org/issues/32601

PR #9363, opened 2021-05-14
Fixes #32578 - Updating a repo without an upstream url errors out

ianballou:
`extract_sles_token` broke the workflow of running update_remote without a url, which is supposed to signal the remote's deletion.  This PR fixes `extract_sles_token` to return `nil`s if there is no `root.url`.

To test, simply try clearing out a repo's upstream url and saving it. You will see the error in the redmine.

Also, check that the remote is actually deleted in Pulp 3. You can do that by recording the repo's `remote_href` from the foreman-console and then checking its existence with `pulp rpm remote show --href <href>`

### DIFF
--- a/app/services/katello/pulp3/repository/yum.rb
+++ b/app/services/katello/pulp3/repository/yum.rb
@@ -18,6 +18,7 @@ module Katello
         end
 
         def extract_sles_token
+          return [nil, nil] if root.url.blank?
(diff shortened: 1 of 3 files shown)


jjeffers (approved):
Verified the original error as suggested, then tested this PR: looks great, works as described.

PR #9365, opened 2021-05-17
Fixes #32604 - pin to psych 3.x

lzap:
VCR is configured to use Psych as the serializer. Psych 4.0 changed load to be safe_load. In tests we don't mind, the codebase does not appear to use any YAML loading.

More info:

https://community.theforeman.org/t/psa-rubygem-psych-4-0-is-out-and-its-a-breaking-change/23604

Edit: We need to pin psych gem.

### DIFF
--- a/katello.gemspec
+++ b/katello.gemspec
@@ -71,6 +71,8 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency "minitest-reporters"
   gem.add_development_dependency "mocha"
   gem.add_development_dependency "vcr", "< 4.0.0"
+  # https://github.com/Katello/katello/pull/9365
+  gem.add_development_dependency "psych", "< 4.0.0"
   gem.add_development_dependency "webmock"
   gem.add_development_dependency "rubocop-checkstyle_formatter"
   gem.add_development_dependency "simplecov"


test/models/docker_manifest_list_test.rb
```diff
@@ -8,7 +8,7 @@ class DockerManifestListTest < ActiveSupport::TestCase
 
     def setup
       FactoryBot.create(:smart_proxy, :default_smart_proxy)
-      @manifest_lists = YAML.load_file(MANIFESTS).values.map(&:deep_symbolize_keys)
+      @manifest_lists = Psych.unsafe_load_file(MANIFESTS).values.map(&:deep_symbolize_keys)
```
ezr-ondrej, line 11: This seems to be backwards incompatible so the tests are failing there.

lzap: Damn so both versions are in use at some point? I guess I need to pin this.

lzap:
Tests are green, we need to unblock CI. Thanks!

lzap:
@ehelms I think the CI is blocked by this.

ezr-ondrej (approved):
Can we please merge this?
We would like to run katello ci in foreman repo.

ehelms (approved):

lzap:
For the record, this patch was reverted in https://github.com/Katello/katello/pull/9374

PR #9368, opened 2021-05-18
Fixes #32619 - pins qpid_proton to resolve gem build issues

jjeffers:


ehelms (approved):
We are seeing build failures across the board with 0.34 version. At the time of this approval, EPEL only has 0.33 devel packages which may be needed to support 0.34.

PR #9369, opened 2021-05-18
Fixes #31825 - Add breadcrumb

sjha4:
Similar to [foreman](https://github.com/theforeman/foreman/blob/dc4201f5514b6b8a76a27577b0e488f07eaf6b70/webpack/assets/javascripts/react_app/components/HostDetails/index.js#L89) implementation..


sjha4:
[test katello]

jeremylenz (changes requested):
Thanks @sjha4, it's great to finally have breadcrumbs! :smile: 

PF4's `BreadcrumbItem` renders an `<a>` tag by default.  This works fine, but causes the whole page to reload when you click.  So I thought it would be better to use React Router's [`<Link>`](https://reactrouter.com/web/api/Link) instead.

```js
import { Link } from 'react-router-dom';
```

I first tried to use the `component` prop provided by Patternfly, like
```js
<BreadcrumbItem component={Link} ... />
```

But this doesn't work because `Link` needs a `to` prop and Patternfly isn't passing `to`, only `href`.

I was able to get it to work by using a render prop, like so:

```js
            <BreadcrumbItem
              aria-label="cv_breadcrumb"
              render={() => (<Link to="/labs/content_views" >{__('Content Views')}</Link>)}
            />
```

I hate render props but I still feel like it's a good idea to have React switch the page rather than start from scratch.  It's just so much faster.  Thoughts?

sjha4:
@jeremylenz : That is a lot lot better. Updated.. 👍🏽

jeremylenz (approved):
Changes LGTM, thanks @sjha4 :+1:

PR #9370, opened 2021-05-19
Fixes #32626 - Ignore empty upstream authentication

sbernhard:


chris1984:
@sbernhard can I test this with Microsoft Edge based on Chromium, or do you prefer straight Chrome?

sbernhard:
We had the issue with Chrome and Chromium. Maybe the same happens on Edge with Chromium but I don't know.

chris1984:
> We had the issue with Chrome and Chromium. Maybe the same happens on Edge with Chromium but I don't know.

Fine, I will install Chrome :P

chris1984 (approved):
ACK, works good on Google Chrome and Microsoft Edge. I was able to create the repo.

PR #9372, opened 2021-05-19
Fixes #32590 - always remove puppet and ostree repos

jlsherrill:
the previous detection of puppet repos did not account
for the situation where the user had a content view puppet
environment, but no longer had any puppet repos. This now
will run the cleanup code regardless


jlsherrill:
To test this, you would need to:

1.  install 3.18
2. create a puppet repo
3. upload a puppet module
4. publish the puppet module in a cv
5. delete the puppet repo
6. upgrade to 4.0
7. observe that an upgrade task fails 'correct_puppet_environments'

jlsherrill:
[test katello]

jjeffers (approved):
Test these changes per the scenario @jlsherrill described - the puppet repositories were removed and the installer completed without error.

PR #9373, opened 2021-05-19
Fixes #32623 - delete_orphaned_content fails if no Pulp 3

ianballou:
On 3.18, it's not enough to check that Pulp 3 is enabled on the smart proxy. We also need to check if Pulp 3 is enabled.

To test:
1) Spin up a 3.18 box
2) Ensure Pulp 2 is being used everywhere and disable the pulpcore-api service
3) Try running `bundle exec rake katello:delete_orphaned_content`
4) Patch in this PR
5) Try again


chris1984 (approved):
ACK works:

```bash
[root@centos7-katello-3-18 ~]# foreman-rake katello:delete_orphaned_content --trace
** Invoke katello:delete_orphaned_content (first_time)
** Invoke dynflow:client (first_time)
** Invoke environment (first_time)
** Execute environment
** Execute dynflow:client
** Execute katello:delete_orphaned_content
Orphaned content deletion started in background.
```

PR #9378, opened 2021-05-24
Fixes #32320 - Candlepin correctly errors out on Job Failures

parthaa:
Candlepin async task was previously not erroring out when a job failed
like manifest-import failure etc.
This commit addresses that


ianballou (approved):
Works for me!

PR #9379, opened 2021-05-24
Fixes #32654 - katello:correct_repositories should use dynflow client

ianballou:
To avoid the uninitialized Dynflow world error.

To test, apply this PR to a production box, delete a Katello repo's Pulp 3 repo with the api or Pulp CLI, then run `rake katello:correct_repositories`.

### DIFF
--- a/lib/katello/tasks/repository.rake
+++ b/lib/katello/tasks/repository.rake
@@ -51,7 +51,7 @@ namespace :katello do
   end
 
   desc "Correct missing pulp repositories. Specify CONTENT_VIEW=name and LIFECYCLE_ENVIRONMENT=name to narrow repositories.  COMMIT=true to perform operation."
-  task :correct_repositories => ["environment", "check_ping"] do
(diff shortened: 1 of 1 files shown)


lib/katello/tasks/repository.rake
```diff
@@ -51,7 +51,7 @@ namespace :katello do
   end
 
   desc "Correct missing pulp repositories. Specify CONTENT_VIEW=name and LIFECYCLE_ENVIRONMENT=name to narrow repositories.  COMMIT=true to perform operation."
-  task :correct_repositories => ["environment", "check_ping"] do
+  task :correct_repositories => ["dynflow:client", "environment", "check_ping"] do
     puts "All operations will be skipped.  Re-run with COMMIT=true to perform corrections." unless commit?
```
jlsherrill, line 55: 'check_ping' already has this prerequisite in master, is this still happening in master?

ianballou: Oh I missed that! It's not, but I thought it didn't always occur. I'll close the PR.

PR #9381, opened 2021-05-24
Fixes #32624 - use correct ca for content guard

jlsherrill:
and properly update content guard if ca changes


app/services/katello/pulp3/api/content_guard.rb
```diff
@@ -20,19 +22,50 @@ def rhsm_api_client
...
         end
 
-        def create(name = "RHSMCertGuard", ca_certificate = Cert::Certs.ca_cert)
-          data = PulpCertguardClient::CertguardRHSMCertGuard.new(name: name, ca_certificate: ca_certificate)
+        def ca_cert
+          Cert::Certs.consumer_ca_cert
+        end
+
+        def refresh
+          found = list(name: 'RHSMCertGuard').results.first
```
jlsherrill, line 30: TODO: i need to update this to use the constant

jlsherrill:
note that this required pretty much a full pulp3 re-record, and since i couldn't re-record the pulp2 tests, i had to delete them

jlsherrill:
Couple of notes:

* i noticed that a lot of tests were using a protected repo, not really intentionally.  This increased vcr cassette size, so i made the non-redhat repo fixtures be unprotected
* note that this required pretty much a full pulp3 re-record, and since i couldn't re-record the pulp2 tests, i had to delete them

app/services/cert/certs.rb
```diff
@@ -8,6 +8,10 @@ def self.ca_cert
       File.read(Setting[:ssl_ca_file])
     end
 
+    def self.consumer_ca_cert
```
ehelms, line 11:
```suggestion
    def self.candlepin_client_ca_cert
    def self.client_ca_cert
    def self.subscription_man_client_ca_cert
```

or

```suggestion

```

Trying to get away from using 'consumer' given I think that just leads to a layer of explanation. I suppose tihs could also be confusing.

app/services/katello/pulp3/api/content_guard.rb
```diff
@@ -20,19 +22,50 @@ def rhsm_api_client
...
+          if found
+            partial_update(found.pulp_href) if found.ca_certificate != ca_cert
```
ehelms, line 32:
```suggestion
         if found && found.ca_certificate != ca_cert
            partial_update(found.pulp_href)
```

app/services/katello/pulp3/api/content_guard.rb
```diff
@@ -20,19 +22,50 @@ def rhsm_api_client
...
+          else
+            create
+          end
+          persist_if_needed(found.pulp_href)
+        end
+
+        def persist_if_needed(href)
+          unless self.smart_proxy.pulp_mirror?
```
ehelms, line 40:
```suggestion
          return if self.smart_proxy.pulp_mirror?
```

Could do a guard clause here to reduce the amount of nesting.

ehelms (approved):
Worked for me, workflow I tested:

* installed nightly with custom certificates
* synced RH Ansible repository
* registered RHEL 7 client
* Got a 403
* Installed RPM with this patch and restarted services
* Re-synced repository to trigger CA certificate update (not ideal but this shouldnt change often)
* Went to client and was able to install ansible package without a 403

PR #9385, opened 2021-05-26
Fixes #32663 - ensure unique name for each migrated repo

jlsherrill:
Steps to Reproduce:
1.  Create ProductA
2.  Create a repo in ProductA with label 'foo' pointed to some repo, sync it
3.  Create Product B
4.  Create a repo in ProductB with label 'foo' pointed to some repo, sync it
5.  Add both 'foo' repos to a content view and publish it
6.  run foreman-maintain content prepare
7.  run foreman-maintain content migration-stats

notice that not all repos are migrated

### DIFF
--- a/app/services/katello/pulp3/migration_plan.rb
+++ b/app/services/katello/pulp3/migration_plan.rb
@@ -105,7 +105,15 @@ module Katello
       end
 
       def name_for_content_view(content_view, root_repo)
-        "#{content_view.label}-#{root_repo.label}"
+        name = "#{content_view.label}-#{root_repo.label}"
+
+        if Katello::RootRepository.where(:label => root_repo.label).group(:label).count(:label)[root_repo.label] > 1
+          repo_query = Katello::Repository.joins(:root, :content_view_version => :content_view).
+              where("#{::Katello::RootRepository.table_name}.id != #{root_repo.id}").
+              where("#{::Katello::RootRepository.table_name}.label" => root_repo.label, "#{::Katello::ContentView.table_name}.label" => content_view.label)
+          name += "-#{root_repo.id}" if repo_query.any?
+        end
+        name
       end
 
       def content_view_migration(content_view, root)
--- a/test/services/katello/pulp3/migration_plan_test.rb
+++ b/test/services/katello/pulp3/migration_plan_test.rb
@@ -1,0 +1,24 @@
+require 'katello_test_helper'
+module Katello
+  module Service
+    module Pulp3
+      class MigrationPlanTest < ActiveSupport::TestCase
+        def setup
+          repo_label = 'foo'
+          repoa = katello_repositories(:fedora_17_x86_64_library_view_2)
+          repob = katello_repositories(:rhel_6_x86_64_dev_archive)
+
+          repoa.root.update_column(:label, repo_label)
+          repob.root.update_column(:label, repo_label)
+        end
+
+        def test_conflicting_labels
+          plan = Katello::Pulp3::MigrationPlan.new(['yum']).generate
+          name_list = plan[:plugins].first[:repositories].map { |i| i[:name] }
+          assert_equal name_list.size, name_list.uniq.size
+          assert_include name_list, 'published_dev_view-foo'
+        end
+      end
+    end
+  end
+end


app/services/katello/pulp3/migration_plan.rb
```diff
@@ -105,7 +105,15 @@ def target_link_map
       end
 
       def name_for_content_view(content_view, root_repo)
-        "#{content_view.label}-#{root_repo.label}"
+        name = "#{content_view.label}-#{root_repo.label}"
+
+        if Katello::RootRepository.where(:label => root_repo.label).group(:label).count(:label)[root_repo.label] > 1
```
gvde, line 110: I don't really know ruby, but doesn't that translate to the SQL (root_repo.label being 2_3_el7_x86_64)

`select count(label) from katello_root_repositories where label = '2_3_el7_x86_64' group by label;`

(I don't understand what the `[root_repo.label]` does after that)

The group seems unnecessary here as all you want to do is to count the number of rows with the specific root_repo.label, i.e. in the end

`select count(*) from katello_root_repositories where label = '2_3_el7_x86_64';`

With my nonexisting ruby knowledge, that should be something like

`Katello::RootRepository.where(:label => root_repo.label).count`

?

app/services/katello/pulp3/migration_plan.rb
```diff
@@ -105,7 +105,15 @@ def target_link_map
...
+          repo_query = Katello::Repository.joins(:root, :content_view_version => :content_view).
+              where("#{::Katello::RootRepository.table_name}.label" => root_repo.label, "#{::Katello::ContentView.table_name}.label" => content_view.label)
+          duplicate_count = repo_query.group(:library_instance_id, :content_view_id).count.select { |k,v| v > 1 }.count
```
gvde, line 113: You are grouping by content_view_id here. So you expect multiple content views to have the same label, too? If that was possible (it is in the database) this would further complicate things.

app/services/katello/pulp3/migration_plan.rb
```diff
@@ -105,7 +105,15 @@ def target_link_map
       end
 
       def name_for_content_view(content_view, root_repo)
-        "#{content_view.label}-#{root_repo.label}"
+        name = "#{content_view.label}-#{root_repo.label}"
+
+        if Katello::RootRepository.where(:label => root_repo.label).group(:label).count(:label)[root_repo.label] > 1
+          repo_query = Katello::Repository.joins(:root, :content_view_version => :content_view).
+              where("#{::Katello::RootRepository.table_name}.label" => root_repo.label, "#{::Katello::ContentView.table_name}.label" => content_view.label)
```
gvde, line 112: Wouldn't it be better to select directly for the content_view.id instead as the content_view is a parameter to this function?

app/services/katello/pulp3/migration_plan.rb
```diff
@@ -105,7 +105,15 @@ def target_link_map
...
+          duplicate_count = repo_query.group(:library_instance_id, :content_view_id).count.select { |k,v| v > 1 }.count
+          name += "-#{root_repo.id}" if duplicate_count > 0
```
gvde, line 114: I guess, this would be a breaking change to previous prepare runs, i.e. I suppose it would be mandatory to reset the migration and start a fresh prepare before you use this code. As far as I can tell, this name makes it into the pulpcore.core_repository table, column name. Thus, if you run another prepare with this change without reset it would add the already migrated repository again with the new name as well.

gvde:
My initial thought how to fix this problem without breaking existing migrations was to only add a counter or identifier to the plan name if there was already one in the list.
As far as I understand (with my non-existing ruby knowledge) this is the function which collects the plans:
https://github.com/Katello/katello/blob/2e555c3f29c49d5d2129fdd9d898bd27c4c7c905/app/services/katello/pulp3/migration_plan.rb#L80-L86
Thus, instead of simply adding another plan to the plans array, my idea would be to scan the plans array for a plan with the same name as plan.name and if it exists, change the plan.name by adding a counter or root.id or something. With some more effort you could only change the plan names already in the plans array which should even allow for a change which does not break previous prepare runs, i.e. it should even work properly to run a prepare/switchover with a migration reset before.

gvde:
Another idea would be (if it is going to be a breaking change anyway) to simply add the product label into the name. The product label must be unique, if I understand correctly. You have the root_repo.id value and can access the root repository join with the products table to find the label and then just put the product label into the name.

app/services/katello/pulp3/migration_plan.rb
jlsherrill: Its possible across organizations, this was partially why i went with root_id, as that should be completely unique.  I do need to test more this scenario!

app/services/katello/pulp3/migration_plan.rb
jlsherrill: i'll write up an explanation about how the data looks and why (i think) this is a good solution

jlsherrill:
So here's my best explanation.  This is what the data model looks like, it may help in this discussion:

image


The query first joins all these tables together, so it could be visualized like this:

| root_repo.id | root_repo.label | repo.id | library_instance_id | content_view.id | content_view_version.id | content_view.label |
|--------------|-----------------|---------|---------------------|-----------------|-------------------------|--------------------|
| 1            | foo             | 4       | 3                   | 10              | 9                       | bar                |
| 1            | foo             | 5       | 3                   | 10              | 9                       | bar                |
| 2            | foo             | 7       | 6                   | 10              | 9                       | bar                |
| 2            | foo             | 8       | 6                   | 10              | 9                       | bar                |

joining on library_instance_id & content_view_id would result in something like this:

| root_repo.id | root_repo.label | repo.id | library_instance_id | content_view.id | content_view_version.id | content_view.label |
|--------------|-----------------|---------|---------------------|-----------------|-------------------------|--------------------|
| 1,1          | foo,foo         | 4,5     | 3                   | 10              | 9,9                     | bar,bar            |
| 2,2          | foo, foo        | 7,8     | 6                   | 10              | 9,9                     | bar,bar            |

and applying counts on it, would return something in ruby:

```
{
  [3,10] =>  2,
  [6,10] => 2
}
```

So this is showing  that there are two different sets of repos with the same label for content_view and repository.  I think this logic works, but i may have one small mistakes, where' i'm filtering out any 'set' that doesn't have more than '1' for the value of the returning hash, i don't think thats needed.

app/services/katello/pulp3/migration_plan.rb
jlsherrill: oh i think you're right! thats overly complicated

app/services/katello/pulp3/migration_plan.rb
jlsherrill: I'm grouping by content view and library instance_id to 'collapse' all the same ones together with that same pairing.

app/services/katello/pulp3/migration_plan.rb
jlsherrill: yes, and for a few repos thats not a big deal.  I was trying to avoid renaming every single repo as this would nearly be a re-migration from scratch.

gvde:
Thanks for the graphics. Very helpful. I have put most of it together going through the postgres database and the foreign keys, as well as the rake console.

jlsherrill:
pushed a small update based on my comment, still have more testing to do though

jlsherrill:
[test katello]

gvde:
My comment to line 114 regarding the duplicate_groups.any? condition which is always true, still applies. All my tests with the rake console show there is always one group. Thus, basically you are replacing the name for all, which have duplicate repo labels. If you do that anyway, the lines before a unnecessary...

jlsherrill:
oh i see what you are saying!

jlsherrill:
i realized i wasn't excluding the current content view when checking for duplicates, so if a content view already was published with a repo it was showing up as a duplicate.  Let me know how that looks to you!

app/services/katello/pulp3/migration_plan.rb
```diff
@@ -105,7 +105,16 @@ def target_link_map
       end
 
       def name_for_content_view(content_view, root_repo)
-        "#{content_view.label}-#{root_repo.label}"
+        name = "#{content_view.label}-#{root_repo.label}"
+
+        if Katello::RootRepository.where(:label => root_repo.label).group(:label).count(:label)[root_repo.label] > 1
+          repo_query = Katello::Repository.joins(:root, :content_view_version => :content_view).
```
gvde, line 111: Now that doesn't work at all for me. This generates the following SQL in my scenario:

`SELECT "katello_repositories".* FROM "katello_repositories" INNER JOIN "katello_root_repositories" ON "katello_root_repositories"."id" = "katello_repositories"."root_id" INNER JOIN "katello_content_view_versions" ON "katello_content_view_versions"."id" = "katello_repositories"."content_view_version_id" INNER JOIN "katello_content_views" ON "katello_content_views"."id" = "katello_content_view_versions"."content_view_id" WHERE (katello_content_views.id != 12) AND "katello_root_repositories"."label" = '2_3_el7_x86_64' AND "katello_content_views"."label" = 'centos7-epel7'`

which is always empty, unless maybe you have the same content view label in different organizations. 12 is the id of the content view 'centos7-epel7', i.e. this is never true for me.

gvde: I think the original statement was good, you just needed different grouping.

```
Katello::Repository.joins(:root, :content_view_version => :content_view)
.where("#{::Katello::RootRepository.table_name}.label" => root_repo.label,
"#{::Katello::ContentView.table_name}.label" => content_view.label)
```

generates the SQL

```
SELECT "katello_repositories".*
FROM "katello_repositories"
 INNER JOIN "katello_root_repositories" ON "katello_root_repositories"."id" = "katello_repositories"."root_id"
 INNER JOIN "katello_content_view_versions" ON "katello_content_view_versions"."id" = "katello_repositories"."content_view_version_id"
 INNER JOIN "katello_content_views" ON "katello_content_views"."id" = "katello_content_view_versions"."content_view_id"
WHERE "katello_root_repositories"."label" = '2_3_el7_x86_64'
      AND "katello_content_views"."label" = 'centos7-epel7'
```

which correctly contains all repositories which have a root_repo_label of '2_3_el7_x86_64' in any content view with label 'centos7-epel7'. Now to find out whether these are in fact different root_repos which would create the identical names in the migration plan, I think you would have to group by "katello_root_repositories".id and "katello_content_views"."id" resulting in an SQL like this:

```
SELECT "katello_root_repositories".id, "katello_content_views"."id"
FROM "katello_repositories"
 INNER JOIN "katello_root_repositories" ON "katello_root_repositories"."id" = "katello_repositories"."root_id"
 INNER JOIN "katello_content_view_versions" ON "katello_content_view_versions"."id" = "katello_repositories"."content_view_version_id"
 INNER JOIN "katello_content_views" ON "katello_content_views"."id" = "katello_content_view_versions"."content_view_id"
WHERE "katello_root_repositories"."label" = '2_3_el7_x86_64'
   AND "katello_content_views"."label" = 'centos7-epel7'
GROUP BY "katello_root_repositories".id, "katello_content_views"."id";
 id  | id 
-----+----
 138 | 12
 140 | 12
(2 rows)
```
So basically the statement before by with the GROUP BY (and adjusted SELECT expression).

These are all root_repo ids with the identical root_repo label connected to the any content view with the identical content view label. If this query results in more than one row, it's a duplicate which needs the name changed.

To compare: I have multiple repositories labeled "gluster-7" which however do not coexist in the same content view (one is centos8 the other is centos7). Due to that, there is no conflict:
```
SELECT "katello_root_repositories".id, "katello_content_views"."id"
FROM "katello_repositories"
 INNER JOIN "katello_root_repositories" ON "katello_root_repositories"."id" = "katello_repositories"."root_id"
 INNER JOIN "katello_content_view_versions" ON "katello_content_view_versions"."id" = "katello_repositories"."content_view_version_id"
 INNER JOIN "katello_content_views" ON "katello_content_views"."id" = "katello_content_view_versions"."content_view_id"
WHERE "katello_root_repositories"."label" = 'gluster-7' AND "katello_content_views"."label" = 'centos7-epel7'
group by "katello_root_repositories".id, "katello_content_views"."id";
 id | id 
----+----
 82 | 12
(1 row)
```

Now if I only knew ruby and were able to get this into a ruby statement...

gvde (changes requested):

app/services/katello/pulp3/migration_plan.rb
gvde: Another note: I only have a single organization so I don't really know if this can happen or not. As far as I understand the content view label is only unique within one organization, i.e. two organizations could have content views with the identical label. There would also be duplicate identical names if it is possible to have the same root repo (i.e. the same katello_root_repositories.id) in several content views with the same label of different organizations. That's the same issue, just not on the root_repo.label part but on the content_view.label.

name = "#{content_view.label}-#{root_repo.label}"

But I don't know if it is at all possible to share a root_repo between organizations...

gvde: O.K. By experimenting with the rake console I think I found the all-in-one statement to find out whether the name must be changed or not:

```
Katello::Repository.joins(:root, :content_view_version => :content_view)
.where("#{::Katello::RootRepository.table_name}.label" => root_repo.label,
       "#{::Katello::ContentView.table_name}.label" => content_view.label)
.select('katello_root_repositories.id')
.group('katello_root_repositories.id','katello_content_views.id').count.count
```

If that is > 1 the name must be changed...

jlsherrill:
@gvde i still saw issues with your suggestion.  We may be hitting different edge cases, as i'm testing across orgs and other situations.   I realized i think that i may have been over thinking this the whole time.  I went back to the basics and thought, 'what do i really need to know'.   When generating the name, we need to know, given a content view and root_repo:

* are there any root repos with the same label, 
* published in a content view with the same label
* excluding this particular root_repo

And i think i came up with something that is both simpler and more accurate. Let me know what you think

jlsherrill:
@gvde also, i just wanted to thank you for your help on this!

app/services/katello/pulp3/migration_plan.rb
gvde: I think that should still simply be 
```
if Katello::RootRepository.where(:label => root_repo.label).count > 1
```

Due to the where, there is always only a single group, making the whole group part unnecessary...

gvde (changes requested):

gvde:
> @gvde i still saw issues with your suggestion. We may be hitting different edge cases, as i'm testing across orgs and other situations. I realized i think that i may have been over thinking this the whole time. I went back to the basics and thought, 'what do i really need to know'. When generating the name, we need to know, given a content view and root_repo:

I don't know what issues you see. I was just looking at the database tables. For that part I think it does exactly what is needed: it finds all combinations of root_repos and content_views with the root_repo.label and content_view.label linked together through the content_view_version table and checks if there is any combo with different id for root_repo or content_view.

I have tested it with this code and the result is fine. It added the root_repo.id to the names which need it. All names in the migration plan are now unique.

```
      def name_for_content_view(content_view, root_repo)
        name = "#{content_view.label}-#{root_repo.label}"

        if Katello::Repository.joins(:root, :content_view_version => :content_view).where("#{::Katello::RootRepository.table_name}.label" => root_repo.label, "#{::Katello::ContentView.table_name}.label" => content_view.label).select('katello_root_repositories.id').group('katello_root_repositories.id','katello_content_views.id').count.count > 1
          name += "-#{root_repo.id}"
        end
        name
      end
```

Of course, I only have one organization. From the SQL code I don't see how this is not complete and correct for all cases. As far as I can see the migration plan doesn't distinguish between organizations for the names. Thus, the name conflict can appear at any organization, even across organizations. That's why it's enough to search for the labels, group by the id and count the number of resulting groups (that's why it's .count.count). So I don't know what issues you may see here. From my understanding and the SQL perspective, it seems correct.

The only potential issue I see at the moment is whether this issue with the name can only happen because of duplicate root_repos or if this issue can also happen with an identical root_repo but different content views using the same name. (See my "Another note:" above).

The name consists of two parts `name = "#{content_view.label}-#{root_repo.label}"`. Currently it's only dealing with the duplicate root_repo.labels. Can the same issue also happen with the content_view.label?

gvde:
By the way: did you test what happens if you run a prepare first with the original code and then another prepare with the patched code? Is it potentially a problem if a repository has been migrated before with the "simple" name and then later again with the name including the id? Does the remove orphan clean this up?

jlsherrill:
root_repos cannot be reused across orgs, but labels can.  In my testing i used a 2nd org with a duplicate cv name and repo label. 

I did test re-migrating after this change, but i haven't done a full switchover and tested that everything was okay.  I'll do that.

gvde:
> root_repos cannot be reused across orgs, but labels can. In my testing i used a 2nd org with a duplicate cv name and repo label.

O.K. In that case and because a cv label within an org must be unique, it should be sufficient to augment the name with the root_repo.id as it should always be two different root_repo.ids if there are two cvs with identical labels.

So just coming from the SQL database point of view my suggested code above is correct. It translates to something like this (even though the sub-query count is not done by SQL in the ruby code, but by counting records).

```
SELECT COUNT(*)
FROM (
  SELECT katello_root_repositories.id,katello_content_views.id,count(*)
  FROM "katello_repositories" INNER JOIN "katello_root_repositories" ON "katello_root_repositories"."id" = "katello_repositories"."root_id" 
        INNER JOIN "katello_content_view_versions" ON "katello_content_view_versions"."id" = "katello_repositories"."content_view_version_id"
        INNER JOIN "katello_content_views" ON "katello_content_views"."id" = "katello_content_view_versions"."content_view_id"
  WHERE "katello_root_repositories"."label" = '2_3_el7_x86_64' AND "katello_content_views"."label" = 'centos7-epel7'
  GROUP BY katello_root_repositories.id, katello_content_views.id ) sub;
```

It links all the root_repos with the content_views through katello_content_view_versions, selects only those with the respective labels and then groups by the root_repo.id and content_view.id which uniquely identify those.

(Sidenote: And this is even correct if it was possible to share root_repos across orgs, even though the augmented name in the next line isn't unique as it is not including the content-view.id).

So from my understanding, which is looking at the database, my suggested code is correct, except maybe for something due to my complete lack of experience with ruby.

jjeffers:
> 6\. run foreman-rake content prepare
I assume this is actually `foreman-maintain content prepare` ?

jlsherrill:
correct

jturel:
Is anything outstanding here? Asking for a friend reviewing the 3.18.4 release :)

jjeffers:
> Is anything outstanding here? Asking for a friend reviewing the 3.18.4 release :)

While testing this on a 3.18 production VM, I ran into issues around pulp tasks (so the migration never completed). I can't say if this is ready to go yet, but I can see if I can retest ASAP.

jjeffers:
> Is anything outstanding here? Asking for a friend reviewing the 3.18.4 release :)

Retested after applying a systemctl restart to the pulp resource manager - success!

```
vagrant@centos7-katello-3-18 katello-3.18.3.1]$ sudo foreman-maintain content migration-stats
Running Retrieve Pulp 2 to Pulp 3 migration statistics
================================================================================
Retrieve Pulp 2 to Pulp 3 migration statistics: 
============Migration Summary================
Migrated/Total RPMs: 0/32
Migrated/Total errata: 0/4
Migrated/Total repositories: 0/1

Estimated migration time based on yum content: fewer than 5 minutes

Note: ensure there is sufficient storage space for /var/lib/pulp/published to triple in size before starting the migration process.
Check the size of /var/lib/pulp/published with 'du -sh /var/lib/pulp/published/'
                                                                      [OK]
--------------------------------------------------------------------------------

[vagrant@centos7-katello-3-18 katello-3.18.3.1]$ sudo foreman-maintain content migration-stats
Running Retrieve Pulp 2 to Pulp 3 migration statistics
================================================================================
Retrieve Pulp 2 to Pulp 3 migration statistics: 
============Migration Summary================
Migrated/Total RPMs: 0/92
Migrated/Total errata: 0/4
Migrated/Total repositories: 0/2

Estimated migration time based on yum content: fewer than 5 minutes

Note: ensure there is sufficient storage space for /var/lib/pulp/published to triple in size before starting the migration process.
Check the size of /var/lib/pulp/published with 'du -sh /var/lib/pulp/published/'
                                                                      [OK]
--------------------------------------------------------------------------------

[vagrant@centos7-katello-3-18 katello-3.18.3.1]$ sudo systemctl restart pulp_resource_manager.service
[vagrant@centos7-katello-3-18 katello-3.18.3.1]$ sudo foreman-maintain content prepare
Running Prepare content for Pulp 3
================================================================================
Prepare content for Pulp 3: 
Checking for valid Katello configuraton.
Starting task.
2021-06-25 15:12:50 +0000: Content migration starting. 
Content Migration completed successfully

Script started on Fri 25 Jun 2021 03:12:37 PM UTC
Checking for valid Katello configuraton.
Starting task.
2021-06-25 15:12:50 +0000: Content migration starting. 
Content Migration completed successfully
                            [OK]
--------------------------------------------------------------------------------

[vagrant@centos7-katello-3-18 katello-3.18.3.1]$ sudo foreman-maintain content migration-stats
Running Retrieve Pulp 2 to Pulp 3 migration statistics
================================================================================
Retrieve Pulp 2 to Pulp 3 migration statistics: 
============Migration Summary================
Migrated/Total RPMs: 92/92
Migrated/Total errata: 4/4
Migrated/Total repositories: 2/2

Estimated migration time based on yum content: fewer than 5 minutes

Note: ensure there is sufficient storage space for /var/lib/pulp/published to triple in size before starting the migration process.
Check the size of /var/lib/pulp/published with 'du -sh /var/lib/pulp/published/'
                                                                      [OK]
--------------------------------------------------------------------------------
```
This LGTM.

jjeffers:
> > Is anything outstanding here? Asking for a friend reviewing the 3.18.4 release :)
> 
> While testing this on a 3.18 production VM, I ran into issues around pulp tasks (so the migration never completed). I can't say if this is ready to go yet, but I can see if I can retest ASAP.

Retested along with an update https://github.com/Katello/katello/pull/9428 and all is well. The migration proceeded and completed, with all repositories migrated to pulpcore. Additionally, after the switchover, content could be installed on a host via yum (from one of the migrated repositories).

gvde:
> > Is anything outstanding here? Asking for a friend reviewing the 3.18.4 release :)
> 
> While testing this on a 3.18 production VM, I ran into issues around pulp tasks (so the migration never completed). I can't say if this is ready to go yet, but I can see if I can retest ASAP.

Have you tested what happens if you made one migration prepare before patching, then one after patching and then switchover? Is the original repository with the unmodified name still in pulpcore? Does it get removed with as orphan later?

I guess not everybody running into this issue will reset the migration after the update but simply try another migrate/switchover which would succeed but will also contain the old migrated repos. The repo_name makes it into the pulpcore.core_repository table...

jjeffers (approved):
Approved pending @jlsherrill 's response to comments from @gvde

jlsherrill:
> Have you tested what happens if you made one migration prepare before patching, then one after patching and then switchover? Is the original repository with the unmodified name still in pulpcore? Does it get removed with as orphan later?

I ran through this again just to make sure and the repos gets published properly in pulp3.  I checked the repodata within the CV after switchover and it all looked sane.  

In this case an extra repo wasn't created in pulp3 because two repos can't have the same name.  But if there was an extra repo it would get cleaned up upon orphan cleanup.


Merging!

PR #9386, opened 2021-05-27
Fixes #32303 - Add show_all endpoint for component content_views

sjha4:
This is similar to repository/show_all endpoint and will be used in the UI to easily show added/not added cvs for a composite cv

The new endpoint is: /katello/api/v2/content_views/$CV_ID/content_view_components/show_all
It should also support pagination and searching.

### DIFF
--- a/app/controllers/katello/api/v2/content_view_components_controller.rb
+++ b/app/controllers/katello/api/v2/content_view_components_controller.rb
@@ -1,9 +1,10 @@
 module Katello
   class Api::V2::ContentViewComponentsController < Api::V2::ApiController
     include Katello::Concerns::FilteredAutoCompleteSearch
-    before_action :find_composite_content_view, :only => [:show, :index]
+    before_action :find_composite_content_view, :only => [:show, :index, :show_all]
     before_action :find_composite_content_view_for_edit, :only => [:add_components, :remove_components, :update]
     before_action :find_authorized_katello_resource, :only => [:show, :update]
+    before_action :find_organization_from_cv, :only => [:show_all]
 
     wrap_parameters :include => %w(composite_content_view_id content_view_version_id content_view_id latest)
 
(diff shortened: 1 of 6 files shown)


app/models/katello/content_view.rb
```diff
@@ -107,6 +107,18 @@ def self.published_with_repositories(root_repository)
...
+    def self.in_organization(org)
+      where(organization_id: org.id)
+    end
+
+    def component_of?(content_view)
+      Katello::ContentViewComponent.where(composite_content_view_id: content_view.id, content_view_id: id).any?
+    end
+
+    def component_content_view_version(content_view)
+      Katello::ContentViewComponent.where(composite_content_view_id: content_view.id, content_view_id: id)&.first
```
jlsherrill, line 119: small nitpic, but i think find_by() would be a tad simpler

app/views/katello/api/v2/content_view_components/show_all.json.rabl
```diff
@@ -0,0 +1,12 @@
+object false
+
+extends "katello/api/v2/common/metadata"
+
+child @collection[:results] => :results do
+  extends 'katello/api/v2/content_views/base'
+  node(:added_to_content_view) { |cv| cv.component_of?(@view) }
+  node(:cv_versions) do |cv|
```
jlsherrill, line 8: i'd prefer something like 'content_view_versions' for the node name

app/views/katello/api/v2/content_view_components/show_all.json.rabl
```diff
@@ -0,0 +1,12 @@
...
+    @version = cv.component_content_view_version(@view)
+    partial('katello/api/v2/content_view_components/show', :object => @version)
+  end
+end
```
jlsherrill, line 12: i'm trying to understand the reasoning behind what's included.  The previous example of this (repositories/show_all) extended the repostiories/base partial and 'enhanced' it by adding 'added_to_content_view'.  Can we do something similar here?

sjha4: Updated as per our discussion yesterday to use a presenter and the existing index template.

app/controllers/katello/api/v2/content_view_components_controller.rb
```diff
@@ -28,6 +29,32 @@ def index_response
...
+         CAST (#{kcc}.composite_content_view_id as BOOLEAN) ASC, #{kc}.name
+      SQL
+
+      query = Katello::ContentView.readable.in_organization(@organization)
+      query = query&.non_composite&.non_default
+      custom_sort = ->(sort_query) { sort_query.joins(join_query).order(order_query) }
+      options = { resource_class: Katello::ContentView, custom_sort: custom_sort }
+      collection = scoped_search(query, nil, nil, options)
+      collection[:results] = ComponentViewPresenter.component_presenter(@view, views: collection[:results])
+      respond_for_index(:collection => collection, :template => "index")
```
jlsherrill, line 55: do you not need a custom rabl to include the existing 'content view component' rabl, but also an extra attribute to show if its in the content view for real or not?

jlsherrill: or maybe the existing rabl is enough? (for example id of the object would be nil).  Maybe being explicit is better, but curious your thoughts

jlsherrill: that works for me!  I'll test this api out a bit.  Code looks good

jlsherrill (approved):
ACK!

sjha4:
Thanks Justin! Might need some updates when I get to the UI bit..Merging this..

PR #9390, opened 2021-06-02
Fixes #32601 - Applicability regen after upload and remove

rverdile:
To test:

1) Register a content host
2) Create empty repo
3) Subscribe content host to product with empty repo
4) Install walrus-0.71 from [https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/](https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/) on content host
5) Add walrus-5.21 from [https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/](https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/) to empty repo
	- walrus-5.21 will be shown as an upgradable package on the content host (bug fix - did not previously update automatically)
6) Remove walrus-5.21 from repo
	- will no longer be shown as an upgradable package on the content host (bug fix - did not previously update automatically)

### DIFF
--- a/app/lib/actions/katello/repository/remove_content.rb
+++ b/app/lib/actions/katello/repository/remove_content.rb
@@ -19,6 +19,8 @@ module Actions
           content_unit_ids = content_units.map(&:id)
           content_unit_type = options[:content_type] || content_units.first.class::CONTENT_TYPE
 
+          generate_applicability = options.fetch(:generate_applicability, repository.yum?)
+
           sequence do
             remove_content_args = {
               :contents => content_unit_ids,
@@ -31,6 +33,7 @@ module Actions
             return if pulp_action.error
             plan_self(:content_unit_class => content_units.first.class.name, :content_unit_ids => content_unit_ids)
             plan_action(CapsuleSync, repository) if sync_capsule
+            plan_action(Actions::Katello::Applicability::Repository::Regenerate, :repo_ids => [repository.id]) if generate_applicability
           end
         end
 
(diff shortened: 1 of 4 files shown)


app/lib/actions/katello/repository/upload_files.rb
```diff
@@ -30,6 +33,7 @@ def plan(repository, files, content_type = nil)
 
             plan_action(FinishUpload, repository, content_type: content_type, upload_actions: upload_actions)
             plan_self(tmp_files: tmp_files)
+            plan_action(Actions::Katello::Applicability::Repository::Regenerate, :repo_ids => [repository.id]) if generate_applicability
           end
         ensure
```
jlsherrill, line 38: There's actually two ways to upload content.  The first takes the entire file and uploads it (i believe thats this action).  The problem with that is that only files of fairly small sizes are accepted.  This is what the UI uses.

The cli uses a much more complicated series of steps which breaks the file into chunks and uploads each of them and then finally triggers an import.  `app/lib/actions/katello/repository/import_upload.rb`  is what handles that. 

You can test this by running the 'hammer repository upload' command

app/lib/actions/katello/repository/remove_content.rb
```diff
@@ -7,6 +7,7 @@ class RemoveContent < Actions::EntryAction
 
         def plan(repository, content_units, options = {})
           sync_capsule = options.fetch(:sync_capsule, true)
+
```
ianballou, line 10: You can remove this newline

ianballou (commented):
Works fine for me! Just one comment.

ianballou (approved):
Looks fine now! Just need to wait for Jenkins.

PR #9393, opened 2021-06-07
Fixes #32740 - Orphan cleanup fails on boxes with only pulp3

sjha4:
To test:
On a pulp3 only box:
1. Run bundle exec rails katello:delete_orphaned_content


jjeffers (approved):
After confirming the original error on a pulp3 only installation, the applied PR resolved the problem. 

AWEP

PR #9394, opened 2021-06-07
Fixes #31616 - Remove content_types requirement in katello.yaml and use of it

ianballou:
This PR removes the need for configuring `content_types` by registering types based only only on what Pulp 3 plugins are installed. The `use_pulp_2_for_content_type` config option has also been removed.

I'm keeping this as a draft for now while I fix all of the tests.

To test:
1) Remove the entire `content_types` section from the Katello configuration.  Try using content types like normal.
2) Shutdown Katello, and remove all Pulp 3 capabilities from your smart proxy with the following:
```ruby
SmartProxy.pulp_primary.smart_proxy_features.where(:feature_id => Feature.find_by(:name => SmartProxy::PULP3_FEATURE).id).update(:capabilities => [])
```
3) Ensure that you can still use all content types like normal without intervention after starting up Katello


app/services/katello/repository_type_manager.rb
```diff
@@ -1,35 +1,78 @@
...
+        when ::Katello::Repository::DOCKER_TYPE
+          'container'
+        when ::Katello::Repository::FILE_TYPE
+          'file'
+        when ::Katello::Repository::ANSIBLE_COLLECTION_TYPE
+          'ansible'
+        when ::Katello::Repository::DEB_TYPE
+          'deb'
+        end
+      end
```
jlsherrill, line 61: I think this mapping may already exist indirectly 

```
Katello::RepositoryTypeManager.repository_types['docker'].pulp3_plugin
```

app/controllers/katello/api/v2/repositories_controller.rb
```diff
@@ -89,7 +89,7 @@ def custom_index_relation(collection)
     param :ostree_branch_id, String, :desc => N_("Id of an ostree branch to find repositories that contain that branch")
     param :library, :bool, :desc => N_("show repositories in Library and the default content view")
     param :archived, :bool, :desc => N_("show archived repositories")
-    param :content_type, RepositoryTypeManager.repository_types.keys, :desc => N_("limit to only repositories of this type")
+    param :content_type, RepositoryTypeManager.usable_repository_types.keys, :desc => N_("limit to only repositories of this type")
```
jlsherrill, line 92: This controls apipie generation which means that the apipie cache might be out of date, especially if the capabilities were updated.  Couple of ideas:

1) use repository_types and advertise all types to hammer
2) use repository_types only if usable_repository_types is empty
3) something else?

app/services/katello/repository_type_manager.rb
```diff
@@ -95,6 +146,15 @@ def find_content_type!(katello_label)
       def enabled?(repository_type)
         find(repository_type).present?
       end
```
jlsherrill, line 148: should this `enabled?` method be updated to mean 'usable'?    (and i guess maybe we could reconcile on one term. usable or enabled)

I believe the navigation configuration needs updating: https://github.com/Katello/katello/blob/master/lib/katello/plugin.rb#L94-L177

to have the check be done at runtime instead of startup time

app/services/katello/repository_type_manager.rb
ianballou: The issue is that I need this mapping before the repository types are registered.

app/services/katello/repository_type_manager.rb
ianballou: I think I like enabled better. I'll switch to that.

app/controllers/katello/api/v2/repositories_controller.rb
ianballou: I think using `defined_repository_types` might be fine.

app/services/katello/repository_type_manager.rb
ianballou: > to have the check be done at runtime instead of startup time

I'm thinking of trying to update the menu every time I update the `enabled_repository_types`.  Does that sound like what you were thinking? I'm still looking into what that entails.

jlsherrill: that might work, the other option is that navigation menu items can accept a lambda https://github.com/theforeman/foreman/blob/develop/app/registries/menu/manager.rb#L60

and it could be looked up on demand.  We'd need to make sure its super fast though

app/services/katello/repository_type_manager.rb
jlsherrill: i dug through the code and it looks like this is called during the repository_type registration, and its actually registered by this point. I tested this change and it worked fine, maybe it didn't on a previous incarnation?

jlsherrill:
code looks good, i will do some testing

ianballou:
Quite a few tests need to be rewritten since I'm removing even more Pulp 2 related code. The one I'm currently on is `test/models/rpm_test.rb`, which is currently quite Pulp 2 focused.

lib/katello/plugin.rb
```diff
@@ -91,89 +91,85 @@
...
-    end
+    menu :top_menu,
+         :debs,
+         :caption => N_('Deb Packages'),
+         :url => '/debs',
+         :url_hash => {:controller => 'katello/api/v2/debs',
+                       :action => 'index'},
+         :engine => Katello::Engine,
+         :turbolinks => false,
+         :if => lambda { ::Katello::RepositoryTypeManager.enabled?(::Katello::Repository::DEB_TYPE) }
```
ianballou, line 102: This `:if` doesn't work in Foreman anymore. It should be fairly straightforward to add it back in: https://projects.theforeman.org/issues/32883

ianballou:
Here's the Foreman PR to get conditional menu support at runtime: https://github.com/theforeman/foreman/pull/8622

ianballou:
The Foreman PR above has been merged.

app/services/katello/repository_type_manager.rb
```diff
@@ -41,29 +75,37 @@ def creatable_by_user?(repository_type)
...
+      def find_defined(repository_type)
+        defined_repository_types[repository_type.to_s]
+      end
+
       def find(repository_type)
-        repository_types[repository_type.to_s]
+        found_repo_type = enabled_repository_types[repository_type.to_s]
+        if found_repo_type.blank?
+          found_repo_type = update_enabled_repository_type(repository_type.to_s)
+        end
```
jlsherrill, line 99: I did some performance testing and calling find() within plugin.rb was taking about ~250ms for each request.  Each call to find() was taking about ~20ms.  I think this is too much and so i dug around a bit and realized that enabled_repository_types() actually calls update_enabled_repository_type() for every single type  (even ones that are already enabled).  

So for 7 content types, and 16 calls to this method through the navidation,  this caused 7x16 = 112 calls to update_enabled_repository_type(), which i think is the expensive part.  

I modified the code to make enabled_repository_types() take an argument to turn off this 'automatic' calling of update_enabled_repository_type(), since find() is doing that already.   And make find() pass in that argument to turn it off.

This then made all the calls go down to a total of ~5 ms.  

One other thing i noticed was that enabled_repository_types() calls update_enabled_repository_type() for all repo types and not just those that are not already enabled.   I don't think there's a reason to do this, you only need to call  update_enabled_repository_type() on those that are disabled

Y

ianballou:
@jlsherrill I've addressed your concerns. I added a new test too.

jlsherrill (approved):
ACK pending tests being green

app/controllers/katello/api/v2/repositories_controller.rb
evgeni: I think you broke nightlies ;)

https://community.theforeman.org/t/katello-nightly-rpm-pipeline-1033-failed/24460/4

PR #9396, opened 2021-06-09
Fixes #32761 - Remove pulp2 call on deb repo content matching

sjha4:


sjha4:
[test katello]

jjeffers (approved):
Reproduced the failure as described in the discussion thread, then applied the PR below. The issue was not seen again and the 2nd publication of the CVV worked.

LGTM

PR #9397, opened 2021-06-09
Fixes #32698 - View component CVs with Added/Not Added filter

sjha4:
To Do:

- [x] Show Environments | Repositories for not added content views using the presenter.
- [x] Add/Remove functionality
- [x] Translations
- [x] Tests

### DIFF
--- a/app/services/katello/component_view_presenter.rb
+++ b/app/services/katello/component_view_presenter.rb
@@ -2,14 +2,26 @@ module Katello
   class ComponentViewPresenter < SimpleDelegator
     attr_accessor :view, :component_view
 
-    def initialize(cv, content_view)
-      @view = content_view
-      @component_view = Katello::ContentViewComponent.where(:composite_content_view_id => cv.id, :content_view_id => @view.id).first_or_initialize
+    def initialize(composite_cv, component_content_view = nil, content_view_component = nil)
+      @view = component_content_view
+      cv_component_record = content_view_component || Katello::ContentViewComponent.where(composite_content_view_id: composite_cv.id, content_view_id: @view.id).first
+      @component_view = cv_component_record || Katello::ContentViewComponent.new(composite_content_view_id: composite_cv.id, content_view_id: @view.id, latest: true)
       super(@component_view)
     end
 
-    def self.component_presenter(cv, views:)
-      views.map { |content_view| ComponentViewPresenter.new(cv, content_view) }
+    def self.component_presenter(composite_cv, status, views:)
+      case status
+      when 'All'
+        views.map { |component_content_view| ComponentViewPresenter.new(composite_cv, component_content_view) }
+      when 'Added'
+        added_cvs = views.map { |component_content_view| Katello::ContentViewComponent.where(composite_content_view_id: composite_cv.id, content_view_id: component_content_view.id).first }
+        added_cvs.compact.map { |content_view| ComponentViewPresenter.new(composite_cv, nil, content_view) }
+      when 'Not added'
+        not_added_cvs = views.reject { |component_content_view| Katello::ContentViewComponent.where(composite_content_view_id: composite_cv.id, content_view_id: component_content_view.id).first }
+        not_added_cvs.map { |component_content_view| ComponentViewPresenter.new(composite_cv, nil, Katello::ContentViewComponent.where(composite_content_view_id: composite_cv.id, content_view_id: component_content_view.id, latest: true).new) }
+      else
+        views.map { |component_content_view| ComponentViewPresenter.new(cv, component_content_view) }
+      end
     end
   end
 end
--- a/webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentContentViewAddModal.js
+++ b/webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentContentViewAddModal.js
@@ -1,0 +1,153 @@
+import React, { useState, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import { Modal, ModalVariant, FormSelect,
+  FormSelectOption, Checkbox, Form, FormGroup,
+  ActionGroup, Button } from '@patternfly/react-core';
+import { useDispatch, useSelector, shallowEqual } from 'react-redux';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { STATUS } from 'foremanReact/constants';
+import { selectCVDetails,
+  selectCVDetailStatus,
+  selectCVDetailError } from '../../Details/ContentViewDetailSelectors';
+import { addComponent } from '../ContentViewDetailActions';
+
+const ComponentContentViewAddModal = ({
+  cvId, componentCvId, componentId, latest, show, setIsOpen,
+}) => {
+  const dispatch = useDispatch();
+  const componentDetails = useSelector(
+    state => selectCVDetails(state, componentCvId),
+    shallowEqual,
+  );
+  const componentStatus = useSelector(
+    state => selectCVDetailStatus(state, componentCvId),
+    shallowEqual,
+  );
+  const componentError = useSelector(
+    state => selectCVDetailError(state, componentCvId),
+    shallowEqual,
+  );
+  const [cvName, setCvName] = useState('');
+  const [options, setOptions] = useState([]);
+  const [formLatest, setFormLatest] = useState(componentId ? latest : false);
+  const [selected, setSelected] = useState(null);
+  const versionsLoading = componentStatus === STATUS.PENDING;
+
+  useEffect(() => {
+    if (!versionsLoading && componentDetails) {
+      const { name, versions } = componentDetails;
+      const versionMutable = versions;
+      setCvName(name);
+      const opt = versionMutable.map(item => ({ value: item.id, label: __(`Version ${item.version}`) }));
+      setOptions([...opt].reverse());
+      setSelected(opt.slice(-1)[0].value);
+    }
+  }, [componentDetails, componentStatus, componentError, versionsLoading]);
+
+  const getAddParams = () => {
+    if (formLatest) {
+      return [{ latest: true, content_view_id: componentCvId }];
+    }
+    return [{ content_view_version_id: selected }];
+  };
+
+  const getUpdateParams = () => {
+    if (formLatest) {
+      return { id: componentId, latest: formLatest, compositeContentViewId: cvId };
+    }
+    return { id: componentId, compositeContentViewId: cvId, content_view_version_id: selected };
+  };
+
+  const updateLatest = (checked) => {
+    setFormLatest(checked);
+    if (checked) setSelected(options[0]);
+  };
+
+  const onSubmit = () => {
+    if (componentId) {
+      dispatch(addComponent({
+        compositeContentViewId: cvId,
+        components: getUpdateParams(),
+      }));
+    } else {
+      dispatch(addComponent({
+        compositeContentViewId: cvId,
+        components: getAddParams(),
+      }));
+    }
+    setIsOpen(false);
+  };
+
+  if (show && versionsLoading) return null;
+
+  return (
+    <Modal
+      title={componentId ? __('Update version') : __('Add component')}
+      variant={ModalVariant.small}
+      isOpen={show}
+      description={__(`Select available version of ${cvName} to use`)}
+      onClose={() => {
+        setIsOpen(false);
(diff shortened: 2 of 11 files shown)


sjha4:
[test katello]

webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentContentViewAddModal.js
```diff
@@ -0,0 +1,136 @@
...
+  useEffect(() => {
+    if (!versionsLoading && componentDetails) {
+      const { name, versions } = componentDetails;
+      const versionMutable = versions;
+      setCvName(name);
+      const opt = versionMutable.map(item => ({ value: item.id, label: `Version ${item.version}` }));
+      setOptions([...opt].reverse());
+    }
+  }, [componentDetails, componentStatus, componentError, versionsLoading]);
+
```
sjha4, line 44: Have to run the dependency array rule against this PR.

sjha4: Ran this..This seems to be the correct array making the rule happy..

sjha4:
[test katello]

app/services/katello/component_view_presenter.rb
```diff
@@ -2,14 +2,26 @@ module Katello
...
     end
 
-    def self.component_presenter(cv, views:)
-      views.map { |content_view| ComponentViewPresenter.new(cv, content_view) }
+    def self.component_presenter(cv, status, views:)
+      return case status
+             when 'All'
+               views.map { |content_view| ComponentViewPresenter.new(cv, content_view) }
+             when 'Added'
+               added_cvs = views.map { |content_view| Katello::ContentViewComponent.where(:composite_content_view_id => cv.id, :content_view_id => content_view.id).first }
```
jeremylenz, line 17: I wonder if @jlsherrill will ever let us enforce modern Ruby hash syntax.. :speak_no_evil:

app/services/katello/component_view_presenter.rb
```diff
@@ -2,14 +2,26 @@ module Katello
...
+      @component_view = component_cv || Katello::ContentViewComponent.where(:composite_content_view_id => cv.id, :content_view_id => @view.id, :latest => true).new
       super(@component_view)
     end
 
-    def self.component_presenter(cv, views:)
-      views.map { |content_view| ComponentViewPresenter.new(cv, content_view) }
+    def self.component_presenter(cv, status, views:)
+      return case status
+             when 'All'
+               views.map { |content_view| ComponentViewPresenter.new(cv, content_view) }
```
jeremylenz, line 15: Could these variables be named to make it more self-explanatory the difference between `cv` and `content_view`? I'm confused..

webpack/scenes/ContentViews/Details/ContentViewDetailActions.js
```diff
@@ -51,6 +57,37 @@ export const updateContentView = (cvId, params) => async dispatch => dispatch(pu
...
+  params: params.components.id ? params.components : params,
+  handleSuccess: () => cvUpdateSuccess({
+    data: { id: params.compositeContentViewId },
+  }, dispatch),
+  handleError: error => dispatch(apiError(null, error)),
+  actionTypes: {
+    REQUEST: ADD_COMPONENT,
+    SUCCESS: ADD_COMPONENT_SUCCESS,
+    FAILURE: ADD_COMPONENT_FAILURE,
+  },
```
jeremylenz, line 73: Specifying action types isn't necessary unless you want to customize the names.  In this case it actually withholds information:

If I comment this out I get action types of `CONTENT_VIEWS_COMPONENTS_5_REQUEST` and `CONTENT_VIEWS_COMPONENTS_5_SUCCESS`.

But with the custom action types it's just `ADD_COMPONENT` and `ADD_COMPONENT_SUCCESS`

Unless we really want to show a different name in the Redux log, I think we should remove the `actionTypes` everywhere.  We should be more clear about that in the usage docs..

webpack/scenes/ContentViews/Details/ContentViewDetailActions.js
```diff
@@ -51,6 +57,37 @@ export const updateContentView = (cvId, params) => async dispatch => dispatch(pu
   },
 }));
 
+export const addComponent = params => async dispatch => dispatch(put({
```
jeremylenz, line 60: `async` is not needed

jeremylenz: Why is this a thunk? So you can have access to `dispatch` inside the success and error handlers?  There must be another way..

jeremylenz (commented):
Thanks @sjha4 

some initial comments below

webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentVersion.js
```diff
@@ -10,33 +10,46 @@ import {
...
+    );
+  }
   return (
     <>
-      <Link to={urlBuilder('labs/content_views', '', id)}> Version {version}</Link>
-      <TextContent>
-        <Text component={TextVariants.small}>{latest ? latestDescription : manualVersionText}</Text>
-      </TextContent>
+      <Link to={urlBuilder('labs/content_views', '', id)}>{latestVersion ? `Version ${latestVersion}` : noVersionText}</Link>
     </>
```
Andrewgdewar, line 39: Remove unneeded fragments

app/services/katello/component_view_presenter.rb
sjha4: Renamed the variables..composite_cv, component_content_view and content_view_component respectively.

The 3rd is the join object between composite and component and the composite/component terminology is driven from the UI but it shouldn't be too confusing now. Although I am sure I'll be confused looking at this in a month 😄

webpack/scenes/ContentViews/Details/ContentViewDetailActions.js
sjha4: Removed..Using a special key and removing these..Let me know how it looks now.

webpack/scenes/ContentViews/Details/ContentViewDetailActions.js
sjha4: Refactored the code to remove the need for the thunk..Added the table refresh as an additionalListener to the Table wrapper..

webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentVersion.js
sjha4: Updated to remove the unnecessary fragments. 👍🏽

Andrewgdewar (approved):
LGTM!

Andrewgdewar:
One question, if a user only has 1 version, why are we showing them the modal popup? 
Is it to allow access to the always use latest version option?

sjha4:
> One question, if a user only has 1 version, why are we showing them the modal popup?
> As selection of that version is redundant.
> Is it to allow access to the 'always use latest version' option?

@Andrewgdewar That's correct. A version 1 without latest flag is different from a Version 1 with the latest flag checked. We associate Version to the composite CV if latest is not checked vs associating content view in case latest is checked and always having the latest version associated by default.  

> If so, we should probably pre-select the only version that is available and disable the drop-down, no?

Feel like the purpose of disabled dropdown is to alert the user that the selection of version doesn't matter. In case of selection of Version 1 without latest flag, it does matter in a sense even though the user has no other options to choose from. Thoughts?

webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentContentViewAddModal.js
```diff
@@ -0,0 +1,148 @@
...
+    state => selectCVDetailStatus(state, componentCvId),
+    shallowEqual,
+  );
+  const componentError = useSelector(
+    state => selectCVDetailError(state, componentCvId),
+    shallowEqual,
+  );
+  const [cvName, setCvName] = useState('');
+  const [options, setOptions] = useState([]);
+  const [useLatest, setUseLatest] = useState(componentId ? latest : false);
```
jeremylenz, line 32: Should probably call this something that doesn't start with `use`, since it's not a custom hook.

webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentContentViewAddModal.js
```diff
@@ -0,0 +1,148 @@
...
+    } else {
+      dispatch(addComponent({
+        compositeContentViewId: cvId,
+        components: getAddParams(),
+      }));
+    }
+    setIsOpen(false);
+  };
+
+  if (show && versionsLoading) return <></>;
```
jeremylenz, line 76: I kinda prefer just returning `null` if you don't want to render anything

jeremylenz (commented):
Getting there! :smile:  The Add modal works as expected now, as does the single remove and bulk remove.

When I click the pencil to change the version selections, the title is still "Add components" even though that's not what I'm doing. Should it be "Select version" if the component is already added?

Likewise, the toast always says "Added component to content view," but it should say something like "Updated component version setting."

When I bulk remove, it says "Removed component" but it should say "Removed components."

I can bulk remove but not bulk add.  Is that the next PR?

jeremylenz:
one more thought I just had: When you check the Automatically update to latest version box, the dropdown is disabled. Which is good. But if you happen to have a non-latest version selected, it implies that it will be set to the old version and yet automatically update, which is not the case.  So, when the checkbox is checked, I think the dropdown should disable _and_ set itself to the latest version.

webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentContentViewAddModal.js
Andrewgdewar: > one more thought I just had: When you check the Automatically update to latest version box, the dropdown is disabled. Which is good. But if you happen to have a non-latest version selected, it implies that it will be set to the old version and yet automatically update, which is not the case. So, when the checkbox is checked, I think the dropdown should disable _and_ set itself to the latest version.

This makes sense to me.

Andrewgdewar:
> > One question, if a user only has 1 version, why are we showing them the modal popup?
> > As selection of that version is redundant.
> > Is it to allow access to the 'always use latest version' option?
> 
> @Andrewgdewar That's correct. A version 1 without latest flag is different from a Version 1 with the latest flag checked. We associate Version to the composite CV if latest is not checked vs associating content view in case latest is checked and always having the latest version associated by default.
> 
> > If so, we should probably pre-select the only version that is available and disable the drop-down, no?
> 
> Feel like the purpose of disabled dropdown is to alert the user that the selection of version doesn't matter. In case of selection of Version 1 without latest flag, it does matter in a sense even though the user has no other options to choose from. Thoughts?

I think than the behaviour in the case of only having one item, is to select it for them, and as mentioned below by @jeremylenz  if the user selects the checkbox, it should also automatically disable/switch to latest version in the dropwdown.

sjha4:
@jeremylenz @Andrewgdewar : Addressed the feedback from above in the last commit.

Andrewgdewar (approved):

webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentContentViewAddModal.js
```diff
@@ -0,0 +1,153 @@
...
+  };
+
+  if (show && versionsLoading) return null;
+
+  return (
+    <Modal
+      title={componentId ? __('Update version') : __('Add component')}
+      variant={ModalVariant.small}
+      isOpen={show}
+      description={__(`Select available version of ${cvName} to add`)}
```
jeremylenz, line 88: Changing the wording here will avoid having to change this text based on if we are adding or updating the version:
```suggestion
      description={__(`Select available version of ${cvName} to use`)}
```

jeremylenz (commented):
Changes look much better, thanks :)

Just one more small change and then I think we are good to go!

webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentContentViewAddModal.js
sjha4: Updated! 👍🏽

jeremylenz (approved):
APJ :+1:

PR #9398, opened 2021-06-09
Fixes #32144 - Support repair repo for all repo types

jlsherrill:


app/lib/actions/pulp3/repository/repair.rb
```diff
@@ -23,6 +23,10 @@ def run_progress_weight
         def presenter
           Presenters::RepairPresenter.new(self)
         end
+
+        def rescue_strategy_for_self
+          Dynflow::Action::Rescue::Skip
+        end
```
jlsherrill, line 29: this is not strictly required, but i noticed it when i reproduced that the task failed and caused a paused task instead of ending

sjha4 (approved):
Able to verify checksum on deb repos now. Rpm repos still work as expected. The option is not available on the UI for any other repo types but I guess it's supported on the backend now. Thoughts on enabling that on the UI?

Ack for this piece. ✔️ 

Screenshot from 2021-06-10 09-28-40


Screenshot from 2021-06-10 09-38-39

sjha4:
@jlsherrill:

The docker repo type strangely fails with error: `Jun 10 15:19:40 centos7-devel3 pulpcore-api: pulp [09531562-532f-4ed3-8da9-8c9a62640453]: django.request:WARNING: Forbidden: /pulp/api/v3/repositories/container/container/85967b32-1347-493a-be90-a4204f3b7468/versions/1/repair/
`

Everything else worked for me. The bulk action now selects all repo types to verify checksum and all repo types have the verify checksum action.

jlsherrill:
i filed: https://pulp.plan.io/issues/8884

sjha4:
Re-Ack on this with the filed issue taking care of the container repair. ✔️

PR #9399, opened 2021-06-10
Fixes #32613 - Update error message on proxy

sjha4:


app/lib/katello/errors.rb
```diff
@@ -53,7 +53,7 @@ def message
 
     class PulpcoreMissingCapabilities < StandardError
       def message
-        _("A smart proxy seems to have been refreshed without pulpcore being running.  You may want to ")
+        _("A smart proxy seems to have been refreshed without pulpcore being running.  You may want to refresh your smart proxy.")
```
jeremylenz, line 56: Will refreshing it again help?  How can the problem and the solution be the same? Do I need to make sure pulpcore is running first?

sjha4: So this is thrown when pulp3 is enabled on the proxy and yet PULP3 as a feature is not found on the proxy. There could be an error on the proxy with the pulp services running that led to this. I believe in most conditions it is caused by a race condition during installation. @jlsherrill can correct me on that.

We could say something like, 'Please refresh the smart proxy after ensuring that pulpcore is running'.

jeremylenz: > Please refresh the smart proxy after ensuring that pulpcore is running.

yes this makes more sense to me!

jlsherrill: yes exactly, the installer couldn't guarantee that pulp was fully up when the smart proxy was being initially refreshed.  We try to detect this situation and run a refresh manually, but this is kinda a last resort suggestion

sjha4: Updated it to say:
`A smart proxy seems to have been refreshed without pulpcore being running. Please refresh the smart proxy after ensuring that pulpcore services are running.`

jeremylenz (approved):

PR #9400, opened 2021-06-11
Fixes #32784 - fix content guard creation

jlsherrill:


sjha4:
Change looks good. Testing now..

sjha4 (approved):
Works well! APJ!

PR #9403, opened 2021-06-11
Fixes #32787 - Ping checks pulpcore api service correctly

parthaa:
The previous check used 'systemctl is-enabled' to figure out. This would
not work well for production systems with selinux content barring
running the systemctl commands.
The change here instead just checks if the file exists.


jlsherrill (approved):

chris1984 (approved):
ACK, works great

PR #9404, opened 2021-06-11
Fixes #32724 - Removes an unnecessary pulp3 ping check

parthaa:
As a part of removing pulp2 references, taking out unnecesary code out
of the ping. Fixeing another issue


parthaa:
[test katello]

chris1984 (approved):
ACK

PR #9406, opened 2021-06-14
Fixes #32712 - Python repo CRUD with generic api/repo classes

rverdile:
Can be tested with hammer commands (create, info, update, delete).

### DIFF
--- a/app/services/katello/pulp3/api/generic.rb
+++ b/app/services/katello/pulp3/api/generic.rb
@@ -1,0 +1,67 @@
+require "pulpcore_client"
+module Katello
+  module Pulp3
+    module Api
+      class Generic < Core
+        attr_accessor :repository_type
+
+        def initialize(smart_proxy, repository_type)
+          @repository_type = repository_type
+          super(smart_proxy)
+        end
+
+        def self.api_exception_class
+          fail NotImplementedError
+        end
+
+        def self.client_module
+          fail NotImplementedError
+        end
+
+        def self.remote_class
+          fail NotImplementedError
+        end
+
+        def self.distribution_class
+          fail NotImplementedError
+        end
+
+        def self.publication_class
+          fail NotImplementedError
+        end
+
+        def self.repository_sync_url_class
+          fail NotImplementedError
+        end
+
+        def self.add_remove_content_class
+          fail NotImplementedError
+        end
+
+        def api_client
+          api_client_class(@repository_type.api_class.new(smart_proxy.pulp3_configuration(@repository_type.configuration_class)))
+        end
+
+        def repositories_api
+          @repository_type.repositories_api_class.new(api_client)
+        end
+
+        def repository_versions_api
+          @repository_type.repository_versions_api_class.new(api_client)
+        end
+
+        def remotes_api
+          @repository_type.remotes_api_class.new(api_client)
+        end
+
+        def publications_api
+          fail NotImplementedError
+        end
+
+        def distributions_api
+          @repository_type.distributions_api_class.new(api_client)
+        end
+      end
+    end
+  end
+end
--- a/app/services/katello/pulp3/repository/generic.rb
+++ b/app/services/katello/pulp3/repository/generic.rb
@@ -1,0 +1,36 @@
+module Katello
+  module Pulp3
+    class Repository
+      class Generic < ::Katello::Pulp3::Repository
+        def copy_content_for_source(source_repository, _options = {})
+          copy_units_by_href(source_repository.files.pluck(:pulp_id))
+        end
+
+        def distribution_options(path)
+          {
+            base_path: path,
+            publication: repo.publication_href,
+            name: "#{generate_backend_object_name}"
+          }
+        end
+
+        def remote_options
+          common_remote_options
+        end
+
+        def partial_repo_path
+          repo.repository_type.partial_repo_path
+        end
+
+        def api
+          @api ||= repo.repository_type.pulp3_api(smart_proxy)
+        end
+
+        def self.api(smart_proxy, repo)
+          api_class = RepositoryTypeManager.find_by(:pulp3_service_class, self).pulp3_api_class
+          api_class ? api_class.new(smart_proxy, repo) : Katello::Pulp3::Api::Core.new(smart_proxy)
+        end
+      end
+    end
+  end
+end
--- a/katello.gemspec
+++ b/katello.gemspec
@@ -59,6 +59,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency "pulp_deb_client", ">= 2.10.0", "< 2.12.0"
   gem.add_dependency "pulp_rpm_client", ">=3.10.0", "< 3.11.0"
   gem.add_dependency "pulp_certguard_client", "< 2.0"
+  gem.add_dependency "pulp_python_client", ">= 3.2.0"
 
   # UI
   gem.add_dependency "deface", '>= 1.0.2', '< 2.0.0'
--- a/lib/katello/repository_types/python.rb
+++ b/lib/katello/repository_types/python.rb
@@ -1,0 +1,16 @@
+require 'pulp_python_client'
+
+Katello::RepositoryTypeManager.register('python') do
+  allow_creation_by_user true
+  pulp3_service_class Katello::Pulp3::Repository::Generic
+  pulp3_api_class Katello::Pulp3::Api::Generic
+  pulp3_plugin 'python'
+  partial_repo_path '' #TODO: add partial repo path
+
+  repositories_api_class PulpPythonClient::RepositoriesPythonApi
+  api_class PulpPythonClient::ApiClient
+  configuration_class PulpPythonClient::Configuration
+  remotes_api_class PulpPythonClient::RemotesPythonApi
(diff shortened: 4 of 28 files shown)


app/services/katello/pulp3/repository/generic.rb
```diff
@@ -0,0 +1,32 @@
...
+
+        def distribution_options(path)
+          {
+            base_path: path,
+            publication: repo.publication_href,
+            name: "#{generate_backend_object_name}"
+          }
+        end
+
+        def remote_options
```
ianballou, line 17: `remote_options` and `partial_repo_path` will need to be made generic as well. Right now it's specific to the File repo type. I'd say for this PR, `remote_options` can just return `common_remote_options` and `partial_repo_path` could be defined in `lib/katello/repository_types/python.rb`.  You could probably just leave it as an empty string until you do the syncing stuff.

app/services/katello/pulp3/api/generic.rb
```diff
@@ -0,0 +1,61 @@
...
+        def self.api_exception_class
+          fail NotImplementedError
+        end
+
+        def self.client_module
+          fail NotImplementedError
+        end
+
+        def self.remote_class
+          fail NotImplementedError
```
ianballou, line 16: Good idea planning these out early.

lib/katello/repository_types/python.rb
```diff
@@ -0,0 +1,17 @@
...
+Katello::RepositoryTypeManager.register('python') do
+  allow_creation_by_user true
+  pulp3_service_class Katello::Pulp3::Repository::Generic
+  pulp3_api_class Katello::Pulp3::Api::Generic
+  pulp3_plugin 'python'
+  repositories_class PulpPythonClient::RepositoriesPythonApi
+  api_class PulpPythonClient::ApiClient
+  configuration_class PulpPythonClient::Configuration
+
+  content_type Katello::FileUnit,
```
ianballou, line 11: `content_type` and `default_managed_content_type` will need to be made generic as well. For now you can leave it all empty instead of having the file units be there.  It shouldn't cause issues with repo CRUD.

katello.gemspec
```diff
@@ -59,6 +59,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency "pulp_deb_client", ">= 2.10.0", "< 2.12.0"
   gem.add_dependency "pulp_rpm_client", ">=3.10.0", "< 3.11.0"
   gem.add_dependency "pulp_certguard_client", "< 2.0"
+  gem.add_dependency "pulp_python_client", ">= 3.2.0"
```
ianballou, line 62: We'll need to either wait to add this or get `pulp_python_client` into foreman-packaging before merging.

ianballou (changes requested):
Looking good! A few comments:

lib/katello/repository_types/python.rb
rverdile: For these guys, an object is a required argument. Is there a way of leaving them empty besides nil or excluding them? I left the File implementation there because it doesn't cause issues and I didn't think a generic version fits within the context of the card.

katello.gemspec
ianballou: Related: https://github.com/theforeman/foreman-packaging/pull/6790

lib/katello/repository_types/python.rb
ianballou: @rverdile you could try removing `content_type` and `default_managed_content_type`.

katello.gemspec
rverdile: @ianballou if that gets merged into foreman, would it make sense to remove this? https://github.com/Katello/katello/blob/89830530de2f46ab6c88f99807b1fe54efa1901a/lib/katello/repository_types/python.rb#L1

ianballou: That PR only has to do with installing production versions of Katello.  You'll still need to require the client somewhere in the code.

app/services/katello/pulp3/api/core.rb
```diff
@@ -6,8 +6,9 @@ module Api
       class Core
         attr_accessor :smart_proxy
 
-        def initialize(smart_proxy)
+        def initialize(smart_proxy, repo = nil)
```
ianballou, line 9: What do you think about overriding the initializer for core in generic.rb and making the repository argument required?

So then initialize in generic.rb could look like:
```ruby
def initialize(smart_proxy, repository)
  @repository = repository
  super(smart_proxy)
end
```

ianballou: I'm suggesting that since it would be good to not modify core unless we really have to. Plus, it'll make it more apparent that the Generic API needs to have the repo passed in.

app/services/katello/pulp3/repository.rb
```diff
@@ -28,17 +28,17 @@ def with_mirror_adapter
         end
       end
 
-      def self.api(smart_proxy)
+      def self.api(smart_proxy, repo = nil)
```
ianballou, line 31: Similarly to my comment above, I think you could override `self.api` (and related methods) in app/services/katello/pulp3/repository/generic.rb so that repository.rb doesn't have to be edited

test/services/katello/pulp3/repository/python/python_test.rb
```diff
@@ -0,0 +1,47 @@
...
+              root_repository_id: @repo.root.id,
+              content_view_id: @repo.content_view.id)
+
+            assert_nil repo_reference
+          end
+        end
+      end
+    end
+  end
+end
```
jlsherrill, line 47: Something i've kinda thought about is to create a test framework to make fully testing a generic repo type (or maybe even other types) quite a bit simpler.  

I imagine it would be something like:

```
Class MytestClass   
   include GenericRepoTesting
    
    test_generic_type  'python'
    test_sync 'https://someserverurl/', :other_repository_option => 'somevalue'
    test_upload './path/to/sample/file/to/upload'
end
```

lets not do it in this Pr, but maybe something to add to work on in the future :)

jlsherrill: also keep in mind 'include GenericRepoTesting' here has all the magic

lib/katello/repository_types/python.rb
```diff
@@ -0,0 +1,15 @@
+require 'pulp_python_client'
+
+Katello::RepositoryTypeManager.register('python') do
+  allow_creation_by_user true
+  pulp3_service_class Katello::Pulp3::Repository::Generic
+  pulp3_api_class Katello::Pulp3::Api::Generic
+  pulp3_plugin 'python'
+  partial_repo_path ''
```
ianballou, line 8: You might want to just put a `# TODO: ...` here since we'll need to decide on the partial_repo_path (and if it should be defined here) later.

ianballou:
[test katello]

rverdile:
Jenkins is failing because of situations like this https://github.com/Katello/katello/blob/47a628e4a35c95f49bd389674ac7987258ec91eb/app/services/katello/pulp3/smart_proxy_repository.rb#L57


python is pulp3 enabled, but Generic overrides self.api to require two args causing `api = repo_type.pulp3_service_class.api(smart_proxy)` to fail. Not sure if I should just use a conditional to avoid this for now, or try to integrate Generic into this?

ianballou:
> Jenkins is failing because of situations like this
> 
> https://github.com/Katello/katello/blob/47a628e4a35c95f49bd389674ac7987258ec91eb/app/services/katello/pulp3/smart_proxy_repository.rb#L57
> 
> python is pulp3 enabled, but Generic overrides self.api to require two args causing `api = repo_type.pulp3_service_class.api(smart_proxy)` to fail. Not sure if I should just use a conditional to avoid this for now, or try to integrate Generic into this?

I think we might just need a conditional to check if the repo type is generic or not and then create the `api` as such. @jlsherrill do you have any thoughts about that?

test/services/katello/pulp3/repository_list_test.rb
```diff
@@ -1,6 +1,5 @@
 require 'katello_test_helper'
 require 'support/pulp3_support'
-
```
ianballou, line 3: Line

lib/katello/repository_types/python.rb
```diff
@@ -0,0 +1,16 @@
+require 'pulp_python_client'
+
+Katello::RepositoryTypeManager.register('python') do
+  allow_creation_by_user true
+  pulp3_service_class Katello::Pulp3::Repository::Generic
+  pulp3_api_class Katello::Pulp3::Api::Generic
+  pulp3_plugin 'python'
+  partial_repo_path '' #TODO: add partial repo path
+
+  repositories_class PulpPythonClient::RepositoriesPythonApi
```
ianballou, line 10: I realized in some cases you have `api` in the variable name and sometimes you don't. I think it would be best to stick to a single convention (in this case probably keeping `api` in there).

ianballou: Edit: we should put API when the Pulp class has API in it.

ianballou:
I think all of my issues have been addressed!

jlsherrill (approved):
ACK from me too!  One of us can merge once its 🟢

PR #9407, opened 2021-06-14
Fixes #32800 - allow Repo delete for RH repos

jlsherrill:


chris1984 (commented):
Angular looks good, will start to test. Since angular is going away, I don't see the need for any new test coverage, thanks for updating the existing one.

chris1984 (approved):
ACK, works great. I was able to delete a RH repo from the UI

PR #9413, opened 2021-06-18
Fixes #32835 - Add RHSM and content URL as info providers

ekohl:
This introduces two Smart Proxy extension methods and exposes them as info providers. This allows clients to use the ENC API to (re)configure subscription-manager.

First of all, this implements logic to determine the RHSM URL. This is constructed by looking at Pulp. While this is really the wrong way, there is currently no alternative. (I started https://github.com/ekohl/smart_proxy_rhsm a while back and that could solve it - another is to bolt it onto the Pulp module).

Ideally this would be exposed as a setting as well. There is a plan to align everything on port 443. If the Smart Proxy exposes a setting, Katello doesn't need to guess and a smooth migration can be realized.

This all becomes more important when the host subscription RPM is phased out.

Then it also introduces pulp_content_url which is already a setting that's exposed via the Smart Proxy Pulp feature. This is also used in the various places where it's needed.

It does highlight a missing setting for pulp_ansible content. Again, Katello should not have to guess URLs but discover it from the infrastructure.

Another potential use case that this allows is using REX to move a host from one content source to another.

Currently a draft since I have not verified this at all. However, I did want to link it and share it. What's also probably missing is provisioning macros.

### DIFF
--- a/app/models/katello/host/info_provider.rb
+++ b/app/models/katello/host/info_provider.rb
@@ -16,6 +16,15 @@ module Katello
         if host.content_facet.present?
           info['parameters']['kickstart_repository'] = host.content_facet.kickstart_repository.try(:label)
         end
+
+        if (rhsm_url = host.content_source&.rhsm_url)
+          info['parameters']['rhsm_url'] = rhsm_url.to_s
+        end
+
+        if (content_url = host.content_source&.pulp_content_url)
+          info['parameters']['content_url'] = content_url.to_s
+        end
+
         info
       end
 


app/lib/katello/resources/registry.rb
```diff
@@ -28,7 +28,7 @@ def load_class
 
             # Pulp 3 has its own registry
             if pulp_primary&.pulp3_repository_type_support?(::Katello::Repository::DOCKER_TYPE)
-              uri = URI(pulp_primary.setting(SmartProxy::PULP3_FEATURE, 'content_app_url'))
+              uri = URI(pulp_primary.pulp_content_url)
               uri.path = "/pulpcore_registry/"
```
ekohl, line 32: I think we should create a separate setting for this in smart_proxy_pulp

app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -118,11 +118,25 @@ def container_gateway_users
         ::User.where(login: usernames['users'])
       end
 
+      def rhsm_url
+        # TODO: We shouldn't relate this to the Pulp feature and also expose it
```
ekohl, line 122: https://github.com/ekohl/smart_proxy_rhsm would do this. However, given there's a `content_source` I think we should also consider a single `smart_proxy_content` module. That will depend on if we ever want to allow splitting the concern.

app/models/katello/repository.rb
```diff
@@ -400,17 +400,21 @@ def group
...
+      pulp_uri.host = pulp_uri.host.downcase
       if docker?
-        "#{pulp_uri.host.downcase}/#{container_repository_name}"
+        "#{pulp_uri.host}/#{container_repository_name}"
       elsif ostree?
-        "#{scheme}://#{pulp_uri.host.downcase}/pulp/content/web/#{relative_path}"
+        pulp_uri.path += "/web/#{relative_path}"
       elsif ansible_collection?
-        "#{scheme}://#{pulp_uri.host.downcase}/pulp_ansible/galaxy/#{relative_path}/api"
+        pulp_uri.path = "/pulp_ansible/galaxy/#{relative_path}/api"
```
ekohl, line 413: This also needs to be exposed as a setting by smart_proxy_pulp

ekohl (commented):
I'm not sure if I can really finish this, but this is IMHO important for our enterprise readiness meetings that we've been having. There's also some features that are starting to depend on this information being present (and correct).
* Host registration
* Removal of the registration RPM
* Aligning RHSM on port 443 for all services

ekohl:
Related PR: https://github.com/Katello/katello/pull/9417

app/lib/katello/resources/registry.rb
jturel: You mean to separate the content app url and the registry url?

ekohl: Today we [expose the content_app_url](https://github.com/theforeman/smart_proxy_pulp/blob/1d4cc85d6b89e292cf613bf0b81ad24e785d4870/lib/smart_proxy_pulp_plugin/pulpcore_plugin.rb#L20). This means Katello doesn't have to guess. I think it would be good if we also added an exposed setting there that is the pulpcore registry URL. We can do that by calculating this value in the Smart Proxy plugin or explicitly setting it in YAML, but Katello shouldn't be responsible for guessing URLs.

Does that clarify it?

jturel: Oh I get that part. I was asking about whether your comment ultimately meant to not derive the registry URL from "pulp_content_url"

ekohl: Probably. The installer already knows the URL so we can set it in the YAML settings.

ekohl: https://github.com/theforeman/smart_proxy_pulp/pull/30 would be that.

ekohl:
That has been merged. I'll rebase this on https://github.com/Katello/katello/pull/9450 once that's been merged.

jturel:
9450 is merged! please drop this out of Draft state when it's ready for review :)

jturel:
@ekohl this is on our board as Ready for Review but I'm gonna move it to our backlog until it's out of draft state. Do you intend to have this ready for 4.2?

ekohl:
I'd love to, but I can't promise anything. Recently https://github.com/theforeman/smart_proxy_pulp/commit/b6582ef999baed26f8e01ed728eaa4b5be5ce2df was introduced in smart_proxy_pulp so at least the prerequisites are merged. However, my current priority is dealing with the Puppet extraction. I'd appreciate it if someone else could take over.

cc @wbclark

jturel:
Should this be closed / backlogged until someone can address it or it's explicitly required?

ekohl:
Given a lot of the related work has been merged already, this is now actually quite trivial. I must admit I haven't tested it since I don't have a development environment, but it should be quite easy. It's exposed in `/api/hosts/:id/enc`. If a content source is assigned to it, the URLs should be present.

The other (unrelated work) will be submitted as a separate PR.

ekohl:
> The other (unrelated work) will be submitted as a separate PR.

https://github.com/Katello/katello/pull/9878

jturel:
Found this while testing:

```
16:15:42 rails.1   | 2022-01-06T16:15:42 [I|app|b8c7adba] Started GET "/api/v2/hosts/4/enc" for 127.0.0.1 at 2022-01-06 16:15:42 +0000
16:15:42 rails.1   | 2022-01-06T16:15:42 [I|app|b8c7adba] Processing by Api::V2::HostsController#enc as JSON
16:15:42 rails.1   | 2022-01-06T16:15:42 [I|app|b8c7adba]   Parameters: {"apiv"=>"v2", "id"=>"4"}
16:15:42 rails.1   | 2022-01-06T16:15:42 [I|app|b8c7adba] Authorized user admin(Admin User)
16:15:42 rails.1   | 2022-01-06T16:15:42 [W|app|b8c7adba] Action failed
16:15:42 rails.1   | 2022-01-06T16:15:42 [I|app|b8c7adba] Backtrace for 'Action failed' error (ActiveModel::MissingAttributeError): can't write unknown attribute `rhsm_url`
16:15:42 rails.1   |  b8c7adba | /home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activemodel-6.0.3.7/lib/active_model/attribute.rb:206:in `with_value_from_database'
16:15:42 rails.1   |  b8c7adba | /home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activemodel-6.0.3.7/lib/active_model/attribute_set.rb:49:in `write_from_user'
16:15:42 rails.1   |  b8c7adba | /home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activerecord-6.0.3.7/lib/active_record/attribute_methods/write.rb:43:in `_write_attribute'
16:15:42 rails.1   |  b8c7adba | /home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activerecord-6.0.3.7/lib/active_record/attribute_methods/write.rb:36:in `write_attribute'
16:15:42 rails.1   |  b8c7adba | /home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activerecord-6.0.3.7/lib/active_record/attribute_methods.rb:337:in `[]='
16:15:42 rails.1   |  b8c7adba | /home/vagrant/katello/app/models/katello/host/info_provider.rb:21:in `host_info'
16:15:42 rails.1   |  b8c7adba | /home/vagrant/foreman/app/models/concerns/host_info_extensions.rb:27:in `block in info'
16:15:42 rails.1   |  b8c7adba | /home/vagrant/foreman/app/models/concerns/host_info_extensions.rb:25:in `each'
16:15:42 rails.1   |  b8c7adba | /home/vagrant/foreman/app/models/concerns/host_info_extensions.rb:25:in `info'
16:15:42 rails.1   |  b8c7adba | /home/vagrant/foreman/app/controllers/api/v2/hosts_controller.rb:182:in `enc'

```

ekohl:
Oh that's my bad. They should be parameters rather than top level. Updated now.

app/models/katello/host/info_provider.rb
```diff
@@ -16,6 +16,15 @@ def host_info
         if host.content_facet.present?
           info['parameters']['kickstart_repository'] = host.content_facet.kickstart_repository.try(:label)
         end
+
+        if (rhsm_url = host.content_source&.rhsm_url)
+          host['parameters']['rhsm_url'] = rhsm_url.to_s
+        end
+
+        if (content_url = host.content_source&.pulp_content_url)
+          host['parameters']['content_url'] = content_url.to_s
```
jturel, line 25: s/host/info ?

ekohl: I really suffer from not having a proper dev env running. My bad :(

jturel: No worries! For awareness, forklift has a centosX-katello-devel-stable box that is built nightly and published to vagrant cloud. It spins up as fast as you can download the image!

app/models/katello/host/info_provider.rb
```diff
@@ -16,6 +16,15 @@ def host_info
         if host.content_facet.present?
           info['parameters']['kickstart_repository'] = host.content_facet.kickstart_repository.try(:label)
         end
+
+        if (rhsm_url = host.content_source&.rhsm_url)
```
jturel, line 20: I learned that simply registering a content host directly to the server does not establish a content source and so these won't be set. Is that compatible with how you see this being used, or should they get the values from the default smart proxy at that point? Or is it a bug that content_source is nil in that scenario?

ekohl: I would consider that a bug but I'm not sure how others see it. IMHO every host that has content should have a `content_source` assigned. Otherwise you always need to remember to do `host.content_source || pulp3_primary`.

It also means that https://github.com/Katello/katello/pull/9780 may be based on an incorrect assumption and currently can't be used to reconfigure if the host is assigned to the primary.

jturel: #9780 works OK. I can create an issue to track that unless you want to dive into it on this PR

jturel:
@ekohl I don't think this needs to be a draft any longer!

ekohl:
Then I'd like it if this was merged.

jturel:
Just realized the katello tests failed but can't see the build any longer. running again.

[test katello]

jturel (approved):

PR #9415, opened 2021-06-18
Fixes #32815 - Display checkbox/select column for grouped subscriptions

lfu:
Fix an issue with disconnected Satellite where checkbox/select column is not displayed even with grouped subscriptions.

### DIFF
--- a/webpack/components/pf3Table/formatters/selectionHeaderCellFormatter.js
+++ b/webpack/components/pf3Table/formatters/selectionHeaderCellFormatter.js
@@ -1,10 +1,11 @@
 import React from 'react';
 import TableSelectionHeaderCell from '../components/TableSelectionHeaderCell';
 
-export default (selectionController, label) => (
+export default (selectionController, label, selectionEnabled) => (
   <TableSelectionHeaderCell
     label={label}
     checked={selectionController.allRowsSelected()}
     onChange={() => selectionController.selectAllRows()}
+    disabled={!selectionEnabled}
   />
 );
--- a/webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTableSchema.js
+++ b/webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTableSchema.js
@@ -30,13 +30,15 @@ export const createSubscriptionsTableSchema = (
     property: 'select',
     header: {
       label: __('Select all rows'),
-      formatters: [label => selectionHeaderCellFormatter(selectionController, label)],
+      formatters: [
+        label => selectionHeaderCellFormatter(selectionController, label, hasPermission),
+      ],
     },
     cell: {
       formatters: [
         (value, additionalData) => {
           // eslint-disable-next-line no-param-reassign
-          additionalData.disabled = additionalData.rowData.available === -1;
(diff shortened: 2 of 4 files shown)


webpack/scenes/Subscriptions/components/SubscriptionsTable/SubscriptionsTableSchema.js
```diff
@@ -30,13 +30,13 @@ export const createSubscriptionsTableSchema = (
     property: 'select',
     header: {
       label: __('Select all rows'),
```
jturel, line 32: I don't think we should carry the `hasPermission` variable name into this part of the code since it's not related to permissions at all at this point. Would `selectionEnabled` be a better name?

jturel:
I'm leaving some feedback here but would like someone else to own the review since I had a small hand in the solution :)

webpack/components/pf3Table/formatters/selectionHeaderCellFormatter.js
```diff
@@ -1,10 +1,11 @@
 import React from 'react';
 import TableSelectionHeaderCell from '../components/TableSelectionHeaderCell';
 
-export default (selectionController, label) => (
+export default (selectionController, label, hasPermission) => (
```
jturel, line 4: I don't think we should carry the hasPermission variable name into this part of the code since it's not related to permissions at all at this point. Would selectionEnabled be a better name?

webpack/scenes/Subscriptions/components/SubscriptionsTable/components/Table.js
```diff
@@ -34,9 +34,12 @@ const Table = ({
     toggle: ({ rowData }) => toggleSubscriptionGroup(rowData.product_id),
   };
 
+  const collapsibleGroup = () =>
```
jturel, line 37: I think the `some` function (returns a boolean) might be advantageous here: https://www.w3schools.com/jsref/jsref_some.asp

`const collapsibleGroup = Object.values(groupedSubscriptions).some(v => v.subscriptions.length > 1);`

webpack/scenes/Subscriptions/components/SubscriptionsTable/components/Table.js
```diff
@@ -34,9 +34,11 @@ const Table = ({
     toggle: ({ rowData }) => toggleSubscriptionGroup(rowData.product_id),
   };
 
+  const collapsibleGroup = Object.values(groupedSubscriptions).some(v => v.subscriptions.length > 1);
```
jeremylenz, line 37: Looks like there's one eslint issue here.  After fixing it you can confirm by running eslint locally from the katello directory like this:
```bash
npx eslint webpack/
```

jeremylenz (approved):
Congrats on your first React contribution! :smile:  Side note: I know it's a pain sifting through all the formatters in the Patternfly 3 table components.  The need for formatters will be eliminated with the new [TableComposable](https://patternfly-react.surge.sh/components/table#composable-basic) component, which @sjha4 is already using. So this should be getting easier eventually :)

Tested with and without disconnected mode, works as expected.

Just one eslint issue below, other than that APJ :+1:

jturel (approved):
APJ

PR #9416, opened 2021-06-18
Fixes #32839 - Make with_ansible? check more precise

jturel:
A recent foreman change re-used the `ForemanAnsible` namespace previously used only by the foreman_ansible plugin. The presence of the `ForemanAnsible` module threw a false positive for our `with_ansible?`. The fix here is to check specifically for the foreman_ansible engine itself.

See: https://github.com/theforeman/foreman/pull/8491

### DIFF
--- a/lib/katello/engine.rb
+++ b/lib/katello/engine.rb
@@ -260,6 +260,6 @@ module Katello
   end
 
   def self.with_ansible?
-    (ForemanAnsible rescue false) ? true : false
+    (ForemanAnsible::Engine rescue false) ? true : false
   end
 end
--- a/lib/katello/engine.rb
+++ b/lib/katello/engine.rb
@@ -248,7 +248,7 @@ module Katello
 
   # check whether foreman_remote_execution to integrate is available in the system
   def self.with_remote_execution?
-    (RemoteExecutionFeature rescue false) ? true : false
+    Foreman::Plugin.installed?("foreman_remote_execution")
   end
 
   def self.with_katello_agent?
@@ -260,6 +260,6 @@ module Katello
   end
 
   def self.with_ansible?
-    (ForemanAnsible::Engine rescue false) ? true : false
+    Foreman::Plugin.installed?("foreman_ansible")
   end
 end


ehelms:
First, yuck on Foreman core for re-using a namespace of a long lived plugin.
Second, should we be more precise here and checking for the presence of the plugin in the plugin registry? This shouldn't happen again, but we do a have a plugin registry that ought to be usable for this purpose.

jturel:
> First, yuck on Foreman core for re-using a namespace of a long lived plugin.
> Second, should we be more precise here and checking for the presence of the plugin in the plugin registry? This shouldn't happen again, but we do a have a plugin registry that ought to be usable for this purpose.

Maybe it wasn't used in the first place for a reason? Updated. 

BONUS: also updated `with_remote_execution?` !!!

jturel:
If you're wondering whether these are the valid plugin names per the registry:
```
irb(main):002:0> Foreman::Plugin.installed?("foreman_ansible")
=> true
irb(main):005:0> Foreman::Plugin.installed?("foreman_remote_execution")
=> true
```

jeremylenz (approved):
APJ

ehelms:
> Maybe it wasn't used in the first place for a reason? Updated.

You could be quite right and there may be dragons in a production setup. I should have codified more heavily my *ought to be usable for this purpose.*. I think we should take this change for a test drive in the devel scenario and in a production scenario. I can try out the latter via a scratch build.

ekohl (commented):
> A recent foreman change re-used the `ForemanAnsible` namespace previously used only by the foreman_ansible plugin. The presence of the `ForemanAnsible` module threw a false positive for our `with_ansible?`. The fix here is to check specifically for the foreman_ansible engine itself.

For what it's worth: that's really intended to be temporary. I only allowed it to be merged with the promise that it would be refactored.

That said, using `Foreman::Plugin.installed?` is IMHO the correct way.

jturel:
@ehelms you still attempting to test this in prod or should we just merge and see?

parthaa (approved):
ACK worked for me

PR #9417, opened 2021-06-21
Fixes #32841 - Host registration - :rhsm_base_url helper

stejskalleos:
For https://github.com/theforeman/foreman/pull/8563

### DIFF
--- a/app/controllers/katello/concerns/api/v2/registration_controller_extensions.rb
+++ b/app/controllers/katello/concerns/api/v2/registration_controller_extensions.rb
@@ -24,6 +24,31 @@ module Katello
 
         super
       end
+
+      def context_urls
+        super.merge(rhsm_url: rhsm_url, pulp_content_url: pulp_content_url)
+      end
+
+      private
+
+      def smart_proxy
+        @smart_proxy ||= begin
+          proxy = params[:url] ? SmartProxy.find_by(url: params[:url]) : SmartProxy.pulp_primary
+
+          fail Foreman::Exception, N_('Smart proxy content source not found!') unless proxy
+          fail Foreman::Exception, N_('Pulp 3 is not enabled on Smart proxy!') unless proxy.pulp3_enabled?
+
+          proxy
+        end
+      end
+
+      def rhsm_url
+        URI(smart_proxy.rhsm_url)
+      end
+
+      def pulp_content_url
+        smart_proxy.setting(SmartProxy::PULP3_FEATURE, 'content_app_url')
+      end
     end
   end
 end
--- a/app/models/katello/concerns/smart_proxy_extensions.rb
+++ b/app/models/katello/concerns/smart_proxy_extensions.rb
@@ -401,6 +401,14 @@ module Katello
       def smart_proxy_service
         @smart_proxy_service ||= Pulp::SmartProxyRepository.new(self)
       end
+
+      def rhsm_url
+        if pulp_primary?
+          "https://#{URI.parse(url).host}/rhsm"
+        elsif pulp_mirror?
+          "https://#{URI.parse(url).host}:8443/rhsm"
+        end
+      end
     end
   end
 end
--- a/test/models/concerns/smart_proxy_extensions_test.rb
+++ b/test/models/concerns/smart_proxy_extensions_test.rb
@@ -58,6 +58,15 @@ module Katello
       assert_equal @proxy.lifecycle_environments.all, Katello::KTEnvironment.all
       assert_equal @proxy_mirror.lifecycle_environments.all.count, 0
     end
+
+    def test_rhsm_url_pulp_primary
+      assert_includes @proxy.rhsm_url, "/rhsm"
+      assert_not_includes @proxy.rhsm_url, ":8443"
+    end
+
+    def test_rhsm_url_pulp_mirror
+      assert_includes @proxy_mirror.rhsm_url, ":8443/rhsm"
+    end
   end
 
   class SmartProxyPulp3Test < ActiveSupport::TestCase


app/controllers/katello/concerns/api/v2/registration_controller_extensions.rb
```diff
@@ -24,6 +24,25 @@ def host_setup_extension
...
+
+      def rhsm_base_url
+        smart_proxy = if params[:url]
+          SmartProxy.find_by(url: params[:url])
+        else
+          SmartProxy.pulp_primary
+        end
+
+        fail Foreman::Exception.new('Smart proxy content source not found!') unless smart_proxy
+        URI.join(smart_proxy.url, 'pulp/content')
```
ekohl, line 44: I'd expect RHSM on `/rhsm`

ekohl (changes requested):
Oh, another complication: the RHSM port is 8443 on Pulp mirrors. I really think https://github.com/Katello/katello/pull/9413 (which implements it as a function on the Smart Proxy) is better. For reference, the relevant bit:
```
      def rhsm_url
        # TODO: We shouldn't relate this to the Pulp feature and also expose it
        # as a setting
        if pulp_mirror?
          "https://#{URI.parse(url).host}:8443/rhsm"
        elsif pulp_primary?
          "https://#{URI.parse(url).host}/rhsm"
        end
      end
```

app/controllers/katello/concerns/api/v2/registration_controller_extensions.rb
stejskalleos: When I install subscription manager with `katello_ca_consumer.rpm` the `/etc/rhsm/rhsm.conf` config file is set like this:
```
[rhsm]
# Content base URL:
baseurl = https://centos7-katello-devel-stable.example.com/pulp/content/
```

So I thought that it should be same

ekohl: That's the (Pulp) content URL, not RHSM itself. I'd consider the RHSM URL the basis for `subscription-manager config --server.hostname "$HOSTNAME" --server.port "$PORT" --server.prefix "$PREFIX"`.

Also, RHSM is always HTTPS while the Smart Proxy URL may be HTTP.

stejskalleos: Hmm, I'm getting super confused from all of the URLs

Right now there are:
* Foreman URL
* Smart Proxy URL
* Primary Pulp content URL
* Mirror pulp content URL
* RHSM URL

If I want to configure subscription-manager, what URLs should I use for these options:

```
subscription-manager config \
      --server.hostname="" \
      --server.port="" \
      --server.prefix="/rhsm" \
      --rhsm.baseurl=""
```

ekohl: It is my understanding (based on https://github.com/theforeman/puppet-foreman_proxy_content/blob/master/lib/puppet/provider/rhsm_reconfigure_script/rhsm_reconfigure_script.rb) that `--server.*` is the RHSM URL. `--rhsm.baseurl` is the Pulp content URL.

ehelms: Correct.

`server.hostname`: is the hostname subscription-manager should communicate with when making API calls
`server.port`: is the port subscription-manager should communicate with when making API calls
`server.prefix`: is a URL prefix that subscription-manager should talk to for all its API communication back to the server, in our case that is always `/rhsm`
`rhsm.baseurl`: is the location at which all content is located, this is prepended to all content paths when constructing the `redhat.repo` file

For example, if Candlepin knows a repo path is `/Default_Organization/Library/Mv_CV/repo_a/` and the `rhsm.baseurl` is `https://mycapsule.com/pulp/content` then the redhat.repo will get configured with:

```
baseurl=https://mycapsule.com/pulp/content/Default_Organization/Library/My_CV/repo_a
```

app/controllers/katello/concerns/api/v2/registration_controller_extensions.rb
```diff
@@ -24,6 +24,25 @@ def host_setup_extension
...
+        urls = super
+        urls[:rhsm_base_url] = rhsm_base_url
+        urls
+      end
+
+      private
+
+      def rhsm_base_url
+        smart_proxy = if params[:url]
+          SmartProxy.find_by(url: params[:url])
```
ehelms, line 38: If we are being technically correct, we should ensure that the found smart proxy has the Pulp 3 feature and throw an error if not.

app/controllers/katello/concerns/api/v2/registration_controller_extensions.rb
```diff
@@ -24,6 +24,32 @@ def host_setup_extension
...
+        fail Foreman::Exception, N_('Smart proxy content source not found!') unless @smart_proxy
+        fail Foreman::Exception, N_('Pulp 3 is not enabled on Smart proxy!') unless @smart_proxy.pulp3_enabled?
+      end
+
+      def rhsm_url
+        URI(@smart_proxy.rhsm_url)
+      end
+
+      def pulp_content_url
+        URI.join(@smart_proxy.url, '/pulp/content')
```
ekohl, line 51:
```suggestion
        @smart_proxy.setting(SmartProxy::PULP3_FEATURE, 'content_app_url')
```

ekohl (changes requested):

jturel (commented):
Left two comments on code style but testing in conjunction with the Foreman PR works well. Here's the resulting bits from my rhsm.conf:

```
hostname = centos7-katello-devel.jturel.example.com
prefix = /rhsm
port = 443
[rhsm]
# Content base URL:
baseurl = https://centos7-katello-devel.jturel.example.com/pulp/content

```

app/controllers/katello/concerns/api/v2/registration_controller_extensions.rb
```diff
@@ -24,6 +24,29 @@ def host_setup_extension
...
+        super.merge(rhsm_url: rhsm_url, pulp_content_url: pulp_content_url)
+      end
+
+      private
+
+      def find_smart_proxy
+        @smart_proxy ||= params[:url] ? SmartProxy.find_by(url: params[:url]) : SmartProxy.pulp_primary
+
+        fail Foreman::Exception, N_('Smart proxy content source not found!') unless @smart_proxy
+        fail Foreman::Exception, N_('Pulp 3 is not enabled on Smart proxy!') unless @smart_proxy.pulp3_enabled?
```
ekohl, line 40: A common pattern for this is
```suggestion
      def smart_proxy
        @smart_proxy ||= begin
          proxy = params[:url] ? SmartProxy.find_by(url: params[:url]) : SmartProxy.pulp_primary

          fail Foreman::Exception, N_('Smart proxy content source not found!') unless proxy
          fail Foreman::Exception, N_('Pulp 3 is not enabled on Smart proxy!') unless proxy.pulp3_enabled?
          
          proxy
        end
```

Then you can just call `smart_proxy.rhsm_url` without first having to call `find_smart_proxy`. Not having to rely on side effects is IMHO a good pattern.

jturel (approved):
APJ

ekohl (approved):

jturel:
[test katello]

PR #9422, opened 2021-06-23
Fixes #32862 - the comment is moved to the core

ares:
see https://projects.theforeman.org/issues/32858 for the change in core
that adds this to the safe list


ofedoren:
@ares, you then probably want to remove the dsl documentation entry as well :)

https://github.com/Katello/katello/blob/d705496b0fcf8243bb93c7520f9f372facd2cc08/app/models/katello/concerns/host_managed_extensions.rb#L88

jeremylenz (changes requested):
LGTM except for @ofedoren's feedback above!

ares:
great catch, updated

ofedoren (approved):

jeremylenz (approved):

PR #9423, opened 2021-06-23
Fixes #32867 - Add validations around collection repos

sjha4:
To test:

1. Create an ansible collection repo with :
Remote URL: https://galaxy.ansible.com
 
Requirements.yaml:
```
---
collections:
- name: testing.ansible_testing_content
  version: ">=1.0.0,<=2.0.0"
  source: https://galaxy-dev.ansible.com/
- testing.k8s_demo_collection
```

2. Create a repo with:
Remote URL: https://galaxy.ansible.com/
 Requirements.yaml:
Upload any non yaml file

Corner case:

Pulp validates URL inside requirement.yaml to end with trailing / and also parses the requirements.yaml to meet galaxy requirements. Simple YAML syntax validation will not catch those. Skipping failed remote create action in those cases (instead of the error and pause) and the repository still gets created.


app/services/katello/pulp3/repository/ansible_collection.rb
```diff
@@ -8,7 +8,7 @@ def remote_options
           if root.url.blank?
             super
           else
-            common_remote_options.merge(url: root.url,
+            common_remote_options.merge(url: root.url.chomp('/').concat('/'),
                                         requirements_file: root.ansible_collection_requirements.blank? ? nil : root.ansible_collection_requirements,
```
sjha4, line 12: Not sure about this or if we should be adding the trailing slash if things slip through cracks!

chris1984 (approved):
ACK, works good was able to reproduce the issue and then with the patch I was able to create a collection

PR #9424, opened 2021-06-24
Fixes #32882 - consider pulp3 for download policy

jlsherrill:


vsedmik:
Patch tested downstream on Sat 6.10.0 snap 5 (katello-4.1.0-0.6.rc2.el7sat.noarch) - the download policy is back and configurable. :+1:

chris1984 (approved):
ACK, works for me i can see the policy back

chris1984 (commented):
Do we have unit tests for this?

jlsherrill:
We don't really have tests for templates, thanks for the review!

PR #9426, opened 2021-06-25
Fixes #16154 - Limit package version count to sync

sbernhard:


app/services/katello/pulp3/repository/yum.rb
```diff
@@ -15,6 +15,7 @@ def remote_options
           options = common_remote_options
           options.merge!(sles_auth_token: sles_token) if sles_token
           options.merge!(url: url, policy: root.download_policy)
+          options.merge!(retain_package_versions: root.retain_package_versions_count) unless root.retain_package_versions_count.nil? || root.retain_package_versions_count.to_i == 0
```
jlsherrill, line 18: This 'unless' causes the whole method to return nil unless you are using  retain_package_versions.  I think this is causing the test failures.

However, i think we always need to add retain_package_versions_count to the remote_options otherwise the user doesn't have the ability to unset that option i don't think?  However i think always setting this will mean you have to re-record a lot of the pulp3 vcr tests :)

sbernhard:
[test katello]

sbernhard:
[test katello]

sbernhard:
[test katello]

app/services/katello/pulp3/repository/yum.rb
sbernhard: Thanks a lot!

sbernhard:
Looks like the VCRs are the "last" remaining thing. Can you have a look at this PR again @jlsherrill? Is there a documentation available how to re-record the VCRs?

app/services/katello/pulp3/repository/ansible_collection.rb
```diff
@@ -16,6 +16,10 @@ def remote_options
           end
         end
 
+        def create_options
+          common_create_options
+        end
```
jlsherrill, line 21: this can be moved to app/services/katello/pulp3/repository.rb and then the yum part can override it

jlsherrill: and this prevents the need for every single class to define it

jlsherrill:
One small comment, otherwise looks good.

to record vcr cassettes (for pulp3), you can just run:
```
bundle exec rake test:katello:test:pulpcore  mode=all
```
I think that should cover all the tests

app/services/katello/pulp3/repository/ansible_collection.rb
sbernhard: Thanks. I have revised the section again.

jlsherrill:
[test katello]

sbernhard:
Sorry, its frustrating. I can not find a pulp python content type rpm.

jlsherrill:
I opened a pr here: https://github.com/ATIX-AG/katello/pull/2  to hopefully resolve the vcr errors

sbernhard:
Thanks @jlsherrill - Tests are green :)

jlsherrill:
I noticed that i created a yum repo, set the limit (after unsetting mirror on sync), and then went back and turned on mirror on sync and got the error saying that both the package limit and mirror on sync could not be used together.  It turns out that repo update isn't updating the package limit properly:

https://github.com/Katello/katello/blob/master/app/services/katello/pulp3/repository.rb#L189  

needs a small change:
```
api.repositories_api.update(repository_reference.try(:repository_href), create_options)
```

sbernhard:
thanks again.

jlsherrill:
I think test/actions/pulp3/orchestration/yum_update_test.rb  needs to be re-recorded, wanna give that a shot @sbernhard ?

jlsherrill (approved):
Thanks @sbernhard !!

sbernhard:
Thank you for your help @jlsherrill

PR #9428, opened 2021-06-25
Fixes #32887 - abort all pulp3 migration tasks during abort

jlsherrill:
task


jjeffers (approved):
Tested this as part of 9385 testing, was able to clear out a waiting task, etc.

jjeffers:
[test katello]

PR #9429, opened 2021-06-25
Fixes #32889 - Add setting to move Actions::Katello::Host::UploadPackageProfile out of dynflow

jeremylenz:
- [x] make it work with Actions::Katello::Host::UploadPackageProfile
- [x] make it work with Actions::Katello::Host::UploadProfiles
- [x] add a Setting
- [x] update tests

This PR makes a few changes:
1. Extract logic from `Actions::Katello::Host::UploadPackageProfile` into a new `PackageProfileUploader`
2. Extract logic from `Actions::Katello::Host::UploadProfiles` into a new `ProfilesUploader`
3. Add a Setting to choose whether to use Dynflow for these actions or to let Rails execute the logic directly (default is the latter)
4. Call the new uploaders both from the CandlepinDynflowProxyController and from the Dynflow actions, depending on the Setting

### To test

To trigger package upload from a host (EL7):
```bash
$ katello-package-upload --force
```

To force using UploadPackageProfile instead of UploadProfiles:
Edit app/controllers/katello/api/rhsm/candlepin_proxies_controller.rb and comment out line 246
```rb
# candlepin_response[:managerCapabilities] << 'combined_reporting'
```

#### Test `katello-package-upload --force` using the following matrix:
------------------------
| | 'combined_reporting' line commented out | 'combined_reporting' line uncommented | 
| ------------- | ------------- | ---------------------|
|  Setting set to true | | |
| Setting set to false | | |

I would just put some Rails.logger.info 's in the controller and in the Dynflow action, so you know which it's using. Trigger uploads from the client and make sure everything works OK, with and without the Dynflow actions.

### DIFF
--- a/app/models/setting/content.rb
+++ b/app/models/setting/content.rb
@@ -157,7 +157,9 @@ class Setting::Content < Setting
       self.set('autosearch_delay', N_('If Autosearch is enabled, delay in milliseconds before executing searches while typing.'),
                500, N_('Autosearch delay')),
       self.set('bulk_load_size', N_('The number of items fetched from a single paged Pulp API call.'), 2000,
-               N_('Pulp bulk load size'))
+               N_('Pulp bulk load size')),
+      self.set('upload_profiles_without_dynflow', N_('Allow Candlepin profile uploads to use Rails directly instead of being wrapped in Dynflow actions'), true,
+               N_('Upload profiles without Dynflow'))
     ]
   end
 
--- a/app/models/setting/content.rb
+++ b/app/models/setting/content.rb
@@ -158,7 +158,7 @@ class Setting::Content < Setting
                500, N_('Autosearch delay')),
       self.set('bulk_load_size', N_('The number of items fetched from a single paged Pulp API call.'), 2000,
                N_('Pulp bulk load size')),
-      self.set('upload_profiles_without_dynflow', N_('Allow Candlepin profile uploads to use Rails directly instead of being wrapped in Dynflow actions'), true,
+      self.set('upload_profiles_without_dynflow', N_('Allow Katello to update host installed packages, enabled repos, and module inventory directly instead of wrapped in Dynflow tasks (try turning off if Puma processes are using too much memory)'), true,
                N_('Upload profiles without Dynflow'))
     ]
   end
--- a/app/services/katello/candlepin/package_profile_uploader.rb
+++ b/app/services/katello/candlepin/package_profile_uploader.rb
@@ -7,7 +7,6 @@ module Katello
         Rails.logger.info "SIZE: #{profile_string.size.to_s(:human_size)}"
         #free the huge string from the memory
         profile_string = 'TRIMMED'.freeze
-        Rails.logger.info "PROFILEUPLOADER ********************************"
         import_package_profile_for_host(host_id, profile)
         ::Katello::Host::ContentFacet.trigger_applicability_generation(host_id)
       end
--- a/app/services/katello/candlepin/package_profile_uploader.rb
+++ b/app/services/katello/candlepin/package_profile_uploader.rb
@@ -12,22 +12,21 @@ module Katello
         ::Katello::Host::ContentFacet.trigger_applicability_generation(host_id)
       end
 
-      private
-        def self.import_package_profile_for_host(host_id, profile)
-          host = ::Host.find_by(:id => host_id)
-          if host.nil?
-            Rails.logger.warn("Host with ID %s not found, continuing" % host_id)
-          elsif host.content_facet.nil? || host.content_facet.uuid.nil?
-            Rails.logger.warn("Host with ID %s has no content facet, continuing" % host_id)
-          else
-            begin
-              simple_packages = profile.map { |item| ::Katello::Pulp::SimplePackage.new(item) }
-              host.import_package_profile(simple_packages)
-            rescue ActiveRecord::InvalidForeignKey # this happens if the host gets deleted in between the "find_by" and "import_package_profile"
-              Rails.logger.warn("Host installed package list with ID %s was not able to be written to the DB (host likely is deleted), continuing" % host_id)
-            end
+      def self.import_package_profile_for_host(host_id, profile)
+        host = ::Host.find_by(:id => host_id)
+        if host.nil?
+          Rails.logger.warn("Host with ID %s not found, continuing" % host_id)
+        elsif host.content_facet.nil? || host.content_facet.uuid.nil?
+          Rails.logger.warn("Host with ID %s has no content facet, continuing" % host_id)
+        else
+          begin
+            simple_packages = profile.map { |item| ::Katello::Pulp::SimplePackage.new(item) }
+            host.import_package_profile(simple_packages)
+          rescue ActiveRecord::InvalidForeignKey # this happens if the host gets deleted in between the "find_by" and "import_package_profile"
+            Rails.logger.warn("Host installed package list with ID %s was not able to be written to the DB (host likely is deleted), continuing" % host_id)
           end
         end
+      end
     end
   end
 end
--- a/app/services/katello/candlepin/package_profile_uploader.rb
+++ b/app/services/katello/candlepin/package_profile_uploader.rb
@@ -1,6 +1,6 @@
 module Katello
   module Candlepin
-    class ProfileUploader
+    class PackageProfileUploader
       def self.upload(profile_string:, host:)
         host_id = host.id
         profile = JSON.parse(profile_string)
--- a/app/services/katello/candlepin/profiles_uploader.rb
+++ b/app/services/katello/candlepin/profiles_uploader.rb
@@ -1,0 +1,61 @@
+module Katello
+  module Candlepin
+    class ProfilesUploader
+      def self.upload(profile_string:, host: nil, host_id: nil)
+        profiles = JSON.parse(profile_string)
+        #free the huge string from the memory
+        profile_string = 'TRIMMED'.freeze
+        if host.nil?
+          Rails.logger.warn("Host with ID %s not found; continuing" % host_id)
+        elsif host.content_facet.nil? || host.content_facet.uuid.nil?
+          Rails.logger.warn("Host with ID %s has no content facet; continuing" % host_id)
+        elsif profiles.try(:has_key?, "deb_package_profile")
+          # remove this when deb_package_profile API is removed
+          payload = profiles.dig("deb_package_profile", "deb_packages") || []
+          import_deb_package_profile(host, payload)
+        else
+          module_streams = []
+          profiles.each do |profile|
+            payload = profile["profile"]
+            case profile["content_type"]
+            when "rpm"
+              PackageProfileUploader.import_package_profile_for_host(host_id, payload)
+            when "deb"
+              import_deb_package_profile(host, payload)
+            when "enabled_repos"
+              host.import_enabled_repositories(payload)
+            else
+              module_streams << payload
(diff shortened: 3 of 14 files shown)


app/models/setting/content.rb
```diff
@@ -157,7 +157,9 @@ def self.default_settings
       self.set('autosearch_delay', N_('If Autosearch is enabled, delay in milliseconds before executing searches while typing.'),
                500, N_('Autosearch delay')),
       self.set('bulk_load_size', N_('The number of items fetched from a single paged Pulp API call.'), 2000,
-               N_('Pulp bulk load size'))
+               N_('Pulp bulk load size')),
+      self.set('upload_profiles_without_dynflow', N_('Allow Candlepin profile uploads to use Rails directly instead of being wrapped in Dynflow actions'), true,
```
jturel, line 161: The description of this is too low level as far as the user is concerned. We might word it to articulate when they might change the setting

app/services/katello/candlepin/package_profile_uploader.rb
```diff
@@ -0,0 +1,31 @@
+module Katello
+  module Candlepin
```
jturel, line 2: Would changing the Candlepin namespace to Host be more accurate? The package profile doesn't get stored in Candlepin

app/services/katello/candlepin/profiles_uploader.rb
```diff
@@ -0,0 +1,61 @@
+module Katello
+  module Candlepin
```
jturel, line 2: Ditto on Candlepin vs Host namespace

app/services/katello/candlepin/profiles_uploader.rb
```diff
@@ -0,0 +1,61 @@
...
+            else
+              module_streams << payload
+            end
+          end
+
+          module_streams.each do |module_stream_payload|
+            import_module_streams(module_stream_payload, host)
+          end
+
+          ::Katello::Host::ContentFacet.trigger_applicability_generation(host.id)
```
jturel, line 36: I think applicability regeneration should be decoupled from this class

app/services/katello/candlepin/package_profile_uploader.rb
```diff
@@ -0,0 +1,31 @@
...
+    class PackageProfileUploader
+      def self.upload(profile_string:, host:)
+        host_id = host.id
+        profile = JSON.parse(profile_string)
+        Rails.logger.info "SIZE: #{profile_string.size.to_s(:human_size)}"
+        #free the huge string from the memory
+        profile_string = 'TRIMMED'.freeze
+        import_package_profile_for_host(host_id, profile)
+        ::Katello::Host::ContentFacet.trigger_applicability_generation(host_id)
```
jturel, line 11: Ditto on separating applicability regen from profiles upload

app/services/katello/candlepin/package_profile_uploader.rb
```diff
@@ -0,0 +1,31 @@
...
+      end
+
+      def self.import_package_profile_for_host(host_id, profile)
+        host = ::Host.find_by(:id => host_id)
```
jturel, line 15: The `upload` method already had the host object, and this code takes the ID and looks it up again

app/services/katello/candlepin/profiles_uploader.rb
```diff
@@ -0,0 +1,61 @@
+module Katello
+  module Candlepin
+    class ProfilesUploader
+      def self.upload(profile_string:, host: nil, host_id: nil)
```
jturel, line 4: If we're passing a Host every time, is there a need to pass host_id separately?

jturel (changes requested):
Gave it a very quick code review pass, no testing performed

app/services/katello/candlepin/package_profile_uploader.rb
jeremylenz: This is because I was trying to preserve the following logic and not alter it (my mindset was that this is a refactor, so only make changes that are semantically necessary due to the code moves.)

```ruby
if host.nil?
          Rails.logger.warn("Host with ID %s not found; continuing" % host_id)
        elsif host.content_facet.nil? || host.content_facet.uuid.nil?
          Rails.logger.warn("Host with ID %s has no content facet; continuing" % host_id)
        else
...
```

however, I'm failing to see how `host` would ever be nil anyway. Should those warning logs be moved to the controller, maybe? Open to suggestions..

app/models/setting/content.rb
jeremylenz: @jturel I changed it but it's still quite wordy. open to suggestions (also @parthaa)

app/services/katello/candlepin/package_profile_uploader.rb
jeremylenz: makes sense! Done

app/services/katello/candlepin/profiles_uploader.rb
jeremylenz: Okay, but that means I have to un-DRY it a bit and call it both from the controller and from the task.  Updated.  (or did you mean something else?)

app/services/katello/candlepin/profiles_uploader.rb
jeremylenz: This was more refactoring oddities. Works much better if we just pass `host`.  Updated.

app/services/katello/candlepin/profiles_uploader.rb
jturel: That's just what I meant! I see upload profiles and regen applicability as two steps in a workflow. Honestly it's not the end of the world to have it the way you did (the way it *was*) but I do like the explicit-ness of calling them both in serial

jturel:
Thanks for making the updates! This is shaping up very nicely but can we take it a bit further? Here's what I mean:

There's a significant amount of code duplication in the new profile uploader classes and I see an opportunity to address this now since we are at a point where this all needs to be tested end-to-end due to the refactor. Here's some pseudo-code to illustrate what this could look like in the end. It's not perfect either because it combines the notions of the old and new profiles, but having all this functionality in one place is also an advantage. Anyway I just wanted to see if any of this appeals to you...

```
class Uploader

  # this method will raise an error if no host is given, and it should because we can't do anything without it
  def initialize(host:, profiles_string: nil, legacy_profile_string: nil)
    @host = host
    @profiles_string = profiles_string
    @legacy_profile_string = legacy_profile_string

     # since we need once of the profile types ...
     fail("Why did you do this?") unless profiles_string || legacy_profile_string
  end

  def upload
    if @profiles_string
      upload_profiles
    else
      upload_legacy_profile
    end

    # another way could be to coerce the old profile to look like the new profiles hash, but just with package data
  end

  def regenerate_host_applicability
    # this can be called separately, but still encapsulate how to calc applicability as far as the Uploader sees it
  end

  private

  def upload_legacy_profile
    # do the old stuff
  end
  
   def upload_profiles
     # do the new stuff
   end
end
```

Example usage:

```
 uploader = Uploader.new(host: my_host, profiles_string: my_profiles_string)
 uploader.upload
 uploader.regenerate_host_applicability
```

I like using an instance vs static method for this kind of thing, because it really is stateful. The initialize method can make sure we didn't give it complete junk, and the `upload` method can be concerned simply with uploading! In the future if we need to track more state that becomes easier. This worked well for me with the `OrganizationCreator` recently

jeremylenz:
```rb
def regenerate_host_applicability
    # this can be called separately, but still encapsulate how to calc applicability as far as the Uploader sees it
end
```

Wouldn't that re-couple applicability generation to the class? :wink:

jturel:
> ```ruby
> def regenerate_host_applicability
>     # this can be called separately, but still encapsulate how to calc applicability as far as the Uploader sees it
> end
> ```
> 
> Wouldn't that re-couple applicability generation to the class?

Yep, but as you mentioned before - that's the way it was. That was a relatively minor detail of my suggested refactor, just to illustrate the possibilities. Throwing things at a wall, seeing what sticks, etc :)

jeremylenz:
Ok @jturel, I combined both new classes into an `Uploader` class and converted to the new usage in the candlepin_dynflow_proxies_controller.  But I feel like it kind of makes `uploader.rb` rather large and hard to digest, and makes the purposes less clear (before there was a class for the legacy case and the modern case, now there's just one class and different methods are used.)  Thoughts?

jturel:
> Ok @jturel, I combined both new classes into an `Uploader` class and converted to the new usage in the candlepin_dynflow_proxies_controller. But I feel like it kind of makes `uploader.rb` rather large and hard to digest, and makes the purposes less clear (before there was a class for the legacy case and the modern case, now there's just one class and different methods are used.) Thoughts?

I'm glad you left the original classes you had because looking at it now, I like what you did better. Sorry for over-complicating. One day we'll stop supporting the Old Way and being able to just delete the related class is advantageous.

jeremylenz:
@jturel thanks for giving me some ideas :)

I erased those two commits so I could go back to what I had before.  But Rubocop was complaining (something about incorrect use of the `private` keyword.)  So I decided to convert to using class instances here too, but keeping the two classes separate.  I think it works well.. let me know your thoughts.

app/services/katello/host/package_profile_uploader.rb
```diff
@@ -0,0 +1,41 @@
...
+        #free the huge string from the memory
+        @profile_string = 'TRIMMED'.freeze
+        import_package_profile(profile)
+      end
+
+      def import_package_profile(profile)
+        self.class.import_package_profile_for_host(@host.id, profile)
+      end
+
+      def trigger_applicability_generation
```
jeremylenz, line 20: I kept `trigger_applicability_generation` as two un-DRY methods in each class. My reasoning process was

1. I should probably DRY this up and have both classes call a single `trigger_applicability_generation`
2. Which class should it live in? Probably this one, since it already has another shared method (`import_package_profile_for_host`)
3. But we're going to delete this class eventually anyway, so maybe it should go in the other one?
4. But if it goes in `ProfilesUploader`, I'd have one shared method there and one shared method here. Makes more sense to pick one of the classes to have all the shared methods
5. It would make more sense for the shared methods to live in the class that's not getting deleted eventually. So it should live in `ProfilesUploader` and not here.
6. But it's a lot of work to move the logic for the other shared method over there (import_package_profile_for_host)
7. Also, `@host` might be nil in that class, but will never be nil here.
8. So it's just easier to have two `trigger_applicability_generation` methods, one for each class, and leave the rest as is.

jturel:
Any idea of any throughput gained here? ie time difference between 10k profile uploads with the setting true and false?

jeremylenz:
@jturel nope, I've only tested it one upload at a time so far!

jturel (commented):
I'm OK with ultimately doing a lift+shift of the code as it is here. It's been running in production as it is for a very long time and it works

jeremylenz:
[test katello]

jeremylenz:
Updated existing tests and added new tests.

parthaa (approved):
Before this pr. PackageProfile uploads on a 1000 hosts
```
real	9m22.570s
user	0m8.767s
sys	0m1.241s
```

After the PR
```
real	6m32.302s
user	0m8.420s
sys	0m1.214s
```
Not much difference in the memory / cpu usage

ACK

jeremylenz:
@jturel Any other concerns?

jturel:
> 
> 
> @jturel Any other concerns?

LGTM :)

jeremylenz:
@jturel you had requested changes, can you please approve? :)

jturel (approved):

PR #9432, opened 2021-06-28
Fixes #31853 - Ignore timeout events to avoid errors if they cannot be delivered

lfu:
Related to https://github.com/Dynflow/dynflow/pull/383.

### DIFF
--- a/app/lib/actions/katello/agent_action.rb
+++ b/app/lib/actions/katello/agent_action.rb
@@ -47,14 +47,14 @@ module Actions
 
             dispatch_message(history) unless input[:bulk]
 
-            schedule_timeout(timeout)
+            schedule_timeout(timeout, optional: true)
           end
         when Dynflow::Action::Skip
           # Do not fail and goto a paused state, instead skip and send the state to warning so we do not block other host actions
         when Dynflow::Action::Timeouts::Timeout
           process_timeout
         when 'accepted'
-          schedule_timeout(finish_timeout)
+          schedule_timeout(finish_timeout, optional: true)
           suspend
         else
           fail_on_errors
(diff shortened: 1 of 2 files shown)


chris1984:
To test, turn off `qpid` with `systemctl` and then send a task.

`systemctl stop qpidd`
* Send a action

lfu:
@jeremylenz Could you comment how to test this? Thanks.

jturel:
I think we need to pin dynflow to >= 1.5.0 since that's the version which enables this capability

lfu:
Re-produce the issue with the following steps:
- Register client as content host to Katello
- Install katello-agent on client
- From UI select the client and try to install a package eg. `screen`
- Remove installed package from the client

Just keep installing and removing the package. Very soon the error message such as the following would show up in the rails console when running in development mode:
```
E, [2021-06-29T20:38:35.509420 #19860] ERROR -- /parallel-executor-core: no manager for Dynflow::Director::Event[request_id: fa1d807c-696e-4940-8115-4456d79be2a8-10, execution_plan_id: 2731bb36-f0b3-46ca-9ca3-5b52eb848869, step_id: 2, event: Dynflow::Action::Timeouts::Timeout, result: #<Concurrent::Promises::ResolvableFuture:0x0000000016cfdeb8 pending>, optional: false] (Dynflow::Error)
```

Tested again with this commit and there is no more error messages like above.

jeremylenz (changes requested):
I did the test described above several times. works great! :+1: 

:heavy_check_mark: Reproduced the error message by installing & removing screen package
:heavy_check_mark: Checked out this PR and the error message did not occur.

I agree with Jonathan that we should make sure we're using the correct dynflow.  In the `katello.gemspec` file, please change

```rb
gem.add_dependency "dynflow", ">= 1.2.0"
```
to
```rb
gem.add_dependency "dynflow", ">= 1.5.0"
```

Otherwise LGTM

jturel (approved):
LGTM based on testing by @jeremylenz

jeremylenz (approved):
APJ

PR #9433, opened 2021-06-29
Refs #32904 - fix new pf4 veresion snapshots

MariaAga:


jturel:
Hi @MariaAga - thank you for the PR. Unfortunately we didn't see this before we merge in another PR which fixed the react failures => https://github.com/Katello/katello/pull/9435 Seems like this one can be closed now; will leave it to you!

PR #9438, opened 2021-07-01
Fixes #32922 - drop Foreman::Environment::Destroy

ezr-ondrej:
This action class is no longer used and can be removed.
Tests would fail on this once we get on Foreman 3.0.

This is blocking https://github.com/theforeman/foreman/pull/8510


tbrisker (approved):
Thanks @ezr-ondrej, merging. Test failures are unrelated (settings race condition which seems to be too common)

PR #9440, opened 2021-07-01
Fixes #32936 - use bulk_host_ids parameter for traces

jturel:
The host_ids parameter was changed to bulk_host_ids on the controller side but the supporting code for traces never got the memo!


jjeffers:
> The host_ids parameter was changed to bulk_host_ids on the controller side but the supporting code for traces never got the memo!

Mea culpa! We thought we caught all uses of this controller, but we clearly did not. :(

jturel:
> > The host_ids parameter was changed to bulk_host_ids on the controller side but the supporting code for traces never got the memo!
> 
> Mea culpa! We thought we caught all uses of this controller, but we clearly did not. :(

It's too easy to introduce bugs in the angular app :-1:

jturel:
[test katello]

jturel:
[test katello]

jeremylenz (approved):
:heavy_check_mark: Checking out this change avoids the referenced Rails error

Didn't test with a full working REX setup, but otherwise LGTM :+1:

chris1984 (approved):
ACK, works great

PR #9441, opened 2021-07-02
Fixes #32901 - inc update doesn't generate metadata

ianballou:
Inc update was missing the metadata generation and content indexing.

To test:
1) Incrementally update a content view version with a new package and/or erratum
2) Ensure the new content view version has a LCE and look at the repo metadata + content
3) See that the new package(s) are missing
4) Apply this PR
5) Run the incremental update again
6) Check the repo metadata and packages again. See that the appropriate content is now there


jjeffers (approved):
Worked! After applying this change, I was able to install packages in the incremental update on a registered host. The changes are also minimal, which makes this pretty easy on the eyes.

APJ

PR #9442, opened 2021-07-02
Fixes #32938 - Added components for new host details page

parthaa:
Includes work from PR 8948

Very barebones place holders for Content and Subscription tabs

### DIFF
--- a/webpack/components/extensions/HostDetails/Tabs/ContentTab.js
+++ b/webpack/components/extensions/HostDetails/Tabs/ContentTab.js
@@ -1,0 +1,56 @@
+import React, { useState } from 'react';
+import EmptyPage from 'foremanReact/components/common/EmptyState/EmptyStatePattern';
+import { Tabs, Tab, TabTitleText } from '@patternfly/react-core';
+import { translate as __ } from 'foremanReact/common/I18n';
+
+const tabs = [
+  {
+    key: 'packages',
+    title: __('Packages'),
+    content: <EmptyPage
+      icon="enterprise"
+      header="WIP Packages"
+      description="This is a demo for adding content to the new host details page"
+    />,
+  },
+  {
+    key: 'errata',
+    title: __('Errata'),
+    content: <EmptyPage
+      icon="enterprise"
+      header="WIP Errata"
+      description="This is a demo for adding content to the new host details page"
+    />,
+  },
+  {
+    key: 'modulestreams',
+    title: __('Module Streams'),
+    content: <EmptyPage
+      icon="enterprise"
+      header="WIP Module Streams"
+      description="This is a demo for module streams on new host details page"
+    />,
+  },
+];
+
+const ContentTab = () => {
+  const [activeTab, setActiveTab] = useState(tabs[0].key);
+  const handleTabClick = (event, tabIndex) => setActiveTab(tabIndex);
+  return (
+    <Tabs
+      isSecondary
+      activeKey={activeTab}
+      onSelect={handleTabClick}
+    >
+
+      { tabs.map(tab => (
+        <Tab eventKey={tab.key} title={<TabTitleText>{tab.title}</TabTitleText>}>
+          {tab.content}
+        </Tab>
+            ))
+          }
+    </Tabs>
+  );
+};
+
+export default ContentTab;
--- a/webpack/components/extensions/HostDetails/Tabs/ContentTab.js
+++ b/webpack/components/extensions/HostDetails/Tabs/ContentTab.js
@@ -3,38 +3,8 @@ import EmptyPage from 'foremanReact/components/common/EmptyState/EmptyStatePatte
 import { Tabs, Tab, TabTitleText } from '@patternfly/react-core';
 import { translate as __ } from 'foremanReact/common/I18n';
 
-const tabs = [
-  {
-    key: 'packages',
-    title: __('Packages'),
-    content: <EmptyPage
-      icon="enterprise"
-      header="WIP Packages"
-      description="This is a demo for adding content to the new host details page"
-    />,
-  },
-  {
-    key: 'errata',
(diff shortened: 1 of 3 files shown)


parthaa:
screenshot

sjha4:
Would it be helpful to use RoutedTabs here similar to the CV UI: https://github.com/Katello/katello/blob/master/webpack/scenes/ContentViews/Details/ContentViewDetails.js#L119 ?

parthaa:
> Would it be helpful to use RoutedTabs here similar to the CV UI:

So looks like routed only works top level and is going to take a lot fighting with `fills` to get it to work. I think using regular tabs is simple enough (even though I love reusing components)

webpack/components/extensions/HostDetails/Tabs/ContentTab.js
```diff
@@ -0,0 +1,62 @@
+import React, { useEffect, useState } from 'react';
+import RoutedTabs from '../../../../components/RoutedTabs';
+import EmptyPage from 'foremanReact/components/common/EmptyState/EmptyStatePattern';
+import { Tabs, Tab, TabTitleText} from '@patternfly/react-core';
+
+
+const ContentTab = () => {
+  const [activeTab, setActiveTab] = useState('Content');
+  const tabs = [
```
Andrewgdewar, line 9: This can be moved above the function scope, as it is a constant that doesn't use state.

webpack/components/extensions/HostDetails/Tabs/ContentTab.js
```diff
@@ -0,0 +1,62 @@
+import React, { useEffect, useState } from 'react';
+import RoutedTabs from '../../../../components/RoutedTabs';
+import EmptyPage from 'foremanReact/components/common/EmptyState/EmptyStatePattern';
+import { Tabs, Tab, TabTitleText} from '@patternfly/react-core';
+
+
+const ContentTab = () => {
+  const [activeTab, setActiveTab] = useState('Content');
```
Andrewgdewar, line 8:
```suggestion
  const [activeTab, setActiveTab] = useState(tabs[0].key);
```

webpack/components/extensions/HostDetails/Tabs/ContentTab.js
```diff
@@ -0,0 +1,62 @@
...
+    },
+  ];
+
+  const handleTabClick = (event, tabIndex) => {
+    setActiveTab(tabIndex);
+  };
+
+  useEffect(() => {
+    setActiveTab(tabs[0].key)
+  }, []);
```
Andrewgdewar, line 45: This can be removed as you are using a constant, just add it as the default state.

webpack/components/extensions/HostDetails/Tabs/ContentTab.js
```diff
@@ -0,0 +1,62 @@
...
+                  icon="enterprise"
+                  header="WIP Module Streams"
+                  description="This is a demo for module streams on new host details page"
+                />,
+    },
+  ];
+
+  const handleTabClick = (event, tabIndex) => {
+    setActiveTab(tabIndex);
+  };
```
Andrewgdewar, line 41: As a suggestion,I like the look of direct returns, nothing wrong with this but I'd usually do: 
const handleTabClick = (event, tabIndex) => setActiveTab(tabIndex)

parthaa:
Aslo filed https://projects.theforeman.org/issues/32964  for @amirfefer @laviro @jeremylenz @sjha4  to look into...

laviro (commented):
Thanks @parthaa, code wise looks good to me,
I would suggest using a routable tabs so we could link into one of the tabs from outside :)

jeremylenz:
@laviro Yeah that's the plan, I was thinking we'd use RoutedTabs once it's in Foreman, but until then just keep it as is so we have consistent behavior for the whole host details page.

webpack/components/extensions/HostDetails/Tabs/ContentTab.js
```diff
@@ -0,0 +1,56 @@
+import React, { useState } from 'react';
+import EmptyPage from 'foremanReact/components/common/EmptyState/EmptyStatePattern';
+import { Tabs, Tab, TabTitleText } from '@patternfly/react-core';
+import { translate as __ } from 'foremanReact/common/I18n';
+
+const tabs = [
```
jeremylenz, line 6: Patternfly tables have instilled a bad habit in us of intertwining our JSX markup and our data objects. I think code looks a lot cleaner when the JSX is only at the bottom in render/return, and all the other stuff is here at the top.  Instead of the `map` down on line 46 I feel like it will make this component simpler and easier to read if it's just the literal JSX. So like

```js
return (
    <Tabs
      isSecondary
      activeKey={activeTab}
      onSelect={handleTabClick}
    >
      <Tab title={__('Packages')}...>
        <EmptyPage header="WIP Packages" ... />
      </Tab>
      <Tab title={__('Errata')} ...>
          <EmptyPage header="WIP Errata" ... />
     </Tab>
     ...
    </Tabs>
  );
```
and then you won't need the `tabs` const at all. This way you can easily see what the markup structure on the page will look like too.

jeremylenz (commented):
Thanks for diving in @parthaa!

Works as expected.  One comment below; let me know your thoughts

webpack/components/extensions/HostDetails/Tabs/ContentTab.js
parthaa: Addressed

jeremylenz:
Got some console warnings in testing but I'm confident they're not related. Opened https://projects.theforeman.org/issues/32967 in Foreman.

jeremylenz (approved):
Changes look great! Thanks @parthaa 
APJ

PR #9445, opened 2021-07-07
Refs #32955 - Remove old export from repositories controller

chris1984:


jturel:
lib/katello/permission_creator.rb probably needs an update too!

jlsherrill:
here it is:  https://github.com/Katello/katello/blob/master/lib/katello/permission_creator.rb#L320-L325

we also need a db migration to remove it i think?

chris1984:
sounds good, will update it

chris1984:
@jlsherrill updated with a migration and permissins_creator fixed. Migration went well:

```bash
== 20210714140440 RemoveRepoExportPermission: migrating =======================
== 20210714140440 RemoveRepoExportPermission: migrated (0.0371s) ==============
```

jlsherrill (approved):

PR #9449, opened 2021-07-08
Fixes #32965 - Have repo delete go directly to task instead of toast

chris1984:
This PR removes the toast notification and sends the user to the task upon a repo delete

### DIFF
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/product-repositories.controller.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/product-repositories.controller.js
@@ -65,10 +65,7 @@ angular.module('Bastion.products').controller('ProductRepositoriesController',
             var success, error, params = getParams(), removalPromise;
 
             success = function (response) {
-                var message = translate('Removal of selected repositories initiated successfully.');
-                var link = ("/foreman_tasks/tasks/%taskId").replace('%taskId', response.task.id);
-                var alertBody = { children: translate("Click to view task"), href: link };
-                Notification.setSuccessMessage(message, {link: alertBody});
+                $state.go('product.tasks.details', {taskId: response.task.id});
             };
 
             error = function (response) {


ares:
Is there a way to keep user on the page, keeping the notification but remove it from the page as it's at least being deleted? Redirecting to the task detail is not great from UX perspective, most likely I want to continue working with the product but I have to navigate through at least 2 pages to get back.

jlsherrill:
I agree, that would be nice, however its a bit trickier than that as:

* some subset of repos could fail to delete (for examine if one repo was published in a cv)
* an error could occur that has to be resolved
* This is an async task and could take a while to complete (for example it may have to wait for pulp to finish syncing some other set of repos)
* its actually only a single click to get back to the 'repositories tab' 

@ares any objection to going ahead and merging this?  This was a 6.10 bz opened against behavior that has existed for many years :)   If you feel that the proposed behavior is worse than the current, i think i'd work to push the z off of 6.10.

ares:
Ah I misread the code. It does not go to generic tasks but products tasks right? I agree then it's a single click. It's still pretty bad UX. I'm OK with merging for now, but it would be great to track futrther improvements of this flow. It's not unique and would be handy on many other places.

So yeah, go ahead, I'll close my eyes.

jlsherrill:
correct, it goes to the product tasks page.  Its not ideal i agree, but i think this is simple improvement.   Thanks @ares !

jlsherrill (approved):

chris1984:
> Ah I misread the code. It does not go to generic tasks but products tasks right? I agree then it's a single click. It's still pretty bad UX. I'm OK with merging for now, but it would be great to track futrther improvements of this flow. It's not unique and would be handy on many other places.
> 
> So yeah, go ahead, I'll close my eyes.

I will make a tracker redmine and attach this issue to it and link it here.

chris1984:
Tracker for visibility

https://projects.theforeman.org/issues/33016

PR #9450, opened 2021-07-09
Fixes #32972: Drop RegistryResource supporting crane

ehelms:


app/lib/katello/resources/registry.rb
```diff
@@ -26,18 +26,12 @@ def load_class
...
-
-              # Assume the registry uses the same CA as the Smart Proxy
-              ca_cert_file = Setting[:ssl_ca_file]
-            elsif (container_config = SETTINGS.dig(:katello, :container_image_registry))
-              registry_url = container_config[:crane_url]
-              ca_cert_file = container_config[:registry_ca_cert_file]
-            end
+            uri = URI(pulp_primary.setting(SmartProxy::PULP3_FEATURE, 'content_app_url'))
+            uri.path = "/pulpcore_registry/"
+            registry_url = uri.to_s
```
ekohl, line 31: Why is this needed? On line 38 it's parsed again and a new `uri` object created. What's the point?

app/lib/katello/resources/registry.rb
```diff
@@ -26,18 +26,12 @@ def load_class
...
+
+            # Assume the registry uses the same CA as the Smart Proxy
+            ca_cert_file = Setting[:ssl_ca_file]
 
             fail Errors::ContainerRegistryNotConfigured unless registry_url
```
ekohl, line 36: Is this still reachable code? I don't think it is. I think it should fail `unless pulp_primary.setting(SmartProxy::PULP3_FEATURE, 'content_app_url')` is set.

ekohl (changes requested):
Yay for cleaning up, but I think you can cleanup a bit more.

ekohl (approved):
Now we're getting somewhere :)

jlsherrill:
code looks good, let me do a quick test

jlsherrill (approved):

PR #9451, opened 2021-07-09
Fixes #32974 - monkey patch missing enum

jlsherrill:
for pulp_rpm_client 3.13.3


jlsherrill:
and these enum classes essentially do some validation and return the same string, here's an example of what one looks like:

```
require 'date'

module PulpRpmClient
  class MetadataChecksumTypeEnum
    UNKNOWN = "unknown".freeze
    MD5 = "md5".freeze
    SHA1 = "sha1".freeze
    SHA224 = "sha224".freeze
    SHA256 = "sha256".freeze
    SHA384 = "sha384".freeze
    SHA512 = "sha512".freeze

    # Builds the enum from string
    # @param [String] The enum value in the form of the string
    # @return [String] The enum value
    def self.build_from_hash(value)
      new.build_from_hash(value)
    end

    # Builds the enum from string
    # @param [String] The enum value in the form of the string
    # @return [String] The enum value
    def build_from_hash(value)
      constantValues = MetadataChecksumTypeEnum.constants.select { |c| MetadataChecksumTypeEnum::const_get(c) == value }
      raise "Invalid ENUM value #{value} for class #MetadataChecksumTypeEnum" if constantValues.empty?
      value
    end
  end
end

```

I don't think the validation is really that helpful, so in my version it just returns the exact value.

jlsherrill:
To reproduce, you would either need to:

1) install katello with 3.11 and create and sync at least 1 repo, upgrade to 3.14
OR
2) with katello 3.14, create and sync at least 1 repo and run:  
```
$ sudo -u postgres psql pulpcore
update rpm_rpmrepository set metadata_checksum_type = 'sha256';
```

ianballou (approved):
works for me! I suppose we need an issue to remove this once the bindings are un-bugged.

jlsherrill:
opened https://projects.theforeman.org/issues/32976

PR #9454, opened 2021-07-12
Fixes #32994 - always refresh remote

jlsherrill:
we cannot reliably detect if a remote needs update due to
proxy auth and upstream auth information being removed from the pulp3 api.
As a result there is not much we can do other than to always refresh it


app/services/katello/pulp3/repository.rb
```diff
@@ -138,21 +138,6 @@ def get_remote(href = repo.remote_href)
         api.remotes_api.read(href)
       end
 
-      def remote_needs_updates?
-        if repo.remote_href
-          remote = get_remote
-          # The proxy auth creds are not returned by the Pulp API.
-          # The creds don't need to be checked here because they
-          # won't be updated outside of UpdateRemote.
-          computed = compute_remote_options.slice!(:proxy_username, :proxy_password)
```
rverdile, line 147: Just curious, why not remove username/password from computed rather than remove this whole method?

rverdile (commented):
Looks good overall. I just had one question about the code. In terms of functionality everything seems okay to me.

app/services/katello/pulp3/repository.rb
jlsherrill: Good question!  The goal was to update the remote if it needed an update, so we would fetch the remote compare it to what it should look like and then update if it didn't match.  

With the change in pulp that caused this, username and password are no longer included in the response, nor is proxy username or proxy password (from a few months ago), so now with 4 attributes that we can no longer detect changes for It seemed simpler and less error prone to just always update the remote no matter what.  Its also a relatively cheap operation, so it shouldnt' be an issue on sync.

rverdile (commented):
Doubled checked, everything looks good to me! Does it make sense to have someone else do a final review as a safeguard? If not I'll mark it as approved.

jlsherrill:
looks like i need to re-record a ton of cassettes.  But yes, @ianballou or @jjeffers ya'll mind doing a short code review?

jjeffers (approved):
LGTM

PR #9455, opened 2021-07-12
Fixes #33000 - Display checkbox/select column for non-admin users

lfu:
- Use `hasUpstreamConnection` instead of `content_disconnected` from `Setting` to determine if checkbox/select column should be enabled.
- Remove all `content_disconnected` related code since it's only being used by Subscription. 

`view_settings` permission is required to retrieve `content_disconnected` from `Setting`.  A non-admin user may not have this permission.
`hasUpstreamConnection` is set by `pingUpstreamSubscriptions` which does not require a permission.

### Steps to Test
- Create a role and assign 'Organization', 'Location' and 'Subscription' permissions.
- Associate the role with a user. 
- Login as this user and navigate to 'Subscriptions' page. Verify the checkboxes next to each subscription are displayed.


jturel:
Since you and I worked together on this fix I'm not going to formally review it, but have some other recommendations:

There's probably some cleanup to be done here:

```
webpack/scenes/Settings/SettingsConstants.js:export const CONTENT_DISCONNECTED = 'content_disconnected';
webpack/scenes/Settings/SettingsReducer.js:  CONTENT_DISCONNECTED,
webpack/scenes/Settings/SettingsReducer.js:        case CONTENT_DISCONNECTED:
```

The `loadSetting` prop can likely be removed as well - the SubscriptionPage no longer makes any reference to it as far as I can tell

app/services/katello/upstream_connection_checker.rb
```diff
@@ -22,7 +21,6 @@ def can_connect?
     end
 
     def assert_connection
-      assert_connected
```
jturel, line 25: Looking at the Ruby changes as a whole, is it really valid to remove all of this? We relied on this to check if we're in disconnected mode via the #ping method in UpstreamSubscriptionsController

lfu: SubscriptionsPage is the only consumer for the disconnected mode and won't care about it any more after this PR. 
Is the method still necessary if no one cares about  it?

jturel: I think it needs to stay because what changed is:

1. removed call to /api/v2/settings/content_disconnected from the UI to determine if we are in disconnected mode
2. rely totally on /api/v2/upstream_subscriptions/ping to determine if we are in disconnected mode from UI perspective

This means that removing this code breaks the very solution that this PR presents, because now there is *no* way to know if we are disconnected mode in the subscriptions UI

lfu: I get your point and think you are right.
My concern is `assert_connected` does not work from non-admin users.

jturel: I think it will work. The reason is that on the backend side - we aren't filtering any of that based on permissions, and also the disconnected setting check is not using the API. Since we're on the backend side it's just looking in the DB!

jturel: Of course, better to verify that than take my work for it ;)

webpack/scenes/Subscriptions/SubscriptionsPage.js
```diff
@@ -130,27 +118,27 @@ class SubscriptionsPage extends Component {
...
       subscriptionTableSettings,
+      pingUpstreamSubscriptions,
     } = this.props;
 
     pollTasks();
-    loadSetting(CONTENT_DISCONNECTED);
     loadSubscriptions();
     await loadTables();
     loadTableColumns(subscriptionTableSettings);
+    pingUpstreamSubscriptions();
```
jturel, line 133: This call was previously wrapped in an `isManifestImported` check which prevents us from needlessly calling this API. Did it need to move here from its original location?

lfu: I'd prefer collect all necessary data at one point. I can revert it back if it's not a good idea.

jturel: If it works that's OK, but I am curious about what happens when it's not wrapped in the `isManifestImported` check

lfu: I did not notice any behavior change. `pingUpstreamSubscriptions` does not depend on `isManifestImported`.

jturel: I think the idea behind checking `isManifestImported` is that we can reduce the # of API calls the page makes. If there is no manifest imported then there is no need to perform `pingUpstreamSubscriptions` or at least that was the case when the code was written. Since it's not related to your bugfix I see value in not changing that specific behavior. What do you think?

lfu: Revert back to its place.

jturel (approved):
ACK from my perspective, but someone else should test it

Nice work!

chris1984 (approved):
ACK, works great. Screenshot showing user with limited permissions and I can see the checkbox:

https://nimb.ws/LhWV4g

PR #9468, opened 2021-07-15
Refs #32923 - drop last Envrionment references

ezr-ondrej:
We are still testing Puppet Environment in tests.
These should be the last references to get tests passing after puppet is
extracted.

Hopefully last katello blocker for #8510


jturel (approved):
ACK

Thanks @ezr-ondrej

PR #9469, opened 2021-07-15
Fixes #33028 - print postgres space considerations for migration

jlsherrill:
as part of migration-stats command


chris1984 (approved):
ACK, thanks for adding this

PR #9470, opened 2021-07-15
Fixes #33036 - Make via Katello-agent option clickable on the content host errata page

lfu:
[find_bulk_errata_ids](https://github.com/Katello/katello/blob/02e2151a90717faf54cb6c3918d6d1dc2f777ae5/app/controllers/katello/concerns/api/v2/host_errata_extensions.rb#L6) expects `bulk_errata_ids` in JSON string format.

### DIFF
--- a/app/controllers/katello/api/v2/host_errata_controller.rb
+++ b/app/controllers/katello/api/v2/host_errata_controller.rb
@@ -115,7 +115,7 @@ module Katello
         fail HttpErrors::NotFound, _("Couldn't find errata ids '%s'") % missing.to_sentence if missing.any?
         @errata_ids = params[:errata_ids]
       else
-        @errata_ids = find_bulk_errata_ids([@host], params[:bulk_errata_ids])
+        @errata_ids = find_bulk_errata_ids([@host], params[:bulk_errata_ids].to_json)
       end
     end
   end


chris1984 (approved):
ACK, works great. I guess this broke when we did some refactoring.

jturel:
This change seems strange (I'm obviously lacking the context). I wonder if it's really that the UI needs to be fixed to send the proper data format? I'm asking because the wrong change here could easily break other areas. We might have to consult @jjeffers who was modifying code in this area somewhat recently in 02e2151a90717faf54cb6c3918d6d1dc2f777ae5

jjeffers:
It's likely that the changes made previously (which included refactoring the parameters for bulk host ids) did force consumers to use json.

jturel:
> It's likely that the changes made previously (which included refactoring the parameters for bulk host ids) did force consumers to use json.

I fixed a (maybe similar to this) issue recently -> https://github.com/Katello/katello/commit/a186876abe8ba1218834d403e65d4fa4cde59a4e

Seems like the UI should be sending the JSON string, and the server turns it back into a hash?

jturel:
This other bug on our backlog could be another case where we regressed due to the API change in a186876

https://bugzilla.redhat.com/show_bug.cgi?id=1977249

PR #9472, opened 2021-07-15
Fixes #33039 - to_s docker whitelist tags

jlsherrill:
as pulp does not like floats/ints

When using hammer, you can pass `--docker-tags-whitelist "16.1"`  and hammer will treat 16.1 as a float when creating an array to be sent to the api:  [16.1]

If you pass a non-float value, it properly passes a string, for example `--docker-tags-whitelist "16.1a"`   would properly pass ["16.1a"] to the api.

Pulp however doesn't seem to handle floats properly, and treats them as if nothing was passed.   

Since our api does not exclude floats (or integers) today within the array, i thought it would be simple to just fix the integration with pulp to convert it to a string prior to passing to pulp.

### DIFF
--- a/app/services/katello/pulp3/repository/docker.rb
+++ b/app/services/katello/pulp3/repository/docker.rb
@@ -15,7 +15,7 @@ module Katello
         def remote_options
           options = {url: root.url, upstream_name: root.docker_upstream_name}
           if root.docker_tags_whitelist&.any?
-            options[:include_tags] = root.docker_tags_whitelist
+            options[:include_tags] = root.docker_tags_whitelist.map(&:to_s)
           else
             options[:include_tags] = nil
           end


jlsherrill:
[test katello]

jturel:
I'm hitting an unrelated error while testing this: https://projects.theforeman.org/issues/33075

However, I can't seem to reproduce the original problem ... I created the container repo in the UI, set the docker-tags-whitelist with hammer and I can see the float via API:

```
19:17:34 rails.1   | 2021-07-20T19:17:34 [I|app|89b815ab] Started PUT "/katello/api/repositories/4" for 127.0.0.1 at 2021-07-20 19:17:34 +0000
19:17:34 rails.1   | 2021-07-20T19:17:34 [I|app|89b815ab] Processing by Katello::Api::V2::RepositoriesController#update as JSON
19:17:34 rails.1   | 2021-07-20T19:17:34 [I|app|89b815ab]   Parameters: {"docker_tags_whitelist"=>[16.1], "api_version"=>"v2", "id"=>"4", "repository"=>{"docker_tags_whitelist"=>[16.1]}}
```

The content syncing seems to work - I see 2 manifests, 1 manifest list, and 1 tag corresponding to 16.1

I must be misunderstanding the bug report :)

jlsherrill:
i cannot reproduce this!

PR #9473, opened 2021-07-15
Fixes #32739 - module stream applicability incorrect w/ enabled+inactive modules

ianballou:
See the related sub-man PR for a good description of the issue: https://github.com/candlepin/subscription-manager/pull/2708

This PR uses the proposed "active" key for the module stream profile to only import enabled+active module streams.

To test:

1) Register a CentOS 8 client
2) Sync AppStream
3) Enable the `perl:5.26` module
4) See that there are perl-IO-Socket-SSL modular packages applicable.  There should likely be no applicable modular packages.
5) Apply this PR and see the wrong applicable modular packages are still there.
  -> This tests that the PR won't break the profile upload if the sub-man fix isn't present.
6) Apply the sub-man patch, force a new profile upload, and recalculate applicability.  The wrong applicable packages should be gone now.

### DIFF
--- a/app/models/katello/concerns/host_managed_extensions.rb
+++ b/app/models/katello/concerns/host_managed_extensions.rb
@@ -214,10 +214,15 @@ module Katello
         new_ids = new_available_module_streams.keys - old_associated_ids
         new_ids.each do |new_id|
           module_stream = new_available_module_streams[new_id]
+          status = module_stream["status"]
+          # Set status to "unknown" only if the active field is in use and set to false and the module is enabled
+          if enabled_module_stream_inactive?(module_stream)
+            status = "unknown"
+          end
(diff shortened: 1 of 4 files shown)


chris1984 (approved):
ACK works great by doing the testing steps on Centos8

ianballou:
I realized the HostAvailableModuleStreams could still refer to the inactive types if they're imported by another host. I need to ensure  that these extra AvailableModuleStreams don't mess up applicability. It looks like they still are.

ianballou:
I've updated the PR to set module streams as "unknown" if they're enabled but inactive.

We'll need a workaround (likely in another PR) to fix that fact that the sub-man side of this issue won't be released in a new version until September at best.

chris1984 (approved):
ACK, working still with the updated changes. Thanks @ianballou

PR #9474, opened 2021-07-16
Fixes #33040 - Enables upgrade-all via agent actions

parthaa:


parthaa:
Steps to test:

- Have a katello-agent enabled  content host that is feeding off https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/
- `yum install walrus`
- `yum downgrade walrus`
- Now go to the Package Actions on the  Content Host in the UI
- Click on `Update All Packages`

Before PR the walrus rpm will not get updated
```bash
# rpm -q walrus
walrus-0.71-1.noarch
```
After PR 
```bash
# rpm -q walrus
walrus-5.21-1.noarch
```

app/lib/katello/agent/base_message.rb
```diff
@@ -30,6 +27,12 @@ def json
         }
       end
 
+      def options
```
jeremylenz, line 30: May be helpful to put a comment here. I could see myself wondering "why does this need to be a method?"
```suggestion
      def options # may be overridden by children
```

jeremylenz (commented):
Thanks @parthaa!

Testing steps work exactly as expected :+1: 

I don't see any test coverage for this area, might be nice to have some?

parthaa:
> I don't see any test coverage for this area, might be nice to have some?

added tests

jturel:
I think this originally should have been implemented as a separate action -> https://github.com/Katello/katello/blob/ea44c8eab2597b6234f1a4b0efbb1bd6cd7e4ffc/app/controllers/katello/api/v2/host_packages_controller.rb#L62

Notice that update and update_all share the same dynflow action. Separating those would mean there could be a 'UpgradeAllPackagesMessage'. Just sharing what a better design(in the past) would have been. Please don't change the code to do this - it's already living on borrowed time.

parthaa:
> I think this originally should have been implemented as a separate action

thats a neat idea.

parthaa:
[test katello]

test/lib/agent/update_package_message_test.rb
```diff
@@ -0,0 +1,19 @@
+require 'katello_test_helper'
+
+module Katello
+  module Agent
+    class UpdatePackageMessageTest < ActiveSupport::TestCase
+      def test_all_package_update
+        json = UpdatePackageMessage.new(content: [], consumer_id: "100000").json
+        assert json[:request][:args][1][:all]
```
jeremylenz, line 8:
```suggestion
        options = json[:request][:args][1]
        assert options[:all]
```

test/lib/agent/update_package_message_test.rb
```diff
@@ -0,0 +1,19 @@
...
+        assert_empty json[:request][:args][0][0][:unit_key]
+      end
+
+      def test_content_package_update
+        json = UpdatePackageMessage.new(content: ["foo.rpm"], consumer_id: "100000").json
+        assert_nil json[:request][:args][1][:all]
```
jeremylenz, line 14: Works fine, but might be nice to use the same variable names as the corresponding code so it's a bit easier to see what's going on:
```suggestion
        options = json[:request][:args][1]
        units = json[:request][:args][0]
        assert_nil options[:all]
        assert_equal "foo.rpm", units[0][:unit_key][:name]
```

jeremylenz (approved):
Thanks @parthaa!

I'll APJ, but also leave a small suggestion about test readability below.

PR #9475, opened 2021-07-16
Fixes #33014 - Sync pulpcore capsules after promote

sjha4:
To test:
Publish/Promote a CV to an env which is tied to a pulpcore only smart proxy.
With PR: The capsule should auto sync.
Without PR: The capsule doesn't auto sync.

### DIFF
--- a/app/models/katello/concerns/smart_proxy_extensions.rb
+++ b/app/models/katello/concerns/smart_proxy_extensions.rb
@@ -89,10 +89,19 @@ module Katello
         end
 
         def self.with_environment(environment, include_default = false)
+          (pulp2_proxies_with_environment(environment, include_default) + pulpcore_proxies_with_environment(environment)).try(:uniq)
+        end
+
+        def self.pulp2_proxies_with_environment(environment, include_default = false)
           features = [PULP_NODE_FEATURE]
           features << PULP_FEATURE if include_default
 
           unscoped.with_features(features).joins(:capsule_lifecycle_environments).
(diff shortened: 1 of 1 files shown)


app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -96,8 +96,13 @@ def self.with_environment(environment, include_default = false)
...
         end
 
+        def self.pulpcore_proxy_with_environment(environment)
+          unscoped.where(id: unscoped.select { |p| p.pulp_mirror? }.pluck(:id)).joins(:capsule_lifecycle_environments).
+              where(katello_capsule_lifecycle_environments: { lifecycle_environment_id: environment.id })
+        end
+
         def self.sync_needed?(environment)
-          Setting[:foreman_proxy_content_auto_sync] && unscoped.with_environment(environment).any?
+          Setting[:foreman_proxy_content_auto_sync] && (unscoped.with_environment(environment).any? || pulpcore_proxy_with_environment(environment).any?)
```
jlsherrill, line 105: i kinda think we should either:  a) make with_environment() check both pulpcore and pulp2, or b) rename with_environment() to something like  'pulp2_with_environment()' 

also it look like with_environment is used in a couple  other places:
```
app/lib/actions/katello/content_view/capsule_sync.rb:              smart_proxies = SmartProxy.with_environment(environment)
app/lib/actions/katello/repository/capsule_sync.rb:              smart_proxies = ::SmartProxy.with_environment(repo.environment)
```

ianballou (approved):
Works for me ! I saw my two capsules get synced after CVV promotion.

PR #9478, opened 2021-07-19
Fixes #33058 - do not error on RH ostree content removal

jlsherrill:


jlsherrill:
To test you'd need to install 3.18 (or 6.9),  enable the RH atomic repos,   upgrade to 4.0 (or 6.10), notice this upgrade task failed  ( or run this rake task and see it fail).  Apply the patch and re-try.

jlsherrill:
there's another issue i need to investigate too

jlsherrill:
this is a little tricky to test out as you need to modify the code prior to running the installer.  to reproduce:

1.  install 6.9
2. create some ostree and puppet repos, promote the ostree repo to CV
3. run 'foreman-rake katello:pulp3_content_switchover'
3. run:  foreman-maintain packages unlock
4. setup 6.10 repos
5. yum update
6. apply this patch to katello
7. run 'satellite-installer'

I think that should do it

db/migrate/20210119162528_delete_puppet_and_ostree_repos.rb
```diff
@@ -39,13 +39,23 @@ def up
     FakeContentViewPuppetEnvironment.delete_all
     FakePuppetModule.delete_all
 
-    ::Katello::Repository.delete(puppet_repositories) if puppet_repositories.any?
+    if puppet_repositories.any?
+      User.as_anonymous_admin do
+        ::Katello::Repository.delete(puppet_repositories)
+        ::Katello::RootRepository.destroy_all
```
parthaa, line 45: Is this intended ? Wouldnt you want to restrict it by type

parthaa (approved):
Works well for me

PR #9480, opened 2021-07-19
Fixes #33057 - apipie cache repo & content type params missing

ianballou:
TODO:
- [x] Write tests

The params in the repository controller cannot be based off of `RepositoryTypeManager.enabled_repository_types`.  I've changed them to be based off of `RepositoryTypeManager.defined_repository_types`.

I've also hardened our failure conditions for doing things with repo types that don't exist. I also noticed that, for uploads and removals, there is no check that the content type matches the repository type. I've fixed this.

Things to test:

1) Repo creation in the API and Hammer.  Test repo types that don't exist repo types that exist but aren't in the `enabled_repository_types`.
2) Repo uploads and deletions.   Test repo types that don't exist repo types that exist but aren't in the `enabled_repository_types`.  Also test a combination of repo type and content type that don't match ("yum" with "ostree" for example).
3) Remove your smart proxy's capabilities with ` Feature.find_by(name: "Pulpcore").smart_proxy_features.first.update(capabilities: [])` (ensure that first smart proxy feature does belong to your Pulp primary) and make sure that the API docs still show the proper repository and content type params.
  -> Don't forget to check for all of the upload-related API endpoints too like upload_content, remove_content, import_upload, etc.

### DIFF
--- a/app/controllers/katello/api/v2/content_uploads_controller.rb
+++ b/app/controllers/katello/api/v2/content_uploads_controller.rb
@@ -10,9 +10,11 @@ module Katello
     param :repository_id, :number, :required => true, :desc => N_("repository id")
     param :size, :number, :required => true, :desc => N_("Size of file to upload")
     param :checksum, String, :required => false, :desc => N_("Checksum of file to upload")
-    param :content_type, RepositoryTypeManager.uploadable_content_types.map(&:label), :required => false, :desc => N_("content type ('deb', 'docker_manifest', 'file', 'ostree', 'rpm', 'srpm')")
+    param :content_type, RepositoryTypeManager.uploadable_content_types(false).map(&:label), :required => false, :desc => N_("content type ('deb', 'docker_manifest', 'file', 'ostree', 'rpm', 'srpm')")
     def create
       content_type = params[:content_type] || ::Katello::RepositoryTypeManager.find(@repository.content_type).default_managed_content_type.label
+      RepositoryTypeManager.check_content_matches_repo_type!(@repository, content_type)
+
       unit_type_id = SmartProxy.pulp_primary.content_service(content_type).content_type
       render :json => @repository.backend_content_service(::SmartProxy.pulp_primary).create_upload(params[:size], params[:checksum], unit_type_id)
     end
--- a/app/controllers/katello/api/v2/repositories_controller.rb
+++ b/app/controllers/katello/api/v2/repositories_controller.rb
@@ -67,7 +67,7 @@ module Katello
     def_param_group :repo_create do
       param :label, String, :required => false
       param :product_id, :number, :required => true, :desc => N_("Product the repository belongs to")
-      param :content_type, RepositoryTypeManager.creatable_repository_types.keys, :required => true, :desc => N_("type of repo")
+      param :content_type, RepositoryTypeManager.creatable_repository_types(false).keys, :required => true, :desc => N_("type of repo")
     end
 
     api :GET, "/repositories", N_("List of enabled repositories")
@@ -95,7 +95,7 @@ module Katello
     param :description, String, :desc => N_("description of the repository")
     param :available_for, String, :desc => N_("interpret specified object to return only Repositories that can be associated with specified object.  Only 'content_view' & 'content_view_version' are supported."),
           :required => false
-    param :with_content, RepositoryTypeManager.enabled_content_types, :desc => N_("only repositories having at least one of the specified content type ex: rpm , erratum")
+    param :with_content, RepositoryTypeManager.enabled_content_types(false), :desc => N_("only repositories having at least one of the specified content type ex: rpm , erratum")
     param :download_policy, ::Runcible::Models::YumImporter::DOWNLOAD_POLICIES, :desc => N_("limit to only repositories with this download policy")
     param :username, String, :desc => N_("only show the repositories readable by this user with this username")
     param_group :search, Api::V2::ApiController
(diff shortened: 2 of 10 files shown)


app/controllers/katello/api/v2/repositories_controller.rb
```diff
@@ -217,8 +217,9 @@ def index_relation_content_unit(query)
     param_group :repo
     def create
       repo_params = repository_params
-      unless RepositoryTypeManager.creatable_by_user?(repo_params[:content_type])
-        msg = _("Invalid params provided - content_type must be one of %s") % RepositoryTypeManager.creatable_repository_types.keys.join(",")
+      unless RepositoryTypeManager.creatable_by_user?(repo_params[:content_type], false)
+        msg = _("Invalid params provided - content_type must be one of %s") %
+          RepositoryTypeManager.creatable_repository_types(false).keys.join(",")
```
jlsherrill, line 222: Shouldn't this validation catch this?  https://github.com/Katello/katello/blob/master/app/models/katello/root_repository.rb#L83-L87

app/controllers/katello/api/v2/repositories_controller.rb
```diff
@@ -521,8 +522,14 @@ def find_organization_from_product
...
 
     def find_content
-      if params[:content_type]
-        @content = @repository.units_for_removal(params[:ids], params[:content_type])
+      content_type = params[:content_type]
+      repo_content_types = @repository.repository_type.content_types.collect { |type| type.label }
+      unless repo_content_types.include?(content_type)
+        fail HttpErrors::BadRequest, _("Content type %{content_type} is incompatible with repositories of type %{repo_type}" %
+                                       { content_type: content_type, repo_type: @repository.content_type })
+      end
```
jlsherrill, line 530: This logic looks like its appears twice, maybe we could move it to a function in api_controller?

app/lib/actions/katello/repository/create_root.rb
```diff
@@ -3,6 +3,9 @@ module Katello
     module Repository
       class CreateRoot < Actions::EntryAction
         def plan(root)
+          unless ::Katello::RepositoryTypeManager.find(root.content_type)
+            fail _("Content type %{content_type} is not enabled." % { content_type: root.content_type })
+          end
```
jlsherrill, line 8: I think this validation will handle this right?  https://github.com/Katello/katello/blob/master/app/models/katello/root_repository.rb#L83-L87   (we could maybe change the wording there)

app/lib/actions/katello/repository/remove_content.rb
```diff
@@ -18,6 +18,12 @@ def plan(repository, content_units, options = {})
           content_unit_ids = content_units.map(&:id)
           content_unit_type = options[:content_type] || content_units.first.class::CONTENT_TYPE
 
+          repo_content_types = repository.repository_type.content_types.collect { |type| type.label }
+          unless repo_content_types.include?(content_unit_type)
+            fail _("Content type %{content_type} is incompatible with repositories of type %{repo_type}" %
+                   { content_type: content_unit_type, repo_type: repository.content_type })
+          end
```
jlsherrill, line 25: hah 3 times, maybe it should be in RepositoryTypeManager :)

app/lib/actions/katello/repository/create_root.rb
ianballou: I'll probably change the wording. I want to make sure there's a distinction between available defined types and what's enabled on the proxy.

app/controllers/katello/api/v2/content_uploads_controller.rb
```diff
@@ -10,9 +10,15 @@ class Api::V2::ContentUploadsController < Api::V2::ApiController
...
     param :size, :number, :required => true, :desc => N_("Size of file to upload")
     param :checksum, String, :required => false, :desc => N_("Checksum of file to upload")
-    param :content_type, RepositoryTypeManager.uploadable_content_types.map(&:label), :required => false, :desc => N_("content type ('deb', 'docker_manifest', 'file', 'ostree', 'rpm', 'srpm')")
+    param :content_type, RepositoryTypeManager.uploadable_content_types(false).map(&:label), :required => false, :desc => N_("content type ('deb', 'docker_manifest', 'file', 'ostree', 'rpm', 'srpm')")
     def create
       content_type = params[:content_type] || ::Katello::RepositoryTypeManager.find(@repository.content_type).default_managed_content_type.label
+      repo_content_types = @repository.repository_type.content_types.collect { |type| type.label }
+      unless repo_content_types.include?(content_type)
+        fail _("Content type %{content_type} is incompatible with repositories of type %{repo_type}" %
+               { content_type: content_type, repo_type: @repository.content_type })
```
ekohl, line 19:
```suggestion
        fail _("Content type %{content_type} is incompatible with repositories of type %{repo_type}") %
               { content_type: content_type, repo_type: @repository.content_type }
```

app/lib/actions/katello/repository/create_root.rb
```diff
@@ -3,6 +3,9 @@ module Katello
     module Repository
       class CreateRoot < Actions::EntryAction
         def plan(root)
+          unless ::Katello::RepositoryTypeManager.find(root.content_type)
+            fail _("Content type %{content_type} is not enabled." % { content_type: root.content_type })
```
ekohl, line 7:
```suggestion
            fail _("Content type %{content_type} is not enabled.") % { content_type: root.content_type }
```

app/controllers/katello/api/v2/repositories_controller.rb
ianballou: This is a little different since it's using `creatable_repository_types`.  I only changed the `creatable_repository_types` to have it use the defined types with that new `false` I'm passing in.

jlsherrill: ah, makes sense

jlsherrill:
seeing one issue:
```
#  hammer repository upload-content --id=1 --path ./foreman-2.5.0-1.el7.noarch.rpm 
Could not upload the content:
  Content type  is incompatible with repositories of type yum
```

--content-type was optional and it would default to a 'default' uploadable type (https://github.com/Katello/katello/blob/master/lib/katello/repository_types/yum.rb#L8)  i think the check needs to handle that in case its not passed in.

jlsherrill:
otherwise i this this look good

ianballou:
@jlsherrill I had a typo which caused the issue you found. Should be all good now.

jlsherrill (approved):
ack!

ianballou:
[test katello]

PR #9483, opened 2021-07-20
[CP] Fixes #33044 - re-use existing publication on sync

jlsherrill:
As part of pulp 3.14 publications can now be created during sync time
in this case, if a publication appears, we should re-use it for
distribution instead of regenerating it.


jlsherrill:
4.1 backport

jeremylenz (approved):
ACK

PR #9484, opened 2021-07-21
Fixes #33102 - Allow validate to return if disc mode is enabled

chris1984:
To test:

On your devel box or a production install do the following:

* Create an export org
* Import a manifest and sync a RH repo as immediate
* Run the export command `hammer content-export complete library --organization-id x`
* Move the export from `/var/lib/pulp/exports` to `/var/lib/pulp/imports` so that we can import
* Create a new import org
* Change the `disconnected mode` setting to `true`
* Import a different manifest
* If using a devel box, stop the server and run the firewalld commands below:
* `systemctl start firewalld`
* `firewall-cmd --direct --add-rule ipv4 filter OUTPUT 0 -p tcp -m tcp --dport=443 -j DROP`
* `firewall-cmd --direct --add-rule ipv4 filter OUTPUT 1 -j ACCEPT`
* Run the import command and see if it goes through
* `import library --organization-id x --path /var/lib/pulp/imports/xx`

Before patch we would see this:

```bash
Could not import the archive.:
a) Forbidden - server refused to process the request.
b) Request Timeout
```

With patch we should see it go through:

```bash
[vagrant@centos7-katello-devel-stable hammer-cli-katello]$ hammer content-import library --organization-id 4 --metadata-file metadata-1.json --path /var/lib/pulp/imports/Default_Organization/Export-Library/1.0/2021-07-21T17-41-17-00-00/
[.............................................................................................................................................................................................................................................................................................................................................] [100%]
```

### DIFF
--- a/app/models/katello/candlepin/repository_mapper.rb
+++ b/app/models/katello/candlepin/repository_mapper.rb
@@ -44,7 +44,7 @@ module Katello
       end
 
       def validate!
-        return if katello_content_type == Repository::OSTREE_TYPE
+        return if katello_content_type == Repository::OSTREE_TYPE || Setting[:content_disconnected]
         substitutor.validate_substitutions(content, substitutions)
       end
 
--- a/test/models/candlepin/repository_mapper_test.rb
+++ b/test/models/candlepin/repository_mapper_test.rb
@@ -33,6 +33,13 @@ module Katello
       assert_equal mapper.unprotected?, false
     end
 
+    def test_disconnected_no_http
+      Setting[:content_disconnected] = true
+      mapper = Candlepin::RepositoryMapper.new(@product_content.product, @product_content.content, {})
+      mapper.expects(:substitutor).never
+      mapper.validate!
+    end
+
     def test_download_policy
       mapper = Candlepin::RepositoryMapper.new(@product_content.product, @product_content.content, {})
 


lfu:
Test passed 👍🏻 

**Before**
```
[DEBUG 2021-07-23T14:08:47 Exception] Using exception handler HammerCLIKatello::ExceptionHandler#handle_bad_request
[ERROR 2021-07-23T14:08:47 Exception] Request Timeout
Could not import the archive.:
  Request Timeout
[ERROR 2021-07-23T14:08:47 Exception] 

RestClient::BadRequest (400 Bad Request):
...
    /home/vagrant/.gem/ruby/gems/hammer_cli-2.5.1/lib/hammer_cli/apipie/command.rb:53:in `send_request'
```
**After**
```
14:23:48 rails.1   | 2021-07-23T14:23:48 [I|bac|93ee8c8c] Task {label: Actions::Katello::ContentViewVersion::ImportLibrary, id: d430a5fc-d090-4202-8332-598a10eb399b, execution_plan_id: c7876e3a-439d-4254-8bf9-da6bbc8c74b7} state changed: stopped  result: success

[DEBUG 2021-07-23T14:25:15 API] Response: {
                   "id" => "d430a5fc-d090-4202-8332-598a10eb399b",
                "label" => "Actions::Katello::ContentViewVersion::ImportLibrary",
              "pending" => false,
               "action" => "Import Default Content View organization 'Default Organization'",
             "username" => "admin",
...
}
[DEBUG 2021-07-23T14:25:15 API] Response headers: {
                      :foreman_version => "2.6.0-develop",
                  :foreman_api_version => "2",
         :foreman_current_organization => "; ANY",
             :foreman_current_location => "; ANY",
...

                    :transfer_encoding => "chunked"
}
[.....................................................................................................................................................................................] [100%]
```

lfu (approved):

jturel:
In the before case, I'd like to know the stack trace of where the error was raised from. The reason is that I feel the RepositoryMapper knowing about disconnected mode is a bit odd, and I wonder if we can do something earlier in the import/export part of the code (which to me, makes more sense to be aware of 'disconnected'). Trying to think beyond "this fixes the error" and get to "is this the correct fix for the error" :)

lfu:
[Back trace.](http://pastebin.test.redhat.com/981857)

jturel:
Here's the interesting part of the stack trace:

```
18:27:52 rails.1   | 2021-07-23T18:27:52 [E|bac|9bee76da] Request Timeout (RestClient::RequestTimeout)
18:27:52 rails.1   |  9bee76da | /home/vagrant/katello/app/lib/katello/resources/cdn.rb:118:in `rescue in get'
18:27:52 rails.1   |  9bee76da | /home/vagrant/katello/app/lib/katello/resources/cdn.rb:98:in `get'
18:27:52 rails.1   |  9bee76da | /home/vagrant/katello/app/lib/katello/util/cdn_var_substitutor.rb:79:in `valid_path?'
18:27:52 rails.1   |  9bee76da | /home/vagrant/katello/app/lib/katello/util/cdn_var_substitutor.rb:51:in `block in any_valid_metadata_file?'
18:27:52 rails.1   |  9bee76da | /home/vagrant/katello/app/lib/katello/util/cdn_var_substitutor.rb:51:in `any?'
18:27:52 rails.1   |  9bee76da | /home/vagrant/katello/app/lib/katello/util/cdn_var_substitutor.rb:51:in `any_valid_metadata_file?'
18:27:52 rails.1   |  9bee76da | /home/vagrant/katello/app/lib/katello/util/cdn_var_substitutor.rb:43:in `validate_substitutions'
18:27:52 rails.1   |  9bee76da | /home/vagrant/katello/app/models/katello/candlepin/repository_mapper.rb:48:in `validate!'
18:27:52 rails.1   |  9bee76da | /home/vagrant/katello/app/lib/actions/katello/repository_set/enable_repository.rb:13:in `plan'
18:27:52 rails.1   |  9bee76da | /home/vagrant/foreman/.vendor/ruby/2.7.0/gems/dynflow-1.5.0/lib/dynflow/action.rb:515:in `block (3 levels) in execute_plan'
```

You can see it originates from the `Actions::Katello::RepositorySet::EnableRepository` action which is also called by the Repository Sets API (think Red Hat Repositories page). So, this change is affecting not only the import/export use case but the repo sets API too.

What does this change do? It completely short-circuits the validation of whether the repository set we are enabling exists on the Red Hat CDN (if I am reading the code correctly...) and is a valid repo when in disconnected mode. The question that comes from that is: do we want to do that for the entire repository sets API, or just the import/export use case?

lfu:
Can a disconnected server communicate to Red Hat CDN and enable repositories? @jturel

jturel:
> Can a disconnected server communicate to Red Hat CDN and enable repositories? @jturel

If someone simply sets the Disconnected setting to true there is nothing stopping outside communication. Users can take it upon themselves (totally separate from the setting) to disable outgoing network traffic on their box which I think is valid if they want to be in the disconnected use case.

chris1984:
@jturel what kind of fix are you thinking about to add to this pr?

jturel:
> @jturel what kind of fix are you thinking about to add to this pr?

Not proposing any change but highlighting that this pr has repercussions broader than one might believe by just looking at the code, and for those involved in the review here to be aware of that in case it raises any concerns, etc.

app/models/katello/candlepin/repository_mapper.rb
```diff
@@ -44,7 +44,7 @@ def build_repository
       end
 
       def validate!
-        return if katello_content_type == Repository::OSTREE_TYPE
+        return if katello_content_type == Repository::OSTREE_TYPE || Setting[:content_disconnected]
```
jturel, line 47: The repository mapper should have a test which asserts "validate! does not issue any http requests when in disconnected mode"

jturel:
cc @parthaa since this change was proposed (or at least identified) by him!

app/models/katello/candlepin/repository_mapper.rb
parthaa: Agree. That should be simple to add

chris1984: Sounds good, will add a test for this scenario

test/models/candlepin/repository_mapper_test.rb
```diff
@@ -33,6 +33,13 @@ def test_unprotected_suse
       assert_equal mapper.unprotected?, false
     end
 
+    def test_disconnected_no_http
+      Setting[:content_disconnected] = true
+      mapper = Candlepin::RepositoryMapper.new(@product_content.product, @product_content.content, {})
+
+      assert_nil mapper.validate!
```
parthaa, line 40:
```suggestion
       mapper.expects(:substitutor).never
      assert_nil mapper.validate!
```

test/models/candlepin/repository_mapper_test.rb
```diff
@@ -33,6 +33,13 @@ def test_unprotected_suse
       assert_equal mapper.unprotected?, false
     end
 
+    def test_disconnected_no_http
```
jturel, line 36: This test doesn't actually execute the changed method

chris1984: If I change the disconnected setting to try I get a vcr error though

jturel: Take a look at the stack trace and see where from / why the HTTP request is going out. Maybe there's another bug?

chris1984: Oh I see, I need to call the `validate!` method in my test. Sorry after reading your comment and looking at it again, I see what needs updated.

parthaa: The test updates lgtm. If @lfu already tested your changes, I am ok acking

chris1984: @parthaa she did here: https://github.com/Katello/katello/pull/9484#issuecomment-885687772

parthaa (approved):

PR #9485, opened 2021-07-21
Refs #33044 - properly check publication_href on SP sync

jlsherrill:


ianballou (approved):
I was able to sync a smart proxy!

PR #9488, opened 2021-07-22
Registration - Remove 'Foreman' from error msg

stejskalleos:


jjeffers (approved):
LGTM

app/controllers/katello/concerns/api/v2/registration_controller_extensions.rb
```diff
@@ -7,7 +7,7 @@ def prepare_host
         if params['uuid']
           @host = Katello::Host::SubscriptionFacet.find_by(uuid: params['uuid'])&.host
           if @host.nil?
-            msg = N_("Host was not found by the subscription UUID: '%s', this can happen if the host is registered already, but not to this Foreman") % params['uuid']
+            msg = N_("Host was not found by the subscription UUID: '%s', this can happen if the host is registered already, but not to this instance") % params['uuid']
             fail ActiveRecord::RecordNotFound, msg
```
jlsherrill, line 11: i suspect this is not the best way to solve this.  Any mention of Foreman in a translated string should be converted to satellite as part of the downstream branding plugin (via i18n).  I think this is not properly marked for translation adn should be using _() instead of N_()

jlsherrill: i opened a pr here:  https://github.com/Katello/katello/pull/9512

i left 'instance' there, but i think that change to `_()` would have resolved it with the magic that happens in the downstream branding plugin

PR #9489, opened 2021-07-22
Fixes #32509 - Promote Content Views

sjha4:
- [X] UI skeleton
- [x] Connect everything to Redux
- [x] Task presenter - https://github.com/Katello/katello/pull/9494 (Not a clean rebase with this PR unfortunately)
- [x] Tests

### DIFF
--- a/webpack/scenes/ContentViews/Details/Promote/ContentViewVersionPromote.js
+++ b/webpack/scenes/ContentViews/Details/Promote/ContentViewVersionPromote.js
@@ -1,0 +1,163 @@
+import React, { useState, useMemo, useCallback } from 'react';
+import useDeepCompareEffect from 'use-deep-compare-effect';
+import { useDispatch, useSelector } from 'react-redux';
+import { STATUS } from 'foremanReact/constants';
+import PropTypes from 'prop-types';
+import { Form, FormGroup, TextArea, ActionGroup, Button, Modal, ModalVariant, Alert } from '@patternfly/react-core';
+import { translate as __ } from 'foremanReact/common/I18n';
+
+import {
+  selectEnvironmentPaths,
+  selectEnvironmentPathsStatus,
+} from '../../components/EnvironmentPaths/EnvironmentPathSelectors';
+import EnvironmentPaths from '../../components/EnvironmentPaths/EnvironmentPaths';
+import { getContentViewVersions, promoteContentViewVersion } from '../ContentViewDetailActions';
+import {
+  selectPromoteCVVersionError, selectPromoteCVVersionResponse,
+  selectPromoteCVVersionStatus,
+} from './ContentViewVersionPromoteSelectors';
+import ComponentEnvironments from '../ComponentContentViews/ComponentEnvironments';
+import Loading from '../../../../components/Loading';
+
+const ContentViewVersionPromote = ({
+  cvId, versionIdToPromote, versionNameToPromote, versionEnvironments, setIsOpen,
+}) => {
+  const [description, setDescription] = useState('');
+  const [userCheckedItems, setUserCheckedItems] = useState([]);
+  const [promoting, setPromoting] = useState(false);
+  const [forcePromote, setForcePromote] = useState([]);
+  const environmentPathResponse = useSelector(selectEnvironmentPaths);
+  const environmentPathStatus = useSelector(selectEnvironmentPathsStatus);
+  const environmentPathLoading = environmentPathStatus === STATUS.PENDING;
+  const promoteResponse = useSelector(state =>
+    selectPromoteCVVersionResponse(state, versionIdToPromote, versionEnvironments));
+  const promoteStatus = useSelector(state =>
+    selectPromoteCVVersionStatus(state, versionIdToPromote, versionEnvironments));
+  const promoteError = useSelector(state =>
+    selectPromoteCVVersionError(state, versionIdToPromote, versionEnvironments));
+  const promoteResolved = promoteStatus === STATUS.RESOLVED;
+  const dispatch = useDispatch();
+
+  const onPromote = () => {
+    setPromoting(true);
+    dispatch(promoteContentViewVersion({
+      id: versionIdToPromote,
+      description,
+      versionEnvironments,
+      environment_ids: userCheckedItems.map(item => item.id),
+      force: true,
+    }));
+  };
+
+  useDeepCompareEffect(() => {
+    if (promoteResolved && promoteResponse) {
+      setIsOpen(false);
+      dispatch(getContentViewVersions(cvId));
+    }
+    if (promoteError) {
+      setPromoting(false);
+    }
+  }, [promoteResponse, promoteResolved, promoteError, setPromoting, setIsOpen, dispatch, cvId]);
+
+  const envPathFlat = useMemo(() => {
+    if (!environmentPathLoading) {
+      const { results } = environmentPathResponse || {};
+      return results.map(result => result.environments).flatten();
+    }
+    return [];
+  }, [environmentPathResponse, environmentPathLoading]);
+
+  const prior = useCallback(
+    env => envPathFlat.find(item => item.id === env.prior.id),
+    [envPathFlat],
+  );
+
+  const isChecked = useCallback(
+    env => (userCheckedItems.includes(env) ||
+      versionEnvironments.filter(item => item.id === env.id).length),
+    [userCheckedItems, versionEnvironments],
+  );
+
+  const isValid = useCallback((env) => {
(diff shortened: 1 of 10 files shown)


webpack/scenes/ContentViews/components/EnvironmentPaths/EnvironmentPaths.js
```diff
@@ -41,9 +43,11 @@ const EnvironmentPaths = ({ userCheckedItems, setUserCheckedItems }) => {
             <FormGroup key={`fg-${count}`} isInline fieldId="environment-checkbox-group">
               {environments.map(env =>
                 (<Checkbox
-                  isChecked={env.library ||
-                  userCheckedItems.filter(item => item.id === env.id).length}
-                  isDisabled={env.library}
+                  isChecked={(publishing && env.library) ||
```
Andrewgdewar, line 46: Can we create a variablefor these: 
 promotedEnvironments.filter(item => item.id === env.id).length

It's repeated twice and for outside readability, if it was named something like const hasPromotedEnvironment?
Thoughts?

webpack/scenes/ContentViews/Details/Versions/ContentViewVersions.js
```diff
@@ -113,6 +133,30 @@ const ContentViewVersions = ({ cvId }) => {
...
+    setVersionIdToPromote(cvVersionId);
+    setVersionNameToPromote(cvVersionName);
+    setVersionEnvironments(cvVersionEnvironments);
+    setPromoting(true);
+  };
+
+  const actionResolver = () => [
+    {
+      title: __('Promote'),
+      onClick: (_event, rowId, rowInfo) => {
```
Andrewgdewar, line 146: do we need to check on anything here before promoting?

webpack/scenes/ContentViews/Details/Versions/ContentViewVersions.js
```diff
@@ -137,7 +182,19 @@ const ContentViewVersions = ({ cvId }) => {
       autocompleteEndpoint={`/content_view_versions/auto_complete_search?content_view_id=${cvId}`}
       fetchItems={useCallback(params => getContentViewVersions(cvId, params), [cvId])}
       additionalListeners={[pollingFinished]}
-    />);
+    >
+      {promoting &&
```
Andrewgdewar, line 186: this seems redundant: perhaps instead of conditionally showing it here, you could just do a check on the ContentViewVersion like: 
if (!show) return null;
at the top of the functional scope. 

You could then remove the conditional {promoting && <thing/>} logic.

webpack/scenes/ContentViews/components/EnvironmentPaths/EnvironmentPaths.js
jturel: +1 it would communicate the intent of the code much more clearly

webpack/scenes/ContentViews/Details/Promote/ContentViewVersionPromote.js
```diff
@@ -0,0 +1,165 @@
...
+    <Modal
+      title={__(`Promote version ${versionNameToPromote}`)}
+      isOpen={show}
+      variant={ModalVariant.large}
+      onClose={() => {
+        setIsOpen(false);
+      }}
+      appendTo={document.body}
+    >
+      {saving && <Loading loadingText={__('Please wait while task starts..')} />}
```
jturel, line 100: Would naming this variable 'promoting' instead of 'saving' be more clear or accurate?

webpack/scenes/ContentViews/Details/Promote/ContentViewVersionPromote.js
```diff
@@ -0,0 +1,165 @@
...
+            setUserCheckedItems={setUserCheckedItems}
+            promotedEnvironments={versionEnvironments}
+            publishing={false}
+          />
+          <Form>
+            <ActionGroup>
+              <Button
+                aria-label="promote_content_view"
+                variant="primary"
+                isDisabled={saving}
```
jturel, line 135: I'm able to press the Promote button even if the version has been promoted to all available environments. It seems like that should be disabled and perhaps an 'empty state' could be applied to this component "This version is already published to all available environments" or something like that

webpack/scenes/ContentViews/Details/Promote/ContentViewVersionPromote.js
jturel, line 100: s/while task starts/while the task starts

jturel (commented):
This is working really well. Nice work keeping the scope of this PR exactly what is described in the PR title!

Since it's working so nicely my feedback is more on the nit-pick side :)

webpack/scenes/ContentViews/Details/Versions/ContentViewVersions.js
sjha4: The conditional return null at top of functional scope makes react hook rules unhappy. It would have to go below state, useEffects and other hooks in the file. Given that, I think we should stick to the pattern of conditional rendering..Thoughts?

`React Hook "useState" is called conditionally. React Hooks must be called in the exact same order in every component render`

webpack/scenes/ContentViews/components/EnvironmentPaths/EnvironmentPaths.js
sjha4: Updated to use method for these checks..

webpack/scenes/ContentViews/Details/Promote/ContentViewVersionPromote.js
Andrewgdewar: +1 This was discussed, I think we decided to add this view to a separate ticket as not to bloat this one? @sjha4

webpack/scenes/ContentViews/Details/Versions/ContentViewVersions.js
Andrewgdewar: As discussed, we'll likely be changing our modal pattern to: 
https://github.com/Katello/katello/pull/9553/files#diff-dc6c9c6fb2119021430864e20e82157272fb9a2a55accf48285743eb8781353fR66
https://github.com/Katello/katello/pull/9553/files#diff-04ef9c1add8b6ffd04659bb6b4e9c32bc4bea75f751dd4d392573ebe09981d2dR97-R101

webpack/scenes/ContentViews/Details/Promote/ContentViewVersionPromote.js
```diff
@@ -0,0 +1,165 @@
...
+      title={__(`Promote version ${versionNameToPromote}`)}
+      isOpen={show}
+      variant={ModalVariant.large}
+      onClose={() => {
+        setIsOpen(false);
+      }}
+      appendTo={document.body}
+    >
+      {promoting && <Loading loadingText={__('Please wait while the task starts..')} />}
+      {!promoting &&
```
Andrewgdewar, line 101: {promoting ? thing1 : thing2}

webpack/scenes/ContentViews/Details/Promote/ContentViewVersionPromote.js
jturel: Totally works for me as long as there's a Redmine filed for it

webpack/scenes/ContentViews/Details/Versions/ContentViewVersions.js
Andrewgdewar: The answer is no, as we don't have the API's data prior to hitting this button to know if it is disabled or not.

webpack/scenes/ContentViews/Details/Promote/ContentViewVersionPromote.js
sjha4: Since the `thing2` is a big JSX, I think the current form is more readable than having it inside a ternary operator?

webpack/scenes/ContentViews/Details/Promote/ContentViewVersionPromote.js
sjha4: This would be a nice to have addition. Opened https://projects.theforeman.org/issues/33275 for this.

webpack/scenes/ContentViews/Details/Promote/ContentViewVersionPromote.js
sjha4: Updated to use ternary. 👍🏽

Andrewgdewar (approved):
ACK!

sjha4:
Merging! Thanks @Andrewgdewar and @jturel !

PR #9490, opened 2021-07-22
Fixes #33120 - Expose ability to sync individual content view and repository to smart proxy

phess:
The Actions::Katello::CapsuleContent::Sync task is already capable of syncing only an individual Lifecycle Environment and Content View and Repository to an individual Smart Proxy.
However, the Katello API exposes only the ability to specify an individual Lifecycle Environment.
It would be nice to expose the Content View and Repository capabilities through the API as well.

Next step once this is accepted: add hammer options to allow passing CV and repos to capsule syncs.


chris1984:
ok to test

chris1984 (changes requested):
ACK, works great @phess 

Here is a link to the unit tests I made if you want to add them to your pr so we have test coverage around the new params.

https://gist.github.com/chris1984/8a95d15dae9b9bc733a8ea523706cd32

Sorry for the long delay in getting back to you, once you add them I will ack it and merge.

PR #9494, opened 2021-07-26
Fixes #33118 - Task presenter for publish/promote

sjha4:
To test:
On the new UI:
Publish a CV from the versions page and close the task presenter final page on the Publish wizard as soon as task starts.

### DIFF
--- a/webpack/scenes/ContentViews/Details/Versions/ContentViewVersions.js
+++ b/webpack/scenes/ContentViews/Details/Versions/ContentViewVersions.js
@@ -1,6 +1,6 @@
 import React, { useState, useCallback } from 'react';
 import useDeepCompareEffect from 'use-deep-compare-effect';
-import { useSelector } from 'react-redux';
+import { useDispatch, useSelector } from 'react-redux';
 import { TableVariant, TableText } from '@patternfly/react-table';
 import { translate as __ } from 'foremanReact/common/I18n';
 import { urlBuilder } from 'foremanReact/common/urlHelpers';
@@ -18,12 +18,16 @@ import {
   selectCVVersionsStatus,
   selectCVVersionsError,
 } from '../ContentViewDetailSelectors';
+import TaskPresenter from '../../components/TaskPresenter/TaskPresenter';
+import { startPollingTask } from '../../../Tasks/TaskActions';
 
 const ContentViewVersions = ({ cvId }) => {
   const response = useSelector(state => selectCVVersions(state, cvId));
   const status = useSelector(state => selectCVVersionsStatus(state, cvId));
   const error = useSelector(state => selectCVVersionsError(state, cvId));
+  const [pollingFinished, setPollingFinished] = useState(false);
   const loading = status === STATUS.PENDING;
+  const dispatch = useDispatch();
 
   const [rows, setRows] = useState([]);
   const [metadata, setMetadata] = useState({});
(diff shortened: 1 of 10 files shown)


webpack/scenes/ContentViews/Publish/CVPublishFinish.js
```diff
@@ -113,46 +115,48 @@ const CVPublishFinish = ({
...
-              {' View task details '}
-              <ExternalLinkAltIcon />
-            </Button>
-          </Bullseye>
-        </GridItem>
-      </Grid>
+                variant="primary"
+                aria-label="publish_content_view"
+              >
+                Close
```
Andrewgdewar, line 144: __()

webpack/scenes/ContentViews/Publish/CVPublishFinish.js
```diff
@@ -113,46 +115,48 @@ const CVPublishFinish = ({
...
+              </Button>
+              <Button
+                component="a"
+                aria-label="view tasks button"
+                href={`/foreman_tasks/tasks/${pollResponse.id}`}
+                target="_blank"
+                variant="link"
+              >
+                {' View task details '}
```
Andrewgdewar, line 153: __()

webpack/scenes/ContentViews/Details/Versions/ContentViewVersions.js
```diff
@@ -38,29 +42,61 @@ const ContentViewVersions = ({ cvId }) => {
...
+    const {
+      version,
+      description,
+      id: versionId,
+      environments,
+      rpm_count: packageCount,
+      errata_counts: errataCounts,
+    } = cvVersion;
+    return [
+      { title: <a href={urlBuilder(`content_views/${cvId}/versions/${versionId}`, '')}>{`Version ${version}`}</a> },
```
Andrewgdewar, line 55: __()

webpack/scenes/ContentViews/Details/Versions/ContentViewVersions.js
```diff
@@ -38,29 +42,61 @@ const ContentViewVersions = ({ cvId }) => {
...
+      version,
+      description,
+      id: versionId,
+      active_history: activeHistory,
+    } = cvVersion;
+    const { task } = activeHistory[0];
+    dispatch(startPollingTask(task.id, task));
+
+    return [
+      { title: <a href={urlBuilder(`content_views/${cvId}/versions/${versionId}`, '')}>{`Version ${version}`}</a> },
```
Andrewgdewar, line 75: Ditto

webpack/scenes/ContentViews/Details/Versions/ContentViewVersions.js
```diff
@@ -38,29 +42,64 @@ const ContentViewVersions = ({ cvId }) => {
     __('Description'),
   ];
 
-
   useDeepCompareEffect(() => {
+    const buildCells = (cvVersion) => {
```
jeremylenz, line 46: You've ended up with an absolutely enormous effect here.  What are your thoughts on using `useCallback` instead? That would allow you to move `buildCells`, `buildActiveTaskCells`, and `buildRows` back outside this effect, and make this file a little more readable.

Having done this several dozen times in my work on the last PR, the procedure I've gotten in the habit of using is

1. Move the function outside the effect. `eslint` will tell you
```
React Hook useDeepCompareEffect has a missing dependency: 'buildCells'. Either include it or remove the dependency array. (react-hooks/exhaustive-deps)
```
2. Do as it says, and add `buildCells` to the array. Now, `eslint` says
```
The 'buildCells' function makes the dependencies of useDeepCompareEffect Hook (at line 116) change on every render. Move it inside the useDeepCompareEffect callback. Alternatively, wrap the definition of 'buildCells' in its own useCallback() Hook. (react-hooks/exhaustive-deps)
```
3. We're going to choose the latter option. Add `useCallback`:
```js
  const buildCells = useCallback((cvVersion) => {
    ...
  });
```
4. Now `eslint` will tell you
```
React Hook useCallback does nothing when called with only one argument. Did you forget to pass an array of dependencies? (react-hooks/exhaustive-deps)
```
5. Add `[]` to the end of the function:
```js
  }, []);
```
Now, `eslint` complains again and tells you exactly what is needed:
```
React Hook useCallback has a missing dependency: 'cvId'. Either include it or remove the dependency array. (react-hooks/exhaustive-deps)
```
6. Add `cvId` and you're good to go!

It's really not too bad once you get used to it ;)

jeremylenz (commented):
Thanks @sjha4 

The new toast notifications are absolutely gorgeous!!

Nothing happens when I close the modal right after starting the task. Is it supposed to pop back up? I just see the new toasts when it's done. But if I leave the modal up, the new progress bar looks and works as expected.

webpack/scenes/ContentViews/Details/Versions/ContentViewVersions.js
sjha4: > Nothing happens when I close the modal right after starting the task.

Ya..If you close the modal before the publish API returns a task response and the modal displays the loading spinner, nothing happens..Andrew and I were discussing about this and probably giving a warning below Loading like Please do not close or something but it seems unnecessary cause in the worst case if you do close it when loading, the page needs a refresh manually. 

Working on the useCallback.. I was waiting for you to come tell me what to do to avoid making the effect so huge.. :smile:

jeremylenz: > in the worst case if you do close it when loading, the page needs a refresh manually.

On the Subscriptions page we always show a progress bar if there's a manifest-related task in progress. And it will pop up even if you navigate to the page from somewhere else. 

We could do something similar here, but instead of having a persistent noncloseable progress bar we could just use the polling to automatically refresh the table when the task is done.  Just a thought..

sjha4: The task progress on the table will show up if you navigate from someplace else that involves a refresh of the rows..That happens also when you close the wizard from the task presenter page..

However the case where the modal is closed when the publish call hasn't returned a response is when reloading the table doesn't help cause publish method creates actual version fairly late in the planning phase..And if you make a call to versions API while publish is still in planning, it returns the versions list without the new version that is yet to be created and the task that is yet to be started.

jeremylenz (approved):
Changes look good

Thanks @sjha4!

PR #9495, opened 2021-07-26
Feature#32912

rverdile:
Can test removing content units through hammer or API.

Can test getting generic content units from API by making GET request to "generic". For example:
`curl --user admin:changeme https://centos7-katello-devel-nfs.localhost.example.com/katello/api/generic`

The other potential content unit API requests are shown on the apidoc.

Upload adds new a param to the controller that specifies "generic_content_type", which for Python would be "python_package". To test that, need to make an API call to upload a file.

Example of uploading to repository API:

`require 'rest-client'`

`request = RestClient::Request.new(
          :method => :post,
          :url => 'https://centos7-katello-devel-nfs.localhost.example.com/katello/api/repositories/1/upload_content',
          :user => 'admin',
          :password => 'changeme',
          :payload => {
            :multipart => true,
            :content => [File.new("notificationcenter-1.0.0b2.whl", 'rb')],
            :content_type => 'generic',
            :generic_content_type => 'python_package'
          })`

`request.execute`


ianballou (commented):
The first thing I noticed was that the generic endpoint for listing units doesn't return the version.  We'll want that to be included.

I tested the basic functionality for everything else though and it's looking good! I need to do a code review to dig more into it.

app/controllers/katello/api/v2/content_uploads_controller.rb
```diff
@@ -11,12 +11,16 @@ class Api::V2::ContentUploadsController < Api::V2::ApiController
     param :size, :number, :required => true, :desc => N_("Size of file to upload")
     param :checksum, String, :required => false, :desc => N_("Checksum of file to upload")
     param :content_type, RepositoryTypeManager.uploadable_content_types(false).map(&:label), :required => false, :desc => N_("content type ('deb', 'docker_manifest', 'file', 'ostree', 'rpm', 'srpm')")
+    param :generic_content_type, String, :required => false, :desc => N_("Content type for generic content unit")
     def create
       content_type = params[:content_type] || ::Katello::RepositoryTypeManager.find(@repository.content_type).default_managed_content_type.label
-      RepositoryTypeManager.check_content_matches_repo_type!(@repository, content_type)
-
+      RepositoryTypeManager.check_content_matches_repo_type!(@repository, content_type) unless content_type == "generic"
```
ianballou, line 17: After some looking, I think we could fix `check_content_matches_repo_type!` to make it work for generic content types.

app/services/katello/pulp3/content.rb
```diff
@@ -4,12 +4,18 @@ module Pulp3
...
       class << self
-        def create_upload(size = 0, checksum = nil, content_type = nil)
+        def create_upload(size = 0, checksum = nil, content_type = nil, generic_content_type = nil, repository = nil)
           content_unit_href = nil
           if checksum
             content_backend_service = SmartProxy.pulp_primary.content_service(content_type)
-            content_list = content_backend_service.content_api.list("sha256": checksum)
-            content_unit_href = content_list.results.first.pulp_href unless content_list.results.empty?
+            if repository&.generic?
+              # pulp python doesn't support filter by sha256 (i.e. list method), so need to get the result differently
```
ianballou, line 12: I'm surprised the filtering doesn't work since each package does have a sha256 hash.  I think we should make a feature request: https://github.com/pulp/pulp_python/issues

rverdile: It looks like they added filtering for sha256 after I made the first issue. I guess this code will have to be updated whenever the next version of pulp python is released/supported.

ianballou: Let's make an issue for it after this PR merges.

rverdile:
I'm wondering if there are places I could replace passing "generic" for `content_type` or `unit_type_id` with the equivalent of "python_package", rather than creating a new parameter - usually `generic_content_type` - for the equivalent of "python_package".

For example, in `check_content_matches_repo_type!` in `content_uploads_controller.rb` I can pass `param[:generic_content_unit]` for the `content_type` parameter when the repository is generic. I wonder if there are other places where this is a better approach than creating a new parameter. Thoughts?

ianballou (commented):
From reading through the code, I feel like it would be beneficial to remove the `:generic_content_type` param.  If we could just rely on the content type, it would be a better user experience.  You might just need to make some sort of helper function to see if a content type is a generic one or not. At that point you should be able to replace all uses of the extra parameter.

Let me know if this seems reasonable or if it ends up looking weirdly complicated.

rverdile:
> From reading through the code, I feel like it would be beneficial to remove the `:generic_content_type` param. If we could just rely on the content type, it would be a better user experience. You might just need to make some sort of helper function to see if a content type is a generic one or not. At that point you should be able to replace all uses of the extra parameter.
> 
> Let me know if this seems reasonable or if it ends up looking weirdly complicated.

I think this makes sense and might even be less complicated. I was actually just gonna leave a comment with a similar thought :)

app/services/katello/repository_type_manager.rb
```diff
@@ -79,6 +79,19 @@ def indexable_content_types
...
+        repo_types = enabled_only ? enabled_repository_types : defined_repository_types
+        list = repo_types.values.map do |type|
+          type.content_types.select { |ct| ct.model_class::CONTENT_TYPE == Katello::GenericContentUnit::CONTENT_TYPE }
+        end
+        list.flatten
+      end
+
+      def generic_content_type?(content_type, enabled_only = true)
+        types = generic_content_types(enabled_only)
+        !types.select { |ct| ct.content_type == content_type }.empty?
```
ianballou, line 92: I think you could just do `types.include?(content_type)` if you changed `generic_content_types` to return a list of content type labels.

app/services/katello/repository_type_manager.rb
```diff
@@ -79,6 +79,19 @@ def indexable_content_types
                   flatten
       end
 
+      def generic_content_types(enabled_only = true)
+        repo_types = enabled_only ? enabled_repository_types : defined_repository_types
+        list = repo_types.values.map do |type|
+          type.content_types.select { |ct| ct.model_class::CONTENT_TYPE == Katello::GenericContentUnit::CONTENT_TYPE }
+        end
+        list.flatten
```
ianballou, line 87: If you do change it to return just the strings, you could change this to:
```ruby
list.flatten.map(&:content_type)
```

rverdile:
Not sure what the uses are for the content units APIs, but I'm wondering if the generic content units API should instead be something like `/katello/api/generic/python_package` or `/katello/api/python_package`. Created generically, of course.

jlsherrill:
you can use generic matches like this (from the rhsm routes file):
```
  match '/consumers/:id' => 'candlepin_proxies#consumer_show', :via => :get
```

which transforms :id into a param.  So something like:

match 'content_types/:content_type/' => 'generic_contents#index', :via => :get

within v2.rb of the routes should allow you to pass in any content_type via the url.  We'd have to define each route manually, but that might be kinda nice.

app/lib/actions/pulp3/orchestration/repository/upload_content.rb
```diff
@@ -7,8 +7,14 @@ class UploadContent < Pulp3::Abstract
           def plan(repository, smart_proxy, file, unit_type_id)
             sequence do
               content_backend_service = SmartProxy.pulp_primary.content_service(unit_type_id)
-              content_list = content_backend_service.content_api.list("sha256": Digest::SHA256.hexdigest(File.read(file[:path])))
-              content_href = content_list&.results&.first&.pulp_href
+              if repository.generic?
+                # pulp python doesn't support filter by sha256 (i.e. list method), so need to get the result differently
+                content_list = content_backend_service.content_api(repository.repository_type, unit_type_id).list.results.find { |result| result.sha256 == Digest::SHA256.hexdigest(File.read(file[:path])) }
```
jlsherrill, line 12: I don't think this will work (or at least won't work if there is more than a page worth of python modules).

It looks like this was now added:  https://github.com/pulp/pulp_python/commit/aeb205db4a9cf02a6c65674894832cb5bad5295e

maybe we could just go ahead and patch locally for testing and ask them for a release?

app/services/katello/pulp3/content.rb
jlsherrill: i mentioned this above, but i think we should just go ahead and use the new feature before merging this since this won't always work, open to other thoughts though

app/views/katello/api/v2/generic_content_units/base.json.rabl
```diff
@@ -0,0 +1,5 @@
+object @resource
+
+attributes :id
+attributes :name
+attributes :pulp_id
```
jlsherrill, line 5: isn't there a version attribute too?

app/lib/actions/pulp3/orchestration/repository/upload_content.rb
rverdile: There's an optional parameter for specifying the amount of results. I can't find what happens if that number isn't specified, but it isn't necessarily all of them. But yeah I can change it to depend on that change.

lib/katello/repository_types/python.rb
```diff
@@ -34,4 +34,5 @@
                        pulp3_service_class: Katello::Pulp3::GenericContentUnit,
                        removable: true,
                        uploadable: true
+  default_managed_content_type Katello::GenericContentUnit
```
jlsherrill, line 37: Let me explain the history of this argument.  Originally there was no way for the user to pass in a content type to the api to remove content or upload content.  The user would just say for repository x, remove content Y.  This meant that only one type of content was supported for removal/uploading.  We added the ability to pass in the type, but needed to keep api compatibility so we added this option to specify the default if no type is passed in. 

So here, i don't think passing in Katello::GenericContentUnit makes sense, consider the possibility of a repo having 2 different content units.    We may need to modify this method to take a label instead of a class, so now we'd have:
```
default_managed_content_type :python_package
```
we'd need to update that method to handle it, and probably modify the other instances of it to use the symbol instead of the class

rverdile:
I updated `default_managed_content_type` to take a label as its input instead of a class.

I changed the API URLs for generic content units. _Still need to figure out how update the apidoc and the test now that it's different._

```
/katello/api/content/ ==> list all generic content units
/katello/api/content/python_package ==> list all generic content units with content type "python_package"

/katello/api/repositories/content/python_package ==> list all generic content units with content type "python_package"
/katello/api/repositories/1/content ==> list all generic content units in repository 1
/katello/api/repositories/1/content/python_package ==> list all generic content units with content type "python_package" in repository 
```
They also return the `version` and `content_type` attributes now.

The sha256 checks in `content.rb` and `upload_content.rb` are changed to work with the sha256 filtering. To work around it for testing you can replace these blocks

content.rb
```
if checksum
            content_backend_service = SmartProxy.pulp_primary.content_service(content_type)
            if repository&.generic?
              # pulp python doesn't support filter by sha256 (i.e. list method), so need to get the result differently
              content_list = content_backend_service.content_api(repository.repository_type, content_type).list.results.find { |result| result.sha256 == checksum }
              content_unit_href = content_list&.pulp_href
            else
              content_list = content_backend_service.content_api.list("sha256": checksum)
              content_unit_href = content_list.results.first.pulp_href unless content_list.results.empty?
            end
            return {"content_unit_href" => content_unit_href} if content_unit_href
          end
```
upload_content.rb
```
sequence do
              content_backend_service = SmartProxy.pulp_primary.content_service(unit_type_id)
              if repository.generic?
                # pulp python doesn't support filter by sha256 (i.e. list method), so need to get the result differently
                content_list = content_backend_service.content_api(repository.repository_type, unit_type_id).list.results.find { |result| result.sha256 == Digest::SHA256.hexdigest(File.read(file[:path])) }
                content_href = content_list&.pulp_href
              else
                content_list = content_backend_service.content_api.list("sha256": Digest::SHA256.hexdigest(File.read(file[:path])))
                content_href = content_list&.results&.first&.pulp_href
              end

              unless content_href
                upload_action_output = plan_action(Pulp3::Repository::UploadFile, repository, smart_proxy, file[:path]).output
                artifact_action_output = plan_action(Pulp3::Repository::SaveArtifact, file, repository, smart_proxy, upload_action_output[:pulp_tasks], unit_type_id).output
                content_href = artifact_action_output[:pulp_tasks]
              end
              action_output = plan_action(Pulp3::Repository::ImportUpload, content_href, repository, smart_proxy).output
              plan_action(Pulp3::Repository::SaveVersion, repository, tasks: action_output[:pulp_tasks]).output
              plan_self(:subaction_output => action_output)
            end
```

ianballou (commented):
A couple notes:

Should `/katello/api/content/` be `/katello/api/generic_content/` if it's limited to generic content?

I think the content unit label should be plural, that's what we usually stick with: 
```
/katello/api/repositories/:id/content/python_packages
```

Listing individual content units seems to be missing:
```
/katello/api/content/python_packages/:id/
```

The same goes for the repository endpoint for content. I think we should be able to list single content units per repository.

ianballou (commented):
For some reason I'm having a hard time now uploading files. Here's what I tried:

```ruby
request = RestClient::Request.new( :method => :post, :url => 'https://centos7-katello-devel-3.cannolo.example.com/katello/api/repositories/61/upload_content', :user => 'admin', :password => 'changeme', :payload => { :multipart => true, :content => [File.new("scipy-1.1.0-cp36-none-win_amd64.whl", 'rb')], :content_type => 'python_package' })
request.execute
```

I see the task in Pulp (`"pulpcore.app.tasks.repository.add_and_remove"`) but no content unit is uploaded and there isn't a new repository version created.

rverdile:
I changed the API to work for all generic content types. It's all listed on the API doc, but two examples are:
`/katello/api/python_packages` and `/katello/api/repositories/:id/python_packages`

I still need to fix the generic_content_units_controller test.

rverdile:
> A couple notes:
> 
> Should `/katello/api/content/` be `/katello/api/generic_content/` if it's limited to generic content?
> 
> I think the content unit label should be plural, that's what we usually stick with:
> 
> ```
> /katello/api/repositories/:id/content/python_packages
> ```
> 
> I know it's a bit different from just creating an endpoint like `rpms`, so if making it plural is unreasonably complicated I think it's fine to leave it singular.
> 
> Listing individual content units seems to be missing:
> 
> ```
> /katello/api/content/python_packages/:id/
> ```
> 
> The same goes for the repository endpoint for content. I think we should be able to list single content units per repository.

See my other comment, but Justin made a couple more suggestions and this is completely different now :)

rverdile:
> For some reason I'm having a hard time now uploading files. Here's what I tried:
> 
> ```ruby
> request = RestClient::Request.new( :method => :post, :url => 'https://centos7-katello-devel-3.cannolo.example.com/katello/api/repositories/61/upload_content', :user => 'admin', :password => 'changeme', :payload => { :multipart => true, :content => [File.new("scipy-1.1.0-cp36-none-win_amd64.whl", 'rb')], :content_type => 'python_package' })
> request.execute
> ```
> 
> I see the task in Pulp (`"pulpcore.app.tasks.repository.add_and_remove"`) but no content unit is uploaded and there isn't a new repository version created.

Did you replace those two blocks I mentioned in my comment? I removed the work-around code. I just tried uploading with the a similar request to the one that you give here and it worked for me.

rverdile:
New work around in `upload_content.rb` is 

```
if repository.generic?
                duplicate_sha_path_content_list = content_backend_service.content_api(repository.repository_type, unit_type_id).list.results.find { |result| result.sha256 == Digest::SHA256.hexdigest(File.read(file[:path])) }
                duplicate_content_href = duplicate_sha_path_content_list&.pulp_href
              else
                duplicate_sha_path_content_list = content_backend_service.content_api.list(
                  "sha256": Digest::SHA256.hexdigest(File.read(file[:path])),
                  "relative_path": file[:filename])
                duplicate_content_href = duplicate_sha_path_content_list&.results&.first&.pulp_href
              end
```

sjha4:
> New work around in `upload_content.rb` is
> 
> ```
> if repository.generic?
>                 duplicate_sha_path_content_list = content_backend_service.content_api(repository.repository_type, unit_type_id).list.results.find { |result| result.sha256 == Digest::SHA256.hexdigest(File.read(file[:path])) }
>                 duplicate_content_href = duplicate_sha_path_content_list&.pulp_href
>               else
>                 duplicate_sha_path_content_list = content_backend_service.content_api.list(
>                   "sha256": Digest::SHA256.hexdigest(File.read(file[:path])),
>                   "relative_path": file[:filename])
>                 duplicate_content_href = duplicate_sha_path_content_list&.results&.first&.pulp_href
>               end
> ```

@rverdile: Can you check if duplicate content upload with same name/ different name works for generic content with this. change?

jlsherrill:
Oh this is so great!

```
[vagrant@devel2 foreman]$ curl  -u admin:changeme http://`hostname`/katello/api/v2/python_packages/   |jq
{
  "total": 2,
  "subtotal": 2,
  "page": 1,
  "per_page": 20,
  "error": null,
  "search": null,
  "sort": {
    "by": "name",
    "order": "asc"
  },
  "results": [
    {
      "id": 2,
      "name": "shelf-reader",
      "pulp_id": "/pulp/api/v3/content/python/packages/7ef7942c-b0d5-482e-a3aa-00b467149ec8/",
      "version": "0.1",
      "content_type": "python_package"
    },
    {
      "id": 1,
      "name": "shelf-reader",
      "pulp_id": "/pulp/api/v3/content/python/packages/eb3e84f9-6ee9-4f83-b793-68c2dd7c6beb/",
      "version": "0.1",
      "content_type": "python_package"
    }
  ]
}
```

app/controllers/katello/api/v2/generic_content_units_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+    def default_sort
+      %w(name asc)
+    end
+
+    def resource_class
+      ::Katello::GenericContentUnit.where(content_type: params[:content_type].singularize)
+    end
+
+    def filter_by_content_type(collection)
+      collection.where(:content_type => params[:content_type])
```
jlsherrill, line 21: I don't think this method is needed anymore right? (with the resource_class method?)

jlsherrill:
I saw the same upload issue that @ianballou saw, in addition i tried uploading via `hammer repository upload` and got an error:
```
2021-08-18T19:52:38 [E|bac|01f92257] wrong number of arguments (given 0, expected 2) (ArgumentError)
 01f92257 | /home/vagrant/git/katello/app/services/katello/pulp3/generic_content_unit.rb:15:in `content_api'
 01f92257 | /home/vagrant/git/katello/app/lib/actions/pulp3/orchestration/repository/import_upload.rb:15:in `block in plan'
 01f92257 | /home/vagrant/foreman/.vendor/ruby/2.7.0/gems/dynflow-1.5.0/lib/dynflow/execution_plan.rb:384:in `switch_flow'
 01f92257 | /home/vagrant/foreman/.vendor/ruby/2.7.0/gems/dynflow-1.5.0/lib/dynflow/action.rb:423:in `sequence'
 01f92257 | /home/vagrant/git/katello/app/lib/actions/pulp3/orchestration/repository/import_upload.rb:12:in `plan'
 01f92257 | /home/vagrant/foreman/.vendor/ruby/2.7.0/gems/dynflow-1.5.0/lib/dynflow/action.rb:515:in `block (3 levels) in execute_plan'
 01f92257 | /home/vagrant/foreman/.vendor/ruby/2.7.0/gems/dynflow-1.5.0/lib/dynflow/middleware/stack.rb:27:in `pass'
 01f92257 | /home/vagrant/foreman/.vendor/ruby/2.7.0/gems/dynflow-1.5.0/lib/dynflow/middleware.rb:19:in `pass'
 01f92257 | /home/vagrant/git/katello/app/lib/actions/middleware/backend_services_check.rb:23:in `plan'
 01f92257 | /home/vagrant/foreman/.vendor/ruby/2.7.0/gems/dynflow-1.5.0/lib/dynflow/middleware/stack.rb:23:in `call'
```

Note that these are two different code paths, as the UI does a 'one shot' upload in a single api call.  Hammer uses a more complex api set that supports uploading larger files in chunks and then importing the upload.  (providing some context)

rverdile:
> > New work around in `upload_content.rb` is
> > ```
> > if repository.generic?
> >                 duplicate_sha_path_content_list = content_backend_service.content_api(repository.repository_type, unit_type_id).list.results.find { |result| result.sha256 == Digest::SHA256.hexdigest(File.read(file[:path])) }
> >                 duplicate_content_href = duplicate_sha_path_content_list&.pulp_href
> >               else
> >                 duplicate_sha_path_content_list = content_backend_service.content_api.list(
> >                   "sha256": Digest::SHA256.hexdigest(File.read(file[:path])),
> >                   "relative_path": file[:filename])
> >                 duplicate_content_href = duplicate_sha_path_content_list&.results&.first&.pulp_href
> >               end
> > ```
> 
> @rverdile: Can you check if duplicate content upload with same name/ different name works for generic content with this. change?

It looks like the duplicates check depends on a `relative_path` attribute which doesn't exist for pulp python, so duplicates aren't duplicated. I need to see if there's a way around that.

ianballou:
Upload worked once I put in the new workaround.

ianballou:
Listing single packages is working, yay

```
vagrant@centos7-katello-devel-3 ~/katello $ curl --user admin:changeme https://centos7-katello-devel-3.cannolo.example.com/katello/api/python_packages/70 | jq
100   168    0   168    0     0    521      0 --:--:-- --:--:-- --:--:--   523
{
  "id": 70,
  "name": "pulpcore",
  "pulp_id": "/pulp/api/v3/content/python/packages/b2da3306-030e-4808-a0f6-0cedf94a8789/",
  "version": "3.12.1",
  "content_type": "python_package"
}
```

rverdile:
Pulp python doesn't allow for duplicate uploads: https://github.com/pulp/pulp_python/blob/3436fab7c82449056714da0de5119f4c175c6707/pulp_python/app/serializers.py#L261

> > > New work around in `upload_content.rb` is
> > > ```
> > > if repository.generic?
> > >                 duplicate_sha_path_content_list = content_backend_service.content_api(repository.repository_type, unit_type_id).list.results.find { |result| result.sha256 == Digest::SHA256.hexdigest(File.read(file[:path])) }
> > >                 duplicate_content_href = duplicate_sha_path_content_list&.pulp_href
> > >               else
> > >                 duplicate_sha_path_content_list = content_backend_service.content_api.list(
> > >                   "sha256": Digest::SHA256.hexdigest(File.read(file[:path])),
> > >                   "relative_path": file[:filename])
> > >                 duplicate_content_href = duplicate_sha_path_content_list&.results&.first&.pulp_href
> > >               end
> > > ```
> > 
> > 
> > @rverdile: Can you check if duplicate content upload with same name/ different name works for generic content with this. change?
> 
> It looks like the duplicates check depends on a `relative_path` attribute which doesn't exist for pulp python, so duplicates aren't duplicated. I need to see if there's a way around that.

Pulp python doesn't allow for duplicate uploads: https://github.com/pulp/pulp_python/blob/3436fab7c82449056714da0de5119f4c175c6707/pulp_python/app/serializers.py#L261.

I'm thinking I should make a boolean for that in the registration file.

ianballou:
> Pulp python doesn't allow for duplicate uploads: https://github.com/pulp/pulp_python/blob/3436fab7c82449056714da0de5119f4c175c6707/pulp_python/app/serializers.py#L261
> 
> > > > New work around in `upload_content.rb` is
> > > > ```
> > > > if repository.generic?
> > > >                 duplicate_sha_path_content_list = content_backend_service.content_api(repository.repository_type, unit_type_id).list.results.find { |result| result.sha256 == Digest::SHA256.hexdigest(File.read(file[:path])) }
> > > >                 duplicate_content_href = duplicate_sha_path_content_list&.pulp_href
> > > >               else
> > > >                 duplicate_sha_path_content_list = content_backend_service.content_api.list(
> > > >                   "sha256": Digest::SHA256.hexdigest(File.read(file[:path])),
> > > >                   "relative_path": file[:filename])
> > > >                 duplicate_content_href = duplicate_sha_path_content_list&.results&.first&.pulp_href
> > > >               end
> > > > ```
> > > 
> > > 
> > > @rverdile: Can you check if duplicate content upload with same name/ different name works for generic content with this. change?
> > 
> > 
> > It looks like the duplicates check depends on a `relative_path` attribute which doesn't exist for pulp python, so duplicates aren't duplicated. I need to see if there's a way around that.
> 
> Pulp python doesn't allow for duplicate uploads: https://github.com/pulp/pulp_python/blob/3436fab7c82449056714da0de5119f4c175c6707/pulp_python/app/serializers.py#L261.
> 
> I'm thinking I should make a boolean for that in the registration file.

I think a boolean would be fine to check if duplicate uploads are supported.

ianballou:
Support for `/katello/api/repositories/:id/python_packages/` would be good as part of this PR.

sjha4:
>> I'm thinking I should make a boolean for that in the registration file.

There are other content types that don't support duplicate upload. For example a duplicate RPM (same NVRA) will fetch the duplicate content and associate that to the repo. I think python packages behave the same and uniqueness is defined by other fields which are not handled. This should be fine IMO as long as it is able to associate the existing duplicate to the repo.

rverdile:
Uploading should work with either UI or hammer if you patch the pulp python code.

To do that edit `/usr/local/lib/python3.6/site-packages/pulp_python/app/viewsets.py` with this change: https://github.com/pulp/pulp_python/commit/aeb205db4a9cf02a6c65674894832cb5bad5295e

Then edit `/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/pulp_python_client-3.4.0/lib/pulp_python_client/api/content_packages_api.rb` line ~230 under "query_parameters" with:

```
query_params[:'sha256'] = opts[:'sha256'] if !opts[:'sha256'].nil?
query_params[:'sha256__in'] = @api_client.build_collection_param(opts[:'sha256__in'], :csv) if !opts[:'sha256__in'].nil?
```

I'm adding this `/katello/api/repositories/:id/python_packages/` now.

app/lib/actions/pulp3/orchestration/repository/upload_content.rb
```diff
@@ -7,17 +8,25 @@ class UploadContent < Pulp3::Abstract
...
-              duplicate_sha_path_content_list = content_backend_service.content_api.list(
+              duplicates_allowed = ::Katello::RepositoryTypeManager.find_content_type(unit_type_id).try(:duplicates_allowed)
+              if repository.generic? && duplicates_allowed
+                filename_key = ::Katello::RepositoryTypeManager.find_content_type(unit_type_id).filename_key
+                duplicate_sha_path_content_list = content_backend_service.content_api(repository.repository_type, unit_type_id).list(
+                  "sha256": file[:sha256],
+                  filename_key => file[:filename])
+              elsif repository.generic?
+                duplicate_sha_path_content_list = content_backend_service.content_api(repository.repository_type, unit_type_id).list(
+                  "sha256": file[:sha256])
```
jlsherrill, line 19: I don't think file[:sha256] is actually present here, notice that it calculates it in the block below

app/lib/actions/pulp3/orchestration/repository/upload_content.rb
```diff
@@ -7,17 +8,25 @@ class UploadContent < Pulp3::Abstract
           def plan(repository, smart_proxy, file, unit_type_id)
             sequence do
               content_backend_service = SmartProxy.pulp_primary.content_service(unit_type_id)
-              duplicate_sha_path_content_list = content_backend_service.content_api.list(
+              duplicates_allowed = ::Katello::RepositoryTypeManager.find_content_type(unit_type_id).try(:duplicates_allowed)
+              if repository.generic? && duplicates_allowed
```
jlsherrill, line 12: maybe move this big if statement to a new function, find_duplciate_unit or something like that?

PR #9496, opened 2021-07-26
Fixes #33146 - Rename gpg model and update references

chris1984:


jturel:
Found a reference to GpgKey model:

```
app/lib/katello/validators/gpg_key_content_validator.rb:          record.errors[attribute] << _("must contain valid  Public GPG Key") if line.length > GpgKey::MAX_CONTENT_LINE_LENGTH
```

Seems that validator is lacking test coverage - I propose adding a happy-path test at least to cover this part that is not causing a test to fail

app/lib/katello/validators/gpg_key_content_type_validator.rb
```diff
@@ -5,7 +5,7 @@ def validate(record)
         # need to verify, that gpg_key is of GpgKey.content type "gpg_key" and
         # ssl_ca_cert, ssl_client_cert, ssl_client_key of GpgKey.content type "cert"
 
-        if !record.gpg_key.blank? && record.gpg_key.content_type != "gpg_key"
+        if !record.gpg_key_id.blank? && record.gpg_key.content_type != "gpg_key"
```
jeremylenz, line 8: should this be updated? Why is it still `gpg_key_id`?

jturel: That's definitely a bigger change, but it might be the best thing to do to get us past this gpg keys vs content credentials once and for all. Chris, if you do this, please add those updates in a separate commit so we reviewers can carefully see how the code has changed in that regard since it means renaming table columns, updating associations, and usage of those associations

chris1984: @jlsherrill said to keep them the same, since there are still things that reference a `gpg_key` The root repository has a column for `gpg_key` and so does `product` I pinged him so he can shed some more light on it but he told me to keep those as is for now.

jturel: Since that would presumably affect the API as well, and therefore all clients, I'm OK with leaving the associations as-is

jlsherrill: so a little history here

Originally we had Gpgkeys that you could create in the UI and associate to a repository and all was well, hence the association named 'gpg_key'
Atix contributed enhancements to this feature that enabled us to associate other things like ssl_ca_cert, ssl_client_key, ssl_client_cert to a repository.  They implemented a new api (Content Credentials), but at the time we didn't want to rename the model.  So now you have 4 associations between  the gpg_keys table and the repositories table:  gpg_key_id, ssl_ca_cert_id, etc....  

With what chris is doing, he's finally renaming the gpg_keys table/model to content_credentials.  It woudlnt' make any sense to also rename the association to content_credentials_id because root_repository has 4 different associations to content_credentials and that would just cause tons of confusion

jturel, line 8: Why did this have to change? /me probably missing something obvious :)

jturel:
LGTM - just left one comment!

app/lib/katello/validators/gpg_key_content_type_validator.rb
chris1984: Something else that needed to be changed that errored out after I changed the table name with the migration. @jlsherrill do you know why?

jturel: Does the unit test you wrote fail without it?

chris1984: My unit test passed with changing it back to `gpg_key` so maybe I misread something. I am letting the rest of the test suite run, if everything passes, I will push up the updated fix.

jlsherrill: I think this should work either way, you might have just changed it when you were debugging the table name issues +1 to changing it back

chris1984: Updated

jturel (approved):
APJ

PR #9497, opened 2021-07-26
Refs #32867 - Create Remote in plan phase

sjha4:
Example invalid requirements.yaml that passes normal yaml syntax check and fails in pulp:

```
---
collections:
- name: testing.ansible_testing_content
  version: ">=1.0.0,<=2.0.0"
  source: https://galaxy-dev.ansible.com
- testing.k8s_demo_collection
```

### DIFF
--- a/app/lib/actions/pulp3/repository/create_remote.rb
+++ b/app/lib/actions/pulp3/repository/create_remote.rb
@@ -3,12 +3,7 @@ module Actions
     module Repository
       class CreateRemote < Pulp3::Abstract
         def plan(repository, smart_proxy)
-          plan_self(:repository_id => repository.id, :smart_proxy_id => smart_proxy.id) if repository.root.url?
-        end
-
-        def run
-          repo = ::Katello::Repository.find(input[:repository_id])
-          output[:response] = repo.backend_service(smart_proxy).create_remote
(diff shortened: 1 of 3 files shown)


sjha4:
[test katello]

sjha4:
[test katello]

app/lib/actions/pulp3/repository/create_remote.rb
```diff
@@ -3,12 +3,7 @@ module Pulp3
     module Repository
       class CreateRemote < Pulp3::Abstract
         def plan(repository, smart_proxy)
-          plan_self(:repository_id => repository.id, :smart_proxy_id => smart_proxy.id) if repository.root.url?
-        end
-
-        def run
-          repo = ::Katello::Repository.find(input[:repository_id])
-          output[:response] = repo.backend_service(smart_proxy).create_remote
+          repository.backend_service(smart_proxy).create_remote
```
jlsherrill, line 6: i think we need a 'if repository.root.url?'  here.  I get an error when i try to create a repo without a url.

sjha4: Updated..Had forgotten we allowed that..

jlsherrill (approved):

PR #9500, opened 2021-07-26
Fixes #33147 - Add validation around Ansible auth url and token

sjha4:


jturel:
This is slightly problematic in the UI in one case:

Consider an ansible collection repo that has no auth url or auth token. Given that each field is saved individually in the UI, it's impossible to specify the token and/or url because the new validation error is raised!

This also does not address ansible collections which are already invalid per your new validations. Not sure of the solution. Could the sync call .valid? on the root repo and raise errors at sync time? That might help us to fail faster in general

sjha4:
> This is slightly problematic in the UI in one case:
> 
> Consider an ansible collection repo that has no auth url or auth token. Given that each field is saved individually in the UI, it's impossible to specify the token and/or url because the new validation error is raised!

Mimicking upstream authorization behavior for Ansible collection authorization behavior. This should resolve the UX issue of having to save twice with error on first save.

> This also does not address ansible collections which are already invalid per your new validations. Not sure of the solution. Could the sync call .valid? on the root repo and raise errors at sync time? That might help us to fail faster in general

That's a larger issue. This is not a new validation. https://pulp.plan.io/issues/7802. Why this was being allowed on repository update is a separate issue where we don't track the updateRemote task which is async. In the short term, there's a refreshRemote task that runs before every sync that will validate this in Pulp. @jlsherrill ^^

sjha4:
Rebased with master to resolve conflicts from the other PR.

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/repository-details-info.controller.js
```diff
@@ -113,7 +113,7 @@ angular.module('Bastion.repositories').controller('RepositoryDetailsInfoControll
                 repository.os_versions = $scope.osVersionsParam();
                 repository.$update(function (response) {
                     deferred.resolve(response);
-                    $scope.repository.ignore_srpms = $scope.repository.ignorable_content.includes("srpm");
+                    $scope.repository.ignore_srpms = $scope.repository.ignorable_content && $scope.repository.ignorable_content.includes("srpm");
```
jturel, line 116: Was there another bug here? :o

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/repository-details-info.controller.js
```diff
@@ -195,6 +195,13 @@ angular.module('Bastion.repositories').controller('RepositoryDetailsInfoControll
                 $scope.save($scope.repository);
             };
 
+            $scope.clearAnsibleCollectionAuth = function () {
+                $scope.repository['ansible_collection_auth_url'] = null;
+                $scope.repository['ansible_collection_auth_exists'] = false;
```
jturel, line 200: Since the accuracy of `ansible_collection_auth_exists` can be affected by the actions of the UI I think it may be worth removing that attribute from the API response and just calculating it on demand via a new function in this controller. Is there any other advantage in carrying that calculated value from the server side?

app/models/katello/root_repository.rb
```diff
@@ -275,6 +276,20 @@ def ensure_valid_upstream_authorization
       end
     end
 
+    def ensure_valid_auth_url_token
```
jturel, line 279: There's quite a bit of behavior in this method - do you mind adding a unit test?

jturel:
> > This also does not address ansible collections which are already invalid per your new validations. Not sure of the solution. Could the sync call .valid? on the root repo and raise errors at sync time? That might help us to fail faster in general
> 
> That's a larger issue. This is not a new validation. https://pulp.plan.io/issues/7802. Why this was being allowed on repository update is a separate issue where we don't track the updateRemote task which is async. In the short term, there's a refreshRemote task that runs before every sync that will validate this in Pulp. @jlsherrill ^^

I see that now based on the call this morning. Sounds like an issue for another time - one approach being to create the remote in the plan phase as a way to be more certain that updating the pulp remote would be a success (Hope I summarized that suggestion correctly - it wasn't my own).

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/repository-details-info.controller.js
```diff
@@ -27,7 +27,7 @@ angular.module('Bastion.repositories').controller('RepositoryDetailsInfoControll
 
             $scope.repository.$promise.then(function () {
                 $scope.uploadURL = 'katello/api/v2/repositories/' + $scope.repository.id + '/upload_content';
-                $scope.repository['ignore_srpms'] = $scope.repository['ignorable_content'].includes("srpm");
+                $scope.repository['ignore_srpms'] = $scope.repository['ignorable_content']?.includes("srpm");
```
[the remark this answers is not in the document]
sjha4: Ya 😄 . I was seeing a (harmless other than annoying console) error on updates when there were no content types to ignore and this array was empty. That's always the case with all non-yum types.

The ?. doesn't work on our angular pages so went with the older way in the last commit. Probably older ES support..

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/repository-details-info.controller.js
sjha4: Same as above..Console error when $scope.repository.ignorable_content is null..

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/repository-details-info.controller.js
sjha4: I mostly copied the pattern of Username/Password field combo to not have to do too much angular.

sjha4: Let me try and update that to be UI only as there doesn't seem to be much value in the API returning it..

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/repository-details-info.controller.js
```diff
@@ -27,7 +27,8 @@ angular.module('Bastion.repositories').controller('RepositoryDetailsInfoControll
 
             $scope.repository.$promise.then(function () {
                 $scope.uploadURL = 'katello/api/v2/repositories/' + $scope.repository.id + '/upload_content';
-                $scope.repository['ignore_srpms'] = $scope.repository['ignorable_content'].includes("srpm");
+                $scope.repository['ignore_srpms'] = $scope.repository['ignorable_content'] && $scope.repository['ignorable_content'].includes("srpm");
+                $scope.repository['ansible_collection_auth_exists'] = $scope.repository['ansible_collection_auth_url'] && $scope.repository['ansible_collection_auth_token'];
```
jturel, line 31: If you make a function ie `ansibleCollectionAuthExists()` then there's no need to track or nullify it, or is there a reason not to do that? :thinking:

sjha4:
@jturel: Thanks for the reviews! Updated based on comments..

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/repository-details-info.controller.js
sjha4: I tried that..The function doesn't wait for `promise.then` before the 1st render ignoring what the actual repository record has..

jturel: Ah! Ok, let's forget about that then :)

jturel (approved):
ACK! Nice work @sjha4

PR #9501, opened 2021-07-26
Fixes #32956 -  Table checkbox fix

Andrewgdewar:
Checkbox selection within the table did not work for usage with bulk action buttons above table.

Note: Bulk checkbox did work to select all and action on all.

### DIFF
--- a/webpack/components/Table/helpers.js
+++ b/webpack/components/Table/helpers.js
@@ -5,7 +5,7 @@ const onSelect = (rows, setRows) => (_event, isSelected, rowId) => {
     newRows = rows.map(row => ({ ...row, selected: isSelected }));
   } else {
     newRows = [...rows];
-    newRows[rowId].selected = isSelected;
+    newRows[rowId] = { ...newRows[rowId], selected: isSelected };
   }
 
   setRows(newRows);


jeremylenz (approved):
LGTM :+1: 

The bulk actions menu wasn't being properly enabled when table rows are selected.  This fixes that.

It wasn't clear to me why mutating the array element wasn't working, yet reassigning it worked fine.  But I did some investigation and found that what was happening is that inside this effect,

https://github.com/Katello/katello/blob/1afd1c59f566e24a3c1e6d01eba77ae57d35db75/webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js#L124

the `rows` dependency wasn't registering as having changed, and so the effect was not getting run when the user clicked on checkboxes.  Changing the code here to reassign the element using ` = ` seemed to fix it.

Interestingly, I found that if you change this line

https://github.com/Katello/katello/blob/1afd1c59f566e24a3c1e6d01eba77ae57d35db75/webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js#L121

from `useDeepCompareEffect` to `useEffect`, it makes it work fine the "old" way as well. It must be a difference in the comparison functions used by the two.  :thinking:

jeremylenz:
[test katello]

Andrewgdewar:
I've attempted using the new imported method of useDeepCompareEffect a few times, and seen strange/errors appear. 
I'm uncertain of what is happening under the hood, but am uncertain of it's stability.

PR #9503, opened 2021-07-26
Fixes #33103 - Content view publish for generic types

rverdile:
Adds generic repositories as allowable types for content views.

Test by updating content view to include a python repository (Not supported in UI. I used API request). Then interact with content view as usual.

### DIFF
--- a/app/models/katello/content_view_version.rb
+++ b/app/models/katello/content_view_version.rb
@@ -311,11 +311,13 @@ module Katello
 
     def update_content_counts!
       self.content_counts = {}
-      RepositoryTypeManager.indexable_content_types.map(&:model_class).each do |content_type|
-        if content_type::CONTENT_TYPE == DockerTag::CONTENT_TYPE
+      RepositoryTypeManager.indexable_content_types.each do |content_type|
+        if content_type&.model_class::CONTENT_TYPE == DockerTag::CONTENT_TYPE
           content_counts[DockerTag::CONTENT_TYPE] = docker_tags.count
+        elsif content_type&.model_class::CONTENT_TYPE == GenericContentUnit::CONTENT_TYPE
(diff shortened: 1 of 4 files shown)


jlsherrill:
This works! can we update https://github.com/Katello/katello/blob/master/app/models/katello/content_view_version.rb#L314-L321 
to update the content counts on the CVV?

app/models/katello/content_view_version.rb
```diff
@@ -311,11 +311,13 @@ def package_groups
 
     def update_content_counts!
       self.content_counts = {}
-      RepositoryTypeManager.indexable_content_types.map(&:model_class).each do |content_type|
-        if content_type::CONTENT_TYPE == DockerTag::CONTENT_TYPE
+      RepositoryTypeManager.indexable_content_types.each do |content_type|
+        if content_type&.model_class::CONTENT_TYPE == DockerTag::CONTENT_TYPE
           content_counts[DockerTag::CONTENT_TYPE] = docker_tags.count
+        elsif content_type&.model_class::CONTENT_TYPE == GenericContentUnit::CONTENT_TYPE
+          content_counts[content_type.content_type] = content_type&.model_class.in_repositories(self.repositories.archived).select { |unit| unit.content_type == content_type.content_type }.count
```
jlsherrill, line 318: .select{} would retrieve all the content types and then iterate over each one which works but would be slow.  A better way would be to use the query structure:
```
.where(:content_type => content_type.content_type).count
```

that executes it in the database via sql and should be much faster

jlsherrill:
these errors are very strange...  
[test katello]

jlsherrill (approved):
ACK thanks @rverdile !!

jlsherrill:
also FYI i tested with docker and yum content to confirm no breakage there!

PR #9506, opened 2021-07-28
Refs #33040 - Readable upgrade-all tests

parthaa:
Added a couple of suggestions from a prior pr that got missed.
Pr #9474


jeremylenz (approved):
APJ

PR #9511, opened 2021-07-28
Refs #32221 - Add back useEffect dependencies

jeremylenz:
https://github.com/theforeman/foreman/pull/8699 is now merged, so we can reenable the eslint rules in these two places.


jturel (approved):
Works well! APJ

jturel:
@jeremylenz I think this can be merged :)

PR #9516, opened 2021-07-29
Fixes #33182 - make --errata-ids required for host errata apply

lfu:


app/controllers/katello/api/v2/host_errata_controller.rb
```diff
@@ -53,7 +53,7 @@ def index
 
     api :PUT, "/hosts/:host_id/errata/apply", N_("Schedule errata for installation using katello-agent. %s") % katello_agent_deprecation_text, deprecated: true
     param :host_id, :number, :desc => N_("Host ID"), :required => true
-    param :errata_ids, Array, :desc => N_("List of Errata ids to install. Will be removed in %s") % katello_agent_removal_release, required: false, deprecated: true
+    param :errata_ids, Array, :desc => N_("List of Errata ids to install. Will be removed in %s") % katello_agent_removal_release, required: true, deprecated: true
```
jturel, line 56: This is going to make hammer enforce that this param is required even though `--ids` is the proper argument to provide, so I'm not sure this is what we want

lfu: But `--ids` is broken. The only way is to use `--errata-ids`.

jturel: It's not truly required though, because the API itself takes the included & excluded parameters. --ids is only a hammer thing.

It may actually be worth investigating if we can somehow fix --ids from the hammer side. ie make sure when ids is used only included['ids'] is populated, and not included['ids'] and excluded['ids']

Curious what @parthaa and @chris1984 think

lfu: Hammer core  does the options parse. The options should be unique. But we have `--ids` for both `included` and `excluded` and that is why it got populated for both.

If we rename `excluded ids`, then it would affect API too.

jturel: Maybe we can just mark the errata-ids as no longer deprecated and remove --ids from the hammer command if possible? that shouldn't break anyone and is a no-op for --ids since it's already broken. Not a great solution but trying to brainstorm

jturel: That makes me wonder if --ids is broken for every hammer command backed by an api endpoint that's using bulk_errata_params :-/

chris1984: We can hide the `name` parameter by adding the controller here: https://github.com/Katello/hammer-cli-katello/blob/6a9f329fb097154750fae4842bd7ae99a3e5147f/lib/hammer_cli_katello/id_resolver.rb#L5-L34 as far as id's there might be a way let me look.

lfu: Anything to be updated? @jturel @chris1984

jturel: My suggestions (please let me know if you agree / disagree) is this:

- remove the deprecation warning from 'errata-ids' (keeping the required: true)
- open a separate pr to hammer-cli-katello which refers to the same issue that will hide `--ids` from hammer host errata apply (and we should check if we need to do the same for other commands...)

That way errata-ids becomes the only way to specify errata with this command.

In the hammer-cli-katello PR you can use `build_options(without: ['ids'])` to hide that specific flag

jturel (approved):
ACK - Thanks Lucy!

PR #9523, opened 2021-08-02
Refs #32901 - updates for Katello 4.0 where pulp2 is no longer supported

jjeffers:
This PR addresses a bad CP into the origin 4.0.2 release. Without it, CVV incremental updates will not work.

### DIFF
--- a/app/lib/actions/katello/content_view_version/incremental_update.rb
+++ b/app/lib/actions/katello/content_view_version/incremental_update.rb
@@ -78,10 +78,7 @@ module Actions
                                                        dependency_solving: dep_solve).output
                     repos_to_clone.each do |source_repos|
                       if separated_repo_map[:pulp3_yum].keys.include?(source_repos)
-                        copy_repos(repository_mapping[source_repos],
-                          new_content_view_version,
-                          content,
-                          dep_solve)
+                        copy_repos(repository_mapping[source_repos])
                       end
                     end
                   end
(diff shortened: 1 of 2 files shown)


app/lib/actions/katello/content_view_version/incremental_update.rb
```diff
@@ -169,9 +168,6 @@ def copy_repos(new_repo, new_version, content, dep_solve)
             copy_output += copy_yum_content(new_repo, solve_dependencies,
                                             content[:package_ids],
                                             content[:errata_ids])
-
-            plan_action(Katello::Repository::MetadataGenerate, new_repo)
```
ianballou, line 173: Why are these removed here? I believe they're still needed for types other than `:pulp3_yum`.

jjeffers: In 4.0, there was no call to `copy_repos`, but I can see why they would still be needed there. 

The original 9441 was picked, but there was a dependent PR that was the "pulp2 action removal" work, which never made it into 4.0.

ianballou: Does it not work to make it look like it does in master right now? I realized also that `copy_deb_content` would bug out if there are Pulp 3 deb repos to be copied.

jjeffers: No because master has diverged sufficiently from 4.0 the change set can't be the same. That was the problem with the CP as it is.

test/actions/katello/content_view_version_test.rb
```diff
@@ -28,98 +28,6 @@ class IncrementalUpdateTest < TestBase
       katello_repositories(:fedora_17_x86_64)
     end
 
-    it 'plans' do
```
ianballou, line 31: We kept this test in master but changed it up a little. Will matching it up with master work? https://github.com/Katello/katello/blob/master/test/actions/katello/content_view_version_test.rb#L31

ianballou (commented):
Looks good, just one question about a test that was deleted.

test/actions/katello/content_view_version_test.rb
```diff
@@ -39,87 +39,16 @@ class IncrementalUpdateTest < TestBase
...
-        :author => "author1",
-        :version => "2.2.3"
-      )
-
-      assert_raise RuntimeError, "Adding multiple versions of the same Puppet Module is not supported by incremental update.  The following Puppet Modules have duplicate versions in the incremental update content list: [\"name1-author1\"]" do
-        action.send(:check_puppet_module_duplicates, [module1.id, module2.id])
-      end
-    end
-
+
```
ianballou, line 51: I think rubocop is unhappy with his line.

ianballou (approved):
Assuming the tests pass, it looks fine to me!

PR #9529, opened 2021-08-03
Refs #33147 - Allow Auth token without Auth URL

sjha4:
In the PR: https://github.com/Katello/katello/pull/9500 we wrongly added the validation that auth_token can not be provided without an auth URL.

The only validation needed is that auth_url can not be provided without a token. The vice-versa is fine and both fields blank is also valid.

Another issue I notice around collections is updating auth_url and token wasn't triggering a pulp remote update. That is fixed with this PR too.

### DIFF
--- a/app/models/katello/repository.rb
+++ b/app/models/katello/repository.rb
@@ -413,7 +413,7 @@ module Katello
       elsif ostree?
         "#{scheme}://#{pulp_uri.host.downcase}/pulp/content/web/#{relative_path}"
       elsif ansible_collection?
-        "#{scheme}://#{pulp_uri.host.downcase}/pulp_ansible/galaxy/#{relative_path}/api"
+        "#{scheme}://#{pulp_uri.host.downcase}/pulp_ansible/galaxy/#{relative_path}/api/"
       else
         "#{scheme}://#{pulp_uri.host.downcase}/pulp/content/#{relative_path}/"
       end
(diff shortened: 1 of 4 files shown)


app/models/katello/repository.rb
```diff
@@ -413,7 +413,7 @@ def full_path(smart_proxy = nil, force_http = false)
       elsif ostree?
         "#{scheme}://#{pulp_uri.host.downcase}/pulp/content/web/#{relative_path}"
       elsif ansible_collection?
-        "#{scheme}://#{pulp_uri.host.downcase}/pulp_ansible/galaxy/#{relative_path}/api"
+        "#{scheme}://#{pulp_uri.host.downcase}/pulp_ansible/galaxy/#{relative_path}/api/"
       else
```
sjha4, line 417: Allows users to copy paste this path as remote for other pulp instance, (ex: hub) with a trailing slash.

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/views/repository-info.html
```diff
@@ -96,7 +96,7 @@ <h4 translate>Sync Settings</h4>
           <input id="requirementFile" type="file" name="file_path" onclick="this.value = null" onchange="angular.element(this).scope().handleFiles(this)"/>
           <p bst-alert='info' ng-show="repository.content_type == 'ansible_collection'">
           <span translate>
-            You can upload a requirements.yml file above to auto-fill contents <b>OR</b> paste contents of <a ng-href="https://docs.ansible.com/ansible/devel/dev_guide/collections_tech_preview.html#install-multiple-collections-with-a-requirements-file" target="_blank"> Requirements.yml </a>below.
+            You can upload a requirements.yml file above to auto-fill contents <b>OR</b> paste contents of <a ng-href="https://docs.ansible.com/ansible/latest/galaxy/user_guide.html#install-multiple-collections-with-a-requirements-file" target="_blank"> Requirements.yml </a>below.
```
sjha4, line 99: Broken link in help text.

jlsherrill (approved):

PR #9530, opened 2021-08-03
Fixes #33206 - only refresh remote for needed types

jlsherrill:
on migration

### DIFF
--- a/app/lib/actions/pulp3/content_guard/refresh_all_distributions.rb
+++ b/app/lib/actions/pulp3/content_guard/refresh_all_distributions.rb
@@ -6,7 +6,8 @@ module Actions
           sequence do
             plan_action(Actions::Pulp3::ContentGuard::Refresh, smart_proxy)
 
-            roots = ::Katello::RootRepository.where.not(:content_type => ::Katello::Repository::DOCKER_TYPE).where(:unprotected => false)
+            protected_types = [::Katello::Repository::YUM_TYPE, ::Katello::Repository::FILE_TYPE, ::Katello::Repository::DEB_TYPE]
+            roots = ::Katello::RootRepository.where(:content_type => protected_types).where(:unprotected => false)
             repositories = ::Katello::Repository.where(:root => roots)
             if repositories.any?
               plan_action(::Actions::BulkAction, Actions::Pulp3::Repository::RefreshDistribution, repositories, smart_proxy.id, assume_content_guard_exists: true)


app/lib/actions/pulp3/content_guard/refresh_all_distributions.rb
```diff
@@ -6,7 +6,8 @@ def plan(smart_proxy)
           sequence do
             plan_action(Actions::Pulp3::ContentGuard::Refresh, smart_proxy)
 
-            roots = ::Katello::RootRepository.where.not(:content_type => ::Katello::Repository::DOCKER_TYPE).where(:unprotected => false)
+            protected_types = [::Katello::Repository::YUM_TYPE, ::Katello::Repository::FILE_TYPE, ::Katello::Repository::DEB_TYPE]
```
jturel, line 9: Is this something we'd want to know outside of this class, or is duplicated anywhere, or alter in the context of this action (given as an input)? ie maybe the protected types should be more centrally located if the answer to either of those is yes

jturel:
I logged in to the reproducer on the BZ and saw a bunch of Refresh Distribution tasks weren't successful for reasons like this:

```
Couldn't find Katello::Repository with 'id'=154
```

and one of them:

```
undefined method `pulp3_service_class' for nil:NilClass
```

Sharing in case you hadn't seen them

jlsherrill:
So this is solving this issue:  undefined method `pulp3_service_class' for nil:NilClass

I didn't see the other error, i'll dig into that!  Thanks!

app/lib/actions/pulp3/content_guard/refresh_all_distributions.rb
jlsherrill: I think it makes a lot of sense for each repository type to declare if it supports/makes sense to provide protected content via a content guard (or really if it doesn't make sense.)  That would probably belong more in master rather than 3.18.  I'm happy to just move this and define it somewhere like katello/pulp3/migration.rb or something here if that makes sense? (and file an issue to enhance master)

jlsherrill:
So i believe what has happened is that during the upgrade:

1) foreman_maintain checks for running tasks and finds none
2) during switchover katello mistakenly calls this refresh distribution on ostree & puppet repos:
D, [2021-07-30 04:19:52-0400 #2079] DEBUG -- : Running command foreman-rake katello:pulp3_content_switchover with stdin nil

This fails on these with:
```
2021-07-30T04:22:16 [E|bac|] undefined method `new' for nil:NilClass (NoMethodError)
 | /opt/theforeman/tfm/root/usr/share/gems/gems/katello-3.18.1.29/app/services/katello/pulp3/repository.rb:90:in `instance_for_type'
 | /opt/theforeman/tfm/root/usr/share/gems/gems/katello-3.18.1.29/app/models/katello/repository.rb:188:in `backend_service'
```
3) the rest of the installer runs, and upgrades to 6.10
4) after the foreman-installer run, the upgrade rake tasks get run which delete ostree and puppet repos (or are supposed to), https://github.com/Katello/katello/pull/9478  fixes an issue where by not all were deleted. 
4) foreman_maintain agains checks  for running tasks and tries to resume them all, meaning the paused resume tasks are re-run.  This causes the errors we see (notice its an hour later after most of upgrade has run):
```
2021-07-30T05:42:48 [E|bac|] undefined method `pulp3_service_class' for nil:NilClass (NoMethodError)
/var/log/foreman/production.log-20210731.gz:2021-07-30T05:42:47 [E|bac|] Couldn't find Katello::Repository with 'id'=237 (ActiveRecord::RecordNotFound)
/var/log/foreman/production.log-20210731.gz:2021-07-30T05:42:48 [E|bac|] Couldn't find Katello::Repository with 'id'=153 (ActiveRecord::RecordNotFound)
/var/log/foreman/production.log-20210731.gz:2021-07-30T05:42:48 [E|bac|] Couldn't find Katello::Repository with 'id'=154 (ActiveRecord::RecordNotFound)
/var/log/foreman/production.log-20210731.gz:2021-07-30T05:42:48 [E|bac|] Couldn't find Katello::Repository with 'id'=152 (ActiveRecord::RecordNotFound)
/var/log/foreman/production.log-20210731.gz:2021-07-30T05:42:48 [E|bac|] Couldn't find Katello::Repository with 'id'=298 (ActiveRecord::RecordNotFound)
```

Here's an example of an ostree repo running through similar code:
```
irb(main):006:0> Katello::Repository.new(:root => Katello::RootRepository.new(:content_type => 'puppet')).backend_service(SmartProxy.pulp_primary, true)
        5: from lib/tasks/console.rake:5:in `block in <top (required)>'
        4: from (irb):5
        3: from (irb):6:in `rescue in irb_binding'
        2: from katello (4.1.1.1) app/models/katello/repository.rb:192:in `backend_service'
        1: from katello (4.1.1.1) app/services/katello/pulp3/repository.rb:94:in `instance_for_type'
NoMethodError (undefined method `pulp3_service_class' for nil:NilClass)
```
And i believe some ostree repos DID get deleted, which is what led to the ' Couldn't find Katello::Repository' errors

jlsherrill:
[test katello]

jlsherrill:
So to simplify, i think this fix will actually solve all the issues seen with RefreshDistributions.  I'll setup a more complicated reproducer to verify that and run through it with ostree & puppet repos promoted to cvs

jlsherrill:
[test katello]

app/lib/actions/pulp3/content_guard/refresh_all_distributions.rb
jturel: Per standup, let's leave it as-is on this branch. Updating master with what you mentioned sounds like the best way to go, so +1 from me

jturel (approved):
I tested this on a 3.18 box with a puppet repo and without the patch I saw the error reported in the BZ. Application of the patch fixed it!

PR #9532, opened 2021-08-04
Ref #33199 - Try nock restore for pipeline failure

sjha4:
Going down the list of things we can do to resolve the intermittent katello-master-source-release pipeline failure.

The 2nd take is to use nock.restore() and nock.activate() between tests in this file. 

Also nudging @jeremylenz and @Andrewgdewar for awareness and ideas cause my next idea is to use jest --no-cache and it will slow down the test run times. It would probably be better to write new tests for the CVDetailsRepo component even though I don't see the issue with the way it is written today.

See for reference: https://github.com/Katello/katello/pull/9526


jlsherrill:
[test katello]

jlsherrill (approved):

PR #9533, opened 2021-08-04
Fixes #33211 - changes smart proxy label to avoid confusion

jjeffers:
It is often confusing to users that the smart proxy displays "Pulpcore" plugin version, when it is in fact the smart proxy content plugin version.

image


jjeffers:
[test katello]

chris1984 (approved):

chris1984:
ACK

PR #9534, opened 2021-08-04
Fixes #33212 - force repo creation for katello:correct_repositories

ianballou:
In the Pulp 3 service class for repositories, the `create` method checks if a repository reference exists and skips creation if so. This causes the Pulp 3 repo to not be recreated during `katello:correct_repositories`.  This PR addresses the issue by introducing a `force` option for the `create` method.

### DIFF
--- a/app/lib/actions/katello/repository/create.rb
+++ b/app/lib/actions/katello/repository/create.rb
@@ -2,7 +2,9 @@ module Actions
   module Katello
     module Repository
       class Create < Actions::EntryAction
-        def plan(repository, clone = false)
+        def plan(repository, args = {})
+          clone = args[:clone] || false
+          force_repo_create = args[:force_repo_create] || false
           repository.save!
           root = repository.root
 
@@ -11,7 +13,7 @@ module Actions
           org = repository.organization
           sequence do
             create_action = plan_action(Pulp3::Orchestration::Repository::Create,
-                                        repository, SmartProxy.pulp_primary)
+                                        repository, SmartProxy.pulp_primary, force_repo_create)
 
             return if create_action.error
 
@@ -22,8 +24,10 @@ module Actions
               if repository.product.redhat?
                 plan_action(Actions::Candlepin::Environment::AddContentToEnvironment, :view_env_cp_id => view_env.cp_id, :content_id => repository.content_id)
               else
-                content_create = plan_action(Katello::Product::ContentCreate, root)
-                plan_action(Actions::Candlepin::Environment::AddContentToEnvironment, :view_env_cp_id => view_env.cp_id, :content_id => content_create.input[:content_id])
+                unless root.content
+                  content_create = plan_action(Katello::Product::ContentCreate, root)
+                  plan_action(Actions::Candlepin::Environment::AddContentToEnvironment, :view_env_cp_id => view_env.cp_id, :content_id => content_create.input[:content_id])
+                end
               end
             end
 
--- a/app/services/katello/pulp3/repository.rb
+++ b/app/services/katello/pulp3/repository.rb
@@ -166,13 +166,16 @@ module Katello
         computed_options.except(:name, :client_key)
       end
 
-      def create
-        unless repository_reference
+      def create(force = false)
+        if force || !repository_reference
           response = api.repositories_api.create(create_options)
-          RepositoryReference.create!(
-           root_repository_id: repo.root_id,
-           content_view_id: repo.content_view.id,
-           repository_href: response.pulp_href)
+          RepositoryReference.where(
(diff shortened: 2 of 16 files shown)


app/lib/actions/katello/repository/create.rb
```diff
@@ -22,8 +24,10 @@ def plan(repository, clone = false)
               if repository.product.redhat?
                 plan_action(Actions::Candlepin::Environment::AddContentToEnvironment, :view_env_cp_id => view_env.cp_id, :content_id => repository.content_id)
               else
-                content_create = plan_action(Katello::Product::ContentCreate, root)
-                plan_action(Actions::Candlepin::Environment::AddContentToEnvironment, :view_env_cp_id => view_env.cp_id, :content_id => content_create.input[:content_id])
+                unless root.content
```
ianballou, line 27: Is it okay to skip these actions like I'm doing here if the root already has a `Content` ?  Otherwise, the `Content` fails to get created during the "Finalize" step.

ianballou:
Looks like this PR will also address this BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1788964

app/lib/actions/katello/repository/create.rb
jlsherrill: i realized that this rake task handles the situation where the repository is missing in pulp, but doesn't handle if the content is missing in candlepin (chiggins actually hit this recently iirc when testing this script ).  Which is handy because you're assuming here that if force gets passed in we MUST recreate the repository.  If we ever have to handle missing content, this might fall apart.  Maybe we should rename the option to 'force_repo_create' or something like that?  And then we could later pass in force_content_create to handle that particular situation (I think he filed a bug)

jlsherrill: sorry i wrote a paragraph and didn't answer your question.   I think what you have is fine because we're not handling a missing content object in candlepin.

jlsherrill:
do you think you could write a test for this?  We have a ton of tests for this task, but i wonder if none are really using vcr to do a full test.

jlsherrill:
oh and if you end up writing a vcr test, can you put it somewhere like: test/lib/tasks/pulpcore/  so that we can more easily add it to https://github.com/Katello/katello/blob/master/lib/katello/tasks/test.rake#L159

app/lib/actions/katello/repository/create.rb
ianballou: Cool, sounds good to me. I'll change up the naming.

ianballou:
@jlsherrill yep I'll add that test.

app/services/katello/pulp3/repository.rb
```diff
@@ -166,10 +166,10 @@ def compute_remote_options(computed_options = remote_options)
         computed_options.except(:name, :client_key)
       end
 
-      def create
-        unless repository_reference
+      def create(force = false)
+        if force || !repository_reference
           response = api.repositories_api.create(create_options)
-          RepositoryReference.create!(
+          RepositoryReference.first_or_create!.update(
```
ianballou, line 172: FIXME: This is failing when the repo reference doesn't already exist because there are no arguments to create the first one.

ianballou:
@jlsherrill the main code should be good to test now. I just need to write those tests.

app/services/katello/pulp3/repository.rb
```diff
@@ -166,13 +166,18 @@ def compute_remote_options(computed_options = remote_options)
...
           response = api.repositories_api.create(create_options)
-          RepositoryReference.create!(
-           root_repository_id: repo.root_id,
-           content_view_id: repo.content_view.id,
-           repository_href: response.pulp_href)
+          if force
+            RepositoryReference.where(
+              root_repository_id: repo.root_id,
+              content_view_id: repo.content_view.id).destroy_all
+          end
```
jlsherrill, line 176: would it make sense to just destroy regardless of the force option?   I think it'd simplify the code and prevent possible error conditions

jlsherrill:
i noticed that this only works with rpm repos due to: https://github.com/Katello/katello/blob/master/lib/katello/tasks/repository.rake#L122

I don't see an easy generic way to catch this, so maybe catching a 'StandardError' and if it replies with 'code',check to see if  ifs a 404.  If it is a 404, assume its gone, else raise the exception?

app/services/katello/pulp3/repository.rb
ianballou: Yeah I think so, that'll simplify things.

jlsherrill (approved):
ACK, tested with an ansible and yum repo, repo was re-created properly

PR #9535, opened 2021-08-04
Fixes #33213 - if erratum rules are empty, return an empty list of ids

jjeffers:


chris1984 (approved):
ACK, looks good. Thanks for adding a test @jjeffers

PR #9542, opened 2021-08-06
Fixes #33084 - Add content view details to Host overview

parthaa:


parthaa:
cv-card

webpack/components/extensions/HostDetails/Cards/ContentViewDetailsCard.js
```diff
@@ -0,0 +1,70 @@
...
+import ContentViewIcon from '../../../../scenes/ContentViews/components/ContentViewIcon';
+
+
+const ContentViewDetailsCard = ({ hostDetails }) => {
+  let contentViewDetails;
+  if (hostDetails.content_facet_attributes) {
+    const cv = hostDetails.content_facet_attributes.content_view;
+    const env = hostDetails.content_facet_attributes.lifecycle_environment;
+    const versionId = hostDetails.content_facet_attributes.content_view_version_id;
+    const version = hostDetails.content_facet_attributes.content_view_version;
```
jeremylenz, line 28: Destructuring in combo with the `propsToCamelCase` helper can help quite a bit here:
```suggestion
  import { propsToCamelCase } from 'foremanReact/common/helpers';
  ...
  const hostDetailsData = propsToCamelCase(hostDetails.content_facet_attributes);
  const { contentView, lifecycleEnvironment, contentViewVersionId, contentViewVersion } = hostDetailsData;
  
```

webpack/components/extensions/HostDetails/Cards/ContentViewDetailsCard.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  return (
+    <GridItem xl2={3} md={6} lg={5}>
+      <Card isHoverable>
+        <CardHeader>
+          <CardTitle>{__('Content View Details')}</CardTitle>
+        </CardHeader>
+        <CardBody>
+          {contentViewDetails}
```
jeremylenz, line 63: A (kinda strong) personal preference here: I don't like when JSX is outside the `return` or render method.  Instead of defining a const `contentViewDetails` above, I'd prefer using conditional rendering here, something like

```jsx
{contentViewDetails.content_facet_attributes &&
<DescriptionList isAutoColumnWidths>
// all the other JSX
}
```

This does raise the question though, if there is no data, should we even be showing the card?

jeremylenz (changes requested):
Nice work adding our first host details card! :smile:  Tested and seems to work great.

Looks like you still need to add a `propTypes` and `defaultProps` object. Let me know if you need help with making eslint happy..

Would also like to see a simple React Testing Library test.

webpack/components/extensions/HostDetails/Cards/ContentViewDetailsCard.js
```diff
@@ -0,0 +1,84 @@
...
+  Label,
+} from '@patternfly/react-core';
+
+import { urlBuilder } from 'foremanReact/common/urlHelpers';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { propsToCamelCase } from 'foremanReact/common/helpers';
+import PropTypes from 'prop-types';
+import ContentViewIcon from '../../../../scenes/ContentViews/components/ContentViewIcon';
+
+const renderContentViewDetails = (hostDetails) => {
```
jeremylenz, line 23: This is a function that returns JSX, therefore I think this is really another React component but named like a regular function. Let's fix that:
```suggestion
const HostContentViewDetails = (hostDetails) => {
```

webpack/components/extensions/HostDetails/Cards/ContentViewDetailsCard.js
```diff
@@ -0,0 +1,84 @@
...
+          </DescriptionList>
+        </CardBody>
+      </Card>
+    </GridItem>
+  );
+};
+
+const ContentViewDetailsCard = ({ hostDetails }) => {
+  if (hostDetails.content_facet_attributes) {
+    return renderContentViewDetails(hostDetails);
```
jeremylenz, line 69:
```suggestion
    return <HostContentViewDetails hostDetails={hostDetails} />;
```

webpack/components/extensions/HostDetails/Cards/ContentViewDetailsCard.js
```diff
@@ -0,0 +1,84 @@
...
+import PropTypes from 'prop-types';
+import ContentViewIcon from '../../../../scenes/ContentViews/components/ContentViewIcon';
+
+const renderContentViewDetails = (hostDetails) => {
+  const hostDetailsData = propsToCamelCase(hostDetails.content_facet_attributes);
+  const {
+    contentView,
+    lifecycleEnvironment,
+    contentViewVersionId,
+    contentViewVersion,
```
jeremylenz, line 29: Now that we're refactoring this to be a component, I think it would be good to add a `propTypes` object for these as well.  Could be a shape, like
```js
hostDetails: PropTypes.shape(
{
  content_view: PropTypes.shape({}),
  lifecycle_environment: PropTypes.shape({}),
  content_view_version_id: PropTypes.string,
...
}
)
```

Another option is to move everything up one level and have those items be props (not under the `hostDetails` object). Then you'd have to use `propsToCamelCase` with props spread syntax in the parent component, like
```js
<HostContentViewDetails {...propsToCamelCase(hostDetails.content_facet_attributes)} />
```
Completely up to you which makes more sense to you.

jeremylenz (commented):
Added some more comments

webpack/components/extensions/HostDetails/Cards/ContentViewDetailsCard.js
```diff
@@ -0,0 +1,94 @@
...
+      </CardBody>
+    </Card>
+  </GridItem>
+);
+
+const ContentViewDetailsCard = ({ hostDetails }) => {
+  if (hostDetails.content_facet_attributes) {
+    return <HostContentViewDetails {...propsToCamelCase(hostDetails.content_facet_attributes)} />;
+  }
+  return (<></>);
```
jeremylenz, line 66: Thoughts on having a Skeleton here?

webpack/components/extensions/HostDetails/Cards/ContentViewDetailsCard.js
```diff
@@ -0,0 +1,102 @@
...
+  CardHeader,
+  CardTitle,
+  CardBody,
+  Flex,
+  FlexItem,
+  GridItem,
+  Label,
+} from '@patternfly/react-core';
+
+import Skeleton from 'react-loading-skeleton';
```
jeremylenz, line 17: I meant the Patternfly skeleton :wink: 
```suggestion
import { Skeleton } from '@patternfly/react-core';
```

parthaa:
@jeremylenz added tests. Can ya check it again

webpack/components/extensions/HostDetails/Cards/ContentViewDetailsCard.js
```diff
@@ -0,0 +1,102 @@
...
+      </Card>
+    </GridItem>
+  );
+};
+
+const ContentViewDetailsCard = ({ hostDetails }) => {
+  if (hostDetails.content_facet_attributes) {
+    return <HostContentViewDetails {...propsToCamelCase(hostDetails.content_facet_attributes)} />;
+  }
+  return <Skeleton />;
```
jeremylenz, line 74: Even with the Patternfly skeleton this doesn't render anything useful.  I think we should move the `GridItem`, `Card`, `CardHeader`, and `CardBody` here, to the outer `ContentViewDetailsCard` component. And then just conditionally render either the `Skeleton` or the `DescriptionList`.

(And after you refactor all that, enjoy not having to change your React testing library test ;)

jeremylenz (commented):
Tests are looking good, thanks @parthaa!

Getting close; let's just fix this one thing

webpack/components/extensions/HostDetails/Cards/ContentViewDetailsCard.js
parthaa: Hmm don't agree on moving the grid item and card here. Currently the card does not show up if you don't pass the `content-facet-attributes`.  If we set grid item we'll be seeing empty card which we don't want.

jeremylenz: We do want an empty card! Otherwise it renders a skeleton on the page and not in the card.

jeremylenz: If you don't want anything to render at all, you wouldn't need a Skeleton

parthaa: I basically need the equivalent of "dont render anything". If `Skeleton` doesn't do that we need something else.

jeremylenz:
```js
return null;
```
would do that. And I'm fine with that in this case

jeremylenz: Skeleton lets the user know that data is still loading, so that makes sense to me here as well. But I'm fine either way

parthaa: Oh I should ve just returned null looks like.

parthaa: updated!

PR #9547, opened 2021-08-10
Fixes #33235 - Allow duplicate uploads to katello

sjha4:
Steps to Reproduce:
1. Create a product and a file-type repo in it.
2. In the repo page click Choose files, select 20 or so files (Use these 10 : https://fixtures.pulpproject.org/file-large/ and make a copy of the 10 to have several duplicates)
3. Click upload

Actual results:
The task succeeded but only 10 or less files were really synced

Expected results:
All files being synced or some notification it did not happen, not pass silently.

### DIFF
--- a/app/lib/actions/pulp3/repository/commit_upload.rb
+++ b/app/lib/actions/pulp3/repository/commit_upload.rb
@@ -10,8 +10,31 @@ module Actions
           repo = ::Katello::Repository.find(input[:repository_id])
           repo_backend_service = repo.backend_service(smart_proxy)
           uploads_api = repo_backend_service.core_api.uploads_api
-          upload_commit = repo_backend_service.core_api.upload_commit_class.new(sha256: input[:sha256])
-          output[:pulp_tasks] = [uploads_api.commit(input[:upload_href], upload_commit)]
+          duplicate_sha_artifact_list = ::Katello::Pulp3::Api::Core.new(smart_proxy).artifacts_api.list("sha256": input[:sha256])
+          duplicate_sha_artifact_href = duplicate_sha_artifact_list&.results&.first&.pulp_href
+          if duplicate_sha_artifact_href
+            uploads_api.delete(upload_href)
+            output[:artifact_href] = duplicate_sha_artifact_href
+            output[:pulp_tasks] = nil
+          else
+            output[:artifact_href] = nil
+            upload_commit = repo_backend_service.core_api.upload_commit_class.new(sha256: input[:sha256])
+            output[:pulp_tasks] = [uploads_api.commit(input[:upload_href], upload_commit)]
+          end
+        end
+
+        def check_for_errors
+          combined_tasks.each do |task|
+            if unique_error task.error
+              warn _("Duplicate artifact detected")
+            else
+              super
+            end
+          end
+        end
+
+        def unique_error(message)
+          message&.include?("code='unique'")
         end
       end
     end
--- a/app/lib/actions/pulp3/repository/upload_file.rb
+++ b/app/lib/actions/pulp3/repository/upload_file.rb
@@ -6,6 +6,7 @@ module Actions
           plan_self(:repository_id => repository.id, :smart_proxy_id => smart_proxy.id, :file => file)
         end
 
+        # rubocop:disable Metrics/MethodLength
         def invoke_external_task
           repo = ::Katello::Repository.find(input[:repository_id])
           repo_backend_service = repo.backend_service(smart_proxy)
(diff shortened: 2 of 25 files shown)


jlsherrill:
this worked well for the UI, but when i use hammer to do something similar:

```
hammer repository upload-content --id=3 --path uploads/
```

i have it upload 3 files (2 with the same checksum) and i only get 2 files total.

test/actions/pulp3/orchestration/file_upload_test.rb
```diff
@@ -50,14 +51,16 @@ def test_duplicate_upload
...
-        action_result = ForemanTasks.sync_task(::Actions::Pulp3::Orchestration::Repository::UploadContent, @repo, @primary, @file, "file")
+        action_result = ForemanTasks.sync_task(::Actions::Pulp3::Orchestration::Repository::UploadContent, @repo, @primary, @file1, "file")
       end
       assert_equal "success", action_result.result
+      @repo.reload
       repository_reference = Katello::Pulp3::RepositoryReference.find_by(
           :root_repository_id => @repo.root.id,
           :content_view_id => @repo.content_view.id)
-      assert_equal repository_reference.repository_href + "versions/1/", @repo.version_href
+      assert_equal repository_reference.repository_href + "versions/2/", @repo.version_href
```
jlsherrill, line 63: this does validate that the repository was updated 2x, but it might be better to fetch the repository version and verify the # of files in it

jlsherrill:
I tried to think of ways to streamline or simplify the code, but really couldn't come up with any 👍🏻  to the code overall

sjha4:
> this worked well for the UI, but when i use hammer to do something similar:
> 
> ```
> hammer repository upload-content --id=3 --path uploads/
> ```
> 
> i have it upload 3 files (2 with the same checksum) and i only get 2 files total.

Handled the hammer duplicate uploads flow in the last commit. Should allow duplicate file uploads now from UI and Hammer.

test/actions/pulp3/orchestration/file_upload_test.rb
sjha4: Updated to check content_count.. 👍🏽

app/lib/actions/pulp3/repository/commit_upload.rb
```diff
@@ -10,8 +10,31 @@ def invoke_external_task
           repo = ::Katello::Repository.find(input[:repository_id])
           repo_backend_service = repo.backend_service(smart_proxy)
           uploads_api = repo_backend_service.core_api.uploads_api
-          upload_commit = repo_backend_service.core_api.upload_commit_class.new(sha256: input[:sha256])
-          output[:pulp_tasks] = [uploads_api.commit(input[:upload_href], upload_commit)]
+          duplicate_sha_artifact_list = ::Katello::Pulp3::Api::Core.new(smart_proxy).artifacts_api.list("sha256": input[:sha256])
+          duplicate_sha_artifact_href = duplicate_sha_artifact_list&.results&.first&.pulp_href
```
jlsherrill, line 14: Is there ever a chance that this 'first' duplicate artifact is actually associated with a content unit in pulp?  Does it matter if its the first or the second?

jlsherrill: or are these only ever going to be artifacts that were uploaded as part of *this* upload?

app/lib/actions/pulp3/repository/upload_file.rb
```diff
@@ -34,10 +35,21 @@ def invoke_external_task
 
             if response
               upload_href = response.pulp_href
-              output[:pulp_tasks] = [uploads_api.commit(upload_href, sha256: sha256)]
+              #Check for any duplicate artifacts created in parallel subtasks
+              duplicate_sha_artifact_list = ::Katello::Pulp3::Api::Core.new(smart_proxy).artifacts_api.list("sha256": sha256)
+              duplicate_sha_artifact_href = duplicate_sha_artifact_list&.results&.first&.pulp_href
```
jlsherrill, line 40: upload file gets run in parallel it looks like, i'm kinda worried this could open us up to a race condition, but maybe its not worth worrying about right now.  Or maybe we should just switch upload files to upload them in sequence instead of in parallel?

jlsherrill: and that will decrease performance for web-ui based uploads, but those have to be small anyways, so its maybe not a big deal?

sjha4: Ya..this additional check at the last minute before commit is to avoid a race consition I was running into..There is a possibility of running into race conditions with duplicate files in the same batch even with this. But I think this is such a specific use-case that we should be OK? Thoughts?

app/lib/actions/pulp3/repository/commit_upload.rb
sjha4: So there can only be one artifact with a sha..But we can have multiple content units with the same artifact at different paths..

sjha4: The duplicate_sha_artifact_list can only have 0 or 1 artifacts in the results since we are looking up with sha which is unique.

jlsherrill (approved):

app/lib/actions/pulp3/repository/upload_file.rb
jlsherrill: after some discussion we decided to serialize webui uploads

sjha4:
[test katello]

PR #9549, opened 2021-08-10
Fixes #33181 - pagination info at the bottom & top

parthaa:
This commit updates the tablewrapper react component and renders the page controls on both top and bottom of the main table.

### DIFF
--- a/webpack/components/Table/TableWrapper.js
+++ b/webpack/components/Table/TableWrapper.js
@@ -1,6 +1,6 @@
 import React, { useState, useCallback } from 'react';
 import useDeepCompareEffect from 'use-deep-compare-effect';
-import { Pagination, Flex, FlexItem } from '@patternfly/react-core';
+import { Pagination, PaginationVariant, Flex, FlexItem } from '@patternfly/react-core';
 
 import PropTypes from 'prop-types';
 import { useDispatch } from 'react-redux';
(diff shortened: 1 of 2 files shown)


parthaa:
cv-paging

webpack/components/Table/TableWrapper.js
```diff
@@ -86,19 +102,10 @@ const TableWrapper = ({
...
-            itemCount={total}
-            page={page}
-            perPage={perPage}
-            onSetPage={(_evt, updated) => onPaginationUpdate({ page: updated })}
-            onPerPageSelect={(_evt, updated) => onPaginationUpdate({ per_page: updated })}
-            perPageOptions={usePaginationOptions().map(p => ({ title: p.toString(), value: p }))}
-            variant="top"
-          />
-        </FlexItem>
+        {ShowPageControls(PaginationVariant.top)}
```
jeremylenz, line 105: A function that returns JSX is a React component. Should be invoked not with `{...}` but instead
```js
<ShowPageControls variant={PaginationVariant.top} />
```

However, I wouldn't have this as a separate component anyway.

jeremylenz: Or you may choose to keep it as a separate component.. I don't feel as strongly about that. As long as it's notated correctly ;)

Andrewgdewar (changes requested):

jeremylenz:
Think this needs a rebase now.. sorry! :grimacing:

webpack/components/Table/TableWrapper.js
```diff
@@ -75,6 +75,22 @@ const TableWrapper = ({
     updatePagination(updatedPagination);
   };
 
+  const ShowPageControls = variant => (
```
jeremylenz, line 78: I actually like Andrew's name better than mine:
```suggestion
  const PageControls = variant => (
```

webpack/components/Table/TableWrapper.js
```diff
@@ -113,6 +119,7 @@ const TableWrapper = ({
       >
         {children}
       </MainTable>
+      <Flex><ShowPageControls variant={PaginationVariant.bottom} /></Flex>
```
jeremylenz, line 122: Prefer the child(ren) of `<Flex>` to be on a separate line.. just looks nicer to me

jeremylenz (commented):
Looking good! Didn't test it but just leaving a couple minor code comments

jeremylenz:
oh, looks like the tests also need to be updated!

jeremylenz:
@parthaa I'm still seeing the full pagination component at the top. Should be the compact one at the top (without the text field)

jeremylenz (approved):
Nice work @parthaa 

APJ :+1:

parthaa:
[test katello]

PR #9550, opened 2021-08-10
Fixes #33238 - Extend TableWrapper and MainTable to allow using PF TableComposable

jeremylenz:
This change does the following:

* Add a `composable` prop to `<TableWrapper>`
* if `composable` prop is passed, `MainTable` returns `<TableComposable>` instead of `<Table>` and doesn't accept `rows` or `cells` props-- so you can do either 
```js
<TableWrapper rows={...} cells={...} />
```
 or
```js
<TableWrapper composable > ... </TableWrapper>
```

Existing pages that use TableWrapper will require __no change__ whatsoever. Behavior will be exactly the same.

As a proof of concept I've refactored `ContentViewHistories` to use the composable TableWrapper. You can see how much it's simplified:

* No `buildRows` function
* No `useDeepCompareEffect`
* No use of state for `rows` or `metadata`
* Everything is visible in the markup

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,13 +23,14 @@
     "url": "https://projects.theforeman.org/projects/katello/issues"
   },
   "devDependencies": {
+    "@babel/eslint-parser": "^7.15.0",
+    "@babel/eslint-plugin": "^7.14.5",
     "@sheerun/mutationobserver-shim": "^0.3.3",
     "@testing-library/jest-dom": "^5.3.0",
     "@testing-library/react": "^10.0.2",
     "@theforeman/builder": ">= 6.0.0",
     "@theforeman/find-foreman": "^4.8.0",
     "axios-mock-adapter": "^1.10.0",
-    "babel-eslint": "^10.0.3",
     "eslint": "^6.2.2",
     "eslint-config-airbnb": "^16.0.0",
     "eslint-plugin-babel": "^5.1.0",
--- a/webpack/components/Table/MainTable.js
+++ b/webpack/components/Table/MainTable.js
@@ -3,7 +3,9 @@ import {
   Table,
   TableHeader,
   TableBody,
+  TableComposable,
 } from '@patternfly/react-table';
+import { translate as __ } from 'foremanReact/common/I18n';
 import { STATUS } from 'foremanReact/constants';
 import PropTypes from 'prop-types';
 
@@ -13,24 +15,43 @@ import Loading from '../../components/Loading';
 const MainTable = ({
   status, cells, rows, error, emptyContentTitle, emptyContentBody,
   emptySearchTitle, emptySearchBody, searchIsActive, activeFilters,
-  ...extraTableProps
+  composable, rowsCount, children, ...extraTableProps
 }) => {
+  if (!composable && (!cells || !rows)) {
+    console.error(__('The <MainTable> component requires either a composable prop, or cells & rows props.')); // eslint-disable-line no-console
+  }
+
+  const tableHasNoRows = () => {
+    if (composable) return rowsCount === 0;
+    return rows.length === 0;
+  };
   const isFiltering = activeFilters || searchIsActive;
   if (status === STATUS.PENDING) return (<Loading />);
   // Can we display the error message?
   if (status === STATUS.ERROR) return (<EmptyStateMessage error={error} />);
-  if (status === STATUS.RESOLVED && isFiltering && rows.length === 0) {
+  if (status === STATUS.RESOLVED && isFiltering && tableHasNoRows()) {
     return (<EmptyStateMessage
       title={emptySearchTitle}
       body={emptySearchBody}
       search
     />);
   }
-  if (status === STATUS.RESOLVED && rows.length === 0) {
+  if (status === STATUS.RESOLVED && tableHasNoRows()) {
     return (<EmptyStateMessage title={emptyContentTitle} body={emptyContentBody} />);
   }
 
   const tableProps = { cells, rows, ...extraTableProps };
+  if (composable) {
+    return (
+      <TableComposable
+        aria-label="Content View Table"
+        className="katello-pf4-table"
+        {...extraTableProps}
+      >
+        {children}
+      </TableComposable>
+    );
+  }
   return (
     <Table
       aria-label="Content View Table"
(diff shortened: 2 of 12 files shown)


webpack/global_index.js
```diff
@@ -7,8 +7,11 @@ import RegistrationCommands from './components/extensions/RegistrationCommands';
 import ContentTab from './components/extensions/HostDetails/Tabs/ContentTab';
```
jeremylenz, line 7: Renamed this from `global_fills.js` to `global_index.js` as this is now the best practice according to the Foreman plugin template.

webpack/components/Table/MainTable.js
```diff
@@ -13,24 +15,43 @@ import Loading from '../../components/Loading';
 const MainTable = ({
   status, cells, rows, error, emptyContentTitle, emptyContentBody,
   emptySearchTitle, emptySearchBody, searchIsActive, activeFilters,
-  ...extraTableProps
+  composable, rowsCount, children, ...extraTableProps
 }) => {
+  if (!composable && (!cells || !rows)) {
+    console.error(__('The <MainTable> component requires either a composable prop, or cells & rows props.')); // eslint-disable-line no-console
```
jeremylenz, line 21: I had to remove the `.isRequired` from `cells` and `rows` prop types, because they aren't required for composable tables.  So we're left needing to conditionally require props - either cells/rows, or `composable`.  (PropTypes does have a custom function, but it didn't really work for this purpose)

webpack/components/Table/MainTable.js
```diff
@@ -13,24 +15,43 @@ import Loading from '../../components/Loading';
...
+  }
+
+  const tableHasNoRows = () => {
```
jeremylenz, line 24: I added a `rowsCount` prop to make up for the fact that we might not have a `rows` to check the length of. But I still check `rows.length` as before for non-composable tables, because that guarantees that the behavior won't have changed and that feels safer.

webpack/components/Table/TableWrapper.js
```diff
@@ -73,6 +75,8 @@ const TableWrapper = ({
     updatePagination(updatedPagination);
   };
 
+  const rowsCount = metadata?.subtotal ?? 0;
```
jeremylenz, line 78: Enjoying my first use of https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator :smile:

webpack/components/Table/TableWrapper.js
```diff
@@ -83,9 +87,11 @@ const TableWrapper = ({
             getAutoCompleteParams={getAutoCompleteParams}
           />
         </FlexItem>
-        <FlexItem>
-          {children}
-        </FlexItem>
+        {!composable &&
```
jeremylenz, line 90: needed to avoid rendering the children twice

webpack/scenes/ContentViews/__tests__/contentViewPage.test.js
```diff
@@ -330,7 +330,7 @@ test('No results message is shown for empty search', async (done) => {
 
   fireEvent.change(getByLabelText(/text input for search/i), { target: { value: query } });
 
-  await patientlyWaitFor(() => expect(getByText(/No matching Content Views found/i)).toBeInTheDocument());
+  await patientlyWaitFor(() => expect(getByText(/No matching content views found/i)).toBeInTheDocument());
```
jeremylenz, line 333: for some reason this test was failing because it was still looking for the capitalized "Content Views"

jeremylenz:
For some reason registering Katello's reducer is causing Katello CSS to load and override Foreman pages. The `overflow: auto` on `<td>` elements causes major problems with dropdown menu visibility. Trying to discover the cause..

jeremylenz:
@laviro this is the PR I was bugging you about today :)

amirfefer:
@jeremylenz,
We can use the [scope](https://webpack.js.org/loaders/css-loader/#scope) feature of webpack's css-loader, with scoping a plugin's css is loaded without overriding core's css classes.

jeremylenz:
@sjha4 Updated to add a class to the Pf3 table.  See what you think..

jeremylenz:
@tbrisker I can't tag @theforeman/packaging here but I think this will need an ack, since it [modifies package.json](https://github.com/Katello/katello/pull/9550#discussion_r686913172).

package.json
```diff
@@ -23,13 +23,14 @@
     "url": "https://projects.theforeman.org/projects/katello/issues"
   },
   "devDependencies": {
+    "@babel/eslint-parser": "^7.15.0",
+    "@babel/eslint-plugin": "^7.14.5",
```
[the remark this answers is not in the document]
MariaAga: Should we also change it in foreman js and use `eslint-plugin-foreman` in Katello? (not in this pr)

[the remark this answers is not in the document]
jeremylenz: @MariaAga yes this sounds good to me, I think it should be part of @Andrewgdewar's work on https://projects.theforeman.org/issues/33172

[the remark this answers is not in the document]
jeremylenz: (but also I don't want that to delay merge of this change, since a lot of our upcoming work depends on it)

webpack/components/Table/TableWrapper.js
Andrewgdewar: ohhhh... me likey...

webpack/components/Table/MainTable.js
Andrewgdewar: #typescript ;)

package.json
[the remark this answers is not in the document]
evgeni: As you asked for someone with packaging powers: this is eslint, so only really used for development/linting, right? It's not needed during asset compilation? If that's correct: here is a packaging blessing: :dove:

webpack/components/Table/MainTable.js
jeremylenz: PropTypes is poor man's Typescript!  It's just as good :smile:

sjha4:
Opened an issue for weird pagination bug where the table goes into a loop sending API calls with new per_page value and default pagination value. Issue is on master so unrelated to these changes: https://projects.theforeman.org/issues/33276

sjha4:
Tested and everything is working fine..Code looks good to me..
You'll probably need to take care of the commit messages to make this green. Ack pending that.. 👍🏽

sjha4 (approved):
APJ..Wrapper works as expected with composable and on existing non-composable tables. Thanks Jeremy! 👍🏽

webpack/scenes/ContentViews/__tests__/contentViewPage.test.js
sjha4: So weird it wasn't failing on master? 😕

sjha4:
[test katello]

PR #9552, opened 2021-08-11
Fixes #33166 - UI Remote Options for Generic Types

rverdile:
This adds dynamically generated remote options to the UI for generic types for the create repository page and the repository details page.

Test create by going to new repository page and selecting 'python' as the repository type. The remote options will appear (includes, excludes, package types).

Test update by going to the details for a python repository and editing the fields for the remote options.

### DIFF
--- a/app/services/katello/repository_type.rb
+++ b/app/services/katello/repository_type.rb
@@ -16,10 +16,12 @@ module Katello
               :pulp3_skip_publication, :configuration_class, :partial_repo_path, :pulp3_api_class,
               :repositories_api_class, :api_class, :remotes_api_class, :repository_versions_api_class,
               :distributions_api_class, :remote_class, :repo_sync_url_class, :client_module_class,
-              :distribution_class, :publication_class, :publications_api_class, :model_name, :model_version
+              :distribution_class, :publication_class, :publications_api_class, :model_name, :model_version,
+              :url_description
 
     attr_accessor :metadata_publish_matching_check, :index_additional_data_proc
     attr_reader :id, :unique_content_per_repo
+    attr_writer :url_description
 
     def initialize(id)
       @id = id.to_sym
@@ -39,6 +41,15 @@ module Katello
 
     def generic_remote_options
       @generic_remote_options.sort_by(&:name)
+    end
+
+    def translated_generic_remote_options
+      translated = generic_remote_options.deep_dup
+      translated.map do |option|
+        option.title = _(option.title)
+        option.description = _(option.description)
+      end
+      translated
     end
 
     def content_types_to_index
@@ -90,7 +101,9 @@ module Katello
         :name => self.id.to_s,
         :id => self.id,
         :creatable => @allow_creation_by_user,
-        :pulp3_support => SmartProxy.pulp_primary.pulp3_repository_type_support?(self)
+        :pulp3_support => SmartProxy.pulp_primary.pulp3_repository_type_support?(self),
+        :generic_remote_options => translated_generic_remote_options,
+        :url_description => _(@url_description)
       }
     end
 
@@ -145,12 +158,15 @@ module Katello
     end
 
     class GenericRemoteOption
-      attr_accessor :name, :type, :description
+      attr_accessor :name, :title, :type, :description, :input_type, :delimiter
 
       def initialize(options)
         self.name = options[:name]
+        self.title = options[:title]
         self.type = options[:type]
         self.description = options[:description]
+        self.input_type = options[:input_type]
+        self.delimiter = options[:delimiter]
       end
     end
   end
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/new/new-repository.controller.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/new/new-repository.controller.js
@@ -87,6 +87,16 @@ angular.module('Bastion.repositories').controller('NewRepositoryController',
                 }
             });
 
+            $scope.$watch('repository.content_type', function () {
+                $scope.genericRemoteOptions = RepositoryTypesService.getAttribute($scope.repository, "generic_remote_options");
+                $scope.urlDescription = RepositoryTypesService.getAttribute($scope.repository, "url_description");
+                if ($scope.genericRemoteOptions && $scope.genericRemoteOptions !== []) {
+                    $scope.genericRemoteOptions.forEach(function(option) {
+                        option.value = "";
+                    });
+                }
+            });
+
             $scope.handleFiles = function (element) {
                 var reader = new FileReader();
                 reader.addEventListener("loadend", function() {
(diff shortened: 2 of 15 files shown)


ianballou:
Not doing a full review yet, but I used this to sync a Python repo for more testing on another Python PR and it definitely worked!

lib/katello/repository_types/python.rb
```diff
@@ -20,9 +20,9 @@
   publication_class PulpPythonClient::PythonPythonPublication
   publications_api_class PulpPythonClient::PublicationsPypiApi
 
-  generic_remote_option :includes, type: Array, description: "A list containing project specifiers for Python packages to include."
-  generic_remote_option :excludes, type: Array, description: "A list containing project specifiers for Python packages to exclude."
-  generic_remote_option :package_types, type: Array, description: "A list of package types to sync for Python content. Leave blank to get every package type."
+  generic_remote_option :includes, type: Array, description: "A list containing project specifiers for Python packages to include.", input_type: "textarea", delimiter: "\\n"
+  generic_remote_option :excludes, type: Array, description: "A list containing project specifiers for Python packages to exclude.", input_type: "textarea", delimiter: "\\n"
+  generic_remote_option :package_types, type: Array, description: "A list of package types to sync for Python content. Leave blank to get every package type.", input_type: "text", delimiter: ","
```
ianballou, line 25: We should have a list of the available types and a note about what the delimiter character is.

lib/katello/repository_types/python.rb
```diff
@@ -20,9 +20,9 @@
   publication_class PulpPythonClient::PythonPythonPublication
   publications_api_class PulpPythonClient::PublicationsPypiApi
 
-  generic_remote_option :includes, type: Array, description: "A list containing project specifiers for Python packages to include."
-  generic_remote_option :excludes, type: Array, description: "A list containing project specifiers for Python packages to exclude."
-  generic_remote_option :package_types, type: Array, description: "A list of package types to sync for Python content. Leave blank to get every package type."
+  generic_remote_option :includes, type: Array, description: "A list containing project specifiers for Python packages to include.", input_type: "textarea", delimiter: "\\n"
```
ianballou, line 23: The delimiter character should be included here too.

ianballou (commented):
Some initial comments:

rverdile:
Updated descriptions of remote options including specifying delimiter and package types.

rverdile:
Pulp python has several remote options that I haven't included (https://docs.pulpproject.org/pulp_python/restapi.html#operation/remotes_python_python_create). I'm thinking I should include the `prereleases` and `policy` options to create a generic precedent for booleans and drop down selectors.

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/new/new-repository.controller.js
```diff
@@ -169,5 +188,11 @@ angular.module('Bastion.repositories').controller('NewRepositoryController',
                 return OSVersions.osVersionsParam($scope.repository.os_versions);
             };
 
+            $scope.fixRemoteOptionName = function(str) {
```
rverdile, line 191: I'm going to remove this and add a 'title' key to each generic_remote_option. I don't think it's good to assume that the name of the parameter will make the best title in the UI in every case.

rverdile:
Changed to include new "url_description" attribute in repository type definition, which adds help block to upstream URL.

Leaving second suggestion in Justin's comment for another issue since we decided it would fit better there. Changing how text is displayed in the repository info part of the UI will affect all repository types and may require small rework for good result.

jlsherrill:
I realized that i think there are some issues with i18n (internationalization) with the names and descriptions here, gonna add some comments to try to explain some of that

lib/katello/repository_types/python.rb
```diff
@@ -20,9 +20,16 @@
   publication_class PulpPythonClient::PythonPythonPublication
   publications_api_class PulpPythonClient::PublicationsPypiApi
 
-  generic_remote_option :includes, type: Array, description: "A list containing project specifiers for Python packages to include."
-  generic_remote_option :excludes, type: Array, description: "A list containing project specifiers for Python packages to exclude."
-  generic_remote_option :package_types, type: Array, description: "A list of package types to sync for Python content. Leave blank to get every package type."
+  generic_remote_option :includes, title: "Includes", type: Array, input_type: "textarea", delimiter: "\\n",
+                        description: "Python packages to include from the upstream URL, names separated by newline. You may also specify versions, for example: django~=2.0. Leave empty to include every package."
```
jlsherrill, line 24: To properly handle i18n, we need to mark the title and description as translatable, you can do that with this notation:
```
title: N_("Includes")
```

This records the string as translatable, but returns the untranslated string so it can be translated later using _("") in ruby or other constructs in the UI

jlsherrill: and actually i think that the UI wouldn't have the needed translations so we'd need to pass it to the UI at render time, which i think we can do

lib/katello/repository_types/python.rb
```diff
@@ -20,9 +20,16 @@
...
+
+  generic_remote_option :excludes, title: "Excludes", type: Array, input_type: "textarea", delimiter: "\\n",
+                        description: "Python packages to exclude from the upstream URL, names separated by newline. You may also specify versions, for example: django~=2.0."
+
+  generic_remote_option :package_types, title: "Package Types", type: Array, input_type: "text", delimiter: ",",
+                        description: "Package types to sync for Python content, separated by comma. Leave empty to get every package type. Package types are: bdist_dmg, bdist_dumb, bdist_egg, bdist_msi, bdist_rpm, bdist_wheel, bdist_wininst, sdist."
+
+  url_description "URL of a PyPI content source such as https://pypi.org."
```
jlsherrill, line 32: we'd also want to mark this for translation too  N_()

rverdile:
Added N_("") to titles and descriptions.

rverdile:
Removed the remote update fix that will be solved by #9584

rverdile:
[test katello]

app/services/katello/repository_type.rb
```diff
@@ -90,7 +92,9 @@ def as_json(_options = {})
         :name => self.id.to_s,
         :id => self.id,
         :creatable => @allow_creation_by_user,
-        :pulp3_support => SmartProxy.pulp_primary.pulp3_repository_type_support?(self)
+        :pulp3_support => SmartProxy.pulp_primary.pulp3_repository_type_support?(self),
+        :generic_remote_options => @generic_remote_options,
```
jlsherrill, line 96: alright, i think to get the translated strings correctly, we'd want this method to return a translated version of those strings, all it'd need to do is return a copy of @generic_remote_options, but looping through and returning _(option[:title]) and _(option[:description]) ,  You might make a 'translated_generic_remote_options' method to help with this

rverdile:
Added `translated_generic_remote_options`

app/services/katello/repository_type.rb
```diff
@@ -90,7 +101,9 @@ def as_json(_options = {})
         :name => self.id.to_s,
         :id => self.id,
         :creatable => @allow_creation_by_user,
-        :pulp3_support => SmartProxy.pulp_primary.pulp3_repository_type_support?(self)
+        :pulp3_support => SmartProxy.pulp_primary.pulp3_repository_type_support?(self),
+        :generic_remote_options => translated_generic_remote_options,
+        :url_description => @url_description
```
jlsherrill, line 106: need to translate @url_description as well, otherwise this looks good!

jlsherrill (approved):
ACK!!

PR #9556, opened 2021-08-12
Fixes #33254 - CVV errata count wrong with date filter and when errata packages aren't in the source repo

ianballou:
```ruby
if erratum.packages.any? && (erratum.packages.pluck(:filename) - rpm_filenames).empty?
```
made the assumption that, if an erratum has any package filenames that aren't in the source repository, it shouldn't be included in the content view version. This caused low errata counts because errata can include RPMs that aren't in the source repository.

This issue occurs when multiple repositories share an erratum.  One repo could have the erratum and 2 of its 3 packages.  The other repo could have the 1 other package.  Neither repo has all the packages, so the erratum is incorrectly excluded from the content view version.  I fixed this by doing an intersection (`&`) instead of subtraction.

Also, as a secondary issue, the date filters were being mostly ignored.

2 things to test:

First:

1) Sync "Red Hat Enterprise Linux 7 Server - Update Services for SAP Solutions - Optional RPMs x86_64 7.7" and "Red Hat Enterprise Linux 7 Server - Update Services for SAP Solutions RPMs x86_64 7.7".
2) Stick them in a content view. Add an inclusion filter that should include most of the errata (like a date filter).
3) Publish the new content view version. See that many errata are missing (thousands vs hundreds).
4) Patch in my PR and publish again. There should be many more errata now.
5) Check that other content view publishing with filters makes sense.

Second:

1) Sync a repo with errata but no modules (like https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/).
2) Add that repo to a content view.
3) Create a date filter so that only a couple errata are included.
4) Publish, and see that the date filter is ignored.
5) Patch in my PR and publish again. The date filter should be obeyed.
6) Try the same thing with an exclusion filter.
7) Do the same thing with a repo that has errata and modules (like https://fixtures.pulpproject.org/rpm-with-modules/)

Perhaps try testing the first and second cases together too to be safe.

### DIFF
--- a/app/models/katello/erratum.rb
+++ b/app/models/katello/erratum.rb
@@ -106,7 +106,7 @@ module Katello
       query_clauses = clauses.map do |clause|
         "(#{clause.to_sql})"
       end
-      statement = query_clauses.join(" OR ")
+      statement = query_clauses.join(" AND ")
 
       Katello::ErratumPackage.joins(:erratum => :repository_errata).
           where("#{RepositoryErratum.table_name}.repository_id" => repo.id).
@@ -117,7 +117,7 @@ module Katello
       query_clauses = clauses.map do |clause|
         "(#{clause.to_sql})"
       end
-      statement = query_clauses.join(" OR ")
+      statement = query_clauses.join(" AND ")
       ModuleStream.where(:id => ModuleStreamErratumPackage.joins(:erratum_package => {:erratum => :repository_errata}).
           where("#{RepositoryErratum.table_name}.repository_id" => repo.id).
           where(statement).select("#{ModuleStreamErratumPackage.table_name}.module_stream_id"))
--- a/app/services/katello/pulp3/repository/yum.rb
+++ b/app/services/katello/pulp3/repository/yum.rb
@@ -319,6 +319,11 @@ module Katello
             dest_repo_map[:content_unit_hrefs] = content_unit_hrefs.uniq.sort
           end
 
+          errata_to_include = errata_to_include_from_map(repo_id_map)
+          repo_id_map.each do |_, dest_repo_map|
+            dest_repo_map[:content_unit_hrefs] |= errata_to_include.flat_map { |erratum| erratum.repository_errata.pluck(:erratum_pulp3_href) }
+          end
+
           dependency_solving = options[:solve_dependencies] || false
 
           multi_copy_units(repo_id_map, dependency_solving)
@@ -353,11 +358,17 @@ module Katello
           blacklist_ids += modular_packages(source_repository, exclusion_modular_filters) unless exclusion_modular_filters.empty?
           content_unit_hrefs = whitelist_ids - blacklist_ids
           if content_unit_hrefs.any?
-            content_unit_hrefs += additional_content_hrefs(source_repository, content_unit_hrefs)
+            content_unit_hrefs += additional_content_hrefs(source_repository, content_unit_hrefs, true)
           end
           content_unit_hrefs += source_repository.srpms.pluck(:pulp_id)
           dependency_solving = options[:solve_dependencies] || false
           copy_units(source_repository, content_unit_hrefs.uniq, dependency_solving)
+        end
+
+        def errata_to_include_from_map(repo_id_map)
+          all_errata = ::Katello::Erratum.in_repositories(repo_id_map.keys.flatten.uniq)
+          all_content_units = repo_id_map.values.pluck(:content_unit_hrefs).flatten.uniq
+          filter_errata_by_pulp_href(all_errata, all_content_units)
         end
 
         def modular_packages(source_repository, filters)
@@ -368,12 +379,14 @@ module Katello
           list_ids
         end
 
-        def additional_content_hrefs(source_repository, content_unit_hrefs)
+        def additional_content_hrefs(source_repository, content_unit_hrefs, copy_errata = false)
           repo_service = source_repository.backend_service(SmartProxy.pulp_primary)
           options = { :repository_version => source_repository.version_href }
 
-          errata_to_include = filter_errata_by_pulp_href(source_repository.errata, content_unit_hrefs)
-          content_unit_hrefs += errata_to_include.collect { |erratum| erratum.repository_errata.pluck(:erratum_pulp3_href) }.flatten
+          if copy_errata
+            errata_to_include = filter_errata_by_pulp_href(source_repository.errata, content_unit_hrefs)
+            content_unit_hrefs += errata_to_include.collect { |erratum| erratum.repository_errata.pluck(:erratum_pulp3_href) }.flatten
+          end
 
           package_groups_to_include = filter_package_groups_by_pulp_href(source_repository.package_groups, content_unit_hrefs)
           content_unit_hrefs += package_groups_to_include.pluck(:pulp_id)
--- a/test/fixtures/vcr_cassettes/actions/pulp3/copy_all_unit_yum_package_groups_repository/package_groups_as_a_filter_rule.yml
+++ b/test/fixtures/vcr_cassettes/actions/pulp3/copy_all_unit_yum_package_groups_repository/package_groups_as_a_filter_rule.yml
@@ -2,168 +2,7 @@
 http_interactions:
 - request:
     method: get
-    uri: https://devel2.balmora.example.com/pulp/api/v3/repositories/rpm/rpm/?name=2_duplicate
-    body:
-      encoding: US-ASCII
-      base64_string: ''
-    headers:
-      Content-Type:
-      - application/json
-      User-Agent:
-      - OpenAPI-Generator/3.13.3/ruby
-      Accept:
-      - application/json
-      Authorization:
-      - Basic Og==
-      Accept-Encoding:
-      - gzip;q=1.0,deflate;q=0.6,identity;q=0.3
-  response:
-    status:
-      code: 200
-      message: OK
-    headers:
-      Date:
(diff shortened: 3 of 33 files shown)


ianballou:
The intersection that I'm doing currently won't work because it causes errata to be included when it shouldn't.  The real fix will likely be to incorporate content units from all source repos when calculating the `dest_repo_map` in `copy_content_from_mapping`.

Since `copy_content_from_mapping` goes source repo by source repo to calculate the hrefs, errata that have some rpms in different repositories will be completely excluded.

jturel:
> The intersection that I'm doing currently won't work because it causes errata to be included when it shouldn't. The real fix will likely be to incorporate content units from all source repos when calculating the `dest_repo_map` in `copy_content_from_mapping`.

Does this PR reflect the 'real fix' now?

ianballou:
@jturel yes indeed, it's ready now.

jjeffers (commented):
NVM!

> ```ruby
> if erratum.packages.any? && (erratum.packages.pluck(:filename) - rpm_filenames).empty?
> ```
> 
> [...]Neither repo has all the packages, so the erratum is incorrectly excluded from the content view version. I fixed this by doing an intersection (`&`) instead of subtraction.

I think I missed where this intersection was done.

app/services/katello/pulp3/repository/yum.rb
```diff
@@ -368,12 +381,14 @@ def modular_packages(source_repository, filters)
           list_ids
         end
 
-        def additional_content_hrefs(source_repository, content_unit_hrefs)
+        def additional_content_hrefs(source_repository, content_unit_hrefs, copy_errata = false)
```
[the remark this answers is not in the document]
ianballou: I don't want the errata copied here since I copy it later: https://github.com/Katello/katello/pull/9556/files#diff-893a1c298832d5213af1a68f9aaae82f2507bd2414371bce47fa7c962d477089R312

ianballou:
> NVM!
> 
> > ```ruby
> > if erratum.packages.any? && (erratum.packages.pluck(:filename) - rpm_filenames).empty?
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > [...]Neither repo has all the packages, so the erratum is incorrectly excluded from the content view version. I fixed this by doing an intersection (`&`) instead of subtraction.
> 
> I think I missed where this intersection was done.

No worries about that, it was my old solution that was still broken.

app/services/katello/pulp3/repository/yum.rb
```diff
@@ -319,6 +319,11 @@ def copy_content_from_mapping(repo_id_map, options = {})
             dest_repo_map[:content_unit_hrefs] = content_unit_hrefs.uniq.sort
           end
 
+          errata_to_include = errata_to_include_from_map(repo_id_map)
+          repo_id_map.each do |_, dest_repo_map|
+            dest_repo_map[:content_unit_hrefs] |= errata_to_include.collect { |erratum| erratum.repository_errata.pluck(:erratum_pulp3_href) }.flatten
```
jturel, line 324: I think you could use flat_map here, but no need to change

app/services/katello/pulp3/repository/yum.rb
```diff
@@ -353,13 +358,21 @@ def copy_content_for_source(source_repository, options = {}) # rubocop:disable M
...
+            content_unit_hrefs += additional_content_hrefs(source_repository, content_unit_hrefs, true)
           end
           content_unit_hrefs += source_repository.srpms.pluck(:pulp_id)
           dependency_solving = options[:solve_dependencies] || false
           copy_units(source_repository, content_unit_hrefs.uniq, dependency_solving)
         end
 
+        def errata_to_include_from_map(repo_id_map)
+          all_errata = ::Katello::Erratum.joins("inner join katello_repository_errata on katello_repository_errata.erratum_id = katello_errata.id").
+            joins("inner join katello_repositories on katello_repositories.id = katello_repository_errata.repository_id").
```
jturel, line 370: We typically do `$MODEL.table_name` rather than hard coding the table names

app/services/katello/pulp3/repository/yum.rb
```diff
@@ -353,13 +358,21 @@ def copy_content_for_source(source_repository, options = {}) # rubocop:disable M
...
           if content_unit_hrefs.any?
-            content_unit_hrefs += additional_content_hrefs(source_repository, content_unit_hrefs)
+            content_unit_hrefs += additional_content_hrefs(source_repository, content_unit_hrefs, true)
           end
           content_unit_hrefs += source_repository.srpms.pluck(:pulp_id)
           dependency_solving = options[:solve_dependencies] || false
           copy_units(source_repository, content_unit_hrefs.uniq, dependency_solving)
         end
 
+        def errata_to_include_from_map(repo_id_map)
```
jturel, line 368: I was staring at this query for a while because I like to see how we can break ones like it (multiple, hand-written joins) into a more railsy approach which may also lend us some nice utility later. I came up with this modification to errata.rb

```
@@ -16,6 +16,7 @@ module Katello
     has_many :bugzillas, :class_name => "Katello::ErratumBugzilla", :dependent => :destroy, :inverse_of => :erratum
     has_many :cves, :class_name => "Katello::ErratumCve", :dependent => :destroy, :inverse_of => :erratum
     has_many :packages, :class_name => "Katello::ErratumPackage", :dependent => :destroy, :inverse_of => :erratum
+    has_many :repository_errata, :class_name => "Katello::RepositoryErratum", :dependent => :destroy, :inverse_of => :erratum
 
     scoped_search :on => :errata_id, :only_explicit => true
     scoped_search :on => :errata_id, :rename => :id, :complete_value => true, :only_explicit => true
@@ -50,6 +51,7 @@ module Katello
     scope :enhancement, -> { of_type(Erratum::ENHANCEMENT) }
     scope :modular, -> { where(:id => joins(:packages => :module_stream_errata_packages)) }
     scope :non_modular, -> { where.not(:id => modular) }
+    scope :with_errata_by_repo_ids, ->(repo_ids) { joins(:repository_errata).merge(Katello::RepositoryErratum.where(repository_id: repo_ids)) }

```

For someone like me this is way easier to understand! Now you can re-implement this method to use the new scope:

```
all_errata = Katello::Erratum.with_errata_by_repo_ids(repo_id_map.keys.flatten.uniq)
all_content_units  = ...
```

Let me know what you think - seems it generates an identical query to what you came up with

jturel: I don't think the scope name is very good btw- maybe `in_repositories` is better?

jturel:
I tested the first scenario and it works. Will test the second set of scenarios if you'd like to include the suggestions I made around some of the querying

app/services/katello/pulp3/repository/yum.rb
ianballou: I think I like `in_repositories`, I'll give it a good stare and see if I come up with anything else. Going to try out the joins, it looks much cleaner to me too.

ianballou:
> I tested the first scenario and it works. Will test the second set of scenarios if you'd like to include the suggestions I made around some of the querying

Sounds fine to me to include your suggestion in your testing.

app/services/katello/pulp3/repository/yum.rb
jturel: Turns out `repository_errata` is defined through the PulpDatabaseUnit concern. so you really just need the new scope!

jturel: May be worth adding a .distinct to the query too?

app/models/katello/erratum.rb
```diff
@@ -16,6 +16,7 @@ class Erratum < Katello::Model
     has_many :bugzillas, :class_name => "Katello::ErratumBugzilla", :dependent => :destroy, :inverse_of => :erratum
     has_many :cves, :class_name => "Katello::ErratumCve", :dependent => :destroy, :inverse_of => :erratum
     has_many :packages, :class_name => "Katello::ErratumPackage", :dependent => :destroy, :inverse_of => :erratum
+    has_many :repository_errata, :class_name => "Katello::RepositoryErratum", :dependent => :destroy, :inverse_of => :erratum
```
jturel, line 19: Turns out this is defined already by PulpDatabaseUnit and adding this here breaks stuff, so you can just remove it!

app/models/katello/erratum.rb
```diff
@@ -50,6 +51,7 @@ def self.of_type(type)
     scope :enhancement, -> { of_type(Erratum::ENHANCEMENT) }
     scope :modular, -> { where(:id => joins(:packages => :module_stream_errata_packages)) }
     scope :non_modular, -> { where.not(:id => modular) }
+    scope :in_repositories, -> (repo_ids) { joins(:repository_errata).merge(Katello::RepositoryErratum.where(repository_id: repo_ids)) }
```
jturel, line 54: It probably makes sense to add .distinct here too (I tested it that way)

jturel:
This change works perfectly! Will ack once it gets one more push with the last tweaks

ianballou:
@jturel thanks for the comments, I've fixed them up.

jturel (approved):
APJ!

jturel:
rubocop :roll_eyes:

ianballou:
Looks like we don't even need the scope, `in_repositories` already exists!

jturel:
> 
> 
> Looks like we don't even need the scope, `in_repositories` already exists!

Wow, I missed that - sorry for making this PR more laborious!

PR #9557, opened 2021-08-16
Pin @types/jest package

sjha4:


sjha4:
Closing. Fix has to be in foreman: https://github.com/theforeman/foreman/pull/8716

PR #9558, opened 2021-08-16
Fixes #31590 - Add last changed timestamp

maccelf:
When you look at a repo via the api (GET /katello/api/repositories/:id), the data does include a "last_synced" timestamp, but no indication about "changes".
last sync could be a noop (because upstream didn't change) or not set at all (because the repo doesn't have an upstream at all), but it would be good to know (e.g. for automated publishes of CVs) when the last change happened.


jturel:
[test katello]

jturel (approved):
Looks like a good enhancement to me! I'll merge once the test build finishes and passes

jturel:
Thanks @maccelf !

PR #9562, opened 2021-08-18
Fixes #33286 - uses unscoped find query for smart proxy

jjeffers:
This change fixes an issue where a location that has no smart proxy assigned is used when creating a new repository. Metadata generation will fail as the RefreshDistribrution attempted to query (via a location scope `find`) a smart proxy by id.

### DIFF
--- a/app/lib/actions/pulp3/repository/refresh_distribution.rb
+++ b/app/lib/actions/pulp3/repository/refresh_distribution.rb
@@ -6,13 +6,16 @@ module Actions
         middleware.use Actions::Middleware::ExecuteIfContentsChanged
 
         def plan(repository, smart_proxy, options = {})
-          smart_proxy = SmartProxy.find_by(id: smart_proxy) #support bulk actions
+          smart_proxy = SmartProxy.unscoped.find_by(id: smart_proxy) #support bulk actions
           sequence do
             if !repository.unprotected && !options[:assume_content_guard_exists]
               plan_action(::Actions::Pulp3::ContentGuard::Refresh, smart_proxy)
             end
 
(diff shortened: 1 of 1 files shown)


jturel:
@jjeffers did you mean 'smart proxy with no location assigned' vs 'location with no smart proxy assigned' (wanna make sure I'm testing the right thing)

jjeffers:
> @jjeffers did you mean 'smart proxy with no location assigned' vs 'location with no smart proxy assigned' (wanna make sure I'm testing the right thing)

In this case, "location with no smart proxy assigned". The original issue is reproduced by:
1. Create a new organization
2. Create a new location, adding the organization to the new location.
3. Note that the current location is now the new location.
4. Create a new product.
5. Create a new repository.

The error will be `undefined method `id' for nil:NilClass` in the refresh distribution plan.

jturel:
Alright, this change makes sense! Two questions:

- Do we need to audit the code for more cases like this where a default scope might be affecting our SmartProxy lookups? Separate Redmine of course.

- I was able to create a new repo without an upstream URL no problem. When I create one per the above scenario with an upstream URL it times out with everything stuck in Pending state while in Run. Does it work for you?

jjeffers:
> Alright, this change makes sense! Two questions:
> 
> * Do we need to audit the code for more cases like this where a default scope might be affecting our SmartProxy lookups? Separate Redmine of course.

I think so. This was a specific update from about 18 months ago, and is now a regression. But I think a review of where this is being done is an order.

jturel (approved):
ACK

I filed https://projects.theforeman.org/issues/33308 to track looking for other such cases

PR #9567, opened 2021-08-19
Fixes #33164 - Allow clearing upstream auth and password

sjha4:
To reproduce:
Create a yum repo with upstream username/password set.
Try to update it after clearing the fields and save.

Same behavior exists on auth_url/token fields for collection repos.
Create a repo with auth_url set and token empty.
See the expected error.
Set a token and clear the auth_url
The error you see on this is invalid and shouldn't happen on this branch.

### DIFF
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/repository-details-info.controller.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/repository-details-info.controller.js
@@ -90,6 +90,7 @@ angular.module('Bastion.repositories').controller('RepositoryDetailsInfoControll
 
             $scope.save = function (repository, saveUpstreamAuth) {
                 var deferred = $q.defer();
+                var fields = ['upstream_password', 'upstream_username', 'ansible_collection_auth_token', 'ansible_collection_auth_url', 'ansible_collection_requirements'];
                 if (repository.content_type === 'yum' && typeof repository.ignore_srpms !== 'undefined') {
                     if (repository['ignore_srpms']) {
                         repository['ignorable_content'] = ["srpm"];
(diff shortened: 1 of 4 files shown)


jturel:
[test katello]

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/repository-details-info.controller.js
```diff
@@ -110,7 +110,22 @@ angular.module('Bastion.repositories').controller('RepositoryDetailsInfoControll
                 } else {
                     repository["docker_tags_whitelist"] = [];
                 }
-                /* eslint-disable camelcase */
+              /* eslint-disable camelcase */
+                if (repository.upstream_password === '') {
```
jturel, line 114: Maybe an unnecessary optimization but it might be good to maintain a list of these that we want to check in this same manner...

```
var fields = ['upstream_password', ...]

angular.forEach(fields, function(field) {
   if (repository[field] === '') {
     repository[field] = null;
   }
});
```

jturel (commented):
Looks good!

I noticed something weird which may be a separate issue but I'm OK to review it as part of this PR if you want to fix here: when adding Ansible Collection Auth fields when neither are currently set, typing in the URL disables the save button but filling in the Token field doesn't re-enable the save button :-1:

sjha4:
> Looks good!
> 
> I noticed something weird which may be a separate issue but I'm OK to review it as part of this PR if you want to fix here: when adding Ansible Collection Auth fields when neither are currently set, typing in the URL disables the save button but filling in the Token field doesn't re-enable the save button

Ya..this is because there's a UI validation that the Auth URL needs to be a URL..Save is enabled when the input value is a URL.

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/repository-details-info.controller.js
sjha4: Also gets rid of all if blocks around this and makes change cleaner..Updated..

jturel (approved):
Works well! APJ

PR #9570, opened 2021-08-19
Fixes #32664 - ignore deleting the candlepin instance with missing cp_id on activation key delete

lfu:
When creating an activation key, users may run into situation where the instance is created on Katello but not on Candlepin.
So this Katello activation key has `cp_id = nil` which makes it impossible to delete this object.

To Test
1. Create an activation key from Web UI
2. Set `cp_id` to `nil` for this activation key from rails console
3. `hammer activation-key delete --organization-id 1 --id <activation_key_id>

### DIFF
--- a/test/actions/katello/activation_key_test.rb
+++ b/test/actions/katello/activation_key_test.rb
@@ -70,12 +70,22 @@ module ::Actions::Katello::ActivationKey
   class DestroyTest < TestBase
     let(:action_class) { ::Actions::Katello::ActivationKey::Destroy }
 
-    it 'plans' do
+    it 'does not plan with missing cp_id' do
       action = create_action(action_class)
       action.expects(:plan_self)
       action.expects(:action_subject).with(activation_key)
       plan_action(action, activation_key)
-      assert_action_planed(action, ::Actions::Candlepin::ActivationKey::Destroy)
+      refute_action_planed(action, ::Actions::Candlepin::ActivationKey::Destroy)
+      assert_nil(activation_key.cp_id)
+    end
+
+    it 'plans' do
+      action = create_action(action_class)
+      activation_key.update!(cp_id: 'something')
+      action.expects(:plan_self)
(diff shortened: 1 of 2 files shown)


lfu:
[test katello]

test/actions/katello/activation_key_test.rb
```diff
@@ -70,12 +70,23 @@ class UpdateWithoutCandlepinTest < TestBase
...
+      action.expects(:plan_self)
+      action.expects(:action_subject).with(activation_key)
+      plan_action(action, activation_key)
+      refute_action_planed(action, ::Actions::Candlepin::ActivationKey::Destroy)
+      assert_equal(activation_key.cp_id, nil)
+    end
+
     it 'plans' do
       action = create_action(action_class)
+      activation_key.update!(cp_id: 'something')
```
jturel, line 84: The way you changed the test makes me think that the value of cp_id is significant since you later assert that's what the value is, but it doesn't seem like that's the case since the test doesn't actually _use_ that value anywhere

Two ways to (imo) improve this test:
- either refute_nil activation_key.cp_id (ie this is a precondition of this test)
- or assert that ::Actions::Candlepin::ActivationKey::Destroy was planned with a particular cp_id - the one from the activation key

lfu: A present `activation_key.cp_id` is required to schedule the plan with this change:
```
plan_action(Candlepin::ActivationKey::Destroy, cp_id: activation_key.cp_id) if !skip_candlepin && activation_key.cp_id.present?
```

jturel: OK, in that case I think the path forward is:

- keep the update! call as-is
- instead of asserting down below that the cp_id is the value you set, rather test the params given to the `::Actions::Candlepin::ActivationKey::Destroy` action reflect the cp_id (tons of examples of this in the test code)

PR #9571, opened 2021-08-20
Fixes #33114 - Bulk Add/Remove filters rules

sjha4:


webpack/scenes/ContentViews/Details/ContentViewDetailActions.js
```diff
@@ -27,7 +27,7 @@ import {
   cvRemoveComponentKey,
   addComponentSuccessMessage,
   removeComponentSuccessMessage,
-  cvVersionPromoteKey,
+  cvVersionPromoteKey, DELETE_CONTENT_VIEW_FILTER_RULES_KEY,
```
Andrewgdewar, line 30: For the love of cheese, can this me inline?

app/controllers/katello/api/v2/content_view_filters_controller.rb
```diff
@@ -82,6 +82,49 @@ def destroy
...
+        rules += (rule_params[:errata_ids] || []).map do |errata_id|
+          rule_clazz.create!(rule_params.except(:errata_ids)
+                                 .merge(filter: @filter, errata_id: errata_id))
+        end
+        if rule_params[:module_stream_ids]
+          rules += (rule_params[:module_stream_ids] || []).map do |module_stream_id|
+            rule_clazz.create!(rule_params.except(:module_stream_ids)
+                                   .merge(filter: @filter, module_stream_id: module_stream_id))
+          end
+        end
```
jlsherrill, line 118: how many rules do you see being created at once?   I wonder if we should use active record import to do a bulk insert?

sjha4: I would say a maximum of 50 (max records per page) if a user selects everything on a page and adds, given multi page selection is not supported..I am not sure if people do that though with filter rules..I imagine only a few package groups added to a filter cause more than that would be a really complicated content view.

app/controllers/katello/api/v2/content_view_filters_controller.rb
```diff
@@ -82,6 +82,49 @@ def destroy
...
+      rule_clazz = ContentViewFilter.rule_class_for(@filter)
+      rule_clazz.where(id: params[:rule_ids]).try(:destroy_all)
+      respond_for_show(resource: @filter)
+    end
+
+    api :put, "/content_views/:content_view_id/filters/:id/add_filter_rules", N_("bulk add filter rules")
+    api :put, "/content_view_filters/:id/add_filter_rules", N_("bulk add filter rules")
+    param :content_view_id, :number, :desc => N_("content view identifier")
+    param :id, :number, :desc => N_("filter identifier"), :required => true
+    param :rules_params, Array, :desc => N_("Rules to be added"), :required => false
```
jlsherrill, line 100: we should probably include the params that this accepts, similar to this:

```
    param :uploads, Array, :desc => N_("Array of uploads to import") do
      param 'id', String, :required => true
      param 'content_unit_id', String
      param 'size', String
      param 'checksum', String
      param 'name', String, :desc => N_("Needs to only be set for file repositories or docker tags")
      param 'digest', String, :desc => N_("Needs to only be set for docker tags")
    end
```

app/controllers/katello/api/v2/content_view_filters_controller.rb
jlsherrill: i guess we can hold off for now, easy to add later on

app/controllers/katello/api/v2/content_view_filters_controller.rb
sjha4: Updated! 👍🏽

sjha4:
[test katello]

jlsherrill (approved):

PR #9573, opened 2021-08-20
Fixes #33315 - Redefine pulp3 dynflow step statuses

rverdile:
This fixes the dynflow statuses printed in the dynflow console when the task is in the suspended state. Previously, pulp tasks were not being correctly identified as "started" or "finished'. 

Also, the status message has been updated to give slightly more precise details. For example, rather than "waiting for Pulp to finish the task" it might say "waiting for Pulp to finish the task synchronize (53ef3ad)". The characters at the end are part of the pulp href, to give the task an identifier.

To see the messages, do a complete sync on a decent sized repo (like a centos repo), and monitor the dynflow console while it's syncing. If that doesn't work, publish a content view with several decent sized repos and filters. Monitor the dynflow console while it's publishing.

### DIFF
--- a/app/lib/actions/pulp3/abstract_async_task.rb
+++ b/app/lib/actions/pulp3/abstract_async_task.rb
@@ -20,10 +20,16 @@ module Actions
             _("checking Pulp task status")
           end
         when :suspended
-          if combined_tasks.any?(&:started?)
-            _("waiting for Pulp to finish the task")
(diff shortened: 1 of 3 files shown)


jjeffers:
I am able to observe the dynflow message updates!

`Actions::Pulp3::Repository::Sync (waiting for Pulp to finish the task synchronize (ef98512daa94))`

PR #9574, opened 2021-08-21
Fixes #28491 - Move stuff from base.json.rabl to show.json.rabl

maccelf:
Move stuff from content_view_versions/base.json.rabl to content_view_versions/show.json.rabl to reduce call time of index.
List page should need less details than details page

### DIFF
--- a/app/views/katello/api/v2/content_view_versions/base.json.rabl
+++ b/app/views/katello/api/v2/content_view_versions/base.json.rabl
@@ -13,10 +13,6 @@ node do |version|
   version.content_counts_map
 end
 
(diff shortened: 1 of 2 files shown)


jturel:
@parthaa can you take a look here? You filed the initial Redmine but we kept moving it from release to release because there was some kind of hidden cost/effort involved but it wasn't mentioned on the Redmine.  Do you remember what that was?

jjeffers:
@maccelf I applied your patch and I noticed this on a CV view details page:
image

Can you recheck the changes and adjust accordingly?

PR #9576, opened 2021-08-24
Fixes #33326 - use uncached when searching for docker tags

jlsherrill:
this attempts to fix an error seen during syncing lots of
docker repos at once with lots of tags.  Following the logic
it seems that the tags were inserted but for some reason this query
caused an empty values array to be returned by the group_by.  One theory
is that it was caused by a query cache similar to what we have seen
with errata applicability insertions


jlsherrill:
This doesn't actually help

PR #9578, opened 2021-08-25
Fixes #32816 - Leave smart proxy sync in warning if gateway container update fails

rverdile:
Previously, during a smart proxy sync if a gateway container update failed the whole sync would fail, because the container update was happening during planning.

This moves the container update from the planning phase to the running phase, so if the update fails, the smart proxy sync will end with a warning.

One way to test is creating a smart proxy to test with and adding it in Infrastructure > Smart Proxy. Then put `fail [fail message]` in the `sync_container_gateway` method of `sync_capsule.rb`. When syncing the smart proxy, the task should fail but be left in the warning state.

### DIFF
--- a/app/models/katello/concerns/smart_proxy_extensions.rb
+++ b/app/models/katello/concerns/smart_proxy_extensions.rb
@@ -4,6 +4,7 @@ require 'proxy_api/pulp_node'
 
 module Katello
   module Concerns
+    # rubocop:disable Metrics/ModuleLength
     module SmartProxyExtensions
       extend ActiveSupport::Concern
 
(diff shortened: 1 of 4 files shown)


ianballou:
Looking good to me! I agree with @jlsherrill that we should move those two methods.

rverdile:
@ianballou updated

app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -110,18 +111,44 @@ def self.sync_needed?(environment)
...
+            next if repo.container_repository_name.nil?
+            inner_repo_list << { repository: repo.container_repository_name,
+                                 auth_required: !unauthenticated_container_repositories.include?(repo.id) }
+          end
+          user_repo_map[:users] << { user.login => inner_repo_list }
+        end
         ProxyAPI::ContainerGateway.new(url: self.url).user_repository_mapping(user_repo_map)
       end
 
+      def update_unauthenticated_repo_list(repo_names)
```
ianballou, line 148: I believe this method can actually be removed. We aren't using it anymore.

test/models/concerns/smart_proxy_extensions_test.rb
```diff
@@ -69,6 +71,21 @@ def test_rhsm_url_pulp_primary
     def test_rhsm_url_pulp_mirror
       assert_includes @proxy_mirror.rhsm_url, ":8443/rhsm"
     end
+
+    def test_sync_container_gateway
```
ianballou, line 75: This test is no longer testing arguments like it used to.  I think we could still test the args by not stubbing out `:update_container_repo_list` and `:update_container_repo_list` and instead putting expectations on `ProxyAPI::ContainerGateway.new(url: self.url).repository_list` and `ProxyAPI::ContainerGateway.new(url: self.url).user_repository_mapping(user_repo_map)`.   Those methods could have their input args tested like the old test did.

test/models/concerns/smart_proxy_extensions_test.rb
```diff
@@ -69,6 +71,21 @@ def test_rhsm_url_pulp_primary
...
+      environment = katello_environments(:library)
+
+      with_pulp3_features(capsule_content.smart_proxy)
+      capsule_content.smart_proxy.add_lifecycle_environment(environment)
+
+      capsule_content.smart_proxy.expects(:update_container_repo_list).returns(true)
+      capsule_content.smart_proxy.expects(:update_user_container_repo_mapping).returns(true)
+      capsule_content.smart_proxy.expects(:container_gateway_users).returns(::User.where(login: 'secret_admin'))
+
+      capsule_content.smart_proxy.update_container_repo_list
```
ianballou, line 85: You could probably move `sync_container_gateway` into the smart proxy extensions as well.  Then you could run it directly.

test/models/concerns/smart_proxy_extensions_test.rb
```diff
@@ -69,6 +71,21 @@ def test_rhsm_url_pulp_primary
...
+      users = capsule_content.smart_proxy.update_user_container_repo_mapping
```
ianballou, line 86: `users` should be fetched from `smart_proxy.container_gateway_users`.  That won't matter though if you run `sync_container_gateway` directly instead of running its individual methods.

ianballou:
Once the test is updated I'm all good with this. The PR is working well now with the container gateway

rverdile:
@ianballou updated

rverdile:
[test katello]

ianballou (approved):
Looks good to me! Those errors are happening on master too...

rverdile:
[test katello]

rverdile:
[test katello]

rverdile:
[test katello]

PR #9587, opened 2021-08-27
Fixes #33365 - When the package group names are empty, reject that package group.

jjeffers:
This change addresses a bug where a package group with no package group names would get copied incorrectly into a new content view version after publication.

To reproduce: 
```
Pulp 3 versions:

pulp-2to3-migration (0.11.1)
pulp-certguard (1.0.3)
pulp-container (2.1.2)
pulp-file (1.3.0)
pulp-rpm (3.10.0)
pulpcore (3.7.6)

Steps to Reproduce:
1. Sync 'Red Hat Enterprise Linux 8 for x86_64 - BaseOS RPMs 8'
2. Make a content view with the BaseOS repo and an RPM include filter that includes 'vim-minimal' with all architectures & versions
3. Publish the content view
```


ianballou (approved):
Worked for me!

PR #9592, opened 2021-08-30
Fixes #33374 - Hostgroup now shows content fields

parthaa:
Before this PR
* Configure -> Hostgroup
* Create host group
This would not show cv, lce and cs.
This error occurs because
hostgroup_puppet_environment_field was removed
in foreman.

This PR fixes this issue by changing the insert before
to the appropriate field .


jturel (approved):
APJ!

I tested a bunch of stuff around hostgroups w.r.t the katello fields (including nesting), updating them, etc. and it all works well with no errors in the browser console

parthaa:
[test katello]

parthaa:
[test katello]

parthaa:
[test katello]

PR #9594, opened 2021-08-31
Fixes #33078 - Add Traces bulk select & restart

jeremylenz:
This adds bulk table row selection to the Traces tab, and allows you to restart traces via remote execution.

- [x] finish tests

Added a `<SelectAllCheckbox>` with the following features:
* Select multiple traces across multiple pages
* Select all, select page, select none options
* Grammatically correct text re: how many items are selected
* The Select all option is not implemented yet, since that will require Katello-wide changes. However, the checkbox is still accurately checked when all items are selected

select_options
one_selected
multi_page_select
all_selected

none_selected


webpack/components/extensions/HostDetails/Tabs/HostTracesActions.js
```diff
@@ -1,12 +1,47 @@
...
+      dispatch({
+        type: 'TOASTS_ADD',
+        payload: {
+          key: id,
+          message: {
+            type: 'success',
+            message: `Restarting ${pluralize(traceCount, 'trace')}.`,
+            link: {
+              children: 'View task',
+              href: `/foreman_tasks/tasks/${id}`,
```
jeremylenz, line 38: This took me like 2 hours to figure out.. I didn't know about `webpack/scenes/Tasks/helpers.js` :sob:  I will update this

jeremylenz:
To test:

1. Get the host ID of your host
2. In rails console,
```rb
100.times { |idx| Katello::HostTracer.create!(host_id: 12, application: "tracer_#{idx}", helper: "sudo systemctl restart blah", app_type: "daemon") }
```

To test end-to-end:

1. Set up remote execution on your development environment
2. Run a `yum update` on your client - you'll probably get some Traces
3. Restart one or more traces (note: the task might hang, this is unrelated)
4. Check `journalctl -f` on the client for some `sshd` activity
5. Refresh the page and verify the traces are no longer listed

jeremylenz:
This PR uses JavaScript `Set`s to keep track of which traces are selected. 

Advantages:
1. The API for Sets is beautiful - you can do
```js
const mySet = new Set([1,2,3])
mySet.add(3) // no change; 3 is already in the set
mySet.delete(2) // Set { 1, 3 }
mySet.size // 2
mySet.has(2) // false
mySet.has(1) // true
[...mySet] // [1, 3] - returns the Set as an array
mySet.clear() // Set { }
```
2. When adding items, you don't have to worry about if they're already in the set
3. When removing items, you don't have to iterate through the set, filter the set, or find the item to remove
4. (according to MDN) `mySet.has()` is faster than `myArray.includes()` for a set / array of the same size

Disadvantages:
1. Working with Sets requires that you mutate a value directly. Therefore, it doesn't work well with React `useState`. For this reason, I used `useRef` so that you would have a mutable value.
2. However, changes to a Ref do _not_ mean that props or state have changed, and therefore React will not automatically rerender when our Set changes. To get around this, I made the custom `useSet` hook which returns a `forceRender()` function that you must manually run every time you update the Set. 

Mostly because of the above, I'm still ambivalent  about whether this is a good approach. But I do really love being able to use the Set API for what it's meant for -- keeping track of a unique list of items and modifying that list easily.

MariSvirik:
Hey, it looks great.  I'm really glad we managed to do this.
I have a couple of comments:

> select_options

- No text is needed where there is nothing selected (even if it's no longer selected). It also saves space within a toolbar. 
Screenshot 2021-09-03 at 11 09 44

- Correct order in the list should be always hierarchical: 1. Select none (0) 2. Select page (nb.) 3. Select all (nb) 
- Menu items should always contain a number of selected items so a user is aware of how many they are going to select.

> one_selected

- Don't use 'traces' in the bulk selector even though it may be grammatically correct. No space to spare.
- Keep partially checked checkbox blue
Screenshot 2021-09-03 at 11 09 51

- Restart button should be secondary (blue outlined) not red. We almost never use red buttons anymore unless it's an extremely destructive action such as Delete my entire system which always comes with extra confirmation.
For more information see [PF4 guidelines](https://www.patternfly.org/v4/guidelines/bulk-selection/)
**[Padding](https://www.patternfly.org/v4/components/toolbar/design-guidelines)**
24px-bulk selector- 24px - search input - 16px - button ----- pagination (numbers) -8px- pagination (carets)-24 px.

jeremylenz:
Thanks for looking at it @MariSvirik :)

> Menu items should always contain a number of selected items so a user is aware of how many they are going to select.

Not sure what is meant by this?  Can you provide examples of correct wording for the menu items?  "Select all / select page / select none" was directly from the Patternfly doc.

> Keep partially checked checkbox blue

The partially checked state being non-blue was just the way the Patternfly React component came. I didn't change anything, maybe it's being overridden with our CSS or something.. I'll look into it.

Also, will update to address all your other comments.

webpack/components/extensions/HostDetails/Tabs/TracesTab.js
```diff
@@ -43,20 +85,29 @@ const TracesTab = () => {
...
+        areAllRowsSelected={areAllRowsSelected}
         variant={TableVariant.compact}
         status={status}
         metadata={meta}
+        selectedCount={selectedTraces.size}
+        modelName="trace"
       >
         <Thead>
           <Tr>
+            <Th />
```
Andrewgdewar, line 103: Empy row, but why?

jeremylenz: It's the checkbox column header!

sjha4: I can't understand in the code how the select all is working..Would you need any updates to the rails traces controller or is it already supported?

webpack/components/Table/TableWrapper.js
```diff
@@ -164,6 +189,15 @@ TableWrapper.propTypes = {
...
   activeFilters: PropTypes.bool,
+  displaySelectAllCheckbox: PropTypes.bool,
+  selectedCount: PropTypes.number,
+  modelName: PropTypes.string,
+  modelNamePlural: PropTypes.string,
+  selectAll: PropTypes.func,
+  selectNone: PropTypes.func,
+  selectPage: PropTypes.func,
+  areAllRowsSelected: PropTypes.func,
+  areAllRowsOnPageSelected: PropTypes.func,
```
Andrewgdewar, line 200: If all of these are only used by the 'Select all checkbox" then we should likely have a selectAllCheckBoxProps instead to clean this up a bit, the list of props on this file is getting messy.

webpack/components/Table/TableWrapper.js
```diff
@@ -96,14 +111,24 @@ const TableWrapper = ({
...
 
-  const rowsCount = metadata?.subtotal ?? 0;
-  const unresolvedStatus = !!allTableProps?.status && allTableProps.status !== STATUS.RESOLVED;
-  const unresolvedStatusOrNoRows = unresolvedStatus || rowsCount === 0;
-  const searchNotUnderway = !(searchQuery || activeFilters);
-
   return (
     <>
       <Flex>
+        {displaySelectAllCheckbox &&
```
Andrewgdewar, line 117: Can this be determined by the presents of the prop I mention below?

webpack/components/SelectAllCheckbox/index.js
```diff
@@ -0,0 +1,117 @@
...
+    if (selectedCount > 0) newCheckedState = null; // null is partially-checked state
+    if (areAllRowsSelected) newCheckedState = true;
+    setSelectAllChecked(newCheckedState);
+  }, [selectedCount, areAllRowsSelected]);
+
+  const selectAllDropdownItems = [
+    <DropdownItem key="select-all" component="button" isDisabled onClick={handleSelectAll}>
+      {__('Select all')}
+    </DropdownItem>,
+    <DropdownItem key="select-page" component="button" isDisabled={areAllRowsOnPageSelected} onClick={handleSelectPage}>
```
Andrewgdewar, line 62: There was talk of removing items instead of disabling them when not relevant, not sure if Maria had weighed in on that for the dropdown items?

webpack/components/SelectAllCheckbox/index.js
```diff
@@ -0,0 +1,117 @@
...
+  modelName,
+  modelNamePlural,
+  areAllRowsOnPageSelected,
+  areAllRowsSelected,
+}) => {
+  const [isSelectAllChecked, setSelectAllChecked] = useState(null);
+  const [isSelectAllDropdownOpen, setSelectAllDropdownOpen] = useState(false);
+
+  const onSelectAllCheckboxChange = () => {
+    if (isSelectAllChecked === false) {
```
Andrewgdewar, line 25: for false checks can we just bang it '! isSelectAllChecked"?

Andrewgdewar: Or in this case, you could do:
```suggestion
    if (isSelectAllChecked ) {
      return selectNone();
    }
     return selectPage();
```

jeremylenz: For this one I was explicit because it can have three values (`false`, `null`, `true`)

webpack/components/SelectAllCheckbox/index.js
```diff
@@ -0,0 +1,117 @@
...
+    modelName,
+    modelNamePlural,
+  );
+
+  const handleSelectAll = () => {
+    selectAll();
+    setSelectAllDropdownOpen(false);
+  };
+  const handleSelectPage = () => {
+    selectPage();
```
Andrewgdewar, line 42: same as above if applicable

webpack/components/SelectAllCheckbox/index.js
```diff
@@ -0,0 +1,117 @@
...
+    setSelectAllDropdownOpen(false);
+  };
+  const handleSelectNone = () => {
+    selectNone();
+    setSelectAllDropdownOpen(false);
```
Andrewgdewar, line 47: same as above if applicable

webpack/components/extensions/HostDetails/Tabs/HostTracesActions.js
Andrewgdewar: K, yeah I'd like to see the dispatch call made into it's own function.

webpack/components/extensions/HostDetails/Tabs/TracesTab.js
```diff
@@ -2,33 +2,75 @@ import React, { useState, useCallback } from 'react';
...
-  const actionButtons = <Button variant="danger" isDisabled> {__('Restart app')} </Button>;
   const fetchItems = useCallback(
     params =>
       (hostId ? getHostTraces(hostId, params) : null),
     [hostId],
   );
   const response = useSelector(state => selectAPIResponse(state, 'HOST_TRACES'));
   const { results, ...meta } = response;
+  const onRestartApp = () => {
+    dispatch(resolveHostTraces(hostId, { trace_ids: [...selectedTraces] }, dispatch));
```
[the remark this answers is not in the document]
Andrewgdewar: I thought we were doing the onSuccess in the component? Not passing the dispatch up to the API.

webpack/components/SelectAllCheckbox/index.js
Andrewgdewar: if it is null you want it to hit select page?

webpack/components/SelectAllCheckbox/index.js
MariaAga: I think you're looking for @MariSvirik :)

webpack/components/extensions/HostDetails/Tabs/TracesTab.js
jeremylenz: No updates to the backend were needed.  We just keep track on the front end of which trace IDs are selected, and render accordingly.

jeremylenz: Backend updates will be needed to get 'Select all' functionality working, however. That is left for a future PR.

jeremylenz: > I can't understand in the code how the select all is working

To be clear, the 'Select all' checkbox is correctly rendering as checked, but the 'Select all' menu item doesn't work yet.

The checkbox rendering works just by comparing the number of selected items and the total from pagination metadata.

webpack/components/SelectAllCheckbox/index.js
jeremylenz: The checkbox states are:
Unchecked: false
Partially checked: null
Checked: true

The flow is:

All are selected (true) -> click -> selectNone()
Some are selected, or the page is selected (null) -> click -> selectNone()
None are selected (false) -> click -> selectPage()

jturel (commented):
I tested the change and it looks great and works well. Compared the JobInvocations that get created and they are created identically so I'm confident this would work end-to-end. Left some really minor comments on the code :)

webpack/components/extensions/HostDetails/Tabs/TracesTab.js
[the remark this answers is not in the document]
jeremylenz: @Andrewgdewar thanks, updated! turns out this was left over anyway, and not even doing anything :smile:

webpack/components/extensions/HostDetails/Tabs/HostTracesActions.js
jeremylenz: Turns out this same thing can also be accomplished in a single line
```js
handleSuccess: response => renderTaskStartedToast(response.data),
```
(or close enough to the same thing, at least.) So I'm doing that instead.

package.json
```diff
@@ -48,6 +48,7 @@
   "dependencies": {
     "angular": "1.8.2",
     "bootstrap-select": "1.13.6",
+    "dequal": "^2.0.2",
```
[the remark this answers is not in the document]
jeremylenz: It's a dependency of `use-deep-compare-effect` so I'd guess yes?  Packaging is an eternal mystery to me though. @tbrisker @evgeni @ehelms  ?

webpack/components/SelectAllCheckbox/index.js
```diff
@@ -0,0 +1,99 @@
...
+    selectPage();
+  };
+  const handleSelectNone = () => {
+    setSelectAllDropdownOpen(false);
+    selectNone();
+  };
+
+  useEffect(() => {
+    let newCheckedState;
+    if (selectedCount === 0) newCheckedState = false;
```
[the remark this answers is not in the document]
jeremylenz: @jturel gotcha, updated :)

webpack/components/extensions/HostDetails/Tabs/TracesTab.js
```diff
@@ -43,20 +84,28 @@ const TracesTab = () => {
...
+        selectNone={selectNone}
+        selectAll={selectAll}
         rowsCount={results?.length}
+        areAllRowsOnPageSelected={areAllRowsOnPageSelected}
+        areAllRowsSelected={areAllRowsSelected}
         variant={TableVariant.compact}
         status={status}
         metadata={meta}
+        selectedCount={selectedTraces.size}
       >
```
sjha4, line 98: These new props when defined in other pages should start supporting the multi-page select on those pages correct? Can we also have an option to hide select all across pages dropdown item..Not all APIs will immediately support that and the disabled option might be disabled for some time on those pages..

webpack/utils/helpers.js
```diff
@@ -100,6 +100,11 @@ export const capitalize = s => s && s[0].toUpperCase() + s.slice(1);
 
 export const truncate = (str, truncateLimit = 50) => (str.length > truncateLimit ? `${str.substring(0, truncateLimit - 3)}...` : str);
 
+export const pluralize = (quantity, singular, plural = `${singular}s`) => {
+  if (quantity === 1) return `${quantity} ${singular}`;
+  return `${quantity} ${plural}`;
+};
+
```
sjha4, line 107: 🥳 🥳

webpack/components/extensions/HostDetails/Tabs/TracesTab.js
jeremylenz: @sjha4 Yeah, this is covered by the`displaySelectAllCheckbox` prop which is false by default :)

webpack/components/SelectAllCheckbox/index.js
```diff
@@ -0,0 +1,107 @@
...
+
+  const selectAllDropdownItems = [
+    <DropdownItem key="select-none" component="button" onClick={handleSelectNone}>
+      {`${__('Select none')} (0)`}
+    </DropdownItem>,
+    <DropdownItem key="select-page" component="button" isDisabled={areAllRowsOnPageSelected} onClick={handleSelectPage}>
+      {`${__('Select page')} (${pageRowCount})`}
+    </DropdownItem>,
+    <DropdownItem key="select-all" component="button" isDisabled onClick={handleSelectAll}>
+      {`${__('Select all')} (${totalCount})`}
```
sjha4, line 67: @MariSvirik , @jeremylenz : Hiding this instead of disabling it for pages that don't support multi page select would allow us to use this on other pages without having to wait for API support..Not sure how soon that would be..

sjha4 (commented):
Works great. Tested with dummy traces on the UI only cause I don't know enough about the entire flow..Code LGTM..Good to go from my perspective..🎉 🎉 
I'll let @Andrewgdewar turn the button green..

webpack/components/Table/TableWrapper.js
```diff
@@ -24,38 +27,64 @@ const TableWrapper = ({
...
-    const { subtotal: newTotal, page: newPage, per_page: newPerPage } = data;
-    if (newTotal !== undefined) setTotal(Number(newTotal));
-    if (newPage !== undefined) setPage(Number(newPage));
-    if (newPerPage !== undefined) setPerPage(Number(newPerPage));
-  };
-  const paginationParams = useCallback(() => ({ per_page: perPage, page }), [perPage, page]);
+  const perPage = Number(metadata?.per_page ?? foremanPerPage);
+  const page = Number(metadata?.page ?? 1);
+  const total = Number(metadata?.subtotal ?? 0);
+  const { pageRowCount } = getPageStats({ total, page, perPage });
```
Andrewgdewar, line 44: happy you pulled this out.

webpack/components/SelectAllCheckbox/index.js
jeremylenz: @sjha4 Updated.

package.json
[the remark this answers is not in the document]
jeremylenz: Update: turns out Lodash can do the same thing via `isEqual`, and we already depend on Lodash, so I updated to use that instead. Crisis averted :smile:

jturel (approved):
Code + functionality ACK! 

Nice work @jeremylenz !

Andrewgdewar (approved):
Looks good!

package.json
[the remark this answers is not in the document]
evgeni: It would not have existed (we vendor most of the deps in packaging), but glad you thought of if and sorted it out ;)

jeremylenz:
@MariSvirik I was finally able to spin up Firefox with this, and it confirmed something I suspected:
blue-partial-select

The blue partially-selected checkbox comes from how the browser styles it.  In Firefox it looks like the above.  The non-blue one is just the way Chrome does it.

PR #9595, opened 2021-08-31
Fixes #33383 - fix error in logs during repo enable

jlsherrill:


jlsherrill:
Steps to reproduce:

1.  import a manifest
2.  go to the RH repos page, scan repos for the 'rhel-7-server-extras-rpms' content set
3.  Notice you see no errors, and see the repository as expected.  You can enable it with no issue
4.  Check /var/log/foreman/production.log for errors, notice the errors mentioned here.

chris1984 (approved):
ACK, works good

jlsherrill:
hah i think @jturel fixed this slightly differently as part of https://github.com/Katello/katello/commit/d40c039b808da04029f21788575f28a2cce446e2#diff-430b5eb5e02b769285564038646f712691c634b5f58d4e5700a03f164d7d5b34R63

jturel:
Funny we both came across that! You fix is better than mine, so I think it's worth rebasing and removing the reject nils I added. TIL about the second argument to in_groups_of! https://apidock.com/rails/Array/in_groups_of

jlsherrill:
sounds good!

PR #9596, opened 2021-08-31
Fixes #32660 - adds an authentication token for SLES repositories

jjeffers:
This PR adds an explicit auth token field to support remote URLs that happen to have query parameters, but are interpreted as SLES auth tokens. Currently any query params are considered to be auth tokens, even when this is not the case.

### DIFF
--- a/app/controllers/katello/api/v2/repositories_controller.rb
+++ b/app/controllers/katello/api/v2/repositories_controller.rb
@@ -56,6 +56,7 @@ module Katello
       param :verify_ssl_on_sync, :bool, :desc => N_("if true, Katello will verify the upstream url's SSL certifcates are signed by a trusted CA")
       param :upstream_username, String, :desc => N_("Username of the upstream repository user used for authentication")
       param :upstream_password, String, :desc => N_("Password of the upstream repository user used for authentication")
+      param :upstream_authentication_token, String, :desc => N_("Password of the upstream authentication token.")
       param :ostree_upstream_sync_policy, ::Katello::RootRepository::OSTREE_UPSTREAM_SYNC_POLICIES, :desc => N_("policies for syncing upstream ostree repositories")
       param :ostree_upstream_sync_depth, :number, :desc => N_("if a custom sync policy is chosen for ostree repositories then a 'depth' value must be provided")
       param :deb_releases, String, :desc => N_("whitespace-separated list of releases to be synced from deb-archive")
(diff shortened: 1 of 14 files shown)


jjeffers:
[test katello]

jlsherrill:
+1 to the overall strategy, i wonder if we should encrypt it like we do the upstream_password in root_repository?

jjeffers:
> +1 to the overall strategy, i wonder if we should encrypt it like we do the upstream_password in root_repository?

I was wondering the same thing. I don't know how sensitive the auth token is, but based on what I've read, the SLES provisioning seems to be similar to a protected value, like an API key.

jjeffers:
> > +1 to the overall strategy, i wonder if we should encrypt it like we do the upstream_password in root_repository?


I am also not sure about the tip text on repository creation. Is there a better way to explain the new field?

jjeffers:
@m-bucher I noticed back last year you helped @jlsherrill with testing SLES authentication tokens in Katello. This PR modifies how those tokens are specified (we make it more explicit on repo creation, etc). I don't have a SLES account to test this with, so I am wondering if you can help with that?

ianballou (commented):
I tried syncing http://mirrorlist.centos.org/?release=7&arch=x86_64&repo=os&infra=stock and Pulp errored out with "An invalid remote URL was provided: http://mirrorlist.centos.org/".

~Looking on the remote, it was shortened to http://mirrorlist.centos.org/.  I'm guessing we still need those query params there?  Without this PR, the repo syncs fine.~

Edit:  This doesn't make sense to me, but I cannot sync the repo unless there is "sles_auth_token": "release=7&arch=x86_64&repo=os&infra=stock on the remote.

ianballou:
Also @jjeffers do you have a starting place for testing out the new "Upstream Authentication Token" field?

ianballou (commented):
It appears that the download policy field has disappeared from the new repo creation page.

m-bucher:
> I don't have a SLES account to test this with, so I am wondering if you can help with that?

@jjeffers, sorry, I missed this :sweat_smile: 

I just tested this and it worked with an SCC-Token.

We will have to make changes for this in https://github.com/ATIX-AG/foreman_scc_manager as well, so please open an issue there, too.

jjeffers:
> Also @jjeffers do you have a starting place for testing out the new "Upstream Authentication Token" field?

I can send you an auth token, but keep it secret, keep it safe!

jjeffers:
> It appears that the download policy field has disappeared from the new repo creation page.

I noticed I omitted a `div`, should be corrected now.

ianballou:
I noticed that you get an error if you include the question mark at the beginning of the upstream auth token. Should we be removing those for people?

jjeffers:
> I noticed that you get an error if you include the question mark at the beginning of the upstream auth token. Should we be removing those for people? I'm not sure if it should be obvious or not that the `?` shouldn't be there, but I can see people making that mistake.

AFAIK you should not include the query parameter delimeter (`?`). Perhaps we should change the help text or provide an example?

ianballou:
> AFAIK you should not include the query parameter delimeter (`?`). Perhaps we should change the help text or provide an example?

We're probably fine then.

ianballou (approved):
This is working fine for me! Code looks fine too. Just need a fix for the master test issue.

jjeffers:
[test katello]

jlsherrill:
[test katello]

PR #9598, opened 2021-08-31
Fixes #33386 - Add repository type counts to CV version response

sjha4:
To test:
Add some repositories of different content types and check the /content_view_versions/:id endpoint for new fields that look like this:
yum_repository_count , ansible_collection_repository_count

### DIFF
--- a/app/views/katello/api/v2/content_view_versions/base.json.rabl
+++ b/app/views/katello/api/v2/content_view_versions/base.json.rabl
@@ -13,6 +13,9 @@ node do |version|
   version.content_counts_map
 end
 
+node do |version|
+  version.repository_type_counts_map
+end
 node :errata_counts do |version|
   partial('katello/api/v2/errata/counts', :object => Katello::RelationPresenter.new(version.errata))
 end
--- a/app/models/katello/content_view_version.rb
+++ b/app/models/katello/content_view_version.rb
@@ -1,4 +1,5 @@
 module Katello
+  # rubocop:disable Metrics/ClassLength
   class ContentViewVersion < Katello::Model
     include Authorization::ContentViewVersion
     include ForemanTasks::Concerns::ActionSubject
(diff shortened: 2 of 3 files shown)


Andrewgdewar (approved):
Ack for functionality!

sjha4:
[test katello]

app/views/katello/api/v2/content_view_versions/base.json.rabl
```diff
@@ -13,6 +13,9 @@ node do |version|
   version.content_counts_map
 end
 
+node do |version|
+  version.repository_type_counts_map
+end
```
jlsherrill, line 18: would it make sense to have a structure more like this:
```
"repository_counts": {
     yum: 5,
     docker: 3
     ansible_collection: 1
    etc...
  }
```  
?

sjha4: So @adewar is writing a component which accepts some fields to pluck out of the response in React. Having nested results like the above adds complexity to what he is trying to achieve. That's why the counts are at the top level.

sjha4: Oh..wrong Adewar.. @Andrewgdewar 😮

Andrewgdewar: @sjha4  Also asked if this would be okay. Although possible to do, the other counts are available from the main response. 
I am using a composed component, that will consume the API, I don't want to build a custom mapper to get the data if I can avoid it.
Here's the value represented in the config file for my other PR: https://github.com/Katello/katello/pull/9608/files#diff-c223792afe94f0173ba83b0cad193dffe5adf5a701cfbc8c36a0e6a343c03c26R24

Andrewgdewar: TLDR: I could do it, it would add a small amount of complexity that I'd prefer to avoid. 
I'm a 4/10 on preference so if you're more passionate about it swing away! :smile:

jlsherrill: nah, its a bit cleaner, but not a big deal, and the errata counts are already like this

jlsherrill (approved):

PR #9599, opened 2021-08-31
Fixes #33385,33054,33144 - Fixes host group inheritance

parthaa:
Before the PR -

On the  create  hostgroups page the inheritance hierarchy behavior was different
between katello attribute like content_view/lce/content source etc vs
things like os and arch
For example if you create a child hostgroup and pointed to a parent
with content view, you would see the cv automatically selected as the
default, instead of 'Inherit Content View' as os/arch fields behave.

More over if you had cv, lce, cs, arch filled in the parent hostgroup
and enabled operating system in the child the media section would not
get automatically populated (because the notion of parent_id was
 missing)

After:

Both the above cases behave appropriately.

### DIFF
--- a/app/assets/javascripts/katello/hosts/host_and_hostgroup_edit.js
+++ b/app/assets/javascripts/katello/hosts/host_and_hostgroup_edit.js
@@ -4,7 +4,7 @@ KT.hosts = {};
 $(document).on('ContentLoad', function(){
     KT.hosts.onKatelloHostEditLoad();
     window.tfm.hosts.registerPluginAttributes("os",
-         ['lifecycle_environment_id', 'content_view_id', 'environment_id', 'content_source_id', 'architecture_id']);
+         ['lifecycle_environment_id', 'content_view_id', 'environment_id', 'content_source_id', 'architecture_id', 'parent_id']);
 
     $("#hostgroup_lifecycle_environment_id").change(KT.hosts.environmentChanged);
     $("#host_lifecycle_environment_id").change(KT.hosts.environmentChanged);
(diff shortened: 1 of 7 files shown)


parthaa:
To test this:

- Create and sync a ks repo -> ( eg: https://partha.fedorapeople.org/test-repos/kickstart-zoo/ )
- Create a parent hg with CV, Library, Content Source and Architecture (`x86_64`) set (leave the OS blank)
- Create another host group and have it inherit from parent hostgroup
- Go to `operating system` tab and select `RedHat 7`

Before PR:
- You'd see that the content view, lce and cs parameters are populated with the names of those entities intead of `Inherit <entity_name>`
Create-Host-Group -pre
- You'd see that selecting the OS as RH doesn't populate media.
Create-Host-Group -os-pre

After PR:
- You'd see `Inherit <entity_name>`
Create-Host-Group
- Selecting the OS automatically populates the synced area
Create-Host-Group_OS

Play with it more saving hg's etc.

jlsherrill:
does this also fix this issue? https://projects.theforeman.org/issues/33144

parthaa:
> does this also fix this issue? https://projects.theforeman.org/issues/33144

Yes!

sbernhard:
BTW, works for us! Thanks @parthaa

sbernhard (approved):

jlsherrill:
image

jlsherrill:
seeing an issue where:

1. Create LCE prod
2. create cv and publish into just library
3. create a host group assigned to library/cv
4. create a child hostgroup of parent and select prod  LCE, leave env as inherit. 
5. save => see no error even though its invalid

jlsherrill:
ignore the previous issue, its really not a big deal.     but one issue i'm seeing is that after selecting the lifecycle env in the hostgroup form, the inherit option goes away

parthaa:
>  but one issue i'm seeing is that after selecting the lifecycle env in the hostgroup form, the inherit option goes away

Fixed!

jlsherrill (approved):

PR #9600, opened 2021-08-31
Fixes #33382 - ensure inc update occurs before REX applicable errata install

ianballou:
The incremental update Bastion Katello code does not actually call incremental update before starting a Remote Execution job.  I considered calling the existing `incrementalUpdate()` before submitting `#errataActionForm`, but I was worried that having to wait for the incremental update before spawning the REX job might get complicated.  So, I refactored the `IncrementalUpdates` action to decide if REX or Katello Agent should be used to apply the hosts' errata.

To test:
1) Set Katello to use Remote Execution by default in the Content Settings
2) Set up some content hosts to have the same applicable (but not installable) errata
3) Try applying those errata by incrementally updating them through the UI (Content -> Errata)
4) See that the REX job appears but the content view version is not incrementally updated
5) Apply my patch and re-do the errata application. Now, the content view version will be incrementally updated before running the REX job
6) Check that Katello Agent still works as well

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -131,7 +131,7 @@ Metrics/ModuleLength:
 # Offense count: 3
 # Configuration parameters: CountKeywordArgs.
 Metrics/ParameterLists:
-  Max: 6
+  Max: 7
 
 # Offense count: 122
 Metrics/PerceivedComplexity:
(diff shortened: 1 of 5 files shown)


jjeffers (approved):
LGTM, worked well to fix the original problem.

The task monitor shows the incremental update created. A subsequent task shows that the REX job was invoked to install that errata on the content host.

PR #9601, opened 2021-09-01
Fixes #33393 - remove Katello's customConfirm for .erb

laviro:
Foreman also overrides the jQuery UJS confirm with a PF4 confirm modal in https://github.com/theforeman/foreman/pull/8619


tbrisker (approved):
Thanks @laviro ! initially I thought we'll need to convert `customConfirm` to use the new modal, but looks like the only place that actually uses it is the default rails handlers.

Andrewgdewar (approved):
<s>Much delete, Ack</s>
Test failures with your other PR https://github.com/theforeman/foreman/pull/8619
Fix the failing tests in Katello before merge.

sjha4:
The VCR error seems unrelated..
[test katello]

ezr-ondrej:
I think this can go in first actually. We should remove related tests and merge

laviro:
> Much delete, Ack
> Test failures around the mocked rails methods when combined with your other PR [theforeman/foreman#8619](https://github.com/theforeman/foreman/pull/8619)
> Fix the failing tests in Katello before merge.

Can you try again with the updated core branch? thanks

tbrisker:
Looks like the updated version fixes the issue with katello, and in any case this can go in as @ezr-ondrej mentioned - worst case katello will have default confirmation dialogues until core is merged (which will be soon).

PR #9609, opened 2021-09-02
Fixes #33407 - Package dependency UI values not formatted correctly

rverdile:
The values for the dependencies were incorrect.

For example, what should be "package_name = 3.1.3-14.el7_7", was being displayed as "package_name = 0.3.1.3.14.el7_7".

Also, the less than sign was flipped for any dependencies with a less than requirement.

To test: 
- go to Content > Packages
- select a package
- observe the values under "Requires" and "Provides" in "Dependencies"

You can check against the values against, for example, `rpm -q --requires tfm-rubygem-hammer_cli`, where `tfm-rubygem-hammer_cli` is the package name.

foreman_concrete from here: http://yum.theforeman.org/plugins/nightly/el7/x86_64/ is a good package to check.

### DIFF
--- a/app/services/katello/pulp3/rpm.rb
+++ b/app/services/katello/pulp3/rpm.rb
@@ -26,41 +26,11 @@ module Katello
       end
 
       def requires
-        results = []
-        flags = {'GT' => '>', 'LT' => '>', 'EQ' => '=', 'GE' => '>=', 'LE' => '<='}
-
-        backend_data['requires']&.each do |requirement|
-          requires_str = ""
-          if requirement.count < 3
-            requires_str = requirement.first
-            results << requires_str
-          else
-            requirement[1] = flags[requirement[1]]
-            requirement[0...2].each { |requirement_piece| requires_str += "#{requirement_piece} " }
-            requirement[2...-1].each { |requirement_piece| requires_str += "#{requirement_piece}." }
-            results << requires_str[0...-1]
-          end
-        end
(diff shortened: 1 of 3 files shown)


app/services/katello/pulp3/rpm.rb
```diff
@@ -26,41 +26,13 @@ def self.page_options(page_opts = {})
...
-          else
-            provided[1] = flags[provided[1]]
-            provided[0...2].each { |provided_piece| provides_str += "#{provided_piece} " }
-            provided[2...-1].each { |provided_piece| provides_str += "#{provided_piece}." }
-            results << provides_str[0...-1]
-          end
-        end
-        results.uniq
+        results = Util::Package.parse_dependencies(backend_data['provides'])
+        results
```
jlsherrill, line 35: no need to grab these into a variable and return them, you can just call `Util::Package.parse_dependencies(backend_data['provides'])`

same for the requires method too

jlsherrill:
works great and i like the refactor!  one comment and its good!

app/services/katello/pulp3/rpm.rb
rverdile: updated!

jlsherrill (approved):
thanks @rverdile will merge once it goes green

rverdile:
[test katello]

jlsherrill:
[test katello]

PR #9611, opened 2021-09-02
Fixes #31590 - add repo timestamps to api response

jlsherrill:


app/views/katello/api/v2/repositories/show.json.rabl
```diff
@@ -28,6 +28,7 @@ glue(@resource.root) do
   attributes :http_proxy_id
   attributes :http_proxy_name
   attributes :retain_package_versions_count
+  attributes :last_contents_changed, :last_applicability_regen
```
evgeni, line 31: is that *additional* to what's in line 61 (as this is glued to the root repo, while the other is not, or is the other wrong?

jlsherrill: hah  me-- its already there!

PR #9613, opened 2021-09-03
Fixes #32556 - disable scopes on cv promote errata calc

jlsherrill:


jlsherrill:
to reproduce:

1) create a new lifecycle environment
2) create a new location
3) create a cv, publish, promote it to that LCE
4) register a host to that cv/lce
5) make sure your location is switched to the new location
4) publish a new CV version and promote it to the LCE, notice the promote task is left in 'warning' state.  Should have errors in the logs

rverdile (commented):
Definitely works for the expected case! Just gonna dig through the code a little and see if I find anything interesting.

app/lib/actions/katello/content_view/promote_to_environment.rb
```diff
@@ -46,10 +46,14 @@ def rescue_strategy_for_self
 
         def finalize
           # update errata applicability counts for all hosts in the CV & LE
-          ::Katello::Host::ContentFacet.where(:content_view_id => input[:content_view_id],
-                                              :lifecycle_environment_id => input[:environment_id]).each do |facet|
-            facet.update_applicability_counts
-            facet.update_errata_status
+          Location.no_taxonomy_scope do
+            User.as_anonymous_admin do
```
rverdile, line 50: It seems like this fix works even without this `User.as_anonymous_admin` block. What does this block do?

rverdile (commented):
Messing around with the code, it seems like just adding the `Location.no_taxonomy_scope` block, without adding the `User` block or adding unscoped to `::Katello::Host::ContentFacet.where...`, avoids the error. 

Just curious, what did you find to be the cause of the bug and where do these two additions come into play?

app/lib/actions/katello/content_view/promote_to_environment.rb
jlsherrill: Great question!  To fix this bug, its not needed.  However a similar bug would happen if some user has permissions to publish/promote a content view but does not have permissions to view some set of hosts.  

The location/organization and current user are preserved in the task, so the task would get execute that code as that uesr and hit the same issue as being in the wrong location

jlsherrill:
[test katello]

rverdile (commented):
I was trying to find a way to make a test for this, but it seems like testing doesn't really hit on users and/or locations the way that's needed to recreate this bug. Is that a fair analysis?

Other than that, this looks good to me! I think we just need someone else to make the final approval.

jlsherrill:
> I was trying to find a way to make a test for this, but it seems like testing doesn't really hit on users and/or locations the way that's needed to recreate this bug. Is that a fair analysis?

Actually there is a way to test it, but testing a method in a finalize method is difficult which gave me the idea to move it!

So i moved it to the content_view model and wrote a test!

app/lib/actions/katello/content_view/promote_to_environment.rb
```diff
@@ -45,17 +45,12 @@ def rescue_strategy_for_self
...
 
         def finalize
-          # update errata applicability counts for all hosts in the CV & LE
-          ::Katello::Host::ContentFacet.where(:content_view_id => input[:content_view_id],
-                                              :lifecycle_environment_id => input[:environment_id]).each do |facet|
-            facet.update_applicability_counts
-            facet.update_errata_status
-          end
+          environment = ::Katello::KTEnvironment.find(input[:environment_id])
+          Katello::ContentView.find(input[:content_view_id]).update_host_statuses(environment)
```
rverdile, line 49: Should be `::Katello` here

test/models/content_view_test.rb
```diff
@@ -607,5 +607,17 @@ def test_import_only_immutable
...
       refute cv.valid?
     end
+
+    def test_update_host_statuses
+      other_location = taxonomies(:location2)
+      facet = katello_content_facets(:content_facet_two)
+
+      refute_equal facet.host.location, other_location
+      assert_nothing_raised do
+        Location.as_taxonomy(facet.host.organization, other_location) do
```
rverdile, line 617: Just curious, what does `as_taxonomy` do?

rverdile (commented):
Test works perfectly, if I undo the fix I get the same error!

Everything looks good except for the one error.

test/models/content_view_test.rb
jlsherrill: The foreman application has 3 'states' that determine the scope of some query:
```
User.current
Location.current
Organization.current
```

(Note that these are Thread local variables, not class variables)  These are usually determined by what user is logged in, or the current location or organization that is set.  In this case, Location.as_taxonomy() sets the current Location and Organization temporarily within the block of code that is passed.  So even if Location.current is set to Location1 (or nil), it will override that and temporarily set Location.current to what is passed in.

app/lib/actions/katello/content_view/promote_to_environment.rb
jlsherrill: fixed!

rverdile:
[test katello]

jjeffers (approved):
ACK

PR #9615, opened 2021-09-03
Fixes #32460,#33375 speed up yum filter copy and stop debs from failing filtered CV publishes

ianballou:
This PR speeds up content  view version publishing by only using the multi-copy tasks if yum depsolving is enabled.  If it isn't, we can safely use the "old" method of copying content.  This allows for better concurrency, and yum repositories that are not affected by filters will simply have their repository versions copied.

I also moved to using the RPM modify API instead of the copy API because there is a bug where the copy API does not allow the Pulp workers to work concurrently.  This should also help filtered CV publishing to be a bit faster without depsolving.
Note: I need to test this to check that concurrency is going on.

This PR also fixes an issue where deb repositories in filtered CVVs caused failures due to no `copy_content_for_source` support.

TODO:

- [x] Check that two content view publishes can run at the same time using the modify API.

To test:

Bug [32460](https://projects.theforeman.org/issues/32460)

1) Check that the "Multi" copy tasks aren't used for content view publishes with dependency solving disabled.
2) Check that the content view filtering is still accurate.
3) Check that incremental update is still accurate.  Incremental update still relies on the Multi copy tasks.
4) Check that repositories without filters in content view versions are only having repository versions copied over, not actual content.
  -> You can tell by checking that `Actions::Pulp3::Repository::CopyContent` was not run for that repository during the publish.

Bug [33375](https://projects.theforeman.org/issues/33375)

1) Add a deb repo to a content view that has any filter assigned to it. Ensure the content view publish does not fail.

### DIFF
--- a/app/lib/katello/util/errata.rb
+++ b/app/lib/katello/util/errata.rb
@@ -46,13 +46,18 @@ module Katello
         end
       end
 
-      def filter_errata_by_pulp_href(errata, package_pulp_hrefs)
+      def filter_errata_by_pulp_href(errata, package_pulp_hrefs, source_repo_rpm_filenames)
         return [] if package_pulp_hrefs.empty?
         rpms = Katello::Rpm.where(:pulp_id => package_pulp_hrefs)
         rpm_filenames = rpms.map { |rpm| File.basename(rpm.filename) }
+        source_repo_rpm_filenames = source_repo_rpm_filenames.map { |rpm| File.basename(rpm) }
         matching_errata = []
         errata.each do |erratum|
-          if erratum.packages.any? && (erratum.packages.pluck(:filename) - rpm_filenames).empty?
+          # The erratum should be copied if package_pulp_hrefs has all of its packages that are available in the source repo.
+          next if erratum.packages.empty?
+          rpms_in_erratum_and_source_repo = erratum.packages.pluck(:filename) & source_repo_rpm_filenames
+          next if rpms_in_erratum_and_source_repo.empty?
+          if (rpms_in_erratum_and_source_repo - rpm_filenames).empty?
             matching_errata << erratum
           end
         end
--- a/app/services/katello/pulp3/repository/apt.rb
+++ b/app/services/katello/pulp3/repository/apt.rb
@@ -47,9 +47,8 @@ module Katello
           "/pulp/deb/#{repo.relative_path}/".sub('//', '/')
         end
 
-        def copy_content_for_source
-          # TODO
-          fail NotImplementedError
+        def copy_content_for_source(source_repository, _options = {})
+          copy_units_by_href(source_repository.debs.pluck(:pulp_id))
         end
 
         def regenerate_applicability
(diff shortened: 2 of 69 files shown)


app/services/katello/pulp3/repository/apt.rb
```diff
@@ -47,9 +47,8 @@ def partial_repo_path
           "/pulp/deb/#{repo.relative_path}/".sub('//', '/')
         end
 
-        def copy_content_for_source
-          # TODO
-          fail NotImplementedError
+        def copy_content_for_source(source_repository, _options = {})
+          copy_units_by_href(source_repository.files.pluck(:pulp_id))
```
jlsherrill, line 51: shouldn't thinks be '.debs'

ianballou: Yep it should! Not sure why it looks like the debs got copied anyway...

ianballou:
Sometimes there are timeouts when using the modify API with many units. Chunking helps, but increasing the API workers to 4 seemed to do the trick. There is also an issue in Pulp to improve the efficiency of the modify copy: https://pulp.plan.io/issues/9266

ianballou:
Question I just thought of.  Will the non-multi repo copy actions properly include errata that have packages spanning multiple repositories? I'm thinking no, so that might need a solution.

ianballou:
The fix to this issue (https://projects.theforeman.org/issues/32435) and this PR are going to have conflicting code.

app/lib/katello/util/errata.rb
```diff
@@ -46,13 +46,18 @@ def filter_by_state(errata_list, errata_state)
...
         rpms = Katello::Rpm.where(:pulp_id => package_pulp_hrefs)
         rpm_filenames = rpms.map { |rpm| File.basename(rpm.filename) }
+        source_repo_rpm_filenames = source_repo_rpm_filenames.map { |rpm| File.basename(rpm) }
         matching_errata = []
         errata.each do |erratum|
-          if erratum.packages.any? && (erratum.packages.pluck(:filename) - rpm_filenames).empty?
+          # The erratum should be copied if package_pulp_hrefs has all of its packages that are available in the source repo.
+          next if erratum.packages.empty?
+          rpms_in_erratum_and_source_repo = erratum.packages.pluck(:filename) & source_repo_rpm_filenames
+          next if rpms_in_erratum_and_source_repo.empty?
```
jlsherrill, line 59: do these to 'next' statement ignore empty errata?  I think we have to support those

jlsherrill: ignore this comment!

ianballou:
For publishing CentOS 7 Kickstart + CentOS 8 BaseOS together with a 'kernel' exclude filter, I'm finding this PR cuts the time by 40-50%.  6 minutes vs 10-11 minutes.

jlsherrill (approved):
ACK!

jlsherrill:
don't forget to merge!

PR #9616, opened 2021-09-07
Fixes #33425 - Optionally prefer registered through proxy for REX

adamruzicka:


jlsherrill:
[test katello]

jturel:
ping @adamruzicka - rebase needed :)

jturel:
@adamruzicka I need some help understanding / evaluating this change. Maybe the content_source would be a better choice than registered through? Was the "Registered Through" smart proxy chosen because that's something that the user or support saw in the UI?

Regarding the new setting - it seems necessary but if possible I think we should avoid it (I'm thinking of reducing complexity, code paths, making debugging easier). Can you think of a way we can do without it? I guess making the strategy you added be the default would not be good?

adamruzicka:
> Was the "Registered Through" smart proxy chosen because that's something that the user or support saw in the UI?

It felt a lot like people didn't understand how rex picks a proxy and were hitting our support with "the host is registered to that proxy, why isn't it using that proxy for rex?"

> Can you think of a way we can do without it? I guess making the strategy you added be the default would not be good?

Not really, it would be a bad default. From my point of view this is implemented purely for people who can't be bothered to set things up properly.

app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -28,6 +28,16 @@ def update_os_from_facts
             content_facet.kickstart_repository_id = valid_repos.first
           end
         end
+
+        def remote_execution_proxies(provider, *_rest)
```
jturel, line 32: This method doesn't seem to be getting called (my breakpoint isn't hit when kicking off a REX job)

lib/katello/engine.rb
```diff
@@ -226,6 +226,10 @@ class Engine < ::Rails::Engine
       ::AuditSearch::ClassMethods.prepend Katello::Concerns::AuditSearch
       ::Widget.singleton_class.prepend Katello::Concerns::WidgetExtensions::ClassMethods
 
+      if with_remote_execution?
```
jturel, line 229: This is throwing an error - it should be Katello.with_remote_execution

jturel:
I understand the reasoning now but it doesn't seem to work. Even with the setting set to true the default smart proxy is getting selected rather than the registered_through proxy which I've configured for my content host

lib/katello/engine.rb
adamruzicka: Right, not sure how I managed to let this slip

app/models/katello/concerns/host_managed_extensions.rb
adamruzicka: Not sure what your environment looks like, but if it is a hijacked production setup, then all the dynflow-sidekiq@* services need to be cycled to pick up the changes.

jturel: Just the usual katello devel environment - centos7-katello-devel.

I guess it's that these extensions are prepended first, and then the REX extensions are prepended and effectively overriding this method?

::Host.first.remote_execution_proxies(nil) isn't hitting this method in the console :(

jturel:
[test katello]

jturel (approved):
ACK!

I was able to toggle the setting and see the action move between my default and external smart proxies. Thanks @adamruzicka

jturel:
@adamruzicka rubocop :(

PR #9618, opened 2021-09-07
Fixes #33430 - do not error on pulpcore with 1 worker

jlsherrill:


jlsherrill:
easy to test, shut down all but 1 worker and try to use pulp3

app/models/katello/ping.rb
```diff
@@ -195,10 +195,7 @@ def pulp3_without_auth(url)
         end
 
         workers = json["online_workers"] || []
-
-        unless workers.count > 1
-          fail _("Not all necessary pulp workers running at %s.") % url
-        end
+        fail _("Not all necessary pulp workers running at %s.") % url if workers.empty?
```
rverdile, line 198: Could the error message be "No pulp workers running..." if it's failing when workers is empty?

jlsherrill: updated!

jjeffers (approved):
Test all 3 conditions - >1, 1, and 0 pulp workers - worked as indicated. 

If there are no pulp workers, the error message appears. 1 or more pulp workers allows the operation to complete.

PR #9621, opened 2021-09-09
Fixes #33442 - Old Registration URL doesn't redirect to the new URL

stejskalleos:


jlsherrill (approved):
ack! works for me! Thanks @stejskalleos

PR #9622, opened 2021-09-09
Fixes #33443 - adds last_indexed to repos

jjeffers:
This PR adjust the IndexContent action to compare the `last_content_changed` and a new `last_indexed` timestamp. Content will be indexed if the contents have been changed since the last index time, otherwise, they are not.

Previously the `contents_changed` option was not being set in a way that would reflect the state of the repository in some circumstances (such as a failed index content action in a sync).

### DIFF
--- a/app/lib/actions/katello/repository/index_content.rb
+++ b/app/lib/actions/katello/repository/index_content.rb
@@ -2,21 +2,23 @@ module Actions
   module Katello
     module Repository
       class IndexContent < Actions::EntryAction
-        middleware.use Actions::Middleware::ExecuteIfContentsChanged
-
         input_format do
           param :id, Integer
-          param :dependency, Hash
-          param :contents_changed
-          param :matching_content
           param :source_repository_id
           param :full_index
+          param :force_index
         end
 
         def run
           source_repository = ::Katello::Repository.find(input[:source_repository_id]) if input[:source_repository_id]
           repo = ::Katello::Repository.find(input[:id])
-          repo.index_content(source_repository: source_repository, full_index: input[:full_index].present?)
+
+          if input[:force_index] || (repo.last_contents_changed >= repo.last_indexed)
+            repo.index_content(source_repository: source_repository, full_index: input[:full_index].present?)
+            repo.update(:last_indexed => DateTime.now)
+          else
+            output[:index_skipped] = true
+          end
         end
       end
     end
--- a/app/lib/actions/katello/repository/sync.rb
+++ b/app/lib/actions/katello/repository/sync.rb
@@ -47,9 +47,7 @@ module Actions
                                         pulp_sync_options)
               output = sync_action.output
 
-              contents_changed = skip_metadata_check || output[:contents_changed]
-
-              plan_action(Katello::Repository::IndexContent, :id => repo.id, :contents_changed => contents_changed)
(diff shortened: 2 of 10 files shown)


jjeffers:
[test katello]

app/lib/actions/katello/repository/index_content.rb
```diff
@@ -2,21 +2,20 @@ module Actions
   module Katello
     module Repository
       class IndexContent < Actions::EntryAction
-        middleware.use Actions::Middleware::ExecuteIfContentsChanged
-
         input_format do
           param :id, Integer
-          param :dependency, Hash
-          param :contents_changed
```
jlsherrill, line 10: overall this looks good, but the main sync action, uses the contents_changed as a 'force' flag here: 

https://github.com/Katello/katello/blob/master/app/lib/actions/katello/repository/sync.rb#L50-L52

so it might be helpful to have an actual 'force' flag that can be passed in to skip the if timestamp check (that would be more clear too)

app/lib/actions/katello/repository/index_content.rb
```diff
@@ -2,21 +2,20 @@ module Actions
...
           param :source_repository_id
           param :full_index
         end
 
         def run
           source_repository = ::Katello::Repository.find(input[:source_repository_id]) if input[:source_repository_id]
           repo = ::Katello::Repository.find(input[:id])
-          repo.index_content(source_repository: source_repository, full_index: input[:full_index].present?)
+
+          if repo.last_contents_changed >= repo.last_indexed
```
jlsherrill, line 15: just a note if i review this (or someone else), we need to make sure that CV publish/promote also shows the correct content with this change

app/lib/actions/katello/repository/index_content.rb
jjeffers: That is the sort of thing that was nagging at my mind!

app/lib/actions/katello/repository/sync.rb
```diff
@@ -49,7 +49,7 @@ def plan(repo, options = {})
 
               contents_changed = skip_metadata_check || output[:contents_changed]
 
-              plan_action(Katello::Repository::IndexContent, :id => repo.id, :contents_changed => contents_changed)
+              plan_action(Katello::Repository::IndexContent, :id => repo.id, :force_index => contents_changed)
```
jlsherrill, line 52: i think here, we really just want:  :force_index => skip_metadata_check
I don't think we want contents_changed to affect it or not (relying on the timestamp for that)

app/lib/actions/katello/repository/index_content.rb
```diff
@@ -2,21 +2,21 @@ module Actions
...
 
         def run
           source_repository = ::Katello::Repository.find(input[:source_repository_id]) if input[:source_repository_id]
           repo = ::Katello::Repository.find(input[:id])
-          repo.index_content(source_repository: source_repository, full_index: input[:full_index].present?)
+
+          if input[:force_index] || (repo.last_contents_changed >= repo.last_indexed)
+            repo.index_content(source_repository: source_repository, full_index: input[:full_index].present?)
+            repo.update(:last_indexed => DateTime.now)
+          end
```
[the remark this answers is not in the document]
jjeffers: That is an excellent suggestion.

jlsherrill (approved):
ACK, tested and worked great.  Don't forget to squash and remove the [WIP] :)

PR #9623, opened 2021-09-09
Fixes #33450 - remove broken katello:refresh_sync_schedule

ianballou:
It doesn't do anything.


jlsherrill (approved):
But it does something! It throws an error!

PR #9630, opened 2021-09-10
Fixes #33465 - don't show links to default org view in proxy UI

ianballou:
We've had this issue before in the old UI: https://projects.theforeman.org/issues/27783

This PR makes the default org view redirect to `/products` instead of the actual default org view since it's bugged in the UI.


jlsherrill:
this issue # is incorrect

sjha4:
Correct issue is: https://projects.theforeman.org/issues/33465

ianballou:
Ah oops, that's the old issue when this first occurred :)

ianballou:
[test katello]

jlsherrill (approved):

PR #9631, opened 2021-09-10
Fixes #33466 - show proper repo count of capsule content

jlsherrill:


sjha4:
This works as expected for non default CVs. 

Couple of questions: we are also showing Default Organization View in the content tab for Library environment on the capsule..Does that seem right?

And since we are showing the default view currently, the number of archived repos always comes to 0 from this: 
`::Katello::ContentViewVersion.in_environment(env).find_by(:content_view => content_view)&.repositories&.archived&.count`

jlsherrill:
good point, i think we should probably still show it, and show the correct amount

jlsherrill:
turns out there was a helper method to do that exact thing

sjha4:
[test katello]

sjha4 (approved):
Tested and work great. APJ 👍🏽

PR #9632, opened 2021-09-10
Fixes #33467 - Add component view count to Versions API

sjha4:
Adds a new field in versions API response with component view counts for composite content views. Value will be 0 for non-composite CVs and also 0 for empty Composite CV versions.

Screenshot from 2021-09-10 14-40-58

Note: An unpublished component CV is not considered as a component of Composite CV **version**. Count will be 0 and the API `/katello/api/v2/content_view_versions?composite_version_id=$VERSION_ID` will not return unpublished component CVs either.

### DIFF
--- a/app/views/katello/api/v2/content_view_versions/base.json.rabl
+++ b/app/views/katello/api/v2/content_view_versions/base.json.rabl
@@ -13,6 +13,10 @@ node do |version|
   version.content_counts_map
 end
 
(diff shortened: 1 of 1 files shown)


Andrewgdewar (approved):
:sunglasses:  :beers:  Thank you! Works perfectly!

sjha4:
[test katello]

PR #9635, opened 2021-09-13
Refs #33078 - Move forceRender inside hook

jeremylenz:
I used JavaScript Sets in TracesTab.js to keep track of which table rows are selected.  The biggest disadvantage of this was that you would have to call a `forceRender()` function each time you mutate the set, or else React wouldn't know when to update the DOM.  This PR automates that and removes the need to call `forceRender()` yourself.

Usage before:
```js
const [mySet, forceRender] = useSet([]);
...
mySet.add(3);
forceRender();
```

Usage after:
```js
const mySet = useSet([]);
...
mySet.add(3); // forceRender is called for you
```

### DIFF
--- a/webpack/components/Table/TableHooks.js
+++ b/webpack/components/Table/TableHooks.js
@@ -1,11 +1,46 @@
 import { useState, useRef } from 'react';
 
+class ReactConnectedSet extends Set {
+  constructor(initialValue, forceRender) {
+    super();
+    this.forceRender = forceRender;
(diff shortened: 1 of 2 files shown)


Andrewgdewar (approved):
Ack! 
Although I do wish there were a way to do this that was simplified. 
The more code in the system that is doing a thing that does a thing that does a thing, the more difficult it is to maintain and fix when bugs come along.

PR #9636, opened 2021-09-13
Fixes #33472 - Show unstarted sync task as waiting

jlsherrill:


rverdile (commented):
Not sure, but does this need an update as well? 
https://github.com/Katello/katello/blob/master/app/lib/actions/pulp3/repository/presenters/repair_presenter.rb#L42

jlsherrill:
@rverdile  that particular task doesn't show up on the sync status page, so i don't think its needed there IMO.  Although i guess it doesn't hurt anything, i'll update it!

rverdile:
> @rverdile that particular task doesn't show up on the sync status page, so i don't think its needed there IMO. Although i guess it doesn't hurt anything, i'll update it!

when is `repair_presenter.rb` displayed?

jlsherrill:
its not shown on the "sync status" page where this is more prominent.  It is still shown if you initiate a repository repair operation from the repo details page and then watch the task details page that it redirects you to

ianballou (approved):
Looks good to me.  It's a little weird that a "Verify Content Checksum" sync with nothing to do will look like this at the end:

```
Waiting to start.
--------------------------------
Identify corrupted units: 0/0
Identify missing units: 0/0
Repair corrupted units: 0/0
```

It's better than `Total steps: 0/0` though.  Anyway, it's a separate issue.

PR #9642, opened 2021-09-16
Refs #33496 - increase qpid connection security

evgeni:
qpid_proton 0.34 defaulted to "ANONYMOUS_PEER", which means
"do not require a certificate nor cipher authorization"

0.35 switched to "VERIFY_PEER_NAME", which means
"require valid certificate and matching name"

the *name* part is what's breaking our setup, but we still can verify
the certificate! so let's use "VERIFY_PEER", which means
"require peer to provide a valid identifying certificate"


evgeni:
@jturel this is what happens if Evgeni has enough coffee and reads the docs again ;-)

evgeni:
even better: https://github.com/Katello/katello/pull/9643

evgeni:
closing in favor of #9643

PR #9643, opened 2021-09-16
Refs #33496 - set the peers host name to be able to verify it

evgeni:
instead of not checking the name in the cert, correctly set it, so that
it actually can be verified

I have no idea why qpid_proton doesn't automatically parse this from the
URL.

### DIFF
--- a/app/lib/katello/qpid/connection.rb
+++ b/app/lib/katello/qpid/connection.rb
@@ -74,12 +74,13 @@ module Katello
       def initialize(url:, ssl_cert_file:, ssl_key_file:, ssl_ca_file:)
         @url = url
         ssl_domain = ::Qpid::Proton::SSLDomain.new(::Qpid::Proton::SSLDomain::MODE_CLIENT)
-        ssl_domain.peer_authentication(::Qpid::Proton::SSLDomain::ANONYMOUS_PEER)
+        ssl_domain.peer_authentication(::Qpid::Proton::SSLDomain::VERIFY_PEER_NAME)
         ssl_domain.credentials(ssl_cert_file, ssl_key_file, nil) if ssl_cert_file && ssl_key_file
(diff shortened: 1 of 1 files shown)


evgeni:
FWIW, I've opened https://issues.apache.org/jira/browse/PROTON-2434

ekohl (commented):
There is no minimum version on the gem:
https://github.com/Katello/katello/blob/4e80f885481415961494563cf2b097855817f617/katello.gemspec#L40

Should there be? I think it'd be nice to guarantee certificate verification. I think that either means setting 0.35 as a minimum version or explicitly enabling peer authentication, right?

evgeni:
> Should there be? I think it'd be nice to guarantee certificate verification. I think that either means setting 0.35 as a minimum version or explicitly enabling peer authentication, right?

Yeah, on 0.34 this would be silently ignored, as it defaults to `ANONYMOUS_PEER`. Forcing 0.35 just because of that seems silly.
Setting `VERIFY_PEER_NAME` explicitly sounds good.

@jturel what'cha think?

evgeni:
On the other hand, we're updating all stable branches to 0.35 anyways, so 🤷‍♀️

jturel:
Considering we're seeing default behavior change, +1 to setting `VERIFY_PEER_NAME` explicitly and not including a min gem version

evgeni:
> Considering we're seeing default behavior change, +1 to setting `VERIFY_PEER_NAME` explicitly and not including a min gem version

Updated.

ekohl:
Is this now a separate Redmine issue? Other than that :+1:

jturel (approved):
APJ

jturel:
> Is this now a separate Redmine issue? Other than that

I'm good as-is. The redmine could be renamed to reflect the evolution perhaps

evgeni:
🥦

PR #9644, opened 2021-09-16
Fixes #33498 - use CSS to hide tooltips

jjeffers:
This is a workaround to the other option to address tooltips that are not disappearing after a click. 

The alternative includes replacing existing react table components with newer ones.

### DIFF
--- a/webpack/containers/Application/overrides.scss
+++ b/webpack/containers/Application/overrides.scss
@@ -75,3 +75,22 @@ html .pagination-pf-pagesize.btn-group {
     margin-left: 200px;
   }
 }
+
+@keyframes hideme {
+  0%, 70% {
+    opacity: 1;
+  }
+
+  100% {
+    opacity: 0;
+  }
+}
+
+.tooltip {
+  margin:auto;
+  animation-direction: normal;
+  animation-name: hideme;
+  animation-duration: 2s;
+  animation-iteration-count: 1;
+  animation-fill-mode: forwards;
+}


webpack/containers/Application/overrides.scss
```diff
@@ -75,3 +75,38 @@ html .pagination-pf-pagesize.btn-group {
...
+    opacity: 1;
+  }
+
+  90% {
+    opacity: 1;
+  }
+
+  100% {
+    opacity: 0;
+  }
```
jeremylenz, line 102: I [learned](https://css-tricks.com/snippets/css/keyframe-animation-syntax/) that you can comma-separate values so it's a bit less repetitive.  Also I think it looks nicer if the fadeout starts at 70% instead of 90:
```suggestion
  0%, 70% {
    opacity: 1;
  }

  100% {
    opacity: 0;
  }
```

webpack/containers/Application/overrides.scss
```diff
@@ -75,3 +75,38 @@ html .pagination-pf-pagesize.btn-group {
...
+}
+
+.tooltip {
+  margin:auto;
+  animation-direction: normal;
+  animation-name: hideme;
+  animation-duration: 6s;
```
jeremylenz, line 109:
```suggestion
  animation-duration: 2s;
```
6 seconds is too long to wait!

jeremylenz (commented):
Thanks @jjeffers for looking into it!

Comments below

jeremylenz (approved):
LGTM! :+1:

PR #9647, opened 2021-09-16
Fixes #33509 - show added content counts on sync

jlsherrill:
This adds a new content type attribute called primary_content
if marked to true, during indexing the count of added units
will be recorded on the index steps output and added
to the task output along with the other status messages


ianballou (approved):
Looks good to me!

Not totally related, but I noticed on the sync status page that the "Total steps" are always x/x instead of x/y.  Maybe that was just a weird coincidence for me...

PR #9648, opened 2021-09-16
Fixes #33510 - Composite CV publishing w/ modify api is broken

ianballou:
Composite CV publishes were wrong because `copy_units` was assuming to clear out the destination repo before starting copying.  In the composite case, the destination repo will have content from the previous source repo in it.

To test:

Create a composite content view from two content views with multiple repositories in them. Ensure that all content is in the resulting composite content view version.


evgeni:
```
test_0007_fully plans out unit copying with multiple source repositories – Actions::Katello::Repository::CloneToVersion
Error
Unexpected input [:remove_all] present.
/home/jenkins/workspace/katello-pr-test/test/actions/katello/repository/clone_to_version_test.rb:121
```

This test failure looks related?

jlsherrill (approved):
ACK!

PR #9629, opened 2021-09-10
Fixes #33454 - Add repositories node to generic content unit results

sjha4:

PR #9650, opened 2021-09-17
Fixes #33435 - Other Content Types Landing Page UI

rverdile:
This adds a new tab to the navigation bar: `Content Types > Other Content Types`. This links to a listing of content types, always filtered by one type. The only type supported here is Python Packages, which is selected by default.

Navigating to this page, once you have python packages created, you'll see a table with  "Name" and "Version" headers and those packages listed. It's also searchable with autocomplete. If not content has been created, there's a message saying so.

It should be possible to add more content types to this page by adding them to `ContentConfig.js`, in a similar structure to what is already there for python packages. That might be a fun test.

### DIFF
--- a/app/services/katello/repository_type.rb
+++ b/app/services/katello/repository_type.rb
@@ -127,7 +127,7 @@ module Katello
 
     class ContentType
       attr_accessor :model_class, :priority, :pulp2_service_class, :pulp3_service_class, :index, :uploadable, :removable,
-                    :primary_content, :index_on_pulp3
+                    :primary_content, :index_on_pulp3, :generic_browser
 
       def initialize(options)
         self.model_class = options[:model_class]
@@ -139,10 +139,18 @@ module Katello
         self.uploadable = options[:uploadable] || false
         self.removable = options[:removable] || false
         self.primary_content = options[:primary_content] || false
+        self.generic_browser = options[:generic_browser]
       end
 
       def label
         self.model_class::CONTENT_TYPE
+      end
+
+      def as_json(_options)
+        {
+          label: label,
+          generic_browser: generic_browser
+        }
       end
     end
 
@@ -150,13 +158,7 @@ module Katello
       attr_accessor :pulp3_api, :pulp3_model, :content_type, :filename_key, :duplicates_allowed
 
       def initialize(options)
-        self.model_class = options[:model_class]
-        self.priority = options[:priority] || 99
-        self.pulp3_service_class = options[:pulp3_service_class]
-        self.index = options[:index].nil? ? true : options[:index]
-        self.index_on_pulp3 = options[:index_on_pulp3].nil? ? true : options[:index_on_pulp3]
-        self.uploadable = options[:uploadable] || false
-        self.removable = options[:removable] || false
+        super
         self.pulp3_api = options[:pulp3_api]
         self.pulp3_model = options[:pulp3_model]
         self.content_type = options[:content_type]
--- a/webpack/scenes/Content/ContentActions.js
+++ b/webpack/scenes/Content/ContentActions.js
@@ -1,0 +1,18 @@
+import { API_OPERATIONS, get } from 'foremanReact/redux/API';
+import api from '../../services/api';
+import { CONTENT_KEY, CONTENT_TYPES_KEY } from './ContentConstants';
+
+export const getContent = (contentType, params) => get({
+  type: API_OPERATIONS.GET,
+  key: CONTENT_KEY,
+  params,
+  url: api.getApiUrl(`/${contentType}`),
+});
+
+export const getContentTypes = () => get({
+  type: API_OPERATIONS.GET,
+  key: CONTENT_TYPES_KEY,
+  url: api.getApiUrl('/repositories/content_types'),
+});
+
+export default getContent;
--- a/webpack/scenes/Content/ContentConfig.js
+++ b/webpack/scenes/Content/ContentConfig.js
@@ -1,0 +1,11 @@
+import { translate as __ } from 'foremanReact/common/I18n';
+
+export default () => [
+  {
+    names: { title: __('Python Packages'), plural: 'python_packages', singular: 'python_package' },
+    columnHeaders: [
+      { title: __('Name'), getProperty: unit => unit?.name },
+      { title: __('Version'), getProperty: unit => unit?.version },
+    ],
+  },
+];
--- a/webpack/scenes/Content/ContentPage.js
+++ b/webpack/scenes/Content/ContentPage.js
@@ -1,0 +1,73 @@
+import React, { useEffect, useState } from 'react';
+import { useSelector, useDispatch } from 'react-redux';
+import { Grid, GridItem, TextContent, Text, TextVariants } from '@patternfly/react-core';
+import { isEmpty } from 'lodash';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { STATUS } from 'foremanReact/constants';
+import ContentTable from './Table/ContentTable';
+import { selectContentTypes, selectContentTypesStatus } from './ContentSelectors';
+import ContentConfig from './ContentConfig';
+import { getContentTypes } from './ContentActions';
+import Loading from '../../components/Loading';
+import EmptyStateMessage from '../../components/Table/EmptyStateMessage';
+
+const ContentPage = () => {
+  const dispatch = useDispatch();
+  const contentTypesResponse = useSelector(selectContentTypes);
+  const contentTypesStatus = useSelector(selectContentTypesStatus);
+  const [selectedContentType, setSelectedContentType] = useState(null);
+  const [contentTypes, setContentTypes] = useState(null);
+
+  useEffect(() => {
+    // Set list of enabled content types and initial content type
+    const buildContentTypes = () => {
+      const types = {};
+      contentTypesResponse.forEach((type) => {
+        if (type.generic_browser) {
+          const typeConfig = ContentConfig().find(config => config.names.singular === type.label);
+          if (typeConfig) {
+            const { names } = typeConfig;
+            types[names.title] = [names.singular, names.plural];
+          }
+        }
+      });
+      return types;
+    };
+
+    if (contentTypesStatus === STATUS.RESOLVED) {
+      const enabledContentTypes = buildContentTypes();
+      setSelectedContentType(Object.keys(enabledContentTypes)[0]);
+      setContentTypes(enabledContentTypes);
+    }
+  }, [contentTypesStatus, contentTypesResponse]);
+
+  useEffect(() => {
+    dispatch(getContentTypes());
+  }, [dispatch]);
+
+  if (contentTypesStatus === STATUS.PENDING) {
+    return <Loading />;
+  } else if (isEmpty(contentTypes)) {
+    return (
+      <EmptyStateMessage
+        title="Browsable content types not found."
+        body="No content types are enabled for this page."
+      />
+    );
+  }
+
+  return (
+    <Grid className="grid-with-margin">
+      <GridItem span={12}>
+        <TextContent>
+          <Text component={TextVariants.h1}>{__(`${selectedContentType}`)}</Text>
+        </TextContent>
+      </GridItem>
+      <GridItem span={12}>
+        <ContentTable {...{ selectedContentType, setSelectedContentType, contentTypes }} />
+      </GridItem>
+    </Grid>
+  );
+};
+
+export default ContentPage;
--- a/webpack/scenes/Content/Table/ContentTable.js
+++ b/webpack/scenes/Content/Table/ContentTable.js
@@ -1,0 +1,90 @@
+import React, { useState, useCallback } from 'react';
+import { useSelector } from 'react-redux';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { TableVariant, Thead, Tbody, Tr, Th, Td } from '@patternfly/react-table';
+import PropTypes from 'prop-types';
+import TableWrapper from '../../../components/Table/TableWrapper';
+import { getContent } from '../ContentActions';
+import { selectContent, selectContentStatus, selectContentError } from '../ContentSelectors';
+import SelectableDropdown from '../../../components/SelectableDropdown';
+import contentConfig from '../ContentConfig';
+
+const ContentTable = ({ selectedContentType, setSelectedContentType, contentTypes }) => {
+  const status = useSelector(selectContentStatus);
+  const error = useSelector(selectContentError);
+  const response = useSelector(selectContent);
+  const [searchQuery, updateSearchQuery] = useState('');
+  const { results, ...metadata } = response;
+
+  const { columnHeaders } = contentConfig().find(type =>
+    type.names.singular === contentTypes[selectedContentType][0]);
+
+  /* eslint-disable react/no-array-index-key */
+  function buildRows(details) {
+    const rows = [];
+    if (details) {
+      columnHeaders.forEach((header, index) =>
+        rows.push(<Td key={index}>{header.getProperty(details)}</Td>));
+    }
+    return rows;
+  }
+
+  return (
+    <TableWrapper
+      {...{
+        metadata,
+        searchQuery,
+        updateSearchQuery,
+        error,
+        status,
+      }}
+      key={selectedContentType}
+      variant={TableVariant.compact}
+      autocompleteEndpoint={`/${contentTypes[selectedContentType][1]}/auto_complete_search`}
+      emptyContentTitle={__(`You currently don't have any ${selectedContentType}.`)}
+      emptySearchTitle={__(`No matching ${selectedContentType} found`)}
+      emptyContentBody={__(`${selectedContentType} will appear here when created.`)}
+      emptySearchBody={__('Try changing your search settings.')}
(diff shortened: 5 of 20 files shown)


rverdile:
Not sure how to resolve the remaining eslint issues.

rverdile:
this should merge first: https://github.com/Katello/katello/pull/9629

sjha4:
Empty page with `Uncaught Error: Too many re-renders. React limits the number of renders to prevent an infinite loop.` on loading page with no python packages..

webpack/scenes/Content/ContentActions.js
```diff
@@ -0,0 +1,17 @@
+import { API_OPERATIONS, get } from 'foremanReact/redux/API';
+import api from '../../services/api';
+import { CONTENT_KEY, CONTENT_TYPES_KEY } from './ContentConstants';
+
+export const getContent = contentType => get({
+  type: API_OPERATIONS.GET,
+  key: CONTENT_KEY,
+  url: api.getApiUrl(`/${contentType}`),
+});
```
sjha4, line 9:
```suggestion
export const getContent = (contentType, params) => get({
  type: API_OPERATIONS.GET,
  key: CONTENT_KEY,
  params,
  url: api.getApiUrl(`/${contentType}`),
});
```

This will make pagination work..The pagination props are passed as params to the API.

webpack/scenes/Content/Table/ContentTable.js
```diff
@@ -0,0 +1,117 @@
...
+        }
+      });
+      setContentTypes(types);
+    }
+  }, [enabledTypesResponse, enabledTypesStatus]);
+
+  const fetchItems = useCallback(
+    () => getContent(contentTypes[contentTypeSelected][1]),
+    [contentTypes, contentTypeSelected],
+  );
```
sjha4, line 62:
```suggestion
  const fetchItems = useCallback(
    (params) => getContent(contentTypes[contentTypeSelected][1], params),
    [contentTypes, contentTypeSelected],
  );
```
Passing params to fetchItem will make pagination and search work..

webpack/scenes/Content/Table/ContentTable.js
```diff
@@ -0,0 +1,117 @@
...
+  const [columnHeaders, setColumnHeaders] = useState([]);
+  const [contentTypes, setContentTypes] = useState(null);
+
+  const { results, ...metadata } = response;
+
+  function buildRows(details) {
+    const rows = [];
+    if (details) {
+      columnHeaders.forEach(header =>
+        rows.push(<Td>{header.getProperty(details)}</Td>));
```
sjha4, line 31:
```suggestion
      columnHeaders.forEach((header, index) =>
        rows.push(<Td key={index}>{header.getProperty(details)}</Td>));
```
This will resolve the console error about needing unique keys. You will need to disable an eslint rule to use this.
`/* eslint-disable react/no-array-index-key */`

rverdile:
updated. thanks @sjha4

webpack/scenes/Content/Table/ContentTable.js
```diff
@@ -0,0 +1,118 @@
...
+    if (details) {
+      columnHeaders.forEach((header, index) =>
+        rows.push(<Td key={index}>{header.getProperty(details)}</Td>));
+    }
+    return rows;
+  }
+
+  useEffect(() => {
+    dispatch(getContentTypes());
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
```
sjha4, line 39: You can add dispatch as dependency here and remove the eslint comment..

webpack/scenes/Content/Table/ContentTable.js
```diff
@@ -0,0 +1,118 @@
...
+import Loading from '../../../components/Loading';
+
+const ContentTable = () => {
+  const dispatch = useDispatch();
+  const status = useSelector(selectContentStatus);
+  const error = useSelector(selectContentError);
+  const response = useSelector(selectContent);
+  const enabledTypesResponse = useSelector(selectContentTypes);
+  const enabledTypesStatus = useSelector(selectContentTypesStatus);
+  const [contentTypeSelected, setContentTypeSelected] = useState('Python Packages');
```
sjha4, line 20: You could move this to the parent ContentPage and pass down as a prop..This will let the Header on Page change with type selection.

rverdile: cool!

webpack/scenes/Content/Table/ContentTable.js
```diff
@@ -0,0 +1,118 @@
...
+        searchQuery,
+        updateSearchQuery,
+        error,
+        status,
+        fetchItems,
+      }}
+      variant={TableVariant.compact}
+      autocompleteEndpoint={`/${contentTypes[contentTypeSelected][1]}/auto_complete_search`}
+      emptyContentTitle={`You currently don't have any ${contentTypeSelected}.`}
+      emptyContentBody={`${contentTypeSelected} will appear here when created.`}
```
sjha4, line 87: if you add `additionalListeners={[contentTypeSelected]}` to the Table here, it will update when content type is picked in the selector. That will be needed when we use this page for other content types..

webpack/scenes/Content/Table/ContentTable.js
```diff
@@ -0,0 +1,118 @@
...
+      <Thead>
+        <Tr>
+          {columnHeaders.map(col =>
+            <Th key={col.title}>{col.title}</Th>)}
+        </Tr>
+      </Thead>
+      <Tbody>
+        {results?.map(details => (
+          <Tr key={`${details.pulp_id}`}>
+            {buildRows(details)}
```
sjha4, line 109: Not all of our content rabls return a pulp_id but I believe all return an id which will also be needed to route to details page. Maybe use details.id instead of pulp_id.

webpack/scenes/Content/ContentConfig.js
```diff
@@ -0,0 +1,11 @@
+import { translate as __ } from 'foremanReact/common/I18n';
+
+export default () => [
+  {
+    names: { title: __('Python Packages'), plural: 'python_packages', singular: 'python_package' },
```
jlsherrill, line 5: is there any way we can have this information pulled from the api?

rverdile: Just the names or all of the information? Having the separate config file for this might be necessary to allow different content types to render information with different react components.

rverdile:
Updated. Added test/fixtures and @sjha4's suggestions.

webpack/scenes/Content/ContentConfig.js
jlsherrill: just the names, but potentially the columns in the future as well.  @sjha4 do you have any ideas how we could do this?  the idea is still that we can drop in the repository_type definition on the ruby server and thats it, the UI will adapt as needed

rverdile: I don't know how much variation in the react components is expected, but maybe when a react component is desired it can just be associated with a keyword from the api.

sjha4: I like this approach of having a UI config file to control what shows on the UI..

But if we want to make it more generic, we could do a default list of columns to display (name, version) to begin with and get this config file to loop over content types and form this config automatically. We could still override a content config if we wanted to.

jlsherrill: I'm not necessarily opposed to it, it just increases the footprint of where things need to be added.  If we do go this route, we'll need something that disables content types that aren't actually enabled on the server (for example, we wouldn't want to show python packages if python isn't enabled)

rverdile: What if I add a kind of default that can have column headers added to it from the registration file, this way the only time it would need to be reconfigured on the javascript side is if different react components need to be specified.

And I'll make the default selected type whatever the first type returned from the API is, rather than python packages specifically.

rverdile:
I added a default config that uses information from the registration file. It will let you add an arbitrary number of column headers to the table. You can also override the default config by adding the configuration to `ContentConfig.js` instead. 

The `names` hash still needs to be specified in the registration file. An example of what an override config looks like is in a comment in `ContentConfig.js`

I also moved the API logic from `ContentTable.js` to `ContentPage.js`

webpack/scenes/Content/ContentConfig.js
```diff
@@ -0,0 +1,62 @@
...
+}
+
+export function getConfiguredTypes(enabledTypes) {
+  const types = [];
+  Object.entries(enabledTypes).forEach(([key, value]) => {
+    if (configuredTypes().includes(value[0])) {
+      types.push(key);
+    }
+  });
+  return types;
```
sjha4, line 45: What is this doing exactly?

app/services/katello/repository_type.rb
```diff
@@ -154,11 +160,19 @@ def initialize(options)
         self.content_type = options[:content_type]
         self.filename_key = options[:filename_key]
         self.duplicates_allowed = options[:duplicates_allowed]
+        self.ui_config = options[:ui_config]
```
sjha4, line 163: Can we move this to parent ContentType so it's available for non-generic types too..You could add ui_config to to_json of parent and remove the overriding that limits it to generic types..Also would super() be helpful here for the common bits of initialize..

webpack/scenes/Content/ContentPage.js
```diff
@@ -0,0 +1,73 @@
...
+import { selectContentTypes, selectContentTypesStatus } from './ContentSelectors';
+import { contentConfig, defaultConfig, getConfiguredTypes, parseEnabledTypesResponse } from './ContentConfig';
+import Loading from '../../components/Loading';
+
+const ContentPage = () => {
+  const dispatch = useDispatch();
+  const enabledTypesResponse = useSelector(selectContentTypes);
+  const enabledTypesStatus = useSelector(selectContentTypesStatus);
+  const [contentTypeSelected, setContentTypeSelected] = useState(null);
+  const [typeSelectedConfig, setTypeSelectedConfig] = useState(null);
```
sjha4, line 18: nitpick: Thoughts on renaming these to selectedContentType and selectedContentTypeConfig?

webpack/scenes/Content/ContentConfig.js
rverdile: It's checking the enabled content types against what has a configuration defined in `ContentConfig()`. It returns a list of keys for content types that are enabled _and_ have a configuration specified. If a content type is in this list, it won't use `DefaultConfig()`.

It might benefit from a different name.

webpack/scenes/Content/ContentConfig.js
sjha4: > And I'll make the default selected type whatever the first type returned from the API is, rather than python packages specifically.

This will probably need to be plucked out of the URL in the future..
For example: If user is going to ostree nav link, url changes to /ostree and that'd be the selection in the dropdown..

webpack/scenes/Content/ContentPage.js
rverdile: I agree :)

Feel free to comment on the names of other things too. Sometimes naming is hard.

webpack/scenes/Content/ContentConfig.js
rverdile: In that case, when someone clicks "Other Content Types" in the navbar, should it go to `/content/:content_type` or `/:content_type`, rather than just `/content`?

webpack/scenes/Content/ContentConfig.js
sjha4: I wasn't able to get it working with an override config locally..Can you verify that it works for you and let me know the config you used?

webpack/scenes/Content/Table/ContentTable.js
```diff
@@ -0,0 +1,102 @@
...
+      }
+    >
+      <Thead>
+        <Tr>
+          {columnHeaders.map(col =>
+            <Th key={col.title}>{col.title}</Th>)}
+        </Tr>
+      </Thead>
+      <Tbody>
+        {results?.map(details => (
```
jeremylenz, line 74: :heart_eyes:

rverdile:
Reverted most of the previous changes. API logic is still in `ContentPage.js` to check that the list of selectable content types contains only enabled content types.

rverdile:
Added message to display when no content types are found.

webpack/scenes/Content/ContentPage.js
```diff
@@ -0,0 +1,73 @@
...
+import EmptyStateMessage from '../../components/Table/EmptyStateMessage';
+
+const ContentPage = () => {
+  const dispatch = useDispatch();
+  const contentTypesResponse = useSelector(selectContentTypes);
+  const contentTypesStatus = useSelector(selectContentTypesStatus);
+  const [selectedContentType, setSelectedContentType] = useState(null);
+  const [contentTypes, setContentTypes] = useState(null);
+
+  function buildContentTypes(response) {
```
sjha4, line 21:
```suggestion
  const buildContentTypes = () => {
```
The param shouldn't need to be passed here. Can directly use the contentTypesResponse

webpack/scenes/Content/ContentPage.js
```diff
@@ -0,0 +1,73 @@
...
+      setSelectedContentType(Object.keys(enabledContentTypes)[0]);
+      setContentTypes(enabledContentTypes);
+    }
+  }, [contentTypesStatus, contentTypesResponse]);
+
+  useEffect(() => {
+    dispatch(getContentTypes());
+  }, [dispatch]);
+
+  if (!contentTypes) {
```
sjha4, line 48:
```suggestion
  if (contentTypesStatus === STATUS.LOADING) {
```

jlsherrill (approved):
ACK!  Good work @rverdile !

webpack/scenes/Content/Table/ContentTable.js
```diff
@@ -0,0 +1,90 @@
...
+  return (
+    <TableWrapper
+      {...{
+        metadata,
+        searchQuery,
+        updateSearchQuery,
+        error,
+        status,
+      }}
+      variant={TableVariant.compact}
```
sjha4, line 41:
```suggestion
      key={contentTypes[selectedContentType][1]}
      variant={TableVariant.compact}
```

So I noticed the autosearch params are not updated when the selectedType changes. This will fix that.

webpack/scenes/Content/ContentConfig.js
sjha4: /content is fine for now..We will need to do some changes when we route explicit content_type types directly to this page.. For ex: routing /ansible_collection would route to this page with contentTypeSelected reading from the URL..

webpack/scenes/Content/ContentPage.js
Andrewgdewar: also! this is es5 syntax, we tend to want to avoid that in our javascript react code now since about 2012. 
Here's an in depth breakdown of what I mean: 
[es5 vs es6 functions (the big jump)](https://engineering.carsguide.com.au/es5-vs-es6-syntax-6c8350fa6998)
[all the more modern methods](https://www.freecodecamp.org/news/es5-to-esnext-heres-every-feature-added-to-javascript-since-2015-d0c255e13c6e/)

rverdile:
updated

webpack/scenes/Content/ContentPage.js
```diff
@@ -0,0 +1,73 @@
...
+const ContentPage = () => {
+  const dispatch = useDispatch();
+  const contentTypesResponse = useSelector(selectContentTypes);
+  const contentTypesStatus = useSelector(selectContentTypesStatus);
+  const [selectedContentType, setSelectedContentType] = useState(null);
+  const [contentTypes, setContentTypes] = useState(null);
+
+  useEffect(() => {
+    // Set list of enabled content types and initial content type
+    const buildContentTypes = () => {
```
rverdile, line 23: I moved `buildContentTypes` inside of the `useEffect` because it became a dependency after I changed the function syntax. Not sure if this is the best way to handle it.

webpack/scenes/Content/Table/ContentTable.js
jeremylenz: Changing the key will cause the entire TableWrapper to unmount and remount.  It would work but seems a bit drastic.  I'm confused because I would think adding `selectedContentType` to `additionalListeners` should be all you need?

sjha4: So the search component inside tableWrapper is not refreshed with the additionalListener..That causes the autosearch Endpoint to not change. When switching between different content types we need to update the search bar with correct endpoints..Hence the complete refresh of the TableWrapper.

webpack/scenes/Content/ContentPage.js
sjha4: Ya..That's good..

webpack/scenes/Content/Table/ContentTable.js
sjha4: We should actually remove the selectedContentType from additionalListeners since it's not needed anymore given we are re-rendering the whole table instead.

sjha4:
[test katello]

rverdile:
removed additionalListeners prop in `ContentTable` since it's being re-rendered with key

sjha4 (approved):
Great work @rverdile !! Works great and code looks good. 👍🏽 🥳

PR #9651, opened 2021-09-17
Fixes #33516 - (New Host details) Add errata recalculate action

jeremylenz:
- [ ] tests

This adds a recalculate action to the Errata table.


chris1984:
@jeremylenz I know the old page does not do this, but do you think it would be worth it upon clicking the recalcuate to refresh the page for the user, instead of them doing a refresh manually or search?

jeremylenz:
> but do you think it would be worth it upon clicking the recalculate to refresh the page for the user

I thought about this, but if we triggered a page refresh immediately on recalculate, I'm pretty sure the refresh would be complete before the applicability calculation finished. Short of finding the task and polling it, I wasn't sure how to get around that.  thoughts?

PR #9652, opened 2021-09-17
Errata select

parthaa:


webpack/components/extensions/HostDetails/Tabs/TracesTab.js
```diff
@@ -110,16 +87,10 @@ const TracesTab = () => {
...
-                disable: false,
-                props: {
-                  'aria-label': `check-${application}`,
-                },
-                isSelected: selectedTraces.has(id),
-                onSelect: (event, isSelected) => onRowSelect({ isSelected, traceId: id }),
-                rowIndex,
-                variant: 'checkbox',
-              }}
+              <TdSelect id = {id}
```
Andrewgdewar, line 90: I don't think this really warrants it's own component.

webpack/components/extensions/HostDetails/Tabs/ErrataTab.js
```diff
@@ -41,6 +43,11 @@ export const ErrataTab = () => {
   const response = useSelector(state => selectAPIResponse(state, HOST_ERRATA_KEY));
   const { results, ...metadata } = response;
   const status = useSelector(state => selectHostErrataStatus(state));
+
+  const selectAll = new SelectAllManager(results, metadata, useSet([]))
```
Andrewgdewar, line 47: I get what is happening here with the external class, but this is bringing external language principles into react, and obfuscating things more than is needed. 

I understand this is a principle of OOP that is appreciated in other languages, but in my experience, it usually ends in complicated difficult code to maintain when applied to React. 

If there is a simpler path that can have the code be contained within a component. I would prefer that.

parthaa: Working on it. Thanks!

webpack/components/extensions/HostDetails/Tabs/TracesTab.js
parthaa: agreed

parthaa:
This PR implements the following functions. 
* Errata Bulk Selection and Apply. 
* Select All Errata  and way to apply this to other future components
* Some of the methods in used in the previous TracerTab have been moved to TableHooks so that other components can use the same.

Select All follows the guildelines listed by https://www.patternfly.org/v4/guidelines/bulk-selection/#bulk-selection
* With nothing selected
nothing-selected

* With some items selected.
partial-selection

* With All items selected
select-all

* You can also deselect items after doing the select all
deselect

* You can filter your search results and select them all (even if they last multiple pages.)

select-all-search


Behavior
-------

Select All works in the following way
* We have 2 sets an `inclusionSet` and an `exclusionSet` that keep track of the selections
* The `inclusionSet` is used when Select All is not on. If its a `Select page` or `Select none` and you just selected some items those will be stored in  inclusionSet
* When Select All is checked the `inclusionSet` is cleared and `exclusionSet` comes into play. Any "de-selections" get added to the `exclusionSet`.
* Finally when there is an action on these selections the server gets sent a special hash for the bulk parameters that looks something like

```javascript
  {
    included: {
      ids: [<ids in the inclusionSet>],
      search: <search box entry>,
    },
    excluded: {
      ids: [<ids in the exclusionSet>],
    },
  };
```

The backend api then figures out how to use this to make the appropriate calls.

Expectation For Future bulk selects components
-------------------------------------

* Import useBulkSelect hook 
```import { useBulkSelect } from '../../../../components/Table/TableHooks';``

* In the component initialize this
```javascript
  const {
    onRowSelect, isSelected, searchQuery, selectedCount,
    updateSearchQuery, clearSelections, fetchBulkIds, ...selectAll
  } = useBulkSelect(results, metadata, [], 'errata_id');
```

When finally when the table is rendered pass some of the parameters

```jsx
        <TableWrapper
          {...{
                metadata,
                emptyContentTitle,
                emptyContentBody,
                emptySearchTitle,
                emptySearchBody,
                status,
                actionButtons,
                searchQuery,
                updateSearchQuery,
                selectedCount,
                }
          }
          {...selectAll}
          ....
          displaySelectAllCheckbox
        >
```

* When an action is applied to the selections use the `fetchBulkIds` call to get 
```javascript
  {
    included: {
      ids: [<ids in the inclusionSet>],
      search: <search box entry>,
    },
    excluded: {
      ids: [<ids in the exclusionSet>],
    },
  }
```

* Make your backend endpoint use the BulkExtensions concern. Look for at HostErrataExtensions on how to do this.

* Make your request.






This works the following way

webpack/components/Table/TableHooks.js
```diff
@@ -41,14 +42,143 @@ class ReactConnectedSet extends Set {
...
+    selectionSet,
+  };
+};
+
+export const useBulkSelect = (results, metadata, initialArry = [], idColumn = 'id') => {
+  const { selectionSet: inclusionSet, ...selectOptions } =
+                useSelectionSet(results, metadata, initialArry, idColumn);
+  const exclusionSet = useSet([]);
+  const [searchQuery, updateSearchQuery] = useState('');
+  const [isSelectAllChecked, setSelectAll] = useState(false);
```
jeremylenz, line 96: I always try to be wary of storing a value in state that could be computed instead.  Instead of trying to keep this value in sync with all the methods below, could we instead make it a function that looks at `inclusionSet`, `exclusionSet`, and `metadata.total` ?

webpack/components/Table/TableHooks.js
```diff
@@ -41,14 +42,143 @@ class ReactConnectedSet extends Set {
...
+
+  const isSelected = (id) => {
+    if (isSelectAllChecked) {
+      return !exclusionSet.has(id);
+    }
+    return inclusionSet.has(id);
+  };
+
+  const selectPage = () => {
+    setSelectAll(false);
```
jeremylenz, line 111: What if the total number of items happens to be the number of items on the page?

webpack/components/extensions/HostDetails/HostErrata/HostErrataConstants.js
```diff
@@ -1,3 +1,5 @@
 export const HOST_ERRATA_KEY = 'HOST_ERRATUM';
 export const HOST_ERRATA_APPLICABILITY_KEY = 'HOST_ERRATA_APPLICABILITY';
+export const HOST_ERRATA_APPLY_KEY = 'HOST_ERRATA_APPLY_KEY';
```
jeremylenz, line 3:
```suggestion
export const HOST_ERRATA_APPLY_KEY = 'HOST_ERRATUM_APPLY';
```

webpack/components/extensions/HostDetails/Tabs/ErrataTab.js
```diff
@@ -47,7 +46,7 @@ export const ErrataTab = () => {
     __('Type'),
     __('Severity'),
     __('Synopsis'),
-    __('Published Date'),
+    __('Published date'),
```
jeremylenz, line 49: :+1:  :+1:

webpack/components/Table/TableHooks.js
```diff
@@ -41,14 +42,143 @@ class ReactConnectedSet extends Set {
...
+    selectedCount,
+    areAllRowsOnPageSelected,
+    selectPage,
+    selectNone,
+    isSelected,
+    selectionSet,
+  };
+};
+
+export const useBulkSelect = (results, metadata, initialArry = [], idColumn = 'id') => {
```
jeremylenz, line 91: `useBulkSelect` and `useSelectionSet` are so important that I think they warrant their own Jest tests.  Or, make sure to thoroughly test all functionality via the ErrataTab tests.

webpack/components/extensions/HostDetails/Tabs/ErrataTab.js
```diff
@@ -81,12 +87,22 @@ export const ErrataTab = () => {
...
+    }
+  };
+
   const actionButtons = (
     <Split hasGutter>
       <SplitItem>
         <ActionList isIconList>
           <ActionListItem>
-            <Button isDisabled> {__('Apply')} </Button>
+            <Button isDisabled={selectedCount === 0} onClick={applyByKatelloAgent}> {__('Apply')} </Button>
```
jeremylenz, line 105: This Apply button works via Katello agent, and the one on Traces works via REX.  We should probably think about how we can be more consistent there, but that's probably outside the scope here

webpack/components/extensions/HostDetails/Tabs/TracesTab.js
```diff
@@ -6,14 +6,13 @@ import { useSelector, useDispatch } from 'react-redux';
 import { selectAPIResponse } from 'foremanReact/redux/API/APISelectors';
 import EnableTracerEmptyState from './EnableTracerEmptyState';
 import TableWrapper from '../../../Table/TableWrapper';
-import useSet from '../../../Table/TableHooks';
+import { useSelectionSet } from '../../../Table/TableHooks';
```
jeremylenz, line 9: How difficult would it be to add the concern to the host traces controller and go all out with `useBulkSelect` here?

jeremylenz (commented):
Great work @parthaa 

I'm excited about it!  Left comments below

webpack/components/Table/TableHooks.js
jeremylenz: If you want to be clever you could even put it in a `useMemo` :wink:

webpack/components/SelectAllCheckbox/index.js
```diff
@@ -12,64 +13,82 @@ const SelectAllCheckbox = ({
...
-  const onSelectAllCheckboxChange = () => {
-    if (isSelectAllChecked === false) {
-      return selectPage();
+  const onSelectAllCheckboxChange = (checked) => {
+    setSelectionToggle(checked);
+    if (checked) {
+      if (selectionState === 'page') {
+        return selectPage();
+      }
+      return setSelectAllChecked(true);
```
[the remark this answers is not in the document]
parthaa: trying to remember. I think I  removed the returns originally and needed them for some reason. Will play around.

app/controllers/katello/concerns/api/v2/bulk_extensions.rb
```diff
@@ -0,0 +1,36 @@
...
+          fail HttpErrors::BadRequest, _("No items have been specified.")
+        end
+
+        items = model_scope
+        if bulk_params[:included][:ids]
+          items = model_scope.where(key => bulk_params[:included][:ids])
+        end
+
+        if bulk_params[:included][:search]
+          search_items = model_scope.search_for(bulk_params[:included][:search])
```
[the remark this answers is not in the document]
parthaa: Thats a good point . The caller is doing that filtering though
https://github.com/Katello/katello/pull/9652/files#diff-fca4d0425961fd92d2a474e34045372e00a96c1c1ea9a59fa3da8da52740570aR132-R152

``` javascript
const fetchBulkIds = () => {
    const selected = {
      included: {
        ids: [],
        search: null,
      },
      excluded: {
        ids: [],
      },
    };

    if (isSelectAllChecked) {
      selected.included.search = searchQuery;
      selected.excluded.ids = [...exclusionSet];
    } else if (!isEmpty(inclusionSet)) {
      selected.included.ids = [...inclusionSet];
    } else {
      return {};
    }
    return selected;
  };

```

webpack/components/SelectAllCheckbox/index.js
```diff
@@ -12,64 +13,82 @@ const SelectAllCheckbox = ({
...
+    <DropdownItem key="select-none" id="none" component="button" onClick={handleSelectNone} >
       {`${__('Select none')} (0)`}
     </DropdownItem>,
-    <DropdownItem key="select-page" component="button" isDisabled={areAllRowsOnPageSelected} onClick={handleSelectPage}>
+    <DropdownItem key="select-page" id="page" component="button" isDisabled={areAllRowsOnPageSelected} onClick={handleSelectPage}>
       {`${__('Select page')} (${pageRowCount})`}
     </DropdownItem>,
   ];
+  if (isSelectAllChecked !== null) {
+    const selectAll = (
```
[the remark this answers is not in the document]
parthaa: more a formatting thing ?.. Will figure out

webpack/components/SelectAllCheckbox/index.js
```diff
@@ -12,64 +13,82 @@ const SelectAllCheckbox = ({
...
   const handleSelectNone = () => {
     setSelectAllDropdownOpen(false);
+    setSelectionToggle(false);
     selectNone();
   };
 
+  const onSelect = (event) => {
+    const { id } = event.currentTarget;
+    if (id) {
+      setSelectionState(id);
```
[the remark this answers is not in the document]
parthaa: ID values are `all`,  `page` or `none`. So basically I introduced a `selectionState` state value to keep track of "are we in page mode or none more or all mode"
Notice the `id` values in the code below.

``` javascript
   <DropdownItem key="select-none" id="none" component="button" onClick={handleSelectNone} >
      {`${__('Select none')} (0)`}
    </DropdownItem>,
    <DropdownItem key="select-page" component="button" isDisabled={areAllRowsOnPageSelected} onClick={handleSelectPage}>
    <DropdownItem key="select-page" id="page" component="button" isDisabled={areAllRowsOnPageSelected} onClick={handleSelectPage}>
      {`${__('Select page')} (${pageRowCount})`}
    </DropdownItem>,
      <DropdownItem key="select-all" id="all" component="button" isDisabled={isSelectAllChecked} onClick={handleSelectAll}>
        {`${__('Select all')} (${totalCount})`}
      </DropdownItem>
```

So when you change the selection, 
- the `selectionState` is reset
- the `useEffect` is invoked when the state is changed which updates the `selectionToggle` appropriately.

webpack/components/SelectAllCheckbox/index.js
```diff
@@ -12,64 +13,82 @@ const SelectAllCheckbox = ({
...
     if (selectedCount === 0) {
-      newCheckedState = false;
-    } else if (selectedCount > 0) {
-      newCheckedState = null; // null is partially-checked state
-    } else if (areAllRowsSelected) {
-      newCheckedState = true;
+      checkedState = false;
+    } else if (selectedCount !== totalCount) {
+      if (selectionState === 'page' && areAllRowsOnPageSelected) {
+        checkedState = true;
```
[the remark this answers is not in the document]
parthaa: checkedState is `null` by default. It only gets set to true if `selectionState === 'page' && areAllRowsOnPageSelected`

So think of the 3 cases here
- Nothing is selected `checkedState` is `false`
- Everything is selected `checkedState` is `true`
- Something is selected  
``` javascript 
     If the selectionState === "page" mode and all rows on the page are selected show  checkedState = true
     else checkedState = null
```
- Basically this if/else pattern removes the need for `allRowsSelected` param.

webpack/components/Table/TableHooks.js
parthaa: Agreed. Will work on em with your help tomorrow :)

webpack/components/Table/TableHooks.js
parthaa: Interesting idea. More than the `isSelectAllChecked` I need the `setSelectAllChecked`. How would I figure out the event where selection was set by the `SelectAllCheckbox` component. This state here gives me a handy way to enforce that.

webpack/components/Table/TableHooks.js
parthaa: If you chose `selectPage` you clearly revoking the idea of `selectAll` and instead doing `selectAll on this page`
setSelectAll is purely for marking the `selectAll mode` since I need that to keep track of the `exclusionSet`.

`selectPage` is still using only the `inclusionSet`

webpack/components/SelectAllCheckbox/index.js
```diff
@@ -12,64 +13,82 @@ const SelectAllCheckbox = ({
...
+  const [selectionToggle, setSelectionToggle] = useState(false);
 
-  // Checkbox states: false = unchecked, null = partially-checked, true = checked
-  // Flow: All are selected -> click -> none are selected
-  // Some are selected -> click -> none are selected
-  // None are selected -> click -> page is selected
-  const onSelectAllCheckboxChange = () => {
-    if (isSelectAllChecked === false) {
-      return selectPage();
+  const onSelectAllCheckboxChange = (checked) => {
```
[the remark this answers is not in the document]
parthaa: I 'll add em.. That may be useful for future debugging.

webpack/components/SelectAllCheckbox/index.js
```diff
@@ -12,64 +13,82 @@ const SelectAllCheckbox = ({
...
-  //   selectAll();
-  // };
+  const handleSelectAll = () => {
+    setSelectAllDropdownOpen(false);
+    setSelectionToggle(true);
+    setSelectAllChecked(true);
+  };
   const handleSelectPage = () => {
     setSelectAllDropdownOpen(false);
+    setSelectionToggle(true);
```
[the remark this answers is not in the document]
parthaa: uh ??

webpack/components/extensions/HostDetails/Tabs/ErrataTab.js
parthaa: Agreed. I think once we update the Errata page with your Rex work, we can try to be more consistent.

webpack/components/extensions/HostDetails/Tabs/TracesTab.js
parthaa: I can try adding here. Shouldn't be hard. But isn't this PR large already? 
That being said I can add it to this PR if you prefer.

app/controllers/katello/concerns/api/v2/bulk_extensions.rb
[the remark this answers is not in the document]
jeremylenz: It's good that it sends the request correctly, but I think it's worth it to add a line in the backend to handle it. 
 Either an error or ignore the search if there are included IDs.

webpack/components/SelectAllCheckbox/index.js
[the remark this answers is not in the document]
jeremylenz: `checkedState = true` means that the box is checked.  `null` means it shows partially checked.

If I select page, but not all items, the box shouldn't be checked (`true`).  It should be partially checked (`null`).

(unless there's only one page of results, in which case it should be checked (`true`)).

webpack/components/Table/TableHooks.js
jeremylenz: Gotcha.  Yeah, per my previous comment I think it's a good idea to rename that `selectAllMode` then.

webpack/components/extensions/HostDetails/Tabs/TracesTab.js
jeremylenz: I would only want it if it's super easy!  Let's punt this to the next PR for now

webpack/components/SelectAllCheckbox/index.js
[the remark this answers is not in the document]
jeremylenz: Perhaps it's clearer from a user perspective:

I have per_page set to 10 and there are 20 items total.  I do 'Select page (10)'  (or I manually select all 10 items on the first page).

Expected: The box is partially checked (the minus sign thingy)
Actual: The box is fully checked

webpack/components/SelectAllCheckbox/index.js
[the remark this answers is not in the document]
jeremylenz: I'm :-1: on this idea. We shouldn't need a "none mode" or a "page mode."  Having a "mode" for selectAll is complex enough.

also, this ties the logic of the hook to the DOM markup of the component, and makes assumptions that may or may not be true about the ID attributes that may be used in future components.  Hooks shouldn't be coupled to their components (that's kinda the point ;)

I don't think we need to keep track of if the user has clicked on all, page, or none.  We can easily compute the value of that state by looking at the selected items.

webpack/components/SelectAllCheckbox/index.js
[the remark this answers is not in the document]
parthaa: Oh are you saying `selectionState` being `page` should not matter at all ?  Only If everything across ALL pages is selected should there be `+` ?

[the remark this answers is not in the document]
jeremylenz: Righto.

webpack/components/Table/TableHooks.js
```diff
@@ -41,14 +42,143 @@ class ReactConnectedSet extends Set {
       this.delete(id);
     }
   }
+
+  addAll(ids) {
```
jeremylenz, line 46: Should also update the constructor to use `addAll` :)

parthaa: Yeh this sped up selectPage by a big margin.

webpack/components/SelectAllCheckbox/index.js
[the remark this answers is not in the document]
parthaa: I think `npx eslint webpack` didn't like the `push(<...>)`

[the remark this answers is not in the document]
jeremylenz: You can fix it... show me the error :laughing:

webpack/components/SelectAllCheckbox/index.js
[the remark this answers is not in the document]
parthaa: OK updated :) . No `selectionState` anymore

webpack/components/SelectAllCheckbox/index.js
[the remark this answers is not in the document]
parthaa: updated

webpack/components/SelectAllCheckbox/index.js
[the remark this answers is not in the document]
parthaa:
```bash
$ npx eslint webpack --fix

/home/vagrant/katello/webpack/components/SelectAllCheckbox/index.js
  79:33  error  Expected indentation of 4 space characters but found 32  react/jsx-indent

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.

```

webpack/components/SelectAllCheckbox/index.js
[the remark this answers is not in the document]
parthaa:
```bash
$ npx eslint webpack --fix

/home/vagrant/katello/webpack/components/SelectAllCheckbox/index.js
  79:33  error  Expected indentation of 4 space characters but found 32  react/jsx-indent

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.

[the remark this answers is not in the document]
Comment (jeremylenz) [reply webpack/components/SelectAllCheckbox/index.js:77]: I see what you mean.. there were two eslint rules battling.  Putting parenthesis around the JSX expression (and fixing the indentation) fixed it for me:
```js
    selectAllDropdownItems.push((
      <DropdownItem key="select-all" id="all" component="button" isDisabled={isSelectAllChecked} onClick={handleSelectAll}>
        {`${__('Select all')} (${totalCount})`}
      </DropdownItem>));
```

webpack/components/Table/TableHooks.js
```diff
@@ -41,14 +42,151 @@ class ReactConnectedSet extends Set {
...
   // needed because mutating a Ref won't cause React to rerender
   const forceRender = () => setToggle(prev => prev + 1);
   const set = useRef(new ReactConnectedSet(initialArry, forceRender));
   return set.current;
 };
 
-export default useSet;
+export const useSelectionSet = (results, metadata, initialArry = [], idColumn = 'id') => {
+  const selectionSet = useSet(initialArry);
+  const ids = results?.map(result => result[idColumn]) ?? [];
```
jeremylenz, line 62: Seems this is a list of ids on the current page, so should rename something like `pageIds`

webpack/components/Table/TableHooks.js
```diff
@@ -41,14 +42,151 @@ class ReactConnectedSet extends Set {
...
+    selectionSet,
+  };
+};
+
+export const useBulkSelect = (results, metadata, initialArry = [], idColumn = 'id') => {
+  const { selectionSet: inclusionSet, ...selectOptions } =
+                useSelectionSet(results, metadata, initialArry, idColumn);
+  const exclusionSet = useSet([]);
+  const [searchQuery, updateSearchQuery] = useState('');
+  const [isSelectAllMode, setSelectAll] = useState(false);
```
jeremylenz, line 100: The naming is closer but let's standardize it a bit :)
```suggestion
  const [selectAllMode, setSelectAllMode] = useState(false);
```

webpack/components/Table/TableHooks.js
```diff
@@ -41,14 +42,151 @@ class ReactConnectedSet extends Set {
...
+  const clearSelections = () => {
+    setSelectAll(false);
+    inclusionSet.clear();
+    exclusionSet.clear();
+  };
+
+  return {
+    ...selectOptions,
+    selectPage,
+    selectNone,
```
jeremylenz, line 179: This should also expose a `selectAll` function.  No need to delegate or expose `selectAllMode` to the component-- just have the component call `selectAll` and handle it here.

webpack/components/Table/TableHooks.js
```diff
@@ -41,14 +42,151 @@ class ReactConnectedSet extends Set {
...
+    isSelected,
+    selectedCount,
+    isSelectAllMode,
+    setSelectAllMode,
```
jeremylenz, line 183:
```suggestion
```
see comment above, let's hide this ugliness ;)

jeremylenz (commented):
Changes look good, I resolved the comments I think are resolved and left some open.

Tests to think about:
#### Dropdown actions
* Select page: all table rows are selected; checkbox is partially checked
* Select page (but there's only a single page): all table rows selected; checkbox fully checked
* Select none: no table rows selected

#### Table row actions
* Check boxes across multiple pages: selected items are preserved when switching pages
* Start with 100 items, do a search which returns 5, Select All: 5 should be selected
* Do the above, but clear the search: the selection should be cleared
* Select All, then deselect a single item: box should be partially checked
* Select All, then manually deselect all items: box should be unchecked

webpack/components/Table/TableHooks.js
parthaa: well I need the `isSelectAllMode` still for things like
```
  if (isSelectAllMode !== null) {
    selectAllDropdownItems.push(
      ( <DropdownItem key="select-all" component="button" isDisabled={isSelectAllMode} onClick={handleSelectAll}>
         {`${__('Select all')} (${totalCount})`}
        </DropdownItem>
      )
    );
  }

```

jeremylenz: From my reading it looks like you're using this check in `selectAllCheckbox` to determine if the `useBulkSelect` hook is in use, or just the regular `useSelectionSet`.

Seems simpler to me just to add a `showSelectAll` prop to `selectAllCheckbox` and use that.  I don't think `selectAllCheckbox` should need to know about `selectAllMode`. In fact, the only code that should know about it is TableHooks.js.

jeremylenz: Don't put anything in internal state which can be computed ;)

```js
isDisabled={areAllRowsSelected}
```

PR #9654, opened 2021-09-20
Fixes #33517 - add force option to metadata regeneration

ianballou:
Removes the metadata generation options for repositories and content view versions from the UI.  It is currently dangerous with Pulp 3 & repositories that have mirror on sync set to true.

To test:

1) Try the `/katello/api/repositories/:id/republish`and `/katello/api/content_view_versions/:id/republish_repositories` endpoints
  -> Ensure both do not publish without the force option being set to `true`
2) Try republishing content view version metadata via Hammer
  -> Note that there isn't an option for republishing single repo metadata in Hammer. Only for content view versions.
  -> The force option should be necessary there too
  -> Check that the warning messages look fine
3) Check that there is no way to republish metadata (repository or content view version) via the UI

### DIFF
--- a/app/controllers/katello/api/v2/repositories_controller.rb
+++ b/app/controllers/katello/api/v2/repositories_controller.rb
@@ -264,8 +264,12 @@ module Katello
 
     api :PUT, "/repositories/:id/republish", N_("Forces a republish of the specified repository, regenerating metadata and symlinks on the filesystem.")
     param :id, :number, :desc => N_("Repository identifier"), :required => true
+    param :force, :bool, :desc => N_("Force metadata regeneration to proceed.  Dangerous when repositories use mirror on sync."), :required => true
     def republish
-      task = async_task(::Actions::Katello::Repository::MetadataGenerate, @repository, :force => true)
+      unless ::Foreman::Cast.to_bool(params[:force])
+        fail HttpErrors::BadRequest, _('Metadata republishing must be forced because it is a dangerous operation.')
+      end
+      task = async_task(::Actions::Katello::Repository::MetadataGenerate, @repository)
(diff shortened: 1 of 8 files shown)


app/controllers/katello/api/v2/repositories_controller.rb
```diff
@@ -264,8 +264,14 @@ def repository_types
 
     api :PUT, "/repositories/:id/republish", N_("Forces a republish of the specified repository, regenerating metadata and symlinks on the filesystem.")
     param :id, :number, :desc => N_("Repository identifier"), :required => true
+    param :force, :bool, :desc => N_("Force metadata regeneration to proceed.  Dangerous when repositories use mirror on sync."), :required => true
     def republish
-      task = async_task(::Actions::Katello::Repository::MetadataGenerate, @repository, :force => true)
+      unless ::Foreman::Cast.to_bool(params[:force])
+        render_error('custom_error', :status => :unprocessable_entity,
```
jturel, line 270: The pattern I've seen for raising controller errors has been more like `fail HttpErrors::BadRequest("why u do this?")`. Was there an advantage to going this route 'cause I think raising an error like BadRequest ultimately does what you've done here automatically!

ianballou: No reason, that looks better to me.

ianballou:
[test katello]

jturel (approved):
It seems like it accomplishes the intended purpose :)

APJ

ianballou:
@jturel I just have to fix up the tests.

ianballou:
[test katello]

ianballou:
[test katello]

PR #9661, opened 2021-09-22
Fixes #33518 - do not regenerate metadata on force sync

jlsherrill:
for mirrored repos

* Updates capsule syncing to never regenerate metadata for yum repos
  * Also cleans and simplifies the code by using a lookup to find the
    publication href, instead of passing it around
* Updates repo sync to never regenerate metadata on a force sync

### DIFF
--- a/app/lib/actions/katello/repository/finish_upload.rb
+++ b/app/lib/actions/katello/repository/finish_upload.rb
@@ -13,7 +13,7 @@ module Actions
             unit_type_id = SmartProxy.pulp_primary.content_service(content_type)::CONTENT_TYPE
           end
           generate_metadata = options.fetch(:generate_metadata, true)
-          plan_action(Katello::Repository::MetadataGenerate, repository, :dependency => import_upload_task) if generate_metadata
+          plan_action(Katello::Repository::MetadataGenerate, repository, :dependency => import_upload_task, :force_publication => true) if generate_metadata
 
           recent_range = 5.minutes.ago.utc.iso8601
           plan_action(Katello::Repository::FilteredIndexContent,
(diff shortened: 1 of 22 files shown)


jlsherrill:
Some things i tested:

- Sync a library yum repo with mirror on sync turned on via normal sync and force sync, notice the timestamp in the repomd.xml file in the repo does not change
- Sync a library yum repo with mirror on sync turned off via normal sync and force sync, the timestamp should change on every force sync, but not a normal sync
- sync file/docker/yum repos to a capsule, notice that no matter how many syncs or force syncs you do, the timestamp in the repomd.xml file does not change on the capsule yum repo and matches the satellite's repomd.xml file's timestamp

When i say timestamp above, i'm referring to the '<revision>1632317374</revision>'  field in the ./repodata/repomd.xml file

ianballou:
[test katello]

ianballou:
I found an issue with a missing publication:

1) After creating a repo and syncing it both with mirror on and off, I added it to a content view with some filters
2) I published, promoted, and synced the environment to a smart proxy
3) Next, I set the repository to use mirroring again
4) After changing up the filters and republishing + promoting, I noticed that the proxy was now getting 404s on syncing and the distribution was missing a publication.
5) The distribution + publication for the Library repo is fine, but not the content view repos

jlsherrill:
A couple of notes on the latest updates:

1) some actions like uploading content or removing content must force a metadata generation, thus i added a :force_publication option to be passed in.  In the future we should forbid uploading/removing content for 'mirrored' repositories.  Then this would not be needed
2) I changed the check about whether to create a publication to check if force_publication is passed or the current publication_href is nil.  this was metadata generates after repo creation will get executed across the app and tests

jlsherrill:
and this is ready for testing again :)

ianballou (approved):
This is working fine for me!

PR #9663, opened 2021-09-22
Fixes #33534 - registry name pattern update fails

ianballou:
This PR fixes a couple issues:
1) Syncing a smart proxy doesn't update distribution paths
2) `repos_to_sync` returns nil because the smart proxy sync histories are not cleared out after changing the registry name pattern

To test:
1) Sync some docker repos and stick them in a content view.
2) Publish that view and ensure it's promoted to some LCE
3) Sync that LCE to a smart proxy
4) Check the container distributions via the Pulp CLI on both the Katello server and the smart proxy and note the base paths
5) Change the registry name pattern
6) After the tasks finish, check all of the distributions on both the Katello server and the smart proxy to ensure that the base paths changed
7) Also check that there are no failed tasks during the testing

TODO:
- [x] Add tests

### DIFF
--- a/app/services/katello/pulp3/repository.rb
+++ b/app/services/katello/pulp3/repository.rb
@@ -205,7 +205,7 @@ module Katello
       def distribution_needs_update?
         if distribution_reference
           expected = secure_distribution_options(relative_path).except(:name).compact
-          actual = get_distribution.to_hash
+          actual = get_distribution&.to_hash || {}
           expected != actual.slice(*expected.keys)
         elsif repo.environment
           true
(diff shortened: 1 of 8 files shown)


ianballou:
[test katello]

app/services/katello/pulp3/repository.rb
```diff
@@ -205,7 +205,9 @@ def get_distribution(href = distribution_reference.href)
       def distribution_needs_update?
         if distribution_reference
           expected = secure_distribution_options(relative_path).except(:name).compact
-          actual = get_distribution.to_hash
+          actual = {}
+          dist = get_distribution
+          actual = dist.to_hash unless dist.nil?
```
jlsherrill, line 210: could you do:
```
actual = get_distribution&.to_hash || nil 
```

app/services/katello/pulp3/repository_mirror.rb
```diff
@@ -169,7 +169,7 @@ def refresh_distributions(options = {})
         if (distro = repo_service.lookup_distributions(base_path: path).first) ||
           (distro = repo_service.lookup_distributions(name: "#{backend_object_name}").first)
           # update dist
-          dist_options = dist_options.except(:name, :base_path)
+          dist_options = dist_options.except(:name)
```
jlsherrill, line 172: could you explain why this change is needed?

ianballou: Without it, the `base_path` was not getting updated on the smart proxies.

app/services/katello/pulp3/repository.rb
ianballou: Actually, the reason I didn't do this was that `actual` being `nil` would break the call to `slice`.

app/services/katello/pulp3/repository_mirror.rb
jlsherrill: ohhh yes, that makes sense.  I think i was reading it backwards

app/services/katello/pulp3/repository.rb
jlsherrill: oh, then why not?  
```
actual = get_distribution&.to_hash || {}
```

ianballou: Yeah that works :)

jlsherrill (approved):

PR #9668, opened 2021-09-24
Fixes #33561 - drop Puppet Environment migration

ezr-ondrej:
Foreman core wants to remove the Environments migrations and Katello useless migration is blocking it.
We can safely remove the migration and let ForemanPuppet remove the column.

ForemanPuppet is dealing with this in https://github.com/theforeman/foreman_puppet/pull/200

This is blocking https://github.com/theforeman/foreman/pull/8790

### DIFF
--- a/db/migrate/20140110000001_update_environments_add_katello_id.rb
+++ b/db/migrate/20140110000001_update_environments_add_katello_id.rb
@@ -1,10 +1,8 @@
 class UpdateEnvironmentsAddKatelloId < ActiveRecord::Migration[4.2]
   def up
-    add_column :environments, :katello_id, :string, :limit => 255
-    add_index :environments, :katello_id
+    # not doing anything, it is not needed anymore in supported versions
   end
 
   def down
-    remove_column :environments, :katello_id
   end
 end


db/migrate/20140110000001_update_environments_add_katello_id.rb
```diff
@@ -1,10 +0,0 @@
-class UpdateEnvironmentsAddKatelloId < ActiveRecord::Migration[4.2]
-  def up
-    add_column :environments, :katello_id, :string, :limit => 255
-    add_index :environments, :katello_id
-  end
-
-  def down
-    remove_column :environments, :katello_id
-  end
-end
```
jlsherrill, line 10: does it make more sense to leave the migration file as is but change was it does?

jlsherrill:
the test failures don't make a ton of sense, will re-test it once the current barrage of tests are finished.

jlsherrill:
[test katello]

db/migrate/20140110000001_update_environments_add_katello_id.rb
ezr-ondrej: You tell me :)
I can probably just have a check there `if table_exists?` that way we wouldn't need to worry about keeping the migration version is `schema_migrations` table.

jlsherrill: I think it does, i'm not sure of the consequences of deleting a migration that Rails thinks was migrated.  And yeah a check of table_exists? i think is how we handled this stuff in the past

ezr-ondrej: I went with doing nothing, as we don't want the `katello_id` on environments.
I have https://github.com/theforeman/foreman_puppet/pull/200 to remove it for users who continue using puppet.

PR #9671, opened 2021-09-27
Fixes #33525 - Add debs packages_restrict_latest

m-bucher:
This should implement the `packages_restrict_latest` parameter, but I am not sure I really want to have this merged, because it does not scale well at all.
This originates in the way we compare debian versions and the fact that we have to compare them and are not able to simply sort them.

Maybe some kind of restriction to a max number of packages is in order for this to be usable.

### DIFF
--- a/app/models/katello/deb.rb
+++ b/app/models/katello/deb.rb
@@ -103,8 +103,13 @@ module Katello
       Katello::Deb.in_repositories(repos).where.not(name: host.installed_debs.pluck(:name)).order(:name)
     end
 
-    def self.latest(_relation)
-      fail 'NotImplemented'
+    def self.latest(relation)
+      # This might be very slow
+      return relation.joins(
+        "LEFT OUTER JOIN(#{relation.to_sql}) AS katello_debs2 ON " \
+        'katello_debs.name = katello_debs2.name AND katello_debs.architecture = katello_debs2.architecture AND ' \
+        'deb_version_cmp(katello_debs.version, katello_debs2.version) < 0 ' \
+      ).where('katello_debs2.id IS NULL')
     end
   end
 end
--- a/test/controllers/api/v2/debs_controller_test.rb
+++ b/test/controllers/api/v2/debs_controller_test.rb
@@ -7,6 +7,7 @@ module Katello
       @version = ContentViewVersion.first
       @deb = katello_debs(:one)
       @host = hosts(:one)
+      @org = get_organization
     end
 
     def setup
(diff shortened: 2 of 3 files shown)


sbernhard:
Can you add some benchmark values here? https://ruby-doc.org/stdlib-2.5.3/libdoc/benchmark/rdoc/Benchmark.html

How can we get this fast?

app/models/katello/deb.rb
```diff
@@ -103,8 +103,13 @@ def self.apt_installable_for_host(host)
...
 
-    def self.latest(_relation)
-      fail 'NotImplemented'
+    def self.latest(relation)
+      # This might be very slow
+      return relation.joins(
+        "LEFT OUTER JOIN(#{relation.to_sql}) AS katello_debs2 ON " \
+        'katello_debs.name = katello_debs2.name AND katello_debs.architecture = katello_debs2.architecture AND ' \
+        'deb_version_cmp(katello_debs.version, katello_debs2.version) < 0 ' \
+      ).where('katello_debs2.id IS NULL')
```
adamruzicka, line 112: Capturing things we talked about live: If I were doing this I'd probably drop the explicit return and used a squiggly heredoc, but that's just personal preference so feel free to ignore.

adamruzicka (approved):
Ack from the logic point of view

app/models/katello/deb.rb
m-bucher: Since most of the code was shamelessly copied over from app/models/katello/rpm.rb, I would keep it that way for being consistent.

Eventually, we should think about enforcing a certain style through rubocop. AFAIK Katello's rubocop rules are currently pretty flexible.

qcjames53:
Hi @m-bucher! Is this PR still relevant to Katello 4.17+? We're wondering if you'd like us to close this out.

app/models/katello/deb.rb
```diff
@@ -103,8 +103,13 @@ def self.apt_installable_for_host(host)
       Katello::Deb.in_repositories(repos).where.not(name: host.installed_debs.pluck(:name)).order(:name)
     end
 
-    def self.latest(_relation)
-      fail 'NotImplemented'
+    def self.latest(relation)
+      # This might be very slow
```
ianballou, line 107: Some LLM fun to try avoiding the left outer join:

```
  def self.latest(relation)
    # Create a CTE to avoid double subquery execution
    sql = "WITH base_data AS (#{relation.to_sql})
           SELECT bd1.* FROM base_data bd1
           WHERE NOT EXISTS (
             SELECT 1 FROM base_data bd2 
             WHERE bd1.name = bd2.name 
             AND bd1.architecture = bd2.architecture 
             AND deb_version_cmp(bd1.version, bd2.version) < 0
           )"
    relation.from("(#{sql}) optimized_latest")
  end
```

ianballou:
Closing due to inactivity.

PR #9672, opened 2021-09-27
Fixes #33571 - Add a PR template

ianballou:
Sample:

What is the context for this pull request?

This PR adds a PR template file to remind contributors to add context and test steps to PRs.

What are the testing steps for this pull request?

N/A

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,3 @@
+### What are the changes introduced in this pull request?
+
+### What are the testing steps for this pull request?


ianballou:
I'm going back and forth on if the template lines should be formatted to stand out.  Perhaps bold, or larger font.  Doesn't seem completely necessary though,,,

rverdile (commented):
I think it makes sense to add something to make the questions stand out, like you mentioned.

jeremylenz:
+1 to @rverdile above, I'd recommend prepending `### `

Do you think it would be helpful to add "What are the changes introduced in this pull request?" as the first question, since we mostly include that anyway?

rverdile (commented):
Would "context" here be like a narrative description of the problem/solution for the PR? Maybe that should be slightly more specific.

ianballou:
I updated the template to ask for an explanation of the changes.  I think that's better than asking for context, which should hopefully be in the redmine and/or BZ anyway.

jeremylenz (approved):
ACK :+1:

rverdile (approved):
Looks good to me too :)

PR #9673, opened 2021-09-27
Fixes #33572 - Refreshes the host statuses on 'built'

parthaa:
During a global registration
1) System is registered to host
2) The build mode is set to true
3) Another script runs to set the build mode to false if nothing fails

This code tracks that time where the hosts' build mode goes from false
-> true and refreshes the status appropriately.

### DIFF
--- a/app/models/katello/concerns/host_managed_extensions.rb
+++ b/app/models/katello/concerns/host_managed_extensions.rb
@@ -56,6 +56,9 @@ module Katello
         after_validation :queue_reset_content_host_status
         register_rebuild(:queue_reset_content_host_status, N_("Content_Host_Status"))
 
+        after_validation :queue_refresh_content_host_status
+        register_rebuild(:queue_refresh_content_host_status, N_("Refresh_Content_Host_Status"))
+
         scope :with_pools_expiring_in_days, ->(days) { joins(:pools).merge(Katello::Pool.expiring_in_days(days)).distinct }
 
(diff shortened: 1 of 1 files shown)


app/services/katello/registration_manager.rb
```diff
@@ -223,9 +223,15 @@ def finalize_registration(host)
...
             consumer_attributes.except(:guestIds, :facts))
         host.subscription_facet.save!
-        host.subscription_facet.update_subscription_status
-        host.content_facet.update_errata_status
-        host.refresh_global_status!
+
+        # we update the status only if build mode is not on.
+        # If it is on we refresh the statuses automatically elsewhere
+        # when build mode is changed from true to false.
+        unless host.build?
```
jturel, line 230: Ooops, this was misguided. We actually never seem to hit this block because we go into build mode *after* subman-register is called :(

the script has:

```
subscription-manager register
register_katello_host
```

Sorry :(

jturel:
Left one comment otherwise it works well!

app/services/katello/registration_manager.rb
jturel: Actually, I guess it would work in the non global-reg workflow of setting host to build mode -> restarting to begin the build -> registration happens (skipping status calls) -> build completes and triggers your new method

parthaa: updated :)

jturel (approved):
ACK

Nice :)

PR #9674, opened 2021-09-27
Fixes #33573 - catch RestClient::NotFound in UpstreamConsumer

lfu:
QE ran into a 404 not found error when using cloned manifest.

**Test Steps**
In web UI, navigate Administer > Organization and select an organization's name.
 Error was displayed:
```
 Oops, we're sorry but something went wrong 404 Not Found
```

### DIFF
--- a/app/lib/katello/resources/candlepin/upstream_consumer.rb
+++ b/app/lib/katello/resources/candlepin/upstream_consumer.rb
@@ -13,6 +13,8 @@ module Katello
             resource.head
           rescue RestClient::Unauthorized, RestClient::Gone
             raise ::Katello::Errors::UpstreamConsumerGone
+          rescue RestClient::NotFound
+            raise ::Katello::Errors::UpstreamConsumerNotFound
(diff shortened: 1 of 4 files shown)


app/lib/katello/resources/candlepin/upstream_consumer.rb
```diff
@@ -22,6 +22,8 @@ def get(params)
             JSON.parse(super(path + hash_to_query(params) + includes, self.default_headers).body)
           rescue RestClient::Gone
             raise ::Katello::Errors::UpstreamConsumerGone
+          rescue RestClient::NotFound
```
jturel, line 25: This should actually be handled in the `ping` method of this class. See how `UpstreamConnectionChecker` uses it

Also, is there any test coverage to be updated?

jturel (changes requested):

jturel (approved):
Works well! APJ

Thanks @lfu

PR #9675, opened 2021-09-27
Fixes #33077 - Enable Tracer via remote execution

jeremylenz:
- [x] via remote execution
- [x] via customized remote execution
- [x] tests

This adds "Enable Tracer" empty state and modal to the Traces tab.
* allows you to enable Tracer via remote execution
* allows you to enable Tracer via customized remote execution
* Adds reusable RemoteExecution Redux actions

#### Get a working REX setup
1. In Infrastructure > Smart Proxies, verify that the 'SSH' feature is listed
2. If needed, run `foreman-installer --enable-foreman-proxy-plugin-remote-execution-ssh --disable-system-checks` on the server
1. In Settings > Authentication, turn off "Require SSL for smart proxies"
2. Spin up a `centos7-katello-client` and make sure it's registered
3. On the smart proxy / server: `cat ~foreman-proxy/.ssh/id_rsa_foreman_proxy.pub`
4. Copy the output to the clipboard
5. On the client, as root, paste the contents of the file into `~/.ssh/authorized_keys`
6. Confirm that it's working: On the smart proxy / server, `ssh -i ~foreman-proxy/.ssh/id_rsa_foreman_proxy root@target.example.com`
7. Run `yum update` on the client and you'll probably have some Traces (optional)

#### Test with Tracer installed
1. Hosts > All Hosts > (your client) > Edit dropdown > New details page > Traces tab
2. Verify that the Traces table loads normally

#### Test without Tracer installed
1. Hosts > All Hosts > (your client) > Edit dropdown > New details page > Traces tab
2. Verify that the "Enable Tracer" empty state shows
3. Click the button and try "via remote execution" > Enable Tracer
4. View the task results and then return to the Traces tab: the table should load the traces normally
5. On the client, `yum -y remote katello-host-tools-tracer` (OR you could just use REX ;)
6. Refresh the page and you should see the Enable Tracer empty state again
7. This time try the "via customized remote execution" option - you should be taken to the New Job Invocations page with the hostname and package name filled in


jeremylenz:
@jturel you were supposed to tell me what not to do re: REX, hope I didn't already do some of those things :grimacing: 

I did use the remote execution controller directly and not the Katello one

jeremylenz:
[test katello]

webpack/components/extensions/HostDetails/Tabs/HostTracesActions.js
```diff
@@ -24,3 +25,14 @@ export const resolveHostTraces = (hostId, params) => put({
   errorToast: error => errorToast(error),
   params,
 });
+
+export const getTracerStatus = hostId => get({
+  type: API_OPERATIONS.GET,
+  key: KATELLO_HOST_TOOLS_TRACER_KEY,
+  url: foremanApi.getApiUrl(`/hosts/${hostId}/packages?search=name=katello-host-tools-tracer`),
```
jturel, line 32: Probably good to make the package name a constant?

webpack/components/extensions/HostDetails/Tabs/RemoteExecutionActions.js
```diff
@@ -0,0 +1,35 @@
...
+};
+
+const jobInvocationParams = ({ hostname, packageName }) => ({
+  job_invocation: {
+    feature: 'katello_package_install',
+    targeting_type: 'Static Query',
+    inputs: {
+      package: packageName,
+    },
+    search_query: `name ^ (${hostname})`,
```
jturel, line 19: Searching by the host id might be faster / more precise?

webpack/components/extensions/HostDetails/Tabs/RemoteExecutionActions.js
```diff
@@ -0,0 +1,35 @@
...
+
+const errorToast = (error) => {
+  const message = getResponseErrorMsgs(error.response);
+  return message;
+};
+
+const jobInvocationParams = ({ hostname, packageName }) => ({
+  job_invocation: {
+    feature: 'katello_package_install',
+    targeting_type: 'Static Query',
```
jturel, line 15: Does targeting_type need to be required? Nevertheless I'm not sure 'Static Query' is a valid option - I thought it was something like 'static' ? Could be wrong but want to confirm

webpack/components/extensions/HostDetails/Tabs/RemoteExecutionActions.js
jeremylenz: When you select "via customized remote execution" on the old page, this is the search that's filled out for you.  I feel like it's better for the user to see the hostname there as opposed to, say, `id = 12247`.  So that's why I did it this way.  For the immediate action here ("via remote execution") I suppose that's not as important, but it would at least be consistent.

jturel: i didn't realize the user sees this. all good then

webpack/components/extensions/HostDetails/Tabs/RemoteExecutionActions.js
jeremylenz: I removed it and it still works fine, so guess not! :+1:

webpack/components/extensions/HostDetails/Tabs/EnableTracerModal.js
```diff
@@ -0,0 +1,126 @@
...
+          {title}
+        </Button>
+      );
+    }
+    return (
+      <Button
+        key="enable_button"
+        component="a"
+        isLoading={redirect}
+        onClick={() => setRedirect(true)}
```
sjha4, line 75: Don't know enough to test this but what is `redirect` doing here other than the isLoading bit?

sjha4: oh nvm..It's a href button so I am assuming it acts as a link..

jeremylenz: I was just using it to put the spinner on the button when you click it.  It was badly named.  Updated the name to `setButtonLoading`. :+1:

jturel:
Looks really good! Is it easy to get the page to reload when the package install completes in order to avoid manually refreshing to see the Traces table?

jeremylenz:
>  Is it easy to get the page to reload when the package install completes in order to avoid manually refreshing to see the Traces table?

@jturel It requires polling the task; I wouldn't say it's "easy".. :confused: 

@sjha4 thoughts?

jturel:
Thanks for the details Samir - I think we can handle that later and focus on other core functionality until then!

jturel (approved):
ACK!

Nice :)

PR #9676, opened 2021-09-28
Fixes #33578 - add download_policy for deb-repos

m-bucher:
Since pulpcore supports it for all, we could actually enable `download_policy` for all repo types :thinking:


quba42:
I should just leave a slight caveat: While I am confident the Pulp feature works for deb content, It does not make sense to use it for upstream repos that frequently purge old package versions (since they can then no longer be downloaded "on demand").

I am not 100% sure to what extent official Debian/Ubuntu repositories are affected by this, but I imagine it is true at least for security repos. IMHO it should nevertheless be a user choice to sync stable upstream repos "on demand".

m-bucher:
[test katello]

jjeffers:
@m-bucher Looks like there are some test failures due to VCR conflicts. Can you check on that and re-push?

jturel:
[test katello]

jturel:
not sure if test failures are related...


[test katello]

m-bucher:
> @m-bucher do you need one of the admins to merge this?

I am not allowed to do it, therefore I guess so :laughing:

PR #9678, opened 2021-09-29
Fixes #33581 - Correctly sets the host attributes on Edit Host

parthaa:
Prior to this commit `Edit Host` would give a preference to the values
of the associated hostgroup (this was only in display.)

This meant if you created a host with a hostgroup and changed things
like LCE/CV/kickstart repository id etc those overrides
would not correctly get reflected when Editing the host. It would only
show what was in the hostgroup attached to the host.

This fix magically solves that issue by making it corrently look at
the host and hostgroup.

### DIFF
--- a/app/models/katello/concerns/host_managed_extensions.rb
+++ b/app/models/katello/concerns/host_managed_extensions.rb
@@ -33,6 +33,8 @@ module Katello
       included do
         prepend Overrides
 
+        delegate :content_source_id, :content_view_id, :lifecycle_environment_id, :kickstart_repository_id, to: :content_facet, allow_nil: true
+
         has_many :dispatch_histories, :class_name => "::Katello::Agent::DispatchHistory", :foreign_key => :host_id, :dependent => :delete_all
 
         has_many :host_installed_packages, :class_name => "::Katello::HostInstalledPackage", :foreign_key => :host_id, :dependent => :delete_all


parthaa:
To test:
1) Sync https://partha.fedorapeople.org/test-repos/kickstart-zoo/ or sync any kickstart tree
2) Create 2 content views cv1 and cv2
3) Add the repo to both cv's and publish
4) Create a hostgroup hg1 with CV1 and ks repo selected
5) Go to `Infrastructure -> Domain` in the UI and create a new domain `foo.com`
6) `Hosts -> Create Host`
7) Select the host group to be hg1
8) Change the Content VIew to cv2
9) Go to Interface and add a fake mac address like `00:11:22:33:44:55` and use the `foo.com` domain (all this to get the host creation running.)
10) Finally submit (you might be required to select partition table + provide a 8 letter password - do whatever is necessary,)
11) You should see the host provisioning
12) Now Edit this host in the UI

Before PR 
You 'd see cv1 as selected (side stepping the cv2  you selected when creating the host.)

After PR
You should see cv2 correctly selected. 

Try playing with other attributes etc.

lfu:
LGTM 👍🏻 

**Before**
Screenshot from 2021-09-29 14-38-46

**After**
Screenshot from 2021-09-29 14-55-34

jturel (approved):
LGTM :)

PR #9680, opened 2021-09-30
Fixes #33598 - Correctly detect Candlepin async hypervisors failures

adamruzicka:


adamruzicka:
It would probably be even better to do the error checking in `#done?`. Is there a reason why it is being done as part of the poll in `Candlepin::AbstractAsyncTask`?

jeremylenz:
>  Is there a reason why it is being done as part of the poll in Candlepin::AbstractAsyncTask?

I doubt it; I think this error checking was [added](https://github.com/Katello/katello/pull/9378) after we started using the async Candlepin API for manifest imports.  If `done?` is a more correct way to do it then by all means :)

PR #9687, opened 2021-10-01
Fixes #33081 - Restart traces via remote execution

jeremylenz:
Rebased on https://github.com/Katello/katello/pull/9675

- [ ] Bulk restart via remote execution
- [ ] Bulk restart via customized REX
- [ ] Single trace restart via REX
- [ ] Single trace restart via customized REX
- [ ] tests -  restart via remote execution
- [ ] tests - restart via customized REX
- [ ] tests - Single trace restart via REX
- [ ] tests - Single trace restart via customized REX


webpack/components/extensions/HostDetails/Tabs/__tests__/tracesTab.test.js
```diff
@@ -1,127 +1,296 @@
...
+    const viaCustomizedRex = queryByText('via customized remote execution');
+    expect(viaCustomizedRex).toBeVisible();
+    viaCustomizedRex.click();
+    expect(queryByText('via remote execution')).not.toBeInTheDocument();
+
+    const enableTracesModalLink = getByRole('link', { name: 'Enable Tracer' });
+    expect(enableTracesModalLink)
+      .toHaveAttribute(
+        'href',
+        '/job_invocations/new?feature=katello_package_install&inputs%5Bpackage%5D=katello-host-tools-tracer&host_ids=name%20%5E%20(client.example.com)',
```
jeremylenz, line 291: @MariaAga For the new job invocation wizard, will we have / do we already have the capability to pass in job details via url like this?

MariaAga: Yeah, should be made available in this pr: https://github.com/theforeman/foreman_remote_execution/pull/629

jeremylenz:
rebased to current master

jeremylenz:
> I noticed when resolving a single trace, the /resolve api is used, but when resolving bulk, the native REX api is used. can native rex be used for both?

* A PUT to `/hosts/${hostId}/traces/resolve` is used for both single and bulk resolves ('Restart via remote execution').  Are you saying both of these should use a POST to `/job_invocations`?  I don't really feel strongly about it but I'm happy to change it if you want.

* The 'Restart via customized remote execution' actions don't contact the backend at all; rather, the `bulkCustomizedRexUrl` method is used to build the href for an `<a>` which takes you to the new job invocation page.  That's why we needed to do the helper-building logic on the front end.

jeremylenz:
side note, A nice bonus of using `/hosts/${hostId}/traces/resolve` is that the `reboot` part of the helper logic is done for you there in the controller.  (It appears the Angular page did the `session` part.)  I agree this logic should probably all be in one place.. :thinking:

PR #9688, opened 2021-10-04
Fixes #33618 - Forward the 'If-Modified-Since' header to Candlepin

hao-yu:


app/controllers/katello/api/rhsm/candlepin_proxies_controller.rb
```diff
@@ -71,9 +71,23 @@ def drop_api_namespace(original_request_path)
     end
 
     def get
-      r = Resources::Candlepin::Proxy.get(@request_path)
-      logger.debug filter_sensitive_data(r)
-      render :json => r
+      extra_headers = {}
+      modified_since = request.headers['If-Modified-Since']
```
jturel, line 75: Might be good to move 'if-modified-since' to a constant. Maybe RestClient already has one to use (although I have no issues creating our own in this file).

hao-yu:
I think I will hold this patch until I know what should be done for this Candlepin bug https://bugzilla.redhat.com/show_bug.cgi?id=2010251

hao-yu:
@jturel Additionally, it will mitigate the number of redundant calls for subscription-manager without my recent patch.

jturel:
Alright, I'm testing this out and when I test "can I get a 304 response from Katello", I can indeed.

But this does not seem to work with subscription-manager because the If-Modified-Since header it sends seem to be identical to the lastUpdated value on the accessible_content response:

```
[11] pry(main)> JSON.parse(r.body)['lastUpdate']
=> "2021-10-14T17:43:50+0000"

[12] pry(main)> r = Katello::Resources::Candlepin::Proxy.get('/consumers/55c00b95-3675-40c2-989a-6ee684423d34/accessible_content', {"If-Modified-Since"=>"Thu, 14 Oct 2021 17:43:50 GMT"})
2021-10-14T18:26:18 [D|kat|] Sending GET request to Candlepin: /consumers/55c00b95-3675-40c2-989a-6ee684423d34/accessible_content
=> <RestClient::Response 200 "{\"lastUpdat...">

[13] pry(main)> r = Katello::Resources::Candlepin::Proxy.get('/consumers/55c00b95-3675-40c2-989a-6ee684423d34/accessible_content', {"If-Modified-Since"=>"Thu, 14 Oct 2021 17:43:50 GMT"})
2021-10-14T18:26:21 [D|kat|] Sending GET request to Candlepin: /consumers/55c00b95-3675-40c2-989a-6ee684423d34/accessible_content

=> <RestClient::Response 200 "{\"lastUpdat...">
[14] pry(main)> r = Katello::Resources::Candlepin::Proxy.get('/consumers/55c00b95-3675-40c2-989a-6ee684423d34/accessible_content', {"If-Modified-Since"=>"Thu, 14 Oct 2021 17:43:50 GMT"})
2021-10-14T18:26:22 [D|kat|] Sending GET request to Candlepin: /consumers/55c00b95-3675-40c2-989a-6ee684423d34/accessible_content
=> <RestClient::Response 200 "{\"lastUpdat...">
```

So, this change definitely sends the headers, the caching is not being leveraged.

hao-yu:
I was thinking to make subscription-manager to set the "lastUpdate" + 1 second for if-modified-since but just not confident whether it is safe to do that. 

Or I think it is enough to fix this condition in the Candlepin so that it will return 304 if the "if-modified-since" is equal to the "cert.getUpdated()"

https://github.com/candlepin/candlepin/blob/candlepin-4.0-HOTFIX/server/src/main/java/org/candlepin/controller/ContentAccessManager.java#L657-L659

jturel:
[test katello]

PR #9693, opened 2021-10-04
Fixes #33629 - 4.0 upgrade migration fails due to validation

ianballou:
Issue at hand: when the migration below actually updates a root repo, there is a validation failure on `:ensure_valid_upstream_authorization` because the `upstream_authentication_token` column does not exist yet.  This PR works around the issue by stubbing out this validation.

To test:

1) Set up a Katello 3.18 instance and make a custom repository
2) Edit the repository like so in the `console`:

```ruby
::Katello::Repository.find(<id>).root.update(arch: 'x86_64')
```
3) Make a VM snapshot
4) Update Katello to 4.0 and run the database migration
5) See the migration error from the redmine
6) Revert back to Katello 3.18
7) Update Katello to 4.0 again
8) Patch in this PR
9) See the migration run successfully

### DIFF
--- a/db/migrate/20210512192745_fix_red_hat_root_repository_arch.rb
+++ b/db/migrate/20210512192745_fix_red_hat_root_repository_arch.rb
@@ -2,7 +2,7 @@ class FixRedHatRootRepositoryArch < ActiveRecord::Migration[6.0]
   def up
     ::Katello::RootRepository.
       joins("INNER JOIN katello_contents ON katello_contents.cp_content_id = katello_root_repositories.content_id").
-      where.not(arch: 'noarch').where.not("katello_contents.content_url ILIKE '%$basearch%'").update(arch: 'noarch')
+      where.not(arch: 'noarch').where.not("katello_contents.content_url ILIKE '%$basearch%'").update_all(arch: 'noarch')
   end
 
   def down


db/migrate/20210512192745_fix_red_hat_root_repository_arch.rb
```diff
@@ -1,8 +1,18 @@
...
-      joins("INNER JOIN katello_contents ON katello_contents.cp_content_id = katello_root_repositories.content_id").
-      where.not(arch: 'noarch').where.not("katello_contents.content_url ILIKE '%$basearch%'").update(arch: 'noarch')
+    ::Katello::RootRepository.class_eval do
+      def upstream_authentication_token=(_)
+        nil
+      end
+
+      def upstream_authentication_token_changed?
+        false
+      end
```
jlsherrill, line 10: one common way that we handle this is by defining a fake class, like:

https://github.com/Katello/katello/blob/ea9570ae2e2529b54f20f113abdab86b83371040/db/migrate/20180920214134_create_repository_root.rb#L46-L48

this way the class uses the same table, but has no validations/hooks

jlsherrill: the benefit is that it doesn't modify the class which could affect other migrations down the line

jlsherrill (approved):
ACK! reproduced and saw that this fixed it!

jlsherrill:
going ahead and merging as the build status seems to be having issues

PR #9689, opened 2021-10-04
Fixes #33625 - use correct remote href for generic repos

jlsherrill:

PR #9695, opened 2021-10-05
Refs #33625 - fix proxy sync and orphan cleanup

jlsherrill:
by using newer method for fetching the api class


jlsherrill:
the previous pr: https://github.com/Katello/katello/pull/9689 was merged prematurely.  This fixes the issues it caused

jjeffers:
Capsule sync appeared to get content onto the capsule (yum content, for example). The katello server contained ostree and python content, but the capsule did not support either. I did see a 404 from the katello server as a ping to `pulp` failed with a 404, which generated an error message. The capsule sync is listed as "Never synced" although the content tab shows the yum content (and the default content view).

jlsherrill:
i think you are seeing this: https://projects.theforeman.org/issues/33325  which is expected at this point.  I believe it shows never sync because one or more failed.

PR #9701, opened 2021-10-06
Fixes #33651 - Cache resource list API responses

jturel:

These changes cache the API responses for two endpoints in the RootController. My recent investigation into increased memory usage got me looking at how many objects our APIs are allocating per request and due to the high volume of the rhsm_resource_list (used by subman) I saw benefit in caching the responses. The idea being that over a period of time we can avoid hundreds of thousands of object allocations, if not millions, and slightly faster response time

Compare the response time and allocation numbers for the RootController endpoints. My stats show this kind of improvement:

Before
```
REQUEST                                                                 REQUESTS        TOTAL ALLOCATIONS       AVG ALLOCATIONS
Katello::Api::V2::RootController#rhsm_resource_list                     501             6023296                 12022


response times
type                                            count   min     max     avg     mean    sum             percentage
--------------------------------------------------------------------------------------------------------------------
RootController#rhsm_resource_list               501     11      34      13      13      6753            100.00 %
```

After
```
response times
type                                            count   min     max     avg     mean    sum             percentage
--------------------------------------------------------------------------------------------------------------------
RootController#rhsm_resource_list               501     4       25      4       4       2247            100.00 %


REQUEST                                                                 REQUESTS        TOTAL ALLOCATIONS       AVG ALLOCATIONS
Katello::Api::V2::RootController#rhsm_resource_list                     501             927748                  1851
```

### DIFF
--- a/app/controllers/katello/api/v2/root_controller.rb
+++ b/app/controllers/katello/api/v2/root_controller.rb
@@ -1,6 +1,8 @@
 module Katello
   class Api::V2::RootController < Api::V2::ApiController
     skip_before_action :authorize # ok - only shows URLs available
+    skip_before_action :load_settings # no Settings used here, avoid DB calls
+    after_action :skip_session
 
     resource_description do
       api_version 'v2'
@@ -8,7 +10,7 @@ module Katello
     end
 
     def resource_list
-      respond_for_index :collection => RESOURCE_LIST
+      render json: self.class.rhsm_resource_list
     end
 
     def rhsm_resource_list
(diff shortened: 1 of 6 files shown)


app/controllers/katello/api/v2/root_controller.rb
```diff
@@ -8,6 +8,19 @@ class Api::V2::RootController < Api::V2::ApiController
     end
 
     def resource_list
```
jturel, line 10: I thought this endpoint was used by hammer but now I'm not so sure. Any ideas?

jlsherrill: i don't see this in foreman at all, and looking at history it was moved over from the v1 api, so i imagine it can be removed?

jturel:
Oh I messed this branch up badly. Fixing

jturel:
@jlsherrill removed the old resource_list

jlsherrill (approved):
ACK, tested and looked at the general rails allocations reported.  It was a bit lower on the initial request after restart, but subsequent requests were about 1/5th - 1/6th of the pre-patch allocations.

PR #9702, opened 2021-10-06
Fixes #33650 - Add filename attribute to generic content units

rverdile:
This adds a `filename` column to the `katello_generic_content_units` table and makes filename viewable and searchable in the python packages browser UI. The purpose of this is to help differentiate between python packages, as described here: https://projects.theforeman.org/issues/33650
1. Create and sync a python repository with `https://pypi.org` as `upstream url` and `numpy==1.21.1` as `includes`.
2. Go  `Other Content Types` in the navbar and observe the python packages listed. There should be "name", "version", and "filename" columns.
3. Check that these packages are searchable by filename.
4. Without the changes, there would only be "name" and "version" columns and you would not be able to distinguish between the files.


rverdile:
moved the model lambdas to the content type

rverdile:
> @rverdile does this PR require a previous PR to see the other content tab?

nope, go to `https://hostname/content`

rverdile:
I think the newly merged repositories content browser should also show the filename. Still figuring out how to add that in.

PR #9703, opened 2021-10-06
Fixes #33648 - errata & repo association models not being updated

ianballou:

Errata and their repo association models will now be updated at sync time if the upstream repository made updates to them.

1) Use the python simple http server to host the "zoo" repo in the Katello source code's test/fixtures/test_repos
2) Sync that local repo
3) Look at one of the repo's errata and note down its Pulp 3 href (using the repo association class)
4) Edit the same erratum (the title or description perhaps) in the repo's updateinfo.xml, delete everything in repodata, run ./create, and host again.
5) Sync the repo again using a complete sync. This forces content indexing to run.
6) Look at the erratum in the repository. Note that the erratum has not been updated with what you expect
7) Look at the repo association model again. Note that the `erratum_pulp3_href` has not changed (it should've!)
8) To be extra sure a new erratum was generated, use the pulp cli to check the advisories in the repo's latest version. You should see that your edited erratum/advisory is correct in Pulp
9) Test that you can run index_content multiple times on any given repo without error.

Patch in this PR and re-check steps (6) and (7).

Also, please test that other indexing things have not broken.  Try other repository types.  Try Red Hat content.

TODO:

- [x] Tests

### DIFF
--- a/app/models/katello/concerns/pulp_database_unit.rb
+++ b/app/models/katello/concerns/pulp_database_unit.rb
@@ -173,6 +173,7 @@ module Katello
                          !repository.repository_type.unique_content_per_repo &&
                          service_class.supports_id_fetch?
 
+        erratum_updated_ids = []
         service_class.pulp_units_batch_for_repo(repository, fetch_identifiers: fetch_only_ids, content_type: generic_content_type).each do |units|
           units.each do |unit|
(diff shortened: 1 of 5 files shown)


jjeffers:
Preliminary testing looks good. I will attempt with a different repo to confirm we get what we expect. :+1:

PR #9707, opened 2021-10-07
Fixes #33292 - OSTree CLI upload content

jjeffers:

1. install and ready the OSBuild utility and CLI tools: https://www.osbuild.org/guides/user-guide/installation.html
2. follow the pulp3 ostree workflows to create a sample OSTree archive .tar file https://docs.pulpproject.org/pulp_ostree/workflows/import.html

Using hammer:
```
hammer repository upload-content --id <repostory_id> --content-type ostree_ref --path <path to ostree .tar file> --ostree-repository-name repo
```

There are additional options that can be passed to the content import process for ostree content (commit refs): `ref`, `repository_name`, and `parent_commit`. Repository name is *required*, although the default is typically "repo", so that can be used in most example.

```
Could not upload the content:
```

### DIFF
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/views/repository-info.html
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/views/repository-info.html
@@ -514,23 +514,7 @@
                   name="ostree_repository_name"
                   autocomplete="off"/>
           <p class="help-block" translate>
-            The name of a repository that contains the compressed OSTree content.
-          </p>
-
-          <div translate>Ref</div>
-          <input id="ostree_ref"
-                  name="ostree_ref"
-                  autocomplete="off"/>
-          <p class="help-block" translate>
-            The name of a ref branch that holds the reference to the last commit.
-          </p>
-
-          <div translate>Parent Commit</div>
-          <input id="ostree_parent_commit"
-                  name="ostree_parent_commit"
-                  autocomplete="off"/>
-          <p class="help-block" translate>
-            The checksum of a parent commit with which the content needs to be associated.
+            The name of a repository that contains the compressed OSTree content. For example, "repo".
           </p>
         </div>
       </form>


jjeffers:
> Which api endpoint are we trying to access to see if there are duplicates? I would think here we'd be hitting the artifact api. I'm not fully sure it makes sense to hit a 'ref' api and ask it for a checksum of the tarball we just uploaded.

I was just looking at the current implementation. So I don't think there was a particular reason beyond that.
 
> I remember we were told this workflow may be... different than others. It may be that we only need to check for duplicate artifacts (that is already being done), and further checking of the 'unit' isn't needed?

That's an interesting idea, I'll try that.

app/controllers/katello/api/v2/repositories_controller.rb
```diff
@@ -361,6 +361,9 @@ def remove_content
     param :id, :number, :required => true, :desc => N_("repository ID")
     param :content, File, :required => true, :desc => N_("Content files to upload. Can be a single file or array of files.")
     param :content_type, RepositoryTypeManager.uploadable_content_types(false).map(&:label), :required => false, :desc => N_("content type ('deb', 'docker_manifest', 'file', 'ostree', 'rpm', 'srpm')")
+    param :ostree_repository_name, String, :desc => N_("Name of repository in OSTree archive")
+    param :ostree_ref, String, :desc => N_("OSTree ref branch that holds the reference to the last commit")
+    param :ostree_parent_commit, String, :desc => N_("Checksum of a parent commit with which the OSTreee content needs to be associated")
```
jlsherrill, line 366: i wonder if we could add support for this in generic content type registry?

app/lib/actions/pulp3/orchestration/repository/upload_content.rb
```diff
@@ -4,28 +4,58 @@ module Orchestration
...
               duplicate_content_href = duplicate_sha_path_content_list&.results&.first&.pulp_href
 
               unless duplicate_content_href
-                duplicate_sha_artifact_list = ::Katello::Pulp3::Api::Core.new(smart_proxy).artifacts_api.list("sha256": Digest::SHA256.hexdigest(File.read(file[:path])))
+                if unit_type_id == 'ostree_ref'
+                  duplicate_sha_artifact_list = ::Katello::Pulp3::Api::Core.new(smart_proxy).artifacts_api.list(
+                    'checksum': Digest::SHA256.hexdigest(File.read(file[:path])))
+                else
+                  duplicate_sha_artifact_list = ::Katello::Pulp3::Api::Core.new(smart_proxy).artifacts_api.list(
+                    'sha256': Digest::SHA256.hexdigest(File.read(file[:path])))
```
jlsherrill, line 22: does the artifacts api really support checksum and sha256 arguments?  Looking at the client bindings ruby (pulpcore 3.15.2), i don't see checksum listed

app/services/katello/pulp3/content.rb
```diff
@@ -9,11 +9,17 @@ def create_upload(size = 0, checksum = nil, content_type = nil, repository = nil
           if checksum
             content_backend_service = SmartProxy.pulp_primary.content_service(content_type)
             if repository&.generic?
-              content_list = content_backend_service.content_api(repository.repository_type, content_type).list("sha256": checksum)
+              if content_type == 'ostree'
+                content_list = content_backend_service.content_api(repository.repository_type, content_type).list('checksum': checksum)
+              else
+                content_list = content_backend_service.content_api(repository.repository_type, content_type).list('sha256': checksum)
```
jlsherrill, line 15: i do see a 'checksum' argument for the content_refs_api, but not for the artifacts api

app/controllers/katello/api/v2/repositories_controller.rb
jlsherrill: I could envision the repository type registry supporting this via something like this:
~~~
  generic_content_type 'ostree_ref',
                       model_class: Katello::GenericContentUnit,
                       pulp3_api: PulpOstreeClient::ContentRefsApi,
                       pulp3_service_class: Katello::Pulp3::GenericContentUnit,
                       model_name: lambda { |pulp_unit| pulp_unit["name"] },
                       model_version: lambda { |pulp_unit| pulp_unit["version"] },
                       import_attribute :parent_commit, :api_param => :ostree_parent_commit, :description => "Checksum of a parent commit blahblahl", :type => String, :required => false
                       import_attribute :repository_name, :api_param => :ostree_repository_name, :description => "Name of the repository in the ostree archive", :type => String, :required => true


~~~

and then these options could be rendered using the repository type manager, and could be fetched from the params and injected dynamically based on content type.

jlsherrill:
differences with ostree (after accounting for generic import params):

1) uses checksum instead of sha256 param for listing duplicate content units
2) doesn't use import_upload.rb to associate with a repository, instead it happens as part of content unit creation
3) uses different method for content unit creation

lets see how we can minimize the 'if ostree' checks on these

app/lib/actions/pulp3/repository/import_upload.rb
```diff
@@ -4,19 +4,31 @@ module Actions
...
+
+          Rails.logger.debug("Pulp3::Repository::ImportUpload input  #{input}")
+
+          if input[:options][:unit_type_id] == 'ostree_ref'
+            artifact_href = input[:options][:artifact_href]
+            repo_name = input[:options][:repository_name]
+            ref = input[:options][:ref]
+            parent_commit = input[:options][:parent_commit]
+            output[:pulp_tasks] = [repo_backend_service.import_content(artifact_href, repo_name, ref, parent_commit)]
+          else
```
jlsherrill, line 26: we mentioned this on our call, but repeating it here.  I think this would go better in save_artifact.rb, as that is what is actually importing the artifact and creating a content unit which this action is responsible for (even if its poorly named)

jjeffers:
Closing in favor of https://github.com/Katello/katello/pull/9768

PR #9714, opened 2021-10-11
Fixes #33547 - Switch old and new Content View URLS

sjha4:
The old UI is now available on legacy/content_views/
The new UI replaces the nav item and /content_views urls.
Go to content > content views
You should be able to run through all flows with the new UI

Go to /legacy/content_views
You should be able to run through all flows with the old UI


Andrewgdewar:
Should this also remove the reference to the lab feature link in the sidebar? 
That goes nowhere after these changes.

chris1984:
I echo what Mr Lenz said. :+1: :heavy_check_mark:

PR #9716, opened 2021-10-11
Fixes #33376 - catch & return new Pulp mirror on sync error

ianballou:

Pulp introduced a new error for repositories that don't work with mirror-on-sync (https://github.com/dralley/pulp_rpm/commit/d4d088ae8d2e20b6d69d8b321b57ebee2afb3163#diff-fb406f31027edc1c5b65abe7635bbd2830a7cc05c24a487344d7649017d14628L107).  This PR catches that error and makes it more friendly for Katello users.

1) Create a yum repo syncing from https://dl.fedoraproject.org/pub/epel/7/x86_64/
2) Set mirror on sync to on
3) Sync the repo and see the error: "Please disable 'mirror on sync' because the upstream repository refers to external resources."

### DIFF
--- a/app/lib/actions/pulp3/abstract_async_task.rb
+++ b/app/lib/actions/pulp3/abstract_async_task.rb
@@ -105,8 +105,17 @@ module Actions
       def check_for_errors
         combined_tasks.each do |task|
           if (message = task.error)
-            fail ::Katello::Errors::Pulp3Error, message
+            fail ::Katello::Errors::Pulp3Error, overwrite_pulp_error(message)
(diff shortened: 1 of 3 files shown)


jjeffers:
Initially testing this on pulp 3.15.2, pulp_rpm 3.15.0 didn't produce this error. In fact the sync proceeded to pull down packages, etc.

Testing with `{"component":"core","version":"3.14.7"},{"component":"rpm","version":"3.14.5"}` the error was trapped and the appropriate message appears.

PR #9354, opened 2021-05-11
Fixes #32396 - Permissions for Import/Export

lfu:
### Changes in this PR
- Replace permission `export_content_views` with `export_content`
- Rename permission `export_library_content` to `export_content`
- Rename permission `import_library_content` to `import_content`
- Add roles `Content Importer` and `Content Exporter`
- Add rake task `katello:upgrades:4.1:update_content_import_export_perms`

### Steps to import/export contents
1. Run `rake db:seed` to sync the new roles into your DB
2. Create a user with role `Content Importer` and `Content Exporter` and assign the `Organization` where your contents will be exported from and imported to.
3. Run `bundle exec hammer -u lucy -p xxx content-export complete version --id 3 --organization=export-7316`
4. Copy the contents in `/var/lib/pulp/exports` exported from step 3 into a directory in `/var/lib/pulp/imports`
5. `bundle exec hammer -u lucy -p xxx content-import version --organization=import --path /var/lib/pulp/imports/2021-05-13T13-54-14-00-00 --metadata-file metadata-17.json` 
6. Verify the contents are imported in UI.

**For existing users with permission `export_content_views`, `export_library_content` or `import_library_content`**
1. Run `bundle exec rake katello:upgrades:4.1:update_content_import_export_perms` 
2. Verify those users should be assigned permission `export_content` for `export_content_views` or `export_library_content`, `import_content` for `import_library_content`

PR #9717, opened 2021-10-11
Fixes #33678 - Fixes import/export permissions on an upgrade task

parthaa:
https://github.com/Katello/katello/pull/9354 as a part of issue https://projects.theforeman.org/issues/32396 renamed `export_library_content` and `export_content_views permission` to `export_content`.

As a part of that PR is this command.
`bundle exec rake katello:upgrades:4.1:update_content_import_export_perms`

This task correctly renamed `export_library_content` to `export_content`  but kept the old export_content permission intact in the filter causing duplicate values in the filter.

Updates import/export permissions correctly on the `bundle exec rake katello:upgrades:4.1:update_content_import_export_perms`

This PR deletes the `import_library_content` and `export_library_content` in the filter if the filter already has `import_content`, and `export_content`

There are 2 ways listed in the redmine issue. I recommend using rails console.

- Add the following permissions
```
[11] pry(main)> Permission.create(name: :import_library_content, resource_type: 'Organization')
=> #<Permission:0x0000000013477af8 id: 227, name: "import_library_content", resource_type: "Organization", created_at: Mon, 11 Oct 2021 18:13:42 UTC +00:00, updated_at: Mon, 11 Oct 2021 18:13:42 UTC +00:00>
[12] pry(main)> Permission.create(name: :export_library_content, resource_type: 'Organization')
=> #<Permission:0x0000000018edbe18 id: 228, name: "export_library_content", resource_type: "Organization", created_at: Mon, 11 Oct 2021 18:13:51 UTC +00:00, updated_at: Mon, 11 Oct 2021 18:13:51 UTC +00:00>
```
- Create a role with permissions on resource `organization` => `import_library_content, export_library_content, import_content, export_content`
- `bundle exec rake katello:upgrades:4.1:update_content_import_export_perms`
- Do a `hammer filter list` and look at the permissions of the last row

### Before PR ###

Expected
`import_content, export_content`
Actual:
`import_content, import_content, export_content, export_content`

### After PR ###
Expected == Actual:
`import_content, export_content`

### DIFF
--- a/db/seeds.d/111-upgrade_tasks.rb
+++ b/db/seeds.d/111-upgrade_tasks.rb
@@ -5,7 +5,7 @@ UpgradeTask.define_tasks(:katello) do
     {:name => 'katello:upgrades:4.0:remove_ostree_puppet_content'},
     {:name => 'katello:upgrades:4.1:sync_noarch_content'},
     {:name => 'katello:upgrades:4.1:fix_invalid_pools'},
-    {:name => 'katello:upgrades:4.1:update_content_import_export_perms'},
+    {:name => 'katello:upgrades:4.1:reupdate_content_import_export_perms'},
     {:name => 'katello:upgrades:4.2:remove_checksum_values'}
   ]
 end
--- a/lib/katello/tasks/upgrades/4.1/reupdate_content_import_export_perms.rake
+++ b/lib/katello/tasks/upgrades/4.1/reupdate_content_import_export_perms.rake
@@ -2,7 +2,7 @@ namespace :katello do
   namespace :upgrades do
     namespace '4.1' do
       desc "Update for content import/export permissions."
-      task :update_content_import_export_perms => ['environment'] do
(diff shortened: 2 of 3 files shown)


jturel:
@parthaa it seems to work but this file seems to be written in a way that's a bit too clever for my feeble mind to fully understand. Can you add some test coverage for this in test/lib/tasks/upgrades/4.1/update_content_import_export_perms_test.rb ?

I'm not sure this code will fix users who have already upgraded to 4.1 upstream. Can you make sure they will get fixed? It's important that we keep the DB clean.

Since we're modifying an existing script, i think we need to add `always_run => true` in `db/seeds.d/111-upgrade_tasks.rb` to ensure users get cleaned up when they upgrade again.

jlsherrill:
i would probably propose we rename the upgrade task in order to force it again rather than setting always_run to true, as then it will run on every run of foreman-installer iirc

parthaa:
[test katello]

PR #9719, opened 2021-10-12
Fixes #33684 - wrong condition on host_filter

ezr-ondrej:
The host filter has been flipped and hosts that matched the filter has been left out.
This fixes the condition so the proper Hosts according to the filter get in.


ezr-ondrej:
Failures are definitelly unrelated as this is not covered in tests :)

PR #9721, opened 2021-10-13
Refs #33648 - Ensure errata indexing works during cv publish and scope copied errata to source repo

ianballou:

This PR is a continuation of https://github.com/Katello/katello/pull/9703/

Two things were not fixed:
1) After re-syncing a repo that got updated errata, content view publishing failed
2) Errata were not scoped to the source repository during the content copying step, which caused the discovery of https://pulp.plan.io/issues/9503

PART 1

1) Sync the zoo, zoo_dup, and zoo_dup_dup repositories from the filesystem or by serving them with an HTTP server
  -> Check out the instructions in PR #9703 
2) Edit `UNIT_LIMIT` to equal 2 in app/services/katello/pulp3/repository/yum.rb
  -> This is a quicker way to test large repositories.  The Pulp bug noted above is triggered when an erratum is copied to a repository that has 2 or more errata with the same `errata_id`.  For example, repo "A" has errata "1" and "1".  Repo "B" has erratum "1" as well and copies it to repo "A".  This causes an error.  Pulp only supports duplicate errata in new repositories.
3) Add all of these repositories to a content view
4) Create a filter that includes only the "Armadillo" erratum
5) Publish the content view and see that there is no error

PART 2

1) Edit one of the repositories so that the erratum has a new title or description (or whatever else, just ensure the `errata_id` is the same)
2) Sync the repo again
3) Publish the content view again.  Ensure there is no error.

PART 3

Check that other content view publishing still works. Try other filters and other repositories.  Try filtered content view publishing with larger repositories, like this one for instance: https://download-ib01.fedoraproject.org/pub/epel/7/x86_64/


rverdile:
The little python server is so cool!

PR #9726, opened 2021-10-15
Fixes #33700 - Added/Not added component CVs API fix

sjha4:
Filter component CV query before sending to ComponentViewPresenter
Create 40 cvs.
Add 20 to a Composite CV
Send out a call to katello/api/v2/content_views/:composite_cv_id/content_view_components/show_all?status=Not+added

The API returns empty result cause the first batch 20 cvs is added and the API tries to process and rejects those leading to empty results.


sjha4:
failure seems related..looking..

PR #9727, opened 2021-10-15
Fixes #33276 - Improve TableWrapper change tracking for search & pagination

jeremylenz:
The `spawnFetch` function within `TableWrapper` tracks several changes and prevents sending duplicate API requests for search and pagination.  The changes tracked are
* `prevRequest` - Don't send a request if it has the same metadata as the previously-sent request
* `paginationChangePending` - Don't send a request with stale data if the user has clicked on a pagination button and the request is still in flight
* `prevSearch` - Don't send duplicate requests with the same search query
* `additionalListeners` - Send a new request if `additionalListeners` has changed, even if all of the above conditions would otherwise prevent it
* Also, pagination params are supposed to reset back to page 1 whenever search or active filters change.

`activeFilters` was supposed to be tracking changes, but it wasn't.  This caused pagination to be reset to page 1 when it shouldn't be.

Also, `paginationChangePending` wasn't getting reset when it should be.

* Change `activeFilters` from a boolean prop to an array of strings
* Use a ref to track changes in `activeFilters`
* Add variable names to `spawnFetch` so that logic is more readable
* Clear `paginationChangePending` when resetting to page 1
* Clear `paginationChangePending` before triggering a user pagination change (`onPaginationChange`)
* Send the request through if `additionalListeners` have changed, even if stale pagination data would otherwise prevent it

You may want to check out the first commit, which has a lot of useful console logs added.

Load a page with filters, such as the Content Views tab of a composite content view
Change to page 2 > this should respond on the first click
Change filter from 'All' to 'Not added' > this should respond normally and not cause an infinite loop
Change filter from 'Not added' to 'Added' > this should respond normally and not cause an infinite loop
also make sure changing per_page works everywhere

### DIFF
--- a/webpack/scenes/ContentViews/Details/Filters/CVModuleStreamFilterContent.js
+++ b/webpack/scenes/ContentViews/Details/Filters/CVModuleStreamFilterContent.js
@@ -71,6 +71,7 @@ const CVModuleStreamFilterContent = ({
     ADDED,
     NOT_ADDED,
   ];
+  const selectedAdded = allAddedNotAdded[selectedIndex];
 
   const fetchItems = useCallback((params) => {
     const adjustedParams = { ...params };
@@ -215,6 +216,8 @@ const CVModuleStreamFilterContent = ({
               actionResolver,
             }}
             additionalListeners={[selectedIndex]}
+            activeFilters={[selectedAdded]}
+            defaultFilters={[allAddedNotAdded[0]]}
             status={status}
             onSelect={onSelect(rows, setRows)}
             cells={columnHeaders}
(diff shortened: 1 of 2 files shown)


sjha4:
Noticed one remaining issue..
Screencast from 10-18-2021 05_19_38 PM

jeremylenz:
@sjha4 That component also doesn't pass `activeFilters`.  I'm going to make it a required prop; stand by...

jeremylenz:
Ended up not changing the required props. We probably should think about a solution though, as we now have 2 unenforced requirements:

1. Any filters must be passed as `activeFilters` and `defaultFilters` props
2. Filters must also be passed as additionalListeners

anyway, it's ready for review again :)

jeremylenz:
> Filters must also be passed as additionalListeners

I'm thinking maybe it's safer to have activeFilters always treated as if they are additionalListeners, and not require you to add them explicitly.  thoughts?  (Could be a follow-up PR.)

PR #9732, opened 2021-10-19
Fixes #33548 - Add Bookmarks

Andrewgdewar:

- New CV UI Bookmark Component added only to Content View list for now
  @jlsherrill May need to check with you where else it's needed in the new CV UI work.

- Updated the Search component to match design, rewrote related components using modern React.
- Updated the debounce method used on search, should reduce needless api calls while typing.


sjha4:
The bookmarks work right and code looks good to me..
There's one unrelated issue with the PF4 search where the up and down button don't work when selecting things in the autocomplete drop-down list but that's also in master without these changes..
Ack from me :+1:

I'd appreciate a look from @jeremylenz since it's modifying a few core components.

webpack/components/Bookmark/index.js
```diff
@@ -0,0 +1,89 @@
...
+
+  const setSelectItem = (query) => {
+    if (selectedItem !== query) {
+      selectItem(query);
+    }
+    setDropdownOpen(false);
+  };
+
+  const dropDownItems = [
+    ...results.map(({ name, id, query }) => (
```
jeremylenz, line 36: Is the spread needed? Why not just
```js
const dropdownItems = results.map(...
```

webpack/components/Search/Search.js
```diff
@@ -38,74 +39,55 @@ class Search extends Component {
       params.params || {},
     ];
 
-    if (autoCompleteParams[0] !== '') {
+    if (params.endpoint !== '') {
```
jeremylenz, line 42: What if it's undefined?

webpack/components/Search/Search.js
```diff
@@ -38,74 +39,55 @@ class Search extends Component {
...
 
-  onSearch = (search) => {
-    if (this.props.updateSearchQuery) this.props.updateSearchQuery(search);
-    this.props.onSearch(search);
-  };
+  useEffect(() => {
+    loadSetting(AUTOSEARCH_DELAY);
+    loadSetting(AUTOSEARCH_WHILE_TYPING);
+    // This is needed to prevent cross-repo hook dependencies.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
jeremylenz, line 63: I see `loadSetting` is defined in Foreman like this:
```js
export const loadSetting = settingName => async dispatch => {
  dispatch({ type: GET_SETTING_REQUEST });
  try {
    const { data } = await API.get(`/api/v2/settings/${settingName}`);
    return dispatch({
      type: GET_SETTING_SUCCESS,
      response: data,
    });
  } catch (error) {
    return dispatch({
      type: GET_SETTING_FAILURE,
      result: error,
    });
  }
};
```

* I'm unclear on why it's being passed down here as a prop, rather than simply imported from `foremanReact`.
* Also, it's using our old action style of thunk dispatch -> await -> success or failure dispatch.  We've moved on from that and I think our API middleware pattern is much better.

With that said, what are your thoughts on either (a) importing it directly as-is, or (b) writing a new action here in Katello, the modern way?  Or even (c) fixing the one in Foreman :)  I think any one of these options would remove the need to disable the rule here.

webpack/components/TypeAhead/TypeAhead.js
```diff
@@ -9,81 +9,91 @@ import { default as TypeAheadSearchPf4 } from './pf4Search/TypeAheadSearch';
...
-      inputValue: this.props.initialInputValue,
-    };
-  }
+  const debouncedValue = useDebounce(inputValue, autoSearchDelay);
+  useEffect(
+    () => {
+      onInputUpdate(debouncedValue);
+    },
+    // This is needed to prevent cross-repo hook dependencies.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
jeremylenz, line 34: `onInputUpdate` is defined in `Search.js` in Katello..

webpack/components/Bookmark/AddBookmarkModal.js
```diff
@@ -0,0 +1,87 @@
...
+            value={query}
+            onChange={setQuery}
+          />
+        </FormGroup>
+        <FormGroup fieldId="latest">
+          <Checkbox
+            style={{ marginTop: 0 }}
+            id="public"
+            name="public"
+            label={__('Public')}
```
jeremylenz, line 64: What does "Public" mean?

jeremylenz (commented):
Looking great, thanks @Andrewgdewar!

When I create a bookmark, the bookmark component disappears momentarily and then reappears. I would expect it to persist ;)

webpack/components/Bookmark/AddBookmarkModal.js
Andrewgdewar: No idea! Isn't feature parody great!

Andrewgdewar: In all seriousness though.. I truthfully don't know...

webpack/components/Bookmark/index.js
Andrewgdewar: I append the add bookmark at the end, its not just the map in this case.

webpack/components/Search/Search.js
Andrewgdewar: Touché

webpack/components/Search/Search.js
Andrewgdewar: I chose the first option, worked well.

webpack/components/Search/Search.js
Andrewgdewar: Question mark added.

webpack/components/Bookmark/index.js
jeremylenz: Ahhhh I see it now, I had missed that! :+1:

webpack/components/Search/Search.js
jeremylenz: If `params` exists but doesn't have an `endpoint` property, `params?.endpoint` would evaluate to `undefined` and we'd be comparing `undefined !== ''`, which is true.  I assume that's not what you want to happen.  I feel like a simple
```js
if (params?.endpoint) {
```
is a better check?

webpack/components/TypeAhead/TypeAhead.js
Andrewgdewar: Ahh yes, the dependencies here are optionally passed down from props to tableWrapper > onSearch > here. 
Removing this line.

webpack/components/Search/Search.js
Andrewgdewar: After examination of the output of getAutoCompleteParams, this entire block seems pointless.
Rewrote using your suggestion/destructuring.

Andrewgdewar:
@jeremylenz I noticed the disappearing as well!
This led me to discover the issue with the Search not being disabled correctly. 
I 'fixed' both that and made the bookmark portion act in the same manner.

webpack/components/Search/Search.js
```diff
@@ -11,104 +12,79 @@ import {
...
+      setItems(newItems);
     }
 
-    const params = getAutoCompleteParams(searchTerm);
-    const autoCompleteParams = [
-      params.endpoint,
-      params.headers || {},
-      params.params || {},
-    ];
+    const { params, endpoint } = getAutoCompleteParams(searchTerm);
```
Andrewgdewar, line 37: @jeremylenz >>

jeremylenz:
Changes look good but I'm seeing a different issue which is super weird..
I have enough content views that I need to scroll to get to the bottom of the first page.  While the modal is on the screen, initially the modal is ... a modal. But sometimes a scroll bar appears and I am able to scroll the list in the background, with the modal still on screen.  And a few seconds later, the scroll bar disappears and I am unable to scroll again.

This is disconcerting because sometimes (but not always) the appearance/disappearance of the scroll bar affects the position of the modal and makes it shift left or right.

webpack/components/TypeAhead/TypeAhead.js
```diff
@@ -9,81 +9,90 @@ import { default as TypeAheadSearchPf4 } from './pf4Search/TypeAheadSearch';
...
-    this.state = {
-      inputValue: this.props.initialInputValue,
-    };
-  }
+  const debouncedValue = useDebounce(inputValue, autoSearchDelay);
+  useEffect(
+    () => {
+      onInputUpdate(debouncedValue);
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
jeremylenz, line 33: Nice try!  I'm saying you should wrap `onInputUpdate` in a useCallback so you can properly add it to the dep array and not disable the rule here.

jeremylenz: When you said "I'll remove the line" I thought you were referring to the actual rule disablement and not just the comment about cross-repo dependencies :smile:

Andrewgdewar: I understand what you'd like me to do here, I tried following the dependencies up, I wasn't able to prevent the cyclic behaviour. This is a syntactic choice for me, I prefer my functionality as is, instead of removing it for the sake of lint.

Andrewgdewar:
@jeremylenz 
I'll take a look at the modal issue, is it not present on Master? 
If so I can make another ticket for that.

webpack/components/Bookmark/AddBookmarkModal.js
Andrewgdewar: I like this addition, will do!

jeremylenz:
So far I only see the issue on this modal.  I checked the new filter modal and the publish new version modal, and neither had the issue.

jeremylenz:
based on the above, I'm +1 on making a new issue for the modal weirdness

Andrewgdewar:
@jeremylenz I'll see if it's just a minor css change, it may be due to some CSS I wrote around modal sizing a while back not being correct in certain cases. If it's a small change, I'll include it along with the addition of the information piece in the modal.

webpack/components/TypeAhead/TypeAhead.js
Andrewgdewar: :crying_cat_face:  :heavy_check_mark:

Andrewgdewar:
I wasn't able to easily replicate the behaviour you describe. I'm pupup question mark to the modal for the 'public' field and we can make another ticket if needed.

jeremylenz (approved):
Works great, and the tooltip helps :)

Thanks @Andrewgdewar 

ACK :+1:

PR #9743, opened 2021-10-25
Fixes #33705 - only look at pulp primary for repo types

jlsherrill:
The previous code was intended to look up the pulp primary (non-mirrored) smart proxies, but it also was returning smart proxies with pulp not installed.  It was also doing it at startup, instead of on demand.  So this pr:

1.  Correctly only looks at smart proxies that have a pulpcore feature
2.   Moves the calculation of 'enabled' types to happen when its first requested, so it doesn't happen at startup
3.   fixed a bunch of references that used  `enabled_only: false`, to load defined types that was incorrect since the method signatures were using 'enabled_only = true' as a positional arg

Reproducer steps here: https://github.com/Katello/katello/pull/9743#issuecomment-952900529


jlsherrill:
this is now green!

ianballou (approved):
Looks good!

PR #9744, opened 2021-10-25
Fixes #33779 - UI support for force_deletion of Repositories

sjha4:

Adds UI bits for force deleting a repository.

Screenshot from 2021-10-25 14-05-45

This needs to be tested with https://github.com/Katello/katello/pull/9738
Get both commits on a branch and try deleting a repository published in a content view.

### DIFF
--- a/app/models/katello/repository.rb
+++ b/app/models/katello/repository.rb
@@ -336,6 +336,10 @@ module Katello
       content_counts
     end
 
+    def published_in_versions
+      Katello::ContentViewVersion.with_repositories(self.library_instances_inverse).distinct
+    end
+
     def self.errata_with_package_counts(repo)
       repository_rpm = Katello::RepositoryRpm.table_name
       repository_errata = Katello::RepositoryErratum.table_name
--- a/app/views/katello/api/v2/repositories/base.json.rabl
+++ b/app/views/katello/api/v2/repositories/base.json.rabl
@@ -25,6 +25,16 @@ node :content_label do |repo|
   repo.content.try(:label)
 end
 
+child :published_in_versions => :content_view_versions do |_version|
+  attributes :id, :version
+  node :content_view_id do |object|
+    object.content_view.id
+  end
+  node :content_view_name do |object|
+    object.content_view.name
+  end
+end
+
 child :latest_dynflow_sync => :last_sync do |_object|
   attributes :id, :username, :started_at, :ended_at, :state, :result, :progress
 end
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/repository-details.controller.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/repository-details.controller.js
@@ -30,6 +30,17 @@
             $scope.page.loading = false;
         }
 
+        $scope.repositoryVersions = function () {
+            return _.groupBy($scope.repository.content_view_versions, function(version) {
+                return version.content_view_id;
+            });
+        };
+
+        $scope.repositoryWrapper = {
+            repository: $scope.repository,
+            repositoryVersions: {}
+        };
+
         $scope.product = Product.get({id: $scope.$stateParams.productId}, function () {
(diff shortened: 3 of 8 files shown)


engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/repository-details.controller.js
```diff
@@ -30,6 +30,33 @@
             $scope.page.loading = false;
         }
 
+        $scope.repositoryVersions = function () {
+            var publishedViews = {};
+            angular.forEach($scope.repository.versions, function (version) {
+                if (publishedViews[version.cv_id]) {
```
jturel, line 36: Could a `reduce` method eliminate the need to two nearly identical  if/else blocks?

app/models/katello/repository.rb
```diff
@@ -336,6 +336,25 @@ def content_counts
       content_counts
     end
 
+    def in_published_content_views
+      published_views = {}
+      versions = Katello::ContentViewVersion.joins(:repositories).includes(:content_view).merge(self.library_instances_inverse).distinct
+      versions.each do |version|
```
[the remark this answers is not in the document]
sjha4: Updated to address these reviews.. 👍🏽

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/repository-details.controller.js
sjha4: Could be a mental block but would still need to check if key exists to push to existing array vs initializing an array to add.

```
  var versionHash = {
        versionId: version.id,
        cvId: version.cv_id,
        cvName: version.cv_name,
        version: version.version
  };

  if (publishedViews[version.cv_id]) {
      publishedViews[version.cv_id].push(versionHash);
  } else {
      publishedViews[version.cv_id] = [versionHash];
  }
```
Does this make it better?

jturel: Looks like underscore.js gives us a groupBy!

```
        $scope.repositoryVersions = function () {                                                                                                                                                                                             
          return _.groupBy($scope.repository.versions, function(version) {                                                                                                                                                                    
            return version.cv_id;                                                                                                                                                                                                             
          });                                                                                                                                                                                                                                 
        }; 
```

I'm not sure why you needed to rename the keys of the API response for versions

app/views/katello/api/v2/repositories/base.json.rabl
```diff
@@ -25,6 +25,16 @@ node :content_label do |repo|
   repo.content.try(:label)
 end
 
+child :published_in_versions => :versions do |version|
```
jturel, line 28: What do you think about naming this `content_view_versions` instead? 'versions' without any context makes me think "repository versions" which obviously isn't right. may be just me though.

app/views/katello/api/v2/repositories/base.json.rabl
```diff
@@ -25,6 +25,16 @@ node :content_label do |repo|
...
+    attributes :id, :version
+    node :cv_id do |object|
```
jturel, line 30: would you be OK with expanding these field names ie content_view_id instead of cv_id?

app/views/katello/api/v2/repositories/base.json.rabl
```diff
@@ -25,6 +25,16 @@ node :content_label do |repo|
...
+        object.content_view.id
+    end
+    node :cv_name do |object|
```
jturel, line 33: Ditto on expanding to content_view_name

jturel (changes requested):

sjha4:
@jturel : Updated with the suggested changes..

jturel:
> @jturel : Updated with the suggested changes..

Awesome! I'll give it another test today

jturel (approved):
Works great! Thanks @sjha4

PR #9745, opened 2021-10-25
Fixes #33780 - additional metadata json for generic content unit

rverdile:
This adds an `additional_metadata` column to the generic content units table. This is a json to allow storing miscellaneous metadata values that won't be consistent between generic content types. For python packages specifically, this adds `sha256` and `package type`, including in the python package details UI.

1. Database migration
2. Sync a python repository.
    - Make upstream url `https://pypi.org`
    - Add the name(s) of a package(s) to the includes option, `django` for example.
    - Sync
3. Check the `additional_metadata` field is correctly filled in the console by doing something like `Katello::Repository.find(:repo_id).generic_content_units`. 
4. Check against the pulp database, something like `pulp python content list --filename "filename"`
5. Check the details tab displays `sha256` and `package type` by going to `https://hostname/python_packages/:id`

### DIFF
--- a/test/services/katello/pulp3/repository/python/python_test.rb
+++ b/test/services/katello/pulp3/repository/python/python_test.rb
@@ -52,6 +52,13 @@ module Katello
             post_unit_count = Katello::GenericContentUnit.all.count
             post_unit_repository_count = Katello::RepositoryGenericContentUnit.where(:repository_id => @repo.id).count
 
+            unit = @repo.generic_content_units.first
+
+            assert_equal unit.content_type, "python_package"
+            assert_includes unit.filename, "shelf-reader"
+            assert_not unit.pulp_id.nil? and unit.version.nil?
+            assert unit.additional_metadata['package_type'] and unit.additional_metadata['sha256']
+
             assert_equal post_unit_count, 2
             assert_equal post_unit_repository_count, 2
           end
--- a/webpack/scenes/Content/ContentConfig.js
+++ b/webpack/scenes/Content/ContentConfig.js
@@ -31,6 +31,8 @@ export default () => [
           { title: __('Name'), getProperty: unit => unit?.name },
           { title: __('Version'), getProperty: unit => unit?.version },
           { title: __('Filename'), getProperty: unit => unit?.filename },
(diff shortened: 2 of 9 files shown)


webpack/scenes/Content/ContentConfig.js
```diff
@@ -31,6 +31,8 @@ export default () => [
           { title: __('Name'), getProperty: unit => unit?.name },
           { title: __('Version'), getProperty: unit => unit?.version },
           { title: __('Filename'), getProperty: unit => unit?.filename },
+          { title: __('Package Type'), getProperty: unit => unit?.additional_metadata.package_type },
+          { title: __('sha256'), getProperty: unit => unit?.additional_metadata.sha256 },
```
jlsherrill, line 35: should this cause it to show up in the UI?

rverdile: yeah, on the details tab

rverdile:
Added to sync and index test case

test/services/katello/pulp3/repository/python/python_test.rb
```diff
@@ -52,6 +52,13 @@ def test_index_on_sync
             post_unit_count = Katello::GenericContentUnit.all.count
             post_unit_repository_count = Katello::RepositoryGenericContentUnit.where(:repository_id => @repo.id).count
 
+            unit = Katello::GenericContentUnit.all.first
```
jlsherrill, line 55: doing a .first like this is very dangerous, it might work for now, but in the future could break in unexpected/transient ways.  It'd be better to look it up by the repository:  @repo.generic_content_units.first

rverdile: because if there were more generic content units added in the future, Katello::GenericContentUnit.all would return something else?

jlsherrill (approved):

test/services/katello/pulp3/repository/python/python_test.rb
rverdile: @jlsherrill small question ^^

jlsherrill: i was more thinking that if we added fixtures later on that cause entries to appear in the database, it could fail

PR #9748, opened 2021-10-27
Fixes #33799 - Redirect to select org from CV UI

sjha4:

On the new CV UI, when no org is selected, user should be redirected to select organization page before accessing CV UI.

Select Any Organization in as organization
Go to Content > Content View
You should see the org selector page.


chris1984 (approved):
ACK, works good

PR #9749, opened 2021-10-28
Fixes #33797 - incremental update requirement not detected

ianballou:

The `CloneContents` action now passes the `force_sync = true` option to `IndexContent` to ensure that all of the content view version repositories have the correct content counts.

Applicable errata are being shown as installable because the environmental repositories tied to a host's content view version (its bound repositories) weren't indexing content when they should. These repositories had incorrect content counts that stayed the same after the first content view publish.  So, in the case of this bug, it's likely the full repository was published with all errata and then it was filtered and published again.  This bug is likely related to the changes to the content view publishing logic for content views without dependency solving.  It may also be related to the indexing improvements that introduced the `force_index` option.  For the simplest fix possible, I'm passing in the `force_index = true` option.

1) Create a content host that's on some LCE other than Library.
2) Make a content view and publish + promote any number of repositories with no filter.
3) Install an old version of some package that will trigger an erratum and check that the newer version is applicable and installable.
4) Add in a filter that removes the newer package(s) and publish + promote again.
5) Without this PR, the filtered package's erratum will still be installable. With this PR, the filtered package's erratum will only be applicable.
6) Try all the above with dependency solving on.

### DIFF
--- a/app/lib/actions/katello/repository/clone_contents.rb
+++ b/app/lib/actions/katello/repository/clone_contents.rb
@@ -23,7 +23,7 @@ module Actions
             matching_content = check_matching_content(new_repository, source_repositories)
             metadata_generate(source_repositories, new_repository, filters, rpm_filenames, matching_content) if generate_metadata
 
-            index_options = {id: new_repository.id}
+            index_options = {id: new_repository.id, force_index: true}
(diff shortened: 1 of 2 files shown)


jeremylenz (approved):
Upon CVV publish, this updated my applicable and installable errata counts correctly!

ACK :+1:

PR #9713, opened 2021-10-11
Fixes #33673 - Migrate host to new proxy

stejskalleos:
New API endpoint that generates bash code for host migration to new proxy.
In future this should be replacement for _Foreman Bootstrap script_ new capsule [0].

**User workflow**
* SSH to host
* Run `curl --user admin:changeme "https://foreman.example.com/katello/api/hosts/change_proxy?smart_proxy_id=3&activation_keys=key1" | bash`

**Script workflow**
* Configure `subscription-manager`
* Update host (host group, location, openscap proxy id)

**Related PRs**
* [Documentation](https://github.com/theforeman/foreman-documentation/pull/744)

**Links**
* [0] [Foreman Bootstrap script](https://github.com/Katello/katello-client-bootstrap#system-already-registered-to-a-foreman--katello-server--capsule---new-capsule)

PR #9756, opened 2021-11-02
[WIP] Job template for content source change

stejskalleos:
Waiting for https://github.com/Katello/katello/pull/9780/files

### DIFF
--- a/app/models/katello/concerns/smart_proxy_extensions.rb
+++ b/app/models/katello/concerns/smart_proxy_extensions.rb
@@ -444,10 +444,18 @@ module Katello
 
       def rhsm_url
         if pulp_primary?
-          "https://#{URI.parse(url).host}/rhsm"
+          URI("https://#{URI.parse(url).host}/rhsm")
         elsif pulp_mirror?
-          "https://#{URI.parse(url).host}:8443/rhsm"
+          URI("https://#{URI.parse(url).host}:8443/rhsm")
         end
+      end
+
+      def pulp_content_url
+        URI(setting(SmartProxy::PULP3_FEATURE, 'content_app_url'))
+      end
+
+      class ::SmartProxy::Jail < ::Safemode::Jail
+        allow :rhsm_url, :pulp_content_url, :api_url
       end
     end
   end
--- a/app/views/foreman/job_templates/change_content_source_and_host.erb
+++ b/app/views/foreman/job_templates/change_content_source_and_host.erb
@@ -1,0 +1,46 @@
+<%#
+kind: job_template
+name: Change content source and update host
+job_category: Katello
+model: JobTemplate
+provider_type: SSH
+description_format: Configure subscription manager to new content source (smart proxy) and update host's content source.
+feature: katello_change_content_source
+template_inputs:
+- name: smart_proxy
+  required: true
+  input_type: user
+  value_type: resource
+  resource_type: SmartProxy
+  description: New content source (Smart Proxy)
+  advanced: false
+- name: api_user
+  required: true
+  input_type: user
+  value_type: plain
+  description: Username of user for API callback
+  advanced: false
+- name: api_user_password
+  required: true
+  input_type: user
+  value_type: plain
+  description: User password (or personal access token) for API callback
+  hidden_value: true
+  advanced: false
+%>
+#!/bin/sh
+<%
+    smart_proxy = input_resource('smart_proxy')
+-%>
+
+<%= render_template('Change content source', { smart_proxy: smart_proxy.id }) %>
+
+# Update host
+curl --silent --show-error --fail --output /dev/null \
+   --request PUT --cacert $SSL_CA_CERT \
+   --user <%= input('api_user') %>:<%= input('api_user_password') %> \
+   "https://<%= rhsm_url.host %>:8443/api/hosts/<%= @host.id %>" \
+   -H 'Content-Type: application/json' \
+   --data '{ "host": { "content_source_id": "<%= smart_proxy.id %>", "content_facet_attributes": { "content_source_id": "<%= smart_proxy.id %>" }}}' || exit 1
+
+echo "Content source for [<%= @host.name %>] host have been successfully updated."
--- a/app/controllers/katello/concerns/api/v2/registration_controller_extensions.rb
+++ b/app/controllers/katello/concerns/api/v2/registration_controller_extensions.rb
@@ -28,7 +28,7 @@ module Katello
       end
 
       def context_urls
-        super.merge(rhsm_url: rhsm_url, pulp_content_url: pulp_content_url)
+        super.merge(rhsm_url: smart_proxy.rhsm_url, pulp_content_url: smart_proxy.pulp_content_url)
       end
 
       private
@@ -42,14 +42,6 @@ module Katello
 
           proxy
         end
-      end
-
-      def rhsm_url
-        URI(smart_proxy.rhsm_url)
-      end
-
-      def pulp_content_url
-        smart_proxy.setting(SmartProxy::PULP3_FEATURE, 'content_app_url')
       end
     end
   end
--- a/test/models/concerns/smart_proxy_extensions_test.rb
+++ b/test/models/concerns/smart_proxy_extensions_test.rb
@@ -64,12 +64,12 @@ module Katello
     end
 
     def test_rhsm_url_pulp_primary
-      assert_includes @proxy.rhsm_url, "/rhsm"
-      assert_not_includes @proxy.rhsm_url, ":8443"
+      assert_includes @proxy.rhsm_url.to_s, "/rhsm"
+      assert_not_includes @proxy.rhsm_url.to_s, ":8443"
     end
 
     def test_rhsm_url_pulp_mirror
-      assert_includes @proxy_mirror.rhsm_url, ":8443/rhsm"
+      assert_includes @proxy_mirror.rhsm_url.to_s, ":8443/rhsm"
     end
 
     def test_sync_container_gateway


stejskalleos:
@ares @ehelms @evgeni

app/views/foreman/job_templates/change_content_source.erb
```diff
@@ -0,0 +1,86 @@
...
+test -f $RHSM_CFG.bak || cp $RHSM_CFG $RHSM_CFG.bak
+
+subscription-manager config \
+  --server.hostname="<%= rhsm_url.host %>" \
+  --server.port="<%= rhsm_url.port %>" \
+  --server.prefix="<%= rhsm_url.path %>" \
+  --rhsm.repo_ca_cert="$KATELLO_SERVER_CA_CERT" \
+  --rhsm.baseurl="<%= pulp_content_url %>"
+
+# Update host
```
ekohl, line 78: Why is this needed? I would expect the workflow to be:
* User changes content source on the host
* Runs REX job to match the config on the host to what's in Foreman

Essentially: treat Foreman's DB as the source of truth.

ehelms: The primary idea is as you call out to make use of REX to enable changing the proxy a host is pointed at. The secondary idea is to use this same mechanism to allow changing the proxy from the host itself as users do with `bootstrap.py` since we have not achieved REX everywhere across installations. The workflow would be:

 * User on a host hits API to change source
 * Template is rendered and delivered to the host and ran
 
I imagine this being used by an Ansible module or a CLI tool like the newer RHC RHEL is working to provide.

app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -444,11 +444,19 @@ def smart_proxy_service
 
       def rhsm_url
         if pulp_primary?
-          "https://#{URI.parse(url).host}/rhsm"
+          URI("https://#{URI.parse(url).host}/rhsm")
```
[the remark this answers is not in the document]
stejskalleos: On my katello-devel-setup
```
proxy = SmartProxy.find(1)

proxy.setting(SmartProxy::PULP3_FEATURE, 'rhsm_url')
# => "https://localhost/rhsm"

proxy.rhsm_url
# => "https://centos7-katello-devel-stable.example.com/rhsm"
```

Is the value of `rhsm_url` in proxy plugin set by installer or users have to do it manually?
Because if they have to set it manually, there is chance that they didn't do it and it will break the workflow.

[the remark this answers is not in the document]
ekohl: It will be set by the installer (https://github.com/theforeman/puppet-foreman_proxy_content/pull/390) but that hasn't been merged yet. You will also need a fallback for n-1 support. Foreman Proxy 3.0 won't have it and it's supported to register a 3.0 Proxy with a 3.1 Foreman. This fallback is the current code.

app/views/foreman/job_templates/change_content_source.erb
ekohl: > The primary idea is as you call out to make use of REX to enable changing the proxy a host is pointed at.

I'd be against that if I'm honest. I've always seen Foreman as the source of truth. To me it would make more sense to have a "trigger" that if you change the content source on a host that a REX job can automatically be triggered to update the host.

I also see security concerns. Curl runs on the host and there the username and password will be visible in the process tree (`ps aux`). I hadn't mentioned it inline since I first want to have the higher level discussion.



> The workflow would be:
> 
>     * User on a host hits API to change source
> 
>     * Template is rendered and delivered to the host and ran

This makes sense to me and what I was arguing for: treat Foreman as the source of truth and then rectify hosts to match.

app/models/katello/concerns/smart_proxy_extensions.rb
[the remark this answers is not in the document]
wbclark: It's set by the installer as of https://github.com/theforeman/puppet-foreman_proxy_content/pull/384 but the apache vhost to proxy /rhsm on 443 is deployed in https://github.com/theforeman/puppet-foreman_proxy_content/pull/390

app/views/foreman/job_templates/change_content_source.erb
stejskalleos: > * User on a host hits API to change source
> * Template is rendered and delivered to the host and ran

That was original plan, and done in https://github.com/Katello/katello/pull/9713, but then it was decided to go with REX template - it's more customizable and there is no need to maintain another API endpoint

ekohl: In that PR I also see a HTTP call to Foreman to modify data. That is still my problem with this. I think this REX template should be reduced to reading the host property and applying that info to the host in practice. So essentially this REX template without any job parameters and without this curl request.

stejskalleos: > I also see security concerns.

We could fix it fairly easy. Create new template macro `user_api_auth_token`, that would generate API auth token for the current user. Of course it would allow to set the scope and expiration time, same way as we do in the registration template [0]. 
This way we can call Foreman API without exposing user's credentials. Thoughts?

[0] https://github.com/theforeman/foreman/blob/develop/app/controllers/concerns/foreman/controller/registration_commands.rb#L39

ekohl: It still leaves a window of attack and I don't see the need to. Going Foreman -> Host -> Foreman feels plain wrong to me.

app/models/katello/concerns/smart_proxy_extensions.rb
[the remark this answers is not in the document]
ekohl: Since it should land in 4.3 I opened https://github.com/Katello/katello/pull/9777 to implement this.

app/views/foreman/job_templates/change_content_source.erb
```diff
@@ -0,0 +1,46 @@
...
+description_format: Configure subscription manager to new content source (smart proxy)
+feature: katello_change_content_source
+template_inputs:
+- name: smart_proxy
+  required: true
+  input_type: user
+  value_type: resource
+  resource_type: SmartProxy
+  description: New content source (Smart Proxy)
+  advanced: false
```
ekohl, line 16: IMHO this template should not have any inputs. It should use `host.content_source` as the Smart Proxy.

ekohl (changes requested):

app/views/foreman/job_templates/change_content_source.erb
stejskalleos: That would not work in case of rendering `app/views/foreman/job_templates/change_content_source_and_host.erb` template.

When `change_content_source_and_host.erb` template is rendered, `host.content_source` is still the old one, while `smart_proxy` input is the new one.

So running `change_content_source_and_host` job would not configure `subscription-manager` to the new proxy but to the old one.

ekohl: > That would not work in case of rendering `app/views/foreman/job_templates/change_content_source_and_host.erb` template.

Then I'd recommend to extracted it to a snippet that is properly reusable.

app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -450,12 +450,16 @@ def rhsm_url
         end
       end
 
+      def api_url
+        URI("https://#{URI.parse(url).host}:8443")
```
ekohl, line 454: This isn't quite correct. It only works on content proxies. On a Foreman + Smart Proxy this doesn't exist. So for the common case where there's only a single server you're breaking things. There is no way to detect if this is present or not. It can even be changed via an installer argument.

Also note that in the future we want to move this service away. Then Foreman Proxy can run on port 8443, just as a vanilla Foreman.

This is precisely why we introduced the registration module in Smart Proxy: it's a proper service we can rely on.

Also why I'm hammering on that Foreman -> Client -> Foreman is such a bad idea.

ekohl (changes requested):
At this point I'd strongly urge you to split this PR into two: one with a REX job that only applies the current content source on a host. IMHO that's uncontroversial and can be easily merged.

Then we can talk about the IMHO controversial part of changing host properties via REX.

app/views/foreman/job_templates/change_content_source.erb
stejskalleos: > Then I'd recommend to extracted it to a snippet that is properly reusable.

But we want to have it as job, so users can make changes on the hosts after they changed content source in UI. We can't run snippets as job right?

app/models/katello/concerns/smart_proxy_extensions.rb
stejskalleos: Hmm, so how to solve the problem with calling Foreman API here: https://github.com/Katello/katello/pull/9756/files#diff-dc69411f7cfa18aebde7d457a2d44098317c9d33755139b130ca5c8bd9fd88eaR42

Should we call only Foreman API with the risk that it is not accessible from the host?

stejskalleos:
Split into two PRs: https://github.com/Katello/katello/pull/9780 and this one

app/views/foreman/job_templates/change_content_source.erb
ekohl: I was suggesting to extract the actual logic to a snippet and then create two job templates.

app/views/foreman/job_templates/change_content_source_and_host.erb
```diff
@@ -0,0 +1,46 @@
...
+    smart_proxy = input_resource('smart_proxy')
+-%>
+
+<%= render_template('Change content source', { smart_proxy: smart_proxy.id }) %>
+
+# Update host
+curl --silent --show-error --fail --output /dev/null \
+   --request PUT --cacert $SSL_CA_CERT \
+   --user <%= input('api_user') %>:<%= input('api_user_password') %> \
+   "https://<%= rhsm_url.host %>:8443/api/hosts/<%= @host.id %>" \
```
evgeni, line 42: this wont work if `rhsm_url` points at Foreman (and not a Proxy)

stejskalleos: Hmm, good point. But after seeing https://github.com/Katello/katello/pull/9756#discussion_r746658820, the only way we can do this is just call to the Foreman directly. We can't use proxy

evgeni: shouldn't `rhsm_url.port` point at the right port?!

jturel:
Looks like this needs a rebase now. Please let me know when it's ready for final review

jturel:
@stejskalleos is this no longer needed then?

stejskalleos:
@jturel The feature is still in plan but I will post it in the new PR.

jturel:
@stejskalleos ok. please use a new redmine for that one since the job template is going into 4.3 and that's going to be GA here soon (few days time)

stejskalleos:
Follow PR: https://github.com/Katello/katello/pull/9873

PR #9760, opened 2021-11-03
Fixes #33325 - warn about unsupported content types during proxy sync

ianballou:

1) Repositories of content types belonging to Pulp plugins not installed on a smart proxy will not be proxy synced.
2) The user will be warned of types that won't be synced in the UI

Screenshot from 2021-11-03 12-51-41

In the SyncCapsule action, proxy syncs are planned individually for each repo that should be synced. Now, repos that have missing plugins on the proxy will be skipped. It's pretty straight-forward.

For the UI, I added a new `node` in the API view to return the repo types to skip.  I remove the types that aren't applicable to the smart proxy because, for example, it would be confusing to see "ostree" being skipped when it was never used.

Then, there's simply a new `<div>` that displays the types to skip with a warning message.  It's hidden if there are no types to skip.

1) Set up a smart proxy to sync.
2) Sync a few repositories of different types and ensure they're in the proxy's LCE.
3) Remove the Pulp plugins for those repositories' types and restart the Pulpcore services on the proxy.
4) Refresh the proxies features.
5) See that the types to skip warning is there.
6) Sync the proxy and ensure there's no error.
7) Reinstall the proxy plugins, refresh the proxy, and see that the warning message goes away.

TODO:

- [x] Add action test

### DIFF
--- a/app/lib/actions/katello/capsule_content/sync_capsule.rb
+++ b/app/lib/actions/katello/capsule_content/sync_capsule.rb
@@ -3,6 +3,7 @@ module Actions
     module CapsuleContent
       class SyncCapsule < ::Actions::EntryAction
         include Actions::Katello::PulpSelector
+        # rubocop:disable Metrics/MethodLength
         def plan(smart_proxy, options = {})
           plan_self(:smart_proxy_id => smart_proxy.id)
           action_subject(smart_proxy)
@@ -17,11 +18,11 @@ module Actions
             repos.in_groups_of(Setting[:foreman_proxy_content_batch_size], false) do |repo_batch|
               concurrence do
                 repo_batch.each do |repo|
-                  plan_pulp_action(
-                    [Actions::Pulp::Orchestration::Repository::SmartProxySync,
-                     Actions::Pulp3::CapsuleContent::Sync],
-                                     repo, smart_proxy,
-                                     skip_metadata_check: skip_metadata_check)
+                  if smart_proxy.backend_service_type(repo) == Actions::Pulp3::Abstract::BACKEND_SERVICE_TYPE
+                    plan_action(Actions::Pulp3::CapsuleContent::Sync,
+                      repo, smart_proxy,
+                      skip_metadata_check: skip_metadata_check)
+                  end
                 end
               end
 
@@ -39,6 +40,7 @@ module Actions
             end
           end
         end
+        # rubocop:enable Metrics/MethodLength
 
         def repos_to_sync(smart_proxy, environment, content_view, repository, skip_metatadata_check = false)
           smart_proxy_helper = ::Katello::SmartProxyHelper.new(smart_proxy)
--- a/app/models/katello/concerns/smart_proxy_extensions.rb
+++ b/app/models/katello/concerns/smart_proxy_extensions.rb
@@ -422,7 +422,7 @@ module Katello
         repos = Katello::Repository
         repos = repos.in_environment(environment) if environment
         repos = repos.in_content_views([content_view]) if content_view
-        repos
+        repos.respond_to?(:to_a) ? repos : repos.none
       end
 
       def repos_in_sync_history
--- a/app/views/foreman/smart_proxies/_content_sync.html.erb
+++ b/app/views/foreman/smart_proxies/_content_sync.html.erb
@@ -19,6 +19,13 @@
   </div>
 
   <br>
+  <div ng-hide="syncStatus.unsyncable_content_types.length == 0">
+    <span translate>
+      Pulp plugin missing for synchronizable content types: <b>{{ syncStatus.unsyncable_content_types.join(", ") }}.</b><br />
+      Repositories containing these content types will not be synced.
+    </span>
+  </div>
+
   <div ng-show="syncState.is(syncState.SYNCING, syncState.SYNC_TRIGGERED, syncState.CANCEL_TRIGGERED)">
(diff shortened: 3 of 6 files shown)


ianballou:
This PR will break smart proxy syncing for Pulp 2 proxies! That should be fine since it's targeted for 4.3, but it's worth mentioning.

app/views/katello/api/v2/capsule_content/sync_status.json.rabl
```diff
@@ -2,6 +2,15 @@ object @capsule
 
 attribute :last_sync_time
 
+node :types_to_skip do
```
jturel, line 5: Something about the word 'skip' isn't really jiving with me. It almost gives the impression that this was somehow configured or determined by the user but that doesn't seem to be. Is  `unsyncable_content_types` more clear? This is nitpicking but it's worth being as deliberate as possible. Let me know if you really like the current naming too.

app/views/foreman/smart_proxies/_content_sync.html.erb
```diff
@@ -14,6 +14,10 @@
     </a>
   </div>
 
+  <div ng-hide="syncStatus.types_to_skip.length == 0">
+    <span translate>Pulp plugin missing for synchronizable types:</span> {{ syncStatus.types_to_skip.join(", ") }}.
```
jturel, line 18: I think the user might want to know what this implies - and I think the answer is that these content types will be ignored/skipped when syncing

app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -422,7 +422,7 @@ def repos_in_env_cv(environment = nil, content_view = nil)
         repos = Katello::Repository
         repos = repos.in_environment(environment) if environment
         repos = repos.in_content_views([content_view]) if content_view
-        repos
+        repos.class == Class ? [] : repos
```
jturel, line 425: It'd be nice if this code returned an active record relation albeit an empty one in this case. side note: this method is really weird because it seems to implicitly require cv and/or env params which your change uncovered. I don't wanna tackle 'improving' this method here though.

You could have it return an empty relation like this: `repos.respond_to?(:to_a) ? repos : repos.none` I thought not checking the class may be *slightly* better because we don't actually are about the class, but if it is array-like. Thoughts?

jturel (changes requested):
I'm not asking for as much change as the number of comments might lead you to think :)

app/lib/actions/katello/capsule_content/sync_capsule.rb
```diff
@@ -17,11 +18,11 @@ def plan(smart_proxy, options = {})
             repos.in_groups_of(Setting[:foreman_proxy_content_batch_size], false) do |repo_batch|
               concurrence do
                 repo_batch.each do |repo|
-                  plan_pulp_action(
-                    [Actions::Pulp::Orchestration::Repository::SmartProxySync,
-                     Actions::Pulp3::CapsuleContent::Sync],
-                                     repo, smart_proxy,
-                                     skip_metadata_check: skip_metadata_check)
+                  if smart_proxy.backend_service_type(repo) == 'pulp3'
```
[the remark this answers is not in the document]
ianballou: I can use `Actions::Pulp3::Abstract::BACKEND_SERVICE_TYPE`

app/views/katello/api/v2/capsule_content/sync_status.json.rabl
ianballou: `unsyncable_content_types` works, sure.

app/models/katello/concerns/smart_proxy_extensions.rb
ianballou: works for me!

jturel:
Code looks good! I'll spin up a proxy to test

app/views/foreman/smart_proxies/_content_sync.html.erb
```diff
@@ -14,6 +14,13 @@
     </a>
   </div>
 
+  <div ng-hide="syncStatus.unsyncable_content_types.length == 0">
+    <span translate>
+      Pulp plugin missing for synchronizable content types: {{ syncStatus.unsyncable_content_types.join(", ") }}.<br />
+      These content types will be skipped.
```
jturel, line 20: "Repositories containing these content types will not be synced" ?

app/views/foreman/smart_proxies/_content_sync.html.erb
```diff
@@ -14,6 +14,13 @@
     </a>
   </div>
 
+  <div ng-hide="syncStatus.unsyncable_content_types.length == 0">
```
jturel, line 17: The spacing of the divs on this page really buries this message. Some ways to make this "pop" more (which I would say is the goal of the PR) would be to:

- increase the spacing around these divs
- perhaps move this new div to the top instead of in between the sync time and status divs **or** put it right next to the Synchronize button?

jturel (changes requested):
Works perfectly - added some feedback in the visual aspects

app/views/foreman/smart_proxies/_content_sync.html.erb
ianballou: @jturel what do you think of this?

Screenshot from 2021-11-11 10-27-43

jturel: Nice!

jturel (approved):
Nice improvement @ianballou  - APJ

PR #9761, opened 2021-11-03
Fixes #33834 - Filter errata by type and severity

jeremylenz:

* Add Type and Severity dropdowns to the new host detail page Errata tab
* Change the order of the toggle group and dropdowns to reflect the mockup

These dropdowns are part of the design, but haven't been added yet.

I added params to the `index` method of the controller, so you can pass `type` and `severity` params.  Special case for Errata: On the backend, a single errata_type may be multiple strings (for instance, see https://github.com/Katello/katello/blob/574136ae75b96b2ece56a0f7c8a6233547e1551b/app/models/katello/erratum.rb#L6)

I made it so that you can pass a `type` param of `bugfix` and it will still search for `.where(:type => ["bugfix", "recommended"])`.

An alternate approach would have been to add "errata_type = " and "severity = " to the user's search on the frontend before sending the request, but that seemed messy and hacky to me. I saw that the CV UI pages don't do this. Rather, they have a similar approach to the one I went with, modifying the API to reflect the dropdowns.

#### Get a host with applicable errata of all types

The easiest way I've found to get "real" errata to play with is
1. Enable the Red Hat repository "Red Hat Enterprise Linux 7 Server RPMs x86_64 7Server"
2. Spin up a `rhel7` virtual machine using Forklift (copy over the definition from sat-deploy)
3. register the client to your Katello using global registration
3. `yum clean all` on the client if needed

#### Test that everything works
* Filter by each type
* Filter by each severity
* Filter by several combinations of type & severity
* Perform a search and then filter it
* Change pages while filtering
* Select items across pages to apply

### DIFF
--- a/app/controllers/katello/api/v2/host_errata_controller.rb
+++ b/app/controllers/katello/api/v2/host_errata_controller.rb
@@ -46,20 +46,9 @@ module Katello
     param :severity, String, :desc => N_("Return only errata of a particular severity (None, Low, Moderate, Important, Critical)"), :required => false
     param_group :search, Api::V2::ApiController
     def index
-      if (params[:content_view_id] && params[:environment_id].nil?) || (params[:environment_id] && params[:content_view_id].nil?)
-        fail _("Either both parameters 'content_view_id' and 'environment_id' should be specified or neither should be specified")
-      end
+      validate_index_params!
 
-      filtered_relation = index_relation
-      if params[:type].present?
-        fail _("Type must be one of: %s" % TYPES_FROM_PARAMS.keys.join(', ')) unless TYPES_FROM_PARAMS.key?(params[:type].to_sym)
-        filtered_relation = filtered_relation.where(:errata_type => TYPES_FROM_PARAMS[params[:type].to_sym])
-      end
-      if params[:severity].present?
-        fail _("Severity must be one of: %s") % Katello::Erratum::SEVERITIES.join(', ') unless Katello::Erratum::SEVERITIES.include?(params[:severity])
-        filtered_relation = filtered_relation.where(:severity => params[:severity])
-      end
-      collection = scoped_search(filtered_relation, 'updated', 'desc', :resource_class => Erratum, :includes => [:cves])
+      collection = scoped_search(index_relation, 'updated', 'desc', :resource_class => Erratum, :includes => [:cves])
 
       @installable_errata_ids = []
       if @host.content_facet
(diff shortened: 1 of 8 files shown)


app/controllers/katello/api/v2/host_errata_controller.rb
```diff
@@ -3,6 +3,12 @@ class Api::V2::HostErrataController < Api::V2::ApiController
     include Katello::Concerns::FilteredAutoCompleteSearch
     include Katello::Concerns::Api::V2::HostErrataExtensions
 
+    TYPES_FROM_PARAMS = {
```
parthaa, line 6: cant you just use https://github.com/Katello/katello/blob/master/app/models/katello/erratum.rb#L9 ?

parthaa: I guess you need a map ??

app/controllers/katello/api/v2/host_errata_controller.rb
```diff
@@ -135,5 +147,17 @@ def find_errata_ids
         @errata_ids = find_bulk_errata_ids([@host], params[:bulk_errata_ids])
       end
     end
+
+    def validate_index_params!
+      if (params[:content_view_id] && params[:environment_id].nil?) || (params[:environment_id] && params[:content_view_id].nil?)
+        fail _("Either both parameters 'content_view_id' and 'environment_id' should be specified or neither should be specified")
+      end
+      if params[:type].present?
+        fail _("Type must be one of: %s" % TYPES_FROM_PARAMS.keys.join(', ')) unless TYPES_FROM_PARAMS.key?(params[:type].to_sym)
```
parthaa, line 156: Why do you need TYPES_FROM_PARAMS. You can do what you did for severity below right?

parthaa (commented):
Seems to work well. Just tested the PR with Rhel7 . Working as designed.

Couple of questions but going to approve once they are answered.

parthaa:
Usability issue:
Need a way to clear/unset Severity and Type. Right now if you chose a type/severity, there is no way to "unset" it. So if you want the original master list you will have to reload the content errata page.

parthaa:
> Need a way to clear/unset Severity and Type. Right now if you chose a type/severity, there is no way to "unset" it. So if you want the original master list you will have to reload the content errata page.

Actually seems to work fine now. Ignore the above comment

parthaa (approved):

app/controllers/katello/api/v2/host_errata_controller.rb
jeremylenz: Yeah, for example I need to be able to send a `type` param of just `bugfix`, but the backend still needs to know to search for types matching either `bugfix` or `recommended`.  So passing in the param as-is wouldn't work here.

app/controllers/katello/api/v2/host_errata_controller.rb
jeremylenz: This is validating the params, which should only be `bugfix`, `security`, or `enhancement`.  If I just did `Katello::Erratum::TYPES`, it would also allow you to send params of `optional` or `recommended`, which I don't want.

jeremylenz (commented):
answers below

PR #9762, opened 2021-11-04
Fixes #33847 - use pulpcore-manager for reset

jlsherrill:
change to use pulpcore-manager for pulp 3.15

Try a rake katello:reset


jjeffers (approved):
I ran into the same problems! Same fix!

PR #9765, opened 2021-11-04
Fixes #33303 - ostree and python FPC sync support

jlsherrill:
Support for generic content types being synced to a smart proxy.  These generic types had some support for a customizable way to define 'relative_path', which is the external path to access the content.  The support however wasn't fully implemented, so i've removed it since neither ostree nor python actually need it. 

In addition there was an error seen when looking up the ContainerGateway constant that may only occur in dev, but i was seeing it consistently so went ahead and fixed it.

1.  Apply this patch to pulp_ostree: https://github.com/pulp/pulp_ostree/commit/baf301ea8abf38e6fe39e7df4a799a05c6afeeda
2. create and sync an ostree repo
3. setup and configure a smart proxy, pointing at library
4. apply the same patch to the smart proxy
4. sync the smart proxy
5. you can curl the ostree repo's path on the smart proxy to see its contents:  for example:  curl https://mycapsule/pulp/content/Default_Organization/Library/custom/ostree/ostree2/

Python support is broken due to https://github.com/pulp/pulp_python/issues/462  but ostree is the main drive here, so i don't think it should block this.

### DIFF
--- a/app/models/katello/concerns/smart_proxy_extensions.rb
+++ b/app/models/katello/concerns/smart_proxy_extensions.rb
@@ -1,6 +1,7 @@
 require 'proxy_api'
 require 'proxy_api/pulp'
 require 'proxy_api/pulp_node'
+require 'proxy_api/container_gateway'
 
 module Katello
   module Concerns
@@ -128,7 +129,7 @@ module Katello
                            auth_required: !unauthenticated_container_repositories.include?(repo.id) }
           end
         end
-        ProxyAPI::ContainerGateway.new(url: self.url).repository_list({ repositories: repo_list })
+        ::ProxyAPI::ContainerGateway.new(url: self.url).repository_list({ repositories: repo_list })
       end
 
       def update_user_container_repo_mapping(users)
--- a/app/models/katello/repository.rb
+++ b/app/models/katello/repository.rb
@@ -440,8 +440,6 @@ module Katello
       scheme = force_http ? 'http' : 'https'
       if docker?
         "#{pulp_uri.host.downcase}/#{container_repository_name}"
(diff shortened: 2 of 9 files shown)


jlsherrill:
ostree is broken until https://github.com/pulp/pulp_ostree/issues/143 is resolved

jlsherrill:
I see a similar issue with python:  https://github.com/pulp/pulp_python/issues/462

rverdile:
Just curious, what's the purpose of the partial repo path change?

Also, would it work in all cases? Because, for example, the file type looks a bit different: https://github.com/Katello/katello/blob/master/app/services/katello/pulp3/repository/file.rb#L28-L30

jlsherrill:
Its the path that a remote pulp uses to 'sync' from this pulp.  File is different because for pulp_file you 'sync' from the PULP_MANIFEST file.

rverdile:
Is `"/pulp/content/#{repo.relative_path}/".sub('//', '/')` universally usable as a partial_repo_path? For example, if File were generic, would that need to be updated?

jlsherrill:
Its not necessarily universal (for example docker and ansible collections would have a different url to sync from).  File would be the same (and rpm would too).  We may hit a type that needs something special in the future, and i think we can add in repository customization at that point IMO

jjeffers (approved):
Tested the given pulp patched katello and proxy, synced a freshly made ostree repository with content from `https://fixtures.pulpproject.org/ostree/small/`. 

Post-capsule sync the proxy is reported as synchronized. Checked the content on the proxy via curl which was a nice mirror of the path indicated on the repository info page on the katello server.

Overall a rather clean update.

PR #9766, opened 2021-11-05
Fixes #33852 - Content - Errata - Add REX actions

parthaa:
On the new host errata page
- Added support for installing both individual and multiple errata (including select all) via Katello Agent or Rex
- Updated the errata install template with new added fields to support `Select All` and excludes
- For applicable errata you will be able to go to the appropriate `Apply Erratum` page when you click on the kebab.

##### On the dev env #####
-  Go to Hosts -> Job Templates
-  Unlock the `Install Errata - Katello SSH Default` template using the Actions Dropdown
-  Run this from the rails console
```ruby
    JobTemplate.without_auditing do
      template_files = Dir[File.join("#{Katello::Engine.root}/app/views/foreman/job_templates/**/*.erb")]
      template_files.reject! { |file| file.end_with?('_ansible_default.erb') } unless Katello.with_ansible?
      template_files.each do |template|
        sync = !Rails.env.test? && Setting[:remote_execution_sync_templates]
        # import! was renamed to import_raw! around 1.3.1
        if JobTemplate.respond_to?('import_raw!')
          template = JobTemplate.import_raw!(File.read(template), :default => true, :lock => true, :update => sync)
        else
          template = JobTemplate.import!(File.read(template), :default => true, :lock => true, :update => sync)
        end

        template.organizations << Organization.unscoped.all if template&.organizations&.empty?
        template.locations << Location.unscoped.all if template&.locations&.empty?
      end
    end
```

- This should update the install errata erb template. 
- In the UI Go to Hosts -> Job Templates
- Click on `Install Errata - Katello SSH Default` template
- Click on the `Inputs` tab. You should see 3 inputs with name `Filte Errata`, `Inclusion Type`, and `errata`

##### Set up the host ##### 

These steps will give you both `Applicable` and `Installable` errata on the host. 

1. Create and Sync this repo -> https://partha.fedorapeople.org/test-repos/multi-errata/
2. Create another repo in the same product. Leave the feed empty
3. Create  a Dev lifecycle environment
4. Create a content view with the repos created in steps 1 and 2
5. Publish and Promote it to dev
6. Create an activation that points dev/cv. Make sure the subscriptions are turned on
7. Register a rhel7/centos7 with that activation key
8. On client `yum install armadillo-0.1-1.noarch`
9. Also `rpm -ivh https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/walrus-0.71-1.noarch.rpm https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/whale-0.2-1.noarch.rpm https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/shark-0.1-1.noarch.rpm https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/stork-0.12-2.noarch.rpm`
10. Now edit the repo in step 2 and add the feed url -> https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/
11. Sync this repo
12. Now you should have 38 Applicable Errata and 37 installable errata
13. Go to Hosts -> New Host Details -> Content -> Errata tab

Now there are 4 bulk apply operations available on this page 
1. Apply by Katello Agent
1. Apply by Remote Execution
1. Apply by Customized Remote Execution
1. Apply button - if remote_execution_by default is enabled then it will use rex. Else it 'll use katello-agent to install the errata.

01-bulk-errata

There are also 3 apply operations available on each installable errata
1. Apply by Katello Agent
1. Apply by Remote Execution
1. Apply by Customized Remote Execution

01-single-errata

On Applicable errata we should see an `Apply Erratum` link that should take you to `Errata - Apply to Content Hosts` page.

01-bulk-errata

- Select All Errata 
- Select `Apply by Customized Remote Execution`
- You should be redirected to the Job Invocations errata install page, where you shoud be able to see elements of the new template
01-select all

- Notice the `Inclusion Type` entry saying `Include all errata EXCEPT below` . The select all resolves to this
- On same page `Filter Errata` will get populated  if you have search query set and done a `Select All`
- Try out different errata operations and make sure they get applied correctly on the host.

#### Backend Changes #####
- Moved Bulk operations to a `BulkItemsHelper` class so that Errata/Traces can make use of the `select all`.
- Updated the `install_errata.erb` template. Added inputs to handle bulk selection operations.
- Added an `advisory_helpers` method in host managed extensions to aid with the new template.

#### Front End Changes ####
* In *ErrataTab* mainly added functionality for errata Apply operations.

### DIFF
--- a/app/views/foreman/job_templates/install_errata.erb
+++ b/app/views/foreman/job_templates/install_errata.erb
@@ -6,18 +6,8 @@ description_format: 'Install errata %{errata}'
 feature: katello_errata_install
 provider_type: SSH
 template_inputs:
-- name: Filter Errata
+- name: Errata Search Query
   description: Filter criteria for Errata to be installed.
-  input_type: user
-  required: false
-- name: Inclusion Type
-  description: Include/Exclude the errata ids specified below.
-  input_type: user
-  required: true
-  options: "Include the errata below\r\nInclude all errata EXCEPT the below"
-  default: 'Include the errata below'
-- name: errata
-  description: A comma separated list of errata to install
   input_type: user
   required: false
 foreign_input_sets:
@@ -29,9 +19,7 @@ foreign_input_sets:
     <% advisories = input(:errata).split(',').join(' ') %>
     <%= render_template('Package Action - SSH Default', :action => 'install -n -t patch', :package => advisories) %>
 <% else %>
-    <% advisory_ids = @host.advisory_helpers(errata: input(:errata).split(','),
-                                          all: "Include all errata EXCEPT the below" == input("Inclusion Type"),
-                                          search: input("Filter Errata")) %>
+    <% advisory_ids = @host.advisory_ids(search: input("Errata Search Query")) %>
 
     <% advisories = advisory_ids.map { |e| "--advisory=#{e}" }.join(' ') %>
     <%= render_template('Package Action - SSH Default', :action => 'update-minimal', :package => advisories) %>
--- a/app/views/foreman/job_templates/install_errata.erb
+++ b/app/views/foreman/job_templates/install_errata.erb
@@ -2,11 +2,11 @@
 kind: job_template
 name: Install Errata - Katello SSH Default
 job_category: Katello
-description_format: 'Install errata %{errata}'
+description_format: 'Install errata %{Errata search query}'
 feature: katello_errata_install
 provider_type: SSH
 template_inputs:
-- name: Errata Search Query
+- name: Errata search query
   description: Filter criteria for Errata to be installed.
   input_type: user
   required: false
@@ -19,7 +19,7 @@ foreign_input_sets:
     <% advisories = input(:errata).split(',').join(' ') %>
     <%= render_template('Package Action - SSH Default', :action => 'install -n -t patch', :package => advisories) %>
 <% else %>
-    <% advisory_ids = @host.advisory_ids(search: input("Errata Search Query")) %>
+    <% advisory_ids = @host.advisory_ids(search: input("Errata search query")) %>
 
     <% advisories = advisory_ids.map { |e| "--advisory=#{e}" }.join(' ') %>
     <%= render_template('Package Action - SSH Default', :action => 'update-minimal', :package => advisories) %>
--- a/app/views/foreman/job_templates/install_errata.erb
+++ b/app/views/foreman/job_templates/install_errata.erb
@@ -7,7 +7,7 @@ feature: katello_errata_install
 provider_type: SSH
 template_inputs:
 - name: Errata search query
-  description: Filter criteria for Errata to be installed.
+  description: Filter criteria for errata to be installed.
   input_type: user
   required: false
 foreign_input_sets:
--- a/app/views/foreman/job_templates/install_errata.erb
+++ b/app/views/foreman/job_templates/install_errata.erb
@@ -6,10 +6,20 @@ description_format: 'Install errata %{errata}'
 feature: katello_errata_install
 provider_type: SSH
 template_inputs:
+- name: Filter Errata
+  description: Filter criteria for Errata to be installed.
+  input_type: user
+  required: false
+- name: Inclusion Type
+  description: Include/Exclude the errata ids specified below.
+  input_type: user
+  required: true
+  options: "Include the errata below\r\nInclude all errata EXCEPT the below"
+  default: 'Include the errata below'
 - name: errata
   description: A comma separated list of errata to install
   input_type: user
-  required: true
+  required: false
 foreign_input_sets:
 - template: Package Action - SSH Default
   exclude: action,package
@@ -19,6 +29,10 @@ foreign_input_sets:
     <% advisories = input(:errata).split(',').join(' ') %>
     <%= render_template('Package Action - SSH Default', :action => 'install -n -t patch', :package => advisories) %>
 <% else %>
-    <% advisories = input(:errata).split(',').map { |e| "--advisory=#{e}" }.join(' ') %>
+    <% advisory_ids = @host.advisory_helpers(errata: input(:errata).split(','),
+                                          all: "Include all errata EXCEPT the below" == input("Inclusion Type"),
+                                          search: input("Filter Errata")) %>
+
+    <% advisories = advisory_ids.map { |e| "--advisory=#{e}" }.join(' ') %>
     <%= render_template('Package Action - SSH Default', :action => 'update-minimal', :package => advisories) %>
 <% end %>
--- a/app/views/katello/api/v2/content_facet/show.json.rabl
+++ b/app/views/katello/api/v2/content_facet/show.json.rabl
@@ -25,6 +25,14 @@ child :content_facet => :content_facet_attributes do
   node :katello_tracer_installed do |content_facet|
     content_facet.tracer_installed?
   end
+
+  node :katello_agent_enabled do
+    Katello.with_katello_agent?
+  end
+
+  node :remote_execution_by_default do
+    Katello.remote_execution_by_default?
+  end
 end
 
 attributes :description, :facts
--- a/webpack/components/Table/TableHooks.js
+++ b/webpack/components/Table/TableHooks.js
@@ -178,7 +178,34 @@ export const useBulkSelect = ({
     }
   };
 
-  const fetchBulkParams = () => {
+  const fetchBulkParamsQueryForm = () => {
+    const query = [];
+    if (selectAllMode) {
+      if (searchQuery) {
+        query.push(searchQuery);
+      }
+      if (!isEmpty(exclusionSet)) {
+        if (!isEmpty(query)) {
+          query.push('and');
+        }
+
+        query.push(idColumn);
+        query.push('!^');
+        query.push(`(${[...exclusionSet].join(',')})`);
+      }
+    } else {
+      query.push(idColumn);
+      query.push('^');
+      query.push(`(${[...inclusionSet].join(',')})`);
+    }
+
+    return query.join(' ');
+  };
+
+  const fetchBulkParams = (queryForm = false) => {
+    if (queryForm) {
+      return fetchBulkParamsQueryForm();
+    }
     const selected = {
       included: {
         ids: [],
@@ -199,6 +226,7 @@ export const useBulkSelect = ({
     } else {
       return {};
     }
+
     return selected;
   };
 
--- a/webpack/components/Table/TableHooks.js
+++ b/webpack/components/Table/TableHooks.js
@@ -178,56 +178,19 @@ export const useBulkSelect = ({
     }
   };
 
-  const fetchBulkParamsQueryForm = () => {
-    const query = [];
-    if (selectAllMode) {
-      if (searchQuery) {
-        query.push(searchQuery);
-      }
-      if (!isEmpty(exclusionSet)) {
-        if (!isEmpty(query)) {
-          query.push('and');
-        }
-
-        query.push(idColumn);
-        query.push('!^');
-        query.push(`(${[...exclusionSet].join(',')})`);
-      }
-    } else {
-      query.push(idColumn);
-      query.push('^');
-      query.push(`(${[...inclusionSet].join(',')})`);
-    }
-
-    return query.join(' ');
-  };
-
-  const fetchBulkParams = (queryForm = false) => {
-    if (queryForm) {
-      return fetchBulkParamsQueryForm();
-    }
-    const selected = {
-      included: {
-        ids: [],
-        search: null,
-      },
-      excluded: {
-        ids: [],
-      },
-      all: false,
+  const fetchBulkParams = () => {
+    const searchQueryWithExclusionSet = () => {
+      const query = [searchQuery,
+        !isEmpty(exclusionSet) && `${idColumn} !^ (${[...exclusionSet].join(',')})`];
+      return query.filter(item => item).join(' and ');
     };
 
-    if (selectAllMode) {
-      selected.included.search = searchQuery;
-      selected.excluded.ids = [...exclusionSet];
-      selected.all = true;
-    } else if (!isEmpty(inclusionSet)) {
-      selected.included.ids = [...inclusionSet];
-    } else {
-      return {};
-    }
+    const searchQueryWithInclusionSet = () => {
+      if (isEmpty(inclusionSet)) throw new Error('No Items Selected');
+      return `${idColumn} ^ (${[...inclusionSet].join(',')})`;
+    };
 
-    return selected;
+    return selectAllMode ? searchQueryWithExclusionSet() : searchQueryWithInclusionSet();
   };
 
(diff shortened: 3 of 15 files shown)


parthaa:
03-ApplicableButNotInstallable
01-single-errata
01-bulk-errata

app/views/foreman/job_templates/install_errata.erb
```diff
@@ -6,10 +6,20 @@ description_format: 'Install errata %{errata}'
...
 template_inputs:
+- name: Filter Errata
+  description: Filter criteria for Errata to be installed.
+  input_type: user
+  required: false
+- name: Inclusion Type
+  description: Include/Exclude the errata ids specified below.
+  input_type: user
+  required: true
+  options: "Include the errata below\r\nInclude all errata EXCEPT the below"
```
jturel, line 17: I think using words like 'below' here is quite odd because we're describing this input field based on some aspect of the UI..

app/views/foreman/job_templates/install_errata.erb
```diff
@@ -19,6 +29,10 @@ foreign_input_sets:
     <% advisories = input(:errata).split(',').join(' ') %>
     <%= render_template('Package Action - SSH Default', :action => 'install -n -t patch', :package => advisories) %>
 <% else %>
-    <% advisories = input(:errata).split(',').map { |e| "--advisory=#{e}" }.join(' ') %>
+    <% advisory_ids = @host.advisory_helpers(errata: input(:errata).split(','),
+                                          all: "Include all errata EXCEPT the below" == input("Inclusion Type"),
```
jturel, line 33: ditto on 'below'

app/views/katello/api/v2/content_facet/show.json.rabl
```diff
@@ -25,6 +25,14 @@ child :content_facet => :content_facet_attributes do
   node :katello_tracer_installed do |content_facet|
     content_facet.tracer_installed?
   end
+
+  node :katello_agent_enabled do
```
jturel, line 29: I know as a dev it's easy to add these here our sake but it feels a bit weird. Random idea: could the /katello/api/v2 (or similar endpoint) expose some amount of configuration? Multiple places may need this info and a content facet may not be in scope at that point (I don't have an example off the top of my head)?

Or maybe Foreman or other plugin has established something we can use?

app/views/foreman/job_templates/install_errata.erb
parthaa: Well this came of a discussion with @jeremylenz . I am ok with other verbiage. Feel free to suggest alternatives.

app/views/katello/api/v2/content_facet/show.json.rabl
parthaa: This was discussed also. I could separately get the configuration separately but having that information embedded in the host details made best sense for me. Given we already have `katello_tracer_installed` and `katello_agent_installed` there. I dont see why adding `katello_agent_enabled` and `remote_execution_by_default` attributes should be a red flag. Considering these are details pertinent to this host?

jturel: The examples you gave are reflective of the state of the content host - whether tracer or agent is installed. The fields you've added are related to the configuration of Katello which I guess is pertinent but also has nothing to do with the content facet itself. If you and Jeremy are good then let's go with it

app/views/foreman/job_templates/install_errata.erb
jturel: My suggestion would be to remove words like 'below' which relate to the UI.

app/views/katello/api/v2/content_facet/show.json.rabl
jeremylenz: I think ForemanContext would be the ideal place to have something like this. It already contains other global app settings. The problem is that adding values there requires Foreman code changes, and also we'd be adding Katello-specific data.  Until we can come up with a way to address both of those issues, I think adding it here is... okay but not great.  Adding to the / API endpoint is an interesting idea too, but then we'd have to make an additional API call.

I think going with what we have is fine for now, because at least it's an API call that we're already making.  Unless someone can think of another brilliant idea :)

jturel: First time I've seen ForemanContext and it really does seem like the right place. I'm surprised that component wasn't designed with plugin extensibility in mind. That is something we should look into in the future!

app/views/foreman/job_templates/install_errata.erb
jeremylenz: @jturel The problem is that this _does_ show up in the UI and this is our only way to control the form fields that appear there. (see the screenshot in the PR description above) I couldn't think of a wording that is truly context-agnostic but also makes sense and is clear enough to be self explanatory.  Do you have any other wording suggestions?

jturel: Could we refer to the field name which seems to be 'errata' (should it be Errata IDs instead?) instead of its location on the screen

parthaa:
@jturel @jeremylenz  Updated this PR and tests to use the `Errata Search Query`. Check the latest commit.

webpack/components/extensions/HostDetails/Tabs/__tests__/errataTab.test.js
```diff
@@ -783,3 +790,418 @@ test('Can filter by severity', async (done) => {
...
+  });
+
+  const scope = nockInstance
+    .get(hostErrata)
+    .query(defaultQuery)
+    .reply(200, mockErrata);
+
+  const postBody = ({ bulk_errata_ids: bulkErrataIds }) => {
+    const { included: { ids }, all } = JSON.parse(bulkErrataIds);
+    return isEqual(ids, [results[0].errata_id, results[1].errata_id]) && !all;
```
jeremylenz, line 848: Whenever you use `[0]` and `[1]` array destructuring can make it a bit cleaner
```suggestion
    const [firstResult, secondResult] = results;
    return isEqual(ids, [firstResult.errata_id, secondResult.errata_id]) && !all;
```

jeremylenz (commented):
looking great!

webpack/components/Table/TableHooks.js
```diff
@@ -178,7 +178,7 @@ export const useBulkSelect = ({
     }
   };
 
-  const fetchBulkParamsQueryForm = () => {
+  const fetchBulkParams = () => {
     const query = [];
     if (selectAllMode) {
       if (searchQuery) {
```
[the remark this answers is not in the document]
parthaa: what do you think of
```javascript
  const fetchBulkParams = () => {
    const searchQueryWithExclusionSet = () => {
      const query = [searchQuery,
                     !isEmpty(exclusionSet) && `${idColumn} !^ (${[...exclusionSet].join(',')})`];
      return query.filter(item => item).join(" and ");
    }

    const searchQueryWithInclusionSet = () => isEmpty(inclusionSet) ? '' :
                                              `${idColumn} ^ (${[...inclusionSet].join(',')})`;

    return selectAllMode ? searchQueryWithExclusionSet() : searchQueryWithInclusionSet();
  };
```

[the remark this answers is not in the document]
jeremylenz: looks good! except we still have to account for the blank search with inclusion set :)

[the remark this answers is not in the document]
parthaa: Fixed :) check now

app/views/foreman/job_templates/install_errata.erb
```diff
@@ -6,10 +6,10 @@ description_format: 'Install errata %{errata}'
 feature: katello_errata_install
 provider_type: SSH
 template_inputs:
-- name: errata
-  description: A comma separated list of errata to install
+- name: Errata Search Query
```
jeremylenz, line 9:
```suggestion
- name: Errata search query
```

On all the new pages we've been instructed to use this style of capitalization.

app/views/foreman/job_templates/install_errata.erb
```diff
@@ -19,6 +19,8 @@ foreign_input_sets:
     <% advisories = input(:errata).split(',').join(' ') %>
     <%= render_template('Package Action - SSH Default', :action => 'install -n -t patch', :package => advisories) %>
 <% else %>
-    <% advisories = input(:errata).split(',').map { |e| "--advisory=#{e}" }.join(' ') %>
+    <% advisory_ids = @host.advisory_ids(search: input("Errata Search Query")) %>
```
jeremylenz, line 22:
```suggestion
    <% advisory_ids = @host.advisory_ids(search: input("Errata search query")) %>
```

webpack/components/Table/TableHooks.js
```diff
@@ -179,27 +179,18 @@ export const useBulkSelect = ({
...
-      selected.excluded.ids = [...exclusionSet];
-      selected.all = true;
-    } else if (!isEmpty(inclusionSet)) {
-      selected.included.ids = [...inclusionSet];
-    } else {
-      return {};
-    }
-    return selected;
+    const searchQueryWithInclusionSet = () => {
+      if (isEmpty(inclusionSet)) throw new Error('No Items Selected');
```
jeremylenz, line 189:
```suggestion
      if (isEmpty(inclusionSet)) throw new Error('Cannot build a search query with no items selected');
```

webpack/components/extensions/HostDetails/Tabs/ErrataTab.js
```diff
@@ -273,6 +329,36 @@ export const ErrataTab = () => {
                 installable: isInstallable,
               } = erratum;
               const isExpanded = erratumIsExpanded(id);
+              let rowActions;
+              if (isInstallable) {
+                rowActions = [
```
jeremylenz, line 334: For non-installable errata, I think these items should still be there, just disabled.

app/views/foreman/job_templates/install_errata.erb
jeremylenz: I think this broke the display of the job name. All my tasks now say
```
Install errata ''
```
but they should say the errata ID:
```
Install errata '2021-blahblah-30148u'
```
I think if you change line 5 above to
```
description_format: 'Install errata %{Errata search query}'
```
it will maybe hopefully fix it.

jeremylenz (commented):
Haven't been able to test select all yet but I tested the REX actions and they work as expected.

a few comments below :)

webpack/components/extensions/HostDetails/Tabs/ErrataTab.js
parthaa: Its a different list for non installable but applicable.

jeremylenz: not sure what you mean here?

app/views/foreman/job_templates/install_errata.erb
parthaa: fixed

webpack/components/extensions/HostDetails/Tabs/ErrataTab.js
parthaa: For applicable errata you want to show `Apply Erratum` which takes em to errata page.   
For installable you want to show `Apply vi Katello agent` etc.

jeremylenz: Ahhh gotcha.

jturel (approved):
ACK from code perspective. Looks great and the updates appear to map to the meeting we had last week all while maintaining backward compatibility with the existing REX template. That's awesome!

webpack/components/extensions/HostDetails/Tabs/ErrataTab.js
```diff
@@ -104,38 +106,92 @@ export const ErrataTab = () => {
...
+    selectNone();
+    dispatch(applyViaKatelloAgent(hostId, { search: selected }));
+  };
+
+  const applyErratumViaKatelloAgent = id => dispatch(applyViaKatelloAgent(
+    hostId,
+    { errata_ids: [id] },
+  ));
+
+  const apply = () => (contentFacet.remoteExecutionByDefault ? applyViaRemoteExecution() :
```
jeremylenz, line 147: The Apply button uses different logic to decide whether to use Katello agent, compared to the dropdown items.  This uses
```js
 contentFacet.remoteExecutionByDefault
```
whereas the dropdowns use
```js
contentFacet.katelloAgentInstalled && contentFacet.katelloAgentEnabled
```

Was this intended? In my environment, the Apply button tries to use Katello agent, but I don't have any Katello agent actions in the dropdowns. This could be confusing.

webpack/components/extensions/HostDetails/Tabs/ErrataTab.js
```diff
@@ -104,38 +106,92 @@ export const ErrataTab = () => {
...
-  const applyByKatelloAgent = () => {
-    const selected = fetchBulkParams();
-    if (!isEmpty(selected)) {
-      const parameters = { bulk_errata_ids: JSON.stringify(selected) };
-      setIsBulkActionOpen(false);
-      selectNone();
-      dispatch(applyViaKatelloAgent(hostId, parameters));
-    }
-  };
+  if (contentFacet.katelloAgentInstalled && contentFacet.katelloAgentEnabled) {
```
jeremylenz, line 161:
```suggestion
  const katelloAgentAvailable = (contentFacet.katelloAgentInstalled && contentFacet.katelloAgentEnabled);
  if (katelloAgentAvailable) {
```

app/views/foreman/job_templates/install_errata.erb
```diff
@@ -2,14 +2,14 @@
...
 job_category: Katello
-description_format: 'Install errata %{errata}'
+description_format: 'Install errata %{Errata search query}'
 feature: katello_errata_install
 provider_type: SSH
 template_inputs:
-- name: errata
-  description: A comma separated list of errata to install
+- name: Errata search query
+  description: Filter criteria for Errata to be installed.
```
jeremylenz, line 10:
```suggestion
  description: Filter criteria for errata to be installed.
```

jeremylenz (commented):
Thanks for your patience @parthaa 

everything seems to work great! Just a couple Katello agent comments

webpack/components/extensions/HostDetails/Tabs/ErrataTab.js
parthaa: I guess we should just disable if `contentFacet.katelloAgentEnabled`  and ` contentFacet.remoteExecutionByDefault` are false

parthaa: Changed this condition a little. `(contentFacet.remoteExecutionByDefault || !katelloAgentAvailable)`
So if the thing cant be installed via katello agent it will still try rex (even if you have rexByDefault false).

jeremylenz (approved):
Changes LGTM

APJ :+1:

parthaa:
[test katello]

parthaa:
[test katello]

PR #9767, opened 2021-11-05
Fixes #33855 - Allow adding filter on same RPM name with different architectures

sjha4:
Allow adding filter on same RPM name with different architectures
Steps to Reproduce:
1. Create/ Use a CV
2. Create a Package filter on the CV
3. Add rule: name = test, architecture = x86_64, version = all version
4. Hit save
5. Add another rule: name = test, architecture = i686, version = all version
6. Hit save

User should be allowed to create 2 filter rules with diff architectures.


ianballou (approved):
Worked fine for me!

PR #9768, opened 2021-11-08
Fixes #33292 - modifications to enable ostree commit ref content uploads

jjeffers:

1. wget --no-parent -r https://fixtures.pulpproject.org/ostree/small/
2. tar --exclude="index.html" -cvf "fixtures_small_repo.tar" -C fixtures.pulpproject.org/ostree "small"

Using hammer:
```
```
After the import process completes, hammer will report the file was uploaded. Check for 2 things in Katello - 1. that the upload task completed (which includes both the upload and import actions), and 2. that the new content is present in the repository. The "small" repo will have 2 commit refs present with labels "stable" and "rawhide". 

There are additional options that can be passed to the content import process for ostree content (commit refs): `ref`, `repository_name`, and `parent_commit`. Repository name is *required*, although the default is typically "repo", so that can be used in most other examples.

Some more things to check:
```
Could not upload the content:
```

### DIFF
--- a/app/controllers/katello/api/v2/content_uploads_controller.rb
+++ b/app/controllers/katello/api/v2/content_uploads_controller.rb
@@ -10,7 +10,7 @@ module Katello
     param :repository_id, :number, :required => true, :desc => N_("repository id")
     param :size, :number, :required => true, :desc => N_("Size of file to upload")
     param :checksum, String, :required => false, :desc => N_("Checksum of file to upload")
-    param :content_type, RepositoryTypeManager.uploadable_content_types(false).map(&:label), :required => false, :desc => N_("content type ('deb', 'docker_manifest', 'file', 'ostree', 'rpm', 'srpm')")
+    param :content_type, RepositoryTypeManager.uploadable_content_types(false).map(&:label), :required => false, :desc => N_("content type ('deb', 'docker_manifest', 'file', 'ostree_ref', 'rpm', 'srpm')")
     def create
       fail Katello::Errors::InvalidRepositoryContent, _("Cannot upload Ansible collections.") if @repository.ansible_collection?
       content_type = params[:content_type] || ::Katello::RepositoryTypeManager.find(@repository.content_type)&.default_managed_content_type&.label
--- a/app/services/katello/pulp3/pulp_content_unit.rb
+++ b/app/services/katello/pulp3/pulp_content_unit.rb
@@ -116,24 +116,30 @@ module Katello
         content_unit_list page_opts
       end
 
+      # rubocop:disable Lint/UselessAssignment
       def self.find_duplicate_unit(repository, unit_type_id, file, checksum)
+        filter_label = 'sha256'
+        if unit_type_id == 'ostree_ref'
+          filter_label = 'checksum'
+        end
         content_backend_service = SmartProxy.pulp_primary.content_service(unit_type_id)
         duplicates_allowed = ::Katello::RepositoryTypeManager.find_content_type(unit_type_id).try(:duplicates_allowed)
         if repository.generic? && duplicates_allowed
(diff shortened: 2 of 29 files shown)


jlsherrill:
I took a stab at simplifying the workflow a bit:  https://github.com/jlsherrill/katello/commit/273f507d4c4087d0d7d0f4493ec2fe182febc6ba

after playing around with it, i realized that these workflows are quite different.  Duplicate uploads had to use duplicate artifacts and couldn't re-use existing content units (this didnt' error, but behaved really oddly).  So my changes account for this by introducing a new content type attribute to indicate that this is a 'repository import' upload, it then adds a separate upload orchestration action to orchestrate it in a different way.  Let me know what you think!

app/services/katello/pulp3/pulp_content_unit.rb
```diff
@@ -116,24 +116,30 @@ def self.fetch_content_list(page_opts)
         content_unit_list page_opts
       end
 
+      # rubocop:disable Lint/UselessAssignment
       def self.find_duplicate_unit(repository, unit_type_id, file, checksum)
+        filter_label = 'sha256'
+        if unit_type_id == 'ostree_ref'
```
rverdile, line 122: What if `filter_label` was one if the fields for the content type? Then we wouldn't need a conditional. And maybe it could be 'sha256' by default since that's what everything else uses.

jlsherrill: actually i meant to comment on this.  Given the current changes, i don't think this is needed.  we never actually want to re-use a content unit within the upload process, only an artifact (which we can still lookup via sha256).  So i don't think this code path will ever actually be hit.  thoughts @jjeffers ?

app/controllers/katello/api/v2/content_uploads_controller.rb
```diff
@@ -10,7 +10,7 @@ class Api::V2::ContentUploadsController < Api::V2::ApiController
     param :repository_id, :number, :required => true, :desc => N_("repository id")
     param :size, :number, :required => true, :desc => N_("Size of file to upload")
     param :checksum, String, :required => false, :desc => N_("Checksum of file to upload")
-    param :content_type, RepositoryTypeManager.uploadable_content_types(false).map(&:label), :required => false, :desc => N_("content type ('deb', 'docker_manifest', 'file', 'ostree', 'rpm', 'srpm')")
+    param :content_type, RepositoryTypeManager.uploadable_content_types(false).map(&:label), :required => false, :desc => N_("content type ('deb', 'docker_manifest', 'file', 'ostree_ref', 'rpm', 'srpm')")
```
rverdile, line 13: What if we used `uploadable_content_types` to create the description? Then other generic content types would automatically show up here.

jlsherrill: you just can't do that within an N_() block for translation, granted these values don't need to be translated :)

Maybe lets just leave out the 'cotnent type' text and just list the possible values.

rverdile:
I'm hitting this following the upload example: ArgumentError: Missing the required parameter 'upload_href' when calling UploadsApi.commit

Is there a patch I missed?

jlsherrill:
@jjeffers have you and @rverdile figured out his issue?

app/controllers/katello/api/v2/content_uploads_controller.rb
jlsherrill: @jjeffers i think we can remove this right?

jlsherrill: oops sorry ignore me ;)

rverdile:
I was able to reproduce the bug with a fresh database doing this:
1. sync https://fixtures.pulpproject.org/ostree/small
2. upload that to a new repo

I could not reproduce if I uploaded first without syncing. Perhaps once you upload successfully the bug isn't reproducible?

PR #9769, opened 2021-11-08
Fixes #33860 - Show 404 when cv is not found for versions API

sjha4:
Throw 404 when content view is not found for versions API
Go to:
katello/api/v2/content_view_versions?content_view_id=NaN&per_page=20&page=1


jeremylenz (approved):
Works as expected

ACK :+1:

PR #9771, opened 2021-11-08
Fixes #33858 - Truncate long env names

sjha4:
For really really long environment names, the labels in the new UI should be truncated.
Create an env with a really really really long name.
Publish and promote CV to that env. The versions page, CV index page, publish/promote pages etc should have the env name truncated..


Andrewgdewar (approved):
Works as advertised, now my environment name, "Really long environment name with multiple words" has a comfortable place in the world.

PR #9772, opened 2021-11-08
Fixes #33874 - UI for Connected ISS

jturel:

Adds a UI to the backend changes introduced by #8821 

I followed our typical react patterns (API middleware...) to implement the feature

See context of the overall feature here: https://github.com/Katello/katello/pull/8821

You should be able to execute the workflow outlined there but using the UI instead of any direct API calls / the script I shared in that PR. You can find the new fields in Content -> Subscriptions -> Manage Manifest

Most importantly: try to break the UI or get it or the CDN configuration into a weird state and let me know how I can improve it (improved error message, etc)

**TODO**
- [x] test coverage (waiting on a first pass of feedback)


parthaa:
Doesn't seem to save the content credential when I reload the page.

- Go to Cdn Configuration and pick  a credential, 
- Click update
- Click `Manage Manifest` again and notice that the credential is empty

parthaa (approved):
Works well ACK.

jturel:
Added some tests. @jeremylenz curious for your feedback on the tests especially (see latest commit) because I:

- Removed snapshot testing for ManageManifestModal
- Copied SimpleContentAccessTest over ManageManifestModalTest. SCA test was misnamed as it was testing the entire modal
- Added my new test as a react-testing-library test

My approach sound ... sound ?

webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
```diff
@@ -57,12 +64,46 @@ class ManageManifestModal extends Component {
...
+    this.setState({ cdn_url: event.target.value });
+  };
+
+  updateCdnUsername = (event) => {
+    this.setState({ cdn_username: event.target.value });
+  };
+
+  hidePasswordPlaceholder = (event) => {
+    const { target } = event;
+    target.value = this.state.cdn_password;
```
jeremylenz, line 77: Mutating the value of a DOM element controlled by React in this way can cause unexpected behavior, since React is the one that decides when it wants to render or update (or unmount) the element. I prefer making it a [controlled component](https://reactjs.org/docs/forms.html#controlled-components), in other words storing that FormControl's value in state (by giving `<FormControl>` a `value` prop) and updating the state instead.

jturel: That makes sense. I'll play around with it

PR #9775, opened 2021-11-09
Fixes #33893 - fix smart proxy sync with blank url

jlsherrill:
We stopped checking remotes to see if they need updates due to the issue around that check. (Thinks like password and certs are not provided by pulp so will always show as needing updates).  It seems the code for smart proxy syncing was comparing the remote config on the SP to the one that was supposed to be on the main pulp server.  This was incorrect, and would have always been returning true!  Except it seems that this completely errors if the repository has no url

Skip the remote comparison like we do for regular repo syncs

1) create some yum repository in Library without a url set
2) try to sync it to a smart proxy (either associating the SP to library or promoting to an LCE)


jjeffers (approved):
Reproduced the original condition then retested with latest updates - a yum repository with an empty url didn't cause a problem during a smart proxy sync.

PR #9776, opened 2021-11-09
Fixes #33838 - Make Generic Content Units Controller Apipie Compatible

rverdile:
- This adds the `katello/api/v2/content_units` (and other similar) route(s). 
- This route has a required `content_type` parameter to select which content type to return. 
- It also changes the apidoc such that the routes for `content_units`, `python_packages` and `ostree_refs` are all under the "Content Units" subheading.
Apipie was previously overwriting the GenericContentUnitController's resource with whichever route was defined last. This changes the controller so that the route is "content_units", with a param to specify the content type. The main motivation for this is to make viewing generic content units possible with hammer.

The `katello/api/v2/python_packages`, `katello/api/v2/ostree_refs`, and other related routes are still valid.
1. Sync a python and/or ostree repository.
2. Check these endpoints (and ostree_refs):
   - `katello/api/v2/python_packages/`
   - `katello/api/v2/python_packages/:id`
   - `katello/api/v2/repositories/:repository_id/python_packages`
   - `katello/api/v2/repositories/:repository_id/python_packages/:id`
3. Replace "python_packages" with "content_units" and pass the param `content_type=python_packages`. The results should be the same.
4. Check the Apidoc and see if the routes under "Content Units" are displaying properly.

### DIFF
--- a/app/controllers/katello/api/v2/generic_content_units_controller.rb
+++ b/app/controllers/katello/api/v2/generic_content_units_controller.rb
@@ -1,10 +1,15 @@
 module Katello
   class Api::V2::GenericContentUnitsController < Api::V2::ApiController
+    resource_description do
+      name 'Content Units'
+      param :content_type, String, desc: N_("Possible values: #{Katello::RepositoryTypeManager.generic_content_types.join(", ")}"), required: true
+    end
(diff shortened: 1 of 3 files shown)


rverdile:
[test katello]

jjeffers (approved):
Tested this change, looks good with python package content, etc. Nice job!

PR #9777, opened 2021-11-10
Fixes #33899 - Use Smart Proxy exposed rhsm_url setting

ekohl:

The smart_proxy_pulp plugin can expose a setting `rhsm_url` and Katello should prefer this over logic that tries to derive it.

Foreman Installer 3.1+ sets the rhsm_url setting in smart_proxy_pulp. This should be used if available. Compatibility with older proxies is still required.

This allows the installer to migrate the port to 443. Multi homing also becomes a possibility.

Install a Smart Proxy with a non-standard URL (don't forget to refresh features) and observe rhsm_url (used in the registration template) is reflected.

### DIFF
--- a/app/models/katello/concerns/smart_proxy_extensions.rb
+++ b/app/models/katello/concerns/smart_proxy_extensions.rb
@@ -443,7 +443,11 @@ module Katello
       end
 
       def rhsm_url
-        if pulp_primary?
+        # Since Foreman 3.1 this setting is set
+        if (rhsm_url = setting(SmartProxy::PULP3_FEATURE, 'rhsm_url'))
+          rhsm_url
+        # Compatibility fall back
+        elsif pulp_primary?
           "https://#{URI.parse(url).host}/rhsm"
         elsif pulp_mirror?
           "https://#{URI.parse(url).host}:8443/rhsm"


app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -443,7 +443,11 @@ def smart_proxy_service
       end
 
       def rhsm_url
-        if pulp_primary?
+        # Since Foreman 3.1 this setting is set
+        if (url = setting(SmartProxy::PULP3_FEATURE, 'rhsm_url'))
```
ehelms, line 447: Looks like this can resolve to nil and cause the tests to fail. While it might be good for some updated fixtures to reflect this, I think primarily this should not be allowed to resolve to nil and break.

ehelms: I think what is happening is you are shadowing the instance attribute with a local variable of the same name. The smart proxy has a `url` property, which this function is using. This if statement sets `url` to nil and then when this cascades down into the elsif blocks the value of url is nil and the `URI.parse` call fails:

```
irb(main):004:0> url = 'test';if (url = nil); puts 'hi'; else; puts url; end

=> nil
```

ekohl: Oh, that's a good point. I think this should be better then.

ianballou (commented):
Once difference in functionality  that I noticed was with the default URL.  Since I had no `rhsm_url` set, the default seemed to be `localhost`: https://github.com/theforeman/smart_proxy_pulp/blob/master/lib/smart_proxy_pulp_plugin/pulpcore_plugin.rb#L10

Should we continue defaulting to the smart proxy's url?

ekohl:
So that's why I suggested to not have a default in https://github.com/theforeman/smart_proxy_pulp/pull/31#discussion_r695847507. We can still do that.

ianballou:
As long as the `rhsm_url` doesn't end up being invalid after installation that's fine by me. I agree with this: "I suspect there are 0 users where https://localhost/rhsm is a valid value".

Should we make sure the installer is updated to have a proper rhsm_url (if it isn't) before merging this?

ekohl:
> Should we make sure the installer is updated to have a proper rhsm_url (if it isn't) before merging this?

This should take care of it:
https://github.com/theforeman/puppet-foreman_proxy_content/commit/57ac645b9ff24163cc8d37648dbdfe225b3f0271

ianballou:
> > Should we make sure the installer is updated to have a proper rhsm_url (if it isn't) before merging this?
> 
> This should take care of it: [theforeman/puppet-foreman_proxy_content@57ac645](https://github.com/theforeman/puppet-foreman_proxy_content/commit/57ac645b9ff24163cc8d37648dbdfe225b3f0271)

Okay, awesome. I'll ack the PR here but let's wait to merge until the proxy side is taken care of.

ianballou (approved):

ehelms:
What is the other PR this is blocked by?

ianballou:
Oops, for some reason I thought the commit Ewoud mentioned was an open PR. I'll merge then!

PR #9779, opened 2021-11-10
Refs #33073 - Add Repository Sets tab to new host details

jeremylenz:

* Add Repository Sets tab as per mockup
* Add actions menus on each table row
* Add toggle group

This was originally going to be just a basic table, but it was so easy to add the bookmarks and actions that I went ahead and did it :)

I changed the Show All / Limit to Environment a bit so it's hopefully more clear now.

repo_sets

#### repository_sets_controller#index
The index endpoint includes the params below. I updated the descriptions to reflect my understanding of what actually happens in the code:
```rb
param :content_access_mode_all, :bool, :desc => N_("Get all content available, not just that provided by subscriptions.")
    param :content_access_mode_env, :bool, :desc => N_("Limit content to just that available in the host's or activation key's content view version and lifecycle environment.")
```
The important point is that `content_access_mode_env` limits content based on _both_ content view _and_ lifecycle environment.

#### Legacy page
The legacy page has two checkboxes:
__Show All__ - Controls the `content_access_mode_all` param.  Not shown when in SCA mode.
__Limit to environment__ - Controls the `content_access_mode_env` param.

#### New page
In the new page, there is only a single toggle group.
toggle_group
This controls the `content_access_mode_env` param.

The `content_access_mode_all` param is no longer a user option.  It is `true` if in SCA mode, and `false` if not.

Based on the status of these two params, the banner text changes:

__When in Simple Content Access mode:__
Show all: `Showing all available repositories.`
Limit to environment: `Showing only repositories in the host's content view and lifecycle environment.`

__When in non-SCA mode:__
Show all: `Showing all repositories available through subscriptions.`
Limit to environment: `Showing repositories in the host's content view and lifecycle environment that are available through subscriptions.`

You'll want to test with a host in an SCA org and a non-SCA org.
* Test that all three actions work as expected (Override to Enabled, Override to disabled, Reset to default)
* Note how the banner text changes depending on SCA and toggle group state


MariSvirik:
@jeremylenz great work and also thorough description!

- Just use the **search bar** from here: https://github.com/theforeman/foreman/pull/8743#issuecomment-939957119 (I guess this is what we decided on in the end).
- Second inline alert should be inlined with the Repo sets table/content -that means the same length as the first inline alert. (No pattern for shrinking alerts in PF4 in order to establish alert hierarchy)
- "Repository Sets" in the tab should be capitalized "**Repository sets**"
- Are we going to add also a description from [here](https://marvelapp.com/prototype/995cc27/screen/80441609)?

Screenshot 2021-11-12 at 14 42 23

jeremylenz:
> Just use the search bar from here

I think this is easier said than done.  Katello has our search bar baked in to TableWrapper so we'd have to change it there.  Thoughts @Andrewgdewar @sjha4 ?

jeremylenz:
> Second inline alert should be inlined with the Repo sets table/content -that means the same length as the first inline alert. (No pattern for shrinking alerts in PF4 in order to establish alert hierarchy)

I think the alert width is determined by the width of the container element.  I didn't shrink it.  I'll see what I can do via CSS..

MariSvirik:
> > Just use the search bar from here
> 
> I think this is easier said than done. Katello has our search bar baked into TableWrapper so we'd have to change it there. 

Correct me if I'm wrong, but wasn't this solved via @Ron-Lavi's template?

jeremylenz:
> Correct me if I'm wrong, but wasn't this solved via @Ron-Lavi's template?

I think it was probably solved for Foreman but not Katello..

jeremylenz:
@MariSvirik Added the message that I forgot to add and fixed the capitalization.  Made some CSS changes and I think the alert is pretty close now:
css_changes

parthaa:
Mostly works great. Found this bug though

reposets
h

webpack/components/extensions/HostDetails/Tabs/RepositorySetsTab/RepositorySetsTab.js
```diff
@@ -66,14 +66,15 @@ const RepositorySetsTab = () => {
     content_facet_attributes: contentFacetAttributes,
   } = hostDetails;
   const contentFacet = propsToCamelCase(contentFacetAttributes ?? {});
-  const nonLibraryHost = !contentFacet.contentViewDefault &&
-  !contentFacet.lifecycleEnvironmentLibrary;
+  const nonLibraryHost = contentFacet.contentViewDefault === false &&
+    contentFacet.lifecycleEnvironmentLibrary === false;
```
jeremylenz, line 70: When the page is loading, these two values were `undefined` which was causing `alertShowing`'s default state to be incorrect. The result was that the alert would show even on a Library host if you reloaded the page.  This fixes that.

webpack/components/extensions/HostDetails/Tabs/RepositorySetsTab/RepositorySetsTab.js
```diff
@@ -225,6 +225,8 @@ const RepositorySetsTab = () => {
                 toggleGroup,
                 }
           }
+          activeFilters={[toggleGroupState]}
+          defaultFilters={[defaultToggleGroupState]}
```
jeremylenz, line 229: TableWrapper has [logic](https://github.com/jeremylenz/katello/blob/219c364525c0f99db389a52831682d7cd18a21d7/webpack/components/Table/TableWrapper.js#L94) that checks whether active filters have changed, and if so resets you to page 1. I neglected to add activeFilters here, so that was the source of the bug you found.

jeremylenz (commented):
Thanks @parthaa, updated

webpack/components/extensions/HostDetails/Tabs/RepositorySetsTab/RepositorySetsTab.js
jeremylenz: side note, Usually the `defaultFilters` are just hard-coded strings. But here they are dynamic, because the default toggle group state changes depending on if it's a library host or not.

parthaa (approved):
Works well. For extra credit I wonder though if you could add the name of the content view and lifecycle environment in 
`Info alert:Showing only repositories in the host's content view and lifecycle environment.`
Might help :) 

Going to approve anyway.

jeremylenz:
@parthaa Added CV & LCE name to the alert. Turned out to be a bit more involved because of translations, but oh well.

cool_alert


Also fixed one more bug (hope it's the last one) with the alert.

webpack/components/extensions/HostDetails/Tabs/RepositorySetsTab/RepositorySetsTab.js
```diff
@@ -0,0 +1,347 @@
...
+          buttonId="limit-to-env-toggle"
+          aria-label="Limit to environment"
+          isSelected={toggleGroupState === limitToEnvironment}
+          onChange={() => setToggleGroupState(limitToEnvironment)}
+        />
+      </ToggleGroup>
+    );
+  }
+
+  const hostEnvText = 'the "{contentViewName}" content view and "{lifecycleEnvironmentName}" environment';
```
parthaa, line 191: Is there a better way to do this, Feels liek we'd lose out on i18n

jeremylenz: @parthaa Check out the documentation for FormattedMessage from react-intl.  I think what I have will still be translated correctly.

parthaa: lgtm ack

PR #9780, opened 2021-11-11
Job template - Configure subman to new content source

stejskalleos:
New REX job template for configuring `subscription-manager` to new content source

Partial replacement for Bootstrap.py `--new-capsule` feature

Run the job and check that subman's configuration have been changed.

### DIFF
--- a/app/models/katello/concerns/smart_proxy_extensions.rb
+++ b/app/models/katello/concerns/smart_proxy_extensions.rb
@@ -446,13 +446,21 @@ module Katello
       def rhsm_url
         # Since Foreman 3.1 this setting is set
         if (rhsm_url = setting(SmartProxy::PULP3_FEATURE, 'rhsm_url'))
-          rhsm_url
+          URI(rhsm_url)
         # Compatibility fall back
         elsif pulp_primary?
-          "https://#{URI.parse(url).host}/rhsm"
+          URI("https://#{URI.parse(url).host}/rhsm")
         elsif pulp_mirror?
-          "https://#{URI.parse(url).host}:8443/rhsm"
+          URI("https://#{URI.parse(url).host}:8443/rhsm")
         end
+      end
+
+      def pulp_content_url
+        URI(setting(SmartProxy::PULP3_FEATURE, 'content_app_url'))
+      end
+
+      class ::SmartProxy::Jail < ::Safemode::Jail
+        allow :rhsm_url, :pulp_content_url
       end
     end
   end
--- a/app/views/foreman/job_templates/change_content_source.erb
+++ b/app/views/foreman/job_templates/change_content_source.erb
@@ -1,0 +1,42 @@
+<%#
+kind: job_template
+name: Change content source
+job_category: Katello
+model: JobTemplate
+provider_type: SSH
+description_format: Configure subscription manager to new content source
+feature: katello_change_content_source
+%>
+#!/bin/sh
+<%
+    content_source = @host.content_source
+-%>
+
+<% if content_source -%>
+SSL_CA_CERT=$(mktemp)
+cat << EOF > $SSL_CA_CERT
+<%= foreman_server_ca_cert %>
+EOF
+
+KATELLO_SERVER_CA_CERT=/etc/rhsm/ca/katello-server-ca.pem
+RHSM_CFG=/etc/rhsm/rhsm.conf
+
+# Prepare SSL certificate
+mkdir -p /etc/rhsm/ca
+cp -f $SSL_CA_CERT $KATELLO_SERVER_CA_CERT
+chmod 644 $KATELLO_SERVER_CA_CERT
+
+# Configure subscription-manager
+test -f $RHSM_CFG.bak || cp $RHSM_CFG $RHSM_CFG.bak
+
+subscription-manager config \
+  --server.hostname="<%= content_source.rhsm_url.host %>" \
+  --server.port="<%= content_source.rhsm_url.port %>" \
+  --server.prefix="<%= content_source.rhsm_url.path %>" \
+  --rhsm.repo_ca_cert="$KATELLO_SERVER_CA_CERT" \
+  --rhsm.baseurl="<%= content_source.pulp_content_url %>"
+
+<% else -%>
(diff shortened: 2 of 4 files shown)


app/views/foreman/job_templates/change_content_source.erb
```diff
@@ -0,0 +1,46 @@
...
+provider_type: SSH
+description_format: Configure subscription manager to new content source (smart proxy)
+feature: katello_change_content_source
+template_inputs:
+- name: smart_proxy
+  required: true
+  input_type: user
+  value_type: resource
+  resource_type: SmartProxy
+  description: New content source (Smart Proxy)
```
ares, line 15: This "Smart Proxy" may be hard to translate to Capsule, if possible, just avoid that string here.

app/views/foreman/job_templates/change_content_source.erb
```diff
@@ -0,0 +1,46 @@
+<%#
+kind: job_template
+name: Change content source
+job_category: Katello
+model: JobTemplate
+provider_type: SSH
+description_format: Configure subscription manager to new content source (smart proxy)
```
ares, line 7: same comment re smart proxy as below

app/views/foreman/job_templates/change_content_source.erb
```diff
@@ -0,0 +1,46 @@
...
+feature: katello_change_content_source
+template_inputs:
+- name: smart_proxy
```
ares, line 10: The name can be human friendly, e.g. "Smart Proxy", then below in the template you refer to it as <%= input("Smart Proxy") %>. The name is used in the form and users should not see UI with underscores.

Also I'd consider renaming it to Capsule (as mentioned below)

ekohl (changes requested):
As I asked for in the other PR: there should not be a content source job template parameter. Instead, it should only apply what's already in the Foreman DB. This means using `host.content_source` as a Smart Proxy.

stejskalleos:
> As I asked for in the other PR: there should not be a content source job template parameter. Instead, it should only apply what's already in the Foreman DB. This means using host.content_source as a Smart Proxy.

As I said in https://github.com/Katello/katello/pull/9756#discussion_r746633858, this change would make it impossible to reuse this template other one from https://github.com/Katello/katello/pull/9756.

ekohl:
And you can use snippets for that. A snippet is reusable from both jobs.

stejskalleos:
> And you can use snippets for that. A snippet is reusable from both jobs.

Yes, but you can do the same using `render_template` macro. If I change it to the snippet, we would have snippet and two templates. Now we have just two templates, no snippet, one file less.

app/views/foreman/job_templates/change_content_source.erb
```diff
@@ -0,0 +1,46 @@
...
+  advanced: false
+%>
+#!/bin/sh
+<%
+    smart_proxy = input_resource('Smart Proxy')
+    rhsm_url = smart_proxy.rhsm_url
+    pulp_content_url = smart_proxy.pulp_content_url
+-%>
+
+SSL_CA_CERT=$(mktemp)
```
ekohl, line 25: Why is a temporary file needed?

ekohl (commented):
> Now we have just two templates, no snippet, one file less.

That's also fine by me.

app/views/foreman/job_templates/change_content_source.erb
ares: I guess to avoid race condition and to auto-clean up after the use, we use the same concept in the registration if I'm not mistaken

ares:
I have no more comments, looks good to me :+1:

app/views/foreman/job_templates/change_content_source.erb
```diff
@@ -0,0 +1,46 @@
...
+description_format: Configure subscription manager to new content source
+feature: katello_change_content_source
+template_inputs:
+- name: Smart Proxy
+  required: true
+  input_type: user
+  value_type: resource
+  resource_type: SmartProxy
+  description: New content source
+  advanced: false
```
ekohl, line 16: I still insist we need a template without input sources and only use `host.content_source`.

ekohl (changes requested):

app/views/foreman/job_templates/change_content_source.erb
stejskalleos: What will happen if `@host` doesn't have any `content_source`? Where in the UI user can set the `content_source` for the host? I can't find it anywhere

ekohl: > What will happen if `@host` doesn't have any `content_source`?

That's a good point. I would check if the content URL and the RHSM URL are non-empty or exit with a failure. Either in the template or on the host. So either `raise Exception unless url` or `if [ -n "$url" ] ; then echo "No URL" ; exit 1 ; fi`.

The alternative would be to unconfigure subman but that feels illogical to me at this point. And it's a template so users can still modify it if they want to implement that.

> Where in the UI user can set the `content_source` for the host? I can't find it anywhere

In the Edit Host form on the `Host` tab there a `Content Source` field. That should configure it.

ehelms (approved):
@jlsherrill or @jturel mind taking a look

jturel:
Added a card to our board for review

stejskalleos:
Any updates, can we merge?

jturel:
Will review & test tomorrow

app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -446,14 +446,22 @@ def smart_proxy_service
       def rhsm_url
         # Since Foreman 3.1 this setting is set
         if (rhsm_url = setting(SmartProxy::PULP3_FEATURE, 'rhsm_url'))
-          rhsm_url
+          URI(rhsm_url)
```
jturel, line 449: I'm not sure of the impact overall of this change. Why was it necessary for this PR as my assumption this has been returning a string for some time

jturel: I see that you're using the URI object in the template. Got it

jturel (approved):
It worked for me! Thanks @stejskalleos

PR #9781, opened 2021-11-11
Fixes #33912 - content is not destroyed with force deleted repo

ianballou:

`::Katello::Content` entries were not being deleted due to the check for repositories tied to a root before destroying.  This PR further utilizes the `remove_from_content_view_versions` flag to force the content deletion as well.

Since the force delete option should remove all traces of the repo, it should be safe to delete the content before the `library_instances_inverse` repositories are all destroyed.

1) Create some repo and publish it in a CVV.
2) Create a content host registered to the new CVV.
3) Force destroy the new repo.
4) Run `subscription-manager refresh` on the host.
5) Check that the host no longer sees the deleted repo in `/etc/yum.repos.d/redhat.repo`.

### DIFF
--- a/app/lib/actions/katello/repository/destroy.rb
+++ b/app/lib/actions/katello/repository/destroy.rb
@@ -37,7 +37,7 @@ module Actions
               handle_redhat_content(repository) unless skip_environment_update
             else
               if destroy_content && !skip_environment_update
-                handle_custom_content(repository)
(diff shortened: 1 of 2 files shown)


ianballou:
[test katello]

jlsherrill (approved):
ACK, tested with a client and correctly saw the repo be removed

PR #9782, opened 2021-11-11
Fixes #33856 - Allow switching back to versions tab when polling

sjha4:
Looks up existing intervals before starting a new interval for polling tasks

Switching between tabs and returning to versions tab was re-registering the interval and causing the following error:
```
Uncaught Error: There is already an interval running and registered for: b360630d-cfe5-4b16-b91a-cb20d9bf47c3_POLL_TASK.
at registeredIntervalException (IntervalHelpers.js:4)
at IntervalMiddleware.js:17
at Object.dispatch (index.js:11)
at dispatch (<anonymous>:14608:80)
at ContentViewVersions.js:95
at ContentViewVersions.js:125
at buildRows (ContentViewVersions.js:116)
at ContentViewVersions.js:139
```

1. Add `sleep 60` to run method in katello/app/lib/actions/katello/content_view/publish.rb for easier testing.
2. Publish a cv and close publish modal and go to Versions tab. There should be active task polling.
3. Go to Repositories tab.
4. Come back to versions tab. The polling should be there as normal.
Without this change:
Blank page appears with console error as above.

### DIFF
--- a/webpack/scenes/ContentViews/Details/Versions/ContentViewVersions.js
+++ b/webpack/scenes/ContentViews/Details/Versions/ContentViewVersions.js
@@ -7,6 +7,7 @@ import { urlBuilder } from 'foremanReact/common/urlHelpers';
 import { STATUS } from 'foremanReact/constants';
 import { Link } from 'react-router-dom';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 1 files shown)


Andrewgdewar (approved):
ACK

PR #9796, opened 2021-11-18
Fixes #33951 - Don't show RH Repos page in disconnected mode

jeremylenz:

- [x] Don't show the Red Hat Repositories page in the nav when in disconnected mode
- [x] Update repository sets api to not provide available repository sets if disconnected

Since one does not have access to available repositories we can't really enable anything via this page.

I added a 403 response to `/repository_sets` with the explanatory error.  I chose 403 Forbidden because [the spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html) says this means "The server understood the request, but is refusing to fulfill it." I think that fits pretty well.

* Are there other endpoints I should also block, besides index?
* I decided not to change any wording etc. on the RH Repositories page, because it won't be displayed anyway

1. Settings > Content > Disconnected mode > Yes
2. Upon page refresh, verify that Red Hat Repositories is not listed in the navigation
3. Verify API error when trying to list repository sets (you can still go to /redhat_repositories manually to try it)
4. Turn the setting back to No and refresh the page; RH Repos should reappear in the nav

### DIFF
--- a/app/controllers/katello/api/v2/repository_sets_controller.rb
+++ b/app/controllers/katello/api/v2/repository_sets_controller.rb
@@ -33,6 +33,10 @@ module Katello
     param_group :search, Api::V2::ApiController
     add_scoped_search_description_for(Katello::ProductContent)
     def index
+      if Setting['content_disconnected']
+        respond_for_index(:collection => { :error => _("Repositories are not available for enablement while in disconnected mode.") }, :status => :forbidden)
+        return
+      end
       collection = scoped_search(index_relation, :name, :asc, :resource_class => Katello::ProductContent)
       collection[:results] = ProductContentFinder.wrap_with_overrides(
           product_contents: collection[:results],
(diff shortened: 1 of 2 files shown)


jlsherrill:
I assume we don't want to support the situation where:

1) a satellite is disconnected
2) they are syncing from locally available CDN source, such as previous iso content?  (i agree this use case may be a stretch, but wanted to bring it up)

Also how does this affect ISS?  Couldn't a satellite be disconnected (i.e. not able to talk to the portal), but still want to enable repos from another satellite?

jeremylenz:
@parthaa would also like your input on how this affects disconnected Satellites :)

parthaa:
> parthaa would also like your input on how this affects disconnected Satellites :)

@jlsherrill brings up a good point.
That being said I 'd think `Disconected` should really only apply for `Airgapped Disconnected` setups. If its talking to a different CDN I 'd consider that a `Connected` case

jlsherrill:
I feel like there are two aspects:
Disconnected subscriptions
Disconnected content

and we're trying to combine them both with one setting, which maybe isn't appropriate ?

jeremylenz:
Closing in favor of a future PR from Partha after the design discussion today.

PR #9585, opened 2021-08-26
Fixes #33361 - Initial stab at the host errata page

parthaa:

PR #9742, opened 2021-10-21
Fixes #33762 - Add toggle group for All/Installable  errata 

parthaa:
### What are the changes introduced in this pull request?
On the new host errata page
- Added toggle group with options for All / Installable
- The "All" option is what was previously known as "Applicable" errata
- Added an Installable column to indicate if the errata is installable or not "No"
- Action menu items (apply etc.) should be disabled for applicable (but not installable) errata
- Also ability to handle "unselectable" rows when using BulkSelect and SelectionSet hooks

### What are the testing steps for this pull request?

1. Create and Sync this repo -> https://partha.fedorapeople.org/test-repos/multi-errata/
2. Create another repo in the same product. Leave the feed empty
3. Create  a Dev lifecycle environment
4. Create a content view with the repos created in steps 1 and 2
5. Publish and Promote it to dev
6. Create an activation that points dev/cv. Make sure the subscriptions are turned on
7. Register a rhel7/centos7 with that activation key
8. On client `yum install armadillo-0.1-1.noarch`
9. Also `rpm -ivh https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/walrus-0.71-1.noarch.rpm https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/whale-0.2-1.noarch.rpm https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/shark-0.1-1.noarch.rpm https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/stork-0.12-2.noarch.rpm`
10. Now edit the repo in step 2 and add the feed url -> https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/
11. Sync this repo
12. Now you should have 38 Applicable Errata and 37 installable errata
13. Go to Hosts -> New Host Details -> Content -> Errata tab
14. Notice the new toggle group.
15. Choose all. Notice that the first check box must be disabled. You should be able hover on 'X' in the installable column and and notice a tooltip that explains why that is the case
16. Try all the select options , add search etc. 
17. Report any inconsistencies
18.
toggle

#### Backend Changes #####

- Added a new parameter to host_errata_controller to accept a `restrict_applicable` flag. Setting this to true will automatically include applicable errata in the index call
- Added a `selectable` attribute to `_metadata.json.rabl`. This will hold the count of selectable items in main list. We need this information for select all to correctly display how many items are going to be selected.
[...]

PR #9799, opened 2021-11-18
Fixes #33849 - Host Packages - basic table

lfu:
Host Packages basic table

Followed the host errata initial PR: https://github.com/Katello/katello/pull/9585

### DIFF
--- a/app/controllers/katello/api/v2/host_packages_controller.rb
+++ b/app/controllers/katello/api/v2/host_packages_controller.rb
@@ -20,10 +20,12 @@ module Katello
 
     api :GET, "/hosts/:host_id/packages", N_("List packages installed on the host")
     param :host_id, :number, :required => true, :desc => N_("ID of the host")
+    param :include_latest_upgradable, :boolean, :desc => N_("Also include the latest upgradable package version for each host package")
     param_group :search, Api::V2::ApiController
     add_scoped_search_description_for(Katello::InstalledPackage)
     def index
       collection = scoped_search(index_relation, :name, :asc, :resource_class => ::Katello::InstalledPackage)
+      collection[:results] = HostPackagePresenter.with_latest(collection[:results], @host) if ::Foreman::Cast.to_bool(params[:include_latest_upgradable])
       respond_for_index(:collection => collection)
     end
 
--- a/app/presenters/katello/host_package_presenter.rb
+++ b/app/presenters/katello/host_package_presenter.rb
@@ -1,0 +1,21 @@
+module Katello
+  class HostPackagePresenter < SimpleDelegator
+    attr_accessor :installed_package, :upgradable_version, :rpm_id
+
+    def initialize(installed_package, upgradable_version, rpm_id)
+      @installed_package = installed_package
+      @upgradable_version = upgradable_version
+      @rpm_id = rpm_id
+      super(@installed_package)
+    end
+
+    def self.with_latest(packages, host)
+      upgradable_packages_map = ::Katello::Rpm.installable_for_hosts([host]).select(:id, :name, :nvra, :evr).order(evr: :desc).all.group_by(&:name)
+      installed_packages_map = ::Katello::Rpm.where(nvra: packages.map(&:nvra)).select(:id, :name).group_by(&:name)
+
+      packages.map do |p|
+        HostPackagePresenter.new(p, upgradable_packages_map[p.name]&.first&.nvra, installed_packages_map[p.name]&.first&.id)
+      end
+    end
+  end
+end
--- a/app/views/katello/api/v2/host_packages/base.json.rabl
+++ b/app/views/katello/api/v2/host_packages/base.json.rabl
@@ -2,3 +2,5 @@ object @resource
 
 attributes :id, :name
 attributes :nvrea, :nvra
+attributes :upgradable_version
+attributes :rpm_id
--- a/webpack/components/extensions/HostDetails/Tabs/PackagesTab.js
+++ b/webpack/components/extensions/HostDetails/Tabs/PackagesTab.js
@@ -1,0 +1,127 @@
+import React, { useCallback, useState } from 'react';
+import { useSelector } from 'react-redux';
+import { Button, Hint, HintBody } from '@patternfly/react-core';
+import { TableVariant, Thead, Tbody, Tr, Th, Td } from '@patternfly/react-table';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { selectAPIResponse } from 'foremanReact/redux/API/APISelectors';
+
+import { urlBuilder } from 'foremanReact/common/urlHelpers';
+import TableWrapper from '../../../../components/Table/TableWrapper';
+import { PackagesStatus, PackagesLatestVersion } from '../../../../components/Packages';
+import { getInstalledPackagesWithLatest } from '../HostPackages/HostPackagesActions';
+import { selectHostPackagesStatus } from '../HostPackages/HostPackagesSelectors';
+import { HOST_PACKAGES_KEY } from '../HostPackages/HostPackagesConstants';
+import './PackagesTab.scss';
+
+export const PackagesTab = () => {
+  const hostDetails = useSelector(state => selectAPIResponse(state, 'HOST_DETAILS'));
+  const { id: hostId } = hostDetails;
+  const actionButtons = <Button isDisabled> {__('Upgrade')} </Button>;
+
+  const [searchQuery, updateSearchQuery] = useState('');
+
+  const emptyContentTitle = __('This host does not have any packages.');
+  const emptyContentBody = __('Packages will appear here when available.');
+  const emptySearchTitle = __('No matching packages found');
+  const emptySearchBody = __('Try changing your search settings.');
+  const columnHeaders = [
+    __('Package'),
+    __('Status'),
+    __('Installed Version'),
+    __('Upgradable To'),
+  ];
+
+  const fetchItems = useCallback(
+    params => (hostId ? getInstalledPackagesWithLatest(hostId, params) : null),
+    [hostId],
+  );
+
+  const response = useSelector(state => selectAPIResponse(state, HOST_PACKAGES_KEY));
+  const { results, ...metadata } = response;
+  const status = useSelector(state => selectHostPackagesStatus(state));
+  const rowActions = [
+    {
+      title: __('Upgrade via remote execution'), disabled: true,
+    },
+    {
+      title: __('Upgrade via customized remote execution'), disabled: true,
+    },
+  ];
+
+  return (
+    <div>
+      <div id="packages-hint">
+        <Hint>
+          <HintBody>
+            {__('Packages management functionality on this page is incomplete')}.
+            <br />
+            <Button component="a" variant="link" isInline href={urlBuilder(`content_hosts/${hostId}/packages/installed`, '')}>
+              {__('Visit the previous Packages page')}.
+            </Button>
+          </HintBody>
+        </Hint>
+      </div>
+      <div id="packages-tab">
+        <TableWrapper
+          {...{
+                metadata,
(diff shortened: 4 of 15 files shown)


lfu:
Screenshot from 2021-11-18 12-32-51

lfu:
Screenshot from 2021-11-18 12-38-45

lfu:
On the mock-up page, both installed version and latest upgradable version are displayed. 

This might be a challenge at this moment as there is no existing API endpoint that would provide both versions.

To get the latest upgradable version, a call to `Katello::Api::V2::PackagesController` was made by `Katello::Api::V2::HostPackagesController` to retrieve the required latest version. This attempt brings another issue caused by  pagination. As a result, a newer version of `python3-pip` package was not displayed in this [page](https://github.com/Katello/katello/pull/9799#issuecomment-973106269). 

Other than this issue, the performance may also get a hit as we need to manipulate every record of the installed packages to insert the latest version in the UI. The amount of installed packages might be huge.  

We might need to reconsider the layout of the new packages page.

Screenshot from 2021-11-18 12-47-20

app/controllers/katello/api/v2/host_packages_controller.rb
```diff
@@ -20,10 +20,12 @@ class Api::V2::HostPackagesController < Api::V2::ApiController
 
     api :GET, "/hosts/:host_id/packages", N_("List packages installed on the host")
     param :host_id, :number, :required => true, :desc => N_("ID of the host")
+    param :packages_restrict_latest, :boolean, :desc => N_("Return only the latest version of each package")
     param_group :search, Api::V2::ApiController
     add_scoped_search_description_for(Katello::InstalledPackage)
     def index
       collection = scoped_search(index_relation, :name, :asc, :resource_class => ::Katello::InstalledPackage)
+      collection[:results] = HostPackagePresenter.with_latest(collection[:results], latest_upgradables) if ::Foreman::Cast.to_bool(params[:packages_restrict_latest])
```
jturel, line 28: I'd recommend migrating this controller to work solely with your new presenter rather than two alternate paths (effectively).

app/controllers/katello/api/v2/host_packages_controller.rb
```diff
@@ -91,6 +93,11 @@ def resource_class
 
     private
 
+    def latest_upgradables
```
jturel, line 96: Curious about this approach of getting these packages through a controller vs the underlying ruby code? This is a new one for me!

parthaa: Very interesting. Wouldn't the following achieve the same?
``` ruby
    def latest_upgradables
      ::Katello::Rpm.installable_for_hosts([@host])
    end   
```

parthaa:
You could try something like with the `with_latest` method taking the host and computing the appropriate pairs

```diff
diff --git a/app/controllers/katello/api/v2/host_packages_controller.rb b/app/controllers/katello/api/v2/host_packages_controller.rb
index 439bedcc3a..a3a294a615 100644
--- a/app/controllers/katello/api/v2/host_packages_controller.rb
+++ b/app/controllers/katello/api/v2/host_packages_controller.rb
@@ -25,7 +25,7 @@ module Katello
     add_scoped_search_description_for(Katello::InstalledPackage)
     def index
       collection = scoped_search(index_relation, :name, :asc, :resource_class => ::Katello::InstalledPackage)
-      collection[:results] = HostPackagePresenter.with_latest(collection[:results], latest_upgradables) if ::Foreman::Cast.to_bool(params[:packages_restrict_latest])
+      collection[:results] = HostPackagePresenter.with_latest(collection[:results], @host) if ::Foreman::Cast.to_bool(params[:packages_restrict_latest])
       respond_for_index(:collection => collection)
     end
 
@@ -93,11 +93,6 @@ module Katello
 
     private
 
-    def latest_upgradables
-      res = Katello::Api::V2::PackagesController.dispatch(:index, request, response)
-      JSON.parse(res[2].body).dig("results")
-    end
-
     def find_editable_host_with_facet
       @host = resource_finder(::Host::Managed.authorized("edit_hosts"), params[:host_id])
       fail HttpErrors::NotFound, _("Couldn't find host with host id '%s'") % params[:host_id] if @host.nil?
diff --git a/app/presenters/katello/host_package_presenter.rb b/app/presenters/katello/host_package_presenter.rb
index 4733b03e24..fa636455ab 100644
--- a/app/presenters/katello/host_package_presenter.rb
+++ b/app/presenters/katello/host_package_presenter.rb
@@ -5,18 +5,18 @@ module Katello
     def initialize(installed_package, upgradables)
       @installed_package = installed_package
       @upgradable_packages = upgradables
-      @upgradableVersion = latest_upgradable_version
+      @upgradableVersion = upgradables[installed_package.name]&.first&.nvra
       super(@installed_package)
     end
   
-    def latest_upgradable_version
-      return '' if upgradable_packages.blank?
-      found = upgradable_packages.find { |p| p["name"] == installed_package.name }
-      found.nil? ? '' : found["nvrea"]
-    end
-
-    def self.with_latest(packages, latest_packages)
-      packages.map { |p| HostPackagePresenter.new(p, latest_packages) }
+    def self.with_latest(installed_packages, host)
+      upgradable_packages_map = ::Katello::Rpm.
+                                installable_for_hosts([host]).
+                                select(:name, :nvra, :evr).
+                                order(evr: :desc).
+                                all.
+                                group_by(&:name)
+      installed_packages.map { |p| HostPackagePresenter.new(p, upgradable_packages_map) }
     end
   end
 end
```

app/presenters/katello/host_package_presenter.rb
```diff
@@ -0,0 +1,22 @@
+module Katello
+  class HostPackagePresenter < SimpleDelegator
+    attr_accessor :installed_package, :upgradable_packages, :upgradableVersion
```
parthaa, line 3:
```suggestion
    attr_accessor :installed_package, :upgradable_packages, :upgradable_version
```

app/views/katello/api/v2/host_packages/base.json.rabl
```diff
@@ -2,3 +2,4 @@ object @resource
 
 attributes :id, :name
 attributes :nvrea, :nvra
+attributes :upgradableVersion
```
parthaa, line 5:
```suggestion
attributes :upgradable_version
```

app/presenters/katello/host_package_presenter.rb
```diff
@@ -0,0 +1,21 @@
...
+    end
+
+    def self.with_latest(packages, host)
+      upgradable_packages_map = ::Katello::Rpm.
+                                installable_for_hosts([host]).
+                                select(:name, :nvra, :evr).
+                                order(evr: :desc).
+                                all.
+                                group_by(&:name)
+      packages.map { |p| HostPackagePresenter.new(p, upgradable_packages_map) }
```
jturel, line 18: It seems unnecessary that every instance of HostPackagePresenter needs the entire upgradeable_packages_map - what do you think about moving the lookup into this method and passing just the upgradeable_package itself as the parameter?

app/controllers/katello/api/v2/host_packages_controller.rb
```diff
@@ -20,10 +20,12 @@ class Api::V2::HostPackagesController < Api::V2::ApiController
 
     api :GET, "/hosts/:host_id/packages", N_("List packages installed on the host")
     param :host_id, :number, :required => true, :desc => N_("ID of the host")
+    param :packages_restrict_latest, :boolean, :desc => N_("Return only the latest version of each package")
```
jturel, line 23: Is this parameter correctly named? It seems like less of a restriction but more of an "also include this extra data"

jturel (commented):
@lfu this looks excellent!! Left some minor input :)

app/controllers/katello/api/v2/host_packages_controller.rb
```diff
@@ -20,10 +20,12 @@ class Api::V2::HostPackagesController < Api::V2::ApiController
...
     param_group :search, Api::V2::ApiController
     add_scoped_search_description_for(Katello::InstalledPackage)
     def index
       collection = scoped_search(index_relation, :name, :asc, :resource_class => ::Katello::InstalledPackage)
+      collection[:results] = HostPackagePresenter.with_latest(collection[:results], @host) if ::Foreman::Cast.to_bool(params[:packages_restrict_latest])
```
jturel, line 28: Would be good to have a controller test which uses your new param to verify the Presenter

app/controllers/katello/api/v2/host_packages_controller.rb
jeremylenz: I'm guessing this was to be consistent with other params; @parthaa and I had a discussion here: https://github.com/Katello/katello/pull/9742#discussion_r739527215

jturel: But I don't think the name or description matches what the param actually does. I prefer accuracy over consistency!

jeremylenz: I agree!

jturel: Pending meethig this morning: `include_latest_upgradable` - Also include the latest upgradable package version for each host package

webpack/components/extensions/HostDetails/Tabs/__tests__/packagesTab.test.js
```diff
@@ -0,0 +1,91 @@
...
+        response: {
+          id: 1,
+          content_facet_attributes: { ...facetAttributes },
+        },
+        status: 'RESOLVED',
+      },
+    },
+  },
+});
+
```
jturel, line 30: Nice test coverage!!!

jturel (changes requested):
Minor feedback. I'll test it all out once everything is addressed :)

parthaa:
Would recommend adding a host package presenter test along the lines of -> https://github.com/Katello/katello/blob/master/test/presenters/host_subscriptions_presenter_test.rb

webpack/components/extensions/HostDetails/Tabs/PackagesTab.js
```diff
@@ -0,0 +1,125 @@
...
+          </Thead>
+          <Tbody>
+            {results?.map((packages) => {
+                const {
+                  id,
+                  name: packageName,
+                  nvra: installedVersion,
+                  upgradable_version: upgradableVersion,
+                } = packages;
+                const newVersion = upgradableVersion || '-';
```
jeremylenz, line 99: To keep maria happy, I think this should use `<MinusIcon />` from `@patternfly/react-icons` :)

lfu: Done.

webpack/components/extensions/HostDetails/Tabs/PackagesTab.js
```diff
@@ -0,0 +1,123 @@
...
+            {results?.map((packages) => {
+                const {
+                  id,
+                  name: packageName,
+                  nvra: installedVersion,
+                } = packages;
+                return (
+                  <Tr key={`${id}`}>
+                    <Td>
+                      <a href={urlBuilder(`packages/${id}`, '')}>{packageName}</a>
```
jturel, line 101: From what I can tell these ids are the HostInstalledPackage ID rather than a Katello::Rpm. Remember that a HostInstalledPackage can possibly (and probably not uncommonly) have no corresponding Katello::Rpm. That is that case for me in my dev env cause I don't have any repos synced that my client host also has installed packages for. 

AFAIK the fix would be to 1) not have a link when there is no corresponding Katello::Rpm (Content -> Packages) for the given HostInstalledPackage and 2) Use the Katello::Rpm ID here in the URL rather than the HostInstalledPackage ID

jturel (approved):
ACK! Terrific work @lfu - thank you

PR #9800, opened 2021-11-18
Fixes #33859 - Remove Greedy DepSolving from UI

rverdile:
Removing UI option to choose greedy as the dependency solving option in Administer > Settings > Content >  Content View Dependency Solving Algorithm. It also drops `dependency_solving_algorithm` from Setting, which was used to store 'greedy' or 'conservative'

1. Migrate
2. Go to Administer > Settings > Content
3. greedy should no longer be an option for Content View Dependency Solving Algorithm
4. `Setting.all.find { |s| s.name === "dependency_solving_algorithm" }` in the console should return nil.

### DIFF
--- a/app/models/setting/content.rb
+++ b/app/models/setting/content.rb
@@ -13,7 +13,6 @@ class Setting::Content < Setting
     download_policies = proc { hashify_parameters(::Katello::RootRepository::DOWNLOAD_POLICIES) }
 
     proxy_download_policies = proc { hashify_parameters(::SmartProxy::DOWNLOAD_POLICIES) }
-    dependency_solving_options = proc { hashify_parameters(['conservative', 'greedy']) }
     cdn_ssl_versions = proc { hashify_parameters(Katello::Resources::CDN::SUPPORTED_SSL_VERSIONS) }
     http_proxy_select = [{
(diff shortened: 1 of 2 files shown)


app/models/setting/content.rb
```diff
@@ -127,8 +127,7 @@ def self.default_settings
                false, N_('Content View Dependency Solving Default')),
       self.set('dependency_solving_algorithm',
                N_("How the logic of solving dependencies in a Content View is managed. Conservative will only add " \
-               "packages to solve the dependencies if the package needed doesn't exist. Greedy will pull in the " \
-               "latest package to solve a dependency even if it already does exist in the repository."),
+               "packages to solve the dependencies if the package needed doesn't exist."),
                'conservative', N_('Content View Dependency Solving Algorithm'), nil,
                :collection => dependency_solving_options),
```
jlsherrill, line 132: I think we can just completely get rid of the setting altogether.  You might need a migration to delete the setting, need to test it without the setting defined

rverdile:
It looks like a migration was necessary, which I _think_ I did correctly. Also removed the setting entirely instead of just 'greedy'.

jlsherrill (approved):
ACK

PR #9801, opened 2021-11-19
Fixes #33959 - correct incorrect client_certificate setting

jlsherrill:
for pulp communication

fixing a typo in which setting to look for.  here's the setting that gets set: https://github.com/theforeman/puppet-foreman_proxy/blob/4775f3f3adb26892beac0e5ea75d88abb3c75be8/manifests/plugin/pulp.pp#L32
testing any pulp actions should be fine


jlsherrill:
[test katello]

PR #9803, opened 2021-11-22
Fixes #33966 - Capsule sync task failed to refresh repos

hao-yu:
This patch fixed the following 2 issues:
- Repositories in Capsule always get refreshed even their remotes
are not changed.
- Failed to refresh a repository in Capsule if its root.url is
empty, such as custom repository without feed url.

### DIFF
--- a/app/services/katello/pulp3/repository_mirror.rb
+++ b/app/services/katello/pulp3/repository_mirror.rb
@@ -38,8 +38,8 @@ module Katello
       def needs_updates?
         remote = fetch_remote
         return true if remote.blank?
-        options = repo_service.compute_remote_options
-        options.keys.any? { |key| remote.send(key) != options[key] }
+        options = compute_remote_options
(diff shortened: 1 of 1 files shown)


jlsherrill:
i believe this was actually already resolved by https://github.com/Katello/katello/pull/9775

hao-yu:
Yeah, I think #9775 just remove the check completely. Based on the description in the PR, "password and certs are not provided by pulp" seems to be partially correct only. I can see Pulp respond with certs except the key.  In my opinion, not providing key is probably ok because if key is changed, cert will also be changed.

jlsherrill:
yeah, i think that was the logic, to just always update the remote.  We do this for the primary pulp server because things like upstream auth, proxy auth, and certs cannot be checked, so we just always update the remote.  

I'll go ahead and close this, but let me know if you think we need to re-open it, or if the existing fix is not sufficient

PR #9804, opened 2021-11-22
Fixes #34104 - Add Errata overview card

chris1984:

With the new host details work being done and the errata work finished up, the next step is to add an errrata card to the host overview page.

Nothing major, we are returning our own custom legend that was made with the `ErrataSummary` function inside `Errata/index.js` We are using Piechart from Patternfly.

* Register a host either el7/el8
* Attach a custom sub for a product/repo that has errata
* Install some packages so there is applicable errata
* Enable the new host ui in the experimental settings
* goto hosts, new details page, look at the errata overview card.

### DIFF
--- a/webpack/components/Errata/index.js
+++ b/webpack/components/Errata/index.js
@@ -5,8 +5,65 @@ import {
   BugIcon,
   SecurityIcon,
   EnhancementIcon,
+  SquareIcon,
 } from '@patternfly/react-icons';
 import PropTypes from 'prop-types';
+import { urlBuilder } from 'foremanReact/common/urlHelpers';
+
+export const ErrataMapper = ({ data, id }) => data.map(({ x: type, y: count }) => <ErrataSummary count={count} type={type} key={`${count} ${type}`} id={id} />);
+
+export const ErrataSummary = ({ type, count, id }) => {
+  let ErrataIcon;
+  let label;
+  let name;
+  let url;
+  let color;
+  switch (type) {
+    case 'security':
+      label = __('Security');
+      ErrataIcon = SecurityIcon;
+      name = 'security advisories';
+      color = '#0066cc';
+      url = <a href={urlBuilder(`content_hosts/${id}/errata`, '')}> {count} {name} </a>;
+      break;
+    case 'recommended':
+    case 'bugfix':
+      label = __('Bugfix');
+      ErrataIcon = BugIcon;
+      name = 'bug fixes';
+      color = '#8bc1f7';
+      url = <a href={urlBuilder(`content_hosts/${id}/errata`, '')}> {count} {name} </a>;
+      break;
+    case 'enhancement':
+    case 'optional':
+      label = __('Enhancement');
+      ErrataIcon = EnhancementIcon;
+      name = 'enhancements';
+      color = '#002f5d';
+      url = <a href={urlBuilder(`content_hosts/${id}/errata`, '')}> {count} {name} </a>;
+      break;
+    default:
+  }
+  if (!ErrataIcon) return null;
+
+  return (
+    <span style={{ whiteSpace: 'nowrap' }}>
+      <TableText wrapModifier="nowrap">
+        <SquareIcon size="sm" color={color} />
+        <span style={{ marginLeft: '8px' }}>
+          <ErrataIcon title={label} />
+          {url}
+        </span>
+      </TableText>
+    </span>
+  );
+};
+
+ErrataSummary.propTypes = {
+  type: PropTypes.string.isRequired,
+  count: PropTypes.number.isRequired,
+  id: PropTypes.number.isRequired,
+};
 
 export const ErrataType = ({ type }) => {
   let ErrataIcon;
@@ -29,6 +86,7 @@ export const ErrataType = ({ type }) => {
     default:
   }
   if (!ErrataIcon) return null;
+
 
   return (
     <TableText wrapModifier="nowrap">
--- a/webpack/components/extensions/HostDetails/Cards/ErrataOverviewCard.js
+++ b/webpack/components/extensions/HostDetails/Cards/ErrataOverviewCard.js
@@ -1,0 +1,109 @@
+import React from 'react';
+import {
+  Card,
+  CardHeader,
+  CardTitle,
+  CardBody,
+  Flex,
+  FlexItem,
+  GridItem,
+} from '@patternfly/react-core';
+import { urlBuilder } from 'foremanReact/common/urlHelpers';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { propsToCamelCase } from 'foremanReact/common/helpers';
+import PropTypes from 'prop-types';
+import { ChartPie } from '@patternfly/react-charts';
+import { ErrataMapper } from '../../../../components/Errata';
+
+function HostInstallableErrata({
+  id, errataCounts,
+}) {
+  const errataTotal = errataCounts.total;
+  const errataSecurity = errataCounts.security;
+  const errataBug = errataCounts.bugfix;
+  const errataEnhance = errataCounts.enhancement;
+  const chartData = [{
+    w: 'security advisories', x: 'security', y: errataSecurity, z: errataTotal,
+  }, {
+    w: 'bug fixes', x: 'bugfix', y: errataBug, z: errataTotal,
+  }, {
+    w: 'enhancements', x: 'enhancement', y: errataEnhance, z: errataTotal,
+  }];
+  return (
+    <GridItem rowSpan={2} md={3} lg={4}>
+      <Card isHoverable>
+        <CardHeader>
+          <CardTitle>{__('Installable Errata')}</CardTitle>
+        </CardHeader>
+        <CardBody>
+          <Flex direction="column">
+            <FlexItem>
+              <a href={urlBuilder(`content_hosts/${id}/errata`, '')}>
+                {errataTotal} errata
+              </a>
+            </FlexItem>
+            <Flex flexWrap={{ xl: 'nowrap' }} direction="row" alignItems={{ default: 'alignItemsCenter' }}>
+              <div className="piechart-overflow" style={{ overflow: 'visible', minWidth: '140px', maxHeight: '155px' }}>
+                <div className="erratachart" style={{ minWidth: '300px', minHeight: '300px' }}>
+                  <ChartPie
+                    ariaDesc="errataChart"
+                    data={chartData}
+                    constrainToVisibleArea
+                    labels={({ datum }) => `${datum.y} ${datum.w}`}
+                    padding={{
+                      bottom: 20,
+                      left: 20,
+                      right: 140,
+                      top: 20,
+                    }}
+                    width={250}
+                    height={130}
+                  />
+                </div>
+              </div>
+              <div className="erratalegend">
+                <FlexItem>
+                  <ErrataMapper data={chartData} id={id} />
+                </FlexItem>
+              </div>
+            </Flex>
+          </Flex>
+        </CardBody>
+      </Card>
+    </GridItem>
+  );
+}
+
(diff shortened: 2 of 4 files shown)


webpack/components/extensions/HostDetails/Cards/ErrataOverviewCard.js
```diff
@@ -0,0 +1,101 @@
...
+          <CardTitle>{__('Installable Errata')}</CardTitle>
+        </CardHeader>
+        <CardBody>
+          <Flex direction="column">
+            <FlexItem>
+              <a href={urlBuilder(`content_hosts/${id}/errata`, '')}>
+                {errataTotal} errata
+              </a>
+            </FlexItem>
+            <Flex flexWrap={{ lg: 'nowrap' }} direction="row">
```
jeremylenz, line 39: I found that adding `align-items: center;` to the flex container will fix the vertical spacing.  I think giving the PF `<Flex>` the following props will accomplish that:
```suggestion
            <Flex flexWrap={{ lg: 'nowrap' }} direction="row" alignItems={{ default: 'alignItemsCenter' }}>
```

But yeah, the CSS to remember is, on the flex container `align-items: center;` :)

webpack/global_index.js
```diff
@@ -23,3 +24,4 @@ addGlobalFill('host-details-page-tabs', 'Content', <ContentTab key="content" />,
 addGlobalFill('host-details-page-tabs', 'Traces', <TracesTab key="traces" />, 100);
 
 addGlobalFill('details-cards', 'Content View Details', <ContentViewDetailsCard key="content-view-details" />);
+addGlobalFill('details-cards', 'Content View Details', <ErrataOverviewCard key="errata-overview" />);
```
jeremylenz, line 27: If the name isn't unique apparently it erases the previous card.

```suggestion
addGlobalFill('details-cards', 'Installable errata', <ErrataOverviewCard key="errata-overview" />);
```

webpack/components/Errata/index.js
```diff
@@ -5,8 +5,62 @@ import {
...
+  let name;
+  let url;
+  let color;
+  switch (type) {
+    case 'security':
+      label = __('Security');
+      ErrataIcon = SecurityIcon;
+      name = 'security advisories';
+      color = '#0066cc';
+      url = <a href={urlBuilder(`content_hosts/${id}/errata`, '')}> {count} {name} </a>;
```
jeremylenz, line 27: I'd say let's link to the new page for now. Since we're staying in React land, we can
```js
import { Link } from 'react-router-dom';
```
and then
```suggestion
      url = <Link to={urlBuilder(`Content/Errata`, '')}> {count} {name} </Link>;
```
Apparently the host ID isn't needed in the url with `<Link>` because it only changes the part of the url after the `#`. Added bonus!

webpack/components/extensions/HostDetails/Cards/ErrataOverviewCard.js
```diff
@@ -0,0 +1,101 @@
...
+          </Flex>
+        </CardBody>
+      </Card>
+    </GridItem>
+  );
+}
+
+const ErrataOverviewCard = ({ hostDetails }) => {
+  if (hostDetails.content_facet_attributes) {
+    const hostId = hostDetails.id;
```
jeremylenz, line 72: Destructuring tips :)

If we have 
```js
const id = hostDetails.id;
```
we know we can destructure that like
```js
const { id } = hostDetails;
```

But if we want to rename our variable at the same time (like you do here), we can still use destructuring:

```suggestion
    const { id: hostId } = hostDetails;
```
This is the destructuring equivalent of what you wrote.

jeremylenz (commented):
Great work @chris1984!

Left some comments below.

webpack/components/extensions/HostDetails/Cards/ErrataOverviewCard.js
chris1984: One day I will get this Destructuring, I just need to take a few functions and practice with it.

chris1984:
@jeremylenz here is the code so far for the card, if you want to look at the css. Also do you have a good starting point for the tests or tell me what i'm doing wrong in the small one i made up?

webpack/components/extensions/HostDetails/Cards/__tests__/errataOverviewCard.test.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import { render } from 'react-testing-lib-wrapper';
+import ErrataOverviewCard from '../ErrataOverviewCard';
+
+test('does not show piechart when content facet is not set', () => {
+  const { getByLabelText } = render(<ErrataOverviewCard />);
+  expect(getByLabelText('errataChart').not.toBeInTheDocument());
```
jeremylenz, line 7:
```suggestion
  expect(getByLabelText('errataChart')).not.toBeInTheDocument();
```
you just need to move the closing parentheses :)

Also, if you use `.not` then I prefer `queryByLabelText` over `getByLabelText`.  `get` will error if it can't find the element, whereas `query` will just return `undefined`.  So if you use `get` here, the `not` will never be reached.

chris1984: Thanks for the pointer, it got me started in the right direction and I learned something about the testing functions :)

chris1984:
Updated with tests

webpack/components/extensions/HostDetails/Cards/ErrataOverviewCard.js
```diff
@@ -0,0 +1,109 @@
...
+  const errataEnhance = errataCounts.enhancement;
+  const chartData = [{
+    w: 'security advisories', x: 'security', y: errataSecurity, z: errataTotal,
+  }, {
+    w: 'bug fixes', x: 'bugfix', y: errataBug, z: errataTotal,
+  }, {
+    w: 'enhancements', x: 'enhancement', y: errataEnhance, z: errataTotal,
+  }];
+  return (
+    <GridItem rowSpan={2} md={3} lg={3}>
```
jeremylenz, line 33:
```suggestion
    <GridItem rowSpan={2} md={3} lg={4}>
```

webpack/components/extensions/HostDetails/Cards/ErrataOverviewCard.js
```diff
@@ -0,0 +1,109 @@
...
+        <CardBody>
+          <Flex direction="column">
+            <FlexItem>
+              <a href={urlBuilder(`content_hosts/${id}/errata`, '')}>
+                {errataTotal} errata
+              </a>
+            </FlexItem>
+            <Flex flexWrap={{ lg: 'nowrap' }} direction="row" alignItems={{ default: 'alignItemsCenter' }}>
+              <FlexItem>
+                <div className="erratachart" style={{ width: '350', marginLeft: '-10%' }}>
```
jeremylenz, line 47:
```suggestion
            <Flex flexWrap={{ xl: 'nowrap' }} direction="row" alignItems={{ default: 'alignItemsCenter' }}>
              <div className="piechart-overflow" style={{ overflow: 'visible', minWidth: '140px', maxHeight: '155px' }}>
                <div className="erratachart" style={{ minWidth: '300px', minHeight: '300px' }}>
```

webpack/components/extensions/HostDetails/Cards/ErrataOverviewCard.js
```diff
@@ -0,0 +1,109 @@
...
+                      left: 20,
+                      right: 140,
+                      top: 20,
+                    }}
+                    width={250}
+                    height={130}
+                  />
+                </div>
+              </FlexItem>
+              <div className="erratalegend" style={{ marginLeft: '-40%' }}>
```
jeremylenz, line 64:
```suggestion
              </div>
              <div className="erratalegend">
```

webpack/global_index.js
```diff
@@ -30,3 +31,4 @@ addGlobalFill(
   'Content View Details',
   <ContentViewDetailsCard key="content-view-details" />,
 );
+addGlobalFill('details-cards', 'Installable errata', <ErrataOverviewCard key="errata-overview" />);
```
jeremylenz, line 34:
```suggestion
  <ContentViewDetailsCard key="content-view-details" />,
  2000,
);
addGlobalFill('details-cards', 'Installable errata', <ErrataOverviewCard key="errata-overview" />, 1900);
```
Adding these weights will put our cards ahead of Recent Jobs. thoughts?

jeremylenz (commented):
okay, I think I arrived at a CSS solution that looks okay.  I've added all the code changes I made as suggestions.

The main problem was that the pie chart content is approx. 140px wide, but comes in a container that is 300px wide.  But when you size down the container, the pie chart is also sized down. I couldn't find any CSS width properties causing this percentage, so my theory is that it's just the way the SVG is drawn.  (They're probably counting on having their legend there, which we're not using.)  This ridiculous size of the container is why we were having to use negative margins to move the legend over.

The solution I found was to put the gigantic pie chart container in _another_ container div, and then limit the size of that. That way, we don't have to choose between having a good-sized pie chart and having the legend positioning behave as you would expect; we get both.  So here are the changes I made: 

1. Replace the `erratachart` FlexItem with a regular `<div>` so we can style it. This will be our overflow container so that the pie chart inside it won't affect the position of the legend.
2. Add three styles to this div:
* `overflow: visible;` - So that the tooltips don't get cut off.
* `min-width: 140px;` - This is the width of the container; the horizontal position of the legend will be determined by this.
* `max-height: 155px;` - Since the flexbox container is set to `align-items: center;`, the height of the tallest element (in our case, this one) will determine the vertical positioning of the legend.  Adding max-height here limits the height of the container so that the legend is visually vertically centered on the pie chart.
3. For the inner `erratachart` container, add a `min-width` and `min-height` of 300px. This determines the actual visual size of the pie chart.
4. Remove the negative margins since we don't need them anymore.
4. Increase card width to 4 for `lg` screens so the legend text doesn't overflow.

I think this looks good both on `xl` and `l` screens. The legend wraps nicely to the second row when the screen isn't wide enough.

webpack/global_index.js
jeremylenz: update: see also https://github.com/theforeman/foreman/pull/8995

webpack/components/extensions/HostDetails/Cards/ErrataOverviewCard.js
```diff
@@ -0,0 +1,109 @@
...
+              </a>
+            </FlexItem>
+            <Flex flexWrap={{ lg: 'nowrap' }} direction="row" alignItems={{ default: 'alignItemsCenter' }}>
+              <FlexItem>
+                <div className="erratachart" style={{ width: '350', marginLeft: '-10%' }}>
+                  <ChartPie
+                    ariaDesc="errataChart"
+                    data={chartData}
+                    constrainToVisibleArea
+                    labels={({ datum }) => `${datum.y} ${datum.w} of ${datum.z}`}
```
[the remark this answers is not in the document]
chris1984: I was just going off the mockup, I will remove this

chris1984:
@jeremylenz updated

jeremylenz (approved):
Works well, thanks for sticking with it @chris1984 

One thing to consider is it looks like PF recently added a [new doc](http://patternfly-react.surge.sh/charts/legend-chart/#legend-links) for adding links to legends, as brought to our attention by Dan Labrecque.  Not sure if this will also cover our use case including errata icons, but may be worth looking into.

Totally up to you whether you want to try to refactor this, or just leave as-is and merge. But either way, wanted to document here that we should have another look at that PF documentation the next time we need a pie chart :)

chris1984:
@jeremylenz 

I am fine with keeping this as is since we both put a lot of work into it making sure it sizes correctly with different resolutions and getting the icons to show up as well. I also would like to get this into 7.0 and unblock your PR you brought up.

I think going forward like you said when we need to do another task that involves charts we use the docs and go that route.

PR #9805, opened 2021-11-22
Fixes #33685 - null generic remote options not updated correctly

rverdile:
Entering a blank remote option for python repositories in the repository update/details UI will now successfully clear the param in pulp. Presumably, this bug could occur for any content type using the generic framework.
This revealed two issues:
1. Params with array types need to be assigned `[]` instead of `null` when entered as blank in the repository details UI, otherwise pulp returns an error.
2. Previously, any generic remote options being updated with `null` were not actually being updated, because they were being filtered out: https://github.com/Katello/katello/blob/master/app/services/katello/pulp3/repository/generic.rb#L27

I can't remember why nil options were being filtered out, but I don't see any reason.
1. Create a python repo with `https://pypi.org` as the remote url (or anything, you don't need to sync to test this). Also test without creating with the remote url and instead update it later.
2. Go to the details page and add a value to the excludes/includes/package_types option
    - update the remote url so that the remote refreshes
    - Check that this value is updated in the pulp database. Get href with: `Katello::Repository.find(id).remote_href`
3. Remove the value and check that is has been cleared successfully (in this case, empty list).
4. Check anything else you can think of related to null values and python remote options.

### DIFF
--- a/app/services/katello/pulp3/repository/generic.rb
+++ b/app/services/katello/pulp3/repository/generic.rb
@@ -24,7 +24,7 @@ module Katello
         def remote_options
           generic_remote_options = JSON.parse(root.generic_remote_options)
           if generic_remote_options.any?
-            common_remote_options.merge(generic_remote_options.select { |_, v| !v.nil? }).symbolize_keys
+            common_remote_options.merge(generic_remote_options).symbolize_keys
           else
             common_remote_options
           end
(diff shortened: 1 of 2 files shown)


jlsherrill (approved):
worked well!   saw excludes get updated from  ['django'] to   to [] in the pulp db

PR #9806, opened 2021-11-22
Fixes #33971 - CVV Import fails due to new Pulp 3.16 task return

ianballou:
This PR allows for the polling of Pulp tasks that return only a task group.  The import and the ACS refresh endpoints do this now.

https://docs.pulpproject.org/pulpcore/changes.html#removals

I tried to make the smallest change possible to allow for task groups only. It seems to be working fine since the BATS tests pass successfully.  I've been learning the tasking code as I go, so please let me know if you think there's a better way to do this.

1) Perform a content view version import in any way you see fit
2) Ensure it errors out
3) Patch in my PR here
4) Try another content view version import on a different organization
5) Ensure that the import completes successfully
6) Check that other tasks are not broken by these changes

### DIFF
--- a/app/lib/actions/pulp3/abstract_async_task.rb
+++ b/app/lib/actions/pulp3/abstract_async_task.rb
@@ -46,7 +46,7 @@ module Actions
       end
 
       def combined_tasks
-        return nil if pulp_tasks.nil? || task_groups.nil?
+        return nil if pulp_tasks.nil? && task_groups.nil?
         pulp_tasks + task_groups
       end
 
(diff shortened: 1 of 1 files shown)


ianballou:
This fixes the current BATS failures.

jlsherrill (approved):
ACK,  it would be nice if we could test this, but maybe when we get into the ACS work, that will be a lot easier

jlsherrill:
went ahead and merged to get the pipeline flowing

ianballou:
Yeah, ACS refresh should be a lot easier to test than Import.

PR #9807, opened 2021-11-23
Fixes #33978 - Show content views when deleting environment

sjha4:
1. Adds a confirmation modal on environment delete.
2. Shows content views tied to environment if any.

Inform users there are content views tied to environment for confirmation before delete. 
Users are still allowed to delete env and the content view is removed from the environment. However, user is not allowed to delete env tied to a content host. That validation is still in place.
Create a CV
Publish and promote it to an env.
Go to env and delete.
You should see a modal like below:
Screenshot from 2021-11-23 07-03-40

### DIFF
--- a/app/views/katello/api/v2/environments/show.json.rabl
+++ b/app/views/katello/api/v2/environments/show.json.rabl
@@ -46,4 +46,13 @@ node :permissions do |env|
   }
 end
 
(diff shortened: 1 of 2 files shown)


Andrewgdewar (approved):
LGTM

PR #9808, opened 2021-11-23
Fixes #33292 - ostree repo upload

jlsherrill:

1. wget --no-parent -r https://fixtures.pulpproject.org/ostree/small/
2. tar --exclude="index.html" -cvf "fixtures_small_repo.tar" -C fixtures.pulpproject.org/ostree "small"

Using hammer:
```
```
After the import process completes, hammer will report the file was uploaded. Check for 2 things in Katello - 1. that the upload task completed (which includes both the upload and import actions), and 2. that the new content is present in the repository. The "small" repo will have 2 commit refs present with labels "stable" and "rawhide". 

There are additional options that can be passed to the content import process for ostree content (commit refs): `ref`, `repository_name`, and `parent_commit`. Repository name is *required*, although the default is typically "repo", so that can be used in most other examples.

Some more things to check:
```
Could not upload the content:
```


jlsherrill:
[test katello]

rverdile (commented):
Fixes the sync then upload bug!

If you sync the "small" repository, upload that same content to it, then sync it again, it looks like this is what happens:
1. uploading replaces the content with the uploaded pulp objects
2. syncing again replaces the 2 units with new pulp objects, rather than using the ones from the original sync

I'm basing that mainly on the pulp hrefs being different. This is confusing so I'm not 100% I'm understanding it correctly :)

Is this expected?

jlsherrill:
I believe this is expected.  Its not really a normal use case to do both syncing and importing on the same repository refs they would get overwritten

jjeffers (approved):
ACK

PR #9811, opened 2021-11-23
Fixes #33919 - support on_demand repo cleaning

ianballou:
Adds support for:
1) Reclaiming space from a single repository
2) Reclaiming space from many repositories
3) Reclaiming space from an entire smart proxy

Reclaiming space means deleting any downloaded on-demand content in Pulp.  Only on-demand repositories are supported.

Most of it is straight-forward except for reclaiming space from a smart proxy. I integrated the space reclamation button into the existing sync dropdown.  I also used the same progress bar that is used during syncing.  I was considering just adding a separate button for reclaiming space, but I couldn't decide. Let me know what you think!

1) Create some on-demand and some immediate repos and sync at least the on-demand ones.
2) Download some on-demand RPMs from your repos' hosted contents
3) Clean contents via the UI (try bulk and single)
4) Check the Pulp task output to ensure the correct number of RPMs were cleaned up (look for "total: <num>" and "done: <num>")  
5) Sync a smart proxy and repeat steps 2-4 with the proxy.
  -> Clean the smart proxy by clicking the "Reclaim Space" button on the Synchronize drop-down.
6) Try cleaning only immediate repositories / smart proxies. It should fail with a warning.
7) Try cleaning a mix of immediate and on-demand repositories. It should succeed.

Questions:

~Would it be easy to know if the smart proxy's download policy in `capsule-content.controller.js`?  I suppose I could hide the clean option depending on how the smart proxy is configured.~
-> I fixed this by sneaking the download policy into the sync status view.

TODO:

- [x] Add tests
- [x] Add warning about custom repositories

### DIFF
--- a/app/services/katello/pulp3/api/core.rb
+++ b/app/services/katello/pulp3/api/core.rb
@@ -55,6 +55,10 @@ module Katello
           repository_type.distributions_api_class.new(api_client)
         end
 
+        def core_repositories_api
+          PulpcoreClient::RepositoriesApi.new(core_api_client)
+        end
+
         def repositories_api
           repository_type.repositories_api_class.new(api_client)
         end
@@ -79,6 +83,10 @@ module Katello
           self.class.ignore_409_exception do
             tasks_api.tasks_cancel(task_href, data)
           end
+        end
+
+        def repositories_reclaim_space_api
+          PulpcoreClient::RepositoriesReclaimSpaceApi.new(core_api_client)
         end
 
         def exporter_api
@@ -190,6 +198,12 @@ module Katello
 
         def delete_distribution(href)
           ignore_404_exception { distributions_api.delete(href) }
+        end
+
+        def core_repositories_list_all(options = {})
+          self.class.fetch_from_list do |page_opts|
+            core_repositories_api.list(page_opts.merge(options))
+          end
         end
 
         def list_all(options = {})
--- a/app/views/foreman/smart_proxies/_reclaim_space.html.erb
+++ b/app/views/foreman/smart_proxies/_reclaim_space.html.erb
@@ -1,0 +1,12 @@
+<div>
+  <h3><%= _('Reclaim Space') %></h3>
+  <p translate>
+    <strong>Warning</strong>: reclaiming space will delete all cached content units in "On Demand" repositories on this Smart Proxy.<br>
+    Take precaution when cleaning custom repositories whose upstream parents don't keep old package versions.
+  </p>
+  <div class="button">
+    <button class="btn btn-default" type="button" id="reclaimSpaceButton" ng-click="reclaimSpace()" translate>
+      Reclaim Space
+    </button>
+  </div>
+</div>
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/views/repository-details.html
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/views/repository-details.html
@@ -38,6 +38,19 @@
           </a>
         </li>
 
+        <li role="menuitem" ng-hide="hideSyncButton(repository, false)" ng-class="{disabled: disableSyncLink()}">
+          <a ng-click="openReclaimSpaceModal(repository)" disable-link="disableSyncLink()" translate>
+            Reclaim Space
+          </a>
+
+          <span class="disabled" ng-show="syncInProgress(repository.last_sync)" translate>
+            Cannot clean Repository, a sync is already in progress.
+          </span>
+
+          <span class="disabled" ng-show="denied('sync_products', product)" translate>
+            Cannot clean Repository without the proper permissions.
+          </span>
+        </li>
         <li role="menuitem" ng-hide="syncInProgress(repository.last_sync) || denied('edit_products', product)">
           <span class="disabled" ng-show="syncInProgress(repository.last_sync)" translate>
             Cannot republish Repository, a sync is already in progress.
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/product-repositories-reclaim-space-modal.controller.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/product-repositories-reclaim-space-modal.controller.js
@@ -1,0 +1,35 @@
+/**
+ * @ngdoc object
+ * @name  Bastion.repositories.controller:ProductRepositoriesReclaimSpaceModalController
+ *
+ * @requires $scope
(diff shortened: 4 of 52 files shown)


engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/views/repository-details.html
```diff
@@ -38,6 +38,13 @@ <h2 translate>{{ repository.name }}</h2>
           </a>
         </li>
 
+        <li role="menuitem" ng-hide="hideSyncButton(repository, false)" ng-class="{disabled: disableSyncLink()}">
+          <a ng-click="reclaimSpace(repository)" disable-link="disableSyncLink()" translate>
+            Reclaim Space
+          </a>
+        </li>
+
+
```
rverdile, line 47: There's an extra newline here

app/services/katello/pulp3/api/core.rb
```diff
@@ -55,6 +55,10 @@ def distributions_api
           repository_type.distributions_api_class.new(api_client)
         end
 
+        def untyped_repositories_api
```
rverdile, line 58: what does untyped mean?

rverdile (commented):
Cleaning with local repos is looking good. Still messing around with it to see if I hit some weird case. Not sure about smart proxies yet because spinning one up is being difficult.

To address your question, I think having reclaim space in the drop down instead of as a button would make sense if this is a niche feature.

rverdile (commented):
If you bulk select immediate and on-demand repos, it looks like this will reclaim space on the immediate repos as well. I saw total and done as 35 when I expected 0. 35 was the amount of packages in my immediate repo. Is this intended?

app/services/katello/pulp3/api/core.rb
ianballou: It's not specific to repository type like the `repositories_api`.  Perhaps I could call it `core_repositories_api` instead.

rverdile: that's what I was figured, I just wasn't sure. The name might be fine as is.

ianballou:
> If you bulk select immediate and on-demand repos, it looks like this will reclaim space on the immediate repos as well. I saw total and done as 35 when I expected 0. 35 was the amount of packages in my immediate repo. Is this intended?

There shouldn't be any affect on the immediate repos, but we should check. I might just add in a filter to only include the on demand repositories.

ianballou:
I just pushed up some draft-y code for adding a button to the pulp primary proxy, just FYI.

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/product-repositories-reclaim-space-modal.controller.js
```diff
@@ -0,0 +1,41 @@
...
+ * @requires $uibModalInstance
+ * @requires reclaimParams
+ *
+ * @description
+ *   A controller for the modal that warns about reclaiming on demand repository space
+ */
+angular.module('Bastion.repositories').controller('ProductRepositoriesReclaimSpaceModalController',
+    ['$scope', '$state', 'translate', 'Notification', 'RepositoryBulkAction', '$uibModalInstance', 'reclaimParams',
+        function ($scope, $state, translate, Notification, RepositoryBulkAction, $uibModalInstance, reclaimParams) {
+            /*var errorHandler = function errorHandler(response) {
```
ianballou, line 19: TODO: Delete me.

lib/katello/plugin.rb
```diff
@@ -246,6 +246,10 @@
       :name => _("Content Sync"),
       :partial => "foreman/smart_proxies/content_sync",
       :onlyif => proc { |proxy| proxy.pulp_mirror? }
+    context.add_pagelet :reclaim_space,
```
jlsherrill, line 249: this should be :details_content

i think this is describing which 'section' to add the pagelet to

rverdile (commented):
This is looking good! Not finding any issues the smart proxies. I guess I'll check back again when you're done pushing test updates.

ianballou:
This PR is ready for a full review now!

rverdile (approved):
Still looking functional, I can't find a case that breaks it. I think the UI choices you made look good. The ruby code looks good to me. 

**The angular code looks good to me too, but I know less about that so it's up to you if you want a second opinion there.**

Definitely doing what the issue is looking for, nice job!

app/services/katello/pulp3/api/core.rb
```diff
@@ -192,6 +200,12 @@ def delete_distribution(href)
           ignore_404_exception { distributions_api.delete(href) }
         end
 
+        def core_repositories_list_all(options = {})
+          self.class.fetch_from_list do |page_opts|
+            core_repositories_api.list(page_opts.merge(options))
+          end
+        end
```
jlsherrill, line 207: didn't  an issue get filed to make the cleanup default to all repositories?  did we open a katello issue to track that too?

app/views/foreman/smart_proxies/_reclaim_space.html.erb
```diff
@@ -0,0 +1,12 @@
+<div>
+  <h3><%= _('Reclaim Space') %></h3>
+  <p>
```
jlsherrill, line 3: i think this should be   <p translate>

app/views/foreman/smart_proxies/_reclaim_space.html.erb
```diff
@@ -0,0 +1,12 @@
...
+    <strong>Warning</strong>: reclaiming space will delete all cached content units in "On Demand" repositories on this Smart Proxy.<br>
+    Take precaution when cleaning custom repositories whose upstream parents don't keep old package versions.
+  </p>
+  <div class="button">
+    <button class="btn btn-default" type="button" id="reclaimSpaceButton" ng-click="reclaimSpace()">
```
jlsherrill, line 8: need a translate here too

app/services/katello/pulp3/api/core.rb
ianballou: What do you mean by default to all repositories? You can clean all repos for any smart proxy by doing it from their UI pages.

ianballou: Oh I remember what you mean now, the ability to not need to provide any repository hrefs.  I'll make an issue.

ianballou: Redmine: https://projects.theforeman.org/issues/34080

jlsherrill (approved):
angular changes look good to me!

PR #9814, opened 2021-11-24
Fixes #34003 - Add repo_type to pulp3_migration

m-bucher:
Usage:
```
repository_types='rpm,deb' foreman-rake katello:pulp3_migration
```

### DIFF
--- a/app/lib/actions/pulp3/content_migration.rb
+++ b/app/lib/actions/pulp3/content_migration.rb
@@ -5,13 +5,15 @@ module Actions
 
       def plan(smart_proxy, options)
         sequence do
-          action = plan_self(smart_proxy_id: smart_proxy.id)
+          action = plan_self(smart_proxy_id: smart_proxy.id, repository_types: options[:repository_types])
           plan_action(Actions::Pulp3::ImportMigration, options.merge(:dependency => action.output))
         end
       end
 
       def invoke_external_task
-        migration_service = ::Katello::Pulp3::Migration.new(smart_proxy)
+        options = {}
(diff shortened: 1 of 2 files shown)


app/lib/actions/pulp3/content_migration.rb
```diff
@@ -5,13 +5,15 @@ class ContentMigration < Pulp3::AbstractAsyncTask
...
+          action = plan_self(smart_proxy_id: smart_proxy.id, repository_types: options[:repository_types])
           plan_action(Actions::Pulp3::ImportMigration, options.merge(:dependency => action.output))
         end
       end
 
       def invoke_external_task
-        migration_service = ::Katello::Pulp3::Migration.new(smart_proxy)
+        options = {}
+        options[:repository_types] = input['repository_types'] unless input['repository_types'].nil?
+        migration_service = ::Katello::Pulp3::Migration.new(smart_proxy, options)
```
jlsherrill, line 16: If you don't specify the env variable, I believe this will cause options to include:  ` repository_types: nil  ` if they are not passed in, which would cause:

https://github.com/Katello/katello/blob/KATELLO-3.18/app/services/katello/pulp3/migration.rb#L35
to set the repository_types to nil instead of defaulting to `Migration.repository_types_for_migration`

jlsherrill:
[test katello]

jlsherrill:
[test katello]

jlsherrill:
@m-bucher are you still interested in this PR?

m-bucher:
I guess with migrations being stable enough it is kinda obsolete :thinking:

PR #9816, opened 2021-11-29
Refs #33874 - improve password field + use PF4

jturel:

Improves the behavior around the CDN Configuration password field by using the EditableTextInput component which has been expanded to support passwords. The main goal is for clearing a previously set (as far as the API is concerned) password for the cdn config to be easily understood & be straightforward

Migrated the form to PF4 and moving it into a separate component made that easier

Needs a regression test of the CDN configuration form.

Create a new org in your environment and adjust the cdn configuration UI to point to some other org, like Default_Organization. Use your katello CA cert as the content credential. Create it from the one on disk at /etc/pki/katello/certs/katello-default-ca


webpack/components/EditableTextInput/EditableTextInput.js
```diff
@@ -27,6 +38,18 @@ const EditableTextInput = ({
...
     setEditing(true);
     if (setCurrentAttribute) setCurrentAttribute(attribute);
+
+    if (isPassword) {
+      if (passwordPlaceholder) {
+        setPasswordPlaceholder(null);
+        setPasswordReset(true);
+      }
+
+      if (inputValue === null) {
```
jturel, line 48: I wonder if this piece should be moved to the onEdit handler function in CdnConfiguration? ie is this behavior specific to my use case here for CdnConfiguration, or should it be generalized for all password usage of this component?

webpack/scenes/Subscriptions/Manifest/__tests__/CdnConfiguration.test.js
```diff
@@ -0,0 +1,106 @@
...
+
+const cdnConfiguration = {
+  ssl_ca_credential_id: 2,
+};
+
+const defaultProps = {
+  cdnConfiguration: {
+
+  },
+};
```
jeremylenz, line 20: Shouldn't this match the required PropTypes shape?
```js
cdnConfiguration: PropTypes.shape({
    url: PropTypes.string,
    username: PropTypes.string,
    upstream_organization_label: PropTypes.string,
    ssl_ca_credential_id: PropTypes.number,
    password_exists: PropTypes.bool,
  }).isRequired,
```
Or if it doesn't matter, at least clean up a bit ;)
```suggestion
const defaultProps = {
  cdnConfiguration: {},
};
```

webpack/components/EditableTextInput/EditableTextInput.js
jeremylenz: I'd say probably yes, it should be moved. The only way it will be null is if the parent component supplies `value={null}` via the prop, so that's pretty specific.

jeremylenz (commented):
Looking great @jturel, I really like the way this came together!

I haven't taken it for an end-to-end spin yet, but I've played around with the React form and left some code comments.

The ManageManifestModal on my screen now requires scrolling, or else I can't see the close button at the bottom. Do you think it's worth it to consider putting CdnConfiguration in a new tab? I don't think the change would be too bad..

webpack/components/EditableTextInput/EditableTextInput.js
```diff
@@ -1,20 +1,31 @@
...
   component, currentAttribute, setCurrentAttribute, disabled,
 }) => {
   // Tracks input box state
   const [inputValue, setInputValue] = useState(value);
   const [editing, setEditing] = useState(false);
   const [submitting, setSubmitting] = useState(false);
-
+  const [passwordPlaceholder, setPasswordPlaceholder] = useState(hasPassword
+    ? PASSWORD_MASK : null);
+  const [passwordReset, setPasswordReset] = useState(false);
```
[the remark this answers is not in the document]
jturel: It's not strictly necessary - will remove

webpack/components/EditableTextInput/EditableTextInput.js
```diff
@@ -1,20 +1,31 @@
...
+  TimesIcon,
+  CheckIcon,
+  PencilAltIcon,
+} from '@patternfly/react-icons';
 import { translate as __ } from 'foremanReact/common/I18n';
 import PropTypes from 'prop-types';
 import Loading from '../Loading';
 import './editableTextInput.scss';
 
+const PASSWORD_MASK = '******';
```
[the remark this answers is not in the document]
jturel: Are you also requesting that the length of the mask also be increased from 6 to 8?

webpack/scenes/Subscriptions/Manifest/CdnConfiguration.js
```diff
@@ -0,0 +1,158 @@
...
+  } = props;
+
+  const dispatch = useDispatch();
+
+  const [url, setUrl] = useState(cdnConfiguration.url);
+  const [username, setUsername] = useState(cdnConfiguration.username);
+  const [password, setPassword] = useState(null);
+  const [organizationLabel, setOrganizationLabel] =
+    useState(cdnConfiguration.upstream_organization_label);
+  const [sslCaCredentialId, setSslCaCredentialId] = useState(cdnConfiguration.ssl_ca_credential_id);
```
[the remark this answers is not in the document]
jturel: I've renamed the component to CdnConfigurationForm which helps this concern - I think ?

webpack/components/EditableTextInput/EditableTextInput.js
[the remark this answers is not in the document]
jeremylenz: Can't hurt to encourage better password practices!

webpack/scenes/Subscriptions/Manifest/__tests__/CdnConfiguration.test.js
jturel: Not according to my tests - but I can add one

jturel:
@jeremylenz updated!

package.json
```diff
@@ -28,6 +28,7 @@
     "@sheerun/mutationobserver-shim": "^0.3.3",
     "@testing-library/jest-dom": "^5.3.0",
     "@testing-library/react": "^10.0.2",
+    "@testing-library/user-event": "^13.5.0",
```
jturel, line 31: Added for testing the select dropdown

https://testing-library.com/docs/ecosystem-user-event/#selectoptionselement-values-options

webpack/components/EditableTextInput/EditableTextInput.js
```diff
@@ -140,7 +132,7 @@ const EditableTextInput = ({
         </SplitItem>
         { isPassword ?
           <SplitItem>
-            <Button aria-label={`show-password ${attribute}`} variant="plain" isDisabled={inputValue?.length < 1} onClick={toggleShowPassword}>
+            <Button aria-label={`show-password ${attribute}`} variant="plain" isDisabled={!inputValue?.length > 0} onClick={toggleShowPassword}>
```
jeremylenz, line 135:
```js
isDisabled={!inputValue?.length > 0}
```
Do you mean `!(inputValue.length > 0)` or `!(inputValue.length) > 0`? I think without the parenthesis you're comparing `true` and `0`, which works (thanks JavaScript) but is too confusing for me.

Either
```js
!(inputValue?.length > 0)
```
or simply
```js
!inputValue.length
```
(since `0` is falsey in JS)
would make more sense to me.

webpack/components/EditableTextInput/EditableTextInput.js
```diff
@@ -64,10 +87,20 @@ const EditableTextInput = ({
...
+        setPasswordPlaceholder(PASSWORD_MASK);
+      }
+    }
+
     setInputValue(value);
     setEditing(false);
   };
 
+  const toggleShowPassword = () => {
+    setShowPassword(!showPassword);
```
jeremylenz, line 101:
```suggestion
    setShowPassword(prevPassword => !prevPassword);
```
When setting a state that depends on the previous state, I feel like using the function form of `setState` is safer because it guarantees you won't be operating on a stale value.

webpack/scenes/Subscriptions/Manifest/CdnConfigurationForm.js
```diff
@@ -0,0 +1,166 @@
...
+      password,
+      upstream_organization_label: organizationLabel,
+      ssl_ca_credential_id: sslCaCredentialId,
+    }));
+  };
+
+  const updateButtonContent = (
+    updatingCdnConfiguration ? (
+      <React.Fragment>
+        <Spinner size="md" />
```
jeremylenz, line 67: You can just add an `isLoading` prop to the `<Button>` component and get all this for free.  (and with better spacing between the spinner and text :wink: )

webpack/scenes/Subscriptions/Manifest/CdnConfigurationForm.js
```diff
@@ -0,0 +1,166 @@
...
+          <EditableTextInput
+            attribute="cdn-password"
+            value={password}
+            isPassword
+            hasPassword={hasPassword}
+            onEdit={editPassword}
+          />
+        </FormGroup>
+        <FormGroup
+          label={__('Organization Label')}
```
jeremylenz, line 110:
```suggestion
          label={__('Upstream Organization Label')}
```
Thoughts?  This would've saved me some confusion.

jeremylenz (commented):
Tested and works great. Some more comments below.

Extra credit: it might be nice to add front-end validations so the user doesn't try to submit with one of the three required fields blank (but of course, if _all_ of them are blank then it would have to show as valid.)  Check out PF's props list for FormGroup on [this page](https://patternfly-react.surge.sh/components/form). For the `validated` prop I'm thinking "success" if all 3 are filled in, "info" if 1-2 are filled in, and "default" if none are filled in and just the Red Hat config is being used.

webpack/scenes/Subscriptions/Manifest/CdnConfigurationForm.js
jturel: Mostly -1 on this because I think you got the word upstream from the api parameter. that is named with upstream_ because hammer would have gotten confused when doing this: `hammer organization update-cdn-configuration --organization-id=X ...`

Confusion may be getting added to the mix here due to testing this locally against your own Katello.  A user would never do that and the documentation will always reflect an external source, so "upstream" is implied

jturel:
Peek 2021-12-01 12-16


@jeremylenz thoughts on this validation? ^^^

jeremylenz:
@jturel looks good to me!

jturel:
@jeremylenz ty. updated!

webpack/scenes/Subscriptions/Manifest/CdnConfigurationForm.js
```diff
@@ -0,0 +1,178 @@
...
+
+  const editPassword = (value) => {
+    if (value === null) {
+      setPassword('');
+    } else {
+      setPassword(value);
+    }
+  };
+
+  const hasPassword = (cdnConfiguration.password_exists && password === null)
```
jeremylenz, line 52: For this condition,
```js
cdnConfiguration.password_exists && password === null
```
my form is still showing invalid even though it's valid.  (I had an existing cdn configuration saved already.)
When I enter a password in the box and click :ballot_box_with_check:, the alert disappears.  I think the validation needs to somehow check `hasPassword` and not just `password`..

jeremylenz (changes requested):
Looking good except one thing!

webpack/scenes/Subscriptions/Manifest/CdnConfigurationForm.js
jturel: nice catch - will fix

jeremylenz (approved):
Gorgeous!

ACK :+1:

PR #9817, opened 2021-11-29
Fixes #33576 - Add upgrade rake task to change repo url auth

rverdile:
This adds a rake task that checks all the existing repositories for URLs that include username/password like `https://user:pass@example.com`, then replaces said URL with `https://example.com` and assigns the username and password fields to `upstream_username` and `upstream_password` in the repository.

It also adds an error check to repository create and update, so that the user cannot include username and password in the URL.
The task performs the Update action which should make sure that everything that needs to get updated is, including in pulp.
1. Try creating and updating a repository and put the username/password in the URL. Notice you'll get an error.
2. Remove the error check and create a repository with the username/password in the URL.
3. Run the rake task and verify that the repository has been updated correctly in katello and pulp.

### DIFF
--- a/app/controllers/katello/api/v2/repositories_controller.rb
+++ b/app/controllers/katello/api/v2/repositories_controller.rb
@@ -234,6 +234,10 @@ module Katello
         fail HttpErrors::UnprocessableEntity, msg
       end
 
+      if !repo_params[:url].nil? && URI(repo_params[:url]).userinfo
+        fail "Do not include the username/password in the URL. Use the username/password settings instead."
+      end
+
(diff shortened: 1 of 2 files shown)


rverdile:
[test katello]

jjeffers (approved):
ACK!

Tested this by creating some repositories with old style pulp2 urls with auth.

Next, applied this PR, then tried to create new repos with the same urls via hammer and the UI. Both were rejected as expected.

Then I ran the rake upgrade task - inspected the existing repos and checked that the root url was fixes and the upstream username and passwords were set.

Great job!

PR #9822, opened 2021-12-01
Fixes #34059 - disable validation when seeding existing org

jturel:

Disables validation when seeding organizations

Disabling validation should be safe because any existing validation errors are not resolvable at this point in execution (db seed)

I haven't been through these steps, but they should get you started. The goal is to create an org that is not "created in katello" and will hit this code path when seeding, and also has host mismatches which is defined as an org having a host in a location that the org itself does not belong to.

- create a host in some organization
- assign that host to a location
- ensure the host's org is not in that location
- the above steps may need to be done via rails console to disable validations
- use the rails console to set created_in_katello on the org to false
- running db:seed should fail with an error "ActiveRecord::RecordInvalid: Validation failed: Environments expecting environments used by hosts or inherited (check mismatches report)."
- Checking out this PR and seeding again should succeed (check that the org now has created_in_katello=true)

### DIFF
--- a/app/services/katello/organization_creator.rb
+++ b/app/services/katello/organization_creator.rb
@@ -16,7 +16,7 @@ module Katello
       User.as_anonymous_admin do
         Organization.not_created_in_katello.each do |org|
           creator = self.new(org)
-          creator.create!
+          creator.create!(raise_validation_errors: false)
         end
       end
     end
@@ -42,7 +42,7 @@ module Katello
       end
     end
 
-    def create!
+    def create!(raise_validation_errors: true)
       ActiveRecord::Base.transaction do
         seed!
 
@@ -50,8 +50,11 @@ module Katello
 
         @organization.created_in_katello = true
 
-        # existing validation errors are not resolvable here, so don't validatate
-        @organization.save(validate: false)
+        begin
+          @organization.save!
+        rescue => e
+          raise e if raise_validation_errors
+        end
       end
     end
 
--- a/app/services/katello/organization_creator.rb
+++ b/app/services/katello/organization_creator.rb
@@ -28,7 +28,9 @@ module Katello
     def seed!
       ActiveRecord::Base.transaction do
         @organization.setup_label_from_name
-        @organization.save!
+
+        # existing validation errors are not resolvable here, so don't validatate
+        @organization.save(validate: false)
 
         create_library_environment
         create_library_view
@@ -47,7 +49,9 @@ module Katello
         create_backend_objects!
 
         @organization.created_in_katello = true
-        @organization.save!
+
+        # existing validation errors are not resolvable here, so don't validatate
+        @organization.save(validate: false)
       end
     end
 
--- a/test/controllers/api/v2/organizations_controller_test.rb
+++ b/test/controllers/api/v2/organizations_controller_test.rb
@@ -71,8 +71,11 @@ module Katello
     end
 
     def test_create_duplicate_name
+      stub_organization_creator
+
       post(:create, params: { :organization => {"name" => @organization.name} })
       assert_response :unprocessable_entity
+      assert_match(/Name has already been taken/, response.body)
     end
 
     def test_controller_path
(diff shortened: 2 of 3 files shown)


chris1984 (approved):
Works for me, ACK

Testing steps below:

Did the steps in the Rails console other than getting a host in my box:

Made an organization called test and a location called foo, with no organizations in it

Moved the host to that org and location

set created_in_katello to false

ran db seed and got
```ruby
[vagrant@devel foreman]$ bundle exec rake db:seed RAILS_ENV=development
Rake aborted!
ActiveRecord::RecordInvalid: Validation failed: Environments expecting environments used by hosts or inherited (check mismatches report).
```
After applying your patch I get:

`2021-12-14T14:53:41 [D|app|] Acquiring Advisory-Transaction-Lock 'database_seed' -> 2715469950095921260`

chris1984:
[test katello]

jturel:
@chris1984 Dang, I didn't realize I had a test failure here which actually caught a bug. The bug was that creating an organization was no longer validating - creating thru the UI or API for example. That still needs to happen, but we don't want validation when the middleware at lib/katello/middleware/organization_creator.rb runs. Pushed the fix as a separate commit - please let me know if you have any questions!

The testing is the same, with one additional case:

- do the data setup as before
- restart the rails server with the new code
- hit the ui/api and make sure created_in_katello gets set to true
- also make sure creating an org via API still validates (ie can't create an org with name or label of existing org)

chris1984 (approved):
Works for me, ACK

Testing steps below:

Did the steps in the Rails console other than getting a host in my box:

Made an organization called Test and a location called Pizza, with no organizations in it

Moved the host to that org and location

set created_in_katello to false

ran db seed and got

[vagrant@devel foreman]$ bundle exec rake db:seed RAILS_ENV=development
Rake aborted!
ActiveRecord::RecordInvalid: Validation failed: Environments expecting environments used by hosts or inherited (check mismatches report).
After applying your patch I get:

2022-01-06T11:53:41 [D|app|] Acquiring Advisory-Transaction-Lock 'database_seed' -> 2715469950095921489

Here is the output of my db before and after for the organization:
```json
{
  "RECORDS": [
    {
      "id": 4,
      "name": "Pizza",
      "type": "Location",
      "created_at": "6/1/2022 11:01:41.140275",
      "updated_at": "6/1/2022 11:01:41.140275",
      "ignore_types": "---\n- ProvisioningTemplate\n- Hostgroup\n",
      "description": "",
      "label": null,
      "ancestry": null,
      "title": "Pizza",
      "manifest_refreshed_at": null,
      "created_in_katello": "f"
    },
    {
      "id": 3,
      "name": "Test",
      "type": "Organization",
      "created_at": "6/1/2022 11:59:01.363731",
      "updated_at": "6/1/2022 11:59:08.193071",
      "ignore_types": null,
      "description": "",
      "label": "Test",
      "ancestry": null,
      "title": "Test",
      "manifest_refreshed_at": null,
      "created_in_katello": "f"
    }
  ]
}
```

After hitting an api endpoint, we can see the `created_in_katello` go from `f` to `t`

```json
{
  "RECORDS": [
    {
      "id": 4,
      "name": "Pizza",
      "type": "Location",
      "created_at": "6/1/2022 16:01:41.140275",
      "updated_at": "6/1/2022 16:01:41.140275",
      "ignore_types": "---\n- ProvisioningTemplate\n- Hostgroup\n",
      "description": "",
      "label": null,
      "ancestry": null,
      "title": "Pizza",
      "manifest_refreshed_at": null,
      "created_in_katello": "f"
    },
    {
      "id": 3,
      "name": "Test",
      "type": "Organization",
      "created_at": "6/1/2022 15:59:01.363731",
      "updated_at": "6/1/2022 15:59:08.193071",
      "ignore_types": null,
      "description": "",
      "label": "Test",
      "ancestry": null,
      "title": "Test",
      "manifest_refreshed_at": null,
      "created_in_katello": "t"
    }
  ]
}
```

I tested validation as well to make sure we get an error with creating an org with the same name and we do in the screenshot below:

https://imgur.com/a/OKtfIn6

PR #9823, opened 2021-12-01
Fixes #34060 -  Content View Version Task polling bug

Andrewgdewar:
- Fix the task polling bug

- Promote a version on the ContentViewVersion page, keep an eye on the network tab after completion. 
- All polling for tasks should stop.

### DIFF
--- a/app/lib/katello/resources/registry.rb
+++ b/app/lib/katello/resources/registry.rb
@@ -32,7 +32,7 @@ module Katello
             self.prefix = "/pulpcore_registry/"
             self.site = "#{uri.scheme}://#{uri.host}:#{uri.port}"
             self.ca_cert_file = Setting[:ssl_ca_file]
-            pulp_primary.pulp3_ssl_configuration(self)
+            pulp_primary.pulp3_ssl_configuration(self, :net_http)
 
             self
           end
--- a/app/models/katello/concerns/smart_proxy_extensions.rb
+++ b/app/models/katello/concerns/smart_proxy_extensions.rb
@@ -185,17 +185,17 @@ module Katello
         end
       end
 
-      def pulp3_ssl_configuration(config)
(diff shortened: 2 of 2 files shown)


webpack/scenes/ContentViews/Details/Versions/ContentViewVersions.js
```diff
@@ -94,8 +94,20 @@ const ContentViewVersions = ({ cvId, details }) => {
...
+
     if (result !== 'error' && !pollIntervals[pollTaskKey(task.id)]) {
-      dispatch(startPollingTask(task.id, task));
+      dispatch(startPollingTask(
+        task.id, task,
+        ({
+          data,
+        }) => {
+          if (!data) return;
+          const { available_actions: { cancellable, resumable } } = data;
```
sjha4, line 105: I'm not too sure if checking cancellable and resumable is an accurate reflection of all tasks in general. I think checking for result === error or result === success will make this easier to follow.

jeremylenz: I was going to say the same thing :smile:

jeremylenz (commented):
makes sense, thanks @Andrewgdewar!

As an extra precaution it might be good to also add a cleanup in a useEffect return, so that we cancel any active polling when the component is unmounted. I noticed the errant polling continues even after I switch to a different React page.

webpack/scenes/ContentViews/Details/Versions/ContentViewVersions.js
Andrewgdewar: > makes sense, thanks @Andrewgdewar!
> 
> As an extra precaution it might be good to also add a cleanup in a useEffect return, so that we cancel any active polling when the component is unmounted. I noticed the errant polling continues even after I switch to a different React page.

Will Do

PR #9824, opened 2021-12-01
Make jest timeout configurable

jturel:

Reverts the 300sec jest timeout back to 10 seconds in dev, but preserves the 300sec timeout for our github actions

Make development less painful when writing react tests with nock requests

- break a react test by changing something like the expected request body
- with this pr checked out, notice the test fails in ~10 seconds
- run the test again, prepending your command with JEST_TIMEOUT=300000 and notice it now takes 300 secs to time out

### DIFF
--- a/webpack/global_test_setup.js
+++ b/webpack/global_test_setup.js
@@ -8,7 +8,7 @@ global.console.error = (error, stack) => {
 };
 
 // Increase jest timeout as some tests using multiple http mocks can time out on CI systems.
-jest.setTimeout(300000);
+jest.setTimeout(process.env.JEST_TIMEOUT || 15000);
 
 afterAll(() => {
   jest.resetModules();


jturel:
Does this address my concern, or do I need to update some jenkins job somewhere too?

jturel:
Oh, I don't think this helps the nightly pipeline. We'd need to set the JEST_TIMEOUT there before merging this. Let me have a look into that.

.github/workflows/react_tests.yml
```diff
@@ -8,8 +8,9 @@ defaults:
 
 jobs:
   build:
-
     runs-on: ubuntu-latest
+    env:
+      JEST_TIMEOUT: 300000
```
jturel, line 13: Do I need to add this? ie were the timeouts only in the nightly pipeline, or did we also see them in the github workflow?

jturel:
See https://github.com/theforeman/jenkins-jobs/pull/156

Andrewgdewar:
Screen Shot 2021-12-02 at 10 19 20 AM

Testing locally I saw this.

.github/workflows/react_tests.yml
Andrewgdewar: @sjha4  I don't remember... :(

webpack/global_test_setup.js
```diff
@@ -8,7 +8,7 @@ global.console.error = (error, stack) => {
 };
 
 // Increase jest timeout as some tests using multiple http mocks can time out on CI systems.
-jest.setTimeout(300000);
+jest.setTimeout(process.env.JEST_TIMEOUT || 10000);
```
Andrewgdewar, line 11:
```suggestion
jest.setTimeout(process.env.JEST_TIMEOUT || 20000);
```

.github/workflows/react_tests.yml
sjha4: We never saw the timeouts on github..This was a nightly only issue.

webpack/global_test_setup.js
sjha4: 2nd test run passed on github with 10k..20k should be good.. 👍🏽

jturel:
@sjha4 @Andrewgdewar updated the default timeout to 15 seconds rather than 20. We may hit an occasional failure on the github action or locally but I think that's a good thing - to identify slow tests sooner than on Jenkins and potentially fix them via some amount of refactoring.

jturel:
The jenkins-job PR using the new env var is merged so this is safe to merge once final ACKs are received :)

sjha4 (approved):
15 secs is long enough for individual unit tests. Longer tests if any probably need refactoring and this can raise awareness whenever that happens. ACK 👍🏽

PR #9826, opened 2021-12-02
don't configure minitest-reporters, Foreman does already

evgeni:

Dropping the use of `Minitest::Reporters` in Katello tests, as these are now centrally managed in Foreman.

In https://github.com/theforeman/foreman/pull/8960 we're switching
Foreman to use minitest-reporters and also enable the MeantimeReporter,
so the special setup in Katello isn't needed anymore.

Especially, calling Minitest::Reporters.use! twice breaks things.

Ensure test output still contains MeantimeReporter output.

### DIFF
--- a/lib/katello/tasks/jenkins.rake
+++ b/lib/katello/tasks/jenkins.rake
@@ -2,8 +2,6 @@ require File.expand_path("../engine", File.dirname(__FILE__))
 
 begin
   namespace :jenkins do
-    ENV['USE_MEAN_TIME_REPORTER'] = '1' unless ENV['USE_MEAN_TIME_REPORTER'] == '0'
-
     task :katello do
       Rake::Task['jenkins:setup:minitest'].invoke
       Rake::Task['rake:test:katello'].invoke
--- a/katello.gemspec
+++ b/katello.gemspec
@@ -71,7 +71,6 @@ Gem::Specification.new do |gem|
   # Testing
   gem.add_development_dependency "factory_bot_rails", "~> 4.5"
   gem.add_development_dependency "minitest-tags"
-  gem.add_development_dependency "minitest-reporters"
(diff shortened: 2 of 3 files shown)


evgeni:
```
ModuleNotFoundError: Module not found: Error: Can't resolve 'foremanReact/redux/actions/toasts' in '/home/jenkins/workspace/katello-pr-test/webpack/utils'
```

Failure unrelated

tbrisker (approved):

lib/katello/tasks/jenkins.rake
```diff
@@ -2,8 +2,6 @@ require File.expand_path("../engine", File.dirname(__FILE__))
 
 begin
   namespace :jenkins do
-    ENV['USE_MEAN_TIME_REPORTER'] = '1' unless ENV['USE_MEAN_TIME_REPORTER'] == '0'
```
evgeni, line 5: @jturel do you remember why this env var was introduced, instead of running the reporter all the time?

In Foreman we now implemented it without this var, but I wanted to check back if there was any reason to want this disabled?

jturel: Oh, I'm not actually sure. Though I don't think it was ever set to 0 until later on when we found the mean time reporter somehow causing timeouts in jenkins - or at least that was the theory at the time.

jturel (approved):
ACK! It looks like a clean revert of 8d0c2fe1dc1004f826512635c2ed165be3d34b46

jturel:
[test katello]

lib/katello/tasks/jenkins.rake
evgeni: Yeah, I don't think we ever made Jenkins or anything else set it to 0.

evgeni:
🥦

PR #9828, opened 2021-12-02
Fixes #34071 - Stop using deprecated redux toast actions

ofedoren:
Removing of deprecated toast actions in favor of a new way of using toats.
Updated tests/snapshots, but please, make sure I didn't break anything.

### DIFF
--- a/webpack/__mocks__/foremanReact/components/ToastsList/index.js
+++ b/webpack/__mocks__/foremanReact/components/ToastsList/index.js
@@ -1,7 +1,8 @@
 export const addToast = toast => ({
-  type: 'TOASTS_ADD',
+  type: 'toasts/addToast',
   payload: {
-    message: toast,
+    key: 'addToast',
+    toast,
   },
 });
 
--- a/webpack/utils/helpers.js
+++ b/webpack/utils/helpers.js
@@ -1,4 +1,4 @@
-import { addToast } from 'foremanReact/redux/actions/toasts';
+import { addToast } from 'foremanReact/components/ToastsList';
 import { translate as __ } from 'foremanReact/common/I18n';
 import { head } from 'lodash';
 import { SUBSCRIPTIONS_QUANTITIES_FAILURE } from '../scenes/Subscriptions/SubscriptionConstants';
@@ -81,6 +81,7 @@ export const sendErrorNotifications = messages => (dispatch) => {
     const message = typeof msg === 'string' ? msg : `${msg.message}: ${msg.details}`;
     dispatch(addToast({
       type: 'error',
+      key: 'toastError',
       message,
       sticky: true,
     }));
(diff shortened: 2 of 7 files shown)


webpack/__mocks__/foremanReact/redux/actions/toasts.js
```diff
@@ -1,5 +1,5 @@
 export const addToast = toast => ({
```
tbrisker, line 1: This file should be renamed to `webpack/__mocks__/foremanReact/components/ToastsList/index`

ofedoren: :facepalm: sure, done.

tbrisker:
@ofedoren looks like the snapshots were generated with the real action (that adds a key and has a bit different structure to the mock), but in CI the mock is being used. Yet another example showing why snapshot tests are mostly useless...

webpack/utils/helpers.js
```diff
@@ -81,6 +81,7 @@ export const sendErrorNotifications = messages => (dispatch) => {
     const message = typeof msg === 'string' ? msg : `${msg.message}: ${msg.details}`;
     dispatch(addToast({
       type: 'error',
+      key: 'toastError',
```
tbrisker, line 84: Could there be more than on error thrown at the same time? not sure we should use a fixed key here.

ofedoren: Well, I'm not sure either. But I'd find a bad UX in multiple toasts thrown at me with error. On the other hand, it's possible that different components would call different APIs and each of them fails with a toast... :/

tbrisker: Indeed, they might even have different error messages

Andrewgdewar: this seems like a good spot to use the forEach((msg, index) => 
... 
key: `toastError${index}`,
... 
and disable the eslint rule around such things for this line. no?

webpack/__mocks__/foremanReact/components/ToastsList/index.js
```diff
@@ -1,7 +1,8 @@
 export const addToast = toast => ({
```
Andrewgdewar, line 1: Why are you moving this file into mocks? is this intentional

Andrewgdewar:
So after testing this PR against the latest foreman develop, I am unable to start katello, I am getting the following errors. 
Screen Shot 2021-12-02 at 12 09 00 PM

Andrewgdewar:
Upon removing both references to the issues seen above, I can now load into katello, but I am seeing the following foreman-task gem failure
Screen Shot 2021-12-02 at 12 37 21 PM
es

adding the following line to update to the latest foreman-tasks in foreman was necessary to get the application to work.

Andrewgdewar:
Closing this, have opened another PR [here](https://github.com/Katello/katello/pull/9829) to address these changes.
Note: A PR/release updating foreman-tasks is needed to prevent katello failures.

PR #9834, opened 2021-12-03
Fixes #33939 - support mirroring policy on repositories

jlsherrill:
Prior to pulp 3.14 (including our time with pulp2), mirror on sync meant that if content was removed from an upstream repository, it would be removed from the local repository at sync time.  If this was disabled, stuff would just be continually added to the repository and never removed.  This is implemented by us calling sync in pulp3 with mirror=true

As part of 3.14 pulp added a 'metadata mirroring' support and reused the same option to now trigger metadata mirroring in addition to content mirroring.  It turns out that some repositories cannot actually have their metadata mirrored due to various reasons, and so pulp started throwing an error suggesting that the user not use the mirror option. 

Now pulp supports a new option 'sync_policy' that lets the user decide between:
* additive content, metadata generation - 'additive'
* mirrored content, metadata generation - 'mirror_content_only'
* mirrored content, mirrored metadata - 'mirror_complete'

This exposes this option as a 'mirroring policy'.

On upgrade, we migrate all redhat repos set to 'mirror_on_sync' to mirror_complete, since all of them can have their metadata mirrored.
For all custom repos we migrate anything with mirror_on_sync set to true to mirror_content_only
If mirror_on_sync is not true, we migrate it to additive.

The mirror_on_sync option is also handled via the repo create/update api for backwards compatibility.  

1.  create some repos, some with mirror on sync set to true, some to false, mix of redhat and custom
2. checkout pr and run db:migrate
3. verify that migrated repos match what i mention above
4. test updating the 'mirror policy' of yum repos and other repos 
5. test creating of yum repos and other repos.


jlsherrill:
[test katello]

db/migrate/20211202172358_add_repository_mirroring_policy.rb
```diff
@@ -0,0 +1,23 @@
+class AddRepositoryMirroringPolicy < ActiveRecord::Migration[6.0]
+  def up
+    add_column :katello_root_repositories, :mirroring_policy, :string
+
+    Katello::RootRepository.where(:content_type => 'yum').where(:mirror_on_sync => true).update_all(:mirroring_policy => ::Katello::RootRepository::MIRRORING_POLICY_COMPLETE)
+    Katello::RootRepository.where.not(:content_type => 'yum').where(:mirror_on_sync => true).update_all(:mirroring_policy => ::Katello::RootRepository::MIRRORING_POLICY_COMPLETE)
```
rverdile, line 6: Would `Katello::RootRepository.where(:mirror_on_sync => true).update_all(:mirroring_policy => ::Katello::RootRepository::MIRRORING_POLICY_COMPLETE)` get the same result as these two lines?

rverdile (commented):
After migrating, if I try to update any repository I get: `An error occurred saving the Repository: can't write unknown attribute mirror_on_sync`. I think this is caused by removing the `mirror_on_sync` column in the migration?

On a similar note, should that column be removed if the issue says "old option should still 'work' via controller code"?

db/migrate/20211202172358_add_repository_mirroring_policy.rb
jlsherrill: ah yes, i meant to make custom repos content_only!!

jlsherrill:
@rverdile hrmm, i will retry that, did you refresh your UI after checking out the PR?

rverdile:
@jlsherrill yeah I refreshed the page. The more precise error message is `can't write unknown attribute mirror_on_sync (ActiveModel::MissingAttributeError)`. 

I might be missing something, but I think it's because `def update` in `repositories_controller.rb` is calling `repository_params`, which includes `mirror_on_sync` as a key. So there would be the value "mirror_on_sync: nil" when update is called, but the mirror_on_sync column no longer exists.

rverdile (approved):
Code and functionality looks good to me!. Migration works both ways and I'm able to create/update repositories' mirroring policies.

PR #9835, opened 2021-12-06
Fixes #34087 - Revert "Fixes #31173 - no deb repo in sub-man"

bastian-src:

This reverts commit ed89bc5cfb4e9ffec082be68a1c40d8efd25bbca.

Instead of transmitting the corresponding debian architecture to the host, a default value has been sent. From now on, the _actual_ architecture must be sent to enable debian architecture support.

The mentioned commit reverted changes due to a delayed implementation of features in `subscription-manager`. Since the changes have been [merged](https://github.com/candlepin/subscription-manager/pull/2213) by now, this commit can be "un-reverted" to enable the debian architecture functionality.

### DIFF
--- a/app/models/katello/root_repository.rb
+++ b/app/models/katello/root_repository.rb
@@ -396,11 +396,7 @@ module Katello
 
     def format_arches
       if content_type == ::Katello::Repository::DEB_TYPE
-        # FIXME: This should be set to self.deb_architectures but it needs to have the
-        # subscription-manager PR https://github.com/candlepin/subscription-manager/pull/2213
-        # merged. Otherwise subscription-manager returns _NO_ debian repository as described in
-        # https://community.theforeman.org/t/katello-3-16-1-1-el7-subscription-manager-doesnt-create-rhsm-repos-for-ubuntu/20928/38
-        nil
+        self.deb_architectures
       else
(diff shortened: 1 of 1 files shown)


jturel:
@bastian-src hello! do we need an upgrade script to update the arches of existing deb content?

See lib/katello/tasks/upgrades if you're unfamiliar. The question came to mind while looking at where format_arches is used:  https://github.com/Katello/katello/blob/c45b0fc21b531e9911251b47d5d1de433d5e0fda/app/lib/actions/katello/repository/update.rb#L27

bastian-src:
Hey @jturel! Thanks for you feedback.
There is no update needed, since the `deb_architectures` has been introduced before already. This commit just enables the functionality again (by reverting the revert of the feature) since we've been waiting on getting some `subscription-manager` PR merged as described in the beginning.

jturel:
[test katello]

jturel:
Works well. I tested two workflows:

- Org created via hammer (seed and create! happen together) - I got all the right artifacts & no double audit
- Simple org created in rails console. Ran db:seed. verified created_in_katello is still false. Hit the API which triggers the rest of the create. Candlepin owner,etc got created

jturel (approved):
Thanks for following up on this @bastian-src !

PR #9836, opened 2021-12-06
Fixes #34060 -  Content View Version Task polling bug

Andrewgdewar:
## What are the changes introduced in this pull request?
Fix the task polling bug
## What are the testing steps for this pull request?
Promote a version on the ContentViewVersion page, keep an eye on the network tab after completion.
All polling for tasks should stop.

### DIFF
--- a/webpack/scenes/ContentViews/Details/Versions/ContentViewVersions.js
+++ b/webpack/scenes/ContentViews/Details/Versions/ContentViewVersions.js
@@ -9,6 +9,7 @@ import { Link } from 'react-router-dom';
 import PropTypes from 'prop-types';
 import { selectIntervals } from 'foremanReact/redux/middlewares/IntervalMiddleware/IntervalSelectors.js';
 
+import { useSet } from '../../../../components/Table/TableHooks';
 import TableWrapper from '../../../../components/Table/TableWrapper';
 import InactiveText from '../../components/InactiveText';
 import ContentViewVersionEnvironments from './ContentViewVersionEnvironments';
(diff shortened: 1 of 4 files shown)


webpack/scenes/ContentViews/Details/Versions/ContentViewVersions.js
```diff
@@ -66,6 +68,14 @@ const ContentViewVersions = ({ cvId, details }) => {
     [dispatch],
   );
 
+  useEffect(() => () => {
```
jeremylenz, line 71: I expected to see the explicit `return` keyword here, but realized that eslint won't allow it because of https://eslint.org/docs/rules/arrow-body-style.  I think someone who knows Hooks should be able to grok what's going on if the `return` keyword is there, but the implicit return of arrow functions makes it confusing and obfuscates what's going on. For that reason I think we should disable that eslint rule here and use `return`.
```suggestion
  useEffect(() => { // eslint-disable-line arrow-body-style
    return () => {
```

jeremylenz (commented):
Works great!  One comment below

webpack/scenes/ContentViews/Details/Versions/ContentViewVersions.js
Andrewgdewar: I suppose it is good to make the return implicit for posterity. 
May our ancestors me be appeased! 🥲

sjha4 (approved):
Worked well for me..Test based ACK 👍🏽 

Feel free to merge if @jeremylenz is fine with the code..

jeremylenz (approved):
ACK :+1:

PR #9840, opened 2021-12-07
Fixes #34103 - lower get length for fetch objects by id

jlsherrill:

When fetching docker manifest list as part of the migration, we fetch them by pulp href, and since docker_manifest_list is a longer name than say 'rpm', we can't fetch as many.  This lowers that threshold. 

This may slow down the migration slightly.

This solution has been tested and shown to solve the issue: https://bugzilla.redhat.com/show_bug.cgi?id=2025948#c1
But for manual testing, you could sync https://quay.io/repository/openshift-release-dev/ocp-release  down, that should generate enough manifest lists, but you may need a lot of disk space

### DIFF
--- a/app/services/katello/pulp3/migration.rb
+++ b/app/services/katello/pulp3/migration.rb
@@ -6,7 +6,7 @@ module Katello
       attr_accessor :smart_proxy, :reimport_all, :task_id
       attr_reader :repository_types
 
-      GET_QUERY_ID_LENGTH = 90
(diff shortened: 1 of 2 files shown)


jturel (approved):
I'm acking this 'cause the code looks good and I trust that it works due to comments saying as much on the BZ. I haven't tested it it, but I'm comfortable with forgoing that in this case. Lemme know if you disagree!

jlsherrill:
I agree completely, i don't think we need to reproduce the issue

PR #9841, opened 2021-12-07
Fixes #34102 - Bump up distributor version to 6.10

lfu:
Bump up distributor version to 6.10
- create a manifest and select a version prior to '6.10'.
- import that manifest and refresh it.
- portal should reflect 6.10 as the version instead of whatever was set intially


jeremylenz (approved):
- [x] Refreshed my Satellite 6.7 manifest
- [x] Customer portal now shows "Satellite 6.10"

ACK :+1:

PR #9844, opened 2021-12-09
Default download policy for deb repos if not set

nadjaheitmann:
**What are the changes introduced in this pull request?**
At the moment, setting a default for download policy is only available for yum content. We want to add the same policy for deb content. We also add a setting that allows to set the default download policy for deb content type, only.

**What are the testing steps for this pull request?**
Create repository of type deb and check if default download policy is set.

**EDIT: Considerations taken when implementing this change?**
Creating a `deb` repo without specifying a `download_policy` results in an error that the `download_policy` field is not supposed to be empty.

### DIFF
--- a/db/migrate/20220127120843_fix_debian_download_policy.rb
+++ b/db/migrate/20220127120843_fix_debian_download_policy.rb
@@ -1,0 +1,11 @@
+class FixDebianDownloadPolicy < ActiveRecord::Migration[6.0]
+  def up
+    Katello::RootRepository.where(content_type: "deb")
+                           .where(download_policy: [nil, ""])
+                           .update_all(:download_policy => "immediate")
+  end
+
+  def down
+    fail ActiveRecord::IrreversibleMigration
+  end
+end


nadjaheitmann:
Not sure if test failure is related. The same test passes for me on a 4.3 stable branch.

sbernhard:
[test katello]

sbernhard:
LGTM!

jturel:
> @jlsherrill @jturel Can one of you have a look at this one, please?

Added this to our sprint for review. I think someone should get to it this week :)

app/models/setting/content.rb
```diff
@@ -92,6 +92,8 @@ def self.default_settings
                N_('Default Custom Repository download policy'), nil, :collection => download_policies),
       self.set('default_redhat_download_policy', N_("Default download policy for enabled Red Hat repositories (either 'immediate' or 'on_demand')"), "on_demand",
                        N_('Default Red Hat Repository download policy'), nil, :collection => download_policies),
+      self.set('default_deb_download_policy', N_("Default download policy for enabled repositories of type 'deb' (either 'immediate' or 'on_demand')"), "immediate",
+                       N_('Default Debian content download policy'), nil, :collection => download_policies),
```
jlsherrill, line 96: Is there a reason to not just use 'default_download_policy' for deb as well?  There's nothing about that is specific to yum

jlsherrill:
Should their be a migration to migrate existing repositories (i assume to immediate) ?

app/models/setting/content.rb
nadjaheitmann: We consider `ondemand` policy as problematic. Maybe @m-bucher can elaborate on this.

nadjaheitmann:
> Should their be a migration to migrate existing repositories (i assume to immediate) ?
I don't think that we should migrate existing repos if they already have a download_policy set.

jlsherrill:
For yum, a download_policy is required to be set on the object.  If the user doesn't set it, it defaults to the Setting.  It looks like here, its not being set as required, and thus it could be 'nil' for debian repos that are upgraded.    Nil is the same as immediate i presume, so i think it makes sense to:
1) require the download policy be set, similar to yum:  https://github.com/Katello/katello/blob/master/app/models/katello/root_repository.rb#L90-L93
2) on migration, set each deb repos download_policy to immediate.  Note, that this isn't actually changing anything, as nil is treated as immediate

nadjaheitmann:
Actually, the main reason why we want to set a default policy for deb repos is that creating a deb repo without setting a download_policy results in an error. I should have put this in the description of the ticket.

app/models/setting/content.rb
m-bucher: I currently suspect that it depends on the repository, whether using `on_demand` is a good idea.
If the repository does remove packages from the pool that are no longer in the current package-indizes, older versions that were synced in katello with `on_demand` will no longer work.
It is not a problem, if the upstream repository's package-pool only gets new packages and never looses older ones.

The first case might be especially problematic if the repo-versions are Content-View versions unless the packages are synced, when a Content-View version is created from the repository.
I might lack some deeper knowledge about how `on_demand` really works :sweat_smile:

jlsherrill: @m-bucher do you think it needs to be a different setting entirely?  currently 'default_download_policy' defaults to immediate for similar reasons you mentioned (if older packages are removed from upstream repos, it can break content views).  So we default custom repos to immediate as well.

m-bucher: In that case I think we can get away with keeping it in one setting and revisit it if users demand it.

It may depend on how high the percentages are of repositories removing old packages versus repos that do not.
My gut tells me deb-repos in the wild tend to be more `remove old packages`.
Not a clue about what kind is used more often with yum-repos.
But then again I do not have reliable data for deb-repos either, just a feeling :innocent:

jlsherrill: okay, how about we start off using the same setting, and we can always split it out later on.  I'd rather err on the side of fewer settings.  @nadjaheitmann do you mind updating the pr to just use  'default_download_policy' ?

nadjaheitmann: So it will be just the migration, then?

jlsherrill: migration plus the chnage to app/models/katello/glue/pulp/repos.rb to also set it for dep repos

nadjaheitmann: Really change in repos.rb? If I look in https://github.com/Katello/katello/blob/d2c20af6911adb02ca9005350a9f2c1ad1f8140b/app/models/katello/root_repository.rb#L23, it looks like it is already checking for `deb` repos in https://github.com/Katello/katello/blob/d4f2976f4d28e574d291dcbb793afa02b7b8a363/app/models/katello/glue/pulp/repos.rb#L135.

jlsherrill: oh right, good call!  just the migration then!

nadjaheitmann: At least I don't have to close the whole PR because it became obsolete ;)

nadjaheitmann: Updated the PR and also the redmine issue.

nadjaheitmann:
@jlsherrill Do you mind having a final look?

jlsherrill (approved):
Thanks @nadjaheitmann !

PR #9846, opened 2021-12-10
Fixes #34127 - Purge empty content before generating metadata

adamruzicka:
kudos to @pmoravec for figuring this one out

Apparently due to order in which we perform steps when publishing a CV can lead to CV filter being ignored and then unwated packages ending up in the repository.

@pmoravec's analysis looks solid, publishing this as a jump-off point to start discussion. To be honest I can't really tell if the steps were intentionally ordered the way they were.

Sadly at this time we don't have a reliable reproducer.


jeremylenz:
[test katello]

jeremylenz:
ah, guess that pipeline is no more?

ianballou:
Closing this since the issue is more of a depsolving problem.

PR #9847, opened 2021-12-10
Fixes #34128 - Smart proxy syncing of a specific LCE fails on Katello 4.2

ianballou:

When https://github.com/Katello/katello/commit/10b33c1a362651e7a7a9b098593bb97acf92c445 was cherry-picked in, the breaking change of having active record objects passed to `sync_capsule.rb`'s `plan_self` was brought in.  We simply need to stop passing the unused options into `plan_self`.

The options are unused, so they can be deleted with no side effects.

Trigger a smart proxy auto-sync by adding an LCE to a proxy and publishing a CVV with that LCE.  Check for the proxy sync task in the tasks viewer. It should not error out.


ianballou:
[test katello]

sjha4 (approved):
LGTM..Ack 👍🏽

PR #9850, opened 2021-12-13
Fixes #34143 - Show synced audit result in sync status page

sjha4:
Use audit log to display sync date/time when associated task has been deleted instead of "Never synced"

Never synced message can be confusing for users when the repo has been synced but the task has been deleted. We use audit of sync task in these cases on repo and product page already.

Create and sync a repo.
1. Note task ID and then in foreman console, run : 
`ForemanTasks::Task.find('$TASK_ID').destroy!`
2. Go to sync status page.

**Without this change:** 
Result will be "Never Synced"

**With this change:**
Screenshot from 2021-12-13 14-45-33

### DIFF
--- a/app/views/katello/sync_management/_repo.html.erb
+++ b/app/views/katello/sync_management/_repo.html.erb
@@ -6,37 +6,48 @@
     </label>
   </td>
   <td class="start_time">
-    <%= @repo_status[repo.id][:start_time] %>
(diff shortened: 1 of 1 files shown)


rverdile (approved):
lgtm!

jlsherrill:
i was curious about the performance hit, and testing some queries on dogfood, it looks like with 148 repos (which is quite a bit) it would add about half a second to the request.  

I think this is okay for now!  +1 from me

PR #9852, opened 2021-12-13
Fixes #32379 -  Autopublish switch reloads entire page

Andrewgdewar:
Prevents full page reload when updating information on the content view info page, as well as on the repository review after adding or removing a repo.

- Go to the contentView details page and change name, description and autopublish. 
- Only the details sub-page should update.

- Go to repositories and add/remove repos. (bulk or not)
- Only the repo sub-page should update.

### DIFF
--- a/webpack/scenes/ContentViews/Details/ContentViewDetailActions.js
+++ b/webpack/scenes/ContentViews/Details/ContentViewDetailActions.js
@@ -1,5 +1,4 @@
 import { API_OPERATIONS, APIActions, get, put, post } from 'foremanReact/redux/API';
-import { addToast } from 'foremanReact/components/ToastsList';
 import { translate as __ } from 'foremanReact/common/I18n';
 import {
   RPM_PACKAGES_CONTENT,
@@ -49,7 +48,7 @@ import {
   DOCKER_TAGS_CONTENT,
 } from '../ContentViewsConstants';
 import api, { foremanApi, orgId } from '../../../services/api';
-import { getResponseErrorMsgs, apiError } from '../../../utils/helpers';
+import { getResponseErrorMsgs } from '../../../utils/helpers';
 import { renderTaskStartedToast } from '../../Tasks/helpers';
 import { cvErrorToast } from '../ContentViewsActions';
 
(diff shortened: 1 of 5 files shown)


webpack/scenes/ContentViews/Details/Repositories/__tests__/contentViewAddRemove.test.js
```diff
@@ -92,16 +92,19 @@ test('Can add repositories', async (done) => {
...
   getByLabelText('Select all rows').click();
   getByLabelText('add_repositories').click();
+  await patientlyWaitFor(() => expect(getByText(firstRepo.name)).toBeInTheDocument());
+
   assertNockRequest(repoAddscope);
 
   assertNockRequest(cvDetailScope);
   assertNockRequest(autocompleteScope);
   assertNockRequest(scope, done);
+  act(done);
```
sjha4, line 102: I still don't understand this.. 😕

Andrewgdewar: We add this to tell the test to ignore further actions on the render, so in this case there is still another api call after the "scope, done". 

But we don't care about that, so we specifically tell the test that we are done all actions: act(done). 
In this case, without the act(done) call, the test would wait forever for a corresponding api call to the follow up one that we don't care about.

Andrewgdewar: This is one of those cases where a change we have made, has a few chained api calls, sometimes we want to wait and confirm them, I didn't in this case, so I just told the test to STAHP.

webpack/scenes/ContentViews/Details/ContentViewDetailActions.js
```diff
@@ -86,19 +75,21 @@ export const getFiles = params => get({
...
   type: API_OPERATIONS.PUT,
   key: cvDetailsKey(cvId),
   url: api.getApiUrl(`/content_views/${cvId}`),
-  params,
-  handleSuccess: response => cvUpdateSuccess(response, dispatch),
-  handleError: error => dispatch(apiError(null, error)),
+  handleSuccess,
+  params: { include_permissions: true, ...params },
+  successToast: () => __(' Content view updated'),
+  errorToast: error => error,
```
sjha4, line 85: This will yield a better error message: `${getResponseErrorMsgs(error.response)}`)`

webpack/scenes/ContentViews/Details/ContentViewDetailActions.js
```diff
@@ -86,19 +75,21 @@ export const getFiles = params => get({
...
+  updateData: (prevState, respState) => ({ ...prevState, ...respState }),
   actionTypes: {
```
sjha4, line 87: This works too:
`updateData: (prevState, respState) => respState,`

webpack/scenes/ContentViews/Details/ContentViewDetailActions.js
Andrewgdewar: Roger

webpack/scenes/ContentViews/Details/ContentViewDetailActions.js
Andrewgdewar: Sweet

PR #9853, opened 2021-12-14
Fixes #34144 - Connected ISS from a CVE

parthaa:
This PR enables one to sync from an upstream satellite's content view
environment

In a Connected Inter Satellite Sync a downstream satellite can directly pull content from an upstream sat. This workflow is not enhanced in this PR to facilitate fetching content from a content view environment of the upstream sat. 

- Adds 2 entries to the CdnConfiguration
1. Content View label
2. Lifecycle Environment label 
- Uses this information to determine the CVE to pull content from the correct repository.
- Enables one to sync from a CVE
- 

- Can it handle custom content.? -  As of now Custom content can be easily imported in the master using the debug and ca certificate of the upstream sat. So this was not considered as a part of this PR.
- Should we  let you enable a RH Repo even if it is not in the CVE of the upstream satellite. This PR allows you to enable a RH Repo as long as the RH repo is enabled in the Library environment upstream. One will not be able to pull the content from a content view environment if the repo is not part of the content view environment. 

Very similar to the testing steps of https://github.com/Katello/katello/pull/8821 and https://github.com/Katello/katello/pull/9772
- Run the migration
- Go to `Subscriptions` -> `Manage Manifest` ->`Cdn Configurator` 
- Set the content view environment of the upstream
- Enable and sync :)

### DIFF
--- a/app/lib/katello/resources/cdn/katello_cdn.rb
+++ b/app/lib/katello/resources/cdn/katello_cdn.rb
@@ -4,6 +4,8 @@ module Katello
       class KatelloCdn < CdnResource
         def initialize(url, options)
           @organization_label = options.delete(:organization_label)
+          @content_view_label = options.delete(:content_view_label)
+          @lifecycle_environment_label = options.delete(:lifecycle_environment_label)
           fail ArgumentError, "No upstream organization was specified" if @organization_label.nil?
 
           super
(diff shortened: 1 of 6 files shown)


app/lib/katello/resources/cdn/katello_cdn.rb
```diff
@@ -2,8 +2,11 @@ module Katello
   module Resources
     module CDN
       class KatelloCdn < CdnResource
+        attr_reader :organization_label, :content_view_label, :lifecycle_environment_label
```
jturel, line 5: Sorry if I missed it but if these don't need to be accessed from outside the instance of this class, I prefer using the instance variables.

webpack/scenes/Subscriptions/Manifest/CdnConfigurationForm.js
```diff
@@ -130,7 +136,27 @@ const CdnConfigurationForm = (props) => {
...
+          />
+        </FormGroup>
+        <FormGroup
+          label={__('Lifecycle Environment Label')}
+        >
+          <TextInput
+            aria-label="cdn-lifecycle-environment-label"
+            type="text"
+            value={lifecycleEnvironmentLabel || ''}
+            onChange={setLifecycleEnvironmentLabel}
```
jlsherrill, line 159: I'd probably put Lifecycle Environment first, and content view 2nd.

Its also not obvious what i'm supposed to do if i just want to use 'Library'.  Any thought on how to make that clearer?

jlsherrill: and i get that i just leave those two fields blank, as they aren't required, i don't think a user will figure that out immediately thought.

parthaa: RkLMy2O

Added tool tips to make it clear. Will address it better in PR https://github.com/Katello/katello/pull/9812

parthaa:
[test katello]

app/services/katello/organization_creator.rb
```diff
@@ -86,8 +92,8 @@ def create_content_view_environment
 
     def create_library_environment
       @library_environment = Katello::KTEnvironment.where(
-        :name => "Library",
-        :label => "Library",
+        :name => DEFAULT_LIFECYCLE_ENV_NAME,
+        :label => DEFAULT_CONTENT_VIEW_LABEL,
```
jlsherrill, line 96: i think this should be DEFAULT_LIFECYCLE_ENV_NAME

which is probably the cause of the test failures

jlsherrill: errr  DEFAULT_LIFECYCLE_ENV_LABEL

jlsherrill:
everything looks and works well, just gotta fix those tests

parthaa:
> I'd probably put Lifecycle Environment first, and content view 2nd.

Reversed the order. LCE comes before CV

jlsherrill (approved):
I found some usability issues (one pretty major one), filed here: https://projects.theforeman.org/issues/34256

its unrelated to the changes in this pr, so ACK

PR #9854, opened 2021-12-14
Fixes #34145 - Errata href not updated

hao-yu:
Like any other content in Pulp 3, advisories are immutable. So when
there is some change to it, a new one is created, it is then
associated to a repository as a part of a new repository version and
the old one is unassociated and becomes orphaned.

There are some advisory's fields that Katello doesn't store in the
database but Pulp does, such as version, pushcount, fromstr, rights
etc. If any of those fields is changed after syncing repo, Katello
can't detect the change and it is still referring to the orphaned
hrefs.

When user publishing an incremental content view, Katello will use the
orphaned hrefs to copy advisories from source repo to the destination
repo. If the orphaned advisories are removed by Pulp then nothing will
be copied.

### DIFF
--- a/app/services/katello/pulp3/erratum.rb
+++ b/app/services/katello/pulp3/erratum.rb
@@ -26,7 +26,7 @@ module Katello
       end
 
       # rubocop:disable Metrics/AbcSize
-      def update_model(model)
+      def update_model(model, repository = nil)
         updated = false
         keys = %w(title id severity issued_date type description reboot_suggested solution updated_date summary)
         custom_json = backend_data.slice(*keys)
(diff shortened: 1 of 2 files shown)


hao-yu:
can someone review this by any chance?

app/services/katello/pulp3/erratum.rb
```diff
@@ -55,6 +55,14 @@ def update_model(model)
         update_packages(model, backend_data['pkglist']) unless backend_data['pkglist'].blank?
         update_modules(model, backend_data['pkglist']) unless backend_data['pkglist'].blank?
 
+        if !updated && repository.present?
```
ianballou, line 58: I have yet to test this, but what would you think about integrating this into the checks above on line 45 & 46?

```ruby
if model.updated.blank ||
  model.attributes.excluding(model.attributes.keys - custom_json.keys) != custom_json
...
```
I don't see any reason why it would have to run after the body of that `if` statement.  I'm thinking it could be the third or check.

ianballou (commented):
I tested this with `pushcount` and it worked fine. My only suggested change would be to integrate the new if statement into the existing ones above as I mentioned.

app/services/katello/pulp3/erratum.rb
hao-yu: In my opinion, having the long third 'or' check with parenthesis and split across lines is quite hard to read. That is why I split into 2 'if' which looks cleaner and easier to read and understand in my opinion.

ianballou: Alrighty, I don't feel strongly one way or the other.

ianballou (approved):

PR #9858, opened 2021-12-15
Fixes #32379 -  Autopublish switch reloads entire page

Andrewgdewar:
Prevents full page reload when updating information on the content view info page, as well as on the repository review after adding or removing a repo.

- Go to the contentView details page and change name, description and autopublish. 
- Only the change detail item should show a spinner.

- Go to repositories and add/remove repos. (bulk or not)
- Only the repo sub-page should update.

### DIFF
--- a/webpack/components/ActionableDetail.js
+++ b/webpack/components/ActionableDetail.js
@@ -4,6 +4,7 @@ import {
   TextListItemVariants,
   Tooltip,
   TooltipPosition,
+  Spinner,
 } from '@patternfly/react-core';
 import { OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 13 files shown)


sjha4 (approved):

sjha4:
[test katello]

sjha4:
[test katello]

jeremylenz (changes requested):
Thanks for taking a look at this @Andrewgdewar!

I notice the page now doesn't need to reload when you make changes on either Details or Repositories tab. :heavy_check_mark:   But I'm still hoping it can be one step better. As I wrote in the original bug,

"When I toggle autopublish for a content view, it shouldn't have to display a spinner for the entire page. The spinner should be localized to the switch element only."

I'm still seeing the spinner visually replace the entire form. I was hoping the visual disruption could be either eliminated, or localized to only the form field you're editing-- whether that's a spinner, skeleton, or etc.

sjha4:
[test katello]

Andrewgdewar:
@jeremylenz  Done!

jeremylenz (changes requested):
When I submit an edit, there's one render where the spinner is ridiculously large:

spinner_huge

To see it, put a debugger on line 58 of `EditableTextInput.js` before the `await`.  Or set the CPU to 6x slowdown in the Performance tab of chrome dev tools.

I think it's something to do with the `<EmptyState>` inside `<Loading>`, but not sure..

Andrewgdewar:
> When I submit an edit, there's one render where the spinner is ridiculously large:
> 
> spinner_huge
> 
> To see it, put a debugger on line 58 of `EditableTextInput.js` before the `await`. Or set the CPU to 6x slowdown in the Performance tab of chrome dev tools.
> 
> I think it's something to do with the `<EmptyState>` inside `<Loading>`, but not sure..

I was hoping to get away without touching details. I'll take a look.

Andrewgdewar:
@jeremylenz  Updated so that the same spinner is used in all inline editable fields.
You can see this working in filterDetails (for name/description) as well as in ContentViewVersionDetails for description.

jeremylenz (approved):
Works well now, LGTM :+1:

PR #9860, opened 2021-12-15
Fixes #34155 - block pulp actions if switchover has run

jlsherrill:
but we are still using pulp2

It is possible that users could run the content migration and switchover, but accidentally stay on pulp2.   If this happens and the user syncs content, repositories will be reindexed with pulp2 content units, while the current content units have been migrated to pulp3.   Then when they try to do a switchover again, it will fail.  This adds a check that prevents some operations if they are running in 'pulp2' mode, but some rpm content has already been 'switched' over.

1.  install katello 3.18/6.9 (with pulp2 enabled)
2. sync some repositories
3. run  'foreman-maintain content prepare'
4. run 'systemctl start pulpcore-api pulpcore-content pulpcore-resource-manager pulpcore-worker@1 pulpcore-worker@2'
5. run foreman-rake katello:pulp3_switchover
6. try to sync a repository, publish a cv, promote a cv, create a repository

### DIFF
--- a/app/lib/actions/katello/content_view/promote.rb
+++ b/app/lib/actions/katello/content_view/promote.rb
@@ -2,6 +2,8 @@ module Actions
   module Katello
     module ContentView
       class Promote < Actions::EntryAction
+        middleware.use Actions::Middleware::SwitchoverCheck
+
         def plan(version, environments, is_force = false, description = nil, incremental_update = false)
           action_subject(version.content_view)
           version.check_ready_to_promote!(environments)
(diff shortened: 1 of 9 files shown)


jturel:
@jlsherrill remind me please how to put my 3.18 back into pulp2 mode

jlsherrill:
@jturel i would honestly just test on 6.9, but the option to use 3.18 is:

```
foreman-installer --foreman-proxy-content-proxy-pulp-isos-to-pulpcore=false \
              --foreman-proxy-content-proxy-pulp-yum-to-pulpcore=false \
              --katello-use-pulp-2-for-file=true \
              --katello-use-pulp-2-for-docker=true \
              --katello-use-pulp-2-for-yum=true
```

jturel (approved):
Works well!

PR #9861, opened 2021-12-15
Fixes #34157 - Accept URL params on Host details page tabs

jeremylenz:

When loading host detail tabs, take into account URL params.  So
```
#/Repository%20sets?search=name+~+empty&show=noLimit
```

will get your search input filled in with
```
name ~ empty
```
and your toggle group selected "Show all".

Capabilities added to the following tabs:
* Errata
* Repository sets
* Traces
* Packages

This is done through a new hook, `useUrlParams`, which can take the URL search params and transform them into a useful object.

* Also, added links to the Errata Overview card demonstrating usage.

* Interacting with the page won't change the URL. I wanted to avoid infinite loops and multiple sources of truth.
* I realized that switching tabs clears out the URL params, so we got that for free.

Try the following URLs and variations of them (note you may have to refresh the page after entering each url):
* `#/Repository%20sets?search=name+~+empty&show=noLimit` <-- Do this with a non-library host. Note that it will select 'No limit' in the toggle group
* `#/Traces?search=app_type+=+session`
* `#/Content/errata?severity=none` <-- note "none" and not "N/A". Also note that this selects N/A in the dropdown.
* `#/Content/errata?search=issued++%3C++%22Dec+09,2021%22` <-- note that the URL params are transformed into friendly text
* `#/Content/errata?type=security&search=issued++<++"Dec+09,2021"` <-- note multiple params (type & search)
* `#/Content/packages?search=acl`

### DIFF
--- a/webpack/components/Table/TableHooks.js
+++ b/webpack/components/Table/TableHooks.js
@@ -117,6 +117,8 @@ export const useBulkSelect = ({
   results,
   metadata,
   initialArry = [],
+  initialSearchQuery = '',
+  routerHistory,
   idColumn = 'id',
   isSelectable,
 }) => {
@@ -125,7 +127,7 @@ export const useBulkSelect = ({
                   results, metadata, initialArry, idColumn, isSelectable,
                 });
   const exclusionSet = useSet([]);
-  const [searchQuery, updateSearchQuery] = useState('');
+  const [searchQuery, updateSearchQuery] = useState(initialSearchQuery);
   const [selectAllMode, setSelectAllMode] = useState(false);
   const selectedCount = selectAllMode ?
     Number(metadata.selectable) - exclusionSet.size : selectOptions.selectedCount;
@@ -203,6 +205,13 @@ export const useBulkSelect = ({
       selectNone();
     }
   }, [searchQuery, selectAllMode, prevSearchRef, selectNone]);
+
+  // useEffect(() => {
+  //   routerHistory.replace({
+  //     ...routerHistory,
+  //     search: `?search=${encodeURIComponent(searchQuery)}`,
+  //   });
+  // }, [searchQuery, routerHistory]);
 
   return {
     ...selectOptions,
--- a/webpack/components/Table/TableHooks.js
+++ b/webpack/components/Table/TableHooks.js
@@ -223,11 +223,11 @@ export const useBulkSelect = ({
   };
 };
 
-// takes a url query like ?type=security&search=name=foo
+// takes a url query like ?type=security&search=name+~+foo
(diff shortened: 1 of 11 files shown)


webpack/utils/helpers.js
```diff
@@ -121,3 +121,5 @@ export default {
   getResponseErrorMsgs,
   apiError,
 };
+
+export const friendlySearchParam = searchParam => decodeURIComponent(searchParam.replace(/\+/g, ' '));
```
jeremylenz, line 125: see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent#decoding_query_parameters_from_a_url

jeremylenz:
When https://github.com/Katello/katello/pull/9804 is merged I will update the links on the overview card to link directly to the Errata with types filtered.

jeremylenz:
Updated to fix a Redux bug.  In all our tabs we have something like
```js
    params => (hostId ? getInstalledPackagesWithLatest(hostId, params) : null),
    [hostId],
  );
```

But `fetchItems` should never return `null` because it's dispatched as a Redux action, and Redux actions must always be objects with a `type` key. This results in errors like this when you're loading the page from scratch:
```
Uncaught TypeError: Cannot read properties of null (reading 'type')
    at IntervalMiddleware.js:12
```

In place of `null`, I made a `NOOP_HOST_ID_NOT_READY`  action

webpack/components/Table/TableHooks.js
```diff
@@ -219,3 +222,20 @@ export const useBulkSelect = ({
...
   };
 };
+
+// takes a url query like ?type=security&search=name+~+foo
+// and returns an object
+// {
+//   type: 'security',
+//   searchParam: 'name ~ foo'
+// }
+export const useUrlParams = () => {
```
jeremylenz, line 232: side note, This is a good candidate to donate to Foreman :)

jeremylenz:
Errata card links added, marking ready for review

jeremylenz:
Rebased & added one more link

jeremylenz:
[test katello]

jturel:
Looks good & works well! I noticed one quirk: after entering a url in the browser the tab I've navigated to isn't selected. In the screenshot below I changed from the Traces to the Repository sets tab by changing the url directly:

image


Should see a blue highlight on the repo sets tab

jeremylenz:
@jturel Good catch! Turns out that requires a code change in Foreman: https://github.com/theforeman/foreman/pull/9035

jturel (approved):

PR #9864, opened 2021-12-16
API support for bulk removing versions

sjha4:
Extends CV controller to accept bulk params for deleting versions


sjha4:
Closing this PR and will open with just the API changes and possibly the action. The rest of the JS changes are useful for testing the API if needed.

PR #9868, opened 2021-12-20
Fixes #34183  -  Add generic content pages to CVV details

Andrewgdewar:

- [x] Add new subtab on CVV page to support python & ostree
- [x] Make all generic content for the CVV page generated dynamically from the webpack/scenes/Content/ContentConfig.js

- View the new ostree page at content/ostree_refs
- View CVV page after publishing ostree/python content to a contentView to see those appear there.

Screen Shot 2022-01-05 at 3 23 12 PM
Screen Shot 2022-01-05 at 3 23 31 PM
Screen Shot 2022-01-05 at 3 24 47 PM

#### Additional Changes: 
- Fixed a bug around translations that would make routing not work for contentViewVersionDetails.

### DIFF
--- a/lib/katello/repository_types/ostree.rb
+++ b/lib/katello/repository_types/ostree.rb
@@ -27,6 +27,7 @@ Katello::RepositoryTypeManager.register('ostree') do
                        model_name: lambda { |pulp_unit| pulp_unit["name"] },
                        model_version: lambda { |pulp_unit| pulp_unit["checksum"] },
                        uploadable: true,
+                       generic_browser: true,
                        repository_import_on_upload: true
 
   import_attribute :ref, :content_type => 'ostree_ref',
--- a/webpack/scenes/Content/ContentConfig.js
+++ b/webpack/scenes/Content/ContentConfig.js
@@ -1,5 +1,4 @@
 import React from 'react';
-import { Link } from 'react-router-dom';
 import { urlBuilder } from 'foremanReact/common/urlHelpers';
 import { translate as __ } from 'foremanReact/common/I18n';
 import ContentInfo from './Details/ContentInfo';
@@ -7,7 +6,8 @@ import LastSync from '../ContentViews/Details/Repositories/LastSync';
 import ContentRepositories from './Details/ContentRepositories';
 import ContentCounts from '../ContentViews/Details/Repositories/ContentCounts';
 
-export default () => [
+// Keep in mind when editing this file that the ContentViewVersionDetailConfig consumes this array.
+export default [
   {
     names: {
       pluralTitle: __('Python Packages'),
@@ -18,7 +18,7 @@ export default () => [
       singularLabel: 'python_package',
     },
     columnHeaders: [
-      { title: __('Name'), getProperty: unit => (<Link to={urlBuilder(`content/python_packages/${unit?.id}`, '')}>{unit?.name}</Link>) },
+      { title: __('Name'), getProperty: unit => (<a href={urlBuilder(`content/python_packages/${unit?.id}`, '')}>{unit?.name}</a>) },
       { title: __('Version'), getProperty: unit => unit?.version },
       { title: __('Filename'), getProperty: unit => unit?.filename },
     ],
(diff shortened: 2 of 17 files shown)


lib/katello/repository_types/ostree.rb
```diff
@@ -27,6 +27,7 @@
                        model_name: lambda { |pulp_unit| pulp_unit["name"] },
                        model_version: lambda { |pulp_unit| pulp_unit["checksum"] },
                        uploadable: true,
+                       generic_browser: true,
```
rverdile, line 30: if we want to enable the browser on this PR, should the issue/commit also be updated to say something like "also enables generic content browser for ostree refs"?

webpack/scenes/Content/ContentConfig.js
```diff
@@ -82,6 +83,58 @@ export default () => [
...
+    ],
+    tabs: [
+      {
+        tabKey: 'details',
+        title: __('Details'),
+        getContent: (contentType, id, tabKey) => <ContentInfo {...{ contentType, id, tabKey }} />,
+        columnHeaders: [
+          { title: __('Name'), getProperty: unit => unit?.name },
+          { title: __('Version'), getProperty: unit => unit?.version },
+          { title: __('Additional metadata'), getProperty: unit => unit?.additional_metadata },
```
rverdile, line 98: `additional_metadata` is a nested hash/object, so to use this you'd want to do something like `additional_metadata.sha256`. But in the case of OSTree,  additional_metadata is unused, so I would just remove this.

If you're curious what `additional_metadata` is, look at `lib/katello/repository_types/python.rb` and you can see how additional_metadata is defined. That whole hash is stored in the database as jsonb. For `python_packages` it has those two values, for `ostree_refs` it's `nil`.

Andrewgdewar: awesome, will update. Thanks for this.

lib/katello/repository_types/ostree.rb
Andrewgdewar: I'm happy to add that to the issue if you think it's important. I thought it it was inherent with this change.

rverdile: Now that I've looked at the CVV changes more, I agree with you

rverdile (commented):
When I click on the name of a python package or ostree ref in the CVV details, it takes me to a blank screen. I think the link there could be wrong, maybe it was supposed to be something like `content/python_packages/2/#/details`?

Andrewgdewar:
> 

investigating!

Andrewgdewar:
> When I click on the name of a python package or ostree ref in the CVV details, it takes me to a blank screen. I think the link there could be wrong, maybe it was supposed to be something like `content/python_packages/2/#/details`?

Fixed, pushing shortly.

webpack/scenes/ContentViews/Details/ContentViewDetailActions.js
```diff
@@ -59,6 +60,14 @@ const getContentViewDetails = (cvId, extraParams = {}) => get({
   url: api.getApiUrl(`/content_views/${cvId}`),
 });
 
+export const getGenericContent = (pluralLabel, params) => get({
+  type: API_OPERATIONS.GET,
+  key: genericContentKey(pluralLabel),
+  url: api.getApiUrl(`/${pluralLabel}`),
+  params,
+  errorToast: error => __(`Something went wrong while fetching ${lowerCase(pluralLabel)}! ${getResponseErrorMsgs(error.response)}`),
+});
```
rverdile, line 69: I'm not sure about having "generic" as a prefix to the the variables/functions you have here and elsewhere. I say that because, in the back end, Ansible Collections isn't a "generic" type, which could make the naming pattern here misleading. 

In the other "generic content" UI stuff, it's just called "content". What do you think?

Andrewgdewar: Love that you are bringing the API into my frame of reference, much thank!

rverdile:
[test katello]

Andrewgdewar:
[test katello]

PR #9867, opened 2021-12-17
Fixes #34178 - add generic repository testing

jlsherrill:
and enable an ostree test

#### What are the changes introduced in this pull request?
Adds the ability to have tests as part of a repository type without having to write any tests.

#### Considerations taken when implementing this change?
This adds some testing code (declaration of fixture repo) into the production repository_type definition.  To me this is okay, but wanted to get opinions.

#### What are the testing steps for this pull request?
Tests pass, you like the approach

PR #9871, opened 2021-12-21
Fixes #34187 - OSTree repository update error

rverdile:
Fixes update bug for ostree repositories or any generic repository that's not python.
The generic remote options are supposed to be filtered by content type, but in the case of update "content_type" is not a parameter. The code has been updated to get the content type from either the repository object _or_ the params.

This should normally get a test, but since https://github.com/Katello/katello/pull/9867 is enabling OSTree tests, we may want to leave that out of this.
1. Trying creating/updating an OSTree repository and upstream URL, see there's no error.
2. Try it for Python too just to be sure that's still working.


jjeffers (approved):
Works great!

PR #9872, opened 2021-12-21
Refs #34191 - New host details - Install Packages modal

jeremylenz:
- [x] fix tests
- [x] add katello-agent support
- [x] add tests

package_install_modal

* Adds 'Install packages' action to kebab
* Action brings up modal as per mockup
* You can select packages and choose how to install them

package_install_action_menu

* Existing remote execution feature `katello_package_install` was left alone
* New REX feature `katello_package_install_by_search_query` added - this accepts package IDs as part of the search query so it works with our existing SelectAll infrastructure

To import the new job template:
```rb
# in Rails console
ForemanInternal.all.first.delete
```
Then run
```
bundle exec rails db:seed
```
In Hosts > Job Templates, you'll see a new locked template:
```
Install packages by search query - Katello SSH Default
```

Now you're ready to test.

1. Register a rhel7 host and sync some Red Hat repositories
2. Visit the new host details page > Content > Packages
3. Action kebab > Install packages
4. Test both 'Install via remote execution' and 'Install via customized remote execution'

things to test:
- make sure the list of available packages looks sane - test this with both a library and non-library host
- open the modal, close the modal, open it again - The modal reappears as before, without making a second API request and reloading the data
- When no items are selected, the 'Install' button and dropdown are disabled - it becomes enabled when you select an item
- Click on the 'Install' button - immediately executes 'Install via remote execution'
- Click the dropdown and then click 'Install via customized remote execution' - you are taken to the job invocation page with IDs filled out as the package search query
- A package search query such as `id ^ (32372, 33474)` should result in the correct command in the rendered template: `yum -y install duck armadillo` - test this both with regular and customized REX


app/models/katello/rpm.rb
```diff
@@ -215,6 +215,19 @@ def self.applicable_to_hosts(hosts)
...
+    def self.yum_installable_for_host(host)
+      env = host.lifecycle_environment
+      content_view = host.content_view
+      repos = if env.library? && content_view.default?
+                host.content_facet.bound_repositories.pluck(:id)
+              else
+                Katello::Repository.in_environment(env).in_content_views([content_view])
+              end
+      Katello::Rpm.in_repositories(repos).where.not(name: host.installed_packages.pluck(:name)).order(:name)
+    end
```
jeremylenz, line 229: @jlsherrill Can you tell me if I'm on the right track or off the rails here?

(This method is based off of https://github.com/jeremylenz/katello/blob/276371854df4323200b9b9414a9f7fea36fe5162/app/models/katello/concerns/pulp_database_unit.rb#L77 but I needed the same thing without the `applicable_rpms` part.)

jlsherrill: that looks right, but i'm not sure why you wouldn't just use host.content_facet.bound_repositories ?  That sould always point to the repos the system is using.

app/controllers/katello/api/v2/api_controller.rb
```diff
@@ -149,15 +149,15 @@ def scoped_search_query(query, group)
     end
 
     def scoped_search_total(query, group)
-      scoped_search_query(query, group).length
+      scoped_search_query(query, group).count
```
jeremylenz, line 152: Changing these from `length` to `count` improved response time by huge amounts, like ~ 2.5 seconds. I think it's because `count` does a SQL query where `length` is just Ruby calculating the length.

webpack/components/extensions/HostDetails/Tabs/PackageInstallModal.js
```diff
@@ -0,0 +1,236 @@
...
+      {__('Install via customized remote execution')}
+    </DropdownItem>,
+  ];
+  return (
+    <Dropdown
+      direction={DropdownDirection.up}
+      onSelect={onActionSelect}
+      toggle={
+        <DropdownToggle
+          toggleVariant="primary"
```
jeremylenz, line 50: This is documented in Patternfly but doesn't seem to work here. I think our CSS is overriding it somewhere.

app/models/katello/rpm.rb
jeremylenz: > not sure why you wouldn't just use host.content_facet.bound_repositories

I didn't know what "bound repositories" means :speak_no_evil: 
```rb
repos = if env.library? && content_view.default?
                host.content_facet.bound_repositories.pluck(:id)
              else
                Katello::Repository.in_environment(env).in_content_views([content_view])
              end
```
So you're saying I don't need the `else`?

jeremylenz:
at least one of the test failures is related; feel free to review in the meantime

app/models/katello/rpm.rb
jlsherrill: i believe so yes.  bound_repositories are the repos reported to be in use by the client.

jeremylenz: The method from `pulp_database_unit.rb` that I linked above has that same `else` for non-library hosts, so that's where it came from. I'll remove it, thanks :+1:

app/controllers/katello/api/v2/api_controller.rb
```diff
@@ -141,19 +141,19 @@ def skip_session
...
     def scoped_search_total(query, group)
-      result = scoped_search_query(query, group).count
-      if result.instance_of?(Hash)
-        result.keys.length
+      count_query = scoped_search_query(query, group)
+      if query_has_group_by?(count_query)
+        count_query.length
       else
-        result
+        count_query.count
```
jeremylenz, line 156: After some hours I decided that using `.count` with `.group` just causes too many SQL errors

(example below)
```
PG::UndefinedFunction: ERROR:  function count(character varying, character varying) does not exist
LINE 1: SELECT COUNT(name, stream) AS count_name_stream, "katello_mo...
               ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

Reverted back to using `.length`, but only when there's a GROUP BY clause. This seems to fix it.

jeremylenz:
Test failures are now unrelated :tada:

app/models/katello/rpm.rb
```diff
@@ -215,6 +216,19 @@ def self.applicable_to_hosts(hosts)
         where("#{Katello::Host::ContentFacet.table_name}.host_id" => hosts).distinct
     end
 
+    # Return RPMs that are not installed on a host, but could be installed
+    # the word 'installable' has a different meaning here than elsewhere
```
jeremylenz, line 220: 'Installable' (and 'applicable') have essentially the same meaning in our code base as they do for errata-- 'applicable' means there is a package installed on the host that needs an upgrade, and 'installable' means it's both applicable and available to upgrade based on the host's CV/LCE.  However, here I needed a third term that means "_not_ installed right now, but _could_ be."  I coined "yum_installable" because the idea is that you could run "yum install <package_name>" on the host and it would be successful.  I named everything here accordingly, but suggestions to improve the name are welcome.

jeremylenz:
Tests completed, ready for review :smile:

jeremylenz:
__Changes to testing steps:__

* Test Katello-agent actions
  - Katello-agent option is hidden unless Katello-agent is enabled and available (you can hack this by setting `showKatelloAgent={true}` in `PackagesTab.js`)
  - Katello-agent option is disabled when in select all mode
* Changes to REX action used
  - When _not_ in select all mode, 'Install via customized remote execution' now uses the simpler `katello_package_install` feature, rather than `katello_package_install_by_search_query`. This results in a better user experience, because the user will see the package names on the job invocation form.

app/controllers/katello/api/v2/api_controller.rb
lfu: How about count_query.[size](https://medium.com/@craigsheen/count-vs-length-in-rails-4308e83f6292)?

jeremylenz: `size` does a SQL `SELECT COUNT` query if the relation isn't already loaded into memory. I want to avoid that because if the query also has a `group`, it results in the error above. Using `length` in that case guarantees that it won't do a SQL query.

webpack/components/extensions/HostDetails/HostPackages/HostPackagesConstants.js
```diff
@@ -1,4 +1,5 @@
 export const HOST_PACKAGES_KEY = 'HOST_PACKAGES';
+export const HOST_PACKAGES_INSTALL_KEY = 'HOST_PACKAGES_KATELLO_AGENT_INSTALL';
```
lfu, line 2: How do you find out the value for the key? Where is it got set?

app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -388,6 +388,16 @@ def traces_helpers(search:)
         ::Katello::HostTracer.helpers_for(traces)
       end
 
+      def package_names_for_job_template(feature:, search:)
+        features = ['katello_package_install_by_search']
```
lfu, line 392: Maybe better not tight the feature name into the method which cares only if it is to remove, install or update the packages. So in case the feature name got changed later, this method won't be affected if the first parameter is based on feature type (like remove/install/update) instead of feature name.
But it is up to you.

app/models/katello/rpm.rb
```diff
@@ -215,6 +216,13 @@ def self.applicable_to_hosts(hosts)
...
+    def self.yum_installable_for_host(host)
```
lfu, line 221: There is another method `Katello::Rpm.installable_for_hosts` which gets all upgradable packages for hosts.
Maybe we should normalize all these method names so it is more meaningful and clear to the users.

webpack/components/extensions/HostDetails/HostPackages/HostPackagesConstants.js
jeremylenz: It's set right here!  The API middleware then puts the response in the Redux store under that key. More info here: https://github.com/theforeman/foreman/blob/develop/webpack/stories/docs/api-middleware.stories.mdx

app/models/katello/rpm.rb
jeremylenz: yeah.. Naming things is hard.  see https://github.com/Katello/katello/pull/9872#discussion_r789647196

@jturel @jlsherrill thoughts?

app/models/katello/concerns/host_managed_extensions.rb
jeremylenz: Sure, but I don't think I've ever seen a REX feature name actually change. I'd lean towards the side of trying not to prematurely optimize. But tbh I don't feel super strongly either way.

On the other hand, 'install' and 'remove' are a lot easier to type and read than 'katello_package_install_by_search' and 'katello_package_remove_by_search'. So I'll change it for that reason :)

jeremylenz: Updated

app/models/katello/rpm.rb
lfu: How about `available_for_hosts`? Just some thought.

lfu (approved):
Test worked well. Learned a lot from this PR. Made some comments but no changes are required.
LGTM 👍🏻

app/models/katello/rpm.rb
jturel: I am coming into this conversation out of left field, so please take what I am saying with a grain of salt.

I'd write the method this way:

```
def self.installable_for_host(host:, only_missing_packages: false)
  repos = host.content_facet.bound_repositories.pluck(:id)
  rpms = self.in_repositories(repos).order(:name) # why do we order by name here? if it's for the sake of the view then I feel it's in the wrong place
  rpms = rpms.where.not(name: host.installed_packages.pluck(:name)) if only_missing_packages
  rpms
end
```

Why? I think the method name becomes more consistent with how we use installable elsewhere, and the extra flavor of making it only return packages which are not installed on the host is controlled by the param. Not sure I like the word 'missing' ...

jeremylenz: > `why do we order by name here? if it's for the sake of the view then I feel it's in the wrong place`

It's for readability of the yum install command rendered by the template. Does that count as the view? I don't want to deal with trying to do that in the template and adding more methods to the jail, etc.

`Katello::Rpm` already has an [`installable_for_hosts`](https://github.com/jeremylenz/katello/blob/7961da661073d22c9ae8e3f320186777aa85aa8b/app/models/katello/concerns/pulp_database_unit.rb#L86) method (via pulp_database_unit.rb), but it gets _applicable_ RPMs which by definition are already installed, and then filters them to only enabled repos. I feel like `installable_for_host` as a name is too similar to that method since what it does is so different.

PR #9875, opened 2022-01-04
Fixes #34208 - include mirroring_policy in export metadata

evgeni:

Include mirroring_policy in export metadata.

Without it, import doesn't work as you can't create a repository without a mirroring_policy:

    Validation failed: Mirroring policy Invalid mirroring policy for repository type yum, only additive, mirror_complete, mirror_content_only are valid.

This means that imports of *older* versions of Katello (without this patch) still won't work, but I am not sure we define the exports as a "stable API" at all or require users to be on the same version across the board.

perform a full export/import cycle, it should succeed.

### DIFF
--- a/app/services/katello/pulp3/content_view_version/metadata_generator.rb
+++ b/app/services/katello/pulp3/content_view_version/metadata_generator.rb
@@ -43,7 +43,7 @@ module Katello
 
         def generate_repository_metadata(repo)
           repo.slice(:name, :label, :description, :arch, :content_type, :unprotected,
-                     :checksum_type, :os_versions, :major, :minor).
+                     :checksum_type, :os_versions, :major, :minor, :mirroring_policy).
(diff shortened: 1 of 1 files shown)


evgeni:
cc @jlsherrill - you broke it ;)

evgeni:
[test katello]

jlsherrill:
i agree, its not :/   @parthaa any reasons this isn't testable via vcr tests?

jlsherrill:
on import we may want to just default to some sane mirroring policy, since its really not all that important here.  That would give us backwards compat for this option.

evgeni:
Yeah, I thought about that, but then couldn't find a good spot for it. So feel free to fix it in a better way ;)

jlsherrill:
will work on this here: https://github.com/Katello/katello/pull/9876

PR #9878, opened 2022-01-06
Align on pulp_content_url

ekohl:

The Pulp content URLs should always respect what's exposed in the pulp_content_url setting. This includes the /pulp/content part.

It does highlight a missing setting for pulp_ansible content. Again, Katello should not have to guess URLs but discover it from the infrastructure.

It was extracted from https://github.com/Katello/katello/pull/9413, but it's probably incomplete. Consider it a start of a discussion.

There should be no changes for the user in default setup. If the user does change the content URL, it should be properly reflected.

### DIFF
--- a/app/helpers/katello/katello_urls_helper.rb
+++ b/app/helpers/katello/katello_urls_helper.rb
@@ -45,18 +45,16 @@ module Katello
     end
     def repository_url(content_path, _content_type = nil, schema = 'http')
       return content_path if content_path =~ %r|^([\w\-\+]+)://|
-      url = if @host.content_source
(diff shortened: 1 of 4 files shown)


ianballou:
I agree that it would be best to rely on `pulp_content_url` as the source of truth. I think this PR is generally going in a good direction

qcjames53:
@ekohl Would you like us to close out this PR? We're doing a cleanup of the open requests in Katello this afternoon.

ekohl:
@qcjames53 fundamentally I still think Katello should not assume `/pulp/content` is used and derive it from the configured URL, but I'm not sure I have the time to finish this.

PR #9880, opened 2022-01-06
Fixes #34229 - pin faraday until we can update to 2.0

jlsherrill:
the pulp bindings do not support faraday 2.0, work is going on here to support it: https://github.com/pulp/pulp-openapi-generator/pull/56

until then, we need to pin to a lower version

1.9 was also released with a breaking change


jturel (approved):
APJ

PR #9882, opened 2022-01-07
Fixes #34235 - reindex repository after recreating it

jlsherrill:
When a user runs the correct_repositories rake task, and a library repo is found to be missing, it is re-created.  However! if content is indexed into it, it will remain indexed even though the repository is empty.  By reindexing it after creation, we can reset the indexed content.

1) create a yum repository and sync it (i used https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/) , notice the content counts show errata & rpms 
2) in the rails console, find the repo href:  `Katello::Repository.find(9).root.repository_references.first.repository_href`
3) Delete the repo:  
```
curl -X DELETE https://`hostname`/pulp/api/v3/repositories/rpm/rpm/920e7484-212c-4009-8324-3f77fe46caab/ --cert /etc/pki/katello/certs/pulp-client.crt --key /etc/pki/katello/private/pulp-client.key
```
4) run:   bundle exec rake katello:correct_repositories COMMIT=true
5) without this pr, the content counts (rpm, errata) stay the same, with it, they are reset to zero

### DIFF
--- a/lib/katello/tasks/repository.rake
+++ b/lib/katello/tasks/repository.rake
@@ -119,7 +119,10 @@ namespace :katello do
     puts "Repository #{repo.id} Missing"
     if repo.content_view.default?
       puts "Recreating #{repo.id}"
-      ForemanTasks.sync_task(::Actions::Katello::Repository::Create, repo, force_repo_create: true) if commit?
+      if commit?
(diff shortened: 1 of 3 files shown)


ianballou (approved):
Works great for me! Just needs a new VCR.

PR #9884, opened 2022-01-12
Fixes #34168 - Extend CV API to support bulk removing versions

jlsherrill:
What are the changes introduced in this pull request?

Adds support for bulk removing versions from a CV
Considerations taken when implementing this change?

    Ensure the versions deletion works.
    Ensure the reassignment of hosts and activation keys from deleted versions works
    Can select one cv/env combo for reassigning all hosts affected by bulk delete. Same for all affected activation keys.

What are the testing steps for this pull request?

One easy way to test this:
There's the closed PR: https://github.com/Katello/katello/pull/9864/files which switches to use bulk remove API.
Copy the JS changes from there and replace IDs on https://github.com/Katello/katello/pull/9864/files#diff-61fb2d50f776e164f9f3da714007d908978c2f22d80b3c915a576a2930e45d6fR72 with whatever versions you want to bulk delete.

### DIFF
--- a/app/controllers/katello/api/v2/content_views_controller.rb
+++ b/app/controllers/katello/api/v2/content_views_controller.rb
@@ -188,26 +188,22 @@ module Katello
       respond_for_async :resource => task
     end
 
-    api :PUT, "/content_views/:id/bulk_remove", N_("Bulk remove versions from a content view and reassign systems and keys")
+    api :PUT, "/content_views/:id/bulk_delete_versions", N_("Bulk remove versions from a content view and reassign systems and keys")
     param_group :bulk_content_view_version_ids
     param :id, :number, :desc => N_("content view numeric identifier"), :required => true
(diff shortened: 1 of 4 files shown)


app/controllers/katello/api/v2/content_views_controller.rb
```diff
@@ -175,6 +188,46 @@ def remove
       respond_for_async :resource => task
     end
 
+    api :PUT, "/content_views/:id/bulk_remove", N_("Bulk remove versions from a content view and reassign systems and keys")
```
Andrewgdewar, line 191: This api's name 'bulk_remove' isn't really indicative of the resulting actions it takes as it doesn't perform a 'removal' but rather a 'deletion' of versions. 

After discussing with @jlsherrill, this should likely be updated to better reflect its actions.

Andrewgdewar (approved):

PR #9798, opened 2021-11-18
Fixes #33952 - do not audit internal columns

ezr-ondrej:
Disable auditing of Katello internal column.

Found as https://github.com/Katello/katello/pull/9270#issuecomment-972763008

#### What are the changes introduced in this pull request?
Stop auditing internal column `created_in_katello`

#### What are the testing steps for this pull request?
Create org

PR #9888, opened 2022-01-14
Refs #33952 - revert change in org creator

ezr-ondrej:
Revert part of #9798 that was wrong, not auditing this column should be sufficient.


jturel (approved):
Works well. I tested two workflows:

- Org created via hammer (seed and create! happen together) - I got all the right artifacts & no double audit
- Simple org created in rails console. Ran db:seed. verified created_in_katello is still false. Hit the API which triggers the rest of the create. Candlepin owner,etc got created

Thanks @ezr-ondrej !

PR #9889, opened 2022-01-14
Fixes #34257 - cv docker tags unpullable after resync with tag limiting

ianballou:
Now, during content indexing, only the `RepositoryDockerMetaTags` related to the repository being re-indexed are deleted as necessary.

I had to edit one of the test fixtures because it had an invalid repository label. This was stopping me from editing the tags whitelist in the middle of the test.

The bug overall is pretty simple.  When reindexing, the `RepositoryDockerMetaTags` were not being grouped by their respective repositories before being deleted.
A test was added so this wouldn't happen again.

1) Sync a docker repo
2) Add that docker repo to a content view and publish
3) Ensure you can pull from the CV docker repo
4) Edit the repository to have a tag whitelist ("Limit Sync Tags")
5) Re-sync
6) Ensure that the blacklisted tags cannot be pulled from the library+default org cv repo
7) Try pulling the blacklisted tags from the CV docker repo and notice that you cannot.  This is the bug
8) Patch in this PR
9) Remove the whitelist entirely from the docker repo
10) Repeat steps 1-6
11) Try pulling the blacklisted tags again from the CV docker repo. Notice that you can. This means the bug is fixed

### DIFF
--- a/app/models/katello/docker_meta_tag.rb
+++ b/app/models/katello/docker_meta_tag.rb
@@ -153,7 +153,7 @@ module Katello
         end
 
         unless params_to_query_for_delete.empty?
-          RepositoryDockerMetaTag.where(:docker_meta_tag => DockerMetaTag.where(params_to_query_for_delete.join(" OR "))).delete_all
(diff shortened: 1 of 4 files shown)


rverdile (approved):
works well!

PR #9891, opened 2022-01-18
Fixes #34279 - Use APT verbatim publisher for deb content on Pulp 3 foreman-proxy syncs

Manisha15:
Use Apt verbatim publisher so as to eliminate the need to configure signing services on foreman proxies

### DIFF
--- a/app/services/katello/pulp3/repository_mirror.rb
+++ b/app/services/katello/pulp3/repository_mirror.rb
@@ -79,7 +79,11 @@ module Katello
       end
 
       def publication_href
-        api.publications_api.list(:repository_version => version_href).results.first&.pulp_href
+        if repo_service.repo.content_type == "deb"
+          api.publications_verbatim_api.list(:repository_version => version_href).results.first&.pulp_href
+        else
+          api.publications_api.list(:repository_version => version_href).results.first&.pulp_href
+        end
       end
 
       def create_version(options = {})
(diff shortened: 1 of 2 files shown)


app/services/katello/pulp3/repository_mirror.rb
```diff
@@ -171,8 +171,13 @@ def publication_options(repository_version)
 
       def create_publication
         if (href = version_href)
-          publication_data = api.publication_class.new(publication_options(href))
-          api.publications_api.create(publication_data)
+          if repo_service.repo.content_type == "deb" && !smart_proxy.pulp_primary?
```
ianballou, line 174: `!smart_proxy.pulp_primary?` is redundant since `with_mirror_adapter` prevents `repository_mirror.rb` code from talking to the Pulp primary smart proxy.

ianballou (changes requested):
The method `publication_href` needs to be edited to used the verbatim API as well:

```ruby
def publication_href
  api.publications_verbatim_api.list(:repository_version => version_href).results.first&.pulp_href
end
```

quba42:
@ianballou I can see the `publication_href` method you mentioned being used as part of `refresh_distributions`. Do you know what exactly `refresh_distributions` does? I am trying to understand how we missed this during our own internal testing.

app/services/katello/pulp3/repository_mirror.rb
Manisha15: Fixed!

ianballou:
`refresh_distributions` essentially ensures that the distribution is created and up to date for a repository. It's part of metadata generation.

I'm not sure how it would've worked before without it. I believe I caught this after testing a normal Debian content sync and saw an error.

ianballou (approved):
Looking fine to me now!

PR #9892, opened 2022-01-18
Fixes #33670 - Get pulp username/password from katello settings

rverdile:
Previously, the username/password for pulp would be fetched from the smart proxy. Instead, this will get the username/password from `foreman/config/settings.plugins.d/katello.yml`.

Like this:

> :pulp_auth:
    - url: "https://proxy-1.localhost.example.com"
      username: "proxy1 username"
      password: "proxy1 password"
    - url: "https://proxy-2.localhost.example.com"
      username: "proxy2 username"
      password: "proxy2 password"

This fix is to avoid exposing the password on the smart proxy's `/v2/features` endpoint. Although exposing the password is the main issue, I changed both the username and password fields because they were originally exposed for the same purpose. This feature is only intended for developers as a convenient way to set up an external pulp server. 

1. Don't apply this PR yet.
2. Load up a capsule: https://github.com/theforeman/forklift/blob/master/docs/development.md#capsule-development. Make sure it's connected to your katello devel server properly.
3. SSH into capsule. Add username/password fields to this file: `/etc/foreman-proxy/settings.d/pulpcore.yml`
4. In Katello box, refresh smart proxy, then check username/password of smart proxy in console with: `PulpcoreClient::ApiClient.new(SmartProxy.find(2).pulp3_configuration(PulpcoreClient::Configuration))`.
5. See that it matches what you made the fields in step 4.
6. Apply this PR
7. Add `:pulp_auth:` to `katello.yml` as shown above, using the url of the capsule and a different username/password
8. In the console, check that the username/password for the capsule matches what you put in `katello.yml` (and not on the smart proxy).

### DIFF
--- a/app/models/katello/concerns/smart_proxy_extensions.rb
+++ b/app/models/katello/concerns/smart_proxy_extensions.rb
@@ -180,8 +180,9 @@ module Katello
           pulp3_ssl_configuration(config)
           config.debugging = false
           config.logger = ::Foreman::Logging.logger('katello/pulp_rest')
-          config.username = self.setting(PULP3_FEATURE, 'username')
-          config.password = self.setting(PULP3_FEATURE, 'password')
+          pulp_auth = SETTINGS[:katello][:pulp_auth].find { |proxy| proxy['url'] == uri.scheme + "://" + uri.hostname }
+          config.username = pulp_auth['username']
+          config.password = pulp_auth['password']
         end
       end
 


ekohl (changes requested):
:-1: on this change tbh. Please don't move forward with it.

In production there are no username and password settings. It's only for development.

We also have multiple servers, so this implies all servers have the same credentials. I'd also like to get rid of configuration in katello.yml.

rverdile:
> -1 on this change tbh. Please don't move forward with it.
> 
> In production there are no username and password settings. It's only for development.
> 
> We also have multiple servers, so this implies all servers have the same credentials. I'd also like to get rid of configuration in katello.yml.


Thanks for your feedback! I just want to get some clarity on what you're saying. 

First, like you mentioned, this change is only meant for use in development. What is the cause for concern there?

Also, why would this imply multiple servers have the same credentials? I used multiple servers in the configuration example in my intial PR comment. Perhaps I am misunderstanding something about the change.

ekohl:
> First, like you mentioned, this change is only meant for use in development. What is the cause for concern there?

It means the security aspect isn't that critical.

> Also, why would this imply multiple servers have the same credentials? I used multiple servers in the configuration example in my intial PR comment. Perhaps I am misunderstanding something about the change.

This is building yet another mechanism for configuration. With exposed settings we're finally getting to a point where the static config is no longer used and everything is dynamic. This is a huge step back.

Let me describe the larger vision that I have. Smart Proxies are essentially also a form for service discovery and make the application extendible at runtime.

Ideally speaking you can install a Foreman, add plugins (such as Katello). It will detect that there is no Smart Proxy with the content feature. The UI tells the user some actions can't be performed until such a proxy is added. So you create a new Smart Proxy with the content feature and now you can use content features.

Note that at no point the admin has changed anything in Foreman's configuration. This will allow deployment of Foreman in environments where you can't (or don't want to) manage configuration files. Examples are load balanced setups where there are multiple application servers. Those load balanced setups may also live in containers in Kubernetes or similar environments. Within the platform team we've had discussions about some SaaS model where a new instance is provisioned via some automation. That really needs this self management from the application.

We're far from that point, but that is the direction I'm trying to push things. This is moving in the opposite direction which is why I said :-1: on it.

jlsherrill:
@ekohl i agree with a lot of your criticisms.  Do you have any idea how to enable this?  Being able to use a username/password makes development against a sandbox pulp server much simpler.  Atix added this functionality for their own development and it has made our development simpler as well.    (And i agree it should allow for different usernames/passwords for each smart proxy).

This is never intended to be used in a production scenario and was seen as a security improvement over the current model. We can take this further to prevent it from being used in production.  I agree with your vision but its a vision for a production deployment, a development environment is very different, has a different 'user'.    Is the security complaint low enough that we should ignore it? 

What do you suggest?

ekohl:
Other than that, there's some things I disagree with in the Redmine issue. For example:

> I think the plugin should just not `expose_setting :password`, as there is really no reason for Foreman to know the password.

As we can see in this PR, there very much is a reason.

Another thing to consider is that right now exposed settings are still internal. They should not be exposed with Foreman's REST API for example. We explicitly chose not to expose the Smart Proxy's exposed settings in Foreman's API because they could contain secrets. And given that at some point the Foreman codebase does need the username/password, it must be present unencrypted somehow.

If you want to treat it as secret, perhaps the field should be encrypted. That would address the SQL query issue that was mentioned in Redmine (which is probably valid). Foreman has a mechanism to encrypt fields. For example:
https://github.com/theforeman/foreman/blob/7b1395c2ce0a200ebf549c156ba132678ebbb99e/app/models/http_proxy.rb#L12

You will probably need a DB migration to convert it, otherwise it gets done the first time the value is written again. I can't find an example right now.

So my advice would be to consider it a Foreman core issue that the settings at rest are unencrypted while they can contain secrets. Other than that I'd consider it working as intended.

PR #9893, opened 2022-01-18
Fixes #34285 - Fix inconsistent Katello test failures

Andrewgdewar:

- With the new Search.js Component functionality, the default autosearch settings we had set within our tests needed to be updated. 

- Spin up the branch, run `npx jest webpack` and ensure that all tests consistently pass.


webpack/components/extensions/HostDetails/Tabs/__tests__/errataTab.test.js
```diff
@@ -81,11 +81,10 @@ let searchDelayScope;
 let autoSearchScope;
 
 beforeEach(() => {
-  // jest.resetModules();
   const { results } = mockErrataData;
   [firstErrata, , thirdErrata] = results;
-  searchDelayScope = mockSetting(nockInstance, 'autosearch_delay', 500);
-  autoSearchScope = mockSetting(nockInstance, 'autosearch_while_typing', true);
+  searchDelayScope = mockSetting(nockInstance, 'autosearch_delay', 0);
```
jturel, line 86: I guess the idea here is that we're not testing the auto_delay functionality in all of these tests, so why use a value of 500 at all?

jturel (approved):
LGTM. I bumped my nock timeout down to 7 secs and ran the tests a few times to see if we have good 'headroom' with timing out. The only failures I saw were around the select and deselect tests (which are not the ones we saw fail in nightly nor are addressed here)

ACK!

webpack/components/extensions/HostDetails/Tabs/__tests__/errataTab.test.js
Andrewgdewar: 500 is default yeah, so we avoid using the delay. 
In most cases that delay is never actually seen in the tests, but I did notice a small reduction in test times. 
The main change was the` 'autosearch_while_typing', true);` bit. 
That makes all of the tests use the autcompleteSearch functionality as it is seen today within contenViews.

PR #9896, opened 2022-01-20
Fixes #34294 - container resync tags test failing on VCR

ianballou:
#### What are the changes introduced in this pull request?
Adds a proper VCR for a new docker limit sync tags test.

#### Considerations taken when implementing this change?
Nope.

#### What are the testing steps for this pull request?
See that the tests pass.

PR #9897, opened 2022-01-20
Fixes #34295 - add default download policy

rverdile:
Adds default download policy to non-yum types using download policy. Also adds migration to assign download policy to old docker repos.
The default download policy behavior is the same as the yum type. Without this, when creating a repository with hammer, the download policy is "nil" if you don't specify it.

Without the migration, the download policy for docker repos would be "". I made the migration irreversible because I figured if someone accidentally rolled-back their database, resetting the policy to "" would be an issue.
1. To test the default download policy, user hammer to create a docker repo, without using the --download-policy option.
  - `hammer repository create --product-id [id] --content-type docker --url "https://quay.io" --docker-upstream-name "quay/busybox" --name "docker repo"`
 2. Sync that repo. Without the change, you would get an error saying "policy cannot be nil".
 3. To test the migration, set your katello branch to before the recent "Add On Demand Policy for Docker" commit.
 4. Create a docker repo.
 5. Reload this PR
 6. Notice the docker repo you created has "" as the policy
 7. Run migration
 8. Now it should be "immediate"

you could also use this `Katello::RootRepository.where(:content_type => "docker").update_all(:download_policy => "")` to get "" as the download policy.

### DIFF
--- a/app/controllers/katello/api/v2/repositories_controller.rb
+++ b/app/controllers/katello/api/v2/repositories_controller.rb
@@ -51,7 +51,7 @@ module Katello
       param :checksum_type, String, :desc => N_("Checksum of the repository, currently 'sha1' & 'sha256' are supported")
       param :docker_upstream_name, String, :desc => N_("Name of the upstream docker repository")
       param :docker_tags_whitelist, Array, :desc => N_("Comma-separated list of tags to sync for Container Image repository")
-      param :download_policy, ["immediate", "on_demand"], :desc => N_("download policy for yum repos (either 'immediate' or 'on_demand')")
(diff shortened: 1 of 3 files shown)


jlsherrill:
overall looks good, i think that one test needs a vcr update

rverdile:
@jlsherrill should this merge first? https://github.com/Katello/katello/pull/9896

jlsherrill:
yeah, probably so!

jlsherrill:
[test katello]

jlsherrill:
[test katello]

transient failure is transient

jlsherrill (approved):

PR #9901, opened 2022-01-24
Fixes #34220 - Packages bulk select & remove and upgrade

lfu:
- Packages bulk select & remove
- Packages Upgrade

- Followed the implementation in https://github.com/Katello/katello/pull/9652 and https://github.com/Katello/katello/pull/9872   
- Add REX feature katello_packages_remove_by_search - this accepts package IDs as part of the search query
- Add REX feature katello_packages_update_by_search - this accepts package IDs as part of the search query

- Expect packages bulk select and remove should work.
  REX and Katello agent.  
- Expect removing a single package should work.
  Both REX and Katello Agent should work 
- Expect packages bulk select and upgrade should work.
- Expect upgrading a single package should work. 

**To setup the testing environment**
To import the new job templates:
```
ForemanInternal.last.destroy
```
Then run
```
bundle exec rails db:seed
```
In Hosts > Job Templates, you'll see the new templates:
```
Remove Packages by search query - Katello SSH Default
Update Packages by search query - Katello SSH Default
```


lfu:
Screenshot from 2022-01-26 10-23-08

lfu:
Screenshot from 2022-01-26 10-23-36

lfu:
Screenshot from 2022-01-26 10-23-59

lfu:
Screenshot from 2022-01-26 10-26-51

webpack/components/extensions/HostDetails/Tabs/__tests__/packagesTab.test.js
```diff
@@ -141,3 +165,363 @@ test('Can filter by package status', async (done) => {
   assertNockRequest(autoSearchScope);
   assertNockRequest(scope2, done); // Pass jest callback to confirm test is done
 });
+
+test('Can select one package', async (done) => {
```
jeremylenz, line 169: I do hope it wasn't too much work adapting all these tests from ErrataTab.  All of the tests you've added here aren't really specific to Packages. Rather, they test the Select All functionality, and so I don't think we need to duplicate them everywhere. I'll leave it up to you whether to delete them or not, but if they weren't there I would not ask for them to be added.  Perhaps it would have helped if the existing tests were in a "describe" block:
```js
describe('Select All functionality'), () =>
  test('Can select one', async (done) => {
...
```

webpack/components/extensions/HostDetails/Tabs/__tests__/packagesTab.test.js
```diff
@@ -141,3 +165,363 @@ test('Can filter by package status', async (done) => {
...
+  assertNockRequest(scope, done); // Pass jest callback to confirm test is done
+});
+
+test('Can select all packages across pages through checkbox', async (done) => {
+  // Setup autocomplete with mockForemanAutoComplete since we aren't adding /katello
+  const autocompleteScope = mockForemanAutocomplete(nockInstance, autocompleteUrl);
+  const mockPackages = makeMockPackages({ page: 1 });
+  // return packages data results when we look for packages
+
+  // return packages data results when we look for packages
```
jeremylenz, line 208:
```suggestion
```

webpack/components/extensions/HostDetails/Tabs/__tests__/packagesTab.test.js
```diff
@@ -117,7 +141,7 @@ test('Can filter by package status', async (done) => {
     getByText,
   } = renderWithRedux(<PackagesTab />, renderOptions());
 
-  // Assert that the errata are now showing on the screen, but wait for them to appear.
+  // Assert that the packages are now showing on the screen, but wait for them to appear.
```
jeremylenz, line 144: In general I think these test comments aren't too useful. We should probably delete them so we stop copy/pasting them everywhere. :smile:

jeremylenz (commented):
Looks great @lfu!  Tested and all REX actions are generated as expected. All of our established patterns are followed nicely and I don't really have too much to say about the code. But I did leave one comment below about tests.

Also, I'd like to see tests added for the specific Package functionality:
* Can upgrade a package via remote execution
* Can upgrade a package via customized remote execution
* Can bulk upgrade via remote execution (this one will test the bulk action kebab at the top)
* Can bulk upgrade via customized remote execution
* Upgrade is disabled when there are non-upgradable packages selected

webpack/components/extensions/HostDetails/Tabs/PackagesTab.js
```diff
@@ -73,20 +89,117 @@ export const PackagesTab = () => {
...
+    contentFacet.katelloAgentEnabled);
+
+  const removeBulk = () => {
+    if (contentFacet.remoteExecutionByDefault || !katelloAgentAvailable) {
+      removePackagesViaRemoteExecution();
+    }
+  };
+
+  const removeKatelloPackage = (packageName) => {
+    if (contentFacet.remoteExecutionByDefault || !katelloAgentAvailable) {
```
jeremylenz, line 136: This was just merged; I suggest reusing it here: https://github.com/Katello/katello/blob/d433a901c0247b699ff742791367ecd73dd304e5/webpack/components/extensions/HostDetails/hostDetailsHelpers.js#L6

app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -401,10 +401,16 @@ def traces_helpers(search:)
...
-        actions = ['install']
+        actions = %w(install remove update).freeze
         case action
         when 'install'
           ::Katello::Rpm.yum_installable_for_host(self).search_for(search).distinct.pluck(:name)
+        when 'remove'
+          installed_packages.search_for(search).distinct.pluck(:name)
+        when 'update'
+          pkg_names = installed_packages.search_for(search).distinct.pluck(:name)
+          upgrades = ::Katello::Rpm.installable_for_hosts([self]).select(:id, :name, :nvra, :evr).order(evr: :desc).group_by(&:name)
```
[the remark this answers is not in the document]
lfu: https://stackoverflow.com/questions/28444892/what-does-name-mean-in-ruby

webpack/components/extensions/HostDetails/Tabs/__tests__/packagesTab.test.js
```diff
@@ -133,10 +160,522 @@ test('Can filter by package status', async (done) => {
...
+
+    expect(queryByText('20 selected')).toBeInTheDocument();
+    expect(getByLabelText('Select row 0').checked).toEqual(true);
+
+    assertNockRequest(autocompleteScope);
+    assertNockRequest(scope, done); // Pass jest callback to confirm test is done
+  });
+});
+
+test('Can upgrade a package via remote execution', async (done) => {
```
jeremylenz, line 461: This is where the Select All-related tests end, so this and subsequent tests should be outside the describe block.

lfu: Right. That is the case.

jeremylenz: Oops, missed that second `});` :smile:

app/models/katello/concerns/host_managed_extensions.rb
jeremylenz, line 412: This is resulting in lots of whitespace in certain situations:
```
yum -y update                                       armadillo-2.1-1.noarch
```
The reason is that we need to remove `nil`s from what it returns here:
```
[3] pry(main)> mh.package_names_for_job_template(action: 'update', search: '').inspect
=> "[nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, \"armadillo-2.1-1.noarch\", nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil]"
```

app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -401,10 +401,16 @@ def traces_helpers(search:)
...
+          pkg_names.map { |p| upgrades[p]&.first&.nvra }
```
jeremylenz, line 413: When the search is blank (meaning all are selected), I'd go even further and return `[]`. That way the resulting command is just
```
yum -y update
```
instead of 
```
yum -y update package1 package2 package3 .... package572
```

jeremylenz (commented):
Left a couple more comments / issues below.

For the tests, let's do it right and extract the Select All functionality into its own file. Let's say `SelectAll.test.js`. This file can contain the Select All tests either from Packages, or from Errata (you pick).  And we can remove the other Select All tests.

app/views/foreman/job_templates/remove_packages_by_search_query.erb
```diff
@@ -0,0 +1,19 @@
+<%#
+kind: job_template
+name: Remove Packages by search query - Katello SSH Default
+job_category: Katello
+description_format: 'Remove packages %{Packages search query}'
+feature: katello_packages_remove_by_search
+provider_type: SSH
+template_inputs:
+- name: Packages search query
+  description: Filter criteria for packages to be removed.
```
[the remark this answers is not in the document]
lfu: If no package is specified, I think it should not remove anything.
Nice catch on the edge case. 👍🏻

app/models/katello/concerns/host_managed_extensions.rb
lfu: Nice catch.

app/views/foreman/job_templates/remove_packages_by_search_query.erb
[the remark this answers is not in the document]
jturel: Agreed - that would be incredibly dangerous :)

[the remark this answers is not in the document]
jeremylenz: Should we disable select all in the UI for bulk package remove?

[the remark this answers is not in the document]
jeremylenz: ...and make 'Packages search query' a required field in the template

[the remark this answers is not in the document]
lfu: @jeremylenz I like these ideas.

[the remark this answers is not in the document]
jeremylenz: I've convinced myself.  I think the 'remove' option should be disabled when in select all mode, and 'Packages search query' should be a required field for the remove packages template.

webpack/components/extensions/HostDetails/Tabs/PackagesTab.js
```diff
@@ -96,6 +226,42 @@ export const PackagesTab = () => {
...
+    <Split hasGutter>
+      <SplitItem>
+        <ActionList isIconList>
+          <ActionListItem>
+            <Dropdown
+              onSelect={onActionSelect}
+              toggle={
+                <DropdownToggle
+                  splitButtonItems={[
+                    <DropdownToggleAction key="action" aria-label="upgrade_actions" onClick={upgradeBulkViaRemoteExecution}>
```
jeremylenz, line 239: The Upgrade button should use Katello agent if that's what `defaultRemoteActionMethod` returns.  (This will align it with the behavior of the Apply button on the ErrataTab.)

Also, since the katello agent API takes package names, it should be disabled when in select all mode.

webpack/components/extensions/HostDetails/Tabs/PackagesTab.js
```diff
@@ -73,20 +86,137 @@ export const PackagesTab = () => {
...
+
+  const removeBulk = () => {
+    if (defaultRemoteAction === KATELLO_AGENT) {
+      removePackagesViaKatelloAgent();
+    } else {
+      removePackagesViaRemoteExecution();
+    }
+  };
+
+  const removeKatelloPackage = (packageName) => {
```
jeremylenz, line 147: There's no such thing as a "Katello package".. :thinking: 

How about `handlePackageRemove`?
```suggestion
  const handlePackageRemove = (packageName) => {
```

jeremylenz (commented):
Upgrade disablement looking better :+1: 

Left some more comments below

webpack/components/extensions/HostDetails/Tabs/PackagesTab.js
jeremylenz: But I think Katello agent should still work. I will ask Maria and see if I can get some feedback..

jeremylenz (commented):
@lfu Thanks for your patience!

We're almost at the finish line with katello-agent. Just a few more things and then I think we're golden.

webpack/components/extensions/HostDetails/Tabs/PackagesTab.js
```diff
@@ -73,20 +91,157 @@ export const PackagesTab = () => {
...
+    if (areAllRowsSelected()) {
+      dispatch(upgradeAllViaKatelloAgent(hostId));
+    } else {
+      dispatch(upgradePackageViaKatelloAgent(hostId, { packages: selectedUpgradableVersions() }));
+    }
+    selectNone();
+  };
+
+  const upgradeBulk = () => {
+    if (!selectAllMode && defaultRemoteAction === KATELLO_AGENT) {
```
jeremylenz, line 187: The other code changes look good, but I think this still needs to be addressed?
```suggestion
    if (defaultRemoteAction === KATELLO_AGENT) {
```
I tested it and without this change, it still tries to use remote execution if I select all and click Upgrade.

jeremylenz (commented):
Sorry for the delay! One comment below.

jeremylenz (approved):
Awesome work @lfu!

ACK :+1:

PR #9902, opened 2022-01-24
Fixes #34306 - Update how we parse CN from subman identity certs

chris1984:

With the new build of Candlepin 4.1.10 they have changed the way entitlement certificates are generated. Before they would look like:

`Subject: CN=eb48d5a8-b759-417c-97f7-93dc2369de29`

Now the new cert looks like:

`Subject: O=Default_Organization, CN=eb48d5a8-b759-417c-97f7-93dc2369de29`

So, we get unauthorized because of the way we parse the ID which now comes back looking like:

`"/O=Default_Organizationeb48d5a8-b759-417c-97f7-93dc2369de29"`

This issue is to address the way we grab the CN from the cert, so it works for the latest version of Candlepin and the older versions.

Make sure to test with the current version of Candlepin and the updated version here:

https://bugzilla.redhat.com/show_bug.cgi?id=2034349

Steps to upgrade your devbox are here:

https://theforeman.org/plugins/katello/developers.html#upgrading-candlepin

ignore the glue tests and the other manual testing steps as I am doing that as part of the candlepin upgrade once this gets merged. See steps below what needs to be done.

Here are the benchmark stats from the way purposed in the pr to the first way I created. Props to jturel++ for finding the 2nd way.

```ruby
Benchmark.bm { |x| x.report { @cert.subject.to_a.select{|name, _, _| name == 'CN' }.first[1] }}
17:39:22 rails.1   |        user     system      total        real
17:39:22 rails.1   |    0.000127   0.000019   0.000146 (  0.000140)
17:39:22 rails.1   | => [#<Benchmark::Tms:0x00007fdd00c90c18
17:39:22 rails.1   |   @cstime=0.0,
17:39:22 rails.1   |   @cutime=0.0,
17:39:22 rails.1   |   @label="",
17:39:22 rails.1   |   @real=0.0001398190506733954,
17:39:22 rails.1   |   @stime=1.8999999999991246e-05,
17:39:22 rails.1   |   @total=0.00014599999999997948,
17:39:22 rails.1   |   @utime=0.00012699999999998823>]
```

```ruby
Benchmark.bm { |x| x.report { subject_string.sub!(/\/O=Default_Organization/i, '') }}
[1] pry(#<Cert::RhsmClient>)> Benchmark.bm { |x| x.report { subject_string.sub!(/\/O=Default_Organization/i, '') }}
17:47:39 rails.1   |        user     system      total        real
17:47:39 rails.1   |    0.000017   0.000006   0.000023 (  0.000015)
17:47:39 rails.1   | => [#<Benchmark::Tms:0x000000000b9fab90
17:47:39 rails.1   |   @cstime=0.0,
17:47:39 rails.1   |   @cutime=0.0,
17:47:39 rails.1   |   @label="",
17:47:39 rails.1   |   @real=1.4631019439548254e-05,
17:47:39 rails.1   |   @stime=6.0000000000060005e-06,
17:47:39 rails.1   |   @total=2.3000000000050758e-05,
17:47:39 rails.1   |   @utime=1.7000000000044757e-05>]
```

```ruby
Benchmark.bm { |x| x.report { subject_string.sub!(/\/CN=/i, '') }}
17:48:12 rails.1   |        user     system      total        real
17:48:12 rails.1   |    0.000017   0.000005   0.000022 (  0.000016)
17:48:12 rails.1   | => [#<Benchmark::Tms:0x000000000e63b9c0
17:48:12 rails.1   |   @cstime=0.0,
17:48:12 rails.1   |   @cutime=0.0,
17:48:12 rails.1   |   @label="",
17:48:12 rails.1   |   @real=1.6019039321690798e-05,
17:48:12 rails.1   |   @stime=4.999999999977245e-06,
17:48:12 rails.1   |   @total=2.2000000000022002e-05,
17:48:12 rails.1   |   @utime=1.7000000000044757e-05>]
```

Output showing the older version of Candlepin with this PR works with clients:

```bash
[root@centos7-katello-old-candlepin ~]# subscription-manager register
Registering to: centos7-katello-old-candlepin.area52.example.com:443/rhsm
Username: admin
Password: 
The system has been registered with ID: 3f139c78-281a-4252-a6eb-6dcebfb4c95b
The registered system name is: centos7-katello-old-candlepin.area52.example.com
```

I added in the disabled cops since I don't think or know how to do what it wants and return the value without making it unhappy. 

1. ctrl + c, apply PR and start foreman web server
2. Try to register a client and verify it does not return a 401
3. Attach a subscription of if you are running SCA try to attach a custom product and see if you can see the changes on the client with the new custom repo and make sure subscription manager refresh works
4. Follow the steps above to upgrade Candlepin and repeat the process to make sure the client communication works correctly.

### DIFF
--- a/app/services/cert/rhsm_client.rb
+++ b/app/services/cert/rhsm_client.rb
@@ -10,7 +10,7 @@ module Cert
     end
 
     def uuid
-      drop_cn_prefix_from_subject(@cert.subject.to_s)
+      @uuid ||= @cert.subject.to_a.detect { |name, _, _| name == 'CN' }&.second
     end
 
     private
@@ -24,10 +24,6 @@ module Cert
 
         OpenSSL::X509::Certificate.new(cert)
       end
-    end
-
-    def drop_cn_prefix_from_subject(subject_string)
-      subject_string.sub(/\/CN=/i, '')
     end
 
     def strip_cert(cert)
--- a/test/services/cert/rhsm_client_test.rb
+++ b/test/services/cert/rhsm_client_test.rb
@@ -4,7 +4,9 @@ require File.expand_path("#{Katello::Engine.root}/app/services/cert/rhsm_client.
 
 module Katello
   class RhsmClientTest < Minitest::Test
-    CERT = '
+    # Subject of old cert is
+    # Subject: CN = 14e98155-731a-4cae-b151-5c504cc30e1a
+    OLD_CERT = '
       -----BEGIN CERTIFICATE-----
       MIIEaTCCA1GgAwIBAgIIMAikOB+/HpowDQYJKoZIhvcNAQEFBQAwezELMAkGA1UE
       BhMCVVMxFzAVBgNVBAgTDk5vcnRoIENhcm9saW5hMRAwDgYDVQQHEwdSYWxlaWdo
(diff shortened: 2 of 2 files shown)


app/services/cert/rhsm_client.rb
```diff
@@ -10,7 +10,9 @@ def initialize(cert)
     end
 
     def uuid
-      drop_cn_prefix_from_subject(@cert.subject.to_s)
+      # rubocop:disable Performance/Detect
```
jturel, line 13: Could this be written in a way to avoid disabling the cop?

Also, if I call .uuid multiple times this could be quite slow. Maybe you can use some memoization to put the value into an instance variable

test/services/cert/rhsm_client_test.rb
```diff
@@ -5,36 +5,48 @@
 module Katello
   class RhsmClientTest < Minitest::Test
     CERT = '
-      -----BEGIN CERTIFICATE-----
```
jturel, line 8: It's critical that your change works in a backward compatible way. Best way to ensure that is to have two certs with the different subject formats and verify each of them with your new technique!

test/services/cert/rhsm_client_test.rb
```diff
@@ -5,36 +5,48 @@
...
+    5d/rtZUAUQEgS1hdrPlzeNsG5H8Mf05lJMghH8GFJfjkqm6OHkGYcJGCns1Mpai0
+    4RS8QEA83rtEmSdY/oubjNGUMeXk0mjkOrSIblSUYixJJYdi+66IjePyCGMRhtd4
+    0TV5WB3ONylUv777yuHRICFZyny63C1vtUIZMGtu67qYjKk4uMKnHC5lWMaSmepG
+    ifSe+u12fpxjD0UirYSMj38vDtCbcFSPcAAu/1YKw4uvN5zcv+YI91tjOkMJuVah
+    -----END CERTIFICATE-----'.freeze
 
     def test_uuid
       rhsm_cert = ::Cert::RhsmClient.new(CERT)
-      assert_equal rhsm_cert.uuid, '14e98155-731a-4cae-b151-5c504cc30e1a'
+      refute_equal rhsm_cert.uuid, rhsm_cert.cert.subject.to_s
```
jturel, line 48: If you've already asserted what it should be equal to I wonder what value another assertion that states it should _not_ equal something else adds

jturel (changes requested):

test/services/cert/rhsm_client_test.rb
chris1984: Fair point will remove the other

jeremylenz: I think it can add a lot of value to a test to assert both that something _is_ one way and _is not_ another way! It of course doesn't make any difference to the test's functionality, but I feel like tests' secondary purpose is being read by humans, and being explicit in this way can be a great way to project your intentions.

jeremylenz: I suppose it could be a comment instead of an actual assertion, if that's more palatable :yum:

chris1984: That is what I was going for, to make sure we don't see the org in there, but I also see the other side where we are just grabbing the CN

jturel (approved):
works well. very nice change!

evgeni:
did that just break the nightlies?
https://ci.theforeman.org/job/katello-nightly-rpm-pipeline/1214//artifact/artifacts/foreman-pipeline-katello-nightly-centos8-stream-install/consoleText.txt
https://ci.theforeman.org/job/katello-nightly-rpm-pipeline/1214//artifact/artifacts/foreman-pipeline-katello-nightly-centos7-install/consoleText.txt

sub-man started throwing "HTTP error code 403: Access denied"

jturel:
> did that just break the nightlies? https://ci.theforeman.org/job/katello-nightly-rpm-pipeline/1214//artifact/artifacts/foreman-pipeline-katello-nightly-centos8-stream-install/consoleText.txt https://ci.theforeman.org/job/katello-nightly-rpm-pipeline/1214//artifact/artifacts/foreman-pipeline-katello-nightly-centos7-install/consoleText.txt
> 
> sub-man started throwing "HTTP error code 403: Access denied"

are you sure this change was present in that pipeline? this PR should be fixing 403 errors not adding them :)

evgeni:
no, after looking, the pipe was running with f3daacdb4f2cb5d13e11401db708b41a0ed1644d

but candelpin changed from 4.1.8 to 4.1.10

is that change *needed* for the new CP build?

jturel:
> no, after looking, the pipe was running with [f3daacd](https://github.com/Katello/katello/commit/f3daacdb4f2cb5d13e11401db708b41a0ed1644d)
> 
> but candelpin changed from 4.1.8 to 4.1.10
> 
> is that change _needed_ for the new CP build?

Yes, and so I tagged the candlepin build after this was merged. Am I missing something here?

evgeni:
no, then we wait (until this change lands in an rpm, and we run the pipe again)

app/services/cert/rhsm_client.rb
```diff
@@ -10,7 +10,7 @@ def initialize(cert)
     end
 
     def uuid
-      drop_cn_prefix_from_subject(@cert.subject.to_s)
+      @uuid ||= @cert.subject.to_a.detect { |name, _, _| name == 'CN' }&.second
```
ekohl, line 13: It would be nice if we could align this with [Foreman's CertificateExtract class](https://github.com/theforeman/foreman/blob/develop/app/services/certificate_extract.rb). It looks like we could actually just have a single class between both projects.

evgeni:
just to close the circle here, with the new CP and this patch in, install pipes are green again, thanks!

PR #9903, opened 2022-01-25
Fixes #34202 - Incorrect rpm is uploaded

rverdile:
Fixes bug where uploading an RPM file results in a "successful" upload, but either the wrong file or no file was uploaded.

Also adds to the rpm test_upload test to check that 2 RPMs can be uploaded successfully.
The checksum attribute to filter through the pulp database for duplicate content was not being passed correctly, so the first rpm content added to the database was being considered a duplicate for everything. 

Note that any type that cannot use "relative_path" to filter would be most noticeably affected by this, not just RPMs. This isn't as noticable for the "file" type because "relative_path" is still able to filter content correctly. 
1. Don't checkout PR yet
2. Bulk upload a few RPMs to an empty repository; you should see that the result is only 1 rpm is uploaded (and possibly not even one that you tried uploading).
3. Checkout PR
4. Bulk upload again to the repository; now the upload should work successfully.
 5. Test with python too

### DIFF
--- a/app/services/katello/pulp3/pulp_content_unit.rb
+++ b/app/services/katello/pulp3/pulp_content_unit.rb
@@ -116,30 +116,28 @@ module Katello
         content_unit_list page_opts
       end
 
-      # rubocop:disable Lint/UselessAssignment
       def self.find_duplicate_unit(repository, unit_type_id, file, checksum)
-        filter_label = 'sha256'
+        filter_label = :sha256
         if unit_type_id == 'ostree_ref'
-          filter_label = 'checksum'
+          filter_label = :checksum
         end
         content_backend_service = SmartProxy.pulp_primary.content_service(unit_type_id)
         duplicates_allowed = ::Katello::RepositoryTypeManager.find_content_type(unit_type_id).try(:duplicates_allowed)
         if repository.generic? && duplicates_allowed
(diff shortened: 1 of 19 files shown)


app/services/katello/pulp3/pulp_content_unit.rb
```diff
@@ -116,30 +116,28 @@ def self.fetch_content_list(page_opts)
...
-          filter_label = 'checksum'
+          filter_label = :checksum
         end
         content_backend_service = SmartProxy.pulp_primary.content_service(unit_type_id)
         duplicates_allowed = ::Katello::RepositoryTypeManager.find_content_type(unit_type_id).try(:duplicates_allowed)
         if repository.generic? && duplicates_allowed
           filename_key = ::Katello::RepositoryTypeManager.find_content_type(unit_type_id).filename_key
           duplicate_sha_path_content_list = content_backend_service.content_api(repository.repository_type, unit_type_id).list(
-            filter_label: checksum,
+            filter_label => checksum,
```
jlsherrill, line 129: i really do not like the ":" style hashes for this reason!  Still not sure why ruby added it

jlsherrill (approved):
ACK tested:

* rpm uploads via ui (with existing rpms, present) 
* rpm uploads via hammer (with existing rpms, present)
* python uploads via ui
* python uploads via hammer

PR #9904, opened 2022-01-25
Refs #33951 - correct AIRGAPPED constant reference

jturel:

Ran into an error while migrating my local DB:

```
NameError: uninitialized constant Katello::CdnConfiguration::AIRGAPPED
/home/vagrant/katello/db/migrate/20220124191056_add_type_to_cdn_configuration.rb:10:in `block in change'
```

This PR references the correct constant

set Setting[:subscription_connection_enabled] to false and migrate the DB. You may need to roll back the migrations if you've already applied AddTypeToConfiguration


parthaa (approved):
ACK

PR #9905, opened 2022-01-25
Fixes #34314 - Rake task to set import_only flag

parthaa:
A rake task that allows the admin user to set/unset the import_only flag of a
content view.

Whether it should be an api change or a rake task. The rake task was chosen to make it safer for the user since we don't want users changing between import_only and vice-versa.

- Create a content view without the  `import_only` flag
- Run the following command
```bash
# bundle exec rake katello:set_content_view_import_only ID=2
Content View 'great's import_only value updated to true

# bundle exec rake katello:set_content_view_import_only ID=2 VALUE=false
Content View 'great's import_only value updated to false
```

### DIFF
--- a/app/services/katello/pulp3/content_view_version/import.rb
+++ b/app/services/katello/pulp3/content_view_version/import.rb
@@ -121,11 +121,20 @@ module Katello
                         }
           end
 
+          fail _("Content View label not provided.") if metadata[:label].blank?
+
           cv = ::Katello::ContentView.find_by(label: metadata[:label],
-                                              organization: organization,
-                                              import_only: true)
+                                              organization: organization)
           if cv.blank?
             ::Katello::ContentView.create!(metadata.merge(organization: organization, import_only: true))
+          elsif !cv.import_only?
+            msg = _("Unable to import in to Content View specified in the metadata - '%{name}'. "\
+                     "The 'import_only' attribute for the content view is set to false. "\
+                     "To mark this Content View as importable, have your system administrator"\
+                     " run the following command on the satellite. "\
+                        % { name: cv.name })
+            command = "foreman-rake katello:set_content_view_import_only ID=#{cv.id}"
+            fail msg + "\n" + command
           else
             cv.update!(description: cv_metadata[:description]) if cv.description != metadata[:description]
             cv
--- a/app/services/katello/pulp3/content_view_version/import.rb
+++ b/app/services/katello/pulp3/content_view_version/import.rb
@@ -131,7 +131,7 @@ module Katello
             msg = _("Unable to import in to Content View specified in the metadata - '%{name}'. "\
                      "The 'import_only' attribute for the content view is set to false. "\
                      "To mark this Content View as importable, have your system administrator"\
-                     " run the following command on the satellite. "\
+                     " run the following command on the server. "\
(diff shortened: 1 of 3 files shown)


jturel:
@parthaa this works well!

What do you think about:

1. Throwing a clearer error rather than a validation error in the scenario this solves for the user. It could mention the rake task to set import_only basically guiding the user how to resolve the situation rather than leaving them without any recourse
2. Updating the import/export documentation to mention the existence of this rake task so that users may be able to preempt running into the problem this solves

parthaa:
> 1. Throwing a clearer error rather than a validation error in the scenario this solves for the user. It could mention the rake task to set import_only basically guiding the user how to resolve the situation rather than leaving them without any recourse

Well it should never happen. I have added individual checks for composite and default.
 I cant think of a scenario where this fails. If it does at the least we'll have a validation error that we can help with.

parthaa:
> 2\. Updating the import/export documentation to mention the existence of this rake task so that users may be able to preempt running into the problem this solves

Agreed

jturel:
> > 1. Throwing a clearer error rather than a validation error in the scenario this solves for the user. It could mention the rake task to set import_only basically guiding the user how to resolve the situation rather than leaving them without any recourse
> 
> Well it should never happen. I have added individual checks for composite and default. I cant think of a scenario where this fails. If it does at the least we'll have a validation error that we can help with.

I mean in the dynflow task which is described in the related BZ as raising validation errors on Label and Name - not in your rake task

parthaa:
> I mean in the dynflow task which is described in the related BZ as raising validation errors on Label and Name - not in your rake task

Ah. Good call. Will update the error message.

parthaa:
Updated
```
$ hammer content-import version --organization=import-22010 --path=/var/lib/pulp/exports/export-31523/view/1.0/2022-01-26T21-31-51-00-00 --metadata-file=metadata-1.json
Could not import the archive.:
  Unable to import in to Content View specified in the metadata - 'view'. The 'import_only' attribute for the content view is set to false. 
  To mark this Content View as importable, have your system administrator run the following command on the satellite. 
  foreman-rake katello:set_content_view_import_only ID=9
```

app/services/katello/pulp3/content_view_version/import.rb
```diff
@@ -121,11 +121,20 @@ def self.find_or_create_import_view(organization:, metadata:, library: false)
...
-                                              organization: organization,
-                                              import_only: true)
+                                              organization: organization)
           if cv.blank?
             ::Katello::ContentView.create!(metadata.merge(organization: organization, import_only: true))
+          elsif !cv.import_only?
+            msg = _("Unable to import in to Content View specified in the metadata - '%{name}'. "\
+                     "The 'import_only' attribute for the content view is set to false. "\
+                     "To mark this Content View as importable, have your system administrator"\
+                     " run the following command on the satellite. "\
```
jturel, line 134: Nice! Recommend removing the 'on the satellite' part so it works better up and down-stream

parthaa: Does this work ??


```suggestion
                     " run the following command on the main server. "\
```

jturel (approved):
Works well! ACK

PR #9910, opened 2022-01-28
Fixes #34345 - Download rate setting for remotes

rverdile:
Adds global setting "Download Rate Limit" which is used to set maximum rate limit when syncing.
The setting is global to affect all repositories.

"Use 0 for no limit" may be an unnecessary clarification. Not sure if that's a known interaction with settings.
1. Go to Administer > Settings > Content. Set a "Download rate limit".
2. Create or update a repository with a remote so that the remote refreshes.
3. Use `pulp show --href <remote href>` to see that the remote gets the download rate.
4. You might also try setting a low download rate and syncing something large to see it in action.


jturel:
@rverdile sadly, this needs a rebase. shouldn't affect my testing though since it looks isolated to the vcr cassettes!

PR #9911, opened 2022-01-28
Fixes #34382 - Add Module stream basic table

chris1984:
With the new host details, we are converting the Module streams to the new layout, this PR serves to add a basic table with actions disabled, pagination and auto complete working as well what this mockup requests:

https://marvelapp.com/prototype/995cc27/screen/78342213

* Sync this repo: https://partha.fedorapeople.org/test-repos/pteradactyl/
* Register a EL8 client to your devel box
* Follow this to enable a stream: https://docs.pagure.org/modularity/
* Verify things look good, pagination etc looks ok and works

### DIFF
--- a/app/models/katello/host_available_module_stream.rb
+++ b/app/models/katello/host_available_module_stream.rb
@@ -37,6 +37,16 @@ module Katello
                                              :stream => available_module_stream.stream}).exists?
     end
 
+    def install_status
+      return 'Not installed' if installed_profiles.blank?
+      case status
+      when 'disabled'
+        'Installed'
+      when 'enabled'
+        upgradable? ? 'Upgradable' : 'Up-to-date'
+      end
+    end
+
     def self.upgradable(host)
       upgradable_module_name_streams = ModuleStream.installable_for_hosts([host]).select(:name, :stream)
 
--- a/webpack/components/extensions/HostDetails/Tabs/ModuleStreamsTab/ModuleStreamsActions.js
+++ b/webpack/components/extensions/HostDetails/Tabs/ModuleStreamsTab/ModuleStreamsActions.js
@@ -1,0 +1,16 @@
+import { API_OPERATIONS, get } from 'foremanReact/redux/API';
+import { foremanApi } from '../../../../../services/api';
+import { getResponseErrorMsgs } from '../../../../../utils/helpers';
+import { MODULE_STREAMS_KEY } from './ModuleStreamsConstants';
+
+const errorToast = error => getResponseErrorMsgs(error.response);
+
+export const getHostModuleStreams = (hostId, params) => get({
+  type: API_OPERATIONS.GET,
+  key: MODULE_STREAMS_KEY,
+  url: foremanApi.getApiUrl(`/hosts/${hostId}/module_streams`),
+  errorToast: error => errorToast(error),
+  params,
+});
+
+export default getHostModuleStreams;
--- a/webpack/components/extensions/HostDetails/Tabs/ModuleStreamsTab/ModuleStreamsTab.js
+++ b/webpack/components/extensions/HostDetails/Tabs/ModuleStreamsTab/ModuleStreamsTab.js
@@ -1,0 +1,241 @@
+import React, { useCallback } from 'react';
+import { useSelector } from 'react-redux';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { Skeleton, Label, Hint, HintBody, Button } from '@patternfly/react-core';
+import PropTypes from 'prop-types';
+import { upperFirst } from 'lodash';
+import { TableText, TableVariant, Thead, Tbody, Tr, Th, Td } from '@patternfly/react-table';
+import {
+  LongArrowAltUpIcon,
+  CheckIcon,
+} from '@patternfly/react-icons';
+import { urlBuilder } from 'foremanReact/common/urlHelpers';
+import { selectModuleStreamStatus, selectModuleStream } from './ModuleStreamsSelectors';
+import { useBulkSelect, useUrlParams } from '../../../../Table/TableHooks';
+import { getHostModuleStreams } from './ModuleStreamsActions';
+import InactiveText from '../../../../../scenes/ContentViews/components/InactiveText';
+import TableWrapper from '../../../../../components/Table/TableWrapper';
+import hostIdNotReady from '../../HostDetailsActions';
+import { selectHostDetails } from '../../HostDetailsSelectors';
+
+export const ModuleStreamsTab = () => {
+  const { id: hostId, name: hostName } = useSelector(selectHostDetails);
+
+  const emptyContentTitle = __('This host does not have any Module streams.');
+  const emptyContentBody = __('Module streams will appear here when available.');
+  const emptySearchTitle = __('Your search returned no matching Module streams.');
+  const emptySearchBody = __('Try changing your search criteria.');
+  const { searchParam } = useUrlParams();
+  const columnHeaders = [
+    __('Name'),
+    __('State'),
+    __('Stream'),
+    __('Installation status'),
+    __('Installed profile'),
+  ];
+
+  const fetchItems = useCallback(
+    (params) => {
+      if (!hostId) return hostIdNotReady;
+      return getHostModuleStreams(
+        hostId,
+        params,
+      );
+    },
+    [hostId],
+  );
+
+  const response = useSelector(selectModuleStream);
+  const { results, ...metadata } = response;
+  const status = useSelector(state => selectModuleStreamStatus(state));
+  /* eslint-disable no-unused-vars */
+  const {
+    selectOne, isSelected, searchQuery, selectedCount, isSelectable,
+    updateSearchQuery, selectNone, fetchBulkParams, ...selectAll
+  } = useBulkSelect({
+    results,
+    metadata,
+    isSelectable: _result => false,
+    initialSearchQuery: searchParam || '',
+  });
+  /* eslint-enable no-unused-vars */
+
+  if (!hostId) return <Skeleton />;
+
+  const rowActions = [
+    {
+      title: __('Install'), disabled: true,
+    },
+    {
+      title: __('Update'), disabled: true,
+    },
+    {
+      title: __('Enable'), disabled: true,
+    },
+    {
+      title: __('Disable'), disabled: true,
+    },
+    {
+      title: __('Reset'), disabled: true,
+    },
+    {
+      title: __('Remove'), disabled: true,
+    },
+  ];
+
+  const EnabledIcon = ({ streamText, streamInstallStatus, upgradable }) => {
+    const INSTALLED_STATE = {
+      INSTALLED: __('Installed'),
+      UPGRADEABLE: __('Upgradable'),
+      UPTODATE: __('Up-to-date'),
+      NOTINSTALLED: __('Not installed'),
+    };
+
+    switch (true) {
+      case (streamInstallStatus?.length > 0 && streamText === 'disabled'):
+        return <TableText wrapModifier="nowrap">{INSTALLED_STATE.INSTALLED}</TableText>;
+      case (streamInstallStatus?.length > 0 && streamText === 'enabled' && upgradable !== true):
+        return <><CheckIcon color="green" /> {INSTALLED_STATE.UPTODATE}</>;
+      case (streamInstallStatus?.length > 0 && streamText === 'enabled' && upgradable):
+        return <><LongArrowAltUpIcon color="blue" /> {INSTALLED_STATE.UPGRADEABLE}</>;
+      default:
+        return <InactiveText text={INSTALLED_STATE.NOTINSTALLED} />;
+    }
+  };
+
+  EnabledIcon.propTypes = {
+    streamText: PropTypes.string.isRequired,
+    streamInstallStatus: PropTypes.arrayOf(PropTypes.string).isRequired,
+    upgradable: PropTypes.bool.isRequired,
+  };
+
+  const StreamState = ({ moduleStreamStatus }) => {
+    let streamText = moduleStreamStatus?.charAt(0)?.toUpperCase() + moduleStreamStatus?.slice(1);
+    streamText = streamText?.replace('Unknown', 'Default');
+    switch (true) {
+      case (streamText === 'Default'):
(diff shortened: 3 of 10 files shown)


webpack/components/extensions/HostDetails/Tabs/ModuleStreamsTab/ModuleStreamsActions.js
```diff
@@ -0,0 +1,13 @@
+import { API_OPERATIONS, get } from 'foremanReact/redux/API';
+import { foremanApi } from '../../../../../services/api';
+import { MODULE_STREAMS_KEY } from './ModuleStreamsConstants';
+
+
+export const getHostModuleStreams = (hostId, params) => get({
+  type: API_OPERATIONS.GET,
+  key: MODULE_STREAMS_KEY,
+  url: foremanApi.getApiUrl(`/hosts/${hostId}/module_streams`),
+  params,
```
Andrewgdewar, line 10: Not sure how you guys are handling errors in the hosts pages, but if not otherwise handled I would suggest adding the errorToast to this in case of an issue.

webpack/components/extensions/HostDetails/Tabs/ModuleStreamsTab/ModuleStreamsTab.js
```diff
@@ -0,0 +1,200 @@
...
+import { getHostModuleStreams } from './ModuleStreamsActions';
+import InactiveText from '../../../../../scenes/ContentViews/components/InactiveText';
+import TableWrapper from '../../../../../components/Table/TableWrapper';
+
+export const ModuleStreamsTab = () => {
+  const hostDetails = useSelector(state => selectAPIResponse(state, 'HOST_DETAILS'));
+  const {
+    id: hostId,
+    name: hostName,
+  } = hostDetails;
```
Andrewgdewar, line 24: you don't use hostDetails anywhere else, so you can do 
`
 const {
    id: hostId,
    name: hostName,
  }  = useSelector(state => selectAPIResponse(state, 'HOST_DETAILS'));
`

webpack/components/extensions/HostDetails/Tabs/ModuleStreamsTab/ModuleStreamsTab.js
```diff
@@ -0,0 +1,200 @@
...
+      if (!hostId) return null;
+      return getHostModuleStreams(
+        hostId,
+        params,
+      );
+    },
+    [hostId],
+  );
+
+  const response = useSelector(state => selectAPIResponse(state, MODULE_STREAMS_KEY));
```
Andrewgdewar, line 50: ditto above

webpack/components/extensions/HostDetails/Tabs/ModuleStreamsTab/ModuleStreamsTab.js
```diff
@@ -0,0 +1,200 @@
...
+  const { results, ...metadata } = response;
+  const status = useSelector(state => selectModuleStreamStatus(state));
+  const {
+    selectOne, isSelected, searchQuery, selectedCount, isSelectable,
+    updateSearchQuery, selectNone, fetchBulkParams, ...selectAll
+  } = useBulkSelect({
```
Andrewgdewar, line 56: like this... good.

webpack/components/extensions/HostDetails/Tabs/ModuleStreamsTab/ModuleStreamsTab.js
```diff
@@ -0,0 +1,200 @@
...
+              if (streamStatus === 'Disabled' || streamStatus === 'Default') {
+                disabledText = true;
+              }
+
+            return (
+              <Tr key={`${id} ${index}`}>
+                <Td><a href={`/module_streams?search=module_spec%3D${moduleSpec}+and+host%3D${hostName}`}>{name}</a></Td>
+                <Td><Label color={color} variant={labelVariant}>{streamStatus}</Label></Td>
+                <Td>{stream}</Td>
+                <Td><EnabledIcon stream={streamStatus} streamStatus={installStatus} /></Td>
```
Andrewgdewar, line 181: this variable assignment looks a bit confusing because of the names.. maybe correlate the naming a bit better?

webpack/components/extensions/HostDetails/Tabs/ModuleStreamsTab/ModuleStreamsTab.js
```diff
@@ -0,0 +1,200 @@
...
+              if (installedProfiles.length === 0) {
+                installedProfile = 'No profile installed';
+              }
+              if (installedProfiles.length > 0) {
+                installedProfile = installedProfiles.map(x => x[0].toUpperCase() + x.substr(1));
+                installedProfile = installedProfile.join(', ');
+              }
+              if (streamStatus === 'Disabled' || streamStatus === 'Default') {
+                disabledText = true;
+              }
```
Andrewgdewar, line 174: we should clean this up a bit, either with self-evoking functions/switches or through normal functions defined outside of the render.

webpack/components/extensions/HostDetails/Tabs/ModuleStreamsTab/ModuleStreamsTab.js
```diff
@@ -0,0 +1,200 @@
...
+                installed_profiles: installedProfiles,
+                install_status: installStatus,
+                module_spec: moduleSpec,
+              } = moduleStream;
+
+              let color;
+              let labelVariant;
+              let disabledText = false;
+              /* eslint-disable max-len */
+              let streamStatus = moduleStreamStatus.charAt(0).toUpperCase() + moduleStreamStatus.slice(1);
```
Andrewgdewar, line 152: I would also safe these types of values in case of "" or undefined. 
moduleStreamStatus?.charAt(0)?.etc...

webpack/components/extensions/HostDetails/Tabs/ModuleStreamsTab/ModuleStreamsTab.js
Andrewgdewar: also, now that I look at this, why are you not using the predefined selector in "HostDetailsSelectors.js"? 
export const selectHostDetails = state =>
  selectAPIResponse(state, HOST_DETAILS_KEY) || {};

webpack/components/extensions/HostDetails/Tabs/ModuleStreamsTab/ModuleStreamsTab.js
Andrewgdewar: Screen Shot 2022-01-28 at 10 32 39 AM
Yeah I would suggest adding to the pre-existing selectors, or creating your own as that is the current pattern.

webpack/components/extensions/HostDetails/Tabs/ModuleStreamsTab/ModuleStreamsTab.js
Andrewgdewar: As a general rule, I like to pull out complicated logic from within the render to make the render look as clean as possible. When this is done across the app it makes it much easier to navigate later.

webpack/components/extensions/HostDetails/Tabs/ModuleStreamsTab/ModuleStreamsTab.js
```diff
@@ -0,0 +1,200 @@
...
+    __('Name'),
+    __('State'),
+    __('Stream'),
+    __('Installation status'),
+    __('Installed profile'),
+  ];
+
+  const fetchItems = useCallback(
+    (params) => {
+      if (!hostId) return null;
```
jeremylenz, line 41: This function is a Redux action creator so it should always return a plain object and never `null`. You can return `hostIdNotReady` instead (see other tabs for examples-- you may need to rebase) :)

webpack/components/extensions/HostDetails/Tabs/ModuleStreamsTab/ModuleStreamsTab.js
```diff
@@ -0,0 +1,200 @@
...
+import { selectAPIResponse } from 'foremanReact/redux/API/APISelectors';
+import { MODULE_STREAMS_KEY } from './ModuleStreamsConstants';
+import { selectModuleStreamStatus } from './ModuleStreamsSelectors';
+import { useBulkSelect } from '../../../../Table/TableHooks';
+import { getHostModuleStreams } from './ModuleStreamsActions';
+import InactiveText from '../../../../../scenes/ContentViews/components/InactiveText';
+import TableWrapper from '../../../../../components/Table/TableWrapper';
+
+export const ModuleStreamsTab = () => {
+  const hostDetails = useSelector(state => selectAPIResponse(state, 'HOST_DETAILS'));
```
jeremylenz, line 20: Extra credit: If you can, now is a good time to use the `useUrlParams` hook: https://github.com/Katello/katello/blob/4a363fd945d56ad48f6a30446d9c16320819d4a5/webpack/components/Table/TableHooks.js#L251

This will allow you to accept URL params and transform them into a pre-filled search box.

See the ErrataTab.js for an example of how to use it. (basically you `useUrlParams` to get the searchParam and then pass that as `initialSearchQuery` to `useBulkSelect`)

jeremylenz (commented):
hi! Won't bother you too much here but wanted to bring your attention to two improvements that were recently merged that I think you should incorporate :)

app/models/katello/host_available_module_stream.rb
```diff
@@ -37,6 +37,13 @@ def upgradable?
                                              :stream => available_module_stream.stream}).exists?
     end
 
+    def install_status?
+      return 'Not installed' if installed_profiles.empty?
+      return 'Installed' if installed_profiles && status == 'disabled'
+      return 'Up-to-date' if installed_profiles && status == 'enabled' && upgradable? == false
+      return 'Upgradable' if installed_profiles && status == 'enabled' && upgradable?
+    end
```
jeremylenz, line 45: The `return` keyword isn't very idiomatic in Ruby (except for early returns). I think this `case` statement should be the same:
```suggestion
    def install_status
      return 'Not installed' if installed_profiles.blank?
      case status
      when 'disabled'
        'Installed'
      when 'enabled'
        upgradable? ? 'Upgradable' : 'Up-to-date'
      end
    end
```

also note that I removed the question mark from `install_status` because question marks should only be used if the method returns `true` or `false`.

webpack/components/extensions/HostDetails/Tabs/ModuleStreamsTab/ModuleStreamsActions.js
```diff
@@ -0,0 +1,19 @@
+import { API_OPERATIONS, get } from 'foremanReact/redux/API';
+import { foremanApi } from '../../../../../services/api';
+import { getResponseErrorMsgs } from '../../../../../utils/helpers';
+import { MODULE_STREAMS_KEY } from './ModuleStreamsConstants';
+
+const errorToast = (error) => {
+  const message = getResponseErrorMsgs(error.response);
+  return message;
+};
```
Andrewgdewar, line 9: this could be simplified to a direct return method.

webpack/components/extensions/HostDetails/Tabs/ModuleStreamsTab/ModuleStreamsTab.js
```diff
@@ -0,0 +1,228 @@
...
+
+  EnabledIcon.propTypes = {
+    streamText: PropTypes.string.isRequired,
+    streamInstallStatus: PropTypes.string.isRequired,
+  };
+
+  const StreamState = ({ moduleStreamStatus }) => {
+    let color;
+    let labelVariant;
+    let streamText = moduleStreamStatus?.charAt(0)?.toUpperCase() + moduleStreamStatus?.slice(1);
```
Andrewgdewar, line 111: ?.charAt(0)?.toUpperCase() < we can probably use the "upperfirst" method to clarify/simplify this.

Andrewgdewar: thus can likely be changed to: 
 upperFirst(moduleStreamStatus)

webpack/components/extensions/HostDetails/Tabs/ModuleStreamsTab/ModuleStreamsTab.js
```diff
@@ -0,0 +1,228 @@
...
+          variant={TableVariant.compact}
+        >
+          <Thead>
+            <Tr>
+              {columnHeaders.map(col =>
+                <Th key={col}>{col}</Th>)}
+            </Tr>
+          </Thead>
+          <Tbody>
+            {results?.map((moduleStream, index) => {
```
Andrewgdewar, line 191:
```suggestion
            {results?.map(({
                id,
                status: moduleStreamStatus,
                name,
                stream,
                installed_profiles: installedProfiles,
                install_status: installStatus,
                module_spec: moduleSpec,
              }, index) => {
```

webpack/components/extensions/HostDetails/Tabs/ModuleStreamsTab/ModuleStreamsTab.js
```diff
@@ -0,0 +1,228 @@
...
+              const {
+                id,
+                status: moduleStreamStatus,
+                name,
+                stream,
+                installed_profiles: installedProfiles,
+                install_status: installStatus,
+                module_spec: moduleSpec,
+              } = moduleStream;
```
Andrewgdewar, line 200:
```suggestion
      
```

webpack/components/extensions/HostDetails/Tabs/ModuleStreamsTab/ModuleStreamsTab.js
```diff
@@ -0,0 +1,231 @@
...
+
+  EnabledIcon.propTypes = {
+    streamText: PropTypes.string.isRequired,
+    streamInstallStatus: PropTypes.arrayOf(PropTypes.string).isRequired,
+    upgradable: PropTypes.bool.isRequired,
+  };
+
+  const StreamState = ({ moduleStreamStatus }) => {
+    let color;
+    let labelVariant;
```
Andrewgdewar, line 114: I think we can now just put your values directly within the <Label variables below, as you only have 3 different outcomes.

Andrewgdewar (approved):
Looking good!

PR #9912, opened 2022-01-28
Fixes #34347 - Content view version list API being called multiple times

Andrewgdewar:

- Fixes multiple renders due to improper use of a useEffect hook on the DetailsContainer.

- In the new CV UI, navigate to` /content_views/<id>#/details `
- Open dev tools and click on the network tab
- (optional) under the filter type "content_views/"
- Refresh the page, without the fix 3 - 7 api calls could occur. With it, only 1 should be seen.

### DIFF
--- a/webpack/scenes/ContentViews/Details/DetailsContainer.js
+++ b/webpack/scenes/ContentViews/Details/DetailsContainer.js
@@ -1,27 +1,27 @@
 import React, { useEffect } from 'react';
-import { useDispatch, useSelector, shallowEqual } from 'react-redux';
-import PropTypes from 'prop-types';
 import { STATUS } from 'foremanReact/constants';
-import EmptyStateMessage from '../../../components/Table/EmptyStateMessage';
+import PropTypes from 'prop-types';
+import {
+  shallowEqual,
+  useDispatch,
+  useSelector,
+} from 'react-redux';
 import Loading from '../../../components/Loading';
+import EmptyStateMessage from '../../../components/Table/EmptyStateMessage';
 import getContentViewDetails from './ContentViewDetailActions';
 import {
-  selectCVDetails,
-  selectCVDetailStatus,
   selectCVDetailError,
+  selectCVDetailStatus,
 } from './ContentViewDetailSelectors';
 
-const DetailsContainer = ({ children, cvId, isOpen }) => {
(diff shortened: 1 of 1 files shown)


webpack/scenes/ContentViews/Details/DetailsContainer.js
```diff
@@ -1,27 +1,27 @@
...
   const status = useSelector(state => selectCVDetailStatus(state, cvId), shallowEqual);
   const error = useSelector(state => selectCVDetailError(state, cvId), shallowEqual);
 
   useEffect(() => {
-    if (isOpen && Object.keys(details).length === 0) {
-      dispatch(getContentViewDetails(cvId));
-    }
-  });
+    dispatch(getContentViewDetails(cvId));
+  }, [cvId, dispatch]);
```
Andrewgdewar, line 24: A new linter makes this look messier than it is, but the selection above is the only change.

webpack/scenes/ContentViews/Details/DetailsContainer.js
```diff
@@ -31,11 +31,6 @@ const DetailsContainer = ({ children, cvId, isOpen }) => {
 DetailsContainer.propTypes = {
   children: PropTypes.element.isRequired,
   cvId: PropTypes.number.isRequired,
-  isOpen: PropTypes.bool,
-};
-
-DetailsContainer.defaultProps = {
-  isOpen: true,
```
Andrewgdewar, line 38: This variable wasn't being used, so was removed.

jturel, line 38: What was this isOpen prop about in the first place? It's not a modal dialog... It might be a pre-Andrew-ism :)

jturel (approved):
Nice fix! APJ

webpack/scenes/ContentViews/Details/DetailsContainer.js
Andrewgdewar: ummm This is a mystery to me, I went back through the git history and this was NEVER really used. 
Perhaps the idea was that this details view eventually be accessible via a modal?  
But in its current implementation  they might as well have just done `const isOpen = true; `at the top of this functional scope.

PR #9914, opened 2022-01-31
Refs #30280 - Fixup yum repository copy_units_test.rb

wbclark:
This appears to have been missed in 540e69 where
the method in smart_proxy_extensions.rb was renamed


ianballou:
The fact that the tests are passing means the line can probably actually be deleted.

wbclark:
> The fact that the tests are passing means the line can probably actually be deleted.

Thanks, it is updated

ianballou (approved):

ianballou:
Thanks @wbclark !

PR #9915, opened 2022-02-01
Fixes #34372 - ignore path on proxy url

jlsherrill:
for repo enablement

Ignore the proxy path when trying to enable a repository

http proxy urls don't really have a path that makes sense

1.  use these steps to setup a proxy: https://projects.theforeman.org/projects/katello/wiki/HttpProxyTesting
2. create an http proxy and point it to your proxy
3. set it as the default http proxy setting  on the  content tab of the administer > settings page
4. import a manifest
5. try to enable a repository (no repos will be listed)

### DIFF
--- a/app/lib/katello/resources/cdn.rb
+++ b/app/lib/katello/resources/cdn.rb
@@ -186,7 +186,7 @@ module Katello
         def net_http_class
           if proxy
             uri = URI(proxy.url) #Net::HTTP::Proxy ignores port as part of the url
-            Net::HTTP::Proxy("#{uri.host}#{uri.path}", uri.port, proxy.username, proxy.password)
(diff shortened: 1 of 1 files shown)


rverdile (approved):
this is working for me.

i can't find similar tests, but should this get a test somewhere?

jlsherrill:
For this kind of thing an integration test is the best solution, as no unit test is going to verify that this works in the end.  Our integration tests fall into 2 categories 1) bats tests and 2) qe automation tests

I think this would fall more into 2) as this is very advanced for a bats test (we on purpose don't test everything little thing).  I'm more than open to disagreements about any of this though!

rverdile:
Sounds good to me. I was just curious.

PR #9919, opened 2022-02-02
Fixes #34384 - publish metadata for imported repos

jlsherrill:
Causes a metadata generate to occur for archive cv version repositories on library import and CV import

* create a repo and sync it
* create an export of library:
   `hammer -u admin -p changeme content-export  complete  library  --destination-server=myfoo --organization-id=1`
* create a 2nd org 'foo'
* Deploy a smart proxy associate it to  foo's library
* copy the export from its location to /var/lib/pulp/imports/
* run the import
 `sudo hammer -u admin -p changeme content-import library  --organization=foo --path /var/lib/pulp/imports/022-02-02T15-52-04-00-00/`
 * sync the smart proxy, it should sync correction.
 * subscribe a system to the created 'import' cv, verify it can see the repo via yum repolist & install a package
 
* repeat the above but for a content view export (and a different org)

### DIFF
--- a/app/lib/actions/katello/content_view/publish.rb
+++ b/app/lib/actions/katello/content_view/publish.rb
@@ -200,6 +200,11 @@ module Actions
                 plan_action(Katello::Repository::IndexContent, id: id, full_index: true)
               end
             end
+            concurrence do
+              version.importable_repositories.each do |repo|
+                plan_action(::Actions::Katello::Repository::MetadataGenerate, repo)
+              end
+            end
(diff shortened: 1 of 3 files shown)


parthaa (approved):
Works well ACK!

parthaa:
Was wondering if you needed any sort of migration for existing users who use this fix,
I am guessing we can have an upgrade  task to run metadata generate for each "import-only" version in  in the importing sat ?

jlsherrill:
added an upgrade rake task

parthaa (approved):
I tested the upgrade rake task and it  works well!

PR #9924, opened 2022-02-04
Fixes #34399 - copy partial package groups

jlsherrill:
as part of cv publish

This is a partial revert of this:  https://github.com/Katello/katello/pull/9206
As part of pulp 3.14, the behavior in pulp changed:  https://pulp.plan.io/issues/9055  so we can now revert back to that behavior that was present in pulp2 as well.

1.  sync a rhel 7 kickstart repo
2. create a cv and add the repo to it
3. add an exclude filter and exclude all versions of java-11-openjdk
4. publish the content view
5. check its package groups, it won't include 'Base' package group
after the patch it should


sjha4 (approved):
Works as described. 

Version without the change without Base package group:
Screenshot from 2022-02-09 14-10-33

Version with the change with Base package group:
Screenshot from 2022-02-09 14-09-59

PR #9926, opened 2022-02-04
Fixes #34398 - exclude source redhat containers by default

ianballou:
- Adds the ability to both exclude and include container tags when syncing.
- The `exclude_tags` are pre-populated with `*-source` to exclude the large source tags by default.
- `docker_tags_whitelist` is deprecated.

`docker_tags_whitelist` should still work in the API for now until we completely remove it.

1) Create a repo with no include or exclude tags via Hammer and the UI.  Note that the exclude tags get pre-populated with `*-source`.
2) Create a repo with include and exclude tags via Hammer and the UI.
3) Update a repo's include and exclude tags the Hammer and the UI.
4) Ensure that syncing still works properly.
5) Check the Pulp remote to see the updated include and exclude tags.

### DIFF
--- a/app/controllers/katello/api/v2/repositories_controller.rb
+++ b/app/controllers/katello/api/v2/repositories_controller.rb
@@ -50,7 +50,9 @@ module Katello
       param :unprotected, :bool, :desc => N_("true if this repository can be published via HTTP")
       param :checksum_type, String, :desc => N_("Checksum of the repository, currently 'sha1' & 'sha256' are supported")
       param :docker_upstream_name, String, :desc => N_("Name of the upstream docker repository")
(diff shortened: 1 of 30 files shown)


rverdile (commented):
code looks good so far, gonna test next

rverdile (commented):
Do you think the UI should have additional help text to explain what "*-source" is? Perhaps someone unfamiliar would delete that without thinking about it.

ianballou:
> Do you think the UI should have additional help text to explain what "*-source" is? Perhaps someone unfamiliar would delete that without thinking about it.

Sure, I can add something in the UI.

rverdile (approved):
Looks good! I did not find any weirdness with creating/update/syncing repos of docker or other types. Leaving the `includes` and `exlcudes` empty caused no issues in the UI and Hammer.

PR #9927, opened 2022-02-07
Fixes #33804 - Change size limit of deb attributes based on pulp deb

Manisha15:
Changes size limit of deb attributes based on pulp deb issue https://pulp.plan.io/issues/9277


Manisha15:
[test katello]

ianballou (approved):

PR #9928, opened 2022-02-07
Fixes #34407 - Repo filtering shows all repos in different organizations

sjha4:
The API call to get repos available for CV should be scoped by org.
1. Create 2 orgs (Default could be one)
2. Create and sync a product and repo within each of those orgs.
3. Create a cv in one org
4. Drill to repositories tab
5. filter status to "not added"
6. Repos from all orgs are listed.


ianballou (approved):
Works fine for me!

PR #9929, opened 2022-02-07
Fixes #34408 - Fix package page initial repo filtering

sjha4:
Adds repo id to params of initial table load on packages page.

There's a watcher that handles repository filtering on the Packages page. That works fine when the filter is actually changed using the repo dropdown. However on initial page load with repository selected, the behavior is not consistent.

1. Sync couple of yum repos.
2. Add them to a CV and publish a version
3. Go to version details > Repository
4. Go to _x_ Rpm packages link in the Content column for the repository.
5. The package page should open with repository pre-selected and results filtered by repository.
6. Refresh the page a couple of times and you should keep seeing repo pre-selected and filtered results.

Without this change, the package page with repositoryId specified results in all packages being listed. Refreshing that page will also work erratically.


ianballou (approved):
Works well for me! I wasn't able to reproduce the issue with this patch.

PR #9931, opened 2022-02-08
Fixes #34414 - Errata mail should only list new errata

sjha4:
Errata mailer should only list newly added errata if any for sync summary emails.
content_changed value should be set correctly.
1. a) Set a valid email id for the admin user in your box and in Email Preferences > Subscribe to Sync errata emails.
    b) Go to Administer>settings>Email and click on Test email to verify you're getting emails from the dev box.
2. Sync a new repo with errata.
3. Check sync summary email.
4. Resync. You shouldn't get an email.

Without this change: You'd see an email on every sync regardless of contents_changed or errata being added.

### DIFF
--- a/app/lib/actions/katello/repository/errata_mail.rb
+++ b/app/lib/actions/katello/repository/errata_mail.rb
@@ -4,17 +4,16 @@ module Actions
       class ErrataMail < Actions::EntryAction
         middleware.use Actions::Middleware::ExecuteIfContentsChanged
 
-        def plan(repo, last_updated = nil, contents_changed = nil)
-          last_updated ||= repo.repository_errata.order('updated_at ASC').last.try(:updated_at) || Time.now
-          plan_self(:repo => repo.id, :last_updated => last_updated.to_s, :contents_changed => contents_changed)
+        def plan(repo, contents_changed = nil)
+          plan_self(:repo => repo.id, :contents_changed => contents_changed)
         end
 
         def run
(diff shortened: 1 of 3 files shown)


rverdile (commented):
This is working, but should there be a test? I can't find a similar test currently. What do you think?

rverdile (approved):
looks good!

app/lib/actions/katello/repository/errata_mail.rb
```diff
@@ -4,17 +4,16 @@ module Repository
...
-          plan_self(:repo => repo.id, :last_updated => last_updated.to_s, :contents_changed => contents_changed)
+        def plan(repo, contents_changed = nil)
+          plan_self(:repo => repo.id, :contents_changed => contents_changed)
         end
 
         def run
           ::User.current = ::User.anonymous_admin
-
           repo = ::Katello::Repository.find(input[:repo])
+          last_updated = repo.repository_errata.order('updated_at ASC').last.try(:updated_at) || Time.now
```
hao-yu, line 14: I think there is an issue in this line. It is getting the 'last_updated' in "def run" which means it is getting the "last_updated" from the updated index as this action runs after the 'index content' action. This means, we will never get any errata notification because it won't be any beyond the "last_updated".   It should retrieve the "last_updated" in "def plan" just like the older version of Katello (Satellite 6.9)

PR #9932, opened 2022-02-08
Fixes #34417 - Don't set global RestClient proxy during repo discover

jturel:

RestClient.proxy is a global value and should not be set unless the intention is to change the global http proxy (unlikely). Doing this in RepoDiscovery for docker repos forces even localhost traffic to go through the HTTP proxy meant for external traffic.

Avoid setting global rest client proxy which I found to be doable with RestClient::Request.execute!

You can follow the testing steps in the BZ if you want to go all the way: https://bugzilla.redhat.com/show_bug.cgi?id=2046337
Otherwise, you could also verify this via the rails console by calling repo discovery, and then calling a Candlepin resource.

### DIFF
--- a/app/lib/katello/repo_discovery.rb
+++ b/app/lib/katello/repo_discovery.rb
@@ -1,10 +1,12 @@
 module Katello
   class RepoDiscovery
+    include Katello::Util::HttpProxy
+
     attr_reader :found, :crawled, :to_follow
 
     # rubocop:disable Metrics/ParameterLists
     def initialize(url, content_type = 'yum', upstream_username = nil,
-      upstream_password = nil, search = '*', proxy = {}, crawled = [],
+      upstream_password = nil, search = '*', crawled = [],
       found = [], to_follow = [])
       @uri = uri(url)
       @content_type = content_type
@@ -14,7 +16,6 @@ module Katello
       @found = found
       @crawled = crawled
       @to_follow = to_follow
-      @proxy = proxy
     end
     # rubocop:enable Metrics/ParameterLists
 
@@ -42,14 +43,6 @@ module Katello
     private
 
     def docker_search
-      proxy_uri = nil
-      if @proxy && @proxy[:proxy_host]
-        proxy_uri = URI(@proxy[:proxy_host])
-        proxy_uri.user = @proxy[:proxy_user] if @proxy[:proxy_user]
-        proxy_uri.password = @proxy[:proxy_password] if @proxy[:proxy_password]
-        proxy_uri.port = @proxy[:proxy_port] if @proxy[:proxy_port]
-      end
-
       request_params = {
         method: :get,
         headers: { accept: :json },
@@ -58,7 +51,7 @@ module Katello
 
       request_params[:headers][:user] = @upstream_username unless @upstream_username.empty?
       request_params[:headers][:password] = @upstream_password unless @upstream_password.empty?
-      request_params[:proxy] = proxy_uri.to_s if proxy_uri
+      request_params[:proxy] = proxy_uri if proxy
 
       begin
         results = RestClient::Request.execute(request_params)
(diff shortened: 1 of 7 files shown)


app/lib/katello/repo_discovery.rb
```diff
@@ -42,31 +42,33 @@ def run(resume_point)
...
 
+      request_params = {
+        method: :get,
+        headers: { accept: :json },
+        url: "#{@uri}v1/search?q=#{@search}"
+      }
+
+      request_params[:headers][:user] = @upstream_username unless @upstream_username.empty?
+      request_params[:headers][:password] = @upstream_password unless @upstream_password.empty?
+      request_params[:proxy] = proxy_uri.to_s if proxy_uri
```
jlsherrill, line 61: when using a proxy password with special characters "@password#@"   i get an error on repo discovery:

Discovery failed. Error: bad password component

Its likely because rest client doesn't support this natively, but i think https://github.com/theforeman/foreman/commit/a196afb125a1af710b69bafd68d1803f7b630620 was added to help support this, we just have to escape it.  I believe that the upstream subscription stuff uses a helper method: 

https://github.com/Katello/katello/blob/master/app/lib/katello/resources/candlepin.rb#L99

 defined here:
https://github.com/Katello/katello/blob/master/app/lib/katello/util/http_proxy.rb#L4

jlsherrill: i only checked this because literally everytime we add proxy support to something, this break ;)

jlsherrill: also this happens on master too, but probably good to fix here

app/models/katello/concerns/organization_extensions.rb
```diff
@@ -197,7 +197,7 @@ def parent_id=(_parent_id)
 
         def latest_repo_discovery
           ForemanTasks::Task::DynflowTask.for_action(::Actions::Katello::Repository::Discover)
-            .for_resource(::User.current).order("started_at").last
+            .where(user: ::User.current).order("started_at").last
```
jturel, line 200: @adamruzicka I see that for_resource is looking for Links against the given task type, but Repository::Discover tasks has no Links so nothing was returned here. Was that an intentional change with the locking rework in foreman tasks? Do we need to set an action_subject, or is using `where user` appropriate here?

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/products.routes.js
```diff
@@ -42,13 +42,14 @@ angular.module('Bastion.products').config(['$stateProvider', function ($statePro
...
         url: '/products/discovery',
-        controller: 'DiscoveryController',
-        templateUrl: 'products/discovery/views/discovery-base.html'
+        templateUrl: 'products/discovery/views/discovery-base.html',
+        permission: 'edit_products',
+        redirectTo: 'product-discovery.scan'
     })
     .state("product-discovery.scan", {
         url: '/scan',
+        controller: 'DiscoveryController',
```
jturel, line 52: This seemed to do the trick w.r.t the browser console error. I can still discover and enable repos by hitting the page directly or navigating from /products list

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/products.routes.js
```diff
@@ -42,13 +42,14 @@ angular.module('Bastion.products').config(['$stateProvider', function ($statePro
     });
 
     $stateProvider.state("product-discovery", {
-        abstract: true,
         url: '/products/discovery',
-        controller: 'DiscoveryController',
-        templateUrl: 'products/discovery/views/discovery-base.html'
+        templateUrl: 'products/discovery/views/discovery-base.html',
```
jturel, line 46: Small adjustment here now makes /products/discovery redirect to /products/discovery/scan which avoids the UI trying to display an nonexistent product called 'discovery'

app/models/katello/concerns/organization_extensions.rb
jturel: Thank you Adam!

jlsherrill (approved):
ACK!

PR #9935, opened 2022-02-09
Fixes #34425 - Update errata severity colors

jeremylenz:

Per request from maria, update errata severity colors to be consistent with console
Screenshot 2022-02-09 at 12 44 03

I used the specific color values because I wasn't sure if the CSS variables would work in the patternfly components. It's fine!

View errata on the new host details page and confirm colors match the mockup.
Tip: In React dev tools, find one of the `<ErrataSeverity>` components and change the `severity` prop to "Low", "Critical", etc. to momentarily see that icon.

### DIFF
--- a/webpack/components/Errata/index.js
+++ b/webpack/components/Errata/index.js
@@ -1,5 +1,11 @@
 import React from 'react';
 import { TableText } from '@patternfly/react-table';
+import {
+  chart_color_black_500 as pfBlack,
+  chart_color_gold_400 as pfGold,
+  chart_color_orange_300 as pfOrange,
+  chart_color_red_200 as pfRed,
+} from '@patternfly/react-tokens';
 import { translate as __ } from 'foremanReact/common/I18n';
 import {
   BugIcon,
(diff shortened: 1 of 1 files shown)


Andrewgdewar:
Are these colors accessible via the patternfly/react-tokens package? 
Example: 
`import {
  global_warning_color_100 as warningColor,
} from '@patternfly/react-tokens';
`
If so the colors should likely originate from there, so that if patternfly changes any of them in a later release they will be updated.

jeremylenz:
Thanks @Andrewgdewar 

I fought with patternfly a bit and wrangled the colors out of them :)

Only took 2 guesses to find the correct imports. Turns out they will give me an object like
```js
{name: '--pf-chart-color-black-500', value: '#6a6e73', var: 'var(--pf-chart-color-black-500)'}
```

of which I can then use the `value`. Updated.

Andrewgdewar (approved):
Looking good

lfu (approved):
Worked well!

PR #9937, opened 2022-02-10
Fixes #34434 - Modulemd-defaults metadata not copied to filtered CVs

rverdile:
This adds an API method for modulemd-defaults in the yum API class so modulemd-defaults can be explicitly added to content view repositories with filters. Previously, modulemd-defaults were missing from the modules.yaml metadata.  
The implementation here closely follows the package environments code path, which copies package environments to filtered repositories.

To reproduce:
1. Sync a yum repository that has module streams, like this one: https://fixtures.pulpproject.org/rpm-with-modules/.
2. Create a content view and add this repository with a filter to include all module streams.
3. Publish content view.
4. In the "...modules.yml" file for your published content view at the `https://hostname/pulp/content/path/to/your/repo/` URL, "modulemd-defaults" are not present in the file.
5. By comparison, if you look at the modules.yml metadata for the original repository, or a content view without filters, you should see "modulemd-defaults" present in the file.

To see fix:
1. Apply PR
2. Following the reproduction steps, in the modules.yml file you should now see "modulesmd-defaults" present.
3. Be sure to test with and without dependency solving on.

### DIFF
--- a/app/services/katello/pulp3/api/yum.rb
+++ b/app/services/katello/pulp3/api/yum.rb
@@ -36,6 +36,10 @@ module Katello
           PulpRpmClient::ContentPackageenvironmentsApi.new(api_client)
         end
 
+        def content_modulemd_defaults_api
+          PulpRpmClient::ContentModulemdDefaultsApi.new(api_client)
+        end
+
         def content_repo_metadata_files_api
           PulpRpmClient::ContentRepoMetadataFilesApi.new(api_client)
         end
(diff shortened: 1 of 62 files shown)


jeremylenz (approved):
Thanks for the great testing steps!

Tested without PR: `modulemd-defaults` not found in xxx-modules.yaml, as described.
Published new version with depsolving off: `modulemd_defaults` present in xxx-modules.yaml :heavy_check_mark: 
Published new version with depsolving on: `modulemd-defaults` present in xxx-modules.yaml :heavy_check_mark: 

APJ :+1:

PR #9938, opened 2022-02-10
Fixes #34436 - dont delete shared pulp3 repos for composites

jlsherrill:

1. Install 3.18/6.9
2. create & sync a yum repo
3. publish it in a content view with a filter
4. create a composite CV that includes that CV and publish it
5. (repeat 3 & 4 a couple times)
6. upgrade to 4.0/6.10
7. delete one of the composite content views
8. try to publish a new version of the component that was in that composite


ianballou (approved):
Worked fine for me!

PR #9944, opened 2022-02-14
Fixes #34501 - Refactor CV import to use MetadataMap + support cp_id product lookup

jturel:

The labels of Red Hat products can sometimes change. Given that the import process is driven by labels (looking up products in the importing org) this would break when the product labels of the exporting and importing org no longer match. While the solution would be to refresh the manifest in both orgs and re-export, this is undesirable. This change allows Red Hat products to be looked up by their cp_id when importing, if present. For backward compatibility, the product lookup will use the label from the export metadata if the cp_id is missing.

While looking into how to handle cp_id on the importing side I came across this method: 
https://github.com/Katello/katello/blob/master/app/services/katello/pulp3/content_view_version/import.rb#L64-L77

Given an export metadata hash, it produces two differently shaped hashes depending on the parameters given. Looking at this method and things like complexity introduced in areas where `metadata_map` was used, and the vague shape of the data (it's hard to tell what's inside) I created the MetadataMap class. Its role is to parse the export metadata json into a collection of objects that are easy to understand and pass around. It simplified a lot of the dynflow actions, import services code, and tests. More importantly it makes the code easier to follow and change in the future - especially if fields get added or removed from the metadata, or if its general shape should change.

I also cleaned up (what I found to be) encapsulation issues. There were a number of attr_accessors which should be attr_readers (or not present at all) and the ContentViewVersion::Import service class had a lot of methods moved from class methods to instance methods given that the org and metadata seemed to be very common arguments to pass to these, and the instance of Import was already aware of those values.

- Create an org for import and export
- Import a manifest in both orgs
- In the export org, enable a Red Hat repo (satellite tools is a good, small one)
- Export the repository

- Simulate a change of the label thru the rails console:
```
prod = Katello::Product.where(label: 'Red_Hat_Enterprise_Linux_for_x86_64', organization_id: $IMPORT_ORG_ID).first
prod.update_attribute(:label, prod.label + "-updated")
```
- Import the exported repository and notice the failure reported in the BZ: "The organization manifest does not contain the subscriptions required to enable the following repositories"

- Check out this PR and try the import again - it should succeed!

- Due to the extensive refactoring on the import side, I recommend validating the other workflows by importing various kinds of exports. I've already done so, but it would be good to double-check this.


app/services/katello/pulp3/content_view_version/importable_repositories.rb
```diff
@@ -2,63 +2,115 @@ module Katello
...
-            if repositories_in_library.include? [product_label, repo_label]
-              repo = ::Katello::RootRepository.find_by(product: product, label: repo_label)
-              updatable << { repository: repo, options: params.except(:label, :name, :content_type) }
-            elsif redhat
-              product_content = product_content_by_label(content[:label])
+          @metadata_repositories.each do |repo|
+            product = product_for_metadata_repo(repo)
+            fail _("Unable to find product '%s' in organization '%s'" % [repo.product.label, @organization.name]) if product.blank?
+
+            # TODO: is it valid to look up the root this way?
```
jturel, line 25: Open question here

jturel: @parthaa  this code previously worked by consulting an array called `repositories_in_library` which contained values like `[product_label, repo_label]`. I'm not totally sure why that list needed to be consulted for the sake of this method. Currently it's working a bit differently - once it finds the product it then finds the root repository from those loaded via association, and the end result seems the same. Is there some gotcha that I'm not seeing?

jturel:
@ianballou thanks for picking up the review here! since @parthaa wrote all of this code it'd be good for him to at least give his official seal of approval before merge. Let me know if you wanna hop on a meet or something like that to talk thru the change as well.

app/services/katello/pulp3/content_view_version/import_validator.rb
```diff
@@ -26,113 +27,107 @@ def check!
...
-          unless products_not_in_library.blank?
-            # make a list of [product_name, repo_name] pairs to report to the user
-            product_repos = []
-            products_not_in_library.each do |prod|
-              products_in_metadata_map[prod].each do |repo|
-                product_name = metadata[:products][prod][:name]
-                repo_name = repo[:name]
-                product_repos << [product_name, repo_name]
-              end
+        def katello_product_for_metadata_repo(metadata_repo)
```
ianballou, line 104: I wasn't able to fix the original error by patching in this PR. It looks like it's due to my manifest.json not having a `cp_id` for the product.  That causes this method to search by label, which fails.  Why would the manifest.json file be missing the `cp_id`?

ianballou: I did a complete library export:

```
hammer content-export complete library --organization "export org"
```

and then a complete library import:

```
hammer content-import library --organization "import org" --metadata-file metadata-5.json  --path '/var/lib/pulp/exports/export_org/Export-Library/5.0/2022-03-08T21-11-36-00-00'
```

ianballou: My metadata file:

```
[vagrant@centos7-hammer-devel ~]$ cat metadata-5.json | jq
{
  "organization": "export org",
  "repositories": {
    "Red_Hat_Satellite_Tools_6_10_for_RHEL_8_x86_64_RPMs-1746498": {
      "name": "Red Hat Satellite Tools 6.10 for RHEL 8 x86_64 RPMs",
      "label": "Red_Hat_Satellite_Tools_6_10_for_RHEL_8_x86_64_RPMs",
      "description": null,
      "arch": "noarch",
      "content_type": "yum",
      "unprotected": false,
      "checksum_type": null,
      "os_versions": null,
      "major": null,
      "minor": null,
      "download_policy": "immediate",
      "mirroring_policy": "mirror_complete",
      "product": {
        "label": "Red_Hat_Enterprise_Linux_for_x86_64"
      },
      "gpg_key": {},
      "content": {
        "id": "12092",
        "label": "satellite-tools-6.10-for-rhel-8-x86_64-rpms"
      },
      "redhat": true
    }
  },
  "content_view": {
    "name": "Export-Library",
    "label": "Export-Library",
    "description": null,
    "generated_for": "library_export"
  },
  "content_view_version": {
    "major": 5,
    "minor": 0,
    "description": ""
  },
  "incremental": false,
  "products": {
    "Red_Hat_Enterprise_Linux_for_x86_64": {
      "name": "Red Hat Enterprise Linux for x86_64",
      "label": "Red_Hat_Enterprise_Linux_for_x86_64",
      "description": null,
      "gpg_key": {},
      "redhat": true
    }
  },
  "gpg_keys": {},
  "toc": "export-e6fca860-ad70-46d1-bed9-80f434abe6d4-20220308_2111-toc.json"
}
```

ianballou: Huh. I exported again with your PR in place and now I see the `cp_id`:

```
[vagrant@centos7-hammer-devel ~]$ cat metadata-6.json | jq
{
  "organization": "export org",
  "repositories": {
    "Red_Hat_Satellite_Tools_6_10_for_RHEL_8_x86_64_RPMs-1746498": {
      "name": "Red Hat Satellite Tools 6.10 for RHEL 8 x86_64 RPMs",
      "label": "Red_Hat_Satellite_Tools_6_10_for_RHEL_8_x86_64_RPMs",
      "description": null,
      "arch": "noarch",
      "content_type": "yum",
      "unprotected": false,
      "checksum_type": null,
      "os_versions": null,
      "major": null,
      "minor": null,
      "download_policy": "immediate",
      "mirroring_policy": "mirror_complete",
      "product": {
        "label": "Red_Hat_Enterprise_Linux_for_x86_64"
      },
      "gpg_key": {},
      "content": {
        "id": "12092",
        "label": "satellite-tools-6.10-for-rhel-8-x86_64-rpms"
      },
      "redhat": true
    }
  },
  "content_view": {
    "name": "Export-Library",
    "label": "Export-Library",
    "description": null,
    "generated_for": "library_export"
  },
  "content_view_version": {
    "major": 6,
    "minor": 0,
    "description": ""
  },
  "incremental": false,
  "products": {
    "Red_Hat_Enterprise_Linux_for_x86_64": {
      "name": "Red Hat Enterprise Linux for x86_64",
      "label": "Red_Hat_Enterprise_Linux_for_x86_64",
      "description": null,
      "cp_id": "479",
      "gpg_key": {},
      "redhat": true
    }
  },
  "gpg_keys": {},
  "toc": "export-275d3181-d043-4ee0-9f1e-cd6e45d8ce61-20220308_2151-toc.json"
}
```

ianballou: Oh I see, you added the `cp_id`!  Okay, so I guess then in the reproducer steps there should be another export.

jturel: Sorry, I've been looking @ this pr a bit too long. You would need to re-export just as you found to get the cp_id in the metadata.json !

jturel: Your test was actually a good one though, to test that error would still show up in absence of cp_id + mismatched repo labels

app/lib/actions/pulp3/content_view_version/create_import.rb
```diff
@@ -1,22 +1,22 @@
 module Actions
   module Pulp3
     module ContentViewVersion
-      class Import < Pulp3::AbstractAsyncTask
+      class CreateImport < Pulp3::AbstractAsyncTask
```
ianballou, line 4: There are some places where the name change isn't propagated, like in the finalize step for [app/lib/actions/katello/content_view_version/import.rb](https://github.com/Katello/katello/pull/9944/files#diff-95d4432c1a79855e7a9c358ab57dcdb08066f32163216c5ac1c4425b1e3fcadd)

ianballou (commented):
I ran through the initial test steps and it worked! I had to rename a task call though, details below:

app/lib/actions/pulp3/content_view_version/create_import.rb
jturel: Ah missed that in the rebase I did just before pushing. I think that was the only one? Fixed now

ianballou: Looks like that was it!

app/services/katello/pulp3/content_view_version/metadata_map.rb
```diff
@@ -0,0 +1,119 @@
...
+              name: p[:name],
+              label: p[:label],
+              cp_id: p[:cp_id],
+              description: p[:description],
+              gpg_key: gpg_key_for_product(p),
+              redhat: ::Foreman::Cast.to_bool(p[:redhat])
+            )
+          end
+        end
+
```
parthaa, line 70:
```ruby
        def parse_products(products)
          products.values.map do |p|
            MetadataProduct.new(
              p.merge(
                gpg_key: gpg_key_for_product(p),
                redhat: ::Foreman::Cast.to_bool(p[:redhat])
              )
            )
          end
        end
```

app/services/katello/pulp3/content_view_version/metadata_map.rb
```diff
@@ -0,0 +1,119 @@
...
+        def parse_content_view(content_view)
+          MetadataContentView.new(
+            name: content_view['name'],
+            label: content_view['label'],
+            description: content_view['description'],
+            generated_for: content_view['generated_for']
+          )
+        end
+
```
parthaa, line 57:
```ruby
        def parse_content_view(content_view)
          MetadataContentView.new(content_view)
        end
```

app/services/katello/pulp3/content_view_version/metadata_map.rb
```diff
@@ -0,0 +1,119 @@
...
+        def parse_gpg_keys(gpg_keys)
+          gpg_keys.values.map do |g|
+            MetadataGpgKey.new(
+              name: g[:name],
+              content_type: g[:content_type],
+              content: g[:content]
+            )
+          end
+        end
+
```
parthaa, line 48:
```ruby
        def parse_gpg_keys(gpg_keys)
          gpg_keys.values.map do |g|
            MetadataGpgKey.new(g)
          end
        end
```

app/services/katello/pulp3/content_view_version/metadata_map.rb
```diff
@@ -0,0 +1,119 @@
...
+              redhat: repo[:redhat],
+              product: product_for_repo(repo),
+              gpg_key: gpg_key_for_repo(repo),
+              content: repo[:content] ? MetadataRepositoryContent.new(id: repo[:content][:id], label: repo[:content][:label]) : nil
+            )
+          rescue => e
+            raise _("Invalid repository in the metadata %{repo} error=%{error}") % { repo: repo, error: e.message }
+          end
+        end
+
```
parthaa, line 104:
```ruby
        def parse_repositories(repositories)
          repositories.map do |pulp_name, repo|
            MetadataRepository.new(
              repo.merge(
                pulp_name: pulp_name,
                product: product_for_repo(repo),
                gpg_key: gpg_key_for_repo(repo),
                content: repo[:content] ? MetadataRepositoryContent.new(id: repo[:content][:id], label: repo[:content][:label]) : nil
              )
            )
          rescue => e
            raise _("Invalid repository in the metadata %{repo} error=%{error}") % { repo: repo, error: e.message }
          end
        end

```

app/services/katello/pulp3/content_view_version/metadata_map.rb
parthaa: Kind of mixed on this approach. I don't find the need to elaborate every  parameter.

app/services/katello/pulp3/content_view_version/importable_repositories.rb
parthaa: I think what you have here is good enough. The previous implementation looks like tried too hard to minimize the number of db calls.

app/services/katello/pulp3/content_view_version/importable_repositories.rb
```diff
@@ -2,63 +2,115 @@ module Katello
...
+            root = product.root_repositories.find { |r| r.label == repo.label }
```
parthaa, line 26: Do you want to match arch and minor also here in addition to label ?

app/services/katello/pulp3/content_view_version/importable_repositories.rb
```diff
@@ -2,63 +2,115 @@ module Katello
...
-              creatable << { repository: product.add_repo(params) }
+              creatable << { repository: product.add_repo(create_repo_params(repo)) }
             end
           end
         end
+
+        private
+
+        def product_for_metadata_repo(repo)
+          if repo.redhat && repo.product.cp_id
```
parthaa, line 48: what happens if it is a custom repo with the same name as a redhat repo ?

parthaa:
Code wise looks good to me (other than some of my optional suggestions and questions). Seems to be smart enough to handle older exports which dont have cp_id but have labels. If @ianballou has tested this I am happy with Acking :)

app/services/katello/pulp3/content_view_version/importable_repositories.rb
jturel: That makes things simpler! ty

app/services/katello/pulp3/content_view_version/importable_repositories.rb
jturel: I think that implies more change to the ImportValidator too. it might be valid, but this change is huge already so I'd prefer to have it come after this is merged

app/services/katello/pulp3/content_view_version/importable_repositories.rb
jturel: I guess it'll be treated as a custom repo! Hopefully no one is doing that

jturel:
@ianballou @parthaa pushed a cleanup commit and left notes on comments. If all is well, please approve the change :)

app/services/katello/pulp3/content_view_version/metadata_map.rb
parthaa: > It's reasonable to me that just because we add some value in the exported metadata doesn't mean we want it exposed verbatim in the MetadataMap
I don't know if we need to be that picky. We trust what is on the disk is correct in any case. That being said if we make metadatamap also generate the json in a future pr, we can look to fixing that.

jturel:
[test katello]

ianballou (approved):
Works for me!

PR #9945, opened 2022-02-14
Fixes #34158 - Add status value translations

sjha4:
Add status translations.

There are several pages on CV UI that use Status - Added/Not Added/All filters like Repository tab, Content View tab on Composite CVs, Fome types of filters like Module Streams, Errata by Id, etc. Make sure the translation doesn't break the filter.

1. Go to Admin User > Edit
2. Set language to something other than English.
3. Go to CV UI and test out status filters and make sure they work correctly.

### DIFF
--- a/webpack/scenes/ContentViews/ContentViewsConstants.js
+++ b/webpack/scenes/ContentViews/ContentViewsConstants.js
@@ -56,9 +56,9 @@ export const addComponentSuccessMessage = component => (component ? __('Updated 
 
 
 // Repo added to content view status display and key
-export const ADDED = 'Added';
-export const NOT_ADDED = 'Not added';
-export const ALL_STATUSES = 'All';
+export const ADDED = __('Added');
+export const NOT_ADDED = __('Not added');
+export const ALL_STATUSES = __('All');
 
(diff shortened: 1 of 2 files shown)


webpack/scenes/ContentViews/ContentViewsConstants.js
```diff
@@ -56,9 +56,9 @@ export const addComponentSuccessMessage = component => (component ? __('Updated
 
 
 // Repo added to content view status display and key
-export const ADDED = 'Added';
-export const NOT_ADDED = 'Not added';
-export const ALL_STATUSES = 'All';
+export const ADDED = __('Added');
```
Andrewgdewar, line 59: these statuses are used as responses to the API in a few places. 
For example in ContentViewComponents.js the "statusSelected" value is one of these translated values. 
We send this directly to the api within the getContentViewComponents functions "status" param. 
Unless the API expects translated values we need to change all usages of these variables to constants. 
An enum pattern like 
`
const STATUSES = { 
added: __('Added'),
not_added: __('Not Added'),
...
}
`
The keys will obviously have to reflect whatever the api expects for statuses of course.

PR #9947, opened 2022-02-14
Fixes #34448 - Truncate long audit comments for import/export

jeremylenz:

* Adds a new Rails concern, `AuditCommentExtensions`, which powers up your Katello class with a new method `truncate_audit_comment`. This method prevents the error `Error: PG::StringDataRightTruncation: ERROR: value too long for type character varying(255)` by doing the following
    - Check if the audit comment is going to be longer than 255 characters
    - If not, pass the comment thru as-is
    - If so,
      - Write a Rails logger message with the untruncated comment. (This is because the user that performed the import is near the end of the comment, and long content view names could be used to conceal the username that performed the action.)
      - Truncate the comment so that it's < 254 characters with a "..." at the end.

The bug was only filed for imports. But I wouldn't want to see the same type of thing pop up months down the road for exports as well. So I applied the fix to both, and also made it reusable should we ever need to do this somewhere else.

1. create a content view with a long name, for example:

`XXXX_Red_Hat_Software_Collections_RPMs_for_Red_Hat_Enterprise_Linux_7_Server_x86_64_7Server`

2. Add and sync a repo

2. Export and copy the archive to the corresponding place in `/var/lib/pulp/imports`

3. Create an import org and run the import:

```
hammer content-import version --organization-id 1 --path /var/lib/pulp/imports/XXX/XXXX_Red_Hat_Software_Collections_RPMs_for_Red_Hat_Enterprise_Linux_7_Server_x86_64_7Server/1.0/2022-01-29T02-54-30-10-30/
```

Before applying this patch:

The import will fail with:

```
ActiveRecord::ValueTooLong: PG::StringDataRightTruncation: ERROR: value too long for type character varying(255)
```

After applying this patch:

The import should be successful, and you should see logs and audit comments as described above.

### DIFF
--- a/app/models/katello/content_view_version_import_history.rb
+++ b/app/models/katello/content_view_version_import_history.rb
@@ -32,9 +32,8 @@ module Katello
     scoped_search :on => :id, :only_explicit => true, :validator => ScopedSearch::Validators::INTEGER
     scoped_search :on => :import_type, :rename => :type, :complete_value => ContentViewVersionExportHistory::EXPORT_TYPES
 
-    def self.generate_audit_comment(user:, path:, content_view_name:)
-      truncate_audit_comment(_("Content imported from %{path} into content view '%{name}' by %{user}") % {
-        path: path,
+    def self.generate_audit_comment(user:, content_view_name:)
+      truncate_audit_comment(_("Content imported by %{user} into content view '%{name}'") % {
         user: user.to_label,
         name: content_view_name
       })
--- a/app/models/katello/content_view_version_import_history.rb
+++ b/app/models/katello/content_view_version_import_history.rb
@@ -1,6 +1,7 @@
 module Katello
   class ContentViewVersionImportHistory < Katello::Model
     include Authorization::ContentViewVersionImportHistory
+    include Concerns::AuditCommentExtensions
 
     audited except: :metadata
     delegate :organization, to: :content_view_version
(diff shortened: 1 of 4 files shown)


jeremylenz:
truncated_audit_comment

app/models/katello/content_view_version_import_history.rb
```diff
@@ -32,11 +33,11 @@ class ContentViewVersionImportHistory < Katello::Model
     scoped_search :on => :import_type, :rename => :type, :complete_value => ContentViewVersionExportHistory::EXPORT_TYPES
 
     def self.generate_audit_comment(user:, path:, content_view_name:)
-      _("Content imported from %{path} into content view '%{name}' by %{user}") % {
+      truncate_audit_comment(_("Content imported from %{path} into content view '%{name}' by %{user}") % {
```
parthaa, line 36: I like this truncation idea. I also wonder if we can take the path out of this comment since it shows up as a property anyway?
Audits

May be we could put the content view name also as a property.

parthaa: And may be add like "Generated" : `Yes/No ` just to highlight

parthaa: Worked well in my test.

chris1984:
Worked well, ACK from me, if you want to address @parthaa comments

chris1984 (approved):

app/models/katello/content_view_version_import_history.rb
jeremylenz: > May be we could put the content view name also as a property.

In order to add properties to that table I think I'd have to add fields to the model. I did remove the `path` from the comment though.

jeremylenz:
@parthaa updated :)

parthaa (approved):
ACK

PR #9958, opened 2022-02-17
Fixes #34469 - Update retain package versions in pulp

sjha4:
1. Ensure changes to retain_package_versions_count flows to pulp
2. Ensure retain_package_versions_count has valid non-negative values for only yum repos.
A)
1. Create/Update a yum repo with a negative retain_package_versions_count value.
2. Should throw a validation error.
B)
1. Create a yum repo with retain_package_versions_count set to some valid value
2. Check repo in pulp `curl https://`hostname`/pulp/api/v3/repositories/rpm/rpm/   --cert /etc/pki/katello/certs/pulp-client.crt  --key /etc/pki/katello/private/pulp-client.key | python -m json.tool`
3. The retain_package_versions_count should be set
4. Update retain_package_versions_count value on the repo
5. The updated value should reflect on the pulp repo.

### DIFF
--- a/app/models/katello/root_repository.rb
+++ b/app/models/katello/root_repository.rb
@@ -83,6 +83,7 @@ module Katello
     validate :ensure_valid_deb_constraints, :if => :deb?
     validate :ensure_no_checksum_on_demand
     validate :ensure_valid_mirroring_policy
+    validate :ensure_valid_retain_package_versions_count
     validates :checksum_type, :inclusion => {:in => CHECKSUM_TYPES}, :allow_blank => true
     validates :product_id, :presence => true
     validates :content_type, :inclusion => {
@@ -306,6 +307,16 @@ module Katello
       end
     end
 
+    def ensure_valid_retain_package_versions_count
+      return unless self.retain_package_versions_count
+      unless yum?
+        errors.add(:retain_package_versions_count, N_("is only allowed for Yum repositories."))
+      end
+      if self.retain_package_versions_count.to_i < 0
+        errors.add(:retain_package_versions_count, N_("must not be a negative value."))
+      end
+    end
+
     def custom_content_path
       parts = []
       # We generate repo path only for custom product content. We add this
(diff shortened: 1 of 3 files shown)


sjha4:
[test katello]

rverdile (commented):
In the new repository screen, if I put a negative value and click "save" I get the error message. _However_, when I replace it with a positive value and click "save" nothing happens and the repository does not create. I don't see errors in the console.  I have to refresh the page to save it with a valid value.

Are you seeing that too?

app/models/katello/root_repository.rb
```diff
@@ -306,6 +307,16 @@ def ensure_valid_deb_constraints
       end
     end
 
+    def ensure_valid_retain_package_versions_count
+      return unless self.retain_package_versions_count
+      unless yum?
+        errors.add(:retain_package_versions_count, N_("is only allowed for Yum repositories."))
```
rverdile, line 313: When should I see this error pop up? I'm able to successfully do something like this: 

`hammer repository create --organization-id 1 --product-id 1 --content-type file --name "file 3" --retain-package-versions-count 1`

sjha4: Do you also see the retain package version being passed to katello and being set on the repo? I remember seeing a BZ around hammer ignoring that param.

rverdile: it gets passed to katello, but it isn't set on the repo because of this conditional: https://github.com/Katello/katello/blob/master/app/controllers/katello/api/v2/repositories_controller.rb#L571.

But it looks like updating the repository will trigger the error message.

sjha4: > In the new repository screen, if I put a negative value and click "save" I get the error message. _However_, when I replace it with a positive value and click "save" nothing happens and the repository does not create. I don't see errors in the console. I have to refresh the page to save it with a valid value.
> 
> Are you seeing that too?

I see this..But I see this in master too with any validation error on the repo page, ex: duplicate name.

rverdile: okay, not a problem here then :)

PR #9959, opened 2022-02-17
Fixes #34470 - Hide columns that are nil on lce page

chris1984:

Added `loading=true` to the start of the page until the data gets loaded in
Found out that errata count `total` and `security` were coming back as null instead of 0 so fixed that in the rabl file for errata
Added an `ng-hide` to the table to hide it until the data gets loaded in from the API

Verified the errata controller and module tests still function
Verified the lce page shows more columns when there is content loaded

* Create an org with 0 content in it
* Goto Content -> Lifecycle Environments
* Watch it show ostree/docker/errata then quickly hide ostree/docker
* Apply patch and restart devel box
* View page and columns should be blank except for content views and products

### DIFF
--- a/app/views/katello/api/v2/errata/_counts.json.rabl
+++ b/app/views/katello/api/v2/errata/_counts.json.rabl
@@ -1,7 +1,7 @@
 totals = @object.relation.group(:errata_type).count.with_indifferent_access
 
 node :security do |_presenter|
-  totals[:security]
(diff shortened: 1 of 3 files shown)


parthaa:
[test katello]

lfu:
[test katello]

chris1984:
@lfu all green

lfu (approved):

PR #9961, opened 2022-02-17
Fixes #34477 - Add act key and host ids to CV Api

sjha4:
Add hosts and activation key Ids to the environment node for CV. This is to support the bulk deletion flow for versions.
Add an activation key and host to an env and a cv.
Go to CV index page/details page and check the API response. There should be activation_keys and hosts keys in the environment node with an array of Ids.

### DIFF
--- a/app/views/katello/api/v2/content_views/base.json.rabl
+++ b/app/views/katello/api/v2/content_views/base.json.rabl
@@ -37,11 +37,15 @@ node :last_published do |content_view|
   end
 end
 
-child :environments => :environments do
-  attributes :id, :name, :label
-  node :permissions do |env|
+node :environments do |cv|
+  cv.environments.map do |env|
     {
-      :readable => env.readable?
+      id: env.id,
+      label: env.label,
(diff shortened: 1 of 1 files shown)


parthaa:
[test katello]

app/views/katello/api/v2/content_views/base.json.rabl
```diff
@@ -39,6 +43,12 @@ end
 
 child :environments => :environments do
   attributes :id, :name, :label
+  node :activation_keys do |env|
+     @resource&.activation_keys&.in_environment(env)&.ids
+  end
+  node :hosts do |env|
+     @resource&.hosts&.in_environment(env)&.ids
+    end
```
sjha4, line 51: @resource is not being set in the environment child scope for the base rabl when coming from index template. It is set when coming from show template. 
Trying to debug that. For now, environments node has the activation key and hosts Ids set only in CV details API call.

Andrewgdewar (approved):
ACK!

PR #9962, opened 2022-02-17
Fixes #34158 - Status translation fix

Andrewgdewar:

Fixes an issue with status translations that would either cause a lack of translation in the UI or a translated (incorrect value) as a status sent to the API.

As their appears to be a lack of translation in other languages, the only way I was able to confirm my change was the following...

- append a value to one of STATUS_TRANSLATIONS_ENUM's values 
- Navigate to a composite content view, select a status from the dropdown.
- check network tab, the status values should be the same as hardcoded within the enum above.

### DIFF
--- a/webpack/scenes/ContentViews/ContentViewsConstants.js
+++ b/webpack/scenes/ContentViews/ContentViewsConstants.js
@@ -56,9 +56,15 @@ export const addComponentSuccessMessage = component => (component ? __('Updated 
 
 
 // Repo added to content view status display and key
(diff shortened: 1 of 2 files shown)


Andrewgdewar:
I am seeing consistent failures in react test on CI that I can't reproduce locally.

Andrewgdewar:
Trying a rebase.

sjha4 (approved):
Works well and change looks good.

PR #9963, opened 2022-02-21
Fixes #34483 - Pulp 3 migration should fail earlier with missing errata

ianballou:
Unmigrated errata are reported after the `content prepare` step now.

What is important for a user to see? I think the repository ID and name are the most helpful.

I put the check right in the rake task because it seemed the most straight-forward.

1) Run the migration on 3.18 with **a few** yum repositories.
2) Edit the repository errata to have `nil` `erratum_pulp3_href`s.
3) Run the migration again and the switchover.
4) See the error from the bug at the beginning of the switchover.
5) Patch in my PR.
6) Run the migration again.  See the new error.
7) Run the migration yet again with `reimport_all=true` this time.
8) The migration and switchover should now complete successfully if run.

### DIFF
--- a/lib/katello/tasks/pulp3_migration.rake
+++ b/lib/katello/tasks/pulp3_migration.rake
@@ -47,6 +47,15 @@ namespace :katello do
               exit(-1)
             end
           end
+          repos_missing_errata = ::Katello::Repository.joins("inner join katello_repository_errata on katello_repositories.id = katello_repository_errata.repository_id").
+            joins("inner join katello_root_repositories on katello_repositories.root_id = katello_root_repositories.id").
+            where("katello_repository_errata.erratum_pulp3_href is null").pluck(:repository_id, :name).uniq
+          if repos_missing_errata.present?
+            puts "\nRepositories with the following IDs and names have unmigrated errata:\n"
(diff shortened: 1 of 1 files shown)


lib/katello/tasks/pulp3_migration.rake
```diff
@@ -47,6 +47,14 @@ namespace :katello do
               exit(-1)
             end
           end
+          repos_missing_errata = ::Katello::Repository.joins("inner join katello_repository_errata on katello_repositories.id = katello_repository_errata.repository_id").
+            joins("inner join katello_root_repositories on katello_repositories.root_id = katello_root_repositories.id").
+            where("katello_repository_errata.erratum_pulp3_href is null").pluck(:repository_id, :name).uniq
+          if repos_missing_errata.present?
+            puts "\nRepositories with the following IDs and names have unmigrated errata:\n"
```
jlsherrill, line 54: I wonder if we should give them a recommendation step, such as 'Resync the repositories'?

jlsherrill: and maybe re-running the 'content prepare' step with reimport_all=true

ianballou:
@jlsherrill I've updated the wording!

jlsherrill (approved):

PR #9964, opened 2022-02-22
Fixes #34355 - Remove parent_commit for ostree

jlsherrill:
its no longer needed, nor supported. So we can remove it.  

Test ostree upload:

wget --no-parent -r https://fixtures.pulpproject.org/ostree/small/
tar --exclude="index.html" -cvf "fixtures_small_repo.tar" -C fixtures.pulpproject.org/ostree "small"\


chris1984 (approved):
ACK, code looks good and it works as expected

PR #9965, opened 2022-02-22
Fixes #34488 - Delete any errata tied to ostree or puppet repos

sjha4:
For some reason, if ostree or puppet repos end up having errata associated, the deletion of those repos as part of the above migration fails.
Create a Katello::RepositoryErratum record with an erratum and a puppet/ostree repo.
Run the migration.
The repo should get deleted.


sjha4:
[test katello]

rverdile (approved):
lgtm!

PR #9966, opened 2022-02-22
Fixes #34491 - Redirect to repo/cv tab on CV create

sjha4:
Creating repo should redirect to repository/content views tab for component/composite cvs instead of versions page
Create a component CV
You should land on repository tab of new CV.
Create a composite CV
You should land on content view tab of the new CV.


chris1984 (approved):
ACK, works good

PR #9967, opened 2022-02-22
Fixes #34490 - Changing architecture setting for some rh repos causes issues

rverdile:
This has two changes:
1. The `find_repository` method used to find red hat repositories will now search without "arch", if it cannot find a repository with a particular "arch".
2. Deleting from the UI will delete by repository ID instead of filtering by product_id, content_id, minor, and arch.

This is to fix bugs where a repository couldn't be disabled, or the repository could be enabled despite already being enabled.
The arch setting is filtered out when using certain repo sets e.g. Red Hat Enterprise Linux 8 for x86_64 - AppStream RPMs, because it is not needed as a substitution in the content url (see here: https://projects.theforeman.org/issues/28644). When this happens, repositories from this set are always handled as having a "noarch" arch setting. If you change the arch setting to "x86_64" for a repository from this set, that repository cannot be disabled and is also not detected as an enabled repository (and so you can enable this repository a second time).

The reason for the first change is to make repositories find-able regardless of the arch setting. 

The reason for the second change is to prevent the first change from causing more weird behavior. If someone, because of this bug, enabled a duplicate repository, then the first change could prevent that user from easily disabling one of their repositories (since the query would find 2 repositories).
**Without PR**
1. Enable a repository from Red Hat Enterprise Linux 8 for x86_64 - AppStream RPMs.
  - `hammer repository-set enable  --name "Red Hat Enterprise Linux 8 for x86_64 - AppStream (RPMs)" --organization-id 1 --product "Red Hat Enterprise Linux for x86_64" --releasever "8.3"` or UI.
 2. Change the "Restrict to architecture" setting of this repository to "x86_64".
 3. In the UI, look at this list of enable-able repositories for this repository set. Notice, 8.3 is once again enable-able, despite already being enabled.
 4. Try to disable this repository; you can't.

**With PR**
1. Follow the same steps as above. This time, the repository should disable and you shouldn't see a duplicate listing.

### DIFF
--- a/app/controllers/katello/api/v2/repository_sets_controller.rb
+++ b/app/controllers/katello/api/v2/repository_sets_controller.rb
@@ -100,6 +100,7 @@ module Katello
     api :PUT, "/repository_sets/:id/disable", N_("Disable a repository from the set")
     api :PUT, "/products/:product_id/repository_sets/:id/disable", N_("Disable a repository from the set")
     param :id, :number, :required => true, :desc => N_("ID of the repository set to disable")
+    param :repository_id, :number, :required => false, :desc => N_("ID of the repository within the set to disable")
     param :product_id, :number, :required => false, :desc => N_("ID of the product containing the repository set")
     param :basearch, String, :required => false, :desc => N_("Basearch to disable")
     param :releasever, String, :required => false, :desc => N_("Releasever to disable")
@@ -186,7 +187,7 @@ module Katello
     end
 
     def substitutions
-      params.permit(:basearch, :releasever).to_h
+      params.permit(:basearch, :releasever, :repository_id).to_h
     end
 
     def find_authorized_activation_key
--- a/app/lib/actions/katello/repository_set/disable_repository.rb
+++ b/app/lib/actions/katello/repository_set/disable_repository.rb
@@ -7,9 +7,14 @@ module Actions
         end
 
         def plan(product, content, options)
-          repository = repository_mapper(product,
-                                         content,
-                                         options).find_repository
+          if options[:repository_id]
+            repository = ::Katello::Repository.find(options[:repository_id])
+          else
+            repository = repository_mapper(product,
+                                           content,
+                                           options).find_repository
+          end
+
           if repository
             action_subject(repository)
             plan_action(Repository::Destroy, repository)
--- a/app/models/katello/candlepin/repository_mapper.rb
+++ b/app/models/katello/candlepin/repository_mapper.rb
@@ -10,10 +10,17 @@ module Katello
       end
 
       def find_repository
-        root = ::Katello::RootRepository.where(product_id: product.id,
-                                    content_id: content.cp_content_id,
-                                    minor: minor,
-                                    arch: arch).first
+        if substitutions[:basearch]
+          root = ::Katello::RootRepository.where(product_id: product.id,
+                                                 content_id: content.cp_content_id,
+                                                 minor: minor,
+                                                 arch: arch).first
+        else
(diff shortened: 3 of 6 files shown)


app/controllers/katello/api/v2/repository_sets_controller.rb
```diff
@@ -100,6 +100,7 @@ def enable
     api :PUT, "/repository_sets/:id/disable", N_("Disable a repository from the set")
     api :PUT, "/products/:product_id/repository_sets/:id/disable", N_("Disable a repository from the set")
     param :id, :number, :required => true, :desc => N_("ID of the repository set to disable")
+    param :repo_id, :number, :required => false, :desc => N_("ID of the repository within the set to disable")
```
jlsherrill, line 103: is this used?

app/lib/actions/katello/repository_set/disable_repository.rb
```diff
@@ -7,9 +7,14 @@ def humanized_name
         end
 
         def plan(product, content, options)
-          repository = repository_mapper(product,
-                                         content,
-                                         options).find_repository
+          if options[:repo_id]
```
jlsherrill, line 10: oh i see it here

app/controllers/katello/api/v2/repository_sets_controller.rb
jlsherrill: is the idea that the user could just pass in the repo_id in case other stuff isn't working?

jlsherrill: could we change this to repository_id ?

app/models/katello/candlepin/repository_mapper.rb
```diff
@@ -13,7 +13,10 @@ def find_repository
         root = ::Katello::RootRepository.where(product_id: product.id,
                                     content_id: content.cp_content_id,
                                     minor: minor,
-                                    arch: arch).first
+                                    arch: arch).or(
+               ::Katello::RootRepository.where(product_id: product.id,
+                                               content_id: content.cp_content_id,
+                                               minor: minor)).first
```
jlsherrill, line 19: this seemed to cause a bug, if i go to the RH repos page and search for 'label = rhel-6-server-rpms'  
Then enable the 'x86_64 6Server' repo, it won't let me enable the 'i386 6Server' repo.  I think because its not finding the arch i specified (i386), so then it excludes the arch in the search and finds the x86_64 one, os it thinks its already enabled.  Should we search for the arch passed in or nil?

jlsherrill: something like this (untested)
```
::Katello::RootRepository.where(product_id: product.id,
                                    content_id: content.cp_content_id,
                                    minor: minor,
                                    arch: [arch, nil]).first
```

app/controllers/katello/api/v2/repository_sets_controller.rb
jlsherrill: i do like this approach though, it means its always able to be disabled from the UI!!!

app/models/katello/candlepin/repository_mapper.rb
rverdile: Oh, good catch. Yeah, I think searching for `nil` in the second search would work. I'm not sure I'll check :).

rverdile: so it looks the issue with that would be the arch is set to "noarch" by default, so it arch would never be nil.

Fixing the duplicate repository issue might be a tricky. I have to dig around again.

rverdile:
updated

jlsherrill (approved):

PR #9969, opened 2022-02-22
Fixes #34494 - Make primary actions on tabs Primary buttons

sjha4:
All primary actions on CV tabs should be primary buttons to make them more obvious and prominent on the UI.
The secondary buttons we currently use are white/grey and not immediately obvious to users based on initial feedback.
All primary action buttons on tabs, ex: Add repositories, add content view, add filter, add package group to filter etc etc should be primary buttons(Blue/Grey).


Andrewgdewar (approved):
giphy

PR #9972, opened 2022-02-22
Fixes #34497 - Pulp 3 migration underestimates migration time

ianballou:
Adds a suggested migration time for servers with slower hardware and/or higher server loads.  Currently just a factor of 4x based on the information in the redmine issue.

There isn't enough time to do more testing to determine a proper migration timing. The data will need to come from actual users if it's desired to get a more accurate time.

Run the migration stats command with a decent number of RPMs so it says something other than "fewer than 5 minutes".  Check that the timing increase matches the factor in the code and the hours + minutes make sense.


ianballou:
[test katello]

ianballou:
[test katello]

PR #9974, opened 2022-02-23
Fixes #34508 - Creating repo stop working after any validation error

sjha4:
Creating a repo fails if there has been a validation error once. ex: create a repo with duplicate name.
Also, updating a docker repo fails if you add a comma at the end of include tags or exclude tags in the form.

1. Create a repo with an existing name, or negative retain package count value etc to trigger a validation error.
2. Correct the repo in the form to be valid
3. It should be able to save properly.

1. Create a docker repo with some include tags and exclude tags.
2. You should be able to update the same repo with changes to the tag lists or otherwise.


chris1984 (approved):
ACK, was able to create a repo after changing the name and adjust the tags on a docker repo I made.

PR #9978, opened 2022-02-24
Fixes #34519 - Check CV and LCE while updating CDN config

sjha4:
When updating Upstream CDN config, check for existence of CV and LCE.
Have the ISS setup in 2 boxes.
In the downstream server, Under CDN configuration for Upstream server > Enter an invalid CV or LCE.
The update should fail.
Update should pass with valid CV/LCE.


app/lib/katello/resources/cdn/katello_cdn.rb
```diff
@@ -36,12 +36,16 @@ def valid_path?(_path, _postfix)
           true
         end
 
+        def validate_cv_lce!
```
parthaa, line 39: wondering if can call this `validate!` instead and also include the checks for the organization like 
``` organization && content_view_id && lifecycle_environment_id```

File: app/controllers/katello/api/v2/content_views_controller.rb [hunk 2]
```diff
@@ -54,6 +54,7 @@ def filtered_associations
     param :composite, :bool, :desc => N_("Filter only composite content views")
     param :without, Array, :desc => N_("Do not include this array of content views")
     param :name, String, :desc => N_("Name of the content view"), :required => false
+    param :label, String, :desc => N_("Label of the content view"), :required => false
```
parthaa, line 57: seems unnecessary given you can use `search`

app/lib/katello/resources/cdn/katello_cdn.rb
```diff
@@ -36,12 +36,16 @@ def valid_path?(_path, _postfix)
...
+          content_view_id && lifecycle_environment_id
+        end
+
         def debug_certificate
           get("/katello/api/v2/organizations/#{organization['id']}/download_debug_certificate")
         end
 
         def content_view_id
-          rs = get("/katello/api/v2/organizations/#{organization['id']}/content_views?name=#{CGI.escape(@content_view_label)}")
+          rs = get("/katello/api/v2/organizations/#{organization['id']}/content_views?search=#{CGI.escape(`label=#{@content_view_label}`)}")
```
parthaa, line 48: whats up with the back ticks. I think ruby tries to run that as an executable
```suggestion
          rs = get("/katello/api/v2/organizations/#{organization['id']}/content_views?search=#{CGI.escape('label=#{@content_view_label}')}")
```

sjha4: Hmm..React hangover..

app/controllers/katello/api/v2/content_views_controller.rb
sjha4: I'd propose leaving it here unless you see an issue with it?

app/lib/katello/resources/cdn/katello_cdn.rb
```diff
@@ -36,12 +36,16 @@ def valid_path?(_path, _postfix)
           true
         end
 
+        def validate!
+          content_view_id && lifecycle_environment_id && organization
```
parthaa, line 40: Check for org first since the other 2 depend on org
```suggestion
          organization && content_view_id && lifecycle_environment_id
```

sjha4:
[test katello]

parthaa:
Almost there. Works mostly well. Suggest one patch. Basically if its airgapped we want the upstream url tin the repo to be nil. 

```patch
diff --git a/app/lib/actions/katello/cdn_configuration/update.rb b/app/lib/actions/katello/cdn_configuration/update.rb
index de223da360..1ab406d6ef 100644
--- a/app/lib/actions/katello/cdn_configuration/update.rb
+++ b/app/lib/actions/katello/cdn_configuration/update.rb
@@ -4,7 +4,6 @@ module Actions
       class Update < Actions::EntryAction
         def plan(cdn_configuration, options)
           cdn_configuration.update!(options)
-          return if cdn_configuration.airgapped?
 
           if cdn_configuration.upstream_server?
             resource = ::Katello::Resources::CDN::CdnResource.create(cdn_configuration: cdn_configuration)
@@ -21,7 +20,7 @@ module Actions
           roots.each do |root|
             full_path = if cdn_configuration.redhat_cdn?
                           root.product.repo_url(root.library_instance.generate_content_path)
-                        else
+                        elsif cdn_configuration.upstream_server?
                           resource.repository_url(content_label: root.content.label)
                         end
             plan_action(::Actions::Katello::Repository::Update, root, url: full_path
```

parthaa:
rebase this PR to latest master react should go green.

sjha4:
> ```diff
> cdn_configuration
> ```

Is the intent of this to update all repos in airgapped setup with url null?

parthaa:
> Is the intent of the suggested change to update all redhat repos in airgapped setup with url null? That makes sense based on definition of the 3 modes. 

Ideally yes. But I don't think its necessary to update existing rh repos because import operation will automatically correct the URL.

app/controllers/katello/api/v2/content_views_controller.rb
parthaa: I am ok leaving it. May be you could add this parameter to environments controller index also?

parthaa: Just to be consistent

parthaa (approved):
ACK works well

parthaa:
[test katello]

sjha4:
[test katello]

PR #9979, opened 2022-02-24
Fixes #34520 - Promote on contentViewVersion Details page

Andrewgdewar:

This converts a menu item to a primary button and makes the "Actions" selection button a kebab to match consistency in other areas.

Open ContentViewVersion Details page and press the fancy new "Promote" button. Ensure it still works as intended.
Screen Shot 2022-02-24 at 2 34 52 PM

### DIFF
--- a/webpack/scenes/ContentViews/Details/Versions/VersionDetails/ContentViewVersionDetailsHeader.js
+++ b/webpack/scenes/ContentViews/Details/Versions/VersionDetails/ContentViewVersionDetailsHeader.js
@@ -2,6 +2,7 @@ import React, { useState, useEffect } from 'react';
 import { useDispatch } from 'react-redux';
 import PropTypes from 'prop-types';
 import {
+  Button,
   Grid,
   GridItem,
   TextContent,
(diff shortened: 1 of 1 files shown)


jeremylenz (commented):
Works well, just one comment:

disabled_promote

Shouldn't we disable the Promote button if none of the environments would be available? 
(I mean the Promote button that brings up the modal.)

Andrewgdewar:
> Works well, just one comment:
> 
> disabled_promote
> 
> Shouldn't we disable the Promote button if none of the environments would be available? (I mean the Promote button that brings up the modal.)

I think this is a design decision. As simply disabling the promote button without an explicit (to the user) reason wouldn't be the best.

You could think of a situation where many environments exist. Perhaps all are selected for promotion.. but the user doesn't know that. How do we tell them, other than a disabled button. @jeremylenz

jeremylenz (approved):
yeah, you make good points!  I won't hold this up further :)

APJ :+1:

jeremylenz:
[test katello]

PR #9980, opened 2022-02-24
Fixes #34496 - Redirect to versionDetails after publish

Andrewgdewar:

It redirects the user to the CVV details page when the publish completes.

- Publish a new content view version
- Wait for the modals progress bar to reach 100%
- The user should now be redirected to the contentViewVersionDetails of the freshly created version. 

NOTE: This will not attempt to redirect the user if they exit the modal while the task is unfinished.


webpack/scenes/ContentViews/Publish/CVPublishFinish.js
```diff
@@ -46,9 +49,8 @@ const CVPublishFinish = ({
     if (currentStep !== 1) {
       dispatch(stopPollingTask(cvVersionPublishKey(cvId, versionCount)));
       setCurrentStep(1);
+      setPolling(false);
       setIsOpen(false);
-      dispatch(getContentViewDetails(cvId));
-      dispatch(getContentViews);
```
sjha4, line 51: We do need these to happen to trigger a second publish without refresh. Without these trying to publish fails without refresh after a publish.

webpack/scenes/ContentViews/Publish/CVPublishFinish.js
```diff
@@ -161,7 +163,8 @@ const CVPublishFinish = ({
   }
   if (status === STATUS.PENDING) return (<Loading />);
   if (status === STATUS.ERROR) return (<EmptyStateMessage error={error} />);
-  return <Loading />;
+  const { input: { content_view_version_id: cvvID } } = response;
+  return <Redirect to={`/content_views/${cvId}#/versions/${cvvID}`} />;
```
sjha4, line 167: This doesn't handle the case where the wizard is closed manually. It lands on the same tab and doesn't refresh the view.

Andrewgdewar: If they close tab manually, I would expect the user should not be redirect.. In my tests it does refresh upon closing it manually.

sjha4:
Works well. However, seeing 3 calls to version details when the task finishes.

sjha4 (approved):
Ack 👍🏼

PR #9981, opened 2022-02-25
Fixes #34522 - Add deprecation warnings for non-SCA

jeremylenz:

Traditional (non-SCA) entitlements and subscription management will be deprecated starting in Katello 4.5 and possibly removed in 4.7.  This updates all subscription-related UI pages with deprecation warning banners.

Deprecation warning banners in the following locations:

* Content hosts detail pages
* Activation Keys > detail page
* Subscriptions main page

Wording is not yet finalized.
The link opens in a new tab. I thought that would be appropriate since it's the way the Subscription Watch link works as well.

Register a host in both an SCA and a non-SCA org and view the banners.  Or just look at the screen shots:

non_sca_content_host_warn
non_sca_subs_page_warn
non_sca_ak_warning

These banners display in the same place as the SCA banners.  So if SCA is turned on, you'll instead see those:
sca_content_host
sca_subs_page
sca_ak

### DIFF
--- a/engines/bastion/app/views/bastion/layouts/assets.html.erb
+++ b/engines/bastion/app/views/bastion/layouts/assets.html.erb
@@ -19,7 +19,7 @@
     angular.module('Bastion').value('currentLocale', '<%= I18n.locale %>');
     angular.module('Bastion').value('CurrentOrganization', "<%= Organization.current.id if Organization.current %>");
     angular.module('Bastion').value('simpleContentAccessEnabled', <%= Organization.current.simple_content_access? if Organization.current %>);
-    angular.module('Bastion').value('isManifestImported', <%= !!Organization.current&.owner_details&.[](:upstreamConsumer)&.[](:webUrl) %>)
+    angular.module('Bastion').value('isManifestImported', <%= !!Organization.current&.manifest_imported? %>)
     angular.module('Bastion').value('foreman', tfm);
     angular.module('Bastion').value('repositoryTypes', angular.fromJson('<%= Katello::RepositoryTypeManager.enabled_repository_types.values.to_json.html_safe %>'));
     angular.module('Bastion').value('deleteHostOnUnregister', angular.fromJson('<%= Setting[:unregister_delete_host] %>'));
--- a/engines/bastion/app/views/bastion/layouts/assets.html.erb
+++ b/engines/bastion/app/views/bastion/layouts/assets.html.erb
@@ -19,6 +19,7 @@
     angular.module('Bastion').value('currentLocale', '<%= I18n.locale %>');
     angular.module('Bastion').value('CurrentOrganization', "<%= Organization.current.id if Organization.current %>");
     angular.module('Bastion').value('simpleContentAccessEnabled', <%= Organization.current.simple_content_access? if Organization.current %>);
+    angular.module('Bastion').value('isManifestImported', <%= !!Organization.current&.owner_details&.[](:upstreamConsumer)&.[](:webUrl) %>)
     angular.module('Bastion').value('foreman', tfm);
     angular.module('Bastion').value('repositoryTypes', angular.fromJson('<%= Katello::RepositoryTypeManager.enabled_repository_types.values.to_json.html_safe %>'));
     angular.module('Bastion').value('deleteHostOnUnregister', angular.fromJson('<%= Setting[:unregister_delete_host] %>'));
--- a/engines/bastion/app/views/bastion/layouts/assets.html.erb
+++ b/engines/bastion/app/views/bastion/layouts/assets.html.erb
@@ -19,7 +19,7 @@
     angular.module('Bastion').value('currentLocale', '<%= I18n.locale %>');
     angular.module('Bastion').value('CurrentOrganization', "<%= Organization.current.id if Organization.current %>");
     angular.module('Bastion').value('simpleContentAccessEnabled', <%= Organization.current.simple_content_access? if Organization.current %>);
-    angular.module('Bastion').value('isManifestImported', <%= !!Organization.current&.manifest_imported? %>)
+    angular.module('Bastion').value('isManifestImported', <%= !!Organization.current&.manifest_imported?(cached: true) %>)
     angular.module('Bastion').value('foreman', tfm);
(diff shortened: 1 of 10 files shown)


jlsherrill:
do non-redhat users have the ability to activate SCA yet?

jeremylenz:
> do non-redhat users have the ability to activate SCA yet?

@jlsherrill No, this has been talked about but it's not a feature yet.

jlsherrill:
Does it make sense to show this warning only if a manifest is imported?    If i'm an upstream user without a Red Hat manifest, this will only cause me confusion

jeremylenz:
> Does it make sense to show this warning only if a manifest is imported?

Definitely yes. Will update.

jeremylenz:
updated to only show banner when a manifest is imported.

webpack/scenes/Subscriptions/SubscriptionsPage.js
```diff
@@ -271,7 +272,7 @@ class SubscriptionsPage extends Component {
             />
 
             <div id="subscriptions-table" className="modal-container">
-              {SCAAlert}
+              {isManifestImported && SCAAlert}
```
jlsherrill, line 275: you can just use this as defined in angular?  how does that work?

jeremylenz: It just happens to be named the same ;) 

going from Angular archaeology to React archaeology..

It's a Redux selector used here -> https://github.com/jeremylenz/katello/blob/0e11585c3859472a2decc4e6579cda25f9753855/webpack/scenes/Subscriptions/index.js#L38-L39

jeremylenz: It was very convenient that we were already using it on this page, so that code was already there

jlsherrill: ahahha makes sense!

engines/bastion/app/views/bastion/layouts/assets.html.erb
```diff
@@ -19,6 +19,7 @@
     angular.module('Bastion').value('currentLocale', '<%= I18n.locale %>');
     angular.module('Bastion').value('CurrentOrganization', "<%= Organization.current.id if Organization.current %>");
     angular.module('Bastion').value('simpleContentAccessEnabled', <%= Organization.current.simple_content_access? if Organization.current %>);
+    angular.module('Bastion').value('isManifestImported', <%= !!Organization.current&.owner_details&.[](:upstreamConsumer)&.[](:webUrl) %>)
```
parthaa, line 22: Can you move the logic to a method in owner.rb
```suggestion
    angular.module('Bastion').value('isManifestImported', <%= Organization.current&.manifest_imported? %>)
```

jlsherrill: the method itself already exists! https://github.com/Katello/katello/blob/master/app/models/katello/concerns/organization_extensions.rb#L94

jeremylenz: @parthaa updated!

jeremylenz: @jlsherrill good catch!  Deleted the method I wrote in glue, which used
```rb
owner_details&.[](:upstreamConsumer)&.[](:webUrl)
```
in favor of the existing method which only uses
```rb
owner_details['upstreamConsumer'].present?
```

Now we're using two different sets of logic to decide if a manifest is imported. Should I also change the React selector to use the latter logic?

jeremylenz: cc @jturel I feel like you were the one who originally used `webUrl`..

jlsherrill: also i noticed in this pr: https://github.com/Katello/katello/pull/8859/files#diff-3f8d4b7127ce10e6e2bc95774bd0b8df3822f0c4075e08efd96d4b2fae6f39b4R38-R41  that some caching comes into play, i presume to reduce calls to candlepin.  I wonder if we should do that here since this is happening on every angular page load (some background on that change from @jturel and @chris1984 might be good if they remember)

jturel: > cc @jturel I feel like you were the one who originally used `webUrl`..

I don't think so because `&.[]` is not in my repertoire but I did write `manifest_imported?`! I'm +1 to using that method here. I also think you could enhance it to use caching similar to the `simple_content_access?` method. However, I'd make the the default `cached: false` so that the whole app doesn't need to be audited to ensure that it's not incorrectly reading a cached value when it really needs an uncached value. So you can just call it with `cached: true` here

jeremylenz: Updated :+1:

jturel:
That branch name!!! :rofl: :rofl: :rofl: :rofl:

engines/bastion/app/views/bastion/layouts/assets.html.erb
```diff
@@ -19,6 +19,7 @@
     angular.module('Bastion').value('currentLocale', '<%= I18n.locale %>');
     angular.module('Bastion').value('CurrentOrganization', "<%= Organization.current.id if Organization.current %>");
     angular.module('Bastion').value('simpleContentAccessEnabled', <%= Organization.current.simple_content_access? if Organization.current %>);
+    angular.module('Bastion').value('isManifestImported', <%= !!Organization.current&.manifest_imported?(cached: true) %>)
```
jturel, line 22: The `!!` is now vestigial thanks to your change

jeremylenz: I almost deleted it. But if `Organization.current` returns `nil`, without the `!!` this may evaluate to `nil` due to the `&.`, and the  `!!` guarantees it'll be a boolean.

jeremylenz: in other words I didn't feel like adding `if Organization.current` at the end.

jturel: The teacher becomes the pupil! That's fair enough

jlsherrill:
looks like you'll need to update the angular tests for the controllers you modified to a pass in isManifestImported

chris1984 (approved):
Angular code looks good, ack

PR #9983, opened 2022-02-25
Fixes #34529 - Lint failure due to uncontrolled eslint-plugin

Andrewgdewar:

The lint package, eslint-plugin-react, was updated from: "7.28.0" to "7.29.1" due to the lack of a committed package-lock in foreman.

Related changelog for eslint-plugin-react [here](https://github.com/yannickcr/eslint-plugin-react/blob/master/CHANGELOG.md).

This fixes some breaking lint changes introduced by that update.

- I thought it prudent to use the updated rules instead of just turning them off for a solution. 
- If this becomes problematic or slows down the dev process, one can simply change two lines in the eslintrc.js file.

`
'indent': 'off',
'react/forbid-prop-types': 'off',
`

- Check if the Katello lint and react tests pass on CI.
- ???? 
- Profit

### DIFF
--- a/webpack/components/WithOrganization/withOrganization.js
+++ b/webpack/components/WithOrganization/withOrganization.js
@@ -1,3 +1,4 @@
+/* eslint-disable react/jsx-indent */
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { bindActionCreators } from 'redux';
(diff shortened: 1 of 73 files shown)


webpack/components/WithOrganization/withOrganization.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react/jsx-indent */
```
sjha4, line 1: Do we need this even after ignoring all the JSX* nodes in the lint rule?

Andrewgdewar: Yes. In this one case it is needed. I have no idea why.

Andrewgdewar: First question (longer answer), there are probably a hundred "ignoredNodes" options for the indent lint rule. I'm not sure which one was being applied in this file. It was behaving differently than the rest. Unsure as to why. 

Second question: 
 eslint --fix

jeremylenz (approved):
APJ :+1:

PR #9985, opened 2022-02-28
Fixes #34528 - Show puppet module count on composite cv add page

chris1984:

* Looking through the code, we deleted a method for the puppet module count, so recreated that and added a test on the method.

* Made sure nothing else broke around that area.

* Spin up a Katello 3.18 or Sat 6.9 box
* Create a custom product and repo of puppet type
* Upload a puppet module 
* Create a content view with the puppet module and publish
* Create a composite content view and add the content view, see that the puppet count is 0
* Apply patch and restart services
* Verify module count is correctly showing

### DIFF
--- a/app/models/katello/content_view_version.rb
+++ b/app/models/katello/content_view_version.rb
@@ -167,6 +167,10 @@ module Katello
       (!content_counts.blank? && content_counts.dig(PuppetModule::CONTENT_TYPE) > 0) || self.content_view.force_puppet_environment?
     end
 
+    def puppet_module_count
+      content_counts&.dig(PuppetModule::CONTENT_TYPE)
+    end
+
     def archived_repos
       self.default? ? self.repositories : self.repos(nil)
     end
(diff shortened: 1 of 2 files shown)


chris1984:
[test katello]

app/models/katello/content_view_version.rb
```diff
@@ -167,6 +167,10 @@ def promote_puppet_environment?
       (!content_counts.blank? && content_counts.dig(PuppetModule::CONTENT_TYPE) > 0) || self.content_view.force_puppet_environment?
     end
 
+    def puppet_module_count
+      content_counts.dig(PuppetModule::CONTENT_TYPE) if content_counts?
```
lfu, line 171: This line can be written as:
```
content_counts&.dig(PuppetModule::CONTENT_TYPE)
```

Where is `puppet_module_count` get called?

chris1984: @lfu I will fix that up, here is where we are looking:

https://github.com/Katello/katello/blob/KATELLO-3.18/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-views/details/components/views/content-view-composite-available-content-views.html#L77

chris1984: @lfu updated

lfu:
Screenshot from 2022-03-07 11-22-19
Puppet Modules count went from 0 to 1 after applying the fix.

lfu:
APJ

lfu (approved):

PR #9986, opened 2022-02-28
Fixes #34432 - Remove drpm from ignorable_content for roots

sjha4:
Adds a migration to remove drpm from irgnorable_content list for root repositories.
How I tested this:
1. Change katello/app/models/katello/root_repository.rb L20 to allow drpm as ignorable_content type:
`IGNORABLE_CONTENT_UNIT_TYPES = %w(srpm drpm).freeze`
2. Change katello/app/lib/actions/katello/repository/update.rb file L13 to add 
`repo_params[:ignorable_content] = ['srpm', 'drpm']`
3. The changes above mean that any yum repo you update will get srpm and drpm as ignorable content. So update a couple yum repos, ex: name in your setup.
4. Go to foreman console: bundle exec rails c
5. Run `Katello::RootRepository.select {|r| r&.ignorable_content&.include? "drpm"}`
6. You should see a few root repository with drpm in the ignorable content list
7. Checkout this PR and run bundle exec rails db:migrate
8. Check `Katello::RootRepository.select {|r| r&.ignorable_content&.include? "drpm"}`. You shouldn't have any records.

### DIFF
--- a/db/migrate/20220228173251_remove_drpm_from_ignorable_content.rb
+++ b/db/migrate/20220228173251_remove_drpm_from_ignorable_content.rb
@@ -1,0 +1,12 @@
+class RemoveDrpmFromIgnorableContent < ActiveRecord::Migration[6.0]
+  def up
+    Katello::RootRepository.select { |r| r&.ignorable_content&.include? "drpm" }.each do |root|
+      root.ignorable_content = root.ignorable_content - ["drpm"]
+      root.save!
+    end
+  end
+
+  def down
+    fail ActiveRecord::IrreversibleMigration
+  end
+end


rverdile (commented):
The migration forward works! Did you want it to be able to rollback? Or should it be irreversible? Right now if you rollback it doesn't do anything.

sjha4:
I don't think there's any way to make this reversible cause once the migration is run, we have no way of knowing which repos had drpm in ignorable content during rollback.

sjha4:
@rverdile : Updated.. 👍🏼

rverdile (approved):
lgtm!

PR #9987, opened 2022-02-28
Fixes #34530 - update smart proxy after setting unauthenticated pull

rverdile:
Smart proxy will automatically sync after updating "Unauthenticated Pull" in LCE details. Addresses bug where manual sync is necessary for unauthenticated pull update to take affect on smart proxy.
On Katello:
1. Create and sync a container repo
2. Create a CV and add the container repo to it
3. Publish and Promote the CV and make sure Capsule sync is triggered.
4. Wait for the Capsule sync to finish
5. Go the the LCE page and set "Unauthenticated Pull: Yes".

On Capsule:
1. `sqlite3 /var/lib/foreman-proxy/smart_proxy_container_gateway.db`
2. sqlite> `.header on`
3. sqlite> `select * from repositories;`
id|name|auth_required
6|busybox|1 <======== 0 when false, 1 when true. Should be updated after changing unauthenticated pull setting.


ianballou (approved):
Works for me!

PR #9942, opened 2022-02-11
Fixes #34320 - Hide Katello tabs & cards on new host detail page when host is unregistered

jeremylenz:
#### What are the changes introduced in this pull request?

We need to allow Katello to hide its tabs and cards from the new host detail page.
* Adds new helper functions `hostIsNotRegistered` and `hostIsRegistered`
* Check if host is registered and if not, hide Errata and Content View overview cards
* Also hide Content, Traces, and Repository sets tabs for non-registered hosts. The way it works is as follows:
  - Adds a new slot metadata attribute, `hideTab`, that is sent to Foreman. This allows us to pass a function to Foreman which Foreman will then call with the host details when setting up the page.
  - If https://github.com/theforeman/foreman/pull/9104 is applied, and `hideTab` returns true for that tab, Foreman will hide the tab from the host details page.
  - I'm using `hostIsNotRegistered` as the `hideTab` function here, but other functions could be used for other slots in the future.

* Also, recalculates errata counts on the host when unregistering to ensure they're properly cleared out.

#### Considerations taken when implementing this change?

Hiding cards I can do from Katello, but I found that changing Foreman code is the only clean way to hide an entire tab.
I found that passing the `hideTab` function to Foreman is a good way to keep Katello code in Katello and Foreman code in Foreman. This way Katello can be the one that cares about _why_ we're hiding the tab, and Foreman doesn't have to touch such ugliness as subscription facets :scream:.

#### What are the testing steps for this pull request?

* Apply this Katello change
* Register a host and get some installable errata on it
* Find its content_facet in Rails console
* Unregister the host
* View the content facet's attributes and verify that the errata counts are cleared out:
```
installable_security_errata_count: 0,
installable_enhancement_errata_count: 0,
installable_bugfix_errata_count: 0,
```

* Apply the Foreman change: https://github.com/theforeman/foreman/pull/9104
* Reload the host detail page and verify:
  - Installable Errata card does not display
  - Content View overview card does not display
  - Content tab does not display
  - Traces tab does not display
  - Repository sets tab does not display

* Reregister the host and verify that
  - all tabs are back
  - errata counts are back to what they were before

PR #9998, opened 2022-03-03
Fixes #34548 - Host collections Add & Remove modals

jeremylenz:

add_hc
remove_hc

* Add Host collections Add & Remove modals to Host collections overview card.
* Refactors `errorToast` in HostDetails (see relevant commit)

I was pleased to discover that there is already a single API for adding or removing host collections from a host.  The API is kinda weird (instead of saying which host collections to add or remove, you instead have to send _what you want the new list of host collections to be_) but works quite well and didn't require any changes!

Knowing that the two modals would be very similar, I combined them into one component. The use of ternaries etc. is a _bit_ higher than I'd like, but I think the tradeoffs are worth it.

Because it would cause significant complexity and is not a common use case, the tables don't have Select All available-- only Select Page. (this is ok per Maria)

* Create some host collections
* On the host collections overview, click the kebab > __Add host to collections__
* Select one or more available host collections and click Add
* Note that they appear on the card after the success toast appears
* Click the kebab >  __Remove host from collections__  - verify that the collections you just added are now showing as available to remove
* Select one or more available host collections and click Remove
* Click the kebab > __Add host to collections__ - verify that the host collection you just removed is available to add
* Make sure that any host collection at its host limit is not selectable in the table
* Change hosts using the breadcrumb switcher and verify everything still works correctly for the new host


jeremylenz:
rebased after Packages reorg PR merged.

jeremylenz:
rebased & incorporated card hiding feature from https://github.com/Katello/katello/pull/9942 :)

jeremylenz:
[test katello]

jturel (approved):
Wow, that was a surprisingly small amount of code (excluding tests) to wire this up. Nice!

Code looks good, everything works as far as I can tell, and I agree with the design considerations.

One observation: I noticed that certain actions led to the modal resizing itself. For example, I had one collection in the Add modal. Then I searched for a bogus collection which made the modal smaller. Any way to make that more stable?

jeremylenz:
> Any way to make that more stable?

Tried a few things with `minHeight` but didn't get very far. But then I discovered that `position: top` makes it a lot less jarring because the search box stays in the same place!  Also I think the other issue with the wonky shifting left/right should be fixed by https://github.com/theforeman/foreman/pull/9124.

jeremylenz:
[test katello]

jeremylenz:
[test katello]

PR #9999, opened 2022-03-04
Fixes #34559 - Patternfly update causes tabs to navigate twice on click

Andrewgdewar:

An update to how <a> tags are handled within the <Tabs> component in patternfly has caused a previous work-around to navigate to pages twice.
This removes the previous work-around that is no longer needed, and corrects the duplicate api-call/navigation behaviour.

On master: 

- Pull latest foreman/katello, 
- Delete node_modules/package-lock inside foreman. 
- npm install

- navigate to contentViewDetails (content_views/<int>#/XXXX)  
- go through versions/Repositories/filters/history tabs
- Without this patch you should see double api calls on all pages after navigation.

### DIFF
--- a/webpack/components/RoutedTabs/index.js
+++ b/webpack/components/RoutedTabs/index.js
@@ -1,33 +1,18 @@
 import React from 'react';
 import { shape, string, number, element, arrayOf } from 'prop-types';
 import { Tab, Tabs, TabTitleText } from '@patternfly/react-core';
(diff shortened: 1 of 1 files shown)


sjha4 (approved):
Fixes the double API calls.. 👍🏼

jeremylenz (approved):
Thanks @Andrewgdewar 

Without patch:

:white_check_mark: Verified double API calls on content view details

With patch:
:heavy_check_mark: No double API calls
:heavy_check_mark: Can still middle-click to open in a new browser tab

Yay for good old fashioned `<a>` tags! ACK

PR #10000, opened 2022-03-04
Fixes #34561 - Fix grammar on sync plan page

sjha4:
Create sync plan and go to products tab. 😄


jeremylenz (approved):
Such a momentous occasion for only a four-character change... :man_shrugging: 

ACK

PR #10002, opened 2022-03-08
Refs #34072 - Address useBulkSelect slowness when per_page is high

jeremylenz:

In tables that use `useBulkSelect` and `useSelectionSet`, table row selection was slow when per_page is set to a custom value of 100 or more. In other words, the time between the mouse click on the checkbox and the visual change to checked could be as much as 4-500ms. It was slowest on the Errata page, but the issue was also present on other tables. I was able to get it down to ~150, testing with 100 items per page. Not ideal but a significant improvement.

* Add `useCallback`, `useMemo`, and other optimizations to several `useBulkSelect` internal functions
* Add `useCallback` to one function in `ErrataTab.js`
* Don't render ErrataTab expanded table rows when the row is not expanded (This was the biggest source of improvement by far)

Ideally the goal should be 50-100ms for a user interaction (see https://web.dev/rail/?utm_source=devtools#goals-and-guidelines) so it's not perfect. The next thing I can think of is to have some sort of infinite scrolling component that only renders what's on screen (or just offscreen), but that would take a significant dev effort. Having a custom `perPage` value set seems like too small a use case to make that worth it.

1. Enable the Red Hat Repository "Red Hat Enterprise Linux 7 Server RPMs x86_64 7Server"
2. Make sure your Foreman VM has at least 16GB of RAM
3. Sync the repo and go have some coffee
4. Spin up a new RHEL7 box and register it. You should now have 100+ errata.
5. Administer > Settings > Set "Entries per page" to 100 or more
6. Hosts > All hosts > new host details for your new host > Content > Errata
7. Click a checkbox. Note the time from the mouse click to the box visually responding and becoming checked. Notice how excruciatingly slow it is.
8. Also look at Packages (you should have plenty) - The issue is still there but not as pronounced.  If you like, you can use DevTools Performance tab > CPU Throttling to magnify the issue (but remember to turn it off before doing the testing procedure below; it kept crashing on my system otherwise)
9. Navigate to the last page (which probably doesn't have a lot of items) and see how much faster it is
10. Perform the DevTools testing procedure below
11. Check out this change and perform the testing procedure again; Note your results.

#### DevTools testing procedure

1. Navigate to Content > Errata if you're not already there
2. In DevTools, click the Performance tab
3. Click the Record button in the upper left
4. Click to check two or three checkboxes - make sure to wait until the box is checked before clicking the next one
5. Click the select none button to clear the selection
6. In DevTools, click Stop to stop the recording
7. Look for the yellow mountains of CPU usage that coincide with your clicks:
cpu_mountains
8. Hover your mouse and look at the screenshots. You should see the checkbox becoming checked close to the right side of the yellow mountain.
devtools_screenshot
9. Click and _drag_ to select the mountain, and a bit before and after it. You should see something like this:
devtools_dragged
Note the red tag in the corner; this indicates that particular task took too long in Chrome's opinion. 
10. Scroll down (not with your mouse wheel! Use the scroll bar on the right) until you see something similar to this:
workLoopSync
The tiny items are React rendering the table after your click. Notice the `workLoopSync` function call, each of which contains several `performUnitOfWork` calls. These are the React internal functions doing their work. I find it useful to measure the `workLoopSync` because there's usually only one per click (on this one there are two, so add them up.) In this picture it's 269.87ms, plus the one to the left which was 60-something ms.  (Or you could just measure `performSyncWorkOnRoot` instead)

Using this testing procedure, we now have a repeatable, measurable way to measure click performance of the checkboxes. Repeat this with the change checked out, and hopefully you'll see a big improvement.

More DevTools tips:
* You can use the up and down arrow buttons (Load Profile and Save Profile) to save and load profiles after you've recorded them
* I recommend having DevTools in its own window for this (click the kebab / three dots menu > Dock Side icons)
* You can use the keyboard shortcut Alt+Tilde (Alt+~) to switch between Chrome windows. Similar to Alt+Tab but within the same app.

### DIFF
--- a/webpack/components/Table/TableHooks.js
+++ b/webpack/components/Table/TableHooks.js
@@ -1,4 +1,4 @@
-import { useState, useRef, useEffect, useCallback } from 'react';
+import { useState, useRef, useEffect, useCallback, useMemo } from 'react';
 import { isEmpty } from 'lodash';
 import { useLocation } from 'react-router-dom';
 import { friendlySearchParam } from '../../utils/helpers';
@@ -53,9 +53,9 @@ class ReactConnectedSet extends Set {
 }
 
 export const useSet = (initialArry) => {
-  const [, setToggle] = useState(0);
+  const [, setToggle] = useState(Date.now());
   // needed because mutating a Ref won't cause React to rerender
-  const forceRender = () => setToggle(prev => prev + 1);
+  const forceRender = () => setToggle(Date.now());
   const set = useRef(new ReactConnectedSet(initialArry, forceRender));
   return set.current;
 };
@@ -68,10 +68,13 @@ export const useSelectionSet = ({
 }) => {
   const selectionSet = useSet(initialArry);
   const pageIds = results?.map(result => result[idColumn]) ?? [];
-  const selectableResults = results?.filter(result => isSelectable(result)) ?? [];
-  const selectableIds = new Set(selectableResults.map(result => result[idColumn]));
+  const selectableResults = useMemo(() => results?.filter(result =>
+    isSelectable(result)) ?? [], [results, isSelectable]);
   const selectedResults = useRef({}); // { id: result }
-  const canSelect = id => selectableIds.has(id);
+  const canSelect = useCallback((id) => {
+    const selectableIds = new Set(selectableResults.map(result => result[idColumn]));
+    return selectableIds.has(id);
+  }, [idColumn, selectableResults]);
   const areAllRowsOnPageSelected = () =>
     Number(pageIds?.length) > 0 &&
         pageIds.every(result => selectionSet.has(result) || !canSelect(result));
@@ -108,7 +111,8 @@ export const useSelectionSet = ({
 
   const selectedCount = selectionSet.size;
 
-  const isSelected = id => canSelect(id) && selectionSet.has(id);
+  const isSelected = useCallback(id =>
+    canSelect(id) && selectionSet.has(id), [canSelect, selectionSet]);
 
   return {
     selectOne,
@@ -157,7 +161,7 @@ export const useBulkSelect = ({
   const areAllRowsSelected = () => (selectAllMode && exclusionSet.size === 0) ||
                                    selectOptions.areAllRowsSelected();
 
-  const isSelected = (id) => {
+  const isSelected = useCallback((id) => {
     if (!selectOptions.isSelectable(id)) {
       return false;
     }
@@ -165,7 +169,7 @@ export const useBulkSelect = ({
       return !exclusionSet.has(id);
     }
     return inclusionSet.has(id);
-  };
+  }, [exclusionSet, inclusionSet, selectAllMode, selectOptions]);
 
   const selectPage = () => {
     setSelectAllMode(false);
--- a/webpack/components/extensions/HostDetails/Tabs/ErrataTab/ErrataTab.js
+++ b/webpack/components/extensions/HostDetails/Tabs/ErrataTab/ErrataTab.js
@@ -135,6 +135,14 @@ export const ErrataTab = () => {
     initialSearchQuery: searchParam || '',
   });
 
+  const tdSelect = useCallback((errataId, rowIndex) => ({
+    disable: !isSelectable(errataId),
+    isSelected: isSelected(errataId),
+    onSelect: (event, selected) => selectOne(selected, errataId),
+    rowIndex,
+    variant: 'checkbox',
+  }), [isSelectable, isSelected, selectOne]);
+
   if (!hostId) return <Skeleton />;
 
   const applyErratumViaRemoteExecution = id => dispatch(installErrata({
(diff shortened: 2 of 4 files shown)


webpack/components/Table/TableHooks.js
```diff
@@ -53,9 +53,9 @@ class ReactConnectedSet extends Set {
 }
 
 export const useSet = (initialArry) => {
-  const [, setToggle] = useState(0);
+  const [, setToggle] = useState(Date.now());
   // needed because mutating a Ref won't cause React to rerender
-  const forceRender = () => setToggle(prev => prev + 1);
+  const forceRender = () => setToggle(Date.now());
```
jeremylenz, line 58: It helped a little bit to not create a function every time

webpack/components/extensions/HostDetails/Tabs/ErrataTab/ErrataTab.js
```diff
@@ -452,16 +453,20 @@ export const ErrataTab = () => {
...
-                      <ExpandableRowContent>
-                        <ErratumExpansionContents erratum={erratum} />
-                      </ExpandableRowContent>
-                    </Td>
-                    <Td colSpan={4}>
-                      <ExpandableRowContent>
-                        <ErratumExpansionDetail erratum={erratum} />
-                      </ExpandableRowContent>
-                    </Td>
+                    {isExpanded && (
```
jeremylenz, line 456: The conditional rendering here was by far the most helpful change of all. Not having to render invisible table rows just makes sense :smile:

jturel:
I've been waiting to see the solution for this one. Elegant!

webpack/components/extensions/HostDetails/Tabs/ErrataTab/ErrataTab.js
```diff
@@ -135,6 +135,14 @@ export const ErrataTab = () => {
...
   });
 
+  const tdSelect = useCallback((errataId, rowIndex) => ({
+    disable: !isSelectable(errataId),
+    isSelected: isSelected(errataId),
+    onSelect: (event, selected) => selectOne(selected, errataId),
+    rowIndex,
+    variant: 'checkbox',
+  }), [isSelectable, isSelected, selectOne]);
+
```
Andrewgdewar, line 145: We likely need a useCallback like this in all places that are using the useSelectionSet hook no? or does your useCallback on the isSelected method handling that issue by itself. 
Would this change then be needed? or is this for exhaustive deps only?

jeremylenz: Tbh I don't think the callback does much here- I can test it and see what the improvement is. I kept it mainly because of your preference for putting that object in a function rather than having an object literal down in the JSX..

webpack/components/extensions/HostDetails/Tabs/ErrataTab/ErrataTab.js
Andrewgdewar: This is good, and we should probably do that [here](https://github.com/Katello/katello/blob/master/webpack/scenes/ContentViews/Table/ContentViewsTable.js#L241) 
And any other places we use expandable content.

webpack/components/extensions/HostDetails/Tabs/ErrataTab/ErrataTab.js
jeremylenz: I was fully ready to delete this and revert back to the object in JSX, but then I tested it..

Tested with `tdSelect useCallback`: 174ms 178ms 171ms, average 174.33ms
With object literal in JSX: 202ms 185ms 200ms, average 195.67ms

Adds 21ms on average.

Also I feel like I can _feel_ the difference a bit as well!  For that reason I'm for keeping the useCallback. thoughts?

(Fyi, PF's doc for it is here: https://www.patternfly.org/v4/components/table/#tdselecttype)

webpack/components/extensions/HostDetails/Tabs/ErrataTab/ErrataTab.js
jeremylenz: Agree. But I feel like only tables that use `useBulkSelect` are in scope of this PR, right?

Andrewgdewar (approved):
This does fix some of the performance issues seen in the useBulkSelect hook! 
Users that manually set high per_page values (of above 100 or more) will likely still see some performance issues on their checkbox selection. 
As that is likely a more rarely seen use case, ACK!

webpack/components/extensions/HostDetails/Tabs/ErrataTab/ErrataTab.js
Andrewgdewar: You can see performance issues for both.  If the fix applied here is only affecting useBulkSelect's behaviour than that's totally reasonable.

jeremylenz:
[test katello]

PR #10005, opened 2022-03-09
Fixes #34582 - Add policy streamed for smart-proxy

m-bucher:
Adds SmartProxy download_policy `streamed`.
See https://docs.pulpproject.org/pulpcore/workflows/on-demand-downloading.html#overview

### DIFF
--- a/app/helpers/katello/concerns/smart_proxy_helper_extensions.rb
+++ b/app/helpers/katello/concerns/smart_proxy_helper_extensions.rb
@@ -36,6 +36,10 @@ module Katello
             :label => ::Katello::RootRepository::DOWNLOAD_IMMEDIATE
           },
           {
+            :name => _("Streamed"),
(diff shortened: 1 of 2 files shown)


chris1984:
@m-bucher what are the testing steps for this?

quba42:
@chris1984 On the smart proxy, select "Edit" (top right), then on the "Smart Proxy" tab select "Streamed" for the "Download Policy" field. Submit, and synchronize content to the smart proxy. To verify that streamed was actually used, you can check the Pulp instance on the proxy, or else request the same content from the proxy multiple time while checking the Pulp logging on the main server to make sure it is downloaded from there each time it is requested.

chris1984 (approved):
ACK, works great

```bash
Smart proxy is synchronized
Last sync: 2022-03-21 12:12:16 -0400
```

Thanks @m-bucher

PR #10006, opened 2022-03-09
Fixes #34576 - Additional steps before deleting repo from CVs

sjha4:
Force deletion of repo should have additional layers of security when deleting repos from cv versions.
This PR adds an additional checkbox on the modal to confirm delete, and a setting which can be turned off to disallow force deletions of repositories.

1. Create a repo.
2. Add it to CVs and publish a few versions of those.
3. Go to Repo, action dropdown > Delete
4. You should see the modal with a checkbox.
5. Check it. Delete button should get enabled. Don't click it yet! Close the modal.
6. Go to settings > content > Allow deleting repositories in published content views : Set to No
7. Go back to same repo and Actions > Delete should not be an option. Instead you should see a Cannot remove with a help text.
8. Go to settings > content > Allow deleting repositories in published content views : Set to Yes
9. Go back to repository.
10. Delete.

### DIFF
--- a/engines/bastion/app/assets/javascripts/bastion/components/bst-modal.directive.js
+++ b/engines/bastion/app/assets/javascripts/bastion/components/bst-modal.directive.js
@@ -59,6 +59,7 @@ angular.module('Bastion.components').directive('bstModal',
 
                     modalInstance.result.then(function () {
                         scope.action();
+                    }, function() {
                     });
                 };
 
--- a/app/models/katello/authorization/repository.rb
+++ b/app/models/katello/authorization/repository.rb
@@ -4,7 +4,7 @@ module Katello
 
     delegate :editable?, to: :product
 
-    def deletable?(remove_from_content_view_versions = false)
+    def deletable?(remove_from_content_view_versions = true)
       product.editable? && (remove_from_content_view_versions || !promoted? || !self.content_views.generated_for_none.exists?)
     end
 
--- a/app/models/setting/content.rb
+++ b/app/models/setting/content.rb
@@ -158,7 +158,10 @@ class Setting::Content < Setting
                N_('Upload profiles without Dynflow')),
       self.set('orphan_protection_time', N_('Time in minutes to consider orphan content as orphaned.'), 1440, N_('Orphaned Content Protection Time')),
       self.set('remote_execution_prefer_registered_through_proxy', N_('Prefer using a proxy to which a host is registered when using remote execution'), false,
-               N_('Prefer registered through proxy for remote execution'))
+               N_('Prefer registered through proxy for remote execution')),
+      self.set('delete_repo_across_cv', N_("If this is enabled, repositories can be deleted even when they belong to published "\
+                 "content views. The deleted repository will be removed from all content view versions."),
+               true, N_('Allow deleting repositories in published content views'))
     ]
   end
 
--- a/app/views/katello/api/v2/repositories/show.json.rabl
+++ b/app/views/katello/api/v2/repositories/show.json.rabl
@@ -70,7 +70,8 @@ end
 
 node :permissions do |repo|
   {
-    :deletable => repo.deletable?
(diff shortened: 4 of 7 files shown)


ianballou:
I'm trying to think where a user permission might be better than a setting.  I'm imagining this:

A sysadmin doesn't want the people with repo delete permissions to be force deleting repos from content views.  However, they like to use the new feature because they know what they're doing.  So, they keep the setting off, and then go flip it whenever they want to force delete a repo.  There is a chance however that an unauthorized user could then force delete a repo because the setting is global. Plus, the sysadmin could forget to flip it back. So, it ends up being annoying for the sysadmin and potentially doesn't suit their authorization needs.

If there's a user permission that starts off as admin only, then it's sure that only the sysadmin can force delete repos.  And, if there are other trusted users, they could do it too.

sjha4:
> I'm trying to think where a user permission might be better than a setting. I'm imagining this:
> 
> A sysadmin doesn't want the people with repo delete permissions to be force deleting repos from content views. However, they like to use the new feature because they know what they're doing. So, they keep the setting off, and then go flip it whenever they want to force delete a repo. There is a chance however that an unauthorized user could then force delete a repo because the setting is global. Plus, the sysadmin could forget to flip it back. So, it ends up being annoying for the sysadmin and potentially doesn't suit their authorization needs.
> 
> If there's a user permission that starts off as admin only, then it's sure that only the sysadmin can force delete repos. And, if there are other trusted users, they could do it too.

So I discussed the permissions vs settings approach with Sayan and Justin. Thought here is that delete_repo permission should be sufficient to manage permissions for deleting repos cause that is intuitive. A new permission would be need to be added to existing users who have already been granted delete_repo permissions. The setting is global and on by default making this feature immediately available for use without any actions on the part of the admin. Assumption is that admins and non-admins with delete permissions on repo will be alerted enough with the additional checkbox that needs a conscious decision on their part to delete it across CVs..

**TLDR: A new permission would be overly restrictive and require work to enable this feature. A setting which is on by default makes it immediately available for all users with delete permissions while providing flexibility to turn it off.**

ianballou:
> So I discussed the permissions vs settings approach with Sayan and Justin. Thought here is that delete_repo permission should be sufficient to manage permissions for deleting repos cause that is intuitive. A new permission would need to be added to existing users who have already been granted delete_repo permissions. The setting is global and on by default making this feature immediately available for use without any actions on the part of the admin. Assumption is that admins and non-admins with delete permissions on repo will be alerted enough with the additional checkbox that needs a conscious decision on their part to delete it across CVs..
> 
> **TLDR: A new permission would be overly restrictive and require work to enable this feature. A setting which is on by default makes it immediately available for all users with delete permissions while providing flexibility to turn it off.**

I gotcha, so we're not digging too much into the aspect of limiting who can force delete repos. That's fine then since it's on by default.

engines/bastion/app/assets/javascripts/bastion/components/bst-modal.directive.js
```diff
@@ -59,6 +59,7 @@ angular.module('Bastion.components').directive('bstModal',
 
                     modalInstance.result.then(function () {
                         scope.action();
+                    }, function() {
```
ianballou, line 62: Why'd you need to add this in?

sjha4: Uggh.. 😬  Ya, the bastion library somehow was suppressing a weird console warning on modal close. When I overrode the modal footer for repo deletion to add the customization with checkbox and button disabling, it started appearing. 
This is to suppress that error back. More details here: https://github.com/angular-ui/bootstrap/issues/6501#issuecomment-292881004

ianballou: Oh weird! Okay, makes sense then.

PR #10007, opened 2022-03-10
Fixes #34586 - Add autocomplete for rpm and container rule modals

rverdile:
Adds autocomplete for the Name and Architecture fields when adding/editing a CV filter rpm rule.  Adds autocomplete for name field when adding CV filter Container Tag rule.
The search component (and related components) had 99% of what was needed to do this, so I tried to adapt it for text inputs instead of search bars.
1. Create yum and docker repositories with content.
1. In UI, Content -> Content View
2. Select any CV -> Go to "Filters" tab
3. Create "Create filter" -> Type: RPM -> Add RPM rule
4. Click in the Name box, should see autocomplete pop up for the names of the packages.
5. Click the architecture box, should see autocomplete.
6. Same thing when editing the rule.
7. Test similarly for Container Tag filter.

### DIFF
--- a/webpack/components/Search/Search.js
+++ b/webpack/components/Search/Search.js
@@ -23,6 +23,8 @@ const Search = ({
   foremanApiAutoComplete,
   bookmarkController,
   placeholder,
+  isTextInput,
+  setTextInputValue,
 }) => {
   const [items, setItems] = useState([]);
   const dispatch = useDispatch();
@@ -48,9 +50,16 @@ const Search = ({
       }
       // Checking whether the current component is mounted before state change events
       if (!mountedRef.current) return;
-      setItems(data?.data?.filter(({ error }) => !error).map(({ label }) => ({
-        text: label.trim(),
-      })));
+      switch (true) {
+      case endpoint.includes('auto_complete_arch'):
+      case endpoint.includes('auto_complete_name'):
+        setItems(data?.data?.map(label => ({ text: label.trim() })));
+        break;
+      default:
+        setItems(data?.data?.filter(({ error }) => !error).map(({ label }) => ({
+          text: label.trim(),
+        })));
+      }
     }
 
     if (autoSearchEnabled && patternfly4) {
@@ -83,6 +92,8 @@ const Search = ({
         patternfly4={patternfly4}
         autoSearchEnabled={autoSearchEnabled}
         placeholder={placeholder}
+        isTextInput={isTextInput}
+        setTextInputValue={setTextInputValue}
       />
     </div>
   );
@@ -102,6 +113,8 @@ Search.propTypes = {
   }),
   bookmarkController: PropTypes.string,
   placeholder: PropTypes.string,
+  isTextInput: PropTypes.bool,
+  setTextInputValue: PropTypes.func,
 };
 
 Search.defaultProps = {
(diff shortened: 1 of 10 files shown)


webpack/scenes/ContentViews/Details/Filters/Rules/ContainerTag/AddEditContainerTagRuleModal.js
```diff
@@ -2,17 +2,22 @@ import React, { useState } from 'react';
...
 const AddEditContainerTagRuleModal = ({
-  onClose, filterId, selectedFilterRuleData,
+  onClose, filterId, selectedFilterRuleData, repositoryIds,
 }) => {
   const { name, id } = selectedFilterRuleData;
   const dispatch = useDispatch();
   const [tagName, setTagName] = useState(name);
   const [saving, setSaving] = useState(false);
   const isEditing = name && id;
+  const [, updateSearchQuery] = useState('');
```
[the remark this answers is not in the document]
Andrewgdewar: Is it to rerender the component?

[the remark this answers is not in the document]
rverdile: It looks like this is unnecessary. I will remove it.

webpack/components/Search/Search.js
```diff
@@ -48,9 +50,13 @@ const Search = ({
       }
       // Checking whether the current component is mounted before state change events
       if (!mountedRef.current) return;
-      setItems(data?.data?.filter(({ error }) => !error).map(({ label }) => ({
-        text: label.trim(),
-      })));
+      if (endpoint.includes('auto_complete_arch') || endpoint.includes('auto_complete_name')) {
```
Andrewgdewar, line 53: We should likely set this up as a switch for future use (preference) 

```
switch(true){
case endpoint.includes('auto_complete_arch'): 
case endpoint.includes('auto_complete_name'): 
  setItems(data?.data?.map(label => ({ text: label.trim() })));
default: 
 setItems(data?.data?.filter(({ error }) => !error).map(({ label }) => ({
          text: label.trim(),
        })));
```

rverdile: I'm okay with that. I'll change this.

webpack/components/TypeAhead/pf4Search/TypeAheadInput.js
```diff
@@ -31,19 +31,19 @@ const TypeAheadInput = ({
         onFocus={onInputFocus}
         aria-label="text input for search"
         onChange={onChangeWrapper}
-        type="search"
-        iconVariant={autoSearchEnabled && 'search'}
+        type="text"
+        iconVariant={(autoSearchEnabled && !isTextInput) && 'search'}
```
Andrewgdewar, line 35: This can produce a case where false is returned. 
(Which may not cause any harm, but not the best standard).
We should likely be explicit and use a ternary operator
example `autoSearchEnabled && !isTextInput ? 'search' : undefined`

Andrewgdewar (approved):

rverdile:
[test katello]

sjha4:
Tested and works great. 🥳 🎉

PR #10008, opened 2022-03-11
Fixes #34429 - properly return gpg key in repos json

jlsherrill:
it seems like :child doesn't work within a glue() block in rabl, i'm unsure why, but other content credentials are done this way.

1.  create a content credential gpg key
2. create a repository, set the gpg key on the repository
3. visit the repo details page, verify that you can see the gpg key name in the ui


ianballou (approved):
Works fine!

PR #10010, opened 2022-03-14
Fixes #34591 - migrate deb-repo attributes

m-bucher:
For APT repositories this fix makes sure that:

1. `download_policy` is set (default: immediate): Introduced in Katello-4.3 (https://github.com/Katello/katello/commit/53a31d6b8aaf4ba3a296e31b1dda7ee125b383ad)
2. `releases` is set (default: stable): requirement of pulp3; was implicitly assumed as stable by pulp2
3. `releases`, `architectures`, `components` are space-separated lists: pulp2 used comma-separation

We tried to do this as an upgrades-task but that failed due to dynflow not being available at the time they run during foreman-installer


ianballou (approved):
This is working fine for me! We'll want to add this to the upstream docs to run during upgrade.

PR #10011, opened 2022-03-14
Fixes #34608 - use only public setting interface

ezr-ondrej:
Use only public setting interface to manipulate settings.

### DIFF
--- a/lib/katello/tasks/update_content_default_http_proxy.rake
+++ b/lib/katello/tasks/update_content_default_http_proxy.rake
@@ -3,7 +3,6 @@ require File.expand_path("../engine", File.dirname(__FILE__))
 namespace :katello do
   desc "Sets the content default HTTP proxy to an existing HTTP proxy based on supplied URL."
   task :update_default_http_proxy => :environment do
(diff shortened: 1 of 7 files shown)


ezr-ondrej:
[test katello]

jeremylenz:
Test failure seems related:

```
Failure:

[2022-03-16T22:35:54.878Z] Katello::SettingTest#test_foreman_proxy_content_auto_sync_setting [/home/jenkins/workspace/katello-pr-test/test/models/setting_test.rb:14]

[2022-03-16T22:35:54.878Z] Minitest::Assertion: Expected true to not be truthy.
```

ezr-ondrej:
> Test failure seems related:

It is, I tried to fix, but it is a pointless test for the new DSL as we have static typing there, so I've decided just to skip in hope Katello will go with the DSL soon, which this helps to unblock :)

jeremylenz (approved):
Thanks @ezr-ondrej!

will make a note to fix that test when we finally tackle https://projects.theforeman.org/issues/33608.

PR #10013, opened 2022-03-14
Fixes #34609 - Entitlement report double counts consumed entitlements

ianballou:
Adds a method to grab the number of entitlements that a host in consuming for a particular pool.  Intended to be tested in conjunction with the report template change over in the Foreman repo: https://github.com/theforeman/foreman/pull/9151

The main issue is that `pool.consumed` is a global value, not per-host.  The report template is generating values per-host, so it makes much more sense to have this value be per-host as well.

It doesn't appear that we store this entitlement "quantity" value since it's also fetched from the Candlepin API for a host's add/list subscriptions page.

1) Register two content hosts.
2) Attach some subscriptions to both the hosts. Make sure that the hosts get some overlapping subscriptions.
3) Generate the Subscription - Entitlement Report.
4) See that the "Consumed" amount is both the same for all hosts and incorrect (too high and is actually the count across all hosts).
5) Apply this PR and the Foreman PR.
6) `bundle exec rake db:seed`
7) Ensure the report template changed and regenerate it. 
8) Check that the consumed amount is now correctly reflected for each host.

### DIFF
--- a/app/models/katello/concerns/host_managed_extensions.rb
+++ b/app/models/katello/concerns/host_managed_extensions.rb
@@ -424,6 +424,12 @@ module Katello
         ::Katello::Erratum.installable_for_hosts([self]).search_for(search).pluck(:errata_id)
       end
 
+      def filtered_entitlement_quantity_consumed(pool)
+        entitlements = subscription_facet.candlepin_consumer.filter_entitlements(pool.cp_id)
+        return nil if entitlements.empty?
+        entitlements.sum { |e| e[:quantity] }
+      end
+
       protected
 
       def update_trace_status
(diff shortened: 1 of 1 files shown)


app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -424,6 +424,12 @@ def advisory_ids(search:)
         ::Katello::Erratum.installable_for_hosts([self]).search_for(search).pluck(:errata_id)
       end
 
+      def filtered_entitlement_quantity_consumed(pool)
+        entitlement = subscription_facet.candlepin_consumer.filter_entitlements(pool.cp_id)&.first
```
rverdile, line 428: Is it safe to use `first` here? What if there's more than 1 entitlement returned? I'm not sure how to manually test that scenario, but I see a test with it: https://github.com/Katello/katello/blob/master/test/services/katello/candlepin/consumer_test.rb#L27-L36

rverdile (commented):
this works, but I have one question that I left in my other comment. There's also a suggestion from jeremy on your foreman PR.

rverdile (approved):
I'm gonna ack this since it seems okay. Let me know if you end up making changes.

PR #10014, opened 2022-03-14
Vscode changes

Andrewgdewar:


.nvmrc
```diff
@@ -0,0 +1 @@
+v12.22.7
```
sjha4, line 1: We don't want this.Do we?

PR #10018, opened 2022-03-15
Fixes #34619 - Remotes should have username and password cleared out if set blank

ianballou:
Ensures that Pulp remotes have their usernames and passwords cleared out if the user sets them to be empty.

This thread spawned the need for this fix: https://community.theforeman.org/t/pulprpmclient-apierror-http-500-during-sync-repository/27480/24

It's not confirmed, but it should fix the problem that folks are having.

1) Update a repo's username and password
2) Check the username and password in from the DB:
```
vagrant@centos8-katello-devel ~/katello $ sudo -i -u postgres
postgres@centos8-katello-devel ~ $ psql
psql (12.9)
Type "help" for help.

postgres=# \c pulpcore 
You are now connected to database "pulpcore" as user "postgres".
pulpcore=#  select pulp_id, name, url, username, password, length(username) from core_remote;
...
```
3) Clear out the username and password on the repo
4) Sync the repo
5) The username and password should now be empty in postgres


sjha4 (approved):
Ack 👍🏼

PR #10019, opened 2022-03-15
Fixes #34620 - Show toast linking to REX job

jeremylenz:

rex_toast

* Add `renderRexJobStartedToast` to go alongside `renderTaskStartedToast`. This will link directly to the Remote execution job details page, not just to its wrapper Foreman Task.
* Change PackageTab install button to primary
* also some other light refactoring

_BONUS REFACTOR:_ Now that we've updated Patternfly, `toggleVariant="primary"` actually works now. So it allows us to have the PackagesTab 'Upgrade' button turn blue when enabled, as intended in the original mockup.
packages-upgrade

Perform REX actions on the new host detail page
Verify that toast comes up and links to job invocation detail page, not Foreman Tasks page
Also see if you can test an API error and verify that it shows up properly as well

### DIFF
--- a/webpack/components/extensions/HostDetails/Tabs/PackagesTab/PackageInstallModal.js
+++ b/webpack/components/extensions/HostDetails/Tabs/PackagesTab/PackageInstallModal.js
@@ -64,7 +64,6 @@ const InstallDropdown = ({
       onSelect={onActionSelect}
       toggle={
         <DropdownToggle
-          isPrimary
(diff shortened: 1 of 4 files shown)


chris1984:
ACK, works good @jeremylenz 

I like the new blue button. Toast works fine and I can see errors if I call them.

Web capture_28-3-2022_104144_192 168 121 207

chris1984 (approved):

PR #10020, opened 2022-03-16
Fixes #34626 - Default to Content only mirroring policy on the UI

sjha4:
The webUI default for mirroring policy is Additive while the controller default is Content only. The correct mirroring policy to default to would be Content only on the UI as well.

1. Create a new repository.
2. Select type: yum and scroll down to check the default value in the mirroring policy field.
3. It should default to "Content only"


sjha4:
[test katello]

ianballou (approved):
Works for me!

PR #10021, opened 2022-03-16
Fixes #34628 - Add TranslatedPlural component

jeremylenz:

Adds two components:
* `<TranslatedPlural>` - Renders a `<FormattedMessage>` with proper translation and pluralization. So we can display messages like this

errata_singular

errata_zero

You can also customize what it says when there are zero items.

* `<TranslatedAnchor>` which renders an `<a>` with a `<TranslatedPlural>` inside, with optional styling.

see my inline review below

Get a host with one erratum and view the ErrataOverviewCard.
Do the same with a host with no errata, and with several errata.

Or, find the `<HostInstallableErrata>` component in React Dev Tools, and change the `errataCounts` prop!

### DIFF
--- a/webpack/components/Errata/index.js
+++ b/webpack/components/Errata/index.js
@@ -37,6 +37,7 @@ export const ErrataSummary = ({ type, count }) => {
         plural="security advisories"
         singular="security advisory"
         zeroMsg="# security advisories"
+        ariaLabel={`${count} security advisories`}
       />
     );
     break;
@@ -45,7 +46,6 @@ export const ErrataSummary = ({ type, count }) => {
     label = __('Bugfix');
     ErrataIcon = BugIcon;
     color = '#8bc1f7';
-    // url = <a href="#/Content/errata?type=bugfix"> {count} {plural} </a>;
     url = (
       <TranslatedAnchor
         id="errata-card-bugfix-count"
@@ -55,6 +55,7 @@ export const ErrataSummary = ({ type, count }) => {
         plural="bug fixes"
         singular="bug fix"
         zeroMsg="# bug fixes"
+        ariaLabel={`${count} bug fixes`}
       />
     );
     break;
@@ -63,7 +64,6 @@ export const ErrataSummary = ({ type, count }) => {
     label = __('Enhancement');
     ErrataIcon = EnhancementIcon;
     color = '#002f5d';
-    // url = <a href="#/Content/errata?type=enhancement"> {count} {plural} </a>;
     url = (
       <TranslatedAnchor
         id="errata-card-enhancement-count"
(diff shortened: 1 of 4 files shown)


webpack/components/Errata/index.js
```diff
@@ -13,39 +14,68 @@ import {
...
   case 'security':
     label = __('Security');
     ErrataIcon = SecurityIcon;
-    name = 'security advisories';
     color = '#0066cc';
-    url = <a href="#/Content/errata?type=security"> {count} {name} </a>;
+    url = (
+      <TranslatedAnchor
+        id="errata-card-security-count"
+        style={{ marginLeft: '0.4rem' }}
```
jeremylenz, line 34: We were using space characters for spacing, which is never ideal. So I added the `style` prop and passed it thru to the `<a>`.

webpack/components/Errata/index.js
```diff
@@ -13,39 +14,68 @@ import {
...
+        href="#/Content/errata?type=security"
+        count={count}
+        plural="security advisories"
+        singular="security advisory"
+        zeroMsg="# security advisories"
+        ariaLabel={`${count} security advisories`}
```
jeremylenz, line 40: React Testing Library won't render the text properly now that it's translated:

```html
<a
      aria-label="0 enhancements"
      href="#/Content/errata?type=enhancement"
      style="margin-left: 0.4rem;"
 >
 {count, plural, =0 {# enhancements} one {# enhancement} other {# enhancements}}
</a>
```
(Looks fine on screen, though.)

Because of this, I had to find another way to locate these elements. I came up with adding an `ariaLabel` prop here which is passed to the `<a>` tag as the `aria-label` HTML attribute. This allows finding the element easily with `getByLabelText`.

webpack/components/Table/components/TranslatedPlural.js
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import { FormattedMessage } from 'react-intl';
+import PropTypes from 'prop-types';
+
+// a defaultMessage like
+// "{count, plural, =0 {No errata} one {# erratum} other {# errata}}"
+// will give us properly translated pluralized strings!
+// see https://formatjs.io/docs/react-intl/components/#message-syntax
+export const TranslatedPlural = ({
```
lfu, line 9: I'm curious why we don't use [pluralize](https://apidock.com/rails/ActionView/Helpers/TextHelper/pluralize) method directly.

jeremylenz: That's a back-end method; it wouldn't help us here..

lfu (approved):
LGTM 👍🏻

jeremylenz:
rebased just to re-kick React tests :)

PR #10023, opened 2022-03-16
Fixes #34633 - Add Epoch to CV version RPM packages page

chris1984:

Nothing major, Epoch was already being returned in the results, so just made it visible in the table.

Nothing major, Epoch was already being returned in the results, so just made it visible in the table.

* Apply patch
* Sync a repo with RPMS
* Create a content view and publish it
* View the version tab and select RPM packages
* Verify that the Epoch column is present in the table

See screenshot for example:

Web capture_16-3-2022_141115_192 168 121 207


lfu (approved):
LGTM 👍🏻

PR #10026, opened 2022-03-17
Fixes #34638 - Fail if errata search query matches nothing

adamruzicka:

Install errata by search query job template now fails during rendering if a search query is provided, but does not match any errata at all.

The broken part was that if user provided a filter which matched nothing, the final template just had `yum update-minimal` which installed all available errata. If someone provides a filter then it should only act on what the filter matches. 

1) Run a job using the Install errata by search query template
2) Provide a bogus query which will match no errata


lfu:
**Before** 

In case of no matching errata, all errata were updated on the target.
```
[vagrant@rhel8 ~]$ sudo yum history
Updating Subscription Management repositories.
ID   | Command line                            | Date and time      | Action(s)      | Altered
---------------------------------------------------------------------------------------------------------
  38 | -y update-minimal                       | 2022-03-25 19:28   | Upgrade        |    3   
```
**After**
In case of no matching errata, there is nothing got updated on the target.
Screenshot from 2022-03-25 16-10-59

lfu (approved):

PR #10031, opened 2022-03-23
Fixes #33591 - Wrong kickstart repo shown in Edit Host page

ianballou:

The `kickstart_repository_options` method made the assumption that the host's content facet would have the same attributes as the host group.  If this was false, the wrong attributes would be assigned to the host's content facet and `new_host.operatingsystem.kickstart_repos(new_host)` would return `[]`.   This was caused by the `distribution_repositories` method seeing a mismatch between the host's OS and the host's LCE and CVE, causing no kickstart repositories to show up.

Now, there is a check to see if the host has a different LCE, CV, or content source from the host group.

None really, it's pretty straight-forward.

1) Follow the instructions on the redmine to check the main bug
2) Try playing around with the host group create + edit pages and the host create + edit pages and ensure everything makes sense.

### DIFF
--- a/app/helpers/katello/hosts_and_hostgroups_helper.rb
+++ b/app/helpers/katello/hosts_and_hostgroups_helper.rb
@@ -205,9 +205,7 @@ module Katello
         return [] unless new_host.operatingsystem.is_a?(Redhat)
 
         if (host.is_a? ::Hostgroup)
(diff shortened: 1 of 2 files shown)


rverdile (approved):
lgtm! although something weird is going on with the CI.

ianballou:
Looks like the tests all passed but the checks are still having problems.

ianballou:
[test katello]

PR #10032, opened 2022-03-24
Fixes #34676 - Incremental CV update does not auto-publish CCV

ianballou:
"Auto publish" was not considered during incremental update.  This is fixed by adding a check for `auto_publish?` where components are added for `propagate_all_composites`.

None really, it's straight forward.

1) Incremental update a component CV that's within a composite that has auto publish on
  -> Ensure that the composite gets a new point-release version
2) Try the same thing using propagate-all-composites
3) Ensure that inc update still works fine in the UI


ianballou:
[test katello]

sjha4:
Works well. 1 thing I'm seeing (and am not sure if it's expected) is that the new version of the component CV itself doesn't get promoted to Library on incremental update whereas the composite CV minor version does.

ianballou:
@sjha4 I think that should be fine, that's how `--propagate-all-composites` worked in the past.

ianballou:
Actually, I'm starting to wonder if I should rethink the code. Right now I'm making auto publish do the same thing as `--propagate-all-composites`.  This isn't quite the same as auto publish though, since a new composite version will only be created if the component version that's incrementally updated is the same one that's in the composite content view version.  With true auto publish, there will be a new composite content view version every time.

ianballou:
I'm going to look into changing up the strategy so that the composite is actually auto published.  That closer fits the BZ request.  I also asked for clarification on the BZ to double check the desired fix.

ianballou:
I changed up the strategy a bit.  Now, if a component is incrementally updated and it generates a new latest version, then the composites will be auto published. This will happen in tandem with `--propagate-all-composites`.

So, to test:
1) Ensure that a composite content view always auto publishes to match the latest component versions
  -> Via component publish or incremental update
2) Ensure that `--propagate-all-components` still functions like before.  The only difference will be that the composite gets two new versions: the incremental and the latest (if the inc update created a new 'latest' version in the component)
  -> In this case, library will always get assigned to the very latest composite version
  -> Note that, to propagate inc update to composites, the composite version to be updated needs to be in an LCE.

ianballou:
This issue might be a contender for a BATS test.

sjha4:
I am testing this but I'm having a tough time with publishes after an incremental update even on master where Setting[:bulk_load_size] randomly returns nil and the whole task fails. Can you reproduce this or is it my dev box only (which is very likely)?
 
```
Error:

TypeError

nil can't be coerced into Integer

---
- "/home/vagrant/katello/app/services/katello/pulp3/pulp_content_unit.rb:98:in `+'"
- "/home/vagrant/katello/app/services/katello/pulp3/pulp_content_unit.rb:98:in `block
  (2 levels) in pulp_units_batch_for_repo'"
- "/home/vagrant/katello/app/services/katello/pulp3/pulp_content_unit.rb:85:in `loop'"
- "/home/vagrant/katello/app/services/katello/pulp3/pulp_content_unit.rb:85:in `block
  in pulp_units_batch_for_repo'"
- "/home/vagrant/katello/app/services/katello/content_unit_indexer.rb:27:in `each'"
- "/home/vagrant/katello/app/services/katello/content_unit_indexer.rb:27:in `each'"
- "/home/vagrant/katello/app/services/katello/content_unit_indexer.rb:27:in `import_all'"
- "/home/vagrant/katello/app/models/katello/repository.rb:921:in `block (2 levels)
  in index_content'"
```

ianballou:
@sjha4 weird! I'll take a look into it.

sjha4 (approved):
Works as defined in comment above and change looks good to me✅ 
Ack..Squash commits..

One note:
Still seeing warnings like this:
`11:35:48 rails.1   | 2022-04-07T11:35:48 [W|app|23dd59ea] Setting bulk_load_size has no definition, please define it before using`
causing some actions to fail randomly. I think it might be happening because of updates to Settings: [deprecate non-DSL setting definitions](https://github.com/theforeman/foreman/pull/9150#) but I have to test that theory independently of this PR.

PR #10035, opened 2022-03-25
Fixes #34383 - No hammer output for composite incremental update

rverdile:
Adds "Added Content" output for composite content view when incremental updating CV with `--propagate-all-composites true`.
Also removed some code that was meant for pulp 2.
Compared the output to the 6.9 output. This bug was caused mainly by the "Incremental Update" task returning before `plan_self` for composites. This meant there was no task output, unlike with component content views, which is needed for the presenter to print the results.

The original issue also mentions not seeing the new content in the composite content view, but I could not reproduce that.
1. Create a content view with a yum repository, e.g. https://fixtures.pulpproject.org/rpm-with-modules/
2. Add filters to the content view to exclude some packages and/or errata.
3. Publish the content view
4. Create a composite content view, containing the content view and content view version you just published (_not_ "always latest version").
5. Publish the content view
6. Incrementally update the component content view with hammer, i.e. `hammer content-view version incremental-update --propagate-all-composite true --content-view-version-id [id] --package-ids [package ids]`
    - Without PR: Only shows "Added Content" output for component content view.
    - With PR: Shows "Added Content" output for component _and_ composite content view.
8. Check that a new incremental version was successfully published/promoted for the composite content view.
7. Incrementally update again and you should see the "Added Content" output for _only_ the component content view.

### DIFF
--- a/app/lib/actions/katello/content_view_version/incremental_update.rb
+++ b/app/lib/actions/katello/content_view_version/incremental_update.rb
@@ -3,7 +3,7 @@ module Actions
     module ContentViewVersion
       class IncrementalUpdate < Actions::EntryAction
         include ::Katello::ContentViewHelper
-        attr_accessor :new_content_view_version
+        attr_accessor :new_content_view_version, :new_content_view_version_id
 
         HUMANIZED_TYPES = {
           ::Katello::Erratum::CONTENT_TYPE => "Errata",
@@ -18,6 +18,7 @@ module Actions
         # rubocop:disable Metrics/MethodLength
         # rubocop:disable Metrics/AbcSize
         # rubocop:disable Metrics/CyclomaticComplexity
+        # rubocop:disable Metrics/PerceivedComplexity
         def plan(old_version, environments, options = {})
           dep_solve = options.fetch(:resolve_dependencies, true)
           description = options.fetch(:description, '')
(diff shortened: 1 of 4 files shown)


app/lib/actions/katello/content_view_version/incremental_update.rb
```diff
@@ -213,44 +222,23 @@ def run # rubocop:disable Metrics/CyclomaticComplexity
           base_repos = ::Katello::ContentViewVersion.find(input[:old_version]).repositories
           new_repos = ::Katello::ContentViewVersion.find(input[:new_content_view_version_id]).repositories
 
-          if input[:copy_action_outputs].present? && input[:copy_action_outputs].last[:pulp_tasks].present?
-            if input[:copy_action_outputs].last[:pulp_tasks].last[:pulp_href]&.include?("/pulp/api/v3/")
-              new_repos.each do |new_repo|
-                matched_old_repo = base_repos.where(root_id: new_repo.root_id).first
```
ianballou, line 219: I hit an issue:

```
 Error:

NoMethodError

undefined method `errata' for nil:NilClass

---
- "/home/vagrant/katello/app/lib/actions/katello/content_view_version/incremental_update.rb:229:in
  `block in run'"
- "/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activerecord-6.0.4.7/lib/active_record/relation/delegation.rb:88:in
  `each'"
- "/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activerecord-6.0.4.7/lib/active_record/relation/delegation.rb:88:in
  `each'"
- "/home/vagrant/katello/app/lib/actions/katello/content_view_version/incremental_update.rb:226:in
```

From a quick debug I found my `matched_old_repo` was `nil`.  In my case, I was incrementally updating an older content view version that didn't include a repository that I added later.  So, my `new_repo` was defined, but not the old one.

I think these testing steps should reproduce the issue:
1) Have a component CV with one filtered repo
2) Publish the component
3) Publish the composite
4) Add a new repo to the component CV and publish again
5) Publish the composite CV again for good measure
6) Now, promote the old 1-repo composite CV version to Library
7) Incrementally update the old 1-repo component CV version with `--propagate-all-composites` and see the error.

rverdile:
updated

ianballou (approved):
This is working fine for me! I found another bug but it seems unrelated to your PR, so I'll file that separately

PR #10040, opened 2022-03-29
Fixes #34696 - rename ssh to script provider

jlsherrill:

### DIFF
--- a/db/seeds.d/150-module_job_templates.rb
+++ b/db/seeds.d/150-module_job_templates.rb
@@ -1,7 +1,7 @@
 if Katello.with_remote_execution?
   User.as_anonymous_admin do
     JobTemplate.without_auditing do
-      module_template = JobTemplate.find_by(name: 'Module Action - SSH Default')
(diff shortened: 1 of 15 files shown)


jlsherrill:
this will likely fail until the next REX release

db/seeds.d/150-module_job_templates.rb
```diff
@@ -1,7 +1,7 @@
 if Katello.with_remote_execution?
   User.as_anonymous_admin do
     JobTemplate.without_auditing do
-      module_template = JobTemplate.find_by(name: 'Module Action - SSH Default')
+      module_template = JobTemplate.find_by(name: 'Module Action - Script Default')
```
chris1984, line 4: Do we need a migration to change the name as well for existing installs?

adamruzicka: It is in rex

adamruzicka:
Rex 6.1.0 is now out

chris1984 (approved):

chris1984:
ACK

PR #10046, opened 2022-04-04
Fixes #34721 - Use dropdown for Traces & Errata REX actions

chris1984:

* Changed the button to a dropdown with the actions that are in the kebab to make it like we have it on the packages page

* Made sure all tests run and the REX actions function like before. 

* Click the drop-down on the restart app button and make sure the actions work
* Verify there are not any console errors

### DIFF
--- a/app/views/foreman/job_templates/install_errata_by_search_query.erb
+++ b/app/views/foreman/job_templates/install_errata_by_search_query.erb
@@ -20,6 +20,7 @@ foreign_input_sets:
     <%= render_template('Package Action - SSH Default', :action => 'install -n -t patch', :package => advisories) %>
 <% else %>
     <% advisory_ids = @host.advisory_ids(search: input("Errata search query")) %>
+    <% raise "No errata matching given search query" if !input("Errata search query").blank? && advisory_ids.empty? %>
 
     <% advisories = advisory_ids.map { |e| "--advisory=#{e}" }.join(' ') %>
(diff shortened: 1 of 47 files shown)


Andrewgdewar (approved):
Looks good, except for lint!

jeremylenz (changes requested):
Thanks @chris1984! The dropdowns do look much better!  However..

1. When there are no items in the list, the button and dropdown are still enabled. This would be a regression. It should be disabled unless (a) the list contains items; _and_ (b) item(s) are selected.
2. The kebab is still showing the same items as the dropdown. The idea was to move all the REX-related actions to the dropdown, and whatever is left will stay in the kebab, if applicable. This means that for Traces, the kebab should be removed entirely. For Errata, the kebab should contain only 'Recalculate'.
3. For Errata, 'Recalculate' should be in the kebab, not the dropdown.

chris1984:
> Thanks @chris1984! The dropdowns do look much better! However..
> 
> 1. When there are no items in the list, the button and dropdown are still enabled. This would be a regression. It should be disabled unless (a) the list contains items; _and_ (b) item(s) are selected.
> 2. The kebab is still showing the same items as the dropdown. The idea was to move all the REX-related actions to the dropdown, and whatever is left will stay in the kebab, if applicable. This means that for Traces, the kebab should be removed entirely. For Errata, the kebab should contain only 'Recalculate'.
> 3. For Errata, 'Recalculate' should be in the kebab, not the dropdown.

You are killing me lol, ok will update

PR #10047, opened 2022-04-04
Fixes #34587 - Delete unused settings from DB

sjha4:
Delete certain left-over setting values from DB
Run the migration. You may or may not have the settings being removed from the db depending on your box and how recently the db was created and seeded.

### DIFF
--- a/db/migrate/20220404190836_delete_old_setting_data.rb
+++ b/db/migrate/20220404190836_delete_old_setting_data.rb
@@ -1,0 +1,9 @@
+class DeleteOldSettingData < ActiveRecord::Migration[6.0]
+  def up
+    Setting.where(name: ['default_location_puppet_content', 'pulp_sync_node_action_accept_timeout',
+                         'pulp_sync_node_action_finish_timeout', 'subscriptions_return_host_data']).delete_all
+  end
+
+  def down
+  end
+end


db/migrate/20220404190836_delete_old_setting_data.rb
```diff
@@ -0,0 +1,10 @@
+class DeleteOldSettingData < ActiveRecord::Migration[6.0]
+  def up
+    Setting.where(name: ['default_location_puppet_content', 'pulp_sync_node_action_accept_timeout',
+                         'pulp_sync_node_action_finish_timeout', 'subscriptions_return_host_data']).delete_all
+  end
+
+  def down
+    fail ActiveRecord::IrreversibleMigration
```
ekohl, line 8: I don't think this is irreversible. You'll just get the default settings again. Should this just be empty?
```suggestion
```

sjha4: @ekohl : Updated to use change and let rollback be a no-op.

ekohl: I'm a bit confused why you went with `change` as a method name, mostly because I don't know what it will do. Isn't `change` also called on rollback? https://edgeguides.rubyonrails.org/active_record_migrations.html#using-the-change-method states:
> If you're going to need to use any other methods, you should use reversible or write the up and down methods instead of using the change method.

So I'd stick to using `up` and `down`, leaving `down` empty. Mostly because it's predictable and easy to understand as a reader.

ekohl (approved):

chris1984 (approved):

PR #10048, opened 2022-04-05
Refs #34698 - Add kebab to change a host's content view

jeremylenz:

* Add modal to Content view details card to allow you to change the host's CV & LCE
* Don't display 'Version x.x' when using Default Organization View
* Add tooltips to Content view details card
* Add tooltips to CV icons (previously they just had `title` attributes) *
* Change environment selector to use radio buttons instead of checkboxes when not using multiSelect *

\* These changes will also affect CV UI screens!

The content view selection dropdown displays the version number of the content view that will be in use if that CV is selected. Along with the change to radio buttons, this is new functionality that wasn't previously anywhere else.

Test with several CV's and environments:
* at least one environment with _no_ CV's promoted to it
* at least one CV which has different versions promoted to different environments
* etc

Make sure that
* modal interactions are as expected
* CV version numbers are correct in the dropdown
* Save button works as expected
* CV version number does not show on CV card when using Default Organization View

#### ~~Still on the to-do list~~
- [x] Tests
- [x] Change environment selector to radio buttons

### DIFF
--- a/webpack/components/extensions/HostDetails/Cards/ContentViewDetailsCard/ChangeHostCVModal.js
+++ b/webpack/components/extensions/HostDetails/Cards/ContentViewDetailsCard/ChangeHostCVModal.js
@@ -1,6 +1,7 @@
 import React, { useState, useCallback, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { FormattedMessage } from 'react-intl';
+import { useDispatch, useSelector } from 'react-redux';
 import { Modal, Button, Select, SelectOption } from '@patternfly/react-core';
 import { translate as __ } from 'foremanReact/common/I18n';
 import { useAPI } from 'foremanReact/common/hooks/API/APIHooks';
@@ -8,6 +9,8 @@ import EnvironmentPaths from '../../../../../scenes/ContentViews/components/Envi
 import { ENVIRONMENT_PATHS_KEY } from '../../../../../scenes/ContentViews/components/EnvironmentPaths/EnvironmentPathConstants';
 import api from '../../../../../services/api';
 import CONTENT_VIEWS_KEY from '../../../../../scenes/ContentViews/ContentViewsConstants';
+import getContentViews from '../../../../../scenes/ContentViews/ContentViewsActions';
+import { selectContentViewError, selectContentViews, selectContentViewStatus } from '../../../../../scenes/ContentViews/ContentViewSelectors';
 
 const ENV_PATH_OPTIONS = { key: ENVIRONMENT_PATHS_KEY };
 const CV_OPTIONS = { key: CONTENT_VIEWS_KEY };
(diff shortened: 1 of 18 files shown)


webpack/components/extensions/HostDetails/Cards/ContentViewDetailsCard/ChangeHostCVModal.js
```diff
@@ -0,0 +1,258 @@
...
+  return (
+    <span style={descriptionStyle}>
+      <FormattedMessage
+        id={`content-view-${cv.id}-version-${cv.latest_version}`}
+        defaultMessage="Version {versionNumber}"
+        values={{ versionNumber }}
+      />
+    </span>
+  );
+};
```
jeremylenz, line 42: You may be wondering, Why not just use the `description` prop of `<SelectOption>`? It's because it makes it impossible to align with the icon as per the mockup, and instead puts it on the far left, under the icon. So I instead did the above, duplicating the styling and importing the same text color from Patternfly, and then using `<Flex>` below to get the layout correct.

webpack/components/extensions/HostDetails/Cards/ContentViewDetailsCard/ChangeHostCVModal.js
```diff
@@ -0,0 +1,258 @@
...
+      host: {
+        content_facet_attributes: {
+          content_view_id: selectedCVForHost,
+          lifecycle_environment_id: selectedEnvId,
+        },
+      },
+    };
+    dispatch(updateHostContentViewAndEnvironment(
+      requestBody, hostId,
+      refreshHostDetails, handleModalClose,
```
jeremylenz, line 135: I noticed that when there is an error, the modal didn't close, making it hard to see the error toast. So I passed `handleModalClose` here as a `handleError`.

webpack/components/extensions/HostDetails/Cards/ContentViewDetailsCard/ChangeHostCVModal.js
```diff
@@ -0,0 +1,258 @@
...
+    return (contentViewsInEnv.length === 0) ? __('No content views available') : __('Select a content view');
+  }, [contentViewsInEnv.length, contentViewsInEnvStatus]);
+
+  const modalActions = ([
+    <Button
+      key="add"
+      variant="primary"
+      onClick={handleSave}
+      isDisabled={!canSave}
+      isLoading={hostUpdateStatus === STATUS.PENDING}
```
jeremylenz, line 151: An earlier version of me would have tried to keep track of this in component state. But I've found it's so much easier just to look at the status directly and render based on that :)

webpack/components/extensions/HostDetails/Cards/ContentViewDetailsCard/ContentViewDetailsCard.js
```diff
@@ -0,0 +1,202 @@
+import React, { useState } from 'react';
```
jeremylenz, line 1: This file shows as added but really it's just moved (and added the modal to it).

webpack/scenes/ContentViews/components/EnvironmentLabels.js
```diff
@@ -2,14 +2,15 @@ import React from 'react';
 import { Label } from '@patternfly/react-core';
 
 const EnvironmentLabels = (environments) => {
-  const { environments: singleEnvironment } = environments || {};
+  const { environments: singleEnvironment, isDisabled } = environments || {};
```
jeremylenz, line 5: Not sure if this was intended, but deconstructing props in this way (rather than one line above in the function declaration) is a sneaky way to avoid eslint requiring PropTypes. don't tell anyone :hand_over_mouth:

webpack/scenes/ContentViews/components/EnvironmentPaths/EnvironmentPaths.js
```diff
@@ -47,16 +49,16 @@ const EnvironmentPaths = ({
               {index === 0 && <hr />}
               <FormGroup key={`fg-${index}`} isInline fieldId="environment-checkbox-group">
                 {environments.map(env =>
-                  (<Checkbox
+                  (<CheckboxOrRadio
```
jeremylenz, line 52: I was pleasantly surprised that this is the only change needed to add radio buttons. Turns out `<Checkbox>` and `<Radio>` both take the same props that we're passing here.

jeremylenz:
Updated with tests :+1:

webpack/utils/helpers.js
```diff
@@ -106,6 +106,8 @@ export const apiError = (actionType, result, additionalData = {}) => (dispatch)
   return resultWithSuccessFlag(result);
 };
 
+export const uniq = items => [...new Set(items)];
+
```
sjha4, line 110: ❤️

webpack/components/extensions/HostDetails/Cards/ContentViewDetailsCard/ChangeHostCVModal.js
sjha4: Can you export this or move it to a component directory so we can add it on the CV UI?

sjha4 (approved):
Works great on the host details and the changes this makes on the CV-UI are also working great.
Code looks good..Ack 👍🏼

PR #10053, opened 2022-04-07
Fixes #34749 - Add ouia-ids to CV buttons

sjha4:
This is to aid in airgun automation. These are the buttons that needs to have ouia-ID:
- create cv
- publish cv
- next, back, and cancel button in the publish screen
- add/remove repositories
- close button in the progress bar workflow screen
- add, remove repo from the kebab menu
Some components in PF4 don't support ouia-id. Example: Wizard footer buttons in basic wizard setup, Kebab dropdown button.
The publish wizard next/close buttons will not have static ouia-id in this case.
Make sure all primary/secondary buttons visible on the CV UI pages have ouia-ids.
For Repository tab:
Check ouia-id for bulk button, bulk action dropdown, the dropdown actions.


chris1984 (approved):
ACK, I was able to see ouia-ids on all the things listed to check

PR #10057, opened 2022-04-07
Fixes #34736 - Show added filter rule content even if repository is removed

sjha4:
Add all added content to filter rules irrespective of repository being in CV or not.
Only 3 content type rules are impacted:
1. Module Stream
2. Errata by Id
3. Package groups
These rules add actual content unit records to the filter rule and not conditions like date & type or nvra to filter. 
Steps to reproduce for errata.
1. Sync a repo with errata.
2. Create a CV and add the repo to it.
3. Create an errata filter, add any errata from the synced repo to the filter.
4. Remove repo from CV.
5. The Errata filter does not show the added errata from the repo although it is still applied in the backend.

Repeat the steps for module streams, package groups.
Verify that everything works properly.

### DIFF
--- a/app/controllers/katello/api/v2/errata_controller.rb
+++ b/app/controllers/katello/api/v2/errata_controller.rb
@@ -35,7 +35,9 @@ module Katello
     end
 
     def all_for_content_view_filter(filter, _collection)
-      Erratum.joins(:repositories).merge(filter.applicable_repos)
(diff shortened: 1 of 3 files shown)


chris1984 (approved):
ACK, works great.

Tested Errata, Module Stream and Package group.

PR #10060, opened 2022-04-12
Fixes #34760 - Hide profiles for module streams

sjha4:
Hide Module Stream profiles from UI/Hammer/API since we don't have that data anymore.
Pulp no longer provides the profile information for module streams for katello to index.
We have an issue open for pulp to consider adding it https://github.com/pulp/pulp_rpm/issues/2456
In the meantime, we should hide user-facing profile details since it's empty.
On UI:

Sync a repo with modules.
Go to > Content > Module Streams > module_stream_record
You shouldn't see the profiles tab.
Also check the breadcrumb bar. Due to some changes in foreman, the breadcrumb on the module stream page had stopped working.

In hammer
hammer -r module-stream info --id 1
should not show profile section anymore.

### DIFF
--- a/app/views/katello/api/v2/module_streams/show.json.rabl
+++ b/app/views/katello/api/v2/module_streams/show.json.rabl
@@ -6,13 +6,6 @@ child :artifacts => :artifacts do
   attributes :id, :name
 end
 
-child :profiles => :profiles do
-  attributes :id, :name
-  child :rpms => :rpms do
-    attributes :id, :name
-  end
-end
-
 child :library_repositories => :repositories do
   attributes :id, :name
   glue :product do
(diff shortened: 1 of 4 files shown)


sjha4:
[test katello]

sjha4:
[test katello]

chris1984 (approved):
ACK,

Works great hammer output:
```bash
[vagrant@centos7-hammer-devel ~]$ hammer -r module-stream info --id 1
Id:                 1
Module Stream Name: walrus
Stream:             5.21
Uuid:               /pulp/api/v3/content/rpm/modulemds/b2900713-6980-4de9-9258-ff829b531b49/
Version:            20180704144203
Architecture:       x86_64
Context:            deadbeef
Repositories:       
 1) Id:   1
    Name: Food
Artifacts:          
 1) Id:   1
    Name: walrus-0:5.21-1.noarch
```

Profile section is gone under the Content -> Module Streams tab. Tested the breadcrumbs on the new host detail and was able to see other hosts and go back to all hosts.

Do you think it's worth making a migration to remove the `katello_module_profiles` table or just keep it since there is an rfe open against Pulp?

sjha4:
> Do you think it's worth making a migration to remove the katello_module_profiles table or just keep it since there is an rfe open against Pulp?

We are not entirely removing the data model and data right away cause we want to see the resolution on the pulp issue. If we can refactor the code a little based on pulp implementation in the future to support it, it'd be nice.

chris1984:
ship it :+1:

PR #10061, opened 2022-04-12
Fixes #34766 - Show sync progress for never synced repos

sjha4:
For some reason, the page doesn't update with changing data if we use the if/else to render the td.
What happens because of that is if @repo_status[repo.id][:sync_id] is not present(In case of never synced repos. If a repo has been synced atleast once, this is true and hence we do see status updates reflect for synced repos.)
1. Create a new repo/ Enable a redhat repo
2. Go to sync status page
3. Select new repo (should say "never synced") and synchronize on the page.
4. With this PR, you should see a progress bar for the sync.
5. It should behave similarly for already synced repos for a resync.


chris1984:
[test katello]

sjha4:
[test katello]

PR #10064, opened 2022-04-14
Fixes #33608 - Use settings DSL

jeremylenz:

* Use the new Settings DSL to define Katello settings
* Before: Settings were defined in `content.rb` and loaded from `engine.rb`. Now: Settings are defined in `plugin.rb` directly. 
* Move active record callbacks and actions to setting_extensions
* Move our one setting validator to `app/lib/katello/validators`
* Move setting helper methods to `settings` block of `plugin.rb`
* clean up some capitalization etc.
* hopefully not break anything

see the inline comments I will add

* Navigate to Administer > Settings and take note of some Content setting values, particularly non-default ones (or set some, if you like)
* Stop your server; Run the migration; Start your server (Warning: This migration is not reversible)
* Visit Administer > Settings and verify that all Content settings are still present and still have their values
* Verify that you can change setting values without error

* Test the validator: try doing `Setting[:content_default_http_proxy] = 'foo'`; you should get `ActiveRecord::RecordInvalid: Validation failed: There is no such HTTP proxy`
* Test the after-action: Set that same setting to a valid value (nil is also valid) and verify that `SettingExtensions#update_global_proxies` runs `::Actions::BulkAction` and `::Actions::Katello::Repository::UpdateHttpProxyDetails`
* Do a `rake katello:reset` and verify that all the migrations can complete without error

__Important:__ After switching away from this branch you'll probably need to rebase on this PR or things will start breaking.


app/models/katello/concerns/setting_extensions.rb
```diff
@@ -6,8 +6,11 @@ module SettingExtensions
       included do
         validates :value, inclusion: { in: ::Katello::RootRepository::DOWNLOAD_POLICIES },
           if: ->(setting) { setting.name == 'default_download_policy' }
+        validates_with ::Katello::Validators::ContentDefaultHttpProxySettingValidator,
```
jeremylenz, line 9: I found that calling the validator from here actually works. @ezr-ondrej hopefully that's ..valid? :yum:

db/migrate/20220419193414_content_settings_to_dsl_category.rb
```diff
@@ -0,0 +1,5 @@
+class ContentSettingsToDslCategory < ActiveRecord::Migration[6.0]
+  def up
+    Setting.where(category: 'Setting::Content').update_all(category: 'Setting')
+  end
```
jeremylenz, line 4: I modeled this after https://github.com/theforeman/foreman-tasks/pull/659 and others which seem to be irreversible, but don't have `down` methods that throw `ActiveRecord::IrreversibleMigration`. I guess it's okay?

lib/katello/plugin.rb
```diff
@@ -302,6 +303,370 @@
...
+        type: :integer,
+        default: 60 * 20,
+        full_name: N_('Timeout when refreshing a manifest (in seconds)'),
+        description: N_("Manifest refresh timeout")
+
+      setting 'content_action_accept_timeout',
+        type: :integer,
+        default: 20,
+        full_name: N_('Accept action timeout'),
+        description: N_("Time in seconds to wait for a host to pick up a katello-agent action")
```
jeremylenz, line 422: changed these descriptions to make it more clear that they apply only for katello-agent and not REX

lib/katello/plugin.rb
```diff
@@ -302,6 +303,370 @@
     multiple_actions_provider :hosts_change_content_source
   end
 
+  settings do
+    category :katello, N_('Content') do
```
jeremylenz, line 307: The `:katello` here doesn't seem to affect the category of the settings, which is something I don't really understand.. @ezr-ondrej

ezr-ondrej: What do you think by doesn't affect? It is internal category name, that users do not really see, so what have you expected it to affect?

db/migrate/20220419193414_content_settings_to_dsl_category.rb
ezr-ondrej: Yeah, the error means, that it is point of no return for users, we should think twice before introducing such a heavy caliber. Tho it would not hurt to have such thinks, because there are certainly some points of no return :)

app/models/setting/content.rb
```diff
@@ -1,13 +1,13 @@
 class Setting::Content < Setting
```
ezr-ondrej, line 1: Keeling this class might be a bad idea as it is still creating settings in a DB on app start, so you'd have the settings with old and new category. Do we need to keep it?

app/models/katello/concerns/setting_extensions.rb
ezr-ondrej: It is "valid" but there is a `validates_with` setting DSL: https://github.com/theforeman/foreman/blob/develop/developer_docs/how_to_create_a_plugin.asciidoc#validate-setting-values

in a long run, I think you should not need patching the setting class :)

app/models/katello/concerns/setting_extensions.rb
```diff
@@ -6,8 +6,11 @@ module SettingExtensions
       included do
         validates :value, inclusion: { in: ::Katello::RootRepository::DOWNLOAD_POLICIES },
           if: ->(setting) { setting.name == 'default_download_policy' }
```
ezr-ondrej, line 8: This can also be rewritten using [setting validations DSL](https://github.com/theforeman/foreman/blob/develop/developer_docs/how_to_create_a_plugin.asciidoc#validate-setting-values), that will abstract the if condition for you.

lib/katello/plugin.rb
jeremylenz: I expected it to affect the category here
```
[1] pry(main)> Setting.find_by(name: 'content_default_http_proxy')
=> #<Setting:0x000055ee680cb7a8
 id: 1,
 name: "content_default_http_proxy",
 value: nil,
 description: "Default HTTP Proxy for syncing content",
 category: "Setting",
 settings_type: nil,
 default: "--- \n...",
 created_at: Fri, 18 Mar 2022 18:10:51 UTC +00:00,
 updated_at: Tue, 19 Apr 2022 20:32:06 UTC +00:00,
 full_name: "Default HTTP proxy",
 encrypted: false>
```

ezr-ondrej: no, it does not, it is kept only in memory, as category field in the DB will be dropped and is already not considered for any User related purposes.

app/models/katello/concerns/setting_extensions.rb
jeremylenz: For some reason it wasn't working when I tried to call `validates_with` from the `settings do` block.

Our setting class was already patched which also allows the Rails after-actions, does the DSL have a solution for this too?

jeremylenz: (I'm talking about this: )
```rb
after_save :recalculate_errata_status
after_save :add_organizations_and_locations_if_global_http_proxy
after_commit :update_global_proxies
```

ezr-ondrej: Yeah, that's why I mentioned in a long run... these are unfortunatelly not possible ATM through DSL, but it is definitelly a great idea to add and is not so hard IMHO :)

jeremylenz:
[test katello]

jeremylenz:
I realized we had lots of past migrations that reference the `Setting::Content` category (see the commit I just pushed.) I _think_ that removing the category from the query should be safe and not break migrations..

ezr-ondrej:
> I _think_ that removing the category from the query should be safe and not break migrations..

It should be safe, `name` should be unique throughout the app, so no other categories should have the same name, deleting without category context is fine :+1:
And also those migrations will do nothing in two releases, as those affect upgrades only :)

parthaa (approved):
ACK

PR #10065, opened 2022-04-14
Fixes #34779 - Call upsert_all with uniq records

lfu:
Call upsert_all with uniq records. 
PostgreSQL refuses to insert and update in the same command: https://github.com/rails/rails/issues/35519


ianballou (approved):
I'm cool with this! Ran a bunch of CV copy and repo sync tests and everything checks out.

PR #10067, opened 2022-04-20
Fixes #34788 - render smart_proxies/show also for non-content

evgeni:

This adds back the `else` branch that was removed in #9811

None ;-)

Deploy a Katello with a non-content proxy, without this patch the info page under "Infrastructure → Smart Proxies" renders empty for that particular proxy, while it works fine for those with content. After the patch, all should render correctly.

See https://community.theforeman.org/t/smart-proxy-page-is-empty/28189 for the original report and discussion.


Ron-Lavi (approved):
Thanks @evgeni, nice fix :)

ehelms (approved):

jeremylenz (approved):
Thanks @evgeni !

PR #10075, opened 2022-04-25
Fixes #34816 - Add Installed Products card

jeremylenz:

Add an Installed products card to the new Details tab.
installed_products_card

The mockup had the products as links. But I don't know what they would link to. And the old content host page doesn't have links.

Right now there seems to be a bug where if a card is not expanded, it hides its data but still takes the height of the tallest card. (this may be a PF issue or a Foreman issue..)

* Settings > Show experimental labs - Yes
* View the new host details page > Details tab


chris1984:
Tested with a RHEL host and was able to see the new card:

screenshot-192 168 121 28-2022 04 26-14_05_37

Tested with a CentOS host and didn't see a card so confirmed the `null` is returning correct:
noprod

Code is clear and easy to understand and didn't see any console errors when looking at the pages. ACK :beers:

chris1984 (approved):
Look forward to using this as a template on my card :)

PR #10076, opened 2022-04-26
Fixes #34818 - Fix Upstream authentication autofill issue

Manisha15:
This PR fixes autofill issue in Katello repository forms caused by some browsers.

Test Cases:
1. Create Katello repository and upstream_username/password shouldn't be auto-filled by browser.
2. Editing Katello repository which doesn't contain upstream_username/password set by user should have empty upstream_authentication fields.
3. Editing Katello repository which contains upstream_username/password set by user should have show upstream_authentication fields as username and hidden password.

### DIFF
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/repository-details-info.controller.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/repository-details-info.controller.js
@@ -89,7 +89,7 @@ angular.module('Bastion.repositories').controller('RepositoryDetailsInfoControll
                 return deferred.promise;
             };
 
-            $scope.save = function (repository, saveUpstreamAuth) {
+            $scope.save = function (repository) {
                 var deferred = $q.defer();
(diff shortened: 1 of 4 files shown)


sjha4 (commented):
Test failure is related.

Also, could you add 
```
if(!repository['upstream_username']) {
    repository['upstream_password'] = null;
}
```
right below the code that does this: 
```
angular.forEach(fields, function(field) {
                    if (repository[field] === '') {
                        repository[field] = null;
                    }
                });
```
Without that, a user cannot edit upstream auth fields to clear out username and password cause password is hidden.

sjha4:
Looks good. One issue from the added code from review above is it lets me save with only password by clearing the password before sending the update call. We should probably remove it and use the x button to clear out username/password instead as that feels like a better UX than silently deleting password and not actually saving it/throwing an error.
 
Sorry for the back and forth.

Manisha15:
> Looks good. One issue from the added code from review above is it lets me save with only password by clearing the password before sending the update call. We should probably remove it and use the x button to clear out username/password instead as that feels like a better UX than silently deleting password and not actually saving it/throwing an error.
> 
> Sorry for the back and forth.

I have updated the code. Although I didn't remove the suggested code as it helps in removing password field if username doesn't exist, I updated the code so that if user erases the credentials it properly removes whole credential and not just username.

sjha4 (approved):

PR #10077, opened 2022-04-26
Fixes #34781 - Add rake script to push all repos from katello to pulp2/3

phess:
This rake script was motivated by some events where we could not ascertain that pulp was hosting all repositories and content views that katello knew about. In these cases, the existing katello:correct_pulp_repositories rake script would sometimes end up removing from katello the repositories that were missing from pulp, instead of pushing those repos from katello to pulp.

This rake script here triggers a repo creation task on the local pulp/pulp3 for every repository known to katello. Existing repositories will usually succeed immediately or fail silently -- both these outcomes are OK.

This rake script can optionally target an individual Lifecycle Env or a CV or a repo as a way to narrow down the contents to be pushed to pulp in case one knows exactly which repos are missing on pulp.

I look forward to this team's feedback.

### DIFF
--- a/lib/katello/tasks/create_pulp_repos.rake
+++ b/lib/katello/tasks/create_pulp_repos.rake
@@ -1,0 +1,110 @@
+namespace :katello do
+  desc <<-DESC
+    Create on Satellite's internal Pulp instance all repositories known to Foreman/Katello. ENV variables:
+
+      Optional:
+        * LIFECYCLE_ENVIRONMENT : name or numeric ID of the Lifecycle Environment to limit repositories to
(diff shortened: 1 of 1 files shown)


chris1984:
ok to test

chris1984:
[test katello]

chris1984:
@phess looks like there are some rubocop failures

phess:
> @phess looks like there are some rubocop failures

@chris1984 , thank you for the warning. I can't view which rubocop tests failed, or why they failed. Is there a way to re-run the failed tests so I can view their output?

qcjames53:
We're conducting a review of the open PRs in Katello this afternoon and thought this would be a good candidate for closure.

PR #10087, opened 2022-05-04
Fixes #34862 - Rails 6.1 updates

parthaa:
In rails 6.1 the following line in repository controller was causing a
'Frozen' error.

RootRepository.attribute_names.concat(....) + generic_repo_wrap_params

In 6.1 looks like RootRepository.attribute_names is frozen so tweaked
this method a little. Also rremoved the unused ignore_global_proxy
attribute from the wrap params


chris1984 (approved):
Works well, thanks @parthaa

PR #10089, opened 2022-05-04
Fixes #34856 - Add TracerEnabler progress indicator and poll Rex jobs

jeremylenz:
- [x] rebase
- [x] fix button enablement bug
- [x] add sticky toast
- [x] un-stick toast after job succeed or fail
- [x] remove cursor that happens when mousing over disabled checkboxes

traces_enable

1. Add a loading screen for enabling Traces:

Succeed:
https://www.awesomescreenshot.com/video/8739856?key=9654f43a18a51c1da28a013fa47d4333

Fail:
https://www.awesomescreenshot.com/video/8739884?key=3e443826b2dab30f6b460dfa1f647be9

2. On all host details tabs, poll Rex jobs. While in progress, display a spinner. When they succeed, display a success toast and reload results. When they fail, display an error toast.

Package remove with results reload:
https://www.awesomescreenshot.com/video/8774035?key=6b9e8e0b8d1cf1edca62e39c91fbfb40

All this is achieved via a new `useRexJobPolling` hook.

After adding the loading indicator to Traces, I couldn't resist spreading the love!

Sadly, I was unable to use the `isLoading` prop of `<Button>` because of upcoming changes in https://github.com/Katello/katello/pull/10046. So I just added a standard spinner instead, which will allow consistency across all tabs.

* Get a working REX setup
* Enable the Satellite Tools repo so you can install Tracer (you'll probably have to override to enabled in Repo sets as well)
* New host detail page > Traces > Enable Tracer
* Make sure everything works ok and you get toasts as in the videos
* Try it again with a different host which is not running and view the failure mode

Also test the following pages and confirm that

1. you get toasts for starting, success, and failure of the job
2. results are reloaded ONLY on success (not on starting the job and not on failure)
3. Polling stops if you navigate away from the tab
4. When results are reloaded, the current search and pagination options are remembered
5. checkboxes are disabled while job is in progress, with small tooltip on mouse hover explaining the reason
6. immediate remote execution actions are disabled while job is in progress (this is to prevent users from trying the same job twice in a row, etc.)
7. customized remote execution actions remain unchanged

* Errata application
* Package remove
* Package upgrade
* Package install
* Traces restart
* Module stream actions

### DIFF
--- a/webpack/components/extensions/HostDetails/HostDetailsActions.js
+++ b/webpack/components/extensions/HostDetails/HostDetailsActions.js
@@ -2,11 +2,11 @@ import HOST_DETAILS_KEY from './HostDetailsConstants';
 
 const hostIdNotReady = { type: 'NOOP_HOST_ID_NOT_READY' };
 
-export const refreshHostDetails = ({ hostName }) => ({
+export const refreshHostDetails = ({ hostname }) => ({
   type: 'API_GET',
   payload: {
     key: HOST_DETAILS_KEY,
-    url: `/api/hosts/${hostName}`,
+    url: `/api/hosts/${hostname}`,
   },
 });
 
--- a/webpack/components/extensions/HostDetails/HostDetailsActions.js
+++ b/webpack/components/extensions/HostDetails/HostDetailsActions.js
@@ -2,7 +2,7 @@ import HOST_DETAILS_KEY from './HostDetailsConstants';
 
 const hostIdNotReady = { type: 'NOOP_HOST_ID_NOT_READY' };
 
-export const refreshHostDetails = ({ hostname }) => ({
+export const getHostDetails = ({ hostname }) => ({
   type: 'API_GET',
   payload: {
     key: HOST_DETAILS_KEY,
--- a/webpack/components/extensions/HostDetails/HostDetailsActions.js
+++ b/webpack/components/extensions/HostDetails/HostDetailsActions.js
@@ -1,2 +1,13 @@
+import HOST_DETAILS_KEY from './HostDetailsConstants';
+
 const hostIdNotReady = { type: 'NOOP_HOST_ID_NOT_READY' };
+
+export const refreshHostDetails = ({ hostName }) => ({
+  type: 'API_GET',
+  payload: {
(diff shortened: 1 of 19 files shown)


webpack/components/extensions/HostDetails/Tabs/TracesTab/TracesEnabler.js
```diff
@@ -0,0 +1,104 @@
+import React, { useState } from 'react';
```
jeremylenz, line 1: file formerly known as `EnableTracerEmptyState.js`

webpack/components/extensions/HostDetails/HostDetailsActions.js
```diff
@@ -1,2 +1,13 @@
+import HOST_DETAILS_KEY from './HostDetailsConstants';
+
 const hostIdNotReady = { type: 'NOOP_HOST_ID_NOT_READY' };
+
+export const getHostDetails = ({ hostname }) => ({
+  type: 'API_GET',
+  payload: {
+    key: HOST_DETAILS_KEY,
+    url: `/api/hosts/${hostname}`,
```
jeremylenz, line 9: This action was left over from an intermediate commit, but I decided that it still makes more sense for the host collections modal to import the action from here rather than just doing it in that component. This way the action can be easily reused in the future.

jeremylenz:
@sjha4 updated and ready for final review :)

webpack/components/extensions/HostDetails/Tabs/TracesTab/TracesTab.js
```diff
@@ -193,9 +208,21 @@ const TracesTab = () => {
...
             } = result;
+            // if (actionInProgress) {
+            //   return (
+            //     <Tr>
+            //       <Td key={id} colSpan={5}>
+            //         <Skeleton />
+            //       </Td>
+            //     </Tr>
+            //   );
+            // }
```
sjha4, line 219: Fine to remove this I guess..

jeremylenz:
@sjha4 thanks, updated!

sjha4 (approved):

PR #10058, opened 2022-04-08
Fixes #9397 - add ouiaId prop to all editable fields in CDN Configuration

lfu:
#### What are the changes introduced in this pull request?
Add `ouiaId` prop to all editable fields in CDN Configuration
#### What are the testing steps for this pull request?
Content -> Subscriptions -> Manage Manifest -> CDN Configuration
Go to each one of  the following: 
  `Red Hat CDN`, `Upstream Foreman server`, and `Air-gapped`
Inspect all editable fields to ensure that the `data-ouia-component-id` prop is now static.

PR #10092, opened 2022-05-05
Refs #34754 - add ouiaId prop to EditableTextInput

lfu:
Add ouiaId prop to EditableTextInput which is used in CDN Configuration

Followup of https://github.com/Katello/katello/pull/10058

1. Go to CDN configuration > Network Sync
2. Inspect element for pencil icon next to password

Expected Result:
Pencil icon, check mark, and cancel buttons all have unique and static data-ouia-component-id.

### DIFF
--- a/webpack/components/EditableTextInput/EditableTextInput.js
+++ b/webpack/components/EditableTextInput/EditableTextInput.js
@@ -86,6 +86,7 @@ const EditableTextInput = ({
       </SplitItem>
       <SplitItem>
         <Button
+          ouiaId={`submit-button-${attribute}`}
           aria-label={`submit ${attribute}`}
           variant="plain"
           onClick={onSubmit}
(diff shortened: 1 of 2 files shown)


webpack/components/EditableTextInput/EditableTextInput.js
```diff
@@ -86,6 +86,7 @@ const EditableTextInput = ({
       </SplitItem>
       <SplitItem>
         <Button
+          ouiaId={`submit-button ${attribute}`}
```
jeremylenz, line 89:
```suggestion
          ouiaId={`submit-button-${attribute}`}
```
Feels a bit more consistent if we add the dash here so it's `edit-button-cdn-password` and not `edit-button cdn-password`

(applies to the buttons below as well)

jeremylenz (commented):
Thanks @lfu !

I tested and confirmed that the `data-ouia-component-id` attributes are static and present as expected.

Just one minor comment below.

jeremylenz (approved):
ACK :+1:

PR #9973, opened 2022-02-23
Refs #33010 - fix slot's id

amirfefer:
Fixing the overview tab cards slot's id, once we add the new details tab, the `details-cards` id is too ambiguous.

 blocked by https://github.com/theforeman/foreman/pull/9054

PR #10095, opened 2022-05-06
Refs #34877 - fix slot's id

jeremylenz:

stop use of deprecated slot id. Taking over from https://github.com/Katello/katello/pull/9973

If all your plugins are updated, you should no longer see the slot deprecation warning in the JS console when visiting the Overview tab of the new host details page.

also, if your cards were out of order, this should fix it

### DIFF
--- a/webpack/global_index.js
+++ b/webpack/global_index.js
@@ -32,18 +32,18 @@ addGlobalFill('host-details-page-tabs', 'Repository sets', <RepositorySetsTab ke
 
 // Overview tab cards
 addGlobalFill(
-  'details-cards',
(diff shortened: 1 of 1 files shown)


jeremylenz:
made a new Redmine so it will be in the correct project and 'Redmine issues' will no longer be sad

jeremylenz:
I think the intermittent test failure is addressed in https://github.com/Katello/katello/pull/10092

lfu:
With this patch, the following warning in the JS console was gone.

Screenshot from 2022-05-10 13-36-07

lfu (approved):

PR #10096, opened 2022-05-09
Fixes #32104 - Retain tasks used for errata reporting for longer time

adamruzicka:
Erratum::ApplicableErrataInstall and Erratum::Install tasks are kept around for 90 days instead of 30.

This is a workaround and as such has a few hairy bits:

Foreman-tasks has a configuration file where this can be configured, however currently the configuration file comes straight from the packages (no installer involved), in packaging we take the example config file without making any changes to it. On update people would (maybe) still need to deal with .rpmnew files. 

If we changed the defaults in the config file, we would shipping configuration for other plugins, which may not be enabled. This would cause warnings to appear in output when running the cleanup rake task without katello or remote_execution.

Alternatively we could make the installer manage that configuration file, but that could get out of hand really quickly.

Of course the right solution would be to consider tasks ephemeral and store the important bits elsewhere, but here we are.

1) Have errata applications older than 1 month but younger than 3 months
2) Run rake foreman_tasks:cleanup

Alternatively just run rake foreman_tasks:cleanup with SQL logging enabled and watch the queries it does.

### DIFF
--- a/app/lib/actions/katello/extensions/rex_jobs_extensions.rb
+++ b/app/lib/actions/katello/extensions/rex_jobs_extensions.rb
@@ -1,0 +1,11 @@
+module Actions
+  module Katello
+    module Extensions
+      module RexJobsExtensions
+        def cleanup_rules
+          ::ForemanTasks::ActionRule.new(self, '90d', 'remote_execution_feature.label = katello_errata_install')
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


ares (approved):
Makes sense, thanks! I don't have merge permissions here but :+1:

adamruzicka:
Now requires https://github.com/theforeman/foreman-tasks/pull/687

adamruzicka:
Rebased and bumped dependency on tasks

adamruzicka:
It was decided this should not be hardcoded and left to the user to configure. Closing

PR #10097, opened 2022-05-09
Fixes #34803 - Display 0 package counts

lfu:
To display package count and icon even if the count is 0 on the content hosts table.

Hosts - Content Hosts

Expected
Screenshot from 2022-05-09 09-29-50
Screenshot from 2022-05-09 09-30-21


chris1984 (approved):
ACK. Code looks good and works well on my system with centos7 and ubuntu 20.04. I was able to see the icons with rpm/deb based on the os type.

PR #10101, opened 2022-05-10
Fixes #34880 - auto-create missing distributions and improve errors

ianballou:
This PR improves distribution auto-creation and error wording when publications are missing.

I removed `middleware.use Actions::Middleware::ExecuteIfContentsChanged` from `SaveDistributionReferences` because it was getting in the way of saving references when a distribution is recreated. It seems perfectly safe to run even if no distributions are created and shouldn't waste any noticible time.

1) Sync a repo and set the mirror policy to 'mirror complete'
2) Destroy the repo's distribution by looking up the distribution reference
```ruby
::Katello::Repository.find(id).backend_service(SmartProxy.pulp_primary).distribution_reference.href
```
Then destroy it via the Pulp API or the Pulp CLI (recommended! https://docs.pulpproject.org/pulp_cli/install/#Katello)
3) Sync the repo and check that the distribution reference changed.
  -> Also check that content is hosted properly.
4) Destroy the repo's publication in Pulp now.  Find the publication with `::Katello::Repository.find(id).publication_href`
5) Add the repo to a content view without filters and try to publish. See the new helpful error.  Skip the failing publish steps and destroy the CVV.
6) Sync the repo normally. Note the new helpful error.  Follow the error's instructions and see that the repo syncs and has properly hosted content.
7) Try publishing the CV again and see that it works.
8) Run steps 1-7 on a new repo with a different mirroring policy ('additive' or 'content only').


ianballou:
The test failures should all just be VCRs and tests that need to be adapted to this new error.

sjha4:
[test katello]

test/services/katello/pulp3/repository_integration_test.rb
```diff
@@ -1,6 +1,7 @@
 require 'katello_test_helper'
 require 'support/pulp3_support'
 require 'katello/pulp3/repository'
+require 'pry-byebug'
```
sjha4, line 4: This is causing the tests to fail.

sjha4:
Able to see error message with name of repo with missing publication during publish and sync. Refresh distribution is called on every sync which shouldn't be too bad with 1 additional call per repo. 
Screenshot from 2022-05-18 13-11-23

Screenshot from 2022-05-18 13-10-35

Looks good to me other than the one jenkins error in the previous comment.

ianballou:
> Refresh distribution is called on every sync which shouldn't be too bad with 1 additional call per repo.

I thought so too, plus if there's nothing to save the tasks will essentially do nothing.

sjha4 (approved):
APJ 👍🏼

PR #10105, opened 2022-05-11
Fixes #34904 - Add Details tab - HW properties card

chris1984:

* Adding info on hardware such as model, ram/cpu and storage

* Making sure all data is reported correctly from the facts and reported data object
* Moved storage to bottom since it was 4th on the mockup so that way incase the value is `null` we hide it, it does not mess up the layout of the rest of the card. 

* Checkout PR
* Settings > Show experimental labs - Yes
* Register a content host with subscription manager
* Verify there are no console errors and check to see if the card is correct.
* Storage field is hidden unless `totalDisks` comes back with a value, to test to make sure that works you can edit the following value `disks_total` in the `host_facets_reported_data_facets` table to 1 or 2 and you can see the singluar/plural disks value populate the filed correctly.

### DIFF
--- a/webpack/components/extensions/HostDetails/DetailsTabCards/HwPropertiesCard.js
+++ b/webpack/components/extensions/HostDetails/DetailsTabCards/HwPropertiesCard.js
@@ -1,0 +1,100 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { propsToCamelCase } from 'foremanReact/common/helpers';
+import { translate as __ } from 'foremanReact/common/I18n';
+import {
+  DescriptionList,
+  DescriptionListTerm,
+  DescriptionListGroup,
+  DescriptionListDescription,
+  Text,
+  TextVariants,
+} from '@patternfly/react-core';
+import CardTemplate from 'foremanReact/components/HostDetails/Templates/CardItem/CardTemplate';
+import { TranslatedPlural } from '../../../Table/components/TranslatedPlural';
+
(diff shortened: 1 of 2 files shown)


webpack/components/extensions/HostDetails/DetailsTabCards/HwPropertiesCard.js
```diff
@@ -0,0 +1,99 @@
...
+  TextVariants,
+} from '@patternfly/react-core';
+import CardTemplate from 'foremanReact/components/HostDetails/Templates/CardItem/CardTemplate';
+
+const HostDisks = ({ totalDisks }) => {
+  if (!totalDisks) return null;
+  return (
+    <>
+      <DescriptionListTerm>{__('Storage')}</DescriptionListTerm>
+      <Text component={TextVariants.h4}>{totalDisks} {totalDisks > 1 ? __('disks') : __('disk')}</Text>
```
jeremylenz, line 20: I know we had trauma around this before, but this should be the simple use case:

```suggestion
      <Text component={TextVariants.h4}><TranslatedPlural count={totalDisks} singular={__('disk')} id="total-disks" /></Text>
```

jeremylenz (commented):
Thanks @chris1984 !

Works well. Also verified that the Storage section shows up by using React dev tools.

Just a couple comments

chris1984:
@jeremylenz updated sir

jeremylenz (approved):
Nice work! Thanks @chris1984 

ACK :+1:

PR #10106, opened 2022-05-12
Refs #34749 - ouia-id to ouiaId

sjha4:
Fixes snake case vs hyphenated ouiaId typos in CV UI.
Go to CV UI.
Click on create CV > Right click -> Inspect element on Create content view button.
You should see data-ouia-component-id being set for the button.
Same for Cancel button on the form.


chris1984 (approved):
ACK, works great can see the lables in the inspect

PR #10107, opened 2022-05-12
Fixes #34905 - override_components don't make it to composite CV publ…

ianballou:
Override components weren't being sent properly to the publish task during incremental update with `--propagate-all-composites=true`.  This was due to a simple typo!

It's a simple typo!

~I need to figure out a test that'll cover this.~

Follow these reproducer steps here: https://bugzilla.redhat.com/show_bug.cgi?id=2049799#c0


sjha4 (approved):
Works as advertised. APJ
Screenshot from 2022-05-13 12-39-40

PR #10109, opened 2022-05-12
Fixes #34743 - Allow RPM updateAll package action

sbernhard:
Instead of disabling the 'Update all packages' button if katello-agent is not present, use REX to update packages.

REX is nice and standard use case for new hosts.

Try to update all packages on a host
a) only having katello-agent (should work)
b) only having REX (should work, too)


sbernhard:
Can someone has a look at this PR @Katello/katello-core ?

jeremylenz:
@sbernhard Thanks for the reminder. I added it to our queue for review, so you should be hearing back in the next few days.

chris1984:
Starting to review

chris1984 (approved):
ACK, works great. Thanks @sbernhard

PR #10111, opened 2022-05-13
Fixes #34920 - Rename Component content view to content view

sjha4:
Component content view should be renamed to Content view.
Verify that you don't see the term Component content view on the UI. 😄 

There's a tootip on the component content view icon that now should say Content view. Screenshot won't let me get that.

Other screenshots below:
Screenshot from 2022-05-13 09-19-22
Screenshot from 2022-05-13 09-19-49
Screenshot from 2022-05-13 09-35-01

### DIFF
--- a/webpack/scenes/ContentViews/Create/CreateContentViewForm.js
+++ b/webpack/scenes/ContentViews/Create/CreateContentViewForm.js
@@ -115,7 +115,7 @@ const CreateContentViewForm = ({ setModalOpen }) => {
               aria-label="component_tile"
               icon={<ContentViewIcon composite={false} />}
               id="component"
-              title={__('Component content view')}
+              title={__('Content view')}
               onClick={() => { setComponent(true); setComposite(false); }}
               isSelected={component}
             >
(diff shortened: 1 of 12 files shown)


sjha4:
Updated tests and I think I covered most of the component content view references on the UI.

sjha4:
> * In error messages - https://github.com/jeremylenz/katello/blob/fdbc7f2057ddf7dec1b6aacc564875d88fd494b7/app/models/katello/content_view.rb#L488-L506 - I sort of like the "non-composite content view" term; what do you think about changing the third error to use that?
> * In Setting description - https://github.com/jeremylenz/katello/blob/fdbc7f2057ddf7dec1b6aacc564875d88fd494b7/app/models/katello/content_view.rb#L488-L506 - I think this should be updated
> * In API descriptions -

I think the error message with the term component in the given context inside a composite content view is not that confusing. We have had the term components or component content views in the context of composite CVs for a while now without causing any confusions.

jeremylenz (approved):

PR #10114, opened 2022-05-13
Refs #33592 - Host details permissions audit

jeremylenz:
- [x] host collections card
- [x] content views card
- [x] Packages
- [x] Errata
- [x] Module streams
- [x] Repository sets
- [x] Traces
- [x] Bookmarks
- [x] tests

* Check permissions of the current user
* Don't show actions in the new host details UI that the user doesn't have permission to do

This mostly consists of hiding kebabs and action buttons, but in one case I hid an entire tab.

see inline comments below.

In the new host details tab, create at least one bookmark.

In Administer > Roles, create a custom role with the permissions `'view_lifecycle_environments', 'view_content_views',
  'promote_or_remove_content_views_to_environments'`
In Administer > Users, create 3 users

1. a user with the Viewer role
2. a user with your custom role
3. a user with the Host viewer role

The "Viewer" user should see no action kebabs, checkboxes, or buttons anywhere in the host details page. But should be able to view everything, and perform searches / filters.
The custom role user should see a kebab on the Content View Details card, but no kebab on the Host collections card.
The "Host viewer" user should not see the Repository sets tab.

The "Viewer" and "Host viewer" users should be able to use existing bookmarks, but not see the actions to create new or bookmark this search. If there are no bookmarks created, these users should not see the bookmark component at all. (But admin users should still see it no matter what.)

Play around with other combinations of roles and permissions and make sure everything works as expected. You should not be able to do anything that gets you an "Access denied" error.


app/views/katello/api/v2/content_facet/show.json.rabl
```diff
@@ -33,6 +33,17 @@ child :content_facet => :content_facet_attributes do
...
+  user = User.current # current_user is not available here
+  child :permissions do
+    node(:view_lifecycle_environments) { user.can?("view_lifecycle_environments") }
+    node(:view_content_views) { user.can?("view_content_views") }
+    node(:promote_or_remove_content_views_to_environments) { user.can?("promote_or_remove_content_views_to_environments") }
+    node(:view_host_collections) { user.can?("view_host_collections") }
+    node(:create_job_invocations) { user.can?("create_job_invocations") }
+    node(:view_activation_keys) { user.can?("view_activation_keys") }
+    node(:view_products) { user.can?("view_products") }
+  end
```
jeremylenz, line 46: It seemed more correct to add Katello permissions to one of the facets (content facet or subscription facet) rather than add them in the Foreman codebase, but of the two options (content facet or subscription facet) I wasn't sure which is more relevant. I went with content facet.

This also includes `create_job_invocations` which technically shouldn't even be in Katello at all, as it's from the foreman_remote_execution plugin. But it seems silly to make another plugin extend the host rabl in the same way, so I just did it here.

webpack/components/extensions/HostDetails/Tabs/RepositorySetsTab/RepositorySetsTab.js
```diff
@@ -59,7 +59,19 @@ import { selectRepositorySetsStatus } from './RepositorySetsSelectors';
 import './RepositorySetsTab.scss';
 import SortableColumnHeaders from '../../../../Table/components/SortableColumnHeaders';
 import SelectableDropdown from '../../../../SelectableDropdown';
+import { hasRequiredPermissions, hostIsNotRegistered, missingRequiredPermissions, userPermissionsFromHostDetails } from '../../hostDetailsHelpers';
 
+const requiredViewPermissions = [
+  'view_hosts', 'view_activation_keys', 'view_products',
```
jeremylenz, line 65: If you don't have all 3 of these permissions, the Repository sets API won't even _show_ you anything. So I just hide the tab.

When the tab moves to Content this will need some refactoring..

jeremylenz:
rebased and squashed

ianballou:
I'm guessing the custom role should also have `view_hosts` ? Otherwise I can't see the host details page.

ianballou:
I noticed that the "Viewer" role user can enable traces.  I'm guessing that should be protected since it installs something on the host?

jeremylenz:
> I'm guessing the custom role should also have view_hosts ? Otherwise I can't see the host details page.

yep!

jeremylenz:
> I noticed that the "Viewer" role user can enable traces. I'm guessing that should be protected since it installs something on the host?

good catch! Will fix

ianballou:
I noticed the "Host Viewer" user and the "Custom" user gets a 403 Forbidden on the Reports tab.

jeremylenz:
> I noticed the "Host Viewer" user and the "Custom" user gets a 403 Forbidden on the Reports tab.

That's something we need to bug @Ron-Lavi about :smile:

ianballou:
> The "Viewer" and "Host viewer" users should be able to use existing bookmarks, but not see the actions to create new or bookmark this search.

Is the "Custom" user part of that too? They can see the bookmarks as well.

ianballou (approved):
I tested this a bunch, read through the code, and it all looks fine now!

Only thing I questioned for a second was the permissions for enabling traces, it seems you need some number of permissions besides `create_job_invocations`.  I think just basing the decision off of that permission is fine though, since otherwise we'd need to maintain that entire list of permissions.

ianballou:
[test katello]

PR #10115, opened 2022-05-16
Refs #34923 - fix CloneToVersion generate_metadata

m-bucher:
This fix ensures that the code does, what originally was intended.

I do not know, whether this really is what should be done, since no one seems to have been complaining about this, yet.
Please verify, if setting `generate_metadata=false` for incremental updates should be done (unfortunately, I lack knowledge on how this is all supposed to work :wink: )

This might not be the only occurrence of `generate_metadata` vs. `metadata_generate`, therefore, I used 'Refs' instead of 'Fixes'.

### DIFF
--- a/app/lib/actions/katello/repository/clone_to_version.rb
+++ b/app/lib/actions/katello/repository/clone_to_version.rb
@@ -23,7 +23,7 @@ module Actions
                         rpm_filenames: rpm_filenames,
                         copy_contents: copy_contents,
                         solve_dependencies: content_view.solve_dependencies,
-                        metadata_generate: !incremental)
+                        generate_metadata: true)
           end
         end
 
--- a/app/lib/actions/katello/repository/multi_clone_to_version.rb
+++ b/app/lib/actions/katello/repository/multi_clone_to_version.rb
@@ -10,7 +10,7 @@ module Actions
             plan_action(::Actions::Katello::Repository::MultiCloneContents, extended_repo_map,
                         copy_contents: true,
                         solve_dependencies: true,
-                        metadata_generate: !incremental)
+                        generate_metadata: true)
           end
         end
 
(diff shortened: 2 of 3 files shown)


sbernhard (approved):
Well, it looks like, that all actions which are called are using generate_metadata instead of metadata_generate:
`options.fetch(:generate_metadata, true)`
So, I'm pretty sure that this is "more true" than before :-)

Your thoughts @ianballou?

ianballou:
I'll get this added to the review queue so that someone tests out incremental update and CV publishes. Here's what needs testing:

1) Incrementally update CVs that have filters
2) Incrementally update CVs that don't have filters
3) Try 1 and 2 with dependency solving on/off
4) Publish CVs with filters
5) Publish CVs without filters
6) Try 4 and 5 with dependency solving on/off

m-bucher:
First tests show that with my current fix IncrementalUpdates do no longer create publications at all.
So more work is necessary :innocent:

jeremylenz:
> So more work is necessary

Ok, please comment here again when it's ready for a look :)

ianballou:
@m-bucher, we still have this on our review board.  I'm not sure why we wouldn't want to run metadata generation for incremental updates.  Personally, since we've been running with this code flow for so long, I'd opt to remove the conditional metadata generation (`metadata_generate: !incremental`) entirely.  It's been forever "true" in the clone contents actions this whole time anyway.

m-bucher:
@ianballou sorry, this PR somehow completely dropped of my radar :see_no_evil: 

I rebased and changed the `!incremental` to `true`, as you suggested.

ianballou:
I'll add a reminder to take a look again later, thanks for the update. I forgot all about it too :)

ianballou (approved):

PR #10123, opened 2022-05-19
Fixes #34659 - Change Katello bookmark to Foremans

chris1984:

Maria said should change the Katello bookmark to use Foremans icon

Removed usage of the `<i className="fas fa-bookmark" />`

* Check out PR
* Visit any of the new host detail pages where there is content like packages/errata
* Verify the book mark is the new one


webpack/components/Bookmark/index.js
```diff
@@ -4,7 +4,7 @@ import { useDispatch, useSelector, shallowEqual } from 'react-redux';
 import { translate as __ } from 'foremanReact/common/I18n';
 import { STATUS } from 'foremanReact/constants';
 import { Dropdown, DropdownItem, DropdownToggle, DropdownSeparator } from '@patternfly/react-core';
-import { CaretDownIcon } from '@patternfly/react-icons';
+import { OutlinedBookmarkIcon } from '@patternfly/react-icons';
```
lfu, line 7: It is not the same icon Foreman uses.

chris1984: @lfu I pulled that one from here:

https://github.com/theforeman/foreman/blob/639b7a366ff1e556d710e24e240186e8e5f4c6f7/webpack/assets/javascripts/react_app/components/PF4/Bookmarks/Bookmarks.js#L4

PR #10129, opened 2022-05-25
Fixes #34957 - Put manifest into a shared temp directory

adamruzicka:
On production deployments, dynflow workers have private tmp directories,
meaning they cannot use /tmp as a place for shared data. This could lead
to manifest refresh failing on scaled-up deployments.

Manifest refresh actions now place the exported manifest into Rails.root/tmp (~foreman/tmp on production deployments) instead of /tmp, which may be private to each worker.

Alternatively we could make all the workers share /tmp, private from the rest of the system, but the approach take here felt more correct.

Steps to reproduce are described in the redmine issue.


chris1984 (approved):
ACK

PR #10130, opened 2022-05-25
Fixes #34966

jpasqualetto:
Change order of initialization of the variables, so to avoid having a parser hitting a nil variable.

Fixes an issue that blocks the import of any custom repositories configured to use GPG Keys.

Try importing custom content that has GPG Keys defined.


chris1984:
ok to test

parthaa (approved):
ACK

PR #10132, opened 2022-05-26
Fixes #32282 - Repository restricted to an OS showed in a label on Repository Sets screen

lfu:
Add a label in `Status` column for repositories with `Restrict to OS version` set on Repository Sets screen.

1. Set a value to `Restrict to OS version` for a repository
2. All Hosts - pick your host - Repository Sets 
    Check Status for a label if `Restrict to OS version` is set

### DIFF
--- a/webpack/components/extensions/HostDetails/Tabs/RepositorySetsTab/RepositorySetsTab.js
+++ b/webpack/components/extensions/HostDetails/Tabs/RepositorySetsTab/RepositorySetsTab.js
@@ -96,6 +96,29 @@ EnabledIcon.propTypes = {
   isOverridden: PropTypes.bool.isRequired,
 };
 
+const OsRestrictedIcon = ({ osRestricted }) => (
+  <Tooltip
+    position="right"
+    content={<FormattedMessage
+      id="os-restricted-tooltip"
+      defaultMessage={__('OS restricted to {osRestricted}. If host OS does not match, the repository will not be available on this host.')}
+      values={{ osRestricted }}
+    />}
+  >
+    <Label color="blue" className="os-restricted-label" style={{ marginLeft: '8px' }}>
+      {__(osRestricted)}
+    </Label>
+  </Tooltip>
+);
+
+OsRestrictedIcon.propTypes = {
+  osRestricted: PropTypes.string,
+};
+
+OsRestrictedIcon.defaultProps = {
+  osRestricted: null,
+};
+
 const RepositorySetsTab = () => {
   const hostDetails = useSelector(state => selectAPIResponse(state, 'HOST_DETAILS'));
   const {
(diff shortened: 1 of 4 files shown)


lfu:
Screenshot from 2022-05-26 13-01-36

webpack/components/extensions/HostDetails/Tabs/RepositorySetsTab/RepositorySetsTab.js
```diff
@@ -96,6 +96,25 @@ EnabledIcon.propTypes = {
   isOverridden: PropTypes.bool.isRequired,
 };
 
+const OsRestrictedIcon = ({ osRestricted }) => (
+  <Tooltip
+    position="right"
+    content={__(`OS restricted to ${osRestricted}`)}
+  >
```
sjha4, line 103: Translate will be able to handle string interpolation. Need to use <FormattedMessage> for this.

sjha4: An example of usage is here: https://github.com/Katello/katello/blob/master/webpack/components/extensions/HostDetails/Tabs/ModuleStreamsTab/ModuleStreamsTab.js#L121

sjha4 (approved):
Works nicely.. Ack 👍🏼

webpack/components/extensions/HostDetails/Tabs/RepositorySetsTab/RepositorySetsTab.js
```diff
@@ -96,6 +96,29 @@ EnabledIcon.propTypes = {
   isOverridden: PropTypes.bool.isRequired,
 };
 
+const OsRestrictedIcon = ({ osRestricted }) => (
+  <Tooltip
+    position="right"
+    content={<FormattedMessage
+      id="os-restricted-tooltip"
+      defaultMessage={__('OS restricted to {osRestricted}')}
```
jeremylenz, line 104: I'd be for a more informative message here:
```suggestion
      defaultMessage={__('OS restricted to {osRestricted}. If host OS does not match, the repo will not be available on this host.')}
```

webpack/components/extensions/HostDetails/Tabs/RepositorySetsTab/RepositorySetsTab.js
```diff
@@ -96,6 +96,29 @@ EnabledIcon.propTypes = {
...
+      values={{ osRestricted }}
+    />}
+  >
+    <Label color="blue" className="os-restricted-label">
```
jeremylenz, line 108: The label is too close to the previous label. Recommend adding a style tag with a `marginLeft`

jeremylenz (commented):
Looking at the linked user story,
```
As a user I want to be aware if a custom repository is using a Restrict to Architecture or Restrict to OS version and the values don't match my content host so that I will understand why the repo is not available
```
This takes care of the "be aware" and "Restrict to OS version" parts, but doesn't cover 
* Restrict to Architecture
* "the values don't match my content host"

I think this work should be expanded to cover everything. (I'd be fine with moving the above to a follow-up card.) Thoughts?

lfu:
> I think this work should be expanded to cover everything. (I'd be fine with doing the above in a follow-up card or cards.) Thoughts?

A follow-up card sounds good.

lfu:
@sjha4 @jeremylenz Updated.

PR #10133, opened 2022-05-26
Fixes #34974 - Add & in enable_dynflow db seed to prevent undefined method update!

chris1984:


db/seeds.d/107-enable_dynflow.rb
```diff
@@ -4,5 +4,5 @@
 # !!! PLEASE KEEP THIS SCRIPT IDEMPOTENT !!!
 #
 User.as(::User.anonymous_api_admin.login) do
-  Setting.find_by(:name => "dynflow_enable_console").update!(:value => true) if Rails.env.development?
+  Setting.find_by(:name => "dynflow_enable_console")&.update!(:value => true) if Rails.env.development?
```
lfu, line 7: When would this setting not exist?

chris1984: @lfu I have no idea @jeremylenz told me to do this since it's used in other places and blocking the katello-devel boxes

chris1984: I don't even know what `&` does lol

jeremylenz: > When would this setting not exist?

Apparently it doesn't exist when building a new dev box :shrug:

parthaa, line 7:
```suggestion
    Setting["dynflow_enable_console"] = true if Rails.env.development? && !Setting["dynflow_enable_console"]
```

parthaa (approved):
APJ

chris1984:
[test katello]

db/seeds.d/107-enable_dynflow.rb
parthaa: I noticed the following behavior on my dev box
```ruby
[1] pry(main)> Setting['dynflow_enable_console']
=> true
[2] pry(main)> Setting.find_by(:name => "dynflow_enable_console")
=> nil
```

looking at https://github.com/theforeman/foreman/blob/develop/app/models/setting.rb#L83-L90 foreman is using a different mechanism.  My guess is `Foreman.settings` is yet to be run when the seed script gets run. So the suggested fix tries to do `Setting["dynflow_enable_console"] = true` if its not already set.

lfu: `dynflow_enable_console` is not a Katello setting. Why should we add it here if its owner can take care of it?

lfu: When foreman-tasks is loaded, I expect `dynflow_enable_console` setting is set to `true` no matter if it's product or development environment. So this file may be out of date already. 
Could we try remove this file and set up the devel environment then check if it works?

jeremylenz: Sounds good to me. @chris1984 do you mind trying this?

PR #10136, opened 2022-05-27
Fixes #34983 - Enable RH Repos on cdn redirect

parthaa:
Co-authored-by: Joniel Pasqualetto <jpasqualetto@gmail.com>

If the redhat cdn url redirects, the redirected CDN path is not followed when scanning and enabling repositories. This commit tries to address that.

Needed to make sure that redirect of a redirect will also work.

In your cdn configuration
1. Modify the CDN url to http://cdn.redhat.com/ (note the HTTP, not HTTPS)
2. Go to Content -> Red Hat Repositories. Try expanding RHEL 7 Server/Optional (can be any repo with substitutions).

Master:
 Nothing will be loaded.

PR:
Repos are shown  and you should be able to see entries like this in the dev logs

```
2022-05-27T20:13:56 [I|app|] CDN: Requesting path http://cdn.redhat.com:80/content/dist/rhel/server/7/listing
2022-05-27T20:14:03 [I|app|] CDN: Redirecting to https://cdn.redhat.com/content/dist/rhel/server/7/listing
2022-05-27T20:14:03 [I|app|] CDN: Requesting path https://cdn.redhat.com/content/dist/rhel/server/7/listing
```

### DIFF
--- a/app/lib/katello/resources/cdn.rb
+++ b/app/lib/katello/resources/cdn.rb
@@ -86,11 +86,11 @@ module Katello
           ::HttpProxy.default_global_content_proxy
         end
 
-        def http_downloader
-          net = net_http_class.new(@uri.host, @uri.port)
-          net.use_ssl = @uri.is_a?(URI::HTTPS)
+        def http_downloader(uri = @uri)
+          net = net_http_class.new(uri.host, uri.port)
+          net.use_ssl = uri.is_a?(URI::HTTPS)
 
-          if @uri.is_a?(URI::HTTPS)
+          if uri.is_a?(URI::HTTPS)
             net.cert = @options[:ssl_client_cert]
             net.key = @options[:ssl_client_key]
             net.ca_file = @options[:ssl_ca_file]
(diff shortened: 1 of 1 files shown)


app/lib/katello/resources/cdn.rb
```diff
@@ -137,6 +144,9 @@ def get(path, _headers = {})
               code = res.code.to_i
               if code == 200
                 return res.body
+              elsif [301, 302].include?(code) && res.header.key?('location')
+                Rails.logger.info "CDN: Redirecting to #{res.header['location']}"
+                return get(res.header['location'])
               else
```
sjha4, line 150: 👀

jpasqualetto:
Hi, I'm the one who started this and after some extra troubleshooting I agree that it's better to update the CDN address. However, some considerations:

1. There's no clear error on logs telling what is the address the CDN is redirecting them to (it takes manual steps outside foreman to figure it out). 
2. Simply modifying the CDN address only works if we also modify the option :redhat_repository_url: (on katello.yaml) to match the new CDN address. While this can be done manually, there's no option to do that in a permanent way. So this would required an issue for foreman-installer.

parthaa:
> Hi, I'm the one who started this and after some extra troubleshooting I agree that it's better to update the CDN address. However, some considerations:
> 
> 1. There's no clear error on logs telling what is the address the CDN is redirecting them to (it takes manual steps outside foreman to figure it out).
> 2. Simply modifying the CDN address only works if we also modify the option :redhat_repository_url: (on katello.yaml) to match the new CDN address. While this can be done manually, there's no option to do that in a permanent way. So this would required an issue for foreman-installer.

Agree going to close this PR. Lets do an installer change instead.

PR #10137, opened 2022-05-27
Fixes #34984 - ActionController::BadRequest when uploading RPMs via Hammer

ianballou:
Stops parameter encoding when uploading binary content (RPMs).

We're not sure of the change that occurred to cause this, but the issue is blocking nightly so this seems like the safest option.

Upload an RPM to a repository via Hammer.


sjha4 (approved):
Thoughts about adding a reference to https://github.com/rails/rails/issues/42278#issuecomment-918079123 in case we need to change this on a future rails upgrade.

Ack 👍🏼

sjha4 (approved):

PR #10140, opened 2022-05-31
Fixes #34993 - Update Recommended Repositories

wbclark:
This commit adds the Satellite Utils repositories for Satellite 6.11 on
RHEL7 and RHEL8 to the Recommended Repositories. It also adds the RHEL9
Satellite Client repository.

NOTE: Although the redmine/BZ also reference Puppet Client repositories, at this time the puppet agent will continue to ship within Satellite Client repos instead.


chris1984 (approved):
ACK, repo names match up with the ones on the product repo sheet from Frank

PR #10143, opened 2022-06-02
Fixes #35004: Drop use of pulp_client_cert and pulp_client_key

ehelms:
Foreman 3.0 introduced use of Foreman client certificates to talk to
Pulp 3 and therefore the older Pulp 2 based client certificates are
no longer needed.

See also:

 - https://github.com/theforeman/puppet-katello/pull/450
 - https://github.com/theforeman/puppet-certs/pull/399

### DIFF
--- a/app/services/katello/pulp/server.rb
+++ b/app/services/katello/pulp/server.rb
@@ -17,8 +17,8 @@ module Katello
             :debug => true
           },
           :cert_auth => {
-            :ssl_client_cert => ::Cert::Certs.ssl_client_cert(use_admin_as_cn_pulp_cert: true),
-            :ssl_client_key => ::Cert::Certs.ssl_client_key(use_admin_as_cn_pulp_cert: true)
+            :ssl_client_cert => '/etc/pki/katello/certs/pulp-client.crt',
+            :ssl_client_key => '/etc/pki/katello/private/pulp-client.key'
           }
         }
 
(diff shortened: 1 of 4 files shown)


app/services/katello/pulp/server.rb
```diff
@@ -17,8 +17,8 @@ def self.config(url, user_remote_id)
             :debug => true
           },
           :cert_auth => {
-            :ssl_client_cert => ::Cert::Certs.ssl_client_cert(use_admin_as_cn_pulp_cert: true),
-            :ssl_client_key => ::Cert::Certs.ssl_client_key(use_admin_as_cn_pulp_cert: true)
+            :ssl_client_cert => '/etc/pki/katello/certs/pulp-client.crt',
+            :ssl_client_key => '/etc/pki/katello/private/pulp-client.key'
```
ehelms, line 21: I hard coded this for now to allow removal of the setting, the real change here is to drop all of the Pulp 2 related code paths from the code base and that seemed a more monumental task. I think this is "dead" code.

chris1984: Hardcoding is fine for now. I think it would be good though to have a redmine to reference so we can go back later on and remove the hard code.

chris1984:
Starting to test

chris1984:
@ehelms acked all 3, i'll let you do the merge because you know the order they need to go in better than i do.

app/services/katello/pulp/server.rb
ehelms: I wrote up https://projects.theforeman.org/issues/35006 -- would that suffice? How do you want it referenced?

PR #6005, opened 2016-04-28
Fixes #14873 - Repo syncs are successful when capsule is unresponsive

johnpmitsch:
When a capsule is down, repository syncs should still be successful on the main Katello server. A separate capsule sync will run if there are capsules to sync to and that task will fail if the capsule is unreachable.

To test:
set up a capsule, sync repositories to it, associate it with library env. Then shut off the capsule and sync a new repository. That task should be successful. Also promote and publish content views. These tasks should be succesful, and if there is a CapsuleGenerateAndSync task, it should fail since the capsule is unreachable. Testing a promote when the capsule is in a dev env would be helpful as well

PR #10148, opened 2022-06-06
Fixes #35025 - Use task hooks for capsule sync

sjha4:
Use after action hooks to trigger Capsule sync upon publish/promote
1. Have a smart proxy tied to Library and one more environment.
2. Publish + Promote a CV to Library and the other environment.
3. Notice logs and make sure the Capsule Sync task doesn't create logs with RecordNotFound errors like it does on master.
4. Make sure the capsule sync happens correctly.

### DIFF
--- a/app/lib/actions/katello/content_view/promote_to_environment.rb
+++ b/app/lib/actions/katello/content_view/promote_to_environment.rb
@@ -1,8 +1,8 @@
-# rubocop:disable Lint/SuppressedException
 module Actions
   module Katello
     module ContentView
       class PromoteToEnvironment < Actions::EntryAction
+        execution_plan_hooks.use :trigger_capsule_sync, :on => :success
         def plan(version, environment, description, incremental_update = false)
           history = ::Katello::ContentViewHistory.create!(:content_view_version => version, :user => ::User.current.login,
                                                           :environment => environment, :task => self.task,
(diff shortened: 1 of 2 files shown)


sjha4:
[test katello]

sjha4:
@adamruzicka : Let me know if it makes sense as far as the use of hooks for triggering the task goes..

adamruzicka:
Makes sense to me.

Just out of curiostiy, which part does raise the `::Katello::Errors::CapsuleCannotBeReached` exception that is rescued at the end of `trigger_capsule_sync`?

sjha4:
> Makes sense to me.
> 
> Just out of curiostiy, which part does raise the `::Katello::Errors::CapsuleCannotBeReached` exception that is rescued at the end of `trigger_capsule_sync`?

The `def` serves as the `begin` in this case. I refactored the code as is to keep that rescue.

adamruzicka:
Well I know that, but I don't see anything in there that would raise such exception.

sjha4:
> Well I know that, but I don't see anything in there that would raise such exception.

That was all I knew about it. 😄  
Went digging a little and it seems we added it here: https://github.com/Katello/katello/pull/6005 
It might be redundant at this point based on refactoring we have done in other code. We also have 
```
def rescue_strategy
    Dynflow::Action::Rescue::Skip if self.state == 'error'
end
```
in the capsule content sync task so I think we should be safe removing the rescue here. I will test a little more.

sjha4:
Updated to remove the unnecessary rescue block as the async task invocation itself doesn't fail for an unreachable proxy.
2 task results, bottom 3 with rescue and top 3 without rescue with same results.

Screenshot from 2022-06-07 10-00-18

adamruzicka (commented):
I don't have a katello environment at hand so I can't test it right now, but from code and Dynflow perspective this is good :+1:

sjha4:
[test katello]

tazhibaevaaliya (approved):
ACK : +1

tazhibaevaaliya:
To test, I've run through all the steps described on the PR by @sjha4. 
On master: able to see "Record not found" error on the logs
On this branch: "Record not found" error is no longer on the logs

parthaa (approved):
ACK

PR #10149, opened 2022-06-07
Fixes #34792 - Add module version to db and show in UI

chris1984:
* Added a db migration to add `version` to the `katello_available_module_streams` table
* Remove the index `katello_available_module_streams_name_stream_contex` so we can lay ours down
* Added an index https://github.com/Katello/katello/compare/master...chris1984:streams?expand=1#diff-f4d99100291d8b178a9143f229b31fee9dd0a1565161a34b17d30f34654f5c64R5

* Made sure Module stream applicabilty was not affected
* Made sure there was not any performance hits by adding an index on version

* Apply patch and run db migration
* Sync CentOS8 stream repo from anywhere or this mirror (https://mirror.steadfast.net/centos/8-stream/BaseOS/x86_64/os/)
* Register a CentOS 8 stream machine to your Katello server
* Verify you see Module streams in the new ui and now see versions instead of duplicated streams

### DIFF
--- a/app/models/katello/concerns/host_managed_extensions.rb
+++ b/app/models/katello/concerns/host_managed_extensions.rb
@@ -257,7 +257,8 @@ module Katello
         module_streams.each do |module_stream|
           stream = AvailableModuleStream.where(name: module_stream["name"],
                                                context: module_stream["context"],
-                                               stream: module_stream["stream"]).first_or_create!
+                                               stream: module_stream["stream"],
+                                               version: module_stream["version"]).first_or_create!
           streams[stream.id] = module_stream
         end
         sync_available_module_stream_associations(streams)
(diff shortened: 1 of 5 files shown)


chris1984:
> I thought we had decided to only show the latest version, and not show a Version column?
> 
> Is there a large enough use case to show all of them? I thought there wasn't..
> 
> If we do go this route, we would also have to make sure that any module actions apply to the specific version that is selected. I'd vote to just keep it simple instead.

Oh I thought you wanted all the versions lol. Would I do that sort inside of the React page or the model?

jeremylenz:
> Would I do that sort inside of the React page or the model?

Seems like it'd keep it simple if it's done on the backend. Just have the API return only the latest version. Then the JS doesn't have to worry about sorting or picking.

PR #10150, opened 2022-06-07
Fixes #35033 - Cannot update network CDN if repository already enabled

sjha4:
Fix error on updating network CDN config with enabled repos
1. Enable a few repos on downstream server.
2. Enable the same repos on upstream server.
3. Update CDN config to the downstream server.


parthaa (approved):
Works well for me

rmynar:
Patch works on my box where i encountered this issue. CDN config can be changed from both UI and CLI.

PR #10151, opened 2022-06-08
Fixes #35027 add link to clear search

tazhibaevaaliya:
Added links to clear the search in situations when there is no results to display from the search. The following tabs have search/filtering functionalities: 
1. Host Collections -> Add Host to Host Collections 
2. Content -> Packages
3. Content -> Module Streams
4. Repository Sets

The following feature would increase user experience since users would be able to clear search/filter right away from the page itself and do not need to go back to the search box to do so. 

For the corresponding tab:
1. Search for non-existing item in the search bar OR/AND filter out to have no results 
2. Click the link "Clear search" or/and "Clear filter" to go back to the default search/filter values

### DIFF
--- a/webpack/components/Table/EmptyStateMessage.js
+++ b/webpack/components/Table/EmptyStateMessage.js
@@ -1,3 +1,4 @@
+/* eslint-disable prefer-const */
 import React from 'react';
 import {
   EmptyState,
--- a/webpack/components/Table/EmptyStateMessage.js
+++ b/webpack/components/Table/EmptyStateMessage.js
@@ -130,6 +130,7 @@ EmptyStateMessage.propTypes = {
   customIcon: PropTypes.elementType,
   happy: PropTypes.bool,
   searchIsActive: PropTypes.bool,
+  resetFilters: PropTypes.oneOfType([PropTypes.func, null]).isRequired,
 };
 
 EmptyStateMessage.defaultProps = {
--- a/webpack/components/Table/EmptyStateMessage.js
+++ b/webpack/components/Table/EmptyStateMessage.js
@@ -130,7 +130,23 @@ EmptyStateMessage.propTypes = {
   customIcon: PropTypes.elementType,
   happy: PropTypes.bool,
   searchIsActive: PropTypes.bool,
-  resetFilters: PropTypes.oneOfType([PropTypes.func, null]).isRequired,
+  activeFilters: PropTypes.arrayOf(PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.arrayOf(PropTypes.string),
+  ])),
+  defaultFilters: PropTypes.arrayOf(PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.arrayOf(PropTypes.string),
+  ])),
+  // eslint-disable-next-line react/require-default-props
+  resetFilters: (props, propName) => {
+    if (props.defaultFilters?.length || props.activeFilters?.length) {
+      if (typeof props[propName] !== 'function') {
+        return new Error(`A ${propName} function is required when using activeFilters or defaultFilters`);
+      }
+    }
+    return null;
+  },
 };
 
 EmptyStateMessage.defaultProps = {
@@ -141,6 +157,8 @@ EmptyStateMessage.defaultProps = {
   customIcon: undefined,
   happy: false,
   searchIsActive: false,
+  activeFilters: [],
+  defaultFilters: [],
 };
 
 export default EmptyStateMessage;
--- a/webpack/components/Table/EmptyStateMessage.js
+++ b/webpack/components/Table/EmptyStateMessage.js
@@ -1,4 +1,3 @@
-/* eslint-disable prefer-const */
 import React from 'react';
 import {
   EmptyState,
@@ -14,6 +13,8 @@ import PropTypes from 'prop-types';
 import { translate as __ } from 'foremanReact/common/I18n';
 import { CubeIcon, ExclamationCircleIcon, SearchIcon, CheckCircleIcon } from '@patternfly/react-icons';
 import { global_danger_color_200 as dangerColor, global_success_color_100 as successColor } from '@patternfly/react-tokens';
+import { useDispatch, useSelector } from 'react-redux';
+import { selectHostDetailsClearSearch } from '../extensions/HostDetails/HostDetailsSelectors';
 
 const KatelloEmptyStateIcon = ({
   error, search, customIcon, happyIcon,
@@ -26,13 +27,15 @@ const KatelloEmptyStateIcon = ({
 };
 
 const EmptyStateMessage = ({
-  title, body, error, search, customIcon, happy, ...extraTableProps
+  title, body, error, search,
+  customIcon, happy, ...extraTableProps
 }) => {
   let emptyStateTitle = title;
   let emptyStateBody = body;
   const {
     primaryActionTitle, showPrimaryAction, showSecondaryAction,
-    secondaryActionTitle, primaryActionLink, secondaryActionLink,
+    secondaryActionTitle, primaryActionLink, secondaryActionLink, searchIsActive, resetFilters,
+    filtersAreActive, requestKey,
   } = extraTableProps;
   if (error) {
     if (error?.response?.data?.error) {
@@ -45,6 +48,21 @@ const EmptyStateMessage = ({
       emptyStateBody = error?.response?.data?.displayMessage || __('Something went wrong! Please check server logs!');
     }
   }
+  const secondaryActionText = searchIsActive ? __('Clear search') : __('Clear filters');
+  const dispatch = useDispatch();
+  const clearSearch = useSelector(selectHostDetailsClearSearch);
+  const handleClick = () => {
+    if (searchIsActive) {
+      clearSearch();
+    }
+    if (filtersAreActive) {
+      resetFilters();
+    }
+    dispatch({
+      type: `${requestKey}_REQUEST`,
+      key: requestKey,
+    });
+  };
   return (
     <Bullseye>
       <EmptyState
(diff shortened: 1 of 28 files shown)


webpack/components/extensions/HostDetails/Tabs/ModuleStreamsTab/ModuleStreamsTab.js
```diff
@@ -324,6 +330,16 @@ export const ModuleStreamsTab = () => {
...
   const activeFilters = [statusSelected, installStatusSelected];
   const defaultFilters = [MODULE_STREAM_STATUS, MODULE_STREAM_INSTALLATION_STATUS];
+
+  const filterFunctions = () => {
+    if (statusSelected !== MODULE_STREAM_STATUS) {
+      setStatusSelected(MODULE_STREAM_STATUS);
+    }
+    if (installStatusSelected !== MODULE_STREAM_INSTALLATION_STATUS) {
+      setInstallStatusSelected(MODULE_STREAM_INSTALLATION_STATUS);
+    }
```
jeremylenz, line 340: I don't think the `if` statements are needed.

```suggestion
    setStatusSelected(MODULE_STREAM_STATUS);
    setInstallStatusSelected(MODULE_STREAM_INSTALLATION_STATUS);
```

webpack/components/extensions/HostDetails/Tabs/PackagesTab/PackagesTab.js
```diff
@@ -387,6 +387,12 @@ export const PackagesTab = () => {
     </Split>
   );
 
+  const filterFunctions = () => {
```
jeremylenz, line 390:
```suggestion
  const resetFilters = () => {
```

webpack/components/extensions/HostDetails/Tabs/PackagesTab/PackagesTab.js
```diff
@@ -387,6 +387,12 @@ export const PackagesTab = () => {
...
+    if (packageStatusSelected !== PACKAGE_STATUS) {
+      setPackageStatusSelected(PACKAGE_STATUS);
+    }
+  };
```
jeremylenz, line 394:
```suggestion
  const resetFilters = () => setPackageStatusSelected(PACKAGE_STATUS);
```

webpack/components/extensions/HostDetails/Tabs/RepositorySetsTab/RepositorySetsTab.js
```diff
@@ -212,6 +218,14 @@ const RepositorySetsTab = () => {
 
   const hostDetailsStatus = useSelector(state => selectHostDetailsStatus(state));
 
+  const filterFunctions = () => {
```
jeremylenz, line 221: same comment as above

webpack/components/Table/EmptyStateMessage.js
```diff
@@ -55,11 +77,47 @@ const EmptyStateMessage = ({
...
+            <Button variant="link">
+              <a href={secondaryActionLink} style={{ textDecoration: 'none' }}>{secondaryActionTitle}</a>
+            </Button>
+          </EmptyStateSecondaryActions>
+        }
+
+        {searchIsActive && !filtersAreActive &&
+          <EmptyStateSecondaryActions>
+            <Button variant="link" onClick={clearSearch}>
+              Clear search
```
jeremylenz, line 95: These need to be translated
```suggestion
              {__('Clear search')}
```

webpack/components/Table/EmptyStateMessage.js
```diff
@@ -55,11 +77,47 @@ const EmptyStateMessage = ({
...
+            </Button>
+          </EmptyStateSecondaryActions>}
+
+        { !!filtersAreActive && !searchIsActive &&
+          <EmptyStateSecondaryActions>
+            <Button variant="link" onClick={clearFilters}>
+              Clear filters
```
jeremylenz, line 102: needs translation

webpack/components/Table/EmptyStateMessage.js
```diff
@@ -55,11 +77,47 @@ const EmptyStateMessage = ({
...
+            </Button>
+          </EmptyStateSecondaryActions>
+        }
+
+        {searchIsActive && !!filtersAreActive &&
+          <EmptyStateSecondaryActions>
+            <Button variant="link" onClick={clearFiltersAndSearch}>
+              Clear search
```
jeremylenz, line 110: needs translation

webpack/components/Table/EmptyStateMessage.js
```diff
@@ -55,11 +77,47 @@ const EmptyStateMessage = ({
...
+            </Button>
+          </EmptyStateSecondaryActions>
+        }
```
jeremylenz, line 113: This section could be simplified with something like
```js
const secondaryActionText = filtersAreActive ? __('Clear filters') : __('Clear search')
```
Then below, you can just write
```js
{secondaryActionText}
```

jeremylenz: I think it can be simplified even more, so you don't have to repeat the `<EmptyStateSecondaryActions />`, if you write a `handleClick` function that looks at `filtersAreActive` and `searchIsActive`.

jeremylenz (commented):
Thanks @tazhibaevaaliya!

Initial comments below :)

jeremylenz:
@tazhibaevaaliya also, The clear search link isn't working for me in Errata tab when I have both a search and filters active.

tazhibaevaaliya:
@jeremylenz I think Errata was not working because I've not realized that we need the same functionality for that tab too. Since the dropdowns were disabled, I assumed that I do not need to implement it there. My bad.

webpack/components/Table/EmptyStateMessage.js
tazhibaevaaliya: Ok, gotcha.

jeremylenz:
[test katello]

webpack/components/Table/EmptyStateMessage.js
```diff
@@ -44,6 +46,22 @@ const EmptyStateMessage = ({
...
+    if (searchIsActive && !filtersAreActive) {
+      updateSearchQuery('');
+    }
+    if (!!filtersAreActive && !searchIsActive) {
+      resetFilters();
+    }
+    if (searchIsActive && !!filtersAreActive) {
+      resetFilters();
+      updateSearchQuery('');
+    }
```
jeremylenz, line 62: I think this can still be simplified more.

If `searchIsActive` you always call `updateSearchQuery('');`
if `filtersAreActive` you always call `resetFilters()`

So could we rewrite it without using any `&&`?

jeremylenz (commented):
Changes look good, except:

(Maybe I didn't notice this before) When I clear a search, the "you don't have any items, period" empty-state message flashes for a single render. (It's easier to see if you go to the dev tools Performance tab and set CPU throttling to 6x).  It gives it a little bit of a wonky feel when you clear a search.

Can you find a way to not have that message flash?

One more small comment below, as well

webpack/components/Table/EmptyStateMessage.js
jeremylenz: I mean the "call-to-action" empty state completed in the last PR. That flashes on the screen.

tazhibaevaaliya: Would you mind hopping on to the quick call after our standup since I still did not get what you mean? On my end, it seems that nothing is flashing.

jeremylenz:
[test katello]

jeremylenz:
[test katello]

jeremylenz (approved):
Thanks for hanging in there!

APJ :+1:

sjha4:
Can you test it on CV repositories tab? I seem to be hitting :

image

jeremylenz:
The CV Repositories page issue I see on master. However, I do see this similar error on Host details:
cleanup_error

To reproduce:

1. Clear a search on one tab
2. switch to another tab
3. attempt to clear another search.

If I add a cleanup function to `TypeAhead.js` like this, it seems to fix it:

```js
  // eslint-disable-next-line arrow-body-style
  useEffect(() => {
    return function cleanupClearSearch() {
      dispatch({
        type: 'SET_CLEAR_SEARCH',
        payload: {},
      });
    };
  }, [dispatch]);
```

jeremylenz (changes requested):
temporarily rescinding my acceptance (but I think I fixed it! see above)

jeremylenz:
The "Clear filters" link isn't working for me on the Content view details > Repositories page.

This is because there isn't a `resetFilters` function defined in the CV pages.

I feel like `resetFilters` should be a required prop on EmptyStateMessage.. Then run the CV tests and see where we need to add it.

tazhibaevaaliya:
@jeremylenz, makes sense! I did not know that I need to add to other pages besides HostDetails pages. Thanks for the early heads-up :smiling_face_with_tear:

tazhibaevaaliya:
@jeremylenz, where else does this link need to work? How about tabs like "Monitor" or "Infrastructure"?

jeremylenz:
> where else does this link need to work?

It needs to work anywhere `<EmptyStateMessage>` is used. Currently that's in Content Views and new host details pages.

If you make it a required prop and then run the tests I think you should get a good idea of where it's needed..

jeremylenz:
[test katello]

jeremylenz:
[test katello]

jeremylenz (approved):
Clicked around both host details and CV pages. Everything works and saw no console errors.

Great work @tazhibaevaaliya! 

@sjha4 want to take a final look?

tazhibaevaaliya:
@sjha4  made the changes accordingly. Would you mind taking a look? Thanks!

jeremylenz:
[test katello]

sjha4 (approved):
Works great for CV pages. Nice work @tazhibaevaaliya 🎉

PR #10154, opened 2022-06-08
Generate package-lock and run npm ci separately for react UI tests

ehelms:
This is just an idea to get better granular data on the time each step takes for tests to later inform ways we might speed up tests. I do not think this should increase test runtime (and locally it had small improvements).


chris1984 (approved):
Makes sense, ACK

ehelms:
[test katello]

ehelms:
30 minute run of the Katello Jenkins tests! @evgeni smooookkinnnnn

evgeni:
I've been called worse things!

😽

PR #10156, opened 2022-06-08
Fixes #35039 - use correct slot for recent communication card item

jeremylenz:

The 'Last check-in' card extension was using the slot `host-overview-cards` when it should have been using `recent-communication-card-item`. This causes text to be displayed outside any card.

make sure the Recent communication card is present and shows last checkin


sjha4 (approved):
😸

PR #10158, opened 2022-06-09
use a normal version of grunt-bower-task, not a git checkout

evgeni:
the git checkout takes ~6 minutes to fetch and inspect on our CI nodes
and has exactly zero benefit.

This was introduced in
https://github.com/Katello/bastion/commit/db728a7cd1cf53572e06b5c53f2855742b350117


evgeni:
before:
image

after:
image

:kissing_cat:

chris1984 (approved):
Makes sense to me

PR #10162, opened 2022-06-14
Fixes #34992 - Use content ID to verify if a redhat_repo already exists

parthaa:
Using labels to verify if a red hat repository already exists create issues when
labels differ between exporting and importing servers. Keep the
validation to use labels as fallback for custom products, which doesn't keep the
same content id between importing and exporting server and for the
case where the export didn't include content_id

Co-authored-by: Joniel Pasqualetto <jpasqualetto@gmail.com>

Instead of using repository labels to verify if a repository being import already exists, this PR uses content ID.

Simulate a label mismatch between exporting and importing servers
Export content once and import it (this works fine)
Export content again (incremental export or a simple new complete export) and try importing again. Without this PR, the second import fails. With this patch it should work.


chris1984 (approved):
ACK, tested on both RH repos with changing the label and got both versions to import correctly with the patch compared to without the patch and it not importing the RH repo into the published view. 

Tried with a custom repo, changed the label to `-XXX` and with the patch it imported correctly. 

Nice work ALL!!!!

PR #10164, opened 2022-06-14
Fixes #34804 - Fix upload debian package with invalid relative_path error

Manisha15:
Fixes following error when uploading certain debian packages:
``Error during upload: Katello::Errors::Pulp3Error: {'non_field_errors': [ErrorDetail(string='Invalid relative_path provided, filename does not match.', code='invalid')]} ``

### DIFF
--- a/app/services/katello/pulp3/deb.rb
+++ b/app/services/katello/pulp3/deb.rb
@@ -9,6 +9,7 @@ module Katello
       end
 
       def self.content_api_create(opts = {})
+        opts.delete(:relative_path) if opts.key?(:relative_path)
         self.content_api.create(opts)
       end
 
--- a/app/services/katello/pulp3/pulp_content_unit.rb
+++ b/app/services/katello/pulp3/pulp_content_unit.rb
@@ -146,6 +146,8 @@ module Katello
         elsif repository.generic?
           duplicate_sha_path_content_list = content_backend_service.content_api(repository.repository_type, unit_type_id).list(
             filter_label => checksum)
+        elsif unit_type_id == 'deb'
+          duplicate_sha_path_content_list = content_backend_service.content_api.list(filter_label => checksum)
         else
           duplicate_sha_path_content_list = content_backend_service.content_api.list(
             filter_label => checksum,


app/lib/actions/pulp3/repository/save_artifact.rb
```diff
@@ -26,10 +26,16 @@ def invoke_external_task
...
               []
             else
-              output[:pulp_tasks] = [content_backend_service.content_api_create(relative_path: input[:options][:file_name],
-                                                                                repository_id: repository.id,
-                                                                                artifact: artifact_href,
-                                                                                content_type: content_type)]
+              if content_type == 'deb'
+                output[:pulp_tasks] = [content_backend_service.content_api_create(repository_id: repository.id,
+                                                                                  artifact: artifact_href,
+                                                                                  content_type: content_type)]
```
sjha4, line 32: Could you make this change in https://github.com/katello/katello/blob/master/app/services/katello/pulp3/deb.rb#L11 and drop the relative_path key from params instead of the generic save_artifact method to avoid the if-else here...

Manisha15: Added the changes.

sjha4:
Does this need to be fixed: https://projects.theforeman.org/issues/35060

Manisha15:
> Does this need to be fixed together with this: https://projects.theforeman.org/issues/35060 ?

I discovered this error (#35060) while testing this PR (#34804). Issue #35060 happened somewhere in Katello 4.4.x and its unrelated to this PR.

sjha4:
I think it would make sense to fix it here as the uploaded package shows up with null values in a lot of fields.
Also we will likely need to add support for uploading duplicate content. We do that for the other content types.

On this branch, trying to do that results in:

ForemanTasks::TaskError: Task c159011e-31eb-48ca-a5b7-0319119ede66: Katello::Errors::Pulp3Error: {'non_field_errors': [ErrorDetail(string="There is already a deb package with relative path 'pool/g/google-chrome-stable/google-chrome-stable_100.0.4896.127-1_amd64.deb' and sha256 'd012bc3e5e000bffbd83cf20c48add187c52b065cc37408862c42d127aa4a656'.", code='invalid')]}

quba42:
@sjha4 How do you do this for the other content types? If there is already a mechanism for this why is deb not using it? Is this something where each Pulp plugin is different?

Note that I told https://github.com/hstct to look into if it makes sense to fix this on the pulp_deb side, by making the logic there more "create or retrieve" like. However, I only want to do this if it brings the pulp_deb workflow more in line with other Pulp plugins. If this is the same for all Pulp plugins, then there should be a single mechanism on the Katello side, that works for all Pulp plugins.

sjha4:
> @sjha4 How do you do this for the other content types? If there is already a mechanism for this why is deb not using it? Is this something where each Pulp plugin is different?
> 
> Note that I told https://github.com/hstct to look into if it makes sense to fix this on the pulp_deb side, by making the logic there more "create or retrieve" like. However, I only want to do this if it brings the pulp_deb workflow more in line with other Pulp plugins. If this is the same for all Pulp plugins, then there should be a single mechanism on the Katello side, that works for all Pulp plugins.

We do have a duplicate check for all content types: https://github.com/Katello/katello/blob/master/app/services/katello/pulp3/pulp_content_unit.rb#L129

This should execute for debian uploads. It might have something to do with https://projects.theforeman.org/issues/35060 ?

Manisha15:
I have added code for duplicate code as well in this PR. For GUI issue of packages not visible after upload, it can be worked on another PR as its no related to deb package upload but something created this bug in Katello 4.4+.

quba42:
@sjha4 To me this fix is separate from the [display issue](https://projects.theforeman.org/issues/35060). The issues fixed by this PR also affects older Katello versions, whereas the display issue only started with Katello 4.4. I would like to address them separately, so that this fix could in principle be back-ported to Katello 4.3.

Also, I don't think we have any good ideas how to approach the display issue. Perhaps we can try to find the change in Katello 4.4 that broke it, and start there?

sjha4:
Deb upload with pulp3 hasn't been in any previous releases, has it? Since the empty deb package is only tied to uploaded packages, it would make sense to test them both together?

quba42:
> Deb upload with pulp3 hasn't been in any previous releases, has it?

I can upload deb packages on Katello 4.1 and 4.3. Just not any deb packages with the "wrong" filename, or repeat packages. All of that is (as far as I can tell) fixed by the current state of this PR.

In addition there is an issue starting with Katello 4.4, that uploaded packages are not displayed (other than as an empty line not containing any values). To me this seems like a separate issue introduced in 4.4, for something that was working up to 4.3. That being said, we could test them both together (if we had a fix for the display thing). I have no idea what caused or how to fix the display thing though.

sjha4:
> > Deb upload with pulp3 hasn't been in any previous releases, has it?
> 
> I can upload deb packages on Katello 4.1 and 4.3. Just not any deb packages with the "wrong" filename, or repeat packages. All of that is (as far as I can tell) fixed by the current state of this PR.
> 
> In addition there is an issue starting with Katello 4.4, that uploaded packages are not displayed (other than as an empty line not containing any values). To me this seems like a separate issue introduced in 4.4, for something that was working up to 4.3. That being said, we could test them both together (if we had a fix for the display thing). I have no idea what caused or how to fix the display thing though.

I am looking at that issue now. Last I checked briefly, pulp is storing the content correctly but katello isn't indexing those fields correctly. I will take a look.

sjha4:
Created a PR which fixes a (silly) issue with uploaded deb content indexing. Please test that and let me know if it looks good.
https://github.com/Katello/katello/pull/10190

Manisha15:
> Created a PR which fixes a (silly) issue with uploaded deb content indexing. Please test that and let me know if it looks good. #10190

I tested it with latest Katello and it fixes the issue. Thank you.

sjha4:
This PR looks good. I will ack and merge this once the other PR is merged.

PR #10170, opened 2022-06-17
Fixes #35079 - Pulp 3 migration stats timing is too low for very large deployments

ianballou:
There was an extra `% 60` in the timing estimate for the hours calculation. This PR removes that.

This isn't noticeable if the calculated migration hours are under 60. However, since the "slow" migration calculation multiplies it by 4, the calculation error will be more noticeable there.

None really.

1) Either create enough dummy RPM / repository records to get the `migration_minutes` up above 3600, or copy and paste the relevant code into a method and hack in the RPM and repository values.

2) Try varying values below 3600 just to make sure everything makes sense.  Compare the results with an online minutes to hours:minutes calculator online.


sjha4 (approved):
Code change makes sense. The `% 60` in the old code would only work for hours under 60 and give false hopes to users with > 60 hours of migrations..😸

sjha4:
[test katello]

ianballou:
[test katello]

PR #10172, opened 2022-06-17
Fixes #34336 - CRUD simplified ACSs via the API

ianballou:
Adds support for Simplified Alternate Content Sources. Users simply need to select which products they want to be included in an ACS and Katello will automatically maintain ACSs for every related repository (of the chosen content type) on all the selected smart proxies.  If repositories are created or destroyed, new ACSs will be created.

I'm updating hammer alongside to include the Products in the show command and also to reduce the amount of smart proxy parameters shown (https://github.com/Katello/hammer-cli-katello/pull/853).

I'm pretty set on keeping the content type requirement for simplified ACSs. Alternatively, we could just create ACSs for all supported types, but that makes things a bit complicated. Plus I think users will like having control over ignoring file repos, for example.

1) Create a couple smart proxies and a couple products
2) Create varying simplified smart proxies with different combinations of products and smart proxies
3) Try updating the simplified smart proxies in various ways, removing/adding smart proxies and products, etc
4) Delete simplified ACSs
5) Ensure that hammer works by testing this: https://github.com/Katello/hammer-cli-katello/pull/853
6) Refresh simplified ACSs and ensure that all the child ACSs were refreshed by checking in Dynflow and/or Pulp
7) Ensure that creating custom ACSs still works.

TODO:
- [x] Tests
- [x] Look into simplifying `app/lib/actions/katello/alternate_content_source/update.rb`
- [x] Audit simplified ACS products
- [x] Validations for fields like product and base url to ensure they don't exist for the wrong type
- [x] Check that HTTP proxy info from repositories isn't copied to ACS remotes.
- [x] Remove `product-names` param since there's no organizational scoping.
- [x] Don't create ACSs for repositories that don't have upstream URLs.


evgeni:
[test katello]

(sorry for the noise, I need to test something)

sjha4:
Does it make sense to audit products assigned to ACS..could be done as part of this.

ianballou:
> Does it make sense to audit products assigned to ACS..could be done as part of this.

I don't see why not! I'll add a TODO.

sjha4:
Do we need some validations around type and fields that can be set? For example custom can't have products and simplified can't have url and subpaths. Does that make sense?

sjha4:
I am confused about where the URL for the remote is coming for the ACS in this case. If I am not misunderstanding the code, it is going to be same as repository upstream URL? So a separate remote for each ACS smart proxy repository which is equal to repository upstream URL..Is that accurate?

ianballou:
> Do we need some validations around type and fields that can be set? For example custom can't have products and simplified can't have url and subpaths. Does that make sense?

Yeah, I could take a look at that.



> If I am not misunderstanding the code, it is going to be same as repository upstream URL?

Yep, exactly the same remote options as the repository minus the download policy.  The http proxy should be different too,  I need to check that bit.



> So a separate remote for each ACS smart proxy repository which is equal to repository upstream URL..Is that accurate?

Yep, the remotes should look the same for a specific repository across different smart proxies.  For each product in an ACS we create smart_proxy_acs records (remote+ACS in Pulp) for each repository in every associated smart proxy.

ianballou:
The audits are giving me a hard time. If I add `products` to the audit associations, nothing is logged. If I audit `alternate_content_source_products`, I can see when products are associated, but for some reason not when they are unassociated. It's really weird.

ianballou:
@sjha4 I added the audits for products. I had to write a custom audit when the products get updated by the update task.

ianballou:
@sjha4 I believe I've made all the necessary updates.

sjha4:
[test katello]

sjha4:
[test katello]

app/lib/actions/katello/repository/create.rb
```diff
@@ -33,6 +34,12 @@ def plan(repository, args = {})
...
               plan_self(:repository_id => repository.id, :clone => clone)
+              repository.product.alternate_content_sources.with_type(repository.content_type).each do |acs|
+                acs.smart_proxies.each do |smart_proxy|
+                  smart_proxy_acs = ::Katello::SmartProxyAlternateContentSource.create(alternate_content_source_id: acs.id, smart_proxy_id: smart_proxy.id, repository_id: repository.id)
+                  plan_action(Pulp3::Orchestration::AlternateContentSource::Create, smart_proxy_acs)
+                end
+              end
             end
           end
         end
```
sjha4, line 45: This doesn't behave as intended if the ACS was created with an empty product and repos added later to it.

app/views/katello/api/v2/alternate_content_sources/base.json.rabl
```diff
@@ -2,18 +2,23 @@ object @resource
 
 extends 'katello/api/v2/common/identifier'
 
-attributes :name, :base_url, :subpaths, :content_type, :alternate_content_source_type, :smart_proxies, :upstream_username, :verify_ssl
-
-child :ssl_ca_cert => :ssl_ca_cert do |_object|
-  attributes :id, :name
-end
+if @resource.respond_to?(:alternate_content_source_type)
```
sjha4, line 5: @resource is null here for some reason so the index doesn't return all expected field. One workaround I tested locally is to not use object @resource and reference @object instead of @resource in the file.

app/lib/actions/pulp3/alternate_content_source/create_remote.rb
```diff
@@ -2,13 +2,14 @@ module Actions
...
+        def plan(smart_proxy_acs)
+          plan_self(smart_proxy_id: smart_proxy_acs.smart_proxy_id, smart_proxy_acs_id: smart_proxy_acs.id)
         end
 
         def run
-          acs = ::Katello::AlternateContentSource.find(input[:acs_id])
-          output[:response] = acs.backend_service(smart_proxy).create_remote
+          smart_proxy_acs_id = input[:smart_proxy_acs_id]
+          smart_proxy_acs = ::Katello::SmartProxyAlternateContentSource.find(smart_proxy_acs_id)
+          output[:response] = smart_proxy_acs.backend_service.create_remote
```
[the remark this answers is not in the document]
ianballou: Ah, if the repository doesn't have a URL then an ACS shouldn't be created. I'll update that.

app/lib/actions/katello/repository/create.rb
ianballou: I see why this is happening. With a product that has no repositories, no smart_proxy_alternate_content_sources get created. So, `acs.smart_proxies` returns `[]`.

ianballou:
@sjha4 this is ready for another round of testing! My latest updates are in the last commit.

sjha4:
Probably a pulp bug..I ran into some OOMs on my 12G memory VM trying to refresh a remote with some large repos. (Rhel8 base, rhel8 appstream, rhel9 base) Logs: http://pastebin.test.redhat.com/1066415

The rubocop failure is relevant..Looking good otherwise..

app/lib/actions/katello/repository/create.rb
```diff
@@ -33,6 +34,12 @@ def plan(repository, args = {})
 
             concurrence do
               plan_self(:repository_id => repository.id, :clone => clone)
+              repository.product.alternate_content_sources.with_type(repository.content_type).each do |acs|
+                acs.smart_proxies.each do |smart_proxy|
+                  smart_proxy_acs = ::Katello::SmartProxyAlternateContentSource.create(alternate_content_source_id: acs.id, smart_proxy_id: smart_proxy.id, repository_id: repository.id)
+                  plan_action(Pulp3::Orchestration::AlternateContentSource::Create, smart_proxy_acs)
```
sjha4, line 40: We should verify existence of repo.url here as well before creating ACSs for new repos.

ianballou:
@sjha4 I've pushed my changes

app/lib/katello/validators/alternate_content_source_products_validator.rb
```diff
@@ -0,0 +1,17 @@
+module Katello
+  module Validators
+    class AlternateContentSourceProductsValidator < ActiveModel::EachValidator
+      def validate_each(record, attribute, value)
+        if value
+          if attribute == :product_id
+            product = ::Katello::Product.find(value)
+            content_type = record.alternate_content_source.content_type
+            if product.repositories.with_type(content_type).empty?
+              record.errors[attribute] << N_("The product %s has no repositories to add to the alternate content source.") % product.name
```
sjha4, line 10:
```suggestion
              record.errors[attribute] << N_("The product %s has no %s repositories to add to the alternate content source.") % [product.name, content_type]
```

ianballou:
TODO: check that products without any usable repositories (including those without URLs) cannot be added to ACSs.

ianballou:
@sjha4 my latest commit has the changes necessary to ensure that ACS products have repositories with URLs. I added some tests too.

sjha4:
One workflow to support would be to allow updating repository to clear out URL if it's been added to an ACS. Currently it doesn't seem to be supported. Task error below: 

image

image

ianballou:
@sjha4 I fixed the URL removal issue.

ianballou:
TODO: ACSs need to be refreshed at the same time as repositories during manifest actions. Otherwise, entitlement certificate info on the Pulp remotes can get out of date.

sjha4:
> TODO: ACSs need to be refreshed at the same time as repositories during manifest actions. Otherwise, entitlement certificate info on the Pulp remotes can get out of date.

Are you planning to add this to this PR? I haven't tested this workflow but everything else looks good.

ianballou:
> > TODO: ACSs need to be refreshed at the same time as repositories during manifest actions. Otherwise, entitlement certificate info on the Pulp remotes can get out of date.
> 
> Are you planning to add this to this PR? I haven't tested this workflow but everything else looks good.

I just pushed up that change in a new commit.

sjha4 (approved):
Tested and works well.. Can see the ACS remote refresh task as part of manifest refresh.

APJ 👍🏼

sjha4:
[test katello]

PR #10173, opened 2022-06-21
Fixes #35063 - Allow copying activation keys with create_activation_keys permissions

sjha4:
Create a user with view and create activation keys permissions.
Go to Activation key > your activation_key > 
1. You should be able to see a copy action in the action dropdown
2. You should be able to successfully copy the key without any permissions issues.


sjha4:
> With only permissions for activation keys, got this error when clicking Content - Activation Keys. Screenshot from 2022-06-21 15-35-48

Was this on the branch or master? I can click on Content > Activation keys in master and on this branch. I added all activation key permissions, edit/create/view and create Activation keys to the role.

lfu:
> Was this on the branch or master? 

This happens when Any Organization is used and I think it is by default. 
After selecting a specific organization, the issue didn't occur.

lfu (approved):
I think the issue I hit is not related to this PR.
LGTM 👍🏻

lfu:
The routing issue was caused by missing permissions to view and assign organizations for the limited user.

PR #10176, opened 2022-06-27
Refs #35006 - Remove unused Pulp 2 classes

ekohl:

This is unfinished, but I wanted to share. My goal was to remove runcible as a dependency, but this turned out to be larger than I thought.

I tried to look at the source and git log to see when it was unused. It may be possible to squash a bunch of these commits and there are many more Pulp 2 actions that can be removed.

I do not know if just removing an action works, or if it also needs a DB migration to purge it from foreman-tasks.

Test suite should pass, but the suite may not cover everything.


ekohl:
This still leaves parts around, like `Katello.pulp_server` is still used in a bunch of places. I have a strong suspicion that `PulpTaskStatus` and related friends are also unused. Also the RegistryProxiesController looks suspicious.

ekohl:
I think this is now fairly complete. There are 3 `[INCOMPLETE]` commits which have the TODO points in their commit messages. Especially the last one is interesting because it appears there is actively some Pulp 2 connection work going on but I don't know how to remove that without taking an axe to it and kill everything.

As for the other commits, they typically refer to some other commit which removed the only use of it. That means large portions of the code were already unused for years.

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -243,7 +243,8 @@ def push_manifest
     end
 
     def pulp_content
-      Katello.pulp_server.resources.content
+      # TODO: Remove this
+      fail 'Pulp 2 content was being used'
```
ekohl, line 247: How to proceed with this? I don't know enough about this controller to say how to replace this.

app/lib/actions/katello/repository/destroy_package_group.rb
```diff
@@ -7,7 +7,8 @@ def plan(repository, pkg_group_id)
           criteria = { type_ids: ["package_group"], filters: {"unit": {"name": pkg_group_id} } }
 
           sequence do
-            ::Katello.pulp_server.extensions.repository.unassociate_units(repository.pulp_id, criteria)
+            # TODO: Replace with Pulp 3?
+            fail 'Pulp 2 was used'
             plan_action(IndexPackageGroups, repository)
             plan_action(Katello::Repository::MetadataGenerate, repository)
```
ekohl, line 13: Rubocop failures are here, but really the question is whether this action is still used. It's called here from a REST API controller which suggests it can still be called by users:
https://github.com/Katello/katello/blob/b056c7960fe99f709cb225c68cd1a321ec105840/app/controllers/katello/api/v2/package_groups_controller.rb#L54

app/controllers/katello/api/registry/registry_proxies_controller.rb
ianballou: This code relates to tagging manifests and uploading manifests via Pulp 2. We already have a different code path for tagging manifests with Pulp 3 (https://github.com/Katello/katello/pull/9054/files), but no way to upload manifests. I actually already have an issue for this: https://projects.theforeman.org/issues/33767

For the sake of this PR, I'd say you can rip all the code out relating to uploading manifests. So that would be `pulp_content`, `upload_manifest`, and `upload_tag`. That also means we'll need to remove mention of uploading docker manifests from other bits of the code:

- https://github.com/Katello/katello/blob/master/app/lib/actions/katello/repository/import_upload.rb#L82-L91
- Some of the controllers mention uploading manifests. Those references can be removed.
- https://github.com/Katello/katello/blob/master/lib/katello/repository_types/docker.rb#L30

ianballou: Note that we do still support removing manifests.

app/lib/actions/katello/repository/destroy_package_group.rb
ianballou: We don't support deleting package groups from the UI or Hammer, so I think it's safe to just drop it.

Re-implementing it would be relatively easy, but it would need to be tested and such. We'd use the remove_content API endpoint for repositories. The related tasks can accept any sort of content unit, but package groups would need to be added as an acceptable type.

app/models/katello/glue/pulp/repo.rb
```diff
@@ -70,166 +29,32 @@ def initialize(attrs = nil)
       end
 
       def srpm_count
-        pulp_repo_facts['content_unit_counts']['srpm']
+        # TODO: Pulp 3 replacement?
```
ianballou, line 32: No need to replace, you can get rid of it entirely.

test/controllers/api/registry/registry_proxies_controller_test.rb
```diff
@@ -589,11 +589,7 @@ def setup_permissions
     end
 
     def mock_pulp_server(content_hash)
-      content = mock
-      content_hash.each do |method|
-        content.stubs(method[:name]).times(method[:count]).returns(method[:result])
-      end
-      @controller.stubs(:pulp_content).returns(content)
+      # TODO: this breaks
```
ianballou, line 592: This and the manifest upload  test can be removed.

ianballou (commented):
Added some comments.

Also, about the PulpSyncStatus and related classes, they all look so old and Pulp 2 specific that I think they could all be pulled out. I know we mentioned before that the sync status page might use it, but I'm starting to doubt that. These classes must be completely broken by now.

app/models/katello/glue/pulp/repo.rb
ekohl: The main reason I didn't do this yet was that I then also needed to make UI adjustments and I wasn't ready for that yet.

ianballou:
Hey @ekohl, I wanted to check if you're planning on getting back to this.  We could always work on top of what you've started and put it back on the Katello dev backlog.

ekohl:
I submitted https://github.com/Katello/katello/pull/10215 as a first step. I took all the commits up until the first incomplete one and then squashed a bunch of them that were all fixing up the same original commit.

It's my proposal to first get that merged and then iterate on it further.

app/models/katello/glue/pulp/repo.rb
ianballou: You shouldn't have to worry about UI changes regarding any of the counts in this file.  The `srpm_counts` mentioned in the UI refer to a content view versions `content_counts`.  The code where we dynamically load these counts (which is why it's not grep-able) is here: https://github.com/Katello/katello/blob/master/app/models/katello/content_view_version.rb#L333

The `content_counts` get passed to the API through the method `content_counts_map`: https://github.com/Katello/katello/blob/master/app/views/katello/api/v2/content_view_versions/base.json.rabl#L13

ianballou:
Sure, we can focus on your https://github.com/Katello/katello/pull/10215 PR first.

ekohl:
I think you included all commits in https://github.com/Katello/katello/pull/10400, so closing this.

PR #10177, opened 2022-06-27
Fixes #35119 - acs refresh rake task fails with no ACSs

ianballou:
Adds a check for existing ACSs in the refresh ACS rake task.

Run `rake katello:refresh_alternate_content_sources` with and without ACSs in the DB.


chris1984 (approved):
ACK, works well:

With sources it worked and i get the message below when i don't have any.

```ruby
2022-06-28T11:26:08 [W|app|] Scoped order is ignored, it's forced to be batch order.
No alternate content sources to refresh.
```

PR #10181, opened 2022-06-30
Fixes #35091 - Python content isn't synced to smart proxies

ianballou:
Updates pulp-python to 3.7.1.  Also adds a fix where updating Python repositories fails because nil isn't accepted for the update api.

I considered just calculating the packages to include, but that would be difficult to do in a generic way (and Python is a generic type).

1) Sync Python content to a smart proxy. Notice that the distributed Python content has no packages.
2) Upgrade pulp-python to 3.7.1 + `bundle update`.
3) Sync again (complete).
4) See that there is Python content being served now on the smart proxy.
5) Try updating a Python repository to have empty includes, excludes, and package types.


wbclark (approved):
This works great for me.

I created a custom product and python package repository with an include filter for the package `apypie`. Then I assigned the Library environment to my content-proxy.

After syncing the repository, I could also see a capsule content sync task with the matching repository_id as input.

Then in the browser, I can see the packages on both the katello server and the content-proxy, at the URLs https://centos8-stream-wclark-katello-devel.example.com/pulp/content/Default_Organization/Library/custom/approved_python_packages/python_packages/ and https://centos8-stream-wclark-fpc-devel.example.com/pulp/content/Default_Organization/Library/custom/approved_python_packages/python_packages/

Then I tested installing packages with `pip install -i https://centos8-stream-wclark-fpc-devel.example.com/pulp/content/Default_Organization/Library/custom/approved_python_packages/python_packages/simple/ apypie`

This successfully searches the index on the content-proxy, downloads the package from content-proxy, and tries to download dependencies (that last part fails, since the include filter didn't include any dependencies)

wbclark (commented):
Other comments I had

1. Style idea (not super important, feel free to disregard)

For the `generic_remote_options_hash` method in `app/controllers/katello/api/v2/repositories_controller.rb` what do you think about shortening it a bit with this approach?

```ruby
    def generic_remote_options_hash(repo_params)
      content_type = @repository&.content_type || repo_params[:content_type]
      RepositoryTypeManager.generic_remote_options(content_type: content_type).to_h do |option|
        [option.name, repo_params[option.name].nil? ? option&.default : repo_params[option.name]]
      end
    end
```

I tested this by adding an additional package to my include on the repository, and it worked nicely.

Similar to the current approach, it does make 2-3 calls to `option.name` per option which doesn't seem costly but I guess you could optimize it further by first doing `option_name = option.name` at the top of the block and then using the stored value. /shrug

2. Also, it needs fixtures re-recorded on top of a rebase

ianballou:
@wbclark your change there makes sense to me!

ianballou:
@wbclark I pushed the changes

chris1984 (approved):
Looks good

PR #10182, opened 2022-06-30
Fixes #35078 - Add 'System purpose' card

jeremylenz:

Add 'System purpose' card to the new host details page.
- [x] tests
- [x] permissions
- [x] fix dropdown sorting
- [x] rebase

syspurpose_card
syspurpose_modal

When building options to choose from, our existing Angular page assembled them from three sources:
* the current value
* a hard-coded list of default values
* a list of available values from the organization (`org.system_purposes`)

This list was then deduped to produce the final list of options available to the user.  I'm doing the same sort of thing here.

Test the following:
1. You can change syspurpose values to any of the options, and the card will update
2. You can unset any value, and the card will show a blank value
3. No duplicate options should appear in the modal
4. You can set an arbitrary value on the host with, e.g. `subscription-manager syspurpose addons --add "Addon 2"` or `subscription-manager role --set "My Special Role"`. When a value is set like this, it should show up as the currently selected option, and as an available option in the dropdown.
5. All dropdown options should be the same as those on the legacy Content host details page (for the same host)
6. The Edit button is not present for a user without `edit_hosts` permission

(see `subscription-manager syspurpose --help` for more info)


jeremylenz:
Added tests; ready for review :heavy_check_mark:

jeremylenz:
> * How is Release Version to be used? Where it can be set?

It's used when calculating the host's bound repositories. I think it's also used for Subscription Watch reporting.  You should be able to set it

a) here on the SystemPurposeEditModal; and 
b) on the old content host detail page.

jeremylenz:
> 3\. Seems Addon can't be add from WebUI but can be done from subscription-manager. Just to confirm this is by design

Yes, Addons can come from specific subscriptions or can be manually set from the host. But not from web UI.

jeremylenz:
> Old content host page does not have Release Version

Release version is in a different location on the old page:
releasever-old-content-host

jeremylenz:
> I can't set Release Version in the new page.

I think the available release versions depend on which repos you have synced (e.g. if you sync RHEL8 you'll have "8" as an option).  If you see different options on the old and new pages, that's a bug.  If they're the same, that's what's expected.

lfu:
LGTM 👍🏻 
Screenshot from 2022-07-11 13-12-56

lfu (approved):

PR #10183, opened 2022-06-30
Fixes #34861 - Syncable yum-format repository exports

parthaa:
This commit adds the concept of `syncable` exports. 

```
$ hammer content-export complete version --format=syncable --id=81
[....................................................................................................................................................................] [100%]
Generated /var/lib/pulp/exports/Default_Organization/sample-content-view/4.0/2022-06-30T22-19-34-00-00
```

The files are exported in a different format. cd into `/var/lib/pulp/exports/Default_Organization/sample-content-view/4.0/2022-06-30T22-19-34-00-00` and you should see a directory structure that looks like the following. 
```
$ ls -l /var/lib/pulp/exports/Default_Organization/sample-content-view/4.0/2022-06-30T22-27-11-00-00/content/dist/rhel/server/7/7.5/x86_64/openstack-tools/10/os
total 24300
-rw-rw-rw-+ 1 pulp pulp   19484 Jun 30 22:27 babel-2.3.4-1.el7ost.noarch.rpm
-rw-rw-rw-+ 1 pulp pulp   71976 Jun 30 22:27 openshift-heat-templates-0.9.9-2.el7ost.noarch.rpm
-rw-rw-rw-+ 1 pulp pulp   72448 Jun 30 22:27 openshift-heat-templates-0.9.9-5.el7ost.noarch.rpm
.....
drwxrwxrwx+ 2 pulp pulp    4096 Jun 30 22:27 repodata

$ 
```

The export directory should also display generated listing files (note you need to run hammer as root for this part.)
```
$ find /var/lib/pulp/exports/Default_Organization/sample-content-view/4.0/2022-06-30T22-27-11-00-00/content/ -name listing
....
/var/lib/pulp/exports/Default_Organization/sample-content-view/4.0/2022-06-30T22-27-11-00-00/content/dist/rhel/server/7/listing
/var/lib/pulp/exports/Default_Organization/sample-content-view/4.0/2022-06-30T22-27-11-00-00/content/dist/rhel/server/listing
/var/lib/pulp/exports/Default_Organization/sample-content-view/4.0/2022-06-30T22-27-11-00-00/content/dist/rhel/listing
/var/lib/pulp/exports/Default_Organization/sample-content-view/4.0/2022-06-30T22-27-11-00-00/content/dist/listing
/var/lib/pulp/exports/Default_Organization/sample-content-view/4.0/2022-06-30T22-27-11-00-00/content/listing
....
```

This is similar to the cdn structure. The typical user would
1. Copy the contents of  `/var/lib/pulp/exports/Default_Organization/sample-content-view/4.0/2022-06-30T22-19-34-00-00` to a disconnected server 
2. Update the other server's cdn to `file:///var/lib/pulp/exports/Default_Organization/sample-content-view/4.0/2022-06-30T22-19-34-00-00`
3. Enable and Sync repos available in this archive

#### Constraints
- No `metadata.json` is generated.
- No import command available here. The disconnected user is responsible for the consistency of data.
- No incremental exports. Only complete for starts.

This will be ideally used in environments where one wants to share the exports to the world and the downstream server may be old and not having pulp3 installed. 

Pulp 3.14 and above had facilities to do a regular complete and incremental exports. However it expects the importer to match versions. In many disconnected environments user may not know what part the downstream may use. They just want a dump of the data like iso's that could then be hand delivered to the disconnected environment.

Pulp 3.16 and later added a new filesystem exports => https://docs.pulpproject.org/pulpcore/restapi.html#tag/Exporters:-Filesystem-Exports
Katello is taking advantage of this feature implement `syncable` exports.

Initially started the PR making this available for repository exports only. However in the current state all three content-export options are available to be extracted in syncable format.
- `hammer content-export complete version --format=syncable ....`
- `hammer content-export complete library --format=syncable ....`
- `hammer content-export complete repository --format=syncable ....`

The export rpm files are generated as hardlinks to the actual file on disk. This there by avoids the 2X space we may need for regular importable exports.  

There are couple of downsides as stated above
- No import command available here. The disconnected user is responsible for the consistency of data.
- No incremental exports. Only complete for starts.

- Make sure you have https://github.com/Katello/hammer-cli-katello/pull/856 checked out along with this pr.
- On your dev env you may need to run `sudo setfacl --recursive --modify u:vagrant:rwX,d:u:vagrant:rwX /var/lib/pulp/exports` so that listing files can be generate when hammer is running as a vagrant user.
- On your katello instance `Administer -> Settings -> Content`. 
- set both `Default Custom Repository download policy` and `Default Red Hat Repository download policy` to immediate
- Create an org
- Add/Enable + sync few repos (recommend Ansible/Openstack as they tend to be smaller.) 
- Create a content view with these repos an publish.

Try any of the 3 hammer commands listed here
- `hammer content-export complete version --format=syncable ....`
- `hammer content-export complete library --format=syncable ....`
- `hammer content-export complete repository --format=syncable ....`

- Check if the listing files generated ok

### DIFF
--- a/app/controllers/katello/api/v2/content_exports_controller.rb
+++ b/app/controllers/katello/api/v2/content_exports_controller.rb
@@ -34,12 +34,18 @@ module Katello
                                                "no greater than the specified size in gigabytes."), :required => false
     param :fail_on_missing_content, :bool, :desc => N_("Fails if any of the repositories belonging to this version"\
                                                          " are unexportable. False by default."), :required => false
+    param :format, ::Katello::Pulp3::ContentViewVersion::Export::FORMATS,
+                   :desc => N_("Export formats. Choose syncable if content is to be imported via repository sync. "\
+                               "Choose importable if content is to be imported via hammer content-import.
+                                Defaults to importable."),
+                   :required => false
     def version
       tasks = async_task(::Actions::Katello::ContentViewVersion::Export,
                           content_view_version: @version,
                           destination_server: params[:destination_server],
                           chunk_size: params[:chunk_size_gb],
-                          fail_on_missing_content: ::Foreman::Cast.to_bool(params[:fail_on_missing_content]))
+                          fail_on_missing_content: ::Foreman::Cast.to_bool(params[:fail_on_missing_content]),
+                          format: find_export_format)
       respond_for_async :resource => tasks
     end
 
@@ -50,12 +56,18 @@ module Katello
                                                "no greater than the specified size in gigabytes."), :required => false
     param :fail_on_missing_content, :bool, :desc => N_("Fails if any of the repositories belonging to this organization"\
                                                          " are unexportable. False by default."), :required => false
+    param :format, ::Katello::Pulp3::ContentViewVersion::Export::FORMATS,
+                   :desc => N_("Export formats. Choose syncable if content is to be imported via repository sync. "\
+                               "Choose importable if content is to be imported via hammer content-import.
+                                Defaults to importable."),
+                   :required => false
     def library
       tasks = async_task(::Actions::Pulp3::Orchestration::ContentViewVersion::ExportLibrary,
                           @organization,
                           destination_server: params[:destination_server],
                           chunk_size: params[:chunk_size_gb],
-                          fail_on_missing_content: ::Foreman::Cast.to_bool(params[:fail_on_missing_content]))
+                          fail_on_missing_content: ::Foreman::Cast.to_bool(params[:fail_on_missing_content]),
+                          format: find_export_format)
       respond_for_async :resource => tasks
     end
 
@@ -63,10 +75,16 @@ module Katello
     param :id, :number, :desc => N_("Repository identifier"), :required => true
     param :chunk_size_gb, :number, :desc => N_("Split the exported content into archives "\
                                                "no greater than the specified size in gigabytes."), :required => false
+    param :format, ::Katello::Pulp3::ContentViewVersion::Export::FORMATS,
+                   :desc => N_("Export formats. Choose syncable if content is to be imported via repository sync. "\
+                               "Choose importable if content is to be imported via hammer content-import.
+                                Defaults to importable."),
+                   :required => false
     def repository
       tasks = async_task(::Actions::Pulp3::Orchestration::ContentViewVersion::ExportRepository,
                           @repository,
-                          chunk_size: params[:chunk_size_gb])
+                          chunk_size: params[:chunk_size_gb],
+                          format: find_export_format)
       respond_for_async :resource => tasks
     end
 
@@ -88,6 +106,18 @@ module Katello
       find_organization
       unless @organization.can_export_content?
         throw_resource_not_found(name: 'organization', id: params[:organization_id])
+      end
+    end
+
+    def find_export_format
+      if params[:format]
+        unless ::Katello::Pulp3::ContentViewVersion::Export::FORMATS.include?(params[:format])
+          fail HttpErrors::UnprocessableEntity, _('Invalid export format provided. Format must be one of  %s ') %
+                                            ::Katello::Pulp3::ContentViewVersion::Export::FORMATS.join(',')
+        end
+        params[:format]
+      else
+        ::Katello::Pulp3::ContentViewVersion::Export::IMPORTABLE
       end
     end
   end
--- a/app/lib/actions/pulp3/content_view_version/create_syncable_export_history.rb
+++ b/app/lib/actions/pulp3/content_view_version/create_syncable_export_history.rb
@@ -1,57 +1,43 @@
 module Actions
   module Pulp3
     module ContentViewVersion
-      class CreateExportHistory < Actions::EntryAction
+      class CreateSyncableExportHistory < Actions::EntryAction
         input_format do
           param :smart_proxy_id, Integer
-          param :exporter_data, Hash
+          param :base_path, String
           param :content_view_version_id, Integer
-          param :from_content_view_version_id, Integer
           param :destination_server, String
         end
 
         output_format do
           param :export_history_id, Integer
           param :path, String
-          param :exported_file_checksum, String
         end
 
         def run
           smart_proxy = ::SmartProxy.unscoped.find(input[:smart_proxy_id])
-          api = ::Katello::Pulp3::Api::Core.new(smart_proxy)
-          export_data = api.export_api.list(input[:exporter_data][:pulp_href]).results.first
-          output[:exported_file_checksum] = export_data.output_file_info
-          file_name = output[:exported_file_checksum].first&.first
-          path = File.dirname(file_name.to_s)
-          output[:path] = path
+          output[:path] = input[:base_path]
           cvv = ::Katello::ContentViewVersion.find(input[:content_view_version_id])
-          from_cvv = ::Katello::ContentViewVersion.find(input[:from_content_view_version_id]) unless input[:from_content_view_version_id].blank?
-
           export_metadata = ::Katello::Pulp3::ContentViewVersion::Export.create(
                                                      content_view_version: cvv,
                                                      smart_proxy: smart_proxy,
-                                                     format: ::Katello::Pulp3::ContentViewVersion::Export::IMPORTABLE,
-                                                     from_content_view_version: from_cvv).generate_metadata
-
-          toc_path_info = output[:exported_file_checksum].find { |item| item.first.end_with?("toc.json") }
-          export_metadata[:toc] = File.basename(toc_path_info.first)
+                                                     format: input[:format]).generate_metadata
 
           history = ::Katello::ContentViewVersionExportHistory.create!(
             content_view_version_id: input[:content_view_version_id],
             destination_server: input[:destination_server],
-            path: path,
+            path: input[:base_path],
             metadata: export_metadata,
             audit_comment: ::Katello::ContentViewVersionExportHistory.generate_audit_comment(content_view_version: cvv,
                                                                                              user: User.current,
-                                                                                             from_version: from_cvv,
                                                                                              metadata: export_metadata)
           )
           output[:export_history_id] = history.id
-          output[:format] = ::Katello::Pulp3::ContentViewVersion::Export::IMPORTABLE
+          output[:format] = ::Katello::Pulp3::ContentViewVersion::Export::SYNCABLE
         end
 
         def humanized_name
-          _("Create Export History")
+          _("Create Syncable Export History")
         end
       end
     end
--- a/app/lib/actions/pulp3/content_view_version/create_exporter.rb
+++ b/app/lib/actions/pulp3/content_view_version/create_exporter.rb
@@ -6,13 +6,20 @@ module Actions
           param :smart_proxy_id, Integer
           param :content_view_version_id, Integer
           param :destination_server, String
+          param :format, String
+          param :base_path, String
+          param :repository_id, Integer
         end
 
         def run
           cvv = ::Katello::ContentViewVersion.find(input[:content_view_version_id])
-          output[:exporter_data] = ::Katello::Pulp3::ContentViewVersion::Export.new(smart_proxy: smart_proxy,
+          repository = ::Katello::Repository.find(input[:repository_id]) unless input[:repository_id].blank?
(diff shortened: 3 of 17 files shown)


killroy1971:
All, from what I'm reading, I think we are on the right track. 

We need a Red Hat CDN formatted output for any repository in a given Satellite Organization and that Organization's Content View (CV). 
In the interests of sustainability, and continuity, I tell my customers to keep it simple and not use CVs on the Connected Satellite for the Disconnected CDN (Apache) server repository tarball creation tasks. That's better handled by the "--since" date flag as it's possible to keep that in the shell script. Plus, Content Views can get complicated very quickly. What's simple to one system admin is byzantine to another. Then there's the eventual "need" to refactor or redo the CVs, which makes updating the Disconnected CDN a lot more challenging. Move that complexity to the Disconnected Satellite server(s), and use CVs there. I don't want my customers replicating exported content into CV and Organization paths on the Disconnected CDN because the incorrectly think they need to do so. Simpler is better.

Question: What about repository metadata? One issue I had with Pulp 2 repository exports with the "--since" flag was that it created incremental repository metadata, and I can't use these files to update the Disconnected CDN's repository metadata. 

Can we add a flag that would export a fully copy of the repository metadata with incremental repository exports? 

I'm on a NAPS engagement, so I'll have limited access to Red Hat resources if there are any questions.

app/controllers/katello/api/v2/content_exports_controller.rb
```diff
@@ -90,5 +108,17 @@ def find_exportable_organization
         throw_resource_not_found(name: 'organization', id: params[:organization_id])
       end
     end
+
+    def find_export_format
+      if params[:format]
+        unless ::Katello::Pulp3::ContentViewVersion::Export::FORMATS.include?(params[:format])
+          fail HttpErrors::UnprocessableEntity, _('Invalid Export Format provided. Format has to be one of  %s ') %
```
jeremylenz, line 115:
```suggestion
          fail HttpErrors::UnprocessableEntity, _('Invalid export format provided. Format must be one of  %s ') %
```

app/lib/actions/pulp3/content_view_version/create_syncable_export_history.rb
```diff
@@ -0,0 +1,44 @@
...
+          history = ::Katello::ContentViewVersionExportHistory.create!(
+            content_view_version_id: input[:content_view_version_id],
+            destination_server: input[:destination_server],
+            path: input[:base_path],
+            metadata: export_metadata,
+            audit_comment: ::Katello::ContentViewVersionExportHistory.generate_audit_comment(content_view_version: cvv,
+                                                                                             user: User.current,
+                                                                                             metadata: export_metadata)
+          )
+          output[:export_history_id] = history.id
```
jeremylenz, line 35: I think it would be convenient to add `format: syncable` to the task output. I was looking at the dynflow output and trying to confirm that my export was syncable, and that's where I instinctively looked.

app/lib/actions/pulp3/content_view_version/create_syncable_export_history.rb
```diff
@@ -0,0 +1,44 @@
...
+        end
+
+        def humanized_name
+          _("Create Export History")
```
jeremylenz, line 39:
```suggestion
          _("Create Syncable Export History")
```

jeremylenz (commented):
thanks @parthaa! Left a few small comments on the hammer PR as well

parthaa:
@killroy1971 Thank you for taking the time to review

> We need a Red Hat CDN formatted output for any repository in a given Satellite Organization and that Organization's Content View (CV).

So for that case all you would need to do would be `hammer content-export complete library --organization=foo --format=syncable`, 
You'll get the dump of the library as if it came from cdn with the listing files and the works.

> In the interests of sustainability, and continuity, I tell my customers to keep it simple and not use CVs on the Connected Satellite for the Disconnected CDN (Apache) server repository tarball creation tasks. That's better handled by the "--since" date flag as it's possible to keep that in the shell script. Plus, Content Views can get complicated very quickly. What's simple to one system admin is byzantine to another. Then there's the eventual "need" to refactor or redo the CVs, which makes updating the Disconnected CDN a lot more challenging. Move that complexity to the Disconnected Satellite server(s), and use CVs there. I don't want my customers replicating exported content into CV and Organization paths on the Disconnected CDN because the incorrectly think they need to do so. Simpler is better.

We are trying to leverage Pulp3's versioning and changesets for import/export. That being said you can use the `--format=syncable` from this pr (the user will only see a dump directory and the generated content view will be hidden). 
Also note pulp3 does not have '--since', it will more likely have a "get me every thing that changed between publication/version A and publication/version B". Any way what we call the incremental export case will be the natural progression once this complete export work is completed.

> Question: What about repository metadata? One issue I had with Pulp 2 repository exports with the "--since" flag was that it created incremental repository metadata, and I can't use these files to update the Disconnected CDN's repository metadata.
> 
> Can we add a flag that would export a fully copy of the repository metadata with incremental repository exports?

This is a very good point. We are yet to implement incremental and that is in works. One of the suggestions I heard from pulp folks was "incremental exports will generate the rpms that changed, but will copy the complete metadata." That should address your concern. 
 
I tested the generated output dump on a new sat, set the cdn url to this repo and enabled/synced content etc. The main focus of this PR is the complete export. Will ping ya when it is more stable and you can give me pointers on improving it.

parthaa:
[test katello]

app/lib/actions/pulp3/content_view_version/create_syncable_export_history.rb
parthaa: should be there now

jeremylenz (approved):
Thanks @parthaa!

Just thought of one thing; I'll leave it up to you whether to merge or change it.

app/controllers/katello/api/v2/content_exports_controller.rb
```diff
@@ -34,12 +34,18 @@ def index
                                                "no greater than the specified size in gigabytes."), :required => false
     param :fail_on_missing_content, :bool, :desc => N_("Fails if any of the repositories belonging to this version"\
                                                          " are unexportable. False by default."), :required => false
+    param :format, ::Katello::Pulp3::ContentViewVersion::Export::FORMATS,
+                   :desc => N_("Export formats. Choose syncable if content is to be imported via repository sync. "\
```
jeremylenz, line 38:
```suggestion
                   :desc => N_("Export format. Choose syncable if content is to be imported via repository sync. "\
```

app/controllers/katello/api/v2/content_exports_controller.rb
```diff
@@ -50,23 +56,35 @@ def version
...
     end
 
     api :POST, "/content_exports/repository", N_("Performs a full-export of the repository in library.")
     param :id, :number, :desc => N_("Repository identifier"), :required => true
     param :chunk_size_gb, :number, :desc => N_("Split the exported content into archives "\
                                                "no greater than the specified size in gigabytes."), :required => false
+    param :format, ::Katello::Pulp3::ContentViewVersion::Export::FORMATS,
+                   :desc => N_("Export formats. Choose syncable if content is to be imported via repository sync. "\
+                               "Choose importable if content is to be imported via hammer content-import.
+                                Defaults to importable."),
```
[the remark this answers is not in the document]
wbclark: (It's OK if the answer is "no", I'm not sure if you already evaluated some tradeoff here)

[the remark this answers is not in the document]
wbclark: I originally came here from the hammer-cli-katello PR trying to figure out where `--help` text is generated, thinking users might be curious to know what is actually different about the thing beyond what next steps you take with it. But I can't think of any way to do that without making it way too wordy -- seems like foreman-documentation is a better fit.

It also got me thinking that maybe this API description should be more API-centric to be neutral towards hammer/FAM. But then it would have to be overwritten to show something hammer-specific to hammer users. (FAM already has its own documentation, I guess only hammer just re-uses the descriptions from here)

Probably 99% of the users who read this would have gotten here from `hammer content-export --help` anyway so it's likely not worth the extra effort. With that perspective, I think this all looks great as-is :+1:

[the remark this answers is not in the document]
parthaa: > This description text is written out three separate times. Is there any sufficiently good strategy for avoiding the duplication, to reduce the diffs and make it easier to keep consistency in the future?

Thats a good point we may be able to dry it up by creating a String constant somewhere and copy pasting that 3 times. But usually I dont like to over optimize on API doc comments. Reason being if we wanted to change message from ` Choose syncable if content is to be imported` to  `Choose syncable if content is to be imported into a repository` etc I should be able to do that independently.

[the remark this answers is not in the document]
parthaa: > I originally came here from the hammer-cli-katello PR trying to figure out where `--help` text is generated, thinking users might be curious to know what is actually different about the thing beyond what next steps you take with it. But I can't think of any way to do that without making it way too wordy -- seems like foreman-documentation is a better fit.
> 
> It also got me thinking that maybe this API description should be more API-centric to be neutral towards hammer/FAM. But then it would have to be overwritten to show something hammer-specific to hammer users. (FAM already has its own documentation, I guess only hammer just re-uses the descriptions from here)
> 
> Probably 99% of the users who read this would have gotten here from `hammer content-export --help` anyway so it's likely not worth the extra effort. With that perspective, I think this all looks great as-is 👍

We could in theory add the hammer specific docs to hammer and leave the api as it is..

would some thing like ` "Choose importable if content is to be imported via the import-export process"`. work better than ` "Choose importable if content is to be imported via hammer content-import."`

killroy1971:
> @killroy1971 Thank you for taking the time to review
> 
> > We need a Red Hat CDN formatted output for any repository in a given Satellite Organization and that Organization's Content View (CV).
> 
> So for that case all you would need to do would be `hammer content-export complete library --organization=foo --format=syncable`, You'll get the dump of the library as if it came from cdn with the listing files and the works.
> 
> > In the interests of sustainability, and continuity, I tell my customers to keep it simple and not use CVs on the Connected Satellite for the Disconnected CDN (Apache) server repository tarball creation tasks. That's better handled by the "--since" date flag as it's possible to keep that in the shell script. Plus, Content Views can get complicated very quickly. What's simple to one system admin is byzantine to another. Then there's the eventual "need" to refactor or redo the CVs, which makes updating the Disconnected CDN a lot more challenging. Move that complexity to the Disconnected Satellite server(s), and use CVs there. I don't want my customers replicating exported content into CV and Organization paths on the Disconnected CDN because the incorrectly think they need to do so. Simpler is better.
> 
> We are trying to leverage Pulp3's versioning and changesets for import/export. That being said you can use the `--format=syncable` from this pr (the user will only see a dump directory and the generated content view will be hidden). Also note pulp3 does not have '--since', it will more likely have a "get me every thing that changed between publication/version A and publication/version B". Any way what we call the incremental export case will be the natural progression once this complete export work is completed.
> 
> > Question: What about repository metadata? One issue I had with Pulp 2 repository exports with the "--since" flag was that it created incremental repository metadata, and I can't use these files to update the Disconnected CDN's repository metadata.
> > Can we add a flag that would export a fully copy of the repository metadata with incremental repository exports?
> 
> This is a very good point. We are yet to implement incremental and that is in works. One of the suggestions I heard from pulp folks was "incremental exports will generate the rpms that changed, but will copy the complete metadata." That should address your concern.
> 
> I tested the generated output dump on a new sat, set the cdn url to this repo and enabled/synced content etc. The main focus of this PR is the complete export. Will ping ya when it is more stable and you can give me pointers on improving it.

@parthaa Thank you for the reply. I have a Pulp 3 question: When a Pulp 3 Content View version is created and that version is exported, does that Content View version have a full copy of each repository's metadata or is it an incremental copy containing only the changes since the last Content View version? 
When do you think the full metadata export flag could be added to the 'content-export complete library options?

parthaa:
> @parthaa Thank you for the reply. I have a Pulp 3 question: When a Pulp 3 Content View version is created and that version is exported, does that Content View version have a full copy of each repository's metadata or is it an incremental copy containing only the changes since the last Content View version?

Full Copy of each repositories metadata only right now. I am guessing this will be same even when we do incremental. Only artifacts may be different with incremental (which we don't have right now anyway.) Metadata will be full copy.

PR #10184, opened 2022-07-05
Fixes #35026 - As a user, I can edit ACS on the UI

sjha4:
Screenshot from 2022-07-29 09-21-58

### DIFF
--- a/app/views/katello/api/v2/alternate_content_sources/show.json.rabl
+++ b/app/views/katello/api/v2/alternate_content_sources/show.json.rabl
@@ -9,6 +9,10 @@ if @resource.respond_to?(:simplified?)
         { id: product.id, organization_id: product.organization.id, name: product.name, label: product.label }
       end
     end
+  else
+    node :upstream_password_exists do |_acs|
+      @resource.upstream_password.present?
+    end
   end
 end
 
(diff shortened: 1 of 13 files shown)


ianballou:
Also, it would be good to add a blurb telling folks that ACSs affect all repository syncs on any particular smart proxy.

sjha4:
> Some comments about smart proxies and products with respect to organizations:
> 
> We'll need to remove the filtering on smart proxies and products since organization scope doesn't apply here.
> 
> Smart proxy names can still be used, but product names cannot because product names aren't unique across organizations.

I will get the simplified ACS create and view PR in so we can get edits to work for both in this PR.

sjha4:
> Also, it would be good to add a blurb telling folks that ACSs affect all repository syncs on any particular smart proxy.

Ian, can you give this a test now and also where do you think this notice is appropriate? On the create wizard > Smart proxy screen + Review screen and the edit smart proxy modal?

ianballou:
> where do you think this notice is appropriate? On the create wizard > Smart proxy screen + Review screen and the edit smart proxy modal?

That sounds perfect to me!

ianballou (commented):
One thing I noticed was how the edit wizard handled errors. The error popped up in the background and the wizard stayed open.  I tried adding a product to an yum ACS that had only a a file repo in it.

ianballou (commented):
Everything is looking good to me now! Let me know when the last tests make it in.

sjha4:
@ianballou : Added some more tests for simplified acs edits. Should be ready for re-review.

ianballou (approved):

PR #10187, opened 2022-07-05
Fixes #35165 - Jest & Nock JS test improvements

jeremylenz:

Some JS test improvements:

1. In global_test.setup.js we call

```js
afterEach(() => {
  nock.restore();
});
```

`restore()` restores the HTTP interceptor to the normal unmocked behaviour, which is not desired or needed after each test. Instead, changed it to

```js
afterEach(() => {
  nock.cleanAll();
});
```

which clears all Nock scopes, and as a side bonus allows us to remove `nock.cleanAll()` from individual test files.

2. In NockWrapper.js there's a `setInterval` which calls the jest `done()` function. The problem is that `setInterval` does not call its passed function until after `delay` milliseconds - see https://developer.mozilla.org/en-US/docs/Web/API/setInterval

So the minimum time before calling jest `done()` will be 500 milliseconds. This means we are wasting countless seconds on our test runs.

Changed `nockWrapper.js` so that Jest `done()` is also checked and called before the first interval.

With these improvements, __Test run on CI seems to be reduced from >210 to ~160 seconds.__

see if the tests pass

### DIFF
--- a/webpack/components/Search/__tests__/search.test.js
+++ b/webpack/components/Search/__tests__/search.test.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import { renderWithRedux, patientlyWaitFor, fireEvent } from 'react-testing-lib-wrapper';
-import nock, {
+import {
   nockInstance, assertNockRequest, mockAutocomplete, mockSetting,
 } from '../../../test-utils/nockWrapper';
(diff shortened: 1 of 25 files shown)


sjha4:
👍🏼 Going to re-run a few times just to make sure.. :)

sjha4 (approved):
Succeeded 5 times in re-runs. Change makes sense..Ack.. 👍🏼

PR #10188, opened 2022-07-06
Fixes #35169 - Call Pulp RefreshIfNeeded concurrently

pmoravec:

During manifest refresh, call pulp refresh repos concurrently
to leverage pulp tasking system to execute the independent pulp tasks.

Limit the level of concurrency by batches likewise Capsule sync, due to
similar scalability reasons.

Level of concurrency is already limited. A potential impact of multiple dynflow workers shall be reviewed (see #[10129](https://github.com/Katello/katello/pull/10129)), but I feel nothing like that is applicable.

Enable few tens of repositories with certificates from manifest (i.e. Red Hat repos). Then refresh the manifest and check its duration.

After the change, the duration will be shorter.

### DIFF
--- a/app/lib/actions/katello/organization/manifest_refresh.rb
+++ b/app/lib/actions/katello/organization/manifest_refresh.rb
@@ -42,10 +42,14 @@ module Actions
 
         def plan_refresh_repos(import_products_action, org)
           repositories = ::Katello::Repository.in_default_view.in_product(::Katello::Product.redhat.in_org(org))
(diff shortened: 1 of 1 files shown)


jeremylenz (approved):
Testing results:

On master, tasks in sequence: 34 sec
sequence-full-page

On this branch, tasks running concurrently: 28 sec
concurrence-full-page

Thanks @pmoravec!  Code looks good and works well. :+1:

PR #10194, opened 2022-07-12
Create wizard and details view for simplified alternate content sources

sjha4:


ianballou:
This is working just fine for me! Do we need to add any tests?

ianballou (approved):
Seems ready to me! Just need to wait on the core PR now.

PR #10196, opened 2022-07-14
Fixes #35031 - Delete ModuleProfile and ModuleProfileRpms before ModuleStreams

sjha4:
Move deletion of child records before parent in migration.
I have a feeling destroy_all instead of delete_all for child records will resolve this too but it will be a costlier operation.


sjha4:
[test katello]

chris1984:
Code looks good. starting to test

parthaa (approved):
ACK

chris1984 (approved):

chris1984:
Works good, merge away

PR #10198, opened 2022-07-15
Fixes #35233 - ForeignKeyViolation Error with docker_meta_tags during Pulp migration

ianballou:

Updates DockerMetaTag indexing to handle situations where 3 docker tags exist in a single Pulp repository that have the same name. Usually we expect two: one v1 tag and one v2 tag.  However, if a repository had a v2 tag pointed at a manifest and then updated it to point at a manifest list, Pulp will hold on to the manifest and its tag. So, we need to ignore this old tag.

This bug caused data integrity issues with DockerMetaTags. Due to the bug, some DockerMetaTags may have two v2 tags instead of a v1 tag and a v2 tag. It's random.

To clean this up, I've included a migration script. Let me know if the script should exist somewhere besides a migration. It seems to make sense for everyone getting this fix to run the migration, since it's likely there are data errors with older Docker repos.

There may be more efficient ways to write the cleanup script, but I tested it on a reproducer and it's slow to retest, so I've left it as-is. I'm up for suggestions on replacing blocks with queries, but we'll have to retest on the reproducer.

Also, this script simply orphans the container image tags that point to the old unused v2 image manifests.  I don't see a reason to delete them since we'd have to remove them from Pulp as well. At this point, since it's right before switchover, it doesn't seem worth removing anything from the Pulp 2 DB.  The orphaned tags don't seem to cause any data issues.

On a reproducer environment:

1) Run this method to confirm the error: https://github.com/Katello/katello/blob/KATELLO-3.18/app/services/katello/pulp3/migration_switchover.rb#L101

2) Revert the reproducer to ensure the data is back to normal

3) Patch in the change and run the migration

4) Run `combine_duplicate_docker_tags` again and see that it succeeds
  -> If all unmigrated content is deleted, a perhaps better alternative would be to try running the switchover:
```ruby
switchover_service = Katello::Pulp3::MigrationSwitchover.new(SmartProxy.pulp_primary)
switchover_service.run
```

### DIFF
--- a/app/models/katello/docker_meta_tag.rb
+++ b/app/models/katello/docker_meta_tag.rb
@@ -173,12 +173,12 @@ module Katello
                                         select(:docker_tag_id))
       dups = tags.group_by(&:name)
 
+      # Values can have more than 2 values if the upstream container repo changed manifest organization
+      # E.g. if a v2 manifest turned into a v2 manifest list
       dups.map do |name, values|
-        if values.first.docker_manifest.schema_version == 1
-          schema1, schema2 = values
-        else
-          schema2, schema1 = values
-        end
+        schema1 = values.find { |t| t.docker_manifest.schema_version == 1 }
+        schema2 = values.find { |t| t.docker_manifest.schema_version == 2 && t.docker_taggable_type == 'Katello::DockerManifestList' }
+        schema2 ||= values.find { |t| t.docker_manifest.schema_version == 2 }
         [schema1.try(:id), schema2.try(:id), name]
       end
     end
--- a/app/services/katello/pulp3/migration_switchover.rb
+++ b/app/services/katello/pulp3/migration_switchover.rb
@@ -16,6 +16,7 @@ module Katello
       def run
         Katello::Logging.time("CONTENT_SWITCHOVER - Total Switchover Process") do
           Katello::Logging.time("CONTENT_SWITCHOVER - check_already_migrated_content") { check_already_migrated_content }
+          Katello::Logging.time("CONTENT_SWITCHOVER - correct_docker_meta_tags") { correct_docker_meta_tags } if docker_migration?
           Katello::Logging.time("CONTENT_SWITCHOVER - cleanup_v1_docker_tags") { cleanup_v1_docker_tags } if docker_migration?
           Katello::Logging.time("CONTENT_SWITCHOVER - migrated_content_type_check") { migrated_content_type_check }
           Katello::Logging.time("CONTENT_SWITCHOVER - combine_duplicate_content_types") { combine_duplicate_content_types }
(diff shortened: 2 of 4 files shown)


parthaa:
Code wise lgtm

ianballou:
TODO: move to a rake task and include the cleanup in the switchover.

wbclark (commented):
Summary of some topics discussed in the group code review session yesterday:

1. It's necessary to resolve the issue to allow the pulp data migration to complete, so it should run during the pulp data migration automatically even when users did not know they were affected

2. Users who are aware of the issue may want to run the fix separately and in advance of the pulp data migration, if they are not yet ready to upgrade

3. Pulp is is required to be up, which can't be cleanly guaranteed in the db:migrate task

4. A standalone rake task can check that back end services are available before proceeding, but can't be easily run during migration_switchover.rb

5. A method in migration_switchover.rb can resolve the issue automatically during the content migration, even for users unaware they are affected, and can still be wrapped in an external rake task. This design should satisfy all of the above requirements.

ianballou:
I'm testing updating the logic so that repositories don't get reindexed on a second run of this script.

lib/katello/tasks/clean_docker_meta_tags.rake
```diff
@@ -0,0 +1,8 @@
+require File.expand_path("../engine", File.dirname(__FILE__))
+
+namespace :katello do
+  desc "Reindex DockerMetaTags in response to Redmine #35233"
```
wbclark, line 4: Some downstream users may find this confusing. Doesn't seem like too much but at the end of the day it can actually generate support cases with inquiries about what does that mean.

In the future I think we need a better way to customize these types of strings for upstream and downstream users, like having defaults that are tailored for the experience of upstream users but can be overridden by drop-in files that can be included in downstream builds. That way an upstream user can see a link to a Redmine issue, or just explanation in upstream terminology "katello, content-proxy" and downstream builds could overwrite these with downstream terminology, knowledgebase links, etc.

Obviously that's not something we can solve right now. I could do a one-off overwrite when building an RPM for this but then someone else would have to do repeat it later if it goes into a z-stream release.

So it might be best to be a bit more general here for now.

```suggestion
  desc "Reindex DockerMetaTags"
```

wbclark: Actually, on second thought. Since this rake task would only run in the one-off scenario, this string should never be displayed during the migration switchover and would only be found by an user that went looking for this. So I think it's probably fine as it is?

parthaa (approved):

app/models/katello/docker_meta_tag.rb
```diff
@@ -173,12 +173,12 @@ def self.get_tag_table_values(repo)
...
+      # Values can have more than 2 values if the upstream container repo changed manifest organization
+      # E.g. if a v2 manifest turned into a v2 manifest list
       dups.map do |name, values|
-        if values.first.docker_manifest.schema_version == 1
-          schema1, schema2 = values
-        else
-          schema2, schema1 = values
-        end
+        schema1 = values.find { |t| t.docker_manifest.schema_version == 1 }
+        schema2 = values.find { |t| t.docker_manifest.schema_version == 2 && t.docker_taggable_type == 'Katello::DockerManifestList'}
```
wbclark, line 180:
```suggestion
        schema2 = values.find { |t| t.docker_manifest.schema_version == 2 && t.docker_taggable_type == 'Katello::DockerManifestList' }
```

to make rubocop happy

lib/katello/tasks/clean_docker_meta_tags.rake
ianballou: yeah they would need to run `foreman-rake --tasks` to see this.

ianballou:
```
# foreman-rake --tasks
...
rake katello:clean_docker_meta_tags                                                  # Reindex DockerMetaTags in response to Redmine #35233
rake katello:clean_old_file_repos                                                    # Cleans up file repos that were moved to /pulp/isos/<org name>/
rake katello:clean_published_repo_directories                                        # Cleans up obsolete directories and estimates time to republish current directories
...
```

ianballou:
`DockerMetaTags` can be orphaned (no repository) in some cases. With this fix, those orphaned `DockerMetaTags` are not deleted. This may be causing cases where the Pulp migration will still fail with the original error. It needs to be investigated.

PR #10199, opened 2022-07-18
Fixes #34792 - Do not show duplicate module streams in the new UI

chris1984:

* Added a new method to filter out duplicate streams and only return one of each for the new ui.

* Made sure Module stream applicabilty was not affected

* Apply patch
* Sync CentOS8 stream repo from anywhere or this mirror (https://mirror.steadfast.net/centos/8-stream/BaseOS/x86_64/os/) - I synced RHEL 8 BaseOS/AppStream and registered a RHEL 8 host but it should work with CentOS
* Register a CentOS 8 stream machine to your Katello server
* Verify you do not see duplicate module streams in the new host UI

screenshot-192 168 121 231-2022 07 18-12_11_38


app/controllers/katello/api/v2/host_module_streams_controller.rb
```diff
@@ -22,7 +22,7 @@ def index
     def index_relation
       return HostAvailableModuleStream.upgradable([@host]) if params[:status] == HostAvailableModuleStream::UPGRADABLE
 
-      rel = @host.host_available_module_streams
+      rel = @host.host_available_module_streams.where(available_module_stream_id: find_available_module_stream_ids)
```
jeremylenz, line 25: I'd like to see a test (or expectation added to `test_index_relation`) that simulates multiples being returned by `host_available_module_streams` that fails without this change.

app/controllers/katello/api/v2/host_module_streams_controller.rb
```diff
@@ -38,6 +38,14 @@ def resource_class
 
     private
 
+    def find_available_module_stream_ids
+      items = ::Katello::AvailableModuleStream.select(:name, :stream, :id)
+      grouped = items.group_by { |item| [item.name, item.stream] }
```
jeremylenz, line 43: If we see performance problems in the future with this query, it's probably because of the `group_by` here which uses Ruby and not SQL to organize the data. There's probably a way to do it with Active Record instead (`.group`), but I'd say let's not prematurely optimize. So I'm fine with this for now :)

jeremylenz (commented):
Thanks @chris1984!

Tested and works great.

one comment and one non-comment below

app/controllers/katello/api/v2/host_module_streams_controller.rb
parthaa: @jeremylenz Good point but I don't think we have to worry about that. We are grouping module streams available for the host.  For my test centos 8 host there are 459 module streams.
I tried the following test
```ruby
def time_it
  start = (Time.now.to_f * 1000).to_i
  yield
  (Time.now.to_f * 1000).to_i - start
end  

>  x = [5] * 10000;x.count
=> 10000

time_it { x.group_by {|x|x} }
=> 3
```
 So it takes 3 milliseconds to group 10K items. Dont think we have to worry about this.

That being said lets ensure that we are only grouping available module streams for one host and not all.

@chris1984 would recommend
```ruby
    def find_available_module_stream_ids
      items = @host.host_available_module_streams.includes(:available_module_stream).map(&:available_module_stream) 
      grouped = items.group_by { |item| [item.name, item.stream] }
      grouped.values.map do |item|
        item.first.id
      end
    end
```

app/controllers/katello/api/v2/host_module_streams_controller.rb
chris1984: @jeremylenz added

test/fixtures/models/katello_available_module_streams.yml
```diff
@@ -17,3 +17,8 @@ available_module_stream_three:
   name: three
   stream: noitsariver
   context: buckeye
+
+available_module_stream_three_diff_context:
```
parthaa, line 21: This change is not necessary. Can you remove it ?

chris1984:
@jeremylenz updated with parthas suggestion and fixed the fixture issue partha commented on

jeremylenz (approved):
re-tested, still works

APJ :+1:

PR #10201, opened 2022-07-18
Fixes #35238 - Remove sca toggle from manifest modal

chris1984:

* With the start of the new candlepin upgrade feature work, toggling sca is no longer part of the manifest process, so this pr removes the toggle in the modal. You can still toggle it in the org settings

* The test file I removed only tested the toggle of sca, so just removed it since it was not used by anything else.

* Check out PR and verify there are no console errors on the subscription page or the manage manifest modal
* Import a manifest/refresh/delete and check to make sure that all works correctly


lfu (approved):
LGTM 👍🏻

PR #10202, opened 2022-07-19
Fixes #35246 - Default cv creation error during seed

sjha4:
1. Adds a validation to ensure default CVs have Default Content View as name and Default_content_view as label.
2. Ensure db:seed doesn't fail for default CVs with wrong label and corrects the label.

1. Create a test org
4. In your console: 
cv = Katello::ContentView.where(default: true, organization_id: 4)&.first
cv.update_attribute(:label, "abc")
5. org.update_attribute(:created_in_katello, false) 
6. Run bundle exec rails db:seed
7. Without this change, it should fail.

### DIFF
--- a/app/models/katello/content_view.rb
+++ b/app/models/katello/content_view.rb
@@ -33,49 +33,50 @@ module Katello
     has_many :content_view_repositories, :class_name => 'Katello::ContentViewRepository',
              :dependent => :destroy, :inverse_of => :content_view
     has_many :repositories, :through => :content_view_repositories, :class_name => "Katello::Repository",
-                            :after_remove => :remove_repository
+             :after_remove => :remove_repository
 
     has_many :filters, :dependent => :destroy, :class_name => "Katello::ContentViewFilter"
 
     has_many :activation_keys, :class_name => "Katello::ActivationKey", :dependent => :restrict_with_exception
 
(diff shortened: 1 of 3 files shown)


sjha4:
[test katello]

sjha4:
[test katello]

sjha4:
[test katello]

sjha4:
[test katello]

app/lib/actions/katello/repository/upload_files.rb
```diff
@@ -29,7 +30,7 @@ def plan(repository, files, content_type = nil, options = {})
             tmp_files.each do |file|
               sequence do
                 upload_action = plan_action(Pulp3::Orchestration::Repository::UploadContent,
-                                                 repository, SmartProxy.pulp_primary!, file, unit_type_id)
+                                            repository, SmartProxy.pulp_primary!, file, unit_type_id)
```
parthaa, line 33: how is this related to this PR ?

parthaa:
Looks like the only relevant that changed in `models/katello/content_view.rb`  is a new method `check_default_label_name` and a validates on that. Could we not do the formatting in this file, will make it easier to review and cherry pick?

parthaa (approved):

PR #10206, opened 2022-07-21
Fixes #35262: Update redhat-uep.pem from RHEL 8

ehelms:
A newer version of redhat-uep.pem is needed for enhanced crypto
policies and FIPS enabled modes on EL. The prior version was from
RHEL 7 and only contained sha1 signature algorithms within the chain.

Simple verification is to spin up a RHEL 8 machine and check that /etc/rhsm/ca/redhat-uep.pem matches what I have in this PR.


chris1984 (approved):
ACK, looks the same

PR #10209, opened 2022-07-25
Fixes #35276 - Add LCE controller to AngularJS so bookmarks work

chris1984:

In AngularJS we are missing the `$scope.controllerName` which allows bookmarks to be saved/created.

N/A

* Apply PR
* Try to create a bookmark under any of the tabs except details.
* Verify it saved and we get a success response from the Foreman bookmarks controller.


lfu (approved):
LGTM 👍🏻

PR #10208, opened 2022-07-22
Fixes #35265 - When toggling SCA, change the owner and don't refresh manifest

lfu:
#### What are the changes introduced in this pull request?
1. Don't refresh manifest when toggling SCA on the Organization edit page
2. To toggle SCA on an org, hit the /candlepin/owners endpoint and not the consumers/:uuid.

#### What are the testing steps for this pull request?
Verify simple content access toggling on organization works.

To upgrate environment to 4.2 candlepin for simple content access change                                                                                              
```   
systemctl stop tomcat
yum install -y http://download.eng.bos.redhat.com/brewroot/vol/rhel-7/packages/candlepin/4.2.3/1.el7sat/noarch/candlepin-4.2.3-1.el7sat.noarch.rpm
yum install -y http://download.eng.bos.redhat.com/brewroot/vol/rhel-7/packages/candlepin/4.2.3/1.el7sat/noarch/candlepin-selinux-4.2.3-1.el7sat.noarch.rpm
/usr/share/candlepin/cpdb --update
systemctl start tomcat
```
EL8 packages: http://download.eng.bos.redhat.com/brewroot/vol/rhel-8/packages/candlepin/4.2.3/1.el8sat/noarch/

PR #10211, opened 2022-07-26
Fixes #35284 - Unhide SCA checkbox in org settings for edit

chris1984:

* Removed the logic that hides the sca checkbox in org settings, so it can be used as part of the new SCA work.

* Making sure the org edit page, did not break

* Apply PR
* Don't import a manifest
* Try to edit an org without a manifest and see if you see the checkbox
* Create an org and when you get to the edit org page in the wizard, make sure checkbox is present

### DIFF
--- a/app/controllers/katello/concerns/organizations_controller_extensions.rb
+++ b/app/controllers/katello/concerns/organizations_controller_extensions.rb
@@ -6,12 +6,6 @@ module Katello
       include Foreman::Controller::Flash
 
       module Overrides
-        def edit
-          @can_toggle_sca = Katello::UpstreamConnectionChecker.new(@taxonomy).can_connect? &&
-            (@taxonomy.upstream_consumer.simple_content_access_eligible? || @taxonomy.simple_content_access?)
-          super
-        end
-
         def destroy
           if @taxonomy.is_a?(Organization)
             begin
@@ -50,20 +44,16 @@ module Katello
       end
 
       def update
-        super # we run super first here so that flash messages won't be in a confusing order
         return if params[:simple_content_access].nil?
         sca_param = ::Foreman::Cast.to_bool(params[:simple_content_access])
-        if sca_param && !@taxonomy.simple_content_access?
+        if sca_param && !@taxonomy.simple_content_access?(cached: false)
           # user has requested SCA enable
-          task = async_task(::Actions::Katello::Organization::SimpleContentAccess::Enable, params[:id])
-          info "Enabling Simple Content Access for organization #{@taxonomy.name}.",
-            link: { text: "View progress on the Tasks page", href: "/foreman_tasks/tasks/#{task&.id}" }
-        elsif !sca_param && @taxonomy.simple_content_access?
+          sync_task(::Actions::Katello::Organization::SimpleContentAccess::Enable, params[:id])
+        elsif !sca_param && @taxonomy.simple_content_access?(cached: false)
           # user has requested SCA disable
-          task = async_task(::Actions::Katello::Organization::SimpleContentAccess::Disable, params[:id])
-          info "Disabling Simple Content Access for organization #{@taxonomy.name}.",
-            link: { text: "View progress on the Tasks page", href: "/foreman_tasks/tasks/#{task&.id}" }
+          sync_task(::Actions::Katello::Organization::SimpleContentAccess::Disable, params[:id])
         end
+        super
       end
 
       included do
--- a/app/views/overrides/organizations/_edit_override.html.erb
+++ b/app/views/overrides/organizations/_edit_override.html.erb
@@ -12,13 +12,10 @@
                'data-url' => "/katello/api/v2/organizations/#{@taxonomy.id}/download_debug_certificate"
   end %>
 
-  <% if @can_toggle_sca %>
-    <%= field(f, _('Simple Content Access'),
-              :help_inline => _('Toggling Simple Content Access will refresh your manifest.')) do
-      tag1 = hidden_field_tag 'simple_content_access', '0'
-      tag2 = check_box_tag 'simple_content_access', '1', @taxonomy.simple_content_access?
-      tag1 + tag2
-    end %>
-  <% end %>
+  <%= field(f, _('Simple Content Access')) do
+    tag1 = hidden_field_tag 'simple_content_access', '0'
+    tag2 = check_box_tag 'simple_content_access', '1', @taxonomy.simple_content_access?(cached: false)
+    tag1 + tag2
+  end %>
 
 <% end %>
--- a/app/views/overrides/organizations/_index_row_override.html.erb
+++ b/app/views/overrides/organizations/_index_row_override.html.erb
@@ -1,3 +1,3 @@
 <% if taxonomy.is_a?(Organization) %>
-  <td><%= checked_icon taxonomy.simple_content_access? %></td>
+  <td><%= checked_icon taxonomy.simple_content_access?(cached: false) %></td>
 <% end %>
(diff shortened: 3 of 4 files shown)


lfu:
Do we need to clean up [this](https://github.com/Katello/katello/blob/597b385ba32e13d0e9764bd03c7e45644bab9482/app/controllers/katello/concerns/organizations_controller_extensions.rb#L10)?

chris1984:
Removed the duplicate notification too

jeremylenz:
Had to upgrade Candlepin using instructions on https://github.com/Katello/katello/pull/10208 to get this working, just fyi

app/views/overrides/organizations/_edit_override.html.erb
```diff
@@ -12,13 +12,10 @@
...
-    <%= field(f, _('Simple Content Access'),
-              :help_inline => _('Toggling Simple Content Access will refresh your manifest.')) do
-      tag1 = hidden_field_tag 'simple_content_access', '0'
-      tag2 = check_box_tag 'simple_content_access', '1', @taxonomy.simple_content_access?
-      tag1 + tag2
-    end %>
-  <% end %>
+  <%= field(f, _('Simple Content Access')) do
+    tag1 = hidden_field_tag 'simple_content_access', '0'
+    tag2 = check_box_tag 'simple_content_access', '1', @taxonomy.simple_content_access?
```
jeremylenz, line 17:
```suggestion
    tag2 = check_box_tag 'simple_content_access', '1', @taxonomy.simple_content_access?(cached: false)
```

app/controllers/katello/concerns/organizations_controller_extensions.rb
```diff
@@ -55,14 +49,10 @@ def update
         sca_param = ::Foreman::Cast.to_bool(params[:simple_content_access])
         if sca_param && !@taxonomy.simple_content_access?
```
jeremylenz, line 50:
```suggestion
        if sca_param && !@taxonomy.simple_content_access?(cached: false)
```

app/controllers/katello/concerns/organizations_controller_extensions.rb
```diff
@@ -55,14 +49,10 @@ def update
...
           # user has requested SCA enable
-          task = async_task(::Actions::Katello::Organization::SimpleContentAccess::Enable, params[:id])
-          info "Enabling Simple Content Access for organization #{@taxonomy.name}.",
-            link: { text: "View progress on the Tasks page", href: "/foreman_tasks/tasks/#{task&.id}" }
+          async_task(::Actions::Katello::Organization::SimpleContentAccess::Enable, params[:id])
         elsif !sca_param && @taxonomy.simple_content_access?
```
jeremylenz, line 53:
```suggestion
        elsif !sca_param && @taxonomy.simple_content_access?(cached: false)
```

app/controllers/katello/concerns/organizations_controller_extensions.rb
```diff
@@ -55,14 +49,10 @@ def update
         sca_param = ::Foreman::Cast.to_bool(params[:simple_content_access])
         if sca_param && !@taxonomy.simple_content_access?
           # user has requested SCA enable
-          task = async_task(::Actions::Katello::Organization::SimpleContentAccess::Enable, params[:id])
-          info "Enabling Simple Content Access for organization #{@taxonomy.name}.",
-            link: { text: "View progress on the Tasks page", href: "/foreman_tasks/tasks/#{task&.id}" }
+          async_task(::Actions::Katello::Organization::SimpleContentAccess::Enable, params[:id])
```
jeremylenz, line 52:
```suggestion
          sync_task(::Actions::Katello::Organization::SimpleContentAccess::Enable, params[:id])
```
I think this is quick enough that these can be sync tasks anyway

app/controllers/katello/concerns/organizations_controller_extensions.rb
```diff
@@ -55,14 +49,10 @@ def update
...
         elsif !sca_param && @taxonomy.simple_content_access?
           # user has requested SCA disable
-          task = async_task(::Actions::Katello::Organization::SimpleContentAccess::Disable, params[:id])
-          info "Disabling Simple Content Access for organization #{@taxonomy.name}.",
-            link: { text: "View progress on the Tasks page", href: "/foreman_tasks/tasks/#{task&.id}" }
+          async_task(::Actions::Katello::Organization::SimpleContentAccess::Disable, params[:id])
```
jeremylenz, line 55:
```suggestion
          sync_task(::Actions::Katello::Organization::SimpleContentAccess::Disable, params[:id])
```

jeremylenz (commented):
Thanks @chris1984!

I noticed an existing bug which I think we should fix here: Turning off SCA in my Default Organization, _sometimes_ the SCA checkbox in the organizations _list_ table didn't update quickly enough. So I'd be looking at the notification "successfully updated Default Organization" but the box would still be checked. A page refresh would correct the issue.

I first thought that maybe it's because it wasn't waiting for the task, so I changed `async_task` to `sync_task`. But it turns out the real issue was that `simple_content_access?` is a cached value. It only updates every 1 minute unless you pass `cached: false` to it.

I added suggestions for 3 places where it's probably good to add it, but it also needs to be added in `app/views/overrides/organizations/_index_row_override.html.erb` - this is the place that was actually causing the issue.

by the way, if you're unable to turn off SCA for an org, you can do this in Rails console:
```
Katello::Resources::Candlepin::Owner.update('Default_Organization', { 'contentAccessModeList' => 'entitlement,org_environment' })
```
(replacing Default_Organization with your org's label.)

chris1984:
@jeremylenz updated sir

chris1984:
This passes locally on my box, i'm guessing it's because we are fixing things with Candlepin 4.2 and jenkins is running 4.1

```bash
# Running tests with run options --seed 5896:

........../home/vagrant/katello/test/support/controller_support.rb:63: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/home/vagrant/katello/test/support/controller_support.rb:6: warning: The called method `check_permission' is defined here
/home/vagrant/katello/test/support/controller_support.rb:68: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/home/vagrant/katello/test/support/controller_support.rb:6: warning: The called method `check_permission' is defined here
............Writing XML reports to /home/vagrant/foreman/jenkins/reports/unit


Finished tests in 7.536992s, 2.9189 tests/s, 10.7470 assertions/s.


22 tests, 81 assertions, 0 failures, 0 errors, 0 skips
```

jeremylenz:
I thought 4.2 was tagged back into nightly?

chris1984:
@jeremylenz 

http://yum.theforeman.org/katello/nightly/candlepin/el8/x86_64/ looks like it got reverted

chris1984:
[test katello]

chris1984:
@jeremylenz all green :smile:

app/controllers/katello/concerns/organizations_controller_extensions.rb
```diff
@@ -53,16 +47,12 @@ def update
         super # we run super first here so that flash messages won't be in a confusing order
```
jeremylenz, line 47: Now that we don't have multiple flash messages any more, this `super` can be moved to the end of the method and the comment removed.

jeremylenz (commented):
One nit below, but I tested again and works great :+1:

chris1984:
@jeremylenz updated and got rid of the comment and moved super to the bottom.

jeremylenz (approved):
Thanks @chris1984!

APJ :+1:

PR #10214, opened 2022-07-28
Fixes #35290 - Katello rpm search via nvra also

pmoravec:
Add scoped_search on nvra. This helps user experience when filename does
not follow NVRA pattern.

Signed-off-by: Pavel Moravec <pmoravec@redhat.com>

Added scoped search on nvra over rpm objects.

??

Import a package with filename not following NVRA pattern (e.g. google-chrome-stable_current_x86_64.rpm) to a custo repo. Then search the repo by NVRA.


chris1984 (approved):
ACK works, thanks @pmoravec

PR #10215, opened 2022-07-28
Refs #35006 - Clean up unused Pulp 2 code

ekohl:
While this doesn't completely solve it (https://github.com/Katello/katello/pull/10176 aims for that), it already gets rid of a lot of unused code.


lib/katello/tasks/repository.rake
```diff
@@ -108,7 +108,7 @@ namespace :katello do
       return false unless backend_service&.repository_reference&.repository_href
       backend_service.api.repositories_api.read(backend_service.repository_reference.repository_href)
     else
-      Katello.pulp_server.extensions.repository.retrieve(repo.pulp_id)
```
ianballou, line 111: If I'm remembering right, this line would trigger a 404 error and get caught below in `rescue StandardError`.  The false there would get replaced by the true on line 113 as the return object in the odd case that Pulp 3 isn't supported.

I'd replace the entire method with something like:

```ruby
def repo_exists?(repo)
  backend_service = repo.backend_service(SmartProxy.pulp_primary!)
  return false unless backend_service&.repository_reference&.repository_href
  backend_service.api.repositories_api.read(backend_service.repository_reference.repository_href)
  true
rescue StandardError => e
  return false if e.code == 404
end
```

test/support/scenario_support.rb
```diff
@@ -72,9 +72,7 @@ def record(name, options = {})
   end
 
   def destroy_repo_if_exists(repo)
-    if exists? { Katello.pulp_server.resources.repository.retrieve(repo.pulp_id) }
-      Katello.pulp_server.resources.repository.delete(repo.pulp_id)
-    end
+    # TODO: Implement for Pulp 3
```
ianballou, line 75: Don't worry about re-implementing this, it can be taken out.  We have all the support methods we need in `test/support/pulp3_support.rb`. The only method that uses it is `create_repo` in the same file. I believe all other uses of `create_repo` in the tests actually refer to the `create_repo` in the Pulp 3 test support class. Too bad it has the same name!

ianballou:
From the test failure, it looks like you'll need to rerecord the VCR for `test_create_remote_with_http_proxy_creds` in `yum_test.rb`. I'm not sure what changed, we'd have to rerecord the new VCR and then look at the POST body diff.

test/support/scenario_support.rb
ekohl: Not sure, It is referenced in https://github.com/Katello/katello/blob/a9e44e26abb667ed474fc562b97172a62ea98a2a/test/scenarios/scenario_test.rb#L18-L39 but that does have a `skip`. If it was completely unreferenced I'd have removed it.

What's your preference here?

ekohl:
> From the test failure, it looks like you'll need to rerecord the VCR for `test_create_remote_with_http_proxy_creds` in `yum_test.rb`. I'm not sure what changed, we'd have to rerecord the new VCR and then look at the POST body diff.

In 29cb32ef32182860b2524826bcbf00dae5ef5841 I did remove the inclusion of `RepositorySupport` from `YumVcrTest`. Perhaps test class relied on it to include things? I can see this bit may have been required.
```ruby
module RepositorySupport
  extend ActiveSupport::Concern

  included do
    include TaskSupport
    include VCR::TestCase
  end
`
```

ianballou:
@ekohl ah I missed that, yeah we'll need to make sure those items keep getting included to the tests that used to use RepositorySupport.

test/support/scenario_support.rb
ianballou: We can remove the test, it looks to me like everything tested in there is tested elsewhere.

ianballou:
Just checking up on this, and to follow up on my last comment, I think you can just add `include VCR::TestCase` to anything that used `RepositorySupport`.  `include Katello::Pulp3Support` works too since that imports `VCR::TestCase`.

ianballou:
Ping ping, @ekohl do you want us to take this one over?

ekohl:
@ianballou did you make any progress on taking this over?

ekohl:
Great. I didn't see it linked back, so wasn't sure if you did. Glad to see. Perhaps good to get this in early so we get a full upstream release to test it out?

ianballou:
Yes indeed, I'm going to ask to get it reviewed soon.

ekohl:
Closing in favor of https://github.com/Katello/katello/pull/10400.

PR #10217, opened 2022-07-29
Fixes #35296 - New Custom CDN type

parthaa:
Now that we have the syncable yum exports available, we need the ability to pull in the exported content.
One should be able to enable via http / https to a custom cdn location.

However while enabling a repo via https we hit an SSL verifcation issue. This is because we only pass redhat ca cert while talking to cdn. So the connection is not able to verify the custom cdn url.
This commit adds a new 'custom_cdn'  type configuration where you specify both your custom cdn url and ssl cert.
 
- New Custom CDN configuration type. API + UI Changes
- Updated the RedHat CDN type and set its url as read only.

Initially thought of just using the default trust store to accept custom cert. However from a security perspective we probably shouldn't be making use of those.

Feel free to ping me for the dump url or do the following 7 steps to generate one.
- Enable a RH Repo like `rhel-7-server-ansible-2.9-rpms`
- Go to products -> repo details of this repo and mark the download policy  immediate
- Sync repo
- On your dev env you may need to run sudo setfacl --recursive --modify u:vagrant:rwX,d:u:vagrant:rwX /var/lib/pulp/exports so that listing files can be generate when hammer is running as a vagrant user.
- `hammer content-export complete repository --id=<repo-id> --format=syncable` -> This should generate a dump
- `mv <dump-dir>  /var/www/html/pub/repos`
- make sure `https://<...>/pub/repos` is browsable.

Now in a new Katello instance which will be the importing sat
- create new org
- import manifest
- Goto Subscriptions => Manage Manifest => CDN Configuration => Redhat CDN
-  Change the URL to https://<...>/pub/repos
- Click updateb
- Now go to Redhat Repos page and try enabling the ansible repo

Before this PR - 
You should see an error like 
```
2022-07-28T18:01:33 [I|app|37e3cb19] CDN: Requesting path https://<webserver>:443/pub/repos/content/dist/rhel/server/7/listing
/opt/rh/rh-ruby27/root/usr/share/ruby/net/protocol.rb:44: warning: exception in verify_callback is ignored
2022-07-28T18:01:33 [E|app|37e3cb19] Failed at scanning for repository: SSL_connect returned=1 errno=0 state=error: certificate verify failed (self signed certificate in certificate chain)
```

After PR
- `bundle exec rake db:migrate`
- Go to `Content->Content Credentials`
- Create a new cert with contents of -> `https://<...>/pub/katello-server-ca.crt`
- Go to Subscriptions => Manage Manifest => CDN Configuration => Custom CDN
- Update the url to `https://<...>/pub/repos`  and content credentials to the one you created above
- Now try enabling and syncing the desired repo.

### DIFF
--- a/webpack/scenes/Subscriptions/Manifest/CdnConfigurationTab/CustomCdnTypeForm.js
+++ b/webpack/scenes/Subscriptions/Manifest/CdnConfigurationTab/CustomCdnTypeForm.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useRef } from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import { useDispatch, useSelector } from 'react-redux';
 import { FormattedMessage } from 'react-intl';
@@ -18,40 +18,46 @@ import { noop } from 'foremanReact/common/helpers';
 import { CUSTOM_CDN } from './CdnConfigurationConstants';
 import { updateCdnConfiguration } from '../../../Organizations/OrganizationActions';
 import {
+  selectOrgLoading,
   selectUpdatingCdnConfiguration,
 } from '../../../Organizations/OrganizationSelectors';
 import './CdnConfigurationForm.scss';
 
 const CustomCdnTypeForm = ({
-  showUpdate, onUpdate, contentCredentials, cdnConfiguration,
+  typeChangeInProgress, onUpdate, contentCredentials, cdnConfiguration,
 }) => {
   const dispatch = useDispatch();
   const urlValue = cdnConfiguration.type === CUSTOM_CDN ? cdnConfiguration.url : '';
   const [url, setUrl] = useState(urlValue);
-  const [updateEnabled, setUpdateEnabled] = useState(showUpdate);
   const updatingCdnConfiguration = useSelector(state => selectUpdatingCdnConfiguration(state));
-  const firstUpdate = useRef(true);
+  const orgIsLoading = useSelector(state => selectOrgLoading(state));
   const [sslCaCredentialId, setSslCaCredentialId] = useState(cdnConfiguration.ssl_ca_credential_id);
 
-  useEffect(() => {
-    if (firstUpdate.current) {
-      firstUpdate.current = false;
-      return;
+  const disableUpdate = () => {
+    if (updatingCdnConfiguration || orgIsLoading) {
+      return true;
     }
-    setUpdateEnabled(true);
-  }, [sslCaCredentialId, cdnConfiguration]);
 
-  const onError = () => setUpdateEnabled(true);
+    const sslCaCredentialIdOrNull = sslCaCredentialId === '' ? null : sslCaCredentialId;
+    if (!url || (url === cdnConfiguration.url
+              && sslCaCredentialIdOrNull?.toString() ===
+                      cdnConfiguration.ssl_ca_credential_id?.toString())) {
+      return true;
+    }
+
+    return false;
+  };
 
   const performUpdate = () => {
-    setUpdateEnabled(false);
     dispatch(updateCdnConfiguration({
       url,
       ssl_ca_credential_id: sslCaCredentialId,
       type: CUSTOM_CDN,
-    }, onUpdate, onError));
+    }, onUpdate));
   };
 
+  const sslCaCredentialValue =
+    cdnConfiguration.type === CUSTOM_CDN || typeChangeInProgress ? sslCaCredentialId : null;
 
   return (
     <Form isHorizontal>
@@ -59,13 +65,13 @@ const CustomCdnTypeForm = ({
         <Text>
           <FormattedMessage
             id="cdn-configuration-type"
-            defaultMessage={__('Red Hat content will be consumed from the {type}.')}
+            defaultMessage={__('Red Hat content will be consumed from {type}.')}
             values={{
-              type: <strong>{__('Custom CDN Url')}</strong>,
+              type: <strong>{__('a custom CDN URL')}</strong>,
             }}
           />
           <br />
-          {showUpdate &&
+          {typeChangeInProgress &&
           <FormattedMessage
             id="cdn-configuration-type-cdn"
             defaultMessage={__('Click {update} below to save changes.')}
@@ -97,7 +103,7 @@ const CustomCdnTypeForm = ({
         <FormSelect
           ouiaId="custom-cdn-ca-content-credential-input"
           aria-label="cdn-ssl-ca-content-credential"
-          value={sslCaCredentialId || ''}
+          value={sslCaCredentialValue || ''}
           isDisabled={updatingCdnConfiguration}
           onChange={value => setSslCaCredentialId(value)}
         >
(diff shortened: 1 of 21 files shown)


parthaa:
[test katello]

parthaa:
[test katello]

webpack/scenes/Subscriptions/Manifest/CdnConfigurationTab/CustomCdnTypeForm.js
```diff
@@ -0,0 +1,148 @@
...
+  };
+
+
+  return (
+    <Form isHorizontal>
+      <div id="update-hint-cdn" className="margin-top-16">
+        <Text>
+          <FormattedMessage
+            id="cdn-configuration-type"
+            defaultMessage={__('Red Hat content will be consumed from the {type}.')}
```
jeremylenz, line 62:
```suggestion
            defaultMessage={__('Red Hat content will be consumed from {type}.')}
```

webpack/scenes/Subscriptions/Manifest/CdnConfigurationTab/CustomCdnTypeForm.js
```diff
@@ -0,0 +1,148 @@
...
+            values={{
+              type: <strong>{__('Custom CDN Url')}</strong>,
```
jeremylenz, line 64:
```suggestion
              type: <strong>{__('a custom CDN URL')}</strong>,
```

jeremylenz (commented):
Are we sure we want to keep the Red Hat CDN URL field disabled? I remember reading somewhere there's also `cdn6.redhat.com` for ipv6..

parthaa:
> Are we sure we want to keep the Red Hat CDN URL field disabled? I remember reading somewhere there's also `cdn6.redhat.com` for ipv6..

Oh that sounds important. Will look into it.

parthaa:
> Are we sure we want to keep the Red Hat CDN URL field disabled? I remember reading somewhere there's also `cdn6.redhat.com` for ipv6..

So I think we are ok with cdn url field disabled. The reason being that the cdn url is derived from a config setting `SETTINGS[:katello][:redhat_repository_url]`

webpack/scenes/Subscriptions/Manifest/CdnConfigurationTab/CustomCdnTypeForm.js
```diff
@@ -0,0 +1,148 @@
...
+  const firstUpdate = useRef(true);
+  const [sslCaCredentialId, setSslCaCredentialId] = useState(cdnConfiguration.ssl_ca_credential_id);
+
+  useEffect(() => {
+    if (firstUpdate.current) {
+      firstUpdate.current = false;
+      return;
+    }
+    setUpdateEnabled(true);
+  }, [sslCaCredentialId, cdnConfiguration]);
```
jeremylenz, line 42: Can you help me understand why this effect was needed?

jeremylenz (commented):
Code looks good & works well.

Is an SSL CA content credential always required? Would a user ever sync to an insecure CDN?

If a user has no content credential configured, what do you think about linking to the content credential create page? (The link could replace the dropdown in the form which would otherwise be empty anyway.)

also one curiosity question below

parthaa:
> Is an SSL CA content credential always required? Would a user ever sync to an insecure CDN?

CA Credential is not necessary. You can use a http server instead of https or https with no_ca. These cases will not require a CA Credential.

webpack/scenes/Subscriptions/Manifest/CdnConfigurationTab/CustomCdnTypeForm.js
parthaa: Enable/Disable update if `sslCaCredentialId` or `cdnConfiguration` changes? Are you questioning the `firstUpdate.current` part ?

jeremylenz: Right, why did it need an effect and why keep track of if it's the first update?

Is there a way that `updateEnabled` could be turned into a computed value rather than a state? If so it could make things much cleaner.

jeremylenz: https://beta.reactjs.org/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state

jeremylenz:
```diff
diff --git a/webpack/scenes/Organizations/OrganizationSelectors.js b/webpack/scenes/Organizations/OrganizationSelectors.js
index 03868e14dc..f3148ae00b 100644
--- a/webpack/scenes/Organizations/OrganizationSelectors.js
+++ b/webpack/scenes/Organizations/OrganizationSelectors.js
@@ -9,6 +9,7 @@ import {
 } from './OrganizationConstants';
 
 export const selectOrganizationState = state => state.katello.organization;
+export const selectOrgLoading = state => state.katello.organization.loading;
 
 export const selectManifestName = state =>
   selectOrganizationState(state).owner_details?.upstreamConsumer?.name;
diff --git a/webpack/scenes/Subscriptions/Manifest/CdnConfigurationTab/CdnTypeForm.js b/webpack/scenes/Subscriptions/Manifest/CdnConfigurationTab/CdnTypeForm.js
index a56165dcae..47a095e2fe 100644
--- a/webpack/scenes/Subscriptions/Manifest/CdnConfigurationTab/CdnTypeForm.js
+++ b/webpack/scenes/Subscriptions/Manifest/CdnConfigurationTab/CdnTypeForm.js
@@ -17,18 +17,20 @@ import { CDN_URL, CDN } from './CdnConfigurationConstants';
 import { updateCdnConfiguration } from '../../../Organizations/OrganizationActions';
 import {
   selectUpdatingCdnConfiguration,
+  selectOrgLoading,
 } from '../../../Organizations/OrganizationSelectors';
 import './CdnConfigurationForm.scss';
 
 const CdnTypeForm = ({ typeChangeInProgress, onUpdate }) => {
   const dispatch = useDispatch();
   const updatingCdnConfiguration = useSelector(state => selectUpdatingCdnConfiguration(state));
+  const orgIsLoading = useSelector(state => selectOrgLoading(state));
   const performUpdate = () => {
     dispatch(updateCdnConfiguration({
       type: CDN,
     }, onUpdate));
   };
-
+  console.table({ disableForTypeChange: !typeChangeInProgress, updatingCdnConfiguration })
   return (
     <Form isHorizontal>
       <div id="update-hint-cdn" className="margin-top-16">
@@ -68,7 +70,7 @@ const CdnTypeForm = ({ typeChangeInProgress, onUpdate }) => {
           aria-label="update-cdn-configuration"
           variant="secondary"
           onClick={performUpdate}
-          isDisabled={updatingCdnConfiguration || !typeChangeInProgress}
+          isDisabled={updatingCdnConfiguration || orgIsLoading || !typeChangeInProgress}
           isLoading={updatingCdnConfiguration}
         >
           {__('Update')}
```

jeremylenz (approved):
Great work @parthaa!

ACK :+1: :rocket:

PR #10218, opened 2022-08-01
Fixes #35312 - Write contentAccessModeList on org create or update; add checkbox to create screen

jeremylenz:

`contentAccessModeList` is a list of values on the Candlepin owner (what Katello calls an organization) which determines whether we are allowed to enable / disable SCA. Possible values may be

* `'entitlement,org_environment'` - User can disable or enable Simple Content Access at will
* `'entitlement'` - User cannot enable Simple Content Access and will get a Candlepin error if attempted
* `'org_environment'` - User cannot disable Simple Content Access and will get a Candlepin error if attempted

Please note this is in contrast to `contentAccessMode`, which describes whether the owner _currently_ has SCA enabled.

The `contentAccessModeList` may have only one value due to previous Candlepins writing it in the manifest that way. This change sets `contentAccessModeList` for all owners to `"entitlement,org_environment"` so that users don't lose the ability to toggle SCA.

The second change is adding a "Simple Content Access" checkbox on the Organization Create screen.

I asked the Candlepin team and they don't see any issues with writing the contentAccessModeList in this way.

The default value of the checkbox is checked, since that matches how Candlepin creates owners now.

Upgrade to Candlepin 4.2 using instructions from https://github.com/Katello/katello/pull/10208

#### Testing: checkbox
Visit the Organizations -> New page and observe the Simple Content Access checkbox

Create some organizations and verify that the Simple Content Access status matches the checkbox.  ~~(You may wish to wait until https://github.com/Katello/katello/pull/10211 is merged, or rebase on it. Otherwise the only way to see the SCA status for orgs without manifests will be in the Rails console with `org.simple_content_access?`)~~

#### Testing: contentAccessModeList

To view an organization's contentAccessModeList:
```rb
# in Rails console
[21] pry(main)> ownr = Katello::Resources::Candlepin::Owner.find('Default_Organization')
...
[22] pry(main)> ownr['contentAccessModeList']
...
=> "entitlement,org_environment"
```
__Important:__ Be sure that you are viewing the contentAccessModeList value for the owner itself, not its upstreamConsumer.

For organizations with manifests imported, you may see only one value in `contentAccessModeList`.
Toggle SCA for these organizations using the web UI.  Verify that SCA is toggled as expected, and that the `contentAccessModeList` is updated to `'entitlement,org_environment'`.

If you don't have any orgs to test with a single value in contentAccessModeList, you can fake this via
```rb
Katello::Resources::Candlepin::Owner.update('Default_Organization, { 'contentAccessModeList' => 'org_environment' })
```
or
```rb
Katello::Resources::Candlepin::Owner.update('Default_Organization, { 'contentAccessModeList' => 'entitlement' })
```


jeremylenz:
updated the API controller too, should probably test with Hammer

also looks like I need to record VCRs..

chris1984:
> updated the API controller too, should probably test with Hammer
> 
> also looks like I need to record VCRs..

Let's see what happens when candlepin 4.2 is in the repos, i'm getting vcr failures in my tests. the task to republish the repos is still going.

jeremylenz:
> Let's see what happens when candlepin 4.2 is in the repos, i'm getting vcr failures in my tests

The VCR failure I saw is about the owners endpoint, so we may still need to record them.

jeremylenz:
[test katello]

chris1984:
what does docker_sync have to do with anything :fire:

ianballou:
So I'm seeing the following request is causing a VCR error:

```
GET https://node04.jenkins.aws.theforeman.org/pulp/api/v3/repositories/rpm/rpm/?name=2_duplicate
```

It would be worth looking into why `ensure_createable` is throwing an extra GET request: https://ci.theforeman.org/blue/organizations/jenkins/katello-pr-test/detail/katello-pr-test/11403/pipeline/128/#step-132-log-692

app/models/katello/glue/candlepin/owner.rb
```diff
@@ -30,7 +30,6 @@ def candlepin_owner_exists?
       def owner_details
         @owner_details ||= find_owner
         @owner_details['virt_who'] ||= self.subscriptions.using_virt_who.any?
-
```
chris1984, line 33: Any reason for deleting this line, or just cleaning up things?

chris1984 (commented):
Overall looks pretty good, just left a few questions. Will start to test

app/controllers/katello/api/v2/organizations_controller.rb
```diff
@@ -68,9 +68,11 @@ def update
     param :organization, Hash do
       param :label, String, :required => false
     end
+    param :simple_content_access, :bool, :desc => N_('Whether to turn on Simple Content Access for the organization.'), :required => false, :default => true
```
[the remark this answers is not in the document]
jeremylenz: Required is false, will that still cause failures?

app/models/katello/glue/candlepin/owner.rb
jeremylenz: oh I think that was just left over from a log while debugging, I'll put it back in!

app/controllers/katello/api/v2/organizations_controller.rb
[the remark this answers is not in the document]
chris1984: Ignore me, I read required as true, i see default is true. coffee++

app/controllers/katello/api/v2/organizations_controller.rb
```diff
@@ -68,14 +68,16 @@ def update
...
       @organization = Organization.new(resource_params)
-      creator = ::Katello::OrganizationCreator.new(@organization)
-      creator.create!
-      @organization.reload
       # @taxonomy instance variable is necessary for foreman side
       # app/views/api/v2/taxonomies/show.json.rabl is using it.
       @taxonomy = @organization
+      sca = params.key?(:simple_content_access) ? ::Foreman::Cast.to_bool(params[:simple_content_access]) : true
+      creator = ::Katello::OrganizationCreator.new(@taxonomy, sca: sca)
+      creator.create!
```
jeremylenz, line 79: Moving `creator.create!` after `@taxonomy = @organization` is what fixed all the organizations_controller tests. (The problem was that `@taxonomy` was nil for some reason, so `creator.create!` was trying to create based on nothing.)  I don't see any harm in assigning `@taxonomy` earlier, but I may not fully understand why it was that way before. @parthaa @sjha4 thoughts?

jeremylenz:
:green_apple:

app/controllers/katello/api/v2/organizations_controller.rb
sjha4: Not sure but the earlier code was calling reator = ::Katello::OrganizationCreator.new(@organization) and then @organization.reload upon create before assigning to taxonomy..

I don't know enough to opine what issues that might cause.

If this works, you'll be able to maintain the old flow:

creator = ::Katello::OrganizationCreator.new(@organization, sca: sca)

jeremylenz: you're right, that's probably better! Will update

chris1984:
@jeremylenz should I wait till you make a push based on @sjha4 comment or keep on testing?

jeremylenz:
@chris1984 You can keep testing (unless you were testing initial DB seeding which is unlikely)

jeremylenz:
@sjha4 updated :+1: 
@chris1984 no need to re-run any tests you've already done :smile:

chris1984 (approved):
ACK, works great. Thanks @jeremylenz. Checkbox works with different org/sca create combos and the `contentAccessModeList` worked great. Reset my db as well and no issues with the db reset.

PR #10221, opened 2022-08-03
Fixes #35329 - useTableSort should translate params

MariaAga:
It receives the untranlsated value which is not a part of the columnsToSortParams object.
Because of this users cant open Katello tabs in the new host details
Error:
`TableHooks.js:275 Uncaught Error: initialSortColumnName 'Name' must also be defined in columnsToSortParams`

Translating the value of `initialSortColumnName` in `useTableSort`

Thought it would be better to translate at the function declaration instead of in each function call so if someone wrote a new call to `useTableSort` they won't accidentally re-do this bug.
`allColumns[0]` is already translated and it might cause issues to translate it twice.

set locale=cs_cz and open Errate - Module streams tab

### DIFF
--- a/webpack/components/Table/TableHooks.js
+++ b/webpack/components/Table/TableHooks.js
@@ -1,6 +1,7 @@
 import { useState, useRef, useEffect, useCallback, useMemo } from 'react';
 import { isEmpty } from 'lodash';
 import { useLocation } from 'react-router-dom';
+import { translate as __ } from 'foremanReact/common/I18n';
 import { friendlySearchParam } from '../../utils/helpers';
 
 class ReactConnectedSet extends Set {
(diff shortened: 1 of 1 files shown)


webpack/components/Table/TableHooks.js
```diff
@@ -268,12 +269,15 @@ export const useUrlParams = () => {
 export const useTableSort = ({
   allColumns,
   columnsToSortParams,
-  initialSortColumnName = allColumns[0],
+  initialSortColumnName,
 }) => {
-  if (!Object.keys(columnsToSortParams).includes(initialSortColumnName)) {
-    throw new Error(`initialSortColumnName '${initialSortColumnName}' must also be defined in columnsToSortParams`);
+  const tranlatedInitialSortColumnName = initialSortColumnName
```
jeremylenz, line 274:
```suggestion
  const translatedInitialSortColumnName = initialSortColumnName
```

jeremylenz (commented):
Thanks @MariaAga!

Saw the error and confirmed this change fixes it.

Just one spelling nit, looks good otherwise.

webpack/components/Table/TableHooks.js
MariaAga: Thanks, fixed!

jeremylenz (approved):
Thanks @MariaAga!

ACK :+1:

PR #10222, opened 2022-08-03
Fixes #35045 - fixed the issue that all errata are applied when only certain errata selected

lfu:
Add `fetchItemsQuery` to `useBulkSelect` for dropdown filters.

@jeremylenz has done an excellent job to explain the issue. I'd like to save it here for the record. Thanks @jeremylenz.

`fetchBulkParams` returns a scoped search query that matches whatever items are selected.

If you're in regular mode and have ids 1 and 3 selected, it will return `"id ^ (1, 3)"`.
if you're in `selectAll` mode, it uses `exclusionSet` instead. 

Let's say you select all, the front end hasn't downloaded all the IDs yet from all the pages so it doesn't know what they are. 
but then lets say you go to page 4 of 99 and deselect a couple items. It knows you deselected those.  So `searchQueryWithExclusionSet` will return a query like "id !^ (396, 398)" which means "id is not equal to those two that the user deselected". So that search would match all IDs _except_ those.

Now, it gets more complicated when the user has already entered a search in the search box. That's why you see it joined with `searchQuery` [there](https://github.com/Katello/katello/blob/a9e44e26abb667ed474fc562b97172a62ea98a2a/webpack/components/Table/TableHooks.js#L210). It actually joins that `id !^ (396, 398)` with the search query so you'll get something like `id !^ (396, 398) AND name ~ panda` or whatever your search was.

The bug is that it should _also_ do that with whatever filter dropdowns you have selected. Right now it ignores them.
But it should translate those dropdowns into a scoped search query and add that to `searchQueryWithExclusionSet` similar to the way `searchQuery` is now.

Get a host with multiple types of errata
On the new host detail page, filter the errata to only a single type (e.g. Bugfix)
Click the Select All checkbox
Click Apply

Expected: Only the selected errata type will be applied instead of all errata are applied.

### DIFF
--- a/webpack/components/Table/__test__/useBulkSelect.test.js
+++ b/webpack/components/Table/__test__/useBulkSelect.test.js
@@ -1,0 +1,99 @@
+import { act, renderHook } from '@testing-library/react-hooks';
+import { useBulkSelect } from '../TableHooks';
+
+const isSelectable = () => true;
+const idColumn = 'errata_id';
+const metadata = {
+  error: null, selectable: 2, subtotal: 2, total: 2,
+};
+const results = [
+  {
+    errata_id: 'RHSA-2022:2031',
+    id: 311,
+    severity: 'Low',
+    type: 'security',
+  },
+  {
+    errata_id: 'RHSA-2022:2110',
+    id: 17,
+    severity: 'Low',
+    type: 'security',
+  },
+];
+
+it('returns a scoped search string based on inclusionSet', () => {
+  const { result } = renderHook(() => useBulkSelect({
+    results,
+    metadata,
+    idColumn,
+    isSelectable,
+  }));
+
+  act(() => {
+    result.current.selectOne(true, 'RHSA-2022:2031');
+  });
+
+  expect(result.current.fetchBulkParams()).toBe('errata_id ^ (RHSA-2022:2031)');
+});
+
+it('returns a scoped search string based on exclusionSet', () => {
+  const { result } = renderHook(() => useBulkSelect({
+    results,
+    metadata,
+    idColumn,
+    isSelectable,
+  }));
+
+  act(() => {
+    result.current.selectAll(true);
+  });
+
+  act(() => {
+    result.current.selectOne(false, 'RHSA-2022:2031');
+  });
+
+  expect(result.current.fetchBulkParams()).toBe('errata_id !^ (RHSA-2022:2031)');
+});
+
+it('adds search query to scoped search string based on exclusionSet', () => {
+  const { result } = renderHook(() => useBulkSelect({
+    results,
+    metadata,
+    idColumn,
+    isSelectable,
+  }));
+
+  act(() => {
+    result.current.updateSearchQuery('type=security');
+  });
+
+  act(() => {
+    result.current.selectAll(true);
+  });
+
+  act(() => {
+    result.current.selectOne(false, 'RHSA-2022:2031');
+  });
+
+  expect(result.current.fetchBulkParams()).toBe('type=security and errata_id !^ (RHSA-2022:2031)');
+});
+
+it('adds filter dropdown query to scoped search string', () => {
+  const { result } = renderHook(() => useBulkSelect({
+    results,
+    metadata,
+    idColumn,
+    isSelectable,
+    filtersQuery: 'severity=Low',
+  }));
+
+  act(() => {
+    result.current.selectAll(true);
+  });
+
+  act(() => {
+    result.current.selectOne(false, 'RHSA-2022:2031');
+  });
+
+  expect(result.current.fetchBulkParams()).toBe('severity=Low and errata_id !^ (RHSA-2022:2031)');
+});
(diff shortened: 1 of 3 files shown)


lfu:
Package upgrade does not have the same issue.
Seems errata is the only one affected.

webpack/components/extensions/HostDetails/Tabs/ErrataTab/ErrataTab.js
```diff
@@ -155,6 +167,7 @@ export const ErrataTab = () => {
     results,
     metadata,
     idColumn: 'errata_id',
+    fetchItemsQuery: fetchItemsQuery(),
```
jeremylenz, line 170:
```suggestion
    dropdownsQuery: dropdownsQuery(),
```
Since `searchQuery` is what the user entered in the search box, I feel `dropdownsQuery` or `filterDropdownsQuery` may be a better name here, since it's what the user has selected in the dropdowns.

webpack/components/extensions/HostDetails/Tabs/ErrataTab/ErrataTab.js
```diff
@@ -120,6 +120,18 @@ export const ErrataTab = () => {
     initialSortColumnName: 'Errata',
   });
 
+  const fetchItemsQuery = () => {
+    let query = '';
```
jeremylenz, line 124:
```suggestion
    let query = [];
```
Doing this with an array (and then `return query.join(' & ')`) will be a bit cleaner and easier to read.

jeremylenz (commented):
Tested and works well, thanks @lfu!

Left a couple suggestions below.

Also, it seems we don't have any test coverage for `TableHooks.js`. It's such an important area of the code; I think it would be great if you could add a few tests for `fetchBulkParams` describing different search inputs and outputs:

* returns a scoped search string based on selectionSet
* returns a scoped search string based on exclusionSet
* adds search query to scoped search string based on exclusionSet
* adds filter dropdown search query to scoped search string

You won't have to render any React components - these will be pure JavaScript tests, so they should be pretty lightweight and easy to write!

jeremylenz:
Seems like the issue also happens in Repository sets, could you look into that?

webpack/components/extensions/HostDetails/Tabs/ErrataTab/ErrataTab.js
lfu: I don't like to include component type in the name. Suppose in the future we'd like to add a checkbox to the filters then the name would become not appropriate.
How about `filtersQuery`?

jeremylenz: I'm fine with `filtersQuery` :+1:

lfu:
> Seems like the issue also happens in Repository sets, could you look into that?

Repository sets does have the same issue.
Unless we fix [this method](app/controllers/katello/api/v2/host_subscriptions_controller.rb) first to accept override status as scoped search attribute, we can't apply the `filtersQuery` change to Repository sets.

jeremylenz:
@lfu can you open a Redmine for the repository sets issue?

webpack/components/extensions/HostDetails/Tabs/ErrataTab/ErrataTab.js
```diff
@@ -120,6 +120,17 @@ export const ErrataTab = () => {
...
 
+  const filtersQuery = () => {
+    let query = [];
+    if (errataTypeSelected !== ERRATA_TYPE) {
+      query.push(`type=${TYPES_TO_PARAM[errataTypeSelected]}`);
+    }
+    if (errataSeveritySelected !== ERRATA_SEVERITY) {
+      query.push(`severity=${SEVERITIES_TO_PARAM[errataSeveritySelected]}`);
+    }
+    return query.join('&');
```
jeremylenz, line 131: I'm not sure if `&` works the same as `and`. And we use `and` in most other places, I think, including `fetchBulkParams`.

```suggestion
    return query.join(' and ');
```

webpack/components/Table/__test__/useBulkSelect.test.js
```diff
@@ -0,0 +1,89 @@
...
+
+  act(() => {
+    result.current.selectAll(true);
+  });
+
+  act(() => {
+    result.current.updateSearchQuery('errata_id=RHSA-2022:2031');
+  });
+
+  expect(result.current.fetchBulkParams()).toBe('errata_id=RHSA-2022:2031');
```
jeremylenz, line 72: I think this test and the test below would make a bit more sense to the reader if it first had a non-blank exclusion set to add the query to. As it is now, it starts with nothing, so it doesn't really illustrate the "adding" part.

jeremylenz (commented):
Thanks for adding tests @lfu! just a couple nits below, but this is looking good

lfu:
> @lfu can you open a Redmine for the repository sets issue?

https://projects.theforeman.org/issues/35357

webpack/components/Table/__test__/useBulkSelect.test.js
lfu: `exclusionSet` is checked with when `selectAllMode` is `true`. 
I can add item to `exclusionSet` if you prefer an non-empty `exclusionSet`.

jeremylenz: Yeah, I'd prefer to see the expectation be "xxx and xxx" instead of just "xxx"

jeremylenz (approved):
Changes look good and still works great, thanks @lfu!

ACK :+1:

PR #10223, opened 2022-08-03
Fixes #35410 - Fix a pathing issue on exports

parthaa:
When exporting a content view version with a space its name we include that space in the generated path. Its not necessarily a bug but ideally we'd like path to be something like -> 
`/var/lib/pulp/exports/Test_Organization/Test_CV/1.0/2022-08-03T00-17-11-00-00/metadata.json`
vs 
`/var/lib/pulp/exports/Test_Organization/Test CV/1.0/2022-08-03T00-17-11-00-00/metadata.json`

The cv name without and with space.

This issue was introduced when the previous syncable export work got merged and got caught on bats tests. So useful.

- Create a CV name with a space in it
- Add a simple custom repo or rh repo with `immediate` download policy
- Publish
- Export the content view version via hammer. 
 - `hammer content-export complete version --id=<cvv-id>`
 
Before this PR, you 'd see something like

`/var/lib/pulp/exports/Test_Organization/Test CV/1.0/2022-08-03T00-17-11-00-00/metadata.json` with the space

After this PR

`/var/lib/pulp/exports/Test_Organization/Test_CV/1.0/2022-08-03T00-17-11-00-00/metadata.json` with the space

### DIFF
--- a/app/services/katello/pulp3/content_view_version/export.rb
+++ b/app/services/katello/pulp3/content_view_version/export.rb
@@ -53,9 +53,9 @@ module Katello
 
         def generate_exporter_path
           return base_path if base_path
-          export_path = "#{content_view_version.content_view}/#{content_view_version.version}/"
-          export_path += "#{destination_server}/" unless destination_server.blank?
-          export_path += "#{date_dir}".gsub(/\s/, '_')
+          export_path = "#{content_view_version.content_view.label}/#{content_view_version.version}/"
+          export_path += "#{destination_server}/".gsub(/\s/, '_') unless destination_server.blank?
+          export_path += "#{date_dir}"
           @base_path = "#{Setting['pulpcore_export_destination']}/#{content_view_version.organization.label}/#{export_path}"
         end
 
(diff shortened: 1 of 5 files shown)


ianballou:
Looks like a VCR error was hit

parthaa:
[test katello]

parthaa:
[test katello]

app/services/katello/pulp3/content_view_version/export.rb
```diff
@@ -55,8 +55,8 @@ def generate_exporter_path
           return base_path if base_path
           export_path = "#{content_view_version.content_view}/#{content_view_version.version}/"
```
sjha4, line 56: `#{content_view_version.content_view} `evaluates to name but it is also rails magic..would make it clearer to use label to avoid whitespaces etc..

parthaa: Good suggestion. Also gsubing at the end is a bad idea, because it may underscore  `Setting['pulpcore_export_destination']`  which means it will error out if that directory had a space in it.

parthaa: updated

sjha4 (approved):
Change looks good and works as advertised.
Before change: `/var/lib/pulp/exports/Default_Organization/An example content view/1.0/2022-08-23T15-08-26-00-00`
After change: `/var/lib/pulp/exports/Default_Organization/An_example_content_view/1.0/2022-08-23T15-07-33-00-00`

PR #10224, opened 2022-08-05
Fixes #35336 - System purpose card should not show when host is unregistered

jeremylenz:

The system purpose card was trying to render even for unregistered hosts. This breaks the entire host details page.

This change will check if the host is registered before rendering the system purpose card.

Check that the new host detail page renders properly with both registered and unregistered hosts.

### DIFF
--- a/webpack/components/extensions/HostDetails/Cards/SystemPurposeCard/SystemPurposeCard.js
+++ b/webpack/components/extensions/HostDetails/Cards/SystemPurposeCard/SystemPurposeCard.js
@@ -27,7 +27,7 @@ import { propsToCamelCase } from 'foremanReact/common/helpers';
 import './SystemPurposeCard.scss';
 import SystemPurposeEditModal from './SystemPurposeEditModal';
 import { selectHostDetailsStatus } from '../../HostDetailsSelectors';
-import { hasRequiredPermissions } from '../../hostDetailsHelpers';
+import { hasRequiredPermissions, hostIsNotRegistered } from '../../hostDetailsHelpers';
 
 const SystemPurposeCard = ({ hostDetails }) => {
   const showEditButton = hasRequiredPermissions(['edit_hosts'], hostDetails?.permissions);
@@ -56,6 +56,8 @@ const SystemPurposeCard = ({ hostDetails }) => {
       </GridItem>
     );
   }
+
+  if (hostIsNotRegistered({ hostDetails })) return null;
 
   return (
     <GridItem rowSpan={1} md={6} lg={4} xl2={3}>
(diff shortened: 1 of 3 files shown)


jeremylenz:
Updated to also hide the HW Properties card for unregistered hosts

webpack/components/extensions/HostDetails/DetailsTabCards/HwPropertiesCard.js
```diff
@@ -32,6 +33,7 @@ HostDisks.defaultProps = {
 };
 
 const HwPropertiesCard = ({ isExpandedGlobal, hostDetails }) => {
+  if (hostIsNotRegistered({ hostDetails })) return null;
```
lfu, line 36: 👍🏻

webpack/components/extensions/HostDetails/Cards/SystemPurposeCard/SystemPurposeCard.js
```diff
@@ -57,6 +57,8 @@ const SystemPurposeCard = ({ hostDetails }) => {
     );
   }
 
+  if (hostIsNotRegistered({ hostDetails })) return null;
```
lfu, line 60: Why not check it at the beginning of the function?

jeremylenz: @lfu updated :+1:

jeremylenz: I guess this is why:
```
Error:   48:29  error  React Hook "useSelector" is called conditionally. React Hooks must be called in the exact same order in every component render  react-hooks/rules-of-hooks
Error:   51:33  error  React Hook "useState" is called conditionally. React Hooks must be called in the exact same order in every component render     react-hooks/rules-of-hooks
```

jeremylenz:
Tried moving that check to the beginning, but that would mean calling hooks conditionally. So I changed it back.

webpack/components/extensions/HostDetails/Cards/SystemPurposeCard/SystemPurposeCard.js
lfu: We can move them together.
```
const SystemPurposeCard = ({ hostDetails }) => {
  const hostDetailsStatus = useSelector(selectHostDetailsStatus);
  const [editing, setEditing] = useState(false);
  if (hostIsNotRegistered({ hostDetails })) return null;
```

jeremylenz: But then we would have logic in the middle of assigning all of the consts. Seems cleaner to just assign all the consts, and then have your `if`s..

lfu (approved):
LGTM 👍🏻 

APJ

jeremylenz:
[test katello]

PR #10225, opened 2022-08-05
Fixes #35337 - Index ModuleStreamErratumPackages correctly in first run

sjha4:
Allow ErratumPackages to be saved before building and saving ModuleStreamErratumPackages

1. Sync Appstream repo for example.
2. In console: `Katello::ModuleStreamErratumPackage.count` 
3. This will return 0 after first sync on master.
4. On this branch, it should correctly index and Katello::ModuleStreamErratumPackage.count should be > 0


chris1984 (approved):
ACK, looks good and works per testing steps.

PR #10220, opened 2022-08-02
Fixes #35323 - CV Version Compare - add view by different/same

tazhibaevaaliya:
#### What are the changes introduced in this pull request?
Added the functionality for the dropdown for viewing by "Different/Same/All" for the content types comparison
#### What are the testing steps for this pull request?
**"View by Different"**
1. Select two versions to compare
2. Hit "Compare" button
3. Select "Different" in the dropdown "View by" 
4. Check that all the tabs show the content which is in either of the versions but not both 

**"View by Same"**
1. Select two versions to compare
2. Hit "Compare" button
3. Select "Same" in the dropdown "View by" 
4. Check that all the tabs show the content which is in common for both versions

**"View by All"**
1. Select two versions to compare
2. Hit "Compare" button
3. Select "All" in the dropdown "View by" 
4. Check that all the tabs show the all the content comparison for both versions

PR #10227, opened 2022-08-08
Fixes #35323 - CV Version Compare - add view by different/same

sjha4:
Added a new commit with the 2 test fixes.
Follow original PR here: Original PR from Aliya here: https://github.com/Katello/katello/pull/10220


jeremylenz (approved):
Thanks @sjha4 (and @tazhibaevaaliya) :)  Works great.

Found a few small issues, but they're not enough to block merging. I will let you decide if you want to address them here or later.

1. Clicking on the Versions tab should take me back to the versions list. It does nothing.
2. The first version I click becomes versionOne, so I got "Version 4.0" and "Version 3.0" in my table.  But I would expect to see the older version on the left. Should we sort the initial versions?
3. The vertical alignment of the dropdown isn't consistent:
Content-View-Details

sjha4:
> 2. The first version I click becomes versionOne, so I got "Version 4.0" and "Version 3.0" in my table.  But I would expect to see the older version on the left. Should we sort the initial versions?
> 3. The vertical alignment of the dropdown isn't consistent:

Pushed changes to address these.

> 1. Clicking on the Versions tab should take me back to the versions list. It does nothing.

This is a bit complicated. Clicking versions tab from compare page actually stays on the same URL and hence path doesn't change. I can't think of an easy way to get this to work other than to add a new URL path for compare component.

PR #10231, opened 2022-08-11
Fixes #35360 - Add all certificates from provided bundle to cert_store

jpasqualetto:
Current code only adds the first certificate from a bundle to the
cert_store. With this change, it will iterate over all certificates
from a provided CA bundle and add each one of them o the cert_store
used by katello to communicate with an upstream server.

### DIFF
--- a/app/lib/katello/resources/cdn.rb
+++ b/app/lib/katello/resources/cdn.rb
@@ -42,13 +42,18 @@ module Katello
 
           if options[:ssl_ca_cert].present?
             begin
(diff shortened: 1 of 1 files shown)


parthaa:
Hoping  https://github.com/theforeman/foreman/pull/9348 makes it in soon. We'd then use the add_file method.

jeremylenz:
@parthaa That's in now..

parthaa:
Closing this because -> https://github.com/Katello/katello/pull/10217 will address it.

PR #10236, opened 2022-08-15
Fixes #35368 - Prevent multiple envs in host registration

chris1984:

* New subscription manager version starting with EL8 now support being registered to multiple envs. We have a new version of Candlepin that allows this, until we add in support for multiple envs, we want to throw an error if multiples are being passed in.

* For hosts with EL8+ subscription-manager has changed the `environment_id` param to `environments` so added that in the controller if/else logic and added a line to throw an error if multiple envs are being passed in.
* I tried registering with multiple envs with EL7 which uses `environment_id` and it does not support passing in multiples so did not touch that legacy area.

* Check out PR
* Create a content view so you have a 2nd environment
* Spin up an EL8 client and try to register with 1 environment and make sure that works
* Unregister and try to register again passing in multiple envs IE `Library, Library/view`
* Verify you get a failure on your client and see the error in the rails logs

### DIFF
--- a/app/controllers/katello/api/rhsm/candlepin_proxies_controller.rb
+++ b/app/controllers/katello/api/rhsm/candlepin_proxies_controller.rb
@@ -313,6 +313,9 @@ module Katello
 
       if params.key?(:environment_id)
         environment = get_content_view_environment("cp_id", params[:environment_id])
+      elsif params.key?(:environments)
+        fail HttpErrors::BadRequest, _('Multiple environments are not supported.') if params['environments'].length > 1
+        environment = get_content_view_environment("cp_id", params['environments'].first['id'])
       elsif params.key?(:organization_id) && !params.key?(:environment_id)
         organization = Organization.current
         environment = organization.library.content_view_environment
--- a/test/controllers/api/rhsm/candlepin_proxies_controller_test.rb
+++ b/test/controllers/api/rhsm/candlepin_proxies_controller_test.rb
@@ -96,6 +96,27 @@ module Katello
         assert_response :success
       end
 
+      it "should register with new environments param" do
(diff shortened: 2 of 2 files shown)


app/controllers/katello/api/rhsm/candlepin_proxies_controller.rb
```diff
@@ -313,6 +313,9 @@ def find_content_view_environment
 
       if params.key?(:environment_id)
         environment = get_content_view_environment("cp_id", params[:environment_id])
+      elsif params.key?(:environments)
+        fail HttpErrors::BadRequest, _('Multiple environments are not currently supported at this time.') if params['environments'].length > 1
```
jeremylenz, line 317:
```suggestion
        fail HttpErrors::BadRequest, _('Multiple environments are not supported.') if params['environments'].length > 1
```

jeremylenz (commented):
Thanks @chris1984!

Would like to see test coverage of this as well.

chris1984:
@jeremylenz updated

jeremylenz (approved):
Thanks @chris1984!

✔️ `--environments Library,Library/CV_One` errors as expected
```
[root@rhel8c vagrant]# subscription-manager register --username admin --password changeme --environments Library,Library/CV_One --org Default_Organization
Registering to: centos8-katello-devel-stable.example.com:443/rhsm
Multiple environments are not supported. (HTTP error code 400: Bad Request)
```
✔️ `--environment` still registers as expected
✔️ `--environments` registers as expected when passing a single environment

Nice work! Ack 👍

chris1984:
[test katello]

parthaa:
[test katello]

PR #10237, opened 2022-08-15
Fixes #35369 - Moved the syncable repo exports to a different cv

parthaa:

Whenever we do an export repository or library we would
1) Find or Create the associated repo/library cv
2) Publish the version
3) Export the published version in a syncable format

With the addition of syncable format we do the same.
This has some unintended consequences
For example
1) Export the repository in an importable format
2) Update the repository contents
3) Now export the same repo in syncable format
4) Finally do an incremental export of the same repo
Notice that it has 0 content to export. That is because it is
incrementally exporting between the repo state in 3 and 4 instead of 1
and 4

This commit fixes this by making the syncable exports happen in an
entirely different content view, so that syncable and non syncable
exports are not mixed.

1) Export the repository in an importable format => `hammer content-export complete repository --id=<repo-id>`
2) Update the repository contents 
3) Now export the same repo in syncable format =>  `hammer content-export complete repository --id=<repo-id> --format=syncable`
4) Finally do an incremental export of the same repo  => `hammer content-export incremental repository --id=<repo-id>`
Notice that it has 0 content to export. That is because it is
incrementally exporting between the repo state in 3 and 4 instead of 1
and 4

Follow the same steps with this PR checked out on a New Repository. 
It should correctly incrementally export between the repo in 1 and 4

### DIFF
--- a/app/controllers/katello/api/v2/content_views_controller.rb
+++ b/app/controllers/katello/api/v2/content_views_controller.rb
@@ -66,7 +66,7 @@ module Katello
 
     def index_relation
       content_views = ContentView.readable
-      content_views = content_views.not_generated_for_repository unless Foreman::Cast.to_bool(params[:include_generated])
+      content_views = content_views.ignore_generated unless Foreman::Cast.to_bool(params[:include_generated])
       content_views = content_views.where(:organization_id => @organization.id) if @organization
       content_views = content_views.in_environment(@environment) if @environment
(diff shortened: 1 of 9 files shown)


sjha4:
Code change makes sense based on description of the PR. Starting to test..Thanks @parthaa for the commands to ctrl c + ctrl v..

parthaa:
> Code change makes sense based on description of the PR..Could use an assert test to see if the correct CVs are being picked for each format.

Added tests

sjha4 (approved):
APJ..works as described and code change looks good. 👍🏼

sjha4:
[test katello]

PR #10242, opened 2022-08-17
Fixes #35382 - Add rake task to refresh repos on all pulp proxies

chris1984:
Co-authored-by: Joniel Pasqualetto jpasqual@redhat.com

* When changing the certs from self installed certs to ones signed by a custom CA, the problem is that the `ca_cert` on the remote (inside pulp3) does not get updated after changing the certificate.
* Added a rake task to loop through the proxies and katello to refresh the repos

* N/A

* Check out PR
* Try running the rake task and see if you see the task get generated in the UI/logs
* Bonus if you have a smart proxy attached to your devel box to make sure it hits that.

### DIFF
--- a/lib/katello/tasks/refresh_repos.rake
+++ b/lib/katello/tasks/refresh_repos.rake
@@ -1,0 +1,8 @@
+namespace :katello do
+  desc 'Refresh all repositories in all smart proxies and main server'
+  task :refresh_repos => ["environment", "dynflow:client"] do
+    User.current = User.anonymous_api_admin
+    ::ForemanTasks.async_task(::Actions::BulkAction, ::Actions::Pulp3::Orchestration::Repository::RefreshRepos, SmartProxy.all)
+    puts _("Repos are being refreshed in the background.")
+  end
+end
--- a/test/lib/tasks/refresh_repos_test.rb
+++ b/test/lib/tasks/refresh_repos_test.rb
@@ -1,0 +1,18 @@
+require 'katello_test_helper'
+require 'rake'
+
+module Katello
+  class RefreshReposTest < ActiveSupport::TestCase
+    def setup
+      Rake.application.rake_require 'katello/tasks/refresh_repos'
+      Rake::Task['katello:refresh_repos'].reenable
+      Rake::Task.define_task(:environment)
+      Rake::Task.define_task('dynflow:client')
+    end
+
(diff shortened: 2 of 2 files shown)


lib/katello/tasks/refresh_repos.rake
```diff
@@ -0,0 +1,8 @@
+namespace :katello do
+  task :refresh_repos => ["environment"] do
```
sjha4, line 2:
```suggestion
  desc 'Refresh all repositories in all smart proxies and main server'
  task :refresh_repos => ["environment"] do
```

lib/katello/tasks/refresh_repos.rake
```diff
@@ -0,0 +1,8 @@
...
+    User.current = User.anonymous_api_admin
+    SmartProxy.all.each do |proxy|
+      ForemanTasks.sync_task(Actions::Pulp3::Orchestration::Repository::RefreshRepos, proxy)
+    end
```
sjha4, line 6: Thoughts on using async bulk action and run this in the background?
```
::ForemanTasks.async_task(::Actions::BulkAction, ::Actions::Pulp3::Orchestration::Repository::RefreshRepos, proxies)
```

lib/katello/tasks/refresh_repos.rake
```diff
@@ -0,0 +1,7 @@
+namespace :katello do
+  desc 'Refresh all repositories in all smart proxies and main server'
+  task :refresh_repos => ["environment"] do
```
sjha4, line 3:
```suggestion
  task :refresh_repos => ["environment", "dynflow:client"] do
```

lib/katello/tasks/refresh_repos.rake
```diff
@@ -0,0 +1,7 @@
...
+    User.current = User.anonymous_api_admin
+    ::ForemanTasks.async_task(::Actions::BulkAction, ::Actions::Pulp3::Orchestration::Repository::RefreshRepos, SmartProxy.all)
```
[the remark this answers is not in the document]
sjha4: This could use a message displayed to the user. We do it in other tasks with async actions.

chris1984:
@sjha4 updated

sjha4 (approved):
Looks good and works as expected. Adds and runs tasks on smart proxies to refresh repos.
Ack 👍🏼

PR #10243, opened 2022-08-18
Fixes #35386 - Prevents AK destroy if associated to HG

parthaa:
Adds validation to AK destroy to ensure it won't get deleted if any hostgroup is using it
Also prints a helpful message giving the user the parameter to  search for.

1) Create an AK
2) Create a hostgroup and associate the AK to it
3) Delete AK

Before PR:
This delete would be successful but the hostgroup would have a dangling reference to the deleted activation key.

After PR:
You should see a message along 
```
This activation key is associated to one or more Hosts/Hostgroups. Search and unassociate Hosts/Hostgroups using params.kt_activation_keys ~ "<ak name>" before deleting.
````

### DIFF
--- a/app/models/katello/activation_key.rb
+++ b/app/models/katello/activation_key.rb
@@ -28,7 +28,7 @@ module Katello
     alias_method :lifecycle_environment, :environment
 
     before_validation :set_default_content_view, :unless => :persisted?
-
+    before_destroy :validate_destroyable!
     accepts_nested_attributes_for :purpose_addons
 
     validates_lengths_from_database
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by parthaa** [comment]
[test katello]

File: app/models/katello/activation_key.rb [hunk 1]
```diff
@@ -177,6 +177,17 @@ def self.find_by_purpose_addons(_key, operator, value)
       return_activation_keys_by_id(activation_keys.pluck(:id))
     end
 
+    def validate_destroyable!
+      unless organization.being_deleted?
+        if Parameter.where(name: 'kt_activation_keys').pluck(:value).any? { |value| value.split(",").include?(name) }
+          fail _("This activation key is associated to one or more Hosts/Hostgroups. "\
+                  "Search and unassociate Hosts/Hostgroups using params.kt_activation_keys ~ \"%{name}\" "\
```
lfu, line 184: How/where should the users do the search and disassociation?

chris1984 (approved):
ACK works well:

```bash
An error occurred removing the Activation Key: This activation key is associated to one or more Hosts/Hostgroups. Search and unassociate Hosts/Hostgroups using params.kt_activation_keys ~ "test" before deleting.
```

app/models/katello/activation_key.rb
lfu: For the record, `params.kt_activation_keys ~ "key1"` works in the hostgroups'  search box.

lfu (approved):

PR #10244, opened 2022-08-19
Fixes #35235 - add name/label field validation to CreateContentViewForm

lfu:
Add name/label field validation to CreateContentViewForm

- Create Content View
- Name could have any characters. There is no validation from UI side at all. But the special characters in name would be replaced with `_` in label.
- Input `Label` with special characters other than `-` and `_` which should make an error message under the Label field and the `Create Content View` button should be disabled.

### DIFF
--- a/webpack/scenes/ContentViews/Create/CreateContentViewForm.js
+++ b/webpack/scenes/ContentViews/Create/CreateContentViewForm.js
@@ -25,6 +25,18 @@ const CreateContentViewForm = ({ setModalOpen }) => {
   const [redirect, setRedirect] = useState(false);
   const [saving, setSaving] = useState(false);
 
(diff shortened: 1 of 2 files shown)


sjha4:
So this PR does some things we don't necessarily need. We don't want to disallow special characters in the CV name as that would be a behavior change..We want to sub those special characters to _ in the label field.

We need to change the regex sub here in the effect to replace special characters in the label. 
https://github.com/Katello/katello/pull/10244/files#diff-60b7be6db7f763676158b3f98e13c95d3ed9db4b69a4b42a744622d0b54cb5c3R81

lfu:
There is no intention for any behavior change regarding CV name in this PR. Nice catch, @sjha4

sjha4 (approved):
Works great!

PR #10256, opened 2022-08-30
Fixes #35412 - Allow syncing collections without tags

sjha4:
Allow syncing collections without tags
1. Create an ansible repo with upstream url as https://galaxy.ansible.com/ and requirements field as:
```
---
collections:
- name: alikins.collection_inspect
```
2. Sync repo

Without this PR, you'll see the error in index_content step of sync: 
```
ArgumentError
Empty list of attributes passed
```

With this PR: Sync should happen successfully.


chris1984 (approved):
ACK, with patch, repo syncs correctly.

PR #10257, opened 2022-08-30
Refs #35233 - ignore docker tags that have no meta tags during switchover

ianballou:
This PR builds on https://github.com/Katello/katello/pull/10198

During `combine_duplicate_docker_tags`, one duplicate is chosen to be the `main_tag` and the others are updated so they can be deleted.  After the cleanup script runs from #10198 , there are docker tags that have no schema 1 or schema 2 `DockerMetaTags`.  In `combine_duplicate_docker_tags`, the duplicate chosen to be "the one" is simply the last one on the list. This chosen tag could be one without a schema 1 or schema 2 meta tag, which means that the other tags won't be updated correctly (assuming they do have a schema 1 or schema 2 meta tag).  This, in turn, means that there will be `RepositoryDockerMetaTag` references still floating around for the duplicates, and thus they cannot be deleted.

The data could look like the following (after the fix script runs):

```
----------------------------------------
irb(main):195:0> Katello::DockerTag.where(:migrated_pulp3_href => "/pulp/api/v3/content/container/tags/41c251e7-7405-4f35-8969-f7cfdc12de48/").includes(:schema1_meta_tag, :schema2_meta_tag).to_a.each {|tag| pp "tag_id: #{tag.id}, schema1: #{tag.schema1_meta_tag.try(:id)}, schema2: #{tag.schema2_meta_tag.try(:id)}"}
"tag_id: 5798286, schema1: , schema2: " 
"tag_id: 477107, schema1: , schema2: 218462"
"tag_id: 499350, schema1: , schema2: "
"tag_id: 5923743, schema1: , schema2: 2567691"
"tag_id: 5780071, schema1: , schema2: "
"tag_id: 5840200, schema1: , schema2: 2610528"
"tag_id: 5759496, schema1: , schema2: "
"tag_id: 5950913, schema1: , schema2: "   <========= last duplicate tag has no schema1 or schema2 
----------------------------------------
```

This PR fixes `combine_duplicate_docker_tags` to choose a `main_tag` that actually has a meta tag associated. The empty ones are ignored.

I haven't seen it, but it might be possible for all of the duplicates to have no meta tags. If this is the case, we fall back to the old `pop` method.  It shouldn't matter because no docker meta tags means no `RepositoryDockerMetaTags`, which is the model that causes the migration error.

1) Sync some docker repo on `centos7-katello-3.18-stable`

2) Pick a docker tag that has a schema 2 meta tag and duplicate it:
```ruby
tag = ::Katello::DockerMetaTag.first.schema2
dup = tag.dup
dup.pulp_id = "dup-tag"
dup.save!
dup.update(migrated_pulp3_href: 'dup-href')
tag.update(migrated_pulp3_href: 'dup-href')
```
The duplicate will have no schema 1 or schema 2 meta tag, so it replicates the error conditions.

3) Use the following code to check the duplicate tag ordering. If the duplicate tag isn't last, it won't be picked up, and the error won't be replicated.  If the order is incorrect, hack `combine_duplicate_docker_tags` with a `main_tag = tags.reverse.pop`.  That will ensure that the duplicate is grabbed if the ordering is wrong.
```ruby
Katello::DockerTag.where(:migrated_pulp3_href => "dup-href").includes(:schema1_meta_tag, :schema2_meta_tag).to_a.each {|tag| pp "tag_id: #{tag.id}, schema1: #{tag.schema1_meta_tag.try(:id)}, schema2: #{tag.schema2_meta_tag.try(:id)}"}
```
4) Run `combine_duplicate_docker_tags`:
```ruby
Katello::Pulp3::MigrationSwitchover.new(SmartProxy.pulp_primary).combine_duplicate_docker_tags
```
You should see an error like the following:
```
ActiveRecord::InvalidForeignKey: PG::ForeignKeyViolation: ERROR:  update or delete on table "katello_docker_meta_tags" violates foreign key constraint "fk_rails_af711a226d" on table "katello_repository_docker_meta_tags"
DETAIL:  Key (id)=(1) is still referenced from table "katello_repository_docker_meta_tags".

from /home/vagrant/foreman/.vendor/ruby/2.5.0/gems/activerecord-6.0.3.4/lib/active_record/connection_adapters/postgresql_adapter.rb:675:in `exec_params'
Caused by PG::ForeignKeyViolation: ERROR:  update or delete on table "katello_docker_meta_tags" violates foreign key constraint "fk_rails_af711a226d" on table "katello_repository_docker_meta_tags"
DETAIL:  Key (id)=(1) is still referenced from table "katello_repository_docker_meta_tags".

from /home/vagrant/foreman/.vendor/ruby/2.5.0/gems/activerecord-6.0.3.4/lib/active_record/connection_adapters/postgresql_adapter.rb:675:in `exec_params'
```

5) Apply this PR.  Don't forget to add the `reverse` back in if it was needed.

6) Re-run `combine_duplicate_docker_tags`.  It should run without error.

### DIFF
--- a/app/services/katello/pulp3/migration_switchover.rb
+++ b/app/services/katello/pulp3/migration_switchover.rb
@@ -116,7 +116,13 @@ module Katello
         to_delete = []
         Katello::DockerTag.having("count(migrated_pulp3_href) > 1").group(:migrated_pulp3_href).pluck(:migrated_pulp3_href).each do |duplicate_href|
           tags = Katello::DockerTag.where(:migrated_pulp3_href => duplicate_href).includes(:schema1_meta_tag, :schema2_meta_tag).to_a
-          main_tag = tags.pop
+          # The duplicates should either have a schema 1 meta tag or a schema 2 meta tag. Skip those with neither.
+          main_tag = tags.detect { |tag| tag.schema1_meta_tag.present? || tag.schema2_meta_tag.present? }
+          if main_tag.present?
+            tags -= [main_tag]
+          else
+            main_tag = tags.pop
+          end
           main_meta_v1 = main_tag.schema1_meta_tag
           main_meta_v2 = main_tag.schema2_meta_tag
 


parthaa (approved):
ACK works well

PR #10258, opened 2022-08-31
Fixes #35693 - Use optimize and complete sync for APT repos

quba42:

They allow for Katello to make use of the pulp_deb optimize sync mode, which uses exactly the same interface as the pulp_rpm optimize sync.

This requires pulp_deb 2.20.0 and pulpcore 3.21 as well as updated client rubygems. These versions will probably not be available in Katello for some time, so this PR is more for discussion for now.

There are three places where these changes can be used from:

- smart proxy page > Synchronize (drop down) > Optimize Sync or Complete Sync
- Repository page (of a deb type repo) > Select Action > Advanced Sync > Optimize Sync or Complete Sync > Sync
- Products page > Select products that include deb type repos > Select Action > Advanced Sync > Optimize Sync or Complete Sync

In each case, the corresponding underlying pulp sync task needs to be called with the correctly set `optimize` flag.

### DIFF
--- a/app/controllers/katello/api/v2/products_bulk_actions_controller.rb
+++ b/app/controllers/katello/api/v2/products_bulk_actions_controller.rb
@@ -34,7 +34,7 @@ module Katello
       syncable_products = @products.syncable
       syncable_roots = RootRepository.where(:product_id => syncable_products).has_url
 
-      syncable_roots = syncable_roots.yum_type if skip_metadata_check || validate_contents
+      syncable_roots = syncable_roots.skipable_metadata_check if skip_metadata_check || validate_contents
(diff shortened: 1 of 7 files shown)


sjha4:
[test katello]

sjha4 (commented):
This looks good to me. Even though this doesn't add the flag to smart proxy mirror repos, I'll hold off on merging this as this flag not being available in older pulp_deb gems. I am working on a patch to fix proxy syncs without the optimize flag.

sjha4:
[test katello]

sjha4 (approved):
APJ..Not expecting VCR failures on this because there are no changes to the tests but re-running suite just to make sure.

PR #10259, opened 2022-08-31
Fixes #35459 - Non-enabled repository types make it into the apipie help-text

ianballou:
Removes references to defined repository types in the repositories API apipie.  These repository types may not be enabled on the server and thus are misleading. Instead, we point the user to use the repository types API.

1) Look at the apipie help-text for the following, ensuring that repository or content unit types are not there:
```
GET /repositories
POST /repositories
PUT /repositories/:id/remove_content
POST /repositories/:id/upload_content
```
2) Try sending the `content_type` or `with_content` parameters to ensure they still work.  Try sending types that don't exist. I had to move the validations into their own manual checks.

3) Check apipie in general to ensure types like "deb" don't show up if they're not enabled.

TODO:
- [x] Make a hammer PR that either shows the available types or suggests a different command to find them.

### DIFF
--- a/app/controllers/katello/api/v2/repositories_controller.rb
+++ b/app/controllers/katello/api/v2/repositories_controller.rb
@@ -81,7 +81,7 @@ module Katello
     def_param_group :repo_create do
       param :label, String, :required => false
       param :product_id, :number, :required => true, :desc => N_("Product the repository belongs to")
-      param :content_type, RepositoryTypeManager.creatable_repository_types(false).keys, :required => true, :desc => N_("type of repo")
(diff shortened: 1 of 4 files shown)


ianballou:
I noticed that `--with-content` allows the user to include non-indexed content types, which cause unhandled errors. I'm going to remove the ability to pass unindexed types into that field.

ianballou:
Related hammer PR: https://github.com/Katello/hammer-cli-katello/pull/864

chris1984 (approved):
ACK, works well no issues with validations. Thanks @ianballou

ekohl:
As noted on IRC, since Foreman 3.3 (https://projects.theforeman.org/issues/34643) we disable the statically generated files so I wonder if this really is still needed.

PR #10261, opened 2022-09-02
Fixes #35482 - Add ACS name and label validation

sjha4:
Add unique validation to name/label of ACSs.
Create an ACS.
Create another ACS with same name.
You should see a validation error instead of the db error.


chris1984 (approved):
ACK, Can we add a test for this @sjha4

sjha4:
Yes..let me add a validation test..

PR #10262, opened 2022-09-02
Fixes #35484 - Unable to apply all Errata via Remote Execution on Web UI with "Select All"

ianballou:
Fixes `BulkItemsHelper.fetch()` so that empty includes/excludes correctly return all items unfiltered.  The code accepted `[]` as true, which caused the wrong `if` statements to execute.

The new test should help ensure this doesn't break again.

1) Register a content host that has some errata.
2) Apply errata with REX by using the "Select all #" option in both the old content host details page AND the "Manage Errata" modal from the content host index page.
3) Try applying all errata again, this time ensuring that un-checked errata are not included in the REX job.
4) Try applying single errata without using "Select all #".
5) Ensure that applying all errata from the new host details page works as well.


jeremylenz (approved):
LGTM! Thanks @ianballou

and thanks Rails for giving us `.present?` and making Ruby behave how we think it should :smile:

jeremylenz:
Katello test failures unrelated

React lint failures unrelated, looks like we must've updated eslint rules or something..

ianballou:
[test katello]

ianballou:
[test katello]

PR #10265, opened 2022-09-07
Fixes #35501 - Hides generated component cvvs in composite

parthaa:

When we ask for a list of component cvv's available for a composite, the backend api also returns component cvvs that are from export/import generated content views (like export-library/export-repository) etc. These versions are not meant to be included in the composite.

This commit changes this behavior by filtering out generated content views all together.

Also adds validation to not add generated content view versions to a composite

1. Export a repository or library
2. Create a Composite CV
3. Go to 'Content Views' tab of composite cv details page

Before PR
You'd notice the `Export-*` content views

After PR
No `Export-*` cvs

Additional ideas
0. Checkout PR
1. Export a content view version.
2. Go to 'Content Views' tab of composite cv details page
3. Make sure the exported CVV is still there on that page.


chris1984 (approved):
ACK, works well. Thanks @parthaa looks like there is a rubocop failure

PR #10266, opened 2022-09-07
Fixes #35503 - Calculate errata_mail errata list via updated_at per erratum

wbclark:
Previously this was done per the updated_at attribute on katello_repository_errata

CC: @pmoravec @hao-yu @sjha4


pmoravec:
I *think* that is the idea of fix I had some time ago, to match updated_at against errata record and not repository_errata record, so +1 for that. (I dont have time to try it to be more confident).

hao-yu:
I think we discussed about only report new errata that are added to the repo.

This scratch PR should do it: https://github.com/Katello/katello/pull/10285

I only did a rough test against the PR so feel free to test it to see if it will work.

wbclark:
Closing in favor of newer attempts at the issue.

PR #10269, opened 2022-09-09
Fixes #35511 - Navigating to content view page from the left panel after creating a cv does not work

sjha4:
Go to Content > Content Views
Create a new CV.
You should land on repositories tab of new CV. ( or content view tab of new Composite Content View.)
Try going to Content > Content Views from navigation.


jeremylenz (approved):
Spent some time trying to find a clean React Router way to accomplish this, but alas..

ACK for now, to unblock us, until I am further educated.

PR #10270, opened 2022-09-09
Fixes #35512 - Don't trigger autosearch until you've typed an operator

jeremylenz:

Description of problem:
In any content search bar, the initial click shows a dropdown of filter keys. These are literal search terms and not key designators.

How reproducible:
100%

Steps to Reproduce:
1. Register host. 
2. Click on host > Content > Packages. 
3. Click on search bar. 

Actual results:
Search is launched for the Key and not the Value of the filter.

Expected results:
Search is launched for the Value of the filter.

#### Solution

This change prevents the autosearch from triggering until the user has typed
1. an operator - `['>=', '<=', '>', '<', '~', '^', '!^', '=', '!=', '~', '!~']`
2. at least part of a value.

This will eliminate most of the premature searches.

The only downside is that if you're searching _not_ using scoped_search syntax (which by the way you can do and scoped_search supports!), you'll now have to press the Enter key to complete the search.

Go to any content search (new host details or CV UI)
Click in the search bar and select a field, then select the operator (for example `errata_id = `)
After applying this change, notice that searches are not executed immediately
Type a value, and notice that the autosearch is now triggered

### DIFF
--- a/webpack/components/Search/Search.js
+++ b/webpack/components/Search/Search.js
@@ -12,6 +12,25 @@ import {
   AUTOSEARCH_WHILE_TYPING,
 } from '../../scenes/Settings/SettingsConstants';
 
+// Don't trigger auto-search until you've typed an operator and begun to type a value
+const searchContainsOperator = (search = '') => {
+  const operators = ['>=', '<=', '>', '<', '~', '^', '!^', '=', '!=', '~', '!~'];
+  return operators.some(operator => search.includes(operator) && !search.trim().endsWith(operator));
+};
+// Apply the above function only to the last 'key = value' pair, if there are multiple
+const allowSearch = (originalSearch = '') => {
+  const search = originalSearch.toLowerCase().trim();
+  // find the last occurrence of 'or' or 'and'
+  const lastOr = originalSearch.lastIndexOf('or');
+  const lastAnd = originalSearch.lastIndexOf('and');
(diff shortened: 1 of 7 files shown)


webpack/components/Search/Search.js
```diff
@@ -12,6 +12,12 @@ import {
   AUTOSEARCH_WHILE_TYPING,
 } from '../../scenes/Settings/SettingsConstants';
 
+// Don't trigger auto-search until you've typed an operator and begun to type a value
+const searchContainsOperator = (search = '') => {
+  const operators = ['>=', '<=', '>', '<', '~', '^', '!^', '=', '!=', '~', '!~'];
```
sjha4, line 17: Does it make sense to add a check that search doesn't end with 'and' or 'or'?

jeremylenz: I tried this, but by the time 'and' or 'or' is added to a search it already contains an operator, so it doesn't help anything..

sjha4: We'd probably want to split on 'and' and 'or' and apply the same operator check on all different parts of search or maybe just the last part after the and/or.

jeremylenz:
Updated addressing @sjha4 comments :)

sjha4:
Works great. Test failures seem related.

sjha4 (approved):
APJ 👍🏼

jeremylenz:
[test katello]

PR #10273, opened 2022-09-12
Fixes #35458 - Add environments param to candlepin update call

sjha4:
The 
so it used to be that if you wanted to update the consumer's enviornment you would send

{ "environment": "newEnv" }

but now because you can specify multiple environments
you have to send

{ "environments" : ["name" : "newEnv" ] }
or

{ "environments" : ["id" : "newEnvId" ] }

1. Create and assign a host to a CV/Env.
2. On the hosts UI update the CV/ENV on the host.
3. On the host run sub-man refresh
4. Make sure the enabled repos  are updated in /etc/yum.repos.d/redhat.repo

### DIFF
--- a/app/lib/katello/resources/candlepin/consumer.rb
+++ b/app/lib/katello/resources/candlepin/consumer.rb
@@ -69,6 +69,9 @@ module Katello
             if params.empty?
               true
             else
+              if params.key?(:environment) && params[:environment].key?(:id)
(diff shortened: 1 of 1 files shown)


chris1984 (approved):
ACK, works well.

sjha4:
[test katello]

linuxphone:
This needs to be committed very urgent to fix the lifecycle environment refresh issue, please push this as soon as possible. Thank you.

PR #10274, opened 2022-09-12
Fixes #35517 - Redirect to library repo from CV version and hide actions on non-library repos

sjha4:
1. CV version details repo links should redirect to library repo.
2. We shouldn't be able to perform normal repo actions on non-library repos at least from the UI.
a) 
1. Create a repo, add it to CV and publish a new version.
2. Go to version > Repositories> repo.
3. Check the IDs on the 2: CV version > Repositories> repo and CV > Repositories. These should be the same with this PR.
b)
1. Right click on UI and check the networks tab when landing on CV > versions> repositories sub-tab.
2. For the API call : https://centos8-katello-devel-stable.example.com/katello/api/v2/repositories?content_view_version_id=.....
From the results, pick id of one of the results and go to /products/1/repositories/:id.
3. Actions should be hidden on the page.
4. For regular library repos, you should still see actions on the page.

### DIFF
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/repository-details.controller.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/repository-details.controller.js
@@ -84,6 +84,13 @@
             return result;
         };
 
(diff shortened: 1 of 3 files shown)


chris1984 (approved):
ACK, looks good and works well.

PR #10280, opened 2022-09-19
Fixes #35539 - update recommended repos

wbclark:

1. Recommended repositories are updated to include rhel 6 els repository
2. Recommended repositories are updated for the Satellite 6.12 release

1. Satellite 6.12 is el8 only
2. Some Capsule 6.11 (including el7) repositories are preserved under recommended repositories, due to Capsule N-1 support
3. Satellite 6.11 repository is removed from recommended repositories

1. Upload a subscription manifest containing Satellite Infrastructure and RHEL6 Extended Lifecycle Support SKUs
2. Go to Red Hat Repositories page and check that the repositories are visible under recommended repositories

### DIFF
--- a/webpack/redux/actions/RedHatRepositories/helpers.js
+++ b/webpack/redux/actions/RedHatRepositories/helpers.js
@@ -33,15 +33,10 @@ const recommendedRepositoriesSatTools = [
 ];
 
 const recommendedRepositoriesMisc = [
-  'rhel-server-rhscl-7-rpms',
-  'rhel-7-server-satellite-capsule-6.11-rpms',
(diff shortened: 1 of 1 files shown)


wbclark:
After a brief discussion, we settled on removing the Capsule 6.11 repositories, which follows the precedent. This makes some sense based on the assumption that the primary use case for Capsule N-1 support is upgrading Satellite without immediately upgrading Capsules, in which case the Capsule 6.11 repositories would have already been enabled.

ianballou (changes requested):
The second commit references the wrong issue

ianballou (approved):

PR #10281, opened 2022-09-20
Refs #35546 - don't translate archRestricted

jeremylenz:

archRestricted and osRestricted values were being translated when they shouldn't be. This was causing the Repository Sets page to error when a repo has an undefined value for arch restriction.

No bugs were filed about OS restriction, but I removed that incorrect translation as well.

To simulate the issue, make this change:
```diff
diff --git a/app/views/katello/api/v2/repository_sets/show.json.rabl b/app/views/katello/api/v2/repository_sets/show.json.rabl
index 396702c898..8a333a53b5 100644
--- a/app/views/katello/api/v2/repository_sets/show.json.rabl
+++ b/app/views/katello/api/v2/repository_sets/show.json.rabl
@@ -38,7 +38,7 @@ child @resource.repositories => :repositories do
 end
 
 node :archRestricted do |pc|
-  pc.repositories&.first&.arch
+  nil
 end
```

On master, you'll get a "No translation key found" error. With this PR, the error should be eliminated.


sjha4 (approved):
Ack 👍🏼

PR #10282, opened 2022-09-21
Fixes #35549 - Delete repo even when included in composite content view

sjha4:
Allows deleting repos even when included in composite content views.
1. Create and sync a repo
2. Add repo to a component content view and publish
3. Create a composite content view and add the component content view from above with latest version.
4. Publish the composite content view.
5. Verify you see the repository in version details of the newly published version.
6. Go to repository page and delete the repository.
7. You should see the composite view in the confirmation modal.
8. Delete. Task should succeed.

### DIFF
--- a/app/lib/actions/katello/repository/destroy.rb
+++ b/app/lib/actions/katello/repository/destroy.rb
@@ -23,7 +23,7 @@ module Actions
                            repository,
                            SmartProxy.pulp_primary)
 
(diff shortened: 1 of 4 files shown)


chris1984 (approved):
ACK, works well and code looks good. Was able to delete the repo from both the ccv/cv without a failed task

PR #10284, opened 2022-09-21
Fixes #35504 - Simplified ACSs are being created during content view publishing

ianballou:
Stops content view publishes from creating more alternate content sources in Pulp.

A simplified ACS should have Pulp ACSs created when its products receive new repositories. However, new ACSs shouldn't be created at CV publish time. ACSs have essentially nothing to do with CVs.

CV situations should always use `clone`, which I believe is a safe assumption.

1) Create a product
2) Add a repo to that product
3) Create a simplified ACS with that product
4) Add another repo and ensure that more ACSs are created (check the repo create Dynflow)
5) Publish a CV with any of the product's repositories and see that more ACSs are NOT create (check the publish task Dynflow)

### DIFF
--- a/app/lib/actions/katello/repository/create.rb
+++ b/app/lib/actions/katello/repository/create.rb
@@ -34,7 +34,7 @@ module Actions
 
             concurrence do
               plan_self(:repository_id => repository.id, :clone => clone)
-              if repository.url.present?
+              if !clone && repository.url.present?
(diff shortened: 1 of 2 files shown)


sjha4 (approved):
Works well. ACS no longer created for clone repos for publish/promote and deletion/removing from env does not trigger the ACS deletion task either.
Ack 👍🏼

PR #10271, opened 2022-09-12
Fixes #35516 - Reconfigure yggdrasil on content source change

adamruzicka:
if present

PR #10286, opened 2022-09-22
Fixes #35548 - Change content source and registered_through proxy

stejskalleos:
Change content source should change `registered_through` proxy as well

See steps to reproduce in https://bugzilla.redhat.com/show_bug.cgi?id=2123975

See steps to reproduce in https://bugzilla.redhat.com/show_bug.cgi?id=2123975

### DIFF
--- a/app/helpers/katello/content_source_helper.rb
+++ b/app/helpers/katello/content_source_helper.rb
@@ -37,6 +37,8 @@ module Katello
           --server.prefix="#{content_source.rhsm_url.path}" \
           --rhsm.repo_ca_cert="$KATELLO_SERVER_CA_CERT" \
           --rhsm.baseurl="#{content_source.pulp_content_url}"
+
+        subscription-manager facts --update
       CMD
     end
   end


app/controllers/katello/api/v2/hosts_bulk_actions_controller.rb
```diff
@@ -328,6 +328,10 @@ def change_content_source
         host.content_facet.content_view = content_view
         host.content_facet.content_source = content_source
 
+        if Katello.with_remote_execution?
+          host.subscription_facet.registered_through =  URI.parse(content_source.url).host
```
jeremylenz, line 332:
```suggestion
          host.subscription_facet.registered_through = URI.parse(content_source.url).host
```

jeremylenz (commented):
Thanks @stejskalleos, tested and works perfectly.

Would also like to see the `test_change_content_source` updated with an assertion describing this change.

ianballou:
@adamruzicka 's comment here (https://github.com/Katello/katello/pull/10271#issuecomment-1255971489) is suggesting that this PR will break the REX pull provider.  If yggdrasil isn't updated before the smart proxy gets updated, then the next request to reconfigure yggdrasil on the host will fail becase it'll be attempting to connect via the wrong smart proxy

ianballou:
I confirmed that this is going to be an issue for the pull provider. yggdrasil needs to be reconfigured before changing the registered-through smart proxy.  I wonder if we could change the registered-through proxy from the host with subscription-manager via a job...

ianballou (approved):
Worked for me with https://github.com/theforeman/foreman/pull/9435 !

jeremylenz:
@stejskalleos Are you making any more changes here then?

stejskalleos:
> Are you making any more changes here then?

Nope, it should be enough I hope, plus @ianballou confirmed that it works.

jeremylenz:
Thanks @stejskalleos!

PR #10289, opened 2022-09-26
Fixes #35565 - ACS UX fixes - Wizard

sjha4:
Several changes based on UX review for ACS UI
Note the changes highlighted with yellow. Those have been addressed with this PR.
Test and make sure everything behaves as per review doc in issue.

### DIFF
--- a/webpack/scenes/AlternateContentSources/Create/ACSCreateWizard.js
+++ b/webpack/scenes/AlternateContentSources/Create/ACSCreateWizard.js
@@ -16,6 +16,8 @@ import { getSmartProxies } from '../../SmartProxy/SmartProxyContentActions';
 import { CONTENT_CREDENTIAL_CERT_TYPE } from '../../ContentCredentials/ContentCredentialConstants';
 import { getProducts } from '../ACSActions';
 import ACSProducts from './Steps/ACSProducts';
+import { areSubPathsValid, isValidUrl } from '../helpers';
+
 
 const ACSCreateWizard = ({ show, setIsOpen }) => {
   const [acsType, setAcsType] = useState(null);
@@ -49,6 +51,16 @@ const ACSCreateWizard = ({ show, setIsOpen }) => {
     [dispatch],
   );
 
+  const credentialsFilled = () => {
+    if (authentication === 'manual') {
+      return username !== '';
+    }
+    return true;
+  };
+
+  const subPathValidated = areSubPathsValid(subpaths) ? 'default' : 'error';
+  const urlValidated = (url === '' || isValidUrl(url)) ? 'default' : 'error';
+
   const sourceTypeStep = {
     id: 1,
     name: __('Select source type'),
@@ -60,6 +72,7 @@ const ACSCreateWizard = ({ show, setIsOpen }) => {
     id: 2,
     name: __('Name source'),
     component: <NameACS />,
+    canJumpTo: acsType && contentType,
     enableNext: name !== '',
   };
 
@@ -67,6 +80,7 @@ const ACSCreateWizard = ({ show, setIsOpen }) => {
     id: 3,
     name: __('Select smart proxy'),
     component: <ACSSmartProxies />,
+    canJumpTo: name !== '',
     enableNext: smartProxies.length,
   };
 
@@ -74,6 +88,7 @@ const ACSCreateWizard = ({ show, setIsOpen }) => {
     id: 4,
     name: __('Select products'),
     component: <ACSProducts />,
+    canJumpTo: smartProxies.length,
     enableNext: productIds.length,
   };
 
@@ -81,14 +96,16 @@ const ACSCreateWizard = ({ show, setIsOpen }) => {
     id: 5,
     name: __('URL and paths'),
     component: <AcsUrlPaths />,
-    enableNext: url !== '',
+    canJumpTo: acsType === 'custom' && (smartProxies.length),
+    enableNext: url !== '' && urlValidated !== 'error' && subPathValidated !== 'error',
   };
 
   const credentialsStep = {
     id: 6,
     name: __('Credentials'),
     component: <ACSCredentials />,
-    enableNext: (url !== '' || productIds.length) && smartProxies.length && name !== '' && acsType && contentType,
+    canJumpTo: url !== '' && urlValidated !== 'error' && subPathValidated !== 'error',
+    enableNext: (url !== '' || productIds.length) && smartProxies.length && name !== '' && acsType && contentType && credentialsFilled(),
   };
 
   const reviewStep = {
@@ -96,6 +113,7 @@ const ACSCreateWizard = ({ show, setIsOpen }) => {
     name: __('Review details'),
     component: <ACSReview />,
     nextButtonText: __('Add'),
+    canJumpTo: (url !== '' || productIds.length) && smartProxies.length && name !== '' && acsType && contentType && credentialsFilled(),
     enableNext: (url !== '' || productIds.length) && smartProxies.length && name !== '' && acsType && contentType,
   };
 
--- a/webpack/scenes/AlternateContentSources/Create/Steps/ACSReview.js
+++ b/webpack/scenes/AlternateContentSources/Create/Steps/ACSReview.js
@@ -1,4 +1,5 @@
 import React, { useContext } from 'react';
+import { capitalize } from 'lodash';
 import { translate as __ } from 'foremanReact/common/I18n';
 import { TextContent, TextList, TextListItem, TextListItemVariants, TextListVariants } from '@patternfly/react-core';
 import ACSCreateContext from '../ACSCreateContext';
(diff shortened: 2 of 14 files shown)


webpack/scenes/AlternateContentSources/Create/ACSCreateWizard.js
```diff
@@ -38,6 +38,8 @@ const ACSCreateWizard = ({ show, setIsOpen }) => {
   const [productIds, setProductIds] = useState([]);
   const [productNames, setProductNames] = useState([]);
   const [currentStep, setCurrentStep] = useState(1);
+  const [urlValidated, setUrlValidated] = useState('default');
+  const [subPathValidated, setSubPathValidated] = useState('default');
```
jeremylenz, line 42: I know this may not be strictly in scope for this PR, but I would _really_ love to see these states converted to values computed at render instead. It just makes the code so much simpler. Highly recommend reading this https://beta.reactjs.org/learn/you-might-not-need-an-effect#how-to-remove-unnecessary-effects-how-to-remove-unnecessary-effects

webpack/scenes/AlternateContentSources/Create/Steps/ACSReview.js
```diff
@@ -16,9 +17,14 @@ const ACSReview = () => {
...
     <>
       <WizardHeader
-        title={__('Review Details')}
-        description={__('Review the information below and click Add to add your source. ' +
-                'To edit details in previous steps, click Back or any step on the left.')}
+        title={__('Review details')}
+        description={
+          <>
+            {__('Review the information below and click ')}<b>{__('Add')}</b>{__(' to add your source. ' +
+                'To edit details in previous steps, click ')}<b>{__('Back')}</b>{__(' or any step on the left.')
```
jeremylenz, line 24: The `<b>` tags should be `<strong>` instead; see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b

jeremylenz (commented):
Thanks @sjha4 

Some initial comments (I'll quote the requirements from the doc and then say if it meets them :)

> I want to see new ACS without being forced to refresh manually. Sometimes it works sometimes it doesn’t

It doesn't (for me.) The page doesn't refresh on create after I finish the wizard..

> Refresh the page and hide details when deleting expanded ACS

:heavy_check_mark: 

>  URL field should validate a correct format

I was able to enter `http://ee`; is this valid?

> Can we validate incorect format of paths on subpaths step?

:heavy_check_mark: 

> Steps in wizard can't be enabled if there are required fields elsewhere. 

:heavy_check_mark: 

> If “Manual authentication” is selected - “Next” should be disabled

:heavy_check_mark: 

> Can’t proceed in Simplified flow using “Next” - it’s disabled in Products

:heavy_check_mark: Button becomes enabled once I select a product.

sjha4:
> > I want to see new ACS without being forced to refresh manually. Sometimes it works sometimes it doesn’t
> 
> x It doesn't (for me.) The page doesn't refresh on create after I finish the wizard..

Sorry, moved it to part 2. Ignore this for the scope of this PR.

sjha4:
> I was able to enter `http://ee`; is this valid?

I didn't want to make too restrictive of validations around this. For ex: http://ee is valid if you update your /etc/hosts to point it to an IP.. There's also IP addresses etc which might be used. The only check I have here is that it contains a protocol:// 🐱

webpack/scenes/AlternateContentSources/Create/ACSCreateWizard.js
sjha4: I had these as variables inside the URL and subpath step in the 1st iteration on this. However, we need the urlValidated and subPathValidated values from the child step in the parent component, i.e the wizard to decide if next button should be enabled/disabled.

sjha4:
[test katello]

webpack/scenes/AlternateContentSources/Create/ACSCreateWizard.js
jeremylenz: If we move the validation functions to the parent component, we no longer need the states or the effect! Like this:

```diff
diff --git a/webpack/scenes/AlternateContentSources/Create/Steps/ACSReview.js b/webpack/scenes/AlternateContentSources/Create/Steps/ACSReview.js
index 90846aaad5..132fd577dc 100644
--- a/webpack/scenes/AlternateContentSources/Create/Steps/ACSReview.js
+++ b/webpack/scenes/AlternateContentSources/Create/Steps/ACSReview.js
@@ -20,7 +20,7 @@ const ACSReview = () => {
         title={__('Review details')}
         description={
           <>
-            {__('Review the information below and click ')}<b>{__('Add')}</b>{__(' to add your source. ' +
+            {__('Review the information below and click ')}<strong>{__('Add')}</strong>{__(' to add your source. ' +
                 'To edit details in previous steps, click ')}<b>{__('Back')}</b>{__(' or any step on the left.')
             }
           </>
diff --git a/webpack/scenes/AlternateContentSources/Create/Steps/AcsUrlPaths.js b/webpack/scenes/AlternateContentSources/Create/Steps/AcsUrlPaths.js
index 27fca58061..a6ca331d36 100644
--- a/webpack/scenes/AlternateContentSources/Create/Steps/AcsUrlPaths.js
+++ b/webpack/scenes/AlternateContentSources/Create/Steps/AcsUrlPaths.js
@@ -13,50 +13,9 @@ import WizardHeader from '../../../ContentViews/components/WizardHeader';
 const AcsUrlPaths = () => {
   const {
     url, setUrl, subpaths, setSubpaths, verifySSL, setVerifySSL,
-    urlValidated, setUrlValidated, subPathValidated, setSubPathValidated,
+    urlValidated, subPathValidated,
   } = useContext(ACSCreateContext);
 
-  const isValidUrl = (urlString) => {
-    try {
-      const urlFromString = new URL(urlString);
-      return (urlFromString.protocol === 'https:' || urlFromString.protocol === 'http:' || urlFromString.protocol === 'file:');
-    } catch (e) {
-      return false;
-    }
-  };
-
-  const areSubPathsValid = (subpathsString) => {
-    try {
-      if (subpathsString === '') {
-        return true;
-      }
-      return subpathsString.split(',').every((subpath) => {
-        const trimmedPath = subpath.trim();
-        return (!trimmedPath.startsWith('/') && trimmedPath.endsWith('/'));
-      });
-    } catch (e) {
-      return false;
-    }
-  };
-
-  const handleURLChange = (newUrl, _event) => {
-    setUrl(newUrl);
-    if (isValidUrl(newUrl)) {
-      setUrlValidated('default');
-    } else {
-      setUrlValidated('error');
-    }
-  };
-
-  const handleSubpathsChange = (newSubpaths, _event) => {
-    setSubpaths(newSubpaths);
-    if (areSubPathsValid(newSubpaths)) {
-      setSubPathValidated('default');
-    } else {
-      setSubPathValidated('error');
-    }
-  };
-
   return (
     <>
       <WizardHeader
@@ -81,7 +40,7 @@ const AcsUrlPaths = () => {
             placeholder="https:// or file://"
             value={url}
             validated={urlValidated}
-            onChange={value => handleURLChange(value)}
+            onChange={setUrl}
           />
         </FormGroup>
         <FormGroup
@@ -95,7 +54,7 @@ const AcsUrlPaths = () => {
             placeholder="test/repo1/, test/repo2/,"
             value={subpaths}
             validated={subPathValidated}
-            onChange={(value) => { handleSubpathsChange(value); }}
+            onChange={setSubpaths}
             name="acs_subpath_field"
             id="acs_subpath_field"
             aria-label="acs_subpath_field"
```

sjha4: I kind of prefer the validation logic for the step component to stay encapsulated in the component itself instead of the parent. Thoughts?

jeremylenz: Those validation logic functions are actually completely independent. So you could just put them outside the component altogether.

sjha4: That makes sense..Will move the 2 validators out and reuse them in the edit modal as well so we don't get another review from Maria.. :)

webpack/scenes/AlternateContentSources/ACSConstants.js
```diff
@@ -18,4 +18,27 @@ export const ACS_TYPE_TRANSLATIONS_ENUM = {
   [FILE]: 'file',
 };
 
+export const isValidUrl = (urlString) => {
```
jeremylenz, line 21: okay, technically it's a `const` but I feel like these fit better in a `_helpers.js` file rather than a `Constants.js` file :smile:

jeremylenz (commented):
Validation code looks much better! :heavy_check_mark: 

Should some of the modal titles be fixed? For example I think

```
Edit Alternate content source URL and subpaths
```

should be

```
Edit alternate content source URL and subpaths
```

webpack/scenes/AlternateContentSources/ACSConstants.js
sjha4: Ya..I was thinking of a helpers file..Let me put these in there..

sjha4:
> Edit alternate content source URL and subpaths

Does it make sense to drop "alternate content source" altogether and just have `Edit URL and subpaths`

jeremylenz (approved):
Code changes LGTM! :+1:  Thanks @sjha4

PR #10291, opened 2022-09-27
[CP] 4.5.1 release

ianballou:


parthaa:
[test katello]

parthaa (approved):
APJ

ianballou:
It appears that the npm packages `expect` and `pretty-format` are too new and thus incompatible with node 12. We need to figure out how to properly pin back these versions.

package.json
```diff
@@ -27,6 +27,8 @@
     "@babel/eslint-plugin": "^7.14.5",
     "@sheerun/mutationobserver-shim": "^0.3.3",
     "@testing-library/jest-dom": "^5.3.0",
+    "expect": "^28.1.3",
+    "pretty-format": "^28.1.3",
```
sjha4, line 31: This shouldn't be needed.

ianballou: These packages are causing the `npm ci` part of the pipeline to fail since they don't support node 12.  If that won't cause issues with the actual release, we could merge with the broken pipeline. I just don't know enough to make that call

ianballou: Looks like even pinning them here isn't sufficient since remote execution has them too as dependencies

ianballou: (I see now that I'm pinning them wrong anyway)

ianballou:
Holding off on merging until we figure out https://community.theforeman.org/t/applicability-issue-with-katello-4-5-0/30650/7

PR #10295, opened 2022-09-28
Fixes #35575 - Always show toggle group when host is non-library

jeremylenz:

new host details UI - When a host has applicable errata but none of them are installable, the 'All / Installable' toggle group is incorrectly hidden. As a result, you are unable to navigate to 'All' to show the applicable errata.

This fixes that by always showing the toggle group unless the host is in the default CV and environment.

Used the pattern already existing for `alwaysShowActionButtons` and added a new prop `alwaysShowToggleGroup`. The default is false so this shouldn't affect anything besides ErrataTab.

Get a host with applicable & installable errata, then change that host to a CV/E that doesn't have those errata.
Navigate to Content > Errata.

_Before:_ You'll see an error that the host has errata that are applicable, but not installable. (which is true) But you can't do anything about it.
_After:_ You'll see the same error, but the toggle group will show and you can click 'All'.

### DIFF
--- a/webpack/components/Table/TableWrapper.js
+++ b/webpack/components/Table/TableWrapper.js
@@ -19,6 +19,7 @@ import { orgId } from '../../services/api';
 const TableWrapper = ({
   actionButtons,
   alwaysShowActionButtons,
+  alwaysShowToggleGroup,
   toggleGroup,
   children,
   metadata,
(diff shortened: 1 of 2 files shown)


lfu (approved):
LGTM 👍🏻

PR #9873, opened 2022-01-03
Fixes #34211 - Hosts - Change content source

stejskalleos:
#### What are the changes introduced in this pull request?
New feature for changing host's content source & generating `bash` code for configuration on the host machines

#### Considerations taken when implementing this change?
* This PR is follow up to https://github.com/Katello/katello/pull/9756

#### What are the testing steps for this pull request?
**UI:**
* Go to Hosts > All hosts, select hosts and choose _Change Content Source_ option in menu
* Change the CS
* Run job invocation (or do it manually)
* Verify that subscription-manager have been configured properly

**API:**
* Call the API endpoint from host machine and run bash code on it
```
curl -X 'PUT 'http://katello.example.com/api/hosts/change_content_source' \
  --user admin:changeme \
  -H 'Content-Type: application/json' \
  --data '{"environment_id":"0","content_view_id":"0","content_source_id":"0","host_ids":[0]}' | bash
```

#### TODO
* React tests

PR #10297, opened 2022-10-03
Fixes #35446 - Change content source improvements

stejskalleos:
Follow up to https://github.com/Katello/katello/pull/9873

Bug fixes:
* Select all hosts from host index (not ignoring `search` parameter anymore)
* Load proxies with content AND environment

Improvements
* Select proxy first, avoid mismatch of environment, content view and capsule
* Lifecycle environment in '/katello/api/capsule'
* Better UI for list of hosts with search input

* Discussion and notes from the CEE demo

* Select one or more hosts (with content & without content)
* Go to the form
* Verify that displaying hosts works (to update & ignored)
* Check loaded data in the form - now it shouldn't be possible to select incorrect proxy / environment / content view
* Change the content source & verify that hosts consume content from new endpoint


stejskalleos:
Hi @jturel @jeremylenz, you guys did reviews for the https://github.com/Katello/katello/pull/9873, could you please look at this one too?

jeremylenz:
@stejskalleos thanks! Added to our review board; should see a review within the next few days.

webpack/scenes/Hosts/ChangeContentSource/components/Hosts.js
```diff
@@ -0,0 +1,83 @@
...
+          value={search}
+          onChange={v => setSearch(v)}
+          onClear={() => setSearch('')}
+        />
+        <List isPlain isBordered className="margin-top-16">
+          {(search ? modalHosts.filter(h => (`${h}`).includes(search)) : modalHosts).map(h => (
+            <ListItem key={h}>
+              <Button
+                component="a"
+                href={foremanUrl(`/hosts/${h}`)}
```
jeremylenz, line 63: This links to the legacy host detail page, even though my "New host details UI" setting is turned on.

webpack/scenes/Hosts/ChangeContentSource/components/Hosts.js
```diff
@@ -0,0 +1,83 @@
...
+
+const Hosts = ({
+  contentHosts, hostsWithoutContent,
+}) => {
+  const [modal, handleModal] = useState(false);
+  const [modalHosts, setModalHosts] = useState([]);
+  const [modalTitle, setModalTitle] = useState('Hosts');
+  const [search, setSearch] = useState('');
+
+  const loadModal = (items, title) => {
```
jeremylenz, line 25: It's generally not a good idea to store values in state that can be computed or derived from other data. It looks like we're using the `loadModal` function below to decide if this should be a "Hosts to update" or "Ignored hosts" modal. Instead, I think the code will be much cleaner if you move the `contentHosts` and `hostsWithoutContent` logic to the parent component.  There, you can render a `<Hosts>` component (or two if necessary) and decide _there_ what props to pass to it. This will be much cleaner than using the local state here.

webpack/scenes/Hosts/ChangeContentSource/helpers.js
```diff
@@ -15,6 +15,11 @@ export const getHostIds = () => {
   return [];
 };
 
+export const getSearchParam = () => {
+  const url = new URL(window.location);
+  return (url.searchParams.get('search'));
+};
```
jeremylenz, line 21: You may consider reusing `useUrlParams` from [here](https://github.com/Katello/katello/blob/6b36dea1b9be7208c89a468703f7f1d9c56045c5/webpack/components/Table/TableHooks.js#L259); it has some URL decoding and other niceties that you might find useful :)

jeremylenz (commented):
Thanks @stejskalleos!

Tested and works well. Left some comments about the React code below

stejskalleos:
@jeremylenz Rebased & updated, sorry for delay, I've been busy with `3.5`

jeremylenz (commented):
Changes look good, thanks! Just a couple nits below

stejskalleos:
@jeremylenz rebased & updated, it's :green_apple: now

jeremylenz (approved):
Thanks @stejskalleos 

LGTM! :+1:

PR #10300, opened 2022-10-03
Fixes #34897 - select ACS HTTP proxies per smart proxy

ianballou:

HTTP proxies for alternate content sources are now per-smart-proxy.  The idea is that users will likely want ACSs on each smart proxy to have a different HTTP proxy. At the moment, all ACSs in Pulp that are part of a Katello ACS will share the same HTTP proxy, regardless of the assigned smart proxies. It probably wouldn't make sense for a content proxy in Asia to use the same HTTP proxy as a content proxy in North America.

So, the user can select an "ACS HTTP Proxy" for each smart proxy.  This is done on the smart proxy edit page.

The user can also now select if they want their ACS to use HTTP proxies at all.

ACS remote configuration will be updated at refresh time to make managing remote update errors better. The smart proxy page doesn't handle async actions very well, so updating the remote at smart proxy configuration time could lead to a missed error.  On the flip side, if the error happens during refresh, the user is more likely to see it.

1) Should simplified ACSs inherit the HTTP proxies from the parent repositories?
  - I think it would be a good future RFE, but I'm not certain that users will find it necessary.  I feel like it would be unlikely that users would want different HTTP proxies for each ACS.  Typically they would use one global HTTP proxy, which this PR supports.
2) Should a test remote be created at smart proxy update time?
  - I'm thinking that creating test ACS remotes is not worth it for HTTP proxies because
    we're editing 'n' remotes rather than just one. At sync time, if a remote fails to
    update, the user can look at the error in Dynflow and likely figure out that something
    is wrong with the HTTP proxy config.

1) Register a smart proxy
2) Create an HTTP proxy (just install `squid` and then use  `http://fqdn:3128`) on the Katello server and the smart proxy.
3) Set the HTTP proxies on both smart proxies  according to (2)
4) Create custom and simplified ACSs
  - Toggle on "Use HTTP Proxies"
5) `tail -f /var/log/squid/access.log` on both the server and the smart proxy
6) Refresh the ACSs and watch the squid logs for requests
7) Create an immediate repo that uses the same RPMs as the ACSs
8) Sync the repo and watch the squid logs for requests
9) Toggle "Use HTTP Proxies" on and off and make sure the Pulp remotes do indeed get updated
10) **Take a look at the overall ACS HTTP Proxy workflow. Does it blend in well with existing HTTP proxy workflow?**

#### Questions:
1) The "Use HTTP Proxies" display on the ACS details is putting the boolean on a new line. I have yet to figure out why. The JSX seems identical for the similar "Verify SSL" on the same page.

#### TODO

- [x] Write tests


sjha4:
> The "Use HTTP Proxies" display on the ACS details is putting the boolean on a new line. I have yet to figure out why. The JSX seems identical for the similar "Verify SSL" on the same page.

Digging into this. The UI does look a little off the way it set up currently.

> Create an on-demand repo that uses the same RPMs as the ACSs
> Sync the repo and watch the squid logs for requests
Not sure I am seeing this happening. I duplicated a repo that is part of a simplified ACS product and synced the duplicate repo and I would expect it to log requests to the HTTP proxy.

webpack/scenes/AlternateContentSources/Details/ACSExpandableDetails.js
```diff
@@ -200,6 +201,17 @@ const ACSExpandableDetails = () => {
               <InactiveText text="N/A" />
                             }
             </List>
+            <TextContent className="margin-0-24">
+              <TextListItem component={TextListItemVariants.dt}>
```
sjha4, line 205: I have a PR up: https://github.com/Katello/katello/pull/10293 to add proper styling to some of these expandable sections so I am not worried about how it looks here. I will rebase and get that sorted as part of my PR as that also introduces some CSS classes not available in master currently. Wrapping the TextListItem inside TextList will resolve the newline issue.
```suggestion
          <TextList component={TextListVariants.dl}> 
              <TextListItem component={TextListItemVariants.dt}>
```

ianballou:
> Not sure I am seeing this happening. I duplicated a repo that is part of a simplified ACS product and synced the duplicate repo and I would expect it to log requests to the HTTP proxy.

I meant to say "immediate" there.  Do you see things logged if you download RPMs?   Also, you may have cached those RPMs already in Pulp, which might also explain a lack of downloads across the proxy.

If it's still being weird, check the Pulp ACS remote and see if the HTTP proxy did indeed make it to the remote.  Keep in mind that you need to refresh the ACS again to see the remote get updated.

webpack/scenes/AlternateContentSources/Details/ACSExpandableDetails.js
ianballou: Committed!

sjha4 (approved):
Looks and works well on pulpcore 3.18 for me. I ran into issues with 3.21 but we are not on 3.21 yet so let's merge this and revisit if we see any issues then. Ack. 👍🏼

PR #10304, opened 2022-10-05
Fixes #35596 - Repository sets shows reposets for sca

parthaa:
Prior to this PR if the subscription status of a host is cleared the repository sets page for the host does not show repository sets even if sca is set to true
This PR addresses that by tracking the sca status from the org instead of subscription status of the host.

Reused the org selector used by system purpose to track sca

1. Ensure Org has simple content access
1. Create a Custom repository
1. Register a host
1. Go to host => manage statuses
1. Remove the subscription status
1. Now go to the repository sets page

Before PR:
No repo sets shown

After PR:
All repo sets visible

### DIFF
--- a/webpack/components/extensions/HostDetails/Tabs/RepositorySetsTab/RepositorySetsTab.js
+++ b/webpack/components/extensions/HostDetails/Tabs/RepositorySetsTab/RepositorySetsTab.js
@@ -55,7 +55,9 @@ import {
   setContentOverrides,
 } from './RepositorySetsActions';
 
-import { selectOrganization } from '../../Cards/SystemPurposeCard/SystemPurposeSelectors';
+import { selectOrganization, selectOrganizationStatus } from '../../Cards/SystemPurposeCard/SystemPurposeSelectors';
+import { getOrganization } from '../../Cards/SystemPurposeCard/SystemPurposeActions';
+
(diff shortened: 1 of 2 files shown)


parthaa:
Added https://projects.theforeman.org/issues/35603 to DRY up systempurpose selectors . Moved it to a separate issue since its a refactor.

webpack/components/extensions/HostDetails/Tabs/RepositorySetsTab/RepositorySetsTab.js
```diff
@@ -247,6 +257,12 @@ const RepositorySetsTab = () => {
       simpleContentAccess, apiSortParams, statusSelected, STATUS_LABEL],
   );
 
+  useEffect(() => {
+    if (orgId && orgStatus !== STATUS.RESOLVED) {
```
jeremylenz, line 261: I think this logic probably needs to be different in order to guarantee the `getOrganization` call is made when needed.

parthaa: noted. Going to address it here -> https://projects.theforeman.org/issues/35604 do you think i should remove the useffects and get organization bits?

jeremylenz (approved):
Thanks @parthaa!

I'm okay with the smaller issue being in a separate redmine as noted.
ACK so we can get this fix in ASAP :+1:

parthaa:
[test katello]

PR #10305, opened 2022-10-05
Fixes #35595 - Add content profile to selectable columns

ofedoren:

Coulmns from `Content Hosts` page are added as selectable columns to `Hosts` page.

Basically it allows to proceed with merging `Content Hosts` page into `Hosts` page.

Currently, the only way to see the columns is to use `hammer user table-preference` command to create a preference for `hosts` table with certain columns. Core's and new `Content` profiles can be used. Until https://github.com/theforeman/foreman/pull/9323 is done, there is no other way to select the whole profile.

You can also use https://github.com/theforeman/foreman/pull/9323 to test the feature out as well as provide some feedback while it's still in early stage.

### DIFF
--- a/lib/katello/plugin.rb
+++ b/lib/katello/plugin.rb
@@ -265,6 +265,29 @@ Foreman::Plugin.register :katello do
       :onlyif => proc { |proxy| proxy.pulp_primary? }
   end
 
+  extend_page 'hosts/_list' do |context|
+    context.with_profile :content, _('Content'), default: false do
+      common_th_class = 'hidden-tablet hidden-xs'
+      common_td_class = "#{common_th_class} ellipsis"
+      use_pagelet :hosts_table_column_header, :name
+      use_pagelet :hosts_table_column_content, :name
+      add_pagelet :hosts_table_column_header, key: :subscription_status, label: _('Subscription status'), sortable: true, class: common_th_class, width: '10%'
+      add_pagelet :hosts_table_column_content, key: :subscription_status, class: common_td_class, callback: ->(host) { host_status_icon(host.subscription_global_status) }
+      add_pagelet :hosts_table_column_header, key: :installable_updates, label: _('Installable updates'), class: common_th_class, width: '15%'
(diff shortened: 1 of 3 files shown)


lib/katello/plugin.rb
```diff
@@ -265,6 +265,29 @@
...
+  extend_page 'hosts/_list' do |context|
+    context.with_profile :content, _('Content'), default: false do
+      common_th_class = 'hidden-tablet hidden-xs'
+      common_td_class = "#{common_th_class} ellipsis"
+      use_pagelet :hosts_table_column_header, :name
+      use_pagelet :hosts_table_column_content, :name
+      add_pagelet :hosts_table_column_header, key: :subscription_status, label: _('Subscription status'), sortable: true, class: common_th_class, width: '10%'
+      add_pagelet :hosts_table_column_content, key: :subscription_status, class: common_td_class, callback: ->(host) { host_status_icon(host.subscription_global_status) }
+      add_pagelet :hosts_table_column_header, key: :status, label: _('Installable updates'), sortable: true, class: common_th_class, width: '15%'
+      add_pagelet :hosts_table_column_content, key: :status, class: common_td_class, callback: ->(host) { errata_counts(host) }
```
adamruzicka, line 277: I'm not 100% sure how this works, but isn't `status` a little bit vague?

Also I'm getting `the field 'status' in the order statement is not valid field for search` when trying to order by it

ofedoren:
Thanks, @adamruzicka, it was just a copy-paste typo :( this column can't be sortable.

Also, here is a screenshot to make the initial review easier:
ScreenShot-1666020370119

chris1984:

chris1984:
@ofedoren what was your hammer command for the table preference columns?

chris1984 (approved):
ACK, tested each column and it looks correct. @adamruzicka has all of your concerns been addressed?

adamruzicka (approved):
Yes, they were, let's get this in

PR #10306, opened 2022-10-06
Fixes #35600 - Fix repository repair API call

sjha4:
Updates repair to use new Repair object instead of RepositoryVersion.
1. Verify checksum on any repository.


ianballou (commented):
Works fine! Think we should add a small VCR test to run the repository repair?

ianballou (approved):
Looks good!

PR #10307, opened 2022-10-08
Fixes #35606 - Can syncably import from webserver

parthaa:
This commit adds some code changes to facilitate syncably importing from a webserver.
The idea here is that users instead of copying bits across multiple disconnected importing satellites, copy it to a webserver instead that is accessible to all the satellites. They then run the content-import providing the url for the `--path` attribute.

 `hammer content-import version --organization=import-org --path=https://<fqdn>/<syncable_dump>`

The main difference between existing syncable import and this is that, this PR facilitates pulling from a webserver instead of disk.
 
The code does the following:

1. Pulls the metadata json from the webserver
2. Auto creates products
3. Auto creates repositories with the feed url pointing to the http webserver
4. Syncs the repositories
5. Creates a CV if needed and adds the repos to this
6. Publishes the CV if one was created. 

This is all code that is already in main branch except for step 3
This only works for syncable exported content not regular

1. Check out this branch and https://github.com/Katello/hammer-cli-katello/pull/867
2. Create a few custom repos + redhat repos with download policy set to immediate
3. Sync those repos
4. Add them to a content view and then publish
5. Now export that version syncably ->
```
$ hammer content-export complete version --id=100 --format=syncable

[....................................................................................................................................................................] [100%]
Generated /var/lib/pulp/exports/<....>/metadata.json
$ cd  /var/lib/pulp/exports/<....>/
$ python3 -m http.server 5050
Serving HTTP on 0.0.0.0 port 5050 (http://0.0.0.0:5050/) ...
```
6. Go to `http://<dev box>:5050` on the browser. Make sure you see the `manifest.json`
7. Now create a new org 
8.  import a manifest in the new org
9. Set the cdn configuration to `Export Sync`
```
$ hammer content-import version --organization=new-org --path=http://<devel-box>:5050
[....................................................................................................................................................................] [100%]

```

You should see 
1. repos created/enabled with url pointing to your http server,
2. cv and cvv created 
3. Verify the content is the same.

### DIFF
--- a/app/lib/actions/katello/content_view/publish.rb
+++ b/app/lib/actions/katello/content_view/publish.rb
@@ -13,7 +13,7 @@ module Actions
           action_subject(content_view)
 
           content_view.check_ready_to_publish!(options.slice(:importing, :syncable))
-          unless options[:importing]
+          unless options[:importing] || options[:syncable]
             ::Katello::Util::CandlepinRepositoryChecker.check_repositories_for_publish!(content_view)
           end
 
--- a/app/services/katello/pulp3/content_view_version/import_validator.rb
+++ b/app/services/katello/pulp3/content_view_version/import_validator.rb
@@ -17,6 +17,7 @@ module Katello
             fail _("Content view not provided in the metadata")
           end
 
+          ensure_non_syncable_path_valid! unless @metadata_map.syncable_format?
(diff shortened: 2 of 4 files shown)


chris1984 (approved):
ACK, works well

PR #10310, opened 2022-10-10
Fixes #33463 - Add RHUI ACS create wizard

lfu:
Create a new ACS wizard for RHUI ACSs that is the same as Custom, but includes help text and copy/paste buttons specifically for RHUI.  RHUI ACS instructions are in this [KCS](https://access.redhat.com/articles/6972161#highlighted-use-case-1-sync-a-capsule-from-an-alternate-source-3).

Also, add descriptions for the different ACS types on their cards.

### DIFF
--- a/webpack/scenes/AlternateContentSources/Create/Steps/ACSCredentials.js
+++ b/webpack/scenes/AlternateContentSources/Create/Steps/ACSCredentials.js
@@ -18,7 +18,7 @@ import Loading from '../../../../components/Loading';
 
 const ACSCredentials = () => {
   const {
-    authentication, setAuthentication, username, setUsername, password, setPassword,
+    acsType, authentication, setAuthentication, username, setUsername, password, setPassword,
     sslCert, setSslCert, sslKey, setSslKey, caCert, setCACert,
     setSslCertName, setSslKeyName, setCACertName, verifySSL, setVerifySSL,
   } = useContext(ACSCreateContext);
@@ -32,13 +32,18 @@ const ACSCredentials = () => {
 
   const getCertName = id => contentCredentials?.filter(cc => Number(cc.id) === Number(id))[0]?.name;
 
+  const description = acsType === 'rhui' ?
+    __('Choose content credentials if required for this RHUI source.') :
+    __('Enter basic authentication information or choose content credentials if required for this source.');
+
   return (
     <>
       <WizardHeader
         title={__('Credentials')}
-        description={__('Enter basic authentication information or choose content credentials if required for this source.')}
+        description={description}
       />
       <Form>
+        {(acsType !== 'rhui') &&
         <Radio
           label={__('Manual authentication')}
           id="manual_auth"
@@ -53,6 +58,7 @@ const ACSCredentials = () => {
             setSslKeyName('');
           }}
         />
+        }
         {(authentication === 'manual') &&
           <>
             <FormGroup
--- a/webpack/scenes/AlternateContentSources/Create/Steps/AcsUrlPaths.js
+++ b/webpack/scenes/AlternateContentSources/Create/Steps/AcsUrlPaths.js
@@ -1,6 +1,7 @@
 import React, { useContext } from 'react';
 import { translate as __ } from 'foremanReact/common/I18n';
 import {
+  ClipboardCopy,
   Form,
   FormGroup,
   TextInput,
@@ -12,24 +13,43 @@ import { areSubPathsValid, isValidUrl } from '../../helpers';
 
 const AcsUrlPaths = () => {
   const {
-    url, setUrl, subpaths, setSubpaths,
+    acsType, url, setUrl, subpaths, setSubpaths,
   } = useContext(ACSCreateContext);
 
   const subPathValidated = areSubPathsValid(subpaths) ? 'default' : 'error';
-  const urlValidated = isValidUrl(url) ? 'default' : 'error';
+  const [urlValidated, setUrlValidated] = React.useState('default');
+  const handleUrlChange = (newUrl, _event) => {
+    setUrl(newUrl);
+    if (isValidUrl(newUrl, acsType)) {
+      setUrlValidated('success');
+    } else {
+      setUrlValidated('error');
+    }
+  };
+
+  const baseURLplaceholder = acsType === 'rhui' ?
+    'https://rhui-server.example.com/pulp/content' :
+    'http:// or https://';
+  const helperTextInvalid = acsType === 'rhui' ?
+    'http://rhui-server.example.com/pulp/content or https://rhui-server.example.com/pulp/content' :
+    'http://, https:// or file://';
+  let headerDescription =
+    __('Enter in the base path and any subpaths that should be searched for alternate content.');
+  headerDescription = acsType === 'rhui' ?
+    `${headerDescription}${__(' The base path must be a web address pointing to the root RHUI content directory.')}` :
+    `${headerDescription}${__(' The base path can be a web address or a filesystem location.')}`;
 
   return (
     <>
       <WizardHeader
         title={__('URL and paths')}
-        description={__('Enter in the base path and any subpaths that should be searched for alternate content. ' +
-          'The base path can be a web address or a filesystem location.')}
+        description={headerDescription}
       />
       <Form>
         <FormGroup
           label={__('Base URL')}
           fieldId="acs_base_url"
-          helperTextInvalid="http://, https:// or file://"
+          helperTextInvalid={helperTextInvalid}
           validated={urlValidated}
           isRequired
         >
@@ -39,12 +59,24 @@ const AcsUrlPaths = () => {
             id="acs_base_url_field"
             name="acs_base_url_field"
             aria-label="acs_base_url_field"
-            placeholder="https:// or file://"
+            placeholder={baseURLplaceholder}
             value={url}
-            validated={url !== '' && urlValidated}
-            onChange={value => setUrl(value)}
+            validated={urlValidated}
+            onChange={handleUrlChange}
           />
         </FormGroup>
+        {acsType === 'rhui' &&
+        <>
+          {__('On the RHUA Instance, check the available repositories.')}
+          <ClipboardCopy hoverTip="Copy" clickTip="Copied" variant="inline-compact" isBlock>
+            rhui-manager repo list
+          </ClipboardCopy>
+          {__('Find the relative path for each RHUI repository and combine them in a comma-separated list.')}
+          <ClipboardCopy hoverTip="Copy" clickTip="Copied" variant="inline-compact" isBlock>
+            rhui-manager repo info --repo_id your_repo_id
+          </ClipboardCopy>
+        </>
+        }
         <FormGroup
           label={__('Subpaths')}
           type="string"
--- a/webpack/scenes/AlternateContentSources/Create/Steps/SelectSource.js
+++ b/webpack/scenes/AlternateContentSources/Create/Steps/SelectSource.js
@@ -1,16 +1,32 @@
 import React, { useContext } from 'react';
 import { translate as __ } from 'foremanReact/common/I18n';
-import { Flex, FlexItem, Form, FormGroup, FormSelect, FormSelectOption, Tile } from '@patternfly/react-core';
+import {
+  Alert,
+  ClipboardCopy,
+  Grid,
+  GridItem,
+  Form,
+  FormGroup,
+  FormSelect,
+  FormSelectOption,
+  Tile,
+} from '@patternfly/react-core';
+import { FormattedMessage } from 'react-intl';
 import ACSCreateContext from '../ACSCreateContext';
 import WizardHeader from '../../../ContentViews/components/WizardHeader';
 
 const SelectSource = () => {
   const {
-    acsType, setAcsType, contentType, setContentType,
+    acsType, setAcsType, contentType, setContentType, setAuthentication,
   } = useContext(ACSCreateContext);
 
   const onSelect = (event) => {
     setAcsType(event.currentTarget.id);
+    if (event.currentTarget.id === 'rhui') {
+      setAuthentication('content_credentials');
+    } else {
+      setAuthentication('');
+    }
   };
   const onKeyDown = (event) => {
     if (event.key === ' ' || event.key === 'Enter') {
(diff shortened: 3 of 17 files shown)


webpack/scenes/AlternateContentSources/Create/Steps/AcsUrlPaths.js
```diff
@@ -71,6 +90,15 @@ const AcsUrlPaths = () => {
             onChange={checked => setVerifySSL(checked)}
           />
         </FormGroup>
+        {verifySSL && acsType === 'rhui' &&
+          <>
+            {__('Generate certificates for the chosen repositories')}
```
ianballou, line 95: I'd instead put this on the "Credentials" part of the wizard under the "Content credentials" bullet.

I think we could make the "Content credentials" bullet the default selection too. I don't know when someone would connect to rhui with basic auth.  In fact -- I'm pretty sure we can remove the basic auth section ("Manual authentication") entirely.

image

webpack/scenes/AlternateContentSources/Create/Steps/AcsUrlPaths.js
```diff
@@ -71,6 +90,15 @@ const AcsUrlPaths = () => {
...
+            <ClipboardCopy hoverTip="Copy" clickTip="Copied" variant="inline-compact" isBlock>
+              rhui-manager client cert --name rhui-acs-certs --days 365 --dir /root --repo_label rhui-repo-1,rhui-repo-2
+            </ClipboardCopy>
+            {__('Create content credentials in server with the created SSL certificate and key via Content -> Content Credentials')}
```
ianballou, line 99: I think "in the server" would be implied by Content -> Content Credentials
```suggestion
            {__('Create content credentials with the generated SSL certificate and key via Content -> Content Credentials.')}
```

webpack/scenes/AlternateContentSources/Create/Steps/AcsUrlPaths.js
```diff
@@ -26,11 +33,23 @@ const AcsUrlPaths = () => {
         description={__('Enter in the base path and any subpaths that should be searched for alternate content. ' +
           'The base path can be a web address or a filesystem location.')}
       />
+      {acsType === 'rhui' &&
+      <>
+        {__('On the RHUA Instance, check the available repositories')}
+        <ClipboardCopy hoverTip="Copy" clickTip="Copied" variant="inline-compact" isBlock>
+          rhui-manager repo list
+        </ClipboardCopy>
+        {__('Find the relative path for the repository, the value for subpath')}
```
ianballou, line 42:
```suggestion
        {__('Find the relative path for each RHUI repository and combine them in a comma-separated list.')}
```

webpack/scenes/AlternateContentSources/Create/Steps/AcsUrlPaths.js
```diff
@@ -26,11 +33,23 @@ const AcsUrlPaths = () => {
         description={__('Enter in the base path and any subpaths that should be searched for alternate content. ' +
           'The base path can be a web address or a filesystem location.')}
       />
+      {acsType === 'rhui' &&
+      <>
+        {__('On the RHUA Instance, check the available repositories')}
```
ianballou, line 38:
```suggestion
        {__('On the RHUA Instance, check the available repositories.')}
```

webpack/scenes/AlternateContentSources/Create/Steps/AcsUrlPaths.js
ianballou, line 95: We've been standardizing around adding periods, so I've added some suggestions.

```suggestion
            {__('Generate certificates for the chosen repositories.')}
```

ianballou (commented):
Some first observations:

1) I realize that the ACS type can still be chosen with RHUI.  Since File won't ever be used, we can force the user to use the Yum type
2) In the Base URL, file:// is allowed. We can take that out since rhui will always be downloaded over http/https

webpack/scenes/AlternateContentSources/Create/Steps/SelectSource.js
```diff
@@ -19,10 +24,10 @@ const SelectSource = () => {
     }
   };
 
-  const typeOptions = [
-    { value: 'yum', label: __('Yum') },
-    { value: 'file', label: __('File') },
-  ];
+  const typeOptions = [{ value: 'yum', label: __('Yum') }];
+  if (acsType !== 'rhui') {
```
ianballou, line 28: I think we could actually hide the content type selection entirely since Yum will be the only option.

webpack/scenes/AlternateContentSources/Create/Steps/ACSCredentials.js
```diff
@@ -37,7 +39,18 @@ const ACSCredentials = () => {
         title={__('Credentials')}
         description={__('Enter basic authentication information or choose content credentials if required for this source.')}
       />
+      {verifySSL && acsType === 'rhui' &&
```
ianballou, line 42: `verifySSL` doesn't need to be true here. That's for peer validation rather than authentication with the server.  (In fact, the original Pulp KCS says to keep it false -- I'll need to double check that it's even okay for it to be true)

ianballou (commented):
Some of the header help text will need to change:

1) URL and paths:
Enter in the base path and any subpaths that should be searched for alternate content. The base path can be a web address or a filesystem location.

to

Enter in the base path and any subpaths that should be searched for alternate content. The base path must be a web address pointing to the root RHUI content directory.

and

Enter basic authentication information or choose content credentials if required for this source.

to

Choose content credentials if required for this RHUI source.

ianballou:
Also, for some reason, in the review details page, the source type is showing up as "Rhui". It looks a little funny being capitalized like a normal word:

image

webpack/scenes/AlternateContentSources/Create/Steps/ACSCredentials.js
lfu: > Also, for some reason, in the review details page, the source type is showing up as "Rhui"...

This was just implemented in https://github.com/Katello/katello/pull/10289.

lfu:
> Also, for some reason, in the review details page, the source type is showing up as "Rhui". It looks a little funny being capitalized like a normal word:

This was just implemented in https://github.com/Katello/katello/pull/10289.

ianballou:
This is looking great! Now we just need an ACK from UX.

ianballou:
Notes from UX meeting:

Content type: turn into a "disabled dropdown" (greyed out)

URL and paths
  - Step display on left is wrong for some reason

  - Red warning under Base URL shouldn't show up before the users types anything in
    - Need another description about the URL? 
      - As a popover or a subtext perhaps (Maria needs to check on this)
    - Error should happen only when they are done typing
      - After a pause and/or moving the cursor away
  - Put base URL above the new help text to improve the flow (since users will enter the Base URL and then Subpaths)

Content credentials
  - Put credentials information in the very first step with something that draws their attention
    - ...because once a user sees that they need to create content credentials, they've already wasted time on entering the previous information. The page will need to be reloaded to see the newly created content credentials.

Maria will help us get new suggestions for the ACS Type descriptions and a draft for the first page with the content credentials information (since we want it to really pop and look nice).

webpack/scenes/AlternateContentSources/Create/Steps/SelectSource.js
```diff
@@ -68,23 +99,34 @@ const SelectSource = () => {
...
+                  value={option.value}
+                  label={option.label}
+                />
+              ))
+            }
           </FormSelect>
         </FormGroup>
+        {acsType === 'rhui' &&
+        <>
+          <Alert variant="info" title={__('Generate certificates for the chosen repositories.')} />
```
ianballou, line 122:
```suggestion
          <Alert variant="info" title={__('Generate RHUI certificates for the desired repositories as necessary.')} />
```

ianballou (commented):
Looking good! Just a wording description for now

webpack/scenes/AlternateContentSources/Create/Steps/SelectSource.js
```diff
@@ -68,23 +99,34 @@ const SelectSource = () => {
...
           </FormSelect>
         </FormGroup>
+        {acsType === 'rhui' &&
+        <>
+          <Alert variant="info" title={__('Generate RHUI certificates for the desired repositories as necessary.')} />
+          <ClipboardCopy hoverTip="Copy" clickTip="Copied" variant="inline-compact" isBlock>
+            rhui-manager client cert --name rhui-acs-certs --days 365 --dir /root
+            --repo_label rhui-repo-1,rhui-repo-2
+          </ClipboardCopy>
+          {__('Create content credentials with the generated SSL certificate and key via Content -> Content Credentials.')}
```
ianballou, line 127: From our meeting, the main improvement suggested was adding a link here. What do you think about having the following text be an entire link?

"Create content credentials with the generated SSL certificated and key."

ianballou (commented):
We'll get some final word on what Verify SSL does and then see if any improvements can be done here to make it more clear.

ianballou:
So Verify SSL essentially checks that the server is who we think they are. The CA cert is required for that to work.  Samir's recent PR moves Verify SSL and the CA Cert together, which I think is enough.

webpack/scenes/AlternateContentSources/Create/Steps/SelectSource.js
lfu: Good idea.

ianballou:
I think this is looking great! We just need one more pass by UX and then we should be good to go.

ianballou:
I want to take another shot at the ACS type descriptions:

Custom: "Define any web or filesystem path as an alternate source for syncing content."

Simplified: "Sync smart proxy content directly from upstream repositories by simply selecting the desired products."

RHUI: "Guided steps for syncing smart proxy content from RHUI."

ianballou:
[test katello]

jeremylenz:
> I want to take another shot at the ACS type descriptions:
> 
> Custom: "Define any web or filesystem path as an alternate source for syncing content."
> 
> Simplified: "Sync smart proxy content directly from upstream repositories by simply selecting the desired products."
> 
> RHUI: "Guided steps for syncing smart proxy content from RHUI."

this is great! Maybe it would help if we break it up further: (a) what's an ACS; and (b) where does it come from. Like have a blurb at the top

Alternate Content Sources allow you to define an alternate source for syncing content in the event the primary server is not available.

Custom: "Define any web or filesystem path."

Simplified: "Sync smart proxy content directly from upstream repositories by simply selecting the desired products."

RHUI: "Guided steps for syncing smart proxy content from RHUI."

ianballou:
@jeremylenz I like the idea of having a blurb up top.  "...in the event the primary server is not available" interestingly isn't the main use case -- the alternate content source actually becomes the main content source.  It's more for importing content that is closer to you.  Like an override.  It's too late now, but I wonder if it should've been called override content source...

Anyway, here's my idea for the top blurb:  "Alternate content sources define new locations to download content from at repository or smart proxy sync time."

Then, Custom could become: "Define web or filesystem paths hold repositories."
  -> I'm making this more specific because refresh will fail if the path doesn't actually contain a repository that Pulp understands.

ianballou:
Also, I want to change RHUI because it's the only one that doesn't start with a verb:

"Sync smart proxy content from RHUI with guided steps"

ianballou:
So, altogether:

New top blurb:  "Alternate content sources define new locations to download content from at repository or smart proxy sync time."

Custom: "Define repositories structured under a common web or filesystem path."
Simplified: "Sync smart proxy content directly from upstream repositories by selecting the desired products."
RHUI: "Define RHUI repository paths with guided steps."

The simplified description feels different, but I can't think of a succinct way to describe it starting with "Define...".

webpack/scenes/AlternateContentSources/Create/Steps/SelectSource.js
```diff
@@ -38,25 +53,41 @@ const SelectSource = () => {
           isRequired
         >
           <Flex>
-            <FlexItem>
```
sjha4, line 41: This layout makes it look good.
<Grid hasGutter>
           <GridItem span={4} rowSpan={2}>

webpack/scenes/AlternateContentSources/Create/Steps/SelectSource.js
```diff
@@ -38,25 +53,41 @@ const SelectSource = () => {
...
               <Tile
                 title={__('Custom')}
                 isStacked
                 id="custom"
                 onClick={onSelect}
                 onKeyDown={onKeyDown}
                 isSelected={acsType === 'custom'}
-              />{' '}
+              >
+                Manually configure content source paths and authentication credentials
```
sjha4, line 65: Translations.

jeremylenz:
> the alternate content source actually becomes the main content source. It's more for importing content that is closer to you. Like an override.

Wow, TIL!

Maybe it's worth adding a sentence to the top blurb: "Content will be synced from the alternate content source first, then the original source if the ACS is not reachable."

ianballou:
I have one last request, can we completely capitalize RHUI on the index page?

image

ianballou (approved):
I'm cool with this! Thanks!

webpack/scenes/AlternateContentSources/Create/Steps/ACSCredentials.js
```diff
@@ -32,13 +32,18 @@ const ACSCredentials = () => {
 
   const getCertName = id => contentCredentials?.filter(cc => Number(cc.id) === Number(id))[0]?.name;
 
+  const description = acsType === 'rhui' ?
+    'Choose content credentials if required for this RHUI source.' :
```
sjha4, line 36: Can you add translations here instead of ```description={__(description)}``` below.

webpack/scenes/AlternateContentSources/Create/Steps/AcsUrlPaths.js
```diff
@@ -12,24 +13,43 @@ import { areSubPathsValid, isValidUrl } from '../../helpers';
...
+    `${headerDescription} The base path must be a web address pointing to the root RHUI content directory.` :
+    `${headerDescription} The base path can be a web address or a filesystem location.`;
 
   return (
     <>
       <WizardHeader
         title={__('URL and paths')}
-        description={__('Enter in the base path and any subpaths that should be searched for alternate content. ' +
-          'The base path can be a web address or a filesystem location.')}
+        description={__(headerDescription)}
```
sjha4, line 46: Same here for translations. These dynamic translations won't work. We can only translate static strings.

sjha4 (approved):
Awesome work @lfu . Ack 👍🏼

PR #10311, opened 2022-10-11
CVV Compare - Sorting compare tables

sjha4:
Adds sorting to CVV compare subtabs table
1. Select 2 CV versions and click compare.
2. On the compare screen > check the tabs and notice the sorting on pre-defined columns.
3. Click on the sorting enabled column and watch table refresh to sort data on that field.

### DIFF
--- a/webpack/scenes/ContentViews/Details/Versions/Compare/CVVersionCompareConfig.js
+++ b/webpack/scenes/ContentViews/Details/Versions/Compare/CVVersionCompareConfig.js
@@ -120,6 +120,10 @@ export default ({
         { title: __(`Version ${versionOne}`), getProperty: item => compareContent(item, versionOneId) },
         { title: __(`Version ${versionTwo}`), getProperty: item => compareContent(item, versionTwoId) },
       ],
+      sortConfig: {
+        [__('Name')]: 'name',
+        [__('Type')]: 'content_type',
+      },
     },
     {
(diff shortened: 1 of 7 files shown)


webpack/scenes/ContentViews/Details/Versions/Compare/CVVersionCompareConfig.js
```diff
@@ -297,6 +315,10 @@ export default ({
         { title: __(`Version ${versionOne}`), getProperty: item => compareContent(item, versionOneId) },
         { title: __(`Version ${versionTwo}`), getProperty: item => compareContent(item, versionTwoId) },
       ],
+      sortConfig: {
+        [__('Name')]: 'name',
+        [__('Architecture')]: 'architecture',
```
ianballou, line 320: Why can debs get Arch sorting but nothing else?

ianballou (commented):
I noticed a weird issue  with the search bar. If the search is cleared out manually by deleting all the text rather than using the "x" button, the search doesn't actually clear but the "x" is gone. To clear the search, you have to search for something else and then clear it with the "x".  Would that be more of a patternfly bug?

ianballou (approved):
The sorting looks good to me! The searching is still a little weird, I also noticed that `errata_id` and `title` aren't being autocompleted. That's unrelated to sorting though I think!

PR #10312, opened 2022-10-11
Fixes #35624 - Overwrite installation media

sbernhard:
The issue is because of two bugs:
1. The following JS method in app/assets/javascripts/katello//hosts/host_and_hostgroup_edit.js doesn't work as expected and therefore it does not clean Synced Content / All-Media selection if you switch between Media Selection:
```
KT.hosts.get_synced_content_dropdown` = function() {
    return $('select[data-kickstart-repository-id]');
};
```

The code does not work for new hosts, because kickstart_repo_id is empty and therefore, the `data-kickstart-repository-id` is not set => JQuery select can not work

2. Even if medium_id is not given, due to the inherited values from Hostgroup, the medium_id is set from the selected Hostgroup. This needs to be prevented if the kickstart_repository_id is given

### DIFF
--- a/app/models/katello/concerns/host_managed_extensions.rb
+++ b/app/models/katello/concerns/host_managed_extensions.rb
@@ -10,6 +10,12 @@ module Katello
           (content_source_id.blank? || (content_facet && content_facet.kickstart_repository.blank?)) && super
         end
 
+        def inherited_attributes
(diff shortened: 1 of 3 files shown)


chris1984 (approved):
Looks good to me

PR #10315, opened 2022-10-13
Fixes #35627 - Use proxy registration_url from Proxy in registration

stejskalleos:
* Use registration_url from proxy's `registration` module while registering hosts

[RFC: Host registration and Load balancers](https://community.theforeman.org/t/rfc-host-registration-and-load-balancers/30462)

* Configure proxy & haproxy
* Try to register hosts to the Foreman
* Verify that register host is configured to the haproxy & can consume data

#### Related PRs
* https://github.com/theforeman/foreman/pull/9464
* https://github.com/theforeman/smart-proxy/pull/850

### DIFF
--- a/app/controllers/katello/concerns/api/v2/registration_controller_extensions.rb
+++ b/app/controllers/katello/concerns/api/v2/registration_controller_extensions.rb
@@ -2,6 +2,7 @@ module Katello
   module Concerns
     module Api::V2::RegistrationControllerExtensions
       extend ActiveSupport::Concern
+      include ::Foreman::Controller::SmartProxyAuth
 
       def prepare_host
         if params['uuid']
@@ -35,13 +36,18 @@ module Katello
 
       def smart_proxy
         @smart_proxy ||= begin
-          proxy = params[:url] ? SmartProxy.unscoped.find_by(url: params[:url]) : SmartProxy.pulp_primary
+          proxy = params[:url] ? find_smart_proxy : SmartProxy.pulp_primary
 
           fail Foreman::Exception, _('Smart proxy content source not found!') unless proxy
           fail Foreman::Exception, _('Pulp 3 is not enabled on Smart proxy!') unless proxy.pulp3_enabled?
 
           proxy
         end
+      end
+
+      def find_smart_proxy
+        auth_smart_proxy
+        @detected_proxy
       end
     end
   end


app/controllers/katello/concerns/api/v2/registration_controller_extensions.rb
```diff
@@ -35,14 +35,28 @@ def context_urls
...
         end
       end
+
+      # Find proxy by URL configured in templates module
+      # or by proxy's URL
+      def find_smart_proxy_by_url
+        feature = Feature.find_by(name: 'Templates')
+
+        if feature
+          feature.smart_proxy_features
```
ekohl, line 53: Can't you query for `SmartProxyFeature` directly to save a DB query? Something like `SmartProxyFeature.where(feature: Feature.find_by(name: 'Templates'))`

app/controllers/katello/concerns/api/v2/registration_controller_extensions.rb
```diff
@@ -35,14 +35,28 @@ def context_urls
...
           proxy
         end
       end
+
+      # Find proxy by URL configured in templates module
+      # or by proxy's URL
+      def find_smart_proxy_by_url
+        feature = Feature.find_by(name: 'Templates')
+
+        if feature
```
ekohl, line 52: I think this logic doesn't work since the feature will always exist, even if no Smart Proxies using it exist. So it will never fall back to the `else` branch.

ekohl (changes requested):

app/controllers/katello/concerns/api/v2/registration_controller_extensions.rb
```diff
@@ -35,14 +35,26 @@ def context_urls
...
 
           proxy
         end
       end
+
+      # Find proxy by URL configured in templates module
+      # or by proxy's URL
+      def find_smart_proxy_by_url
+        proxy_from_feature = SmartProxyFeature.where(feature: Feature.find_by(name: 'Templates'))
+                                              .where('settings LIKE ?', "%#{params[:url]}%")
```
ekohl, line 51: https://www.binarywebpark.com/query-json-data-rails-postgresql/ suggests the native way to query JSON is:
```suggestion
                                              .where('settings ->> 'template_url' = ?', params[:url])
```

app/controllers/katello/concerns/api/v2/registration_controller_extensions.rb
```diff
@@ -35,14 +35,26 @@ def context_urls
...
+                                              .first&.smart_proxy
```
ekohl, line 52: Doesn't this still generate 2 queries? 1 to get `SmartProxyFeature` and then another one for the SmartProxy instance.

ekohl (changes requested):

ekohl (commented):
I'm going to throw out a completely different solution here, which I already suggested way back in the beginning but what was not implemented. However, it's going to be cleaner I think.

The request coming from the Smart Proxy can be authenticated using client certificates. This is how a Smart Proxy identifies itself to Foreman.

To start, we have the request issues by the Smart Proxy. Luckily, that is already done [here](https://github.com/theforeman/smart-proxy/blob/a48da738dd0f41cb28c7042e5f2773a8d9a29293/lib/proxy/request.rb#L65-L87).

Then on the receiving end you have to authenticate. We have the [SmartProxyAuth concern](https://github.com/theforeman/foreman/blob/develop/app/controllers/concerns/foreman/controller/smart_proxy_auth.rb). I'm not sure if `require_smart_proxy_or_login` method is correct for this use case (was authentication required or not?), but if it it's then it's simple. After that you get the authenticated `detected_proxy` property, regardless of whatever URL was used.

stejskalleos:
> The request coming from the Smart Proxy can be authenticated using client certificates. This is how a Smart Proxy identifies itself to Foreman.

Sorry I don't follow much, can you please explain how is this related to the load balancing and registration and why to do it? Do I understand correctly that it's an another level of security for registration?

ekohl:
A Smart Proxy has SSL client certificates and can use those to authenticate. On the certificate is the name of the Smart Proxy (stored in the `commonName`). What happens is:
* Apache validates the client certificate (authentication)
* Apache passes that information on to Foreman
* Foreman can use the authentication data to authorize. Most controllers in Foreman ignore the data, but where it does need to the `SmartProxyAuth` concern is used.

Now you have an authenticated way to determine which Smart Proxy was used, regardless of which URL was passed in.

One example is the config reports endpoint. It first adds a filter that checks, additionally that it has the right features enabled:
https://github.com/theforeman/foreman/blob/debea6c156b9508188e03cabc63093bc19f62956/app/controllers/api/v2/config_reports_controller.rb#L11

Then in the actual method it can use the detected proxy:

https://github.com/theforeman/foreman/blob/debea6c156b9508188e03cabc63093bc19f62956/app/controllers/api/v2/config_reports_controller.rb#L39-L47

Note it's still optional there since a regular user can also log in and use the API endpoint as well.

So it does add an additional layer of security, but it also makes it very reliable in detecting which Smart Proxy since it no longer relies on the passed in URL which could go out of sync. For example, when a user changes the registration URL on a Smart Proxy but forgets to refresh the features. Then your detection would fail, but the certificate authentication still works.

stejskalleos:
@ekohl I added `SmartProxyAuth` for smart proxy search, however I wasn't able to make the SSL authentication work on my setup, [this condition](https://github.com/theforeman/foreman/blob/develop/app/controllers/concerns/foreman/controller/smart_proxy_auth.rb#L52) is always `false`, even when I send the certs, no idea why.

But with mocked condition for the certificate registration works and I've been able to register and subscribe hosts without any problems.

ekohl:
> however I wasn't able to make the SSL authentication work on my setup, [this condition](https://github.com/theforeman/foreman/blob/develop/app/controllers/concerns/foreman/controller/smart_proxy_auth.rb#L52) is always `false`, even when I send the certs, no idea why.

I guess it depends on your setup. If you use Apache as a reverse proxy in front of it, you do need to set variables to pass it on. These are the settings that are used to determine the environment variables:
https://github.com/theforeman/foreman/blob/b2d06d0ddbf01923d527feef16c4aef97aada881/app/services/foreman/client_certificate.rb#L49-L59
Our installer sets them here:
https://github.com/theforeman/puppet-foreman/blob/master/files/settings-reverse-proxy-headers.yaml

If you use Puma directly, the values may be different. Be sure to pass `verify_mode` and set it to `peer`. I think that the raw certificate is passed raw as the env variable `puma.peercert`. At least if I look at https://github.com/puma/puma/blob/03ed6c8e78b9ba0c76055af2ddb0bdd532807d6c/lib/puma/request.rb#L58-L60 & https://github.com/puma/puma/blob/03ed6c8e78b9ba0c76055af2ddb0bdd532807d6c/lib/puma/const.rb#L194

app/controllers/katello/concerns/api/v2/registration_controller_extensions.rb
```diff
@@ -35,14 +36,19 @@ def context_urls
...
+          proxy = params[:url] ? find_smart_proxy_by_url : SmartProxy.pulp_primary
 
           fail Foreman::Exception, _('Smart proxy content source not found!') unless proxy
           fail Foreman::Exception, _('Pulp 3 is not enabled on Smart proxy!') unless proxy.pulp3_enabled?
 
           proxy
         end
       end
+
+      def find_smart_proxy_by_url
```
ekohl, line 48: This doesn't really find it by URL anymore so I'd say this should have a better name

app/controllers/katello/concerns/api/v2/registration_controller_extensions.rb
```diff
@@ -35,14 +36,19 @@ def context_urls
...
+        auth_smart_proxy
+        return @detected_proxy if @detected_proxy
```
ekohl, line 50: I think it's sufficient to use
```suggestion
        @detected_proxy
```

app/controllers/katello/concerns/api/v2/registration_controller_extensions.rb
```diff
@@ -35,14 +36,19 @@ def context_urls
 
       def smart_proxy
         @smart_proxy ||= begin
-          proxy = params[:url] ? SmartProxy.unscoped.find_by(url: params[:url]) : SmartProxy.pulp_primary
+          proxy = params[:url] ? find_smart_proxy : SmartProxy.pulp_primary
```
ekohl, line 39: Reviewers note: I was a bit tempted to suggest always authenticating the Smart Proxy and seeing if it can, but that can raise warnings. `params[:url]` should be set if a Smart Proxy sends a request so that's good enough for now. In the future we may want to revisit this.

ekohl (commented):
Code wise :+1: though I'd probably update the commit message to reflect the current change. Describe that it now uses client certificates to identify a Smart Proxy. I'll also admit that I don't have a setup to verify this right now.

Any Katello developer who could verify this?

stejskalleos:
@jeremylenz Looking for Katello dev to do the final ACK, could you add this to your review board?

ianballou (commented):
Perhaps I'm missing something, but trying to register a host through a smart proxy via the usual method (just entering a smart proxy in the host registration form) is failing:

```
21:19:01 rails.1   | 2022-11-03T21:19:01 [I|app|a812306b] Started GET "/register?activation_keys=Default+Library+Key&force=true&location_id=2&operatingsystem_id=3&organization_id=1&setup_remote_execution_pull=true&update_packages=false&url=https%3A%2F%2Fcentos8-proxy-devel.cannolo.example.com%3A9090" for 192.168.122.243 at 2022-11-03 21:19:01 +0000
21:19:01 rails.1   | 2022-11-03T21:19:01 [I|app|a812306b] Processing by Api::V2::RegistrationController#global as HTML
21:19:01 rails.1   | 2022-11-03T21:19:01 [I|app|a812306b]   Parameters: {"activation_keys"=>"Default Library Key", "force"=>"true", "location_id"=>"2", "operatingsystem_id"=>"3", "organization_id"=>"1", "setup_remote_execution_pull"=>"true", "update_packages"=>"false", "url"=>"https://centos8-proxy-devel.cannolo.example.com:9090", "registration"=>{}}
21:19:01 rails.1   | 2022-11-03T21:19:01 [I|app|a812306b] Authorized user admin(Admin User)
21:19:01 rails.1   | 2022-11-03T21:19:01 [D|app|a812306b] Post-login processing for admin
21:19:01 rails.1   | 2022-11-03T21:19:01 [D|app|a812306b] Unpermitted parameters: :operatingsystem_id, :update_packages, :url, :registration
21:19:01 rails.1   | 2022-11-03T21:19:01 [W|app|a812306b] No SSL cert with CN supplied - request from 192.168.122.243
21:19:01 rails.1   | 2022-11-03T21:19:01 [W|app|a812306b] Action failed
21:19:01 rails.1   | 2022-11-03T21:19:01 [I|app|a812306b] Backtrace for 'Action failed' error (Foreman::Exception): ERF42-9394 [Foreman::Exception]: Smart proxy content source not found!
21:19:01 rails.1   |  a812306b | /home/vagrant/katello/app/controllers/katello/concerns/api/v2/registration_controller_extensions.rb:41:in `smart_proxy'
```

My Foreman and smart proxy have both been updated with the related code.

I'm not totally understanding how the smart proxy URL is automatically decided by `auth_smart_proxy`.  I'm seeing the `No SSL cert with CN supplied - request from 192.168.122.243` error... is more setup necessary?

I'm testing with a Katello development environment and one of the proxy-devel vagrant boxes.

ekohl:
Looking at puppet-katello_devel I think some settings may be missing (which means the Smart Proxy auth never worked in the devel setup?) Can you check if https://github.com/theforeman/puppet-foreman/blob/master/files/settings-reverse-proxy-headers.yaml is present in your `settings.yaml`?

ianballou:
Those settings are indeed missing from my development environment. I'll add them in and see what changes.

ianballou (approved):
I didn't test with haproxy, but after adding the correct Foreman settings this worked just fine for me. I tested registering through a smart proxy and changing the content source a couple times.

If you think an actual test with haproxy is needed, I'd appreciate some steps for configuring it how a user actually would.

Anyway, from my standpoint, I'm cool with merging it. We'll just need to update the katello_devel puppet module.

ekohl:
I don't think it's needed to test it out with HAProxy now. Please merge it. And we do indeed also need to update the katello_devel module. Thanks for that.

PR #10318, opened 2022-10-14
Fixes #35633 - expose search params for product_content endpoints

evgeni:
Fixes: 49799266289075aa71bb5832c841373fd7d04f7b

This adds the `search` param group (so `per_page`, `search`, etc parameters) to the `activation_keys/:id/product_content` and `host_subscriptions/:id/product_content` endpoints. Without these params api clients won't send them to the server, thus only getting the default number of results instead of the requested one.

"make it work" ;-)

The generated apidoc should contain search related (especially `per_page`) params for the two endpoints


evgeni:
before:
Screenshot from 2022-10-14 10-57-06

after:
Screenshot from 2022-10-14 10-57-23

chris1984 (approved):
Looks good and can see the changes in the apidoc when I regen the apipie cache.

PR #10325, opened 2022-10-21
Fixes #33464 - Bulk ACS delete and refresh

ianballou:
Allows bulk refreshing and deleting ACSs

No permissions yet, but it sounds like we need to do an entire ACS permission audit anyway.
For the checkbox `<Th>` tag, I set the width to 0 to match the look of other pages. Is that fine?

1) Create some ACSs
2) Try refreshing them all via a bulk action
3) Try deleting them all via a bulk action
4) Ensure viewing ACS details still works
5) Ensure single ACS refreshing, editing, and deleting still works

TODO:
- [x] Tests

### DIFF
--- a/webpack/scenes/AlternateContentSources/ACSActions.js
+++ b/webpack/scenes/AlternateContentSources/ACSActions.js
@@ -8,6 +8,8 @@ import ACS_KEY, {
   DELETE_ACS_KEY,
   EDIT_ACS_KEY,
   PRODUCTS_KEY,
+  BULK_ACS_REFRESH_KEY,
+  BULK_ACS_DELETE_KEY,
 } from './ACSConstants';
 import { getResponseErrorMsgs } from '../../utils/helpers';
 import { renderTaskStartedToast } from '../Tasks/helpers';
@@ -76,6 +78,30 @@ export const refreshACS = (acsId, handleSuccess) => post({
     return renderTaskStartedToast(response.data);
   },
   errorToast: error => acsErrorToast(error),
+});
+
+export const bulkRefreshACS = (params, handleSuccess) => post({
+  type: API_OPERATIONS.POST,
+  key: BULK_ACS_REFRESH_KEY,
+  url: api.getApiUrl('/alternate_content_sources/bulk/refresh'),
+  params,
+  handleSuccess: (response) => {
+    renderTaskStartedToast(response?.data, __('Bulk alternate content source refresh has started.'));
+    return handleSuccess();
+  },
+  errorToast: error => __('Something went wrong while refreshing alternate content sources: ') + getResponseErrorMsgs(error.response),
+});
+
+export const bulkDeleteACS = (params, handleSuccess) => APIActions.put({
+  type: API_OPERATIONS.PUT,
+  key: BULK_ACS_DELETE_KEY,
+  url: api.getApiUrl('/alternate_content_sources/bulk/destroy'),
+  params,
+  handleSuccess: (response) => {
+    renderTaskStartedToast(response?.data, __('Bulk alternate content source delete has started.'));
+    return handleSuccess();
+  },
+  errorToast: error => __(`Something went wrong while deleting alternate content sources: ${getResponseErrorMsgs(error.response)}`),
 });
 
 export const getProducts = () => get({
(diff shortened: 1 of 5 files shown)


ianballou:
I got the tests passing and added one that checks for the "Select all" to be disabled. Let me know if I should add others. I didn't find too many tests specifically around checkbox selection for the content views page...

webpack/scenes/AlternateContentSources/MainTable/ACSTable.js
```diff
@@ -249,9 +275,10 @@ const ACSTable = () => {
             variant={TableVariant.compact}
             additionalListeners={[activeSortColumn, activeSortDirection]}
             autocompleteEndpoint="/alternate_content_sources/auto_complete_search"
+            {...selectionSetVars}
             actionButtons={
               <>
-                {status === STATUS.RESOLVED && results?.length !== 0 &&
+                {status === renderActionButtons &&
```
sjha4, line 281:
```suggestion
               {renderActionButtons &&
```

webpack/scenes/AlternateContentSources/ACSActions.js
```diff
@@ -78,6 +80,30 @@ export const refreshACS = (acsId, handleSuccess) => post({
   errorToast: error => acsErrorToast(error),
 });
 
+export const bulkRefreshACS = (params, handleSuccess) => post({
+  type: API_OPERATIONS.POST,
+  key: BULK_ACS_REFRESH_KEY,
+  url: api.getApiUrl('/alternate_content_sources/bulk/refresh'),
+  params,
+  handleSuccess: (response) => {
+    renderTaskStartedToast(response?.data);
```
sjha4, line 89: Since this renderTaskStartedToast is contained inside bulkRefreshACS , we should probably try to get it to say that Bulk refresh started or Bulk delete started.  renderTaskStartedToast will probably need to be updated a little. Feel free to tackle it here or open an issue for this so we can handle and test this on other bulk task toasts.

webpack/scenes/AlternateContentSources/ACSActions.js
```diff
@@ -78,6 +80,30 @@ export const refreshACS = (acsId, handleSuccess) => post({
...
+    return handleSuccess();
+  },
+  errorToast: error => __(`Something went wrong while refreshing alternate content sources: ${getResponseErrorMsgs(error.response)}`),
```
sjha4, line 92: We cannot translate dynamic values like getResponseErrorMsgs(error.response)
This should work except for translating error response message.
```suggestion
  errorToast: error => __(`Something went wrong while refreshing alternate content sources: `) + getResponseErrorMsgs(error.response),
```

webpack/scenes/AlternateContentSources/ACSActions.js
ianballou: I added an override string to renderTaskStartedToast

webpack/scenes/AlternateContentSources/MainTable/ACSTable.js
```diff
@@ -80,6 +93,19 @@ const ACSTable = () => {
       dispatch(getAlternateContentSources())));
   };
 
+  const onBulkDelete = (ids) => {
+    setDeleting(true);
+    dispatch(bulkDeleteACS({ ids }, () => {
+      setDeleting(false);
+      dispatch(getAlternateContentSources());
```
sjha4, line 100: We'd want here is a check to see if one of the selected ACS to be deleted is also expanded with URL as /details/:id cause that throws an error. You can re-purpose the handleSuccess function code in onDelete to check if acsId is in the ids list and push('/alternate_content_sources') if it is.


```
const onBulkDelete = (ids) => {
    setDeleting(true);
    dispatch(bulkDeleteACS({ ids }, () => {
      setDeleting(false);
      if (acsId && ids.has(Number(acsId))) {
        push('/alternate_content_sources');
      } else {
        dispatch(getAlternateContentSources());
      }
    }));
  };
```

ianballou: Pushed the fix!

sjha4 (approved):
Ack 👍🏼 🥳

PR #10326, opened 2022-10-24
Fixes #35661 - Handle ACS permissions on UI

sjha4:
Show/hide actions on ACS UI based on user permissions
Create a user with viewer role and refresh ACS UI after impersonating the user.
Make sure actions like refresh/delete/create/edit are hidden for the read-only user.
Try other combinations of ACS permissions and make sure actions are displayed/hidden accordingly.


ianballou:
It looks like the viewer for some reason doesn't see the Alternate Content Sources link on the side menu. Unless your branch here is out of date and doesn't have that change?

ianballou:
I tested the permissions and it looks fine! Only the permissions issue with actually seeing the Alternate  Content Sources link is there.

app/views/katello/api/v2/alternate_content_sources/permissions.rabl
```diff
@@ -0,0 +1,8 @@
+collection @alternate_content_sources
+
+if params.key?(:include_permissions)
```
ianballou, line 3: `/katello/api/v2/alternate_content_sources?include_permissions=false` also causes the permissions to be returned.

ianballou: Why is the create permission not included by default?

ianballou (commented):
Just one comment, otherwise looks good!

app/views/katello/api/v2/alternate_content_sources/permissions.rabl
sjha4: > Why is the create permission not included by default?

It would be an additional meaningless(?) field for others consuming our API. The permission is only needed for the UI. 

Will update this: `if params.key?(:include_permissions)`

ianballou: I meant more like why do you only need "can_create" and not "can_edit" or "can_view" ?

sjha4: `can_edit `and `can_view` are more record specific so they live in the permissions of individual records whereas for create, especially with empty results when there are no ACSs, we still want to know if user has create permissions for ACS in general to show/hide the create button on the UI.

app/views/katello/api/v2/alternate_content_sources/permissions.rabl
```diff
@@ -0,0 +1,8 @@
+collection @alternate_content_sources
+
+if params[:include_permissions]
```
ianballou, line 3: The params here might be strings, which will always be truthy.

I curled with `/katello/api/v2/alternate_content_sources?include_permissions=false"`

app/views/katello/api/v2/alternate_content_sources/base.json.rabl
```diff
@@ -28,7 +28,7 @@ child :latest_dynflow_refresh_task => :last_refresh do |_object|
   end
 end
 
-if params.key?(:include_permissions)
+if params[:include_permissions]
```
ianballou, line 31: Same issue as the other comment.

sjha4: Updated to cast falsy values to bool.

sjha4:
@ianballou : Rebased and added permissions for bulk actions. Hides bulk action when can_edit and can_delete are both false, and disable refresh/delete bulk action when one of them is false.

ianballou (commented):
For some reason, the bulk "Delete" button is staying greyed out for my user that has ACS delete permissions. The non-bulk one is not grey.

app/views/katello/api/v2/alternate_content_sources/permissions.rabl
```diff
@@ -0,0 +1,10 @@
+collection @alternate_content_sources
+
+if ::Foreman::Cast.to_bool(params.fetch(:include_permissions, false))
+  user = User.current # current_user is not available here
+  node do
+    node(:can_create) { user.can?("create_alternate_content_sources") }
+    node(:can_edit) { user.can?("edit_alternate_content_sources") }
+    node(:can_delete) { user.can?("delete_alternate_content_sources") }
```
ianballou, line 8:
```suggestion
    node(:can_delete) { user.can?("destroy_alternate_content_sources") }
```

ianballou (changes requested):

ianballou:
Should the right side kebab also be removed when the user cannot refresh nor delete?

sjha4:
> Should the right side kebab also be removed when the user cannot refresh nor delete?

Updated to hide side kebab unless user has atleast edit or delete perms. And updated to refer destroy_alternate_content_sources.

webpack/scenes/AlternateContentSources/MainTable/ACSTable.js
```diff
@@ -389,9 +404,13 @@ const ACSTable = () => {
                       emptyMessage="N/A"
                     />
                     </Td>
-                    <Td isActionCell>
-                      <ActionsColumn items={rowDropdownItems(acs)} />
-                    </Td>
+                    {(hasPermission(permissions, 'destroy_alternate_content_sources') ||
```
ianballou, line 407: why can't you use `canDelete` and `canEdit` here?

ianballou (approved):
Looks good to me! I had one question, but I'm not blocking on it.

webpack/scenes/AlternateContentSources/MainTable/ACSTable.js
sjha4: CanDelete and canEdit are object level. The hasPermissions will check record level permissions for the row actions which is more accurate.

PR #10328, opened 2022-10-27
Fixes #35503 - Notify user only when new errata are added to repo

wbclark:
Co-authored-by: William Bradford Clark <wclark@redhat.com>
Co-authored-by: Pavel Moravec <pmoravec@redhat.com>

Instead of using contents_changed for input (and corresponding middleware) we use new inputs associated_errata_before_syncing and new_associated_errata; in planning, we set the former to the current list of associated errata and initialize the latter as an empty array. Later in the run step (after the repository has synced) we re-calculate the associated errata and set new_associated_errata as the difference -- only sending the notification email when this is non-empty.

To avoid keeping potentially large arrays in the database (and to create better UX viewing the action in the dynflow console), after we are done _using_ the new inputs to determine what (if any) emails to send, we truncate them and replace with a message displaying the count of errata rather than the complete list of IDs. More than 3 errata was chosen (somewhat arbitrarily) as a cutoff point, because the full information will be available in the email itself. This prevents us from seeing very large inputs for the action when viewing the action in dynflow console _after the action has run_.

1. Install sendmail/postfix on development environment
2. systemctl enable --now postfix
3. Configure email address and notification subscriptions for user
4. Enable and sync a repository containing errata
5. After sync is complete, observe task in dynflow console and validate that we have `associated_errata_before_syncing: []` and some number of `new_associated_errata`
6. Read mail in `/var/spool/mail/$user`
7. Immediately re-sync the repository
8. This time the inputs are reversed -- we observe some number of existing `associated_errata_before_syncing` but `new_associated_errata` is `[]`
9. There is no additional mail in `/var/spool/mail/$user`

### DIFF
--- a/app/lib/actions/katello/repository/errata_mail.rb
+++ b/app/lib/actions/katello/repository/errata_mail.rb
@@ -2,18 +2,21 @@ module Actions
   module Katello
     module Repository
       class ErrataMail < Actions::EntryAction
-        middleware.use Actions::Middleware::ExecuteIfContentsChanged
-
-        def plan(repo, contents_changed = nil)
-          last_updated = repo.repository_errata.order('updated_at ASC').last.try(:updated_at) || Time.now
-          plan_self(:repo => repo.id, :contents_changed => contents_changed, :last_updated => last_updated.to_s)
+        def plan(repo)
+          plan_self(:repo => repo.id, :associated_errata_before_syncing => repo.repository_errata.pluck(:erratum_id).uniq.sort.reverse, :new_associated_errata => [])
         end
 
         def run
           ::User.current = ::User.anonymous_admin
           repo = ::Katello::Repository.find(input[:repo])
+          input[:new_associated_errata] = repo.repository_errata.pluck(:erratum_id).uniq.sort.reverse - input[:associated_errata_before_syncing]
+
           users = ::User.select { |user| user.receives?(:sync_errata) && user.organization_ids.include?(repo.organization.id) && user.can?(:view_products, repo.product) }.compact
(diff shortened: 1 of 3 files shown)


pmoravec:
Credits / co-author should rather go to @hao-yu than me :)

app/lib/actions/katello/repository/errata_mail.rb
```diff
@@ -2,18 +2,21 @@ module Actions
   module Katello
     module Repository
       class ErrataMail < Actions::EntryAction
-        middleware.use Actions::Middleware::ExecuteIfContentsChanged
-
-        def plan(repo, contents_changed = nil)
-          last_updated = repo.repository_errata.order('updated_at ASC').last.try(:updated_at) || Time.now
-          plan_self(:repo => repo.id, :contents_changed => contents_changed, :last_updated => last_updated.to_s)
+        def plan(repo)
+          plan_self(:repo => repo.id, :associated_errata_before_syncing => repo.repository_errata.pluck(:erratum_id).uniq.sort.reverse, :new_associated_errata => [])
```
pmoravec, line 6: Why the `.reverse`?

wbclark:
> Credits / co-author should rather go to @hao-yu than me :)

I didn't list @hao-yu as co-author because he is the primary author (of the commit itself). If that is confusing or contrary to convention I could add his information in both places (first time using this GH feature, but it seems very useful)

app/lib/actions/katello/repository/errata_mail.rb
pmoravec: Ah I see, then it makes sense to me.

pmoravec (approved):

ianballou (approved):
Worked fine for me! I couldn't get Foreman to connect to postfix for some reason, but the emails in the logs show the correct thing.  Only the added errata are shown in the email.

I like the way the dynflow input is setup, makes for easy debugging!

PR #10331, opened 2022-10-28
Fixes #35688 - Monkey patch pulp_container binding to allow syncing from 2.14 -> 2.10

sjha4:
Monkey patch Container RepositorySyncURL class to not set signed_only.
pulp_container 2.14 adds a signed_only field to RepositorySyncURL class and send that as part of API call to sync repos on the proxy. That throws an error on proxies with pulpcore3.18 as the field is not recognized on that version

We have an issue filed with pulp: https://github.com/pulp/pulp_container/issues/1122
If this PR solves the issue in the short term for us on current versions, it will unblock our nightlies.

1. Get a main server with pulpcore3.21
2. Get a proxy with pulpcore3.18
3. Sync a container repo on main server
4. Try a capsule sync and notice the error.

On this branch, the capsule sync for container repo should pass.

### DIFF
--- a/config/initializers/monkeys.rb
+++ b/config/initializers/monkeys.rb
@@ -2,3 +2,4 @@
 require 'monkeys/anemone'
 require 'monkeys/ar_postgres_evr_t'
 require 'monkeys/fx_sqlite_skip'
+require 'monkeys/try_pulp_container_path'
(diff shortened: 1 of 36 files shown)


sjha4:
Expected VCR failures due to change in API call. WIll re-record once this is tested as working OK for a setup other than mine.

sjha4:
> Expected VCR failures due to change in API call. WIll re-record once this is tested as working OK for a setup other than mine.

Added a commit with updated cassettes for the monkey patched gem.

ianballou (approved):
Worked for me!

PR #10332, opened 2022-11-02
Fixes #35709 - index application/vnd.oci.image.manifest.v1+json manifests

ianballou:
Adds support for the 'application/vnd.oci.image.manifest.v1+json' type so indexing doesn't fail. Also updates docker tags to be "mutable" so that, during indexing, they get "upserted" and thus corrected in case the user has bad docker tags in their DB.  Without this, docker tags that are missing manifests would not be updated.

Setting docker tags to be mutable should be fine because they are a mutable concept anyway.

Try to sync ansible/ansible-runner from quay.io.  It should not fail.

I also noticed that the docker UI really needs some help. I made an issue: https://projects.theforeman.org/issues/35710

### DIFF
--- a/app/services/katello/pulp3/docker_manifest.rb
+++ b/app/services/katello/pulp3/docker_manifest.rb
@@ -16,7 +16,8 @@ module Katello
 
       def self.content_unit_list(page_opts)
         page_opts[:media_type] = ['application/vnd.docker.distribution.manifest.v1+json',
-                                  'application/vnd.docker.distribution.manifest.v2+json']
+                                  'application/vnd.docker.distribution.manifest.v2+json',
+                                  'application/vnd.oci.image.manifest.v1+json']
         self.content_api.list(page_opts)
       end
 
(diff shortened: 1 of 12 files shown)


app/services/katello/pulp3/docker_manifest.rb
```diff
@@ -16,7 +16,8 @@ def self.ids_for_repository(repo_id)
 
       def self.content_unit_list(page_opts)
         page_opts[:media_type] = ['application/vnd.docker.distribution.manifest.v1+json',
-                                  'application/vnd.docker.distribution.manifest.v2+json']
+                                  'application/vnd.docker.distribution.manifest.v2+json',
+                                  'application/vnd.oci.image.manifest.v1+json']
```
parthaa, line 20: Should we have more entries  https://docs.pulpproject.org/pulp_container/restapi.html#tag/Content:-Manifests/operation/content_container_manifests_list

Seem to see more media types there

ianballou: The remaining two types relate to manifest lists it appears. One of which is missing from the manifest lists filter... I should add it.

parthaa (approved):
Works well. Was able to sync

PR #10339, opened 2022-11-08
Fixes #35734 - switch to async endpoint for manifest refresh

jeremylenz:

In Katello, a manifest refresh is comprised of a few steps:

1. Download the manifest from hosted Candlepin
2. Write the contents to a zip file in a temporary location
3. Import the manifest from the zip file

The change here is that the endpoint to retrieve manifests is now split into two: an async endpoint to start "exporting" the manifest, and a second endpoint to download the exported manifest. Previously this was a single, synchronous endpoint.

So the new "step 1" now looks like
1. Call `/subscription/consumers/{consumerId}/export/async` to start the export
2. Poll the upstream job by calling `/subscription/jobs/{jobId}`
3. When it's complete, retrieve the exported manifest by calling `/subscription/consumers/{consumerId}/export/{exportId}`

I guess this is the first time we've polled async jobs in Hosted (upstream) Candlepin. (We've been using an async endpoint for manifest imports, but that's in local Candlepin.) So, I had to make some new Dynflow actions and Candlepin resources to cover upstream job polling.

Test manifest refresh
Test adding subscriptions to a manifest from the Subscriptions page

```
tail -f /home/vagrant/foreman/log/development.log | grep "Sending GET request to upstream Candlepin"
```

If you get any Candlepin errors:
```
cat /var/log/candlepin/error.log
```


app/lib/actions/candlepin/abstract_async_task.rb
```diff
@@ -30,6 +30,10 @@ def done?
         !::Katello::Resources::Candlepin::Job.not_finished?(external_task)
       end
 
+      def candlepin
+        "Candlepin"
```
parthaa, line 34:
```suggestion
        _("Candlepin")
```

parthaa (approved):
Works well with my tests ACK

PR #10340, opened 2022-11-08
Fixes #35657 - Fix orphan delete

sjha4:
1. Allows indexing ansible collections without tags
2. Allow orphan cleanup on proxies
Pulp deb has an issue with syncing content to proxies I ran into during working on this. I have filed an issue for that.
Create repos of all types.
Sync a proxy with library environment
Move proxy to a different env and sync again
On console run bundle exec rails katello:delete_orphaned_content.

### DIFF
--- a/app/services/katello/pulp3/ansible_collection.rb
+++ b/app/services/katello/pulp3/ansible_collection.rb
@@ -30,21 +30,24 @@ module Katello
       end
 
       def self.insert_child_associations(units, pulp_id_to_id)
-        tag_names = units.map { |unit| unit['tags'].map { |tag| tag[:name] } }.flatten
-        tag_rows = tag_names.map { |name| {name: name } }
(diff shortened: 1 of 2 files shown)


sjha4:
[test katello]

sjha4:
Tested on proxy with 3.18 and 3.21..

ianballou:
I can't  sync my ansible collections repo to my smart proxy, so I filed this: https://projects.theforeman.org/issues/35746

ianballou (approved):
Other issues aside, this is fine by me!

gvde:
@sjha4 Is there are particular reason why this patch is not in 4.7?

ianballou:
@gvde it looks like the release config for 4.7 did not include the bugs triaged to 4.6.1. The config will need to be updated to include 4.6.1 and another release will be needed.

For some reference, here is the config for Katello 4.7: https://github.com/theforeman/tool_belt/blob/master/configs/katello/4.7.yaml#L10

PR #10308, opened 2022-10-10
Fixes #35610 - CV filter copies modules not in the source repo

hao-yu:

PR #10343, opened 2022-11-09
Fixes #35737 - Filter by id includes module streams of other arch

hao-yu:
The issue is not completely fix in PR #10308. It is still reproducible when filtering errata by ids.

This PR is to fix the remaining issue.


hao-yu:
@ianballou correct!

ianballou (approved):
Worked fine for me!

PR #10344, opened 2022-11-09
Bump dep warning on mirror_on_sync to 4.8

chris1984:


ianballou:
We're moving this since it isn't completely straight forward to fix and we need to get 4.7 out. Not all Pulp content types support the 3 different mirroring types.

PR #10341, opened 2022-11-08
Fixes #35735 - Stub pulp3support? for test smart proxy

sjha4:
#### What are the changes introduced in this pull request?
stub :pulp3_support? to return true for intermittently failing test.
#### Considerations taken when implementing this change?
This is safe to do as the mock proxy should return true for docker support anyway. It sometimes randomly returns false due to race conditions or whatever reason on CI. Safest option is to stub to return true.
#### What are the testing steps for this pull request?
Run the CI tests a few (hundred?) times. 😝

PR #10349, opened 2022-11-10
Fixes #35614 - extra modular packages and errata copied into CVV

ianballou:
Causes errata to no longer be copied into content view versions unless both its module streams and packages are satisfied.

This affects both copy_all_units and multi_copy_all_units. They should both work however since the parent method is called the same way.

Follow the reproducer steps on the issue: https://projects.theforeman.org/issues/35614

### DIFF
--- a/test/lib/util/errata_test.rb
+++ b/test/lib/util/errata_test.rb
@@ -36,29 +36,60 @@ module Katello
       @erratum2.save!
 
       @empty_erratum = Katello::Erratum.create(:errata_id => "empty", :pulp_id => "empty")
+
+      @modular_erratum = katello_errata(:modular)
+      @modular_rpm = katello_rpms(:modular_with_erratum_package)
+      @modular_erratum_package = @modular_erratum.packages.first
+      @module_stream = @modular_erratum_package.module_streams.first
+    end
+
+    # https://projects.theforeman.org/issues/35614
+    def test_filter_by_pulp_href_excludes_errata_with_missing_module_streams
+      # If the erratum's modular RPM count is no longer 1, the test no longer makes sense.
+      # Ensure that all modular packages are accounted for in the final assertion.
+      # The same goes for the number of module streams that the single RPM belongs to.
+      assert_equal 1, @modular_erratum.packages.count
+      assert_equal 1, @modular_erratum_package.module_streams.count
+      assert_equal [@erratum], filter_errata_by_pulp_href([@modular_erratum, @erratum],
+                                                          [@fedora_one.pulp_id, @fedora_two.pulp_id, @modular_rpm.pulp_id],
+                                                          [@fedora_one.filename, @fedora_two.filename, @modular_erratum_package.filename],
+                                                          [@module_stream.module_spec])
+    end
+
+    # https://projects.theforeman.org/issues/35614
+    def test_filter_by_pulp_href_includes_errata_with_proper_module_streams
+      # If the erratum's modular RPM count is no longer 1, the test no longer makes sense.
+      # Ensure that all modular packages are accounted for in the final assertion.
+      # The same goes for the number of module streams that the single RPM belongs to.
+      assert_equal 1, @modular_erratum.packages.count
+      assert_equal 1, @modular_erratum_package.module_streams.count
+      assert_equal [@erratum, @modular_erratum].sort, filter_errata_by_pulp_href([@modular_erratum, @erratum],
+                                                          [@fedora_one.pulp_id, @fedora_two.pulp_id, @modular_rpm.pulp_id, @module_stream.pulp_id],
+                                                          [@fedora_one.filename, @fedora_two.filename, @modular_erratum_package.filename],
+                                                          [@module_stream.module_spec])
     end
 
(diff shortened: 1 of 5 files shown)


test/lib/util/errata_test.rb
```diff
@@ -36,29 +36,60 @@ def setup
       @erratum2.save!
 
       @empty_erratum = Katello::Erratum.create(:errata_id => "empty", :pulp_id => "empty")
+
+      @modular_erratum = ::Katello::Erratum.find_by(errata_id: 'RHEA-2019-002')
+      @modular_rpm = ::Katello::Rpm.create(name: "foobarmod", filename: "foobarmod-1.3.4.rpm", pulp_id: 'foobarmodularrpm')
```
wbclark, line 41: nit: should this be defined in fixtures (like `@modular_erratum` was) or use this existing fixture? https://github.com/Katello/katello/blob/ea9570ae2e2529b54f20f113abdab86b83371040/test/fixtures/models/katello_rpms.yml#L40-L51

test/lib/util/errata_test.rb
```diff
@@ -36,29 +36,60 @@ def setup
...
+                                                          [@module_stream.module_spec])
+    end
+
+    # https://projects.theforeman.org/issues/35614
+    def test_filter_by_pulp_href_includes_errata_with_proper_module_streams
+      # If the erratum's modular RPM count is no longer 1, the test no longer makes sense.
+      # Ensure that all modular packages are accounted for in the final assertion.
+      # The same goes for the number of module streams that the single RPM belongs to.
+      assert_equal 1, @modular_erratum.packages.count
+      assert_equal 1, @modular_erratum_package.module_streams.count
```
wbclark, line 65: these are repeated from the test case above.... can they be extracted from both test cases into a 3rd test case that runs only once?

wbclark (approved):
Left some small questions/comments on the tests, but overall this worked very nicely for me, testing via the steps at https://projects.theforeman.org/issues/35614

The one nuance of testing was, for the "Include all packages without errata" filter, I had to create a dummy rule to make the checkbox appear properly in the UI, then select it and remove the dummy rule. (Thanks @ianballou for the tip!)

I then validated that this option persisted on the filter:

```
[1] pry(main)> ::Katello::ContentViewFilter.find(3)
=> #<Katello::ContentViewPackageFilter:0x00005624d3ddadd0
 id: 3,
 content_view_id: 2,
 name: "testfilter3",
 created_at: Mon, 28 Nov 2022 16:15:58.331257000 UTC +00:00,
 updated_at: Mon, 28 Nov 2022 21:09:03.939056000 UTC +00:00,
 type: "Katello::ContentViewPackageFilter",
 inclusion: true,
 original_packages: true,
 description: "include all packages without errata",
 original_module_streams: false>
[2] pry(main)>
```

(The `original_packages: true` is what we needed to see)

Overall, it looks good to me.

test/lib/util/errata_test.rb
ianballou: I kept those test cases duplicated because they're testing that the test data is set up correctly.  They're meta-tests in a way. It seems weird to me to have separate unit tests dependent on each other.

test/lib/util/errata_test.rb
ianballou: Fixed!

wbclark:
From my PoV this is ready for merge, we just need a :green_apple: ACK. @sjha4 would you kindly give this a look?

test/lib/util/errata_test.rb
```diff
@@ -36,29 +36,60 @@ def setup
...
     def test_filter_by_pulp_id_returns_nothing_with_empty_list_of_errata
-      assert_empty filter_errata_by_pulp_href([], [@fedora_one.pulp_id], [@fedora_one.filename, @fedora_two.filename])
+      assert_empty filter_errata_by_pulp_href([], [@fedora_one.pulp_id], [@fedora_one.filename, @fedora_two.filename], [])
     end
 
-    # "Proper" errata here means errata with no RPMs that exist in its source repo any empty errata.
+    # "Proper" errata here means errata with no RPMs that exist in its source repo and any empty errata.
     def test_filter_by_pulp_id_returns_proper_errata_if_no_matching_packages
-      assert_equal [@erratum2, @empty_erratum].sort, filter_errata_by_pulp_href([@erratum, @erratum2, @empty_erratum], ["floop"], [@fedora_one.filename, @fedora_two.filename]).sort
+      assert_equal [@erratum2, @empty_erratum].sort, filter_errata_by_pulp_href([@erratum, @erratum2, @empty_erratum], ["floop"], [@fedora_one.filename, @fedora_two.filename], []).sort
```
sjha4, line 82: We had some issues with array sorts on activerecords not being consistent. I am not sure if that applies here.
https://github.com/Katello/katello/pull/10341#issuecomment-1310844946 but it might be better to use order_by(:id) just in case this introduces ambiguity.

ianballou: the `order` method only works on active record objects rather than arrays.  I did a quick test and the ordering seemed to be consistent for errata.

sjha4 (approved):
Ship it!👍🏼

PR #10354, opened 2022-11-14
Fixes #35752 - Make errata search filtered with ID work in Web UI

lfu:
Add `db_id` to allow errata search on errata's DB `id`.

Before
Try to search the Errata from GUI with Errata ID : " errata_id = RHSA-2020:2031"
Try to search the Errata from GUI with ID: "id  =  RHSA-2020:2031"

After
Try to search the Errata from GUI with: " errata_id = RHSA-2020:2031"
Try to search the Errata from GUI with: " id = RHSA-2020:2031"
Try to search the Errata from GUI with: "db_id  =  6597"

### DIFF
--- a/app/models/katello/erratum.rb
+++ b/app/models/katello/erratum.rb
@@ -24,7 +24,8 @@ module Katello
     has_many :cves, :class_name => "Katello::ErratumCve", :dependent => :destroy, :inverse_of => :erratum
     has_many :packages, :class_name => "Katello::ErratumPackage", :dependent => :destroy, :inverse_of => :erratum
 
-    scoped_search :on => :errata_id, :only_explicit => true
+    scoped_search :on => :id, :rename => :db_id, :only_explicit => true, :validator => ScopedSearch::Validators::INTEGER
+    scoped_search :on => :errata_id, :complete_value => true, :only_explicit => true
     scoped_search :on => :errata_id, :rename => :id, :complete_value => true, :only_explicit => true
     scoped_search :on => :title
     scoped_search :on => :title, :rename => :synopsis, :complete_value => true, :only_explicit => true


app/models/katello/erratum.rb
```diff
@@ -25,7 +25,7 @@ class Erratum < Katello::Model
     has_many :packages, :class_name => "Katello::ErratumPackage", :dependent => :destroy, :inverse_of => :erratum
 
     scoped_search :on => :errata_id, :only_explicit => true
-    scoped_search :on => :errata_id, :rename => :id, :complete_value => true, :only_explicit => true
+    scoped_search :on => :id, :rename => :db_id, :complete_value => true, :only_explicit => true, :validator => ScopedSearch::Validators::INTEGER
```
jeremylenz, line 28: `:complete_value => true` means that autocomplete will suggest values for it when searching in the UI. I don't find this useful for database ID, but I think we should have it for `errata_id`.

jeremylenz:
Looks better now, thanks :)

Will it cause problems if we keep the `:rename => :id` for `errata_id`?  That way you'd be able to search by errata_id with either 'id' or 'errata_id' (no change in current behavior), and would also now be able to search by `db_id`.

lfu:
> Will it cause problems if we keep the `:rename => :id` for `errata_id`? That way you'd be able to search by errata_id with either 'id' or 'errata_id' (no change in current behavior), and would also now be able to search by `db_id`.

It's doable. But don't you think `id` is confusing along with `errata_id` and `db_id`?

jeremylenz:
> It's doable. But don't you think `id` is confusing along with `errata_id` and `db_id`?

A bit. But users may be used to it, because both are available today.

jeremylenz:
Specifically I'm worried that if a user has some automation that performs a search by 'id', it will stop working after this change.

jeremylenz (approved):
LGTM :+1:

PR #10355, opened 2022-11-15
Fixes #35754 - ACS refresh by name via hammer does not work

ianballou:
Adds the ability to list ACSs by name. ACS refresh via hammer failed previously because hammer first calls the index API endpoint to find the ACS with the name in question. Then it calls the refresh API using the ID of the returned ACS.  The ACS index endpoint didn't support a name argument other than in the scoped search, so all ACSs were returned, causing the error.

I moved the `assert_response_ids` method out of the repositories controller test so it could be utilized by all tests.

1) Make an ACS
2) Try refreshing it via hammer
3) Try listing ACSs with their name via the API and hammer.


chris1984 (approved):
ACK, works well. Thanks for moving the test so it can be used by other tests.

PR #10356, opened 2022-11-16
Fixes #35730 - Show include artifact checkboxes for rpm and module stream

sjha4:
Move Include/Exclude all RPMs with no errata and Include/Exclude all module streams with no errata checkboxes on RPM and Module stream filter details view

1. Create a RPM filter and a module stream filter.
2. The Include/Exclude all RPMs with no errata and Include/Exclude all module streams with no errata should be visible and editable even without any filter rules added.

### DIFF
--- a/webpack/components/Table/EmptyStateMessage.js
+++ b/webpack/components/Table/EmptyStateMessage.js
@@ -11,7 +11,7 @@ import {
 } from '@patternfly/react-core';
 import PropTypes from 'prop-types';
 import { translate as __ } from 'foremanReact/common/I18n';
-import { CubeIcon, ExclamationCircleIcon, SearchIcon, CheckCircleIcon } from '@patternfly/react-icons';
+import { CubeIcon, ExclamationCircleIcon, SearchIcon, CheckCircleIcon, PlusCircleIcon } from '@patternfly/react-icons';
 import { global_danger_color_200 as dangerColor, global_success_color_100 as successColor } from '@patternfly/react-tokens';
 import { useDispatch, useSelector } from 'react-redux';
 import { selectHostDetailsClearSearch } from '../extensions/HostDetails/HostDetailsSelectors';
@@ -76,8 +76,8 @@ const EmptyStateMessage = ({
       >
         <KatelloEmptyStateIcon
           error={!!error}
-          search={search}
-          customIcon={customIcon}
+          search={search && !showPrimaryAction}
+          customIcon={PlusCircleIcon}
           happyIcon={happy}
         />
         <Title headingLevel="h2" size="lg" ouiaId="empty-state-title">
--- a/webpack/scenes/ContentViews/Details/Filters/ArtifactsWithNoErrata.js
+++ b/webpack/scenes/ContentViews/Details/Filters/ArtifactsWithNoErrata.js
@@ -1,7 +1,7 @@
 import React, { useState } from 'react';
 import { useDispatch } from 'react-redux';
 import PropTypes from 'prop-types';
-import { Checkbox } from '@patternfly/react-core';
+import { Switch } from '@patternfly/react-core';
 import { translate as __ } from 'foremanReact/common/I18n';
 import { editCVFilter } from '../ContentViewDetailActions';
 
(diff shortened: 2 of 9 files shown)


jeremylenz:
What's been bothering me about "Include all RPMs with no errata" is that it's unclear in a couple ways:

1. Does it mean "Include all RPMs and don't include any errata?" Or "Include all RPMs which do not contain any errata?" I think it's the latter. Is there any better wording we could use?
1. In the context of the explicit list below: It actually means "In addition to the list explicitly defined below, include [all RPMs with no errata]." So in effect, it affects what RPMs are on the Include list, even though those are not explicitly displayed.

How about "In addition to the list below, include all RPMs which do not contain any errata"?

sjha4:
> 1. Include all RPMs which do not contain any errata?

That's the right one. The usecase for this is mostly together with errata filters. Errata filters add errata + packages for that errata to the new version. To include/exclude all packages that would be filtered out because they don't have any errata, we have this checkbox on package filters.

We could add a help tooltip with a detailed description for what that field does..

jeremylenz:
> We could add a help tooltip with a detailed description for what that field does..

+1 to that!

ianballou:
> How about "In addition to the list below, include all RPMs which do not contain any errata"?

Other way around, "...include all RPMs that 'are not listed in' / 'do not belong to' any errata."

sjha4:
Any thoughts on whether we should rename the field?

jeremylenz:
> Other way around, "...include all RPMs that 'are not listed in' / 'do not belong to' any errata."

That makes so much more sense!!

ianballou:
> Any thoughts on whether we should rename the field?

Which field do you mean @sjha4 ?

ianballou:
This PR is making me realize we have no way to delete filters from their details pages...

ianballou:
I'm seeing an issue where sometimes the filter's "exclude all" or "include all" switch isn't in the correct position. A couple times I've flipped it on or off, published a version, looked back at it, and it's showing the opposite.

MariSvirik:
@sjha4 empty state looks good. Just change an icon to "plus". 
https://www.patternfly.org/v4/components/empty-state/design-guidelines

Screenshot 2022-11-18 at 15 13 40

For a better microcopy, I'll ping Akshay.

MariSvirik:
@sjha4 
1. remove the period at the end of the label - it's viewed as a setting.
2. switcher and label text should be center aligned https://www.patternfly.org/v4/components/switch 

Screenshot 2022-11-18 at 16 00 10
 not a sentence.

sjha4:
> > Any thoughts on whether we should rename the field?
> 
> Which field do you mean @sjha4 ?

I meant the `Include all RPMs without errata` label for the switch. It still makes sense to me but I am open to changing it if that's the consensus.

sjha4:
Screenshot from 2022-11-18 10-23-40

sjha4:
> @sjha4 empty state looks good. Just change an icon to "plus". https://www.patternfly.org/v4/components/empty-state/design-guidelines
> 
> Screenshot 2022-11-18 at 15 13 40
> 

Updated. 
@jeremylenz : Change for this makes sense to me but changes pages outside CV UI. Let me know if it makes sense.

ianballou:
> I meant the `Include all RPMs without errata` label for the switch. It still makes sense to me but I am open to changing it if that's the consensus.

What do we think about a slight rewording to indicate that errata own RPMs and not the other way around?

`Include/Exclude all RPMs not associated to errata`

ianballou:
> @ianballou would this be for the label? "Switch on (switcher) Include/Exclude all RPMs not associated to errata"

@MariSvirik yeah, the text that goes next to the switch. And for include filters it would start with "Include", and for exclude filters it would start with "Exclude" (just to be clear that I wasn't asking to actually put "Include/Exclude" with the slash).

sjha4:
Added kebab next to dropdown with delete action.
Screenshot from 2022-11-18 12-08-19

AkshayGadhaveRH:
@MariSvirik slightly rewording @ianballou's suggestion: ` Include all RPMs not associated with any errata`; Does this sound better?

sjha4:
> @MariSvirik slightly rewording @ianballou's suggestion: ` Include all RPMs not associated with any errata`; Does this sound better?

@AkshayGadhaveRH .. That sounds good. I will update the text.

ianballou (approved):
This is working well for me!

sjha4:
Merging this..Thanks all for the feedback and reviews!

PR #10360, opened 2022-11-18
Fixes #35168 - stop using single quotes that were breaking translations

jeremylenz:

In `engines/bastion/app/views/bastion/layouts/assets.html.erb`, we were calling

```js
angular.fromJson('<%= Katello::RepositoryTypeManager.enabled_repository_types.values.to_json.html_safe %>')
```

and other similar ERB expressions, enclosing them in single quotes. But if one of the translation values contained a single quote or apostrophe character, JavaScript would interpret this as ending the string, causing lots of problems.

I realized that the single quotes aren't even required. So I removed them.

```js
angular.fromJson(<%= Katello::RepositoryTypeManager.enabled_repository_types.values.to_json.html_safe %>)
```

I also considered using template strings - the (`) character. Both approaches seem to work fine; I'm still somewhat on the fence which is best.

1. Create a user whose language is 'fr'  - Administer > Users > create new > "Language"
2. Impersonate this user
3. Visit Angular Katello pages: Activation Keys, Lifecycle Environments, Products, Content Credentials, Sync Plans
4. Verify that the pages load

### DIFF
--- a/engines/bastion/app/views/bastion/layouts/assets.html.erb
+++ b/engines/bastion/app/views/bastion/layouts/assets.html.erb
@@ -21,17 +21,17 @@
     angular.module('Bastion').value('simpleContentAccessEnabled', <%= Organization.current.simple_content_access? if Organization.current %>);
     angular.module('Bastion').value('isManifestImported', <%= !!Organization.current&.manifest_imported?(cached: true) %>)
     angular.module('Bastion').value('foreman', tfm);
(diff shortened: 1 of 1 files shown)


ianballou (commented):
> I also considered using template strings - the (`) character. Both approaches seem to work fine; I'm still somewhat on the fence which is best.

I actually needed these backticks to get the page to load. Without it, on the activation keys page at least, I got a JS error regarding some JSON parsing issue.

jeremylenz:
@ianballou Ok, switched to using backticks. See if that's better?

ianballou (approved):
Works for me!

PR #10361, opened 2022-11-18
Fixes #35773 - Checks for empty cacert on connecting via Http Proxy

parthaa:
Given an http proxy, katello tries to add proxy's cacert while connecting to cdn/upstream. Unfortunately it checked for blank cacert incorrectly.
This PR adds a `.present?` to check for blank certs  (`""` is considered true while `"".present?` is false in ruby.) 

Check out the steps listed in the redmine
- Setup a proxy. My ruby one which runs proxy on port 8777
```ruby
#!/usr/bin/ruby
require 'webrick'
require 'webrick/httpproxy'

proxy = WEBrick::HTTPProxyServer.new Port: 8777

trap 'INT'  do proxy.shutdown end
trap 'TERM' do proxy.shutdown end

proxy.start
```

- Setup an http proxy in ui with url set to `http://localhost:8777` and make sure you can connect via test connection
- Mark this http proxy as the default http proxy in settings
- Try any of the following operations

1. Enable repositories
2. Check out Subscriptions
3. Sync stuff

Before PR:
```Katello::HttpErrors::BadRequest: no certificate or crl found```

After PR:
No such errors.

### DIFF
--- a/test/lib/resources/cdn_test.rb
+++ b/test/lib/resources/cdn_test.rb
@@ -17,6 +17,17 @@ class CdnResourceTest < ActiveSupport::TestCase
     assert_nil cdn_resource.http_downloader.ssl_version
   end
 
+  def test_http_proxy_no_cacert
+    proxy = FactoryBot.create(:http_proxy, :url => 'http://foo.com:1000',
+                              :username => 'admin',
+                              :password => 'password',
+                              :cacert => "")
+    Katello::Resources::CDN::CdnResource.any_instance.stubs(:proxy).returns(proxy)
+    OpenSSL::X509::Store.any_instance.stubs(:add_file)
+    Foreman::Util.expects(:add_ca_bundle_to_store).never
+    Katello::Resources::CDN::CdnResource.new('http://foo.com', ssl_ca_file: "lol")
+  end
+
   def test_http_downloader_bad_param
     Setting[:cdn_ssl_version] = 'Foo'
     assert_raise RuntimeError do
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Review by ianballou — COMMENTED** [review]
Tested and works well!

Do you think we should add a test for checking that the cacert is included appropriately?  I can see 3 places for potential tests: 
- https://github.com/katello/katello/blob/master/test/lib/resources/candlepin_test.rb#L28
  - To test `UpstreamCandlepinResource.resource`
- https://github.com/Katello/katello/blob/master/test/lib/resources/cdn_test.rb#L19
  - To test `CdnResource.initialize`
- https://github.com/Katello/katello/blob/master/test/services/katello/pulp3/repository/yum/yum_test.rb
  - To test `append_proxy_cacert`

**Review by ianballou — APPROVED** [review]
Looks good!

File: test/lib/resources/cdn_test.rb [hunk 1]
```diff
@@ -17,6 +17,17 @@ def test_http_downloader_no_version
     assert_nil cdn_resource.http_downloader.ssl_version
   end
 
+  def test_http_proxy_no_cacert
+    proxy = FactoryBot.create(:http_proxy, :url => 'http://foo.com:1000',
+                              :username => 'admin',
+                              :password => 'password',
+                              :cacert => "")
+    Katello::Resources::CDN::CdnResource.any_instance.stubs(:proxy).returns(proxy)
+    OpenSSL::X509::Store.any_instance.stubs(:add_file)
```
parthaa, line 26:
```suggestion
    OpenSSL::X509::Store.any_instance.expects(:add_file)
```

ianballou (approved):

PR #10363, opened 2022-11-22
Fixes #35640 - fix broken hammer host subscription content-override

lfu:
Fix hammer to override repository sets correctly.

```
hammer host subscription content-override --host-id 2 --content-label Default_Organization_fake-product_fake-repo --override-name 'enabled' --value=1
hammer host subscription content-override --host-id 2 --content-label Default_Organization_fake-product_fake-repo --override-name 'enabled' --value=0
hammer host subscription content-override --host-id 2 --content-label Default_Organization_fake-product_fake-repo --remove
```
Check Host - Content - Repository sets from UI.

Before
All repos were overridden to disabled when trying to disable/enable one repo.
All repos were reset too default when trying to reset one repo.

After 
Specified repo will be enabled/disabled/reset.

### DIFF
--- a/app/controllers/katello/api/v2/host_subscriptions_controller.rb
+++ b/app/controllers/katello/api/v2/host_subscriptions_controller.rb
@@ -232,7 +232,7 @@ module Katello
     end
 
     def find_content_overrides
-      if params[:content_overrides_search]
+      if params.dig(:content_overrides_search, :search).present?
         content_labels = ::Katello::Content.joins(:product_contents)
                             .where("#{Katello::ProductContent.table_name}.product_id": @host.organization.products.subscribable.enabled)
                             .search_for(params[:content_overrides_search][:search])


app/controllers/katello/api/v2/host_subscriptions_controller.rb
```diff
@@ -232,7 +232,7 @@ def action_permission
     end
 
     def find_content_overrides
-      if params[:content_overrides_search]
+      if params.dig(:content_overrides_search, :search).present?
         content_labels = ::Katello::Content.joins(:product_contents)
                             .where("#{Katello::ProductContent.table_name}.product_id": @host.organization.products.subscribable.enabled)
                             .search_for(params[:content_overrides_search][:search])
```
lfu, line 238: When calling hammer to enable/disable a repo, `params` has `"content_overrides_search" => {}`.
When calling hammer to reset a repo, `params` has `"content_overrides_search" => {"remove"=>true}`.

For above two cases, `params[:content_overrides_search]` exists but `params[:content_overrides_search][:search]` is `nil` and this search returns back all repos. That is why all repos were affected by the hammer commands.

chris1984 (approved):
Tested and looks good, thanks @lfu

PR #10364, opened 2022-11-23
Fixes #35782 - Add Gregorian translations in Katello

chris1984:


chris1984:
@sjha4 when you get time, can this get a review, this is from my exploring :)

ekohl (changes requested):
@chris1984 this should be done in master first and then cherry picked. Otherwise it'll be gone in the next release (4.8). Dirk also pointed out it's Georgian. That was my bad.

chris1984:
> @chris1984 this should be done in master first and then cherry picked. Otherwise it'll be gone in the next release (4.8). Dirk also pointed out it's Georgian. That was my bad.

@ekohl Ok I will close this and open it in master then cherry-pick it

PR #10365, opened 2022-11-23
Fixes #35782 - Add Georgian translations in Katello

chris1984:
@sjha4 here is the updated pr.


sjha4 (approved):

locale/bn/katello.po
```diff
@@ -8,7 +8,7 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: katello 2.5.0\n"
+"Project-Id-Version: katello 2.4.0-RC1\n"
```
ekohl, line 11: I'd recommend syncing `locale/Makefile` from https://github.com/theforeman/foreman_plugin_template/pull/38. It pulls the version from the gemspec instead of the tag.

That it only finds 2.4.0-RC1 is interesting though. It's probably the last version that was tagged while in master. After that it started to be tagged in branches.

PR #10366, opened 2022-11-23
Fixes #35783 - Remove old Pulp 2 docker port setting

ianballou:
Removes the old `pulp_docker_registry_port` setting from Pulp 2. Also cleans up some of the related Pulp 2 methods.

This should not affect any of the Pulp 3 container workflows.

1) Edit the `pulp_docker_registry_port` setting
2) Patch in this PR and run the DB migrations
3) Check that the setting is gone
4) Check that there are no traces of the setting in the database
5) Check that there are no traces of the setting in the code
6) Double check that container smart proxy syncing still works (and thus repo syncing too)
  -> And `podman pull` from a client


app/services/katello/pulp/repository/docker.rb
```diff
@@ -24,26 +24,6 @@ def generate_primary_importer
...
-        def generate_mirror_importer
-          config = {
-            feed: docker_registry_host,
-            upstream_name: repo.container_repository_name,
-            enable_v1: false
-          }
-          importer_class.new(config.merge(mirror_importer_connection_options))
-        end
-
         def generate_distributors
```
sjha4, line 27: Can we delete this entire file. I don't see why we need this.

PR #10370, opened 2022-12-01
Registration - use registration_url for RHSM configuration

stejskalleos:
Follow up to https://github.com/Katello/katello/pull/10315, if Smart Proxy Registration module 
is configured to load balancing (`registration_url` is set),
use this value for RHSM configuration when registering host, see [global registration](https://github.com/theforeman/foreman/blob/develop/app/views/unattended/provisioning_templates/registration/global_registration.erb#L186-L189)

* Configure smart proxy to load balancing
* Register host
* Check that correct RHSM URLs are set and it works

### DIFF
--- a/app/models/katello/concerns/smart_proxy_extensions.rb
+++ b/app/models/katello/concerns/smart_proxy_extensions.rb
@@ -456,14 +456,16 @@ module Katello
       end
 
       def rhsm_url
+        proxy_url = setting('Registration', 'registration_url') || url
+
         # Since Foreman 3.1 this setting is set
         if (rhsm_url = setting(SmartProxy::PULP3_FEATURE, 'rhsm_url'))
           URI(rhsm_url)
           # Compatibility fall back
         elsif pulp_primary?
-          URI("https://#{URI.parse(url).host}/rhsm")
+          URI("https://#{URI.parse(proxy_url).host}/rhsm")
         elsif pulp_mirror?
-          URI("https://#{URI.parse(url).host}:8443/rhsm")
+          URI("https://#{URI.parse(proxy_url).host}:8443/rhsm")
         end
       end
 
--- a/test/factories/smart_proxy_factory.rb
+++ b/test/factories/smart_proxy_factory.rb
@@ -40,5 +40,15 @@ FactoryBot.modify do
         smart_proxy_feature.settings[:mirror] = true
       end
     end
+
+    trait :with_load_balancer do
+      after(:build) do |proxy, _evaluator|
+        feature = Feature.find_or_create_by(:name => 'Registration')
+        proxy.features << feature unless proxy.features.include?(feature)
+        smart_proxy_feature = proxy.smart_proxy_features.find { |spf| spf.feature_id == feature.id }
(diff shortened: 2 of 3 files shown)


stejskalleos:
@ehelms

ehelms:
Seems pretty straight forward and would solve the problem to me. @ekohl / @evgeni since we discussed this amongst ourselves, I'd like your thoughts.

app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -456,14 +456,16 @@ def smart_proxy_service
...
         # Since Foreman 3.1 this setting is set
         if (rhsm_url = setting(SmartProxy::PULP3_FEATURE, 'rhsm_url'))
           URI(rhsm_url)
           # Compatibility fall back
         elsif pulp_primary?
-          URI("https://#{URI.parse(url).host}/rhsm")
+          URI("https://#{URI.parse(proxy_url).host}/rhsm")
         elsif pulp_mirror?
-          URI("https://#{URI.parse(url).host}:8443/rhsm")
+          URI("https://#{URI.parse(proxy_url).host}:8443/rhsm")
```
ekohl, line 468: These code paths should be unused since Foreman 3.1.

ekohl (changes requested):
This is not doing what you think it's doing. You're changing a dead code path and won't work without changes to the installer to STOP setting the rhsm_url.

Then comes the point: when do you set rhsm_url and when don't you. I'd say this leads to more problems than it solves so :-1: from my point of view.

app/models/katello/concerns/smart_proxy_extensions.rb
ehelms: I grok from that only the initial `if` statement is ever triggered now, and is really not needed anymore. That, with this idea this function could be shortened to:

```
def rhsm_url

    url = setting('Registration', 'registration_url') || setting(SmartProxy::PULP3_FEATURE, 'rhsm_url')
    URI(url)

end
```

ekohl: Correct that only the initial `if` is triggered. Still, big :-1: from my side on this.

stejskalleos: @ekohl so what would be the proper fix  / update?

ehelms:
> Then comes the point: when do you set rhsm_url and when don't you. I'd say this leads to more problems than it solves so  from my point of view.

Can you expand on this? Sounds like the devil is in the details, but we need the details to understand why this won't work.

ekohl:
The short answer is that we have a mechanism to declare what the RHSM URL is. I think we should really use that end-to-end. This PR effectively gets rid of the mechanism, which I think is a step back. It's something we should solve in our configuration layer (Puppet/installer in this case), not on the application layer. There we already have a registration URL setting - we could derive the value there and put it statically in the right Smart Proxy's YAML file.

stejskalleos:
Okay then, I'm closing the PR and let's continue in https://github.com/theforeman/puppet-foreman_proxy_content/pull/432

PR #10371, opened 2022-12-01
Fixes #35810 - Use 'search' url param instead of 'host_ids' for job wizard

jeremylenz:

Change the 'via customized remote execution' options in the new host details page to use the `search` url param instead of `host_ids`. This will make them compatible with the new job wizard.

Perform any customized REX action on the new host details page
ensure that the host search is filled in for you.  (see the BZ https://bugzilla.redhat.com/show_bug.cgi?id=2149990 for more details)


chris1984:
Code looks sane, since it's just flipping url params, will start to test and ack based on that.

chris1984 (approved):
Tested and worked great, I installed a package on my client and in the rex job, I was able to see the host search correctly.

PR #10373, opened 2022-12-02
Fixes #35795,35213 - fix Repo Sets pagination, sorting, filtering

jeremylenz:

This PR addresses 2 issues with the RepositorySetsController#index method:

1. When sorting by `enabled_by_default`, the custom sort method was applied only to the single page of results, not to the entire result list before pagination.
2. When filtering by `enabled_by_default`, the results were paginated first, then filtered. Rather, the correct way is to  filter first, then paginate.

New host details page > Content > Repository sets

Sort by "Enabled" column
Filter by Enabled / Disabled / Overridden

Verify that sorting is correct:
1. Enabled (overridden)
2. Enabled
3. Disabled
4. Disabled (overridden)

Verify that filtering pagination is correct (see description in redmine)


app/controllers/katello/api/v2/repository_sets_controller.rb
```diff
@@ -236,18 +241,24 @@ def sort_score(pc) # sort order for enabled
...
+        status: params[:status])
+      sorted_pcps = if params[:sort_by] == 'enabled_by_default' && params[:sort_order] == 'desc'
+                      product_content_finder.sort { |pca, pcb| sort_score(pca) <=> sort_score(pcb) }.reverse!
+                    elsif params[:sort_by] == 'enabled_by_default'
+                      product_content_finder.sort { |pca, pcb| sort_score(pca) <=> sort_score(pcb) }
+                    else
+                      product_content_finder
+                    end
+      sort_order = sorted_pcps.map(&:id)
+      unsorted_relation.reorder(Arel.sql("array_position('{#{sort_order.join(',')}}'::int[], #{Katello::ProductContent.table_name}.id)"))
```
jeremylenz, line 261: This only worked once I realized that I needed to call `reorder` instead of `order`, since that eliminates a default `ORDER BY` that was causing the custom sort to be ignored.

ianballou: I'm a bit  confused about how this is working. `array_position` is supposed to return a single position of an element found in an array.  But here, instead of an element, a table's ID field is being passed in as the second argument.

ianballou (commented):
It's working as intended! One comment to clear up some confusion:

app/controllers/katello/api/v2/repository_sets_controller.rb
jeremylenz: `{#{sort_order.join(',')}}'::int[]` <-- So this creates a sorted array of product_content IDs, sorted in the way we want them sorted. The second argument is saying, "find where this ID is in that array, and then sort by that."

jeremylenz: So for example, it'll make an array of IDs like
```
[8142, 8132, 726]
```
and then the SQL will be like
```sql
SELECT blah FROM table ORDER BY array_position('{#{sort_order.join(',')}}'::int[], #{Katello::ProductContent.table_name}.id)
```

which evaluates to something like
```sql
SELECT blah FROM table ORDER BY array_position([8142, 8132, 726], katello_product_contents.id)
```

jeremylenz: > Instead, would it be possible to ensure that `relation` is actually a relation when it's sent to `scoped_search`? Then, the ordering can be applied as usual. It's a little odd that `relation` isn't actually a relation by the time it is passed into `scoped_search` anyway (right?).

`relation` is always a relation. I made sure of this, since that's what `scoped_search` needs. Here's the journey it goes through:

1. `index_relation` is a relation at the end of the method. It may or may not continue onto `index_relation_with_consumable_overrides`. If it does,
2. During `index_relation_with_consumable_overrides`, it makes `ProductContentFinder`s to adorn the items with the necessary override data. It then converts it back to a relation before exiting the method.
3. It's then passed into `scoped_search` as a relation, as required.

The reason all this is necessary is because

1. Content overrides are associated with hosts or activation keys, but this controller is for product contents, which is _very_ far away from those in terms of database tables.
2. Also, content overrides are a regular Ruby class, not a model. This is because they're just a snippet of Candlepin data.
3. Because of the above, there isn't a viable join we can do to "sort normally." But we already have the ProductContentFinder and ProductContentPresenter which addresses these two issues, so it works out.

ianballou: I'm convinced! Thanks for the details.

ianballou (approved):

PR #10377, opened 2022-12-05
Fixes #35046 - pluralize and translate errata tooltip

jeremylenz:

1. Translate the pie chart tooltips from the errata overview card
2. Run the counts through our simple pluralizer function to prevent incorrect pluralization such as "1 bug fixes"

I tried to use `react-intl` but we apparently don't have access to the imperative API documented [here](https://formatjs.io/docs/react-intl/api). So I used the simpler `pluralize` helper function instead.

Also I realized we should probably be translating these tooltips as well.

Change https://github.com/Katello/katello/blob/4950081967a99de4b68825cbe86ea8845334b155/webpack/components/extensions/HostDetails/Cards/ErrataOverviewCard.js#L30 to

```js
const errataBug = 1;
```

and view the tooltip displayed on mouse hover.


chris1984 (approved):
ACK works well, thanks @jeremylenz 

Singular:
errata-pr

Plural:
errata-normal

jeremylenz:
Sure, but the change isn't in the chart legend, it's in the tooltip that appears when you hover over the pie chart :wink:

chris1984:
errata-pr

Tooltip is working as intended :)

chris1984 (approved):

PR #10379, opened 2022-12-06
Fixes #35829 - Mirror complete sync policy no longer allowed with ignored content types

sjha4:
1. Adds a migration to clear ignored content types when repository mirroring policy is mirror_complete
2. Ignores skip_types for all repo mirrors which default to mirror_complete as mirroring policy.
1. Create a repo with Ignore SRPMs checked and Mirroring Policy: Complete mirroring
2. Sync > You should hit the error.
3. If you have a capsule in the environment, try syncing the capsule and you'll see the error for any repos with Ignore SRPMs checked regardless of mirroring policy on the repo.

Switch to PR and run db:migrate.
You shouldn't see the error anymore on sync as data gets cleaned.
You should see a validation error on create/update when setting Ignore SRPM with Complete mirroring policy.

### DIFF
--- a/app/models/katello/root_repository.rb
+++ b/app/models/katello/root_repository.rb
@@ -227,6 +227,8 @@ module Katello
         errors.add(:ignorable_content, N_("Invalid value specified for ignorable content."))
       elsif ignorable_content.any? { |item| !IGNORABLE_CONTENT_UNIT_TYPES.include?(item) }
         errors.add(:ignorable_content, N_("Invalid value specified for ignorable content. Permissible values %s") % IGNORABLE_CONTENT_UNIT_TYPES.join(","))
+      elsif self.mirroring_policy == MIRRORING_POLICY_COMPLETE
+        errors.add(:ignorable_content, N_("Ignore SRPMs can not be set in combination with 'Complete Mirroring' mirroring policy."))
       end
     end
 
--- a/app/services/katello/pulp3/repository.rb
+++ b/app/services/katello/pulp3/repository.rb
@@ -200,7 +200,7 @@ module Katello
 
       def sync_url_params(_sync_options)
         params = {remote: repo.remote_href, mirror: repo.root.mirroring_policy == Katello::RootRepository::MIRRORING_POLICY_CONTENT}
-        params[:skip_types] = skip_types if skip_types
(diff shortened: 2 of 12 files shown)


chris1984 (commented):
Code looks sane, starting to test on my devel box with a smart proxy

chris1984 (approved):
ACK works well, hit all the issues. After applying the patch, I was able to sync the repo, sync the smart proxy and when testing the repo create I got the following error:
```bash
Validation failed: Ignorable content Ignore SRPMs can not be set in combination with 'Complete Mirroring' mirroring policy.
```

@sjha4 are you able to make a cherry-pick pr to 4.7 directly?

app/models/katello/root_repository.rb
```diff
@@ -227,6 +227,8 @@ def ensure_valid_ignorable_content
         errors.add(:ignorable_content, N_("Invalid value specified for ignorable content."))
       elsif ignorable_content.any? { |item| !IGNORABLE_CONTENT_UNIT_TYPES.include?(item) }
         errors.add(:ignorable_content, N_("Invalid value specified for ignorable content. Permissible values %s") % IGNORABLE_CONTENT_UNIT_TYPES.join(","))
+      elsif self.mirroring_policy == MIRRORING_POLICY_COMPLETE
+        errors.add(:ignorable_content, N_("Ignore SRPMs can not be set in combination with 'Complete Mirroring' mirroring policy."))
```
sbernhard, line 231: Currently, only SRPM is a ignoreable content type. But, this could change in the future. Therefore, I would suggest to write the error message as it was in
```("Invalid value specified for ignorable content. Permissible values %s") % IGNORABLE_CONTENT_UNIT_TYPES.join(","))``` 

errors.add(:ignorable_content, N_("Ignore %s can not be set in combination with 'Complete Mirroring' mirroring policy.") % IGNORABLE_CONTENT_UNIT_TYPES.join(","))

chris1984:
@sbernhard We can do a follow up pr, this had to get in for a downstream bz and to not block Katello 4.7 GA. We can bring the updated pr/fix into 4.7.1

sbernhard:
> @sbernhard We can do a follow up pr, this had to get in for a downstream bz and to not block Katello 4.7 GA. We can bring the updated pr/fix into 4.7.1

Done with https://github.com/Katello/katello/pull/10386

PR #10380, opened 2022-12-07
Fixes #35834 - select content source before LCE and CV 

lfu:
In Create Host, show Content source field before both Lifecycle environment and Content view fields.

When `Edit Host`, all Content source, Lifecycle environment and Content view fields should be disabled to honor `Change content source` task which is available in the kebab dropdown list.

Lifecycle environment, Content view and Content source are all saved into a host's content facet.
Content facet would not be created if either Lifecycle environment or Content view is blank due to [this checking](https://github.com/Katello/katello/blob/4950081967a99de4b68825cbe86ea8845334b155/app/models/katello/concerns/content_facet_host_extensions.rb#L62).

[Decision](https://docs.google.com/document/d/18lB3EPatXtCi-GZRNUxT7ujr_LniiInp0Tk7G30RR7U/edit) made during a meeting.

All Hosts - Create Host
Select Content source and all other required fields in all tabs

Before
With either Lifecycle environment or Content view blank, selected content source is not persistent.

After
Selected Content source field is preserved.

### DIFF
--- a/app/assets/javascripts/katello/hosts/host_and_hostgroup_edit.js
+++ b/app/assets/javascripts/katello/hosts/host_and_hostgroup_edit.js
@@ -13,6 +13,13 @@ $(document).on('ContentLoad', function(){
     KT.hosts.set_media_selection_bindings();
 });
 
+KT.hosts.contentSourceChanged = function() {
+  $("#hostgroup_lifecycle_environment_id").val("");
+  $("#host_lifecycle_environment_id").val("");
+  KT.hosts.fetchEnvironments();
+  KT.hosts.environmentChanged();
+};
+
 KT.hosts.environmentChanged = function() {
   // if we don't save the currently selected view it's likely
   // it will be undefined in toggle_installation_medium due to the CV dropdown reload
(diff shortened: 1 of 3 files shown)


lfu:
Content source is not displayed when either environment or content view is not set.
Screenshot from 2022-12-13 13-24-31

Content source is displayed when both environment and content view are selected.
Screenshot from 2022-12-13 13-24-57

jeremylenz (commented):
On the Change Content Source page, selection of content source is required first, and then you can choose an LCE and content view.  This makes sense to me because Smart Proxies with Content (capsules) are synced to a particular lifecycle environment, and that LCE determines which content views are available for selection.  But on this page it's the opposite, and I'm afraid that might not be correct.

cc @stejskalleos for thoughts

stejskalleos:
Jeremy is correct, first you should select capsule, and then select views & environments, but only those that are available on the capsule.
Otherwise users could select view and environment that are not available on capsule and that will cause issues later when host will start consuming data from the capsule.

stejskalleos:
Linking https://github.com/Katello/katello/pull/10297 with the change of field order

jeremylenz:
Now that https://github.com/Katello/katello/pull/10457 is merged with the improvements to the new CV selector, maybe we can use it here as well!

jeremylenz (commented):
Thanks for getting back to this @lfu!

image

When editing a host, this button looks out of place. I would suggest a link-style (not button-style), a simple "Change" link below the "Content Source" text.

Also, the disabled fields should reflect the host's current values. But for me, they're just blank.

When creating a new host, are the LCE and CV options reflecting the currently selected content source? (I couldn't test this because I don't have a smart proxy setup)

jeremylenz:
When I click the Change content source button, I'm getting this error:

image

lfu:
> When creating a new host, are the LCE and CV options reflecting the currently selected content source? (I couldn't test this because I don't have a smart proxy setup)

LCE options does not reflect the selected content source with current code.
CV options depends on the selected LCE.

jeremylenz (commented):
The new link to Change content source looks better, thanks!

Still seeing some issues with Edit. When I load the edit page, the form should reflect all 3 of the host or hostgroup's current values -- Content source, Lifecycle environment, Content view.

Host edit: Current Content source and Lifecycle environment are not reflected.
Hostgroup edit: Current Content view is not reflected.

I think this is what the original bug was describing when it said "Selecting a content source is not persistent." It actually does get saved (I confirmed in Rails console), it just doesn't show on the Edit form.

lfu:
@jeremylenz Updated. It looks good on my screen. :)

PR #10384, opened 2022-12-08
Refs #35774 - further refactor EventDaemon::Runner start

evgeni:
this eliminates the need for `.initialize` as it moves all the code into `.start`

also uses a block for the lockfile, so no file descriptor is leaked

I mostly listened to Ewouds suggestions :)

The following should still be true:
1. The EventDaemon is properly started only once in the app
2. The EventDaemon does not try to start in Rake


app/lib/katello/event_daemon/runner.rb
```diff
@@ -54,8 +49,11 @@ def stop
...
           return unless runnable?
-          lockfile = File.open(lock_file, 'r')
-          begin
+
+          FileUtils.mkdir_p(tmp_dir)
+          FileUtils.touch(lock_file)
+
+          File.open(lock_file, 'r') do |lockfile|
             lockfile.flock(File::LOCK_EX)
             return if started? # ensure it wasn't started while we waited for the lock
```
evgeni, line 58:
```
[2022-12-08T09:55:52.120Z] /home/jenkins/workspace/katello-pr-test/app/lib/katello/event_daemon/runner.rb:58:13: W: Lint/NonLocalExitFromIterator: Non-local exit from iterator, without return value. next, break, Array#find, Array#any?, etc. is preferred.
```

I don't see no iterator here?

ehelms: Suppose you could just return nil explicitly.

evgeni: *that* is what it complains about? ACAB!

ehelms (approved):

chris1984 (approved):
Code looks sane, tested a few things to make sure nothing broke and `/tmp` was still created ACK

PR #10387, opened 2022-12-08
Fixes #35818 - Fix repository sets NilClass error when all items selected

jeremylenz:

In `Api::V2::HostSubscriptionsController#find_content_overrides`, the value from `params[:content_overrides_search][:search]` was checked using `.present?`. Since Rails' `#present?` considers an empty string to be not "present," this was breaking the Select All functionality on the new host details page Repository Sets tab. To fix it, we switch to checking explicitly for `nil` instead.

This shouldn't interfere with the fix in https://github.com/Katello/katello/pull/10363.

New host details > Repository Sets tab:

1. Select All repository sets with the checkbox -> Override to Disabled / Enabled / Reset to default. Make sure it completes with no errors.
2. Make sure other repo set operations continue to work (non-select-all)
3. Make sure the hammer fix mentioned above is still working.

### DIFF
--- a/app/controllers/katello/api/v2/host_subscriptions_controller.rb
+++ b/app/controllers/katello/api/v2/host_subscriptions_controller.rb
@@ -176,7 +176,6 @@ module Katello
       content_override_values = @content_overrides.map do |override_params|
         validate_content_overrides_enabled(override_params)
       end
-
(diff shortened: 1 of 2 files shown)


jeremylenz:
[test katello]

lfu (approved):
LGTM 👍🏻 
Both webUI and hammer worked as expected.

PR #10389, opened 2022-12-12
Fixes #35854 - Columns should be empty if there's no data

pkoprda:
On host index page `Registered` and `Last checkin` columns should show up as empty if there's no data.

before:
registered-last-checkin-before

after:
registered-last-checkin-after

In Column selector check  `Registered` and `Last checkin` column.

### DIFF
--- a/app/helpers/katello/hosts_and_hostgroups_helper.rb
+++ b/app/helpers/katello/hosts_and_hostgroups_helper.rb
@@ -297,13 +297,13 @@ module Katello
     end
 
     def host_registered_time(host)
-      return _('Never registered') unless host.subscription_facet_attributes&.registered_at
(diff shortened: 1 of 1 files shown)


jeremylenz (commented):
Hi @pkoprda 

Pretty straightforward change, but can you add some context behind it? Was it requested by someone? Is there a particular problem it's solving?

pkoprda:
Hi @jeremylenz 

It's the same way as we want to achieve here: https://github.com/theforeman/foreman/pull/9535.
And I totally agree what @adamruzicka already mentioned.

jeremylenz (approved):
Thanks for the explanations @pkoprda and @adamruzicka 

LGTM :+1:

PR #10390, opened 2022-12-12
Fixes #35857 - don't assume single_content_view etc. is present in RABL templates

jeremylenz:

In our RABL templates, `single_content_view` and `single_lifecycle_environment` were being relied on to be non-nil. This caused some NilClass errors when a host had a cleared-out content facet.

hopefully this will fix nightly

ermm.. run the pipeline I guess?

### DIFF
--- a/app/views/katello/api/v2/content_facet/show.json.rabl
+++ b/app/views/katello/api/v2/content_facet/show.json.rabl
@@ -3,20 +3,22 @@ child :content_facet => :content_facet_attributes do
 
   node do |content_facet|
     content_facet.single_content_view # just so Rails gives a warning and we remember to fix this
-    version = content_facet.content_view_environments.first.content_view_version
-    {
-      :content_view_version => version.version,
-      :content_view_version_id => version.id,
-      :content_view_version_latest => version.latest?
-    }
+    version = content_facet.content_view_environments.first&.content_view_version
+    if version.present?
+      {
+        :content_view_version => version.version,
+        :content_view_version_id => version.id,
+        :content_view_version_latest => version.latest?
+      }
+    end
   end
 
   node :content_view_default? do |content_facet|
-    content_facet.single_content_view.default?
+    content_facet.single_content_view&.default? || false
   end
 
   node :lifecycle_environment_library? do |content_facet|
-    content_facet.single_lifecycle_environment.library?
+    content_facet.single_lifecycle_environment&.library? || false
   end
 
   node :katello_agent_installed do |content_facet|
--- a/test/controllers/api/v2/hosts_controller_test.rb
+++ b/test/controllers/api/v2/hosts_controller_test.rb
@@ -41,6 +41,13 @@ class Api::V2::HostsControllerTest < ActionController::TestCase
     host_index_and_show(host)
   end
 
+  def test_no_content_view_environments
+    host = FactoryBot.create(:host, :with_content, :with_subscription)
+    assert_empty host.content_facet.content_view_environments
+
+    host_index_and_show(host)
+  end
+
   def test_with_subscriptions
     host = FactoryBot.create(:host, :with_subscription)
     host_index_and_show(host)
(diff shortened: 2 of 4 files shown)


app/views/katello/api/v2/content_facet/show.json.rabl
```diff
@@ -4,19 +4,21 @@ child :content_facet => :content_facet_attributes do
   node do |content_facet|
     content_facet.single_content_view # just so Rails gives a warning and we remember to fix this
     version = content_facet.content_view_environments.first.content_view_version
```
ekohl, line 6: I'm guessing `content_facet.content_view_environments.first` can be `nil` too. Not sure if `content_facet.content_view_environments` is always a list of even that can be `nil`.
```suggestion
    version = content_facet.content_view_environments.first&.content_view_version
```

ekohl (commented):
Is `nil.present?` a Rails enhancement? In plain Ruby that doesn't work. There you can just use `if nil`.

jeremylenz:
Yes, `.present?` is from Rails. But I think that should be fine, right?

ekohl:
Yes, I'm just not used to Rails' enhancements since I often write in vanilla Ruby.

app/views/katello/api/v2/content_facet/show.json.rabl
```diff
@@ -3,20 +3,22 @@ child :content_facet => :content_facet_attributes do
...
+        :content_view_version => version.version,
+        :content_view_version_id => version.id,
+        :content_view_version_latest => version.latest?
+      }
+    end
   end
 
   node :content_view_default? do |content_facet|
-    content_facet.single_content_view.default?
+    !!content_facet.single_content_view&.default?
```
parthaa, line 17: This seems a very javascripty thing and not ruby?
Why is this needed. The code prior to your single-cv change
`content_facet.content_view.default?` seems to think content facet has to have a CV

app/views/katello/api/v2/content_facet/show.json.rabl
```diff
@@ -3,20 +3,22 @@ child :content_facet => :content_facet_attributes do
...
+        :content_view_version => version.version,
+        :content_view_version_id => version.id,
+        :content_view_version_latest => version.latest?
+      }
+    end
   end
 
   node :content_view_default? do |content_facet|
-    content_facet.single_content_view.default?
+    ::Foreman::Cast.to_bool(content_facet.single_content_view&.default?)
```
ekohl, line 17: I'd keep it simple:
```suggestion
    content_facet.single_content_view&.default? || false
```

ekohl (changes requested):
Please squash the commits.

app/views/katello/api/v2/content_facet/show.json.rabl
jeremylenz: `content_view_default?` should return a Boolean; I wanted to ensure it's not `nil`. Switched to `::Foreman::Cast.to_bool`.

jeremylenz: ...and now  to `|| false`

ekohl:
Code looks ok, but can we please also cover this using tests so we don't regress in the future?

test/controllers/api/v2/hosts_controller_test.rb
```diff
@@ -41,6 +41,13 @@ def test_with_content
     host_index_and_show(host)
   end
 
+  def test_no_cves
```
evgeni, line 44:
```suggestion
  def test_no_content_view_environments
```

CVEs is kinda ambiguous in our context ;-)

evgeni:
LGTM besides one wording nitpick ;-)
(didn't run it, but it reads correct)

jeremylenz:
@evgeni updated :)

evgeni (approved):

parthaa (approved):
ack

PR #10391, opened 2022-12-13
Fixes #35861 - CV publish error when host tied to content facet is not found

sjha4:
1. Throw an error if the content host tied to a CV has had it's host deleted for some reason.
2. Add a task to clean orphaned content and subscription facets.
Working on it to avoid adding steps that involve delete FK constraints in the DB. 🐱 
Adding steps to reproduce on dev.
1. Create a CV and add an activation key for the CV.
2. Register a host with the activation key.
3. Go to `psql -d katello -U katello`
4. DELETE from host where id='id'
5. ALTER table to remove the FK constraints that fail.
 Follow steps here for the queries to run:  https://github.com/Katello/katello/pull/10391#issuecomment-1371230005
7. Once the host has been deleted, try publishing the CV again. You should see the error.
Screenshot from 2023-01-04 12-42-48

8. Checkout this branch and publish again.
9. Run the rails task as the error message states.
10. The orphaned content facet and subscription facet will be deleted.
11. Publish again. It should succeed.

### DIFF
--- a/app/models/katello/content_view.rb
+++ b/app/models/katello/content_view.rb
@@ -604,6 +604,7 @@ module Katello
         fail _("Import-only content views can not be published directly") if import_only? && !syncable
         check_composite_action_allowed!(organization.library)
         check_docker_repository_names!([organization.library])
+        check_orphaned_content_facets(environments: self.environments)
       end
 
       true
@@ -648,6 +649,17 @@ module Katello
         end
       end
       true
+    end
+
+    def check_orphaned_content_facets(environments: [])
+      ::Katello::Host::ContentFacet.in_content_views_and_environments(
+        content_views: [self],
+        lifecycle_environments: environments
+      ).each do |facet|
+        unless facet.host
+          fail _("Orphaned content facets for deleted hosts exist for the content view and environment. Please run rake task : katello:clean_orphaned_facets and try again!")
+        end
+      end
     end
 
     def check_remove_from_environment!(env)
--- a/app/models/katello/host/content_facet.rb
+++ b/app/models/katello/host/content_facet.rb
@@ -38,7 +38,7 @@ module Katello
 
       validates_with ::AssociationExistsValidator, attributes: [:content_source]
       validates_with Katello::Validators::GeneratedContentViewValidator
-      validates :host, :presence => true, :allow_blank => false
+      validates :host, :presence => true, :allow_blank => true
 
       attr_accessor :cves_changed
 
--- a/lib/katello/tasks/clean_orphaned_facets.rb
+++ b/lib/katello/tasks/clean_orphaned_facets.rb
@@ -1,0 +1,20 @@
+namespace :katello do
+  desc "Remove orphaned and unneeded content/subscription facets."
+  task :clean_orphaned_facets => ["environment"] do
+    User.current = User.anonymous_admin
+    remove_orphan_facets
+  end
+
+  def remove_orphan_facets
+    ::Katello::Host::ContentFacet.select { |c| c.host.nil? }&.each do |content_facet|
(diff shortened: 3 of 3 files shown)


app/controllers/katello/concerns/api/v2/registration_controller_extensions.rb
```diff
@@ -18,16 +18,6 @@ def prepare_host
...
-      def host_setup_extension
-        if params['host']['lifecycle_environment_id']
-          new_lce = KTEnvironment.readable.find(params['host']['lifecycle_environment_id'])
-          @host.content_facet.lifecycle_environment = new_lce
-          @host.update_candlepin_associations
-        end
-
-        super
-      end
-
```
jeremylenz, line 30: We were going to move this to another issue, right?

sjha4: @jeremylenz : Yes..I filed this for the registration change: https://projects.theforeman.org/issues/35874 
Will update this PR.

sjha4:
Drop all foreign key constraints:

```
ALTER TABLE fact_values DROP CONSTRAINT fact_values_host_id_fk;
ALTER TABLE nics DROP CONSTRAINT nics_host_id_fk;
ALTER TABLE host_status DROP CONSTRAINT host_status_hosts_host_id_fk;
ALTER TABLE katello_subscription_facets DROP CONSTRAINT katello_subscription_facets_host_id;
ALTER TABLE katello_content_facets DROP CONSTRAINT katello_content_facets_host_id;
ALTER TABLE katello_host_installed_packages DROP CONSTRAINT katello_host_installed_packages_host_id;
ALTER TABLE katello_host_available_module_streams DROP CONSTRAINT katello_hems_host_id_fk;
ALTER TABLE host_facets_reported_data_facets DROP CONSTRAINT fk_rails_a82ac722f1;
```

Now delete the host record:

`DELETE from hosts where id=$ID`

app/models/katello/content_view.rb
```diff
@@ -604,6 +604,7 @@ def check_ready_to_publish!(importing: false, syncable: false)
         fail _("Import-only content views can not be published directly") if import_only? && !syncable
         check_composite_action_allowed!(organization.library)
         check_docker_repository_names!([organization.library])
+        check_orphaned_content_facets(organization.library)
```
parthaa, line 607: Can we run this for all environments instead of just library ?
```ruby
        organization.kt_environments.each do |env|
             check_orphaned_content_facets(env)
        end
```

sjha4: Updated to use cv.environments.

app/models/katello/content_view.rb
```diff
@@ -650,6 +651,17 @@ def check_composite_action_allowed!(env)
       true
     end
 
+    def check_orphaned_content_facets(env)
```
parthaa, line 654: nit pick

```suggestion
    def check_orphaned_content_facets(envs)
```

or


```suggestion
    def check_orphaned_content_facets(environments: [])
```
to make it amply clear

sjha4: Updated with 2nd suggestion.

app/models/katello/content_view.rb
```diff
@@ -604,6 +604,7 @@ def check_ready_to_publish!(importing: false, syncable: false)
         fail _("Import-only content views can not be published directly") if import_only? && !syncable
         check_composite_action_allowed!(organization.library)
         check_docker_repository_names!([organization.library])
+        check_orphaned_content_facets(self.environments)
```
parthaa, line 607:
```suggestion
        check_orphaned_content_facets(environments: self.environments)
```

parthaa: will ignore otherwise :)

parthaa (approved):

PR #10392, opened 2022-12-13
Fixes #35863 - Stop using #hosts with KTEnvironments

jeremylenz:

Previously, the path from KTEnvironment to Host was

```
KTEnvironment > ContentFacet > Host
```

This is only 2 hops (e.g. there is only one model between KTEnvironment and Host.)

Now, in our new multiple environment world, it's

```
KTEnvironment > ContentViewEnvironment > ContentViewEnvironmentContentFacet > ::Host::ContentFacet > Host
```

Because there are now 4 hops, modifying the association directly will no longer work:

```rb
environment.hosts.clear # this won't work anymore
```

Instead, we must work around it by finding the host IDs similar to what we already do [here](https://github.com/jeremylenz/katello/blob/899516c90490f86df39c38fbe6052d8d82b2d66f/app/lib/actions/katello/content_view/remove.rb#L79).

I used `delete_all` rather than `destroy_all` because that's equivalent to `clear` and won't run callbacks.

Create an organization
Register a host to it
Delete the organization

Before this change:
```
Cannot modify association 'Katello::KTEnvironment#hosts' because it goes through more than one other association.
```

After this change: org should be deleted with no error.

### DIFF
--- a/app/lib/actions/katello/environment/destroy.rb
+++ b/app/lib/actions/katello/environment/destroy.rb
@@ -24,12 +24,18 @@ module Actions
             end
 
             if organization_destroy
-              env.hostgroups.clear
-              env.hosts.clear
+              delete_host_and_hostgroup_associations(environment: env)
             end
 
             plan_self
           end
+        end
+
+        def delete_host_and_hostgroup_associations(environment:)
+          environment.hostgroups.delete_all
+          host_ids = environment.hosts.ids
+          ::Katello::Host::ContentFacet.where(:host_id => host_ids).delete_all
+          ::Katello::Host::SubscriptionFacet.where(:host_id => host_ids).delete_all
         end
 
         def humanized_name
--- a/app/lib/actions/katello/content_view/remove.rb
+++ b/app/lib/actions/katello/content_view/remove.rb
@@ -71,17 +71,16 @@ module Actions
                       content_view_history_ids: cv_histories.map { |history| history.id })
 
             if organization_destroy
-              destroy_hosts_and_hostgroups(content_view: content_view)
+              destroy_host_and_hostgroup_associations(content_view: content_view)
             end
           end
         end
 
-        def destroy_hosts_and_hostgroups(content_view:)
+        def destroy_host_and_hostgroup_associations(content_view:)
           content_view.hostgroups.destroy_all
           host_ids = content_view.hosts.ids
           ::Katello::Host::ContentFacet.where(:host_id => host_ids).destroy_all
           ::Katello::Host::SubscriptionFacet.where(:host_id => host_ids).destroy_all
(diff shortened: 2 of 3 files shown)


app/lib/actions/katello/environment/destroy.rb
```diff
@@ -10,6 +10,7 @@ class Destroy < Actions::EntryAction
         end
 
         def plan(env, options = {})
+          puts options
```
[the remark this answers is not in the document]
jeremylenz: :see_no_evil:

jeremylenz:
:green_apple:

app/lib/actions/katello/environment/destroy.rb
```diff
@@ -24,14 +24,21 @@ def plan(env, options = {})
...
             plan_self
           end
         end
 
+        def delete_hosts_and_hostgroups(environment:)
+          environment.hostgroups.delete_all
+          host_ids = environment.hosts.ids
+          ::Katello::Host::ContentFacet.where(:host_id => host_ids).delete_all
+          ::Katello::Host::SubscriptionFacet.where(:host_id => host_ids).delete_all
+          ::Host::Managed.where(:id => host_ids).delete_all
```
parthaa, line 39: Don't delete the hosts. The default behavior today is if you delete the org the host remains unassigned but still visible via 'Any Organization' => 'All Hosts'

```suggestion
```

jeremylenz: The previous code was 

```rb
env.hosts.clear
```

[This](https://apidock.com/rails/ActiveRecord/Associations/CollectionProxy/clear) says `clear` is equivalent to `delete_all`. Wouldn't that also delete the hosts?

parthaa: They may be equivalent but remember that env to host association looked lie this
```ruby
 has_many :hosts,      :class_name => "::Host::Managed", :through => :content_facets,
                          :inverse_of => :lifecycle_environment
```

The `env.hosts.clear` would delete the associated content facets not the host itself.

jeremylenz: `env.hosts.clear` is equivalent to `env.hosts.delete_all` which would indeed delete the hosts. I'm not understanding what you're saying here..

jeremylenz: TIL that `delete_all` and `clear` don't delete the named records if they're called at the end of a chain. wow.

Updated to not call `delete_all` for hosts (also fixed it in the other location)

jeremylenz:
Updated; also renamed the methods to be more accurate

parthaa (approved):
ACK

PR #10395, opened 2022-12-16
Fixes #35878 - CV and LCE name show empty strings in RepoSets banner

jeremylenz:

When you set the toggle to 'Limit to environment', the Repository Sets tab has a banner

```
Showing only repositories in the "My CV" content view and "dev" environment.
```

However, since the multi-CV work it has said

```
Showing only repositories in the "" content view and "" environment.
```

This is because Rails no longer provides the methods `content_view_name` and `lifecycle_environment_name` on content facets.

This fixes that.

When we start updating the UI for multi-CV, we'll have to change this text again. But I didn't want to do that prematurely.

put a host in a non-library CV/LCE
view the Repo Sets tab


sjha4 (approved):
Before PR:
Screenshot from 2022-12-16 15-04-23
After PR:
Screenshot from 2022-12-16 15-06-05

PR #10399, opened 2022-12-22
Fixes #35232 - Fix repo sets empty state when Limit to Environment selected

jeremylenz:

Add a call-to-action for the "Limit to Environment" repo sets empty state.
image

I first tried just adding the toggle group states to `activeFilters` and `defaultFilters`, but it didn't work well. It feels better to track them separately.

1. Assign your host to an empty content view
2. New host details page > Content > Repository sets -- see the new empty state and verify that the link works

### DIFF
--- a/webpack/components/extensions/HostDetails/Tabs/RepositorySetsTab/RepositorySetsTab.js
+++ b/webpack/components/extensions/HostDetails/Tabs/RepositorySetsTab/RepositorySetsTab.js
@@ -15,6 +15,7 @@ import {
   useDispatch,
   useSelector,
 } from 'react-redux';
+import { isEqual } from 'lodash';
 
 import {
   ActionList,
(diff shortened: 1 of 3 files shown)


webpack/components/extensions/HostDetails/Tabs/RepositorySetsTab/RepositorySetsTab.js
```diff
@@ -280,9 +284,17 @@ const RepositorySetsTab = () => {
...
+  const isFiltering = activeFilters?.length &&
+    !isEqual(new Set(activeFilters), new Set(defaultFilters));
+  const emptyContent = (results && results.length === 0) && !searchQuery && !isFiltering;
+  const showPrimaryAction = (toggleGroupState === SHOW_ALL) && emptyContent;
+  const showSecondaryAction = showPrimaryAction;
+
   const resetFilters = () => {
     setStatusSelected(STATUS_LABEL);
-    setToggleGroupState(defaultToggleGroupState);
+    if (emptyContent) setToggleGroupState(SHOW_ALL);
```
lfu, line 297: Would this method `resetFilters()` ever be called when there is some content with Limit to Environment?

lfu (approved):
Nice work!
Other than a minor question, LGTM 👍🏻 
Tests all looked good.

webpack/components/extensions/HostDetails/Tabs/RepositorySetsTab/RepositorySetsTab.js
jeremylenz: No, I think it's only called when the table is empty.

PR #10401, opened 2023-01-05
Fixes #35763 - Switch to standard PF4 search input in Katello

lfu:
Katello is switching to Foreman's standard PF4 search input.

Depends on https://github.com/theforeman/foreman/pull/9571

Check all the search field work well in new host details pages and content view components.
Verify this PR fixes https://projects.theforeman.org/issues/35988.


jeremylenz (commented):
Thanks @lfu 

First impression is that this looks great! More comments coming soon, but for now I'll just say I love that we were able to remove the searchDelay and autoSearch scopes from all the tests.

Now that we don't use those settings anymore, I think we should also remove the `autosearch_while_typing` and `autosearch_delay` settings themselves.  (it's in `plugin.rb`.)

lfu:
> Now that we don't use those settings anymore, I think we should also remove the `autosearch_while_typing` and `autosearch_delay` settings themselves. (it's in `plugin.rb`.)

I have not deleted anything since the Katello search field is being used by some other components.

jeremylenz:
> I have not deleted anything since the Katello search field is being used by some other components.

Where is it still used? Shouldn't we use Foreman's everywhere?

lfu:
> Where is it still used? Shouldn't we use Foreman's everywhere?

SubscriptionsToolbar and redhatRepositoriesPage.
Should they be changed as well? Thought this change is only for new host details pages.

jeremylenz:
> SubscriptionsToolbar and redhatRepositoriesPage.
> Should they be changed as well?

Ideally yes, we should standardize everywhere!

jeremylenz (approved):
Great work @lfu !

Let's continue on in the next PR.  ACK :+1:

PR #10407, opened 2023-01-10
Fixes #35915 - Add RHEL to the OS filter list for Module Stream Tab

chris1984:

* On a production install of Katello(on RHEL)/Satellite the host itself is generated with Puppet and creates an OS name called RHEL, Katello creates the name normally.
* Adding RHEL to the filter list

* N/A

* Register a RHEL 8.7 host and verify you can see the module streams tab still
* Bonus points for changing the redux state to operatingsystem_name(pin):"RHEL 8.7"


lfu (approved):
Don't have a RHEL 8.7 to test with. But LGTM 👍🏻

PR #10409, opened 2023-01-11
Fixes #35926 - Include SmartProxyAuth into RegistrationController

stejskalleos:
When https://github.com/Katello/katello/pull/10315 patch is applied on production environment with older
Katello versions (like katello-4.5.0), registration will start failing on following error:

undefined local variable or method `auth_smart_proxy' for #<Api::V2::RegistrationController:0x00007f84842a5728>

Fix is to include SmartProxyAuth into the RegistrationController
instead of RegistrationControllerExtensions module.

See https://bugzilla.redhat.com/show_bug.cgi?id=2158959

* Register & subscribe host through the proxy
* Verify that it has been registered and it's subscribed to the Forman

### DIFF
--- a/app/controllers/katello/concerns/api/v2/registration_controller_extensions.rb
+++ b/app/controllers/katello/concerns/api/v2/registration_controller_extensions.rb
@@ -2,7 +2,6 @@ module Katello
   module Concerns
     module Api::V2::RegistrationControllerExtensions
       extend ActiveSupport::Concern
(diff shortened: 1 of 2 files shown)


stejskalleos:
Hi @ianballou, I'm looking for reviewers, could you please take a look?

ianballou (approved):
Works! Tested on master Katello and Satellite 6.11.

PR #10410, opened 2023-01-12
Fixes #35929 - destroy the product after *all* content is gone

evgeni:
Otherwise a content delete step can happen *after* the product is already deleted, resulting in `Product with ID "X" could not be found` errors raised from Candlepin.

Product deletion happens not in parallel with Content deletion anymore.

None

The following script was rather successful (80-90%) in reproducing the issue on Katello 4.5 with Candlepin 4.1.19 (the check is not present in earlier versions of Candlepin 4.1 branch):

```
hammer organization create --name lol
hammer product create --name lol --organization lol
hammer repository create --name lol --content-type yum --product lol --organization lol
hammer content-export complete library --organization lol
hammer content-export incremental library --organization lol
hammer repository delete --name lol --product lol --organization lol
hammer product delete --name lol --organization lol
hammer organization delete --name lol
```

### DIFF
--- a/app/lib/actions/katello/product/destroy.rb
+++ b/app/lib/actions/katello/product/destroy.rb
@@ -46,8 +46,9 @@ module Actions
                               product_id: product.cp_id,
                               content_id: pc.content.cp_content_id)
                 end
-                plan_action(Candlepin::Product::Destroy, cp_id: product.cp_id, :owner => product.organization.label)
               end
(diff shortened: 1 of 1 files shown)


chris1984 (approved):
ACK, works great when testing with the script. I can't really think of any edge cases where this would mess with anything.

PR #10411, opened 2023-01-12
Fixes #35935 - add info about reclaiming space from orphans

ianballou:
Adds a clarification to ensure users that disk space used by deleted repositories will be cleaned automatically.

This is in regards to confusion caused by trying to run space reclamation on smart proxies that have no repositories on them.

I don't think the added text makes the UI look bulky, but it's a consideration.

Find my new warnings in the UI for both Pulp Primary smart proxies and Smart Proxy Mirrors / Capsules / whatever you want to call every other smart proxy

### DIFF
--- a/app/views/foreman/smart_proxies/_content_sync.html.erb
+++ b/app/views/foreman/smart_proxies/_content_sync.html.erb
@@ -56,6 +56,7 @@
         <a ng-click="reclaimSpace()" ng-hide="syncStatus.download_policy != 'on_demand'">
           <span translate><strong>Reclaim Space</strong></span>
           <p translate> Delete cached content units from an "On Demand" Smart Proxy.</p>
+          <p translate> Storage used by deleted repositories will be reclaimed during the next orphan cleanup run.</p>
         </a>
       </li>
     </ul>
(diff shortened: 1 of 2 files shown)


sjha4:
Is this accurate? Does content from deleted repositories also get cleaned up when reclaiming space? In that case, do we need this clarification?

sjha4:
> Is this accurate? Does content from deleted repositories also get cleaned up when reclaiming space? In that case, do we need this clarification?

Oh I misread the text..Orphaned content will not be cleaned during reclaim space but will be automatically cleaned up during orphan cleanup run. I wonder if we should specify that orphaned content will not be cleaned with reclaim space task if it doesn't take too much UI space.

ianballou:
> I wonder if we should specify that orphaned content will not be cleaned with reclaim space task if it doesn't take too much UI space.

I was hoping that people would make the inference that space reclamation misses orphan cleanup because of this new warning.  Adding any more text would probably be too much.

sjha4:
> > I wonder if we should specify that orphaned content will not be cleaned with reclaim space task if it doesn't take too much UI space.
> 
> I was hoping that people would make the inference that space reclamation misses orphan cleanup because of this new warning. Adding any more text would probably be too much.

How does this sound? It makes it clear IMO that we are talking about a separate job.

`Storage used by deleted repositories will be reclaimed when the next orphan cleanup job runs.`

sjha4 (approved):
Acking..Feel free to update the text from the previous comment if it looks good.

ianballou:
@sjha4 I updated the text. Slight modification from yours to shorten it a little.

sjha4 (approved):
Ack..

PR #10412, opened 2023-01-12
Fixes #35936 - change to create_or_find_by!

jeremylenz:

When importing module streams, `first_or_create!` was creating a race condition that could occur during concurrent registrations. This would cause a key constraint error. 

Using `create_or_find_by!` seems to prevent this race condition.

Pablo has tested this and confirmed it works on two downstream builds.

### DIFF
--- a/app/models/katello/concerns/host_managed_extensions.rb
+++ b/app/models/katello/concerns/host_managed_extensions.rb
@@ -271,9 +271,9 @@ module Katello
       def import_module_streams(module_streams)
         streams = {}
         module_streams.each do |module_stream|
-          stream = AvailableModuleStream.where(name: module_stream["name"],
+          stream = AvailableModuleStream.create_or_find_by!(name: module_stream["name"],
                                                context: module_stream["context"],
-                                               stream: module_stream["stream"]).first_or_create!
+                                               stream: module_stream["stream"])
           streams[stream.id] = module_stream
         end
         sync_available_module_stream_associations(streams)


app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -271,9 +271,9 @@ def import_enabled_repositories(repos)
       def import_module_streams(module_streams)
         streams = {}
         module_streams.each do |module_stream|
-          stream = AvailableModuleStream.where(name: module_stream["name"],
+          stream = AvailableModuleStream.create_or_find_by!(name: module_stream["name"],
```
parthaa, line 274: Do you need the exclamation here ?

jeremylenz: We used one in `first_or_create!` so I just kept it.

sjha4:
This is so interesting. They get around race condition of stale reads by creating and catching the error instead of reading and create like first_or_create! does.
https://github.com/rails/rails/blob/f44d6fc197d0ccda811c88a7ea4e11ba9aee2cb0/activerecord/lib/active_record/relation.rb#L208

app/models/katello/concerns/host_managed_extensions.rb
parthaa:
```ruby
irb(main):006:0> ::Katello::ContentCredential.create_or_find_by!(name: "baz", organization: org, content_type: "cert", content: "justin")
=> #<Katello::ContentCredential id: 1, name: "baz", organization_id: 1, content: "justin", created_at: "2023-01-12 20:05:28", updated_at: "2023-01-12 20:05:28", content_type: "cert">
irb(main):007:0> ::Katello::ContentCredential.create_or_find_by!(name: "baz", organization: org, content_type: "cert", content: "justin")
        2: from lib/tasks/console.rake:5:in `block in <top (required)>'
        1: from (irb):7
ActiveRecord::RecordInvalid (Validation failed: Name has already been taken)
```
vs
```ruby
irb(main):009:0> ::Katello::ContentCredential.create_or_find_by(name: "baz", organization: org, content_type: "cert", content: "justin")
=> #<Katello::ContentCredential id: nil, name: "baz", organization_id: 1, content: "justin", created_at: nil, updated_at: nil, content_type: "cert">
```
I prefer the exclamation version.

parthaa (approved):
ACK

PR #10414, opened 2023-01-12
Refs #35236 - ak update requires the existence of organization

lfu:
Make sure `organization` exists to avoid an error like:
```
Could not update the activation key:
  undefined method `library' for nil:NilClass
```
Followup of https://github.com/Katello/katello/pull/10200

hammer activation-key update --name=test5 --organization-id=1 --release-version='willthiswork'
Before
```
Could not update the activation key:
  undefined method `library' for nil:NilClass
```
After
```
Could not update the activation key:
  Invalid release version: [willthiswork]
```

### DIFF
--- a/test/controllers/api/v2/activation_keys_controller_test.rb
+++ b/test/controllers/api/v2/activation_keys_controller_test.rb
@@ -295,6 +295,11 @@ module Katello
       assert_match 'Validation failed: Max hosts is not a number', @response.body
     end
 
+    def test_should_not_update_with_invalid_release
+      put(:update, params: { :organization_id => @organization.id, :id => @activation_key.id, :release_version => "foo" })
+      assert_response :bad_request
+    end
+
     test_attributes :pid => 'ec225dad-2d27-4b37-989d-1ba2c7f74ac4'
     def test_update_auto_attach
       new_auto_attach = !@activation_key.auto_attach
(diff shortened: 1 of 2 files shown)


jeremylenz (commented):
Works well!

Before:
```
$ hammer activation-key update --name=dev --organization-id=1 --release-version=8
Could not update the activation key:
  undefined method `library' for nil:NilClass
```

After:
```
$ hammer activation-key update --name=dev --organization-id=1 --release-version=8
Activation key updated.

$ hammer activation-key update --name=dev --organization-id=1 --release-version="hoo"
Could not update the activation key:
  Invalid release version: [hoo]
```

I'd also like to see a test added that reflects this change.

test/controllers/api/v2/activation_keys_controller_test.rb
```diff
@@ -295,6 +295,16 @@ def test_should_not_update_existing_with_invalid_max_hosts
       assert_match 'Validation failed: Max hosts is not a number', @response.body
     end
 
+    def test_should_not_update_with_invalid_release
+      activation_key = ActivationKey.new(
+        :name => 'new key', :organization => @organization
+      )
+      assert activation_key.save
```
jeremylenz, line 302: We're not testing the model's validity here, so I'd say the `assert .. save` isn't needed. (It's probably not needed in the previous test either.) I'd just go with `ActivationKey.create!`

jeremylenz (approved):
@lfu Test looks good, but there's a Rubocop issue.

APJ :+1:

PR #10415, opened 2023-01-12
Fixes #35927 - ensure upstream auth is nil if empty string

ianballou:
Pulp does not allow empty string username and passwords on remotes. So, if the username or password is blank, change it to nil so that Pulp clears the field out completely.

1) Create a repo with a username and password.
2) Check the remote in the pulpcore database to check that username and password exist
3) Clear out the username and password in the UI and check the DB again

### DIFF
--- a/app/services/katello/pulp3/repository.rb
+++ b/app/services/katello/pulp3/repository.rb
@@ -385,6 +385,8 @@ module Katello
         remote_options[:download_concurrency] = root.download_concurrency unless root.download_concurrency.blank?
         remote_options.merge!(username: root&.upstream_username,
                               password: root&.upstream_password)
+        remote_options[:username] = nil if remote_options[:username] == ''
+        remote_options[:password] = nil if remote_options[:password] == ''
         remote_options.merge!(ssl_remote_options)
       end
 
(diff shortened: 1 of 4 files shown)


sjha4:
Do we need to worry about repos which are in this state. The issue mentions hammer repository update --id XXX --upstream-username='' --upstream-password='' fixes this. Should we have an upgrade task to do this for repos?

ianballou:
@sjha4 sure I'll add a migration to be safe. We do update the remote at every sync, so it's not totally necessary, but it would be good to get the database in a safer state.

ianballou:
Updated!

ianballou:
@sjha4 , it appears that Katello 4.1 Pulp accepted empty string, and sometime after that they stopped accepting it.

sjha4 (approved):
APJ 👍🏼

sjha4:
[test katello]

sjha4:
@ianballou : Apt seems to have the same test for blank vs nil username/passwords..
https://github.com/Katello/katello/blob/master/test/services/katello/pulp3/repository/apt/apt_test.rb#L34

ianballou:
@sjha4 it should be fixed, we shall see

sjha4 (approved):

PR #10419, opened 2023-01-16
Fixes #35954 - Properly propagate org create errors

adamruzicka:

We are now forcing validations to run on a model before trying to display any errors it may have.

Try creating an organization with invalid label through api or hammer. Before this fix it fails without any indication what went wrong, the request ends up with an ISE. After the fix, the validation errors should be returned, request should fail with 422.

### DIFF
--- a/app/controllers/katello/api/v2/organizations_controller.rb
+++ b/app/controllers/katello/api/v2/organizations_controller.rb
@@ -82,6 +82,9 @@ module Katello
       respond_for_create :resource => @organization
     rescue => e
       ::Foreman::Logging.exception('Could not create organization', e)
+      # Force @organization.errors to be populated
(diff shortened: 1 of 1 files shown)


ofedoren (approved):
Thanks, @adamruzicka, fixes the issue.

chris1984 (approved):
ACK, works well

```bash
[vagrant@centos7-hammer-devel ~]$ hammer organization create --name "Test Organization" --label "Test Organization" 
Could not create the organization:
  resource have no errors
[vagrant@centos7-hammer-devel ~]$ hammer organization create --name "Test Organization" --label "Test Organization" 
Could not create the organization:
  Label is invalid
  Label cannot contain characters other than ascii alpha numerals, '_', '-'.
```

PR #10420, opened 2023-01-17
Fixes #35959 - prepare to use structured publisher for deb content

sbernhard:

### DIFF
--- a/app/models/katello/root_repository.rb
+++ b/app/models/katello/root_repository.rb
@@ -454,7 +454,7 @@ module Katello
           params = []
           params << "comp=#{deb_components.gsub(" ", ",")}"
           params << "rel=#{deb_releases.gsub(" ", ",")}"
-          path += "/?#{params.join('&')}"
+          path += "?#{params.join('&')}"
         else
           Rails.logger.warn("deb_use_structured_content is set, but repository #{self.name} is lacking deb_releases or deb_components, so we are defaulting to using simple content.")
           unless Setting['deb_use_simple_publish']
--- a/app/models/katello/root_repository.rb
+++ b/app/models/katello/root_repository.rb
@@ -447,6 +447,24 @@ module Katello
       end
     end
 
+    def format_custom_content_path
+      path = custom_content_path
+      if content_type == ::Katello::Repository::DEB_TYPE && Setting['deb_use_structured_content']
+        if deb_components.present? && deb_releases.present?
+          params = []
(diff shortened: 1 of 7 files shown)


sjha4:
@sbernhard : Is this part of a bigger RFE? For this PR is there a way to test this? Code LGTM as far as non-deb content goes but I am not sure how it impacts deb repos.

sbernhard:
> @sbernhard : Is this part of a bigger RFE? For this PR is there a way to test this? Code LGTM as far as non-deb content goes but I am not sure how it impacts deb repos.

To use this, we need to change the subscription-manager, too. We are working on it.

app/models/katello/root_repository.rb
```diff
@@ -434,6 +434,23 @@ def format_arches
       end
     end
 
+    def format_custom_content_path
+      path = custom_content_path
+      if content_type == ::Katello::Repository::DEB_TYPE && SETTINGS[:katello][:use_deb_structured_publisher] == true
```
sbernhard, line 439: Long term story: we want to get rid of this setting again. 
We need this setting so that old sub-man works. If "all users" use the new sub-man (see https://github.com/candlepin/subscription-manager/pull/3223 ) then we would remove this check setting check again. Additionally, we would remove to do a simple publish in pulp which would increase the performance a lot.

ianballou:
Is this ready to be merged? I'm curious what the testing steps would be from the debian side. After applying the sub-man patch and we sync a debian repo, what would we be looking for in the distribution path?

ianballou:
Thanks for moving it to a draft! Mind pinging us when it's ready?

app/services/katello/pulp3/repository/apt.rb
```diff
@@ -31,22 +31,19 @@ def sync_url_params(sync_options)
...
-            {
-              distributions: repo.deb_releases + "#{' default' unless repo.deb_releases.include? 'default'}"
-            }
-          )
+          distributions = "#{repo.deb_releases}"
+          if Setting['deb_use_simple_publish'] and not distributions.split(' ').include? 'default'
+            distributions.concat(' default')
+          end
+
+          super.merge({policy: policy, distributions: distributions})
```
quba42, line 39:
```suggestion
          super.merge({distributions: distributions})
```
This is a rebase error.

sbernhard:
Closed in favor of https://github.com/Katello/katello/pull/11058

PR #10421, opened 2023-01-17
Refs #35861 - Do not allow blank for content facet host

sjha4:
Should not allow nil as host for content facet host.
This is basically reverting a change from the parent PR for [35861](https://github.com/Katello/katello/pull/10391)
No real testing required because the db validations exist to not allow this. However, make sure you can't delete a host tied to a content facet from foreman console.


parthaa (approved):
apj

PR #10422, opened 2023-01-17
Fixes #35961 - Simple Content Access API shouldn't check upstream connection

jeremylenz:

The `organizations/:id/simple_content_access/` API endpoints were never updated after the upgrade to Candlepin 4.2. This means that

* you could not enable or disable SCA via the API
* you could not check SCA eligibility via the API

This change reflects that SCA is now tied to the organization, and no longer requires a connection to the upstream Candlepin consumer.

I'm not entirely sure if the `/eligible` endpoint is necessary any more. But I left it in for backward compatibility.

Get an organization with no manifest imported

```
curl -sku admin:changeme -H "content-type: application/json" -X PUT <server url>/katello/api/organizations/1/simple_content_access/disable
```

```
curl -sku admin:changeme -H "content-type: application/json" -X PUT <server url>/katello/api/organizations/1/simple_content_access/enable
```

```
curl -sku admin:changeme -H "content-type: application/json" -X GET <server url>/katello/api/organizations/1/simple_content_access/eligible
```

Before:
```
{"displayMessage":"Current organization does not have a manifest imported.","errors":["Current organization does not have a manifest imported."]}
```

After:

API responses as expected


jeremylenz:
Fixed tests, ready for review :heavy_check_mark:

app/controllers/katello/api/v2/simple_content_access_controller.rb
```diff
@@ -11,7 +10,7 @@ class Api::V2::SimpleContentAccessController < Api::V2::ApiController
     api :GET, "/organizations/:organization_id/simple_content_access/eligible",
       N_("Check if the specified organization is eligible for Simple Content Access")
     def eligible
-      eligible = @organization.upstream_consumer.simple_content_access_eligible?
+      eligible = @organization.simple_content_access_eligible?
```
chris1984, line 13: We use this here https://github.com/Katello/hammer-cli-katello/blob/master/lib/hammer_cli_katello/simple_content_access.rb#L5

Since SCA is the default now and no longer tied to upstream, do you think it's worth it to add a deprecation warning, then in 4.9 we can look at pulling it out.

chris1984 (commented):
Code looks sane, left one comment. Will start to test

app/controllers/katello/api/v2/simple_content_access_controller.rb
jeremylenz: yeah, we should probably do that and remove both the Katello and Hammer parts in 4.9.

jeremylenz:
@chris1984 Added deprecation warning

chris1984 (approved):
ACK, works great. Thanks for adding the dep warning @jeremylenz

PR #10423, opened 2023-01-17
Refs #35873 - More ouia-ids for ACS

sjha4:
Add ouiaId prop to supported PF4 components in ACS UI.
Uncomment `webpack/global_test_setup.js` as desired
Run `npx jest webpack/scenes/AlternateContentSources/`
See if any components complain about ouia-ids.

### DIFF
--- a/webpack/global_test_setup.js
+++ b/webpack/global_test_setup.js
@@ -16,7 +16,7 @@ global.console.error = (error, stack) => {
   //    'The prop `ouiaId` is marked as required in `Modal`, but its value is `undefined`',
   //    'created by WizardHeader', 'created by Navigation',
   //    'created by ActionsColumn', 'created by InactiveText', 'created by Select',
-  //    'created by Context.Consumer'];
+  //    'created by Context.Consumer', 'created by DualListSelector'];
 
   /* eslint-disable-next-line no-console */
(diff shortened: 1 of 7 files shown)


webpack/global_test_setup.js
```diff
@@ -16,7 +16,7 @@ global.console.error = (error, stack) => {
   //    'The prop `ouiaId` is marked as required in `Modal`, but its value is `undefined`',
   //    'created by WizardHeader', 'created by Navigation',
   //    'created by ActionsColumn', 'created by InactiveText', 'created by Select',
-  //    'created by Context.Consumer'];
+  //    'created by Context.Consumer', 'created by DualListSelector'];
```
chris1984, line 19: What is `DualListSelector`

sjha4: It's this patternfly component: https://www.patternfly.org/v4/components/dual-list-selector/

We use it on ACS creation wizard to select products and smart proxies. The component internally renders some child components which are ouia-id enabled but we can't set those in our code directly. Hence ignoring those for now.

chris1984 (approved):
ACK, don't see any complaints during the npx jest run about `ouia-ids` and went into the browser and inspected the ones in the pr and confirmed i can see them in the DOM

PR #10425, opened 2023-01-19
Fixes #35965 - properly clean orphaned smart proxy remotes

ianballou:
Properly causes orphaned remotes to be deleted from smart proxies during orphan cleanup.  Before, the filtering included all repositories that are in the database. Now, it only considers repositories that are available to the smart proxy.

None worth mentioning.

1) Register a smart proxy and create a new pulp-cli profile.
2) Look at the remotes on the smart proxy.
```
pulp --profile proxy rpm remote list
```
3) Sync a smart proxy to some environment.
4) Look at the created remotes.
5) Remove all (or some) environments from the smart proxy
6) Run orphan cleanup: `katello:delete_orphaned_content`
7) Check the remotes again. The remotes associated to the now-removed environments should be gone.
8) Try the process again after re-adding the environments.

### DIFF
--- a/app/services/katello/pulp3/smart_proxy_mirror_repository.rb
+++ b/app/services/katello/pulp3/smart_proxy_mirror_repository.rb
@@ -96,7 +96,8 @@ module Katello
 
       def delete_orphan_remotes
         tasks = []
-        repo_names = Katello::Repository.pluck(:pulp_id)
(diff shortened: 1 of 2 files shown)


sjha4 (approved):
Ack 👍🏼

PR #10430, opened 2023-01-23
Refs #35861 - Add empty content facet check to CV promote

sjha4:
Adds empty content facet check to promote.
Noticed an error when promoting a cv to an env with content facet where the host has been deleted.
1. Create and publish a CV twice and promote one version to environemnt env1 and add an activation key for the CV tied to env1.
2. Register a host with the activation key.
3. Go to `psql -d katello -U katello`
4. DELETE from hosts where id='id'
5. ALTER table to remove the FK constraints that fail.
 Follow steps here for the queries to run:  https://github.com/Katello/katello/pull/10391#issuecomment-1371230005
7. Once the host has been deleted, try promoting one version to env1 . You should see the error.
Screenshot from 2023-01-04 12-42-48

8. Checkout this branch and try promoting the other version to env1.
9. Run the rails task as the error message states.
10. The orphaned content facet and subscription facet will be deleted.
11. Promote a version to env1 again. It should succeed.

### DIFF
--- a/app/models/katello/content_view.rb
+++ b/app/models/katello/content_view.rb
@@ -604,7 +604,7 @@ module Katello
         fail _("Import-only content views can not be published directly") if import_only? && !syncable
         check_composite_action_allowed!(organization.library)
         check_docker_repository_names!([organization.library])
-        check_orphaned_content_facets(environments: self.environments)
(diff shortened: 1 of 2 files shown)


chris1984 (approved):
ACK, works well. Gave me a reason to reset my devel box :)

PR #10432, opened 2023-01-26
Fixes #36007 - Migration error 'column settings.category does not exist'

stejskalleos:
See:

https://github.com/theforeman/foreman/pull/9524
https://github.com/theforeman/foreman/pull/9524#issuecomment-1404661119

* Create fresh DB
* Run migrations
* Check that migration didn't fail with:
```
rails aborted!
StandardError: An error has occurred, this and all later migrations canceled:

PG::UndefinedColumn: ERROR:  column settings.category does not exist
LINE 1: ... IN (SELECT "settings"."id" FROM "settings" WHERE "settings"...
```


chris1984 (approved):
ACK, works well.

PR #10433, opened 2023-01-31
Fixes #36025 - Ensure custom products don't use Red Hat provider

jeremylenz:

Thanks to @hao-yu for doing most of the heavy lifting here :)

There are two situations where Katello needs to import product data from Candlepin and create new `::Katello::Product`s:

1. during manifest refresh
2. during `rake katello:reimport`

When doing this import, Katello was assuming that any product in Candlepin is automatically a Red Hat product. This is an incorrect assumption. If a _custom_ product has been deleted from Katello but remains in Candlepin, it was getting incorrectly reimported as a Red Hat product. This results in a few problems:

1. a seemingly unrelated error message about a Subscription being blank;
2. a second error that a Pool and a Subscription must belong to the same organization; and
3. most importantly, the inability to complete a successful manifest import or refresh.

This PR tackles this problem in a few ways:

1. As it turns out, the Candlepin ID of custom products is actually set in Katello, and is (usually) a 12-digit numeric string. When reimporting, we can use this to recognize custom products and ~~avoid assigning them to a Red Hat provider~~ avoid reimporting them. 
2. If your custom products have already been assigned an incorrect provider, you need a way to clean them up. I added this capability in a new upgrade rake task.
3. The error "A Pool and its Subscription cannot belong to different organizations" is actually irrelevant here. It was being triggered because Subscription was `nil` and `nil` is not equal to whatever the Pool was. I improved the validation by suppressing this message unless both Pool and Subscription are present.

The mechanism through which Katello assigns IDs to custom products was previously

```rb
SecureRandom.random_number(999_999_999_999)
```

At first glance you may think this will give you a 12-digit random number. But it actually gives you an integer _between_ 0 and 999,999,999,999. As a consequence, most custom products will have assigned `cp_id`s that are 12 digits. But occasionally one will be 11 digits. And less occasionally, it'll be 10. And even less than that, 9 digits. And so on.

So I altered Katello's definition of a custom product ID to allow between 8 and 12 digits. Based on probability, this seems reasonable, as less than 1 in 10,000 custom products will have an ID less than 8 digits:
```
[18] pry(main)> result = []
=> []
[19] pry(main)> 10000.times { result << SecureRandom.random_number(999_999_999_999).to_s.length }
=> 10000
[20] pry(main)> result.count(12)
=> 9019
[21] pry(main)> result.count(11)
=> 885
[22] pry(main)> result.count(10)
=> 85
[23] pry(main)> result.count(9)
=> 10
[24] pry(main)> result.count(8)
=> 1
[25] pry(main)> result.count(7)
=> 0
[26] pry(main)> result.count(6)
=> 0
[27] pry(main)> result.count(5)
=> 0
[28] pry(main)> result.count(4)
=> 0
[29] pry(main)> result.count(3)
=> 0
[30] pry(main)> result.count(2)
=> 0
[31] pry(main)> result.count(1)
=> 0
```

I also altered the way Katello generates IDs so that they will always be 12 digits going forward. But that won't help us with already-existing custom products.

Steps to reproduce:
1. Create a custom product.

2. Run the following command to delete the custom product from Katello database only
```rb
bundle exec rails console
product = Katello::Product.find_by_name("<PRODUCT NAME>")
product.provider    <====================== This should be an "Anonymous" provider
product.destroy
exit
```

3. Before checking out this patch, attempt a manifest refresh. It should fail with warning.
4. You can also try `rake katello:reimport` which will also fail.
5. Check out the product in Rails console again:
```rb
product = Katello::Product.find_by_name("test_custom_product")
```

See that it now has a 'Red Hat' provider.

6. Check out the patch and run a manifest refresh. It may fail with "Subscription can't be blank", but you shouldn't see "A Pool and its Subscription cannot belong to different organizations."
7. Run `bundle exec rake katello:upgrades:4.8:fix_incorrect_products`. It should note that it fixed 1 incorrect product.
8. Manifest refresh should now complete successfully.
9. Create another custom product, then delete it in Rails console.
10. Do the manifest refresh again. It should succeed. The log will have a warning "Found orphaned custom product".
11. Run `rake katello:delete_orphaned_custom_products`. It will tell you that it deleted 1 orphaned custom product for your organization.
12. Manifest refresh should now succeed always, whether or not you have orphaned products in Candlepin.

### DIFF
--- a/app/models/katello/glue/provider.rb
+++ b/app/models/katello/glue/provider.rb
@@ -6,6 +6,30 @@ module Katello
       base.send :include, InstanceMethods
     end
 
+    def self.orphaned_custom_product?(cp_id, organization)
+      return unless cp_id.present? && orphaned_product?(cp_id, organization)
+      product_provider = provider_for_cp_id(cp_id: cp_id, organization: organization)
+      if product_provider.redhat_provider?
+        false
+      else
+        Rails.logger.warn "Found orphaned object with id #{cp_id} in Candlepin. Skipping import into Katello; run rake katello:delete_orphaned_custom_products to remove it from Candlepin."
+        true
+      end
+    end
+
+    def self.orphaned_product?(cp_id, organization)
+      !organization.products.where(:cp_id => cp_id).exists?
+    end
+
+    def self.provider_for_cp_id(cp_id:, organization:)
+      return organization.redhat_provider unless ::Katello::Glue::Candlepin::Product.engineering_product_id?(cp_id)
+      if ::Katello::Glue::Candlepin::Product.custom_product_id?(cp_id)
+        organization.anonymous_provider
+      else
+        organization.redhat_provider
+      end
+    end
+
     module InstanceMethods
       API_URL = 'https://subscription.rhsm.redhat.com/subscription/consumers/'.freeze
 
@@ -126,6 +150,7 @@ module Katello
 
         Katello::Logging.time("Imported #{cp_products.size} products") do
           cp_products.each do |product_json|
+            next if ::Katello::Glue::Provider.orphaned_custom_product?(product_json['id'], organization)
             product = import_product(product_json)
             prod_content_importer.add_product_content(product, product_json['productContent']) if product.redhat?
           end
@@ -144,7 +169,7 @@ module Katello
         if product&.redhat?
           product.update!(:name => product_json['name']) unless product.name == product_json['name']
         elsif product.nil?
-          product = Glue::Candlepin::Product.import_from_cp(product_json, organization)
+          product = Glue::Candlepin::Product.import_redhat_product_from_cp(product_json, organization)
         end
         product
       end
--- a/app/models/katello/pool.rb
+++ b/app/models/katello/pool.rb
@@ -56,7 +56,10 @@ module Katello
     validate :subscription_matches_organization
 
     def subscription_matches_organization
-      errors.add(:base, _("A Pool and its Subscription cannot belong to different organizations")) unless subscription&.organization_id == self.organization_id
+      return if errors[:subscription].any? || errors[:organization].any? # let other validations catch this
+      unless subscription&.organization_id == self.organization_id
+        errors.add(:base, _("A Pool and its Subscription cannot belong to different organizations."))
+      end
     end
 
     DAYS_RECENTLY_EXPIRED = 30
--- a/app/models/katello/product.rb
+++ b/app/models/katello/product.rb
@@ -33,6 +33,7 @@ module Katello
 
     validates_lengths_from_database :except => [:label]
     validates :provider_id, :presence => true
+    validate :ensure_provider_type_matches_id
     validates_with Validators::KatelloNameFormatValidator, :attributes => :name
     validates_with Validators::KatelloLabelFormatValidator, :attributes => :label
     validates_with Validators::ProductUniqueAttributeValidator, :attributes => :name
@@ -236,7 +237,7 @@ module Katello
     end
 
     def self.unused_product_id
-      id = SecureRandom.random_number(999_999_999_999)
+      id = (SecureRandom.random_number * (10**12 - 10**11) + 10**11).to_i # guarantee that we get 12 digits
       if ::Katello::Product.find_by(:cp_id => id)
         unused_product_id
       else
@@ -252,6 +253,15 @@ module Katello
     end
     class Jail < ::Safemode::Jail
       allow :name, :label
+    end
+
+    private
+
+    def ensure_provider_type_matches_id
+      return if errors[:provider_id].any?
+      if provider.redhat_provider? && ::Katello::Glue::Candlepin::Product.custom_product_id?(cp_id)
+        errors.add(:base, _("Cannot associate a Red Hat provider with a custom product"))
+      end
     end
   end
 end
--- a/lib/katello/tasks/clean_candlepin_orphaned_products.rake
+++ b/lib/katello/tasks/clean_candlepin_orphaned_products.rake
@@ -1,0 +1,38 @@
+namespace :katello do
+  desc "Removes custom products (and their associated pools and subscriptions) from Candlepin that are not in Katello"
+  task :clean_candlepin_orphaned_products => ["dynflow:client", "environment"] do
+    User.current = User.anonymous_admin #set a user for orchestration
+
+    deleted_products_counts = {}
+    Organization.all.each do |org|
+      print "Cleaning Candlepin orphaned custom products for organization #{org.name}\n"
+
+      cp_products = Katello::Resources::Candlepin::Product.all(org.label)
+      cp_products = cp_products.select { |prod| Katello::Glue::Candlepin::Product.engineering_product_id?(prod['id']) }
+      cp_product_ids = cp_products.map { |cp_product| cp_product['id'] }
+
+      katello_product_ids = Katello::Product.where(:organization_id => org.id).pluck(:cp_id)
+      orphaned_cp_product_ids = cp_product_ids - katello_product_ids
+      orphaned_cp_product_ids.each do |cp_product_id|
+        print "Deleting Candlepin orphaned custom product #{cp_product_id} (and its associated pools and subscriptions)\n"
+        ForemanTasks.sync_task(
+          ::Actions::Candlepin::Product::DeletePools,
+          cp_id: cp_product_id, organization_label: org.label)
+
+        ForemanTasks.sync_task(
+          ::Actions::Candlepin::Product::DeleteSubscriptions,
+          cp_id: cp_product_id, organization_label: org.label)
+
+        ForemanTasks.sync_task(
+          ::Actions::Candlepin::Product::Destroy,
+          owner: org.label, cp_id: cp_product_id)
+      end
+
+      deleted_products_counts[org.name] = orphaned_cp_product_ids.length
+    end
+
+    deleted_products_counts.each do |org_name, deleted_products_count|
+      print "Deleted #{deleted_products_count} Candlepin orphaned custom products for organization #{org_name}\n"
+    end
+  end
+end
--- a/lib/katello/tasks/reimport.rake
+++ b/lib/katello/tasks/reimport.rake
@@ -14,7 +14,6 @@ namespace :katello do
   desc "Reimports information from backend systems"
   task :reimport => ["dynflow:client", "katello:check_ping"] do
     User.current = User.anonymous_admin #set a user for orchestration
-    Dir.glob(Katello::Engine.root.to_s + '/app/models/katello/*.rb').each { |file| require file }
 
     models = [
       Katello::Subscription,
@@ -32,7 +31,7 @@ namespace :katello do
       ack_key.import_pools
     end
 
-    print "Importing Linked Repositories"
+    print "Importing Linked Repositories\n"
     Katello::Repository.linked_repositories.each(&:index_content)
   end
 end
--- a/lib/katello/tasks/upgrades/4.8/fix_incorrect_providers.rake
+++ b/lib/katello/tasks/upgrades/4.8/fix_incorrect_providers.rake
@@ -1,0 +1,29 @@
+namespace :katello do
+  namespace :upgrades do
+    namespace '4.8' do
+      desc "Fix custom products incorrectly assigned a Red Hat provider"
+      task :fix_incorrect_providers => ["dynflow:client", "environment"] do
+        User.current = User.anonymous_admin #set a user for orchestration
+
+        print "Fixing incorrect providers\n"
+        incorrect_provider_count = 0
+        error_msgs = []
+        Katello::Product.redhat.includes(:organization).each do |product|
+          if ::Katello::Glue::Candlepin::Product.custom_product_id?(product.cp_id)
+            print "Fixing provider for #{product.name}\n"
+            incorrect_provider_count += 1
+            product.provider = product.organization.anonymous_provider
+            product.save!
+          end
+        end
+
+        print "Fixed #{incorrect_provider_count} incorrect providers\n"
+        if error_msgs.any?
+          print "Errors while fixing providers: #{error_msgs.join("\n")}\n"
+        end
+
+        Rake::Task['katello:clean_candlepin_orphaned_products'].invoke
+      end
+    end
+  end
+end
--- a/app/models/katello/glue/candlepin/pool.rb
+++ b/app/models/katello/glue/candlepin/pool.rb
@@ -56,6 +56,12 @@ module Katello
         end
       end
 
+      def import_candlepin_records(pools, org)
+        # Skip import of pools that were associated with an orphaned custom product
+        pools = pools.reject { |cp_pool| ::Katello::Glue::Provider.orphaned_custom_product?(cp_pool['productId'], org) }
+        super(pools, org)
+      end
+
       def import_candlepin_record(record:, organization:)
         subscription = determine_subscription(
           product_id: record['productId'],
--- a/app/models/katello/glue/candlepin/product.rb
+++ b/app/models/katello/glue/candlepin/product.rb
@@ -32,7 +32,7 @@ module Katello
       id.match(/^\d+$/) #engineering products are numeric
     end
 
-    def self.import_from_cp(attrs, organization)
+    def self.import_redhat_product_from_cp(attrs, organization)
       import_logger = attrs[:import_logger]
 
       product_attrs = {'name' => attrs['name'],
@@ -48,6 +48,12 @@ module Katello
         logger&.error "Failed to create product #{attrs['name']}: #{e}"
       end
       raise e
+    end
+
+    def self.custom_product_id?(id)
+      # Engineering products with 12 digits are custom products (see Katello::Product#unused_product_id)
+      # however, previously generated ids are random and can be shorter than 12 digits
+      id =~ /^\d{8,12}$/
     end
 
     module InstanceMethods
--- a/app/models/katello/glue/candlepin/subscription.rb
+++ b/app/models/katello/glue/candlepin/subscription.rb
@@ -21,6 +21,11 @@ module Katello
           !Glue::Candlepin::Product.engineering_product_id?(product['id']) || Katello::Product.find_by(:cp_id => product['id']).try(:custom?)
         end
       end
+
+      def import_candlepin_records(cp_subs, org)
+        cp_subs = cp_subs.reject { |cp_subscription| ::Katello::Glue::Provider.orphaned_custom_product?(cp_subscription['productId'], org) }
+        super(cp_subs, org)
+      end
     end
 
     module InstanceMethods
--- a/app/services/katello/pulp3/content_view_version/export_validation_error.rb
+++ b/app/services/katello/pulp3/content_view_version/export_validation_error.rb
@@ -1,0 +1,7 @@
+module Katello
+  module Pulp3
+    module ContentViewVersion
+      class ExportValidationError < HttpErrors::BadRequest; end
+    end
+  end
+end
--- a/app/services/katello/pulp3/content_view_version/export_validator.rb
+++ b/app/services/katello/pulp3/content_view_version/export_validator.rb
@@ -1,7 +1,6 @@
 module Katello
   module Pulp3
     module ContentViewVersion
-      class ExportValidationError < HttpErrors::BadRequest; end
       class ExportValidator
         delegate :content_view_version, :from_content_view_version, :format, :repositories,
                  :smart_proxy, :version_href_to_repository_href, to: :@export_service
--- a/db/seeds.d/111-upgrade_tasks.rb
+++ b/db/seeds.d/111-upgrade_tasks.rb
@@ -7,6 +7,7 @@ UpgradeTask.define_tasks(:katello) do
     {:name => 'katello:upgrades:4.1:fix_invalid_pools'},
     {:name => 'katello:upgrades:4.1:reupdate_content_import_export_perms'},
     {:name => 'katello:upgrades:4.2:remove_checksum_values'},
-    {:name => 'katello:upgrades:4.4:publish_import_cvvs'}
+    {:name => 'katello:upgrades:4.4:publish_import_cvvs'},
+    {:name => 'katello:upgrades:4.8:fix_incorrect_providers'}
   ]
 end
--- a/spec/models/product_spec.rb
+++ b/spec/models/product_spec.rb
@@ -89,12 +89,13 @@ module Katello
       specify { Product.new(:label => "shoo", :name => 'contains space', :provider => @provider).must_be :valid? }
       specify { Product.new(:label => "bar foo", :name => "foo", :provider => @provider).wont_be :valid? }
       it "should not be successful when creating a product with a duplicate name in one organization" do
-        @p = Product.create!(ProductTestData::SIMPLE_PRODUCT.merge(:organization_id => @organization.id))
-
-        Product.new(:name => @p.name, :label => @p.name,
-                    :id => @p.cp_id,
-                    :provider => @p.provider
-                   ).wont_be :valid?
+        @p = Product.create!(ProductTestData::SIMPLE_PRODUCT.merge(:organization_id => @organization.id, :provider_id => @provider.id))
+        np = Product.new(:name => @p.name, :label => @p.name,
+                    :cp_id => "42424242424",
+                    :provider => @p.provider,
+                    :organization => @organization
+                   )
+        np.wont_be :valid?
       end
     end
 
@@ -114,7 +115,13 @@ module Katello
     end
 
     it 'should be destroyable' do
-      product = create(:katello_product, :fedora, provider: create(:katello_provider), organization: @organization)
+      product = create(
+                       :katello_product,
+                       :fedora,
+                       name: 'to be destroyed',
+                       label: 'to_be_destroyed',
+                       provider: create(:katello_provider),
+                       organization: @organization)
       assert product.destroy
     end
   end
--- a/test/actions/katello/alternate_content_source_test.rb
+++ b/test/actions/katello/alternate_content_source_test.rb
@@ -111,7 +111,10 @@ module ::Actions::Katello::AlternateContentSource
 
     it 'fails to create simplified during update with empty product' do
       action.expects(:action_subject).with(simplified_acs)
-      empty_product = ::Katello::Product.create(name: 'empty', organization_id: ::Organization.first)
+      empty_product = ::Katello::Product.create(
+        name: 'empty',
+        organization_id: ::Organization.first.id,
+        provider: ::Organization.first.anonymous_provider)
       assert_raises ActiveRecord::RecordInvalid do
         plan_action action, simplified_acs, [proxy, mirror], [empty_product], {}
       end
--- a/test/factories/product_factory.rb
+++ b/test/factories/product_factory.rb
@@ -2,12 +2,17 @@ FactoryBot.define do
   factory :katello_product, :class => Katello::Product do
     sequence(:name) { |n| "Product #{n}" }
     sequence(:label) { |n| "product_#{n}" }
+    sequence(:cp_id) { |n| (100_000_000_000 + n).to_s }
     association :provider, :factory => :katello_provider, :strategy => :build
 
     trait :fedora do
       name { "Fedora" }
       description { "The open source Linux distribution." }
       label { "fedora_label" }
+    end
+
+    trait :redhat do
+      sequence(:cp_id) { |n| "RH00000#{n}" }
     end
 
     trait :with_provider do
--- a/test/fixtures/models/katello_products.yml
+++ b/test/fixtures/models/katello_products.yml
@@ -2,7 +2,7 @@ debian:
   name:            Debian
   description:     The universal operating system
   label:           debian_label
-  cp_id:           debian
+  cp_id:           123123123000
   provider_id:     <%= ActiveRecord::FixtureSet.identify(:anonymous) %>
   organization_id: <%= ActiveRecord::FixtureSet.identify(:empty_organization) %>
 
@@ -10,7 +10,7 @@ fedora:
   name:         Fedora
   description:  An open source Linux distribution.
   label:        fedora_label
-  cp_id:        fedora
+  cp_id:        123123123001
   provider_id:  <%= ActiveRecord::FixtureSet.identify(:anonymous) %>
   organization_id: <%= ActiveRecord::FixtureSet.identify(:empty_organization) %>
 
@@ -25,7 +25,7 @@ redhat:
 puppet_product:
   name:         Puppet Product
   description:  A product for puppet content
-  cp_id:        puppet
+  cp_id:        123123123002
   label:        puppet_product
   provider_id:  <%= ActiveRecord::FixtureSet.identify(:anonymous) %>
   organization_id: <%= ActiveRecord::FixtureSet.identify(:empty_organization) %>
@@ -33,14 +33,14 @@ puppet_product:
 ostree_product:
   name:         OSTree Product
   description:  A product for ostree content
-  cp_id:        ostree
+  cp_id:        123123123003
   label:        ostree_product
   provider_id:  <%= ActiveRecord::FixtureSet.identify(:anonymous) %>
   organization_id: <%= ActiveRecord::FixtureSet.identify(:empty_organization) %>
 
 empty_product:
   name:         Empty Product
-  cp_id:        empty_product
+  cp_id:        123123123004
   label:        empty_product
   provider_id:  <%= ActiveRecord::FixtureSet.identify(:anonymous) %>
   organization_id: <%= ActiveRecord::FixtureSet.identify(:empty_organization) %>
@@ -55,7 +55,7 @@ empty_redhat:
 
 empty_product_2:
   name:         Empty Product_2
-  cp_id:        empty_product_2
+  cp_id:        123123123005
   label:        empty_product_2
   provider_id:  <%= ActiveRecord::FixtureSet.identify(:anonymous) %>
   organization_id: <%= ActiveRecord::FixtureSet.identify(:empty_organization) %>
@@ -64,6 +64,6 @@ oracle:
   name:         Oracle
   description:  An open source Linux distribution.
   label:        oracle_label
-  cp_id:        oracle
+  cp_id:        123123123006
   provider_id:  <%= ActiveRecord::FixtureSet.identify(:anonymous) %>
   organization_id: <%= ActiveRecord::FixtureSet.identify(:empty_organization) %>
--- a/test/models/pool_test.rb
+++ b/test/models/pool_test.rb
@@ -139,6 +139,7 @@ module Katello
       Katello::Resources::Candlepin::Pool.expects(:get_for_owner).returns([pool_data])
       Katello::Resources::Candlepin::Pool.expects(:find).returns(pool_data)
       Katello::Pool.any_instance.expects(:import_managed_associations).returns
+      Organization.any_instance.expects(:redhat_provider).returns(katello_providers(:redhat))
       Pool.import_all(org)
 
       refute_empty Katello::Pool.where(organization: org)
--- a/test/services/katello/pulp3/content_view_version/export_test.rb
+++ b/test/services/katello/pulp3/content_view_version/export_test.rb
@@ -96,7 +96,7 @@ module Katello
             ::Katello::Pulp3::ContentViewVersion::ExportValidator.any_instance.expects(:version_href_to_repository_href).with("1").returns("1")
             ::Katello::Pulp3::ContentViewVersion::ExportValidator.any_instance.expects(:version_href_to_repository_href).with(nil).returns(nil).at_least_once
 
-            exception = assert_raises(Katello::Pulp3::ContentViewVersion::ExportValidationError) do
+            exception = assert_raises(::Katello::Pulp3::ContentViewVersion::ExportValidationError) do
               export.validate!(fail_on_missing_content: true, validate_incremental: true)
             end
             assert_match(/cannot be incrementally updated/, exception.message)
@@ -169,7 +169,7 @@ module Katello
           it "fails on validate! if chunk_size is >= 1_000_000GB" do
             export = setup_environment
 
-            exception = assert_raises(Katello::Pulp3::ContentViewVersion::ExportValidationError) do
+            exception = assert_raises(::Katello::Pulp3::ContentViewVersion::ExportValidationError) do
               export.validate!(fail_on_missing_content: false, validate_incremental: false, chunk_size: 1e6)
             end
             assert_match(/Specify an export chunk size less than 1_000_000 GB/, exception.message)


lib/katello/tasks/reimport.rake
```diff
@@ -16,6 +16,22 @@ namespace :katello do
     User.current = User.anonymous_admin #set a user for orchestration
     Dir.glob(Katello::Engine.root.to_s + '/app/models/katello/*.rb').each { |file| require file }
 
+    print "Fixing incorrect providers\n"
+    incorrect_provider_count = 0
+    error_msgs = []
```
jeremylenz, line 21: Originally I was just printing these rather than tallying them, but the task output is bulldozed by several screens of info. So I remember the data here and then output it below, when we're done.

jturel:
@jeremylenz can I leave some notes here? :)

parthaa:
curious to hear what @jturel had to say :)

I am wondering if we are approaching this wrong. Seems like dirty data is present in Candlepin but not in Katello. If thats the case I would suggest ignoring Custom products on import and may be warn + have user run a rake task to remove it from Candlepin.

app/models/katello/pool.rb
```diff
@@ -50,13 +50,16 @@ class Pool < Katello::Model
 
     validates_lengths_from_database
 
-    validates :subscription, :presence => true
+    validates :subscription, :presence => { :message => N_("cannot be blank. If this is during a manifest import or refresh, try running foreman-rake katello:reimport.") }
```
jturel, line 53: This isn't an error that a user should ever see! As you found, it was (incorrectly) being thrown due to some underlying bad data. Not great, but IMO better that it failed so that you could fix the underlying problem. What I'm getting at is that I don't think this message is needed and katello:reimport wouldn't fix it (right?)

app/models/katello/pool.rb
```diff
@@ -50,13 +50,16 @@ class Pool < Katello::Model
...
     validates :organization, :presence => true
 
     validate :subscription_matches_organization
 
     def subscription_matches_organization
-      errors.add(:base, _("A Pool and its Subscription cannot belong to different organizations")) unless subscription&.organization_id == self.organization_id
+      return if subscription&.organization_id.blank? || self.organization_id.blank? # let other validations catch this
```
jturel, line 59: I had the same thought as you did when I wrote this validation. But then I thought: when would either of those be blank? The underlying issue you're working on has triggered this validation so I think it's good to leave it alone and in its more strict form even if it's triggered by Some Other Problem

app/models/katello/product.rb
```diff
@@ -31,8 +31,11 @@ class Product < Katello::Model
     has_many :alternate_content_source_products, dependent: :delete_all, inverse_of: :product, :class_name => "Katello::AlternateContentSourceProduct"
     has_many :alternate_content_sources, :through => :alternate_content_source_products, :inverse_of => :products, :class_name => "Katello::AlternateContentSource"
 
+    before_validation :set_provider_type_from_id, only: [:create]
```
jturel, line 34: I think you already fixed the create path, or maybe it should go into some Dynflow action or similar?

app/models/katello/product.rb
```diff
@@ -253,5 +256,18 @@ def self.unused_product_id
     class Jail < ::Safemode::Jail
       allow :name, :label
     end
+
+    private
+
+    def ensure_provider_type_matches_id
+      if provider&.redhat_provider? && ::Katello::Glue::Candlepin::Product.custom_product_id?(cp_id)
```
jturel, line 263: when can provider be nil?

app/models/katello/product.rb
```diff
@@ -253,5 +256,18 @@ def self.unused_product_id
...
+        errors.add(:base, _("Cannot associate a Red Hat provider with a custom product"))
+      end
+    end
+
+    def set_provider_type_from_id
+      return unless self.organization&.present? && self.organization.anonymous_provider.present? && self.organization.redhat_provider.present?
```
jturel, line 269: When can any of these 3 entities not be present? Seems like something is really wrong in that case

lib/katello/tasks/reimport.rake
```diff
@@ -16,6 +16,22 @@ namespace :katello do
...
+    Katello::Product.redhat.each do |product|
```
jturel, line 22: I think the spirit of your PR is to fix this once and for all but this reimport task is more general purpose to fix out-of-sync things. What you likely want to do is add an upgrade rake task like this
 https://github.com/Katello/katello/commit/df86940f1ab1116b8b3126424ba6064aa69e4853#diff-adc47a3fb24f22bba9fa2f908cc1e0676804459186bd60f27dc94f1febbd1f06

That'd get run when the system is updated or if it gets hotfixed. Your the rest of your PR will prevent that script from needing to run again

jturel (commented):
@jeremylenz I hope my comments match the reality of what you want to fix here which I understand to be the product <-> provider mismatch.

Hao mentions in the bug: "This issue can happen when one or more custom products have been deleted from the Katello but not yet deleted from Candlepin" That seems weird but I guess people can mess around in their dbs and skip dynflow steps. etc, and you're doing this to insulate fallout from that & fix data integrity in general.

My question is - while this PR seems valid - is it possible to identify and delete or otherwise stop importing those orphaned products in candlepin?

app/models/katello/product.rb
jeremylenz: I actually hit a NilClass error here, but I don't remember exactly where. It may have been in a test.

lib/katello/tasks/reimport.rake
ianballou: I agree, since the incorrect providers fix should only be a one time fix thanks to the changes you've made. You could just yank out the changes here and stick them in an upgrade task.

jeremylenz: The reimport task (without the new code) will fail if you've got products with incorrect providers, though. I have no problem moving it, but I'd probably want to direct users to that new rake task.

app/models/katello/product.rb
jeremylenz: I think this was encountered in a unit test as well.

app/models/katello/pool.rb
jeremylenz: But aren't we both a fan of useful, accurate error messages? This was triggered not because `subscription.organization_id` was blank, but because `subscription` was blank. So it's not true that the organizations are different-- the error message is just inaccurate.  In this case it would fall back to the already-existing `validates :subscription, :presence => true` validation-- which is why I added the message to that.

jeremylenz: basically the flow is
1. you try to import a bad manifest and hit this issue
2. you see the manifest refresh/import fail with "If this is during a manifest import or refresh, try running foreman-rake katello:reimport"
3. you run the rake task and all is fixed!

app/models/katello/product.rb
jeremylenz: Yeah, I wasn't sure I needed both this and the validation, but it's nice to have. I'll see if anything breaks without it :P

lib/katello/tasks/reimport.rake
ianballou: By the time the user gets your new code here, the upgrade task should have already run. They shouldn't even need to run the rake task manually.  The installer I believe is responsible for running `foreman-rake upgrade:run`.

jeremylenz: ooh, I see what you're saying now!

jeremylenz:
> is it possible to identify and delete or otherwise stop importing those orphaned products in candlepin?

If we go with @parthaa's suggestion here, and skip importing custom products altogether, we'd also need to make sure those are deleted from Candlepin. I was thinking to just do it during the reimport, but we could also direct users to a rake task.

I'm still not sure if that's the right solution though. Can we always be sure that Katello is the source of truth for custom products and not Candlepin? Do we make that decision differently for Red Hat products? Thoughts?

app/models/katello/pool.rb
jturel: I see your point! +1 to including it.

jturel:
> > is it possible to identify and delete or otherwise stop importing those orphaned products in candlepin?
> 
> If we go with @parthaa's suggestion here, and skip importing custom products altogether, we'd also need to make sure those are deleted from Candlepin. I was thinking to just do it during the reimport, but we could also direct users to a rake task.

I think partha was suggesting that for a given custom product that can't be imported, just ignore it for the sake of allowing the user's action to succeed

> I'm still not sure if that's the right solution though. Can we always be sure that Katello is the source of truth for custom products and not Candlepin? Do we make that decision differently for Red Hat products? Thoughts?

app/models/katello/product.rb
jturel: I think the providers get created on organization create so they should be in place which brings the accuracy of those unit tests into question! of course, gotta double check what i'm saying

ianballou:
> I'm still not sure if that's the right solution though. Can we always be sure that Katello is the source of truth for custom products and not Candlepin? Do we make that decision differently for Red Hat products? Thoughts?

Users already have the expectation that the reimport task will fix their broken "backend objects". As such I think a user might be disappointed to see a message saying "hey something is broken, let's purge it" rather than "let's fix it and reimport it".

jeremylenz:
Okay, so two paths lie before me:

Partha Path:
- Change reimport task to not import custom products from Candlepin at all
- Provide a rake task to clean up orphaned products in Candlepin

Jturel path:
- Leave reimport task unchanged (import orphaned custom products from Candlepin, making them un-orphaned)

In both paths:
- For cleanup of products with incorrect providers, move this code to an upgrade rake task
- update error message to direct to the upgrade rake task (can I do that? Can you run just one of the upgrade subtasks? or maybe I should just remove the error message hint..)

I'm kind of leaning towards Jturel's path since it seems safer not to delete user data.  Thoughts?

jeremylenz:
> Is the data in Candlepin actually dirty for this issue? My thought was the data was wrongly imported into Katello, but Candlepin is still correct.

Yeah, this is my understanding too!

parthaa:
> In both paths:
> 
> * For cleanup of products with incorrect providers, move this code to an upgrade rake task
> * update error message to direct to the upgrade rake task (can I do that? Can you run just one of the upgrade subtasks? or maybe I should just remove the error message hint..)
> 
> I'm kind of leaning towards Jturel's path since it seems safer not to delete user data. Thoughts?

I think the fact that data exists in candlepin and not katello clearly tells me a Custom Product delete gone wrong and skipped on dynflow.  I'd argue Katello is the right source for custom products because it is responsible for creating the products and unlimited subscriptions associated to that product.
One other thing to note is if custom product delete went wrong we don't know if there is other data in  services also impacted. It has to be dirty data  (unless you can find another cause for this behavior.)  
I'd vouch for 
- Upgrade task 
- Warn user

Creating a shell custom product to address this seems to be imho weird.

jturel:
I do like Partha's idea - it just takes care to make that change. What if a custom product is being created at the same time that `reimport` is running, or a manifest refresh is happening? It'd be wrong to delete that half-baked data at that time, etc. So the code would need to check for those things

app/models/katello/pool.rb
jturel: Lightbulb moment? Rather than checking for presence/absence of org and subscription, what about checking if there any _errors_ for those fields? This changes the logic from: 'skip this validation if these fields are blank' to 'skip this validation if we already have errors on subscription and org'

I know this is a minor detail in the bigger picture here, but wanted to share my thought now that I slept on it

jeremylenz: Sounds like this will work, but only if Rails runs custom validations after stock validations. I'll check on that.

ianballou:
Partha's idea sounds fine, we'll just need a way to be 100% sure that the entity in Candlepin is an orphan and not something being freshly created.  I suppose we would need to query for running Dynflow tasks (or things in a queue?).

jeremylenz:
I wonder if Foreman Tasks' lock system could help us here.. @adamruzicka

jturel:
Depending on the severity of this issue you might consider splitting it into two halves -  completing this PR as it is defined which would unblock users and hotfixing as needed. Then a second fix to add a deeper solution around handling the bad Candlepin data. That way you have more time to solve that more sensitive aspect

adamruzicka:
Now that I had time to think on it, it doesn't feel like locking in tasks is the right tool for the job. It allows a single task to exclusively lock a resource, which will make all other tasks attempting to lock the same resource fail. It doesn't seem to be fine-grained enough for this situation where you might want to lock only certain combinations of things.

jeremylenz:
We decided to do

1. Upgrade rake task that fixes incorrect providers on existing products;
2. When reimporting, don't recreate orphaned custom products. Instead, log warnings and suggest that the user run
3. Regular rake task that finds orphaned custom products in Candlepin and calls Candlepin API to clean them up.

jeremylenz:
Also, we confirmed that when custom products are created in Katello (https://github.com/Katello/katello/blob/master/app/lib/actions/katello/product/create.rb#L11), the Katello database entry is created before Candlepin is called. So race conditions involving product creates during reimport shouldn't be as much of an issue.

app/models/katello/glue/provider.rb
```diff
@@ -144,14 +150,39 @@ def import_product(product_json)
...
+          # we have an orphaned Red Hat product
+          product = Glue::Candlepin::Product.import_redhat_product_from_cp(product_json, organization)
         end
         product
       end
 
-      def index_subscriptions
-        Katello::Subscription.import_all(organization)
-        Katello::Pool.import_all(organization, false)
+      def orphaned_custom_product?(product_json, organization)
```
jturel, line 159: Now that these are instance methods I think you already have access to `organization` and don't need to send it as a param

app/models/katello/glue/provider.rb
```diff
@@ -144,14 +150,39 @@ def import_product(product_json)
...
+        unless product_provider.redhat_provider?
+          Rails.logger.info "Found orphaned custom product #{product_json['id']} in Candlepin: '#{product_json['name']}'. Skipping import into Katello; run rake katello:delete_orphaned_custom_products to remove it from Candlepin."
+          true
+        else
+          false
+        end
+      end
+
+      def orphaned_product?(product_json, organization)
+        organization.products.find_by(:cp_id => product_json['id']).blank?
```
jturel, line 171: If you need to query this per-product, consider using a `where(...).exists?` this way you avoid unnecessarily instantiating the product if it does exist

app/models/katello/glue/provider.rb
```diff
@@ -144,14 +150,39 @@ def import_product(product_json)
...
+        return organization.redhat_provider unless ::Katello::Glue::Candlepin::Product.engineering_product_id?(cp_id)
+        if ::Katello::Glue::Candlepin::Product.custom_product_id?(cp_id)
+          organization.anonymous_provider
+        else
+          organization.redhat_provider
+        end
+      end
+
+      def index_subscriptions(skip_cp_ids: [])
+        Katello::Subscription.import_all(organization, nil, skip_cp_ids)
```
jturel, line 184: Would it make sense for `import_all` to have a param called `skip_orphaned` that then does the thing instead of passing them down through `import_all` ?

jeremylenz: not sure I understand this suggestion.?

jturel: It really was just a question! Worded differently: whose responsibility is it to know what constitutes an orphaned (or skipped) product? Should the import_all implementation care about determining that  (rather than being told those values), or is this the best place for it - the answer might be yes :) I'm a little fuzzy on the whole execution path here so I can't say

jeremylenz: ohh, gotcha! I did play around with this a bit, and it does clean up the code quite a bit not to have to pass around `skip_product_ids` or `skip_cp_ids` anywhere. In addition, I discovered that `Subscription.import_all` is also called in a previous upgrade rake task, fix_invalid_pools.rake, so we must account for that too. This made me realize that whenever we call `import_all` in the future, it _will_ break if you have orphaned CP products and don't pass them in. So good catch!

But making that change introduces two complications: (a) now, in addition to `import_all`, `import_products` also needs to know which ones to skip; and (b) in the case of `import_all`, since it's for Pools and Subscriptions the field to look at is `object['productId']` whereas for Products it's `object['id']`. This makes it either impossible or really difficult to keep that knowledge in `import_all`. For this reason I think it's better to keep it as is, but make it a required keyword arg. I'll push shortly with what I mean. Thoughts?

lib/katello/tasks/upgrades/4.8/fix_incorrect_providers.rake
```diff
@@ -0,0 +1,32 @@
+namespace :katello do
+  namespace :upgrades do
+    namespace '4.8' do
+      desc "Fix custom products incorrectly assigned a Red Hat provider"
+      task :fix_incorrect_providers => ["dynflow:client", "katello:check_ping"] do
```
jeremylenz, line 5: Do I need the `katello:check_ping` here? it doesn't seem to work anyway (complains about foreman-tasks..)

app/models/katello/pool.rb
```diff
@@ -50,13 +50,16 @@ class Pool < Katello::Model
 
     validates_lengths_from_database
 
-    validates :subscription, :presence => true
+    validates :subscription, :presence => { :message => N_("cannot be blank. If this is during a manifest import or refresh, try running foreman-rake katello:upgrades:4.8:fix_incorrect_providers.") }
```
jturel, line 53: Given that:

- the upgrade rake task has been run already when this code hits via installer (correcting any incorrect providers)
- you have put in place validations that prevent mismatched providers (should give validation errors of their own)
- you no longer import orphaned products

=> This validation should never, ever fire since it's not based on any kind of user input. And if it does fire, I don't think it would be because of anything to do w/ the providers at that point but some other bug the user can't work around themselves.

The upgrade rake tasks are also periodically removed, so the message would direct the user to do something totally un-doable in a few releases

app/models/katello/product.rb
jturel: Definitely work looking into. The code communicates that provider can be nil. If that's true,  then 👍 . But if that's not true then it's confusing (and wrong)!

lib/katello/tasks/upgrades/4.8/fix_incorrect_providers.rake
jturel: I think you need `environment`

https://github.com/Katello/katello/blob/master/lib/katello/tasks/upgrades/4.4/publish_import_cvvs.rake#L5

lib/katello/tasks/upgrades/4.8/fix_incorrect_providers.rake
```diff
@@ -0,0 +1,32 @@
...
+        User.current = User.anonymous_admin #set a user for orchestration
+        Dir.glob(Katello::Engine.root.to_s + '/app/models/katello/*.rb').each { |file| require file }
```
jturel, line 7: Once you add `environment` above I highly doubt you need this line

lib/katello/tasks/upgrades/4.8/fix_incorrect_providers.rake
```diff
@@ -0,0 +1,32 @@
+namespace :katello do
```
jturel, line 1: Gotta add this task to https://github.com/Katello/katello/blob/master/db/seeds.d/111-upgrade_tasks.rb in order to get picked up by the installer

lib/katello/tasks/upgrades/4.8/fix_incorrect_providers.rake
```diff
@@ -0,0 +1,32 @@
...
+    namespace '4.8' do
+      desc "Fix custom products incorrectly assigned a Red Hat provider"
+      task :fix_incorrect_providers => ["dynflow:client", "katello:check_ping"] do
+        User.current = User.anonymous_admin #set a user for orchestration
+        Dir.glob(Katello::Engine.root.to_s + '/app/models/katello/*.rb').each { |file| require file }
+
+        print "Fixing incorrect providers\n"
+        incorrect_provider_count = 0
+        error_msgs = []
+        Katello::Product.redhat.each do |product|
```
jturel, line 12: Maybe an `includes(:organization)` here to avoid N+1

lib/katello/tasks/upgrades/4.8/fix_incorrect_providers.rake
```diff
@@ -0,0 +1,32 @@
...
+          if ::Katello::Glue::Candlepin::Product.custom_product_id?(product.cp_id)
+            print "Fixing provider for #{product.name}\n"
+            incorrect_provider_count += 1
+            product.provider = product.organization.anonymous_provider
+            product.save!
+          end
+        rescue => e
+          print "Error fixing provider: #{e.message}\n"
```
jturel, line 20: This seems like it would be pretty bad & undesirable (an error here).

Should it really be rescued and ignored - I'm actually leaning toward no but that may just be me

Did you actually encounter this case? What was the root cause?

jeremylenz: nah, was just trying to be safe. I can remove it :+1:

jeremylenz:
@jturel thanks for the ongoing reviews! I'll get to addressing them tomorrow. In the meantime I've added the new rake task ([item 3 here](https://github.com/Katello/katello/pull/10433#issuecomment-1428564829))

lib/katello/tasks/clean_candlepin_orphaned_products.rake
```diff
@@ -0,0 +1,61 @@
...
+      orphaned_cp_product_ids.each do |cp_product_id|
+        
+        pools = Katello::Resources::Candlepin::Pool.get_for_product(cp_product_id, org.label)
+        pools.each do |pool|
+          pool_subscription_id = pool['subscriptionId']
+          if pool_subscription_id.present?
+            print "Deleting Candlepin orphaned subscription #{pool_subscription_id}\n"
+            deleted_subscriptions_count += 1
+            Katello::Resources::Candlepin::Subscription.destroy(pool_subscription_id)
+          end
```
jeremylenz, line 41: Unlike for pools, Candlepin doesn't have an API endpoint that allows you to retrieve subscriptions by product ID. So I noticed that `subscriptionId` is included in the API response [here](https://petstore.swagger.io/?url=https://raw.githubusercontent.com/candlepin/candlepin/master/api/candlepin-api-spec.yaml#/owner/listOwnerPools), and thought we could use that.

However, `subscriptionId` always seems to be nil. I attached my custom product's subscription to a host and to an AK, but it still doesn't find any subscriptions to delete from Candlepin. Maybe we don't need to clean up subscriptions after all..? Or maybe I'm missing something. :thinking:

app/models/katello/product.rb
jeremylenz: Removed the before_validation hook.

app/models/katello/product.rb
jeremylenz: It was in a test where `provider` wasn't assigned. It was just easier to add one character than it is to update a bunch of test fixtures.

app/models/katello/glue/provider.rb
jeremylenz: for the record, I'm still not sure about this. Even with this new way, if you pass `[]` just to satisfy the required argument, but you actually have orphaned CP products for which you're reimporting their pools, it will still break. Still need to think some more.

jturel: What's the use case for calling import_all without omitting the orphans? Having a clear answer to that might help here?

jeremylenz: The only use case I can think of for calling `import_all` without passing in the orphans is if you don't know about them. If I'm in the process of reimporting products, and I come across an orphaned product, then I know that its associated pools are also orphaned, so I can skip them on import as well. But if I'm only importing pools or subscriptions (like in fix_invalid_pools.rake), I won't know what's valid or not, and my only choice is to try to reimport everything.  :thinking:

jturel: I see what you're saying. It feels like there's a missing layer of abstraction such as a service class that encapsulates the logic around all of that (which itself would call `import_all` for one or more candlepin-backed models) - cause it seems to be spread out and somewhat unclear. Not that you should attempt to solve that here, but not having it forces you to pay attention to changing code in more areas of the app.

lib/katello/tasks/clean_candlepin_orphaned_products.rake
ianballou: Do we delete subscriptions when deleting custom repository / products? If yes, then I suppose we'd need to support cleaning up orphans. Is there any change that Candlepin detects them as orphans themselves and does its own cleanup?

jeremylenz: > Do we delete subscriptions when deleting custom repository / products?

you're a genius!!

Just discovered `Actions::Candlepin::Product::DeletePools` and `Actions::Candlepin::Product::DeleteSubscriptions` - turns out I've been trying to reinvent these and they were there all along. Now my rake task can just run those! I can just pass in the product's cp_id.

ianballou: Hey awesome!

app/models/katello/pool.rb
jeremylenz: okay, good points. :P  Removing.

jeremylenz:
Updated PR description and testing steps.

ianballou (commented):
Where should I see the new error on manifest refresh? What I saw in dynflow was:

```
Validation failed: Subscription can't be blank 
```

Granted, I had some more errors that were unrelated to your issue, but I wasn't seeing the new one you introduced.

app/models/katello/glue/candlepin/candlepin_object.rb
```diff
@@ -12,8 +12,9 @@ def candlepin_records_by_id(organization)
         records_by_id
       end
 
-      def import_candlepin_records(candlepin_objects, org)
+      def import_candlepin_records(candlepin_objects, org, skip_product_ids = [])
         candlepin_objects.each do |object|
+          next if skip_product_ids.include?(object['productId'])
```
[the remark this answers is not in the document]
jeremylenz: Thanks! I think this will end up being a better solution (see latest commit)

jeremylenz:
> Where should I see the new error on manifest refresh?

@ianballou Oh! I removed it based on Jonathon's convincing [here](https://github.com/Katello/katello/pull/10433#discussion_r1106392150). Seeing only the "Subscription can't be blank" is expected now.

jeremylenz:
Refactored the import methods so we don't have to pass around cp_ids to skip.

jturel:
> > Where should I see the new error on manifest refresh?
> 
> @ianballou Oh! I removed it based on Jonathon's convincing [here](https://github.com/Katello/katello/pull/10433#discussion_r1106392150). Seeing only the "Subscription can't be blank" is expected now.

Given all the other changes here I thought we wouldn't be seeing that validation error at all! What are the circumstances that create that situation?

jeremylenz:
> Given all the other changes here I thought we wouldn't be seeing that validation error at all! What are the circumstances that create that situation?

If you've reproduced the issue (by destroying a product and then refreshing a manifest before checking out this branch) and have products in Katello with incorrect providers that you haven't yet cleaned up, I'd expect that to happen. For 'real' users the upgrade rake task should take care of that situation.

lib/katello/tasks/clean_candlepin_orphaned_products.rake
```diff
@@ -0,0 +1,41 @@
+require File.expand_path("../engine", File.dirname(__FILE__))
+
+namespace :katello do
```
jturel, line 3: Would it make sense to invoke this rake task from inside your upgrade task? It doesn't seem mandatory given your other changes but it might be useful. You'll know that any orphaned products that occur will have happened after this code is running which might help you narrow down the root cause of where they come from. ie you wouldn't have to worry about sifting through orphaned products that could be half a decade old. 🤔

jeremylenz: We were actually talking a bit about that before.. Now that you bring up that scenario, it sounds like a good idea to me!

app/models/katello/glue/candlepin/candlepin_object.rb
```diff
@@ -28,6 +28,8 @@ def import_candlepin_record(record:, organization:)
 
         persisted = nil
         Katello::Util::Support.active_record_retry do
+          # new_record = self.where(db_attrs).first_or_initialize
```
ianballou, line 31: Comments

ianballou (commented):
Comments!

app/models/katello/glue/provider.rb
```diff
@@ -123,9 +147,15 @@ def import_products_from_cp
         cp_products = cp_products.select { |prod| Glue::Candlepin::Product.engineering_product_id?(prod['id']) }
 
         prod_content_importer = Katello::ProductContentImporter.new(cp_products)
+        # orphaned_ids = ["559524071636"]
```
ianballou, line 150: Comment

app/models/katello/content_view.rb
```diff
@@ -110,15 +110,15 @@ class ContentView < Katello::Model
...
-    enum generated_for: {
-      none: 0,
-      library_export: 1,
-      repository_export: 2,
-      library_import: 3,
-      repository_import: 4,
-      library_export_syncable: 5,
-      repository_export_syncable: 6
-    }, _prefix: true
+    # enum generated_for: {
```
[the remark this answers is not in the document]
jeremylenz: aack I did `git restore --staged` soooo many times here! and then I missed it this time :smile:

ianballou:
Confirmed that manifest refresh works now with orphaned custom repos.

jturel (approved):
This looks amazing! Thanks for indulging me :)

ACK LGTM etc

jeremylenz:
:green_apple:

ianballou:
The upgrade rake task is failing for me:

```
rake aborted!
ArgumentError: You tried to define an enum named "generated_for" on the model "Katello::ContentView", but this will generate a instance method "generated_for_none?", which is already defined by another enum.
/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activerecord-6.1.7.2/lib/active_record/enum.rb:294:in `raise_conflict_error'
/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activerecord-6.1.7.2/lib/active_record/enum.rb:289:in `detect_enum_conflict!'
/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activerecord-6.1.7.2/lib/active_record/enum.rb:234:in `define_enum_methods'
/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activerecord-6.1.7.2/lib/active_record/enum.rb:207:in `block (3 levels) in enum'
/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activerecord-6.1.7.2/lib/active_record/enum.rb:201:in `each_pair'
/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activerecord-6.1.7.2/lib/active_record/enum.rb:201:in `each'
/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activerecord-6.1.7.2/lib/active_record/enum.rb:201:in `block (2 levels) in enum'
/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activerecord-6.1.7.2/lib/active_record/enum.rb:187:in `module_eval'
/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activerecord-6.1.7.2/lib/active_record/enum.rb:187:in `block in enum'
/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activerecord-6.1.7.2/lib/active_record/enum.rb:166:in `each'
/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activerecord-6.1.7.2/lib/active_record/enum.rb:166:in `enum'
/home/vagrant/katello/app/models/katello/content_view.rb:113:in `<class:ContentView>'
/home/vagrant/katello/app/models/katello/content_view.rb:3:in `<module:Katello>'
/home/vagrant/katello/app/models/katello/content_view.rb:1:in `<main>'
/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/bootsnap-1.16.0/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:32:in `require'
/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/bootsnap-1.16.0/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:32:in `require'
/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/polyglot-0.3.5/lib/polyglot.rb:65:in `require'
/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activesupport-6.1.7.2/lib/active_support/dependencies.rb:332:in `block in require'
/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activesupport-6.1.7.2/lib/active_support/dependencies.rb:299:in `load_dependency'
/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activesupport-6.1.7.2/lib/active_support/dependencies.rb:332:in `require'
/home/vagrant/katello/lib/katello/tasks/clean_candlepin_orphaned_products.rake:7:in `block (3 levels) in <main>'
/home/vagrant/katello/lib/katello/tasks/clean_candlepin_orphaned_products.rake:7:in `each'
/home/vagrant/katello/lib/katello/tasks/clean_candlepin_orphaned_products.rake:7:in `block (2 levels) in <main>'
/home/vagrant/katello/lib/katello/tasks/upgrades/4.8/fix_incorrect_providers.rake:25:in `block (4 levels) in <main>'
/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/rake-13.0.6/exe/rake:27:in `<top (required)>'
/usr/share/gems/gems/bundler-2.2.24/lib/bundler/cli/exec.rb:63:in `load'
/usr/share/gems/gems/bundler-2.2.24/lib/bundler/cli/exec.rb:63:in `kernel_load'
/usr/share/gems/gems/bundler-2.2.24/lib/bundler/cli/exec.rb:28:in `run'
/usr/share/gems/gems/bundler-2.2.24/lib/bundler/cli.rb:475:in `exec'
/usr/share/gems/gems/bundler-2.2.24/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
/usr/share/gems/gems/bundler-2.2.24/lib/bundler/vendor/thor/lib/thor/invocation.rb:127:in `invoke_command'
/usr/share/gems/gems/bundler-2.2.24/lib/bundler/vendor/thor/lib/thor.rb:392:in `dispatch'
/usr/share/gems/gems/bundler-2.2.24/lib/bundler/cli.rb:31:in `dispatch'
/usr/share/gems/gems/bundler-2.2.24/lib/bundler/vendor/thor/lib/thor/base.rb:485:in `start'
/usr/share/gems/gems/bundler-2.2.24/lib/bundler/cli.rb:25:in `start'
/usr/share/gems/gems/bundler-2.2.24/libexec/bundle:46:in `block in <top (required)>'
/usr/share/gems/gems/bundler-2.2.24/lib/bundler/friendly_errors.rb:128:in `with_friendly_errors'
/usr/share/gems/gems/bundler-2.2.24/libexec/bundle:34:in `<top (required)>'
/usr/local/bin/bundle:23:in `load'
/usr/local/bin/bundle:23:in `<main>'
Tasks: TOP => katello:clean_candlepin_orphaned_products
(See full trace by running task with --trace)
```

jeremylenz:
This is the reason I was commenting out that enum. thought it was just me!

I don't know what's going on there..

lib/katello/tasks/clean_candlepin_orphaned_products.rake
```diff
@@ -0,0 +1,41 @@
...
+  desc "Removes custom products (and their associated pools and subscriptions) from Candlepin that are not in Katello"
+  task :clean_candlepin_orphaned_products => ["dynflow:client", "environment"] do
+    User.current = User.anonymous_admin #set a user for orchestration
+    Dir.glob(Katello::Engine.root.to_s + '/app/models/katello/*.rb').each { |file| require file }
```
ianballou, line 7: Why is this required? It seems to be the cause of the enum error.

jeremylenz: I thought I removed that already! I bet you're right, that's making it try to load everything twice..

jeremylenz: It was in the `katello:reimport` task which I used as a template for this. I didn't really think much about it.

jeremylenz:
@ianballou Squashed and updated!

jeremylenz:
:green_apple: :green_apple:

lib/katello/tasks/reimport.rake
```diff
@@ -32,7 +32,7 @@ namespace :katello do
       ack_key.import_pools
     end
```
ianballou, line 34: It would also be worth removing the `Dir.glob(Katello::Engine.root.to_s + '/app/models/katello/*.rb').each { |file| require file }` from this file. It errors out as well!

jeremylenz: Removed.

ianballou (approved):
Looking good to me!

jeremylenz:
[test katello]

app/models/katello/glue/provider.rb
```diff
@@ -144,13 +173,13 @@ def import_product(product_json)
...
         elsif product.nil?
-          product = Glue::Candlepin::Product.import_from_cp(product_json, organization)
+          product = Glue::Candlepin::Product.import_redhat_product_from_cp(product_json, organization)
         end
         product
       end
 
       def index_subscriptions
-        Katello::Subscription.import_all(organization)
+        Katello::Subscription.import_all(organization, nil)
```
jturel, line 182: unnecessary?

app/models/katello/glue/provider.rb
```diff
@@ -123,9 +147,14 @@ def import_products_from_cp
         cp_products = cp_products.select { |prod| Glue::Candlepin::Product.engineering_product_id?(prod['id']) }
 
         prod_content_importer = Katello::ProductContentImporter.new(cp_products)
+        orphaned_ids = []
 
         Katello::Logging.time("Imported #{cp_products.size} products") do
           cp_products.each do |product_json|
+            if ::Katello::Glue::Provider.orphaned_custom_product?(product_json['id'], organization)
+              orphaned_ids << product_json['id']
```
jturel, line 155: looks like these ids arent being used now

lib/katello/tasks/clean_candlepin_orphaned_products.rake
```diff
@@ -0,0 +1,40 @@
+require File.expand_path("../engine", File.dirname(__FILE__))
```
jturel, line 1: don't need this (right?)

jeremylenz: Apparently not!

app/models/katello/product.rb
jeremylenz: oh no I typed up a long comment here and now I lost it, let me know if you don't see it!

jeremylenz: Found it: https://github.com/Katello/katello/commit/792f3b464d57e81c650d549686f24d6e6070a831#r101064846

jeremylenz:
Thanks @ianballou for testing, and thanks @jturel for the review! good to see your insightful commentary again :slightly_smiling_face: 

Don't be a stranger!

PR #10435, opened 2023-02-02
Fixes #36038 - Add validations for ACS creation and update in hammer

qcjames53:
Previously, alternate content sources would silently sanitize certain fields in create and update requests through hammer, depending on the ACS type (custom vs simplified vs rhui). The request would go through, using only the applicable fields for the ACS type. These changes now cause these requests to fail and return appropriate error messages. Alternate Content Sources now will also disallow modifying the content-type or alternate-content-source-type fields once created.

Additionally, tests have been added to check this functionality is handled appropriately in the future.

The controller now passes all inputs to the model layer for validation during creation and for most params during updates. However, due to how --product-ids is handled, logic for rejecting these during update needed to be handled at the controller level. If supplied on an invalid type of ACS, this logic manually throws UnprocessableEntity.

Testing creation: Using hammer, attempt to create a simplified ACS or custom ACS using restricted arguments. Creation should fail.
Testing update: Create an ACS. Using hammer, attempt to update the ACS using restricted args. The update should fail.

Restricted arguments for custom/rhui ACS: 

- --product-ids

Restricted arguments for simplified ACS: 

- --base-url
- --ssl-ca-cert-id
- --ssl-client-cert-id
- --subpaths
- --upstream-password
- --upstream username
- --verify-ssl

Additionally please attempt to update the alternate-content-source-type and the content-type using curl. Steps below:
```
curl -sku admin:password -H "Content-type: application/json" -X PUT -d '{"content_type": "file"}' https://$(hostname)/katello/api/alternate_content_sources/191
```

The update should fail:

### DIFF
--- a/app/controllers/katello/api/v2/alternate_content_sources_controller.rb
+++ b/app/controllers/katello/api/v2/alternate_content_sources_controller.rb
@@ -88,9 +88,6 @@ module Katello
         find_smart_proxies
       end
 
-      # Check for invalid params
-      check_params_for_invalid_updates
-
       if params[:product_ids].nil?
         @products = @alternate_content_source.products
       elsif params[:product_ids] == []
@@ -129,21 +126,6 @@ module Katello
       ]
 
       params.require(:alternate_content_source).permit(*keys).to_h.with_indifferent_access
-    end
-
-    def check_params_for_invalid_updates
-      # Check parameters which cannot be validated at the model level, throwing
-      # errors where neccessary
-
-      # Check that the combination of params[:product_ids] and ACS type is allowed:
-      #            | simplified | custom  | rhui
-      # -----------+------------+---------+---------
-      # nil        | ok         | ok      | ok
-      # []         | ok         | invalid | invalid
-      # [foo, ...] | ok         | invalid | invalid
-      unless @alternate_content_source&.simplified? || params[:product_ids].nil?
-        (fail HttpErrors::UnprocessableEntity, "Products must remain blank for ACS of type #{@alternate_content_source&.alternate_content_source_type}")
-      end
     end
 
     def find_smart_proxies
--- a/app/controllers/katello/api/v2/alternate_content_sources_controller.rb
+++ b/app/controllers/katello/api/v2/alternate_content_sources_controller.rb
@@ -70,7 +70,6 @@ module Katello
     param_group :acs
     def create
       @alternate_content_source = ::Katello::AlternateContentSource.new(acs_params.except(:smart_proxy_ids, :smart_proxy_names, :product_ids))
-      @alternate_content_source.verify_ssl = nil if @alternate_content_source.simplified?
       sync_task(::Actions::Katello::AlternateContentSource::Create, @alternate_content_source, @smart_proxies, @products)
       @alternate_content_source.reload
       respond_for_create(resource: @alternate_content_source)
@@ -88,6 +87,9 @@ module Katello
       else
         find_smart_proxies
       end
+
+      # Check for invalid params
+      check_params_for_invalid_updates
 
       if params[:product_ids].nil?
         @products = @alternate_content_source.products
@@ -117,10 +119,30 @@ module Katello
     protected
 
     def acs_params
-      keys = [:name, :label, :description, {smart_proxy_ids: []}, {smart_proxy_names: []}, :content_type, :alternate_content_source_type, :use_http_proxies]
-      keys += [:base_url, {subpaths: []}, :upstream_username, :upstream_password, :ssl_ca_cert_id, :ssl_client_cert_id, :ssl_client_key_id, :verify_ssl] if params[:action] == 'create' || @alternate_content_source&.custom? || @alternate_content_source&.rhui?
-      keys += [{product_ids: []}] if params[:action] == 'create' || @alternate_content_source&.simplified?
+      # In the past, this method flitered acs keys depending on what type the
+      # acs repo was. This caused silent failures. We're now passing everything
+      # to the model layer to mitigate.
+      keys = [
+        :name, :label, :description, {smart_proxy_ids: []}, {smart_proxy_names: []}, :content_type,
+        :alternate_content_source_type, :use_http_proxies, :base_url, {subpaths: []}, :upstream_username,
+        :upstream_password, :ssl_ca_cert_id, :ssl_client_cert_id, :ssl_client_key_id, :verify_ssl, {product_ids: []}
+      ]
+
       params.require(:alternate_content_source).permit(*keys).to_h.with_indifferent_access
+    end
+
+    def check_params_for_invalid_updates
+      # Check parameters which cannot be validated at the model level, throwing
+      # errors where neccessary
+
+      # Disallow users from updating ACS type or content type: these should be static
+      (fail HttpErrors::UnprocessableEntity, "Content type cannot be modified once ACS is created") if params[:content_type].nil?
+      (fail HttpErrors::UnprocessableEntity, "ACS type cannot be modified once ACS is created") if params[:alternate_content_source_type].nil?
+
+      # Check that this acs is simplified before allowing products to be cleared / updated
+      unless @alternate_content_source&.simplified? || params[:product_ids].nil?
+        (fail HttpErrors::UnprocessableEntity, "Products must remain blank for ACS of type #{@alternate_content_source&.alternate_content_source_type}")
+      end
     end
 
     def find_smart_proxies
--- a/app/controllers/katello/api/v2/alternate_content_sources_controller.rb
+++ b/app/controllers/katello/api/v2/alternate_content_sources_controller.rb
@@ -87,7 +87,6 @@ module Katello
       else
         find_smart_proxies
       end
-
       if params[:product_ids].nil?
         @products = @alternate_content_source.products
       elsif params[:product_ids] == []
@@ -95,6 +94,7 @@ module Katello
       else
         find_products
       end
+
       sync_task(::Actions::Katello::AlternateContentSource::Update, @alternate_content_source, @smart_proxies, @products, acs_params.except(:smart_proxy_ids, :smart_proxy_names, :product_ids))
       respond_for_show(:resource => @alternate_content_source)
     end
--- a/app/controllers/katello/api/v2/alternate_content_sources_controller.rb
+++ b/app/controllers/katello/api/v2/alternate_content_sources_controller.rb
@@ -116,9 +116,6 @@ module Katello
     protected
 
     def acs_params
-      # In the past, this method flitered acs keys depending on what type the
-      # acs repo was. This caused silent failures. We're now passing everything
-      # to the model layer to mitigate.
       keys = [
         :name, :label, :description, {smart_proxy_ids: []}, {smart_proxy_names: []}, :content_type,
         :alternate_content_source_type, :use_http_proxies, :base_url, {subpaths: []}, :upstream_username,
--- a/app/controllers/katello/api/v2/alternate_content_sources_controller.rb
+++ b/app/controllers/katello/api/v2/alternate_content_sources_controller.rb
@@ -135,11 +135,12 @@ module Katello
       # Check parameters which cannot be validated at the model level, throwing
       # errors where neccessary
 
-      # Disallow users from updating ACS type or content type: these should be static
-      (fail HttpErrors::UnprocessableEntity, "Content type cannot be modified once ACS is created") if params[:content_type].nil?
-      (fail HttpErrors::UnprocessableEntity, "ACS type cannot be modified once ACS is created") if params[:alternate_content_source_type].nil?
-
-      # Check that this acs is simplified before allowing products to be cleared / updated
+      # Check that the combination of params[:product_ids] and ACS type is allowed:
+      #            | simplified | custom  | rhui
+      # -----------+------------+---------+---------
+      # nil        | ok         | ok      | ok
+      # []         | ok         | invalid | invalid
+      # [foo, ...] | ok         | invalid | invalid
       unless @alternate_content_source&.simplified? || params[:product_ids].nil?
         (fail HttpErrors::UnprocessableEntity, "Products must remain blank for ACS of type #{@alternate_content_source&.alternate_content_source_type}")
       end
--- a/app/lib/actions/katello/alternate_content_source/create.rb
+++ b/app/lib/actions/katello/alternate_content_source/create.rb
@@ -7,7 +7,9 @@ module Actions
         def plan(acs, smart_proxies, products = nil)
           acs.save!
           action_subject(acs)
-          acs.products << products if products.present?
+          if products.present?
+            acs.update!(products: products)
+          end
           smart_proxies = smart_proxies.present? ? smart_proxies.uniq : []
           concurrence do
             smart_proxies.each do |smart_proxy|
--- a/app/lib/actions/katello/alternate_content_source/update.rb
+++ b/app/lib/actions/katello/alternate_content_source/update.rb
@@ -19,12 +19,14 @@ module Actions
           products_to_associate = []
           products_to_disassociate = []
 
-          products = products.uniq
-          products_to_associate = products - acs.products
-          products_to_disassociate = acs.products - products
-          old_product_ids = acs.products.pluck(:id)
-          acs.products = products
-          acs.audit_updated_products(old_product_ids) unless products_to_associate.empty? && products_to_disassociate.empty?
+          if products.present?
+            products = products.uniq
+            products_to_associate = products - acs.products
+            products_to_disassociate = acs.products - products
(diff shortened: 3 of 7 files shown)


jeremylenz:
[test katello]

jeremylenz (commented):
Congrats @qcjames53 on your first Katello PR! :tada: 

I'm seeing some weirdness when trying to create via hammer:

On master:
```
$ hammer alternate-content-source create --alternate-content-source-type simplified
Could not create the Alternate Content Source.:
  PG::NotNullViolation: ERROR:  null value in column "name" violates not-null constraint
  DETAIL:  Failing row contains (14, null, null, null, null, null, null, null, {}, yum, simplified, null, null, null, null).
```
It seems there's a non-null constraint for the "name" column, but no validation that says "name" must be present. Maybe we should add that?

Still on master:
```
$ hammer alternate-content-source create --alternate-content-source-type simplified --name myACS
Alternate Content Source created.
```

On this branch:
```
$ hammer alternate-content-source create --alternate-content-source-type simplified --name myACS2
Could not create the Alternate Content Source.:
  Validation failed: Verify ssl must be blank
```

I'm not sure why it's saying this, since I didn't pass a `verify_ssl` param?

jeremylenz:
[test katello]

app/controllers/katello/api/v2/alternate_content_sources_controller.rb
```diff
@@ -92,10 +91,17 @@ def update
       if params[:product_ids].nil?
         @products = @alternate_content_source.products
       elsif params[:product_ids] == []
+        # Check that this acs is simplified before allowing products to be cleared
+        fail HttpErrors::UnprocessableEntity, _("Products must remain blank for acs of type #{@alternate_content_source&.alternate_content_source_type}") unless @alternate_content_source&.simplified?
```
jeremylenz, line 95:
```suggestion
        fail HttpErrors::UnprocessableEntity, _("Products must remain blank for ACS of type #{@alternate_content_source&.alternate_content_source_type}") unless @alternate_content_source&.simplified?
```

app/controllers/katello/api/v2/alternate_content_sources_controller.rb
```diff
@@ -92,10 +91,17 @@ def update
...
+
         @products = []
       else
+        # Check that this acs is simplified before allowing products to be added
+        fail HttpErrors::UnprocessableEntity, _("Products must remain blank for acs of type #{@alternate_content_source&.alternate_content_source_type}") unless @alternate_content_source&.simplified?
```
jeremylenz, line 100:
```suggestion
        fail HttpErrors::UnprocessableEntity, _("Products must remain blank for ACS of type #{@alternate_content_source&.alternate_content_source_type}") unless @alternate_content_source&.simplified?
```

jeremylenz (commented):
Looking good! The behavior I'm seeing in Hammer makes sense now.

Some extra credit:
```
Could not create the Alternate Content Source.:
  Validation failed: Base url can't be blank, Verify ssl is reserved
```

"Verify ssl is reserved" isn't a very useful message. Maybe add a custom message "is required" for that case?

One small capitalization issue below, other than that LGTM

jeremylenz:
[test katello]

jeremylenz:
test failures look related!

Let me know if you need help with either of them

app/controllers/katello/api/v2/alternate_content_sources_controller.rb
```diff
@@ -92,10 +91,17 @@ def update
       if params[:product_ids].nil?
         @products = @alternate_content_source.products
       elsif params[:product_ids] == []
+        # Check that this acs is simplified before allowing products to be cleared
```
ianballou, line 94: Hopefully this doesn't complicated things too much, but we should also disallow users from changing the alternate content source type or the content type on update. It would require recreating all of the backend Pulp entities and the base url + subpaths would likely no longer make sense.

ianballou (commented):
One bonus request :)

app/controllers/katello/api/v2/alternate_content_sources_controller.rb
```diff
@@ -134,6 +131,20 @@ def acs_params
       params.require(:alternate_content_source).permit(*keys).to_h.with_indifferent_access
     end
 
+    def check_params_for_invalid_updates
+      # Check parameters which cannot be validated at the model level, throwing
+      # errors where neccessary
```
jeremylenz, line 136: Why can't these be validated at the model level?

app/controllers/katello/api/v2/alternate_content_sources_controller.rb
qcjames53: Features added :)

app/controllers/katello/api/v2/alternate_content_sources_controller.rb
qcjames53: Fixed, thank you :)

app/controllers/katello/api/v2/alternate_content_sources_controller.rb
qcjames53: Fixed, thanks!

app/controllers/katello/api/v2/alternate_content_sources_controller.rb
ianballou: I think the example you brought up could work very well here (https://github.com/Katello/katello/blob/master/app/models/katello/provider.rb#L48-L56).  If that does indeed work, it would be much nicer to get all of this out of the controller!

qcjames53: Thank you for the suggestion + resources. Implemented the necessary changes.

app/controllers/katello/api/v2/alternate_content_sources_controller.rb
```diff
@@ -117,12 +119,33 @@ def refresh
...
+
+      # Check that the combination of params[:product_ids] and ACS type is allowed:
+      #            | simplified | custom  | rhui
+      # -----------+------------+---------+---------
+      # nil        | ok         | ok      | ok
+      # []         | ok         | invalid | invalid
+      # [foo, ...] | ok         | invalid | invalid
+      unless @alternate_content_source&.simplified? || params[:product_ids].nil?
+        (fail HttpErrors::UnprocessableEntity, "Products must remain blank for ACS of type #{@alternate_content_source&.alternate_content_source_type}")
```
jeremylenz, line 145: I'm not sure how Rails deals with associations in cases like this. Can this be removed as well and just rely on the existing validation in the model `validates :products, if: -> { custom? || rhui? }, absence: true` ? Or does that not work?

If we have to rely on the controller, it's not ideal because that would prevent changes in the API, but you could still do something like 
```
my_acs.product_ids = [42, 43]
```
in Rails console and it wouldn't complain.

qcjames53:
[test katello]

qcjames53:
[test react]

ianballou:
The snapshot issues can be fixed with:

```
git fetch upstream
git rebase upstream/master
git push -ff <remote> [36038-simplified-acs-validations](https://github.com/qcjames53/katello/tree/36038-simplified-acs-validations)
```

Tested on my local development environment.

app/controllers/katello/api/v2/alternate_content_sources_controller.rb
jeremylenz: @qcjames53 This is the only unresolved thread here. It still seems like this could be covered with a custom validation in the model. were you able to try this?

app/lib/actions/katello/alternate_content_source/create.rb
```diff
@@ -7,7 +7,9 @@ class Create < Actions::EntryAction
         def plan(acs, smart_proxies, products = nil)
           acs.save!
           action_subject(acs)
-          acs.products << products if products.present?
+          if products.present?
+            acs.update!(products: products)
+          end
```
jeremylenz, line 12: If `acs` previously had products,
```rb
acs.products << products
```
would add `products` to its existing products.

But 
```rb
acs.update!(products: products)
```
won't preserve existing products. Is this intended?

app/controllers/katello/api/v2/alternate_content_sources_controller.rb
```diff
@@ -117,9 +116,15 @@ def refresh
     protected
 
     def acs_params
-      keys = [:name, :label, :description, {smart_proxy_ids: []}, {smart_proxy_names: []}, :content_type, :alternate_content_source_type, :use_http_proxies]
-      keys += [:base_url, {subpaths: []}, :upstream_username, :upstream_password, :ssl_ca_cert_id, :ssl_client_cert_id, :ssl_client_key_id, :verify_ssl] if params[:action] == 'create' || @alternate_content_source&.custom? || @alternate_content_source&.rhui?
-      keys += [{product_ids: []}] if params[:action] == 'create' || @alternate_content_source&.simplified?
+      # In the past, this method flitered acs keys depending on what type the
+      # acs repo was. This caused silent failures. We're now passing everything
+      # to the model layer to mitigate.
```
jeremylenz, line 121: I'd say this comment isn't necessary

app/lib/actions/katello/alternate_content_source/update.rb
```diff
@@ -19,14 +19,14 @@ def plan(acs, smart_proxies, products, acs_params)
...
 
-          if acs.simplified?
-            products = products.uniq
-            products_to_associate = products - acs.products
-            products_to_disassociate = acs.products - products
-            old_product_ids = acs.products.pluck(:id)
-            acs.products = products
-            acs.audit_updated_products(old_product_ids) unless products_to_associate.empty? && products_to_disassociate.empty?
-          end
+          products = products.uniq
```
jeremylenz, line 22:
```suggestion
          products.uniq!
```

ianballou, line 22: We should still have an `if` statement, how about a check first if products were given at all? 

```ruby
if products.present?
...
```

app/models/katello/alternate_content_source.rb
```diff
@@ -113,5 +120,15 @@ def audit_updated_products(old_product_ids)
     def self.humanize_class_name
       "Alternate Content Sources"
     end
+
+    # Disallow static properties from being modified on update
+    def constraint_acs_update
```
ianballou, line 125: Nice.

app/controllers/katello/api/v2/alternate_content_sources_controller.rb
```diff
@@ -88,14 +87,14 @@ def update
       else
         find_smart_proxies
       end
-
```
ianballou, line 91: Try to avoid having random "style" changes in a PR.

ianballou (commented):
Some small comments. Looking pretty good I think!

jeremylenz (commented):
thanks @qcjames53 ! Just a couple more nits

ianballou:
Also you might want to add an instruction to test changing ACS type and content type since you made that nice addition.

app/lib/actions/katello/alternate_content_source/create.rb
qcjames53: This is happening on create so I don't believe either approach would make a difference?

app/controllers/katello/api/v2/alternate_content_sources_controller.rb
qcjames53: ok

app/lib/actions/katello/alternate_content_source/update.rb
qcjames53: Thank you

app/lib/actions/katello/alternate_content_source/create.rb
jeremylenz: ok, sounds good!

Does `update!` work with `products` as an attribute like that? Since it's an association and not a column on the alternate_content_sources table, I wonder if `products = ` isn't more correct. But I may just be nitpicking. If it works I'm fine with either.

app/lib/actions/katello/alternate_content_source/update.rb
qcjames53: Would you mind explaining a little more about what this is accomplishing?

app/controllers/katello/api/v2/alternate_content_sources_controller.rb
qcjames53: Good to know, thanks!

app/lib/actions/katello/alternate_content_source/create.rb
qcjames53: `update!` is used so we can validate the list of products. `acs.products << products` and `acs.products = products` wouldn't validate until another `acs.save!` was called.

jeremylenz: ahhh good point! Carry on then :)

app/lib/actions/katello/alternate_content_source/update.rb
ianballou: Yeah, so as you have it, the code to calculate the products to add and remove will be run regardless of  whether or not the product list is empty.  That means those CPU cycles are being wasted.

The code leading up to `acs.save!` could be put in an `if` statement that checks if there are any products at all.

ianballou: The specific code block:

```ruby
          products = products.uniq
          products_to_associate = products - acs.products
          products_to_disassociate = acs.products - products
          old_product_ids = acs.products.pluck(:id)
          acs.products = products
          acs.audit_updated_products(old_product_ids) unless products_to_associate.empty? && products_to_disassociate.empty?
```

qcjames53: Have to deny this change because it unfortunately breaks the program. Went down a bit of a rabbit hole trying to figure out why; If you have time for a meeting I can explain.

jeremylenz: No worries!

ianballou (commented):
I found an issue with active record: validating for absence with booleans also accepts false. This is funky, but you can add the following validation to fix it:

```ruby
    validates :verify_ssl,  if: :simplified?, inclusion: { in: [nil] }
```

And then remove the `verify_ssl` absence check from the simplified validation.

test/models/katello/alternate_content_source_test.rb
```diff
@@ -115,6 +115,24 @@ def test_with_type
...
+      assert_raises(ActiveRecord::RecordInvalid) do
+        @simplified_acs.update!(alternate_content_source_type: "custom")
+      end
+      assert_raises(ActiveRecord::RecordInvalid) do
+        @rhui_acs.update!(alternate_content_source_type: "simplified")
+      end
+    end
+
+    def test_update_content_type
+      assert_raises(ActiveRecord::RecordInvalid) do
```
ianballou, line 129: I don't think it's strictly necessary, but you can match the error message too:

```ruby
        exception = assert_raises(RuntimeError) do
          plan_action(action, organization: organization, path: path, metadata: metadata)
        end
        assert_match(/'#{content_view_version.name}' already exists/, exception.message)
```

jeremylenz: I like things like this! This way if the record is invalid for an unrelated reason, the test will still do what it's supposed to.

ianballou (approved):
Looks good to me!

app/models/katello/alternate_content_source.rb
```diff
@@ -29,13 +29,18 @@ class AlternateContentSource < Katello::Model
...
     validates :products, if: -> { custom? || rhui? }, absence: true
     validates :label, :uniqueness => true
     validates :name, :uniqueness => true, presence: true
-    validates :verify_ssl, if: :custom?, exclusion: [nil]
+    # verify ssl must be validated this way due to presence: <bool> failing on a value of false
+    validates :verify_ssl, if: -> { custom? || rhui? }, inclusion: {
+      in: [true, false],
+      message: "can't be blank"
+    }
+    validates :verify_ssl, if: :simplified?, inclusion: { in: [nil] }
```
jeremylenz, line 43: This results in a message "Verify ssl is not included in the list." can you add a custom message "must be blank" ?

jeremylenz (commented):
```
[9] pry(main)> Katello::AlternateContentSource.all.first.update!(alternate_content_source_type: "simplified", content_type: "file")
ActiveRecord::RecordInvalid: Validation failed: Base url must be blank, Verify ssl is not included in the list, Content type cannot be modified once an ACS is created, Alternate content source type cannot be modified once an ACS is created
```

Works great!  One more small item below

app/models/katello/alternate_content_source.rb
qcjames53: Fixed!

jeremylenz (approved):
Nice work @qcjames53! thanks for putting up with the long review cycle :)

APJ :+1:

PR #10437, opened 2023-02-03
Fixes #36043 - Assign content view correctly when reassigning hosts

jeremylenz:

When deleting a content view version that requires reassigning hosts, you get
```
NoMethodError: undefined method `content_view=' for #<Katello::Host::ContentFacet:0x00005614daf63278>
```
This is because it's still trying to set the content view the old way. Instead, `assign_single_environment` should be called.

not sure how we missed this one :man_shrugging: 

1. Make a content view and publish it
2. Assign host(s) to the CV
3. Delete the latest version, and select to reassign the hosts to a different CV
4. observe error or lack thereof

### DIFF
--- a/app/lib/actions/katello/host/reassign.rb
+++ b/app/lib/actions/katello/host/reassign.rb
@@ -3,8 +3,10 @@ module Actions
     module Host
       class Reassign < Actions::Base
         def plan(host, content_view_id, environment_id)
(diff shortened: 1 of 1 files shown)


jeremylenz:
hm, I didn't hit that. Do you know what caused it?

chris1984:
Tested on Production and didn't hit the error, Merge away. I will reset my dev box.

PR #10438, opened 2023-02-03
Fixes #36042 - RHUI ACS content type validation

qcjames53:
RHUI-type ACS will now reject creation and update attempts where the content-type is set to 'file', as they only support 'yum'-type content. Added testing to ensure this functionality survives updates.

N/A

1. Attempt to create a RHUI ACS via hammer or API with --content-type "file":

```
hammer alternate-content-source create --alternate-content-source-type "rhui" --content-type "file" --name "Test RHUI hammer" --base-url "https://rhui.example.com/pulp/content" --subpaths "test/repo1/,test/repo2/" --smart-proxy-ids 1
```

The creation should fail:

```
 Could not create the Alternate Content Source.:
   Validation failed: Content type 'file' is not valid for RHUI ACS
```

2. Attempt to update an existing 'yum' RHUI ACS to type 'file':

```
curl -sku admin:password -H "Content-type: application/json" -X PUT -d '{"content_type": "file"}' https://$(hostname)/katello/api/alternate_content_sources/191
```

The update should fail:

```
{"displayMessage":"Validation failed: Content type 'file' is not valid for RHUI ACS","errors":{"content_type":["'file' is not valid for RHUI ACS"]}}
```

### DIFF
--- a/app/models/katello/alternate_content_source.rb
+++ b/app/models/katello/alternate_content_source.rb
@@ -44,7 +44,13 @@ module Katello
     validates :content_type, inclusion: {
       in: ->(_) { RepositoryTypeManager.defined_repository_types.keys & CONTENT_TYPES },
       allow_blank: false,
-      message: ->(_, _) { _("is not allowed for ACS. Must be one of the following: %s") % (RepositoryTypeManager.defined_repository_types.keys & CONTENT_TYPES).join(',') }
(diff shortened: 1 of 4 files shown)


jeremylenz:
[test katello]

ianballou:
@qcjames53 I think there may be some extra commits in here. You can use the interactive rebase to delete those extra commits. Then you'll just have to deal with the conflicts if there are any (likely not!).

ianballou (approved):
This is looking great!

PR #10440, opened 2023-02-06
Fixes #35963 - Change empty state on blank manifest

chris1984:

* The new workflow is to create a manifest on console.redhat.com
* Import that then add subscriptions
* Before you would get a message saying the manifest has not been imported even though it was
* This changes the empty state by looking at `isManifestImported` and changing the message based on that and removing the import button if an empty manifest is imported.

* Making sure the manifest page didn't break

* Create an empty manifest on console.redhat.com or use mine(pm me for it)
* Import that and verify the message changes and the import button does not show up anymore.
* Delete the manifest to make sure the old message comes up

Pics of standard message:
nomanifest

Pic with PR and empty manifest:
emptymanifest

### DIFF
--- a/webpack/scenes/Subscriptions/SubscriptionsPage.js
+++ b/webpack/scenes/Subscriptions/SubscriptionsPage.js
@@ -201,14 +201,23 @@ class SubscriptionsPage extends Component {
       loadTableColumns({ columns: getEnabledColumns(columns) });
     };
     const columns = subscriptions.selectedTableColumns;
-    const emptyStateData = {
-      header: __('There are no Subscriptions to display'),
-      description: __('Import a Manifest to manage your Entitlements.'),
-      action: {
-        onClick: () => openManageManifestModal(),
-        title: __('Import a Manifest'),
-      },
-    };
+    const emptyStateData = isManifestImported
+      ? {
+        header: __('There are no Subscriptions to display'),
+        description: __('Add Subscriptions using the Add Subscriptions button.'),
+        action: {
+          title: __('Add subscriptions'),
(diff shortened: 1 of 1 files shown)


jeremylenz:
I think it's preferable to have an action button inside the empty state no matter what! But it should just go to 'Add Subscriptions' or 'Import a Manifest' as appropriate. cc @MariSvirik

chris1984:
Button added that functions the same as the PF3 `Add Subscriptions` button to the empty state.

webpack/scenes/Subscriptions/SubscriptionsPage.js
```diff
@@ -209,14 +209,23 @@ class SubscriptionsPage extends Component {
...
-      description: __('Import a Manifest to manage your Entitlements.'),
-      action: {
-        onClick: () => openManageManifestModal(),
-        title: __('Import a Manifest'),
-      },
-    };
+    const emptyStateData = isManifestImported
+      ? {
+        header: __('There are no Subscriptions to display'),
+        description: __('Empty manifest imported, add Subscriptions using the Add Subscriptions button.'),
```
jeremylenz, line 215:
```suggestion
        description: __('Add subscriptions using the Add Subscriptions button.'),
```

jeremylenz (commented):
Works as expected, just one wording comment and then it's good to go

webpack/scenes/Subscriptions/SubscriptionsPage.js
chris1984: @jeremylenz updated the wording

jeremylenz (approved):

PR #10441, opened 2023-02-07
Fixes #36049 - support for issued date in applied report template

snagoor:
Added support for `issued` Erratum date for `Host - Applied Errata` report template.


chris1984 (approved):
ACK, looks good

PR #10442, opened 2023-02-07
Fixes #36053 - Re-enable test sync skipped srpms

wbclark:
https://pulp.plan.io/issues/8812 no longer
applies, so the test is re-enabled.


chris1984 (approved):
ACK, looks fine.

PR #10447, opened 2023-02-10
Fixes #36074 - Saving RHUI alternate content source with a malformed Base URL is possible

ianballou:

Fixes up the ACS create UI for RHUI.  Ensures that a bad RHUI Base URL cannot be passed through. Also corrects the URL validation when editing RHUI ACSs.

There was a funny bug with Wizard steps being skipped. For some reason, the `startAtStep` value was being edited. When I removed that, it fixed the weird skipping.

1) Create a RHUI ACS and check that the enabling of the "Next" button makes sense
2) Same a (1) but also check the steps that can be skipped to.
3) Check that editing a RHUI ACS validates the URL correctly (it should check for the `/pulp/content` at the end)
4) Check all of the above for the other ACS types
5) Try editing other ACS types

### DIFF
--- a/webpack/scenes/AlternateContentSources/Create/ACSCreateWizard.js
+++ b/webpack/scenes/AlternateContentSources/Create/ACSCreateWizard.js
@@ -51,15 +51,17 @@ const ACSCreateWizard = ({ show, setIsOpen }) => {
     [dispatch],
   );
 
+  const subPathValidated = areSubPathsValid(subpaths) ? 'default' : 'error';
+  const urlValidated = (url === '' || isValidUrl(url, acsType)) ? 'default' : 'error';
+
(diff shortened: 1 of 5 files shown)


sjha4:
There's a small weirdness with active steps. If you add a name, smart proxy, go back and remove name, some steps remain active. 
The `enableNext` and `canJumpTo` checks on the create wizard will need to check for all previous steps.

Screenshot from 2023-02-13 05-53-24

ianballou:
@sjha4 should be all good now!

sjha4 (approved):
Tested malformed URL for RHUI ACS.
Tested enabling/disabling of steps based on entered/cleared values on different steps.
Ack 👍🏼

PR #10449, opened 2023-02-12
Fixes #36083 - Set remote_execution_by_default to true

Ron-Lavi:

Changed the `remote_execution_by_default` default value to `true`

The Katello Agent/Goferd has been deprecated and has been replaced by remote execution. The "Satellite WebUI -> Administer -> Settings -> Content -> Use remote execution by default" is by default set to "No". Since Katello Agent/Goferd has been deprecated, this setting should either be set to 'Yes' by default or should be removed.

Verify that the default setting value is true.

#### Additional thoughts
- is the `katello-agent` deprecated-deprecated?
I wonder if we should really remove all `remote_execution_by_default` related code.

- in case we are keeping the related code,
should we create a migration to set this setting value to `true` in case it was explicitly set to `false`?

### DIFF
--- a/lib/katello/plugin.rb
+++ b/lib/katello/plugin.rb
@@ -566,7 +566,7 @@ Foreman::Plugin.register :katello do
 
       setting 'remote_execution_by_default',
         type: :boolean,
-        default: false,
+        default: true,
         full_name: N_('Use remote execution by default'),
         description: N_("If this is enabled, remote execution is used instead of katello-agent for remote actions")
 


jeremylenz:
katello-agent is deprecated, but will still be around for a few more releases (currently the plan is that 6.14 will be the last downstream release that has it.)

Fun fact: here's what Katello does to decide whether to use katello-agent or REX:

1. Look for the setting 'remote_execution_by_default'. If that's turned on, use REX. Otherwise..
2. On the host's content facet, look for the two properties 'katello_agent_installed' (which indicates if katello-agent is listed in the host's installed packages) and 'katello_agent_enabled' (which reflects if Foreman has katello-agent enabled via the installer - !!SETTINGS.dig(:katello, :agent, :enabled).)
3. If both of the above are true, use katello-agent.
4. Otherwise, use REX.

So if katello-agent isn't both enabled on the Foreman server and installed on the host, this setting has less importance.

However, if it _is_ enabled, this setting will change behavior for all those users who are still using Katello agent. So I'm on the fence about changing the default.  Was there a particular reason for raising this?

Ron-Lavi:
Thanks for the explanation @jeremylenz!
the request came from https://bugzilla.redhat.com/show_bug.cgi?id=2167493

would it be safe to say that for new instances we want the setting to be true by default,
and for older instances, we should keep the value that was there before?

jeremylenz:
> would it be safe to say that for new instances we want the setting to be true by default,
> and for older instances, we should keep the value that was there before?

For new installations, katello-agent is disabled by default anyway so this setting won't do anything.
For upgrades, I agree we should keep the value that was there before.

To resolve the BZ, I think the setting should be removed as part of the upcoming work removing katello-agent. So maybe align that Jira with SAT-1877 and close this for now.

Ron-Lavi:
ACK, thanks @jeremylenz !
closing it for now

jeremylenz:
See https://github.com/Katello/katello/pull/10679

PR #10454, opened 2023-02-16
Fixes #36051 - ForeignKeyViolation on ACS create when invalid --ssl-* argument is provided 

qcjames53:
Added logic to the ACS controller to improve the verbosity of errors returned when a user tries to create or update any ssl-* parameters on a simplified ACS.

This PR contains validation logic at the controller level (sort of). Reasons below.
##### Why I think this is ok:
- Active Record Validations will always evaluate a `belongs_to` relation before a `validates` condition. When supplying an invalid --ssl-* param, ARV would first check if the param matched an existing --ssl-* record and would return an error about the mismatched id before returning an error about how the id must be blank with simplified ACS. This essentially meant the Bugzilla ticket was asking for a custom error message to be displayed before the standard mismatched id message. The controller is the place to do something like that.
- There doesn't seem to be a good way to implement changing the order of ARV checks.
- Disabling the `belongs_to` check for simplified ACS's, while probably possible, doesn't seem like the best idea for a relation that needs to always hold.
- Even if the user bypasses the controller, the validation for --ssl-* params still exists for simplified ACS's. While they would  miss the custom error message, they still wouldn't be able to set invalid states. (This was fixed in validation in PR #10435)

##### Reasons this might be an issue:
- If the user bypasses the controller somehow, the custom error message will be missed and the mismatched ssl id message would be displayed instead.
- The logic between simplified and rhui/custom ACS's is diverging quite a bit. Should we consider splitting them into separate classes?

##### Other potentially messy bits:
- I based these changes on #10435 to avoid duplicating work with the --ssl-* param rejection and error handling in the create/update stages. --ssl-* was previously allowed to be passed as an update param with no ill effects. This may cause merge issues.
- `app/controllers/katello/api/v2/alternate_content_sources_controller.rb` contains two nearly duplicate functions: `check_params_for_invalid_create` and `check_params_for_invalid_update`. Due to the slight difference in how params are passed in for the two functions, I needed to separate the logic. If we decide to take a non-controller approach, this can all be deleted.

1. Have a Satellite with some Prods and Capsules you could refer to in 2.
2. Try to create or update an ACS:
- `hammer alternate-content-source create --alternate-content-source-type simplified --name "Test SACS" --ssl-ca-cert-id 2`
- `hammer alternate-content-source update --id 2 --ssl-ca-cert-id 2`
3. Replace [--ssl-ca-cert-id] with [--ssl-client-cert-id] and [--ssl-client-key-id], repeat above testing.

### DIFF
--- a/app/controllers/katello/api/v2/alternate_content_sources_controller.rb
+++ b/app/controllers/katello/api/v2/alternate_content_sources_controller.rb
@@ -71,14 +71,7 @@ module Katello
     def create
       @alternate_content_source = ::Katello::AlternateContentSource.new(acs_params.except(:smart_proxy_ids, :smart_proxy_names, :product_ids))
 
-      # Check that ssl-* params are not included if the ACS is of type 'simplified'
-      # This is handled in the controller to allow us to validate key absence before key value being valid.
-      # See https://bugzilla.redhat.com/show_bug.cgi?id=2159963 for more info
-      if @alternate_content_source.simplified?
-        (fail HttpErrors::UnprocessableEntity, "SSL ca cert must remain blank for ACS of type #{@alternate_content_source&.alternate_content_source_type}") unless @alternate_content_source.ssl_ca_cert_id.nil?
-        (fail HttpErrors::UnprocessableEntity, "SSL client cert must remain blank for ACS of type #{@alternate_content_source&.alternate_content_source_type}") unless @alternate_content_source.ssl_client_cert_id.nil?
-        (fail HttpErrors::UnprocessableEntity, "SSL ca cert must remain blank for ACS of type #{@alternate_content_source&.alternate_content_source_type}") unless @alternate_content_source.ssl_ca_cert_id.nil?
-      end
+      check_params_for_invalid_create
 
       sync_task(::Actions::Katello::AlternateContentSource::Create, @alternate_content_source, @smart_proxies, @products)
       @alternate_content_source.reload
--- a/app/controllers/katello/api/v2/alternate_content_sources_controller.rb
+++ b/app/controllers/katello/api/v2/alternate_content_sources_controller.rb
@@ -125,21 +125,6 @@ module Katello
       params.require(:alternate_content_source).permit(*keys).to_h.with_indifferent_access
     end
 
-    def check_params_for_invalid_updates
-      # Check parameters which cannot be validated at the model level, throwing
-      # errors where neccessary
-
-      # Check that the combination of params[:product_ids] and ACS type is allowed:
-      #            | simplified | custom  | rhui
-      # -----------+------------+---------+---------
-      # nil        | ok         | ok      | ok
-      # []         | ok         | invalid | invalid
-      # [foo, ...] | ok         | invalid | invalid
-      unless @alternate_content_source&.simplified? || params[:product_ids].nil?
-        (fail HttpErrors::UnprocessableEntity, "Products must remain blank for ACS of type #{@alternate_content_source&.alternate_content_source_type}")
-      end
-    end
-
     def find_smart_proxies
       if params[:smart_proxy_ids]
         @smart_proxies = ::SmartProxy.where(id: params[:smart_proxy_ids])
--- a/app/controllers/katello/api/v2/alternate_content_sources_controller.rb
+++ b/app/controllers/katello/api/v2/alternate_content_sources_controller.rb
@@ -70,9 +70,6 @@ module Katello
     param_group :acs
     def create
       @alternate_content_source = ::Katello::AlternateContentSource.new(acs_params.except(:smart_proxy_ids, :smart_proxy_names, :product_ids))
-
-      check_params_for_invalid_create
-
       sync_task(::Actions::Katello::AlternateContentSource::Create, @alternate_content_source, @smart_proxies, @products)
       @alternate_content_source.reload
       respond_for_create(resource: @alternate_content_source)
--- a/app/controllers/katello/api/v2/alternate_content_sources_controller.rb
+++ b/app/controllers/katello/api/v2/alternate_content_sources_controller.rb
@@ -70,6 +70,16 @@ module Katello
     param_group :acs
     def create
       @alternate_content_source = ::Katello::AlternateContentSource.new(acs_params.except(:smart_proxy_ids, :smart_proxy_names, :product_ids))
+
+      # Check that ssl-* params are not included if the ACS is of type 'simplified'
+      # This is handled in the controller to allow us to validate key absence before key value being valid.
+      # See https://bugzilla.redhat.com/show_bug.cgi?id=2159963 for more info
+      if @alternate_content_source.simplified?
+        (fail HttpErrors::UnprocessableEntity, "SSL ca cert must remain blank for ACS of type #{@alternate_content_source&.alternate_content_source_type}") unless @alternate_content_source.ssl_ca_cert_id.nil?
+        (fail HttpErrors::UnprocessableEntity, "SSL client cert must remain blank for ACS of type #{@alternate_content_source&.alternate_content_source_type}") unless @alternate_content_source.ssl_client_cert_id.nil?
+        (fail HttpErrors::UnprocessableEntity, "SSL ca cert must remain blank for ACS of type #{@alternate_content_source&.alternate_content_source_type}") unless @alternate_content_source.ssl_ca_cert_id.nil?
+      end
+
       sync_task(::Actions::Katello::AlternateContentSource::Create, @alternate_content_source, @smart_proxies, @products)
       @alternate_content_source.reload
       respond_for_create(resource: @alternate_content_source)
--- a/app/controllers/katello/api/v2/alternate_content_sources_controller.rb
+++ b/app/controllers/katello/api/v2/alternate_content_sources_controller.rb
@@ -125,6 +125,20 @@ module Katello
       params.require(:alternate_content_source).permit(*keys).to_h.with_indifferent_access
     end
 
+    def check_params_for_invalid_updates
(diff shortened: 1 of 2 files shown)


qcjames53:
@jeremylenz and @ianballou, please advise regarding the considerations section above. Thanks!

jeremylenz:
> * Active Record Validations will always evaluate a `belongs_to` relation before a `validates` condition.

Is this true? What if `validates` is called in the model before the `belongs_to`?

also, If you get a foreign key error, it's not because of when `belongs_to` was evaluated. (`belongs_to` only adds methods to the model; it doesn't do any DB queries.) It's because all the validations _passed_ and Rails attempted to save the record in the database.

> If the user bypasses the controller somehow, the custom error message will be missed and the mismatched ssl id message would be displayed instead

This is comparing a Rails validation message with a Postgres foreign key error, if I'm understanding correctly? I still don't think it's ideal to let it get all the way to Postgres if we can avoid it.

> The logic between simplified and rhui/custom ACS's is diverging quite a bit. Should we consider splitting them into separate classes?

Yes, I think this is worth a discussion. @ianballou maybe we can all meet about it

jeremylenz:
can't remember if I suggested this already.. could `validates_associated` help here? https://guides.rubyonrails.org/active_record_validations.html#validates-associated

qcjames53:
Hi Jeremy! Thanks for taking a look at this.

Regarding the `validates` vs `belongs_to` order of operations, I could be completely wrong. I was personally unable to find a way of getting field validations to occur before the db checked the relations for --ssl-* keys. Additionally, it seems that all validations are checked when the record is attempted to be saved; both the key relation check and the field must be blank check are triggered at this point.

What were you thinking would be a good candidate for `validates_associated`? The --ssl-* objects the keys match to already exist in the database and have been separately validated (provided the --ssl-* key is valid). If we created wrapper objects to validate the key numbers themselves, would that even differ from a standard `validates` call?

> This is comparing a Rails validation message with a Postgres foreign key error, if I'm understanding correctly? I still don't think it's ideal to let it get all the way to Postgres if we can avoid it.

Regarding this, all I was saying is the current validations handle all bad data (including throwing a Postgres error if the key is bad) regardless of whether the error message indicates a bad --ssl-* key or it indicates that the --ssl-* param must be blank. Are you saying you'd like to catch a bad foreign key error before it gets to Postgres for all ACS types?

By the way, thank you for the help! I can schedule a meeting to talk more about this in a little bit.

jeremylenz:
> Are you saying you'd like to catch a bad foreign key error before it gets to Postgres for all ACS types?

Yes. Foreign key errors ideally shouldn't ever be seen by users, but Rails validation errors are meant to be user-facing (at least in this project :)

qcjames53:
Validations have been moved from the controller to the model layer. Thank you for all the help!

app/models/katello/alternate_content_source.rb
```diff
@@ -110,8 +116,35 @@ def audit_updated_products(old_product_ids)
...
+        errors.add(:content_type, "cannot be modified once an ACS is created")
+      end
+      if changes.keys.include? "alternate_content_source_type"
+        errors.add(:alternate_content_source_type, "cannot be modified once an ACS is created")
+      end
+    end
+
+    # Disallow --ssl-* properties from being set for simplified ACS
+    # Must be done in method as workaround for validation printing the incorrect
+    # field string.
```
jeremylenz, line 135: I don't really consider this a workaround :smile:  It's just another option Rails offers for writing validations. I'd remove this comment

app/controllers/katello/api/v2/alternate_content_sources_controller.rb
```diff
@@ -117,12 +119,42 @@ def refresh
...
+        (fail HttpErrors::UnprocessableEntity, "Products must remain blank for ACS of type #{@alternate_content_source&.alternate_content_source_type}")
+      end
+
+      # Check that ssl-* params are not included if the ACS is of type 'simplified'
+      # This is handled in the controller to allow us to display a custom error  message, allowing for the presence of a key to be reported before checking whether the key matches an ssl entry.
+      # See https://bugzilla.redhat.com/show_bug.cgi?id=2159963 for more info
+      if @alternate_content_source.simplified?
+        (fail HttpErrors::UnprocessableEntity, "SSL ca cert must remain blank for ACS of type #{@alternate_content_source&.alternate_content_source_type}") unless params[:ssl_ca_cert_id].nil?
+        (fail HttpErrors::UnprocessableEntity, "SSL client cert must remain blank for ACS of type #{@alternate_content_source&.alternate_content_source_type}") unless params[:ssl_client_cert_id].nil?
+        (fail HttpErrors::UnprocessableEntity, "SSL client key must remain blank for ACS of type #{@alternate_content_source&.alternate_content_source_type}") unless params[:ssl_client_key_id].nil?
```
jeremylenz, line 154: These should still be removed from the controller.

jeremylenz:
@qcjames53 can you rebase this to remove the #36068 commits? :)

qcjames53:
> @qcjames53 can you rebase this to remove the #36068 commits? :)

Haha of course.

jeremylenz (changes requested):
The simplified case is working well, but I still got a ForeignKeyViolation with a custom ACS.

(in this example ACS with id 2 is simplified and 1 is custom)

```
[vagrant@centos8-katello-devel-stable hammer-cli-katello]$ hammer alternate-content-source update --id 2 --ssl-client-cert-id 1
Could not update the Alternate Content Source.:
  Validation failed: Ssl client cert must be blank
[vagrant@centos8-katello-devel-stable hammer-cli-katello]$ hammer alternate-content-source update --id 1 --ssl-client-cert-id 1
Could not update the Alternate Content Source.:
  PG::ForeignKeyViolation: ERROR:  insert or update on table "katello_alternate_content_sources" violates foreign key constraint "katello_alternate_content_sources_ssl_client_cert_id"
  DETAIL:  Key (ssl_client_cert_id)=(1) is not present in table "katello_content_credentials".
```

We need to add these validations to catch it before it hits the database:

```rb
validates :ssl_ca_cert, if: -> { custom? && verify_ssl || rhui? && verify_ssl }, presence: true
validates :ssl_client_cert, if: -> { custom? && verify_ssl || rhui? && verify_ssl }, presence: true
validates :ssl_client_key, if: -> { custom? && verify_ssl || rhui? && verify_ssl }, presence: true

validates :ssl_ca_cert_id, if: -> { custom? && !verify_ssl || rhui? && !verify_ssl }, absence: { message: "ssl_ca_cert_id must be blank when verify_ssl is false" }
validates :ssl_client_cert_id, if: -> { custom? && !verify_ssl || rhui? && !verify_ssl }, absence: { message: "ssl_client_cert_id must be blank when verify_ssl is false" }
validates :ssl_client_key_id, if: -> { custom? && !verify_ssl || rhui? && !verify_ssl }, absence: { message: "ssl_client_key_id must be blank when verify_ssl is false" }

```

After adding these, I got this (which is what we want):

```
[vagrant@centos8-katello-devel-stable hammer-cli-katello]$ hammer alternate-content-source update --id 1 --verify-ssl false --ssl-ca-cert-id 1
Could not update the Alternate Content Source.:
  Validation failed: Alternate Content Sources ssl_ca_cert_id must be blank when verify_ssl is false
[vagrant@centos8-katello-devel-stable hammer-cli-katello]$ hammer alternate-content-source update --id 1 --verify-ssl true --ssl-ca-cert-id 1
Could not update the Alternate Content Source.:
  Validation failed: Ssl ca cert can't be blank, Ssl client cert can't be blank, Ssl client key can't be blank
[vagrant@centos8-katello-devel-stable hammer-cli-katello]$ hammer alternate-content-source update --id 2 --verify-ssl true --ssl-ca-cert-id 1
Could not update the Alternate Content Source.:
  Validation failed: Verify ssl must be blank, Ssl ca cert must be blank

```

app/models/katello/alternate_content_source.rb
```diff
@@ -16,9 +16,22 @@ class AlternateContentSource < Katello::Model
 
     encrypts :upstream_password
 
+    # Validate the ssl_id params.
```
jeremylenz, line 19: Relationships aren't technically validations; I'd remove this comment

app/models/katello/alternate_content_source.rb
```diff
@@ -16,9 +16,22 @@ class AlternateContentSource < Katello::Model
...
     belongs_to :ssl_ca_cert, inverse_of: :ssl_ca_alternate_content_sources, class_name: "Katello::ContentCredential"
     belongs_to :ssl_client_cert, inverse_of: :ssl_client_alternate_content_sources, class_name: "Katello::ContentCredential"
     belongs_to :ssl_client_key, inverse_of: :ssl_key_alternate_content_sources, class_name: "Katello::ContentCredential"
+    # We breakout some validation into a function to allow for us to set
+    # the object name correctly in error messages
+    validate :validate_ssl_ids
+    validates :ssl_ca_cert, if: -> { ssl_ca_cert_id.present? }, presence: {
+      message: "id is invalid"
```
jeremylenz, line 27:
```suggestion
      message: "Couldn't find ssl_ca_cert with ID '#{ssl_ca_cert_id}'"
```

jeremylenz (approved):
I'm happy with this! Tested and seems to work as it should.

just left a couple more comments for extra credit

app/models/katello/alternate_content_source.rb
qcjames53: Done!

app/models/katello/alternate_content_source.rb
qcjames53: Added this functionality, thanks!

PR #10456, opened 2023-02-17
Fixes #36122 - Hide unsyncable content type warning

adamruzicka:
until the information is actually loaded

The warning is supposed to be based on some data, but the condition was failing if the data was not loaded yet. These changes take that option into consideration.

Carrying some default state in the angular controller would probably be cleaner, but this seems to work

1) Have an external content proxy
2) Visit its detail page


ianballou (approved):
The smart proxy details page is looking fine to me! Tested a sync just to be safe.

PR #10457, opened 2023-02-17
Fixes #35559 - Change Content Source LCE dropdown should not show multiple Library entries

lfu:
There should be only one 'Library' selection in the dropdown. Which Library it is depends on the organization selector.

1. Create at least one extra org
2. Create at least one content view in Default Organization (this will make the problem easier to see)
3. Register a host to Default Organization
4. Go to change that host's content source (host details > kebab > Change Content Source)
5. Select the Satellite as the content source
6. Observe the selections for Lifecycle Environment

### DIFF
--- a/webpack/components/extensions/HostDetails/Cards/ContentViewDetailsCard/ChangeHostCVModal.js
+++ b/webpack/components/extensions/HostDetails/Cards/ContentViewDetailsCard/ChangeHostCVModal.js
@@ -1,11 +1,7 @@
 import React, { useState, useCallback } from 'react';
 import PropTypes from 'prop-types';
-import { FormattedMessage } from 'react-intl';
 import { useDispatch, useSelector } from 'react-redux';
-import { Modal, Button, SelectOption, Alert, Flex } from '@patternfly/react-core';
-import {
-  global_palette_black_600 as pfDescriptionColor,
-} from '@patternfly/react-tokens';
+import { Modal, Button, Alert } from '@patternfly/react-core';
 import { translate as __ } from 'foremanReact/common/I18n';
 import { STATUS } from 'foremanReact/constants';
 import { useAPI } from 'foremanReact/common/hooks/API/APIHooks';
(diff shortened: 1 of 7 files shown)


jeremylenz:
@lfu I thought we were going to move to the new CV/LCE selector?

chris1984:
@jeremylenz do you want me to review this so you can focus on other things?

jeremylenz:
@chris1984 feel free to review as you like! But I definitely want to see the results of this, since it will become even more important once the multi-CV work starts up again :)

lfu:
Updated.

jeremylenz:
Change-host-content-source

Can the environment paths be left-aligned?

webpack/scenes/Hosts/ChangeContentSource/index.js
```diff
@@ -45,18 +44,18 @@ const ChangeContentSourcePage = () => {
...
   const contentHosts = useSelector(selectContentHosts);
   const hostsWithoutContent = useSelector(selectContentHostsWithoutContent);
-  const environments = useSelector(selectEnvironments);
   const contentSources = useSelector(selectContentSources);
   const jobInvocationPath = useSelector(selectJobInvocationPath);
 
   const template = useSelector(selectTemplate);
   const contentViews = useSelector(selectContentViews);
 
   const [contentSourceId, setCapsuleId] = useState('');
```
jeremylenz, line 53: When I only have a single content source, it's confusing that I have to click it before the environments component becomes enabled. Can this default to the host's current content source?

webpack/scenes/ContentViews/components/ContentViewSelect/ContentViewSelectOption.js
```diff
@@ -0,0 +1,76 @@
...
+import { translate as __ } from 'foremanReact/common/I18n';
+import PropTypes from 'prop-types';
+import { FormattedMessage } from 'react-intl';
+import {
+  global_palette_black_600 as pfDescriptionColor,
+} from '@patternfly/react-tokens';
+import { uniq } from '../../../../utils/helpers';
+import ContentViewIcon from '../../../../scenes/ContentViews/components/ContentViewIcon';
+
+const relevantVersionObjFromCv = (cv, env) => { // returns the entire version object
```
jeremylenz, line 12: Pretty much this entire file is reused from `ChangeHostCVModal.js`. Instead of duplicating the code, I'd like to see the components and helper methods extracted and used in both places. There are a few ways to do this - you can keep everything in the ChangeHostCVModal area and then import them here; you can put all the common stuff in a new area and import them both places; or you can move everything over here and import stuff in ChangeHostCVModal. Or you could do a combination of these - whatever works best. The important thing is that we don't have two copies of the same code in our code base, so this way if one gets updated we don't have to remember to update the other.

jeremylenz (changes requested):
We've already fixed the bug - confirmed that the environment paths (including Library) are only shown for the current organization context.

This is looking fine once the code is cleaned up; see below :)

webpack/components/extensions/HostDetails/Cards/ContentViewDetailsCard/ChangeHostCVModal.js
```diff
@@ -193,35 +155,7 @@ const ChangeHostCVModal = ({
...
-                  <ContentViewDescription
-                    cv={cv}
-                    versionNumber={relevantVersionFromCv(cv, selectedEnv)}
-                  />
-                </Flex>
-              </Flex>
-            </SelectOption>
-          ))
-          }
+          {contentViewsInEnv?.map(cv => ContentViewSelectOption(cv, selectedEnvForHost[0]))}
```
jeremylenz, line 158:
```suggestion
          {contentViewsInEnv?.map(cv => <ContentViewSelectOption cv={cv} selectedEnvForHost={selectedEnvForHost[0]} />)}
```

jeremylenz (commented):
Looking much better!!

webpack/components/extensions/HostDetails/Cards/ContentViewDetailsCard/ChangeHostCVModal.js
sjha4: Updating it this way makes the selected option show up as Id instead of name for some reason. Pushed a few updates other than this change.

jeremylenz: you'll also need to make the function take an object ;)

jeremylenz:
```js
<ContentViewSelectOption cv={cv} />
```

is the same as
```js
ContentViewSelectOption({ cv })
```

or 
```js
ContentViewSelectOption({ cv: cv })
```

webpack/scenes/Hosts/ChangeContentSource/index.js
sjha4: Updated to use a placeholder text for both bulk and single change forms.

jeremylenz (commented):
Code looks better now, thanks :)

Couple more things:

1. After the update process is complete, I am able to re-select a different content source and env/CV, even though the Update button remains disabled. Can we just disable everything after the update is complete?
2. If no content views are available for the selected environment, can we add the banner like we have in the ChangeHostCVModal?

jeremylenz (approved):
thanks @sjha4 @lfu!  

LGTM :+1:

PR #10458, opened 2023-02-17
Fixes #36124 - fix wording of errata_status_installable setting

jeremylenz:

The setting `errata_status_installable` had a bad description. I had to look at the code to figure out what the setting does. Turns out, it's about whether to base your host's errata status on installable or applicable errata. I'm updating the setting's full name and description so that others don't have to wonder.

The short name `errata_status_installable` isn't changing, so we don't have to worry about a migration or anything.

view Administer > Settings > Content > you should now see the new name "Generate errata status from directly-installable content" and the new description.

### DIFF
--- a/lib/katello/plugin.rb
+++ b/lib/katello/plugin.rb
@@ -475,8 +475,8 @@ Foreman::Plugin.register :katello do
       setting 'errata_status_installable',
         type: :boolean,
         default: false,
-        full_name: N_('Installable errata from content view'),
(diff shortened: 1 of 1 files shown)


jeremylenz:
[test katello]

jeremylenz:
[test katello]

chris1984 (approved):
ACK, looks a lot better than what we had before.

PR #10460, opened 2023-02-21
Fixes #36132 - Add export definitions for columns

adamruzicka:
Requires:
- [x] https://github.com/theforeman/foreman/pull/9618

### DIFF
--- a/lib/katello/plugin.rb
+++ b/lib/katello/plugin.rb
@@ -271,19 +271,22 @@ Foreman::Plugin.register :katello do
       common_td_class = "#{common_th_class} ellipsis"
       use_pagelet :hosts_table_column_header, :name
       use_pagelet :hosts_table_column_content, :name
-      add_pagelet :hosts_table_column_header, key: :subscription_status, label: _('Subscription status'), sortable: true, class: common_th_class, width: '10%'
+      add_pagelet :hosts_table_column_header, key: :subscription_status, label: _('Subscription status'), sortable: true, class: common_th_class, width: '10%', export_key: 'subscription_global_status'
       add_pagelet :hosts_table_column_content, key: :subscription_status, class: common_td_class, callback: ->(host) { host_status_icon(host.subscription_global_status) }
-      add_pagelet :hosts_table_column_header, key: :installable_updates, label: _('Installable updates'), class: common_th_class, width: '15%'
+      add_pagelet :hosts_table_column_header, key: :installable_updates, label: _('Installable updates'), class: common_th_class, width: '15%',
+                  export_data: [:security, :bugfix, :enhancement].map { |kind| CsvExporter::ExportDefinition.new("installable_updates.#{kind}", callback: ->(host) { (host.content_facet_attributes&.errata_counts || {})[kind] }) } +
+                               [:rpm, :deb].map { |kind| CsvExporter::ExportDefinition.new("installable_packages.#{kind}", callback: ->(host) { host&.content_facet_attributes&.public_send("upgradable_#{kind}_count".to_sym) || 0 }) }
       add_pagelet :hosts_table_column_content, key: :installable_updates, class: common_td_class, callback: ->(host) { errata_counts(host) }
       use_pagelet :hosts_table_column_header, :os_title
       use_pagelet :hosts_table_column_content, :os_title
-      add_pagelet :hosts_table_column_header, key: :lifecycle_environment, label: _('Lifecycle environment'), sortable: true, class: common_th_class, width: '10%'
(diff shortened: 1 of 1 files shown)


lib/katello/plugin.rb
```diff
@@ -271,19 +271,21 @@
...
       use_pagelet :hosts_table_column_header, :name
       use_pagelet :hosts_table_column_content, :name
-      add_pagelet :hosts_table_column_header, key: :subscription_status, label: _('Subscription status'), sortable: true, class: common_th_class, width: '10%'
+      add_pagelet :hosts_table_column_header, key: :subscription_status, label: _('Subscription status'), sortable: true, class: common_th_class, width: '10%', export_key: 'subscription_global_status'
       add_pagelet :hosts_table_column_content, key: :subscription_status, class: common_td_class, callback: ->(host) { host_status_icon(host.subscription_global_status) }
-      add_pagelet :hosts_table_column_header, key: :installable_updates, label: _('Installable updates'), class: common_th_class, width: '15%'
+      add_pagelet :hosts_table_column_header, key: :installable_updates, label: _('Installable updates'), class: common_th_class, width: '15%',
+                  export_data: [:security, :bugfix, :enhancement].map { |kind| CsvExporter::ExportDefinition.new("installable_updates.#{kind}", callback: ->(host) { (host.content_facet_attributes&.errata_counts || {})[kind] }) }
```
ofedoren:
```suggestion
                  export_data: [:security, :bugfix, :enhancement].map { |kind| CsvExporter::ExportDefinition.new("installable_updates.#{kind}", callback: ->(host) { (host.content_facet_attributes&.errata_counts || {})[kind] }) } +
```

Otherwise raises the error: undefined `+@ for Array`

lib/katello/plugin.rb
```diff
@@ -271,19 +271,21 @@
...
+                               + [:rpm, :deb].map { |kind| CsvExporter::ExportDefinition.new("installable_packages.#{kind}", callback: ->(host) { host&.content_facet_attributes&.public_send("upgradable_#{kind}_count".to_sym) || 0 }) }
```
ofedoren:
```suggestion
                               [:rpm, :deb].map { |kind| CsvExporter::ExportDefinition.new("installable_packages.#{kind}", callback: ->(host) { host&.content_facet_attributes&.public_send("upgradable_#{kind}_count".to_sym) || 0 }) }
```

ofedoren (commented):
Few issues, but otherwise LGTM. I'd ask Katello guys to take a better look of course :)

ofedoren:
Can someone take a look at this? Core PR was merged, we'd like to have these changes in sync, since it affects CSV export.

sjha4:
[test katello]

sjha4:
@adamruzicka : I don't have enough context on this. Could you provide some steps to verify functionality etc..

adamruzicka:
Go to host index, add content related columns, hit export. Without this change the export would (possibly) just crash. With this change, the export will pass and will contain content related fields

lib/katello/plugin.rb
```diff
@@ -271,19 +271,21 @@
...
       add_pagelet :hosts_table_column_content, key: :installable_updates, class: common_td_class, callback: ->(host) { errata_counts(host) }
       use_pagelet :hosts_table_column_header, :os_title
       use_pagelet :hosts_table_column_content, :os_title
-      add_pagelet :hosts_table_column_header, key: :lifecycle_environment, label: _('Lifecycle environment'), sortable: true, class: common_th_class, width: '10%'
+      add_pagelet :hosts_table_column_header, key: :lifecycle_environment, label: _('Lifecycle environment'), sortable: true, class: common_th_class, width: '10%', export_key: 'single_lifecycle_environment'
       add_pagelet :hosts_table_column_content, key: :lifecycle_environment, class: common_td_class, callback: ->(host) { host.content_facet&.single_lifecycle_environment&.name }
-      add_pagelet :hosts_table_column_header, key: :content_view, label: _('Content view'), sortable: true, class: common_th_class, width: '10%'
+      add_pagelet :hosts_table_column_header, key: :content_view, label: _('Content view'), sortable: true, class: common_th_class, width: '10%', export_key: 'single_content_view'
```
sjha4, line 284: These headers seem to be picking up label based on key instead of the actual label param.
Screenshot from 2023-03-01 16-21-47

adamruzicka: Yes, that's the way it works. The label itself is translated which is not really desirable in csv exports. I mean, there are ways to force it to be named differently if that's what you want

sjha4: That would be ideal. 👍🏼

sjha4 (approved):
Looks good now. Thanks @adamruzicka !

PR #10466, opened 2023-02-23
Fixes #36142 - Allow reclaim space without explicitly setting Organization

wbclark:

Allows running reclaim space without relying on the Organization being set. Previously this used `Organization.current` and had `repositories.first.organization` as a fallback, however the fallback was broken. Since https://github.com/Katello/katello/pull/9952 we pass the `repo` to the task directly, so now we can simply use `repo.organization` as the definitive approach to determine the organization to use for the task.

To test this, I removed defining the organization in the test case setup. The repo should get created now with any current organization, and the task should properly determine the organization from the repo object.

I also removed other setup steps that are no longer used.

Create a repository. Install hammer. Test with `hammer repository reclaim-space --id 1 --organization-id 1`

### DIFF
--- a/app/lib/actions/pulp3/repository/reclaim_space.rb
+++ b/app/lib/actions/pulp3/repository/reclaim_space.rb
@@ -4,7 +4,7 @@ module Actions
       class ReclaimSpace < Pulp3::AbstractAsyncTask
         def plan(repo, smart_proxy = SmartProxy.pulp_primary)
           action_subject(repo)
-          repository_hrefs = ::Katello::Pulp3::RepositoryReference.default_cv_repository_hrefs([repo], Organization.current || repositories.first.organization)
+          repository_hrefs = ::Katello::Pulp3::RepositoryReference.default_cv_repository_hrefs([repo], repo.organization)
           plan_self(repository_hrefs: repository_hrefs, smart_proxy_id: smart_proxy.id)
         end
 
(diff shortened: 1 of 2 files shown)


sjha4:
Works as advertised. 
**Before:**
Before_PR

**After:**
After_PR

One small note..
The `organization-id` param is all but ignored when passed via hammer.
Ex: `hammer repository reclaim-space --id 1 --organization-id 1` and `hammer repository reclaim-space --id 1 --organization-id 2` work the same based on repo-id.

sjha4 (approved):
Ack..Feel free to open a hammer issue to reject org-id or respect that param if passed. Not sure what the correct approach would be on that..

wbclark:
Pushed an update to the commit message only, adding @pmoravec as a co-author as this was originally his suggestion.

I'm leaning towards removing the ability to pass an organization to `hammer repository reclaim-space` since it's better not to mislead the user by silently ignoring it. I will create the card for that follow-up after thinking on it a bit more.

Thanks @sjha4 for the helpful and timely review.

wbclark:
In hammer we could handle this by first printing a warning when the user tries to pass an organization. this approach informs users that the organization won't be used, without immediately breaking any existing scripts. that can be done at a later stage when users have had sufficient time to update scripts, or at least not be surprised by the change.

I created https://projects.theforeman.org/issues/36148 to track this

PR #10471, opened 2023-03-02
Fixes #36137 - Catch content_view_id and LCE id before Rails update

jeremylenz:

The hosts controller in Foreman passes through all attributes with `@host.update(attributes)`. This causes problems with attributes that no longer exist on content hosts, such as `content_view_id` and `lifecycle_environment_id`.

This adds a method that overrides `Host::Managed#update` to account for that.

Theoretically this fix is only temporary, until we finish building out the multi CV feature and stop sending these params to the backend.

See steps in the Redmine issue.

### DIFF
--- a/app/models/katello/concerns/host_managed_extensions.rb
+++ b/app/models/katello/concerns/host_managed_extensions.rb
@@ -6,6 +6,20 @@ module Katello
       include ForemanTasks::Concerns::ActionSubject
 
       module Overrides
+        def update(attrs)
+          if attrs[:content_facet_attributes]
+            cv_id = attrs[:content_facet_attributes].delete(:content_view_id)
+            lce_id = attrs[:content_facet_attributes].delete(:lifecycle_environment_id)
+            if cv_id && lce_id
+              content_facet.assign_single_environment(content_view_id: cv_id, lifecycle_environment_id: lce_id)
+            end
+            if (cv_id.present? && lce_id.blank?) || (cv_id.blank? && lce_id.present?)
+              fail "content_view_id and lifecycle_environment_id must be provided together"
+            end
+          end
+          super
+        end
+
         def validate_media?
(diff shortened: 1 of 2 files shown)


app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -6,6 +6,20 @@ module HostManagedExtensions
...
       module Overrides
+        def update(attrs)
+          if attrs[:content_facet_attributes]
+            cv_id = attrs[:content_facet_attributes].delete(:content_view_id)
+            lce_id = attrs[:content_facet_attributes].delete(:lifecycle_environment_id)
+            if cv_id && lce_id
+              content_facet.assign_single_environment(content_view_id: cv_id, lifecycle_environment_id: lce_id)
+            end
+            if (cv_id.present? && lce_id.blank?) || (cv_id.blank? && lce_id.present?)
+              fail "content_view_id and lifecycle_environment_id must be provided together"
```
ofedoren, line 17: Although the patch seems to fix the issue, I've got a semi-related question:

If I've created a host within a hostgroup which has lifecycle env and content view attached, can I remove prefiled and saved these inherited environment and content view from the host?

jeremylenz: Theoretically you should be able to click the X and remove the value. That whole area is buggy right now though; also @lfu is currently working on https://projects.theforeman.org/issues/35834 which will refactor that whole area of the form.

chris1984:
Code wise it makes sense, will give it a quick run through and ack based on results. 

Do you think we should add a new test for the method? I am not set on it because as you stated it's a temporary fix, just figured I would ask?

jeremylenz:
@chris1984: Updated :+1:

chris1984 (approved):
ACK, works

video of me in the host edit screen being able to save.

https://nimb.ws/uPfAz5

Thanks for adding the test :smile:

PR #10472, opened 2023-03-02
Fixes #36157 - reuse pulpcore registry manifest headers

ianballou:
Stops Katello from calculating container manifest pull headers. Trust the headers coming from Pulp during manifest pull.

The old code is very likely due to Pulp 2 times. Ina from Pulp mentioned we should be able to trust their headers.

1) Try to sync and pull quay.io/riadh_hamdi/ceph-zabbix:latest
2) Try other container images and ensure you can pull them.


ianballou:
Looks like I'll have to adjust some tests

sjha4:
The fix works in my testing. 
Before PR:
Before_PR

After PR: 
After_PR

Change makes sense. Will ack after test fixes.

sjha4 (approved):
APJ 👍🏼

PR #10477, opened 2023-03-03
Fixes #36164 - Incremental imports create proper publication

parthaa:
Fixes an issue with incremental imports, which correctly updates the database and ui counts but does not create a new publication when that data is copied to the library

The `force` variable in MetadataGenerate got renamed to `force_publication` at some point. This updates the import action to now use that,

On the export server

- Enable a RH Repo (recommend `Red Hat Ansible Engine 2 for RHEL 8 x86_64 RPMs`)
- Update its download policy to immediate
- sync
- Create a content view version
- Add this repository
- Add rpm excludes filter to filter out `sshpass`
- Publish
- Update the rpm excludes filter now to filter out `sshpassnothere` instead of `sshpass`
- Publish again
- Export version one with `hammer content-export complete version --id=<...>`
- Export version two as incremental  with `hammer content-export incremental version --id=<...>`

Import Side
-----------------
- Create  a new org
- import manifest
- Import version 1 of the exported content - `hammer content-import version --id=<> --path=<>`
- create an activation key to use Library
- Go to Repository sets for the AK and make sure the Ansible repo is enabled 
- Register a rhel8 host to that key
- On the host run `dnf clean all && dnf repolist -v` 
- You should see the number of packages for ansible repo listed as 57
- Import version 2 of the exported content - `hammer content-import version --id=<> --path=<>`
- On the host run `dnf clean all && dnf repolist -v` 

Before PR
- You should see the number of packages for ansible repo still listed as 57

After PR
- Delete the  import version2 in the cv and import version 2 again
- On the host run `dnf clean all && dnf repolist -v` 
- You should see the number of packages for ansible repo still listed as 58

### DIFF
--- a/lib/katello/tasks/upgrades/4.8/regenerate_imported_repository_metadata.rake
+++ b/lib/katello/tasks/upgrades/4.8/regenerate_imported_repository_metadata.rake
@@ -5,7 +5,7 @@ namespace :katello do
       task :regenerate_imported_repository_metadata => ["dynflow:client", "environment"] do
         User.current = User.anonymous_admin #set a user for orchestration
         versions = Katello::ContentViewVersionImportHistory.all.map(&:content_view_version)
-        repo_ids = versions.map do |ver|
+        repos = versions.map do |ver|
            rps = if ver.default?
                    ver.repositories
                   else
@@ -14,12 +14,11 @@ namespace :katello do
            rps.exportable
               .map(&:library_instance_or_self)
               .select(&:using_mirrored_metadata?)
-              .map(&:id)
         end.flatten.uniq
 
-        if repo_ids.any?
+        if repos.any?
           task = ForemanTasks.async_task(::Actions::Katello::Repository::BulkMetadataGenerate,
-                                  ::Katello::Repository.where(id: repo_ids),
+                                  repos,
                                   force_publication: true)
           puts "Refreshing #{repo_ids.count} repositories.  You can monitor these on task id #{task.id}\n"
         else
--- a/lib/katello/tasks/upgrades/4.8/regenerate_imported_repository_metadata.rake
+++ b/lib/katello/tasks/upgrades/4.8/regenerate_imported_repository_metadata.rake
@@ -23,7 +23,7 @@ namespace :katello do
                                          ::Actions::Katello::Repository::MetadataGenerate,
                                          repos,
                                          force_publication: true)
-          puts "Refreshing #{repo_ids.count} repositories.  You can monitor these on task id #{task.id}\n"
+          puts "Refreshing #{repos.count} repositories. You can monitor these on task id #{task.id}\n"
         else
           puts "No repositories found for regeneration."
         end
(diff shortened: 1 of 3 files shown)


ianballou:
As we discussed in a meet, there should also be an upgrade rake task to run metadata generation on incremental imports for repositories with "complete mirroring". The bug is due to the check we have that stops publication regeneration for complete mirroring repositories. For a normal repository use case, Pulp creates these.

sjha4:
Before change: The import of incremental CV doesn't refresh distribution with updated content count.
Before_PR

After change: The new publication and refreshed distribution updates content count as expected.
After_PR

Change works as expected. I am not sure about an additional rake task cause the workaround of regenerating metadata is easy enough.

lib/katello/tasks/upgrades/4.8/regenerate_imported_repository_metadata.rake
```diff
@@ -0,0 +1,31 @@
...
+        repo_ids = versions.map do |ver|
+           rps = if ver.default?
+                   ver.repositories
+                  else
+                    ver.archived_repos
+                  end
+           rps.exportable
+              .map(&:library_instance_or_self)
+              .select(&:using_mirrored_metadata?)
+              .map(&:id)
```
ianballou, line 17: Do you need to pluck out the IDs since you'll already have all the repository objects anyway from the select? Then you could just pass the repos into the foreman task bulk metadata generate.  That would avoid the `where` statement.

sjha4:
> @sjha4 @ianballou added the rake task

The task selected the one repository I was using for testing correctly based on import history. It's idempotent although it will create new publications everytime it runs for those repositories even after they have been created which shouldn't be a problem with an upgrade task.

I think calling this directly instead of going through the bulk_metadata_generate and finding nothing for repos.archived and repos.in_published_environments makes more sense.
`plan_action(::Actions::BulkAction, ::Actions::Katello::Repository::MetadataGenerate, repos, force_publication: true)`

parthaa:
> The task selected the one repository I was using for testing correctly based on import history. It's idempotent although it will create new publications everytime it runs for those repositories even after they have been created which shouldn't be a problem with an upgrade task.

@sjha4  So can you think of a way for me to detect if a new publication is needed.

parthaa:
@sjha4 updated

lib/katello/tasks/upgrades/4.8/regenerate_imported_repository_metadata.rake
```diff
@@ -0,0 +1,33 @@
...
+        end
+        repos.flatten!
+        repos.uniq!
+
+        if repos.any?
+          task = ForemanTasks.async_task(::Actions::BulkAction,
+                                         ::Actions::Katello::Repository::MetadataGenerate,
+                                         repos,
+                                         force_publication: true)
+          puts "Refreshing #{repo_ids.count} repositories.  You can monitor these on task id #{task.id}\n"
```
sjha4, line 26: `repo_ids` doesn't exist here anymore..

sjha4 (approved):
The task logic works to select imported library repos and trigger the metadata generate task for those repos. The fix works to force publication of the library repo for future imports. ACK 👍🏼

PR #10479, opened 2023-03-06
Fixes #36035 - Add empty content states for CV repositories

jeremylenz:

This PR improves empty states in the new CV UI for ContentViewRepositories, in the following two situations:

1. When your content view doesn't have any repositories added;
2. When there aren't any repositories in your organization.

It also changes the default view from "All" repositories to "Added," and adds call-to-action buttons & links:

1. Show repositories - switches the view from Added to All
2. Enable Red Hat repositories - link
3. Create a custom product - link

The "Show repositories" button was originally "Add repositories," but I thought it would be confusing to have two different "Add Repositories" buttons on the screen.

Test the following scenarios:

1. Empty content view with no repositories added
2. Empty content view in an empty organization with no products or Red Hat repos enabled
3. also make sure I didn't break empty states anywhere else in CV UI or the new host details page

### DIFF
--- a/webpack/components/Table/EmptyStateMessage.js
+++ b/webpack/components/Table/EmptyStateMessage.js
@@ -92,12 +92,17 @@ const EmptyStateMessage = ({
         {showSecondaryActionAnchor &&
           <EmptyStateSecondaryActions>
             <Button variant="link" ouiaId="empty-state-secondary-action-link">
-              <a href={secondaryActionLink} style={{ textDecoration: 'none' }}>{secondaryActionTitle}</a>
+              {extraTableProps.secondaryActionTargetBlank ? (
+                <a href={secondaryActionLink} target="_blank" rel="noreferrer" style={{ textDecoration: 'none' }} >{secondaryActionTitle}</a>
+              ) : (
+                <a href={secondaryActionLink} style={{ textDecoration: 'none' }} >{secondaryActionTitle}</a>
+              )}
             </Button>
           </EmptyStateSecondaryActions>
         }
 
-        {(showSecondaryActionButton || searchIsActive || !!filtersAreActive) &&
+        {(!showSecondaryActionAnchor &&
+          (showSecondaryActionButton || searchIsActive || !!filtersAreActive)) &&
           <EmptyStateSecondaryActions>
             <Button variant="link" onClick={handleClick} ouiaId="empty-state-secondary-action-router-link">
               {secondaryActionText}
--- a/webpack/components/Table/MainTable.js
+++ b/webpack/components/Table/MainTable.js
@@ -21,7 +21,8 @@ const MainTable = ({
   activeToggleState, unfilteredToggleState, actionButtons, rowsCount,
   children, showPrimaryAction, showSecondaryAction, showSecondaryActionButton, primaryActionLink,
   secondaryActionLink, primaryActionTitle, secondaryActionTitle, secondaryActionTextOverride,
-  resetFilters, updateSearchQuery, requestKey, primaryActionButton, ...extraTableProps
+  resetFilters, updateSearchQuery, requestKey, primaryActionButton, secondaryActionTargetBlank,
+  ...extraTableProps
 }) => {
   const tableHasNoRows = () => {
     if (children) return rowsCount === 0;
@@ -76,7 +77,8 @@ const MainTable = ({
         title={emptyContentTitle}
         body={emptyContentBody}
         happy={happyEmptyContent}
-        search={!happyEmptyContent}
+        search={!happyEmptyContent && !emptyContentOverride}
+        secondaryActionTargetBlank={secondaryActionTargetBlank}
         {...clearSearchProps}
         {...callToActionProps}
       />
@@ -162,6 +164,7 @@ MainTable.propTypes = {
   updateSearchQuery: PropTypes.func,
   requestKey: PropTypes.string,
   primaryActionButton: PropTypes.element,
+  secondaryActionTargetBlank: PropTypes.bool,
 };
 
 MainTable.defaultProps = {
@@ -192,6 +195,7 @@ MainTable.defaultProps = {
   updateSearchQuery: undefined,
   requestKey: '',
   primaryActionButton: undefined,
+  secondaryActionTargetBlank: false,
 };
 
 export default MainTable;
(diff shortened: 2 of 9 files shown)


jeremylenz:
image

webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js
```diff
@@ -217,8 +231,37 @@ const ContentViewRepositories = ({ cvId, details }) => {
...
+    <Button ouiaId="empty-state-primary-action-button" onClick={handleShow}>
+      {__('Show repositories')}
+    </Button>
+  ) : (
+    <Button ouiaId="empty-state-primary-action-button">
+      <a href="/redhat_repositories" style={{ color: 'white', textDecoration: 'none' }}>{__('Enable Red Hat repositories')}</a>
+    </Button>
+  );
+
+  const secondaryActionLink = trulyEmpty ? ('/products/') : undefined;
```
jeremylenz, line 259: Should this go to `/products/` or `/products/new`? I changed it to products because what if you've created a product already but haven't added repos to it. But I'm on the fence.

webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js
```diff
@@ -230,6 +273,8 @@ const ContentViewRepositories = ({ cvId, details }) => {
     </DropdownItem>,
   ];
 
+  const loadingStatus = (isLoading || status === STATUS.PENDING) ? STATUS.PENDING : status;
```
jeremylenz, line 276: When you clicked the Show Repositories button, it would flash momentarily with the trulyEmpty state before you saw the loading spinner. This prevents that.

webpack/components/Table/EmptyStateMessage.js
```diff
@@ -97,7 +97,8 @@ const EmptyStateMessage = ({
           </EmptyStateSecondaryActions>
         }
 
-        {(showSecondaryActionButton || searchIsActive || !!filtersAreActive) &&
+        {(!showSecondaryActionAnchor &&
```
jeremylenz, line 100: Added this so that it only shows one secondary action - either the <a> link style or the button with onClick - but not both.

webpack/components/Table/MainTable.js
```diff
@@ -76,7 +76,7 @@ const MainTable = ({
         title={emptyContentTitle}
         body={emptyContentBody}
         happy={happyEmptyContent}
-        search={!happyEmptyContent}
+        search={!happyEmptyContent && !emptyContentOverride}
```
jeremylenz, line 79: had to add this so that it would show the plus icon and not the search icon

webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js
sjha4: `/products/` is good. It has it's own empty state in bastion.. 😝

jeremylenz:
@sjha4 Updated:

1. Shortened wording
2. Updated empty state as per discussion with Maria
3. Updated API endpoints so that we can show the "trulyEmpty" state immediately.

image

webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js
```diff
@@ -217,8 +247,35 @@ const ContentViewRepositories = ({ cvId, details }) => {
...
+  const showPrimaryAction = emptyContentOverride || trulyEmpty;
+  const primaryActionButton = trulyEmpty ? (
+    <span style={{ fontSize: 'larger' }}><TrulyEmptyCallsToAction /></span>
+  ) : (
+    <Button ouiaId="empty-state-primary-action-button" onClick={handleShow}>
+      {__('Show repositories')}
+    </Button>
+  );
+
+  const secondaryActionLink = trulyEmpty ? ('https://docs.theforeman.org/nightly/Managing_Content/index-katello.html#Products_and_Repositories_content-management') : undefined;
```
jeremylenz, line 273: I'm not sure of the correct way to do this documentation link.

webpack/utils/useKatelloDocUrl.js
sjha4, line 1: Can you explain to me how the downstream links will work? Does the branding change the root URL and the version automatically?

jeremylenz: Yes, the downstream branding plugin should do the requisite magic. I'll ping you with more details..

sjha4:
Works well..

Documentation link is formed correctly but picks up next foreman version 3.7 documentation for katello. We would need to go back a version but I guess it's only a problem for folks running master. Released versions should be fine.

A small naming change as discussed and this should be good to go. ✔️

sjha4 (approved):
APJ 👍🏼

webpack/utils/useKatelloDocUrl.js
```diff
@@ -0,0 +1,18 @@
...
+import { foremanUrl } from 'foremanReact/common/helpers';
+
+// useKatelloDocUrl('Managing_Content, '#Products_and_Repositories_content-management') =>
+// https://docs.theforeman.org/3.7/Managing_Content/index-katello.html#Products_and_Repositories_content-management
+export const useKatelloDocUrl = (guide = 'Managing_Content', hash = '') => {
+  // in dev you'll have to replace the Foreman version in the url with the latest published one
+  const foremanVersion = useForemanVersion();
+  const rootUrl = `https://docs.theforeman.org/${foremanVersion}/`;
+
+  const section = `${guide}/index-katello.html${hash}`;
```
ekohl, line 12: This doesn't belong in the frontend. By having it in the LinksController we can apply branding, which is crucial for Red Hat to properly supports its customers. Note that Red Hat also uses a different file name and directory structure.

ekohl (commented):
Can we revert the `useKatelloDocUrl` part so we don't start to rely on it? That should be done before it's branched for the next release, or the work within Red Hat downstream will be a lot harder.

jeremylenz:
@ekohl This won't be in released code until Katello 4.9, so we have some time. It does go through the links controller (since `foremanUrl` leads there.) If you want to raise a PR doing it a more correct way (with the corresponding Foreman changes) I'll be happy to review and make sure they get in.

ekohl:
From now on there are less than 3 weeks left until we branch. In my experience that time flies by. And I think it's also a lot of wasted time to introduce temporary API while it's clear we need a permanent one.

ekohl:
Just as I predicted, that this was never picked up and now Katello 4.9 is GA. With my downstream hat on, this is going to require special attention so downstream products don't link to upstream resources.

ekohl:
Sadly nothing happened with this, but I've just opened https://github.com/Katello/katello/pull/11061 to fix it.

PR #10481, opened 2023-03-07
Fixes #35324 - Reset Syspurpose modal state on BreadcrumbSwitcher host switch

jeremylenz:

Pass a `key` to the `SystemPurposeEditModal`. This will cause its local state to be reset when switching hosts with the breadcrumb switcher. See https://beta.reactjs.org/learn/preserving-and-resetting-state#resetting-a-form-with-a-key

I also considered having the key be `id` but I think it doesn't matter

Setup: Have two hosts with content. 

1. Go to Hosts > All Hosts and click on one of them
2. Click Edit on the System Purpose card and change some values, then click Save
3. The card will update. Click Edit again and notice that the form is prefilled with the values you saved.
4. Switch hosts using the breadcrumb switcher.
5. Click Edit on the System Purpose card.

Before this change: The prefilled values on the SystemPurposeEditModal form are those of the previous host.
After this change: The prefilled values on the form are those of the new host. You can switch between hosts freely and the modal's values always reflect the current host.


sjha4 (approved):
1. Modal is able to pre-populate values with host details of the host. ✅ 
2. After switching to another host, the pre-populated values come from the new host. ✅ 
Ack 👍🏼

PR #10482, opened 2023-03-08
Fixes #36182 - Error generating host applied errata report

hao-yu:
- Use get_task_input to fetch task input.
- Skip invalid/deleted errata to prevent error.


wbclark (approved):
This worked nicely in my testing.

sjha4 (approved):

PR #10483, opened 2023-03-08
Fixes #35310 - Add call to action link on no activation keys in global registration form

sjha4:
Adds a call to action link for creating activation keys on global registration form.
no_ac_keys_to_select


sjha4:
[test katello]

chris1984 (approved):
Looks fine and works well, my only comment would be it does not really stand out. Can we make it bold or somehow stand out a bit more?

screenshot-192 168 121 7-2023 03 09-13_33_36


I am fine if we can't and it can be merged, just was curious.

sjha4:
After some discussions offline with @jeremylenz , keeping this as is. Merging!
Thanks @chris1984 @jeremylenz 😸

PR #10484, opened 2023-03-08
Fixes #36156 - Content -> Errata does not display the year for the dates listed under column Updated

ianballou:

Errata Issued and Updated dates now show the year.

image

image

I changed every `short-date-time` to `long-date-time` that made sense. Our other uses of `short-date-time` make sense -- users likely won't need the year for thinks like seeing when a task ran.

Sync some errata and look at both the errata index page and an erratum's details page.


sjha4 (approved):
Ack 👍🏼

PR #10485, opened 2023-03-08
Fixes #36174 - Added additional context to error messages for certain…

qcjames53:
… ACS conditions.

Changed the 'validation failed' error messages for various alternate content source fields:

- `Validation failed: Ssl ca cert must be blank` -> `Validation failed: Ssl ca cert cannot be set for simplified ACS`
- `Validation failed: Ssl client cert must be blank` -> `Validation failed: Ssl client cert cannot be set for simplified ACS`
- `Validation failed: Ssl client key must be blank` -> `Validation failed: Ssl client key cannot be set for simplified ACS`
- `Validation failed: Products must be blank` -> `Validation failed: Products should not be set for any custom or rhui ACS`

n/a

Attempt to create a simplified ACS with the `ssl-ca-cert-id`, `ssl-client-cert-id`, or `ssl-client-key-id` fields populated and note the presence of new error messages. Attempt the same thing with a custom or rhui ACS and setting the `product-ids` field.

### DIFF
--- a/app/models/katello/alternate_content_source.rb
+++ b/app/models/katello/alternate_content_source.rb
@@ -35,7 +35,9 @@ module Katello
     validates :base_url, :subpaths, :upstream_username,
               :upstream_password, if: :simplified?, absence: true
     validates :base_url, if: -> { custom? || rhui? }, presence: true
-    validates :products, if: -> { custom? || rhui? }, absence: true
+    validates :products, if: -> { custom? || rhui? }, absence: {
+      message: "should not be set for any custom or rhui ACS"
+    }
     validates :label, :uniqueness => true
     validates :name, :uniqueness => true, presence: true
     # verify ssl must be validated this way due to presence: <bool> failing on a value of false
@@ -141,13 +143,13 @@ module Katello
       # Simplified ACS's should never have ssl-* params populated
       if simplified?
         if changes.keys.include? "ssl_ca_cert_id"
-          errors.add(:ssl_ca_cert, "must be blank")
+          errors.add(:ssl_ca_cert, "cannot be set for simplified ACS")
         end
         if changes.keys.include? "ssl_client_cert_id"
-          errors.add(:ssl_client_cert, "must be blank")
+          errors.add(:ssl_client_cert, "cannot be set for simplified ACS")
         end
         if changes.keys.include? "ssl_client_key_id"
-          errors.add(:ssl_client_key, "must be blank")
+          errors.add(:ssl_client_key, "cannot be set for simplified ACS")
         end
 
       # Custom and RHUI ACS's should have valid keys where populated
--- a/app/models/katello/alternate_content_source.rb
+++ b/app/models/katello/alternate_content_source.rb
@@ -19,9 +19,10 @@ module Katello
     belongs_to :ssl_ca_cert, inverse_of: :ssl_ca_alternate_content_sources, class_name: "Katello::ContentCredential"
     belongs_to :ssl_client_cert, inverse_of: :ssl_client_alternate_content_sources, class_name: "Katello::ContentCredential"
     belongs_to :ssl_client_key, inverse_of: :ssl_key_alternate_content_sources, class_name: "Katello::ContentCredential"
-    # We breakout ssl-* validation into a function to allow for us to set
-    # the object name correctly in complex error messages
+    # We breakout ssl-* and product validation into a function to allow for us
+    # to set the object name correctly in complex error messages
     validate :validate_ssl_ids
+    validate :validate_products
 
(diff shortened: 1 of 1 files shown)


app/models/katello/alternate_content_source.rb
```diff
@@ -35,7 +35,9 @@ class AlternateContentSource < Katello::Model
     validates :base_url, :subpaths, :upstream_username,
               :upstream_password, if: :simplified?, absence: true
     validates :base_url, if: -> { custom? || rhui? }, presence: true
-    validates :products, if: -> { custom? || rhui? }, absence: true
+    validates :products, if: -> { custom? || rhui? }, absence: {
+      message: "should not be set for any custom or rhui ACS"
```
sjha4, line 39: Cannot seems better for validations IMO..
```suggestion
      message: "cannot be set for any custom or rhui ACS"
```

sjha4:
[test katello]

sjha4:
Do we need to validate a false presence for verify_ssl. A blank value can be treated as falsy?

qcjames53:
> Do we need to validate a false presence for verify_ssl. A blank value can be treated as falsy?

Sorry if I misinterpret your question. `verify_ssl` is allowed to be true or false for custom/rhui ACS types but cannot be nil. For simplified ACS types it must be nil. `presence: <bool>` validation is weird, and we have to use a workaround to test this behavior [[documentation]](https://guides.rubyonrails.org/active_record_validations.html#presence).

Are you happy with the current error messages for `verify_ssl`?

sjha4:
[test katello]

sjha4:
> > Do we need to validate a false presence for verify_ssl. A blank value can be treated as falsy?
> 
> Sorry if I misinterpret your question. `verify_ssl` is allowed to be true or false for custom/rhui ACS types but cannot be nil.

My question was more around why it can't be nil for custom/rhui ACS types creation and a nil would be falsy. During update if verify_ssl isn't passed, it is ignored. Does that make sense?

ianballou:
> My question was more around why it can't be nil for custom/rhui ACS types creation and a nil would be falsy. During update if verify_ssl isn't passed, it is ignored. Does that make sense?

From an organizational standpoint this sounds confusing to me. I view `nil` as "this record attribute is in either an errored state or should be completely ignored". I view `true` or `false` as a setting that has been set and it should be acted on. So if a record attribute is unused, it should be `nil`.  It's true that `nil` is falsy, but I think we should be doing the proper nil / boolean checks to work around this.

If `verify_ssl` is false for custom or RHUI ACS, we'd also need to check for nil since nil could end up getting passed to Pulp. Sticking to the strict logic that I mentioned above would avoid that situation.

If we decide instead to treat `nil` as `false`, I propose the following: create a database migration to disallow verify_ssl being nil, and then migrate all simplified ACSs to have verify_ssl to be false. If we're treating `nil` as `false`, we might as well just stop allowing `nil` to be a possibility.

app/models/katello/alternate_content_source.rb
```diff
@@ -19,9 +19,10 @@ class AlternateContentSource < Katello::Model
     belongs_to :ssl_ca_cert, inverse_of: :ssl_ca_alternate_content_sources, class_name: "Katello::ContentCredential"
     belongs_to :ssl_client_cert, inverse_of: :ssl_client_alternate_content_sources, class_name: "Katello::ContentCredential"
     belongs_to :ssl_client_key, inverse_of: :ssl_key_alternate_content_sources, class_name: "Katello::ContentCredential"
-    # We breakout ssl-* validation into a function to allow for us to set
-    # the object name correctly in complex error messages
+    # We breakout ssl-* and product validation into a function to allow for us
```
sjha4, line 22: You can remove these comments. I think the code should be self explanatory plus maintaining comments will be an overhead later if things change. Things @jlsherrill taught me when I made my first few PRs.. 😄

app/models/katello/alternate_content_source.rb
```diff
@@ -165,5 +163,13 @@ def validate_ssl_ids
         end
       end
     end
+
+    # Check that products are not present for custom or rhui ACS's
+    # Requires complex/custom error message
```
sjha4, line 168: Comments not necessary..`errors.add(:product_ids, "cannot be set for custom or rhui ACS")` is good enough explanation for readability..

sjha4:
```
validates :verify_ssl, if: -> { custom? || rhui? }, inclusion: {
      in: [true, false],
      message: "can't be blank"
    }
    validates :verify_ssl, if: :simplified?, inclusion: {
      in: [nil],
      message: "must be blank"
    }
```
Can we add the acs type in error messages for verify-ssl as well.

sjha4 (approved):

PR #10486, opened 2023-03-09
Fixes #36176 - Show N/A for empty product list

sjha4:
We show N/A for empty values on ACS details panel. Products should behave the same.
1. Create a simplified ACS with some product.
2. Edit the ACS to remove all products.
3. Details panel > Products should display N/A instead of blank.


jeremylenz (approved):
thanks @sjha4 

Works fine! LGTM :+1:

PR #10488, opened 2023-03-09
Refs #35763 - cleanup for switching to PF4 search input

lfu:
Remove the usage of [webpack](https://github.com/Katello/katello/tree/master/webpack)/[components](https://github.com/Katello/katello/tree/master/webpack/components)/[Search](https://github.com/Katello/katello/tree/master/webpack/components/Search)/Search.js

Followup of https://github.com/Katello/katello/pull/10401

Check all the search field work well in new host details pages and content view components.
Verify CV filter rules add/edit still work for packages and container image tag.
Verify it does not break anything :)


package.json
```diff
@@ -52,6 +52,7 @@
     "downshift": "^5.4.2",
     "ngreact": "^0.5.0",
     "query-string": "^6.1.0",
+    "react-autocomplete": "^1.8.1",
```
jeremylenz, line 55: I think this will need an ok from Packaging. Also, this component seems to be [no longer maintained](https://github.com/reactjs/react-autocomplete).  How painful of an effort would it be to use Patternfly components instead?

lfu: Are you referring to `SearchInput`?

jeremylenz: I'm referring to the `Autocomplete` component which is imported here from `react-autocomplete`. Can this be done without bringing in any new libraries (e.g., with no changes to `package.json`)?

jeremylenz: As for what existing Patternfly component(s) will replace it, I haven't researched that; I'll leave that up to you to tell me if it's doable or not.

lfu:
@jeremylenz Updated.

package.json
lfu: Seems Patternfly doesn't have an equivalent component for `react-autocomplete`. 
`SearchInput` is the best without a new react library. 
I would suggect https://github.com/JedWatson/react-select if current implementation with `SearchInput` does not look good.

jeremylenz (approved):
Great work @lfu!

Autocomplete on filter rule modal works well. As long as tests are passing this LGTM.

APJ :+1:

PR #10494, opened 2023-03-20
Fixes #35358 - simplified acs products not removed

qcjames53:
Products will now be removed from a simplified ACS if the product loses all repositories with populated URLs. This is handled whether the repository's URL is updated to nil or it is removed entirely.

N/A

1.  Using hammer, create a simplified ACS and an empty product.
2. Create two repositories and attatch them to the empty product (replacing product id where needed):
```
hammer repository create --name RepoNoURL --product-id 2 --content-type yum
hammer repository create --name RepoURL --url https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/ --product-id 2 --content-type yum
```
3. Attach this product to the ACS (replacing ids where needed). Query the ACS to ensure the product has been added:
```
hammer alternate-content-source update --id 49 --product-ids 2
hammer alternate-content-source info --id 49
```
4. Update the only repo with a URL so that the field is nil. The product should be removed from the ACS.
```
hammer repository update --name RepoURL --product 35358Product --organization-id 1 --url ""
hammer alternate-content-source info --id 49
```
5. Reset the repo and add it back to the ACS:
```
hammer repository update --name RepoURL --product 35358Product --organization-id 1 --url https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/
hammer alternate-content-source update --id 49 --product-ids 2
```
6. Delete the repo so that the product contains no assigned URLs. The product should be removed from the ACS:
```
hammer repository delete --name RepoURL --product 35358Product --organization-id 1
hammer alternate-content-source info --id 49
```


app/lib/actions/katello/repository/update.rb
```diff
@@ -65,6 +65,8 @@ def run
         private
 
         def handle_alternate_content_sources(repository, create_acs, delete_acs, old_url)
+          # Look at plan_self
+          # research dynflow, foreman tasks
```
sjha4, line 69:
```suggestion
```

app/lib/actions/katello/repository/update.rb
```diff
@@ -76,6 +78,24 @@ def handle_alternate_content_sources(repository, create_acs, delete_acs, old_url
             repository.smart_proxy_alternate_content_sources.each do |smart_proxy_acs|
               plan_action(Pulp3::Orchestration::AlternateContentSource::Delete, smart_proxy_acs, old_url: old_url)
             end
+
+            # If there are no URLs for any of the remaining repos, remove the
+            # product from the ACS. An ACS with only an empty product will not
```
sjha4, line 83:
```suggestion
            # product from the ACS. An ACS product with only repositories with empty URLs will not
```

sjha4 (commented):
Some nitpicks.

app/lib/actions/katello/repository/update.rb
qcjames53: Completely forgot about those, thanks :)

app/lib/actions/katello/repository/destroy.rb
```diff
@@ -56,19 +56,21 @@ def finalize
...
 
-          # Remove the product from the ACS if it's empty.
-          # An ACS with only an empty product will not function correctly
-          ## because there will be no smart_proxy_alternate_content_sources.
-          if product.repositories.with_type(content_type).count == 1
+          # If there are no URLs for any of the remaining repos, remove the
+          # product from the ACS. An ACS with only an empty product will not
+          # function correctly because there will be no
+          # smart_proxy_alternate_content_sources.
+          contains_url = product.repositories.any? do |url_test_repo|
```
sjha4, line 67: We still want to limit this search to repositories of the same content_type as the ACS right?

app/lib/actions/pulp3/alternate_content_source/remove_product.rb
```diff
@@ -0,0 +1,13 @@
+module Actions
+  module Pulp3
+    module AlternateContentSource
+      class RemoveProduct < Pulp3::AbstractAsyncTask
+        def plan(acs, product)
+          acs.products = acs.products - [product]
+          Rails.logger.info _('Removing product %{prod_name} with ID %{prod_id} from ACS %{acs_name} with ID %{acs_id}') %
+            { prod_name: product.name, prod_id: product.id, acs_name: acs.name, acs_id: acs.id }
```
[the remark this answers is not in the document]
sjha4: This task under Pulp3 namespace doesn't make sense to me. I don't see any communications with pulp here. 
Can we move the acs.products = acs.products - [product] and logging to the katello/repo update/destroy actions directly?

app/lib/actions/pulp3/orchestration/alternate_content_source/remove_product.rb
```diff
@@ -0,0 +1,13 @@
...
+      module AlternateContentSource
+        class RemoveProduct < Pulp3::Abstract
+          def plan(acs, product)
+            plan_action(Actions::Pulp3::AlternateContentSource::RemoveProduct, acs, product)
+          end
+        end
+      end
+    end
+  end
+end
```
sjha4, line 13: Same with this file. I am not convinced this needs to be a task.

app/lib/actions/pulp3/alternate_content_source/remove_product.rb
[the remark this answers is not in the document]
qcjames53: Done!

app/lib/actions/katello/repository/update.rb
```diff
@@ -83,6 +91,27 @@ def handle_alternate_content_sources(repository, create_acs, delete_acs, old_url
           end
         end
 
+        def handle_acs_product_removal(repository)
```
sjha4, line 94: can we call this from inside handle_alternate_content_sources inside the `elsif delete_acs` block.

Also, why are we checking this and adding it to content type list, delete_acs being true implies this repo has no url right?
: 
```
if (repository.id == test_repo.id) && repository.url.present?
              repo_content_types.add(repository.content_type)
```

qcjames53: I'll also look into if we can put the function call back where you're suggesting. There have been changes to the tests since I made that move so it's possible everything is kosher now.

qcjames53: I did some testing and it looks like we're all good to put this logic back in the sequence. I believe making the removal of the product an action was what was messing things up. Just pushed fixes, going to mark as resolved.

sjha4:
[test katello]

app/lib/actions/katello/repository/update.rb
sjha4: Still here.. 😸

sjha4 (approved):

sjha4:
Thanks @qcjames53 !! Merging!!🎉

PR #10495, opened 2023-03-22
Fixes #36202 - ACS page shows loading spinner forever

sjha4:
Ensure refresh of page after bulk delete request responds.

The page reloads after the API responds with success on planning of the bulk tasks. That doesn't ensure that all child tasks have been planned as well. The page displays the current state of data it receives, it's possible not all records deleted in bulk, have been deleted in the backend. The reloaded page will still have these ACSs on the UI. A refresh on the page after the bulk task completes will bring the data on the page up to date with backend.

There's a bigger change that could be possible here to track the bulk task. It's not immediately obvious to me how we track child tasks that haven't been planned yet. But I can file an issue to dig deeper into that.

Bulk delete some ACSs.
Page should refresh and you should see a notification with link to bulk task.


jeremylenz (approved):
ACK with reservations :smile: 

Can you please put a link here to the followup issue?

PR #10498, opened 2023-03-24
Fixes #36220 - Use key to reset Foreman SearchBar from Katello

jeremylenz:

Since updating to use Foreman's `<SearchBar />` component, the "Clear search" link has been broken. This is because of two things:

1. The `clearSearch` function was created in `TypeAhead.js`, which we don't use anymore.
2. We don't have access to the Foreman component's internal state from Katello, so we need another way to clear the search input.

So the changes here are

1. Introduce a new `useClearSearch` hook that you'll need to use wherever you use `SearchBar` with `EmptyStateMessage`
2. This hook provides a key you can pass to reset `<SearchBar>` internal state (e.g. clear the text input) when clearSearch is called
6. Moved and renamed `selectHostDetailsClearSearch` to `selectSearchBarClearSearch` to better reflect that its use is not only in host detail pages.

I kept the `clearSearch` function in Redux rather than introducing prop drilling. I feel like this will keep the EmptyStateMessage component more reusable, since it's used both with and without TableWrapper.

I left `TypeAhead.js` (mostly) alone, since that cleanup will be happening in a different PR.

Do a code search for `<SearchBar>` and make sure everything still works everywhere it's used.


sjha4 (approved):
Tested CV and host details. Works as advertised! Also realized the Generic content page doesn't have the common Empty state view implemented. Acking this. Will file an RFE to add it to content page unless it's small enough to add it here. Let me know.

jeremylenz:
> Also realized the Generic content page doesn't have the common Empty state view implemented. Acking this. Will file an RFE to add it to content page unless it's small enough to add it here.

It would require changing `<Table>` to `<TableWrapper>`, so I'll leave that for a separate RFE.

However, I did add a test to ModuleStreamsTab that goes thru the 'Clear search' workflow and confirms that it works. :+1:

webpack/components/extensions/HostDetails/Tabs/__tests__/moduleStreamsTab.test.js
```diff
@@ -93,6 +93,82 @@ test('Can handle no Module streams being present', async (done) => {
...
+    expect(queryByText('Clear search')).not.toBeInTheDocument();
+  });
+
+  assertNockRequest(initialScope);
+  assertNockRequest(badAutoCompleteScope);
+  assertNockRequest(badSearchScope);
+  assertNockRequest(scopeWithoutSearch);
+  assertNockRequest(autocompleteScope, done);
+});
+
```
sjha4, line 171: Test makes sense..Re-ack 👍🏼

PR #10499, opened 2023-03-28
Fixes #36237 - Add associated hosts to activation key

chris1984:

* Added a method named `hosts` to the activation key model to return a mapping of the hosts used by it. The way we do it in the UI is calling the hosts controller and passing in the activation key, hammer is dumb and can't do that unless I make a host command to do that. Since the BZ wanted to see the hosts used by the key, this made more sense.

* N/A

* Check out Hammer PR: https://github.com/Katello/hammer-cli-katello/pull/884
* Create an activation key
* Register a host with an activation key
* run `hammer activation-key info --id X` and see if it shows the associated host

### DIFF
--- a/app/views/katello/api/v2/activation_keys/base.json.rabl
+++ b/app/views/katello/api/v2/activation_keys/base.json.rabl
@@ -32,6 +32,12 @@ child :products => :products do |_product|
   attributes :id, :name
 end
 
+if ::Foreman::Cast.to_bool(params.fetch(:show_hosts, false))
+  child :hosts do
+    attributes :id, :name
+  end
+end
+
 child :host_collections => :host_collections do
   attributes :id
   attributes :name
--- a/app/controllers/katello/api/v2/activation_keys_controller.rb
+++ b/app/controllers/katello/api/v2/activation_keys_controller.rb
@@ -87,6 +87,7 @@ module Katello
     api :GET, "/activation_keys/:id", N_("Show an activation key")
     param :id, :number, :desc => N_("ID of the activation key"), :required => true
     param :organization_id, :number, :desc => N_("organization identifier"), :required => false
+    param :show_hosts, :bool, :desc => N_("Show hosts associated to an activation key"), :required => false
     def show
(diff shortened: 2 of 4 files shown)


sjha4:
Works great!

Left a small question on the hammer PR about adding a flag on the API to show the hosts instead of showing them by default. Something like `show_hosts` which needs to be passed to fetch the hosts.

chris1984:
@sjha4 updated

Without hosts param:
```
[vagrant@hammer ~]$ hammer activation-key show --id 1
Name:                  Custom
Id:                    1
Host Limit:            1 of Unlimited
Auto Attach:           true
Lifecycle Environment: Library
Content View:          Default Organization View
Host Collections:

Content Overrides:

System Purpose:
    Service Level:
    Purpose Usage:
    Purpose Role:
    Purpose Addons:
```

With hosts param:
```
[vagrant@hammer ~]$ hammer activation-key show --show-hosts true --id 1
Name:                  Custom
Id:                    1
Host Limit:            1 of Unlimited
Auto Attach:           true
Lifecycle Environment: Library
Content View:          Default Organization View
Associated Hosts:
 1) Id:   1
    Name: devel.croberts.io
Host Collections:

Content Overrides:

System Purpose:
    Service Level:
    Purpose Usage:
    Purpose Role:
    Purpose Addons:
```

app/views/katello/api/v2/activation_keys/base.json.rabl
```diff
@@ -32,6 +32,12 @@ child :products => :products do |_product|
   attributes :id, :name
 end
 
+if params.key?(:show_hosts)
```
sjha4, line 35:
```suggestion
if ::Foreman::Cast.to_bool(params.fetch(:show_hosts, false))
```

chris1984: Fixed

sjha4 (approved):
Works great. 👍🏼 
Screenshot from 2023-04-13 13-30-02

PR #10501, opened 2023-03-30
Fixes #36256 - Allow user to force profile upload when changing host content view

jeremylenz:

When changing a host's content view, normally the host's package and errata information isn't updated until the next package action or host check-in. This PR adds a checkbox allowing the user to force that update immediately, by running a subscription-manager command on the host via remote execution.

https://www.awesomescreenshot.com/video/16089294?key=99f6cd115d69a6ed97698284c4074dfe

Does this also require a change to the 'Change content source' workflow? I don't think so, because the job invocation that it provides already includes 'subscription-manager facts' which also updates the host. But double-check this to be safe :)

Check out the Foreman Remote Execution PR: https://github.com/theforeman/foreman_remote_execution/pull/803

Load the updated job template: In rails console, run

```
ForemanInternal.all.first.destroy
```

Then run `bundle exec rails db:seed` and restart the server.

Check out this PR and go to the new host detail page > Content view card > Edit content view assignment

Test the following:

1. REX job is started when you check the checkbox and click Save
2. Content view card is updated with the new content view at the beginning of REX job
3. You receive a toast notification when the job is completed
4. On the errata card, Installable errata (NOT applicable errata) will update when the REX job completes
5. The Recent Communication card will update with the new "Last check-in" time
6. None of this happens when you don't check the checkbox

### DIFF
--- a/webpack/components/extensions/HostDetails/Cards/ContentViewDetailsCard/ChangeHostCVModal.js
+++ b/webpack/components/extensions/HostDetails/Cards/ContentViewDetailsCard/ChangeHostCVModal.js
@@ -1,7 +1,7 @@
 import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import { useDispatch, useSelector } from 'react-redux';
-import { Modal, Button, Alert } from '@patternfly/react-core';
+import { Modal, Button, Alert, Checkbox, TextContent, Text, TextVariants } from '@patternfly/react-core';
 import { translate as __ } from 'foremanReact/common/I18n';
 import { STATUS } from 'foremanReact/constants';
 import { useAPI } from 'foremanReact/common/hooks/API/APIHooks';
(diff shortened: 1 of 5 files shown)


jeremylenz:
[test katello]

webpack/components/extensions/HostDetails/Cards/ContentViewDetailsCard/ChangeHostCVModal.js
```diff
@@ -74,7 +77,14 @@ const ChangeHostCVModal = ({
   const { results: contentViewsInEnv = [] } = contentViewsInEnvResponse;
   const canSave = !!(selectedCVForHost && selectedEnvForHost.length);
 
+  const { triggerJobStart } =
+    useRexJobPolling(runSubmanRepos, () => getHostDetails({ hostname: hostName }));
+
   const refreshHostDetails = () => {
+    const shouldForceProfileUpload = forceProfileUpload;
```
sjha4, line 84: Can we work directly with forceProfileUpload here?

jeremylenz:
@sjha4 Updated :+1:

sjha4 (approved):
Ack..Works great..Rex job gets started and updates the /etc/yum.repos.d/redhat.repo file with checkbox checked, updates last checkin and errata cards. Old behavior when left uncehecked as expected. 👍🏼

PR #10504, opened 2023-04-05
Fixes #36277 - Make the event daemon runner safer

ekohl:

Most importantly this wraps the monitor in an executor, which is recommended by Rails. Quoting the [manual](https://guides.rubyonrails.org/threading_and_code_execution.html#wrapping-application-code):

> Each thread should be wrapped before it runs application code, so if your application manually delegates work to other threads, such as via Thread.new or Concurrent Ruby features that use thread pools, you should immediately wrap the block.

In addition to that it uses the atomic `FileUtils.rm_f` instead of first checking if the file exists. Similarly on read it attempts to read the file, and catches the exception if it doesn't exist instead of first checking if it does to avoid a race condition. To be extra safe, it uses the safe operator in case File.read returns nil.

This is based on reading the Rails manual and applying [the Easier to Ask Forgiveness Than Permission (EAFP)](https://realpython.com/python-lbyl-vs-eafp/) approach that's common in Python. I've also seen RuboCop flag `File.unlink(path) if File.exist?(path)` and replace it with `FileUtils.rm_f` since it's atomic.

Hard to say. I've always found this code to be suspect, but https://community.theforeman.org/t/katello-nightly-rpm-pipeline-1633-failed/33071 triggered me to look into it again.

### DIFF
--- a/app/lib/katello/event_daemon/runner.rb
+++ b/app/lib/katello/event_daemon/runner.rb
@@ -11,9 +11,11 @@ module Katello
         end
 
         def pid
-          return unless pid_file && File.exist?(pid_file)
+          return unless pid_file
 
-          File.read(pid_file).to_i
+          File.read(pid_file)&.to_i
+        rescue Errno::ENOENT
+          nil
         end
 
         def pid_file
@@ -44,7 +46,7 @@ module Katello
           @monitor_thread.kill
           @cache.clear
           @services.values.each(&:close)
-          File.unlink(pid_file) if pid_file && File.exist?(pid_file)
+          FileUtils.rm_f(pid_file) if pid_file
         end
 
         def start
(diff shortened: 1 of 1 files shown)


jturel:
While I don't remember where exactly I was adding the `Rails.executor.wrap`, somewhere in my testing with it I was running into deadlocks in development where requests would hang forever. It's something to keep an eye on for a while after this is merged.

@jeremylenz if you wanna give that one more test but start puma with a single worker and thread (`puma -w 1 -t 1` iirc) and verify the app works as usual, this is probably OK to put in nightly and see what happens. 

If you DO see something weird, you can debug using the `Rack::Lock` middleware.

jturel (approved):
APOMTBJ

(ack pending one more test by jeremy)

jeremylenz:
Tested again with `bundle exec puma -w 1 -t 1 -p 3000`

Seems to work fine. Let's get this in and keep monitoring it.

thanks @ekohl @jturel! :slightly_smiling_face:

PR #10505, opened 2023-04-05
Fixes #36278 - Use hostgroup os refresh path in hostgroup edit

adamruzicka:

Previously the code path taken to update OS when one of the katello fields was changed was determined based on presence of dom element matching #hostgroup_parent_id. However, this element was only present if the user was editing a hostgroup and had more than one hostgroup visible in the current context.

The refresh should now get triggered if data-type of element matching #hostgroup_operatingsystem_id is hostgroup which should be a more reliable indicator.

1) Have two LFEs
2) Have a content view
3) Have an OS with available 
4) Promote the content view to both LFEs
5) Create a hostgroup, assign one of the LFEs, the CV, architecture and OS to it
6) Edit the hostgroup, change its LFE

You should see a request being done to /hostgroups/os_selected.

Without this patch, this should only happen if you have multiple hostgroups.

### DIFF
--- a/app/assets/javascripts/katello/hosts/host_and_hostgroup_edit.js
+++ b/app/assets/javascripts/katello/hosts/host_and_hostgroup_edit.js
@@ -125,7 +125,7 @@ KT.hosts.toggle_installation_medium = function(content_view_id) {
       content_view_id = KT.hosts.getSelectedContentView();
     }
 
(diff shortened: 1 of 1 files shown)


chris1984 (approved):
ACK, works great. I am able to see the connection being made with just one hostgroup.

PR #10506, opened 2023-04-06
Fixes #35976 - Fix ACS SSL key content credential association.

chris1984:

* I had the naming wrong of the rabl when I made the initial pr. Fixed the rabl and the html for the ACS main UI.

* Made sure nothing broke

* Check out PR
* Create an SSL cert, key, ca. Let me know if you want what I used or just fill it with dummy data
* Create an ACS and associate the ssl stuff to it
* Visit content credentials and make sure the key shows 1 acs
* Click on the key, and goto ACS and verify it shows the correct ACS

Before:

acs_association_before
acs_main_before

After:

acs_association_after
acs_main_after


chris1984:
@parthaa is this good for an ack?

parthaa (approved):

PR #10507, opened 2023-04-07
Fixes #36178 - Add repository type column to Repo Sets

jeremylenz:

1. Add a _Repository type_ column to the Repository Sets tab in the new host details page. This column is sortable and will show whether a repo is a Red Hat or custom repo.
2. Move the 'Restricted to OS version / arch' labels to this new column.
3. Add a _Repository type_ dropdown, so you can filter results to just custom or just Red Hat repos.
4. Add a `repository_type` URL param, so you can add `repository_type=custom` to the end of the URL and the dropdown will be pre-selected for you.
5. Fix the bug I mention below.

I noticed that if you have repo sets filtered, and use Select All, it would behave as if the selection were unfiltered. Fixing this proved impossible with the Status dropdown*, so I ended up disabling select all for Status. But you can still use the checkbox to select the page. For repository type, I did fix it, but had to add some crazy scoped search custom external methods in a couple places.

*This is due to the complexities around content overrides and how they're not an actual model in our DB that can be queried.

1. Click around and make sure adding and removing overrides works as expected, especially when using the Select All checkbox with filters.
2. Also make sure pagination, Select All, and filter dropdowns work together as expected.

### DIFF
--- a/app/controllers/katello/api/v2/repository_sets_controller.rb
+++ b/app/controllers/katello/api/v2/repository_sets_controller.rb
@@ -26,7 +26,7 @@ module Katello
     param :enabled, :bool, :required => false, :desc => N_("If true, only return repository sets that have been enabled. Defaults to false")
     param :with_active_subscription, :bool, :required => false, :desc => N_("If true, only return repository sets that are associated with an active subscriptions")
     param :organization_id, :number, :desc => N_("organization identifier"), :required => false
-    param :with_custom, :bool, :required => false, :desc => N_("If true, return custom repository sets along with redhat repos")
+    param :with_custom, :bool, :required => false, :desc => N_("If true, return custom repository sets along with redhat repos. Will be ignored if repository_type is supplied.")
     param :activation_key_id, :number, :desc => N_("activation key identifier"), :required => false
     param :host_id, :number, :desc => N_("Id of the host"), :required => false
     param :content_access_mode_all, :bool, :desc => N_("Get all content available, not just that provided by subscriptions.")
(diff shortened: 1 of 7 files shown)


app/controllers/katello/api/v2/repository_sets_controller.rb
```diff
@@ -139,7 +143,10 @@ def index_relation
       end
 
       relation = relation.where(Katello::Content.table_name => {:name => params[:name]}) if params[:name].present?
-      relation = relation.redhat unless ::Foreman::Cast.to_bool(params[:with_custom])
+      unless ::Foreman::Cast.to_bool(params[:repository_type])
```
sjha4, line 146: This seems more consistent with non boolean params. `params[:repository_type].present?`

app/models/katello/content.rb
```diff
@@ -11,6 +11,15 @@ class Content < Katello::Model
...
     scoped_search :on => :cp_content_id, :complete_value => true
+    scoped_search :on => :id, :rename => :redhat, :ext_method => :search_by_redhat, :complete_value => { :true => true, :false => false }, :only_explicit => true
+
+    def self.search_by_redhat(_key, _operator, value)
+      conditions = Arel.sql(value == 'true' ? "#{Provider.table_name}.provider_type = 'Red Hat'" : "#{Provider.table_name}.provider_type != 'Red Hat'")
+      {
+        :conditions => conditions, :order => "#{Product.table_name}.name",
+        :joins => {:product_contents => {:product => :provider}}
+      }
+    end
```
sjha4, line 23: Where is this scoped_search used? Are we using it from the UI?

sjha4 (approved):
Works as described. Able to select/deselect all with repository type filter and page with status filter.
Able to filter out custom/redhat type repository sets.
Able to see restricted by arch/os labels in the repository type column.
Code looks good 👍🏼

PR #10508, opened 2023-04-10
Fixes #36285 - Update katello CI to use node 14

sjha4:
Update katello CI react tests to use node 14.


jeremylenz (approved):
APJ :+1:

PR #10509, opened 2023-04-10
Fixes #36287 - Manifest import uses async candlepin API

lfu:
Fix broken tests to use async candlepin API to import manifest.

Found the broken tests when upgrading Candlepin to 4.3.1.
The issues exist with current Candlepin version 4.2.13.

`record=true mode=all ktest ~/katello/test/scenarios/scenario_test.rb`


jeremylenz (approved):
I'm not sure about the `sleep`s but I suppose the test manifest size won't change, so I guess it's okay. Ack :+1:

chris1984 (approved):
Looks good to me, glad vcr tests hit an issue

PR #10511, opened 2023-04-11
Fixes #36289 - Hidden CV version number in CVv comparison

sjha4:
Steps to Reproduce:
1. Sync a repository that contains some erratas (e.g. RHEL8 BaseOS)
2. Create a Content view, add repositories from previous step, publish at least two Content view versions
3. Compare the two Content view versions and go to "Errata"

Actual results:
Some erratas have too long Title and Type so headers of other columns get shortened. Headers of last two columns that should show the compared versions are shortened to "Version...", so the version numbers are not visible.

Expected results:
The version numbers should be always visible.

Before change:
Before_PR

After change:
After_PR


lfu (approved):
Nice 👍🏻

PR #10514, opened 2023-04-13
Fixes #36297 - Enable tracer on debian/ubuntu is always on

sbernhard:


jeremylenz (approved):
Code LGTM and this has been tested and verified by two upstream community members: https://community.theforeman.org/t/view-traces-for-ubuntu-hosts/32951

PR #10502, opened 2023-03-31
Fixes #36120 - Change default enablement for custom products

jeremylenz:

PR #10515, opened 2023-04-13
Fixes #36301 - Auto-create content overrides when turning on SCA

jeremylenz:

The upgrade rake task in https://github.com/Katello/katello/pull/10502 executes the following as its first step:

_For non-SCA orgs only, add disabled overrides ("Override to disabled") for any custom repos for which hosts and AKs do not have an existing subscription. This will allow repo enablement state to be preserved when switching on SCA._

I'm extracting this part into ~~its own rake task that you can run at will~~ a thing that happens automatically when you turn on SCA.

image

 This operation will be useful for users switching to SCA, _regardless_ of the current state of your custom product enablement default. In other words:

1. Pre-Katello 4.9 users have custom products enabled by default. These users should be able to switch to SCA and prevent hosts suddenly getting access to content that they've never had a subscription for.
2. Users upgrading to Katello 4.9 already receive this benefit as part of https://github.com/Katello/katello/pull/10502.

I realized that this change is only useful for Katello pre-4.9 and Satellite pre-6.14. So I made sure not to touch any of the same files as in the previous PR, for easier backporting. This results in some duplicated code, but it should be harmless. Maybe we can cherry-pick this only into the older branches.

The override creation task runs by default when you enable SCA. While I think this should be safe, I'm adding a param to the simple_content_access controller so that you can skip this if you need to, using Hammer or the API:

```
hammer simple-content-access enable --organization-id 1 --auto-create-overrides=false
```

Get at least one non-SCA organization with a host and AK
Remove any content overrides
Turn on SCA.

You should now have disabled overrides ("Disabled (Overridden)") on your activation keys and hosts for any custom product that did not have a subscription attached.

Custom products with an attached subscription should not have an override.
Custom products that already had an override should be unchanged.

Run the hammer command with `--auto-create-overrides=false` and verify that the Dynflow console's Plan tab does _not_ show the `PrepareContentOverrides` action.

image

### DIFF
--- a/app/controllers/katello/api/v2/simple_content_access_controller.rb
+++ b/app/controllers/katello/api/v2/simple_content_access_controller.rb
@@ -26,8 +26,10 @@ module Katello
     api :PUT, "/organizations/:organization_id/simple_content_access/enable",
       N_("Enable simple content access for a manifest")
     param :organization_id, :number, :desc => N_("Organization ID"), :required => true
+    param :auto_create_overrides, :bool, :desc => N_("Automatically create disabled content overrides for custom products which do not have an attached subscription"), :required => false, :default => true
     def enable
-      task = async_task(::Actions::Katello::Organization::SimpleContentAccess::Enable, params[:organization_id])
+      auto_create = params.key?(:auto_create_overrides) ? ::Foreman::Cast.to_bool(params[:auto_create_overrides]) : true
+      task = async_task(::Actions::Katello::Organization::SimpleContentAccess::Enable, params[:organization_id], auto_create_overrides: auto_create)
       respond_for_async :resource => task
     end
 
--- a/app/lib/actions/katello/organization/simple_content_access/prepare_content_overrides.rb
+++ b/app/lib/actions/katello/organization/simple_content_access/prepare_content_overrides.rb
@@ -1,0 +1,36 @@
+module Actions
+  module Katello
+    module Organization
+      module SimpleContentAccess
+        class PrepareContentOverrides < Actions::Base
+          def plan(organization_id)
+            Rails.logger.info "PrepareContentOverrides plan: #{organization_id.inspect}"
+            organization = ::Organization.find(organization_id.to_i)
+            org_name = organization.name
+
+            plan_self(organization_id: organization_id, organization_name: org_name)
+          end
+
+          def run
+            organization = ::Organization.find(input[:organization_id].to_i)
+            migrator = ::Katello::Util::ContentOverridesMigrator.new(organization: organization)
+
+            output[:migrator_result] = migrator.execute_non_sca_overrides!
+          end
+
+          def rescue_strategy
+            Dynflow::Action::Rescue::Skip
+          end
+
+          def humanized_name
+            N_("Prepare content overrides for Simple Content Access")
+          end
+
+          def humanized_input
+            _("for organization %s") % input[:organization_name]
+          end
+        end
+      end
+    end
+  end
+end
--- a/app/lib/actions/katello/organization/simple_content_access/toggle.rb
+++ b/app/lib/actions/katello/organization/simple_content_access/toggle.rb
@@ -12,9 +12,15 @@ module Actions
           attr_reader :organization
 
           def plan(organization_id)
-            @organization = ::Organization.find(organization_id)
+            organization = ::Organization.find(organization_id.to_i)
+            input[:organization_name] = organization.name
+            input[:organization_label] = organization.label
             action_subject organization
-            ::Katello::Resources::Candlepin::Owner.update(@organization.label, contentAccessMode: content_access_mode_value)
+            plan_self(organization_id: organization_id)
+          end
+
+          def run
+            ::Katello::Resources::Candlepin::Owner.update(input[:organization_label], contentAccessMode: content_access_mode_value)
           end
 
           def failure_notification(plan)
@@ -28,6 +34,10 @@ module Actions
             task_success_notification.deliver!(
               subject_organization
             )
+          end
+
+          def humanized_input
+            _("for organization %s") % input[:organization_name]
           end
 
           private
--- a/app/lib/katello/util/content_overrides_migrator.rb
+++ b/app/lib/katello/util/content_overrides_migrator.rb
@@ -1,0 +1,98 @@
+module Katello
+  module Util
+    class ContentOverridesMigrator # used in Actions::Katello::Organization::SimpleContentAccess::PrepareContentOverrides
+      include ActionView::Helpers::TextHelper
+
+      def initialize(organization:)
+        @organization = organization
+      end
+
+      def execute_non_sca_overrides!
+        host_errors = create_disabled_overrides_for_non_sca_org_hosts(organization: @organization)
+        ak_errors = create_disabled_overrides_for_non_sca_org_activation_keys(organization: @organization)
+
+        total_errors = host_errors + ak_errors
+        finish_message = "Finished creating overrides in non-SCA orgs; #{total_errors == 0 ? "no errors" : "#{pluralize(total_errors, "error")}"}"
+        messages = { result: finish_message, errors: total_errors }
+        messages[:host_errors] = "Hosts - #{pluralize(host_errors, "error")} creating disabled overrides for unsubscribed content; see log messages above" if host_errors > 0
+        messages[:ak_errors] = "Activation keys - #{pluralize(ak_errors, "error")} creating disabled overrides for unsubscribed content; see log messages above" if ak_errors > 0
+        messages[:success_message] = "You may now switch all organizations to Simple Content Access mode without any change in access to content." if total_errors == 0
+        Rails.logger.info finish_message
+        Rails.logger.info messages[:host_errors] if messages[:host_errors]
+        Rails.logger.info messages[:ak_errors] if messages[:ak_errors]
+        Rails.logger.info messages[:success_message] if messages[:success_message]
+        messages
+      end
+
+      def create_disabled_overrides_for_non_sca(consumable:)
+        content_finder = ::Katello::ProductContentFinder.new(
+                match_subscription: false,
+                match_environment: false,
+                consumable: consumable
+              )
+        subscribed_content_finder = ::Katello::ProductContentFinder.new(
+          match_subscription: true,
+          match_environment: false,
+          consumable: consumable
+        )
+        candlepin_resource = consumable.is_a?(::Katello::Host::SubscriptionFacet) ? ::Katello::Resources::Candlepin::Consumer : ::Katello::Resources::Candlepin::ActivationKey
+        consumable_id = consumable.is_a?(::Katello::Host::SubscriptionFacet) ? consumable.uuid : consumable.cp_id
+        repos_with_existing_overrides = candlepin_resource.content_overrides(consumable_id).map do |override|
+          override[:contentLabel]
+        end
+        unsubscribed_content = content_finder.custom_content_labels - subscribed_content_finder.custom_content_labels - repos_with_existing_overrides
+        new_overrides = unsubscribed_content.map do |repo_label|
+          ::Katello::ContentOverride.new(
+            repo_label,
+            { name: "enabled", value: "0" } # Override to disabled
+          )
+        end
+        return if new_overrides.blank?
+        if consumable.is_a? ::Katello::Host::SubscriptionFacet
+          ::Katello::Resources::Candlepin::Consumer.update_content_overrides(
+            consumable.uuid,
+            new_overrides.map(&:to_entitlement_hash)
+          )
+        else
+          ::Katello::Resources::Candlepin::ActivationKey.update_content_overrides(
+            consumable.cp_id,
+            new_overrides.map(&:to_entitlement_hash)
+          )
+        end
+      end
+
+      def create_disabled_overrides_for_non_sca_org_hosts(organization:)
+        errors = 0
+        fail _("Organization must be specified") if organization.blank?
+        return 0 if organization.simple_content_access? # subscription attachment is meaningless with SCA
+        Rails.logger.info("Hosts - Creating disabled overrides for unsubscribed content in organization #{organization.name}")
+        # only registered hosts with content!
+        hosts_to_update = organization.hosts.joins(:subscription_facet).where.not("#{Katello::Host::SubscriptionFacet.table_name}.host_id" => nil)
+        hosts_to_update.each do |host|
+          create_disabled_overrides_for_non_sca(consumable: host.subscription_facet)
+        rescue => e
+          errors += 1
+          Rails.logger.error("Failed to update host #{host.name}: #{e.message}")
+          Rails.logger.debug e.backtrace.join("\n")
+        end
+        errors
+      end
+
+      def create_disabled_overrides_for_non_sca_org_activation_keys(organization:)
+        errors = 0
+        fail _("Organization must be specified") if organization.blank?
+        return 0 if organization.simple_content_access? # subscription attachment is meaningless with SCA
+        Rails.logger.info("Activation keys - Creating disabled overrides for unsubscribed content in organization #{organization.name}")
+        aks_to_update = organization.activation_keys
+        aks_to_update.each do |ak|
+          create_disabled_overrides_for_non_sca(consumable: ak)
+        rescue => e
+          errors += 1
+          Rails.logger.error("Failed to update activation key #{ak.name}: #{e.message}")
+          Rails.logger.debug e.backtrace.join("\n")
+        end
+        errors
+      end
+    end
+  end
+end
--- a/app/lib/actions/katello/organization/simple_content_access/enable.rb
+++ b/app/lib/actions/katello/organization/simple_content_access/enable.rb
@@ -3,6 +3,16 @@ module Actions
     module Organization
       module SimpleContentAccess
         class Enable < Toggle
+          def plan(organization_id, auto_create_overrides: true)
+            input[:auto_create_overrides] = auto_create_overrides
+            sequence do
+              if auto_create_overrides
+                plan_action(PrepareContentOverrides, organization_id)
+              end
+              super(organization_id) # puts plan_self inside the sequence
+            end
+          end
+
           def content_access_mode_value
             SIMPLE_CONTENT_ACCESS_ENABLED_VALUE
           end


ianballou (commented):
It looks like this fails for all non-registered hosts: 
```
2023-04-19T18:28:06 [I|app|] Failed to update host centos8-katello-devel.example.com: undefined method `organization' for nil:NilClass
````

This error blocks the ending message from coming through.

It's a bit difficult to find the errors in the logs above since the logging is quite verbose. Perhaps changing the errors to be `Rails.logger.error` instead of `info` would help.

File: app/lib/katello/util/content_overrides_migrator.rb [hunk 1]
```diff
@@ -0,0 +1,89 @@
...
+        errors = 0
+        Organization.all.each do |org|
+          next if org.simple_content_access? # subscription attachment is meaningless with SCA
+          Rails.logger.info("Hosts - Creating disabled overrides for unsubscribed content in org #{org.name}")
+          hosts_to_update = org.hosts.where.not(subscription_facet: nil)
+          hosts_to_update.each do |host|
+            create_disabled_overrides_for_non_sca(consumable: host.subscription_facet)
+          rescue => e
+            errors += 1
+            Rails.logger.info("Failed to update host #{host.name}: #{e.message}")
```
ianballou, line 60:
```suggestion
            Rails.logger.error("Failed to update host #{host.name}: #{e.message}")
```

app/lib/katello/util/content_overrides_migrator.rb
```diff
@@ -0,0 +1,89 @@
...
+          end
+        rescue => e
+          errors += 1
+          Rails.logger.info("Error while creating host overrides: #{e.message}")
```
ianballou, line 64:
```suggestion
          Rails.logger.error("Error while creating host overrides: #{e.message}")
```

app/lib/katello/util/content_overrides_migrator.rb
```diff
@@ -0,0 +1,89 @@
...
+        errors = 0
+        Organization.all.each do |org|
+          next if org.simple_content_access? # subscription attachment is meaningless with SCA
+          Rails.logger.info("Activation keys - Creating disabled overrides for unsubscribed content in org #{org.name}")
+          aks_to_update = org.activation_keys
+          aks_to_update.each do |ak|
+            create_disabled_overrides_for_non_sca(consumable: ak)
+          rescue => e
+            errors += 1
+            Rails.logger.info("Failed to update activation key #{activation key.name}: #{e.message}")
```
ianballou, line 79:
```suggestion
            Rails.logger.error("Failed to update activation key #{activation key.name}: #{e.message}")
```

app/lib/katello/util/content_overrides_migrator.rb
```diff
@@ -0,0 +1,89 @@
...
+          end
+        rescue => e
+          errors += 1
+          Rails.logger.info("Error while creating non-SCA activation key overrides: #{e.message}")
```
ianballou, line 83:
```suggestion
          Rails.logger.error("Error while creating non-SCA activation key overrides: #{e.message}")
```

ianballou:
I discovered that the Product Name is wrong in the repo sets list! I guess it's an old legacy page anyway...

image

lib/katello/tasks/prepare_content_overrides_for_sca.rake
```diff
@@ -0,0 +1,7 @@
+namespace :katello do
+  desc "Add disabled content overrides for custom products which do not have a subscription attachment"
```
ianballou, line 2: Do you think it would be worth adding a bit about why you would need this rake task?

jeremylenz:
@iballou Thanks, updated!

jeremylenz:
also added the same query fix in https://github.com/Katello/katello/pull/10525

ianballou (commented):
A first-pass test shows that this is working correctly!  It would be good to get this in tomorrow, so someone else can finish the code review and also try testing.

I tested changing an org to SCA that had a non-registered host and a CentOS host with one product subscribed to it. The other product's repo was overridden to be disabled, and the RH products were untouched.

jeremylenz:
fixed rubocop

app/lib/actions/katello/organization/simple_content_access/prepare_content_overrides.rb
```diff
@@ -0,0 +1,26 @@
...
+    module Organization
+      module SimpleContentAccess
+        class PrepareContentOverrides < Actions::Base
+          def plan(organization_id)
+            @organization = ::Organization.find(organization_id.to_i)
+            @org_name = @organization.name
+            migrator = ::Katello::Util::ContentOverridesMigrator.new(organization: @organization)
+
+            input[:migrator_result] = migrator.execute_non_sca_overrides!
+            plan_self(organization_id: organization_id, organization_name: @org_name)
```
sjha4, line 12: This plan_self would look for a run or finalize phase for the task. I don't see any here. Is this needed?

app/controllers/katello/api/v2/simple_content_access_controller.rb
```diff
@@ -26,8 +26,10 @@ def status
     api :PUT, "/organizations/:organization_id/simple_content_access/enable",
       N_("Enable simple content access for a manifest")
     param :organization_id, :number, :desc => N_("Organization ID"), :required => true
+    param :auto_create_overrides, :bool, :desc => N_("Automatically create disabled content overrides for custom products which do not have an attached subscription"), :required => false, :default => true
     def enable
-      task = async_task(::Actions::Katello::Organization::SimpleContentAccess::Enable, params[:organization_id])
+      auto_create = params.key?(:auto_create_overrides) ? params[:auto_create_overrides] : true
```
parthaa, line 31:
```suggestion
      auto_create = params.key?(:auto_create_overrides) ? Foreman::Cast.to_bool(params[:auto_create_overrides]) : true
```

app/lib/actions/katello/organization/simple_content_access/toggle.rb
```diff
@@ -12,7 +12,8 @@ class Toggle < Actions::AbstractAsyncTask
           attr_reader :organization
 
           def plan(organization_id)
-            @organization = ::Organization.find(organization_id)
+            @organization = ::Organization.find(organization_id.to_i)
+            input[:organization_name] = @organization.name
             action_subject organization
             ::Katello::Resources::Candlepin::Owner.update(@organization.label, contentAccessMode: content_access_mode_value)
```
sjha4, line 18: Looking at this whole action chain, there seems to be no run/finalize in any of these. All the work is being done in plan phase which is not ideal. The Actions::AbstractAsyncTask also allows skipping failed tasks in run phase but not in plan. Looking at the code the PrepareContentOverrides does a lot of heavylifting including creating content overrides on all hosts and activation keys in candlepin. We should move that to the run phase. And this final Owner.update to SCA can move to the finalize once the run phase is completed. Thoughts?

jeremylenz: Sounds like a good ... plan! :smile:

parthaa: Yeah recommend moving it to run (since it involves candlepin.)

app/lib/actions/katello/organization/simple_content_access/prepare_content_overrides.rb
```diff
@@ -0,0 +1,26 @@
+module Actions
+  module Katello
+    module Organization
+      module SimpleContentAccess
+        class PrepareContentOverrides < Actions::Base
+          def plan(organization_id)
+            @organization = ::Organization.find(organization_id.to_i)
+            @org_name = @organization.name
+            migrator = ::Katello::Util::ContentOverridesMigrator.new(organization: @organization)
```
parthaa, line 9: To add to what @sjha4 said I would move this to the run phase

app/lib/katello/util/content_overrides_migrator.rb
```diff
@@ -0,0 +1,96 @@
...
+        host_errors = create_disabled_overrides_for_non_sca_org_hosts(organization: @organization)
+        ak_errors = create_disabled_overrides_for_non_sca_org_activation_keys(organization: @organization)
+
+        total_errors = host_errors + ak_errors
+        finish_message = "Finished creating overrides in non-SCA orgs; #{total_errors == 0 ? "no errors" : "#{pluralize(total_errors, "error")}"}"
+        messages = { result: finish_message, errors: total_errors }
+        messages[:host_errors] = "Hosts - #{pluralize(host_errors, "error")} creating disabled overrides for unsubscribed content; see log messages above" if host_errors > 0
+        messages[:ak_errors] = "Activation keys - #{pluralize(ak_errors, "error")} creating disabled overrides for unsubscribed content; see log messages above" if ak_errors > 0
+        messages[:success_message] = "You may now switch all organizations to Simple Content Access mode without any change in access to content." if total_errors == 0
+        Rails.logger.info finish_message
```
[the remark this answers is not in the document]
jeremylenz: I found it helpful when debugging. Also I log the backtrace if there's an error

parthaa:
Suggested patch (feel free to veto)
```patch
diff --git a/app/lib/actions/katello/organization/simple_content_access/prepare_content_overrides.rb b/app/lib/actions/katello/organization/simple_content_access/prepare_content_overrides.rb
index e6b692a4a2..3dc67ca2f9 100644
--- a/app/lib/actions/katello/organization/simple_content_access/prepare_content_overrides.rb
+++ b/app/lib/actions/katello/organization/simple_content_access/prepare_content_overrides.rb
@@ -4,12 +4,14 @@ module Actions
       module SimpleContentAccess
         class PrepareContentOverrides < Actions::Base
           def plan(organization_id)
-            @organization = ::Organization.find(organization_id.to_i)
-            @org_name = @organization.name
-            migrator = ::Katello::Util::ContentOverridesMigrator.new(organization: @organization)
+            plan_self(organization_id: organization_id)
+          end
 
-            input[:migrator_result] = migrator.execute_non_sca_overrides!
-            plan_self(organization_id: organization_id, organization_name: @org_name)
+          def run
+            organization = ::Organization.find(input[:organization_id].to_i)
+            migrator = ::Katello::Util::DefaultEnablementMigrator.new(organizations: [organization],
+                                                            non_sca_disabled_overrides_only: true)
+            input[:migrator_result] = migrator.execute!
           end
 
           def humanized_name
diff --git a/app/lib/katello/util/default_enablement_migrator.rb b/app/lib/katello/util/default_enablement_migrator.rb
index 99ef011c4f..2256731137 100644
--- a/app/lib/katello/util/default_enablement_migrator.rb
+++ b/app/lib/katello/util/default_enablement_migrator.rb
@@ -2,10 +2,19 @@ module Katello
   module Util
     class DefaultEnablementMigrator # used in lib/katello/tasks/upgrades/4.9/update_custom_products_enablement.rake
       include ActionView::Helpers::TextHelper
+      def initialize(organizations: [], non_sca_disabled_overrides_only: false)
+        @organizations = organizations || Organization.all
+        @non_sca_disabled_overrides_only = non_sca_disabled_overrides_only
+      end
 
       def execute!
         ds_errors = create_disabled_overrides_for_non_sca_org_hosts
         dsak_errors = create_disabled_overrides_for_non_sca_org_activation_keys
+        if non_sca_disabled_overrides_only
+          Rails.logger.info("#{pluralize(ds_errors, "error")} updating disabled overrides for unsubscribed content; see log messages above") if ds_errors > 0
+          Rails.logger.info("#{pluralize(dsak_errors, "error")} updating disabled overrides for unsubscribed content in activation keys; see log messages above") if dsak_errors > 0
+          return
+        end
         ak_errors = create_activation_key_overrides
         consumer_errors = create_consumer_overrides
         cp_errors = update_enablement_in_candlepin
@@ -54,9 +63,9 @@ module Katello
         end
       end
 
-      def create_disabled_overrides_for_non_sca_org_hosts
+      def create_disabled_overrides_for_non_sca_org_hosts()
         errors = 0
-        Organization.all.each do |org|
+        @organizations.each do |org|
           next if org.simple_content_access? # subscription attachment is meaningless with SCA
           Rails.logger.info("Creating disabled overrides for unsubscribed content in org #{org.name}")
           hosts_to_update = org.hosts.where.not(subscription_facet: nil)
@@ -75,7 +84,7 @@ module Katello
 
       def create_disabled_overrides_for_non_sca_org_activation_keys
         errors = 0
-        Organization.all.each do |org|
+        @organizations.each do |org|
           next if org.simple_content_access? # subscription attachment is meaningless with SCA
           Rails.logger.info("Creating disabled overrides for unsubscribed content in activation keys in org #{org.name}")
           aks_to_update = org.activation_keys
@@ -114,7 +123,7 @@ module Katello
         Rails.logger.info "Creating content overrides for all activation keys"
         ak_errors = 0
         cp_aks_to_update = ::Katello::Resources::Candlepin::ActivationKey.get.map { |ak| ak['id'] }
-        Organization.all.each do |org|
+        @organizations.each do |org|
           all_repos = ::Katello::RootRepository.custom.in_organization(org).map(&:custom_content_label) # update all custom repos
 
           org_aks_to_update = ::Katello::ActivationKey.where(organization: org, cp_id: cp_aks_to_update)
```

app/lib/actions/katello/organization/simple_content_access/toggle.rb
jeremylenz: Moved everything to `run`. I also kept the `Owner.update` in `run`, since it's Candlepin, is that ok?
image

jeremylenz:
@parthaa I'd prefer to keep the DefaultEnablementMigrator changes completely separate.

jeremylenz:
Updated to move ContentOverridesMigrator and Candlepin calls to `run`.

app/controllers/katello/api/v2/simple_content_access_controller.rb
sjha4: Where is this param being passed? Is there some UI change needed to pass this parameter? Or is it hammer only?

jeremylenz:
and fixed rubocop again

app/controllers/katello/api/v2/simple_content_access_controller.rb
jeremylenz: It's API / CLI-only. The default is true so no UI changes were needed.

app/lib/actions/katello/organization/simple_content_access/prepare_content_overrides.rb
```diff
@@ -0,0 +1,36 @@
...
+          def run
+            organization = ::Organization.find(input[:organization_id].to_i)
+            migrator = ::Katello::Util::ContentOverridesMigrator.new(organization: organization)
+
+            output[:migrator_result] = migrator.execute_non_sca_overrides!
+          end
+
+          def rescue_strategy
+            Dynflow::Action::Rescue::Skip
+          end
```
sjha4, line 23: @wbclark : Hoping to piggyback off your testing here. Once you're done with the happy path testing, can you force an error in app/lib/katello/util/content_overrides_migrator.rb > execute_non_sca_overrides! method and test if Dynflow::Action::Rescue::Skip makes sense as rescue strategy. I think org SCA will end up being enabled even if there's an error in the overrides method cause the output of first action isn't passed to second action so skip chaining won't work..

@jeremylenz : Should we just remove this and not skip and fail instead?

jeremylenz: If it can't create the overrides for some reason, I _want_ SCA to still be enabled. That should be what Skip does, right?

sjha4:
Code flow LGTM other than the small comment above. 👍🏼

app/lib/actions/katello/organization/simple_content_access/prepare_content_overrides.rb
```diff
@@ -0,0 +1,36 @@
...
+  module Katello
+    module Organization
+      module SimpleContentAccess
+        class PrepareContentOverrides < Actions::Base
+          def plan(organization_id)
+            Rails.logger.info "PrepareContentOverrides plan: #{organization_id.inspect}"
+            organization = ::Organization.find(organization_id.to_i)
+            org_name = organization.name
+
+            plan_self(organization_id: organization_id, organization_name: org_name)
```
parthaa, line 11: Why the org_name ? Its not used in run
```suggestion
            plan_self(organization_id: organization_id)
```

wbclark (commented):
My test involves an organization which is originally configured for entitlement-based content. I created two custom products, centos8-stream and EPEL. I created an activation key, added the centos8-stream subscription to it, and overrode the two repositories for that product to enabled. I register the host and then perform the following test steps:

1. check initial subscription status

```
[root@centos8-stream ~]# subscription-manager status
+-------------------------------------------+
   System Status Details
+-------------------------------------------+
Overall Status: Current

System Purpose Status: Not Specified
```

2. Consumed subscriptions shows centos8-stream

```
[root@centos8-stream ~]# subscription-manager list --consumed
+-------------------------------------------+
   Consumed Subscriptions
+-------------------------------------------+
Subscription Name:   centos8-stream
Provides:            centos8-stream
SKU:                 680867728281
Contract:            
Account:             
Serial:              4840248055557758867
Pool ID:             4028f90b879ed1fa0187a4727bda00a4
Provides Management: No
Active:              True
Quantity Used:       1
Service Type:        
Roles:               
Service Level:       
Usage:               
Add-ons:             
Status Details:      Subscription is current
Subscription Type:   Standard
Starts:              04/21/2023
Ends:                12/01/2049
Entitlement Type:    Physical
```

3. Available subscriptions shows EPEL

```
[root@centos8-stream ~]# subscription-manager list --available
+-------------------------------------------+
    Available Subscriptions
+-------------------------------------------+
Subscription Name:   EPEL
Provides:            
SKU:                 522265941628
Contract:            
Pool ID:             4028f90b879ed1fa0187a4823e1700b1
Provides Management: No
Available:           Unlimited
Suggested:           1
Service Type:        
Roles:               
Service Level:       
Usage:               
Add-ons:             
Subscription Type:   Standard
Starts:              04/21/2023
Ends:                12/01/2049
Entitlement Type:    Physical
```

4. Both of the centos8-stream repositories are enabled

```
[root@centos8-stream ~]# subscription-manager repos
+----------------------------------------------------------+
    Available Repositories in /etc/yum.repos.d/redhat.repo
+----------------------------------------------------------+
Repo ID:   Default_Organization_centos8-stream_BaseOS4
Repo Name: BaseOS4
Repo URL:  https://centos8-stream-wclark-katello-devel.example.com/pulp/content/Default_Organization/Libr
           ary/custom/centos8-stream/BaseOS4
Enabled:   1

Repo ID:   Default_Organization_centos8-stream_AppStream
Repo Name: AppStream
Repo URL:  https://centos8-stream-wclark-katello-devel.example.com/pulp/content/Default_Organization/Libr
           ary/custom/centos8-stream/AppStream
Enabled:   1
```

5. I edit the organization and enable Simple Content Access

6. The host's subscription status now shows that SCA has been enabled:

```
[root@centos8-stream ~]# subscription-manager status
+-------------------------------------------+
   System Status Details
+-------------------------------------------+
Overall Status: Disabled
Content Access Mode is set to Simple Content Access. This host has access to content, regardless of subscription status.

System Purpose Status: Disabled
```

6. Interestingly, we get the same consumed and available subscriptions  as before when using `subscription-manager list`

```
[root@centos8-stream ~]# subscription-manager list --consumed
+-------------------------------------------+
   Consumed Subscriptions
+-------------------------------------------+
Subscription Name:   centos8-stream
Provides:            centos8-stream
SKU:                 680867728281
Contract:            
Account:             
Serial:              4840248055557758867
Pool ID:             4028f90b879ed1fa0187a4727bda00a4
Provides Management: No
Active:              True
Quantity Used:       1
Service Type:        
Roles:               
Service Level:       
Usage:               
Add-ons:             
Status Details:      Subscription is current
Subscription Type:   Standard
Starts:              04/21/2023
Ends:                12/01/2049
Entitlement Type:    Physical
```

```
[root@centos8-stream ~]# subscription-manager list --available
+-------------------------------------------+
    Available Subscriptions
+-------------------------------------------+
Subscription Name:   EPEL
Provides:            
SKU:                 522265941628
Contract:            
Pool ID:             4028f90b879ed1fa0187a4823e1700b1
Provides Management: No
Available:           Unlimited
Suggested:           1
Service Type:        
Roles:               
Service Level:       
Usage:               
Add-ons:             
Subscription Type:   Standard
Starts:              04/21/2023
Ends:                12/01/2049
Entitlement Type:    Physical
```

7. There are now two entitlement certificates, one for the centos8-stream product which was already present (note the timestamps), and a new entitlement called 'content_access`

```
# ll /etc/pki/entitlement/
total 16
-rw-r--r--. 1 root root 3272 Apr 21 18:28 1728865362103868440-key.pem
-rw-r--r--. 1 root root 3585 Apr 21 18:28 1728865362103868440.pem
-rw-r--r--. 1 root root 3272 Apr 21 18:18 4840248055557758867-key.pem
-rw-r--r--. 1 root root 3659 Apr 21 18:18 4840248055557758867.pem
```

```
[root@centos8-stream ~]# rct cat-cert /etc/pki/entitlement/1728865362103868440.pem 

+-------------------------------------------+
	Entitlement Certificate
+-------------------------------------------+

Certificate:
	Path: /etc/pki/entitlement/1728865362103868440.pem
	Version: 3.4
	Serial: 1728865362103868440
	Start Date: 2023-04-21 17:28:04+00:00
	End Date: 2024-04-21 17:28:04+00:00
	Pool ID: Not Available

Subject:
	CN: f11e3287-09cb-4f48-a56f-d089b2548f4a
	O: Default_Organization

Issuer:
	C: US
	CN: centos8-stream-wclark-katello-devel.example.com
	L: Raleigh
	O: Katello
	OU: SomeOrgUnit
	ST: North Carolina

Product:
	ID: content_access
	Name:  Content Access
	Version: 
	Arch: 
	Tags: 
	Brand Type: 
	Brand Name: 

Order:
	Name: Content Access
	Number: 
	SKU: content_access
	Contract: 
	Account: 
	Service Type: 
	Roles: 
	Service Level: 
	Usage: 
	Add-ons: 
	Quantity: 
	Quantity Used: 1
	Socket Limit: 
	RAM Limit: 
	Core Limit: 
	Virt Only: False
	Stacking ID: 
	Warning Period: 0
	Provides Management: False

Authorized Content URLs:
	/Default_Organization/Library

Content:
	Type: yum
	Name: EPEL8
	Label: Default_Organization_EPEL_EPEL8
	Vendor: Custom
	URL: /Default_Organization/Library/custom/EPEL/EPEL8
	GPG: 
	Enabled: False
	Expires: 1
	Required Tags: 
	Arches: 

Content:
	Type: yum
	Name: AppStream
	Label: Default_Organization_centos8-stream_AppStream
	Vendor: Custom
	URL: /Default_Organization/Library/custom/centos8-stream/AppStream
	GPG: 
	Enabled: False
	Expires: 1
	Required Tags: 
	Arches: 

Content:
	Type: yum
	Name: BaseOS4
	Label: Default_Organization_centos8-stream_BaseOS4
	Vendor: Custom
	URL: /Default_Organization/Library/custom/centos8-stream/BaseOS4
	GPG: 
	Enabled: False
	Expires: 1
	Required Tags: 
	Arches: 
```

```
[root@centos8-stream ~]# rct cat-cert /etc/pki/entitlement/4840248055557758867.pem 

+-------------------------------------------+
	Entitlement Certificate
+-------------------------------------------+

Certificate:
	Path: /etc/pki/entitlement/4840248055557758867.pem
	Version: 3.4
	Serial: 4840248055557758867
	Start Date: 2023-04-21 15:32:58+00:00
	End Date: 2049-12-01 00:00:00+00:00
	Pool ID: 4028f90b879ed1fa0187a4727bda00a4

Subject:
	CN: 6e27cfc33bdd44ee835923098f3c2102
	O: Default_Organization

Issuer:
	C: US
	CN: centos8-stream-wclark-katello-devel.example.com
	L: Raleigh
	O: Katello
	OU: SomeOrgUnit
	ST: North Carolina

Product:
	ID: 680867728281
	Name: centos8-stream
	Version: 
	Arch: ALL
	Tags: 
	Brand Type: 
	Brand Name: 

Order:
	Name: centos8-stream
	Number: 
	SKU: 680867728281
	Contract: 
	Account: 
	Service Type: 
	Roles: 
	Service Level: 
	Usage: 
	Add-ons: 
	Quantity: Unlimited
	Quantity Used: 1
	Socket Limit: 
	RAM Limit: 
	Core Limit: 
	Virt Only: False
	Stacking ID: 
	Warning Period: 0
	Provides Management: False

Authorized Content URLs:
	/Default_Organization/Library/custom/centos8-stream/AppStream
	/Default_Organization/Library/custom/centos8-stream/BaseOS4

Content:
	Type: yum
	Name: AppStream
	Label: Default_Organization_centos8-stream_AppStream
	Vendor: Custom
	URL: /Default_Organization/Library/custom/centos8-stream/AppStream
	GPG: 
	Enabled: False
	Expires: 1
	Required Tags: 
	Arches: ALL

Content:
	Type: yum
	Name: BaseOS4
	Label: Default_Organization_centos8-stream_BaseOS4
	Vendor: Custom
	URL: /Default_Organization/Library/custom/centos8-stream/BaseOS4
	GPG: 
	Enabled: False
	Expires: 1
	Required Tags: 
	Arches: ALL
```

8. And subscription-manager now shows all 3 repos, with the previously enabled 2 being enabled and EPEL being disabled

```
[root@centos8-stream ~]# subscription-manager repos
+----------------------------------------------------------+
    Available Repositories in /etc/yum.repos.d/redhat.repo
+----------------------------------------------------------+
Repo ID:   Default_Organization_centos8-stream_BaseOS4
Repo Name: BaseOS4
Repo URL:  https://centos8-stream-wclark-katello-devel.example.com/pulp/content/Default_Organization/Libr
           ary/custom/centos8-stream/BaseOS4
Enabled:   1

Repo ID:   Default_Organization_centos8-stream_AppStream
Repo Name: AppStream
Repo URL:  https://centos8-stream-wclark-katello-devel.example.com/pulp/content/Default_Organization/Libr
           ary/custom/centos8-stream/AppStream
Enabled:   1

Repo ID:   Default_Organization_EPEL_EPEL8
Repo Name: EPEL8
Repo URL:  https://centos8-stream-wclark-katello-devel.example.com/pulp/content/Default_Organization/Libr
           ary/custom/EPEL/EPEL8
Enabled:   0
```

9. When I unregister the system, the entitlements are removed

```
[root@centos8-stream ~]# subscription-manager unregister
Unregistering from: centos8-stream-wclark-katello-devel.example.com:443/rhsm
System has been unregistered.
```

```
[root@centos8-stream ~]# ll /etc/pki/entitlement/
total 0
```

10. After re-registering the system, there is only a single new entitlement created for content access

```
[root@centos8-stream ~]# ll /etc/pki/entitlement/
total 8
-rw-r--r--. 1 root root 3272 Apr 21 18:48 8881077338115403415-key.pem
-rw-r--r--. 1 root root 3581 Apr 21 18:48 8881077338115403415.pem
```

11. As expected, we see three repositories, and the same combination as before remains enabled:

```
[root@centos8-stream ~]# subscription-manager repos
+----------------------------------------------------------+
    Available Repositories in /etc/yum.repos.d/redhat.repo
+----------------------------------------------------------+
Repo ID:   Default_Organization_EPEL_EPEL8
Repo Name: EPEL8
Repo URL:  https://centos8-stream-wclark-katello-devel.example.com/pulp/content/Default_Organization/Libr
           ary/custom/EPEL/EPEL8
Enabled:   0

Repo ID:   Default_Organization_centos8-stream_AppStream
Repo Name: AppStream
Repo URL:  https://centos8-stream-wclark-katello-devel.example.com/pulp/content/Default_Organization/Libr
           ary/custom/centos8-stream/AppStream
Enabled:   1

Repo ID:   Default_Organization_centos8-stream_BaseOS4
Repo Name: BaseOS4
Repo URL:  https://centos8-stream-wclark-katello-devel.example.com/pulp/content/Default_Organization/Libr
           ary/custom/centos8-stream/BaseOS4
Enabled:   1
```

So the obvious thing I notice isn't immediately related to your PR perhaps, but a more general observation about enabling SCA. The extra entitlement remained on the client after I enabled SCA mode for the organization: rhsmcertd would still eventually (at next check in) remove the stale entitlement certificate, yes? Note that refreshing alone does not remove entitlement certificates, but with `--force` it does:

```
[root@centos8-stream ~]# subscription-manager refresh
All local data refreshed
```

```
[root@centos8-stream ~]# subscription-manager refresh --force
1 local certificate has been deleted.
All local data refreshed
```

I can't really see a way that this could lead to issues for the content host, however, because the previously not-present repos are always disabled in the new entitlement certificate anyway. So the extra stale entitlement (which should not be accepted by the server anyway) could not possibly grant any content that the client didn't already have.

jeremylenz:
> The extra entitlement remained on the client after I enabled SCA mode for the organization: rhsmcertd would still eventually (at next check in) remove the stale entitlement certificate, yes?

Enabling or disabling SCA doesn't actually affect entitlements / subscription attachments under the hood. This was on purpose, so that you could decide disable SCA and still have the same subscription attachments you did when you enabled it. I believe there's a change in the works around this, though, so it may not be the case for much longer.

Note that `subscription-manager repos` only tells you the _resulting_ enablement of your products (default enablement, plus any overrides.) You can't see if the "Enabled: 0" or "Enabled: 1" is the result of an override. To see overrides added by this PR, you'd have to use the web UI, or use `subscription-manager repo-override --list`.

jeremylenz:
Also note that subscriptions and entitlements have nothing to do with content overrides, so no need to re-check those :)

app/lib/actions/katello/organization/simple_content_access/prepare_content_overrides.rb
sjha4: >> That should be what Skip does, right?

Yes..So SCA will be enabled without the overrides? Is there a way to go back and create those overrides?

jeremylenz: Well... you could've back when this was a rake task :grinning: 

I don't anticipate that this will fail except in very rare circumstances, so I don't really consider that a problem.

app/lib/actions/katello/organization/simple_content_access/prepare_content_overrides.rb
jeremylenz: It's used in `humanized_input`

parthaa (approved):
Works for me

app/lib/actions/katello/organization/simple_content_access/prepare_content_overrides.rb
jeremylenz: Also, you can always create overrides manually if you want (the "normal" way)

jeremylenz: Oh actually.. You can just disable and enable SCA again!

wbclark:
> Also note that subscriptions and entitlements have nothing to do with content overrides, so no need to re-check those :)

The override is reflected in the entitlement, though, which is what matters to users.

Before enabling SCA, we have an entitlement for the custom product that was added to the activation key, listing the repos which exist under that product in its content. When there is no override for those repos (I tested this as well, but didn't include this test in my above output) then these use the default enablement as a fallback (i.e. `enabled: 0`).

After enabling SCA, an additional entitlement is created which contains ALL repos. Those which had an override previously are `enabled: 1` the same as before, and those which were not present at all in non-SCA mode are now present with `enabled: 0`.

An alternate test proposed offline by @parthaa was forcing `enabled: 1` to be the default behavior in SCA mode, and checking that overrides get created for repositories to set `enabled: 0` unless it was getting `enabled: 1` from an already existing override (rather than from the default). But in that case, either we'd already be in SCA mode (so the migrator would not run, because we are not changing the content access mode) or if we'd start in non-SCA mode (and the migrator would run and create the overrides when they didn't already exist, which is already confirmed by my test steps above).

So, for an user that already has SCA mode enabled, they'd not have any new overrides created because the migrator would not run. But their issue (all custom products became enabled with SCA) would disappear anyway because it could still fall back to the default enablement for custom products (which is false).

jeremylenz:
Opened https://github.com/Katello/katello/pull/10528 and https://github.com/Katello/katello/pull/10529 as direct backports to Katello 4.8 and 4.7.

wbclark:
Following up on my earlier question, for anyone reading along after the fact: it turns out that it was _intended_ that the old entitlement certificate remains in place when changing to SCA mode, and it doesn't get removed by rhsmcertd (which doesn't `force` the refresh). This allows for users to revert back to non-SCA mode without having to re-entitle their systems. It may be changed at some point in the future when non-SCA mode gets removed.

PR #10517, opened 2023-04-14
Fixes #36303 - disable scopes on CV publish errata calc

hao-yu:
Disable user and location scopes when calculating errata applicability after publishing a content view.

This also fix false positive orphaned content facet detection.


sjha4:
Oh..I didn't realize you had a PR for this. I had a smaller fix: https://github.com/Katello/katello/pull/10519 but I will close that in favor of this.

sjha4 (approved):
Ack.. 👍🏼

PR #10518, opened 2023-04-14
Fixes #36295 - Auto activation key selection if there's only one

girijaasoni:
This PR is blocked by Foreman PR: 
https://github.com/theforeman/foreman/pull/9688

This PR has changes t
o auto select an activation key when there's just one and to remove the "No activation key selected error" on User's arrival to the register form. This pr also has better validation formatting like the exclamation icon on the left of carter and underline on the Select dropdown when there's an error! Screenshots of the changes are attached below:

Screenshot from 2023-04-18 17-10-21

Screenshot from 2023-04-18 17-09-57

### DIFF
--- a/webpack/components/extensions/RegistrationCommands/__tests__/__snapshots__/ActivationKeys.test.js.snap
+++ b/webpack/components/extensions/RegistrationCommands/__tests__/__snapshots__/ActivationKeys.test.js.snap
@@ -11,13 +11,21 @@ exports[`ActivationKeys renders 1`] = `
       Create new activation key
     </a>
   }
+  helperTextInvalidIcon={
+    <ExclamationCircleIcon
+      color="currentColor"
+      noVerticalAlign={false}
+      size="sm"
+    />
+  }
   isRequired={true}
   label="Activation Keys"
   labelIcon={
     <LabelIcon
-      text="Activation key(s) for Subscription Manager."
+      text="Activation key(s) to use during registration"
     />
   }
+  onFocus={[Function]}
   validated="error"
 >
   <Select
--- a/webpack/components/extensions/RegistrationCommands/fields/ActivationKeys.js
+++ b/webpack/components/extensions/RegistrationCommands/fields/ActivationKeys.js
@@ -1,16 +1,21 @@
-
 /* eslint-disable max-len, react/forbid-prop-types */
 import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import {
   FormGroup,
-  Select, SelectOption, SelectVariant,
+  Select,
+  SelectOption,
+  SelectVariant,
 } from '@patternfly/react-core';
 
 import LabelIcon from 'foremanReact/components/common/LabelIcon';
+import { ExclamationCircleIcon } from '@patternfly/react-icons';
 import { sprintf, translate as __ } from 'foremanReact/common/I18n';
 
-import { validateAKField, akHasValidValue } from '../RegistrationCommandsPageHelpers';
+import {
+  validateAKField,
+  akHasValidValue,
+} from '../RegistrationCommandsPageHelpers';
 
 const ActivationKeys = ({
   activationKeys,
@@ -23,14 +28,22 @@ const ActivationKeys = ({
   handleInvalidField,
 }) => {
   const [isOpen, setIsOpen] = useState(false);
+  const [hasInteraction, setHasInteraction] = useState(false);
 
   const updatePluginValues = (keys) => {
     onChange({ activationKeys: keys });
-    handleInvalidField('Activation Keys', akHasValidValue(hostGroupId, pluginValues?.activationKeys, hostGroupActivationKeys));
+    handleInvalidField(
+      'Activation Keys',
+      akHasValidValue(
+        hostGroupId,
+        pluginValues?.activationKeys,
+        hostGroupActivationKeys,
+      ),
+    );
   };
 
   const onSelect = (_e, value) => {
-    if (selectedKeys.find((key => key === value))) {
+    if (selectedKeys.find(key => key === value)) {
       updatePluginValues(selectedKeys.filter(sk => sk !== value));
     } else {
       updatePluginValues([...selectedKeys, value]);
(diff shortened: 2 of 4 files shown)


webpack/components/extensions/RegistrationCommands/fields/ActivationKeys.js
```diff
@@ -41,7 +41,12 @@ const ActivationKeys = ({
   useEffect(() => {
     handleInvalidField('Activation Keys', akHasValidValue(hostGroupId, pluginValues?.activationKeys, hostGroupActivationKeys));
   }, [handleInvalidField, hostGroupId, hostGroupActivationKeys, pluginValues]);
-
+  
+  useEffect(() => {
+    if(activationKeys.length==1){
```
Ron-Lavi, line 46:
```suggestion
    if (activationKeys.length === 1) {
```

webpack/components/extensions/RegistrationCommands/fields/ActivationKeys.js
```diff
@@ -41,7 +41,12 @@ const ActivationKeys = ({
...
+      onChange({ activationKeys: [activationKeys[0]?.name]})
```
Ron-Lavi, line 47: maybe you could reuse the `onSelect` function above instead by passing the value of the activation key?

webpack/components/extensions/RegistrationCommands/fields/ActivationKeys.js
```diff
@@ -41,7 +41,12 @@ const ActivationKeys = ({
...
+    }
+  }, [activationKeys]);
```
Ron-Lavi, line 49: we should avoid objects/arrays as dependencies: https://dmitripavlutin.com/react-useeffect-infinite-loop/#21-avoid-objects-as-dependencies

Ron-Lavi (commented):
Thanks @girijaasoni !

webpack/components/extensions/RegistrationCommands/fields/ActivationKeys.js
girijaasoni: Thanks @Ron-Lavi

webpack/components/extensions/RegistrationCommands/fields/ActivationKeys.js
stejskalleos: @girijaasoni `npm run lint -- --fix` is your friend

girijaasoni: Thanks @stejskalleos

webpack/components/extensions/RegistrationCommands/fields/ActivationKeys.js
```diff
@@ -38,10 +38,16 @@ const ActivationKeys = ({
   };
 
   // Validate field when hostgroup is changed (host group may have some keys)
-  useEffect(() => {
+  useEffect(() => {
```
stejskalleos, line 41: NIT: Trailing space, check the configuration of VS code, `Trim Trailing Whitespace` should be enabled by default if I am not mistaken

webpack/components/extensions/RegistrationCommands/fields/ActivationKeys.js
```diff
@@ -38,10 +38,16 @@ const ActivationKeys = ({
...
 
   // Validate field when hostgroup is changed (host group may have some keys)
-  useEffect(() => {
+  useEffect(() => { 
     handleInvalidField('Activation Keys', akHasValidValue(hostGroupId, pluginValues?.activationKeys, hostGroupActivationKeys));
   }, [handleInvalidField, hostGroupId, hostGroupActivationKeys, pluginValues]);
 
+  useEffect(() => {
+    if (activationKeys?.length === 1) {
+      updatePluginValues([activationKeys[0]?.name]);
```
stejskalleos, line 47: If `activationKeys` is not empty, we can safely assume that the key has a name attribute:
`updatePluginValues([activationKeys[0].name]);`

webpack/components/extensions/RegistrationCommands/fields/ActivationKeys.js
```diff
@@ -88,7 +94,7 @@ ActivationKeys.propTypes = {
 };
 
 ActivationKeys.defaultProps = {
-  activationKeys: undefined,
+  activationKeys: [],
```
stejskalleos, line 97: Note to myself: Why it was `undefined` in the first place?

stejskalleos (changes requested):
Two nits, otherwise code looks good, but haven't tested it yet.

webpack/components/extensions/RegistrationCommands/fields/ActivationKeys.js
girijaasoni: Activation key is set to "[]" in the useEffect of indexJs

stejskalleos: yeah I know, but it was supposed to be `[]` from the beginning ...

Ron-Lavi (commented):
Nice work @girijaasoni !

- [X] When there is only one activation key available during host registration, the Activation Key is automatically selected.

- [X] When multiple activation keys are available during host registration, the Activation Key is not automatically selected.

- [X] The auto-selection of the Activation Key does not interfere with the ability to manually select or deselect the option.

- [ ] There's a warning alert shown initially when user's have multiple AK (not sure if you are going to address it in this PR):
Screenshot - 2023-04-19T093224 456

girijaasoni:
> Nice work @girijaasoni !
> 
> * [x]  When there is only one activation key available during host registration, the Activation Key is automatically selected.
> * [x]  When multiple activation keys are available during host registration, the Activation Key is not automatically selected.
> * [x]  The auto-selection of the Activation Key does not interfere with the ability to manually select or deselect the option.
> * [ ]  There's a warning alert shown initially when user's have multiple AK (not sure if you are going to address it in this PR):
>   Screenshot - 2023-04-19T093224 456

That is done in the foreman pr which blocks this pr https://github.com/theforeman/foreman/pull/9688

webpack/components/extensions/RegistrationCommands/fields/ActivationKeys.js
```diff
@@ -42,14 +43,21 @@ const ActivationKeys = ({
...
+  }, [activationKeys?.length]);
+
   return (
     <FormGroup
       label={__('Activation Keys')}
       fieldId="reg_katello_ak"
       helperText={hostGroupActivationKeys && sprintf('From host group: %s', hostGroupActivationKeys)}
       helperTextInvalid={activationKeys?.length === 0 ? <a href="/activation_keys/new">{__('Create new activation key')}</a> : __('No Activation Keys selected')}
-      validated={validateAKField(hostGroupId, pluginValues?.activationKeys, hostGroupActivationKeys)}
+      helperTextInvalidIcon={<ExclamationCircleIcon />}
```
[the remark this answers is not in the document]
stejskalleos: Why remove it?

webpack/components/extensions/RegistrationCommands/fields/ActivationKeys.js
```diff
@@ -42,14 +43,21 @@ const ActivationKeys = ({
     handleInvalidField('Activation Keys', akHasValidValue(hostGroupId, pluginValues?.activationKeys, hostGroupActivationKeys));
   }, [handleInvalidField, hostGroupId, hostGroupActivationKeys, pluginValues]);
 
+  useEffect(() => {
+    if (activationKeys?.length === 1) {
+      updatePluginValues([activationKeys[0].name]);
+    }
+  }, [activationKeys?.length]);
```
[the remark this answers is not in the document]
stejskalleos: Note to myself: Optional chaining might be required for the case when `activationKeys` is `undefined`

webpack/components/extensions/RegistrationCommands/fields/ActivationKeys.js
stejskalleos: `undefined` is for first form load so the field is not marked as invalid

webpack/components/extensions/RegistrationCommands/index.js
```diff
@@ -73,7 +73,7 @@ export const RegistrationActivationKeys = ({
   isLoading,
 }) => {
   useEffect(() => {
-    onChange({ activationKeys: [] });
+    onChange({ activationKeys: undefined });
```
stejskalleos, line 76: `PropTypes.oneOfType([PropTypes.undefined, PropTypes.array])`

stejskalleos (changes requested):
One nit about `PropTypes` otherwise works fine.

stejskalleos:
ok to test

stejskalleos:
ok to test

webpack/components/extensions/RegistrationCommands/index.js
Ron-Lavi: imho it's not a best practice to pass the value as `undefined`
if we are doing this in order to hide/display proper error on the field, e.g:
hide the error when the user opens the page,
and show it upon interaction with the field,

I propose to add a `const [ wasInteracted, setInteraction ] = useState(false)`
and then use the `wasInteracted` in the right places.

This should give us more control over what we are trying to achieve.

stejskalleos: I'm not against it, it makes sense, just didn't know that it's a bad practice.

@jeremylenz @MariaAga How do you handle touched / untouched form fields in React components?

MariaAga: I used `onBlur={() => setWasFocus(true)`  here: https://github.com/theforeman/foreman_remote_execution/pull/784

Ron-Lavi (commented):
@girijaasoni snapshot test is failing, could you update it?
`npm test -- -u`

webpack/components/extensions/RegistrationCommands/fields/ActivationKeys.js
```diff
@@ -39,47 +52,95 @@ const ActivationKeys = ({
...
   }, [handleInvalidField, hostGroupId, hostGroupActivationKeys, pluginValues]);
 
+  useEffect(() => {
+    if (activationKeys?.length === 1) {
+      updatePluginValues([activationKeys[0].name]);
+    }
+    if (activationKeys?.length === 0) {
+      setHasInteraction(true);
+    }
+  }, [activationKeys?.length]);
```
jeremylenz, line 72: In general, setting state in effects should be avoided if possible.

Rather, setting state in event handlers is much better. For `hasInteraction`, I think we can achieve the same thing by adding an `onFocus` to the `<FormGroup>`:

```js
<FormGroup
      onFocus={() => setHasInteraction(true)}
```

And for the `pluginValues`, I would strongly prefer if that could be updated outside an effect as well, but it seems slot and fill makes that too difficult. So I'm ok with keeping that part.

Either way, this is really good reading: https://react.dev/learn/you-might-not-need-an-effect

jeremylenz, line 72: However this effect ends up, the dependency array needs to be correct :)

webpack/components/extensions/RegistrationCommands/fields/ActivationKeys.js
```diff
@@ -39,47 +52,95 @@ const ActivationKeys = ({
...
 ActivationKeys.propTypes = {
   activationKeys: PropTypes.array,
   selectedKeys: PropTypes.array,
-  hostGroupActivationKeys: PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
+  hostGroupActivationKeys: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.array,
+  ]),
   hostGroupId: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
   pluginValues: PropTypes.objectOf(PropTypes.shape({})),
```
jeremylenz, line 145: I got this console error; seems like this needs an update:

```
Warning: Failed prop type: Invalid prop `pluginValues.activationKeys` of type `array` supplied to `ActivationKeys`, expected `object`
```

jeremylenz (changes requested):
Hi @girijaasoni, I left a couple code comments below :)

jeremylenz:
@girijaasoni any update here? This is still on our review board.

girijaasoni:
> @girijaasoni any update here? This is still on our review board.

Since it was for 6.15, I paused the work on it, back on it now again :D

webpack/components/extensions/RegistrationCommands/__tests__/__snapshots__/ActivationKeys.test.js.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`ActivationKeys renders 1`] = `
 <FormGroup
-  fieldId="activation_keys_field"
+  fieldId="reg_katello_ak"
```
jeremylenz, line 5: I thought QE wanted us to remove "katello" from these? Or did I have that switched around?

webpack/components/extensions/RegistrationCommands/fields/ActivationKeys.js
```diff
@@ -39,48 +52,93 @@ const ActivationKeys = ({
...
+      helperTextInvalid={
+        activationKeys?.length === 0 ? (
+          <a href="/activation_keys/new">{__('Create new activation key')}</a>
+        ) : (
+          __('No Activation Keys selected')
+        )
+      }
+      helperTextInvalidIcon={<ExclamationCircleIcon />}
+      labelIcon={
+        <LabelIcon text={__('Activation key(s) for Subscription Manager.')} />
```
jeremylenz, line 89: Since subscription-manager is going away at some point, this might be a good time to update the wording here.

```suggestion
        <LabelIcon text={__('Activation key(s) to use during registration.')} />
```

webpack/components/extensions/RegistrationCommands/fields/ActivationKeys.js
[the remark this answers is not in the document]
jeremylenz: I don't see the "Create new activation key" link any more. I see no reason to remove it; can you please put it back?

jeremylenz (commented):
The useEffect code is looking much better and the page is mostly working well! :+1:  Just a couple more minor comments below.

jeremylenz (approved):
Tested all three scenarios

:heavy_check_mark: No AKs available
:heavy_check_mark: Only one AK
:heavy_check_mark: Multiple AKs

All works great. Nice work! APJ :+1:

jeremylenz:
[test katello]

jeremylenz:
Test failures unrelated

jeremylenz:
@girijaasoni If you rebase to current master the test failures should go away now :)

girijaasoni:
[test katello]

girijaasoni:
[test katello]

PR #10520, opened 2023-04-14
Refs #36285 - Revert katello CI to use node 12

sjha4:
Reverting to using node 12 for our pipeline since the dependency error should now be resolved.
See if React tests are green on this.


ianballou (approved):
(sn)ack!

PR #10522, opened 2023-04-17
Fixes #36302 - Add option to skip dep solving during inc update in the UI

ianballou:

Adds a checkbox to skip dependency solving during incremental update in the UI.

This was determined to be the best way to speed up incremental update for users.

To the reviewer: how does the wording sound?

1) Subscribe a host to a content view version with errata filtered out.
2) Apply some errata that were filtered out. First, skip dependency solving with the new checkbox.
3) See that dependency solving `false` is sent to Pulp in the MultiCopyContents Dynflow.
4) Apply the errata again (after promoting the original CVV  to Library) using dependency solving this time.
5) Check that dependency solving is `true` this time.


sjha4 (approved):
Screenshot from 2023-04-19 12-46-45
Screenshot from 2023-04-19 12-46-25

Able to control dep solving through the new checkbox. Ack 👍🏼

PR #10524, opened 2023-04-19
Fixes #36316 - Scope LCE selection to content source and organization

jeremylenz:

1. When assigning a content view environment to a host, check that its lifecycle environment is available on the host's content source, and raise an error if not.
2. On the Create Host page, the 'Lifecycle environment' dropdown showed multiple Library entries if you had multiple organizations. This change scopes the lifecycle environment dropdown so it only shows environments in the currently selected organization.
3. In the environment path selector on the Change Content Source page and Edit content view assignment modal, disable lifecycle environments not available in the host's content source.

image

I'm separating the commits to hopefully make this more backportable

If you're ambitious, you can actually set up one or more Smart Proxies with Content  / capsules / content sources / whatever you wanna call 'em.  If not, just edit my code to simulate things :)

Example of simulating:
```rb
--- a/app/views/katello/api/v2/environments/show.json.rabl
+++ b/app/views/katello/api/v2/environments/show.json.rabl
@@ -11,7 +11,7 @@ if @content_source.present?
     {
       :name => @content_source.name,
       :id => @content_source_id,
-      :environment_is_associated => env.content_source_associated?(@content_source)
+      :environment_is_associated => env.name == 'dev' ? false : env.content_source_associated?(@content_source)
     }
   end
 end
```

Verify the following:

1. API error is raised if you try to assign a content view environment to the host, but that lifecycle environment is not associated with the host's content source.
2. On the Create Host page, the lifecycle environment dropdown shows only the currently selected org's environments -- NOT all environments.
3. See item 3 above and make sure that that's happening. Check both locations!! Also notice the pretty info alert

### DIFF
--- a/app/controllers/katello/api/v2/environments_controller.rb
+++ b/app/controllers/katello/api/v2/environments_controller.rb
@@ -132,6 +132,7 @@ module Katello
 
     api :GET, "/organizations/:organization_id/environments/paths", N_("List environment paths")
     param :organization_id, :number, :desc => N_("organization identifier")
+    param :content_source_id, :number, :desc => N_("Show whether each lifecycle environment is associated with the given Smart Proxy id.")
     param :permission_type, String, :desc => <<-DESC
       The associated permission type. One of (readable | promotable)
       Default: readable
@@ -142,7 +143,10 @@ module Katello
                   else
                     @organization.readable_promotion_paths
                   end
-
+      if params[:content_source_id]
+        @content_source_id = params[:content_source_id].to_i
+        @content_source = SmartProxy.authorized(:view_smart_proxies).find(@content_source_id)
+      end
       paths = env_paths.inject([]) do |result, path|
         result << { :environments => [@organization.library] + path.select(&:readable?) }
       end
--- a/app/models/katello/content_view_environment_content_facet.rb
+++ b/app/models/katello/content_view_environment_content_facet.rb
@@ -5,5 +5,17 @@ module Katello
 
     validates :content_view_environment_id, presence: true
     validates :content_facet_id, presence: true, unless: :new_record?
+    validate :ensure_valid_content_source
+
+    def ensure_valid_content_source
+      source = self.content_facet&.content_source
+      return if source&.pulp_primary? # pulp_primary smart proxy always has all content
+      env = self.content_view_environment&.environment
+      hostname = self.content_facet&.host&.name
+      return unless [source, env].all? { |x| x.present? }
+      unless source.lifecycle_environments.include?(env)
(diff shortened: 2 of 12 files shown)


app/views/katello/api/v2/environments/show.json.rabl
```diff
@@ -6,6 +6,16 @@ extends 'katello/api/v2/common/org_reference'
...
 
+if @content_source.present?
+  node :content_source do |env|
+    {
+      :name => @content_source.name,
+      :id => @content_source_id,
+      :environment_is_associated => env.content_source_associated?(@content_source)
+    }
+  end
+end
```
sjha4, line 17: This doesn't seem to work. Content source for an env is supposed to be an array given multiple smart proxies can have multiple environments. Hence will need to change this here..

app/models/katello/content_view_environment_content_facet.rb
```diff
@@ -5,5 +5,17 @@ class ContentViewEnvironmentContentFacet < Katello::Model
 
     validates :content_view_environment_id, presence: true
     validates :content_facet_id, presence: true, unless: :new_record?
+    validate :ensure_valid_content_source
+
+    def ensure_valid_content_source
```
ianballou, line 10: Relating to my other comments, I think this validation is somehow running too early. Perhaps we're not updating the content source, CV, and LCE at the same time? If we update only the content source but not the LCE, I could see this error happening.

sjha4: That makes sense..I don't remember hitting this issue..Let me spin up some boxes.. :)

jeremylenz: Yeah, I don't think I took into account if you're _currently in the process of changing_ the content source.

thanks for picking this up while I was gone :)

sjha4: Pushed a change for this. Basically following the order in the UI. Update content_source first and then update CVEs with the validations.

app/controllers/katello/api/v2/environments_controller.rb
```diff
@@ -132,6 +132,7 @@ def destroy
 
     api :GET, "/organizations/:organization_id/environments/paths", N_("List environment paths")
     param :organization_id, :number, :desc => N_("organization identifier")
+    param :content_source_id, :number, :desc => N_("Show whether each lifecycle environment is associated with the given Smart Proxy id.")
```
ianballou, line 135: Time to make a smart proxies tab on the LCE? XD

ianballou (approved):
Looking fine to me! Tested again the scenario that previously failed and it's all good.

PR #10526, opened 2023-04-21
Refs #30371 - Fix date/time display of sync plans

sbernhard:
The method https://github.com/Katello/katello/blob/master/engines/bastion_katello/app/assets/javascripts/bastion_katello/sync-plans/details/sync-plan-details-info.controller.js#L33 requires that the date-time string look like "2021-10-04 18:57:00 -0400" (two whitespaces, first: date, second: time, third: timezone spec)

If you set the Users timezone to anything else than "Browser timezone", you will get the error:
```
bastion-a61e08734ad66c57bc535a2e0c25acc398bb44a8ac0e0ecd3541a62fa8bccafb.js:5 Timepicker directive: "ng-model" value must be a Date object, a number of milliseconds since 01.01.1970 or a string representing an RFC2822 or ISO 8601 date.```


sbernhard:
Can you please give it a try @sjha4 / @chris1984

chris1984:
> Can you please give it a try @sjha4 / @chris1984

I can look at it

chris1984 (approved):
Works great, ship it

PR #10527, opened 2023-04-21
Fixes #36322 - Added filters_applied flag to the CV version API.

qcjames53:
Fixes #36322 - Added support for old records which don't contain applied_filters data. Will return nil/null as per spec.

Refs #36322 - Added comment + cleanup

When using the API, content view versions will now display a flag for whether the content view contains any applied filters. For new content view versions, this field will be boolean; older versions without the applied_filters field will return `null`.

n/a

This test will require an older content view version created before the applied_filters field was introduced...

1. Create and publish a content view version without any applied filters. Do the same for a CVV with filters applied.
2. Using curl (or a tool like postman), query the CVV without any applied filters at `http://<host>/katello/api/content_view_versions/<CVV ID>`. Note the returned "filters_applied" field will be `false`.

    Command:
    `curl --request GET --user admin:changeme http://<host>/katello/api/content_view_versions/<CVV ID> | jq .filters_applied`

    Result:
    `false`

3. Repeat the above step for the content view version with applied filters. "filters_applied" will be `true`.
4. Repeat the above step for a content view version predating the applied_filters field. "filters_applied" will be `null` or blank depending on your JSON reader.


app/models/katello/content_view_version.rb
```diff
@@ -447,6 +447,11 @@ def add_applied_filters!
       save!
     end
 
+    def filters_applied?
+      # If applied_filters doesn't exist, return nil
+      self.applied_filters.nil? ? nil : self.applied_filters["applied_filters"].length != 0
```
parthaa, line 452: Why not just 
```suggestion
      self.applied_filters.present? 
```

sjha4: +1

parthaa: I guess you want it to be nil for older content view versions where its "unknown"
I wonder if  we can do this

```ruby
def applied_filter_state
    if self.applied_filters.nil? 
       "unknown"
   elsif applied_filters.present?
     "applied"
   else
      "not_applied"
  end
end
```

qcjames53: I wasn't sure if this was a hidden feature of `.present?` so I went ahead and tested. For older CVV's that don't contain an applied_filters field, `.present? ` outputs `false` instead of the spec'ed `null`.

qcjames53: `self.applied_filters.nil? ? nil : self.applied_filters["applied_filters"].present?` works however

app/models/katello/content_view_version.rb
```diff
@@ -447,6 +447,11 @@ def add_applied_filters!
       save!
     end
 
+    def filters_applied?
+      # If applied_filters doesn't exist, return nil
```
parthaa, line 451:
```suggestion
      # For older content view versions we do not know if filters were applied.
      # return nil for those.  
```

app/models/katello/content_view_version.rb
parthaa: How about this
```ruby
   def filters_applied?
       # For older content view versions we do not know if filters were applied.
       # return nil for those. 
       return nil if applied_filters.nil?
       applied_filters.present? 
   end
```

app/models/katello/content_view_version.rb
qcjames53: I pushed down these changes, thanks!

app/models/katello/content_view_version.rb
qcjames53: That looks great to me, thank you. Just pushed in your suggestions.

parthaa (approved):

sjha4:
[test katello]

PR #10530, opened 2023-04-24
Fixes #36324 - Add bookmarks button for subscriptions

kmalyjur:
Adding a bookmark search in content -> subscriptions -> search.

### DIFF
--- a/webpack/scenes/Subscriptions/components/SubscriptionsToolbar/SubscriptionsToolbar.js
+++ b/webpack/scenes/Subscriptions/components/SubscriptionsToolbar/SubscriptionsToolbar.js
@@ -4,6 +4,7 @@ import { Row, Col, Form, FormGroup, Button } from 'patternfly-react';
 import { LinkContainer } from 'react-router-bootstrap';
 import { noop } from 'foremanReact/common/helpers';
 import SearchBar from 'foremanReact/components/SearchBar';
+import { getControllerSearchProps } from 'foremanReact/constants';
 import { translate as __ } from 'foremanReact/common/I18n';
 import { SUBSCRIPTIONS_SERVICE_URL } from '../../SubscriptionConstants';
 
@@ -34,11 +35,8 @@ const SubscriptionsToolbar = ({
         <FormGroup className="toolbar-pf-filter">
           <SearchBar
             data={{
-              autocomplete: {
-                url: '/katello/api/v2/subscriptions/auto_complete_search',
-                apiParams: { ...autocompleteQueryParams },
-              },
-              bookmarks: {},
+              ...getControllerSearchProps('/katello/api/v2/subscriptions', 'searchBar-katello_subscriptions', true, autocompleteQueryParams),
+              controller: 'katello_subscriptions',
             }}
             onSearch={onSearch}
             onSearchChange={updateSearchQuery}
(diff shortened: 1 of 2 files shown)


chris1984:
@kmalyjur did you see this on your box when typing in the search box from the JS console
`to load resource: the server responded with a status of 404 ()`

chris1984 (changes requested):
See my comment above

chris1984:
ok to test

chris1984:
# #
[test katello]

webpack/scenes/Subscriptions/components/SubscriptionsToolbar/SubscriptionsToolbar.js
```diff
@@ -34,11 +35,11 @@ const SubscriptionsToolbar = ({
         <FormGroup className="toolbar-pf-filter">
           <SearchBar
             data={{
+              ...getControllerSearchProps('katello_subscriptions'),
```
lfu, line 38: Correct way to add bookmark here:
```
         <FormGroup className="toolbar-pf-filter">
           <SearchBar
             data={{
-              autocomplete: {
-                url: '/katello/api/v2/subscriptions/auto_complete_search',
-                apiParams: { ...autocompleteQueryParams },
-              },
-              bookmarks: {},
+              ...getControllerSearchProps('/katello/api/v2/subscriptions', 'searchBar-katello_subscriptions', true, autocompleteQueryParams),
+              controller: 'katello_subscriptions',
             }}
             onSearch={onSearch}
             onSearchChange={updateSearchQuery}

**Comment by chris1984** [comment]
@lfu does the react part look ok? If so, I am fine with acking it as well. The Katello test failure is unrelated so will keep kicking it until it passes.

**Comment by chris1984** [comment]
[test katello]

File: webpack/scenes/Subscriptions/components/SubscriptionsToolbar/SubscriptionsToolbar.js [hunk 2]
```diff
@@ -5,6 +5,7 @@ import { LinkContainer } from 'react-router-bootstrap';
 import { noop } from 'foremanReact/common/helpers';
 import SearchBar from 'foremanReact/components/SearchBar';
 import { translate as __ } from 'foremanReact/common/I18n';
+import { getControllerSearchProps } from 'foremanReact/constants';
```
lfu, line 8: Minor, but I'd like move this line up as it goes with `SearchBar`.

chris1984 (approved):
ACK, works great.

bookmark

Will let @lfu give the final ack one the suggestion she wrote is addressed, then we can merge this.

webpack/scenes/Subscriptions/components/SubscriptionsToolbar/SubscriptionsToolbar.js
kmalyjur: Thank you very much for your input. Could I please just have some questions so I understand it better?
- Isn't the line `bookmarks: {},` unnecessary when we override it with `getControllerSearchProps` anyway?
- Overall, why is putting the `autocomplete` first the correct way if the result is the same?

kmalyjur: Thank you very much for your input. I'm sorry about my last response if you read it, I somehow haven't noticed the +/- at the beginning of your lines so I was confused. I thought that my solution is less complicated, but now I see that your way is already used somewhere else in the code and it makes sense.

lfu (approved):
LGTM 👍🏻 
APJ

kmalyjur:
Thank you for your reviews

PR #10531, opened 2023-04-24
Fixes #36326 - Sync redhat-uep.pem with EL8

ekohl:

The Red Hat CDN has rotated their certificate. This should work towards compatibility with the FUTURE crypto-policy.

This takes redhat-uep.pem from CentOS Stream 8. To be precise, subscription-manager-rhsm-certificates-1.28.35-1.el8.x86_64. Similar to 66d9e4e7291d1041b5c2da3d22f10de773e77f18.

Interaction with the Red Hat CDN still works.


ekohl:
Now I do wonder if we should actually patch it in packaging. Make rubygem-katello depend on `subscription-manager-rhsm-certificates` and symlink the file. Though I wonder what that means for development.

ehelms (approved):
Untested.

This is our third update in a decade, so I'm inclined to do nothing more than just perform this update here. If they start rotating it more frequently I'd lean towards a better solution.

ianballou (approved):

PR #10533, opened 2023-04-26
Fixes #35710 - Update AngularJS Docker pages to display correctly

chris1984:
* Created a new repositories tab for each tag when you go into the details
* Fixed Container Image Manifest Lists to show the tag correctly since it was showing anything before
* Verified Container Image Manifests shows tags on a container without a manifest
* Fixed a broken test because of my change and added a test for the new AngularJS controller I added.
* Added a new route in the AngularJS route for the new repositories tab
* Since we are no longer showing the repositories on the index page, we got rid of an awful JS function we were using to grab the repos that didn't work half the time.

* Removed the repositories table from the index page of Container Tags because I added the repositories tab in the details of each tag since it was not pulling the info some of the time
* Removed the repositories label from the details section of the tag since it never worked and was pulling a single value instead of looping through the results and listing them

* Apply the PR
* Create a repo with the Registry URL of `https://quay.io/` and the Upstream Repository `Prometheus/busybox`
* Create a repo with the Registry URL of `https://quay.io/` and the Upstream Repository `quay/busybox`
* Sync said repos
* Look around at the Container Image Tags, Container Image Manifests, Container Image Manifests Lists and make sure everything looks good and the tags are showing up.
* On the Container Image Tags, click on a tag and look at the new repositories tab to make sure it's filled with the correct info.
* Check the `quay/busybox' Container Image Manifests to make sure it shows the tag instead of saying `Part of a manifest`


engines/bastion_katello/app/assets/javascripts/bastion_katello/docker-tags/details/views/docker-tag-repositories.html
```diff
@@ -0,0 +1,62 @@
+
```
lfu, line 1: Extra line :)

lfu (approved):
Other than the blank line, it looks good 👍🏻

PR #10534, opened 2023-04-26
Fixes #36335 - Migrate Transifex Config for v3 API

wbclark:

1. Created transifex account and joined foreman organization
2. Created API token at https://app.transifex.com/user/settings/api/
3. Installed https://github.com/transifex/cli and copied binary to /usr/local/bin/
4. ran `tx migrate` and entered API token when prompted

1. You need a transifex account and API token as described above
2. You need to install https://github.com/transifex/cli as described above
3. You can test the new configuration with `tx pull --minimum-perc 50 --all` or similar


.tx/config
```diff
@@ -1,14 +1,14 @@
 [main]
 host = https://www.transifex.com
```
wbclark, line 2: This is still wrong, it turns out....

...but I fixed it with https://github.com/transifex/cli/pull/187

I'll re-run the migration with the fix so that this gets properly updated to `host = https://app.transifex.com`

ianballou (approved):
tx-pull worked for me!

PR #10539, opened 2023-05-01
Fixes #36333 - Calculate needs_publish for composite content views

sjha4:
Calculate needs_publish for composite content views
The reasons a composite content view (CCV) would need a publish are:
1. CCV has no published versions.
2. A component CV was added/removed to the CCV.
3. A component CV added with `Always update to latest version` set to true got a new version published.
4. A component CV was updated to edit version associated with CCV or set to update to latest when earlier CCV version was published with an older version.

Test the above paths that would need a publish on CCV and make sure the needs_publish returns the correct value for both Composite and component CVs.

### DIFF
--- a/app/models/katello/content_view.rb
+++ b/app/models/katello/content_view.rb
@@ -777,7 +777,18 @@ module Katello
       audited_filter_rules
     end
 
+    def composite_cv_components_changed?
+      return true unless latest_version_object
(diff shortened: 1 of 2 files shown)


ianballou (commented):
Changes look good! Is it worth adding something to `models/content_view_test.rb`?

ianballou (approved):
Looks good!

PR #10542, opened 2023-05-02
Fixes #36334 - Delete orphaned content units after indexing and improve missing content units error on Pulp 3 copy

ianballou:
Ensures that content units belonging to no repositories are destroyed at repo indexing time.  This is necessary because content units that have been removed from Pulp can still remain attached to content view filters, which will cause blocking errors at CV publish time.

I also improved the error returned from Pulp that occurs when trying to copy units that don't exist.

Note that I didn't rewrite the special rpm copy API error. In testing, I found that the rpm copy API ignores content units that don't exist.

I considered if content units should be destroyed during content unit importing, or perhaps if units could be disassociated from filters instead. Finding the filter relations at sync time would be too query-intensive. Determining the content units to remove during `ContentUnitIndexer.import_all` would require searching other repositories to find out if it's truly orphaned, which I ended up doing higher up in the stack anyway.

1) Create a content view with some module streams and container image tags
2) Make filters that filter in those tags and module streams
3) Modify the pulp_hrefs for some module stream and some container image tag to be something that doesn't exist in Pulp
4) Try publishing the CV. You should see a failed task in Dynflow with the error message rewrite that this PR implements.
5) Perform a complete sync on the related repositories.
6) The publish should now work. I think you'll have to skip the failing Dynflow task and try the publish again.

Also, try verifying that orphaned content units are indeed destroyed using something other than module streams and container image tags.

### DIFF
--- a/app/models/katello/concerns/pulp_database_unit.rb
+++ b/app/models/katello/concerns/pulp_database_unit.rb
@@ -167,6 +167,14 @@ module Katello
         where("#{self.table_name}.id in (?) or #{self.table_name}.pulp_id in (?)", id_integers, ids)
       end
 
+      def orphaned
+        if many_repository_associations
+          where.not(:id => repository_association_class.where(:repository_id => ::Katello::Repository.all).select(unit_id_field))
+        else
+          where.not(:repository_id => ::Katello::Repository.all)
+        end
+      end
+
       def in_repositories(repos)
         if many_repository_associations
           where(:id => repository_association_class.where(:repository_id => repos).select(unit_id_field))
--- a/app/services/katello/content_unit_indexer.rb
+++ b/app/services/katello/content_unit_indexer.rb
@@ -154,6 +154,7 @@ module Katello
 
     def clean_filter_rules(repo_associations_to_destroy)
       affected_content_view_ids = @repository.content_views.non_default.pluck(:id)
+      return false if affected_content_view_ids.empty?
       case @model_class.to_s
       when 'Katello::ModuleStream'
         module_stream_ids = repo_associations_to_destroy.pluck(:module_stream_id)
--- a/app/services/katello/content_unit_indexer.rb
+++ b/app/services/katello/content_unit_indexer.rb
@@ -159,20 +159,16 @@ module Katello
       when 'Katello::ModuleStream'
         module_stream_ids = repo_associations_to_destroy.pluck(:module_stream_id)
         filter_rules = ::Katello::ContentViewModuleStreamFilterRule.
-          joins('INNER JOIN katello_content_view_filters ON katello_content_view_module_stream_filter_rules.content_view_filter_id = katello_content_view_filters.id').
-          where("katello_content_view_filters.content_view_id IN (#{affected_content_view_ids.join(',')})").where(module_stream_id: module_stream_ids)
+          in_content_views(affected_content_view_ids).where(module_stream_id: module_stream_ids)
         filter_rules.delete_all
       when 'Katello::Erratum'
         errata_ids = ::Katello::Erratum.where(id: repo_associations_to_destroy.select(:erratum_id)).pluck(:errata_id)
-        filter_rules = ::Katello::ContentViewErratumFilterRule.
-          joins('INNER JOIN katello_content_view_filters ON katello_content_view_erratum_filter_rules.content_view_filter_id = katello_content_view_filters.id').
-          where("katello_content_view_filters.content_view_id IN (#{affected_content_view_ids.join(',')})").where(errata_id: errata_ids)
+        filter_rules = ::Katello::ContentViewErratumFilterRule.in_content_views(affected_content_view_ids).where(errata_id: errata_ids)
         filter_rules.delete_all
       when 'Katello::PackageGroup'
         package_group_uuids = ::Katello::PackageGroup.where(id: repo_associations_to_destroy.select(:package_group_id)).pluck(:pulp_id)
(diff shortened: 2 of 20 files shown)


app/models/katello/concerns/pulp_database_unit.rb
```diff
@@ -167,6 +167,14 @@ def with_identifiers(ids)
         where("#{self.table_name}.id in (?) or #{self.table_name}.pulp_id in (?)", id_integers, ids)
       end
 
+      def orphaned
+        if many_repository_associations
+          where.not(:id => repository_association_class.where(:repository_id => ::Katello::Repository.all).select(unit_id_field))
+        else
+          where.not(:repository_id => ::Katello::Repository.all)
+        end
+      end
```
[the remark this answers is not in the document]
ianballou: You're right, in that case they wouldn't get removed. To fix this, I'll look again at having some sort of "correct filters" method that will run after a repository is synced. That'll ensure that ID filter rules exist only if they relate to content that is in the library repositories.

app/services/katello/content_unit_indexer.rb
```diff
@@ -142,14 +142,42 @@ def fetch_only_ids
...
+          where("katello_content_view_filters.content_view_id IN (#{affected_content_view_ids.join(',')})").where(errata_id: errata_ids)
+        filter_rules.delete_all
+      when 'Katello::PackageGroup'
+        package_group_uuids = ::Katello::PackageGroup.where(id: repo_associations_to_destroy.select(:package_group_id)).pluck(:pulp_id)
+        filter_rules = ::Katello::ContentViewPackageGroupFilterRule.
+          joins('INNER JOIN katello_content_view_filters ON katello_content_view_package_group_filter_rules.content_view_filter_id = katello_content_view_filters.id').
+          where("katello_content_view_filters.content_view_id IN (#{affected_content_view_ids.join(',')})").where(uuid: package_group_uuids)
+        filter_rules.delete_all
+      else
+        return false
```
sjha4, line 177: We have a helper method "rules" for content view filters that is a generic method to return whatever rule class exists on the filter of any type. ` Katello::ContentViewFilter.find(37).rules` 
Would using that simplify the code here and have everything handled with ActiveRecord? I am not against raw SQL but it hurts my brain more than ActiveRecord.

ianballou:
@sjha4 I updated the code to clean filter rules after disassociating module streams, package groups, and errata. I didn't bother putting it in an async task because the disassociation should be pretty rare.

app/services/katello/content_unit_indexer.rb
ianballou: Maybe, I'll check if it's applicable

ianballou: I'm not sure that's useful here since I need to collect rules for an arbitrary number of content view filters.

sjha4: Also I was thinking a task that does this on any given CV without access to indexing data might be helpful. If a module stream exists as a rule in a CV but the library repo has no association to the module stream, we delete it. Shouldn't be a long running task for a single repo CVs especially if we make it async in the finalize. Added benefit would be we can run it directly on CVs if needed to clean up filters rather than rely on indexing action on repo.

ianballou: If we make a rake task I could include it as an upgrade task I suppose. We'll just need to be very sure that it works well since we're deleting things.

ianballou:
@sjha4 I added the rake task.

app/services/katello/content_unit_indexer.rb
```diff
@@ -142,14 +142,42 @@ def fetch_only_ids
     def sync_repository_associations(assocication_tracker, additive: false)
       unless additive
         ActiveRecord::Base.connection.uncached do
-          @model_class.repository_association_class.where(repository_id: @repository.id).where.
-            not(@model_class.unit_id_field => assocication_tracker.unit_ids).delete_all
+          repo_associations_to_destroy = @model_class.repository_association_class.where(repository_id: @repository.id).where.
+            not(@model_class.unit_id_field => assocication_tracker.unit_ids)
+          clean_filter_rules(repo_associations_to_destroy) if repo_associations_to_destroy.present? && [::Katello::ModuleStream, ::Katello::Erratum, ::Katello::PackageGroup].include?(@model_class)
```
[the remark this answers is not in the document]
ianballou: Added a return statement below.

ianballou:
@sjha4 I added rule cleanup when removing repositories from content views. I also did a little refactoring with the content view filter rule queries to dry up the  code.

sjha4:
Code looks good..Will test the updates. 👍🏼

parthaa (approved):
APJ

sjha4 (approved):
1. Able to reproduce the error of missing content unit and task fails with updated error message:
Failed_to_publish
2. After complete sync on failing repository, able to publish successfully
Screenshot from 2023-05-23 16-03-52
3. Filter rules by Id are removed if their repo association is removed during indexing
4. Tested the rake task on missing module streams from master and then switching to this branch and running rake task. Able to delete stream filter rules with the task.

Code looks good. Added a commit to fix tests etc which was APJ'ed by @parthaa 
Ack on the PR. Will merge after tests are green!

sjha4:
[test katello]

ianballou:
@sjha4 my latest push should've fixed the test.

sjha4 (approved):
Ack.. 👍🏼

PR #10543, opened 2023-05-03
Fixes #36362 - Reporting results of docker v2 repo discovery

stbergmann:
A Repository Discovery with custom container images URL <https://registry.fedoraproject.org/> (which does not support the v1 API) reported "No discovered repositories".

### DIFF
--- a/app/lib/katello/repo_discovery.rb
+++ b/app/lib/katello/repo_discovery.rb
@@ -62,7 +62,9 @@ module Katello
         # Note: v2 endpoint does not support search
         request_params[:url] = "#{@uri}v2/_catalog"
         results = RestClient::Request.execute(request_params)
-        @found = JSON.parse(results)['repositories']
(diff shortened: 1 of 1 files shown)


ianballou:
@stbergmann mind making a redmine issue here? https://projects.theforeman.org/projects/katello/issues/new

That way we can triage it to a release properly.

ianballou:
[test katello]

ianballou:
The commit and PR title need the redmine number in it once you make it, you can use this as an example https://github.com/Katello/katello/pull/10544

stbergmann:
> The commit and PR title need the redmine number in it once you make it, you can use this as an example #10544

@ianballou So I need to file some issue in some redmine instance first?  Where exactly?

chris1984:
ok to test

chris1984 (approved):
ACK works, thanks @stbergmann Will merge once Jenkins finishes and is green.

docker

PR #10544, opened 2023-05-03
Fixes #36352 - Add metadata_expire attribute for custom yum repos

parthaa:
Default value for custom yum repo metadata expiration is 1. This means every yum transaction will check on its entitlements, accessible repositories etc if the yum cache is more than a second old.  This works for most cases but may be not ideal for high latency environments. 

This PR provides a clear way to setup  metadata expiration values for custom yum repositories . 
- Added a new metadata expire column to root repository
- This value will be used to set on the repository content in Candlepin.
- Added API controllers
- Added validations (Must be yum and custom, must be >0 and < 86400s =24 hours)

We were trying to solve [Community Issue](https://community.theforeman.org/t/yum-commands-are-very-slow/30174/17 ) 

Initially we were going to do a rake task to set the metadata expiry. But repository update will reset that back to 1. So had to go to this approach

- Checkout PR
- `bundle exec rake db:migrate`
- Create a Custom Yum repo and a content host consuming it
- Check the redhat repo file on the content host and make sure the `metadata_expire` value is set to 1
- Now update this via hammer or api => `hammer repository update --id=10 --metadata-expire=1000`
- On your content host `subscription-manager refresh`
-  check your redhat repo and make sure the `metadata_expire` value is set to 1000
- Make sure you are not able to update repository if its redhat or non yum. Also check for boundary conditions (>0, <=86400)

### DIFF
--- a/app/controllers/katello/api/v2/repositories_controller.rb
+++ b/app/controllers/katello/api/v2/repositories_controller.rb
@@ -71,7 +71,7 @@ module Katello
       param :http_proxy_id, :number, :desc => N_("ID of a HTTP Proxy")
       param :arch, String, :desc => N_("Architecture of content in the repository")
       param :retain_package_versions_count, :number, :desc => N_("The maximum number of versions of each package to keep.")
-
+      param :metadata_expire, :number, :desc => N_("Time to expire yum metadata in seconds. Only relevant for custom yum repositories.")
       RepositoryTypeManager.generic_remote_options(defined_only: true).each do |option|
         param option.name, option.type, :desc => N_(option.description)
       end
@@ -580,7 +580,7 @@ module Katello
     # rubocop:disable Metrics/CyclomaticComplexity
     def repository_params
       keys = [:download_policy, :mirroring_policy, :sync_policy, :arch, :verify_ssl_on_sync, :upstream_password,
-              :upstream_username, :download_concurrency, :upstream_authentication_token,
+              :upstream_username, :download_concurrency, :upstream_authentication_token, :metadata_expire,
               {:os_versions => []}, :deb_releases, :deb_components, :deb_architectures, :description,
               :http_proxy_policy, :http_proxy_id, :retain_package_versions_count, {:ignorable_content => []}
              ]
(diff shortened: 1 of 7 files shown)


app/controllers/katello/api/v2/repositories_controller.rb
```diff
@@ -71,7 +71,7 @@ def custom_index_relation(collection)
       param :http_proxy_id, :number, :desc => N_("ID of a HTTP Proxy")
       param :arch, String, :desc => N_("Architecture of content in the repository")
       param :retain_package_versions_count, :number, :desc => N_("The maximum number of versions of each package to keep.")
-
+      param :metadata_expire, :number, :desc => N_("time to expire yum metadata in seconds. Only relevant for custom yum repositories.")
```
chris1984, line 74: `param :metadata_expire, :number, :desc => N_("Time to expire yum metadata in seconds. Only relevant for custom yum repositories.")`

app/models/katello/root_repository.rb
```diff
@@ -322,6 +333,13 @@ def ensure_valid_retain_package_versions_count
       end
     end
 
+    def ensure_metadata_expire_valid
+      return if self.metadata_expire.blank?
+      unless yum? && custom?
+        errors.add(:metadata_expire, N_("is only allowed for Non-Redhat Yum repositories."))
```
chris1984, line 339: `errors.add(:metadata_expire, N_("Is only allowed for Non-Redhat Yum repositories."))`

chris1984 (approved):
ACK left 2 comments on grammar fixes, works great:

Before:
```bash
[Default_Organization_Custom_Zoo]
name = Zoo
baseurl = https://devel.croberts.io/pulp/content/Default_Organization/Library/Partha/custom/Custom/Zoo
enabled = 1
gpgcheck = 0
sslverify = 1
sslcacert = /etc/rhsm/ca/katello-server-ca.pem
sslclientkey = /etc/pki/entitlement/5562356167126193581-key.pem
sslclientcert = /etc/pki/entitlement/5562356167126193581.pem
metadata_expire = 1
enabled_metadata = 1
```

After:
```bash
[Default_Organization_Custom_Zoo]
name = Zoo
baseurl = https://devel.croberts.io/pulp/content/Default_Organization/Library/Partha/custom/Custom/Zoo
enabled = 1
gpgcheck = 0
sslverify = 1
sslcacert = /etc/rhsm/ca/katello-server-ca.pem
sslclientkey = /etc/pki/entitlement/5562356167126193581-key.pem
sslclientcert = /etc/pki/entitlement/5562356167126193581.pem
metadata_expire = 1000
enabled_metadata = 1
```

Used a host connected to a content view

PR #10547, opened 2023-05-04
Fixes #36270 - Update UI to reflect needs_publish on CV

sjha4:
Update CV UI to reflect needs_publish flag and add icons to versions and notice to publish wizard.
We have many pages where we avoid reloading the store from API for CV to avoid a refresh of the page. On such components, added a local store variable to reflect needs_publish based on UI actions.
Caveat: Repository level updates like content changes will need a refresh as there's no way to track that on the UI alone.
Play around with CV versions and check if the needs_publish icon shows up against the latest version. Also test the publish wizard to see if notice explaining that a publish is not needed shows up when needs_publish is false.

### DIFF
--- a/webpack/scenes/ContentViews/Publish/CVPublishForm.js
+++ b/webpack/scenes/ContentViews/Publish/CVPublishForm.js
@@ -1,4 +1,5 @@
 import React, { useState } from 'react';
+import { useSelector } from 'react-redux';
 import PropTypes from 'prop-types';
 import { translate as __ } from 'foremanReact/common/I18n';
 import {
@@ -10,6 +11,7 @@ import EnvironmentPaths from '../components/EnvironmentPaths/EnvironmentPaths';
 import ComponentEnvironments from '../Details/ComponentContentViews/ComponentEnvironments';
 import './cvPublishForm.scss';
 import WizardHeader from '../components/WizardHeader';
+import { selectCVNeedsPublish } from '../Details/ContentViewDetailSelectors';
 
 const CVPublishForm = ({
   description,
@@ -25,6 +27,7 @@ const CVPublishForm = ({
 }) => {
   const [alertDismissed, setAlertDismissed] = useState(false);
   const [needsPublishAlertDismissed, setNeedsPublishAlertDismissed] = useState(false);
+  const needsPublishLocal = useSelector(state => selectCVNeedsPublish(state));
 
   const checkPromote = (checked) => {
     if (!checked) {
@@ -38,7 +41,7 @@ const CVPublishForm = ({
         title={__('Publish')}
         description={
           <>
-            {!needsPublishAlertDismissed && !needsPublish && (
+            {!needsPublishAlertDismissed && !(needsPublish || needsPublishLocal) && (
             <Alert
               ouiaId="needs-publish-alert"
               variant="info"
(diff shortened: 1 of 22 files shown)


ianballou:
I'm not seeing `needs_publish` propagate to the UI until I fully refresh the page. I tried updating a filter and then immediately publishing.

ianballou:
One thought -- why is the "update available" icon on the version rather than somewhere at the top of the CV? If it will only ever be on the latest content view version, I wonder if the upgrade icon should be by the CV name instead.

sjha4:
> 

That makes sense..Let me run that by Maria and see if we can update that. 👍🏼

ianballou:
> I'm not seeing `needs_publish` propagate to the UI until I fully refresh the page. I tried updating a filter and then immediately publishing.

I think this may only be happening when deleting filters or filter rules rather than creating them.

sjha4:
> I'm not seeing `needs_publish` propagate to the UI until I fully refresh the page. I tried updating a filter and then immediately publishing.



> 

Ya..We have some pages on the UI where for responsiveness we do not refresh CV details upon some operations. I guess it's time to change that and fetch CV details on those operations. Let me check where things will need updating. 👍🏼

sjha4:
@ianballou : This is ready for re-review with the UI updates on needs_publish changes without refresh.

sjha4:
Updated to clear needs_publish on UI on publish. 👍🏼

webpack/scenes/ContentViews/Publish/CVPublishForm.js
```diff
@@ -36,11 +40,31 @@ const CVPublishForm = ({
...
+                __('No available component content view updates') :
+                __('No available repository or filter updates')}
+              actionClose={
+                <AlertActionCloseButton
+                  onClose={() => setNeedsPublishAlertDismissed(true)}
+                />
+            }
+              style={{ marginBottom: '24px' }}
+            >
+              <TextContent>{__('Newly published version will be the same as previous version.')}</TextContent>
```
ianballou, line 59:
```suggestion
              <TextContent>{__('Newly published version will be the same as the previous version.')}</TextContent>
```

webpack/scenes/ContentViews/components/NeedsPublishIcon.js
```diff
@@ -0,0 +1,27 @@
...
+import { Tooltip } from '@patternfly/react-core';
+import { ArrowCircleUpIcon } from '@patternfly/react-icons';
+
+const NeedsPublishIcon = ({ composite }) => (
+  <Tooltip
+    position="auto"
+    enableFlip
+    entryDelay={400}
+    content={composite ? __('Available content view version updates.') :
+      __('Available repositories and/or filters updates.')}
```
ianballou, line 13:
```suggestion
      __('Publish update available: repositories and/or filters have changed.')}
```
what do you think about something like this instead?

ianballou (commented):
some small comments, works well otherwise!

webpack/scenes/ContentViews/components/NeedsPublishIcon.js
sjha4: Updated. 👍🏼

ianballou (approved):
This is working for me now!

sjha4:
Merging! 🎉

PR #10548, opened 2023-05-05
Fixes #36250 - Alert user about past and current filters in the UI

ianballou:

Adds:
- A flag to each CV version that had filters applied
- A warning to the CV publish modal that tells users if filters are applied
- A link that opens up the CV filters page in a new link

This was built on top of @sjha4 's  needs_publish update.

1) Make filters and check that the CV review modal shows the warning correctly.
2) Remove the filters and check that the warning goes away.
3) Publish some CVVs with filters and ensure the filter icon appears.
4) Check that the filters link works.
5) Check that the "needs publish" icon works well with the filters applied icon.


sjha4:
Merge conflicts 😺

webpack/scenes/ContentViews/Publish/CVPublishReview.js
```diff
@@ -61,6 +74,10 @@ const CVPublishReview = ({
             <Td>
               <ComponentEnvironments environments={promotedToEnvironments} />
             </Td>
+            {filtered
+              ? <Td><Link to={`/content_views/${id}#/filters`} target="_blank" rel="noopener noreferrer">Filters </Link> </Td>
```
sjha4, line 78:
```suggestion
              ? <Td><Link to={`/content_views/${id}#/filters`} target="_blank" rel="noopener noreferrer">{__('View filters')} </Link> </Td>
```

sjha4:
One thing @parthaa mentioned was adding the filterIcon to version details screen as well..

ianballou:
> Would you be up for adding a param to the CV version API and returning the value in applied_filters when true?

@sjha4 what do you think about just always returning it? It's not a complicated query or anything on the backend to generate the `applied_filters` result.

sjha4:
> 

It can be a very big value in the returned object and make the return bulky unless someone asks for it? I am thinking one filter will all errata included with bulk select for example..Not that users should be creating such huge filters.. 😄

ianballou:
> It can be a very big value in the returned object and make the return bulky unless someone asks for it? I am thinking one filter with all errata included with bulk select for example..Not that users should be creating such huge filters.. smile

Ah applied filters is just a boolean! So no bulky data there.

ianballou:
@sjha4 I just added:
- The new API return with applied filters for content view versions
- Filters applied icon to content view versions header

app/views/katello/api/v2/content_view_versions/base.json.rabl
```diff
@@ -89,3 +89,9 @@ end
 node :filters_applied do |cv_version|
   cv_version.filters_applied?
 end
+
+if params.key?(:include_applied_filters)
```
sjha4, line 93:
```suggestion
if ::Foreman::Cast.to_bool(params.fetch(:include_applied_filters, false))
```

sjha4:
Looking good..Also have a few unaddressed suggestions from prior review, one of which should fix the react tests.

ianballou:
> Looking good..Also have a few unaddressed suggestions from prior review, one of which should fix the react tests.

Which suggestions? I looked em over and it seems I hit  them all.

sjha4:
> > Looking good..Also have a few unaddressed suggestions from prior review, one of which should fix the react tests.
> 
> Which suggestions? I looked em over and it seems I hit them all.

Oh..I did not see the updates and the test failures properly..We need to update some mocks for tests..

ianballou:
I updated all of the CV and CVV filters to have filters, lemme know if it would be better to have some "unfiltered" ones.

sjha4:
Could you update the webpack/scenes/ContentViews/components/NeedsPublishIcon.js to have 
`<ArrowCircleUpIcon size="sm" style={{ color: 'var(--pf-global--primary-color--100)', marginLeft: '9px' }} />`

That makes gap between the needs_publish icon and filter_applied icon as 9 px instead of 18px. Note spacing between icons with/without the change.
Screenshot from 2023-05-10 12-56-39
Screenshot from 2023-05-10 12-55-57

ianballou:
@sjha4 I added in your style suggestion!

sjha4:
Rubocop failure seems related. Looks good otherwise..Will ack once rubocop is happy. 🎉

sjha4:
[test]

sjha4 (approved):

sjha4:
@ianballou Testing closing-reopening.. 🥲

sjha4:
[test katello]

PR #10551, opened 2023-05-08
Fixes #35722 - Selector to enable/disable custom/rh repos

parthaa:
Adds a Repository Type selector to AK repo-sets page. This will enable one select/deselect all  custom/rh repos and add override. 

Initially thought of using scoped search. Sounds like the backend api already has code to handle this so no scoped search.

- Enable some RH Repos
- Create some custom repos
- Create Activation Key pointing to Library/Default CV
- Go to Activation Key repository sets page
- You should see a drop down after the "limit to environment" checkbox.
- Select Redhat/Custom/blank  and make sure the items are filtered correctly.


chris1984 (approved):
ACK, works

Custom:
custom


RH:
rh

chris1984:
Looks better now with the divider
ak

@jeremylenz how does that look?

jeremylenz (approved):
APJ :+1:

PR #10552, opened 2023-05-10
Fixes #36378 - Use plugin dsl for gettext

adamruzicka:

Switch to a new plugin dsl for registering translations

Extracting strings should work just the same


ianballou (approved):
Tested gettext and po_to_json for katello, they both work!

PR #10553, opened 2023-05-10
Fixes #36338 - display Module streams tab correctly

lfu:
Change the regex to parse OS name and version correctly.
Display Module streams tab for OS when regex does not match it.
 
Verify module stream tab is displayed for hosts that supports module streams.


chris1984 (approved):
ACK, works well

Rocky 8:
rocky8

Rocky 9:
rocky9

PR #10554, opened 2023-05-11
Fixes #36390 - Orphan cleanup on capsules

sjha4:
Run pulp orphan cleanup after removing other records like unneeded repositories etc which create orphaned content of their own.
Run capsule orphan cleanup with orphan_protection_time as 0.
Set up a capsule assigned to Default organization that also has some repo content (Synced with immediate download policy)
Sync.
Notice files in /var/lib/pulp/media/artifact/*/*
Remove Default Organization from the capsule and run `bundle exec rails katello:delete_orphaned_content`
Notice files in /var/lib/pulp/media/artifact/*/* . It should be empty since all the downloaded content is orphaned.

### DIFF
--- a/lib/katello/plugin.rb
+++ b/lib/katello/plugin.rb
@@ -669,7 +669,7 @@ Foreman::Plugin.register :katello do
         type: :integer,
         default: 1440,
         full_name: N_('Orphaned Content Protection Time'),
-        description: N_('Time in minutes to consider orphan content as orphaned.')
+        description: N_('Time in minutes before content that is not contained within a repository and has not been accessed is considered orphaned.')
 
       setting 'remote_execution_prefer_registered_through_proxy',
         type: :boolean,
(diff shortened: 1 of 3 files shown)


sjha4:
[test katello]

ianballou (commented):
Should we address this comment too? https://bugzilla.redhat.com/show_bug.cgi?id=2043089#c5

"I would also update the description of the parameter from " Time in minutes to consider orphan content as orphaned." to "Time in minutes before content not contained within a repo nor recently touched will be considered as orphaned"

The existing description doesn't really capture what is going on."

lib/katello/plugin.rb
```diff
@@ -669,7 +669,7 @@ def katello_template_setting_values(name)
         type: :integer,
         default: 1440,
         full_name: N_('Orphaned Content Protection Time'),
-        description: N_('Time in minutes to consider orphan content as orphaned.')
+        description: N_('Time in minutes before content not contained within a repo nor recently touched will be considered as orphaned.')
```
ianballou, line 672: How about "Time in minutes before content that is not contained within a repository and has not been accessed is considered orphaned."

ianballou (commented):
Working well! Just one note:

ianballou (approved):

PR #10556, opened 2023-05-15
Fixes #36397 - Disabled needs_publish flag for old CVs and CVs with audits cleaned up

sjha4:
Show a disabled needs_publish icon when we can not determine if a CV needs publish or not.
This is the case when:
1. CV is a component CV.
2. CV audits have been cleaned since the last publish.
3. CV version was published pre-upgrade when logic for creating several of the audits were added.
For a CV version, add a few repositories after publish.
This would mean needs_publish should be true.
Notice active(primary-color) icon next to version on table/version details.
In console> Audit.delete_all
Now the CV versions page should show disabled needs_publish icon with a tooltip text.
The publish wizard shouldn't tell user there are no updates in case of nil (soft-yes) needs_publish

### DIFF
--- a/app/models/katello/content_view.rb
+++ b/app/models/katello/content_view.rb
@@ -787,7 +787,28 @@ module Katello
       published_component_version_ids.compact.uniq.sort != unpublished_component_version_ids.compact.uniq.sort
     end
 
+    def last_publish_task_success?
+      last_publish_result = latest_version_object&.history&.publish&.first&.task&.result
+      return last_publish_result.present? && last_publish_result == 'success'
+    end
+
     def needs_publish?
+      #Returns
+      # True:
+      #     a) When content/repo/filter change audit records exist
+      #     b) CV hasn't ever been published
+      # nil:
+      #     a) When CV version creation audit is missing(Indicating audit cleanup)
+      #     b) Version doesn't have audited_filters set indicating
+      #     it was published before 4.9 upgrade when we started auditing changes on the CV.
+      #     c) Last publish task failed leaving us with no way of knowing if all content in the version is correct.
+      # False:
+      #     a) No changes were detected via audits *and*
+      #        Audit for CV publish exists (Audits haven't been cleaned up)
+      #        *and* applied_filters field is set(Published after upgrade)
+      #
+      return true unless latest_version_object
+      return nil unless last_publish_task_success?
       return composite_cv_components_changed? if composite?
       # return true if the audit records clearly show we have unpublished changes
       return true if (audited_cv_repositories_since_last_publish.present? || audited_cv_repository_changed.present? ||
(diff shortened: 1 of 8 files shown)


wbclark:
1. created test_cv, added a repo, published version 1.0

2. added an additional repo, observed the blue `needs_publish` icon on the versions page for the cv:

Screenshot from 2023-05-19 07-31-07

3.

```
[7] pry(main)> Katello::ContentView.find(2).latest_version_object.audits.delete_all
=> 3
[8] pry(main)> Katello::ContentView.find(2).latest_version_object.audits.exists?
=> false
```

4. refresh the CV versions page, however, at this point I didn't see any change in the status

5. it appears something is not correct with my implementation of #10560 , since `needs_publish?` still returns true

```
[9] pry(main)> Katello::ContentView.find(2).latest_version_object.audits
=> []
[10] pry(main)> Katello::ContentView.find(2).latest_version_object.applied_filters
=> {"applied_filters"=>[], "dependency_solving"=>false}
[11] pry(main)> Katello::ContentView.find(2).needs_publish?
=> true
```

wbclark:
@sjha4 ++ ++ ++ !!

it turns out this is proper behavior, since adding a repository created audits on the cv, and we can return true based on those

```
[3] pry(main)> Katello::ContentView.find(2).audits.count
=> 4
[4] pry(main)> Katello::ContentView.find(2).audits&.pluck(:action)
=> ["create", "update", "update", "update"]
[5] pry(main)> Katello::ContentView.find(2).audits.delete_all
=> 4
[6] pry(main)> Katello::ContentView.find(2).needs_publish?
=> nil
```

Screenshot from 2023-05-19 08-05-50

app/models/katello/content_view.rb
```diff
@@ -787,10 +787,29 @@ def composite_cv_components_changed?
       published_component_version_ids.compact.uniq.sort != unpublished_component_version_ids.compact.uniq.sort
     end
 
+    def last_publish_task_success?
+      last_publish_result = latest_version_object&.history&.publish&.first&.task&.result
+      return last_publish_result&.equal?('success')
+    end
+
     def needs_publish?
```
parthaa, line 795:
```suggestion
    def needs_publish?
          # returns 
          # true - if audit records indicate changes since latest version
          # nil - if we have insufficient data 
         # false - otherwise
```

parthaa: format it correctly :) but theres  value in stating what the expected behavior is here

app/models/katello/content_view.rb
```diff
@@ -787,10 +787,29 @@ def composite_cv_components_changed?
...
-        audited_cv_filters_changed.present? || audited_cv_filter_rules_changed.present?
+      # return true if the audit records clearly show we have unpublished changes
+      return true if (audited_cv_repositories_since_last_publish.present? || audited_cv_repository_changed.present? ||
+        audited_cv_filters_changed.present? || audited_cv_filter_rules_changed.present?)
+      # if we didn't return `true` already, either the audit records show that we don't need to publish, or we may
+      # have insufficient data to make the determination (either audits were cleaned, or never got created at all).
+      # first, check for the `create` audit record; its absence indicates that audits were cleaned some time after
+      # the cv version was created (i.e. the first indeterminate state) so we return `nil` in that case.
+      return nil unless latest_version_object&.audits&.where(action: "create")&.exists?
+      # even when the `create` audit exists, the other audits could still be absent due to the latest cv version
```
parthaa, line 807:
```suggestion
      # even when the `create` audit exists, audits related to filters that made the latest version could have changed and we may not have records of 
```

parthaa:
Code LGTM. If some one has tested this I am ready to ack :)

app/models/katello/content_view.rb
parthaa: Updates look good

app/models/katello/content_view.rb
parthaa: Updates look good

parthaa (approved):
ACK

PR #10557, opened 2023-05-17
Fixes #36409 - Update webpack translations and edit mo-files command to ignore .gitignore

ianballou:

Adds the new webpack translation po files and updates the makefile to tell release engineers when to create the mo files. I'm going to update the tool_belt config for branching and releasing as well.

Try `make -C local tx-update` and `make -C local mo-files`


locale/Makefile
```diff
@@ -50,14 +50,16 @@ tx-pull: $(EDITFILES)
...
 	@echo
 
 mo-files: $(MOFILES)
-	git add $(POFILES) $(POTFILE) ../locale/*/LC_MESSAGES
-	git commit -m "i18n - pulling from tx"
+	git add -f $(POFILES) $(POTFILE) ../locale/*/LC_MESSAGES
+	git commit -m "i18n - creating .mo files"
 	@echo
-	@echo Changes commited!
+	@echo Changes commited! Do not push this commit to GitHub as it contains .mo files.
```
jeremylenz, line 62: Any reason we can't just add `.mo` to `.gitignore`?

ianballou: They are indeed ignored

ianballou:
I just made a change to stop us from committing the .mo files at all. That should be fine, we don't need to commit the files to git to create the gem.

locale/Makefile
jeremylenz: In that case, is this warning needed? I guess you'd have a blank commit then, okay :)

chris1984 (approved):
ACK it works, thanks @ianballou 

```bash
mkdir -p ta/LC_MESSAGES
msgfmt -o ta/LC_MESSAGES/katello.mo ta/katello.po
touch ta/LC_MESSAGES/katello.mo
mkdir -p te/LC_MESSAGES
msgfmt -o te/LC_MESSAGES/katello.mo te/katello.po
touch te/LC_MESSAGES/katello.mo
mkdir -p zh_CN/LC_MESSAGES
msgfmt -o zh_CN/LC_MESSAGES/katello.mo zh_CN/katello.po
touch zh_CN/LC_MESSAGES/katello.mo
mkdir -p zh_TW/LC_MESSAGES
msgfmt -o zh_TW/LC_MESSAGES/katello.mo zh_TW/katello.po
touch zh_TW/LC_MESSAGES/katello.mo

.mo files created. Do not push these up to GitHub.
```

locale/Makefile
```diff
@@ -50,14 +50,14 @@ tx-pull: $(EDITFILES)
...
 tx-update: tx-pull
 	@echo
-	@echo Run rake plugin:gettext[$(DOMAIN)] from the Foreman installation, then make -C locale mo-files to finish
+	@echo Run rake plugin:gettext[$(DOMAIN)] from the Foreman installation. \
+		If this is for a release, run make -C locale mo-files from the Katello directory to finish. \
+		.mo files should not be uploaded to GitHub, they should only be packed into the gem.
 	@echo
 
 mo-files: $(MOFILES)
-	git add $(POFILES) $(POTFILE) ../locale/*/LC_MESSAGES
```
evgeni, line 57: I think the PO files should still be committed? Otherwise people without access to TX can't recreate the gem (as git now would contain some old PO with old translations)?

ekohl: An alternative approach I've considered is a separate `build-mo-files` task, but failed to submit this as a PR earlier. https://github.com/theforeman/foreman_plugin_template/pull/44 has lived as local change on my disk for too long.

Then in packaging we would call `make -C locale build-mo-files` (and resolve https://projects.theforeman.org/issues/35789). I just haven't figured out how to do that exactly. We probably need to pull in some additional build time dependencies (like `gettext`).

It was my intention to play with this, but didn't find the time.

ianballou: I always `git add` the PO files myself, I'll double check that there's something about this in tool belt. We haven't used the mo-files command for ages because git would complain about the gitignore.

ianballou: I'm going to split out two commands, build-mo-files and then po-files

ianballou: https://github.com/Katello/katello/pull/10574

https://github.com/theforeman/tool_belt/pull/542

PR #10558, opened 2023-05-17
Fixes #36402 - Closing publish task wizard doesn't reset needs_publish local state

sjha4:
Updates page to hide needs_publish icon when CV is published and wizard is closed while task is running.
In case of errors, the icon will not appear and if you skip and resume tasks. This feels like a gap and I am thinking about opening an issue to track that.
1. Add/remove a filter on a cv. (Do not refresh anytime during the process)
2. Publish and close the wizard while task is running.
3. After the task polling completes, the needs_publish icon shouldn't appear on the latest version.

### DIFF
--- a/webpack/scenes/ContentViews/Publish/CVPublishFinish.js
+++ b/webpack/scenes/ContentViews/Publish/CVPublishFinish.js
@@ -68,6 +68,7 @@ const CVPublishFinish = ({
             setSaving(false);
           },
         ));
+        dispatch({ type: CONTENT_VIEW_NEEDS_PUBLISH_RESET });
       }, () => { setSaving(false); }));
     }
   }, [POLLING_TASK_KEY, currentStep, cvId, description, dispatch, forcePromote,


qcjames53 (commented):
Hi Samir! I just verified the PR works as described and the changed line looks fine to me :)

I noticed the needs_publish icon is removed before a new version is completely published. If that process fails for some reason, would we run into any issues?

sjha4:
> Hi Samir! I just verified the PR works as described and the changed line looks fine to me :)
> 
> I noticed the needs_publish icon is removed before a new version is completely published. If that process fails for some reason, would we run into any issues?

I have a PR out for that case:
https://github.com/Katello/katello/pull/10562

qcjames53:
> 

That sounds great. Ack'ed :+1:

qcjames53 (approved):

chris1984 (approved):
🚢 it

PR #10559, opened 2023-05-17
Fixes #36411 - add ability to skip treeinfo files

ianballou:
- Adds a new checkbox to skip treeinfo files
- Rewrites a treeinfo Pulp error to give better workaround instructions

1) Sync https://www.keepersecurity.com/kcm/2/el/8/x86_64
2) Verify the error in the Dynflow action
3) Enable treeinfo skipping
4) Sync again
5) Verify that the sync is successful

### DIFF
--- a/app/lib/actions/pulp3/abstract_async_task.rb
+++ b/app/lib/actions/pulp3/abstract_async_task.rb
@@ -114,6 +114,8 @@ module Actions
         case message
         when 'This repository uses features which are incompatible with \'mirror\' sync. Please sync without mirroring enabled.'
           'The "Complete Mirroring" mirroring policy is not compatible with this repository.  You may want to update it to use "Content Only"'
+        when 'Treeinfo file should have INI format'
+          'This repository has a malformed or inaccessible treeinfo file. To sync, try enabling \'Ignore treeinfo\'. All kickstart contents will be skipped.'
         else
           message
         end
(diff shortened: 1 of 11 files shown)


sjha4:
Test failure looks related.

sjha4:
Do we need to call pulp at all when updating ignored_types since pulp doesn't store it anywhere. We should probably remove it from here to avoid the pulp actions: https://github.com/Katello/katello/blob/3efcc7cf5d2aa7f007c7a9f81b34dcd941cce096/app/models/katello/root_repository.rb#L410

sjha4 (approved):
Acking first commit 👍🏼 
Without ignore_treeinfo checked:
Screenshot from 2023-05-23 13-22-30

With ignore treeinfo checked:
Screenshot from 2023-05-23 13-23-50

wbclark (commented):
is the change to `ensure_valid_ignorable_content` related to https://projects.theforeman.org/issues/36428 ?

right now it's attached to the commit with tests

sjha4:
> is the change to ensure_valid_ignorable_content related to https://projects.theforeman.org/issues/36428 ?

No..So that was a small change to make validation error more useful. If you set any of the 2 ignore flags with mirror complete download policy, the error should only report the Ignore flag that was checked.

sjha4:
That change also fixes 2 tests that were failing on the 1st commit which brought this to my notice:
https://ci.theforeman.org/blue/organizations/jenkins/katello-pr-test/detail/katello-pr-test/12784/tests

wbclark (approved):
ACK to the 2nd commit

sjha4:
Thanks @wclark and @ianballou ! Merging..

PR #10560, opened 2023-05-18
Fixes #36320 - Return nil when needs_publish? can't be determined by …

wbclark:
…audits

Can be tested in conjunction with https://github.com/Katello/katello/pull/10556


sjha4 (approved):
Looks good to me. 👍🏼 
Merge once @parthaa has had a chance to proof read the logic.

wbclark:
closing, since the commit was merged with https://github.com/Katello/katello/pull/10556

PR #10561, opened 2023-05-18
i18n - extracting new, pulling from tx

lfu:

- Update config for the transifex-cli
- Update *.po files


chris1984 (approved):
Nice work @lfu ACK

ianballou:
This doesn't fully solve the issue -- these are the steps that we take every branching. The issue is that the mo files were not being included at gem build time. Also the React javascript translation files aren't included. It's fine since my other PR included them, but the Katello translations problems aren't fully resolved yet.

I'll make some more PRs to help ensure that the mo files get included.

PR #10562, opened 2023-05-18
Fixes #36410 - Set needs_publish to true for versions published with a failed task

sjha4:
When a publish task ends in a failed state, we can not determine if all repository contents were successfully snapshotted in the CV.
In such a case, we should fall back on the indeterminate needs_publish value of null instead of determinate true/false.
1. Create and publish a CV with a modular repo and some modules selected in an include filter
2. In console, update one of the module streams in include filter href to junk value
3. Publish the CV again.
4. The publish will fail. In dynflow console, skip the failed step and resume the task. Task will end in warning state.
5. Check if needs_publish is returned as null for the CV in the API.

### DIFF
--- a/app/models/katello/content_view.rb
+++ b/app/models/katello/content_view.rb
@@ -787,7 +787,14 @@ module Katello
       published_component_version_ids.compact.uniq.sort != unpublished_component_version_ids.compact.uniq.sort
     end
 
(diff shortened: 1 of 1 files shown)


sjha4:
Closing as the commit went into another PR above.

PR #10563, opened 2023-05-18
i18n - extracting new, pulling from tx AngularJS

chris1984:
* Added missing `translate` to 15 AngularJS html pages
* Updated the `.pot` and `translations.js` Bastion AngularJs files with `grunt i18n:compile` and `grunt i18n:extract`


lfu (approved):
LGTM 👍🏻

PR #10564, opened 2023-05-18
Fixes #36417 - Bring back 'Republish repository metadata'

jeremylenz:

1. Add a 'Republish Repository Metadata' action to the repository details actions:
image
2. Add a 'Republish Repository Metadata' action to the Content view version details actions:
image
 
As we know, this action is dangerous for repositories using the 'Complete' mirroring policy, and the API requires the `force` option to be passed. With this in mind,

1. Any repositories that use the complete mirroring policy will have the action disabled.
2. For content view versions, the action will be disabled if _any_ of the content view version's repositories use complete mirroring.

You can still republish repo metadata in such cases; however, you'll have to use API or Hammer and not the web UI.

One caveat (described in more detail in https://bugzilla.redhat.com/show_bug.cgi?id=2184278#c8) is that this won't take into account whether a repository's mirroring policy has changed. We just have to live with that for now.

1. Verify that republishing repository metadata works as expected on the Repository detail page
2. Verify that republishing repository metadata works as expected on the Content view version details page
4. Verify that the option is disabled in both places when complete mirroring is set


engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/repository-details.controller.js
```diff
@@ -77,7 +77,7 @@
         });
 
         $scope.hideSyncButton = function(repository, advancedSync) {
-            var result = $scope.syncInProgress(repository.last_sync) || !repository.url || $scope.denied('sync_products', $scope.product);
+            var result = $scope.syncInProgress(repository.last_sync) || !repository.url || repository.mirroring_policy === 'mirror_complete' || $scope.denied('sync_products', $scope.product);
```
sjha4, line 80: Need to test it but this looks wrong maybe due to the name of the function hideSyncButton..Reading this code makes you think we hide sync button for mirror_complete repositories. Does it do that?

sjha4: If you just add your bitcoin wallet as the disabled reason for sync, it would level up from a bug to ransomware.. :stuck_out_tongue_closed_eyes:

jeremylenz: I think you may be right.. Let me look at that again..

jeremylenz:
@sjha4 Updated to fix the link disablement on the repository details page.

sjha4 (approved):
Ack..Tested various workflows ex:
1. Repo with complete_mirroring (Republish disabled) 👍🏼 
2. Repo without complete_mirroring (Republish enabled) 👍🏼 
3. Component CV version with child repo with complete_mirroring. (Republish disabled) 👍🏼 
4. Component CV with no child repo with complete_mirroring (Republish enabled) 👍🏼 
5. Composite CV with Component CV version with child repo with complete_mirroring. (Republish disabled) 👍🏼
6. Composite CV with Component CV with no child repo with complete_mirroring (Republish enabled) 👍🏼 

Actions get correctly planned and upon execution, publication is updated as needed on repositories.
Ack ✔️

PR #10565, opened 2023-05-19
Fixes #36420 - Correct aria-label for description

sjha4:
Correct aria-label of description field value on ACS details.

1. Go to ACS details page.
2. Inspect description value in the details.
3. Check if aria-label changed to description_text_value


chris1984 (approved):
ACK, I can see the change in the screenshot below
ouia

PR #10566, opened 2023-05-22
add packit config

evgeni:

adds support for packit (see https://community.theforeman.org/t/packit-for-foreman-get-production-rpms-from-prs/32412)

none

spin up a box
```
# dnf copr enable packit/katello-katello-<PR_NUMBER> centos-stream-8-x86_64
# dnf upgrade rubygem-katello
```
should give you a .rpm build of Katello from a PR :)

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -25,7 +25,7 @@ actions:
     - bash -c "ls -1t ./katello-*.gem | head -n 1"
   fix-spec-file:
     - 'bash -c "sed -i \"s/Source0:.*/Source0: ${PACKIT_PROJECT_ARCHIVE}/\" rubygem-katello.spec"'
-    - bash -c 'sed -i "/global release/a%global nightly $PACKIT_RPMSPEC_RELEASE" rubygem-katello.spec'
+    - bash -c 'sed -i "/global release/a%global nightly ${PACKIT_RPMSPEC_RELEASE:1}" rubygem-katello.spec'
 
 jobs:
   - job: copr_build
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,0 +1,43 @@
+# See the documentation for more information:
+# https://packit.dev/docs/configuration/
+
+specfile_path: rubygem-katello.spec
+
+# add or remove files that should be synced
+files_to_sync:
+    - rubygem-katello.spec
+    - .packit.yaml
+
+# name in upstream package repository or registry (e.g. in PyPI)
+upstream_package_name: katello
+# downstream (Fedora) RPM package name
+downstream_package_name: rubygem-katello
+
+upstream_tag_template: "{version}"
+
+actions:
+  post-upstream-clone:
+    - "wget https://raw.githubusercontent.com/theforeman/foreman-packaging/rpm/develop/packages/katello/rubygem-katello/rubygem-katello.spec -O rubygem-katello.spec"
+    - "sed -i '/%global prerelease/d' rubygem-katello.spec"
+  get-current-version:
+    - "sed -i 's/-master//' lib/katello/version.rb"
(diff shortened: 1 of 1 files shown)


evgeni:
seems someone (@ehelms?) gotta approve the org in Packit first -.-

evgeni:
That built, but uses a version that will "always" higher than nightly, not sure I like it

evgeni:
> That built, but uses a version that will "always" higher than nightly, not sure I like it

stupid pre-release gems, I couldn't fix it without doing almost *everything* packit does manually, so, uh, let's not :)

ehelms:
> > That built, but uses a version that will "always" higher than nightly, not sure I like it
> 
> stupid pre-release gems, I couldn't fix it without doing almost _everything_ packit does manually, so, uh, let's not :)

What is it that packit is doing that we don't have control over? As we onboard more plugins as nightly projects, we are going to see this over and over. So I am left wondering if we need a fix from Packit?

evgeni:
It's replacing the `Version` and `Release` entries, thus breaking our macro magic.

I found a way to make it stop, but then it's also not doing a few other changes to the spec file and that broke other stuff. I dan try having another look next week

.packit.yaml
```diff
@@ -0,0 +1,43 @@
...
+files_to_sync:
+    - rubygem-katello.spec
+    - .packit.yaml
+
+# name in upstream package repository or registry (e.g. in PyPI)
+upstream_package_name: katello
+# downstream (Fedora) RPM package name
+downstream_package_name: rubygem-katello
+
+upstream_tag_template: "{version}"
```
ofedoren, line 16: Since I'm also trying to add packit to some project, I've got a question: is this line necessary? I mean, by https://packit.dev/docs/configuration/#upstream_tag_template it makes more sense if the tag contains something else along the version.

evgeni: It is not. It's just part of the template (as most projects/plugins use v{version}) and it felt more natural to set it "correctly" vs dropping

lachmanfrantisek:
> What is it that packit is doing that we don't have control over? As we onboard more plugins as nightly projects, we are going to see this over and over. So I am left wondering if we need a fix from Packit?


Hello all, Packit PO here..;)

I am happy to see you are configuring Packit for more and more Foreman projects! So, let us know if there is something we can improve to make your life easier. It makes sense to make a change on our side if there is something you are fighting in every single package.

Please, either [create an issue](https://github.com/packit/packit/issues/new/choose) or we can also set up a virtual meeting to discuss this. (If you want, we can also discuss how you can get more out of Packit...)

František

evgeni:
> > > That built, but uses a version that will "always" higher than nightly, not sure I like it
> > 
> > 
> > stupid pre-release gems, I couldn't fix it without doing almost _everything_ packit does manually, so, uh, let's not :)
> 
> What is it that packit is doing that we don't have control over? As we onboard more plugins as nightly projects, we are going to see this over and over. So I am left wondering if we need a fix from Packit?

I think it now does "the right thing" when mangling around the spec file, and I've opened https://github.com/packit/packit/issues/2038 and https://github.com/packit/packit/issues/2040 to make those steps easier in the future.

ehelms:
Where does the buildroot get specified? When we switch to Copr we will have repositories that were built against RHEL and I'd imagine we'd want to do the same across Packit?

ehelms (approved):
Did not mean to close this earlier, I have a later question I posted this as it stands works and is useful so I am going to merge.

evgeni:
The `targets` is what effectively maps to buildroots/chroots. 
If I read https://packit.dev/docs/configuration/upstream/copr_build#available-copr-build-targets correctly, as long as Copr offers rhel, we can configure it.

PR #10568, opened 2023-05-23
Fixes #36425 - request backend status with a trailing slash

evgeni:

Candlepin doesn't mind either way, but Pulp really wants that slash, which results in a redirect if we request the non-slashed URL.
So this PR changes the status URL to end with a `/`

None.

Ensure `Katello::Ping` still works with Candlepin and Pulp, e.g. by calling `hammer ping`


ekohl (approved):
I noticed this too and mentioned it to @ianballou (I think on IRC). Simple fix, but does indeed save resources.

evgeni:
another optimization could be to reduce the number of requests, right now there are two (one for `pulp3`, one for `pulp3_content`, but both use the same URL, just parse different things out of it)

jeremylenz:
[test katello]

evgeni:
heh, I guess there is some more mocking/vcring to do :>

jeremylenz (approved):
LGTM! :+1: 

Thanks @evgeni

PR #10571, opened 2023-05-24
Fixes #36435 - Make metadata_expire an optional field

sjha4:
1. Go to create new repository
2. Select type Yum and fill in fields other than metadata_expire.
3. You should be able to save the repository.

On the details view for the repository
1. Metadata expire field is now under Publishing Settings section and also lowest value you can set is 1 similar to on the new repository page.


chris1984 (approved):
Looks good :ship: it

PR #10572, opened 2023-05-25
Fixes #36436 - Don't sync generated repos to capsules

sjha4:
Opening this up for reviews.
I am not so sure if we need to ignore all generated CV repos. That's what I have here but if we want to narrow it down to all exported repos only, I can make that change.
In hammer do a repo export, library export, import and other operations.
These generated repos should not get synced to the capsule tied to the environment or library.


app/models/katello/content_view.rb
```diff
@@ -95,11 +95,13 @@ class ContentView < Katello::Model
                                :repository_import,
                                :repository_export_syncable])
     }
-    scope :ignore_generated, -> {
-      where.not(:generated_for => [:repository_export,
-                                   :repository_import,
-                                   :library_export_syncable,
-                                   :repository_export_syncable])
+    scope :ignore_generated, ->(include_library_generated = false) {
```
ianballou, line 98:
```suggestion
    scope :ignore_generated, ->(ignore_library_generated = false) {
```

include_library_generated made me think that it wasn't ignoring library generated repos.

ianballou (commented):
This is working fine for me! I'm not seeing any of the generated repositories get synced to the capsule.

app/models/katello/content_view.rb
sjha4: So when ignore_library_generated is true we return library_export and library_import as part of the ignore_generated list and vice-versa. 
With include_library_generated , I was thinking of it as including library_generated in the ignore_generated list.

ianballou: Alrighty, I could see it either way

chris1984 (commented):
I think having a setting to toggle the importable content view would be best and solve the case of preventing the user from synchronizing it to the capsule and for a scenario like a customer can't physically get to the capsule to import the files, edge cases, then they can enable the setting and have it sync.

ianballou:
> a scenario like a customer can't physically get to the capsule to import the files, edge cases, then they can enable the setting and have it sync.

With how import/export is, can a customer do anything with the generated repos synced to a capsule? We'd need to make sure that use case is documented so users don't get more confused.

Also, from the bug description on the Redmine, it sounds like the certs can get messed up.

app/models/katello/content_view.rb
parthaa, line 98: Shouldn't this be
```suggestion
    scope :ignore_generated, ->(include_library_generated:false) {
```

app/models/katello/content_view.rb
```diff
@@ -95,11 +95,13 @@ class ContentView < Katello::Model
                                :repository_import,
                                :repository_export_syncable])
     }
-    scope :ignore_generated, -> {
-      where.not(:generated_for => [:repository_export,
-                                   :repository_import,
-                                   :library_export_syncable,
-                                   :repository_export_syncable])
+    scope :ignore_generated, ->(include_library_generated: false) {
```
parthaa, line 98: Can you add a simple model unit test for this scope (just ensure the params are getting set correctly)

sjha4:
Pushed the update. 
To recap:
All generated CVs will be excluded from Capsule syncs.
Imported content should be added to a regular content view and follow regular lifecycle paths of publish/promote to make it to a capsule.

parthaa:
ACK

PR #10573, opened 2023-05-25
Fixes #36442 - Capsule is auto-synced on CV promotion

sjha4:
1. Check the value for setting: Sync Smart Proxies after content view promotion
2. Promote a CV to an env that has a capsule attached.
3. The capsule sync task should get triggered after promotion based on the value of that setting.


ianballou (approved):
Looks great!

PR #10575, opened 2023-05-26
Fixes #36445 - remove extra th element from repository sets page

trev-allison03:
Removing the extra th element from the repository sets page
Steps to Reproduce:
1. Register host
2. Navigate to All hosts->$hostname->Content->Repository sets
3. Check DOM for the table, the number of header columns should be the same as the number of columns in rows


jeremylenz:
[test katello]

jeremylenz:
@trev-allison03 Looks like eslint failed. `npx eslint webpack/ --fix` should fix it :)

jeremylenz (approved):
The extra `<td>` element is removed as advertised.

APJ :+1:

jeremylenz:
[test katello]

PR #10579, opened 2023-05-30
Fixes #36453 - Prevent regenerating metadata for repositories that use complete mirroring

sjha4:
1. Deprecates --force flag on repository republish metadata and CV version metadata regenerate.
2. Errors out for republishing metadata on complete mirroring repositories.
3. Skips Complete Mirroring repositories of a CV version when republishing metadata of version repos
On the UI, we disable regenerate metadata button for 'Complete Mirroring' repos and CV versions with any 'Complete Mirroring' repositories.
On the API and hammer, we now fail on republishing metadata on 'Complete Mirroring' repos and 'Complete Mirroring' repositories in a CV version will be skipped.
1. Create a couple repos, one with Complete Mirroring and another with content only or additive.
2. In hammer, try hammer -r repository republish --id=`id` for both. 
3. Add both repos to a CV and publish a version.
4. In hammer, try hammer -r content-view version republish-repositories --id=`id` . Notice that the foreman task only republishes metadata on the version repos for the non-complete mirroring repo.


jeremylenz (commented):
A couple things I'm thinking about here..

1. This blocks access to the "bad regenerates" in the API controller for both CVV and Repository endpoints. But at the Dynflow action level, access is only blocked for the CVV action and the single Regenerate action remains unchanged. Should the single action also be updated, so that everything is consistent?
2. What happens in the [edge case](https://bugzilla.redhat.com/show_bug.cgi?id=2184278#c8) we discussed on the previous BZ, where a complete-mirroring repo has already gotten its local metadata regenerated incorrectly? Will the next repo sync fix this? Or is more needed here to address this case?

sjha4:
>> This blocks access to the "bad regenerates" in the API controller for both CVV and Repository endpoints. But at the Dynflow action level, access is only blocked for the CVV action and the single Regenerate action remains unchanged. Should the single action also be updated, so that everything is consistent?

We block complete mirroring repo metadata regeneration for library repositories. I don't think we need to worry about action layer if the API level blocks it.
For CVV, we do not block it at any level. The action just skips version repositories with complete mirroring and allows metadata generate on the version for other repositories.
We can think about not disabling the action on CV version on the UI for UI/Hammer consistency.

sjha4:
>> What happens in the [edge case](https://bugzilla.redhat.com/show_bug.cgi?id=2184278#c8) we discussed on the previous BZ, where a complete-mirroring repo is changed to non-complete and its local metadata is regenerated incorrectly? Is more needed here to address this case? cc @ianballou

We don't have a history of mirroring policy of a repo when published in a version or in library. I think it's fine to trust the library repo mirroring policy at the time of requested operation.

sjha4:
Test failures expected. Will push an update.

app/controllers/katello/api/v2/repositories_controller.rb
```diff
@@ -323,12 +323,12 @@ def repository_types
...
-    api :PUT, "/repositories/:id/republish", N_("Forces a republish of the specified repository, regenerating metadata and symlinks on the filesystem.")
+    api :PUT, "/repositories/:id/republish", N_("Forces a republish of the specified repository, regenerating metadata and symlinks on the filesystem. Not allowed for repositories with 'Complete Mirroring' mirroring policy.")
     param :id, :number, :desc => N_("Repository identifier"), :required => true
-    param :force, :bool, :desc => N_("Force metadata regeneration to proceed.  Dangerous when repositories use the 'Complete Mirroring' mirroring policy."), :required => true
+    param :force, :bool, :desc => N_("Force metadata regeneration to proceed. (Deprecated)"), deprecated: true
     def republish
-      unless ::Foreman::Cast.to_bool(params[:force])
-        fail HttpErrors::BadRequest, _('Metadata republishing must be forced because it is a dangerous operation.')
+      if @repository.mirroring_policy == Katello::RootRepository::MIRRORING_POLICY_COMPLETE
+        fail HttpErrors::BadRequest, _("Metadata republishing is not allowed on repositories with 'Complete Mirroring' mirroring policy.")
```
ianballou, line 331:
```suggestion
        fail HttpErrors::BadRequest, _("Metadata republishing is not allowed on repositories with the 'Complete Mirroring' mirroring policy.")
```

ianballou:
> and CV versions with any 'Complete Mirroring' repositories.

Can we allow the regenerating, but just skip the complete mirroring ones like we do in the API ?

ianballou (approved):
Looking great to me! Tested UI and hammer.

PR #10580, opened 2023-05-31
Fixes #36459 - Hide import_only and generated fields on CV

sjha4:
1. Set up tabs better. (On master, for generated CVs, there's a hidden tab which still shows up as clickable where filters tab would be)
2. Hide import_only and generated fields on CV details when false
3. Remove import_only from CV create form.
1. Create a CV from UI. You shouldn't see the import only flag
2. On the created CV, you shouldn't see import_only or generated fields.
3. Export library environment. Make sure you see the disabled generated flag.
4. Import the library environment to a staellite and ensure you see the disabled import_only flag set to true on the UI.


sjha4:
[katello]

jeremylenz (approved):
Approving based on @qcjames53, thanks!

parthaa (approved):
ACK

PR #10581, opened 2023-06-01
Fixes #36430 - Get arch restriction data from all repos, not just those with URL

jeremylenz:

image

On the Repository Sets tab, the arch restriction label ("x86_64" in the image above) was based on

```rb
pc.repositories&.first&.arch
```

(where `pc` is a `Katello::ProductContent`).

`pc.repositories` runs the query

```rb
def repositories
      Katello::Repository.in_default_view.where(:root_id => product.root_repositories.has_url.where(:content_id => content.cp_content_id))
end
```

Because of the `has_url`, `pc.repositories` will not return any repository that doesn't have a URL. So if you have a product with an empty repository, or a repository that doesn't yet have an Upstream URL set, `pc.repositories` would return `[]`, meaning the arch restriction label would not be properly displayed.

With this change, the query doesn't filter out any repositories, thus allowing a correct arch restriction label in the UI.

Originally the new query was using `pc.product.root_repositories.first` but I realized that would cause an incorrect arch restriction label if you had more than one repo in your product.

Create custom products with repos that don't have an upstream URL
Restrict the architecture on one of these repos, but leave the other unrestricted
Verify that the arch restriction label shows up correctly on all custom repos.

### DIFF
--- a/app/models/katello/product_content.rb
+++ b/app/models/katello/product_content.rb
@@ -57,17 +57,17 @@ module Katello
       Katello::Repository.in_default_view.where(:root_id => product.root_repositories.has_url.where(:content_id => content.cp_content_id))
     end
 
-    def root_repositories
+    def unfiltered_repositories
       # don't filter by url, as we want to show all repos in the product
       Katello::Repository.in_default_view.where(:root_id => product.root_repositories.where(:content_id => content.cp_content_id))
     end
 
     def arch
-      root_repositories.first&.arch
+      unfiltered_repositories.first&.arch
     end
 
     def os_versions
-      root_repositories.first&.os_versions || []
+      unfiltered_repositories.first&.os_versions || []
     end
 
     def enabled_value_from_candlepin


jeremylenz:
Updated to handle the case where a product has >1 repo.

ready for review

app/models/katello/product_content.rb
```diff
@@ -52,11 +52,24 @@ def self.with_valid_subscription(organization)
       where(:product_id => Katello::PoolProduct.where(:pool_id => organization.pools).select(:product_id))
     end
 
-    # used by Katello::Api::V2::RepositorySetsController#index
+    # following 4 methods used by Katello::Api::V2::RepositorySetsController#index
     def repositories
       Katello::Repository.in_default_view.where(:root_id => product.root_repositories.has_url.where(:content_id => content.cp_content_id))
     end
 
+    def root_repositories
```
lfu, line 60: Could we have a better name here since `root_repositories` has its own meaning in Katello already?
How about `all_repositories`, `source_repositories`, `base_repositories`?

app/models/katello/product_content.rb
```diff
@@ -52,11 +52,24 @@ def self.with_valid_subscription(organization)
       where(:product_id => Katello::PoolProduct.where(:pool_id => organization.pools).select(:product_id))
     end
 
-    # used by Katello::Api::V2::RepositorySetsController#index
+    # following 4 methods used by Katello::Api::V2::RepositorySetsController#index
     def repositories
       Katello::Repository.in_default_view.where(:root_id => product.root_repositories.has_url.where(:content_id => content.cp_content_id))
```
lfu, line 57: Is it possible to drop `has_url` here?

jeremylenz: `repositories` is used in the existing API response / rabl for repository sets, so I didn't want to change that output.

app/models/katello/product_content.rb
jeremylenz: They _are_ root repositories, though! Maybe `all_root_repositories` to highlight the fact that it's not filtered?

app/models/katello/product_content.rb
lfu: Why repos without url are displayed?

app/models/katello/product_content.rb
lfu: These are `Katello::Repository` instead of `Katello::RootRepository`.

app/models/katello/product_content.rb
jeremylenz: You can have a custom repository that doesn't sync to an upstream repo -- for example, if you upload all the rpms manually.

app/models/katello/product_content.rb
jeremylenz: My bad. `all_repositories` sounds good to me, then.

jeremylenz:
@lfu Updated.

lfu (approved):
Nice work!

PR #10582, opened 2023-06-01
Fixes #36461 - 'Remove orphans' task fails on DeleteOrphanAlternateContentSources step

sjha4:
Only triggers orphan ACS cleanup of types enabled on proxy. This is a special case where users diable file plugin on proxy.

1. The long way to test this is to disable file content on a proxy.
2. And then try running bundle exec rails katello:delete_orphaned_content
3. You'll see the remove orphan task for proxy fail.
4. On this PR, it should not fail.

A shortcut to test that I used, is to change L82 and 83 of katello/app/services/katello/pulp3/smart_proxy_mirror_repository.rb, replace 'file' with 'file2'
and notice it doesn't fail with this PR.

### DIFF
--- a/app/services/katello/pulp3/repository.rb
+++ b/app/services/katello/pulp3/repository.rb
@@ -39,6 +39,7 @@ module Katello
 
       def self.api(smart_proxy, repository_type_label)
         repo_type = RepositoryTypeManager.enabled_repository_types[repository_type_label]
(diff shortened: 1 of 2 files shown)


ianballou (approved):
Looks good to me, I disabled file type on the pulp primary and the mirror. After applying the PR (and remembering to refresh the proxies + restart the server), the orphan deletion worked without error.

PR #10584, opened 2023-06-04
Add an internal API for Katello Event Queue and Candlepin Events

jturel:

The news has spread like wildfire that Katello 4.10 will be the release which removes Katello Agent, leaving only two subsystems - Event Queue and Candlepin Events - using the EventDaemon thing that runs those services inside the Puma process. Now could be the right time to move away from the EventDaemon completely and do something a little more _normal_ in terms of how these services fit into the bigger picture.

- Remove the need for the Katello Event Daemon
- Use an established pattern (API) to allow event queue and candlepin events to be processed
- A new repository tentatively called `katello_events` will house the Ruby code that calls these APIs. The two services therein can be installed on the system using systemd ie `systemctl start/stop/restart/status katello-event-queue` and `systemctl start/stop/restart/status katello-candlepin-events`
- Does nothing to EventDaemon or Katello Agent itself. Those can be cleaned up whenever the time is right.

- Check out this PR
- Follow the steps in the README of https://github.com/jturel/katello_events (the instructions assume checkout is in /home/vagrant)
- Set Katello "with_event_daemon" to false:
```
[vagrant@centos8-katello-devel foreman]$ grep event_daemon ~/foreman/config/settings.plugins.d/katello.yaml -A1
  :event_daemon:
    :enabled: false
```
- start katello
- `systemctl start katello-candlepin-events` (check it's running with `status`)
- `systemctl start katello-event-queue` (check it's running with `status`)
- Register a system with subman which exercises both of these. You can see the activity in the logs:
```
[vagrant@centos8-katello-devel foreman]$ ack "api/internal" log/development.log 
2023-08-05T15:59:40 [I|app|46f1df7f] Started POST "/api/internal/event_queue/next" for 192.168.122.191 at 2023-08-05 15:59:40 +0000
2023-08-05T15:59:41 [I|app|49c0ffd5] Started POST "/api/internal/event_queue/heartbeat" for 192.168.122.191 at 2023-08-05 15:59:41 +0000
2023-08-05T15:59:45 [I|app|c8e8e077] Started POST "/api/internal/event_queue/next" for 192.168.122.191 at 2023-08-05 15:59:51 +0000
2023-08-05T15:59:52 [I|app|883ddd76] Started POST "/api/internal/event_queue/heartbeat" for 192.168.122.191 at 2023-08-05 15:59:52 +0000
2023-08-05T15:59:54 [I|app|b8328d6e] Started POST "/api/internal/event_queue/next" for 192.168.122.191 at 2023-08-05 15:59:54 +0000
2023-08-05T15:59:57 [I|app|0bb633f9] Started POST "/api/internal/candlepin_events/heartbeat" for 192.168.122.191 at 2023-08-05 15:59:57 +0000
2023-08-05T15:59:57 [I|app|d7615ed5] Started POST "/api/internal/event_queue/next" for 192.168.122.191 at 2023-08-05 15:59:57 +0000
2023-08-05T16:00:03 [I|app|d913349f] Started POST "/api/internal/event_queue/heartbeat" for 192.168.122.191 at 2023-08-05 16:00:03 +0000
2023-08-05T16:00:06 [I|app|c7c5d367] Started POST "/api/internal/event_queue/next" for 192.168.122.191 at 2023-08-05 16:00:06 +0000
2023-08-05T16:00:16 [I|app|7fff6325] Started POST "/api/internal/event_queue/heartbeat" for 192.168.122.191 at 2023-08-05 16:00:16 +0000
2023-08-05T16:00:18 [I|app|2b0aabd0] Started POST "/api/internal/candlepin_events/handle" for 192.168.122.191 at 2023-08-05 16:00:18 +0000
2023-08-05T16:00:18 [I|app|caec1fdc] Started POST "/api/internal/candlepin_events/handle" for 192.168.122.191 at 2023-08-05 16:00:18 +0000
2023-08-05T16:00:19 [I|app|61c4a269] Started POST "/api/internal/event_queue/next" for 192.168.122.191 at 2023-08-05 16:00:19 +0000
2023-08-05T16:00:23 [I|app|ac0a597d] Started POST "/api/internal/event_queue/next" for 192.168.122.191 at 2023-08-05 16:00:23 +0000
```

These services should be resilient to things like Katello going down, or returning unexpected status codes from the API so be clever in trying to break it!

**Known Issues**
- Ping API is not reporting OK
- Logging could be improved
- ???

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1,0 +1,21 @@
+# Katello Architecture
+
+## Event Queue
+The Event Queue enables asynchronous event processing with a few other requirements:
+
+### Requirements
+
+#### Deduplication
+Certain operations in Katello may trigger a flurry of events that would cause the same operation to be executed several times in a row. One example is generating Content Host Errata applicability. When an action triggering applicability generation occurs, an event is placed on the queue. Several applicability events may be queued for the same host in a brief period. The next time the Event Queue is drained it will calculate applicability for that host only once and remove the duplicate events afterward since their action has already been completed.
+
+#### Scheduling
+Certain operations need to happen at some point in the future from the time an event is queued. Events can be marked with a timestamp indicating when they should be handled. Until that point they'll remain untouched in the queue. One example of this is removal of a Content Host's Qpid queue (Katello Agent) which happens ten minutes after unregistering. If the host re-registers in that window its queue will be preserved.
+
+#### Retrying
+Certain events which might fail can be marked as needing to be retried. The functionality works on top of event scheduling. Each event can configure its retry wait individually. The Event Queue controls the total elapsed time that any event can be retried but doesn't limit the number of retries in that window.
+
+### Operation
+The Event Queue is drained by the Event Daemon which is different subsystem. Draining occurs every three seconds. The drain interval was chosen to strike a balance between timely handling of events and gaining benefit from deduplication. The Daemon runs the Event Queue [here](https://github.com/Katello/katello/blob/master/app/services/katello/event_monitor/poller_thread.rb).
+
+### Implementation
+The logic is consolidated in the [Event Queue Service](https://github.com/Katello/katello/blob/master/app/services/katello/event_queue.rb). It ultimately controls all ingress/egress of the [Katello::Event](https://github.com/Katello/katello/blob/master/app/models/katello/event.rb) table. All events must be [registered](https://github.com/Katello/katello/blob/master/lib/katello/engine.rb#L225-L230) in order to be processed.
--- a/app/controllers/katello/api/internal/candlepin_events_controller.rb
+++ b/app/controllers/katello/api/internal/candlepin_events_controller.rb
@@ -1,0 +1,21 @@
+module Katello
+  module Api
+    module Internal
+      class CandlepinEventsController < ::Katello::Api::InternalApiController
+        before_action :authorize_foreman_client
+
+        def handle
+          event = OpenStruct.new(params.slice(:subject, :content))
+          handler = ::Katello::Candlepin::EventHandler.new(Katello.internal_events_logger)
+          handler.handle(event)
+        end
+
+        def heartbeat
+          if params[:status].present?
+            Rails.cache.write(::Katello::Ping::CANDLEPIN_EVENTS_CACHE_KEY, params[:status], expires_in: 1.minute)
+          end
+        end
+      end
+    end
+  end
+end
--- a/app/controllers/katello/api/internal/event_queue_controller.rb
+++ b/app/controllers/katello/api/internal/event_queue_controller.rb
@@ -8,6 +8,17 @@ module Katello
           EventQueue.reset_in_progress
         end
 
+        def subscribe
+          subscription = Katello::EventQueue::Subscription.new
+          result = subscription.wait
+
+          if result
+            render json: 'ok'
+          else
+            head :no_content
+          end
+        end
+
         def next
           event = ::Katello::EventQueue.next_event
 
--- a/app/controllers/katello/api/internal/event_queue_controller.rb
+++ b/app/controllers/katello/api/internal/event_queue_controller.rb
@@ -1,0 +1,34 @@
+module Katello
+  module Api
+    module Internal
+      class EventQueueController < ::Katello::Api::InternalApiController
+        before_action :authorize_foreman_client
+
+        def reset
+          EventQueue.reset_in_progress
+        end
+
+        def next
+          event = ::Katello::EventQueue.next_event
+
+          unless event
+            return head :no_content
+          end
+
+          ::User.as_anonymous_admin do
+            handler = Katello::EventQueue::Handler.new(event)
+            handler.handle
+          end
+
+          render json: { event: event }
+        end
+
+        def heartbeat
+          if params[:status].present?
+            Rails.cache.write(::Katello::Ping::CANDLEPIN_EVENTS_CACHE_KEY, params[:status], expires_in: 1.minute)
+          end
+        end
+      end
+    end
+  end
+end
--- a/app/controllers/katello/api/internal_api_controller.rb
+++ b/app/controllers/katello/api/internal_api_controller.rb
@@ -1,0 +1,17 @@
+module Katello
+  module Api
+    class InternalApiController < ApplicationController
+      skip_before_action :verify_authenticity_token
+      skip_before_action :require_login
+      skip_before_action :session_expiry
+      skip_before_action :authorize
+
+      private
+
+      def authorize_foreman_client
+        client_cert = ::Foreman::ClientCertificate.new(request: request)
+        render json: nil, status: :forbidden unless client_cert.verified?
+      end
+    end
+  end
+end
--- a/app/lib/katello/event_queue/handler.rb
+++ b/app/lib/katello/event_queue/handler.rb
@@ -1,0 +1,34 @@
+module Katello
+  class EventQueue
+    class Handler
+      def initialize(event, logger = EventQueue.logger)
+        @event = event
+        @event_instance = ::Katello::EventQueue.create_instance(@event)
+        @logger = logger
+      end
+
+      def handle
+        Katello::Logging.time("katello event handled") do |data|
+          data[:type] = @event.event_type
+          data[:object_id] = @event.object_id
+          data[:expired] = false
+          data[:rescheduled] = false
+
+          begin
+            @event_instance.run
+          ensure
+            if @event_instance.try(:retry)
+              result = ::Katello::EventQueue.reschedule_event(@event)
+              if result == :expired
+                @logger.warn("event_queue_event_expired: type=#{@event.event_type} object_id=#{@event.object_id}")
+              elsif !result.nil?
+                @logger.warn("event_queue_rescheduled: type=#{@event.event_type} object_id=#{@event.object_id}")
+              end
+            end
+            ::Katello::EventQueue.clear_events(@event.event_type, @event.object_id, @event.created_at)
+          end
+        end
+      end
+    end
+  end
+end
--- a/app/lib/katello/event_queue/subscription.rb
+++ b/app/lib/katello/event_queue/subscription.rb
@@ -1,0 +1,26 @@
+module Katello
+  class EventQueue
+    class Subscription
+      TIMEOUT = 120 # Base this on httpd timeout?
+
+      def wait
+        start = Time.zone.now
+        loop do
+          if Katello.shutting_down?
+            Rails.logger.info "Canceling event queue subscription due to shutdown."
+            return
+          end
+
+          return true if Katello::EventQueue.runnable_events.any?
+
+          elapsed = Time.zone.now - start
+          break if elapsed > TIMEOUT
+
+          sleep 2
+        end
+
+        nil
+      end
+    end
+  end
+end
--- a/app/models/katello/ping.rb
+++ b/app/models/katello/ping.rb
@@ -4,6 +4,9 @@ module Katello
     FAIL_RETURN_CODE = 'FAIL'.freeze
     PACKAGES = %w(katello candlepin pulp qpid foreman tfm hammer).freeze
 
+    KATELLO_EVENTS_CACHE_KEY = 'katello_events_heartbeat'.freeze
+    CANDLEPIN_EVENTS_CACHE_KEY = 'candlepin_events_heartbeat'.freeze
+
     class << self
       def services(capsule_id = nil)
         proxy = fetch_proxy(capsule_id)
@@ -39,7 +42,7 @@ module Katello
         }
       end
 
-      def event_daemon_status(status, result)
+      def events_status(status, result)
         running = status&.dig(:running)
 
         if running
@@ -52,22 +55,20 @@ module Katello
 
       def ping_katello_events(result)
         exception_watch(result) do
-          status = Katello::EventDaemon::Runner.service_status(:katello_events)
-          event_daemon_status(status, result)
+          events_status(katello_events_status, result)
         end
       end
 
       def ping_candlepin_events(result)
         exception_watch(result) do
-          status = Katello::EventDaemon::Runner.service_status(:candlepin_events)
-          event_daemon_status(status, result)
+          events_status(candlepin_events_status, result)
         end
       end
 
       def ping_katello_agent(result)
         exception_watch(result) do
           status = Katello::EventDaemon::Runner.service_status(:katello_agent_events)
-          event_daemon_status(status, result)
+          events_status(status, result)
         end
       end
 
@@ -226,6 +227,22 @@ module Katello
       end
 
       private
+
+      def katello_events_status
+        if Katello.with_event_daemon?
+          Katello::EventDaemon::Runner.service_status(:katello_events)
+        else
+          Rails.cache.read(KATELLO_EVENTS_CACHE_KEY)
+        end
+      end
+
+      def candlepin_events_status
+        if Katello.with_event_daemon?
+          Katello::EventDaemon::Runner.service_status(:candlepin_events)
+        else
+          Rails.cache.read(CANDLEPIN_EVENTS_CACHE_KEY)
+        end
+      end
 
       def failed_services(result)
         result[:services].select do |_name, details|
--- a/app/services/katello/event_queue.rb
+++ b/app/services/katello/event_queue.rb
@@ -12,6 +12,10 @@ module Katello
       end
     end
 
+    def self.logger
+      ::Foreman::Logging.logger('katello/katello_events')
+    end
+
     def self.queue_depth
       ::Katello::Event.all.size
     end
--- a/config/routes/api/internal.rb
+++ b/config/routes/api/internal.rb
@@ -5,6 +5,7 @@ Katello::Engine.routes.draw do
         match '/heartbeat' => 'event_queue#heartbeat', via: :post
         match '/next' => 'event_queue#next', via: :post
         match '/reset' => 'event_queue#reset', via: :post
+        match '/subscribe' => 'event_queue#subscribe', via: :get
       end
 
       scope path: :candlepin_events, as: :candlepin_events do
--- a/config/routes/api/internal.rb
+++ b/config/routes/api/internal.rb
@@ -1,0 +1,16 @@
+Katello::Engine.routes.draw do
+  scope module: :api, path: :api do
+    scope module: :internal, path: :internal do
+      scope path: :event_queue, as: :event_queue do
+        match '/heartbeat' => 'event_queue#heartbeat', via: :post
+        match '/next' => 'event_queue#next', via: :post
+        match '/reset' => 'event_queue#reset', via: :post
+      end
+
+      scope path: :candlepin_events, as: :candlepin_events do
+        match '/handle' => 'candlepin_events#handle', via: :post
+        match '/heartbeat' => 'candlepin_events#heartbeat', via: :post
+      end
+    end
+  end
+end
--- a/lib/katello/engine.rb
+++ b/lib/katello/engine.rb
@@ -262,4 +262,8 @@ module Katello
   def self.with_ansible?
     Foreman::Plugin.installed?("foreman_ansible")
   end
+
+  def self.shutting_down?
+    ::Puma::Server.current.shutting_down?
+  end
 end
--- a/lib/katello/engine.rb
+++ b/lib/katello/engine.rb
@@ -121,6 +121,7 @@ module Katello
 
     initializer "katello.paths", :before => :sooner_routes_load do |app|
       app.routes_reloader.paths << "#{Katello::Engine.root}/config/routes/api/v2.rb"
+      app.routes_reloader.paths << "#{Katello::Engine.root}/config/routes/api/internal.rb"
       app.routes_reloader.paths << "#{Katello::Engine.root}/config/routes/api/rhsm.rb"
       app.routes_reloader.paths << "#{Katello::Engine.root}/config/routes/api/registry.rb"
       app.routes_reloader.paths.unshift("#{Katello::Engine.root}/config/routes/overrides.rb")
@@ -244,6 +245,14 @@ module Katello
 
   def self.with_katello_agent?
     !!SETTINGS.dig(:katello, :agent, :enabled)
+  end
+
+  def self.with_event_daemon?
+    !!SETTINGS.dig(:katello, :event_daemon, :enabled)
+  end
+
+  def self.internal_events_logger
+    ::Foreman::Logging.logger('katello/internal_events')
   end
 
   def self.remote_execution_by_default?
--- a/lib/katello/plugin.rb
+++ b/lib/katello/plugin.rb
@@ -242,6 +242,7 @@ Foreman::Plugin.register :katello do
   logger :registry_proxy, :enabled => true
   logger :katello_events, :enabled => true
   logger :candlepin_events, :enabled => true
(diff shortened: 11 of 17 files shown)


jturel:
@jeremylenz I think this is ready for a test run if you're interested. I've been working on this over a period of months so I probably forgot something in the steps. Let me know if you hit any issues!

jeremylenz (commented):
Thanks @jturel ! I gave this a spin, and everything seems to be working fine.

```
# sudo journalctl -f -u katello-candlepin-events -u katello-event-queue
Aug 07 13:20:40 centos8-katello-devel-stable.example.com katello-candlepin-events[230643]: Subscribed to katello.candlepin.candlepin_events
Aug 07 13:20:40 centos8-katello-devel-stable.example.com katello-candlepin-events[230643]: started candlepin event listener
Aug 07 13:21:21 centos8-katello-devel-stable.example.com katello-event-queue[230644]: event queue processed=1
Aug 07 13:28:51 centos8-katello-devel-stable.example.com katello-candlepin-events[230643]: Handled owner_content_access_mode.modified
Aug 07 13:28:51 centos8-katello-devel-stable.example.com katello-candlepin-events[230643]: Handled compliance.created
Aug 07 13:30:26 centos8-katello-devel-stable.example.com katello-candlepin-events[230643]: Handled owner_content_access_mode.modified
Aug 07 13:34:51 centos8-katello-devel-stable.example.com katello-event-queue[230644]: event queue processed=1
Aug 07 13:35:21 centos8-katello-devel-stable.example.com katello-event-queue[230644]: event queue processed=1
Aug 07 13:43:19 centos8-katello-devel-stable.example.com katello-event-queue[230644]: event queue processed=1
```

Here are some observations:

This is what it looks like when I trigger a host applicability generation from the new host details page. This generates a Katello Event. According to your instructions I would expect to see a call to `/api/internal/handle`, but I don't. Is this expected?

```
13:35:21 rails.1   | 2023-08-07T13:35:21 [I|app|20ad7efa] Started PUT "/api/v2/hosts/1/errata/applicability" for 192.168.122.1 at 2023-08-07 13:35:21 +0000
13:35:21 rails.1   | 2023-08-07T13:35:21 [I|app|20ad7efa] Processing by Katello::Api::V2::HostErrataController#applicability as JSON
13:35:21 rails.1   | 2023-08-07T13:35:21 [I|app|20ad7efa]   Parameters: {"api_version"=>"v2", "host_id"=>"1", "host_erratum"=>{}}
13:35:21 rails.1   | 2023-08-07T13:35:21 [D|app|20ad7efa]   Rendering layout /home/vagrant/katello/app/views/katello/api/v2/layouts/resource.json.erb
13:35:21 rails.1   | 2023-08-07T13:35:21 [D|app|20ad7efa]   Rendering /home/vagrant/katello/app/views/katello/api/v2/common/async.json.rabl within katello/api/v2/layouts/resource
13:35:21 rails.1   | 2023-08-07T13:35:21 [I|app|20ad7efa]   Rendered /home/vagrant/katello/app/views/katello/api/v2/common/async.json.rabl within katello/api/v2/layouts/resource (Duration: 0.2ms | Allocations: 124)
13:35:21 rails.1   | 2023-08-07T13:35:21 [I|app|20ad7efa]   Rendered layout /home/vagrant/katello/app/views/katello/api/v2/layouts/resource.json.erb (Duration: 0.4ms | Allocations: 290)
13:35:21 rails.1   | 2023-08-07T13:35:21 [I|app|20ad7efa] Completed 202 Accepted in 16ms (Views: 0.7ms | ActiveRecord: 6.8ms | Allocations: 5290)
13:35:21 rails.1   |  20ad7efa | 
13:35:21 rails.1   |  20ad7efa | 
13:35:21 rails.1   | 2023-08-07T13:35:21 [D|app|20ad7efa] With body:   []
13:35:21 rails.1   |  20ad7efa | 
13:35:21 rails.1   |  20ad7efa | 
13:35:21 rails.1   | 2023-08-07T13:35:21 [I|app|b680395f] Started POST "/api/internal/event_queue/next" for 127.0.0.1 at 2023-08-07 13:35:21 +0000
13:35:21 rails.1   | 2023-08-07T13:35:21 [I|app|b680395f] Processing by Katello::Api::Internal::EventQueueController#next as HTML
13:35:21 rails.1   | 2023-08-07T13:35:21 [I|app|b680395f]   Parameters: {"event_queue"=>{}}
13:35:21 rails.1   | 2023-08-07T13:35:21 [I|bac|b680395f] Task {label: , execution_plan_id: fba8c0e0-9939-4e8e-9835-bbaead9bad7c} state changed: pending 
13:35:21 rails.1   | 2023-08-07T13:35:21 [I|bac|b680395f] Task {label: Actions::Katello::Applicability::Hosts::BulkGenerate, id: 22cb33e3-6142-4c77-9522-ce725bbdc0f9, execution_plan_id: fba8c0e0-9939-4e8e-9835-bbaead9bad7c} state changed: planning 
13:35:21 rails.1   | 2023-08-07T13:35:21 [I|bac|b680395f] Task {label: Actions::Katello::Applicability::Hosts::BulkGenerate, id: 22cb33e3-6142-4c77-9522-ce725bbdc0f9, execution_plan_id: fba8c0e0-9939-4e8e-9835-bbaead9bad7c} state changed: planned 
13:35:21 rails.1   | 2023-08-07T13:35:21 [I|app|b680395f] katello event handled success=true type=generate_host_applicability object_id=0 expired=false rescheduled=false duration=63.44
13:35:21 rails.1   | 2023-08-07T13:35:21 [I|app|b680395f] Completed 200 OK in 85ms (Views: 0.3ms | ActiveRecord: 21.3ms | Allocations: 22195)
```

However, I do see
```
# sudo journalctl -f -u katello-candlepin-events -u katello-event-queue
Aug 07 13:43:19 centos8-katello-devel-stable.example.com katello-event-queue[230644]: event queue processed=1
```

and the task completes fine, so it seems to be working?

I do see this when I toggle SCA (a Candlepin event):
```
13:59:55 rails.1   | 2023-08-07T13:59:55 [I|kat|b2d8a788] candlepin event handled success=true subject=owner_content_access_mode.modified entity_id=ff8080818997807a018997928ea20001 duration=145.43
13:59:55 rails.1   | 2023-08-07T13:59:55 [I|app|b2d8a788] No template found for Katello::Api::Internal::CandlepinEventsController#handle, rendering head :no_content
13:59:55 rails.1   | 2023-08-07T13:59:55 [I|app|b2d8a788] Completed 204 No Content in 159ms (ActiveRecord: 7.7ms | Allocations: 21255)
```

Is `No template found for Katello::Api::Internal::CandlepinEventsController#handle, rendering head :no_content` expected?


Observation #2 - This is what my `development.log` tail looks like now:
```
13:55:18 rails.1   |  cd6eabe0 | 
13:55:19 rails.1   | 2023-08-07T13:55:19 [I|app|0d4cff51] Started POST "/api/internal/event_queue/next" for 127.0.0.1 at 2023-08-07 13:55:19 +0000
13:55:19 rails.1   | 2023-08-07T13:55:19 [I|app|0d4cff51] Processing by Katello::Api::Internal::EventQueueController#next as HTML
13:55:19 rails.1   | 2023-08-07T13:55:19 [I|app|0d4cff51]   Parameters: {"event_queue"=>{}}
13:55:19 rails.1   | 2023-08-07T13:55:19 [I|app|0d4cff51] Completed 204 No Content in 11ms (ActiveRecord: 6.3ms | Allocations: 1283)
13:55:19 rails.1   |  0d4cff51 | 
13:55:19 rails.1   |  0d4cff51 | 
13:55:22 rails.1   | 2023-08-07T13:55:22 [I|app|586f7754] Started POST "/api/internal/event_queue/next" for 127.0.0.1 at 2023-08-07 13:55:22 +0000
13:55:22 rails.1   | 2023-08-07T13:55:22 [I|app|586f7754] Processing by Katello::Api::Internal::EventQueueController#next as HTML
13:55:22 rails.1   | 2023-08-07T13:55:22 [I|app|586f7754]   Parameters: {"event_queue"=>{}}
13:55:22 rails.1   | 2023-08-07T13:55:22 [I|app|586f7754] Completed 204 No Content in 12ms (ActiveRecord: 5.9ms | Allocations: 1283)
13:55:22 rails.1   |  586f7754 | 
13:55:22 rails.1   |  586f7754 | 
13:55:25 rails.1   | 2023-08-07T13:55:25 [I|app|9cff1b95] Started POST "/api/internal/event_queue/next" for 127.0.0.1 at 2023-08-07 13:55:25 +0000
13:55:25 rails.1   | 2023-08-07T13:55:25 [I|app|9cff1b95] Processing by Katello::Api::Internal::EventQueueController#next as HTML
13:55:25 rails.1   | 2023-08-07T13:55:25 [I|app|9cff1b95]   Parameters: {"event_queue"=>{}}
13:55:25 rails.1   | 2023-08-07T13:55:25 [I|app|9cff1b95] Completed 204 No Content in 8ms (ActiveRecord: 4.7ms | Allocations: 1283)
13:55:25 rails.1   |  9cff1b95 | 
13:55:25 rails.1   |  9cff1b95 | 
13:55:28 rails.1   | 2023-08-07T13:55:28 [I|app|91fe51f3] Started POST "/api/internal/event_queue/next" for 127.0.0.1 at 2023-08-07 13:55:28 +0000
13:55:28 rails.1   | 2023-08-07T13:55:28 [I|app|91fe51f3] Processing by Katello::Api::Internal::EventQueueController#next as HTML
13:55:28 rails.1   | 2023-08-07T13:55:28 [I|app|91fe51f3]   Parameters: {"event_queue"=>{}}
13:55:28 rails.1   | 2023-08-07T13:55:28 [I|app|91fe51f3] Completed 204 No Content in 6ms (ActiveRecord: 3.1ms | Allocations: 1283)
13:55:28 rails.1   |  91fe51f3 | 
13:55:28 rails.1   |  91fe51f3 | 
```

Especially with the calls to `/next` every 3 seconds, I wonder if there's a way to tame those logs?

also tagging @ehelms for review here because Packaging will be involved if we spin up these new services.

jeremylenz:
oh also, can you pls make a Redmine? thx :innocent:

jeremylenz:
Oh also, I observed that Katello agent actions stopped working when I spun this up.

```
Host did not respond within 20 seconds. The task has been cancelled. Is katello-agent installed and goferd running on the Host?
```

 But I'm assuming that's because I disabled the Katello event daemon, and there's no corresponding service to call our new APIs.

jturel:
@jeremylenz thanks for the quick feedback! responses below

> This generates a Katello Event. According to your instructions I would expect to see a call to /api/internal/handle, but I don't. Is this expected?

All of the Event Queue handling is done through the `next` endpoint which you can see was called in the log output you shared, so that looks good to me.

> Is No template found for Katello::Api::Internal::CandlepinEventsController#handle, rendering head :no_content expected?

It's not a big deal (API behavior isn't affected), but I can fix that cause it's not nice to look at

> I wonder if there's a way to tame those logs?

There probably is. It's a little bit nuanced because I think the logging is valuable - it shows when events are being handled, their execution time, and (importantly) the number of Allocations for the request which is really valuable for debugging & being aware of performance problems. To silence those logs entirely might be too aggressive. In dev, it could be a matter of polling every 10-15 seconds instead. I wonder if it's possible to log those requests to a different file  or not stdout, although that would be pretty unconventional. Just spitballing - and definitely something for you all to think about and probably make the call on here.

> Oh also, I observed that Katello agent actions stopped working when I spun this up.

Exactly right for the reasons you mentioned. This code can be shipped in a disabled state until Katello Agent is removed and that'd be my recommendation rather than keeping the Daemon running only for the Agent for a very brief period while running the new services in tandem. Order of operations could be:
1. merge all of this work and get it into a deliverable (packaged, installed, disabled) state
2. wrap up katello agent removal
3. turn new services on & disable the event daemon
4. remove event daemon code in katello / any other cleanup after a period of time in case a fallback needs to happen

You could even wait until 3.11 altogether to do the switchover and reduce the amount of change in 3.10

jturel:
Another option to quiet the Event Queue logs down would be for a file to be touched by Katello somewhere on the disk. The new service would check the presence of this file to know when to start sending API requests to clear the queue. This adds another layer of complexity to the picture which I think takes away from the simplicity of the way I have it written currently, but it's certainly possible. Totally open to feedback. We can take our time to come up with the right thing.

jturel:
> oh also, can you pls make a Redmine so that Focaccia can bake us a review card? thx 😇

As soon as I hear "yes, let's do this" I can build out a tracker!

ehelms:
A big +1 to the progressive approach, that makes merging and orchestrating the changes much easier over the course of 1 or more releases and provides better test predictability with smaller changes.

These would run as systemd services now and as such we should follow our pattern for including those in the git repo (e.g. https://github.com/theforeman/foreman/tree/develop/extras/systemd). You can do that here or as a follow up towards operationalizing them.

What is the expectation for how these would work in a development environment?

jturel:
> What is the expectation for how these would work in a development environment?

I'd like them to run as systemd services as well, but running from a git checkout in /home/vagrant for easy development without having to hack the service definition, etc.

ehelms:
The idea is that you store the systemd file in the repository to manage the standard definition that can be used on any Linux system. Then use systemd dropin files to configure it for the environment which the installer handles: https://github.com/theforeman/puppet-foreman/blob/0a2b3100dbac7c1003d261220329cdb3148003a4/manifests/config.pp#L88-L93

jturel:
> The idea is that you store the systemd file in the repository to manage the standard definition that can be used on any Linux system. Then use systemd dropin files to configure it for the environment which the installer handles: https://github.com/theforeman/puppet-foreman/blob/0a2b3100dbac7c1003d261220329cdb3148003a4/manifests/config.pp#L88-L93

TIL! Thanks - should work well

jturel:
@jeremylenz I added a commit to this PR and a branch to katello_events (https://github.com/jturel/katello_events/tree/long_poll) that uses a 'long polling' technique that allows the service to hold an http connection open until an event is received at which point it starts to process (call /next). It "resubscribes" every two minutes which means 1 request every 2 minutes assuming there are no events coming by. I think that addresses your concern about log noise. There are probably bugs - I whipped this up at 2 in the morning - but give it a try.

The downside is that the two-minute subscription, left unchecked, blocks puma from stopping since it waits for all requests to complete. I added `Katello.shutting_down?` to detect shutdown which is dependent on Puma so this is now an app-server specific solution. I think it's OK - you're not moving away from Puma any time soon ;)

jturel:
@ehelms @jeremylenz with the long-polling approach here for the event queue (katello holding a 1+ minute HTTP connection open, waiting for the signal to start event processing) I had to work around puma shutdown being delayed (see my above comment on that). The approach also delays httpd from stopping because it's also waiting for incoming requests to finish. Can the new services talk to Puma directly instead of going through httpd? That should prevent anything from taking a long time to shut down or restart. Long-polling stops a lot of log noise and keeps the traffic patterns easier to analyze/interpret so I'm a fan of using it and getting around this quirk.

ehelms:
Talking to Puma directly leaves a hole open where if we need to scale we might hit issues and I try to avoid piling on if possible. Can you explain more why the long-polling is needed?

jturel:
> Talking to Puma directly leaves a hole open where if we need to scale we might hit issues and I try to avoid piling on if possible.

I'm not sure how these things are related - can you elaborate?

> Can you explain more why the long-polling is needed?

It's not needed, but it avoids polling the server every several seconds to know when there are events to clear out which generates a lot of log noise. However, I'm kind of losing interest in plugging the holes around potentially blocking Puma or Apache shutdowns.


> Remind me -- what is the event queue handling? Is it just applicability?

A bunch of stuff actually: https://github.com/Katello/katello/blob/master/lib/katello/engine.rb#L225-L230

jturel:
@ehelms other potential avenues:

1. Use a ~20 second long poll. The server will be blocked by at most 20 seconds from shutting down and not generate too much log noise. Any shorter than that and might as well do interval-based polling, imo.

2. The external service connects to the db to run a query to see if the queue needs to drain. I wouldn't want to hard-code this query in the service - there can be an API endpoint to tell the service which query to run that way all queue logic stays in Katello itself. If the query is positive, the service issues http requests to drain.

3. Katello periodically writes something to /var/spool. Tricky part is when to write this. It could be coupled to (certain) HTTP requests, but I don't really like that.

4. Rewrite the event queue to use redis, beanstalkd, etc

5. <someones better idea here!>

ehelms:
At a design level what would a Redis based solution look like? We have it
already in the architecture for a few features.

jturel:
> At a design level what would a Redis based solution look like? We have it already in the architecture for a few features.

A lot of the logic in the event queue is around de-duplicating event handling ie if a flood of Generate Applicability events for a single host are present in the queue and ready to process, only handle one and delete the rest. Another main feature is to schedule an event to process in the future, and possibly to retry handling a few times across a few re-schedules. AFAIK all of those things should be more natural to do with Redis. The service which itself would listen for events and send those events to Katello as a payload for processing.

This would blow the scope up of what I'm trying to do here, so I think it would need to be reduced, for now, to deal only with removing Candlepin from the Event Daemon construct inside Katello, and leave the event queue improvements for later

jturel:
@ehelms I think option 2 is well worth considering: it avoids a rewrite, doesn't block the server from shutting down, and only generates API traffic when handling actual events though I suppose there would be a periodic heartbeat too. It also uses existing the existing queue implementation on the server with no disturbance

ehelms:
> gosh, I didn't mean to cause such a ruckus!

If it helps, I think these discussions would have come around one day eventually, @jturel is just nicely bringing them up ahead of time. And while my questions have the potential to expand the scope and breadth, do feel strongly any re-design needs to ensure it consider scalability looking ahead so we do not have to repeat this process again. Progressive enhancement is good, complete re-write again would be bad.

Would the external service connecting to the database re-use the Rails database configuration and connection configuration?

jturel:
> Would the external service connecting to the database re-use the Rails database configuration and connection configuration?

I'm thinking more like:

- create a read-only user for the purpose
- service gets the db host/user/pass from environment config
-  service uses the pg gem to connect and run the select

The same credentials could be used but it seems trivial to create dedicated user. If by config you're talking so far as pooling - this would be a single connection in a single thread

ehelms:
After the mention of Redis, it got me wondering about if this is unique to us or a problem set that (should) affect more Rails applications. I know we had looked a the ActiveJob method but I believe deemed is as trying to put a square block in a round hole. I saw this project https://github.com/RailsEventStore/rails_event_store and wondered if anybody had ever looked at it?

I know I am adding potentially more scope, but architectural changes I like to be more pedantic with these days.

jturel:
> After the mention of Redis, it got me wondering about if this is unique to us or a problem set that (should) affect more Rails applications. I know we had looked a the ActiveJob method but I believe deemed is as trying to put a square block in a round hole. I saw this project https://github.com/RailsEventStore/rails_event_store and wondered if anybody had ever looked at it?
> 
> I know I am adding potentially more scope, but architectural changes I like to be more pedantic with these days.

Appears to be much heavier than what Katello needs, and doesn't seem like it offers de-duplication and (re)scheduling out of the box. It looks more like a typical pub/sub thing which isn't what the Event Queue is.

ehelms:
> > After the mention of Redis, it got me wondering about if this is unique to us or a problem set that (should) affect more Rails applications. I know we had looked a the ActiveJob method but I believe deemed is as trying to put a square block in a round hole. I saw this project https://github.com/RailsEventStore/rails_event_store and wondered if anybody had ever looked at it?
> > I know I am adding potentially more scope, but architectural changes I like to be more pedantic with these days.
> 
> Appears to be much heavier than what Katello needs, and doesn't seem like it offers de-duplication and (re)scheduling out of the box. It looks more like a typical pub/sub thing which isn't what the Event Queue is.

That goes to show I don't exactly know what Event Queue is even though I thought I did :smile: @jturel @jeremylenz what do you think about including somewhere (Katello README? discourse post? we don't really have a great place at present for these) a write up of what Event Queue is and how it works. I think that will help me, and others who end up reviewing some of the parts and pieces of the architectural changes as it gets modified.

jeremylenz:
> what do you think about including somewhere (Katello README? discourse post? we don't really have a great place at present for these) a write up of what Event Queue is and how it works.

Maybe it's time for Katello to add an [ARCHITECTURE.md](https://matklad.github.io//2021/02/06/ARCHITECTURE.md.html)...

otherwise, https://community.theforeman.org/c/development/9 would be a great place for it!

jturel:
I like the architecture.md idea - I can write a quick summary and add it to this PR reflecting the current state

jturel:
@ehelms @jeremylenz https://github.com/Katello/katello/blob/c75ad67851b2560fd31d76bc2c4154658907aff8/ARCHITECTURE.md

ehelms:
Awesome. Now to tie back the parts of that document, and so I don't misinterpret, what parts of the process will the new service handle and what parts will be within Katello? This might be good to add to it, or a diagram once we know the back and forth clearly.

jturel:
> Awesome. Now to tie back the parts of that document, and so I don't misinterpret, what parts of the process will the new service handle and what parts will be within Katello? This might be good to add to it, or a diagram once we know the back and forth clearly.

My main goal here is to separate the Event Queue from the Event Daemon (to remove this piece). I wrote an API (this PR) that something can call to drain the queue. The total functionality of the Event Queue still remains untouched and in Katello. The new service pings this API to drain the queue and knows/does nothing except that. It's not sending the events itself and that's the beauty of it.

This discussion started with how to ping that API intelligently (only when there are events). Right now I'm convinced that the service going to the db to detect if the queue needs a drain is a straightforward and non-problematic way to go. Any future evolution of the event queue such as a Redis rewrite (I don't see the point for such a stable component unless the requirements become more complex) will more than likely need a separate service anyway, so this is compatible in that way.

jturel:
@jeremylenz @ehelms Another crazy idea is to merge this PR only with the Candlepin Events side of the changes since they're not subject to anything we're talking about here. The reason for this would be to see if we can remove the Event Queue altogether. Yep, I said it. It's probably time to evaluate where the thing stands and if there are better ways to achieve the same end results that the EQ gives us without actually having that construct. That would take quite a bit of time, though. Additional benefit is that it even further reduces the total amount of change at one time.

ehelms:
> @jeremylenz @ehelms Another crazy idea is to merge this PR only with the Candlepin Events side of the changes since they're not subject to anything we're talking about here. The reason for this would be to see if we can remove the Event Queue altogether. Yep, I said it. It's probably time to evaluate where the thing stands and if there are better ways to achieve the same end results that the EQ gives us without actually having that construct. That would take quite a bit of time, though. Additional benefit is that it even further reduces the total amount of change at one time.

I like where your head is at. Smaller changes and taking a moment to step back and ask the broader question. Obviously the EQ has and continues to work and do it's job, but there's no harm in the light of other changes to re-assess assumptions from before and new information to ask if we can simplify things. 

That is what I found interesting about the capabilities in the `architecture.md` and the list of actions here https://github.com/Katello/katello/blob/master/lib/katello/engine.rb#L225-L230 that is governed by the EQ.

jturel:
> That is what I found interesting about the capabilities in the `architecture.md` and the list of actions here https://github.com/Katello/katello/blob/master/lib/katello/engine.rb#L225-L230 that is governed by the EQ.

Yep, need to evaluate those individually to see what can happen.

In order to preserve the conclusions here I'm gonna close this PR and open a new one which addresses only the Candlepin Events side. I'll bring in the architecture.md there too.

ehelms:
There is a related, interesting conversation happening within an RFC, I pinged on it but posting here to close the loop -- https://community.theforeman.org/t/rfc-redis-as-the-default-cache/34650/7

PR #10587, opened 2023-06-06
Fixes #36478 - add ouiaId to Tab components

MariaAga:


sjha4:
Able to see missing ouia-id on Tabs with https://github.com/theforeman/foreman-js/pull/470.
ouia-id_failures
This PR fixes the missing ouia-id warnings in katello. Ack 👍🏼

PR #10588, opened 2023-06-07
Fixes #36484 - Add :view_params permission to 'Register hosts' role

stejskalleos:
 Add `:view_params` permission to the 'Register hosts' role

Global Registration using a user with the `Register Hosts` role ignores all the setup options (host parameters)

1. Create a user with the `Register Hosts` role.
2. Using the new user, generate a global registration command with all the Setup options set to No.
3. Execute the command on the client.

**Actual results:** REX & insights client is installed, params have the value `true`
**Expected result:** The opposite


stejskalleos:
@ianballou Hi, could you please take a look? We'd like to have the fix in the next stream snap if possible.

ianballou (approved):

PR #10595, opened 2023-06-09
Fixes #36498 - Don't try to unset a host's CV/LCE when inheriting from hostgroup

jeremylenz:

Redefines Foreman's `apply_inherited_attributes` method (from `Host::Managed`) to address the case where either `content_view_id` or `lifecycle_environment_id` are blank. Since they must now be provided together, having a hostgroup where only one of them is set causes problems.

This whole issue should go away once we make the same multi-CV changes to `Hostgroup::ContentFacet` that have already been made to `Host::ContentFacet`.

Create a hostgroup with CV and LCE
Edit the hostgroup and remove the LCE (click the X)
Go to one of your hosts and click Edit
Assign the host to the hostgroup and click Submit

Before patch: an error `content_view_id and lifecycle_environment_id must be provided together`
After patch: no errors

Click around other host and hostgroup edit flows and make sure nothing is broken
- host assigned to hostgroup with no CV and no LCE
- host assigned to hostgroup with both CV and LCE
- host not assigned to hostgroup

### DIFF
--- a/app/models/katello/concerns/host_managed_extensions.rb
+++ b/app/models/katello/concerns/host_managed_extensions.rb
@@ -33,10 +33,10 @@ module Katello
         def apply_inherited_attributes(attributes, initialized = true)
           attributes = super(attributes, initialized)
           facet_attrs = attributes['content_facet_attributes']
-          return attributes unless facet_attrs
+          return attributes if facet_attrs.blank?
           cv_id = facet_attrs['content_view_id']
           lce_id = facet_attrs['lifecycle_environment_id']
-          if cv_id.blank? || lce_id.blank?
+          if initialized && (cv_id.blank? || lce_id.blank?)
             if cv_id.blank?
               Rails.logger.info "Hostgroup has no content view assigned; using host's existing content view"
               facet_attrs['content_view_id'] = content_facet&.single_content_view&.id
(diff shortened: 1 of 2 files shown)


app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -30,6 +30,28 @@ def inherited_attributes
           inherited_attrs
         end
 
+        def apply_inherited_attributes(attributes, initialized = true)
+          attributes = super(attributes, initialized)
+          facet_attrs = attributes['content_facet_attributes']
+          return attributes if facet_attrs.blank?
+          cv_id = facet_attrs['content_view_id']
+          lce_id = facet_attrs['lifecycle_environment_id']
+          if initialized && (cv_id.blank? || lce_id.blank?)
```
jeremylenz, line 39: If both `cv_id` and `lce_id` are blank, what then?

I considered adding explicit logic for this. But I think we're okay without it, because the two cases are

1. Host with content facet: will already have a valid LCE and CV, and the host will use its existing values
2. Host without content facet: `facet_attrs` will be `nil`, meaning this code path isn't touched

anyway, I couldn't get it to error in my testing. But let me know if I'm wrong :D

ianballou (approved):
Working fine for me!

PR #10598, opened 2023-06-09
Fixes #36477 - Correctly imports  repos with arch restrictions

parthaa:
Red Hat repositories incorrectly allow users to set architecture restrictions on the repo details page. This causes issues when exporting and importing these repositories due to the way enablement worked.

Initially we went the route of no allowing user to update arch information on a redhat repo. Which would imply that the they'd need to run an upgrade task to correct stuff.

This PR allows for exports and imports for the wrong arch via being smart on enablement. 

Originally thought of adding some complexity in the behavior by adding a `original arch substitutions` key to the export `metadata.json`. 

But instead decided to go with the simpler "Enable Repo and then override the arch to what comes from the metadata json" approach

Before you check out the PR
- Enable a couple of redhat repos (suggest something you can quickly sync.)
- Change their architectures
- Set the download policy to immediate 
- Sync them
- Export that repo and try to import it in a new org 
- You should see the task completing successfully but the imported repo has zero content.

After checking out the PR
- Go to the repo details of one of the redhat repos
- Finally reimport the content in a  new org 
- Verify that the content appears correctly.

### DIFF
--- a/app/lib/actions/katello/content_view_version/auto_create_redhat_repositories.rb
+++ b/app/lib/actions/katello/content_view_version/auto_create_redhat_repositories.rb
@@ -15,7 +15,8 @@ module Actions
             helper.creatable.each do |root|
               plan_action(::Actions::Katello::RepositorySet::EnableRepository,
                             root[:product], root[:content], root[:substitutions],
-                            override_url: root[:override_url])
+                            override_url: root[:override_url],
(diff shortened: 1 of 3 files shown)


parthaa:
Discussed with @sjha4 . Going to remove the imports part of this PR since the import errors won't show up if the arches were correct in the original repository in the first place.

PR #10600, opened 2023-06-12
Fixes #36440 - Redefine #attributes= to hijack old content facet attributes

jeremylenz:

When updating a host using the `Api::V2::HostController`, Foreman doesn't call `@host.update` - instead, it does

```rb
@host.attributes = host_attributes(host_params, @host)
...
@host.save
```

The consequence of this is that the host's `update` method redefined in `host_managed_extensions` was never getting called. This is what removed the old deprecated content facet attributes `content_view_id` and `lifecycle_environment_id`, so as a consequence, those attributes were still present when Foreman called `@host.attributes=`. Turns out the `#attributes=` method checks and validates all the nested attributes, so when we try to include imaginary ones like `content_view_id`, it would throw an error:

```
Backtrace for 'Action failed' error (ActiveModel::UnknownAttributeError): unknown attribute 'content_view_id' for Katello::Host::ContentFacet.
```

With this change, the `attributes=` method of `Host::Managed` gets the same enhancement that `update` already had. This will hopefully prevent this problem until we can properly finish the multi-CV related work.

Just like https://github.com/Katello/katello/pull/10595, this issue should go away completely when we update `Hostgroup::ContentFacet` to truly handle multiple content views.

Create a hostgroup and assign both CV and LCE
Assign a host to it
Update the host via hammer (this is important because hammer will use the API hosts controller, not the regular one) and include a hostgroup update:

```
hammer host update --id 5 --lifecycle-environment-id 1 --content-view-id 2 --hostgroup "My hostgroup"
```

Before: you should get the error above in the Rails log, and Hammer will complain:

```
Could not update the host:
  Internal Server Error: the server was unable to finish the request. This may be caused by unavailability of some required service, incorrect API call or a server-side bug. There may be more information in the server's logs.
```

After:

```
Host updated.
```

### DIFF
--- a/app/models/katello/concerns/host_managed_extensions.rb
+++ b/app/models/katello/concerns/host_managed_extensions.rb
@@ -6,7 +6,7 @@ module Katello
       include ForemanTasks::Concerns::ActionSubject
 
       module Overrides
-        def update(attrs)
+        def check_cve_attributes(attrs)
           if attrs[:content_facet_attributes]
             cv_id = attrs[:content_facet_attributes].delete(:content_view_id)
             lce_id = attrs[:content_facet_attributes].delete(:lifecycle_environment_id)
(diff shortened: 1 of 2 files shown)


ianballou (commented):
Working great.  Maybe add a quick test that runs attributes= on a host?

PR #10604, opened 2023-06-13
Fixes #36401 - Update content_facet rabl to show multi_env like in db

chris1984:

* Updated the rabl to return cv/env how we have it in the DB
```bash
             "content_view_environments" => [
            [0] {
                         "content_view" => {
                           "id" => 2,
                         "name" => "Zoo",
                    "composite" => false
                },
                "lifecycle_environment" => {
                      "id" => 1,
                    "name" => "Library"
                }
            },
            [1] {
                         "content_view" => {
                           "id" => 3,
                         "name" => "Plumbing",
                    "composite" => false
                },
                "lifecycle_environment" => {
                      "id" => 2,
                    "name" => "Custom"
                }
            }
        ],
```

* Check out PR
* Register a host to Multiple envs and check out hammer PR https://github.com/Katello/hammer-cli-katello/pull/894
* `hammer host info --id X`
* Also check to make sure it looks good on a host with just 1 CV/ENV

### DIFF
--- a/app/views/katello/api/v2/content_facet/base.json.rabl
+++ b/app/views/katello/api/v2/content_facet/base.json.rabl
@@ -1,6 +1,4 @@
 attributes :id, :uuid
-attributes :content_view_id, :content_view_name
-attributes :lifecycle_environment_id, :lifecycle_environment_name
 attributes :content_source_id, :content_source_name
 attributes :kickstart_repository_id, :kickstart_repository_name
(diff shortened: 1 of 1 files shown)


jeremylenz (approved):
Checked hammer with single and multiple environments

:heavy_check_mark: displays info correctly
:heavy_check_mark: displays environments in the same order I registered them with

ACK :+1:

PR #10607, opened 2023-06-15
Fixes #36506 - allow installable errata counts in safe mode

snagoor:
Allow installable errata counts in safe mode for foreman reporting.


wbclark:
Merging based on result of testing w/ https://github.com/theforeman/foreman/pull/9734 (there was an issue, but it was related to the actual template implementation on the foreman side)

PR #10611, opened 2023-06-15
Fixes #36511 - Update aria-label for last_refresh component

sjha4:
Update aria-label last_refresh value as recommended.
Go to ACS details > Inspect Last refresh value and make sure aria-label is updated as requested by automation.


jeremylenz (approved):
The aria-label is now correct.

PR #10612, opened 2023-06-15
Fixes #36512 - Update ids on ssl client key select

sjha4:
Go to ACS details > Credentials > Edit
On the modal inspect the "SSL client key" FormSelect option.
Aria-label and ouia-id should be updated as requested.


jeremylenz (approved):
ouiaId=LGTM
aria-label=LGTM

PR #10614, opened 2023-06-15
Fixes #36514 - Update ids on ACS edit product modal action buttons

sjha4:
Create a simplified ACS with some products.
Go to ACS details > Edit products > Inspect Save and cancel buttons and ensure aria-label and ouia-ids are correctly updated.


jeremylenz (approved):
Works!

:+1:

PR #10616, opened 2023-06-16
Fixes #36482 - Use string keys for accessing 'arch' and 'minor' in re…

wbclark:
…po hash

### DIFF
--- a/app/lib/katello/resources/cdn/katello_cdn.rb
+++ b/app/lib/katello/resources/cdn/katello_cdn.rb
@@ -11,10 +11,14 @@ module Katello
           super
         end
 
-        def fetch_paths(content_path)
+        def fetch_repo_set(content_path)
           url = "/katello/api/v2/repository_sets?organization_id=#{organization['id']}&search=#{CGI.escape("path = #{content_path}")}"
           response = get(url)
-          repo_set = JSON.parse(response)['results'].first
+          JSON.parse(response)['results'].first
+        end
+
+        def fetch_paths(content_path)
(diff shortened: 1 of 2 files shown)


app/lib/katello/resources/cdn/katello_cdn.rb
```diff
@@ -33,8 +33,8 @@ def fetch_paths(content_path)
           results = json_body['results']
 
           results.map do |repo|
-            Katello::Content.substitute_content_path(arch: repo[:arch],
-                                                     releasever: repo[:minor],
+            Katello::Content.substitute_content_path(arch: repo['arch'],
+                                                     releasever: repo['minor'],
                                                      content_path: content_path)
```
sjha4, line 38: Can we do something like:
repo_hash = HashWithIndifferentAccess.new(repo) and then we won't have to worry about symbol or string key access?
Thoughts?

wbclark: I initially wanted this approach, but decided against it because I wasn't sure about performance characteristics in case `results` or `repo` is particularly large.

If you think it's fine, I'll happily make the change.

parthaa: I like @wbclark 's approach better. Only thing I 'd suggest adding is a Unit test.

sjha4:
Able to enable upstream repos with substitutions now. 
After_PR

Test based Ack 👍🏼

wbclark:
Let's give this a shot...

parthaa (approved):
ACK pending rubocop

PR #10617, opened 2023-06-16
Fixes #36515 - Don't show promote out of order for in-order envs

sjha4:
If the env to be promoted to has prior env checked or already promoted to, even if the prior promoted env is out of order, don't show out of order warning
Scenarios in issue
Consider a promotion like this:
Screenshot from 2023-06-16 13-57-57
Env qa is now considered 'in-order' because version is already published to previous env test (which was out of order, notice missing env dev). We don't have to consider test being out of order anymore. As a result qa is in-order. If you uncheck qa, check prod, prod is still out of order. Hope that makes sense.. :)


jeremylenz (commented):
Thanks @sjha4 

Works as expected!

Would be good to have a test for this new functionality.

jeremylenz (approved):
APJ :+1:

PR #10618, opened 2023-06-18
Fixes #36516 - Add trailing / to CV RPM search autocomplete endpoint

lumarel:

There is a bug in Content views -> <content-view> -> Versions -> compare 2 versions -> RPM packages, the search box, which makes it not show the autocompletion.
I searched a bit in the code and found that the RPM packages part is missing the trailing / compared to the others.
This makes the autocompletion work.

I looked for side-effects, but didn't see any. So it does just this fix.

I setup a forklift devel system with the 3.7/4.8 repos (and this change, on top of KATELLO-4.8), without this change it doesn't show up:
image

As I tested it there, might be a candidate for backporting as well.
Thanks!


jeremylenz:
[test katello]

jeremylenz (approved):
Good catch @lumarel, thanks!

ACK :+1:

lumarel:
Great, thanks!

PR #10622, opened 2023-06-20
Fixes #36288 - Wrong listing of Content Views which contain Files

sjha4:
Add file_id as a param to cv version API as bastion expects the API call to accept the parameter
1. Create and sync file repo.
2. Create some CVs with and some without file repo and publish a few versions.
3. Go to Menu > Content > Files > One of the file units > Content views tab

Before PR:
Views shows all CV versions in the system. No file filter applied.
After PR:
Only CV versions that contain file unit are shown.

### DIFF
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/files/details/views/file-content-views.html
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/files/details/views/file-content-views.html
@@ -24,7 +24,7 @@
       <tbody>
         <tr bst-table-row ng-repeat="version in table.rows">
           <td bst-table-cell>
-            <a ui-sref="content-view.info({contentViewId: version.content_view.id})">
+            <a href="/content_views/{{version.content_view.id}}">
               {{ version.content_view.name }}
             </a>
           </td>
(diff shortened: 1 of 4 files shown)


ianballou (commented):
Looking good! Saw one issue that may be outside the scope of this PR:

The links aren't actually links for the content views:
image

sjha4:
> Looking good! Saw one issue that may be outside the scope of this PR:
> 
> The links aren't actually links for the content views: image

I will update the table. We are returning versions and trying to map them to CV names without versions I think in the tab.

sjha4:
Ended up being a small change. Old code was linking to old CV UI.. 💀 😅

engines/bastion_katello/app/assets/javascripts/bastion_katello/files/details/views/file-content-views.html
```diff
@@ -24,7 +24,7 @@ <h3 translate>Content Views that contain this File</h3>
       <tbody>
         <tr bst-table-row ng-repeat="version in table.rows">
           <td bst-table-cell>
-            <a ui-sref="content-view.info({contentViewId: version.content_view.id})">
+            <a href="/content_views/{{version.content_view.id}}">
```
ianballou, line 27: We just need a statement here that'll link to the products page if the content view ID is the default org one.

ianballou: image

sjha4: Oh..We should probably not be showing Default organization CV at all..Let's just hide it..Thoughts?

sjha4: Updated to return only non_default CV versions.

PR #10625, opened 2023-06-21
Fixes #36221 - Allow removing products with empty URL repos from ACS

sjha4:
Check for presence of ACSSmartProxy before planning delete task
Create a product.
Create a yum repository with URL, sync and everything you'd do normally.
Create another yum repo without a URL.

Create a simplified ACS and add the product to it.
Try removing that product from the ACS. It fails without this PR cause we don't create SmartProxyACS records for empty URL repos.


qcjames53 (approved):
Hi Samir.

I was able to test out the new functionality added in this PR and it works as advertised. The ability to manually remove products with bad URLs was a small oversight in #35358, thank you for fixing. Everything looks good to me.

jeremylenz (approved):
Approving based on @qcjames53, thanks @sjha4!

PR #10626, opened 2023-06-21
Fixes #36526 - Add banners warning about katello-agent removal

jeremylenz:

Add several banners in the web UI warning users about impending katello-agent removal.
Also, I removed the banners on the content host details / Packages and Errata tabs in favor of a single banner at the top of the page.

image
image
image
image

I didn't add a single banner at the top of the new host details page because that would require a Foreman code change.

There were already banners on the Content Host Bulk Actions modals for errata and packages.

Can you think of any other places where I should add a banner?

Downstream, "Katello 4.10" will be translated to "Satellite 6.15".

The PF4 banners have a close button which is nice! but the Angular ones don't, sorry!

In order to see these banners, the UI has to think that you have katello-agent enabled. So, you either have to actually set up katello-agent, or fake it somehow.

### DIFF
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/common/views/katello-agent-notice.html
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/common/views/katello-agent-notice.html
@@ -1,7 +1,7 @@
 <section>
   <p bst-alert="warning" >
     <span translate>
-      Katello-agent is deprecated and will be removed in a future release.
+      Katello-agent is deprecated and will be removed in Katello 4.10.
     </span>
   </p>
 </section>
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/content-hosts.controller.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/content-hosts.controller.js
@@ -15,6 +15,7 @@
  * @requires ContentHostsHelper
  * @requires simpleContentAccessEnabled
  * @requires newHostDetailsUI
+ * @requires BastionConfig
  *
  * @description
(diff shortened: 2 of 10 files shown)


engines/bastion_katello/app/assets/javascripts/bastion_katello/common/views/katello-agent-notice.html
```diff
@@ -1,7 +1,7 @@
 <section>
   <p bst-alert="warning" >
     <span translate>
-      Katello-agent is deprecated and will be removed in a future release.
+      Katello-agent is deprecated and will be removed in Katello 4.10.
```
jeremylenz, line 4: This hard-coding was missed in the last PR, which updated the removal version in the API deprecation warnings only.

jeremylenz:
> do you think we should drop the period at the end?

Nah, it's the period at the end of the sentence!

You said you left 2 questions, but I don't see them?

chris1984:
> Package actions - should we add it here along with the group package deprecation warning message or do you think the notice at the top is enough?

Ah sorry it got mixed up in my review comment: 

Package actions - should we add it here along with the group package deprecation warning message or do you think the notice at the top is enough?

jeremylenz:
> Package actions - should we add it here along with the group package deprecation warning message or do you think the notice at the top is enough?

Ohh, this was the second question!

I actually removed the banner from package actions because there were two of the same banner on the screen and it looked silly.

chris1984:
> > Package actions - should we add it here along with the group package deprecation warning message or do you think the notice at the top is enough?
> 
> Ohh, this was the second question!
> 
> I actually removed the banner from package actions because there were two of the same banner on the screen and it looked silly.

Sounds good, then yeah looks good to me on all the areas I checked, :ship: it

PR #10628, opened 2023-06-23
Fixes #36529 - CV page needs refresh to get the current filters state

sjha4:
1. Reload cv filters on publish screen when wizard is opened to get accurate state of filters on the CV.
2. Needs_publish is now reset if you move from one CV to another.

1. Go to a CV without filters.
2. Go to filters tab. Create a new filter.
3. Hit Publish new version button and go to Review step. You should see the notice that filters will be applied.
4. Publish.
5. Go back to filters tab. Delete the single filter and hit publish.
6. You should not see the warning now.

Small bug fix with needs_publish UI:
Have 2 CVs with some published versions in the box. Make sure none need publish at the moment. (Aka no needs_publish icon on the version)
1. Go to a CV and create a new filter.
2. Go to versions tab and see that needs_publish icon is shown.
3. Go to another CV and make sure the needs_publish icon is not shown.


qcjames53 (approved):
Hi Samir!

I was able to confirm the two new features through testing. I threw a couple different scenarios at the CVs to see if I could break anything but everything held up :)

The source looks good to me. I'm giving this a solid stamp of approval.

PR #10635, opened 2023-06-30
Fixes #36562 - Clear RootRepo checksum if on_demand repo

sayan3296:

Upgrade migration to clear checksum_type from RootRepositories for repos with on_demand download policy, before saving the repos.

In certain cases where users are upgrading their instance from a very old version of katello (at some point they either have done **validate content sync** or due to some other really weird reason), They may have a few on_demand repos left with some checksum_type set. But as this combination should not exist and can fail the [RemoveDrpmFromIgnorableContent](https://github.com/Katello/katello/blob/master/db/migrate/20220228173251_remove_drpm_from_ignorable_content.rb) migration, we want to unset the checksum_type for such repos before executing `root.save!`

* Install Katello 4.4
* Enable a few repos ( having on_demand download policy )
* Use a DB query like this to set a specific checksum_type for them i.e. 
```
echo "update katello_root_repositories set checksum_type = 'sha256' where content_type='yum' ;" | su - postgres -c "psql foreman"
```
* Upgrade to 4.5 and observe the failure during db:migrate
* Apply the changes from this PR
* Re-run db:migrate and it should be successful 
* Re-run the upgrade and that should be completed successfully as well.

### DIFF
--- a/db/migrate/20220228173251_remove_drpm_from_ignorable_content.rb
+++ b/db/migrate/20220228173251_remove_drpm_from_ignorable_content.rb
@@ -6,6 +6,7 @@ class RemoveDrpmFromIgnorableContent < ActiveRecord::Migration[6.0]
       else
         root.ignorable_content = []
       end
+      root.checksum_type = nil if root.download_policy == ::Katello::RootRepository::DOWNLOAD_ON_DEMAND
       root.save!
     end
   end


chris1984 (approved):
ACK works,

```bash
Upgrade Step 2/2: katello:clean_backend_objects. This may take a long while.
0 orphaned consumer id(s) found in candlepin.
Candlepin orphaned consumers: []
  Success!
  * Foreman is running at https://centos8-stream-katello-4-4.area52.example.com
  * To install an additional Foreman proxy on separate machine continue by running:

      foreman-proxy-certs-generate --foreman-proxy-fqdn "$FOREMAN_PROXY" --certs-tar "/root/$FOREMAN_PROXY-certs.tar"
  * Foreman Proxy is running at https://centos8-stream-katello-4-4.area52.example.com:9090
 ```

PR #10638, opened 2023-07-07
Fixes #36563 - Remove orphaned content unit cleanup from index_content

sjha4:
Move remove content unit orphan deletion to async task to avoid increasing indexing times which impacts both syncs and publishes.
1. Sync a repo. You should see the task: "Remove orphaned content units" run after completion of sync or publish CV actions asynchronously.

### DIFF
--- a/app/lib/actions/katello/orphan_cleanup/remove_orphaned_content_units.rb
+++ b/app/lib/actions/katello/orphan_cleanup/remove_orphaned_content_units.rb
@@ -1,0 +1,32 @@
+module Actions
+  module Katello
+    module OrphanCleanup
+      class RemoveOrphanedContentUnits < Actions::Base
+        def plan(options = {})
+          plan_self(id: options[:repo_id], destroy_all: options[:destroy_all])
+        end
+
+        def run
+          content_types_to_index = []
+          if input[:destroy_all]
+            ::Katello::RepositoryTypeManager.enabled_repository_types.each_value do |repo_type|
+              content_types_to_index << repo_type.content_types_to_index
+            end
+          elsif input[:id]
+            repo = ::Katello::Repository.find(input[:id])
+            content_types_to_index = repo.repository_type.content_types_to_index
+          else
+            fail "Pass either a repository to determine content type or destroy_all to destroy all orphaned content units"
+          end
+          content_types_to_index.flatten.each do |type|
+            type.model_class.orphaned.destroy_all
+          end
+        end
+
+        def rescue_strategy
+          Dynflow::Action::Rescue::Skip
+        end
+      end
+    end
+  end
+end
--- a/app/lib/actions/katello/orphan_cleanup/remove_orphaned_content_units.rb
+++ b/app/lib/actions/katello/orphan_cleanup/remove_orphaned_content_units.rb
@@ -2,24 +2,14 @@ module Actions
   module Katello
     module OrphanCleanup
       class RemoveOrphanedContentUnits < Actions::Base
-        def plan(options = {})
-          plan_self(id: options[:repo_id], destroy_all: options[:destroy_all])
-        end
-
         def run
-          content_types_to_index = []
-          if input[:destroy_all]
-            ::Katello::RepositoryTypeManager.enabled_repository_types.each_value do |repo_type|
-              content_types_to_index << repo_type.content_types_to_index
-            end
-          elsif input[:id]
-            repo = ::Katello::Repository.find(input[:id])
-            content_types_to_index = repo.repository_type.content_types_to_index
-          else
-            fail "Pass either a repository to determine content type or destroy_all to destroy all orphaned content units"
+          models = []
+
+          ::Katello::RepositoryTypeManager.enabled_repository_types.each_value do |repo_type|
+            models << repo_type.content_types_to_index
           end
-          content_types_to_index.flatten.each do |type|
-            type.model_class.orphaned.destroy_all
+          models.flatten.each do |content_type|
+            content_type.model_class.orphaned.destroy_all
           end
         end
 
--- a/app/lib/actions/katello/orphan_cleanup/remove_orphans.rb
+++ b/app/lib/actions/katello/orphan_cleanup/remove_orphans.rb
@@ -9,7 +9,7 @@ module Actions
           sequence do
             if proxy.pulp_primary?
               ::Katello::RootRepository.orphaned.destroy_all
-              plan_action(RemoveOrphanedContentUnits, destroy_all: true)
+              plan_action(RemoveOrphanedContentUnits)
             end
             if proxy.pulp3_enabled?
               plan_action(
--- a/app/lib/actions/katello/orphan_cleanup/remove_orphans.rb
+++ b/app/lib/actions/katello/orphan_cleanup/remove_orphans.rb
@@ -7,27 +7,16 @@ module Actions
         end
         def plan(proxy)
           sequence do
+            if proxy.pulp_primary?
+              ::Katello::RootRepository.orphaned.destroy_all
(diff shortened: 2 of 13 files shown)


app/models/katello/concerns/pulp_database_unit.rb
```diff
@@ -169,7 +169,7 @@ def with_identifiers(ids)
 
       def orphaned
         if many_repository_associations
-          where.not(:id => repository_association_class.where(:repository_id => ::Katello::Repository.all).select(unit_id_field))
+          where.not(:id => repository_association_class.select(unit_id_field))
         else
           where.not(:repository_id => ::Katello::Repository.all)
```
parthaa, line 174:
```suggestion
          where.not(:repository_id => ::Katello::Repository.select(:id))
```

parthaa (approved):
Seems to work ok . Checked sync and saw the new action

parthaa (approved):
Tested the remove orphan calls. Behavior looked good. ACK

app/lib/actions/katello/repository/index_content.rb
```diff
@@ -31,6 +31,10 @@ def run
             output[:new_content][content_type.label] = new_count - initial_counts[content_type.label]
           end
         end
+
+        def finalize
+          ForemanTasks.async_task(OrphanCleanup::RemoveOrphanedContentUnits, {repo_id: input[:id]})
```
ianballou, line 36: I'm questioning now if we even need to destroy orphaned units after syncing.

For the orphaned filter rule problem, the solution is this: https://github.com/Katello/katello/blob/05f9de2ae38093f5e30e96eed21882671bab6ea2/app/services/katello/content_unit_indexer.rb#L155

Otherwise, I don't think orphaned content units should cause any other problems.

Plus, if a user syncs 20 repositories, they're going to get 20 orphan cleanup tasks running, which could eat up resources on their system.

Perhaps we should just rely on the weekly orphan cleanup?

If we do that, I just suggest that we change the following error messages to tell the users to run orphan cleanup instead of run a complete sync:

- https://github.com/Katello/katello/blob/master/app/services/katello/pulp3/repository.rb#L500
- https://github.com/Katello/katello/blob/master/app/services/katello/pulp3/repository.rb#L512
- https://github.com/Katello/katello/blob/master/app/services/katello/pulp3/repository/docker.rb#L73

ianballou: Actually, maybe we should change those error messages with your PR here regardless. Complete sync no longer deletes the orphaned units itself.

app/lib/actions/katello/orphan_cleanup/remove_orphaned_content_units.rb
```diff
@@ -0,0 +1,32 @@
+module Actions
+  module Katello
+    module OrphanCleanup
+      class RemoveOrphanedContentUnits < Actions::Base
+        def plan(options = {})
+          plan_self(id: options[:repo_id], destroy_all: options[:destroy_all])
+        end
+
+        def run
+          content_types_to_index = []
```
ianballou, line 10: `content_types_to_index` makes it sound a bit like the types will be indexed instead of checked for orphans.

app/lib/actions/katello/orphan_cleanup/remove_orphaned_content_units.rb
```diff
@@ -0,0 +1,32 @@
...
+          if input[:destroy_all]
+            ::Katello::RepositoryTypeManager.enabled_repository_types.each_value do |repo_type|
+              content_types_to_index << repo_type.content_types_to_index
+            end
+          elsif input[:id]
+            repo = ::Katello::Repository.find(input[:id])
+            content_types_to_index = repo.repository_type.content_types_to_index
+          else
+            fail "Pass either a repository to determine content type or destroy_all to destroy all orphaned content units"
```
ianballou, line 19: What do you think about removing the `destroy_all` param?  That way only `repo_id` would be needed and the logic would be a bit simplified. Orphan cleanup should only be deleting useless data, so I don't think a safety net is needed.

sjha4:
Updated the PR to run orphan content unit deletion only in orphan cleanup task which should run once a week in deployed environments.
Removed inputs from task to make it run on all content units during the orphan cleanup.
Updated error messages to point to the rake task.

jeremylenz:
PR title needs an update now ;)

sjha4:
Should be ready for re-review.

app/services/katello/pulp3/repository.rb
```diff
@@ -497,7 +497,8 @@ def add_content(content_unit_href, remove_all_units = false)
       rescue api.client_module::ApiError => e
         if e.message.include? 'Could not find the following content units'
           raise ::Katello::Errors::Pulp3Error, "Content units that do not exist in Pulp were requested to be copied."\
-            " Please run a complete sync on the following repository: #{repository_reference.root_repository.name}. Original error: #{e.message}"
+             " Please run `foreman-rake katello:delete_orphaned_content` to fix the following repository:"\
+            " #{::Katello::Pulp3::RepositoryReference.find_by(repository_href: repository_reference.repository_href).root_repository.name}. Original error: #{e.message}"
```
ianballou, line 501:
```suggestion
            " #{repository_reference.root_repository.name}. Original error: #{e.message}"
```

ianballou: The `repository_reference` can be used here since the method is adding content to the service class's own repo.

app/lib/actions/katello/orphan_cleanup/remove_orphans.rb
```diff
@@ -7,27 +7,16 @@ class RemoveOrphans < Actions::Base
         end
         def plan(proxy)
           sequence do
+            if proxy.pulp_primary?
+              ::Katello::RootRepository.orphaned.destroy_all
```
ianballou, line 11: Any reason why this was pulled out of the run phase?

sjha4: Not really..This was the only operation in the run phase after refactoring the content unit deletion..and this doesn't seem like something that needs to be in the run phase..

jeremylenz: If there's a DB operation, should try not to put it in the run phase because `run` doesn't run in a transaction :)

ianballou: Would a heavy DB operation in the planning phase eat up a Puma worker? Also we shouldn't need transactions here at least since we're deleting orphans.  But even if we did need one, couldn't we do a `transaction` block?

ianballou (approved):
Works for me!

app/lib/actions/katello/orphan_cleanup/remove_orphans.rb
jeremylenz: > I'm a little confused because we stick heavy DB ops in tasks in other places

Usually they're in `plan` or `finalize` though, right? Those are fine because they run in transactions by default.

sjha4: Ack to that..We tend to put only things that need API calls to other products in run phase generally. For this task, I'd be ok keeping this in run phase over the plan phase. Any failures on this are set to rescue=skip and also we do want to still plan other pulp orphan deletion actions even if the orphan deletion in katello fails. A plan phase failure wouldn't plan those pulp tasks. 

We could put this in finalize if we want. Let me know if we feel strongly about that and I can update this.

jeremylenz: Makes sense to me. I trust your judgment to choose the phase correctly :)

PR #10640, opened 2023-07-11
Fixes #36577 - Filter gets applied to all the repository upon removal

sjha4:
1. Create a couple of repos and add them to filters.
2. Go to repo > Remove repository
3. You should see a section in the deletion confirmation modal if the repo is the last repo that a filter is applied to.
4. You should be able to choose to delete the filter/or make it apply to all repositories in the CV depending on your selection.
5. This section shouldn't appear for repos that are not the last affected repo in any filters.

### DIFF
--- a/app/controllers/katello/api/v2/repositories_controller.rb
+++ b/app/controllers/katello/api/v2/repositories_controller.rb
@@ -416,9 +416,12 @@ module Katello
     api :DELETE, "/repositories/:id", N_("Destroy a custom repository")
     param :id, :number, :required => true
     param :remove_from_content_view_versions, :bool, :required => false, :desc => N_("Force delete the repository by removing it from all content view versions")
+    param :delete_empty_repo_filters, :bool, :required => false, :desc => N_("Delete content view filters that have this repository as the last associated repository. Defaults to true. If false, such filters will now apply to all repositories in the content view.")
     def destroy
       sync_task(::Actions::Katello::Repository::Destroy, @repository,
-        remove_from_content_view_versions: ::Foreman::Cast.to_bool(params.fetch(:remove_from_content_view_versions, false)))
+                remove_from_content_view_versions: ::Foreman::Cast.to_bool(params.fetch(:remove_from_content_view_versions, false)),
+                delete_empty_repo_filters: ::Foreman::Cast.to_bool(params.fetch(:delete_empty_repo_filters, true))
+                )
       respond_for_destroy
     end
 
--- a/app/lib/actions/katello/product/destroy.rb
+++ b/app/lib/actions/katello/product/destroy.rb
@@ -2,21 +2,11 @@ module Actions
   module Katello
     module Product
       class Destroy < Actions::EntryAction
-        # rubocop:disable Metrics/MethodLength
         def plan(product, options = {})
           organization_destroy = options.fetch(:organization_destroy, false)
           skip_environment_update = options.fetch(:skip_environment_update, false) ||
               options.fetch(:organization_destroy, false)
-
-          unless organization_destroy || product.user_deletable?
-            if product.redhat?
-              fail _("Cannot delete Red Hat product: %{product}") % { :product => product.name }
-            elsif !product.published_content_view_versions.not_ignorable.empty?
-              fail _("Cannot delete product with repositories published in a content view.  Product: %{product}, %{view_versions}") %
-                       { :product => product.name, :view_versions => view_versions(product) }
-            end
-          end
-
+          check_ready_to_delete(product, organization_destroy)
           action_subject(product)
 
           # Candlepin::Product::ContentRemove is called with Katello::Repository::Destroy, so we only want to run ContentRemove
@@ -98,6 +88,20 @@ module Actions
             _("Content View %{view}: Versions: %{versions}") % {:view => view, :versions => versions.join(', ')}
           end
           results.join(', ')
+        end
+
+        def check_ready_to_delete(product, organization_destroy)
+          unless organization_destroy || product.user_deletable?
+            if product.redhat?
+              fail _("Cannot delete Red Hat product: %{product}") % { :product => product.name }
+            elsif !product.published_content_view_versions.not_ignorable.empty?
+              fail _("Cannot delete product with repositories published in a content view.  Product: %{product}, %{view_versions}") %
+                     { :product => product.name, :view_versions => view_versions(product) }
+            elsif product.repositories.any? { |repo| repo.filters.any? { |filter| filter.repositories.size == 1 } }
+              fail _("Cannot delete product: %{product} with repositories that are the last affected repository in content view filters. Delete these repositories before deleting product.") %
+                     { :product => product.name }
+            end
+          end
         end
       end
     end
--- a/app/lib/actions/katello/repository/destroy.rb
+++ b/app/lib/actions/katello/repository/destroy.rb
@@ -15,6 +15,7 @@ module Actions
               options.fetch(:organization_destroy, false)
           destroy_content = options.fetch(:destroy_content, true)
           remove_from_content_view_versions = options.fetch(:remove_from_content_view_versions, false)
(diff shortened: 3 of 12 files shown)


app/controllers/katello/api/v2/repositories_controller.rb
```diff
@@ -416,9 +416,12 @@ def update
     api :DELETE, "/repositories/:id", N_("Destroy a custom repository")
     param :id, :number, :required => true
     param :remove_from_content_view_versions, :bool, :required => false, :desc => N_("Force delete the repository by removing it from all content view versions")
+    param :delete_orphaned_filters, :bool, :required => false, :desc => N_("Delete filters that have this repository as the last associated repository. Such filters will apply to all repositories in CV otherwise. Defaults to true.")
```
jeremylenz, line 419:
```suggestion
    param :delete_orphaned_filters, :bool, :required => false, :desc => N_("Delete content view filters that have this repository as the last associated repository. Defaults to true. If false, such filters will now apply to all repositories in the content view. ")
```

app/lib/actions/katello/repository/destroy.rb
```diff
@@ -81,6 +83,11 @@ def handle_acs_product_removal(repository)
           end
         end
 
+        def delete_orphaned_filters(repository)
```
jeremylenz, line 86:
```suggestion
        def delete_empty_filters(repository)
```

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/views/repository-details.html
```diff
@@ -127,13 +127,60 @@ <h2 translate>{{ repository.name }}</h2>
...
+                   ng-model="repositoryWrapper.forceDeleteCV"
+                   ng-change="updateForceDeleteCV()"/>
+            <span translate
+                  ng-show="repositoryWrapper.repository.content_view_versions.length !== 0"
+                  style="margin-bottom: 2em"
+            >
+                Are you sure you want to remove repository {{ repositoryWrapper.repository.name }} from all content views?
+            </span>
+            <div ng-show="repositoryWrapper.showLastFilterDeletion" style="margin-top: 1em; margin-bottom: 1em">
+                <b><span translate> Below filters have this repository as the last affected repository!</span></b>
```
jeremylenz, line 139:
```suggestion
                <b><span translate> The filters below have this repository as the last affected repository!</span></b>
```

jeremylenz (commented):
just a few wording suggestions :)

app/lib/actions/katello/repository/destroy.rb
sjha4: Both don't accurately reflect which filters are being deleted. Thoughts on this:  `delete_empty_repo_filters` ? Also we should change the name of param of controller if we change this since that is user facing.

engines/bastion_katello/test/products/details/repositories/details/repository-details.controller.test.js
```diff
@@ -202,4 +210,63 @@ describe('Controller: RepositoryDetailsController', function() {
               }]
           });
   });
+
+    it('should provide a way to view orphaned filters on a repository', function() {
```
jeremylenz, line 214:
```suggestion
    it('should provide a way to view empty-repo filters on a repository', function() {
```

jeremylenz (commented):
When clicking Remove Repository from the repo details screen / Select Action,
:heavy_check_mark: I see the section added to the modal when it's the last repo
:heavy_check_mark: I don't see the section added to the modal when it's not the last repo

One question.. When selecting a repo from the repo list screen and clicking Remove Repositories, I don't see anything special. Is this intended?

image

jeremylenz:
> For filters we would default to deleting the filter.

Have we always deleted the filter in this case? Or this PR changes that behavior?

If it's a change, should probably add some notice to that modal.

sjha4:
> > For filters we would default to deleting the filter.
> 
> Have we always deleted the filter in this case? Or this PR changes that behavior?
> 
> If it's a change, should probably add some notice to that modal.

We do change the behavior. Earlier it would make the filter apply to all repos..

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/views/product-repositories.html
```diff
@@ -45,7 +45,10 @@
       <div data-block="modal-body"
            translate
            translate-n="table.numSelected"
-           translate-plural="Are you sure you want to remove {{ table.getSelected().length }} repositories?">
+           translate-plural="Repositories that are published in a content view
+                             versions or are the last affected repository in a content view filter will be skipped. You can delete those from the repository details page.
```
jeremylenz, line 49:
```suggestion
           translate-plural="Repositories that are published in content view
                             versions or are the last affected repository in a content view filter will be skipped. You can delete those from the repository details page.
```

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/views/product-details.html
```diff
@@ -46,6 +46,11 @@ <h2 translate>{{ product.name }}</h2>
                  tooltip-animation="false"
                  tooltip-append-to-body="true">
               </i>
+              <i class="fa fa-question-circle" ng-switch-when="last_affected_repo_on_filter"
+                 uib-tooltip="{{ 'You cannot remove this product because it has repositories that are last affected repository on content view filters' | translate }}"
```
jeremylenz, line 50:
```suggestion
                 uib-tooltip="{{ 'You cannot remove this product because it has repositories that are the last affected repository on content view filters' | translate }}"
```

app/lib/actions/katello/product/destroy.rb
```diff
@@ -99,6 +89,20 @@ def view_versions(product)
...
+
+        def check_ready_to_delete(product, organization_destroy)
+          unless organization_destroy || product.user_deletable?
+            if product.redhat?
+              fail _("Cannot delete Red Hat product: %{product}") % { :product => product.name }
+            elsif !product.published_content_view_versions.not_ignorable.empty?
+              fail _("Cannot delete product with repositories published in a content view.  Product: %{product}, %{view_versions}") %
+                     { :product => product.name, :view_versions => view_versions(product) }
+            elsif product.repositories.any? { |repo| repo.filters.any? { |filter| filter.repositories.size == 1 } }
+              fail _("Cannot delete product: %{product} with repositories that are the last affected repository in content view filters. Delete these repositories before deleting product.") %
```
jeremylenz, line 101: Does it have to be this way? Can't we just default to deleting the filters like we do with individual repo deletes?

jeremylenz (commented):
I was expecting a different message on both the bulk product delete confirmation and the bulk repository delete confirmation modals, but I don't see them.

See my [question/suggestion below](https://github.com/Katello/katello/pull/10640#discussion_r1265639501) about auto-deleting filters during bulk delete. Maybe we should change the behavior as per that, then update these modals to describe it. Thoughts?

app/lib/actions/katello/product/destroy.rb
sjha4: This would be the better behavior in case someone wanted to retain their filters etc. This is also the behavior for repos in published CVs so it makes sense to me to follow the same user flow. Deleting filters would be a non-reversible action and affect content in CVs, probably something user should have to look at and decide per CV.

jeremylenz: ahh, gotcha! Ok, just fix the typos then and I'm good :)

engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/views/product-details.html
sjha4: Updated!

jeremylenz (approved):
thanks @sjha4!

ACK :+1:

PR #10645, opened 2023-07-12
Fixes #36586 - pin ostree bindings below 2.1.1

ianballou:
Pins the ostree client version to before 2.1.0.

As of 2.1.0, there is a new hidden distribution attribute that is breaking the VCRs.

Re-recording the VCRs to fix this would be incorrect because Katello's packaged version of pulp ostree likely does not support the hidden distribution feature.

See the tests pass.

### DIFF
--- a/katello.gemspec
+++ b/katello.gemspec
@@ -60,7 +60,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency "pulp_rpm_client", ">= 3.19.0", "< 3.20.0"
   gem.add_dependency "pulp_certguard_client", "< 2.0"
   gem.add_dependency "pulp_python_client", ">= 3.8.0", "< 3.9"
-  gem.add_dependency "pulp_ostree_client"
+  gem.add_dependency "pulp_ostree_client", "< 2.1.1"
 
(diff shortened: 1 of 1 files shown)


jeremylenz (approved):
APJ :+1:

sjha4:
:ship: 🚢

sjha4:
So this http://koji.katello.org/releases/yum/katello-nightly/katello/el8/x86_64/rubygem-pulp_ostree_client-2.1.0-1.el8.noarch.rpm says 2.1.0 is packaged for nightly..Not sure if that conflicts with < 2.1.0

ianballou:
> So this http://koji.katello.org/releases/yum/katello-nightly/katello/el8/x86_64/rubygem-pulp_ostree_client-2.1.0-1.el8.noarch.rpm says 2.1.0 is packaged for nightly..Not sure if that conflicts with < 2.1.0

@sjha4 we shouldn't be packaging that  version of the ostree client yet, I'll have to ask to have it pulled out (or at least make sure we still have the older versions).

chris1984:
🟢

PR #10646, opened 2023-07-13
Fixes #36589 - Added ability to refresh all alternate content sources…

qcjames53:
… using API POST call at /alternate_content_sources/bulk/refresh_all

Sending a POST request to `http://centos8-katello-devel.redhat.example.com/katello/api/alternate_content_sources/bulk/refresh_all` will refresh all alternate content sources. This is the first half of the implementation of #36589 (adding refresh all support to hammer); the remaining changes will need to be made on hammer's end to support bulk actions.

n/a

Create a Katello instance with no alternate content sources. Run the following command to request to refresh all ACS's:
```
curl --request POST --user admin:changeme -H "Content-type: application/json" http://<host>/katello/api/alternate_content_sources/bulk/refresh_all

```
Since there are no ACS's, the command should fail:
```
{"displayMessage":"Unable to refresh any alternate content source. You either do not have the permission to refresh, or no alternate content sources exist.","errors":["Unable to refresh any alternate content source. You either do not have the permission to refresh, or no alternate content sources exist."]}
```
Now add a few alternate content sources and monitor the last refreshed time (I used the web UI). Run the same command as before:
```
curl --request POST --user admin:changeme -H "Content-type: application/json" http://<host>/katello/api/alternate_content_sources/bulk/refresh_all
```
The alternate content sources should all have been refreshed.

### DIFF
--- a/app/controllers/katello/api/v2/alternate_content_sources_bulk_actions_controller.rb
+++ b/app/controllers/katello/api/v2/alternate_content_sources_bulk_actions_controller.rb
@@ -1,6 +1,6 @@
 module Katello
   class Api::V2::AlternateContentSourcesBulkActionsController < Api::V2::ApiController
-    before_action :find_alternate_content_sources
+    before_action :find_alternate_content_sources, except: [:refresh_all_alternate_content_sources]
 
     api :PUT, '/alternate_content_sources/bulk/destroy', N_('Destroy one or more alternate content sources')
     param :ids, Array, desc: N_('List of alternate content source IDs'), required: true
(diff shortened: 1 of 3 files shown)


app/controllers/katello/api/v2/alternate_content_sources_bulk_actions_controller.rb
```diff
@@ -34,6 +34,21 @@ def refresh_alternate_content_sources
       end
     end
 
+    api :POST, '/alternate_content_sources/bulk/refresh_all', N_("Refresh all alternate content sources")
+    def refresh_all_alternate_content_sources
+      refreshable_alternate_content_sources = AlternateContentSource.readable.editable
```
sjha4, line 39: I think editable should be enough for permissions here. Editable but not readable doesn't make much sense..I don't know how that would work..

qcjames53: Added in your suggested changes, thank you.

sjha4:
Code looks good..Will do some testing.. 👍🏼

sjha4 (approved):
Works great. Tested with hammer PR: https://github.com/Katello/hammer-cli-katello/pull/902 and both worked as advertised. Ack. Please merge after @ofedoren acks the hammer PR.

PR #10647, opened 2023-07-13
Fixes #36590 - fix ACS error message for duplicate words

lfu:
1. Add new simplified ACS
2. Have an active product that does not have a file repository and try to add it to a new ACS (i.e. RHEL)

### DIFF
--- a/app/lib/katello/validators/alternate_content_source_products_validator.rb
+++ b/app/lib/katello/validators/alternate_content_source_products_validator.rb
@@ -7,7 +7,7 @@ module Katello
             product = ::Katello::Product.find(value)
             content_type = record.alternate_content_source.content_type
             if product.repositories.with_type(content_type).has_url.empty?
-              record.errors[attribute] << N_("The product %{name} has no %{type} repositories with upstream URLs to add to the alternate content source.") % { name: product.name, type: content_type }
+              record.errors.add(attribute, _("%{name} has no %{type} repositories with upstream URLs to add to the alternate content source.") % { name: product.name, type: content_type })
             end
           end
         end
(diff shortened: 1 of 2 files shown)


app/lib/katello/validators/alternate_content_source_products_validator.rb
```diff
@@ -7,7 +7,7 @@ def validate_each(record, attribute, value)
             product = ::Katello::Product.find(value)
             content_type = record.alternate_content_source.content_type
             if product.repositories.with_type(content_type).has_url.empty?
-              record.errors[attribute] << N_("The product %{name} has no %{type} repositories with upstream URLs to add to the alternate content source.") % { name: product.name, type: content_type }
+              record.errors.add(attribute, N_("%{name} has no %{type} repositories with upstream URLs to add to the alternate content source.") % { name: product.name, type: content_type })
```
lfu, line 10: Do you think this should be `N_` or `_`?

sjha4: Not entirely sure, but documentation suggests N_ for dynamic translations..
https://github.com/grosser/gettext_i18n_rails#unfound-translations-with-rake-gettextfind

lfu: Just tested it out. 

With `N_`
Screenshot at 2023-07-13 17-11-28

With `_`
Screenshot at 2023-07-13 17-14-12

sjha4 (approved):
Ack. Can see updated error message:
Before chnage:
Screenshot from 2023-07-14 16-40-52
---
After change:
Screenshot from 2023-07-14 16-38-08

PR #10648, opened 2023-07-13
Refs #36540 - Removed the unnecessary enablement migrator in master

parthaa:
This commit tries to address the issues listed [here](https://community.theforeman.org/t/katello-4-9-custom-products-changes-with-sca/33763/10)

PR for Issue #36120 changed default enablement for custom repos. By default all custom repos got disabled in candlepin. However a migration task was provided to add overrides for existing custom repos.

The enablement migrator introduced in that PR is no longer necessary for 4.10 (which points to the correct master) so removing it.

We basically want say Default Enablement Migrator must only run for anything upto 4.9 but not after. 

- 4.9 -> 4.9.1 upgrade migration is handled [here](https://github.com/Katello/katello/pull/10644)
- 4.9 => 4.10 upgrade is a addressed by this PR.

- `bundle exec rake katello:upgrades:4.9:update_custom_products_enablement` should not change the enablement of anything. 
- Should error with no such task :).


jeremylenz (approved):
APJ :+1:

PR #10649, opened 2023-07-14
Fixes #36580 - Content view dependency solving should be tracked for needs_publish

sjha4:
Track changes to solve_dependency on CV for needs_publish determination.
1. Create and publish a CV.
2. Update the CV's solve dependency from details tab.
3. Go back to versions tab and confirm that last version shows needs_publish as true.
4. Switch CV's dependency solving to original value.
5. Go back to versions tab and confirm that last version shows needs_publish as false.


jeremylenz (approved):
Works for me!

ACK :+1:

PR #10651, opened 2023-07-14
Fixes #36520 - Enforce authorization on capsule syncs

sjha4:
Only sync capsules users have permissions on.
Add scoping and permissions to smart proxies that can be synced by a user.
User Setup: Create a user created for API use only, selected role is only "sync_repositories"

Steps to reproduce:
Push Content into Repo via cmd `hammer repository upload-content --organization <org> --product <product> --name <reponame> --path <directory_path_to_rpms>`
The capsule content sync task in dynflow triggered by this will error out.

With this PR: 
The capsule content sync task will only start on capsules that user has permissions on.

### DIFF
--- a/app/lib/actions/katello/content_view/capsule_sync.rb
+++ b/app/lib/actions/katello/content_view/capsule_sync.rb
@@ -9,7 +9,7 @@ module Actions
         def plan(content_view, environment)
           sequence do
             concurrence do
-              smart_proxies = SmartProxy.with_environment(environment)
+              smart_proxies = SmartProxy.unscoped.with_environment(environment).select { |sp| sp.authorized?(:manage_capsule_content) && sp.authorized?(:view_capsule_content) }
               unless smart_proxies.blank?
                 plan_action(::Actions::BulkAction, ::Actions::Katello::CapsuleContent::Sync, smart_proxies.sort,
                             :content_view_id => content_view.id, :environment_id => environment.id)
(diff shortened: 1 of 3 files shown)


app/lib/actions/katello/content_view/capsule_sync.rb
```diff
@@ -9,7 +9,7 @@ def humanized_name
         def plan(content_view, environment)
           sequence do
             concurrence do
-              smart_proxies = SmartProxy.with_environment(environment)
+              smart_proxies = SmartProxy.unscoped.authorized(:view_capsule_content).with_environment(environment)
```
ianballou, line 12: I think these should all be:

```suggestion
              smart_proxies = SmartProxy.unscoped.authorized(:manage_capsule_content).with_environment(environment)
```

sjha4: That specifically didn't work so I had to tweak it a little while using manage_capsule_content..

app/lib/actions/katello/content_view/capsule_sync.rb
```diff
@@ -9,7 +9,7 @@ def humanized_name
         def plan(content_view, environment)
           sequence do
             concurrence do
-              smart_proxies = SmartProxy.with_environment(environment)
+              smart_proxies = SmartProxy.unscoped.with_environment(environment).select { |sp| sp.authorized?(:manage_capsule_content) }
```
ianballou, line 12: Looks like users also need `view_capsule_content`. I couldn't sync with just `manage..`

ianballou (approved):
Looks good!

sjha4:
[test katello]

PR #10652, opened 2023-07-17
Fixes #36597 - Rename 'Change content source' job template

jeremylenz:

The user was confused because running the job template on the host didn't change the host's content source. In actuality, it updates subscription-manager configuration on the host based on the host's _current_ content source. It's meant to only be run just after changing the content source on the server, as part of the 'Change Content Source' feature.

So, the changes here are

1. Change the name of the job template `Change content source` to `Configure host for new content source`. This reflects what the job template actually does. Includes a DB migration similar to https://github.com/theforeman/foreman/blob/develop/db/migrate/20220204155632_rename_job_report_template.rb (thanks for the tip @ianballou)
2. Change the name of the method `change_content_source` to `configure_host_for_new_content_source`
3. Rename files accordingly
4. Add (hopefully) helpful comments to the job template

I didn't update the name of the remote execution feature, since that would require a code change in foreman_remote_execution.

Run migrations:
```
bundle exec rails db:migrate
```

Run seeds and restart the server:
```
bundle exec rails runner "ForemanInternal.all.first.destroy" && bundle exec foreman start
```

Test the following:

1. In Hosts > Job Templates, verify that the "Change content source" template is gone and the new "Configure host for new content source" template is there.
2. test changing a host's content source and verify everything still works I guess


jeremylenz:
I changed the ERB comments to regular Bash comments.

See https://github.com/theforeman/foreman_remote_execution/pull/822

ianballou (commented):
Code is looking fine, now I just need to actually test it.

ianballou (approved):
Working for me! I tested the case of having a same-named job template already and it was backed up properly. New template works as well.

PR #10654, opened 2023-07-17
Fixes #36599 - Add hostname to hypervisor task with validation error

chris1984:

* In `Actions::Katello::Host::Hypervisors` returned the name of the hypervisor hostname reporting in instead of just `hostname`
* Returned an error like Foreman's validation citing the RFCs around hostnames
* Fixed a test that was failing because of my added validation

* Wrote the `create_host_for_hypervisor` method in a `begin/rescue` statement so that if the host does not have any issues we just move along like normal.

* Check out PR
* Install `virt-who` with `dnf`
* Create a `virt-who` config file in `/etc/virt-who.d/` I called mine `fake.conf` Here is what mine looks like:
```bash
[fakevirt]
type=fake
file=/tmp/fake.json
is_hypervisor=True
owner=Default_Organization
hypervisor_id=$&
```

* Now in `/tmp` create a `fake.json` that looks like this
```json
{
        "hypervisors": [{
                "uuid": "$&",
                "guests": [{
                                "guestId": "a8f73be7-c6a9-4c2a-8590-0ec2457c975d",
                                "state": 1
                        },
                        {
                                "guestId": "3e62be70-2365-4b53-8d2c-84c89cc1373a",
                                "state": 1
                        },
                        {
                                "guestId": "a746244b-125a-4f09-80bd-c043413a9153",
                                "state": 1
                        },
                        {       "guestId": "b5e07903-f5d8-44e6-b2dd-922743c402e2",
                                "state": 1
                        },
                        {       "guestId": "be263884-7c5c-4d02-a268-0cfadef3c6b4",
                                "state": 1
                        },
                        {       "guestId": "da172ae4-1b82-4900-975b-17ed24553f09",
                                "state": 1
                        },
                        {       "guestId": "ecf50029-a084-4c0e-813b-51f6b977f2f1",
                                "state": 1
                        }
                ]
        }]
}
```
* Register your system running virt-who to your devel box ( I just used my own devel box to register to itself)
* Start up `virt-who` 
  * `systemctl start virt-who`
* Check in the production log to make sure you see the new error `InvalidVirtWhoHost` with the hostname `virt-who-xxx` like so
`ESC[32m2023-07-12T20:51:33ESC[0m [ESC[31mEESC[0m|ESC[36mbacESC[0m|7f084d2d] The hostname virt-who-$&-1 can contain only lowercase letters, numbers, dashes and dots according to RFC921, RFC952 and RFC1123 (Katello::Errors::InvalidVirtWhoHost)`

### DIFF
--- a/app/models/katello/concerns/host_managed_extensions.rb
+++ b/app/models/katello/concerns/host_managed_extensions.rb
@@ -121,6 +121,8 @@ module Katello
 
         has_many :hypervisor_pools, :class_name => '::Katello::Pool', :foreign_key => :hypervisor_id, :dependent => :nullify
 
+        validates :name, format: { with: Net::Validations::HOST_REGEXP, message: _("%{value} can contain only lowercase letters, numbers, dashes and dots.") }
+
         before_validation :correct_kickstart_repository
         before_update :check_host_registration, :if => proc { organization_id_changed? }
 
--- a/test/controllers/api/v2/hosts_controller_test.rb
+++ b/test/controllers/api/v2/hosts_controller_test.rb
@@ -106,7 +106,7 @@ class Api::V2::HostsControllerTest < ActionController::TestCase
   def test_create_with_permitted_attributes
     cf_attrs = {:content_view_id => @content_view.id, :lifecycle_environment_id => @environment.id}
     sf_attrs = {:purpose_addons => ["Addon"]}
-    attrs = @host.clone.attributes.merge("name" => "contenthost", "content_facet_attributes" => cf_attrs, "subscription_facet_attributes" => sf_attrs).compact!
+    attrs = @host.clone.attributes.merge("name" => "contenthost.example.com", "content_facet_attributes" => cf_attrs, "subscription_facet_attributes" => sf_attrs).compact!
 
     assert_difference('Host.unscoped.count') do
       post :create, params: attrs
@@ -119,7 +119,7 @@ class Api::V2::HostsControllerTest < ActionController::TestCase
                 :lifecycle_environment_id => @environment.id,
                 :uuid => "thisshouldntbeabletobesetbyuser"
                }
-    attrs = @host.clone.attributes.merge("name" => "contenthost1", "content_facet_attributes" => cf_attrs).compact!
+    attrs = @host.clone.attributes.merge("name" => "contenthost1.example.com", "content_facet_attributes" => cf_attrs).compact!
 
     post :create, params: attrs
     assert_response :success # the uuid is simply filtered out which allows the host to be still saved
--- a/test/controllers/foreman/hosts_controller_test.rb
+++ b/test/controllers/foreman/hosts_controller_test.rb
@@ -53,7 +53,7 @@ class HostsControllerTest < ActionController::TestCase
   test 'empty content facet parameters are removed' do
     orig_cves = @host.content_facet.content_view_environment_ids.to_a
     post :create, params: { :host => {
-      :name => 'test_content',
+      :name => 'test-content',
       :content_facet_attributes => {
         :lifecycle_environment_id => "",
         :content_source_id => ""
--- a/test/models/host/content_facet_test.rb
+++ b/test/models/host/content_facet_test.rb
@@ -6,7 +6,7 @@ module Katello
     let(:dev) { katello_environments(:dev) }
     let(:view)  { katello_content_views(:library_dev_view) }
     let(:environment) { katello_environments(:library) }
-    let(:empty_host) { ::Host::Managed.create!(:name => 'foobar', :managed => false) }
+    let(:empty_host) { ::Host::Managed.create!(:name => 'foobar.example.com', :managed => false) }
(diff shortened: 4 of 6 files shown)


app/lib/actions/katello/host/hypervisors_update.rb
```diff
@@ -142,10 +142,16 @@ def duplicate_name(hypervisor, consumer)
...
-                                     :location => location, :managed => false, :enabled => false)
-          host.save!
-          host
+          begin
+            host = ::Host::Managed.new(:name => name, :organization => organization,
+                                       :location => location, :managed => false, :enabled => false)
+            host.save!
+          rescue
+            # Taken from Foreman's active record error message
+            raise ::Katello::Errors::InvalidVirtWhoHost, "The hostname #{name} can contain only lowercase letters, numbers, dashes and dots according to RFC921, RFC952 and RFC1123"
```
jeremylenz, line 151: This should probably be a regular Rails validation, not a rescue.

Check out `app/models/katello/concerns/organization_extensions.rb`, see how it has an included block and then has some validations added to the Organization model?

```rb
included do
...
validates_with Validators::KatelloLabelFormatValidator, :attributes => :label
validates :label, :uniqueness => true
```

In this case we could do the same thing, except add it to `host_managed_extensions`. That would allow us to add validations to `Host::Managed`.

As for the validation itself, you can look at

https://github.com/Katello/katello/blob/master/app/models/katello/root_repository.rb#L88

and

https://github.com/Katello/katello/blob/master/app/models/katello/root_repository.rb#L182C5-L187C8

and write something similar to that.

jeremylenz (changes requested):
See comment below :)

Also, I think the original BZ asked for the hypervisor hostname and ID to be included in the error, but I don't see it added here?

chris1984:
@jeremylenz let me know if this looks better? 

I am still getting the desired result:
```bash
18:06:48 rails.1   | 2023-07-20T18:06:48 [E|bac|92627384] The hostname virt-who-$&-1 can contain only lowercase letters, numbers, dashes and dots
18:06:48 rails.1   |  92627384 |                                                         according to RFC921, RFC952 and RFC1123 (Katello::Errors::InvalidVirtWhoHost)
18:06:48 rails.1   |  92627384 | /home/vagrant/katello/app/models/katello/concerns/host_managed_extensions.rb:210:in `ensure_no_special_chars'
```

jeremylenz:
> let me know if this looks better?

A step in the right direction!

But should use `errors.add`, like in https://github.com/Katello/katello/blob/master/app/models/katello/root_repository.rb#L182C5-L187C8

This way you won't have to create a new error class (which isn't standard for validations), and it also may help you get around the long line length, since you can put a newline after each argument if you need to.

chris1984:
@jeremylenz updated to using the `errors.add` method and shortened the error message so it's not a 1:1 dupe of the validation from Foreman.

chris1984:
@jeremylenz updated to using the `errors.add` method and shortend the error message so it's not a 1:1 dupe of the validation from Foreman.

app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -202,6 +204,13 @@ def reset_katello_status
         true
       end
 
+      def ensure_no_special_chars
+        hostname = self.name
+        unless hostname.scan(/[!@#$%^&*()_+{}\[\]:;'"\/\\?><,]/).empty?
```
jeremylenz, line 209: This regex seems to allow capital letters and spaces.

I looked on the internet and found 
```suggestion
        unless hostname.match?(/(?=^.{4,253}$)(^((?!-)[a-zA-Z0-9-]{1,63}(?<!-)\.)+[a-zA-Z]{2,63}$)/)
```

https://stackoverflow.com/questions/11809631/fully-qualified-domain-name-validation

Thoughts?

jeremylenz (commented):
Validation code looks better now, thanks!

One more comment/question and then I'll leave it to another reviewer to do a functional test.

chris1984:
@jeremylenz updated

chris1984:
@jeremylenz the tests are failing with the updated code, since this is not important (set to a target milestone) and I am not good with regex, just going to close this and put it on overflow for someone else to pickup.

chris1984:
@jeremylenz how does this look? I still have a few places I need to update tests, but does it look fine for APJ? @sjha4 signed up to test it

lfu:
Got error message with hostname displayed.

10:06:40 rails.1   | 2023-08-04T10:06:40 [E|bac|b157dd57] Validation failed: Name hostname can contain only lowercase letters, numbers, dashes and dots according to RFC921, RFC952 and RFC1123, Name hostname can contain only lowercase letters, numbers, dashes and dots according to RFC921, RFC952 and RFC1123, Name virt-who-$&-1 can contain only lowercase letters, numbers, dashes and dots. (ActiveRecord::RecordInvalid)

lfu (approved):
LGTM 👍🏻

PR #10657, opened 2023-07-18
Fixes #36604 - Manifest delete confirmation should be SCA-aware

jeremylenz:

In the Delete Manifest confirmation modal, most of the words are no longer relevant if you're using SCA. Update the wording of the `<DeleteManifestModalText>` to remove outdated wording.

Subscriptions > Manage Manifest > Delete > look at the wording for an SCA org vs. a non-SCA org with manifest imported.


parthaa (approved):
Works well.

PR #10663, opened 2023-07-24
Fixes #36610 - Add system purpose card to new Activation Key details page

trev-allison03:
Refactor the system purpose card to make it work with activation keys and hosts

Implementing an indicator between when host details or ak details are passed it

See if system, purpose card is populated if activation key has correlating data, see if data can be edited using edit modal, see if card is still fully functional on host details page.


jeremylenz:
See https://github.com/Katello/katello/pull/10697

PR #10668, opened 2023-07-28
Fixes #36622 - Mark needs_publish after reindexing

sjha4:
Set needs_publish true when repo is indexed.
An indexing in katello will now show up as needs_publish=true in the CV. Thanks to optimizations, when repo contents haven't changed, indexing is skipped so in most cases, CV won't set needs_publish=true when contents haven't changed. But if a forced index is run on repo wthout adding new content, CV will have needs_publish true. You can see this by going to console and running: repo.index_content manually. 
The reproducer steps are a little weird because you need to fake an indexing error in katello to reproduce this.
1: (Fake the error)
In app/models/katello/repository.rb , method index_content add `fail _('Forced error')` as first line. (Downstream requires server restart to reload updates: `foreman-maintain service restart`)
2. Restart the server and create a new yum repo. Do not sync the repo yet.
3. Add new repo to a CV and publish the Content View.
4. Sync the repo. Skip the indexing failure from Dynflow console of the task and let the task finish with warning. You shouldn't see any content in the repo.
5. Publish the CV again.
6. Remove the  `fail _('Forced error')` from  app/models/katello/repository.rb , method index_content
7. Restart server (Downstream requires server restart to reload updates: `foreman-maintain service restart`)
8. Sync the repo now.
9. You should see the CV have the needs_publish icon show up once contents are added correctly in katello.

### DIFF
--- a/app/models/katello/content_view.rb
+++ b/app/models/katello/content_view.rb
@@ -794,12 +794,20 @@ module Katello
       return last_publish_result.present? && last_publish_result == 'success'
     end
 
+    def cv_repo_indexed_after_last_published?
+      repositories.any? { |repo| repo.last_indexed && repo.last_indexed > latest_version_object.created_at }
+    end
+
     def needs_publish?
       #Returns
       # True:
       #     a) When content/repo/filter change audit records exist
       #     b) CV hasn't ever been published
       #     c) CV dependency_solving != latest_version.applied_filters.dependency_solving
+      #     d) If repo was indexed after cv publish. This can happen under 3 cases:
+      #       i) Index runs because last index(before publish) had failed and repo is picked up for index even if pulp publication hasn't changed.
+      #       ii) Complete sync runs or sync adds/removes new content (Already true because new pulp publication/version gets created)
+      #       iii) repo.index_content is run. (This doesn't necessarily indicate contents changed. Corner case where we play safe and return true)
       # nil:
       #     a) When CV version creation audit is missing(Indicating audit cleanup)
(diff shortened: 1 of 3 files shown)


app/models/katello/content_view.rb
```diff
@@ -794,12 +794,20 @@ def last_publish_task_success?
...
     def needs_publish?
       #Returns
       # True:
       #     a) When content/repo/filter change audit records exist
       #     b) CV hasn't ever been published
       #     c) CV dependency_solving != latest_version.applied_filters.dependency_solving
+      #     d) If repo was indexed after cv publish. This can happen under 3 cases:
+      #       i) Index runs because last index(before publish) had failed and repo is picked up for index even if pulp publication hasn't changed.
+      #       ii) Complete sync runs or sync adds/removes new content (Already true because new pulp publication/version gets created)
+      #       iii) repo.index_content(force_index: true) is run. (This doesn't necessarily indicate contents changed. Corner case where we play safe and return true)
```
ianballou, line 810:
```suggestion
      #       iii) repo.index_content is run. (This doesn't necessarily indicate contents changed. Corner case where we play safe and return true)
```

From the code and testing it looks like `force_index: true` isn't needed?

sjha4: Hah..I had that meant for the action Repository::IndexContent which takes that param but then I moved the indexed_at datetime update to index_content method to be more accurate..Will update the text here.

ianballou (approved):
Looking good!

PR #10669, opened 2023-07-28
Fixes # 36623 - Add buttons to change content source screen

jeremylenz:

Add buttons to the Change Content Source screen so it's less confusing.

image

When you click either of the buttons, the content source is updated on the Foreman / Satellite server.

In the case of 'Run job invocation', you are then redirected to the job wizard.
In the case of 'Update hosts manually', the following is presented:

image

Other changes:

* Add a description blurb explaining why you might want to change a content source
* Change 'Environment' to 'Lifecycle environment'. I feel like 'Environment' can mean too many things, and want to start getting away from using it to describe lifecycle environments.

You should be able to test this without actually setting up multiple content sources.

### DIFF
--- a/app/controllers/katello/concerns/hosts_controller_extensions.rb
+++ b/app/controllers/katello/concerns/hosts_controller_extensions.rb
@@ -91,7 +91,7 @@ module Katello
                                       .distinct
 
           if Katello.with_remote_execution?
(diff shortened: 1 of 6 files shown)


jeremylenz:
GH isn't picking up my new commit, let's try another PR

jeremylenz:
see https://github.com/Katello/katello/pull/10670

PR #10671, opened 2023-07-31
Fixes #36625 - allow verify checksum for all repo types

ianballou:
Allows verifying checksums for all repo types. Looking at the APIs for different Pulp plugins, it appears they all support version repair.

All types should support version repair

Try a verify checksum sync for all kinds of Katello repositories. Yum and Debian were only supported in the past.


sjha4 (approved):
`hammer sync --validate-content true` now works for all the below content types I had on my box. Was able to verify the repair task in pulp and see the task progress etc as expected.
Screenshot from 2023-08-01 15-45-56

Ack 👍🏼

ianballou:
[test katello]

PR #10672, opened 2023-08-01
Fixes #36628 - update audit is not created after updating CV name

sjha4:
Adds audit back to name,label,description fields. (label can't be changed but no benefit of excluding it specifically.)
1. Create a CV
2. Update name/description of CV
3. Go to Monitor > Audits
4. Check if an audit record is created for the updates on name/description.


chris1984 (approved):
ACK, code looks fine and it works well

Screenshot 2023-08-02 at 12-05-17 Audits

PR #10674, opened 2023-08-01
Fixes #36629 - Add cvv synced to env created at on capsule sync page

chris1984:

* The published date should be for the version that's tied to the env in that capsule
* Before it was the last published date of the latest version of the CV

N/A

* Check out PR
* Create a CV, publish and promote it to an env associated with a smart-proxy so it syncs
* Go to Infrastructure --> Smart Proxies --> <your smartproxy) --> Content --> Expand environment
* Check if the created at matches the cv version synced to the env instead of the initial cv created at time


sjha4 (approved):
Before PR (Published date shows latest version publish date):
Before_PR
After PR (Published date shows assigned version publish date):
After_PR

Ack 👍🏼

PR #10675, opened 2023-08-02
Fixes #36634 - Fail when wrong/missing ids are passed to bulk ACS actions

sjha4:
Fail when invalid ids are passed to ACS bulk actions
Try bulk actions like below with invalid IDs:

`hammer alternate-content-source bulk refresh --ids 1,2,3`
or 
`hammer alternate-content-source bulk destroy --ids 1,2,3`


chris1984 (approved):
ACK, works well

PR #10676, opened 2023-08-02
Fixes #36631 - add file:// to ACS path types

ianballou:
Adds file:// to the list of suggested path types for ACSs.

...
Create and edit an ACS. In the greyed out empty-text for the base URLs, ensure that `file://` is there.

image


sjha4 (approved):
Ship it!

PR #10680, opened 2023-08-04
Remove katello-agent API endpoints and rabl nodes

jeremylenz:


jeremylenz:
see https://github.com/Katello/katello/pull/10690

PR #10681, opened 2023-08-04
Fixes #36638 - Allow metadata regenerate with --force flag on API/Hammer

sjha4:
The plan for Regenerating metadata on "complete_mirroring" is as follows:
a) **Disallowed on the UI** : Republishing on UI on repos and cv versions with complete_mirroring repos will not be allowed.
b) **Allowed in hammer** : only with `--force true`
We want to discourage users from republishing metadata locally on Pulp for repositories with complete mirroring. Disallowing on UI and only allowing via hammer/api with --force flag makes it so.
1. Create a yum repo with mirroring policy = Complete Mirroring and another with not mirror complete.
2. Sync the repos, add them to a CV and publish.
3. Check to see that  Republishing Metadata action is disabled on repo with complete mirroring and enabled for the other on the UI.
4. In hammer, you should be able to regenerate metadata on both, however only with --force 1 on complete mirroring repo.
Check `hammer -r repository republish --id $REPO_ID --force false ` vs `hammer -r repository republish --id $REPO_ID --force true`
5. For the CV version, republishing metadata from Version Details > Actions will be disabled. (Since a mirror complete repo exists in the version)
6. Via hammer, you'll get an error message when republishing metadata without force flag.
7. Hammer should allow republishing metadata with --force flag. Check number of sub-actions in the bulk task to make sure all repos get republished with 
`hammer -r content-view version republish-repositories --id $VERSION_ID --force 1` and `hammer -r content-view version republish-repositories --id $VERSION_ID --force 0`

### DIFF
--- a/app/controllers/katello/api/v2/content_view_versions_controller.rb
+++ b/app/controllers/katello/api/v2/content_view_versions_controller.rb
@@ -82,9 +82,15 @@ module Katello
 
     api :PUT, "/content_view_versions/:id/republish_repositories", N_("Forces a republish of the version's repositories' metadata")
     param :id, :number, :desc => N_("Content view version identifier"), :required => true
-    param :force, :bool, :desc => N_("Force metadata regeneration to proceed. (Deprecated)"), deprecated: true
+    param :force, :bool, :desc => N_("Force metadata regeneration to proceed. Dangerous operation when version has repositories with the 'Complete Mirroring' mirroring policy")
     def republish_repositories
-      task = async_task(::Actions::Katello::ContentViewVersion::RepublishRepositories, @content_view_version)
+      mirror_complete_repos = @content_view_version.repositories.joins(:root).where(root: { mirroring_policy: ::Katello::RootRepository::MIRRORING_POLICY_COMPLETE })
+      if mirror_complete_repos.size > 0 && !::Foreman::Cast.to_bool(params[:force])
+        fail HttpErrors::BadRequest, _("Metadata republishing is dangerous on content view versions with repositories with the 'Complete Mirroring' mirroring policy.
+Change the mirroring policy on these repositories: #{mirror_complete_repos.pluck(:name)} and try again.
+Alternatively, use the 'force' parameter to regenerate metadata locally. New versions of the content view will continue to use upstream metadata for repositories with 'Complete Mirroring'.")
+      end
+      task = async_task(::Actions::Katello::ContentViewVersion::RepublishRepositories, @content_view_version, force: ::Foreman::Cast.to_bool(params[:force]))
       respond_for_async :resource => task
     end
 
--- a/app/controllers/katello/api/v2/repositories_controller.rb
+++ b/app/controllers/katello/api/v2/repositories_controller.rb
@@ -325,10 +325,11 @@ Pass [] to make repo available for clients regardless of OS version. Maximum len
 
     api :PUT, "/repositories/:id/republish", N_("Forces a republish of the specified repository, regenerating metadata and symlinks on the filesystem. Not allowed for repositories with the 'Complete Mirroring' mirroring policy.")
     param :id, :number, :desc => N_("Repository identifier"), :required => true
-    param :force, :bool, :desc => N_("Force metadata regeneration to proceed. (Deprecated)"), deprecated: true
+    param :force, :bool, :desc => N_("Force metadata regeneration to proceed. Dangerous when repositories use the 'Complete Mirroring' mirroring policy")
     def republish
-      if @repository.mirroring_policy == Katello::RootRepository::MIRRORING_POLICY_COMPLETE
-        fail HttpErrors::BadRequest, _("Metadata republishing is not allowed on repositories with the 'Complete Mirroring' mirroring policy.")
+      if @repository.mirroring_policy == Katello::RootRepository::MIRRORING_POLICY_COMPLETE && !::Foreman::Cast.to_bool(params[:force])
+        fail HttpErrors::BadRequest, _("Metadata republishing is risky on 'Complete Mirroring' repositories. Change the mirroring policy and try again.
+Alternatively, use the 'force' parameter to regenerate metadata locally. On the next sync, the upstream repository's metadata will overwrite local metadata for 'Complete Mirroring' repositories.")
       end
       task = async_task(::Actions::Katello::Repository::MetadataGenerate, @repository)
       respond_for_async :resource => task
--- a/app/lib/actions/katello/content_view_version/republish_repositories.rb
+++ b/app/lib/actions/katello/content_view_version/republish_repositories.rb
@@ -2,10 +2,16 @@ module Actions
   module Katello
     module ContentViewVersion
       class RepublishRepositories < Actions::EntryAction
-        def plan(content_view_version)
+        def plan(content_view_version, options = {force: false})
+          force = options[:force]
           action_subject(content_view_version.content_view)
           plan_self(:version_id => content_view_version.id)
-          plan_action(::Actions::Katello::Repository::BulkMetadataGenerate, content_view_version.repositories.joins(:root).where.not(root: { mirroring_policy: ::Katello::RootRepository::MIRRORING_POLICY_COMPLETE }))
+          repositories = if force
+                           content_view_version.repositories
+                         else
+                           content_view_version.repositories.joins(:root).where.not(root: { mirroring_policy: ::Katello::RootRepository::MIRRORING_POLICY_COMPLETE })
+                         end
+          plan_action(::Actions::Katello::Repository::BulkMetadataGenerate, repositories)
         end
 
         def run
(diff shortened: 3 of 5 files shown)


app/controllers/katello/api/v2/content_view_versions_controller.rb
```diff
@@ -82,9 +82,14 @@ def update
 
     api :PUT, "/content_view_versions/:id/republish_repositories", N_("Forces a republish of the version's repositories' metadata")
     param :id, :number, :desc => N_("Content view version identifier"), :required => true
-    param :force, :bool, :desc => N_("Force metadata regeneration to proceed. (Deprecated)"), deprecated: true
+    param :force, :bool, :desc => N_("Force metadata regeneration to proceed. Dangerous operation when version has repositories with the 'Complete Mirroring' mirroring policy")
     def republish_repositories
-      task = async_task(::Actions::Katello::ContentViewVersion::RepublishRepositories, @content_view_version)
+      mirror_complete_repos = @content_view_version.repositories.joins(:root).where(root: { mirroring_policy: ::Katello::RootRepository::MIRRORING_POLICY_COMPLETE })
```
jeremylenz, line 87: It might need to be a different kind of join, to exclude repos outside the CV?

app/controllers/katello/api/v2/content_view_versions_controller.rb
```diff
@@ -82,9 +82,14 @@ def update
...
+      if mirror_complete_repos.size > 0 && !::Foreman::Cast.to_bool(params[:force])
+        fail HttpErrors::BadRequest, _("Metadata republishing is dangerous on content view versions with repositories with the 'Complete Mirroring' mirroring policy.
+          Update mirroring policies on these repositories #{mirror_complete_repos.pluck(:name)} or Use 'force' parameter to stop mirroring upstream metadata and regenerate metadata locally.")
```
jeremylenz, line 90:
```suggestion
          Update mirroring policies on these repositories #{mirror_complete_repos.pluck(:name)}, or use the 'force' parameter to stop mirroring upstream metadata and regenerate metadata locally.")
```

app/controllers/katello/api/v2/repositories_controller.rb
```diff
@@ -325,10 +325,11 @@ def repository_types
...
     param :id, :number, :desc => N_("Repository identifier"), :required => true
-    param :force, :bool, :desc => N_("Force metadata regeneration to proceed. (Deprecated)"), deprecated: true
+    param :force, :bool, :desc => N_("Force metadata regeneration to proceed. Dangerous when repositories use the 'Complete Mirroring' mirroring policy")
     def republish
-      if @repository.mirroring_policy == Katello::RootRepository::MIRRORING_POLICY_COMPLETE
-        fail HttpErrors::BadRequest, _("Metadata republishing is not allowed on repositories with the 'Complete Mirroring' mirroring policy.")
+      if @repository.mirroring_policy == Katello::RootRepository::MIRRORING_POLICY_COMPLETE && !::Foreman::Cast.to_bool(params[:force])
+        fail HttpErrors::BadRequest, _("Metadata republishing is dangerous on repositories with the 'Complete Mirroring' mirroring policy.
+          Change mirroring policy on the repository and try again or use 'force' parameter via hammer/API to stop mirroring upstream metadata and regenerate metadata locally.")
```
jeremylenz, line 332:
```suggestion
          Change the mirroring policy on the repository and try again, or use the 'force' parameter to stop mirroring upstream metadata and regenerate metadata locally.")
```

app/lib/actions/katello/content_view_version/republish_repositories.rb
```diff
@@ -2,10 +2,16 @@ module Actions
...
-        def plan(content_view_version)
+        def plan(content_view_version, options = {force: false})
+          force = options[:force]
           action_subject(content_view_version.content_view)
           plan_self(:version_id => content_view_version.id)
-          plan_action(::Actions::Katello::Repository::BulkMetadataGenerate, content_view_version.repositories.joins(:root).where.not(root: { mirroring_policy: ::Katello::RootRepository::MIRRORING_POLICY_COMPLETE }))
+          repositories = if force
+                           content_view_version.repositories
+                         else
+                           content_view_version.repositories.joins(:root).where.not(root: { mirroring_policy: ::Katello::RootRepository::MIRRORING_POLICY_COMPLETE })
```
jeremylenz, line 12: Same thing for the join here..

app/controllers/katello/api/v2/repositories_controller.rb
jeremylenz, line 332: This implies that using `force` will actually change the repo's mirroring policy. I know that's what happens in effect, but it doesn't _actually_ change the mirroring policy, right? Should it?

jeremylenz (commented):
Thanks @sjha4 

Everything works as advertised, except this:

Seems the hammer error message here includes all mirror-complete repos:

```
Could not republish the Content View:
  Metadata republishing is dangerous on content view versions with repositories with the 'Complete Mirroring' mirroring policy.
            Update mirroring policies on these repositories ["Red Hat Enterprise Linux 8 for x86_64 - BaseOS RPMs 8", "Red Hat Enterprise Linux 8 for x86_64 - AppStream RPMs 8", "JustinComplete"] or Use 'force' parameter to stop mirroring upstream metadata and regenerate metadata locally.
```

But I don't have the RHEL ones included in the content view:

image

app/controllers/katello/api/v2/repositories_controller.rb
sjha4: It does not change the mirroring policy on repo object..I'll remove the `stop mirroring upstream metadata and` part of the message.

jeremylenz: Using force will regenerate metadata locally for now, but then continue mirroring upstream metadata on the next sync, right? Maybe we should mention that instead..

app/controllers/katello/api/v2/content_view_versions_controller.rb
sjha4: Hmm..Not sure I can reproduce. Did you not have those repos in the CV version?

jeremylenz: Nope, only the custom repos were added.

app/controllers/katello/api/v2/repositories_controller.rb
sjha4: `Metadata republishing is dangerous on repositories with the 'Complete Mirroring' mirroring policy. Change mirroring policy on the repository and try again if you want to stop using upstream metadata for this repository. Alternatively, you can use the 'force' parameter via hammer/API to regenerate metadata locally. This local metadata will be overwritten by upstream metadata on the next complete sync if repository mirroring policy is 'Complete Mirroring'.`

app/controllers/katello/api/v2/content_view_versions_controller.rb
jeremylenz: I was using the wrong id.. ignore me :see_no_evil:

app/controllers/katello/api/v2/content_view_versions_controller.rb
```diff
@@ -82,9 +82,14 @@ def update
...
     param :id, :number, :desc => N_("Content view version identifier"), :required => true
-    param :force, :bool, :desc => N_("Force metadata regeneration to proceed. (Deprecated)"), deprecated: true
+    param :force, :bool, :desc => N_("Force metadata regeneration to proceed. Dangerous operation when version has repositories with the 'Complete Mirroring' mirroring policy")
     def republish_repositories
-      task = async_task(::Actions::Katello::ContentViewVersion::RepublishRepositories, @content_view_version)
+      mirror_complete_repos = @content_view_version.repositories.joins(:root).where(root: { mirroring_policy: ::Katello::RootRepository::MIRRORING_POLICY_COMPLETE })
+      if mirror_complete_repos.size > 0 && !::Foreman::Cast.to_bool(params[:force])
+        fail HttpErrors::BadRequest, _("Metadata republishing is dangerous on content view versions with repositories with the 'Complete Mirroring' mirroring policy.
+          Update mirroring policies on these repositories #{mirror_complete_repos.pluck(:name)}, or use the 'force' parameter to stop mirroring upstream metadata and regenerate metadata locally.")
```
jeremylenz, line 90:
```suggestion
          Change the mirroring policy on these repositories #{mirror_complete_repos.pluck(:name)} and try again, or use the 'force' parameter to regenerate metadata locally. On the next sync, the upstream repository's metadata will overwrite local metadata for 'Complete Mirroring' repositories.")
```

sjha4: `On the next sync, the upstream repository's metadata will overwrite local metadata for 'Complete Mirroring' repositories.`
This is not accurate for CV version repos..Those will not be overwritten by library repo syncs..

sjha4: `New versions of the content view will continue to use upstream metadata for repositories with 'Complete Mirroring'`

jeremylenz (approved):
ACK :+1:

PR #10683, opened 2023-08-07
Fixes #36641 - use Minitest namespace

evgeni:

making old code work with modern minitest

letting the code bit rot and pin minitest is not an option

CI + https://github.com/theforeman/foreman/pull/9786


sjha4:
Is this running against foreman develop branch and passing on older gems? Should we merge the foreman PR before or this one to get foreman CI test green?

evgeni:
I'd prefer to merge this (right now it ran with 5.18 minitest) and then via https://github.com/theforeman/foreman/pull/9786 verify it works with newer minitest too.

(or you can verify the later locally too, of course)

PR #10687, opened 2023-08-08
Fixes #36654 - Remove dead content_view_solve_dependencies setting

jeremylenz:

1. Remove the setting `content_view_solve_dependencies`, which was previously used to set whether depsolving was enabled by default for new content views, but now is apparently not used anymore.
2. Remove reference to this setting in bastion_katello

Verify that the setting "content view Dependency Solving Default" is no longer there


chris1984 (approved):
ACK, don't see the setting anymore in the UI or listing it with hammer

jeremylenz:
@parthaa @sjha4  Just double-checking, we're ok with removing this setting, right?

sjha4:
Seems good to me..The default is false for dep_solving and we shouldn't be making it easier to turn it on for all CVs..

PR #10690, opened 2023-08-09
Fixes #36665 - Remove katello-agent API endpoints and rabl nodes

jeremylenz:

Remove Katello API endpoints that use katello-agent:

* `/hosts/:host_id/errata/apply`
* `/hosts/:host_id/packages/install`
* `/hosts/:host_id/packages/upgrade`
* `/hosts/:host_id/packages/upgrade_all`
* `/hosts/:host_id/packages/remove`
* `/hosts/bulk/install_content`
* `/hosts/bulk/update_content`
* `/hosts/bulk/remove_content`

Also remove katello-agent from `api/ping`.

This will break all related Hammer commands, so we should probably wait and merge the hammer-cli-katello PR at the same time as this one

Try existing hammer commands and confirm they're broken, I guess


test/services/katello/registration_manager_test.rb
```diff
@@ -275,29 +275,15 @@ def test_registration_existing_host
...
         @host = FactoryBot.create(:host, :with_content, :with_subscription, :content_view => @content_view,
                                    :lifecycle_environment => @library, :organization => @content_view.organization)
-        ::Katello.expects(:with_katello_agent?).returns(false)
         @host.content_facet.expects(:cves_changed?).returns(false)
         ::Katello::Resources::Candlepin::Consumer.expects(:destroy)
         ::Katello::EventQueue.expects(:push_event).never
         ::Katello::RegistrationManager.unregister_host(@host, :unregistering => true)
       end
 
-      def test_unregister_host_with_katello_agent
```
jturel, line 288: I was admiring your work when it occurred to me that there's always a small chance that after all this cleanup is done, there may be "delete host agent queue" events in the db which will break the event queue when they get processed because their handler class will be removed. You'll  need an upgrade rake task to delete those from the events table

jeremylenz: Yeah, I was just thinking about delete_host_agent_queue and how to handle that..  :cry:

parthaa: Do we need that? Isn't the queue emptied when upgrading any way?

app/controllers/katello/api/v2/hosts_bulk_actions_controller.rb
```diff
@@ -117,37 +116,6 @@ def installable_errata
                                                      :resource_class => Erratum))
     end
 
-    api :PUT, "/hosts/bulk/install_content", N_("Install content on one or more hosts using katello-agent. %s") % katello_agent_deprecation_text, deprecated: true
```
parthaa, line 120: Need to remove from => https://github.com/Katello/katello/blob/master/config/routes/overrides.rb#L72

parthaa: Basically you need to do this for every action you removed.

test/services/katello/registration_manager_test.rb
jturel: The queue isn't explicitly emptied on upgrade and there could be events on it especially for this specific case because it's scheduled to run ten minutes in the future. Any hosts unregistered ~10 minutes before shutdown will put the events on queue and if Katello is upgraded with that event removed it'll raise an error around an unsupported event. There could also be events stuck on the queue due to this bug: #10692 

Looking at the code: https://github.com/Katello/katello/blob/c9fe25d6705d5cf84dc60fefb521e8d3e939688b/app/services/katello/event_queue.rb#L73-L75

An error gets raised if an unsupported event is found on queue which is what will happen when the event's class is removed from the app. However, that error may be rescued, logged, and the event removed: https://github.com/Katello/katello/blob/c9fe25d6705d5cf84dc60fefb521e8d3e939688b/app/services/katello/event_monitor/poller_thread.rb#L61-L70

Definitely test that to make sure it doesn't break the queue though.

jeremylenz:
@parthaa Updated route overrides :+1:

test/services/katello/registration_manager_test.rb
jeremylenz: So, do you think I should add to the rescue here instead of the upgrade rake task?

Either way, can this be addressed in a followup issue?

jturel: I think the rescue that's there will handle the situation already - but that needs a confirmation - definitely outside of the scope here!

jturel (approved):
LGTM ;)

parthaa (approved):

jeremylenz:
Waiting until https://github.com/theforeman/forklift/pull/1695 is merged.

jeremylenz:
[test katello]

PR #10694, opened 2023-08-11
Fixes #36674 - don't update Candlepin consumer when force-reregistering

jeremylenz:

Mark CVEs unchanged in `remove_host_artifacts`. This prevents `backend_update_needed?` from erroneously returning `true` during the process of host removal.

This was manifesting when you do a force reregister with subscription-manager without having consumer certs.

I considered passing around some sort of `destroying_host?` variable instead, but that seemed like too many hops to me, and reminded me of React prop drilling.

```
rm -rf /etc/pki/consumer/*
subscription-manager register --force --username=admin --password=changeme --org=Default_Organization --environment=Library
```

**Before**: The first attempt fails with 
```
Registering to: centos8-katello-devel-stable.example.com:443/rhsm
Unit 547cf5fd-3ff0-44d9-b26c-5944bbbb9fbd has been deleted (HTTP error code 410: Gone)
```

The second attempt succeeds.

**After**: All attempts succeed.

also test other registration flows and make sure nothing is broken

### DIFF
--- a/app/models/katello/host/content_facet.rb
+++ b/app/models/katello/host/content_facet.rb
@@ -61,6 +61,10 @@ module Katello
         self.cves_changed = true
       end
 
+      def mark_cves_unchanged
+        self.cves_changed = false
+      end
+
       def cves_changed?
         cves_changed
       end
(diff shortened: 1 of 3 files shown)


ianballou (approved):
Looking great to me!

PR #10695, opened 2023-08-14
Fixes #36678 - Support Link header in docker v2 repo discovery

stbergmann:
...as e.g. needed by a Repository Discovery with custom container images URL <https://registry.fedoraproject.org/>, which only includes the first 100 results in the original response.


app/lib/katello/repo_discovery.rb
```diff
@@ -42,6 +44,70 @@ def run(resume_point)
 
     private
 
+    def parse_parameters(field_value)
```
ianballou, line 47: I have a proposition for simplifying how we determine if the Link header needs to be used.

Looking at the spec from Docker for how a container registry should communicate pagination (https://docs.docker.com/registry/spec/api/#pagination), it says: 
```
To get the next result set, a client would issue the request as follows, using the URL encoded in the described Link header:

...

The above process should then be repeated until the Link header is no longer set.
```

So, perhaps we can simply directly check for `rel="next"` ?  Or, if we're being super careful, we could do:
```ruby
link.include?('rel=\"next\"') || link.include?('rel="next"')
```
The same could be said for `anchor`.

app/lib/katello/repo_discovery.rb
```diff
@@ -42,6 +44,70 @@ def run(resume_point)
...
+        break if i.nil?
+        target_string = field_value[0, i]
+        field_value.slice!(0, i + 1)
+        params = parse_parameters(field_value)
+        if params[:has_next_rel]
+          # To keep it simple, ignore a link with an (unlikely) anchor parameter; but the RFC
+          # mandates that we "MUST NOT process the link without applying the anchor", so just raise
+          # an exception in that (unlikely) case:
+          raise "anchor not supported" if params[:has_anchor]
+          return target_string
```
ianballou, line 105: `target_string` could potentially be found more easily with regex:

```ruby
[31] pry(main)> regex = /\<(.*)>/
=> /\<(.*)>/
[32] pry(main)> str = ";<https://website.org?thing=1&otherthing=2>;"
=> ";<https://website.org?thing=1&otherthing=2>;"
[33] pry(main)> str[regex, 1]
=> "https://website.org?thing=1&otherthing=2"
```

ianballou (commented):
Finally got around to reviewing this. I haven't tested it yet, but I had some suggestions:

ianballou:
Rubocop is failing: https://ci.theforeman.org/blue/organizations/jenkins/katello-pr-test/detail/katello-pr-test/13324/pipeline/118

You can check this locally by running the following in the Foreman dir: `bundle exec rake katello:rubocop`

app/lib/katello/repo_discovery.rb
ianballou: Also I'm assuming the Link has a form like: `"link"=>["<http://pulpcore-api/v2/container-image-name/tags/list?n=100&last=last-tag-name>; rel=\"next\""]`

stbergmann: The Docker spec at <https://docs.docker.com/registry/spec/api/#pagination> says:

> If there are indeed more results, the URL for the next block is encoded in an [RFC5988open_in_new](https://tools.ietf.org/html/rfc5988) Link header, as a "next" relation.

It is silent on whether a response could carry further, different Link header data.  Therefore, I would assume that it is better that a client explicitly checks for a "next" link encoded as per the Link spec (RFC 8288 obsoleting RFC 5988), rather than extrapolating from example Link headers seen in the wild.

(Though I do understand the sentiment to keep the code simple.  And I wasn't impressed that I apparently had to code this all up ad-hoc here, instead of being able to reuse some existing code.)

stbergmann:
> Rubocop is failing: https://ci.theforeman.org/blue/organizations/jenkins/katello-pr-test/detail/katello-pr-test/13324/pipeline/118

addressed with  acc9221 now

app/lib/katello/repo_discovery.rb
stbergmann: addressed with 04eecce now

(Though a variant using `$1` and `$'` didn't appear to work and left those empty; but I'm very ignorant of Ruby.  And the quoted `\<` in your sample regexp looks like it wasn't necessary, or was it?)

app/lib/katello/repo_discovery.rb
```diff
@@ -42,6 +44,73 @@ def run(resume_point)
 
     private
 
+    def parse_parameter(field_value)
+      field_value.sub!(/^[ \t]*/, '')
+      i = field_value.index(/[ \t=;,]/) || field_value.length
```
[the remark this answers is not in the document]
ianballou: One caveat I found:

```ruby
s = "<http://pulpcore-api/v2/container-image-name/tags/list?n=100&last=last-tag-name>; rel=\"next\""

m  = s.match(/<(.*)>/)
=> "; rel=\"next\""

m.post_match.split(/[ \t=;,]/, 2).map(&:strip)
=> ["", "rel=\"next\""]



```

The non-whitespace delimiters (here a semicolon) between the URL and the key-value pairs need to be cut out.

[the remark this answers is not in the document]
ianballou: Actually, one amendment:

if `field_value` is `"rel=\"next\" anchor=\"blah\""`, then the anchor part will end up in `value`.

If we take out the limit of `2`, then the first and second entries into the array that `strip` returns would definitely include the name and value of the first argument.

```ruby
params = "rel=\"next\" anchor=\"blah\""
[59] pry(main)> n,v = params.split(/[ \t=;,]/).map(&:strip)
=> ["rel", "\"next\"", "anchor", "\"blah\""]
[61] pry(main)> n
=> "rel"
[62] pry(main)> v
=> "\"next\""
```

I realize the split method won't jiving perfectly with the slicing that you're doing to the args string so that you can loop through it and collect up all of the key-value pairs, but hopefully it serves as some good inspiration/learning/etc.

app/lib/katello/repo_discovery.rb
parthaa: > It is silent on whether a response could carry further, different Link header data. Therefore, I would assume that it is better that a client explicitly checks for a "next" link encoded as per the Link spec (RFC 8288 obsoleting RFC 5988), rather than extrapolating from example Link headers seen in the wild.

I understand 8288 obsoletes 5988 but isn't the docker spec saying it has to be in 5988 format ?

For example I ran the following curl command. By "next" link are ya referring to `link: </v2/_catalog?last=f32%2Fflatpak-sdk&n=100>; rel="next"` below ?

```bash
$ curl -i -H "Accept: application/json"  -H "Content-type: application/json" "https://registry.fedoraproject.org/v2/_catalog"
HTTP/2 200 
strict-transport-security: max-age=31536000; includeSubDomains; preload
x-frame-options: SAMEORIGIN
x-xss-protection: 1; mode=block
x-content-type-options: nosniff
referrer-policy: same-origin
content-type: application/json; charset=utf-8
docker-distribution-api-version: registry/2.0
link: </v2/_catalog?last=f32%2Fflatpak-sdk&n=100>; rel="next"
vary: Accept
date: Tue, 12 Sep 2023 19:26:23 GMT
content-length: 1533
apptime: D=161655
x-fedora-proxyserver: proxy10.iad2.fedoraproject.org
x-fedora-requestid: ZQC7X4t6dbwX308xkqp_NwAEQQo
server: Apache

{"repositories":["0ad","abiword","aisleriot","alligator","angelfish","aqualung","ardour5","ardour6","arianna","ark","armacycles-ad","artikulate","astromenace","atlantik","atomix","audacious","audacity","baobab","berusky","bijiben","bitstower-markets","blackbox-terminal","blinken","bomber","bovo","calindori","calligraplan","celestia","celestia-qt","cheese","compneuro","convertall","d-feet","d-spy","darktable","dconf-editor","devhelp","dialect","dolphin","dosbox","dosbox-staging","drawing","easytag","elisa-player","endless-sky","eog","eol/astromenace","eol/baobab","eol/meld","epiphany","evince","evolution","exaile","exfalso","extremetuxracer","f29/cassandra","f29/cockpit","f29/fedora-toolbox","f29/golang","f29/httpd","f29/mirrormanager2-mirrorlist","f29/nginx","f29/origin-base","f29/origin-cli","f29/origin-control-plane","f29/origin-docker-builder","f29/origin-egress-dns-proxy","f29/origin-egress-router","f29/origin-f5-router","f29/origin-hyperkube","f29/origin-hypershift","f29/origin-keepalived-ipfailover","f29/origin-nginx-router","f29/origin-recycler","f29/origin-template-service-broker","f29/origin-tests","f29/postgresql","f29/ruby","f29/s2i-base","f29/s2i-core","f30/fedora-toolbox","f30/flatpak-runtime","f30/flatpak-sdk","f30/mariadb","f30/s2i-base","f30/s2i-core","f31/fedora-toolbox","f31/flatpak-runtime","f31/flatpak-sdk","f31/mariadb","f31/postgresql","f31/python-classroom","f31/python3","f31/redis","f31/ruby","f31/s2i-base","f31/s2i-core","f32/fedora-toolbox","f32/flatpak-runtime","f32/flatpak-sdk"]}


```

ianballou: @parthaa yep, that's the Link we're talking about.

stbergmann: > I understand 8288 obsoletes 5988 but isn't the docker spec saying it has to be in 5988 format ?

But none of the differences between the two RFCs should be of practical concern for us here, see <https://www.rfc-editor.org/rfc/rfc8288#appendix-C> "Changes from RFC 5988".

app/lib/katello/repo_discovery.rb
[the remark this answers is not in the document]
stbergmann: I think, for a general case like `<http://example.com>; title=foo; rel="bar next baz"` we don't get away with anything simpler than tediously following the RFC 8288 algorithm.

app/lib/katello/repo_discovery.rb
```diff
@@ -42,6 +44,73 @@ def run(resume_point)
 
     private
 
+    def parse_parameter(field_value)
+      field_value.sub!(/^[ \t]*/, '')
```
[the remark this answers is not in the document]
stbergmann: done in  4fa3cde now

app/lib/katello/repo_discovery.rb
```diff
@@ -42,6 +44,73 @@ def run(resume_point)
...
+            break if field_value.delete_prefix!('"')
+            field_value.delete_prefix!("\\")
+            value += field_value.slice!(0, 1) || ''
+          end
+        else
+          i = field_value.index(/[;,]/) || field_value.length
+          value = field_value.slice!(0, i)
+        end
+      else
+        value = ''
```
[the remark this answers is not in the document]
stbergmann: done in  4fa3cde now

app/lib/katello/repo_discovery.rb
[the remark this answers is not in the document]
ianballou: Alright, I'm fine with keeping it as-is to follow the spec guidelines.

app/lib/katello/repo_discovery.rb
```diff
@@ -42,6 +44,71 @@ def run(resume_point)
...
+          has_next_rel = param[:value].downcase(:ascii).split(/[ \t]/).include?('next')
+        when param[:name] == 'anchor'
+          has_anchor = true
+        end
+      end
+      {has_next_rel: has_next_rel, has_anchor: has_anchor}
+    end
+
+    def get_next_link(link_header)
+      # See Appendix B "Algorithms of Parsing Link Header Fiels" of RFC 8288 "Web Linking" (that RFC
```
ianballou, line 89:
```suggestion
      # See Appendix B "Algorithms of Parsing Link Header Fields" of RFC 8288 "Web Linking" (that RFC
```

ianballou: Think it's worth adding a link to https://www.rfc-editor.org/rfc/rfc8288.html#appendix-B.2 ?

ianballou (commented):
Code is looking fine to me! Tested, working with https://registry.fedoraproject.org/.

This test line will need updating to return an object with headers rather than a hash: https://github.com/Katello/katello/blob/master/test/lib/repo_discovery_test.rb#L103

stbergmann:
[test katello]

> Code is looking fine to me! Tested, working with https://registry.fedoraproject.org/.
> 
> This test line will need updating to return an object with headers rather than a hash: https://github.com/Katello/katello/blob/master/test/lib/repo_discovery_test.rb#L103

addressed with [ed016e0](https://github.com/Katello/katello/commit/ed016e05db49ede59d09654685ad38487744659b) now (though that might not be the most idiomatic approach?)

ianballou:
[test katello]

ianballou (approved):
Looking good! Thanks for the contribution @stbergmann

PR #10696, opened 2023-08-14
Fixes #36679 - Sort CV versions when adding component views

sjha4:
Sorts versions for component CVs in the Composite CV -> Add/Bulk Add/Edit Component CVs modals in descending order.
Without the sorting, the versions in the component CV add modals are in random order in single add/ and ascending order in bulk modal. 
Have a few CVs and publish a few versions of each.
Create a composite CV and go to Content view subtab in the Composite CV.
Try single add(row action)/ bulk add(From table action button)/Edit(From pencil icon on added CV row version)
Make sure the versions for component CV in the dropdown are correctly sorted with latest version first down to the earliest version in correct order.

### DIFF
--- a/webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentContentViewAddModal.js
+++ b/webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentContentViewAddModal.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState, useMemo } from 'react';
 import PropTypes from 'prop-types';
 import {
   Flex, Modal, ModalVariant, Select, SelectVariant,
@@ -12,13 +12,12 @@ import { STATUS } from 'foremanReact/constants';
 import {
   selectCVDetails,
   selectCVDetailStatus,
-  selectCVDetailError,
 } from '../../Details/ContentViewDetailSelectors';
 import { addComponent } from '../ContentViewDetailActions';
 import { CONTENT_VIEW_NEEDS_PUBLISH } from '../../ContentViewsConstants';
 
 const ComponentContentViewAddModal = ({
-  cvId, componentCvId, componentId, latest, show, setIsOpen,
+  cvId, componentCvId, componentId, latest, componentVersionId, show, setIsOpen,
 }) => {
   const dispatch = useDispatch();
   const componentDetails = useSelector(
@@ -29,29 +28,20 @@ const ComponentContentViewAddModal = ({
     state => selectCVDetailStatus(state, componentCvId),
     shallowEqual,
   );
-  const componentError = useSelector(
-    state => selectCVDetailError(state, componentCvId),
-    shallowEqual,
-  );
-  const [cvName, setCvName] = useState('');
-  const [options, setOptions] = useState([]);
+  const cvName = componentDetails?.name ?? '';
+  const options = useMemo(() => (componentDetails?.versions ?? []).map(item => ({
+    value: item.id, label: __(`Version ${item.version}`), description: item.description, publishedAtWords: __(` (${item.published_at_words} ago)`),
+  })), [componentDetails?.versions]);
   const [formLatest, setFormLatest] = useState(componentId ? latest : false);
   const [selected, setSelected] = useState(null);
+  const [prevOptions, setPrevOptions] = useState(options);
   const [cvVersionSelectOpen, setCvVersionSelectOpen] = useState(false);
   const versionsLoading = componentStatus === STATUS.PENDING;
 
-  useEffect(() => {
-    if (!versionsLoading && componentDetails) {
-      const { name, versions } = componentDetails;
-      const versionMutable = versions;
-      setCvName(name);
-      const opt = versionMutable.map(item => ({
-        value: item.id, label: __(`Version ${item.version}`), description: item.description, publishedAtWords: __(` (${item.published_at_words} ago)`),
-      }));
-      setOptions([...opt].reverse());
-      setSelected(opt.slice(-1)[0].value);
-    }
-  }, [componentDetails, componentStatus, componentError, versionsLoading]);
+  if (options !== prevOptions) {
+    setPrevOptions(options);
+    setSelected(componentVersionId ?? options[0]?.value);
+  }
 
   const getAddParams = () => {
     if (formLatest) {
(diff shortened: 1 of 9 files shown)


webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentContentViewAddModal.js
```diff
@@ -43,15 +43,15 @@ const ComponentContentViewAddModal = ({
   useEffect(() => {
     if (!versionsLoading && componentDetails) {
       const { name, versions } = componentDetails;
-      const versionMutable = versions;
```
jeremylenz, line 46: This wouldn't even have made a copy of the array.. :smile:

webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentContentViewAddModal.js
```diff
@@ -43,15 +43,15 @@ const ComponentContentViewAddModal = ({
...
+      const versionMutable = [].concat(versions).sort((a, b) => (a.id < b.id ? 1 : -1));
```
jeremylenz, line 46: Why sort on the frontend?

webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentContentViewAddModal.js
```diff
@@ -43,15 +43,15 @@ const ComponentContentViewAddModal = ({
   useEffect(() => {
```
jeremylenz, line 43: We have a bad habit (especially in CV UI) of using effects to compute what should be rendered. 

https://react.dev/learn/you-might-not-need-an-effect

Could this useEffect be eliminated? Or at least remove a lot of the data normalization logic from it and do that during render?

jeremylenz (commented):
Thanks @sjha4 

Some initial comments below :innocent:

webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentContentViewAddModal.js
sjha4: Good point..Let me move this to the API..

webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentContentViewAddModal.js
sjha4: Interesting..I can see removing useMemo in the bulk modal cause we pass all details to render the modal as props so there aren't any API calls that need to propogate effects on success.

The single add/edit modal does send out a call to the API and useEffect looks at API return status. That seems necessary.

sjha4:
React failures expected as fixture needs to be updated to be sorted.

webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentContentViewAddModal.js
sjha4: Updated bulk modal..

jeremylenz:
```
diff --git a/webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentContentViewAddModal.js b/webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentContentViewAddModal.js
index c09d09de43..7309f894f5 100644
--- a/webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentContentViewAddModal.js
+++ b/webpack/scenes/ContentViews/Details/ComponentContentViews/ComponentContentViewAddModal.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, useMemo } from 'react';
 import {
   Flex, Modal, ModalVariant, Select, SelectVariant,
@@ -33,24 +33,20 @@ const ComponentContentViewAddModal = ({
     state => selectCVDetailError(state, componentCvId),
     shallowEqual,
   );
-  const [cvName, setCvName] = useState('');
-  const [options, setOptions] = useState([]);
   const [formLatest, setFormLatest] = useState(componentId ? latest : false);
   const [selected, setSelected] = useState(null);
   const [cvVersionSelectOpen, setCvVersionSelectOpen] = useState(false);
   const versionsLoading = componentStatus === STATUS.PENDING;
+  const cvName = componentDetails?.name ?? '';
+  const options = useMemo(() => (componentDetails?.versions ?? []).map(item => ({
+    value: item.id, label: __(`Version ${item.version}`), description: item.description, publishedAtWords: __(` (${item.published_at_words} ago)`),
+  })), [componentDetails?.versions]);
 
   useEffect(() => {
     if (!versionsLoading && componentDetails) {
-      const { name, versions } = componentDetails;
-      setCvName(name);
-      const opt = versions.map(item => ({
-        value: item.id, label: __(`Version ${item.version}`), description: item.description, publishedAtWords: __(` (${item.published_at_words} ago)`),
-      }));
-      setOptions(opt);
-      setSelected(componentVersionId || opt[0]?.value);
+      setSelected(componentVersionId || options[0]?.value);
     }
-  }, [componentDetails, componentStatus, componentError, versionsLoading, componentVersionId]);
+  }, [versionsLoading, componentDetails, componentVersionId, options]);
 
   const getAddParams = () => {
     if (formLatest) {
```

This simplifies the effect and removes 2 states. This is what I meant by remove the data normalization logic from the effect.

jeremylenz: Now that I think about it, we could remove the effect entirely if we move `componentVersionId || options[0]?.value` into the default value for `selected`:

```js
- const [selected, setSelected] = useState(null);
+ const [selected, setSelected] = useState(componentVersionId || options[0]?.value);
```

(you'd also have to move that line to after we declare `options`)

jeremylenz (commented):
One other thing I noticed.. On the individual add modal, when I check the latest version checkbox, the selection automatically sets itself to the latest version. But this doesn't happen on the bulk add modal.

sjha4:
> One other thing I noticed.. On the individual add modal, when I check the latest version checkbox, the selection automatically sets itself to the latest version. But this doesn't happen on the bulk add modal.

Added this to bulk modal in the last commit.

sjha4:
>> we could remove the effect entirely if we move componentVersionId || options[0]?.value into the default value for selected:

This part doesn't seem to work outside of the effect so I left it in the useEffect. Let me know if the updated code looks good.

jeremylenz:
I was able to remove the effect with [this trick](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes), which I hadn't seen before:

```js
const [selected, setSelected] = useState(null);
const [prevOptions, setPrevOptions] = useState(options);
...
if (options !== prevOptions) {
    setPrevOptions(options);
    setSelected(componentVersionId ?? options[0]?.value);
  }
```

jeremylenz (approved):
Thanks for going on this journey!

ACK :+1:

PR #10699, opened 2023-08-16
Fixes #36684 - allow single_content_view in ContentFacet::Jail

evgeni:

Allow single_content_view in ContentFacet::Jail

Otherwise the Ansible Inventory report template doesn't render properly when Safe Mode is on.

Same for single_lifecycle_environment and KTEnvironment#id.

none

1. have a Katello with at least one system that has a content facet and safe mode on
2. render the "Ansible Inventory" report and it should not fail anymore


evgeni:
:eyes: @ianballou @jeremylenz in https://github.com/theforeman/foreman/pull/9478 you tested this -- was this with Safe Mode off?

jeremylenz:
Looks like `:single_content_view` and `:single_lifecycle_environment` were added to the jail for `Host::Managed` in that PR, but not to `ContentFacet`. I guess the reports we tested with didn't try that on content facet directly..

evgeni:
Can we get this picked to 4.8 and 4.9? 😻

PR #10700, opened 2023-08-16
consume katello repo from koji when doing packit builds

evgeni:
otherwise dependencies are sometimes out of date if nightly is red


evgeni:
How to test? See this PR build fine in Packit, while #10699 did not :)

ehelms (approved):

PR #10701, opened 2023-08-16
Fixes #36686 - temporarily patch out Pulp dist hidden field

ianballou:

Removes the hidden field from Pulpcore 3.28 supported clients so that n-1 smart proxy syncing works.

This should be removed as soon as possible.

Try syncing an n-1 smart proxy (that has Pulpcore 3.22) with varying repository types.


sjha4:
I am seeing this on an ansible collection repo sync on capsule:
https://gist.github.com/sjha4/7497317c51d01dc63d026dce207064ee

ianballou:
@sjha4 this is all good now.

sjha4:
We could pin the gems to the versions in koji.

ianballou:
> We could pin the gems to the versions in koji.

That's alright, we should be able to take in whatever the latest z version is. However, there really shouldn't be major API changes. I'll see if I find any suspicious changes.

sjha4 (approved):
Ack..One known issue of collections we can get to later once the patch is out, This looks good to go!

PR #10703, opened 2023-08-21
Fixes #36690 - Validate *all* products before saving to sync plans

sjha4:
Adds a validate_and_update_products method to sync plans which runs a complete validation of all products of sync plan to remove any possible invalid products left around due to errors in product deletion/repository disable actions etc for any reason. 
To test, you'll have to create invalid data to reproduce the error. To do so, you'll need to disable a validation on sync plans model.
1. Comment out the line `validate :product_enabled` like `#validate :product_enabled`
2. Restart the server.
3. Make sure your org has a valid subscription and create a sync plan (Menu > Content> Sync Plans) to run the following steps on. 
4.In foreman-rake console:
Do this:
```
sync_plan = Katello::SyncPlan.first
sync_plan.product_ids = [1,2,3,4]
sync_plan.save!
```
5. Uncomment the  `#validate :product_enabled` like  `validate :product_enabled`
6. Restart server
7. Try enabling a couple redhat products (Content > Redhat repositories > Enable some random products, don't have to sync or anything) and adding them to the sync plan above.
8. See the error on master branch.
9. With changes above, you should be able to correct the sync plan and see messages in foreman log when sync plan is updated.

### DIFF
--- a/app/controllers/katello/api/v2/sync_plans_controller.rb
+++ b/app/controllers/katello/api/v2/sync_plans_controller.rb
@@ -8,6 +8,7 @@ module Katello
                                                                :add_products, :remove_products]
     before_action :set_organization, :only => [:update, :show, :destroy, :sync,
                                                :add_products, :remove_products]
+    before_action :validate_sync_plan_products, :only => [:update, :add_products, :remove_products]
 
     def_param_group :sync_plan do
       param :name, String, :desc => N_("sync plan name"), :required => true, :action_aware => true
@@ -126,6 +127,10 @@ module Katello
 
     def set_organization
       @organization ||= @sync_plan.try(:organization)
+    end
+
+    def validate_sync_plan_products
+      @sync_plan.validate_and_update_products force_update: true
     end
   end
 end
--- a/app/models/katello/sync_plan.rb
+++ b/app/models/katello/sync_plan.rb
@@ -174,5 +174,22 @@ module Katello
     def cron_status_mismatch?
       self.interval != CUSTOM_CRON && !(self.cron_expression.nil? || self.cron_expression.eql?(''))
     end
+
+    def validate_and_update_products(force_update: false)
+      sync_plan_products = Product.where(id: self.product_ids).select { |p| p.enabled? }
+      return if sync_plan_products.length == self.product_ids.length
+      sync_plan_product_ids = sync_plan_products.pluck(:id)
+      if sync_plan_product_ids.length < self.product_ids.length
+        missing_ids = self.product_ids - sync_plan_product_ids
+        Rails.logger.warn "Sync plan products with following ids are either disabled or don't exist: #{missing_ids}"
+        if force_update
+          Rails.logger.info "Updating sync plan with valid and enabled product ids: #{sync_plan_product_ids}"
+          self.product_ids = sync_plan_product_ids
+          self.save!
+        else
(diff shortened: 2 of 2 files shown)


sjha4:
[test katello]

app/models/katello/sync_plan.rb
```diff
@@ -174,5 +174,25 @@ def rec_logic_changed?
     def cron_status_mismatch?
       self.interval != CUSTOM_CRON && !(self.cron_expression.nil? || self.cron_expression.eql?(''))
     end
+
+    def validate_and_update_products(force_update: false)
+      sync_plan_products = Product.where(id: self.product_ids).select { |p| p.enabled? }
+      if sync_plan_products.length == self.product_ids.length
+        Rails.logger.info "Successfully validated presence of enabled sync plan products!"
```
qcjames53, line 181: Do we need this to be logged? This will be the case nearly 100% of the time.

app/controllers/katello/api/v2/sync_plans_controller.rb
```diff
@@ -92,6 +92,7 @@ def destroy
     param :id, String, :desc => N_("ID of the sync plan"), :required => true
     param :product_ids, Array, :desc => N_("List of product ids to add to the sync plan"), :required => true
     def add_products
+      @sync_plan.validate_and_update_products force_update: true
```
qcjames53, line 95: It looks like running an update on the sync plan also causes invalid products to error out, even when changing innocuous fields like the description. Perhaps we should add this check for updates as well.

app/models/katello/sync_plan.rb
```diff
@@ -174,5 +174,25 @@ def rec_logic_changed?
     def cron_status_mismatch?
       self.interval != CUSTOM_CRON && !(self.cron_expression.nil? || self.cron_expression.eql?(''))
     end
+
+    def validate_and_update_products(force_update: false)
```
qcjames53, line 178: Will we ever need the option to not force update and just log the warning? It looks like in the current implementation, we are always force updating. What was the rationale behind the decision to include that?

qcjames53 (commented):
Hi Samir!

I was able to confirm the changes work as advertised for a variety of scenarios with enabled and disabled products. It all seems very robust. Outside of a couple questions, this looks great to me.

app/models/katello/sync_plan.rb
sjha4: I was not sure if we should or should not keep it..The more logging the better but I guess we can remove it since it's the happy path and we don't need to log that. 👍🏼

app/models/katello/sync_plan.rb
sjha4: I was thinking for support purposes and an option to figure out which sync plans have the issue if we needed to..One question I had which is more a PX kind of question is if we should run a script to fix these "broken" sync plans in user environments..

qcjames53: Ok, I understand why you included it. That seems reasonable to keep the toggle, then. The support script isn't a bad idea at all; I'm a big fan.

app/models/katello/sync_plan.rb
sjha4: Pushed a commit to remove logging in happy path.

app/controllers/katello/api/v2/sync_plans_controller.rb
sjha4: Pushed a commit to validate before update.

qcjames53 (approved):
The changes look great to me. Ack'ed :)

PR #10706, opened 2023-08-22
Fixes #36693 - Add new host status for RHEL lifecycle

jeremylenz:

* Add a new host status, `Katello::RhelLifecycleStatus` which tells you if your RHEL system is getting close to EOS (end of support).

* Hard-code a list of EOS schedules, which we'll have to update whenever they change and whenever a new major RHEL is released.


app/services/katello/registration_manager.rb
```diff
@@ -243,9 +243,7 @@ def finalize_registration(host)
         host.subscription_facet.update_from_consumer_attributes(host.subscription_facet.candlepin_consumer.
             consumer_attributes.except(:guestIds, :facts))
         host.subscription_facet.save!
-        host.subscription_facet.update_subscription_status
-        host.content_facet.update_errata_status
-        host.refresh_global_status!
+        host.refresh_statuses(::Katello::HostStatusManager::STATUSES)
```
jeremylenz, line 246: `refresh_statuses` calls `refresh!` on every `relevant?` status, and then calls `refresh_global_status!`. So it seems like it's better just to do this.

jeremylenz:
[test katello]

app/models/katello/rhel_lifecycle_status.rb
```diff
@@ -0,0 +1,169 @@
...
+    end
+
+    def to_status(operatingsystem: nil)
+      self.class.to_status(operatingsystem: operatingsystem || self.host&.operatingsystem)
+    end
+
+    # this status is only relevant for RHEL
+    def relevant?(_options = {})
+      host&.operatingsystem&.rhel_eos_schedule
+    end
```
jeremylenz, line 167: If `relevant?` returns false, `refresh_host_statuses` will skip this status when refreshing. This is what causes it to remain "N/A" for non-RHEL hosts.

app/models/katello/concerns/operatingsystem_extensions.rb
```diff
@@ -44,6 +44,24 @@ def set_atomic_attributes
       def atomic?
         name.match(/.*atomic.*/i)
       end
+
+      def rhel_eos_schedule
+        return nil unless name == "RedHat" # using name and not title so we get specifically RHEL, not rebuilds
```
jeremylenz, line 49: Is it safe to assume all RHEL OSes use "RedHat" as the name? Or do some use "RHEL" ?

app/models/katello/rhel_lifecycle_status.rb
lfu: Could we display this RHEL lifecycle information only for RHEL hosts? 
So nothing would be added for non-RHEL hosts at all.

app/models/katello/concerns/operatingsystem_extensions.rb
lfu: Seems they are all `RedHat` for RHEL 9, 8 and 7.

app/models/katello/concerns/operatingsystem_extensions.rb
```diff
@@ -44,6 +44,23 @@ def set_atomic_attributes
       def atomic?
         name.match(/.*atomic.*/i)
       end
+
+      def rhel_eos_schedule(arch_name: nil)
```
lfu, line 48: We could have a better name for this method 😄 
It is really `rhel_release` or `rhel_release_index`. Definitely doesn't look like a schedule :)

app/models/katello/rhel_lifecycle_status.rb
```diff
@@ -0,0 +1,175 @@
+module Katello
+  class RhelLifecycleStatus < HostStatus::Status
+    UNKNOWN = 0
```
lfu, line 3: Don't think this is necessary.

app/models/katello/rhel_lifecycle_status.rb
```diff
@@ -0,0 +1,175 @@
...
+      'RHEL7 (POWER9)' => {
+        'full_support' => Date.parse('2019-08-06'),
+        'maintenance_support' => Date.parse('2021-05-31')
+      },
+      'RHEL6' => {
+        'full_support' => Date.parse('2016-05-10'),
+        'maintenance_support' => Date.parse('2020-11-30'),
+        'extended_support' => Date.parse('2024-06-30')
+      },
+      'RHEL5' => {
```
lfu, line 43: Seems all 3 dates for RHEL5 are all wrong :)

app/models/katello/rhel_lifecycle_status.rb
jeremylenz: As far as I can tell, all registered statuses display for all hosts. This is what it looks like for a non-relevant host:

image

Since there's no way to register a host status only for certain hosts, I think this is the best we can do.

app/models/katello/rhel_lifecycle_status.rb
```diff
@@ -0,0 +1,175 @@
...
+    def self.status_map
+      map = {
+        full_support: FULL_SUPPORT,
+        maintenance_support: MAINTENANCE_SUPPORT,
+        extended_support: EXTENDED_SUPPORT,
+        approaching_end_of_support: APPROACHING_END_OF_SUPPORT,
+        support_ended: SUPPORT_ENDED
+      }
+
+      map.default = UNKNOWN
```
lfu, line 61: I think a RHEL host must be in one of the `full_support`, `maintenance_support`, `extended_support` and `support_ended` states at any given point of time. 
And optionally may be also in `approaching_end_of_support` state when one of the 3 end dates is approaching.
Don't think `UNKNOWN` state is needed here.

app/models/katello/rhel_lifecycle_status.rb
```diff
@@ -0,0 +1,175 @@
...
+    UNKNOWN = 0
+    FULL_SUPPORT = 1
+    MAINTENANCE_SUPPORT = 2
+    EXTENDED_SUPPORT = 3
+    APPROACHING_END_OF_SUPPORT = 4
+    SUPPORT_ENDED = 5
+
+    RHEL_EOS_SCHEDULES = { # dates that each support category ends
+      'RHEL9' => {
+        'full_support' => Date.parse('2027-05-31'),
```
lfu, line 12: Can we keep it as DateTime UTC format as it is used in foreman?
```
DateTime.parse('2027-05-31').end_of_day.utc
```

app/models/katello/concerns/operatingsystem_extensions.rb
jeremylenz: I originally had it named something similar, but changed it to this. It does refer to the keys of the `RHEL_EOS_SCHEDULES` elsewhere in this PR:

```rb
RHEL_EOS_SCHEDULES = { # dates that each support category ends
      'RHEL9' => {
 'full_support' => Date.parse('2027-05-31'),
        'maintenance_support' => Date.parse('2032-05-31'),
        'extended_support' => Date.parse('2035-05-31')
 ...
```

For example, here, "RHEL9" is the name of the EOS schedule with those schedule dates.

I thought "release" wasn't as accurate because we have EOS schedules for `'RHEL7'`, `'RHEL7 (System z (Structure A))'`, and `'RHEL7 (ARM)'` and those are three different architectures of the same release.

Naming it a "schedule" also opens up the possibility of adding more EOS schedules in the future that are not necessarily the same across an entire release-- say, for example, if we wanted different support schedules in different regions, we could add `'RHEL10 (North America)'` and `'RHEL10 (EMEA)'`.

app/models/katello/rhel_lifecycle_status.rb
```diff
@@ -0,0 +1,175 @@
...
+
+    def self.eos_date(eos_schedule: nil)
+      return nil unless eos_schedule
+      RHEL_EOS_SCHEDULES[eos_schedule]&.[]('extended_support') ||
+        RHEL_EOS_SCHEDULES[eos_schedule]&.[]('maintenance_support')
+    end
+
+    def self.warn_date(eos_schedule: nil)
+      return nil unless eos_schedule
+      end_of_support_date = eos_date(eos_schedule: eos_schedule)
```
lfu, line 112: I know RHEL 7 is the one should get the alert this round.
But the full support end date for RHEL 8 is May 31, 2024.
We could display the message when any one of the 3 end of support dates is approaching without too much efforts. What do you think?
` end_of_support =  RHEL_EOS_SCHEDULES[eos_schedule].select { |k, v| (Time.now.utc .. Time.now.utc + EOS_WARNING_THRESHOLD).cover?(v) }`

jeremylenz: Do we want to alert users when a system is transitioning from full support to maintenance support, etc?  I'm not sure I see how that would add value?

The only time a user actually needs to take action (update to a new OS major) is when approaching the end of support date, unless I'm not understanding something.

app/models/katello/concerns/operatingsystem_extensions.rb
lfu: How about `rhel_eos_schedule_index`?

app/models/katello/rhel_lifecycle_status.rb
lfu: Alright. Red Hat releases usually follow a ten-year life cycle. Then it is possible to pay for Extended Life Cycle Support (ELS).  Guess that is why we want the alert for RHEL 7 whose maintenance support ends  on June 30, 2024.
Looks like we just need to display the alert message for approaching maintenance support end date or ELS end date.

I don't see the eos approaching message for RHEL 7 yet.
image

jeremylenz: Okay, I can add a `maintenance_warn_date` to go alongside `eos_warn_date`. It will warn 1 year before, is that good?

One thing I wasn't sure about: A couple of the RHEL7 schedules have a "Maintenance support" end date, but no "Extended support" end date. I assume that means that _all_ support for those RHEL7 architectures ends at the Maintenance support end date. Is that correct?

app/models/katello/rhel_lifecycle_status.rb
jeremylenz: All of our other host statuses have an `UNKNOWN`. And for non-RHEL hosts, the status won't be one of those three, so it made sense to me to calculate that as `UNKNOWN`.

app/models/katello/rhel_lifecycle_status.rb
lfu: > Okay, I can add a `maintenance_warn_date` to go alongside `eos_warn_date`. It will warn 1 year before, is that good?

👍🏻 

> * Support ended (error)

Not sure this one should be an error or warning.

> One thing I wasn't sure about: A couple of the RHEL7 schedules have a "Maintenance support" end date, but no "Extended support" end date. I assume that means that _all_ support for those RHEL7 architectures ends at the Maintenance support end date. Is that correct?

I would agree with you.

app/models/katello/concerns/operatingsystem_extensions.rb
jeremylenz: @chris1984 do you happen to know what the strings are for RHEL6 and RHEL5? :smile:

app/models/katello/rhel_lifecycle_status.rb
jeremylenz: See my other comment.

app/models/katello/rhel_lifecycle_status.rb
jeremylenz: > Not sure this one should be an error or warning.

If "Support ended" is only a warning, there won't be any differentiation between "Approaching end of support" and "Support ended." For this reason I'd vote to keep it an error status.

(To be clear, error status just means it shows red on host statuses; it doesn't actually throw any errors.) Thoughts?

lfu: Error status is fine with me. Let's start with that.

app/models/katello/rhel_lifecycle_status.rb
```diff
@@ -0,0 +1,179 @@
...
+    MAINTENANCE_SUPPORT = 2
+    EXTENDED_SUPPORT = 3
+    APPROACHING_END_OF_SUPPORT = 4
+    SUPPORT_ENDED = 5
+
+    def self.end_of_day(date)
+      DateTime.parse(date.to_s).end_of_day.utc
+    end
+
+    RHEL_EOS_SCHEDULE_INDEXES = { # dates that each support category ends
```
lfu, line 14: This one should be `RHEL_EOS_SCHEDULE`.

app/models/katello/host/content_facet.rb
```diff
@@ -206,7 +206,7 @@ def calculate_and_import_applicability
         ::Katello::Applicability::ApplicableContentHelper.new(self, ::Katello::Erratum, bound_repos).calculate_and_import
         ::Katello::Applicability::ApplicableContentHelper.new(self, ::Katello::ModuleStream, bound_repos).calculate_and_import
         update_applicability_counts
-        self.update_errata_status
+        self.host&.refresh_statuses(::Katello::HostStatusManager::STATUSES)
```
lfu, line 209: Personally I don't think this change is appropriate. Why should an errata status update affect all other statuses?

jeremylenz: If we don't do it here, where should we update the host's RHEL lifecycle status? And how can we ensure that it's updated regularly?

I thought about adding some sort of cron schedule or something, but this seemed much easier since we're already doing a regular status update of errata.

lfu: How about Host status card? Could we call this method when the card is loaded?

jeremylenz: There are other places besides the host status card where we'd need updated information-- `hammer host status`, for one. :thinking:

app/models/katello/rhel_lifecycle_status.rb
jeremylenz: One more question: Is a host only covered by extended support if the customer has purchased extended support on it?  Currently we have no way to know if extended support has been purchased, so our host statuses must be based on the schedule dates only.

jeremylenz:
Added maintenance_warn_date so RHEL7 should now show `"Approaching end of maintenance support (2024-06-30)"`

jeremylenz:
fixed tests

app/models/katello/host/content_facet.rb
lfu: I thought the epic is for UI notification. 
We can add it for `hammer host status` if required.

lfu:
> Added maintenance_warn_date so RHEL7 should now show `"Approaching end of maintenance support (2024-06-30)"`

I saw this message in green color 😃

app/models/katello/rhel_lifecycle_status.rb
```diff
@@ -0,0 +1,210 @@
...
+      if between_dates?(maintenance_approach_date, maintenance_end_of_support_date)
+        return APPROACHING_END_OF_MAINTENANCE
+      end
+      return MAINTENANCE_SUPPORT if Date.today <= maintenance_end_of_support_date
+      if between_dates?(approach_date, end_of_support_date)
+        return APPROACHING_END_OF_SUPPORT
+      end
+      return EXTENDED_SUPPORT if Date.today <= end_of_support_date
+      return SUPPORT_ENDED
+    end
```
lfu, line 97: Is this clearer?
```
   def self.approaching_eos(eos_schedule_index: release)
     RHEL_EOS_SCHEDULE[eos_schedule_index].select { |k, v| (Time.now.utc .. Time.now.utc + EOS_WARNING_THRESHOLD).cover?(v) }
   end

    def self.to_status(operatingsystem: nil)
      #return UNKNOWN unless operatingsystem.is_a?(::Operatingsystem)   # this check should be in operatingsystem.rhel_eos_schedule_index
      release = operatingsystem.rhel_eos_schedule_index
      return UNKNOWN if release.nil? || RHEL_EOS_SCHEDULE[release].nil?

      approach = approaching_eos(eos_schedule_index: release)
      if approach.present?
        case approach.keys.first
        when 'maintenance_support'
          return APPROACHING_END_OF_MAINTENANCE 
        when 'extended_support'
          return APPROACHING_END_OF_SUPPORT 
        end
      end

      full_support_date = full_support_end_date(eos_schedule_index: release)
      end_of_support_date = eos_date(eos_schedule_index: release)
      maintenance_end_of_support_date = maintenance_support_end_date(eos_schedule_index: release)

      case DateTime.current
      when full_support_date-5.years..full_support_date
        return FULL_SUPPORT
      when full_support_date..maintenance_end_of_support_date
        return MAINTENANCE_SUPPORT
      when maintenance_end_of_support_date..end_of_support_date
        return EXTENDED_SUPPORT
      else
        return SUPPORT_ENDED
      end
    end
```

app/models/katello/rhel_lifecycle_status.rb
```diff
@@ -0,0 +1,210 @@
...
+        support_ended: SUPPORT_ENDED
+      }
+
+      map.default = UNKNOWN
+      map
+    end
+
+    def self.to_status(operatingsystem: nil)
+      return UNKNOWN unless operatingsystem.is_a?(::Operatingsystem)
+      release = operatingsystem.rhel_eos_schedule_index
```
lfu, line 79: This release schedule index is passed around a lot. 
Is it better to save it into an instance variable? Hopefully the code could be cleaner.

app/models/katello/rhel_lifecycle_status.rb
lfu: I think this is fine. We are just displaying the RHEL lifecyce status not their subscriptions.

app/models/katello/rhel_lifecycle_status.rb
jeremylenz: If it were an instance method, I'd agree. But this is a class method (`def self.`) and I wouldn't want to store the same eos_schedule_index for the entire class..

jeremylenz: I do have `def rhel_eos_schedule_index` as an instance method.

jeremylenz:
@lfu Made approaching maintenance a warning status (should be yellow now) and cleaned up `self.to_status`.

lfu (approved):
Nice job 👍🏻

app/models/katello/concerns/operatingsystem_extensions.rb
chris1984: @jeremylenz Should be good to go with 6 too:

```
irb(main):004:0> Operatingsystem.second
=> #<Redhat id: 6, major: "6", name: "RedHat", minor: "10", nameindicator: nil, created_at: "2023-09-01 13:51:18.161260000 +0000", updated_at: "2023-09-01 13:51:18.161260000 +0000", release_name: nil, type: "Redhat", description: nil, password_hash: [FILTERED], title: "RedHat 6.10">
```

jeremylenz: _narrator voice_  They were not good to go.

PR #10709, opened 2023-08-25
Refs #35573 - Improve UX when enabling Tracer

jeremylenz:

Detect if the `katello-host-tools-tracer` package can be installed on the host.

If it's available, and REX is working, then things should go smoothly when you click 'Enable Tracer.'  But if it's not available, one of two things is the case:

1. One or more of the prerequisites to install the package have not been met; _or_
2. All the prerequisites have been met, but the host hasn't yet uploaded a new package profile and recalculated the installed / available package list.

Because of no. 2 above, we're never 100% sure that clicking 'Enable Tracer' is going to fail. But we can _at least_ remind you about the prerequisites if we're showing that the package isn't yet available.

So, if the package isn't listed in the host's `yum_installable` packages, we show this:
image

And if it is, we skip all of that and show this:
image

Originally I wanted to do a lot of this for you automatically. But it turned out to be more complex than you might think, since we'd have to do it

* across different OS versions
* across different arches
* also, the repository names & labels are different upstream vs. downstream

So I settled on this compromise.

Test with two hosts, neither of which has katello-host-tools-tracer installed

* one which has the repo available and enabled
* one which does not

### DIFF
--- a/app/models/katello/host/content_facet.rb
+++ b/app/models/katello/host/content_facet.rb
@@ -8,6 +8,9 @@ module Katello
       HOST_TOOLS_PACKAGE_NAME = 'katello-host-tools'.freeze
       HOST_TOOLS_TRACER_PACKAGE_NAME = 'katello-host-tools-tracer'.freeze
       SUBSCRIPTION_MANAGER_PACKAGE_NAME = 'subscription-manager'.freeze
+      ALL_TRACER_PACKAGE_NAMES = [ "python-#{HOST_TOOLS_TRACER_PACKAGE_NAME}",
+                                   "python3-#{HOST_TOOLS_TRACER_PACKAGE_NAME}",
+                                   HOST_TOOLS_TRACER_PACKAGE_NAME ].freeze
 
       belongs_to :kickstart_repository, :class_name => "::Katello::Repository", :inverse_of => :kickstart_content_facets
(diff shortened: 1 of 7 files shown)


lfu:
Should this one be removed?  `Remote execution is enabled.`

What does `Foreman Client repository ` refer to? Should [this instruction](https://theforeman.org/plugins/katello/nightly/installation/clients.html) be followed here?

jeremylenz:
> Should this one be removed? `Remote execution is enabled.`

Remote execution must be enabled to install the Tracer package, since it gets installed via a REX job.

>
> What does `Foreman Client repository ` refer to?

It refers to this repository:
image

Downstream, the word "Foreman" will be branded to "Satellite."

> Should [this instruction](https://theforeman.org/plugins/katello/nightly/installation/clients.html) be followed here?

For upstream, yes I think that's correct.

jeremylenz:
> Since katello agent is removed, isn't REX the only option and by default?

Upstream I think you still must explicitly enable it.

lfu (approved):
Tested with RHEL 8. It looked good.

lfu:
please rebase the PR.

damoore044 (approved):
looks good! - pending checks after rebase

jeremylenz:
[test katello]

PR #10710, opened 2023-08-28
Refs #36706 - Remove remaining katello-agent code including Agent Events & Dynflow actions

jeremylenz:

Remove remaining katello-agent code, including

- katello-agent events from Event Daemon
- katello-agent Dynflow actions
- katello-agent Dispatcher and Dispatch Histories
- miscellaneous stragglers

Probably not enough

This removes Agent Events, but the other two event types handled by the Event Daemon (Katello Events and Candlepin Events) should be unaffected.

Run migrations
Verify that tests pass and the remaining tests make sense
Verify that basic functionality still works, particularly incremental updates
Verify that Katello Events and Candlepin Events still work properly


jeremylenz:
currently rebased on the PR that removes API endpoints; I'll rebase on master when that's in.

app/models/katello/ping.rb
```diff
@@ -143,13 +142,6 @@ def exception_watch(result)
         result
       end
 
-      # get package information for katello and its components
-      def packages
-        names = PACKAGES.join("|")
-        packages = `rpm -qa | egrep "#{names}"`
-        packages.split("\n").sort
-      end
```
jeremylenz, line 159: I couldn't see anywhere where this was used. :shrug:

katello.gemspec
```diff
@@ -37,7 +37,6 @@ Gem::Specification.new do |gem|
   gem.add_dependency "foreman_remote_execution", ">= 7.1.0"
   gem.add_dependency "dynflow", ">= 1.6.1"
   gem.add_dependency "activerecord-import"
-  gem.add_dependency "qpid_proton"
```
jeremylenz, line 40: oh hey, @Katello/packagers :innocent:

jturel (approved):
You got all the things I thought someone might miss like removing the DispatchHistory table and qpid gem. Nice!

jeremylenz:
rebased

parthaa:
have you thought about removing this https://github.com/Katello/katello/blob/master/app/controllers/katello/api/rhsm/candlepin_proxies_controller.rb#L189

jeremylenz:
> have you thought about removing this https://github.com/Katello/katello/blob/master/app/controllers/katello/api/rhsm/candlepin_proxies_controller.rb#L189

I'm a bit nervous about that one. Are we sure it's not also used by the subscription-manager plugin somewhere? It's in the rhsm_proxies_controller so it probably felt safer not to touch it. I could update the description..

I did do some spot checks just now, and I can't get sub-man to hit that endpoint (it doesn't with `subscription-manager repos` or `subscription-manager repo-override`) so maybe it is safe to remove.. thoughts?

parthaa:
> I did do some spot checks just now, and I can't get sub-man to hit that endpoint (it doesn't with `subscription-manager repos` or `subscription-manager repo-override`) so maybe it is safe to remove.. thoughts?

On 2nd thoughts i think it is used by katello-host-tools.
https://github.com/Katello/katello-host-tools/blob/master/src/katello/repos.py#L31

So unless we are removing dnf upload, I say we leave this route alone.

jeremylenz:
> So unless we are removing dnf upload, I say we leave this endpoint alone.

Sounds good to me. in which case, is this good to go now? :smile:

parthaa:
Need to remove this line https://github.com/Katello/katello/blob/master/engines/bastion_katello/lib/bastion_katello/engine.rb#L43

jeremylenz:
removed

engines/bastion_katello/lib/bastion_katello/engine.rb
```diff
@@ -40,9 +40,8 @@ class Engine < ::Rails::Engine
         :config_generator =>  lambda do
           { 'consumerCertRPM' => consumer_cert_rpm,
             'defaultDownloadPolicy' => !Foreman.in_rake? && db_migrated && Setting['default_download_policy'],
-            'katelloAgentPresent' => ::Katello.with_katello_agent?,
             'remoteExecutionPresent' => ::Katello.with_remote_execution?,
-            'hostToolingEnabled' => (::Katello.with_katello_agent? || ::Katello.with_remote_execution?) ? true : false
+            'hostToolingEnabled' => ::Katello.with_remote_execution? ? true : false
```
parthaa, line 44:
```suggestion
            'hostToolingEnabled' => ::Katello.with_remote_execution? 
```

ehelms:
I know you are already pretty far along, something to think about in the future for these kind of changes is to split them into two parts:

 1) Do the bare minimum to stop exposing whatever the feature is to users anymore
 2) Clean-up the code at some future date

This both can allow you to move faster by getting in the actual drop faster (as the code changes are much smaller, and less risky) and it can provide a fall-back mechanism for some amount of time if the change or timeline is risky.

I mentioned a similar thing here for the installer part as well -- https://github.com/theforeman/puppet-foreman_proxy_content/pull/455#issuecomment-1695734613

parthaa (approved):
ACKed
Verified that incremental updates does the right thing and always tries remote execution.

PR #10711, opened 2023-08-28
Fixes #36538 - Don't delete manifests referenced by tags or manifest lists

sjha4:
Ignore manifests during deletion that are referenced by docker tags or manifest lists
1. Sync a docker repo.
2. Go to repository > manifest list.
3. Select a few from the list and try deleting.
4. Deletion should ignore any manifests with tags associated. (ex: quay/busybox should ignore the only manifest and also fail saying "No docker manifests to delete after ignoring manifests with tags or manifest lists")
5. If your docker repo has more than one valid manifest selected, that doesn't contain tags or lists, the removal will continue and ignore manifests with tags or lists.


chris1984 (approved):
ACK, works great.

PR #10714, opened 2023-08-31
Fixes #36702 - store package counts after smart proxy sync

ianballou:
In draft form for now until I write some tests.

Adds the endpoints:

```
GET /katello/api/capsules/2/content/counts
POST /katello/api/capsules/2/content/update_counts
```

so that smart proxies can store their content counts.

Also adds calculation of content counts during smart proxy sync.

I was considering making the counts get calculated per-repository, but with how many api calls that would require, it seems simpler and potentially more efficient to just re-calc all the counts at once. The json columns shouldn't get too large, and I'm making the API calls to Pulp more efficient by setting `limit: 1` to grab only the number of content units for special cases like SRPMs and DockerManifestLists.

Sync many repository types to a smart proxy and try out the two new endpoints.

### DIFF
--- a/app/models/katello/concerns/smart_proxy_extensions.rb
+++ b/app/models/katello/concerns/smart_proxy_extensions.rb
@@ -121,6 +121,36 @@ module Katello
         SmartProxy.joins(:smart_proxy_alternate_content_sources).where('katello_smart_proxy_alternate_content_sources.smart_proxy_id' => self.id)
       end
 
+      def update_content_counts!
+        # {:content_view_versions=>{87=>{:repositories=>{1=>{:rpms=>98, :module_streams=>9898}}}}}
+        new_content_counts = { content_view_versions: {} }
+        smart_proxy_helper = ::Katello::SmartProxyHelper.new(self)
+        repos = smart_proxy_helper.repositories_available_to_capsule
+        return new_content_counts if repos.empty?
+
+        repos.each do |repo|
+          repo_mirror_service = repo.backend_service(self).with_mirror_adapter
+          repo_content_counts = repo_mirror_service.latest_content_counts
+          translated_counts = {}
+          repo_content_counts.each do |name, count|
+            count = count[:count]
+            # Some content units in Pulp have the same model
+            if name == 'rpm.package' && repo.content_counts['srpm'] > 0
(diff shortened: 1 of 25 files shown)


app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -121,6 +121,36 @@ def alternate_content_sources
         SmartProxy.joins(:smart_proxy_alternate_content_sources).where('katello_smart_proxy_alternate_content_sources.smart_proxy_id' => self.id)
       end
 
+      def update_content_counts!
```
sjha4, line 124: Wondering if this should be a separate task which is called as an async task at end of capsule sync and the update_counts API can also trigger this as a task since there are all the capsule calls involved. We got capsules on ⛵ to tend to.. 😸

ianballou: Sure I can do that :+1:

ianballou: Actually, this will be done at the end of the smart proxy sync since it's being run as part of `SyncCapsule`. It's run right before the container gateway data sync.

sjha4:
Code good looks and flow makes sense. Will test! 👍🏼

sjha4:
> Code good looks and flow makes sense. Will test! 👍🏼

Had to run it to figure the `UpdateContentCounts` action isn't in the commit..

ianballou:
I'll make a push soon to get this out of draft form.

ianballou:
One thing that might not be obvious -- I'm only translating the content counts for content that we'll show on the UI. Things like "package categories" aren't indexed into Katello, so I'm not bothering to translate them to Katello terms. However, I think it's good to leave them in in case it's useful.

sjha4 (commented):
Tested and works great 👍🏼 

Do you think we should add the content_counts field to katello/app/views/katello/api/v2/capsule_content/sync_status.json.rabl and return it via API as part of this PR?

ianballou:
> Do you think we should add the content_counts field to katello/app/views/katello/api/v2/capsule_content/sync_status.json.rabl and return it via API as part of this PR?

I can do that!

sjha4 (approved):
Ack.. 👍🏼

PR #10715, opened 2023-08-31
Fixes #36712 - Add true select all to errata legacy UI

chris1984:

* Adding a select all option in the errata legacy UI when you start to select errata on the table

prerrata

* It looks like Bastion does not keep track of records selected when switching pages in the pagination on multiple pages so this is as close to fixing it as I can. I can raise a Redmine issue to try and fix the other areas, but outside of the scope of this PR

* Sync some repos that have errata in them
* Goto content - Errata and select a few and see if the new select all works
* Change the tab to the next table page and make sure they are all selected still
* Click apply and see if it takes you to the next screen without an error

### DIFF
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/errata/errata.controller.js
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/errata/errata.controller.js
@@ -35,6 +35,7 @@ angular.module('Bastion.errata').controller('ErrataController',
         }
 
         nutupane = $scope.nutupane = new Nutupane(Erratum, params);
+        nutupane.enableSelectAllResults();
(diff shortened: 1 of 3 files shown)


Ron-Lavi:
Hey @chris1984, can you rebase?

Ron-Lavi (commented):
Hey @chris1984, can you rebase?

chris1984:
@Ron-Lavi this approach does not work, Jeremy tested it a while ago. It's not sending the params correctly when you get to the last page, so I gave up on it hence why it's so old.

chris1984:
I thought it would be an easy fix but it turned out to not be as easy as I thought it would be. With us getting rid of the old UI eventually, should we even bother to fix this?

PR #10697, opened 2023-08-15
Fixes #36610 - Add system purpose to activation key details

jeremylenz:

PR #10716, opened 2023-08-31
Fixes #36720 - Composable Expandable Table for Smart Proxy content

sjha4:

Refactor the smart proxy content table as expandable table to prep for more changes on the page in follow-up PRs.


jeremylenz:
I think this is the one where I was complaining about the expansion carat, right? I addressed something similar in https://github.com/Katello/katello/pull/10697/files

PR #10720, opened 2023-09-06
Fixes #36722 - restore Katello::Ping.packages

evgeni:

Brings back `Katello::Ping.packages` which was wrongly removed during katello-agent removal.

Instead of removing this, update the package list to not list qpid.

The data provided is still valuable, so instead of removing it, make it work again.

Go to `/about` on your Katello installation.
Without the patch, you get a 502, with the patch it works ;-)


jeremylenz (approved):
Thanks for cleaning up after me, @evgeni :)

ACK :+1:

ekohl:
I suspect this also broke 4.10.0-rc1: https://community.theforeman.org/t/katello-4-10-rpm-pipeline-3-failed/35046

PR #10722, opened 2023-09-06
Fixes 36726 - Non-admin users with the view_smart_proxies permission …

qcjames53:
…can now access capsules from the API.

Non admin users before this change could not access capsules using the `/katello/api/capsules/<id>` endpoint. This has been fixed and now requires the user to have the `view_smart_proxies` permission.

Users with this permission could previously see capsule content using the `/katello/api/capsules/` endpoint. This change brings the other endpoint into parity.

1. Launch a Katello instance with at least one capsule.
2. Create two non-admin users. Make a new role with the `view_smart_proxies` permission. Add this role to one of the two new users.
3. Check the `/capsules` endpoint with the admin user and the two new users. The user without the view permission should receive an error, while the other two should list the contents of all capsules.
```
curl -k -u <user>:<password> <address>/katello/api/capsules/
```
4. Check the `/capsules/<id>` endpoint for a specific capsule. The user without the view permission should receive an error, while the other two should list the contents of the capsule.
```
curl -k -u <user>:<password> <address>/katello/api/capsules/<id>
```


qcjames53:
Force pushed to trigger test re-run. Jenkins errored out :(

app/controllers/katello/api/v2/capsules_controller.rb
```diff
@@ -31,5 +33,13 @@ def resource_class
     def authorized
       User.current.allowed_to?(params.slice(:action, :id).merge(controller: 'api/v2/smart_proxies'))
     end
+
+    # By default, Foreman required users to be admin or have the "view_capsule" permission.
```
ianballou, line 37: Maybe something like this would be more clear:
```suggestion
    # Without this method, Foreman required non-admin users to have the non-existent "view_capsule" permission.
```

ianballou (commented):
Looking good, just need to do a test.

ianballou (commented):
Works for me! It would be good to have a test for this. The test could just change `User.current` to a non-admin, `GET` the smart proxy, and then ensure that the response is a 200 instead of a 404.

ianballou (approved):
Looking great! Thanks for improving the tests!

PR #10725, opened 2023-09-07
Fixes #36713 - Fix errata workflow in host collections

chris1984:

* I got to trigger happy with removing stuff for katello-agent and removed too much AngularJS stuff, so the errata form never got submitted.
* I was sending an array instead of a hash so once I fixed the errata form, we got a `deep symbolize_keys` error on an array, so corrected my mistake
* Fixed up a test that I broke when I switched to an array instead of a hash

* Made sure we transitioned to the rex job for all three buttons
* I didn't test the actual rex since I don't have a client setup with it

* Check out PR
* Have a client with REX working and some errata that are avail to the host
* Add the host to a host collection
* Apply errata through the host collection bulk window and see if the buttons work


lfu:
Install Errarta from Host collection worked.

image

lfu (approved):

PR #10727, opened 2023-09-07
Fixes #36732 - Add a RHEL Lifecycle Status host column

parthaa:
Adds a new  RHEL Lifecycle Status host column in the Host Index page

- Register rhel host
- Go to Hosts index
- Select Manage Columns and Expand Content
- Notice the new RHEL Lifecycle Status
- Select that
- The Lifecycle status should now show up.


jeremylenz (approved):
Works great! Thanks @parthaa 

APJ :+1:

PR #10729, opened 2023-09-12
Fixes #36737 - use timeouts also for smart-proxy

m-bucher:
Apply the content-sync timeout-values also for smart-proxy syncs.


quba42:
@ianballou What do you think about this approach (setting the same sync timeouts for smart proxy and remote syncs)?

I think the current situation where there is no way of changing the timeouts on the proxies is certainly not great.

ianballou (approved):
Looks fine to me!

PR #10731, opened 2023-09-12
Fixes #36739 - Product without any repo is added to a Sync Plan regardless the error message

sjha4:
1. Create an empty product.
2. Try adding empty product to sync plan
3. You should see an error message and product should not get added to sync plan.

Before PR: You'll see an error but the sync plan gets updated despite the error message.

### DIFF
--- a/app/models/katello/product.rb
+++ b/app/models/katello/product.rb
@@ -34,6 +34,7 @@ module Katello
     validates_lengths_from_database :except => [:label]
     validates :provider_id, :presence => true
     validate :ensure_provider_type_matches_id
+    validate :ensure_no_sync_plans_on_empty_product
     validates_with Validators::KatelloNameFormatValidator, :attributes => :name
     validates_with Validators::KatelloLabelFormatValidator, :attributes => :label
     validates_with Validators::ProductUniqueAttributeValidator, :attributes => :name
@@ -261,6 +262,13 @@ module Katello
       return if errors[:provider_id].any?
       if provider.redhat_provider? && ::Katello::Glue::Candlepin::Product.custom_product_id?(cp_id)
         errors.add(:base, _("Cannot associate a Red Hat provider with a custom product"))
+      end
+    end
+
+    def ensure_no_sync_plans_on_empty_product
+      return if sync_plan_id.nil?
+      if redhat? && !enabled?
+        errors.add(:base, _("Cannot add disabled Red Hat product %s to sync plan!") % name)
       end
     end
   end
--- a/app/controllers/katello/api/v2/sync_plans_controller.rb
+++ b/app/controllers/katello/api/v2/sync_plans_controller.rb
@@ -94,8 +94,9 @@ module Katello
     param :product_ids, Array, :desc => N_("List of product ids to add to the sync plan"), :required => true
     def add_products
       products = ::Katello::Product.where(:id => params[:product_ids]).editable
+      disabled_products = products.select { |p| p.redhat? && !p.enabled? }
+      fail _("Cannot add disabled products to sync plan!") if disabled_products.present?
       @sync_plan.product_ids = (@sync_plan.product_ids + products.collect { |p| p.id }).uniq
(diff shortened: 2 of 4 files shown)


app/models/katello/product.rb
```diff
@@ -263,5 +264,12 @@ def ensure_provider_type_matches_id
         errors.add(:base, _("Cannot associate a Red Hat provider with a custom product"))
       end
     end
+
+    def ensure_no_sync_plans_on_empty_product
+      return if sync_plan_id.nil?
+      if repositories.empty?
+        fail _("Cannot add empty product %s to sync plan!") % name
```
jeremylenz, line 271: Why not `errors.add`?

I feel like if this just raises an error (`fail`) rather than adding errors, it's not a validation and shouldn't be triggered by `validate`.

jeremylenz (commented):
Interesting that this isn't a many-to-many relationship like we were assuming the other day, but the same sort of thing was happening with the skipping validations.

app/models/katello/product.rb
jeremylenz: > can not be overridden given we don't have a way to catch and customize the message

You can pass a custom message to `errors.add`..

jeremylenz: also, if you need the message to be completely custom you can `errors.add(:base` instead of `errors.add(:sync_plan`

sjha4:
Updated the PR to look at disabled products instead of 'empty' products. That is what it should've been doing.

app/models/katello/product.rb
sjha4: So the errors.add(:base,..) makes the product record invalid and fails to save sync_plan but the error thrown is default activerecord error from here: https://apidock.com/rails/v6.0.0/ActiveRecord/Associations/CollectionAssociation/replace_records

Screenshot from 2023-09-13 15-20-21

jeremylenz: Ah.. 

I still think `errors.add` is a more correct way to handle it.

I see it raises a `RecordNotSaved`. Maybe we could rescue that and re-raise it just to make the message better..

app/models/katello/product.rb
```diff
@@ -107,7 +108,7 @@ def repos(env, content_view = nil, include_feedless = true)
     end
 
     def enabled?
-      self.repositories.present?
+      self.redhat? && self.repositories.present?
```
jeremylenz, line 111: Wouldn't this prevent you from adding custom products to a sync plan?

app/models/katello/product.rb
jeremylenz: Also,

Would it help to add `validates_associated :products` to `sync_plan.rb`? Seems that should have been there already. And you can add a `message:` there.

sjha4:
Updated the PR to get a mix of validations for API and model.

When using the API, the controller catches the default error and raises the custom error. That covers the UI/Hammer workflows.

When in console, doing a Product.sync_plan_id = $Id will mark product invalid as expected. The save! will throw the activerecord error message instead of the custom error.

app/models/katello/product.rb
```diff
@@ -263,5 +264,13 @@ def ensure_provider_type_matches_id
         errors.add(:base, _("Cannot associate a Red Hat provider with a custom product"))
       end
     end
+
+    def ensure_no_sync_plans_on_empty_product
+      return if sync_plan_id.nil?
+      if redhat? && !enabled?
+        #raise ActiveRecord::RecordNotSaved, "Cannot add disabled redhat product #{name} to sync plan!"
+        errors.add(:base, _("Cannot add disabled redhat product %s to sync plan!") % name)
```
jeremylenz, line 272:
```suggestion
        errors.add(:base, _("Cannot add disabled Red Hat product %s to sync plan!") % name)
```

jeremylenz (commented):
Looking good, just one final comment :)

jeremylenz (approved):
Works great!

APJ :+1:

PR #10734, opened 2023-09-14
Fixes #36748 - fix usage of deb copy api

m-bucher:
Pulp-Deb does not support dependency-solving.

Also we should use the correct name of the copy-api class.

### DIFF
--- a/app/services/katello/pulp3/repository/apt.rb
+++ b/app/services/katello/pulp3/repository/apt.rb
@@ -73,18 +73,27 @@ module Katello
           "/pulp/deb/#{repo.relative_path}/".sub('//', '/')
         end
 
-        def multi_copy_units(repo_id_map, dependency_solving)
+        def multi_copy_units(repo_id_map, _dependency_solving)
           tasks = []
 
           if repo_id_map.values.pluck(:content_unit_hrefs).flatten.any?
             data = PulpDebClient::Copy.new
-            data.dependency_solving = dependency_solving
+            data.dependency_solving = false
             data.config = []
             repo_id_map.each do |source_repo_ids, dest_repo_id_map|
               dest_repo = ::Katello::Repository.find(dest_repo_id_map[:dest_repo])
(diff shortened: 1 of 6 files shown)


m-bucher:
[test_katello]

m-bucher:
[test katello]

app/services/katello/pulp3/repository/apt.rb
```diff
@@ -203,9 +212,9 @@ def add_debs(source_repo_ids, filters, filter_list_map)
           filter_list_map
         end
 
-        def copy_content_from_mapping(repo_id_map, options = {})
+        def copy_content_from_mapping(repo_id_map, _options = {})
```
ianballou, line 215: I don't think `copy_content_from_mapping` will even run for Debian content, so these methods can likely be taken out entirely. If you trace back references to `copy_content_from_mapping`, you'll eventually get here: https://github.com/Katello/katello/blob/07d11cd10d3edbab9f1f73c66494b3ff1b53ca04/app/lib/actions/katello/content_view/publish.rb#L62

That shows that we only do the whole "multi copy" workflow for yum content anyway.

This wasn't always the case, but it has been since Katello 4.3.

ianballou: Although, if dep solving is something that is a TODO for debian content, then I'd recommend keeping these methods of course.

m-bucher: I believe we will need it eventually for the IncrementalUpdate to work : https://github.com/Katello/katello/pull/10847

Dependency Solving for debian is currently out-of-scope and due to various reasons like lacking support for efficient handling (sorting and comparing) of debian version within PostgreSQL.
This is also the reason I added a PR to make sure dep-solving with debs is never even tried :smiling_imp: : https://github.com/Katello/katello/pull/10734

ianballou (approved):
Looking fine to me! I haven't tested it since the deb incremental update bits aren't there, will do so in the next PR.

PR #10737, opened 2023-09-14
Fixes #36736 - Correct memory fact on HW properties card

chris1984:

* We had the wrong param to grab the RAM on the HW properties card so wrong info was being shown
* Corrected to the right value
* This came out of this support thread https://community.theforeman.org/t/hosts-ram-details-buggy/35050/5?u=cintrix84

* N/A

* Register a client with 4GB of ram and it should show 5GB (might work with libvirt if not I can give you a test client and test devel box to try it against)
* Apply PR
* Refresh the page and it should show the correct amount.

### DIFF
--- a/webpack/components/extensions/HostDetails/DetailsTabCards/HwPropertiesCard.js
+++ b/webpack/components/extensions/HostDetails/DetailsTabCards/HwPropertiesCard.js
@@ -10,6 +10,7 @@ import {
   Text,
   TextVariants,
 } from '@patternfly/react-core';
+import { number_to_human_size as NumberToHumanSize } from 'number_helpers';
 import CardTemplate from 'foremanReact/components/HostDetails/Templates/CardItem/CardTemplate';
 import { TranslatedPlural } from '../../../Table/components/TranslatedPlural';
 import { hostIsNotRegistered } from '../hostDetailsHelpers';
@@ -43,7 +44,7 @@ const HwPropertiesCard = ({ isExpandedGlobal, hostDetails }) => {
   const coreSocket = facts?.['cpu::core(s)_per_socket'];
   const reportedFacts = propsToCamelCase(hostDetails?.reported_data || {});
   const totalDisks = reportedFacts?.disksTotal;
-  const memory = facts?.['dmi::memory::maximum_capacity'];
+  const memory = facts?.['memory::memtotal'];
 
   return (
     <CardTemplate
@@ -71,7 +72,9 @@ const HwPropertiesCard = ({ isExpandedGlobal, hostDetails }) => {
         </DescriptionListGroup>
         <DescriptionListGroup>
           <DescriptionListTerm>{__('RAM')}</DescriptionListTerm>
-          <DescriptionListDescription>{memory}</DescriptionListDescription>
+          <DescriptionListDescription>{NumberToHumanSize(memory*1024, {
+          strip_insignificant_zeros: true, precision: 2
+        })}</DescriptionListDescription>
         </DescriptionListGroup>
         <DescriptionListGroup>
           <HostDisks totalDisks={totalDisks} />
@@ -89,7 +92,7 @@ HwPropertiesCard.propTypes = {
       cpuCount: PropTypes.number,
       cpuSockets: PropTypes.number,
       coreSocket: PropTypes.number,
-      memory: PropTypes.string,
+      memory: PropTypes.number,
     }),
     reported_data: PropTypes.shape({
       totalDisks: PropTypes.number,


webpack/components/extensions/HostDetails/DetailsTabCards/HwPropertiesCard.js
```diff
@@ -43,7 +43,7 @@ const HwPropertiesCard = ({ isExpandedGlobal, hostDetails }) => {
   const coreSocket = facts?.['cpu::core(s)_per_socket'];
   const reportedFacts = propsToCamelCase(hostDetails?.reported_data || {});
   const totalDisks = reportedFacts?.disksTotal;
-  const memory = facts?.['dmi::memory::maximum_capacity'];
+  const memory = facts?.['dmi::memory::size'];
```
sbernhard, line 46: I think, this is wrong and should stay at maximum_capacity:

sub-man facts:
image


puppet facts:

image

chris1984: What about having a max capacity field and then a size field?

chris1984: If the user does not have puppet facts, then I guess we just only show max capacity?

sbernhard: the question is, is it about "hardware properties" as the headline describes or hardware resources the OS can handle?
There are some other facts, e.g. from subscription manager.
Maybe just use what is already there:
https://github.com/theforeman/foreman/blob/31956e5267cc74b7e06eec73042a24505ba45c6f/app/services/katello/rhsm_fact_parser.rb#L107

or 
https://github.com/Katello/katello/blob/506c5282e72d1c4dcbd5a579d189641b0ec84f4b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/details/content-host-details-info.controller.js#L159

this is what users expect I guess.

chris1984:
@sbernhard fixed, let me know if this looks better?
prmemory

webpack/components/extensions/HostDetails/DetailsTabCards/HwPropertiesCard.js
```diff
@@ -43,7 +44,7 @@ const HwPropertiesCard = ({ isExpandedGlobal, hostDetails }) => {
   const coreSocket = facts?.['cpu::core(s)_per_socket'];
```
ekohl, line 44: I think this is `reportedFacts?.cores`. I can't comment on it, but above there's also `reportedFacts?.sockets`

ekohl (changes requested):
If this comes from facts, why isn't this in Foreman itself? And everything should be stored as reported_data (see https://github.com/theforeman/foreman/blob/develop/app/models/host_facets/reported_data_facet.rb). Direct use of facts is something we should avoid.

chris1984:
> If this comes from facts, why isn't this in Foreman itself? And everything should be stored as reported_data (see https://github.com/theforeman/foreman/blob/develop/app/models/host_facets/reported_data_facet.rb). Direct use of facts is something we should avoid.

So this would need a pr in Foreman then to have that fact come as a reported value?

ekohl:
IMHO yes, though I'd even say that this entire card should be dropped from Katello and implemented in Foreman itself. I see no reason why it would be Katello-specific, except that it uses RHSM facts directly instead of normalizing.

webpack/components/extensions/HostDetails/DetailsTabCards/HwPropertiesCard.js
```diff
@@ -71,7 +72,9 @@ const HwPropertiesCard = ({ isExpandedGlobal, hostDetails }) => {
         </DescriptionListGroup>
         <DescriptionListGroup>
           <DescriptionListTerm>{__('RAM')}</DescriptionListTerm>
-          <DescriptionListDescription>{memory}</DescriptionListDescription>
+          <DescriptionListDescription>{NumberToHumanSize(memory*1024, {
+          strip_insignificant_zeros: true, precision: 2
+        })}</DescriptionListDescription>
         </DescriptionListGroup>
         <DescriptionListGroup>
           <HostDisks totalDisks={totalDisks} />
```
ekohl, line 80: I just noticed this on the screenshot. This value is the total number of bytes of all disks combined, not the number of disks.

ekohl (changes requested):

webpack/components/extensions/HostDetails/DetailsTabCards/HwPropertiesCard.js
chris1984: I have that fixed in a separate pr

PR #10738, opened 2023-09-14
Fixes #36731 - Use facts, not operatingsystem name, to identify RHEL hosts

jeremylenz:

When determining the host status for RHEL lifecycle, do what's described in the PR title.

We now rely on sub-man facts instead of operatingsystem name, which is a bit more reliable. These facts won't be present for "unmanaged" hosts, but that's okay because we only care about managed hosts.

In rails console, run
```
Host.all.each(&:refresh_statuses)
```

Then check all host statuses and make sure they are still as expected.

### DIFF
--- a/app/models/katello/concerns/host_managed_extensions.rb
+++ b/app/models/katello/concerns/host_managed_extensions.rb
@@ -108,8 +108,6 @@ module Katello
         delegate :content_source_id, :single_content_view, :single_lifecycle_environment, :default_environment?, :single_content_view_environment?, :multi_content_view_environment?, :kickstart_repository_id, :bound_repositories,
           :installable_errata, :installable_rpms, to: :content_facet, allow_nil: true
 
-        delegate :rhel_eos_schedule_index, to: :operatingsystem, allow_nil: true
-
         has_many :content_view_environment_content_facets, through: :content_facet, class_name: 'Katello::ContentViewEnvironmentContentFacet'
         has_many :content_view_environments, through: :content_view_environment_content_facets
         has_many :content_views, through: :content_view_environments
@@ -522,6 +520,33 @@ module Katello
       def traces_helpers(search:)
         traces = host_traces.selectable.search_for(search)
         ::Katello::HostTracer.helpers_for(traces)
+      end
+
+      def probably_rhel?
+        # Get the os name from sub-man facts rather than operatingsystem. This is
+        # less likely to have been changed by the user.
+        os_name = facts['distribution::name']
+        # if this fact isn't there, we can ignore it because the host is not "managed"
+        os_name.present? && os_name.start_with?('Red Hat Enterprise Linux')
+      end
+
+      def rhel_eos_schedule_index
+        return nil unless probably_rhel?
+        major = operatingsystem&.major
+        arch_name = architecture&.name
+        return "RHEL#{major}" unless major == "7"
+        case arch_name
+        when "x86_64", nil
+          "RHEL7"
+        when "ppc64le"
+          "RHEL7 (POWER9)"
+        when "aarch64"
+          "RHEL7 (ARM)"
+        when "s390x"
+          "RHEL7 (System z (Structure A))"
+        else
+          "RHEL#{major}"
+        end
       end
 
       def package_names_for_job_template(action:, search:, versions: nil)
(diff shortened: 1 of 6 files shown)


app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -524,6 +522,33 @@ def traces_helpers(search:)
...
+      end
+
+      def rhel_eos_schedule_index
+        return nil unless probably_rhel?
+        major = operatingsystem&.major
+        arch_name = architecture&.name
+        return "RHEL#{major}" unless major == "7"
+        case arch_name
+        when "x86_64", nil
+          "RHEL7"
```
ekohl, line 540: This is already captured in the `else` case
```suggestion
```

app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -524,6 +522,33 @@ def traces_helpers(search:)
...
+        os_name = facts['distribution::name']
+        # if this fact isn't there, we can ignore it because the host is not "managed"
+        os_name.present? && os_name.start_with?('Red Hat Enterprise Linux')
+      end
+
+      def rhel_eos_schedule_index
+        return nil unless probably_rhel?
+        major = operatingsystem&.major
+        arch_name = architecture&.name
+        return "RHEL#{major}" unless major == "7"
```
ekohl, line 537: I'm not 100% sure if `architecture` triggers a foreign key load, but I'd swap the order just in case.
```suggestion
        return "RHEL#{major}" unless major == "7"
        arch_name = architecture&.name
```

app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -524,6 +522,33 @@ def traces_helpers(search:)
...
+        # Get the os name from sub-man facts rather than operatingsystem. This is
+        # less likely to have been changed by the user.
+        os_name = facts['distribution::name']
+        # if this fact isn't there, we can ignore it because the host is not "managed"
+        os_name.present? && os_name.start_with?('Red Hat Enterprise Linux')
+      end
+
+      def rhel_eos_schedule_index
+        return nil unless probably_rhel?
+        major = operatingsystem&.major
```
ekohl, line 535: What if `major` is `nil`? You'll end up with just `RHEL` as the name. If `operatingsystem` can't be `nil` then just `operatingsystem.major` should suffice, right?

ekohl (commented):
Some context: there is a setting `ignore_facts_for_operatingsystem` which can be used to either update the OS based on facts or disable it. The default is `false`, so (double negative) the default is to update the OS based on facts.

Some users don't like that syncing very much, so they disable it. There has been discussion about always storing it in a reported data facet. Then you can, even if you disable syncing, always easily use the parsed data. The idea being that application code should never use the raw facts but always handled processed data.

My biggest worry is cost. I think loading facts is pretty expensive, especially if you have a lot of them.

Perhaps this is a good excuse to move forward storing the OS as a reported data facet and then consume it here.

jeremylenz:
> My biggest worry is cost. I think loading facts is pretty expensive, especially if you have a lot of them.

I've been noticing this as well, as I'm also working on https://github.com/theforeman/foreman/pull/9819

Once that's merged I can update this to take advantage of it.

Also, this particular use case of facts shouldn't be _too_ bad, since the status is only updated on host check-in, and will occur in a staggered fashion and not for all hosts at once.

jeremylenz:
Updated to take advantage of https://github.com/theforeman/foreman/pull/9819 which is now merged.

app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -524,6 +522,31 @@ def traces_helpers(search:)
         ::Katello::HostTracer.helpers_for(traces)
       end
 
+      def probably_rhel?
+        # Get the os name from sub-man facts rather than operatingsystem. This is
+        # less likely to have been changed by the user.
+        os_name, = facts('distribution::name').values # only query for that one fact, then get its value
```
jeremylenz, line 528: I'm using `facts('distribution::name')` here instead of `facts['distribution::name']` because this guarantees that the SQL query will only retrieve that one fact, not all facts for the host.

With the comma, this line is equivalent to 

```
os_name = facts('distribution::name').values.first
```

but avoids the need for safe navigation (`&.`).

lfu (approved):
LGTM 👍🏻

lfu:
Would you sqash your commits?

jeremylenz:
@lfu squashed and merged, thanks! :+1:

PR #8198, opened 2019-07-05
Fixes #27241 - duplicate content view CCV message is misleading

ianballou:
The mention of filters is a bit misleading here since it can be interpreted that the CCV undoes the filtering.

PR #10741, opened 2023-09-15
Fixes #36492 - duplicate content warning for composite CVs for non-docker repos

sjha4:
Adds a notice to composite CVs publish screen if component CV versions have the same repositories added.
Screenshot from 2023-09-18 13-37-43

1, Create 2 CV versions with same repo
2.Add both to a composite CV
3. Publish. You should see a notice explaining duplicate content will be merged into the Composite CV version. The text is copied from earlier: https://github.com/Katello/katello/pull/8198/files

### DIFF
--- a/webpack/scenes/ContentViews/Publish/CVPublishForm.js
+++ b/webpack/scenes/ContentViews/Publish/CVPublishForm.js
@@ -17,7 +17,10 @@ const CVPublishForm = ({
   description,
   setDescription,
   details: {
-    name, composite, next_version: nextVersion, needs_publish: needsPublish,
+    name, composite,
+    next_version: nextVersion,
+    needs_publish: needsPublish,
+    duplicate_repositories_to_publish: duplicateRepos,
   },
   userCheckedItems,
   setUserCheckedItems,
@@ -27,6 +30,7 @@ const CVPublishForm = ({
 }) => {
   const [alertDismissed, setAlertDismissed] = useState(false);
   const [needsPublishAlertDismissed, setNeedsPublishAlertDismissed] = useState(false);
+  const [duplicateReposAlertDismissed, setDuplicateReposAlertDismissed] = useState(false);
   const needsPublishLocal = useSelector(state => selectCVNeedsPublish(state));
 
   const checkPromote = (checked) => {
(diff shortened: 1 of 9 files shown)


sjha4:
Test failures expected. Will update those.

sjha4:
Updated tests..Should be green..

webpack/scenes/ContentViews/Publish/CVPublishForm.js
```diff
@@ -61,6 +65,24 @@ const CVPublishForm = ({
...
                 </Alert>)
             }
+            {!duplicateReposAlertDismissed && composite &&
+                (duplicateRepos !== null || duplicateRepos.length > 0) &&
+                (
+                <Alert
+                  ouiaId="duplicate-repos-alert"
+                  variant="info"
+                  isInline
+                  title={__('Duplicate repositories in component content views')}
```
jeremylenz, line 75:
```suggestion
                  title={__('Duplicate repositories in content view versions')}
```

webpack/scenes/ContentViews/Publish/CVPublishForm.js
```diff
@@ -61,6 +65,24 @@ const CVPublishForm = ({
...
+                  actionClose={
+                    <AlertActionCloseButton
+                      onClose={() => setDuplicateReposAlertDismissed(true)}
+                    />
+                        }
+                  style={{ marginBottom: '24px' }}
+                >
+                  <TextContent>{__('Duplicate repositories between the component content view versions will merge, resulting in a composite content view with all the content that exists among the duplicates.')}</TextContent>
```
jeremylenz, line 83:
```suggestion
                  <TextContent>{__('Repositories in common between the content view versions will merge, resulting in a composite content view with all the content that exists among the duplicates.')}</TextContent>
```

jeremylenz (changes requested):
We don't use the term "component content views" any more..

webpack/scenes/ContentViews/Publish/CVPublishForm.js
jeremylenz: What do we mean by "all the content that exists among the duplicates?" Does this mean that if one CV version has something filtered out in a repository, but the other doesn't, I'll get the unfiltered content?

sjha4: Ya..So a union of content in the 2 repository versions between the 2 CVs..

sjha4: The message comes from the old UI which was here: https://github.com/Katello/katello/pull/8198/files

jeremylenz: Hmm.. How about
```
The composite content view will contain a combination of all the repositories from each of the content view versions.
```

sjha4: The composite content view will contain a combination of _all the content_ from the repositories from each of the content view versions.

jeremylenz: We decided on
```
Repositories common to the selected content view versions will merge, resulting in a composite content view that is a union of all content from each of the content view versions.
``

sjha4: Updated to the freshly baked text..

jeremylenz (approved):
Works as expected :+1: 

APJ

PR #10744, opened 2023-09-19
Fixes #35713 - Host details tab for Debian packages

nadjaheitmann:
Host details tab for Debian packages

Try viewing, installing, removing and updating deb packages for hosts.

For code review: There are actually a lot of code pieces that were taken from the PackagesTab. Any re-structuring of the code needed?

What is missing:

- [x] Tests

### DIFF
--- a/app/controllers/katello/api/v2/debs_controller.rb
+++ b/app/controllers/katello/api/v2/debs_controller.rb
@@ -88,9 +88,13 @@ module Katello
     def custom_index_relation(collection)
       applicable = ::Foreman::Cast.to_bool(params[:packages_restrict_applicable]) || params[:host_id]
       upgradable = ::Foreman::Cast.to_bool(params[:packages_restrict_upgradable])
+      not_installed = ::Foreman::Cast.to_bool(params[:packages_restrict_not_installed])
 
       if upgradable
         collection = collection.installable_for_hosts(@hosts)
+      elsif not_installed && params[:host_id]
+        host = @hosts.first
+        collection = Katello::Deb.apt_installable_for_host(host)
       elsif applicable
         collection = collection.applicable_to_hosts(@hosts)
       end
--- a/app/models/katello/concerns/host_managed_extensions.rb
+++ b/app/models/katello/concerns/host_managed_extensions.rb
@@ -534,6 +534,14 @@ module Katello
       end
 
       def package_names_for_job_template(action:, search:, versions: nil)
+        if self.operatingsystem.family == 'Debian'
+          deb_names_for_job_template(action, search)
+        else
+          yum_names_for_job_template(action, search, versions)
+        end
+      end
+
+      def yum_names_for_job_template(action:, search:, versions: nil)
         actions = %w(install remove update).freeze
         case action
         when 'install'
@@ -557,6 +565,24 @@ module Katello
           pkg_name_archs.map { |name, arch| versions_by_name_arch[[name, arch]] || upgrades[[name, arch]]&.first&.nvra }.compact
         else
           fail ::Foreman::Exception.new(N_("package_names_for_job_template: Action must be one of %s"), actions.join(', '))
+        end
+      end
+
+      def deb_names_for_job_template(action:, search:)
+        actions = %w(install remove update).freeze
+        case action
+        when 'install'
+          ::Katello::Deb.apt_installable_for_host(self).search_for(search).distinct.pluck(:name)
+        when 'remove'
+          return [] if search.empty?
+
+          installed_debs.search_for(search).distinct.pluck(:name)
+        when 'update'
+          return [] if search.empty?
+
+          installed_debs.search_for(search).distinct.pluck(:name)
+        else
+          fail ::Foreman::Exception.new(N_("deb_names_for_job_template: Action must be one of %s"), actions.join(', '))
         end
       end
 
--- a/app/models/katello/deb.rb
+++ b/app/models/katello/deb.rb
@@ -10,6 +10,7 @@ module Katello
              :dependent => :destroy, :inverse_of => :deb
     has_many :content_facets, :through => :content_facet_applicable_debs, :class_name => "Katello::Host::ContentFacet"
 
+    scoped_search :on => :id, :complete_value => true
     scoped_search :on => :name, :complete_value => true
     scoped_search :on => :version, :complete_value => true
     scoped_search :on => :architecture, :complete_value => true
@@ -93,6 +94,13 @@ module Katello
     def self.applicable_to_hosts(hosts)
       self.joins(:content_facets).
         where("#{Katello::Host::ContentFacet.table_name}.host_id" => hosts).distinct
+    end
+
+    # Return deb packages that are not installed on a host, but could be installed
+    # the word 'installable' has a different meaning here than elsewhere
+    def self.apt_installable_for_host(host)
+      repos = host.content_facet.bound_repositories.pluck(:id)
+      Katello::Deb.in_repositories(repos).where.not(name: host.installed_debs.pluck(:name)).order(:name)
     end
 
     def self.latest(_relation)
--- a/webpack/components/extensions/HostDetails/Tabs/DebsTab/DebInstallModal.scss
+++ b/webpack/components/extensions/HostDetails/Tabs/DebsTab/DebInstallModal.scss
@@ -1,0 +1,3 @@
+#deb-install-modal .pf-c-input-group input {
+  min-width: 15em;
+}
--- a/webpack/components/extensions/HostDetails/Tabs/DebsTab/DebsTab.js
+++ b/webpack/components/extensions/HostDetails/Tabs/DebsTab/DebsTab.js
@@ -1,5 +1,5 @@
 import React, { useCallback, useState, useRef } from 'react';
-import { useSelector, useDispatch } from 'react-redux';
+import { useSelector } from 'react-redux';
 import {
   ActionList,
   ActionListItem,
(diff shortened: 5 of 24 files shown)


nadjaheitmann:
Hey @lfu , I saw that you created most of the host details packages tab and I am trying to reuse your code for Debian packages, but I got stuck at the tests. Do you have any idea what needs adjustment? Any hints are highly appreciated!

nadjaheitmann:
> Hey @lfu , I saw that you created most of the host details packages tab and I am trying to reuse your code for Debian packages, but I got stuck at the tests. Do you have any idea what needs adjustment? Any hints are highly appreciated!

Or maybe @jturel ? Not sure, who would be the right person to ping.

lfu:
cc @jeremylenz

jeremylenz:
```
Error: Error: Nock: No match for request {
```

means the test is making an API request that wasn't planned for with a Nock instance. Each API request in a test must have a corresponding scope that matches the request, and then an `assertNockRequest` at the end of the test. You can run the tests with

`DEBUG=nock.* npx jest <filename>`

to see how Nock is doing its matching.  Probably some adjustment of the nock `.query` is needed.

nadjaheitmann:
Thanks @jeremylenz That debug output helped me a lot!

app/controllers/katello/api/v2/debs_controller.rb
```diff
@@ -88,9 +88,13 @@ def available_for_content_view_version(version)
     def custom_index_relation(collection)
       applicable = ::Foreman::Cast.to_bool(params[:packages_restrict_applicable]) || params[:host_id]
       upgradable = ::Foreman::Cast.to_bool(params[:packages_restrict_upgradable])
+      not_installed = ::Foreman::Cast.to_bool(params[:packages_restrict_not_installed])
 
       if upgradable
         collection = collection.installable_for_hosts(@hosts)
+      elsif not_installed && params[:host_id]
+        host = @hosts.first
+        collection = Katello::Deb.deb_installable_for_host(host)
```
jeremylenz, line 97: Wouldn't you want to run `deb_installable_for_host` on `Host.where(id: params[:host_id]).first` and not just `@hosts.first`?

webpack/components/extensions/HostDetails/Tabs/DebsTab/DebsTab.js
```diff
@@ -0,0 +1,597 @@
...
+} from '@patternfly/react-core';
+import { TableVariant, Thead, Tbody, Tr, Th, Td, TableText } from '@patternfly/react-table';
+import PropTypes from 'prop-types';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { selectAPIResponse } from 'foremanReact/redux/API/APISelectors';
+
+import { urlBuilder } from 'foremanReact/common/urlHelpers';
+import SelectableDropdown from '../../../../SelectableDropdown';
+import TableWrapper from '../../../../../components/Table/TableWrapper';
+import { useBulkSelect, useTableSort, useUrlParams, useSet } from '../../../../../components/Table/TableHooks';
```
jeremylenz, line 27: After https://github.com/Katello/katello/pull/10783 is merged you'll have to import these from `foremanReact`, just fyi

jeremylenz (commented):
Haven't tested this yet but here's a couple things I noticed :)

app/controllers/katello/api/v2/debs_controller.rb
nadjaheitmann: It is already found correctly in `find_hosts` method below. I agree that this is not straight forward from the logic point of view. I copied the logic from https://github.com/Katello/katello/blob/master/app/controllers/katello/api/v2/packages_controller.rb#L61 .

webpack/components/extensions/HostDetails/Tabs/DebsTab/DebsTab.js
nadjaheitmann: Thanks, applied!

nadjaheitmann:
@jeremylenz Got the react tests running! Might need some Ruby tests, though. I copied a lot of code from the packages tab. Shall I mention the developer - mostly @lfu I think - in the commit message? I think this should be credited. Another option is to have a common packages tab component for rpm packages and for deb packages and re-use it for the specific tabs. Would be the cleaner implementation, I guess, but would require quite some afford to change the code.

lfu:
> ... Shall I mention the developer - mostly @lfu I think - in the commit message? I think this should be credited...

Not necessary at all :)

jeremylenz:
> Another option is to have a common packages tab component for rpm packages and for deb packages and re-use it for the specific tabs. Would be the cleaner implementation, I guess, but would require quite some afford to change the code.

In our downstream product we rip out or disable most of the Debian-related stuff anyway, so this implementation with separate Deb components and files actually seems cleaner to me with that in mind. :+1:

nadjaheitmann:
> In our downstream product we rip out or disable most of the Debian-related stuff anyway, so this implementation with separate Deb components and files actually seems cleaner to me with that in mind. 👍

Okay, then let's leave the structure as is then. Theoretically, the PR should not affect rpm repos if it works correctly. It just adds a package tab to Debian repositories.

nadjaheitmann:
@jeremylenz This is ready for review :)

app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -560,6 +560,24 @@ def package_names_for_job_template(action:, search:, versions: nil)
...
+        when 'remove'
+          return [] if search.empty?
+
+          installed_debs.search_for(search).distinct.pluck(:name)
+        when 'update'
+          return [] if search.empty?
+
+          installed_debs.search_for(search).distinct.pluck(:name)
+        else
+          fail ::Foreman::Exception.new(N_("package_names_for_job_template: Action must be one of %s"), actions.join(', '))
```
jeremylenz, line 577:
```suggestion
          fail ::Foreman::Exception.new(N_("deb_names_for_job_template: Action must be one of %s"), actions.join(', '))
```

webpack/components/extensions/HostDetails/Tabs/DebsTab/DebInstallModal.scss
```diff
@@ -0,0 +1,3 @@
+#package-install-modal .pf-c-input-group input {
```
jeremylenz, line 1:
```suggestion
#deb-install-modal .pf-c-input-group input {
```

If we don't want this CSS to also affect PackageInstallModal, maybe it's better to use a distinct ID.

app/models/katello/deb.rb
```diff
@@ -95,6 +96,13 @@ def self.applicable_to_hosts(hosts)
         where("#{Katello::Host::ContentFacet.table_name}.host_id" => hosts).distinct
     end
 
+    # Return deb packages that are not installed on a host, but could be installed
+    # the word 'installable' has a different meaning here than elsewhere
+    def self.deb_installable_for_host(host)
```
jeremylenz, line 101: Thoughts on calling it this instead? It's more in line with "yum installable" which is what we called it on the rpm side.

```suggestion
    def self.apt_installable_for_host(host)
```

app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -560,6 +560,24 @@ def package_names_for_job_template(action:, search:, versions: nil)
         end
       end
 
+      def deb_names_for_job_template(action:, search:)
```
jeremylenz, line 563: I wonder if we could avoid adding a new method here, and just check `self.operatingsystem.family` within `package_names_for_job_template` instead?

jeremylenz (commented):
Thanks @nadjaheitmann !

Spun up a Debian 11 host and everything seems to work ok. Some minor comments below.

app/models/katello/concerns/host_managed_extensions.rb
nadjaheitmann: @jeremylenz I think, it is a good idea and I have a added a commit on top of the actual commit for the host details page. However, rubocop does not like the changes. Not sure if I can make this any simpler. 

[2024-01-05T09:07:35.947Z] Offenses:

[2024-01-05T09:07:35.947Z] /home/jenkins/workspace/katello-pr-test/app/models/katello/concerns/host_managed_extensions.rb:`541:7: C: Metrics/AbcSize: Assignment Branch Condition size for package_names_for_job_template is too high. [<9, 56, 21> 60.48/60]`

[2024-01-05T09:07:35.947Z]       def package_names_for_job_template(action:, search:, versions: nil) ...

[2024-01-05T09:07:35.947Z]       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

[2024-01-05T09:07:35.947Z] /home/jenkins/workspace/katello-pr-test/app/models/katello/concerns/host_managed_extensions.rb:`541:7: C: Metrics/CyclomaticComplexity: Cyclomatic complexity for package_names_for_job_template is too high. [16/14]`

[2024-01-05T09:07:35.947Z]       def package_names_for_job_template(action:, search:, versions: nil) ...

[2024-01-05T09:07:35.947Z]       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

[2024-01-05T09:07:35.947Z] /home/jenkins/workspace/katello-pr-test/app/models/katello/concerns/host_managed_extensions.rb:`541:7: C: Metrics/MethodLength: Method has too many lines. [34/30]`

[2024-01-05T09:07:35.947Z]       def package_names_for_job_template(action:, search:, versions: nil) ...

jeremylenz: @nadjaheitmann Two options here. I would be fine with either one:

1. Extract your `if / else` block to its own method. I would probably also do it as a case statement `(case operatingsystem.family`)
2. Or, just disable that cop and enable it at the end of the method. We do that in several places.

nadjaheitmann: I tried option one, although I did not want to use a case statement. Seems safer to me to have Debian and everything else :)

jeremylenz:
@nadjaheitmann Please squash and/or rename your commit to add 'Refs #35713', then we can merge.

jeremylenz (approved):
Thanks @nadjaheitmann!

ACK :+1:

PR #10745, opened 2023-09-19
Fixes #36758 - Add content counts to capsule content view UI

sjha4:
Add UI bits for capsule content redesigned page with content counts.
1. Add some CVs to environments tied to a capsule and run a sync.
2. Check to see if content counts show up on the UI.

Have added all necessary changes to commits here for easier testing.

### DIFF
--- a/app/models/katello/concerns/smart_proxy_extensions.rb
+++ b/app/models/katello/concerns/smart_proxy_extensions.rb
@@ -145,23 +145,12 @@ module Katello
               translated_counts[::Katello::Pulp3::PulpContentUnit.katello_name_from_pulpcore_name(name, repo)] = count
             end
           end
-          new_content_counts[:content_view_versions][repo.content_view_version_id] ||= { repositories: {}, cv_version_content_counts: {}}
-          new_content_counts[:content_view_versions][repo.content_view_version_id][:repositories][repo.id] = translated_counts
-          new_content_counts = aggregated_cv_version_count!(repo.content_view_version_id, new_content_counts, translated_counts)
+          new_content_counts[:content_view_versions][repo.content_view_version_id] ||= { repositories: {}}
+          # Store counts on capsule of archived repos which are reused across environment copies
+          # of the archived repo corresponding to each environment CV version is promoted to.
+          new_content_counts[:content_view_versions][repo.content_view_version_id][:repositories][repo.content_view_version.archived_repos.find_by(library_instance_id: repo.library_instance_id)&.id] = translated_counts
         end
         update(content_counts: new_content_counts)
-      end
-
-      def aggregated_cv_version_count!(cv_version_id, cvv_content_counts, repo_counts)
-        repo_counts.keys.each do |content_type|
-          cvv_content_counts[:content_view_versions][cv_version_id][:cv_version_content_counts][content_type] =
-            if cvv_content_counts[:content_view_versions][cv_version_id][:cv_version_content_counts][content_type]
-              cvv_content_counts[:content_view_versions][cv_version_id][:cv_version_content_counts][content_type] + repo_counts[content_type]
-            else
-              repo_counts[content_type]
-            end
-        end
-        cvv_content_counts
       end
 
       def sync_container_gateway
--- a/webpack/scenes/Content/ContentConfig.js
+++ b/webpack/scenes/Content/ContentConfig.js
@@ -16,6 +16,7 @@ export default [
       singularLowercase: __('Python package'),
       pluralLabel: 'python_packages',
       singularLabel: 'python_package',
+      capsuleCountLabel: 'python_package',
     },
     columnHeaders: [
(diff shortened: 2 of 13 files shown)


ianballou (commented):
I think the counts are looking good! One thing I just noticed is that we have a discontinuity between our content unit naming on content views / smart proxies and repositories. Like "Container Image Tags" versus "Container tags".

Anyway, from a testing perspective it's looking good.

I'm currently looking into something weird, but it'll be unrelated to this PR... my CVV package count is different from the capsule package count, but CVV.packages is also a different number, and that number is ALSO different from what I get if I count up the packages reported in each of the CVVs repos... I'll report back if it's something that needs addressing here...

ianballou:
> I'm currently looking into something weird, but it'll be unrelated to this PR... my CVV package count is different from the capsule package count, but CVV.packages is also a different number, and that number is ALSO different from what I get if I count up the packages reported in each of the CVVs repos... I'll report back if it's something that needs addressing here...

It looks like the issue is that, on a content view version, we show the count of unique packages.  However, here, are are counting up the total number of packages. This is a little problematic because we only get the counts from Pulp, not the NVRAs.

Some potential solutions:

1. Don't aggregate the content units by content view version. Show the counts for each repositories within the content view version.
2.  Stop showing the unique package count on the content view version and instead count duplicates.
  -> While this would be a change in behavior that users might notice, at least it would get the count closer to the truth.  A package in one repository isn't the same package as in the other repository. However, it might mean changing how `in_repositories` works...
3. Add a "raw count" to content view versions that users can use to specifically compare content on the Foreman vs on the smart proxy.

ianballou (commented):
I noticed that some repositories are missing content type names:

image

webpack/scenes/Content/ContentConfig.js
```diff
@@ -142,6 +144,7 @@ export default [
       singularLowercase: __('Ansible collection'),
       pluralLabel: 'ansible_collections',
       singularLabel: 'ansible_collection',
+      capsuleCountLabel: 'ansible.collection_version',
```
ianballou, line 147: Can this be fixed on the Ruby side so that it gets translated from Pulp words to Katello words?

I thought it was `ansible.collection`... `PULPCORE_CONTENT_TYPE = "ansible.collection".freeze`.

We can change that to `ansible.collection_version`.

webpack/scenes/Content/ContentConfig.js
```diff
@@ -81,6 +82,7 @@ export default [
       singularLowercase: __('OSTree ref'),
       pluralLabel: 'ostree_refs',
       singularLabel: 'ostree_ref',
+      capsuleCountLabel: 'ostree.refs',
```
ianballou, line 85:
```suggestion
      capsuleCountLabel: 'ostree_ref',
```

app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -146,7 +146,7 @@ def update_content_counts!
             end
           end
           new_content_counts[:content_view_versions][repo.content_view_version_id] ||= { repositories: {}, cv_version_content_counts: {}}
-          new_content_counts[:content_view_versions][repo.content_view_version_id][:repositories][repo.id] = translated_counts
+          new_content_counts[:content_view_versions][repo.content_view_version_id][:repositories][repo.content_view_version.archived_repos.find_by(library_instance_id: repo.library_instance_id)&.id] = translated_counts
```
ianballou, line 149: What do you think about adding a comment somewhere saying that the repos returned are only the archived ones?

webpack/scenes/Content/ContentConfig.js
sjha4: Updated 👍🏼

webpack/scenes/Content/ContentConfig.js
```diff
@@ -81,6 +82,7 @@ export default [
       singularLowercase: __('OSTree ref'),
       pluralLabel: 'ostree_refs',
       singularLabel: 'ostree_ref',
+      capsuleCountLabel: 'ostree_refs',
```
ianballou, line 85:
```suggestion
      capsuleCountLabel: 'ostree_ref',
```

Than ostree should show up!

image

ianballou (changes requested):
Just found one typo:

ianballou:
Looking good now! 

image

sjha4:
[test katello]

ianballou (commented):
Everything is looking good! Only one comment:

Should the repository entries be alphabetical?

image

sjha4:
Sorting the repos by name is a little tricky given we only have access to ids in the content count json. We do have list of repositories on the UI and it should be possible to sort those. Opened an issue here: https://projects.theforeman.org/issues/36806 to work on that and move this PR along. Thanks @ianballou !

ianballou (approved):
Ack!

PR #10748, opened 2023-09-20
Fixes #36762 - Katello css overrides foreman

MariaAga:
wrapping generic css rules in specific classes/ids

Went over all `.scss` files to see which have css rules that override css in non-katello pages 

Please open the pages effected to make sure they still look ok


MariaAga:
@jeremylenz  Can you take a look?

jeremylenz (approved):
Works for me! :+1: 

Thanks @MariaAga

PR #10750, opened 2023-09-20
Fixes #36766 - Set disks to amount of disks instead of bytes

chris1984:

* We had the number of disks using bytes instead of an actual count
* I fixed the count by looking at the block devices and filtering out the cdrom or sr0

* https://projects.theforeman.org/issues/36344 is assigned to someone from the Foreman team who will add the facts from Puppet.
* Since we are fixing the total disks, I kept this pr doing the same thing instead of taking over the 36344 card.

* Apply PR
* Register a client
* Check to see if it reports the correct disk counts

Before PR: https://community.theforeman.org/t/hosts-ram-details-buggy/35050/7?u=cintrix84

After PR:

prdisk

### DIFF
--- a/webpack/components/extensions/HostDetails/DetailsTabCards/HwPropertiesCard.js
+++ b/webpack/components/extensions/HostDetails/DetailsTabCards/HwPropertiesCard.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { propsToCamelCase } from 'foremanReact/common/helpers';
 import { translate as __ } from 'foremanReact/common/I18n';
 import {
   DescriptionList,
@@ -14,7 +13,12 @@ import CardTemplate from 'foremanReact/components/HostDetails/Templates/CardItem
 import { TranslatedPlural } from '../../../Table/components/TranslatedPlural';
 import { hostIsNotRegistered } from '../hostDetailsHelpers';
 
-const HostDisks = ({ totalDisks }) => {
+const HostDisks = ({ blockDevices }) => {
+  if (!blockDevices) return null;
+  // blockDevices fact will look like this by default 'sr0,sda' and increment like sdb etc
+  const disks = blockDevices.split(',');
+  // We are filtering out the CDROM drive that gets added by default a lot of the time
+  const totalDisks = disks.filter(e => !e.startsWith('sr')).length;
   if (!totalDisks) return null;
   return (
     <>
@@ -27,11 +31,11 @@ const HostDisks = ({ totalDisks }) => {
 };
 
 HostDisks.propTypes = {
-  totalDisks: PropTypes.number,
+  blockDevices: PropTypes.string,
 };
 
 HostDisks.defaultProps = {
-  totalDisks: null,
+  blockDevices: '',
 };
 
 const HwPropertiesCard = ({ isExpandedGlobal, hostDetails }) => {
@@ -41,8 +45,7 @@ const HwPropertiesCard = ({ isExpandedGlobal, hostDetails }) => {
   const cpuCount = facts?.['cpu::cpu(s)'];
   const cpuSockets = facts?.['cpu::cpu_socket(s)'];
   const coreSocket = facts?.['cpu::core(s)_per_socket'];
-  const reportedFacts = propsToCamelCase(hostDetails?.reported_data || {});
-  const totalDisks = reportedFacts?.disksTotal;
+  const blockDevices = facts?.blockdevices;
   const memory = facts?.['dmi::memory::maximum_capacity'];
 
   return (
@@ -74,7 +77,7 @@ const HwPropertiesCard = ({ isExpandedGlobal, hostDetails }) => {
           <DescriptionListDescription>{memory}</DescriptionListDescription>
         </DescriptionListGroup>
         <DescriptionListGroup>
-          <HostDisks totalDisks={totalDisks} />
+          <HostDisks blockDevices={blockDevices} />
         </DescriptionListGroup>
       </DescriptionList>
     </CardTemplate>
@@ -90,9 +93,7 @@ HwPropertiesCard.propTypes = {
       cpuSockets: PropTypes.number,
       coreSocket: PropTypes.number,
       memory: PropTypes.string,
-    }),
-    reported_data: PropTypes.shape({
-      totalDisks: PropTypes.number,
+      blockdevices: PropTypes.string,
     }),
   }),
 };


chris1984:
@sbernhard How does this look? I took a different approach since there is another redmine that someone else is working on for adding puppet support. I just altered the disks to be the correct total like we had before but with a different method of getting the disks.

The girl working on the puppet support will update this card in the future

webpack/components/extensions/HostDetails/DetailsTabCards/HwPropertiesCard.js
```diff
@@ -41,8 +45,7 @@ const HwPropertiesCard = ({ isExpandedGlobal, hostDetails }) => {
   const cpuCount = facts?.['cpu::cpu(s)'];
   const cpuSockets = facts?.['cpu::cpu_socket(s)'];
   const coreSocket = facts?.['cpu::core(s)_per_socket'];
-  const reportedFacts = propsToCamelCase(hostDetails?.reported_data || {});
-  const totalDisks = reportedFacts?.disksTotal;
+  const blockDevices = facts?.blockdevices;
```
ekohl, line 48: If I look at Facter you have a fact `disks` which does exactly what you'd expect. Doesn't RHSM have something similar?

webpack/components/extensions/HostDetails/DetailsTabCards/HwPropertiesCard.js
```diff
@@ -14,8 +13,13 @@ import CardTemplate from 'foremanReact/components/HostDetails/Templates/CardItem
 import { TranslatedPlural } from '../../../Table/components/TranslatedPlural';
 import { hostIsNotRegistered } from '../hostDetailsHelpers';
 
-const HostDisks = ({ totalDisks }) => {
-  if (!totalDisks) return null;
+const HostDisks = ({ blockDevices }) => {
```
ekohl, line 16: I still think it's poor practice to apply this logic in the frontend. We've settled on the reported data facet to normalize all fact uploads. Ideally it would be implemented in https://github.com/theforeman/foreman/blob/develop/app/services/katello/rhsm_fact_parser.rb

webpack/components/extensions/HostDetails/DetailsTabCards/HwPropertiesCard.js
ekohl: Probably yes.

sbernhard:
Maybe a more generic question: is the amount of disks a system has really interesting? Or should it just list the size of the storage?

chris1984:
> Maybe a more generic question: is the amount of disks a system has really interesting? Or should it just list the size of the storage?

I think having both would be beneficial, I will do the total disks and the other Redmine is being looked at for adding disk info from puppet_facts.

sbernhard:
What's the reason to close this PR @chris1984 ? 
I think we should still try to fix this isuse.

ekohl:
Thinking more about it: it would probably still be a valid bug fix cherry pick.

sbernhard:
Are the failed tests related to this change? Would be good to finish this soonish.

chris1984:
#
/
[test katello]

chris1984:
@sbernhard tests are green if you want to test it

webpack/components/extensions/HostDetails/DetailsTabCards/HwPropertiesCard.js
```diff
@@ -14,8 +13,13 @@ import CardTemplate from 'foremanReact/components/HostDetails/Templates/CardItem
...
+  if (!blockDevices) return null;
+  // blockDevices fact will look like this by default 'sr0,sda' and increment like sdb etc
+  const disks = blockDevices.split(',');
+  // We are filtering out the CDROM drive that gets added by default a lot of the time
+  const totalDisks = disks.filter(e => e !== 'sr0').length;
```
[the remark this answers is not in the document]
sbernhard:
```suggestion
  const totalDisks = disks.filter(e => !e.startsWith('sr')).length
```

sbernhard:
> @sbernhard tests are green if you want to test it

I tested it together with my suggestion and it worked:

image

sbernhard (approved):
Thanks. Works as expected.

chris1984:
Awesome, thanks for testing it. Will get you change in and pushed up so we can get it in.

chris1984:
@sbernhard pushed up your fix, will put this on the review board for someone with merge access to ack it so it can be merged

chris1984:
[test katello]

webpack/components/extensions/HostDetails/DetailsTabCards/HwPropertiesCard.js
```diff
@@ -14,8 +13,13 @@ import CardTemplate from 'foremanReact/components/HostDetails/Templates/CardItem
...
 
-const HostDisks = ({ totalDisks }) => {
-  if (!totalDisks) return null;
+const HostDisks = ({ blockDevices }) => {
+  if (!blockDevices) return null;
+  // blockDevices fact will look like this by default 'sr0,sda' and increment like sdb etc
+  const disks = blockDevices.split(',');
+  // We are filtering out the CDROM drive that gets added by default a lot of the time
+  const totalDisks = disks.filter(e => !e.startsWith('sr')).length;
+  if (totalDisks.length === 0) return null;
```
jeremylenz, line 22: With

```js
const totalDisks = disks.filter(e => !e.startsWith('sr')).length;
```

`totalDisks` will be a number, right? because we're taking the `length` of the filtered array

So on the next line, why are we again checking `totalDisks.length`?  

```js
const totalDisks = 0
undefined
totalDisks.length
undefined
totalDisks.length === 0
false
```

chris1984: @jeremylenz Updated

```javascript
const totalDisks = 0
undefined
if (!totalDisks) console.log('hi')
VM247:1 hi
```

Went with this approach

webpack/components/extensions/HostDetails/DetailsTabCards/HwPropertiesCard.js
```diff
@@ -27,11 +31,11 @@ const HostDisks = ({ totalDisks }) => {
 };
 
 HostDisks.propTypes = {
-  totalDisks: PropTypes.number,
+  blockDevices: PropTypes.string,
 };
 
 HostDisks.defaultProps = {
-  totalDisks: null,
+  blockDevices: null,
```
jeremylenz, line 38: If `blockDevices` comes in as `null`, this will cause a PropTypes error. Should set the default to `''` here and then make sure that's handled-- _or_ make it a required prop.

chris1984:
[test katello]

chris1984:
@jeremylenz how does this look?

jeremylenz (approved):
LGTM! :+1:

PR #10752, opened 2023-09-25
Fixes #36776 - Fixed issue where composite content views promoted to …

qcjames53:
…lifecycle environments referencing the env name would error on promotion.

Changed env replacement source.

Composite content views containing archived repositories have their environment set to nil. When promoting this CVV to a lifecycle environment with a registry name containing the lifecycle environment's name, this would previously cause the action to error out. After the changes, if the env is unset, Katello will grab a valid environment from the first non-archived repository in the repository's content view.

I'm not 100% positive on this solution. This line is sketchy, but I haven't been able to break it in my testing:
```
environment ||= repository.content_view.repositories.first.environment
```
Is there a scenario where the archived repo's content view's first repo would belong to a different environment than the unarchived repo? Is there a better way to link an archived repo to it's unarchived version?

1. Create a product:
```
hammer product create --description "Product Busybox Container" --name Product_Busybox_Container --label "Product_Busybox_Container" --organization myorg
```
2. Create a repository and sync it:
```
hammer repository create \
  --content-type "docker" \
  --organization myorg \
  --product "Product_Busybox_Container" \
  --url "https://registry-1.docker.io/" \
  --name "library/busybox" \
  --docker-upstream-name "library/busybox" \
  --label "library_busybox" \
  --include-tags "stable, 1.35.0"
hammer repository synchronize --name "library/busybox" --product "Product_Busybox_Container" --organization myorg
```
3. Create a content view with this repo and publish it:
```
hammer content-view create --label ContentView_Busybox --name ContentView_Busybox --organization-label myorg --repository-ids <repo_id>
hammer content-view publish --organization myorg --name ContentView_Busybox
```
4. Using the UI, go to Content -> Lifecycle -> Content Views. Create a content view of type composite content view. I left auto-publish unchecked.
5. Once created, add the ContentView_Busybox CV to this composite CV.
6. Go to Content -> Lifecycle -> Lifecycle Environments. Create an environment path. Click on the newly-created environment path. On the details tab, modify the registry name pattern to `<%= repository.name %>`.
7. Create another environment path. Set its registry name pattern to `<%= lifecycle_environment.label %>/<%= repository.name %>`
8. Go to Content -> Lifecycle -> Content Views. Click on the kebab menu next to your composite content view and select promote. Select the first lifecycle environment and publish. There should be no errors because the environment isn't being referenced.
9. Refresh the page. Click on the kebab menu next to your composite content view and select promote. Select the second lifecycle environment and publish. This PR should not error out, since the environment is now being pulled from the un-archived repo instead of using the archived repo's nil environment.


ianballou:
> After the changes, if the env is unset, Katello will grab a valid environment from the first non-archived repository in the repository's content view.

I haven't fully looked through the code yet, but I don't think the special registry naming scheme should even apply to archived repositories. The archived repositories shouldn't be exposed to users besides for debugging. While I'm guessing this stops the error, the env picked likely doesn't match the repository. 

Could we instead stop using the special naming scheme for archived repositories? Perhaps just ignore the registry naming pattern?

qcjames53:
> > After the changes, if the env is unset, Katello will grab a valid environment from the first non-archived repository in the repository's content view.
> 
> I haven't fully looked through the code yet, but I don't think the special registry naming scheme should even apply to archived repositories. The archived repositories shouldn't be exposed to users besides for debugging. While I'm guessing this stops the error, the env picked likely doesn't match the repository.
> 
> Could we instead stop using the special naming scheme for archived repositories? Perhaps just ignore the registry naming pattern?

I implemented the suggestions we talked about in our PM, meaning if the environment does not exist (as in archived repos), render the name as if a pattern is not provided. I reworked the logic a little to make it easier to understand and added a big table in the documentation to explain everything.

app/models/katello/repository.rb
```diff
@@ -860,20 +860,40 @@ def self.search_by_content_label(_key, operator, value)
...
+      # env exists /  00|   4 | n/a | n/a |   4
+      # is cv default 01|   2 | n/a | n/a |   4
+      #               11|   2 |   1 |   1 |   1
+      #               10|   3 |   1 |   1 |   1
+      #
+      # This diagram shows the name to render given the properties of
+      # the container provided. Branches numbered as ordered below.
+      is_pattern_provided = pattern && !pattern.blank?
+      env_exists = repository.environment.present?
+      is_env_pattern_provided = env_exists && !repository.environment.registry_name_pattern.empty?
```
ianballou, line 875:
```suggestion
      is_env_pattern_provided = env_exists && repository.environment.registry_name_pattern.present?
```

app/models/katello/repository.rb
```diff
@@ -860,20 +860,40 @@ def self.search_by_content_label(_key, operator, value)
...
+      #                 |  00 |  01 |  11 |  10
+      # ----------------+-----+-----+-----+------
+      # env exists /  00|   4 | n/a | n/a |   4
+      # is cv default 01|   2 | n/a | n/a |   4
+      #               11|   2 |   1 |   1 |   1
+      #               10|   3 |   1 |   1 |   1
+      #
+      # This diagram shows the name to render given the properties of
+      # the container provided. Branches numbered as ordered below.
+      is_pattern_provided = pattern && !pattern.blank?
```
ianballou, line 873: Could be upgraded with `pattern.present?`

ianballou (commented):
I haven't tested it, but the logic looks solid to me! Thanks for the table.

test/models/content_view_test.rb
```diff
@@ -391,15 +391,18 @@ def test_docker_repo_container_names
...
 
-      composite.update(component_ids: [version1.id])
+      composite.update(component_ids: [version1.id, version2.id])
       assert composite.valid?
-      @dev.registry_name_pattern = "abcdef"
       assert composite.check_docker_repository_names!([@dev])
 
-      composite.update(component_ids: [version1.id, version2.id])
-      assert composite.valid?
-      assert_raises(RuntimeError) do
```
ianballou, line 401: This test used to verify that `check_docker_repository_names!` would error out if there is a clash between the names in an environment. Since I don't see any other test doing that anymore, I think a new test will need to be created that does just that.

It might be worth just splitting your new assertions here into a new test and remove the extra bits from `test_docker_repo_container_names` that you don't need. Then you could return `test_docker_repo_container_names` to doing what it was before.

ianballou (commented):
Functionality-wise it works like a charm

test/models/content_view_test.rb
qcjames53: To me, it seems like the tail end of `test_docker_promote` handles that? I could be incorrect though. I can absolutely add/repair this if you think it's the correct thing to do.

ianballou: Looks like the original test fails because the repositories do not have any environment. If you add environments to the repositories that would be published into the composite, the test should succeed.

ianballou: > To me, it seems like the tail end of `test_docker_promote` handles that? I could be incorrect though. I can absolutely add/repair this if you think it's the correct thing to do.

`test_docker_promote` tests that `check_docker_repository_names!` succeeds, but it doesn't test that it fails.

qcjames53: I believe I fixed this test but I'd love a second pair of eyes on this. I couldn't find a way to change the container repo name after creation, so I ended up just testing for an invalid record error on the attempted creation of the second identical cv.

qcjames53: Because the creation fails on the creating a cv step, there are no content view versions in use here. That said, I believe it still tests whether two docker repos will error out when they share a container repository name.

ianballou (approved):
Ship it!

app/models/katello/repository.rb
```diff
@@ -860,20 +860,45 @@ def self.search_by_content_label(_key, operator, value)
...
+      # ----------------+-----+-----+-----+------
+      # env exists /  00|   4 | n/a | n/a |   4
+      # is cv default 01|   2 | n/a | n/a |   4
+      #               11|   2 |   1 |   1 |   1
+      #               10|   3 |   1 |   1 |   1
+      #
+      # This table shows the name to render given the properties of
+      # the container provided. Branches numbered as ordered below.
+      #
+      # 1 - Render promotion pattern (or env pattern if no promo pattern)
```
ianballou, line 874: By  promotion pattern, do you mean the variable `pattern` in this method?

qcjames53: I do. This pattern is set based on the registry name pattern of the lifecycle environment to which the CVV is being added. I believe it is blank and the env pattern takes over in some other cases. Is this still good to ship?

ianballou: All good!

PR #10755, opened 2023-09-27
Fixes #36787 - Delete oprhan content task doesn't remove orphaned remotes in the Capsule

sjha4:
Delete *all* orphaned remotes on capsule when deleting orphaned content.
1. Create >100 repositories
2. Add and publish the repo to a CV version tied to an env on a capsule.
3. Sync capsule.
4. Edit capsule to remove the env.
5. Run "bundle exec rails katello:delete_orphaned_content" and make sure all remotes are deleted on capsule.
You can check that using pulp-cli from your server with:
"pulp --profile=proxy rpm remote list"

### DIFF
--- a/app/services/katello/pulp3/api/core.rb
+++ b/app/services/katello/pulp3/api/core.rb
@@ -224,7 +224,7 @@ module Katello
           remotes_api.list(args).results
         end
 
-        def remotes_list_all(_smart_proxy, options)
+        def remotes_list_all(_smart_proxy, options = {})
           self.class.fetch_from_list do |page_opts|
             remotes_api.list(page_opts.merge(options))
           end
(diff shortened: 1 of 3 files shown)


qcjames53 (approved):
Hi Samir! Everything in your PR works as advertised; I couldn't find a way to break anything. I have a couple of conceptual questions but for the functionality presented here I'm ok giving the go-ahead.

- I think it would be beneficial to include a delete orphan content button in the web UI. Perhaps this could be a future card.
- Is there a reason orphaned content isn't removed automatically when an env is taken off a smart proxy? If not, I think we should add this feature as well. Different card.
- I think it would be useful (at least in the context of this card) to have a repo count in the UI and hammer on the smart proxy info pages. Would this be expensive to calculate? Just would be a nice-to-have.

PR #10756, opened 2023-09-28
Fixes #36790 - Show recalculate action in happy empty state

jeremylenz:

Adds a "Recalculate" link to the "happy" empty state for ErrataTab.

There's no change in functionality for this action (for now), so you're going to have to just wait for the task to complete and then refresh the browser. It doesn't poll the task like it does for REX jobs.

image

I first just brought back the table toolbar, but it looked super ugly with the disabled "Apply" button.
Then I thought about hiding the apply button and just having the kebab in its own row, but that would also be ugly.
This is the only non-ugly solution I could think of.

Make sure the ErrataTab continues to work properly in all situations (no errata, some errata, no search results)


jeremylenz:
[test katello]

ianballou (approved):
Works great!

PR #10757, opened 2023-09-29
Fixes #36789 - Add actions to force applicability refresh

jeremylenz:
Goes with https://github.com/theforeman/foreman/pull/9847

Until now, the 'Recalculate' button on the Errata tab only did the server-side recalculation. This means that if you've changed the host content view, content overrides, etc. it might use an out-of-date installed package list to calculate applicability.

See https://community.theforeman.org/t/list-of-packages-to-install-not-up-to-date/35153

In this PR, we add new kebab actions to the host details page:
* main kebab - 'Refresh applicability'
* Packages kebab - 'Refresh package applicability'
* Errata kebab - 'Refresh errata applicability'. This replaces the  "Recalculate" button on the Errata kebab.
* Errata happy empty state - 'Refresh errata applicability' link. This replaces the 'Recalculate' link that was recently added.

The button will do the same thing in all four places - run 'subscription-manager repos' on the host via REX. This will also trigger (as it always has) errata applicability recalculation, and the web UI will be updated immediately.

Turns out the hardest part of this was _getting the main kebab menu to close when you click on it_. Never woulda thought. (See the linked PR) I considered several solutions including Redux and slot and fill, but the React context solution I arrived at seems the most elegant.

On the ErrataTab, the new action _replaces_ the "Recalculate" action, which used to just do the server-side recalculation. We thought this is what most users would want, since recalculating without refreshing the host's installed package list is only useful in certain situations.

Server-side-only recalculation will still be available via Hammer. (maybe we should also make that available as a host bulk action in the new UI)

Check out https://github.com/theforeman/foreman/pull/9847 in Foreman
Get REX working on at least one host
Test the recalculate action in all three places
Make sure everything works


jeremylenz:
rebased after merging https://github.com/Katello/katello/pull/10756

ianballou (commented):
Working great! We'll probably want to make sure that there's some documentation around this change at Katello 4.11 release time.

Is it okay to assume that users will have remote execution set up? The applicability calculation button is sorta useless now since we always update applicability unless there is some error, so I think it's okay. Just makes me wonder if users are going to miss the old button for whatever reason.

jeremylenz:
> Is it okay to assume that users will have remote execution set up? The applicability calculation button is sorta useless now since we always update applicability unless there is some error, so I think it's okay. Just makes me wonder if users are going to miss the old button for whatever reason.

I think it's somewhat safe to assume that REX is present, since it's now the only way to patch hosts with Katello and is included by default with Satellite. I suppose I could make sure to hide the actions if REX isn't enabled; let me know if you want me to add that.

ianballou (approved):
@jeremylenz after some thought I think this is good as-is. Mind just putting in a request for a docs update to warn that server-side applicability calcs are going hammer-only? Probably a release note.  I can't believe that that button is useful for anyone anymore, especially after the move to Pulp 3.

PR #10762, opened 2023-10-06
Fixes #36808 - Add sudo to Resolve Traces template

jeremylenz:

In the Resolve Traces - Katello Script Default template, use `sudo` when running `katello-tracer-upload`. This will allow REX to use a non-root user when executing the job.

We also have a "Resolve Traces - Katello Ansible Default" but I didn't change that one because I think Ansible defaults to `become: true` anyway, is that right?

Check out the PR.
Refresh your job templates:
```
# bundle exec rails console
ForemanInternal.all.last.destroy
```

Then go look at the locked job template and verify that it's updated. Bonus: actually run it via REX as a non-root user

### DIFF
--- a/app/views/foreman/job_templates/resolve_traces.erb
+++ b/app/views/foreman/job_templates/resolve_traces.erb
@@ -20,5 +20,5 @@ reboot = commands.delete('reboot')
 shutdown -r +1
 <% else -%>
 <%= commands.join("\n") %>
-katello-tracer-upload
+sudo katello-tracer-upload
 <% end %>


jeremylenz:
[test katello]

jeremylenz:
[test katello]

chris1984:
Code wise it looks fine and your comment about `become: true` is correct so we are becoming sudo anyways. Starting to test

chris1984 (approved):
ACK works great.

Screenshot showing locked template is updated:
prtemplate

Screenshot showing vagrant user being used instead of root:
Screenshot 2023-10-10 at 14-26-59 Run job


Screenshot showing job completed:
Screenshot 2023-10-10 at 14-28-26 Resolve Traces

chris1984:
[test katello]

ianballou:
[test katello]

chris1984:
@jeremylenz All green :)

PR #10766, opened 2023-10-12
Fixes #36807 - UX fixes and add update count action

sjha4:
Added task link
Added action on page to refresh counts (Also sets up code to add syncs in the future)
UX fixes
1. Go to Capsule> content tab and make sure the new action works.
2. Also some minor cosmetic changes.


sjha4:
[test katello]

sjha4:
[test katello]

app/services/katello/pulp3/repository_mirror.rb
```diff
@@ -75,7 +75,7 @@ def fetch_repository
       end
 
       def version_href
-        fetch_repository.latest_version_href
+        fetch_repository&.latest_version_href
```
ianballou, line 78: Even with this, `latest_content_counts` tries to do `api.repository_versions_api.read(version_href)` which errors out. We could just skip that repository perhaps if version_href is `nil`.

ianballou:
I also noticed that failing content counts updates pauses the task. I  think we could just make it stop completely and turn into a warning since the json update is all or  nothing.

app/services/katello/pulp3/repository_mirror.rb
ianballou: To reproduce:

1) Create a CVV assigned to some env with some yum content
2) Sync that CVV to the smart proxy
3) Promote the CVV to a new LCE
4) Add that LCE to the smart proxy
5) Run the content counts update
6) See the error:
```
Exception:

ArgumentError: Missing the required parameter 'rpm_rpm_repository_version_href' when calling RepositoriesRpmVersionsApi.read

Backtrace:

/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/pulp_rpm_client-3.22.5/lib/pulp_rpm_client/api/repositories_rpm_versions_api.rb:238:in `read_with_http_info'
/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/pulp_rpm_client-3.22.5/lib/pulp_rpm_client/api/repositories_rpm_versions_api.rb:221:in `read'
/home/vagrant/katello/app/services/katello/pulp3/repository_mirror.rb:228:in `latest_content_counts'
/home/vagrant/katello/app/models/katello/concerns/smart_proxy_extensions.rb:133:in `block in update_content_counts!'
/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activerecord-6.1.7.6/lib/active_record/relation/delegation.rb:88:in `each'
/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activerecord-6.1.7.6/lib/active_record/relation/delegation.rb:88:in `each'
/home/vagrant/katello/app/models/katello/concerns/smart_proxy_extensions.rb:131:in `update_content_counts!'
/home/vagrant/katello/app/lib/actions/katello/capsule_content/update_content_counts.rb:15:in `run'
```

sjha4:
The latest commit should:
1. Allow content_counts to be updated even with presence of repositories not synced on capsule.
2. Use skip over pause as rescue strategy for failed content_count update task.

app/services/katello/pulp3/repository_mirror.rb
sjha4: Updated to skip repos not synced to capsule, i.e version_href = nil

sjha4:
[test katello]

sjha4:
[test katello]

sjha4:
[test katello]

ianballou:
The updated content count logic seems to be messing up somewhere for me. I have a content view version with a yum repo and a python repo that is promoted to 5 LCEs including Library. All of the LCEs except Library are assigned to my capsule, and only one of those are actually synced.

What I'm seeing is this:

```
               "content_counts" => {
                    "content_view_versions" => {
                        "2" => {
                            "repositories" => {
                                "2" => {
                                        "rpm" => 32,
                                    "erratum" => 4
                                }
                            }
                        },
                        "8" => {
                            "repositories" => {
                                "62" => {
                                        "rpm" => 32,
                                    "erratum" => 4
                                },
                                "63" => {}
                            }
                        }
                    }

```

63 is my python repo. For some reason, the  counts are completely blank even though the smart proxy repo has python packages. The yum content is being recorded, however.

sjha4:
> The updated content count logic seems to be messing up somewhere for me. I have a content view version with a yum repo and a python repo that is promoted to 5 LCEs including Library. All of the LCEs except Library are assigned to my capsule, and only one of those are actually synced.
> 
> What I'm seeing is this:
> 
> ```
>                "content_counts" => {
>                     "content_view_versions" => {
>                         "2" => {
>                             "repositories" => {
>                                 "2" => {
>                                         "rpm" => 32,
>                                     "erratum" => 4
>                                 }
>                             }
>                         },
>                         "8" => {
>                             "repositories" => {
>                                 "62" => {
>                                         "rpm" => 32,
>                                     "erratum" => 4
>                                 },
>                                 "63" => {}
>                             }
>                         }
>                     }
> ```
> 
> 63 is my python repo. For some reason, the counts are completely blank even though the smart proxy repo has python packages. The yum content is being recorded, however.

What did I do now?! Let me look.. 😭 😄

ianballou:
I think I see the issue, `latest_content_counts` is `nil` for a repo not yet synced on the smart proxy, so it's replacing the actual content counts that were calculated for a good repo earlier.

The fact that we loop through unarchived repositories might be making this difficult. Do you think we could be looping over archived repositories instead?

ianballou:
So it seems we keep overwriting the content counts as we loop along because we're storing the counts only for the archived repository.  However, on the smart proxy, we do sync the promoted repositories separately.

Should we go back to showing counts for all promoted repositories? Those are the ones that users consume from at least... I know it confused something in the UI, but I wonder if now the counts aren't showing enough.

sjha4:
Updated PR to:
1. Return env_repo id as key in content_counts.
2. Added metadata section to json to hold env_id of repo and library_instance_id of env_repo.
3. Updated UI/tests to handle the JSON change.

sjha4:
[test katello]

webpack/scenes/SmartProxy/ExpandedSmartProxyRepositories.js
```diff
@@ -4,17 +4,40 @@ import PropTypes from 'prop-types';
...
-    cellList.push(<DataListCell key={`${repo.id}-count`}><AdditionalCapsuleContent counts={contentCounts[repo]} /></DataListCell>);
+    if (syncedToCapsule) {
+      cellList.push(<DataListCell key={`${repo.id}-name`}><span><a href={`/products/${envContentCounts[repo].metadata.product_id}/repositories/${envContentCounts[repo].metadata.library_instance_id}/`}>{getRepositoryNameById(envContentCounts[repo].metadata.library_instance_id)}</a></span></DataListCell>);
+      cellList.push(<DataListCell key={`${repo.id}-type`}><RepoIcon type={envContentCounts[repo].metadata.content_type} /></DataListCell>);
+      cellList.push(<DataListCell key={`${repo.id}-rpm`}><span>{envContentCounts[repo].counts.rpm ? `${envContentCounts[repo].counts.rpm} Packages` : 'N/A'}</span></DataListCell>);
+      cellList.push(<DataListCell key={`${repo.id}-count`}><AdditionalCapsuleContent counts={envContentCounts[repo].counts} /></DataListCell>);
+    } else {
+      cellList.push(<DataListCell key={`${repo.id}-not-synced`}><InactiveText text={__('Content view is not synced to capsule')} /></DataListCell>);
```
ianballou, line 39: image

What do you think about showing the repository name and moving the "not synced" text to the count field? That would match up with how I'm doing it in Hammer.

ianballou (commented):
Overall this is looking good! Gonna do some more in-depth testing tomorrow, but I think it should be good soon.

PR #10767, opened 2023-10-12
Fixes #35901 - Allow searching for CCVs based on component CVs

jeremylenz:

Add a new field to Content View scoped_search. This will allow you to search for a composite content view that contains a particular content view.

For example:
```
content_views = My-Content_view
```

will return _composite_ content views that have My-Content_view as one of their components.

I'm still not sure about any performance implications; suggestions welcome

Create at least one composite CV but preferably hundreds lol
Search on the Content Views page
Make sure everything works as expected

### DIFF
--- a/app/models/katello/content_view.rb
+++ b/app/models/katello/content_view.rb
@@ -111,6 +111,27 @@ module Katello
     scoped_search :on => :composite, :complete_value => true
     scoped_search :on => :generated_for, :complete_value => true
     scoped_search :on => :default # just for ordering
+    scoped_search :on => :name, :complete_value => true,
+                  :rename => :content_views,
+                  :operators => ['='],
+                  :ext_method => :find_components_by_cv_name
+
+    def self.find_components_by_cv_name(_key, operator, value)
+      kcv = Katello::ContentView.table_name
+      kcvc = Katello::ContentViewComponent.table_name
+      { :conditions => "#{kcv}.composite = 't' AND #{kcv}.id IN (SELECT #{kcvc}.composite_content_view_id FROM #{kcvc} WHERE #{kcvc}.content_view_id IN (SELECT #{kcv}.id FROM #{kcv} WHERE #{kcv}.name #{operator} ?))",
+        :parameter => [value]
+      }
+    end
+
+    def self.completer_scope_options(search)
+      if search.include?('content_views')
+        # Don't autocomplete CCV names when searching for components
+        { :value_filter => { :composite => false } }
+      else
+        {}
+      end
+    end
 
     enum generated_for: {
       none: 0,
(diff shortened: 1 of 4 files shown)


app/models/katello/content_view.rb
```diff
@@ -111,6 +111,27 @@ class ContentView < Katello::Model
...
+    end
+
+    def self.completer_scope_options(search)
+      if search.include?('content_views')
+        # Don't autocomplete CCV names when searching for components
+        { :value_filter => { :composite => false } }
+      else
+        {}
+      end
+    end
```
jeremylenz, line 134: Autocomplete was including composite CV names in the suggested search. For example, if I was searching for `content_views = xxx` and had typed `content_views = `, the suggestions would include all of my content view names, including _composite_ content views. But a composite content view cannot contain itself, so this made no sense. To prevent this I added this method, which modifies scoped_search's value filter. I had to modify it to pass in the search because otherwise it would filter out composite CVs when I searched for `name = `, which we don't want.

app/models/katello/content_view.rb
```diff
@@ -111,6 +111,27 @@ class ContentView < Katello::Model
     scoped_search :on => :composite, :complete_value => true
     scoped_search :on => :generated_for, :complete_value => true
     scoped_search :on => :default # just for ordering
+    scoped_search :on => :name, :complete_value => true,
+                  :rename => :content_views,
+                  :operators => ['=', '!=', '~', '!~'],
```
jeremylenz, line 116: I filtered out the operators like `<` that don't make much sense here, as well as `^` which just don't seem to work well for this use case

app/models/katello/content_view.rb
```diff
@@ -111,6 +111,27 @@ class ContentView < Katello::Model
     scoped_search :on => :composite, :complete_value => true
     scoped_search :on => :generated_for, :complete_value => true
     scoped_search :on => :default # just for ordering
+    scoped_search :on => :name, :complete_value => true,
+                  :rename => :content_views,
```
jeremylenz, line 115: I considered naming this `components`, but we've moved away from using the "component" terminology recently, and I don't want to reverse that.

jeremylenz:
fixed rubocop & added unused method args

jeremylenz:
[test katello]

jeremylenz:
[test katello]

sjha4:
The ~ operator seems to be acting weird with content_views..

Screenshot from 2023-10-13 13-03-13
Screenshot from 2023-10-13 13-02-51
Screenshot from 2023-10-13 13-02-24

jeremylenz:
Updated to only suggest the `=` operator, per our discussion :+1:

sjha4 (approved):
Ack.. 👍🏼 
Works as expected with = operator. Ignoring other operators for now and as discussed offline, those can be revisited if someone needs it and it's deemed worthwhile to add custom operator queries for those..

PR #10772, opened 2023-10-19
Fixes #36846 - Override REX job names for package install by search

jeremylenz:

Previously, REX job names for bulk package installs were Katello database IDs, e.g. the description for the job would be "Install package(s) id ^ (2196)" or some such.

This change overrides that default description to actually display the package names, _in most cases._ (see considerations below)

image

Descriptions have been updated for

- bulk update
- bulk install
- bulk remove

There are some .. nuances here. As you may know, we have a `useBulkSelect` React hook that allows us to store information about which items are selected. It can either be in "normal" mode, or "Select all" mode. The behavior changes depending on the mode.

"Normal" mode - A set of numeric IDs is stored (`selectionSet`), representing every item the user has selected. (These are typically the database ID, not the table row number.) This is how we start out. You can also click "select page" to select all items on the current page, and you'll remain in normal mode. Conveniently, when using `selectionSet`, we also record the selected items' other data, including package names in this case, and store it in a variable called `selectedResults`.

"Select all" mode - You can click the Select All checkbox to select all items on all pages. Since the front end hasn't yet downloaded the list of all the IDs for all pages, we don't yet know the complete list-- we just know that the user has indicated to select everything. When we enter "Select all" mode, we stop using the `selectionSet` and just say, "ok, everything is selected."  But what if the user now wants to _deselect_ some items? We then have to start populating an `exclusionSet`, which you can think of as the opposite of `selectionSet`. It's basically saying that the user has selected everything _except_ these IDs. However, when using `exclusionSet` we don't currently have the benefit of having stored `selectedResults`.

Long story short: In this PR, we take advantage of `selectedResults` to provide nice REX job descriptions _where we can._ So what about "Select all" mode?

* Bulk update: If you've actually selected everything, the message will read 'Update all packages.' Otherwise, we don't know the package names of the ones you've deselected, so the message will just read 'Update lots of packages.'
* Bulk install: Technically "Select all" mode works if you want to install _all_ of the thousands of packages available to a host. But that's, like, insane and real users would never do that, right? so I didn't do anything special here. I think it would just say 'Install package(s)'
* Bulk remove: "Select all" mode is actually disabled for bulk remove (we _do_ have limits!) so no worries here.

You do have to set up REX to test this. But you don't actually have to be able to communicate with the host.

Test all of the scenarios and make sure you get the expected messages


lfu:
Normal mode:
image

All mode:
image

lfu (approved):
LGTM 👍🏻

jeremylenz:
fixed tests

jeremylenz:
[test katello]

PR #10775, opened 2023-10-24
Fixes #36857 - Run content count update on smart proxy after orphan cleanup

sjha4:
Trigger a content count update task after running orphan cleanup
1. Have a proxy and some content views/envs tied to it and sync the capsule.
2. Check the result from endpoint `/katello/api/capsules/:id/content/counts`
3. Remove an env from capsule and don't sync.
4. Run `bundle exec rails katello:delete_orphaned_content`
5. Check the result from endpoint `/katello/api/capsules/:id/content/counts`
6. The content_counts will remain the same.
7. Checkout this PR.
8. Run `bundle exec rails katello:delete_orphaned_content`
9. The content_counts should get updated to reflect only envs still tied to smart_proxy.

### DIFF
--- a/app/lib/actions/katello/orphan_cleanup/remove_orphans.rb
+++ b/app/lib/actions/katello/orphan_cleanup/remove_orphans.rb
@@ -15,7 +15,15 @@ module Actions
               plan_action(
                 Actions::Pulp3::Orchestration::OrphanCleanup::RemoveOrphans,
                 proxy)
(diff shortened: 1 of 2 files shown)


sjha4:
Need to update the update_content_count method to update content_count instead of returning when repos is empty on capsule. Will do after either merging the other PR or will need to rebase.

ianballou (approved):
Works great for me!

PR #10776, opened 2023-10-25
Fixes #36859 - Fix inconsistent repo publication

m-bucher:
Fix repo not having correct publication due to e.g. killed pulp-publication task.
Or rather make sure next Optimized Sync will fix this, by identifying the inconsistency.

### DIFF
--- a/app/lib/actions/pulp3/repository/save_version.rb
+++ b/app/lib/actions/pulp3/repository/save_version.rb
@@ -34,7 +34,12 @@ module Actions
               repo.update(:version_href => version_href)
             end
           else
-            output[:contents_changed] = false
+            # get publication and check if repo-version and publication match. Otherwise, contents_changed: false
+            if !repo.repository_type.pulp3_skip_publication && fetch_current_published_version_href(repo) != repo.version_href
+              output[:contents_changed] = true
+            else
+              output[:contents_changed] = false
+            end
           end
         end
 
@@ -43,6 +48,13 @@ module Actions
           repo_backend_service = repo.backend_service(SmartProxy.pulp_primary)
           repo_href = repo_backend_service.repository_reference.repository_href
           repo_backend_service.api.repositories_api.read(repo_href).latest_version_href
+        end
+
+        def fetch_current_published_version_href(repo)
+          # Fetch latest Pulp 3 repo version
+          return nil if repo.publication_href.nil?
+          repo_backend_service = repo.backend_service(SmartProxy.pulp_primary)
+          repo_backend_service.api.publications_api.read(repo.publication_href).repository_version
         end
       end
     end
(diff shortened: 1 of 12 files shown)


quba42:
> Or rather make sure next Optimized Sync will fix this, by identifying the inconsistency.

This is not actually related to optimized vs. complete sync. Both types won't repair an affected repo without this change, unless there is new content in the upstream repo being synced. If there is no new content in the upstream, both types of sync will not create a new pulp repo version (full sync will simply take longer to not create a new repo version). And in both cases Katello will not republish if there is no new repo version.

test/actions/pulp3/repository/save_versions_test.rb
```diff
@@ -81,5 +81,24 @@ def test_save_new_version_from_lookup
       assert_equal task.output[:contents_changed], true
       assert_empty task.output[:updated_repositories] - [@repo1.id, @repo2.id]
     end
+
+    def test_save_version_with_outdated_publication
```
m-bucher, line 85: Is this the correct file for the test?

I kinda doubt it, but it seemed to be the best fir without creating a new test-class  :innocent:

sbernhard:
Can you please have a look at this @jeremylenz ? Any thoughts as we think, this should fix a pretty bad bug.

jeremylenz:
Added a review card to our sprint, thanks!

quba42:
For the benefit of any potential reviewers: Note that I put fairly detailed reproducer instructions on the accompanying issue: https://projects.theforeman.org/issues/36859

test/actions/pulp3/repository/save_versions_test.rb
ianballou: Works for me!

app/lib/actions/pulp3/repository/save_version.rb
```diff
@@ -34,7 +34,12 @@ def run
               repo.update(:version_href => version_href)
             end
           else
-            output[:contents_changed] = false
+            # get publication and check if repo-version and publication match. Otherwise, contents_changed: false
+            if fetch_current_published_version_href(repo) != repo.version_href
```
ianballou, line 38: Since not all repository types have publications, this is causing `contents_changed` to be true for all syncs of those types. One example of a type without publications is the container/docker type.

I'd recommend checking if `repo.repository_type.pulp3_skip_publication` to fix the issue.

ianballou (changes requested):
Thanks for the contribution! Just one bug found:

app/lib/actions/pulp3/repository/save_version.rb
quba42: Well spotted! I did not consider there are types without publications.

m-bucher: Thanks, added a condition.

m-bucher:
katello_test

ianballou (approved):
Working for me! I tested it by editing the `publication_href` for a repository to a new one that I created. After syncing, the repository's `publication_href` was returned to normal.

PR #10780, opened 2023-10-27
Fixes #37095 - Use shared GitHub workflow to run tests

evgeni:

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -1,0 +1,27 @@
+---
+name: CI
+
+on:
+  pull_request:
+  push:
+    branches:
+      - 'master'
+      - 'KATELLO-*'
+
+concurrency:
+  group: ${{ github.ref_name }}-${{ github.workflow }}
+  cancel-in-progress: true
+
+jobs:
+  rubocop:
+    name: Rubocop
+    uses: theforeman/actions/.github/workflows/rubocop.yml@v0
+
+  test:
+    name: Ruby
+    needs: rubocop
+    uses: theforeman/actions/.github/workflows/foreman_plugin.yml@v0
+    with:
+      plugin: katello
+      postgresql_container: ghcr.io/theforeman/postgresql-evr
+      test_existing_database: false
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,7 @@
 source 'https://rubygems.org'
 
 gemspec
+
+Dir[File.join(__dir__, 'gemfile.d', '*.rb')].each do |bundle|
+  eval_gemfile(bundle)
+end
--- a/gemfile.d/test.rb
+++ b/gemfile.d/test.rb
@@ -1,0 +1,4 @@
+group :test do
+  gem "vcr", "~> 6.1"
+  gem 'theforeman-rubocop', '~> 0.0.6', require: false
+end
--- a/katello.gemspec
+++ b/katello.gemspec
@@ -64,14 +64,4 @@ Gem::Specification.new do |gem|
   # UI
   gem.add_dependency "deface", '>= 1.0.2', '< 2.0.0'
   gem.add_dependency "angular-rails-templates", "~> 1.1.0"
-
-  # Testing
-  gem.add_development_dependency "factory_bot_rails", "~> 4.5"
-  gem.add_development_dependency "mocha"
-  gem.add_development_dependency "vcr", "~> 6.1"
-  gem.add_development_dependency "webmock"
-  gem.add_development_dependency "robottelo_reporter"
-
-  # Rubocop
-  gem.add_development_dependency 'theforeman-rubocop', '~> 0.0.6'
 end


ekohl:
Looks like RuboCop isn't in the dependencies. I assume previously it ran from the Foreman codebase. On the DB side we're missing the EVR extension.

katello.gemspec
```diff
@@ -67,13 +67,13 @@ Gem::Specification.new do |gem|
...
-  gem.add_development_dependency "simplecov-rcov"
-  gem.add_development_dependency "robottelo_reporter"
+  #gem.add_dependency "factory_bot_rails", "~> 4.5"
+  gem.add_dependency "minitest-tags"
+  gem.add_dependency "mocha"
+  gem.add_dependency "vcr", "< 4.0.0"
+  gem.add_dependency "webmock"
+  gem.add_dependency "rubocop-checkstyle_formatter"
+  gem.add_dependency "simplecov"
+  gem.add_dependency "simplecov-rcov"
```
evgeni, line 77: That's a terrible hack, and I wonder how to "correctly" fix this.

`add_development_dependency` adds things to the `development` group of Bundler, but we do not install this group as we only want to run tests. There is no `add_test_dependency`.

I see three possible solutions, and not quite sure which one I like most:
1. install `development` group (and pull in all those really dev-only deps of foreman, meh)
2. set `development_group` to something else (like `${plugin}_dev`) when including the plugin in our actions, this would install all plugin dev deps as the new group is not excluded
3. provide a `gemfile.d/test.rb` that clearly defines the test deps in the `test` group.

I think *3* is the cleanest, but will also require the most changes (to every plugin that needs to define test deps)

evgeni: 2 is what Jenkins does today, btw:
https://github.com/theforeman/jenkins-jobs/blob/6dd3bfe1fa010ecbb4750d9e909246a148c410fb/theforeman.org/pipelines/lib/foreman.groovy#L7

katello.gemspec
```diff
@@ -66,14 +66,6 @@ Gem::Specification.new do |gem|
...
-  gem.add_development_dependency "minitest-tags"
-  gem.add_development_dependency "mocha"
-  gem.add_development_dependency "vcr", "< 4.0.0"
-  gem.add_development_dependency "webmock"
-  gem.add_development_dependency "rubocop-checkstyle_formatter"
-  gem.add_development_dependency "simplecov"
-  gem.add_development_dependency "simplecov-rcov"
-  gem.add_development_dependency "robottelo_reporter"
+  # Rubocop
+  gem.add_development_dependency 'theforeman-rubocop', '~> 0.0.6'
```
ekohl, line 70: We have 0.1.2 as well, which is actually new enough for GH annotations.
```suggestion
  gem.add_development_dependency 'theforeman-rubocop', '~> 0.1.0'
```

gemfile.d/test.rb
```diff
@@ -0,0 +1,7 @@
+group :test do
+  gem "minitest-tags"
+  gem "vcr", "< 4.0.0"
+  gem "rubocop-checkstyle_formatter"
```
ekohl, line 4: AFAIK we don't actually need this anymore. This used to be needed for Jenkins, but nowadays it can parse the normal progress format. Also, `theforeman-rubocop` still pulls it in.

Gemfile
```diff
@@ -1,3 +1,7 @@
 source 'https://rubygems.org'
 
 gemspec
+
+Dir[File.join(__dir__, 'gemfile.d', '*.rb')].each do |bundle|
+  eval_gemfile(bundle)
```
ekohl, line 6: It is a shame dependabot doesn't understand this, where it did understand the previous format.

evgeni: in this very particular case, we could just write `eval_gemfile('gemfile.d/test.rb')` and Dependabot would work, but otherwise I agree.

katello.gemspec
evgeni: yeah, it was just a test to pull something in, and then I erred on the side of "what's in foreman today"

.github/workflows/ruby.yml
```diff
@@ -0,0 +1,24 @@
...
+  cancel-in-progress: true
+
+jobs:
+  rubocop:
+    name: Rubocop
+    uses: theforeman/actions/.github/workflows/rubocop.yml@v0
+
+  test:
+    name: Ruby
+    uses: theforeman/actions/.github/workflows/foreman_plugin.yml@evr
```
ekohl, line 22: We merged this and deleted the branch. There is now only a parameter for it

ekohl (changes requested):

gemfile.d/test.rb
```diff
@@ -0,0 +1,7 @@
+group :test do
+  gem "minitest-tags"
```
ekohl, line 2: This hasn't had an update since 2012. Granted, it's tiny but I wonder if anyone actually uses this. It's only for developers to run specific tests.

gemfile.d/test.rb
```diff
@@ -0,0 +1,7 @@
...
+  gem "vcr", "< 4.0.0"
+  gem "rubocop-checkstyle_formatter"
+  gem "simplecov"
+  gem "simplecov-rcov"
```
ekohl, line 6: I was considering a patch to make simplecov loading optional. If we don't publish the coverage, is there any point to having it? Or really, does anyone look at it?

.github/workflows/ruby.yml
```diff
@@ -0,0 +1,25 @@
...
+  cancel-in-progress: true
+
+jobs:
+  rubocop:
+    name: Rubocop
+    uses: theforeman/actions/.github/workflows/rubocop.yml@v0
+
+  test:
+    name: Ruby
+    uses: theforeman/actions/.github/workflows/foreman_plugin.yml@custom-db
```
ekohl, line 22: https://github.com/theforeman/actions/commit/e4e8fb7c0e9907b6e0859306c4efd4240b763109 was merged
```suggestion
    uses: theforeman/actions/.github/workflows/foreman_plugin.yml@v0
```

gemfile.d/test.rb
ehelms: My vote is to remove simplecov. @ianballou @parthaa please weigh in

gemfile.d/test.rb
```diff
@@ -0,0 +1,7 @@
+group :test do
```
ehelms, line 1: Why move these into this group in a Gemfile rather than keep them where they are?

evgeni: Today, they are defined as development dependencies, but we do not install the "development" group in CI as that would be silly (we don't need pry and friends from https://github.com/theforeman/foreman/blob/develop/bundler.d/development.rb).
By moving them to a bundler.d file, we can define them as a group (you can't do that in the gemspec) and install cleanly.

evgeni: More in my old comment at https://github.com/Katello/katello/pull/10780#pullrequestreview-1752348091

gemfile.d/test.rb
ekohl: Proposing to drop it: https://github.com/Katello/katello/pull/10836

gemfile.d/test.rb
ekohl: Let's move that over to https://github.com/Katello/katello/pull/10837.

gemfile.d/test.rb
ekohl: Fun fact: we don't even use `katello:rubocop:jenkins` today and don't even parse the results. In https://github.com/theforeman/jenkins-jobs/pull/400 I switch over to properly parsing the results.

Dropping it in https://github.com/Katello/katello/pull/10838

db/migrate/20160404132250_remove_katello_from_notification_name.rb
```diff
@@ -1,26 +1,8 @@
 class RemoveKatelloFromNotificationName < ActiveRecord::Migration[4.2]
-  class FakeMailNotification < ApplicationRecord
-    self.table_name = 'mail_notifications'
-  end
-
   def up
-    FakeMailNotification.all.each do |notification|
-      if notification_names.keys.include?(notification.name)
```
ekohl, line 8: Another approach could be to filter with a `where` (so pushing it to the DB instead of client side). That would be an empty list in most cases and not cause it to break.

gemfile.d/test.rb
ehelms: I agree and disagree. These testing frameworks are, I would argue, part of development. We shouldn't divorce testing from development.

evgeni: But are we divorcing?
`bundle install` still installs them by default, we just can avoid *pure* dev deps (pry) by using "without development" config if we want to (in CI).
And this follows the same pattern we have in core.

ekohl: Technically development dependencies in a gemspe end up in a bundler group called development. This just further split it.

It should be noted you can no longer load them via gemspec in a bundler file. Perhaps that's something that will show up in gems that depend on katello and want to use its tests

ianballou:
When a redmine gets created for this, mind linking this redmine to the new one? https://projects.theforeman.org/issues/36998

db/migrate/20160505181337_rename_katello_settings.rb
```diff
@@ -1,9 +1,9 @@
 class RenameKatelloSettings < ActiveRecord::Migration[4.2]
   def up
-    Setting.where(category: 'Setting::Katello').update_all(:category => 'Setting::Content')
+    Setting.where(category: 'Setting::Katello').update_all(:category => 'Setting::Content') if column_exists?(:settings, :category)
```
evgeni, line 3: https://github.com/Katello/katello/pull/10852

db/migrate/20131014225132_add_users_fields.rb
```diff
@@ -3,7 +3,6 @@ def change
     add_column :users, :helptips_enabled, :boolean, :default => true
     add_column :users, :hidden, :boolean, :default => false, :null => false
     add_column :users, :page_size, :integer, :default => 25, :null => false
-    add_column :users, :disabled, :boolean, :default => false
```
evgeni, line 6: https://github.com/Katello/katello/pull/10853

db/migrate/20160404132250_remove_katello_from_notification_name.rb
evgeni: technically, yeah. but it felt weird to spend time on "fixing" this where it really is not needed anymore.

https://github.com/Katello/katello/pull/10854

evgeni:
> When a redmine gets created for this, mind linking this redmine to the new one? https://projects.theforeman.org/issues/36998

I've split out the Rubocop related fixes into https://github.com/Katello/katello/pull/10857 for easier review.

gemfile.d/test.rb
evgeni: I've opened https://github.com/theforeman/jenkins-jobs/pull/410 to at least fix Jenkins support for this setup.

evgeni:
[test katello]

evgeni:
[test katello]

evgeni:
[test katello]

ekohl:
Because the issue is getting quite long, a list of split off PRs:
* https://github.com/Katello/katello/pull/10836
* https://github.com/Katello/katello/pull/10837
* https://github.com/Katello/katello/pull/10852
* https://github.com/Katello/katello/pull/10853
* https://github.com/Katello/katello/pull/10854
* https://github.com/Katello/katello/pull/10857

katello.gemspec
```diff
@@ -64,18 +64,4 @@ Gem::Specification.new do |gem|
   # UI
   gem.add_dependency "deface", '>= 1.0.2', '< 2.0.0'
   gem.add_dependency "angular-rails-templates", "~> 1.1.0"
-  gem.add_development_dependency "uglifier"
```
evgeni, line 67: https://github.com/Katello/katello/pull/10859

gemfile.d/test.rb
evgeni: @ehelms do the recent change to move all devel-like deps to `gemfile.d/test.rb` satisfy your need for "consistency"?

evgeni:
This is now :green_apple: and ready!

gemfile.d/test.rb
ehelms: I sense snark in the force. And yes it does. Thank you.

.github/workflows/ruby.yml
```diff
@@ -0,0 +1,26 @@
...
+    name: Rubocop
+    uses: theforeman/actions/.github/workflows/rubocop.yml@v0
+
+  test:
+    name: Ruby
+    uses: theforeman/actions/.github/workflows/foreman_plugin.yml@v0
+    with:
+      plugin: katello
+      postgresql_container: ghcr.io/theforeman/postgresql-evr
+      test_existing_database: false
```
evgeni, line 26: One thing to consider for branching. When Katello gets branched, you'll have to add a `foreman_version: X.Y-stable` here, as this is what's driving which Foreman branch is used for testing, and you want them to match ;-)

ehelms: Can that be handled somehow centrally through some mapping or matrix file?

evgeni: There is no such thing today (which of course doesn't mean we can't invent it…)

Today, this mapping lives in https://github.com/theforeman/jenkins-jobs/blob/master/theforeman.org/pipelines/test/testKatello.groovy#L1 and needs to be edited by whoever does the branching. So it's not a workflow change as such.

ianballou: @wbclark @qcjames53 assuming this merges as-is, we'll need a new branching step to update the `foreman_version` in this `ruby.yml` file.

ianballou (commented):
Does this require a matching Foreman PR to work? I'm hitting issues loading `vcr` now when running tests locally:

```
vagrant@centos8-katello-devel-stable ~/katello $ ktest test/services/katello/pulp3/rpm_test.rb 
/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activerecord-6.1.7.6/lib/active_record/connection_adapters/postgresql_adapter.rb:883: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/pg-1.5.4/lib/pg/text_decoder/timestamp.rb:8: warning: The called method `initialize' is defined here
Emptying /home/vagrant/foreman/jenkins/reports/unit
   -> 0.0003s
	19: from -e:1:in `<main>'
	18: from -e:1:in `each'
	17: from -e:1:in `block in <main>'
	16: from -e:1:in `require'
	15: from /home/vagrant/katello/test/services/katello/pulp3/rpm_test.rb:1:in `<top (required)>'
	14: from /home/vagrant/katello/test/services/katello/pulp3/rpm_test.rb:1:in `require'
	13: from /home/vagrant/katello/test/katello_test_helper.rb:12:in `<top (required)>'
	12: from /home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activesupport-6.1.7.6/lib/active_support/dependencies.rb:332:in `require'
	11: from /home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activesupport-6.1.7.6/lib/active_support/dependencies.rb:299:in `load_dependency'
	10: from /home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activesupport-6.1.7.6/lib/active_support/dependencies.rb:332:in `block in require'
	 9: from /home/vagrant/foreman/.vendor/ruby/2.7.0/gems/zeitwerk-2.6.12/lib/zeitwerk/kernel.rb:38:in `require'
	 8: from /home/vagrant/foreman/.vendor/ruby/2.7.0/gems/polyglot-0.3.5/lib/polyglot.rb:65:in `require'
	 7: from /home/vagrant/foreman/.vendor/ruby/2.7.0/gems/polyglot-0.3.5/lib/polyglot.rb:65:in `require'
	 6: from /home/vagrant/katello/test/support/vcr.rb:1:in `<top (required)>'
	 5: from /home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activesupport-6.1.7.6/lib/active_support/dependencies.rb:332:in `require'
	 4: from /home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activesupport-6.1.7.6/lib/active_support/dependencies.rb:299:in `load_dependency'
	 3: from /home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activesupport-6.1.7.6/lib/active_support/dependencies.rb:332:in `block in require'
	 2: from /home/vagrant/foreman/.vendor/ruby/2.7.0/gems/zeitwerk-2.6.12/lib/zeitwerk/kernel.rb:38:in `require'
	 1: from /home/vagrant/foreman/.vendor/ruby/2.7.0/gems/polyglot-0.3.5/lib/polyglot.rb:65:in `require'
/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/polyglot-0.3.5/lib/polyglot.rb:65:in `require': cannot load such file -- vcr (LoadError)
```

evgeni:
> Does this require a matching Foreman PR to work? I'm hitting issues loading `vcr` now when running tests locally:

Ah, this is a good point.
You (or whatever sets up the devel box) would need to know about gemfile.d too.

https://github.com/theforeman/actions/blob/v0/.github/workflows/foreman_plugin.yml#L103

a `cat` like this could do (or a symlink, then changes to the file would propagate automatically)

ianballou:
> a `cat` like this could do (or a symlink, then changes to the file would propagate automatically)

Sounds puppet-katello_devel should handle this https://github.com/theforeman/puppet-katello_devel

ianballou (approved):
I'm cool with merging this in -- the necessary additions to katello.gemspec.local are likely to cause a bit of confusion among katello devs, but I suppose it doesn't make sense to wait for a new puppet-katello_devel PR first.

PR #10781, opened 2023-10-27
Fixes #36871 - Fix incremental update concurrency

m-bucher:
Do not run Incremental Updates on the same Content View in parallel

### DIFF
--- a/app/lib/actions/katello/content_view/incremental_updates.rb
+++ b/app/lib/actions/katello/content_view/incremental_updates.rb
@@ -4,28 +4,41 @@ module Actions
       class IncrementalUpdates < Actions::EntryAction
         include Helpers::Presenter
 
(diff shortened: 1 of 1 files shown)


ianballou:
Tested on Katello 4.7 and seems to be working fine. Tested with normal CVs and composite CVs.  Going to test now on the latest code.

ianballou (approved):
Working fine for me! Tested again on the latest Katello code. Created a CV with two versions that contain the same repo with different content. The content difference was created via RPM deletion in the UI.  Each content view version has a different LCE.

I subscribed one host to each LCE and ensured that each had one applicable erratum to apply. I then ran the incremental update via the UI. The incremental update generate two new content view versions, each had the appropriate content in it. I also confirmed that applying the new content to the host after incremental update still works.

PR #10783, opened 2023-10-31
Fixes #36866 - Remove table hooks from katello and use foreman

parthaa:

Now that TableHooks.js code has been ported to Foreman, we need to have Katello consume it from there so code isn't duplicated. This PR does the clean up / walden job 

* Check out this PR
* Try some pages that use Table hooks. 
* Make sure they look ok


jeremylenz:
@parthaa Do we know if Foreman TableHooks has diverged at all from Katello?

chris1984:
@jeremylenz do you mind reviewing the code, while I test the pages to make sure there are not any breakages?

webpack/components/extensions/HostDetails/Tabs/ErrataTab/ErrataTab.js
```diff
@@ -228,7 +229,7 @@ export const ErrataTab = () => {
   });
 
   const installErrataAction = () => installErrata({
-    hostname, search: fetchBulkParams(),
+    hostname, search: fetchBulkParams({}),
```
jeremylenz, line 232: The need to pass `{}` here is eliminated in https://github.com/theforeman/foreman/pull/9878, but I suppose it's harmless either way. If that gets merged first, you could remove all of these.

jeremylenz (commented):
Code LGTM :+1:

chris1984:
@sjha4 said he would test the CV pages, working on testing the other stuff that was changed

chris1984 (approved):
ACK works well, all the tables/pages/modals render that I tested

webpack/components/extensions/HostDetails/Tabs/ErrataTab/ErrataTab.js
parthaa: Agreed!. May be I should wait for your PR before merging

PR #10784, opened 2023-11-01
Fixes #36884 - Make more strings translatable and extract strings for Katello 4.11

ianballou:
Besides extracting strings, this PR makes (what should be) the remaining unextractable strings extractable.  Angular breadcrumb labels use the translate filter to become translated, but the gettext grunt extraction task cannot pick them up for shipment to transifex. That means that, if translations are already existing in the locale files, the breadcrumbs get translated, but if not, they go completely untranslated.

To work around this, I've added variables containing the needed strings in related controllers with some comments to help future maintainers. This is a short term workaround for the remaining time that Angular has in our ecosystem.

To test, just poke around the UI and make sure Angular breadcrumbs still look fine. They shouldn't be affected unless I messed up a translation import somewhere.


engines/bastion_katello/app/assets/javascripts/bastion_katello/activation-keys/activation-keys.controller.js
```diff
@@ -19,6 +19,9 @@ angular.module('Bastion.activation-keys').controller('ActivationKeysController',
     ['$scope', '$location', 'translate', 'Nutupane', 'ActivationKey', 'CurrentOrganization', 'experimentalLabsSetting',
     function ($scope, $location, translate, Nutupane, ActivationKey, CurrentOrganization, experimentalLabsSetting) {
 
+        // Labels so breadcrumb strings can be translated
```
chris1984, line 22: Thanks for leaving comments :)

chris1984 (approved):
Looks good, send it.

ianballou:
Just need to fix the tests

ianballou:
@chris1984 my latest push should fix all the broken tests.

chris1984 (approved):
Code looks fine, thanks for taking the time to do this. Went through and checked a few pages and the breadcrumbs look fine. ACK

Thanks for putting comments in the JS Controllers too

PR #10786, opened 2023-11-01
Fixes #36883 - Wrong date when using an errata by date filter

sjha4:
Updates CV -> Filters -> Errata by date range filter to ignore timezone by adding timezone offset to time to show times in UTC instead of local timezone
Create an errata by date filter.
In the console, update the date to something like this : cvf.start_date = "2023-10-01"
On the UI, the date will show up as 9/30 instead of 10/01
With this change, you should see 9/30


ianballou (approved):
Looks good to me!

PR #10789, opened 2023-11-06
Fixes #36873 - Store all env_ids for smart_proxy complete sync in task

sjha4:
Added a new environment_ids value in capsule sync task input which stores all env_ids o proxy when a full sync task is started

1. Sync a smart proxy and add some env/cvs and run a full sync.
2. Go to smart proxy > content tab and check sync time against envs.
3. Wait a minute and then run an env specific capsule sync task from hammer : ex: `hammer capsule content synchronize --id 2 --lifecycle-environment qa --organization-id 1`
4. Go to smart proxy > content tab and check sync time against env you ran sync on. it should show latest task and others should show older complete task.
5. Add a new env to smart proxy from edit smart proxy action.
6. Go back to content tab for proxy and confirm that you see N/A for sync time against the env.

### DIFF
--- a/app/controllers/katello/api/v2/capsule_content_controller.rb
+++ b/app/controllers/katello/api/v2/capsule_content_controller.rb
@@ -73,12 +73,16 @@ module Katello
       find_content_view if params[:content_view_id]
       find_repository if params[:repository_id]
       skip_metadata_check = ::Foreman::Cast.to_bool(params[:skip_metadata_check])
+      sync_options = {
+        :environment_id => @environment.try(:id),
+        :content_view_id => @content_view.try(:id),
+        :repository_id => @repository.try(:id),
+        :skip_metadata_check => skip_metadata_check
+      }
+      sync_options[:environment_ids] = @capsule.lifecycle_environments&.pluck(:id) unless (@environment || @content_view || @repository)
       task = async_task(::Actions::Katello::CapsuleContent::Sync,
                         @capsule,
-                        :environment_id => @environment.try(:id),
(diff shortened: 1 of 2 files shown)


ianballou:
I created a fresh new environment and promoted it and then added it to my smart proxy. Immediate it says that there is a last sync "about 3 hours ago".

I think it happened because there were no `environment_ids` and no other IDs either:

```
options:
  environment_id: 
  content_view_id: 
  repository_id: 
```

Is it better to assume that the env was synced, or just say N/A ?

ianballou:
After syncing with this code and then making a new env + adding to the smart proxy, then I see the N/A properly.

sjha4:
> Is it better to assume that the env was synced, or just say N/A if there is no `environment_ids` data?

For this release, it's better to assume env was synced (In the case of new envs CVs will show as not synced) because all complete sync tasks till the release will have no env_ids in the task and show all envs as N/A even if they are synced.

ianballou:
I also noticed after deleting all tasks that the  content views are showing up as synced
image

It seems a bit misleading since we cannot tell if it was actually synced. Would this be a good spot for another N/A ?

sjha4:
>> It seems a bit misleading since we cannot tell if it was actually synced. Would this be a good spot for another N/A ?

I'd err on the side of showing CV as synced with the content counts if we have the data available. N/A for task details makes sense in this case after task cleanup since we no longer have that info..The CV itself being synced is still accurate info here based on smart_proxy_sync_history..

ianballou (approved):
Looking good to me!

PR #10794, opened 2023-11-09
Fixes #36911 - Org scoping with container tags

sjha4:
Add scoping by org to Docker tags on UI which includes products, repos, env etc.
Have 2 or more organizations
In each, org create the same docker repo and possibly a couple unique docker repos.
Using org switcher, switch orgs and make sure, the tags and product links, environment and repositories tab on tag details are scoped by organization.


app/views/katello/api/v2/docker_tags/_base.json.rabl
```diff
@@ -18,8 +18,26 @@ child :docker_manifest => :manifest do
   attributes :schema_version, :digest, :manifest_type
 end
 
-child :repositories => :repositories do
```
ianballou, line 21: Can we retain the `:repositories` list even if no organization is specified? Asking since this will be a breaking API change.

ianballou (commented):
Working well, just one question:

app/views/katello/api/v2/docker_tags/_base.json.rabl
sjha4: Good point..Updated..

ianballou (approved):
Looking fine to me!

sjha4:
[test katello]

PR #10795, opened 2023-11-10
Fixes #36557 - ouiaId does not apply to PF3

lfu:
Remove `ouiaId` from PF3 components.
Make sure `@theforeman/eslint-plugin-rules` `12.2.0` is installed.
`npx eslint webpack` passes.
No error in `console` like:
```
react-dom.development.js:88 Warning: React does not recognize the `ouiaId` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `ouiaid` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "@testing-library/user-event": "^13.5.0",
     "@theforeman/builder": ">= 6.0.0",
     "@theforeman/find-foreman": "^4.8.0",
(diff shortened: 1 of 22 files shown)


jeremylenz (approved):
Green tests mean this is doing the needful. Thanks @lfu !

ACK :+1:

PR #10796, opened 2023-11-13
Fixes #36914 - skip missing hosts during applicability generate

ianballou:
Adds error handling to the block in BulkGenerate that calculates applicability for all hosts.

The issue here is that we cannot use resource locking because it's a list of hosts. So, we can't stop hosts from being deleted during applicability calculation.

We thought we handled this before with am existence check, but turns out some folks still hit the issue because the host is deleted part way  through `calculate_and_import_applicability`.

NoMethodError and all PG errors are handled here to cover our bases but not be too broad. Plus the errors are raised back to the logs so debugging shouldn't be too much harder.

Also makes bulk generate skip itself on an error so the task doesn't get paused.  

I don't see any reason why applicability generation needs to result in a paused task. I don't see it causing a data integrity issue.

1) Edit the code to error out somewhere in `calculate_and_import_applicability`
2) Trigger applicability generation for a host, ideally the "natural" way via subscription-manager
3) See that the error is handled and that it pops up in the logs.

### DIFF
--- a/app/lib/actions/katello/applicability/hosts/bulk_generate.rb
+++ b/app/lib/actions/katello/applicability/hosts/bulk_generate.rb
@@ -8,14 +8,24 @@ module Actions
           end
 
           def run
+            error = false
             input[:host_ids].each do |host_id|
               content_facet = ::Katello::Host::ContentFacet.find_by_host_id(host_id)
               if content_facet.present?
+                # Catch errors and log them, but continue processing the rest of the hosts
                 content_facet.calculate_and_import_applicability
               else
                 Rails.logger.warn(_("Content Facet for host with id %s is non-existent. Skipping applicability calculation.") % host_id)
               end
+            rescue NoMethodError, PG::Error => e
+              Rails.logger.error("Error calculating applicability for host #{host_id}: #{e.message}")
+              error = true
             end
+            fail "Error calculating applicability for one or more hosts" if error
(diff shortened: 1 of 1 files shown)


app/lib/actions/katello/applicability/hosts/bulk_generate.rb
```diff
@@ -8,14 +8,20 @@ class BulkGenerate < Actions::EntryAction
...
+            error = false
             input[:host_ids].each do |host_id|
               content_facet = ::Katello::Host::ContentFacet.find_by_host_id(host_id)
-              if content_facet.present?
-                content_facet.calculate_and_import_applicability
-              else
-                Rails.logger.warn(_("Content Facet for host with id %s is non-existent. Skipping applicability calculation.") % host_id)
-              end
+              # Catch errors and log them, but continue processing the rest of the hosts
+              content_facet.calculate_and_import_applicability
```
sjha4, line 15: Should we still keep the `present?` to not have to fail when host gets deleted before getting here?

ianballou: I'll put that back

sjha4 (approved):
Code looks good. Task now skips and ends upon failure and is able to log errors when error introduced in `calculate_and_import_applicability` method instead of pausing.

PR #10797, opened 2023-11-13
Fixes #36840 - Clear content source when unregistering host

jeremylenz:

* When unregistering (but not destroying the host), reset the host's content source to the Pulp Primary smart proxy.
 * Also correctly display the content source on the host details page (previously it was showing the
  Registered Through smart proxy mislabeled as content source.)

This can happen when you re-register a host using `--force`.

1. Set up an external smart proxy (I know, sorry)
2. Set up a new lifecycle environment, and a content view promoted to that environment
3. Sync this LCE to the smart proxy (but don't sync Library)
4. Register the host
 - to this smart proxy
 - using global registration
 - and an AK that assigns your newly created LCE/CV
6. Register the host again
 - to the Pulp Primary smart proxy (your Foreman server)
 - using global registration
 - and the __Force__ checkbox checked
 - and an AK that assigns the Library / Default Organization View

Before checking out this patch: you'll get an error like

```
Validation failed: Host rhel9b.fedora.example.com: Cannot add content view environment to content facet. The host's content source 'rhel8b.fedora.example.com' does not sync lifecycle environment 'Library'. (HTTP error code 422: Unprocessable Entity)
```

If you re-register again after checking out the patch, you should be able to register without errors, and the host's content facet should show the Foreman server as its content source.

### DIFF
--- a/app/services/katello/registration_manager.rb
+++ b/app/services/katello/registration_manager.rb
@@ -306,6 +306,7 @@ module Katello
           host.content_facet.applicable_errata = []
           host.content_facet.uuid = nil
           host.content_facet.content_view_environments = []
+          host.content_facet.content_source = ::SmartProxy.pulp_primary
           host.content_facet.save!
           Rails.logger.debug "remove_host_artifacts: marking CVEs unchanged to prevent backend update"
           host.content_facet.mark_cves_unchanged
--- a/test/services/katello/registration_manager_test.rb
+++ b/test/services/katello/registration_manager_test.rb
@@ -308,6 +308,19 @@ module Katello
         assert_empty @host.rhsm_fact_values
       end
 
+      def test_unregister_host_resets_content_source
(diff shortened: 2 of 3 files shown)


ianballou:
[test katello]

app/services/katello/registration_manager.rb
```diff
@@ -306,6 +306,7 @@ def remove_host_artifacts(host, clear_content_facet: true)
           host.content_facet.applicable_errata = []
           host.content_facet.uuid = nil
           host.content_facet.content_view_environments = []
+          host.content_facet.content_source = ::SmartProxy.pulp_primary
```
ianballou, line 309: Think this is worth a test? Perhaps in `test/services/katello/registration_manager_test.rb`

ianballou (commented):
Works well for me! Just one comment.

jeremylenz:
@iballou Added test :+1:

ianballou (approved):
Looks good!

PR #10798, opened 2023-11-14
Fixes #36920 - Katello::Reimport fails due to format error

sjha4:
Uses correct format of candlepin cp_product results to find product with specified content_id
Added some comments to the code explaining the cp_product format in candlepin result.
Locally I tested with a box with some redhat repos synced.
In rake console try this:

```
org = Organization.first
cp_products = ::Katello::Resources::Candlepin::Product.all(org.label, [:id, :productContent])
product_hash = cp_products.group_by { |prod| prod['id'] }
prod_content_importer = Katello::ProductContentImporter.new(product_hash)
```

Look up a redhat content id from katello as below:
```
Katello::Content.where(vendor: "Red Hat").pluck(:cp_content_id)
```
Use any id in the list. (We are using redhat content so make sure you have redhat product enabled/synced.)

Then try:
```
prod_content_importer.find_product_for_content("9268")

```
On PR branch, you'll see the right product returned.

### DIFF
--- a/app/services/katello/product_content_importer.rb
+++ b/app/services/katello/product_content_importer.rb
@@ -20,6 +20,47 @@ module Katello
     #    },
     #    "enabled":false
     # }
+    #
+    #Example cp_products json structure (key:values where values are arrays and productContent is an array of contents inside value array)
+    # {
+    #    "230"=>
+    #   [{"id"=>"230",
+    #     "productContent"=>
+    #      [{"content"=>
+    #         {"created"=>"2023-11-01T18:02:54+0000",
+    #          "updated"=>"2023-11-01T18:02:54+0000",
+    #          "uuid"=>"ff8080818b715841018b8c0d76d90190",
+    #          "id"=>"3996",
+    #          "type"=>"yum",
+    #          "label"=>"rhel-7-server-rt-htb-rpms",
+    #          "name"=>"Red Hat Enterprise Linux for Real Time HTB (RHEL 7 Server) (RPMs)",
+    #          "vendor"=>"Red Hat",
+    #          "contentUrl"=>"/content/htb/rhel/server/7/$basearch/rt/os",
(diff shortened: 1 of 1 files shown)


app/services/katello/product_content_importer.rb
```diff
@@ -35,12 +76,13 @@ def add_product_content(product, product_content_json)
     end
 
     def find_product_for_content(content_id)
-      prod = @cp_products.find do |prod_json|
-        prod_json['productContent'].any? do |product_content_json|
+      #Hash.find converts hash into ennuberable iterating over each key,value pair which we can capture based on find result
```
jeremylenz, line 79: :stuck_out_tongue_winking_eye: 

```suggestion
      #Hash.find converts hash into enumerable iterating over each key,value pair which we can capture based on find result
```

jeremylenz (commented):
I wonder if there was some Candlepin change that changed the shape of the data.. anyway, I looked at the BZ and we're not backporting any further than the version the error was encountered, so I think this should be okay. Will test shortly.

jeremylenz (commented):
Tested and works well :+1:

app/services/katello/product_content_importer.rb
```diff
@@ -35,12 +76,13 @@ def add_product_content(product, product_content_json)
...
+      cp_product_id, __cp_product_value = @cp_products.find do |prod_json|
+        prod_json&.last&.first&.[]('productContent')&.any? do |product_content_json|
```
jeremylenz, line 81: Curious why the `&.last` needed to be added?

sjha4: The  `@cp_products.find` create an enumerable in the form an array which is [key, value] for the hash key:value pair and works on that. The last is to target the value in that. Weird formatting really on the return + the way .find iterates over it.

jeremylenz (approved):

PR #10782, opened 2023-10-30
Fixes #36878 - Override host bulk destroy

jeremylenz:
#### What are the changes introduced in this pull request?

We're adding bulk host deletion to the Foreman V2 API, to support the new All Hosts index page. Need to add overrides to Katello to ensure hosts are unregistered and artifacts are properly destroyed.

Also includes some misc. items:

- removed the icon from the 'Change content source' menu item per UX guidance
- Consume bulk host extensions from Foreman instead of duplicating it in Katello

#### What are the testing steps for this pull request?

Check out https://github.com/theforeman/foreman/pull/9878
Delete some hosts and ensure they're properly unregistered and artifacts removed (content facet, etc. -check registration_manager.rb#remove_host_artifacts)

PR #10799, opened 2023-11-15
Fixes #36921 - Host queue not added to sidekiq

hao-yu:
The host queue is added too early during the initialization which is before the settings for Foreman plugins are loaded.

Fix it by removing the setting check because pool size is no longer define here. Since Dynflow migrated to Sidekiq, the
concurrency is defined externally in the Sidekiq workers configuration.

### DIFF
--- a/lib/katello/engine.rb
+++ b/lib/katello/engine.rb
@@ -68,9 +68,7 @@ module Katello
 
     initializer "katello.register_actions", :before => :finisher_hook do |_app|
       ForemanTasks.dynflow.require!
-      if (Setting.table_exists? rescue(false)) && Setting['host_tasks_workers_pool_size'].to_i > 0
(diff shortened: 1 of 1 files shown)


wbclark:
rpm-build test keeps failing due to an issue mentioned at https://github.com/Katello/katello/pull/10782

That should be resolved, but it's still happening when I re-run that step

Maybe I need to re-run all steps for the change to be propagated

wbclark:
[test katello]

wbclark (approved):
That did the trick. I guess the buildroot on fedora copr is associated with the PR, so re-running it didn't fix the issue.

As for the actual functionality, I previously tested this using the katello nightly box on forklift and found that it works fine (devel box doesn't work to test it since it doesn't have `dynflow-sidekiq@worker-hosts-queue-1.service`)

wbclark:
Thanks @hao-yu !

PR #10800, opened 2023-11-15
use yum.tfm.o for packit

evgeni:
koji.katelo.org is not updated anymore


jeremylenz (approved):
APJ :+1:

evgeni:
> Will need a Redmine too.

We are pirates, we need no stinkin' Redmines! :)

PR #10801, opened 2023-11-16
Fixes #36924 - Added breadcrumb bar to the header of subscription details pages

qcjames53:
The subscription details page now features a functional breadcrumb bar allowing the user to change the subscription they are investigating or navigate back to the subscription overview page.

n/a

1. Launch Katello with a few subscriptions.
2. Go to the subscriptions page and click on a subscription to view its details. The header should contain a breadcrumb bar allowing you to both navigate back to the subscription overview page and to switch between possible subscriptions.
3. Ensure similarity between this breadcrumb bar and others in Katello (module stream details page, content view details page, etc).

### DIFF
--- a/webpack/scenes/Subscriptions/Details/SubscriptionDetails.js
+++ b/webpack/scenes/Subscriptions/Details/SubscriptionDetails.js
@@ -59,24 +59,21 @@ class SubscriptionDetails extends Component {
 
     return (
       <div id="subscription-details">
-        {!subscriptionDetails.loading &&
-          <BreadcrumbsBar
-            onSwitcherItemClick={(e, url) => this.handleBreadcrumbSwitcherItem(e, url)}
-            data={{
-              isSwitchable: true,
-              breadcrumbItems: [
-                {
-                  caption: __('Subscriptions'),
-                  onClick: () =>
-                    this.props.history.push('/subscriptions'),
-                },
-                {
-                  caption: String(subscriptionDetails.name || __('Subscription Details')),
-                },
-              ],
-              resource,
-            }}
-          />}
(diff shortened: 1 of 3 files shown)


webpack/scenes/Subscriptions/Details/SubscriptionDetails.js
```diff
@@ -58,75 +58,74 @@ class SubscriptionDetails extends Component {
     }
 
     return (
-      <div id="subscription-details">
-        {!subscriptionDetails.loading &&
-          <BreadcrumbsBar
+      <div style={{ margin: '24px' }}>
+        <div id="subscription-details">
```
jeremylenz, line 62: I see we already have a `SubscriptionDetails.scss` file alongside this one. And the `#subscription-details {` in there indicates that the file applies to this already-existing `<div>` with `id="subscription-details"`.

In general it's preferable to add CSS in `scss` files when possible, instead of style tags. I think maybe we could get rid of the outer div (and style tag) and instead just add a new `margin: 24px;` entry in `SubscriptionDetails.scss`. Thoughts?

jeremylenz (commented):
Tested and works well! left one comment above

webpack/scenes/Subscriptions/Details/SubscriptionDetails.js
qcjames53: That sounds ok to me :+1:

qcjames53: Done!

jeremylenz (approved):
Thanks @qcjames53 !

APJ :+1:

PR #10802, opened 2023-11-16
Fixes #36803 - Add capsule verify content checksum

m-bucher:

### DIFF
--- a/app/controllers/katello/api/v2/capsule_content_controller.rb
+++ b/app/controllers/katello/api/v2/capsule_content_controller.rb
@@ -100,6 +100,14 @@ module Katello
       end
     end
 
+    api :POST, '/capsules/:id/content/verify_content', N_('Validate the checksums of all content.')
+    param :id, :number, :required => true, :desc => N_('Id of the smart proxy')
+    def validate_content
+      find_capsule(true)
+      task = async_task(::Actions::Pulp3::CapsuleContent::ValidateContent, @capsule)
+      respond_for_async :resource => task
+    end
+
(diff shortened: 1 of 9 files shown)


sbernhard (commented):
LGTM

ianballou:
Thanks for the contribution! After a discussion in the katello-atix-pulp community meeting, we came to the conclusion that most users would want this repair for specific repositories synced to the smart proxy rather than for the whole thing. Users might have many other repositories on there of different content types, so the repair could take forever.

Selecting a repository that's available on a smart proxy isn't as easy as selecting a Library repository, but I think we can use content views and lifecycle environments as identifiers.  If a user knows that their EPEL 9 repo in LCE "Dev" and CV "CentOS 9" is failing, then naturally they would want to use that information to specify the repository on the smart proxy that needs fixing. So, perhaps we need a workflow that allows them to select the repository to repair by selecting those 3 items.

An alternative method would be to try presenting a list of repositories that are on the smart proxy and to allow the user to pick one. We might be able to use each repository's `backend_object_name` there to help.  I suppose the repository could be listed by repo name / content view / LCE.

@sjha4 I'm curious what you think about my ideas here.

sjha4:
I agree with @ianballou here..Long running repair task can cause deadlocks with syncs etc. We should keep it to hammer with a --complete flag to make sure it's run only when user is certain about it..On the UI we should scope the repair task to individual repositories with corrupted content.

m-bucher:
OK, so I am closing this one in favour of a more fine-grained solution that has yet to be developed.

PR #10803, opened 2023-11-16
Fixes #36926 - RefreshRepos called for relevant repos only

pmoravec:

For unscoped Capsule Sync, call RefreshRepos only for repos to be synced, not for all that exist on the Capsule. That makes the RefreshRepos step significantly faster (from minutes to seconds). E.g. when I had 500 repos on a Capsule and was about to sync just one or a few repos in unsoped Capsule sync, RefreshRepos execution time dropped from 155s to 3s.

`:repository_ids_list` option can be merged with `:repository`. I left the later one as it is already used, to propose a minimalistic change. But I am open to update `:repository` to :repositories` as one common option and amend existing code.

Testing steps described in #36926 . Another test scenarios worth to check: syncing a LE or CV or repository to Capsule - the behaviour should be unchanged there.

### DIFF
--- a/app/lib/actions/katello/capsule_content/refresh_repos.rb
+++ b/app/lib/actions/katello/capsule_content/refresh_repos.rb
@@ -18,7 +18,8 @@ module Actions
           plan_self(:smart_proxy_id => smart_proxy.id,
                     :environment_id => options[:environment]&.id,
                     :content_view_id => options[:content_view]&.id,
-                    :repository_id => options[:repository]&.id)
+                    :repository_id => options[:repository]&.id,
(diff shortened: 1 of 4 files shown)


sjha4:
Rubocop failures related..

pmoravec:
How I could overlook those indentation errors.. should be fixed now.

sjha4:
New task flow looks good.. 💚  New param makes sense since repository param is already being used for isolating repositories to sync individually. 

Test failures are related.

sjha4 (approved):
Ack..works well..Thanks @pmoravec ! Merging..

PR #10804, opened 2023-11-16
Fixes #36928 - (sorta) handle load balanced smart proxies when registering hosts

jeremylenz:

As we know, currently

- full load balanced setups are not supported
- Smart proxy load balancing is _only_ supported for Pulp content, not other smart proxy services
- When registering a host, you pass the `--serverurl` parameter to `subscription-manager`, which contacts the server you specify for registration.
- In such a case, the host's `registered_through` value is set to the hostname of the load balancer, which is not a smart proxy.
- Therefore, the `registered_through` value won't match the name of any known Smart Proxy.
- Previously, when calculating remote execution smart proxies available to the host, that caused the host's `registered_through` Smart Proxy to be listed as `[]`, breaking Remote Execution on certain setups depending on your settings.

With this change, some methods are added to attempt to surmise
  * when you're registering to a load-balanced smart proxy;
  * when a certain smart proxy is behind a load balancer; and
  * which smart proxies are behind a given load balancer.

In the case of smart proxies, this is determined by comparing the smart proxy's `registration_url` hostname with its `url` hostname. If they're different, we assume it's a load balanced setup.

In the case of host registration, this is still determined (for now) by comparing the request headers to known smart proxy names.

Additionally, we add the following behavior changes when a host is registered through a load balancer:

* Set the host's remote execution `registered_through` to _all_ known smart proxies behind the load balancer
* Set the host's content source to the first found smart proxy behind the load balancer. This should be safe because (theoretically) all smart proxies behind the load balancer should be syncing the same content. It's also better than setting content source to `nil`, because `nil` assumes the host has access to all library content, when actually it only has access to content synced by its smart proxy(ies).

Ideally a load balancer would be a Foreman entity which holds a list of smart proxies, and the user would explicitly tell us which smart proxies go with which load balancer, and we could just refer to that. But that would require a rather large effort, so this is being offered as somewhat of a stopgap to fix the immediate bug. None of the changes here preclude a more proper feature design from happening later.

I actually haven't tested this with a load balancer yet. You may leave my hard-codings in, or ideally set up an actual load balancer and take them out.


app/controllers/katello/api/rhsm/candlepin_proxies_controller.rb
```diff
@@ -273,6 +272,7 @@ def get_parent_host(headers)
       hostnames = headers["HTTP_X_FORWARDED_HOST"]
       host = hostnames.split(/[,,:]/)[0].strip if hostnames
       host || URI.parse(Setting[:foreman_url]).host
+      "loadbalancer.example.com"
```
ehelms, line 275: I assume this is for testing and will get cleaned up.

app/controllers/katello/api/rhsm/candlepin_proxies_controller.rb
```diff
@@ -474,7 +474,21 @@ def client_authorized?
     def update_host_registered_through(host, headers)
       parent_host = get_parent_host(headers)
       host.subscription_facet.update_attribute(:registered_through, parent_host)
-      content_source_id = get_content_source_id(parent_host)
+      set_host_content_source(host, parent_host)
+    end
+
+    def registering_thru_load_balancer?(hostname)
+      # If the request doesn't come from any known smart proxy, it came from a load balancer.
+      get_content_source_id(hostname).nil?
```
ehelms, line 482: Is this a safe assumption? Could you instead check all smart-proxies registration URLs and see if they match the hostname?

app/models/katello/concerns/smart_proxy_extensions.rb
```diff
@@ -121,6 +133,15 @@ def alternate_content_sources
         SmartProxy.joins(:smart_proxy_alternate_content_sources).where('katello_smart_proxy_alternate_content_sources.smart_proxy_id' => self.id)
       end
 
+      def registration_fqdn
+        url = self.setting('Registration', 'registration_url').presence || self.url
+        name == "rhel8b.fedora.example.com" ? "loadbalancer.example.com" : URI.parse(url).host
```
ehelms, line 138: Since you are grabbing the `.host`, you could opt to call this method `registration_host` instead

ehelms:
@ekohl given you had a lot of opinion around the implementation where we used registration_url to indicate a load-balancer rather than introducing a dedicated parameter to indicate load-balancing I think you should review this.

app/controllers/katello/api/rhsm/candlepin_proxies_controller.rb
jeremylenz: Yeah, that would be better. At this point I should probably just change it not to use the header, too..

app/controllers/katello/api/rhsm/candlepin_proxies_controller.rb
jeremylenz: Yeah, this is one of the two "hard codings" I refer to in the description :)

app/controllers/katello/api/rhsm/candlepin_proxies_controller.rb
jeremylenz: Updated to check if hostname is a known load balancer. :+1:

jeremylenz:
Added some tests; marking ready for review.

adamruzicka (commented):
I don't have a setup with load balanced capsules so I can't really test this, but in general this reads like it should do the trick, assuming my mental model of the whole thing is correct.

ehelms (approved):

sayan3296:
@adamruzicka @jeremylenz I have done some testings on a Sat\Cap 6.15 + Haproxy setup with two RHEL clients .. The results are looking great. 

Details can be found in the private comment 11 of the main Bugzilla https://bugzilla.redhat.com/show_bug.cgi?id=2240956

ianballou (approved):
After poking around the rails console on a load-balanced set up, this is looking good to me! Credit to @sayan3296 for the setup.

lib/katello.rb
```diff
@@ -6,6 +6,7 @@
 require "oauth"
 require "gettext_i18n_rails"
 require "foreman-tasks"
+require "foreman_remote_execution"
```
evgeni, line 9: Given https://github.com/Katello/katello/blob/6d8d3ca36e1469d1f7c2c8e180e42467176ac1a4/katello.gemspec#L37
I am rather confident we never have a Katello install w/o REX.

But at the same time it indicates that the way we load stuff differs so heavily in prod/dev/test, that things go weird.

evgeni: https://community.theforeman.org/t/requiring-rex-in-katello-4-0/19582 indicates that dep is here to stay :)

ianballou (approved):
Tests are good, so ACK!

PR #10805, opened 2023-11-17
Fixes #36850 - Avoid duplicate database indexes

hao-yu:
Some downstream users have already created the indexes manually to relieve the issue. When upgrading Katello, the migration script will create duplicate indexes to the table with different index names or the migration will fail with PG duplicate index error due to index name conflict with the one migration script creates.

Thus, modify this migration script to check for existing indexes in the table and then drop them before creating the new indexes.

### DIFF
--- a/db/migrate/20231020153017_create_indexes_for_nvra_and_epoch.rb
+++ b/db/migrate/20231020153017_create_indexes_for_nvra_and_epoch.rb
@@ -1,6 +1,19 @@
 class CreateIndexesForNvraAndEpoch < ActiveRecord::Migration[6.1]
-  def change
+  def up
+    if index_exists?(:katello_installed_packages, :nvra)
+      remove_index :katello_installed_packages, :nvra
+    end
+
+    if index_exists?(:katello_installed_packages, [:nvra, :epoch])
+      remove_index :katello_installed_packages, [:nvra, :epoch]
+    end
+
     add_index :katello_installed_packages, :nvra
     add_index :katello_installed_packages, [:nvra, :epoch]
+  end
+
+  def down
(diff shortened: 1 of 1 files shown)


hao-yu:
@jeremylenz As discussed, I submitted the changes I proposed in  [PR 10773](https://github.com/Katello/katello/pull/10773),

lfu:
This is not necessary.
You can remove a migration using:
```
rake db:migrate:down VERSION=20230816154510
```
Replace the Migration ID with yours.

hao-yu:
@lfu we don't want to do a manual job here

lfu:
Could you try something like:
```
add_index(:suppliers, :name, if_not_exists: true)
```

lfu:
@hao-yu Can you tell us how you have created the index manually for customers?

hao-yu:
>  Can you tell us how you have created the index manually for customers?

@lfu 
```
su - postgres -c "psql foreman -c \"create index index_katello_installed_packages_nvra on katello_installed_packages(nvra);\""
su - postgres -c "psql foreman -c \"create index index_katello_installed_packages_nvra_epoch on katello_installed_packages(nvra, epoch);\""
```

hao-yu:
how is this review going?

hao-yu:
I am not sure how user friendly we are asking users to remove duplicate indexes manually.

Unfortunately, the proposed PR is sitting here for 4 months without proper or motivation to progress

It is no point and too late to fix this any more because https://github.com/Katello/katello/pull/10773 has been released in downstream Satellite 6.14. For affected users who already upgraded to 6.14 will have duplicate indexes.

PR #10811, opened 2023-11-24
Fixes #36946 - cache if tracer/host_tools is installed

sbernhard:
Access to hosts page "/hosts" with about 72 hosts.

```
Without cached data:
Completed 200 OK in 9369ms (Views: 8113.6ms | ActiveRecord: 640.9ms | Allocations: 5036689)
Completed 200 OK in 8512ms (Views: 7289.0ms | ActiveRecord: 669.4ms | Allocations: 5044464)
SQL Queries just for installed packages: 264

With cached data:
Completed 200 OK in 9833ms (Views: 8730.5ms | ActiveRecord: 628.5ms | Allocations: 4952378)
Completed 200 OK in 9131ms (Views: 8049.9ms | ActiveRecord: 637.1ms | Allocations: 4950737)
No query to installed packages
```

### DIFF
--- a/app/services/katello/host/profiles_uploader.rb
+++ b/app/services/katello/host/profiles_uploader.rb
@@ -6,15 +6,20 @@ module Katello
         @host = host
       end
 
+      # rubocop:disable Metrics/MethodLength
       def upload
+        if @host.nil?
+          Rails.logger.warn("Host was not specified; skipping")
+          return false
+        elsif @host.content_facet.nil? || @host.content_facet.uuid.nil?
+          Rails.logger.warn("Host with ID %s has no content facet; skipping" % @host.id)
+          return false
+        end
+
         profiles = JSON.parse(@profile_string)
         #free the huge string from the memory
         @profile_string = 'TRIMMED'.freeze
-        if @host.nil?
-          Rails.logger.warn("Host was not specified; continuing")
-        elsif @host.content_facet.nil? || @host.content_facet.uuid.nil?
-          Rails.logger.warn("Host with ID %s has no content facet; continuing" % @host.id)
-        elsif profiles.try(:has_key?, "deb_package_profile")
+        if profiles.try(:has_key?, "deb_package_profile")
           # remove this when deb_package_profile API is removed
           payload = profiles.dig("deb_package_profile", "deb_packages") || []
           import_deb_package_profile(payload)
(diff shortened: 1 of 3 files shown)


sbernhard:
I don't think the failed checks are related to this change.

app/services/katello/host/profiles_uploader.rb
```diff
@@ -6,15 +6,20 @@ def initialize(profile_string:, host: nil)
...
         profiles = JSON.parse(@profile_string)
         #free the huge string from the memory
         @profile_string = 'TRIMMED'.freeze
         if @host.nil?
           Rails.logger.warn("Host was not specified; continuing")
+          return
         elsif @host.content_facet.nil? || @host.content_facet.uuid.nil?
           Rails.logger.warn("Host with ID %s has no content facet; continuing" % @host.id)
-        elsif profiles.try(:has_key?, "deb_package_profile")
+          return
```
m-bucher, line 19: Two things:
1. since we do not do anything with it, should the message not end with `skipping`?
2. given that we return `true` at the end, should we explicitly return `false` here? Although it is not evaluated anywhere, yet

jeremylenz:
From the benchmarks you posted it looks like getting the cached values is actually a bit slower compared to SQL queries, is that right? If so how is it better?

sbernhard:
> From the benchmarks you posted it looks like getting the cached values is actually a bit slower compared to SQL queries, is that right? If so how is it better?

I'm not sure if the measured execution times are really countable because there are so many other factors.

jeremylenz:
> I'm not sure if the measured execution times are really countable because there are so many other factors.

Have you seen other improvements from this that are harder to measure? Can you help me understand how this change makes sense?

sbernhard:
> > I'm not sure if the measured execution times are really countable because there are so many other factors.
> 
> Have you seen other improvements from this that are harder to measure? Can you help me understand how this change makes sense?

Lets try to benchmark it:

Old way:

```
irb(main):003:0> total=0; 100.times do x=Benchmark.measure { Host.all.each do |h| h.content_facet.nil? || h.content_facet.host_tools_installed? end }; total = total + x.total; end
=> 100
irb(main):004:0> puts total
30.919935000000027

irb(main):005:0> total=0; 100.times do x=Benchmark.measure { Host.all.each do |h| h.content_facet.nil? || h.content_facet.tracer_installed? end }; total = total + x.total; end
=> 100
irb(main):006:0> puts total
41.89736400000003
```

New way:
```
irb(main):001:0> total=0; 100.times do x=Benchmark.measure { Host.all.each do |h| h.content_facet.nil? || h.content_facet.host_tools_installed_new? end }; total = total + x.total; end
=> 100
irb(main):002:0> puts total
13.782863000000017

irb(main):007:0> total=0; 100.times do x=Benchmark.measure { Host.all.each do |h| h.content_facet.nil? || h.content_facet.tracer_installed_new? end }; total = total + x.total; end
=> 100
irb(main):008:0> puts total
12.510359000000003
```

jeremylenz (approved):
OK, seems harmless enough. Let's get it in.

Ack based on your testing results in https://github.com/Katello/katello/pull/10811#issuecomment-1867470853

PR #10814, opened 2023-11-28
Fixes #11938 - Drop WebMock.allow_net_connect!

ekohl:
Reportedly this was only done to allow ElasticSearch tests to pass, but that's been gone for a long time.


chris1984:
[test katello]

chris1984 (approved):

PR #10816, opened 2023-11-29
Fixes #37005 - Allow to set the the default name of template change…

Et7f3:
… content source

Add a setting
use that setting

I have added js part (because I grepped and identified the same constant their) but it might not be needed.

I haven't considered how migration should add this setting (If I read correctly it is pure runtime so no migration and the feature will be upserted)

I did test in foreman-rake console (by copying the lines)
What is remaining:
- On fresh instance with remote execution feature enabled:
Their is a new setting `Katello: Configure host for new content source` that allow to change the default template used when changing content source
- On old instance updated: the option should appear

### DIFF
--- a/app/controllers/katello/concerns/hosts_controller_extensions.rb
+++ b/app/controllers/katello/concerns/hosts_controller_extensions.rb
@@ -95,7 +95,7 @@ module Katello
                                       .distinct
 
           if Katello.with_remote_execution?
-            template_id = JobTemplate.find_by(name: 'Configure host for new content source')&.id
+            template_id = RemoteExecutionFeature.feature!(:katello_change_content_source).job_template_id
             job_invocation_path = new_job_invocation_path(template_id: template_id, host_ids: content_hosts.map { |h| h[:id] }) if template_id
           end
 
--- a/lib/katello/plugin.rb
+++ b/lib/katello/plugin.rb
@@ -685,6 +685,7 @@ Foreman::Plugin.register :katello do
     RemoteExecutionFeature.register(:katello_errata_install_by_search, N_("Katello: Install errata by search query"), :description => N_("Install errata using scoped search query"), :provided_inputs => ['Errata search query'])
     RemoteExecutionFeature.register(:katello_service_restart, N_("Katello: Service Restart"), :description => N_("Restart Services via Katello interface"), :provided_inputs => ['helpers'])
     RemoteExecutionFeature.register(:katello_host_tracer_resolve, N_("Katello: Resolve Traces"), :description => N_("Resolve traces via Katello interface"), :provided_inputs => ['ids'])
+    RemoteExecutionFeature.register(:katello_change_content_source, N_("Katello: Configure host for new content source"), :description => N_("Replace content source on the target machine"), :provided_inputs => [])
     RemoteExecutionFeature.register(:katello_module_stream_action, N_("Katello: Module Stream Actions"),
                                     :description => N_("Perform a module stream action via Katello interface"),
                                     :provided_inputs => ['action', 'module_spec', 'options'])


Et7f3:
Oh no I created in the [RedHat bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=2252121) should I add it in the Redmine also ?

adamruzicka:
Hi, thank you for sending this in. I haven't tested this yet but at first glance it looks like a step in the right direction

> should I add it in the Redmine also ?

Please do

jeremylenz (commented):
Thanks for the contribution @Et7f3!

image

I'm able to see the new feature here, but it doesn't seem to have a default template. 

It looks like [the job template](https://github.com/Katello/katello/blob/df80f7bc568d90d81a8ed4397630b457897b933e/app/views/foreman/job_templates/configure_host_for_new_content_source.erb) already has

```
feature: katello_change_content_source
```

but you called your feature `katello_host_change_content_source`. I think if you change it to match the job template, the default template should show up correctly.

Et7f3:
> but you called your feature `katello_host_change_content_source`. I think if you change it to match the job template, the default template should show up correctly.

I don't remember if I had choosed `katello_change_content_source` at first and renamed to be more descriptive or it it was directly that name. Or used that name without seeing this name used.

> I think if you change it to match the job template, the default template should show up correctly.

IIRC it was also empty for me but the drop down showed the basic job template and one custom for us: don't you see it in the drop-down ? I thought it was ok and the value could be changed in the settings or by the installer/migration script.

jeremylenz:
I do see it in the drop-down. But in order to be the default option (without having to click the dropdown) I believe the name will need to match the template.

jeremylenz:
> See [Install Errata - Katello Script Default](https://github.com/Katello/katello/blob/a4b8414151959d5af79c6904455647011025ead9/app/views/foreman/job_templates/install_errata.erb#L3) or [Install Errata - Katello Ansible Default](https://github.com/Katello/katello/blob/a4b8414151959d5af79c6904455647011025ead9/app/views/foreman/job_templates/install_errata_-_katello_ansible_default.erb#L2) it doesn't use the same name:

I'm not talking about the `name:` field; rather, I'm talking about the `feature:` field. The `feature:` field in the job template should match the `RemoteExecutionFeature.register` in this PR.

jeremylenz:
[test katello]

webpack/components/extensions/HostDetails/Tabs/RemoteExecutionConstants.js
```diff
@@ -9,5 +9,6 @@ export const REX_FEATURES = {
   KATELLO_HOST_TRACER_RESOLVE: 'katello_host_tracer_resolve',
   KATELLO_HOST_ERRATA_INSTALL_BY_SEARCH: 'katello_errata_install_by_search',
   KATELLO_HOST_MODULE_STREAM_ACTION: 'katello_module_stream_action',
+  KATELLO_HOST_CHANGE_CONTENT_SOURCE: 'katello_change_content_source',
```
jeremylenz, line 12: This will be unused in the JavaScript so it's not needed.

jeremylenz (commented):
Looking good, just one more comment!

webpack/components/extensions/HostDetails/Tabs/RemoteExecutionConstants.js
Et7f3: Thanks for confirmation (I just did when I compared how other features were implemented). Just to understand the others constants are needed because the ui provides run action X and customise REX. Since the ui for change content source already open the REX page for customisation: it won't be needed.

I will remove it.

jeremylenz: Yeah, we use the other constants in JavaScript when we're composing our API requests to call the REX API directly from the frontend (`/job_invocations/new`), for example [here in customizedRexUrlHelpers](https://github.com/Katello/katello/blob/0250c25c0ca44fff077983aea3541e989bd047c0/webpack/components/extensions/HostDetails/Tabs/customizedRexUrlHelpers.js#L52-L53). Whereas for the change content source feature [here](https://github.com/Katello/katello/blob/0250c25c0ca44fff077983aea3541e989bd047c0/webpack/scenes/Hosts/ChangeContentSource/actions.js#L24-L25), we're actually calling a different API endpoint `/api/v2/hosts/bulk/change_content_source`, and that endpoint doesn't require specifying the REX feature name.

jeremylenz:
[test katello]

jeremylenz (approved):
Works well. I was able to clone the default template, and change the default for the feature to my new template. The template was rendered and executed correctly.

Thanks for your contribution @Et7f3 !

ACK :+1:

PR #10817, opened 2023-11-30
Fixes #36960 - Assign default content view environment during migration when no existing CVE found

jeremylenz:

If a content facet has an assigned lifecycle environment which does not have any content views promoted to it, there will be no associated ContentViewEnvironment objects in the database. This breaks a migration and thus breaks upgrades.

With this change, such hosts are assigned to the default content view environment during migration.

This condition may appear if you've assigned a host/content facet to a lifecycle environment, but haven't promoted any content views to that environment.

It seems it may also be triggered if at some point in the past you've deleted a content view or lifecycle environment, and the deletion task failed.

1. If you don't already have a host assigned to a content view, create a content view and assign a host to it
2. Create a lifecycle environment, but do NOT promote the content view to it. This will ensure that the LCE does not have an associated ContentViewEnvironment.
3. Reverse the migration:

```
bundle exec rails db:migrate:down VERSION=20220929204746
```

4. Find your host and change its lifecycle environment to the one you created. We need to skip validations because of the reverted migration, so you'll have to do it in rails console:

```
rhel8 = Host::Managed.find_by(name: 'rhel8.fedora.example.com')
[2] pry(main)> cf = rhel8.content_facet
=> #<Katello::Host::ContentFacet:0x00005599477823b0
 id: 1,
 host_id: 2,
...
 content_view_id: 1,
 lifecycle_environment_id: 1>
[3] pry(main)> cf.update_columns(lifecycle_environment_id: <your new LCE id>)
```

5. Checkout master, if you like, and run the up migration:

```
bundle exec rails db:migrate:up VERSION=20220929204746
```

You'll get an error

```
StandardError: An error has occurred, this and all later migrations canceled:

undefined method `id' for nil:NilClass
```

Now, check out the PR and run the migration again. You should see something like this, with no errors:

```
== 20220929204746 AddContentViewEnvironmentContentFacet: migrating ============
-- create_table(:katello_content_view_environment_content_facets)
   -> 0.0070s
2023-12-06T22:11:24 [I|app|] Found 1 hosts whose lifecycle environment does not have a corresponding ContentViewEnvironment
2023-12-06T22:11:24 [I|app|] Creating CVE for host 2 with content_view_id 2 and lifecycle_environment_id 3
2023-12-06T22:11:24 [I|app|] Found 0 hosts whose content facet is missing either content_view_id or lifecycle_environment_id
-- remove_column(:katello_content_facets, :content_view_id)
   -> 0.0006s
-- remove_column(:katello_content_facets, :lifecycle_environment_id)
   -> 0.0004s
== 20220929204746 AddContentViewEnvironmentContentFacet: migrated (0.1426s) ===
```

### DIFF
--- a/app/lib/katello/util/cvecf_migrator.rb
+++ b/app/lib/katello/util/cvecf_migrator.rb
@@ -1,0 +1,34 @@
+module Katello
+  module Util
+    class CVECFMigrator # used in db/migrate/20220929204746_add_content_view_environment_content_facet.rb
+      def execute!
+        hosts_with_no_cve = []
+        hosts_with_missing_cve = []
+
+        ::Host::Managed.all.each do |host|
+          next if host.content_facet.blank?
+          if ::Katello::ContentView.exists?(id: host.content_facet.content_view_id) && ::Katello::KTEnvironment.exists?(host.content_facet.lifecycle_environment_id)
+            cve = ::Katello::ContentViewEnvironment.find_by(content_view_id: host.content_facet.content_view_id, environment_id: host.content_facet.lifecycle_environment_id)
+            if cve.blank?
+              hosts_with_no_cve << host
+            end
+          else
+            hosts_with_missing_cve << host
+          end
+        end
+
+        if hosts_with_missing_cve.present? || hosts_with_no_cve.present?
+          Rails.logger.warn "Found #{hosts_with_no_cve.count} hosts whose lifecycle environment does not have a corresponding ContentViewEnvironment"
+          Rails.logger.warn "Found #{hosts_with_missing_cve.count} hosts whose content facet is missing either content_view_id or lifecycle_environment_id"
+          Rails.logger.info "You may want to change the content view / lifecycle environment for these hosts manually."
+        end
+        (hosts_with_no_cve + hosts_with_missing_cve).each do |host|
+          default_content_view = host.organization.default_content_view
+          library = host.organization.library
+          Rails.logger.info "Updating host #{host.name} with default content_view_id and lifecycle_environment_id"
+          host.content_facet&.update_columns(content_view_id: default_content_view&.id, lifecycle_environment_id: library&.id)
+        end
+      end
+    end
+  end
+end
--- a/db/migrate/20220929204746_add_content_view_environment_content_facet.rb
+++ b/db/migrate/20220929204746_add_content_view_environment_content_facet.rb
@@ -8,12 +8,13 @@ class AddContentViewEnvironmentContentFacet < ActiveRecord::Migration[6.1]
       t.references :content_view_environment, :null => false, :index => false, :foreign_key => { :to_table => 'katello_content_view_environments' }
       t.references :content_facet, :null => false, :index => false, :foreign_key => { :to_table => 'katello_content_facets' }
     end
+    ::Katello::Util::CVECFMigrator.new.execute!
     FakeContentFacet.all.each do |content_facet|
       cve_id = ::Katello::KTEnvironment.find(content_facet.lifecycle_environment_id)
           .content_view_environments
           .find_by(content_view_id: content_facet.content_view_id)
-          .id
-      unless ::Katello::ContentViewEnvironmentContentFacet.create(
+          &.id
+      unless cve_id.present? && ::Katello::ContentViewEnvironmentContentFacet.create(
         content_facet_id: content_facet.id,
         content_view_environment_id: cve_id
       )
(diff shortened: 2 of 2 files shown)


app/lib/katello/util/cvecf_migrator.rb
```diff
@@ -0,0 +1,33 @@
...
+        Rails.logger.info "Found #{hosts_with_no_cve.count} hosts whose lifecycle environment does not have a corresponding ContentViewEnvironment"
+        hosts_with_no_cve.each do |host|
+          Rails.logger.info "Creating CVE for host #{host.id} with content_view_id #{host.content_facet.content_view_id} and lifecycle_environment_id #{host.content_facet.lifecycle_environment_id}"
+          ::Katello::ContentViewEnvironment.create(environment_id: host.content_facet.lifecycle_environment_id, content_view_id: host.content_facet.content_view_id)
+        end
+
+        Rails.logger.info "Found #{hosts_with_missing_cve.count} hosts whose content facet is missing either content_view_id or lifecycle_environment_id"
+        hosts_with_missing_cve.each do |host|
+          Rails.logger.info "Updating host #{host.id} with default content_view_id and lifecycle_environment_id"
+          host.content_facet&.update(content_view_id: ::Katello::ContentView.default.first.id, lifecycle_environment_id: ::Katello::KTEnvironment.library.first.id)
```
parthaa, line 28: You need to make sure the CV and Environment belong to the same org as the content_facet

app/lib/katello/util/cvecf_migrator.rb
```diff
@@ -0,0 +1,33 @@
+module Katello
+  module Util
+    class CVECFMigrator # used in db/migrate/20220929204746_add_content_view_environment_content_facet.rb
+      def execute!
+        hosts_with_no_cve = []
+        hosts_with_missing_cve = []
+
+        ::Host::Managed.all.each do |host|
```
parthaa, line 8:
```suggestion
        ::Host::Managed.all.each do |host|
                 next unless host.content_facet.present?
```

db/migrate/20220929204746_add_content_view_environment_content_facet.rb
```diff
@@ -37,16 +38,47 @@ def down
 
     ::Katello::Host::ContentFacet.reset_column_information
 
+    default_cv_id = ::Katello::ContentView.default.first.id
```
parthaa, line 41: Are these changes necessary.. I ;d think down should just fail with not implemented error.?

jeremylenz: Down is needed for devs who don't want to torture themselves :) it doesn't make much difference for users or customers

I did have to make these changes so that my down migration would work.

jeremylenz:
@parthaa updated

jeremylenz:
fixed Rubocop

app/lib/katello/util/cvecf_migrator.rb
```diff
@@ -0,0 +1,42 @@
...
+            end
+          else
+            hosts_with_missing_cve << host
+          end
+        end
+
+        Rails.logger.info "Found #{hosts_with_no_cve.count} hosts whose lifecycle environment does not have a corresponding ContentViewEnvironment"
+        hosts_with_no_cve.each do |host|
+          Rails.logger.info "Creating CVE for host #{host.id} with content_view_id #{host.content_facet.content_view_id} and lifecycle_environment_id #{host.content_facet.lifecycle_environment_id}"
+          ::Katello::ContentViewEnvironment.create(environment_id: host.content_facet.lifecycle_environment_id, content_view_id: host.content_facet.content_view_id)
```
parthaa, line 23: The assumption here that a CVE should have existed is dangerous imo. I'd rather you assign these the default content view and environment than create one (i.e same as missing cve).

parthaa:
I tried previous versions of the katello prior to your cve work and found it literally impossible to have a content_facet with a lifecycle environment but No content view. I had to use the rails console as pointed out literally to create a possible dirty data. These content_facets are bad data. I would not want to create a content_view_environment if one did not exist before. I 'd rather assign the default content view and environment.
suggested patch
```patch
diff --git a/app/lib/katello/util/cvecf_migrator.rb b/app/lib/katello/util/cvecf_migrator.rb
index b8d2c7937a..f5dbaf5ba0 100644
--- a/app/lib/katello/util/cvecf_migrator.rb
diff --git a/app/lib/katello/util/cvecf_migrator.rb b/app/lib/katello/util/cvecf_migrator.rb
index b8d2c7937a..f5dbaf5ba0 100644
--- a/app/lib/katello/util/cvecf_migrator.rb
+++ b/app/lib/katello/util/cvecf_migrator.rb
@@ -16,27 +16,20 @@ module Katello
             hosts_with_missing_cve << host
           end
         end
-
+        default_content_view = host.organization.default_content_view
+        library =  host.organization.library
         Rails.logger.info "Found #{hosts_with_no_cve.count} hosts whose lifecycle environment does not have a corresponding ContentViewEnvironment"
         hosts_with_no_cve.each do |host|
-          Rails.logger.info "Creating CVE for host #{host.id} with content_view_id #{host.content_facet.content_view_id} and lifecycle_environment_id #{host.content_facet.lifecycle_environment_id}"
-          ::Katello::ContentViewEnvironment.create(environment_id: host.content_facet.lifecycle_environment_id, content_view_id: host.content_facet.content_view_id)
+          Rails.logger.info "Updating host #{host.id} with default content_view_id and lifecycle_environment_id"
+          host.content_facet.update(content_view_id: default_content_view.id, lifecycle_environment_id: library.id )
         end
 
         Rails.logger.info "Found #{hosts_with_missing_cve.count} hosts whose content facet is missing either content_view_id or lifecycle_environment_id"
         hosts_with_missing_cve.each do |host|
           Rails.logger.info "Updating host #{host.id} with default content_view_id and lifecycle_environment_id"
-          host.content_facet&.update(content_view_id: default_cv_for_host(host)&.id, lifecycle_environment_id: default_lce_for_host(host)&.id)
+          host.content_facet.update(content_view_id: default_content_view.id, lifecycle_environment_id: library.id )
         end
       end
-
-      def default_cv_for_host(host)
-        ::Katello::ContentView.default.where(organization_id: host.organization_id).first
-      end
-
-      def default_lce_for_host(host)
-        ::Katello::KTEnvironment.library.where(organization_id: host.organization_id).first
-      end
     end
   end
 end
```

jeremylenz:
> I would not want to create a content_view_environment if one did not exist before. I 'd rather assign the default content view and environment.

The way multi CV works is that content facets are assigned to multiple content view environments, instead of having content_view_id and lifecycle_environment_id fields. So assigning a host to the default CV/LCE is not correct. We need the host to be assigned to the same content view and environment as before the migration.

parthaa:
> ContentViewEnvironment: purpose before migration
> 
> * Keep track of what content view has been promoted to a given LCE, and thus what content view version to use.
> 
> ContentViewEnvironment purpose after migration:
> 
> Same as above, and ALSO
> 
> * allow hosts to be assigned to multiple CVEs.

CVE table has existed long before CVECF.  There are pulp and candlepin entities that were associated to this CVE  which may have been removed if delete failed. 

> ContentViewEnvironment: purpose before migration
> 
> * Keep track of what content view has been promoted to a given LCE, and thus what content view version to use.

Yes the purpose in terms of model. But if there is a CVE I would want to make sure repos belonging to the CVE are represented in pulp and candlepin. Using the default org view guarantees that they exist and so dont have to worry about tallying.

jeremylenz:
Ok I see your point. But it will cause hosts to magically change themselves to the default CVE which isn't a great user experience. I have to think if there's a reasonable way to make it better..

parthaa:
Look here for example
https://github.com/Katello/katello/blob/master/app/lib/actions/katello/content_view_environment/destroy.rb#L33

` cv_env.destroy!` is happening in the finalize step after everything else has been accomplished. If the customer skipped `Candlepin::Environment::Destroy` but kept `plan_action(Repository::Destroy, repo, skip_environment_update: true, docker_cleanup: false)` and finalize step OR if repo destroy failed due to pulp - we can't really remediate this situation. To me its safer to assume that CVE should not have existed in the first place and assign it to default CVE

jeremylenz:
@parthaa updated :+1:

app/lib/katello/util/cvecf_migrator.rb
```diff
@@ -0,0 +1,34 @@
...
+        if hosts_with_missing_cve.present? || hosts_with_no_cve.present?
+          Rails.logger.warn "Found #{hosts_with_no_cve.count} hosts whose lifecycle environment does not have a corresponding ContentViewEnvironment"
+          Rails.logger.warn "Found #{hosts_with_missing_cve.count} hosts whose content facet is missing either content_view_id or lifecycle_environment_id"
+          Rails.logger.info "You may want to change the content view / lifecycle environment for these hosts manually."
+        end
+        (hosts_with_no_cve + hosts_with_missing_cve).each do |host|
+          default_content_view = host.organization.default_content_view
+          library = host.organization.library
+          Rails.logger.info "Updating host #{host.name} with default content_view_id and lifecycle_environment_id"
+          host.content_facet&.update(content_view_id: default_content_view&.id, lifecycle_environment_id: library&.id)
```
parthaa, line 29:
```suggestion
          host.content_facet&.update_columns(content_view_id: default_content_view&.id, lifecycle_environment_id: library&.id)
```

parthaa (approved):
Works well. APJ

sjha4:
[test katello]

jeremylenz:
[test katello]

PR #10820, opened 2023-12-06
Fixes #11760 - Update repo deletion message to include --remove-from-content-view-versions help text

sjha4:
Update the repo deletion error message to include --remove-from-content-view-versions flag suggestion.
Create a CV and add a repo and publish.
From hammer, try hammer repository delete --id=$REPO_ID
The error should now include mention of the flag to delete the repo from all cv versions.
Use the --remove-from-content-view-versions true option in the hammer repo delete command and the deletion should succeed.


chris1984 (approved):
ACK, works great

PR #10822, opened 2023-12-08
Fixes #36975 - Pass cp_products as array to ProductContentImporter instead of hash

sjha4:
~~We have a second code path that initializes ProductContentImporter. This second code path was passing an array to the initializer vs the first path of import_all which passes a hash.~~

UPDATE: So instead of converting the 2nd code path to pass a hash, we are revrting to using array as input for ProductContentImporter and correcting the import_all flow to array instead of Hash.

1. Import an expired manifest into an org.
2. Delete the manifest
3. Import a newer, valid manifest.

Refer to steps in https://bugzilla.redhat.com/show_bug.cgi?id=2253621 for more details.

### DIFF
--- a/app/services/katello/product_content_importer.rb
+++ b/app/services/katello/product_content_importer.rb
@@ -21,46 +21,6 @@ module Katello
     #    "enabled":false
     # }
     #
-    #Example cp_products json structure (key:values where values are arrays and productContent is an array of contents inside value array)
-    # {
-    #    "230"=>
-    #   [{"id"=>"230",
-    #     "productContent"=>
-    #      [{"content"=>
-    #         {"created"=>"2023-11-01T18:02:54+0000",
-    #          "updated"=>"2023-11-01T18:02:54+0000",
-    #          "uuid"=>"ff8080818b715841018b8c0d76d90190",
-    #          "id"=>"3996",
-    #          "type"=>"yum",
-    #          "label"=>"rhel-7-server-rt-htb-rpms",
-    #          "name"=>"Red Hat Enterprise Linux for Real Time HTB (RHEL 7 Server) (RPMs)",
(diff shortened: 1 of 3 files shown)


jeremylenz:
I want to make sure that if something in the data shape changed at some point, this handles both cases.

app/services/katello/product_content_importer.rb
```diff
@@ -76,13 +36,12 @@ def add_product_content(product, product_content_json)
     end
 
     def find_product_for_content(content_id)
-      #Hash.find converts hash into enumerable iterating over each key,value pair which we can capture based on find result
-      cp_product_id, __cp_product_value = @cp_products.find do |prod_json|
-        prod_json&.last&.first&.[]('productContent')&.any? do |product_content_json|
+      prod = @cp_products.find do |prod_json|
```
parthaa, line 39:
```suggestion
      prod = @cp_products.find do |prod_json|
      next unless prod_json&. []('productContent').present?
```

app/services/katello/product_content_importer.rb
```diff
@@ -76,13 +36,12 @@ def add_product_content(product, product_content_json)
...
+        prod_json['productContent'].any? do |product_content_json|
           product_content_json["content"]["id"] == content_id
```
parthaa, line 41:
```suggestion
          product_content_json&.dig("content", "id") == content_id
```

app/services/katello/product_content_importer.rb
parthaa: May be even 
`prod_json&.[]('productContent')&.any? do |product_content_json|` instead of `prod_json['productContent'].any? do |product_content_json|`

parthaa (approved):
Works well APJ

PR #10824, opened 2023-12-12
Fixes #36979 - Change cdn_ssl_version setting to cdn_min_tls_version

wbclark:
This uses the min_version attribute from Net::HTTP instead of ssl_version, which in turn uses OpenSSL::SSL::SSLContext#min_version and not the older OpenSSL::SSL::SSLContent#ssl_version. As a result the there is no longer a blanket SSLv23 setting and the data format for specifying TLS versions has changed.

It also sets a default value of the new setting for TLSv1.2, which is reasonable because older versions of the standard are deprecated. Users requiring a deprecated protocol version due to their network infrastructure can lower this value as needed.

1. What is a reasonable default value? I went with TLSv1.2 since TLSv1.1 and older standards are officially deprecated
2. How to best handle the difference in allowed values for `min_version` and `ssl_version` in OpenSSL, since `ssl_version` is deprecated? I opted to introduce a new setting entirely since the functionality has changed and we can't directly map the old allowed values onto the new allowed values.
3. I removed the related tests from `test/models/setting_test.rb` as it seems that `test/lib/resources/cdn_test.rb` provides adequate coverage.

1. Test uploading a manifest, syncing from CDN, etc
4. Test inter-server sync from a server that has newer TLS disabled... this should fail because older TLS is now blocked by default
5. Lower the `CDN min TLS version` value and re-try inter-server sync. Now it should succeed once older TLS has been permitted.

#### Open Questions

1. Is there any need for continued support for insecure protocols such as SSLv3 ?
2. Is a DB migration needed to clean up the removed setting? I tested giving a non-nil value to the older setting before applying the change, it doesn't seem to cause any issue, and it looks like there is precedent for skipping this, as in https://github.com/Katello/katello/commit/283c63dc301b52f08ff6e09d45f0e817e811f9a0

### DIFF
--- a/app/lib/katello/resources/cdn.rb
+++ b/app/lib/katello/resources/cdn.rb
@@ -1,7 +1,18 @@
 module Katello
   module Resources
     module CDN
-      SUPPORTED_SSL_VERSIONS = ['SSLv23', 'TLSv1'].freeze
+      # the if defined? check is included for TLSv1.3 because it is not available on all platforms.
+      # at the time of writing, TLSv1 through TLSv1.2 should be available on all supported platforms,
+      # but we check regardless in case they are removed from any platform in the future.
+      # any unavailable protocols will be removed from the final Hash by the compact method.
+      SUPPORTED_TLS_VERSIONS_MAPPING = {
+        'TLSv1' => (OpenSSL::SSL::TLS1_VERSION if defined?(OpenSSL::SSL::TLS1_VERSION)),
+        'TLSv1.1' => (OpenSSL::SSL::TLS1_1_VERSION if defined?(OpenSSL::SSL::TLS1_1_VERSION)),
+        'TLSv1.2' => (OpenSSL::SSL::TLS1_2_VERSION if defined?(OpenSSL::SSL::TLS1_2_VERSION)),
+        'TLSv1.3' => (OpenSSL::SSL::TLS1_3_VERSION if defined?(OpenSSL::SSL::TLS1_3_VERSION))
+      }.compact.freeze
+
+      SUPPORTED_TLS_VERSIONS = SUPPORTED_TLS_VERSIONS_MAPPING.keys.freeze
 
       class Utils
         # takes releasever from contentUrl (e.g. 6Server, 6.0, 6.1)
@@ -24,9 +35,9 @@ module Katello
         end
 
         def initialize(url, options = {})
-          @ssl_version = Setting[:cdn_ssl_version]
-          if @ssl_version && !SUPPORTED_SSL_VERSIONS.include?(@ssl_version)
-            fail("Invalid SSL version specified. Check the 'CDN SSL Version' setting")
+          @cdn_min_tls_version = Setting[:cdn_min_tls_version]
+          if @cdn_min_tls_version && !SUPPORTED_TLS_VERSIONS.include?(@cdn_min_tls_version)
+            fail("Invalid TLS min_version specified. Check the 'CDN Minimum Allowed TLS Version' setting")
           end
 
           options.reverse_merge!(:verify_ssl => 9)
@@ -106,12 +117,10 @@ module Katello
             net.cert_store = @cert_store
           end
 
-          # NOTE: This was added because some proxies dont support SSLv23 and do not handle TLS 1.2
-          # Valid values in ruby 1.9.3 are 'SSLv23' or 'TLSV1'
-          # Run the following command in rails console to figure out other
-          # valid constants in other ruby versions
-          # "OpenSSL::SSL::SSLContext::METHODS"
-          net.ssl_version = @ssl_version
+          # NOTE: This was originally added because some proxies dont support SSLv23 and do not handle TLS 1.2
+          # Run the following command in rails console to figure out valid constants on your platform:
+          # "OpenSSL::SSL.constants.select { |sym| sym.to_s.include?('VERSION') }"
+          net.min_version = SUPPORTED_TLS_VERSIONS_MAPPING[@cdn_min_tls_version]
 
           if (@options[:verify_ssl] == false) || (@options[:verify_ssl] == OpenSSL::SSL::VERIFY_NONE)
             net.verify_mode = OpenSSL::SSL::VERIFY_NONE
--- a/lib/katello/plugin.rb
+++ b/lib/katello/plugin.rb
@@ -377,12 +377,12 @@ Foreman::Plugin.register :katello do
         collection: proc { http_proxy_select },
         include_blank: N_("no global default")
 
-      setting 'cdn_ssl_version',
+      setting 'cdn_min_tls_version',
         type: :string,
-        default: nil,
-        full_name: N_('CDN SSL version'),
-        description: N_("SSL version used to communicate with the CDN"),
-        collection: proc { hashify_parameters(Katello::Resources::CDN::SUPPORTED_SSL_VERSIONS) }
+        default: Katello::Resources::CDN::SUPPORTED_TLS_VERSIONS.include?('TLSv1.3') ? 'TLSv1.3' : 'TLSv1.2',
+        full_name: N_('CDN minimum allowed TLS version'),
+        description: N_("Minimum allowed TLS version used to communicate with the CDN. WARNING: Older versions of the TLS standard than TLSv1.2 are deprecated and should only be enabled when required for backwards compatibility with HTTP proxy servers."),
+        collection: proc { hashify_parameters(Katello::Resources::CDN::SUPPORTED_TLS_VERSIONS) }
 
       setting 'katello_default_provision',
         type: :string,
--- a/test/lib/resources/cdn_test.rb
+++ b/test/lib/resources/cdn_test.rb
@@ -1,20 +1,41 @@
 require 'katello_test_helper'
 class CdnResourceTest < ActiveSupport::TestCase
-  def test_http_downloader_v2
-    Setting[:cdn_ssl_version] = 'SSLv23'
+  def test_http_downloader_tlsv1
+    Setting[:cdn_min_tls_version] = 'TLSv1'
     cdn_resource = Katello::Resources::CDN::CdnResource.new('http://foo.com')
-    assert_equal cdn_resource.http_downloader.ssl_version, 'SSLv23'
+    assert_equal cdn_resource.http_downloader.min_version, OpenSSL::SSL::TLS1_VERSION
   end
 
-  def test_http_downloader_tls
-    Setting[:cdn_ssl_version] = 'TLSv1'
+  def test_http_downloader_tlsv11
+    Setting[:cdn_min_tls_version] = 'TLSv1.1'
     cdn_resource = Katello::Resources::CDN::CdnResource.new('http://foo.com')
-    assert_equal cdn_resource.http_downloader.ssl_version, 'TLSv1'
+    assert_equal cdn_resource.http_downloader.min_version, OpenSSL::SSL::TLS1_1_VERSION
   end
 
-  def test_http_downloader_no_version
+  def test_http_downloader_tlsv12
+    Setting[:cdn_min_tls_version] = 'TLSv1.2'
     cdn_resource = Katello::Resources::CDN::CdnResource.new('http://foo.com')
-    assert_nil cdn_resource.http_downloader.ssl_version
+    assert_equal cdn_resource.http_downloader.min_version, OpenSSL::SSL::TLS1_2_VERSION
+  end
+
+  def test_http_downloader_tlsv13
+    # TLSv1.3 is unavailable in some environments, for example the EL7 builders at
+    # ci.theforeman.org. This check should be removable when those are upgraded to
+    # EL8 or later. See https://github.com/theforeman/foreman-infra/issues/1706
+    unless defined?(OpenSSL::SSL::TLS1_3_VERSION)
+      skip "TLSv1.3 is unavailable in the current environment"
+    end
+
+    Setting[:cdn_min_tls_version] = 'TLSv1.3'
+    cdn_resource = Katello::Resources::CDN::CdnResource.new('http://foo.com')
+    assert_equal cdn_resource.http_downloader.min_version, OpenSSL::SSL::TLS1_3_VERSION
+  end
+
+  def test_http_downloader_default_version
+    cdn_resource = Katello::Resources::CDN::CdnResource.new('http://foo.com')
+
+    default_min_tls_version = defined?(OpenSSL::SSL::TLS1_3_VERSION) ? OpenSSL::SSL::TLS1_3_VERSION : OpenSSL::SSL::TLS1_2_VERSION
+    assert_equal cdn_resource.http_downloader.min_version, default_min_tls_version
   end
 
   def test_http_proxy_no_cacert
@@ -29,7 +50,7 @@ class CdnResourceTest < ActiveSupport::TestCase
   end
 
   def test_http_downloader_bad_param
-    Setting[:cdn_ssl_version] = 'Foo'
+    Setting[:cdn_min_tls_version] = 'Foo'
     assert_raise RuntimeError do
       Katello::Resources::CDN::CdnResource.new('http://foo.com')
     end
--- a/test/models/setting_test.rb
+++ b/test/models/setting_test.rb
@@ -17,15 +17,6 @@ module Katello
       assert setting.valid?
     end
 
-    def test_cdn_ssl_setting
-      # TODO: assert an error raised by the SettingRegistry
-      # setting = Foreman.settings.set_user_value('cdn_ssl_version', nil)
-      # assert setting.valid?
-
-      setting = Foreman.settings.set_user_value('cdn_ssl_version', 'SSLv23')
-      assert setting.valid?
-    end
-
     def test_recalculate_errata_status
       ForemanTasks.expects(:async_task).with(::Actions::Katello::Host::RecalculateErrataStatus)
       Setting['errata_status_installable'] = !Setting['errata_status_installable']


wbclark:
I disabled the TLSv1.3 unit test because CI infra doesn't know about TLSv1.3.

I set TLSv1.2 as the default minimum version (the previous feature had no default)

wbclark:
Requesting reviews from @ehelms for general protocol wizardry and from @hao-yu and @pmoravec for familiarity with the related downstream bug

app/lib/katello/resources/cdn.rb
```diff
@@ -1,7 +1,12 @@
 module Katello
   module Resources
     module CDN
-      SUPPORTED_SSL_VERSIONS = ['SSLv23', 'TLSv1'].freeze
+      SUPPORTED_TLS_VERSIONS = [
+        'TLSv1',
+        'TLSv1.1',
+        'TLSv1.2',
+        *('TLSv1.3' if defined?(OpenSSL::SSL::TLS_1_3_VERSION))
```
pmoravec, line 8: Shouldn't be there `OpenSSL::SSL::TLS1_3_VERSION` ? (`s/TLS_1/TLS1/`)?

wbclark: Yes! Good catch, thanks

pmoravec:
Dunno about impact on upgrades (i.e. does db migration make sense?), and apart of the typo it sounds great!

test/lib/resources/cdn_test.rb
```diff
@@ -1,20 +1,39 @@
...
     cdn_resource = Katello::Resources::CDN::CdnResource.new('http://foo.com')
-    assert_nil cdn_resource.http_downloader.ssl_version
+    assert_equal cdn_resource.http_downloader.min_version, 771
+  end
+
+  def test_http_downloader_tlsv13
+    # TLSv1.3 is unavailable in some environments, for example the EL7 builders at
+    # ci.theforeman.org. This check should be removable when those are upgraded to
+    # EL8 or later. See https://github.com/theforeman/foreman-infra/issues/1706
+    unless defined?(OpenSSL::SSL::TLS_1_3_VERSION)
```
wbclark, line 25: And the same change is needed here

wbclark:
> Dunno about impact on upgrades (i.e. does db migration make sense?), and apart of the typo it sounds great!

That's a good question.

I tested this by setting a value for the original setting then applying my patch and restarting the server, and found that it didn't cause any immediate issue.

I also found the earlier commit https://github.com/Katello/katello/commit/283c63dc301b52f08ff6e09d45f0e817e811f9a0 which removes a setting without any migration to remove the previous value from the DB.

Maybe @ehelms or @parthaa can say if there is any reason a migration should be included with this one.

wbclark:
I pushed an update which fixes the typos.

pmoravec (approved):

ehelms:
> I disabled the TLSv1.3 unit test because CI infra doesn't know about TLSv1.3.

What were the issues? Are there are other reasons not to defer to 1.3 as the default?

app/lib/katello/resources/cdn.rb
```diff
@@ -24,9 +29,9 @@ def substitutor
         end
 
         def initialize(url, options = {})
-          @ssl_version = Setting[:cdn_ssl_version]
-          if @ssl_version && !SUPPORTED_SSL_VERSIONS.include?(@ssl_version)
-            fail("Invalid SSL version specified. Check the 'CDN SSL Version' setting")
+          @cdn_min_tls_version = Setting[:cdn_min_tls_version]
+          if @cdn_min_tls_version && !SUPPORTED_TLS_VERSIONS.include?(@cdn_min_tls_version)
+            fail("Invalid TLS min_version specified. Check the 'CDN Minimum Allowed TLS Version' setting")
```
ehelms, line 34: How will users see this failure? Does it propagate to the UI and API?

app/lib/katello/resources/cdn.rb
```diff
@@ -107,11 +112,16 @@ def http_downloader
...
           # Run the following command in rails console to figure out other
           # valid constants in other ruby versions
-          # "OpenSSL::SSL::SSLContext::METHODS"
-          net.ssl_version = @ssl_version
+          # "OpenSSL::SSL.constants.select { |sym| sym.to_s.include?('VERSION') }"
+          net.min_version = OpenSSL::SSL.const_get(
+            @cdn_min_tls_version.gsub('TLSv', 'TLS')
+                                .gsub('.', '_')
+                                .concat('_VERSION')
+                                .to_sym
```
ehelms, line 123: I'd consider pulling this out to a method to represent translation of our representation of TLS versions to the Ruby required ones and then it can be tested directly.

ehelms: And I would consider swapping these so that the list is based on the Ruby naming convention with a translation to something human readable. And, consider just having a map of values rather than a method that can break when naming convention breaks.

app/lib/katello/resources/cdn.rb
ehelms: The other tricky part being that an organization admin will run into this issue and then have to contact a Satellite admin in order to modify this. I realize that's how the current functionality already works, but my point being we should make sure this is as actionable by a user encountering it as possible.

app/lib/katello/resources/cdn.rb
wbclark: Thanks for the suggestions. I updated it to use the map of values approach, which should be easy to maintain because the number of supported protocols is small.

I kept the human readable version strings as the keys and the Ruby naming convention as the values (we convert to the Ruby naming convention only once).

I also created a separate Array from the human readable version strings since those only are referenced in multiple files and it's easier than calling `SUPPORTED_TLS_VERSIONS_MAPPING.keys` each time.

wbclark:
> What were the issues? Are there are other reasons not to defer to 1.3 as the default?

There is some discussion on it here: https://github.com/theforeman/foreman-infra/pull/2001

The differences between TLSv1.2 and TLSv1.3 are small, and TLSv1.2 is still a supported standard, while as we learned with our own CI TLSv1.3 is not yet available everywhere. I opted for TLSv1.2 as the default minimum allowed version as a sane choice for most organizations, while those desiring to enforce TLSv1.3 only have the choice of raising that minimum allowed version via the option if their platform supports it.

app/lib/katello/resources/cdn.rb
```diff
@@ -1,7 +1,14 @@
 module Katello
   module Resources
     module CDN
-      SUPPORTED_SSL_VERSIONS = ['SSLv23', 'TLSv1'].freeze
+      SUPPORTED_TLS_VERSIONS_MAPPING = {
+        'TLSv1'   => OpenSSL::SSL::TLS1_VERSION,
```
wbclark, line 5: Our rubocop seems configured to not allow TableAlignment.

That's unfortunate, but not a big deal.

I'll make this line adequately ugly to satisfy rubocop so that the rest of the tests can run.

app/lib/katello/resources/cdn.rb
```diff
@@ -1,7 +1,14 @@
 module Katello
   module Resources
     module CDN
-      SUPPORTED_SSL_VERSIONS = ['SSLv23', 'TLSv1'].freeze
+      SUPPORTED_TLS_VERSIONS_MAPPING = {
+        'TLSv1' => OpenSSL::SSL::TLS1_VERSION,
```
ehelms, line 5: I wonder if we need the same if defined check for this older version in the event that a Ruby version removes it.

app/lib/katello/resources/cdn.rb
```diff
@@ -1,7 +1,14 @@
...
+        'TLSv1.1' => OpenSSL::SSL::TLS1_1_VERSION,
+        'TLSv1.2' => OpenSSL::SSL::TLS1_2_VERSION,
+        'TLSv1.3' => (OpenSSL::SSL::TLS1_3_VERSION if defined?(OpenSSL::SSL::TLS1_3_VERSION))
```
ehelms, line 8: Does this work? What value does it put into the hash?

wbclark: Yes, it works. If it's not defined, `defined?` returns `nil`, the "then" statement also evaluates to `nil`, and the corresponding KVP is removed by the `Hash#compact` method.

https://ruby-doc.org/2.7.8/Hash.html#method-i-compact
https://ruby-doc.org/2.7.8/syntax/miscellaneous_rdoc.html#label-defined-3F

The relevant behavior of modifier `if` wasn't documented at https://ruby-doc.org/2.7.8/syntax/control_expressions_rdoc.html , but `(expr if nil)` evaluates to `nil`:

```
irb(main):001:0> defined?(foo)
=> nil
irb(main):002:1* myhash = {
irb(main):003:1*   'one' => 'oneval',
irb(main):004:1*   'two' => ('twoval' if defined?(foo))
irb(main):005:0> }
=> {"one"=>"oneval", "two"=>nil}
irb(main):006:0> myhash.compact
=> {"one"=>"oneval"}
```

app/lib/katello/resources/cdn.rb
wbclark: I suppose there is no reason I couldn't include the relevant check for each protocol version, just to future proof everything.

wbclark: I pushed an update that demonstrates what that looks like.

app/lib/katello/resources/cdn.rb
wbclark: `fail()` should cause an ISE. I'm not sure if one would need to look at production.log to see the exact message.

That said, I think the only way this condition could be met is if the user has a value already set, and then that value later becomes unavailable (an upgrade to ruby or to the operating system removes the currently configured TLS version).

As long as we test on all supported platforms, I think our testing would catch that and appropriate intervention would be made before an end user would ever face that issue.

I kept and updated the check from the previous implementation (which seems also unlikely to be reached in real world usage) just in case.

If you still have concerns about it, maybe we can create a new issue to follow up? Let me know what you think, thanks.

app/lib/katello/resources/cdn.rb
```diff
@@ -1,7 +1,18 @@
 module Katello
   module Resources
     module CDN
-      SUPPORTED_SSL_VERSIONS = ['SSLv23', 'TLSv1'].freeze
+      # the if defined? check is included for TLSv1.3 because it is not available on all platforms.
+      # at the time of writing, TLSv1 through TLSv1.2 should be available on all supported platforms,
+      # but we check regardless in case they are removed from any platform in the future.
+      # any unavailable protocols will be removed from the final Hash by the compact method.
```
ehelms, line 7: TLSv1.3 is available in our production platform, the one that end users use and I think that should aim to be a driving force. I'd agree that being able to test is important but is it the right choice for our ends users is an important question. For example, when we move to Github Actions testing or Ruby 3, this whole issue goes away. Therefore I think we should either:

 1. Default to TLSv1.3 as the newest standard
 2. Update this wording to indicate this is due to an issue with our Jenkins test environment and make sure we follow up on it when GHA or Ruby 3 happens for Katello
 
 @ekohl @evgeni y'all got any thoughts?

evgeni: Before the GHA move, we could try running Katello CI on el8 Jenkins nodes, which should also fix the problem if not available openssl-new-enough.

ekohl: Does the Red Hat CDN provide TLS 1.3? Would it cause issues with any proxy the customer is using? I think that's what determines if it can be the default.

Though I wonder about this setting at all. Why do we have it? I know it's there because in the past it needed to be changed to allow insecure corporate proxies that didn't support the latest versions, but why do we only have it for the CDN? Not for other HTTP connections? Or does it also affect custom repos?

ekohl:
Another way to approach this: get rid of the whole setting and respect crypto-policies? https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/security_hardening/using-the-system-wide-cryptographic-policies_security-hardening describes how users can define these policies system wide. We need to consider container use cases, but I think it's worth taking a step back and look at the big picture.

test/lib/resources/cdn_test.rb
```diff
@@ -1,20 +1,39 @@
 require 'katello_test_helper'
 class CdnResourceTest < ActiveSupport::TestCase
-  def test_http_downloader_v2
-    Setting[:cdn_ssl_version] = 'SSLv23'
+  def test_http_downloader_tlsv1
+    Setting[:cdn_min_tls_version] = 'TLSv1'
     cdn_resource = Katello::Resources::CDN::CdnResource.new('http://foo.com')
-    assert_equal cdn_resource.http_downloader.ssl_version, 'SSLv23'
+    assert_equal cdn_resource.http_downloader.min_version, 769
```
ekohl, line 6: I would not use these magic numbers. Instead, rely on constants.

app/lib/katello/resources/cdn.rb
```diff
@@ -107,11 +118,11 @@ def http_downloader
           end
 
           # NOTE: This was added because some proxies dont support SSLv23 and do not handle TLS 1.2
-          # Valid values in ruby 1.9.3 are 'SSLv23' or 'TLSV1'
+          # Valid values in ruby 2.7.8 are :TLS1, :TLS1_1, :TLS1_2, :TLS1_3
```
ekohl, line 121: It really doesn't depend on Ruby, but more on OpenSSL.

ekohl:
> Another way to approach this: get rid of the whole setting and respect crypto-policies? https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/security_hardening/using-the-system-wide-cryptographic-policies_security-hardening describes how users can define these policies system wide. We need to consider container use cases, but I think it's worth taking a step back and look at the big picture.

Some initial testing suggests that Ruby will happily connect to TLS 1.0 servers even if the system policy is TLS 1.2+ though https://access.redhat.com/articles/3666211 implies it should, given it's not listed in the exceptions. According to strace it doesn't open the policies:

```console
# grep crypto output
openat(AT_FDCWD, "/lib64/libcrypto.so.1.1", O_RDONLY|O_CLOEXEC) = 5
openat(AT_FDCWD, "/proc/sys/crypto/fips_enabled", O_RDONLY) = 5
stat("/etc/crypto-policies/back-ends/opensslcnf.config", {st_mode=S_IFREG|0644, st_size=647, ...}) = 0
openat(AT_FDCWD, "/etc/crypto-policies/back-ends/opensslcnf.config", O_RDONLY) = 6
openat(AT_FDCWD, "/etc/crypto-policies/back-ends/openssl.config", O_RDONLY) = 6
```

So it reads `/etc/pki/tls/openssl.cnf` and that has an `include` statement. `/etc/crypto-policies/back-ends/opensslcnf.config` then has `TLS.MinProtocol = TLSv1.2`, but that isn't respected. After some digging I've opened https://github.com/ruby/openssl/issues/709.

The short version is that `min_version` is set unconditionally to `OpenSSL::SSL::TLS1_VERSION` [here](https://github.com/ruby/openssl/blob/1fa9fc54ebc480dcab435212d020ec29068bac48/lib/openssl/ssl.rb#L25). When I comment that line out it does respect `TLS.MinProtocol`.

ekohl:
> Is there any need for continued support for insecure protocols such as SSLv3 ?

https://github.com/ruby/openssl/pull/710#pullrequestreview-1809357248 pointed out OpenSSL 1.1.0+ (as shipped in EL8) is compiled without SSL 3.0 so this setting is effectively useless since we dropped EL7 support.

AFAIK there haven't been bug reports so I think it raises the question: can we just drop this setting altogether?

wbclark:
Thanks @ekohl . Let me make sure I'm following: you are proposing that in the future once https://github.com/ruby/openssl/pull/710 is available in our ruby, we could remove these settings entirely and rely on the system wide crypto policy to say what protocols are allowed, is that right?

ekohl:
I'd argue for removal without replacement now, even if my PR (which is now merged) isn't released yet. Today's setting limits TLS version but in a way that the default is better. So we wouldn't regress. Over time we'd gain the more secure default

lib/katello/plugin.rb
```diff
@@ -377,12 +377,12 @@ def katello_template_setting_values(name)
...
         include_blank: N_("no global default")
 
-      setting 'cdn_ssl_version',
+      setting 'cdn_min_tls_version',
         type: :string,
-        default: nil,
-        full_name: N_('CDN SSL version'),
-        description: N_("SSL version used to communicate with the CDN"),
-        collection: proc { hashify_parameters(Katello::Resources::CDN::SUPPORTED_SSL_VERSIONS) }
+        default: Katello::Resources::CDN::SUPPORTED_TLS_VERSIONS.include?('TLSv1.3') ? 'TLSv1.3' : 'TLSv1.2',
```
wbclark, line 382: @ehelms 

I set a default of TLSv1.3, with a fallback to TLSv1.2 in case TLSv1.3 is unavailable for some reason

I then reprovisioned my dev environment to perform a clean test. Indeed, when I loaded the WebUI for the first time, the setting was already populated with the value for TLSv1.3

Following that, I performed basic testing like uploading a subscription manifest and enabling and syncing repositories. All of this was working as expected with TLSv1.3

With the fallback to TLSv1.2 if TLSv1.3 is unavailable, the test suite runs just fine. The TLSv1.3 specific test is skipped since the protocol is unavailable, and `test_http_downloader_default_version` confirmed that TLSv1.2 was the default version.

wbclark:
> I'd argue for removal without replacement now, even if my PR (which is now merged) isn't released yet. Today's setting limits TLS version but in a way that the default is better. So we wouldn't regress. Over time we'd gain the more secure default

What about a scenario where a user wants to use inter-server sync to sync from an older server still running on an EL7 host (for example, they are migrating that content off of the EL7 host and onto their new host). They may not want to reconfigure the host's crypto policy just for this one task (and our approach here is easier).

Also, the latest version I've pushed here sets TLSv1.3 as the default wherever it is available. I'm not sure when the ruby openssl fix will be available, or if it will be backported, but I'm assuming this change could be delivered to users much faster.

ekohl:
How many users actually want this feature? It was introduced for compatibility. Since we moved to EL8 the SSLv3 option doesn't do anything (since OpenSSL 1.1.0+ is compiled without it) and I don't think people asked about this. I'd argue simplifying the code base is more important than the edge case where someone wants so specific control over the connections. And if they do, why only this one

wbclark:
> How many users actually want this feature? It was introduced for compatibility. Since we moved to EL8 the SSLv3 option doesn't do anything (since OpenSSL 1.1.0+ is compiled without it) and I don't think people asked about this. I'd argue simplifying the code base is more important than the edge case where someone wants so specific control over the connections. And if they do, why only this one

Well, there was a downstream BZ so someone is definitely trying to use it. I'd defer to @pmoravec about whether this is potentially useful or important since he has deep knowledge of user deployments and different scenarios they are facing, and to @ianballou or @parthaa about whether there's a concrete benefit for the health of the katello code base to remove it.

ekohl:
> Well, there was a downstream BZ so someone is definitely trying to use it. I'd defer to @pmoravec about whether this is potentially useful or important since he has deep knowledge of user deployments and different scenarios they are facing, and to @ianballou or @parthaa about whether there's a concrete benefit for the health of the katello code base to remove it.

https://bugzilla.redhat.com/show_bug.cgi?id=2216445 is from @pmoravec and the problem there is that the current default is *less* secure than if we dropped it (and can cause problems). I'll let him answer if allowing TLS 1.0+ (which will include TLS 1.3) but not giving explicit control is (as I suggested) solving the issue.

wbclark:
> https://bugzilla.redhat.com/show_bug.cgi?id=2216445 is from @pmoravec and the problem there is that the current default is _less_ secure than if we dropped it (and can cause problems). I'll let him answer if allowing TLS 1.0+ (which will include TLS 1.3) but not giving explicit control is (as I suggested) solving the issue.

The new setting would give users the _option_ of reducing the min_version if they absolutely need to for compatibility with other infrastructure (syncing from the CDN via an older proxy, syncing from an older katello server using ISS) but by _default_ it would enforce the more secure TLS1.3 standard.

Given the existence of upgrade paths that involve building a new server on a newer OS and migrating content via ISS, which we've done in the past and could do again in the future, I think it's better to allow admins to quickly and easily modify the allowed min_tls version in the UI, rather than the more error-prone and higher risk process of modifying the system wide crypto policy.

Given that the current version of `cdn_ssl_version` has default value `nil`, I think this means that the user must have tried to enable the setting to cause the issue (indicating to me that at least someone is trying to use this?)

Also as you already pointed out, Net::http has ssl_version (which we are setting nil unless the user changes that default) and min_version, corresponding to ssl_version (deprecated) and min_version in OpenSSL::SSL::SSLContext. So what really happens today would depend on the issue you discovered and fixed with https://github.com/ruby/openssl/pull/710 , which would suggest we are allowing the insecure TLS 1.0 protocol for CDN syncs (...unless Net::http sets any defaults of its own instead of falling back to the ruby openssl library default?)

Anyway, to summarize my concerns about removing it:

1. We don't know how long it will take for the ruby fix to be shipped, and your proposal could potentially delay us from fixing the issue of allowing insecure TLS 1.0 by default.

2. We could set TLS 1.3 as the default today instead of waiting for the ruby fix to give users TLS 1.2 as a default

3. It's easier and less error prone to set this via the UI than to modify system wide crypto policy, particularly for something like one-time ISS to migrate content during an upgrade.

4. There was a use case for a setting like this in the past, and the stated reasons seem likely to persist today and into the future, which is supported by the bug report which says that someone was trying to use a non-default value.

wbclark:
> But what do you win with it? System wide crypto policies are probably better for compliance. For example, [NIST SP 800-52 Rev. 2](https://csrc.nist.gov/pubs/sp/800/52/r2/final) mandates TLS 1.2 as the minimum version with support for TLS 1.3. So anyone who wants compliance with that should already set it via a system wide crypto policy.

I know that you are aware of this, but to emphasize for anyone else reading along, if we don't set the `min_version` then _the system wide crypto policy is not used and we get an insecure protocol by default, and the user has no way to change it_. For that reason alone, this PR should be merged immediately. _Taking this step would not preclude further discussion about redesigning or dropping this feature in the future_.

> I don't think you really win a lot with this. TLSv1.2 is still considered secure and EL7 hosts don't have TLSv1.3 support, so you'd likely break a lot of deployments out of the box. Like where there's an EL7 HTTP proxy.

This is why I previously had TLSv1.2 as the default value, not TLSv1.3, although I can see the argument for setting TLSv1.3 anyway and accepting that users who can't use TLSv1.3 will need to intervene by lowering it to TLSv1.2. Either of these choices is preferable to the current status quo, and either of these choices is also preferable to dropping this feature immediately without going through design and planning, RFC, communicating the change, etc.

wbclark:
N.B. I split this portion of my response into a separate comment, to emphasize that I see these details as being less urgent than the primary points I made above.

> It was there to allow for a downgrade in security. It was introduced in https://github.com/Katello/katello/commit/280a77fe77f9ac811107018127151a31f77ae033 and https://projects.theforeman.org/issues/8441 tells us that some proxies lacked TLS 1.2 support. I don't think that's relevant anymore.

> This patch would only partially solve it, not for the whole platform. I don't think a partial fix is worth all the complexity. Both in maintenance and testing, but also for users consuming the product.

So it was introduced in https://github.com/Katello/katello/pull/4823 and some proxies at the time lacked TLS1.2 support. While it is true that this exact circumstance is no longer exactly the case right now, I think it's very likely that a similar circumstance will arise again in the future as the world gradually (and unevenly) migrates to future OSes and TLS protocol versions.

I think a partial fix is better than no fix at all. Previously this was configured via yaml and the installer which would have been a fine way to orchestrate katello and pulp for a consistent configuration, but that was removed and it was moved to the DB and controlled via settings API with https://github.com/Katello/katello/pull/8552 , https://github.com/theforeman/puppet-katello/pull/320 , and https://github.com/theforeman/foreman-installer/pull/477 .

ekohl:
> I think a partial fix is better than no fix at all

And I disagree with this. Remember that this is from the time when we didn't have unified proxy code. Today if you set up a proxy, you go through Foreman's HTTPProxy. That has code that patches every connection adapter.

Imagine this scenario. You set the minimum TLS version for the CDN to be 1.3, that works. Then later you set a global proxy that doesn't understand TLS 1.3. Suddenly the CDN connection breaks, but all other connections work. How do you expect the user and anyone supporting the user to figure out why that one aspect doesn't work?

My thought is that this current feature helps exactly nobody and introduces a maintenance and usability burden. It's making a tiny aspect of the application special and that's more likely to cause problems than solve any.

I'm not a Katello maintainer, but with my platform hat on I'd say the broken setting needs to be removed without replacement.

ehelms:
## Understanding

I am going to attempt to try and summarize what I think I understand the requirements to be to ensure I properly understand them before delving into the implementation details:

 1. Any TLS connection should default to any system-wide crypto policies.
 2. If no system-wide setting exists, TLS connections should default to the most secure protocol possible.
 3. Users should have a method by which to lower the TLS version in use in the event their infrastructure prevents the use of the defaults.

For #3, I avoided what I think is a key point, and requirement up for debate:

 * Should the user have to downgrade *all* connections or be able to selectively downgrade connections based on what is breaking?

For #1, I think this is a best practice that we should be designing for across the entirety of our application.
For #2, I think fall backs just make sense, and we should also strive to provide the securest method possible by default.
For #3, environments vary from use case to use case, and we best serve our customers by providing some supported and sane fallback that allows them to continue operations within reason. For example, we should not let the user downgrade the connection to something that is beyond deprecated.


I believe the design hinges on #3 and the point that is up for debate. I have no love for over doing it with settings. I am wary of a solution that would give users no way to adjust the configuration for their environment (even if we think they are making poor security choices). 

Another way to put it is, if we opted not to provide a method to lower the TLS version, what would we tell a user who comes looking for help?

ianballou:
> Current PR: The default value (`nil`) uses the Ruby default for the CDN connection, which doesn't respect the minimum TLS version (setting it to TLS 1.0) but does respect the maximum TLS version.

I thought the default min value was hardcoded for the current PR?

```
default: Katello::Resources::CDN::SUPPORTED_TLS_VERSIONS.include?('TLSv1.3') ? 'TLSv1.3' : 'TLSv1.2',
```

Perhaps the code should change the setting default to `nil` instead of choosing TLS 1.2 or 1.3. That should use the system default, right?

> Imagine this scenario. You set the minimum TLS version for the CDN to be 1.3, that works. Then later you set a global proxy that doesn't understand TLS 1.3. Suddenly the CDN connection breaks, but all other connections work. How do you expect the user and anyone supporting the user to figure out why that one aspect doesn't work?

This does indeed sound like a support annoyance, but isn't the setting for those few very special setups where the user should know they're doing something funky with their CDN connection anyway?

This is definitely one of those weird corner-case features, but since it deals with import/export there are probably users doing some interesting and strange things related to disconnected environments.

> This would be what the current PR is implementing. But I'm questioning if it's adequate. If we implement this, what's the QE plan to verify it really does what it's supposed to?

I think the test would be to run through the original reproducer steps and verify that they now work:

```
Steps to Reproduce:
1. Set Administer -> Settings -> Content -> "CDN SSL version" to the highest possible version (TLSv1)
2. Set ISS from another Satellite 6.13 or higher: Content -> Subscriptions -> Manage Manifest -> CDN Configuration -> Network Sync -> provide upstream Sat details (incl. SSL debug cert)
3. Attempt to Update fails with some "tls mismatch" like error.
```

---

Besides turning this into a full-blown feature where we let you set the min and max TLS version for all connections via the HTTP Proxy model (or something like that), it sounds like the other options are to accept this PR as-is or remove the setting entirely.

Removing the setting entirely means saying "WONTFIX" to the disconnected users who have interesting TLS requirements, but we do gain some maintainability.
  -> We apparently have at least one user who uses this setting, so can we remove it? Perhaps more  disconnected folks use it?

I'm thinking removing the setting could cause complaints.

ekohl:
On mobile, so limited quoting capabilities. But on that last point: as I understand it today the setting can enforce TLS 1.0 (no less, no more, thus insecure) or the SSLv3 setting (which I think is SSLv3 or TLSv1, but can't check now). OpenSSL 1.1.0 included in EL8 does NOT have SSLv3 compiled in. That means essentially it's TLSv1 or nothing else. In other words, it's broken

wbclark:
Thanks for the thoughtful comments, everyone. So that we are all working from the same understanding, I want to clarify some very important points:

1. The current setting `min_ssl_version` (what this PR replaces) is broken worse than we realized at first. The two possible values today are `TLSv1.0` and `nil` (default), and both of these are broken.

1.a. If the value `TLSv1.0` is used, then we are forcing an insecure, out of date protocol to be used.

This happens because we are currently using (deprecated) `Net::http#ssl_version=` instead of `Net::http#min_version=`, and the deprecated method both enforces a single version instead of allowing a range of versions, and also can't set protocols newer than TLSv1.0. This first issue affects only those customers that attempt to use `TLSv1.0` as the value of `min_ssl_version` instead of the default `nil`.

1.b. The second breakage (what Ewoud discovered) is that _even when the setting is not used_ (the current default behavior, which is the `nil` value), we are still _allowing_ (but not _forcing_) insecure TLSv1.0 to be used. _Almost all users are affected by this flaw, regardless of any system-wide crypto policy they implement_; the only users not affected by this version of the flaw, are the ones who are affected by the even worse version (1.a).

2. There are two proposals for how to move forward:

2.a. (Ewoud's proposal) drop `cdn_ssl_version`, removing the need to fix 1.a by deprecating the feature. Do not take additional action to fix 1.b, just wait for the upstream Ruby fix (recently merged) to make it into our Ruby. For the time being, customers would not be able to lower the minimum TLS version, because _they are already allowing insecure TLS versions and there is no lower to go_. In the future, when Ruby starts respecting the system-wide crypto policy, then customers who require it could lower the minimum allowed TLS version only by modifying the system-wide crypto policy.

2.b. (This PR) drop `cdn_ssl_version` and replace it with `cdn_min_tls_version` fixing 1.a with a proper working implementation. Set a default value for `cdn_min_tls_version` so that 1.b is fixed immediately as well. Once Ruby starts respecting the system-wide crypto policy, we then still have the choice of dropping `cdn_min_tls_version` and system-wide crypto policy becomes the single source of truth (i.e. we change to 1.a) OR we could allow `nil` and make it the default once again, which preserves the ability for users to selectively downgrade connections.

My evaluation is that 2.b is the better approach for now, because it fixes both issues right now, _while still allowing us to pursue either design in the long term_ (i.e. once the Ruby bug is fixed). My increasing concern is that we are debating the design merits of 2.a vs. 2.b as long term ideals, when we should instead move forward with the approach that fixes both flaws today and have a separate conversation about potentially deprecating the feature in the future (ideally, not in a PR review).

What follows are in-line responses to various comments, which I've re-ordered for clarity. First, @ekohl

>Current PR: The default value (nil) uses the Ruby default for the CDN connection, which doesn't respect the minimum TLS version (setting it to TLS 1.0) but does respect the maximum TLS version. Once my ruby-openssl fix makes it to production it will. Any selected value will divert from the system-wide crypto policy for the minimum version.

This could be the (at least partial) source of our disconnect? @ianballou got it right here:

> I thought the default min value was hardcoded for the current PR?
> 
> ```
> default: Katello::Resources::CDN::SUPPORTED_TLS_VERSIONS.include?('TLSv1.3') ? 'TLSv1.3' : 'TLSv1.2',
> ```

That's correct. What I'm proposing is that we have to hardcode this for now to fix 1.b, and after sufficient time we could then either remove `cdn_min_tls_version` and only allow changes via the system-wide crypto policy OR we could introduce a new value, with a migration making it default, that respects the system-wide crypto policy out of the box but still allows the user to selectively downgrade the connection.

In response to @ehelms and @ianballou , who suggested

> Any TLS connection should default to any system-wide crypto policies.

> Perhaps the code should change the setting default to `nil` instead of choosing TLS 1.2 or 1.3. That should use the system default, right?

Neither 2.a nor 2.b would achieve this today, because if we don't explicitly set `min_version` here then `Net::http` won't set it either[1]; in that case, ruby's insecure hardcoded default[2][3] will be used.

[1] https://github.com/ruby/net-http/blob/edc99a54b2c2888759068e2627f2f26c5f505352/lib/net/http.rb#L1635-L1643
[2] https://github.com/ruby/openssl/issues/709
[3] https://github.com/ruby/openssl/blob/fcda6cf9d5f69f6dc0c297ca76feff71f9021f00/lib/openssl/ssl.rb#L25

Until we have a Ruby version with Ewoud's fix, we must explicitly set `min_version` ourselves or else it will always permit the insecure, deprecated TLSv1.0 protocol.

Finally in response to these comments by @ehelms

> Users should have a method by which to lower the TLS version in use in the event their infrastructure prevents the use of the defaults.

> Should the user have to downgrade all connections or be able to selectively downgrade connections based on what is breaking?

This is a good question to ask about the future, and IMO strong arguments have been presented in favor of both approaches. With that said, we don't need to decide that right now, and it would be better to have that discussion on Discourse.

pmoravec:
(2c from somebody who is _not_ a typical developer)

In a long term, relying only on system-wide crypto policy might seem a better approach to me, since that policy covers *all* connections (not only from katello) that are required for getting some content. I have no experience with such policies but I assume it supersets any setting we (ideally) allow by this feature - but system-wide, not "katello speaks to CDN" scenario only. If the policies allow what this feature is supposed to allow, how much sense it has to replicate the same functionality, for fine-tuned user scenario? (not a rhetorical question)

Could there be user scenarios "we need lower SSL/TLS version just for this and that, but not globally / system-wide"? If not, we have an argument for dropping the feature. If so, we shall evaluate such corner cases if they deserve this feature.

That is for a long term. For a short term, until the ruby PR from @ekohl ++ gets into the ruby version that katello uses, we can treat this PR as a "hotfix". So (at least) for a short term, fixing this feature seems to me as a better option than dropping it - we can re-evaluate once the ruby fix gets to katello.

ekohl:
> And comments like these are the other reason I think we should take that aspect of the conversation to Discourse/RFC rather than making a hasty decision to deprecate a feature during a PR review. Deprecating a feature is a complex decision and should be carefully considered

Just jumping on this: the current setting is harmful and there is no logical migration path from one to the other. It's not like you want to store the value that the user has today and transform it something else so we can drop it without deprecation IMHO. Having a discussion about a replacement is a good RFC topic though.

ianballou:
> the current setting is harmful and there is no logical migration path from one to the other.

I haven't caught up with the other replies yet but this stood out -- I thought the setting in its current state did help those few people who used insecure proxies? Or is it actually totally broken?

ianballou:
Now that I understand the situation better, I get that we have an immediate need to bump up the minimum TLS version so that TLSv1 isn't used even for users who do not set the setting in question.

The current PR sets TLS to be a secure setting by default. It also lets users do insecure things for whatever unique environment they have. By that logic, I'm happy with the general strategy. Please let me know if I'm missing something.

I can do an actual functionality test on the code once there's a consensus.

wbclark:
> the current setting is harmful and there is no logical migration path from one to the other.

The really important point to me is that the harm exists even if the setting is removed. I am worried about something like a MITM attack that exploits the (always allowed) insecure protocol. Maybe that needs to be chained with other exploits to compromise the supply chain. I am not comfortable with that risk.

> It's not like you want to store the value that the user has today and transform it something else so we can drop it without deprecation

The current value of `cdn_ssl_version` is still stored in the DB if it was defined previously. IF we later decide to keep `cdn_min_tls_version` and introduce a new default value `"Use system-wide crypto policy"` then we have the option of a migration that considers both the stale `cdn_ssl_version` value (if any) and `cdn_min_tls_version` in determining whether to swap to the new default.

We aren't forced to do this, we still have (in the future) the option of dropping `cdn_min_tls_version` entirely.

> For a short term, until the ruby PR from @ekohl ++ gets into the ruby version that katello uses, we can treat this PR as a "hotfix". So (at least) for a short term, fixing this feature seems to me as a better option than dropping it - we can re-evaluate once the ruby fix gets to katello.

:+1: from me. I want to emphasize again that I am hearing some good arguments in favor of dropping `cdn_min_tls_version` _in the long term_. I think there are potentially some good arguments for keeping it too, but either way, since we _can't_ drop it entirely right now, we don't need to decide that yet.

> The current PR sets TLS to be a secure setting by default. It also lets users do insecure things for whatever unique environment they have. By that logic, I'm happy with the general strategy. Please let me know if I'm missing something.

This is how I see it too.

Sincerely, thank you to everyone for the time, effort, and patience involved in figuring this out.

ekohl:
> The really important point to me is that the harm exists even if the setting is removed. I am worried about something like a MITM attack that exploits the (always allowed) insecure protocol. Maybe that needs to be chained with other exploits to compromise the supply chain. I am not comfortable with that risk.

The whole of the project is vulnerable to that. Why is the CDN connection special?

To perform a successful MITM attack you would also need to spoof the DNS resolution and the certificate. If you can do that, providing it over TLS 1.2 or 1.3 isn't a big challenge.

Preventing MITM can best be done by certificate pinning and verifying the certificate. AFAIK certificate pinning is what many mobile applications do and it turns out we already do this in Katello. cdn.redhat.com doesn't use a globally signed certificate so we bundle it in Katello (https://github.com/Katello/katello/blob/master/ca/redhat-uep.pem). I think the verification happens, but not entirely sure what this code does:
https://github.com/Katello/katello/blob/cc1b7cc0a624e00a6b59b5bae2be298a8ba6331e/app/lib/katello/resources/cdn.rb#L32

I would have expected it to use some constant, but value 9 is the bitwise OR of 1 and 8. Reverse engineering this, we know these are the various values:
* 0 `OpenSSL::SSL::VERIFY_NONE`
* 1 `OpenSSL::SSL::VERIFY_PEER`
* 2 `OpenSSL::SSL::VERIFY_FAIL_IF_NO_PEER_CERT`
* 4 `OpenSSL::SSL::VERIFY_CLIENT_ONCE`

Since 8 undefined, I think this is equivalent to `OpenSSL::SSL::VERIFY_PEER` so the certificate is verified.

We should change this magic `9` value to some constant instead. It was introduced in f9d4b104fbfbd435dcc70e8714d8b770271a5236 so it's seriously been there for a long time.

So 'm not worried about MITM attacks because we apply better protections already. But perhaps I'm wrong and with TLS 1.0 you can MITM even with certificate verification.

> The current value of `cdn_ssl_version` is still stored in the DB if it was defined previously.

If we do this, the applications raises a warning on every start up, which is a bad user experience.

> IF we later decide to keep `cdn_min_tls_version` and introduce a new default value `"Use system-wide crypto policy"` then we have the option of a migration that considers both the stale `cdn_ssl_version` value (if any) and `cdn_min_tls_version` in determining whether to swap to the new default.
> 
> We aren't forced to do this, we still have (in the future) the option of dropping `cdn_min_tls_version` entirely.

How are you going to convert the current value? I assumed you can't really differentiate between the current value and the new, but now that you mention it, perhaps you can.

ekohl:
I forgot about this until @evgeni reminded me of it: https://github.com/ruby/net-http/pull/55 means that HTTPS proxies are currently not supported. That means the only way I can think of why this was needed is [Deep Packet Inspection](https://en.wikipedia.org/wiki/Deep_packet_inspection). In the past many of these middleware boxes broke old TLS support. If you can't support TLS 1.2 these days, large parts of the internet are no longer reachable. So I wouldn't be really concerned with supporting them. I'd also say it's questionable if they were ever supported.

wbclark:
> I forgot about this until @evgeni reminded me of it: [ruby/net-http#55](https://github.com/ruby/net-http/pull/55) means that HTTPS proxies are currently not supported. That means the only way I can think of why this was needed is [Deep Packet Inspection](https://en.wikipedia.org/wiki/Deep_packet_inspection). In the past many of these middleware boxes broke old TLS support. If you can't support TLS 1.2 these days, large parts of the internet are no longer reachable. So I wouldn't be really concerned with supporting them. I'd also say it's questionable if they were ever supported.

That's not my reading of https://bugs.ruby-lang.org/issues/16482 ... "Note that this issue is not about the connection that is routed through the proxy but about the connection to the proxy itself."

ekohl:
That's exactly my point. From the Foreman perspective, if you use a proxy today you MUST have:

Foreman -> HTTP -> Proxy -> HTTP(S) -> target

You can't have:

Foreman -> HTTPS -> Proxy -> HTTP(S) -> target

Foreman can't enforce a minimal TLS version on the Proxy -> target connection.

So `cdn_min_tls_version` would not apply to Pulp and not apply if you use a HTTP proxy. All in all I'm still convinced this setting is inadequate while adding a burden to both users and maintainers.

wbclark:
> That's exactly my point. From the Foreman perspective, if you use a proxy today you MUST have:
> 
> Foreman -> HTTP -> Proxy -> HTTP(S) -> target
> 
> You can't have:
> 
> Foreman -> HTTPS -> Proxy -> HTTP(S) -> target
> 
> Foreman can't enforce a minimal TLS version on the Proxy -> target connection.
> 
> So `cdn_min_tls_version` would not apply to Pulp and not apply if you use a HTTP proxy. All in all I'm still convinced this setting is inadequate while adding a burden to both users and maintainers.

OK, and what if you _don't_ have an HTTP proxy?

> If there is a system-wide config option for the same, that would make much more sense (and be the argument to just remove this setting).

There is, but Ruby doesn't respect it currently, so unless we are explicit in specifying it ourselves, TLSv1.0 (an insecure, out of date protocol) is always used as the min_tls_version.

I think we've all been in agreement that this setting could be removed in the long term when the Ruby bug is fixed.

ianballou:
HTTPS proxies aside, this setting still enables syncing Foreman from some anciently configured Foreman that requires old TLS right?

ianballou:
There hasn't been much movement for a while, and understandably so given the conversation.

However, in following the debate above on "keep the cdn_ssl_version setting" vs "remove the setting", I'd like to confirm:

Removing "cdn_ssl_version" will stop users from being able to sync an older (TLS v1) upstream Foreman with a newer downstream Foreman.

If this is true, I'd like more outside opinions about if we need to continue supporting this feature.

If it's false and either that feature is already broken or maybe I'm just misunderstanding the debate above, then I would be more in support of removing the setting entirely and raising the min tls version to the lowest secure version.

wbclark:
Closing this, as we addressed it with https://github.com/Katello/katello/pull/10926

Thanks all!

PR #10825, opened 2023-12-13
Fixes #36986 - Warn users on traces that require reboot

jeremylenz:

Add a warning on the Traces tab when one of the selected traces is "static" type, and thus requires a reboot of the host.

Also
* added a warning icon next to the helper, and
* changed the button text to "Reboot Host" just in case you didn't see _any_ of those other things

image

With select all mode, we haven't downloaded the data for all selected traces, so we don't know for sure if one of the undownloaded traces requires a reboot. The best we can do in that case is see if any of the results we _have_ downloaded require a reboot. If so, it'll show the warning in select all mode as well. However, there are two edge cases where the warning will be a lie:

 - If you select all, and then deselect all of the static-type traces, the warning will still show.
 - If you select all, and the current page of results happens not to have any static traces, the warning will _not_ show.

I consider these cases to be rare enough not to need special treatment.

Set up a box with traces (easiest way for me is to spin up a new rhel8, then apply all errata)
Go to the host details / Traces tab and verify the warning shows

### DIFF
--- a/webpack/components/extensions/HostDetails/Tabs/TracesTab/TracesTab.js
+++ b/webpack/components/extensions/HostDetails/Tabs/TracesTab/TracesTab.js
@@ -2,10 +2,11 @@ import React, { useState, useCallback } from 'react';
 import { FormattedMessage } from 'react-intl';
 import {
   Skeleton, Split, SplitItem, ActionList, ActionListItem, Dropdown,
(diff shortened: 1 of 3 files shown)


ehelms:
[test katello]

ianballou (approved):
Works great, checked static and daemon trace types. The warnings made sense when selecting a number of them in different combinations.

PR #10827, opened 2023-12-15
Fixes #36987 - Remove param :source_url since it's no longer supported

sjha4:
We have not supported :source_url param to override repository URL during sync which was supported in pulp2 world since upgrading to pulp3.
Make sure all references to :source_url are removed.


sjha4:
[test katello]

chris1984 (approved):
ACK

PR #10830, opened 2024-01-02
Fixes #37018 - Use correct function signature for fetchBulkParams

jeremylenz:
In https://github.com/Katello/katello/pull/10783, the `fetchBulkParams` method was removed from Katello.

In https://github.com/theforeman/foreman/pull/9860, it was moved to Foreman.

In the course of doing so, the function signature was changed from

```js
const fetchBulkParams = (idColumnName = idColumn) => {
```

to

```js
const fetchBulkParams = ({
    idColumnName = idColumn,
    selectAllQuery = '',
  }) => {
```

This means that you now have to pass an object instead of a string. This breaks bulk repository set overrides in the web UI.

* Correctly pass an object to `fetchBulkParams`


parthaa (approved):
Couldn't find other places that were impacted by the `fetchBulkParams`  signature change. 
Works well ACK

PR #10831, opened 2024-01-03
Fixes #36995 - bump recommended Sat repos to 6.15

ianballou:
Bumps recommended Satellite repositories to 6.15

I'm assuming that the repository label scheme is still the same.

Nothing for now, the repositories won't show up until they're published on the RH CDN.


jeremylenz (approved):
LGTM!

PR #10742, opened 2023-09-18
Opt-in email notifications when {repo sync, cv promote, cv publish, proxy sync} fails

adamruzicka:
#### What are the changes introduced in this pull request?
Opt-in email notifications when {repo sync, cv promote, cv publish, proxy sync} fails

#### What are the testing steps for this pull request?
1) Subscribe to each of the things mentioned above
2) Make them fail. In general `raise "HALT"` somewhere in the action works pretty well, but it cannot happen before action_subject is called

TODO:
- [x] redmine

PR #10835, opened 2024-01-04
Fixes #37023 - check for CapsuleContent::Sync options existence in last_env_sync_task

ianballou:
The `CapsuleContent::Sync` task input didn't contain an `options` value until https://github.com/Katello/katello/pull/10742/files#diff-a34bd4b978d218a78d31d4dfc93bd71f3a50368135c47fc6b0b4650b7e72e092R25

The code currently assumes it is there. So, after upgrading, until a sync is done, the smart proxy page will throw an error.

...

1) Create a Katello 4.10 machine with a smart proxy mirror
2) Sync the smart proxy mirror
3) Upgrade to 4.11 and see the `undefined method `[]' for nil:NilClass` error pop up on the smart proxy details page
4) See the error again with `hammer proxy content info ...`
5) See the error yet again with `hammer proxy content synchronization-status ...`
6) Patch in this PR and the errors should go away.


sjha4 (approved):
LGTM 👍🏼

PR #10836, opened 2024-01-05
Fixes #37032 - Drop minitest-tags development dependency

ekohl:

minitest-tags allows developers to run a single test, but it hasn't been updated since 2012.

Some developers may be using this functionality. Perhaps there are alternatives. Foreman also uses minitest so it would be better to discuss it in a wider group.

Nothing. It drops some developer functionality, so more of a question: does anyone use that?

### DIFF
--- a/katello.gemspec
+++ b/katello.gemspec
@@ -68,7 +68,6 @@ Gem::Specification.new do |gem|
 
   # Testing
   gem.add_development_dependency "factory_bot_rails", "~> 4.5"
-  gem.add_development_dependency "minitest-tags"
   gem.add_development_dependency "mocha"
   gem.add_development_dependency "vcr", "< 4.0.0"
   gem.add_development_dependency "webmock"


ianballou:
We should check that `ktest` doesn't use this.

ekohl:
I don't see it in https://github.com/theforeman/puppet-katello_devel/blob/master/templates/ktest.sh.erb. Having said that, I wonder why it's needed. I'd assume you can use `TEST=input_to_file_list`. That's defined here:
https://github.com/ruby/rake/blob/5476cda5c368773c5198a7157d032fe4fc93d795/lib/rake/testtask.rb#L163-L164

So I suspect for most uses `ktest` could be replaced with:
```sh
#!/bin/bash
if [[ -n $1 ]] ; then
  TEST=$1
  export TEST
fi

bundle exec rake test:katello
```

And IMHO that's also easy enough to remember that it doesn't need to be a separate script. Of course, that's a separate discussion from this one.

ianballou (approved):

PR #10837, opened 2024-01-05
Fixes #37084 - Drop simplecov dependency

ekohl:

This is a test dependency, but marked as a development dependency. We currently install all development dependencies while testing katello, but don't do that for any other plugin. This is an effort to reduce the development dependencies to pure development dependencies.

We may want this, but then we need to figure that out. Currently this is marked as a draft because it's only up for discussion now.

One option is to rescue `LoadError` if it's unavailable and gracefully deal with it.

Big question is: does anyone use these coverage reports?

### DIFF
--- a/katello.gemspec
+++ b/katello.gemspec
@@ -72,7 +72,5 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency "mocha"
   gem.add_development_dependency "vcr", "< 4.0.0"
   gem.add_development_dependency "webmock"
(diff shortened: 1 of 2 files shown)


jeremylenz:
> does anyone use these coverage reports?

No one from Katello that I know of uses them. I would be fine dropping it entirely. @ianballou ?

ianballou:
I'm cool with dropping it, I don't think the reports are used.

parthaa (approved):

PR #10838, opened 2024-01-05
Fixes #37033 - Drop unused katello:rubocop:jenkins rake task

ekohl:

The task is dropped and the unused `rubocop-checkstyle_formatter` dependency is removed.

Jenkins uses katello:rubocop and doesn't parse the results right now, so the dependency can be dropped safely.

https://github.com/theforeman/jenkins-jobs/pull/400 enables parsing of the results again, but these days the default formatter (progress) output can be parsed by Jenkins.

Verify CI still passes.


ianballou (approved):
Thanks for the cleanup!

PR #10839, opened 2024-01-09
Fixes #36829 - supprimer used incorrectly in French translations

ianballou:
Updates all translations to be up-to-date with Transifex. Also fixes the wrong translations noted in https://projects.theforeman.org/issues/36829 and https://projects.theforeman.org/issues/36945

...

1) Switch to French
2) Check that the "Create Activation Key" button says "Creer une cle d'activation" (with accent marks)
3) Check that the "Create Product" button says "Creer un produit" (with accent marks)


ekohl (approved):
The GitHub UI does not like this PR. It may be worth looking at the translation files and look at their percentages. If they're below a certain number, perhaps drop them?

For now I'd merge this though.

chris1984 (approved):
ACK

ianballou:
> The GitHub UI does not like this PR. It may be worth looking at the translation files and look at their percentages. If they're below a certain number, perhaps drop them?

I agree, no sense really keeping anything that isn't well-translated. Hmm maybe 50% ?

PR #10840, opened 2024-01-10
Fixes #37040 - Fix outdated registration wording for AKs

jeremylenz:

1. Fix outdated wording on the Activation Key details page which recommends a deprecated host registration method.
2. Fix outdated wording on the Activation Key tab of the Host Group details page which recommends a deprecated host registration method.
3. In those places, add links to the new Global Registration page.
4. :partying_face: Bonus! :partying_face: Improve the Global Registration form so that we can have a certain activation key selected immediately on page load.

Activation Key details page:
image

Host Group details page:
image

Create at least one activation key
Create a host group
Verify the new wording on both pages

Verify the new features of the Host Registration form:

1. Link from Activation Key details page works and pre-selects the activation key
2. Link from Host Group details page works and does not pre-select any key (unless there's only one in the system, see no. 5 below)
3. Loading the page with `?initialAKSelection=` and a valid activation key pre-selects that key
4. Loading the page with `?initialAKSelection=` and an _invalid_ activation key does not error and does not pre-select a key
5. Loading the page when only one activation key exists still pre-selects that activation key, with or without  `?initialAKSelection=`

### DIFF
--- a/app/views/overrides/activation_keys/_host_tab_pane.html.erb
+++ b/app/views/overrides/activation_keys/_host_tab_pane.html.erb
@@ -21,9 +21,11 @@
     </div>
 
     <p><%= _('Activation keys and subscriptions can be managed') %>
-      <b><a href="/activation_keys" target="_blank"> <%= _('here') %></a></b>
+      <b><a href="/activation_keys" target="_blank"> <%= _('here.') %></a></b>
     </p>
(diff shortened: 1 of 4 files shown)


ianballou (approved):
Working well for me! Next up I think pre-filled host group as well would be cool :)

PR #10841, opened 2024-01-10
Fixes #36897 - Hosts now consistently follow rules when inheriting at...

qcjames53:
…tributes from their hostgroups

Despite the simplicity of the bug, there were some big changes introduced to how we handle host inheritance from host groups. On host creation, a host should still inherit the lifecycle environment and the content view from its hostgroup; these two properties (among a few others) are referred to as 'content facets'. The new changes introduced are as follows:

- On creation, hosts which do not have certain content facets assigned will inherit these properties from their host group (if it exists).
- Hosts can be manually requested to have certain content facets differing from the host group on creation. These requests will now be honored.
- Hosts that are given either a CV or an LE at creation can now inherit the missing content facets from the host group.
- Hosts can be updated to have different CVs and LEs from their host group. Unlike before, no updates (either to the host group or the host) will modify these properties unless the CV or LE itself is directly updated.

There are a lot of ramifications to this change. Jeremy and I talked about the expected behavior for a while but there is likely stuff that we missed or were wrong about. I'd be happy to chat about this PR any time.

One big consideration is that updating host groups does not update the host content LE or CV. Is this a feature users need/want or are they indifferent?

1. Create two different LEs, each with their own CV.
2. Create a host group assigned to LE 1.
3. Create a host assigned to the host group and register it. Ensure that the CV and LE match the host group. I had to use a VM host to get the registration to work properly, but allegedly there is an easier way to do this?
4. On the host details page for your host, click the menu next to 'Content view details'. Select 'Edit content view assignment' and assign the other LE and CV. Upon saving, you should see the content view details window update.
5. Click 'Edit' in the upper corner to edit the host. Modify the comments field in 'Additional information' and submit the changes. The CV and LE selected before should not have changed.

That takes care of the bug portion of this PR. Due to the changes with how inheritance is handled, it would be beneficial to check the functionality of a few other styles of host:
- Is behavior normal for hosts without a hostgroup?
- Is behavior normal for hostgroups that don't have a content view assigned?
- Check that hostgroup changes do not change existing hosts. Is this ok?

You can take a look at `host_managed_extensions_test.rb` for the theory behind my testing. Please ensure that these tests are sound.


app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -48,28 +48,51 @@ def inherited_attributes
...
-            attributes['content_facet_attributes'] = facet_attrs
-          else
-            Rails.logger.debug "Hostgroup has content view and lifecycle environment assigned; using those"
+        def apply_inherited_attributes(input_attributes, initialized = true)
+          attributes = super(input_attributes.clone, initialized) || {}
+
+          # When the input explicitly sets content_facet_attributes, keep them
+          # regardless of what any higher function calls added
+          if initialized && input_attributes.key?(:content_facet_attributes) &&
+            attributes.key?(:content_facet_attributes)
```
ekohl, line 57: This line formatting is confusing. I'd expect RuboCop to disagree with it, because it gives the impression to be part of the code block rather than the condition.

app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -48,28 +48,51 @@ def inherited_attributes
...
           attributes
         end
 
+        def apply_inherited_content_view(attributes)
+          facet_cv_id = content_facet&.single_content_view&.id
+          is_missing_cv = !attributes.key?(:content_facet_attributes) ||
+              attributes[:content_facet_attributes][:content_view_id].blank?
+          if is_missing_cv && facet_cv_id.present?
+            Rails.logger.info "Host has no content view assigned; using hostgroups's existing content view"
+            attributes[:content_facet_attributes] = {} unless attributes.key?(:content_facet_attributes)
```
ekohl, line 78:
```suggestion
            attributes[:content_facet_attributes] ||= {}
```

app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -48,28 +48,51 @@ def inherited_attributes
...
+            attributes = apply_inherited_content_view(attributes)
+            attributes = apply_inherited_lifecycle_environment(attributes)
           end
           attributes
         end
 
+        def apply_inherited_content_view(attributes)
+          facet_cv_id = content_facet&.single_content_view&.id
+          is_missing_cv = !attributes.key?(:content_facet_attributes) ||
+              attributes[:content_facet_attributes][:content_view_id].blank?
```
ekohl, line 75: Wouldn't `dig` work here?
```suggestion
          is_missing_cv = attributes.dig(:content_facet_attributes, content_view_id).blank?
```

app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -48,28 +48,51 @@ def inherited_attributes
...
+          #     type key = first stage host creation, skip inheritance here
+          unless initialized || attributes.key?(:type)
+            attributes = apply_inherited_content_view(attributes)
+            attributes = apply_inherited_lifecycle_environment(attributes)
           end
           attributes
         end
 
+        def apply_inherited_content_view(attributes)
+          facet_cv_id = content_facet&.single_content_view&.id
```
ekohl, line 73: This line is raising an exception in the tests. I suspect you can't call `content_facet` on a host that doesn't exist yet or the host is created in the wrong way.

ekohl (commented):
Not a proper review, just some coding style things I noticed.

qcjames53:
Just made some big changes after a few meetings with the team. Will update PR description later today.

app/models/katello/concerns/host_managed_extensions.rb
qcjames53: Thank you for the coding style notes, I appreciate it. I'm going to resolve these since the source has changed so much in the last few days due to new information.

qcjames53:
@parthaa This should be good for review. Jeremy suggested you could get a head start while I work on the test cases.

ShimShtein (commented):
Looks good as far as facets/hostgroup relations go.

jeremylenz:
I wonder if this also addresses https://projects.theforeman.org/issues/35215

jeremylenz:
```
C: Layout/TrailingWhitespace: Trailing whitespace detected.
```

:see_no_evil:

parthaa (approved):
I tested inheritance of host groups (4 levels) and they worked as desired,
Nice work @qcjames53 and @jeremylenz !
Thank you @ShimShtein  for your guidance :)

APJ

PR #10843, opened 2024-01-11
Fixes #37014 - set correct Host owner during registration

nofaralfasi:
This commit modifies the prepare_host method to include the host owner. 
This change ensures that the host owner is set to the current user, providing consistency with the prepare_host method in the Foreman [repository](https://github.com/theforeman/foreman/blob/develop/app/controllers/api/v2/registration_controller.rb#L120).

### DIFF
--- a/app/controllers/katello/concerns/api/v2/registration_controller_extensions.rb
+++ b/app/controllers/katello/concerns/api/v2/registration_controller_extensions.rb
@@ -11,6 +11,7 @@ module Katello
             fail ActiveRecord::RecordNotFound, msg
           end
           @host.assign_attributes(host_params('host'))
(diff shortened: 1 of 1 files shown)


stejskalleos:
ok to test

stejskalleos (approved):
:green_apple: **LGTM**

It follows the same logic as the core:

https://github.com/theforeman/foreman/blob/develop/app/controllers/api/v2/registration_controller.rb#L120

stejskalleos:
@jeremylenz @ianballou Hi guys, can anyone of you please check the PR? I tested it and it worked fine, the change is small & safe.

chris1984 (approved):
ACK, LGTM

PR #10845, opened 2024-01-11
Refs #35713 - correctly call methods with kwargs

jeremylenz:

In https://github.com/Katello/katello/pull/10744 package_names_for_job_template was extracted into two new methods, `yum_names_for_job_template` and `deb_names_for_job_template`.

These new methods accept only keyword arguments, but are called only with regular non-kwargs.

This change fixes that and properly calls the methods with kwargs.

Install or remove a package via the new host details UI and make sure that the job template renders without errors.

### DIFF
--- a/app/models/katello/concerns/host_managed_extensions.rb
+++ b/app/models/katello/concerns/host_managed_extensions.rb
@@ -542,9 +542,9 @@ module Katello
 
       def package_names_for_job_template(action:, search:, versions: nil)
         if self.operatingsystem.family == 'Debian'
-          deb_names_for_job_template(action, search)
+          deb_names_for_job_template(action: action, search: search)
         else
-          yum_names_for_job_template(action, search, versions)
+          yum_names_for_job_template(action: action, search: search, versions: versions)
         end
       end
 


ekohl:
@ofedoren I think this would also be relevant to our Ruby 3 work.

ofedoren (approved):
Yeah, thanks for notifying.

LGTM, and I'd just merge it. There is no reason to hold it.

And now I'm really curious why it worked at the first place...

jeremylenz:
It was broken by https://github.com/Katello/katello/pull/10744 because that PR both added the methods and called them incorrectly, and I didn't catch it before merging.

ekohl:
@jeremylenz then shouldn't this be `Refs #35713` instead of opening a new issue?

jeremylenz:
I usually don't use "Refs" because I don't trust our automations (mostly Jira/Bugzilla) not to duplicate or mess up tickets. I can change it if you feel strongly.

ekohl:
Refs should be used if it's the same issue, within the same release. Otherwise you mess up automation, like cherry picks

jeremylenz:
updated Redmine and commit messages.

ianballou (approved):
ACK via https://github.com/Katello/katello/pull/10846#pullrequestreview-1828143114

PR #10848, opened 2024-01-12
Fixes #37050 - restrict module copying to source repository

ianballou:
Ensures that module streams are only copied from the source repository during CV publish.

Had to check that other content types aren't also affected.

From the BZ:

Steps to Reproduce:
1. Create a content view containing 1 modular and 1 non-modular repository (IE: 'Red Hat Enterprise Linux 8 for x86_64 - AppStream RPMs 8' & 'Red Hat Ansible Engine 2 for RHEL 8 x86_64 RPMs')
2. Create a content-view filter for module streams with the following settings:
- Include: true
3. Add a rule to this CV filter to:
- include all module streams
- original-module-streams: false
- Apply to all repositories in the CV: true
4. Publish a new version
5. See that module streams + modular RPMs got copied to the non-modular repo
6. Patch in my fix and publish again. See that the non-modular repo stays non-modular.

### DIFF
--- a/test/actions/pulp3/orchestration/copy_all_units_test.rb
+++ b/test/actions/pulp3/orchestration/copy_all_units_test.rb
@@ -568,6 +568,28 @@ module ::Actions::Pulp3
       assert_equal @repo.module_streams.pluck(:id).sort, @repo_clone.module_streams.pluck(:id).sort
     end
 
+    def test_module_streams_copied_from_actual_source_repo
+      # Try to copy a module stream from a source repository that doesn't have it.
+      filter = FactoryBot.build(:katello_content_view_module_stream_filter, :inclusion => true)
+      duck = @repo.module_streams.where(:name => "duck").first
+      FactoryBot.create(:katello_content_view_module_stream_filter_rule,
+                                   :filter => filter,
+                                   :module_stream => duck)
+      @repo.root.update!(:url => 'https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/')
+
+      sync_args = {:smart_proxy_id => @primary.id, :repo_id => @repo.id}
+      ForemanTasks.sync_task(::Actions::Pulp3::Orchestration::Repository::Sync, @repo, @primary, sync_args)
+
+      index_args = {:id => @repo.id}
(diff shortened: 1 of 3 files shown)


test/actions/pulp3/orchestration/copy_all_units_test.rb
```diff
@@ -568,6 +568,28 @@ def test_all_module_streams_copied_if_empty_modular_filter_rules
...
     end
 
+    def test_module_streams_copied_from_actual_source_repo
+      # Try to copy a module stream from a source repository that doesn't have it.
+      filter = FactoryBot.build(:katello_content_view_module_stream_filter, :inclusion => true)
+      duck = @repo.module_streams.where(:name => "duck").first
+      FactoryBot.create(:katello_content_view_module_stream_filter_rule,
+                                   :filter => filter,
+                                   :module_stream => duck)
+      @repo.root.update!(:url => 'https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/')
```
qcjames53, line 578: It looks like we're using this URL in a few places in Katello. I'm not too sure about referencing external URLs in our dev pipeline just in case this ever goes down. Since there is precedent for using `needed-errata`, I'm ok with it; I think we should at least consider doing away with this in the future.

qcjames53 (approved):
Apologies for getting this in so late; hopefully we can squeeze this one in before freeze.

I was able to verify the original bug was present when this fix was not applied. I verified that with this PR, a non-modular repository remains non-modular after a CV publish. This is confirmed to work as expected with both the UI and the API (hammer used) with both on-demand and immediate repos. I think the single test case still manages to cover the bases for this PR.

I have one concern about using the jlsherrill dummy url, but seeing as this is a resource referenced in several other test cases across Katello, I'm happy to ship this. I think we should consider making those calls local in the future for pipeline stability. Otherwise, everything looks great to me! Ack'ing

test/actions/pulp3/orchestration/copy_all_units_test.rb
sambible: we do this in QE as well, might be something I can bring up in a future JPL call and see what everyone else thinks.

PR #10850, opened 2024-01-12
Fixes #37051 - Add Katello Ansible default job templates

jeremylenz:

In the new host details UI, remote execution jobs use, by default, the 'xxx by search - Katello Script Default' job templates for package install / remove / upgrade, and errata apply. For those that use remote execution via Ansible, previously there was no way to use the web UI because there were no corresponding 'by search' Ansible job templates. This PR adds those.

For the package actions, I was originally looking at inheriting from "Package Action - Ansible Default." But that template only takes a single package name, and I needed multiple. So I'm just rendering the Ansible directly in Katello's template, instead of raising a PR to foreman_ansible. You can yell at me if you must.

1. Set up REX with Ansible (enable foreman_ansible and make sure your smart proxy shows the "Ansible" feature)
2. Check out the PR
3. `bundle exec rails runner "ForemanInternal.all.first.destroy"` to trigger seeds, then restart your Foreman server
4. Go to the Remote Execution Features page and click on each feature
5. Select the '`<feature name>` by search - Katello Ansible Default' as the default template for each feature. Example:
image

From the new host details Packages and Errata tabs, try

- package remove
- package install
- errata apply
- package upgrade
- package upgrade but the one where you have the little dropdown and you select which version to upgrade to (I didn't test this yet)

Make sure that templates render the Ansible correctly (check the 'Preview templates' tab on the REX job details). There's no need to actually communicate with the host, though this is nice to have.

### DIFF
--- a/app/views/foreman/job_templates/install_errata_by_search_query_-_katello_ansible_default.erb
+++ b/app/views/foreman/job_templates/install_errata_by_search_query_-_katello_ansible_default.erb
@@ -1,0 +1,24 @@
+<%#
+kind: job_template
+name: Install errata by search query - Katello Ansible Default
+job_category: Katello
+description_format: 'Install errata %{Errata search query}'
+feature: katello_errata_install_by_search
+provider_type: Ansible
+template_inputs:
+- name: Errata search query
+  description: Filter criteria for errata to be installed.
+  input_type: user
+  required: false
+%>
+
+<% advisory_ids = @host.advisory_ids(search: input("Errata search query")) -%>
+<% render_error(N_("No errata matching given search query")) if !input("Errata search query").blank? && advisory_ids.blank? -%>
+# RESOLVED_ERRATA_IDS=<%= advisory_ids.join(',') %>
+
+<% if @host.operatingsystem.family == 'Suse' -%>
+<%= render_template('Run Command - Ansible Default', :command => "zypper -n install -t patch #{advisory_ids.join(' ')}") %>
+<% else -%>
+<% advisories = advisory_ids.map { |e| "--advisory=#{e}" }.join(' ') -%>
+<%= render_template('Run Command - Ansible Default', :command => "yum -y update-minimal #{advisories}") %>
+<% end -%>
--- a/app/views/foreman/job_templates/install_packages_by_search_query_-_katello_ansible_default.erb
+++ b/app/views/foreman/job_templates/install_packages_by_search_query_-_katello_ansible_default.erb
@@ -1,0 +1,28 @@
+<%#
+kind: job_template
+name: Install packages by search query - Katello Ansible Default
+job_category: Katello
+description_format: 'Install package(s) %{Package search query}'
+feature: katello_package_install_by_search
+provider_type: Ansible
+template_inputs:
+- name: Package search query
+  description: Filter criteria for packages to be installed. IMPORTANT- If left blank, the job will attempt to install all possible packages.
+  input_type: user
+  required: false
+%>
+<% package_names = @host.package_names_for_job_template(
+  action: 'install',
+  search: input('Package search query')
+) -%>
+---
+- hosts: all
+  tasks:
+    - package:
+<% if package_names.empty? -%>
+        name: []
+<% else -%>
+        name: <%= indent(10) { to_yaml(package_names).gsub(/---/, "") } -%>
+        state: present
+<% end -%>
+
--- a/app/views/foreman/job_templates/remove_packages_by_search_query_-_katello_ansible_default.erb
+++ b/app/views/foreman/job_templates/remove_packages_by_search_query_-_katello_ansible_default.erb
@@ -1,0 +1,26 @@
+<%#
+kind: job_template
+name: Remove packages by search query - Katello Ansible Default
+job_category: Katello
+description_format: 'Remove package(s) %{Packages search query}'
+feature: katello_package_remove_by_search
+provider_type: Ansible
+template_inputs:
+- name: Packages search query
+  description: Filter criteria for packages to be removed.
+  input_type: user
+  required: false
+%>
+<% package_names = @host.package_names_for_job_template(
+  action: 'remove',
+  search: input('Packages search query')
+) -%>
+---
+- hosts: all
+  tasks:
+    - package:
+        name: 
+<% package_names.each do |package_name| -%>
+          - <%= package_name %>
+<% end -%>
+        state: absent
--- a/app/views/foreman/job_templates/update_packages_by_search_query_-_katello_ansible_default.erb
+++ b/app/views/foreman/job_templates/update_packages_by_search_query_-_katello_ansible_default.erb
@@ -1,0 +1,33 @@
+<%#
+kind: job_template
+name: Update packages by search query - Katello Ansible Default
+job_category: Katello
+description_format: 'Update package(s) %{Packages search query}'
+feature: katello_package_remove_by_search
+provider_type: Ansible
+template_inputs:
+- name: Packages search query
+  description: Filter criteria for packages to be updated.
+  input_type: user
+  required: false
+- name: Selected update versions
+  description: JSON string of selected package versions to be updated, in the format [ nvra ]. Leave blank to upgrade to latest available version.
+  input_type: user
+  required: false
+  value_type: plain
+%>
+<% package_names = @host.package_names_for_job_template(
+  action: 'update',
+  search: input('Packages search query'),
+  versions: input('Selected update versions')
+) -%>
+<% if package_names.empty? -%>
+<%= render_template('Run Command - Ansible Default', :command => "yum -y update") %>
+<% else -%>
+---
+- hosts: all
+  tasks:
+    - package:
+        name: <%= indent(10) { to_yaml(package_names).gsub(/---/, "") } -%>
+        state: latest
+<% end -%>


jeremylenz:
Marking draft for now because this also needs https://github.com/Katello/katello/pull/10845. When that's merged I'll rebase this without it. But feel free to review any time.

app/views/foreman/job_templates/install_errata_by_search_query_-_katello_ansible_default.erb
```diff
@@ -0,0 +1,24 @@
...
+  input_type: user
+  required: false
+%>
+
+<% advisory_ids = @host.advisory_ids(search: input("Errata search query")) -%>
+<% render_error(N_("No errata matching given search query")) if !input("Errata search query").blank? && advisory_ids.blank? -%>
+# RESOLVED_ERRATA_IDS=<%= advisory_ids.join(',') %>
+
+<% if @host.operatingsystem.family == 'Suse' -%>
+<%= render_template('Run Command - Ansible Default', :command => "zypper -n install -t patch #{advisory_ids.join(' ')}") %>
```
Et7f3, line 20:
```suggestion
<%= render_template('Run Command - Ansible Default', :command => "zypper -n install -t patch #{advisory_ids.join(' ')}") -%>
```

app/views/foreman/job_templates/install_errata_by_search_query_-_katello_ansible_default.erb
```diff
@@ -0,0 +1,24 @@
...
+<% else -%>
+<% advisories = advisory_ids.map { |e| "--advisory=#{e}" }.join(' ') -%>
+<%= render_template('Run Command - Ansible Default', :command => "yum -y update-minimal #{advisories}") %>
```
Et7f3, line 23:
```suggestion
<%= render_template('Run Command - Ansible Default', :command => "yum -y update-minimal #{advisories}") -%>
```

app/views/foreman/job_templates/install_packages_by_search_query_-_katello_ansible_default.erb
```diff
@@ -0,0 +1,26 @@
...
+) -%>
+---
+- hosts: all
+  tasks:
+    - package:
+        name: 
+<% package_names.each do |package_name| -%>
+          - <%= package_name %>
+<% end -%>
+        state: present
```
Et7f3, line 26:
```suggestion
<%= render_template('Package Action - Ansible Default', :state => 'present', :name => package_names) %>
```

At least it will be on par with the [Katello without Ansible](https://github.com/Katello/katello/blob/1e385f23bbc9d95c112f8a91d99f04d2bae7389c/app/views/foreman/job_templates/install_packages_by_search_query.erb)

app/views/foreman/job_templates/update_packages_by_search_query_-_katello_ansible_default.erb
```diff
@@ -0,0 +1,32 @@
...
+) -%>
+---
+- hosts: all
+  tasks:
+    - package:
+        name: 
+<% package_names.each do |package_name| -%>
+          - <%= package_name %>
+<% end -%>
+        state: present
```
Et7f3, line 32:
```suggestion
<%= render_template('Package Action - Ansible Default', :action => 'update', :package => package_names.join(' ')) %>
```
You have written `state: present`

app/views/foreman/job_templates/remove_packages_by_search_query_-_katello_ansible_default.erb
```diff
@@ -0,0 +1,26 @@
...
+) -%>
+---
+- hosts: all
+  tasks:
+    - package:
+        name: 
+<% package_names.each do |package_name| -%>
+          - <%= package_name %>
+<% end -%>
+        state: absent
```
Et7f3, line 26:
```suggestion
<%= render_template('Package Action - Ansible Default', :state => 'absent', :name => package_names) %>
```

app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -542,9 +542,9 @@ def rhel_eos_schedule_index
 
       def package_names_for_job_template(action:, search:, versions: nil)
         if self.operatingsystem.family == 'Debian'
-          deb_names_for_job_template(action, search)
+          deb_names_for_job_template(action: action, search: search)
         else
-          yum_names_for_job_template(action, search, versions)
+          yum_names_for_job_template(action: action, search: search, versions: versions)
```
Et7f3, line 547: why not

Et7f3 (commented):
Sorry for my suggestion (I wanted to upstream it but the PR process of opening a issue refrained me to upstream them: however we tested those in prod and they works ok) The only culprit is no package = all package for update. I don't know if we should force user to specify '*'. One thing we did is run this job on machine with query `applicable_rpms ~ yum` and we use package query `name = yum` however it resolved to empty query and then a `yum update` was fired (hopefully no machine were broken).

I don't think if we use the UI we would see the warning.

app/models/katello/concerns/host_managed_extensions.rb
Et7f3: Oh this seems to come from this PR https://github.com/Katello/katello/pull/10845

app/views/foreman/job_templates/remove_packages_by_search_query_-_katello_ansible_default.erb
jeremylenz: I didn't do this because I don't think that `:name => package_names` will render a correct YAML list.

app/views/foreman/job_templates/install_errata_by_search_query_-_katello_ansible_default.erb
jeremylenz: Why would the hyphen be needed at the end?

jeremylenz:
> The only culprit is no package = all package for update. I don't know if we should force user to specify '*'

The reason for the empty string is that if you run `dnf update` with no arguments, it updates all packages. I don't think it would understand `dnf update *`? (I could be wrong)

app/views/foreman/job_templates/install_errata_by_search_query_-_katello_ansible_default.erb
Et7f3: Just cosmetic.

app/views/foreman/job_templates/remove_packages_by_search_query_-_katello_ansible_default.erb
Et7f3: IIRC it generate a valid YAML list (the compact form) each package name separated by comma. Ansible understand it. I will recheck.

Et7f3:
> The reason for the empty string is that if you run `dnf update` with no arguments, it updates all packages. I don't think it would understand `dnf update *`? (I could be wrong)

if we use the ansible module yes: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/dnf_module.html other dnf command like provides,list accepts glob. I can test it but almost sure it supports it.

jeremylenz:
Rebased per https://github.com/Katello/katello/pull/10850#issuecomment-1889986708, marked ready for review

app/views/foreman/job_templates/remove_packages_by_search_query_-_katello_ansible_default.erb
Et7f3: Rendered output:
```yaml

# For Windows targets use the win_package module instead.
---
- hosts: all
  tasks:
    - package:
        name: ["libssh2", "libssh-devel"]
        state: absent
```

query: name = libssh-devel or name = libssh2
We have a trailling line empty but that can be changed by adding minus

I launched the jobs and confirm that Ansible is able to read this YAML.

app/views/foreman/job_templates/update_packages_by_search_query_-_katello_ansible_default.erb
Et7f3: There is really an error here: `state: present` -> `state: latest` or use `Package Action - Ansible Default` with `latest`
I don't put the rendered template because my test machine is outdated (long list) but we have the same layout than the others.

Scenario tested:
- update one package by clicking and it works correctly.
- update two packages by clicking and it works correctly.
- update with search query * and I see the full list and it works correctly
- update with blank search query => the lists is [] so a correct yaml. rc=0 and "Nothing to do"
- update with a query to a uptodate package `name ~ acl` same than blank query. In the proposed file: it would cause to trigger a `yum update` because the list is empty.

Et7f3 (changes requested):
I really think my suggestion is worthwhile at least for the last comment.

ianballou:
@ianballou 's testing tracker

Tested so far: 
Single package remove, install, update
Single errata apply
Install numerous packages

jeremylenz:
> There is really an error here: state: present -> state: latest or use Package Action - Ansible Default with latest

Sometimes, you need to be able to upgrade to a version that is not the latest. There will be multiple upgrade versions to choose from, and you could choose one that's not the latest. This feature was added in https://github.com/Katello/katello/pull/10263. You can see the community demo of this feature [here](https://youtu.be/SBzU8bmGAMI?si=iiZ_tmzpHc2iTBtT&t=177). My idea with using `present` instead of `latest` for package updates was so this would continue to work. The package `name` in the rendered template _should_ be the full package with version number, not just the base name. If you or @ianballou can test this scenario, please let me know since I haven't tested it yet.

ianballou:
Re: updating to an intermediate version:

```
---
- hosts: all
  tasks:
    - package:
        name: 
          - c-ares-1.17.1-5.el9_2.1.x86_64
          - crypto-policies-20230731-1.git94f0e2c.el9_3.1.noarch
        state: present
```

Here's an example. Neither of these are the latest.

ianballou:
@jeremylenz should clicking the hamburger button on a package to update and selecting "Upgrade via remote execution" use the ansible script if it's the default? In my case it used bash.

ianballou:
I'm hitting a weird bug that, so far, isn't reproducible with all RPMs:

image

The ansible script gets the latest version of the package, but the input is an earlier version to update.

jeremylenz:
> should clicking the hamburger button on a package to update and selecting "Upgrade via remote execution" use the ansible script if it's the default?

In my testing it did! Keep in mind you do have to set the default script for _all_ the features on the Features page (katello_package_install_by_search, katello_packages_update_by_search, katello_errata_install_by_search, etc.)

jeremylenz:
> The ansible script gets the latest version of the package, but the input is an earlier version to update.
> 
> Reproducible via a RHEL 9 host and RHEL 9 appstream/baseos

Can you also test the same with the defaults set back to Script? This def looks like a bug but I don't see how it could be ansible-related

ianballou:
> Can you also test the same with the defaults set back to Script? This def looks like a bug but I don't see how it could be ansible-related

Confirmed it fails with Script as well. [redmine.](https://projects.theforeman.org/issues/37072?issue_count=4&issue_position=2&next_issue_id=37071&prev_issue_id=37058)


> In my testing it did! Keep in mind you do have to set the default script for _all_ the features on the Features page (katello_package_install_by_search, katello_packages_update_by_search, katello_errata_install_by_search, etc.)

All good, I got confused by the switch to `Update Package - Katello Ansible Default` and thought it was script since I was bash, oops.

app/views/foreman/job_templates/install_errata_by_search_query_-_katello_ansible_default.erb
ianballou: https://www.puppet.com/docs/puppet/5.5/lang_template_erb.html#tags

The template renders the same, so maybe not needed?

```
# RESOLVED_ERRATA_IDS=RHBA-2023:0341

---
- hosts: all
  tasks:
    - shell:
        cmd: |
          yum -y update-minimal --advisory=RHBA-2023:0341
      register: out
    - debug: var=out
```

vs

```
# RESOLVED_ERRATA_IDS=RHBA-2023:1491

---
- hosts: all
  tasks:
    - shell:
        cmd: |
          yum -y update-minimal --advisory=RHBA-2023:1491
      register: out
    - debug: var=out
```

ianballou (approved):
This is working well for me, tested via the host details page.

Et7f3:
> @Et7f3 Sometimes, you need to be able to upgrade to a version that is not the latest. There will be multiple upgrade versions to choose from, and you could choose one that's not the latest.

(I just discovered randomly while showcasing foreman to colleagues)

latest c-ares can vary but latest c-ares-1.17.1-5.el9_2.1.x86_64 (the one rendered should be unique). I will test it soon but think latest/present in Ansible will honour the version we specify.

I don't recall the script version take care of this (command yum update/yum install)

app/views/foreman/job_templates/install_errata_by_search_query_-_katello_ansible_default.erb
Et7f3: IIRC it was just one end with \n while other have one \n less

Et7f3:
> I will test it soon but think latest/present in Ansible will honour the version we specify.

<details><summary>Test "Update Packages by search query - Katello Ansible Default" with latest and selected version not latest</summary>

On a machine with libcurl-7.76.1-23.el9.x86_64 upgradable to libcurl-7.76.1-23.el9_2.4.x86_64 or libcurl-7.76.1-26.el9_3.2.x86_64
I triggered `Update package(s) id ^ (18828)`
```yaml
# For Windows targets use the win_package module instead.
---
- hosts: all
  tasks:
    - package:
        name: ["libcurl-7.76.1-23.el9_2.4.x86_64"]
        state: latest
```

Packages search query 	id ^ (18828)
Selected update versions 	["libcurl-7.76.1-23.el9_2.4.x86_64"]
</details>

I changed my template to `present` like this PR:
<details><summary>Test "Update Packages by search query - Katello Ansible Default" with present and selected version not latest</summary>
On a machine with open-vm-tools-12.1.5-1.el9_2.2.x86_64 upgradable to open-vm-tools-12.1.5-1.el9_2.3.x86_64, open-vm-tools-12.2.5-3.el9_3.x86_64 or open-vm-tools-12.2.5-3.el9_3.2.x86_64
I triggered `Update package(s) id ^ (19232)`
```yaml
# For Windows targets use the win_package module instead.
---
- hosts: all
  tasks:
    - package:
        name: ["open-vm-tools-12.1.5-1.el9_2.3.x86_64"]
        state: present
```

Packages search query 	id ^ (19232)
Selected update versions 	["open-vm-tools-12.1.5-1.el9_2.3.x86_64"]
</details>

> I don't recall the script version take care of this (command yum update/yum install)

<details><summary>Test "Update Packages - Katello Ansible Default" selected version not latest</summary>

On a machine with glibc-2.34-60.el9.x86_64 upgradadable to glibc-2.34-60.el9_2.7.x86_64 or glibc-2.34-83.el9_3.7.x86_64
I triggered `Update package(s) glibc-2.34-60.el9_2.7.x86_64` that executed: `yum -y update glibc-2.34-60.el9_2.7.x86_64`
</details>

I am suprised that these template use raw ansible command when `Package action - Ansible Default` is present

Here is the result:
```
yum list installed glibc libcurl open-vm-tools | cat
Mise à jour des référentiels de gestion des abonnements.
Paquets installés
glibc.x86_64             2.34-60.el9_2.7       @rhel-9-for-x86_64-baseos-rpms   
libcurl.x86_64           7.76.1-23.el9_2.4     @rhel-9-for-x86_64-baseos-rpms   
open-vm-tools.x86_64     12.1.5-1.el9_2.3      @rhel-9-for-x86_64-appstream-rpms
```

Ok since `package_names_for_job_template` always fill the name+version present/latest does the same thing. I haven't tested on a debian host but think it also fill name+version. Ok so I have no strong argument about using `latest` instead of `present` (maybe just avoid future reader to think it is incorrect). And that is the reason the template `Update Packages - Katello Ansible Default` always works even if it begin with `yum update -y`.

One more argument about depending on `Package Action - Ansible Default` is that if one day we want to add a feature (like enable repos in pre_task, pass other options, ...). With the current implementation you have to repeat for each jobs instead of doing it once. So it is not DRY friendly.

jeremylenz:
> Ok so I have no strong argument about using `latest` instead of `present` (maybe just avoid future reader to think it is incorrect)

I think this is a good reason. I had originally used `present` instead of `latest` because I thought `latest` would not work with a non-latest version, but as your testing shows, `present` and `latest` are equivalent as long as the full package name with version is specified. So `latest` better signifies the intent. Since this is a package upgrade, I will change it to `latest`.

> One more argument about depending on Package Action - Ansible Default is that if one day we want to add a feature (like enable repos in pre_task, pass other options, ...). With the current implementation you have to repeat for each jobs instead of doing it once. So it is not DRY friendly.

Using Package Action - Ansible Default would require the ['short', 'form'] of writing lists. I did not even know this was possible before now. To be honest, I just prefer the long form of YAML lists (one item per line, beginning with `-`) which I see everywhere and I believe is easier to read. So if you don't mind I would like to keep it this way.

jeremylenz:
Changed the package update template to use `latest`.

Et7f3:
> So if you don't mind I would like to keep it this way

I don't mind short form/long form. Can you modify the `Package Action - Ansible Default` to print in long form then ?

You already have the code just put in the right file.

> since these templates do not change often

We use the `pre_script` and thus we would need to add x4

app/views/foreman/job_templates/install_packages_by_search_query_-_katello_ansible_default.erb
```diff
@@ -0,0 +1,26 @@
...
+  search: input('Package search query')
+) -%>
+---
+- hosts: all
+  tasks:
+    - package:
+        name: 
+<% package_names.each do |package_name| -%>
+          - <%= package_name %>
+<% end -%>
```
Et7f3, line 25:
```suggestion
<% if package_names.empty? -%>
        name: []
<% else -%>
        name:
<% package_names.each do |package_name| -%>
          - <%= package_name %>
<% end -%>
<% end -%>
```

Oh one thing about short form/long form for list: if list is empty (because we use the REX tab) [you need to use the short form](https://stackoverflow.com/a/5110341/7227940) otherwise the yaml won't be valid and we will get an syntax error.

So either print only short for the empty case or render_error that forbid it.

jeremylenz:
> I don't mind short form/long form. Can you modify the `Package Action - Ansible Default` to print in long form then ?

That would require a code change in https://github.com/theforeman/foreman_ansible which I am trying to avoid.

Et7f3:
> That would require a code change in https://github.com/theforeman/foreman_ansible which I am trying to avoid.

Oh didn't see it was another repo. Nevertheless this change is appreciable: will open a PR for it.

I tried to add argument about consistency but some code in Katello use it:
https://github.com/Katello/katello/blob/f281fc5bb15dbd35bc6e944d23bcdaac8ebbb2d8/app/views/foreman/job_templates/remove_package_-_katello_ansible_default.erb#L16
https://github.com/Katello/katello/blob/f281fc5bb15dbd35bc6e944d23bcdaac8ebbb2d8/app/views/foreman/job_templates/install_package_-_katello_ansible_default.erb#L16
but then found upgrade used another mechanism:
https://github.com/Katello/katello/blob/f281fc5bb15dbd35bc6e944d23bcdaac8ebbb2d8/app/views/foreman/job_templates/update_package_-_katello_ansible_default.erb#L16

jeremylenz:
```
---
- hosts: all
  tasks:
    - package:
        name: []
```

It seems this doesn't work. (dnf thinks there is nothing to do.) I will need to handle the "update all" case separately. Will update shortly.

Et7f3:
> ```
> ---
> - hosts: all
>   tasks:
>     - package:
>         name: []
> ```
> 
> It seems this doesn't work. (dnf thinks there is nothing to do.) I will need to handle the "update all" case separately. Will update shortly.

It is in the scenario I tested: https://github.com/Katello/katello/pull/10850#discussion_r1456643812 and I think it is correct behavior as it avoid a global `yum upgrade`

jeremylenz:
> It is in the scenario I tested: [#10850 (comment)](https://github.com/Katello/katello/pull/10850#discussion_r1456643812) and I think it is correct behavior as it avoid a global `yum upgrade`

When you select all upgradable packages in the web UI, the resulting command should be `dnf -y upgrade` (with no more arguments). This is the way the select all feature works - an empty search query matches all packages.

jeremylenz:
@ianballou This should be ready for another look, if you wish!

ianballou (approved):
Working well!

Re: update script updates all vs none for an empty package list, I agree that an empty list should update all because it mimics the `yum update` command.

jeremylenz:
Merging to ensure it gets into Katello 4.11.1.

Thanks for the feedback and reviews @ianballou @Et7f3 !

PR #10851, opened 2024-01-17
Fixes #37058 - Optimize query to identify orphaned content units

jpasqualetto:

Optimizes query used to identify orphaned content units

Query generated by the method was slow and inefficient. This attempts to offer an alternative with better performance.

Execution time of Remove Orphans task should improve drastically.

### DIFF
--- a/app/models/katello/concerns/pulp_database_unit.rb
+++ b/app/models/katello/concerns/pulp_database_unit.rb
@@ -169,7 +169,7 @@ module Katello
 
       def orphaned
         if many_repository_associations
-          where.not(:id => repository_association_class.select(unit_id_field))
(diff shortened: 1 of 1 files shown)


parthaa:
Copy pasting testing steps from 
https://github.com/Katello/katello/pull/10638#issue-1793844611


0.  Check `Katello::Rpm.count` in foreman-rake console.
1. Create a lot of orphaned content. You can create these by running below in foreman-rake console:
```
i = 1
while(i<50000)
f = Katello::Rpm.new(name: i, pulp_id: i)
f.save!
i +=1
end
```
2. Check `Katello::Rpm.count` in console.
3. Run `bundle exec rails katello:delete_orphaned_content` . Go to foreman task in UI and wait for it to complete.
4. Check `Katello::Rpm.count` again. The orphaned dummy rpm records should get cleaned up..

PR #10852, opened 2024-01-22
Fixes #37073 - only rename Katello settings if category column exists

evgeni:

make the migration work even when the `category` column was removed from Foreman core

none

in theory: install foreman, db:migrate, install katello, db:migrate, but there are other things that prevent this from working :(


evgeni:
[test katello]

ekohl (approved):

ehelms (approved):

PR #10854, opened 2024-01-22
Fixes #37075 - make RemoveKatelloFromNotificationName a noop

evgeni:

making an own migration a noop, so that Rails remains happy (the migration still "exists"), but the code does not error out

The change from this migration is irrelevant by now as
* We do not support upgrades from a 2016 Katello straight to master
* Any installation that needed this migration was updated by now and thus has the "real" migration applied

in theory: install foreman, db:migrate, install katello, db:migrate, but there are other things that prevent this from working :(


ehelms (approved):
Makes me wonder if we are due for a migration squish.

PR #10855, opened 2024-01-22
Fixes #37080 - content counts index goes over max size

ianballou:
Removes the index on smart proxies and content counts because the size can grow too large.

Since we read back the entire content count and write it in its entirety, an index likely doesn't help much anyway.

I think you can make indexes on items within the jsonb type, but I'm not sure that would help us here because, again, we always use the entire json blob. We don't have queries looking into the json blob (yet).

Run through the testing steps in the redmine https://projects.theforeman.org/issues/37080?issue_count=10&issue_position=5&next_issue_id=37077&prev_issue_id=37058


ianballou:
I tested this fix on the internal machine that found the error and can confirm that it works.

PR #10857, opened 2024-01-23
Fixes #36998 - declare development dependency on theforeman-rubocop

evgeni:

1. adds development dependency on `theforeman-rubocop ~> 0.0.6` (version picked to match todays Foreman develop branch)
2. fixes config and cops to pass `bundle exec rubocop` in a plain Katello checkout

Only confusion why the fixed cops weren't flagged before.

1. CI still passes
2. clean checkout of `katello.git` passes `bundle install && bundle exec rubocop`

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -11,6 +11,9 @@ inherit_from:
 Rails:
   Enabled: true
 
+Rails/ContentTag:
+  Enabled: false
+
 AllCops:
   Exclude:
     - db/migrate/20131014135042_katello_tables.rb
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -23,6 +23,7 @@ AllCops:
     - Gemfile
     - locale/action_names.rb
     - lib/monkeys/remove_hidden_distribution.rb
+    - vendor/**/*
   Include:
     - '**/*.rb'
     - app/views/**/*.rabl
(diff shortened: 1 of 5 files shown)


evgeni:
Feel free to squash those commits if you feel like, but I think the individual commits and their commit messages have valuable content that should be preserved :)

.rubocop.yml
```diff
@@ -11,6 +11,9 @@ inherit_from:
 Rails:
   Enabled: true
 
+Rails/ContentTag:
```
ekohl, line 14: What's the reason you disable https://docs.rubocop.org/rubocop-rails/cops_rails.html#railscontenttag here? Too many hits? If so, shouldn't it be in the TODO file?

evgeni: The commit message says:

The cop is buggy before rubocop-rails 2.12.0 [1][2], but to get that we'd need a newer theforeman-rubocop, which generates more offenses than I am willing to fix today. Disable the cop like it was disabled in foreman_ansible too [3].

[1] https://github.com/rubocop/rubocop-rails/issues/260
[2] https://github.com/rubocop/rubocop-rails/pull/526
[3] https://github.com/theforeman/foreman_ansible/pull/357

ekohl: Would it be a good idea to mention that in the file itself so future maintainers don't need to start the same archeological search?

evgeni: Can do that, sure.

jeremylenz (approved):
Confirmed `bundle exec rubocop` from Katello directory works (after I got my gems in order) and seems to run the same checks as `bundle exec rubocop ../katello` from foreman dir.

```
1920 files inspected, no offenses detected
```

PR #10865, opened 2024-01-26
Fixes #37109 - allow to use eager loading on latest_version_object

sbernhard:
Strips away 480 sql queries because of needs_publish? and especially the implementation of latest_version_object

Display 20 Content Views on UI (index page)

Total amount of queries:
[root@or v2]# grep SELECT /tmp/cv_without_opt1.log | wc -l
1234
[root@or v2]# grep SELECT /tmp/cv_opt1.log | wc -l
755

[root@or v2]# grep SELECT /tmp/cv_without_opt1.log | grep content_view_version    | wc -l
641
[root@or v2]# grep SELECT /tmp/cv_opt1.log | grep content_view_version    | wc -l
181

Details: https://pawelurbanek.com/rails-eager-ordered (and thank you so much for this page)

### DIFF
--- a/app/models/katello/content_view.rb
+++ b/app/models/katello/content_view.rb
@@ -18,6 +18,7 @@ module Katello
 
     has_many :content_view_versions, :class_name => "Katello::ContentViewVersion", :dependent => :destroy
     alias_method :versions, :content_view_versions
+    has_one :latest_version_object, -> { latest }, :class_name => "Katello::ContentViewVersion", :dependent => :destroy
     # Note the difference between content_view_components and component_composites both refer to
     # ContentViewComponent but mean different things.
     # content_view_components -> Topdown, given I am a composite CV get the associated components belonging to me
@@ -341,10 +342,6 @@ module Katello
       latest_version_object.try(:environments) || []
     end
 
-    def latest_version_object
-      self.versions.order('major DESC').order('minor DESC').first
-    end
-
     def last_task
       last_task_id = history.order(:created_at)&.last&.task_id
       last_task_id ? ForemanTasks::Task.find_by(id: last_task_id) : nil
@@ -610,6 +607,14 @@ module Katello
                                            :content_view => self,
                                            :components => components
       )
+
+      # TODO: If a controller creates a new version and then uses latest_version_object, the old data is displayed.
+      #       To prevent this, a 'reload' would currently be necessary, but this is not very performant.
+      #       However, this is currently not a problem because after your create_new_version there is no immediate
+      #       access to latest_version_object, but the ContentView object is first completely reloaded.
+      #
+      #       In Rails 7.1, individual connections can be reloaded:
+      #       https://www.shakacode.com/blog/rails-7-1-allows-resetting-singular-associations/
 
       update(:next_version => major.to_i + 1) unless major.to_i < next_version
       version
--- a/app/models/katello/content_view_version.rb
+++ b/app/models/katello/content_view_version.rb
@@ -66,6 +66,8 @@ module Katello
       joins(:repositories).includes(:content_view).merge(repositories).distinct
     end
 
+    scope :latest, -> { order('major DESC', 'minor DESC').limit(1) }
+
     scoped_search :on => :content_view_id, :only_explicit => true, :validator => ScopedSearch::Validators::INTEGER
     scoped_search :on => :major, :rename => :version, :complete_value => true, :ext_method => :find_by_version
     serialize :content_counts
--- a/test/models/content_view_component_test.rb
+++ b/test/models/content_view_component_test.rb
@@ -229,6 +229,7 @@ module Katello
                                                       :action => "publish")
       assert_equal @composite.needs_publish?, true
       @composite.create_new_version
+      @composite.reload
       cv_history.update!(katello_content_view_version_id: @composite.latest_version_object.id)
       assert_equal @composite.reload.needs_publish?, false
       version1 = create(:katello_content_view_version, :content_view => view1)
@@ -236,6 +237,7 @@ module Katello
                                          :content_view_version => version1, :latest => false)
       assert_equal @composite.reload.needs_publish?, true
       @composite.create_new_version
+      @composite.reload
       cv_history.update!(katello_content_view_version_id: @composite.latest_version_object.id)
       assert_equal @composite.reload.needs_publish?, false
       create(:katello_content_view_version, :content_view => view2)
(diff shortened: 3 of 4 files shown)


ekohl:
Test failures look related.

app/models/katello/content_view.rb
```diff
@@ -18,6 +18,7 @@ class ContentView < Katello::Model
 
     has_many :content_view_versions, :class_name => "Katello::ContentViewVersion", :dependent => :destroy
     alias_method :versions, :content_view_versions
+    has_one :latest_version_object, -> { latest }, :class_name => "Katello::ContentViewVersion"
```
jeremylenz, line 21: The `latest` method you have added to `content_view_version.rb` returns all the versions, sorted. But `latest_version_object` returns only one record, which surprised me. I guess Active Record is smart enough to see `has_one` and return only one?

app/models/katello/content_view_version.rb
```diff
@@ -66,6 +66,8 @@ class ContentViewVersion < Katello::Model
       joins(:repositories).includes(:content_view).merge(repositories).distinct
     end
 
+    scope :latest, -> { order('major DESC').order('minor DESC') }
```
jeremylenz, line 69: Since this returns a list of versions, maybe `by_latest` or `by_version_number` is a better name?

jeremylenz (commented):
Great find @sbernhard, thanks!

I tested this on my setup. With my ~20 CVs with 2 versions each, this improved my loading time from >3 sec every time to about 2.5 sec.

just a couple minor comments/questions

app/models/katello/content_view.rb
jeremylenz: I think I found the answer to my own question [in Active Record](https://github.com/rails/rails/blob/68eade83c87ae309191add6dfa4959d7d7e07464/activerecord/lib/active_record/associations/association_scope.rb#L30)

app/models/katello/content_view_version.rb
ekohl: https://guides.rubyonrails.org/active_record_querying.html#ordering says you can use a single order statement

sbernhard:
Katello tests are now green.

jeremylenz:
@sbernhard looking good, please squash or reword your commit to make 'Redmine issues' happy :)

test/models/content_view_component_test.rb
```diff
@@ -229,25 +229,29 @@ def test_needs_publish
                                                       :action => "publish")
       assert_equal @composite.needs_publish?, true
       @composite.create_new_version
+      @composite.reload
```
ekohl, line 232: Can `create_new_version` somehow invalidate the `latest_version_object` cache on the object?

sbernhard: - Rails load the `@composite` content view. During this, it already loads the latest_version_object relation "has_one -> latest"
- create_new_version creates a new version but the reference in the already loaded `@composite` is not updated.

ekohl: Yes, I got that. That's why I was asking if you can somehow flush that cached `latest_version_object` from Rails. Because I'm worried you might end up serving some outdated data somewhere. Looks like this is a Rails 7.1 feature: https://www.shakacode.com/blog/rails-7-1-allows-resetting-singular-associations/

So I think you can some something like:
```ruby
# Rails 7.1+
reset_latest_version_object if respond_to?(:reset_latest_version_object)
```

jeremylenz: I seem to be on Rails 6.1.7.6. (I didn't think we had upgraded to 7.1 already..)

Anyway, `reload` seems fine to me for the time being.

ekohl: > I seem to be on Rails 6.1.7.6. (I didn't think we had upgraded to 7.1 already..)

Yes, we're not there yet. I wish we were, but not yet. We're working on it.

> Anyway, `reload` seems fine to me for the time being.

My point was that this can also break things at runtime. Previously you could count on `latest_version_object` being up to date, but after this it may be stale. For example, if a controller creates the new object and then uses `latest_version_object` it will show the old data.

I don't know if that happens and if we have any coverage for it, but it's an API change that can be breaking.

sbernhard: run a `reload` inside create_new_version?

ekohl: There is a performance overhead with `reload`, because it actually loads an object which is an additional DB query that it doesn't need in some cases (which Rails 7.1 solves with a new method).

At this point I'm a bit uncertain, but it's a risk. When I take a look where it's called, it's probably a small risk.

sbernhard: Is it really necessary to reload it? AFAIK, if API / UI  is used, you run a action, like "Create a new Version" and then you run another action - which would be a completely new request and therefore all the structure will be reloaded automatically. I think, we are trying to solve a "unit-test-issue" which doesn't occur in production.

ekohl: No, I'm not saying you need to call `reload`. All I'm saying is that there is a cache invalidation that doesn't happen and that's a risk.

For example, if you have a REST API that is something like `POST /content_views/:id/versions` (so effectively an API controller) and the implementation is roughly:
```ruby
@content_view.create_new_version
{
  'id' => @content_view.id,
  'latest' => @content_view.latest_version_object,
}
```

Before this patch it would work. After this patch it would show incorrect data.

It would be very easy to do this, because the actual use is in RABL. Also note that there are other indirections, such as `latest_version` and `latest_version_id`. So these do show up here:

https://github.com/Katello/katello/blob/3dcae86ca41974534ec67771364ed6b312351e92/app/views/katello/api/v2/content_views/base.json.rabl#L8

So all I'm saying is that we must verify such code doesn't exist, or if it does, deal with it. If it does exist, there is a very good chance there are no tests for such a thing, so it doesn't show up as a failure. It's very easy to introduce a regression here.

sbernhard: So, should we:
1. do a reload in create_new_version or
2. run a reload after create_new_version in the test?

ekohl: I think that from reading the source of the actions it's probably safe to assume it's not a problem, but I'd still leave at least a TODO comment for Rails 7.1 that it can safely invalidate the cached entry.

sbernhard: ok

jeremylenz:
I think all we're waiting on here is a TODO comment, right?

sbernhard:
Broken React Test not related to this change.

jeremylenz (approved):
LGTM pending green tests. Failures are unrelated but let's wait just to be safe.

jeremylenz:
[test katello]

jeremylenz:
@sbernhard I think this may need another rebase due to the recent CI changes.

jeremylenz:
actually no, it ain't a required check! Merging

PR #10866, opened 2024-01-26
Fixes #37111 - use preloaded count

sbernhard:
Use size instead of count which just reuse the preloaded data if already available. See https://www.speedshop.co/2019/01/10/three-activerecord-mistakes.html

Before
[root@or v2]# grep SELECT /tmp/cv_opt2.log | grep COUNT | wc -l
47

Afterwards:
[root@or v2]# grep SELECT /tmp/cv_opt2.log | grep COUNT | wc -l
27

(20 CVs)

### DIFF
--- a/app/models/katello/content_view.rb
+++ b/app/models/katello/content_view.rb
@@ -188,7 +188,7 @@ module Katello
     end
 
     def content_host_count
-      hosts.count
+      hosts.size
     end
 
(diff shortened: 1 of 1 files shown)


sbernhard:
[test katello]

sjha4 (approved):

PR #10870, opened 2024-01-30
Fixes #37118 - Remove Virtual Guests in legacy UI for non hypervisors

chris1984:

* Virtual guests is returning an empty array when the host does not have any virtual guests. This is not present in the new UI, and I don't want to disrupt anything in the consumer.rb which could break something. Since the old UI will be going away soon and the issue is only present there, I just changed that section to hide that field/value if the array is empty.
* This does work correctly when `virt-who` reports in because then we do have values in the `virtual_guest` array, I did check this on my box and confirms the field shows up correctly.

* See above

* Register your devel box or a client to your Katello devel box
* Check out the old UI and see that is has a section called Virtual Guests (0)
* Check out PR
* Verify that section is gone

Screenshot before:
guestbefore

Screenshot after:
guestafter

### DIFF
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/details/views/content-host-info.html
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/content-hosts/details/views/content-host-info.html
@@ -51,8 +51,8 @@
           </button>
         </dd>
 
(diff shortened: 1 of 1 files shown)


jeremylenz (approved):
Tested with my webpack5 setup, works well

thanks @chris1984 ! ACK :+1:

PR #10873, opened 2024-01-31
Refs #37104 - fix sync_management

MariaAga:
* in `app/assets/javascripts/katello/sync_management/sync_management.js`
the real change is to move `KT.content =` and `KT.content_actions =` to be at the top, and remove `$(document).ready(function() {` as it will only run the script when the document is ready because of the `content_for`
* moving `katello.common.js` from `sync_management/index.js` to `sync_management/index.html.erb` to make sure its fully loaded in time
* Wrapping all the scripts in `content_for` so they wont run too early

### DIFF
--- a/app/views/katello/sync_management/index.html.erb
+++ b/app/views/katello/sync_management/index.html.erb
@@ -1,17 +1,18 @@
-<%= javascript 'katello/sync_management' %>
+<% content_for(:katello_javascripts) do -%>
+  <%= javascript_include_tag 'katello/sync_management' %>
+  <script>
+    localize({
+      "cancel": "<%= escape_javascript(_('Cancel')) %>",
+      "cancelled": "<%= escape_javascript(_('Sync Canceled')) %>",
+      "error": "<%= escape_javascript(_('Error')) %>",
+      "complete": "<%= escape_javascript(_('Sync complete.')) %>",
+      "no_start_time": "<%= escape_javascript(_('No start time currently available.')) %>"
+    });
 
-<%= javascript do %>
-  localize({
-    "cancel": "<%= escape_javascript(_('Cancel')) %>",
-    "cancelled": "<%= escape_javascript(_('Sync Canceled')) %>",
-    "error": "<%= escape_javascript(_('Error')) %>",
-    "complete": "<%= escape_javascript(_('Sync complete.')) %>",
-    "no_start_time": "<%= escape_javascript(_('No start time currently available.')) %>"
-  });
-
-    KT.repo_status = $.parseJSON('<%= escape_javascript(@repo_status.to_json.html_safe) %>');
-    KT.permissions = { "syncable" : <%= any_syncable? %> };
-<% end %>
+      KT.repo_status = $.parseJSON('<%= escape_javascript(@repo_status.to_json.html_safe) %>');
+      KT.permissions = { "syncable" : <%= any_syncable? %> };
+  </script>
+<% end -%>
 
 <% if Organization.current -%>
     <%= hidden_field_tag :organization_id, Organization.current.id%>
--- a/app/assets/javascripts/katello/common/katello.global.js
+++ b/app/assets/javascripts/katello/common/katello.global.js
@@ -1,5 +1,5 @@
 //Katello global object namespace that all others should be attached to
-var KT = {};
+var KT = KT ? KT : {};
 
 //i18n global variable
 var katelloI18n = {};
--- a/app/assets/javascripts/katello/sync_management/index.js
+++ b/app/assets/javascripts/katello/sync_management/index.js
@@ -1,3 +1,2 @@
 //= require "katello/jquery.treeTable"
-//= require "katello/common/katello.common.js"
 //= require "katello/sync_management/sync_management"
--- a/app/assets/javascripts/katello/sync_management/sync_management.js
+++ b/app/assets/javascripts/katello/sync_management/sync_management.js
@@ -1,343 +1,400 @@
-$(document).ready(function() {
-  // Setup initial state
+KT.content = (function () {
+  var draw_syncing = function (repo_id, progress, task_id) {
+      var element = $("#repo-" + repo_id).find(".result"),
+        cancelButton = $("<a/>").attr("class", "cancel_sync").text(katelloI18n.cancel),
+        progressBar = $("<a/>").attr("class", "progress").text(" ");
 
-    var ids = [];
-    $.each(KT.repo_status, function(repo_id, status){
-      if (status.is_running) {
-          ids.push(repo_id);
-          KT.content.draw_syncing(repo_id, status.progress.progress, status.sync_id);
+      if (task_id !== undefined) {
+        progressBar.attr("href", tfm.tools.foremanUrl("/foreman_tasks/tasks/" + task_id));
       }
-    });
-    KT.content.reset_products(KT.repo_status);
-    KT.content_actions.addSyncing(ids);
 
-    $('#select_all').click(KT.content.select_all);
-    $('#select_none').click(KT.content.select_none);
-    $('#collapse_all').click(KT.content.collapse_all);
-    $('#expand_all').click(KT.content.expand_all);
+      progress = progress ? progress : 0;
+      progressBar.progressbar({
+        value: progress,
+      });
+      element.find(".result-info").html("");
+      element.find(".result-info").append(progressBar);
+      element.find(".info-tipsy").addClass("hidden");
 
-    KT.content.showAll();
-    KT.content.select_repo();
+      if (KT.permissions.syncable) {
+        element.find(".result-info").append(cancelButton);
+      }
+    },
+    updateRepo = function (
+      repo_id,
+      starttime,
+      duration,
+      progress,
+      display_size,
+      packages,
+      size,
+      task_id
+    ) {
+      var repo = $("#repo-" + repo_id);
+      update_item(repo, starttime, duration, progress, display_size, packages, size, task_id);
+    },
+    finishRepo = function (repo_id, state, duration, raw_state, error_details, task_id) {
+      var element = $("#repo-" + repo_id);
+      var messages = [];
+      var url = tfm.tools.foremanUrl("/foreman_tasks/tasks/" + task_id);
+      state = '<a href="' + url + '">' + state + "</a>";
+      element.find(".result .result-info").html(state);
+      fadeUpdate(element.find(".duration"), duration);
 
-    $("#products_table").delegate(".cancel_sync", "click", function(){
-        var repo_id = $(this).parents("tr").attr("data-id");
-        KT.content_actions.cancelSync(repo_id, $(this));
-    });
+      if (raw_state === "error") {
+        element.find(".result .info-tipsy").removeClass("hidden");
 
-    $('#sync_product_form').bind("ajax:success",
-      function(evt, data, status, xhr){
-       var ids = [];
-       $.each(data, function(index, item){
-          ids.push(item.id);
-          KT.content.draw_syncing(item.id, 0, undefined);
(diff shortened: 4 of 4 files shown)


evgeni:
Hmm, trying the packit build, it now fails to load common.js:
```
GET https://centos8-stream-katello-nightly.tanso.example.com/assets/katello/common/index.js [HTTP/2 404  1ms]
```

app/views/katello/sync_management/index.html.erb
```diff
@@ -1,17 +1,19 @@
-<%= javascript 'katello/sync_management' %>
+<% content_for(:javascripts) do -%>
+  <script src="/assets/katello/common/index.js"></script>
+  <script src="/assets/katello/sync_management.js"></script>
```
evgeni, line 3:
```suggestion
  <%= javascript 'katello/common/index' %>
  <%= javascript 'katello/sync_management' %>
```
this makes things load the right files but in the wrong order:

```
<script type="text/javascript">
      const html = "  \n  \n  <script>\n    localize({\n      \"cancel\": \"Cancel\",\n      \"cancelled\": \"Sync Canceled\",\n      \"error\": \"Error\",\n      \"complete\": \"Sync complete.\",\n      \"no_start_time\": \"No start time currently available.\"\n    });\n\n      KT.repo_status = \$.parseJSON(\'{}\');\n      KT.permissions = { \"syncable\" : false };\n  <\/script>\n  <script src=\"/assets/katello/common/vendor-9f3c77c20de190d174986979ad7d0d5cee9362a9a280adabf643e57ee4259155.js\"><\/script>\n  <script src=\"/assets/katello/common-41006bb366e1429ef7397de66336e76035be453bc9bac600cd845b9571266232.js\"><\/script>\n  <script src=\"/assets/katello/common/index-41006bb366e1429ef7397de66336e76035be453bc9bac600cd845b9571266232.js\"><\/script><script src=\"/assets/katello/sync_management-db168612191a6dc850c8d5cd28375ee6af25d8fd011e62638f11f7fd903f5025.js\"><\/script>\n"
      load_dynamic_javascripts(html);
    </script>
```

and well, this gives the localize error again

evgeni:
```suggestion
  <%= javascript_include_tag 'katello/common/index' %>
  <%= javascript_include_tag 'katello/sync_management' %>
```

This works.
Now, wth is the difference between `javascript` and `javascript_include_tag`?!

evgeni: Ah, it's already wrapping it in `content_for`… Weird

https://github.com/theforeman/foreman/blob/262b5caceab9c3f71d0b80f189cc724389593fb0/app/helpers/layout_helper.rb#L98-L100

evgeni: Thinking more, this makes sense!

* Your original intent was "have the script load tags and the following embedded code in one block, so that it gets executed in the right order", but had the bug that it ignored asset pipeline paths.
* Using `javascript` creates a `content_for(:javascripts…)` block, which gets yielded into the `:javascripts` section individually (so the `=` of `<%=` didn't really do anything) and thus the order is fucked up
* Using `javascript_include_tag` creates script tags that honor the asset pipeline *AND* prints them directly where the code is, thus resulting in the correct order as one block, which then gets yielded into `:javascripts`.

app/views/overrides/activation_keys/_host_environment_select.html.erb
```diff
@@ -1,5 +1,7 @@
 <% f ||= form %>
-<%= javascript "katello/hosts/host_and_hostgroup_edit" %>
+<% content_for(:javascripts) do -%>
+  <%= javascript "katello/hosts/host_and_hostgroup_edit" %>
+<% end %>
```
evgeni, line 4: this is redundant, `<%= javacript …` already does `content_for(:javascripts)`: https://github.com/theforeman/foreman/blob/262b5caceab9c3f71d0b80f189cc724389593fb0/app/helpers/layout_helper.rb#L98-L100

app/views/overrides/activation_keys/_host_tab_pane.html.erb
```diff
@@ -1,5 +1,6 @@
-
-<%= javascript "katello/hosts/activation_key_edit" %>
+<% content_for(:javascripts) do -%>
+  <%= javascript "katello/hosts/activation_key_edit" %>
+<% end %>
```
evgeni, line 4: this is redundant, `<%= javacript …` already does `content_for(:javascripts)`: https://github.com/theforeman/foreman/blob/262b5caceab9c3f71d0b80f189cc724389593fb0/app/helpers/layout_helper.rb#L98-L100

app/views/overrides/organizations/_edit_override.html.erb
```diff
@@ -1,7 +1,7 @@
 <% if @taxonomy.is_a?(Organization) %>
-
-  <%= javascript_include_tag 'katello/organizations/download_certificate' %>
-
+  <% content_for(:javascripts) do -%>
+    <%= javascript_include_tag 'katello/organizations/download_certificate' %>
+  <% end %>
```
evgeni, line 4: should we standardize on either
```
<%= javascript … %>
```
OR
```
  <% content_for(:javascripts) do -%>
    <%= javascript_include_tag … %>
  <% end %>
```

using both seems odd?

app/views/katello/sync_management/index.html.erb
evgeni:
```
>>> KT
<<< Object { utils: $(n), common: {…}, widget: {}, getData: getData(n), orgswitcher: {…}, menu: {…}, object: {…} }

>>> load_dynamic_javascripts("<script src=\"/assets/katello/sync_management-db168612191a6dc850c8d5cd28375ee6af25d8fd011e62638f11f7fd903f5025.js\"><\/script>")
<<< undefined

>>> KT
<<< Object { utils: $(n), common: {…}, widget: {}, getData: getData(n), orgswitcher: {…}, menu: {…}, object: {…}, content: {…}, content_actions: {…} }
```

So reloading that sync management script works. WTH.

app/views/overrides/organizations/_edit_override.html.erb
MariaAga: I rather use 
```
  <% content_for(:javascripts) do -%>
    <%= javascript_include_tag … %>
  <% end %>
  ```
  just because when I see `<%= javascript … %>` I assumed it was something default from ruby and not a custom thing

File: app/views/katello/sync_management/index.html.erb (hunk 1 above)
Comment (evgeni) [reply app/views/katello/sync_management/index.html.erb:3]: Could it be something re-initializes `KT`, thus overwriting what this code did?

File: app/views/katello/sync_management/index.html.erb [hunk 5]
```diff
@@ -1,17 +1,19 @@
-<%= javascript 'katello/sync_management' %>
+<% content_for(:javascripts) do -%>
```
evgeni, line 1:
```suggestion
<% content_for(:katello_javascripts) do -%>
```

THIS fixes it.

the reset happens here:
https://github.com/Katello/katello/blob/136a73e8c007d4ed5969adfb6411dc2c68e9880d/app/views/katello/layouts/katello.html.erb#L9-L13

evgeni:
```
<% content_for(:katello_javascripts) do -%>
  <%= javascript_include_tag 'katello/sync_management' %>
  <script>
    localize({
      "cancel": "<%= escape_javascript(_('Cancel')) %>",
      "cancelled": "<%= escape_javascript(_('Sync Canceled')) %>",
      "error": "<%= escape_javascript(_('Error')) %>",
      "complete": "<%= escape_javascript(_('Sync complete.')) %>",
      "no_start_time": "<%= escape_javascript(_('No start time currently available.')) %>"
    });

      KT.repo_status = $.parseJSON('<%= escape_javascript(@repo_status.to_json.html_safe) %>');
      KT.permissions = { "syncable" : <%= any_syncable? %> };
  </script>
<% end -%>
```

this is how it looks in full on my box now

app/views/overrides/organizations/_edit_override.html.erb
evgeni: I do agree that ths makes it more explicit.

app/views/katello/sync_management/index.html.erb
MariaAga: I still get "Uncaught TypeError: Cannot read properties of undefined (reading 'select_repo')" on select none with this?

evgeni: Works for me (on a prod install, don't have a devel katello right now)

evgeni: This is the code generated for loading (I added line breaks for readability):
```
<script type="text/javascript">
      const html = "
        <script src=\"/assets/katello/common/vendor-9f3c77c20de190d174986979ad7d0d5cee9362a9a280adabf643e57ee4259155.js\"><\/script>\n
        <script src=\"/assets/katello/common-41006bb366e1429ef7397de66336e76035be453bc9bac600cd845b9571266232.js\"><\/script>\n
        <script src=\"/assets/katello/sync_management-db168612191a6dc850c8d5cd28375ee6af25d8fd011e62638f11f7fd903f5025.js\"><\/script>\n
        <script>\n    localize({\n      \"cancel\": \"Cancel\",\n      \"cancelled\": \"Sync Canceled\",\n      \"error\": \"Error\",\n      \"complete\": \"Sync complete.\",\n      \"no_start_time\": \"No start time currently available.\"\n    });\n\n      KT.repo_status = \$.parseJSON(\'{\\\"1\\\":{\\\"id\\\":1,\\\"product_id\\\":1,\\\"progress\\\":{},\\\"state\\\":\\\"Never Synced\\\",\\\"raw_state\\\":\\\"never_synced\\\"}}\');\n      KT.permissions = { \"syncable\" : true };\n  <\/script>\n\n"
      load_dynamic_javascripts(html);
    </script>
```

evgeni:
FWIW, I'd omit the "[Refs #37104 - wrap scripts in content_for(:javascripts)](https://github.com/Katello/katello/pull/10873/commits/a1fd6f44ee884d84762118a965f29aed2d3941a8)" commit from this PR and only have the "fix sync_management" changes

(but I am neither author, nor reviewer, so just stating my thoughts ;) )

app/views/katello/sync_management/index.html.erb
MariaAga: I put it in the wrong file :facepalm: 
but setting `var KT = KT ? KT : {};` instead of `var KT = {};`  in `app/assets/javascripts/katello/common/katello.global.js` also fixed it :upside_down_face:

evgeni: Yeah… but that still means that the code gets executed twice, which seems wrong :)

evgeni: I mean… It's probably still a good idea to have such a safety guard, but it should not be the "real" fix

MariaAga: which code gets executed twice?

evgeni: `katello/common` gets executed twice if we use `content_for(:javascripts)` (and not `:katello_javascripts`), as that generates:
- katello/common (from sync_management/index.html)
- katello/sync_management
- katello/common (from layouts/katello)

MariaAga: > I mean… It's probably still a good idea to have such a safety guard, but it should not be the "real" fix

Yeah, I thought you meant adding it will cause the double execution.
Added it to the pr and used your solution, thanks!

evgeni (approved):

jeremylenz (approved):

jeremylenz:
merging based on @evgeni's review

Thanks @MariaAga !

PR #10874, opened 2024-01-31
i18n - pulling from tx

qcjames53:
Updated translations from Transifex. Please ensure that `engines/bastion_katello/app/assets/javascripts/bastion_katello/i18n/bastion_katello.pot` is correct; I needed to add it manually.


locale/action_names.rb
```diff
@@ -1,33 +1,36 @@
...
-_("Check for long running tasks")
 _("Combined Profile Update")
 _("Copy version units to library")
 _("Create Alternate Content Source")
 _("Create Export History")
 _("Create Import History")
 _("Create Syncable Export History")
 _("Create")
+_("Create")
+_("Create")
```
wbclark, line 15: @ianballou any idea what's going on with these duplicates?

locale/action_names.rb
```diff
@@ -1,33 +1,36 @@
...
 _("Delete Activation Key")
 _("Delete Lifecycle Environment")
 _("Delete Product")
 _("Delete")
-_("Deliver notifications about long running tasks")
+_("Delete")
+_("Delete")
```
wbclark, line 21: and here

locale/action_names.rb
```diff
@@ -78,6 +78,12 @@
 _("Update redhat repository")
 _("Update release version for host")
 _("Update")
+_("Update")
+_("Update")
+_("Update")
+_("Update")
+_("Update")
```
wbclark, line 85: same

locale/action_names.rb
qcjames53: When re-running the translation pulls from Transifex, we encountered the exact same duplicate strings in the exact same locations. See https://github.com/Katello/katello/pull/10877

qcjames53: This led William and I to believe these duplicates are somehow intentional. We're going to resolve these issues.

wbclark (approved):
we don't know why the dupes appeared but they are appearing consistently when we re-run the procedure and it seems they're not breaking anything

PR #10875, opened 2024-01-31
Fixes #37131 - Migrate all organizations to SCA

jeremylenz:

This PR is the first in the "SCA-Only" feature, and introduces the following Candlepin-related changes: 

* Adds an upgrade rake task to migrate all organizations to SCA on upgrade.
* Adds a dry run option to the rake task. This can be used later if needed in a possible foreman-maintain pre-upgrade check.
* Updates the _content access mode list_ for all orgs so that entitlement mode is not allowed. 

You will recall definitions for two terms:

_Content access mode_: Either "entitlement" (non-SCA) or "org_environment" (SCA). The subscription management strategy used by a particular org.
_Content access mode list_: The list of allowed _content access modes_. Before this change, it was `'entitlement','org_environment'`. Now, it will simply be `'org_environment'`.

`::Katello::Resources::Candlepin::Owner#update` already edits an organization's _content access mode list_ on every update. I have simply changed it to remove `'entitlement'`:

```rb
:contentAccessModeList => 'org_environment'
```

And the rake task in this PR updates all orgs like so:

```rb
::Katello::Resources::Candlepin::Owner.update(org.label, contentAccessMode: 'org_environment')
```

So, the effect will be that after the rake task is run, all organizations (a) use SCA; and (b) can _only_ use SCA.

1. Have at least one org which does not use SCA
2. Check out this PR
3. Try to edit a description (or other field) of your org - You'll get an error about content access mode list
4. Try to create a new organization with SCA off - You'll get an error about content access mode list
5. Run the rake task:

```
rake katello:upgrades:4.12:update_content_access_modes[dry_run]
```

You should see something like

```
----------------------------------------
1 organizations would be migrated to Simple Content Access on upgrade
----------------------------------------
```

Now commit the changes:
```
rake katello:upgrades:4.12:update_content_access_modes
```

It should now say

```
----------------------------------------
Set content access mode for 1 organizations
----------------------------------------
```

The web UI etc. should now also reflect this, and nothing should be broken.

You can also verify in Rails console that both contentAccessMode and contentAccessModeList have the correct values (neither should say 'entitlement'):

```
[7] pry(main)> org.owner_details['contentAccessModeList']
=> "org_environment"
[8] pry(main)> org.owner_details['contentAccessMode']
=> "org_environment"
```


wbclark (commented):
Before migrating:
`Katello::Resources::Candlepin::Owner: 400 Bad Request {"displayMessage":"Content access mode is not present in the content access mode list for this org: entitlement","requestUuid":"2caee74b-c128-4a46-a022-a8ce47b59f0d"} (PUT /candlepin/owners/test)`

Dry run:
```
Checking content access mode for test
----------------------------------------
1 organizations would be migrated to Simple Content Access on upgrade
----------------------------------------
```

I also noticed that with a manifest uploaded to this organization, this stage spit out debug logs to the console which included, among other things, the private key under `upstreamConsumer`. This is... probably fine but maybe users won't like it?

Once more, with conviction:

```
----------------------------------------
Set content access mode for 1 organizations
----------------------------------------
```

(As before, I could see sensitive details about my manifest printed to the console)

After migrating the org to SCA mode, I tried to edit my org again and got:

```
ActiveRecord::RecordInvalid in OrganizationsController#update
Validation failed: Taxonomy has already been taken
```


After a restart, that issue went away and I was again able to edit org description, etc.

wbclark:
> I also noticed that with a manifest uploaded to this organization, this stage spit out debug logs to the console which included, among other things, the private key under `upstreamConsumer`. This is... probably fine but maybe users won't like it?

We chatted about this offline. Since these are debug logs, they shouldn't spook users running in production mode. :)

> After a restart, that issue went away and I was again able to edit org description, etc.

Regardless, the actual upgrades will always involve a restart so this is fine.

wbclark:
LGTM pending fix of

```
[2024-01-31T21:54:28.778Z] An HTTP request has been made that VCR does not know how to handle:
[2024-01-31T21:54:28.778Z]   POST https://localhost:8443/candlepin/owners/
```

jeremylenz:
Tests :green_apple: :green_apple: now :+1:

wbclark (approved):

app/lib/katello/resources/candlepin/owner.rb
```diff
@@ -18,7 +18,7 @@ def create(key, description, content_access_mode: 'org_environment')
               :displayName => description,
               :contentPrefix => "/#{key}/$env",
               :contentAccessMode => content_access_mode,
-              :contentAccessModeList => ['entitlement', 'org_environment'].join(',')
+              :contentAccessModeList => 'org_environment'
```
wbclark, line 21: (non-blocking) since the :contentAccessModeList is now just a single item, should we create a card to simplify and refactor in the future? or does this construction come from candlepin and changes would need to be made on their first?

let me know what you think @jeremylenz

jeremylenz: Candlepin takes a comma-separated list, which was previously `"entitlement,org_environment"` but is now simply `"org_environment"`. So I don't see any refactoring that could be done on our end, unless I'm missing something?

wbclark: No, that answered my question. Thanks!

PR #10876, opened 2024-02-01
try to fix sync_management page

evgeni:

I need a packit build


evgeni:
nope

PR #10861, opened 2024-01-25
Fixes #37145 - Ruby 3 support

ofedoren:
#### What are the changes introduced in this pull request?

Updates for kwargs usage across the code base to comply with Ruby 3.0.

Please, review carefully, since there might be a lot of unnecessary changes as well as some breaking changes (such as changing signatures for action planning methods).

#### Considerations taken when implementing this change?

Depends on
 - https://github.com/Katello/katello/pull/10780
 - https://github.com/theforeman/foreman/pull/9989
 - https://github.com/theforeman/foreman-tasks/pull/739 (including Dynflow changes)

#### What are the testing steps for this pull request?

Run CI and hope for the best.

PR #10878, opened 2024-02-01
Fixes #37137 - Use insert_all when creating available module streams

jeremylenz:

During a profile upload, a host's module streams are updated in Katello's database.

In https://github.com/Katello/katello/pull/10412 we changed to using `create_or_find_by!` to create those records. This solved an issue where concurrent host registrations would sometimes fail due to a PostgreSQL error:

```
ERROR:  duplicate key value violates unique constraint "katello_available_module_streams_name_stream_context"
```

However, these errors are still inserted into postgres logs. They are harmless, but log files can fill up quickly. This is because there will be one log for _every_ module stream creation attempt, for every host with module streams, _not just at registration_ but also during every profile upload. This results in log files that can quickly build to 200000 lines with just those error logs.

With this change, we're now using `insert_all` to insert all of the records in a single SQL query, which should avoid the race condition.

This required some major refactoring, because

1. `import_module_streams` takes module stream data in a list of Hashes, but those don't include Katello's ID. (There's no way they could, as this info is coming from sub-man/dnf.)
2. In our previous implementation, we relied on the `id`s returned from `create_or_find_by` etc. to pass on to the next method, `sync_available_module_stream_associations`. `sync_available_module_stream_associations` requires a Hash indexed by database ID. 
3. Previously, the data passed to `sync_available_module_stream_associations` was _all_ of the host's available module streams, indexed by ID, _including_ the ones we just newly added. Thus, `new_available_module_streams` was a bit of a misnomer. I did some refactoring to recreate this format of the indexed hash.

1. `tail -f /var/lib/pgsql/data/log/postgresql-Thu.log` (replace Thu with whatever day it is today)
6. Register a host (without this patch) - Notice that the file fills up with error logs

```
2024-02-01 21:57:16 UTC STATEMENT:  INSERT INTO "katello_available_module_streams" ("name", "stream", "context") VALUES ($1, $2, $3) RETURNING "id"
2024-02-01 21:57:16 UTC ERROR:  duplicate key value violates unique constraint "katello_available_module_streams_name_stream_context"
2024-02-01 21:57:16 UTC DETAIL:  Key (name, stream, context)=(virt, rhel, 30b713e6) already exists.
```

You can also check how many logs there are:

```
grep 'duplicate key value violates unique constraint "katello_available_module_streams_name_stream_context"' /var/lib/pgsql/data/log/postgresql-Thu.log | wc -l
---
227950
---
```

7. Enable or disable a module stream - Notice that more error logs are created.
8. Check out this PR
9. Enable or disable a module stream - No new logs should be created.
10. Register a host - No new logs should be created.

### DIFF
--- a/app/models/katello/concerns/host_managed_extensions.rb
+++ b/app/models/katello/concerns/host_managed_extensions.rb
@@ -326,11 +326,21 @@ module Katello
       end
 
       def import_module_streams(module_streams)
+        # create_or_find_by avoids race conditions during concurrent registrations but clogs postgres logs with harmless errors.
+        # So we'll use create_or_find_by! during registration and first_or_create! otherwise.
+        registered_time = subscription_facet&.registered_at
+        use_create_or_find_by = registered_time.nil? || registered_time > 1.minute.ago
         streams = {}
         module_streams.each do |module_stream|
-          stream = AvailableModuleStream.create_or_find_by!(name: module_stream["name"],
-                                               context: module_stream["context"],
-                                               stream: module_stream["stream"])
+          if use_create_or_find_by
+            stream = AvailableModuleStream.create_or_find_by!(name: module_stream["name"],
+                                                context: module_stream["context"],
+                                                stream: module_stream["stream"])
+          else
+            stream = AvailableModuleStream.where(name: module_stream["name"],
+            context: module_stream["context"],
+            stream: module_stream["stream"]).first_or_create!
+          end
           streams[stream.id] = module_stream
         end
         sync_available_module_stream_associations(streams)
--- a/app/models/katello/concerns/host_managed_extensions.rb
+++ b/app/models/katello/concerns/host_managed_extensions.rb
@@ -342,12 +342,13 @@ module Katello
             context: module_stream["context"]
           }
         end
-        AvailableModuleStream.insert_all(
-          streams,
-          unique_by: %w[name stream context],
-          returning: %w[id name stream context]
-        )
-
+        if streams.any?
+          AvailableModuleStream.insert_all(
+            streams,
+            unique_by: %w[name stream context],
+            returning: %w[id name stream context]
+          )
+        end
         indexed_module_streams = module_streams.index_by do |module_stream|
           available_module_stream_id_from(
                   name: module_stream["name"],
@@ -355,7 +356,6 @@ module Katello
                   context: module_stream["context"]
                 )
         end
-
         sync_available_module_stream_associations(indexed_module_streams)
       end
 
--- a/app/models/katello/concerns/host_managed_extensions.rb
+++ b/app/models/katello/concerns/host_managed_extensions.rb
@@ -326,15 +326,15 @@ module Katello
       end
 
       def available_module_stream_id_from(name:, stream:, context:)
-        AvailableModuleStream.find_by!(name: name, stream: stream, context: context).id
-      rescue ActiveRecord::RecordNotFound
-        Rails.logger.warn("Module stream not found: name: #{name}, stream: #{stream}, context: #{context}")
-        nil
+        @indexed_available_module_streams ||= Katello::AvailableModuleStream.all.index_by do |available_module_stream|
+          "#{available_module_stream.name}-#{available_module_stream.stream}-#{available_module_stream.context}"
+        end
+        @indexed_available_module_streams["#{name}-#{stream}-#{context}"]&.id
       end
 
       def import_module_streams(module_streams)
-        Rails.logger.debug "INSERT_ALL-------------------------------------"
-        Rails.logger.debug module_streams.first(3)
+        # module_streams looks like this
+        # {"name"=>"389-ds", "stream"=>"1.4", "version"=>"8030020201203210520", "context"=>"e114a9e7", "arch"=>"x86_64", "profiles"=>[], "installed_profiles"=>[], "status"=>"default", "active"=>false}
         streams = module_streams.map do |module_stream|
           {
             name: module_stream["name"],
@@ -342,15 +342,11 @@ module Katello
             context: module_stream["context"]
           }
         end
-        newly_added_records = AvailableModuleStream.insert_all(
+        AvailableModuleStream.insert_all(
           streams,
           unique_by: %w[name stream context],
           returning: %w[id name stream context]
         )
-        Rails.logger.debug "newly_added_records-------------------------------------"
-        Rails.logger.debug newly_added_records.to_a
-        # module_streams looks like this
-        # {"name"=>"389-ds", "stream"=>"1.4", "version"=>"8030020201203210520", "context"=>"e114a9e7", "arch"=>"x86_64", "profiles"=>[], "installed_profiles"=>[], "status"=>"default", "active"=>false}
 
         indexed_module_streams = module_streams.index_by do |module_stream|
           available_module_stream_id_from(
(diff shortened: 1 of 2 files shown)


jeremylenz:
fixed rubocop

chris1984:
Your PR is having the same test failures as mine is

jeremylenz:
> Your PR is having the same test failures as mine is

Yeah, apparently https://github.com/Katello/katello/pull/10861 has the fixes. We may have to ignore those for a bit.

jeremylenz:
[test katello]

sbernhard:
Ready?

jeremylenz:
Yes, this should be ready for review again :+1:

app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -348,11 +348,21 @@ def import_enabled_repositories(repos)
       end
 
       def import_module_streams(module_streams)
+        # create_or_find_by avoids race conditions during concurrent registrations but clogs postgres logs with harmless errors.
```
sbernhard, line 351: instead of two different implementation, would it be possible to catch the "postgres logs with harmless errors" and handle this accordingly?

jeremylenz: No, that's the way the race condition is avoided - by letting the Postgres constraint handle it rather than ruby/rails.

sbernhard: ok. make sense.

app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -348,11 +348,21 @@ def import_enabled_repositories(repos)
...
+        # So we'll use create_or_find_by! during registration and first_or_create! otherwise.
+        registered_time = subscription_facet&.registered_at
+        use_create_or_find_by = registered_time.nil? || registered_time > 1.minute.ago
         streams = {}
         module_streams.each do |module_stream|
```
ianballou, line 356: I know this code is old, but I wonder if now would be a good time to switch to `insert_all` or `upsert_all`. It feels odd to do so many creations one postgres insert at a time in a Ruby loop.

Bulk insertion lets you create all entities in a single Postgres insert call. It takes into account duplicates too from what I remember. You can pass in what the records should be `unique_by` (in this case name, stream, and context).

There might be a catch, but I think it's worth a thought.

https://apidock.com/rails/v6.0.0/ActiveRecord/Persistence/ClassMethods/insert_all
https://apidock.com/rails/v6.0.0/ActiveRecord/Persistence/ClassMethods/upsert_all

We use them in a couple places: 

https://github.com/search?q=repo%3AKatello%2Fkatello%20insert_all&type=code
https://github.com/search?q=repo%3AKatello%2Fkatello+upsert_all&type=code

ianballou (commented):
Why won't we hit `PG::UniqueViolation` errors during concurrent package profile uploads? Is there a difference in concurrency between 10 systems registering at the same time and 10 systems `dnf updating` at the same time?

jeremylenz:
We haven't had any complaints of that, for whatever reason. It could be because most package profile uploads are from rhsmcertd checkins which are staggered automatically with random delay intervals to prevent too many concurrent requests, whereas registrations can happen at any time.

app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -326,26 +326,77 @@ def import_enabled_repositories(repos)
...
-      end
-
-      def sync_available_module_stream_associations(new_available_module_streams)
+        Rails.logger.debug "INSERT_ALL-------------------------------------"
+        Rails.logger.debug module_streams.first(3)
+        streams = module_streams.map do |module_stream|
+          {
+            name: module_stream["name"],
+            stream: module_stream["stream"],
+            context: module_stream["context"]
```
ianballou, line 335: AvailableModuleStreams also have `profiles`:

```
[6] pry(main)> ::Katello::AvailableModuleStream
=> Katello::AvailableModuleStream(id: integer, name: string, stream: string, context: string, profiles: )
```

app/models/katello/concerns/host_managed_extensions.rb
```diff
@@ -326,26 +326,77 @@ def import_enabled_repositories(repos)
...
+        untransformed_keys = indexed_module_streams.keys
+        Rails.logger.debug "UNTRANSFORMED_KEYS-------------------------------------"
+        Rails.logger.debug untransformed_keys
+        # now we transform the keys of indexed_module_streams to the ids from indexed_result
+        # { 1 => {"name"=>"389-ds", "stream"=>"1.4", "version"=>"8030020201203210520", "context"=>"e114a9e7", "arch"=>"x86_64", "profiles"=>[], "installed_profiles"=>[], "status"=>"default", "active"=>false} }
+        indexed_module_streams.transform_keys! { |k| indexed_result[k]&.[]("id") }
+
+        Rails.logger.debug "TRANSFORMED KEYS-------------------------------------"
+        Rails.logger.debug indexed_module_streams.keys
+        # remove the keys that were not transformed
```
ianballou, line 369: Maybe I'm missing something, but what do these hash transformations get us that `result` doesn't already have? The only module streams that will have DB IDs will be the ones returned by `upsert_all`, which are all of the records in `result`.

As I understand the the code here, `result` and `module_streams` get matched up by the hash of their (name, stream, context) set.  Then, the hashes get replaced by DB IDs if they exist. However, `result` already has these database IDs, and `indexed_module_streams.except!(untransformed_keys)` throws out all of the ones that don't have database IDs.

 Doesn't this just leave us with what is in `result` but in a slightly different format?

I see that `indexed_module_streams` will also get you the profiles, but that could be returned from `upsert_all`.

jeremylenz: > Maybe I'm missing something, but what do these hash transformations get us that `result` doesn't already have?

`result` is an ActiveRecord::Result and thus can only give us columns that are in our tables; in other words, attributes of the AvailableModuleStream model. However, the argument passed into this method contains several fields that are _not_ attributes of that model. The other difference is that `result` contains _only_ newly-created records, not existing records.   So we have

`result` - newly inserted Katello DB records, containing only id, name, stream, context
`module_streams` - module stream data containing name, stream, version, context, arch, profiles, installed_profiles, status, active -- but NOT id

The next method, `sync_available_module_stream_associations`, in its original form, expected an indexed Hash by database ID containing _both_ existing _and_ newly created AvailableModuleStreams. It called this `new_available_module_streams` even though it contained both existing and new records. It used this in two ways:

1. Composing a list of AvailableModuleStreams to delete, by comparing the host's `available_module_stream_ids` with `new_available_module_streams.keys`
2. Composing lists of new and upgradable module streams, by using `new_available_module_streams.keys`

The difficult part is that `sync_available_module_stream_associations` relies on our hash meeting _both_ of the following requirements:

1. Indexed by Katello database ID; and
2. contains _both_ new records _and_ existing records.

Now that I think some more, it may be simpler to just look at the database after doing our `upsert_all`, and not rely anymore on the passed-in module stream data.

app/models/katello/concerns/host_managed_extensions.rb
jeremylenz: Indeed they do. I wonder how `profiles` is getting set, since we've never done so here.

ianballou: I think it may be unused!

app/models/katello/concerns/host_managed_extensions.rb
jeremylenz: update: Refactored to index the hash using DB queries, see what you think

jeremylenz:
fixed rubocop

ianballou:
Gonna test first then code review.

ianballou (commented):
This is working great! So far I'm not seeing any issues with the code.

I'll give it a final review once the dev logging is cleaned up.

ianballou (approved):
Still working fine for me! Ruby tests are passing now too.

PR #10879, opened 2024-02-02
Fixes #37139 - Block content view publish during repository publication tasks

sjha4:
Blocks content view publish task while certain repo actions (Sync/upload/remove content) are in progress. Similarly block these repo actions when corresponding CV publish is in progress.
There are multiple places in CV publish orchestration where we fetch repository contents. If repository contents are changed during a publication, different steps in the publish workflow will receive different versions of the repository with possibly different contents. This would lead to improper cv publications. 
Have a few repos and cvs with those repos. 
Try a combination of actions like sync/upload/remove content/generate metadata and associated CV publish to make sure the 2-way blocking works as expected.

### DIFF
--- a/app/models/katello/content_view.rb
+++ b/app/models/katello/content_view.rb
@@ -639,10 +639,12 @@ module Katello
     end
 
     def check_repositories_blocking_publish!
-      repositories_with_blocking_tasks = repositories.select { |repo| repo.blocking_tasks }
+      blocking_tasks = repositories&.map { |repo| repo.blocking_task }&.compact
 
-      if repositories_with_blocking_tasks.any?
-        fail _("Pending tasks detected in repositories of this content view. Please wait before publishing.")
+      if blocking_tasks&.any?
+        errored_tasks = blocking_tasks.uniq.map { |task| "- #{Setting['foreman_url']}/foreman_tasks/tasks/#{task&.id}" }.join("\n")
+        fail _("Pending tasks detected in repositories of this content view. Please wait for the tasks: " +
+                 errored_tasks + " before publishing.")
       end
     end
 
@@ -888,6 +890,17 @@ module Katello
 
     def filtered?
       filters.present?
+    end
+
+    def blocking_task
+      blocking_task_labels = [
+        ::Actions::Katello::ContentView::Publish.name
+      ]
+      ForemanTasks::Task::DynflowTask.where(:label => blocking_task_labels)
+                                     .where.not(state: 'stopped')
+                                     .for_resource(self)
+                                     .order(:started_at)
+                                     .last
     end
 
     protected
--- a/app/models/katello/content_view.rb
+++ b/app/models/katello/content_view.rb
@@ -629,12 +629,21 @@ module Katello
         check_ready_to_import!
       else
         fail _("Import-only content views can not be published directly") if import_only? && !syncable
+        check_repositories_blocking_publish!
         check_composite_action_allowed!(organization.library)
         check_docker_repository_names!([organization.library])
         check_orphaned_content_facets!(environments: self.environments)
       end
 
       true
+    end
+
+    def check_repositories_blocking_publish!
+      repositories_with_blocking_tasks = repositories.select { |repo| repo.blocking_tasks }
+
+      if repositories_with_blocking_tasks.any?
+        fail _("Pending tasks detected in repositories of this content view. Please wait before publishing.")
+      end
     end
 
     def check_docker_repository_names!(environments)
--- a/app/models/katello/repository.rb
+++ b/app/models/katello/repository.rb
@@ -641,16 +641,31 @@ module Katello
                                 for_resource(self).order(:started_at).last
     end
 
-    def blocking_tasks
+    def blocking_task
       blocking_task_labels = [
         ::Actions::Katello::Repository::Sync.name,
-        ::Actions::Katello::Repository::UploadFiles.name
+        ::Actions::Katello::Repository::UploadFiles.name,
+        ::Actions::Katello::Repository::RemoveContent.name,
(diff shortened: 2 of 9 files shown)


app/models/katello/content_view.rb
```diff
@@ -637,6 +638,14 @@ def check_ready_to_publish!(importing: false, syncable: false)
       true
     end
 
+    def check_repositories_blocking_publish!
+      repositories_with_blocking_tasks = repositories.select { |repo| repo.blocking_tasks }
+
+      if repositories_with_blocking_tasks.any?
+        fail _("Pending tasks detected in repositories of this content view. Please wait before publishing.")
```
adamruzicka, line 645: Would it make sense to list the tasks which are preventing this one from running? Something like we did [on lock conflicts](https://github.com/theforeman/foreman-tasks/blob/89864f50d472ac719997b2e52eb15219ad9fcbad/app/models/foreman_tasks/lock.rb#L10)

app/models/katello/content_view.rb
```diff
@@ -637,6 +638,14 @@ def check_ready_to_publish!(importing: false, syncable: false)
       true
     end
 
+    def check_repositories_blocking_publish!
+      repositories_with_blocking_tasks = repositories.select { |repo| repo.blocking_tasks }
+
+      if repositories_with_blocking_tasks.any?
```
adamruzicka, line 644: Not saying you have to use it, but `#any?` can take a block to save some allocations

app/models/katello/content_view.rb
sjha4: Updated..

app/models/katello/content_view.rb
sjha4: Using `any?` more as a conditional check to see if there are any entries in the task array rather than an enumerable to act on..

sjha4:
[test katello]

app/models/katello/repository.rb
```diff
@@ -641,6 +641,32 @@ def latest_dynflow_sync
...
+        ::Actions::Katello::Repository::Sync.name,
+        ::Actions::Katello::Repository::UploadFiles.name,
+        ::Actions::Katello::Repository::RemoveContent.name
+      ]
+      ForemanTasks::Task::DynflowTask.where(:label => blocking_task_labels)
+                                     .where.not(state: 'stopped')
+                                     .for_resource(self)
+                                     .order(:started_at)
+                                     .last
+    end
```
adamruzicka, line 655: And maybe one more thing, the method name is in plural, but it returns either a task or nil, that's slightly misleading

wbclark:
Attempted to publish a CV while one if its repos was syncing:

```
500
Pending tasks detected in repositories of this content view. Please wait for the tasks: - http://centos8-stream-wclark-katello-devel.example.com/foreman_tasks/tasks/7590b00d-5532-49bf-9cbc-2377d1bca5de before publishing.
```

The CV publish task immediately went to stopped/error. The repository sync task remained running/pending.

wbclark (commented):
Summarizing an offline chat we had about this:

1. Should this blocking behavior be controlled by a setting, allowing users with hundreds of repos, automated sync plans, etc to opt out of the blocking behavior to avoid these collisions? Is the current pre-blocking behavior that the CV can be published with the exact repo content that existed prior to the sync, or is there any data inconsistency possible due to the repo contents changing during publishing?

2. Is there a way that we can block these tasks, still clearly informing the user of what is happening when they try to initiate it, without having a "scary" red stopped/error task showing up on the monitor tasks page thereafter? E.g. could we check for blockers and surface them before the task even gets created, or else have the task go to stopped/warning instead (without changing the entire rescue_strategy for other modes of failure, and still surfacing a UI notification explaining what is happening?) @adamruzicka do you have any ideas about this?

sjha4:
>> Should this blocking behavior be controlled by a setting, allowing users with hundreds of repos, automated sync plans, etc to opt out of the blocking behavior to avoid these collisions? Is the current pre-blocking behavior that the CV can be published with the exact repo content that existed prior to the sync, or is there any data inconsistency possible due to the repo contents changing during publishing?

After looking at the CV publish code flow, I am leaning towards blocking it for everyone to prevent any changes between repo contents when the publish starts to when it finishes. Cause many actions in the code path independently looks up repository and contents during the flow and changing repository between these sub actions would lead to unexpected results in the published version. 

>> E.g. could we check for blockers and surface them before the task even gets created.

We could potentially move the validations to the API layer and block creation of task. However syncs and other blocked repo actions are started from multiple places outside of API. ex: sync plans which directly trigger sync action. Having the blocking at action layer would handle all cases including foreman-rake invocations or any rake tasks. I don't think we need exclusive locks on repo during sync or any other actions however and a custom validation to only block publish and nothing else would suffice for this.

sjha4:
[test katello]

wbclark (approved):
LGTM

We can change one commit from "Fixes" to "Refs" or squash on merge

app/models/katello/repository.rb
```diff
@@ -641,6 +641,33 @@ def latest_dynflow_sync
...
+        ::Actions::Katello::Repository::UploadFiles.name,
+        ::Actions::Katello::Repository::RemoveContent.name,
+        ::Actions::Katello::Repository::MetadataGenerate.name
+      ]
+      ForemanTasks::Task::DynflowTask.where(:label => blocking_task_labels)
+                                     .where.not(state: 'stopped')
+                                     .for_resource(self)
+                                     .order(:started_at)
+                                     .last
+    end
```
wbclark, line 656: do you anticipate that we may use this in other areas in the future?

sjha4: Not necessarily..However, this is the list of tasks with potential to create a new publication. So if we need to block anything on itthe future, a repo.check_ready_to_act! should work for that use-case.

PR #10881, opened 2024-02-06
Fixes #37144 - Use OS family for host details page packages tab

nadjaheitmann:
At the moment, the packages sub tab on the new host details page is shown based on the description of the operating system (e.g. checking whether it should load rpm or debs). The condition checks for various strings, but the description itself can be randomly assigned by the user. Hence, this is not a safe method to determine which OS was used.

We need to find an unchangeable attribute to test against for the content packages tabs which is the operatingsystem family. 

Have a host with packages and some OS. Change OS name to something like "Test OS" --> Host packages should still be available.

### DIFF
--- a/app/views/katello/api/v2/hosts/base.json.rabl
+++ b/app/views/katello/api/v2/hosts/base.json.rabl
@@ -4,6 +4,15 @@ object @resource
 @facet = @resource.content_facet
 
 attributes :id, :name, :description
+
+node :operatingsystem_family do |resource|
+  resource.operatingsystem.family
+end
+
+node :operatingsystem_major do |resource|
+  resource.operatingsystem.major
+end
+
 if @facet
   node :content_view do
     content_view = @facet&.single_content_view
--- a/lib/katello/plugin.rb
+++ b/lib/katello/plugin.rb
@@ -303,6 +303,7 @@ Foreman::Plugin.register :katello do
 
   extend_rabl_template 'api/v2/smart_proxies/main', 'katello/api/v2/smart_proxies/pulp_info'
   extend_rabl_template 'api/v2/hosts/show', 'katello/api/v2/hosts/host_collections'
+  extend_rabl_template 'api/v2/hosts/show', 'katello/api/v2/hosts/show'
 
   # Katello variables for Host Registration
   extend_allowed_registration_vars :activation_keys
--- a/test/controllers/api/v2/host_subscriptions_controller_test.rb
+++ b/test/controllers/api/v2/host_subscriptions_controller_test.rb
@@ -8,7 +8,7 @@ module Katello
     tests Katello::Api::V2::HostSubscriptionsController
 
     def models
-      @host = FactoryBot.create(:host, :with_subscription)
+      @host = FactoryBot.create(:host, :with_subscription, :with_operatingsystem)
       users(:restricted).update_attribute(:organizations, [@host.organization])
       users(:restricted).update_attribute(:locations, [@host.location])
       @pool = katello_pools(:pool_one)
--- a/test/controllers/api/v2/hosts_controller_test.rb
+++ b/test/controllers/api/v2/hosts_controller_test.rb
@@ -12,7 +12,7 @@ class Api::V2::HostsControllerTest < ActionController::TestCase
     @environment = katello_environments(:library)
     @dev = katello_environments(:dev)
     @cv2 = katello_content_views(:library_view_no_version)
-    @host = FactoryBot.create(:host)
+    @host = FactoryBot.create(:host, :with_operatingsystem)
   end
 
   def host_index_and_show(host)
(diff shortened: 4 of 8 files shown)


nadjaheitmann:
@jeremylenz We found a bug partly introduced in #10744,  here is a proposed fix :) Test failure seems unrelated if I am not mistaken.

jeremylenz (commented):
Thanks @nadjaheitmann! Yes, I agree this should be addressed. 

I see we're adding an additional API call here. There are a couple other ways this could be accomplished, without adding an API call.

1. The host details response already includes `operatingsystem_name` and `operatingsystem_id`. We could add `operatingsystem_family` to that, either (a) via a Foreman code change so other plugins could take advantage; or (b) by extending it from Katello.
2. The response also already includes `facts`. Foreman uses the `distribution::name` fact to determine the operating system's family. We could look at that fact here as well.  (see https://github.com/theforeman/foreman/blob/e87de8d1a30aeba3da6d0cb2518d94aa450db55e/app/services/katello/rhsm_fact_parser.rb#L49-L51)

nadjaheitmann:
Also, we would need to test the operatingsystem's major version for module streams, so we would also need to expose that one. I agree it would be much simpler if it was part of one API response.

jeremylenz:
> I was actually looking into that one as well, but `Host` class does not have any attribute called `operatingsystem_family` and I don't know how to add it such that it could be exposed in the `.rabl` file for the API. Any hints?

The problem is that the `operatingsystems` table doesn't actually have a `family` attribute; rather it's a column called `type` which is renamed to `family`. So rails doesn't add the `operatingsystem_family` convenience method like it did for `name` and `id`.

```rb
# app/models/operatingsystem.rb
def family
    self[:type]
  end

  def family=(value)
    self.type = value
  end
```

To add family to the rabl in Foreman, this would probably work:

```rb
# app/views/api/v2/hosts/main.json.rabl
node :operatingsystem_family do |host|
  host.operatingsystem.family
end
```

Or, to extend it from Katello, I think you could add the same to `app/views/katello/api/v2/hosts/base.json.rabl`

nadjaheitmann:
Thanks for the detailed explanation, much appreciated! Do you actually need the operating system family anywhere outside Katello?

jeremylenz:
If you're asking where the code should live, I'd be fine with keeping it in Katello.

nadjaheitmann:
Yeah, thanks. I don't get it work in Katello, though. It does not seem to load the `app/views/katello/api/v2/hosts/base.json.rabl`

jeremylenz:
https://github.com/Katello/katello/blob/519564630bda213bc7ab9034657f95b5938493ed/app/controllers/katello/concerns/api/v2/hostgroups_controller_extensions.rb#L31-L35

Hm, maybe you'll need to add something similar to that ^, but in hosts_controller_extensions

jeremylenz:
Another thing to try is `extend_rabl_template`. Seems like `foreman_remote_execution` does this:

https://github.com/theforeman/foreman_remote_execution/blob/e97021ca4417d737f6d60a7586efd786413041db/lib/foreman_remote_execution/engine.rb#L274-L278

nadjaheitmann:
> https://github.com/Katello/katello/blob/519564630bda213bc7ab9034657f95b5938493ed/app/controllers/katello/concerns/api/v2/hostgroups_controller_extensions.rb#L31-L35
> 
> Hm, maybe you'll need to add something similar to that ^, but in hosts_controller_extensions

This actually overrides the original template. 


> Another thing to try is `extend_rabl_template`. Seems like `foreman_remote_execution` does this:

This seems to do the trick, thanks!

nadjaheitmann:
@jeremylenz Looks much better doesn't it :)

app/views/katello/api/v2/hosts/base.json.rabl
```diff
@@ -1,9 +1,19 @@
 object @resource
+object @host
```
jeremylenz, line 2:
```suggestion
```

This seems to not be needed

app/views/katello/api/v2/hosts/base.json.rabl
```diff
@@ -1,9 +1,19 @@
...
 
 attributes :id, :name, :description
+
+node :operatingsystem_family do |host|
+  host.operatingsystem.family
+end
+
+node :operatingsystem_major do |host|
+  host.operatingsystem.major
+end
```
jeremylenz, line 15: This seems to work also..

```suggestion
node :operatingsystem_family do
  @resource.operatingsystem.family
end

node :operatingsystem_major do
  @resource.operatingsystem.major
end
```

jeremylenz (commented):
Looking much better :smile: 

Also tested it and it's working well.

Just a couple small suggestions

app/views/katello/api/v2/hosts/base.json.rabl
nadjaheitmann: True, fixed

jeremylenz:
Once https://github.com/Katello/katello/pull/10884 is merged let's see how the tests look. LGTM pending green tests. :+1:

jeremylenz:
[test katello]

nadjaheitmann:
[2024-02-11T22:26:17.378Z] Error:
[2024-02-11T22:26:17.378Z] Katello::Api::Rhsm::CandlepinDynflowProxyControllerIntegrationTest#test_0001_params are not parsed in the controller:
[2024-02-11T22:26:17.378Z] Errno::ENOENT: No such file or directory @ rb_sysopen - /home/jenkins/workspace/katello-pr-test/foreman/test/fixtures/hosts.yml
[2024-02-11T22:26:17.378Z]     test/test_helper.rb:77:in `before_setup'

Looks like a fixture cannot be found in foreman itself. Any idea how to fix this?

nadjaheitmann:
Now the katello test is green... New failure which I assume is not related (looks funny though):

Failure:
Actions::Pulp3::CopyAllUnitsDockerRepositoryTest#test_exclusion_docker_filters [katello/test/actions/pulp3/orchestration/copy_all_units_test.rb:89]
Minitest::Assertion: Expected: ["musl", "glibc"]
  Actual: ["glibc", "musl"]

jeremylenz:
I re-ran that check and it's green now. Was just an array ordering failure that we get intermittently. :heavy_check_mark:

jeremylenz (approved):
Works well

thanks @nadjaheitmann !

Will merge once we get the React failures sorted.

PR #10900, opened 2024-02-21
Fixes #37191 - upload profile via REX

sbernhard:

### DIFF
--- a/app/views/foreman/job_templates/upload_profile.erb
+++ b/app/views/foreman/job_templates/upload_profile.erb
@@ -1,0 +1,16 @@
+<%#
+kind: job_template
+name: Upload profile - Katello Script Default
+job_category: Katello
+model: JobTemplate
+provider_type: script
+description_format: Upload package profile for a host
+feature: katello_upload_profile
+%>
+#!/bin/sh
+<% if @host.operatingsystem.family == 'Redhat' -%>
+dnf uploadprofile --force-upload
+<% else -%>
+package-profile-upload --force-upload
+<% end -%>
+subscription-manager repos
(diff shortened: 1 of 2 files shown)


app/views/foreman/job_templates/upload_profile.erb
```diff
@@ -0,0 +1,15 @@
...
+model: JobTemplate
+provider_type: script
+description_format: Upload profile for a host
+feature: katello_upload_profile
+%>
+#!/bin/sh
+<% if @host.operatingsystem.family == 'Redhat' -%>
+dnf uploadprofile --force-upload
+<% else -%>
+package-profile-upload --force-upload
```
jeremylenz, line 14: `dnf uploadprofile --force` will upload a package profile, but does not trigger Katello to recalculate errata applicability or upgradable packages. For this reason we run `subscription-manager repos` using the "Run Command" template in a couple places on the host details page, as this will do both. Depending on your use case you may want to consider that here.

app/views/foreman/job_templates/upload_profile.erb
```diff
@@ -0,0 +1,15 @@
+<%#
+kind: job_template
+name: Upload profile - Katello Script Default
+job_category: Katello
+model: JobTemplate
+provider_type: script
+description_format: Upload profile for a host
```
jeremylenz, line 7:
```suggestion
description_format: Upload package profile for a host
```

app/views/foreman/job_templates/upload_profile.erb
sbernhard: Sounds to me, that we should do a `dnf uploadprofile` first and a `subscription-manager repos` afterwards. Am I correct?

jeremylenz: If you do `subscription-manager repos` you don't have to run uploadprofile. Sub-man repos is all that's needed, in my experience.

jeremylenz: Another thing to consider: I have not tested how it behaves if you've changed the sub-man default config for `package_profile_on_trans` from 1 to 0.

sbernhard: I tried with sub-man only but this does not allow to force to upload the profile. sub-man repos does not push the profile everytime and it does not start to calculate errata applicability. If the host is not able to reach foreman/katello the first time, in case of debugging or if a vmware rollback is done for the host, I want to force to upload the list sometimes.

jeremylenz: We discussed offline and determined we should probably do both, to be safe. Sub-man repos only uploads profiles when it knows there has been a change, and the lack of applicability calculations may be solved by more recent katello code changes.

sbernhard:
Done @jeremylenz

sbernhard:
@jeremylenz I think its better to have another PR to update the frontend. Would be good if someone else can work on this. Or I will come back to in about 2-3 weeks (sorry for that - priorities....)

jeremylenz (approved):
@sbernhard that's fair. Opened https://projects.theforeman.org/issues/37225.

Tested and works as advertised. ACK :+1:

PR #10901, opened 2024-02-21
Fixes #37192 - Use with_enabled_email scope

adamruzicka:
Instead of typing out the same query explicitly.

Reusing a scope introduced in https://github.com/theforeman/foreman/pull/9892 from foreman instead of doing the same thing explicitly.

Email notifications on capsule sync, cv promote, cv publish and repo sync should still work exactly as they did, but now with less code.


sjha4 (approved):
Ack..Test failures unrelated. Will merge once master pipelines are green..

PR #10902, opened 2024-02-21
Fixes #37194 - Uses ENV option for RHSM and UPSTREAM_RHSM api url

parthaa:
Provides a way for setting an alternate rhsm url to test against staging environments.

`REDHAT_RHSM_API_URL=http://alternate-rhsm-fqdn  bundle exec foreman start`

You should see upstream request subscriptions go to the  alternate server.

### DIFF
--- a/app/models/katello/glue/provider.rb
+++ b/app/models/katello/glue/provider.rb
@@ -34,7 +34,7 @@ module Katello
       API_URL = 'https://subscription.rhsm.redhat.com/subscription/consumers/'.freeze
       def api_url(upstream = {})
         # Default to Red Hat
-        upstream['apiUrl'] || API_URL
+        ENV['REDHAT_RHSM_API_URL'] || upstream['apiUrl'] || API_URL
       end
 
       def sync
@@ -65,7 +65,6 @@ module Katello
           Rails.logger.error "Upstream identity certificate not available"
           fail _("Upstream identity certificate not available")
         end
-
         params = {}
         params[:capabilities] = Resources::Candlepin::CandlepinPing.ping['managerCapabilities'].inject([]) do |result, element|
           result << {'name' => element}
--- a/app/lib/katello/resources/candlepin/upstream_job.rb
+++ b/app/lib/katello/resources/candlepin/upstream_job.rb
@@ -11,7 +11,7 @@ module Katello
           end
 
           def get(id, upstream)
-            url = API_URL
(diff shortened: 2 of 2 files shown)


app/models/katello/glue/provider.rb
```diff
@@ -74,9 +74,13 @@ def owner_upstream_update(upstream, _options)
                                                       upstream['idCert']['key'], ca_file, params)
       end
 
+      def upstream_api_url(upstream)
+        upstream['apiUrl'] || ENV['UPSTREAM_REDHAT_RHSM_API_URL'] ||  API_URL
```
jeremylenz, line 78: In `upstream_job.rb` the variable is called `REDHAT_RHSM_API_URL`; here it's `UPSTREAM_REDHAT_RHSM_API_URL`. Is that intended?

parthaa: Moved it to the same value. No need to keep it separate.

app/models/katello/glue/provider.rb
```diff
@@ -34,7 +34,7 @@ module InstanceMethods
       API_URL = 'https://subscription.rhsm.redhat.com/subscription/consumers/'.freeze
       def api_url(upstream = {})
         # Default to Red Hat
-        upstream['apiUrl'] || API_URL
+        upstream['apiUrl'] || ENV['REDHAT_RHSM_API_URL'] || API_URL
```
jeremylenz, line 37: This means that you won't be able to control this via environment variable if the upstream metadata already contains `apiUrl`. Is this what we want?

if not then maybe should switch the order

parthaa: hmm good point :)

```suggestion
        ENV['REDHAT_RHSM_API_URL'] || upstream['apiUrl'] || API_URL
```

jeremylenz:
@parthaa please clean up the commits etc., code LGTM :+1:

parthaa:
[test katello]

jeremylenz:
> [test katello]

That was a Jenkins thing :yum:

jeremylenz:
@parthaa looks like this is the Mocha failure which should be fixed now if you rebase again

jeremylenz:
To be green this will also need to have the change discussed here -> https://github.com/Katello/katello/pull/11085#issuecomment-2251284626

jeremylenz (approved):

PR #10899, opened 2024-02-20
Fixes #37187 - Update ACS refresh Pulp fixtures + fix repository_test test

ianballou:
#### What are the changes introduced in this pull request?
Fixes failing testes on master (hopefully)

#### Considerations taken when implementing this change?
This will need to be backported to 4.11.1 and 4.12.0.

#### What are the testing steps for this pull request?
See that the tests pass.

PR #10905, opened 2024-02-22
Fixes #37169 - Managing a Hosts Repository Sets does not behave as expected

Thorben-D:
Ticket with information on how to reproduce and pictures: [Bug #37169](https://projects.theforeman.org/issues/37169)

This pull request fixes #37169 by adding a new "limit_to_env" parameter to `/api/v2/hosts/<id>/subscriptions/content_override`.

**Detailed problem description**

When entering the Repository sets page, a request is made to the _RepositorySetsController_, which returns the according repository sets.
When the user selects "**Limit to environment"**, a new request is made. This time however, the parameter `content_access_mode_env` is set to true and included in the request. _RepositorySetsController_ now returns the subset of repositories which all belong to the environment.

When the user selects "Override to enabled/disabled" from the kebab-menu in the toolbar, a function called `fetchBulkParams`, which is part of `tableHooks` is called. This function queries the table and constructs a valid search query, which is then passed to the _HostSubscriptionsController_, where repositories described by the query are enabled/disabled.

When the user selects every table-entry by using the "select all" checkbox in the toolbar, a flag `selectAllMode` is set and `fetchBulkParams` evaluates to `''`. The _HostSubscriptionsController_ evaluates this query and enables/disables every repository.
While this works fine for "**Show all**", if "**Limit to environment**" is selected in the UI, the information that the `''`-query should only be applied to a subset is not passed to _HostSubscriptionsController_ and instead of only repositories belonging to the environment, all repositories are enabled/disabled.

**This subset problem may very well be still an issue elsewhere**

I found that the easiest way to address this issue without breaking existing usages, was to partially replicate the functionality of _RepositorySetsController_ and add a parameter to indicate that a query should only be applied to items belonging to the environment.

### DIFF
--- a/app/controllers/katello/api/v2/host_subscriptions_controller.rb
+++ b/app/controllers/katello/api/v2/host_subscriptions_controller.rb
@@ -170,6 +170,7 @@ module Katello
     param :content_overrides_search, Hash, :desc => N_("Content override search parameters") do
       param_group :search, Api::V2::ApiController
       param :enabled, :bool, :desc => N_("Set true to override to enabled; Set false to override to disabled.'"), :required => false
+      param :limit_to_env, :bool, :desc => N_("Limit actions to content in the host's environment."), :required => false
       param :remove, :bool, :desc => N_("Set true to remove an override and reset it to 'default'"), :required => false
     end
     def content_override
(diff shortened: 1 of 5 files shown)


app/controllers/katello/api/v2/host_subscriptions_controller.rb
```diff
@@ -232,10 +233,22 @@ def action_permission
...
+
+        if Foreman::Cast.to_bool(params.dig(:content_overrides_search, :limit_to_env))
+          env_content = ProductContentFinder.new(
+              :match_subscription => false,
+              :match_environment => true,
+              :consumable => @host.subscription_facet
+          ).product_content
+          env_content_labels = ::Katello::Content.find(env_content.pluck(:content_id)).pluck(:label)
+          content_labels &= env_content_labels
+        end
```
m-bucher, line 250: I am wondering if this also needs to be done in case `:search` is `nil`?

m-bucher:
[test katello]

sbernhard (approved):

sbernhard:
This doesn't look like the failed Katello tests are related.

jeremylenz:
[test katello]

jeremylenz:
> This doesn't look like the failed Katello tests are related.

No, I think https://github.com/Katello/katello/pull/10899 should fix that.

sbernhard:
[test katello]

sbernhard (approved):
All tests green. LGTM! Thank you @Thorben-D

Thorben-D:
@jeremylenz Could you take one last look?

jeremylenz:
Thanks for the contribution @Thorben-D! Tested and works correctly :+1: 

I would like to see a test added in `host_subscriptions_controller_test.rb` which asserts that the `limit_to_env` param works as designed. Will ack once that's added.

sbernhard:
[test katello]

jeremylenz (approved):
Don't know how I missed that before, thanks! Verified that `test_find_content_overrides_with_empty_string_search_limited_to_environment` fails on master without the PR and passes here. And thanks for adding the React test.

Thanks @Thorben-D! 

APJ :+1:

PR #10897, opened 2024-02-20
Fixes #37178 - use safe navigation for operatingsystem

jeremylenz:
#### What are the changes introduced in this pull request?

Use Ruby safe navigation operator for `operatingsystem`. This fixes a bug introduced in https://github.com/Katello/katello/pull/10881

#### What are the testing steps for this pull request?

Clone a registered rhel8 host
Make sure you can display the host details page without getting a 500

PR #10908, opened 2024-02-26
Fixes #37206 - Allow hosts to not have an os

adamruzicka:
Fixes f14f58e

Hosts are no longer required to have an OS assigned in order for host details to show for them.

While having an OS assigned makes sense from katello point of view, we didn't require it prior to https://github.com/Katello/katello/commit/f14f58e3d7dc2a3c5a14b7b64592d5c9a9900868#

1) Have a host without an OS, see the hammer command below
2) Go to /hosts/myhost.example.com

```shell
    hammer host create\
        --name "myhost.example.com" \
        --managed false \
        --build false \
        --organization "Default Organization" \
        --location "Default Location"
```

Without this page the backend returns 500 and frontend redirects to the not found page.

### DIFF
--- a/app/views/katello/api/v2/hosts/base.json.rabl
+++ b/app/views/katello/api/v2/hosts/base.json.rabl
@@ -6,11 +6,11 @@ object @resource
 attributes :id, :name, :description
 
 node :operatingsystem_family do |resource|
-  resource.operatingsystem.family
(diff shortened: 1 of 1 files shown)


ianballou:
Closing since https://github.com/Katello/katello/pull/10897 is merged and has the same exact changes.

PR #10911, opened 2024-02-28
Fixes #36976 - Too many audit records slow down CV loading

sjha4:
Query based on version.created_at instead of filtering through records after fetching.
Too many audit records in a system can make the filtering in-memory really slow.
Have a box with a very large number of audit records. 
Try loading CV index page. 
It shouldn't take forever to load. 😸

### DIFF
--- a/app/models/katello/content_view.rb
+++ b/app/models/katello/content_view.rb
@@ -770,45 +770,6 @@ module Katello
       content_views
     end
 
-    def audited_cv_repositories_since_last_publish
-      audited_repositories = self.audits.filter do |a|
-        !a.audited_changes["repository_ids"].nil?
-      end
-      if latest_version_object
-        audited_repositories.filter! { |a| a.created_at > latest_version_object.created_at }
-      end
-      audited_repositories
-    end
-
-    def audited_cv_repository_changed
-      audited_repositories_changes = Audit.where(auditable_id: repositories, auditable_type: "Katello::Repository").filter do |a|
-        a.audited_changes["publication_href"].present? || a.audited_changes["version_href"].present?
-      end
-      if latest_version_object
-        audited_repositories_changes = audited_repositories_changes.filter { |a| a.created_at > latest_version_object.created_at }
-      end
-      audited_repositories_changes
-    end
-
-    def audited_cv_filters_changed
-      audited_filters = Audit.where(auditable_type: "Katello::ContentViewFilter",
-                                    associated_id: id,
-                                    associated_type: "Katello::ContentView")
-      if latest_version_object
-        audited_filters = audited_filters.filter { |a| a.created_at > latest_version_object.created_at }
-      end
-      audited_filters
-    end
-
-    def audited_cv_filter_rules_changed
-      audited_filter_rules = Audit.where(associated_id: id,
-                                    associated_type: "Katello::ContentView").where("auditable_type LIKE '%FilterRule%'")
-      if latest_version_object
-        audited_filter_rules = audited_filter_rules.filter { |a| a.created_at > latest_version_object.created_at }
-      end
-      audited_filter_rules
-    end
-
     def composite_cv_components_changed?
       return true unless latest_version_object
       published_component_version_ids = latest_version_object.components.pluck(:id) || []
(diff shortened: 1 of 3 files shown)


app/models/katello/content_view.rb
```diff
@@ -770,43 +770,30 @@ def related_composite_cvs
       content_views
     end
 
-    def audited_cv_repositories_since_last_publish
-      audited_repositories = self.audits.filter do |a|
+    def audited_cv_repositories_since_last_publish?
+      self.audits.where("created_at > ?", latest_version_object.created_at).any? do |a|
         !a.audited_changes["repository_ids"].nil?
       end
```
jeremylenz, line 776: The `.any?` here uses Ruby (which is slow) to do the bulk of the work. I see that `audited_changes` is a `text` field that holds JSON, which Rails/Ruby converts into a Hash for us. I wonder if there's a way to use SQL to accomplish the equivalent of `!a.audited_changes["repository_ids"].nil?` If we could incorporate that into our `where` somehow, Postgres (which is much faster than Ruby) could do all these queries.

app/models/katello/content_view.rb
```diff
@@ -874,8 +861,8 @@ def needs_publish?
     end
 
     def audited_changes_present?
-      audited_cv_repositories_since_last_publish.present? || audited_cv_repository_changed.present? ||
-        audited_cv_filters_changed.present? || audited_cv_filter_rules_changed.present?
+      audited_cv_repositories_since_last_publish? || audited_cv_repository_changed? ||
+        audited_cv_filters_changed? || audited_cv_filter_rules_changed?
```
jeremylenz, line 865: Should make sure this is ordered from least expensive to most expensive.

jeremylenz (commented):
This does help load time with my 20 CVs, but only by about 1 second (~3.5sec --> ~2.5sec). Maybe I don't have enough audits to make a big difference..

sjha4:
>> This does help load time with my 20 CVs, but only by about 1 second (~3.5sec --> ~2.5sec). Maybe I don't have enough audits to make a big difference..

Ya..The load time increases a lot based on audit records..The reproducer on the issue has 200k.. I was testing with 20k and it was a noticeable difference..

app/models/katello/content_view.rb
```diff
@@ -770,43 +770,30 @@ def related_composite_cvs
       content_views
     end
 
-    def audited_cv_repositories_since_last_publish
-      audited_repositories = self.audits.filter do |a|
+    def audited_cv_repositories_since_last_publish?
+      self.audits.where("created_at > ?", latest_version_object.created_at).any? do |a|
```
sbernhard, line 774: this would run another SQL query. Can this be prevented?

app/models/katello/content_view.rb
sjha4: Pushed a commit with this. This is what I have with ~70k audit records:
Screenshot from 2024-02-29 14-43-55

app/models/katello/content_view.rb
sjha4: The `self.audit` should be fast since the foreign key fields are indexed. The alternative would be to run a `Audit.where..` query which could merge those.

Rails also seems to be optimizing the self.audits relation and where clause into one query.
```
[2] pry(main)> cv.audited_cv_repositories_since_last_publish?
2024-02-29T19:57:37 [D|sql|]   Katello::ContentViewVersion Load (0.6ms)  SELECT "katello_content_view_versions".* FROM "katello_content_view_versions" WHERE "katello_content_view_versions"."content_view_id" = $1 ORDER BY major DESC, minor DESC LIMIT $2  [["content_view_id", 2], ["LIMIT", 1]]
2024-02-29T19:57:37 [D|sql|]   Audited::Audit Load (0.8ms)  SELECT "audits".* FROM "audits" WHERE "audits"."auditable_id" = $1 AND "audits"."auditable_type" = $2 AND (created_at > '2024-02-29 17:12:01.448064') ORDER BY "audits"."version" ASC  [["auditable_id", 2], ["auditable_type", "Katello::ContentView"]]
=> false

```

sbernhard: Well, better would be, that all this stuff is not loaded in separate queries and at least not n+1. Use eager loading and use already loaded data if possible.

sjha4: Aah..I see..I see your PR for cv components along the same lines. Le me try that for audits..
https://github.com/Katello/katello/pull/10911

app/models/katello/content_view.rb
```diff
@@ -770,43 +770,30 @@ def related_composite_cvs
...
-      end
-      if latest_version_object
-        audited_repositories_changes = audited_repositories_changes.filter { |a| a.created_at > latest_version_object.created_at }
-      end
-      audited_repositories_changes
+    def audited_cv_repository_changed?
+      Audit.where(auditable_id: repositories, auditable_type: "Katello::Repository")
+           .where("created_at > ?", latest_version_object.created_at)
+           .where("audited_changes LIKE :publication_href OR audited_changes LIKE :version_href", publication_href: "%publication_href%", version_href: "%version_href%")
+        .any?
```
parthaa, line 783:
```suggestion
        .exists?
```

sjha4:
The latest commit has following chnages:
1. Removes needs_publish from index rabl and only shows it on details page
2. Updates audited_changes_present? to merge all audit queries into one to reduce number of queries.

parthaa (approved):
Tested the following scenarios
1.  Add/Remove repositories
2. Add/Remove filters
3. Force Synced repos

All worked well. Nice Work!. ACK

PR #10920, opened 2024-03-08
 Refs #37214 - Use Limit to Environment by default

sbernhard:
Limit to Environment should be used by default. Unfortunatley, while loading the page the "Limit to Environment" checkbox is checked but the list of repository sets isn't filtered accordingly

Fix it :-) 

1. Verify, that the Limit to Environment checkbox is enabled
2. Verify, that the repositories are filtered by the selected Content View
3. Verify, that the list of repositories are unfilterd, if Limit to Enviroment is un-checked


chris1984 (approved):
ACK, looks good:

When loading up the page we see the checkbox checked and filtered to my CV of RHEL 8 repos:

prak1load

When I uncheck, it I can see all the other repos:
prkakunchecked

PR #10925, opened 2024-03-08
Fixes #37238 - show upstream auth token only for yum repos

ianballou:
Hides the upstream authentication token for non-yum repositories. This field is only helpful for SLES repositories for now, but could potentially be used for other purposes in the future. By hiding it for other repositories, it's less likely that users will get confused.

I considered renaming the field to "SLES token" or something like that, but SLES users by now are likely used to that naming convention. Plus, the field could be expanded in the future to be useful for other repositories.

1. Make sure that the upstream auth token field is only settable for yum repos
2. Make sure that the upstream auth token field is still settable for yum repos (check the Pulp remote)

---

There might be some tests I need to add / updates.


ianballou:
Note that fields not usable by certain repo types still show up in hammer help-text. That's normal since the API definition isn't content type dependent.

ianballou:
@sjha4: 
```ruby
param :upstream_authentication_token, String, :desc => N_("Upstream authentication token string for yum repositories.")
```

sjha4:
Do we care about the 2 failures on CI?

ianballou:
Will fix with https://projects.theforeman.org/issues/37277

ianballou:
[test katello]

sjha4 (approved):
Ack🍏

PR #10926, opened 2024-03-11
Fixes #36979 - Remove cdn_ssl_version setting

wbclark:
This was originally added to allow downgrading the CDN connection SSL version for compatibility with much older proxy servers. That should be less of a concern now.

We do still set a value of TLS v1.2 for the min_version, but only because https://github.com/ruby/openssl/issues/709 prevents using the system-wide crypto policy for now. In the future, that can be removed as well, restoring control to the user at the OS level.

This change is a simpler alternative to https://github.com/Katello/katello/pull/10824

### DIFF
--- a/app/lib/katello/resources/cdn.rb
+++ b/app/lib/katello/resources/cdn.rb
@@ -1,8 +1,6 @@
 module Katello
   module Resources
     module CDN
-      SUPPORTED_SSL_VERSIONS = ['SSLv23', 'TLSv1'].freeze
-
       class Utils
         # takes releasever from contentUrl (e.g. 6Server, 6.0, 6.1)
         # returns hash e.g. {:major => 6, :minor => "6.1"}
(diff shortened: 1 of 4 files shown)


ianballou:
I'm pretty sure those test failures are unrelated. Perhaps  your branch isn't rebased on master? The most important part is that https://ci.theforeman.org/blue/organizations/jenkins/katello-pr-test/detail/katello-pr-test/14131/tests is passing. Also the failing VCR test passes locally for me.

ianballou (approved):
Looking fine to me! Tested syncing from the CDN before and after applying the patch. Also set a cdn_ssl_version before patching just to make sure nothing funky happened.

wbclark:
> Perhaps your branch isn't rebased on master?

It is, so I'm not sure what's happening. I've tried re-running the failed jobs to see if it happens consistently.

ianballou:
We should be good to merge, the errors don't make much sense.

PR #10882, opened 2024-02-06
Fixes #37148 - Remove subscription and system purpose host statuses

jeremylenz:
#### What are the changes introduced in this pull request?

image

With SCA-Only the following host statuses are now irrelevant:

Katello::SubscriptionStatus,
Katello::PurposeSlaStatus,
Katello::PurposeRoleStatus,
Katello::PurposeUsageStatus,
Katello::PurposeAddonsStatus,
Katello::PurposeStatus

This PR makes the following changes:

1. Remove the host statuses above
2. Via a DB migration, remove both the status records and the user table preferences for subscription_status
3. Remove the two subscription-related dashboard widgets
4. Remove handling of Candlepin events whose sole purpose was to update the now-removed statuses.

#### Considerations taken when implementing this change?

Note that system purpose _attributes_ are not removed here. You can still set release version, role, usage, etc. and these are still useful for reporting. What is removed are the statuses that tell you whether your syspurpose attributes are "correctly supported by a subscription" -- because that's not a thing with SCA always enabled.

Rather than remove Candlepin event handling entirely, I chose to still log these events in case that's needed for debugging. I figure once there are no more Candlepin events to handle, that will be a good time to actually remove the event handler.

#### What are the testing steps for this pull request?

Make sure you've run 
```
bundle exec rake katello:upgrades:4.12:update_content_access_modes
```

before beginning to test.

Create a TablePreference by going to the old host UI and clicking 'Manage columns,' then select "Subscription status" and a few others from the Content section.

You can see this in `bundle exec rails console` via

```
::TablePreference.all
```

You will see that `"subscription_status"` is included in the `columns` field.

Run migrations:

```
bundle exec rails db:migrate
```

Now, you should see that  `"subscription_status"` is removed from your TablePreference. Also, this should now return nothing:

```
obsolete_statuses = [
      "Katello::SubscriptionStatus",
      "Katello::PurposeStatus",
      "Katello::PurposeAddonsStatus",
      "Katello::PurposeRoleStatus",
      "Katello::PurposeSlaStatus",
      "Katello::PurposeUsageStatus"
    ]

    ::HostStatus::Status.where(type: obsolete_statuses)
```

Make sure everything still works:

- Legacy content host details - Setting syspurpose attributes
- New host details - Setting syspurpose attributes
- Activation keys - setting syspurpose attributes
- Dashboard - widgets
- Host status views

PR #10927, opened 2024-03-11
Refs #37148 - Remove removed_widgets override

jeremylenz:

https://github.com/Katello/katello/pull/10882 removed subscription-related widgets, but we still had some code that refers to them. This code was an override of Foreman's `removed_widgets` method and is no longer needed.

This is causing the Dashboard page to break with

```
uninitialized constant SUBSCRIPTION_TEMPLATES
Extracted source (around line #16):

        if Organization.current&.simple_content_access?
          widgets.reject! { |widget| ::Widget.singleton_class::SUBSCRIPTION_TEMPLATES.include? widget[:template] }
        end

        widgets

Rails.root: /home/vagrant/foreman

Application Trace | Framework Trace | Full Trace
/home/vagrant/katello/app/helpers/katello/concerns/dashboard_helper_extensions.rb:16:in `block in removed_widgets'

```

Both of the referenced widgets are removed, so removing this is correct

On master: Load the dashboard page. You will get that error.

On this branch: Load the dashboard page. You should get no error.

### DIFF
--- a/app/helpers/katello/concerns/dashboard_helper_extensions.rb
+++ b/app/helpers/katello/concerns/dashboard_helper_extensions.rb
@@ -8,16 +8,6 @@ module Katello
       def total_host_count
         host_query.size
       end
-
-      def removed_widgets
-        widgets = super
-
-        if Organization.current&.simple_content_access?
-          widgets.reject! { |widget| ::Widget.singleton_class::SUBSCRIPTION_TEMPLATES.include? widget[:template] }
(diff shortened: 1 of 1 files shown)


m-bucher (approved):
Fixes the issue on my devel-forklift, thanks.

ianballou (approved):
Works for me! To reproduce, you must have some widgets that are deleted.

PR #10928, opened 2024-03-12
Fixes #37240 - Fix CCV duplicate repo warning

m-bucher:
only show duplicate repo warning on CCV publish, if it is actually true.

duplicate repo value is usually `[]` if no duplicates are present.

publish CCV without duplicate repos => message should not be there.


m-bucher:
@sjha4 this was caused by https://github.com/Katello/katello/pull/10741

sjha4 (approved):
Thanks @m-bucher ! Ack..Can reproduce and the commit fixes it.

PR #10931, opened 2024-03-13
pin victory-core to ~36.8.6

evgeni:

pinning victory-core to the same version as foreman, so that CI passes

removing victory-core from our dep-chain was not an option

"React Tests (non-blocking) (Required)" apss


jeremylenz:
wait why is it failing

evgeni:
https://github.com/theforeman/foreman/pull/10095

PR #10933, opened 2024-03-13
Fixes #37254 - Unpublishable content views should always return false for needs_publish?

sjha4:
Always return false as needs_publish for default, import_only and generated content views.
Default, import only and other generated CVs can not be published by user. We should return false for such CVs immediately and not look for change etc.
In console:
```
cv = Katello::ContentView.default.first
cv.needs_publish?
```
should be false
Export/import a cv
In console:
```
cv = Katello::ContentView.find(:id)
cv.needs_publish?
```
Should also be false.


parthaa (approved):
Works well. ACK but can you add a quick model test for unpublishable?

PR #10934, opened 2024-03-13
Fixes #37259 - Only repair library instance repos for products

sjha4:
Only repair library instances of repositories when product verify checksum is run.
Steps to Reproduce:
1. Sync a repository
2. Add this repo to a content view
3. Create 1 or more content view versions with different filters to exclude any package to cause pulp to create a clone of the repository
4. WebUI > Content > Product > [select a checkbox for a product] > Select Action > Verify Content Checksum, execute a verify checksum action on the product
5. Watch the tasks that spawn in foreman-tasks and make sure you see only repair tasks for library instance repos and not all clones and version repos.


chris1984 (approved):
ACK works well:

Before the pr I got 68 tasks:
beforeprtask

After the PR I got 9 tasks:
afterprtask

PR #10935, opened 2024-03-14
Fixes #37266 - Regenerate upstream identity certificate on manifest refresh

jeremylenz:

When a manifest is imported into Katello, a consumer is created in hosted Candlepin. We call this the "upstream consumer." This upstream consumer has an identity certificate which is stored in Katello's database and sent along with any requests to upstream Candlepin (adding subscriptions to the manifest, refreshing, etc.) The normal expiration for this certificate is 1 year from the time it was exported from hosted Candlepin.

When the upstream consumer's identity certificate expires, Katello calls this an "expired manifest." With an expired manifest, your hosts can no longer access content. You must import a new manifest with a nonexpired identity certificate, which can be quite inconvenient for users.

With this change, Katello will request hosted Candlepin to regenerate the upstream consumer identity certificate as the first step of a manifest refresh action. Thus, if you refresh your manifest at least once a year, you'll never be caught with an expired manifest. We do this by calling the upstream Candlepin endpoint `POST consumers/{consumer_uuid}`.

This PR also adds a `manifest_expiration_date` method to Organizations.

This step isn't essential for a manifest refresh itself, so I set the `rescue_strategy` to Skip. This way it won't halt the entire manifest refresh task if something fails with regenerating the identity certificate.

In Rails console, check your manifest expiration date:

```
org = Organization.find 1
org.manifest_expiration_date

...

=> Fri, 01 Nov 2024 16:31:23
```

Now, refresh your manifest and check again:
```
org = Organization.find 1
org.manifest_expiration_date

...

=> Fri, 14 Mar 2025 16:33:29
```

### DIFF
--- a/app/lib/katello/resources/candlepin/upstream_consumer.rb
+++ b/app/lib/katello/resources/candlepin/upstream_consumer.rb
@@ -40,19 +40,31 @@ module Katello
             raise ::Katello::Errors::UpstreamEntitlementGone
           end
 
-          def get_export(url, client_cert, client_key, ca_file)
+          def start_upstream_export(url, client_cert, client_key, ca_file)
             logger.debug "Sending GET request to upstream Candlepin: #{url}"
-            return resource(url: url, client_cert: client_cert, client_key: client_key, ca_file: ca_file).get
+            resource(url: url, client_cert: client_cert, client_key: client_key, ca_file: ca_file).get
           rescue RestClient::Exception => e
             raise e
           end
 
+          alias_method :retrieve_upstream_export, :start_upstream_export
+
           def update(url, client_cert, client_key, ca_file, attributes)
             logger.debug "Sending PUT request to upstream Candlepin: #{url} #{attributes.to_json}"
-            return resource(url: url, client_cert: client_cert, client_key: client_key, ca_file: ca_file).put(attributes.to_json,
-                                                                       'accept' => 'application/json',
-                                                                       'accept-language' => I18n.locale,
-                                                                       'content-type' => 'application/json')
+            resource(
+              url: url,
+              client_cert: client_cert,
+              client_key: client_key,
+              ca_file: ca_file).put(
+                attributes.to_json,
+                'accept' => 'application/json',
+                'accept-language' => I18n.locale,
+                'content-type' => 'application/json')
+          end
+
+          def regenerate_upstream_identity(url, client_cert, client_key, ca_file)
+            logger.debug "Sending POST request to upstream Candlepin: #{url}"
+            resource(url: url, client_cert: client_cert, client_key: client_key, ca_file: ca_file).post(nil)
           end
 
           def bind_entitlement(**pool)
--- a/app/models/katello/glue/provider.rb
+++ b/app/models/katello/glue/provider.rb
@@ -32,6 +32,10 @@ module Katello
 
     module InstanceMethods
       API_URL = 'https://subscription.rhsm.redhat.com/subscription/consumers/'.freeze
+      def api_url(upstream = {})
+        # Default to Red Hat
+        upstream['apiUrl'] || API_URL
+      end
 
       def sync
         Rails.logger.debug "Syncing provider #{name}"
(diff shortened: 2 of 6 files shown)


app/lib/katello/resources/candlepin/upstream_consumer.rb
```diff
@@ -42,19 +42,24 @@ def remove_entitlement(entitlement_id)
 
           def get_export(url, client_cert, client_key, ca_file)
```
qcjames53, line 43: These three methods (get_export, update, post_consumer) are all named inconsistently. Since you're introducing the third, this is probably a good time for a refactor. Is there anything preventing this?

app/lib/katello/resources/candlepin/upstream_consumer.rb
```diff
@@ -42,19 +42,24 @@ def remove_entitlement(entitlement_id)
...
-            return resource(url: url, client_cert: client_cert, client_key: client_key, ca_file: ca_file).get
+            resource(url: url, client_cert: client_cert, client_key: client_key, ca_file: ca_file).get
           rescue RestClient::Exception => e
             raise e
           end
 
           def update(url, client_cert, client_key, ca_file, attributes)
             logger.debug "Sending PUT request to upstream Candlepin: #{url} #{attributes.to_json}"
-            return resource(url: url, client_cert: client_cert, client_key: client_key, ca_file: ca_file).put(attributes.to_json,
+            resource(url: url, client_cert: client_cert, client_key: client_key, ca_file: ca_file).put(attributes.to_json,
```
qcjames53, line 52: Any way we can break this up? 123 characters on this line and some wild formatting on the next few lines.

app/models/katello/glue/provider.rb
```diff
@@ -62,32 +66,33 @@ def owner_upstream_update(upstream, _options)
...
           upstream['idCert']['key'], ca_file)
         JSON.parse(response)
       end
 
       def retrieve_owner_upstream_export(upstream, zip_file_path, export_id)
         validate_upstream_identity_cert!(upstream)
-        url = upstream['apiUrl'] || API_URL
-
-        data = Resources::Candlepin::UpstreamConsumer.get_export("#{url}#{upstream['uuid']}/export/#{export_id}", upstream['idCert']['cert'],
+        data = Resources::Candlepin::UpstreamConsumer.get_export("#{api_url(upstream)}#{upstream['uuid']}/export/#{export_id}", upstream['idCert']['cert'],
```
qcjames53, line 95: Can we fix the formatting here as well? 156 chars plus a wild line afterwards.

qcjames53 (changes requested):
I have verified that the refresh action resets time until expiry on the manifest.  The `manifest_expiration_date` property in console seems to work properly for old, refreshed, missing, and malformed manifests. The logging works and (I think) is an appropriate amount of detail.

I had a few style and naming issues but other than that everything looks great to me!

app/lib/katello/resources/candlepin/upstream_consumer.rb
jeremylenz: let's see what Rubocop will let me do :)

We do have a pretty generous max line length in our config for some reason. :man_shrugging:

app/lib/katello/resources/candlepin/upstream_consumer.rb
jeremylenz: They all do have reasons for the naming :stuck_out_tongue: 

If you look vertically it makes a bit more sense:
`Actions::Candlepin::Owner::RegenerateUpstreamIdentityCert` calls
`Katello::Glue::Provider#owner_upstream_regenerate_identity_cert` which calls
`Resources::Candlepin::UpstreamConsumer.post_consumer` which calls
`Katello::Resources::Candlepin::UpstreamCandlepinResource#post`

`Actions::Candlepin::Owner::StartUpstreamExport` calls
`Katello::Glue::Provider#start_owner_upstream_export` which calls
`Resources::Candlepin::UpstreamConsumer.get_export` which calls
`Katello::Resources::Candlepin::UpstreamCandlepinResource#get`

In each of these chains the more "general" name (get, post, etc.) is at the end. These methods are named specifically rather than generally because it made more sense to me to keep that pattern. I'd be open to changing them to something else (also non-general) if that's what you think is needed? But I wouldn't know what to call them. I guess the answer to why they're not named consistently is that they're not part of the same family and don't do the same things.

`get_export` is so named because it's used specifically to start the export of a manifest. `post_consumer` is similar, in that it's used specifically to regenerate the identity cert, and is not a general 'post' method.

 Let me know your thoughts..

jeremylenz:
Addressed formatting suggestions.

app/lib/katello/resources/candlepin/upstream_consumer.rb
qcjames53: Thanks for the write up on the rationale and implementing the other changes, I appreciate it.

I'm feeling better about the naming as of now, but I still have some issues. This isn't something you started, but I'm not a fan of the Candlepin endpoint requests that have real generic-looking method parameters and log messages but they expect a certain format of data back. I'd be happier removing the web request type from the method name entirely if these are single-purpose methods.

Maybe `get_export` could be `start_upstream_export` and `post_consumer` could be `regenerate_upstream_identity`? Just some ideas. If I was working with this file and needed another get or post I would have been tempted to use these two methods as generics (and they might have worked); I think changing the names might help.

qcjames53: Or alternatively `get_export` and `post_consumer` could be refactored into general-purpose get and post methods. I think the first option is better but I'm happy with either tbh.

jeremylenz: > Maybe `get_export` could be `start_upstream_export` and `post_consumer` could be `regenerate_upstream_identity`?

sounds good to me! Let's go with that. :+1:

jeremylenz: I ran into a bump here:

```rb
def retrieve_owner_upstream_export(upstream, zip_file_path, export_id)
        validate_upstream_identity_cert!(upstream)
        data = Resources::Candlepin::UpstreamConsumer.start_upstream_export(
          "#{api_url(upstream)}#{upstream['uuid']}/export/#{export_id}",
          upstream['idCert']['cert'],
          upstream['idCert']['key'], ca_file
        )
        File.write(zip_file_path, data, mode: 'wb')

        true
      end
```

`start_upstream_export` doesn't make sense as a method name here. I'm going to alias the method; see what you think when I push the changes.

jeremylenz:
@qcjames53 Updated method names; see comments above

app/lib/katello/resources/candlepin/upstream_consumer.rb
qcjames53: Do we typically alias stuff like this? I'm not sure about best practices. Everything else looks great, thank you!

qcjames53: I guess what I'm asking is would it make more sense to choose a different name in this case?

jeremylenz: > would it make more sense to choose a different name for `start_upstream_export` in this case?

That's one option, but I can't think of anything besides `start_or_retrieve_upstream_export` which is cumbersome. That, or a more general name, like (wait for it) `get_export`. :smile:  

We do use `alias_method` a few other places in Katello. If we can't think of a better name, I think alias is fine personally.

qcjames53: Haha ok I think that sounds great then. Thanks Jeremy!

qcjames53 (approved):
Looks good to me 👍

PR #10938, opened 2024-03-15
Fixes #37268 - add repository_url back

ianballou:
Re-adds `repository_url` for liveimg provisioning (https://access.redhat.com/documentation/en-us/red_hat_satellite/6.14/html-single/provisioning_hosts/index#Using_an_Image_Builder_Image_for_Provisioning_provisioning) and makes it work with multiple cv environments.

Currently, two approaches are both coded in. I'd like feedback on which is better:

1) Force the user to use a single content view environment only when provisioning liveimg files. This is a safer and more simple option.

2) Allow the user to have multiple CVs on their host. If that is the case, search for the correct repository among the multiple CVs. If there are multiple root repositories with this file, throw an error.

Ideally, run through the steps in the docs above to provision a host using the template that uses this method.
Alternatively, simulate the provisioning by only rendering the script, or perhaps just run `repository_url` with the right inputs to trigger all of the errors.

To go through the host provisioning process, first sync some repo with kickstart content and create a domain. Then, create a host and fill in all the fields. Don't forget the domain and fake mac address on the interface. Lastly, add the parameter `kickstart_liveimg = some/path/to/a/file/that/does/not/need/to/exist.jpg`

Then build the host and look at the rendered kickstart default provisioning template. You can find it on the Details tab by clicking "Kickstart default".

Also test that `relative_path` in `repository_mapper.rb` still works since I moved one method from pulp glue to the repository model.

### DIFF
--- a/app/helpers/katello/katello_urls_helper.rb
+++ b/app/helpers/katello/katello_urls_helper.rb
@@ -37,11 +37,36 @@ module Katello
       "#{prefix}/pub/#{config}"
     end
 
-    apipie :method, 'Generates an absolute path to the file' do
+    apipie :method, 'Generates an absolute path to the file for liveimg kickstart templates' do
       required :content_path, String, desc: "Relative path to the file or it's name"
       optional :schema, String, desc: 'Optional URL schema for the content source', default: 'http'
       optional :content_type, String, desc: 'Content type', default: 'repos'
       returns String, desc: 'Absolute path to a file'
+    end
+    def repository_url(content_path, _content_type = nil, schema = 'http')
+      return content_path if content_path =~ %r|^([\w\-\+]+)://|
+      url = if @host.content_source
+              "#{schema}://#{@host.content_source.hostname}"
+            else
+              foreman_settings_url(schema)
+            end
+      content_path = content_path.sub(%r|^/|, '')
+
+      lifecycle_environment = nil
+      if @host.default_environment?
+        # Don't update the content_path if the host is using the default org view / library
+        lifecycle_environment = ::Katello::KTEnvironment.library.find_by(organization_id: @host.organization_id)
+      elsif !@host.single_content_view_environment?
+        fail ::Katello::Errors::MultiEnvironmentNotSupportedError,
+          "Host #{@host.name} must be subscribed to only a single content view & environment or subscribe to the default organization content view for liveimg provisioning."
+      else
+        lifecycle_environment = @host.single_lifecycle_environment
+        content_path = [@host.single_content_view.label, content_path].join('/')
+      end
+
+      path = ::Katello::Repository.repo_path_from_content_path(
+        lifecycle_environment, content_path)
+      "#{url}/pulp/content/#{path}"
     end
   end
 end
--- a/app/lib/katello/errors.rb
+++ b/app/lib/katello/errors.rb
@@ -10,6 +10,10 @@ module Katello
 
     class RegistrationError < StandardError; end
 
+    class InvalidRepositoryTypeError < StandardError; end
+
+    class MultiEnvironmentNotSupportedError < StandardError; end
+
     # unauthorized access
     class SecurityViolation < StandardError; end
 
--- a/app/models/katello/candlepin/repository_mapper.rb
+++ b/app/models/katello/candlepin/repository_mapper.rb
@@ -74,7 +74,7 @@ module Katello
       end
 
       def relative_path
-        ::Katello::Glue::Pulp::Repos.repo_path_from_content_path(product.organization.library, path)
+        ::Katello::Repository.repo_path_from_content_path(product.organization.library, path)
       end
 
       def prune_substitutions(subs, url)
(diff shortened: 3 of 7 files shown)


app/helpers/katello/katello_urls_helper.rb
```diff
@@ -37,11 +37,73 @@ def subscription_manager_configuration_url(host = nil, rpm = true, hostname: nil
...
+          lifecycle_environment = file_repo.first.environement
+          content_path = [file_repo.first.content_view.label, content_path].join('/')
+        end
+      else
+        # If the host is subscribed to the default org view, use Library.
+        lifecycle_environment = ::Katello::KTEnvironment.library.where(organization_id: @host.organization.id)
+      end
+
+      path = ::Katello::Glue::Pulp::Repos.repo_path_from_content_path(
+        lifecycle_environment, content_path)
```
sjha4, line 105: Can we move this out of glue while we're touching this and put this in repository model layer?

ianballou:
After a chat with @jeremylenz , we determined that we should wait to support mutli-cv until it's implemented further. Once content view environment prioritization is available, we can use that to choose a CV / LCE without the user needing to enter one.

So, for now, we will error out if the user is using multi CVs (unless they have default content view / library among them).

ianballou:
For reference, here is the old implementation that was removed:

```ruby
    def repository_url(content_path, _content_type = nil, schema = 'http')
      return content_path if content_path =~ %r|^([\w\-\+]+)://|
      url = if @host.content_source
              "#{schema}://#{@host.content_source.hostname}"
            else
              foreman_settings_url(schema)
            end
      content_path = content_path.sub(%r|^/|, '')
      if @host.content_view && !@host.content_view.default?
        content_path = [@host.content_view.label, content_path].join('/')
      end
      path = ::Katello::Glue::Pulp::Repos.repo_path_from_content_path(
        @host.lifecycle_environment, content_path)
      "#{url}/pulp/content/#{path}"
    end
```

app/helpers/katello/katello_urls_helper.rb
```diff
@@ -37,11 +37,38 @@ def subscription_manager_configuration_url(host = nil, rpm = true, hostname: nil
...
+      url = if @host.content_source
+              "#{schema}://#{@host.content_source.hostname}"
+            else
+              foreman_settings_url(schema)
+            end
+      content_path = content_path.sub(%r|^/|, '')
+
+      lifecycle_environment = nil
+      if @host.content_views.present?
+        if @host.content_views.to_a.intersection(::Katello::ContentView.default).present?
```
jeremylenz, line 57: There's a method that can simplify this:

https://github.com/Katello/katello/blob/679f08e6bffad977e0ffc70be4fc78e76582932a/app/models/katello/host/content_facet.rb#L128-L133
```suggestion
      if @host.default_environment?
```

ianballou: Fixed!

app/helpers/katello/katello_urls_helper.rb
```diff
@@ -37,11 +37,36 @@ def subscription_manager_configuration_url(host = nil, rpm = true, hostname: nil
...
+            else
+              foreman_settings_url(schema)
+            end
+      content_path = content_path.sub(%r|^/|, '')
+
+      lifecycle_environment = nil
+      if @host.default_environment?
+        # Don't update the content_path if the host is using the default org view / library
+        lifecycle_environment = ::Katello::KTEnvironment.library.find_by(organization_id: @host.organization_id)
+      elsif !@host.single_content_view_environment?
```
jeremylenz, line 59:
```suggestion
      elsif @host.multi_content_view_environment?
```

jeremylenz: and now that I'm here, I wonder if we should add a `!default_environment?` check to that method.. :thinking:

ianballou: I feel like there's good reason to check for single content view environment separately from checking the default environment.

jeremylenz: on second thought I agree.

jeremylenz (approved):
I also had to make sure the host had a `kt_activation_keys` parameter.

Before: Template displays "N/A" in the web UI and I see the error in the log:

```
[E|app|9cb31ee5] There was an error rendering the Kickstart default template: Safemode doesn't allow to access 'repository_url' on #<Safemode::ScopeObject>
```

After: Template renders correctly in UI :+1: 

Also registered a host to multiple non-default environments and confirmed that `repository_url` errors correctly:

```
[15] pry(main)> repository_url('bad/path', nil, 'https')
Katello::Errors::MultiEnvironmentNotSupportedError: Host rhel9.fedora.example.com must be subscribed to only a single content view & environment or subscribe to the default organization content view for liveimg provisioning.
from /home/vagrant/katello/app/helpers/katello/katello_urls_helper.rb:60:in `repository_url'
```

ACK :+1:

PR #10942, opened 2024-03-19
Fixes #37279 - Re-create remotes when the type needs to change

quba42:
When updating a remote, we may need to re-create it instead if the type has changed from uln to rpm or vice versa.

For operations on a remote href, the type is now determined from the href, rather than indirectly from the url. This is more robust since it is the actually relevant source of truth.

All other changes are for readability, deduplication, or consistency

### DIFF
--- a/app/services/katello/pulp3/api/core.rb
+++ b/app/services/katello/pulp3/api/core.rb
@@ -47,6 +47,11 @@ module Katello
           fail NotImplementedError
         end
 
+        # Method is called with either :url or :href parameters for the sake of yum content.
+        def get_remotes_api(*)
+          remotes_api
+        end
+
         def publications_api
           repository_type.publications_api_class.new(api_client) #Optional
         end
--- a/app/services/katello/pulp3/api/yum.rb
+++ b/app/services/katello/pulp3/api/yum.rb
@@ -32,6 +32,17 @@ module Katello
           PulpRpmClient::RemotesUlnApi.new(api_client)
         end
 
+        def get_remotes_api(href: nil, url: nil)
+          fail 'Provide exactly one of href or url for yum remote selection!' if url.blank? && href.blank?
+          fail 'The href must be a pulp_rpm remote href!' if href && !href.start_with?('/pulp/api/v3/remotes/rpm/')
+
+          if href&.start_with?('/pulp/api/v3/remotes/rpm/uln/') || url&.start_with?('uln')
+            remotes_uln_api
+          else
+            remotes_api
+          end
+        end
+
         def copy_api
           PulpRpmClient::RpmCopyApi.new(api_client)
         end
--- a/app/services/katello/pulp3/repository.rb
+++ b/app/services/katello/pulp3/repository.rb
@@ -91,16 +91,21 @@ module Katello
       end
 
       def remote_partial_update
-        if remote_options[:url]&.start_with?('uln')
-          api.remotes_uln_api.partial_update(repo.remote_href, remote_options)
-        else
-          api.remotes_api.partial_update(repo.remote_href, remote_options)
+        url_type = remote_options[:url]&.start_with?('uln') ? 'uln' : 'default'
+        remote_type = repo.remote_href.start_with?('/pulp/api/v3/remotes/rpm/uln/') ? 'uln' : 'default'
+        href = repo.remote_href
+
+        if url_type == remote_type
+          api.get_remotes_api(href: href).partial_update(href, remote_options)
+        else # We need to recreate a remote of the correct type!
+          create_remote
+          delete_remote(href: href)
         end
       end
 
       def delete_remote(options = {})
         options[:href] ||= repo.remote_href
-        ignore_404_exception { remote_options[:url]&.start_with?('uln') ? api.remotes_uln_api.delete(options[:href]) : api.remotes_api.delete(options[:href]) } if options[:href]
+        ignore_404_exception { api.get_remotes_api(href: options[:href]).delete(options[:href]) } if options[:href]
       end
 
       def self.instance_for_type(repo, smart_proxy)
@@ -139,7 +144,7 @@ module Katello
       end
 
       def get_remote(href = repo.remote_href)
-        repo.url&.start_with?('uln') ? api.remotes_uln_api.read(href) : api.remotes_api.read(href)
+        api.get_remotes_api(href: href).read(href)
       end
 
       def get_distribution(href = distribution_reference.href)
--- a/app/services/katello/pulp3/repository_mirror.rb
+++ b/app/services/katello/pulp3/repository_mirror.rb
@@ -24,11 +24,9 @@ module Katello
       def refresh_entities
         href = remote_href
         if href
-          if remote_options[:url]&.start_with?('uln')
-            [api.remotes_uln_api.partial_update(href, remote_options)]
-          else
-            [api.remotes_api.partial_update(href, remote_options)]
-          end
+          # Do not consider remotes_uln_api, since the Katello server is not a ULN server. Even if the sync
+          # to Katello used ULN, the sync from Katello server to smart proxy will use a normal RPM remote!
+          [api.remotes_api.partial_update(href, remote_options)]
         else
           create_remote
           []
@@ -111,13 +109,10 @@ module Katello
       end
 
       def create_remote
-        if remote_options[:url]&.start_with?('uln')
-          remote_file_data = @repo_service.api.class.remote_uln_class.new(remote_options)
-          api.remotes_uln_api.create(remote_file_data)
-        else
-          remote_file_data = @repo_service.api.remote_class.new(remote_options)
-          api.remotes_api.create(remote_file_data)
-        end
+        # Do not consider remotes_uln_api, since the Katello server is not a ULN server. Even if the sync
+        # to Katello used ULN, the sync from Katello server to smart proxy will use a normal RPM remote!
+        remote_file_data = @repo_service.api.remote_class.new(remote_options)
+        api.remotes_api.create(remote_file_data)
       end
 
       def compute_remote_options
--- a/test/services/katello/pulp3/repository/file/update_remote_test.rb
+++ b/test/services/katello/pulp3/repository/file/update_remote_test.rb
@@ -20,7 +20,7 @@ module Katello
           @file_repo_service = @file_repo.backend_service(@mock_smart_proxy)
           @file_repo.root.update(url: 'my-files.org')
           @file_repo_service.stubs(:api).returns(@mock_api_wrapper)
-          @mock_api_wrapper.stubs(:remotes_api).returns(@mock_pulp3_api)
+          @mock_api_wrapper.stubs(:get_remotes_api).returns(@mock_pulp3_api)
 
           @file_repo.remote_href = '193874298udsfsdf'
           refute_empty @file_repo.remote_href
--- a/app/services/katello/pulp3/alternate_content_source.rb
+++ b/app/services/katello/pulp3/alternate_content_source.rb
@@ -88,18 +88,16 @@ module Katello
       end
 
       def get_remote(href = smart_proxy_acs.remote_href)
-        acs.base_url&.start_with?('uln') ? api.remotes_uln_api.read(href) : api.remotes_api.read(href)
+        api.get_remotes_api(href: href).read(href)
       end
 
-      def update_remote
-        api.remotes_api.partial_update(smart_proxy_acs.remote_href, remote_options)
+      def update_remote(href = smart_proxy_acs.remote_href)
+        api.get_remotes_api(href: href).partial_update(href, remote_options)
       end
 
-      # The old repo URL is needed to determine which remote API to use.
       def delete_remote(options = {})
         options[:href] ||= smart_proxy_acs.remote_href
-        options[:old_url] ||= remote_options[:url]
-        ignore_404_exception { options[:old_url]&.start_with?('uln') ? api.remotes_uln_api.delete(options[:href]) : api.remotes_api.delete(options[:href]) } if options[:href]
+        ignore_404_exception { api.get_remotes_api(href: options[:href]).delete(options[:href]) } if options[:href]
       end
 
       def create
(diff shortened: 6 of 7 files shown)


app/services/katello/pulp3/repository.rb
```diff
@@ -91,16 +91,21 @@ def update_remote
...
-          api.remotes_api.partial_update(repo.remote_href, remote_options)
+        url_type = remote_options[:url]&.start_with?('uln') ? 'uln' : 'default'
+        remote_type = repo.remote_href.start_with?('/pulp/api/v3/remotes/rpm/uln/') ? 'uln' : 'default'
+        href = repo.remote_href
+
+        if url_type == remote_type
+          api.get_remotes_api(href: href).partial_update(href, remote_options)
+        else # We need to recreate a remote of the correct type!
+          create_remote
+          delete_remote(href: href)
```
quba42, line 102: FYI: This is what actually fixes the bug.

All the rest of the changes are essentially a refactor aimed at improving readability and robustness, using the correct source of truth for the remote type (in most cases the remote href itself), and also de-duplicating the remote API selection code.

app/services/katello/pulp3/api/core.rb
```diff
@@ -47,6 +47,10 @@ def remotes_uln_api
           fail NotImplementedError
         end
 
+        def get_remotes_api(*)
```
m-bucher, line 50: I am thinking how to communicate that we expect a hash-parameter containing either `:href` or `:url`, here.
Would a comment be sufficient or do we want to specify the parameter list as required.
The latter will probably make rubocop complain about it.

m-bucher (changes requested):
Overall pretty nice, the `get_remotes_api()` really cleans up the handling of ULN-remotes.

app/services/katello/pulp3/api/core.rb
quba42: I would argue, that in this parent class, we don't expect any particular parameters here.

The yum type version of this method now start with:
```
        def get_remotes_api(opts = {})
          fail 'Provide exactly one of href or url for yum remote selection!' unless !opts[:href] != !opts[:url]
```
This code makes pretty clear what type of arguments we expect in the yum case.

So maybe we can leave things as is? In case developers working on other content types wonder why we need the core version of this method at all, they can always use `git grep 'def get_remotes_api'` or `git grep get_remotes_api` to study up.

m-bucher: > I would argue, that in this parent class, we don't expect any particular parameters here.

The problem is other content-types that might need to implement their own `get_remotes_api()` are not free to choose different parameters, because the method is called with `:href` or `:url` in the generic code (like app/services/katello/pulp3/repository.rb).

Therefore, the parent-method must IMHO define what can be evaluated. It is always possible to look for this in existing code that uses it or defines child-methods, but the first location I would look for is the method-definition in the parent-class.

quba42: If another content type might want to add a second remote type some time in the future, there is no guarantee that that content types discriminator for choosing the remote type would also be the URL. It follows that there would likely need to be some refactoring/extending at that point. This coupled with the fact that `*` seems semantically correct to me (for the parent class), I feel like the best way to address your concern might really be a comment. Something along the lines of:
```
# This function is currently called with either opts[:url] or opts[:href] parameters for the benefit of the yum content type.
```
Or maybe simply:
```
# See the yum version of this function to see how it can be used.
```
What do you think?

app/services/katello/pulp3/repository_mirror.rb
```diff
@@ -24,11 +24,8 @@ def backend_object_name
       def refresh_entities
         href = remote_href
         if href
-          if remote_options[:url]&.start_with?('uln')
-            [api.remotes_uln_api.partial_update(href, remote_options)]
-          else
-            [api.remotes_api.partial_update(href, remote_options)]
-          end
+          # Do not consider remotes_uln_api, since the Katello server will never serve ULN!
+          [api.remotes_api.partial_update(href, remote_options)]
```
quba42, line 28: Even though it is not really necessary we could consider using `get_remotes_api` here anyway, just to keep things more generic so this won't be overlooked in case the `get_remotes_api` use case is extended further in the future. I would then use:
```suggestion
          [api.get_remotes_api(href: href).partial_update(href, remote_options)]
```
Any opinions? I am unsure what is best.

app/services/katello/pulp3/api/core.rb
quba42: Added a comment.

app/services/katello/pulp3/repository_mirror.rb
m-bucher: Since for rpm/uln it should be correct (because it is created correctly), I guess we can do that.

m-bucher (approved):

ianballou:
A Katello maintainers should get to reviewing this this week.

ianballou:
@quba42 do you have a `uln` repo URL that I can test with?

app/services/katello/pulp3/repository_mirror.rb
```diff
@@ -24,11 +24,8 @@ def backend_object_name
       def refresh_entities
         href = remote_href
         if href
-          if remote_options[:url]&.start_with?('uln')
-            [api.remotes_uln_api.partial_update(href, remote_options)]
-          else
-            [api.remotes_api.partial_update(href, remote_options)]
-          end
+          # Do not consider remotes_uln_api, since the Katello server will never serve ULN!
```
ianballou, line 27: What does "...since the Katello server will never serve ULN" mean? Are ULN repos not supported on smart proxies?

ianballou (commented):
Overall this is working well! Just one more question, and I'd like input on my other comment from before about the args for `get_remotes_api`.

app/services/katello/pulp3/repository_mirror.rb
quba42: "ULN repo" just means an RPM repo with a specific type of authentication mechanism on it. Katello will sync the ULN repo to the Katello server, and then publish a normal Pulp RPM repo on the Katello server. This normal RPM repo is then synced to smart proxy the normal way. The sync from Katello server to smart proxy will never use a ULN URL. Therefore it will never use a ULN remote. We will never implement ULN authentication for the repos served in pulp_rpm/Katello.

app/services/katello/pulp3/api/yum.rb
```diff
@@ -32,6 +32,17 @@ def remotes_uln_api
           PulpRpmClient::RemotesUlnApi.new(api_client)
         end
 
+        def get_remotes_api(opts = {})
```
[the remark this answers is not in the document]
quba42: This makes sense to me. I will try to implement it today.

app/services/katello/pulp3/repository_mirror.rb
quba42: I suppose checking the URL that is synced to the proxy for ULN would not hurt, it's just that test will always come up: "not ULN".

quba42: Would my comment be more clear with something like:
```suggestion
          # Do not consider remotes_uln_api, since the Katello server is not a ULN server. Even if the sync to Katello used ULN, the sync from Katello server to smart proxy will use a normal RPM remote!
```

app/services/katello/pulp3/api/yum.rb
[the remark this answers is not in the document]
quba42: Made and tested the relevant adjustments.

app/services/katello/pulp3/repository_mirror.rb
ianballou: Thanks for the clarification, makes sense now.  I didn't realize that the ULN remote pretty much existed just for their special authentication. 

I think the updated comment would be helpful.

app/services/katello/pulp3/repository_mirror.rb
ianballou: I like using `remotes_api` specifically if this case should always use the Rpm remote.

ianballou: I think it makes it more obvious that there is a special case here -- that ULN remotes will never be used on smart proxies.

test/services/katello/pulp3/repository/file/update_remote_test.rb
```diff
@@ -20,7 +20,7 @@ def setup
           @file_repo_service = @file_repo.backend_service(@mock_smart_proxy)
           @file_repo.root.update(url: 'my-files.org')
           @file_repo_service.stubs(:api).returns(@mock_api_wrapper)
-          @mock_api_wrapper.stubs(:remotes_api).returns(@mock_pulp3_api)
+          @mock_api_wrapper.stubs(:get_remotes_api).returns(@mock_pulp3_api)
```
ianballou, line 23: Perhaps this entire method shouldn't be stubbed out so it gets tested? `remotes_api` should still be stub-able.

Maybe something like  `blah_repo_service.expects(:remotes_api).once.returns(@mock_pulp3_api)`

I realize in this file only `File` repository types are tested. A new test for yum/uln could be in order :)

ianballou (commented):
Still working, left some code comments.

app/services/katello/pulp3/repository_mirror.rb
quba42: Then I will leave this as is. Except for the improved comment discussed below.

app/services/katello/pulp3/repository_mirror.rb
quba42: I may have oversimplified a little bit, but overall the "special authentication" mental model has served me fine.

I updated the comment.

test/services/katello/pulp3/repository/file/update_remote_test.rb
quba42: If I allow it to run `get_remotes_api` (revert this change), I get:
```
[vagrant@centos8-katello-devel-stable2 foreman]$ ktest ../katello/test/services/katello/pulp3/repository/file/update_remote_test.rb
/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/activerecord-6.1.7.7/lib/active_record/connection_adapters/postgresql_adapter.rb:883: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/home/vagrant/foreman/.vendor/ruby/2.7.0/gems/pg-1.5.6/lib/pg/text_decoder/timestamp.rb:8: warning: The called method `initialize' is defined here
Emptying /home/vagrant/foreman/jenkins/reports/unit
   -> 0.0009s

# Running tests with run options --seed 65021:

FF.Writing XML reports to /home/vagrant/foreman/jenkins/reports/unit


Finished tests in 2.028559s, 1.4789 tests/s, 7.3944 assertions/s.


Failure:
Katello::Service::Repository::UpdateRemoteTest#test_feed_url_exists_and_remote_href_exists_updates_remote [/home/vagrant/katello/app/services/katello/pulp3/repository.rb:99]
Minitest::Assertion: unexpected invocation: #<Mock:api_wrapper>.get_remotes_api(:href => "193874298udsfsdf")
unsatisfied expectations:
- expected exactly once, invoked never: #<Mock:pulp3_api>.partial_update(any_parameters)
satisfied expectations:
- allowed any number of times, invoked never: #<AnyInstance:Resolv::DNS>.getaddresses(any_parameters)
- allowed any number of times, invoked never: #<AnyInstance:Resolv::DNS>.getaddress(any_parameters)
- allowed any number of times, invoked never: #<AnyInstance:Resolv::DNS>.getnames(any_parameters)
- allowed any number of times, invoked never: #<AnyInstance:Resolv::DNS>.getname(any_parameters)
- allowed any number of times, invoked never: #<AnyInstance:Katello::Pulp3::Repository>.test_remote_name(any_parameters)
- allowed any number of times, invoked never: Katello::Ping.ping(any_parameters)
- allowed any number of times, invoked never: Cert::Certs.ssl_client_key(any_parameters)
- allowed any number of times, invoked never: Cert::Certs.ssl_client_cert(any_parameters)
- allowed any number of times, invoked never: Cert::Certs.ca_cert(any_parameters)
- allowed any number of times, invoked never: #<Mock:response>.pulp_href(any_parameters)
- allowed any number of times, invoked never: #<Mock:api_wrapper>.remotes_api(any_parameters)
- allowed any number of times, invoked never: #<Mock:pulp3_api>.create(any_parameters)
- allowed any number of times, invoked never: #<Mock:smart_proxy>.pulp_primary?(any_parameters)
- allowed any number of times, invoked never: #<Mock:smart_proxy>.pulp2_preferred_for_type?(any_parameters)
- allowed any number of times, invoked once: #<Mock:smart_proxy>.pulp3_support?(any_parameters)
- allowed any number of times, invoked once: #<Katello::Pulp3::Repository::File:0x249c8>.api(any_parameters)


Failure:
Katello::Service::Repository::UpdateRemoteTest#test_feed_url_is_missing_but_remote_href_exists_deletes_remote [/home/vagrant/katello/app/services/katello/pulp3/service_common.rb:44]
Minitest::Assertion: unexpected invocation: get_remotes_api({:href=>"193874298udsfsdf"})

3 tests, 15 assertions, 2 failures, 0 errors, 0 skips

Minitest Reporters: Mean Time Report (Samples: 4, Order: :avg :desc)
Avg: 2.07335646   Min: 1.994977516  Max: 2.256206166  Last: 2.026355225  Description: Katello::Service::Repository::UpdateRemoteTest
```

quba42: Adding entirely new tests for additional content types would make a lot of sense.

I will put it on the [agenda](https://hackmd.io/U9Q0Js0bRPez7oCgxUawng) for our upcoming meeting.

ianballou:
```
- expected exactly once, invoked never: #<Mock:pulp3_api>.partial_update(any_parameters)
```
Is there a bit in your code that is expecting `partial_update` to run?

ianballou: You should be able to set it up so at least you can run get_remotes_api and stub/expect the internal method calls instead.

ianballou:
@quba42 just let me know if you're ready for this to merge and I'll do so.

ianballou:
Merging based on a conversation today where the tests around RPM remote creation will be bolstered outside this PR.

PR #10943, opened 2024-03-19
Fixes #37195 - Remove dynflow actions for profile uploads

jeremylenz:

Remove Dynflow actions
* `::Actions::Katello::Host::UploadPackageProfile`
* `::Actions::Katello::Host::UploadProfiles`

Also remove a Setting, `upload_profiles_without_dynflow`, which defaulted to true and has never been known to be needed since its debut.

The reason there are two actions is because Katello supports two endpoints for uploading profiles.

- The legacy one is `:PUT, "/consumers/:id/packages"` and `:PUT, "/consumers/:id/profile"` and uses `::Katello::Host::PackageProfileUploader`.
- The current one is `:PUT, "/consumers/:id/profiles"` and uses `::Katello::Host::ProfilesUploader`. This endpoint also supports modules.

I believe we can remove the legacy endpoint and `::Katello::Host::PackageProfileUploader` when RHEL7 support ends.

pretty much just make sure the tests pass. You could also test some package profile uploads just to make sure everything continues to work.

```
# EL8+
subscription-manager repos # uploads package profile if there have been any changes
dnf uploadprofile --force-upload # always uploads package profile

# EL7
subscription-manager repos # uploads package profile if there have been any changes
katello-enabled-repos-upload -f # uploads package profile if katello-host-tools is installed
```


sbernhard:
One setting to be removed. This is good. Is there a documentation change necessary? Were there reasons, why user want to run the package profile upload in a dynflow task?

jeremylenz:
> Is there a documentation change necessary? Were there reasons, why user want to run the package profile upload in a dynflow task?

I don't think a doc change is needed since this setting isn't documented anyway.

The only reason a user would want to do this would be if Puma couldn't handle the sheer volume of package profile uploads, which is a thing we thought might happen when it was designed but it never actually happened. We added the setting in https://bugzilla.redhat.com/show_bug.cgi?id=1771724 around Katello 4.2 and made profile uploads Dynflow-less by default, and we've had exactly zero complaints and bug reports since then, so I think this removal is pretty safe.

chris1984 (approved):
ACK, code looks good.

Tested with EL7 and EL8 and didn't see any tracebacks when running the commands you posted, so I would say this is good to go.

PR #10947, opened 2024-03-22
Drop anemone and use Spidr for Repo discovery

sjha4:
To Do:

- [x] Support basic auth
- [x] Support proxy

bundle install
Go to Content > Product > Repo discovery
Run repo discovery.


sjha4:
@evgeni : Thoughts on [spidr gem](https://github.com/postmodern/spidr/) as replacement for anemone?

evgeni:
Doesn't look crazy? Only dep is nokogiri, which we have anyway, tested on modern rubies. Why not.

app/lib/katello/repo_discovery.rb
```diff
@@ -172,14 +174,70 @@ def http_crawl(resume_point)
...
+
+    def spidr_crawl(url)
+      user = @upstream_username if @upstream_username
+      password = @upstream_password if @upstream_password
+      #Spidr crawl the url over every url and add urls with repodata to found
+      Spidr.site(url, proxy: spidr_proxy_details) do |spider|
+        spider.authorized.add(url, user, password)
+        spider.every_url do |url|
+          if @found.include?(url.to_s.split('repodata/').first)
+            puts ">>> Skipping #{url.to_s}"
```
jeremylenz, line 206: These should probably be `Rails.logger.debug`s, no?

sjha4: Ya..I will clean this up and move it to ready for review.. 😻

sjha4:
I am seeing significant performance difference with what's on the PR vs the existing workflow..Looking at ways to speed this up..Will push updates when I get the performance sorted.

app/lib/katello/repo_discovery.rb
```diff
@@ -172,14 +174,70 @@ def http_crawl(resume_point)
...
+            puts ">>> Found #{url.to_s.split('repodata/').first}"
+          end
+        end
+      end
+    end
+
+    def spidr_crawl_url(url)
+      Spidr.site(url) do |spider|
+        spider.every_page do |page|
+          next if page.url.to_s.include?('bigzoo') || page.url.to_s.include?('ostree') || page.url.to_s.include?('?')
```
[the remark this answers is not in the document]
sjha4: This was junk code I was playing with when testing with a specific repo..Removed all of this..

app/lib/katello/repo_discovery.rb
```diff
@@ -8,8 +9,8 @@ class RepoDiscovery
 
     # rubocop:disable Metrics/ParameterLists
     def initialize(url, content_type = 'yum', upstream_username = nil,
-      upstream_password = nil, search = '*', crawled = [],
-      found = [], to_follow = [])
+                   upstream_password = nil, search = '*', crawled = [],
+                   found = [], to_follow = [])
```
evgeni, line 13: The indentation seems off here

katello.gemspec
```diff
@@ -49,6 +49,7 @@ Gem::Specification.new do |gem|
 
   # Pulp
   gem.add_dependency "anemone"
```
evgeni, line 51: Can we drop anemone and webrick deps now?

sjha4: Removed anemone.. 🎉

app/lib/katello/repo_discovery.rb
```diff
@@ -142,40 +143,38 @@ def docker_search
...
+          @crawled << page.url.to_s
+          process_page_urls(page.urls)
+          spider.skip_page!
+        end
+      end
+    end
+
+    def process_page_urls(urls)
+      urls.each do |url|
+        url = url.to_s
```
ekohl, line 173: I think this is a bad idea. The URL object is way more valuable and I'd only add it to `@to_follow` with `to_s`.

app/lib/katello/repo_discovery.rb
```diff
@@ -194,13 +193,14 @@ def file_crawl(resume_point)
...
-      # * link's path starts with the base url
+      # * link's path includes the base url
       # * link hasn't already been crawled
       # * link ends with '/' so it should be a directory
       # * link doesn't end with '/Packages/', as this increases
       #       processing time and memory usage considerably
-      return path.starts_with?(@uri.path) && !@crawled.include?(path) &&
-           path.ends_with?('/') && !path.ends_with?('/Packages/')
+      #
+      return path.include?(@uri.hostname) && !@crawled.include?(path) &&
```
ekohl, line 202: Looks like the `path` is really the `url`. Please change the name, but better: pass it as a `URI` object and then compare `uri.hostname` to `@uri.hostname`. Also note that hostnames are case insensitive and plain string comparisons are wrong.

app/lib/katello/repo_discovery.rb
```diff
@@ -142,40 +143,38 @@ def docker_search
...
+          process_page_urls(page.urls)
+          spider.skip_page!
+        end
+      end
+    end
+
+    def process_page_urls(urls)
+      urls.each do |url|
+        url = url.to_s
+        if url.include? 'repodata/'
```
ekohl, line 174: This should only look at `url.path`. And previously it was `ends_with?`. Now it will also find files within. Is that desirable?

katello.gemspec
```diff
@@ -47,12 +47,10 @@ Gem::Specification.new do |gem|
 
   gem.add_dependency "pg"
 
-  # Pulp
-  gem.add_dependency "anemone"
-  # required by anemone: https://github.com/chriskite/anemone/blob/next/lib/anemone/page.rb#L3
-  # webrick is no longer a part of Ruby's stdlib: https://bugs.ruby-lang.org/issues/17303
-  gem.add_dependency "webrick"
-  #pulp3
+  #Required for repo discovery
```
ekohl, line 50:
```suggestion
  # Required for repo discovery
```

katello.gemspec
```diff
@@ -47,12 +47,10 @@ Gem::Specification.new do |gem|
...
+  gem.add_dependency "spidr"
+
+  #Pulp dependencies
```
ekohl, line 53:
```suggestion
  # Pulp dependencies
```

ekohl (commented):
Implementation wise I think you should separate the crawler and Docker search into separate classes. Perhaps even the file crawl as well. Right now it's confusing.

app/lib/katello/repo_discovery.rb
sjha4: We don't necessarily use the URL object properties once we have the list of URLs in the page. It's just matching strings after that point..Becomes easier to pass that around and work with it IMO?

sjha4: Updated it in latest push to pass URL object.

app/lib/katello/resources/discovery/file_discovery.rb
```diff
@@ -0,0 +1,34 @@
+module Katello
+  class FileDiscovery < RepoDiscovery
+    attr_reader :found, :crawled, :to_follow
+    def initialize(url, upstream_username = nil,
+                   upstream_password = nil, search = '*', crawled = [],
+                   found = [], to_follow = [])
+      @uri = uri(url)
+      @upstream_username = upstream_username.empty? ? nil : upstream_username
+      @upstream_password = upstream_password.empty? ? nil : upstream_password
+      @search = search
```
ekohl, line 10: These parameters are unused. Perhaps drop them? If not, then should the constructor be part of the base class?

app/lib/katello/resources/discovery/file_discovery.rb
```diff
@@ -0,0 +1,34 @@
...
+      @to_follow = to_follow
+    end
+
+    def run(resume_point)
+      file_crawl(resume_point)
+    end
+
+    private
+
+    def file_crawl(resume_point)
```
ekohl, line 22: Since there is no recursion in the method, perhaps just make this `run(resume_point)`?

ekohl (commented):
I think this heads in the right direction.

Review wise I wonder if it makes sense to split it up: 1 to create the 3 classes and a follow up that replaces anemone with spidr.

app/lib/katello/resources/discovery/file_discovery.rb
sjha4: I will drop these here..
I was trying to get the constructor as part of base class but the foreman task output which is read in chunks doesn't work when inherited for some reason. Had to go with constructors in child classes and the class_for construct.

sjha4: Updated to make credentials and search optional additional params for container and yum classes.

app/lib/katello/resources/discovery/file_discovery.rb
sjha4: Updated to remove this extra method.

app/lib/katello/resources/discovery/container_discovery.rb
```diff
@@ -0,0 +1,123 @@
+module Katello
+  class ContainerDiscovery < RepoDiscovery
+    attr_reader :found, :crawled, :to_follow
+    def initialize(url, crawled = [], found = [], to_follow = [],
+                   upstream_credentials_and_search = {
+                     upstream_username: nil,
+                     upstream_password: nil,
+                     search: '*'
```
ekohl, line 8: This will be `nil` if `upstream_credentials_and_search` is overridden by a caller. If you want it to default to `*` you can use `@search = upstream_credentials_and_search.fetch(:search, '*')`.

app/lib/katello/resources/discovery/yum_discovery.rb
```diff
@@ -0,0 +1,73 @@
...
+  class YumDiscovery < RepoDiscovery
+    attr_reader :found, :crawled, :to_follow
+    def initialize(url, crawled = [], found = [], to_follow = [],
+                   upstream_credentials_and_search = {
+                     upstream_username: nil,
+                     upstream_password: nil
+                   })
+      @uri = uri(url)
+      @upstream_username = upstream_credentials_and_search[:upstream_username].empty? ? nil : upstream_credentials_and_search[:upstream_username]
+      @upstream_password = upstream_credentials_and_search[:upstream_password].empty? ? nil : upstream_credentials_and_search[:upstream_password]
```
ekohl, line 11: I know it was already this way, but isn't this equivalent?
```suggestion
      @upstream_username = upstream_credentials_and_search[:upstream_username].presence
      @upstream_password = upstream_credentials_and_search[:upstream_password].presence
```

app/lib/katello/repo_discovery.rb
```diff
@@ -1,206 +1,33 @@
 require 'uri'
+require 'spidr'
```
ekohl, line 2: I think this should live in the yum class

ekohl (commented):
Review wise I always prefer to first have a refactor and then an actual code change. Right now if I read the commits then it's the other way around. That makes this whole change harder to follow, but I don't have commit permissions here so I'll let other reviewers weigh in on that aspect.

app/lib/katello/repo_discovery.rb
```diff
@@ -1,206 +1,33 @@
...
-      @content_type = content_type
-      @upstream_username = upstream_username.empty? ? nil : upstream_username
-      @upstream_password = upstream_password.empty? ? nil : upstream_password
-      @search = search
-      @found = found
-      @crawled = crawled
-      @to_follow = to_follow
+    def self.class_for(content_type)
+      case content_type
+      when 'file'
```
parthaa, line 14: Can we leverage this => https://github.com/Katello/katello/tree/master/lib/katello/repository_types 
Let the content type tell you the discovery class

lib/katello/repo_discovery.rb
```diff
@@ -0,0 +1,3 @@
+Dir["#{File.expand_path('../../../app/lib/katello/resources/discovery', __FILE__)}/*.rb"].each do |file|
```
ekohl, line 1: You don't need `expand_path`, but I wonder why you even need this `load` in the first place. I think it's very much incompatible with the Zeitwerk autoloader.

app/lib/katello/resources/discovery/container_discovery.rb
```diff
@@ -0,0 +1,123 @@
+module Katello
+  class ContainerDiscovery < RepoDiscovery
```
ekohl, line 2: I think for the autoloader to work this needs to be `Katello::Resources::Discovery::ContainerDiscovery`. Then given the repetition I'd suggest to rename this to `Katello::Resources::Discovery::Container`.

app/lib/katello/repo_discovery.rb
```diff
@@ -4,203 +4,17 @@ module Katello
...
-      found = [], to_follow = [])
-      @uri = uri(url)
-      @content_type = content_type
-      @upstream_username = upstream_username.empty? ? nil : upstream_username
-      @upstream_password = upstream_password.empty? ? nil : upstream_password
-      @search = search
-      @found = found
-      @crawled = crawled
-      @to_follow = to_follow
+    def self.class_for(content_type = 'yum')
```
ekohl, line 7: I wouldn't assume a content type and make it a mandatory argument
```suggestion
    def self.class_for(content_type)
```

app/lib/katello/repo_discovery.rb
```diff
@@ -4,203 +4,17 @@ module Katello
...
+      repo_discovery_class = RepositoryTypeManager.find_repository_type(content_type)&.repo_discovery_class
+      fail _ "Content type does not support repo discovery" unless repo_discovery_class
```
ekohl, line 9: is `_` a method call? If so, I'd prefer braces because it's very easy to miss that it's a method call
```suggestion
      fail _("Content type does not support repo discovery") unless repo_discovery_class
```

app/lib/katello/resources/discovery/container_discovery.rb
```diff
@@ -0,0 +1,123 @@
...
+    attr_reader :found, :crawled, :to_follow
+    def initialize(url, crawled = [], found = [], to_follow = [],
+                   upstream_credentials_and_search = {
+                     upstream_username: nil,
+                     upstream_password: nil,
+                     search: '*'
+                   })
+      @uri = uri(url)
+      @upstream_username = upstream_credentials_and_search[:upstream_username].empty? ? nil : upstream_credentials_and_search[:upstream_username]
+      @upstream_password = upstream_credentials_and_search[:upstream_password].empty? ? nil : upstream_credentials_and_search[:upstream_password]
```
ekohl, line 12: Like elsewhere, you can use `.presence`
```suggestion
      @upstream_username = upstream_credentials_and_search[:upstream_username].presence
      @upstream_password = upstream_credentials_and_search[:upstream_password].presence
```

app/lib/katello/resources/discovery/container_discovery.rb
```diff
@@ -0,0 +1,123 @@
...
+
+    def docker_search
+      request_params = {
+        method: :get,
+        headers: { accept: :json },
+        url: "#{@uri}v1/search?q=#{@search}"
+      }
+
+      request_params[:user] = @upstream_username unless @upstream_username.empty?
+      request_params[:password] = @upstream_password unless @upstream_password.empty?
```
ekohl, line 99: It's already `nil` or a non-empty string (constructor guarantees that) so why not just:
```suggestion
      request_params[:user] = @upstream_username if @upstream_username
      request_params[:password] = @upstream_password if @upstream_password
```

test/lib/repo_discovery_test.rb
```diff
@@ -58,13 +58,15 @@ def test_docker_with_v1_search_no_proxy
...
       found = []
       to_follow = [base_url]
-      search = 'busybox'
+      upstream_credentials_and_search = {
+        search: 'busybox'
+      }
 
       RestClient::Request.expects(:execute)
-        .with({ method: :get, url: base_url.to_s + "v1/search?q=#{search}", headers: {:accept => :json} })
+        .with({ method: :get, url: base_url.to_s + "v1/search?q=#{upstream_credentials_and_search[:search]}", headers: {:accept => :json} })
```
ekohl, line 66: I'd just hardcode this because it's a better guarantee. If something modifies the hash in place you can silently end up with `q=` instead of `q=busybox`. Previously with a string that could also happen, but it was less likely.
```suggestion
        .with({ method: :get, url: base_url.to_s + "v1/search?q=busybox", headers: {:accept => :json} })
```

ekohl (changes requested):

app/lib/katello/resources/discovery/container_discovery.rb
sjha4: Aah..Ya..That works..The earlier manual loading was cause the file wasn't getting loaded..Removed that and renamed this..

sjha4:
@ekohl : Updated the PR to address reviews..

lib/katello/repo_discovery.rb
sjha4: That was because of what you said in the review below around naming the modules correctly. It loads now without this explicit loading.

evgeni:
@sjha4 will you tackle the packaging side of this, or shall I?

sjha4:
@evgeni : I will give it a try and you can rescue me from time to time..Good learning opportunity cause this is not an urgent thing..

evgeni:
Sounds good to me!

evgeni:
(just don't merge this PR until the packaging is done, otherwise it'll block nightlies)

parthaa (approved):
Tested with

- Http
- Http With basic auth
- Container images
- local katello instance

Worked great!. ACK


image

https://en.wikipedia.org/wiki/Spyder_(film)

sjha4:
Will rebase and push for tests and merge once packaging is in..

ekohl (approved):
Packaging will fail until this is merged. Technically you can go through the complex dance of making it work, but then you need to revert all those changes again when you merge it. I'd suggest to merge this.

PR #10949, opened 2024-03-26
Pulpcore upgrade

sjha4:
Related pulpcore upgrade PRs:

https://github.com/theforeman/foreman-packaging/pull/10658
https://github.com/theforeman/pulpcore-packaging/pull/931
https://github.com/theforeman/foreman-packaging/pull/10667

Update pulp bindings and VCRs
On a box with upgraded pulp packages, check this out and run through some smoke testing.


ianballou (approved):
Looks good! Just need to wait to merge until the packaging PRs (like https://github.com/theforeman/foreman-packaging/pull/10658) are ready.

sjha4:
Added teardown code to some of our tests which didn't have those. Also added code path to look up dangling distributions during orphan cleanup and version deletion to resolve this: https://projects.theforeman.org/issues/37314

app/services/katello/pulp3/smart_proxy_repository.rb
```diff
@@ -33,15 +33,45 @@ def current_repositories(environment_id = nil, content_view_id = nil)
         katello_repos.select { |repo| repo_ids.include? repo.pulp_id }
       end
 
+      def delete_dangling_distributions
+        distribution_deletion_tasks = []
+        distro_hrefs = []
+        orphan_repository_versions.each do |api, version_hrefs|
+          next if api.repository_type.id == :deb
```
ianballou, line 40: Is there no workaround for Debian?

app/services/katello/pulp3/smart_proxy_repository.rb
```diff
@@ -33,15 +33,45 @@ def current_repositories(environment_id = nil, content_view_id = nil)
...
+            end
+            distro_hrefs.flatten!
+            distribution_deletion_tasks << distro_hrefs.collect do |href|
+              api.delete_distribution(href)
+            end
+
+          else
+            distribution_deletion_tasks << version_hrefs.collect do |href|
+              api.publications_api.list({repository_version: href}).results.collect do |publication|
+                api.distributions_api.list({publication: publication.pulp_href}).results.collect do |distribution|
```
ianballou, line 57: I don't think any Pulp plugin API supports searching by publication. When I try it by hand, I'm seeing random things get returned that don't make sense. For example, I synced a Python repo twice and ran orphan cleanup. The distribution for the latest publication got deleted for some reason.

https://docs.pulpproject.org/pulp_python/restapi.html#tag/Distributions:-Pypi/operation/distributions_python_pypi_list

Even RPM doesn't seem to support listing distributions by publication.

I think we're going to have to either request the feature and wait or work around deleting these floating distributions.

Did you find that there are floating distributions often? I feel like this shouldn't be the case, since we delete distributions when repositories get deleted. I wonder if removing content view versions from an environment could cause that since the LCE distribution would no longer point to any repository.

ianballou (commented):
I found one bug where un-orphaned distributions were deleted, details below:

app/services/katello/pulp3/smart_proxy_repository.rb
sjha4: Ack..Removing this piece for now. Strangely the VCR re-recording ran without errors. There were some test repos left in pulp after the test run however.  Probably some tests that still don't have proper teardowns.

app/services/katello/pulp3/smart_proxy_repository.rb
```diff
@@ -63,6 +61,10 @@ def orphan_repository_versions
         repo_version_map
       end
 
+      def publication_for_repository_href(repository_version_href)
```
ianballou, line 64: Is this used anywhere?

ianballou (commented):
Looking good! Just one question:

app/services/katello/pulp3/smart_proxy_repository.rb
sjha4: This was a helper for the distribution lookup I had earlier. Let me remove this.

ianballou (approved):
Acking again! Still needs to wait for the other packaging ones to be good to go to merge.

PR #10951, opened 2024-03-28
Test node 16 workflow

sjha4:


sjha4:
Well..

MariaAga:
I recommend adding `.npmrc` file with `legacy-peer-deps=true` to make this smoother

PR #10952, opened 2024-04-01
Fixes #37302 - push containers through Katello

ianballou:
Adds support for pushing content through Katello to Pulp. Does not index any information from Pulp. As such, only pushing is possible through Katello.

The REMOTE_USER header must be set to perform remote auth with the Pulpcore registry. Apache will set this rather than Katello, so it's left out here.

0) Cherry-pick @sjha4 's changes here before pulling in this PR and upgrade to Pulpcore nightly: https://github.com/theforeman/foreman-packaging/pull/10658
  - Pulpcore Nightly repo: https://stagingyum.theforeman.org/pulpcore/nightly/
1) Due to https://github.com/pulp/pulp_container/issues/1577, hacking is required for auth with Pulp to work. In `/usr/lib/python3.11/site-packages/pulp_container/app/token_verification.py`, add `return RemoteUserRegistryAuthentication().authenticate(request)` to the top of `RegistryAuthentication.authenticate.`
  - Don't forget to restart Pulpcore after making the changes.
2) Apache must also be configured to send the REMOTE_USER header over to Pulp. Edit `/etc/httpd/conf.d/05-foreman-ssl.conf` so that the `/pulpcore_registry/v2/` location loses `RequestHeader set REMOTE_USER "%{SSL_CLIENT_S_DN_CN}s" env=SSL_CLIENT_S_DN_CN` and gains `RequestHeader set REMOTE_USER "admin"`.
  - https://github.com/theforeman/puppet-pulpcore/pull/337
3) Add the following to `foreman/config/settings.plugins.d/katello.yaml`:
```yaml
# The katello section should exist already.
:katello:
  :container_image_registry:
    :allow_push: true
```
4) Pull some image to your system using `podman pull` so there is something to push.
5) Push to Katello:
```
# Example -- 76435261c6d3 must point to an image on your machine. Find it with 'podman image ls' after pulling.
podman push 76435261c6d3 centos8-katello-devel.manicotto.example.com/organization/product/quarkus-micro-image
```
6) Check that a container push repo was created in Pulp:
```
pulp container repository -t push list
```
  - Make sure that actual content was uploaded.
7) Try a few images. Try some big ones!

### DIFF
--- a/app/controllers/katello/api/registry/registry_proxies_controller.rb
+++ b/app/controllers/katello/api/registry/registry_proxies_controller.rb
@@ -4,14 +4,16 @@ module Katello
     before_action :disable_strong_params
     before_action :confirm_settings
     before_action :confirm_push_settings, only: [:start_upload_blob, :upload_blob, :finish_upload_blob,
-                                                 :chunk_upload_blob, :push_manifest]
+                                                 :push_manifest]
     skip_before_action :authorize
     before_action :optional_authorize, only: [:token, :catalog]
     before_action :registry_authorize, except: [:token, :v1_search, :catalog]
     before_action :authorize_repository_read, only: [:pull_manifest, :tags_list]
-    before_action :authorize_repository_write, only: [:push_manifest]
+    # TODO: authorize_repository_write commented out until Katello indexes Pulp container push repositories.
+    # before_action :authorize_repository_write, only: [:start_upload_blob, :upload_blob, :finish_upload_blob,
+    #                                                   :push_manifest]
     skip_before_action :check_media_type, only: [:start_upload_blob, :upload_blob, :finish_upload_blob,
-                                                 :chunk_upload_blob, :push_manifest]
+                                                 :push_manifest]
 
     wrap_parameters false
 
@@ -182,15 +184,8 @@ module Katello
     end
 
     def check_blob
-      begin
-        r = Resources::Registry::Proxy.get(@_request.fullpath, 'Accept' => request.headers['Accept'])
-        response.header['Content-Length'] = "#{r.body.size}"
-      rescue RestClient::NotFound
-        digest_file = tmp_file("#{params[:digest][7..-1]}.tar")
-        raise unless File.exist? digest_file
-        response.header['Content-Length'] = "#{File.size digest_file}"
-      end
-      render json: {}
+      pulp_response = Resources::Registry::Proxy.get(@_request.fullpath, 'Accept' => request.headers['Accept'])
+      head pulp_response.code
     end
 
     def redirect_client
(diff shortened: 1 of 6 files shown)


qcjames53:
Hi Ian, thanks for the earlier fix. It looks like small/medium images are pushed properly but larger images run into issues. I have been seeing `502 bad gateway` errors on pushes for images like [this one](https://quay.io/repository/keycloak/keycloak) (407 MB). Rails console output shows the following:
```
HTTP parse error, malformed request ("PATCH /v2/organization/product/keycloak/blobs/uploads/018eaadd-7f75-7ae8-958c-5db6a1ea456c" - (192.168.121.94)): #<Puma::HttpParserError: maximum size of chunk header exceeded>
```

config/routes/api/registry.rb
```diff
@@ -11,15 +11,15 @@ class ActionDispatch::Routing::Mapper
       match '/v2/token' => 'registry_proxies#token', :via => :post
       match '/v2/*repository/manifests/:tag' => 'registry_proxies#pull_manifest', :via => :get
       # Push-related routes are disabled until there is support for pushing to Pulp 3.
```
qcjames53, line 13: I believe this comment should be removed.

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -182,15 +184,8 @@ def pull_manifest
     end
 
     def check_blob
```
qcjames53, line 186: I have a feeling the image size issue is something with this method. We _were_ setting a content length value in the response header that doesn't seem to exist anymore? Qualifier on this advice: I may not know what I'm talking about.

qcjames53 (changes requested):
A few comments...

app/controllers/katello/api/registry/registry_proxies_controller.rb
ianballou: Our code actually errors out before we even reach this method, the issue is between podman and puma rather than our code. The MAX_CHUNK_HEADER_SIZE is so small that puma rejects the request that podman sends to us. I'm going to report an issue to puma to see what they say.

ianballou: https://github.com/puma/puma/issues/3362

qcjames53: Good to know, thanks! Interesting issue. Are you wanting to keep this on the back burner until they get back to you or push this through under the pretense larger image sizes should be fine in the future?

qcjames53: Awesome, thanks!

ianballou:
Reminder for myself to see about adding / modifying tests.

ianballou:
[test katello]

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -258,42 +233,37 @@ def status_upload_blob
...
-    def chunk_upload_blob
-      response.header['Location'] = "#{request_url}/v2/#{params[:repository]}/blobs/uploads/#{params[:uuid]}"
-      render plain: '', status: :accepted
+    def translated_headers_for_proxy
+      current_headers = {}
+      env = request.env.select do |key, _value|
+        key.match("^HTTP_.*")
+      end
+      env.each do |header|
+        current_headers[header[0].split('_')[1..-1].join('-')] = header[1]
```
sjha4, line 242: @ianballou : Could you explain why this is needed?

ianballou: This is one bit of the code that was created a while back. I believe it's here because request.env contains environment variables, and some of which translate to headers. The headers as env vars have underscores and are prefixed with "HTTP_".  To turn them into true headers, they need the underscores turned into hyphens also the HTTP prefix needs removing. 

I need to read up more on it to verify, but one example I saw was that our apache must set the Remote-User header, which eventually reaches Pulp/Django as "HTTP_REMOTE_USER".

ianballou: I was trying to remember why `start_upload_blob` doesn't use `translated_headers_for_proxy`, and remembered that uploading a blob doesn't require special headers, unlike actually uploading the blob. For example: https://specs.opencontainers.org/distribution-spec/?v=v1.0.0#DISTRIBUTION-SPEC-93

ianballou: Some more helpful material about the web server environment variables https://medium.com/@cipeinado/unlock-http-info-in-rails-with-rack-request-env-and-uri-c7835397dc14
It clarifies my thinking about how the HTTP_ variables relate to headers.

ianballou: Another thing that's interesting -- so `request.headers` doesn't list anything by default. You need to look up a specific header to get info. If you type `request.headers.to_a`, for example, you end up just getting the normal `request.env` information more or less. It would need to be verified, but I think `request.headers['key']` looks for headers by various naming conventions. For example:

```ruby
[13] pry(#<Katello::Api::Registry::RegistryProxiesController>)> request.headers
=> #<ActionDispatch::Http::Headers:0x000055afdecc0450
 @req=
  #<ActionDispatch::Request POST "https://centos8-katello-devel.manicotto.example.com/v2/organization/product/hello-world/blobs/uploads/" for 192.168.122.147>>
[14] pry(#<Katello::Api::Registry::RegistryProxiesController>)> request.headers['Authorization']
=> "Bearer $2a$09$1b6453892473a467d0737uL4crR75S3rqz44qQcdMGK68gYLHX/Ba"
[15] pry(#<Katello::Api::Registry::RegistryProxiesController>)> request.headers['AUTHORIZATION']
=> "Bearer $2a$09$1b6453892473a467d0737uL4crR75S3rqz44qQcdMGK68gYLHX/Ba"
[16] pry(#<Katello::Api::Registry::RegistryProxiesController>)> request.headers['HTTP_AUTHORIZATION']
=> "Bearer $2a$09$1b6453892473a467d0737uL4crR75S3rqz44qQcdMGK68gYLHX/Ba"
[17] pry(#<Katello::Api::Registry::RegistryProxiesController>)> request.headers['HTT_AUTHORIZATION']
=> nil
[18] pry(#<Katello::Api::Registry::RegistryProxiesController>)> request.headers['HTTP_AUTHORIZATION']
=> "Bearer $2a$09$1b6453892473a467d0737uL4crR75S3rqz44qQcdMGK68gYLHX/Ba"
[19] pry(#<Katello::Api::Registry::RegistryProxiesController>)> request.headers['x-authorization']
=> nil
[20] pry(#<Katello::Api::Registry::RegistryProxiesController>)> request.headers['X-Authorization']
=> nil
```

ianballou: I did realize that some headers, like the Content-Length, doesn't make it in with the HTTP_ env vars. So I'll add explicit checks for those.

ianballou:
@qcjames53 my PR is ready for another round of reviews.

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -210,94 +205,67 @@ def pull_blob
...
+      end
+      current_headers
     end
 
     def upload_blob
-      File.open(tmp_file("#{params[:uuid]}.tar"), 'ab', 0600) do |file|
-        file.write request.body.read
+      headers = translated_headers_for_proxy
+      headers['Content-Type'] = request.headers['Content-Type'] if request.headers['Content-Type']
```
qcjames53, line 244: It really feels like there should be a more concise way to do all these calls but neither ||= or &&= works. Perhaps you could split all these test blocks into a method?

qcjames53 (changes requested):
Hi Ian. I was able to test this PR again with a large image (I used keycloak again, 461MB). Using [the puma patch you mentioned](https://github.com/puma/puma/pull/3338/files), I'm unfortunately still seeing 502 bad gateway errors. Still the same puma error as well: `#<Puma::HttpParserError: maximum size of chunk header exceeded>`. I'm hoping I'm doing something incorrect with this testing; please let me know if this _**should**_ be working on my end.

Thanks for adding test cases for the new functionality; good addition. I was unable to get them to run locally but it looks like the GH automation was happy. As of now I don't think we need a test for large images since you are mocking the pulp responses anyways. 

Thanks for the other small formatting/comment changes. Outside of the one comment below, everything else looks pretty good to me.

app/controllers/katello/api/registry/registry_proxies_controller.rb
ianballou: Lemme take a look about if there's a less verbose way to do this that doesn't lose readability.

qcjames53:
Can confirm that the new puma patch works to allow pushes for larger images. Thank you for the help with that config, Ian.

app/controllers/katello/api/registry/registry_proxies_controller.rb
qcjames53: Based on our video call I'm happy to call this ok. It's very readable and quick as you wrote it.

qcjames53 (approved):
ACK

app/controllers/katello/api/registry/registry_proxies_controller.rb
ianballou: I didn't find a less verbose way to write this, that check for nil seems like a requirement. C'mon Ruby!

PR #10953, opened 2024-04-02
Refs #36924 - Add breadcrumb bar to Add Subscription page, fix add subscription button url

qcjames53:
The add subscription page now has a breadcrumb bar showing the user's path through the subscriptions overview page.

This is the second pull request for [this redmine issue](https://projects.theforeman.org/issues/36924). The first PR failed QA; this PR should correct missing elements from the first PR. Both PRs are necessary to complete the issue.

Open Katello to the subscriptions overview page. Click 'Add Subscription'. The add subscription page should display a breadcrumb bar in the header which allows the user to return to the subscriptions overview page.

### DIFF
--- a/webpack/scenes/Subscriptions/UpstreamSubscriptions/UpstreamSubscriptionsPage.js
+++ b/webpack/scenes/Subscriptions/UpstreamSubscriptions/UpstreamSubscriptionsPage.js
@@ -222,19 +222,20 @@ class UpstreamSubscriptionsPage extends Component {
 
     return (
       <Grid bsClass="container-fluid">
-        <BreadcrumbsBar data={{
-          isSwitchable: false,
-          breadcrumbItems: [
-            {
-              caption: __('Subscriptions'),
-              onClick: () => this.props.history.push('/subscriptions'),
-            },
-            {
-              caption: __('Add Subscriptions'),
-            },
-          ],
-        }}
-        />
+        {!upstreamSubscriptions.loading && <div style={{marginBottom: "10px"}}>
+          <BreadcrumbsBar
+            isLoadingResources={upstreamSubscriptions.loading}
+            breadcrumbItems={[
+              {
+                caption: __('Subscriptions'),
+                url: '/subscriptions/',
+              },
+              {
+                caption: String(__('Add Subscriptions')),
+              },
+            ]}
+          />
+        </div>}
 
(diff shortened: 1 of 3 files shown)


webpack/scenes/Subscriptions/UpstreamSubscriptions/UpstreamSubscriptionsPage.js
```diff
@@ -11,6 +11,7 @@ import { urlBuilder } from 'foremanReact/common/urlHelpers';
 import { LoadingState } from '../../../components/LoadingState';
 import { Table } from '../../../components/pf3Table';
 import { columns } from './UpstreamSubscriptionsTableSchema';
+import api, { orgId } from '../../../services/api';
```
jeremylenz, line 14: These imports don't seem to be used anywhere in this file. I know we saw them yesterday, so not sure what happened?

qcjames53: Good catch. I ended up removing the subscription switcher since I didn't believe it made sense in this context. That orphaned these imports.

jeremylenz: gotcha, makes sense! :+1:

jeremylenz (changes requested):
Code change here looks good, but I found another issue which it would be great to fix here.

When I click the Add Subscriptions button, it links to `/subscriptions/subscriptions/add` but it should be linking to simply `/subscriptions/add`. I bet this is why we were seeing the unexplained blank screens.

The fix is in https://github.com/Katello/katello/blob/d8b76387c3bf2a4d139d16f916e5c71d1548e9bf/webpack/scenes/Subscriptions/components/SubscriptionsToolbar/SubscriptionsToolbar.js#L52 where you just need a slash added before `subscriptions/add`.

That code has been there over 5 years, so I'm not sure what changed.

Side note re: the commit message -  this situation is a near-perfect occasion to use "Refs" rather than "Fixes."

qcjames53:
@jeremylenz Added your suggestion, thanks! The PR and the commit should be reworded appropriately.

jeremylenz (approved):
Works as expected. thx for fixing the url too :slightly_smiling_face: 

Test failures unrelated

Thanks @qcjames53! ACK :+1:

PR #10958, opened 2024-04-11
Fixes #37340 - do not use shell to update packages

maximiliankolb:
In favor of just running a shell command, you can use the Ansible module "ansible.builtin.package" to perform package actions such as installing, updating, and removing packages.

* docs: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/package_module.html
* see also "install_package_-_katello_ansible_default.erb"

Make better use of Ansible to run package manager-indenpendent package actions.

copied code shamelessly from https://github.com/Katello/katello/blob/master/app/views/foreman/job_templates/install_package_-_katello_ansible_default.erb#L16

I noticed that there is also the option to further nest the templates, i.e. call another template from within this template. However, I consider this a worse solution because the `state` in `install_package_-_katello_ansible_default.erb` should be "present", not "latest". (IMHO a bug). Therefore, we should differentiate between "installing" as in "ensuring it's present" and actually updating a package.

Apply the patch manually to the job templates on your Foreman+Katello nightly and then execute a job on multiple hosts, for example on RHEL 8 and Debian 12.

### DIFF
--- a/app/views/foreman/job_templates/update_package_-_katello_ansible_default.erb
+++ b/app/views/foreman/job_templates/update_package_-_katello_ansible_default.erb
@@ -13,4 +13,8 @@ provider_type: Ansible
 kind: job_template
 %>
 
-<%= render_template('Run Command - Ansible Default', :command => "yum -y update #{input('package')}") %>
+<%
+pkgs = input('package')
+pkgs = '"*"' if pkgs.empty?
+-%>
+<%= render_template('Package Action - Ansible Default', :state => 'latest', :name => pkgs) %>
--- a/app/views/foreman/job_templates/update_packages_by_search_query_-_katello_ansible_default.erb
+++ b/app/views/foreman/job_templates/update_packages_by_search_query_-_katello_ansible_default.erb
@@ -3,7 +3,7 @@ kind: job_template
 name: Update packages by search query - Katello Ansible Default
 job_category: Katello
 description_format: 'Update package(s) %{Packages search query}'
-feature: katello_package_remove_by_search
+feature: katello_package_update_by_search
 provider_type: Ansible
 template_inputs:
 - name: Packages search query
@@ -22,7 +22,7 @@ template_inputs:
   versions: input('Selected update versions')
 ) -%>
 <% if package_names.empty? -%>
-<%= render_template('Run Command - Ansible Default', :command => "yum -y update") %>
+<%= render_template('Package Action - Ansible Default', :state => 'latest', :name => '"*"') %>
 <% else -%>
 ---
 - hosts: all


maximiliankolb:
draft PR until I test this locally and fix the typos in the commit message.

maximiliankolb:
@m-bucher I noticed that "Update packages by search query - Katello Ansible Default" runs different code on Debian/Ubuntu compared to EL. I suspect this is due to https://github.com/Katello/katello/pull/10958/files#diff-7a4d0bd5b733f160cfa65a705e51f10f45e9deb5deaf4fb8cdd26feeeedce575R24. Because it's using the "package" Ansible module, do you think we can a) drop the if-statement or b) expand the if-statement to also check for Deb packages?

m-bucher:
[test katello]

m-bucher (approved):
The `if package_names.empty?` statement IMHO has to stay, because that is the way updating all packages is done, if no package_name is specified.

And your change IMHO removes the necessity to do a per packaging-system distinction of case.

Looks good to me over all.

maximiliankolb:
Rebased to "master" and slightly reworked.

app/views/foreman/job_templates/update_packages_by_search_query_-_katello_ansible_default.erb
```diff
@@ -16,14 +16,14 @@ template_inputs:
   required: false
   value_type: plain
 %>
+<% if input('Packages search query').empty? && input('Selected update versions').empty? -%>
+<%= render_template('Package Action - Ansible Default', :state => 'latest', :name => '"*"') %>
+<% else -%>
```
maximiliankolb, line 21: I wonder if we need a third case for "only the version field is empty" aka. a user only inputs a package name. Doing this results in a different template for EL vs DEB:

image

Both jobs succeed; tested on Foreman 3.9 with input "subscription-manager".

What do you think? @nadjaheitmann

app/views/foreman/job_templates/update_packages_by_search_query_-_katello_ansible_default.erb
```diff
@@ -16,14 +16,14 @@ template_inputs:
...
 <% package_names = @host.package_names_for_job_template(
   action: 'update',
   search: input('Packages search query'),
   versions: input('Selected update versions')
 ) -%>
-<% if package_names.empty? -%>
-<%= render_template('Run Command - Ansible Default', :command => "yum -y update") %>
```
nadjaheitmann, line 25: I think it should be enough to replace line 25 with the code that you moved to line 20. 

There is the method here https://github.com/Katello/katello/blob/master/app/models/katello/concerns/host_managed_extensions.rb#L498 that takes care of the package search and it should return the correct packages based on the search query. For empty search query it simply returns an empty array. For yum packages it considers nvra, for deb packages we do not check the version so anything that you put into the version field for deb packages will be ignored.

app/views/foreman/job_templates/update_packages_by_search_query_-_katello_ansible_default.erb
```diff
@@ -3,7 +3,7 @@ kind: job_template
 name: Update packages by search query - Katello Ansible Default
 job_category: Katello
```
nadjaheitmann, line 4: What is a bit funny it that this template has the category `Katello` and the template `Update Package - Katello Ansible Default` has the category `Katello via Ansible`.

This results in a separate section for Katello jobs with the caption Ansible:
image

IMHO all of them should go to `Katello via Ansible`, otherwise this is somehow inconsistent (but this is out of scope of this PR).

app/views/foreman/job_templates/update_packages_by_search_query_-_katello_ansible_default.erb
maximiliankolb: I have applied your feedback; this resulted in a much smaller diff.

app/views/foreman/job_templates/update_packages_by_search_query_-_katello_ansible_default.erb
maximiliankolb: I have created a separate issue: https://projects.theforeman.org/issues/37362

nadjaheitmann:
Now that I think about it, maybe you could change the commit message to something like "Use Ansible package action for updating packages via Ansible"

app/views/foreman/job_templates/update_package_-_katello_ansible_default.erb
```diff
@@ -13,4 +13,10 @@ provider_type: Ansible
 kind: job_template
 %>
 
-<%= render_template('Run Command - Ansible Default', :command => "yum -y update #{input('package')}") %>
+<%
+pkgs = input('package')
+if pkgs.empty?
+  pkgs = '"*"'
+end
```
sbernhard, line 20:
```suggestion
pkgs = '"*"' if pkgs.empty?
```

maximiliankolb: Applied.

maximiliankolb:
Applied all feedback and rebased to "master".

sbernhard (approved):

sbernhard:
[test katello]

app/views/foreman/job_templates/update_packages_by_search_query_-_katello_ansible_default.erb
jeremylenz: Rendering the template with the same inputs should result in updates to the same packages on EL and Debian.

Wouldn't the `name:       []` cause an update to all packages, not just `python3-subscription-manager` ?

jeremylenz: No, the packages search query is a scoped search and should allow all the same searches as the Packages page at `/packages`.

nadjaheitmann: Then I think that the template needs some input validation check. The example above uses `subscription-manager` as input, which is not a valid search query as you said. I think this is out of scope for this PR, though.

jeremylenz: `subscription-manager` is indeed a valid scoped search input! The code looks fine to me; I was just questioning the screenshots since one has `name:       []` and the other one shows the package names.

nadjaheitmann: What I meant is that we have a package name for the `Update Package - Katello Ansible Default` template: 
https://github.com/Katello/katello/blob/master/app/views/foreman/job_templates/update_package_-_katello_ansible_default.erb#L8

And we have `Packages search query` as input for the `Update packages by search query - Katello Ansible Default` template: https://github.com/Katello/katello/blob/master/app/views/foreman/job_templates/update_packages_by_search_query_-_katello_ansible_default.erb#L9

What I am wondering is that if you just enter a package name for the `Packages search query` as input, is this really valid? What I see from the auto-generated inputs is something like `id ^ (1, 2, 3)` when I trigger package update in the new Host details page.

jeremylenz: Yes, this is a valid search. Searches without key/value pairs are totally legit and they're even the first example in [the scoped_search usage doc](https://github.com/wvanbergen/scoped_search?tab=readme-ov-file#usage-). In the UI, this search would return one or more packages, and then the user would select them with the checkboxes. When clicked, the checkboxes [add the ID](https://github.com/Katello/katello/blob/95a55e0da6b5355170a23bed51d6f5065c05cb7f/webpack/components/extensions/HostDetails/Tabs/PackagesTab/PackagesTab.js#L573) to the selection set; that's why you end up with the `id ^ (1, 2, 3)` in the end.

jeremylenz: What I'm still not sure about is why this produces different results on Debian. As I said, the existing code here looks like it should work fine.. :thinking:

jeremylenz: I think I am discovering what's going on. `yum_names_for_job_template` checks `Rpm.installable_for_host` [here](https://github.com/Katello/katello/blob/a1c864acdde85deaff5b22e131086cff5e3ffb4a/app/models/katello/concerns/host_managed_extensions.rb#L526) (which actually returns a list of installed, upgradable packages) and builds a hash of upgradable packages with versions/arches. And `deb_names_for_job_template` just bases it on the name alone, without the version complexities. I suppose for some reason `subscription-manager` wasn't in Katello's list of installed debs for that host, and that's why it returned an empty list.

If you want to add the ability later to select which deb version to upgrade to, a similar change would be needed for `deb_names_for_job_template`. But I'm okay leaving this as-is in this PR.

jeremylenz (approved):
Did a bit of investigation and [commented](https://github.com/Katello/katello/pull/10958#discussion_r1588330419), after which I'm fine with this change. thanks @maximiliankolb! ACK :+1:

jeremylenz:
Can you please rebase? hopefully that will resolve the 'React Tests' failure

maximiliankolb:
Thanks Jeremy. Rebased to "master". :crossed_fingers:

app/views/foreman/job_templates/update_packages_by_search_query_-_katello_ansible_default.erb
nadjaheitmann: > If you want to add the ability later to select which deb version to upgrade to, a similar change would be needed for deb_names_for_job_template

Eventually :) 

@jeremylenz Thanks for the detailed explanation and the investigation!

PR #10960, opened 2024-04-11
Fixes #37346 - Fetching Host's details does not scale wrt Hosts Collections

ianballou:
Changes `length` to `count` in HostCollection#total_hosts since `length` will load all records in hosts where `count` does a `count` query.

This should significantly improve API performance when loading hosts.

I considered if we should remove total_hosts completely from the host collections view snippet, but the count query change should make it much faster.

With my patch, when listing a host via the UI you should see postgres queries like:

```
2024-04-11 15:40:06 UTC LOG:  execute a28: SELECT "katello_host_collections".* FROM "katello_host_collections" INNER JOIN "katello_host_collection_hosts" ON "katello_host_collections"."id" = "katello_host_collection_hosts"."host_collection_id" WHERE "katello_host_collection_hosts"."host_id" = $1
2024-04-11 15:40:06 UTC DETAIL:  parameters: $1 = '1'
2024-04-11 15:40:06 UTC LOG:  execute <unnamed>: SELECT COUNT(*) FROM "hosts" INNER JOIN "katello_host_collection_hosts" ON "hosts"."id" = "katello_host_collection_hosts"."host_id" WHERE "hosts"."type" = $1 AND "katello_host_collection_hosts"."host_collection_id" = $2
2024-04-11 15:40:06 UTC DETAIL:  parameters: $1 = 'Host::Managed', $2 = '1'
2024-04-11 15:40:06 UTC LOG:  execute a29: SELECT "katello_content_view_versions".* FROM "katello_content_view_versions" WHERE "katello_content_view_versions"."id" = $1 LIMIT $2
```
Note the `SELECT COUNT(*) FROM "hosts"`.

Without my patch, you'll see logs like:
```
2024-04-11 15:39:38 UTC LOG:  execute a27: SELECT "katello_host_collections".* FROM "katello_host_collections" INNER JOIN "katello_host_collection_hosts" ON "katello_host_collections"."id" = "katello_host_collection_hosts"."host_collection_id" WHERE "katello_host_collection_hosts"."host_id" = $1
2024-04-11 15:39:38 UTC DETAIL:  parameters: $1 = '1'
2024-04-11 15:39:38 UTC LOG:  execute a28: SELECT "hosts".* FROM "hosts" INNER JOIN "katello_host_collection_hosts" ON "hosts"."id" = "katello_host_collection_hosts"."host_id" WHERE "hosts"."type" = $1 AND "katello_host_collection_hosts"."host_collection_id" = $2
2024-04-11 15:39:38 UTC DETAIL:  parameters: $1 = 'Host::Managed', $2 = '1'
```

The `SELECT "hosts".*` is what should be causing the performance issue.

Try doing a timing test or a SQL complexity comparison to help prove that this is indeed an improvement.

### DIFF
--- a/app/models/katello/host_collection.rb
+++ b/app/models/katello/host_collection.rb
@@ -29,7 +29,10 @@ module Katello
     scoped_search :relation => :hosts, :on => :name, :rename => :host, :complete_value => true
 
     def max_hosts_check
-      if !unlimited_hosts && (hosts.length > 0 && (hosts.length.to_i > max_hosts.to_i)) && max_hosts_changed?
+      # NOTE: max_hosts_check and max_hosts_no_exceeded use size() instead of count() because
+      # the host list exists as an array rather than a DB query when run as a validation.
+      host_count = hosts.size
+      if !unlimited_hosts && (host_count > 0 && (host_count.to_i > max_hosts.to_i)) && max_hosts_changed?
         errors.add :max_host, N_("may not be less than the number of hosts associated with the host collection.")
       end
     end
@@ -60,8 +63,14 @@ module Katello
       type ? query.of_type(type) : query
     end
 
-    def total_hosts
-      hosts.count
+    def cache_key
+      "#{self.class.name}/#{self.id}"
+    end
+
+    def total_hosts(cached: false)
+      Rails.cache.fetch("#{cache_key}/total_hosts", expires_in: 1.minute, force: !cached) do
+        hosts.count
+      end
     end
 
     # Retrieve the list of accessible host collections in the organization specified, returning
--- a/app/models/katello/host_collection.rb
+++ b/app/models/katello/host_collection.rb
@@ -61,7 +61,7 @@ module Katello
     end
 
     def total_hosts
-      hosts.length
+      hosts.count
     end
 
     # Retrieve the list of accessible host collections in the organization specified, returning
--- a/app/views/katello/api/v2/hosts/host_collections.json.rabl
+++ b/app/views/katello/api/v2/hosts/host_collections.json.rabl
@@ -1,3 +1,7 @@
 child :host_collections => :host_collections do
-  attributes :id, :name, :description, :max_hosts, :unlimited_hosts, :total_hosts
+  attributes :id, :name, :description, :max_hosts, :unlimited_hosts
+
+  node :total_hosts do |host_collection|
+    host_collection.total_hosts(cached: true)
+  end
 end


m-bucher (approved):
Nice, performance improvements are always welcome :+1: 

Would it make sense to review the use of `hosts.size` and `hosts.length` in `max_hosts_not_exceeded()` and `max_hosts_check()` as well?

It is rather tricky to choose between the two depending on the application and in this case it also depends on the context (and what else is loaded from the DB), e.g show a host, show a host_collection list host_collections.

Since showing one host is probably the one used the most, this should be the best choice.

adamruzicka:
@ares iirc you mentioned this didn't make all that much of a difference on your test box, correct?

ianballou:
> Would it make sense to review the use of `hosts.size` and `hosts.length` in `max_hosts_not_exceeded()` and `max_hosts_check()` as well?

@m-bucher I agree that I'd be a good idea to review our use of `record.size` in the code. Let's see first if we can get some performance testing done to verify whether or not the change to `length` does what I'm hoping.

jeremylenz:
@ianballou In light of https://bugzilla.redhat.com/show_bug.cgi?id=2249087#c11, what do you think we should do here? IIUC this change helps significantly, but removing the rabl node entirely is even faster.

jeremylenz (approved):
Acking, but also see comment above

ianballou:
The new time is about 50x faster than the old time on average, that's pretty significant.
With removing the total_hosts entirely, it seems there's another 3x speed increase compared to the PR I have here.

I believe the timing that was discovered with this PR alone (the ~50x speed increase) is up-to-spec. While we may believe that the total host count on the response is unhelpful, removing it as a bug fix would break semantic versioning.

One idea could be to use this patch as the bug fix backport, and deprecate the host -> host collection -> total hosts info entirely as a y-stream fix.

I'll wait to merge until we hear back from @adamruzicka and/or @ares since they have a stake in the solution now too.

---

To keep things simpler, here's the paste from the Bugzilla (https://bugzilla.redhat.com/show_bug.cgi?id=2249087#c9):

```
Hello, the PR https://github.com/Katello/katello/pull/10960 does significantly help.

Having 5095 Hosts and 100 Host Collections like in #c0, then:

- original execution times prior adding Hosts to Host Collections:

# for i in $(seq 1 10); do time curl -ks -u admin:redhat 'https://localhost/api/v2/hosts/4790' > /dev/null; done 2>&1 | grep real
real	0m0.440s
real	0m0.342s
real	0m0.342s
real	0m0.280s
real	0m0.251s
real	0m0.287s
real	0m0.249s
real	0m0.257s
real	0m0.244s
real	0m0.249s
#

- Bad times after adding Hosts to Host Collections:

# for i in $(seq 1 10); do time curl -ks -u admin:redhat 'https://localhost/api/v2/hosts/4790' > /dev/null; done 2>&1 | grep real
real	0m27.062s
real	0m38.244s
real	0m26.237s
real	0m43.044s
real	1m8.279s
real	0m16.226s
real	0m52.270s
real	0m22.764s
real	0m22.230s
real	0m24.327s
#

- Applying the patch and restarting services, reasonable small response times:

# for i in $(seq 1 10); do time curl -ks -u admin:redhat 'https://localhost/api/v2/hosts/4790' > /dev/null; done 2>&1 | grep real
real	0m0.887s
real	0m0.759s
real	0m0.700s
real	0m0.783s
real	0m0.951s
real	0m0.812s
real	0m0.768s
real	0m0.793s
real	0m0.806s
real	0m0.814s
#

```

And here's the other comment about removing the total hosts entirely (https://bugzilla.redhat.com/show_bug.cgi?id=2249087#c11):

```
.. testing also another patch proposal:

sed -i "s/, :total_hosts//1" /usr/share/gems/gems/katello-*/app/views/katello/api/v2/hosts/host_collections.json.rabl

gives even better times:

# for i in $(seq 1 10); do time curl -ks -u admin:redhat 'https://localhost/api/v2/hosts/4790' > /dev/null; done 2>&1 | grep real
real	0m0.273s
real	0m0.272s
real	0m0.260s
real	0m0.276s
real	0m0.249s
real	0m0.265s
real	0m0.261s
real	0m0.261s
real	0m0.254s
real	0m0.262s
#
```

ianballou:
@jeremylenz and I were discussing caching, so I've introduced a 1-minute long cache for holding the total_hosts count.  The cache is only used with the hosts endpoint. When looking at actual host collections, the cache will always force a miss. 

I also discovered in our code that total_hosts is run after host collections are updated via the API. So, since `force_update_cache` is `true` be default in my change, the cache will always be updated when a host collection has its host count updated via the API.

I also updated `max_hosts_not_exceeded()` and `max_hosts_check()` to use `count` for consistency.

ianballou:
Considering that loading a host with cached `total_hosts` should take ~ a quarter-second to load based on the performance findings, the 1-minute cache expiration time should cover a good amount of host calls without taking on much risk of having stale data.

app/models/katello/host_collection.rb
```diff
@@ -60,8 +61,14 @@ def errata(type = nil)
...
     end
 
-    def total_hosts
-      hosts.length
+    def cache_key
+      "#{self.class.name}/#{self.id}"
+    end
+
+    def total_hosts(force_update_cache: true)
+      Rails.cache.fetch("#{cache_key}/total_hosts", expires_in: 1.minute, force: force_update_cache) do
```
jeremylenz, line 69: In a few other places in Katello we have the argument as `cached` with a default of `false`, and then we pass `!cached` to `Rails.cache.fetch`. Prob better to stay consistent with that.

https://github.com/Katello/katello/blob/f3891b2b8dacb2f5f10e19b5df22df330019aff2/app/models/katello/concerns/organization_extensions.rb#L103-L104

ianballou: I saw two standards, so I was following this one: https://github.com/Katello/katello/blob/master/app/models/katello/host/content_facet.rb#L315

But `cached` is less verbose, so I like that.

ianballou:
I'm seeing good timing with the new change:

```
[root@pmoravec-sat613-disco katello-4.7.0.32]# for i in $(seq 1 10); do time curl -ks -u admin:redhat 'https://localhost/api/v2/hosts/4790' > /dev/null; done 2>&1 | grep real
real	0m0.925s
real	0m0.272s
real	0m0.272s
real	0m0.294s
real	0m0.298s
real	0m0.265s
real	0m0.271s
real	0m0.261s
real	0m0.273s
real	0m0.269s
```

This is with 5095 Hosts and 100 Host Collections like mentioned in the BZ. The run was taken after a minute to ensure the cache was invalidated.

ianballou:
I ended up  keeping the usage of `size` in `max_hosts_not_exceeded` since it will have the same performance regardless. In the tests only, `hosts` were being returned as an array which returned 0 with count even though it had two hosts in it.

m-bucher:
> I ended up keeping the usage of `size` in `max_hosts_not_exceeded` since it will have the same performance regardless. In testing the method, `hosts` were being returned as an array which returned 0 with count even though it had two hosts in it. I believe this is because `max_hosts_not_exceeded` runs as a validation before the host collection is created, so the count call returns nothing since nothing is committed to the DB yet.

@ianballou would it make sense to add a short code-comment to document this? I do not want to clutter the code with comments, but it might be a useful information. IMHO, performance is something that will be re-evaluated periodically. Therefore, having a reasoning for deliberately choosing count/size/length might save time, in the future :smiley:

PR #10967, opened 2024-04-19
Fixes #37371 - Fix bats and handle distributed archived repo version deletion

sjha4:
Try and delete publication of repo version if version deletion fails.
We have an issue https://github.com/pulp/pulpcore/issues/5218 open with pulp to provide a list of conflicting publication/distribution preventing deletion of repo version.
We will update our version deletion logic based on that implementation.

In the meantime, we have a workflow https://github.com/Katello/katello/blob/cb8c4231bc1db3f1e5630dd8d6e2f13ce41e953b/app/lib/actions/pulp3/orchestration/repository/delete.rb#L14 which seems to be causing bats test failures and in general preventing archived distributed repo versions from being directly deleted causing nightly bats failure. This should fix that.
Not entirely sure where the dangling version+distro was being created but in our bats test this was happening when creating incremental update to CV and then deleting child repos while removing them from versions.
Run bats tests against this. It should pass and if you have nightly, you should see `clean-up modules-rpms and rpm-deps` bats pass and both those repo get deleted. You should also see the log messages being recorded when running the test.

### DIFF
--- a/app/services/katello/pulp3/repository.rb
+++ b/app/services/katello/pulp3/repository.rb
@@ -333,6 +333,17 @@ module Katello
 
       def delete_version
         ignore_404_exception { api.repository_versions_api.delete(repo.version_href) } unless version_zero?
+      rescue api.api_exception_class => e
+        if e.message.include?("are currently being used to distribute content")
+          Rails.logger.warn "Exception when calling repository_versions_api->delete: #{e}"
+          publication_href = repo.publication_href
+          Rails.logger.warn "Trying to delete publication #{publication_href} for repository #{repo.id}}"
+          Rails.logger.error "Could not delete version: #{repo.version_href} because conflicting publication could not be looked up" unless publication_href
+          if publication_href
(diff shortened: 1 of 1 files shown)


evgeni:
I did run a pipeline locally with the packit build of this (via #10970) and it passed the "clean-up" step quite fine.

What I do not like is the hooking on the error message text and then performing additional actions. Can't we detect the fact that there are existing publications *before* trying to delete the version?

sjha4:
This change is fairly limited in scope in that we are only trying to fix what's failing, which is repo version with a direct publication in pulp.

The reason we want to be extra conservative and do the deletion on publications only after the version deletion error in pulp is cause it gives us an added layer of security against deleting valid publications. Sometimes we reuse publications on several versions of the repo in katello and we want to be careful with preemptively deleting any publications unless pulp asks us to which we should have upon resolution of the pulp RFE.

evgeni (approved):

PR #10968, opened 2024-04-19
Use 3.10-stable for GHA based tests, not develop

evgeni:

use the correct Foreman version for tests


ekohl (approved):
We should include this in the branching procedures

evgeni:
> We should include this in the branching procedures

I did mention that in https://github.com/Katello/katello/pull/10780#pullrequestreview-1843086521 ;-)

PR #10971, opened 2024-04-19
Pin foreman_remote_execution to < 13

ekohl:

This pins to < 13 because version 13 is incompatible with Foreman 3.9. It allows https://github.com/theforeman/foreman/pull/10134 to pass again.

https://github.com/theforeman/foreman-packaging/tree/rpm/3.9/packages/plugins/rubygem-foreman_remote_execution contains version 12.0.5, so this matches packaging.

CI passes.

### DIFF
--- a/katello.gemspec
+++ b/katello.gemspec
@@ -34,7 +34,7 @@ Gem::Specification.new do |gem|
 
   gem.add_dependency "rabl"
   gem.add_dependency "foreman-tasks", ">= 5.0"
-  gem.add_dependency "foreman_remote_execution", ">= 7.1.0"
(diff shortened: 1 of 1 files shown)


jeremylenz (commented):
Looks like this will need snapshot updates, for some reason? maybe Patternfly changed something since 4.11..

`npm test -- -u`

ekohl:
> Looks like this will need snapshot updates, for some reason? maybe Patternfly changed something since 4.11..

Can we ignore it in this PR? We also need to fix packit, but I've split that off to https://github.com/Katello/katello/pull/10972.

ianballou (approved):
The snapshots will just need to happen when we do more CPs in the future.

PR #10975, opened 2024-04-23
Fixes #37379 - Index labels and other fields for manifests

sjha4:
Index new fields on container manifests and add them to the API
Sync a container repo with labels/anotations.
ex: https://quay.io/repository/centos-bootc/fedora-bootc 
tag: sha256-4b336aaec054e57549075517ae025a8c3b8b23703d928437dce6a1db068eae5a

Check the API , replace :ID :
```
https://centos8-katello-devel-stable.example.com/katello/api/v2/docker_manifests?organization_id=1&page=1&paged=true&per_page=20&repository_id=:ID&search=
```
Screenshot from 2024-04-23 08-09-20

### DIFF
--- a/db/migrate/20240423112842_add_fields_to_katello_docker_manifest.rb
+++ b/db/migrate/20240423112842_add_fields_to_katello_docker_manifest.rb
@@ -1,0 +1,8 @@
+class AddFieldsToKatelloDockerManifest < ActiveRecord::Migration[6.1]
+  def change
+    add_column :katello_docker_manifests, :annotations, :jsonb, default: {}
+    add_column :katello_docker_manifests, :labels, :jsonb, default: {}
+    add_column :katello_docker_manifests, :is_bootable, :boolean, default: false
+    add_column :katello_docker_manifests, :is_flatpak, :boolean, default: false
+  end
+end
--- a/app/services/katello/pulp3/docker_manifest.rb
+++ b/app/services/katello/pulp3/docker_manifest.rb
@@ -26,7 +26,11 @@ module Katello
         {
           schema_version: unit['schema_version'],
           digest: unit['digest'],
-          pulp_id: unit[unit_identifier]
+          pulp_id: unit[unit_identifier],
+          annotations: unit['annotations'],
+          labels: unit['labels'],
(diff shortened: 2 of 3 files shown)


db/migrate/20240423112842_add_fields_to_katello_docker_manifest.rb
```diff
@@ -0,0 +1,8 @@
+class AddFieldsToKatelloDockerManifest < ActiveRecord::Migration[6.1]
+  def change
+    add_column :katello_docker_manifests, :annotations, :jsonb, default: {}
```
ianballou, line 3: I think it's fine to leave out indexes for now, but we might want to add them later if searching with in the annotations and labels becomes important. For now I think they'll just be used as information.

ianballou (commented):
Working well, I'm able to see the fields get populated.
One issue I hit was with older manifests. I discovered that existing manifests don't get reindexed if you resync the repo or even if you run `index_content(full_index: true)` again on them. I had to destroy the repository first and then resync.

If we add `:mutable => true` to DockerManifest in lib/katello/repository_types/docker.rb, the issue seems to be resolved. However we should double check that that is safe.

ianballou:
I confirmed with Lubos that manifests should be immutable, so I  think for the resync case we can rely on the rake task that reindexes manifests. We'll just need to ensure in that rake task that manifests are treated as if they were mutable.

sjha4:
Testing the manifest getting 

> I confirmed with Lubos that manifests should be immutable, so I think for the resync case we can rely on the rake task that reindexes manifests. We'll just need to ensure in that rake task that manifests are treated as if they were mutable.

I was testing this and saw that pulp doesn't pull in the new info on resync. Is that the behavior you saw or was it on katello end only?

ianballou:
> I was testing this and saw that pulp doesn't pull in the new info on resync. Is that the behavior you saw or was it on katello end only?

The behavior I saw happens when you run the manifest migration in Pulp and then call reindex in Katello. So yeah, since Pulp doesn't pull the info in on resync, I think we're good to do the same.

ianballou (approved):
Based on the comments, ACK!

Re-running that one failing test.

PR #10981, opened 2024-04-30
Fixes #37400 - Add setting to control validation of host/lifecycle environment/content source coherence

jeremylenz:

Adds a Setting to allow you to skip running the validation `ensure_valid_content_source`. In certain edge-case situations, this validation can cause problems with host registrations.

Turning this setting off will erase the guardrails introduced in https://github.com/Katello/katello/pull/10524. Therefore, it defaults to on and the vast majority of users should never need to touch it.

Regardless of the value of the setting, the web UI will still look at coherence of the assigned content source/LCE, and you'll still get UI niceties like those pictured in https://github.com/Katello/katello/pull/10524.

set up an external smart proxy
Don't sync any environments to it (not even Library)
Register the host to the external smart proxy.

With the setting on, you'll get an error. With the setting off, registration will succeed with no errors.

### DIFF
--- a/lib/katello/plugin.rb
+++ b/lib/katello/plugin.rb
@@ -512,6 +512,12 @@ Foreman::Plugin.register :katello do
         description: N_("Default download policy for Smart Proxy syncs (either 'inherit', immediate', or 'on_demand')"),
         collection: proxy_download_policies
 
+      setting 'validate_host_lce_content_source_coherence',
+        type: :boolean,
+        default: true,
+        full_name: N_('Validate host/lifecycle environment/content source coherence'),
+        description: N_("Validate that a host's assigned lifecycle environment is synced by the smart proxy from which the host will get its content. Applies only to API requests; does not affect web UI checks")
+
       setting 'pulpcore_export_destination',
         type: :string,
         default: "/var/lib/pulp/exports",
--- a/app/models/katello/content_view_environment_content_facet.rb
+++ b/app/models/katello/content_view_environment_content_facet.rb
@@ -14,7 +14,9 @@ module Katello
       hostname = self.content_facet&.host&.name
       return unless [source, env].all? { |x| x.present? }
       unless source.lifecycle_environments.include?(env)
-        errors.add(:base, _("Host %{hostname}: Cannot add content view environment to content facet. The host's content source '%{content_source}' does not sync lifecycle environment '%{lce}'.") % { hostname: hostname, content_source: source.name, lce: env.name })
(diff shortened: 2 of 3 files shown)


lib/katello/plugin.rb
```diff
@@ -512,6 +512,12 @@ def katello_template_setting_values(name)
         description: N_("Default download policy for Smart Proxy syncs (either 'inherit', immediate', or 'on_demand')"),
         collection: proxy_download_policies
 
+      setting 'validate_host_lce_content_source_coherence',
+        type: :boolean,
+        default: true,
+        full_name: N_('Validate host/lifecycle environment/content source coherence'),
+        description: N_("Validate that a host's assigned lifecycle environment is synced by the smart proxy from which the host will get its content")
```
ianballou, line 519:
```suggestion
        description: N_("Validate that a host's assigned lifecycle environment is synced by the smart proxy from which the host will get its content at registration time")
```

The change content source UI will still block this.

ianballou (commented):
Works, with the setting on I can register to smart proxies that don't sync LCEs. Without the setting, at registration time, I see:
```
Validation failed: Content view environment content facets is invalid (HTTP error code 422: Unprocessable Entity)
```

lib/katello/plugin.rb
jeremylenz: There are still some places in the UI that won't, like the legacy Content Host UI and Content host bulk actions. So it's not just registration time.

jeremylenz:
> Works, with the setting on I can register to smart proxies that don't sync LCEs. Without the setting, at registration time, I see:
> 
> ```
> Validation failed: Content view environment content facets is invalid (HTTP error code 422: Unprocessable Entity)
> ```

Interesting, is that the error that's propagated to subscription-manager? The other day we were actually seeing the validation message itself ("does not sync lifecycle environment xxx") which would be better.

ianballou:
@jeremylenz yeah that's how it popped up when I tried registering to a smart proxy that didn't sync the LCE.

jeremylenz:
Ah yeah, I'm pretty sure it's because that validation moved from content_facet to content_view_environment_content_facet.

lib/katello/plugin.rb
ianballou: Think it's worth noting that the change content source UI alone won't respect the setting then?

ianballou (approved):
Works great! I see the right errors in bash after registration and in the logs. Am able to work around the errors with the setting.

PR #10982, opened 2024-05-02
Fixes #37402 - add jquery-ui dependency

MariaAga:
we want to remove it in core since its no longer used https://github.com/theforeman/foreman/pull/10142
so adding it here as only Katello seem to use it, and hopefully it will be removed here as well

### DIFF
--- a/katello.gemspec
+++ b/katello.gemspec
@@ -64,6 +64,7 @@ Gem::Specification.new do |gem|
   # UI
   gem.add_dependency "deface", '>= 1.0.2', '< 2.0.0'
   gem.add_dependency "angular-rails-templates", "~> 1.1"
+  gem.add_dependency "jquery-ui-rails", "~> 6.0"
 
   gem.add_development_dependency "theforeman-rubocop", '~> 0.1.0'
 end


ekohl (approved):
Looks like an OK stopgap to me. Did you have a look at how many work it would be to actually drop it from Katello too?

MariaAga:
yeah its used in some parts of the sync management (and maybe more, I didn't look after that). and I dont have capacity now to try to remove it from there since its spread through a few files

ekohl:
Sounds good to me: the Katello team has plans to remove those pages anyway. It may be a waste of time trying to rework them just to drop jquery-ui while they're fully replaced soon after.

jeremylenz:
Sync management is special, though, since it's actually ERB + vanilla JS + jQuery and doesn't have any Angular in it :flushed:

MariaAga:
@jeremylenz Can we move jquery-ui to katello anyway? and makes plans to maybe refactor sync management later?

jeremylenz:
@MariaAga If we're the only ones that use it, I'm fine with moving it to Katello for now. And yes we should also talk about redesigning sync status :)

sjha4:
Wondering if this: https://bugzilla.redhat.com/show_bug.cgi?id=2274132 has anything to do with this change here and if it fixes this.

jeremylenz:
@sjha4 The Foreman PR dropping it is not yet merged, so my guess is it's something else 🤷

MariaAga:
is anything else needed for this pr?

m-bucher:
> is anything else needed for this pr?

This will also require moving the dependency in https://github.com/theforeman/foreman-packaging from foreman to rubygem-katello package.
I assume it would be nice to have that prepared in a PR there as well :wink: .

Once that is done I guess it is only a matter of who is willing and allowed to push the buttons :smile:

MariaAga:
@m-bucher Thanks, added it: https://github.com/theforeman/foreman-packaging/pull/11004

ianballou:
A rebase will be needed to get the tests green.

ianballou:
@MariaAga looks like there's a conflict now with katello.gemspec.  Feel free to ping me when the time to merge is right so we can do it quickly.

MariaAga:
@ianballou Rebased, and https://github.com/theforeman/foreman-packaging/pull/11416 was merged.

ianballou:
@MariaAga do we need to keep waiting for the Foreman PR to merge?

ekohl (approved):
Packaging was done.

ekohl:
> @MariaAga do we need to keep waiting for the Foreman PR to merge?

No, quite the opposite. This needs to go in before Foreman. The Katello codebase really depends on it while Foreman itself doesn't.

ianballou:
Angular failures are unrelated. Thanks!

PR #10983, opened 2024-05-02
Fixes #37403 - make repository rpm ID sequence bigint

ianballou:
Makes katello_repository_rpms_id_seq a bigint since users are hitting the ID limit.

I was thinking if we need to make other fields larger, however no user has complained so I think it should be safe to only increase this one.

1) In `psql`, look at `select * from pg_sequences where sequencename='katello_repository_rpms_id_seq';`
2) See that the sequence has the type of `integer`
3) Run the migration
4) Run the query above again
5) See that the data type is now `bigint`
6) Rollback the query
7) Run `ALTER SEQUENCE katello_repository_rpms_id_seq AS bigint;` in `psql`
8) Run the migration again. Ensure there are no errors
9) Check that the ID sequence data type is still `bigint`.

### DIFF
--- a/db/migrate/20240502192021_change_katello_repository_rpms_id_seq_to_big_int.rb
+++ b/db/migrate/20240502192021_change_katello_repository_rpms_id_seq_to_big_int.rb
@@ -1,0 +1,9 @@
+class ChangeKatelloRepositoryRpmsIdSeqToBigInt < ActiveRecord::Migration[6.1]
+  def up
+    execute 'ALTER SEQUENCE katello_repository_rpms_id_seq AS bigint;'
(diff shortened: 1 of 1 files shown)


m-bucher (approved):
Makes sense, since the `id` column in `katello_repository_rpms` is already bigint.

I am thinking about suggesting to do the same for `katello_repository_debs`. I must admit though that I have not heard from anyone who has hit the MAX_INT, yet.
It has to be a separate ticket and PR anyway, so this one is IMHO good to go.

chris1984 (approved):
ACK, works as expected in the testing steps.

PR #10985, opened 2024-05-06
Fixes #37394 - content_view_filter_id only works for ID-type filters

asteflova:

https://bugzilla.redhat.com/show_bug.cgi?id=1785146, and by extension https://projects.theforeman.org/issues/37394, requests documenting that the --content-view-filter-id flag only works for ID-type filters.

CI checks are successful.


adamruzicka:
Before
```
# hammer -r erratum list -h  | grep content-view-filter
 --content-view-filter[-id] VALUE/NUMBER        Content View Filter name/id
```

After
```
# hammer -r erratum list -h  | grep content-view-filter
 --content-view-filter[-id] VALUE/NUMBER        Content View Filter name/id. Use to filter by name/id
```

chris1984 (approved):

PR #10986, opened 2024-05-06
Fixes #37408 - remove metadata_generate_needed?

ianballou:
Remove an unneeded method.

It has a typo anyway!

Check that nothing actually used this method.


chris1984 (approved):
Ship it

PR #10988, opened 2024-05-07
Fixes #37407 - reindex container manifests to populate label information

sjha4:
On a box with older pulp create a few docker repos with labels and annotations.
Upgrade your pulp.
Run the pulpcore container-handle-image-data migration
Run bundle exec rails katello:import_docker_manifest_labels and verify that katello manifests pull in all the new field info.

### DIFF
--- a/app/services/katello/content_unit_indexer.rb
+++ b/app/services/katello/content_unit_indexer.rb
@@ -55,6 +55,15 @@ module Katello
       end
     end
 
+    def reimport_units
+      units_from_pulp.each do |units|
+        to_update = units.map do |unit|
+          @service_class.generate_model_row(unit)
+        end
+        @model_class.upsert_all(to_update, unique_by: :pulp_id)
+      end
+    end
+
     def import_associations(units)
       pulp_id_to_id = self.class.pulp_id_to_id_map(@content_type, units.map { |unit| unit[@service_class.unit_identifier] })
       @service_class.insert_child_associations(units, pulp_id_to_id) if @service_class.respond_to?(:insert_child_associations)
--- a/lib/katello/tasks/repository.rake
+++ b/lib/katello/tasks/repository.rake
@@ -71,6 +71,12 @@ namespace :katello do
     end
   end
 
+  desc "Re-import all container manifests to populate labels and annotations."
+  task :import_container_manifest_labels => ["dynflow:client", "check_ping"] do
+    User.current = User.anonymous_api_admin
+    handle_manifest_label_updates
+  end
+
   def lookup_repositories
     lifecycle_envs = Katello::KTEnvironment.where(:name => ENV['LIFECYCLE_ENVIRONMENT']) if ENV['LIFECYCLE_ENVIRONMENT']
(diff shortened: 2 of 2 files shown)


app/services/katello/content_unit_indexer.rb
```diff
@@ -21,7 +21,7 @@ def remove_duplicates(unit)
       end
     end
 
-    def import_all(filtered_indexing = false)
+    def import_all(filtered_indexing = false, override_mutable = false)
```
ianballou, line 24: Is `override_mutable` used at all?

app/services/katello/content_unit_indexer.rb
```diff
@@ -55,6 +55,15 @@ def import_all(filtered_indexing = false)
       end
     end
 
+    def import_manifest_labels
```
ianballou, line 58: This method ended up being totally generic, so what do you think about somethink like:

```suggestion
    def reimport_by_type
```

sjha4: Good point..I landed at this version after trying some different ways. Didn't realize it became generic. :)

ianballou:
I did a performance  test with 120 repositories and 26295 manifests:

```
[vagrant@centos8-stream-katello-4-12 ~]$ time sudo foreman-rake katello:import_docker_manifest_labels

real	23m44.868s
user	10m2.752s
sys	0m20.095s
```

Do you think there would be any benefit from only indexing manifests with media types `application/vnd.docker.distribution.manifest.v2+json` and `application/vnd.oci.image.manifest.v1+json`

ianballou:
There also might be many manifests without labels or annotations -- I wonder if we could filter those out in Pulp somehow.

ianballou:
Looks like in Pulp we cannot  filter by labels or annotations, so I think the only improvement we could do API-wise would be to ignore `application/vnd.docker.distribution.manifest.v1+json` manifests. But even in my case, that wouldn't have gained us any more performance.

lib/katello/tasks/repository.rake
```diff
@@ -108,6 +114,18 @@ namespace :katello do
     end
   end
 
+  def handle_manifest_label_updates
+    batch_size = 1000
+    Katello::DockerManifest.where(schema_version: 2).find_in_batches(batch_size: batch_size) do |group|
```
ianballou, line 119: Perhaps we should filter out manifests with annotations, labels, or is_flatpak / is_bootable set to true, since users might have manifests that were synced after the change.

ianballou: This is just for the users who don't run the migration right away.

lib/katello/tasks/repository.rake
```diff
@@ -71,6 +71,12 @@ namespace :katello do
     end
   end
 
+  desc "Re-import all docker manifests to populate labels and annotations."
+  task :import_docker_manifest_labels => ["dynflow:client", "check_ping"] do
```
ianballou, line 75: What do you think about saying `container manifest` instead of `docker manifest` ? While our code still refers to docker, I think the public facing stuff could all say container.

ianballou (commented):
Looking good now, just one small comment

ianballou (approved):
Looks good to me!

PR #10993, opened 2024-05-13
Fixes #37442 - Sync status page presenter doesn't work

sjha4:
Fix sync status page presenter
1. Sync some repos from the Sync status page.
2. Ensure you see the presenter and updates on an active sync.


ianballou (approved):
Working great.

image

evgeni:
Now imagine the same fix had to be done for a React page. ;-)

PR #10995, opened 2024-05-14
Fixes #37455 - Create Katello push repositories as needed at container push time, handle non-auth validation.

qcjames53:

- When using podman to push containers to Katello, ensure that the container name is of a valid format:
  - <org_label>/<product_label>/<container_name>[:tag] (down-case all letters)
  -  id/<org_id>/<product_id>/<container_name>[:tag]
- Limits to container names:
  - Org and product fields must always correspond to a valid, existing org and product, respectively. No fields can be blank or omitted.
  - Org and product labels must be downcased. Container names must be lowercase and contain only letters and underscores (podman rejects these automatically).
  - Org and product labels cannot be ambiguous (for example 'foo_bar' and 'Foo_Bar' both map to 'foo_bar'). For ambiguous orgs or products, the user is directed to use the id format.
  - Org and product IDs must be integers with no leading zeros.
  - A user must stick to a specific format for each org, product, name combination. Pushes are not permitted to alternate between label and id or vice versa.
  - If a label format push becomes ambiguous due to external changes, the user will be directed to destroy either the offending org/product or will be directed to destroy the existing push container and re-push using the id format.
- Create a new push repository matching the container name under the appropriate product where required.
- Update existing push repos where required.
- Run content indexing after completion of the container upload.
- Ensure the root repository fields are correct after upload:
  - name: matches container_name
  - label: matches container_name
  - content_type: docker
  - is_container_push (new field): true
  - container_push_name (new field): name used in podman command e.g. "id/0/0/foo"
  - container_push_name_format (new field): name format used, `label` or `id` 
- Ensure that the instance repository fields are correct after upload:
  - relative_path: matches the container_push_name above
  - container_repository_name: matches the container_push_name above
  - environment_id: correct for product
  - content_view_version_id: correct for product
  - latest_version_href: pulp latest version href
  - pulp_href: pulp href

1. Pull a few different podman images:
 ```
podman pull quay.io/aptible/hello-world
 ```
 2. Launch katello and create an empty product. Log in with podman:
 ```
 podman login <url>
 ```
 3. Get the image ID for one of the test podman images. Push that image to katello using the correct organization label and product label. Make sure to down-case the labels or podman will reject the container name:
 ```
 podman image list
 podman push <img_id> <url>/<organization_label>/<product_label>/<container_name>
 ```
4. Use the rake console to ensure the root repository and repository were created properly:
```
bundle exec rake console
::Katello::Products.find(<id>).root_repositories.where(label: "<container_name>").first
::Katello::Products.find(<id>).root_repositories.where(label: "<container_name>").first.library_instance
```
5. Repeat steps 3 and 4 using the id push format. You will need to change the container name.
```
podman push <img_id> <url>/id/<organization_id>/<product_id>/<container_name>
```
6. Repeat steps 3 and 4 with a [:tag] field. This should work as usual.
7. Repeat steps 3 and 4 while trying every error in the above 'Limits to container names' list. This is an extensive but not exhaustive list of all the error handling included. List duplicated below:
  - Org and product fields must always correspond to a valid, existing org and product, respectively. No fields can be blank or omitted.
  - Org and product labels must be downcased. Container names must be lowercase and contain only letters and underscores (podman rejects these automatically).
  - Org and product labels cannot be ambiguous (for example 'foo_bar' and 'Foo_Bar' both map to 'foo_bar'). For ambiguous orgs or products, the user is directed to use the id format.
  - Org and product IDs must be integers with no leading zeros.
  - A user must stick to a specific format for each org, product, name combination. Pushes are not permitted to alternate between label and id or vice versa for a given container.
  - If a label format push becomes ambiguous due to external changes, the user will be directed to destroy either the offending org/product or will be directed to destroy the existing push container and re-push using the id format.
8. Validate the test cases for coverage and accuracy.

### DIFF
--- a/app/controllers/katello/api/registry/registry_proxies_controller.rb
+++ b/app/controllers/katello/api/registry/registry_proxies_controller.rb
@@ -3,15 +3,14 @@ module Katello
   class Api::Registry::RegistryProxiesController < Api::V2::ApiController
     before_action :disable_strong_params
     before_action :confirm_settings
-    before_action :confirm_push_settings, only: [:start_upload_blob, :upload_blob, :finish_upload_blob,
-                                                 :push_manifest]
     skip_before_action :authorize
     before_action :optional_authorize, only: [:token, :catalog]
     before_action :registry_authorize, except: [:token, :v1_search, :catalog]
     before_action :authorize_repository_read, only: [:pull_manifest, :tags_list]
-    # TODO: authorize_repository_write commented out until Katello indexes Pulp container push repositories.
-    # before_action :authorize_repository_write, only: [:start_upload_blob, :upload_blob, :finish_upload_blob,
-    #                                                   :push_manifest]
+    # TODO: authorize_repository_write commented out due to container push changes. Additional task needed to fix.
+    # before_action :authorize_repository_write, only: [:start_upload_blob, :upload_blob, :finish_upload_blob, :push_manifest]
+    before_action :container_push_prop_validation, only: [:start_upload_blob, :upload_blob, :finish_upload_blob, :push_manifest]
+    before_action :create_container_repo_if_needed, only: [:start_upload_blob, :upload_blob, :finish_upload_blob, :push_manifest]
     skip_before_action :check_media_type, only: [:start_upload_blob, :upload_blob, :finish_upload_blob,
                                                  :push_manifest]
 
@@ -83,6 +82,301 @@ module Katello
       redirect_authorization_headers
       render_error('unauthorized', :status => :unauthorized)
       return false
+    end
+
+    def container_push_prop_validation(props = nil)
+      # Handle validation and repo creation for container pushes before talking to pulp
+      return false unless confirm_push_settings
+      props = parse_blob_push_props if props.nil?
+      return false unless check_blob_push_field_syntax(props)
+
+      # validate input and find the org and product either using downcase label or id
+      if props[:schema] == "label"
+        return false unless check_blob_push_org_label(props)
+        return false unless check_blob_push_product_label(props)
+      else
+        return false unless check_blob_push_org_id(props)
+        return false unless check_blob_push_product_id(props)
+      end
+
+      return false unless check_blob_push_container(props)
+      true
+    end
+
+    def parse_blob_push_props(path_string = nil)
+      # path string should follow one of these formats:
+      #   - /v2/{org_label}/{product_label}/{name}/blobs/uploads...
+      #   - /v2/id/{org_id}/{product_id}/{name}/blobs/uploads...
+      #   - /v2/{org_label}/{product_label}/{name}/manifests/...
+      #   - /v2/id/{org_id}/{product_id}/{name}/manifests/...
+      # inputs not matching format will return {valid_format: false}
+      path_string = @_request.fullpath if path_string.nil?
+      segments = path_string.split('/')
+
+      if segments.length >= 7 && segments[0] == "" && segments[1] == "v2" &&
+        segments[2] != "id" && (segments[5] == "blobs" || segments[5] == "manifests")
+
+        return {
+          valid_format: true,
+          schema: "label",
+          organization: segments[2],
+          product: segments[3],
+          name: segments[4]
+        }
+      elsif segments.length >= 8 && segments[0] == "" && segments[1] == "v2" &&
+        segments[2] == "id" && (segments[6] == "blobs" || segments[6] == "manifests")
+
+        return {
+          valid_format: true,
+          schema: "id",
+          organization: segments[3],
+          product: segments[4],
+          name: segments[5]
+        }
+      else
+        return {valid_format: false}
+      end
+    end
+
+    def check_blob_push_field_syntax(props)
+      # check basic url field syntax
+      unless props[:valid_format]
+        return render_podman_error(
+          "NAME_INVALID",
+          "Invalid format. Container pushes should follow 'organization_label/product_label/name' OR 'id/organization_id/product_id/name' schema.",
+          :bad_request
+        )
+      end
+      return true
+    end
+
+    # rubocop:disable Metrics/MethodLength
+    def check_blob_push_org_label(props)
+      org_label = props[:organization]
+      unless org_label.present? && org_label.length > 0
+        return render_podman_error(
+          "NAME_INVALID",
+          "Invalid format. Organization label cannot be blank.",
+          :bad_request
+        )
+      end
+      org = Organization.where("LOWER(label) = '#{org_label}'") # convert to lowercase
+      # reject ambiguous orgs (possible due to lowercase conversion)
+      if org.length > 1
+        # Determine if the repo already exists in one of the possible products. If yes,
+        # inform the user they need to destroy the existing repo and use the ID format
+        unless props[:product].blank? || props[:name].blank?
+          org.each do |o|
+            products = get_matching_products_from_org(o, props[:product])
+            products.each do |prod|
+              root_repos = get_root_repo_from_product(prod, props[:name])
+              unless root_repos.empty?
+                return render_podman_error(
+                  "NAME_INVALID",
+                  "Due to a change in your organizations, this container name has become "\
+                    "ambiguous (org name '#{org_label}'). If you wish to continue using this "\
+                    "container name, destroy the organization in conflict with '#{o.name} (id "\
+                    "#{o.id}). If you wish to keep both orgs, destroy '#{o.label}/#{prod.label}/"\
+                    "#{root_repos.first.label}' and retry your push using the id format.",
+                  :conflict
+                )
+              end
+            end
+          end
+        end
+
+        # Otherwise tell them to try pushing with ID format
+        return render_podman_error(
+          "NAME_INVALID",
+          "Organization label '#{org_label}' is ambiguous. Try using an id-based container name.",
+          :conflict
+        )
+      end
+      if org.length == 0
+        return render_podman_error(
+          "NAME_UNKNOWN",
+          "Organization not found: '#{org_label}'",
+          :not_found
+        )
+      end
+      @organization = org.first
+      true
+    end
+
+    def check_blob_push_org_id(props)
+      org_id = props[:organization]
+      unless org_id.present? && org_id == org_id.to_i.to_s
+        return render_podman_error(
+          "NAME_INVALID",
+          "Invalid format. Organization id must be an integer without leading zeros.",
+          :bad_request
+        )
+      end
+      @organization = Organization.find_by_id(org_id.to_i)
+      if @organization.nil?
+        return render_podman_error(
+          "NAME_UNKNOWN",
+          "Organization id not found: '#{org_id}'",
+          :not_found
+        )
+      end
+      true
+    end
+
+    def check_blob_push_product_label(props)
+      prod_label = props[:product]
+      unless prod_label.present? && prod_label.length > 0
+        return render_podman_error(
+          "NAME_INVALID",
+          "Invalid format. Product label cannot be blank.",
+          :bad_request
+        )
+      end
+      product = get_matching_products_from_org(@organization, prod_label)
+      # reject ambiguous products (possible due to lowercase conversion)
+      if product.length > 1
+        # Determine if the repo already exists in one of the possible products. If yes,
+        # inform the user they need to destroy the existing repo and use the ID format
+        unless props[:name].blank?
+          product.each do |prod|
+            root_repos = get_root_repo_from_product(prod, props[:name])
+            unless root_repos.empty?
+              return render_podman_error(
+                "NAME_INVALID",
+                "Due to a change in your products, this container name has become ambiguous "\
+                  "(product name '#{prod_label}'). If you wish to continue using this container "\
+                  "name, destroy the product in conflict with '#{prod.name}' (id #{prod.id}). If "\
+                  "you wish to keep both products, destroy '#{@organization.label}/#{prod.label}/"\
+                  "#{root_repos.first.label}' and retry your push using the id format.",
+                :conflict
+              )
+            end
+          end
+        end
+
+        return render_podman_error(
+          "NAME_INVALID",
+          "Product label '#{prod_label}' is ambiguous. Try using an id-based container name.",
+          :conflict
+        )
+      end
+      if product.length == 0
+        return render_podman_error(
+          "NAME_UNKNOWN",
+          "Product not found: '#{prod_label}'",
+          :not_found
+        )
+      end
+      @product = product.first
+      true
+    end
+
+    def check_blob_push_product_id(props)
+      prod_id = props[:product]
+      unless prod_id.present? && prod_id == prod_id.to_i.to_s
+        return render_podman_error(
+          "NAME_INVALID",
+          "Invalid format. Product id must be an integer without leading zeros.",
+          :bad_request
+        )
+      end
+      @product = @organization.products.find_by_id(prod_id.to_i)
+      if @product.nil?
+        return render_podman_error(
+          "NAME_UNKNOWN",
+          "Product id not found: '#{prod_id}'",
+          :not_found
+        )
+      end
+      true
+    end
+
+    def get_matching_products_from_org(organization, product_label)
+      return organization.products.where("LOWER(label) = '#{product_label}'") # convert to lowercase
+    end
+
+    def get_root_repo_from_product(product, root_repo_name)
+      return product.root_repositories.where(label: root_repo_name)
+    end
+
+    def check_blob_push_container(props)
+      unless props[:name].present? && props[:name].length > 0
+        return render_podman_error(
+          "NAME_INVALID",
+          "Invalid format. Container name cannot be blank.",
+          :bad_request
+        )
+      end
+
+      @container_name = props[:name]
+      @container_push_name_format = props[:schema]
+      if @container_push_name_format == "label"
+        @container_path_input = "#{props[:organization]}/#{props[:product]}/#{props[:name]}"
+      else
+        @container_path_input = "id/#{props[:organization]}/#{props[:product]}/#{props[:name]}"
+      end
+
+      # If the repo already exists, check if the existing push format matches
+      root_repo = get_root_repo_from_product(@product, @container_name).first
+      if !root_repo.nil? && @container_push_name_format != root_repo.container_push_name_format
+        return render_podman_error(
+          "NAME_INVALID",
+          "Repository name '#{@container_name}' already exists in this product using a different naming scheme. Please retry your request with the #{root_repo.container_push_name_format} format or destroy and recreate the repository using your preferred schema.",
+          :conflict
+        )
+      end
+
+      true
+    end
+
+    def create_container_repo_if_needed
+      if get_root_repo_from_product(@product, @container_name).empty?
+        root = @product.add_repo(
+          name: @container_name,
+          label: @container_name,
+          download_policy: 'immediate',
+          content_type: Repository::DOCKER_TYPE,
+          unprotected: true,
+          is_container_push: true,
+          container_push_name: @container_path_input,
+          container_push_name_format: @container_push_name_format
+        )
+        sync_task(::Actions::Katello::Repository::CreateRoot, root, @container_path_input)
+      end
+    end
+
+    def blob_push_cleanup
+      # after manifest upload, index content and set version href using pulp api
+      root_repo = get_root_repo_from_product(@product, @container_name)&.first
+      instance_repo = root_repo&.library_instance
+
+      unless root_repo.present? && instance_repo.present?
+        return render_podman_error(
+          "BLOB_UPLOAD_UNKNOWN",
+          "Could not locate local uploaded repository for content indexing.",
+          :not_found
+        )
+      end
+
+      api = ::Katello::Pulp3::Repository.api(SmartProxy.pulp_primary, ::Katello::Repository::DOCKER_TYPE).container_push_api
+      api_response = api.list(name: @container_path_input)&.results&.first
+      latest_version_href = api_response&.latest_version_href
+      pulp_href = api_response&.pulp_href
+
+      if latest_version_href.empty? || pulp_href.empty?
+        return render_podman_error(
+          "BLOB_UPLOAD_UNKNOWN",
+          "Could not locate repository properties for content indexing.",
+          :not_found
+        )
+      end
+
+      instance_repo.update!(version_href: latest_version_href)
+      ::Katello::Pulp3::RepositoryReference.where(root_repository_id: instance_repo.root_id,
+        content_view_id: instance_repo.content_view.id, repository_href: pulp_href).create!
+      instance_repo.index_content
+
+      true
     end
 
     def find_writable_repository
@@ -205,17 +499,6 @@ module Katello
       redirect_client { Resources::Registry::Proxy.get(@_request.fullpath, headers, max_redirects: 0) }
     end
 
-    def push_manifest
-      headers = translated_headers_for_proxy
-      headers['Content-Type'] = request.headers['Content-Type'] if request.headers['Content-Type']
-      body = @_request.body.read
-      pulp_response = Resources::Registry::Proxy.put(@_request.fullpath, body, headers)
-      pulp_response.headers.each do |key, value|
-        response.header[key.to_s] = value
-      end
-      head pulp_response.code
-    end
-
     def start_upload_blob
       headers = translated_headers_for_proxy
       headers['Content-Type'] = request.headers['Content-Type'] if request.headers['Content-Type']
@@ -225,6 +508,7 @@ module Katello
       pulp_response.headers.each do |key, value|
         response.header[key.to_s] = value
       end
+
       head pulp_response.code
     end
 
@@ -264,6 +548,21 @@ module Katello
       pulp_response.headers.each do |key, value|
         response.header[key.to_s] = value
       end
+
+      head pulp_response.code
+    end
+
+    def push_manifest
+      headers = translated_headers_for_proxy
+      headers['Content-Type'] = request.headers['Content-Type'] if request.headers['Content-Type']
+      body = @_request.body.read
+      pulp_response = Resources::Registry::Proxy.put(@_request.fullpath, body, headers)
+      pulp_response.headers.each do |key, value|
+        response.header[key.to_s] = value
+      end
+
+      cleanup_result = blob_push_cleanup if pulp_response.code.between?(200, 299)
+      return false unless cleanup_result
 
       head pulp_response.code
     end
@@ -423,8 +722,11 @@ module Katello
 
     def confirm_push_settings
       return true if SETTINGS.dig(:katello, :container_image_registry, :allow_push)
-      render_error('custom_error', :status => :not_found,
-                   :locals => { :message => "Registry push not supported" })
+      render_podman_error(
+        "UNSUPPORTED",
+        "Registry push is not enabled. To enable, add ':katello:'->':container_image_registry:'->':allow_push: true' in the katello settings file.",
+        :unprocessable_entity
+      )
     end
 
     def request_url
@@ -446,10 +748,19 @@ module Katello
       Rails.logger.debug "With body: #{filter_sensitive_data(response.body)}\n" unless route_name == 'pull_blob'
     end
 
+    def render_podman_error(code, message, status = :bad_request)
+      # Renders a podman-compatible error and returns false.
+      #     code: uppercase string code from opencontainer error code spec:
+      #         https://specs.opencontainers.org/distribution-spec/?v=v1.0.0#DISTRIBUTION-SPEC-140
+      #     message: a custom error string
+      #     status: a symbol in the 400 block of the rails response code table:
+      #         https://guides.rubyonrails.org/layouts_and_rendering.html#the-status-option
+      render json: {errors: [{code: code, message: message}]}, status: status
+      false
+    end
+
     def item_not_found(item)
-      msg = "#{item} was not found!"
-      # returning errors based on registry specifications in https://docs.docker.com/registry/spec/api/#errors
-      render json: {errors: [code: :invalid_request, message: msg, details: msg]}, status: :not_found
+      render_podman_error("NAME_UNKNOWN", "#{item} was not found!", :not_found)
     end
   end
 end
--- a/app/lib/actions/katello/repository/create.rb
+++ b/app/lib/actions/katello/repository/create.rb
@@ -13,10 +13,13 @@ module Actions
 
           org = repository.organization
           sequence do
-            create_action = plan_action(Pulp3::Orchestration::Repository::Create,
-                                        repository, SmartProxy.pulp_primary, force_repo_create)
-
-            return if create_action.error
+            # Container push repositories will already be in pulp. The version_href is
+            # directly updated after a push.
+            unless root.is_container_push
+              create_action = plan_action(Pulp3::Orchestration::Repository::Create,
+                                          repository, SmartProxy.pulp_primary, force_repo_create)
+              return if create_action.error
+            end
 
             # when creating a clone, the following actions are handled by the
             # publish/promote process
@@ -32,13 +35,16 @@ module Actions
               end
             end
 
-            concurrence do
-              plan_self(:repository_id => repository.id, :clone => clone)
-              if !clone && repository.url.present?
-                repository.product.alternate_content_sources.with_type(repository.content_type).each do |acs|
-                  acs.smart_proxies.each do |smart_proxy|
-                    smart_proxy_acs = ::Katello::SmartProxyAlternateContentSource.create(alternate_content_source_id: acs.id, smart_proxy_id: smart_proxy.id, repository_id: repository.id)
-                    plan_action(Pulp3::Orchestration::AlternateContentSource::Create, smart_proxy_acs)
+            # Container push repos do not need metadata generation or ACS (they do not sync)
+            unless root.is_container_push
+              concurrence do
+                plan_self(:repository_id => repository.id, :clone => clone)
+                if !clone && repository.url.present?
+                  repository.product.alternate_content_sources.with_type(repository.content_type).each do |acs|
+                    acs.smart_proxies.each do |smart_proxy|
+                      smart_proxy_acs = ::Katello::SmartProxyAlternateContentSource.create(alternate_content_source_id: acs.id, smart_proxy_id: smart_proxy.id, repository_id: repository.id)
+                      plan_action(Pulp3::Orchestration::AlternateContentSource::Create, smart_proxy_acs)
+                    end
                   end
                 end
               end
--- a/app/models/katello/repository.rb
+++ b/app/models/katello/repository.rb
@@ -118,7 +118,7 @@ module Katello
     end}
 
     before_validation :set_pulp_id
-    before_validation :set_container_repository_name, :if => :docker?
+    before_validation :set_container_repository_name, :unless => :skip_container_name?
 
     scope :has_url, -> { joins(:root).where.not("#{RootRepository.table_name}.url" => nil) }
     scope :on_demand, -> { joins(:root).where("#{RootRepository.table_name}.download_policy" => ::Katello::RootRepository::DOWNLOAD_ON_DEMAND) }
@@ -266,6 +266,10 @@ module Katello
 
     def content_view
       self.content_view_version.content_view
+    end
+
+    def skip_container_name?
+      self.library_instance? && self.root.docker? && self.root.is_container_push
     end
 
     def library_instance?
--- a/app/models/katello/root_repository.rb
+++ b/app/models/katello/root_repository.rb
@@ -112,6 +112,8 @@ module Katello
         only_integer: true
       }
 
+    validates :container_push_name_format, inclusion: { in: ['label', 'id'].freeze, allow_nil: true}
+
     scope :subscribable, -> { where(content_type: RootRepository::SUBSCRIBABLE_TYPES) }
     scope :skipable_metadata_check, -> { where(content_type: RootRepository::SKIPABLE_METADATA_TYPES) }
     scope :has_url, -> { where.not(:url => nil) }
@@ -195,7 +197,7 @@ module Katello
     def ensure_docker_repo_unprotected
       unless unprotected
         errors.add(:base, _("Container Image Repositories are not protected at this time. " \
-                             "They need to be published via http to be available to containers."))
+                            "They need to be published via http to be available to containers."))
       end
     end
 
--- a/app/lib/actions/katello/repository/create_root.rb
+++ b/app/lib/actions/katello/repository/create_root.rb
@@ -2,13 +2,15 @@ module Actions
   module Katello
     module Repository
       class CreateRoot < Actions::EntryAction
-        def plan(root)
+        def plan(root, relative_path = nil)
           root.save!
           repository = ::Katello::Repository.new(:environment => root.organization.library,
                                       :content_view_version => root.organization.library.default_content_view_version,
                                       :root => root)
-          repository.relative_path = repository.custom_repo_path
+          repository.container_repository_name = relative_path
+          repository.relative_path = relative_path || repository.custom_repo_path
           repository.save!
+
           action_subject(repository)
           plan_action(::Actions::Katello::Repository::Create, repository)
         end
--- a/app/models/katello/glue/pulp/repos.rb
+++ b/app/models/katello/glue/pulp/repos.rb
@@ -133,7 +133,14 @@ module Katello
 
         repo_param[:mirroring_policy] = Katello::RootRepository::MIRRORING_POLICY_ADDITIVE if repo_param[:mirroring_policy].blank?
 
-        RootRepository.new(repo_param.merge(:product_id => self.id))
+        repo_param = repo_param.merge(:product_id => self.id)
+
+        # Container push may concurrently call root add several times before the db can update.
+        if repo_param[:is_container_push]
+          RootRepository.create_or_find_by!(repo_param)
+        else
+          RootRepository.new(repo_param)
+        end
       end
     end
   end
--- a/app/services/katello/pulp3/api/docker.rb
+++ b/app/services/katello/pulp3/api/docker.rb
@@ -15,6 +15,10 @@ module Katello
         def recursive_add_api
           PulpContainerClient::ContainerRecursiveAddApi.new(api_client)
         end
+
+        def container_push_api
+          PulpContainerClient::RepositoriesContainerPushApi.new(api_client)
+        end
       end
     end
   end
--- a/db/migrate/20240520142245_add_container_push_props_to_repo.rb
+++ b/db/migrate/20240520142245_add_container_push_props_to_repo.rb
@@ -1,0 +1,7 @@
+class AddContainerPushPropsToRepo < ActiveRecord::Migration[6.1]
+  def change
+    add_column :katello_root_repositories, :is_container_push, :boolean, default: false
+    add_column :katello_root_repositories, :container_push_name, :string, default: nil
+    add_column :katello_root_repositories, :container_push_name_format, :string, default: nil
+  end
+end
--- a/test/controllers/api/registry/registry_proxies_controller_test.rb
+++ b/test/controllers/api/registry/registry_proxies_controller_test.rb
@@ -536,11 +536,16 @@ module Katello
 
     describe 'container push' do
       it 'starts a blob upload to Pulp' do
-        repo_name = 'a repo'
+        repo_name = 'test_org/test_product/test_name'
+        @controller.expects(:check_blob_push_org_label).returns(true)
+        @controller.expects(:check_blob_push_product_label).returns(true)
+        @controller.expects(:check_blob_push_container).returns(true)
+        @controller.expects(:create_container_repo_if_needed)
+
         # The content type should be octet-stream, but action controller
         # throws a non-existence error since Mime::Type.lookup('application/octet-stream').to_sym is nil.
         Resources::Registry::Proxy.expects(:post).with(
-              "/v2/#{ERB::Util.url_encode(repo_name)}/blobs/uploads",
+              "/v2/#{repo_name}/blobs/uploads",
               is_a(StringIO),
               has_entries('MOCK-TEST' => 123,
                           'Content-Type' => 'application/json',
@@ -549,13 +554,18 @@ module Katello
         request.env['HTTP_MOCK_TEST'] = 123
         request.headers['Content-Type'] = 'application/json'
         resp = post :start_upload_blob, params: { repository: repo_name }
-        assert_equal 'Mars', resp.headers['Location']
         assert_equal resp.code, '202'
+        assert_equal 'Mars', resp.headers['Location']
       end
 
       it 'uploads a blob chunk' do
-        repo_name = 'repo'
+        repo_name = 'test_org/test_product/test_name'
         uuid = 'uuid'
+        @controller.expects(:check_blob_push_org_label).returns(true)
+        @controller.expects(:check_blob_push_product_label).returns(true)
+        @controller.expects(:check_blob_push_container).returns(true)
+        @controller.expects(:create_container_repo_if_needed)
+
         # The content type should be octet-stream, but action controller
         # throws a non-existence error since Mime::Type.lookup('application/octet-stream').to_sym is nil.
         Resources::Registry::Proxy.expects(:patch).with(
@@ -571,13 +581,18 @@ module Katello
         request.headers['Content-Range'] = 'bytes 500-1000/65989'
         request.env['HTTP_MOCK_TEST'] = 123
         resp = patch :upload_blob, params: { repository: repo_name, uuid: uuid }
-        assert_equal 'Mars', resp.headers['Location']
         assert_equal resp.code, '202'
+        assert_equal 'Mars', resp.headers['Location']
       end
 
       it 'finishes a blob upload to Pulp' do
-        repo_name = 'repo'
+        repo_name = 'test_org/test_product/test_name'
         uuid = 'uuid'
+        @controller.expects(:check_blob_push_org_label).returns(true)
+        @controller.expects(:check_blob_push_product_label).returns(true)
+        @controller.expects(:check_blob_push_container).returns(true)
+        @controller.expects(:create_container_repo_if_needed)
+
         # The content type should be octet-stream, but action controller
         # throws a non-existence error since Mime::Type.lookup('application/octet-stream').to_sym is nil.
         Resources::Registry::Proxy.expects(:put).with(
@@ -592,13 +607,19 @@ module Katello
         request.headers['Content-Type'] = 'application/json'
         request.headers['Content-Range'] = 'bytes 500-1000/65989'
         resp = put :finish_upload_blob, params: { repository: repo_name, uuid: uuid }
-        assert_equal 'Mars', resp.headers['Location']
         assert_equal resp.code, '201'
+        assert_equal 'Mars', resp.headers['Location']
       end
 
       it 'pushes a manifest' do
-        repo_name = 'repo'
+        repo_name = 'test_org/test_product/test_name'
         tag = 'latest'
+        @controller.expects(:check_blob_push_org_label).returns(true)
+        @controller.expects(:check_blob_push_product_label).returns(true)
+        @controller.expects(:check_blob_push_container).returns(true)
+        @controller.expects(:create_container_repo_if_needed)
+        @controller.expects(:blob_push_cleanup).returns(true)
+
         # The content type should be application/vnd.oci.image.manifest.v1+json, but action controller
         # throws a non-existence error since Mime::Type.lookup('application/octet-stream').to_sym is nil.
         Resources::Registry::Proxy.expects(:put).with(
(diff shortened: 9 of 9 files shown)


app/models/katello/repository.rb
```diff
@@ -261,6 +261,7 @@ def self.attribute_name
     end
 
     def set_container_repository_name
+      # TODO - fix this name render for container push
```
qcjames53, line 264: @ianballou As of now this follows the LE pattern. That is one option, but the container_repository_name could also be this format: <org_label>/<product_label>/<container_name> (including capitals). If it doesn't matter at all I'd like to follow LE pattern since it's significantly easier to leave as-is.

qcjames53:
The errors thrown are a little messy. They are all generic errors and they tend to get truncated on the podman side. I'm happy to improve them if you think it's a good idea.

qcjames53:
Containers with tags such as `default_organization/dummy_product/foo:bar` end up losing the tag when pushed. I think this is okay since the tag is on the podman side; however, this means that users would technically need to pull using a different name than the name they pushed with. This may need further discussion since I could see users expecting either behavior (tag loss vs tag specification on pull).

ianballou (commented):
I saw this error on my first push:
```
container_repository_name: ["for repository 'fedora-bootc' is not unique and cannot be created in 'Library'. Its Container Repository Name (default_organization-buttermilk_biscuits-fedora-bootc) conflicts with an existing repository.  Consider changing the Lifecycle Environment's Registry Name Pattern to something more specific."]
```

I do have another container repo, but its label is `centos-bootc_fedora-bootc`

The error didn't throw until later. Was repo creation attempted twice maybe?

I pushed: 
```
podman push 75eca4dcedfb `hostname`/default_organization/buttermilk_biscuits/fedora-bootc
```

qcjames53:
> I saw this error on my first push:
> 
> ```
> container_repository_name: ["for repository 'fedora-bootc' is not unique and cannot be created in 'Library'. Its Container Repository Name (default_organization-buttermilk_biscuits-fedora-bootc) conflicts with an existing repository.  Consider changing the Lifecycle Environment's Registry Name Pattern to something more specific."]
> ```
> 
> I do have another container repo, but its label is `centos-bootc_fedora-bootc`
> 
> The error didn't throw until later. Was repo creation attempted twice maybe?
> 
> I pushed:
> 
> ```
> podman push 75eca4dcedfb `hostname`/default_organization/buttermilk_biscuits/fedora-bootc
> ```

Well shoot. Podman calls the endpoint that calls `start_upload_blob` multiple times. Sometimes once, sometimes 3 or 4 times. Big pain in the butt. If I were a betting person, the logic in that method is getting tripped up and is unable to find the repo created on the previous run and is trying to create the repo again.

Another option is that the `centos-bootc_fedora-bootc` is named `fedora_bootc` because of a weird LE pattern?

ianballou:
I deleted all of my repos and tried again, ended up hitting:

```
2024-05-14T21:18:25 [E|kat|2a1f0b55] NoMethodError: undefined method `library_instance' for nil:NilClass
 2a1f0b55 | /home/vagrant/katello/app/controllers/katello/api/registry/registry_proxies_controller.rb:397:in `push_manifest'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_controller/metal/basic_implicit_render.rb:6:in `send_action'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/abstract_controller/base.rb:228:in `process_action'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_controller/metal/rendering.rb:30:in `process_action'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/abstract_controller/callbacks.rb:42:in `block in process_action'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/activesupport-6.1.7.7/lib/active_support/callbacks.rb:117:in `block in run_callbacks'
 2a1f0b55 | /home/vagrant/katello/app/controllers/katello/api/registry/registry_proxies_controller.rb:24:in `repackage_message'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/activesupport-6.1.7.7/lib/active_support/callbacks.rb:126:in `block in run_callbacks'
 2a1f0b55 | /home/vagrant/foreman/app/controllers/concerns/foreman/controller/timezone.rb:10:in `set_timezone'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/activesupport-6.1.7.7/lib/active_support/callbacks.rb:126:in `block in run_callbacks'
 2a1f0b55 | /home/vagrant/foreman/app/models/concerns/foreman/thread_session.rb:32:in `clear_thread'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/activesupport-6.1.7.7/lib/active_support/callbacks.rb:126:in `block in run_callbacks'
 2a1f0b55 | /home/vagrant/foreman/app/controllers/concerns/foreman/controller/topbar_sweeper.rb:12:in `set_topbar_sweeper_controller'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/activesupport-6.1.7.7/lib/active_support/callbacks.rb:126:in `block in run_callbacks'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/audited-5.6.0/lib/audited/sweeper.rb:16:in `around'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/activesupport-6.1.7.7/lib/active_support/callbacks.rb:126:in `block in run_callbacks'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/audited-5.6.0/lib/audited/sweeper.rb:16:in `around'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/activesupport-6.1.7.7/lib/active_support/callbacks.rb:126:in `block in run_callbacks'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/activesupport-6.1.7.7/lib/active_support/callbacks.rb:137:in `run_callbacks'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/abstract_controller/callbacks.rb:41:in `process_action'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_controller/metal/rescue.rb:22:in `process_action'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_controller/metal/instrumentation.rb:34:in `block in process_action'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/activesupport-6.1.7.7/lib/active_support/notifications.rb:203:in `block in instrument'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/activesupport-6.1.7.7/lib/active_support/notifications/instrumenter.rb:24:in `instrument'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/activesupport-6.1.7.7/lib/active_support/notifications.rb:203:in `instrument'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_controller/metal/instrumentation.rb:33:in `process_action'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_controller/metal/params_wrapper.rb:249:in `process_action'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/activerecord-6.1.7.7/lib/active_record/railties/controller_runtime.rb:27:in `process_action'
 2a1f0b55 | /home/vagrant/katello/app/controllers/katello/api/registry/registry_proxies_controller.rb:581:in `call'
 2a1f0b55 | /home/vagrant/katello/app/controllers/katello/api/registry/registry_proxies_controller.rb:581:in `process_action'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/abstract_controller/base.rb:165:in `process'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionview-6.1.7.7/lib/action_view/rendering.rb:39:in `process'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_controller/metal.rb:190:in `dispatch'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_controller/metal.rb:254:in `dispatch'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/routing/route_set.rb:50:in `dispatch'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/routing/route_set.rb:33:in `serve'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/journey/router.rb:50:in `block in serve'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/journey/router.rb:32:in `each'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/journey/router.rb:32:in `serve'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/routing/route_set.rb:842:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/railties-6.1.7.7/lib/rails/engine.rb:539:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/railties-6.1.7.7/lib/rails/railtie.rb:207:in `public_send'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/railties-6.1.7.7/lib/rails/railtie.rb:207:in `method_missing'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/routing/mapper.rb:20:in `block in <class:Constraints>'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/routing/mapper.rb:49:in `serve'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/journey/router.rb:50:in `block in serve'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/journey/router.rb:32:in `each'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/journey/router.rb:32:in `serve'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/routing/route_set.rb:842:in `call'
 2a1f0b55 | /home/vagrant/katello/lib/katello/middleware/organization_created_enforcer.rb:18:in `call'
 2a1f0b55 | /home/vagrant/katello/lib/katello/middleware/event_daemon.rb:10:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/middleware/static.rb:24:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/middleware/static.rb:24:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/apipie-dsl-2.6.2/lib/apipie_dsl/static_dispatcher.rb:67:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/apipie-rails-1.3.0/lib/apipie/static_dispatcher.rb:74:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/bullet-7.1.6/lib/bullet/rack.rb:17:in `call'
 2a1f0b55 | /home/vagrant/foreman/lib/foreman/middleware/libvirt_connection_cleaner.rb:9:in `call'
 2a1f0b55 | /home/vagrant/foreman/lib/foreman/middleware/telemetry.rb:10:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/apipie-rails-1.3.0/lib/apipie/middleware/checksum_in_headers.rb:27:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/rack-openid-1.4.2/lib/rack/openid.rb:98:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/rack-2.2.9/lib/rack/tempfile_reaper.rb:15:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/rack-2.2.9/lib/rack/etag.rb:27:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/rack-2.2.9/lib/rack/conditional_get.rb:40:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/rack-2.2.9/lib/rack/head.rb:12:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/http/permissions_policy.rb:22:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/http/content_security_policy.rb:19:in `call'
 2a1f0b55 | /home/vagrant/foreman/lib/foreman/middleware/logging_context_session.rb:22:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/rack-2.2.9/lib/rack/session/abstract/id.rb:266:in `context'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/rack-2.2.9/lib/rack/session/abstract/id.rb:260:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/middleware/cookies.rb:697:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/activerecord-6.1.7.7/lib/active_record/migration.rb:601:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/middleware/callbacks.rb:27:in `block in call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/activesupport-6.1.7.7/lib/active_support/callbacks.rb:98:in `run_callbacks'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/middleware/callbacks.rb:26:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/middleware/executor.rb:14:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/middleware/actionable_exceptions.rb:18:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/middleware/debug_exceptions.rb:29:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/middleware/show_exceptions.rb:33:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/railties-6.1.7.7/lib/rails/rack/logger.rb:37:in `call_app'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/railties-6.1.7.7/lib/rails/rack/logger.rb:28:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/sprockets-rails-3.4.2/lib/sprockets/rails/quiet_assets.rb:13:in `call'
 2a1f0b55 | /home/vagrant/foreman/lib/foreman/middleware/logging_context_request.rb:11:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/middleware/remote_ip.rb:81:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/middleware/request_id.rb:26:in `call'
 2a1f0b55 | /home/vagrant/katello/lib/katello/prevent_json_parsing.rb:12:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/rack-2.2.9/lib/rack/method_override.rb:24:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/rack-2.2.9/lib/rack/runtime.rb:22:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/activesupport-6.1.7.7/lib/active_support/cache/strategy/local_cache_middleware.rb:29:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/middleware/executor.rb:14:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/middleware/static.rb:24:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/rack-2.2.9/lib/rack/sendfile.rb:110:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/actionpack-6.1.7.7/lib/action_dispatch/middleware/host_authorization.rb:148:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/secure_headers-6.5.0/lib/secure_headers/middleware.rb:11:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/railties-6.1.7.7/lib/rails/engine.rb:539:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/railties-6.1.7.7/lib/rails/railtie.rb:207:in `public_send'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/railties-6.1.7.7/lib/rails/railtie.rb:207:in `method_missing'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/rack-2.2.9/lib/rack/urlmap.rb:74:in `block in call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/rack-2.2.9/lib/rack/urlmap.rb:58:in `each'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/rack-2.2.9/lib/rack/urlmap.rb:58:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/puma-6.4.2/lib/puma/configuration.rb:272:in `call'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/puma-6.4.2/lib/puma/request.rb:100:in `block in handle_request'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/puma-6.4.2/lib/puma/thread_pool.rb:378:in `with_force_shutdown'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/puma-6.4.2/lib/puma/request.rb:99:in `handle_request'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/puma-6.4.2/lib/puma/server.rb:464:in `process_client'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/puma-6.4.2/lib/puma/server.rb:245:in `block in run'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/puma-6.4.2/lib/puma/thread_pool.rb:155:in `block in spawn_thread'
 2a1f0b55 | /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/logging-2.3.1/lib/logging/diagnostic_context.rb:474:in `block in create_with_logging_context'
2024-05-14T21:18:25 [I|app|2a1f0b55] Completed 500 Internal Server Error in 221ms (Views: 0.2ms | ActiveRecord: 6.6ms | Allocations: 16103)

```

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -225,6 +325,21 @@ def start_upload_blob
       pulp_response.headers.each do |key, value|
         response.header[key.to_s] = value
       end
+      
+      # if pulp was successful and we need to create a new repository, create it now
+      if 200 <= pulp_response.code && pulp_response.code <= 299 && 
+        @product.root_repositories.where(label:@container_name).empty?
```
ianballou, line 331:
```suggestion
        @product.root_repositories.where(label: @container_name).empty?
```

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -268,6 +383,27 @@ def finish_upload_blob
...
+      headers['Content-Type'] = request.headers['Content-Type'] if request.headers['Content-Type']
+      body = @_request.body.read
+      pulp_response = Resources::Registry::Proxy.put(@_request.fullpath, body, headers)
+      pulp_response.headers.each do |key, value|
+        response.header[key.to_s] = value
+      end
+
+      # if manifest push was successful, trigger content indexing on the repo
+      if 200 <= pulp_response.code && pulp_response.code <= 299
+        unless @product.root_repositories.where(label: @container_name).first.library_instance.index_content
```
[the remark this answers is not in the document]
qcjames53: So you were pushing an image that was manifest only so it only hit the push_manifest controller method? I suppose it would make sense to make the repo creation logic on all four podman-related API endpoints, probably as a before action.

app/models/katello/repository.rb
ianballou: Following the LE pattern means the repository name in Katello and Pulp will differ. That means that users will end up pushing and pulling with different names, which would be different from normal container workflows.

For push, we should be able to completely skip the container repository name setting. One way to tell if a repository is a special push one could be via the version_href, since the version_href has that container-push string in it. Or, you could consider adding a tiny field in the database at creation time to denote that it's a container push repository.

I think it would be helpful to have a way to tell if a repo is a push one regardless -- we might need to lock down things in the UI, for example, which would mean the API would need to return if a repository is a container push one.

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -225,6 +325,21 @@ def start_upload_blob
...
+        @product.root_repositories.where(label:@container_name).empty?
+        
+        root = @product.add_repo(
+          name: @container_name,
+          label: @container_name,
+          download_policy: 'immediate',
+          content_type: Repository::DOCKER_TYPE,
+          unprotected: true
+        )
+        sync_task(::Actions::Katello::Repository::CreateRoot, root, @container_path_input)
```
ianballou, line 340: One issue with using CreateRoot -- it creates entities that are unnecessary like a Pulp repository and a distribution for that repository.  I think this might be breaking you code, since the version_href after the Katello repository gets created incorrectly points to an empty Pulp repository.

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -85,6 +86,116 @@ def registry_authorize
...
+      end      
+    end
+
+    def check_blob_push_props(props=nil)
+      props = get_blob_push_props if props.nil?
+
+      # check basic url field syntax
+      unless props[:valid_format]
+        render_error('custom_error', :status => :unprocessable_entity,
+          :locals => { :message => "Invalid format. Container pushes should follow '/organization_label/product_label/name' OR '/organization_id/product_id/name' schema." })
```
ianballou, line 130: Try to be consistent with rocket syntax and default to colon syntax, I think colon syntax is now the norm in Ruby. Rocket is necessary sometimes still, like when using strings instead of labels.

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -85,6 +86,116 @@ def registry_authorize
...
+          organization: segments[3],
+          product: segments[4],
+          name: segments[5]
+        }
+      else
+        return {valid_format: false}
+      end      
+    end
+
+    def check_blob_push_props(props=nil)
```
ianballou, line 124: Refactoring this down into smaller methods would help a lot with readability. Plus I'm guessing Rubocop is complaining :)

app/controllers/katello/api/registry/registry_proxies_controller.rb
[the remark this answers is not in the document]
ianballou: yeah it seems only the manifest was getting pushed since all the blobs already existed in the Pulp repo

ianballou:
I realize that we will definitely need some way to tell push repositories apart from normal ones. Relying on the version_href to say "container-push" in it could be a pretty good way to do so. When making content views, we'll need it to determine how contents should be copied. We'll also need it for deleting container push repositories, since it's the distribution that needs to be deleted rather than the repo itself.

app/controllers/katello/api/registry/registry_proxies_controller.rb
qcjames53: Done, thank you!

qcjames53: Done, thanks! Was on the to-do list.

app/controllers/katello/api/registry/registry_proxies_controller.rb
qcjames53: Done, thanks!

app/controllers/katello/api/registry/registry_proxies_controller.rb
qcjames53: Fixed, thanks

ianballou:
> Containers with tags such as `default_organization/dummy_product/foo:bar` end up losing the tag when pushed. I think this is okay since the tag is on the podman side; however, this means that users would technically need to pull using a different name than the name they pushed with. This may need further discussion since I could see users expecting either behavior (tag loss vs tag specification on pull).

I missed this comment before. The tag is separate from the repository name -- it is tagging the manifest that the user is pushing up. Under the hood, if not tag is noted, podman will auto-assign latest to the manifest being pushed. I'm guessing in this case then that all pushed content is getting the latest tag. It's important that tags are taken in exactly as they're pushed.

webpack/scenes/Hosts/ChangeContentSource/actions.js
```diff
@@ -47,7 +47,7 @@ export const getContentViews = environmentId =>
     params: {
       environment_id: environmentId,
       full_result: true,
-      order: 'default DESC', // shows the default CV before all other options
```
ianballou, line 50: Extra commit?

qcjames53: Yes it's from 37229. No idea why it was added.

ianballou:
I tested passing in custom tags and it works :)

```
vagrant@centos9-katello-devel ~/foreman $ podman push 66640e9653de  `hostname`/default_organization/buttermilk_biscuits/busybox:tacos
Getting image source signatures
Copying blob 29e363cff31e skipped: already exists  
Copying config 66640e9653 done   | 
Writing manifest to image destination
vagrant@centos9-katello-devel ~/foreman $ sudo pulp show --href /pulp/api/v3/content/container/tags/?repository_version=/pulp/api/v3/repositories/container/container-push/018f9759-7d55-7b1e-82b0-f2f062ebd606/versions/1/
{
  "count": 1,
  "next": null,
  "previous": null,
  "results": [
    {
      "pulp_href": "/pulp/api/v3/content/container/tags/018f9759-8181-7602-b74e-db7c22a9981b/",
      "pulp_created": "2024-05-20T18:52:56.577888Z",
      "pulp_last_updated": "2024-05-20T18:52:56.577901Z",
      "name": "latest",
      "tagged_manifest": "/pulp/api/v3/content/container/manifests/018f9759-816b-7ed1-a0a9-6f9b8aabfbbc/"
    }
  ]
}

vagrant@centos9-katello-devel ~/foreman $ sudo pulp show --href /pulp/api/v3/content/container/tags/?repository_version=/pulp/api/v3/repositories/container/container-push/018f9759-7d55-7b1e-82b0-f2f062ebd606/versions/2/
{
  "count": 2,
  "next": null,
  "previous": null,
  "results": [
    {
      "pulp_href": "/pulp/api/v3/content/container/tags/018f975c-858c-774c-9271-e67a92edcf5e/",
      "pulp_created": "2024-05-20T18:56:14.221447Z",
      "pulp_last_updated": "2024-05-20T18:56:14.221459Z",
      "name": "tacos",
      "tagged_manifest": "/pulp/api/v3/content/container/manifests/018f975c-857f-7709-9ab5-d8e897713f8e/"
    },
    {
      "pulp_href": "/pulp/api/v3/content/container/tags/018f9759-8181-7602-b74e-db7c22a9981b/",
      "pulp_created": "2024-05-20T18:52:56.577888Z",
      "pulp_last_updated": "2024-05-20T18:52:56.577901Z",
      "name": "latest",
      "tagged_manifest": "/pulp/api/v3/content/container/manifests/018f9759-816b-7ed1-a0a9-6f9b8aabfbbc/"
    }
  ]
}

```

ianballou (commented):
Weird errors aside, pushing via name and via IDs are both creating repositories in the proper places :+1:

qcjames53:
> > Containers with tags such as `default_organization/dummy_product/foo:bar` end up losing the tag when pushed. I think this is okay since the tag is on the podman side; however, this means that users would technically need to pull using a different name than the name they pushed with. This may need further discussion since I could see users expecting either behavior (tag loss vs tag specification on pull).
> 
> I missed this comment before. The tag is separate from the repository name -- it is tagging the manifest that the user is pushing up. Under the hood, if not tag is noted, podman will auto-assign latest to the manifest being pushed. I'm guessing in this case then that all pushed content is getting the latest tag. It's important that tags are taken in exactly as they're pushed.

ianballou and I talked about this outside of the PR and we came to the conclusion that we are handling tags correctly at the moment. Tag info shouldn't need to be stored anywhere in the repo; handling everything with Pulp should be completely fine.

app/controllers/katello/api/registry/registry_proxies_controller.rb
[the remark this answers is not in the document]
qcjames53: This issue has been resolved. Repos are now created whenever they do not already exist.

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -85,6 +84,148 @@ def registry_authorize
...
+      if props[:schema] == "label"
+        prod_label = props[:product]
+        unless prod_label.present? && prod_label.length > 0
+          render_error('custom_error', status: unprocessable_entity,
+            locals: {message: "Invalid format. Product label cannot be blank." })
+          return false
+        end
+        product = @organization.products.where("LOWER(label) = '#{prod_label}'") # convert to lowercase
+        if product.length > 1 # reject ambiguous products (possible due to lowercase conversion)
+          render_error('custom_error', status: unprocessable_entity,
```
ianballou, line 186: Like below, the error here reaches the user as a 500 and the message is not carried along. If you start using the OCI error spec, you'll be able to get your error message across.

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -85,6 +84,148 @@ def registry_authorize
...
+      if @product.root_repositories.where(label: @container_name).empty?
+        root = @product.add_repo(
+          name: @container_name,
+          label: @container_name,
+          download_policy: 'immediate',
+          content_type: Repository::DOCKER_TYPE,
+          unprotected: true,
+          is_container_push: true
+        )
+        sync_task(::Actions::Katello::Repository::CreateRoot, root, @container_path_input)
```
ianballou, line 225: `::Actions::Katello::Repository::Create` can now be modified to skip the `Pulp3::Orchestration::Repository::Create` and `Katello::Repository::MetadataGenerate` steps if the repository is a push one.

Currently, an empty repo is still being created in Pulp with its href saved to the `::Katello::Repository`.

ianballou (changes requested):

app/controllers/katello/api/registry/registry_proxies_controller.rb
qcjames53: This should now be taken care of. Thank you for the help!

app/controllers/katello/api/registry/registry_proxies_controller.rb
qcjames53: Error handling is now using OCI spec, thank you.

app/models/katello/repository.rb
qcjames53: `version_href` has been updated to match the pulp contents AND there is a `is_container_push` flag on the root repo. Should be all good to go.

app/controllers/katello/api/registry/registry_proxies_controller.rb
qcjames53: This is now working properly after updating Repository::Create. Thank you for the help!

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -85,6 +84,148 @@ def registry_authorize
...
+            locals: {message: "Invalid format. Product label cannot be blank." })
+          return false
+        end
+        product = @organization.products.where("LOWER(label) = '#{prod_label}'") # convert to lowercase
+        if product.length > 1 # reject ambiguous products (possible due to lowercase conversion)
+          render_error('custom_error', status: unprocessable_entity,
+            locals: {message: "Product label is ambiguous" })
+          return false
+        elsif product.length == 0
+          render_error('custom_error', status: unprocessable_entity,
```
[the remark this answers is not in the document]
qcjames53: Errors now follow OCI spec, thank you.

qcjames53:
@ianballou I really appreciate all the help with content indexing above (too big to quote). The newest update to the PR should fix the duplicate Pulp entry, set the root repo's `version_href` properly, and handle errors so that Podman can display them.

ianballou:
Already can see one improvement:
```
vagrant@centos9-katello-devel ~/katello $ podman push 66640e9653de  `hostname`/default_organization/buttermilk_biscuits/busybox:latest
Getting image source signatures
Copying blob b7d21803af9e [--------------------------------------] 8.0b / 3.0MiB | 1.3 MiB/s
WARN[0001] Failed, retrying in 1s ... (1/3). Error: writing blob: initiating layer upload to /v2/default_organization/buttermilk_biscuits/busybox/blobs/uploads/ in centos9-katello-devel.manicotto.example.com: name invalid: Product label 'buttermilk_biscuits' is ambiguous. Try using an id-based container name. 
Getting image source signatures
Copying blob b7d21803af9e [--------------------------------------] 8.0b / 3.0MiB | 1.8 MiB/s
WARN[0004] Failed, retrying in 1s ... (2/3). Error: writing blob: initiating layer upload to /v2/default_organization/buttermilk_biscuits/busybox/blobs/uploads/ in centos9-katello-devel.manicotto.example.com: name invalid: Product label 'buttermilk_biscuits' is ambiguous. Try using an id-based container name. 
Getting image source signatures
Copying blob b7d21803af9e [--------------------------------------] 8.0b / 3.0MiB | 2.0 MiB/s
WARN[0007] Failed, retrying in 1s ... (3/3). Error: writing blob: initiating layer upload to /v2/default_organization/buttermilk_biscuits/busybox/blobs/uploads/ in centos9-katello-devel.manicotto.example.com: name invalid: Product label 'buttermilk_biscuits' is ambiguous. Try using an id-based container name. 
Getting image source signatures
Copying blob b7d21803af9e [--------------------------------------] 8.0b / 3.0MiB | 1.7 MiB/s
Error: writing blob: initiating layer upload to /v2/default_organization/buttermilk_biscuits/busybox/blobs/uploads/ in centos9-katello-devel.manicotto.example.com: name invalid: Product label 'buttermilk_biscuits' is ambiguous. Try using an id-based container name.

```

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -85,6 +85,188 @@ def registry_authorize
...
+        instance_repo = root_repo&.library_instance
+
+        return render_podman_error(
+          "BLOB_UPLOAD_UNKNOWN", 
+          "Could not locate local uploaded repository for content indexing.",
+          :not_found
+        ) unless root_repo.present? && instance_repo.present?
+
+        api = ::Katello::Pulp3::Repository.api(SmartProxy.pulp_primary, ::Katello::Repository::DOCKER_TYPE).container_push_api
+        latest_version_href = api.list(name: @container_path_input).results.first.latest_version_href
```
ianballou, line 257: `api.list(name: @container_path_input).results.first` was nil for me when pushing via the ID scheme for some reason.

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -85,6 +85,188 @@ def registry_authorize
...
+
+        return render_podman_error(
+          "BLOB_UPLOAD_UNKNOWN", 
+          "Could not locate repository properties for content indexing.",
+          :not_found
+        ) unless latest_version_href.present?
+
+        instance_repo.index_content
+        instance_repo.update!(version_href: latest_version_href)
```
ianballou, line 266: You need to switch these two lines, then indexing content should work.

```suggestion
        instance_repo.update!(version_href: latest_version_href)
        instance_repo.index_content
```

app/controllers/katello/api/registry/registry_proxies_controller.rb
qcjames53: That is interesting. @container_path_input _should_ be exactly what the user entered. I'll look into it.

app/controllers/katello/api/registry/registry_proxies_controller.rb
qcjames53: Just fixed this, thanks

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -85,6 +84,223 @@ def registry_authorize
...
+          unprotected: true,
+          is_container_push: true
+        )
+        sync_task(::Actions::Katello::Repository::CreateRoot, root, @container_path_input)
+      end
+    end
+
+    def handle_blob_push_cleanup
+      # if we're returning a successful code, index content and set version href using pulp api
+      if response.status.between(200, 299)
```
ianballou, line 276: `NoMethodError: undefined method `between' for 201:Integer`

```suggestion
      if response.status.between?(200, 299)
```

ianballou (changes requested):

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -85,6 +84,223 @@ def registry_authorize
...
+        }
+      else
+        return {valid_format: false}
+      end
+    end
+
+    def check_blob_push_field_syntax(props)
+      # check basic url field syntax
+      unless props[:valid_format]
+        return render_podman_error(
```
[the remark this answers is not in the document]
ianballou: The return might not be the issue, but it's one possibility.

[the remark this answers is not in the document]
ianballou: Ah, it says "Also note that neither redirect nor render terminate execution of the action"

So perhaps then there is a return missing that is causing a double render?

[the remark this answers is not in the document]
ianballou: It might have to do with how you're trying to render stuff in an after_action.

In the action, there should already be a render or a redirect, so there's no more API response to send inside after_action.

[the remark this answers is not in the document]
ianballou: ""after" filters are registered via [after_action](https://api.rubyonrails.org/v7.1.3.3/classes/AbstractController/Callbacks/ClassMethods.html#method-i-after_action). They are similar to "before" filters, but because the action has already been run they have access to the response data that's about to be sent to the client. Obviously, "after" filters cannot stop the action from running. Please note that "after" filters are executed only after a successful action, but not when an exception is raised in the request cycle."

https://guides.rubyonrails.org/action_controller_overview.html#after-filters-and-around-filters

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -85,6 +84,223 @@ def registry_authorize
...
+            "BLOB_UPLOAD_UNKNOWN",
+            "Could not locate local uploaded repository for content indexing.",
+            :not_found
+          )
+        end
+
+        api = ::Katello::Pulp3::Repository.api(SmartProxy.pulp_primary, ::Katello::Repository::DOCKER_TYPE).container_push_api
+        latest_version_href = api.list(name: @container_path_input).results&.first&.latest_version_href
+
+        unless latest_version_href.blank?
```
ianballou, line 291: Should this be:

```suggestion
        if latest_version_href.blank?
```

Also one note: I like `empty?` because it covers nil, empty strings, and empty arrays.

ianballou (changes requested):

ianballou:
One thing I noticed is that errors are not easily recovered from -- in one case my push succeeded, but the repository remained and I had to do some cleanup before trying to push again. We could potentially deal with it in a later PR, depends on if it ends up being a bigger issue.

ianballou:
I'm seeing some issues with bigger container images where the repository is created twice. I can reproduce the issue solidly with `quay.io/curl/curl:latest

app/lib/actions/katello/repository/create.rb
```diff
@@ -50,7 +53,10 @@ def run
           ::User.current = ::User.anonymous_api_admin
           unless input[:clone]
             repository = ::Katello::Repository.find(input[:repository_id])
-            ForemanTasks.async_task(Katello::Repository::MetadataGenerate, repository)
+
+            # Container push repos do not need metadata generation
+            ForemanTasks.async_task(Katello::Repository::MetadataGenerate, repository) unless
+              repository.root.is_container_push
```
ianballou, line 59: Rather than skipping just the async_task call for MetadataGenerate, you could actually skip this entire section up on line 38: 
```
            concurrence do
              plan_self(:repository_id => repository.id, :clone => clone)
              if !clone && repository.url.present?
                repository.product.alternate_content_sources.with_type(repository.content_type).each do |acs|
                  acs.smart_proxies.each do |smart_proxy|
                    smart_proxy_acs = ::Katello::SmartProxyAlternateContentSource.create(alternate_content_source_id: acs.id, smart_proxy_id: smart_proxy.id, repository_id: repository.id)
                    plan_action(Pulp3::Orchestration::AlternateContentSource::Create, smart_proxy_acs)
                  end
                end
              end
```

This section runs `plan_self`, which plans the run phase of the action (what's under `def run`).  The run phase only triggers the MetadataGenerate call. By skipping the entire run phase, overhead from creating a job for a Dynflow worker will be completely removed.

You migh task if the other stuff relating to alternate content sources is important. Alternate content sources only help with syncing, and push repositories will never sync anything, so there should be no reason to even create an ACS for them. We'll save some more cycles by skipping that as well.

I don't see anything else in this action that can be skipped for container push repos.

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -85,6 +84,223 @@ def registry_authorize
...
+          return render_podman_error(
+            "BLOB_UPLOAD_UNKNOWN",
+            "Could not locate repository properties for content indexing.",
+            :not_found
+          )
+        end
+
+        instance_repo.update!(version_href: latest_version_href)
```
ianballou, line 299:
```suggestion
        instance_repo.update!(version_href: latest_version_href)
        RepositoryReference.where(
            root_repository_id: instance_repo.root_id,
            content_view_id: instance_repo.content_view.id,
            repository_href: api_response.pulp_href).create!
```

We need to create the repository reference to set us up for content view publishing down the line.

ianballou (changes requested):

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -85,6 +84,223 @@ def registry_authorize
...
+      if @product.root_repositories.where(label: @container_name).empty?
+        root = @product.add_repo(
+          name: @container_name,
+          label: @container_name,
+          download_policy: 'immediate',
+          content_type: Repository::DOCKER_TYPE,
+          unprotected: true,
+          is_container_push: true
+        )
+        sync_task(::Actions::Katello::Repository::CreateRoot, root, @container_path_input)
```
ianballou, line 270: I found an interesting issue here -- there seems to be a race condition.

Pushing content is concurrent. That means that multiple threads could be in this if statement at once time. So, `::Actions::Katello::Repository::CreateRoot` is being run even when `@product.root_repositories.where(label: @container_name).empty?` turns false.

You can reproduce this in the console by sticking a debugger right before the sync_task call. Multiple different debuggers will run at once due to the concurrency, and they'll all queue up before the sync_task call.

This is tricky because we need a way for one of the concurrent calls from podman to become the "chosen one" and for the rest of the calls to wait. One idea that comes to mind would be to implement some sort of lock that is saved in the database.

ianballou: I confirmed that using a single-threaded development environment "solves" the issue, so concurrency does seem to be the problem.

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -85,6 +84,223 @@ def registry_authorize
       return false
     end
 
+    def handle_blob_push(props = nil)
```
ianballou, line 87: It's doing more than handling blobs -- maybe change the name to something like `run_pre_push_checks` ?

app/controllers/katello/api/registry/registry_proxies_controller.rb
qcjames53: Fixed, thanks

app/controllers/katello/api/registry/registry_proxies_controller.rb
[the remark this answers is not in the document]
qcjames53: Fixed, thank you. After discussion, we determined that content indexing only needs to run after manifest upload. after_action has been removed and this method is called from the manifest push.

app/lib/actions/katello/repository/create.rb
qcjames53: Thank you for finding this :) 
Your suggested changes have been implemented.

app/controllers/katello/api/registry/registry_proxies_controller.rb
qcjames53: `if latest_version_href.empty?` thank you!

app/controllers/katello/api/registry/registry_proxies_controller.rb
qcjames53: Added as suggested, thank you.

app/controllers/katello/api/registry/registry_proxies_controller.rb
qcjames53: I ended up splitting this into two before_actions: `blob_push_validation` and `create_container_repo_if_needed`. I think this will also make the lock implementation cleaner. Does that sound ok to you?

qcjames53: I'm looking into if I can move the logic for testing if a container is needed into the `before_action` line

app/controllers/katello/api/registry/registry_proxies_controller.rb
ianballou: One idea would be to use https://apidock.com/rails/v6.0.0/ActiveRecord/Relation/create_or_find_by

(or maybe `create_or_find_by!`)

I wonder if it would be easier to stop using the CreateRoot action and instead create the database entries manually, especially since we already have to add container push conditionals to those actions.

ianballou: Just for inspiration, https://api.rubyonrails.org/v7.1.3.2/classes/ActiveRecord/Relation.html#method-i-find_or_initialize_by is also available -- it would allow to instantiate the RootRepository object OR actually get a record if it exists.

app/controllers/katello/api/registry/registry_proxies_controller.rb
ianballou: My point was more that it's also handling pre-checks for manifests, so having "blob" in the name was a little misleading.

qcjames53: How does `container_push_prop_validation` sound?

app/controllers/katello/api/registry/registry_proxies_controller.rb
qcjames53: > One idea would be to use https://apidock.com/rails/v6.0.0/ActiveRecord/Relation/create_or_find_by
> 
> (or maybe `create_or_find_by!`)
> 
> I wonder if it would be easier to stop using the CreateRoot action and instead create the database entries manually. We'd then be able to remove the conditionals from the current actions.

I've been experimenting with that for testing different approaches. I'm going to try and roll everything back into the foreman tasks unless that breaks concurrency, just so there isn't two different methods in Katello that do the same thing.

qcjames53: Also thank you for the resources. Have been trying to get create_or_find_by to work but there are a lot of layers of foreman tasks to unroll.

qcjames53: As of the newest push, I believe this is working correctly. I used `create_or_find_by` inside of `app/models/katello/glue/pulp/repos.rb`. When you have time, please confirm the issue is fixed on your end (I've been having trouble getting the endpoint to be hit simultaneously on my device).

ianballou (commented):
Finished up a functional test, was able to push the fedora-bootc image multiple times and the content all ended up in the same repository. It seems like the concurrency problem was solved, nice!

---

One issue I noticed -- pushing by IDs to a repository right after pushing by name created a different repository in Pulp. It didn't create another repository in Katello though, which is correct. As a result though, the content pushed by ID in Pulp appears to have disappeared in the Katello UI. 

I was going to suggest translating the IDs to names before pushing content, but that would defeat the whole purpose of being able to differentiate the container names in Pulp.

To fix this, I think we'd need to know if a repository was originally pushed by name in order to decide which repo to push content to -- the "name" repo, or the "ID" repo.

If a user later creates an org / product that makes the naming ambiguous, they'll need to be able to keep pushing content to their original repositories that they pushed to by name. To do so, they'll need to start pushing by ID.

app/models/katello/root_repository.rb
```diff
@@ -71,7 +71,7 @@ class RootRepository < Katello::Model
     validates_with Validators::ContainerImageNameValidator, :attributes => :docker_upstream_name, :allow_blank => true, :if => :docker?
 
     validate :ensure_valid_docker_attributes, :if => :docker?
-    validate :ensure_docker_repo_unprotected, :if => :docker?
+    #validate :ensure_docker_repo_unprotected, :if => :docker?
```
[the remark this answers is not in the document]
ianballou: I'm pretty sure we can completely remove the validation and the validation method.

ianballou:
I wonder if this is starting to get more complicated than we need, especially since this capitalization scenario we're guarding against should be very rare.

I'm trying to think if we could completely throw out the "by ID" style to ensure that Katello and Pulp are always aligned.

ianballou:
Throwing out the ID method entirely will force the rare ambiguous org / product owner to recreate the org / product. I think the safer middle ground would be to check if a user is trying to push content to a repository with a different name / ID scheme and to tell them to destroy it first. It makes for a decent experience for the very rare ambiguous org/product user, and shouldn't over-complicate the code for the 99% use-case.

ianballou (changes requested):
The error messages are looking better! One issue I noticed:

1) Push an image by name/label
2) Create another org with the same name
3) Create a product in the new org with the same name as the product in the old org
4) Push the same image in step (1) to the product in step (3) by ID
5) The ID repo gets created in Pulp, but no repository is created in the new org/product.

It looks like the Root repository was created under the new org. They do have the proper product association as well. However, neither of them have any Repository records.

The erroneous Root repositories also don't have any RepositoryReferences.

The first repository pushed by name is all correct, it seems.

Edit: I was able to push a new repo via ID to the new org, but it doesn't seem to have the version_href set up.

I think in my Pulp the arianna repository already existed from a prior push.
  -> It looks like this didn't matter.

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -85,6 +84,286 @@ def registry_authorize
...
+
+      if latest_version_href.empty? || pulp_href.empty?
+        return render_podman_error(
+          "BLOB_UPLOAD_UNKNOWN",
+          "Could not locate repository properties for content indexing.",
+          :not_found
+        )
+      end
+
+      instance_repo.update!(version_href: latest_version_href)
```
ianballou, line 361: It looks like this (and likely the entire `blob_push_cleanup`) is not running for by-ID pushes.

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -85,6 +84,286 @@ def registry_authorize
...
+      unless root_repo.present? && instance_repo.present?
+        return render_podman_error(
+          "BLOB_UPLOAD_UNKNOWN",
+          "Could not locate local uploaded repository for content indexing.",
+          :not_found
+        )
+      end
+
+      api = ::Katello::Pulp3::Repository.api(SmartProxy.pulp_primary, ::Katello::Repository::DOCKER_TYPE).container_push_api
+      api_response = api.list(name: @container_path_input).results&.first
```
ianballou, line 349: For ID-type repositories, this always returns empty because @container_path_input does not include "id":

```
[18] pry(#<Katello::Api::Registry::RegistryProxiesController>)> @container_path_input
=> "6/344/arianna"

------------------------------------

[19] pry(#<Katello::Api::Registry::RegistryProxiesController>)> api.list(name: @container_path_input)
2024-06-03T19:05:41 [D|kat|334ab382] Calling API: RepositoriesContainerPushApi.list ...
2024-06-03T19:05:41 [D|kat|334ab382] HTTP response body ~BEGIN~
 334ab382 | {"count":0,"next":null,"previous":null,"results":[]}
 334ab382 | ~END~
 334ab382 | 
2024-06-03T19:05:41 [D|kat|334ab382] API called: RepositoriesContainerPushApi#list
 334ab382 | Data: #<PulpContainerClient::PaginatedcontainerContainerPushRepositoryResponseList:0x00007fba5ca9f358 @count=0, @results=[]>
 334ab382 | Status code: 200
 334ab382 | Headers: {"date"=>"Mon, 03 Jun 2024 19:05:41 GMT", "server"=>"gunicorn", "content-type"=>"application/json", "vary"=>"Accept,Cookie", "allow"=>"GET, HEAD, OPTIONS", "x-frame-options"=>"DENY", "content-length"=>"52", "x-content-type-options"=>"nosniff", "referrer-policy"=>"same-origin", "cross-origin-opener-policy"=>"same-origin", "correlation-id"=>"334ab382-a357-4489-98a1-e909c19b1d41", "access-control-expose-headers"=>"Correlation-ID", "via"=>"1.1 centos9-katello-devel.manicotto.example.com"}
=> #<PulpContainerClient::PaginatedcontainerContainerPushRepositoryResponseList:0x00007fba5ca9f358 @count=0, @results=[]>

-----------------------------------------

  {
    "pulp_created": "2024-06-03T19:01:47.382749Z",
    "latest_version_href": "/pulp/api/v3/repositories/container/container-push/018fdf7a-a2f5-7f56-bb36-2e9b66fd0f19/versions/1/",
    "pulp_labels": {},
    "retain_repo_versions": null,
    "pulp_href": "/pulp/api/v3/repositories/container/container-push/018fdf7a-a2f5-7f56-bb36-2e9b66fd0f19/",
    "pulp_last_updated": "2024-06-03T19:01:48.179714Z",
    "versions_href": "/pulp/api/v3/repositories/container/container-push/018fdf7a-a2f5-7f56-bb36-2e9b66fd0f19/versions/",
    "name": "id/6/344/arianna",
    "manifest_signing_service": null,
    "description": null
  },

```

ianballou: I believe this is the cause for ID-type repositories not being created correctly.

qcjames53: This is fixed as of the next update to this PR, thank you

ianballou (commented):
Things have changed a bit -- now I'm seeing:

```
Error: writing manifest: uploading manifest latest to centos9-katello-devel.manicotto.example.com/id/3/2/curl: blob upload unknown: Could not locate local uploaded repository for content indexing.
```
when trying to push the same container repository (curl) to a second org with the same name by ID.

However, if I push a brand new container repository, it seems to work just fine now.

ianballou:
I think the commit and title can be changed to a "Fixes" instead of "Refs"

ianballou:
It looks like pulling the pushed content broke somewhere along the way:

```
vagrant@centos9-katello-devel ~/katello $ podman search `hostname`/
NAME                                                                                          DESCRIPTION
centos9-katello-devel.manicotto.example.com/default_organization-buttermilk_biscuits-curl     
centos9-katello-devel.manicotto.example.com/default_organization-buttermilk_biscuits-busybox
```

It seems Katello still thinks we're distributing the contents using the old container name format.

However, I'd be okay with pushing this to another PR, since in our plan we did have distribution as a separate task. Up to you. I know this PR has been lingering a while!

ianballou:
Pushing the same repo to multiple products seems to be failing, looks like the container repository name is being saved just as the name of the repository and not the full path with org and product:

```
2024-06-04T20:47:41 [D|app|4773c508] With body: {"displayMessage":"Validation failed: Container repository name for repository 'busybox' is not unique and cannot be created in 'Library'. Its Container Repository Name (default_organization-buttermilk_biscuits-busybox) conflicts with an existing repository.  Consider changing the Lifecycle Environment's Registry Name Pattern to something more specific.","errors":{"container_repository_name":["for repository 'busybox' is not unique and cannot be created in 'Library'. Its Container Repository Name (default_organization-buttermilk_biscuits-busybox) conflicts with an existing repository.  Consider changing the Lifecycle Environment's Registry Name Pattern to something more specific."]}}

```

ianballou:
I think I may have fixed all the issues with duplicate org / product, try this patch:

```
diff --git a/app/lib/actions/katello/repository/create_root.rb b/app/lib/actions/katello/repository/create_root.rb
index c3770f8fd9..89f7c07f82 100644
--- a/app/lib/actions/katello/repository/create_root.rb
+++ b/app/lib/actions/katello/repository/create_root.rb
@@ -8,6 +8,7 @@ module Actions
                                       :content_view_version => root.organization.library.default_content_view_version,
                                       :root => root)
 
+          repository.container_repository_name = relative_path
           repository.relative_path = relative_path || repository.custom_repo_path
           repository.save!
 
diff --git a/app/models/katello/repository.rb b/app/models/katello/repository.rb
index a893959ec6..2bae545206 100644
--- a/app/models/katello/repository.rb
+++ b/app/models/katello/repository.rb
@@ -118,7 +118,7 @@ module Katello
     end}
 
     before_validation :set_pulp_id
-    before_validation :set_container_repository_name, :if => :docker?
+    before_validation :set_container_repository_name, :unless => :skip_container_name?
 
     scope :has_url, -> { joins(:root).where.not("#{RootRepository.table_name}.url" => nil) }
     scope :on_demand, -> { joins(:root).where("#{RootRepository.table_name}.download_policy" => ::Katello::RootRepository::DOWNLOAD_ON_DEMAND) }
@@ -268,6 +268,10 @@ module Katello
       self.content_view_version.content_view
     end
 
+    def skip_container_name?
+      self.library_instance? && self.root.docker? && self.root.is_container_push
+    end
+
     def library_instance?
       self.content_view.default?
     end
```

ianballou:
The container repository name was getting duplicated and throwing errors.

app/controllers/katello/api/registry/registry_proxies_controller.rb
qcjames53: This is now fixed, closing :)

app/controllers/katello/api/registry/registry_proxies_controller.rb
qcjames53: This has been changed, closing :)

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -3,15 +3,14 @@ module Katello
...
-    before_action :confirm_push_settings, only: [:start_upload_blob, :upload_blob, :finish_upload_blob,
-                                                 :push_manifest]
     skip_before_action :authorize
     before_action :optional_authorize, only: [:token, :catalog]
     before_action :registry_authorize, except: [:token, :v1_search, :catalog]
     before_action :authorize_repository_read, only: [:pull_manifest, :tags_list]
     # TODO: authorize_repository_write commented out until Katello indexes Pulp container push repositories.
-    # before_action :authorize_repository_write, only: [:start_upload_blob, :upload_blob, :finish_upload_blob,
-    #                                                   :push_manifest]
+    # before_action :authorize_repository_write, only: [:start_upload_blob, :upload_blob, :finish_upload_blob, :push_manifest]
```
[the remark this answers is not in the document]
qcjames53: Like you said here, we're going to leave authorization for a second PR. I'll create the JIRA card soon.

ianballou:
Looks like just a single test failure: `Error: Failure: test_docker_repo_unprotected`

qcjames53:
I just updated the PR description and testing steps. I'll get right on the failing test, thanks.

qcjames53:
The failing test should be fixed. I ran into a few issues with content indexing after making the changes in the most recent push so you may need to reset the db.

ianballou (commented):
This is working well now! I've finished testing a bunch of scenarios, pushing and pulling content with different levels of ambiguity.

I noticed one error that could be improved:
```
vagrant@centos9-katello-devel ~/katello $ podman push 66640e9653de `hostname`/id/1/339/arianna-id:nachos
Getting image source signatures
Copying blob b7d21803af9e [--------------------------------------] 8.0b / 3.0MiB | 1.7 MiB/s
Error: writing blob: initiating layer upload to /v2/id/1/339/arianna-id/blobs/uploads/ in centos9-katello-devel.manicotto.example.com: StatusCode: 404, "{\"displayMessage\":\"Couldn't find Organization with...
```
which happens when pushing by-ID to an org that doesn't exist.

If it's obvious how to fix it,  great. If not, we can leave this for later. I'm guessing the JSON returned for the error is just malformed.

ianballou (commented):
Impressive testing! Besides my comment above everything is looking good. I did some more testing as well. On another note, smart proxy syncing seems to work as well!

Once you reply about the only issue I found above, I can do my FinalFinalV2.txt review.

ianballou (approved):
Nice job, this is looking really solid. ACK (pending test run) !

PR #10998, opened 2024-05-16
Fixes #37476 - update ak results in hostgroup

MariaAga:
selecting lifecycle env does not update the ak results in new host group, since useMemo doesnt update.
placeholder prop was wrong and should be placeholderText to show empty results

### DIFF
--- a/webpack/components/ActivationKeysSearch/index.js
+++ b/webpack/components/ActivationKeysSearch/index.js
@@ -20,27 +20,29 @@ import { foremanUrl } from 'foremanReact/common/helpers';
 import { STATUS } from 'foremanReact/constants';
 import { selectAPIStatus } from 'foremanReact/redux/API/APISelectors';
 
+const getSelectedEnvId = () => {
+  const selectElement = document.querySelector('#hostgroup_lifecycle_environment_id');
+  const selectedOption = selectElement.options[selectElement.selectedIndex];
+  let dataId = selectedOption?.getAttribute?.('data-id');
+  if (!dataId) {
+    dataId = selectElement.value;
+  }
+  return dataId;
+};
+const getSelectedContentViewId = () => {
+  const selectElement = document.querySelector('#hostgroup_content_view_id');
+  const selectedOption = selectElement.options[selectElement.selectedIndex];
+  let dataId = selectedOption?.getAttribute?.('data-id');
+  if (!dataId) {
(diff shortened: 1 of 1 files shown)


adamruzicka:
Can't really speak to what goes on in the code, but it seems to resolve the issue

webpack/components/ActivationKeysSearch/index.js
```diff
@@ -23,24 +23,26 @@ import { selectAPIStatus } from 'foremanReact/redux/API/APISelectors';
 const ActivationKeysSearch = () => {
   const ACTIVATION_KEYS = 'ACTIVATION_KEYS';
   const KT_AK_LABEL = 'kt_activation_keys';
-  const selectedEnvId = useMemo(() => {
+  const getSelectedEnvId = () => {
```
jeremylenz, line 26: This seems to be vanilla JS, can we move it outside the component so the function isn't recreated on each render? That may provide the same benefit that the useMemo was supposed to.

MariaAga: yeah, thats a better place for it

webpack/components/ActivationKeysSearch/index.js
```diff
@@ -20,27 +20,29 @@ import { foremanUrl } from 'foremanReact/common/helpers';
...
-    const selectElement = document.querySelector('#hostgroup_content_view_id');
-    const selectedOption = selectElement.options[selectElement.selectedIndex];
-    let dataId = selectedOption?.getAttribute?.('data-id');
-    if (!dataId) {
-      dataId = selectElement.value;
-    }
-    return dataId;
-  }, []);
+  const [selectedEnvId, setSelectedEnvId] = useState(getSelectedEnvId());
+  const [selectedContentViewId, setSelectedContentViewId] = useState(getSelectedContentViewId);
```
jeremylenz, line 45:
```suggestion
  const [selectedContentViewId, setSelectedContentViewId] = useState(getSelectedContentViewId());
```

jeremylenz (commented):
Works well, just one more comment below

webpack/components/ActivationKeysSearch/index.js
MariaAga: Fixed, thanks

jeremylenz (approved):
Thanks @MariaAga! ACK

PR #10999, opened 2024-05-16
Fixes #37481 - Update cached value for manifest expiration date during import

jeremylenz:
Our stored manifest expiration date relies on a Rails cached value which is held for 1 minute:

```rb
def manifest_expiration_date(cached: true)
          Rails.cache.fetch("#{self.label}_manifest_expiration_date", expires_in: 1.minute, force: !cached) do
```

This means that after a manifest import / refresh, the manifest expiration date may be inaccurate until the cache expiration time (1 minute). Because of this you must refresh the browser page to see the new manifest expiration date.

1. During manifest import (and as a consequence, during manifest refresh), I found that calling `manifest_expiration_date(cached: false)` is enough to prevent stale data from making it to the user's screen.
2. If the manifest expiration date is not populated for some reason, don't display anything.

Delete and reimport a manifest. Visit the Manage Manifest modal.

Before:
You'll see a manifest expiration date of the Unix epoch (31 Dec 1969 or 01 Jan 1970)

After:
You'll see the correct manifest expiration date

### DIFF
--- a/app/lib/actions/katello/organization/manifest_import.rb
+++ b/app/lib/actions/katello/organization/manifest_import.rb
@@ -27,6 +27,11 @@ module Actions
           end
         end
 
+        def run
+          organization = ::Organization.find_by(name: input[:organization_name])
+          organization&.manifest_expiration_date(cached: false) # update the date
+        end
+
         def failure_notification(plan)
           ::Katello::UINotifications::Subscriptions::ManifestImportError.deliver!(
             :subject => subject_organization,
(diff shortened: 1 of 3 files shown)


jturel (approved):
Nice work @jeremylenz - thanks for the PR! :)

jeremylenz:
Updated to better handle importing of already-expired manifests.

jturel:
> > Nice work @jeremylenz - thanks for the PR! :)
> 
> We miss you man ❤️

Likewise!

chris1984 (approved):
Works great, thanks @jeremylenz

PR #11006, opened 2024-05-23
Fixes #37504 - send 404 to v2 client requesting v1/search

ianballou:
Podman no longer has libpod in the HTTP_USER_AGENT header. Now, we should check for v2 clients instead. This should apply to docker as well.

/_catalog and /v1/search are both removed from the OCI spec, but we should keep supporting them for now since the clients allow it.

1) Test podman search with unauthenticated pull off (default) and some search text:
```
$ podman search `hostname`/ubuntu

NAME                                                                                                                     DESCRIPTION
centos9-katello-devel.manicotto.example.com/default_organization-buttermilk_biscuits-docker_ubuntu                       
centos9-katello-devel.manicotto.example.com/default_organization-library-export_view-buttermilk_biscuits-docker_ubuntu   
centos9-katello-devel.manicotto.example.com/default_organization-buttermilk_biscuits-ubuntu                              
centos9-katello-devel.manicotto.example.com/default_organization-library-container_push_view-buttermilk_biscuits-ubuntu 
```

2) Try the code as well with a Docker client, just to be safe. Installed Docker-CE via their online instructions.

### DIFF
--- a/app/controllers/katello/api/registry/registry_proxies_controller.rb
+++ b/app/controllers/katello/api/registry/registry_proxies_controller.rb
@@ -278,10 +278,10 @@ module Katello
     end
 
     def v1_search
-      # Checks for podman client and issues a 404 in that case. Podman
+      # Checks for v2 client and issues a 404 in that case. Podman
       # examines the response from a /v1_search request. If the result
       # is a 4XX, it will then proceed with a request to /_catalog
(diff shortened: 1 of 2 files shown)


sjha4:
Before PR: 200 status code
200_Before_PR

After PR: 404 returned.
404_after_PR

Code makes sense after reading description. Ack 👍🏼

sjha4 (approved):

PR #11011, opened 2024-05-29
Fixes #37518 - Clear KS repo upon unregistration and show inherited hostgroup vals

ianballou:

Clears the kickstart repo when unregistering a host. Also make it more obvious when a cleared LCE, CV, content source, or synced content value on the host edit page is actually inheriting something from a host group.

The "Inherited" text will only pop up for the 4 specific fields mentioned above. With how the UI is, we'd need to specifically add the inherited check to every field that needs it. LCE, CV, and Synced Content (KS Repo) were the ones requested to be highlighted, and I threw content source in there too since it was in the same ERB file.

1) Create a hostgroup with synced content filled out and the fields needed to get that to work (LCE, CV, CS)
2) Register a host with that host group
3) Unregister the host
4) Edit the host
5) Click "manage host"
6) See that the 4 fields mentioned above are still filled out, but say "inherited".

TODO: ensure that it doesn't say "Inherited" if the value from the hostgroup is actually blank.

### DIFF
--- a/app/views/overrides/activation_keys/_host_environment_select.html.erb
+++ b/app/views/overrides/activation_keys/_host_environment_select.html.erb
@@ -16,25 +16,30 @@
 <% cs_select_name =  using_hostgroups_page? ? 'hostgroup[content_source_id]' : 'host[content_facet_attributes][content_source_id]' %>
 <% cs_select_attr = using_hostgroups_page? ? 'content_source' : 'content_facet.content_source' %>
 
-<%= field(f, cs_select_attr, {:label => _("Content Source")}) do
-  if using_hostgroups_page?
-    select_tag cs_select_id, content_source_options(@hostgroup, :include_blank => blank_or_inherit_with_id(f, :content_source)), :data => {"spinner_path" => spinner_path},
-               :class => 'form-control',  :name => cs_select_name
-  else
-    select_tag cs_select_id, content_source_options(@host, :selected_host_group => @hostgroup || @host.hostgroup, :include_blank => blank_or_inherit_with_id(f, :content_source)), :data => {"spinner_path" => spinner_path}, :class => 'form-control',  :name => cs_select_name, :disabled => cv_lce_disabled?
-  end
-end %>
+<%= field(f, cs_select_attr, {:label => _("Content Source"), :help_inline => content_source_inherited? ? 'Inherited' : nil }) do %>
+  <% if using_hostgroups_page? %>
+    <%= select_tag cs_select_id, content_source_options(@hostgroup, :include_blank => blank_or_inherit_with_id(f, :content_source)), :data => {"spinner_path" => spinner_path},
+               :class => 'form-control',  :name => cs_select_name %>
+  <% else %>
+    <%= select_tag cs_select_id, content_source_options(@host, :selected_host_group => @hostgroup || @host.hostgroup, :include_blank => blank_or_inherit_with_id(f, :content_source)), :data => {"spinner_path" => spinner_path}, :class => 'form-control',  :name => cs_select_name, :disabled => cv_lce_disabled? %>
+  <% end %>
+<% end %>
 
 <% env_select_id = using_hostgroups_page? ? :hostgroup_lifecycle_environment_id : :host_lifecycle_environment_id %>
 <% env_select_name =  using_hostgroups_page? ? 'hostgroup[lifecycle_environment_id]' : 'host[content_facet_attributes][lifecycle_environment_id]' %>
 <% env_select_attr = using_hostgroups_page? ? 'lifecycle_environment' : 'content_facet.single_lifecycle_environment' %>
 
-<%= field(f, env_select_attr, {:label => _("Lifecycle Environment")}) do %>
+<%= field(f, env_select_attr, {:label => _("Lifecycle Environment"), :help_inline => lifecycle_environment_inherited? ? 'Inherited' : nil}) do %>
   <% if using_hostgroups_page? %>
     <%= select_tag env_select_id, lifecycle_environment_options(@hostgroup, :include_blank => blank_or_inherit_with_id(f, :lifecycle_environment)), :class => 'form-control',  :name => env_select_name %>
   <% elsif cv_lce_disabled? %>
-    <%= hidden_field_tag 'host[content_facet_attributes][lifecycle_environment_id]', fetch_lifecycle_environment(@host).try(:id) %>
-    <%= select_tag env_select_id, lifecycle_environment_options(@host, :selected_host_group => @hostgroup || @host.hostgroup, :include_blank => blank_or_inherit_with_id(f, :lifecycle_environment)), :class => 'form-control',  :name => env_select_name, :disabled => true %>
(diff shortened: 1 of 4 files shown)


ianballou:
This is currently a bit broken -- the if statements can cause the form to error out because `nil` is getting returned.

app/views/overrides/activation_keys/_host_environment_select.html.erb
```diff
@@ -15,26 +15,37 @@
 <% cs_select_id =  using_hostgroups_page? ? :content_source_id : :content_source_id %>
 <% cs_select_name =  using_hostgroups_page? ? 'hostgroup[content_source_id]' : 'host[content_facet_attributes][content_source_id]' %>
 <% cs_select_attr = using_hostgroups_page? ? 'content_source' : 'content_facet.content_source' %>
+<% if !using_hostgroups_page? && !@host.content_source.present? && @host&.hostgroup&.content_source.present? && cv_lce_disabled? %>
+  <% content_source_inherited = true %>
+<% end %>
```
jeremylenz, line 20: Seems this may be cleaner if it were a method in `host_and_hostgroups_helper.rb`.

```rb
def content_source_inherited?(host)
  !using_hostgroups_page? && !@host.content_source.present? && @host&.hostgroup&.content_source.present? && cv_lce_disabled?
end
```

Then you could just call it here like
```rb
content_source_inherited?(@host)
```

app/views/overrides/activation_keys/_host_environment_select.html.erb
```diff
@@ -15,26 +15,37 @@
...
+  <% else %>
+    <%= select_tag cs_select_id, content_source_options(@host, :selected_host_group => @hostgroup || @host.hostgroup, :include_blank => blank_or_inherit_with_id(f, :content_source)), :data => {"spinner_path" => spinner_path}, :class => 'form-control',  :name => cs_select_name, :disabled => cv_lce_disabled? %>
+  <% end %>
+<% end %>
 
 <% env_select_id = using_hostgroups_page? ? :hostgroup_lifecycle_environment_id : :host_lifecycle_environment_id %>
 <% env_select_name =  using_hostgroups_page? ? 'hostgroup[lifecycle_environment_id]' : 'host[content_facet_attributes][lifecycle_environment_id]' %>
 <% env_select_attr = using_hostgroups_page? ? 'lifecycle_environment' : 'content_facet.single_lifecycle_environment' %>
+<% if !using_hostgroups_page? && @host.lifecycle_environments.empty? && @host&.hostgroup&.lifecycle_environment.present? && cv_lce_disabled? %>
+  <% lifecycle_environment_inherited = true %>
```
jeremylenz, line 35: same comment as above

app/views/overrides/activation_keys/_host_environment_select.html.erb
```diff
@@ -43,7 +54,11 @@ end %>
 <% cv_select_id =  using_hostgroups_page? ? :hostgroup_content_view_id : :host_content_view_id %>
 <% cv_select_name =  using_hostgroups_page? ? 'hostgroup[content_view_id]' : 'host[content_facet_attributes][content_view_id]' %>
 <% cv_select_attr = using_hostgroups_page? ? 'content_view' : 'content_facet.single_content_view' %>
-<%= field(f, cv_select_attr, {:label => _("Content View")}) do %>
+<% if !using_hostgroups_page? && @host.content_views.empty? && @host&.hostgroup&.content_view.present? && cv_lce_disabled? %>
+  <% content_view_inherited = true %>
```
jeremylenz, line 58: same comment as above

app/views/overrides/activation_keys/_host_synced_content_select.html.erb
```diff
@@ -18,10 +18,14 @@
 
 <% spinner_path = asset_path('spinner.gif') %>
 
-<%= field(f, ks_repo_select_attr, {:label => _("Synced Content")}) do
-  select_tag ks_repo_select_id,  view_to_options(kickstart_options, kickstart_repo_id, blank_or_inherit_with_id(f, :kickstart_repository)), :data => {"spinner_path" => spinner_path, "kickstart-repository-id" => kickstart_repo_id},
-             :class => 'form-control',  :name => ks_repo_select_name, :disabled => kickstart_options.empty?
-end %>
+<% if !using_hostgroups_page? && !@host.kickstart_repository_id.present? && @host&.hostgroup&.kickstart_repository.present? && cv_lce_disabled? %>
+  <% kickstart_repo_inherited = true %>
```
jeremylenz, line 22: same comment as above

jeremylenz (commented):
Thanks @ianballou!

Clicked around and everything seems to work as expected. :+1: 

One code comment x4, below :)

ianballou:
@jeremylenz how does that look?

app/views/overrides/activation_keys/_host_environment_select.html.erb
```diff
@@ -15,26 +15,33 @@
 <% cs_select_id =  using_hostgroups_page? ? :content_source_id : :content_source_id %>
 <% cs_select_name =  using_hostgroups_page? ? 'hostgroup[content_source_id]' : 'host[content_facet_attributes][content_source_id]' %>
 <% cs_select_attr = using_hostgroups_page? ? 'content_source' : 'content_facet.content_source' %>
+<% content_source_inherited = true if content_source_inherited? %>
```
jeremylenz, line 18: Now that we've moved the logic to the helper, I don't see a reason to assign these to variables anymore..

app/views/overrides/activation_keys/_host_environment_select.html.erb
```diff
@@ -15,26 +15,33 @@
...
 
-<%= field(f, cs_select_attr, {:label => _("Content Source")}) do
-  if using_hostgroups_page?
-    select_tag cs_select_id, content_source_options(@hostgroup, :include_blank => blank_or_inherit_with_id(f, :content_source)), :data => {"spinner_path" => spinner_path},
-               :class => 'form-control',  :name => cs_select_name
-  else
-    select_tag cs_select_id, content_source_options(@host, :selected_host_group => @hostgroup || @host.hostgroup, :include_blank => blank_or_inherit_with_id(f, :content_source)), :data => {"spinner_path" => spinner_path}, :class => 'form-control',  :name => cs_select_name, :disabled => cv_lce_disabled?
-  end
-end %>
+<%= field(f, cs_select_attr, {:label => _("Content Source"), :help_inline => content_source_inherited ? 'Inherited' : nil }) do %>
```
jeremylenz, line 20: You can just call the methods directly here like this, no? And then you can eliminate the four variable assignments.

```suggestion
<%= field(f, cs_select_attr, {:label => _("Content Source"), :help_inline => content_source_inherited? ? 'Inherited' : nil }) do %>
```

app/views/overrides/activation_keys/_host_environment_select.html.erb
ianballou: yeah good point :)

ianballou:
@jeremylenz now it should be all set!

jeremylenz (approved):
LGTM :+1:

jeremylenz:
RHEL8 full support ended 2024-05-31, so test failures are unrelated. We should fix that in a separate PR tho :stuck_out_tongue: 

```
Error: Failure: test_0001_can update the rhel lifecycle status

Expected: 1
  Actual: 2
```

ianballou:
Only test failure is the lifecycle one, so merging!

PR #11013, opened 2024-05-31
Fixes #37522 - remove sha1 publication checksum and add newer values

ianballou:
Removes SHA1 publication support because it was dropped in pulp-rpm 3.25. SHA1 content can still be synced, this only has to do with what Pulp publishes. EL6+ is still supported.

Also adds support for some more checksum types to match up with Pulp.

Removes the `ensure_no_checksum_on_demand` validation since that's from Pulp 2.

Deprecating the feature would've been better, but the amount of folks using SHA1 for a legitimate reason should be very rare. Hanging back a Pulp version just for this doesn't seem to make sense.

1) Ensure that repositories with a checksum type of SHA1 are migrated to "Default"
  -> To do this, it might be best to start with a Katello 4.12 and then upgrade it.
    -> This would also help test that repositories & CVs are still usable if the underlying publications used SHA1.
2) Ensure that all the new checksum types work
3) Ensure that the API help text makes sense (Hammer too)

### DIFF
--- a/app/controllers/katello/api/v2/repositories_controller.rb
+++ b/app/controllers/katello/api/v2/repositories_controller.rb
@@ -48,7 +48,7 @@ Pass [] to make repo available for clients regardless of OS version. Maximum len
       param :ssl_client_cert_id, :number, :desc => N_("Identifier of the content credential containing the SSL Client Cert"), :allow_nil => true
       param :ssl_client_key_id, :number, :desc => N_("Identifier of the content credential containing the SSL Client Key"), :allow_nil => true
       param :unprotected, :bool, :desc => N_("true if this repository can be published via HTTP")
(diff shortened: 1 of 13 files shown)


chris1984 (commented):

ianballou:
> Removes the `ensure_no_checksum_on_demand` validation since that's from Pulp 2.

I put this back -- it sounds like Pulp ignores the checksum type for on demand repositories, so I think it's still a valid thing to skip.

chris1984 (approved):
ACK works great,

* Repo got migrated from `sha1` to `default` in UI after upgrading from 4.12 to 4.13 and applying pr
* Able to do `yum repolist -v` on client after upgrade to 4.13 and applying packit build without errors. 
* Able to see repos of `sha256/384/512` on my client

PR #11014, opened 2024-06-04
Fixes #37356 - Allow turning automatic content count updates off

sjha4:
Add a setting "automatic_content_count_updates" to control automatic invocation of Update content count task on proxies upon successful syncs. 
The Update content counts action is triggered asynchronously upon a successful sync on a proxy.
This task can be a long running task for large proxies on a slow connection and this can cause slowness in syncs in the queue.
1. Turn the setting "Calculate content counts on smart proxies automatically" on and off and check if Update content counts is run automatically or not depending on the value of the setting.

### DIFF
--- a/lib/katello/plugin.rb
+++ b/lib/katello/plugin.rb
@@ -651,6 +651,12 @@ Foreman::Plugin.register :katello do
         default: true,
         full_name: N_('Distribute archived content view versions'),
         description: N_("If this is enabled, repositories of content view versions without environments (\"archived\") will be distributed at '/pulp/content/<organization>/content_views/<content view>/X.Y/...'.")
+
+      setting 'automatic_content_count_updates',
+        type: :boolean,
+        default: true,
+        full_name: N_('Calculate content counts on smart proxies automatically'),
+        description: N_("If this is enabled, content counts on smart proxies will be updated automatically after content sync.")
     end
   end
 
--- a/app/lib/actions/katello/capsule_content/sync_capsule.rb
+++ b/app/lib/actions/katello/capsule_content/sync_capsule.rb
@@ -69,8 +69,13 @@ module Actions
         end
 
         def update_content_counts(_execution_plan)
-          smart_proxy = ::SmartProxy.unscoped.find(input[:smart_proxy_id])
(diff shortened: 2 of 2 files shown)


sjha4:
Rebased for pipeline..

sbernhard:
> This task can be a long running task for large proxies on a slow connection and this can cause slowness in syncs in the queue.

I wonder, why a async task can slow down other tasks. Instead of adding a new setting and make this feature configuruable, wouldn't it be better to think about how this feature can be improved?

sjha4:
This seems to happen for users with slow master -> proxy connections and the API call to pulp takes a long time. This is a band-aid for such users to turn it off and run this manually when they need to.

We have some improvements to this in the pipeline which include making the task granular and only run on contents that have actually changed but that will be follow up PR.

sbernhard (approved):
LGTM

lib/katello/plugin.rb
```diff
@@ -651,6 +651,12 @@ def katello_template_setting_values(name)
         default: true,
         full_name: N_('Distribute archived content view versions'),
         description: N_("If this is enabled, repositories of content view versions without environments (\"archived\") will be distributed at '/pulp/content/<organization>/content_views/<content view>/X.Y/...'.")
+
+      setting 'automatic_content_count_updates',
+              type: :boolean,
+              default: true,
+              full_name: N_('Calculate content counts on smart proxies automatically'),
+              description: N_("If this is enabled, content counts on smart proxies will be updated automatically after content sync.")
```
wbclark, line 659:
```suggestion
      setting 'automatic_content_count_updates',
        type: :boolean,
        default: true,
        full_name: N_('Calculate content counts on smart proxies automatically'),
        description: N_("If this is enabled, content counts on smart proxies will be updated automatically after content sync.")
```

Nit: to match alignment with the other settings

wbclark (approved):
Sorry for the delay here and thanks for your patience.

Code LGTM other than the alignment nit.

I thought about asking for a test, but looking at test/actions/katello/capsule_content_test.rb and then app/lib/actions/katello/capsule_content/update_content_counts.rb there doesn't seem enough need

lib/katello/plugin.rb
sjha4: Updated to match indentation..Will merge when green..

PR #11015, opened 2024-06-04
Fixes #37534 - Use ~ as default operator for repository content labels

jeremylenz:

On the Red Hat Repositories page, searching for a keyword only (without field names) does a preliminary SQL query to the `katello_contents` table:

```sql
SELECT "katello_contents"."cp_content_id" FROM "katello_contents" WHERE (label = 'coreservices')
```

The problem here is that the operator being used is `=`, which only matches exact results. Instead we should use the `~` operator.

This PR alters the scoped_search call to set the [`default_operator`](https://github.com/wvanbergen/scoped_search/wiki/search-definition#setting-the-default-operator) to `:like`, which changes the query to

```sql
SELECT "katello_contents"."cp_content_id" FROM "katello_contents" WHERE (label ILIKE '%%coreservices%%')
```

This should allow the Red Hat Repositories page to show the same results for both the left column, which searches `Katello::ProductContent`s, and the right column, which searches `Katello::Repository`s.

I don't foresee any detrimental side effects here (famous last words..)

Follow the steps in https://issues.redhat.com/browse/SAT-23768

If you like, you can enable SQL logging by adding the following to your Foreman's `config/settings.yaml`

```yaml
  :sql:
    :enabled: true
    :level: debug
```

### DIFF
--- a/app/models/katello/repository.rb
+++ b/app/models/katello/repository.rb
@@ -166,7 +166,7 @@ module Katello
     scoped_search :on => :name, :relation => :product, :rename => :product_name
     scoped_search :on => :id, :relation => :product, :rename => :product_id, :only_explicit => true
     scoped_search :on => :label, :relation => :root, :complete_value => true, :only_explicit => true
-    scoped_search :on => :content_label, :ext_method => :search_by_content_label
+    scoped_search :on => :content_label, :ext_method => :search_by_content_label, :default_operator => :like
 
     delegate :product, :redhat?, :custom?, :to => :root
     delegate :yum?, :docker?, :deb?, :file?, :ostree?, :ansible_collection?, :generic?, :to => :root


jeremylenz:
test failures are due to https://projects.theforeman.org/issues/37533

app/models/katello/repository.rb
```diff
@@ -166,7 +166,7 @@ class Repository < Katello::Model
     scoped_search :on => :name, :relation => :product, :rename => :product_name
     scoped_search :on => :id, :relation => :product, :rename => :product_id, :only_explicit => true
     scoped_search :on => :label, :relation => :root, :complete_value => true, :only_explicit => true
-    scoped_search :on => :content_label, :ext_method => :search_by_content_label
+    scoped_search :on => :content_label, :ext_method => :search_by_content_label, :default_operator => :like
```
wbclark, line 169: question: is https://www.rubydoc.info/gems/scoped_search/ScopedSearch/Definition/Field#default_operator-instance_method the right place to look, to understand this does?

wbclark: ah, TY!

will give this a shot shortly when my EL9 dev environment finishes provisioning

jeremylenz:
rebased

wbclark:
Code LGTM, re-running test to see if it passes now

wbclark (approved):
it worked well in my testing

I can search, e.g. for `appstream-rpms x86` and it returns `rhel-9-for-x86_64-appstream-rpms` which is nice, `appstream x86 eus` returns `rhel-8-for-x86_64-appstream-eus-rpms` which is great

one line of changed code, one major improvement in usability :+1:

PR #11016, opened 2024-06-05
Fixes #37533 - Stub date in test so it doesn't rely on real system clock

jeremylenz:

RHEL8 full support ended on 2024-05-31, which is causing some unit tests to fail in CI. This PR updates the relevant test to mock `Date.today`, so that the test doesn't rely on the real system clock.

I could have instead updated the test expectations, but then we'd just have to do it again once the next milestone date rolls around.

ensure that CI tests pass


chris1984 (approved):
APJ

PR #11017, opened 2024-06-05
Fixes #37538 - Check orphaned content facet before CV deletion

sjha4:
Check for orphaned content facet before CV deletion, removal from env
We have this check for CV publish and we should also prevent this error and inform users about the cleanup task during deletion/remove from environment.
1. Follow steps in https://github.com/Katello/katello/pull/10391 to get orphaned content host for a CV.
2. Try deleting the CV.

### DIFF
--- a/app/models/katello/content_view.rb
+++ b/app/models/katello/content_view.rb
@@ -710,6 +710,7 @@ module Katello
     end
 
     def check_remove_from_environment!(env)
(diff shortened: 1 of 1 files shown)


chris1984 (approved):
ACK.

After running the rake task I was able to delete the content view:
Screenshot 2024-06-13 at 11-19-13 Tasks

PR #11018, opened 2024-06-06
Fixes #37539 - replace deprecated webpack_asset_paths in smart proxies

MariaAga:
Tested only on mock data by disabling the check on lib/katello/plugin.rb `:onlyif => proc { |proxy| proxy.pulp_mirror? }`  so a real test should be done as well.
'`webpack_asset_paths` is deprecated, use `content_for(:javascripts) { webpacked_plugins_js_for(plugin_name) }` instead.
It looks like the page didnt find the Content component, but it was not loaded since katello/index.js was not loaded.


sjha4:
Do you know why this was working on centos8 and not 9? Node versions?

sjha4 (approved):
Thanks @MariaAga ! The component loads on EL8 and El9 boxes now. Ack 🎉 
Screenshot from 2024-06-06 09-51-06

PR #11019, opened 2024-06-06
Refs #37481 - Update cached value for manifest expiration date during delete

jeremylenz:

In https://github.com/Katello/katello/pull/10999 we invalidated the cache of `manifest_expiration_date` and `manifest_imported?` when importing a new manifest, but neglected to do this for manifest deletes.

Import an expired manifest
Delete the manifest
Open Manage Manifest
Ensure it says "No manifest imported" and does not have an alert or a date

### DIFF
--- a/app/lib/actions/katello/organization/manifest_delete.rb
+++ b/app/lib/actions/katello/organization/manifest_delete.rb
@@ -16,8 +16,13 @@ module Actions
             repositories.each do |repo|
               plan_action(Katello::Repository::RefreshRepository, repo)
             end
-            plan_self
+            plan_self(:organization_name => organization.name)
           end
+        end
+
+        def run
+          organization = ::Organization.find_by(name: input[:organization_name])
+          organization&.manifest_expiration_date(cached: false) # update organization.manifest_imported? value
         end
 
         def failure_notification(plan)
--- a/webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
+++ b/webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
@@ -200,7 +200,7 @@ class ManageManifestModal extends Component {
                           {getManifestName()}
                         </Col>
                       </Row>
-                      {isManifestImported && !!manifestExpirationDate &&
+                      {isManifestImported && Boolean(manifestExpirationDate) &&
                         <Row>
                           <Col sm={5} />
                           <Col sm={7}>
--- a/webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
+++ b/webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
@@ -137,6 +137,8 @@ class ManageManifestModal extends Component {
       return name;
     };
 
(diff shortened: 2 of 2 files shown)


chris1984 (commented):
Code wise it looks fine, starting to test

chris1984:
Oh it looks like @qcjames53 picked this up, I won't test

jeremylenz:
VJ actually already tested this on Packit and he says it works :smile: but feel free to test on dev as well!

qcjames53 (commented):
Hi Jeremy! I was able to confirm this works properly when a manifest is already expired on upload and when a manifest expires 'naturally'. I just had two small questions before acking:

app/lib/actions/katello/organization/manifest_delete.rb
```diff
@@ -16,10 +16,15 @@ def plan(organization)
             repositories.each do |repo|
               plan_action(Katello::Repository::RefreshRepository, repo)
             end
-            plan_self
+            plan_self(:organization_name => organization.name)
```
qcjames53, line 19: I don't see why this would be an issue for lookup, but is there a reason to use name over id here?

webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
```diff
@@ -198,7 +200,7 @@ class ManageManifestModal extends Component {
                           {getManifestName()}
                         </Col>
                       </Row>
-                      {isManifestImported && manifestExpirationDate &&
+                      {isManifestImported && !!manifestExpirationDate &&
```
qcjames53, line 203: I think a direct check for falsiness, `instanceof Date`, or even `!== null` would be clearer. I just haven't encountered this `!!` syntax much before; what are your thoughts?

app/lib/actions/katello/organization/manifest_delete.rb
jeremylenz: Just consistency with `manifest_import.rb`. In the [previous PR](https://github.com/Katello/katello/pull/10999) the `plan_self` was already there with `name` so I just did it the same way here.

webpack/scenes/Subscriptions/Manifest/ManageManifestModal.js
jeremylenz: It's kinda like `Foreman::Cast.to_bool` except not as cool :)

I'm taking advantage here of the fact that `0` is falsy in JavaScript. If `manifestExpirationDate` is `0`, it would evaluate to the 01 Jan 1969 date and I won't want it rendered.

I could just leave it as `manifestExpirationDate` (without the `!!`), but that would risk actually rendering a "0" on the page somewhere. But `false` won't be rendered.

image

Would you be okay with `Boolean(manifestExpirationDate)`? That's a bit more explicit and I think it will provide the same result.

qcjames53: I'm totally ok with the explicit cast; that sounds like a good plan.

app/lib/actions/katello/organization/manifest_delete.rb
qcjames53: Okay, that's totally reasonable. This should be fine 👍

jeremylenz:
@qcjames53 updated :+1:

qcjames53 (approved):
Looks all good to me. Acking 👍

PR #11020, opened 2024-06-06
Fixes #37543 - CV promote during host deletion fails

ianballou:
Adds a rescue to skip host applicability status updates when the host no longer exists.

Follow the reproducer steps in the redmine, or:

1) Stick a debugger right before `facet.update_applicability_counts`
2) Run `update_host_statuses` in a Rails console
3) When the breakpoint is reached, delete the Host via some other means (like the UI or hammer)
4) Continue the code
5) See that the code keeps working and the logger statement is triggered

### DIFF
--- a/app/models/katello/content_view.rb
+++ b/app/models/katello/content_view.rb
@@ -454,6 +454,8 @@ module Katello
           ).each do |facet|
             facet.update_applicability_counts
             facet.update_errata_status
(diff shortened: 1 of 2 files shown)


sjha4 (approved):
Works as described and code looks good.
Screenshot from 2024-06-12 14-40-42

Ack 👍🏼

PR #11021, opened 2024-06-07
Fixes #37544 - properly access host for hostgroup inheritance checks

ianballou:
A continuation from https://github.com/Katello/katello/pull/11011, `@host` was being used where it shouldn't be.

I also adjusted the logic and help text for kickstart repositories. I realized that the kickstart repo is never truly inherited from the hostgroup for edited hosts. Instead, it come from the operating system and other selected fields.

I also noticed that the LCE selector was empty when editing hosts, so I fixed that too.

The new KS repo help text is a bit verbose, but I think it's best to be fully clear about where that value comes from. For a host with no existing kickstart repository, it's not completely clear how the field is generated. We do have a tool-tip next to to the Media Selection, but this makes it even more obvious.

Follow the steps from https://github.com/Katello/katello/pull/11011
Help verify that the help text makes sense based on the previously raised bugs.

Check new hosts + host groups, edited hosts + host groups.

Generally make sure the host and host group pages still work, they're so fragile... I've checked as much as I could think to check.

### DIFF
--- a/app/helpers/katello/hosts_and_hostgroups_helper.rb
+++ b/app/helpers/katello/hosts_and_hostgroups_helper.rb
@@ -12,20 +12,20 @@ module Katello
       edit_action? && !using_discovered_hosts_page?
     end
 
-    def content_source_inherited?
-      !using_hostgroups_page? && @host.content_source.blank? && @host&.hostgroup&.content_source.present? && cv_lce_disabled?
+    def content_source_inherited?(host)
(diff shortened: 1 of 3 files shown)


parthaa:
Code looks good to me

chris1984 (approved):
Able to build a box with the changes and didn't hit the error, code looks fine. ACK

PR #11022, opened 2024-06-07
Fixes #37288 - Host Collections widget now uses errata_status_installable Setting

jeremylenz:

The Host Collections dashboard widget shows you a colored box:

red if there are security errata for the hosts in that host collection;
yellow if there are bugfix or enhancement errata;
and green otherwise.

With this PR, the Host Collections widget now uses the `errata_status_installable` Setting so that these colors can be based on installable errata and not just applicable, if desired by the user.

Also added some tooltips.
image

It made more sense to me to use `applicable_for_hosts` and `installable_for_hosts` side-by-side. Let me know if that's incorrect. also, I added keyword arguments to a method that already had a positional argument; hopefully that doesn't break any Ruby things

Get one or more hosts with applicable but not installable errata
Put those hosts in a host collection
With the `errata_status_installable` setting on: The host collections widget should show a green box if none of the errata are installable.
With the `errata_status_installable` setting off: The host collections widget should show a red or yellow box, depending on what errata types are applicable.

### DIFF
--- a/app/models/katello/host_collection.rb
+++ b/app/models/katello/host_collection.rb
@@ -58,8 +58,12 @@ module Katello
       consumer_ids
     end
 
-    def errata(type = nil)
-      query = Erratum.joins(:content_facets).where("#{Katello::Host::ContentFacet.table_name}.host_id" => self.host_ids)
+    def errata(type = nil, installable_only: false)
+      query = if installable_only
+                Katello::Erratum.installable_for_hosts(self.hosts)
+              else
+                Katello::Erratum.applicable_to_hosts(self.hosts)
+              end
       type ? query.of_type(type) : query
     end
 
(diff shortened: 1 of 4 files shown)


jeremylenz:
> Please remind me how to get an errata to be applicable and not installable to test this? 😸

The packages in the errata must be available in the library instances of the repo, but not the host's CV version's repo.

You can use a content view filter of a package that is part of an errata. Or, sometimes easiest way is to publish a CV version, assign it to your host, then sync the repos in that CV so that some new errata show up (but don't publish again). That's what I did for RHEL8 BaseOS and AppStream I think..

sjha4 (approved):
Able to switch between installable/applicable warnings on host collection widget based on the setting. Ack 👍🏼 
Screenshot from 2024-06-20 15-26-06

PR #11023, opened 2024-06-07
Fixes #37546 - Disable BulkChangeHostCVModal when Any Organization is selected

jeremylenz:
Goes with https://github.com/theforeman/foreman/pull/10199 (not strictly required tho)

The `BulkChangeHostCVModal` was throwing an error on the HostsIndex page when 'Any organization' was selected as the organization context.

We need to make sure the HostsIndex page handles 'Any organization' correctly. 

With this change, 'Change content view environments' is disabled and the modal does not render unless you have selected a specific organization.

You can't change the content view of a host to one from a different org, so that action should be disabled until you select an organization.

Select 'Any organization'
Make sure there are no JS console errors on HostsIndex page
Observe that 'Change content view environments' menu item is disabled


jeremylenz:
> Observe that 'Change content view environments' menu item is disabled - Pass

sorry, forgot to say that normally you must have at least one host selected, for this to be enabled. So observing it with no hosts selected doesn't really confirm anything.

PR #11024, opened 2024-06-10
Fixes #37549 - Failed to index content due to duplicate errata

hao-yu:
Content Import task failed indexing errata from Pulp with the following error.

~~~
13: from katello (4.11.0.9) app/models/katello/repository.rb:960:in `index_content'
       12: from katello (4.11.0.9) app/models/katello/repository.rb:960:in `each'
       11: from katello (4.11.0.9) app/models/katello/repository.rb:961:in `block in index_content'
        7: from katello (4.11.0.9) app/services/katello/content_unit_indexer.rb:26:in `each'
        3: from katello (4.11.0.9) app/services/katello/pulp3/pulp_content_unit.rb:139:in `block (2 levels) in pulp_units_batch_for_repo'
        2: from katello (4.11.0.9) app/services/katello/pulp3/pulp_content_unit.rb:139:in `yield'
        1: from katello (4.11.0.9) app/services/katello/content_unit_indexer.rb:44:in `block in import_all'
ActiveRecord::StatementInvalid (PG::CardinalityViolation: ERROR:  ON CONFLICT DO UPDATE command cannot affect row a second time)
HINT:  Ensure that no rows proposed for insertion within the same command have duplicate constrained values.
~~~

 

Is this issue a regression from an earlier version:

Yes. The following bug is fixed in previous Satellite versions, but the code didn't brought forward after refactoring the index content codes.

https://bugzilla.redhat.com/show_bug.cgi?id=2070535

 

Additional info:

Due to some bugs in Pulp, it is still possible to have duplicate errata associated to a repository version. It is causing issue to Katello because Katello doesn't expect duplicate errata in a repository version. **

https://github.com/pulp/pulp_rpm/issues/2821

https://github.com/pulp/pulp_rpm/issues/3587

### DIFF
--- a/app/services/katello/content_unit_indexer.rb
+++ b/app/services/katello/content_unit_indexer.rb
@@ -38,6 +38,12 @@ module Katello
             end
           end
 
+          # Even after this bug (https://github.com/pulp/pulp_rpm/issues/2821) is fixed,
+          # it is possible to have duplicate errata asosociated to a repo.
+          if @content_type.label == 'erratum'
+            to_insert.uniq! { |row| row["pulp_id"] || row[:pulp_id] }
+          end
+
           next if to_insert.empty?
           insert_timestamps(to_insert)
           if @content_type.mutable


app/services/katello/content_unit_indexer.rb
```diff
@@ -38,6 +38,12 @@ def import_all(filtered_indexing = false)
             end
           end
 
+          # Even after this bug (https://github.com/pulp/pulp_rpm/issues/2821) is fixed,
+          # it is possible to have duplicate errata asosociated to a repo.
+          if @content_type.label == 'erratum'
```
ianballou, line 43: The units in `to_insert` and `@final_values` below both seem to come from `units_from_pulp`.  Here, `to_insert` comes from the iteration through `units_from_pulp`. Below, `@final_values` comes from `@values` through `::Katello::ContentUnitIndexer.pulp_id_to_id_map`.  `@values` is populated through `def push`, which is also fed by `units_from_pulp`.

With all this in mind, would it make sense to filter out the duplicate errata in `units_from_pulp` itself so it doesn't need to be done twice?

ianballou: Likewise, I noticed there's a `remove_duplicate` method being called -- would it make sense for this logic to live there?

ianballou: Although, this duplicate checking should likely happen before the content units are used, so `remove_duplicates` might be too late. Perhaps it could happen in `units_from_pulp` itself.

ianballou: I'm curious why you needed to remove duplicates in two places rather than one.

ianballou: Now this is the only one, nice :)

ianballou:
@hao-yu mind trying a rebase? I'm pretty sure that "CI / Angular bastion_katello" error is unrelated, but it's strange.

Other than the test issue this PR looks great.

hao-yu:
@ianballou rebased!

ianballou (approved):
Looking great! Tested with RHEL content.

PR #11026, opened 2024-06-11
Fixes #37555 - Move capsule index and show to correct :view_smart_proxies permission

chris1984:

* Viewing a smart proxy was giving the wrong permissions needed `manage_capsule_content`

* Check out PR
* Create a non-admin user
* Create a role with only `manage_capsule_content` and assign it to that user
* `curl -k -u test:pass https://localhost/katello/api/capsules/X`
* Verify you get the correct permission to add to see the smart proxies
* Once you add `view_capsule_content` see if the curl returns output

### DIFF
--- a/lib/katello/permission_creator.rb
+++ b/lib/katello/permission_creator.rb
@@ -62,7 +62,8 @@ module Katello
       @plugin.permission :view_capsule_content,
                          {
                            'katello/api/v2/capsule_content' => [:counts, :lifecycle_environments, :available_lifecycle_environments, :sync_status],
-                           'smart_proxies' => [:pulp_storage, :pulp_status, :show_with_content]
+                           'smart_proxies' => [:pulp_storage, :pulp_status, :show_with_content],
+                           'katello/api/v2/capsules' => [:index, :show]
                          },
                          :resource_type => "SmartProxy"
     end


lib/katello/permission_creator.rb
```diff
@@ -51,11 +51,16 @@ def activation_keys_permissions
     end
 
     def capsule_content_permissions
+      @plugin.permission :view_smart_proxies,
```
sjha4, line 54: Thoughts on using view_capsule_content permission to manage this access instead of the new permission?

chris1984: Let me check with Vlad and if approves then I will move it.

chris1984: Vlad said this:

chris1984: @sjha4 I am ok with whatever you want to roll with based on Vlad's feedback

sjha4: I am still leaning towards the old permission. Let's check with @parthaa ..

chris1984: @sjha4 updated

sjha4 (approved):
Works well..Getting right permissions missing messages for the proxies. 👍🏼 
Screenshot from 2024-06-25 09-34-05
Screenshot from 2024-06-25 09-43-47

PR #11027, opened 2024-06-12
Fixes #37174 - Honor selected organizations in hg form

adamruzicka:
when generating options for lifecycle environment field.

A spiritual successor to https://github.com/Katello/katello/pull/10524

When creating/editing a hostgroup, only lifecycle environments belonging to the hostgroup's organization are offered.

This is one of the things that seems to be conceptually broken. On one side we have a hostgroup which can belong to multiple (or none) organizations and we're trying to match it with a lifecycle environment which belongs to exactly one organization.

With current organization set when creating a hostgroup, the current organization gets pre-selected for the hostgroup, leading to what appears to be expected behavior. If the user add more organizations, they'll be able to pick LCEs from all of them.

1) Have 2 organizations
2) Go create a hostgroup
3) Set content source
4) Open the lifecycle environment picker

### DIFF
--- a/app/assets/javascripts/katello/hosts/host_and_hostgroup_edit.js
+++ b/app/assets/javascripts/katello/hosts/host_and_hostgroup_edit.js
@@ -36,15 +36,19 @@ KT.hosts.fetchEnvironments = function () {
   select.find('option').remove();
   if (content_source_id) {
     var url = tfm.tools.foremanUrl('/katello/api/capsules/' + content_source_id);
-    var orgId = $('#host_organization_id').val();
+    var orgIds = $("#hostgroup_organization_ids").val();
+    if(orgIds === undefined || orgIds === null || orgIds.length === 0) {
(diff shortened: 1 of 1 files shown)


wbclark (approved):
Testing with/without this commit, not only do I now see only the LCEs that I assigned to my current Organization, it also prevents me from seeing three different Library environments. :^)

wbclark:
I've confirmed that this works as expected for the Create Host page as well. Steps taken:

I have "Default Organization", "Test_Org1", and "Test_Org2". For each of the non-default Orgs, I have corresponding LCEs Test_Lce1 and Test_Lce2, respectively.

On the Create Host page, I can select any of these Organizations for the new host, regardless of the current context (i.e. organization selection on the top bar). The available LCEs for the new host are properly limited by the currently selected Organization in the form.

So, everything looks good here, and the code itself is very straightforward. LGTM and thanks @adamruzicka !

PR #11028, opened 2024-06-12
Fixes #37552 - container push repo content view support

ianballou:
- Allows publishing container push repositories in content views
- Allows deletion of container push repositories
- Adds authorization checks for container push
- Skips metadata generation for container push repositories
- Adds translations for container push errors

Other update actions on container push repositories will cause weirdness. I considered fixing that here too, but I think it'll be better for another PR. We'll need to change up the repo edit UI for container push repos.

1) Push some container repos to Katello and publish them in a content view
2) Try filtering content with CV publishes
3) Try deleting container push repos
4) Try pushing repositories as a user that shouldn't be able to (perhaps only has ability to create personal access tokens)
5) Try regenerating metadata for container push repositories -- see that nothing fails


app/lib/actions/katello/environment/publish_repositories.rb
```diff
@@ -15,10 +15,12 @@ def plan(env, options = {})
           concurrence do
             repositories.each do |repository|
               sequence do
-                repository.set_container_repository_name
-                repository.clear_smart_proxy_sync_histories
-                plan_action(::Actions::Katello::Repository::InstanceUpdate, repository)
-                plan_action(::Actions::Katello::Repository::CapsuleSync, repository)
+                unless repository.root.is_container_push && repository.library_instance?
```
sjha4, line 18: Thoughts on changing the name of this action/file to specify `Container` since it's only called for those?

ianballou: There's an `unless` there, so it's for all repo types except container push ones that are in Default Organization View.

ianballou:
I realize I need to add some tests.

app/lib/actions/katello/environment/publish_repositories.rb
sjha4: This flow is called only for container repos though right? The only place this is called is in that flow and also I see `repository.set_container_repository_name` as the first thing in this action..

sjha4: The only place I see this getting called from passes Docker as content type. I used simple text search so I might have missed any other calls. Otherwise I think we should streamline that flow and update the name of the action to be more readable.

ianballou: I see what you mean, yeah this is totally a Docker specific action now. I wonder if once upon a time the action was used elsewhere.
I'll change it to `PublishContainerRepositories`

app/lib/actions/katello/environment/publish_container_repositories.rb
```diff
@@ -1,24 +1,26 @@
...
 
         input_format do
           param :id
           param :name
         end
 
-        def plan(env, options = {})
-          repositories = options[:content_type] ? env.repositories.with_type(options[:content_type]) : env.repositories
+        def plan(env)
+          repositories = env.repositories
```
sjha4, line 13:
```suggestion
          repositories = env.repositories.with_type(Katello::Repository::DOCKER_TYPE)
```
We'd need something like above to limit to container repos with this refactoring.

ianballou: Forgot that bit, thanks :+1:

ianballou: Fixed in the latest commit with `.docker_type`.

ianballou:
One test should be fixed, a few more to go in registry proxies controller test. There are quite a few mocks to deal with :)

ianballou:
I made a bunch of test updates. I also made the decision that `unauthenticated_pull` should not allow unauthenticated pushing of repositories -- that sounds unsafe. For now we can just require authentication for any pushing of content. If people request it, we could add an `unauthenticated push` setting on LCEs.

app/lib/actions/pulp3/orchestration/repository/delete.rb
```diff
@@ -9,10 +9,13 @@ def plan(repository, smart_proxy)
               plan_action(Actions::Pulp3::Repository::DeleteDistributions, repository.id, smart_proxy)
 
               if repository.content_view.default?
-                #we're deleting the library instance, so just delete the whole pulp3 repo
+                # Container push repositories must be deleted through the distribution
+                return if repository.root.is_container_push
+
```
sjha4, line 14: Why do we need this? Should we be able to delete unpublished container-push repos in pulp?

sjha4:
Everything looks good here..Was able to go through the general CV workflows with container push repos. 

One note above which is because after deleting the product, I was left with a container repo in pulp untracked by katello.

app/lib/actions/pulp3/orchestration/repository/delete.rb
ianballou: In Pulp Container, a container push repo is tightly-coupled to its distribution. A container push repository cannot be deleted in fact, only its distribution can be deleted.

ianballou:
> One note above which is because after deleting the product, I was left with a container repo in pulp untracked by katello.

Interesting, was the DeleteDistributions step skipped? If it is, then the Pulp repository won't get deleted. I'm curious if you can consistently reproduce the issue.

sjha4:
Will retrace the steps but I think it wa the library push repo without any CVs that I tried deleting first and it left the library repo in pulp.

sjha4:
So I tried reproducing and it happens only when the repo push actually fails. For ex: I tried a push with a v1 manifest podman pull quay.io/aptible/hello-world . That created the repo in katello/pulp and then deleting the katello repo left a pulp repo.

ianballou:
@sjha4 I fixed the issue by allowing distributions to get looked up even if the manifest push part fails. Previously, the blob upload would succeed but the manifest upload would fail. This caused a Pulp repo to get created that Katello wasn't keeping proper track of.

sjha4:
I think there is a workflow that leads to an orphaned pulp repo. If you push a container repo, publish it in a CV version and then delete the repository from the repo details page  while selecting remove from all CV versions, there is still a repo left in pulp db after the task. All the distributions get deleted though.

ianballou:
> I think there is a workflow that leads to an orphaned pulp repo. If you push a container repo, publish it in a CV version and then delete the repository from the repo details page while selecting remove from all CV versions, there is still a repo left in pulp db after the task. All the distributions get deleted though.

I did a bit of digging and found out this is the normal behavior for published repositories in general. I was able to reproduce the same thing on a yum repository with filtered content. If actual repository copies are made, some empty shell of a repository is always left behind. When orphaned cleanup happens, all of its repository versions get deleted and just the repository is left.

Take a look at https://github.com/Katello/katello/blob/master/app/lib/actions/pulp3/orchestration/repository/delete.rb
Inside you'll see that a repository only gets deleted if `repository.content_view.default?`.
If `repository.environment.nil?`, we only delete the repository version, not the repository itself.

This is explained a bit in here: https://github.com/Katello/katello/pull/8594

It seems that logic came from the fact that, if you deleted the CV version and it deleted the entire repository, it would break other CV versions.

---

The next question is then, what should we do about this bug(?).  The repo shells left behind don't take up any space, but they do take up database entries and potential backend_object_name resources.

We might be able to delete these orphaned repositories if they have no repository versions left. Or, we could improve the repo deletion logic to determine that we are trying to purge this repository and thus delete all cloned repository versions.

I'm inclined to make this a separate issue and deal with it outside the scope of this PR so it's not rushed. Touching our repo deletion logic will need some special attention since it's a bit complicated.

sjha4 (approved):
Ack..let's revisit the orphaned repo issue separately.

This PR looks good and works well..Able to do the usual push, repository and CV operations. 🚢

PR #11031, opened 2024-06-12
Expand description for AKs in registration_commands controller

Lennonka:

Expanded description (doc text) for the `activation_keys` parameter.

* Requested by RH support for improved clarity of the API endpoint/Hammer command.
* Cherry-pick as far as possible, please

Only changes inside a string value. Review from TWs and ack from Devs.

### DIFF
--- a/app/controllers/katello/concerns/api/v2/registration_commands_controller_extensions.rb
+++ b/app/controllers/katello/concerns/api/v2/registration_commands_controller_extensions.rb
@@ -28,7 +28,7 @@ module Katello
         update_api(:create) do
           param :registration_command, Hash do
             param :activation_key, String, desc: N_('Activation key for subscription-manager client, required for CentOS and Red Hat Enterprise Linux. For multiple keys use `activation_keys` param instead.'), deprecated: true
-            param :activation_keys, Array, desc: N_('Activation keys for subscription-manager client, required for CentOS and Red Hat Enterprise Linux. Required only if host group has no activation keys.')
+            param :activation_keys, Array, desc: N_('Activation keys for subscription-manager client, required for CentOS and Red Hat Enterprise Linux. Required only if host group has no activation keys or if you do not provide a host group.')
             param :force, :bool, required: false, desc: N_('Clear any previous registration and run subscription-manager with --force.')
             param :ignore_subman_errors, :bool, required: false, desc: N_('Ignore subscription-manager errors for `subscription-manager register` command')
           end


app/controllers/katello/concerns/api/v2/registration_commands_controller_extensions.rb
```diff
@@ -28,7 +28,7 @@ module ApiPieExtensions
         update_api(:create) do
           param :registration_command, Hash do
             param :activation_key, String, desc: N_('Activation key for subscription-manager client, required for CentOS and Red Hat Enterprise Linux. For multiple keys use `activation_keys` param instead.'), deprecated: true
-            param :activation_keys, Array, desc: N_('Activation keys for subscription-manager client, required for CentOS and Red Hat Enterprise Linux. Required only if host group has no activation keys.')
+            param :activation_keys, Array, desc: N_('Activation keys for subscription-manager client, required for CentOS and Red Hat Enterprise Linux. Required only if host group has no activation keys or if host group is not assigned.')
```
Lennonka, line 31: "or if no host group is given"

Would this be better?

jeremylenz: I'm fine with either :+1:

jeremylenz (commented):
Thanks @Lennonka!

Will wait a bit before acking to see if this gets any more comments.

Lennonka:
@jeremylenz I've invited the TWs the doc channel to review. We usually wait 24 hours for comments, so I'm okay with that.

jeremylenz:
@Lennonka This needs a Redmine issue; can you please create one at projects.theforeman.org ? (Project should be "Katello")

Once that's done, you'll have to edit the commit message as per [this](https://github.com/Katello/katello/pull/11031/checks?check_run_id=26142595861)

app/controllers/katello/concerns/api/v2/registration_commands_controller_extensions.rb
```diff
@@ -28,7 +28,7 @@ module ApiPieExtensions
         update_api(:create) do
           param :registration_command, Hash do
             param :activation_key, String, desc: N_('Activation key for subscription-manager client, required for CentOS and Red Hat Enterprise Linux. For multiple keys use `activation_keys` param instead.'), deprecated: true
-            param :activation_keys, Array, desc: N_('Activation keys for subscription-manager client, required for CentOS and Red Hat Enterprise Linux. Required only if host group has no activation keys.')
+            param :activation_keys, Array, desc: N_('Activation keys for subscription-manager client, required for CentOS and Red Hat Enterprise Linux. Required only if host group has no activation keys or if you do not provide a host group.')
```
asteflova, line 31: I think "required for CentOS and RHEL" and "required only if host group has no activation keys" could be somewhat confusing, if not contradictory, statements. Can you think of an alternative word to replace the second instance of "required"? For example:

```suggestion
            param :activation_keys, Array, desc: N_('Activation keys for subscription-manager client, required for CentOS and Red Hat Enterprise Linux. Use only if host group has no activation keys or if you do not provide a host group.')
```

jeremylenz: There are two instances where `activation_keys` is not required:

1. The host is not CentOS or RHEL; or
2. The host group already provides an activation key.

So maybe 'required' is fine for both? I'm okay with 'use' as well

Lennonka:
Doc-wise peer reviewed and good to go.

jeremylenz (approved):
Thanks @Lennonka !

ACK :+1:

PR #11034, opened 2024-06-14
Fixes #37561 - make sure katello-tracer-upload can run without sudo

sbernhard:
Run cmd for root users without sudo to avoid unnecessary pkg dependency on sudo. sudo is not part of e.g. Debian default installation.


sbernhard:
@jeremylenz, can you have a look. Its related to https://github.com/Katello/katello/pull/10762

jeremylenz (approved):
`[ "$(id -u)" -ne 0 ]` seems like a good way to check if the user is root.

LGTM :+1:

PR #11037, opened 2024-06-14
run Angular tests on GHA

evgeni:


sjha4:
Would it be possible to have the bastion directory packages available for the bastion_katello action? It would be something like npm i bastion -> npm i bastion_katello -> grunt ci

evgeni:
Yeah, that's what was missing.

All seem to be working/failing as expected now :)

sjha4 (approved):
Ack 🎉

PR #11042, opened 2024-06-21
Fixes #37588 - Change UI wording for immediate REX

jeremylenz:

During community demo it was suggested that the options

- via remote execution
- via customized remote execution

are outdated because they come from a time when `katello-agent` was a third option. It was instead suggested to say 'immediately' for the first option.

I changed it in the two places where we have the wording above:

image
image

There are also other places where the wording is slightly different:
image

Should I change the wording in these places too? What should it say? "Install immediately?"

I'm also a bit uneasy because the change to the wording implies that "immediate" isn't remote execution, and "customized" is. Is this a legit concern? If so, what are possible solutions?

I mean, do whatever you want. No functional changes here.

### DIFF
--- a/webpack/components/extensions/HostDetails/Tabs/TracesTab/EnableTracerModal.js
+++ b/webpack/components/extensions/HostDetails/Tabs/TracesTab/EnableTracerModal.js
@@ -31,7 +31,7 @@ const EnableTracerModal = ({
   const [buttonLoading, setButtonLoading] = useState(false);
   const toggleDropdownOpen = () => setIsDropdownOpen(prev => !prev);
   const dropdownOptions = [
-    __('via remote execution'),
+    __('immediately'),
     __('via customized remote execution'),
   ];
   const [selectedOption, setSelectedOption] = useState(dropdownOptions[0]);
--- a/webpack/components/extensions/Hosts/BulkActions/BulkPackagesWizard/04_Review.js
+++ b/webpack/components/extensions/Hosts/BulkActions/BulkPackagesWizard/04_Review.js
@@ -7,7 +7,7 @@ import { CaretDownIcon } from '@patternfly/react-icons';
 import { BulkPackagesWizardContext, INSTALL, UPGRADE_ALL } from './BulkPackagesWizard';
 
 export const dropdownOptions = [
-  __('via remote execution'),
+  __('immediately'),
   __('via customized remote execution'),
 ];
 


bangelic:
@jeremylenz I would just have it say **Install**. Not sure the **immediately** is necessary.

jeremylenz:
> I would just have it say **Install**. Not sure the **immediately** is necessary.

Not sure what you mean?

"Selected packages will be updated on 2 hosts Install" doesn't make sense.

ekohl:
> I'm also a bit uneasy because the change to the wording implies that "immediate" isn't remote execution, and "customized" is. Is this a legit concern? If so, what are possible solutions?

I had the exact same feeling. A drop down with 2 options doesn't make a whole lot of sense to me. Is it better and clear enough to instead use a checkbox "Schedule future execution"?

jeremylenz:
> Is it better and clear enough to instead use a checkbox "Schedule future execution"?

It would, but the REX job wizard lets you customize much more than just execution schedule.

ekohl:
Then perhaps "Customize execution"? Since we can be fancy with React, you could even render 2 different (natural) sentences, depending on which option is selected. Without a drop down at the end of it.

webpack/components/extensions/HostDetails/Tabs/TracesTab/EnableTracerModal.js
```diff
@@ -31,7 +31,7 @@ const EnableTracerModal = ({
   const [buttonLoading, setButtonLoading] = useState(false);
   const toggleDropdownOpen = () => setIsDropdownOpen(prev => !prev);
   const dropdownOptions = [
-    __('via remote execution'),
+    __('immediately'),
     __('via customized remote execution'),
```
Lennonka, line 35:
```suggestion
    __('after you customize remote execution'),
```

Lennonka (changes requested):
How about this?

Can we also change the text on the second picture: "After the prerequisites are met, select how you want to install."

And then immediatelly/after you customize REX

ianballou:
> Then perhaps "Customize execution"? Since we can be fancy with React, you could even render 2 different (natural) sentences, depending on which option is selected. Without a drop down at the end of it.

My ideal would be something like this -- rather than a dropdown, you could have a checkbox with something like "Customize remote execution job".

I realize this will require editing the UI in a bunch of places since the dropdown is used a bunch.

If that seems like too much, I actually prefer `via remote execution` and `via customized remote execution`. They are short and to-the-point for having a dropdown with two selection. You know exactly what you're getting.

ianballou:
There haven't been any comments for a while. I think we should either implement the new button/checkbox/etc here, or close out the PR for now and put the issue back on the backlog.

jeremylenz:
> close out the PR for now and put the issue back on the backlog.

There seems to be no consensus here, so I'm going to close this for now.

PR #11044, opened 2024-06-25
Fixes #37595 - Drop defunct deb mirror_publication_options

quba42:
Ever since we switched to verbatim publications for deb content on smart proxies this code path has been unused. We could have cleaned it up as part of the change in 4a363fd945d56ad48f6a30446d9c16320819d4a5.

Cleans up some dead code. From using git grep, it is clear only deb content has `mirror_publication_options` defined, but ever since https://projects.theforeman.org/issues/35959 deb content no longer uses those publication options, because a verbatim publication is used instead.

One could sync some deb content and some non deb content to smart proxy as a test, but it is pretty clear from the code that this change has no effect.

### DIFF
--- a/app/services/katello/pulp3/repository_mirror.rb
+++ b/app/services/katello/pulp3/repository_mirror.rb
@@ -166,21 +166,13 @@ module Katello
         uri.to_s
       end
 
-      def publication_options(repository_version)
-        popts = {repository_version: repository_version}
-        if (type_specific_options = repo_service.try(:mirror_publication_options))
-          popts.merge!(type_specific_options)
-        end
-        popts
-      end
-
       def create_publication
         if (href = version_href)
           if repo_service.repo.content_type == "deb"
             publication_data = api.publication_verbatim_class.new({repository_version: href})
             api.publications_verbatim_api.create(publication_data)
           else
-            publication_data = api.publication_class.new(publication_options(href))
+            publication_data = api.publication_class.new(repository_version: href)
             api.publications_api.create(publication_data)
           end
         end
--- a/app/services/katello/pulp3/repository/apt.rb
+++ b/app/services/katello/pulp3/repository/apt.rb
@@ -51,16 +51,6 @@ module Katello
           popts
         end
 
-        def mirror_publication_options
-          {
-            # Since we are synchronizing the "default" distribution from the simple publisher on the server,
-            # it will be included in the structured publish. Therefore, we MUST NOT use the simple publisher
-            # on the proxy, since this would collide!
-            #simple: true,
-            structured: true # publish real suites (e.g. 'stable')
-          }
-        end
-
         def distribution_options(path)
           {
             base_path: path,


app/services/katello/pulp3/repository_mirror.rb
```diff
@@ -166,21 +166,13 @@ def remote_feed_url
...
-      end
-
       def create_publication
         if (href = version_href)
           if repo_service.repo.content_type == "deb"
             publication_data = api.publication_verbatim_class.new({repository_version: href})
             api.publications_verbatim_api.create(publication_data)
           else
-            publication_data = api.publication_class.new(publication_options(href))
+            publication_data = api.publication_class.new(repository_version: href)
```
quba42, line 175: This essentially reverts to what we had before 5876a01660ff0c1696b7acf13969e625889931d4.

Manisha15 (approved):

Manisha15:
Looks good to me.

wbclark (commented):
> Cleans up some dead code. From using git grep, it is clear only deb content has `mirror_publication_options` defined, but ever since https://projects.theforeman.org/issues/35959 deb content no longer uses those publication options, because a verbatim publication is used instead.

Hey @quba42 ! It's possible I'm missing some context, but from https://projects.theforeman.org/issues/35959 I found the unmerged draft PR https://github.com/Katello/katello/pull/10420 -- should that go in first?

quba42:
@wbclark This change makes sense independently of the changes in #10420 and is ready now. Once this is merged, I will soon rebase an updated version of #10420 on top of this change. I found this change while working on #10420, but decided to open this PR separately, because it is small and unrelated.

app/services/katello/pulp3/repository_mirror.rb
```diff
@@ -166,21 +166,13 @@ def remote_feed_url
         uri.to_s
       end
 
-      def publication_options(repository_version)
```
wbclark, line 169: Musing: do you see any value in keeping this around as

```ruby
def publication_options(repository_version)
  {repository_version: repository_version}
end
```

for reasons like encapsulation, single-responsibility, readability, extensibility, etc?

In that case, we'd still keep `publication_data = api.publication_class.new(publication_options(href))` below (and could do the same for the verbatim, non-deb case, I suppose)

OR, in your experience is this a case of "You Aren't Gonna Need It"?

I don't have rigid, inflexible opinions about either approach, just curious whether you see any value in it based on your experience with this area of the code, & anticipating how it's likely to evolve in the future.

wbclark (approved):
Thanks for the additional context.

The code looks fine and worked for me in some basic testing. I can see that the apt mirror_publication_options is not reachable via any code path. I didn't see anything in test/services/katello/pulp3/* that would require any update to accommodate this change, and the tests are green.

I have one (non-blocking) observation/idea (see comment below), and I'm honestly fine with merging this either way so I'll go ahead and ACK it. ;-)

app/services/katello/pulp3/repository_mirror.rb
quba42: I am going to go with: "We may never need it, so I lean towards cleaning it up". I had a look at the API docs for several Pulp plugins and there is the odd plugin specific option for Pulp distributions (some plugins don't have distributions), but the evidence suggests Katello never made use of any of those.

PR #11009, opened 2024-05-29
Fixes #37517 - Upgrade theforeman-rubocop gem to the  v0.1.0

archanaserver:
#### What are the changes introduced in this pull request?

- Upgrade theforeman-rubocop gem to version 0.1.0, https://community.theforeman.org/t/standardizing-rubocop-with-theforeman-rubocop/37239/1#use-theforeman-rubocop-010-in-foreman-and-all-plugins-1
- Changed the RuboCop configuration to inherit from the `lenient.yml` style provided by `theforeman-rubocop`, https://github.com/theforeman/theforeman-rubocop?tab=readme-ov-file#not-intrusive-style---lenient
- Removed `Rails` and `Style/Documentation` cops from the configuration as they are already included in the `lenient.yml` style.

#### Considerations taken when implementing this change?

- Chose `lenient.yml` because it aligns well with the existing RuboCop style followed by this repo.
- Removed `Rails` and `Style/Documentation` cops to avoid redundancy and conflicts since they are already covered by `lenient.yml`.

#### What are the testing steps for this pull request?

1. Run `bundle install` to ensure all dependencies are correctly installed, including the updated `theforeman-rubocop` gem.
2. Execute `bundle exec rubocop` to verify that the new RuboCop configuration is being applied without errors.
3. Review the output to ensure there are no unexpected offenses or configuration issues.

PR #11047, opened 2024-06-25
Fix Style/CaseLikeIf cop

archanaserver:
Based on https://github.com/Katello/katello/pull/11009
Addresses the Style/CaseLikeIf rule by converting if-elsif statements to case-when statements

Ensured that all if-elsif statements were appropriately converted to case-when.
Make sure of existing logic and functionality of the code during the conversion.

We can run the test to ensure that no functionality is broken and all tests pass.

### DIFF
--- a/app/services/katello/candlepin/consumer.rb
+++ b/app/services/katello/candlepin/consumer.rb
@@ -136,26 +136,26 @@ module Katello
       def self.distribution_to_puppet_os(name)
         return ::Operatingsystem::REDHAT_ATOMIC_HOST_OS if name == ::Operatingsystem::REDHAT_ATOMIC_HOST_DISTRO_NAME
 
-        name = name.downcase
-        if name =~ /red\s*hat/
+        case name.downcase
+        when /red\s*hat/
           'RedHat'
-        elsif name =~ /centos/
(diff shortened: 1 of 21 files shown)


app/services/katello/candlepin/consumer.rb
```diff
@@ -137,25 +137,26 @@ def self.distribution_to_puppet_os(name)
         return ::Operatingsystem::REDHAT_ATOMIC_HOST_OS if name == ::Operatingsystem::REDHAT_ATOMIC_HOST_DISTRO_NAME
 
         name = name.downcase
-        if name =~ /red\s*hat/
+        case name
```
m-bucher, line 140: We could simplify this to `case name.downcase` now.

m-bucher (approved):
Code looks good to me.

app/services/katello/candlepin/consumer.rb
archanaserver: Good catch. Thanks

archanaserver: @m-bucher i've updated the changes, can you take a look?

ianballou (approved):

PR #11049, opened 2024-06-26
Fixes #37599 - ensure KS repo remains when reregistering host

ianballou:
Makes sure that the KS repo doesn't become `nil` for reregistering hosts that have a KS repo. Otherwise, the Host record will become invalid with a missing Medium.

Do we need to clear any host artifacts when reregistering?

1) Make a host with the kt_activation_keys parameter set to some AK
2) Provision it
3) See that provisioning will fail due to the KS repo being cleared out.

### DIFF
--- a/app/services/katello/registration_manager.rb
+++ b/app/services/katello/registration_manager.rb
@@ -125,9 +125,11 @@ module Katello
       #  * organization_destroy: destroy some data associated with host, but
       #    leave items alone that will be removed later as part of org destroy
       #  * unregistering: unregister the host but don't destroy it
+      #  * keep_kickstart_repository: ensure the KS repo ID is not set to nil
       def unregister_host(host, options = {})
         organization_destroy = options.fetch(:organization_destroy, false)
         unregistering = options.fetch(:unregistering, false)
+        keep_kickstart_repository = options.fetch(:keep_kickstart_repository, false)
 
         # if the first operation fails, just raise the error since there's nothing to clean up yet.
         candlepin_consumer_destroy(host.subscription_facet.uuid) if !organization_destroy && host.subscription_facet.try(:uuid)
@@ -138,7 +140,11 @@ module Katello
         host.subscription_facet.try(:destroy!)
 
         if unregistering
-          remove_host_artifacts(host)
+          if keep_kickstart_repository
+            remove_host_artifacts(host, kickstart_repository_id: host&.content_facet&.kickstart_repository_id)
+          else
+            remove_host_artifacts(host)
+          end
         elsif organization_destroy
           host.content_facet.try(:destroy!)
           remove_host_artifacts(host, clear_content_facet: false)
@@ -152,7 +158,9 @@ module Katello
         new_host = host.new_record?
         unless new_host
           host.save!
-          unregister_host(host, :unregistering => true)
+          # Keep the kickstart repository ID so the host's Medium isn't unset
+          # Important for registering a host during provisioning
+          unregister_host(host, :unregistering => true, :keep_kickstart_repository => true)
           host.reload
         end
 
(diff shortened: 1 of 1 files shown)


app/services/katello/registration_manager.rb
```diff
@@ -128,6 +128,7 @@ def joined_hostnames(hosts)
       def unregister_host(host, options = {})
         organization_destroy = options.fetch(:organization_destroy, false)
         unregistering = options.fetch(:unregistering, false)
+        reregistering = options.fetch(:reregistering, false)
```
jeremylenz, line 131: Would be good to add `reregistering` to the comment-documentation 4 lines above here

app/services/katello/registration_manager.rb
```diff
@@ -139,6 +140,8 @@ def unregister_host(host, options = {})
 
         if unregistering
           remove_host_artifacts(host)
+        elsif reregistering
+          remove_host_artifacts(host, kickstart_repository_id: host.content_facet.kickstart_repository_id)
```
jeremylenz, line 144: can never be too cautious :smile: 

```suggestion
          remove_host_artifacts(host, kickstart_repository_id: host.content_facet&.kickstart_repository_id)
```

app/services/katello/registration_manager.rb
jeremylenz, line 131: Is `reregistering` an accurate name? Has the host been registered before at this point?

app/services/katello/registration_manager.rb
chris1984: Should we place a logger.info or debug incase this happens again we can track down the params and host_artifacts for satellite installs?

chris1984: +1

app/services/katello/registration_manager.rb
ianballou: I suppose I could change it to something like `keep_kickstart_repo`. 

I named it `reregistering` originally because it communicates that the unregister code is running before being expected to register again later.

app/services/katello/registration_manager.rb
```diff
@@ -152,7 +158,9 @@ def register_host(host, consumer_params, content_view_environments, activation_k
         new_host = host.new_record?
         unless new_host
           host.save!
-          unregister_host(host, :unregistering => true)
+          # Keep the kickstart repository ID so the host's Medium isn't unset
+          # Important for registering a host during provisioning
+          unregister_host(host, :reregistering => true, :keep_kickstart_repository => true)
```
jeremylenz, line 163: Shouldn't this be `unregistering`?
```suggestion
          unregister_host(host, :unregistering => true, :keep_kickstart_repository => true)
```

app/services/katello/registration_manager.rb
```diff
@@ -294,14 +302,15 @@ def populate_subscription_facet(host, activation_keys, consumer_params, uuid)
         subscription_facet
       end
 
-      def remove_host_artifacts(host, clear_content_facet: true)
+      def remove_host_artifacts(host, clear_content_facet: true, kickstart_repository_id: nil)
```
chris1984, line 305: Can we get a logger debug statement here with the params, just in case support needs to debug a customer case, since we are clearing a lot of stuff from a host.

jeremylenz:
Looks like there are some related test failures.

app/services/katello/registration_manager.rb
ianballou: Done :+1:

ianballou:
Unfortunately the RHEL 9 packit build seems to have failed to produce any RPMs even though it reported success.

jeremylenz (approved):
Approved pending a final functional test :+1:

ianballou:
Merging - I was able to provision RHEL 8 and RHEL 9 with hosts using the kt_activation_keys parameter.

PR #11050, opened 2024-06-26
Fixes #37598 - Fix missing kwargs on repo enable for ISS

chris1984:
* Fixing kwargs missing on repo enable that we missed in Ruby 3

* Make sure it works on EL8/Ruby2

* Import a manifest
* Sync a repo and create a content view with it
* Export the CV with hammer
* Create a 2nd org
* Import a manifest
* Import the cvv into the new org and see if it fails
* Checkout PR
* Try the import again
* ???
* Profit

### DIFF
--- a/app/lib/actions/katello/repository_set/enable_repository.rb
+++ b/app/lib/actions/katello/repository_set/enable_repository.rb
@@ -6,8 +6,7 @@ module Actions
           _("Enable")
         end
 
-        def plan(product, content, substitutions, override_url: nil,
-                 override_arch: nil)
+        def plan(product, content, substitutions, opts = {})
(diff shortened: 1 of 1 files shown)


vsedmik (approved):
The packit/PRT [passed](https://github.com/SatelliteQE/robottelo/pull/15530#issuecomment-2194077298), but it seems it was running against the default RHEL8 (I need to check with CI folks how to properly specify the target RHEL in PRT comment).

Anyway, for that reason I ran it locally against RHEL9 based SAT and it passed too!
So ACK from me.

**Edit:** So the original packit/PRT was run against RHEL9 too, so double ACK :slightly_smiling_face:

ianballou (approved):
Code looks good, ACK based on @vsedmik 's testing.

PR #11052, opened 2024-06-26
Fixes #37600 - Add params to skip Candlepin content on repository remove

chris1984:
* We currently have 2 params available to skip Candlepin content during repo deletion `skip_environment_update` `destroy_content` these params were not exposed to the user.
* Added `skip_candlepin_environment_update` which maps to `skip_environment_update` and `skip_candlepin_remove_content` which maps to `destroy_content`
* Added unit tests

* SSH to reproducer(IP in Jira)
* Try to delete a repo that causes the error `hammer repository destroy --id 1457 --organization-id 1`
* Check out PR or apply patch file at bottom of PR since it's Katello 4.7
  * `cd /usr/share/gems/gems/katello-4.7.0.36`
  * `patch -p1 < <patchfile>`
  * `foreman-maintain service restart`
* Try to delete a repo that would throw an error with new params `hammer repository destroy --id 1458 --organization-id 1 --skip-candlepin-environment-update true --skip-candlepin-remove-content true`

Patch file:
[katello.patch](https://github.com/user-attachments/files/15994877/katello.patch)

### DIFF
--- a/app/controllers/katello/api/v2/repositories_controller.rb
+++ b/app/controllers/katello/api/v2/repositories_controller.rb
@@ -410,12 +410,16 @@ Alternatively, use the 'force' parameter to regenerate metadata locally. On the 
 
     api :DELETE, "/repositories/:id", N_("Destroy a custom repository")
     param :id, :number, :required => true
+    param :skip_candlepin_environment_update, :bool, :required => false, :desc => N_('Skip updating the candlepin environment section in the destroy task.')
+    param :skip_candlepin_remove_content, :bool, :required => false, :desc => N_('Skip remove content in candlepin section in the destroy task.')
     param :remove_from_content_view_versions, :bool, :required => false, :desc => N_("Force delete the repository by removing it from all content view versions")
     param :delete_empty_repo_filters, :bool, :required => false, :desc => N_("Delete content view filters that have this repository as the last associated repository. Defaults to true. If false, such filters will now apply to all repositories in the content view.")
     def destroy
       sync_task(::Actions::Katello::Repository::Destroy, @repository,
                 remove_from_content_view_versions: ::Foreman::Cast.to_bool(params.fetch(:remove_from_content_view_versions, false)),
-                delete_empty_repo_filters: ::Foreman::Cast.to_bool(params.fetch(:delete_empty_repo_filters, true))
+                delete_empty_repo_filters: ::Foreman::Cast.to_bool(params.fetch(:delete_empty_repo_filters, true)),
+                skip_environment_update: ::Foreman::Cast.to_bool(params.fetch(:skip_candlepin_environment_update, false)),
+                destroy_content: ::Foreman::Cast.to_bool(params.fetch(:skip_candlepin_remove_content, true))
                 )
       respond_for_destroy
     end
--- a/test/controllers/api/v2/repositories_controller_test.rb
+++ b/test/controllers/api/v2/repositories_controller_test.rb
@@ -767,6 +767,26 @@ module Katello
       assert_response :success
     end
 
+    def test_skip_candlepin_environment_update
+      assert_sync_task(::Actions::Katello::Repository::Destroy) do |repo|
+        repo.id == @repository.id
+      end
+
+      delete :destroy, params: { :id => @repository.id, :skip_candlepin_environment_update => true }
+
+      assert_response :success
+    end
+
+    def test_skip_candlepin_remove_content
+      assert_sync_task(::Actions::Katello::Repository::Destroy) do |repo|
+        repo.id == @repository.id
+      end
+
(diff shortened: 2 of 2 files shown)


ianballou (commented):
Seems to work fine, ran the instructions in the PR exactly on the reproducer. Just wanted to check -- 1458 on the reproducer was also an erroring repository?

One question -- how will a user know to run this workflow? I wonder if we can catch the `Unable to find content with the ID "8799".` error and tell the user to try deleting the repo with those parameters.

chris1984:
> Seems to work fine, ran the instructions in the PR exactly on the reproducer. Just wanted to check -- 1458 on the reproducer was also an erroring repository?
> 
> One question -- how will a user know to run this workflow? I wonder if we can catch the `Unable to find content with the ID "8799".` error and tell the user to try deleting the repo with those parameters.

I am not sure the best way, we try to find those items in candlepin and we can't find them just skip those items in the task as well as have the params I exposed if the user knows the except repos #'s. To answer your other questions, yeah any of the codebuilder repos were failing in that reproducer, prob some others but that is what was in the bug.

ianballou:
What would you think about adding a rescue where the missing content error pops up and return a custom error saying something like "Try deleting the repository with blah options"?

ianballou:
Otherwise I suppose we'd need to add an entry to the debugging section in the Foreman docs

chris1984:
> What would you think about adding a rescue where the missing content error pops up and return a custom error saying something like "Try deleting the repository with blah options"?

My worry there is then the task will be half complete and the repos will be out of Pulp/Katello so won't it fail the next time on the first part of the delete? Is there a way to do a dry run first before it deletes anything?

ianballou:
> My worry there is then the task will be half complete and the repos will be out of Pulp/Katello so won't it fail the next time on the first part of the delete? Is there a way to do a dry run first before it deletes anything?
Yeah, it's too bad that the  `Unable to find content with the ID "8799"` error is likely the only way a user will know to try deletion with the new options.

If that's unavoidable, I could see a user hitting the error, having one messed up repo, and then going to some document teaching them how to find the rest of the broken repositories and then fixing them following the guide.

I can't remember, do we have a rake task that'll find these repos that are broken and need deleting? If so, I suppose the rake task could teach the user of the new options.

ianballou:
(Also, the PR needs a rebase for the tests to run)

ianballou:
> we can try to find those items in candlepin and we can't find them just skip those items in the task

I missed this before, I think this totally makes sense so the user wouldn't even have to think about it. Maybe just a warning log to say that the content deletion was skipped.

Ron-Lavi:
@chris1984 can you rebase? @ianballou mind having another look? thanks

chris1984:
@Ron-Lavi this approach does not work, Jeremy tested it a while ago. It's not sending the params correctly when you get to the last page, so I gave up on it hence why it's so old.

chris1984:
Ian is helping me get it working, but Sayan said it might be safe to push it to 6.17. I'll keep trying to make it work

sayan3296:
Based on a few of the recent occurrences of this issue, I believe I would be happy with the implementation if it would be something like this :

* There are some orphaned repos present and some good repos present and part of the same CV version [ lifecycle ]
* User started deleting one or more good + orphaned repos
* Some of their content ids are missing but Katello only logs a warning about it and skips the SetContent step
* Completes the deletion successfully without any manual intervention needed 

As long as this would be the workflow for the deletion of a repo, I am happy to have the fix delivered as soon as possible. 

While the issue also could happen during repo creation, 90% of the cases encountered are related to deletion only. So for now, let's fix the deletion part and make it as smooth as we can. 

@chris1984 @ianballou jfyi ^^

chris1984:
> Based on a few of the recent occurrences of this issue, I believe I would be happy with the implementation if it would be something like this :
> 
> * There are some orphaned repos present and some good repos present and part of the same CV version [ lifecycle ]
> * User started deleting one or more good + orphaned repos
> * Some of their content ids are missing but Katello only logs a warning about it and skips the SetContent step
> * Completes the deletion successfully without any manual intervention needed
> 
> As long as this would be the workflow for the deletion of a repo, I am happy to have the fix delivered as soon as possible.
> 
> While the issue also could happen during repo creation, 90% of the cases encountered are related to deletion only. So for now, let's fix the deletion part and make it as smooth as we can.
> 
> @chris1984 @ianballou jfyi ^^

That works for me, I am wrapping up 2 6.16 bugs, then will get back to this one so we can get this possibly into a 6.16.z

ianballou:
@chris1984 I suppose we can close this out now with the other PR making it in.

sayan3296:
Is https://github.com/Katello/katello/commit/fae847127ac78a38e0bbf2da98d804add0674746 the one to consider as the alternate fix?

PR #11053, opened 2024-06-26
Fixes #37560 - rewrite React GH action

jeremylenz:

Change the 'React Tests' GitHub action to use @theforeman/actions in line with Foreman CI.

This will also allow us to use `foreman_version` to test Katello PRs against any Foreman PR or branch.

Also removed '(Non-blocking)' from the name, so it's assumed this is a blocking check. But I think we need a GH org admin to make that actually true.

critique my beautiful YAML
verify that CI passes and that the output reflects actual test runs

### DIFF
--- a/.github/workflows/react_tests.yml
+++ b/.github/workflows/react_tests.yml
@@ -1,45 +1,16 @@
-name: React Tests (non-blocking)
-
+name: React Tests
 on: [pull_request]
 
-defaults:
-  run:
-    working-directory: ./katello
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.ref_name }}-${{ github.workflow }}
+  cancel-in-progress: true
 
 jobs:
-  build:
-
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v2
-      with:
-        path: ${{ github.workspace }}/katello
-    - name: Use Node.js
-      uses: actions/setup-node@v1
-      with:
-        node-version: '14'
-    # We could update the postinstall action for foreman to look for an environment variable for plugin webpack dirs
-    # before kicking off the ruby script to find them, this would eliminate the ruby dep and running `npm i` in Katello.
-    - uses: ruby/setup-ruby@v1
-      with:
-        ruby-version: '2.7'
-    - name: Checkout Foreman repo
-      uses: actions/checkout@v2
-      with:
-        repository: theforeman/foreman
-        path: ${{ github.workspace }}/foreman # checkout Foreman and Katello as siblings, same as dev env
-        ref: develop
-    - name: Generate Foreman dependencies package-lock
-      run:  npm i --package-lock-only --no-audit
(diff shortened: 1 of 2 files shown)


.github/matrix.json
ekohl, line 1: Please don't include this. The whole design we've applied is that we take the Foreman core repository to define these. That way you can simply point to Foreman x.y and get a test matrix.

ekohl (changes requested):

.github/workflows/react_tests.yml
```diff
@@ -1,45 +1,23 @@
...
-    - name: Install Katello dependencies
-      run:  npm ci --no-audit
-    - name: Run Katello linting
-      run: npm run lint
-    - name: Run Katello tests
-      run: npm test
+    uses: theforeman/actions/.github/workflows/foreman_plugin_js.yml@v0
+    with:
+      plugin: katello
+      foreman_version: develop # set to the Foreman release branch after branching :)
```
ekohl, line 23: This is good, but we should also include it by default in `.github/workflows/ruby.yml` so we can simply update it on branching (instead of relying on adding it in branching).

.github/matrix.json
jeremylenz: Removed (was a leftover from a previous iteration).

jeremylenz:
```
build Expected — Waiting for status to be reported
```

I'm not sure what's wrong here.

ekohl:
> I'm not sure what's wrong here.

There's probably a required check configured for this branch that now no longer exists. Some admin needs to fix that.

ekohl (approved):

.github/workflows/react_tests.yml
```diff
@@ -1,45 +1,23 @@
-name: React Tests (non-blocking)
```
evgeni, line 1: This, albeit the name, used to be a required check in GH.
After the rename, it's not matching anymore, so you get the "waiting" job.

I don't think this *needs* to be a "required check", but if you want it to be, it needs to run on every PR, not limited to `paths` like below.

jeremylenz: Ahhh, I thought it was my rename of `build` to `react-tests`. That explains why changing it back to `build` didn't work.

@parthaa I would like to call this "React Tests" (without the "(non-blocking)"). Can you please make "React Tests" the required check and remove "React Tests (non-blocking)" ?

ekohl: @jeremylenz you probably need to merge this first before you can add it as a required check.

jeremylenz:
I had Partha remove the `build` required check and add a new required check for `react-tests`. I renamed the job here to `react-tests` and also removed the `paths` so it now runs on all pull requests. Let's see how it goes..

jeremylenz:
Looks like we might still have to merge this before the expected check is satisfied..

ekohl:
> Looks like we might still have to merge this before the expected check is satisfied..

That's also my experience

PR #11054, opened 2024-06-27
Fixes #37603 - Point API users to the correct content upload endpoint

sjha4:
Update API desc to point to correct endpoint for uploading content.
API description should read this:
Screenshot from 2024-06-27 09-23-29


wbclark (approved):

wbclark:
The redmine reads like it originates from RH bugzilla or JIRA, but I don't see a link to the original issue.

Was it manually created, or do we need to investigate the automation there and make sure it's linking properly?

sjha4:
I created that issue manually from jira. Updated issue to link the 2. Thanks for catching that.Merging.

PR #11056, opened 2024-06-28
Fixes #37609 - Migrate sha1 repos only at the next edit/sync time

ianballou:
Deletes the recent migration from https://github.com/Katello/katello/pull/11013 and makes it more clear how repositories using a sha1 'yum metadata checksum type' will migrate off of sha1 to the default type.

sha1 using repositories will lose the sha1 checksum type when a repository is updated and when it has its contents changed (essentially anything that causes a new publication to get created). This includes syncing, uploading content, and deleting content.

The repository edit UI is acting a little strange if you have a sha1 repository. The selector will correctly say "sha1", but when you click to edit it, there will be a totally blank option available for selection. It will error out, but the fact that it's available is a bit confusing. However, the fact that the sha1 disappears as you click on it does communicate that it is no longer available. There is also new help text below thought to clear any confusions.

Another consideration was actually deleting the migration. Due to the timing of things, the migration will make it into Katello 4.13.0. Some users will thus experience the original sha1 migration. They shouldn't hit any issues, however until their repositories regenerate a publication, the Publication will use sha1 while the UI reports "Default". We can explain this in the release post on Discourse.  cc @parthaa 

1) Create a repository and hack it to sha1 with:
```ruby
rooty = ::Katello::Repository.find(<id>).root
rooty.checksum_type = 'sha1'
rooty.save(validate: false)
```
2) Ensure that the edit UI shows the correct help text.
3) Ensure that the sha1 option disappears when selecting another option.
4) Ensure that you can select another option
5) Ensure that the sha1 is changed to default when:
  a) Editing any other field on the repo,
  b) Uploading content to the repo,
  c) Deleting content from the repo,
  d) Syncing the repo with actual changes


m-bucher (approved):
Works well, as far as I can say.

Tested with:
yum-repo:
1. showed warning
1. showed `sha1`, disappeared in edit-mode (as predicted), which is OK for me.
1. modified to default after changing any value (tested by changing 'Verify SSL')

deb-repo: improbable that this may have 'sha1' in `checksum_type` as the field is not used by deb-repos. Nevertheless, the fix correctly removed the 'sha1' value and I saw no negative impact there.

chris1984 (approved):
ACK, code looks fine and acking based on ATIX testing

PR #11058, opened 2024-07-02
Fixes #35959 - Use structured APT for deb content

quba42:
RFC: https://community.theforeman.org/t/rfc-transitioning-katello-to-structured-apt-deb-content/34262
Issue: https://projects.theforeman.org/issues/35959
Replaces: https://github.com/Katello/katello/pull/10420
Requires: https://github.com/candlepin/subscription-manager/pull/3223

Note that the subscription-manager packages for Debian and Ubuntu published on http://oss.atix.de/ are already patched to support these changes.


db/migrate/20240508100434_add_content_id_to_katello_repository.rb
```diff
@@ -0,0 +1,5 @@
+class AddContentIdToKatelloRepository < ActiveRecord::Migration[6.1]
+  def change
+    add_column :katello_repositories, :content_id, :string
```
quba42, line 3: The core requirement for the feature is for deb type content to have a different candlepin content associated with each katello_repositories record. This way different APT repo distributions/components can be communicated to registered clients, depending on what katello repository they are currently consuming.

app/models/katello/repository.rb
```diff
@@ -207,6 +206,16 @@ def self.repo_path_from_content_path(environment, content_path)
       "#{path_prefix}/#{path}"
     end
 
+    def content_id
+      # Currently deb content will store a content_id on each Reposiotry, while all other content
+      # types will store one on the RootRepository.
+      self[:content_id] || root.content_id
+    end
```
quba42, line 213: The basic idea here is that all content types will now look for the content_id on `Katello::Repository` instead of `Katello::RootRepository`. However, for all non-deb content types, the content_id on a `Katello::Repository` will always be `nil`, and will default right back to `root.content_id`. That way, things don't actually change much for non-deb content types. They will continue to store and use a single `content_id` per root repository.

quba42 (commented):
This "review" (and the below comments) aims to explain the design approach for this PR.

Regarding the current state of this PR:
- I have tested the new "structured APT" functionality as follows:
  - Created an APT repo, synced it, and added it to a CV version and LCENV. Registered a host and checked the APT repo components and distirbutions (structure information) make it all the way into the repo file on the registered host, as appropriate for the environment it is in. This works.
- For deb content, the current state does not yet update the corresponding KatelloContent, whenever some candlepin content is updated. I have not found this to impact the functionality, but I will continue to work on this for the sake of consistency and to avoid problems down the road.

In my local dev environment I currently see two test failures relative to master branch:
```
ktest test/actions/pulp3/orchestration/export_test.rb
ktest test/lib/tasks/pulpcore/repository_vcr_test.rb
```
The failing VCR test does not turn green from re-recording the cassette. It complains about a call to `POST https://localhost:23443/candlepin/environments/1/content`, but only records calls to Pulp.

The other test appears to attempt calling a pulp_deb API endpoint from a Yum API object, which does not make any sense. I was so far unable to figure out why this is, but I think it is likely an issue with the test rather than the code (not 100% sure though).

# What I am hoping for at this point:

It would be great to receive some initial feedback if the design approach for this feature is sound. In particular, an initial review should focus on potential risks to non-deb content types. The idea is that from the point of view of yum and other non-deb types, this is a pure refactor, that does not change any outcomes! Thanks in advance for any and all input!

quba42:
The latest push adds the following fixes:

- Fixed the `test/actions/pulp3/orchestration/export_test.rb` test (by fixing the code).
- Consolidated the several settings to enable this feature into a single setting.
- Fixed two rubocop errrors.

Remaining ToDo's:

- Update the rake task to migrate existing content to structured mode.
- There is still a failing test in `ktest test/lib/tasks/pulpcore/repository_vcr_test.rb`.
- Manually test more deb content edge cases.
- Add new test coverage for the new deb features.

ianballou:
> There is still a failing test in `ktest test/lib/tasks/pulpcore/repository_vcr_test.rb`.

This is the reference used for rerecording Candlepin VCR tests:
```
Running the glue tests: record=true mode=all ktest ~/katello/test/scenarios/scenario_test.rb (save some time by doing other testing concurrently)
```

I noticed that test uses scenario_support.rb, which has its own record definition:
```ruby
  def record(name, options = {})
    VCR.insert_cassette("scenarios/#{name}", options)
    User.current = user
    yield
  rescue
    raise
  ensure
    VCR.eject_cassette
  end
```

This is kinda tricky because that test may need to do both candlepin and Pulp recordings at the same time.

Another idea would be to stub out the live Candlepin bits.

app/lib/actions/candlepin/product/content_create.rb
```diff
@@ -10,13 +10,22 @@ class ContentCreate < Candlepin::Abstract
...
           param :owner
           param :os_versions
+          param :repository_id
         end
 
         def run
+          content_url = input[:content_url]
+          # We must retrieve the repository in the run phase, so the latest Pulp version_href is
+          # already set. This is needed to retrieve the latest repository.deb_content_url_options!
+          repository = ::Katello::Repository.find(input[:repository_id])
```
ianballou, line 20: Can this query go inside the `if` statement below? Just to avoid any unnecessary lookups f or non deb repositories.

ianballou (commented):
Code structure is generally looking fine to me, just one small comment. I need to give it a test now.

ianballou:
Should this PR still be in draft form?

app/lib/actions/candlepin/product/content_create.rb
quba42: If the `:type` parameter tells me whether we have deb content or not, then I suppose it can.

quba42:
> Should this PR still be in draft form?

I just came back from PTO/am continuing on this PR now, currently there are three things I am waiting on to undraft:

- The one failing test I have not yet figured out
- The enable/disable rake task I have yet to update
- It looks like there is a bug in the implementation where activation keys are no longer showing any "repository sets" for deb content.

app/lib/actions/candlepin/product/content_create.rb
ianballou: Ah you have to look up the repo to tell the type, I see. Up to you if you want to refactor it so the repo type gets passed in as a parameter (unless it does already).

quba42:
Update: I rebased to latest master and re-recorded VCR cassettes (because of conflicts).

For me all tests are now locally green.

I have two remaining TODOs before un-drafting:

- The "repository sets" view on activation keys for deb content is broken because of a broken query.
- The enable/disable rake task needs to be reworked.

quba42:
Update: "repository sets" query for deb content (with structured APT enabled) is partially fixed, but needs to be refined further.

quba42:
Update: To my best knowledge the feature itself is now ready for final/full review.

I am currently reworking the enable/disable rake task.

ianballou:
I haven't been finding time lately to test this --  @quba42  if someone is able provide a testing report with steps taken from your team, it'd make it easier for us to get more eyes on it.

quba42:
Update: Unfortunately I found that handling of the Library environment is currently broken. (Non-Library environments work fine). I am working on fixing this. I will also add a testing report what has been tried already.

quba42:
Update: I have fixed the handling of Library content view environments. I have also fixed the enable/disable rake task to the point where it appears to be working in both directions, at least for my relatively simple initial test case (more testing is needed).

This means I now have a state that is to the best of my knowledge both complete and working. I will now rebase to latest master branch and re-record vcr_cassettes. Then I will undraft the PR.

lib/katello/tasks/enable_structured_content_for_deb.rake
```diff
@@ -0,0 +1,38 @@
+namespace :katello do
+  desc "Enable or disable the use of structured content for APT clients."
```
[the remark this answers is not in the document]
quba42: The rake task is now working in both directions (though it could use more extensive testing).

app/lib/actions/candlepin/product/content_create.rb
quba42: Now that the basic functionality is complete I should have another look at implementing this suggestion. I will probably get to it early next week.

app/models/katello/repository.rb
```diff
@@ -1022,6 +1035,35 @@ def in_content_view?(content_view)
       content_view.repositories.include? self
     end
 
+    def deb_content_url_options
+      if self.version_href
+        deb_content_url_options = "/?comp=#{self.deb_pulp_components.join(",")}&rel=#{self.deb_pulp_distributions.join(",")}"
+      else
+        deb_content_url_options = ''
+      end
+      return deb_content_url_options
```
sbernhard, line 1044:
```suggestion
      return '' unless version_href

      components = deb_pulp_components.join(',')
      distributions = deb_pulp_distributions.join(',')
      "/?comp=#{components}&rel=#{distributions}"
```

app/models/katello/repository.rb
```diff
@@ -1022,6 +1035,35 @@ def in_content_view?(content_view)
...
+    end
+
+    def deb_pulp_components(version_href = self.version_href)
+      if version_href.blank?
+        components = []
+      else
+        pulp_api = Katello::Pulp3::Repository.instance_for_type(self, SmartProxy.pulp_primary).api.content_release_components_api
+        components = pulp_api.list({:repository_version => self.version_href}).results.map { |x| x.component }.uniq
+      end
+      return components
```
sbernhard, line 1054:
```suggestion
      return [] if version_href.blank?

      pulp_api = Katello::Pulp3::Repository.instance_for_type(self, SmartProxy.pulp_primary).api.content_release_components_api
      pulp_api.list({:repository_version => version_href}).results.map { |x| x.component }.uniq
```

app/models/katello/repository.rb
```diff
@@ -1022,6 +1035,35 @@ def in_content_view?(content_view)
...
+    end
+
+    def deb_pulp_distributions(version_href = self.version_href)
+      if version_href.blank?
+        distributions = []
+      else
+        pulp_api = Katello::Pulp3::Repository.instance_for_type(self, SmartProxy.pulp_primary).api.content_release_components_api
+        distributions = pulp_api.list({:repository_version => self.version_href}).results.map { |x| x.distribution }.uniq
+      end
+      return distributions
```
sbernhard, line 1064:
```suggestion
      return [] if version_href.blank?
      pulp_api = Katello::Pulp3::Repository.instance_for_type(self, SmartProxy.pulp_primary).api.content_release_components_api
      pulp_api.list({:repository_version => version_href}).results.map { |x| x.distribution }.uniq
```

app/services/katello/pulp3/repository/apt.rb
```diff
@@ -31,22 +39,20 @@ def sync_url_params(sync_options)
...
-          super.merge(
-            {
-              distributions: repo.deb_releases + "#{' default' unless repo.deb_releases.include? 'default'}"
-            }
-          )
+          if Setting['deb_enable_structured_apt']
+            distributions = repo.deb_pulp_distributions.join(' ')
+          else
+            distributions = 'default'
+          end
```
sbernhard, line 46:
```suggestion
          distributions = if Setting['deb_enable_structured_apt']
                            repo.deb_pulp_distributions.join(' ')
                         else
                            'default'
                         end
```

app/services/katello/pulp3/repository/apt.rb
```diff
@@ -31,22 +39,20 @@ def sync_url_params(sync_options)
...
           ss = api.signing_services_api.list(name: SIGNING_SERVICE_NAME).results
           popts = super(repository_version)
-          popts.merge!(
-            {
-              structured: true, # publish real suites (e.g. 'stable')
-              simple: true # publish all into 'default'-suite
-            }
-          )
+          popts.merge!({ structured: true })
+          popts.merge!({ simple: true }) unless Setting['deb_enable_structured_apt']
```
sbernhard, line 55: So we want to publish structured always and simple only if not enable_structured_apt?
Wasn't there an issue that we should only publish either simple or structured?

quba42: We had an issue with publishing both, that led to us internally switching to publishing only one or the other as a hotfix, but that work around never made it into upstream. Eventually we fixed the underlying issue, so the workaround was no longer strictly necessary. The way it is now in this PR does not change the upstream behavior for when structured APT is disabled.

app/services/katello/pulp3/repository/apt.rb
quba42: Applied the suggestion locally.

app/lib/actions/candlepin/product/content_create.rb
quba42: Finally got around to implementing this suggestion.

quba42:
Update: The current version of this PR has now been back-ported to our internal Katello 4.13 build. This means that going forward all internal ATIX testing will be using these changes. It also means we can enable the feature on large internal installations to test under real world conditions. To date I have been testing a lot of "targeted edge cases", but all within Forklift.

We are not currently aware of any issues affecting the current state that are prohibitive for merging. We are aware of several possible enhancements that we would like to defer to follow up tasks.

Since the feature lives behind a feature flag and needs to be enabled before anything changes, we have some flexibility on how to organize roll out. For example, we could document it as "experimental" for the first one or two Katello versions it lands in. I will then add a follow up PR that will make transitioning deb content to structured APT mandatory for a Katello version of our choice. For any content types that are not deb, as well as for deb until a user enables the feature, these changes act as a pure refactor, that dose not change any outcomes. As a result we have plenty of time for refinement after merging the initial PR. We are still relatively close to the Katello 4.14 branching which I would consider a good time to merge large new features. :wink: 

@ianballou I am going to re-request a review from you. I aim to post a detailed list of what we have tested ASAP. I will keep updating how our internal testing is progressing. If there is anything else you would like to have from our end, or if you would like to go over details with me via video conference, please tell! Both @m-bucher and @sbernhard have been involved in internal review and implementation to different degrees. Our internal review of this initial implementation is essentially complete.

quba42:
@sjha4 So I just heard that @ianballou is on holidays, and that you might be a good secondary contact. Hence I am requesting a review from you. :wink: The same thing I said to Ian in my previous comment obviously applies to you as well: I will add a list of what I tested ASAP, and if there is anything you want from us (including a meeting by video conference) please do tell!

quba42:
Update: The latest state adds a method to expose the value of the `deb_enable_structured_apt` for use in templates in safe mode. This is useful for Ubuntu/Debian host deployment templates, that need to configure a client repo from the Katello server on the deployed host differently depending on whether the feature is enabled or disabled.

The tests ran successfully in my local forklift, and the test failures on this PR look unrelated to my changes to me.

I have also opened a new PR for the accompanying subscription-manager change: https://github.com/candlepin/subscription-manager/pull/3454

This is so I can take ownership of both PRs, so as to better pursue their joint finalization.  :stuck_out_tongue_winking_eye:

quba42:
Rebased to resolve merge conflict.

sjha4:
Looking at the changes, I remember some discussions around the refactoring in the PR. I haven't had a chance to look at this closely but a lot of it is around candlepin integration. I'll poke some folks more familiar with the files to take a look at those specifically : @parthaa @jeremylenz 👀

lib/katello/plugin.rb
```diff
@@ -668,6 +668,13 @@ def katello_template_setting_values(name)
         default: false,
         full_name: N_('Hide Reclaim Space Warning'),
         description: N_('If this is enabled, the Smart Proxy page will suppress the warning message about reclaiming space.')
+
+      setting 'deb_enable_structured_apt',
```
jeremylenz, line 672: Since Satellite doesn't support deb, should probably add this to https://github.com/RedHatSatellite/foreman_theme_satellite/blob/develop/app/models/concerns/upstream_only_settings.rb once this is merged.

lib/katello/plugin.rb
```diff
@@ -668,6 +668,13 @@ def katello_template_setting_values(name)
...
+        type: :boolean,
+        default: false,
+        full_name: N_('Enable structured APT for deb content'),
+        description: N_("APT repos in katello will have the same repo structure as the remote repos they are syncronized from. " \
```
jeremylenz, line 676:
```suggestion
        description: N_("APT repos in Katello will have the same repo structure as the remote repos they are synchronized from. " \
```

lib/katello/plugin.rb
```diff
@@ -668,6 +668,13 @@ def katello_template_setting_values(name)
...
+                        "This feature must be enabled/disabled by running 'foreman-rake katello:enable_structured_content_for_deb' in order to migrate existing deb content.")
```
jeremylenz, line 677: If it must be enabled via the rake task, why also have a Setting?

quba42: To record somewhere if it is enabled, and also for users to be able to check somewhere if it is enabled. But if there is a better place for this (one where users cannot inadvertently change the setting without running the rake task), that would be better still. It may well be down to my lack of knowledge about suitable alternatives that I chose a setting.

lib/katello/plugin.rb
quba42: Should I open an PR for that? Note also that the setting is not intended to be kept for ever. Eventually, there will come a Katello release where all users are migrated to structured APT during upgrade. After that the setting can  be dropped from the code.

lib/katello/plugin.rb
quba42: Wups. Will fix.

lib/katello/plugin.rb
jeremylenz: > if there is a better place for this (one where users cannot inadvertently change the setting without running the rake task), that would be better still.

The more I think about it the more I feel like a Setting isn't the right way to express this. Settings are for users to set, and they should take effect when you set them. So having a Setting here that does almost nothing just feels weird.

As I look at the rake task, it looks to me a bit like a database migration -- not exactly, since schema is not changing, but rather, data itself is being migrated. I think having a rake task to do this makes sense, but there should be a way other than a Setting to tell if it's been done or not. Ideally it would be something that could be some sort of computed value -- say, something like, `Katello::RootRepository.deb.any? { |root| root.content_id.present? }` If that's not reliable enough, perhaps we could add a boolean database field to RootRepository or something.

As for how to display it to the user, since it's on Repository maybe it would make sense to show something on the repository edit page. @quba42 @sjha4 thoughts?

quba42: What you are saying definitely makes sense, I was never fully happy with the "setting that must not be altered via the UI"  aspect. However, having just done a round of rake task testing on a large installation, which ran into problems with "re-run robustness" (i.e. the rake task does not run to completion on the first attempt for some reason, and then needs to be re-run, but only for the remaining not yet migrated data), I would like to tweak your suggestion as follows:

- There is a computed value for each deb type repository that tells us if that repostiory is using "structured APT". (This does not require many changes, since `root.content_id.blank?` already answers the exact question we are asking here.)
- I keep the setting, but it is now interpreted as answering the question: should _newly created deb repos_ use "structured APT". (Users may freely toggle this setting in the UI).
- The rake task may be used to migrate any existing repositories that are in the other state from the setting, to match the current state of the setting.
- We could display the current state for each repo "structured APT enabled" vs. "structured APT disabled" on the repository page for each repo. (Not sure how important this is).

This approach decouples the question of "what state is a given deb repo in?" from the question of "would you like me to create new deb repos using the new feature?". This also means a partially migrated system (some repos have been migrated, but others are not) is no longer in a broken state (as is currently the case). This greatly lowers the stakes for the migration rake task. The worst thing the rake task can now possibly do is leave an individual repository in an inconsistent state, but it can't leave your entire DB in a globally inconsistent state that may be hard to recover from. It should be much easier to make the rake task robust for these "lower stakes".

jeremylenz: Ack to this modification. That makes much more sense to me. :+1:

lib/katello/plugin.rb
quba42: Done.

lib/katello/plugin.rb
quba42: The suggestion is now implemented.

quba42:
Update: Implemented the suggestion by @jeremylenz. Also fixed several issues affecting the repository sets tab on activation keys when the feature is in use for deb content (credit to @m-bucher).

It's now off to another round of internal testing at ATIX. If internal testing does not reveal any major issues, I will ask for a final review by an ATIX colleague. Once I get one ATIX ACK, I will ask for final ACK from upstream.

quba42:
Update: Pushed a minor cosmetic change. :wink:

app/models/katello/repository.rb
```diff
@@ -211,6 +210,16 @@ def self.repo_path_from_content_path(environment, content_path)
       "#{path_prefix}/#{path}"
     end
 
+    def content_id
+      # Currently deb content will store a content_id on each Reposiotry, while all other content
```
jeremylenz, line 214:
```suggestion
      # Currently deb content will store a content_id on each Repository, while all other content
```

lib/katello/tasks/migrate_structure_content_for_deb.rake
```diff
@@ -0,0 +1,105 @@
+namespace :katello do
+  desc "Will migrate all APT repo content, to match the deb_enable_structured_apt setting."
```
jeremylenz, line 2:
```suggestion
  desc "Migrate all APT repo content to match the deb_enable_structured_apt setting."
```

lib/katello/tasks/migrate_structure_content_for_deb.rake
```diff
@@ -0,0 +1,105 @@
...
+  task :migrate_structure_content_for_deb => ['environment', 'dynflow:client', "check_ping"] do
+    User.current = User.anonymous_api_admin # Set a user
+
+    deb_enable_structured_apt = Setting['deb_enable_structured_apt']
+    puts "'Enable structured APT for deb content' is currently set to '#{deb_enable_structured_apt}'!"
+    if deb_enable_structured_apt
+      puts "Migrating deb type repos, that are not yet using structured APT to start using the feature."
```
jeremylenz, line 9:
```suggestion
      puts "Enabling structured APT for existing deb type repos"
```

lib/katello/tasks/migrate_structure_content_for_deb.rake
```diff
@@ -0,0 +1,105 @@
...
+    else
+      puts "Migrating deb type repos currently using structured APT to disable the feature."
```
jeremylenz, line 11:
```suggestion
      puts "Disabling structured APT for existing deb type repos"
```

app/lib/katello/concerns/base_template_scope_extensions.rb
```diff
@@ -59,6 +59,14 @@ def host_content_facet(host)
         host.content_facet
       end
 
+      apipie :method, "Returns the value of Setting['deb_enable_structured_apt']" do
+        desc 'Structured APT is a feature for the deb content type that may be disabled'
+        returns one_of: [true, false]
+      end
+      def deb_using_structured_apt?
+        Setting['deb_enable_structured_apt']
+      end
```
jeremylenz, line 68: Is this really needed? I'd think you can just use the existing Settings API [here](https://github.com/theforeman/foreman/blob/develop/app/controllers/api/v2/settings_controller.rb), which enables `hammer host settings info --name deb_enable_structured_apt`

jeremylenz (commented):
Some minor nitpicks below.

 @parthaa Can you take a look specifically at the changes to Dynflow actions and see if they look good to you?

app/lib/katello/concerns/base_template_scope_extensions.rb
quba42: I will remove this again. We use it on our downstream branch so that a provisioning template we deliver can check the setting. However, the template in question does not exist in upstream Katello, and anyone who writes their own template can just adjust it after enabling/disabling "structured APT", so I don't think there is an upstream use case for this.

lib/katello/plugin.rb
quba42: Dropped.

app/lib/katello/concerns/base_template_scope_extensions.rb
quba42: Dropped.

lib/katello/tasks/migrate_structure_content_for_deb.rake
quba42: Fixed.

lib/katello/tasks/migrate_structure_content_for_deb.rake
quba42: Fixed.

app/models/katello/repository.rb
quba42: Fixed.

quba42:
Incomplete list of things I have tested (split up into several parts):

### Testing the change with the feature disabled

The point of these tests is to show that this change is a pure refactor for content types other than deb (as well as for deb with the feature disabled).

For some time now, all our internal builds have been using this change with the feature disabled. This includes large production like systems that were upgraded to this state. This means everything that was done on these systems has tested the "disabled" case. Some things that are explicitly worth mentioning:

- Sync, Publish, and Promote for various content types including deb and yum for both CVs and CCVs.
- Export/Import of yum and deb content.
- Backup and restore of yum and deb content.
- Host deployments and host registration, installing content on hosts.

For all of these reasons, I am very confident the refactor part is just that.

### Testing the enable/disable rake task

- Tested the rake task in both directions in forklift using `bundle exec rake katello:migrate_structure_content_for_deb` (the direction is determined by the UI setting).
- Tested creating some deb content (dsiabled), then flipping the setting to enabled, creating more deb content (enabled), then migrating just the disabled content using the rake task.
- Tested re-running the rake task with nothing left to migrate in both directions (prints a message and does nothing further).
- Tested the rake task on a large production like installation with a large amount of deb content. (More than 40 root repos of various sizes published into various CVs, CCVs, and promoted to various environments).

### Testing of the feature itself (deb content with the feature enabled)

- Tested Sync, Publish, and Promote for deb content with multiple components and releases/distributions.
- Tested re-sync after dropping a component or release/distribution, and promoting that.
- Tested registering a host using an activation key.
- Tested overriding repos on an activation key to disabled/enabled.
- Tested changing the content source for a content host.

### Detailed description of a front to back test case.

- Start a forklift without any deb content on it.
- Enable the setting in the UI.
- Create product `P Ubuntu 22.04 jammy`
  - Create repository "R Ubuntu 22.04 jammy client"
    - Upstream URL: `http://oss.atix.de/Ubuntu22LTS/`
    - Releases/Distributions: `stable`
    - Components: `main`
    - Architectures: `amd64`
  - Sync
- Create product `P_structured_apt_tests`
  - Create repository `R_deb_empty`
  - Create repository `R_deb_upload`
    - Get example package: `wget ftp.de.debian.org/debian/pool/main/v/vim/vim_9.0.1378-2_amd64.deb`
    - Upload it
  - Create repository `R_deb_multi_dist`
    - Upstream URL: `https://fixtures.pulpproject.org/debian/`
    - Releases/Distributions: `ginnungagap ragnarok`
    - Components: `asgard jotunheimr`
    - Sync it.
- Create LCENV DEV
- Create CV `CV_structured_apt_tests`
  - Add all four test repos
  - Publish and promote new version
- Create activation key `AK Ubuntu 22.04 jammy DEV`
- Register an Ubuntu host using the key and the instructions from here: http://oss.atix.de/html/ubuntu.html
- Check the `/etc/apt/sources.list.d/rhsm.sources`
  - Check the `Suites` and `Components` fields. They should contain things like `ginnungagap ragnarok` and not just `default` or `all`.
- Now go back to the repo page for `R_deb_multi_dist`
  - remove `ginnungagap` and `asgard`
  - re-sync
- Verify that nothing changes on your registered host from the sync alone (after `subscription-manager repos`)
- Publish and Prmote a new CV version
- Verify that your host no longer has `ginnungagap` and `asgard` configured (after `subscription-manager repos`)

app/models/katello/host/content_facet.rb
```diff
@@ -164,10 +164,12 @@ def update_repositories_by_paths(paths)
         prefixes = %w(/pulp/deb/ /pulp/repos/ /pulp/content/)
         relative_paths = []
 
-        # paths == ["/pulp/content/Default_Organization/Library/custom/Test_product/test2"]
+        # paths == ["/pulp/content/Default_Organization/Library/custom/Test_product/test2",
+        #           "/pulp/content/Org/Library/custom/Test_product/test2/%3Fcomp%3Dmain%26rel%3Dstable"]
```
sbernhard, line 168: are you sure that its test2**/**%3F...
I thought its test2?comp ...

app/models/katello/host/content_facet.rb
```diff
@@ -164,10 +164,12 @@ def update_repositories_by_paths(paths)
...
         paths.each do |path|
           if (prefix = prefixes.find { |pre| path.start_with?(pre) })
-            relative_paths << path.gsub(prefix, '')
+            # strip prefix and structured_apt postfix before adding to relative_paths
+            relative_paths << path.sub(prefix, '').sub(/\/?%3F.*/, '')
```
m-bucher, line 172: This fixes an issue we identified during review/testing.

The subman will report `baseurl` of enabled_repositories from `/etc/apt/sources.list.d/rhsm.sources`.
With structured APT, these `baseurl`s includes query-parameters to transfer release and components.
Katello would not be able to find the corresponding repositories without first stripping the query-parameters.

I also changed the `gsub()` that strips the prefix to `sub()`. It is not probable that the prefix is also part of a repository's `relative_path`, but better be safe than sorry :wink:

sbernhard, line 172: Would it be safer to parse the code (decode the "%3F" to "?" etc) and then URI parse it like:
```
=> "blub?comp=main"
irb(main):014:0> uri = URI.parse(a)
=> #<URI::Generic blub?comp=main>
irb(main):015:0> uri.path
=> "blub"
irb(main):016:0> query_params = URI.decode_www_form(uri.query).to_h
=> {"comp"=>"main"}
```

app/models/katello/host/content_facet.rb
m-bucher: On our test-system subman had it encoded in the baseurl and therefore also reported it that way to katello.
We may have to investigate, if this is an issue itself :thinking:

ianballou: The weird characters making it into the path is a Candlepin bug, I don't have a link to it but someone should have it somewhere.

ianballou: It was fixed a while back I believe.

ianballou: https://github.com/candlepin/candlepin/pull/4783

quba42: It is definitely not fixed here http://yum.theforeman.org/candlepin/4.4/.

Also, I still think this particular issue should really be fixed on the client side. It also is an issue independently of the encoding issue, because the mere fact that there are URL params (encoded or not) breaks applicability, because Katello does not recognize the repos.

jeremylenz: The fix should be in Candlepin 4.4.12-1 or later. Also you would need to regenerate your entitlement certificates (subscription-manager identity --regenerate --force).

thanks to @evgeni for info

quba42: Our test system is using `candlepin-4.4.14-1.el8.noarch` and I am seeing subman output with `%3F`:
```
# subscription-manager repos
+----------------------------------------------------------+
    Available Repositories in /etc/yum.repos.d/redhat.repo
+----------------------------------------------------------+
Repo ID:   ATIX_Saltstack_for_Debian_12_Saltstack_for_Debian_12
Repo Name: Saltstack for Debian 12
Repo URL:  https://orcharhino.testdmz.atix/pulp/deb/ATIX/Testing/Composite_Debian_12/custom/Saltstack_for_Debian_12/Saltstack_for_Debian_12/%3Fcomp%3Dmain%26rel%3Dbookworm
Enabled:   0

Repo ID:   ATIX_Debian_12_bookworm-updates_main
Repo Name: bookworm-updates main
Repo URL:  https://orcharhino.testdmz.atix/pulp/deb/ATIX/Testing/Composite_Debian_12/custom/Debian_12/bookworm-updates_main/%3Fcomp%3Dmain%26rel%3Dbookworm-updates
Enabled:   1
```
Running `subscription-manager identity --regenerate --force` has no effect on this.

Is it possible that the bug that was fixed for entitlement certificates was just adjacent to what we are using for structured APT?

Either way, we have made the implementation robust both for encoded and not encoded URL options. :shrug:

app/models/katello/host/content_facet.rb
quba42: We have made it robust for both the encoded and not encoded variants.

m-bucher (approved):
Had a meeting with @quba42 just now and looked through/discussed the latest changes again.
As far as I am concerned, this is good to go now :+1:

app/models/katello/host/content_facet.rb
quba42: Also, I am not sure if I made this properly clear, we need the extra processing here regardless of whether candlepin passes strings with `%3F` or strings with `?`. Either way, subman is reporting a list of repositories to Katello with structured APT options appended to the relative path bit. Katello then compares this list to it's repositories without the structured APT options appended, and so finds no matches. In both cases we need to strip away the structured APT options before the comparison. Which is what we are doing several lines down.

Without this fix Katello never shows any applicable or upgradeable packages, because it does not know what repos to look in.

jeremylenz: I've gotten confirmation from Candlepin team that the URL encoding in your example is now expected:

"Yeah, the changes URL encode the segments, so I would expect that you'd see artifacts of that with the underscore -> %3F example."

So I think we're good here, as long as this code works the way you want it to.

sjha4 (approved):
I did some testing for non-deb content and I was able to run the migrations and able to sync different non-deb repo types, have their content be setup correctly in candlepin and artifacts created in pulp..Pulp DB looked rightly setup and registration workflows went through fine and a subscribed host is able to consume the distributed content. Based on the code, most of the tasks updated look to be running successfully. 

Note that this is not an exhaustive test but based on the code changes, ATIX testing, PRT testing and other reviews, this looks in a good state for merge.

 ACK 🎉

dosas (approved):

quba42:
Thanks a ton to everyone involved!

PR #11059, opened 2024-07-02
Fixes #37617 - Deletion of repository not working from "Products" page when repo in published CV

Thorben-D:
[Redmine issue and reproducer](https://projects.theforeman.org/issues/37617)

When a user tries to delete a repository of a product, which is part of a published CV, an error is raised.
This happens despite the `delete_repo_across_cv`-setting being set to true. 
Deletion from the "details" page of the respective repository works as expected.

It appears like it was forgotten to adapt the `RepositoriesBulkActionsController` to account for the `delete_repo_across_cv` setting.

This PR adresses this by adding the missing logic.

### DIFF
--- a/app/controllers/katello/api/v2/repositories_bulk_actions_controller.rb
+++ b/app/controllers/katello/api/v2/repositories_bulk_actions_controller.rb
@@ -6,7 +6,7 @@ module Katello
     param :ids, Array, :desc => N_("List of repository ids"), :required => true
     def destroy_repositories
       deletion_authorized_repositories = @repositories.deletable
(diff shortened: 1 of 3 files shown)


sbernhard:
@chris1984 Can you have a look at this PR? Actually, I'm not able to re-run the failed test as I can in foreman project but I think the failed test is not related to this PR.

chris1984:
> @chris1984 Can you have a look at this PR? Actually, I'm not able to re-run the failed test as I can in foreman project but I think the failed test is not related to this PR.

Rekicked

PR #11061, opened 2024-07-05
Fixes #37632 - Use getDocsURL instead of useKatelloDocUrl

ekohl:

In d400e3996f068841259b6d4274288f547e7a5f06 a useKatelloDocUrl helper was introduced. This should have never been introduced in the first place and instead should have been implemented in Foreman itself. Since Foreman 3.9.1 there is such an API so this should be used.

Downstream concern: https://github.com/RedHatSatellite/foreman_theme_satellite should also point to the correct URL.

* Open page
* Click documentation link
* Verify you end up on the correct documentation


webpack/scenes/ContentViews/Details/Repositories/ContentViewRepositories.js
```diff
@@ -4,6 +4,7 @@ import React, {
   useState,
 } from 'react';
 import { FormattedMessage } from 'react-intl';
+import { getDocsURL } from 'foremanReact/common/helpers'
```
ianballou, line 7:
```suggestion
import { getDocsURL } from 'foremanReact/common/helpers';
```

ekohl: Updated and now the tests are green.

PR #11063, opened 2024-07-05
Fixes #35582 - properly send invalid ULN ACS remote errors

ianballou:
Throw a proper validation error if an ACS has `uln://` in it. Pulp does not support `uln://` ACSs.
Also skip ULN repositories for simplified ACSs.
Some refactoring was done as well to make the code more readable, and to make sure that only ACS-compatible repositories get used in simplified ACSs.

I considered asking Pulp for ULN ACS support, but this will need to be done first anyway to stop ugly errors from hitting users.

1) Try making simplified ACSs for products with yum repositories that have `uln://` in the upstream URL.
  -> Check that changing the URL to/from ULN adds/removes the repo from the simplified ACS. You can check this in the Rails console.
2) Try to make a Custom ACS with a ULN URL. The UI should disallow this, but the API and likely Hammer will not. Ensure no ACS gets half-created and that the error message is readable. It should say that the URL is invalid.

TODO
- [x] Write tests.

### DIFF
--- a/app/models/katello/repository.rb
+++ b/app/models/katello/repository.rb
@@ -121,6 +121,7 @@ module Katello
     before_validation :set_container_repository_name, :unless => :skip_container_name?
 
     scope :has_url, -> { joins(:root).where.not("#{RootRepository.table_name}.url" => nil) }
+    scope :not_uln, -> { joins(:root).where("#{RootRepository.table_name}.url NOT LIKE 'uln%'") }
     scope :on_demand, -> { joins(:root).where("#{RootRepository.table_name}.download_policy" => ::Katello::RootRepository::DOWNLOAD_ON_DEMAND) }
     scope :immediate, -> { joins(:root).where("#{RootRepository.table_name}.download_policy" => ::Katello::RootRepository::DOWNLOAD_IMMEDIATE) }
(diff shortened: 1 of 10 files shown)


app/models/katello/repository.rb
```diff
@@ -121,6 +121,7 @@ class Repository < Katello::Model
     before_validation :set_container_repository_name, :unless => :skip_container_name?
 
     scope :has_url, -> { joins(:root).where.not("#{RootRepository.table_name}.url" => nil) }
+    scope :not_uln, -> { joins(:root).where("#{RootRepository.table_name}.url NOT LIKE 'uln://%'") }
```
sjha4, line 124: Most places in code we only use `start_with?('uln')` and here we have uln:// .. Not really important but we could make it consistent.

sjha4:
I am seeing one issue. The following workflow has the ULN validation applied too late:
1. Create a product > regular yum repo
2. Create simplified ACS with the product.
3. Create a ULN repo inside the product above. That tries to create a ULN acs and errors out during repo create.
Screenshot from 2024-07-18 13-41-36

ianballou:
@sjha4 the issues should be taken care of now. I added a test too to check that newly created ULN repos don't get added to the ACS.

sjha4 (approved):
Looks good..Don't seem to have access to packit logs so not sure what's going on there..The PR works great..APJ 👍🏼

ianballou:
Failures might be related to the recent pulp rpm bindings bump

PR #11065, opened 2024-07-08
xFix Minitest/GlobalExpectations cop

archanaserver:
Base on: https://github.com/Katello/katello/pull/11009
This pull request addresses the Minitest/GlobalExpectations RuboCop offense by replacing global expectation methods with the recommended non-global expectation methods.

https://docs.rubocop.org/rubocop-minitest/cops_minitest.html#minitestglobalexpectations

 Verified that the changes do not impact existing functionality or cause syntax errors.

bundle exec rubocop


ianballou (commented):
It looks like the underscore is confusing gettext for `katello/spec/models/host_collection_spec.rb`.  Perhaps because the underscore is also used to mark translations? The cop guide seems to have some examples without underscores.

ianballou:
> > It looks like the underscore is confusing gettext for `katello/spec/models/host_collection_spec.rb`. Perhaps because the underscore is also used to mark translations? The cop guide seems to have some examples without underscores.
> 
> I have followed different style for that specific file you mentioned, but not sure if this is good to follow different style for different files. wdyt?

It might be best to avoid the underscores entirely if possible since that usually represents translations. I don't feel super strongly about it though since it's in tests only as far as I see.

archanaserver:
> > > It looks like the underscore is confusing gettext for `katello/spec/models/host_collection_spec.rb`. Perhaps because the underscore is also used to mark translations? The cop guide seems to have some examples without underscores.
> > 
> > 
> > I have followed different style for that specific file you mentioned, but not sure if this is good to follow different style for different files. wdyt?
> 
> It might be best to avoid the underscores entirely if possible since that usually represents translations. I don't feel super strongly about it though since it's in tests only as far as I see.

Done! @ianballou

ianballou:
Looks like there are two failures left, I'm seeing `katello/spec/models/provider_spec.rb` mentioned.

archanaserver:
> Looks like there are two failures left, I'm seeing `katello/spec/models/provider_spec.rb` mentioned.

@ianballou not sure what this failure now, can you help me in this?

spec/models/host_collection_spec.rb
```diff
@@ -22,44 +22,44 @@ module Katello
...
+        value(grp2.new_record?).must_equal(false)
+        value(HostCollection.where(:name => "TestHostCollection").count).must_equal(2)
       end
 
       it "should not allow unlimited_hosts=false and max_hosts to be nil at the same time" do
         create = lambda do
           HostCollection.create!(:name => "TestHostCollection", :organization => @org, :unlimited_hosts => false)
         end
-        create.must_raise(ActiveRecord::RecordInvalid)
+        value { create }.must_raise(ActiveRecord::RecordInvalid)
```
ianballou, line 47: I think this is the issue:
```suggestion
        value(create).must_raise(ActiveRecord::RecordInvalid)
```

ianballou (changes requested):

ianballou:
The latest test failures are pretty  odd -- need to see what's going on there. Maybe in the meantime rebase if you haven't.

ianballou:
Please rebase once https://github.com/Katello/katello/pull/11076 is merged.

ianballou (approved):
Looks good, thanks @archanaserver !

archanaserver:
Thanks for the review @ianballou!

PR #11071, opened 2024-07-10
Fixes #37634 - Prevent users from editing container push repositories

qcjames53:
Aside from regular container push operations, users are now no longer able to update fields of container push root repositories and instance repositories. The repository edit page has all editable fields disabled for container push repos.

When Katello needs to internally update a root or instance repo of a container push repository, the developer must first run `root_repo.update!(allow_updates: true)` and then run `root_repo.update!(allow_updates: false)` after operations are complete. If the update operations fail midway, the root and instance repos will remain editable by the end user.

As of now, we only update root and instance repo properties during repo push commands. If the user's repo push fails, the repo remaining editable is moot anyways.

1. Download some container images. Create a product with a few regular repositories of different types.
2. `podman login` to Katello. Push the container to this product (`podman push <image> <url>/<org label lowercase>/<prod label lowercase>/<chosen name>`). The push should complete without error.
3. Push to the same location again. Updating the repo should work without error.
4. Try to edit this new repo using hammer / the API. It should fail with error message.
5. Try to edit this repo using the UI. All fields should have editing disabled.
6. Try to edit the new root repo and instance repo using rake console. Validation should fail with a clear error.
7. Editing the other regular non-container-push repositories should work normally without any errors.

### DIFF
--- a/app/controllers/katello/api/registry/registry_proxies_controller.rb
+++ b/app/controllers/katello/api/registry/registry_proxies_controller.rb
@@ -383,6 +383,7 @@ module Katello
           :not_found
         )
       end
+
       instance_repo.update!(version_href: latest_version_href)
       # The Pulp repository should not change after first creation
       if root_repository.repository_references.empty?
--- a/app/models/katello/repository.rb
+++ b/app/models/katello/repository.rb
@@ -32,6 +32,8 @@ module Katello
 
     EXPORTABLE_TYPES = [YUM_TYPE, FILE_TYPE, ANSIBLE_COLLECTION_TYPE, DOCKER_TYPE, DEB_TYPE].freeze
 
+    ALLOWED_UPDATE_FIELDS = ['version_href', 'last_indexed'].freeze
+
     define_model_callbacks :sync, :only => :after
 
     belongs_to :root, :inverse_of => :repositories, :class_name => "Katello::RootRepository"
@@ -119,6 +121,7 @@ module Katello
 
     before_validation :set_pulp_id
     before_validation :set_container_repository_name, :unless => :skip_container_name?
+    before_update :prevent_updates, :unless => :allow_updates?
 
     scope :has_url, -> { joins(:root).where.not("#{RootRepository.table_name}.url" => nil) }
     scope :on_demand, -> { joins(:root).where("#{RootRepository.table_name}.download_policy" => ::Katello::RootRepository::DOWNLOAD_ON_DEMAND) }
@@ -179,7 +182,7 @@ module Katello
              :deb_components, :deb_architectures, :ssl_ca_cert_id, :ssl_ca_cert, :ssl_client_cert, :ssl_client_cert_id,
              :ssl_client_key_id, :os_versions, :ssl_client_key, :ignorable_content, :description, :include_tags, :exclude_tags,
              :ansible_collection_requirements, :ansible_collection_auth_url, :ansible_collection_auth_token,
-             :http_proxy_policy, :http_proxy_id, :to => :root
+             :http_proxy_policy, :http_proxy_id, :prevent_updates, :to => :root
 
     delegate :content_id, to: :root, allow_nil: true
     delegate :repository_type, to: :root
@@ -1009,6 +1012,7 @@ module Katello
         repository_type.index_additional_data_proc&.call(self)
       end
       self.update!(last_indexed: DateTime.now)
+
       true
     end
 
@@ -1048,6 +1052,12 @@ module Katello
         manifest.destroy
       end
       DockerMetaTag.cleanup_tags
+    end
+
+    def allow_updates?
+      # allow the update if this repo is not in the default view
+      return true unless in_default_view?
+      root.allow_updates?(::Katello::Repository::ALLOWED_UPDATE_FIELDS)
     end
 
     apipie :class, desc: "A class representing #{model_name.human} object" do
--- a/app/models/katello/root_repository.rb
+++ b/app/models/katello/root_repository.rb
@@ -46,6 +46,8 @@ module Katello
     MIRRORING_POLICY_COMPLETE = 'mirror_complete'.freeze
     MIRRORING_POLICIES = [MIRRORING_POLICY_ADDITIVE, MIRRORING_POLICY_COMPLETE, MIRRORING_POLICY_CONTENT].freeze
 
+    ALLOWED_UPDATE_FIELDS = ['updated_at', 'content_id'].freeze
+
     belongs_to :product, :inverse_of => :root_repositories, :class_name => "Katello::Product"
     has_one :provider, :through => :product
 
@@ -115,6 +117,7 @@ module Katello
       }
 
     validates :container_push_name_format, inclusion: { in: ['label', 'id'].freeze, allow_nil: true}
+    before_update :prevent_updates, :unless => :allow_updates?
 
     scope :subscribable, -> { where(content_type: RootRepository::SUBSCRIBABLE_TYPES) }
     scope :skipable_metadata_check, -> { where(content_type: RootRepository::SKIPABLE_METADATA_TYPES) }
@@ -453,6 +456,22 @@ module Katello
       else
         self.arch == "noarch" ? nil : self.arch
       end
+    end
+
+    def allow_updates?(additional_allowed_fields = [])
+      # allow updates for non-container-push repos, repos not in default view, and
+      # repos with a library instance
+      return true unless is_container_push && library_instance.present?
+
+      # let updates that contain ONLY allowed strings through
+      allowed_fields = ::Katello::RootRepository::ALLOWED_UPDATE_FIELDS + additional_allowed_fields
+      return true if (changed - allowed_fields).empty?
+
+      false
+    end
+
+    def prevent_updates
+      fail _("Cannot update properties of a container push repository")
     end
 
     apipie :class, desc: 'A class representing Repository object' do
--- a/app/views/katello/api/v2/repositories/show.json.rabl
+++ b/app/views/katello/api/v2/repositories/show.json.rabl
@@ -28,6 +28,8 @@ glue(@resource.root) do
   attributes :http_proxy_name
   attributes :retain_package_versions_count
   attributes :metadata_expire
+  attributes :allow_updates? => :allow_updates
+  attributes :is_container_push? => :is_container_push
 
   node :http_proxy do
     attributes :id => @resource.root&.http_proxy&.id, :name => @resource.root&.http_proxy&.name, :policy => @resource.root&.http_proxy_policy
--- a/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/views/repository-info.html
+++ b/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/views/repository-info.html
@@ -8,7 +8,7 @@
       <dt translate>Name</dt>
       <dd bst-edit-text="repository.name"
           on-save="save(repository)"
-          readonly="product.redhat || denied('edit_products', product)">
+          readonly="product.redhat || denied('edit_products', product) || !repository.allow_updates || repository.is_container_push">
       </dd>
 
       <dt translate>Label</dt>
@@ -17,7 +17,7 @@
       <dt translate>Description</dt>
       <dd bst-edit-textarea="repository.description"
 	  on-save="save(repository)"
-	  readonly="product.redhat || denied('edit_products', product)">
+	  readonly="product.redhat || denied('edit_products', product) || !repository.allow_updates || repository.is_container_push">
       </dd>
 
       <dt translate>Backend Identifier</dt>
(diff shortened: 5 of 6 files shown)


ianballou (commented):
The UI bits are looking great, you can't edit anything and you can't sync the repositories :+1:

Hammer is also looking good:
```
[vagrant@centos7-hammer-devel hammer-cli-foreman]$ hammer repository update --id 30 --product-id 1 --name potato
...
Could not update the repository:
  Cannot update properties of a container push repository
```
---

Main workflows aside, this is breaking content view publishing since the new CV instance repositories are also getting blocked from being updated at CV publish time.

The validation should probably only affect repositories that are `in_default_view`, which are the repositories that a user can actually edit via the UI/API/CLI.

Some things to consider:
- What happens if the web server crashes between enabling and disabling edits on the repository?  The code isn't atomic, so the database value could go out-of-sync. How would a user recover from that?
- Could we change the validation to always permit editing certain fields? It looks like you are only locking/unlocking so Katello can change `version_href` and perform content indexing. Users can't change either of those directly through the UI/API/CLI -- those fields are only changed under-the-hood. We could safely allow them to be edited always.
- Or, could we use validation skipping at key areas in our code to get around needing a lock in the database?

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -383,13 +383,16 @@ def save_pulp_push_repository_href
...
+
+      enable_container_push_repo_update(root_repository)
       instance_repo.update!(version_href: latest_version_href)
       # The Pulp repository should not change after first creation
       if root_repository.repository_references.empty?
         ::Katello::Pulp3::RepositoryReference.where(root_repository_id: instance_repo.root_id,
                                                     content_view_id: instance_repo.content_view.id,
                                                     repository_href: pulp_repo_href).create!
       end
+      disable_container_push_repo_update(root_repository)
```
jeremylenz, line 395: Especially since `update!` raises errors if anything is invalid, what do you think about putting the `disable_container_push_repo_update` line inside an `ensure` block? This way if that happens, you won't be left with an incorrectly-update-enabled container push repo.

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -428,6 +431,14 @@ def save_push_repo_hrefs
       save_pulp_push_distribution_href(pulp_repo_href)
     end
 
+    def enable_container_push_repo_update(root_repo)
+      root_repo.update!(allow_updates: true)
```
jeremylenz, line 435: The `!` means it will throw an error if anything goes wrong (rather than returning false). That's not what we want here-- I'd go the opposite direction instead:

```suggestion
      root_repo.update_columns(allow_updates: true)
```

Using `update_columns` will skip validations and ensure that we aren't foiled by an error.

app/controllers/katello/api/registry/registry_proxies_controller.rb
```diff
@@ -428,6 +431,14 @@ def save_push_repo_hrefs
...
+    end
+
+    def disable_container_push_repo_update(root_repo)
+      root_repo.update!(allow_updates: false)
```
jeremylenz, line 439: same as above

app/models/katello/repository.rb
```diff
@@ -992,6 +993,8 @@ def index_content(options = {})
       # some times we want to force fetch all data even for non-library repos.
       # Use the full_index for those times
 
+      root.update!(allow_updates: true) if root.is_container_push?
```
jeremylenz, line 996: Same comment as above. Now that I've commented the same thing thrice, maybe this would be better as a method on `root`, too - a method that does the `root.update_columns` for you. and maybe call it `set_allow_updates`

app/models/katello/repository.rb
```diff
@@ -1007,6 +1010,9 @@ def index_content(options = {})
         repository_type.index_additional_data_proc&.call(self)
       end
       self.update!(last_indexed: DateTime.now)
+
+      root.update!(allow_updates: false) if root.is_container_push?
```
jeremylenz, line 1014: same comment as above

app/models/katello/repository.rb
```diff
@@ -1048,6 +1054,14 @@ def remove_docker_content(manifests)
       DockerMetaTag.cleanup_tags
     end
 
+    def allow_updates?
+      root.allow_updates
+    end
+
+    def prevent_updates
+      root.prevent_updates
+    end
```
jeremylenz, line 1063: Cool Rails trick: If you add a line to the top of this like

```rb
delegate :allow_updates?, :prevent_updates, to: :root 
```

it's equivalent to this code you have here. There's tons of examples in Katello so I would lean towards using it.

app/models/katello/root_repository.rb
```diff
@@ -455,6 +456,17 @@ def format_arches
       end
     end
 
+    def allow_updates?
+      # Allow updates for non-container-push repos and for updating exclusively the allow_updates flag
+      return true if allow_updates
+      return true if changed.size == 2 && will_save_change_to_allow_updates?
```
jeremylenz, line 462: I don't see this defined anywhere. Is this a Rails helper?

jeremylenz (commented):
Looking great @qcjames53!

I didn't test but wanted to leave some code comments below

app/models/katello/root_repository.rb
qcjames53: Yes, that's a rails helper. I forget the exact syntax but would using changed.contains('allow_updates') be more readable? 

There's also a chance this will be moot if Ian's suggestion about allowing certain backend fields to be updated pans out.

jeremylenz: > would using changed.contains('allow_updates') be more readable?

I would've understood that on the first try :stuck_out_tongue:

ianballou:
To add to my previous review -- we may also need to support deleting container push contents in the future, along with other potential changes. The lock has the disadvantage of needing to lock/unlock it each time you need to edit the repo, so it might become cumbersome as we add bits on container push repos that you can edit.

qcjames53:
I just pushed a few changes. Sorry about getting this in so late today; I was really hoping to be finished sooner. I was able to remove the locks with some tweaks for bypassing update param checks and the following params explicitly allowed for repo updates:
```
root repo: updated_at, content_id
instance repo: version_href, last_indexed
```

Here's the sad news. The repo details page has reverted to edit icons being visible for some reason. I need a little more time to look into this. The fields still display an error when saving but I think it would be far better if they were disabled instead.

qcjames53:
> The UI bits are looking great, you can't edit anything and you can't sync the repositories 👍
> 
> Hammer is also looking good:
> 
> ```
> [vagrant@centos7-hammer-devel hammer-cli-foreman]$ hammer repository update --id 30 --product-id 1 --name potato
> ...
> Could not update the repository:
>   Cannot update properties of a container push repository
> ```
> 
> Main workflows aside, this is breaking content view publishing since the new CV instance repositories are also getting blocked from being updated at CV publish time.
> 
> The validation should probably only affect repositories that are `in_default_view`, which are the repositories that a user can actually edit via the UI/API/CLI.
> 
> Some things to consider:
> 
> * What happens if the web server crashes between enabling and disabling edits on the repository?  The code isn't atomic, so the database value could go out-of-sync. How would a user recover from that?
>   
>   * Likewise, Foreman has parallelism. Might there be a circumstance where the lock is being accessed by two puma workers at once?
> * Could we change the validation to always permit editing certain fields? It looks like you are only locking/unlocking so Katello can change `version_href` and perform content indexing. Users can't change either of those directly through the UI/API/CLI -- those fields are only changed under-the-hood. We could safely allow them to be edited always.
> * Or, could we use validation skipping at key areas in our code to get around needing a lock in the database?

This should be ok now. The validation is skipped for root repos if they are either not a container push repo or if they aren't in the default view or if they don't have at least one instance repo. Instance repo validation is skipped if their root repo fits any of the above categories or if they themselves are not in the default view.

app/controllers/katello/api/registry/registry_proxies_controller.rb
qcjames53: Thank you for the input with this, Jeremy. I was able to work around the lock entirely with the latest push so I'm marking this as resolved.

app/controllers/katello/api/registry/registry_proxies_controller.rb
qcjames53: Same as above.

app/models/katello/repository.rb
qcjames53: Same as above. Thanks for the thoughts.

app/models/katello/repository.rb
qcjames53: The same as above.

app/models/katello/repository.rb
qcjames53: Appreciate this trick. I implemented it for `prevent_updates` and it works great :)

app/models/katello/root_repository.rb
qcjames53: This no validation longer exists so I'll mark as resolved.

app/models/katello/repository.rb
```diff
@@ -119,6 +119,7 @@ class Repository < Katello::Model
 
     before_validation :set_pulp_id
     before_validation :set_container_repository_name, :unless => :skip_container_name?
+    before_update :prevent_updates, :unless => :allow_updates?
```
jeremylenz, line 122: This line is absolutely hilarious and I love it :smile:

app/models/katello/root_repository.rb
```diff
@@ -455,6 +456,22 @@ def format_arches
       end
     end
 
+    def allow_updates?(additional_allowed_fields=[])          
+      # allow updates for non-container-push repos, repos not in default view, and
+      # repos with a library instance
+      return true unless is_container_push && library_instance.present?
+
+      # let updates that contain ONLY allowed strings through
+      allowed_fields = ['updated_at', 'content_id'] + additional_allowed_fields
```
jeremylenz, line 465: Probably better to put these in a constant

qcjames53: I had the same thought but ruby was very mad with the syntax as-is. It said I couldn't use this data structure specifically. I'll see if there's another way to achieve the result.

jeremylenz: hm, what specific cop was it? Often the cop documentation has examples of how to do it correctly.

jeremylenz: The only failing cop I see is around whitespace:
```
app/models/katello/root_repository.rb:459:49: C: Layout/SpaceAroundEqualsInParameterDefault: Surrounding space missing in default value assignment.
    def allow_updates?(additional_allowed_fields=[])          
                                                ^
app/models/katello/root_repository.rb:459:53: C: Layout/TrailingWhitespace: Trailing whitespace detected.
    def allow_updates?(additional_allowed_fields=[])  
```
looks like you just need to change it to `additional_allowed_fields = []`

qcjames53: Thanks Jeremy :)

The newest push includes the constants (implemented correctly this time) and fixes my Rubocop errors.

qcjames53:
I just force pushed some changes that fix the UI. I had a small oversight where running 'allow_updates' without changing anything would return 'true' in all cases. Container-push-type repos will now return false in this case, meaning the edit tooltips in the UI should once again be disabled.

app/models/katello/repository.rb
ianballou: Ruby art

app/models/katello/repository.rb
```diff
@@ -1050,6 +1054,12 @@ def remove_docker_content(manifests)
       DockerMetaTag.cleanup_tags
     end
 
+    def allow_updates?
+      # allow the update if this repo is not in the default view
+      return true unless in_default_view?
+      root.allow_updates?(ALLOWED_UPDATE_FIELDS)
```
ianballou, line 1060: It's a little bit confusing that a root method is being called to check instance repo attributes, but the only alternatives are to either duplicate code or make some separate helper file, so I think I prefer this.

ianballou (commented):
Tested, working well. Tried to edit the repos, sync them, run reclaim space, and publish them in CVs. Everything checked out.
Looks like the tests need an update.
@jeremylenz in the meantime did you have any final comments ?

app/models/katello/repository.rb
jeremylenz, line 1060: Even though it isn't strictly needed, it may be more clear to be explicit about which constant we're using, since they're both called `ALLOWED_UPDATE_FIELDS`.

```suggestion
      root.allow_updates?(::Katello::Repository::ALLOWED_UPDATE_FIELDS)
```

jeremylenz (approved):
APJ, one final thought below

app/models/katello/repository.rb
ianballou: Nice, +1 to that.

qcjames53:
I fixed up the failing test. Provided everything passes and you're happy, I think we are OK to merge.

jeremylenz (approved):

jeremylenz:
I don't know what to make of the test failures, but code LGTM :+1:

qcjames53:
I'm in the same boat. I don't believe I would have affected anything pulpcore related?

ianballou (changes requested):
New container repo creation is failing. I did some debugging and found the following:
1) The RootRepository creation passes because there is no library instance yet
2) The Repository creation fails because the root is a `push` repo but `changed` is empty.
  -> I'm not sure why changed is empty, but I wonder if it's because `changed` is in reference to the RootRepository rather than the Repository itself because the validation method is in RootRepository.

qcjames53:
This should be fixed now; I just pushed another update. This time we allow updates that contain no changes but also pass the is_container_push parameter to the repository overview page and disable UI editing based on that boolean.

ianballou (commented):
Looking good and working well! Rubocop is unhappy.

qcjames53:
Sorry about that; I thought I checked. That was actually surprisingly tricky to fix for some reason. It took a good 15 minutes.

PR #11072, opened 2024-07-12
Fixes #37530 - Very long CV names/labels display weirdly on CV UI

sjha4:
Adds truncated name/labels to UI where necessary.
Truncate truncates by default to 50 chars which seems like enough information for a CV name. The full name is seen on the CV details tab.
Create a CV with a loooong name
Navigate around the CV UI and make sure the UI doesn't get weird with the long way and truncates and displays CV info properly.


chris1984:

chris1984 (commented):
Two small comments discussed on slack, around areas it's missing

chris1984 (approved):
ACK, works good. APJ. Looks like there are some eslint complaints,

PR #11074, opened 2024-07-17
Fixes #37654 - Use custom pre/post snippets for errata templates

sbernhard:
Before and after Errata installation customers want to run certain tasks like taking a snapshot, write the current state to a 3rd party tool, etc.


sbernhard:
Rebased but some tests still fails. I don't think they are related.

chris1984 (approved):

PR #11075, opened 2024-07-17
Fixes #37656 - Rename "proxy" to "Smart Proxy" in settings.

Thorben-D:
Under Administer > Settings > Content > Prefer registered through proxy for remote execution, the Smart Proxy is referred to by "proxy". "Smart Proxy" would be the correct term here.


maximiliankolb (approved):
Thanks, LGTM. IMHO a useful improvement to differentiate between HTTP proxies and Smart Proxies.

maximiliankolb (approved):
still good ^^

chris1984 (approved):
ACK

PR #11077, opened 2024-07-18
Fixes #37658 - explicitly set type to CUSTOM_CDN_TYPE

evgeni:

When updating CdnConfiguration from the old, deprecated redhat_repository_url parameter, explicitly set the CdnConfiguration type to CUSTOM. If we don't do this, the URL gets reset to redhat.com automatically.

None. The param is deprecated, but I expect it to work until its removed.

Update redhat_repository_url via a PUT request to the org endpoint.
Verify the data changed in the DB -- not only the task was successful (this is what the unit test test and I didn't see a good way to assert the DB state)

### DIFF
--- a/app/controllers/katello/api/v2/organizations_controller.rb
+++ b/app/controllers/katello/api/v2/organizations_controller.rb
@@ -59,7 +59,7 @@ module Katello
     param_group :resource
     def update
       if params[:redhat_repository_url]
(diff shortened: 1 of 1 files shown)


Griffin-Sullivan (approved):

chris1984 (approved):
ACK

PR #11079, opened 2024-07-18
Fixes #37633 - Newly published CV version shows need_published as true

sjha4:
Scope audit query to auditable_type content views
For tracking audits on repository additions/removals on CV, we were not scoping the query to content view objects.
On a fresh box:
1. Create a custom repo.
2. Create a CV and publish
3. You'll see the ⬆️ next to newly published version and cv.needs_publish? will return true in console.
4. With this change, the version will correctly show as not needing publish.

### DIFF
--- a/app/models/katello/content_view.rb
+++ b/app/models/katello/content_view.rb
@@ -864,7 +864,9 @@ module Katello
       }
 
       table = Audit.arel_table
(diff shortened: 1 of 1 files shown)


chris1984 (approved):
ACK, works good and fixes https://issues.redhat.com/browse/SAT-26493

PR #11057, opened 2024-07-01
Fixes #37622 - Bump pulp-rpm to 3.26.1

ianballou:
Bumps pulp-rpm to 3.26.1 and adds VCRs.

Other PRs:
https://github.com/theforeman/pulpcore-packaging/pull/1039
https://github.com/theforeman/foreman-packaging/pull/10981
https://github.com/theforeman/foreman-packaging/pull/10982

PR #11081, opened 2024-07-18
[CP] 4.13.1

parthaa:


chris1984 (approved):
APJ

jeremylenz:
looks like 'React Tests' still uses the old (broken) GH workflow. But there are no Ruby changes in these commits AFAIK, so it's fine to ignore that check.

Odilhao:
@ianballou @parthaa can we include this https://github.com/Katello/katello/pull/11057? At some point we will need to cherrypick the new pulp-rpm into 3.49 branch

ianballou:
@Odilhao yeah we should also include that bindings bump commit here.

ianballou:
@parthaa let me know if CPing in  that pulp-rpm update commit causes VCR errors, if it does we may need to do some re-recording.

parthaa:
> @parthaa let me know if CPing in that pulp-rpm update commit causes VCR errors, if it does we may need to do some re-recording.

CP worked fine. make sure https://github.com/Katello/katello/pull/11081/files#diff-21cb8426e694930953a1fb66d2f1e4c0c4021bcadd760d9582423094a4dcaace looks correct.

ianballou (approved):

PR #11083, opened 2024-07-19
Fixes #37666 - Fix Candlepin reset in reset rake task

chris1984:
* With the recent changes to Candlepin and `cpdb` we forgot to update the reset rake task
```bash
basically, instead of cpdb --create --schema-only you need to be now running cpdb --create  (--schema-only no longer exists), 

instead of cpdb --update the new command is now cpdb --update-schema  (although, cpdb --update
may still work, because bash knows --update is start of --update-schema)
```

* Instead of using `runuser postgres` cpdb can create and drop the database for us, so switched to that for all native candlepin db actions.

* Make sure the Candlepin database gets recreated still

* Check out PR
* Reset your devel box `bundle exec rake katello:reset --trace`
* Verify you do not see the error message anymore

### DIFF
--- a/lib/katello/tasks/reset.rake
+++ b/lib/katello/tasks/reset.rake
@@ -30,10 +30,9 @@ namespace :katello do
       puts "\e[33mStarting Candlepin Reset\e[0m\n\n"
 
       system(service_stop.gsub("%s", 'tomcat'))
-      system("sudo runuser - postgres -c 'dropdb candlepin'")
-      system("sudo runuser - postgres -c 'createdb candlepin'")
-      system("sudo /usr/share/candlepin/cpdb --create --schema-only")
-      system("sudo /usr/share/candlepin/cpdb --update")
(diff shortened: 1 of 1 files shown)


ianballou (dismissed):
Working well for me, just one rubocop issue.

ianballou (approved):
Looks good and works!

evgeni:
Out of curiosity, why don't you use the reset feature of the installer? It has all the same bells and whistles and is actually tested regularly to do the right thing?

chris1984:
> Out of curiosity, why don't you use the reset feature of the installer? It has all the same bells and whistles and is actually tested regularly to do the right thing?

@evgeni I can make a PR to update the katello developer guide on the foreman.org to use the installer instead? I still think we should keep this one as well, because there are times for example on 1 dev box I have where it's setup for provisioning and if I don't pass all the prov params, it does not remember them. I only see this with the katello-devel scenario not on production.

PR #11084, opened 2024-07-19
Fixes #37668 - enable container push by default

ianballou:
Removes the necessity of having a setting for pushing container images into Katello. 

Requires https://github.com/Katello/katello/pull/11071

Try pushing a container without having the old allow_push setting, or any `:container_image_registry` setting at all.


sjha4 (approved):
Ack.. 👍🏼

PR #11089, opened 2024-07-23
Fixes #37674 - Change All Hosts kebab menu to match design [Katello]

jeremylenz:
Requires https://github.com/theforeman/foreman/pull/10252

Rearrange the All Hosts kebab menu to match existing designs.
- [x] Main kebab with flyout menu
- [x] Single-host table row kebab

The fly-out menu in the design required a switch in Foreman from a `<Dropdown>` to a `<Menu>` component, which does not handle its own open/closed state. Therefore, I had to add `menuOpen` to the `ForemanActionsBarContext` so that Katello can close its own submenu when the user clicks on something.

Also, previously we had a solution in place for when users have the "Any Organization" context selected, which would disable menu items with a tooltip explaining why they are disabled. This was accomplished through `Dropdown`'s `isAriaDisabled` and `tooltip` properties, neither of which are available on `Menu`. So I had to come up with a different solution. I ended up using an icon and text in the menu item description.

For the single-host table row kebab menu, Katello must contribute 'Change content view environments' (renamed from 'Lifecycle environment' in the mockup.) Since the table row actions only take a plain object and not a React component (thanks, Patternfly..), I found that accessing the function to open the BulkChangeHostCVModal from the table row action was simply too hard, and React yelled at me. After much strife I went back to the drawing board and discovered that the easiest way to accomplish this was via a URL redirect to the host details page. 

make sure everything works as expected in the menu

### DIFF
--- a/webpack/components/extensions/Hosts/ActionsBar/index.js
+++ b/webpack/components/extensions/Hosts/ActionsBar/index.js
@@ -1,18 +1,37 @@
 import React, { useContext, useEffect } from 'react';
+import PropTypes from 'prop-types';
 import { useDispatch } from 'react-redux';
-import { DropdownItem } from '@patternfly/react-core';
+import { Menu, MenuItem } from '@patternfly/react-core';
+import { BanIcon } from '@patternfly/react-icons';
 import { translate as __ } from 'foremanReact/common/I18n';
 import { foremanUrl } from 'foremanReact/common/helpers';
 import { ForemanHostsIndexActionsBarContext } from 'foremanReact/components/HostsIndex';
 import { useForemanModal } from 'foremanReact/components/ForemanModal/ForemanModalHooks';
 import { addModal } from 'foremanReact/components/ForemanModal/ForemanModalActions';
 import { useForemanOrganization } from 'foremanReact/Root/Context/ForemanContext';
+import './ActionsBar.scss';
+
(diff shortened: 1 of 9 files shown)


jeremylenz:
need to work on failing tests, should be ready for review otherwise

webpack/components/extensions/Hosts/ActionsBar/index.js
```diff
@@ -40,45 +59,51 @@ const HostActionsBar = () => {
...
-
-    </>
+    <MenuItem
+      itemId="content-flyout-item"
+      key="content-flyout"
+      isDisabled={selectedCount === 0}
+      flyoutMenu={(
+        <Menu ouiaId="content-flyout-menu" onSelect={() => setMenuOpen(false)}>
+          <MenuItem
+            itemId="bulk-packages-wizard-dropdown-item"
```
MariaAga, line 69: under the Menu should be `<MenuContent> <MenuList>` wrappers to avoid the console error

jeremylenz: awesome, it works!!

jeremylenz:
Updated to remove console errors when you go into the flyout menu.

MariaAga (approved):
Works great!

chris1984 (approved):
ACK looks good and works well. Thanks @jeremylenz

